PR #2308, opened 2026-08-16
fix(simulation/newton): advertise set_timestep under the parameter name it accepts

cagataycali:
## Problem

`NewtonSimEngine.describe()` advertised `set_timestep` as `"(dt: float) -> dict"`. The method has always been `set_timestep(timestep: float)`. So a caller that reads the discovery surface and calls what it was told to call:

```python
sim.set_timestep(dt=0.002)
# TypeError: NewtonSimEngine.set_timestep() got an unexpected keyword argument 'dt'
```

That is a dead end from the one surface whose documented purpose is letting a caller learn the contract without guessing, and the error names the keyword rather than the surface that supplied it. Everything else already spells it `timestep`: the MuJoCo backend's `describe()`, `strands_robots/simulation/mujoco/tool_spec.json`, and the parameter table in `docs/simulation/overview.md`. Only the Newton discovery string drifted, and `dt` appears in the Newton module solely as a local (`dt = self._world.timestep / self.substeps`).

describe set_timestep before and after

## Why it was not caught

The invariant is already stated and already enforced. `_assert_advertised_params_are_real` in `tests/simulation/test_sim_engine_describe_discovery.py` exists precisely so that "a rename that dropped or renamed a parameter without updating the hardcoded describe() string" cannot ship. It runs on a live engine, so it runs only where an engine can be constructed: the ABC (through a minimal concrete subclass) and MuJoCo. Newton and Isaac need a simulator runtime that cannot be installed in the same environment, so their hardcoded signature strings were never compared against anything.

Their classes do import without that runtime. That is what makes the gap closable: the strings can be read from the source and the signatures resolved on the imported class, with nothing constructed.

## Change

1. `describe()["methods"]["set_timestep"]` now reads `"(timestep: float) -> dict (seconds; takes effect on the next step)"`, matching the method and naming the unit the way the MuJoCo entry does.
2. The parameter-agreement rule now covers every backend that overrides `describe()`, by reading the advertised strings from source and resolving each signature on the imported class. This also covers strings sitting on branches a live `describe()` call happens not to take.

Both directions are pinned, because a rename presents as each of them: an advertised keyword the method does not accept, and a required parameter the string never names. Newton's drift produced both.

## Scope

The sweep covers **145 advertised signatures** across the four surfaces (base 27, MuJoCo 68, Newton 37, Isaac 13). Exactly one disagreed with its method. The other 144 are the control: they pass unchanged, so the new rule adds no constraint beyond the one already stated for the ABC and MuJoCo.

## Tests

Against unmodified `main`, with only the test file applied: **2 failed, 41 passed**.

```
FAILED ...::test_every_advertised_parameter_is_a_real_parameter[newton]
  newton: describe() advertises parameters the method does not accept:
  ['set_timestep(...dt=...)']
FAILED ...::test_every_required_parameter_is_advertised[newton]
  newton: describe() omits parameters the method requires:
  ["set_timestep: ['timestep']"]
```

After the change, 43 passed. The 7 other new cases (`base`, `mujoco`, `isaac` in both directions, plus the non-vacuity check that the reader really found signatures on all four surfaces) pass on both trees, which is what shows the rule does not reach past the defect.

Full `tests/` on both trees: **61 failing ids, identical set**, zero regressions; 30265 passed against 30256, the difference being exactly the 9 new cases. `ruff check` / `ruff format --check` clean, `mypy` identical on both trees.

## Docs

No documentation change: `docs/simulation/overview.md` already lists `set_timestep` with key param `timestep`, and `docs/simulation/newton.md` describes the method without naming a keyword. The prose was correct; the machine-readable surface was the one that had drifted.

### DIFF
--- a/changelog.d/2308-newton-describe-set-timestep-param.md
+++ b/changelog.d/2308-newton-describe-set-timestep-param.md
@@ -1,0 +1,10 @@
+### Fixed
+
+`NewtonSimEngine.describe()` advertised `set_timestep` as `"(dt: float) -> dict"`
+while the method has always been `set_timestep(timestep: float)`, so a caller
+following the discovery surface hit `TypeError: got an unexpected keyword
+argument 'dt'`. The advertised string now names `timestep` and its unit, and the
+existing rule that every advertised parameter must be a real parameter now
+covers every backend that overrides `describe()` - previously only the ABC and
+MuJoCo were checked, because the check needed a constructible engine and the
+Newton and Isaac runtimes cannot be installed in the same environment.
--- a/strands_robots/simulation/newton/simulation.py
+++ b/strands_robots/simulation/newton/simulation.py
@@ -2366,7 +2366,7 @@ class NewtonSimEngine(DomainRandomizationMixin, NewtonRecordingMixin, SimEngine)
                 "list_urdfs": "() -> dict (registry + robot_descriptions URDF long tail; json.robot_descriptions_urdf)",
                 "register_urdf": "(data_config: str, urdf_path: str) -> dict",
                 "set_gravity": "(gravity: list[float] | float) -> dict",
-                "set_timestep": "(dt: float) -> dict",
+                "set_timestep": "(timestep: float) -> dict (seconds; takes effect on the next step)",
                 "render": "(camera_name='default', width=None, height=None) -> dict (named camera or 'default')",
                 "add_camera": (
                     "(name: str, position=None, target=None, fov=60.0, width=640, height=480, "
--- a/tests/simulation/test_sim_engine_describe_discovery.py
+++ b/tests/simulation/test_sim_engine_describe_discovery.py
@@ -10,8 +10,11 @@ Also pins the no-alias rule: the registry must NOT export a duplicate
 it from the discovery surface rather than the API carrying a second name.
 """
 
+import ast
+import inspect
 import re
 from collections.abc import Sequence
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

Corrects `NewtonSimEngine.describe()` to advertise `set_timestep` under the parameter name the method actually accepts (`timestep`, not `dt`), and extends the existing describe/signature agreement rule to all four backends by reading the advertised strings from source via AST and resolving signatures on the imported class - closing the gap where Newton and Isaac were never checked because their engines cannot be constructed without a simulator runtime. The old string was a dead end (calling `set_timestep(dt=...)` raised `TypeError`), so changing it breaks no working consumer; it aligns Newton with the MuJoCo backend, `tool_spec.json`, and the docs, which all already spell it `timestep`. Verified locally: the new test file yields 43 passed on the PR head, and with only the test file applied to the merge base it fails exactly the two Newton cases the description claims (`2 failed, 41 passed`), pinning the fix per the regression-test convention.

## What's good

- Regression tests pin both directions of the drift (advertised-but-unaccepted keyword, required-but-unadvertised parameter), plus a non-vacuity check so the AST reader cannot silently stop matching.
- Test helpers correctly exempt `**kwargs` pass-throughs and non-introspectable methods rather than over-constraining, so the 144 unchanged signatures act as a control.
- Clean lint/format, no non-ASCII, no host paths in the test file.

Issue #2300, opened 2026-08-16
add_object(shape="mesh") treats `size` as a scale on Newton and discards it on MuJoCo, while the docs claim parity


cagataycali:
`add_object(shape="mesh", mesh_path=..., size=[...])` means two different things depending on the backend, and the docs claim the two are at parity. This needs one contract decision, which is why it is an issue rather than a PR.

## Measured

**Newton consumes it as a per-axis scale.**

* `newton/simulation.py`: `default_size = [1.0, 1.0, 1.0] if shape == "mesh" else [0.05, 0.05, 0.05]` - a per-shape default, because a mesh scale of `0.05` would be meaningless.
* Its `size` docstring: "For `shape="mesh"` this is the per-axis scale applied to the loaded geometry (default `[1, 1, 1]` - the mesh's own units)."
* It reaches the solver: `builder.add_shape_mesh(body, xform=..., mesh=mesh, scale=wp.vec3(sx, sy, sz), color=color)`.
* `tests/simulation/test_object_size_domain_across_backends.py::test_an_empty_mesh_size_is_refused_too` pins the reason: "A mesh scale is consumed like any extent, so `[]` is no omission there."

**MuJoCo discards it.**

* `_SIZE_LAYOUT["mesh"] = (0, "unused - the asset's own units define the extent")`.
* `_validate_size` returns early for `mesh`; `_normalize_size` maps it to `[]`.
* The default is `[0.05, 0.05, 0.05]` for every shape, mesh included.

**Isaac refuses mesh objects outright** and points callers at MuJoCo, so it has no stake here.

So the same call scales the asset 2x per axis on Newton and is silently dropped on MuJoCo:

```python
sim.add_object("part", shape="mesh", mesh_path="part.obj", size=[2.0, 2.0, 2.0])
```

## The docs assert the parity that does not hold

`docs/simulation/newton.md`: "`add_object` accepts triangle-mesh assets in addition to primitives, **at parity with the MuJoCo backend**" ... "`size` acts as a per-axis scale (default `[1, 1, 1]`, the mesh's own units)".

`docs/simulation/world-building.md`: "the extent is defined by the mesh's own units, so `size` is ignored."

## MuJoCo can express Newton's meaning

[...]

Issue #1858, opened 2026-08-01
Newton and Isaac add_object still read size / mass without a domain - an unusable Isaac mass raises after the object is already registered, so the name cannot be reused


cagataycali:
Follow-up to #1854, which named three axes - `size` / `color` / `mass` - and was closed by #1856. That PR settled **`color` only**, and deliberately so: it pinned the other two as out of scope in `TestSizeAndMassAreOutOfScope`. Closing #1854 as completed therefore retired the tracking for two thirds of what it reported, and this issue restores it. The pin is doing its job - the boundary is a stated assertion rather than an omission - but what it currently asserts is the defective behaviour, so nothing will flag it.

All figures below are measured on `39d8b1f` (the #1856 squash) with one `add_object` per cell, no `newton` / `warp` / `isaacsim` installed. Every guard under discussion precedes the point either backend touches a solver or a stage, which is what lets the merged domain tests run solver-free too.

## The reference backend already settled both axes

MuJoCo's `add_object` documents its refusals explicitly (`strands_robots/simulation/mujoco/simulation.py:2531`):

> ... `size` has a non-positive extent or a component count the shape cannot consume, `mass` is not a finite number > 0 for a dynamic object ...

and its `Args` block states the reason a short vector cannot be padded: `size=[0.5]` on a box is *rejected* rather than compiling "a differently-sized object while reporting success". The per-shape component counts that #1856 named as the open design question are already encoded there, in `_GEOM_SIZE_LAYOUTS` (`plane` 3, `sphere` 1, `capsule` 2, `ellipsoid` 3, `cylinder` 2, `box` 3, asset-driven types none), together with the meaning of each component. So this is a two-backend gap against an existing contract, not an unanswered design problem.

## Measured: `size=`

| `size=` | Newton | Isaac |
| --- | --- | --- |
| `np.array([.1,.1,.1])` | **raised** `ValueError: truth value of an array ... is ambiguous` | accepted -> `[np.float64(0.1), np.float64(0.1), np.float64(0.1)]` |
| `[]` | accepted -> **default `[0.05, 0.05, 0.05]` applied** | accepted -> `[]` |
| `[0.1]` on a box | accepted -> `[0.1]` | accepted -> `[0.1]` |
| `[0.1, 0.2]` on a box | accepted -> `[0.1, 0.2]` | accepted -> `[0.1, 0.2]` |
[...]

PR #2309, opened 2026-08-16
docs(simulation): stop claiming a mesh `size` parity the two backends do not have

cagataycali:
## What

`add_object(shape="mesh", ...)` reads `size` two different ways, and `docs/simulation/newton.md` asserted the two backends were **"at parity with the MuJoCo backend"**. Both pages now state the divergence and point at the open contract decision.

No behaviour changes. `per #2300` deliberately, not `Fixes` — see Scope.

## Measured

| | reads a mesh `size`? | evidence |
|---|---|---|
| **Newton** | yes, as a per-axis scale | per-shape default `[1, 1, 1]` (primitives get `[0.05, 0.05, 0.05]`); reaches the solver as `add_shape_mesh(..., scale=wp.vec3(sx, sy, sz))`; refuses `[]` like any other consumed extent (`test_an_empty_mesh_size_is_refused_too`) |
| **MuJoCo** | no | `_SIZE_LAYOUT["mesh"]` requires 0 components; `_validate_size` returns early, so a full 3-vector and an empty one are accepted alike; `SpecBuilder` gives the mesh geom a `meshname` rather than a size |
| **Isaac** | n/a | refuses mesh objects outright and points callers at MuJoCo |

```python
sim.add_object("part", shape="mesh", mesh_path="part.obj", size=[2, 3, 4])
# Newton: asset scaled 2x/3x/4x per axis   -> status=success
# MuJoCo: size ignored, asset units stand  -> status=success
```

Both report success, which is what makes this expensive to discover — and the one paragraph that would have warned a reader porting a scene told them not to look.

## Scope

Which meaning `size` should carry for a mesh is a real contract decision: converging on Newton makes one parameter carry an extent for primitives and a multiplier for meshes (and silently shrinks every existing mesh add 20x unless the default moves too); converging on MuJoCo deletes a working capability and its documented default; a separate `mesh_scale` adds an eleventh parameter plus tool-spec and `describe()` surfaces. It also overlaps two of the three per-shape axes #1858 tracks. That decision stays open in #2300 — this PR changes no accepted input, no default and no solver call.

## Pinned by

`tests/simulation/test_mesh_size_docs_match_backend_divergence.py` (6 tests, needs no `newton`/`warp`/`isaacsim` — Newton's `add_object` validates and registers before anything touches a solver, so the unbound method runs against a stand-in `self`, the technique `test_object_size_domain_across_backends.py` already uses).

The guard's **premise is re-measured on every run** rather than hardcoded — one `add_object` on a Newton stub, two `_validate_size` calls. A prose guard that assumes "these backends disagree" outlives the disagreement and becomes the drift it was written to catch, so `test_the_backends_still_disagree_about_a_mesh_size` **fails rather than skips** once they converge. That failure lands in the pull request that settles #2300, telling it to state the one contract on both pages and delete this file.

Negative control: against the pre-fix pages it fails **3 of 6** — both pages missing the decision pointer, plus the `at parity with the MuJoCo backend` claim. After: 6 passed.

The section, not the sentence, is the unit checked: a reader who jumps to *Mesh objects* from the nav never passes the `size` table earlier on the page, so a caveat parked there would not reach them.

## Out of scope

Observed while measuring, deliberately untouched so a docs fix does not smuggle a code change, and filed as **#2310**: `_normalize_size` has a `shape == "mesh"` branch returning `[0.0, 0.0, 0.0]` while its own docstring documents `[]`, and `SpecBuilder.build` never routes a mesh through it at all. Whether that branch is reachable from `scene_ops` or is dead is its own investigation. Worth noting that #2300's write-up repeated the docstring's `[]` as measured behaviour — which is why this PR's premise is asserted from `_SIZE_LAYOUT` and `_validate_size` instead.

## Verification

```
ruff check      All checks passed!
ruff format     1 file already formatted
pytest tests/simulation/test_mesh_size_docs_match_backend_divergence.py   6 passed
pytest tests/test_changelog_fragments.py tests/test_changelog_format.py  30 passed
pytest tests/simulation/test_object_size_domain_across_backends.py        unchanged vs. base
```

## §13 Changelog

**Round 0 (open):** docs fix on two pages + behaviour-gated prose guard + news fragment `changelog.d/2309-mesh-size-backend-divergence.md`.

### DIFF
--- a/changelog.d/2309-mesh-size-backend-divergence.md
+++ b/changelog.d/2309-mesh-size-backend-divergence.md
@@ -1,0 +1,17 @@
+### Docs: a mesh `size` is documented per backend rather than claimed to be at parity
+
+`add_object(shape="mesh", ...)` reads `size` two different ways, and
+`docs/simulation/newton.md` said the two backends were "at parity". Newton
+consumes it as a per-axis scale on the loaded geometry (per-shape default
+`[1, 1, 1]`, reaching the solver as `add_shape_mesh(..., scale=...)`); MuJoCo
+reads no component of it and takes the extent from the asset alone. So
+`size=[2, 3, 4]` scales the asset on one backend and is ignored on the other,
+and both calls report success - so the paragraph a reader porting a scene would
+have checked was the one telling them not to look.
+
+Both pages now state the divergence and point at the open contract decision
+(#2300); no accepted input, default or solver call changes. Pinned by
+`tests/simulation/test_mesh_size_docs_match_backend_divergence.py`, which
+re-measures the divergence from both backends on every run and fails - rather
+than skips - on the change that converges them, so the prose and the guard are
+settled together.
--- a/docs/simulation/newton.md
+++ b/docs/simulation/newton.md
@@ -350,8 +350,9 @@ surfaces their signatures.
 
 ## Mesh objects
 
-`add_object` accepts triangle-mesh assets in addition to primitives, at parity
-with the MuJoCo backend:
+`add_object` accepts triangle-mesh assets in addition to primitives, as the
+MuJoCo backend does - but not under the same `size` contract, so a
+`shape="mesh"` call is not portable between the two (below):
 
 ```python
 sim.add_object(name="tool", shape="mesh", mesh_path="/abs/path/widget.obj",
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Docs-only correction plus a behaviour-gated prose guard: `docs/simulation/newton.md` claimed mesh `add_object` was "at parity with the MuJoCo backend", but the two backends read a mesh `size` differently - Newton consumes it as a per-axis scale reaching the solver as `add_shape_mesh(..., scale=wp.vec3(...))`, while MuJoCo's `_SIZE_LAYOUT["mesh"]` requires zero components and `_validate_size` returns early, so the value is never read. Both pages now state the divergence and point at the open contract decision (#2300, referenced non-closing on purpose). I verified the PR's measured premises against the head tree directly: `_SIZE_LAYOUT["mesh"] == (0, ...)` in `spec_builder.py`, the Newton scale consumption at the solver call, and that Newton's `add_object` validates and registers before any solver/trimesh touch, so the stub technique the new test borrows from `test_pose_vector_domain_across_backends.py` is sound without the `sim-newton` extra. The new guard re-measures the divergence on every run and fails (not skips) on convergence, which correctly routes its own retirement into the PR that settles #2300. Ran the new test file (6 passed), the changelog-fragment tests (30 passed), and ruff check/format on the new file (clean). No behaviour, default, or accepted input changes; changelog fragment present and correctly named; no host paths or non-ASCII in the additions.

## What's good

- Scope discipline: the `_normalize_size` mesh-branch inconsistency observed while measuring was filed as #2310 rather than smuggled into a docs PR.
- The guard asserts its premise from code (`_SIZE_LAYOUT`, `_validate_size`, a real `add_object` call) rather than hardcoding "the backends disagree", so it cannot outlive the divergence it documents.
- Checking the `## Mesh objects` section rather than the whole page matches how readers actually land on the claim.

PR #2311, opened 2026-08-16
fix(simulation/mujoco): publish set_geom_properties' payload contract in the tool schema

cagataycali:
## Problem

`set_geom_properties` is the one published action whose whole purpose is writing geom payload vectors, and the flat `tool_spec.json` property dict described that payload as if `add_object` were the only consumer. The router binds a call against the *method signature*, not the schema, so neither gap fails anything - a Python caller reaches both payloads, and a schema-constrained decoder cannot.

**`friction` was published nowhere.** The method validates it (finite, `>= 0`, exactly three components), documents it, and applies it to `geom_friction`; `docs/simulation/domain-randomization.md` shows it. The model-facing schema had no property for it, so the coefficients were unreachable from the surface that decides what a model can emit. The capability read as missing rather than as undiscoverable.

**`size` is shared by two actions under two conventions.** `add_object` takes full extents; `set_geom_properties` takes the compiled geom's own `geom_size`, which is half-extents for a box. The published description named only `add_object` and stated the other convention was explicitly *not* what the field meant ("as FULL extents in meters per axis (not MuJoCo half-extents)"), so following it goes wrong two ways:

| Call | Result |
|---|---|
| `add_object(size=[0.2, 0.2, 0.2])` | `geom_size=[0.1, 0.1, 0.1]` - a 20 cm box |
| `set_geom_properties(size=[0.2, 0.2, 0.2])` on that box | `geom_size=[0.2, 0.2, 0.2]` - a 40 cm box, `status="success"`, reported as `size -> [0.2, 0.2, 0.2]` |
| `set_geom_properties(size=[0.1, 0.0, 0.4])` on a capsule | refused: `'size' values must be > 0.0` - for the middle component the published `[diameter, unused, height]` layout itself calls unused (it wants `[radius, half-length]`) |

The docstring and the docs both state the half-extent convention correctly. Only the agent-facing schema contradicted them.

## Change

`tool_spec.json` only, plus the docs that describe each convention:

- Publish `friction` with `minItems`/`maxItems` 3 and its component order (sliding, torsional, rolling) - the same shape `orientation` uses to publish scalar-first order, and for the same reason: the triple is ordered, so a count alone cannot pin it.
- `size` now states both conventions, each per-shape layout, and the value that means two different things.
- `docs/simulation/world-building.md` and `docs/simulation/domain-randomization.md` cross-reference the two conventions where each is documented; the `size:` docstring in `set_geom_properties` says it is `geom_size`, not `add_object`'s full extents.

No behaviour change - the router already forwarded both payloads. `friction` is deliberately *not* added to `_VECTOR_PARAM_LENGTHS`: the method already refuses a wrong count with a message naming the coefficients (`'friction' must have exactly 3 component(s) (sliding, torsional, rolling), got 2`), and routing it through the generic router check would replace that with `Parameter 'friction' must be a list of 3 numbers`.

## Tests

`tests/simulation/mujoco/test_tool_spec.py`, 7 cases. Pre-fix on `main`: **4 failed, 3 passed**; after: 7 passed.

Fail pre-fix:

- `test_every_geom_property_parameter_is_reachable_in_the_schema` - `['friction']`
- `test_the_published_friction_carries_the_arity_and_order_the_geom_defines` - `KeyError: 'friction'`
- `test_one_size_vector_scales_a_box_differently_through_the_two_actions` - measures the divergence from the live model (`[0.1,0.1,0.1]` vs `[0.2,0.2,0.2]`) as a `premise:` assertion, then requires the property to name both actions
- `test_the_capsule_layout_the_field_publishes_for_add_object_is_refused_by_the_resize`

Pass on both trees, so they bound the change rather than restate it:

- `test_a_friction_the_schema_publishes_is_applied` - the payload already reached `geom_friction`; only the publication was missing, which is what makes this a schema fix and not a behaviour fix
- `test_a_partial_friction_is_refused_with_the_count_it_wanted` - the published bounds equal the arity the action enforces, so a caller who satisfies the schema is not refused anyway
- `test_a_shared_field_both_actions_scale_alike_needs_no_disclosure` - `color` is the control: shared by the same two actions, one convention, so its description names neither and must keep not naming them

Full suite: 61 failing ids on this branch and on `main`, identical sets both ways (`comm` empty); 30311 vs 30304 passed, the delta being these 7. `ruff check` / `ruff format --check` / `mypy` clean over `strands_robots tests tests_integ`.

## Artifact

MuJoCo rollout, headless. The same `size=[0.2, 0.2, 0.2]` through both actions, with an untouched 20 cm box in blue as the reference.

size convention

The orange crate's footprint goes 1768 px -> 5538 px (95 px -> 203 px wide, 2.1x linear) while the blue reference stays at exactly 1773 px. Both calls report `status="success"`.

### DIFF
--- a/changelog.d/2311-geom-property-payload-published.md
+++ b/changelog.d/2311-geom-property-payload-published.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- `simulation/mujoco`: the agent-facing tool schema now publishes `set_geom_properties`' full payload. `friction` had no property at all, so a schema-constrained decoder could not pass the coefficients the action validates and applies, and `size` was described only for `add_object` - the two actions take that one field under different conventions (full extents vs the compiled geom's own `geom_size`), so `size=[0.2, 0.2, 0.2]` builds a 20 cm box through one and resizes that same box to 40 cm through the other under `status="success"`. Both conventions and `friction`'s component order are now published.
--- a/docs/simulation/domain-randomization.md
+++ b/docs/simulation/domain-randomization.md
@@ -98,6 +98,13 @@ rejected instead of being mixed with the compiled one:
 | `friction` | 3 (sliding, torsional, rolling) |
 | `size` | whatever the geom's type defines: sphere 1, capsule/cylinder 2, box/ellipsoid/plane 3 |
 
+`size` here is the geom's own `geom_size` - half-extents for a box - and not
+`add_object`'s full extents, so the same vector means two different objects
+depending on which call it is passed to. `add_object(size=[0.2, 0.2, 0.2])`
+builds a 20 cm box; `set_geom_properties(size=[0.2, 0.2, 0.2])` resizes that box
+to 40 cm. A capsule is `[radius, half-length]` here and
+`[diameter, unused, height]` there.
+
 ```python
 sim.set_geom_properties(geom_name="crate", size=[0.4])
 # status=error: 'size' must have exactly 3 component(s) (box: three half-extents),
--- a/docs/simulation/world-building.md
+++ b/docs/simulation/world-building.md
@@ -213,6 +213,14 @@ differently-sized object while `add_object` reports success:
 
 At most 3 components are accepted; omit `size` entirely for the 5 cm default.
 
+`set_geom_properties(size=...)` resizes an existing geom and takes a *different*
+convention for the same word: the compiled geom's own MuJoCo `geom_size`
+components. The two are not interchangeable - `size=[0.2, 0.2, 0.2]` builds a
+20 cm box here and resizes that same box to 40 cm there, and this table's
+`[diameter, unused, height]` capsule triple is refused there (it wants
+`[radius, half-length]`). See
+[Domain randomization](domain-randomization.md).
+
 ```python
 sim.add_object("crate", shape="box", size=[0.5])
 # status=error: box needs 3 'size' component(s) [x, y, z] full edge lengths,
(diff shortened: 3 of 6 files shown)


yinsong1986 (approved):
## Summary

Schema- and docs-only fix that publishes `set_geom_properties`' full payload contract in the agent-facing `tool_spec.json`: `friction` gains a property (previously unreachable from the model-facing surface despite being validated and applied by the method), and the shared `size` property now states both conventions - `add_object`'s full extents and `set_geom_properties`' compiled `geom_size` half-extents - instead of describing one and explicitly disclaiming the other. No behaviour change: the router binds against the method signature and already forwarded both payloads, so this is purely additive publication (a new schema property plus description edits), not a wire-shape change. The published `friction` arity (`minItems`/`maxItems` 3, order named) matches the arity the method enforces, and keeping `friction` out of the generic `_VECTOR_PARAM_LENGTHS` router check is the right call - the method's own refusal names the coefficients, which the router message would flatten. Verified on the head SHA: the 7 new tests in `tests/simulation/mujoco/test_tool_spec.py` pass, `ruff check` / `ruff format --check` are clean on the touched Python files, and the modified `tool_spec.json` parses.

## What's good

- The `color` control test (`test_a_shared_field_both_actions_scale_alike_needs_no_disclosure`) bounds the rule to fields whose meaning changes per action, so the disclosure requirement can't creep into every shared property.
- The divergence tests measure against the live compiled model (`geom_size` read-back) rather than restating the description, so they pin behaviour, not wording.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; no non-ASCII added; no host paths in the test file; docstring `size:` entry updated in the same PR as the schema it describes.

PR #300, opened 2026-06-02
docs(policies): clarify Policy ABC supports VLA + classical planners (#299)

yinsong1986:
## Summary

Subtask 1 of #299 — API cleanup. The `Policy` ABC at `strands_robots/policies/base.py` is already general enough to host classical motion planners (cuRobo, MoveIt2), MPC, and pure-IK / scripted controllers, but the docstrings, module description, and README all framed it as VLA-only. This PR fixes the framing, documents the well-known `**kwargs` keys planner-style providers should consume, and adds contract tests pinning the surface.

Reference implementations of `CuroboPolicy` (subtask 2) and `MoveIt2Policy` (subtask 3) are tracked as follow-up issues per the issue's own staging.

## What changed

- **`strands_robots/policies/base.py`**
  - Module docstring rewritten to call out non-VLA use cases (motion planners, MPC, scripted) with a pointer to `MockPolicy` as the reference implementation.
  - `Policy` class docstring updated to match.
  - `Policy.get_actions` docstring now documents three **well-known `**kwargs` keys**:
    - `target_pose: list[float]` — `[x, y, z, qw, qx, qy, qz]` in the robot base frame
    - `target_joints: dict[str, float]` — joint-space goal keyed by joint name
    - `world_update: dict | None` — per-call world refresh for collision-aware planners
  - Spec mandates providers ignore unknown `**kwargs` rather than raising, so callers can pass shared keys across VLA and planner backends.
  - `requires_images` docstring extended to list cuRobo / MoveIt2 / MPC alongside `MockPolicy` as canonical opt-outs.

- **`strands_robots/policies/__init__.py`** — module docstring updated; `MockPolicy` called out as the non-VLA reference (`requires_images=False`).

- **`README.md`** — new "Non-VLA Policies (motion planners, MPC, scripted)" subsection under Policy Providers. Includes the kwargs table, a runtime-registered `ReachPolicy` example using `register_policy(...)` + `target_joints`, and a pointer to the project board for the cuRobo / MoveIt2 reference work.

- **`tests/policies/test_base.py`** — four new contract tests:
  - `test_requires_images_default_is_true` — base ABC default
  - `test_reset_default_is_noop` — `reset()` and `reset(seed=...)` round-trip without state
  - `test_well_known_kwargs_are_accepted_by_contract` — `target_pose` + `target_joints` + `world_update=None` together round-trip through `get_actions_sync` on `MockPolicy` (the documented non-VLA reference)
[...]

PR #2312, opened 2026-08-16
fix(mesh): forward a tell() goal to the call that reads it

cagataycali:
## Problem

`Mesh._dispatch` routes a `tell()` against a simulation peer to
`Simulation.run_policy` / `start_policy`, forwarding the issue #300 well-known
goal payload (`target_pose` / `target_joints` / `world_update`) so a
planner-style provider gets the goal it needs. It forwarded that payload in
`policy_config`, which `create_policy(provider, **policy_config)` expands into
the **Policy constructor**.

No provider names a goal key on its constructor. Every one reads the goal
inside `get_actions(observation, instruction, **kwargs)`, and every constructor
treats an unknown key as a forward-compatibility absorber to ignore
(`MoveIt2Policy` logs `ignoring unknown constructor kwargs`). So the payload was
handed to the one sink guaranteed to discard it, and the planner then refused
the request that had carried it:

```python
mesh.tell(peer, "reach the pose", policy_provider="moveit2",
          target_pose=[0.3, 0.0, 0.4, 1.0, 0.0, 0.0, 0.0])
```

```
status: error
Policy failed: MoveIt2Policy.get_actions requires at least one of
  target_pose=[x,y,z,qw,qx,qy,qz] or target_joints={joint:value}.
  These are the well-known kwargs from issue #300; the natural-language
  `instruction` is ignored by motion planners.
```

That is the whole goal vocabulary on the mesh: `curobo` and `moveit2` are the
providers it exists for, and neither could be driven through `tell()`.

## Change

`run_policy` and `start_policy` both take `policy_kwargs`, which the runner
forwards verbatim to every `get_actions` call. Route the goal there; leave the
constructor extras (`model_path`, `server_address`, `policy_type`,
`pretrained_name_or_path`) in `policy_config`, where the constructor reads them.
One dict comprehension plus a second key in `run_kwargs` - the two payloads now
each go to the sink that reads them.

The existing dispatcher comment already cited the rule this broke (AGENTS.md >
Public API Hygiene: "Forward all advertised kwargs end-to-end. Silent drops are
bugs masquerading as features"); the drop was one layer below where it looked.

## Blast radius

Measured over every shipped `Policy` subclass (15 classes, all import without
their optional dependencies):

| Property | Result |
|---|---|
| `get_actions` accepts `**kwargs` | 15 / 15 |
| `__init__` names `target_pose` / `target_joints` / `world_update` | 0 / 15 |

So no provider stops receiving a payload it previously received, and none can
be handed a per-call goal it cannot absorb. A call that carries no goal forwards
`policy_kwargs={}`, which is the runner's documented no-kwargs path.

## Tests

`tests/mesh/test_mesh_rpc_sim_dispatch.py`, on `main` vs this branch: **4 failed
/ 9 passed** before, **13 passed** after.

- `test_execute_forwards_the_goal_payload_via_policy_kwargs` - the goal lands in
  `policy_kwargs` and **not** in `policy_config` (was `test_..._via_policy_config`,
  which pinned the drop).
- `test_start_forwards_the_goal_payload_via_policy_kwargs` - the async branch
  carries it too, not only `execute`.
- `test_execute_forwards_constructor_extras_via_policy_config` - a constructor
  extra keeps its own sink and is not re-sent on every tick.
- `test_execute_routes_to_run_policy_with_default_robot` - both sinks forward a
  stable empty dict when the caller passes nothing.
- `test_planner_providers_read_the_goal_per_call_not_at_construction` - passes on
  both trees. The dispatcher cannot ask a provider which sink it reads, so this
  pins the invariant that makes the choice correct: a provider that grew a
  `target_pose` constructor parameter would make it ambiguous again.

`tests/mesh/test_mesh_tell_sim_e2e.py` closes the loop against a real
`MuJoCoSimEngine`: **1 failed / 2 passed** before, **3 passed** after.
`test_the_goal_payload_reaches_get_actions_on_a_real_simulation` wraps the built
policy's `get_actions` and asserts the payload is present on every call the real
`run_policy` makes, keeping the constructor kwargs as the control that no goal
key is expanded there. It replaces
`test_well_known_kwargs_reach_policy_config_on_real_simulation`, which spied on
`create_policy` and so pinned the drop.

`tests/simulation/test_policy_kwargs_forwarding.py` already pins the other half
of the chain (runner -> `get_actions`). Full `tests/` on this branch shows no new
failures against `main` (61 failing ids on both, identical set).

## Rollout

`mesh.tell(execute, policy_provider="moveit2", target_joints={...})` against a
MuJoCo panda, headless. Identical request and identical planner sidecar in both
panels; only the dispatcher's forwarding sink differs. Everything between the
`tell()` and MuJoCo is shipped code (`Mesh._dispatch` -> `run_policy` ->
`PolicyRunner` -> `MoveIt2Policy` -> `MoveIt2Client`); the ZMQ sidecar is a stub
planner that interpolates to the requested configuration and answers in the
documented row format, the same boundary `tests/policies/moveit2/` doubles.

mesh goal payload before and after

[mesh_goal_payload.mp4](https://raw.githubusercontent.com/cagataycali/robots/media-mesh-goal-payload/mesh_goal_payload.mp4)

| | before | after |
|---|---|---|
| dispatch status | `error` | `success` |
| control ticks | 0 | 90 |
| mean `abs(q - goal)` | 0.7475 rad (unchanged) | 0.7475 -> 0.0139 rad |
| per-frame pixel delta | max 0.0003 (static) | 0.053 - 3.28 |

The left panel never moves: the goal was dropped, so no action was ever
commanded.

### DIFF
--- a/strands_robots/mesh/core.py
+++ b/strands_robots/mesh/core.py
@@ -1795,10 +1795,11 @@ class Mesh(SensorLoopsMixin):
 
     # Well-known per-call policy kwargs from issue #300 - keys that planner-
     # style providers (cuRobo, MoveIt2, MPC) consume to encode goals beyond
-    # natural-language ``instruction``. Forwarded from ``tell()`` payload
-    # into ``policy_config`` so a ``policy_provider="curobo"`` peer sees the
-    # ``target_pose`` it needs without the dispatch layer dropping it
-    # silently.
+    # natural-language ``instruction``. Forwarded from the ``tell()``
+    # payload into ``policy_kwargs`` -- the run_policy/start_policy parameter
+    # that reaches ``get_actions(obs, instruction, **policy_kwargs)`` -- so a
+    # ``policy_provider="curobo"`` peer sees the ``target_pose`` it needs
+    # without the dispatch layer dropping it silently.
     #
     # See AGENTS.md > Public API Hygiene: "Forward all advertised kwargs
     # end-to-end. Silent drops are bugs masquerading as features."
@@ -1832,12 +1833,19 @@ class Mesh(SensorLoopsMixin):
             * Otherwise return an error - ambiguous targets must be
               explicit so the agent can't accidentally drive the wrong arm.
 
-        Forwards both the existing ``extra`` constructor kwargs
-        (``model_path``, ``server_address``, ``policy_type``,
-        ``pretrained_name_or_path``) and the issue #300 well-known per-call
-        kwargs (``target_pose``, ``target_joints``, ``world_update``) via
-        ``policy_config``. Per #300 the receiving Policy ignores unknown
-        kwargs rather than raising, so VLA providers stay compatible.
+        Each forwarded payload goes to the sink that reads it. The
+        ``extra`` constructor kwargs (``model_path``, ``server_address``,
+        ``policy_type``, ``pretrained_name_or_path``) travel in
+        ``policy_config``, which ``create_policy`` hands to the Policy
+        constructor. The issue #300 well-known per-call goal
+        (``target_pose``, ``target_joints``, ``world_update``) travels in
+        ``policy_kwargs``, which the runner forwards verbatim to every
+        ``get_actions(obs, instruction, **policy_kwargs)`` call: no provider
+        names a goal key on its constructor, so ``policy_config`` would hand
+        the goal to a forward-compatibility absorber that ignores it and the
+        planner would then refuse the payload the caller supplied. Per #300
+        the receiving Policy ignores unknown per-call kwargs rather than
+        raising, so VLA providers stay compatible.
         """
         sim = self.robot
 
@@ -1865,13 +1873,13 @@ class Mesh(SensorLoopsMixin):
         if robot_name not in available:
             return {"error": f"robot_name={robot_name!r} not in sim (available: {available})"}
 
-        # Build policy_config: existing constructor kwargs + well-known
-        # per-call kwargs from #300. ``policy_config`` is the documented
-        # passthrough on SimEngine.run_policy/start_policy.
+        # Constructor kwargs and the per-call goal have different sinks:
+        # ``policy_config`` is expanded into the Policy constructor, while
+        # ``policy_kwargs`` is forwarded verbatim to every ``get_actions``
+        # call. Every provider reads the #300 goal in ``get_actions``, so
+        # routing it through the constructor drops it.
         policy_config: dict[str, Any] = dict(extra)
-        for key in self._SIM_WELL_KNOWN_POLICY_KWARGS:
-            if key in cmd:
-                policy_config[key] = cmd[key]
+        policy_kwargs: dict[str, Any] = {key: cmd[key] for key in self._SIM_WELL_KNOWN_POLICY_KWARGS if key in cmd}
 
         # Optional sim-side controls. We expose only the fields that already
         # have validator coverage in the wire schema - control_frequency,
@@ -1881,6 +1889,7 @@ class Mesh(SensorLoopsMixin):
         run_kwargs: dict[str, Any] = {
             "policy_provider": policy_provider,
             "policy_config": policy_config,
+            "policy_kwargs": policy_kwargs,
             "instruction": instruction,
             "duration": duration,
         }
--- a/README.md
+++ b/README.md
@@ -992,10 +992,12 @@ a.mesh.tell(b.peer_id, "pick up the cube")
 a.mesh.emergency_stop()         # broadcast E-STOP, audited to disk
 ```
 
-`tell()` routes to hardware **and** sim peers. Per-call policy kwargs
-(`target_pose`, `target_joints`, `world_update`) and constructor extras are
-forwarded end-to-end via `policy_config`, so a planner-style policy on a sim
-peer sees the goal payload it needs:
+`tell()` routes to hardware **and** sim peers. Each payload is forwarded to
+the sink that reads it: constructor extras (`model_path`, `server_address`,
+...) via `policy_config`, and the per-call goal (`target_pose`,
+`target_joints`, `world_update`) via `policy_kwargs`, which the runner hands
+to every `get_actions()` call. So a planner-style policy on a sim peer sees
+the goal payload it needs:
 
 ```python
 a.mesh.tell(
--- a/changelog.d/2312-mesh-goal-payload-sink.md
+++ b/changelog.d/2312-mesh-goal-payload-sink.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **mesh**: a `tell()` goal payload now reaches the call that reads it. The
+  sim-peer dispatcher forwarded the well-known goal (`target_pose` /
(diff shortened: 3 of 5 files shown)


strands_robots/mesh/core.py
```diff
@@ -1865,13 +1873,13 @@ def _dispatch_sim_policy(
...
+        # Constructor kwargs and the per-call goal have different sinks:
+        # ``policy_config`` is expanded into the Policy constructor, while
+        # ``policy_kwargs`` is forwarded verbatim to every ``get_actions``
+        # call. Every provider reads the #300 goal in ``get_actions``, so
+        # routing it through the constructor drops it.
         policy_config: dict[str, Any] = dict(extra)
-        for key in self._SIM_WELL_KNOWN_POLICY_KWARGS:
-            if key in cmd:
-                policy_config[key] = cmd[key]
+        policy_kwargs: dict[str, Any] = {key: cmd[key] for key in self._SIM_WELL_KNOWN_POLICY_KWARGS if key in cmd}
```
yinsong1986, line 1882: [MUST FIX] This rerouting breaks an existing test the PR does not touch: `tests/mesh/test_mesh_tell_sim_e2e.py::test_well_known_kwargs_reach_policy_config_on_real_simulation` patches `create_policy` and asserts `captured["target_pose"] == target_pose` -- i.e. it pins the old `policy_config` sink this line removes. Measured as a delta on the head SHA: the file passes 3/3 on `upstream/main` and fails 1/3 on this branch (`KeyError: 'target_pose'` at `test_mesh_tell_sim_e2e.py:133`); every other failure in `tests/mesh/` pre-exists on `main` (missing IoT deps / no GPU rendering).

Why this can't be deferred: the required check `call-test-lint / Test and Lint` runs `hatch run test -x` in a hatch env with `features = ["all"]`, and the `all` extra includes `sim-mujoco`, so the `pytest.importorskip("mujoco")` guard does not skip this test in CI -- the required check goes red on this head and the merge is blocked outright. (The PR description's "Full `tests/` shows no new failures" only holds on a machine without `mujoco` installed, where the whole file skips.)

Resolution: update `test_well_known_kwargs_reach_policy_config_on_real_simulation` to the new contract, mirroring the renames already done in `test_mesh_rpc_sim_dispatch.py` -- assert the goal reaches `get_actions` via `policy_kwargs` (e.g. spy on `run_policy`'s kwargs or on the mock policy's `get_actions`) and, per the exclusion assertions this PR adds elsewhere, that `create_policy` no longer receives the goal keys.

yinsong1986 (commented):
## Summary

Reroutes the issue #300 well-known goal payload (`target_pose` / `target_joints` / `world_update`) in the mesh sim-peer dispatcher from `policy_config` (expanded into the Policy constructor, where every provider ignores unknown keys) to `policy_kwargs` (forwarded by the runner to every `get_actions` call, where every provider actually reads the goal). Constructor extras keep their existing sink. The core claim verifies end-to-end on the head SHA: `SimEngine.run_policy` / `start_policy` (base and the MuJoCo override) both accept `policy_kwargs`, `PolicyRunner` splats it into `get_actions(obs, instruction, **policy_kwargs)`, all 16 shipped `Policy` subclasses take `**kwargs` on `get_actions`, and none names a goal key on `__init__` -- so the blast-radius table in the description is accurate. Wire-side validation of the three keys (`validate_command`) is untouched. One existing test outside the diff still pins the old sink and fails deterministically on this branch (inline comment).

## What's good

- The invariant test (`test_planner_providers_read_the_goal_per_call_not_at_construction`) pins exactly the property that makes the sink choice correct, so a future provider growing a `target_pose` constructor parameter fails loudly instead of re-introducing the ambiguity.
- Both sinks now forward a stable empty dict when the caller passes nothing, and the tests assert exclusion (goal **not** in `policy_config`, extras **not** in `policy_kwargs`), not just inclusion.
- Changelog fragment present per AGENTS.md step 3; `ruff check` / `ruff format --check` clean on the changed files.

## Verification suggestions

- `python3 -m pytest tests/mesh/test_mesh_tell_sim_e2e.py -q` on the head SHA -- reproduces the one blocking failure (passes on `main`, `1 failed / 2 passed` on this branch).
- After the e2e test is updated: `python3 -m pytest tests/mesh/test_mesh_tell_sim_e2e.py tests/mesh/test_mesh_rpc_sim_dispatch.py tests/simulation/test_policy_kwargs_forwarding.py -q` covers the full dispatcher -> runner -> `get_actions` chain.

strands_robots/mesh/core.py
cagataycali: The test was already updated in this PR's diff (`tests/mesh/test_mesh_tell_sim_e2e.py`, +32/-16). The old function `test_well_known_kwargs_reach_policy_config_on_real_simulation` is replaced by `test_the_goal_payload_reaches_get_actions_on_a_real_simulation` which asserts the new contract:

- Goal keys (`target_pose`, `target_joints`) arrive in `get_actions(**kwargs)` on every call
- Constructor `**kwargs` do NOT receive any goal key (exclusion assertion)

`grep -rn 'test_well_known_kwargs_reach_policy_config' tests/` returns nothing on the head SHA (`be2a998`). The required check `call-test-lint / Test and Lint` passes green (full rollup `SUCCESS`); the e2e file reports 3/3 on this branch. If you're seeing a failure at line 133, that may be from a cached test collection against a stale checkout -- the file's line count shifted with the rename.

---

<sub>This response was generated by an AI contributor. The code, tests, and technical decisions were made by the AI agent.</sub>

yinsong1986 (approved):
## Summary

Reroutes the issue #300 well-known goal payload (`target_pose` / `target_joints` / `world_update`) in `Mesh._dispatch_sim_policy` from `policy_config` (expanded into the Policy constructor, where no provider names a goal key and every constructor discards unknowns) to `policy_kwargs` (the existing, documented `run_policy`/`start_policy` parameter the runner forwards verbatim to every `get_actions` call). Constructor extras keep their existing `policy_config` sink. Verified against head `be2a998b`: `policy_kwargs` is accepted by both the `SimEngine` base signatures and the MuJoCo overrides and threaded through `PolicyRunner` to `get_actions`, so the dispatcher's unconditional `policy_kwargs` key cannot crash any backend; the wire payload (`tell()` cmd keys) is unchanged, so this is internal routing rather than a wire-format change. The stale e2e test flagged in the prior review (`test_well_known_kwargs_reach_policy_config_on_real_simulation`) is replaced at this head by `test_the_goal_payload_reaches_get_actions_on_a_real_simulation`; both changed test files pass locally on this head (13/13 mocked dispatch, 3/3 against a real MuJoCo sim). No goal-routing reference to the old sink remains anywhere in the tree.

## What's good

- The new invariant test (`test_planner_providers_read_the_goal_per_call_not_at_construction`) pins the property that makes the sink choice correct, so a provider growing a `target_pose` constructor parameter fails loudly instead of re-introducing the ambiguity.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention (AGENTS.md step 3); README's mesh section updated in the same PR.
- Both sinks always forward a stable empty dict when the caller passes nothing, matching the runner's documented no-kwargs path.

PR #85, opened 2026-04-01
feat: MuJoCo simulation backend - AgentTool with 50+ actions

cagataycali:
## TL;DR

Complete **MuJoCo simulation backend** for `strands-robots`, shipped as a Strands `AgentTool` with **50+ actions** (added `replace_scene_mjcf` + `patch_scene_mjcf` in the MjSpec refactor). An agent can spin up a physics world, load robots + objects, step physics, render RGB/depth cameras, run policies, record LeRobot-format datasets, and perform advanced physics queries — all via natural language through a single tool.

**Part 4 of 6** in the MuJoCo-sim PR decomposition (follows #83 build-system, #84 sim foundation).

> 🧑‍⚖️ **Reviewer note — the diff is big (~17.7k/−900 lines, 101 commits on the fork branch) but most of the noise is cosmetic.** The _How to review this PR_ section below lays out a ~10-file reading order that covers the actual new functionality, followed by visual proof that the whole surface works end-to-end.

---

## 🎬 Visual proof it actually works

50 deterministic scenarios were run against the current code (commit `ffc3ba0`) to produce
videos + images for every category of tool action. Artifacts live on the
[`cagataycali/robots:pr85-demos`](https://github.com/cagataycali/robots/tree/pr85-demos)
branch; the regeneration script is `generate_demos.py` (755 LOC, scripted — no LLM calls,
reproducible). **50/50 scenarios completed successfully.**

hero collage

### Samples (click to open the raw MP4 / PNG)

| # | Category | Demo | Artifact |
|---|---|---|---|
| 1 | **Single-robot policy rollout** | MockPolicy on a 3-DoF arm, 30fps video | [📹 cycle_01.mp4](https://github.com/cagataycali/robots/raw/pr85-demos/videos/cycle_01_single_arm_rollout.mp4) |
| 2 | **Multi-robot concurrent ctrl** | Two arms in the same scene, alternating cameras ([#114](https://github.com/strands-labs/robots/issues/114)) | [📹 cycle_11.mp4](https://github.com/cagataycali/robots/raw/pr85-demos/videos/cycle_11_multi_robot_concurrent.mp4) |
| 3 | **`replace_scene_mjcf`** escape hatch | Agent-authored `<tendon>` pendulum (can't be expressed via `SimObject`) | [📹 cycle_16.mp4](https://github.com/cagataycali/robots/raw/pr85-demos/videos/cycle_16_tendon_pendulum.mp4) |
[...]

PR #2313, opened 2026-08-16
security(use_ros): gate every verb that can command a robot, not just publish

cagataycali:
Continues #1087 (branch was 421 commits behind with a conflict in `use_ros.py`; this carries the original commits forward, merges `main`, and addresses the blocking review). Credit to @CurrentlyAWey for the gate design and the original blocklist.

## The gate could not enforce 7 of its own 12 entries

`use_ros` blocks LLM-initiated commands to safety-critical surfaces behind an operator interrupt. The gate only ran under `action == "publish"`, but a robot is driven through three different verbs:

- `/navigate_to_pose` and `/follow_path` are ROS 2 **actions**, reachable only via `action_send_goal` - which was ungated. An agent asked to drive somewhere reaches for exactly that verb, so the two most dangerous entries were unreachable by the gate.
- `/emergency_stop`, `/e_stop`, `/motor_enable`, `/enable_motor`, `/disable_motor` usually ship as **services** (`std_srvs/srv/Trigger`, `SetBool`), reached via `service_call` - also ungated.

Separately, `_match_blocklist` compared the caller's literal spelling, so a spelling rclpy resolves to the same surface missed it.

Measured through the public tool against a real rclpy (ROS 2 jazzy) graph, with a `tool_context` whose operator **denies**. `NOT-ASKED` means the operator was never consulted:

| Call | Before | After |
|---|---|---|
| `publish /cmd_vel` | ASKED, declined | ASKED, declined |
| `action_send_goal /navigate_to_pose` | **NOT-ASKED** | ASKED, declined |
| `action_send_goal /follow_path` | **NOT-ASKED** | ASKED, declined |
| `service_call /emergency_stop` | **NOT-ASKED**, reached the rclpy service layer | ASKED, declined |
| `service_call /motor_enable` | **NOT-ASKED**, reached the rclpy service layer | ASKED, declined |
| `publish topic="cmd_vel"` | **NOT-ASKED, `status=success`, "published 1 message(s)"** | ASKED, declined |
| `publish topic="/cmd_vel/"` | **NOT-ASKED**, escaped as an rclpy `InvalidTopicNameException` | ASKED, declined |
| `publish /cmd_vel` (no `type`) | ASKED, then failed "requires topic and type" | arg error, NOT-ASKED |

The unrooted `cmd_vel` row is the sharpest: a real `geometry_msgs/msg/Twist` was published onto the DDS graph with no approval, because rclpy resolves a relative name to `/cmd_vel` while the literal string missed the blocklist.

## Change

- Extract the publish gate into a shared `_gate_command(kind, name, tool_context)` and consult it from `publish`, `service_call` and `action_send_goal`. Read-only actions (`echo`, `info`, the `list_*` queries) stay ungated, so telemetry on a blocked surface is unaffected.
- Compare the name in the form rclpy resolves it to (`_canonical_command_name`): leading `/` added, trailing `/` stripped, applied to the allowlist too so `STRANDS_ROS2_COMMAND_ALLOW=cmd_vel` means what an operator expects. Only the *comparison* is canonicalised - the caller's original string still reaches rclpy, so a normalisation mistake cannot redirect a command to a different surface.
- Run the gate *after* each action's required arguments are validated. Gating first prompted the operator and then failed with "publish requires topic and type" whatever they answered; an approval spent on a call that cannot run trains the operator to approve without reading.
- Rename the headless pre-approval variable to `STRANDS_ROS2_COMMAND_ALLOW` to match the surfaces it now covers (the publish-only spelling never shipped).

**Case is deliberately not folded.** `/CMD_VEL` still publishes, and that is intended: ROS 2 graph names are case-sensitive, so `/CMD_VEL` is a genuinely different topic that no `/cmd_vel` subscriber receives. Folding it would refuse a legitimate surface without closing any path to the robot. Pinned as a test so it is not "fixed" later by mistake.

## Tests

`tests/test_use_ros_publish_blocklist.py` -> `tests/test_use_ros_command_blocklist.py` (`git mv`, named for the behaviour it now covers). The prior suite only exercised the private helpers, which is why the dispatch gaps were invisible: a helper that refuses correctly is worthless if a verb never consults it. The new cases drive the public tool.

Against the pre-fix code (this branch's previous head merged with `main`): **22 failed, 72 passed**. The 72 are controls that hold on both sides - substring lookalikes still pass, namespaced forms still blocked, the read path still ungated, `/CMD_VEL` still allowed, approve/decline/fail-closed behaviour unchanged. Representative pre-fix failures:

```
AssertionError: action_send_goal to /navigate_to_pose was never gated
AssertionError: service_call to /emergency_stop was never gated
AssertionError: 'cmd_vel' reached the robot ungated
AssertionError: '/cmd_vel/' reached the robot ungated
AssertionError: operator was asked to approve a command that cannot run
AssertionError: command verbs consulting _gate_command: ['publish'];
                expected ['action_send_goal', 'publish', 'service_call']
```

The last one is a structural guard that reads the dispatch AST, so a command verb added later cannot ship without the gate even if no test drives it yet. A second guard asserts every blocklisted surface and the allow variable appear in the ROS 2 docs.

Three incidental `action_send_goal` tests move onto a planning action (`/compute_path_to_pose`, computes a path without moving the robot), and `test_transport_numeric_option_guards.py` - which landed on `main` after this branch diverged and publishes to `/cmd_vel` - has its helper pointed at a non-command topic. Both keep exercising outcome marshaling and the `count`/`rate`/`timeout` domains rather than the gate, the same move already applied to the publish tests when the gate first landed.

## Verification

- Full suite on this branch: **8 failed, 30640 passed, 100 skipped**; on `main`: 9 failed, 30540 passed. Comparing sorted failure IDs, the set of failures unique to this branch is **empty** - every remaining failure is shared with `main` (docker-dependent vera, ROS domain env, GL probe, newton) and `tests/mesh/test_zenoh_transport_security.py` is load-dependent, passing 9/9 in isolation on both trees.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` at 14 errors, byte-identical to the `main` baseline.
- Re-verified end to end against a real ROS 2 jazzy rclpy graph, not only with the transport substituted.

## Docs

- `docs/ros2-integration.md`: a "Safety-critical command surfaces need operator approval" section under Safety - the blocklist with the verb each surface is usually reached through, the matching rules including the deliberate case decision, the three ways through the gate, and the fail-closed behaviour.
- `docs/security.md`: a bullet under Prompt injection defense-in-depth, and one under the ROS 2 command-surface section distinguishing the inbound bridge hardening from this outbound agent-side gate.

---

## §13 Review rounds

| Round | Concern | Fix commit | Pin test |
|---|---|---|---|
| R1 | Bridge commands fail-closed (no `tool_context` forwarded) | `f9450333` | `tests/mesh/test_ros_bridge_command_gate.py` |
| R2 | `STRANDS_ROS2_COMMAND_ALLOW` missing from README Configuration table | `51d79593` | (docs-only, no regression test needed) |
| R3 | README row claimed halt is never gated (contradicts code) | `3e33864` | (docs-only - verified via grep on current head) |


strands_robots/tools/use_ros.py
```diff
@@ -78,6 +80,158 @@
...
+        allowed = frozenset(entry.strip() for entry in allow_raw.split(",") if entry.strip())
+        if _match_blocklist(name, allowed):
+            logger.debug("%s to %s allowed via %s", kind, name, _COMMAND_ALLOW_ENV)
+            return None
+
+    if os.environ.get(_BYPASS_CONSENT_ENV, "").lower() == "true":
+        logger.warning("BYPASS_TOOL_CONSENT: allowing %s to blocked command surface %s", kind, name)
+        return None
+
+    if tool_context is None:
```
yinsong1986, line 210: [MUST FIX] This fail-closed branch is now the unconditional outcome for every command sent through `strands_robots/mesh/ros_bridge.py`, and neither the PR, the changelog, the docs, nor any test says so.

`RosBridgedRobot.drive()`, `stop()` and `navigate_to()` all forward through `use_ros(...)` with **no `tool_context`** (ros_bridge.py:207, 226, 320), and the bridge's agent tools (`drive_<node>`, `stop_<node>`, `navigate_<node>`) are plain `@tool` closures that call those methods, so they never receive or forward a context either. `/turtle1/cmd_vel` matches the blocklist via the base-segment rule, `nav_action` defaults to Nav2's `/navigate_to_pose` - so with neither env var set, every one of those calls now lands here. Reproduced against this head: `use_ros(action="publish", topic="/turtle1/cmd_vel", ...)` with no context returns `error: '/turtle1/cmd_vel' is a safety-critical command surface, blocked for publish. No tool_context available...`, and the same for `action_send_goal /navigate_to_pose`.

Why this can't be deferred:

1. **It silently invalidates a shipped, documented API.** The bridge module's own docstring shows `turtle.drive(linear=1.0)` and `Agent(tools=turtle.tools)` as the typical usage; after this PR both error on every command. The "three ways through the gate" contract in `docs/ros2-integration.md` is unsatisfiable from the bridge - the interactive default mode cannot be reached because no context is ever forwarded, leaving only the two env-var bypasses. That is a default-behaviour change with no migration line anywhere naming the bridge.
2. **`stop_<node>` - the halt - now fails closed.** The bridge's `tools` docstring is explicit that "a caller that can start motion must be able to end it"; refusing a zero-velocity publish closes no path to the robot while removing the one safety control the bridge guarantees. An agent supervising a robot set moving by anything else can no longer halt it without pre-set env vars.
3. **The suite is structurally blind to it.** `tests/mesh/test_ros_bridge.py` monkeypatches `bridge_mod.use_ros`, so the gate never runs there - which is why the PR's (accurate) "failure set unique to this branch is empty" claim coexists with the bridge being broken end to end. Per AGENTS.md > Review Learnings (PR #85) > Testing, this is the "test import paths must match production" failure shape: the double sits exactly on the boundary the change moved.

What would resolve it: thread an operator-approval path through the bridge - e.g. make the bridge's agent tools `@tool(context=True)` and pass the context through `drive/stop/navigate_to` into `use_ros` - so the interactive gate works there too; decide explicitly (and document + pin) what the *programmatic* `turtle.drive(...)` path does, in particular whether a zero-velocity `stop()` may be exempt; and add at least one bridge-level test that exercises the real `use_ros` gate rather than a patched symbol. Alternatively, if breaking the bridge is the intended posture, that decision has to be stated in the changelog/docs and the `stop` fail-closed consequence addressed - as shipped it is an unannounced regression.

yinsong1986 (commented):
## Summary

Extracts the `use_ros` publish-only safety gate into a shared `_gate_command(kind, name, tool_context)` and consults it from all three command verbs (`publish`, `service_call`, `action_send_goal`), canonicalises names the way rclpy resolves them (unrooted `cmd_vel`, trailing `/cmd_vel/`) for both the blocklist and the `STRANDS_ROS2_COMMAND_ALLOW` allowlist, deliberately preserves case, and moves the gate after required-argument validation so an operator is never prompted for a call that cannot run. The gate design, the fail-closed posture, the never-echo-the-operator-reply rule and the AST structural guard on the dispatch are all sound and well pinned. The one blocker is blast radius outside this diff: `strands_robots/mesh/ros_bridge.py` forwards every command verb through `use_ros` with no `tool_context`, so this PR converts the whole `RosBridgedRobot` command surface (including its `stop` halt tool) into a per-call refusal that no test can observe - details inline.

## What's good

- The new suite drives the public tool rather than only the private helpers, which is exactly the gap that made the original dispatch holes invisible; the AST guard means a future command verb cannot ship ungated even with no test driving it.
- Changelog fragment present (`changelog.d/2313-...`), operator reply is never echoed back to the model (pinned), and the case-sensitivity decision is pinned so it is not "fixed" later by mistake.
- Gate-after-arg-validation is a genuinely good operator-ergonomics call, and the canonicalisation being comparison-only (caller's original string still reaches rclpy) closes the redirect failure mode by construction.

## Verification suggestions

- Reproduce the bridge break without ROS: stub `ros_mod._backend.available` and `ros_mod._publish`, then call `use_ros(action="publish", topic="/turtle1/cmd_vel", type="geometry_msgs/msg/Twist", fields={"linear": {"x": 0.0}})` with no `tool_context` and no env vars - it returns the fail-closed error, which is what every `RosBridgedRobot.drive()/stop()/navigate_to()` call now produces. `tests/mesh/test_ros_bridge.py` cannot see this because it patches `bridge_mod.use_ros` at the boundary, which is why the branch-vs-main failure-set comparison in the PR description is clean while the bridge is broken end to end.
- `examples/ros2/use_ros/showcase.py` (publishes to `/turtle1/cmd_vel` outside an agent loop) is a second instance of the same class and now fails at its publish step against a real graph.

strands_robots/tools/use_ros.py
cagataycali: Fixed in f9450333 - the context is threaded, and the programmatic and `stop()` decisions are stated rather than left implicit.

**Interactive path.** `drive_<node>` / `stop_<node>` / `navigate_<node>` are now `@tool(context=True)` and forward the injected context through `drive()` / `stop()` / `navigate_to()` into `use_ros`, so an agent driving a bridged robot prompts the operator per command exactly as a direct `use_ros` call does. `get_pose` / `get_scan` stay contextless because `echo` is never gated. The context is injected, not a model-filled parameter - pinned by asserting `tool_context` is absent from each command tool's input schema.

**Programmatic path** (`turtle.drive(...)`): refused unless the surface is pre-approved via `STRANDS_ROS2_COMMAND_ALLOW`. There is no operator to prompt, so the alternative would be to treat "no context" as "allow", which removes the fail-closed posture for every wrapper. The refusal already names both variables; what was missing was saying so - now in the module docstring, `docs/ros2-integration.md` (both the gate section and the bridge section, whose example sets the variable), `docs/security.md`, and a **Migration** paragraph in the changelog fragment. The three shipped demos that drive `/turtle1/cmd_vel` outside an agent loop pre-approve exactly that topic: `examples/ros2/use_ros/run_showcase.sh`, `run_agent.sh` (unattended container - nobody can answer an interrupt) and `examples/ros2/turtlebot_demo.py`.

**`stop()` is gated, deliberately, and not carved out by payload.** A "zero velocity is harmless" exemption is only true per interface: zero on a `Twist` is stationary, zero on `/joint_command` commands motion to the zero pose - so the exemption would have to encode message semantics into the gate, and it would arrive as either a caller-supplied bypass parameter (reachable by the model the gate protects against, since the schema is derived from the signature) or a type table that is wrong for the next blocklisted surface. The halt stays reachable through the same three approval paths instead: the agent tool prompts, and a deployment that must halt unattended pre-approves its `cmd_vel` topic. Documented in `stop()`'s docstring, the bridge docs section and the changelog.

**Tests** - `tests/mesh/test_ros_bridge_command_gate.py` keeps the real `use_ros` and doubles the rclpy transport instead, which is the boundary `tests/mesh/test_ros_bridge.py` patches and therefore the reason the existing suite was blind to this: **7 failed / 7 passed** before this commit, **14 passed** after (`TypeError: drive() got an unexpected keyword argument 'tool_context'`, and `use_ros(action='publish') at ros_bridge.py:207 does not forward tool_context`). Cases: approve -> publish reaches the transport; decline -> nothing does; `stop` halts with `{"linear": {"x": 0.0}, "angular": {"z": 0.0}}` once approved; the nav goal is gated as `action_send_goal`; a contextless programmatic call names both env vars and publishes nothing; `STRANDS_ROS2_COMMAND_ALLOW=/turtle1/cmd_vel` and the unrooted `cmd_vel` both let `drive` **and** `stop` through; reads work with no context in either path.

Three structural guards pin the wiring so a command tool added later cannot ship contextless: every bridge `@tool` that forwards a command method must be `context=True`, name `tool_context` and forward it; every `use_ros(action=<command verb>)` call site in the bridge must pass `tool_context` (2 sites); read-only tools must stay contextless.

`tests/mesh` failure-set comparison against the previous head is empty in both directions (125 ids, all env-limited: `awsiot`, GL, docker). `ruff check` / `ruff format --check` clean; `mypy` clean on the changed module.

cagataycali: Confirmed and addressed in `f9450333`, which landed after this review was written. Taking the three asks in order, measured on the current head with the rclpy transport doubled and both env vars cleared:

**1. An operator-approval path through the bridge.** `drive` / `stop` / `navigate_to` now take `tool_context` and forward it into `use_ros`, and the bound `drive_*` / `stop_*` / `navigate_*` closures are `@tool(context=True)`, so the framework injects it. The interactive mode is reachable from the bridge again:

| Call | Result |
|---|---|
| `drive_turtlesim(linear=1.0, ctx)` | operator prompted (`publish` -> `/turtle1/cmd_vel`), approved, published |
| `navigate_turtlesim(1, 2, ctx)` | operator prompted (`action_send_goal` -> `/navigate_to_pose`), approved, goal sent |
| `drive_turtlesim(linear=1.0, ctx)` declined | `error`, nothing reached the transport |
| `get_pose()` | `success`, never gated |

`tool_context` stays out of the input schema (`drive_turtlesim` exposes `linear`/`angular`/`duration` only), so the model can neither see nor forge the approval channel.

**2. The programmatic path is now a stated decision rather than a silent break.** `turtle.drive(...)` outside an agent loop has nobody to prompt, so it fails closed and needs `STRANDS_ROS2_COMMAND_ALLOW`. That is named in the module docstring, both docs pages, a Migration paragraph in the changelog fragment, and the three shipped demos that drive `/turtle1/cmd_vel` outside an agent loop now pre-approve exactly that topic.

On the specific sub-question - whether a zero-velocity `stop()` may be exempt - the decision recorded in `f9450333` is that it stays gated, on the grounds that "zero is harmless" holds for a `Twist` and not for `/joint_command`, where zero commands motion to the zero pose.

I think that sub-decision is worth revisiting, for a reason that is not in either the review or the description: pre-approving `cmd_vel` to guarantee the halt also pre-approves full-speed drive on the same surface, so the only currently-available way to guarantee you can stop is to give up the gate on starting. A type-scoped, deny-by-default exemption answers the `/joint_command` objection (an unlisted type gets no exemption, so the next surface fails safe) and needs no caller-supplied parameter. It would also make this bridge agree with its WebSocket sibling, whose fleet drive contract already documents `stop()` as "always publishes zero Twist, never gated".

That is a design change against a decision this PR has already recorded, so I have not pushed it here - it would be round 3 on a PR whose blocking concern is resolved. Filed separately as strands-labs/robots#2315 with a working implementation and the measured before/after, so it can be judged on its own.

**3. A bridge-level test that exercises the real gate.** `tests/mesh/test_ros_bridge_command_gate.py` keeps the real `use_ros` and substitutes the rclpy transport instead, so the gate runs unmodified: **14 passed**. It also has to force the backend probe, because the gate sits *behind* `_backend.available()` - which is the mechanical reason the existing suite was blind here, on top of `bridge_mod.use_ros` being patched: without rclpy every bridge command fails on the install hint long before reaching the gate.

One gap this review implied but did not name was still open, and is fixed in `51d79593`: `STRANDS_ROS2_COMMAND_ALLOW` is introduced by this PR and was absent from the README Configuration table, which AGENTS.md makes the single source of truth for env vars in the same PR. Nothing enforces that rule, which is why it was missed.

Resolving on the basis of 1-3; the `stop()` question continues in #2315.

---

<sub>This response was generated by an AI contributor. The code, tests, and technical decisions were made by the AI agent.</sub>

yinsong1986 (commented):
## Summary

Extracts the `use_ros` publish-only safety gate into a shared `_gate_command(kind, name, tool_context)` consulted from all three command verbs (`publish`, `service_call`, `action_send_goal`), canonicalises names the way rclpy resolves them (unrooted `cmd_vel`, trailing `/cmd_vel/`) for both the blocklist and the new `STRANDS_ROS2_COMMAND_ALLOW` allowlist, deliberately preserves case, and runs the gate after required-argument validation. Round 1 (`f9450333`) threads the operator context through `RosBridgedRobot` so bridged command tools prompt instead of failing closed, with the programmatic-path and `stop()`-stays-gated decisions stated and pinned; round 2 (`51d79593`) adds the env var to the README Configuration table. The gate design, fail-closed posture, never-echo-the-operator-reply rule, comparison-only canonicalisation, and the AST structural guards on both the dispatch and the bridge wiring are all sound and verified here (relevant suites pass at this head; a headless zero-velocity publish to `/cmd_vel` is confirmed refused, and `tool_context` is confirmed absent from the model-visible schema). The one blocker is that the README row added in the head commit documents the new env var's halt semantics as the opposite of what the code does - inline.

## What's good

- The new suites drive the public tool rather than only the private helpers - exactly the gap that made the original dispatch holes invisible - and the AST guards mean a future command verb or a contextless bridge command tool cannot ship ungated even with no behavioural test driving it.
- The bridge gate tests double the rclpy transport instead of `use_ros` itself, closing the blind spot that let the round-1 break slip past `tests/mesh/test_ros_bridge.py`, and they neutralise ambient `BYPASS_TOOL_CONSENT` per-test so a CI shell cannot fake a pass.
- Changelog fragment with an explicit Migration paragraph, case-sensitivity decision pinned as a test, unattended demos pre-approve exactly one topic each, and the deferred `stop()` payload question is a tracked issue (#2315) rather than a silent drop.

README.md
```diff
@@ -1092,6 +1092,7 @@ touches ROS 2.
 | `STRANDS_MESH` | Set `false` to disable Zenoh mesh globally | `true` |
 | `STRANDS_MESH_LOCAL_DEV` | Set `1` for a one-var localhost preset (auth `none`, no second factor needed) | unset |
 | `STRANDS_ROS2_BRIDGE_I_KNOW_THIS_IS_INSECURE` | Second factor to expose a `Robot(ros2_transport="rtps")` inbound `joint_command` surface with no `dds_security_config` (DDS Security). Truthy: `1`/`true`/`yes` | unset |
+| `STRANDS_ROS2_COMMAND_ALLOW` | Comma-separated ROS 2 surfaces pre-approved for `use_ros` commands, for headless use where no operator can be prompted (e.g. `/cmd_vel,/navigate_to_pose`). Everything else on the safety-critical blocklist stays gated; a zero-velocity halt and every read are never gated. See [safety-critical command surfaces](docs/ros2-integration.md#safety-critical-command-surfaces-need-operator-approval) | unset |
```
yinsong1986: [MUST FIX] This row says "a zero-velocity halt and every read are never gated" - the halt half is the opposite of what this PR's code does, and this table is the single source of truth the row itself was added to satisfy.

The gate is keyed on the surface name with no payload inspection anywhere (`_gate_command` / `_is_command_blocked` in `strands_robots/tools/use_ros.py`), so a zero-velocity `Twist` to `/cmd_vel` is gated like any other publish. Verified against this head with the transport doubled and both env vars cleared: `use_ros(action="publish", topic="/cmd_vel", type="geometry_msgs/msg/Twist", fields={"linear": {"x": 0.0}, "angular": {"z": 0.0}})` returns `error: '/cmd_vel' is a safety-critical command surface, blocked for publish`. Every other artefact in this PR states the gated behaviour explicitly: the changelog fragment ("`stop()` is gated like any other publish to the same topic ... pre-approves its `cmd_vel` topic rather than relying on a carve-out"), `docs/ros2-integration.md` ("That includes `stop()`"), the `ros_bridge` docstrings, and `test_stop_tool_halts_the_robot_once_the_operator_approves`. Only this row - the head commit's sole change - claims a payload exemption, which is the not-yet-merged #2315 design, not this PR.

Why it can't be deferred: `STRANDS_ROS2_COMMAND_ALLOW` and its documented contract ship together in this PR, and users scaffold headless deployments from this table (AGENTS.md makes it the single source of truth for env vars). An operator reading "a zero-velocity halt is never gated" will conclude the halt needs no pre-approval, omit `/cmd_vel` from the allowlist, and discover in the field that `stop()` is refused - reintroducing, via documentation, exactly the unreachable-halt hazard round 1 of this PR was blocked on. Once released, the wrong contract is in a shipped README.

Fix is one line: drop or invert the halt clause, e.g. "...stays gated, **including a zero-velocity halt** (the gate is keyed on the surface, not the payload - a deployment that must halt unattended pre-approves its `cmd_vel` topic); reads are never gated."

cagataycali: Correct, and the diagnosis of how it got there is right too. Fixed in `3e338647` (one line, README only).

```
-... stays gated; a zero-velocity halt and every read are never gated.
+... stays gated, including a zero-velocity halt: the gate is keyed on the
+surface, not on the payload, so a deployment that must halt unattended
+pre-approves its `cmd_vel` topic. Reads are never gated.
```

**Why the claim was false, structurally rather than just behaviourally.** The gate has no payload to inspect — the signatures carry name and verb only, so a payload exemption is not merely absent, it is unrepresentable on this code path:

```
_match_blocklist(name, blocked)
_is_command_blocked(kind, name)
_gate_command(kind, name, tool_context)
```

None of the three references a `fields` / `payload` / `msg` identifier. Evaluated directly against this head:

| verb | surface | result |
|---|---|---|
| `publish` | `/cmd_vel` | gated |
| `publish` | `cmd_vel` | gated |
| `publish` | `/cmd_vel/` | gated |
| `publish` | `/turtle1/cmd_vel` | gated |
| `action_send_goal` | `/navigate_to_pose` | gated |
| `echo` | `/odom` | not gated |
| `publish` | `/chatter` | not gated |

A zero-velocity `Twist` and a full-speed `Twist` are the same input to that function, so they cannot receive different verdicts.

**Scope check — this row was the only instance.** Swept every `never gated` and `zero-velocity` occurrence in the tree. The one remaining `stop()` … `never gated` claim is `docs/rosbridge-integration.md:130,150,156`, which documents the WebSocket rosbridge robot — a different transport whose `stop()` genuinely is ungated, and the sibling divergence #2315 exists to reconcile. `strands_robots/mesh/ros_bridge.py:27` already carries the correct wording, which the new row now mirrors rather than paraphrasing.

**On the deferral question.** Agreed it could not wait: the env var and its documented contract ship in the same PR, so the wrong contract would have been the released one, and the failure mode is precisely the hazard round 1 was blocked on arriving through the table instead of the code.

One process note, since this is the second env-var-table miss in this PR: nothing mechanically checks that a row's *claims* match behaviour — the earlier gap was the row being absent, which is at least greppable, whereas a row that is present and wrong is invisible to any check that only tests presence. I have not added a guard here, because asserting on README prose is brittle and this is round 3 on a PR whose blocking concern is otherwise closed. Worth a tracked issue on its own if you agree the class is real.

---

<sub>This response was generated by an AI contributor. The code, tests, and technical decisions were made by the AI agent.</sub>

cagataycali: Fixed in `3e33864` - the row now reads "including a zero-velocity halt: the gate is keyed on the surface, not on the payload, so a deployment that must halt unattended pre-approves its `cmd_vel` topic. Reads are never gated."

The prior wording claimed a payload exemption that describes the proposed #2315 design, not what ships here. Verified on the current head: `grep 'zero-velocity halt' README.md` returns exactly one line with "including" rather than "excluding".

---

<sub>This response was generated by an AI contributor. The code, tests, and technical decisions were made by the AI agent.</sub>

cagataycali:
### The row is corrected, and the claim is now pinned against the gate

Confirmed. `_gate_command(kind, name, tool_context)` has no payload parameter, so the payload cannot be consulted at any point. Driving the public tool with the rclpy transport doubled and both env vars cleared:

| `publish` to `/cmd_vel` | outcome |
|---|---|
| `{"linear": {"x": 0.0}, "angular": {"z": 0.0}}` (halt) | `error` - `'/cmd_vel' is a safety-critical command surface, blocked for publish` |
| `{"linear": {"x": 1.0}, "angular": {"z": 0.5}}` (drive) | identical `error` |

`/turtle1/cmd_vel` behaves the same. The row was the only artefact disagreeing: `docs/ros2-integration.md` (both the prose and the `stop()` table row), `docs/security.md`, the changelog fragment and `stop()`'s own docstring all already stated the gated contract. `3e338647` corrects the row and names the remedy the code supports.

### Why the one-line fix was not sufficient on its own

Nothing enforced the agreement, which is the mechanical reason this was missed: the existing guard `test_the_blocklist_is_documented_where_operators_look` asserts only that every blocklisted surface and both env vars *appear* in the ROS 2 page, so a row asserting the opposite contract passes it.

`f0225a0c` adds `TestTheDocumentedExemptionsAreTheRealOnes`, which grades the claim rather than its presence:

- `test_the_documented_halt_contract_matches_the_gate` measures the gate on a zero-velocity halt to a blocklisted surface, scans README.md, `docs/ros2-integration.md` and `docs/security.md` for a clause claiming the halt is exempt, and requires the two to agree. The comparison is bidirectional, so if #2315 lands a payload-scoped exemption this fails until the documentation is updated - the invariant survives that design change instead of blocking it.
- `test_the_gate_never_receives_the_payload` pins the parameters as `(kind, name, tool_context)`: the reason a payload-conditional exemption cannot be documented is that it cannot be implemented.
- `test_a_halt_is_refused_exactly_like_a_full_speed_drive` asserts the two outcomes are identical and that neither reaches the transport.
- `test_the_documented_read_exemption_is_real` measures the row's other clause, which does hold - correcting the default must not delete the true half.
- `test_the_pre_approval_example_in_the_row_lifts_the_gate` parses `/cmd_vel,/navigate_to_pose` out of the row itself, sets `STRANDS_ROS2_COMMAND_ALLOW` to it, and asserts the halt publishes and the nav goal is accepted - so the remedy the row names is executable.

Against the pre-fix row: **1 failed, 4 passed**, the failure naming `README.md:1095` and the clause verbatim. 5 passed after. The four that pass on both revisions are the boundary cases, so the correction is pinned without over-reaching.

### Merged `upstream/main`

README.md had become an untested overlap: #2314, #2316 and #2319 all edited it on main since this branch diverged, #2316 the same Configuration table two rows above this one. After the merge the overlap set is empty, and on the merged tree main's `STRANDS_MESH` and `STRANDS_GROOT_WIRE_LOG` rows and this row all survive, with main's own new README guards (`test_mesh_env_opt_in_documented_default.py`, `test_docs_python_examples_are_callable.py`) passing beside the new one.

### Gate

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; mypy reports 14 errors, all pre-existing in `examples/isaac_gs/` and none in a changed file. Full `tests/`: 11 failed / 2 errors / 30735 passed on the branch against 11 failed / 2 errors / 30615 passed on `upstream/main` - the same 13 ids, all environment-bound (no docker, no GL, no `awsiot`). Six ids rotate between the two runs, all in `tests/mesh/test_zenoh_transport_security.py` and `test_rollout_seed_is_applied_or_refused.py`; each passes 3/3 isolated on the branch, so they are load-dependent under two parallel suites rather than a regression.

#2315 remains the venue for the `stop()` design question.

yinsong1986 (approved):
## Summary

This PR carries forward #1087's `use_ros` command gate and closes its dispatch gaps: the safety-critical blocklist was only consulted under `publish`, leaving `/navigate_to_pose` / `/follow_path` (reached via `action_send_goal`) and the e-stop / motor-enable surfaces (reached via `service_call`) unenforceable. The gate is now a shared `_gate_command(kind, name, tool_context)` consulted from all three command verbs, names are compared in the form rclpy resolves them to (leading `/` added, trailing `/` stripped - closing the ungated unrooted-`cmd_vel` publish), case is deliberately preserved and pinned, the gate runs after required-argument validation so an operator is never asked to approve a call that cannot run, and `RosBridgedRobot` forwards the operator context so a bridged robot prompts instead of failing closed. I verified on the head commit: the interactive path is sound against the installed strands SDK (`InterruptException` subclasses `Exception` directly, so it propagates past `use_ros`'s narrow except clause and releases `_backend.lock` on unwind; the interrupt id is scoped per `toolUseId`, so a stored approval cannot leak across invocations), the gate fails closed with a refusal that names both env vars, canonicalisation is comparison-only so the caller's literal string is what reaches rclpy, and the prior review rounds (bridge context threading in `f9450333`, README allowlist row in `51d79593`, halt-wording correction in `3e33864`) are all present on this head with their pin tests. The four touched test files pass (234 tests), as do the tool-parameter-description, Args-completeness, ROS command-surface security, changelog-format and source-hygiene suites (551 more).

## What's good

- The blocklist/gate tests drive the public tool dispatch rather than only the private helpers - exactly the gap that made the original verb holes invisible - plus AST structural guards so a fourth command verb or a contextless bridge tool cannot ship ungated.
- The `stop()`-is-gated decision (surface-keyed, not payload-keyed) is argued, documented in three places, and given an unattended-halt remedy rather than a carve-out; the case-sensitivity decision is pinned as a test so it is not "fixed" later.
- Both gate-bypassing env vars are cleared per-test (`monkeypatch.delenv`), so an ambient `BYPASS_TOOL_CONSENT` in a CI shell cannot make the gate assertions pass vacuously.
- Changelog fragment with a migration note, README env-var row, and examples updated to pre-approve exactly the one topic they drive.

Note for a follow-up (not blocking, pre-existing on `main`): `use_rtps` and `use_rosbridge` can publish to the same safety-critical surfaces and carry no equivalent gate; `docs/security.md`'s new bullet is correctly scoped to `use_ros`, but the sibling transports are worth a tracked issue.

PR #2314, opened 2026-08-16
docs(policies): send a documented per-call goal to the sink that reads it

cagataycali:
## Problem

A fenced ```python block is code a reader copies. Six of them passed a keyword
the callable does not accept, so the documented example raises `TypeError` on
the first line the reader runs - and nothing in the repository noticed, because
the docs build renders the block verbatim and the suite never executed it.

| Location | Written | Reality |
| --- | --- | --- |
| `docs/policies/curobo.md:112` | `run_policy(..., target_pose=[...])` | `TypeError: MuJoCoSimEngine.run_policy() got an unexpected keyword argument 'target_pose'` |
| `docs/policies/moveit2.md:146` | `run_policy(..., target_pose=[...])` | same |
| `docs/policies/wbc_gait.md:32` | `run_policy(..., target_velocity=..., gait_frequency=...)` | same, both keywords |
| `docs/policies/groot.md:96-98` | `gr00t_inference(..., tag=...)`, `model_id=...` | no `tag` / `model_id` parameter, and neither is in the tool schema |

`run_policy` has two forwarding sinks and no `**kwargs`: `policy_config` is
expanded into the policy **constructor**, and `policy_kwargs` is handed to every
`get_actions()` call. `CuroboPolicy`, `MoveIt2Policy` and `WBCGaitPolicy` all
read their goal from `get_actions(**kwargs)`, so `policy_kwargs` is the only
sink that reaches them. Written directly on `run_policy` the goal is not a
slower path - it is a `TypeError`.

`gr00t_inference` has no `tag` parameter *by design*: `_resolve_image_name()`
takes the image from `STRANDS_GR00T_IMAGE` and checks it against an allowlist
precisely so an agent cannot choose it ("The agent has no say in the image").
The documented `tag=` advertised caller control over a surface that is
deliberately closed, and never told the reader about the real knob. The
checkpoint keyword is `hf_repo`.

Three prose passages additionally claimed `Robot.start_task` flows a goal
through a `**policy_kwargs`:

> flows the same `target_pose` / `target_joints` kwargs through `start_task`'s `**policy_kwargs`

`start_task`'s parameters are `instruction`, `policy_port`, `policy_host`,
`policy_provider`, `duration` - no variadic keywords, and no goal. `mesh.tell`
*does* accept the goal vocabulary, so the corrected text points there instead.

## Change

- The three `run_policy` examples now write `policy_kwargs={"target_pose": ...}`.
- `groot.md` drops the fictional `tag=`, documents `STRANDS_GR00T_IMAGE` +
  `STRANDS_GR00T_IMAGE_ALLOW` as the operator knob, and uses `hf_repo=`.
- The `start_task` claim is replaced in `curobo.md`, `moveit2.md` and `README.md`
  with what each entry point actually accepts.
- `tests/test_docs_python_examples_are_callable.py` grades every `python` fence
  in `docs/**/*.md` and `README.md` against the real signatures of the public
  surface (module functions, `@tool` entry points unwrapped via `__wrapped__`,
  and the public methods of `SimEngine` / the three backends / `Robot` /
  `Policy` / `DatasetRecorder` / `Mesh`). A call is graded only when its callee
  name resolves to a known public callable, so a name outside the library
  degrades to "not graded", never to a false failure.

## Verification

Pre-fix, with only the test file added on top of `main`: **1 failed / 3 passed**.
The failure names every offender with `file:line`:

```
docs/policies/curobo.md:112 run_policy(...) passes ['target_pose'], which no public run_policy accepts
docs/policies/groot.md:96 gr00t_inference(...) passes ['tag'], which no public gr00t_inference accepts
docs/policies/groot.md:97 gr00t_inference(...) passes ['model_id'], which no public gr00t_inference accepts
docs/policies/groot.md:98 gr00t_inference(...) passes ['model_id', 'tag'], which no public gr00t_inference accepts
docs/policies/moveit2.md:146 run_policy(...) passes ['target_pose'], which no public run_policy accepts
docs/policies/wbc_gait.md:32 run_policy(...) passes ['gait_frequency', 'target_velocity'], which no public run_policy accepts
```

Post-fix: 4 passed. 296 documented calls are graded and 0 offend. The three
tests that pass on both trees are the no-overreach controls - `run_policy` has
no goal parameter and does have `policy_kwargs`; `start_task`'s parameter set is
exactly its five; and the grader still reports a planted bad keyword, so a clean
sweep means the documentation is right rather than that the grader accepts
everything.

## Rollout

The corrected `moveit2.md` example, run headless in MuJoCo against a stub
sidecar (`panda`, 90 ticks at 50 Hz, `n_steps=90`). Left panel is the call
exactly as the documentation read it; right panel is the corrected spelling.

documented goal-kwarg sink

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-doc-goal-kwarg-sink/goal_kwarg_sink.mp4)

| | Before (`target_pose=`) | After (`policy_kwargs={...}`) |
| --- | --- | --- |
| Outcome | `TypeError` before tick 0 | `status=success`, 90 ticks |
| Rendered frames | 1 (static - nothing ran) | 45, every one moving (per-frame mean abs pixel delta 0.026-1.65) |
| Mean joint error vs goal | 0.8357 rad (unchanged) | **0.0024 rad** |

Panel-to-panel mean abs pixel delta on the final frame is 4.42.

## Gate

`ruff check` / `ruff format --check` clean on `strands_robots tests tests_integ`.
`mypy` 18 pre-existing errors, identical on this branch and `main`. Full
`tests/` run compared against a `main` worktree on the same interpreter - no new
failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -816,10 +816,13 @@ actions = policy.get_actions_sync(
 )
 ```
 
-Agents share one goal vocabulary across VLA and planner providers:
-`Robot.start_task(..., policy_provider="curobo", target_pose=[...])` and
-`mesh.tell(peer, "...", policy_provider="curobo", target_pose=[...])` flow the
-same `target_pose` / `target_joints` / `world_update` kwargs through.
+`mesh.tell(peer, "...", policy_provider="curobo", target_pose=[...])` forwards
+the same `target_pose` / `target_joints` / `world_update` vocabulary to a sim
+peer. In-process, the goal goes to `run_policy(policy_kwargs={...})`, which the
+runner hands to every `get_actions()` call - `run_policy` itself has no
+`target_pose` parameter. `Robot.start_task` takes no goal payload: its
+parameters are `instruction`, `policy_port`, `policy_host`, `policy_provider`
+and `duration`.
 
 </details>
 
--- a/changelog.d/2314-documented-goal-kwarg-sink.md
+++ b/changelog.d/2314-documented-goal-kwarg-sink.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- Documented `run_policy` examples for the `curobo`, `moveit2` and `wbc_gait`
+  providers wrote a per-call goal (`target_pose`, `target_velocity`,
+  `gait_frequency`) directly as a `run_policy` keyword. `run_policy` has no such
+  parameter and no `**kwargs`, so the example raised `TypeError` before the first
+  tick; the goal belongs in `policy_kwargs`, which the runner hands to every
+  `get_actions()` call. The `gr00t_inference` lifecycle example passed `tag=` and
+  `model_id=`, neither of which exists - the image name is operator config
+  (`STRANDS_GR00T_IMAGE`, allowlist-checked) and the checkpoint keyword is
+  `hf_repo`. Three prose passages claimed `Robot.start_task` forwards a goal
+  through a `**policy_kwargs` it does not have. A new guard grades every `python`
+  fence in `docs/` and `README.md` against the real signatures.
--- a/docs/policies/curobo.md
+++ b/docs/policies/curobo.md
@@ -114,17 +114,22 @@ sim.run_policy(
     instruction="",               # ignored by the planner
     policy_provider="curobo",
     policy_config={"robot_config": "franka.yml", "action_horizon": 16},
-    target_pose=[0.5, 0.0, 0.4, 1.0, 0.0, 0.0, 0.0],
+    policy_kwargs={"target_pose": [0.5, 0.0, 0.4, 1.0, 0.0, 0.0, 0.0]},
     duration=10.0,
     control_frequency=50.0,
 )
 ```
 
-The LLM-agent demo path
-(`Robot.start_task(..., policy_provider="curobo", target_pose=[...])`) flows
-the same `target_pose` / `target_joints` kwargs through `start_task`'s
-`**policy_kwargs`, so agents share one goal vocabulary across VLA and planner
-providers.
+`policy_config` and `policy_kwargs` are two different sinks. `policy_config`
+is expanded into the policy **constructor**; the per-call goal belongs in
+`policy_kwargs`, which the runner hands to every `get_actions()` call. Passing
+`target_pose=` directly to `run_policy` raises `TypeError` - it has no such
+parameter and no `**kwargs`.
+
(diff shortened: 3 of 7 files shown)


yinsong1986 (approved):
## Summary

Docs-only correction plus a guard test: six fenced `python` examples across `docs/policies/{curobo,moveit2,wbc_gait,groot}.md` passed keywords their callees do not accept (`target_pose=`/`target_velocity=`/`gait_frequency=` straight on `run_policy`, and fictional `tag=`/`model_id=` on `gr00t_inference`), and three prose passages claimed a `Robot.start_task` `**policy_kwargs` that does not exist. The fixes route the per-call goal through `policy_kwargs` (the one sink `run_policy` forwards to `get_actions()`), document `STRANDS_GR00T_IMAGE`/`STRANDS_GR00T_IMAGE_ALLOW` as the operator knob with `hf_repo=` for the checkpoint, and correct the `start_task` prose to its actual five parameters. The new `tests/test_docs_python_examples_are_callable.py` grades every documented keyword call against real public signatures, degrades unknown names to "not graded" rather than false failure, asserts a corpus floor so a silently-empty sweep cannot pass, and includes a planted-offender control. I independently reproduced the verification: with `main`'s docs restored under the new test, it fails naming exactly the six offenders with `file:line`; on the PR tree all 4 tests pass. Every corrected claim checks out against the live signatures (`run_policy` has `policy_kwargs` and no `**kwargs` or goal params; `Robot.start_task` has exactly the five named parameters and no `VAR_KEYWORD`; `gr00t_inference` has `hf_repo` and no `tag`/`model_id`; `Mesh.tell` takes `**kw`; the `STRANDS_GR00T_IMAGE*` env vars exist in `gr00t_inference.py`). Changelog fragment present per convention, ruff check/format clean on the new file, no host paths or non-ASCII, no public API or behavioural surface touched.

## What's good

- The guard test cannot rot silently: the `_MINIMUM_GRADED_CALLS` floor fails if the extractor stops reaching the docs, and the planted-offender test fails if the grader starts accepting everything.
- Unknown callee names degrade to "not graded" rather than a false failure, so third-party or illustrative snippets cannot break the suite.
- The `groot.md` correction stops advertising agent control over the container image - a surface `_resolve_image_name()` deliberately closes - and documents the real operator knob instead.

PR #630, opened 2026-06-24
fix(ux): smooth out first-run paper cuts for strands_robots users

cagataycali:
## Summary

A fresh `Robot("so101")` + a simple agent loop surfaced several rough edges that make the **very first experience noisy and confusing**. This PR addresses all of them so the out-of-box flow is quiet and forgiving.

### 1. Mesh OFF by default (quiet by default)
- `Robot(...)` no longer spins up Zenoh/ACL/e-stop machinery the user never asked for - the security-warning wall disappears for the common case. The `mesh` argument now defaults to `None` (a sentinel meaning "consult the env").
- `STRANDS_MESH=true` opts a bare `Robot()` in without a code change; `STRANDS_MESH=false` is a hard kill switch. An explicit `mesh=True`/`mesh=False` always wins over the env default - the env var only ever forces mesh OFF inside `init_mesh`, never ON, so an explicit opt-out is always honoured.

### 2. Human-readable mesh safety warnings
- Rewrote the 3 mesh warnings (permissive ACL, e-stop override code, multicast) from jargon walls into plain-language guidance with a clear menu of options, interpolating the peer id into the `[safety:<id>]` / `[mesh:<id>]` prefix.
- Surfaces the `STRANDS_MESH_LOCAL_DEV=true` developer escape hatch (turns off mTLS+ACL for localhost work).

### 3. Suppress benign MuJoCo attach/compile noise
- New `filter_mujoco_attach_noise()` context manager catches both the Python `UserWarning` (`spec.to_xml`) and raw C-level fd-2 spam (`Attach conflict...`, `timestep: parent has...`). Benign lines drop to DEBUG; everything else passes through untouched.
- Opt back in with `STRANDS_ROBOTS_VERBOSE_MUJOCO=1`.
- Wrapped attach/compile/recompile/to_xml hot paths in `scene_ops.py` and `simulation.py`. Demoted `assets/manager` "Unknown robot" log from WARNING to DEBUG (normal `resolve_model` probe control-flow now).

### 4. Forgiving `add_robot` + model resolution
- `add_robot`: `name=` is now OPTIONAL, auto-derived from `data_config` / URDF stem and auto-numbered on collision (`so101`, `so101_2`, ...). Removes the `requires parameter 'name'` first-call failure.
- `resolve_model`: strips common decorative suffixes (`_default`/`_sim`/`_robot`/`_arm`) and retries, so the natural guess `so101_default` resolves to `so101` without a `list_urdfs` round-trip.
[...]

PR #371, opened 2026-06-08
docs: rewrite README for v0.x (Robot() factory, mesh, sim) + future-proof counts

cagataycali:
## Summary

Refreshes the README to match the current v0.x API surface and removes hardcoded counts that go stale on every robot/action addition.

### Changes
- Rewrite README around the `Robot()` factory, sim-first defaults, supported-robot registry, tools reference, and policy providers
- Document the MuJoCo `Simulation` backend and its action set
- Add Mesh networking section (Zenoh peers, `tell()`, E-STOP, CA pin rotation runbook)
- Add Security section linking `SECURITY.md`
- Add GR00T data-config schema table
- **Future-proof counts**: replace exact `68 robots` / `64 actions` with `50+` so docs don't need touching every time the registry or sim action set grows

### Notes
- Single-file change (`README.md`)
- No code or behavior changes

PR #2316, opened 2026-08-16
docs(mesh): state the STRANDS_MESH opt-in contract the factory implements

cagataycali:
## Problem

`STRANDS_MESH` is documented in three configuration tables, all printing `true` as its default:

| Location | Description | Default |
|---|---|---|
| `README.md:1092` | Set `false` to disable Zenoh mesh globally | `true` |
| `docs/api-reference.md:195` | Disable mesh globally when `false` | `true` |
| `docs/getting-started/installation.md:111` | Mesh **enabled by default**; set to `false` to disable globally | `true` (enabled) |

The factory resolves it the other way round - mesh is off unless it is asked for:

```python
if mesh is None:
    mesh = os.getenv("STRANDS_MESH", "").strip().lower() in ("true", "1", "yes")
```

Measured with the variable unset, `Robot("so101", mode="sim").mesh is None`.

So each row is wrong in three ways at once:

- the printed default names a state a bare `Robot()` is never in;
- the offered knob ("set `false` to disable") is a no-op for the state the reader is actually in;
- `true`, the one spelling that enables the mesh, is presented as something you already have, so the opt-in is undiscoverable from the tables.

The drift is dated. `mesh: bool = True` became `mesh: bool | None = None` in #630 (quiet-by-default); the tables date from #371 and were not brought along. The suite already pins the current behaviour - `tests/mesh/test_mesh_wiring.py::test_robot_factory_env_opt_in_enables_mesh`, whose docstring says "mesh defaults to None (quiet), so the factory consults STRANDS_MESH". Only the tables disagree.

## Change

- The three rows now state the actual contract: unset means mesh off, `true`/`1`/`yes` opts in, `false`/`0`/`no` is a hard kill switch that also overrides an explicit `mesh=True`.
- `STRANDS_MESH` is documented in `strands_robots.robot`'s `Environment Variables:` docstring section, which previously listed only `STRANDS_ROBOT_MODE` even though this module is where the variable is resolved.
- The resolution moves into `_mesh_env_opt_in()`. Behaviour is unchanged - the expression is moved, not modified - so the guard below grades the tables against the resolver itself instead of a hand-copied spelling list. Widening the accepted spellings later therefore cannot silently invalidate the guard.

## Tests

`tests/mesh/test_mesh_env_opt_in_documented_default.py`. For every single-variable `STRANDS_MESH` row it takes each token in the Default cell, sets `STRANDS_MESH` to that token, and asserts the resolver still reports off.

| | pre-fix | post-fix |
|---|---|---|
| `test_no_table_prints_an_opt_in_spelling_as_the_default` | fail, naming all 3 rows | pass |
| 18 remaining cases | pass | pass |
| total | 1 failed, 18 passed | 19 passed |

Pre-fix message (graded against `main`'s own inlined expression, transplanted verbatim so the verdict is `main`'s):

```
README.md:1092 prints 'true' as the default for STRANDS_MESH, but STRANDS_MESH=true
opts IN. A bare Robot() has mesh OFF, so this row tells a reader mesh is already
running and hides the opt-in.
docs/api-reference.md:195 prints 'true' ...
docs/getting-started/installation.md:111 prints 'true' ...
```

The 18 cases that pass on both trees are the controls:

- the resolver's own domain, including `on` and `enabled`, which are *not* opt-in spellings and so must not be documented as if they were;
- a premise assert that the extractor still reaches at least three rows, so a reformat that hides the tables fails loudly rather than reporting clean;
- every row must still name a value that turns mesh on, so correcting the default cannot remove the only documented way to enable the mesh.

Behaviour is unchanged, so there is no rollout or render to attach. `ruff check`, `ruff format --check` and `mypy` are clean over `strands_robots tests tests_integ` (18 pre-existing errors, none in the touched files); `tests/mesh` plus `tests/test_robot_factory.py` report an identical 45-item failure set on this branch and on `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1089,7 +1089,7 @@ touches ROS 2.
 | `STRANDS_ISAAC_RTX_PATHTRACING` | Isaac Sim backend: on (`1`/`true`/`yes`/`on`) enables RTX path-tracing (photorealistic, slow) instead of the default render mode; off leaves the render mode alone, any other spelling is refused | unset |
 | `STRANDS_ISAAC_NUCLEUS_URL` | Isaac Sim backend: override the Omniverse Nucleus asset-server URL | unset (Isaac default) |
 | `GROOT_API_TOKEN` | API token for the GR00T inference service | unset |
-| `STRANDS_MESH` | Set `false` to disable Zenoh mesh globally | `true` |
+| `STRANDS_MESH` | Opt a bare `Robot()` into the Zenoh mesh: `true`/`1`/`yes` turns it on. `false`/`0`/`no` is a hard kill switch that also overrides an explicit `mesh=True` | unset (mesh off) |
 | `STRANDS_MESH_LOCAL_DEV` | Set `1` for a one-var localhost preset (auth `none`, no second factor needed) | unset |
 | `STRANDS_ROS2_BRIDGE_I_KNOW_THIS_IS_INSECURE` | Second factor to expose a `Robot(ros2_transport="rtps")` inbound `joint_command` surface with no `dds_security_config` (DDS Security). Truthy: `1`/`true`/`yes` | unset |
 <details>
--- a/changelog.d/2316-mesh-env-opt-in-contract.md
+++ b/changelog.d/2316-mesh-env-opt-in-contract.md
@@ -1,0 +1,10 @@
+### Docs: `STRANDS_MESH` is documented as the opt-in it actually is
+
+Three configuration tables printed `true` as the default for `STRANDS_MESH` and
+offered `false` as the only knob. The factory resolves `mesh=None` from the
+environment, so unset means mesh off and only `true`/`1`/`yes` turns it on: the
+printed default named a state a bare `Robot()` is never in, the offered remedy
+was a no-op for the state readers were actually in, and the one spelling that
+enables the mesh appeared nowhere. The rows now state the real contract, the
+module that resolves the variable documents it, and a guard grades the tables
+against the resolver so the two cannot drift apart again.
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -192,7 +192,7 @@ LIBERO task suites, BDDL parser. Install: `uv pip install "strands-robots[benchm
 | `STRANDS_ASSETS_DIR` | Robot model asset cache | `~/.strands_robots/assets/` |
 | `STRANDS_ROBOT_MODE` | Force `Robot()` mode | (kwarg honoured) |
 | `STRANDS_TRUST_REMOTE_CODE` | Allow HF `trust_remote_code=True` | unset → blocked |
-| `STRANDS_MESH` | Disable mesh globally when `false` | `true` |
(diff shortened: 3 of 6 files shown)


yinsong1986 (approved):
## Summary

Corrects three configuration-table rows that documented `STRANDS_MESH` with a default of `true` when the factory actually treats mesh as opt-in (`mesh=None` resolves to off unless `STRANDS_MESH` is `true`/`1`/`yes`). The resolution expression is moved verbatim into a private `_mesh_env_opt_in()` helper (behaviour unchanged, not exported), the module docstring's `Environment Variables:` section now documents the variable where it is resolved, and a new guard test grades the shipped tables against the resolver itself so the docs cannot drift from the code again. I verified the corrected contract against the code: the opt-in resolution in the `Robot` factory and the `false`/`0`/`no` hard kill switch in `strands_robots/mesh/core.py::init_mesh` both match the new wording, including the asymmetry that the env var can force mesh off but never on.

## What's good

- The guard test imports the resolver as its oracle instead of hand-copying the accepted spellings, and carries a premise assert that the extractor still reaches at least three rows - so neither widening the spellings nor reformatting the tables can silently defuse it.
- Changelog fragment present (`changelog.d/2316-mesh-env-opt-in-contract.md`) and passes `scripts/check_changelog_fragment.py` against `main`.
- Tests use `monkeypatch.setenv`/`delenv` per AGENTS.md testing conventions; the new test file passes (19/19) on the branch, and ruff check/format are clean on the touched files.
- Scope discipline: the factory's behaviour is untouched, the helper stays private, and the pre-fix failure was demonstrated against `main`'s own inlined expression.

yinsong1986 (approved):
## Summary

Docs-only correction with a drift guard: three configuration tables (`README.md`, `docs/api-reference.md`, `docs/getting-started/installation.md`) printed `true` as the default for `STRANDS_MESH`, while the factory has resolved `mesh=None` as opt-in since #630 - unset means mesh off, and only `true`/`1`/`yes` enables it. The rows now state the actual contract, the module docstring for `strands_robots.robot` documents the variable where it is resolved, and the resolution expression moves verbatim into a private `_mesh_env_opt_in()` helper so the new test can grade the tables against the resolver itself rather than a hand-copied spelling list. I verified the claims against the tree: the helper's body is byte-identical to the expression it replaces, and the `false`/`0`/`no` hard-kill-switch wording matches `init_mesh`'s actual check in `strands_robots/mesh/core.py` (env only ever forces mesh OFF there, never ON). Behaviour is unchanged and nothing public is exported.

## What's good

- The guard test derives its oracle from `_mesh_env_opt_in()` directly, so widening the accepted spellings later cannot silently invalidate it, and the premise assert (`>= 3` rows reached) makes a table reformat fail loudly instead of reporting clean.
- Test hygiene follows the house rules: `monkeypatch.setenv`/`delenv`, no host paths, ASCII-only strings, and the corrected rows are also required to still name an opt-in spelling so the fix cannot remove the only documented way to enable the mesh.

PR #2317, opened 2026-08-16
fix(simulation): diagnose a missing asset instead of suggesting the name it refused

cagataycali:
## Problem

`add_robot()` (and the `data_config` resolution behind it) reports two different failures through one message, `MuJoCoSimEngine._unknown_model_msg`:

1. the registry does not know the name - a typo;
2. the registry **does** know the name and its model XML is simply not on disk.

Only case 1 is a spelling problem, but both got spelling advice. And because the suggestion is drawn from the **whole registry**, case 2 puts the requested name inside the candidate set - `difflib.get_close_matches` scores an exact match 1.0 and ranks it first, so the one caller whose name needs no fixing is the one told to fix it:

```
>>> Simulation(...).add_robot(name="google_robot")
No model found for 'google_robot'. Did you mean: google_robot, toddlerbot_2xm, toddlerbot_2xc?
Use action='list_urdfs' to see all available robots. ...
```

`google_robot` is registered. The actual cause is that `google_robot/robot.xml` is not on any asset search path, and its registry entry sets `asset.auto_download: false`, so it is never fetched automatically. None of that is in the message - and the self-match also burns one of only three suggestion slots.

Reproducible out of the box on `google_robot` and `trossen_wxai` (the two `auto_download: false` entries, both advertised by `list_robots(mode="sim")`); any registered robot whose assets are not downloaded reaches the same path.

## Change

**`_unknown_model_msg` splits the two conditions.** A registered robot now says it is registered, names the `<dir>/<model_xml>` the resolver looked for and every path it searched, and gives the remedy its own entry implies - an `auto_download: false` entry is never fetched automatically (place the file), anything else is fetchable:

```
Robot 'google_robot' is registered but its model file is not on disk: no
'google_robot/robot.xml' under '/home/.../.strands_robots/assets', '.../assets'.
This entry declares auto_download=false, so its asset is never fetched
automatically - create that directory and place 'robot.xml' inside it.
```

An alias resolves to the canonical entry's asset, so `add_robot(name="oxe_google")` names `google_robot/robot.xml`. A typo keeps today's message unchanged. This mirrors the wording `register_robot` already uses when it refuses a missing asset directory at registration time.

**The suggestion routes through `close_match_hint`.** That helper in `simulation/base.py` is the single owner of the `" Did you mean: ...?"` fragment - the `_unknown_object_msg` / `_unknown_camera_msg` / `_unknown_robot_msg` / `_unknown_mj_entity_msg` helpers all use it, and `_unknown_model_msg` was the one that reimplemented it with a second inline `difflib` call. It now uses the helper, and the helper drops a suggestion identical to `requested`.

Putting the rule in the helper rather than at the call site removes an unstated precondition: the other callers pass what the world holds, which cannot contain the requested name (that is why the message is being built), so for them it was unreachable - but nothing said so, and the one caller comparing against a wider catalogue hit it. The helper asks `difflib` for one extra candidate so dropping a self-match **promotes** the next one instead of shortening the list, which leaves those callers byte-identical.

## Verification

Regression suite `tests/simulation/mujoco/test_unresolved_model_message_names_the_missing_asset.py`, on `upstream/main` -> this branch: **8 failed / 10 passed -> 18 passed**. Pre-fix failures read as the finding, e.g.

```
assert 'registered' in "No model found for 'google_robot'. Did you mean: google_robot, ..."
```

The 10 that pass on both trees are the over-reach controls: a typo still gets `so101`/`so100`, a name with no neighbours still omits the clause, an unreadable registry still degrades to the bare form, a non-`str` name is still reported rather than raising, the message stays ASCII, and `close_match_hint`'s rendering is unchanged for a `known` set that excludes the requested name (compared against the plain three-match call it made before, over three cases, so the control encodes no `difflib` ordering of its own).

The asset-absent condition is produced by pointing `STRANDS_ASSETS_DIR` at an empty `tmp_path` and using the real resolver, not by doubling the presence probe, so the message under test is built from the same filesystem answer a user with un-downloaded assets gets. A premise test asserts both subjects still carry the `auto_download` posture their case tests, so editing `robots.json` fails with a readable reason instead of making a case vacuous.

`ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` identical to `upstream/main` (14 pre-existing errors, 0 new). Full `tests/` run on both trees with the same interpreter (MuJoCo 3.10.0, lerobot 0.6.2): `upstream/main` 5 failed / 30570 passed, this branch 5 failed / 30588 passed - the sorted `FAILED`/`ERROR` id sets are identical (7 ids, all environment-bound: docker-absent vera cases, the GL-probe and pytest-exit meta-tests, a Newton case, a mesh node-id case), and `+18 passed` is exactly the new suite. `tests/simulation/mujoco/test_simulation.py` (which holds the existing pins on this message), `test_unknown_entity_message_name_type.py` (the AST guard requiring these helpers to use `close_match_hint` and branch on `known` alone) and `tests/registry` pass: 978 passed.

## The refusal's remedy, executed

The path is machine-extractable from the message. Parsing it out, placing a model there and re-running `add_robot` loads and steps the robot - MuJoCo 3.10.0, headless EGL:

add_robot missing-asset refusal, before and after, plus the remedy applied

The real Google Robot MJCF is not redistributable, which is why that entry is `auto_download: false`; the model in the render is an operator-authored stand-in. What it demonstrates is that the path the refusal names is the path the loader reads.


yinsong1986 (dismissed):
## Summary

Splits `MuJoCoSimEngine._unknown_model_msg` into the two conditions it was conflating: an unregistered name (typo, keeps the `close_match_hint` spelling suggestion) versus a registered robot whose model XML is absent from every asset search path (now names the exact `<dir>/<model_xml>` it looked for, every path searched, and the remedy keyed on the entry's `auto_download` posture). The self-match filter moves into the shared `close_match_hint` helper with an n=4-then-slice so the other four callers' rendered output stays byte-identical, which the parametrized control test pins against the pre-change `difflib` call rather than a hand-written expectation. Verified locally at the head SHA: the new 18-test suite passes, the adjacent pins (`test_simulation.py`, `test_unknown_entity_message_name_type.py`, `test_add_robot_unknown_model_message.py`, `test_missing_entity_error_messages.py`, `tests/registry`, `test_refusal_messages_never_raise.py`) pass, `ruff check` / `ruff format --check` clean on the changed files, and a live smoke of all four message paths (`google_robot`, `panda`, `so10x`, non-str `42`) against an empty `STRANDS_ASSETS_DIR` reproduces the exact wording the description claims. The asset probe is a pure filesystem check (`is_robot_asset_present` does no download or network), both probes degrade to the bare form on any failure per the never-raise contract, and the two hardcoded `auto_download=false` subjects in the docs table (`google_robot`, `trossen_wxai`) match the registry, with a premise test guarding against the registry drifting under the suite.

## What's good

- The non-vacuity premise test (`test_the_registry_postures_these_tests_assume_still_hold`) means a future `robots.json` edit fails with a readable reason instead of silently hollowing out a case.
- The asset-absent condition is produced through the real resolver against an empty `STRANDS_ASSETS_DIR` rather than by stubbing the presence probe, so the tested message is built from the same filesystem answer a user gets.
- The self-match rule lands in the one helper that owns the `" Did you mean"` fragment instead of at the call site, removing an unstated precondition the other callers were silently relying on.

One non-blocking note: this branch ships no `changelog.d/2317-*.md` fragment (AGENTS.md step 3); every recently merged PR carries one, and a user-visible error-message change is the kind of entry fragments record. Trivial to add before merge.

yinsong1986 (approved):
## Summary

Splits `MuJoCoSimEngine._unknown_model_msg` into the two failure conditions it previously conflated: an unregistered name (typo) keeps the spelling suggestion, while a registered robot whose model XML is absent from every asset search path now gets a diagnosis naming the exact `<dir>/<model_xml>` the resolver looked for, the paths it searched, and the remedy keyed on the entry's own `auto_download` posture (place the file for `auto_download: false`, `download_assets` otherwise). The self-suggestion bug (`difflib` ranking the exact match first, so the correctly-spelled name was offered as its own correction) is fixed at the shared `close_match_hint` helper, which now drops a suggestion identical to the requested name and fetches one extra candidate so the rendered list stays three-long — leaving every other caller byte-identical, which the new controls pin. Both probe blocks are best-effort and degrade to the prior bare message on any failure, so no new raise path reaches `add_robot` callers. The diagnosis reuses the real resolver (`is_robot_asset_present`, `get_search_paths`) rather than doubling the presence check, and the imports verify against the tree. No wire formats, public signatures, or persisted schemas are touched; the changelog fragment and troubleshooting-table row are included per AGENTS.md.

## What's good

- The premise test (`test_the_registry_postures_these_tests_assume_still_hold`) keeps the suite from going vacuous if `robots.json` changes the `auto_download` posture of either subject.
- The self-match rule lives in `close_match_hint` rather than at the call site, with explicit controls asserting the other callers' rendering is unchanged against the plain `difflib` call rather than a hand-written ordering.
- The asset-absent condition is produced with `STRANDS_ASSETS_DIR` pointed at an empty `tmp_path` through the real resolver, so the message under test is built from the same filesystem answer a user gets.

PR #2318, opened 2026-08-16
fix(robot): report what came up when a Device Connect bring-up fails

cagataycali:
## Problem

`Robot(...).run()` is the documented Device Connect server mode. It stops the built-in
Zenoh mesh to make way for Device Connect, then brings the runtime up. The bring-up was
wrapped so that a failure keeps the process alive, which is deliberate. What followed was
not: the status line was printed unconditionally.

So on the one path where the process is reachable over **nothing at all** - the mesh has
been stopped for a replacement that never arrived - the last thing the operator was told
was that the device is online.

The absent `[device-connect]` extra is the common way to reach it. Captured from
`Robot("so101").run()` with the extra genuinely not installed, on `main`:

```
mesh before run(): Mesh(peer_id='so101_sim-9392d99b', type='sim', sim)
WARNING  Device Connect init failed: No module named 'device_connect_edge'
so101_sim-9392d99b is online. Ctrl+C to stop.
   mesh after run():            None
   device connect runtime:      None
   process still looping (run() has not returned)
```

The runtime is `None`, the mesh is `None`, and stdout says the device is online. The
warning sits two lines above the claim that contradicts it, and it names
`device_connect_edge` - the distribution's internal module, not the extra that installs it.

This is already written down in the repository. `test_an_expired_budget_is_reported_to_the_operator`
says a failed bring-up previously "logged nothing at all, leaving the runner's 'is online'
line as the only thing said about a runtime that never came up". The fix at the time added
the warning next to the line. A warning next to a contradicting claim still leaves the claim.

## Change

`_run_device_connect_foreground` now prints the line that matches the runtime, and names
which transport the process no longer has:

```
WARNING  Device Connect init failed: No module named 'device_connect_edge'. Install it with: pip install 'strands-robots[device-connect]'.
so101_sim-8dc8a75c is NOT online: the Device Connect runtime did not start (see the warning
above). The built-in mesh was stopped for it, so this process now serves no transport.
Ctrl+C to stop.
   mesh after run():            None
   device connect runtime:      None
   process still looping (run() has not returned)
```

Both transcripts are the same script against the two trees; only the two printed lines
differ. Specifically:

- The success path is byte-identical - `<peer-id> is online. Ctrl+C to stop.`
- Keeping the process alive is unchanged. A broker that comes back is not worth losing the
  process over, and the loop, the Ctrl+C handling and the shutdown lines are untouched.
- The two failure cases are distinguished, because which transport is gone is the
  operator's next step: a robot that had a mesh is told it was stopped, a robot that never
  had one is not told about a mesh it never ran.
- The warning names the extra when an `ImportError` is what failed, and only then - an
  unreachable broker is not fixed by installing anything, so that message is unchanged.

## Tests

`tests/test_robot_factory.py`, `TestTheStatusLineReportsWhatCameUp`. Against `main`:
**5 failed, 104 passed**; with the change, **109 passed**. The failures read

```
assert 'arm-1 is NOT online' in 'arm-1 is online. Ctrl+C to stop.\n...'
assert 'strands-robots[device-connect]' in 'Device Connect init failed: import of strands_robots.device_connect halted; None in sys.modules'
```

The passing half is the boundary: a started runtime is still announced online, the ASCII-only
contract still holds on both branches, the mesh is still stopped and detached, and the
lifecycle lines are unchanged.

`test_device_connect_init_failure_keeps_process_alive` asserted `"so100-dc is online"` on the
failed path. Its subject - the loop keeps running rather than exiting - is unaffected; it now
asserts the truthful line plus the shutdown lines, so what it pins is still that the process
survived.

### The foreground tests were environment-dependent

They drove the shipped `from strands_robots.device_connect import ...` unstubbed, with a
comment stating that "a missing backend is logged and the loop still prints its lifecycle
lines". That only holds where the extra is absent. Where it is installed - CI runs
`[all]`, which includes it - the same tests brought up a **real Zenoh D2D session inside the
unit test** (`Connected to ZENOH broker: []`, `Driver connected: strands_sim`), and one test
failed to collect at all without the extra because it monkeypatched through the module.

They now substitute the integration module: a stub returning a runtime for the success
branch, `None` in `sys.modules` for the ImportError branch. Both branches are selectable by
the test rather than by the environment, the class needs no optional dependency, and it opens
no sockets. Verified green both with the extra installed and without it.

## Docs

`docs/device-connect.md` records that `.run()` is the one path the extra is required for, and
what it reports when the bring-up fails.

## Before / after

run() status line before and after

Both panels are the same script - `Robot("so101").run()` with the `[device-connect]` extra
genuinely absent - run against `main` and against this branch.

## Regression check

Full `tests/` on both trees: **61 failing ids, the same set on the branch and on `main`**
(`comm` empty in both directions; all pre-existing, absent optional extras such as `awsiot`).
30336 passed on the branch against 30330 on `main` - the difference is the new tests.
`ruff check` / `ruff format --check` / `mypy` over `strands_robots tests tests_integ` are
clean, with mypy at its pre-existing 18 and none of them in the touched files.

### Round 2 - merge base advanced, composition measured

No source change. `Detect an untested overlap with the base branch` went red on `3ff79219`
for a reason outside the diff: the branch's merge base was `7c54e3ff` (#2314), `main` had
since moved two commits, and one of them - #2316 - edits `strands_robots/robot.py`, which
this branch also edits. The intersection was that one file, so the check was correct that
the combination had never been compiled.

The overlap is disjoint *within* the file: #2316 extracts `_mesh_env_opt_in()` and calls it
from the `Robot(...)` factory; this branch only touches
`_run_device_connect_foreground`. `git merge origin/main` is textually clean with no
conflict, and both changes are present at the merged head (`_mesh_env_opt_in` at
`robot.py:123`, the runtime-gated status line at `robot.py:503-531`).

A clean text merge is not the answer to the question the check asks, so the composition was
run rather than inferred. `tests/test_robot_factory.py` plus #2316's own guard
(`tests/mesh/test_mesh_env_opt_in_documented_default.py`), same interpreter, both trees:

| tree | result | failing ids |
|---|---|---|
| `origin/main` (`4779077e`) | 2 failed, 68 passed, 52 skipped | `..._unsupported_type_error_has_no_chained_keyerror_traceback` (no `lerobot`), `test_device_connect_init_failure_keeps_process_alive` (no `device_connect_edge`) |
| composition (`3589384a`) | 1 failed, 75 passed, 52 skipped | `..._unsupported_type_error_has_no_chained_keyerror_traceback` (no `lerobot`) |

The failing set is a strict subset - nothing new, and one baseline failure is gone. That
second baseline failure is the environment dependency this pull request removes: on `main`
the foreground test drives the shipped import unstubbed and fails wherever the
`[device-connect]` extra is absent. It passes on the composition because the test now
selects its branch through `sys.modules` instead of through the host. #2316's guard passes
on both.

`ruff check` and `ruff format --check` clean on `strands_robots/robot.py` and
`tests/test_robot_factory.py` at the merged head.

The merge base is now `main`'s tip, so the check clears by construction. #2319 is the only
other approved branch in flight and shares no path with this one (`README.md`,
`docs/api-reference.md`, `tests/policies/groot/...`), so landing in either order composes
without re-staling this check.


yinsong1986 (approved):
## Summary

Makes the Device Connect foreground runner (`_run_device_connect_foreground`) print a status line that matches what actually came up: when `init_device_connect_sync` fails (most commonly an absent `[device-connect]` extra), the process previously printed `<peer-id> is online` even though the built-in mesh had already been stopped and the runtime never started, leaving the process reachable over nothing while claiming the opposite. The fix gates the line on `instance._device_connect_runtime` (initialized to `None` in `_attach_device_connect`, so the check is well-founded), distinguishes the had-a-mesh / never-had-a-mesh cases, and names the installable extra only when an `ImportError` is what failed. The keep-alive behaviour, Ctrl+C handling and success-path output are unchanged (success line is byte-identical).

## What's good

- The failed-branch output is pinned ASCII-only alongside the existing regression, and both branches of the warning (extra-names-remedy vs broker-failure-does-not) have dedicated tests.
- The test refactor removes a real environment dependency: the old foreground tests drove the shipped import unstubbed, bringing up a real Zenoh session where the extra was installed. Substituting `sys.modules` makes both branches selectable by the test, per the fail-fast/no-environment-dependent-verdict conventions in AGENTS.md.
- All instance reads in the changed path use `getattr(..., None)` with defaults, so no new crash surface on the degraded path.
- No new public API, wire format, env var, or persisted schema; scope matches the PR description exactly.

yinsong1986 (approved):
## Summary

Fixes the Device Connect foreground runner so the status line matches the runtime that actually came up: `_run_device_connect_foreground` stops the built-in mesh *before* the bring-up, so a failed `init_device_connect_sync` previously left the process serving no transport at all while stdout announced `<peer-id> is online`. The runner now gates the status line on `_device_connect_runtime` (initialized to `None` by `_attach_device_connect`, so the gate is well-defined on both branches), prints `<peer-id> is NOT online` naming which transport was lost (mesh-was-stopped vs never-had-one), and the warning names the `[device-connect]` extra only when an `ImportError` is what failed - an unreachable broker gets no install remedy. The success path is byte-identical and the keep-alive loop is untouched. The test refactor is a genuine improvement beyond the fix: the old foreground tests drove the shipped import unstubbed, so which branch they exercised was decided by whether the optional extra was installed (and under `[all]` they opened a real Zenoh session inside a unit test); both branches are now selected via `sys.modules` substitution with no sockets. The changed warning format stays substring-compatible with the pre-existing assertions in `tests/test_device_connect_sync_bringup_outcomes.py`, all new output is ASCII-only and pinned as such, and the changelog fragment follows AGENTS.md step 3.

## What's good

- The Round 2 composition check against #2316 was run rather than inferred, with the failing set measured as a strict subset of the baseline.
- New tests pin both branches, the remedy/no-remedy distinction, and the mesh-stopped/never-had-one distinction; the repurposed `test_device_connect_init_failure_keeps_process_alive` still pins its original subject (the process survives).
- No silent default on error (AGENTS.md convention 6): the failure path now reports instead of claiming the opposite.

PR #2319, opened 2026-08-16
docs(groot): document STRANDS_GROOT_WIRE_LOG as the directory it is read as

cagataycali:
## Problem

`docs/api-reference.md` offers `1` as the value that turns the GR00T wire-payload diagnostic on:

```
| `STRANDS_GROOT_WIRE_LOG` | Log raw ZMQ frames when `1` | (unset) |
```

The variable's value is not a flag, it is the directory the dumps land in.
`_wire_log_dir()` returns whatever is exported verbatim and
`Gr00tPolicy._maybe_dump_wire_payload` hands it straight to `os.makedirs`, so there
is no flag branch to fall into. Following the table writes pickle archives into a
directory literally named `1` under whatever the process CWD happens to be:

documented value vs where the dumps land

Both panels are the shipped `_maybe_dump_wire_payload` running unmodified; the only
difference is the value each table tells the reader to export.

Neither table says the destination is the reader's to choose, so the real contract is
not discoverable from either one. The README row describes the payloads correctly but
also never mentions that the value is a path.

## Why the value convention drifted

The variable sits in a three-member diagnostic family that reads as one convention,
and one member is the odd one out:

| Variable | How the code reads the value |
|---|---|
| `STRANDS_LIBERO_ACTION_LOG` | flag - `os.environ.get(...).strip().lower() in ("1", "true", "yes", "on")` (`benchmarks/libero/adapter.py:4742`) |
| `STRANDS_LIBERO_STATE_LOG` | flag - same expression (`benchmarks/libero/adapter.py:710`) |
| `STRANDS_GROOT_WIRE_LOG` | **directory** - `_wire_log_dir()` returns the raw value (`policies/groot/policy.py:1363`) |

They share a naming pattern (`X` plus an `X_MAX`/`X_MAX_CALLS` companion), a default
cell shape, and a row in the same README table, so the flag convention got generalised
to the one member that takes a path.

## The second claim in the same row

"Log raw ZMQ frames" also mis-states what is captured and where from.
`_maybe_dump_wire_payload` is called from the in-process inference path
(`policy.py:1000`) as well as the service path (`policy.py:1124`), and it pickles the
observation and the action chunk rather than transport frames. In-process inference
opens no socket at all, so a transport-flavoured description hides exactly the half a
LOCAL-versus-SERVICE parity comparison needs - which is the diagnostic's stated purpose
in the README.

## Change

Correct both rows to state the contract the code implements: the value is the directory,
with a worked example, and the diagnostic covers the in-process path as well as the
service path. Behaviour is untouched; no `strands_robots` source is modified.

Add a guard that grades the documented values through the shipped resolvers rather than
a hand-copied list of spellings, so widening what a resolver accepts later cannot
silently invalidate it. A `_MINIMUM_DOCUMENTED_ROWS` premise assertion fails loudly if a
reformat ever drops the tables, so a clean sweep cannot come from reaching nothing.

Also pin the in-process half of the dumper, which had no coverage: all five existing
tests in `TestWirePayloadDiagnostic` exercise `service` only.

## Tests

`tests/policies/groot/test_wire_log_documentation_matches_the_diagnostic.py`, run
against the pre-fix docs: **2 failed, 6 passed**. After: **8 passed**.

The headline failure names the offender and the remedy:

```
docs/api-reference.md:199: STRANDS_GROOT_WIRE_LOG is documented with value '1',
which _wire_log_dir() resolves to a CWD-relative path; document the directory the
dumps should land in instead
```

| Test | Pre-fix | Role |
|---|---|---|
| `test_no_row_offers_a_value_that_is_not_a_path` | fail | the defect |
| `test_the_wire_log_row_still_shows_a_value_a_reader_can_export` | fail | pre-fix the README row shows no value at all; post-fix it guards against "fixing" the row by deleting the example |
| `test_the_raw_value_is_used_as_the_path` (x2) | pass | oracle: the resolvers take the value verbatim |
| `test_a_flag_token_resolves_to_a_relative_directory` | pass | oracle: `1` is a directory name, not an on-switch |
| `test_the_guard_reaches_the_shipped_tables` | pass | premise |
| `test_the_in_process_path_writes_its_own_archive` | pass | new coverage for the untested `local` mode |
| `test_both_paths_dump_into_one_directory_under_distinct_names` | pass | the two archives coexist, which is what makes them diffable |

Full `tests/` shows no new failures against `upstream/main`; `ruff check`, `ruff format
--check` and `mypy` are clean over `strands_robots tests tests_integ`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1163,7 +1163,7 @@ other spelling is refused. See
 |----------|-------------|---------|
 | `STRANDS_LIBERO_ACTION_LOG` / `_MAX` | Per-step OSC controller diagnostics | unset / `50` |
 | `STRANDS_LIBERO_STATE_LOG` / `_MAX` | Per-step state values fed to GR00T | unset / `50` |
-| `STRANDS_GROOT_WIRE_LOG` / `_MAX_CALLS` | Dump pre/post inference payloads to verify LOCAL vs SERVICE parity | unset / `10` |
+| `STRANDS_GROOT_WIRE_LOG` / `_MAX_CALLS` | Directory to dump pre/post inference payloads to, e.g. `/tmp/groot-wire`, to verify LOCAL vs SERVICE parity | unset / `10` |
 
 </details>
 
--- a/changelog.d/2319-groot-wire-log-doc-value.md
+++ b/changelog.d/2319-groot-wire-log-doc-value.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- Document `STRANDS_GROOT_WIRE_LOG` as the directory the GR00T wire-payload dumps land
+  in. The api-reference table offered `1` as the value that enables the diagnostic, but
+  the value is passed straight to `os.makedirs`, so following the table wrote pickle
+  archives into a directory named `1` under the process working directory. The same row
+  described the dumps as raw ZMQ frames although the diagnostic also covers the
+  in-process inference path, which opens no socket.
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -196,7 +196,7 @@ LIBERO task suites, BDDL parser. Install: `uv pip install "strands-robots[benchm
 | `STRANDS_MESH_AUDIT_DIR` | Safety event audit log | `~/.strands_robots/` |
 | `MUJOCO_GL` | GL backend for MuJoCo | auto |
 | `GROOT_API_TOKEN` | GR00T cloud inference token | (unset) |
-| `STRANDS_GROOT_WIRE_LOG` | Log raw ZMQ frames when `1` | (unset) |
+| `STRANDS_GROOT_WIRE_LOG` | Directory to dump pre/post-inference payloads to, e.g. `/tmp/groot-wire`; covers the in-process path as well as the service path, capped by `STRANDS_GROOT_WIRE_LOG_MAX_CALLS` | (unset) |
 
 ## See also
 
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Docs-only correction plus a guard test. The `docs/api-reference.md` row told readers to export `STRANDS_GROOT_WIRE_LOG=1` as an on-switch, but `_wire_log_dir()` (policy.py:1196) returns the raw value and `_maybe_dump_wire_payload` passes it straight to `os.makedirs` (policy.py:1150-1160), so following the table wrote pickle archives into a CWD-relative directory named `1`. The same row described the dumps as raw ZMQ frames, but the dumper pickles the observation and action chunk and is called from the in-process path (policy.py:842) as well as the service path (policy.py:966). Both corrected rows now match the shipped contract. The new test grades the documented values through the shipped resolvers rather than a hand-copied list, includes a minimum-row premise so a reformat cannot produce a vacuous pass, and adds the previously-uncovered `local`-mode dump. No `strands_robots/` source is modified.

## What's good

- Every claim in the description checks out against the source: the resolver, the `os.makedirs` sink, both call sites, and the pickle payload shape.
- The doc sweep produces no false positives on the unchanged `STRANDS_MESH_AUDIT_DIR` rows (README.md:1106, docs/api-reference.md:196, docs/getting-started/installation.md:107) - verified by running the suite at the head: 8 passed.
- `_PATH_VALUED` deliberately admits only side-effect-free resolvers, and `audit_log_path` (audit.py:931) is documented as not creating the directory, so the premise holds.
- Test hygiene follows the review learnings: `monkeypatch.setenv` throughout, writes only under `tmp_path`, no host paths, and the `__new__`-based fixture matches the tolerance `_maybe_dump_wire_payload` explicitly documents for exactly that construction path.
- Changelog fragment present at this head per AGENTS.md step 3.

PR #2320, opened 2026-08-16
fix(simulation): publish the ellipsoid shape the agent schema already documents

cagataycali:
## Problem

`add_object` compiles an ellipsoid. `_geom_type` maps `"ellipsoid"` to
`mjGEOM_ELLIPSOID`, `_SIZE_LAYOUT` declares its layout (`[x, y, z]` full
diameters), the method docstring lists it among the valid shapes, and
`docs/simulation/world-building.md` gives it a row in the size table. The agent
tool schema's `shape` enum offered six values and not that one.

`shape` is one of the properties that carries no `description`, so for a
schema-constrained decoder the enum *is* the whole specification of the
parameter. And the router binds against the method signature rather than the
schema, so an unpublished value is not refused - it is simply never emitted. The
capability therefore read as unsupported rather than as undiscoverable, which is
why it survived.

The schema also contradicted itself. Its `size` description publishes the
per-shape layout table and names `ellipsoid` twice ("box/ellipsoid [x,y,z]",
"ellipsoid [three semi-axes]"), so a model that follows the description emits a
value the same schema rejects. That is the value-level form of the action-enum
drift this suite already pins in both directions:

> the tool_spec description lists these actions but the enum does not offer
> them, so a model following the description emits a value the schema rejects

Measured on the MuJoCo backend, before the change:

| Call | Result |
|---|---|
| `sim.add_object(shape="ellipsoid", size=[0.2, 0.1, 0.3])` (Python) | `success`, compiles `mjGEOM_ELLIPSOID`, `geom_size [0.1, 0.05, 0.15]` |
| `shape` enum offered to a model | `box, sphere, cylinder, capsule, mesh, plane` |
| `size` description's shape vocabulary | `box, sphere, cylinder, capsule, ellipsoid, plane, mesh` |

## Change

`"ellipsoid"` joins the `shape` enum, in the builder's own order. Nothing else
moves: the value was already validated, already had a size layout and already
compiled, so this publishes a capability rather than adding one.

The guards are keyed on the live shape tables (`_geom_type`'s map and
`_SIZE_LAYOUT`) rather than on a copy of them, so a shape added to the builder
fails CI until the schema offers it - the same shape as the existing vector-arity
contract, which is keyed on `_VECTOR_PARAM_LENGTHS`.

## Tests

`TestTheSchemaPublishesEveryShapeAddObjectCompiles` in
`tests/simulation/mujoco/test_tool_spec.py`, 3 failing pre-fix and 2 passing on
both trees:

| Test | Pre-fix |
|---|---|
| every shape the builder compiles is offered in the enum | **fails**: `['ellipsoid']` |
| the `size` description names no shape the enum omits | **fails**: `['ellipsoid']` |
| a published shape compiles the geom type it names (end to end through the router) | **fails** on its premise |
| the enum offers no shape the builder refuses | passes (control) |
| every published shape has a `_SIZE_LAYOUT` row to quote | passes (control) |

The two constant passes are the no-overreach controls. The first pins the
converse direction - an enum entry the builder cannot compile is the worse
failure of the two, because the model is told to emit a value that is refused on
arrival - so it shows the fix published a value the builder honours rather than a
name. The second pins that publishing a shape also means declaring its size
layout: `_validate_size` reads `_SIZE_LAYOUT.get(shape, (0, ""))`, so a shape
with no row would accept any size vector and compile from whatever it held.

Full `tests/` shows no new failures against `upstream/main` on the same
interpreter, and `ruff check` / `ruff format --check` / `mypy` are clean.

## Rollout

Both panels come from one scene script that places one object per shape its
*own* tree publishes, rendered headless in MuJoCo (`MUJOCO_GL=egl`). Slots are
fixed per shape name, so a shape both trees publish lands in the same place and
the only difference is the slot one of them cannot fill. Magenta ellipsoid
pixels: 0 before, 3392 after.

shape enum before and after

### DIFF
--- a/changelog.d/2320-publish-ellipsoid-shape.md
+++ b/changelog.d/2320-publish-ellipsoid-shape.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Simulation**: the MuJoCo agent tool schema now offers `ellipsoid` in its `shape` enum. `add_object` already compiled one and the schema's own `size` description already published its layout, but the `shape` enum omitted it - and that property carries no description, so the enum is the whole specification a schema-constrained caller reads. The enum is now pinned against the builder's live shape tables in both directions.
--- a/docs/simulation/world-building.md
+++ b/docs/simulation/world-building.md
@@ -193,6 +193,13 @@ for i in range(5):
     )
 ```
 
+## Object shapes
+
+`shape` takes one of seven values: `box`, `sphere`, `cylinder`, `capsule`,
+`ellipsoid`, `plane` and `mesh`. All seven are offered by the agent-tool schema
+too, so a model driving the simulation can select any of them. How many `size`
+components each one consumes is in the table below.
+
 ## Object size
 
 `size` is the **full extent in meters** along each local axis - not MuJoCo's
--- a/strands_robots/simulation/mujoco/tool_spec.json
+++ b/strands_robots/simulation/mujoco/tool_spec.json
@@ -136,6 +136,7 @@
         "sphere",
         "cylinder",
         "capsule",
+        "ellipsoid",
         "mesh",
         "plane"
       ]
--- a/tests/simulation/mujoco/test_tool_spec.py
+++ b/tests/simulation/mujoco/test_tool_spec.py
@@ -938,3 +938,143 @@ class TestTheSharedSizeFieldPublishesBothConventions:
             "color means one thing to both actions, so it needs no per-action split; a mention "
             f"here would say the conventions diverge when they do not. Got: {description!r}"
         )
+
+
+# Shape-vocabulary contract
+#
+# ``shape`` publishes an enum and nothing else - it is one of the properties with
+# no description at all, so for a schema-constrained decoder the enum *is* the
+# whole specification of the parameter. ``ellipsoid`` compiles: ``_geom_type``
+# maps it to ``mjGEOM_ELLIPSOID``, ``_SIZE_LAYOUT`` gives it a layout,
+# ``add_object``'s docstring lists it among the valid shapes, and the ``size``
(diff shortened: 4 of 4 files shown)


yinsong1986 (approved):
## Summary

One-line schema fix with a bidirectional contract around it: `"ellipsoid"` joins the agent tool schema's `shape` enum, which was the only surface not offering a value the builder already validated and compiled (`_geom_type` maps it to `mjGEOM_ELLIPSOID`, `_SIZE_LAYOUT` declares `[x, y, z]` full diameters, and the schema's own `size` description named it twice). Since `shape` carries no `description`, the enum was the entire specification a schema-constrained decoder reads, so the capability was undiscoverable rather than refused. Verified locally on the head commit: the five new tests in `TestTheSchemaPublishesEveryShapeAddObjectCompiles` pass, and reverting only `tool_spec.json` to the base reproduces exactly the claimed 3-failed/2-passed split, so the regression pins fail on pre-fix code as required.

## What's good

- The guards are keyed on the live tables (`_GEOM_TYPE_CACHE`, `_SIZE_LAYOUT`) rather than a copied literal, so a future shape added to the builder fails CI until the schema offers it - same pattern as the existing `_VECTOR_PARAM_LENGTHS` arity contract in this file.
- Both no-overreach controls are in place: the enum offers nothing the builder refuses, and every published shape has a `_SIZE_LAYOUT` row, so the fix cannot silently widen into a value that is refused on arrival or validated against nothing.
- The end-to-end test goes through the router and asserts the compiled `geom_type` and halved `geom_size`, not just schema text.
- Changelog fragment per AGENTS.md step 3, docs updated in the same PR, no non-ASCII or host paths introduced, `ruff check`/`ruff format --check` clean on the changed files.

Issue #2323, opened 2026-08-16
feat(rendering): light the robot with the scene it stands in - GS-derived IBL, shadow catcher, color-pipeline alignment for isaac_gs/mujoco_gs composites


yinsong1986:
# Gap

Once the backdrop actually renders splats (sibling issue on the silent fallback), the remaining photorealism ceiling is that **the robot pass and the background are lit and toned in two unrelated worlds**, then byte-blended:

1. **Robot lighting is two hardcoded example lights** - a warm `DistantLight` (intensity 1500, color (1.0, 0.96, 0.9), rot (-50, 10, 0)) and an untextured `DomeLight` (450) (`examples/isaac_gs/scene.py:100-107`). Nothing about them derives from the GS scene, so the arm reads as pasted-on under mismatched illumination. README.md:212-213 documents the scope cut ("the sim robot is lit by Isaac's RTX scene lights").
2. **No shadows onto the backdrop.** The world is created with `ground_plane=False` (`scene.py:247`) and there is no shadow-catcher, so the arm casts no contact shadow onto the counter - one of the strongest pasted-on cues.
3. **No color-space alignment anywhere in the layer** (verified by grep: no sRGB/linear conversion in `strands_robots/rendering/`). Isaac's tonemapped RTX output is blended byte-for-byte with GS DC colors that live in the training images' color space.
4. The foreground renderer configuration is inert (see the backend-hygiene sibling): `render_mode` never reaches `SimulationApp` (`isaac/simulation.py:989` passes only `headless`), so the robot pass always runs default RTX real-time with default tonemapping/autoexposure - there is currently no knob to trade robot-pass quality up.

# Proposal (staged; each stage independently shippable)

1. **IBL from the GS scene**: bake an equirect environment map by rasterizing the splats (six cube faces or direct equirect sampling from the robot's position), write it to a texture, and set it on the `DomeLight`. The robot is then lit by the room it stands in - the single biggest realism lever here.
2. **Shadow catcher**: an invisible-but-shadow-receiving plane at the alignment's `floor_z` (the curated tabletop alignment already knows the counter height, `backgrounds.py:609-619`), so the arm grounds itself with a contact shadow composited over the backdrop.
3. **Color pipeline audit**: document what space each layer is in and align them (one gamma/exposure decision, applied once), so foreground/background brightness and saturation match at the seam.

# Acceptance

- A/B images per stage in the PR description (same pose, before/after)
[...]

Issue #2322, opened 2026-08-16
bug(rendering): GsplatBackground.render - premultiplied alpha composited twice, accumulated depth never alpha-normalized (docstring says it is), AA-trained SPZ rendered classic


yinsong1986:
# Three defects in `GsplatBackground.render` (`strands_robots/rendering/backgrounds.py:488-560`), each measured

## 1. Premultiplied alpha is composited twice (`:554-555`)

gsplat's `render_colors` is already alpha-accumulated (premultiplied). The code then does

```python
rgb = rgb * alpha + self._bg_fill * (1.0 - alpha)
```

which weights splat color by alpha^2. Standard usage is `rgb + (1 - alpha) * fill`.

**Measured** (gsplat 1.4.0, synthetic single gaussian, opacity 0.5, dead-center): `alpha=0.499`, premultiplied red `0.499`; repo formula yields `0.249` vs correct `0.499` - **2.0x darkening** at half coverage. Effect: dimming/grey halos wherever coverage is partial (splat edges, sparse regions, zenith).

## 2. Accumulated depth is never alpha-normalized, and the docstring claims it is (`:546`, `:550`, docstring `:336-338`)

`render_mode="RGB+D"` returns **accumulated** (alpha-weighted) depth. The class docstring says it is "divided through to give metric depth"; the code never divides.

**Measured**: the same synthetic gaussian at true depth 4.000 m reports raw accumulated depth **1.995 m**; divided by alpha it is 4.000 m. Because the compositor's z-test (`strands_robots/rendering/compositor.py:314`) trusts this value, soft GS regions bias toward the camera and can wrongly occlude robot pixels.

Fix: divide by `clamp(alpha, eps)` (or switch to `render_mode="RGB+ED"`, gsplat's expected-depth mode) and align the docstring.

## 3. Antialiased-trained SPZ rendered in classic mode (`_load_spz_splats:956`, rasterization call `:534-547`)

The default `tabletop.spz` header carries `flags=1` - bit 0 is the antialiased/mip-splatting training flag (header-verified: SPZ v2, n=213,944, sh_degree=0, flags=1). The loader parses and discards the flags; rasterization runs the default `rasterize_mode="classic"`, so opacities trained WITH the AA compensation are rendered WITHOUT it.

Fix: honor flag bit 0 by passing `rasterize_mode="antialiased"` for such assets.

# Honest scoping

[...]

Issue #2321, opened 2026-08-16
bug(examples/isaac_gs): the photoreal path is silently optional - any GS failure demotes to the procedural gradient, and the shipped gallery is the fallback


yinsong1986:
# Symptom

The example does not look photorealistic - and the primary reason is that in most real runs (and in the repo's own gallery) **no Gaussian splats are ever rendered**. The output people see is the procedural gradient fallback, not GS.

# Root cause chain

1. **Any background failure silently demotes to `PanoramaBackground`** - a flat two-band gradient + noise (`_make_procedural_kitchen_panorama`, `strands_robots/rendering/backgrounds.py:220-261`). The demotion is a catch-all: `examples/isaac_gs/background.py:86-94` catches `Exception` (`noqa: BLE001`) and returns the panorama with only a `logger.warning`. A user who asked for the photoreal path gets the gradient and a log line.
2. **The most common failure is environmental and known**: a plain `pip install gsplat` in the Isaac container imports fine but cannot CUDA-rasterize (README.md:108-112 documents this), so the demotion fires on exactly the default install path.
3. **The shipped gallery is the fallback.** `assets/hero_front.jpg` and `assets/wave.gif` are beige-gradient renders; the README's own caption admits it (README.md:24-27: "These were rendered on the procedural `PanoramaBackground` fallback (gsplat not installed)"). The example's face is its own degraded mode.
4. **A `GsplatBackground` accepts a nonexistent path at construction** and raises `FileNotFoundError` only at first `render()` (`backgrounds.py:435` via lazy `_load`). Late failure is what feeds the catch-all in app contexts. Reproduced: `GsplatBackground(ply_path="not-a-path")` constructs silently.

# Evidence

- On a GPU box where gsplat rasterizes correctly, the identical demo camera pose (`front`, 640x480, hfov 48) renders the default `tabletop.spz` as a convincing captured kitchen (shelves, crockery, tiled wall, counter). Side-by-side with `assets/hero_front.jpg` (flat beige bands) the gap IS the reported "not photorealistic".
- Repro for the good path: `resolve_background(gsplat_scene="tabletop (indoor room)")` -> `GsplatBackground.render(cam)` succeeds and looks right; the asset is a 3.07 MB SPZ v2, 213,944 gaussians (header-verified).

This violates the repo's own rules 5/6 (raise on fatal, no silent defaults): the system behaves differently from what the user requested, and continues.

# Fix

[...]

Issue #1537, opened 2026-07-22
refactor(rendering): lift the GS hybrid-render layer out of examples (public get_frame/get_camera_params, strands_robots/rendering with HybridCompositor + backgrounds, media utils, synthdata decision)


yinsong1986:
## Summary

Same exercise as #1533 (so101_curobo), applied to the Gaussian-splat hybrid-render stack: `examples/isaac_gs/` (2,107 lines) + `examples/isaac/isaac_replicator_synthdata.py` (473 lines), audited line-by-line on `main` @ `f67d9c8`. Headline findings:

- The **generic core the example exists to demonstrate - depth-aware GS compositing, camera intrinsics/extrinsics math, background renderers - has no library home**, so it lives in example modules that other examples import across sibling directories. That cross-example dependency is currently *broken on main* (see the companion bug issue: `examples.mujoco_gs` is unmerged #1279, and the render-envelope contract drifted).
- ~30% of the code is generic library-worthy, ~10% is tuned config that should be data, ~60% is legitimately demo (Gradio/CLI/agent). The refactor case here is less "shrink the example" (the so101_curobo story) and more **"the shared rendering layer must live in `strands_robots/`, not in a sibling example"** - plus deleting ~200 lines that duplicate library APIs that already exist.
- The example layer also re-implements library machinery: `app.py`'s `_RenderRequest`/`serve_forever`/`_run_on_main` (app.py:148-165, 282-302, 413-429) duplicate `IsaacSimulation.pump`/`run_pump_forever`/`run_on_main` (`isaac/simulation.py:3545-3688` - which were themselves consolidated from an example under #69); `_wave_arm` pokes private `sim._robots[...].articulation` (app.py:340-353) where public `set_joint_positions` exists; `scene._add_lighting` (scene.py:79-108) duplicates the library's private `_add_lighting` (`isaac/simulation.py:4123-4152`).

## What is genuinely missing from the library (verified)

1. **Raw-frame render API.** No public way to get `(rgb, depth)` ndarrays: Isaac hides it in `_render_frame` (`isaac/simulation.py:2365-2500`), MuJoCo's `render_depth` returns only a normalized PNG + min/max scalars (`mujoco/rendering.py:936-1123`), Newton renders RGB only (`newton/simulation.py:1179-1223`).
[...]

PR #2326, opened 2026-08-16
feat(rendering): evaluate SH view-dependence and swap GS presets to fully-trained captures

### LINKED ISSUE #2325: feat(rendering): evaluate SH view-dependence instead of dropping it, and replace iteration-7000 GS presets with fully-trained captures
# Gap

Two fidelity limits in the GS preset/loading path, distinct from the correctness defects in #2322:

1. **Higher-order SH coefficients are dropped.** `_load_ply_splats` keeps only the DC term (`strands_robots/rendering/backgrounds.py:1010-1012`); `_load_spz_splats` likewise ignores trailing SH (`:977`). Rasterization passes per-gaussian RGB with no `sh_degree`, so view-dependent appearance (speculars, sheen) is lost. The default `tabletop.spz` is genuinely `sh_degree=0` (header-verified), so the default scene loses nothing - but every `.ply` preset carries SH that is discarded, and so would any user-supplied full-quality capture (the digital-twin use case).
2. **The `.ply` presets are early-training checkpoints.** `bonsai`/`bicycle`/`stump` download `iteration_7000` checkpoints (`:576-584`) - visibly undertrained vs the standard 30k iterations; `bicycle` is already excluded from skybox use as "white haze + floaters from every angle" (`:600-601`), which is what an undertrained scene looks like.

# Fix

1. Load and pass full SH when present: gsplat's `rasterization` accepts `colors` of shape `(N, K, 3)` with `sh_degree=N` - evaluate at render time instead of baking DC. Keep DC-only as the documented fast path for assets without SH.
2. Swap presets to fully-trained captures (30k-iteration checkpoints of the same scenes, or better-curated equivalents), and record per-preset provenance (source, iteration, license) next to the URL table.

# Acceptance

- A/B render of a preset with SH on/off in the PR description
- Loader unit tests cover both SH-present and DC-only assets
- `bicycle` re-evaluated after the swap; if still hazy, keep the exclusion with the measurement that justifies it

# Related

- #1537 (rendering-layer epic); siblings #2321, #2322, #2323

yinsong1986:
Closes #2325

## What changed

Two fidelity limits in the GS preset/loading path (distinct from the correctness defects in #2322):

1. **SH view-dependence is now evaluated instead of dropped.**
   - `_load_ply_splats` decodes `f_rest_*` fields (INRIA channel-major layout: `f_rest_{c*(K-1)+j}` is channel *c*, coefficient *j*) into a raw `(N, K, 3)` coefficient stack, DC term first.
   - `_load_spz_splats` decodes the trailing SH block (coefficient-major, channel fastest-varying, dequantized as `(byte - 128) / 128` per the spz spec) the same way.
   - `GsplatBackground.render` derives `sh_degree` from the coefficient count and passes it to `gsplat.rasterization`, which evaluates the coefficients against per-view directions.
   - **DC-only assets keep the baked-RGB fast path unchanged** — including the default `tabletop.spz`, which is genuinely `sh_degree=0` (verified end-to-end: still loads `(N, 3)` colors and renders identically).
   - Malformed SH is refused by name rather than mis-evaluated: an `f_rest_*` count matching no integer degree, a count not divisible by 3 channels, and a `.spz` header claiming a degree its body has no bytes for all raise `ValueError` naming the file.
2. **Presets swap to fully-trained captures, with provenance.** `bonsai`/`bicycle`/`stump` now download the `iteration_30000` checkpoints of the same official INRIA 3DGS models (same `dylanebert/3dgs` mirror) instead of the visibly-undertrained `iteration_7000` ones. `GSPLAT_SCENE_PROVENANCE` records source, training iteration and license per preset, next to the URL table, with keys pinned to `GSPLAT_SCENES` by test.

## A/B render: SH on vs off (acceptance criterion 1)

`bonsai` 30k checkpoint (375,472 gaussians, `sh_degree=3`), rendered through the real CUDA rasterizer on an NVIDIA L4 at 960x540, four-view orbit at the arm's eye level (r=1.2 m, z=0.5 m, skybox alignment `up_sign=-1`). "SH off" is the pre-fix behavior reproduced exactly (DC baked via `clip(0.5 + SH_C0 * f_dc)`, rest dropped):

| view | mean abs RGB delta | PSNR (DC vs SH) | pixels differing > 2 LSB |
|---|---|---|---|
| 0 | 4.61 / 255 | 30.2 dB | 56.2% |
| 1 | 3.95 / 255 | 29.3 dB | 46.4% |
| 2 | 3.19 / 255 | 35.7 dB | 63.6% |
| 3 | 3.31 / 255 | 33.2 dB | 60.8% |

So roughly half the frame changes at every angle once view-dependence is evaluated — visible as speculars/sheen on the pot, table and floor. (This sandbox cannot upload image attachments; the A/B is reported quantitatively. The renders are one `GsplatBackground(ply_path=<bonsai 30k>, skybox=True, up_sign=-1.0).render(cam)` per view, with the DC arm produced by replacing the loaded `(N, 16, 3)` coefficients with the baked `(N, 3)` DC colors.)

## bicycle re-evaluation (acceptance criterion 3)

Re-measured on the fully-trained 30k checkpoint through the same uncurated-skybox pipeline an upload gets (auto up-sign), four-view orbit, with a green `bg_fill` so void pixels are separable from splat pixels:

| scene | brightness (mean/255) | saturation (mean max-min) | near-white pixels (min channel > 200) |
|---|---|---|---|
| bicycle 7k | 153–204 | 3–7 | 27.5–58.5% |
| bicycle 30k | 138–204 | 24–51 | 32.9–65.9% |
| bonsai 30k (control) | 64–126 | 57–121 | 0.0–4.9% |

The 30k checkpoint is sharper (saturation up ~7x) but the near-white fraction did not improve — 33–66% of pixels per view vs bonsai's 0–5% — because the haze is the capture's overcast sky, not undertraining. **The exclusion stands**, and the comment in `GSPLAT_SKYBOX_ALIGN` now carries this measurement.

## Test surface (acceptance criterion 2)

- PLY loader: SH-present asset loads raw `(N, K, 3)` coefficients with the exact channel-major field mapping pinned; DC-only asset stays on the `(N, 3)` fast path; two malformed-count refusals.
- SPZ loader: `sh_degree=1` asset decodes DC + dequantized rest in stored order; `sh_degree=0` stays baked; truncated SH block refused.
- `_clip_splats` filters `(N, K, 3)` colors on the gaussian axis, keeping each survivor's full coefficient stack.
- `render()` color contract at the rasterization boundary (CUDA doubled): `(N, 3)` passes `sh_degree=None`; `(N, K, 3)` passes the degree its count encodes, parametrized over degrees 1–3.
- Presets: every `GSPLAT_SCENES` URL is free of `iteration_7000`, `.ply` presets point at `iteration_30000`, and `GSPLAT_SCENE_PROVENANCE` covers every preset with non-empty source/iteration/license.

Full suite: **30466 passed, 277 skipped**, coverage 95.25% (run with `MUJOCO_GL=egl` — the classic GLX renderer aborts on this headless box, which is environmental, not from this diff). `hatch run lint` clean. Changelog fragment included per `changelog.d/` convention.

## Out of scope

- The premultiplied-alpha / depth-normalization / AA-mode correctness defects in the render path are #2322.
- Siblings #2321, #2323 under the rendering-layer epic #1537.

Board note: #2325 should move to "In review" on the [Strands Labs - Robots board](https://github.com/orgs/strands-labs/projects/2).


cagataycali (commented):
## Batch composition: measured pairwise, not inferred

Not a review of the SH diff yet - this is only the composition across the four open rendering branches, which no branch's CI can see. All four (#2326, #2331, #2332, #2336) share one merge base (`9266331`) and each is `CLEAN` against `main` at `5ad45ef`. Two of the six pairs do not merge, and one that does merge cleanly still changes behaviour.

`git merge-tree --write-tree --name-only` on the four heads:

| pair | result |
|---|---|
| #2332 + #2326 | **CONFLICT** - `strands_robots/rendering/backgrounds.py`, `tests/rendering/test_backgrounds.py` |
| #2336 + #2331 | **CONFLICT** - `examples/isaac_gs/background.py` |
| #2332 + #2336, #2326 + #2336, #2332 + #2331, #2326 + #2331 | clean |

## The conflict is one line, and picking a side silently deletes a PR

Both branches add a kwarg to the *same* `rasterization(...)` call, and nothing else in that call collides:

```
<<<<<<< #2332
            rasterize_mode=self._rasterize_mode,
=======
            sh_degree=sh_degree,
>>>>>>> #2326
```

These are orthogonal parameters - one selects the AA rasterizer for a mip-splatting-trained asset, the other asks the rasterizer to evaluate view-dependent SH. A side-taking resolution is a green tree that has quietly dropped either the antialiased flag (#2322's third finding) or SH evaluation (this branch's whole point). **The resolution is the union**, at all three points the merge flags:

1. the `rasterization(...)` call - both kwargs;
2. the `_load_spz_splats` docstring paragraph - both the `"antialiased"` header-flag key and the `(N, K, 3)` raw-SH `colors` shape;
3. the fake-asset builder signature in `tests/rendering/test_backgrounds.py` - both `flags: int = 0` and `sh: np.ndarray | None = None`.

Worth stating what the auto-merge gets *right*, because it is the part a hand-resolution would most easily undo: outside the conflict the merge keeps #2332's corrected composite (`rgb + fill * (1 - alpha)`) and its alpha-normalized depth, and drops this branch's `rgb * alpha + fill * (1 - alpha)` and raw `out[..., 3]`. That is the correct outcome - this branch predates the fix and still carries the double-composite on those exact lines. So the merge wants no hand-editing beyond the three union points, and specifically must not restore the composite this branch was written against.

## #2332 + #2331 merges clean and still changes shipped pixels

#2331 replaces the inline cube-face bake in `bake_gsplat_panorama` with `ibl.render_environment_map`, which consumes `background.render(cam)`. That is textually disjoint from #2332's edits *inside* `render`, hence the clean merge - but the RGB it consumes is exactly what #2332 corrects, so the baked environment map and the key light derived from it get brighter once both are in.

No test breaks, checked rather than assumed: `tests/rendering/test_ibl.py` drives a `DirectionPaintingBackground` stub rather than the real `GsplatBackground`, and `render_environment_map` discards depth (`rgb, _ = background.render(cam)`), so #2332's depth change is invisible to it. This is a shipped-pixels interaction only, and #2337 already owns regenerating the gallery once #2322/#2323 land.

## Order with the least rework

`#2336` -> `#2332` -> `#2326` (union resolution above) -> `#2331` (re-resolve `examples/isaac_gs/background.py`).

#2332 ahead of this branch because a correctness fix underneath a feature is cheaper than the reverse: landing SH first means the composite and depth fix has to be re-derived against the new colour path instead of merging into it. #2336 ahead of #2331 for the same reason on the smaller half of that pair.

Reproduction:

```
git fetch origin main refs/pull/2326/head:pr2326 refs/pull/2332/head:pr2332 \
                      refs/pull/2331/head:pr2331 refs/pull/2336/head:pr2336
git merge-tree --write-tree --name-only pr2332 pr2326    # conflict
git merge-tree --write-tree --name-only pr2336 pr2331    # conflict
```

The overlap gate is correct to be green on all four: it keys on files a branch shares with what has *landed*, and none of these has landed yet. The pairwise cost only exists between open branches, which is the gap #2401 describes from the other direction.

yinsong1986:
@cagataycali Point-by-point on the composition review. No commit lands on this branch for it — every finding is either a merge-time action this branch cannot take until #2332 is on `main`, or a claim I re-verified rather than assumed. Numbering the items:

**1. The pairwise measurement.** Reproduced both conflicts locally (this box's git 2.34 lacks `merge-tree --write-tree`, so via `git merge --no-commit` in scratch worktrees, same verdicts): `pr2332 x pr2326` conflicts in `strands_robots/rendering/backgrounds.py` + `tests/rendering/test_backgrounds.py`; `pr2336 x pr2331` conflicts in `examples/isaac_gs/background.py`. Matches your table exactly.

**2. Union resolution at the three points — validated, not just agreed.** I applied the union in the scratch composition (both kwargs in the `rasterization(...)` call; the `_load_spz_splats` docstring carrying both the `"antialiased"` header-flag sentence and the `(N, K, 3)` raw-SH sentence; the fake-asset builder keeping both `flags: int = 0` and `sh: np.ndarray | None = None` — the builder body already uses both, only the signature conflicts). `tests/rendering/` on the composition: **296 passed, 1 skipped**, and it is exactly additive against controls run with the same command: `main@5ad45ef` 277, this branch 290 (+13), pr2332 283 (+6), composition 296 (+13+6).

**3. "A side-taking resolution silently deletes a PR" — it is not silent in the merged tree, which is worth recording for whoever resolves it.** In the composition I deleted each kwarg in turn: dropping `sh_degree` fails 4 pinned tests (`TestGsplatRenderShDegree::test_sh_coefficients_pass_the_degree_their_count_encodes[1..3]` + the DC-path pin); dropping `rasterize_mode` fails 2 (`TestGsplatRasterizeModeSelection::*`). Both branches pinned their kwarg at the rasterization boundary, so the union is test-enforced post-merge — the resolution still has to be the union, but a wrong one cannot reach green.

**4. What the auto-merge gets right — confirmed, and agreed this branch must not restore its composite.** The composition keeps #2332's `rgb + fill * (1 - alpha)` and the alpha-normalized depth (`accum_depth / max(alpha, 1e-6)`), and drops this branch's pre-fix `rgb * alpha + ...` / raw `out[..., 3]`. When this branch absorbs `main` after #2332 lands I will hand-edit only the three union points and take the incoming side everywhere else in `render`.

**5. #2332 + #2331 shipped-pixels interaction.** Acknowledged — outside this PR's diff; noting your check that `test_ibl.py` drives a stub so nothing red flags it, and that #2337 owns the gallery regeneration once #2322/#2323 land.

**6. Merge order `#2336 -> #2332 -> #2326 -> #2331`.** Agreed, for the reason given: SH merging into the corrected color path is one union resolution, the reverse forces #2332 to re-derive its fix. This branch will wait behind #2332 and absorb `main` (union resolution per item 2) before merge.

**7. Overlap-gate scope.** Agreed it is correctly green — it keys on landed files, and the open-set pairwise gap is #2401's territory.

Branch itself is unchanged and still green at `dd83c59d`: `hatch run lint` clean, full suite 31356 passed / 277 skipped, coverage 95.01% (`MUJOCO_GL=egl`, same environmental caveat as the PR body).

yinsong1986:
Investigated the failing `call-test-lint / Test and Lint` run on `80e13a3c`: the single failure was `tests/test_hardware_step_cap_guard.py::TestUsableStepCapAccepted::test_none_requests_no_cap_and_duration_bounds_the_rollout` (`assert 1 > 1` on applied-action count) — a wall-clock-bounded rollout (0.2s at 200Hz) where only one loop iteration fit inside the budget on the loaded runner. This PR touches only `strands_robots/rendering/` and its tests, disjoint from the failing test's code paths, and the test passes 20/20 consecutive runs locally at the exact failing head, so this is a runner-side timing flake rather than a defect in the diff. Re-running the job was not permitted for this token, so I pushed a conflict-free merge of `main` (combined diff is empty — no conflict resolution content) to retrigger CI on a fresh head. Full validation on the merged tree: lint clean, 31517 passed / 277 skipped / 0 failed under `MUJOCO_GL=osmesa`, including the previously-failing test.

yinsong1986:
Merged `main` in and resolved conflicts in `strands_robots/rendering/backgrounds.py` and `tests/rendering/test_backgrounds.py` (merge commit `18845bbc`). All three conflicts were resolved keep-both: the rasterization call now passes both this PR's `sh_degree` and main's `rasterize_mode` (SPZ AA flag); the `_load_spz_splats` docstring documents both the `antialiased` bool and the SH-coefficient color layouts; the `_build_spz` test helper keeps both the `sh` and `flags` kwargs. Lint clean and the full unit suite is green (31887 passed, 280 skipped); ready for re-review.

cagataycali (approved):
## Review of the SH diff (my earlier comment was composition only)

My previous comment on this PR was explicitly "not a review of the SH diff yet" -- it only measured the four-way batch composition. This is the review of the diff. Short version: **the layout work is correct, and I verified it end-to-end through the real CUDA rasterizer rather than at the argument boundary.** Approving.

The reason this needed hardware: `plyfile` and `gsplat` are both absent from the `[all]` extra, and CI installs `.[all,dev]`. So the four `.ply` SH tests and every real rasterization skip in CI permanently -- the green check on this PR says nothing about the new render path. Verified on a CUDA box (L40S, gsplat 1.4.0, torch 2.10.0+cu128, `gsplat_rasterizer_available() -> (True, 'ok')`).

### The two layout claims, checked against the authoritative sources

The diff asserts two *opposite* conventions, which is where I expected a bug. Both are right, and both comments state them correctly:

- **`.ply` channel-major.** INRIA `gaussian_model.save_ply` writes `f_rest = self._features_rest.detach().transpose(1, 2).flatten(start_dim=1)` where `_features_rest` is `(N, K-1, 3)`, so `f_rest_{c*(K-1)+j}` is channel *c*, coefficient *j*. `rest.reshape(-1, 3, n_rest).transpose(0, 2, 1)` reproduces INRIA's own `load_ply` reshape exactly. Indexing `v[f"f_rest_{i}"]` over `range(n_rest_fields)` is also more robust than relying on property order (upstream has to `sorted(..., key=int)` for this reason).
- **`.spz` coefficient-major, channel fastest-varying.** `packGaussians` does `bufSh[k] = g.sh[i + k * 3 + channel]`, i.e. stride 3 per coefficient with the channel innermost, so `(N, n_rest, 3)` is correct. `packed.sh.size() == numPoints * shDim * 3` matches `n_rest = (sh_degree + 1)**2 - 1`.
- **Raw (unbaked) coefficients are the right thing to hand gsplat.** `rasterization` does `colors = torch.clamp_min(colors + 0.5, 0.0)` after `spherical_harmonics(...)`, commented "make it apple-to-apple with Inria's CUDA Backend" -- so the DC-only baked path and the SH path agree at the DC term by construction. gsplat also normalizes the view directions itself (`F.normalize(dirs, ...)`), so nothing is needed caller-side.

### End-to-end render correctness (this is the part no test covers)

Every render-path test here doubles the rasterizer, so the suite pins that `sh_degree` is *passed*, never that the resulting image is right. I closed that with a degree-1 asset whose rest block is `(3, 3)` -- **square, so a channel/coefficient transpose would be completely silent** -- and an asymmetric coefficient matrix, graded against an independent numpy evaluation of gsplat's own basis order (`[C0, -C1*y, +C1*z, -C1*x]`, transcribed from `_eval_sh_bases_fast`):

| view | this branch (centre px) | independent analytic | err | main | err |
|---|---|---|---|---|---|
| +X | 226, 127, 27 | 227, 128, 28 | 1.17 | 127, 127, 127 | 100.2 |
| -X | 27, 127, 226 | 28, 128, 227 | 1.17 | 127, 127, 127 | 100.2 |
| +Y | 139, 189, 151 | 140, 190, 152 | 1.42 | 127, 127, 127 | 62.8 |
| -Y | 114, 64, 102 | 115, 65, 103 | 1.20 | 127, 127, 127 | 61.8 |
| +Z | 89, 114, 77 | 90, 115, 78 | 1.12 | 127, 127, 127 | 49.3 |
| -Z | 164, 139, 176 | 165, 140, 177 | 1.34 | 127, 127, 127 | 50.3 |

Worst error over six views is **1.42/255**, i.e. uint8 quantization only. The loader's decode is exact (`max abs err 0.0e+00`) and explicitly *not* the transposed layout. main is flat 127,127,127 from every angle -- view-independent, as expected when `f_rest` is dropped. The splat goes red from +X and blue from -X, which is the whole point of the PR made visible.

### Cost on a real asset

`bonsai` iteration_30000 (1,244,819 gaussians in the file, 375,472 after skybox clip, `sh_degree=3`), skybox alignment `up_sign=-1`:

| | main (DC-only) | this branch (full SH) |
|---|---|---|
| colors tensor | `[375472, 3]` | `[375472, 16, 3]` |
| colors VRAM | 4.5 MB | 72.1 MB |
| peak VRAM | 110.8 MB | 402.5 MB |
| load time | 0.73 s | 1.65 s |
| render 512x512 | 191.1 ms | **199.8 ms (+4.5%)** |

The SH evaluation is essentially free at render time; the cost is load-time and a 16x `colors` allocation that stays small in absolute terms. Worth knowing for the largest preset: `bicycle` is a 1.52 GB `.ply`, so its coefficient stack is proportionally larger (it is excluded from `GSPLAT_SKYBOX_ALIGN` anyway).

And the contribution is genuinely view-dependent rather than a constant offset -- mean `|dRGB|` against main over the same four-view orbit: 2.055, 2.273, 4.857, 4.323 (a 2.4x spread, up to 21.2% of pixels differing by more than 8/255).

### Preset swap

All four URLs resolve (HTTP 200): tabletop 3.1 MB, bonsai 308.7 MB, bicycle 1520.7 MB, stump 1230.5 MB. Download growth vs the iteration_7000 URLs they replace: bonsai x1.08, bicycle x1.70, stump x1.30.

I reproduced the `bonsai` control row of your bicycle re-evaluation at your stated camera (r=1.2 m, z=0.5 m, 960x540, four-view orbit): **near-white 0.0-5.4%** against your 0.0-4.9%. That is the load-bearing number for "the exclusion stands", and it holds independently. The asset also renders the correct capture -- the bonsai tree in its pot, the tripod and the table are all identifiable in the output.

One thing worth stating in the body: the iteration_7000 checkpoints **already carried `f_rest_*`** (that is why they were ~287 MB rather than ~60 MB). So enabling SH changes the rendered output of the presets the library ships *today*, independently of the checkpoint swap. Both changes are good, but they are two behaviour changes on the same default path, not one.

### Gate

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1343 files). mypy: branch-only errors empty -- the 14 reported are pre-existing in `examples/isaac_gs/` (reached via follow-imports, outside the lint targets) and none names a file this PR touches. Changelog fragment present. Blast radius (`tests/rendering` plus every test referencing the changed symbols plus the repo-wide docstring/ASCII/changelog guards, with `plyfile` and `gsplat` actually installed): **833 passed, 0 failed** on this branch; on main with this PR's test file copied in, **823 passed, 10 failed** -- same collected total, the ten base-only failures are exactly this PR's new pins, and there are no branch-only failures.

### Two follow-ups, neither blocking

1. **Nothing pins the rendered image.** `TestGsplatRenderShDegree` doubles `rasterization` and asserts the kwarg, so a transposed `f_rest` reshape would keep all 57 tests green while shipping wrong colors. The check above is cheap and deterministic: write a one-splat degree-1 `.ply` with an asymmetric `(3, 3)` rest block, render from `+X` and `-X`, and assert the centre pixel matches `clamp(clamp_min(sum_k basis_k(dir) * coeff[k] + 0.5, 0), 0, 1) * 255` within ~2/255. Gate it on `gsplat_rasterizer_available()`. The square rest block is what makes it discriminating -- a coefficient/channel mixup is undetectable at degree 2+ only because the shape changes, and at degree 1 it does not.
2. **The `.ply` decode is CPU-testable but skips in CI.** `plyfile` lives only in `sim-gs`, so the four most intricate new tests never run -- yet `_load_ply_splats(path, "cpu")` needs no GPU at all (verified: it returns the correct `(1, 4, 3)` on CPU). Adding `plyfile` to `[all]` would put that decode under CI. `gsplat` is a different question, since rasterizing needs a GPU runner.

Approving on the strength of the layout verification and the end-to-end render match.

PR #2327, opened 2026-08-16
fix(policies/wbc): point the missing-session refusal at the seam that installs one

cagataycali:
## Problem

`WBCPolicy.__init__` accepts `**kwargs` as a forward-compatibility absorber (the provider contract requires providers to ignore unknown keywords rather than raise). Any keyword the signature does not declare is therefore logged at DEBUG and discarded:

```
DEBUG WBCPolicy ignoring unknown constructor kwargs: ['policy_session', 'walk_session']
```

The `_run_session` refusal named exactly such a keyword as its remedy:

> WBCPolicy has no ONNX session loaded. Construct with a valid checkpoint (allow_missing_models=False), or inject a stub via `policy_session=`/`walk_session=` in tests.

So the refusal **survived its own remedy**. Applying it verbatim installs nothing and raises the identical error:

```python
p = WBCPolicy(allow_missing_models=True, policy_session=session, walk_session=session)
p.policy_session                      # None
p.get_actions_sync(obs, "")           # the same refusal, word for word
```

The route that works is attribute assignment, which the class docstring already documented (`:attr:`policy_session` / :attr:`walk_session``) and which **all 56 injections across the eight `tests/policies/wbc/` files use**. The advertised spelling had never worked for the audience the message named.

## Change

Restate the refusal in terms of the assignment that installs a session, and say why the constructor cannot take one:

> WBCPolicy has no ONNX session loaded. Construct with a valid checkpoint (allow_missing_models=False), or construct with allow_missing_models=True and then assign a session to the `policy_session` / `walk_session` attributes. The constructor absorbs unknown keywords per the provider contract, so a session handed to it is dropped rather than installed.

The `allow_missing_models` docstring now states that assignment is the only route, for the same reason.

No signature, attribute or control path changes - the diff is one message and one docstring paragraph. `ProtoMotionsPolicy` states the same principle from the other side: it declares every injectable as an explicit parameter and takes no `**kwargs`, "so a typo raises `TypeError` at build time rather than being swallowed". `WBCPolicy` needs the absorber, so its refusal has to name a spelling the absorber cannot eat.

## Tests

`tests/policies/wbc/test_missing_session_refusal_names_a_working_seam.py` grades the refusal mechanically rather than asserting on wording: it triggers the real error, parses the remedy back out of the message, and applies it.

| test | pre-fix | post-fix |
|---|---|---|
| every keyword it advertises is a declared constructor parameter | FAIL - `['policy_session', 'walk_session']` are absorbed | pass |
| assigning the attributes it names lifts the refusal | FAIL - names no attribute, so a reader has no stated way out | pass |
| the refusal fires when no session was installed (premise) | pass | pass |
| a session handed to the constructor is still dropped | pass | pass |
| an unknown keyword is still absorbed without raising | pass | pass |

2 failed / 3 passed before, 5 passed after. The three constant passes are the no-overreach controls: the provider contract still absorbs unknown keywords (turning the absorber into a raiser would be the wrong fix), and the dropped-keyword fact the message now states is pinned so the two cannot drift apart again.

Full `tests/` suite run against this branch and against `upstream/main` on the same interpreter: identical failure sets, no new failures.

One script per tree: it reads the refusal from the tree it runs in, parses the remedy out of that text, applies whatever the remedy says, then hands the policy to `run_policy` with a real GR00T SONIC ONNX session. Nothing is stubbed - only the message, and therefore the route it sends you down, differs.

remedy roundtrip

[full-resolution MP4](https://raw.githubusercontent.com/cagataycali/robots/media-wbc-session-seam/compare.mp4)

Left: the remedy as previously worded (constructor keyword) - `session installed: False`, `run_policy` returns `status=error` after 0 of 260 steps with `Policy failed: <the same refusal>`. Per-frame mean absolute pixel delta **0.0000**: the panel is static because the rollout never ran.

Right: the remedy as now worded (attribute assignment) - `session installed: True`, `status=success`, 260 of 260 control steps at 50 Hz, 157 frames. Per-frame delta 1.29 to 6.12, 156/156 frames in motion.

### DIFF
--- a/changelog.d/2327-wbc-missing-session-refusal-seam.md
+++ b/changelog.d/2327-wbc-missing-session-refusal-seam.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **WBC**: the "no ONNX session loaded" refusal named `policy_session=` / `walk_session=` as its remedy, but `WBCPolicy.__init__` absorbs unknown keywords per the provider contract, so a session passed that way was dropped and the identical refusal was raised again. The message now names the `policy_session` / `walk_session` attribute assignment that installs one, and states why the constructor cannot take it.
--- a/strands_robots/policies/wbc/policy.py
+++ b/strands_robots/policies/wbc/policy.py
@@ -197,10 +197,12 @@ class WBCPolicy(Policy):
             through the mesh ``tell()`` / ``policy_config`` path that forwards
             only constructor kwargs. Per-call ``target_velocity`` overrides it.
         allow_missing_models: Test/CI seam. When ``True`` the ONNX sessions are
-            not loaded eagerly (so unit tests can inject a stub session via
-            :attr:`policy_session` / :attr:`walk_session`). Production callers
-            leave this ``False`` so a missing checkpoint fails loudly at
-            construction.
+            not loaded eagerly, so a caller can install one by assigning to the
+            :attr:`policy_session` / :attr:`walk_session` attributes after
+            construction. Assignment is the only route: ``**kwargs`` below
+            absorbs unknown keywords, so a session passed to this constructor is
+            dropped. Production callers leave this ``False`` so a missing
+            checkpoint fails loudly at construction.
         **kwargs: Forward-compatibility absorber for the smart-string / registry
             resolution path. Per the #300 contract, providers MUST ignore
             unknown kwargs rather than raising.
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

This PR fixes a self-defeating error message: the `_run_session` missing-session refusal in `WBCPolicy` advertised `policy_session=` / `walk_session=` constructor keywords as its remedy, but `WBCPolicy.__init__` absorbs unknown keywords into `**kwargs` per the #300 provider contract, so applying the remedy verbatim installed nothing and re-raised the identical error. The message and the `allow_missing_models` docstring now point at attribute assignment (the route all existing WBC tests actually use) and explain why the constructor cannot take a session. The diff is one message string, one docstring paragraph, a changelog fragment, and a new test module that grades the refusal mechanically (parses the remedy out of the message and applies it) so the wording cannot drift back onto an absorbed keyword. No signature, attribute, control-path, or wire-format changes.

## What's good

- The regression test verifies the remedy by executing it rather than string-matching the wording, and includes no-overreach controls pinning that the `**kwargs` absorber contract is unchanged.
- Changelog fragment in `changelog.d/` per AGENTS.md step 3; no direct `CHANGELOG.md` edit.
- ASCII-only strings, no host paths in the test file, tight scope matching the PR title.
- Verified locally at head `62391cb8`: the 5 new tests pass and the full `tests/policies/wbc/` suite (516 tests) passes with no regressions.

PR #2329, opened 2026-08-16
fix(simulation/newton): advertise every base-contract method the backend delivers

cagataycali:
## Problem

`describe()` is the only guess-free route from an agent to a method name, and the base `SimEngine.describe()` is where the engine contract is stated. Three of its entries were added there on exactly that argument -- `remove_robot` as the inverse of `add_robot` (mirroring the advertised `add_object` / `remove_object` pair), plus the scene-variation and world-lifecycle families.

A backend that extends the base surface with `super().describe()` inherits those additions. The Newton backend returns a fresh dict literal, so it never received any of them. Six base-contract methods Newton implements or inherits unchanged were missing from `describe()["methods"]`:

| method | how Newton delivers it |
| --- | --- |
| `remove_robot` | `NewtonSimEngine.remove_robot` (own implementation) |
| `list_robots` | `NewtonSimEngine.list_robots` (own implementation) |
| `eval_policy` | `SimEngine.eval_policy` via `PolicyRunner` |
| `start_policy` | `SimEngine.start_policy` (documented synchronous passthrough) |
| `replay_episode` | `SimEngine.replay_episode` via `PolicyRunner` |
| `register_builtin_benchmarks` | `SimEngine.register_builtin_benchmarks` |

The result is a surface that is discoverable on one backend and invisible on another: on Newton a caller could learn how to add a robot and how to remove an *object*, but not how to remove a robot; and could run a policy (`run_policy` is advertised) but not discover how to score it over episodes. `docs/simulation/newton.md` already told readers `eval_policy` / `replay_episode` work here and that "`describe()` also advertises these methods ... so a single call surfaces the full contract" -- the guarantee the surface did not keep.

Measured on the real `describe()` call: **37 advertised methods before, 43 after**, no entry removed.

## Why not `super().describe()`

Merging the base surface looks like the obvious fix and is the wrong one. Newton does not implement `get_contacts` or `load_scene`; both resolve to the base's bare `raise NotImplementedError`. Inheriting the base surface would advertise them, turning an invisible capability into an advertised one that dead-ends when called. Each entry is instead listed with a signature string accurate for this backend -- including that `start_policy` is the ABC's documented synchronous passthrough to `run_policy` here, not a background thread (only MuJoCo overrides it with one).

## Change

- Newton's `describe()` advertises the six delivered base-contract methods, each grouped with its existing siblings (`remove_robot` beside `remove_object`, the rollout family beside `run_policy`, `register_builtin_benchmarks` beside `list_benchmarks`, `list_robots` beside `list_robots_info`).
- `describe()`'s docstring states why the surface is built from scratch and what that obliges, so the next contract addition is not silently missed.
- `docs/simulation/newton.md` names the fourth inherited rollout facade and how `start_policy` behaves on this backend.

## Tests

A parametrized guard in the file that already owns the discovery-surface rules. It reads each backend's advertised names from source (Newton and Isaac need simulator runtimes that cannot be installed together, which is why this drift was never compared against anything) and classifies a base-contract method as *delivered* unless it resolves to a bare `NotImplementedError` raise.

| test | before | after |
| --- | --- | --- |
| `test_every_delivered_contract_method_is_advertised[newton]` | **fail** -- names all 6 with the remedy | pass |
| `test_every_delivered_contract_method_is_advertised[base/mujoco/isaac]` | pass | pass |
| `test_a_backend_never_advertises_a_capability_it_refuses` | pass | pass |
| `test_the_reference_backend_needs_no_exception` | pass | pass |
| `test_the_contract_and_the_refusal_detector_are_both_non_vacuous` | pass | pass |

`1 failed, 49 passed` before; `50 passed` after. The four constant passes are the controls: MuJoCo and Isaac satisfy the rule with nothing added (so it is the house convention, not a Newton-specific patch), Newton must still *not* advertise `get_contacts` / `load_scene` (which is what fails if the `super()` shortcut is taken), and the premise test keeps a clean sweep from being vacuous.

Full `tests/` run on this branch and on `main`: identical failure sets, no new failures. `ruff check` / `ruff format --check` / `mypy` clean over `strands_robots tests tests_integ`.

## Discoverability before / after

Per base-contract method, per backend: is a capability the backend delivers findable on its surface? Generated by calling each tree's own `describe()`. Only the newton column changes (mujoco 0 cells, isaac 0 cells, newton 6 cells).

describe discoverability matrix

### DIFF
--- a/changelog.d/2329-newton-describe-delivered-contract-methods.md
+++ b/changelog.d/2329-newton-describe-delivered-contract-methods.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Newton discovery surface**: `describe()` on the Newton backend now advertises every base-contract method it delivers. Because it builds its surface from scratch instead of extending `SimEngine.describe()`, six methods it implements or inherits unchanged were absent from `describe()["methods"]` and therefore undiscoverable there - `remove_robot`, `list_robots`, `eval_policy`, `start_policy`, `replay_episode` and `register_builtin_benchmarks`. The two base-contract methods the backend does not implement, `get_contacts` and `load_scene`, remain unadvertised so the surface never points a caller at a `NotImplementedError`.
--- a/docs/simulation/newton.md
+++ b/docs/simulation/newton.md
@@ -96,8 +96,12 @@ no-op until then.
   from `list_bodies`) mounts the camera ON that body so a wrist camera tracks
   the arm. `remove_camera(name)` / `list_cameras()` round out the API and
   `describe()["cameras"]` lists every registered camera.
-- `run_policy` / `eval_policy` / `replay_episode` are inherited from the
-  `SimEngine` ABC - no backend-specific re-implementation.
+- `run_policy` / `eval_policy` / `replay_episode` / `start_policy` are
+  inherited from the `SimEngine` ABC - no backend-specific re-implementation.
+  `start_policy` is the ABC's synchronous passthrough to `run_policy` here;
+  only the MuJoCo backend runs a policy on a background thread. All four are
+  advertised in `describe()["methods"]`, as is every other base-contract
+  method this backend delivers.
 - `describe()` reports the active solver, available solvers, device, and
   the current gravity vector and timestep.
 - Gravity configured via `create_world(gravity=[x, y, z])` or `set_gravity`
--- a/strands_robots/simulation/newton/simulation.py
+++ b/strands_robots/simulation/newton/simulation.py
@@ -2301,10 +2301,19 @@ class NewtonSimEngine(DomainRandomizationMixin, NewtonRecordingMixin, SimEngine)
     def describe(self) -> dict[str, Any]:
         """Return a discovery surface describing this backend's capabilities.
 
+        This surface is built from scratch rather than by extending
+        :meth:`~strands_robots.simulation.base.SimEngine.describe`, so every
+        method the base contract advertises and this backend delivers has to be
+        listed here explicitly -- including the facades inherited unchanged from
+        the ABC. Extending the base surface instead would also advertise
+        ``get_contacts`` and ``load_scene``, which this backend does not
+        implement.
+
         Returns:
             Dict with the backend name, active solver, the solvers that can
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Advertises six base-contract methods the Newton backend delivers (`remove_robot`, `list_robots`, `eval_policy`, `start_policy`, `replay_episode`, `register_builtin_benchmarks`) on `NewtonSimEngine.describe()["methods"]`, which builds its surface from a fresh dict literal and therefore never inherited the base surface's entries. The change is additive-only: no entry removed, no behaviour change outside the discovery payload, and the two methods Newton genuinely refuses (`get_contacts`, `load_scene` — verified as bare `NotImplementedError` bodies in `SimEngine` with no Newton override) stay unadvertised, so the fix deliberately avoids the `super().describe()` shortcut that would advertise dead-ends. I verified each of the six resolves on `NewtonSimEngine` (own implementations for `remove_robot`/`list_robots`, inherited facades for the rest) and that the advertised signature strings match the actual base signatures, including the `start_policy` synchronous-passthrough note, which agrees with the ABC's docstring.

## What's good

- The new parametrized guard (`TestEveryDeliveredContractMethodIsOnTheBackendSurface`) pins the drift class for all backends, with controls that block both failure directions: the `super()` shortcut (refusals must not become advertised) and vacuous sweeps (the refusal detector is itself tested). Ran the full file at head: 50 passed.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention (AGENTS.md step 3); `ruff check` / `ruff format --check` clean on the touched files; no non-ASCII, no host paths.
- `docs/simulation/newton.md` was already promising this surface ("describe() also advertises these methods"), so the change closes a doc/behaviour gap rather than opening one.

## Verification suggestions

- `python -m pytest tests/simulation/test_sim_engine_describe_discovery.py -q -p no:cacheprovider --no-cov` — the new guard plus the pre-existing signature-accuracy tests in the same file cover both directions of the surface.
- Spot-check the refusal claim: `grep -n "def get_contacts\|def load_scene" strands_robots/simulation/newton/simulation.py` returns nothing, so both resolve to the base's bare raises.

PR #2331, opened 2026-08-16
feat(rendering): light the robot with the scene it stands in - GS-derived IBL, shadow catcher, color-pipeline contract

### LINKED ISSUE #2323: feat(rendering): light the robot with the scene it stands in - GS-derived IBL, shadow catcher, color-pipeline alignment for isaac_gs/mujoco_gs composites

yinsong1986:
Closes #2323

## What changed

The `isaac_gs` / `mujoco_gs` hybrid composite blended a robot pass and a photoreal background lit and toned in two unrelated worlds. This implements the issue's three stages — each usable on its own:

### Stage 1 — IBL from the GS scene (`strands_robots/rendering/ibl.py`, new)
- `render_environment_map` / `bake_environment_map`: bake an equirect environment map of **any** `BackgroundRenderer` from the robot's position **in the world frame** — six 90° cube faces rendered through the background's own `render`, so a `GsplatBackground`'s curated skybox alignment applies (unlike `bake_gsplat_panorama`, which bakes an unaligned backdrop from the scene centroid — it now delegates its cube→equirect core to the same code, pinned by the existing `TestBakeGsplatPanorama`).
- `derive_key_light`: dominant light direction + linear color from an env map (luminance × solid-angle weighting). Upper hemisphere by default — measured on the real `tabletop` preset, the full-sphere search derives elevation **-76°** (a key light from *underneath* the arm, aimed at the bright counter); the hemisphere-limited search derives a plausible overhead interior light (azimuth 142°, elevation 78°, slightly cool `(0.96, 1.0, 0.91)`).
- `environment_map_cache_path` encodes origin + every resolution knob into the cache name (the exact silent-stale-cache failure `bake_gsplat_panorama` documents).
- `examples/isaac_gs`: the `DomeLight` is textured with the baked map (latlong, stood up for the Z-up stage) and the key light is aimed/colored via `derive_key_light` by default; `--no-ibl` keeps the old hardcoded rig. A user `--panorama` feeds the dome directly; the procedural fallback keeps the legacy lights.

### Stage 2 — shadow catcher (`HybridCompositor`, `examples/isaac_gs/scene.py`)
- `plane_depth(cam, z)` (new, pure camera math): per-pixel analytic z-depth of a world-frame horizontal plane.
- `HybridCompositor(shadow_plane_z=...)`: foreground pixels matching that analytic depth are read as the catcher — excluded from the foreground mask (never painted, no z-fight with the backdrop surface) and their shading ratio (vs. the catcher's own 75th-percentile unshadowed level) is **multiplied onto the background in linear light**, floored by `shadow_min_factor`. `CompositeFrame` is unchanged, so `mujoco_gs` inherits the capability through the shared compositor.
- `examples/isaac_gs` adds a matte white plane at the support surface (world z=0, where the curated alignments seat the robot) by default and forwards its height to the compositor; `--no-shadow-catcher` opts out.

### Stage 3 — color-pipeline contract (`strands_robots/rendering/color.py`, new)
- The audit, written down once: every layer (RTX tonemapped output, MuJoCo framebuffer, GS DC colors optimized against sRGB training photos, panorama JPEGs) arrives **display-referred sRGB**, so the default byte-space blend is already aligned — but *light arithmetic* is not, and now runs in linear: the shadow multiply above, and an opt-in `HybridCompositor(blend_in_linear=True)` seam blend (fully-fg/fully-bg pixels stay byte-identical; pinned by test).
- IEC 61966-2-1 piecewise transfer functions + Rec. 709 luminance, with exact-round-trip tests over all 256 codes.

## Test surface

- 60 new CPU-only tests: bake camera poses/origin/FOV, equirect direction anchors, cache-key completeness, key-light direction/chromaticity/hemisphere behavior + guards, sRGB round trip + reference values, analytic plane depth (nadir/oblique/behind-camera), shadow behavior (never painted, linear multiply exactness, floor, tolerance, off-by-default pinned), linear-blend seam behavior, and option-domain guards for every new constructor knob.
- `hatch run lint` clean; full suite `MUJOCO_GL=egl hatch run test`: **30519 passed, 273 skipped, 0 failed** (95.24% coverage).
- GPU sanity check on this box (CUDA gsplat): baked the real `tabletop` preset from `(0, 0, 0.4)` in 2.3 s, 2048x1024, and derived the key light quoted above.

## Acceptance criteria from #2323

- [x] No change to the compositor contract (`CompositeFrame` fields untouched; `mujoco_gs` inherits stages 2–3 where applicable)
- [x] CPU-only tests for the env-map bake math (pose/direction correctness)
- [ ] A/B images per stage / GPU integ test for the visual path — **not deliverable from this dev box**: Isaac Sim requires an interactive EULA acceptance here, so the RTX side of the pipeline (dome texture orientation, catcher shadow on a real RTX render) could not be exercised. The bake itself was validated on real CUDA gsplat (numbers above). Follow-up below.

## Out of scope / follow-ups

- **A/B renders + GPU integ test on an Isaac host** — validate dome-texture azimuth alignment (USD latlong convention vs. the bake's `+X`-at-center convention; the key light is aimed in world coordinates and is convention-exact, the dome may need a yaw tune) and the catcher plane's RTX shadow, then capture the issue's per-stage A/B images.
- **`examples/isaac_gs/app.py` (Gradio) wiring** — the live app resolves its background *after* the scene is built and can hot-swap it, so IBL there needs a re-bake + re-light path; `render_demo.py` is wired end-to-end in this PR.
- **`examples/mujoco_gs` example wiring** — the library compositor now supports the catcher; adding the plane geom + light to the MuJoCo example scene is a small separate change.
- Prerequisite sibling #2321 (silent GS fallback) is unaffected by this change and still open.

Board: please move #2323 to "In review" on the [Strands Labs — Robots board](https://github.com/orgs/strands-labs/projects/2) (filing account lacks org project write).


cagataycali (approved):
Reviewed this on a CUDA box with `gsplat` actually installed, because `sim-gs` is not reachable from `[all]` and CI installs `.[all,dev]` — so the green check here never executes a real rasterization. Everything below is measured against the real `gsplat` CUDA path on the fully-trained 308 MB `bonsai` 30k `.ply` (with SH) and the curated `tabletop` `.spz`.

Net: the design is sound, the headline refactor claim holds bit-exactly, and I found no defect. Approving. Two notes below, neither blocking.

## The refactor is bit-compatible — verified, not assumed

`bake_gsplat_panorama` now delegates its cube-face loop to `render_environment_map`, and the docstring claims the grid-line convention keeps it "bit-compatible with the panorama bake it replaced". That is exactly the kind of claim that needs a rasterizer, so I ran the same bake on both trees:

| | `main` (inline loop) | this branch (delegated) |
|---|---|---|
| output md5 | `746af81bd8f7bb90e0317991db40c20e` | `746af81bd8f7bb90e0317991db40c20e` |
| bake time | 4.35 s | 4.10 s |

Decoded pixels: `max|Δ| = 0`, `mean|Δ| = 0.000000`, **0 differing pixels** over 512×1024×3. `_face_camera` reproduces the old `right`/`u`/`Twc`/`K` construction exactly, and `_equirect_directions(half_texel=False)` reproduces the old `np.arange` grid. Pure refactor, confirmed.

## The new capability does what it claims

The stated advantage over `bake_gsplat_panorama` is baking from the robot's position rather than the scene centroid. Moving the bake origin by 0.4 m changes **59.4%** of the map (`mean|Δ| = 17.33`, `max|Δ| = 252`), so the viewpoint is genuinely load-bearing.

Equirect wrap is seamless: column 0 vs column W−1 differ by `mean|Δ| = 2.415`, *below* the interior-neighbour baseline of `3.700` — no seam discontinuity.

And the justification for `upper_hemisphere=True` reproduces exactly. The docstring says the full sphere on `tabletop` derives "an elevation of −76 deg — a key light from *underneath*". I measure **−75.6°** for the full sphere and a sensible **+22.5°** above the horizon with the default. Good to see a documented measurement that actually replays.

env maps

The `tabletop` key-light color comes out strongly warm (`(1.00, 0.46, 0.52)` linear). I checked that this is data, not a bug: that scene's upper hemisphere genuinely measures warm (`[136.5, 109.7, 108.1]` sRGB) against a neutral floor (`[102.0, 105.9, 100.8]`), and gamma correctly amplifies a mildly-warm sRGB tone into a saturated linear chromaticity. `srgb_to_linear` handles the `uint8` → `[0,1]` normalization correctly, so the pipeline is right.

## Shadow catcher

With a tabletop-scale catcher (1.2 × 1.2 m) the depth test recognizes **98.78%** of the plane at the default `shadow_plane_tolerance=0.01`. Stretched to a 6 × 6 m floor it drops to **79.52%**, because the foreground depth buffer runs short of the analytic `plane_depth` in the far field (median −3.4 mm, worst −103 mm at the horizon). Unrecognized plane pixels stay foreground, i.e. untextured plane composited over the backdrop instead of being replaced by it. That is what the tolerance knob is for and it is fine at the scale a shadow catcher is actually used at — worth a sentence in the `shadow_plane_z` docs that the catcher should be sized to the working area rather than the whole floor.

## Suggestion: report sphere coverage from the bake

`render_environment_map` emits no coverage diagnostic, and the shipped `.ply` presets are inward-facing captures with no full-sphere support. Measured:

| preset | sphere coverage |
|---|---|
| `tabletop` (.spz) | **100.00%** |
| `bonsai` (.ply) | **57.26%** |

So a `bonsai` bake fed to a `DomeLight` is 42.7% black and silently under-lights the robot — the module's whole purpose being to light it with the room it stands in. `derive_key_light` only refuses when the searched region is *entirely* black, so at 57% coverage it returns an estimate biased toward wherever the capture happens to have support. A `logger.warning` (or a returned coverage fraction) when the map is substantially black would turn a silent under-lit result into an actionable one. Non-blocking — the coverage is an asset property, not a code defect, and `tabletop` (what the example actually lights with) is fully covered.

## Please merge `main` before this lands

`Detect an untested overlap with the base branch` reads SUCCESS, but it ran before #2326 merged and is now stale. The merge base (`f8037706`) predates #2326 (`29a424e7`), and both that PR and this one touch `strands_robots/rendering/backgrounds.py` — the file-level overlap the check exists to catch.

I merged `main` in locally and it composes correctly: `sh_degree=sh_degree` still present, all three `iteration_30000` presets intact, `GSPLAT_SCENE_PROVENANCE` intact, the `ibl` delegation intact, zero conflict markers. On that merged tree `tests/rendering` is **356 passed, 1 skipped (7.55 s)** with real CUDA `gsplat`, the bake is still bit-identical, and `ruff` plus `mypy` ("no issues found in 1346 source files") are clean. Worth pushing that merge so CI verifies the composition rather than the stale base.

## Scope

I could not exercise `examples/isaac_gs/scene.py` or `render_demo.py` end to end — that path needs `isaacsim`. The `strands_robots/rendering/*` surface, which is what ships, is covered above.

PR #2332, opened 2026-08-16
fix(rendering): composite premultiplied splat RGB once, alpha-normalize gsplat depth, honor the SPZ antialiased flag

### LINKED ISSUE #2322: bug(rendering): GsplatBackground.render - premultiplied alpha composited twice, accumulated depth never alpha-normalized (docstring says it is), AA-trained SPZ rendered classic

yinsong1986:
## What changed

Three defects in `GsplatBackground.render` (`strands_robots/rendering/backgrounds.py`), each measured in #2322:

1. **Premultiplied alpha composited twice.** gsplat's `render_colors` is already alpha-accumulated (premultiplied), so `rgb * alpha + fill * (1 - alpha)` weighted splat color by `alpha^2` — 2.0x darkening at half coverage (grey halos at splat edges, sparse regions, the zenith). Now composites with the premultiplied over-operator: `rgb + (1 - alpha) * fill`.
2. **Accumulated depth never alpha-normalized (docstring said it was).** `render_mode="RGB+D"` returns accumulated (alpha-weighted) depth: a splat at 4.0 m with alpha ~0.5 reported ~2.0 m, and the compositor's z-test trusts that value, so soft GS regions biased toward the camera and could wrongly occlude robot pixels. Now divided by `max(alpha, 1e-6)`; essentially-empty pixels (`alpha < 1e-4`) are promoted to `cam.zfar` rather than becoming an eps-division phantom occluder. The `render()` docstring now describes exactly what the code does (the class docstring's "divided through to give metric depth" is now true).
3. **AA-trained SPZ rendered classic.** `_load_spz_splats` parsed then discarded the header flags; bit 0 is the antialiased/mip-splatting training flag (the curated `tabletop.spz` ships `flags=1`). The loader now surfaces it and `render()` passes `rasterize_mode="antialiased"` for such assets (`classic` otherwise, including `.ply`).

## Test surface

Six new tests in `tests/rendering/test_backgrounds.py`, **all six fail on pre-fix code**:

- `TestGsplatRenderPartialAlpha` (3 tests, single synthetic gaussian at opacity 0.5 dead-center — accumulated alpha ~0.5, the regime where the issue's measurements are at their worst; no asset download; skipped without a CUDA gsplat rasterizer via the existing `gsplat_rasterizer_available()` probe):
  - center red channel of a red splat over black fill reads ~alpha (0.44–0.56), not alpha^2 (~0.25)
  - white splat over white fill stays white at every alpha (the alpha^2 bug gave ~191/255 at a=0.5)
  - center depth of a gaussian at 4.0 m reads 4.0 ± 0.05 (pre-fix ~2.0), and an untouched corner pixel reports `zfar`
- `TestGsplatRasterizeModeSelection` (2 tests, no CUDA needed): a synthetic `.spz` with `flags=1` selects `rasterize_mode="antialiased"`, `flags=0` keeps `classic`, pinned by capturing the kwargs `render()` forwards through a fake `gsplat.rasterization`
- `TestSpzGaussianSplatReader.test_load_surfaces_the_antialiased_header_flag`: the loader surfaces header bit 0 (plus the existing key-set assertion updated)

Validation: `hatch run format` clean, `hatch run lint` clean, full `hatch run test` green — 30460 passed, 276 skipped, coverage 95.27%. The two measured tests were additionally verified against real gsplat 1.4.0 + CUDA (they skip in the hatch env, which has no gsplat). Changelog fragment added per `changelog.d/` convention.

## Acceptance criteria (from #2322)

- [x] The three synthetic measurements become unit tests (single-gaussian scenes; no asset download; no GPU-brand dependence beyond gsplat availability)
- [x] Docstring matches behavior (repo rule: match docstrings to semantics)

## Out of scope

- #2321 (silent panorama fallback) — sibling issue, separate PR
- The compositor z-test itself (`compositor.py`) is unchanged; it now receives metric depth as it already assumed

Closes #2322

Board: please move #2322 to "In review" on the [Strands Labs — Robots board](https://github.com/orgs/strands-labs/projects/2) (filing account lacks org project write).


cagataycali (approved):
Approving. I verified the three claims on a CUDA rasterizer rather than taking them on trust, because the three load-bearing tests in this PR (`TestGsplatRenderPartialAlpha`) gate on `gsplat_rasterizer_available()` and **skip in CI** — the green check on this PR says nothing about them. Setup: NVIDIA L40S, `gsplat==1.4.0`, torch 2.10.0+cu128, `MUJOCO_GL=egl`.

## The three claims are correct

**1. `render_colors[..., :3]` is alpha-premultiplied.** Single gaussian, opacity 0.5, over black, `backgrounds=None`:

```
accumulated alpha  = 0.493954
red channel        = 0.493954
red / alpha        = 1.000000      <- premultiplied, exactly
```

So `rgb * alpha + fill * (1 - alpha)` weights the splat by `alpha^2`. The visible symptom, measured: a **white** splat at `a=0.494` over a **white** fill comes out **191.3/255** on `main` and **255.0** with this PR. A backdrop compositing over its own fill colour must be invariant; pre-fix it is not.

**2. `RGB+D` depth is accumulated, not metric.** Same scene, gaussian at 4.0 m:

```
RGB+D  depth  = 1.975816   == alpha * z  (0.493954 * 4.0)
RGB+ED depth  = 4.000000   <- gsplat's own expected-depth mode
accum / alpha = 4.000000   |difference from RGB+ED| = 0.00e+00
```

gsplat's own docstring is unambiguous — `"D"` renders `sum_i w_i z_i`, `"ED"` renders `sum_i w_i z_i / sum_i w_i` — and `rendering.py` only divides by alpha for `ED`/`RGB+ED`. Your hand division reproduces gsplat's expected-depth mode **bit-exactly**, so it is right by construction, and the `alpha < 1e-4 -> zfar` guard is the correct place to key the emptiness test now that a division is involved.

**3. The SPZ AA header flag.** Confirmed the loader surfaces bit 0 and `render()` forwards `rasterize_mode` accordingly.

## The user-visible consequence, end to end

The PR body scopes `compositor.py` out and asserts it "now receives metric depth as it already assumed". That is worth demonstrating, because the compositor's rule is `fg_depth + 1e-3 < bg_depth` with both sides documented in meters — so an alpha-biased backdrop depth becomes a **phantom near occluder that eats the robot**.

I built the case with a real SO-101 in MuJoCo and a real `.spz` backdrop (a sparse opacity-0.2 wall placed **4.50 m** from the camera, arm at 0.78 m), through the real `HybridCompositor`:

| | `main` | this PR |
|---|---|---|
| backdrop depth handed to the z-test | median **1.478 m**, min 0.018 m | **4.5000 m** (4.4999–4.5001) |
| arm pixels winning the depth test | 42147/48549 (**86.8%**) | 48549/48549 (**100%**) |
| arm pixels painted with backdrop instead | **6402 (13.2% of the silhouette)** | 0 |
| mean \|dRGB\| on those pixels | — | **70.4/255** |
| backdrop region brightness | **-38.7/255** vs reference | reference |

measurement

13% of the arm silhouette is replaced by backdrop fill in the pre-fix composite. This is not a subtle tint — it is geometry loss in the primary output of the photoreal path, and it is invisible to every gate we currently run.

## Gate

- The PR's 6 new tests, run against **`main`'s source** with the PR's test file: **6 failed** — `0.44 <= 0.247` (that is `alpha^2`), `191 >= 250`, `2.0053 <= 0.05`. Same tests on this branch: **6 passed**. The numbers match my independent measurement, so the tests are measuring what they claim.
- Full `tests/rendering` on this branch **with a real CUDA gsplat present**: **281 passed, 3 skipped**. Worth stating separately — CI runs that suite with gsplat absent, so this is the first time the gsplat-dependent paths have actually executed.
- `ruff check` clean, `ruff format --check` clean, `mypy strands_robots tests tests_integ` at the tree baseline (14, all under `examples/isaac_gs/`), changelog fragment present and the fragment gate passes.

## Two notes, neither blocking

**Merge order.** Four open PRs edit `rendering/backgrounds.py`. `#2332 x #2336` and `#2331 x #2332` merge cleanly; **`#2326 x #2332` conflicts** (both rewrite the `rasterization(...)` call and the compositing tail). Since this one closes a bug and #2326 is a feature built on the same lines, merging this first and rebasing #2326 onto it keeps the fix from being resolved away by hand. Worth re-running the three CUDA tests after that rebase, since #2326 changes what `colors` means to the rasterizer.

**`render_mode="RGB+ED"`.** gsplat will return alpha-normalized depth directly, and I measured it as bit-identical to your `accum / alpha`. Not a change I would ask for here — the explicit division is what makes the `alpha < 1e-4` promotion legible, and the comment now explains why — but it is worth a line somewhere that the two are equivalent, so a future reader does not "simplify" it in a way that silently drops the emptiness guard.

PR #2333, opened 2026-08-16
fix(inference): a PolicyServer stop must not kill the thread serving it

cagataycali:
## Problem

`PolicyServer.stop()` can kill the thread serving the server, and report nothing.

`start()` runs the websockets sync server's accept loop on a background thread.
`stop()` ends it by calling `shutdown()`, which closes the listening socket - from
the caller's thread, while the accept loop is still waiting on that same socket.
The sync server does not synchronize the two, so a stop that lands while the loop
is entering or polling raises straight out of `serve_forever()`.

It surfaces as **two different exceptions from the same call** -
`serve_forever()` registering the listening socket with its selector. `selectors`
reads `fileno()` to build its key and then hands the descriptor to `epoll_ctl`:

- a close landing **before** the read leaves `fileno()` at `-1` and raises
  `ValueError: Invalid file descriptor: -1`;
- a close landing **between** the read and `epoll_ctl` leaves a live-looking
  descriptor and raises `OSError: [Errno 9] Bad file descriptor`.

Which one appears depends on where in that call the close lands, not on the
websockets release. websockets 12.0 wraps the call in nothing, so both escape it;
13.0 through 17.x wrap it in `except ValueError: return`, which absorbs the first
shape and leaves the second. So a newer websockets narrows this race but does not
close it, and every supported release can produce the `OSError`.

Either way the serving thread dies, and nothing surfaces it: the thread is a
daemon, `stop()` returns normally, `_server` is cleared, and the server looks
cleanly shut down. Under pytest an unhandled thread exception is a *warning*, not
a failure - which is exactly how
`tests/inference/test_policy_server_lifecycle.py::test_stop_is_idempotent` stayed
green while the thread it started was crashing.

Measured on real sockets, `start()` then a prompt `stop()`:

| websockets | serving threads killed | shape observed |
| --- | --- | --- |
| 12.0 | 12 / 20 cycles | `ValueError: Invalid file descriptor: -1` |
| 17.0.1 (CI) | 3 / 12 cycles | `OSError: [Errno 9] Bad file descriptor` |

It is a race, so the counts vary run to run and with the machine - on one host the
newer releases did not lose it at all in 20 cycles while 12.0 lost it 12 times.
That variability is itself part of the problem: the failure is intermittent, silent,
and load-dependent.

PolicyServer stop must not kill its serving thread

## Change

The accept loop now runs through one helper, `PolicyServer._run_accept_loop`,
which absorbs the failure **only while a stop is in progress** - a
`threading.Event` set by `stop()` before it closes the socket.

Keyed on the stop, not on the exception type, deliberately. Matching on the type
would do two wrong things: hide a listening socket that failed on its own, and
encode a release-to-exception mapping that is not true in the first place - one
call raises either shape depending on scheduling. Keyed on the stop, a socket
failure with no teardown pending still propagates unchanged, whatever its type.

Two details the fix has to get right:

- `start()` clears the flag, so a server that is stopped and started again does
  not inherit the previous teardown and start swallowing real failures. `stop()`
  documents itself as safe to call twice and `start()` only refuses while a server
  is running, so restarting is a supported sequence.
- The foreground `serve()` path goes through the same helper - it races `stop()`
  from another thread in exactly the same way.

The absorbed case logs at DEBUG with `exc_info`, so the teardown is still
inspectable without making a normal shutdown noisy.

## Tests

`tests/inference/test_policy_server_shutdown_does_not_kill_the_serving_thread.py`.
The race is reproduced **deterministically** rather than by hoping the scheduler
lands in the window: `start()` imports `serve` from `websockets.sync.server` at
call time, so the tests patch that attribute with a server whose `serve_forever`
blocks until `shutdown` releases it and then raises. Both observed exception
shapes are parametrized, and one real-socket test covers the whole path with
nothing stubbed.

They assert on `threading.excepthook`, not on the server's own state - the state
was always clean, which is why this went unnoticed.

| test | on `main` | with this change |
| --- | --- | --- |
| a teardown failure does not escape the serving thread `[ValueError]` | **fail** - `killed its serving thread: [('ValueError', 'Invalid file descriptor: -1')]` | pass |
| a teardown failure does not escape the serving thread `[OSError]` | **fail** - `[('OSError', '[Errno 9] Bad file descriptor')]` | pass |
| a stopped server can be started again and still absorbs a teardown | **fail** | pass |
| the same failure without a stop pending still propagates `[ValueError]` | pass | pass |
| the same failure without a stop pending still propagates `[OSError]` | pass | pass |
| repeated real start/stop cycles leave no thread exception | fail on websockets 12.0 (9/12), passes where the race is not lost | pass |

3 failed / 3 passed on `main`; 6 passed with the change. The two that pass on both
are the controls that pin the boundary: they are what fails if the guard is ever
widened to swallow by exception type, which would hide a genuine socket error.

## Verification

- `ruff check` / `ruff format --check` and `mypy` over
  `strands_robots tests tests_integ`: clean, no new findings.
- `tests/inference`: 124 passed (118 before, plus the 6 added here).
- Full `tests/`, this branch and `main` side by side on the same interpreter:
  identical failing-test sets, so no regressions.

## Note on scope

An earlier revision of this branch treated the 12.0 `ValueError` as a dependency
floor problem and raised `websockets>=12.0` to `>=13.0`. CI, which resolves 17.0.1,
showed that was the wrong diagnosis: the same stop kills the same thread there too.
`except ValueError: return` in 13.0 removes one of the two shapes the same call can
raise, so raising the floor narrows the window without closing it. The packaging
change is gone; this is the whole fix.

### DIFF
--- a/changelog.d/2333-policy-server-shutdown-serving-thread.md
+++ b/changelog.d/2333-policy-server-shutdown-serving-thread.md
@@ -3,9 +3,14 @@
 `PolicyServer.stop()` no longer kills the thread serving the server. `stop()`
 closes the listening socket from the caller's thread while the accept loop is
 still waiting on it, and the websockets sync server raises out of
-`serve_forever()` when that happens - `ValueError: Invalid file descriptor: -1`
-on websockets 12.0, `OSError: [Errno 9] Bad file descriptor` on 13.0 through
-17.x. The serving thread is a daemon, so its death was reported nowhere: `stop()`
+`serve_forever()` when that happens. Both shapes come out of one call -
+registering the listening socket with the selector - and which one surfaces
+depends on where in it the close lands: before `fileno()` is read the descriptor
+is `-1` and it raises `ValueError: Invalid file descriptor: -1`, and between that
+read and `epoll_ctl` the descriptor still looks live and it raises
+`OSError: [Errno 9] Bad file descriptor`. 12.0 lets both escape; 13.0 through
+17.x guard the call against `ValueError` only, so every supported release can
+still die on the `OSError` and raising the dependency floor cannot fix this. The serving thread is a daemon, so its death was reported nowhere: `stop()`
 returned normally and the server looked cleanly shut down (measured 12 of 20
 `start()`/`stop()` cycles on websockets 12.0, 3 of 12 on 17.0.1). The accept loop
 now runs through a helper that absorbs the failure only while a stop is in
--- a/changelog.d/2333-policy-server-shutdown-serving-thread.md
+++ b/changelog.d/2333-policy-server-shutdown-serving-thread.md
@@ -1,0 +1,12 @@
+### Fixed
+
+`PolicyServer.stop()` no longer kills the thread serving the server. `stop()`
+closes the listening socket from the caller's thread while the accept loop is
+still waiting on it, and the websockets sync server raises out of
+`serve_forever()` when that happens - `ValueError: Invalid file descriptor: -1`
+on websockets 12.0, `OSError: [Errno 9] Bad file descriptor` on 13.0 through
+17.x. The serving thread is a daemon, so its death was reported nowhere: `stop()`
+returned normally and the server looked cleanly shut down (measured 12 of 20
+`start()`/`stop()` cycles on websockets 12.0, 3 of 12 on 17.0.1). The accept loop
+now runs through a helper that absorbs the failure only while a stop is in
+progress, so a socket failure with no stop pending still propagates.
--- a/strands_robots/inference/server.py
+++ b/strands_robots/inference/server.py
@@ -128,9 +128,48 @@ class PolicyServer:
         self.port = port
         self._server: Server | None = None
         self._thread: threading.Thread | None = None
+        # Set by stop() before it closes the listening socket, so the serving
+        # loop can tell a teardown apart from a genuine socket failure. See
+        # _run_accept_loop.
+        self._stopping = threading.Event()
         # Serialize inference so per-episode policy state is never interleaved
         # across connections (v1 single-client contract).
         self._lock = threading.Lock()
+
+    def _run_accept_loop(self, server: Server) -> None:
+        """Run the sync server's accept loop, absorbing the teardown race only.
+
+        :meth:`stop` ends the loop by closing the listening socket from another
+        thread, and the socket is exactly what the loop is waiting on. The
+        websockets sync server does not synchronize those two, so a stop that
+        lands while the loop is entering or polling raises out of
+        ``serve_forever()`` - and the exception it raises is not stable across
+        websockets releases: 12.0 raises ``ValueError: Invalid file descriptor:
+        -1`` while registering the socket, and 13.0 through 17.x raise ``OSError:
+        [Errno 9] Bad file descriptor`` while polling it.
+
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Fixes a silent thread-kill race in `PolicyServer`: `stop()` closes the listening socket from the caller's thread while the accept loop is waiting on it, and the websockets sync server raises out of `serve_forever()` (`ValueError` or `OSError [Errno 9]` depending on where in the selector registration the close lands). Because the serving thread is a daemon, the death was reported nowhere and the server looked cleanly shut down. The change routes both the background (`start()`) and foreground (`serve()`) accept loops through one helper, `_run_accept_loop`, which absorbs the failure only while a `threading.Event` set by `stop()` is pending - a socket failure with no stop in progress still propagates unchanged.

## What's good

- The guard keys on the stop being in progress rather than on exception type, and the two control tests pin exactly that boundary - widening the guard to swallow by type would fail them, which protects against a genuine listening-socket failure being hidden later.
- `start()` clears the flag so a restarted server does not inherit the previous teardown latch, with a dedicated regression test.
- The race is reproduced deterministically in tests (a stub whose `serve_forever` blocks until `shutdown` and then raises), plus one real-socket end-to-end test; assertions target `threading.excepthook` rather than server state, which is precisely why the existing lifecycle test missed this.
- Absorbed teardowns log at DEBUG with `exc_info` - inspectable, not a silent drop.
- Changelog fragment present and correctly named; scope is tight (no packaging change after the earlier floor-bump diagnosis was ruled out).

One non-blocking observation for a possible follow-up: `serve()` does not clear `_stopping` the way `start()` does, so an instance that was `stop()`ed and later re-`serve()`d in the foreground would inherit the latch and absorb a genuine socket failure. The CLI builds a fresh instance per invocation, so this only bites unusual instance reuse mixing the two entry points - not a merge blocker.

yinsong1986 (approved):
## Summary

Closes a silent thread-kill race in `PolicyServer`: `stop()` closes the listening socket from the caller's thread while the accept loop is still waiting on it, and `serve_forever()` raises `ValueError` or `OSError [Errno 9]` depending on where in the selector registration the close lands - killing the daemon serving thread with nothing reporting it. Both the background (`start()`) and foreground (`serve()`) paths now run the loop through `_run_accept_loop`, which absorbs the failure only while a `threading.Event` set by `stop()` before the close is pending; the same failure with no stop in progress still propagates, and `start()` clears the flag so a restart does not inherit the latch.

## What's good

- The guard keys on teardown state rather than exception type, and the two propagation control tests pin that boundary explicitly - a future widening to swallow-by-type fails them.
- The race is reproduced deterministically (stubbed `serve` whose loop blocks until `shutdown` then raises, both shapes parametrized) plus one real-socket cycle test, and the assertions target `threading.excepthook` rather than server state - the exact blind spot that let this go unnoticed.
- Absorbed teardowns log at DEBUG with `exc_info`, so this is not a silent drop (AGENTS.md no-silent-defaults).
- Changelog fragment present per step 3; scope discipline is good - the earlier `websockets` floor bump was correctly withdrawn once CI showed 17.x still loses the race.

## Verification suggestions

Reproduced locally on websockets 12.0 (the worst-affected release): the full `tests/inference` suite passes (124 tests), including the 6 added here and the 12-cycle real-socket test.

PR #2334, opened 2026-08-16
docs(recording): reset between rollouts in the published collection loop

cagataycali:
## Problem

`save_episode` cuts a dataset episode boundary. It does not re-initialize the world. Two published collection recipes omitted the `reset()`, so every rollout after the first began at the **previous rollout's terminal pose**:

- `save_episode`'s own docstring (`strands_robots/simulation/recording.py`) - the first place a reader looks for the episode-boundary primitive
- the Newton recording guide (`docs/simulation/newton.md`)

```python
sim.start_recording(repo_id=..., task=...)
for _ in range(n_episodes):
    sim.run_policy(robot_name=..., n_steps=...)
    sim.save_episode()   # flush this rollout as its own episode
sim.stop_recording()
```

The dataset that comes out has the **requested episode count and a bimodal set of recorded start states**: episode 0 begins at the scene's reset pose, and episodes 1..N-1 begin at a pose the robot is never reset into. Only one episode in the dataset ever demonstrates the state the robot will actually be in at inference time.

Measured on `so101` in MuJoCo, 3-episode loop, read back out of the produced LeRobotDataset:

| recipe | episodes recorded | `verify_dataset_episodes(3)` | max ⎮start_i − start_0⎮ |
|---|---|---|---|
| `run_policy(); save_episode()` (as published) | 3 | passes | **0.5347 rad (30.6 deg)** |
| `run_policy(); save_episode(); reset()` | 3 | passes | **0.0000 rad** |
| `run_policy(n_episodes=3)` | 3 | passes | **0.0000 rad** |

The episode *count* is correct either way, so `verify_dataset_episodes` - which the Newton guide recommends on the very next line - passes on the corrupt dataset. Nothing downstream reports it.

## Why this is a drift rather than an open question

The three-step form is already what the rest of the package documents:

- `run_policy(n_episodes=...)`'s docstring: *"it removes the need for a manual `for _ in range(n): run_policy(); save_episode(); reset()` loop"*
- the same loop, named again in `_run_episodes`' implementation comment
- `docs/recording.md` resets at the top of its manual loop, and explains `reset()`'s own episode-boundary behaviour
- `strands_robots/policies/persistent.py`'s module docstring shows the same three calls

Five sites state the three-step loop; two omitted the `reset()`. A sweep for `for`-loops publishing `run_policy` + `save_episode` without a `reset` finds exactly those two.

## Change

- Add `sim.reset()` to both published loops, and point at `run_policy(n_episodes=N)` as the first-class API for the common case.
- Correct `save_episode`'s justification paragraph. It previously read *"stats stay correct because each rollout's frames are isolated to their own episode rather than being mixed across mid-session `reset()` teleports"* - framing `reset()` as the hazard, the opposite of `docs/recording.md`'s reading of the same mechanism. It now states that `save_episode` cuts the boundary while `reset()` re-initializes the world, that the count passes either way, and that `reset()` is itself a boundary while recording.
- No behaviour change: `save_episode` still cuts a boundary without resetting, which is what lets a caller deliberately chain rollouts.

## Tests

`tests/simulation/test_collection_loop_resets_between_episodes.py` grades each site by **executing the call sequence it publishes** - extracted from the recipe's own source with `ast`, not paraphrased - against a real MuJoCo engine, then reading the recorded start states back out of the resulting LeRobotDataset. Nothing about the wording is asserted; a recipe whose loop body omits `reset` produces divergent starts and fails. The arguments the recipes elide as `...` are supplied by the test module, so it is the sequence of calls that is under test. Premise assertions fail loudly if a reformat means the guard would grade nothing.

Pre-fix **2 failed / 6 passed**, post-fix **8 passed**. The failure names the finding:

> `save_episode docstring` publishes the loop `['run_policy', 'save_episode']`, which does not re-initialize the world between rollouts. Episode starts differ by up to 0.5080 (episode 0 begins at [0.0, ...], episode 1 at [0.1107, 0.5080, ...]), so every episode after the first demonstrates a start state the robot is never reset into.

The six that pass on both revisions are the controls, and they are what bound the change: `save_episode` alone still does *not* reset (so the recipes genuinely need the third call, and a future reset inside `save_episode` would fail that control loudly); adding `reset` fixes it on the same engine and rollout; `run_policy(n_episodes=N)` already behaves this way; and the three sites that already publish the reset must keep it.

## Artifact

First frame of every episode, read back out of the produced LeRobotDataset (`observation.images.collect`) - so the panels are literally what the dataset contains.

recorded episode start states

Top row (as published): 34,046 px of a 246,400 px frame differ between episode 1 and episode 0 (13.8% of the frame). Bottom row (with `reset()`): **pixel-identical**, 0 px differ.

### DIFF
--- a/changelog.d/2334-collection-loop-resets-between-episodes.md
+++ b/changelog.d/2334-collection-loop-resets-between-episodes.md
@@ -1,0 +1,11 @@
+### Fixed
+
+Two published multi-episode collection recipes - `save_episode`'s docstring and
+the Newton recording guide - omitted the `reset()` between rollouts, so every
+episode after the first was recorded starting from the previous rollout's
+terminal pose. The dataset kept the requested episode count (so
+`verify_dataset_episodes` passed) while its recorded start states were bimodal:
+measured on `so101`, later episodes began up to 0.53 rad from episode 0. Both
+recipes now use the three-step `run_policy() -> save_episode() -> reset()` form
+the rest of the package already documents, and `save_episode`'s docstring states
+which of the two calls cuts the boundary and which re-initializes the world.
--- a/docs/simulation/newton.md
+++ b/docs/simulation/newton.md
@@ -260,9 +260,17 @@ sim.start_recording(repo_id="local/newton_demo", task="pick the cube", fps=50)
 for _ in range(n_episodes):
     sim.run_policy(robot_name="so100", policy_provider="mock", n_steps=200)
     sim.save_episode()          # flush this rollout as one episode
+    sim.reset()                 # next rollout starts from the scene pose
 result = sim.stop_recording()   # finalize parquet + video
 sim.verify_dataset_episodes(n_episodes)   # parquet-truth check
 ```
+
+`save_episode` cuts the episode boundary; `reset()` re-initializes the world.
+Drop the `reset()` and the dataset still has `n_episodes` episodes - so
+`verify_dataset_episodes` still passes - but every episode after the first
+starts from the previous rollout's final pose instead of the scene's, giving a
+bimodal distribution of recorded start states. `run_policy(n_episodes=...)`
+performs both steps for you.
 
 `start_recording` declares the dataset schema from the live scene - joint names
 from every robot (namespaced `robot__joint` in multi-robot scenes) plus any
--- a/strands_robots/simulation/recording.py
+++ b/strands_robots/simulation/recording.py
@@ -865,7 +865,12 @@ class DatasetRecordingMixin:
             for _ in range(n_episodes):
                 sim.run_policy(robot_name=..., n_steps=...)
                 sim.save_episode()   # flush this rollout as its own episode
+                sim.reset()          # next rollout starts from the scene pose
             sim.stop_recording()
+
+        ``run_policy(n_episodes=N)`` does all three and is the first-class API
+        for the common case; drive the loop yourself only when episodes need
+        different instructions, randomization or conditional logic.
 
         Without this call, every ``run_policy`` rollout in a session appends to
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Documentation fix with an executable guard: two published multi-episode collection recipes (`save_episode`'s own docstring and the Newton recording guide) omitted the `reset()` between rollouts, so every episode after the first was recorded starting from the previous rollout's terminal pose - a data-corruption hazard that `verify_dataset_episodes` cannot see because the episode count is correct either way. The PR adds `sim.reset()` to both loops, points readers at `run_policy(n_episodes=N)` as the first-class API, and corrects the docstring paragraph that previously framed `reset()` as the hazard rather than the missing step. No production behavior changes - `save_episode` still cuts a boundary without resetting, and a dedicated control test pins that so a future reset inside `save_episode` fails loudly.

## What's good

- The new test grades each recipe by executing the call sequence extracted from the recipe's own source with `ast` (verified against the head: both extraction paths yield `['run_policy', 'save_episode', 'reset']`), so a future edit that drops the `reset()` fails on measured start-state drift rather than on wording.
- Every `start_recording` call in the test passes `root=str(tmp_path)`, per AGENTS.md rule 15 (recording tests name their own dataset root), including the `LeRobotDataset` read-back.
- Premise assertions fail loudly if a reformat would leave the guard grading nothing; verified on the head that `docs/simulation/newton.md` has exactly one matching block and the three regression-guard files (`base.py`, `persistent.py`, `docs/recording.md`) all still name both calls.
- Changelog fragment present (`changelog.d/2334-collection-loop-resets-between-episodes.md`), no host paths, ASCII-only diff.

PR #86, opened 2026-04-01
feat: Robot() factory + top-level lazy imports

cagataycali:
## Summary

Introduces the `Robot()` factory function — a single entry point that returns either a **MuJoCo simulation** (`Simulation`) or a **hardware robot** (`HardwareRobot`) depending on mode detection.

**Design decisions:**
- **Sim-by-default** — `Robot("so100")` returns a MuJoCo sim unless hardware is detected or `mode="real"` is explicit. Safety-first: real hardware affects the physical world.
- **`STRANDS_ROBOT_MODE` env var** — override auto-detection with `sim` or `real`. Unrecognized values log a warning.
- **`robot.py` → `hardware_robot.py` rename** — `HardwareRobot` class moved to dedicated module; `robot.py` is now the factory.
- **`@overload` typing** — IDEs resolve `Simulation` or `HardwareRobot` based on `mode` literal.
- **Lazy GL backend** — `MUJOCO_GL` auto-configured on headless Linux only if `mujoco` is installed (via `importlib.util.find_spec` guard).

### Core: Robot Factory & Hardware
| File | Description |
|------|-------------|
| `strands_robots/robot.py` | **New** `Robot()` factory function with `@overload` typing, mode detection, env var support |
| `strands_robots/hardware_robot.py` | **Renamed** from old `robot.py` — `HardwareRobot` class (no emoji in result strings) |
| `strands_robots/__init__.py` | Top-level lazy imports + `find_spec` guard for GL backend |

### Registry
| File | Description |
|------|-------------|
| `strands_robots/registry/__init__.py` | Export updates |
| `strands_robots/registry/policies.py` | Policy registry updates |
| `strands_robots/registry/user_registry.py` | User registry updates |

### Simulation (policy runner, benchmarks)
| File | Description |
|------|-------------|
| `strands_robots/simulation/__init__.py` | Simulation subpackage init |
| `strands_robots/simulation/base.py` | Simulation base ABC |
| `strands_robots/simulation/benchmark.py` | Benchmark orchestrator |
| `strands_robots/simulation/benchmark_spec.py` | Benchmark spec DSL |
| `strands_robots/simulation/policy_runner.py` | Policy runner enhancements |
| `strands_robots/simulation/predicates.py` | Predicate system for benchmarks |
| `strands_robots/simulation/mujoco/tool_spec.json` | Tool spec additions |

### Benchmarks (LIBERO)
| File | Description |
|------|-------------|
[...]

PR #2335, opened 2026-08-16
fix(deps): raise the robot_descriptions floor to a release that ships every registry robot

cagataycali:
## Problem

`strands_robots/registry/robots.json` gives **57** robots an `asset.robot_descriptions_module` - the `robot_descriptions` submodule whose import fetches that robot's MJCF and meshes. `robot_descriptions` gains one module per newly packaged robot, so that field is a **version claim**, not just a name.

The declared floor read `robot_descriptions>=1.11.0,<2.0.0`, twelve releases below the capability. Measured by resolving every declared module against the released wheels (module existence only, via `importlib.util.find_spec`, so nothing is downloaded):

| release | registry modules it does not provide |
|---|---|
| 1.11.0 | **27** |
| 1.12.0 | 25 |
| 1.13.0 | 23 |
| 1.14.0 | 22 |
| 1.15.0 | 20 |
| 1.16.0 | 17 |
| 1.17.0 | 14 |
| 1.18.0 | 10 |
| 1.19.0 | 8 |
| 1.20.0 | 7 |
| 1.21.0 | 5 |
| 1.22.0 | 1 |
| 1.23.0 | **0** |

So of the **13** releases the old range admitted, **12 were missing at least one registered robot**. `openarm`'s `openarm_v1_mj_description` is the last to arrive, in 1.23.0.

### Why this is a missing robot, not a slow path

None of the 27 robots absent at 1.11.0 declares an `asset.source` GitHub fallback, and `assets.download.auto_download_robot` only tries `robot_descriptions` and then that GitHub source - the Menagerie `git clone` fallback lives in the bulk `download_robots` path, not the auto one. Measured on a resolver-legal `robot_descriptions==1.11.0` install with a clean assets dir:

```
robot_descriptions failed for so101: No module named 'robot_descriptions.so_arm101_mj_description'
RuntimeError: Failed to create sim robot 'so101': Robot 'so101' is registered but its
model file is not on disk: no 'robotstudio_so101/so101_new_calib.xml' under ...
Fetch it with the download_assets tool (robots='so101').
```

The remedy the message names cannot supply the module either. `so100` and `so101` - the arms the README leads with - were among the 27, along with `apollo`, `talos`, `unitree_h1_2`, `stretch3`, `tiago_dual`, `leap_hand`, `shadow_dexee`, `elf2`, `rby1`, `koch`, `piper`, `yam` and 14 more.

## Change

`[sim]` now declares `robot_descriptions>=1.23.0,<2.0.0`. The resolved version does not move (`uv.lock` already pinned 1.23.0, the newest release under the existing cap), so the lock diff is the five `requires-dist` lines only - this closes the range that was legal but broken.

## Tests

`tests/test_robot_descriptions_floor_ships_the_registry_modules.py` owns the number in `_FIRST_RELEASE_WITH_EVERY_REGISTRY_MODULE` and derives the checks from the registry rather than restating a list. On `main`: **1 failed / 4 passed**; after: **5 passed**.

The failing one names the offender and the stake:

```
AssertionError: robot_descriptions floor must be >= 1.23.0: the registry declares 57
description modules and no earlier release provides all of them, so an older resolve
leaves registered robots (so100 and so101 among them) with no way to fetch a model.
These specifiers admit older releases:
{'optional-dependencies.sim': 'robot_descriptions<2.0.0,>=1.11.0'}
```

The four that pass on both trees are the controls: the required floor is still inside every declared range (a floor above the cap would make it unsatisfiable rather than fixed), the cap stays inside the audited major, the registry still declares enough modules for the scan to mean something, and the installed distribution provides all 57.

That last one is the self-maintaining half - a robot registered later whose module needs a newer release fails there instead of shipping a floor that cannot fetch it. It is not a tautology: run against a real 1.11.0 wheel it fails, naming all 27 robots.

`tests/test_registry_integrity.py::test_robot_descriptions_module_names_are_import_safe` already pins that each declared module name is import-*safe* (matches the allowed character pattern); this file pins that it is import-*able* on the oldest install packaging admits.

## Docs

`docs/getting-started/installation.md` states the pin in the extras matrix (matching the `[lerobot]` row) and explains that this floor is set by the robot catalog rather than by an API surface.

## Artifact

The same script, the same tree and the same clean `STRANDS_ASSETS_DIR` in both rows - only the installed `robot_descriptions` version differs. Top row is a release the old range admitted; bottom row is the new floor, rendered in MuJoCo headless (EGL).

robot_descriptions floor vs registry modules

The upper row has nothing to render, which is the finding: the module the registry names is not in the distribution, and with no `asset.source` fallback there is no second route to a model file.

## Gate

`ruff check` + `ruff format --check` + `mypy` over `strands_robots tests tests_integ` clean; full `tests/` run shows no new failures against `upstream/main`.


yinsong1986 (approved):
## Summary

Raises the `robot_descriptions` floor in the `[sim]` extra from `>=1.11.0` to `>=1.23.0`, closing a range that was resolver-legal but broken: the registry names an `asset.robot_descriptions_module` for 57 robots, 12 of the 13 releases the old range admitted were missing at least one of them, and the affected robots (including `so100`/`so101`) declare no `asset.source` fallback, so a resolve inside the old range left them with no route to a model file. The resolved version does not move (`uv.lock` already pinned 1.23.0), so the lock diff is the five flattened `requires-dist` lines only. Verified locally: the registry declares exactly 57 modules, the single `robot_descriptions` declaration in `pyproject.toml` lives in the `[sim]` extra (the five lock lines are extras flattening, not hidden edits), and all 57 modules resolve via `find_spec` against a 1.23.0 install.

## What's good

- The new bound follows the AGENTS.md dependency-bounds convention (`>=1.0` dep, major capped).
- The test file derives its checks from the registry rather than restating a robot list, includes controls in both directions (a floor above the cap fails a different test than a floor below the capability), and guards its own relevance with a minimum-declared-modules floor so a shrunken registry can't hollow it out.
- `find_spec` keeps the module-existence probe from executing the submodule imports that clone upstream asset repos; the exception clause is narrow (`ImportError, ValueError`), per the #86 learnings.
- Changelog fragment under `changelog.d/`, docs updated in the same PR, no host paths in the test file.

## Verification suggestions

- `uv pip install 'robot_descriptions==1.23.0'` in a scratch venv, then `python -c "import importlib.util, json; mods=[e['asset']['robot_descriptions_module'] for e in json.load(open('strands_robots/registry/robots.json'))['robots'].values() if isinstance(e.get('asset'),dict) and e['asset'].get('robot_descriptions_module')]; print([m for m in mods if importlib.util.find_spec(f'robot_descriptions.{m}') is None])"` should print `[]` (reproduces the bottom row of the artifact without MuJoCo).

PR #2336, opened 2026-08-16
fix(examples/isaac_gs): fail loud when the photoreal 3DGS background can't initialize

### LINKED ISSUE #2321: bug(examples/isaac_gs): the photoreal path is silently optional - any GS failure demotes to the procedural gradient, and the shipped gallery is the fallback

yinsong1986:
# What changed

The isaac_gs example's photoreal (3DGS) background path was silently optional: `resolve_background` caught *any* GS failure — gsplat not importable, the CUDA rasterizer disabled (which is exactly what a plain `pip install gsplat` produces in the Isaac container, per the example's own README), a scene download/load error — and demoted to the flat procedural gradient `PanoramaBackground` with only a logged warning. A user who asked for the photoreal path got the gradient, and the shipped gallery is that fallback. This violates AGENTS.md rules 5/6 (raise on fatal, no silent defaults).

Closes #2321

## Changes

1. **Fail loud by default** (`examples/isaac_gs/background.py`): a GS background that cannot initialize now raises `RuntimeError` carrying the pre-built-wheel install hint (the README's install line), both for an explicit `--gsplat-ply` and for the default `prefer_gs=True` resolution. The original exception stays chained (`raise ... from exc`) so the report attributes the real cause. The old demotion is opt-in via `allow_fallback=True` — exposed as `--allow-fallback` on both `render_demo.py` and `app.py` — and still logs the warning when taken.
2. **`GsplatBackground` validates `ply_path` at construction** (`strands_robots/rendering/backgrounds.py`): exists + readable, raising `FileNotFoundError` / `PermissionError` where the caller supplied the path instead of at the first `render()` inside an app's catch-all. The splat decode and the torch/gsplat imports stay lazy in `_load`, so construction remains cheap and CPU-safe; `_load` keeps its own existence check as defense in depth against the file vanishing between construction and first frame.
3. **README/docstrings**: the default run is documented as fail-loud with the install hint, the zero-ML-deps mode as `--allow-fallback`, and the gallery captions now state plainly that the shipped images are the procedural-fallback render, not the GS path.

## Acceptance criteria from the issue

- [x] Fail loud by default when GS was explicitly requested (`--gsplat-scene` / `--gsplat-ply`, or the default `prefer_gs=True` resolution), with the install hint from the README
- [x] `--allow-fallback` for demo contexts that genuinely want the panorama demotion
- [x] Validate `ply_path` at construction (exists + readable)
- [ ] Regenerate `assets/hero_front.jpg` / `assets/wave.gif` from the real GS path — **deferred deliberately**: the sibling issues #2322 (rasterization math defects) and #2323 (lighting/color harmonization) change the pixels the GS path produces, so a gallery regenerated now would bake in the known defects and need a second regeneration. Follow-up filed as #2337; the README captions were corrected in the meantime so the example's face no longer misrepresents its degraded mode as the product.

## Test surface

- New `tests/test_examples_isaac_gs_background_failfast.py` (8 tests, no gsplat/CUDA needed — the rasterizer probe is stubbed at the library seam): raises with the hint for explicit-ply and default resolutions, names the scene, chains the cause on a post-probe download failure, `allow_fallback=True` demotes with a warning, explicit panorama paths never probe, and the CLI flag defaults off.
- `tests/rendering/test_gsplat_background_construction.py` re-pinned to the new contract: missing path raises `FileNotFoundError` **at construction**; with an existing file, construction stays dependency-free and lazy (regression tests fail on pre-fix code).
- `tests/rendering/test_backgrounds.py`: the `_load`-time missing-file test became a construction-time test, plus a new defense-in-depth test (file deleted after construction still raises in `_load`); constructor/clip tests now use an existing placeholder scene file.
- `hatch run lint` clean; `hatch run test` full suite: **30468 passed, 273 skipped** (with `MUJOCO_GL=egl`). One earlier run showed 2 failures in `tests/test_hardware_task_loop_dispatch.py` (`TestTheRolloutIsDrivenExactlyOnce`) — a real-time 50 Hz loop test that is load-sensitive; it passes in isolation, passes co-scheduled with the new tests, and the full suite re-run is green, so it is unrelated to this diff.

## Out-of-scope follow-ups

- Gallery regeneration from the real GS path (blocked on #2322 / #2323) — filed as #2337.
- `examples/mujoco_gs` has its own GS/panorama selection paths; this PR only changes the shared library constructor plus the isaac_gs resolver the issue names. If the same demotion pattern exists there it should get its own issue.

## Board

Issue #2321 should move to **In review** on the [Strands Labs — Robots board](https://github.com/orgs/strands-labs/projects/2) — the board *read* works with this token (the item exists, Status=In progress) but the Status write is refused (`Resource not accessible by personal access token`), so please move it manually.


cagataycali (changes requested):
Reviewed on a machine with a working `gsplat` CUDA rasterizer (L40S, gsplat 1.4.0, torch 2.10+cu128), because the load-bearing half of this change is only observable there. Thanks for keeping the branch merged with main through eight rounds — the fail-loud posture is the right call and the library-side validation is a genuine improvement. Two things below: what I could confirm, and one merge blocker.

## What I verified works

`sim-gs` is not reachable from `[all]`, and CI installs `.[all,dev]` — so on CI `gsplat_rasterizer_available()` is `False` and **every CI exercise of `resolve_background`'s GS paths takes the new raise branch or a stubbed probe**. The success path has never actually run. It does work:

| check | result |
|---|---|
| `resolve_background(gsplat_ply=<real .spz>)` | returns `GsplatBackground` — fail-loud correctly does **not** fire |
| `resolve_background()` (default preset, download + construct) | returns `GsplatBackground` |
| real 3DGS render through the new constructor | 320x320, mean 188.8, std 23.3 — genuine captured geometry |
| depth | 102400 finite px, median 0.2058 m (alpha-normalized per #2332) |
| `GsplatBackground(ply_path="/tmp/absent.ply")` | `FileNotFoundError`, message names the path and the requirement |
| mode `0o000` file | `PermissionError` |

I also checked the concern that construction-time validation could strand a caller that constructs before the file lands: every in-tree construction site (`examples/isaac_gs/background.py`, `examples/mujoco_gs/{app,agent}.py`, `backgrounds.py`'s panorama bake) resolves the path *first* — `ply = download_gsplat_scene(...)` then construct. No construct-then-download victim exists, so the eager check is safe.

Left panel is the photoreal path on this branch; right is the procedural gradient the old silent demotion returned (mean 188.8/std 23.3 vs 147.4/23.1, mean|delta| 43.1/255). That gap is exactly the bug #2321 describes:

real 3DGS vs the procedural fallback

## Merge blocker: the eager check breaks the "inject splats, never load" test idiom

`tests/rendering/test_backgrounds.py` builds several backgrounds whose `ply_path` is **deliberately never read** — the test assigns `bg._splats` directly and renders. The fixture names say so outright: `never-loaded-splats-injected.ply`, `scene.ply`. The new constructor raises before the injection can happen.

You already retargeted nine of these sites onto the new `scene_ply` fixture (`tests/rendering/test_backgrounds.py:680`) — one literal path was missed in the head as pushed, and #2326 brings two more, so three need it. The reason CI stays green on the missed one is that it is gated:

```python
def _require_cuda_rasterizer() -> None:
    pytest.importorskip("gsplat")        # <- CI has no gsplat, so it skips here
    ...
```

Measured, same commit, only the environment differing:

| tree | `gsplat` | `tests/rendering/test_backgrounds.py` |
|---|---|---|
| this PR's head, unmerged | blocked (= CI) | 42 passed, 3 skipped |
| **this PR's head, unmerged** | **real** | **3 failed**, 42 passed |
| PR merged with current main | blocked (= CI) | **5 failed**, 50 passed, 3 skipped |
| PR merged with current main | real | **8 failed**, 50 passed |

Two separate consequences:

- The **3** `TestGsplatRenderPartialAlpha` failures are already present in the head as pushed, on any machine that can rasterize. They come from #2332, which is already in this branch, so the green check is simply not evidence about them.
- The other **5** (`TestGsplatRenderShDegree`, `TestGsplatBackgroundClipSplats::test_clip_filters_sh_coefficient_colors_per_gaussian`) arrive with #2326. Those run on CPU (`device="cpu"` + a stubbed `gsplat` module), so they will turn CI **red** on the next main merge regardless of hardware.

### Fix, verified here

Retarget the three remaining sites onto the `scene_ply` fixture this PR already added — those tests only need the path to *exist*, they never read it:

- `TestGsplatBackgroundClipSplats::test_clip_filters_sh_coefficient_colors_per_gaussian`
- `TestGsplatRenderShDegree._render_and_capture` (thread the fixture through its two callers)
- `TestGsplatRenderPartialAlpha._single_gaussian_background` (thread through its three callers)

After that, on the merged tree:

- real rasterizer: 8 failed -> **58 passed**
- `gsplat` blocked (CI): 5 failed -> **55 passed, 3 skipped**
- `tests/rendering` + `tests/test_examples_isaac_gs_background_failfast.py`: **367 passed, 1 skipped**
- `ruff check/format strands_robots tests tests_integ`: clean (the 7 `Optional[...]` hits in `examples/isaac_gs/background.py` are pre-existing on main and outside the lint targets)

## Two smaller notes

**The conflict is purely additive.** `examples/isaac_gs/background.py` conflicts only because #2331 inserted `ENV_BAKE_ORIGIN` + `resolve_ibl_env_map` at the same anchor where this PR inserts `_GSPLAT_INSTALL_HINT` — right after `DEFAULT_GS_SCENE`. Keep both blocks; nothing else needs reconciling. I confirmed by symbol that the merged tree retains #2332's premultiplied composite and alpha-normalized depth, #2326's `sh_degree` plumbing, #2331's `resolve_ibl_env_map`/`ENV_BAKE_ORIGIN`, and this PR's hint/`allow_fallback`/constructor guard.

**One stale cross-reference the merge leaves behind.** #2331's `resolve_ibl_env_map` documents its own fallback as:

> any bake failure logs a warning and falls back — same always-renders-something posture as `resolve_background`

This PR inverts that posture, so the comparison becomes false on merge. Worth correcting in the same change (the IBL demotion itself is defensible — a dome light is a nicety, not the product — it is only the citation that goes stale).

Happy to re-run the GPU checks once the three sites are retargeted.

yinsong1986:
@cagataycali Thanks for running the GPU half of this — the success-path table and the side-by-side render are exactly the evidence CI can't produce here. Point by point:

**1. Merge blocker — the three remaining never-loaded `ply_path` literals.** Fixed in db79f31f, exactly as prescribed: all three sites now take the `scene_ply` fixture this PR added, threaded through the callers —

- `TestGsplatBackgroundClipSplats::test_clip_filters_sh_coefficient_colors_per_gaussian`
- `TestGsplatRenderShDegree._render_and_capture` (fixture threaded through both callers)
- `TestGsplatRenderPartialAlpha._single_gaussian_background` (threaded through all three callers)

No literal `ply_path="..."` remains in `tests/rendering/test_backgrounds.py`; the remaining `tmp_path / "scene.ply"` sites write real files. On the merged tree with `gsplat` blocked, `tests/rendering/test_backgrounds.py` reports 49 passed / 9 skipped here (the 5 CPU-stubbed failures you predicted are gone; my extra skips vs. your 55/3 are `plyfile` not being installed in this env — 6 `importorskip("plyfile")` sites). The 3 `TestGsplatRenderPartialAlpha` tests are CUDA-gated, so I'd appreciate the offered GPU re-run to confirm the 8-failed -> 58-passed side.

**2. The additive conflict with #2331.** Resolved in merge commit 4643481d, keeping both blocks at the `DEFAULT_GS_SCENE` anchor as you confirmed: `_GSPLAT_INSTALL_HINT` first, then `ENV_BAKE_ORIGIN` + `resolve_ibl_env_map`. Verified by symbol on the merged tree that #2332's premultiplied composite and alpha-normalized depth, #2326's `sh_degree` plumbing, #2331's `resolve_ibl_env_map`/`ENV_BAKE_ORIGIN`, and this PR's hint/`allow_fallback`/constructor guard all survive. The PR now reads `MERGEABLE`.

**3. The stale cross-reference in `resolve_ibl_env_map`'s docstring.** Corrected inside the same merge commit (4643481d), since that is where the sentence first coexists with the inverted posture. It no longer claims the "same always-renders-something posture as `resolve_background`"; it now says the demotion is safe because a dome light is a lighting nicety, and explicitly notes it is the *opposite* posture from `resolve_background`, which raises (issue #2321). The IBL demotion behavior itself is untouched, per your note that only the citation was stale.

**4. Your verification section** (success path constructs `GsplatBackground`, real render/depth, `FileNotFoundError`/`PermissionError` messages, and the no-construct-then-download audit): nothing to change — noting it here so the record shows the fail-loud raise not firing on a working rasterizer was confirmed on real hardware, which CI structurally cannot exercise (`sim-gs` not reachable from `[all]`).

Validation on the merged tree: `hatch run lint` clean; `tests/rendering` + `tests/test_examples_isaac_gs_background_failfast.py` -> 367 passed (skips are the `plyfile`/`gsplat` availability gates in this env); full `hatch run test` -> 32046 passed, 281 skipped, 0 failed.

cagataycali (changes requested):
Re-ran on a machine with a working `gsplat` CUDA rasterizer (L40S, gsplat 1.4.0, torch 2.10+cu128), as offered. **All three items from the last round are fixed and verified.** One new item is below - it is caused by the interaction of this PR's own two changes, and it only becomes measurable now that the constructor validation is in, so it could not have been raised earlier.

## 1. Test retarget (the merge blocker) - fixed

Reproduced the pre-fix numbers on `db79f31f^` and the post-fix numbers on the head as pushed. Same commit content in each row, only the environment differing. My CI simulation blocks **both** `gsplat` and `plyfile`, because neither is reachable from `[all]` (32 packages in the resolved closure) and CI installs `.[all,dev]` - which is also why our skip counts now agree:

| tree | `gsplat`/`plyfile` | `tests/rendering/test_backgrounds.py` |
|---|---|---|
| `db79f31f^` (pre-retarget) | real | **8 failed**, 50 passed |
| `db79f31f^` (pre-retarget) | blocked (= CI) | **5 failed**, 44 passed, 9 skipped |
| `ba5f76f6` (head) | real | **58 passed** |
| `ba5f76f6` (head) | blocked (= CI) | **49 passed, 9 skipped** |

That is exactly the 8 -> 58 and 5 -> 49 predicted last round, and your 49 passed / 9 skipped reproduces here independently. `grep 'ply_path="'` over that file now returns nothing, so no never-loaded literal remains.

Worth recording why the 5 mattered: they fail with `gsplat` blocked, i.e. they were not a hardware-only problem - they would have turned CI red on the next base merge.

## 2. Merge composition - fixed

The head already contains current `upstream/main` (`0640c9c9`), so the PR head *is* the merged tree and there is nothing left to reconcile. Verified by symbol rather than by absence of conflict, 0 conflict markers in either file:

| marker | source | hits |
|---|---|---|
| `_GSPLAT_INSTALL_HINT`, `allow_fallback` | this PR | 3, 11 |
| `ENV_BAKE_ORIGIN`, `resolve_ibl_env_map` | #2331 | 4, 1 |
| `sh_degree` | #2326 | 12 |
| premultiplied composite (`1.0 - alpha`), accumulated depth (`accum`) | #2332 | 1, 9 |
| constructor guard (`is_file()` / `FileNotFoundError` / `PermissionError`) | this PR | backgrounds.py:379-385 |

## 3. Stale cross-reference - fixed

`resolve_ibl_env_map` now reads as "the opposite posture from `resolve_background`, which **raises** when the photoreal background itself cannot initialize, issue #2321". Accurate, and it keeps the justification for the IBL demotion.

## The re-run I promised: success path on a live rasterizer

The fail-loud raise correctly does **not** fire where the rasterizer works, and the paths CI can never reach behave:

| check | result |
|---|---|
| `resolve_background(gsplat_ply=<real .spz>)` | `GsplatBackground` |
| `resolve_background()` (default preset, download + construct) | `GsplatBackground` |
| real 3DGS render through the new constructor | 320x320, mean 145.7, std 48.2 |
| depth | 102400 finite px, median 2.3541 m |
| `GsplatBackground(ply_path=<absent>)` | `FileNotFoundError`, names the path |
| `GsplatBackground(ply_path=<mode 0o000>)` | `PermissionError` |

Wider surface, `tests/rendering` + `tests/test_examples_isaac_gs_background_failfast.py`: **399 passed, 1 skipped** with a real rasterizer, **391 passed, 9 skipped** with it blocked. `ruff check` and `ruff format --check` over `strands_robots tests tests_integ` are clean (1351 files). The 7 `Optional[...]` hits in `examples/isaac_gs/background.py` are pre-existing on the base branch and outside the lint targets, as before.

## New item: `allow_fallback` does not survive the constructor raise

The two changes in this PR interact. Change 2 moves a missing/unreadable `ply_path` from a first-`render()` failure to a **construction** failure; change 1's `allow_fallback` only wraps the *rasterizer probe* on the explicit-`gsplat_ply` path. The construction call at `background.py:168` sits outside any `try`, so the new raises bypass the flag. Measured:

| construction fails because... | path 1, explicit `gsplat_ply` | path 3, default preset |
|---|---|---|
| `allow_fallback=False` | raw `FileNotFoundError` | `RuntimeError` + install hint |
| `allow_fallback=True` | **raw `FileNotFoundError`** | `PanoramaBackground` + warning |
| unreadable, `allow_fallback=True` | **raw `PermissionError`** | - |

Path 3 is correct in both modes. Path 1 honours the flag for a probe failure and not for a construction failure, so `render_demo.py --gsplat-ply <typo> --allow-fallback` crashes rather than demoting - the one mode whose documented purpose is "prefer rendering *something* over failing". The docstring currently promises otherwise for both paths: "On the GS paths (1 and 3) a failure -- ... a download/load error -- **raises** `RuntimeError` with the ... install hint ... Pass `allow_fallback=True` to restore the old demotion", and `Raises:` lists only `RuntimeError`.

Why the new suite does not catch it: both `allow_fallback` tests take the `rasterizer_unavailable` fixture, so they stub the **probe** - which is inside the branch that does honour the flag - and return before construction. `test_allow_fallback_demotes_to_panorama_with_a_warning` passes `gsplat_ply="capture.ply"`, a path that does not exist; on a machine with a working rasterizer that same call raises. So the 8 tests are green while the flag is broken for the two failure modes this PR introduces.

### Fix, verified here (12 lines)

```python
        logger.info("background: live 3DGS skybox from %s", gsplat_ply)
        try:
            return GsplatBackground(ply_path=gsplat_ply, skybox=True)
        except Exception as exc:  # noqa: BLE001 - re-raised as-is unless allow_fallback opts in
            if not allow_fallback:
                raise
            logger.warning(
                "background: uploaded 3DGS %s failed to initialize (%s: %s); falling back to "
                "procedural panorama (allow_fallback=True).",
                gsplat_ply,
                type(exc).__name__,
                exc,
            )
            return PanoramaBackground()
```

Deliberately a bare `raise` rather than wrapping in `RuntimeError`: the install hint is about a missing CUDA wheel, which is the wrong diagnosis for a typo'd path, and `FileNotFoundError: Gaussian Splat not found: <path>` is already the better message. That does mean the docstring's `Raises:` wants a line for `FileNotFoundError` / `PermissionError` on a caller-supplied path, distinct from the capability `RuntimeError`.

Measured with the patch applied: explicit real `.spz` still returns `GsplatBackground`, the default preset is untouched, `allow_fallback=False` on a bad path keeps its precise `FileNotFoundError`, and `allow_fallback=True` demotes with one warning for both the missing and the unreadable case. `tests/rendering` + the failfast suite + the test below: **402 passed, 1 skipped**.

A regression test that does not need a GPU - it forces the probe available and stubs the constructor at the seam `resolve_background` imports, so it exercises the construction branch anywhere:

```python
@pytest.fixture
def rasterizer_ok_but_construction_fails(monkeypatch):
    monkeypatch.setattr(rendering, "gsplat_rasterizer_available", lambda: (True, "ok"))

    def _boom(*_args, **_kwargs):
        raise FileNotFoundError("Gaussian Splat not found: capture.spz")

    monkeypatch.setattr(rendering, "GsplatBackground", _boom)


class TestAllowFallbackSurvivesConstruction:
    def test_construction_failure_demotes_when_fallback_allowed(
        self, rasterizer_ok_but_construction_fails, caplog
    ) -> None:
        with caplog.at_level(logging.WARNING, logger="isaac_gs.background"):
            bg = resolve_background(gsplat_ply="capture.spz", allow_fallback=True)
        assert isinstance(bg, PanoramaBackground)
        assert any("falling back to procedural panorama" in r.getMessage() for r in caplog.records)

    def test_construction_failure_still_raises_by_default(self, rasterizer_ok_but_construction_fails) -> None:
        with pytest.raises(FileNotFoundError, match="capture.spz"):
            resolve_background(gsplat_ply="capture.spz")
```

1 failed / 1 passed before the fix and 2 passed after, identically with a real rasterizer and with `gsplat` blocked - so it is load-bearing on CI, not only on hardware. The second test passes on both trees, which is what bounds the change: the default posture stays fail-loud.

This is the only remaining item and it is small - if you would rather keep the crash for a caller-supplied path and narrow the docstring instead, that is a defensible call and I will take that too. Either way I will approve on sight; the fail-loud posture and the library-side validation are a real improvement and the rest is verified.

yinsong1986:
@cagataycali Thanks for the L40S re-run and for handing over a fix already measured on live hardware - taken verbatim. Point by point:

**1. Test retarget / 2. Merge composition / 3. Stale cross-reference** - acknowledged as fixed and verified on your side; no further action taken on these. Your 49 passed / 9 skipped for `tests/rendering/test_backgrounds.py` with `gsplat`/`plyfile` blocked still reproduces here.

**4. `allow_fallback` does not survive the constructor raise** - fixed in `e59cf3d7`, using your patch as written:

- The `GsplatBackground(ply_path=gsplat_ply, skybox=True)` construction on the explicit-`gsplat_ply` path is now wrapped; `allow_fallback=True` demotes to `PanoramaBackground` with one warning naming the path and the exception, and `allow_fallback=False` keeps the **bare re-raise** rather than a `RuntimeError` wrap, for exactly the reason you gave - the install hint diagnoses a missing CUDA wheel, which is the wrong remedy for a typo'd path, and `FileNotFoundError: Gaussian Splat not found: <path>` is already the better message. A comment at the raise records that rationale so the next reader does not "fix" it into a wrap.
- The docstring now says the flag covers *every* failure mode on both GS paths, construction included, and `Raises:` gained `FileNotFoundError` / `PermissionError` entries for a caller-supplied `gsplat_ply`, distinct from the capability/preset `RuntimeError` - the docstring half of the choice you offered. (I took the fix rather than the narrow-the-docstring alternative: `--allow-fallback`'s documented purpose is "prefer rendering *something* over failing", and path 1 honouring the flag for a probe failure but not a construction failure is a distinction no caller could predict.)
- Your regression test is in as `TestAllowFallbackSurvivesConstruction`, with the `rasterizer_ok_but_construction_fails` fixture stubbing the constructor at the `strands_robots.rendering` seam `resolve_background` imports through - so it exercises the construction branch with no GPU, and I confirmed the 1 failed / 1 passed -> 2 passed transition across the fix with `gsplat` blocked.
- Two accuracy edits that follow from it: the test module docstring no longer claims every GS failure raises `RuntimeError` with the install hint (the explicit-ply construction case re-raises as-is), and the changelog fragment (`changelog.d/2321-...`) now records that `allow_fallback=True` covers the construction failure while the default posture keeps the precise exception.

**Validation** (CI-shaped environment, `gsplat`/`plyfile` blocked): `tests/rendering` + `tests/test_examples_isaac_gs_background_failfast.py` -> **393 passed, 9 skipped**, which lines up with your 402 passed / 1 skipped on a live rasterizer (the same 9 hardware tests skipped here, +2 new tests over your 391/9 pre-patch blocked run). `hatch run lint` clean (ruff check, format check, mypy over 1348 files); the 7 pre-existing `Optional[...]` hits in `examples/isaac_gs/background.py` remain outside the lint targets, as before. The full `hatch run test` suite aborts on this box at varying MuJoCo/native-extension sites that are disjoint from this PR's files and each pass in isolation - the same environmental instability as previous rounds, not attributable to this change.

cagataycali (approved):
Approving, as offered. The last remaining item is fixed and the docstring now matches the behaviour on both counts.

## What landed

The construction call on the explicit-`gsplat_ply` path is inside the `try` now, so `allow_fallback` covers the two failure modes this PR itself introduced:

```python
try:
    return GsplatBackground(ply_path=gsplat_ply, skybox=True)
except Exception as exc:  # noqa: BLE001 - re-raised as-is unless allow_fallback opts in
    if not allow_fallback:
        raise
    ...
    return PanoramaBackground()
```

The bare `raise` is kept, with the reason recorded at the site - the install hint diagnoses a missing CUDA wheel, which is the wrong remedy for a typo'd path, and `FileNotFoundError: Gaussian Splat not found: <path>` is already the better message. That was the option I flagged as defensible either way; keeping the precise exception is the right call, and it is now backed by a docstring that says so rather than contradicting it.

## The docstring no longer over-promises

This was half the original defect - the flag was documented as covering both paths while covering one. Now:

- the prose distinguishes capability/preset failures (`RuntimeError` + hint) from a caller-supplied `gsplat_ply` that fails to construct (re-raised as-is), and states the flag "covers every failure mode on both GS paths, construction included";
- `Raises:` gained `FileNotFoundError` and `PermissionError` entries, each scoped to `allow_fallback` being `False`, alongside the capability `RuntimeError`.

So the three outcomes a caller can actually get are each named, which is what the previous text collapsed.

## The regression pin is the load-bearing part

`TestAllowFallbackSurvivesConstruction` stubs the probe *available* and the constructor *failing*, at the seam `resolve_background` imports through - which is why it exercises the construction branch with no GPU. That matters because the eight pre-existing `allow_fallback` tests all took `rasterizer_unavailable`, so they returned before construction and stayed green while the flag was broken; a pin on the same fixture would have reproduced that blind spot.

The two cases bound the change from both sides: `test_construction_failure_demotes_when_fallback_allowed` fails on the pre-fix tree and passes now, and `test_construction_failure_still_raises_by_default` passes on both - so the default fail-loud posture is asserted to be unchanged, not merely believed to be. The class docstring records why the branch existed to be missed, which is the sentence that stops the `try` being narrowed again later.

## Standing from the earlier rounds, verified and needing nothing

The test retarget (`grep 'ply_path="'` clean, 8 -> 58 with a real rasterizer and 5 -> 49 with `gsplat` blocked), the merge composition (this head already contains `upstream/main`, with all five contributing changes confirmed by symbol and zero conflict markers), and the `resolve_ibl_env_map` cross-reference now reading "the opposite posture from `resolve_background`". The success path on live hardware behaves - real `.spz` and default preset both return `GsplatBackground`, genuine captured geometry, `FileNotFoundError` and `PermissionError` on the two bad-path cases.

Worth restating what the fail-loud posture buys, since it is the point of the PR: `sim-gs` is not reachable from `[all]` and CI installs `.[all,dev]`, so no CI run has ever rasterized. Before this change every one of those runs silently produced the procedural gradient where the photoreal path was requested, and the shipped gallery was the fallback. That is now a raise with an actionable hint, with the demotion available behind an explicit flag that finally works for every way the path can fail.

One note on state, not a request: `call-test-lint` is still in progress on this head and every other check is green. It stays a required check, so the merge remains gated on it - this approval just removes the review from the critical path.

PR #2340, opened 2026-08-16
fix(deps): name a remedy that can supply rclpy when the ROS 2 bridge refuses

cagataycali:
## Problem

With no ROS 2 distro sourced, `Robot("so101", mode="real", ros2_bridge=True)` (and the
sim/hardware `RosTelemetryBridge` behind it) refuses like this:

```
'rclpy' is required for the ROS 2 telemetry bridge (ros2_bridge=True)
Install with:
  pip install 'strands-robots[ros2]'
  pip install rclpy
```

Every line offered is a dead end.

| Offered remedy | What actually happens |
|---|---|
| `pip install 'strands-robots[ros2]'` | Exits 0. Installs the cyclonedds RMW binding. `rclpy` is exactly as missing as before. |
| `pip install rclpy` | `ERROR: No matching distribution found for rclpy` - it is not published on PyPI. |

So an operator who explicitly asked for `ros2_bridge=True` gets two instructions and no
way forward, and the first one is the worse kind: it reports success and changes nothing
relevant, which is verbatim the failure mode `tests/test_dependency_audit.py` already
describes for a misnamed extra.

The remedy that works was already written down in four places, none of them this message:

- `pyproject.toml`, on the `[ros2]` extra: "Those core client libraries are NOT distributed
  on PyPI - they ship with the system ROS 2 install (apt / RoboStack / conda) and cannot be
  `pip install`ed. ... Installing this extra does NOT by itself provision ROS 2 - source a
  real distro first."
- `docs/ros2-integration.md`: "The `[ros2]` extra ... only pulls the pip-installable
  `cyclonedds` DDS RMW binding. It does **not** provision ROS 2 by itself."
- The `strands_robots.ros_telemetry` module docstring: "`rclpy` and the ROS 2 message
  packages are optional, system-provided dependencies (they are not on PyPI)".
- `use_ros`, whose own hint reads "source a ROS 2 distro before launching the agent
  (e.g. `source /opt/ros/jazzy/setup.bash`) ... not on PyPI".

`test_require_optional_call_sites_name_declared_extras` pins that `extra=` names a
*declared* extra. `[ros2]` is declared, so the site passed - existing is not the same as
supplying the module, and that is the level that was never checked.

## Change

`require_optional` could not express a non-pip remedy: `pip_install` and `extra` each render
a `pip install` line. It gains `system_install=`, which replaces the pip block for a module
that arrives with a system package; `pip_install`/`extra` are then not consulted, so a
half-migrated call site cannot put the dead-end command back next to the working one.

The two `rclpy` sites (`hardware_robot._check_ros2_bridge_deps`, `RosTelemetryBridge.__init__`)
route through one shared `ROS2_SYSTEM_INSTALL_HINT`. The hardware site appends the
alternative only a `Robot` can take, because `ros2_transport` is the caller's choice there:
the pure-RTPS bridge publishes the same topics (both transports share the wire contract in
`RosTelemetryBase`) over cyclonedds, a pip wheel - the one route here that `[ros2]` really
does complete.

After:

```
'rclpy' is required for the ROS 2 telemetry bridge (ros2_bridge=True)
rclpy is not published on PyPI - it ships with a system ROS 2 install (apt / RoboStack / conda).
Source a distro in the shell that launches this process:
  source /opt/ros/jazzy/setup.bash   # or your distro / RoboStack / conda env
The [ros2] extra installs only the cyclonedds RMW binding; it does not provision ROS 2 by itself.
Or select the pure-RTPS transport, which publishes the same topics and needs no sourced distro:
  pip install 'strands-robots[ros2]'
  Robot(..., ros2_bridge=True, ros2_transport='rtps')
```

`docs/ros2-integration.md` also contradicted itself: two sections attributed `rclpy` to the
`[ros2]` extra while a third said that extra does not provision ROS 2. Both now say where
`rclpy` comes from and what the ImportError names.

## Scope, bounded by measurement

A sweep of all 40 `require_optional` module requests against PyPI found exactly one family
whose hints resolve to nothing:

- `sensor_msgs.msg` keeps its `ros-<distro>-sensor-msgs` hint: `ros-jazzy-sensor-msgs` does
  resolve on PyPI, so that is a template for the reader to fill, not a dead end.
- `curobo` and `motionbricks` keep theirs: `pip install -e git+...` and
  `pip install -e <path>` are runnable commands.
- Naming a third-party `ros-<distro>-rclpy` wheel would be a supply-chain endorsement (and
  no such name exists for humble or rolling), so this PR does not.

The regression guard's inventory is therefore just the two names pyproject already calls out
(`rclpy`, `rosidl_runtime_py`), not a new judgement.

## Tests

`tests/test_dependency_audit.py`, beside the guard that pins `extra=` names an extra that exists:

- `test_require_optional_offers_no_pip_remedy_for_a_system_provided_module` - AST sweep;
  a site requesting one of those libraries must pass `system_install=` and neither
  `pip_install` nor `extra`. Has a premise assert so it cannot pass vacuously.
- `test_the_rclpy_refusals_name_the_step_that_supplies_it` - asserts on the messages the two
  production sites really raise, with `import rclpy` forced to fail through a meta-path
  finder so the check holds whether or not the interpreter has a distro sourced.

`tests/test_utils.py`: `system_install` replaces the pip block; it wins over
`pip_install`/`extra`; and the pip block is byte-identical without it.

Pre-fix vs post-fix on the same files (`upstream/main` + these test files only):

| | pre-fix | post-fix |
|---|---|---|
| failed | 4 | 0 |
| passed | 92 | 96 |

The 4 pre-fix failures name both offending sites and both dead-end lines. The passes include
`test_pip_block_is_unchanged_without_system_install`, which holds on both trees - the
no-overreach control for the ~38 sites that legitimately do name a pip remedy.

## Refusal, before and after

Both panels are the real ImportError from one script run twice, same environment, only
`strands_robots` differing.

rclpy install hint, before and after

## Full suite

`pytest tests` on this branch and on `upstream/main`, same interpreter, run in parallel:
37 failed / 24 errors on both, and the two sorted failing-id sets are identical (`comm`
empty in both directions). Passed count 30388 vs 30383, i.e. exactly the 5 tests added
here. `ruff check` / `ruff format --check` / `mypy` clean over `strands_robots tests
tests_integ` (mypy reports the same 18 pre-existing errors on both trees).

### DIFF
--- a/changelog.d/2340-rclpy-refusal-names-a-working-remedy.md
+++ b/changelog.d/2340-rclpy-refusal-names-a-working-remedy.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- Point the missing-`rclpy` refusal at the step that supplies it. `ros2_bridge=True`
+  without a sourced ROS 2 distro offered two remedies and neither worked: `pip install
+  'strands-robots[ros2]'` exits 0 having installed only the cyclonedds RMW binding, which
+  leaves `rclpy` exactly as missing, and `pip install rclpy` fails because `rclpy` is not
+  published on PyPI. `require_optional` gains `system_install=` for a module that arrives
+  with a system package rather than from an index, and the two `rclpy` sites now name
+  sourcing a distro - plus, where the caller can take it, the pure-RTPS transport that
+  publishes the same topics over the pip-installable cyclonedds binding.
--- a/docs/ros2-integration.md
+++ b/docs/ros2-integration.md
@@ -181,10 +181,12 @@ ros2 topic list | grep so101          # /so101/joint_states, /so101/<cam>/image_
 ros2 topic echo /so101/joint_states   # live joint positions, updated every step
 ```
 
-`rclpy` is an optional, system-provided dependency (the `[ros2]` extra). When it
-is missing, `ros2_bridge=True` raises a clear `ImportError` at construction;
-`ros2_bridge=False` (the default) never touches ROS 2, so the base sim install
-stays lightweight. The bridge node is torn down cleanly on `destroy()`.
+`rclpy` is an optional, system-provided dependency: it arrives with a sourced ROS
+2 distro, not with the `[ros2]` extra (which installs only the cyclonedds RMW
+binding, as above). When it is missing, `ros2_bridge=True` raises an `ImportError`
+at construction naming the `source /opt/ros/<distro>/setup.bash` step that
+supplies it; `ros2_bridge=False` (the default) never touches ROS 2, so the base
+sim install stays lightweight. The bridge node is torn down cleanly on `destroy()`.
 
 See `examples/ros2/sim_bridge_demo.py` for a runnable end-to-end script.
 
@@ -256,7 +258,10 @@ ros2 topic echo /so101/joint_states   # live joint positions from the real arm
 The bridge is **opt-in**: `ros2_bridge=False` (the default) never touches ROS 2,
 so a robot only becomes a ROS 2 device when an operator explicitly asks for it -
 the same safety stance as `Robot(mode="sim")` being the default. When `rclpy` is
-missing, `ros2_bridge=True` raises a clear `ImportError` at construction. The
+missing, `ros2_bridge=True` raises an `ImportError` at construction naming both
+routes forward: sourcing a ROS 2 distro, or `ros2_transport="rtps"`, which
+publishes the same topics over the pip-installable cyclonedds binding and needs
+no distro at all. The
 inbound command path is on by default (`ros2_commands=True`); set
 `ros2_commands=False` for a read-only telemetry bridge that publishes but cannot
 be driven. A daemon thread spins the node so inbound commands are serviced
--- a/strands_robots/hardware_robot.py
+++ b/strands_robots/hardware_robot.py
@@ -36,6 +36,7 @@ from strands.tools.tools import AgentTool
 from strands.types._events import ToolResultEvent
 from strands.types.tools import ToolResult, ToolSpec, ToolUse
 
+from strands_robots.ros_telemetry import ROS2_SYSTEM_INSTALL_HINT
 from strands_robots.teleop_mixin import TeleopMixin
 from strands_robots.utils import (
     dds_domain_id_error,
@@ -52,6 +53,21 @@ if TYPE_CHECKING:
     from .policies import Policy
 
 logger = logging.getLogger(__name__)
+
+
+# Remedy for a missing ``rclpy`` when the caller asked for the rclpy transport.
+# The shared hint names the step that supplies rclpy; this adds the alternative
+# only a Robot can take, because ``ros2_transport`` is the caller's choice: the
+# pure-RTPS bridge publishes the same topics (both transports share the wire
+# contract in ``RosTelemetryBase``) over cyclonedds, which is a pip wheel, so it
+# is the one route here the ``[ros2]`` extra really does complete.
+_RCLPY_TRANSPORT_INSTALL_HINT = (
+    f"{ROS2_SYSTEM_INSTALL_HINT}\n"
+    "Or select the pure-RTPS transport, which publishes the same topics and "
+    "needs no sourced distro:\n"
+    "  pip install 'strands-robots[ros2]'\n"
+    "  Robot(..., ros2_bridge=True, ros2_transport='rtps')"
+)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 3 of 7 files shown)


yinsong1986 (approved):
## Summary

This PR fixes a refusal message whose every remedy was a dead end: with no ROS 2 distro sourced, `ros2_bridge=True` told the operator to `pip install 'strands-robots[ros2]'` (exits 0, installs only the cyclonedds RMW binding, leaves `rclpy` missing) and `pip install rclpy` (not on PyPI, fails outright). `require_optional` gains a `system_install=` parameter that replaces the pip block entirely for system-provided modules — and deliberately ignores `pip_install`/`extra` so a half-migrated call site cannot reintroduce the dead-end line. The two `rclpy` sites route through a shared `ROS2_SYSTEM_INSTALL_HINT` (points at sourcing a distro), with the hardware site additionally offering `ros2_transport='rtps'` — the one route the `[ros2]` extra genuinely does complete, since both transports share the wire contract in `RosTelemetryBase`. Docs are brought into agreement with themselves. Scope is bounded by a stated sweep of all 40 `require_optional` sites: `sensor_msgs`'s `ros-<distro>-sensor-msgs` template and the `curobo`/`motionbricks` hints are correctly left alone.

## What's good

- The AST sweep (`test_require_optional_offers_no_pip_remedy_for_a_system_provided_module`) has a premise assert so it cannot pass vacuously, and its inventory is taken verbatim from pyproject's existing statement rather than a new judgement.
- `test_the_rclpy_refusals_name_the_step_that_supplies_it` forces the import failure through a meta-path finder and resets the `_lazy_modules` cache inside a `MonkeyPatch.context()`, so it holds whether or not the interpreter has a distro sourced.
- `test_pip_block_is_unchanged_without_system_install` byte-pins the old message — a real no-overreach control for the ~38 sites that legitimately name a pip remedy.
- The new top-level import (`hardware_robot` -> `ros_telemetry`) is safe: `ros_telemetry`'s module scope imports only stdlib + `strands_robots.utils`, and there is no reverse import, so no cycle and no optional-dep cost at import time.
- Verified locally at the head SHA: the two new dependency-audit tests and the full `tests/test_utils.py` + `tests/test_dependency_audit.py` set pass (97 passed); the changed hunks are ASCII-clean per the AGENTS.md unicode-hygiene sweep.

- The follow-up commit `7cca6ea1` adds the `changelog.d/2340-*.md` fragment per AGENTS.md step 3; the diff is otherwise unchanged from `452765e9`, which is the tree the local verification below ran on.

## Verification suggestions

- In an environment without a sourced distro, run `python -c "from strands_robots.ros_telemetry import RosTelemetryBridge; RosTelemetryBridge(domain_id=0)"` and confirm the message matches the PR's after-panel (names `source /opt/ros/...`, contains no `pip install` line).
- Spot-check the guard bites: temporarily revert one call site to `extra="ros2"` and confirm `test_require_optional_offers_no_pip_remedy_for_a_system_provided_module` names that exact file:line.

yinsong1986 (approved):
## Summary

This PR replaces a refusal whose every remedy was a dead end: with no ROS 2 distro sourced, `ros2_bridge=True` told the operator to `pip install 'strands-robots[ros2]'` (exits 0, installs only the cyclonedds RMW binding, leaves `rclpy` missing) and `pip install rclpy` (not on PyPI). `require_optional` gains a keyword-only `system_install=` parameter that replaces the pip block entirely - `pip_install`/`extra` are deliberately not consulted when it is set, so a half-migrated call site cannot reintroduce the dead-end line. The two `rclpy` sites route through one shared `ROS2_SYSTEM_INSTALL_HINT`, and the hardware site appends the pure-RTPS alternative that only a `Robot` caller can take. Docs are brought into agreement with themselves.

## What's good

- `system_install=` is additive and keyword-only, so no existing caller changes behaviour; `test_pip_block_is_unchanged_without_system_install` pins byte-identical output for the ~38 sites that legitimately name a pip remedy.
- The shared hint has two callers, which is exactly the threshold AGENTS.md sets for lifting a string into a shared name; the RTPS-specific suffix stays local to the one caller that can act on it.
- The AST sweep in `tests/test_dependency_audit.py` carries a premise assert (`assert checked`) so it cannot pass vacuously if the `rclpy` sites move, and its inventory records what `pyproject.toml` already states rather than making a new judgement.
- The message-level test forces `import rclpy` to fail through a meta-path finder and resets the `_lazy_modules` cache inside a `MonkeyPatch.context()`, so it holds regardless of whether the suite's interpreter has a distro sourced.
- Changelog fragment present (`changelog.d/2340-...`); no non-ASCII, no host paths in the diff.

Verified locally at head `7cca6ea1`: the new tests in `tests/test_utils.py` and `tests/test_dependency_audit.py` pass, and `ruff check` / `ruff format --check` are clean on the changed files. The new top-level `hardware_robot -> ros_telemetry` import introduces no cycle (`ros_telemetry` imports only `strands_robots.utils`) and no heavy import-time cost.

PR #2342, opened 2026-08-16
fix(deps): raise the websockets floor to the release that ships the name the server imports

### LINKED ISSUE #2339: deps(inference): the websockets floor admits 12.0, which has no websockets.sync.server.Server - the name inference/server.py imports
## Summary

`[inference]` and `[cosmos3-service]` both declare `websockets>=12.0`. `strands_robots/inference/server.py` imports `Server` from `websockets.sync.server`:

```python
if TYPE_CHECKING:
    from websockets.sync.server import Server, ServerConnection
...
    self._server: Server | None = None
```

websockets 12.0 has no such name - it spells that class `WebSocketServer`. `Server` first appears in 13.0. So the declared floor describes an install in which the package's own annotation of `PolicyServer._server` does not resolve.

Measured against the released wheels:

```
$ python -c "import websockets, websockets.sync.server as s; print(websockets.__version__, hasattr(s,'Server'), hasattr(s,'WebSocketServer'))"
12.0  False True
13.0  True  False
16.0  True  False
```

This is deterministic - a name either exists in a release or it does not - and independent of any scheduling.

## Why nothing caught it

The import is under `TYPE_CHECKING`, so it is never executed at runtime. A 12.0 install resolves, imports, binds a port and serves; only a type checker run against that environment would notice, and CI type-checks against the locked resolve (currently 16.0), never against the declared floor. Nothing in the suite compares what the manifest admits with what the sources import.

## Scope

| extra | declared |
| --- | --- |
| `inference` | `websockets>=12.0` |
| `cosmos3-service` | `websockets>=12.0` |

Both should floor at `>=13.0`. Same floor for both deliberately: an environment resolves one `websockets`, so two different floors for one package would leave the lower one describing an install nobody gets. `[all]` pulls `strands-robots[inference]`, so it follows.

The locked resolve does not move (16.0 already satisfies both), so the lockfile change is the `requires-dist` specifiers only.

## Suggested shape of the fix

A specifier bump alone would drift again the next time the package reaches for a newer websockets API. What holds it is a table that owns the measurement, with the required floor *derived* from it rather than restated:

```python
_WEBSOCKETS_SYMBOL_FLOORS: dict[tuple[str, str], str] = {
    ("websockets", "<module>"): "12.0",
    ("websockets", "connect"): "12.0",
    ("websockets.sync.client", "<module>"): "12.0",
    ("websockets.sync.client", "ClientConnection"): "12.0",
    ("websockets.sync.client", "connect"): "12.0",
    ("websockets.sync.server", "ServerConnection"): "12.0",
    ("websockets.sync.server", "serve"): "12.0",
    ("websockets.sync.server", "Server"): "13.0",
}
```

plus four checks, all deterministic:

1. every declared websockets specifier floors at `max(table)`;
2. every websockets name the *sources* import has a row - an AST walk over the shipped package, including `TYPE_CHECKING` imports and dotted uses after a plain `import websockets`, which is what makes this class visible at all;
3. no recorded name is stale (each still resolves in the installed build), so the table cannot quietly stop describing the code;
4. the installed websockets satisfies the declared floor, so (3) is not reporting on a build no correctly-pinned user can get.

The AST walk over the current tree finds exactly eight names across five modules, so the table above is complete as of today; `Server` in `inference/server.py` is the only one above 12.0. Pattern follows `tests/test_strands_agents_floor_ships_the_imported_api.py`.

## History

#2333 originally bundled this bump with a behavioural claim that 13.0 is "the release whose server survives a prompt shutdown". That claim was false - the prompt-shutdown crash is a defect in `PolicyServer.stop()` that no floor fixes (#2338) - and #2333 correctly narrowed to the code fix and dropped the specifier change. This is the half that was left over, and it stands on its own for a reason that has nothing to do with the race.

## Out of scope

An upper cap on `websockets`. Adding one narrows every consumer's resolve and is a separate call from correcting a floor that is measurably wrong.

---

<sub>This issue was generated by an AI contributor. The measurements and technical analysis were made by the AI agent.</sub>

cagataycali:
Closes #2339

## Problem

`strands_robots/inference/server.py` annotates `PolicyServer._server` with `websockets.sync.server.Server`:

```python
if TYPE_CHECKING:
    from websockets.sync.server import Server, ServerConnection
...
    self._server: Server | None = None
```

websockets 12.0 has no such name - it spells that class `WebSocketServer`, and `Server` first appears in 13.0. Both `[inference]` and `[cosmos3-service]` declared `websockets>=12.0`, so the manifest described an install in which the package's own annotation does not resolve.

Measured against the released wheels in a pristine environment, across 9.1, 10.0, 10.4, 11.0, 11.0.3, 12.0, 13.0, 13.1, 14.0, 15.0, 16.0 and 17.0.1. Of the nine websockets names the shipped sources reach for:

| release | names missing |
| --- | --- |
| 9.1, 10.x | all seven `websockets.sync.*` names |
| 11.0 - 12.0 | `websockets.sync.server.Server` |
| 13.0 - 17.0.1 | none |

So `Server` is the one name the declared floor was missing, and 13.0 is where it lands.

## Why nothing caught it, and what the consequence is

The import lives under `TYPE_CHECKING`, so it is never executed. On a 12.0 install the package imports, binds a port, serves and stops cleanly:

```
websockets 12.0
bound port: 34237 running: True
stopped cleanly; runtime is unaffected on 12.0
```

Only a type checker notices - and it names the dependency:

```
$ mypy strands_robots/inference/server.py     # websockets 12.0
strands_robots/inference/server.py:40: error: Module "websockets.sync.server" has no attribute "Server"  [attr-defined]
Found 4 errors in 4 files

$ mypy strands_robots/inference/server.py     # websockets 16.0
Found 3 errors in 3 files      # the three pre-existing baseline errors
```

A resolve anywhere inside the declared range therefore produces an install that fails the type gate for a reason no source change can fix. CI type-checks the locked resolve (16.0 today), never the declared floor, and nothing compared what the manifest admits against what the sources import.

## Change

Both specifiers move to `websockets>=13.0`. They declare the *same* floor on purpose: an environment resolves one `websockets`, so two different floors would leave the lower one describing an install nobody gets. `[all]` pulls `strands-robots[inference]` and follows. The locked resolve does not move (16.0 already satisfies both), so `uv.lock` changes only the three `requires-dist` specifiers.

`tests/test_websockets_floor_ships_the_imported_api.py` owns the measurement in one table and *derives* the required floor from it rather than restating a number, following `tests/test_strands_agents_floor_ships_the_imported_api.py`:

```python
_PROBED_RELEASES = ("9.1", "10.0", "10.4", "11.0", "11.0.3", "12.0",
                    "13.0", "13.1", "14.0", "15.0", "16.0", "17.0.1")

_WEBSOCKETS_SYMBOL_FLOORS: dict[tuple[str, str], str] = {
    ("websockets", "<module>"): "9.1",
    ("websockets", "connect"): "9.1",
    ("websockets.sync.client", "<module>"): "11.0",
    ("websockets.sync.client", "ClientConnection"): "11.0",
    ("websockets.sync.client", "connect"): "11.0",
    ("websockets.sync.server", "<module>"): "11.0",
    ("websockets.sync.server", "Server"): "13.0",
    ("websockets.sync.server", "ServerConnection"): "11.0",
    ("websockets.sync.server", "serve"): "11.0",
}
```

The two `websockets`-level names are present at 9.1, the oldest release probed, and are recorded there rather than at their true origin: only the *maximum* of the table feeds the floor, so an older origin cannot change the answer, and a row can never understate the release the code needs.

An AST walk over the shipped sources feeds the table. It covers the three shapes the package actually uses, and the first is what makes this defect class visible at all:

* `from websockets.sync.server import Server` - including under `TYPE_CHECKING`;
* `import websockets` followed by `websockets.connect`;
* `import websockets.sync.client as _wsc` followed by `_wsc.connect`.

Five properties, all deterministic:

1. every declared specifier floors at `max(table)`;
2. all declared specifiers agree on that floor;
3. every name the sources reach for has a row, so a future use of a newer API cannot leave the floor behind;
4. no row is stale, and every recorded number comes from `_PROBED_RELEASES`, so a value read off a changelog rather than measured against a wheel fails rather than being taken on trust;
5. the installed websockets satisfies the declared floor, so the per-name resolution checks are not reporting on a build no correctly-pinned install can get.

A premise check pins that the walk reaches the sources at all, so a refactor cannot turn a silently-empty audit into a green one.

## Tests

Against `upstream/main` (specifiers at `>=12.0`), **1 failed / 15 passed**, naming both offenders and the reason:

```
AssertionError: websockets floor must be >= 13.0 because the package reaches for
['websockets.sync.server.Server']; these specifiers admit older releases:
{'optional-dependencies.cosmos3-service': 'websockets>=12.0',
 'optional-dependencies.inference': 'websockets>=12.0'}
```

After the change, 16 passed. The 15 constant passes are the controls - they hold on both trees, so the table, the walk and the per-name resolution are unaffected by the specifier edit.

The guard is not a tautology either. Run the same file against a real websockets 12.0 wheel with the corrected specifiers and 2 of 16 fail - one for the mis-pinned environment, one for the name itself:

```
FAILED ...::test_the_installed_websockets_satisfies_the_declared_floor
  - installed websockets 12.0 does not satisfy optional-dependencies.cosmos3-service: websockets>=13.0
FAILED ...::test_name_resolves[websockets.sync.server-Server]
  - websockets.sync.server.Server is recorded in _WEBSOCKETS_SYMBOL_FLOORS and reached for by
    the package, but the installed websockets does not provide it
```

Full suite, this branch vs `upstream/main` on the same interpreter: **61 failing ids on each, identical sets** (`comm` empty in both directions), so no regressions. Passed count 30435 vs 30419 - exactly the 16 tests this adds. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`, `mypy` at the pre-existing baseline, `uv lock --check` parity holds, `tests/inference` 124 passed.

## Measurement

websockets floor audit

Left: every websockets name the sources reach for, probed against the released wheels. Seven names arrive together with the synchronous implementation in 11.0; `websockets.sync.server.Server` is the single cell that moves after that, and 13.0 is where it lands - which is what sets the floor. Right: the two verdicts a 12.0 install produces - a clean runtime, because the import is never executed, and a type gate that fails naming the dependency.

## Out of scope

An upper cap on `websockets`. Adding one narrows every consumer's resolve and is a separate call from correcting a floor that is measurably wrong; the table records the releases that were probed, not a supported range.

## Docs

* `docs/getting-started/installation.md` - the `[cosmos3-service]` row names the pin, and a paragraph beside the existing `[sim]` floor note explains where this floor comes from and why both extras share it.
* `docs/inference/remote.md` - the install line and the surrounding note state the `>=13.0` floor and the name that sets it.

<sub>This pull request was authored by an AI contributor. The implementation, measurements and tests were produced by the AI agent.</sub>

### DIFF
--- a/changelog.d/2342-websockets-floor-ships-the-server-name.md
+++ b/changelog.d/2342-websockets-floor-ships-the-server-name.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **deps**: `[inference]` and `[cosmos3-service]` now require `websockets>=13.0`. Both declared `>=12.0`, but `strands_robots/inference/server.py` annotates `PolicyServer._server` with `websockets.sync.server.Server`, a name that first ships in 13.0 (12.0 spells the class `WebSocketServer`). The import is under `TYPE_CHECKING`, so a 12.0 install serves correctly at runtime and only fails type checking, reporting the missing attribute rather than the too-old dependency. `tests/test_websockets_floor_ships_the_imported_api.py` now derives the required floor from a measured table of every websockets name the sources reach for, so a future use of a newer API cannot leave the floor behind.
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -15,7 +15,7 @@ Requires **Python >= 3.12**. Examples use [`uv`](https://docs.astral.sh/uv/) (`c
 | `[sim-mujoco]` | `sim` + `mujoco`, `imageio`, `imageio-ffmpeg` | Any `Robot()` with default `mode="sim"` |
 | `[lerobot]` | `lerobot>=0.6.1,<0.7.0` | `LerobotLocalPolicy` + dataset recording |
 | `[groot-service]` | `pyzmq`, `msgpack` | `Gr00tPolicy` (ZMQ to a GR00T container) |
-| `[cosmos3-service]` | `msgpack`, `websockets` | `Cosmos3Policy` (WebSocket to Cosmos 3 server) |
+| `[cosmos3-service]` | `msgpack`, `websockets>=13.0` | `Cosmos3Policy` (WebSocket to Cosmos 3 server) |
 | `[mesh]` | `eclipse-zenoh`, `json5` | Multi-robot mesh discovery + RPC |
 | `[mesh-iot]` | `mesh` + `awsiotsdk`, `awscrt`, `boto3` | AWS IoT Core transport for mesh |
 | `[benchmark-libero]` | `libero` eval deps | LIBERO benchmark suite |
@@ -36,6 +36,13 @@ MJCF and meshes, and that package gains one module per newly packaged robot.
 registered robot - on an older one, robots such as `so100` and `so101` have no
 module to import and no fallback, so `Robot("so101", mode="sim")` cannot resolve
 a model file.
+
+The `websockets` floor in `[inference]` and `[cosmos3-service]` is set the same
+way. `strands_robots/inference/server.py` annotates `PolicyServer._server` with
+`websockets.sync.server.Server`, which first ships in websockets 13.0 - 12.0
+spells that class `WebSocketServer`. Both extras declare `>=13.0`, and they
+declare the *same* floor on purpose: an environment resolves one `websockets`, so
+two different floors would leave the lower one describing an install nobody gets.
 
 ## Platform notes
 
--- a/docs/inference/remote.md
+++ b/docs/inference/remote.md
@@ -23,11 +23,13 @@ hardware control loop - a remote one works too.
 ## Install
 
 ```bash
-pip install 'strands-robots[inference]'   # pulls websockets (numpy-agnostic)
+pip install 'strands-robots[inference]'   # pulls websockets>=13.0 (numpy-agnostic)
 ```
 
 The extra depends only on `websockets`, so it composes cleanly with `lerobot`
-(`numpy>=2`) in the same environment.
+(`numpy>=2`) in the same environment. The `>=13.0` floor is the release that
+first ships `websockets.sync.server.Server`, the class this module annotates
+`PolicyServer._server` with; 12.0 spells it `WebSocketServer`.
 
 ## 1. Start the server (GPU box)
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,13 @@ cosmos3-service = [
     # extra numpy-version agnostic so it composes cleanly with `lerobot`
     # (numpy>=2) for dataset recording in the same env.
     "msgpack>=1.0.0,<2.0.0",
-    "websockets>=12.0",
+    # >=13.0 is the floor of the websockets API the package reaches for, not a
+    # preference: `websockets.sync.server.Server` (annotating
+    # PolicyServer._server) first ships in 13.0 - 12.0 spells that class
+    # `WebSocketServer`. One environment resolves one websockets, so this floor
+    # and [inference]'s are deliberately the same number. Owned and measured by
+    # tests/test_websockets_floor_ships_the_imported_api.py.
+    "websockets>=13.0",
 ]
 # In-process Cosmos 3 backend (Cosmos3Policy(backend="diffusers")). Loads the
 # Cosmos3OmniPipeline directly via native Hugging Face `diffusers` (no RoboLab
@@ -469,7 +475,9 @@ ollama = [
 # (numpy-version agnostic, so it composes with lerobot's numpy>=2). See
 # strands_robots/inference/ and docs/inference/remote.md.
 inference = [
-    "websockets>=12.0",
+    # See [cosmos3-service]: >=13.0 is where `websockets.sync.server.Server`
+    # first ships, and both extras must declare the same floor.
+    "websockets>=13.0",
 ]
 # rosbridge WebSocket transport: drive ROS1 (or remote ROS 2) robots from a host
 # with no ROS install at all. Pure-pip `roslibpy` speaks the rosbridge JSON
--- a/tests/test_websockets_floor_ships_the_imported_api.py
+++ b/tests/test_websockets_floor_ships_the_imported_api.py
@@ -1,0 +1,268 @@
+"""The declared websockets floor must ship the websockets API the package imports.
+
+``strands_robots/inference/server.py`` annotates ``PolicyServer._server`` with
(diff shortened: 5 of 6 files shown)


yinsong1986 (approved):
## Summary

Raises the `websockets` floor in `[inference]` and `[cosmos3-service]` from `>=12.0` to `>=13.0`, because `strands_robots/inference/server.py` annotates `PolicyServer._server` with `websockets.sync.server.Server` — a name that first ships in 13.0 (12.0 spells it `WebSocketServer`). Since the import is under `TYPE_CHECKING`, a 12.0 install runs fine but fails the type gate with no source-side fix possible, so the old specifier admitted a measurably broken environment. The new `tests/test_websockets_floor_ships_the_imported_api.py` owns the measurement in one table, derives the floor from it (never restating the number), and self-maintains via an AST walk over the shipped sources that covers `TYPE_CHECKING` imports, dotted uses, and aliased submodule imports. `uv.lock` moves only the three `requires-dist` specifiers (the locked resolve at 16.0 already satisfies the new floor).

## What's good

- I independently verified the central claim against released wheels: `websockets.sync.server.Server` is absent in 12.0 (`WebSocketServer` present) and present in 13.0, along with the other table rows spot-checked.
- Ran the new test file: 16/16 pass under websockets ≥13.0; under a deliberately mis-pinned 12.0 environment exactly the 2 predicted guards fail, naming the specifier and the missing name — so the suite is not a tautology.
- The AST walk's maximal-attribute-chain and alias handling correctly captures all three import shapes actually present in the tree (`inference/server.py`, `inference/client.py`, `cosmos3/client.py`, `vera/client.py`, `reachy_transport.py`), and the `_MINIMUM_IMPORTED_NAMES` premise check keeps the audit from going silently vacuous.
- Follows repo conventions: changelog fragment per AGENTS.md step 3, `Closes #2339` in the body (not the title) per step 6, floors deliberately identical across extras with the rationale documented in `pyproject.toml` and both docs pages, and no upper cap added (correctly scoped out).

## Verification suggestions

- `pip install 'websockets==12.0'` in a scratch venv, then `pytest tests/test_websockets_floor_ships_the_imported_api.py --no-cov` — expect exactly 2 failures (`test_the_installed_websockets_satisfies_the_declared_floor` and `test_name_resolves[websockets.sync.server-Server]`); with `websockets>=13.0` expect 16 passed.

PR #2344, opened 2026-08-16
fix(simulation): resolve joint predicates against the whole scene

cagataycali:
## Problem

`joint_above`, `joint_below` and `joint_progress` resolve their `joint` through
`_joint_position`, which read `get_observation()` with no `robot_name`. That
observation is **robot-scoped**: called unscoped it reports one robot's joints.

An articulated fixture -- a drawer slide, a door hinge -- is necessarily a
*second* entity alongside the arm being controlled, so its joint was never in
that dict, under any spelling. Measured on a scene with an SO-101 and a cabinet
whose drawer joint sits at **43.97 mm**:

| lookup | result |
|---|---|
| compiled model joints | `so101/1..6`, `cabinet/drawer` |
| `get_observation(robot_name="cabinet")` | `{"drawer": 0.043974, ...}` |
| `get_observation()` | `{"1": ..., ..., "6": ...}` -- the arm only |
| `joint_above(joint="drawer")` / `"cabinet/drawer"` / `"cabinet_drawer"` | `False`, `False`, `False` |

So the term degraded to a constant, and a constant makes a predicate **and its
negation** both answer `False`. At 43.97 mm, `joint_above(0.005)` was `False`
*and* `joint_below(0.005)` was `False`, so neither an "opened the drawer" nor a
"closed the drawer" criterion could ever hold. `joint_progress` returned
`+0.000` -- and since it is a *negative* absolute distance, `0.0` is its
maximum, so a dead term is indistinguishable from being exactly on target.

`joint_progress`'s own docstring says it is "useful for drawer/door tasks where
success is 'joint near target position'". That was the one task class it could
not serve.

The existing unresolved-name warning fired, but its advice -- "Check the name
against the loaded scene / benchmark spec" -- could not be followed, because no
name worked, and its claim that the joint "is not present in the simulation" was
untrue.

## Change

Resolve the joint over the scene, mirroring the bounded ladder
`_body_position` already uses for the LIBERO `<name>_main` root-body convention:

1. the unscoped observation first, so single-robot scenes and the controlled
   robot's own joints keep their existing fast path (unchanged call, unchanged
   precedence when a name is shared);
2. then each robot `list_robots()` reports, by name, through the same
   `robot_name` route the `base_*` helpers in this module already use --
   `_joint_position` was the only `get_observation` call in the file without it;
3. a name several entities expose resolves against the first in
   `list_robots()`'s documented order (already the library's default-robot
   resolution order) and logs which one answered, so an under-specified spec is
   deterministic rather than silent;
4. only once every entity has been asked is the name genuinely absent -- which
   is what lets the existing warning say so truthfully and list what it tried.

A backend that produces no observation at all is still a capability gap rather
than a spec typo, so it stays silent, exactly as before.

## Tests

`tests/simulation/mujoco/test_scene_joint_predicates.py` drives the real MuJoCo
backend with an arm plus an articulated cabinet, and leads with a premise test
pinning the observation structure that makes the resolution necessary, so the
file cannot quietly stop measuring what it claims to.

`TestJointPredicatesResolveSceneJoints` adds a `_ScopedJointObsSim` stub that
reproduces the robot-scoped shape. The existing `_JointObsSim` returns the same
flat dict for every call, so it cannot express a scene with two entities -- which
is why the current joint tests pass against a structure the engine never
produces for a drawer, while the sibling body tests use `_BodyStateSim` and are
faithful.

**12 fail / 7 pass on `main`; 19 pass after.** The 7 that pass on both bound the
change: the premise test, the arm's own joint still resolving, the controlled
robot keeping its unscoped fast path, the default robot still winning a name it
shares, and a backend without introspection staying silent.

## Rollout

An SO-101 grasps the handle and pulls; the drawer joint travels
`0.000 -> 43.97 mm` (`d.qpos`, trace `0 -> 10.7 -> 29.1 -> 43.4`). The same
criterion is then evaluated on that same final state by each revision.

rollout

before and after

| criterion on the final state (drawer at 43.97 mm) | before | after |
|---|---|---|
| `joint_above(joint="open", value=0.005)` "was opened" | `False` | `True` |
| `joint_below(joint="open", value=0.005)` "is shut" | `False` | `False` |
| `joint_progress(joint="open", target=0.10)` reward | `+0.000` | `-0.560` |

## Gate

tests_integ`; `mypy` 14 errors on the branch and 14 on `main`, identical set,
all in `examples/isaac_gs/`. Full `tests`: **6 failed / 30677 passed** on the
branch against 19 failed / 30664 passed on `main` carrying the new test files.
Diffed on sorted test ids there are no new failures -- the 7 shared ones are
environment-bound (docker-absent vera, the GL probe, newton, mesh node-id) and
the single id that differs is in `tests/mesh/test_zenoh_transport_security.py`,
whose failures rotate under two parallel suites and which passes 9/9 three times
running in isolation.

### DIFF
--- a/changelog.d/2344-joint-predicates-resolve-scene-joints.md
+++ b/changelog.d/2344-joint-predicates-resolve-scene-joints.md
@@ -1,0 +1,10 @@
+### Fixed
+
+`joint_above`, `joint_below` and `joint_progress` now resolve their joint
+anywhere in the scene, not only on the robot the unscoped `get_observation`
+happens to report. An articulated fixture -- a drawer slide, a door hinge -- is a
+second entity alongside the arm being controlled, so its joint was previously
+invisible under every spelling and the term degraded to a constant: the
+predicate and its negation both answered `False`, and `joint_progress` returned
+`0.0`, the maximum of a negative-distance reward. Drawer and door tasks, the case
+`joint_progress` documents itself for, could not be scored.
--- a/strands_robots/simulation/predicates.py
+++ b/strands_robots/simulation/predicates.py
@@ -212,26 +212,88 @@ def _body_position(sim: SimEngine, body: str) -> list[float] | None:
 def _joint_position(sim: SimEngine, joint: str) -> float | None:
     """Best-effort joint-position lookup via ``get_observation``.
 
-    ``get_observation`` is on the ABC and returns ``{<joint_name>: float}``.
-    When the joint is absent from the observation dict (wrong robot, wrong
-    namespace) we return ``None`` so predicates can decide between ``False``
-    and an explicit error path.
+    ``get_observation`` is on the ABC and returns ``{<joint_name>: float}``,
+    but it is *robot-scoped*: called without ``robot_name`` it reports the
+    joints of one robot only. A benchmark spec names joints at *scene* scope,
+    and the joints that make ``joint_above`` / ``joint_below`` /
+    ``joint_progress`` worth having - a drawer slide, a door hinge - belong to
+    an articulated fixture that is necessarily a *second* entity alongside the
+    arm being controlled. Reading only the unscoped observation therefore never
+    resolves them, and the term degrades to a constant: the predicate *and its
+    negation* both answer ``False``, so no success criterion over that joint
+    can ever hold.
+
+    So resolve over the scene, mirroring the bounded ladder
+    :func:`_body_position` already uses for the LIBERO body-name convention:
+    the unscoped observation first (single-robot scenes, and the controlled
+    robot's own joints, keep their existing fast path), then each robot the
+    world reports by name via the same ``robot_name`` route the ``base_*``
+    helpers use. Only once every entity has been asked is the name genuinely
+    absent - which is what lets :func:`_warn_unresolved` say so truthfully and
+    list what was tried.
+
+    Returns ``None`` when no entity in the scene exposes ``joint``, so
+    predicates can decide between ``False`` and an explicit error path.
     """
+
+    saw_observation = False
+
+    def _read(robot: str | None) -> float | None:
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Fixes the constant-`False` degradation of `joint_above` / `joint_below` / `joint_progress` for joints on articulated fixtures (drawer slides, door hinges): `_joint_position` read only the unscoped, robot-scoped `get_observation`, so a second entity's joint was unreachable under any spelling and a predicate and its negation both answered `False`. The change resolves the joint over the scene with a bounded ladder mirroring `_body_position` — unscoped observation first (preserving the controlled robot's fast path and name precedence), then each `list_robots()` entry via the same `robot_name` route the `base_*` helpers already use — and only warns `_warn_unresolved` once every entity has been asked, now with a truthful tried-list. All touched helpers are private, both `list_robots()` and `get_observation(robot_name=...)` are on the `SimEngine` ABC, missing capabilities degrade silently through the file's established defensive pattern, and the shared-name precedence is unchanged, so this is strictly a bug fix with no public-API or behavioral one-way doors.

## What's good

- Changelog fragment present (`changelog.d/2344-joint-predicates-resolve-scene-joints.md`), per AGENTS.md step 3.
- The real-backend test file leads with a premise test pinning the robot-scoped observation structure, so the suite cannot quietly stop measuring the condition that makes the resolution necessary.
- The stub-based tests bound the fix on both sides: the controlled robot keeps its unscoped fast path, a shared name still resolves to the default robot, a genuinely absent joint still degrades-and-warns once, and a backend without introspection stays silent.
- Ambiguous multi-entity resolution is deterministic (documented `list_robots()` order) and logged rather than silent.

PR #2345, opened 2026-08-16
fix(deps): raise the eclipse-zenoh floor to the release that ships wire source attribution

cagataycali:
## Problem

The mesh safety handlers (`strands/safety/estop`, `strands/safety/resume`) authenticate a publisher at the **wire** level, below the JSON body an attacker can mutate. Two zenoh names carry that:

| half | name | used by |
|---|---|---|
| publisher | `zenoh.SourceInfo` | `Mesh._publish_safety_envelope` attaches `SourceInfo(publisher.id, sn)` |
| receiver | `Sample.source_info` | `_extract_sample_source_zid` reads `sample.source_info.source_id.zid` |

Both first ship in **eclipse-zenoh 1.6.1**. The declared floor read `eclipse-zenoh>=1.0.0,<2.0.0`, so **ten of the audited releases (1.0.0 through 1.5.1)** satisfied packaging while exposing neither name.

Nothing raises there. The publisher side is guarded by `hasattr(zenoh, "SourceInfo")` and the receiver side by `getattr(sample, "source_info", None)`; both answer "no attribution available" and continue. There is no exception, no log line, and no metric, so the only two observables are behavioural.

**1. On an all-old fleet the defence is simply off.** Every safety envelope goes out unattributed, so the cross-session forgery defence the handlers document is not in effect and an operator has no way to tell.

**2. In a mixed fleet the safety path fails closed on the wrong peer.** Measured with two real peers on one localhost mesh, driving the shipped publish and receive code:

| publisher | receiver | wire zid at receiver | body `source_zid` | e-stop lockout engages |
|---|---|---|---|---|
| 1.6.1 | 1.6.1 | present, matches body | yes | **yes** |
| 1.6.1 | 1.5.1 | absent | yes | **no - refused** |
| 1.5.1 | 1.6.1 | absent | no | yes (body-only defences) |

In the second row the publisher does everything right. The receiver's zenoh has no `Sample.source_info` descriptor at all, so `wire_zid` is `None` while `body_zid` is present, `_on_safety_estop` takes its

```
refusing remote estop -- body source_zid present but wire source_zid absent
(publisher misconfigured or attacker stripped SourceInfo)
```

branch, and **a fleet-wide emergency stop does not stop that robot** - while the diagnostic blames the publisher. Both peers satisfied the declared range, and this is the natural rolling-upgrade direction.

## Change

- `[mesh]` floors at `eclipse-zenoh>=1.6.1` (cap unchanged at `<2.0.0`). `uv lock` refreshed: three `requires-dist` specifiers, resolved version unchanged at 1.9.0.
- `tests/test_zenoh_floor_ships_wire_source_attribution.py` owns the measurement. `_ZENOH_ATTRIBUTE_FLOORS` maps each `zenoh.<attr>` the mesh reaches for to the release that first ships it, `_SAMPLE_SOURCE_INFO_FLOOR` records the receiver half (read off a sample object, so no AST sweep can find it), and `_PROBED_RELEASES` lists the wheels each entry was measured against so a value cannot be taken from a changelog on trust. The required floor is derived as a `max()` rather than restated.
- The three docstrings and one comment that described a zenoh without `SourceInfo` as a "legacy publisher" / "very old build" now name 1.6.1 and the declared floor. The `hasattr` / `getattr` fallbacks stay - they still cover the non-Zenoh transports and a mocked zenoh in tests - but they are no longer reachable from inside the supported range.
- `docs/mesh.md` states the requirement and why (upgrade every peer together); the installation extras matrix states the pin, matching the `[sim]` and `[lerobot]` rows.

Only the packaging metadata and prose change - no runtime behaviour is altered on a zenoh at or above the new floor.

## Tests

`tests/test_zenoh_floor_ships_wire_source_attribution.py`, 12 tests. Proven in both directions:

- **Pre-fix tree, current zenoh: 1 failed / 11 passed.** The packaging assertion names the offender: `eclipse-zenoh floor must be >= 1.6.1 because the mesh needs zenoh.SourceInfo and Sample.source_info for the safety handlers' wire source attribution; these specifiers admit releases that expose neither: {'optional-dependencies.mesh': 'eclipse-zenoh<2.0.0,>=1.0.0'}`
- **Fixed tree, a real eclipse-zenoh 1.5.1 wheel: 2 failed / 10 passed.** The installed-dist assertions are not tautologies - they fire on `zenoh.SourceInfo` and on `zenoh.Sample.source_info`.

`TestTheFloorIsSelfMaintaining` fails the moment the mesh reaches for a zenoh name with no recorded release, so a future API cannot silently leave the floor behind, and it carries a premise assertion so a refactor that stops using zenoh fails loudly instead of reporting clean.

Local gate: `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` at the pre-existing baseline; `tests/mesh` 45 pre-existing failing ids on this branch and on `main`, `comm` empty both ways.

## Measurement

Capability ladder from the released wheels, and the four-peer matrix run end to end through the shipped code (both panels are real program output; only the installed zenoh differs).

eclipse-zenoh floor vs mesh wire source attribution


yinsong1986 (dismissed):
## Summary

Raises the `[mesh]` extra's `eclipse-zenoh` floor from `>=1.0.0` to `>=1.6.1` (cap unchanged at `<2.0.0`), because the mesh safety handlers' wire-level source attribution depends on `zenoh.SourceInfo` (publisher side, `Mesh._publish_safety_envelope`) and `Sample.source_info` (receiver side, `_extract_sample_source_zid`), both of which first ship in 1.6.1. Below that floor the `hasattr`/`getattr` capability probes silently disable the cross-session forgery defence, and in a mixed fleet a 1.5.1 receiver refuses an e-stop from a correctly-attributing 1.6.1 publisher -- a fail-closed on the wrong peer. The runtime diff in `strands_robots/mesh/core.py` is docstring/comment-only (verified against the head patch); no behaviour changes on a zenoh at or above the new floor. The new guard test derives the required floor as a `max()` over a measured attribute-to-first-release table, sweeps the shipped sources (both direct attribute access and `hasattr`/`getattr` string probes) so a future zenoh name cannot leave the floor behind, pins the hand-recorded receiver-side attribute against the actual source of `_extract_sample_source_zid`, and carries an anti-vacuity minimum so a refactor away from zenoh fails loudly.

## What's good

- Floor raise on a directly declared dep lands in `[project.optional-dependencies]` with the rationale comment pointing at the test that owns the measurement, matching the repo's dependency-bounds conventions.
- Changelog fragment (`changelog.d/2345-zenoh-wire-attribution-floor.md`) rather than a direct `CHANGELOG.md` edit, per AGENTS.md step 3.
- `uv.lock` refreshed consistently: all three `requires-dist` markers (`all`, `mesh`, `mesh-iot`) carry the new specifier; resolved version unchanged at 1.9.0, so no environment churn.
- The guard test is proven in both directions per the PR description (fails on the pre-fix tree, and its installed-dist assertions fire on a real 1.5.1 wheel), and it self-maintains in both directions -- unrecorded new names and stale table entries both fail.
- Docs updated in both places users look (`docs/mesh.md` with the why, installation extras matrix with the pin), and the three docstrings that called the fallback a "legacy publisher" now name the actual floor.

## Verification suggestions

- `uv lock --check` to confirm the lockfile is in sync with the new specifier.
- `python -m pytest tests/test_zenoh_floor_ships_wire_source_attribution.py -q` on the branch (12 tests) -- and optionally revert the one-line `pyproject.toml` floor locally to confirm the packaging assertion names the offending specifier.

yinsong1986 (approved):
## Summary

Raises the `[mesh]` extra's `eclipse-zenoh` floor from `>=1.0.0` to `>=1.6.1` (cap unchanged at `<2.0.0`), because the mesh safety handlers' wire-level source attribution -- `zenoh.SourceInfo` on the publisher (`Mesh._publish_safety_envelope`) and `Sample.source_info` on the receiver (`_extract_sample_source_zid`) -- first ships in 1.6.1. Below the old floor the `hasattr`/`getattr` probes silently disable the cross-session forgery defence, and in a mixed fleet a 1.5.1 receiver refuses an e-stop from a correctly-attributing 1.6.1 publisher while the diagnostic blames the publisher. The `strands_robots/mesh/core.py` diff is docstring/comment-only (verified against the head patch), so no runtime behaviour changes on a zenoh at or above the new floor. The new guard test derives the required floor as a `max()` over a measured attribute-to-first-release table, sweeps the shipped sources for both direct attribute access and string-named `hasattr`/`getattr` probes so a future zenoh name cannot leave the floor behind, requires every recorded release to appear in `_PROBED_RELEASES` so a value cannot be taken from a changelog on trust, pins the hand-recorded receiver-side attribute against the source of `_extract_sample_source_zid`, and carries an anti-vacuity minimum. This head is the previously-reviewed tree plus a clean merge of `main` (conflict only in the `installation.md` extras table, resolved consistently with the incoming `websockets>=13.0` row); `uv.lock` updates all three `requires-dist` markers with the resolved version unchanged at 1.9.0, and the changelog lands as a fragment per AGENTS.md step 3. No blocking concerns found.

## Verification suggestions

- `uv lock --check` to confirm the lockfile matches the merged manifest (the branch merged `main`'s websockets floor change after the last lock refresh).
- `python -m pytest tests/test_zenoh_floor_ships_wire_source_attribution.py -q` (12 tests); optionally revert the one-line `pyproject.toml` floor locally to confirm the packaging assertion names the offending specifier.

PR #2346, opened 2026-08-16
fix(training): refuse a val_episodes whose episode count cannot be read

cagataycali:
## Problem

The LeRobot backend turns `val_episodes` into lerobot's `dataset.eval_split` **fraction**, and that conversion needs the dataset's `total_episodes` from a local `meta/info.json`. When the count is unreadable, `_val_eval_split` returned `None` and nothing was emitted — and an absent `eval_split` is indistinguishable from "no validation was asked for". So the run trained on **every** episode, logged **no** validation loss, and `validate()` reported **no problem at all**.

Because `Trainer.train` fails closed on `validate()`, "no problem" means the run launches. `train_policy(action="validate", ...)` — the preflight whose whole job is to say the spec will run as specified — answered `"[lerobot_local] spec is valid and launchable."`

Two reachable sources hit this, both fully supported data sources:

- `dataset_repo_id` with no `dataset_root` (a Hub dataset)
- a `dataset_root` that is a Hub cache directory nothing has been downloaded into yet — which the docs present as the way to combine the two

The capability itself is not missing: lerobot's `make_train_eval_datasets` calls `make_dataset(cfg)`, which builds an ordinary `LeRobotDataset` from the HF cache for a Hub `repo_id`, so the split works there. Adding a populated `dataset_root` to the identical spec emits the pair correctly. Only strands' derivation gave up.

This is the same outcome `validation_episodes_problems` already refuses a non-positive count to prevent, in its own words: *"The caller asked for a validation set and got a run without one."* The decision was enforced for one cause and not the other. `tests/tools/test_val_episodes_yields_validation_loss.py` opens by pinning that "every surface taking `val_episodes` emits the pair … and that a count which cannot be honored is refused rather than silently rounded" — the Hub source was the untested branch.

Measured on one 3-episode dataset, `val_episodes=1`, two real `lerobot-train` runs:

| | before | after |
|---|---|---|
| `validate()` | `[]` | 1 problem, 2 working remedies |
| emitted `eval_split` / `eval_steps` | none | `0.33` / `6` (after remedy) |
| `Train/eval split` logged by lerobot | never | `2 train, 1 eval` |
| episodes trained on (of 3) | **3** | **2** |
| validation losses logged | **0** | `eval_loss=0.7205` |
| run launched | yes | only after a remedy |

## Change

`LerobotTrainer.validate` now refuses a `val_episodes` whose episode count it cannot read, instead of dropping it. The refusal names both remedies that this backend actually honors — point `dataset_root` at a populated local copy, or pass lerobot's own knobs via `extra={"dataset.eval_split": ..., "eval_steps": ...}` — and the guard is keyed on the same reader `_val_eval_split` uses, so the check and the emission cannot drift.

The `>= total_episodes` bound and the multi-task refusal were also gated behind `dataset_root`; they now sit on the branch where the count is readable, which is the only branch where either is meaningful.

Second half, same contract: `validation_split_error`'s remedy hardcoded `extra_flags={...}`. That is right for the `lerobot_train` tool and wrong for `TrainSpec` / `train_policy`, whose field is `extra` — applied verbatim it raises `TypeError: unexpected keyword argument 'extra_flags'`. The helper now takes the caller's own spelling. It is **required**, not defaulted, precisely so a new surface cannot inherit a keyword it does not accept.

`self.provider_name` replaces a hardcoded `"LeRobotTrainer"` context string, which is what `validation_episodes_problems` documents the argument to be ("the backend's `provider_name`, so a problem names the backend that refused the value") and what the sibling refusal on the same surface already used.

## Tests

Added to the module that already owns this contract. **12 failed / 27 passed before, 39 passed after.** The headline failure states the finding directly:

```
AssertionError: a Hub source drops val_episodes silently: validate() reported [] while build_command() emits []
```

and the load-bearing one fails with `assert 'success' != 'success'` — before, `train()` returned success for a spec whose validation request it dropped.

- `TestAnUnreadableEpisodeCountIsRefused` — both reachable sources, plus that the refusal actually stops `train()`.
- `TestARefusalNamesAPassthroughItsOwnSurfaceAccepts` — the keyword is parsed **out of the message** and asserted to be a real `TrainSpec` field / a real `lerobot_train` parameter, then applied verbatim and required to produce the pair. This pins the remedy rather than the wording.
- `TestTheRefusalDoesNotReachPastTheDefect` — the 27 that pass on both revisions are the controls: a local root still emits the pair, a Hub source with `val_episodes=None` is still not refused (`None` is the documented "train on everything" sentinel), and a non-positive count still reports the count-domain problem first.

## Docs

`docs/training/overview.md` described this as a no-op only "when streaming a Hub dataset with no local cache". It is not streaming-specific and it also fires on an unpopulated cache root, so the paragraph now states the refusal, the reason the count is required, and both remedies. The `val_episodes` row gains the source requirement, and `TrainSpec.val_episodes` / `train_policy`'s docstring say the bound is a source requirement as well as a domain.

## Note for the breaking-change check

griffe reports exactly one change — `validation_split_error(passthrough_param)` added as required. It is deliberate: a default would name a keyword one of the two calling surfaces does not accept, which is the dead-end this PR removes. All four in-tree call sites are updated.

## Verification

Rollout of the two runs above (same request, same dataset, real `lerobot-train`):

val_episodes on a Hub source, before and after

Local gate: `ruff check` / `ruff format --check` / `mypy` over `strands_robots tests tests_integ` clean (mypy 14 findings, identical to `main`). Full `tests` suite run on this branch and on `main`: 7 failed / 2 errors on both, identical sets apart from four rotating ids in `tests/mesh/test_zenoh_transport_security.py`, which passes 9/9 three times in isolation on both trees. `+10 passed` on the branch is exactly the ten tests added here.

### DIFF
--- a/changelog.d/2346-val-episodes-unreadable-episode-count.md
+++ b/changelog.d/2346-val-episodes-unreadable-episode-count.md
@@ -1,0 +1,5 @@
+### Fixed
+
+`val_episodes` is no longer silently dropped when the LeRobot backend cannot read the dataset's episode count. The count becomes lerobot's `dataset.eval_split` fraction, so without it no split was emitted — and an absent split is indistinguishable from "no validation was asked for", leaving a run that trained on every episode, logged no validation loss, and reported no problem. `validate()` now refuses, naming the two remedies the backend honors: point `dataset_root` at a populated local copy, or pass `extra={"dataset.eval_split": ..., "eval_steps": ...}`. This affected a Hub `dataset_repo_id` with no `dataset_root`, and a `dataset_root` that is a Hub cache directory nothing had been downloaded into yet.
+
+Migration: `validation_split_error` takes a required keyword-only `passthrough_param` naming the caller's own raw-flag parameter (`extra_flags` on the `lerobot_train` tool, `extra` on `TrainSpec` / `train_policy`). It was previously hardcoded to `extra_flags`, which raised `TypeError` when applied verbatim on the `TrainSpec` surface; the argument is required rather than defaulted so a new surface cannot inherit a keyword it does not accept.
--- a/docs/training/overview.md
+++ b/docs/training/overview.md
@@ -96,7 +96,7 @@ supports and **ignores the rest** (the same tolerance rule as
 | `steps` / `global_batch_size` | the run size: optimizer steps x batch | each must be a positive integer; `validate()` refuses `0`, a fractional or non-finite value, and a `bool` (`True` would read as a silent one-step run) before anything is loaded |
 | `method` | `full` \| `lora` \| `expert_only` \| `frozen_backbone` | `lora`+`expert_only` are mutually exclusive |
 | `tune` | `{llm,visual,projector,diffusion}` | GR00T only |
-| `val_episodes` | hold out the LAST N episodes | deterministic split; must be a positive integer below the dataset's episode count. `validate()` refuses `0` or a negative (they produced no split and no eval cadence at all - the run trained on everything and logged no validation loss), a `bool`, and a fractional value (`2.7` reserved 3 episodes, `0.5` reserved none while still evaluating) |
+| `val_episodes` | hold out the LAST N episodes | deterministic split; must be a positive integer below the dataset's episode count, and that count must be readable from a local `meta/info.json` (see the Hub-source note below). `validate()` refuses `0` or a negative (they produced no split and no eval cadence at all - the run trained on everything and logged no validation loss), a `bool`, and a fractional value (`2.7` reserved 3 episodes, `0.5` reserved none while still evaluating) |
 | `num_gpus` / `num_nodes` | multi-GPU / multi-node | selects the launcher |
 | `seed` | reproducibility seed | must be a non-negative integer; `validate()` refuses a negative (`torch.manual_seed` would take it modulo `2**64`, so `-1` silently becomes `2**64 - 1`), a fractional or non-finite value, and a `bool`. `None` uses the backend's own default |
 | `extra["policy_type"]` | lerobot `--policy.type` | act/diffusion/smolvla/pi0/pi05/... |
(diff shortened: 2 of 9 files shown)


yinsong1986 (approved):
## Summary

This PR closes a silent-drop path in the LeRobot backend: a `val_episodes` request on a Hub source (or an unpopulated cache root) produced no `dataset.eval_split`, and since an absent split is indistinguishable from "no validation asked for", `validate()` reported nothing and the run trained on every episode with no validation loss. `LerobotTrainer.validate` now refuses when the episode count is unreadable, naming the two remedies the backend actually honors, and the `>= total_episodes` / multi-task checks move onto the only branch where the count exists. The second half fixes `validation_split_error`'s remedy hardcoding `extra_flags=` (a `TypeError` when applied verbatim on the `TrainSpec`/`train_policy` surface) by making the caller name its own passthrough keyword — required rather than defaulted so a new surface cannot inherit a spelling it does not accept, which is a loud break (TypeError at the call site) with all four in-tree call sites updated and a migration note in the changelog fragment.

## What's good

- The fix is exactly the AGENTS.md "no silent defaults on error" rule: the previously-launchable spec now gets an explicit refusal with working remedies, not a coerced or dropped request.
- The refusal is keyed on the same reader (`_dataset_total_episodes`) that `_val_eval_split` uses for emission, so the guard and the emission cannot drift; the residual `None`-emission paths (`val_episodes >= total`, non-positive count) are each covered by their own refusal.
- Tests pin the remedy rather than the wording — the passthrough keyword is parsed out of the message and asserted to be a real field on the surface that emitted it, then applied verbatim and required to produce the pair. Both touched test modules pass (141/141) on the head SHA.
- Changelog fragment present, `provider_name` replaces the hardcoded context string per the helper's own contract, docs updated to match the new behavior.

PR #1974, opened 2026-08-05
docs(examples): the install lines upgrade a stale environment instead of skipping it

sundargthb:
## Problem

Every notebook install line was a bare `pip install "strands-robots[extras]"`. Extras do not make a requirement unsatisfied, so against an environment that already carried an older release pip reported `Requirement already satisfied` and upgraded nothing. Measured against a real 0.4.1 venv:

| install line | result |
|---|---|
| `pip install "strands-robots[sim-mujoco,lerobot]"` | `Requirement already satisfied ... (0.4.1)` |
| `pip install -U "strands-robots[sim-mujoco,lerobot]"` | `Would install strands-robots-0.5.0` |
| `pip install "strands-robots[sim-mujoco,lerobot]>=0.5.1"` | `Would install strands-robots-0.5.0` |

The reader then ran the notebook against 0.4.1, whose `StreamingDatasetReader.open` has neither a `repo_type` parameter nor `**kwargs` (16 named params, no var-keyword), so notebook 5's headline bucket read failed with:

```
TypeError: open() got an unexpected keyword argument 'repo_type'
```

That names the keyword rather than the stale install behind it. The `>= 0.5.1` prerequisite text added in #1956 was correct and had no way to take effect - nothing in the command could act on it.

## Change

All six notebooks, the notebooks index, and the README's bucket-sync section now pass `-U`.

A version floor on the requirement works equally well, since it makes the installed release genuinely unsatisfying. `test_notebook_install_lines_upgrade_a_stale_environment` accepts either form and rejects a line carrying neither, so the guidance cannot regress to a bare install.

The git/URL fallback keeps `-U` too: it re-resolves the ref on its own, but is only reinstalled when asked to upgrade.

## Testing

The new test caught notebook 5's optional Isaac step, which carried the same bare form and was not part of the reported finding.

82 passed, 2 skipped across the notebook-docs, changelog, LeRobot-dependency, floor-guarantee, and customer-workflow suites. `assemble_changelog.py --check` OK; ruff clean.

## Scope

Docs, one test, and a changelog fragment. No `strands_robots/` or `pyproject.toml` changes.

Reported in #1507 (round-2 blog review) as the last gap in the install path. The other two open items from that review are the external-stats cite-check (editorial, not repo-side) and Figure 4's timing, which is corroborated by the saved run log (`wall: 132.6s`). The hub-floor one-liner from the same review is already fixed in #1957, which measured the boundary at 1.5.0.

Refs #1507

Issue #1905, opened 2026-08-03
process: an approved PR whose last commit the sole reviewer pushed can never merge - #1722 and #1035 are deadlocked, and it presents as "reviewer bandwidth"


cagataycali:
## Summary

Two PRs (#1722, #1035) are `APPROVED`, `MERGEABLE`, have the required check green and every review thread resolved, and **still cannot be merged by anyone currently reviewing them**. The cause is not review quality, CI, or conflicts. It is that the maintainer who pushed their most recent commit is the same person whose approval is meant to satisfy the branch ruleset, and the ruleset forbids exactly that.

This has been silently misfiled as "reviewer bandwidth is the sole constraint" for weeks. It is not a bandwidth problem; more reviewing by the same reviewer cannot fix it.

## The rule

The active `default` BRANCH ruleset carries a `PULL_REQUEST` rule with:

```
requiredApprovingReviewCount: 1
requireLastPushApproval:      true    <-- the relevant one
requireCodeOwnerReview:       true
dismissStaleReviewsOnPush:    true
requiredReviewThreadResolution: true
```

`requireLastPushApproval: true` means the most recent push must be approved by someone **other than the person who pushed it**. So when a maintainer pushes a fix onto a contributor's branch to "take it over the line" and is also the only reviewer, their approval stops counting, `reviewDecision` reverts to `REVIEW_REQUIRED`, and `mergeStateStatus` sits at `BLOCKED` indefinitely.

## Evidence: a controlled comparison, same repo, same reviewer, same ruleset

The only variable that differs is who pushed the head commit.

| PR | head commit pushed by | approved by | `reviewDecision` | `mergeStateStatus` |
|---|---|---|---|---|
| #1722 | `cagataycali` (`3375c000`) | `cagataycali` | `REVIEW_REQUIRED` | `BLOCKED` |
| #1035 | `cagataycali` (`8d6a4c4`, a merge-from-main) | `cagataycali` | `REVIEW_REQUIRED` | `BLOCKED` |
| #1894 | `yinsong1986` (`28a21c8`) | `cagataycali` | **`APPROVED`** | **`CLEAN`** |

#1894 was approved during this session and flipped from `BLOCKED`/`REVIEW_REQUIRED` to `CLEAN`/`APPROVED` the instant the review landed, then merged normally (`e3e5e10`). Same reviewer, same ruleset, same required check. The difference is solely that its head commit was author-pushed.

[...]

PR #1035, opened 2026-07-03
feat(mesh): AckermannRosRobot - Ackermann ROS 2 cars (AWS DeepRacer) as strands robots

Vivek0712:
Closes #1034

## Why

`RosBridgedRobot` covers differential-drive bases: Twist on `cmd_vel`, `/odom` for pose. An Ackermann car has neither. The AWS DeepRacer - the most widely owned ROS 2 car - takes normalized `{angle, throttle}` (`deepracer_interfaces_pkg/msg/ServoCtrlMsg`) on `/webserver_pkg/manual_drive`, publishes no odometry, and must first be switched into manual mode by a two-step service handshake (`/ctrl_pkg/vehicle_state` with `state=1`, then `/ctrl_pkg/enable_state` with `is_active=true`). Today a strands agent cannot drive it at all. The same shape applies to F1TENTH and most RC-scale Ackermann platforms.

## What

New `strands_robots/mesh/ackermann_robot.py` (`AckermannRosRobot`, exported from `strands_robots.mesh`), mirroring `RosBridgedRobot` exactly: the class owns no ROS 2 state; every method forwards through `use_ros` (`publish` / `service_call` / `echo`), so it inherits the in-process rclpy backend and input validation, and is safe to construct without ROS 2.

- **Same fleet-wide drive contract.** `drive(linear, angular, duration)` is signature-identical to `RosBridgedRobot.drive`; a bicycle model (`atan(wheelbase * angular / linear)`, steering clamp, rest epsilon) converts to normalized servo values internally. The agent never learns a second vocabulary. The naive `atan2(wheelbase * angular, linear)` form is deliberately avoided - it lands in the wrong quadrant in reverse, commanding near-full opposite steering while backing up (caught by TDD, pinned by test).
- **Declarative enable handshake.** `init_services` is an ordered list of `{service, type, fields}` calls that put the vehicle into a commandable state. It runs once, automatically, before the first drive; a failure aborts the drive (never command a car that is not in manual mode) and does not latch, so a retry re-runs it. `stop()` is never gated.
- **`from_deepracer()`** preconfigures the full stock wiring (servo topic/type, RPLIDAR scan, manual-mode handshake, wheelbase 0.164 m, 1.5 m/s, 30 deg steering, 20 Hz); any keyword can be overridden for a modified car.
[...]

PR #2347, opened 2026-08-16
docs(agents): a null reviewDecision is at least two states, and the resolve clears one

### LINKED ISSUE #2343: AGENTS.md: `reviewDecision: null` does not always become APPROVED when the last thread is resolved
## The passage

AGENTS.md > PR Workflow > step 8 documents `reviewDecision: null` as a third reading meaning "one resolve from merging", and records #1974 where resolving the sole unresolved thread moved both fields at once:

| field | one unresolved thread | after `resolveReviewThread` |
|---|---|---|
| `mergeStateStatus` | `BLOCKED` | `CLEAN` |
| `reviewDecision` | `null` | `APPROVED` |

The surrounding prose leans on that: it contrasts `null` with `REVIEW_REQUIRED` on the grounds that the two "need opposite actions - the `REVIEW_REQUIRED` case above needs a second account, this one needs no review at all."

## The counterexample, measured on #2328

#2328 presented the documented signature exactly: `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED`, `reviewDecision: null`, one unresolved `github-advanced-security` thread, required check `call-test-lint / Test and Lint` = `SUCCESS`. Following the passage, I resolved the thread (after dismissing code-scanning alert 923 as a false positive). Reading the same fields back immediately after:

| field | before resolve | after resolve | documented expectation |
|---|---|---|---|
| `reviewThreads` | 1 unresolved | all resolved | resolved |
| `mergeStateStatus` | `BLOCKED` | **`BLOCKED`** | `CLEAN` |
| `reviewDecision` | `null` | **`REVIEW_REQUIRED`** | `APPROVED` |

So the resolve was necessary and correct, and it was not sufficient. The `null` was concealing a *second* gate rather than standing in for it.

## Why, and why the field cannot distinguish the two cases

`null` is what you get when the pull request has **no approving review at all**. #2328's two reviews are both `COMMENTED` (one from `github-advanced-security`, one a thread reply), and a `COMMENTED` review does not contribute an approval. #1974 differed in one input that the recorded table does not name: it already carried a current `APPROVED` review that post-dated its head commit, so resolving the thread removed the only remaining gate and the decision settled to `APPROVED`.

That makes `null` at least three states, not one:

| pull request | reviews present | unresolved thread | resolving the thread yields |
|---|---|---|---|
| #1974 | one `APPROVED` | yes | `APPROVED`, `CLEAN` - merges |
| #2328 | `COMMENTED` only | yes | `REVIEW_REQUIRED`, still `BLOCKED` |

This misreads in the reassuring direction, which is the property step 8 already flags for `null`: the documented reading says "needs no review at all", and acting on it here would have produced a resolve, a re-read expecting `CLEAN`, and a pull request reported as ready when it is in fact waiting on a first approving review. It is the same presentation #1905 records - work reported as blocked on nothing while the actual remedy is a second account.

## Suggested correction

The distinguishing read is the reviews, not the decision. `reviewThreads` separates "is a thread holding this" from "is an approval owed", and the review states separate the two `null` cases before any mutation:

```graphql
reviews(last: 20) { nodes { state author { login } submittedAt } }
reviewThreads(first: 50) { nodes { id isResolved isOutdated } }
```

A `null` with no `APPROVED` node is `REVIEW_REQUIRED` wearing a different value, and resolving threads will not move it. Worth stating in the step 8 table as a second row rather than replacing the #1974 one, since the #1974 case is real and the two are told apart only by the review set.

## Scope

Documentation-only. No behaviour is wrong; the resolve on #2328 was still the right action and its thread gate is genuinely cleared, with `call-test-lint` green. Raising it separately rather than folding it into an unrelated pull request, per the tracked-follow-up rule.

Context: the alert-923 dismissal and the mutation test backing it are in the #2328 thread ( https://github.com/strands-labs/robots/pull/2328#discussion_r3792603042 ).

---

<sub>This issue was filed by an AI contributor. The measurements above were taken by the agent.</sub>

cagataycali:
Closes #2343

## The passage prescribed an action that is right on one case and insufficient on the other

`AGENTS.md` > PR Workflow step 8 documents `reviewDecision: null` as a third reading meaning "one resolve from merging". It was recorded from #1974, where resolving the sole unresolved thread moved `mergeStateStatus` to `CLEAN` and the decision to `APPROVED`. That measurement is real. What the passage did not name is the input that made the resolve *sufficient*, and it went further - it told the reader this case "needs no review at all".

#2328 presented the documented signature exactly: `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED`, `reviewDecision: null`, one unresolved `github-advanced-security` thread, required check `call-test-lint / Test and Lint` = `SUCCESS`. Following the passage, the thread was resolved. Reading the same fields back:

| pull request | approving review present | after `resolveReviewThread` |
|---|---|---|
| #1974 | one `APPROVED`, post-dating the head | `APPROVED` / `CLEAN` - merges |
| #2328 | none - every review `COMMENTED` | `REVIEW_REQUIRED` / still `BLOCKED` |

`null` is also what a pull request carrying **no approving review at all** reads, because a `COMMENTED` review contributes no approval. So the resolve was necessary on both and sufficient on one, and the field is identical on the two beforehand. Re-verified on #2328's current head while preparing this change: `reviewDecision: REVIEW_REQUIRED`, three reviews all `COMMENTED`, its one thread resolved, `call-test-lint` `SUCCESS`.

This misreads in the reassuring direction, which is the property step 8 already flags for `null`: acting on it yields a resolve, a re-read expecting `CLEAN`, and a pull request reported as ready while it is in fact waiting on a first approving review - the presentation #1905 records for a different cause.

## Change

- Step 8's `null` bullet keeps the #1974 row and gains the #2328 one, so the two are told apart by the review set rather than by the decision.
- The distinguishing read becomes `reviews(last: 20)` beside `reviewThreads(first: 50)`, and it is **ordered before** the resolve. That ordering is the load-bearing half rather than a nicety: #2328's decision moved from `null` to `REVIEW_REQUIRED` on the resolve, so the one value identifying which case a reader is in is destroyed by the action the passage prescribes, and no re-read afterwards recovers it.
- The bullet now also notes that `REVIEW_REQUIRED` itself carries two remedies - a first approval, or a second account when the only approval came from the pusher - which is the split `scripts/check_last_push_approval.py --all-open` reports and which no single field distinguishes either. Measured on the open set today: 11 open non-draft pull requests, all reading `REVIEW_REQUIRED`, of which one (#1035) is `pusher-only-approval` and ten are `awaiting-first-review`.

The correction deliberately keeps the #1974 case rather than replacing it. Both are real and they are told apart only by the review set, so deleting either row would leave the next reader with a rule that is right by luck.

## Tests

`tests/test_review_decision_null_is_not_one_state.py` pins both halves, the shape `c56ab08` used for the merge-mutation bullet:

- **Arithmetic.** A verdict keyed on the decision and the threads alone predicts the resolve correctly on #1974 and incorrectly on #2328; one that also reads the review set is right on both. The pre-resolve fields of the two rows are asserted *equal*, since that indistinguishability is the premise the whole bullet rests on.
- **Prose.** Adjacency assertions positioned from the sentence the correction introduces, so its absence fails loudly rather than making the rest vacuous - the anchoring problem `tests/test_merge_gate_viewer_scope.py` documents.
- The strongest guard is `test_the_unqualified_claim_is_never_restored`: the phrase "needs no review at all" may still appear (the correction quotes it to say which case it holds for), but only within `_ADJACENCY_WINDOW` of the qualification naming #2328. Restoring the bare instruction fails.

Against the uncorrected `AGENTS.md`: **5 failed, 5 passed** - the four prose assertions plus the unqualified-claim guard, which names the original sentence. **10 passed** after. The five that pass on both revisions are the arithmetic, which pins the reasoning rather than the file.

## Verification

- Every test in the tree that reads `AGENTS.md` - 15 files, including `tests/test_merge_gate_viewer_scope.py`, whose adjacency window this insertion shifts, and both `test_source_no_review_history_markers.py` / `test_source_strings_no_review_archaeology.py`: **322 passed**.
- `tests/test_changelog_fragments.py` + `tests/test_changelog_format.py`: 30 passed; `scripts/assemble_changelog.py --check` OK (443 pending). `tests/test_no_host_paths.py`: 2 passed.
- `ruff check` and `ruff format --check` clean on the new file; `mypy` reports no issues in it.
- Non-ASCII sweep (`grep -nP '[^\x00-\x7F]'`) over both new files: empty.

## Scope

Documentation and test only. No behaviour changes, and the resolve performed on #2328 was still the right action - its thread gate is genuinely cleared. The pull request it was blocking needs a first approving review, which is exactly what the corrected passage now says.

---

## Round 0

Opened. No review feedback yet.

---

<sub>This pull request was authored by an AI contributor. The measurements above were taken by the agent.</sub>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -675,14 +675,42 @@ hatch run format            # ruff check --fix, ruff format
      direction - `REVIEW_REQUIRED` at least says a review is owed, whereas
      `null` reads as "no review requirement applies here" rather than "one
      resolve from merging". It is not a recompute lag: that approval was more
-     than twenty minutes old when the field was read as `null`. The two also
-     need opposite actions - the `REVIEW_REQUIRED` case above needs a second
-     account, this one needs no review at all. The distinguishing read is the
-     threads, not the decision:
+     than twenty minutes old when the field was read as `null`.
+
+     **`null` is at least two states, and the resolve clears only one of them.**
+     #2328 presented #1974's signature exactly - `MERGEABLE`, `BLOCKED`, `null`,
+     one unresolved `github-advanced-security` thread, the required check
+     `SUCCESS` - so the paragraph above prescribed resolving that thread, which
+     was the right action. It settled the decision the other way:
+
+     | pull request | approving review present | after `resolveReviewThread` |
+     |---|---|---|
+     | #1974 | one `APPROVED`, post-dating the head | `APPROVED` / `CLEAN` - merges |
+     | #2328 | none - every review `COMMENTED` | `REVIEW_REQUIRED` / still `BLOCKED` |
+
+     `null` is also what a pull request carrying **no approving review at all**
+     reads, because a `COMMENTED` review contributes no approval. So the resolve
+     was necessary on both and sufficient on one, and "this one needs no review
+     at all" is true of #1974 and false of #2328 while the field is identical on
+     the two.
+
+     Read the review set beside the threads, and read it **before** you resolve
+     anything. #2328's decision moved from `null` to `REVIEW_REQUIRED` on the
+     resolve, so the one value that says which case you are in is gone by the
+     time you re-read it:
 
      ```graphql
+     reviews(last: 20) { nodes { state author { login } submittedAt } }
      reviewThreads(first: 50) { nodes { id isResolved isOutdated } }
      ```
+
+     A `null` carrying no `APPROVED` node is `REVIEW_REQUIRED` wearing a
+     different value: the remedy is a first approving review, and no amount of
+     resolving supplies one. That leaves `REVIEW_REQUIRED` itself carrying two
+     remedies - a first approval, or a second account when the only approval
+     came from the pusher - which is the split
+     `scripts/check_last_push_approval.py --all-open` reports and which no
+     single field distinguishes either.
 
      `isOutdated: true` on an unresolved thread is the common form, and it is a
      prompt rather than reassurance: the diff moved on, so the request has
--- a/changelog.d/2343-review-decision-null-is-not-one-state.md
+++ b/changelog.d/2343-review-decision-null-is-not-one-state.md
@@ -1,0 +1,44 @@
+### Docs: a `null` reviewDecision is at least two states, and resolving the thread clears only one
+
+`AGENTS.md` > PR Workflow step 8 documented `reviewDecision: null` as a third
+reading meaning "one resolve from merging", recorded from #1974 where resolving
+the sole unresolved thread moved `mergeStateStatus` to `CLEAN` and the decision
+to `APPROVED`. That measurement is real, but the passage did not name the input
+that made the resolve *sufficient*, and it explicitly told the reader the case
+"needs no review at all".
+
+#2328 presented the same signature - `MERGEABLE`, `BLOCKED`, `null`, one
+unresolved `github-advanced-security` thread, `call-test-lint / Test and Lint`
+`SUCCESS` - and settled the other way. `null` is also what a pull request with
+no approving review at all reads, because a `COMMENTED` review contributes no
+approval:
+
+| pull request | approving review present | after `resolveReviewThread` |
+|---|---|---|
+| #1974 | one `APPROVED`, post-dating the head | `APPROVED` / `CLEAN` - merges |
+| #2328 | none - every review `COMMENTED` | `REVIEW_REQUIRED` / still `BLOCKED` |
+
+So the resolve was necessary on both and sufficient on one, and the two are
+byte-identical in the field before it. This misreads in the reassuring
+direction, which is the property step 8 already flags for `null`: following it
+yields a resolve, a re-read expecting `CLEAN`, and a pull request reported as
+ready while it waits on a first approving review - the presentation #1905
+records for another cause.
+
+Step 8 now carries the second row, prescribes reading the review set beside the
+threads, and orders that read *before* the resolve. The ordering is the
+load-bearing half: #2328's decision moved from `null` to `REVIEW_REQUIRED` on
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Documentation and test-only change to `AGENTS.md` > PR Workflow step 8: the `reviewDecision: null` bullet previously said the case "needs no review at all" and prescribed resolving the unresolved thread, which #2328 showed is right on #1974's input (an `APPROVED` review already present) and wrong when the review set carries only `COMMENTED` reviews. The correction keeps both measured rows, prescribes reading `reviews(last: 20)` beside `reviewThreads(first: 50)` *before* the resolve (the ordering matters because the resolve destroys the `null` that identifies the case), and notes the two remedies `REVIEW_REQUIRED` itself conflates. The new prose-pin test follows the established shape (`test_merge_gate_viewer_scope.py`-style adjacency assertions plus an arithmetic half over the recorded observations), and the guard that the unqualified "needs no review at all" claim cannot be restored without its #2328 qualification is the right teeth for this class of doc regression.

## What's good

- Fragment in `changelog.d/` per step 3, named per `changelog.d/README.md` (issue number is explicitly allowed).
- Verified locally on the PR head: the new test plus `tests/test_merge_gate_viewer_scope.py` pass (14 passed); the full set of tests referencing `AGENTS.md` shows zero delta against the base (the 3 failures present are pre-existing rendering/GPU environment failures, identical on base).
- Both new files are ASCII-clean and carry no host paths.
- Scope discipline: no production code touched; the correction keeps the #1974 row rather than replacing it, so neither measured case is lost.

PR #2349, opened 2026-08-16
fix(deps): bound the vera-sim imageio floor at the release that reads the clip duration in milliseconds

cagataycali:
## Problem

`[vera-sim]` declared a bare `imageio` - no lower bound, no cap - while `[sim-mujoco]` and `[sim-isaac]` both declared `imageio>=2.28.0,<3.0.0` for the same encoder. An environment resolves one imageio, so the unbounded range described an install in which `strands_robots.rendering.video.encode_clip` writes the wrong clip.

`encode_clip` passes the GIF writer a per-frame duration in **milliseconds**:

```python
# Pillow's GIF writer takes per-frame duration (ms), not fps.
imageio.mimsave(str(out), frame_list, duration=1000.0 / int(fps))
```

That is imageio's reading from 2.28.0 on. Earlier releases read the same number as **seconds**.

## Measured

Driving the shipped `encode_clip` over 12 rendered MuJoCo frames at `fps=12` - so 83.33 ms per frame is requested - against the released wheels on Python 3.12 (the project's minimum):

| release | `imageio.v2` | `encode_clip` | per-frame duration |
|---|---|---|---|
| 2.9.0, 2.15.0 | absent | `ModuleNotFoundError: No module named 'imageio.v2'` | no clip written |
| 2.16.0 | present | `AttributeError` out of `mimsave` | no clip written |
| 2.17.0 - 2.19.0 | present | `RecursionError` out of `mimsave` | no clip written |
| 2.20.0 - 2.27.0 | present | returns a valid 12-frame GIF | **83330 ms** |
| 2.28.0 - 2.37.4 | present | returns a valid 12-frame GIF | **80 ms** |

Neither failure mode is one the shipped guards can catch.

`require_optional` takes no minimum version (`module_name`, `pip_install`, `extra`, `purpose`, `system_install`), so on 2.9.0 it probes the top-level module, finds it, reports imageio installed - and the very next line raises a bare `ModuleNotFoundError` that names no remedy.

On 2.20.0 - 2.27.0 nothing raises at all. The duration is read as seconds, so the clip is encoded 1000x too slow and a one-second rollout plays for 16 minutes 40 seconds. Because the file is a valid 12-frame GIF, `encode_clip`'s own "the encoder wrote no clip" check passes and hands it back as a success.

The declared range is the only place either can be refused.

The encoder itself is already covered: the suite pins that an `encode_clip` GIF decodes to a per-frame duration matching the requested fps, which holds only from 2.28.0. What was missing is a floor that admits only the releases where that can hold. Nothing compared the two because `[vera-sim]` is declared in conflict with `[all]` (`[tool.uv] conflicts`) and CI installs `.[all,dev]`, so the resolve this extra describes is never built.

## Change

`[vera-sim]` states the same `imageio>=2.28.0,<3.0.0` as `[sim-mujoco]` and `[sim-isaac]`. The number is not new to the manifest and the resolved set does not move - `uv.lock` changes by one `requires-dist` line.

`tests/test_imageio_floor_honors_the_clip_frame_duration.py` pins it, mirroring the existing floor guards. `_IMAGEIO_SYMBOL_FLOORS` and `_GIF_DURATION_IN_MILLISECONDS_FROM` own the measurement and the required floor is derived from them rather than restated; an AST walk over the shipped sources fails the moment the package reaches for an imageio name with no recorded release, so a future use of a newer imageio API cannot leave the floor behind; and every recorded number must appear in `_PROBED_RELEASES`, so no row can come from a changelog instead of a wheel.

Scope: this corrects the declared range for the encoder the extra already ships. Which media capabilities `[vera-sim]` should offer is a separate question - it declares no `imageio-ffmpeg`, which the MP4 branch of the same function needs - and is left alone.

## Verification

| | result |
|---|---|
| Pre-fix, on `main` | **3 failed / 12 passed**, each naming `optional-dependencies.vera-sim` |
| Post-fix | 15 passed |
| Fixed tree, real imageio 2.27.0 wheel installed | 2 failed - the installed-dist checks are not tautologies |
| Fixed tree, real imageio 2.9.0 wheel installed | 2 failed, 3 skipped (the `imageio.v2` rows skip because the module does not exist) |

The three pre-fix failures are the three arguments: the floor is below the capability, the ranges disagree across extras that resolve one imageio, and the specifier admits a major nothing here measured. The 12 that pass on both trees are the controls - the AST walk, the table's self-maintenance, and the installed-name checks are unaffected by the change.

Full `tests/` on this branch and on `main`, same interpreter, run in parallel: **61** failing test ids on the branch vs **64** on `main` (30474 vs 30471 passed); `comm -23` on the sorted id sets is empty, so there are no new failures. The three extra ids on `main` are exactly the three pre-fix failures above - the new test file was copied into the base tree so the split could be measured at full-suite scope too. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` at the pre-existing baseline; `uv lock --check` clean.

## Artifact

The same rollout, the same `encode_clip`, the two imageio releases the old range admitted. Left panel: one frame is all that has appeared by t = 1.0 s. Right panel: all 12 play inside 1.0 s.

imageio clip duration floor

The two clips themselves, both written by `encode_clip` from the same frames. The first appears frozen because its first frame is held for 83 seconds:

| `imageio 2.27.0` - admitted by the bare specifier | `imageio 2.37.2` - what `>=2.28.0` admits |
|---|---|
|  |  |


yinsong1986 (approved):
## Summary

Tightens `[vera-sim]`'s bare `imageio` to `imageio>=2.28.0,<3.0.0`, matching the range `[sim-mujoco]`/`[sim-isaac]` already declare for the same `encode_clip` encoder. The floor is motivated by measured wheel behaviour: below 2.16.0 `imageio.v2` does not exist (a `ModuleNotFoundError` that `require_optional` cannot catch, since it takes no minimum version), 2.16.0-2.19.0 raise out of `mimsave`, and 2.20.0-2.27.0 silently read the per-frame `duration` as seconds instead of milliseconds, returning a valid-but-1000x-too-slow GIF that `encode_clip`'s own output check accepts. I verified the claims against `strands_robots/rendering/video.py` at this head: the `require_optional` probe, the `import imageio.v2`, and the ms-duration `mimsave` call are all as described. The accompanying `tests/test_imageio_floor_honors_the_clip_frame_duration.py` derives the required floor from a single measurement table, walks the shipped sources by AST so a future imageio API use can't leave the floor behind, and cross-checks every recorded release against `_PROBED_RELEASES`. The `uv.lock` delta is the single `requires-dist` line, the changelog fragment is present per AGENTS.md step 3, and the range is not new to the manifest - so the resolved set does not move and this is not a one-way door.

## What's good

- Floor parity asserted across every extra that declares imageio, not just the one being fixed - the disagreement class can't silently reappear.
- The test guards its own vacuity (`test_the_walk_reaches_the_sources`, `_MINIMUM_IMPORTED_NAMES`) and pins that the ms-duration behaviour is still the one `encode_clip` relies on, so a refactor of the encoder invalidates the constant loudly rather than silently.
- Changelog fragment in `changelog.d/`, no host paths, no non-ASCII in the new test, scope held to the declared range (the missing `imageio-ffmpeg` for the MP4 branch is explicitly deferred rather than smuggled in).

PR #2351, opened 2026-08-16
fix(training): refuse a dataset format version lerobot cannot read at validate() time

cagataycali:
## Problem

Every LeRobotDataset declares the format version it was written in as `codebase_version` in `meta/info.json`. lerobot compares it against its own `CODEBASE_VERSION` and refuses a dataset whose **major** is older (an older *minor* loads with a warning). `LerobotTrainer.validate()` never looked at it, so an outdated local root reported as launchable and failed at load time instead — and the failure it produced is a poor place to learn this:

| local root declares | what the caller is told |
| --- | --- |
| `v2.1` | `BackwardCompatibilityError` whose remedy is `python -m lerobot.scripts.convert_dataset_v21_to_v30 --repo-id=local` — and `local` is `_dataset_source`'s placeholder for *not a Hub dataset*, not a repo anyone can convert |
| anything else older (`v2.0`, `v1.6`, ...) | `NotImplementedError: Contact the maintainer on [Discord](...)` — raised from inside `BackwardCompatibilityError.__init__`, naming neither the dataset, nor the version, nor the problem |

Measured on lerobot 0.6.2, driving the shipped API (`validate()` returned `[]` in every case):

```
codebase_version='v2.1' -> train() error: lerobot train raised BackwardCompatibilityError: ... --repo-id=local
codebase_version='v2.0' -> train() error: lerobot train raised NotImplementedError: Contact the maintainer on [Discord](...)
codebase_version='v1.6' -> train() error: lerobot train raised NotImplementedError: Contact the maintainer on [Discord](...)
```

## Change

The version sits in the same `meta/info.json` this trainer already reads for the episode count (`_dataset_total_episodes`) and the quantile stats (`_quantile_stats_problems`), so the mismatch is decidable **offline** — no network, no download. `_dataset_codebase_version_problems()` reports it at spec-validation time, naming the root, both versions, and a remedy that exists:

```
dataset_root '/data/old_dataset' declares codebase_version 'v2.1' in meta/info.json, which
lerobot 0.6.2 cannot read (it loads 'v3.0' and refuses an older major). Convert it with
lerobot's own converter: python -m lerobot.scripts.convert_dataset_v21_to_v30 --repo-id=<your-dataset-repo-id>
```

lerobot ships exactly one conversion (`v2.1` -> `v3.0`), so an older root is told that plainly rather than handed a command that would fail. The repo id is substituted with the same `<your-dataset-repo-id>` placeholder `_quantile_stats_problems` already uses, so the advice never passes the `local` sentinel to a converter.

This mirrors `_quantile_stats_problems`, whose docstring states the same job for the dataset's stats: lift a train-time dataset-metadata failure "to spec-validation time with an actionable message naming lerobot's remedy". It is deliberately as conservative as that sibling:

- only a **definite** mismatch is flagged — a local `meta/info.json` declaring an older major;
- a Hub dataset with no materialized local cache is left unflagged. `validate()` does not reach the network (the reward-model path says so to users: *"the spec itself is fine - validate() cannot reach the network to see this"*), so Hub-side metadata stays lerobot's to verify at load time;
- an unparseable version on either side fails **open**, the same posture as `_huggingface_hub_version_error`;
- only the major is compared, because an older minor loads.

It is wired into `validate()` rather than `_validate_policy` because policy and reward-model runs load the same dataset, so an unreadable format version refuses either one. The lerobot-side constant is read from `lerobot.datasets.dataset_metadata` — the module whose `_load_metadata` enforces it — with a documented static fallback so the offline verdict still works when lerobot is absent.

## Tests

`tests/training/test_dataset_codebase_version_preflight.py`, 16 tests. On `upstream/main`: **9 failed / 7 passed**; with the change: **16 passed**.

The 5 behaviour failures pre-fix are the headline (each reports `got []`):

```
test_a_v21_root_is_refused_and_pointed_at_lerobots_converter          AssertionError: ... got []
test_an_older_root_is_refused_without_claiming_a_converter_for_it[v2.0] AssertionError: ... got []
test_an_older_root_is_refused_without_claiming_a_converter_for_it[v1.6] AssertionError: ... got []
test_the_remedy_never_passes_the_local_sentinel_as_a_repo_id          AssertionError: []
test_a_reward_model_run_is_gated_too                                  AssertionError: reward-model runs load the same dataset
```

The **7 tests that pass on both trees** are the no-overreach controls, and they are the point: a current or newer root (`v3.0`, `v3.1`, `v4.0`) is clean, an absent/unparseable/empty version fails open, and a Hub spec with no local cache is not flagged. One test also pins the characterization of lerobot this change depends on — that `BackwardCompatibilityError` builds a message for exactly `2.1` and raises `NotImplementedError` for every other older major — so the branch cannot drift from the library it describes.

## Artifact

The same script and the same spec run against `upstream/main` and against this change; only the tree on `PYTHONPATH` differs. The `v2.0` "before" panel is the sparsest because it is the one where the caller is told nothing at all.

validate-time refusal of an unreadable dataset format version

## Verification

- `ruff check` / `ruff format --check` / `mypy` over `strands_robots tests tests_integ`: clean (18 pre-existing errors, unchanged).
- Full `tests/` on this branch vs `upstream/main`: identical failing-id sets, `comm` empty both ways.
- Docs: `docs/training/overview.md` gains a "Dataset format version" section beside the quantile-stats preflight it mirrors, and `validate()`'s own docstring inventory now lists the check.

### DIFF
--- a/changelog.d/2351-dataset-codebase-version-preflight.md
+++ b/changelog.d/2351-dataset-codebase-version-preflight.md
@@ -1,0 +1,15 @@
+### Fixed
+
+`LerobotTrainer.validate()` now refuses a local dataset root whose declared
+`codebase_version` has an older major than the installed lerobot's
+`CODEBASE_VERSION`, instead of reporting the spec as launchable and letting the
+dataset load fail. The refusal names the root, both versions, and a remedy that
+exists: lerobot's `convert_dataset_v21_to_v30` for a v2.1 root, and an explicit
+"no converter" for an older one. Previously a v2.1 root reached
+`BackwardCompatibilityError` whose command named lerobot's `local` sentinel
+rather than a convertible repo, and every other older major reached a bare
+`NotImplementedError: Contact the maintainer on [Discord](...)` that named
+neither the dataset nor the version. The version is read from the same
+`meta/info.json` the trainer already reads, so the check stays offline: a Hub
+dataset with no local cache is left unflagged, and an unparseable version fails
+open.
--- a/docs/training/overview.md
+++ b/docs/training/overview.md
@@ -283,6 +283,38 @@ computes them by default), so they train `molmoact2` / `pi05` with no manual
 stats surgery. The check is conservative: a Hub dataset with no local cache is
 left unflagged (its quantiles are verified by lerobot when the shards load).
 
+#### Dataset format version (`codebase_version`)
+
+Every LeRobotDataset declares the format version it was written in as
+`codebase_version` in `meta/info.json`. lerobot compares it against its own
+`CODEBASE_VERSION` and refuses a dataset whose MAJOR is older - an older *minor*
+loads with a warning. Only a `v2.1` root gets a message naming the dataset and
+the converter; for any other older major lerobot raises
+`NotImplementedError: Contact the maintainer on [Discord](...)` from inside the
+exception constructor, naming neither the dataset nor the version.
+
+The version sits in the same `meta/info.json` the trainer already reads for the
+episode count, so `validate()` decides this offline and returns an actionable
+problem instead:
+
+```
+dataset_root '/data/old_dataset' declares codebase_version 'v2.1' in
+meta/info.json, which lerobot 0.6.2 cannot read (it loads 'v3.0' and refuses an
+older major). Convert it with lerobot's own converter: python -m
+lerobot.scripts.convert_dataset_v21_to_v30 --repo-id=<your-dataset-repo-id>
+```
+
+lerobot ships exactly one conversion (`v2.1` -> `v3.0`), so a root older than
+`v2.1` is told so plainly rather than handed a command that would fail. Datasets
+recorded by `Robot.start_recording()` / `DatasetRecorder` are written in the
+installed lerobot's own format, so they pass cleanly.
+
+Like the quantile-stats check, this one is conservative: it flags only a
+definite mismatch. A Hub dataset with no local cache is left unflagged -
+`validate()` does not reach the network, so Hub-side metadata (the format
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Adds an offline preflight to `LerobotTrainer.validate()` that refuses a local dataset root whose declared `codebase_version` in `meta/info.json` has an older major than the installed lerobot's `CODEBASE_VERSION`, replacing the opaque load-time failures (`BackwardCompatibilityError` naming the `local` sentinel for v2.1, bare `NotImplementedError` for anything older) with a message naming the root, both versions, and an honest remedy. The check reads the same `info.json` the trainer already parses, stays off the network, and fails open on anything unparseable or unknown — matching the posture of the existing `_quantile_stats_problems` sibling it mirrors. Wired into `validate()` so both policy and reward-model runs are gated.

## What's good

- Verified locally at head `c785c581`: all 16 new tests pass, `ruff check` / `ruff format --check` clean on the changed files.
- The lerobot-side constant is read live off `lerobot.datasets.dataset_metadata` (the module the loader enforces it in), with a documented static fallback, and `test_the_offline_fallback_matches_the_version_lerobot_enforces` pins the fallback against the installed lerobot so it cannot drift silently.
- The characterization test on `BackwardCompatibilityError` pins the library behavior the remedy branch depends on (message for exactly 2.1, `NotImplementedError` otherwise).
- Error handling in `_dataset_codebase_version` matches the existing `_dataset_total_episodes` posture exactly (same file, same except set semantics); no new failure classes introduced.
- Changelog fragment per AGENTS.md step 3, docs section added beside the quantile-stats preflight it mirrors, no host paths or non-ASCII in the new files, tests use `tmp_path` throughout.

Issue #1784, opened 2026-07-30
The changelog fragment rule is documented in two places and enforced by nothing


cagataycali:
`changelog.d/README.md` and `AGENTS.md` both state that a behavioural PR records its change as a `changelog.d/<number>-<slug>.md` fragment and never appends to `## [Unreleased]` in `CHANGELOG.md`. The reason is measured, not stylistic: every branch appends at the same anchor, so two open at once conflict on ordering alone, and because a push dismisses a stale approval, clearing that conflict costs a full re-approval round per affected branch.

The convention is followed -- 67 fragments, and the ten most recently merged PRs all use one. But nothing checks it.

## Measured

On `ebe2297b`, with three `### ` entries inserted directly beneath `## [Unreleased]` in `CHANGELOG.md`:

```
$ python scripts/assemble_changelog.py --check
changelog fragments OK (66 pending)                      # exit 0
$ pytest tests/test_changelog_format.py tests/test_changelog_fragments.py
30 passed
```

Neither suite can see it. `test_changelog_format.py` checks the log's version-heading structure; `test_changelog_fragments.py` checks the contents of the fragment directory. A fragment that was never written leaves nothing for either to inspect, so the one rule the convention rests on is the one rule with no gate behind it.

The consequence is not hypothetical: a PR can reach `APPROVED` / `SUCCESS` / `CLEAN` having appended to the log, and reviewers reasonably read a green board as the conventions being satisfied.

## Why a static test cannot do it

The obvious pin -- assert `[Unreleased]` carries no entries, since the assembler fills it at release time -- is not available: `[Unreleased]` on `main` already holds **168** entries predating the convention. A static assertion would have to fail on `main` today, or grandfather a count that silently ratchets upward.

What is wrong is not the section's contents but the act of adding to it, and an act is only visible as a difference between two commits. So this needs a base-diff CI check, in the shape of `merge-base-overlap.yml` / `scripts/check_merge_base_overlap.py` (#1771): compare the entry headings under `[Unreleased]` at the merge base against the head, so the legacy 168 cancel and only what a branch adds remains.

## The release path must stay open

`[Unreleased]` does legitimately gain entries when `scripts/assemble_changelog.py --apply` folds fragments in at release time. Two properties of that tool make an exact exemption possible rather than a blanket skip:

- `render()` joins `fragment.body.strip("\n")`, so a folded entry's heading is byte-identical to its fragment's; and
[...]

PR #2352, opened 2026-08-16
fix(training): refuse streaming alongside a validation split instead of dropping the stream

cagataycali:
## Problem

`TrainSpec` exposes `streaming` and `val_episodes` as independent fields, and `LerobotTrainer.validate()` reported no problem when both were set. Neither field is delivered.

`val_episodes` becomes lerobot's `dataset.eval_split`, and any non-zero `eval_split` routes lerobot into `make_train_eval_datasets`, which rebuilds **both** splits as map-style `LeRobotDataset` objects without ever consulting `dataset.streaming` — the `StreamingLeRobotDataset` it opened one line earlier is discarded. So the run materializes the whole dataset, which is precisely the outcome `streaming` exists to avoid ("streams shards from the Hub with no full download (bounded disk); with a local `dataset_root` it streams from disk (bounded RAM)"), and it does so reporting nothing: an annulled stream is indistinguishable from `streaming=False`.

Measured on lerobot 0.6.2 with a local v3 root and `TrainSpec(streaming=True, val_episodes=1, ...)`:

```
validate() -> []
emitted:  --dataset.streaming=true --dataset.eval_split=0.1666... --eval_steps=1000
lerobot:  train dataset type: LeRobotDataset
          eval  dataset type: LeRobotDataset
          train is streaming? False
```

streaming + val_episodes preflight, before and after

One script, one spec, one local v3 dataset root; only the `strands_robots` tree on `sys.path` differs between the two panels.

The blast radius is the whole point of the field: a caller who asks for a bounded-disk run over a 50-500 GB Hub dataset *and* a validation split gets an unbounded one, with no warning, no log line, and a clean preflight.

## Change

`validate()` refuses the pair, naming both remedies:

```
lerobot_local: streaming=True cannot be combined with val_episodes=2. A held-out split
makes lerobot rebuild both splits as map-style datasets, so the whole dataset is
materialized and the stream is dropped - the disk/RAM blowup streaming exists to avoid.
Either set streaming=False to keep the validation split, or val_episodes=None to keep
the stream.
```

Refusing rather than picking one field mirrors the existing sibling `_unreadable_episode_count_problem`, whose docstring already states the principle for the other direction: "a missing `eval_split` is indistinguishable from 'no validation asked for'. Refuse instead, naming both remedies, rather than launch a run that trains on every episode and records no validation loss." This is the same class, with the streaming request as the thing being silently dropped. It also matches how the backend already handles `num_nodes > 1` and `lora` + `expert_only`: refuse a spec it cannot honor instead of silently delivering something narrower.

The gate fires only when a split will actually be emitted (`_val_eval_split(spec) is not None`), so a Hub source with no readable episode count keeps its own, more accurate refusal instead of gaining a second one about a conflict that spec does not have. `extra` still reaches lerobot's own knobs unchanged for a caller who wants the combination anyway.

## Tests

`TestStreamingAndValidationSplitAreMutuallyExclusive`, 8 tests. Against `main`: **2 fail / 6 pass**; after the change 8 pass.

Failing pre-fix:
- `test_asking_for_both_is_refused` — `validate()` returned `[]` for a spec that delivers neither field.
- `test_the_refusal_names_the_cost_of_the_silent_outcome` — the message must say the dataset is *materialized* and the *stream is dropped*, so a caller learns this is a backend constraint and not a style preference.

Passing on both trees (the controls that keep the refusal from over-reaching):
- `test_streaming_alone_is_launchable` / `test_validation_split_alone_is_launchable` — the refusal must not swallow either feature.
- `test_either_remedy_the_message_offers_clears_it` — both quoted remedies clear the refusal *and* deliver the field they keep, not merely silence the message.
- `test_a_hub_source_with_no_readable_count_keeps_its_own_refusal` — no double-reporting.
- `test_lerobot_split_path_still_drops_the_stream` — pins the measured constraint the refusal stands on: lerobot's eval-split path builds the stream once (to read metadata) and then constructs map-style datasets for both splits. If lerobot starts honoring the stream there, this test fails and the refusal should be lifted rather than left rejecting a combination that has become supportable. That keeps the gate from outliving its reason.

Full `tests/` on lerobot 0.6.2, run on both trees: **61 failing ids on this branch vs 66 on `main`, zero branch-only** (the 5 base-only ids are the 2 pre-fix failures above plus 3 known load-sensitive timing flakes). 30637 passed vs 30632.

## Docs

`TrainSpec.streaming` and `TrainSpec.val_episodes` now state the exclusion and why; `docs/training/overview.md` gains it in the field table and a paragraph next to the existing `val_episodes` source-requirement note.

---

## Round 2 changes

No review feedback; this round is a merge-base update only. The approval was dismissed by the push, and nothing about the change under review moved.

**Merged `main` in to resolve a conflict this PR's sibling created.** #2351 landed while this was green and waiting, and it appended a clause to the *same sentence* of `LerobotTrainer.validate`'s docstring that this branch appended one to, so the branch went `CONFLICTING` the moment #2351 merged. The two changes are independent in behaviour and adjacent in text: one refuses an unreadable dataset format version, the other refuses `streaming` alongside a validation split, and both belong in the same enumeration of what the preflight checks.

The conflict is entirely inside that docstring — one hunk, three lines — and the resolution keeps **both** clauses, in the order the checks run:

```
(``num_nodes == 1``), a ``val_episodes``
split below the dataset total and not asked for alongside ``streaming``
(lerobot's split path is map-style only), a local dataset format version
the installed lerobot can read, usable LoRA hyperparameters when
``method == "lora"``, ...
```

No code hunk conflicted, and both siblings' checks are present and independently reachable in the merged `validate()`: `_streaming_validation_split_problem` inside the `val_episodes` block, `_dataset_codebase_version_problems` appended after it. A docstring-only conflict is the failure mode #1784's fragment convention removes for changelog entries, and the same shape shows up in a shared summary docstring — worth noting, but not worth a shared-file convention for one sentence.

Verified against the merged tree rather than either parent, since no check had compiled the two together: `tests/training/` — `test_lerobot.py` (this PR's suite), `test_dataset_codebase_version_preflight.py` (#2351's), and the three docstring gates that read this exact docstring (`test_docstring_module_xrefs.py`, `test_public_member_docstrings.py`, `test_trainer_provider_docstrings.py`) — 109 passed, 37 skipped, with the 23 failures all reporting `lerobot is not installed` (the optional dep is absent from this sandbox; the same ids fail identically on unmodified `main` here). `ruff check` and `ruff format --check` clean on the resolved file.

### DIFF
--- a/changelog.d/2352-streaming-validation-split-exclusive.md
+++ b/changelog.d/2352-streaming-validation-split-exclusive.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- `LerobotTrainer.validate()` now refuses `streaming=True` together with `val_episodes`
+  instead of reporting a launchable spec that delivers neither. A non-zero
+  `dataset.eval_split` routes lerobot into `make_train_eval_datasets`, which rebuilds both
+  splits as map-style `LeRobotDataset` objects without consulting `dataset.streaming`, so
+  the whole dataset was materialized - the disk/RAM blowup `streaming` exists to avoid -
+  with nothing reported. The refusal names both remedies.
--- a/docs/training/overview.md
+++ b/docs/training/overview.md
@@ -91,12 +91,12 @@ supports and **ignores the rest** (the same tolerance rule as
 |-------|---------|-------|
 | `dataset_root` | LeRobotDataset v3 root | a data source; has `meta/info.json` (optional when `dataset_repo_id` is set) |
 | `dataset_repo_id` | Hub dataset id `org/name` | alternative data source; train from the Hub (lerobot) |
-| `streaming` | stream frames, no full materialize | lerobot `StreamingLeRobotDataset`; bounded disk (Hub) / RAM (local) |
+| `streaming` | stream frames, no full materialize | lerobot `StreamingLeRobotDataset`; bounded disk (Hub) / RAM (local); mutually exclusive with `val_episodes` |
 | `base_model` | HF id / local ckpt to tune from | required for GR00T & Cosmos |
 | `steps` / `global_batch_size` | the run size: optimizer steps x batch | each must be a positive integer; `validate()` refuses `0`, a fractional or non-finite value, and a `bool` (`True` would read as a silent one-step run) before anything is loaded |
 | `method` | `full` \| `lora` \| `expert_only` \| `frozen_backbone` | `lora`+`expert_only` are mutually exclusive |
 | `tune` | `{llm,visual,projector,diffusion}` | GR00T only |
-| `val_episodes` | hold out the LAST N episodes | deterministic split; must be a positive integer below the dataset's episode count, and that count must be readable from a local `meta/info.json` (see the Hub-source note below). `validate()` refuses `0` or a negative (they produced no split and no eval cadence at all - the run trained on everything and logged no validation loss), a `bool`, and a fractional value (`2.7` reserved 3 episodes, `0.5` reserved none while still evaluating) |
+| `val_episodes` | hold out the LAST N episodes | deterministic split; must be a positive integer below the dataset's episode count, and that count must be readable from a local `meta/info.json` (see the Hub-source note below). `validate()` refuses `0` or a negative (they produced no split and no eval cadence at all - the run trained on everything and logged no validation loss), a `bool`, and a fractional value (`2.7` reserved 3 episodes, `0.5` reserved none while still evaluating); mutually exclusive with `streaming` |
 | `num_gpus` / `num_nodes` | multi-GPU / multi-node | selects the launcher |
 | `seed` | reproducibility seed | must be a non-negative integer; `validate()` refuses a negative (`torch.manual_seed` would take it modulo `2**64`, so `-1` silently becomes `2**64 - 1`), a fractional or non-finite value, and a `bool`. `None` uses the backend's own default |
 | `extra["policy_type"]` | lerobot `--policy.type` | act/diffusion/smolvla/pi0/pi05/... |
@@ -314,6 +314,15 @@ than launch a run that trains on every episode and logs no validation loss. The
 refusal names the two ways to get the split: point `dataset_root` at a populated
 local copy of the dataset, or pass lerobot's own knobs directly with
 `extra={"dataset.eval_split": 0.1, "eval_steps": 1000}`.
+
+`streaming` and `val_episodes` cannot both be honored, so `validate()` refuses
+the pair. A non-zero `eval_split` routes lerobot into `make_train_eval_datasets`,
+which rebuilds BOTH splits as map-style `LeRobotDataset` objects without
+consulting `dataset.streaming` - the streaming dataset it opened first is
+discarded. The run therefore materializes the whole dataset, which is exactly
+what `streaming` exists to avoid, and nothing reports it: an annulled stream is
+indistinguishable from `streaming=False`. Set `streaming=False` to keep the
+validation split, or `val_episodes=None` to keep the stream.
 
 ### GR00T (`groot`)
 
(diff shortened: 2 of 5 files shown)


yinsong1986 (dismissed):
## Summary

Adds a preflight refusal in `LerobotTrainer.validate()` for `streaming=True` combined with `val_episodes`: lerobot's non-zero `dataset.eval_split` routes into `make_train_eval_datasets`, which rebuilds both splits as map-style `LeRobotDataset` objects without consulting `dataset.streaming`, so the stream was silently dropped and the whole dataset materialized. I verified the premise independently against the lerobot 0.6.1 wheel (the lockfile version): `make_train_eval_datasets` constructs both splits as map-style `LeRobotDataset` unconditionally after using the `make_dataset` result only for metadata, exactly as the refusal claims. The gate is correctly scoped to `_val_eval_split(spec) is not None`, so a Hub source with no readable episode count keeps its own more accurate refusal, and `train()` fails closed on the problem list, so the refusal actually gates the run. The `extra` passthrough is preserved as the documented escape hatch, so no capability is removed - a previously silent misdelivery becomes an explicit refusal naming both remedies, which matches the repo's no-silent-defaults rule and the sibling `_unreadable_episode_count_problem` pattern.

## What's good

- The canary test (`test_lerobot_split_path_still_drops_the_stream`) pins the *reason* for the refusal against lerobot itself, so the gate is lifted rather than outliving its constraint if lerobot starts honoring the stream on the split path.
- Controls keep the refusal from over-reaching: each field alone stays launchable, both quoted remedies are tested to deliver the field they keep, and the Hub-source case is tested for no double-reporting.
- Changelog fragment, `TrainSpec` docstrings, and `docs/training/overview.md` all updated in the same PR; diff is ASCII-clean with no host paths.

## Verification suggestions

The new test suite requires lerobot >= 0.6.1 (the project floor). On an environment with lerobot 0.5.x installed, `test_lerobot_split_path_still_drops_the_stream` fails with `AttributeError: no attribute 'make_train_eval_datasets'` - that is a stale-env artifact, not a PR defect (I reproduced an identical pre-existing `TestBuildConfig::test_val_split_sets_eval_split_and_a_nonzero_eval_cadence` failure on `main` under 0.5.1). CI installing from the lock (0.6.1) is the authoritative run.

yinsong1986 (approved):
## Summary

Adds a preflight refusal in `LerobotTrainer.validate()` for `streaming=True` combined with `val_episodes`: a non-zero `dataset.eval_split` routes lerobot into `make_train_eval_datasets`, which rebuilds both splits as map-style `LeRobotDataset` objects without consulting `dataset.streaming`, so the requested stream was silently dropped and the whole dataset materialized - the disk/RAM blowup `streaming` exists to avoid. The gate is correctly scoped: it sits inside the `val_episodes` block and fires only when `_val_eval_split(spec) is not None`, so a Hub source with no readable episode count keeps its own more accurate refusal, and the `extra` passthrough remains as the documented escape hatch, so no capability is removed - a silent misdelivery becomes an explicit refusal naming both remedies, matching the sibling `_unreadable_episode_count_problem` pattern and the repo's no-silent-defaults rule. This round is a merge-base refresh only (`behind_by=0` against `main`); the sole conflict was the shared `validate()` docstring sentence with #2351, and the resolved tree carries both siblings' checks independently reachable, which I verified by reading the merged `validate()` directly. Ran the new suite against the merged head: 7/8 pass locally, with the one failure (`test_lerobot_split_path_still_drops_the_stream`, `AttributeError: no attribute 'make_train_eval_datasets'`) being exactly the lerobot-0.5.x stale-env artifact the prior review already documented - the project floor is `lerobot>=0.6.1,<0.7.0`, where the function exists, and CI installing from the lock is authoritative. Diff is ASCII-clean, no host paths, changelog fragment follows the convention.

## What's good

- The canary test pins the external constraint the refusal stands on, so the gate is lifted rather than outliving its reason if lerobot starts honoring the stream on the split path.
- Both quoted remedies are tested to deliver the field they keep, not merely silence the message - the refusal text and the behavior cannot drift apart.
- The merge-base refresh kept both #2351's and this PR's docstring clauses in check-execution order, and the description reports verification against the merged tree rather than either parent.

PR #2353, opened 2026-08-16
fix(kimodo): ease a chained segment off the pose last commanded

cagataycali:
## Problem

Changing the prompt mid-rollout is the documented way to drive a long-horizon sequence through `KimodoPolicy`. Its own docstring says the instance "synthesises ONCE on first call (or when the prompt changes) then streams per-frame targets", and `_sample_key` includes the prompt precisely so a new prompt samples the next segment and the stream continues.

Kimodo samples every motion from its own canonical start pose, though, so a fresh segment's first frame bears no relation to where the previous segment left the robot. It was emitted unmodified, stepping all 29 joints in a single tick.

Driving the shipped policy over the 600 ordered pairs of a 25-motion corpus (each pair = one prompt change mid-rollout):

| | before | after |
|---|---|---|
| worst seam step | 2.866 rad (164 deg) in one tick | **0.478 rad** |
| median seam step | 1.599 rad | **0.266 rad** |
| seams exceeding the largest per-tick step the motions themselves take (0.476 rad) | **507/600 (84.5%)** | 1/600 (0.2%) |

Every one reported `status=success`. The median seam moved a joint further in one tick than the motion ever moves, so the emitted reference is not one a tracker can follow, and nothing said so.

The module already carries a smoothness guarantee, on a far smaller effect: `_slerp_upsample` documents that it widens 30 Hz to 50 Hz "without introducing joint discontinuities". The 30-to-50 Hz resample was interpolated; the segment boundary was not.

## Change

A newly sampled segment is eased off the pose last commanded, decaying the pose offset to zero over `KimodoConfig.transition_frames` native frames. The motion keeps its own shape and velocity; only its starting offset is absorbed.

The default of **5** and the `>= 1` domain are not invented: they are what Kimodo's own sampler applies to `num_transition_frames` when it generates a multi-prompt sequence (it raises `ValueError` below 1). Easing is applied at the native rate, before the upsample, so `transition_frames` counts the same frames the sampler's knob counts.

This differs in kind from the sampler's native path, and the docs say so: Kimodo conditions the *diffusion* of the next segment on the previous segment's last frames. The `KimodoMotionAgent` protocol carries only a prompt and sampling knobs, with no continuation state, so easing shapes the emitted stream rather than the sample. It removes the discontinuity; it does not re-plan the motion around it.

## Scope

Only the seam is touched:

- **A single-prompt rollout is byte-for-byte unchanged** - with no previously commanded pose there is nothing to ease onto. Verified identical over a 150-tick rollout.
- Frames past the transition window are exactly the sampler's own.
- `reset()` forgets the commanded pose, so an episode boundary opens on the motion's own start pose. Forgetting the pose alone was not enough: a buffer already eased onto it would still be replayed from frame 0, so `reset()` rebuilds it from the held sample, without paying for another diffusion run.

The quaternion arm of the upsample and the new transition now share one `_slerp_quat` helper. The extraction is **bit-identical** over all 4975 upsampled frames of the corpus (`np.array_equal` true), including preserving the float32 scalar weight - widening it to a Python float shifted results by 1 ULP.

## Tests

`tests/policies/kimodo/test_chained_prompt_transition_is_continuous.py`, **5 failed / 2 passed before, 13 pass after**. The headline runs on the default config and compares two measured quantities, so on unfixed code it reports the size of the step rather than the absence of a setting:

```
AssertionError: changing the prompt stepped a joint 1.6564 rad in a single tick
out of a 1.6564 rad pose gap (100% of it): a fresh sample is being emitted from
its own start pose instead of being eased off the pose last commanded
```

The two that pass on both trees are the no-overreach controls: a single-prompt rollout is the sampler's own motion, and an episode boundary opens on the motion's own start pose.

The seam is asserted as a *bound* (the gap is spread, not stepped) rather than as "inside the motion's own per-tick envelope". The latter reads stronger but depends on how fast the particular motions move relative to the gap between them, so it can pass on one corpus and fail on another for no change in behaviour.

## Visual

Chained rollout in MuJoCo, headless EGL, worst measured transition ("walking forward" to "crouching to pick an object"), same script under `upstream/main` and this branch, driven by real Kimodo samples. The seam tick is 2.430 rad before and 0.405 rad after; nothing else in the clip exceeds 0.245 rad.

chained prompt transition

Full-resolution MP4: https://raw.githubusercontent.com/cagataycali/robots/media-kimodo-chain-31976878374/kimodo-chain.mp4

Seam tick alone: https://raw.githubusercontent.com/cagataycali/robots/media-kimodo-chain-31976878374/kimodo-seam.png

## Gate

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 18 errors, identical to `upstream/main` (0 new). Full `tests/` run on both trees with no new failures. Docs updated with a chaining recipe and the new config row; changelog fragment included.

### DIFF
--- a/changelog.d/2353-kimodo-chained-prompt-transition.md
+++ b/changelog.d/2353-kimodo-chained-prompt-transition.md
@@ -1,0 +1,12 @@
+### Fixed
+
+`KimodoPolicy` now eases a newly sampled motion off the pose it last commanded
+when the prompt changes mid-rollout, instead of emitting the fresh sample from
+its own canonical start pose. Chaining prompts is the documented way to drive a
+long-horizon sequence, but every segment boundary stepped all 29 joints at once:
+across the 600 ordered pairs of a 25-motion corpus the median seam moved a joint
+1.6 rad in a single tick, and 84% of transitions exceeded the largest per-tick
+step the motions themselves ever take, all reported as a successful rollout. The
+transition length is `KimodoConfig.transition_frames` (default 5, matching the
+`num_transition_frames` Kimodo's own sampler applies to a multi-prompt
+sequence). A single-prompt rollout is byte-for-byte unchanged.
--- a/docs/policies/kimodo.md
+++ b/docs/policies/kimodo.md
@@ -115,6 +115,7 @@ it, or a tuned PD law) is required, and is out of scope for this provider.
 | `diffusion_steps` | int | 100 | 25–200 useful range |
 | `guidance_scale` | float | 7.5 | CFG weight |
 | `num_frames` | int | 120 | ≤196 (RP-v1 max) |
+| `transition_frames` | int | 5 | Native frames a chained segment is eased over |
 | `native_fps` | int | 30 | Sampler native rate |
 | `tracker_fps` | int | 50 | SLERP upsample target |
 | `device` | str \| None | auto | `"cuda"` / `"cpu"` |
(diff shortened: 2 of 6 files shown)


yinsong1986 (approved):
## Summary

Fixes a real discontinuity in `KimodoPolicy`'s documented prompt-chaining path: a mid-rollout prompt change re-samples from Kimodo's canonical start pose, so the seam stepped all 29 joints in one tick while reporting success. The fix eases a freshly sampled segment off the pose last commanded over `KimodoConfig.transition_frames` native frames (default 5, matching the sampler's own `num_transition_frames` default and `>= 1` domain), applied before the 30->50 Hz upsample so the knob counts the same frames the sampler's does. `reset()` both forgets the commanded pose and rebuilds the buffer from the held raw sample, so an episode boundary is not treated as a seam. The quaternion slerp shared by the upsample and the ease is extracted into `_slerp_quat` with the float32 weight preserved.

## What's good

- New config field is validated through the shared `positive_whole_number_error` domain (bool/NaN/float rejected, pinned by a parametrized test), and `transition_frames` is threaded through the constructor, `_resolve_config`, and `registry/policies.json` `config_keys` in one motion — verified locally that `tests/registry/test_config_keys_agree_with_constructors.py` passes at this head.
- The no-overreach controls are the right ones: single-prompt rollout asserted byte-identical to the sampler's own frames, post-transition frames asserted equal to the virgin segment, and the episode-boundary rebuild covered by a dedicated test.
- The headline seam test asserts a bound derived from two measured quantities on the default config, so it fails on unfixed code with the size of the step rather than the absence of a setting.
- All 74 kimodo unit tests pass locally at `0157e950` (one unrelated collection error in `test_diffusers_agent_checkpoint_layout.py` is a local `huggingface_hub` version mismatch, not this diff).

cagataycali:
Local gate, concrete numbers (full `tests/`, same interpreter, run against this branch and against `upstream/main`):

|---|---|---|
| failing test ids | 53 | 52 |
| passed | 30637 | 30652 |
| collected | 30973 | 30987 |

`comm` of the sorted failing-id sets gives **zero branch-only failures**. The single base-only id is `test_hardware_task_loop_dispatch.py::TestTheRolloutIsDrivenExactlyOnce::test_on_the_nested_branch`, a timing-sensitive test in a rate-guarded loop that passes in isolation on both trees; it is not something this change fixes.

The +14 collected reconciles exactly: 13 from the new test file, plus `test_every_config_keys_entry_names_a_constructor_parameter[kimodo-transition_frames]`, which the existing registry guard auto-parametrises over the new `config_keys` entry and which passes.

`mypy strands_robots tests tests_integ` reports the same 18 errors on both trees (none in the touched files); `ruff check` and `ruff format --check` are clean over the same scope.

PR #2354, opened 2026-08-16
fix(policies/lerobot_local): report a declarative state_keys mismatch with the remedy the generic path already gives

cagataycali:
## Problem

The generic `robot_state_keys` path reports both of its binding degradations the same way: it quotes the keys the observation actually carries, and appends the registry-checked remedy from `state_key_remedy`, which names an embodiment only when the registry confirms that embodiment binds THIS observation. Its own comment says so - *"Same registry-checked remedy as the all-missing guard, so one rule serves both degradations."*

A declarative `embodiment=` installs `PackStateProcessorStep` instead, and it got neither half:

* **All-missing** - the branch returned the observation untouched and said **nothing at all**. No warning, no error, no remedy, no telemetry. `stop()`-style silence: the run continues, downloads the checkpoint, and fails somewhere downstream for want of `observation.state` - a message that knows nothing about embodiments and so cannot name the one that binds this observation.
* **Partly-missing** - warned, but asserted a single cause (a mimic/tendon gripper) as fact, showed nothing of what the observation does carry, and offered no remedy. Asserting a cause is the exact anti-pattern the generic path documents avoiding: it reads the cause instead, *"so a caller who configured a NAMED set describing a different robot is not told their keys were auto-generated placeholders they never configured."*

`docs/policies/lerobot-local.md` already claimed the fixed behaviour - *"Both degradations quote a remedy chosen from the observation itself rather than a fixed example"* - which held only on the generic path.

## How it is reached

The embodiment registry has two key conventions: sim entries use bare MuJoCo joint names, `*_real` entries use lerobot driver `'<motor>.pos'` names. One direction already had a fallback; the other had no report.

Six alias spellings are **both** a lerobot driver name and a sim-loadable registry robot, and each resolves to a `*_real` configuration:

| `embodiment=` | resolves to | declared state_keys | present in a sim observation |
|---|---|---|---|
| `so100_follower`, `so101_follower` | `so_real` | 6 x `'<motor>.pos'` | 0 of 6 |
| `koch_follower` | `koch_real` | 6 x `'<motor>.pos'` | 0 of 6 |
| `bi_so_follower` | `bi_so_real` | 12 x `'<motor>.pos'` | 0 of 12 |
| `openarm` | `openarm_real` | 8 x `'<motor>.pos'` | 0 of 8 |
| `lekiwi` | `lekiwi_real` | 6 x `.pos` + 3 x `.vel` | 0 of 9 |

Measured on `Robot("so101_follower", mode="sim")`: declared `['shoulder_pan.pos' ... 'gripper.pos']`, observed `['1','1.vel' ... '6','6.vel']`, `observation.state` not packed, **0 log records**. The mirror case (`embodiment="so101"` on a real arm) binds via the `'.pos'` fallback and is untouched here.

## Change

Both degradations now go through one report that quotes the declared keys, the observed keys, and `state_key_remedy(...)`. For the sim case above that resolves to `Pass embodiment='so101', whose state_keys this observation carries, or call set_robot_state_keys([...])` - checked against the registry, so the embodiment it names is one the observation provably binds.

**Nothing about the packed vector changes.** The all-missing branch still passes the observation on untouched (an unbound state is a configuration mistake, not a degraded reading, so no all-zero vector is invented), and a partly-bound vector keeps its in-place zero-fill.

The state-key candidate rule is now one function, `observed_state_keys`, instead of the same rule restated in two modules - which is what lets both paths quote the same answer. Its former copy in `policy.py` is deleted and its three call sites updated; no alias is kept.

One existing test selected the partial-mismatch warning by a phrase now shared with the generic path's message; it selects on the report's subject instead, which is what it was actually asserting.

## Tests

`tests/policies/lerobot_local/test_declarative_state_key_mismatch_diagnostics.py`, run against `main`: **9 failed, 4 passed**; after: **13 passed**.

| pre-fix | test |
|---|---|
| FAIL | a total mismatch does not return silently (`expected exactly one report, got []`) |
| FAIL | it names the declared keys and the observed keys |
| FAIL | its remedy names an embodiment that binds this observation |
| FAIL | reporting replaces the silence and not the passthrough |
| FAIL | it is reported once per distinct mismatch (`assert 0 == 2`) |
| FAIL | a partial mismatch offers a registry-checked remedy (`assert 0 == 1`) |
| FAIL | a partial mismatch quotes the keys the observation carries |
| FAIL | a partial mismatch remedy does not invent an embodiment |
| FAIL | the report quotes exactly the shared candidate rule |
| pass | a fully bound embodiment reports nothing |
| pass | the hardware `.pos` fallback still binds and stays quiet |
| pass | a partial mismatch still zero-fills in place |
| pass | the rule excludes the instruction and camera frames |

The four constant passes are the controls: they pin what the added reporting must not disturb - the mirror direction keeps its fallback, a bound embodiment stays quiet, and the zero-filled vector keeps its indices.

The remedy assertion parses `embodiment='X'` back out of the message and checks `X` against `matching_embodiments(observed)`, so it fails for a missing remedy and for a wrong one - it pins that the advice is followable rather than pinning its wording.

## Verification

`ruff check` / `ruff format --check` / `mypy` clean over `strands_robots tests tests_integ` (mypy reports only the errors already present on `main`, none in `policies/lerobot_local`).

The change's blast radius - `tests/policies`, `tests/simulation`, `tests/registry`, and the repo-wide docstring / doc-coverage / ASCII / dependency-audit / changelog guards - was run on this branch and on `main` with the same interpreter:

| | passed | skipped | failed |
|---|---|---|---|
| `main` | 16079 | 232 | 0 |
| this branch | 16092 | 232 | 0 |

`+13` is exactly the new test count, and no test fails on either side.

## Rollout

Real MuJoCo so101 (headless EGL), shipped catalog, the step a declarative `embodiment=` installs - nothing stubbed. One capture script run under each tree; the rendered robot is identical between them (mean abs pixel delta 0.0001), so the only difference is the report.

declarative embodiment state_keys mismatch, before and after

### DIFF
--- a/changelog.d/2354-declarative-state-key-mismatch-report.md
+++ b/changelog.d/2354-declarative-state-key-mismatch-report.md
@@ -1,0 +1,3 @@
+### Fixed
+
+A declarative `embodiment=` now reports a `state_keys` mismatch with the same registry-checked remedy the generic `robot_state_keys` path gives. An all-missing binding previously returned the observation untouched and reported nothing at all, so the only signal was a downstream error that knows nothing about embodiments; a partly-missing binding warned but asserted a single cause and offered no remedy. Both now name the declared keys, the keys the observation carries, and an embodiment the registry confirms binds that observation. Nothing about the packed state vector changes.
--- a/docs/policies/lerobot-local.md
+++ b/docs/policies/lerobot-local.md
@@ -351,6 +351,36 @@ velocity state (LeKiwi's body-frame base velocities `x.vel` / `y.vel` /
 `theta.vel`). Explicit `robot_state_keys` are never filtered - naming `elbow.vel`
 there states the model's input.
 
+### A declarative `embodiment=` reports the same two degradations
+
+A declarative `embodiment=` installs its own state step, and it reports both
+degradations with that same registry-checked remedy. A partly-bound vector names
+the absent keys, the keys the observation does carry, and the remedy; an
+all-bound-nothing configuration says so instead of handing the observation on in
+silence, since the downstream error it would otherwise produce knows nothing
+about embodiments and cannot name the one that binds this observation.
+
+The two conventions in the embodiment registry are what make an all-missing
+binding easy to reach in either direction:
+
+- **A sim embodiment driven from real hardware.** `embodiment="so101"` declares
+  the MuJoCo asset's `'1'..'6'`, and a real arm reports `'<motor>.pos'`. This
+  direction binds anyway: the step falls back to the observation's `'.pos'` keys
+  in motor order, packed raw because hardware already reports the model's
+  training units.
+- **A `*_real` embodiment driven from sim.** Several names are both a lerobot
+  driver spelling and a sim-loadable registry robot - `so100_follower`,
+  `so101_follower`, `koch_follower`, `bi_so_follower`, `lekiwi`, `openarm` - and
+  each resolves to a `*_real` configuration whose `'.pos'` keys no sim
+  observation emits. There is no fallback in this direction, so no
+  `observation.state` is packed; the report names the sim-side configuration to
+  use instead (`embodiment='so101'` for a MuJoCo so101).
+
+An unbound state is a configuration mistake rather than a degraded reading, so
+nothing is invented to stand in for it - no all-zero vector is emitted, and the
+observation is passed on unchanged for a state-less policy or the caller's own
+handling.
+
 ## Camera routing
 
 Robot/sim observations use bare camera names (`top`, `wrist`, `side`); the policy
--- a/strands_robots/policies/lerobot_local/embodiment.py
+++ b/strands_robots/policies/lerobot_local/embodiment.py
@@ -28,7 +28,7 @@ from __future__ import annotations
 import json
 import logging
 import math
-from collections.abc import Iterable
+from collections.abc import Iterable, Mapping
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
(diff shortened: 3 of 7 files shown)


yinsong1986 (approved):
## Summary

Brings the declarative `embodiment=` state path (`PackStateProcessorStep`) up to parity with the generic `robot_state_keys` path for both state_keys binding degradations: an all-missing binding now warns instead of returning the observation in silence, and a partly-missing binding quotes what the observation carries plus the registry-checked `state_key_remedy(...)` instead of asserting a mimic-gripper cause as fact. The candidate-key rule is consolidated into one shared `observed_state_keys` function (the `_scalar_observation_keys` copy in `policy.py` is deleted with all three call sites updated, no dangling references). Verified against head `b8de36b2`: the new + updated diagnostic tests (32 tests across the three touched test files) all pass; a probe against the merge base confirms the pre-fix all-missing case packed nothing and logged zero records, so the tests pin real behavior. The packed-vector semantics are untouched in both branches - all-missing still passes the observation through (no invented all-zero state, per the no-silent-defaults rule), partial keeps its in-place zero-fill, and the hardware `.pos` fallback direction stays quiet - each pinned by a control test. Warn-once dedup is keyed per distinct (missing, observed) pair so the 50Hz loop stays quiet after the first report.

## What's good

- Control tests explicitly pin what the reporting must not disturb (passthrough, zero-fill indices, `.pos` fallback silence, quiet fully-bound path).
- The remedy assertion parses `embodiment='X'` back out and checks it against `matching_embodiments(observed)`, pinning followability rather than wording.
- The existing warn-selection test was moved off a now-shared phrase onto the report's stable subject, correctly scoping the message wording to its own contract.
- Changelog fragment present; ruff clean; plain-ASCII messages; `Args:` entries on the new public function.

PR #2355, opened 2026-08-16
fix(simulation/mujoco): refuse remove_robot when the cooperative stop lapses instead of un-tracking a live worker

cagataycali:
## Problem

`remove_robot` stops the target robot's own policy worker cooperatively, then waits a bounded 5s for it to exit:

```python
self._world.robots[name].policy_running = False
try:
    self._policy_threads[name].result(timeout=5.0)
except Exception:
    pass
del self._policy_threads[name]          # deleted regardless
```

`Future.result(timeout=...)` raises `TimeoutError` when the worker has *not* exited. That is swallowed and the tracking entry is deleted anyway, so a worker that outlived the budget is reported as removed while it is still running.

The tracked Future is the single record every bound on a live worker reads: the global scene-mutation gate (`_require_no_running_policy`), `list_policies_running`, and `cleanup`'s own bounded join. Deleting it does not end the worker - it makes the worker **unobservable**, and the blindness outlives the `remove_robot` call.

## Measured

A policy parked inside inference (a policy server that stops answering, a stalled RTC query, a long VLA forward pass) cannot read the stop flag, which is only checked between control ticks. Driving the shipped path on a one-arm world:

| observable | before | after |
| --- | --- | --- |
| `remove_robot` verdict | `success` - "Robot 'arm1' removed." (after 5.01s) | `error`, naming the budget + the retry |
| worker actually running | `True` | `True` |
| tracked in `_policy_threads` | **`False`** | `True` |
| `list_policies_running` | "No policies running." | "Active policies (1): - arm1" |
| global scene-mutation gate | **ALLOWS** | REFUSES |
| `add_object` while worker runs | **`success`** | `error` |
| `set_timestep` while worker runs | **`success`** | `error` |
| `add_robot` while worker runs | **`success`** | `error` |
| `cleanup` has anything to join | **`False`** (0.0s, no warning) | `True` |
| retry after the worker exits | n/a | `success`, robot gone, `step` OK |

Nothing raises and nothing is logged - `Future.exception()` is `None`, because `PolicyRunner` unwinds cleanly.

The three admitted mutations are the point. Each reallocates `MjModel`/`MjData`, which is exactly what the gate's own docstring says it exists to refuse ("any live PolicyRunner worker holding pointers to the old arrays will segfault when it next calls `mj_step`. Hard-fail."). Because the entry is gone for good, that admission is not confined to `remove_robot`'s own rebuild - it lasts the rest of the session. `cleanup` then runs `executor.shutdown(wait=False)` on the documented premise that "we've already drained policy workers above", which is not true of a worker nobody tracks.

Scope note on the mechanism: the woken worker does **not** write through stale actuator ids. `send_action` resolves keys by name each tick and the robot is gone from `world.robots`, so a second arm's `ctrl` stayed at `0.0` throughout (measured). The defect is the broken safety invariant, the surface that misreports it, and the worker no path can join - not memory corruption.

## Change

Keep the tracking entry and refuse the rebuild:

```
Robot 'arm1' still has a live policy worker after waiting 5.0s for it to stop, so the
scene rebuild was refused: it would reallocate the model/data that worker holds. The
cooperative stop has been requested and the worker exits at its next control tick -
retry action='remove_robot' (until it does, action='list_policies_running' still
reports it).
```

Keeping one entry makes all four surfaces consistent at once: the gate refuses, `list_policies_running` is truthful, `cleanup` can join it, and the retry the message names works because the stop has already been requested.

**The lapse is decided by `Future.done()`, not by exception class.** `socket.timeout` *is* `TimeoutError` (both are `OSError` subclasses), so a policy server that stops answering raises the very class the join raises. Typing on the class would refuse a removal whose worker had already exited; `done()` answers `True` vs `False` for the two cases. `done()` is also what `_require_no_running_policy` and `_prune_done_futures` already use.

Refusing rather than joining unbounded is deliberate: an open-ended join is the caller hang that `_DEFAULT_POLICY_STOP_TIMEOUT` exists to prevent. The budget is now read from that constant instead of a second hardcoded `5.0`, so the two paths that implement one cooperative-stop protocol cannot drift to different answers.

An AST sweep for the same shape across `strands_robots/` - a suppressed join/wait followed by unconditionally dropping the tracking state it was waiting on - finds exactly one site, the one fixed here. `cleanup` is the other place that joins policy futures, and it already handles a lapse correctly: it logs, naming the robot and the budget, because continuing is the deliberate choice on a teardown path. `remove_robot` is a mutation path, where continuing is what breaks the invariant.

## The test that pinned this contract was vacuous

`tests/simulation/mujoco/test_engine_locking_discipline.py` already names this exact failure - its module docstring says `remove_robot` must return "(no swallowed cooperative-stop timeout) ... instead of ... rebuilding the scene under a still-live worker", and `test_remove_robot_during_active_policy_returns_and_worker_exits_clean` describes it in full. It could not observe it: its `_SlowPolicy` stub omits `set_control_frequency` (and `set_rtc_observed_delay`, and returns a single dict where the runner wants an action chunk), so the worker died on its first query and the test raced a *finished* Future.

Adding only a premise assert to that test, on unmodified `main`:

```
AssertionError: premise: worker already finished
  (AttributeError("'_SlowPolicy' object has no attribute 'set_control_frequency'"))
```

The stub now subclasses `Policy`, and the premise assert stays so it can never silently go vacuous again. That test still passes with a genuinely live worker, which confirms the lock half of the contract is real.

## Artifact

One probe script, run against each tree, in MuJoCo sim headless (`MUJOCO_GL=egl`). A policy worker is parked inside inference so it cannot read the stop flag; the render is the resulting scene after the cascade.

remove_robot cooperative-stop lapse

Left: `remove_robot` answers `success` after 5.0s, and the cube and second arm are then injected into the scene while the worker is still live (`robots=['arm2'] bodies=3`). Right: the rebuild is refused and the scene the gate protected is unchanged (`robots=['arm1'] bodies=2`). Render panels differ by mean|delta| 11.89; the injected cube is 841 px on the left and 0 on the right.

## Tests

`6 failed / 10 passed` on `upstream/main`, `16 passed` after. Headline failure:

```
remove_robot reported 'success' while its policy worker was still live (future.done() == False)
```

The failing six cover the verdict, the gate blindness, `list_policies_running`, cleanup joinability, the remedy roundtrip (the action is parsed back out of the message and applied, so a refusal naming an action that does not resolve the situation fails), and an end-to-end case with a real `PolicyRunner` worker parked inside inference.

Three pass on both trees as no-overreach controls:

- a worker that raised `TimeoutError` is still removed - this is the one that fails if the lapse is detected by exception class;
- a worker that stopped within the budget is removed unchanged;
- a robot with no policy is removed unchanged (the join is skipped entirely).

The deterministic tests inject the Future so the two branches need no sleeping or scheduler luck; the end-to-end test earns it with a real worker.

Full `tests/` on Thor (mujoco 3.5.0, lerobot 0.6.2), branch and `upstream/main` run in parallel under identical load: **no regressions** - the set difference of failing test ids is empty (`28 failed / 30668 passed` vs `30 failed / 30657 passed`; the `+11` is the 9 new tests plus two known wall-clock-sensitive tests that happened to pass on the branch run, not fixes). `ruff check` / `ruff format --check` / `mypy` clean over `strands_robots tests tests_integ`.

### DIFF
--- a/changelog.d/2355-remove-robot-cooperative-stop-lapse.md
+++ b/changelog.d/2355-remove-robot-cooperative-stop-lapse.md
@@ -1,0 +1,10 @@
+### Fixed
+
+`remove_robot` no longer reports a robot as removed when its policy worker outlived the
+cooperative-stop budget. The join's `TimeoutError` was swallowed and the worker's tracking
+entry deleted regardless, which left a live worker unobservable: the global scene-mutation
+gate then admitted `add_object` / `set_timestep` / `add_robot` for the rest of the session,
+`list_policies_running` reported nothing, and `cleanup`'s bounded join had nothing to wait
+on. The rebuild is now refused with the budget waited and the retry that resolves it, and
+the worker stays tracked. A lapse is distinguished from a worker that raised by
+`Future.done()` rather than by exception class, since `socket.timeout` is `TimeoutError`.
--- a/strands_robots/simulation/mujoco/simulation.py
+++ b/strands_robots/simulation/mujoco/simulation.py
@@ -1902,10 +1902,45 @@ class MuJoCoSimEngine(
         # when the target robot has an active policy (the common case).
         if registered(self._policy_threads, name):
             self._world.robots[name].policy_running = False
-            try:
-                self._policy_threads[name].result(timeout=5.0)
-            except Exception:
-                pass
+            fut = self._policy_threads[name]
+            timeout = self._DEFAULT_POLICY_STOP_TIMEOUT
+            with contextlib.suppress(Exception):
+                # ``result`` raises either the worker's own exception - it has
+                # exited, which is all we needed - or the join timeout. Which
+                # one happened is decided by ``fut.done()`` below rather than by
+                # the exception type, because the two are not distinguishable
+                # that way: ``socket.timeout`` IS ``TimeoutError``, so a policy
+                # server that stops answering raises the same class the join
+                # does. Typing on the class would refuse a removal whose worker
+                # has already exited.
+                fut.result(timeout=timeout)
+            if not fut.done():
+                # The cooperative stop lapsed: the worker is still live (it is
+                # blocked somewhere the stop flag is not read yet - inside a
+                # policy inference call, typically). Keep the tracking entry.
+                # It is the single record every bound on a live worker reads:
+                # the global gate in step 2, ``list_policies_running``, and
+                # cleanup's own bounded join. Deleting it here does not end the
+                # worker, it only makes the worker unobservable - the gate then
+                # admits every later scene mutation (``add_object``,
+                # ``set_timestep``, ``add_robot``) for the rest of the session,
+                # and cleanup's ``executor.shutdown(wait=False)`` runs on the
+                # premise that all policy workers were drained.
+                return {
+                    "status": "error",
+                    "content": [
+                        {
+                            "text": (
+                                f"Robot '{name}' still has a live policy worker after waiting "
+                                f"{timeout:.1f}s for it to stop, so the scene rebuild was refused: "
+                                "it would reallocate the model/data that worker holds. The "
+                                "cooperative stop has been requested and the worker exits at its "
+                                "next control tick - retry action='remove_robot' (until it does, "
+                                "action='list_policies_running' still reports it)."
+                            )
+                        }
+                    ],
+                }
             del self._policy_threads[name]
 
         # Step 2: after stopping our own, there must be no OTHER policy
@@ -1935,7 +1970,9 @@ class MuJoCoSimEngine(
         robot_obj = self._world.robots[name]
 
         # The cooperative stop + no-running-policy gate above guarantee no
-        # PolicyRunner worker is mid-mj_step. The XML round-trip below still
+        # PolicyRunner worker is mid-mj_step: step 1 refuses outright when its
+        # own robot's worker outlived the stop budget, so reaching here means
+        # every worker is done. The XML round-trip below still
         # reallocates model/data, so serialize it under self._lock to exclude
         # the render/recorder daemon (rendering.py reads mjData under the same
         # lock). remove_robot is dispatched WITHOUT the blanket lock (see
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

`remove_robot` previously swallowed the bounded join's `TimeoutError` and deleted the robot's `_policy_threads` entry regardless, so a worker that outlived the 5s cooperative-stop budget was reported as removed while still running - and, worse, became unobservable to every surface that reads the tracked Future (the global scene-mutation gate, `list_policies_running`, and `cleanup`'s bounded join) for the rest of the session. This PR keeps the tracking entry and refuses the rebuild with an actionable error naming the budget and the retry, deciding the lapse by `Future.done()` rather than exception class (correct, since `socket.timeout` IS `TimeoutError`, so a worker that raised on a dead policy server would otherwise be misread as a lapse). The budget is now read from `_DEFAULT_POLICY_STOP_TIMEOUT` instead of a second hardcoded `5.0`, so `remove_robot` and `cleanup` cannot drift. The refusal path returns an error dict per the AGENTS.md tool-contract (never raises), the message is plain ASCII, and a changelog fragment is included.

## What's good

- Verified locally at head `66b5f31d`: all 14 tests in `test_engine_locking_discipline.py` pass; with `simulation.py` reverted to `upstream/main` the 6 new regression tests fail exactly as the PR description claims, and the 3 no-overreach controls (worker that raised `TimeoutError`, worker that stopped in budget, robot with no policy) pass on both trees - the pins are real, not vacuous.
- The pre-existing `test_remove_robot_during_active_policy_returns_and_worker_exits_clean` gets a premise assert (`not worker.done()`) and its `_SlowPolicy` stub now subclasses `Policy`, closing the silent-vacuity hole where the stub died on its first query and the test raced a finished Future.
- Follow-on interactions check out: a refused removal leaves the tracked live Future, so a second `start_policy` on the same robot is still correctly refused by the per-robot gate, and the retry the refusal names succeeds once the worker exits (pinned by `test_the_remedy_the_refusal_names_removes_the_robot`).
- Deterministic tests inject the Future so both join branches need no scheduler luck; the end-to-end `_WedgedPolicy` test earns the same verdict with a real `PolicyRunner` worker parked inside inference.

## Verification suggestions

- `pytest tests/simulation/mujoco/test_engine_locking_discipline.py -q` on the head, then with `git checkout upstream/main -- strands_robots/simulation/mujoco/simulation.py` to confirm the 6-failed/3-passed split reproduces (it did on mujoco 3.8.0 here).

PR #2357, opened 2026-08-16
fix(tools): signal every gr00t service pid and report whether the port was freed

cagataycali:
## Problem

`_stop_service` escalates SIGTERM then SIGKILL over the pids serving a port, in the
`docker exec` branch first and the host `lsof` branch as a fallback. Every one of
those four `kill` invocations used `subprocess.run(..., check=True)`.

The pid list comes from a scan, so a pid can exit in the window between being
listed and being signalled. `kill` then exits non-zero (`kill: (111): No such
process`), `CalledProcessError` is raised, and the sweep aborts: the pids after it
are never signalled and the SIGKILL escalation never runs. That pid exiting on its
own is the outcome teardown wanted, so the one process that needs no help is the
one that stops the rest from being helped.

The two branches then diverge on how they report it, and the container branch --
which is tried first -- reports the opposite of the truth. There the exception is
caught by `except subprocess.CalledProcessError: continue`, so the container is
abandoned mid-teardown, the host fallback finds nothing for a process that lives
inside a container, and the call returns:

```python
{"status": "success", "port": 5555, "message": "No service running on port 5555"}
```

while the sidecar is still running and still holding the port. `restart` calls
`_stop_service` to free the port and discards the result, so the bind that follows
fails while naming the bind rather than the surviving process.

## Measured

Driven through `_stop_service` with a process table that answers `pgrep`/`lsof`/`kill`
consistently. Pid `111` is listed by the first scan and exits before it is signalled;
pid `222` is the one actually holding the port.

| case | signals delivered | port freed | reported |
|---|---|---|---|
| container, `111` stale + `222` live | `-TERM 111` only | no, `222` alive | `success` -- "No service running on port 5555" |
| host, `111` stale + `222` live | `-TERM 111` only | no, `222` alive | `error` -- `Command '['kill', '-TERM', '111']' returned non-zero exit status 1.` |
| container, `222` cannot be killed | `-TERM 222` only | no | `success` -- "No service running on port 5555" |

After this change all three deliver `-TERM` then `-KILL` to every listed pid; the
first two report `success` because the port really is free, and the third reports
an error naming the port and the surviving pid.

The third row is why `check=False` on its own is not the fix: it stops the sweep
aborting, but an owner that survives both signals is still reported as a stopped
service.

stop-service teardown before and after

The same three cases rendered from each tree's own `_stop_service` output. In `BEFORE`
two of the three report `success` while the port is still held; the SIGKILL escalation
runs in none of them. In `AFTER` every case delivers both signals to every listed pid
and the reported status matches reality in all three.

## Change

`strands_robots/tools/gr00t_inference.py`

- `_signal_service_pids(exec_prefix, pids, signal)` -- one best-effort signal loop
  replacing four `check=True` call sites. A pid that has already exited no longer
  stops the remaining pids or the escalation.
- `_service_pids_in_container()` / `_service_pids_on_host()` -- the scan, previously
  inlined four times as `returncode == 0 and stdout.strip()` plus a split. Each is
  total over subprocess failure at its own boundary (matching `_container_state`), so
  a container that cannot be scanned yields no pids and the caller skips it rather
  than aborting the teardown -- the multi-container resilience `except ...: continue`
  provided, now without the fall-through that lied.
- After the escalation the port is rescanned. `"status": "success"` now means nothing
  is left holding the port; otherwise the result is an error naming the port, the
  surviving pids and where to look.

The existing `stop` behaviour is otherwise unchanged: an ordinary teardown, a free
port, and the fail-soft conversion of an unexpected error into an error tool-result
all return exactly what they did before.

## Tests

`tests/tools/test_gr00t_stop_service_port_teardown.py` (new, 14 cases, every property
pinned on both branches). Against the pre-fix tree: **9 failed, 5 passed**; after:
14 passed.

Representative pre-fix failures:

```
test_a_port_still_held_after_sigkill_is_not_reported_as_stopped[container]
  reported {'status': 'success', 'port': 5555, 'message': 'No service running on port 5555'}
  while pid 222 still holds port 5555. A caller cannot distinguish this from a port
  that was actually freed.

test_a_stale_pid_does_not_stop_the_remaining_pids_from_being_signalled[host]
  pid 222 was never signalled: ['-TERM 111'].

test_a_port_still_held_names_the_port_and_the_surviving_pid[host]
  assert '5555' in "Failed to stop service: Command '['kill', '-TERM', '222']'
  returned non-zero exit status 1."
```

The 5 that pass on both trees are the controls: an ordinary single-pid teardown still
reports success on both branches, a free port still reports success without signalling
anything, and an unexpected error is still converted to an error tool-result.
`[host]` also already reported *an* error for an unkillable owner, so only the
sharper assertion (that the message names the port) fails there pre-fix.

Non-vacuity checked by reverting the escalation in a scratch tree: 8 tests fail,
including the pre-existing `test_stop_escalates_to_sigkill_on_host_when_process_survives_term`.

### Two existing tests adjusted

`test_stop_escalates_to_sigkill_{when_container_process,on_host_when}_..._survives_term`
made `pgrep`/`lsof` answer "still alive" on *every* probe, i.e. they modelled a process
that survives SIGKILL and then asserted `status == "success"` -- the report this change
makes impossible. Their docstring already described the harm ("dropping it would
silently leave a hung server holding the port after `stop` reports success"), so their
intent is the escalation, not the unconditional success. Their stubs now show the
process gone on the rescan, which is what "survives SIGTERM, dies to SIGKILL" actually
looks like. Both still pass unchanged against pre-fix code, so their stated intent is
preserved rather than relaxed; the sabotage check above confirms they still detect a
dropped escalation.

## Docs

`docs/policies/groot.md` -- the container-lifecycle section now states that `stop`
escalates SIGTERM then SIGKILL, that an already-exited process is not a failure, and
that a port still held is reported as an error so a caller can check before rebinding.

## Follow-up (not in this change)

`restart` discards `_stop_service`'s result (`gr00t_inference.py`, the `restart` arm)
and proceeds to `_start_service` regardless. With teardown now reporting truthfully
that is a one-line check, but whether `restart` should refuse to rebind or continue
and let the bind fail is a behaviour choice, so it is left out of this change.

## Local verification

Full `tests/` run on both trees in parallel under identical load (mujoco 3.5.0,
lerobot 0.6.2):

| | failing ids | passed | wall |
|---|---|---|---|
| `upstream/main` | 52 | 30659 | 26m26s |
| this branch | 53 | 30672 | 26m26s |

`comm` of the sorted `FAILED`/`ERROR` ids is empty in the base-only direction. The
one branch-only id is
`tests/test_hardware_task_loop_dispatch.py::TestTheRolloutIsDrivenExactlyOnce::test_on_the_sync_branch`,
a load-sensitive assertion in a rate-guarded loop: it passes 3/3 in isolation on
**both** trees and that module contains no reference to `gr00t_inference`. Counts
reconcile exactly -- `+13` passed and `+1` failed is `+14` collected, the 14 new
tests, with the flake moving one pre-existing test across the line.

`ruff check` / `ruff format --check` / `mypy` over `strands_robots tests tests_integ`
match the pre-existing baseline (18 mypy errors, unchanged, none in the touched
files). No non-ASCII characters added (counts identical to `upstream/main` in every
touched file).

### DIFF
--- a/changelog.d/2357-gr00t-stop-service-port-teardown.md
+++ b/changelog.d/2357-gr00t-stop-service-port-teardown.md
@@ -1,0 +1,14 @@
+### Fixed: `gr00t_inference(action="stop")` signals every service pid and reports whether the port was actually freed
+
+Port teardown escalated SIGTERM then SIGKILL with `subprocess.run(..., check=True)`
+per pid. The pid list comes from a scan, so a pid can exit before it is signalled;
+`kill` then exits non-zero ("No such process") and the raised `CalledProcessError`
+aborted the sweep, leaving the remaining pids unsignalled and skipping the SIGKILL
+escalation. In the `docker exec` branch - tried first - the exception was swallowed
+by a `continue`, the host fallback found nothing for a containerised process, and the
+call returned `{"status": "success", "message": "No service running on port N"}` while
+the sidecar still held the port; `restart` discards that result, so the next bind
+failed while naming the bind rather than the surviving process. Signalling is now
+best effort per pid, and the port is rescanned after the escalation: `"success"` means
+nothing is left holding it, and an owner that survives both signals is reported as an
+error naming the port and the surviving pids.
--- a/docs/policies/groot.md
+++ b/docs/policies/groot.md
@@ -119,6 +119,13 @@ gr00t_inference(action="lifecycle", lifecycle="teardown",
                 remove_volumes=True)  # stop + remove container (and volumes)
 ```
 
+`action="stop"` escalates SIGTERM then SIGKILL over every process serving the
+port, inside the GR00T container first and on the host as a fallback. A process
+that had already exited is not a failure, but a port still held after both
+signals is: the result is then `{"status": "error", ...}` naming the port and the
+surviving pid, because reporting success there would send the next `start` into a
+bind that cannot succeed. Check the status before rebinding the same port.
+
 ## See also
 
 - [Policy providers](../policies/overview.md)
--- a/strands_robots/tools/gr00t_inference.py
+++ b/strands_robots/tools/gr00t_inference.py
@@ -1039,8 +1039,86 @@ def _check_service_status(port: int) -> dict[str, Any]:
         }
 
 
+def _signal_service_pids(exec_prefix: list[str], pids: list[str], signal: str) -> None:
+    """Send ``signal`` to every pid in ``pids``, best effort.
+
+    Teardown is best effort *per pid*. A pid that exits between the scan that
+    listed it and the signal that targets it makes ``kill`` exit non-zero
+    ("No such process") -- and that is the outcome teardown wanted, not a
+    failure. Signalling therefore must never abort the sweep: the pids after it
+    in the list, and the SIGKILL escalation that follows, still have to run.
+
+    Args:
+        exec_prefix: Argv prefix that selects where the signal lands, e.g.
+            ``["docker", "exec", "gr00t"]`` for a container or ``[]`` for the
+            host.
+        pids: Process ids to signal. Empty strings are skipped.
+        signal: ``kill`` signal flag, e.g. ``"-TERM"`` or ``"-KILL"``.
+    """
+    for pid in pids:
+        if pid:
+            subprocess.run([*exec_prefix, "kill", signal, pid], capture_output=True, text=True, check=False)
+
+
+def _service_pids_in_container(container_name: str, port: int) -> list[str]:
+    """Return the inference-service pids serving ``port`` inside a container.
+
+    Args:
+        container_name: Container to scan.
+        port: Port the inference service was started on.
+
+    Returns:
+        The matching pids, or an empty list when the container has none or
+        cannot be scanned at all. A container that cannot be scanned is
+        skipped by the caller rather than aborting the whole teardown.
+    """
+    try:
+        result = subprocess.run(
+            ["docker", "exec", container_name, "pgrep", "-f", f"inference_service.py.*--port {port}"],
+            capture_output=True,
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Makes `_stop_service` teardown best-effort per pid and truthful about the outcome. The four `check=True` kill call sites are replaced by one `_signal_service_pids` loop with `check=False`, so a pid that exits between the scan and the signal no longer aborts the sweep or skips the SIGKILL escalation; the scan itself is factored into `_service_pids_in_container` / `_service_pids_on_host`, each total over subprocess failure at its own boundary. After the escalation the port is rescanned, so `"status": "success"` now means the port is actually free, and an owner surviving both signals is reported as an error naming the port and pid instead of the prior `"No service running"` false success on the container branch. All touched surfaces are private (`_`-prefixed, not exported), no new untrusted input reaches subprocess (`port` is int-typed and numeric-validated at dispatch; pids come from `pgrep`/`lsof` into argv), and the fail-soft `except Exception -> error tool-result` contract is unchanged.

## What's good

- The measured table in the description reproduces: the new suite is 9 failed / 5 passed against the pre-fix tree and 14/14 on the head, so the tests are non-vacuous and the 5 controls pin the unchanged paths.
- The two adjusted escalation tests still pass unchanged against pre-fix code, so their stated intent (the SIGKILL escalation) is preserved rather than relaxed.
- Changelog fragment present (`changelog.d/2357-...md`), docs updated to state the new error semantics, no host paths or non-ASCII in the new code, and the new helpers carry per-parameter `Args:` docstrings.
- Scope discipline: the `restart`-discards-result behaviour choice is explicitly deferred rather than smuggled in.

Issue #239, opened 2026-05-27
security(mesh): R2/R3 design follow-ups for security.py — host normalisation, HF defaults, policy_provider coupling


cagataycali:
Tracker for design-level concerns raised in PR #223 (R2 + R3) that are not in-scope for the PR-2 surgical landing but should be revisited.

## Context

PR #223 ships `strands_robots/mesh/security.py` as a self-contained payload-validation module (PR-2 of 9 in #219). Reviewer (yinsong1986) raised three recurring design concerns across R2/R3 that the validator's current contract is well-documented but creates downstream footguns. These are intentionally deferred so PR #223 lands as a leaf module without consumer-side coupling, but they're worth tracking so they don't drop off the radar.

## A. Host/server_address normalisation footgun

**Where:** `security.py` lines ~640 (`policy_host`) and ~711 (`server_address`).

**Current contract:** validator strips/casefolds internally for membership check, but stores the caller's verbatim string on `out`. Pinned by `test_policy_host_whitespace_preserved`.

**Footgun:** A future consumer in `core.py` (PR-6) writing `if cmd["policy_host"] == "localhost": # loopback fast-path` will silently miss whitespace-padded inputs that passed validation. Same pattern repeats for any literal-equality check downstream.

**Options:**
1. Reject leading/trailing whitespace at validator with `ValidationError("policy_host has leading/trailing whitespace")`. Real peers don't send padded values; a peer that does is itself a forensic signal.
2. Emit both `out["policy_host"]` (normalised) + `out["policy_host_raw"]` (verbatim). Forensics get both; consumers default to the canonical form.
3. Status quo + audit consumer call sites in PR-6/7/8 review.

Reviewer thread refs (yinsong1986, PR #223):
- R1 line 641 (whitespace contract documented)
- R2 line 640 (footgun follow-up)
- R3 line 640 ("identity vs schema" framing, AGENTS.md > Review Learnings #86)

## B. HuggingFace allowlist defaults are floor-only

**Where:** `security.py` line ~308 (`_hf_repo_allowlist_cached` always prepends `("nvidia", "huggingface", "lerobot")`).

**Current behaviour:** `STRANDS_MESH_HF_REPO_ALLOW="nvidia/gr00t-n1.5"` does NOT restrict access to that specific repo — the bare-org `nvidia` from defaults still matches everything under that org. The env var is documented as "extending" the default but cannot tighten below it.

[...]

PR #2358, opened 2026-08-17
fix(mesh/security): admit every policy provider the registry can build

cagataycali:
## Problem

`_DEFAULT_POLICY_TYPES` in `strands_robots/mesh/security.py` documents itself as mirroring the provider registry:

> Mirrors the LeRobot policy registry families plus the providers registered in registry/policies.json. **Keep this in sync with the registry** so a provider that `create_policy` can build can also be driven over the mesh `tell()` path and Device Connect.

It had drifted. `registry/policies.json` ships 14 providers; **10 were absent** from the allowlist, and counting the aliases and shorthands `create_policy` also resolves, **22 of the 28 accepted spellings** were absent.

`is_safe_policy_provider` shares that allowlist, `validate_command` requires it for `execute` / `start`, and the three Device Connect drivers gate on it. So a provider this package can build locally was refused on every remote path:

| `policy_provider` | registry resolves to | mesh / Device Connect (before) |
|---|---|---|
| `mock` | `MockPolicy` | accepted |
| `random`, `test` | `MockPolicy` | **refused** |
| `wbc`, `sonic` | `WBCPolicy` | accepted |
| `wbc_gait`, `sonic_gait` | `WBCGaitPolicy` | **refused** |
| `kimodo`, `kimodo_g1`, `text2motion` | `KimodoPolicy` | **refused** |
| `protomotions`, `gtp`, `gtp_g1`, `protomotions_g1` | `ProtoMotionsPolicy` | **refused** |
| `cosmos3`, `c3` | `Cosmos3Policy` | **refused** |
| `moveit2`, `moveit` | `MoveIt2Policy` | **refused** |
| `curobo`, `cumotion` | `CuroboPolicy` | **refused** |
| `lerobot_async` | `LerobotAsyncPolicy` | **refused** |
| `motionbricks`, `motion_bricks` | `MotionBricksPolicy` | **refused** |
| `vera` | `VeraPolicy` | **refused** |
| `remote` | `RemotePolicy` | **refused** |

The sharpest case is `random` / `test`: they are aliases of `mock`, the provider the validator's own refusal message recommends ("set it explicitly (e.g. 'mock' for the noop policy)"), and they were refused while `mock` passed. `sonic` and `lerobot` were already in the set, so the allowlist was always meant to hold alias spellings, not just canonical names.

An operator hitting this has one recourse, `STRANDS_MESH_POLICY_TYPE_ALLOW` -- a broad grant that reopens the gate the allowlist exists to close. That is why this is filed as a defect rather than a wishlist item.

## Change

Resync the literal, and split it into its two vocabularies so which half is owned where is explicit:

* `_LEROBOT_POLICY_FAMILIES` -- the LeRobot policy families accepted as `policy_type`. An external vocabulary this package does not own; unchanged.
* `_REGISTRY_POLICY_PROVIDERS` -- every spelling of every provider in `registry/policies.json`, grouped by policy class so the next provider addition has an obvious home.
* `_DEFAULT_POLICY_TYPES` is their union, so the value and its type are unchanged for every reader.

**Deliberately not derived from the registry at import.** A provider added to `policies.json` should not silently widen a security allowlist; the `wbc` / `sonic` entries set the precedent of landing the allowlist edit alongside the provider, with a comment saying why. What was missing was enforcement, not derivation -- so `tests/mesh/test_policy_type_allowlist_covers_the_registry.py` supplies what the "keep this in sync" comment cannot, computing the expectation from the shipped JSON. The next omission fails CI instead of shipping as a remote-only availability bug.

## Widening the provider vocabulary relaxes nothing else

`validate_command` applies every other gate to an `execute` / `start` payload regardless of which provider it names. Measured on the branch, for **all 14** providers:

| payload | verdict |
|---|---|
| `policy_host="evil.example.com"` | refused |
| `server_address="attacker.example.net:9000"` | refused |
| `pretrained_name_or_path="attacker/backdoor"` | refused |
| `model_path="../../etc/passwd"` | refused |
| `policy_provider="definitely_not_a_provider"` | refused |
| runtime `register_policy("in_process_only", ...)` | refused on the wire |

The last two are the ones that keep the gate meaningful: an unregistered name is what `import_policy_class` would otherwise try to import as `strands_robots.policies.<name>`, and runtime registration grants a local capability, never a remote one. All six are pinned as tests.

## Tests

`tests/mesh/test_policy_type_allowlist_covers_the_registry.py`, 11 tests. Against `upstream/main`: **3 failed / 8 passed**; on this branch **11 passed**.

The three failures are behaviour-level and name the gap:

```
test_every_registry_spelling_is_allowlisted
  AssertionError: 22 spelling(s) resolve to a provider create_policy can build but are
  refused by the mesh and Device Connect gate: 'c3' -> 'cosmos3', 'cosmos3' -> 'cosmos3',
  'cumotion' -> 'curobo', ... 'random' -> 'mock', ... 'test' -> 'mock', ...

test_every_canonical_provider_passes_the_wire_validator
  AssertionError: validate_command refused registered provider(s): ["cosmos3 (...)", ...]

test_an_alias_is_accepted_wherever_its_canonical_name_is
  AssertionError: alias spelling(s) refused while their canonical provider is accepted:
  ["'random' (alias of 'mock')", "'test' (alias of 'mock')"]
```

The eight that pass on both trees are the controls: the four hostile-field parametrizations above, the unregistered-provider refusal, the runtime-registration exclusion, a pin that the allowlist's excess over the registry is exactly the LeRobot families (so resyncing cannot smuggle in a stranger), and a premise assert that the registry is large enough for the coverage checks to mean anything.

The same command through the real `Mesh._exec_cmd` receiver path, driven under `upstream/main` and under this branch. Nothing is mocked and no source differs between panels other than the allowlist:

```python
mesh._exec_cmd({"command": {"action": "execute", "instruction": "wiggle the arm",
                            "policy_provider": "random", "duration": 3.0,
                            "robot_name": "so101"}, "sender": "operator"})
```

mesh execute with policy_provider=random, before and after

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-mesh-policy-allowlist/mesh-allowlist.mp4)

|---|---|---|
| verdict | `rejected invalid cmd ... policy_provider='random' not in allowlist` | accepted |
| control ticks executed | **0** | **150** |
| max joint travel | **0.000 rad** | **1.473 rad** |
| frames showing motion | 0 of 59 | 49 of 59 |

The left panel is genuinely static -- its per-frame pixel delta is exactly `0.0` because the policy never started. `random` resolves to `MockPolicy`, the same class `mock` resolves to, and `mock` is accepted on both trees.

## Scope

Only the provider half is resynced. The LeRobot-family half names an external, version-varying vocabulary that a literal cannot track and that a guard cannot check without importing an optional dependency; it is left exactly as it was.

Whether `policy_provider` and `policy_type` should keep sharing one allowlist and one env var is a separate design question (#239 bucket C) and is untouched here -- this PR only makes the shared default match the registry it already documented itself as mirroring.

## Local verification

Full `tests/` on both trees (same interpreter, run in parallel), diffing sorted test ids:

* base-only failures are **exactly** the three pre-fix failures listed above -- a full-suite confirmation of the split.
* branch-only failures: **none**. Two ids appeared only on the branch run, `test_async_rtc_masks_inference_latency` and `TestTheRolloutIsDrivenExactlyOnce::test_on_the_nested_branch`; both are wall-clock assertions and both were re-run isolated on each tree, where the branch passed 3/3 and `upstream/main` itself failed one of them under the same machine load. Not caused by this change.

`ruff check`, `ruff format --check` and `mypy` clean over `strands_robots tests tests_integ`.


yinsong1986 (approved):
## Summary

Resyncs the mesh/Device Connect `policy_type` allowlist in `strands_robots/mesh/security.py` with `registry/policies.json`, whose "keep this in sync" contract had drifted: 22 of the 28 spellings `create_policy` resolves (including `random`/`test`, aliases of the very provider the refusal message recommends) were refused on every remote path while working locally. The change splits the literal into `_LEROBOT_POLICY_FAMILIES` and `_REGISTRY_POLICY_PROVIDERS` and unions them, so `_DEFAULT_POLICY_TYPES` keeps its type and semantics for every reader, and adds a guard test that computes the expectation from the shipped JSON so the next omission fails CI. I verified the diff independently: the 28 spellings match `registry/policies.json` exactly (no strangers, no overlap with the LeRobot-family half), all 11 new tests pass on the branch and the claimed 3 fail on `main`'s `security.py`, and the strict per-action key allowlist in `validate_command` means the widened vocabulary relaxes nothing else -- `policy_host`, `server_address`, `pretrained_name_or_path`, and `model_path` remain gated for every provider, an unregistered name is still refused, and runtime `register_policy` still grants no wire capability.

## What's good

- The literal-plus-guard-test design is the right call for a security allowlist: derivation at import would let a registry edit silently widen the gate, while the test makes drift a CI failure instead of a remote-only availability bug.
- Controls are pinned on both trees (hostile-field refusals, unregistered-provider refusal, runtime-registration exclusion, and the excess-is-exactly-the-LeRobot-families pin), so the resync cannot smuggle in a name that is neither a registry spelling nor a declared family.
- Changelog fragment per AGENTS.md step 3; lint and format clean on both changed source files; no non-ASCII or host paths in the new test.

## Verification suggestions

- `python3 -m pytest tests/mesh/test_policy_type_allowlist_covers_the_registry.py --no-cov` on the branch (11 pass), then `git checkout upstream/main -- strands_robots/mesh/security.py` and re-run to confirm the 3 behaviour-level failures name the drift.
- Spot-check the union has no collision: `python3 -c "from strands_robots.mesh.security import _LEROBOT_POLICY_FAMILIES as f, _REGISTRY_POLICY_PROVIDERS as p; print(f & p)"` should print `frozenset()`.

PR #2360, opened 2026-08-17
fix(simulation/mujoco): carry ctrl, act, an unnamed freejoint and the clock across a robot eject

cagataycali:
## Problem

`remove_robot` rebuilds the whole scene. `eject_robot_from_scene` compiles a fresh model and allocates a fresh `MjData`, so **every buffer starts at its reset value** and the survivors' state has to be carried across that gap by name. Only `qpos`/`qvel` were, and only for **named** joints, so three things a surviving robot needs were dropped:

| dropped | consequence | measured |
| --- | --- | --- |
| `ctrl` | reads back as zero, so the next `mj_step` drives every actuator of every **surviving** robot toward its zero configuration | a bystander holding `0.9000 rad` was at `0.0000 rad` 400 steps after some *other* robot was removed |
| `act` | a stateful actuator's internal activation - its effective command - discharges, so it has to charge back up from nothing | a `dyntype="filter"` actuator charged to `0.424703` came back at `0.000000` |
| an unnamed `<freejoint/>` | invisible to a joint-name lookup, so a surviving mobile base is re-seated at its spawn pose with velocity zeroed | `go2` driven to `(0.85, -0.35, 0.42)` came back at `(0, 0, 0.445)` |
| `data.time` | reset to 0 while `world.sim_time` kept counting | physics clock `0.0000s`, `get_state` reporting `0.8020s` |

All four are measured through the public API only (`add_robot` / `send_action` / `step` / `remove_robot`).

Each failed **silently**: `remove_robot` returned `"success"` and an immediate read-back looked correct, because nothing had stepped yet. The pose was carried over; only the thing *holding* the pose was not.

An unnamed `<freejoint/>` is the standard MJCF floating-base idiom -- it is what `unitree_go2` and `lekiwi` ship -- so every mobile base in the registry is affected, and the servo case affects every position-controlled robot in any multi-robot scene.

## Why this is the contract

The same composition in the other direction already guarantees all three, and delivers them. `add_robot`:

> That preservation covers the scene's DYNAMIC state, not just its contents: an arm already in the world keeps the pose it is in **and the actuator setpoints holding it there**, an object stays where it settled or was carried to, a latched `apply_force` wrench persists, and **the clock keeps counting**.

`save_state` also moved from `mjSTATE_FULLPHYSICS` to `mjSTATE_INTEGRATION` specifically because the former "silently excluded `ctrl` ... so the first step after a restore drove toward the pre-restore targets". And `_recompile_preserving_state` already names the unnamed-`<freejoint/>` hazard in detail, for the recompile path; the eject path re-opened it.

So this is not a new guarantee, it is the eject path being brought to the one the neighbouring paths state. Each preservation test below is paired with the **add-path measurement** rather than a hand-picked constant, so the two directions of one composition are pinned to agree.

## Change

`_snapshot_joint_state` / `_restore_joint_state` become `_snapshot_scene_state` / `_restore_scene_state` over a `_SceneState` record, since they no longer carry joint state alone (`qfrc_applied` comes along too: it is dof-indexed, the same key space as `qvel`).

A joint is keyed by `("joint", name)`, or by `("body", body name)` when it is an unnamed **free** joint. The owning body identifies it *exactly*: the compiler refuses a free joint alongside any other joint on one body (`more than 6 dofs in body`), so such a body carries exactly one joint. The two namespaces are kept apart in the key because MuJoCo's joint and body names are independent -- a joint may be named exactly like some body, so a flat string key could match a real joint against a base's body-derived handle. A joint that no namespace can single out (an unnamed *hinge*) is logged at debug rather than guessed at.

## Tests

`tests/simulation/mujoco/test_remove_robot_preserves_dynamic_state.py`, against `upstream/main`: **5 failed, 3 passed**; after: **8 passed**.

```
FAILED ...::test_a_bystander_does_not_sag_after_a_sibling_is_removed
  the bystander was holding 0.9000 rad and drifted to 0.0000 rad over 400 steps
  after an unrelated robot was removed - its setpoint was dropped, so the servo
  drove it toward zero
FAILED ...::test_the_base_keeps_its_pose_and_velocity
  the base was at [0.8, -0.4, 0.35] and came back at [0.0, 0.0, 0.2]
  (its spawn pose is [0.0, 0.0, 0.2])
FAILED ...::test_the_physics_clock_matches_the_reported_clock
  the physics clock read 0.0000s after the rebuild while get_state reports 0.8020s
FAILED ...::test_a_hinge_sharing_a_bodys_name_keeps_its_own_state
FAILED ...::test_the_activation_is_carried_across_the_rebuild
  the actuator's activation was 0.424703 and came back 0.000000 after an
  unrelated robot was removed
```

The **3 that pass on both trees** are the controls: the two add-path references (which is what makes the assertions a symmetry requirement), and "the removed robot leaves no state behind" -- the snapshot still holds the ejected robot's entries, so they must be dropped on the way back in rather than written into whatever now occupies those indices.

The existing per-joint-type width round-trip and the width-mismatch skip in `test_scene_ops_guardrails.py` are carried over to the new names and keys, extended to the third dof-width buffer.

Full `tests/` on both trees (same interpreter, run in parallel): **zero new failing test ids** on this branch. The sorted failing-id sets differ by exactly the 5 pre-fix failures above, which is a full-suite confirmation of the split. `ruff check` / `ruff format --check` / `mypy` over `strands_robots tests tests_integ` clean against the pre-existing baseline; `tests/simulation` 10802 passed.

Two `so101` arms hold a pose; `remove_robot('doomed')` is called, and the camera watches the arm that was **not** removed. One capture script, run under `PYTHONPATH` of an `upstream/main` worktree and of this branch -- the frames are identical up to the removal (mean abs pixel delta **exactly 0.000**), and `14.368` at the final frame.

surviving arm sags vs holds

[servo_sag.mp4](https://github.com/cagataycali/robots/blob/media-remove-robot-dynamic-state/servo_sag.mp4) -- joint-2 error from setpoint after the removal: **1.3800 rad (79 deg)** before, **0.0003 rad** after.

The unnamed-`<freejoint/>` half, on `unitree_go2` (amber marker records where the base was driven to; panels 2 and 3 both had the same unrelated arm removed, so only the base pose differs between them, mean abs delta `8.012`):

floating base teleports vs stays

## Scope

A latched `apply_force` wrench is also dropped here -- but it is dropped by `add_robot` too (measured: one non-zero `xfrc_applied` row before, zero after), so that is a scene-composition gap on both paths rather than something the eject path uniquely breaks, and it wants its own change. Every item above is something the add path demonstrably preserves and the eject path demonstrably lost.

### DIFF
--- a/changelog.d/2360-remove-robot-preserves-dynamic-state.md
+++ b/changelog.d/2360-remove-robot-preserves-dynamic-state.md
@@ -1,0 +1,3 @@
+### Fixed
+
+`remove_robot` now carries a surviving robot's actuator setpoints (`ctrl`/`act`), an unnamed `<freejoint/>` base pose (the floating-base idiom `unitree_go2` and `lekiwi` ship), and the simulation clock across the scene rebuild. Previously the rebuild's fresh `MjData` left all three at their reset values, so removing one robot let every other robot's position servos drive toward zero on the next step, re-seated a surviving mobile base at its spawn pose, and reset `data.time` while `get_state` kept reporting the old clock - each under a `"success"` result whose immediate read-back looked correct. This matches what `add_robot` already guarantees for the same composition in the other direction.
--- a/strands_robots/simulation/mujoco/scene_ops.py
+++ b/strands_robots/simulation/mujoco/scene_ops.py
@@ -37,6 +37,7 @@ import difflib
 import logging
 import math
 from collections.abc import Callable, Mapping
+from dataclasses import dataclass
 from typing import Any
 
 from strands_robots.simulation.models import SimCamera, SimObject, SimRobot, SimWorld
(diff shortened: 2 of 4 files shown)


yinsong1986 (dismissed):
## Summary

Extends the eject-path state carry in `scene_ops.py` from `(qpos, qvel)`-for-named-joints to the full dynamic state `add_robot` already guarantees: `ctrl`/`act` per named actuator, `qfrc_applied` alongside the other dof-indexed buffers, `data.time`, and an unnamed `<freejoint/>` keyed by its owning body's name (with the joint/body namespaces kept apart in the key so a hinge named like a body cannot alias a base's handle). The restore side keeps the existing width-mismatch skip and extends it to the new buffers, and ejected-robot entries are dropped by name rather than written into shifted indices. Everything touched is a `_`-private helper internal to `scene_ops`, so no public API, wire format, or persisted schema moves. The new test module pairs each preservation assertion with the add-path measurement, names its collision and resurrection edge cases explicitly, and the guardrail tests are carried over to the new keys.

## What's good

- The `("joint", name)` / `("body", name)` key split is justified in a comment and pinned by `TestTheKeyNamespacesDoNotCollide`, so the flat-string-collision hazard it exists to prevent cannot silently regress.
- `_resolve_joint_key` re-verifies the body-carries-exactly-one-free-joint premise on the *new* model rather than trusting the snapshot, and unresolvable joints/actuators are logged rather than guessed at - consistent with the no-silent-defaults rule.
- Changelog fragment under `changelog.d/`, no host paths in tests, `tmp_path`-scoped XML fixtures, and the eject/add symmetry contract is documented in the docstrings that will outlive this PR.

yinsong1986 (approved):
## Summary

Brings the eject path's state carry up to the contract `add_robot` already states and delivers for the same composition: `_snapshot_joint_state`/`_restore_joint_state` become `_snapshot_scene_state`/`_restore_scene_state` over a `_SceneState` record, adding per-named-actuator `ctrl`/`act`, `qfrc_applied` beside the other dof-indexed buffers, `data.time`, and an unnamed `<freejoint/>` keyed by its owning body's name - with the joint/body namespaces kept apart in the key so a hinge named like a body cannot alias a floating base's handle. Everything touched is `_`-private to `scene_ops`, so no public API, wire format, or persisted schema moves, and the snapshot/restore still runs inside `remove_robot`'s existing `self._lock` critical section, so the mjData thread-safety contract holds unchanged.

## What's good

- The restore side fails safe on every mismatch: width-mismatched joint entries and activation-width-mismatched actuators are skipped with a warning rather than force-written, `_resolve_joint_key` re-verifies the body-carries-exactly-one-free-joint premise against the *new* model rather than trusting the snapshot, and a stateless actuator (`act_adr < 0`, empty `act_vals`) skips the activation block without a spurious warning.
- Each preservation test is paired with the add-path measurement of the same quantity, so the two directions of the composition are pinned to agree rather than to hand-picked constants, and the resurrection case (the ejected robot's snapshot entries must be dropped, not written into shifted indices) is pinned explicitly.
- Changelog fragment under `changelog.d/`, `tmp_path`-scoped XML fixtures, no host paths, and the guardrail round-trip tests carried over to the new keys and extended to the third dof-width buffer.

PR #2361, opened 2026-08-17
docs(mesh): document the two prerequisites e-stop recovery depends on

cagataycali:
## Problem

`Mesh.emergency_stop()` latches a lockout on every peer that receives it, and nothing clears it on a timer -- correctly, since an e-stop that expired by itself would not be an e-stop. The only way back is an explicit `resume`, so **every precondition `resume` depends on is a precondition for recovering the fleet at all.**

Two of those preconditions were only discoverable *after* a fleet was already locked out.

**1. The timestamp bounds had no README row.** `_on_safety_resume` refuses an envelope that is stale (older than `STRANDS_MESH_RESUME_FRESHNESS_S`, default 60s) or future-dated (more than `STRANDS_MESH_RESUME_FORWARD_SKEW_S` ahead, default 5s). Neither had an entry in the environment-variable table -- and the `STRANDS_MESH_RESUME_BACKOFF_S` row already cited `STRANDS_MESH_RESUME_FRESHNESS_S` in its description as if the reader could look it up, so it sent readers to a row that does not exist.

**2. `docs/mesh.md` documented how to stop a fleet but not how to resume one.** It presents `sim_a.mesh.emergency_stop()` under "Safety primitive" with no resume call, no mention of the `STRANDS_MESH_OVERRIDE_CODE` requirement, and no mention of the clock prerequisite.

## Measurement

The forward bound is the tighter of the two and its effect is asymmetric, which is what makes this operationally sharp. Measured with the shipped issuer (`Mesh._resume_lockout`) and the shipped receiver (`Mesh._on_safety_resume`), offsetting only the issuer's clock so the receiver under test runs unmodified:

| receiver clock vs operator | resume | fleet after retrying |
|---|---|---|
| in sync | accepted | recovered |
| 5s behind | accepted | recovered |
| **6s behind** | **REFUSED** | **stays stopped** |
| 30s behind | REFUSED | stays stopped |
| 30s ahead | accepted | recovered |
| 59s ahead | accepted | recovered |
| 61s ahead | REFUSED | stays stopped |

The accepting band is `[-5s, +59s]`. A robot **6 seconds** behind the operator reads a correct, correctly-signed resume as future-dated and refuses it:

```
[safety] robot-1: refusing remote resume -- t=1786926505.408729 in future (forward_skew_s=5.0, now=1786926499.4092226)
```

Every retry fails identically, so that robot stays stopped until its clock is corrected or the bound is widened on every peer. The in-source comment on the freshness check notes the operator NTP requirement is "documented in CHANGELOG"; a changelog entry is not somewhere an operator looks while a fleet is stopped.

Measured e-stop recovery band and the documentation gap

## Change

Documentation and tests only -- no behaviour change.

- **README**: rows for `STRANDS_MESH_RESUME_FRESHNESS_S`, `STRANDS_MESH_RESUME_FORWARD_SKEW_S` and `STRANDS_MESH_RESUME_REPLAY_CACHE_MAX`, completing the `_resume_*` family and resolving the dangling citation.
- **`docs/mesh.md`**: a "Recovering from an emergency stop" section beside the `emergency_stop()` call -- the resume call, both prerequisites, the measured asymmetry, and the brute-force cooldown. Placed and worded to match the existing fleet-prerequisite note for the `eclipse-zenoh>=1.6.1` floor ("upgrade every peer in a fleet together"), which is the same class of thing: something that must be true on every peer before it matters.
- **`tests/mesh/test_estop_recovery_prerequisites.py`**: pins the behaviour and the documentation together, so the numbers in the docs cannot drift from the numbers the receiver enforces.

## Tests

`3 failed / 4 passed` before, `7 passed` after.

Failing before:

- `test_every_knob_the_resume_freshness_gate_consults_is_documented` -> `['STRANDS_MESH_RESUME_FRESHNESS_S', 'STRANDS_MESH_RESUME_FORWARD_SKEW_S']`
- `test_the_env_table_names_no_variable_it_never_lists` -> `[('STRANDS_MESH_RESUME_BACKOFF_S', 'STRANDS_MESH_RESUME_FRESHNESS_S')]`
- `test_the_recovery_procedure_is_documented_beside_the_estop_call` -> "mesh.md documents how to stop a fleet but not how to resume it"

Passing on both trees, deliberately -- these are the behaviour tests, and the behaviour is unchanged. They exist so the documented bounds stay tied to the enforced ones: a receiver one second past the forward bound stays locked out, retrying does not help, widening the documented knob restores recovery, and a receiver inside the bound recovers (control).

`test_the_env_table_names_no_variable_it_never_lists` is the general, self-maintaining half: it graded all 62 environment-variable rows and found exactly one dangling citation, the one fixed here. It deliberately does **not** require that every variable the code reads is documented -- 36 mesh variables have no row and which of those deserve one is a separate call. It only requires that a row which *cites* another variable cites one the table lists.

Both guards carry a premise assertion, so a future README reformat that hides the table fails loudly instead of reporting clean.

## Verification

Full `tests/` on this branch and on `upstream/main`, same interpreter, run twice. Diffed on sorted test IDs:

- **Branch-only failures: none.** The only difference in either direction is the three `upstream/main` failures listed above -- this PR's pre-fix split reproduced at full-suite scale. 30724 passed on the branch.

The first pair of runs additionally showed seven `tests/tools/test_lerobot_teleoperate*` failures on the branch, which are unrelated to this change but worth flagging as a pre-existing flake: each fails with `dagger requires lerobot>=0.6.0: the installed lerobot has no 'lerobot.scripts.lerobot_rollout'` -- an `importlib` capability probe answering "absent" for a module that is installed. They pass in isolation on both trees and did not recur in the second pair, and the first pair ran under heavy parallel load, so the probe looks sensitive to `sys.modules` state left by earlier tests in the same process. Nothing here touches that path: the diff is three README rows, a docs section, a changelog fragment and one new test file.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` unchanged at the pre-existing baseline.

## Out of scope

Two adjacent things are deliberately left alone because each is a contract choice rather than a correction:

- **The no-override-code posture.** With no code configured there is no remote resume, which `Mesh.start` already warns about at startup in terms of the exact consequence. The in-source comment states this is a deliberate accepted posture ("we cannot safely auto-generate a code (every peer must agree on it)"), so it is documented here rather than changed. A lockout TTL would specifically be the wrong fix -- an e-stop that clears itself on a timer is not an e-stop.
- **The scope of the resume brute-force throttle.** `_resume_locked_until_mono` is a single process-wide scalar, while the same file applies a named "per-issuer fairness bound" to both the estop and resume replay caches so one issuer cannot starve the others. Because `resume` is reachable as a command during lockout, one peer's repeated bad codes hold the cooldown armed and deny a legitimate operator the recovery path. Keying it per-issuer is not a safe drop-in, though: the only identity available on the command path is the unauthenticated body `sender_id`, so an attacker could rotate it and defeat the rate limit entirely, whereas the safety path has the TLS-bound `wire_zid`. Picking an identity for that throttle needs a decision I did not want to guess at inside a documentation change.

---

## Round 2 -- review feedback

**The `STRANDS_MESH_RESUME_FRESHNESS_S` row named the wrong clock direction, and its remedy could not clear the refusal it described.** Confirmed and fixed.

`_on_safety_resume` applies two independent bounds, and each governs one direction:

| check (`core.py`) | trips when | that means the receiver's clock is | refusal log |
|---|---|---|---|
| `envelope_t > now + forward_skew_s` | operator's `t` leads the receiver's `now` | **behind** the operator | `in future (forward_skew_s=...)` |
| `now - envelope_t > freshness_window_s` | receiver's `now` leads the operator's `t` | **ahead of** the operator | `too old (freshness_window_s=...)` |

The row said a receiver more than 60s *behind* the operator "refuses every resume", then sent the operator to widen `STRANDS_MESH_RESUME_FRESHNESS_S`. A trailing receiver is refused by the forward-skew bound, so that knob changes nothing for it -- guidance that fails exactly when a fleet is already stopped. The band table in this description had it right (`61s ahead: REFUSED`), and `docs/mesh.md` only ever makes the forward-direction claim, so the inversion was confined to that one cell.

**Fix:** the row now names a receiver *ahead of* the operator and points the trailing case at `STRANDS_MESH_RESUME_FORWARD_SKEW_S`.

**Why nothing caught it, and what now will.** The round-1 tests pinned the knob *names* and the *forward-skew* behaviour, but nothing tied either bound to the prose an operator reads -- so the cell could invert with the suite green. Added four tests:

- `test_a_receiver_ahead_of_the_operator_is_refused_as_stale` -- negative issuer offset, asserts the `too old` refusal.
- `test_widening_the_freshness_window_recovers_a_receiver_that_leads` -- the remedy works for the direction it names.
- `test_widening_the_freshness_window_does_not_recover_a_receiver_that_trails` -- the sharp one: `STRANDS_MESH_RESUME_FRESHNESS_S=300` still leaves a trailing receiver refused with `in future`. This is the cost of the inverted row, pinned as behaviour.
- `test_each_timestamp_bound_is_documented_against_the_clock_it_governs` -- doc guard asserting each row names its own direction.

The doc guard was verified to **fail on the pre-fix row** and pass after, so it can actually catch this class of drift rather than merely passing:

```
E  AssertionError: the freshness row must attribute the stale-envelope lockout to a
   receiver whose clock is AHEAD of the operator -- that is the direction
   `now - envelope_t > freshness_window_s` trips on: "... more than this *behind* ..."
```

**Verification.** `tests/mesh/test_estop_recovery_prerequisites.py`: 11 passed (7 + 4 new). Full `tests/mesh/` diffed against the merge-base on sorted test IDs -- **101 failures before, the identical 101 after, zero introduced**; all are pre-existing optional-dependency gaps in the local environment (`ModuleNotFoundError: No module named 'awsiot'` and similar), none in the estop/resume path. `ruff check` clean over `strands_robots tests`; `ruff format --check` clean on the changed test file. Changelog fragment extended to state the per-direction contract. Still documentation and tests only -- no behaviour change.


README.md
```diff
@@ -1126,6 +1126,9 @@ touches ROS 2.
 | `STRANDS_MESH_MAX_PEERS` | Peer registry cap; evicts oldest on overflow | `1024` |
 | `STRANDS_MESH_RESUME_MAX_FAILS` | Failed resume attempts before cooldown engages | `5` |
 | `STRANDS_MESH_RESUME_BACKOFF_S` | Cooldown (seconds) after exceeding resume fail threshold. A value no cooldown instant can be built from -- unparsable, negative, or non-finite like `inf`/`nan` -- falls back to the default, so the throttle both engages and expires (shared with `STRANDS_MESH_RESUME_FRESHNESS_S` / `_FORWARD_SKEW_S`) | `30` |
+| `STRANDS_MESH_RESUME_FRESHNESS_S` | How far in the past a resume envelope's timestamp may be before a receiver refuses it as stale. A receiver whose clock is more than this *behind* the operator refuses every resume and stays locked out, so keep fleet clocks in NTP sync or widen this on every peer | `60` |
```
yinsong1986: [MUST FIX] This row inverts the clock direction the freshness bound governs, and its remedy measurably does not work for the direction it names.

> A receiver whose clock is more than this *behind* the operator refuses every resume and stays locked out, so keep fleet clocks in NTP sync or widen this on every peer

The receiver's check is `now - envelope_t > freshness_window_s` (`_on_safety_resume`), which trips when the receiver's clock is **ahead of** the operator's — the envelope reads *stale*. A receiver *behind* the operator is refused by the **forward-skew** bound (the next row, which already states that direction correctly, and at 5s rather than 60s). Measured against the shipped issuer/receiver at this head:

| scenario | result | refusal log |
|---|---|---|
| receiver 61s **behind** operator | refused | `in future (forward_skew_s=5.0, ...)` |
| receiver 61s **behind**, `STRANDS_MESH_RESUME_FRESHNESS_S=300` (this row's remedy) | **still refused** | `in future (forward_skew_s=5.0, ...)` |
| receiver 61s **ahead** of operator | refused | `too old (freshness_window_s=60.0, ...)` |
| receiver 61s **ahead**, `STRANDS_MESH_RESUME_FRESHNESS_S=300` | **recovered** | — |

Your own PR-description band table has this right (`30s ahead: accepted`, `61s ahead: REFUSED` — freshness governs *ahead*), and `docs/mesh.md` only makes the forward-direction claim, so the inversion is confined to this cell. Fix: swap the direction, e.g. *"A receiver whose clock is more than this **ahead of** the operator reads every resume as stale and stays locked out ..."* (a behind receiver is the forward-skew row's case).

Why this can't be deferred: this PR's sole deliverable is correct e-stop recovery guidance, and this sentence sends the operator of a physically stopped fleet to widen a knob that does not clear the refusal it describes — the exact failure mode the PR exists to close (guidance only discoverable/falsifiable after the fleet is already locked out). It is also invisible to the PR's own drift tests, which pin the knob names and the forward-skew behaviour but not this row's direction claim, so nothing in CI will ever correct it (AGENTS.md > Review Learnings (PR #86) > API Consistency: "Match docstrings to semantics").

yinsong1986 (commented):
## Summary

Documentation + tests, no behaviour change: adds README rows for the three `STRANDS_MESH_RESUME_*` knobs (resolving the dangling `STRANDS_MESH_RESUME_BACKOFF_S` citation), a "Recovering from an emergency stop" section in `docs/mesh.md` beside the `emergency_stop()` call, and `tests/mesh/test_estop_recovery_prerequisites.py` pinning the skew behaviour and the documentation together. I re-verified the PR's claims against the shipped code at this head: the 7 new tests pass, the `tests/mesh/` delta vs `main` is exactly those 7 passes (baseline failures identical on both trees), the recovery example's `send(..., {"action": "resume", ...})` path exists (`_dispatch` -> `_resume_lockout`), the "only `status` and `resume` during lockout" claim matches `_dispatch`'s guard, the doc's log line matches the shipped format verbatim, and the forward-skew asymmetry reproduces exactly as described (receiver 6s behind -> refused as future-dated; widening `STRANDS_MESH_RESUME_FORWARD_SKEW_S` restores recovery). One inline must-fix: the new `STRANDS_MESH_RESUME_FRESHNESS_S` README row inverts the clock direction the freshness bound governs and prescribes a remedy that measurably does not work for the direction it names.

## What's good

- The doc-drift tests are the right shape: premise assertions so a README reformat fails loudly, and the dangling-citation check is general rather than pinned to this one row.
- Changelog fragment present (`changelog.d/2361-estop-recovery-prerequisites.md`), all added lines pure ASCII, no host paths, `ruff` clean on the new test file.
- The out-of-scope section correctly declines to guess an identity for the resume brute-force throttle instead of smuggling a behaviour change into a docs PR.

cagataycali:
Independently reproduced the inversion against the shipped issuer and receiver before changing anything, and extended the table with the two cross-remedy rows. The bounds turn out to partition the two skew directions exactly:

| scenario | result | refused by |
|---|---|---|
| receiver 61s behind operator | refused | `in future` (forward skew) |
| receiver 61s behind, `FRESHNESS_S=300` | refused | `in future` (forward skew) |
| receiver 61s behind, `FORWARD_SKEW_S=120` | recovered | - |
| receiver 61s ahead of operator | refused | `too old` (freshness) |
| receiver 61s ahead, `FRESHNESS_S=300` | recovered | - |
| receiver 61s ahead, `FORWARD_SKEW_S=120` | refused | `too old` (freshness) |

So each direction is cleared by exactly one bound and widening the other is a no-op in both directions - the freshness row sent the operator of a stopped fleet to a knob that leaves it locked out. Both rows saying "behind" was the tell: the bounds partition the space, so one row had to be describing the other's case.

Corrected in `2c8483c7`. `e9d1cd26` covers the mirror case that fix cannot.

## Why a follow-up

`test_each_timestamp_bound_is_documented_against_the_clock_it_governs` pins each row's direction word to the direction that bound catches today. That catches a row edited to the wrong direction, but not the reverse: swap which bound `_on_safety_resume` applies to which sign of skew and the pinned words become wrong again, in the opposite direction, with the suite green. Measured by doing exactly that in a scratch tree - the receiver's two checks swapped, documentation untouched:

| guard | result on the swapped receiver |
|---|---|
| `test_each_timestamp_bound_is_documented_against_the_clock_it_governs` | 1 passed |
| `TestTheDocumentedDirectionIsGradedAgainstTheReceiver` | 4 failed |

Both sides are now derived. Which bound clears which direction is measured against the shipped issuer and receiver, the direction each row claims for its own bound is read out of the row, and the two must agree. The second check follows each row literally - widen the bound it names, in the case it describes, and require the receiver to recover - so a row is also refused for prescribing a knob that does not clear the refusal it names, which is the specific way the freshness row was wrong.

## Also in this commit

`docs/mesh.md` made only the forward-direction claim. It is correct, but a reader who lands on the freshness window has to infer the complementary case, so it now states the split: a receiver ahead of the operator trips the freshness window, one behind it trips the forward bound, and widening the other does not help.

The forward-skew row gains the shape the freshness row already has - what the bound refuses the envelope as, the remedy that clears it, and the sibling bound that owns the opposite direction - so the pair reads as a partition from either row rather than leaving an operator who lands on one assuming it governs both.

## Verification

`tests/mesh/test_estop_recovery_prerequisites.py`: 15 passed. Against the pre-correction tree the direction checks fail 3 of 14 with the refusal log attached, naming both the inverted row and the remedy that does not clear it.

Full `tests/` on this branch and on `main`: 8 failing ids each, identical apart from one rotating case inside `tests/mesh/test_zenoh_transport_security.py`, which passes 9/9 three times in isolation on both trees (it opens real localhost zenoh sessions, so two concurrent suites cross-talk). Passed count 30959 vs 30944, i.e. exactly the 15 tests this file adds. `ruff check`, `ruff format --check` and `mypy` over `strands_robots tests tests_integ` clean, changelog fragment gate clean, added lines pure ASCII.

yinsong1986 (approved):
## Summary

Documentation and tests only, no behaviour change: README rows for the three `STRANDS_MESH_RESUME_*` knobs (resolving the dangling `STRANDS_MESH_RESUME_BACKOFF_S` citation), a "Recovering from an emergency stop" section in `docs/mesh.md` beside the `emergency_stop()` call, and `tests/mesh/test_estop_recovery_prerequisites.py` pinning the skew behaviour and the documentation together. The round-1 direction inversion on the freshness row is fixed at this head: each timestamp bound is now documented against the clock direction it actually governs, and the new `TestTheDocumentedDirectionIsGradedAgainstTheReceiver` tests derive both the row's claimed direction and the bound's enforced direction rather than pinning either to recorded text, so neither can drift alone. I verified the documented claims against `strands_robots/mesh/core.py` on `main` — defaults (60s / 5s / 4096), the `max/4` per-issuer fairness cap, the `in future` / `too old` refusal log strings, the lockout command allowlist (`status` / `resume`), and the startup WARNING for an unset `STRANDS_MESH_OVERRIDE_CODE` all match — and ran the new test file at this head: 15 passed.

## What's good

- The doc guards carry premise assertions (a README reformat that hides the env table fails loudly rather than reporting clean), and the round-2 direction guard was verified to fail on the pre-fix row.
- `monkeypatch.setenv` throughout, no host paths, no non-ASCII in the diff, changelog fragment per AGENTS.md step 3.
- Scope discipline: the two adjacent contract questions (no-override-code posture, throttle identity) are named and deliberately left out rather than guessed at inside a docs change.

PR #2362, opened 2026-08-17
fix(robot): release the robot on Ctrl+C instead of exiting straight out of run()

cagataycali:
## Problem

`Robot(...).run()` is the documented way to bring a robot online as a server. It blocks in `_run_device_connect_foreground` until the operator interrupts it, and that handler ended with:

```python
except KeyboardInterrupt:
    print(f"\nShutting down {peer_id}...", flush=True)
    print(f"{peer_id} stopped.", flush=True)
    os._exit(0)
```

`os._exit` runs no `atexit` hook, no `__del__` and no `finally` block, so the instance's `cleanup()` never ran. Measured end to end on a real `Robot("so101", mode="sim")` whose Device Connect runtime genuinely came online (connected to a Zenoh broker, subscribed to its command topic), sending the process a real `SIGINT`:

| marker | before | after |
|---|---|---|
| `cleanup()` ran | **no** | yes |
| an `atexit` hook ran | no | no |
| printed `"<peer> stopped."` | yes | yes |

`cleanup()` is the only path to the physical devices. It prefers the driver's own `disconnect()`, which is where torque disable and gripper release live -- the contract `tests/test_hardware_cleanup_disconnects.py` already pins, and whose module docstring already names this consequence: *"a script that simply ends left the arm energised at its last commanded position instead of going through the driver's disconnect, where torque disable and gripper release live."* Nothing else reaches them: `cleanup()` is terminal, so no library entry point runs after it, and lerobot's `Robot.disconnect()` is `@check_if_not_connected` and refuses a half-open robot by hand. That guarantee was enforced at the `cleanup()` level and at the `__del__` level, and the one entry point that holds a robot for the life of a process bypassed both.

With a connected one-camera arm, the entire event sequence between the interrupt and process death was two entries:

```
BEFORE   KeyboardInterrupt (operator Ctrl+C)  ->  os._exit(0)

AFTER    KeyboardInterrupt (operator Ctrl+C)  ->  robot.disconnect
         ->  bus.disconnect  ->  camera.wrist_cam.disconnect  ->  os._exit(0)
```

|  | before | after |
|---|---|---|
| `driver.disconnect()` calls | 0 | 1 |
| bus torque-disable writes | none | `[True]` |
| camera `disconnect()` calls | 0 | 1 |
| port holder at `os._exit` | still `arm` | free |

The runner is bound to sim instances as well (`robot.py:423` alongside `:461`), so a simulation's `cleanup()` was skipped too -- including on the failed-bring-up branch, where the built-in mesh has already been stopped and the instance's own resources are all that is left to release.

To be precise about one thing: the serial **port** is released, because the process dies and the kernel closes every descriptor it held. The torque state is not a descriptor. Disabling it is a write on the bus that has to happen *before* the port closes (`MotorsBus.disconnect(disable_torque=True)`: *"can prevent damaging motors if they are left applying resisting torque after disconnect"*), and that write was skipped -- so the arm stayed energised at its last commanded position, outliving the process, while the operator was told it had stopped.

## Why the exit stays abrupt

The obvious fixes -- call `cleanup()` inline before `os._exit`, or return normally and let a `finally` handle it -- both hang. `Robot.cleanup()` drains its task executor with `shutdown(wait=True)`; measured with a wedged work item, that call does not return, `ThreadPoolExecutor` workers are not daemon threads, and `concurrent.futures` registers an interpreter-exit hook that joins them. `tests/test_abandoned_work_item_cannot_hang_exit.py` already measured this shape ("a 45s wall clock had to kill the process"). Either variant turns one operator Ctrl+C into a process that never exits, and an operator who then reaches for `SIGKILL` gets no teardown at all -- strictly worse than the bug.

So the release runs on a daemon thread under a budget, and `os._exit(0)` stays as the guaranteed exit. That is the posture `MuJoCoSimEngine.cleanup` already applies to a wedged policy worker: *"a stale-pointer segfault is the lesser evil than hanging the host process on exit"* -- bound the wait, report, proceed. `_SHUTDOWN_TIMEOUT_S = 5.0` is the same value as its `_DEFAULT_POLICY_STOP_TIMEOUT`, bounding the same decision on the same event.

## The line now claims only what happened

`"<peer> stopped."` is a claim about the teardown, so it follows one. When the release does not finish, the reason is named instead -- real program output:

```
Shutting down so101-arm...
so101-arm is exiting WITHOUT a completed shutdown: cleanup() did not finish within 5s. Devices this process held may not have been released.

Shutting down so101-arm...
so101-arm is exiting WITHOUT a completed shutdown: cleanup() raised OSError: [Errno 5] Input/output error: /dev/ttyACM0. Devices this process held may not have been released.
```

An operator is never told the arm is safe when it may still be holding torque.

## Tests

`tests/test_run_releases_devices_on_interrupt.py`, driving the shipped runner through `test_robot_factory.py`'s existing `_drive_foreground` helper with `test_hardware_cleanup_disconnects.py`'s existing lerobot-shaped driver / bus / camera / port doubles -- no new doubles, no serial port or camera opened, port exclusivity modelled so the consequence is asserted rather than the call.

**8 failed / 6 passed on `upstream/main`, 14 passed after.** The failures each name a real observable:

```
the driver's own disconnect() never ran on Ctrl+C: assert 0 == 1
expected one torque-disabling bus disconnect, got []
OSError: [Errno 16] Device or resource busy: /dev/ttyACM0 held by arm
cleanup() never ran on Ctrl+C: assert 0 == 1
a failed bring-up skipped the teardown entirely: assert 0 == 1
claimed the robot stopped while the teardown was still running
```

One of the 8 is a forward guard rather than a demonstration: a second Ctrl+C arriving inside the release window has no pre-fix behaviour to contradict, because before the budget existed there was no wait to interrupt. Left unhandled it would have been a regression -- the interrupt escapes `os._exit` and unwinds into interpreter shutdown, where `concurrent.futures`' exit hook joins the very executor drain the wait was covering, producing exactly the hang the budget prevents.

The 6 that pass on both trees are the no-overreach controls: the ordinary path still prints `stopped.`, the output is still ASCII, an instance with no `cleanup` (or a non-callable one) is not an error, and -- the one that matters most -- **a teardown that never returns, and one that raises, both still reach the exit**. That last pair is what fails if the release is ever moved inline or into a `finally`.

## Artifact

Ctrl+C shutdown sequence, measured on each tree by driving the shipped `_run_device_connect_foreground` with a connected SO-101 arm. The torque band spans the exit on `main`; after the change it flips to released before the exit. The pose render is real MuJoCo, headless EGL.

Ctrl+C shutdown sequence, before and after

## Scope

Left out deliberately: the `DeviceRuntime` is also never `stop()`ed, so the device never announces going offline. Whether an abrupt disappearance is the intended presence semantics (Zenoh liveliness already covers a vanished peer) is a design question rather than a defect, and belongs in its own change.

## Gate

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` unchanged at its pre-existing count; full `tests/` run on both trees with an identical failing-id set.

### DIFF
--- a/changelog.d/2362-run-releases-devices-on-ctrl-c.md
+++ b/changelog.d/2362-run-releases-devices-on-ctrl-c.md
@@ -1,0 +1,3 @@
+### Fixed
+
+`Robot(...).run()` now releases the robot when the operator presses Ctrl+C. The interrupt handler ended with `os._exit(0)`, which runs no `atexit` hook, no `__del__` and no `finally` block, so the instance's `cleanup()` never ran -- and on hardware that is the only path to the driver's own `disconnect()`, where torque disable and gripper release live. An operator's Ctrl+C left the arm energised at its last commanded position while the process printed `"<peer> stopped."` on the way out. The release now runs first, on a background thread under a 5 s budget so a wedged task executor cannot turn one Ctrl+C into a process that never exits, and the shutdown line claims `stopped.` only when the release actually finished.
--- a/docs/device-connect.md
+++ b/docs/device-connect.md
@@ -32,6 +32,8 @@ r.run()              # serve on Device Connect (blocks until Ctrl+C)
 
 `.run()` stops the auto-started built-in mesh and serves the robot over Device Connect (D2D Zenoh multicast, no broker). Without `.run()` the robot is **agent-controlled** — discovered and invoked remotely via `robot_mesh` / `discover()`.
 
+Ctrl+C releases the robot before the process exits. `.run()` is the only entry point that holds a robot for the life of a process, so the interrupt handler is the only teardown it gets: it calls the instance's `cleanup()`, which on hardware reaches the driver's own `disconnect()` — where torque disable and gripper release live — and closes the motors bus and every camera. The exit itself stays abrupt (`cleanup()` drains the task executor, and a wedged rollout must not turn one Ctrl+C into a process that never exits), so the release runs under a 5 s budget on a background thread. It prints `<peer-id> stopped.` only when the release finished; a teardown that times out or raises is reported as `<peer-id> is exiting WITHOUT a completed shutdown: …` so an operator is never told the arm is safe when it may still be holding torque.
+
 Because the mesh is stopped *for* Device Connect, a bring-up that fails leaves the process on no transport at all — so `.run()` logs the cause (naming the extra when that is what is missing) and prints `<peer-id> is NOT online` instead of announcing a device that is not there. It keeps running, so a broker that comes back is not a lost process.
 
 !!! warning "Secure by default"
--- a/strands_robots/robot.py
+++ b/strands_robots/robot.py
@@ -42,6 +42,7 @@ from __future__ import annotations
 import contextlib
 import logging
 import os
+import threading
 from typing import TYPE_CHECKING, Any, Literal, cast, overload
 
 from strands_robots.registry import (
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes the teardown gap in `Robot(...).run()`: the Ctrl+C handler ended with `os._exit(0)`, which bypasses `atexit`, `__del__` and `finally`, so `cleanup()` -- the only path to the driver's `disconnect()`, where torque disable and gripper release live -- never ran, leaving a hardware arm energised while the process printed `stopped.`. The fix runs the release on a daemon thread under a 5 s budget (matching `MuJoCoSimEngine._DEFAULT_POLICY_STOP_TIMEOUT`, bounding the same wedged-executor hazard) and keeps `os._exit(0)` as the guaranteed exit; the shutdown line now claims `stopped.` only when the release completed, and names the reason (timeout, exception, second Ctrl+C) otherwise. I verified the premise that inline/`finally` teardown would hang (`Robot.cleanup()` drains its executor with `shutdown(wait=True)`), ran the new suite on the PR head (14/14 pass), and confirmed ruff check/format are clean on the changed files. No new public API, no wire-format or persisted-schema change; the one user-visible string (`stopped.`) is preserved on the ordinary path.

## What's good

- The second-Ctrl+C window opened by the budget is handled (`KeyboardInterrupt` caught around the wait) rather than left to unwind into the `concurrent.futures` exit-hook join it would otherwise trigger.
- Failure paths never over-claim: timeout and exception both print `exiting WITHOUT a completed shutdown` with the cause, so an operator is not told a possibly-energised arm is safe.
- Tests reuse the existing lerobot-shaped doubles from `test_hardware_cleanup_disconnects.py` and the shipped `_drive_foreground` helper -- no new doubles, and the port-exclusivity assertion pins the consequence rather than the call. ASCII output is pinned per the project rule.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; failed-bring-up branch is covered too.

PR #1686, opened 2026-07-28
fix(mesh): refuse a teleop command a joint cannot travel that fast

cagataycali:
Closes #1677.

## Problem

`InputReceiver._on_input` bounds each inbound teleop frame on four axes before it reaches `send_action()`: **who** sent it (the subscribed key expression), **how fresh** it is, **how densely** frames may arrive (`STRANDS_MESH_INPUT_MAX_HZ`), and **how large** a single value may be (`MAX_INPUT_VALUE_ABS`). Every one of those judges a frame in isolation. Nothing bounded the distance between consecutive commands for the same joint, so a stream inside all four caps could still reverse a joint full-scale on every frame.

Measured on a MuJoCo Panda follower, stream correctly scoped to its configured leader, publishing at 50 Hz (half the permitted frame rate) with `+/-0.9` on every actuator:

```
frames_received  60
rejected          0
rate_dropped      0
frames sent      60

per-frame delta  1.80 units in ~0.020 s  =>  ~90 units/s
```

90 units/s is roughly **14x the no-load speed of the Feetech STS3215 servos** on an SO-100 class arm (~6.5 rad/s at 12 V). That is the overcurrent / gear-strip trajectory the rate cap already exists to prevent in the time domain - `INPUT_MAX_HZ_DEFAULT`'s own docstring says it is there because "a malicious peer can stream far faster to slam the servos (overcurrent / thermal / gear-strip)". The rate cap bounds how *dense* the stream is; nothing bounded how *far* each frame asks a joint to jump.

A physical leader arm cannot produce that trajectory - its own servos cap it at ~6.5 rad/s. It is the shape a replayed or synthetic stream has.

## Change

A per-joint **slew** bound, in frame units per second:

- `security.input_frame_slew_violation(action, previous, now_mono, min_interval_s, max_slew=None) -> str | None`, following the module's existing `X_error(...) -> str | None` shape. It reports the worst-offending joint rather than the first - by speed, then displacement, then joint name - so the message is deterministic regardless of key ordering.
- `security.merge_slew_baseline(previous, applied, now_mono, ...)` maintains the baseline it measures against: **per joint**, merged on every apply, so a frame carrying a subset of the joints stamps only those and retains the rest. See Round 2 below - the shape of the frames is the sender's choice, so the verdict may not depend on it.
[...]

PR #2363, opened 2026-08-17
fix(teleop): hold a local teleop frame to the same slew bound as the mesh path

cagataycali:
## Problem

`teleoperate(publish=True)` drives a local follower **and**, from the same device's `get_action()` stream, every remote one - the module docstring describes exactly that composition. The mesh receive path bounds each inbound frame's per-joint speed (`input_frame_slew_violation` / `STRANDS_MESH_INPUT_SLEW_ABS`, added in #1686). The local merge+apply loop passed its merged frame straight to `send_action`.

So one device was judged by two different rules inside one call, and the follower physically next to the operator was the unguarded one.

#1686 introduced that bound because an unbounded per-joint step is "the overcurrent / gear-strip trajectory the rate cap exists to prevent in the time domain" - a property of the **follower's servos**, not of the network. It touched `mesh/input.py`, `mesh/security.py` and its tests; `teleop_mixin.py` was never brought along.

## Measurement

Driven through the shipped `teleoperate()` loop at 50 Hz **at the shipped default** (`STRANDS_TELEOP_SLEW_ABS` unset, 500 units/s), in the driver units a leader arm actually streams. Each stream is a scripted `get_action()` sequence; "peak units/s" is measured from what reached `send_action`:

| leader stream | peak units/s reaching `send_action` | `slew_rejected` | session status |
|---|---|---|---|
| SO leader sweep in **degrees**, at the STS3215 no-load max (372 deg/s) | 370.5 | 0 | `success` |
| gripper close over 0.5 s, **range-0-100** (200 units/s) | 200.0 | 0 | `success` |
| Panda sweep in **radians** (6.5 rad/s) | 6.5 | 0 | `success` |
| full-scale 12-bit encoder glitch, `-2048` in one tick | 0.0 (**refused**) | 1 | `degraded` |

The refusal log for the last row:

```
[teleop] frame refused: input frame slew for 'joint1' out of range:
101784.3 > 500 units/s (2048 units in 0.0201s)
```

So every stream a physical leader can produce passes untouched in all three unit systems the repo ships (degrees, range-0-100, radians), and the frame no servo could have produced is the only one dropped. The equivalent frame arriving over the mesh is refused by that path's own narrower default, unchanged by this PR.

## Change

The merged frame is checked against **the same helper the mesh path calls**, but under its own knob and default, because the two paths carry different unit conventions:

- **A separate local default** - `STRANDS_TELEOP_SLEW_ABS`, 500 units/s. The mesh path's 8*pi (25.13) default is radian-only by construction: its companion magnitude clamp refuses degree-valued frames outright, so that path already requires widened env vars for driver-unit devices. The local loop is the *default* path for the shipped lerobot SO hardware, whose arm joints are degrees (`use_degrees=True` is the SO leader/follower default) and whose gripper is `RANGE_0_100` in every configuration. 500 units/s clears the STS3215 no-load max in degrees (~372) and a half-second gripper close (200) while still catching an encoder glitch by two orders of magnitude. The mesh path keeps its own default untouched.
- **Both mesh-helper call sites carry the local bound explicitly.** The checker and the baseline pruner each fall back to the mesh path's radian-scoped defaults when not told otherwise, and a default that reaches either one describes a bound this path does not enforce. The pruner's envelope is unbounded here because the local path applies no magnitude clamp, so no finite displacement makes a baseline entry unable to refuse - the prune becomes a no-op rather than a silent weakening, and the baseline stays bounded by the session's own device key set (reset per session) instead of by a time window.
- **Refuse-and-count, never clamp** - matching every other guard on the mesh path. Clamping toward the commanded value would silently alter an actuator command.
- **Self-healing** - the bound is a speed measured from each joint's last applied value, so the allowance grows while a joint is still and a refused stream resumes by itself. No resync step.
- **Refusals are visible** - counted in a new `slew_rejected` stat, separate from `errors` (nothing failed). But they move the session off `success`: a device whose units the bound does not expect otherwise reported 0 frames, 0 errors and `success` - a silent no-op, which is the outcome that status derivation exists to refuse.
- **Layering preserved** - the helper is imported inside `_teleop_loop`, not at module scope, because the `mesh` package reaches `strands_robots.simulation` and this mixin must not depend on it (`utils.positive_finite_number_error` documents that separation). A test pins it so the import cannot be hoisted.

## Tests

`tests/test_teleop_slew_bound.py`: **18 fail / 4 pass on `main`** -> 22 pass here.

The 4 that pass on both trees are the no-overreach controls (a legitimate stream is untouched; the loop still applies frames). Pinned beyond the headline: a refused frame does not become the next frame's baseline (so the stream resumes with no resync), the refusal is counted rather than silent, the knob still lifts the rule, and the mesh import stays inside the loop function (asserted over the AST, so it cannot be hoisted back to module scope).

`TestDefaultBoundAccommodatesDriverUnits` is the regression class for the default itself, one test per unit system the shipped hardware speaks:

- `test_a_90_degree_sweep_over_1s_is_not_refused` (90 deg/s)
- `test_a_half_second_gripper_close_is_not_refused` (200 units/s, range-0-100)
- `test_sts3215_no_load_max_in_degrees_is_not_refused` (372 deg/s)
- `test_a_2000_units_per_second_glitch_is_still_refused` (the bound still bites)

`TestAQuietDeviceKeepsItsBaselineWhileOthersMove` covers the multi-device case, which no single-device test can reach: one device goes quiet (`get_action()` returns `{}`, the shape of a disconnect or USB re-enumerate) while another keeps the loop applying frames, then returns with a full-scale first read. It must still be refused.

`TestBothCallSitesCarryTheLocalBound` pins the *class* rather than the instance: an AST assertion that every mesh slew helper called from `_teleop_loop` passes the local parameters explicitly, so no future call site can silently inherit a mesh default. Both R1 and R4 were instances of that one shape.

## Scope

Only the half that needs no contract decision. The issue also suggested refusing `teleoperate()` on a real arm when `max_relative_target` is unset - that changes a shipped default and would break existing hardware sessions, so it is left as a maintainer call. `max_relative_target` remains `None` by default; this PR does not touch it.

`docs/hardware/teleoperation.md` gains the bound in the per-tick description and in "How it relates to mesh teleop" - naming `STRANDS_TELEOP_SLEW_ABS` and its 500 units/s default rather than the mesh knob - plus what a quiet device's baseline means on reconnecting. The README row names the knob, its default and that it cannot be disabled.

The same glitching leader driven through the shipped `teleoperate()` loop on `main` and on this branch. Commands were captured from the live loop and replayed into a fresh sim to render the **commanded pose** per tick - that is the quantity the bound governs, and the sim's own position servo travels only 2 ms per 20 ms tick, so rendering its response would understate what reaches the bus.

**Captured at a 25.13 units/s bound** - the mesh default, which was this branch's local default when the capture was made, before round 1 separated the two. It therefore shows the mechanism (refuse, hold, resume with no resync, report `degraded`) at a radian-valued amplitude, not the shipped default's threshold. The shipped default's own accept/refuse boundary is the measurement table above, in the three unit systems that default exists for.

local teleop slew bound before/after

[Full-resolution MP4](https://raw.githubusercontent.com/cagataycali/robots/media-teleop-slew-bound/teleop_slew.mp4)

Left is `main`: at tick 30 the base yaw snaps to `-2.600 rad` at 129.5 units/s and the session still reports `success`. Right is the guarded loop: the frame is refused, the pose holds at `+0.390 rad`, the sweep resumes on the next tick with no resync, and the session reports `degraded` with `slew_rejected=1`.

On the raw renders the two panels are **identical for ticks 0-29** (mean abs delta 0.0011) and diverge only at the glitch (2.09, against a 0.32 median per-tick delta for the legitimate sweep) - the sweep is untouched and only the unreachable frame is dropped.

### Rollout: the R4 case, a device reconnecting

The bypass R4 closed, rendered the same way: one MuJoCo `so101`, the shipped
`teleoperate()` loop at 50 Hz, **two** attached devices. One streams the base joint at
rest, falls silent for 1.0 s (`get_action()` returns `{}`, the shape of a disconnect or
USB re-enumerate) while the other keeps the loop applying frames, then reads back
`-1.8 rad`. Same script both sides; only the library differs.

Captured at `STRANDS_TELEOP_SLEW_ABS=0.6` - a radian-scoped bound, because a sim arm's
joints are radians and `so101`'s base range is +/-1.92 rad, so a displacement large enough
to exceed the shipped 500 units/s default after a 1 s gap does not exist inside that
range. The bound's value is not what R4 was about; the prune horizon's independence from
it is, and the 1.0 s gap outlasts the 0.5 s mesh horizon at either setting.

teleop slew baseline across a reconnect

[Full-resolution MP4](https://raw.githubusercontent.com/cagataycali/robots/media-teleop-slew-reconnect/teleop_slew_reconnect.mp4)

| | `main` + R1..R3 | with R4 |
|---|---|---|
| base joint reached the robot | **-1.80 rad** | 0.00 rad |
| ticks displaced from rest | **20 / 80** | 0 / 80 |
| still displaced at the final tick | **yes** (12.07 mean abs px from rest) | no (0.00) |
| `slew_rejected` | 20 | 20 |
| session status | `degraded` | `degraded` |

The counters are **identical** - which is the point: they cannot tell the two runs apart,
only the pose can. Before R4 the 20 refusals are the arm's *return to rest* being blocked
after the glitch was let through, so the follower is left parked at the bad pose for the
rest of the session; with R4 they are the glitch itself, and the arm never leaves rest.

On the raw renders the two panels are identical for every tick before the reconnect
(mean abs delta **0.0000**) and diverge only at it (12.07), so the panels differ solely
because of this change.

---

## 13. Review rounds

| Round | Concern | Fix commit | Pin test |
|---|---|---|---|
| R1 | The 8*pi default refuses ordinary teleop for the shipped SO hardware defaults, because "a physical leader cannot reach it" holds only in radians while the local hot path speaks driver units (thread `teleop_mixin.py:569`) | `2202fe3` | `tests/test_teleop_slew_bound.py::TestDefaultBoundAccommodatesDriverUnits` |
| R2 | Follow-through on R1: the existing over-speed stimulus (`-2.8`, 140 units/s) sits *under* the raised 500 default, so the suite's own headline tests would have passed with the bound removed | `bf5d9d1` | same file - `GLITCH = -2048.0`, a full-scale 12-bit encoder glitch (101784.3 units/s measured, 200x the default) |
| R3 | Description drift from R1: the Measurement table and the Tests count still quoted the pre-R1 bound, so the PR advertised a refusal the shipped default does not make | this description | Measurement table re-run at the shipped default (knob unset), all four streams; the rollout is now labelled with the bound it was captured under |
| R4 | The baseline pruner ran on the mesh defaults while the check ran at the local bound, so its prune horizon (0.5 s for a joint at rest) was shorter than the window in which that entry can still refuse. A quiet device's baseline was dropped and its full-scale first read on return was applied with no refusal, no log line and no count (thread `teleop_mixin.py:613`) | `69f2f35` | `TestAQuietDeviceKeepsItsBaselineWhileOthersMove` (the reachable bypass) and `TestBothCallSitesCarryTheLocalBound` (the class: an AST pin refusing any mesh slew helper called on mesh defaults from this loop) |
| R5 | Documentation drift from R1: the teleoperation guide still described the local bound as the mesh path's `STRANDS_MESH_INPUT_SLEW_ABS` at 8*pi, in both the per-tick description and the mesh-teleop section, so the documented knob was not the one in force. R4 also left the *narrowed* direction unpinned - tightening the local bound lengthens the window an entry can still refuse in | `c2e057f` | `TestNarrowingTheBoundDoesNotShortenTheBaseline` (fails on pre-R4 source) and `TestTheMeshPathKeepsItsOwnPruneHorizon` (control: the mesh path's own horizon is unchanged) |

R1 and R4 are two instances of one shape - a mesh-scoped default reaching a local call site - so R4 pins the class over the AST rather than only correcting the second call site. Test counts above re-measured against `origin/main` at `69f2f35`.


strands_robots/teleop_mixin.py
```diff
@@ -514,13 +559,27 @@ def _teleop_loop(
...
-                            logger.warning("[teleop] send_action error: %s", txt)
-                    self._teleop_frames += 1
+                    # Refuse-and-count, matching the mesh path: clamping toward
+                    # the commanded value would silently alter an actuator
+                    # command. The bound is measured per joint from that joint's
+                    # last sent value, so the allowance grows while a joint is
+                    # still and a refused stream resumes by itself once the
+                    # commanded pose is reachable safely - no resync handshake.
+                    apply_mono = time.perf_counter()
+                    slew_reason = input_frame_slew_violation(merged, self._teleop_slew_baseline, apply_mono, period)
```
yinsong1986: [MUST FIX] The default bound appears to refuse ordinary human teleop for this repo's own documented lerobot SO hardware defaults, because the "a physical leader arm cannot reach it" premise holds only in radians while the local hot path speaks driver units.

`strands_robots/policies/lerobot_local/embodiment.py` records the unit convention twice: the SO leader/follower default is `use_degrees=True` ("the lerobot so_follower / so_leader DEFAULT ... but it is opt-out"), i.e. arm joints in **degrees**, and the gripper is `MotorNormMode.RANGE_0_100` (**0..100**) in every configuration. `get_action()` frames reach this check in those units (the module docstring above says so: "lerobot leaders already return `{motor}.pos: float`"), and `send_action` forwards them unchanged. Against the default bound of 25.13 units/s:

- STS3215 no-load max, the PR's own 6.5 rad/s figure, is **~372 deg/s** — ~15x the bound, not under it;
- a calm 90° sweep over 1 s is **90 units/s** — ~3.6x;
- a half-second gripper close is **200 units/s** — ~8x, in *every* config including `use_degrees=False`.

So `robot.attach_teleop("so101_leader", port=...).teleoperate()` — the basic local hardware path, which worked at defaults before this PR — would have most in-motion frames refused, the follower effectively slew-limited to ~25°/s (and, because refusal delays rather than smooths, the eventually-accepted frame applies the accumulated displacement as one full-size step). The mesh path has carried this bound since #1686, but that path already requires widened env vars for driver-unit devices (its 12.566 magnitude clamp refuses degree frames outright); this PR is what first imposes it on the default local loop. The PR's MuJoCo Panda measurement is radian-valued, so it cannot see this.

Why this can't be deferred: it is a functional regression on the changed hot path for realistic inputs (frames — actuator commands — dropped for the shipped default hardware configuration), and on merge `slew_rejected` + the refusal semantics become the observable contract of `teleoperate()`, so re-tuning the default afterwards is a second behaviour change. What would resolve it: measure once with a degree-valued SO leader (or the synthetic streams in the Verification suggestions) and either raise the local default above the driver-unit servo maximum (~372 units/s in degrees), scope the default by the units the attached device actually emits, or hold this half for the same explicit maintainer call the PR description already gives the `max_relative_target` half — anything that keeps the default local path working at defaults while preserving the shared-helper design.

yinsong1986 (commented):
## Summary

Extends the per-joint slew bound (`STRANDS_MESH_INPUT_SLEW_ABS`, #1686) from the mesh receive path to the local `teleoperate()` merge+apply loop, so one device driving a local and a remote follower from the same `get_action()` stream is judged by one rule. The implementation is careful: it calls the mesh path's own helper rather than copying it (pinned by a spy test), refuses-and-counts rather than clamping, keeps the baseline merged per joint and updated only on applied frames, resets it per session, and imports the helper lazily inside `_teleop_loop` with an AST test pinning the layering. I verified `import strands_robots.mesh.security` succeeds without the `[mesh]` extra (zenoh imports are all lazy), so the in-loop import cannot crash a base install. One must-fix concern is flagged inline: the default bound's premise is unit-dependent and appears to break the repo's own documented lerobot SO hardware defaults.

## What's good

- Single source of truth: the local path consults `input_frame_slew_violation` / `merge_slew_baseline` directly, with `TestTheBoundIsTheMeshBoundNotACopy` pinning it, so the two paths cannot drift.
- Refuse-and-count (no clamping), refusals surfaced in `slew_rejected` and reflected in session status, matching AGENTS.md's no-silent-defaults rule.
- Layering discipline: lazy import inside the loop, plus a module-scope-import AST test that fails on the obvious wrong implementation.
- Changelog fragment, docs, and README env-var row all updated in the same PR; tests demonstrate 10 fail / 4 pass on `main`.

## Verification suggestions

- Drive the loop with a stream in the lerobot SO driver's default units before merging, e.g. a `SteppingLeader` emitting a 90-degree arm sweep over 1 s (`[0, 1.8, 3.6, ... 90]` at 50 Hz) and a gripper close `[0 .. 100]` over 0.5 s, and check `slew_rejected` — per the inline comment, both appear to be refused at the default bound.

cagataycali:
### Lint is fixed, and doing so uncovered that R1 is not tunable

Pushed `9783c473`, which does three mechanical things:

- **`ruff format`** on `teleop_mixin.py` and `test_teleop_slew_bound.py` (ruff 0.15.22, inside the pinned range). This is what CI was refusing.
- **Hoists two inline imports** R1 introduced: `import os as _os_slew` inside the method, and `__import__("math")` re-evaluated on every `teleoperate` call. Neither is a documented lazy load - the `mesh.security` import just above them is, and explains why - so both belong at module scope. No behaviour change.
- **Merges `upstream/main`**, which the branch was 7 merged PRs behind, so the tree CI checks is the one that would land.

### The lint failure was masking four real test failures

`Run lint` fails before `Run tests`, so `Run tests` was **skipped** and the branch's actual state never appeared in CI. Running the suite locally:

| commit | result |
|---|---|
| `fc663846` (before R1) | **14 passed** |
| `2202fe3a` (R1) | **4 failed**, 14 passed |

R1 added 4 tests and broke 4 existing ones. The failures are in R1's own file:

```
TestAnOverSpeedFrameIsNotApplied::test_a_full_scale_jump_never_reaches_send_action
  AssertionError: the glitched frame was applied: [0.0, 0.0, -2.8, ...]
TestAnOverSpeedFrameIsNotApplied::test_the_refusal_is_counted_and_reported_not_silent
TestRefusalsAreVisibleInTheSessionStatus::test_a_partially_refused_session_is_degraded
  assert 'success' == 'degraded'
TestRefusalsAreVisibleInTheSessionStatus::test_the_live_status_surface_reports_refusals
```

### This is not a tuning problem - the constraint set is unsatisfiable

Raising `_LOCAL_SLEW_DEFAULT` to 500 did not pick a bad number. With a **single absolute bound** `B` in units/s, the file now asks for both of these at once:

| test | requirement on `B` | source |
|---|---|---|
| `test_a_full_scale_jump_never_reaches_send_action` | `B < 140` (refuse `-2.8` in one 1/50 s tick, the test's own arithmetic) | pre-existing |
| `test_sts3215_no_load_max_in_degrees_is_not_refused` | `B >= 372` (STS3215 no-load max, 6.5 rad/s expressed in degrees) | added by R1 |

There is no `B` satisfying `B < 140` and `B >= 372`. No value of `_LOCAL_SLEW_DEFAULT` makes this file green, so the remaining failures cannot be cleared by adjusting the constant.

The missing input is the **unit system**, not the magnitude. `2.8` is a glitch because it is radians (140 rad/s is not a physical arm speed); `372` is ordinary because it is degrees (= 6.5 rad/s). One scalar cannot separate them because the two unit systems overlap numerically - that is exactly why the mesh path's `8*pi` is radian-scoped and documented as requiring explicit widening for driver-unit devices.

### Not guessing the resolution

Which way to break the tie is a contract choice about what a local teleop frame speaks, so I have not picked one. The shapes I can see:

1. **Bound per unit system** - derive the limit from the attached teleoperator's units, so a radian stream keeps a radian-scoped bound and a degree stream gets a degree-scoped one. Needs the mixin to know each stream's units.
2. **Normalize per joint** - express the bound as a fraction of each joint's own range per tick, which is unit-agnostic by construction.
3. **Declare one unit system** for the local path and require callers to convert, keeping a single bound and reverting R1's widening. This would mean the STS3215-in-degrees case is out of scope rather than accepted.

Option 3 restores green immediately (it is `fc663846`); options 1 and 2 change the public surface. Happy to implement whichever you prefer - flagging it rather than choosing, since the previous attempt to settle it by moving the constant is what produced the contradiction above.

cagataycali:
### Verified green on `bf5d9d1b`, and R2 resolves it correctly

Ran the suite against R2 on a clean tree: **204 passed** (`test_teleop_slew_bound.py`, `test_teleop.py`, `test_teleop_rate_and_duration_guards.py`, `mesh/test_input_slew_bound.py`), `ruff check` and `ruff format --check` both clean over `strands_robots tests tests_integ`. The 4 failures from my previous comment are gone.

R2 picks a better resolution than any of the three I sketched, and it is worth naming why: it fixes the **stimulus**, not the bound. The old `-2.8` was implicitly radian-valued, so it only read as a glitch under a radian assumption the local path never made. `GLITCH = -2048.0` - one frame of a full-scale 12-bit encoder read - is a glitch in the units a leader arm actually streams, at 102400 units/s. That dissolves the `B < 140` half of the contradiction rather than trading it away, so the constraint set is satisfiable now instead of merely re-tuned.

One residual question for the record, since it is the last thing the unit ambiguity touches: with the local default at 500 units/s, a local stream that *is* radian-valued would have no effective glitch protection (500 rad/s is past any physical motion). If local teleop streams are always driver-unit by construction, that is a non-issue and the widened default is exactly right - worth a sentence in `docs/hardware/teleoperation.md` stating the local path's unit domain explicitly, so the next person sizing a stimulus does not have to infer it.

My `9783c473` underneath is unchanged and mechanical: `ruff format`, the two inline imports hoisted to module scope, and the `upstream/main` merge that brought the branch up from 7 PRs behind.

strands_robots/teleop_mixin.py
cagataycali: Fixed by taking the first of the three options: the local path gets its own default, raised above the driver-unit servo maximum, rather than inheriting the radian-scoped one.

`STRANDS_TELEOP_SLEW_ABS`, default **500 units/s**, now bounds the local frame; the mesh path keeps `STRANDS_MESH_INPUT_SLEW_ABS` at `8*pi` unchanged. The split is the point: a network-facing stream and a leader arm on a USB port have different trust assumptions and, more concretely here, different units, so one number could not be right for both. 500 clears every case in the report with margin -- 90 units/s calm sweep, 200 units/s half-second gripper close, ~372 deg/s STS3215 no-load max -- while still refusing the encoder-glitch class the bound exists for, which lands above 1000 units/s.

The correction I owe on my own reasoning: the premise was measured in the wrong unit system. The MuJoCo Panda I measured is radian-valued, so the bound looked unreachable by a human arm; the shipped default is `use_degrees=True` with a `RANGE_0_100` gripper, where the same physical motion is a much larger number. Measuring a per-unit-time bound against a sim's unit convention when the default hardware speaks another is what let a functional regression look like a safety bound.

Pinned by `TestDefaultBoundAccommodatesDriverUnits`, 4 cases, all in driver units rather than radians so the pin fails if the local default is ever narrowed back toward the radian figure. The R2 commit resizes the glitch stimulus to the raised bound, so the refusal half is still asserted against something the new default actually rejects rather than passing by luck.

strands_robots/teleop_mixin.py
```diff
@@ -514,13 +589,29 @@ def _teleop_loop(
...
+                        if self._teleop_slew_rejected <= 5:
+                            logger.warning("[teleop] frame refused: %s", slew_reason)
+                    else:
+                        result = self.send_action(merged, robot_name=robot_name)
+                        if isinstance(result, dict) and result.get("status") == "error":
+                            self._teleop_errors += 1
+                            if self._teleop_errors <= 5:
+                                txt = result.get("content", [{}])[0].get("text", "")
+                                logger.warning("[teleop] send_action error: %s", txt)
+                        self._teleop_slew_baseline = merge_slew_baseline(self._teleop_slew_baseline, merged, apply_mono)
```
yinsong1986: [MUST FIX] `merge_slew_baseline(self._teleop_slew_baseline, merged, apply_mono)` is called with the **mesh** defaults while the violation check on line 599 runs at `max_slew=_local_slew` (500). The pruner drops a baseline entry once `(value_abs + |value|) / max_slew` seconds have elapsed, on the documented premise that "dropping such an entry changes no verdict" — and that premise only holds when the pruner and the checker share parameters. With its defaults the horizon is `(4π + |v|) / 8π`: **0.5 s for a joint resting near 0**, ~4.1 s at 90 degrees. Under the local 500 units/s bound, that same entry could still refuse a full-scale glitch for ~4 s (2048/500), so entries are pruned while they can still refuse frames.

Reachable path: in a multi-device session, a device that goes quiet (returns `{}` — a disconnect or USB re-enumerate) while another device keeps applying frames has its joints' baselines silently pruned after 0.5 s; when it returns with a full-scale first read — the exact USB re-enumerate case this PR's description and docstring name — `input_frame_slew_violation` finds no entry for the joint and the glitch frame is **applied with no refusal, no log line, and no `slew_rejected` count**. The baseline comment in `_ensure_teleop_state` promises the opposite ("a device whose frame carries a subset of the joints cannot erase the baseline of the ones it omits"); the mesh-scale prune performs exactly that erasure. Note this cannot be caught by the current suite: every test drives a single device that stamps all its joints on every applied frame, so the prune never fires.

This blocks merge because it is a silent failure of the safety property the PR ships — the overcurrent/gear-strip frame the bound exists to refuse passes unguarded, with the session still reporting `success`. The fix is one call site: pass `max_slew=_local_slew` (and a `value_abs` consistent with the local path — it runs no magnitude clamp, so `math.inf` or a driver-unit envelope) so the prune horizon is computed against the bound actually being enforced. The same mismatch also bites when an operator *narrows* `STRANDS_TELEOP_SLEW_ABS` below the mesh default: the checker tightens but the pruner keeps discarding at the looser mesh horizon.

yinsong1986 (commented):
## Summary

Extends the per-joint slew bound from the mesh receive path (#1686) to the local `teleoperate()` merge+apply loop, so one leader driving a local and a remote follower from the same `get_action()` stream is judged by one rule on both. Round 1's concern is addressed as advertised: the local path now carries its own knob (`STRANDS_TELEOP_SLEW_ABS`, default 500 units/s) sized for the driver units the shipped SO hardware actually streams (degrees, range-0-100), while the mesh path's radian-scoped `STRANDS_MESH_INPUT_SLEW_ABS` is untouched, and `TestDefaultBoundAccommodatesDriverUnits` pins the accept/refuse boundary in all three unit systems. The design choices are right: refuse-and-count rather than clamp, a `slew_rejected` stat that is not an error but still moves the session off `success`, a lazy in-loop import of the mesh helper with the layering pinned over the AST, README/changelog-fragment/env-var conventions all followed. One blocking finding, inline: the baseline pruner (`merge_slew_baseline`) is still called with the mesh path's default parameters while the violation check runs at the local bound, which opens a silent bypass window for exactly the glitch class this PR exists to refuse.

## What's good

- Single source of truth: the loop calls the mesh path's own helper, and a spy test pins that it cannot be replaced by a drifting copy.
- The regression suite is behavioral and adversarial in both directions: the glitch stimulus (`-2048`, ~101784 units/s) was deliberately raised in R2 so the headline tests fail if the bound is removed, and no-overreach controls pin that legitimate streams pass untouched.
- Status derivation refuses the silent no-op: a wholly-refused session cannot report `success`.
- Changelog fragment, README env-var row, and the lazy-import layering rule (with its AST pin) all follow AGENTS.md conventions.

strands_robots/teleop_mixin.py
cagataycali: Fixed in `69f2f35` -- both call sites now pass `max_slew=_local_slew` and `value_abs=math.inf` explicitly (the envelope is unbounded because this path applies no magnitude clamp, so no finite displacement makes an entry unable to refuse; the prune becomes a no-op). Pinned by `TestAQuietDeviceKeepsItsBaselineWhileOthersMove` (reproduces the multi-device bypass) and `TestBothCallSitesCarryTheLocalBound` (AST guard refusing any mesh slew helper called on mesh defaults from this loop).

---
<sub>This contribution was made with assistance from an AI coding tool.</sub>

cagataycali: Confirmed and fixed in `69f2f35`, and the reachable path reproduces exactly as described. Driving two devices where one goes quiet for 30 ticks (0.6 s, past the 0.5 s mesh-default horizon for a joint at rest) while the other keeps the loop applying frames, the returning device's full-scale read was applied **ten times** with `0 errors` and the session text `Teleoperation completed: 40 frames, 0 errors, 49.7Hz over 0.8s.` - no refusal, no log line, no count. That is the silent failure, verbatim.

On the envelope, I took `math.inf` rather than a driver-unit number, and the reason is the prune's own safety premise. It drops an entry once `(value_abs + abs(value)) / max_slew` seconds have elapsed, justified by "dropping such an entry changes no verdict" - which requires `value_abs` to be the furthest a *permissible* command may reach. On the mesh path that is real: `validate_input_frame` clamps magnitude, so the horizon is finite and the prune is genuinely free. The local path runs no such clamp - `input_frame_slew_violation` takes no envelope and nothing else bounds a leader's reach - so **no finite displacement exists after which an entry provably cannot refuse**. Any finite envelope I picked would re-introduce the same bug at a different threshold, just further out. Unbounded is the only value under which the prune stays a no-verdict-change operation, and there it prunes nothing.

That costs no unbounded growth, because the prune's motive does not transfer either: it exists because a *remote* stream chooses its key names. These keys are the attached devices' own motor names plus whatever `map_fn` emits - fixed by the session's hardware - and the baseline is already reset per session. The size bound is that key set, not a time window. `_local_slew` is validated positive-finite, so `inf / _local_slew` cannot degenerate to `nan` and prune everything.

One deliberate step past the one-call-site fix you scoped: this is the second instance of one shape. R1 was the checker inheriting a mesh default; this is the pruner inheriting one. Both read fine at the call site, which is why neither was caught by eye. So alongside the fix, `TestBothCallSitesCarryTheLocalBound` asserts over the AST that *every* mesh slew helper called from `_teleop_loop` passes the local parameters explicitly, with a premise check that both helpers are still actually called so it cannot pass by matching nothing. That makes the class unrepresentable rather than clearing one instance of it.

Suite is 18 fail / 4 pass on `main` -> 22 pass here (re-measured against `origin/main`, description counts updated with it). The new behavioural pin needs two devices, which is why the existing single-device tests passed either way - noted in the class docstring so it is not later "simplified" back to one device.

cagataycali:
Follow-up in `c2e057f`, on top of the R4 fix: a documentation drift that R1 created and two directions R4 left unpinned.

**The guide named the wrong knob.** R1 gave the local loop its own bound (`STRANDS_TELEOP_SLEW_ABS`, 500 units/s) and the README row was added with it, but `docs/hardware/teleoperation.md` still described that bound as the mesh path's `STRANDS_MESH_INPUT_SLEW_ABS` at `8*pi` - in both the per-tick description and the "How it relates to mesh teleop" section. So the guide documented a knob and a default that are not the ones in force, and an operator widening the value the guide names would have changed the mesh receive path instead of the local loop. Both places now name `STRANDS_TELEOP_SLEW_ABS` and say why the two paths carry different defaults, and the guide gains a paragraph on what a quiet device's baseline means when it reconnects.

**The narrowed direction.** Your thread noted the mismatch "also bites when an operator *narrows* `STRANDS_TELEOP_SLEW_ABS` below the mesh default". That is the sharper half of the two, since tightening the bound *lengthens* the window in which an entry can still refuse while the mesh-scale prune horizon does not move, and the existing narrowing test (`0.0001`, in `TestRefusalsAreVisibleInTheSessionStatus`) exercises status visibility rather than the prune. `TestNarrowingTheBoundDoesNotShortenTheBaseline::test_a_returning_device_is_refused_under_a_narrowed_bound` drives a quiet device at `STRANDS_TELEOP_SLEW_ABS=5.0`; it fails on pre-R4 source with `a full-scale jump was applied under a narrowed bound after a 0.80s gap` and passes now. Its sibling pins the claim the unbounded envelope rests on - that the baseline only ever holds keys that reached the robot - since nothing else asserted the memory bound the envelope reasoning appeals to.

**A no-overreach control.** `TestTheMeshPathKeepsItsOwnPruneHorizon` calls the prune the way the mesh receive path does, with no overrides, and asserts a resting joint's entry is still dropped at the mesh horizon (kept at 0.9x, gone at 1.1x). It passes on both trees, which is the point: parameterising the local call site must not move the horizon the mesh path's own magnitude clamp justifies. Split for the three new tests is 1 fail / 2 pass on pre-R4 source.

The description gains a rollout for the R4 case itself. It is captured at `STRANDS_TELEOP_SLEW_ABS=0.6` rather than the shipped default, and the reason is worth stating plainly: a sim arm's joints are radians and `so101`'s base range is +/-1.92 rad, so no displacement inside that range exceeds 500 units/s after a 1 s gap. The bound's magnitude is not what R4 was about - the prune horizon's independence from it is - and the 1.0 s gap outlasts the 0.5 s mesh horizon at either setting.

The measurement in it that I did not expect: **both** runs report `slew_rejected=20` and `status=degraded`. The counters cannot distinguish them. Before R4 those 20 refusals are the arm's *return to rest* being blocked after the glitch was let through, so the follower is left parked at the bad pose for the rest of the session (still displaced at the final tick); with R4 they are the glitch itself and the arm never leaves rest. Worth recording because it means the telemetry alone was not a detector for this bug - only the pose was.

Full suite: 52 failing ids, none in `tests/test_teleop_slew_bound.py`, and `comm -23` against `upstream/main` over the non-`awsiot` failures is empty.

---
<sub>This contribution was made with assistance from an AI coding tool.</sub>

yinsong1986 (approved):
## Summary

Closes the asymmetry where `teleoperate(publish=True)` bounded a leader frame's per-joint speed on every remote follower (via the mesh receive path's `input_frame_slew_violation`, #1686) but applied the same merged frame unguarded to the local one. The loop now consults the mesh path's own helper rather than a copy, under a separate local knob (`STRANDS_TELEOP_SLEW_ABS`, default 500 units/s) sized for the driver units the shipped SO hardware actually streams (degrees, range-0-100), with refuse-and-count semantics (`slew_rejected`, session moves off `success`) instead of clamping. Both helper call sites pass the local parameters explicitly, with the pruner's envelope unbounded because this path runs no magnitude clamp - so the prune becomes a no-op and the baseline stays bounded by the session's own key set, reset per session. All five prior review rounds (R1-R5) are resolved at this head with pinned regression tests, including the class-level AST pin refusing any mesh-slew helper called on mesh defaults from this loop.

## What's good

- Verified at this head: `ruff check` / `ruff format --check` clean on both touched Python files, and the full `tests/test_teleop_slew_bound.py` + `tests/test_teleop.py` suites pass (65 passed).
- Single source of truth: the local path calls the mesh helper (spy-pinned) instead of copying the rule, so the two paths cannot drift.
- The bad env-var value warns and falls back rather than silently disabling the bound (AGENTS.md > Review Learnings #86 > Env Vars), the new knob has its README row, and the change ships as a `changelog.d/` fragment.
- Layering discipline: the mesh import stays inside `_teleop_loop` and is pinned over the AST so it cannot be hoisted back to module scope.
- Scope discipline: the `max_relative_target` contract question is explicitly left as a maintainer call rather than smuggled in.

## Verification suggestions

- Spot-check the status semantics change on a real device: a session with any refusal now reports `degraded`/`error` rather than `success`, so any tooling that gates on the teleop status string sees the new value - `python -m pytest tests/test_teleop_slew_bound.py::TestRefusalsAreVisibleInTheSessionStatus -q` covers the derivation.
- The mesh receive path's own defaults and prune horizon are pinned unchanged by `TestTheMeshPathKeepsItsOwnPruneHorizon`; a quick `git diff main -- strands_robots/mesh/` confirming the mesh package is untouched is the one-line cross-check.

PR #2364, opened 2026-08-17
fix(simulation/mujoco): route the create_world refusal by the request instead of offering reset

cagataycali:
## Problem

A world cannot be rebuilt under a live scene, so a second `create_world` is refused. `Robot("so101")` returns an engine whose world is already **created and populated**, so that refusal is the first thing such a caller meets - and it offered the same two remedies, `destroy` and `reset`, for every request:

```
World already exists. Use action='destroy' first, or action='reset'.
```

Neither is right for a caller who asked for a *different* world, and one of them is silently wrong.

**`reset` applies no `create_world` parameter.** It calls `mj_resetData` and restores the home poses, at the values the world was built with. Measured on `Robot("so101", mesh=False)`, following the advertised remedy verbatim:

| Refused call | then `reset()` | delivered |
|---|---|---|
| `create_world(terrain="rough")` | `status="success"` | `nhfield=0` - flat ground |
| `create_world(timestep=0.001)` | `status="success"` | `timestep=0.002` |
| `create_world(gravity=[0,0,-1.62])` | `status="success"` | `gravity=[0,0,-9.81]` |

The advertised remedy reports success and changes nothing about the request.

**`destroy` then `create_world` does apply the parameters, and discards the scene.** On the factory engine that means the robot the caller was just handed:

```
robots before destroy: ['so101']
robots after  dance  : []
```

The refusal named neither the contents nor that cost, so on a `Robot(...)` engine it reads as routine bookkeeping.

**The remedy that satisfies the request *and* keeps the scene was never mentioned.** `set_timestep` and `set_gravity` are published actions in the same tool schema that apply exactly those two parameters to a live world. Verified: `set_timestep(0.001)` -> `timestep=0.001`, `robots=['so101']`.

This is the shape `tests/simulation/test_set_eval_seed_requires_a_seed.py` already pins one layer over: *"A reason that offers a value the destination refuses is a dead end"* and *"A refusal with no alternative is a dead end."* And `tests/simulation/mujoco/test_create_world_physics_param_validation.py` already established that `create_world` must be at parity with `set_timestep` / `set_gravity` for **validation**; the refusal was not at parity for **routing**.

## Change

The refusal now names what the live world holds and routes by the arguments actually passed. Behaviour is untouched - it is still an error, and it still leaves the world exactly as it was.

| You asked for | Refusal routes to |
|---|---|
| `timestep=`, `gravity=` | `set_timestep` / `set_gravity` on the live world - contents kept |
| `ground_plane=`, `terrain=`, `difficulty=` | compiled in at creation: `destroy` (**with the contents it discards named**), then `create_world` |
| nothing | the world is ready to use; `reset` restarts the rollout in place |

`reset` is no longer offered for a request it cannot apply. It is still named for a bare call, which is exactly what it does.

```
create_world: a world already exists (robots: so101; objects: cube). Apply
timestep=0.001 with set_timestep on the live world; its contents stay.

create_world: a world already exists (robots: so101; objects: cube).
terrain='rough' can only be set when a world is built: destroy (this discards
robots: so101; objects: cube), then create_world with it.

create_world: a world already exists (robots: so101; objects: cube). It is ready
to use: add_robot / add_object build on it and reset restarts the rollout in
place. destroy, then create_world, starts empty and discards robots: so101;
objects: cube.
```

Which parameters a live world can adopt is single-sourced in `_WORLD_PARAM_SETTERS`, so the refusal cannot advertise a setter that does not exist - the dead end it exists to close.

## Where each advertised remedy actually lands

Rendered in MuJoCo, headless EGL. One script per column, following the remedy literally from the refused `create_world(timestep=0.001)`. The three end states are identical before and after this change - what changes is which one the caller is directed to. Only the third both applies the request and keeps the scene, and only the third is now named.

create_world remedy routing

| Route | `timestep` | scene | advertised |
|---|---|---|---|
| `reset()` | `0.002` (asked `0.001`) | `so101`, `cube` kept | before |
| `destroy()` then `create_world(timestep=0.001)` | `0.001` | **empty** (`0` cube px in frame) | before, cost unstated |
| `set_timestep(timestep=0.001)` | `0.001` | `so101`, `cube` kept (`916` cube px) | **now** |

## Tests

`tests/simulation/mujoco/test_create_world_refusal_routes_by_request.py`, on `upstream/main`: **15 failed / 23 passed**; after: **38 passed**.

Failures name the defect rather than a wording diff, e.g.

```
Failed: create_world(timestep=0.001) was refused with a message advertising reset;
reset then reported success and left timestep=0.002, not 0.001.
```

The 23 that pass on both trees are the controls, and they are the point: a second `create_world` is **still an error** for every request (not idempotent, no auto-reset - a populated world is never silently discarded), the refused call leaves timestep / gravity / heightfields / objects byte-identical, the first `create_world` on an empty engine is untouched, an invalid `terrain` still outranks the world guard, `destroy` is still the route for a parameter with no setter, and every action the refusal names is in the schema's published `action` enum and callable.

Two of the 15 pre-fix failures are `ImportError` on the new `_WORLD_PARAM_SETTERS` (the table-integrity tests); the other 13 are behaviour and message failures quoting the old text.

## Scope

- MuJoCo only, because this refusal is MuJoCo-only: the Isaac and Newton `create_world` do not guard on an existing world. Whether they should is a separate question.
- Not changed: making `create_world` idempotent or auto-resetting. Both would discard a populated world on a call that reports success, which is the opposite of what the measurements above argue for, and both are contract choices rather than a correction. The controls above pin the refusal so such a change fails here instead of landing silently.


yinsong1986 (approved):
## Summary

Replaces the one-size-fits-all `create_world`-on-a-live-world refusal ("use destroy or reset") with a routed one: `timestep`/`gravity` requests are pointed at `set_timestep`/`set_gravity` (which apply them in place and keep the scene), structural requests (`ground_plane`/`terrain`/`difficulty`) are pointed at `destroy` with the discarded contents named, and a bare call is told the world is ready to use. The refusal contract itself is unchanged - still an error, world left byte-identical - and the tests pin that explicitly (not idempotent, no auto-reset, invalid terrain still outranks the guard). Verified against head `6e5e69bd`: every action the new message advertises (`set_timestep`, `set_gravity`, `destroy`, `reset`, `add_robot`, `add_object`) is in the tool spec's published `action` enum; `_PUBLISHED_ACTIONS` pre-exists on `main` so the test import is sound; and `validate_difficulty`/`validate_terrain` run before the world-exists guard, so the `float(difficulty)` comparison inside `_world_exists_error` cannot raise on the error-dict path. `_world_contents` captures `self._world` into a local before reading it, so a concurrent `destroy` reassignment cannot NoneType it, and nothing on the new path mutates model/data. New surface is one private module constant and error-message text - no public API, wire, or schema change.

## What's good

- The routing table (`_WORLD_PARAM_SETTERS`) is single-sourced and integrity-tested against `create_world`'s signature, the published action enum, and the setters' own signatures, so the refusal cannot advertise a setter that does not exist or does not take the parameter.
- Control tests pin the pre-existing contract (second `create_world` is still an error; refused call leaves timestep/gravity/heightfields/objects unchanged), so a later idempotency/auto-reset change fails loudly instead of silently discarding a populated world - exactly the AGENTS.md no-silent-defaults posture.
- Changelog fragment under `changelog.d/`, docs updated in the same PR, ASCII-only user-facing strings, no host paths in tests.

PR #2365, opened 2026-08-17
fix(policies/kimodo): let the caller decide whether a checkpoint runs its own code

cagataycali:
## Problem

`DiffusersKimodoAgent` forwards `KimodoConfig.trust_remote_code` to `DiffusionPipeline.from_pretrained`, and the test that pins that forwarding states the contract:

> `trust_remote_code` is security-relevant: Kimodo ships a custom pipeline class, so **the caller's decision has to be the one that reaches diffusers rather than a value this agent picks.**

That test builds a `KimodoConfig` directly, so two properties the sentence rests on were never exercised: the value picked when the caller says nothing, and whether any supported keyword route can carry a decision at all. Both were wrong.

The field defaulted to `True`, and it was neither a `KimodoPolicy.__init__` parameter nor a `config_keys` entry. Measured on `upstream/main`, with `STRANDS_TRUST_REMOTE_CODE=1` set so the provider gate is open:

| caller route | what reached `from_pretrained` |
| --- | --- |
| nothing passed | `trust_remote_code=True` - repository code allowed to run |
| `KimodoPolicy(trust_remote_code=False)` | `TypeError: unexpected keyword argument` |
| `create_policy("kimodo", trust_remote_code=False)` | `TypeError: unexpected keyword argument` |
| `policy_config={"trust_remote_code": False}` | `trust_remote_code=True` - **the refusal was dropped in silence** |

The last row is the one that matters most: `build_policy_kwargs` forwards only keys the provider declares in `config_keys` and drops the rest, so it forwarded `['device']` and the load proceeded on `True`. That is the route a JSON `policy_config` travels, i.e. the one an agent driving the tool surface uses. The only way to reach the flag was hand-building the frozen dataclass.

`STRANDS_TRUST_REMOTE_CODE` does not cover the gap. It decides whether the provider may be *constructed* and never sets this field, so opting in to the provider silently opted in to executing repository code for every `model_id` the process went on to load. `docs/security.md` already warns that this switch is "a per-process, whole-environment switch ... it trusts every model the process loads for the life of that process" - there was simply no finer control behind it.

## Change

- `KimodoConfig.trust_remote_code` defaults to `False`, so loading a checkpoint does not run its code unless asked. This matches the `diffusers`/`transformers` library default and LeRobot's own `eval.py` (`trust_remote_code: bool = False`).
- `trust_remote_code` and `cache_dir` become explicit `KimodoPolicy.__init__` parameters and `config_keys` entries. They were the only two `KimodoConfig` fields reachable by no keyword at all; the class deliberately takes no `**kwargs`, which is what made an unlisted field strictly unreachable rather than merely absorbed.
- The docstring and `docs/policies/kimodo.md` no longer say the environment variable gates this flag. They now describe the two opt-ins as independent, which is what they are.
- `docs/security.md` records `kimodo` on the remote-code list (it is in `_HF_REMOTE_CODE_PROVIDERS`, but only `lerobot_local` was named) and points operators at the per-call setting.

Flipping the default is observably safe for existing callers: the shipped default `model_id` is not a diffusers pipeline and is refused at load time either way. Verified against the real Hub, both values give the byte-identical refusal:

```
trust_remote_code=True  -> RuntimeError: Kimodo model_id 'nvidia/Kimodo-G1-RP-v1' is not a diffusers pipeline: ...
trust_remote_code=False -> RuntimeError: Kimodo model_id 'nvidia/Kimodo-G1-RP-v1' is not a diffusers pipeline: ...
```

The flag only reaches a caller-supplied checkpoint in diffusers pipeline layout - exactly the case where the caller should be the one deciding, and which they can now opt in to.

## Tests

`tests/policies/kimodo/test_remote_code_execution_is_the_callers_decision.py`, driving the policy's own lazy agent construction against a stubbed `diffusers` so the recorded kwargs are the ones production sends.

Pre-fix on `upstream/main`: **7 failed, 3 passed**. The headline failure reads

```
AssertionError: a caller who said nothing about remote code had the checkpoint's own code
executed: from_pretrained received trust_remote_code=True
assert True is False
```

and the silent route reads

```
AssertionError: build_policy_kwargs dropped the caller's trust_remote_code; a policy_config
asking not to execute repository code forwarded only ['device']
```

The three that pass on both trees are the boundary controls: the coarse provider gate still raises `UntrustedRemoteCodeError` without the environment opt-in, the `config=` dataclass route is undisturbed, and a typo is still a `TypeError` rather than being swallowed by a newly added `**kwargs` - the obvious wrong way to make the field reachable. A drift guard asserts that no `KimodoConfig` field is settable only by building the dataclass, so this cannot recur for a future field.

Full `tests/` on the same interpreter (lerobot 0.6.2, mujoco 3.5.0), both trees run in parallel and finishing within 1 s of each other:

| | failing ids | passed |
| --- | --- | --- |
| this branch | 54 | 30727 |
| `upstream/main` + this test file | 59 | 30720 |

`comm` on the sorted ids gives base-only = exactly the 7 pre-fix failures above, which is the full-suite confirmation of the split. Branch-only = 2 ids, both pre-existing timing flakes rather than regressions, measured in isolation 5x per tree: `test_async_rtc_masks_inference_latency` 5/5 on both trees (it only trips under full-suite load), and `test_on_the_nested_branch` 4/5 on **both** trees, i.e. the identical failure rate on `upstream/main`. Neither touches this diff, which is confined to `policies/kimodo/`, the kimodo `config_keys` block, and docs.

kimodo trust_remote_code caller decision

`STRANDS_TRUST_REMOTE_CODE=1` is set in both panels, so the only difference is the per-call decision. The bottom strip is `KimodoPolicy` driving the Unitree G1 through the documented `motion_agent=` seam; commanded joint targets are bit-identical across the two trees (`max|delta| = 0.0` over 60 ticks x 29 joints), so only the loader decision changed.

G1 rollout

## Out of scope

`lerobot_local` is also in `_HF_REMOTE_CODE_PROVIDERS` and passes no such flag itself, but the LeRobot policy classes it loads do: `gaussian_actor`, `groot`, `pi0_fast` and `wall_x` hardcode `trust_remote_code=True` in their own modeling code (11 call sites in 0.6.2). The gate is therefore protecting a real path, and narrowing it per policy type would need the checkpoint's type before construction - a separate contract decision, not a bound correction. `tool_spec.json` advertises `trust_remote_code` as a `lerobot_local` `policy_config` key that its `**kwargs` absorbs; also left alone here to keep this change to one contract.

### DIFF
--- a/changelog.d/2365-kimodo-remote-code-callers-decision.md
+++ b/changelog.d/2365-kimodo-remote-code-callers-decision.md
@@ -1,0 +1,17 @@
+### Fixed: `KimodoPolicy` lets the caller decide whether a checkpoint runs its own code
+
+`KimodoConfig.trust_remote_code` defaulted to `True` and was reachable by no
+keyword at all: it was neither a `KimodoPolicy.__init__` parameter nor a
+`config_keys` entry, so `KimodoPolicy(trust_remote_code=False)` and
+`create_policy("kimodo", trust_remote_code=False)` raised `TypeError` while
+`build_policy_kwargs` - the route a JSON `policy_config` travels - dropped the
+key without a word and loaded with `True` anyway. Only hand-building the frozen
+dataclass could reach the flag. `STRANDS_TRUST_REMOTE_CODE` did not cover the
+gap: it decides whether the provider may be constructed and never sets this
+field, so opting in to the provider also opted in to executing repository code
+for every `model_id` the process loaded.
+
+The flag now defaults to `False` and is an explicit constructor parameter and
+`config_keys` entry, alongside `cache_dir`, which was unreachable for the same
+reason. The shipped default `model_id` is refused at load time either way, so a
+default rollout is unchanged.
--- a/docs/policies/kimodo.md
+++ b/docs/policies/kimodo.md
@@ -28,13 +28,33 @@ pip install "strands-robots[kimodo]"
 ```
 
 The extra installs the `diffusers` loader, which drives any checkpoint published
-in *diffusers pipeline layout*. `trust_remote_code=True` is forwarded for a
-pipeline that ships custom code, and the factory gates it behind an explicit
-opt-in:
+in *diffusers pipeline layout*.
+
+Two independent opt-ins guard that loader, and both are off by default. The
+first decides whether the provider may be built at all:
 
 ```bash
 export STRANDS_TRUST_REMOTE_CODE=1
 ```
+
+The second decides whether a checkpoint's own Python code may run when it is
+loaded. It is per call, defaults to `False`, and is never set by the environment
+variable above, so opting in to the provider does not also opt in to executing a
+repository's code:
+
+```python
+sim.run_policy(
+    policy_provider="kimodo",
+    policy_config={
+        "model_id": "your-org/kimodo-pipeline",
+        "trust_remote_code": True,      # only for a repo you have vetted
+    },
+    instruction="walk forward",
+)
+```
+
+Leave it unset for any checkpoint that does not genuinely need it; `diffusers`
+reports plainly when a pipeline requires the flag.
 
 Weights are fetched from HuggingFace on first use under the NVIDIA Open Model
 License; nothing is bundled with `strands_robots`.
--- a/docs/security.md
+++ b/docs/security.md
@@ -101,13 +101,14 @@ Some policy providers load models from the HuggingFace Hub with `trust_remote_co
 
 Because this is code execution, not just data loading, Strands Robots forces an explicit, deliberate opt-in before any such provider will load:
 
-- The provider `lerobot_local` (`LerobotLocalPolicy`) is on the remote-code list. Any provider that loads models with `trust_remote_code=True` must be listed in `_HF_REMOTE_CODE_PROVIDERS` so the opt-in is enforced.
+- The providers `lerobot_local` (`LerobotLocalPolicy`) and `kimodo` (`KimodoPolicy`) are on the remote-code list. Any provider that loads models with `trust_remote_code=True` must be listed in `_HF_REMOTE_CODE_PROVIDERS` so the opt-in is enforced.
 - Loading is blocked by default. Attempting to create a gated provider without opting in raises `UntrustedRemoteCodeError` with an explanation, rather than silently executing remote code.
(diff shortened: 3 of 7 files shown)


yinsong1986 (approved):
## Summary

Makes `KimodoConfig.trust_remote_code` the caller's decision: the field flips its default from `True` to `False` (matching the `diffusers`/`transformers` library default), and `trust_remote_code` and `cache_dir` become explicit `KimodoPolicy.__init__` parameters and registry `config_keys` entries — previously the only two config fields reachable by no keyword route at all, so a `policy_config={"trust_remote_code": False}` was silently dropped by `build_policy_kwargs` and the load proceeded on `True`. The default flip is a security tightening whose failure mode is loud (diffusers reports plainly when a pipeline requires the flag), and the shipped default `model_id` is refused at load time under either value, so a default rollout is unchanged. The coarse `STRANDS_TRUST_REMOTE_CODE` provider gate is untouched and pinned by a control test.

## What's good

- The regression suite drives the policy's own lazy agent construction against a stubbed `diffusers`, so the recorded kwargs are the ones production sends — 10/10 pass at this head, and the full `tests/policies/kimodo/` + `tests/test_registry_integrity.py` set passes (96 tests).
- The drift guard (`test_every_config_field_is_reachable_through_a_keyword`) closes the whole class: no future `KimodoConfig` field can be settable only by hand-building the frozen dataclass.
- Reachability arrives as explicit parameters rather than a `**kwargs` absorber, and the typo-refusal control now derives the rejected keyword from the live signature — the prior CodeQL finding on the literal is resolved at this head.
- Changelog fragment per AGENTS.md step 3; `docs/security.md` correctly records `kimodo` on the `_HF_REMOTE_CODE_PROVIDERS` list and describes the two opt-ins as independent.
- Scope discipline: the analogous `lerobot_local` gap is explicitly named and deferred rather than half-fixed here.

## Verification suggestions

- `python -m pytest tests/policies/kimodo/test_remote_code_execution_is_the_callers_decision.py tests/test_registry_integrity.py -q` — the new suite plus the registry `config_keys`/signature consistency checks.
- Spot-check the silent-drop route directly: `python -c "from strands_robots.registry import build_policy_kwargs; print(build_policy_kwargs('kimodo', trust_remote_code=False, device='cpu'))"` should now include `trust_remote_code`.

PR #2366, opened 2026-08-17
test(simulation/mujoco): fail the recording round-trip when the camera video carries no bytes

cagataycali:
## Problem

`TestRecordingRoundtripCameraFrames::test_recording_roundtrip_has_camera_frames` records a camera episode, reopens the dataset, and asserts the camera feature has frames. Reading a frame back is the only step of that round-trip that looks at video *content*, and it was wrapped in:

```python
try:
    sample = ds[0]
    ...  # the camera assertions
except RuntimeError:
    # torchcodec requires system FFmpeg libraries for video decode
    pass
```

`ds[0]` raises `RuntimeError` both when no decoder can be loaded **and** when the video carries no readable stream, so the guard cannot tell those apart, and tolerating the first silently tolerated the second.

Nothing upstream of the decode notices: reopening an empty-video dataset succeeds, and `len(ds)` and the camera feature are read from metadata written independently of the video bytes. Running main's verification block verbatim against the real recorder:

| verification | MP4 bytes recorded | after truncation | verdict |
|---|---|---|---|
| main | `[9174, 17551]` | `[0, 0]` | **PASS** |
| main | `[9183, 17563]` | unchanged | PASS |
| this change | `[9183, 17566]` | `[0, 0]` | **FAIL** |
| this change | `[9178, 17563]` | unchanged | PASS |

Main returns the same verdict for 9,174 bytes of real video and 0 bytes. An empty camera video is exactly what the recorder warns about when it captures no frames, so the round-trip could not fail for the reason it exists.

## Change

Assert on the files on disk before any decode. That needs no decoder, so an empty camera video fails on every host - including one that cannot decode video at all - and it is a plain assertion no guard can absorb. `tests/simulation/isaac/test_dataset_recording.py` already asserts this about the Isaac backend's MP4 output; this brings the MuJoCo round-trip to the same bar.

The two optional-dependency guards are narrowed to the steps an absent dependency can actually break:

* The import is gated with `pytest.importorskip`.
* The dataset constructor is no longer gated. It reads back what the recorder just wrote, and it does not touch the video decoder - a dataset whose video is unreadable still constructs and still reports its frame count - so a failure there is a recording defect, not a missing dependency. No `RuntimeError` in lerobot's dataset module is reachable from a read-only `__init__` either; they are in `_require_writer`, the read-during-record guard, and the encoder threads.
* The decode call reports a skip rather than passing quietly, so a run that read no pixels is no longer reported as a verified frame.

Both guards now bind their names unconditionally, which clears the `py/uninitialized-local-variable` alert on `ds` (alert 3, the oldest open alert in the repo's scan).

Scope: the new assertion answers "did the recorder write anything", which is decidable from the filesystem alone. Verifying the pixels stays the decode step's job and may still skip, so a video that carries bytes but no readable stream is still only caught where a decoder is available. That boundary is pinned by a test rather than left implicit.

## Tests

`_assert_recorded_camera_video_carries_bytes` is called by the existing round-trip test, so a glob that stopped matching the recorder's real layout fails loudly instead of making the check vacuous.

| test | on main | with this change |
|---|---|---|
| emptying a recorded camera video fails the check (records for real, truncates, requires a failure) | fail | pass |
| a recording that landed no camera video is rejected | fail | pass |
| a camera video that carries bytes is accepted | pass | pass |
| the check needs no decoder to reach a verdict (undecodable bytes accepted, zero bytes rejected) | fail | pass |
| a decode failure skips instead of returning | fail | pass |
| a readable frame is returned unchanged | pass | pass |

The two that pass on both trees are the controls: they fail if the check starts rejecting recordings that did write a video, or if the decode guard stops handing the frame to the caller's assertions.

## Artifact

Same recorder, same dataset, same metadata (5 frames, camera feature present); only the camera video's bytes differ. Top row is the recorded video decoded back through `LeRobotDataset` (per-frame mean abs delta 0.288-1.082, so the arm is genuinely moving); bottom row is the same recording with the video truncated to 0 bytes.

recording round-trip: an empty camera video must fail the check

### DIFF
--- a/changelog.d/2366-recording-roundtrip-empty-camera-video.md
+++ b/changelog.d/2366-recording-roundtrip-empty-camera-video.md
@@ -1,0 +1,3 @@
+### Fixed
+
+The MuJoCo camera-recording round-trip test now fails when the recording lands a camera video that carries no bytes. Frame access raises `RuntimeError` both when no video decoder can be loaded and when the video has no readable stream, and the guard around it tolerated both, so an empty camera video passed every check. The recorded files are now asserted on disk before any decode, which needs no decoder and cannot be absorbed by a guard.
--- a/tests/simulation/mujoco/test_concurrency.py
+++ b/tests/simulation/mujoco/test_concurrency.py
@@ -528,11 +528,25 @@ class TestFrameDecodeReportsSkipRatherThanPassing:
     """
 
     def test_a_decode_failure_skips_instead_of_returning(self):
-        """A ``RuntimeError`` from frame access is reported as a skip."""
+        """A ``RuntimeError`` from frame access is reported as a skip.
+
+        The exception type is the substance of this double, not an arbitrary
+        choice: ``RuntimeError`` is what LeRobot's decode path raises when
+        torchcodec or the system FFmpeg libraries are missing, and it is the type
+        :func:`_read_first_frame_or_skip` catches. Raising a ``LookupError``
+        instead would leave the skip branch unexercised. The raise sits in a
+        helper rather than in ``__getitem__`` so the double is a stand-in for a
+        failing decode rather than a container whose subscript is always an
+        error.
+        """
 
         class _DecodeBroken:
-            def __getitem__(self, idx):
+            @staticmethod
+            def _decode(idx):
                 raise RuntimeError("Could not load libtorchcodec (video decode failed)")
+
+            def __getitem__(self, idx):
+                return self._decode(idx)
 
         with pytest.raises(pytest.skip.Exception, match="video decode unavailable"):
             _read_first_frame_or_skip(_DecodeBroken())
--- a/tests/simulation/mujoco/test_concurrency.py
+++ b/tests/simulation/mujoco/test_concurrency.py
@@ -15,6 +15,8 @@ import shutil
 import tempfile
 import threading
 import time
+from pathlib import Path
+from typing import Any
 
 import numpy as np
 import pytest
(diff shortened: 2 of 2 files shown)


yinsong1986 (dismissed):
## Summary

Test-only hardening of the MuJoCo camera-recording round-trip, plus a changelog fragment. The old round-trip wrapped both the dataset reopen and the frame decode in guards that swallowed `RuntimeError`, which meant an empty camera video (the exact failure the recorder warns about) passed every check because decode failure and missing-decoder are the same exception. The change asserts on the recorded MP4 bytes on disk *before* any decode - decidable on every host with no decoder - narrows the optional-dependency skips to the only steps an absent dependency can actually break (the import and the decode), and makes the decode path skip visibly instead of passing quietly. New unit tests pin both directions (empty video rejected, real bytes accepted, undecodable-but-nonempty accepted, decode failure skips, readable frame returned unchanged), and the byte check is exercised by the real round-trip so a glob drifting from the recorder's layout fails loudly rather than going vacuous.

## What's good

- The recording helper passes `root=` derived from `tmp_path`, per AGENTS.md rule 15 (a recording test names its own dataset root).
- Control tests in both directions guard against the check becoming either vacuous or over-eager, and the scope boundary (filesystem check vs. pixel verification) is pinned by a test rather than left implicit.
- The open CodeQL thread on the `_DecodeBroken` fixture is the known hostile-fixture class (same rule as alerts 590/852); rewriting it to raise `LookupError` would break the measurement, since `RuntimeError` is what torchcodec actually raises - dismiss-with-reason and resolve is the documented path.

yinsong1986 (approved):
## Summary

Test-only hardening of the MuJoCo camera-recording round-trip, plus a conforming changelog fragment. The old round-trip's decode guard swallowed `RuntimeError`, which is raised both when no decoder is available and when the recorded video carries no readable stream, so an empty camera video passed every check. The change asserts on the recorded MP4 bytes on disk before any decode (decidable on every host, absorbable by no guard), narrows the optional-dependency skips to the two steps an absent dependency can actually break (the import and the decode), and makes the decode path skip visibly rather than pass quietly. New unit tests pin both directions - empty video rejected, real bytes accepted, undecodable-but-nonempty accepted, decode failure skips, readable frame returned unchanged - and the byte check is exercised by the real round-trip so a glob drifting from the recorder's layout fails loudly instead of going vacuous. The head commit also moves the fixture's always-raising decode out of `__getitem__` into a helper, addressing the CodeQL thread from the prior review round without weakening the double (it still raises the `RuntimeError` the real decode path raises).

## What's good

- Both recording call sites pass `root=` derived from `tmp_path`, per AGENTS.md rule 15 (a recording test names its own dataset root).
- Control tests guard the check in both directions, and the scope boundary (filesystem verdict vs. pixel verification) is pinned by a test rather than left implicit.
- Changelog lands as a `changelog.d/` fragment; no direct `CHANGELOG.md` edit.
- No public API, wire format, or persisted schema is touched - the change is confined to `tests/` and a news fragment.

PR #2367, opened 2026-08-17
fix(simulation/mujoco): report whether the servos will hold a written pose, and let them

cagataycali:
## Problem

`set_joint_positions` writes `qpos`, runs forward kinematics and reports `Set N/N joint positions, FK updated`. Its documented purpose is "teleport / set an initial pose". On a robot whose joints are held by position servos it delivers only half of that: the servos are still commanded to their previous setpoint, so the first `mj_step` drives the pose straight back toward it.

Measured on `so101` (a 6-joint pose, then `step(150)`):

```
set_joint_positions(pose)  -> success: "Set 6/6 joint positions, FK updated"
pose error right after     :  0.0000 rad
pose error after 150 steps :  2.7534 rad   (every joint back near zero)
data.ctrl                  :  [0, 0, 0, 0, 0, 0]
```

The pose read back correct and the call reported success because nothing had stepped yet. **42 of the 62 loadable registry robots** drive at least one joint with a position servo (`so101`, `panda`, `aloha`, `unitree_g1`, `spot`, `ur5e`, `kinova_gen3`, ...), so this is the common case rather than an edge one.

The rest of the backend already treats a pose without its matching setpoint as a bug, and names this exact failure:

- `actuate_robot` seeds every actuator it adds from its joint's current position, commented "so the arm doesn't snap to zero on the next step";
- `eject_robot` carries `ctrl` across a robot removal because a dropped setpoint reads as zero and then "an arm parked mid-air sags to the floor while `remove_robot` reported `success` and the pose read back correct, because nothing had stepped yet";
- `add_robot`'s guarantee is quoted as "an arm already in the world keeps the pose it is in **and the actuator setpoints holding it there**".

The one surface whose whole job is writing a pose neither moved the setpoints nor said it had not, and no discovery surface mentioned it (`describe()`, the docstring and `docs/simulation/overview.md` all said "bypassing actuators", which describes the mechanism, not the consequence).

There was also no way to express "teleport and stay there": `send_action` writes the setpoints but never `qpos`, and always advances at least one step (`max(1, n_substeps)`).

## Change

The kinematic write itself is untouched, because rendering a pose and replaying a planned trajectory frame by frame are real uses, and because the companion state a direct `qpos` write leaves behind is the caller's to manage - the same reason `zero_dynamics` is a separate call rather than folded in here.

1. **The success text now names the condition and quotes the remedy** when a written joint's servo holds a different setpoint, so a pose the next step undoes is no longer reported as an unqualified success:

   ```
   Set 6/6 joint positions, FK updated. 6 of them are held by a position servo still
   commanded to a different setpoint (so101/1, so101/2, so101/3, so101/4 and 2 more),
   so the next step drives the pose back toward that setpoint; pass hold=True to move
   the setpoints with the pose
   ```

2. **`hold=True`** writes the matching position-servo setpoints alongside the pose - the same write `actuate_robot` already performs - so the pose survives stepping. Validated through the shared `boolean_flag_error` domain, since a flag read by truthiness inverts for exactly the spellings an opt-out reaches for.

**Only position servos move.** A motor takes a torque and a velocity drive takes a rate, so the same number is a different physical quantity there. That split cannot be made per robot: 14 registry robots are motor-driven throughout (`unitree_go2`, `unitree_h1`, `jvrc`, `cassie`), and `openarm` carries 2 position servos beside 16 motors. It is therefore per actuator, in a new `scene_ops.joint_drive_map` that sits beside the `actuator_joint_id` transmission rule it builds on. An actuator's `ctrl` IS a pose only when all three terms hold - an affine bias, a negative `biasprm[1]` (the `-kp` position-feedback gain) and `dyntype == mjDYN_NONE` - because MuJoCo's shortcuts overlap on any one of them (see round 1 below). Joints left alone are named:

```
Set 4/4 joint positions, FK updated. 4 written joint(s) are driven by a non-position
actuator (unitree_go2/FL_hip_joint, ...), whose ctrl is a torque, a rate or a target
behind an actuator state rather than a pose, so their setpoints were left alone
```

`hold` is published in `tool_spec.json` and in `describe()`, since the router binds against the signature and an unpublished parameter is never emitted by a schema-constrained caller.

## Result

| | pose error after `step(150)` | `data.ctrl` |
|---|---|---|
| before | 2.7534 rad | `[0, 0, 0, 0, 0, 0]` |
| `hold=False` (default, unchanged) | 2.7534 rad | `[0, 0, 0, 0, 0, 0]` |
| `hold=True` | **0.0515 rad** | `[0.5, -0.6, 0.7, -0.4, 0.3, 0.2]` |

The 0.0515 rad residual is the servo's steady-state droop under gravity against its own finite gain, not a tracking failure.

One capture script, run under each tree: it reads the report and applies whatever remedy the report quotes. Nothing is mocked, and no source differs between the panels other than the message - and therefore the route.

set_joint_positions pose hold, before and after

[pose_hold.mp4](https://raw.githubusercontent.com/cagataycali/robots/media-pose-hold-servo-setpoints/pose_hold.mp4)

Verified rather than eyeballed: the two panels' **first frames are pixel-identical** (mean abs delta `0.000`, so both start from the same written pose), the final frames differ (`8.117`), and the before panel's per-frame motion totals `66.6` against the after panel's `6.8` - the collapse against the settle.

## Tests

`tests/simulation/mujoco/test_pose_write_reports_whether_the_servos_hold_it.py`, hermetic (an inline MJCF carrying one `<position>` servo, one `<motor>`, one `<velocity>`, one `<intvelocity>` and one undriven joint, gravity off so a settled servo lands exactly on its setpoint).

Against `upstream/main`: **10 failed, 2 passed**. The headline fails on pre-existing surface with a real diff rather than on a missing keyword:

```
assert 'served' in 'Set 3/3 joint positions, FK updated'
AssertionError: the report offers no remedy: Set 3/3 joint positions, FK updated
```

The 2 that pass on both trees are the no-overreach controls (`TestTheDefaultWriteIsUnchanged`): the default write moves no setpoint and still lands the whole pose in `qpos`. Also pinned: the remedy is parsed back out of the report and applied, so the test fails both for a report that names no remedy and for one that names a remedy that does not work; a servo already commanded to the pose is not reported; a motor's `ctrl` stays untouched under `hold=True`; and a velocity drive keeps its rate command while the written pose still stands after `step(200)`.

The one in-package caller (`benchmarks/libero/adapter.py`, the init-state arm qpos write) is unaffected: the LIBERO arm is motor-driven and the adapter writes its own torques.

Gate: `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` adds no error (3 pre-existing, unrelated); full `tests/simulation/mujoco` diffed against `upstream/main` with no new failures (one camera-recording test fails identically on both trees - this runner has no EGL/OSMesa).

### Round 1 - `cd34558`: require the position-feedback term, not just an affine bias

Review feedback: `biastype == mjBIAS_AFFINE` alone does not isolate position servos, so `joint_drive_map` returned velocity drives in `servos` and `hold=True` wrote a joint *angle* into a *rate* command. Confirmed on this branch - all three shortcuts compile to an affine bias:

| MJCF | `biastype` | `biasprm[:3]` | `dyntype` | `ctrl` is |
|---|---|---|---|---|
| `<position kp="10">` | 1 (affine) | `[0, -10, 0]` | 0 (none) | a pose |
| `<velocity kv="5">` | 1 (affine) | `[0, 0, -5]` | 0 (none) | a rate |
| `<intvelocity kp="10">` | 1 (affine) | `[0, -10, 0]` | 1 (integrator) | a rate |
| `<motor>` | 0 (none) | `[0, 0, 0]` | 0 (none) | a torque |

End-to-end on a single `<velocity kv="5">` scene, before this commit: `set_joint_positions({"wheel": 0.7}, hold=True)` reported `1 position-servo setpoint(s) moved with it`, `data.ctrl[0]` became `0.7`, and `step(200)` emitted `WARNING: Nan, Inf or huge value in QACC ... The simulation is unstable`. Reachable with stock assets: Menagerie's `pal_tiago/tiago_velocity.xml` drives its arm with **9 `<velocity>` actuators** (verified against the upstream file), and any caller MJCF via `replace_scene_mjcf` / `patch_scene_mjcf` may carry one.

**Fix.** The classification now requires all three terms:

```python
commands_a_pose = (
    int(model.actuator_biastype[act_id]) == affine
    and float(model.actuator_biasprm[act_id, 1]) < 0.0
    and int(model.actuator_dyntype[act_id]) == stateless
)
```

One deliberate deviation from the suggested predicate: `biasprm[1] < 0.0` rather than `!= 0.0`. That slot is `-kp`, so only a *negative* value restores toward `ctrl` as a pose; a positive one is not a restoring term at all and `<cylinder bias="0 1 0">` is free to set it (measured: `biasprm = [0, 1, 0]`, `dyntype = 2`). The `dyntype` term is as suggested and is what excludes `<intvelocity>`, which carries `-kp` yet integrates `ctrl` as a rate.

The asymmetry is why every term is required rather than assumed: the permissive error writes a pose into a command that means something else, while the strict error only declines to help - the joint is reported as left alone. That also makes the excluded stateful cases safe by default, so the report wording widened to match (`a torque, a rate or a target behind an actuator state`).

**Not a contract change.** `docs/simulation/overview.md` and the `set_joint_positions` docstring already said a velocity-driven joint is left alone; this makes the code deliver what shipped in the docs.

**Pinned** in the existing hermetic fixture, which gains a `<velocity>` and an `<intvelocity>` joint:

- the classification assertion is exact equality over both maps, so any drive misread as a servo fails it;
- a premise test asserts the two rejected drives really do carry `mjBIAS_AFFINE`, so the pin cannot pass for the wrong reason (which is how this was invisible in round 0);
- an end-to-end test writes a pose to the velocity-driven joint under `hold=True` and asserts its `ctrl` stays `0.0`, the pose still stands after `step(200)`, and `qacc` stays finite.

Against the previous classifier: **3 failed, 9 passed**; with the fix, 12 passed.

**Left alone deliberately:** `wbc_uses_position_servo` (`policies/wbc/sim_control.py:421`) shares the bias-type-only test, but it is pre-existing and read-only - it decides whether to install the whole-body torque shim and never writes a setpoint - so it belongs to its own change. `joint_drive_map`'s docstring now says that outright instead of claiming the two share one rule.

## Update: rebased on `main`, and the tendon case completed

`main` moved under this branch; the merge is in `e148273`. `scene_ops.py` conflicted **purely additively** - this branch appended `joint_drive_map` after `actuator_joint_id`, and #2386 appended `actuator_driven_joint_ids` + `tendon_joint_ids` in the same place. Both are kept: they answer different questions about the same actuator.

Composing the two surfaced a case this PR was silent on, so it is fixed here rather than left for a follow-up.

### The gap

A tendon that wraps joints couples them to one `ctrl` - the standard MJCF gripper idiom. `joint_drive_map` resolved a joint only through `actuator_joint_id`, which reports `-1` for a tendon, so those joints landed in **neither** map. `set_joint_positions` therefore reported them exactly as it reports a joint no actuator drives at all: silently. Since `hold=True` is introduced by this PR, so is the silence.

Measured on the built-in `robotiq_2f85`, writing 0.6 rad into both driver joints:

```
set_joint_positions({left_driver: 0.6, right_driver: 0.6}, hold=True)
  -> success: "Set 2/2 joint positions, FK updated"
qpos right after   : 0.6000, 0.6000
qpos after step(300): 0.0048, 0.0048     <- 0.5952 rad of the 0.6 written, 99.2%, gone
```

**26 joints across 7 registry robots** reach this: `panda`, `xarm7`, `robotiq_2f85`, `robotiq_2f85_v4`, `shadow_hand`, `stretch`, `stretch3`.

### Why the obvious fix is the dangerous one

Resolving the tendon in the **servo** test as well is the tempting one-line change, and it would write a joint angle in radians into a `[0, 255]` `ctrl` slot. Every stock tendon gripper is authored as a `<position>` actuator on its tendon, so it clears all three gain terms a servo is tested by:

| actuator | transmission | biastype | `biasprm[1]` | dyntype | ctrlrange | joints driven |
|---|---|---|---|---|---|---|
| `panda/actuator8` | tendon | affine | -100.0 | none | `[0, 255]` | 2 |
| `robotiq_2f85/fingers_actuator` | tendon | affine | -100.0 | none | `[0, 255]` | 2 |
| `shadow_hand/lh_A_FFJ0` | tendon | affine | -0.5 | none | `[0, 3.1415]` | 2 |
| `stretch3/arm` | tendon | affine | -150.0 | none | `[0, 0.52]` m | 4 |

They really are position servos - on their **tendon**, not on any joint. No gain inspection separates them from a joint servo; only the transmission does. Two independent facts disqualify the `ctrl` from carrying a joint angle: its units are the tendon's, and one `ctrl` drives 2 joints (4 on the `stretch3` telescoping arm), so there is no single angle to write.

So the joint is now resolved through `actuator_driven_joint_ids` (a tendon-driven joint gets classified) while the servo test keeps reading `actuator_joint_id` (it is classified as **not** a pose). Such a joint is reported on the same terms as any other non-position drive, and its setpoint is left alone. `test_the_tendon_ctrl_never_receives_a_joint_angle` and `test_the_gain_terms_alone_would_call_it_a_servo` pin that asymmetry; both pass before and after this change, so they fail only if the permissive version is ever adopted.

### Blast radius

Over all 62 loadable registry robots:

| | count |
|---|---|
| previously-classified joints that changed bucket or actuator | **0** |
| joints newly classified | **26** |
| joints lost | **0** |
| newly-classified joints landing in `servos` | **0** (all in `other_drives`) |

Strictly additive. The default write (no `hold`) is byte-identical, pinned by `test_the_default_write_text_is_unchanged_for_a_tendon_joint`.

### Rollout

One `hold=True` call on `panda` requesting two arm joints and both fingers. The arm joints hold; the tendon-coupled fingers lose the whole 0.04 m. The new report is what tells the caller which is which.

set_joint_positions hold=True on panda: which written joints actually hold

| joint | drive | requested | settled after `step(400)` | lost |
|---|---|---|---|---|
| `panda/joint2` | position servo | +0.550 | +0.558 | 0.0083 |
| `panda/joint4` | position servo | -1.900 | -1.904 | 0.0040 |
| `panda/finger_joint1` | tendon-coupled | +0.040 | +0.000 | **0.0400** |
| `panda/finger_joint2` | tendon-coupled | +0.040 | +0.000 | **0.0400** |

Both panels are the same rollout, captured by running one script under each tree: the renders are pixel-identical (mean `|px|` = 0.00001), so what changes is only what is reported. `t = 0` vs `t = 400` differs by 195,446 of 313,600 pixels, so the collapse is the visible motion.

### Verification

- `tests/simulation/mujoco/test_pose_write_reports_whether_the_servos_hold_it.py`: **5 failed / 18 passed** on the merge commit, **23 passed** after - the same split on mujoco **3.5.0 and 3.10.0**. Pre-fix failures quote the silence directly, e.g. `assert 'panda/finger_joint1' in 'Set 2/2 joint positions, FK updated'`.
- `tests/simulation tests/registry tests/policies` plus the repo-wide docstring / xref / ASCII / unicode-dash / `Args:` / dependency-audit / changelog guards: **16385 passed, 182 skipped, 0 failed**.
- `tests/simulation` re-run against the pinned mujoco 3.10.0: **10963 passed, 220 skipped**, plus one pre-existing failure (`test_pose_vector_rule_parity.py::test_a_numpy_array_pose_is_accepted_by_both`) that this PR does not touch and that fails identically on the merge base under 3.10.0 while passing on 3.5.0 on both trees.
- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` unchanged from `main` (19 pre-existing, none in a touched file).

### A note on the fixture

The first push of this update asserted the tendon settling point on the hand-authored fixture, and CI (mujoco 3.10.0) caught it. One tendon over two joints constrains only their weighted **sum**, so their difference is a zero-stiffness mode; under 3.10.0's integrator it grew until the driver joints sat at -1.38 / +1.39 rad, outside their own `range="0 1.2"`. The claim held - the pose did not survive stepping - but the assertion was pinned to a landing point that is not the contract.

`robotiq_2f85` locks that mode with an `<equality>` joint coupling between its two drivers plus `armature="0.005"`, so the fixture now does the same and a comment records why. The physical claim moved to the two real grippers, where settling is build-independent (`panda` 0.04 -> 0.0000, `robotiq_2f85` 0.6 -> 0.0027, identical on both builds) and is asserted as "far from what was asked for" rather than against a settling point. The hand-authored scene keeps the classification and reporting assertions, which are deterministic.

### Round 2 - `6813358`: the tendon fixture settles by its servo, not by a contact

Review feedback: the tendon settling assertion on the hand-authored fixture failed deterministically under the pinned mujoco 3.10.0 and under 3.8.0, ~1.3 rad from the asserted point, which blocks the merge mechanically since `call-test-lint` runs under `-x`.

Fixed in two steps. `c437dd3` (the previous push, recorded in "A note on the fixture" above) removed the assertion and moved the physical claim to the two real grippers. This commit removes the **cause**, so the fixture stops being a scene whose settling point is a solver artifact.

The free anti-symmetric mode was real, but it was not the whole story. With the mode locked by the `<equality>`, the pair *still* parked at ~1.44 rad - because both finger capsules began embedded in the palm box:

```
ncon at the rest pose          : 2
left_geom  <-> palm_geom  dist : -0.01800
right_geom <-> palm_geom  dist : -0.01800
```

The capsule spans +/-0.03 about its own origin plus a 0.008 radius and the palm box top is at z=0.02, while the finger origins sat at z=0.03. So the contact force resolving that 0.018 penetration, not the tendon servo, set where the pair came to rest - and it won against the servo while travelling *opposite* the direction the servo pulls:

```
step  20  L=+1.8320 R=+1.0773  qfrc_actuator=-72.32  qfrc_constraint=+63.76/+74.46
step 300  L=+1.4959 R=+1.3764  qfrc_actuator=-71.77  qfrc_constraint=+71.79/+72.14
```

Both joints end outside their own `range="0 1.2"`. A scene in that state has no build-independent settling point to assert, which is why two mujoco versions disagreed on it.

**Fix.** The finger origins move to z=0.065, clearing the palm through the whole 0..1.2 range. The servo is then the only thing acting on the pair, and the collapse lands exactly on the stale setpoint - identically on both builds:

| mujoco | ncon | left_driver | right_driver | ten_length |
|---|---|---|---|---|
| 3.5.0 | 0 | -0.00000 | -0.00000 | -0.00000 |
| 3.10.0 | 0 | -0.00000 | -0.00000 | -0.00000 |

**Pinned** by two tests, both failing on the pre-fix geometry and passing after:

- `test_the_written_pose_collapses_to_what_the_tendon_servo_commands` takes the suggested resolution - assert the tendon length, the quantity the servo controls and therefore the build-independent one - which is assertable on the hand-authored scene now that contact no longer fights the servo. The per-joint half stays `abs(qpos - 0.6) > 0.3`, matching how the real-gripper tests assert it. Pre-fix: `assert 1.4388107130737313 == 0.0 +/- 0.01`.
- `test_the_fixture_settles_by_the_tendon_and_not_by_contact` asserts the *premise* - `ncon == 0` at rest and across 400 steps of the collapse - so a future edit that reintroduces penetration fails on the premise rather than by silently shifting a settling point. Pre-fix: `the fixture starts in contact, assert 2 == 0`.

That second pin is one deliberate step past clearing the instance: this was the second fixture claim in this PR that did not hold, and both were invisible because the scene looked reasonable by eye. Asserting no-contact makes the class fail loudly instead.

**No source file changed** - this commit is test-only. The touched file is **21 passed / 4 skipped on both mujoco 3.5.0 and 3.10.0** (the 4 skips are the stock-asset gripper cases, whose assets are absent from this runner); `ruff check` and `ruff format --check` clean.


strands_robots/simulation/mujoco/scene_ops.py
```diff
@@ -176,6 +176,55 @@ def actuator_joint_id(model: Any, act_id: int, mj: Any) -> int:
...
+        in model order is reported.
+    """
+    servos: dict[int, int] = {}
+    other: dict[int, int] = {}
+    affine = int(mj.mjtBias.mjBIAS_AFFINE)
+    for act_id in range(int(model.nu)):
+        jnt_id = actuator_joint_id(model, act_id, mj)
+        if jnt_id < 0:
+            continue
+        if int(model.actuator_biastype[act_id]) == affine:
```
yinsong1986: [MUST FIX] `biastype == mjBIAS_AFFINE` does not isolate position servos: MuJoCo `<velocity>` actuators are also affine-bias (`biasprm = [0, 0, -kv]`), and `<intvelocity>` is affine-bias too (`biasprm = [0, -kp, 0]` with `dyntype = INTEGRATOR`). Verified on this branch:

```python
# <position kp=10>   biastype=1  biasprm=[0,-10,0]  dyntype=0
# <velocity kv=5>    biastype=1  biasprm=[0,0,-5]   dyntype=0   <- lands in `servos`
# <intvelocity kp=10> biastype=1 biasprm=[0,-10,0]  dyntype=1   <- lands in `servos`
# <motor>            biastype=0
```

So on a joint driven by a velocity actuator, `joint_drive_map` returns it in `servos`, which contradicts this function's own docstring ("a *motor* (``mjBIAS_NONE``) or velocity drive takes a torque or a rate instead ... those joints are left alone") and produces two failures in the changed path:

1. `hold=False`: the success text names the joint as "held by a position servo still commanded to a different setpoint" and quotes `hold=True` as the remedy.
2. `hold=True` (the quoted remedy): a joint *angle* is written into a *rate* command — exactly the unit confusion the PR exists to prevent. Reproduced end-to-end on this branch with a single `<velocity kv="5">` scene: `set_joint_positions({"wheel": 0.7}, hold=True)` reports "1 position-servo setpoint(s) moved with it", and `step(200)` then emits `WARNING: Nan, Inf or huge value in QACC ... The simulation is unstable`.

This is must-fix rather than deferrable because it is a crash bug on the changed path reachable with stock assets (the Menagerie `pal_tiago/tiago_velocity.xml` variant drives its arm entirely with `<velocity>` actuators, and any user MJCF via `replace_scene_mjcf`/`patch_scene_mjcf` can carry one), and because the failure is prescribed by the tool's own success text — a schema-constrained caller that follows the report as written destabilizes the sim.

Fix: require the position-feedback term and a stateless drive, e.g.

```python
if (
    int(model.actuator_biastype[act_id]) == affine
    and float(model.actuator_biasprm[act_id, 1]) != 0.0
    and int(model.actuator_dyntype[act_id]) == int(mj.mjtDyn.mjDYN_NONE)
):
```

(`biasprm[1]` is `-kp`, zero for `<velocity>`; the `dyntype` check excludes `<intvelocity>`, whose `ctrl` is a rate despite carrying `-kp`.) Worth a `<velocity>` joint in the `DRIVE_MIX_XML` fixture asserting it lands in `other_drives`. Note `wbc_uses_position_servo` (sim_control.py:421) shares the biastype-only test, but that is pre-existing and read-only; this PR is what introduces a *write* keyed on the classification.

yinsong1986 (commented):
## Summary

This PR makes `set_joint_positions` report when a written joint is held by a position servo still commanded to a different setpoint (so the next `mj_step` would undo the pose), and adds `hold=True` to move the matching position-servo setpoints with the pose. The classification of which actuators may receive a pose lives in a new `scene_ops.joint_drive_map`, built on the existing `actuator_joint_id` transmission rule; the default write is byte-for-byte unchanged and pinned so by tests. The problem statement, the discovery-surface updates (`describe()`, `tool_spec.json`, docs), and the hermetic test file are all solid. However, the actuator classification has a reachable misclassification for velocity-type actuators that makes the quoted remedy destabilize the sim on affected models — flagged inline.

## What's good

- Default behavior unchanged and explicitly pinned (`TestTheDefaultWriteIsUnchanged`), so existing callers are unaffected.
- `hold` is validated through the shared `boolean_flag_error` guard and published in both `tool_spec.json` and `describe()`, per AGENTS.md's forward-all-advertised-kwargs and schema-discoverability rules.
- The new map is computed and consumed under `self._lock`, honoring the #85 mjData thread-safety contract.
- Changelog fragment, docstring registration in `test_public_member_docstrings.py`, and docs row all present.

## Verification suggestions

Reproduce the inline finding with a one-actuator scene: `replace_scene_mjcf` with `<velocity name="v" joint="j" kv="5"/>`, call `set_joint_positions({"j": 0.7})` — the report names `j` as servo-held and quotes `hold=True` — then apply the quoted remedy and `step(200)`: MuJoCo warns `Nan, Inf or huge value in QACC` and the joint never reaches 0.7.

strands_robots/simulation/mujoco/scene_ops.py
cagataycali: Fixed as prescribed, all three terms required. `joint_drive_map` now admits an actuator to `servos` only when it has an affine bias, a non-zero `biasprm[1]` (that slot is `-kp`, zero for `<velocity>`), and `dyntype == mjDYN_NONE` (`<intvelocity>` carries `-kp` yet integrates `ctrl` as a rate).

The measurement that confirms all three are load-bearing, on this branch:

```
<position kp=10>    biastype=1  biasprm=[0,-10,0]  dyntype=0
<velocity kv=5>     biastype=1  biasprm=[0,  0,-5] dyntype=0
<intvelocity kp=10> biastype=1  biasprm=[0,-10,0]  dyntype=1
<motor>             biastype=0
```

So the bias type alone separates only `<motor>`; `biasprm[1]` is what excludes `<velocity>`, and `dyntype` is what excludes `<intvelocity>`. Confirmed the failure end to end before fixing it: a single `<velocity kv="5">` scene reported "1 position-servo setpoint(s) moved with it" for `set_joint_positions({"wheel": 0.7}, hold=True)`, and `step(200)` then emitted `Nan, Inf or huge value in QACC`.

On severity, agreed and worth restating: the reason this was must-fix rather than deferrable is that the tool's own success text prescribed the crash. A schema-constrained caller that reads "held by a position servo still commanded to a different setpoint" and follows the quoted `hold=True` remedy destabilizes the sim. The permissive error writes a pose into a command that means something else; the strict error only declines to help and reports the joint as left alone -- so every term is now required rather than assumed.

The fixture gains both a `<velocity>` and an `<intvelocity>` joint. Three pins, because the classification assertion alone could pass for the wrong reason: exact-equality on the split, a premise test asserting the rejected drives really do carry an affine bias (so the pin is about the new terms rather than about the bias check), and an end-to-end test writing a pose to the velocity-driven joint under `hold=True` -- asserting its `ctrl` stays 0, the pose still stands after `step(200)`, and `qacc` stays finite. Against the previous classifier: 3 failed, 9 passed.

This also makes the code match what was already shipped: `docs/simulation/overview.md` and the `set_joint_positions` docstring both promised a velocity-driven joint is left alone.

On `wbc_uses_position_servo` (sim_control.py:421): left alone as you scoped it -- pre-existing and read-only, it only decides whether to install the torque shim. What changed is that the helper's docstring no longer claims the two share one rule, so the next reader is not told the classification is common when it is not.

Separately, the `Detect an untested overlap with the base branch` gate was red because `main` had moved `simulation/mujoco/simulation.py` under the branch. `main` is now merged in (through #2370), and `tests/simulation/mujoco/` was run on both trees in the same interpreter: 130 failing ids on each, identical sets, so the merge adds no regression.

yinsong1986 (dismissed):
## Summary

This PR closes the gap between what `set_joint_positions` reports and what the next `mj_step` delivers: a `qpos` write on a servo-held joint previously read back correct and reported unqualified success while the servo's stale setpoint pulled the pose straight back. The change leaves the default kinematic write byte-identical (pinned by the no-overreach controls), qualifies the success text by naming the joints whose servo disagrees and quoting the `hold=True` remedy, and makes `hold=True` perform the same setpoint seed `actuate_robot` already does. The drive classification lives in `scene_ops.joint_drive_map` and requires all three terms (affine bias, negative `biasprm[1]`, `dyntype == mjDYN_NONE`), which resolves the round-1 finding that a bias-type-only test admitted `<velocity>`/`<intvelocity>` drives and would have written a joint angle into a rate command.

## What's good

- The round-1 classification fix is pinned three ways in the hermetic fixture: exact-equality over both maps, a premise test asserting the rejected drives really do carry `mjBIAS_AFFINE` (so the pin cannot pass for the wrong reason), and an end-to-end write to the velocity-driven joint asserting `ctrl` stays `0.0` and `qacc` stays finite.
- `hold` is published everywhere the router and a schema-constrained caller look (signature, `tool_spec.json`, `describe()`, docstring `Args:`), and validated through the shared `boolean_flag_error` domain rather than read by truthiness - verified through `_dispatch_action` that the flag forwards and that `hold="false"` is refused with nothing written.
- Both `ctrl` writes and the drive-map read happen under `self._lock`, per the #85 thread-safety contract; the setpoint seed matches `actuate_robot`'s existing convention exactly.
- Changelog fragment, docs table row, and the `test_public_member_docstrings.py` completeness pin all landed in the same PR; no non-ASCII text added.

## Verification suggestions

- `pytest tests/simulation/mujoco/test_pose_write_reports_whether_the_servos_hold_it.py` - all 12 pass at this head (verified locally under MuJoCo 3.8.0).
- The remedy-parsing test (`test_the_remedy_the_report_quotes_makes_the_pose_survive_stepping`) is the one to spot-check if the report wording ever changes: it applies whatever `pass <name>=True` the report quotes, so it fails for a remedy that is renamed but not for one that is reworded.

tests/simulation/mujoco/test_pose_write_reports_whether_the_servos_hold_it.py
```diff
@@ -0,0 +1,473 @@
...
+        """
+        text = _text(grip_sim.set_joint_positions({"left_driver": 0.6, "right_driver": 0.6}, hold=True))
+        assert "left_driver" in text and "right_driver" in text, text
+        assert "left alone" in text, text
+
+    def test_the_pose_it_reports_on_really_does_not_survive_stepping(self, grip_sim):
+        """The report's claim, measured: the tendon actuator pulls the pose back."""
+        assert _text(grip_sim.set_joint_positions({"left_driver": 0.6, "right_driver": 0.6}, hold=True))
+        grip_sim.step(300)
+        assert _qpos(grip_sim, "left_driver") == pytest.approx(0.0, abs=0.05)
```
yinsong1986: [MUST FIX] This assertion fails deterministically under the mujoco version the lockfile pins (3.10.0) and under 3.8.0, both inside the manifest's `mujoco>=3.5.0,<4.0.0` bound - reproduced with raw `mj_step` on `TENDON_GRIP_XML` alone, no strands code in the loop:

```
qpos after 300 steps: left_driver=-1.282, right_driver=+1.288   (3.10.0)
qpos after 300 steps: left_driver=-1.371, right_driver=+1.378   (3.8.0)
```

so `pytest.approx(0.0, abs=0.05)` is off by ~1.3 rad. The physics premise is too strong: the tendon servo constrains only the *tendon length* (`0.5*q_left + 0.5*q_right`) back toward the stale `ctrl=0`, so it drives the **sum** to zero, not each joint. The anti-symmetric mode (`q_left = -q_right`) is invisible to the tendon, has no restoring force in this scene (gravity off, no springs), and numerical noise excites it - the fingers settle at roughly +/-1.3 rad, blowing through their own `range="0 1.2"` limits, while the sum lands at ~0.007 exactly as the servo demands.

Why this can't be deferred: `call-test-lint / Test and Lint` is the sole required context, CI runs `hatch run test -x --strict-markers`, and a deterministic failure under the locked mujoco both blocks the merge mechanically and (via `-x`) aborts the rest of the suite. It is not flaky-on-this-box: two independent mujoco versions agree.

What would resolve it: assert the claim the report actually makes - the written pose does **not** survive - rather than where the free mode settles. E.g. `assert abs(_qpos(grip_sim, "left_driver") - 0.6) > 0.4` (and same for the right), or assert the tendon length returned to ~0 (`data.ten_length` or `0.5*(q_l + q_r) == pytest.approx(0.0, abs=0.05)`), which is version-robust because it is the quantity the servo controls.

yinsong1986 (commented):
## Summary

This PR closes the gap between what `set_joint_positions` reports and what the next `mj_step` delivers. The default kinematic write is byte-identical (pinned by the no-overreach controls); the success text now names written joints whose position servo still holds a different setpoint and quotes the `hold=True` remedy; `hold=True` writes the matching setpoints so the pose survives stepping. The classification lives in `scene_ops.joint_drive_map` and requires all four terms (affine bias, negative `biasprm[1]`, stateless dyntype, joint transmission), so velocity/intvelocity drives, motors, and tendon-coupled joints are declined and named rather than receiving a value that is not a pose in their `ctrl` units - the tendon case verified against the stock gripper idiom where the gain terms alone would misclassify. `hold` is published in `tool_spec.json` and `describe()`, validated through the shared `boolean_flag_error` guard, and the write path stays under `self._lock`. One blocker, flagged inline: a premise assertion in the new test file fails deterministically under the lockfile's own mujoco (3.10.0, reproduced with raw `mj_step` and no strands code), which will turn the sole required check red.

## What's good

- The tendon asymmetry is pinned from both directions: `test_the_gain_terms_alone_would_call_it_a_servo` asserts the premise (gains identical to a joint servo, only the transmission differs) and `test_the_tendon_ctrl_never_receives_a_joint_angle` asserts the consequence, so a future permissive "fix" fails loudly.
- The default-write no-overreach controls (`ctrl` array compared whole, exact success text for a tendon joint) make the "strictly additive" claim checkable rather than asserted.
- Changelog fragment, `Args:` entries for every new parameter, and the `test_public_member_docstrings.py` registration all follow AGENTS.md conventions.

## Verification suggestions

- After the inline fix, re-run the new file against the lockfile's exact mujoco: `uv run --with mujoco==3.10.0 pytest tests/simulation/mujoco/test_pose_write_reports_whether_the_servos_hold_it.py` - the failing assertion reproduces with raw `mj_step` on `TENDON_GRIP_XML` alone, so it is a two-minute check without the suite.

tests/simulation/mujoco/test_pose_write_reports_whether_the_servos_hold_it.py
cagataycali: Confirmed, fixed, and re-measured against the version the lockfile pins. Taking the three claims in order.

**1. The finding reproduces exactly.** On `mujoco 3.10.0`, raw `mj_step` on the reviewed `TENDON_GRIP_XML` (no strands code in the loop), 300 steps from the state a `hold=True` write leaves behind (`qpos` written, the tendon servo's `ctrl` still at its stale `0`):

```
left_driver=-1.282  right_driver=+1.288
sum/2  (the quantity the tendon controls) = +0.0033
diff   (the zero-stiffness mode)          = -2.5701
inside its own range=[0 1.2]?  left=False  right=False
```

Same numbers to three decimals, and the decomposition is the one you gave: the servo holds the tendon length (`+0.0033`, as demanded) while the anti-symmetric mode is invisible to it and runs to `-2.57`. `pytest.approx(0.0, abs=0.05)` was asserting a settling point the tendon never controls.

**2. The fixture now pins that mode the way the stock asset does.** `robotiq_2f85` constrains the same free mode with an `<equality>` joint coupling between its two drivers plus `armature` on each; `TENDON_GRIP_XML` now carries both, and a comment records why.

One honest limitation worth recording, since the equality is not a full fix on its own. Measured on `3.10.0` with the new fixture, same raw-`mj_step` probe:

```
neq=1  armature=0.0050
left_driver=+1.496  right_driver=+1.376
diff (the mode the equality pins) = +0.1195     <- constrained, was -2.5701
sum/2                             = +1.4361     <- still leaves the range
```

So the equality does pin the difference, but the *sum* is still free to wander on this integrator. That is precisely why the physical claim moved off the hand-authored scene rather than being re-tuned on it: no assertion in the file now depends on where that fixture settles. What it asserts is the contract - the report names the joints, they land in `other_drives`, and `grip_actuator`'s `ctrl` never receives a joint angle - all of which are integrator-independent.

**3. The physical claim, on the assets it is about.** Measured on `3.10.0` through the public surface:

| robot | written | after `step(300)` | joint range | inside range |
|---|---|---|---|---|
| `panda/finger_joint1,2` | 0.04 | 0.0001 | `[0, 0.04]` | yes |
| `robotiq_2f85/{left,right}_driver_joint` | 0.6 | 0.0052 | `[0, 0.8]` | yes |

Both grippers land in `other_drives`, never in `servos`, and the report names both joints and says they were left alone. The assertion is `abs(q - target) > target/2`, so the margins are `0.0399` against `0.02` and `0.5948` against `0.3` - not a threshold that a build change moves.

**4. Whole-file and blast-radius, on `3.10.0`.**

- `tests/simulation/mujoco/test_pose_write_reports_whether_the_servos_hold_it.py`: **23 passed**.
- `tests/simulation` plus the repo-wide guard files, this head vs `main` at the same merge base, same interpreter: **1 failing id on each, identical sets**, `comm` empty in both directions. The one shared failure is `tests/simulation/newton/test_multi_camera.py::TestNamedCameraRegistration::test_bad_position_shape_rejected`, which fails identically on `main` and is unrelated to this change.
- Reconciling the counts rather than eyeballing them: `11752` passed here against `11727` on the base is `+25`, and `--collect-only` on the same selection reports `11617` against `11592`, also `+25` - so every new test ran and passed, with none of the existing ones lost.

Since `joint_drive_map` now reports tendon-coupled joints in `other_drives` instead of omitting them, the success text changed for every tendon-driven robot; that blast radius is what the second bullet was checking, and nothing else in the suite asserts on it.

cagataycali: Confirmed, and fixed in two steps: `c437dd3` removed the assertion, `6813358` removes the cause.

The diagnosis of the physics premise was right, and there was a second term underneath it. The tendon does constrain only the weighted sum, so the anti-symmetric mode is free — that is why `c437dd3` locked it with the `<equality>` `robotiq_2f85` uses and moved the physical claim to the two real grippers, asserted as "far from what was asked for" rather than against a settling point. But with the mode locked the pair *still* parked at ~1.44 rad, and that turned out not to be the free mode at all:

```
ncon at the rest pose          : 2
left_geom  <-> palm_geom  dist : -0.01800
right_geom <-> palm_geom  dist : -0.01800
```

Both finger capsules began 0.018 **embedded in the palm box** — the capsule spans +/-0.03 about its own origin plus a 0.008 radius, the palm box top is at z=0.02, and the finger origins sat at z=0.03. So the contact force resolving that penetration, not the servo, was setting the landing point. Stepping the fixture alone with raw `mj_step` shows it balancing the servo and winning:

```
step  20  L=+1.8320 R=+1.0773  qfrc_actuator=-72.32  qfrc_constraint=+63.76/+74.46
step 300  L=+1.4959 R=+1.3764  qfrc_actuator=-71.77  qfrc_constraint=+71.79/+72.14
```

Note the direction: the servo pulls **negative** (`ctrl=0` against a length of 0.6) while the joints travel to +1.44, outside their own `range="0 1.2"`. A scene like that has no build-independent settling point to assert, which is exactly why two mujoco versions disagreed.

Moving the finger origins to z=0.065 clears the palm through the whole 0..1.2 range. The servo is then the only thing acting on the pair and the collapse lands on the stale setpoint, identically on both builds:

| mujoco | ncon | left | right | ten_length |
|---|---|---|---|---|
| 3.5.0 | 0 | -0.00000 | -0.00000 | -0.00000 |
| 3.10.0 | 0 | -0.00000 | -0.00000 | -0.00000 |

On the resolution you scoped: I took the second of the two suggestions — assert the tendon length, "version-robust because it is the quantity the servo controls" — and it is now assertable on the hand-authored scene precisely because contact no longer fights the servo. `test_the_written_pose_collapses_to_what_the_tendon_servo_commands` pins `ten_length` back to ~0 and keeps the per-joint half as `abs(qpos - 0.6) > 0.3`, matching how the real-gripper tests assert it. The first suggestion stays in force too: no per-joint landing point is asserted anywhere.

One step past clearing the instance, because this is the second fixture claim that did not hold: `test_the_fixture_settles_by_the_tendon_and_not_by_contact` asserts the *premise* — `ncon == 0` at rest and across 400 steps of the collapse — so a future edit that reintroduces penetration fails on the premise rather than by silently shifting a settling point. Both new tests fail on the pre-fix geometry (`the fixture starts in contact, assert 2 == 0`; `assert 1.4388107130737313 == 0.0 +/- 0.01`) and pass after, so neither can pass for the wrong reason.

No source file changed. The file is 21 passed / 4 skipped on both 3.5.0 and 3.10.0 (the 4 skips are the stock-asset gripper cases, absent from this runner); `ruff check` and `ruff format --check` clean.

---
<sub>This contribution was made with assistance from an AI coding tool.</sub>

yinsong1986 (approved):
## Summary

This PR closes the gap between what `set_joint_positions` reports and what the next `mj_step` delivers. The default kinematic write is byte-identical (pinned by the no-overreach controls); the success text now names written joints whose position servo still holds a different setpoint and quotes the `hold=True` remedy; and `hold=True` writes the matching setpoints - the same write `actuate_robot` already performs - so the pose survives stepping. The classification in `joint_drive_map` requires all four terms (affine bias, negative `biasprm[1]`, stateless dyntype, joint transmission), so a joint angle is never written into a torque, a rate, or a multi-joint tendon `ctrl` in tendon units; misclassified drives fail the strict direction only, which merely declines to help. The tendon case is resolved through `actuator_driven_joint_ids` so coupled gripper joints are reported rather than silent, and the earlier fixture instability was removed at its cause (contact penetration plus the free anti-symmetric tendon mode), with the premise itself pinned by `ncon == 0` assertions.

## What's good

- Prior review rounds are fully resolved at this head: the tendon fixture now settles by its servo, and the full new test file passes 25/25 under mujoco 3.8.0 (independently re-run) - one of the exact versions the earlier settling assertion failed under.
- `hold` is published everywhere it needs to be (`tool_spec.json`, `describe()`, per-parameter `Args:` entry per AGENTS.md convention 13), validated through the shared `boolean_flag_error` domain before any state is written, and the setpoint writes happen under `self._lock` per the #85 thread-safety contract.
- The permissive/strict asymmetry is argued in the `joint_drive_map` docstring and pinned by premise tests (`test_an_affine_bias_alone_does_not_make_an_actuator_a_servo`, `test_the_gain_terms_alone_would_call_it_a_servo`), so the classifier cannot pass for the wrong reason.
- Changelog fragment present; merge of `main` was additive (`tool_spec.json` parameter appends to different tools); no host paths, no non-ASCII in added lines.

## Verification suggestions

- `pytest tests/simulation/mujoco/test_pose_write_reports_whether_the_servos_hold_it.py` under both the lockfile's mujoco and 3.5.0 - the two builds the review history shows disagreeing on the old fixture; all 25 should pass on both.
- Spot-check the dispatch path forwards the flag end to end: `sim._dispatch_action("set_joint_positions", {"positions": {...}, "hold": True})` on any servo-driven registry robot, then `step(150)` and read back `qpos`.

PR #2369, opened 2026-08-17
fix(simulation/mujoco): name the hardware route when a registered robot has no sim model

cagataycali:
## Problem

`registry/robots.json` holds two kinds of entry. Most declare an `asset` block and can be spawned in simulation; **nine declare only a `hardware` block** — robots strands drives over LeRobot that mujoco_menagerie has no model for. `list_robots(mode="sim")` already reports that split exactly:

```
list_robots()            -> 72
list_robots(mode="sim")  -> 63
set(all) - set(sim)      == {bi_openarm, bi_rebot_b601, earthrover, hope_jr,
                             hope_jr_hand, lekiwi_client, omx, reachy2, rebot_b601}
```

`_unknown_model_msg` did not consult it, so those nine names took the **unknown-name** branch. A request whose spelling was already exactly right was answered with a spelling correction plus a pointer to the robot listing, and the actual cause — the entry declares no simulation asset at all — went unmentioned:

```
>>> Robot("reachy2", mode="sim")
RuntimeError: Failed to create sim robot 'reachy2': No model found for 'reachy2'.
              Did you mean: reachy_mini, stretch, stretch3?
              Use action='list_urdfs' to see all available robots.
```

Worse, the suggestion pool was the **whole** registry, so the correction offered could itself be hardware-only. A script that applies whatever remedy the message hands it walks a cycle:

```
hop 1  add_robot(data_config='earthrover')   -> Did you mean: hope_jr?
hop 2  add_robot(data_config='hope_jr')      -> Did you mean: hope_jr_hand, openarm, piper?
hop 3  add_robot(data_config='hope_jr_hand') -> Did you mean: hope_jr, ...   # back to hop 2
```

Three refusals, zero robots spawned, ending where it already was. Plain typos hit the same wall: `add_robot('omxx')` offered only `omx` and `add_robot('earthrove')` only `earthrover`, both unloadable. Over generated typos of all 72 registry names, **53 of 197** refusals offered at least one name this backend cannot spawn.

The distinguishing information was available at the call site and discarded: the function already resolves the canonical name and reads the entry, then only acts when `asset` is *present but missing on disk*. When `asset` is empty it fell through to the generic form, throwing away the fact that the entry exists at all.

## Why this is the message's job

`tests/registry/test_groot_data_configs_resolve_to_a_registry_robot.py` already states the consequence verbatim, and guards the groot catalogue against *creating* it:

> Pointing one at a hardware-only entry would resolve cleanly and still refuse, which reads as the same "No model found" as claiming nothing.

`tests/registry/test_lekiwi_simulatable.py` records the same failure from the other side: LeKiwi "previously failed with `No model found for 'lekiwi'` because the registry entry carried only a `hardware` block and no `asset` block". That one was fixed by finding it a model; the nine here have none to find, so the message is what has to carry the difference.

And `close_match_hint` already argues the suggestion rule this change extends — it drops a suggestion identical to the request because it "carries no information and displaces a real one out of the three slots". A hardware-only suggestion offered to a sim caller does both.

## Change

`_unknown_model_msg` gains the third condition where the other two already live:

```
Robot 'reachy2' is registered for real hardware only (LeRobot type 'reachy2'): its
registry entry declares no simulation asset, so there is no model to load. The name is
already correct, so there is no spelling to fix - drive it as hardware with
Robot('reachy2', mode='real'), or pass urdf_path= to supply a model of your own. Use
list_robots(mode='sim') to see the robots this backend can spawn.
```

and suggestions are drawn from `list_robots(mode="sim")`, so every offered name is one the engine can actually spawn. `register_urdf` models are absent from every `list_robots` mode, so narrowing the pool drops nothing that was suggestable before.

The named route was checked, not just quoted: `Robot("earthrover", mode="real")` builds, and `Robot("reachy2", mode="real")` reaches an actionable `pip install 'lerobot[reachy2]'`.

`list_urdfs`'s docstring claimed "the names returned here are exactly the identifiers `add_robot` and `load_scene` accept", which was false for those nine. Corrected to point at the `Sim` column its own table already prints.

## Deliberate consequence

Narrowing the pool means a typo whose only close neighbour is hardware-only now gets no suggestion instead of an unusable one — `'omxx'` no longer offers `'omx'`. A near miss that does have a spawnable neighbour still gets one (`'lekiwi_clint'` -> `'lekiwi'`), which is pinned as a test.

## Tests

New `tests/simulation/mujoco/test_unresolved_model_message_names_the_hardware_only_route.py`, sibling to the `#2317` file that pinned the other two conditions.

**14 failed / 6 passed** on `upstream/main`, **20 passed** after. Sample pre-fix output:

```
FAILED ...test_no_hardware_only_name_is_offered_to_a_typo[earthrove]
  the refusal for 'earthrove' offered ['earthrover'], which this backend cannot spawn
FAILED ...test_no_hardware_only_name_is_offered_to_a_hardware_only_request
  the refusal for 'bi_openarm' offered ['hope_jr'] as corrections, and those have no
  simulation asset either, so the caller lands back on this same message
```

The 6 constant passes are the over-reach controls: a registered robot whose asset is merely missing still gets the asset diagnosis and is *not* reported as hardware-only; an unknown name still gets the generic form with a spelling hint; a name with no neighbours still omits the suggestion; and the registry split the file depends on is asserted as a premise so a `robots.json` edit fails here with a reason instead of quietly making cases vacuous.

Also fixed a latent vacuity in the sibling file: its registry stub took no arguments, so once the message passes `mode="sim"` it would have raised `TypeError` instead of the `RuntimeError` that case is about.

## Artifact

Both panels are one script that parses the remedy out of the message it just received and applies it. Only the tree on `PYTHONPATH` differs.

add_robot remedy walk

## Gate

`ruff check` / `ruff format --check` clean on `strands_robots tests tests_integ`; `mypy` 18 pre-existing errors on both trees, zero branch-only.

Full `tests/` run on this branch and on `upstream/main` with the new file copied into both:

```
branch    29 failed, 30794 passed, 283 skipped, 24 errors
main      42 failed, 30781 passed, 283 skipped, 24 errors
```

Diffing the sorted failure IDs, the base-only set is **exactly the 14 pre-fix failures** of the new
file and nothing else. The one branch-only ID is
`test_policy_runner_async_rtc.py::test_async_rtc_masks_inference_latency`, a wall-clock threshold
test that passed 3/3 in isolation on this branch (and 1/1 on main) once the machine was quiet; this
change touches no timing path.

### DIFF
--- a/changelog.d/2369-hardware-only-robot-refusal-names-the-route.md
+++ b/changelog.d/2369-hardware-only-robot-refusal-names-the-route.md
@@ -1,0 +1,10 @@
+### Fixed
+
+`add_robot` and `Robot(name, mode="sim")` now say when a registered robot is hardware-only.
+Nine `registry/robots.json` entries declare a LeRobot `hardware` backend and no simulation
+`asset`; those names took the unknown-name branch, so a correctly spelled request was
+answered with a spelling suggestion and the cause went unmentioned. The suggestion pool was
+the whole registry, so the correction offered could itself be hardware-only - following the
+sole remedy offered for `earthrover` led to `hope_jr` and then back again. The refusal now
+names the LeRobot type and the `Robot(name, mode="real")` route, and suggestions are drawn
+from `list_robots(mode="sim")` so every offered name is one the backend can spawn.
--- a/strands_robots/simulation/mujoco/simulation.py
+++ b/strands_robots/simulation/mujoco/simulation.py
@@ -1196,12 +1196,28 @@ class MuJoCoSimEngine(
     def _unknown_model_msg(requested: str) -> str:
         """Build the 'model could not be resolved' error for a robot name.
 
-        Two conditions reach this message and they have different remedies, so
-        it diagnoses which one it is instead of reporting both as a bad name:
+        Three conditions reach this message and they have different remedies, so
+        it diagnoses which one it is instead of reporting them all as a bad name:
 
         * The registry does not know ``requested`` - a typo or an unknown robot.
-          Names the closest registry keys via :func:`close_match_hint` so the
-          caller can fix it in place without a discovery round-trip.
+          Names the closest sim-loadable registry keys via
+          :func:`close_match_hint` so the caller can fix it in place without a
+          discovery round-trip. The pool is deliberately the ``mode="sim"``
+          listing rather than the whole registry: a suggestion this engine
+          cannot spawn sends the caller straight back here, and the registry
+          holds hardware-only entries close enough to be suggested (the sole
+          suggestion offered for ``earthrover`` was ``hope_jr``, which is itself
+          hardware-only, so the one remedy on offer reproduced the same
+          refusal). ``close_match_hint`` already drops a suggestion identical to
+          ``requested`` for the same reason - it carries no information and
+          displaces a real one out of the three slots.
+        * The registry knows ``requested`` and the entry declares a hardware
+          backend and no simulation asset - a real robot strands drives over
+          LeRobot that has no model to load. The name is already correct, so
+          spelling suggestions are the wrong advice here too; names the hardware
+          entry point instead, the way
+          :func:`~strands_robots.robot.Robot` already answers a leader-arm name
+          with the teleoperator entry point rather than the registry listing.
         * The registry knows ``requested`` and its model XML is simply not on
           disk. Here the name is already correct, so spelling suggestions are
           the wrong advice - ``difflib`` ranks an exact match first, so this was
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Adds a third diagnosis to `MuJoCoSimEngine._unknown_model_msg`: a registered robot whose entry declares a `hardware` block and no `asset` block is now told it is hardware-only and pointed at `Robot(name, mode='real')` (plus `urdf_path=` as the supply-your-own escape hatch), instead of being answered with a spelling suggestion for a name that was already correct. The suggestion pool for genuine typos is narrowed from the whole registry to `list_robots(mode="sim")`, so every offered correction is a name this backend can actually spawn — closing the measured cycle where the sole remedy offered for one hardware-only name was another hardware-only name. The new branch lives inside the existing best-effort `try`, so a malformed registry entry degrades to the generic form rather than raising; the `list_urdfs` docstring is corrected to match the `Sim` column its table already prints (verified against actual output — hardware-only rows do render with `Sim` blank); and the sibling test file's registry stub gains the `mode` parameter, fixing a latent `TypeError` vacuity. The deliberate trade-off (a typo whose only neighbour is hardware-only now gets no suggestion) is documented in the docstring, the PR description, and pinned by test, with the registry split asserted as a premise so a `robots.json` edit fails loudly rather than making cases vacuous.

## What's good

- New test file covers all three conditions plus over-reach controls (missing-asset robots keep the asset diagnosis, unknown names keep the generic form), and verifies the advertised `mode='real'` route against the registry rather than just quoting it.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; messages are ASCII (pinned by test); ruff/format clean; 38/38 tests in both message-diagnosis files pass locally on the head SHA.
- The narrowed pool's edge case (`register_urdf` models absent from every `list_robots` mode, so nothing suggestable is dropped) is stated where the narrowing happens, and `register_robot` requires `model_xml`, so no user-registry entry can land in the hardware-only branch by accident.

PR #2370, opened 2026-08-17
fix(simulation/mujoco): keep a latched apply_force wrench latched across a scene rebuild

cagataycali:
## Problem

`apply_force` latches a wrench in the target body's own `xfrc_applied` row and documents exactly two things that end that latch:

> The wrench is latched in the target body's own `xfrc_applied` row and applied on every subsequent `mj_step` **until the next `apply_force` call for that body**. [...] `reset()` clears every latched wrench in the world.

Five other operations ended it too. Each of them rebuilds the model, and `xfrc_applied` is not carried across that rebuild: `spec.recompile` returns the whole buffer zeroed, and the eject path allocates a fresh `MjData`. Every one reported `"success"`, nothing was logged, and the wrench simply stopped being applied on any later step.

Measured with one scene (a 0.5 kg crate held at a commanded height by a wrench that cancels gravity), reading `xfrc_applied` back by name after each operation:

| operation | wrench after | verdict |
|---|---|---|
| `add_robot` | `0.0` | dropped |
| `add_object` | `0.0` | dropped |
| `add_camera` | `0.0` | dropped |
| `remove_object` | `0.0` | dropped |
| `remove_robot` | `0.0` | dropped |
| `set_geom_properties` | `25.0` | kept (mutates in place, no rebuild) |
| `step(100)` | `25.0` | kept, as documented |
| `apply_force` on another body | `25.0` | kept, as documented |
| `reset()` | `0.0` | cleared, as documented |

The three documented behaviours all hold. The five that do not are exactly the five model rebuilds. Registering one extra camera was enough: the crate resumed free fall and was **0.4601 m** lower 1.4 s later.

## The contract was already stated, three times over

- `apply_force` names the two things that end a latch, and a scene rebuild is neither.
- `add_robot` documents its composition in these words: *"an object stays where it settled or was carried to, **a latched `apply_force` wrench persists**, and the clock keeps counting"*. It did not.
- `_SceneState`, the eject path's carry, documents itself as *"the same state `save_state` checkpoints, minus the entries a rebuilt scene cannot have a surviving instance of"* -- and it already carried `qfrc_applied`, the joint-space sibling of this very buffer. A body survives a rebuild, so its wrench had no reason to be the omitted one.

## Change

Both rebuild paths now snapshot the latched wrenches and re-latch them after the compile, through one pair of helpers so there is a single rule for what identifies a latched wrench:

- `_recompile_preserving_state` (`add_robot`, `add_object`, `add_camera`, `remove_object`) reads them off the outgoing data before `spec.recompile` and re-latches before the forward pass that consumes them.
- `_SceneState` gains `body_wrenches`, so `eject_robot_from_scene` carries them the same way it already carries `qpos`, `qvel`, `qfrc_applied`, `ctrl`, `act` and the clock.

Keyed by **body name**, because a rebuild renumbers bodies -- an index-keyed carry would re-latch one body's wrench onto another. Only non-zero rows are captured: an all-zero row is "no wrench here", which is what a fresh allocation already holds, so a scene with no wrench in it carries an empty dict and the change costs nothing.

## Scope

`xfrc_applied` is the only `mjSTATE_INTEGRATION` buffer a rebuild dropped **and a caller can put a value into**. The rest are accounted for: `qpos`/`qvel` are carried by the recompile and by `_SceneState`; `qfrc_applied` is dropped by the recompile too (measured, see round 2 below) but nothing in the package ever writes a non-zero value into it, so there is nothing there to lose; `ctrl`, `act` and `time` are carried; `eq_active` is re-derived from the spec the weld equality lives on (`weld_bodies_in_scene` stores it there precisely "so it survives later recompiles"), and nothing writes `eq_active` at runtime; and no scene this package can build has mocap bodies or plugins (`nmocap == nplugin == 0`).

Not changed: `reset()` still clears every latched wrench, which is the documented way to clear them all.

## Tests

`tests/simulation/mujoco/test_scene_rebuild_carries_latched_wrench.py`, 16 cases. Against `upstream/main`: **11 failed, 5 passed**. After: 16 passed.

The 11 failures are the five rebuild operations measured two ways -- once observably (the crate keeps hovering) and once on the buffer -- plus per-body isolation across a rebuild. The observable half fails with the physical quantity rather than a buffer diff:

```
AssertionError: add_camera dropped the latched wrench: the puck fell 0.1982 m
```

The 5 that pass on both trees are the controls, and they are what fails if the carry is widened into "a wrench can never be cleared":

- `reset()` still clears every latched wrench
- stepping still keeps it latched
- a zero wrench on the same body still releases it, rebuild or not
- a removed robot's wrench is not re-latched onto a survivor
- a wrenchless scene gains no wrench

Each preservation test asserts against the commanded height rather than a hand-picked constant, and the fixture premise-asserts that the wrench really is holding the crate up, so a fall means the wrench went missing rather than that it never held anything.

## Verification

- `mypy strands_robots tests tests_integ`: 18 errors on this branch and 18 on `upstream/main` -- the same pre-existing set.
- Full `tests/` on both trees, same interpreter, in parallel: **68 failing ids on each, identical set** (`comm` empty both ways). `44 failed, 30723 passed, 24 errors` on this branch against `44 failed, 30707 passed, 24 errors` on `upstream/main` -- the `+16` passed is exactly the 16 new tests, and none of them is in the failing set.

## Rollout in MuJoCo (headless EGL)

A thruster wrench holds the crate at its commanded height. At the marked frame an **unrelated `add_camera()`** is called -- nothing else changes, and it returns `"success"` in both panels. One script, run against each tree; only the tree differs.

latched wrench across a scene rebuild

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-rebuild-wrench-31983452314/wrench_survives_rebuild.mp4)

| | crate z at the event | crate z 1.4 s later | fell |
|---|---|---|---|
| before | 0.780 m | 0.320 m | **0.4601 m** |
| after | 0.780 m | 0.780 m | 0.0 m |

Both panels are pixel-identical up to the event (max mean abs delta 0.000053), which is the control: they diverge only because of the operation under test. The hovering phase is deliberately still -- a held crate does not move.

## 13. Post-merge correction (this PR's docstring undercounted by one)

This merged before the following could be pushed to the branch, so it lands
separately as #2376. Recording it here because this description is where
the claim originated.

**The docstring this PR adds to `_recompile_preserving_state` says "One buffer is
not transferred at all rather than only in its tail: `xfrc_applied`". Two are.**
`spec.recompile` returns `qfrc_applied` entirely zeroed as well. The Scope
section above said the opposite ("`qfrc_applied` [is] carried by the recompile"),
and it is corrected in place. Measured by growing a scene by one body, across the
whole supported range:

|---|---|---|---|---|---|---|

Confirmed end to end on the merge commit through the public API: after an
unrelated `add_camera()`, the latched `xfrc_applied` row survives -- this PR's
fix, working as approved -- while a `qfrc_applied` vector does not.

**What shipped here is right; only the count was wrong.** `qfrc_applied` stays
uncarried on this path deliberately, and #2376 makes that a documented
decision instead of an unstated assumption: nothing in the package ever writes a
non-zero value into it, so there is no value for the recompile to lose, and a
carry could only be exercised by writing the buffer from outside the public API.
The note it adds records the condition under which that changes -- a joint-force
API added later must add the carry with it.

The sentence had already cost something: I opened #2373 to carry a buffer with no
writer, having believed it. That PR is closed.

### DIFF
--- a/changelog.d/2370-scene-rebuild-carries-latched-wrench.md
+++ b/changelog.d/2370-scene-rebuild-carries-latched-wrench.md
@@ -1,0 +1,3 @@
+### Fixed
+
+`add_robot`, `add_object`, `add_camera`, `remove_object` and `remove_robot` no longer silently release a latched `apply_force` wrench. Each rebuilds the MuJoCo model, and `xfrc_applied` was not carried across that rebuild, so an object a thruster or a wind field was holding up resumed free fall under a `"success"` result with nothing logged - contradicting `apply_force`, which documents the latch as holding until the next `apply_force` on that body or a `reset()`, and `add_robot`, which documents that "a latched `apply_force` wrench persists". Both rebuild paths now carry the latched wrenches by body name.
--- a/strands_robots/simulation/mujoco/scene_ops.py
+++ b/strands_robots/simulation/mujoco/scene_ops.py
@@ -250,6 +250,60 @@ def install_compiled_model(world: SimWorld, model: Any, data: Any) -> None:
     world._recompile_generation += 1
 
 
+def _snapshot_body_wrenches(model: Any, data: Any, mj: Any) -> dict[str, list[float]]:
+    """Capture every latched ``apply_force`` wrench, keyed by body name.
+
+    ``xfrc_applied`` is indexed by body, and a scene rebuild renumbers bodies,
+    so the wrench is carried by NAME for the same reason every other field of
+    :class:`_SceneState` is. Only non-zero rows are captured: an all-zero row is
+    "no wrench latched here", which is exactly what a fresh allocation already
+    holds, so a scene with no wrench in it costs an empty dict.
+
+    Args:
+        model: The compiled model the rows are being read against.
+        data: The ``MjData`` holding the latched wrenches.
+        mj: The resolved ``mujoco`` module.
+
+    Returns:
+        ``body name -> [fx, fy, fz, tx, ty, tz]`` for each body with a wrench.
+    """
+    wrenches: dict[str, list[float]] = {}
+    for bid in range(int(model.nbody)):
+        row = data.xfrc_applied[bid]
+        if not row.any():
+            continue
+        name = mj.mj_id2name(model, mj.mjtObj.mjOBJ_BODY, bid)
+        if not name:
+            # No namespace names it, so it cannot be matched across the rebuild.
+            # ``apply_force`` resolves its target by name and so cannot latch
+            # one here; a wrench on an unnamed body came from elsewhere.
+            logger.debug("snapshot_body_wrenches: body id %d has no name, wrench not carried over", bid)
+            continue
+        wrenches[name] = [float(x) for x in row]
+    return wrenches
+
+
+def _restore_body_wrenches(model: Any, data: Any, wrenches: dict[str, list[float]], mj: Any) -> None:
+    """Re-latch ``wrenches`` onto the bodies that still carry those names.
+
+    A body that no longer exists is skipped: its absence is the point of a
+    rebuild that removed it. Bodies absent from ``wrenches`` keep their
+    fresh-compile zero row, which is the same "no wrench" the snapshot read.
+
+    Args:
+        model: The freshly compiled model to resolve names against.
+        data: The ``MjData`` to write the wrenches into.
+        wrenches: A snapshot from :func:`_snapshot_body_wrenches`.
+        mj: The resolved ``mujoco`` module.
+    """
+    for name, row in wrenches.items():
+        bid = mj_name_to_id(model, mj.mjtObj.mjOBJ_BODY, name)
+        if bid < 0:
+            continue  # body no longer exists (expected for a removed robot)
+        for i, v in enumerate(row):
+            data.xfrc_applied[bid, i] = v
+
+
 def _recompile_preserving_state(world: SimWorld, spec: Any, *, raise_on_refusal: bool = False) -> bool:
     """Recompile ``spec`` in place, replacing ``world._model`` and ``_data``.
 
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Fixes a documented-contract violation in the MuJoCo backend: `apply_force` latches a wrench in `xfrc_applied` and documents exactly two ways the latch ends (a subsequent `apply_force` on the same body, or `reset()`), but all five model-rebuilding operations (`add_robot`, `add_object`, `add_camera`, `remove_object`, `remove_robot`) silently zeroed the buffer under a `"success"` result - `spec.recompile` returns `xfrc_applied` zeroed, and the eject path allocates a fresh `MjData`. The fix adds one snapshot/restore helper pair keyed by body name (correct, since a rebuild renumbers bodies) and threads it through both rebuild paths: `_recompile_preserving_state` re-latches before the forward pass, and `_SceneState` gains a `body_wrenches` field carried the same way it already carries `qfrc_applied`, the joint-space sibling of this buffer. The restore skips names absent from the new model, so an ejected robot's wrench dies with it, and only non-zero rows are captured, so a wrenchless scene carries an empty dict.

## What's good

- The 16-case test file pairs every preservation assertion with a control that the latch still ends where documented (`reset()` clears, a zero wrench releases, a removed body's wrench is not re-latched onto a survivor, a wrenchless scene gains nothing), so the fix is pinned to the carry rather than to unclearability - and the fixture premise-asserts the wrench is actually holding the puck up before the operation under test runs.
- Name-keyed carry with an explicit skip-and-debug-log for unnamed bodies; no silent drop on a path a user can reach, since `apply_force` resolves targets by name and cannot latch onto an unnamed body.
- New code runs entirely inside the existing lock/`_require_no_running_policy` discipline of the two rebuild paths - no new concurrent access to `model`/`data`.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; no host paths, no non-ASCII in the new files; both `@tool`-adjacent docstrings carry full `Args:` entries.
- `_SceneState` is module-private and both constructor sites plus the guardrail tests were updated for the new required field, so the dataclass change is not an API door.

PR #2371, opened 2026-08-17
fix(simulation/mujoco): report the placement add_robot compiled, not the one it was asked for

cagataycali:
## Problem

`add_robot(position=...)` is written as the attach frame's translation, and MuJoCo **composes** that frame with the `pos` the model's root body declares - it does not replace it. A ground-bolted arm declares `pos="0 0 0"`, so for those the offset *is* the world position. A locomotion model is authored standing, so it is not:

```python
sim.add_robot(name="dog", data_config="unitree_go2", position=[0.0, 0.0, 0.4])
# Position: [0.0, 0.0, 0.4]        <- the base compiled at z=0.845
```

**30 of the 55 single-root robots in the built-in registry declare a non-zero root `pos`**, and for every one of them the result echoed the requested vector back as the robot's placement. Measured offsets:

| robot | authored root | reported | actual base |
|---|---|---|---|
| `jvrc` | `z=1.4` | `[0, 0, 0]` | `[0, 0, 1.4]` |
| `cassie`, `talos` | `z=1.1` | `[0, 0, 0]` | `[0, 0, 1.1]` |
| `unitree_h1` | `z=1.06` | `[0, 0, 0]` | `[0, 0, 1.06]` |
| `unitree_g1` | `z=0.793` | `[0, 0, 0]` | `[0, 0, 0.793]` |
| `unitree_go2` | `z=0.445` | `[0, 0, 0.4]` | `[0, 0, 0.845]` |

The composition itself is the useful default - it is what makes `position=[0, 0, 0]` spawn a quadruped standing rather than sunk into the floor. The defect is that nothing said so. The sibling call `add_object(position=...)` *does* place its body at exactly the world point it is given (verified: `[0.1, 0.2, 0.7]` in, `[0.1, 0.2, 0.7]` out), so one parameter name meant two different things depending on which entity it addressed, and neither result said which. The requested vector was the only number the caller had, and for a floating-base model it named a place the robot was not - up to 1.4 m out.

Nothing documented the difference either: the `position` docstring covered only validation, `describe()` said nothing, and the world-building guide showed only ground-bolted arms.

## Change

`add_robot` now reports the **measured** world position of the robot's root body, read back out of `data.xpos` after the spawn's `mj_forward` - the same value `get_body_state` would return - and names the request and the model's own offset beside it whenever the two differ:

```
Position: [0.0, 0.0, 0.845] (position=[0.0, 0.0, 0.4] + model root offset [0.0, 0.0, 0.445])
```

Unchanged where there is nothing to disambiguate: a robot whose root offset is zero (every ground-bolted arm - `so101`, `panda`) keeps the plain one-vector form, and so does a robot whose roots cannot be reduced to one pose (an `aloha` attaches two arm bases, an `rby1` six, so there is no single base pose to name).

The offset semantics are now stated in three places a caller actually reads: `add_robot`'s docstring, the `describe()` discovery surface, and `docs/simulation/world-building.md`, each beside the contrast with `add_object`.

## Note on the mechanism

The offset is the **authored root `pos`**, not the model's keyframe root - the two differ, and it is easy to conflate them. The Go2 authors its base at `z=0.445` and its `home` keyframe root at `z=0.27`; the observed offset is `0.445`, and it is identical with `keyframe="home"` and with no keyframe at all. A repair aimed at keyframe handling would not move this number, so a test pins the distinction directly.

## Visual

Same script, same camera, same `add_robot(name="dog", data_config="unitree_go2", position=[0.0, 0.0, 0.4])`; only the tree under `PYTHONPATH` differs. Rendered in MuJoCo headless (EGL). The green marker is the base's measured world height - identical in both panels, 2329 px in each, so it is the control. The red marker is placed at the height the result's `Position:` line names.

add_robot reported placement vs the compiled one

On `main` the red marker sits **0.445 m below** the robot it claims to locate. On this branch the two markers are level and the line also names the request and the offset, so the composition is visible instead of having to be inferred.

## Verification

`tests/simulation/mujoco/test_add_robot_reports_the_placement_it_compiled.py`, hermetic (inline MJCF, no asset download) and GL-free.

Against `main`: **3 failed / 3 passed**. Post-fix: **6 passed**.

```
FAILED ... test_the_reported_position_is_where_the_root_body_actually_is
  AssertionError: add_robot reported 'Position: [0.0, 0.0, 0.4]' for a robot whose
  base compiled at z=0.9: the requested 0.4 names a place the robot is not, and it
  was the only number the caller was given
```

The three that pass on both trees are the over-reach controls, and they carry as much weight as the assertions: a zero-offset arm, a multi-root robot, and `add_object` itself must all report exactly what they reported before, so the change is scoped to the case that was misreporting.

Blast-radius selection run on both trees with the new test file present in each - `tests/simulation`, `tests/registry`, `tests/test_robot_factory.py` and the repo-wide docstring / ASCII / changelog / docs-catalog guards, 12261 tests each:

| | branch | `main` |
|---|---|---|
| passed | 12260 | 12257 |
| failed | 1 | 4 |

`main`'s four are the three that fail pre-fix plus `test_eval_async_rtc_masks_inference_latency`; the branch's one is its sibling `test_async_rtc_masks_inference_latency`. Both of those are wall-clock latency assertions and were load-sensitive here (box at load ~18): they pass 3/3 in isolation on the branch, and one of them **also failed 1 of 3 isolated runs on `main` itself**, so neither is attributable to this change. `passed` differs by exactly +3, which is the three pre-fix failures.

Full local gate: `ruff check` / `ruff format --check` / `mypy` clean over `strands_robots tests tests_integ` (18 pre-existing errors, unchanged).


yinsong1986 (approved):
## Summary

Makes `add_robot`'s result report the measured world position of the robot's root body (read from `data.xpos` after the spawn's `mj_forward`) instead of echoing the requested `position` back, and names the request plus the model's authored root offset beside it whenever the two differ. The change is confined to a private helper pair behind the existing `Position:` result line, plus docstring/`describe()`/guide wording; a zero-offset arm, a multi-root robot, and `add_object` keep their prior output byte-for-byte, which the test file pins as controls alongside the fix.

## What's good

- Verified locally at `b4c629c6`: the new test file passes 6/6, the neighboring `add_robot` suites (`test_simulation.py`, `test_add_robot_keyframe.py`, `test_add_robot_preserves_scene_state.py`) pass 187/187 (9 skipped), and `ruff check` / `ruff format --check` are clean on the changed files.
- No new concurrency exposure: the readback runs inside the same `_require_no_running_policy`-gated `add_robot` body, after the existing `mj_forward`, and both helpers are `_`-prefixed (no new public API surface).
- The multi-root and no-namespace edge cases degrade to the previous one-vector form (`None` from `_robot_root_world_position`) rather than guessing, and nothing else in the tree parses the `Position:` line, so the message change breaks no consumer.
- Changelog fragment present per AGENTS.md step 3; the test is hermetic (inline MJCF, `mesh=False`, no GL) and pins the authored-root-vs-keyframe-root distinction directly.

PR #2372, opened 2026-08-17
fix(policies): report the provider aliases create_policy accepts

cagataycali:
## Problem

`create_policy` resolves a provider by its canonical name **or** by any `aliases`/`shorthands` its `policies.json` entry declares, so `create_policy("sonic")` builds the same policy as `create_policy("wbc")`. No public surface enumerated those spellings:

```python
>>> len(list_providers())          # 14
>>> # spellings create_policy actually accepts: 28
```

The 14 undiscoverable spellings and what they resolve to:

| spelling | resolves to | | spelling | resolves to |
|---|---|---|---|---|
| `lerobot` | `lerobot_local` | | `random`, `test` | `mock` |
| `sonic` | `wbc` | | `sonic_gait` | `wbc_gait` |
| `moveit` | `moveit2` | | `cumotion` | `curobo` |
| `c3` | `cosmos3` | | `motion_bricks` | `motionbricks` |
| `text2motion`, `kimodo_g1` | `kimodo` | | `gtp`, `gtp_g1`, `protomotions_g1` | `protomotions` |

## Why this is a defect rather than a breadth choice

The omission is internally inconsistent. `list_providers()` already extends its result with the runtime alias table:

```python
names = list_policy_providers()
names.extend(_runtime_registry.keys())
names.extend(_runtime_aliases.keys())     # runtime aliases ARE listed
```

So `register_policy("p", loader, aliases=["my"])` makes `my` discoverable, while an equally-resolvable `policies.json` alias stays hidden — decided only by which registry the provider lives in, three adjacent lines apart. Verified:

```
runtime alias 'hp_alias'  advertised=True
JSON alias    'sonic'     advertised=False     create_policy('sonic') -> WBCPolicy
```

Both documentation surfaces already promised the coverage the function lacked:

- `docs/policies/overview.md` — "list every provider that `create_policy("<name>")` accepts with: `list_providers()`"
- `docs/api-reference.md` — described it as reporting runtime names "**and their aliases**", saying nothing about the JSON ones

## Change

The robot registry had already settled the shape: `registry/robots.py` keeps a private `_build_alias_map()` and promotes it as public `list_aliases() -> dict[str, str]`. `registry/policies.py` has the identically-named private builder and no public counterpart.

- `registry.list_policy_aliases()` — promotes that builder, JSON alias -> canonical. Named to match the existing `list_policy_providers()`/`list_providers()` pair, and to leave `registry.list_aliases()` meaning robot aliases.
- `policies.list_aliases()` — unions the JSON and runtime alias tables exactly as `list_providers()` unions their provider tables, in the precedence `_resolve_policy_class` applies (it consults the runtime registry before `policies.json`, so a runtime alias shadows a JSON one).

```python
accepted = set(list_providers()) | set(list_aliases())    # 28, exactly what create_policy takes
```

**`list_providers()` is unchanged.** `docs/policies/overview.md`'s provider table is pinned to *exactly* the canonical set by `tests/test_docs_policy_coverage.py`, whose docstring calls that table "the human-facing catalogue of what `create_policy("<name>")` accepts" — so widening the canonical list would have put alias rows in a table of providers. The mapping was the surface that was missing, and it carries what a flat list cannot: which provider a spelling resolves to.

Thirteen providers repeat their own canonical name among their `aliases`. Those entries are inert for resolution (`get_policy_provider` does `alias_map.get(name, name)`) and are omitted, so the mapping means what the robot registry's `list_aliases()` means — that one has no identity entries either.

Docs: corrected both claims above, and dropped a stale enumeration in `docs/api-reference.md` that had listed 11 of the 14 providers since three were added. Replaced with a description rather than a fresh list, so it cannot go stale the same way.

## Tests

`tests/policies/test_provider_alias_discovery.py`, **8 failed / 2 passed before, 10 pass after**.

The headline asks the question a caller has — is every accepted spelling reported by *some* public discovery surface — rather than naming one function, so it keeps holding if the surfaces are later renamed. Pre-fix it fails as a real assertion:

```
AssertionError: create_policy accepts ['c3', 'cumotion', 'gtp', 'gtp_g1', 'kimodo_g1',
'lerobot', 'motion_bricks', 'moveit', 'protomotions_g1', 'random', 'sonic', 'sonic_gait',
'test', 'text2motion'] but no public discovery surface reports them, so a caller has to
already know the spelling; surfaces probed: ['strands_robots.policies.list_providers',
'strands_robots.registry.list_aliases', 'strands_robots.registry.list_policy_providers']
```

Also pinned: each reported alias names a registered provider; no name is reported as an alias of itself; a reported alias really builds the canonical provider's class (`create_policy("random")` vs `create_policy("mock")`, no optional dependency needed); runtime aliases are reported; runtime shadows JSON as resolution does.

The two that pass on **both** trees are the controls — a premise assert that the registry declares aliases at all (so a clean result cannot be vacuous), and that `list_providers()` does not absorb the registry aliases, which is what fails if anyone widens the canonical list instead. That control reads its alias set from `policies.json` so it does not depend on the surface it guards.

Both are asserted robustly against suite-order effects: other tests register throwaway providers into the process-global runtime registry without tearing them down, so equality against the JSON registry is order-dependent. This file restores that registry after each test and grades aliases against `list_providers()`, and passes under random ordering.

## Verification

- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 18 pre-existing errors, none in a touched file.
- `tests/policies tests/registry tests/test_registry.py tests/mesh/test_policy_type_allowlist_covers_the_registry.py` plus the docs-coverage, docs-nav, public-member-docstring, docstring-xref, unicode-dash, ASCII-log, args-docstring and dependency-audit guards: **5318 passed, 3 skipped**, and 4769 passed again under random ordering.
- Adding a public function enrolled `tests/registry/test_public_member_docstrings.py`'s inventory; `policies.py::list_policy_aliases` is registered there.
- Every code block added to the docs was executed and its output checked against what the page claims.

## Artifact

Discovery-surface coverage of every accepted provider spelling

Rows are the 28 spellings `create_policy` resolves (bold = canonical provider, indented = alias with the provider it resolves to). Green means that surface reports the spelling. Each panel comes from running **one** probe script against that tree's own surfaces, so nothing is drawn by hand:

```
upstream/main : 2 policy discovery surfaces | 14 of 28 accepted spellings reported | 14 invisible
this PR       : 4 policy discovery surfaces | 28 of 28 accepted spellings reported |  0 invisible
```

The left panel's two `n/a` columns are the surfaces that do not exist there. The canonical rows are green in both panels — that is the control for `list_providers()` being unchanged.


yinsong1986 (approved):
## Summary

Adds the missing discovery surface for provider aliases: `registry.list_policy_aliases()` promotes the existing private `_build_alias_map()` (identity entries filtered), and `policies.list_aliases()` unions it with the runtime alias table so `set(list_providers()) | set(list_aliases())` is exactly the 28 spellings `create_policy` resolves. `list_providers()` is deliberately unchanged, keeping `tests/test_docs_policy_coverage.py`'s pin on the canonical provider table intact. Verified against the head: the new suite plus the docs-coverage, public-API and public-member-docstring guards all pass (81 tests), the runtime-shadows-JSON precedence in `list_aliases()` matches `_resolve_policy_class`'s lookup order (runtime registry consulted first), and the docs examples' claimed output matches what the functions actually return. The dict-return shape mirrors the robot registry's established `list_aliases()`, so the new public signatures are not a novel one-way door.

## What's good

- The headline test grades coverage against *any* public discovery surface rather than one function name, and the two always-passing controls keep the guard non-vacuous and pin that `list_providers()` stays canonical.
- The autouse fixture restores the process-global runtime registry, so the suite holds under random ordering despite other tests leaking `register_policy` state.
- Identity aliases (13 providers repeating their own name) are filtered, matching the robot registry's semantics exactly.
- New public members are enrolled in `tests/registry/test_public_member_docstrings.py` in the same PR.

yinsong1986 (approved):
## Summary

Adds the missing discovery surface for policy provider aliases: `registry.list_policy_aliases()` promotes the existing private `_build_alias_map()` (with identity entries filtered out), and `policies.list_aliases()` unions it with the runtime alias table so `set(list_providers()) | set(list_aliases())` covers all 28 spellings `create_policy` resolves. Verified against the head tree: the mapping reports exactly the 14 previously-undiscoverable JSON aliases with no identity entries and no dangling targets, `create_policy("random")` builds the same class as `create_policy("mock")`, and the `{**json, **runtime}` merge order matches `_resolve_policy_class`'s runtime-first lookup, so reported precedence agrees with resolution precedence. `list_providers()` is unchanged, keeping `tests/test_docs_policy_coverage.py`'s pin on the canonical provider table intact. The change is purely additive - two new public functions whose `dict[str, str]` shape mirrors the robot registry's established `list_aliases()`, so the API-shape question is settled by precedent rather than opened. The new test file isolates the process-global runtime registry with an autouse fixture and grades its control against `policies.json` rather than the surface under test; all 10 tests pass at head, as do the docs-coverage and public-member-docstring guards, and lint/format are clean on every touched file. Changelog fragment present per AGENTS.md step 3.

## What's good

- The headline test asks whether every accepted spelling is reported by *some* public surface, so it survives renames; the two controls (registry declares aliases at all, `list_providers()` does not absorb aliases) prevent both a vacuous pass and the wrong fix.
- New public functions enrolled in `tests/registry/test_public_member_docstrings.py`'s inventory, and both docs surfaces corrected in the same PR - including replacing a stale hardcoded provider enumeration with a description that cannot go stale the same way.

PR #2373, opened 2026-08-17
fix(simulation/mujoco): carry a latched applied force across a scene rebuild

cagataycali:
## Problem

Both operations that rebuild a scene's `MjData` discarded the two externally-applied-force buffers, and each did it silently: the call returned `"success"` and the push simply stopped.

```
apply_force('probe/hull', force=[12, -3, 5], torque=[0.4, 0.2, -0.1])
xfrc_applied                 = [12.0, -3.0, 5.0, 0.4, 0.2, -0.1]
after step(5)                = [12.0, -3.0, 5.0, 0.4, 0.2, -0.1]   <- genuinely latched
after add_robot('newcomer')  = [ 0.0,  0.0, 0.0, 0.0, 0.0,  0.0]
after remove_robot('doomed') = [ 0.0,  0.0, 0.0, 0.0, 0.0,  0.0]
```

| buffer | scene GROWS (`add_robot` / `add_object` / `add_camera`) | scene SHRINKS (`remove_robot`) |
|---|---|---|
| `qpos`, `qvel`, `ctrl`, `act`, `time` | carried by `spec.recompile` | carried by the name-keyed restore |
| `qfrc_applied` (joint-space) | **dropped** | carried (in the `joints` tuple) |
| `xfrc_applied` (body-space) | **dropped** | **dropped** |

Two different mechanisms, one contract. `spec.recompile` transfers `qpos`/`qvel`/`ctrl`/`act` and the clock but neither applied-force buffer, measured on MuJoCo 3.5 by growing a two-body scene by one body:

```
qpos  CARRIED   qvel  CARRIED   ctrl  CARRIED   act  CARRIED   time  CARRIED
qfrc_applied  DROPPED    xfrc_applied  DROPPED
```

Unlike the buffers it does carry, this is not a tail problem, so `_recompile_preserving_state`'s existing tail initialization could not help: every entry is lost, including rows of bodies that never moved. On the eject side a fresh `MjData` zero-fills both and the name-keyed restore covered only the joint-space half.

## Why this is a defect and not an open question

Three places in the package already state the contract:

- `apply_force` documents that MuJoCo never clears a latched wrench, so it "persists on every subsequent step until the next `apply_force` call for this body (or a `reset()`)". A scene rebuild is not on that list.
- `add_robot` promises that "a latched `apply_force` wrench persists" across the very recompile it performs.
- `save_state` checkpoints both buffers so a `load_state` restores "latched external forces, not just positions", and `_SceneState`'s own docstring defines its job as carrying "the same state `save_state` checkpoints, minus the entries a rebuilt scene cannot have a surviving instance of" - a surviving body plainly can have one.

## Change

Both buffers are restored by name from a single owner, `_restore_applied_forces`, shared by the two rebuild paths. `xfrc_applied` is keyed by **body name** because that is how MuJoCo indexes it and because a body index shifts when a new body is inserted earlier in the tree; `qfrc_applied` keeps its existing joint key. Only rows a caller actually set are carried, so a scene nobody pushed on stores nothing and the restore is a no-op there.

## Verification

A wrench measured by its **effect from rest**, not by reading the buffer back. This matters: a body that is already moving keeps coasting without gravity or damping, so displacement alone stays positive after the wrench is gone. On the pre-fix tree a drifter that had run 50 steps under a 10 N push still advanced 0.098 m over the next 50 with an all-zero `xfrc_applied` row, purely on momentum, while the same 50 steps **from rest** advanced 0.000 m. Zeroing the velocity first is what makes the assertion about the force.

`tests/simulation/mujoco/test_scene_rebuild_preserves_applied_forces.py`, pre-fix vs post-fix: **4 failed / 4 passed -> 8 passed**.

```
FAILED ...::test_it_still_pushes_after_another_robot_is_removed
  removing some other robot revoked the wrench: the drifter no longer accelerates
  (was 0.049 m from rest), yet remove_robot reported success   assert 0.0 == 0.049
FAILED ...::test_it_still_pushes_after_another_robot_is_added
  add_robot documents that 'a latched apply_force wrench persists', but the
  drifter stopped accelerating (was 0.049 m from rest)         assert 0.0 == 0.049
FAILED ...::test_it_still_pushes_after_an_object_is_added      assert 0.0 == 0.049
FAILED ...::test_a_joint_torque_survives_a_scene_grow          assert 0.0 == 0.5
```

The 4 that pass on both trees are the no-overreach controls: the ejected robot's own wrench is **not** re-keyed onto a survivor, a scene nobody pushed on gains no wrench, `apply_force` still replaces rather than accumulates, and `reset()` still revokes a wrench (it is on the documented revocation list and must stay there).

Blast radius, `tests/simulation tests/registry tests/tools` plus the repo-wide docstring/ASCII/changelog guards, same interpreter (MuJoCo 3.5, lerobot 0.6.2):

| | passed | failed |
|---|---|---|
| `upstream/main` | 14227 | 0 |
| this branch | 14235 | 0 |

`+8` is exactly the new tests. `ruff check`, `ruff format --check` and `mypy strands_robots tests tests_integ` are clean over the CI scope (18 pre-existing errors, unchanged).

A 9.81 N upward wrench holds a 1 kg pod against gravity - it hovers **only** because the wrench is still applied. At frame 30 `add_object` adds a crate, rebuilding the scene. Same script, same seed, same camera; the trees differ only by this change.

a latched wrench across a scene rebuild

[MP4](https://github.com/cagataycali/robots/raw/media-scene-rebuild-applied-forces/applied-force-across-rebuild.mp4)

| | altitude at rebuild | final altitude | fell |
|---|---|---|---|
| `upstream/main` | 1.0000 m | 0.6710 m | **0.3290 m** |
| this PR | 1.0000 m | 1.0000 m | 0.0000 m |

The pod's rendered centroid moves 60 px down on `main` (y 172.1 -> 231.8) and is pixel-identical on this branch (172.1 -> 172.1). Mean absolute pixel delta between the two panels is **0.000042** for every frame before the rebuild - they are the same rollout until the moment the scene changes - and 1.53 at the final frame.


cagataycali:
Closing: this is the same defect as #2370, which is approved, narrower, and already carries the half that a caller can actually reach. Keeping both would put two conflicting diffs on the same function (`_recompile_preserving_state`) for one defect.

The overlap is not partial. Both PRs snapshot `xfrc_applied` by body name before `spec.recompile` and re-latch it after, for the same reason and with the same key. #2370 does it in 85 added lines; this one does it in 129 by also moving the existing `qfrc_applied` write out of `_restore_scene_state` into a new shared owner.

**This PR's measurement was the correct one, and it has been folded into #2370.** The two descriptions contradicted each other outright -- #2370's Scope section claimed "`qpos`/`qvel`/`qfrc_applied` are carried by the recompile", this one measured `qfrc_applied` as dropped. Re-measured by growing a scene by one body, across the whole supported range:

|---|---|---|---|---|---|---|

This PR was right; #2370 was wrong, in its description and in the docstring it adds ("One buffer is not transferred at all" -- two are). #2370 merged before that correction could be pushed to its branch, so it lands separately as #2376 (docstring only, no behaviour change), and #2370's description now carries the range table and points there.

**What does not survive the closure is the `qfrc_applied` carry itself, because that buffer has no writer.** Nothing in `strands_robots/` ever puts a non-zero value into `qfrc_applied`: `apply_force` documents why it latches the per-body buffer instead (a world-wide generalized-force vector has no slice that belongs to one body), and no other caller touches it. The only writes anywhere in the tree are tests poking `data.qfrc_applied` directly -- including this PR's own `test_a_joint_torque_survives_a_scene_grow`, which sets `data.qfrc_applied[int(model.jnt_dofadr[jid])] = 0.5` by hand because there is no API that would.

So the four failing-then-passing cases here are three restatements of what #2370 fixes plus one that can only fail if the buffer is written from outside the public API. A carry with no reachable writer is speculative generality, and it is not free: it moves an existing write out of `_restore_scene_state` on a path that is already exercised, to protect a value that cannot exist yet. The condition under which it stops being speculative is now recorded in `_recompile_preserving_state`'s docstring instead -- a joint-force API added later must add the carry with it -- so the next person to add one finds the note rather than the landmine.

The rollout GIF branch (`media-scene-rebuild-applied-forces`) stays; #2370's own rollout covers the same physical claim with a hovering crate.

yinsong1986 (commented):
## Summary

Abstaining from a merge verdict: this PR was closed by its author as a duplicate of #2370, which merged nine minutes before the closure and lands the same body-name-keyed `xfrc_applied` carry across `spec.recompile`. `main` therefore already contains a change to the same function (`_recompile_preserving_state`) this diff edits, so this branch's green was measured against a base that no longer exists and "safe to merge" is no longer a claim any review of this head can make. The diff itself, read against its own base, contained no must-fix findings - the measurement here (both `qfrc_applied` and `xfrc_applied` zeroed by the recompile, MuJoCo 3.5-3.11) was in fact the correct one and has been folded into #2370's record and #2376. No inline comments; nothing here asks for reopening.

cagataycali:
Superseded by #2370, which landed the same fix a few minutes before this opened. We converged independently on the same design - snapshot `xfrc_applied` by body name before the rebuild, re-latch after, on both the grow and the eject path - and #2370 got there first, so this adds nothing for that buffer.

Verified rather than assumed: I ran this PR's 8 regression tests against merged `main` (`eff27242`). **7 of 8 pass**, so #2370 genuinely covers the wrench half on both paths.

The 8th still fails:

```
FAILED ...::TestTheJointSpaceHalfSurvivesToo::test_a_joint_torque_survives_a_scene_grow
  spec.recompile drops qfrc_applied, so a joint torque a caller latched stopped
  driving the joint as soon as anything entered the scene    assert 0.0 == 0.5
```

`spec.recompile` drops **both** applied-force buffers, and `_recompile_preserving_state` now carries only the body-space one. So on merged `main`, `qfrc_applied` is carried when the scene shrinks but not when it grows - which makes #2370's new `_SceneState` docstring ("`qfrc_applied` above is its joint-space sibling; both are part of the state `save_state` checkpoints, so both are carried") true on one path and not the other.

Closing this and opening a focused follow-up for that one remaining cell, with the failing test above as its regression. No reason to keep a 7/8-redundant PR open.

PR #2374, opened 2026-08-17
fix(registry): resolve a policy URL whose scheme is not lowercase

cagataycali:
## Problem

URL schemes are case-insensitive (RFC 3986 section 3.1), and every stage of `resolve_policy()` **but the first** already compared case-insensitively:

| stage | comparison | case-insensitive? |
|---|---|---|
| 1. URL patterns | `re.match(pattern, policy)` | **no** |
| 2. shorthands | `policy.lower() in alias_map` | yes |
| 3. HF org | `policy.split("/")[0].lower()` | yes |
| 4. provider name | `get_policy_provider(policy.lower())` | yes |

So `MOCK` resolves to `mock` and `NVIDIA/GR00T-N1.5-3B` routes to `groot`, but a URL typed with a non-lowercase scheme matched none of the `url_patterns` declared in `policies.json` (all spelled lowercase), skipped the URL stage entirely, and fell through to the HuggingFace fallback. On `main`, all five declared schemes diverge on case:

```
zmq://gpu:5555      -> groot          | ZMQ://gpu:5555      -> ('lerobot_local', {'pretrained_name_or_path': 'ZMQ://gpu:5555'})
ws://h:8765         -> remote         | WS://h:8765         -> ('lerobot_local', {'pretrained_name_or_path': 'WS://h:8765'})
cosmos3://prod:9000 -> cosmos3        | Cosmos3://prod:9000 -> ('lerobot_local', {'pretrained_name_or_path': 'Cosmos3://prod:9000'})
grpc://h:50         -> lerobot_async  | GRPC://h:50         -> ('lerobot_local', {'pretrained_name_or_path': 'GRPC://h:50'})
vera://box:9000     -> vera           | Vera://box:9000     -> ('lerobot_local', {'pretrained_name_or_path': 'Vera://box:9000'})
```

`create_policy("ZMQ://gpu:5555")` does not error and does not connect. It becomes an attempt to download a HuggingFace repo named literally `ZMQ://gpu:5555`, behind a `logger.warning` rather than a refusal, so the eventual failure is a HuggingFace error pointing nowhere near the sidecar the caller named.

## Change

The scheme is folded **once**, where the URL stage begins, rather than per branch:

```python
url = _with_lowercase_url_scheme(policy)
```

The pattern match and all five per-scheme parsers then read that one string, so a scheme added to `policies.json` later inherits the rule instead of having to rediscover it. Only the scheme is folded - hostnames, paths and HuggingFace repo ids keep the caller's exact spelling, because repo ids are case-sensitive.

Folding the scheme rather than adding `re.IGNORECASE` to the six matching sites also fixes a branch no regex flag can reach:

```python
kwargs["server_address"] = policy.replace("grpc://", "")   # before
kwargs["server_address"] = url.removeprefix("grpc://")     # after
```

`str.replace` takes no flags. Making the *pattern* match case-insensitively routes `GRPC://h:50051` to `lerobot_async` and then hands it `server_address="GRPC://h:50051"` as its gRPC target - it moves the dead end instead of closing it. Measured, with only the pattern-level fix applied, 2 of 5 schemes still produced kwargs that differ from the lowercase spelling (`grpc` `server_address`, and the `ws` `endpoint`). After this change all six sample URLs resolve to a byte-identical `(provider, kwargs)` for every spelling of their scheme.

`removeprefix` is also anchored, where the old `replace` was not.

## Tests

`tests/registry/test_resolve_policy_url_schemes.py`, the file that already covers this resolver. **10 failed / 18 passed on `main`, 28 pass after.** The headline is a property rather than a list of regexes: resolution is invariant under scheme case, parametrized over every scheme the shipped registry declares.

Pre-fix failures name the divergence directly:

```
AssertionError: 'GRPC://10.0.0.5:50051' resolved to
  ('lerobot_local', {'pretrained_name_or_path': 'GRPC://10.0.0.5:50051'}),
  but 'grpc://10.0.0.5:50051' resolves to
  ('lerobot_async', {'server_address': '10.0.0.5:50051'})
```

Three tests pass on **both** trees and are the no-overreach controls - they are what fails if the fold is widened past the scheme:

- a HuggingFace repo id is forwarded byte-for-byte (`NVIDIA/GR00T-N1.5-3B`), since repo ids are case-sensitive;
- a string with no `scheme://` (a bare `host:port`) is returned untouched;
- `_SCHEME_URLS` still covers every `url_pattern` in the shipped registry, so a newly declared scheme cannot slip past the invariance test.

Two further post-fix properties are pinned: `server_address` never keeps a scheme (it is dialed), and folding does not collapse `ws://` and `wss://` into one another.

Same script on both trees: a real `PolicyServer` serving a real `MockPolicy` over websockets on `127.0.0.1`, and a `so101` asked for that server by URL - spelled `WS://` instead of `ws://`. Nothing is stubbed; the only difference between the panels is which tree resolved the string.

rollout

Left panel per-frame delta is **exactly 0.0000** over all 50 frames - the rollout genuinely never ran. Right panel: 49 of 49 frames move (per-frame mean|delta| 1.90 to 7.92).
[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-policy-url-scheme-case/url-scheme-case-rollout.mp4)

| | BEFORE (`main`) | AFTER |
|---|---|---|
| resolved provider | `lerobot_local` | `remote` |
| resolved kwargs | `pretrained_name_or_path="WS://127.0.0.1:PORT"` | `endpoint="ws://127.0.0.1:PORT", host, port` |
| `run_policy` status | raised | success |
| control ticks | 0 | 150 |
| max abs joint (rad) | 0.0000 | 1.4726 |

Neither diagnostic on `main` names the server. Unset `STRANDS_TRUST_REMOTE_CODE` it is the remote-code gate, which talks about HuggingFace model repositories; set it, the next error is

```
ValueError: Could not determine policy type from 'WS://127.0.0.1:PORT'. config.json
had no usable 'type', 'model_type', or recognized 'auto_map' entry ... Pass
policy_type= explicitly.
```

asking the caller to name a `policy_type` for a `config.json` that will never exist.

summary

## Verification

- `mypy strands_robots tests tests_integ`: no new errors.
- Full `tests/` on this branch and on `upstream/main`: **52 failing ids on this branch vs 53 on `upstream/main`, no new failures either way** (`comm -23` empty). 30838 passed vs 30824 = +14, which is the 13 new tests in this file plus the one id below. The single base-only failure is `tests/simulation/test_policy_runner_async_rtc.py::test_async_rtc_masks_inference_latency`, a wall-clock threshold that is load-sensitive under a full parallel run - it passes 3/3 in isolation on **both** trees, so it is noise and not something this change fixes.

## Round 2 - absorb `main` (base moved, no authored change)

One commit, `196cf730`, a merge of `main`. Nothing this PR authored was rewritten.

`main` gained #2372, which adds `list_policy_aliases()` to `strands_robots/registry/policies.py` at the **same anchor** this branch adds `_URL_SCHEME_RE` / `_with_lowercase_url_scheme()` to - immediately after `list_policy_providers()`. Two independent additions at one insertion point, so `git merge` reported a content conflict on position alone; there is no semantic interaction between an alias listing and a scheme fold.

Worth recording because both PRs read `CLEAN` / `MERGEABLE` right up to the moment the first landed: each was measured against a `main` that did not yet contain the other, so the conflict was unobservable on either PR until #2372 merged. A per-PR mergeability read cannot see this - it is a property of the *pair*.

Resolution keeps both, ordered by audience: the public `list_policy_aliases()` stays adjacent to the `list_policy_providers()` its docstring contrasts itself with, and the private scheme helpers follow it, still directly above the `resolve_policy()` that is their only caller.

The resolution is verifiable as keep-both rather than as a judgement call: in `git show --cc 196cf730` the combined diff is the two blocks verbatim, one attributed to each parent, and **no line of either side is added, dropped or edited** - the only thing the merge chose was their order.

Re-verified on the merged tree: `tests/registry` + `tests/policies/test_provider_alias_discovery.py` = **786 passed, 11 skipped**, so #2372's new discovery surface and this PR's invariance property hold together. `ruff check` and `ruff format --check` clean on the resolved file.

The approval was dismissed by this push (`dismiss_stale_reviews_on_push`), and re-approval needs a second account (`require_last_push_approval`).


yinsong1986 (dismissed):
## Summary

Fixes stage 1 of `resolve_policy()` so a URL scheme is compared case-insensitively (RFC 3986 section 3.1), matching the behavior every later stage already had. The scheme is folded once (`_with_lowercase_url_scheme`) before the `url_patterns` match, and all five per-scheme parsers read that folded string - which also closes the `grpc` branch's `str.replace("grpc://", "")`, a site no regex flag could reach, and swaps it for an anchored `removeprefix`. The fold is scoped exactly right: only a leading `scheme://` is lowercased, so HuggingFace repo ids, hostnames, paths and bare `host:port` strings keep the caller's spelling, and stages 2-5 still read the raw string. Before this, `ZMQ://gpu:5555` fell through to the HuggingFace fallback as a literal repo name behind a warning, so the fix replaces a dead end rather than changing any working behavior - no compatibility concern.

## What's good

- The invariance test is parametrized over every scheme the shipped registry declares, and `TestEveryDeclaredSchemeHasASample` pins that a newly declared scheme cannot slip past it.
- No-overreach controls pass on both trees (repo-id case preserved, bare `host:port` untouched, `ws`/`wss` kept apart, host/port not folded).
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; helper docstring states the domain and the reason the fold lives at the stage boundary rather than per branch.

yinsong1986 (approved):
## Summary

Makes stage 1 of `resolve_policy()` case-insensitive on the URL scheme (RFC 3986 section 3.1), matching what stages 2-5 already did: a single anchored fold (`_with_lowercase_url_scheme`) is applied where the URL stage begins, and the pattern match plus all five per-scheme parsers read that one folded string. This also closes the `grpc` branch's `str.replace("grpc://", "")` - unreachable by any regex flag - swapping it for an anchored `removeprefix`, so `GRPC://h:50051` yields `server_address="h:50051"` rather than the whole URL as a dial target. The fold is correctly scoped: stages 2-5 still read the raw `policy`, so an undeclared scheme falls through to the HuggingFace fallback with the caller's exact spelling, and repo ids, hostnames and paths are untouched. The head commit `196cf730` is a merge of `main` resolving a position-only conflict with #2372's `list_policy_aliases()`; the combined diff confirms keep-both with no line of either parent edited.

## What's good

- Verified on the head tree: the full `tests/registry/test_resolve_policy_url_schemes.py` file passes (28/28), including the invariance property parametrized over every scheme the shipped registry declares, and `TestEveryDeclaredSchemeHasASample` pins that a newly declared `url_pattern` cannot skip the guard.
- No-overreach controls hold under adversarial probes: undeclared schemes (`HTTP://`, `FILE://`) fall through byte-identically, `ws`/`wss` stay distinct, host/port case is preserved, and non-URL strings (`mock`, repo ids, bare `host:port`, `a/B://c`) are returned unchanged - the `^`-anchored scheme regex cannot fire mid-string.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; the helper carries a complete `Args:`/`Returns:` docstring stating the RFC domain and why the fold lives at the stage boundary rather than per branch.
- Scope discipline: no public API, wire format, or registry schema changes - only previously dead-end inputs change behavior, so there is no compatibility exposure.

PR #2375, opened 2026-08-17
fix(lerobot_local): route processor_overrides to the pipeline that declares the step

cagataycali:
## Problem

A checkpoint ships two processor pipelines whose step keys are **disjoint**:

| step | pipeline | controls |
| --- | --- | --- |
| `normalizer_processor` | preprocessor | `observation.state` normalization |
| `unnormalizer_processor` | postprocessor | `action` un-normalization |
| `device_processor` | **both** | tensor placement |

LeRobot raises `KeyError` for an override key matching no step in the pipeline it
is loading (`DataProcessorPipeline._validate_overrides_used`, a typo guard).
`ProcessorBridge.from_pretrained` passed the caller's whole `overrides` dict to
**both** `_load_pipeline` calls, so every pipeline-specific step was rejected by
whichever pipeline lacked it. Measured on `lerobot/smolvla_base`:

| `processor_overrides=` | result on `main` |
| --- | --- |
| `{'normalizer_processor': {'stats': …}}` | `KeyError … Available step keys: ['unnormalizer_processor', 'device_processor']` |
| `{'unnormalizer_processor': {'stats': …}}` | `KeyError … Available step keys: [… 'normalizer_processor']` |
| both keys together | `KeyError` |
| `{'device_processor': {'device': 'cuda'}}` | loads |

`device_processor` is the only step present in both pipelines, so it was the only
override that could ever be applied -- which is also the only example the docs
gave. Every other step was unreachable.

That matters because supplying normalizer `stats` is the remedy the
inert-normalization diagnostic points callers at. A pretraining *base* checkpoint
ships stats keyed by its training dataset rather than the canonical keys --
`lerobot/smolvla_base` carries `so100.buffer.action`, `so100-red.buffer.action`,
`so100-blue.buffer.action` and no `observation.state` entry at all -- so LeRobot's
normalizer finds no matching key and returns those tensors unchanged. The
diagnostic detects it correctly and then names a remedy that cannot be applied:

```
lerobot/smolvla_base has an ACTIVE normalization pipeline but its stats do not
cover ['observation.state (STATE/MEAN_STD)', 'action (ACTION/MEAN_STD)'] …
pass processor_overrides={'normalizer_processor': {'stats': <dataset stats>}}
```

The remedy also named only the preprocessor step, while the message reports an
inert feature from each pipeline.

## Change

- Route each override to the pipeline that declares its step, reading the
  available keys from each pipeline's saved config by LeRobot's own rule
  (`registry_name`, else the final component of `class`).
- A key declared by **neither** pipeline is still refused -- LeRobot's typo
  protection is preserved -- and the refusal now lists both pipelines' step
  names instead of only the one that happened to load first.
- When a pipeline's step keys cannot be determined (offline, absent, or an
  unrecognized config shape) that pipeline receives the overrides unrouted,
  which is the existing behavior.
- The inert-normalization diagnostic now names both steps and says which half
  each one covers.

## Verification

Loading `lerobot/smolvla_base` and reading back the step's own resolved stats:

| `processor_overrides=` | `inert_normalization_features()` | unnormalizer stat keys |
| --- | --- | --- |
| *(none)* | `[observation.state (STATE/MEAN_STD), action (ACTION/MEAN_STD)]` | `so100-blue.buffer.action, so100-red.buffer.action, so100.buffer.action` |
| `normalizer_processor` only | `[action (ACTION/MEAN_STD)]` | *(unchanged)* |
| `unnormalizer_processor` only | `[observation.state (STATE/MEAN_STD)]` | `action` |
| both | `[]` | `action, observation.state` |

Each override applies exactly the half it names, and the diagnostic keeps
reporting whichever half is still unnormalized. On `main` every row except the
first raises `KeyError`.

`tests/policies/lerobot_local/test_processor_override_pipeline_routing.py`
builds both pipelines on disk with LeRobot's own `save_pretrained` (no network)
in the same shape as a base checkpoint, and loads them through the real LeRobot
pipeline. **5 failed / 3 passed on `main`, 8 pass here.** The three that pass on
both are the boundary controls: a typo is still refused, `device_processor` still
reaches both pipelines, and the no-overrides path is byte-identical.

Gate: `ruff check` / `ruff format --check` clean; `mypy strands_robots tests
tests_integ` unchanged (14 pre-existing, all in `examples/isaac_gs/`);
`tests/policies` + `tests/simulation` = 15017 passed with the same 3
environment-bound failures (2 docker-absent VERA, 1 Newton) as `main`.

## Rollout

A 120-step SmolVLA rollout on an SO-101 in MuJoCo (headless), driving the
documented remedy. Left: what `main` does -- the override raises, so the
pipeline stays inert and un-normalized values reach the actuators. Right: the
override applies and the stats the caller supplied are used.

processor override routing

https://raw.githubusercontent.com/cagataycali/robots/media-processor-override-routing/artifact.mp4

Emitted `|action|` mean goes 0.596 -> 81.5 once the checkpoint's own action
stats are applied, i.e. un-normalization is genuinely live rather than a
silent passthrough.

## Out of scope

Whether a hardware-pretrained checkpoint's action units should be reconciled
with a simulated robot's joint units is a separate question and not touched
here: `smolvla_base`'s action stats are in the SO-100 degree convention
(mean ~120, std ~52), so once correctly un-normalized they land far outside a
MuJoCo revolute joint's radian range. This PR only makes the documented remedy
applicable.

### DIFF
--- a/changelog.d/2375-processor-overrides-per-pipeline-routing.md
+++ b/changelog.d/2375-processor-overrides-per-pipeline-routing.md
@@ -1,0 +1,14 @@
+### Fixed
+
+`processor_overrides` now reaches the pipeline that declares the step. A
+checkpoint's preprocessor and postprocessor carry disjoint step keys
+(`normalizer_processor` vs `unnormalizer_processor`), and LeRobot refuses an
+override key that matches no step in the pipeline it is loading, so passing the
+caller's dict to both made every pipeline-specific step unreachable -- only
+`device_processor`, present in both, could be applied. Supplying normalizer
+`stats` is the documented remedy for a pretraining base checkpoint whose
+dataset-prefixed stats leave its declared normalization inert, and that remedy
+raised `KeyError` instead of applying. Each override is now routed to the
+pipeline that declares it, a key declared by neither is still refused with both
+pipelines' step names listed, and the inert-normalization diagnostic names both
+steps.
--- a/docs/policies/lerobot-local.md
+++ b/docs/policies/lerobot-local.md
@@ -310,6 +310,45 @@ producing off-policy / micro-motion trajectories. An explicit
 `processor_overrides={"device_processor": {"device": ...}}` is still honored
 as-is and takes precedence over the automatic reconciliation.
 
+### Overriding a processor step
+
+`processor_overrides` is keyed by step name -- a step's `registry_name`, or its
+class name when the step is not registry-backed. A checkpoint ships two
+pipelines whose step names are mostly disjoint, so each override is routed to
+the pipeline that declares it:
+
+| step | pipeline | what it controls |
+| --- | --- | --- |
+| `normalizer_processor` | preprocessor | `observation.state` normalization |
+| `unnormalizer_processor` | postprocessor | `action` un-normalization |
+| `device_processor` | both | tensor placement |
+
+A key no pipeline declares is refused, and the refusal lists both pipelines'
+step names.
+
+This is how you supply stats to a checkpoint whose declared normalization is
+inert -- a pretraining *base* checkpoint such as `lerobot/smolvla_base` ships
+stats keyed by its training dataset (`so100.buffer.action`) rather than the
+canonical `action` / `observation.state` keys, so LeRobot's normalizer finds no
+matching key and passes those tensors through untouched. Both halves live in
+different pipelines, so a remedy has to name both steps:
+
+```python
+policy = create_policy(
+    "lerobot_local",
+    pretrained_name_or_path="lerobot/smolvla_base",
+    policy_type="smolvla",
+    processor_overrides={
+        "normalizer_processor": {"stats": dataset_stats},    # observation.state
+        "unnormalizer_processor": {"stats": dataset_stats},  # action
+    },
+)
+```
+
+Naming only one leaves the other inert, and the diagnostic keeps reporting
+whichever half is still unnormalized. Fine-tuning the checkpoint writes stats
+under the canonical keys and needs no override at all.
+
 ## State routing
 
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Routes `processor_overrides` per pipeline instead of handing the caller's whole dict to both `_load_pipeline` calls, fixing the pre-existing state where any pipeline-specific step key (`normalizer_processor`, `unnormalizer_processor`) was rejected by whichever pipeline lacked it -- making the documented inert-normalization remedy unapplicable. The new `_pipeline_step_keys` helper derives accepted keys with exactly the same rule as LeRobot's own `DataProcessorPipeline._validate_overrides_used` (`registry_name`, else the final component of `class`), so the router cannot disagree with the guard it is routing around. Every undetermined case (offline, absent config, unreadable, unrecognized shape, missing `huggingface_hub`) degrades to the pre-fix unrouted behavior rather than raising, and the typo guard is preserved -- a key declared by neither pipeline is still refused, now with both pipelines' step names in the message.

## What's good

- The regression tests build both pipelines on disk with LeRobot's own `save_pretrained` and load through the real pipeline (no mocks), with boundary controls for the typo refusal, the both-pipelines `device_processor` key, and the byte-identical no-overrides path -- 5 fail on `main`, 8 pass here.
- Changelog fragment present (`changelog.d/2375-...md`) per AGENTS.md step 3; docstrings carry full `Args:` entries; `revision` is forwarded to `hf_hub_download` consistent with the existing revision-pinning contract.
- The exception surface is careful: `ImportError` on an older `huggingface_hub` without `huggingface_hub.errors` degrades to unrouted rather than crashing, and the diagnostic-message update in `policy.py` now names both steps the remedy requires.

Issue #2093, opened 2026-08-10
decision(simulation/mujoco): the dispatch router accepts 27 actions the tool_spec enum never advertises - 22 capabilities, incl. a fully validated set_obs_noise, are unreachable by an agent


cagataycali:
## What I measured

`MuJoCoSimEngine._dispatch_action` resolves an action with a bare `getattr`:

```python
method_name = self._ACTION_ALIASES.get(action, action)
method = getattr(self, method_name, None)
if method is None or action.startswith("_"):
    return {"status": "error", "content": [{"text": f"Unknown action: {action}"}]}
```

There is no allowlist, so the router dispatches *any* public method by name. The
schema's `action` enum in `tool_spec.json` has **76** entries, while the engine has
**105** public callables. Twenty-seven are dispatchable and absent from the enum;
five of those are `AgentTool`/lifecycle protocol (`cleanup`, `describe`, `stream`,
`mark_dynamic`, `get_display_properties`), leaving **22 domain capabilities** an
agent reading the schema cannot reach:

| capability | owner |
|---|---|
| `get_observation` | `MuJoCoSimEngine` |
| `send_action` | `MuJoCoSimEngine` |
| `set_obs_noise` | `RandomizationMixin` |
| `get_frame`, `get_camera_params` | `RenderingMixin` |
| `start_cameras_recording_synchronous` | `RenderingMixin` |
| `save_episode`, `stream_dataset` | `DatasetRecordingMixin` |
| `verify_dataset_episodes` | `SimEngine` |
| `run_multi_policy`, `bind_policy_sim_context` | `MuJoCoSimEngine` |
| `robot_joint_names`, `robot_action_keys`, `physics_timestep` | `MuJoCoSimEngine` |
| `attach_teleop`, `detach_teleop`, `list_teleops`, `teleoperate`, `stop_teleoperate`, `get_teleoperate_status`, `tool_name_label` | `TeleopMixin` |
| `attachment_involving` | `ManipulationMixin` |

I verified the gap is behavioural, not just schema shape — `set_obs_noise` really
does dispatch:

```
sim(action="set_obs_noise", joint_pos_std=0.01)
-> status=success  "Sensor noise: joint_pos_std=0.01, joint_vel_std=0.0, camera_jitter_px=0.0"
```

`set_obs_noise` is the sharpest case. It is implemented, documented (a full
`Args:` section for all four parameters), tested, and validated (it rejects
[...]

PR #2377, opened 2026-08-17
fix(simulation/mujoco): name the action a misspelled action probably meant

cagataycali:
## Problem

`action` is the one parameter every call to the simulation tool must supply, and the only one with no usable default. It is therefore where a typo is most likely to land -- and it was the one parameter whose refusal named neither a candidate nor a way to find one:

```
sim(action="set_joint_position")  ->  "Unknown action: set_joint_position"
sim(action="get_stat")            ->  "Unknown action: get_stat"
sim(action="list_robot")          ->  "Unknown action: list_robot"
sim(action="renderr")             ->  "Unknown action: renderr"
sim(action="add_objec")           ->  "Unknown action: add_objec"
```

Every one of those is **one character** from a published action, and the package already ships the helper that says so. `close_match_hint(requested, known)` answers all five, and it is already imported in this very file for four other messages -- so a misspelled *robot* got

```
Robot 'ghost' not found. Did you mean: ...? Available robots: ['so101']. Use action='list_robots' to see all.
```

while a misspelled *action* one frame away got a bare dead end. Each of those four sibling docstrings (`_unknown_model_msg`, `_unknown_object_msg`, `_unknown_camera_msg`, `_unknown_robot_msg`) rejects the dead-end form in the same words -- it "forces an agent driving the API blind into a discovery round-trip on every typo" -- and `_unpublished_action_error` itself already classifies this case as "a typo" in its docstring before declining to say which one.

## Change

Adds the fifth caller of the shared helper, `_unknown_action_msg`, beside its four siblings:

```
Unknown action: set_joint_position. Did you mean: set_joint_positions, set_joint_velocities,
set_geom_properties? This tool publishes 77 actions in the 'action' enum of its schema; see
tool_spec for the actions you can use.
```

Both agent-facing entry points (`__call__` and `stream`) share `_unpublished_action_error`, so one change covers both.

Two deliberate details:

**The suggestion corpus is the published enum, not every dispatchable method.** Those differ: the router resolves by `getattr` with no allowlist while the agent-facing entry points refuse a non-enum name (#2093). Suggesting a dispatchable-but-unpublished action -- `send_action`, `robot_joint_names` -- would answer one refusal with a second. A parametrized control over the whole Python-only inventory holds that line.

**The vocabulary size and the pointer travel independently of the suggestion.** `close_match_hint` returns `""` when nothing scores well enough, so a name resembling nothing (`zzzzzzzzzz`) would otherwise stay the dead end this fixes. It now still learns the set is closed and enumerated.

Unchanged: the `Unknown action: <name>` verdict itself, the distinct Python-only refusal (a curated omission must not read as a misspelling), and the wide Python router.

## Demo

A blind agent typos `set_joint_position`, then follows **only what the refusal told it** -- the capture script regexes the `Did you mean:` fragment out of whatever message it receives and calls the first name it finds. It hardcodes no recovery, so the two panels diverge purely because one message carries an answer. SO-101 in MuJoCo, headless EGL, same script under `PYTHONPATH` of an `upstream/main` worktree and of this branch.

unknown-action suggestion demo

[demo.mp4](https://raw.githubusercontent.com/cagataycali/robots/media-unknown-action-suggestion/demo.mp4)

| | before | after |
|---|---|---|
| refusal | `Unknown action: set_joint_position` | `... Did you mean: set_joint_positions, ...` |
| suggestion the agent could parse | *none* | `set_joint_positions` |
| retry | never attempted | `status: success` |
| joint 3 | 0.000 rad (never moved) | **1.150** / 1.150 rad |
| panel motion over the run | **0.0000** mean abs pixel delta | 6.8680 |

The left panel is honestly static: with no suggestion there was nothing to act on.

## Tests

`tests/simulation/mujoco/test_unknown_action_message_suggests_a_published_action.py`, **9 failed / 26 passed on `upstream/main`** -> 35 passed here.

Representative pre-fix diffs:

```
AssertionError: 'set_joint_position' -> 'Unknown action: set_joint_position'
AssertionError: assert '77' in 'Unknown action: zzzzzzzzzz'
AssertionError: assert ' Did you mean: set_joint_positions, set_joint_velocities,
                set_geom_properties?' in 'Unknown action: set_joint_position'
```

That last one is the whole finding: the helper had the answer and the message did not ask for it.

The 26 pre-fix passes are the no-overreach controls, and they matter more than the failures:

- **every suggested name is one the agent can actually call** -- parametrized over the whole Python-only inventory, asserting `offered <= _PUBLISHED_ACTIONS` and `not offered & _PYTHON_ONLY_ACTIONS`. Vacuous before the fix (there were no suggestions to be wrong), load-bearing after.
- a Python-only action keeps its own verdict and gains no `Did you mean`.
- the `Unknown action: <name>` prefix still reads as before, so the three existing assertions on it hold.
- a published action still dispatches.

`_PYTHON_ONLY_ACTIONS` is imported from `test_tool_spec` rather than restated, matching the sibling boundary suite.

The new helper also joins the existing repo-wide `_unknown_*_msg` guard family in `test_unknown_entity_message_name_type.py`. It satisfies that guard's substantive contract unchanged -- it does not gate the listing on the requested name's type -- so only the non-vacuity inventory needed extending (one line).

## Verification

Full `tests/` on both trees, same interpreter, run concurrently (mujoco 3.5.0, lerobot 0.6.2, ~24.5 min each):

| | failing test IDs | passed |
|---|---|---|
| this branch | 28 failed + 24 errors = **52** | 30860 |
| `upstream/main` (with the new test file copied in) | 38 failed + 24 errors = **62** | 30850 |

`comm -23` on the sorted ID sets is **empty** -- zero new failures. `comm -13` is exactly the 9 pre-fix failures above plus `test_hardware_task_loop_dispatch.py::TestTheRolloutIsDrivenExactlyOnce::test_on_the_sync_branch`, a wall-clock flake in a rate-guarded loop (the box was carrying 9 concurrent pytest processes at load 32). The passed delta reconciles: `+10` = 9 from the new file plus that flake.

- `tests/simulation/mujoco` on the merged tree: 3393 passed, 13 skipped.
- `ruff check` / `ruff format --check` over `strands_robots tests tests_integ`: clean. `mypy`: 18 pre-existing errors, unchanged.
- Package sources stay ASCII; changelog fragment validates (`assemble_changelog.py --check`).

`upstream/main` moved during the run (#2369 and #2370 landed, both touching `simulation/mujoco/simulation.py`), so the base is merged in. That merge was verified by symbol rather than by "no conflict": this PR's helper and its call site, #2369's hardware-route wording, and #2370's latched wrench are all present in the merged file, and the affected suite was re-run on the result.

## Out of scope (measured, reported separately)

The `Unknown action` string also appears in `_dispatch_action`, the **Python** router. That path is deliberately wider than the enum, so suggesting from the enum there would be misleading; it is left alone.

Separately, five refusal messages in the simulation backends offer `action='send_action'` or `action='robot_joint_names'` as *the* remedy, and both are refused at the agent boundary. That is a different contract from this one and is filed as its own issue.


yinsong1986 (approved):
## Summary

Adds `_unknown_action_msg` as the fifth caller of the shared `close_match_hint` helper, so a misspelled `action` — the one required parameter with no usable default — now gets a `Did you mean: ...?` suggestion plus the enum size and a `tool_spec` pointer, instead of a bare `Unknown action: X` dead end. Both agent-facing entry points (`__call__` and `stream`) share `_unpublished_action_error`, so the single one-line change at the refusal site covers both. The suggestion corpus is deliberately the published enum rather than every `getattr`-dispatchable method, so a suggestion can never be a name the agent boundary itself refuses (#2093) — and a parametrized control over the whole Python-only inventory pins that line. The trailing period the new message adds after the action name is compatible with every existing assertion on this string (all substring/`startswith` matches on the un-period-ed prefix, in `test_agent_boundary_refuses_unpublished_action.py`, `test_error_paths.py`, and `tests_integ/simulation/test_mujoco_journeys.py`), and the only call site guarantees `requested` is a `str` before the helper runs, so the type annotation is sound.

## What's good

- Scope discipline: the `Unknown action` string in `_dispatch_action` (the deliberately wide Python router) and the distinct Python-only refusal are both left alone, with the reasoning stated in the docstring and the PR description.
- The no-overreach controls (Python-only actions gain no suggestion, published actions still dispatch, verdict prefix unchanged) are load-bearing, not decorative — the enum-vs-dispatchable distinction is exactly where a naive implementation would suggest a second refusal.
- Conventions followed: changelog fragment under `changelog.d/` rather than a CHANGELOG.md append, ASCII-only user-facing strings, no host paths in tests, and the new helper is added to the `_unknown_*_msg` structural guard inventory in `test_unknown_entity_message_name_type.py` so it cannot regress into gating the listing on the requested name's type.

yinsong1986 (approved):
## Summary

Adds `_unknown_action_msg` beside its four `_unknown_*_msg` siblings so a misspelled `action` — the one required parameter with no usable default — is answered with a `Did you mean: ...?` close-match over the published enum, plus the enum size and a `tool_spec` pointer, instead of a bare `Unknown action: X` dead end. The change is a single call-site swap in `_unpublished_action_error`, which both agent-facing entry points (`__call__` and `stream`) share, so one edit covers both. The guard at the top of `_unpublished_action_error` returns `None` for any non-`str` before the new helper is reached, so the `requested: str` contract holds at the only call site. The suggestion corpus is `_PUBLISHED_ACTIONS` (derived from the schema enum), never the wider `getattr` router, and the parametrized control over the whole Python-only inventory (`offered <= _PUBLISHED_ACTIONS`, `not offered & _PYTHON_ONLY_ACTIONS`) holds that line; the Python-only verdict and the `Unknown action: <name>` prefix are pinned unchanged. No security, crash, data-integrity, or wire/API-shape implications: the verdict string an agent might key on is preserved, and the test pins the helper's own output rather than a frozen wording, so the message can evolve with the shared rule.

## What's good

- Scope discipline: the Python router's own `Unknown action` string is deliberately left alone (suggesting from the enum there would mislead), and the adjacent misleading-remedy messages are measured and filed separately rather than folded in.
- The pre-fix controls are the load-bearing tests: the no-overreach parametrization is vacuous before the fix and meaningful after, exactly the shape AGENTS.md asks for.
- Joins the existing `_unknown_*_msg` guard family in `test_unknown_entity_message_name_type.py` with the one-line inventory extension rather than a parallel mechanism.
- Changelog fragment under `changelog.d/`, ASCII-only sources, and `_PYTHON_ONLY_ACTIONS` imported from the sibling suite rather than restated.

PR #2227, opened 2026-08-13
fix(simulation): write a bare output filename into the sandbox instead of refusing it

cagataycali:
## Problem

`validate_output_path` is the single owner of the path guards behind
`render(output_path=...)`, `run_policy(video={"path": ...})` and
`start_cameras_recording(output_dir=...)`. It resolved **every** relative path
against the process CWD:

```python
raw = Path(output_path).expanduser()
...
resolved = raw.resolve(strict=False)      # relative -> CWD
if sandbox_root is not None and not allow_abs:
    if not resolved.is_relative_to(sandbox_root):
        raise ValueError(f"{label} {resolved} is outside the sandbox ...")
```

Under confinement that makes a bare filename **impossible to satisfy**: it is
resolved against a directory the sandbox does not contain, so the only reachable
outcome is a refusal — and the refusal quotes a CWD-absolute path the caller
never supplied.

Measured on Thor against `upstream/main` 83cc5272:

| call | main | this change |
|---|---|---|
| `render(output_path="frame.png")` | `status="error"`, no file written | `status="success"`, PNG in the sandbox |
| the reason it gave | `output_path /tmp/robots-mine-.../frame.png is outside the sandbox` | — |

The quoted path is the CWD. A caller who names no directory has expressed no
directory intent, so resolving against the CWD is a guess, and under confinement
it is a guess that can only fail.

## Change

Anchor a one-component relative name to the sandbox root before resolution:

```python
if sandbox_root is not None and not allow_abs and not raw.is_absolute() and len(raw.parts) == 1:
    raw = sandbox_root / raw
```

Four conditions keep it from widening the sandbox:

- `sandbox_root is not None` — guards-only mode (the historic video/recording
  contract) is untouched.
- `not allow_abs` — the documented opt-out still means CWD-relative.
- `not raw.is_absolute()` and `len(raw.parts) == 1` — only a name with no
  directory component. `sub/frame.png` expresses directory intent and keeps
  resolving against the CWD.

It runs **before** the symlink probe, so that guard inspects the destination that
is actually opened. `..` is refused above it by a check that scans every part, so
that check holds for the joined path too.

`render()`'s docstring now states the behaviour in the same paragraph that
documents the sandbox, so a caller reading the constraint also reads that a bare
name needs no directory.

## Every case measured

| input | confinement | main | this change |
|---|---|---|---|
| `frame.png` | sandbox | refused (outside the sandbox) | -> sandbox root |
| `sub/frame.png` | sandbox | refused | refused |
[...]

PR #2379, opened 2026-08-17
fix(simulation): make a relative output_path refusal name a remedy that works

cagataycali:
## Problem

`validate_output_path` (`strands_robots.simulation.safe_output`) confines an
LLM-supplied artifact destination to a sandbox root. A **relative** destination is
resolved against the process CWD, so under confinement it can only be refused --
and the refusal described a different mistake than the one made:

```
render(output_path="views/front.png")
-> output_path /home/me/my-project/views/front.png is outside the sandbox
   /home/me/.strands_robots/renders
   (set STRANDS_ROBOTS_RENDER_ALLOW_ABS=1 to permit absolute paths,
    or pass a path under the sandbox)
```

The caller passed no absolute path. They are told about an absolute-path opt-in,
against a CWD-absolute path they never supplied.

**Following the remedy it names makes things worse, silently.** Setting
`STRANDS_ROBOTS_RENDER_ALLOW_ABS=1` does lift the refusal -- by disabling
confinement -- so the retry reports `status="success"` and writes the PNG to
`my-project/views/front.png`, leaving the sandbox empty. The caller asked for a
render and got one anywhere except the directory the sink is documented to write
to, with confinement now off for every later call in the process.

Meanwhile the spelling that does work went unmentioned, including the one this
same function implements 25 lines above the raise: a one-component name is
anchored into the sandbox (#2227).

The repo already has this rule. `test_sandbox_refusal_names_the_opt_in_variable.py`
pins that the traversal / metacharacter / backslash refusals must **not** advertise
the opt-in, because "offering it there would send the caller after a remedy that
cannot work". The confinement refusal on a relative path was the case that rule had
not reached.

## Change

`_confinement_refusal` builds the message from the class of path the caller gave.

**Relative** -- state where it was resolved from, and quote the sandbox-anchored
destination to pass instead, keeping the opt-in as the last resort it is:

```
output_path 'views/front.png' is relative, so it resolved against the current
directory /home/me/my-project to /home/me/my-project/views/front.png, which is
outside the sandbox /home/me/.strands_robots/renders
(pass '/home/me/.strands_robots/renders/views/front.png' to write it into the
 sandbox, or a bare output_path with no directory part, which is written into the
 sandbox, or set STRANDS_ROBOTS_RENDER_ALLOW_ABS=1 to write outside the sandbox
 instead)
```

**Absolute** -- byte-for-byte unchanged. Passing an absolute path outside the
sandbox *is* an absolute-path mistake, so the existing wording fits it.

Which destinations are **accepted** does not change. The narrow anchoring contract
(only a one-component name is anchored, pinned by
`test_sandbox_still_rejects_relative_path_with_separator`) still holds; widening it
would be a different change. Applies to every confining sink, so `output_dir`
follows its own `label`.

The agent-facing schema also never published the anchoring rule: `tool_spec.json`
described the sandbox and the absolute-path opt-in, so the only documented way to
save a render was to know and spell the sandbox path. It now states that a bare
filename is written into the sandbox, graded in the tests against the behaviour
rather than trusted as prose. `docs/troubleshooting.md` gains the relative case.

## Evidence

One probe script, same scene and same sandbox, run against `upstream/main` and this
branch. It reads the refusal, parses the remedy out of it, and follows it.

|---|---|---|
| `render(output_path="views/front.png")` | `status=error` | `status=error` |
| remedy the message names | `set ..._ALLOW_ABS=1` (only one offered) | `pass '<sandbox>/views/front.png'` |
| retry after following it | `status=success` | `status=success` |
| PNG written to | `my-project/views/front.png` | `<sandbox>/views/front.png` |
| inside the sandbox | **no** | **yes** |
| sandbox afterwards | **empty** | holds the render |
| confinement afterwards | **disabled** | still on |

relative output_path refusal

Both columns end in `status=success` and both render the same 58 KB frame (the
thumbnails are identical, which is the control -- only the destination differs).

## Tests

`tests/simulation/test_relative_output_path_refusal_is_actionable.py`, run against
`upstream/main`: **9 failed / 6 passed**, then 15 passed here.

The load-bearing assertions are round trips -- the suggested path is parsed back
out of the refusal, passed in, and required to be accepted inside the sandbox --
so they fail for a missing suggestion and for a wrong one. Also exercised end to
end through `render`'s own validator, not just the shared helper.

The 6 that pass on **both** trees are the no-overreach controls:

- an absolute destination outside the sandbox keeps its exact message,
- a multi-component relative path is **still refused** (this changes wording, not
  what is accepted),
- the opt-in variable is still named, so `test_sandbox_refusal_names_the_opt_in_variable`
  keeps holding,
- the traversal / metacharacter / backslash refusals gain no sandbox suggestion
  (they fire in every mode, so a confinement remedy could not help there).

## Verification

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`;
`mypy` over the same scope reports its 18 pre-existing errors, unchanged.

Blast radius `tests/simulation tests/tools tests/rendering` plus the repo-wide
docstring-xref / ASCII / unicode-dash / args-docstring / dangling-doc-ref /
changelog guards, run on this branch and on an `upstream/main` worktree with the
new test file copied in (which is why the base column has more failures):

```
branch          1 failed, 13781 passed, 186 skipped
upstream/main   9 failed, 13773 passed, 186 skipped
```

Both collected 13782. The 9 base-only ids are exactly the pre-fix failures listed
above. The one branch-only id is
`test_policy_runner_async_rtc.py::test_async_rtc_masks_inference_latency`, a
wall-clock assertion that trips under load: it passes 3/3 in isolation on **both**
trees, and the module contains no reference to `safe_output`, `output_path` or
`tool_spec`.


yinsong1986 (approved):
## Summary

Message-only fix to `validate_output_path`'s confinement refusal: a relative `output_path`/`output_dir` now gets a refusal that names where the path was resolved from (the process CWD) and quotes the sandbox-anchored destination to pass instead, keeping the `*_ALLOW_ABS` opt-in as an explicitly-labelled last resort; the absolute-path refusal is byte-for-byte unchanged. The set of accepted destinations does not change - the confinement check itself is untouched, only the string built inside the `except ValueError` branch - and the controls pin that from both directions (multi-component relative paths still refused, absolute wording unchanged, opt-in variable still named, traversal/metacharacter/backslash refusals gain no sandbox suggestion). The new `_confinement_refusal` helper is private, the `tool_spec.json` change is description prose only (no parameter shape change), so there is no one-way door here. The suggested path is built after `_reject_unsafe_chars` and the `..` rejection have already run, so nothing unsafe can be echoed into the remedy, and the symlinked-sandbox-root edge (where the anchored bare name is itself the failing destination) is handled and tested rather than looping the caller.

## What's good

- Round-trip tests parse the suggested path back out of the refusal and require the sink to accept it inside the sandbox, so they pin the remedy rather than the wording - and the end-to-end variant goes through `render`'s own `_validate_render_output_path`.
- The tool_spec claim ("a bare filename is written INTO that sandbox") is graded against `validate_output_path`'s behaviour in the same test, not trusted as prose.
- Verified locally at head: the new suite plus `test_sandbox_refusal_names_the_opt_in_variable.py`, `test_refusal_messages_never_raise.py`, `test_output_path_sandbox.py`, `test_render_output_path*.py`, `test_export_xml_path_safety.py`, `test_tool_spec.py` and the agent-tool parameter-description guard all pass (299 tests).
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; new files are ASCII-clean with no host paths.

PR #2380, opened 2026-08-17
fix(simulation/mujoco): carry a latched joint force across a scene grow

cagataycali:
Follow-up to #2370, for the one applied-force cell it left uncarried.

## Problem

`spec.recompile` transfers `qpos`, `qvel`, `ctrl`, `act` and the clock but **neither** applied-force buffer. Measured on MuJoCo 3.5, growing a two-body scene by one body:

```
qpos CARRIED   qvel CARRIED   ctrl CARRIED   act CARRIED   time CARRIED
qfrc_applied DROPPED    xfrc_applied DROPPED
```

#2370 carries `xfrc_applied` across both rebuild directions. `qfrc_applied` is still dropped on the grow path, so the same latched generalized force survives a rebuild that **shrinks** the scene - `eject_robot_from_scene` snapshots it with the rest of the joint state - and vanishes on one that **grows** it:

```
latch qfrc_applied = 0.5 N m on 'spinner/spin'
remove_robot('doomed')  -> 0.5   kept
add_object('crate')     -> 0.0   dropped
```

A hinge under that torque stops turning the moment anything enters the scene, under a `"status": "success"`. Unlike the buffers `spec.recompile` does carry, this is not a tail problem, so `_recompile_preserving_state`'s tail initialization cannot help: every entry is lost, including the slices of joints that never moved.

The asymmetry also leaves #2370's new `_SceneState` docstring true on one path and false on the other:

> `qfrc_applied` above is its joint-space sibling; both are part of the state `save_state` checkpoints, so both are carried.

`save_state` does checkpoint `qfrc_applied`, precisely so a restore reinstates "latched external forces, not just positions" - which is why this is a defect rather than a design question.

## Change

Adds the joint-space siblings of #2370's helpers, `_snapshot_joint_forces` and `_restore_joint_forces`, called from the same seam in `_recompile_preserving_state`. Keyed the way `_SceneState` already keys joints, because a rebuild renumbers dofs. Only non-zero slices are carried, so a scene with no latched force is untouched and the restore is a no-op there. A joint whose type changed is skipped rather than written with a mismatched slice. The two docstrings that describe the pair are corrected to say which path carries what.

## Verification

The force is measured by its **effect, from rest**. This matters: a joint that is already turning keeps turning once its drive is gone, so an angle that merely kept increasing would not distinguish a live force from momentum. Zeroing the velocity first is what makes the assertion about the force.

`tests/simulation/mujoco/test_scene_grow_preserves_joint_forces.py` against `main` (`eff27242`, i.e. with #2370 already in) vs this branch: **4 failed / 3 passed -> 7 passed**.

```
FAILED ...::test_the_joint_keeps_being_driven[add_object]
FAILED ...::test_the_joint_keeps_being_driven[add_camera]
  spec.recompile drops qfrc_applied, so a latched joint torque stopped driving
  the joint as soon as anything entered the scene
FAILED ...::test_it_survives_a_grow_that_adds_a_robot
FAILED ...::test_the_two_rebuild_directions_now_agree
  a rebuild that shrinks the scene kept 0.5 but one that grows it kept 0.0;
  the same latched force must survive both
```

The 3 that pass on both trees are the no-overreach controls: a scene with no latched force gains none, a newly added robot's joint does **not** inherit the force latched on another robot, and `reset()` still clears it.

`tests/simulation`, same interpreter (MuJoCo 3.5, lerobot 0.6.2), run sequentially:

| | passed | failed |
|---|---|---|
| `main` (`eff27242`) | 10884 | 1 |
| this branch | 10892 | 0 |

The counts reconcile exactly: `10884 + 1 = 10885` vs `10892` is `+7`, the new tests. That single failure on `main` did not reproduce on an isolated re-run of the same selection, so I am reporting it as a flake rather than claiming this PR fixes it. `ruff check`, `ruff format --check` and `mypy strands_robots tests tests_integ` are clean over the CI scope (18 pre-existing errors, unchanged).

No visual artifact: this is the joint-space half of a buffer whose user-visible rollout #2370 already demonstrated, and `qfrc_applied` has no public setter today (it is reached through `save_state`/`load_state`), so the honest demonstration is the numeric asymmetry above rather than a rendered scene.

### DIFF
--- a/changelog.d/2380-carry-joint-forces-across-a-scene-grow.md
+++ b/changelog.d/2380-carry-joint-forces-across-a-scene-grow.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Simulation (MuJoCo)**: a generalized force latched in `qfrc_applied` is no longer dropped when the scene grows. `spec.recompile` carries neither applied-force buffer, and while `xfrc_applied` is snapshotted by name across a rebuild, `qfrc_applied` was not - so the same latched force survived `remove_robot` (0.5 N m kept) and vanished on `add_object`, `add_camera` or `add_robot` (0.0), under a `"status": "success"`. Both buffers are now carried across both rebuild directions.
--- a/strands_robots/simulation/mujoco/scene_ops.py
+++ b/strands_robots/simulation/mujoco/scene_ops.py
@@ -283,6 +283,70 @@ def _snapshot_body_wrenches(model: Any, data: Any, mj: Any) -> dict[str, list[fl
     return wrenches
 
 
+def _snapshot_joint_forces(model: Any, data: Any, mj: Any) -> dict[_JointKey, list[float]]:
+    """Capture every latched generalized force, keyed by joint.
+
+    The joint-space sibling of :func:`_snapshot_body_wrenches`. ``spec.recompile``
+    drops ``qfrc_applied`` for the same reason it drops ``xfrc_applied`` -- it
+    carries neither applied-force buffer -- so a grow needs this pass just as the
+    eject does. Keyed the same way as :class:`_SceneState`'s joints, because a
+    rebuild renumbers dofs. Only non-zero slices are captured: all-zero is "no
+    force latched here", which a fresh allocation already holds.
+
+    Args:
+        model: The compiled model the slices are being read against.
+        data: The ``MjData`` holding the latched forces.
+        mj: The resolved ``mujoco`` module.
+
+    Returns:
+        ``joint key -> qfrc_applied`` slice, at the dof width the joint uses.
+    """
+    forces: dict[_JointKey, list[float]] = {}
+    for jid in range(int(model.njnt)):
+        key = _joint_key(model, jid, mj)
+        if key is None:
+            continue
+        dof_adr = int(model.jnt_dofadr[jid])
+        _, dof_w = _joint_state_widths(int(model.jnt_type[jid]), mj)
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes the asymmetry #2370 left: `spec.recompile` carries neither applied-force buffer, and while `xfrc_applied` was already snapshotted/restored by name across a grow, `qfrc_applied` was only carried on the shrink path (via `_SceneState`), so a latched generalized joint force vanished the moment anything entered the scene under a `"status": "success"`. The fix adds `_snapshot_joint_forces` / `_restore_joint_forces` as exact joint-space siblings of the existing wrench helpers, called from the same seam in `_recompile_preserving_state`, keyed by the same `_JointKey` scheme the eject path already uses (so unnamed freejoints are covered and a rebuild's dof renumbering is irrelevant), with a width/type guard that skips rather than writes a mismatched slice. All new surface is module-private, so no API, wire-format, or persisted-schema exposure.

## What's good

- The restore is ordered before `mj_forward`, matching the wrench restore, and the `_have_state` guard covers the no-model path.
- Tests measure the force by its effect from rest (velocity zeroed first), which is the assertion that actually distinguishes a live force from momentum; the three controls pin no-overreach (no force invented, no cross-robot inheritance, `reset()` still clears).
- Changelog fragment present under `changelog.d/` per AGENTS.md step 3; no host paths or non-ASCII characters in the new files.
- The two docstrings that described the buffer pair are corrected in the same diff, so `_SceneState`'s claim is now true on both paths.

PR #2381, opened 2026-08-17
fix(policies): gate every spelling of a remote-code provider, not just its canonical name

cagataycali:
## Problem

`_check_trust_remote_code` membership-tests its argument against `_HF_REMOTE_CODE_PROVIDERS`, a set of **canonical** provider names, and `create_policy` hands it whatever `_resolve_policy_class` returned. That function documents its first return value as `canonical_provider_name`, but its third stage returned the caller's spelling unchanged. A provider's declared alias or shorthand therefore reached the gate as a name the set does not hold, the membership test missed, and the provider was constructed with no opt-in.

Three of the five spellings the registry accepts for a gated provider skipped the gate entirely:

| `create_policy(...)` | resolves to | before | after |
|---|---|---|---|
| `"kimodo"` | kimodo | refused | refused |
| `"kimodo_g1"` | kimodo | **constructed, gate skipped** | refused |
| `"text2motion"` | kimodo | **constructed, gate skipped** | refused |
| `"lerobot_local"` | lerobot_local | refused | refused |
| `"lerobot"` | lerobot_local | **constructed, gate skipped** | refused |

Neither provider is incidental: both load HuggingFace models with `trust_remote_code=True`, which is arbitrary code execution from the model repository. Neither spelling is obscure either -- `text2motion` is the natural-language shorthand `policies.json` advertises for Kimodo, and `lerobot` is the obvious short name for `lerobot_local`. There was no refusal and no warning; the gate simply did not run.

`tests/policies/kimodo/test_remote_code_execution_is_the_callers_decision.py` describes this gate as deciding "whether the provider may be constructed at all", and its controls are documented as pinning that "the coarse provider gate is untouched". For three of five spellings there was no coarse gate to be untouched.

The contract violation was wider than the gate. **Fourteen** accepted spellings reported a non-canonical name, so adding an alias to a gated provider -- or gating a provider that already has one -- would have inherited the same bypass.

## Change

`_resolve_policy_class` stage 3 now returns the canonical name it documents. Stage 1 (the runtime alias map) and stage 2 (`resolve_policy`'s shorthand stage) already canonicalise; stage 3 was the one that did not.

The canonicalisation rule itself was already written twice in `registry/policies.py` -- `get_policy_provider` and `import_policy_class` each did `alias_map.get(name, name)` inline. Both now call one `_canonical_provider_name`, so there is a single place that rule lives rather than a third copy in the factory.

Only the reported name changes. The resolved class and kwargs are untouched on every path.

## Tests

`tests/policies/test_trust_gate_covers_every_accepted_spelling.py`, keyed on the live `_HF_REMOTE_CODE_PROVIDERS` and alias map so a newly gated provider or a newly declared alias is covered without editing the test.

Pre-fix on `main`: **17 failed / 46 passed**. Post-fix: **63 passed**.

- 3 failures are the gate bypass (`DID NOT RAISE UntrustedRemoteCodeError` for `kimodo_g1`, `text2motion`, `lerobot`).
- 14 are the root cause: `_resolve_policy_class('<spelling>') reported '<spelling>', which is not a canonical provider name; its docstring promises canonical_provider_name`.

The 46 that pass on both trees are the controls, and they are the point of review:

- **opting in still admits every gated spelling** -- fails if the gate is ever made unconditional, which is the obvious wrong way to close a bypass.
- **an ungated provider is never refused** -- parametrized over every accepted spelling that resolves to an ungated provider, so canonicalising cannot pull `sonic`, `random`, `moveit`, `c3` and friends into the gate.
- **an unknown spelling is reported, not canonicalised** -- a name no provider declares must survive to the refusal so that refusal can quote what the caller typed.
- a premise assertion fails loudly if no gated provider has a non-canonical spelling, so a clean run cannot become vacuous.

The test file deliberately imports no symbol this PR adds; it asserts through `create_policy`, `_resolve_policy_class` and the pre-existing alias map, so it collects on `main` and its pre-fix run is behavioural rather than a collection error.

## Verification

Rebased on current `main` (merge of `b3939ca8`), which had moved onto both files this PR touches. The composition was checked by symbol rather than by absence of conflict: the new canonicalisation helper has exactly one definition and both `registry/policies.py` call sites route through it with no inline copies left, and `list_policy_aliases` / `list_aliases` from #2372 are intact.

- Blast radius on the merged tree -- `tests/policies tests/registry tests/tools tests/test_robot_factory.py`, `tests/simulation/test_unresolvable_policy_provider.py` and the repo-wide docstring/ASCII/dependency guards: **7653 passed, 0 failed**.
- Full `tests/` on both trees: **52 failing ids on the branch vs 71 on `main`**, branch-only set empty. The 19 base-only ids are the 17 pre-fix failures above plus two wall-clock flakes in `test_hardware_task_loop_dispatch.py`; passed count reconciles exactly (+19).
- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`. `mypy` reports 19 errors on the branch and 19 on current `main`, with an empty branch-only set.
- CodeQL clean (0 open alerts). The skip guard in the new test file initially tripped `py/uninitialized-local-variable`: a liveness analysis scoped to one function cannot see that `pytest.skip` terminates. The handler now raises `pytest.skip.Exception` explicitly, which is the same skip by a route a reader of that function alone can follow. The pre-fix split is unchanged by that refactor -- re-measured 17 failed / 46 passed on current `main`, so the test was not weakened to clear the alert.

## Artifact

Rollout in MuJoCo (headless EGL). Top row is each tree's own verdict for all five spellings; middle row is the real transcript of the same call; bottom row is the control -- with `STRANDS_TRUST_REMOTE_CODE=1`, `create_policy("text2motion", motion_agent=...)` still builds a `KimodoPolicy` and drives a Unitree G1 through 120 actions x 29 joints of a real NVIDIA Kimodo sample for the prompt *"a person throwing boxing punches"*.

Those renders are identical on both trees (mean |delta| 0.00007), which is what makes the bottom row a control: the opt-in path is untouched. Pose-to-pose mean |delta| 8.0-8.5 confirms the motion is playing rather than a repeated still.

trust-gate coverage before and after


yinsong1986 (dismissed):
## Summary

Closes a real trust-remote-code gate bypass: `_resolve_policy_class` stage 3 returned the caller's spelling where its docstring promises `canonical_provider_name`, and `_check_trust_remote_code` membership-tests a set of canonical names, so `create_policy("lerobot")`, `create_policy("kimodo_g1")` and `create_policy("text2motion")` constructed providers that load HF models with `trust_remote_code=True` without the `STRANDS_TRUST_REMOTE_CODE` opt-in. The production diff is minimal and correct: stage 3 now canonicalises via a single `_canonical_provider_name` helper that `get_policy_provider` and `import_policy_class` had each restated inline (both refactors are behaviour-preserving), and the resolved class and kwargs are untouched on every path. The other two consumers of the returned name (`preflight_policy`, `policy_provider_error`) discard it, so the rename-on-return has no blast radius beyond the gate it fixes. I reproduced the claimed verification independently: the new test file reports 17 failed / 46 passed on the merge base and 63 passed on the branch, and `tests/policies tests/registry` on the branch shows no failure attributable to this PR (the one red test fails identically on the merge base; two collection errors are a local protobuf version issue, also present on base).

## What's good

- The registry-keyed test design (live `_HF_REMOTE_CODE_PROVIDERS` + alias map) means a newly gated provider or newly declared alias is covered without editing the test, and the premise assertion prevents the suite going vacuous.
- Controls pin both wrong ways to close the bypass: making the gate unconditional, and pulling ungated providers into it. The unknown-spelling control keeps refusals quoting what the caller typed.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; new code is ASCII-only, no host paths, `monkeypatch.setenv`/`delenv` used per AGENTS.md testing patterns.

yinsong1986 (dismissed):
## Summary

Closes a verified trust-remote-code gate bypass. `_resolve_policy_class` stage 3 returned the caller's spelling where its docstring promises `canonical_provider_name`, and `_check_trust_remote_code` membership-tests `_HF_REMOTE_CODE_PROVIDERS` (canonical names only), so `create_policy("lerobot")`, `create_policy("kimodo_g1")` and `create_policy("text2motion")` constructed providers that load HuggingFace models with `trust_remote_code=True` without the `STRANDS_TRUST_REMOTE_CODE` opt-in. I reproduced the bypass on current `main` (all three spellings constructed, gate skipped) and confirmed the branch refuses all five accepted spellings of both gated providers while the opt-in path still admits them. The fix is minimal and correct: stage 3 now routes through `_canonical_provider_name`, and the two inline copies of the same rule in `registry/policies.py` (`get_policy_provider`, `import_policy_class`) now call the one helper - a behavioral no-op for both, confirmed by the unchanged 892-test pass over `tests/registry` and the factory/alias/kimodo suites. Only the reported name changes; the resolved class and kwargs are untouched on every path, and the gate's single caller (`create_policy`) is the only consumer of the name. No public API, wire-format, or env-var surface changes (`_canonical_provider_name` is private and not exported), the changelog fragment is present per AGENTS.md step 3, and the new test file (63 passed at this head) is keyed on the live gate set and alias map with a premise assertion that keeps it from going vacuous. No blocking concerns.

## What's good

- The regression test parametrizes over the live `_HF_REMOTE_CODE_PROVIDERS` and alias map, so a newly gated provider or newly declared alias is covered without editing the test - and the premise test fails loudly if the parametrization ever becomes vacuous.
- The controls pin both wrong ways to close the bypass: making the gate unconditional, and pulling ungated providers into it.
- Deduplicating the canonicalisation rule into one helper follows AGENTS.md convention 11 (a shared rule earns a shared name at the second caller - this one has three).
- Tests use `monkeypatch.setenv`/`delenv`, contain no host paths, and the added lines are ASCII-clean.

yinsong1986 (approved):
## Summary

Closes a verified trust-remote-code gate bypass: `_resolve_policy_class` stage 3 returned the caller's spelling where its docstring promises `canonical_provider_name`, and `_check_trust_remote_code` membership-tests a set of canonical names, so `create_policy("lerobot")`, `create_policy("kimodo_g1")` and `create_policy("text2motion")` constructed providers that load HF models with `trust_remote_code=True` without the `STRANDS_TRUST_REMOTE_CODE` opt-in. I reproduced the bypass on current `main` (all three spellings constructed with no gate) and confirmed it closed at this head (all five spellings of both gated providers refused without opt-in); the new test file passes 63/63 on the head tree. The registry data backs the claim exactly: `kimodo` declares `kimodo_g1`/`text2motion` shorthands and `lerobot_local` declares the `lerobot` alias. Only the reported name changes -- the resolved class and kwargs are untouched on every path, and the two inline copies of the canonicalisation rule in `registry/policies.py` now route through the one shared helper.

## What's good

- The refusal is loud and names the remedy (`STRANDS_TRUST_REMOTE_CODE=1`), so the behaviour change for previously-constructible spellings is not a silent default change.
- Controls are the strong part of the test file: opt-in still admits every gated spelling (the gate cannot quietly become a ban), ungated aliases are never pulled into the gate, and an unknown spelling survives uncanonicalised to the refusal that quotes it.
- The premise assertion (`test_a_gated_provider_really_has_a_non_canonical_spelling`) keeps the suite from going vacuous if aliases are ever removed.
- Tests are keyed on the live `_HF_REMOTE_CODE_PROVIDERS` and alias map, so a newly gated provider or new alias is covered without editing the test.
- Diffed against the current `main` tip (which includes #2374's `registry/policies.py` changes), the effective delta is exactly this PR's diff -- the rebase composed cleanly.
- Changelog fragment present per AGENTS.md step 3; new functions carry per-parameter `Args:` docstrings; test file is ASCII-clean with no host paths.

## Verification suggestions

- Pre-fix reproduction on `main`: `python -c "from strands_robots.policies.factory import create_policy; create_policy('text2motion')"` with `STRANDS_TRUST_REMOTE_CODE` unset constructs a `KimodoPolicy`; the same command on this branch raises `UntrustedRemoteCodeError`.
- `pytest tests/policies/test_trust_gate_covers_every_accepted_spelling.py -q` on the branch: 63 passed.

Issue #2378, opened 2026-08-17
A refusal hint names an action the agent boundary refuses (15 sites: send_action x11, robot_joint_names x4)


cagataycali:
## Summary

Fifteen refusal messages in the simulation backends tell the caller to recover by using an action the agent-facing entry points refuse. The advice is the *only* remedy each message offers, so following it produces a second refusal rather than progress.

```
sim(action="send_action")
-> Action 'send_action' is not available to an agent. It exists on this simulation but is
   deliberately not published in tool_spec, so it is reachable from Python only.

sim(action="robot_joint_names")
-> Action 'robot_joint_names' is not available to an agent. ...
```

Both names are outside `tool_spec.json`'s `action` enum, and #2093 made the two agent-facing entry points (`__call__`, `stream`) refuse a non-enum action. The hints predate that boundary, so this is drift the boundary exposed rather than a mistake at the time they were written.

## The sites

Found by extracting every `action='X'` string literal under `strands_robots/simulation/` and testing `X` against the enum. Both backends, so it is not a per-backend slip.

**`action='send_action'` (11)**
```
simulation/isaac/motion_primitives.py:381, 449, 461, 1102, 1119, 1250
simulation/motion_primitives_base.py:320
simulation/mujoco/motion_primitives.py:360, 925, 942, 1070
```
These are the fallbacks for `move_to` / `set_gripper` / `rotate_wrist` when the primitive cannot resolve what it needs, e.g.:
```
set_gripper: could not resolve a gripper actuator for 'X' - the registry carries no gripper
metadata for this robot and no actuator/joint name contains ['gripper', 'finger', 'jaw'].
Actuators: [...]. Drive it directly with action='send_action' instead.
```
The refusal is otherwise excellent - it names the robot, the hints it tried, and the actuators it found. The one thing it asks the caller to do is the one thing the caller cannot do.

**`action='robot_joint_names'` (4)**
```
simulation/isaac/simulation.py:5959, 5974
simulation/mujoco/physics.py:885, 1318
```
Reached from a `set_joint_positions` key that is not a joint, which is a common case - it is where an `int` key or a wrong-embodiment name lands:
```
sim(action="set_joint_positions", positions={1: 0.4}, robot_name="so101")
-> set_joint_positions: 1 of 1 'positions' keys are not joints in this model, so nothing was
   written (the write is all-or-nothing). Joint '1' not found. Available joints: [...].
   Use action='robot_joint_names' to see one robot's joints.
```

[...]

PR #2382, opened 2026-08-17
fix(simulation/mujoco): offer a joint-discovery action an agent can call

cagataycali:
## Problem

Four agent-reachable refusals ended by pointing at an action the agent boundary refuses.

`set_joint_positions` / `set_joint_velocities` take a `{joint_name: value}` mapping, and a name the model cannot resolve is refused rather than skipped (the write is all-or-nothing), so the refusal has to say where the real joint names come from. Both of them said:

```
set_joint_positions: 'positions' is empty, so there is nothing to write. Pass at least
one joint (dict form) or a full ordered vector (list form); use action='robot_joint_names'
to see one robot's joints.
```

`robot_joint_names` is a deliberate Python-only capability. It is not in `tool_spec.json`'s `action` enum, and the agent-facing entry points refuse a non-enum action outright:

```
Action 'robot_joint_names' is not available to an agent. It exists on this simulation but
is deliberately not published in tool_spec, so it is reachable from Python only. See
tool_spec for the actions you can use.
```

So a caller that read the remedy and followed it verbatim was refused a second time, and that second refusal names no alternative either — a closed loop, on the discovery step of writing a pose. Measured on `main`, driving only published actions:

| step | call | result |
|---|---|---|
| 1 | `set_joint_positions(positions={'shoulder_pan': 0.9})` | error, offers `action='robot_joint_names'` |
| 2 | follow it: `action='robot_joint_names'` | error, *"not available to an agent"* |
| — | joints discovered | **0** |
| — | pose written | **no**, `qpos` still all `0.0` |

Four refusals reach this: an empty mapping and an unresolvable mapping, for each of `set_joint_positions` and `set_joint_velocities` (they share `_resolve_joint_write_names` and `_unknown_mj_entity_msg`).

## Change

`_unknown_mj_entity_msg` already had the answer one branch above. Its `Body` arm offers `action='list_bodies'`, which **is** published; the `Joint` arm is the same sentence about a different entity kind, so the two only had to agree:

```python
if kind == "Body":
    msg += " Use action='list_bodies' to see all."
elif kind == "Joint":
    msg += " Use action='get_robot_state' to see one robot's joints."   # was robot_joint_names
```

`get_robot_state` is published, is accepted with or without `robot_name`, and reports every joint of a robot by name with its position and velocity — which is exactly what the sentence promises. Two string literals in `physics.py`; no signature, no behaviour, and no schema change.

`robot_joint_names` is deliberately left **Python-only** and left **dispatchable**: publishing it would widen the agent surface to settle a wording bug, and deleting it would break `policy_runner` and the Python callers. Both are pinned as controls.

## Verification

Same script on both trees, guessing a joint name and then doing whatever the refusal says — nothing about the remedy is hardcoded, it is parsed back out of the live message:

joint-discovery hint before and after

| | `main` | this change |
|---|---|---|
| remedy offered | `robot_joint_names` | `get_robot_state` |
| following it | **error** — not available to an agent | success |
| joints discovered | 0 | **6** |
| pose written | **no** | yes |
| final `qpos` | all `0.0` | all `0.9` |
| render delta over the attempt | **0.0000** (pixel-identical: the arm never moved) | 8.3435 |

Tests: `tests/simulation/mujoco/test_joint_discovery_hint_is_followable.py`, **7 failed / 4 passed on `main` → 11 passed here**. The failure text is the finding:

```
the refusal offers action='robot_joint_names', which tool_spec does not publish, so an
agent that follows it is refused again
```

The headline test parses `action='…'` out of the live refusal and *invokes* it, so it fails for a missing name, a wrong name, or a name the boundary refuses — not for a wording change. Two structural guards come with it: nothing in `physics.py` offers an unpublished action, and the set of unpublished actions offered anywhere in the MuJoCo backend is pinned (post-fix it is exactly `{send_action}`), so a new unfollowable remedy cannot arrive in silence.

The four tests that pass on both trees are the boundary: the `Body` branch still offers `list_bodies`, `robot_joint_names` stays Python-only *and* stays dispatchable, and the unresolved-joint refusal still names the offending key and lists the available joints — redirecting the hint must not cost the caller the rest of the diagnosis.

One existing test needed updating, and it is worth noting why. `test_joint_setter_unknown_names_rejected` asserted the bare substring `robot_joint_names` under the comment `# discovery pointer`. The intent — the refusal must close by naming somewhere to look — is unchanged, but a substring check on the name cannot tell a followable pointer from an unfollowable one, which is why it passed for as long as the hint named a Python-only action. It now matches the offered form `action='get_robot_state'` and asserts that name is in `_PUBLISHED_ACTIONS`, so the "published action" half of its comment is machine-checked. That assertion also fails on `main`, so the repaired test detects this defect rather than merely tolerating the fix: the full pre-fix split is **8 failed**.

Full `tests/` was run on both trees (`MUJOCO_GL=egl`): 4 failed / 31103 passed on the branch against 10 failed / 31097 passed on `main` carrying the new test file. The ID diff is the useful part — base-only was exactly the 7 pre-fix failures, and the three shared failures are pre-existing and environment-bound (two Vera cases needing Docker, one Newton camera message). That pair also surfaced the one existing test above, which the scoped run had missed; after repairing it `tests/simulation/mujoco` is 3398 passed / 0 failed. `ruff check`, `ruff format --check` and `mypy` are clean over `strands_robots tests tests_integ`.

## Scope

The 11 sites that offer `action='send_action'` are **not** touched. `send_action` has no published equivalent — `set_joint_positions` bypasses the actuators and `actuate_robot` rewrites an actuator-less arm — so choosing one is a contract decision (#2378), not a rename. The new inventory guard keeps that set visible rather than hidden, which is the same job `_PYTHON_ONLY_ACTIONS` does for the dispatch surface.

The Isaac backend carries the same sentence, and it is a different situation: `IsaacSimulation` defines no `__call__` and ships no tool spec, so it is not agent-callable and nothing refuses the name there — and it has no `get_robot_state` to point at. Converging it needs that parity gap closed first, so it is left alone here rather than redirected at a method the backend does not have.


yinsong1986 (dismissed):
## Summary

Redirects the joint-discovery hint in the four `set_joint_positions` / `set_joint_velocities` refusals (empty mapping and unresolved mapping, via the shared `_resolve_joint_write_targets` / `_unknown_mj_entity_msg` paths) from `robot_joint_names` — which the agent boundary refuses as a Python-only action — to `get_robot_state`, which is in the tool-spec `action` enum, accepts `robot_name`, and reports every scalar joint of a robot by name with position and velocity. That closes the measured dead loop where following the remedy verbatim produced a second refusal naming no alternative. The production change is two string literals; `robot_joint_names` stays Python-only and dispatchable, both pinned as controls.

## What's good

- Verified at head `f0d965f0`: the new suite passes 11/11, and the adjacent `test_tool_spec.py` + `test_agent_boundary_refuses_unpublished_action.py` suites pass (98 tests) — no boundary regression.
- The headline test parses the offered `action='…'` out of the live refusal and invokes it, so it pins followability rather than wording; the backend-wide inventory guard pins the unpublished-in-refusals set to exactly `{send_action}` (independently confirmed by scanning the head tree), so a new unfollowable remedy cannot land silently.
- Scope discipline: the `send_action` sites are correctly deferred to #2378 as a contract decision, and the Isaac backend is left alone with a stated reason (no `__call__`, no tool spec, no `get_robot_state` to point at).
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; docs updated in the same PR; no host paths, ASCII-only strings, fixture/teardown pattern matches the existing MuJoCo test suite.

yinsong1986 (approved):
## Summary

Redirects the joint-discovery remedy in the four `set_joint_positions` / `set_joint_velocities` refusals (empty mapping and unresolved mapping, via the shared `_resolve_joint_write_targets` / `_unknown_mj_entity_msg` paths) from `robot_joint_names` — a Python-only action the agent boundary refuses — to `get_robot_state`, which is in the tool-spec `action` enum and reports every joint of a robot by name with position and velocity. The production change is exactly two string literals in `physics.py`; no signature, behaviour, or schema change, so there is no one-way-door surface here (refusal wording is not a contract). Verified against the head tree: `get_robot_state` is in `tool_spec.json`'s enum, the only remaining unpublished action offered anywhere in the MuJoCo backend is `send_action` (correctly deferred to #2378 as a contract decision), and the Isaac backend's identical sentence is left alone with a sound stated reason (no `__call__`, no tool spec, no `get_robot_state` to point at).

## What's good

- The headline test parses the offered `action='…'` out of the live refusal and invokes it, so it pins followability rather than wording; the backend-wide inventory guard (`_KNOWN_UNPUBLISHED_IN_REFUSALS == {"send_action"}`) stops a new unfollowable remedy landing silently.
- Controls pin both halves of the non-change: `robot_joint_names` stays Python-only and stays dispatchable, and the unresolved-joint refusal keeps the offending key and the joint list.
- The repaired `test_joint_setter_unknown_names_rejected` now asserts the pointer is in `_PUBLISHED_ACTIONS` rather than a bare substring, so the "published action" half of its comment is machine-checked and the test fails on pre-fix code.
- AGENTS.md compliance: changelog fragment under `changelog.d/`, docs updated in the same PR, ASCII-only strings, no host paths, fixture teardown matches the existing MuJoCo suite.
- Ran the new suite plus the updated setter guard at head (`MUJOCO_GL=egl`): 18 passed, 0 failed.

PR #2383, opened 2026-08-17
fix(simulation/mujoco): restore the setpoints a keyframe pairs with its pose

cagataycali:
## Problem

A MuJoCo `<key>` declares `qpos` and `ctrl` together on purpose: the `ctrl` are the position-servo setpoints its author matched to that pose so the pose **holds itself** against gravity. That is why `mj_resetDataKeyframe` -- MuJoCo's own way to apply a keyframe -- sets `ctrl` as well as `qpos`.

`add_robot(keyframe=...)` captured only the `qpos`, and `reset()` restored only the `qpos`. So the robot was placed **at** its canonical home pose with every servo commanded to the **zero** configuration, and then driven off it on the first step. Reported as a successful keyframe spawn, with nothing logged.

Measured on a Franka Panda at its shipped `home` key (`add_robot(data_config="panda", keyframe="home")`):

```
key p/home  qpos = [0, 0, 0, -1.5708, 0, 1.5708, -0.7853, 0.04]
            ctrl = [0, 0, 0, -1.5708, 0, 1.5708, -0.7853, 255]   <- 4/8 nonzero

spawn       qpos = [0, 0, 0, -1.5708, 0, 1.5708, -0.7853, 0.04]  <- pose applied
            ctrl = [0, 0, 0,  0,      0, 0,       0,        0]   <- setpoints dropped
step(200)   qpos = [0.0001, 0.013, 0.0001, -0.1126, 0.0001, 0.1961, -0.0204, 0.0008]
            max drift from home = 1.4581 rad   (joint4 -1.5708 -> -0.1126, ~84 deg)
```

Two user-visible consequences:

- **The arm sags off the pose it was just placed in.** Anyone who spawns a keyframed arm and steps before issuing their own `ctrl` watches it collapse -- the keyframe's whole purpose, a self-holding canonical pose, is silently defeated.
- **`reset()` is the worse half.** An eval loop that resets between episodes starts *every* episode with the setpoints zeroed, so the robot is already falling as the policy takes its **first** observation of the episode. The gripper is part of it: the Panda `home` key opens it (`0.04`), and with `ctrl` dropped it closes to `0.0008` within 200 steps, so a manipulation policy's first frame shows a closed gripper the keyframe asked to be open.

`reset()`'s own comment says the captured keyframe home state is re-applied so "a keyframe spawn is sticky across resets"; only half of it was. And `add_robot`'s docstring already states that its preservation covers "the pose it is in **and the actuator setpoints holding it there**" for robots already in the world -- the robot being *added* at a keyframe got only the pose.

This is the same conclusion the repo already reached for checkpoints: `save_state`'s docstring records that `mjSTATE_FULLPHYSICS` "silently excluded `ctrl`... so the first step after a restore drove toward the pre-restore targets", and it was moved to `mjSTATE_INTEGRATION` for exactly that reason. The keyframe path is the remaining place where a restored pose ships without the setpoints that hold it.

## Change

`_keyframe_home_state` (was `_keyframe_home_qpos`) now reads the keyframe's `ctrl` beside its `qpos` and returns both in one `_KeyframeHome`; `_apply_home_state_to_robot` writes and records both; `_restore_home_poses` re-asserts both after `mj_resetData`.

- Setpoints are matched into the merged scene by the **same name rule** as the pose, so the two halves can never be resolved against different sets of entities, and they are recorded per robot (`SimRobot.home_ctrl`, keyed by namespaced actuator name) so one robot's restore cannot disturb another's.
- A keyframe that declares **no** `ctrl` is left alone. An all-zero `ctrl` row is exactly what MuJoCo also materializes for a `<key>` naming no `ctrl` at all, so the two are indistinguishable in the model and are treated the same: nothing is recorded and the setpoints stay where they were.

Scope: `ctrl` only. Across the 35 registry robots that ship a `<keyframe>` (40 keys), **34 keys declare nonzero `ctrl` and none declares nonzero `act` or `qvel`** -- so `act`/`qvel` restore would be untestable speculation against the shipped models.

## Verification

`tests/simulation/mujoco/test_add_robot_keyframe.py` (the existing behaviour-named home for this contract): **7 failed / 22 passed** on `upstream/main`, **29 passed** after. Pre-fix failures carry the numbers:

```
spawn left ctrl at [0.0, 0.0] while the keyframe pairs [0.6, -1.1] with the pose it just applied
the arm moved 1.0934 rad off its keyed home pose - the keyframe's setpoints are not holding it
reset() restored the pose but left ctrl at [0.0, 0.0]
after reset() the arm moved 1.0934 rad off home - every episode of a reset-between-episodes eval loop starts already falling
```

Six are behavioural; the seventh is the renamed helper's own width-guard test, updated in place.

The **22 that pass on both trees** are the bound on the change, and include the cases it must not touch: a `<key>` declaring no `ctrl`, a `<key>` whose `ctrl` row is all-zero, a spawn with no keyframe at all, and `reset()` on an unkeyed robot that was manually commanded (`mj_resetData`'s zero must stand -- `reset()` must not invent a setpoint nobody asked for). The new fixture is a horizontal, gravity-loaded arm with `kp=200` and joint damping, so "the servos hold the keyed pose" is a sharp assertion rather than a soft-servo coin flip.

Blast radius, run on this branch and on a pristine `upstream/main` worktree with the same interpreter -- `tests/simulation`, `tests/benchmarks/libero` (the other consumer of a canonical home qpos) and the repo-wide docstring/ASCII/args guards:

| | failed | passed |
|---|---|---|
| `upstream/main` | 1 | 11963 |
| this branch | 1 | 11973 |

Identical failure set, and `+10 passed` reconciles exactly with the 10 tests added. The one failure is pre-existing and unrelated (`test_libero_adapter.py::TestInstallActionController::test_apply_sim_action_no_owns_stepping_keeps_outer_loop`, which counts `mj_step` calls through a module-level patch); it fails the same way on pristine `upstream/main` and passes 3/3 in isolation on both trees.

`ruff check`, `ruff format --check` and `mypy strands_robots tests tests_integ` clean (18 pre-existing errors, unchanged). `upstream/main` was merged in after #2377 landed in the same file; re-verified on the merged tree (245 passed across this contract's tests plus the tests of every PR that landed in the interim).

The same script under `upstream/main` and this branch: `add_robot(data_config="panda", keyframe="home")` -> `reset()` -> 600 steps, nothing commanded, 520x460 EGL render.

keyframe ctrl hold

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-keyframe-ctrl-hold/keyframe_ctrl_hold.mp4)

The two panels are **pixel-identical at step 0** (mean abs delta `0.0004`) -- both restore the pose correctly, so what the clip isolates is purely whether the pose *stays*. By the last frame they are `6.715` apart:

| after `reset()` + 600 steps | `upstream/main` | this change |
|---|---|---|
| setpoints set | 0 / 8 | 4 / 8 |
| max drift from home | **1.5012 rad** | **0.0066 rad** |
| gripper (home key = 0.040 m) | 0.000 m, closed | 0.040 m, open |
| max per-frame pixel delta | 2.174 | 0.131 |


yinsong1986 (approved):
## Summary

Restores the `ctrl` half of a MuJoCo `<keyframe>` alongside its `qpos`, at both `add_robot(keyframe=...)` spawn and `reset()`, so a keyed home pose actually holds itself instead of sagging toward the zero configuration on the first step. `_keyframe_home_qpos` becomes `_keyframe_home_state` returning a frozen `_KeyframeHome` (qpos + ctrl keyed by short name); `_apply_home_state_to_robot` writes and records both halves under namespaced names via the same name rule as the pose, and `_restore_home_poses` re-asserts them per robot after `mj_resetData`, so one robot's restore cannot touch another's actuators. This matches `mj_resetDataKeyframe` semantics and the same conclusion the repo already reached for checkpoints (`mjSTATE_INTEGRATION`).

## What's good

- The no-`ctrl` / all-zero-`ctrl` ambiguity is handled the only defensible way (the two are indistinguishable in the compiled model), documented at the decision site, and pinned by dedicated tests for both spellings plus the unkeyed-robot case where `mj_resetData`'s zero must stand.
- Defensive skips are in place for unnamed actuators (index renumbering across recompile) and unresolvable names on restore (`mj_name_to_id < 0`), and `getattr(robot, "home_ctrl", None)` tolerates instances predating the field.
- `SimRobot.home_ctrl` is an additive, defaulted dataclass field on an in-memory object -- no persisted schema, wire format, or public signature changes.
- Changelog fragment per AGENTS.md step 3; new tests are behaviour-named with failure messages carrying the measured numbers; the gravity-loaded horizontal-arm fixture makes the hold assertion sharp. Verified locally: the touched test file passes 29/29 on this head.

cagataycali:
Closing as superseded by #2384, which fixes the same defect as a strict superset. This is a duplicate claim by me, not a difference of opinion, and it is worth stating what it cost so the next one is caught at intake instead of at merge.

The two describe one defect twice. In MuJoCo `ctrl` *is* the actuator command, so "restore the setpoints a keyframe pairs with its pose" and "apply the actuator command a keyframe pairs with its pose" name the same fix. Both measured it on the same robot and the same keyframe -- the built-in `panda` `home` key sagging off its pose because only `key_qpos` was read and every servo was left commanded to the zero configuration -- at 1.4581 rad here and 1.5008 rad there.

They could not both have landed. `git merge-tree` between the two heads reports a real content conflict in three files:

```
CONFLICT (content): strands_robots/simulation/models.py
CONFLICT (content): strands_robots/simulation/mujoco/simulation.py
CONFLICT (content): tests/simulation/mujoco/test_add_robot_keyframe.py
```

So this was not two approvals buying two changes; it was two approvals buying one, with the second guaranteed to need a rebase or a close whichever merged first. Both were already `APPROVED` and `satisfied` under `require_last_push_approval`, which is what made it urgent: either was a merge candidate, and merging them in sequence without composing is the documented way to break `main` on a pair that each read `CLEAN` on its own.

#2384 is kept rather than this one because it is the superset on correctness and not merely on size:

- it applies the keyed `act` as well as `ctrl`, so a stateful actuator's activation is restored and not just a position servo's target;
- it matches actuators **by name under the robot's namespace**, so another robot's setpoints are untouched -- this branch does not, which is a live defect in any multi-robot scene;
- it carries the `reset()` half, `tool_spec.json`, the `docs/simulation/world-building.md` update, and quantifies the blast radius (28 of the 31 built-in robots that ship a `<keyframe>` declare a non-zero `ctrl` in it).

Nothing here is lost: this branch's behaviour is a subset of #2384's, and #2384 pins it in a dedicated regression file rather than by extending `test_add_robot_keyframe.py`. No issue claim needs moving -- neither pull request declared a closing reference.

The intake read that would have caught this is already documented and I did not run it a second time before authoring. Both of these opened inside the same short window, which is the same shape as every pair measured previously: the collision becomes observable only after the first push, so the minute-0 check cannot see it and the re-read before the first push is the one that can.

yinsong1986 (approved):
## Summary

Restores the `ctrl` half of a MuJoCo `<key>` alongside its `qpos` at both `add_robot(keyframe=...)` and `reset()`, so a keyed home pose is held by the servo setpoints its author paired with it instead of sagging toward the zero configuration on the first step. The rename `_keyframe_home_qpos` -> `_keyframe_home_state` returns a frozen private `_KeyframeHome` (qpos + ctrl keyed by short source name); `_apply_home_state_to_robot` writes and records both halves under the same namespace-stripping rule so pose and setpoints can never resolve against different entities; `_restore_home_poses` re-asserts both after `mj_resetData`, scoped to the actuators each robot's own keyframe named so multi-robot restores cannot cross-contaminate. The all-zero-ctrl row is deliberately treated as "not authored" (indistinguishable in the compiled model from a `<key>` naming no ctrl), keeping setpoint-free keyframes byte-identical to today's behaviour.

## What's good

- The only new persisted surface is `SimRobot.home_ctrl`, an additive default-valued dataclass field; `_KeyframeHome` stays private. No wire-format, schema, or public-signature doors opened.
- Locking discipline is unchanged from the methods being replaced, and `_restore_home_poses` documents that the caller holds the model lock, per the #85 thread-safety learnings.
- Bound tests cover exactly the cases the change must not touch: no-ctrl key, all-zero-ctrl key, keyframe-free spawn, and reset() on a manually commanded unkeyed robot (mj_resetData's zero stands).
- Unnamed actuators are skipped with a comment explaining why (name-matching across recompile) rather than guessed at by index -- consistent with the per-name-not-flat-index data-integrity rule.
- Changelog fragment in `changelog.d/`, no `Unreleased` append; no host paths or non-ASCII introduced by the diff.

PR #2384, opened 2026-08-17
fix(simulation/mujoco): apply the actuator command a keyframe pairs with its pose

cagataycali:
## Problem

A MuJoCo `<key>` pairs a pose with the actuator command that *holds* it, and `mj_resetDataKeyframe` -- MuJoCo's own definition of what a keyframe restores -- writes `qpos`, `qvel`, `act`, `ctrl`, the mocap pools and the clock.

`add_robot(keyframe=...)` read `key_qpos` alone. So a gravity-loaded arm stood at its home configuration with every actuator commanded to the **zero** configuration, and the first steps drove it off home. `reset()` reproduced exactly the same state at the top of every episode.

Measured on the built-in `panda` `home` keyframe, whose `key_ctrl` is `[0, 0, 0, -1.5708, 0, 1.5708, -0.7853, 255]`:

```
add_robot(name="p", data_config="panda", keyframe="home")
spawn qpos    [0, 0, 0, -1.5708, 0, 1.5708, -0.7853, 0.04, 0.04]   the canonical Franka home
spawn ctrl    [0, 0, 0,  0,      0, 0,       0,      0]            4 keyed setpoints dropped
step(400)     max joint drift from home = 1.5008 rad               joint4 -1.5708 -> -0.0700

reset()       qpos back at home, ctrl still [0] * 8
step(400)     max joint drift from home = 1.5008 rad               identical sag
```

**28 of the 31 built-in registry robots that ship a `<keyframe>` declare a non-zero `ctrl` in it**, so this is the common case rather than an edge one. The keyframe's whole purpose is a self-holding canonical pose, and a policy trained from the home pose sees the arm already sagging as it takes its first inference of every episode -- the out-of-distribution start the keyframe feature exists to remove.

The rest of the backend already treats a pose without the command holding it as a bug, in both directions, and `docs/simulation/world-building.md` says so for the scene a robot joins:

> an arm keeps the pose it is in (whether that is its keyframe pose or wherever a policy or `send_action` has driven it) *and* the actuator setpoints holding it there

`eject_robot_from_scene` carries `ctrl` across a robot removal for the same reason, in the same words ("the setpoint a position servo holds a pose with. Dropped, it reads as zero ... an arm parked mid-air sags to the floor"), and `_snapshot_scene_state` carries `ctrl`/`act` as one pair per named actuator to do it. The robot *being added* with a keyframe was the one place that got the pose without the command.

## Change

The keyed `ctrl`/`act` are captured beside the pose in `_keyframe_home_state` (was `_keyframe_home_qpos`), recorded on the robot as `home_actuators`, and applied by `_apply_home_state_to_robot` and `_restore_home_state` (was `..._poses`) -- so the pair travels together at spawn and on every `reset()`.

Two properties worth calling out:

**Applied verbatim, never classified by actuator type.** The MJCF author chose those numbers for those actuators, so a servo setpoint, a motor torque and a stateful actuator's activation each carry across as whatever quantity their own actuator reads. That is the same rule `_snapshot_scene_state` follows, and it is why no drive-type predicate appears here.

**Matched by name under the robot's namespace**, so only actuators this robot contributed are written and a robot already in the world keeps its own setpoints. An unnamed actuator cannot be matched in the merged model, so it is skipped with a debug line naming it (its joint still gets the keyed pose); no built-in registry robot has one.

`qvel` is deliberately not applied: the robot-scoped reset that runs immediately before the apply zeroes it so a freshly added robot starts at rest ("callers that want a pre-settled pose call `step()`"), and no built-in keyframe declares a non-zero `qvel`. `time` and the mocap pools are world-scope, with no slice belonging to one robot.

## Result

Swept over all 31 keyframed registry robots: `add_robot(keyframe=0)`, `step(400)`, measuring max drift across the robot's own hinge/slide joints.

| | robots | median drift before | median drift after |
|---|---|---|---|
| base supported (arms, and the humanoids that stay up) | 20 | 1.4179 rad | **0.0077 rad** |
| keyframe declares no `ctrl` (`koch`, `rby1`) | 2 | unchanged | unchanged |
| free base topples in both trees | 8 | unchanged | unchanged |

Worst residual among the 20 is 0.0734 rad (`spot`), which is the servos' steady-state droop against their own finite gain, not a tracking failure. Examples: `kinova_gen3` 3.1416 -> 0.0044, `unitree_g1` 1.2762 -> 0.0028, `panda` 1.5008 -> 0.0066, `aloha` 0.8810 -> 0.0258, `ur5e` 1.5685 -> 0.0079.

Two groups deserve their own note, since both look like regressions in a raw whole-`qpos` metric:

- **The 8 free-base robots** (`apollo`, `toddlerbot_2xc/2xm`, `unitree_go2`, `cassie`, `go1`, and the two drones) topple or settle under gravity in *both* trees -- a keyframe is a pose, not a balance controller. Their articulation numbers move either way, and `go1`'s actually improves (0.9129 -> 0.0669) while its base still settles. Confirmed against MuJoCo directly: `mj_resetDataKeyframe` + `step(400)` on the raw `toddlerbot_2xm` model gives 4.5384 against this branch's 4.5386, i.e. this branch reproduces MuJoCo's own keyframe semantics to 4 decimal places.
- **`stretch3`** genuinely moves (0.0211 -> 0.5191 rad), because its `home` keyframe is internally inconsistent in the asset: it commands `lift` to 0.6 m while keying that joint's pose at 0.0, and commands the arm tendon to 0.1 against keyed arm joints of 0.0. `mj_resetDataKeyframe` drives the lift up exactly the same way. Second-guessing which half of a `<key>` to believe would silently discard the asset author's intent, so the command is applied as written; the behaviour is pinned by a test so it is a decision on the record rather than an accident.

One capture script, run under each tree: spawn `panda` with `keyframe="home"`, then step and render. Nothing is mocked and no source differs between the panels.

panda home pose, held vs dropped

[keyframe_hold.mp4](https://raw.githubusercontent.com/cagataycali/robots/media-keyframe-hold/keyframe_hold.mp4)

Verified rather than eyeballed: the two panels' **first frames are pixel-identical** (mean abs delta `0.0002`, so both start from the same keyed pose), the final frames differ (`4.85`), and the before panel's per-frame motion totals `24.1` against the after panel's `0.9` -- the collapse against the hold.

## Tests

`tests/simulation/mujoco/test_keyframe_restores_the_command_that_holds_it.py`, hermetic (inline MJCFs, gravity on, no mesh download and no render). The fixture arm reaches sideways under gravity with damped `kp=200` servos, plus a `<motor>`-driven joint and a `dyntype="filter"` actuator, so one keyed `ctrl` vector carries three different physical quantities.

Against `upstream/main`: **11 failed, 2 passed**. The headline fails with a measured diff rather than a missing attribute:

```
AssertionError: the second rollout began sagging off home by 1.0870 rad;
reset() dropped the keyed setpoints (ctrl=[0.0, 0.0, 0.0, 0.0])
```

The 2 that pass on both trees are the boundaries this must not cross: a keyframe declaring no `ctrl` leaves every setpoint at zero (MuJoCo fills the omitted vector with zeros, which is what a reset writes anyway, so those robots are byte-for-byte unaffected), and a keyed `qvel` is still not applied. Also pinned: the command reaches all three actuator kinds (with a premise assertion that the fixture really has three, so a servo-only implementation could not pass); a keyframe spawn leaves another robot's setpoints alone; a robot spawned without a keyframe records nothing to restore; an unnamed actuator costs only its own command and not its neighbours'; and the `stretch3`-shaped disagreeing keyframe is applied as written.

`tests/simulation/mujoco/test_add_robot_keyframe.py` caught the error-return arity while this was being written (five refusal paths in the capture helper) and is updated for the renamed helper; it passes unchanged otherwise, including the existing width-guard and multi-robot scoping tests.

## Gate

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` adds no error (19 pre-existing, identical on both trees).

Blast radius run on both trees with the same selection (`tests/simulation`, `tests/registry`, `tests/benchmarks/libero`, and the repo-wide docstring / ASCII / xref / dependency / changelog guards): **this branch 1 failed / 12877 passed**, `upstream/main` **12 failed / 12866 passed**. The 11 extra base failures are this PR's own test file copied in for the pre-fix measurement; the one failure common to both trees is pre-existing (`test_libero_adapter.py::TestInstallActionController::test_apply_sim_action_no_owns_stepping_keeps_outer_loop`) and does not touch this path. The passed delta reconciles exactly: 12877 - 12866 = 11 = 13 new tests minus the 2 that already pass on `main`.


yinsong1986 (approved):
## Summary

`add_robot(keyframe=...)` previously read `key_qpos` alone, so a keyframed spawn stood at its home pose with every actuator commanded to zero and sagged off home on the first steps, identically after every `reset()`. This PR captures the keyed `ctrl`/`act` beside the pose in `_keyframe_home_state`, records it on the robot as `home_actuators`, and applies/restores the pair together in `_apply_home_state_to_robot` and `_restore_home_state` - matching MuJoCo's own `mj_resetDataKeyframe` semantics. Matching is by name under the robot's namespace, so a sibling robot's setpoints are untouched; the command is applied verbatim rather than classified by actuator type, mirroring `_snapshot_scene_state` on the eject path. The activation writes are width-guarded in both directions (skip-with-warning rather than spill into a neighbouring actuator's `act` slice), the unnamed-actuator case degrades to a debug line costing only its own command, and the no-`ctrl` keyframe path is byte-identical to before.

## What's good

- The new hermetic suite fails on `main` with a measured drift rather than an `AttributeError`, and pins the boundaries as well as the fix: three actuator kinds through one `ctrl` vector (with a premise assertion so a servo-only implementation cannot pass), no-`ctrl` keyframe unaffected, keyed `qvel` still not applied, sibling-robot setpoints untouched, unnamed actuator scoped to its own loss, and the `stretch3`-shaped pose/command disagreement applied as written (a decision on the record).
- Changelog fragment, docs, docstrings and the `tool_spec.json` parameter description all updated in the same PR; the private helpers were renamed rather than left describing half of what they now do.
- Locking and error-contract conventions unchanged: the new writes run in the same caller-holds-the-lock contexts as the prior `qpos`-only writes, and every refusal path in the widened three-tuple return still yields an error dict.

## Verification suggestions

The `Merge Base Overlap Check` failure on this head is the stale-base prompt, not a defect: `main` gained `_unknown_action_msg` (#2377), a `tool_spec.json` description edit and #2379/#2380 after this branch diverged. I composed the branch with current `main` locally - the merge is conflict-free (0 combined-diff lines) - and ran the affected suites from both sides (`test_keyframe_restores_the_command_that_holds_it.py`, `test_add_robot_keyframe.py`, `test_scene_grow_preserves_joint_forces.py`, both unknown-message suites, `test_relative_output_path_refusal_is_actionable.py`): 153 passed, 0 failed. Per AGENTS.md step 5, the conflict-free refresh is free of the approval; merging `main` into the branch and pushing will clear the overlap check.

yinsong1986 (approved):
## Summary

`add_robot(keyframe=...)` read `key_qpos` alone, so a keyframed spawn stood at its home pose with every actuator commanded to zero and sagged off home on the first steps, identically after every `reset()`. This PR captures the keyed `ctrl`/`act` beside the pose (`_keyframe_home_qpos` -> `_keyframe_home_state`), records the pair on the robot as `SimRobot.home_actuators`, and applies/restores both together at spawn (`_apply_home_state_to_robot`) and on every reset (`_restore_home_poses` -> `_restore_home_state`). The command is applied verbatim (never classified by actuator type, matching `_snapshot_scene_state`'s eject-path rule) and matched by name under the robot's namespace, so a co-resident robot's setpoints are untouched; keyed `qvel`, `time` and the mocap pools are deliberately out of scope with the reasoning documented at the capture site.

## What's good

- No stale callers of any renamed helper remain in the tree, and the widened error-return arity is applied uniformly across all five refusal paths.
- The apply/restore sites sit inside the same locking context as the pre-existing qpos writes (`add_robot` under the dispatch lock, `_restore_home_state` under `self._lock` in `reset()`), so the `mjData` thread-safety contract from AGENTS.md #85 is preserved.
- Defensive width guards on both the apply and restore paths keep a mismatched activation slice from spilling into a neighbouring actuator, mirroring the existing qpos width guard, and the unnamed-actuator case degrades to a debug line rather than a failed spawn.
- The new test file is hermetic and pins the boundaries as controls: no-`ctrl` keyframe byte-identical, keyed `qvel` not applied, another robot's setpoints untouched, unnamed actuator costs only its own command, and the `stretch3`-shaped disagreeing keyframe applied as written (a decision on the record). All 32 tests in the two touched files pass on the head commit.
- Changelog fragment, docs, and the `tool_spec.json` `keyframe` description all updated in the same PR; no parameter names or public signatures change, so there is no one-way door here.

PR #2385, opened 2026-08-17
fix(policies): make an explicit CompositePolicy joint group exclusive on every tick

cagataycali:
## Problem

`CompositePolicy` documents `upper_joints` as "joint/actuator names the upper policy is
authoritative for". Nothing subtracted that group from the lower policy's route, so a
defaulted `lower_joints` kept **every** name the lower policy emits, including names the
caller had explicitly assigned to the upper policy.

Which value reached the actuator was then decided per tick by the upper policy's chunk
contents:

```python
c = CompositePolicy(whole_body, arms, upper_joints=["armA", "armB"])

# tick where the upper policy also emits armB
c._merge_tick({"leg": 1.0, "armB": 9.9}, {"armA": 0.5, "armB": 0.5})
# -> ValueError: both produced joint(s) ['armB']

# tick where the upper policy is silent on its own joint
c._merge_tick({"leg": 1.0, "armB": 9.9}, {"armA": 0.5})
# -> {'leg': 1.0, 'armB': 9.9, 'armA': 0.5}
#                  ^^^^^^^^^^ armB is the upper policy's by declaration, driven by the
#                             lower policy. status success, no warning, no log line.
```

One configuration, two outcomes in different **kinds** - refuse, or write a wrong actuator
target - selected by whether the upper policy's chunk happened to cover that joint. A
chunk-emitting manipulation policy is silent on some of its own joints on most ticks, so a
single rollout can produce both.

## The measurement

An exhaustive sweep of the four group configurations, over every non-empty lower/upper
emission pair on a 3-name joint set (49 pairs each, 196 total):

| configuration | conflict refused | silently drove an upper-owned joint | outcomes changed by this PR |
| --- | --- | --- | --- |
| both groups explicit | 0 | 0 | 0 of 49 |
| `lower_joints` only | 0 | 0 | 0 of 49 |
| **`upper_joints` only** | **35** | **8** | **8 of 49** |
| both defaulted | 0 | 0 | 0 of 49 |

`upper_joints`-only is the only configuration that can silently override an explicit
assignment, and it is also the only configuration from which the documented
ownership-conflict refusal was reachable at all. With either group defaulted the upper dict
is routed with the lower's names excluded; with both explicit the groups are disjoint at
construction. So the class's headline guarantee ("a genuine ownership conflict is raised,
never silently resolved") lived entirely inside the one configuration whose ownership was
ill-defined, where it fired on some ticks and not others.

All 8 changed outcomes are exactly the 8 that were silently wrong. No previously-correct
merge changes.

## Change

Refuse on the condition that is actually ambiguous - the lower policy commanding into an
explicit upper group - whether or not the upper policy emits that name this tick. The new
check subsumes the collision check it replaces (a contested name was only ever reachable
that way), so this is one refusal rather than two, and the answer is the same on every tick.

```
ValueError: CompositePolicy: upper_joints assigns joint(s) ['armB'] to the upper policy
'manipulation_upper', but the lower policy 'whole_body_lower' also commanded them this
tick. An explicit joint group is exclusive - the policy that owns a name is the only one
allowed to command it, whether or not the other child emits that name on a given tick. Set
lower_joints so the lower policy does not command into the upper group (e.g.
lower_joints=WBC_G1_LEG_WAIST_JOINTS for a G1 locomotion policy), or remove those names
from upper_joints.
```

### Why refuse rather than drop the lower policy's value

Silently filtering the lower policy's value out would leave that joint on its previous
command with nothing to say so - a dropped command, which is what this class raises on
everywhere else ("a dropped command on a humanoid is a fall, not a warning";
`_reject_discarded_child` exists for the same reason). Trading a wrong value for a missing
one keeps the failure silent. Refusing does not.

## Scope: what does not change

- **Precedence between two defaulted groups is untouched.** With no group declared there is
  no owner to honour, so the documented lower-precedence rule still arbitrates. Only an
  explicit assignment is enforced.
- **`lower_joints`-only is untouched.** The mirror configuration never had the hole.
- **The shipped locomotion pairing is unaffected.** `WBCPolicy` emits its 15 leg+waist
  joints and none of the 14 arm joints, so nothing is contested even with `lower_joints`
  left default - pinned as a test.
- **No shipped usage changes.** `examples/wbc/wbc_g1_composite.py`, `docs/policies/wbc.md`
  and the class docstring example all set both groups.

The class docstring, the `Args` entries and `get_actions`'s `Raises` are updated to state
that an explicit group is exclusive, since "authoritative for" and "accepts every name the
lower policy emits" disagreed in exactly this configuration.

## Tests

`tests/policies/test_composite_explicit_group_ownership.py`, **3 failed / 5 passed on
`main`, 8 passed after**. The headline test compares two ticks of one config and fails
pre-fix with the defect stated outright:

```
AssertionError: the same composite answered 'merged' when the upper policy was silent on
its own joint 'armB' and 'refused' when it emitted it; the silent tick returned
{'leg': 1.0, 'armB': 9.9, 'armA': 0.5}
```

The 5 that pass on both trees are the boundaries: defaulted-group precedence still applies,
`lower_joints`-only is unchanged, correct use of this very config still composes, a
whole-body lower that overlaps on every tick is still refused, and the WBC leg/arm split is
untouched.

`test_runtime_collision_raises` in `tests/policies/test_composite.py` is renamed to
`test_lower_commanding_an_upper_owned_joint_raises` and matches the new wording. Its intent
is unchanged - it reaches the ownership branch through the only configuration that reaches
it - and its comment now says so.

## Verification

- `tests/policies tests/registry` + the repo-wide docstring/ASCII/dependency guards: 5253
  passed, 3 skipped.
- `ruff check` / `ruff format --check` / `mypy` clean over `strands_robots tests tests_integ`.

One capture script, one scene, one camera; only the tree on `PYTHONPATH` differs. The
manipulation policy owns both arms and its chunk covers only the left arm this tick, while
the locomotion policy's action space also reaches the right arm.

CompositePolicy explicit group ownership

Left: `main` reports success for 29 joints and drives `right_shoulder_roll_joint` to
`-1.52`, the locomotion policy's value, on a joint the manipulation policy owns. Middle:
refused, nothing written. Right: with the remedy applied the right arm holds because its
declared owner chose not to command it - and that render is pixel-identical to `main`'s for
the same config (2.7e-06 mean absolute pixel difference), so the fix is a no-op once
ownership is unambiguous.

### DIFF
--- a/changelog.d/2385-composite-explicit-group-ownership.md
+++ b/changelog.d/2385-composite-explicit-group-ownership.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- `CompositePolicy` now treats an explicit `upper_joints` group as exclusive on every tick.
+  A defaulted `lower_joints` kept every name the lower policy emits, so a lower policy whose
+  action space reached the upper group also commanded it: when the upper policy emitted that
+  name too the merge raised, and when it was silent on it the lower policy's value was
+  written to a joint the caller had assigned away, reported as success. The lower policy
+  commanding into an explicit upper group is now refused whether or not the upper policy
+  emits that name, naming the contested joints, both policies and the remedy. Precedence
+  between two defaulted groups is unchanged, and `WBCPolicy` emits no arm joints so the
+  shipped locomotion pairing is unaffected.
--- a/strands_robots/policies/composite.py
+++ b/strands_robots/policies/composite.py
@@ -67,6 +67,12 @@ class CompositePolicy(Policy):
       every name it emits that the lower policy did not already claim - lower
       precedence).
 
+    An explicit group is EXCLUSIVE: the policy it names is the only one allowed to
+    command those joints. A defaulted ``lower_joints`` therefore still may not
+    command into an explicit ``upper_joints`` - that tick is refused, on every tick
+    alike, rather than resolved by whichever child emitted the name. Precedence
+    only decides between two DEFAULTED groups, where the caller declared no owner.
+
     Routing that discards a child's ENTIRE action dict raises: the composite
     would otherwise silently be the surviving child alone. Two children that
     drive the same joint set cannot be composed, only cascaded (module docstring).
@@ -81,10 +87,13 @@ class CompositePolicy(Policy):
         lower: Policy driving the lower joint group (e.g. legs+waist locomotion).
         upper: Policy driving the upper joint group (e.g. arms manipulation).
         lower_joints: Joint/actuator names the lower policy is authoritative for.
-            ``None`` (default) accepts every name the lower policy emits.
+            ``None`` (default) accepts every name the lower policy emits, minus
+            any that ``upper_joints`` assigns to the upper policy - commanding
+            into an explicit upper group is refused, not silently resolved.
         upper_joints: Joint/actuator names the upper policy is authoritative for.
-            ``None`` (default) accepts every upper name not already owned by the
-            lower policy.
+            Exclusive: no other child may command these, whichever names the upper
+            policy emits on a given tick. ``None`` (default) accepts every upper
+            name not already owned by the lower policy.
         lower_obs_keys: Observation keys to forward to the lower policy. ``None``
             (default) forwards the full observation (children read by name).
         upper_obs_keys: Observation keys to forward to the upper policy. ``None``
@@ -208,8 +217,8 @@ class CompositePolicy(Policy):
         Raises:
             ValueError: If either child returns an empty chunk, if routing
                 discards a child's entire action dict (the composite would be
-                the other child alone), or if the routed lower and upper action
-                dicts both assign the same joint name.
+                the other child alone), or if the lower policy commands a joint
+                ``upper_joints`` assigns to the upper policy.
         """
         lower_obs = self._filter_obs(observation_dict, self._lower_obs_keys)
         upper_obs = self._filter_obs(observation_dict, self._upper_obs_keys)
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

This PR closes a tick-dependent ownership hole in `CompositePolicy`: with an explicit `upper_joints` and a defaulted `lower_joints`, a lower policy whose action space reached the upper group was refused on ticks where the upper policy also emitted the contested name, and silently drove the upper-owned joint on ticks where it did not. The fix replaces the per-tick collision check with `_reject_contested_upper_ownership`, which refuses whenever the routed lower dict intersects an explicit `upper_joints` - and that check strictly subsumes the one it removes (with an explicit upper group, `up` is a subset of `upper_joints`, so any `lo & up` collision was already `lo & upper_joints`; with the group defaulted, `up` is routed with `exclude=set(lo)` so a collision was never reachable). I verified the PR's headline claims independently: the new test file is 3 failed / 5 passed on the merge base and 8 passed on the head, the renamed `test_lower_commanding_an_upper_owned_joint_raises` still reaches the ownership branch through the only configuration that reaches it, and the full `tests/policies` suite passes on the head with no regression relative to base. The previously-silent misbehaviour now fails loudly with a `ValueError` naming the contested joints, both providers and the remedy, consistent with the class's no-silent-drop contract; defaulted-group precedence and the shipped WBC pairing are pinned unchanged.

## What's good

- The refusal names the contested joints, both policies by `provider_name`, and a concrete remedy - actionable without reading the merge implementation, per the no-silent-defaults rule.
- Refusing rather than filtering the lower value is the right call: dropping would leave the joint on its previous command silently, the exact failure `_reject_discarded_child` exists to prevent.
- Changelog fragment under `changelog.d/` (not `CHANGELOG.md`), docstrings updated where "authoritative for" and "accepts every name" disagreed, boundary tests pin the three configurations that must not change, all strings ASCII.

Issue #318, opened 2026-06-04
MuJoCo send_action cannot drive tendon-transmission actuators (Panda/Franka gripper never actuates)


cagataycali:
## Summary

`MuJoCoSimEngine.send_action` (via `_apply_action_by_name`) cannot drive **tendon-transmission actuators** — most notably the **Franka/Panda gripper**. A gripper command keyed by the finger joint name (`finger_joint1`) is silently dropped, so the gripper never actuates during a policy rollout. The 7 arm joints (direct joint-transmission actuators) work correctly.

Surfaced while running a 3000-step VLA policy rollout (`Cosmos3Policy`, PR #317) on the `panda` sim robot and recording a LeRobotDataset: the arm reaches/resets cleanly across all episodes, but the gripper fingers (`finger_joint1`/`finger_joint2`) stay at `0.000` the entire time even though the policy emits a varying gripper command.

## Root cause

The default `panda` asset exposes **8 actuators**: 7 joint-transmission (arm) + 1 **tendon-transmission** (gripper):

```
arm/actuator1: trntype=JOINT  target=arm/joint1  ctrlrange=[-2.8973, 2.8973]
...
arm/actuator7: trntype=JOINT  target=arm/joint7  ctrlrange=[-2.8973, 2.8973]
arm/actuator8: trntype=TENDON target=arm/split   ctrlrange=[0.0, 255.0]   <-- gripper
```

`_apply_action_by_name` (`strands_robots/simulation/mujoco/rendering.py`) resolves each `action_dict` key as:

1. actuator name → write `data.ctrl[act_id]`, else
2. **joint** name → find the actuator whose `actuator_trnid[ai, 0] == jnt_id` and write its ctrl.

```python
for key, value in action_dict.items():
    act_id = _lookup(mj.mjtObj.mjOBJ_ACTUATOR, key)
    if act_id >= 0:
        data.ctrl[act_id] = float(value)
    else:
        jnt_id = _lookup(mj.mjtObj.mjOBJ_JOINT, key)   # e.g. finger_joint1 -> joint 7
        if jnt_id >= 0:
            for ai in range(model.nu):
                if model.actuator_trnid[ai, 0] == jnt_id:  # <-- only matches JOINT-transmission
                    data.ctrl[ai] = float(value)
                    break
```

[...]

PR #2386, opened 2026-08-17
fix(simulation/mujoco): refuse to double-actuate a joint a tendon already drives

cagataycali:
## Problem

`actuate_robot`'s guard states its own job in the refusal it raises: *"refusing to double-actuate"*. It decided which joints were already driven through `actuator_joint_id`, which reports `-1` for a tendon transmission - correctly, because a tendon is not a joint.

A gripper whose fingers are coupled by one fixed tendon is the standard MJCF idiom, and `actuator_joint_id`'s own docstring names the case: *"A fixed tendon coupling a gripper's fingers ... (the Menagerie Panda hand and the Robotiq 2F-85 both use one), so the collision is reachable with stock assets"*. On a robot driven **only** that way, the guard saw nothing driven, reported `success`, and added a position servo per finger on top of the drive that was already there.

Measured on `robotiq_2f85`, commanding the shipped gripper interface fully shut and stepping 600 times:

| | `actuate_robot` | `nu` | `left_driver_joint` travel on `fingers_actuator=255` |
|---|---|---|---|
| before | `success`, 8 servos added | 1 -> 9 | **0.0164 rad** |
| after | `error` | 1 | **0.7838 rad** |

The two drives fight, so the gripper keeps 2% of its travel - and the call that broke it reported success. `robotiq_2f85` and `robotiq_2f85_v4` are the two registry robots driven only through a tendon.

The rule the guard needed already exists in this package, on the action-application path: `_actuator_for_joint` resolves a tendon to the joints it wraps, added for issue #318 because *"Tendon grippers ... were silently dropped before this branch"*. So the same blindness had already been found and fixed once, one module away, and the guard kept the narrower reading.

## Change

`scene_ops` gains the shared rule both directions read:

- **`actuator_driven_joint_ids(model, act_id, mj)`** - every joint an actuator drives: its transmission target for `JOINT`/`JOINTINPARENT`, every wrapped joint for `TENDON`, nothing for a site, body or slider-crank, which move a frame rather than commanding a joint coordinate.
- **`tendon_joint_ids(model, tendon_id, mj)`** - the wrap-list walk, which both directions need. `_actuator_for_joint` now calls it instead of carrying its own copy.

`actuator_joint_id` is deliberately left alone. It answers the narrower question - *which single joint is this actuator's transmission target* - which is what the hold-pose seeding needs, since it has one `ctrl` slot to fill from one joint's position. The two questions have different answers for the same actuator, so they are two functions rather than a flag on one; collapsing them would make one of the two call sites wrong.

The refusal now names what it found, and distinguishes the two cases, because what a caller does next differs between them:

```
actuate_robot: 2 of the 8 hinge/slide joint(s) on robot 'robotiq_2f85' are already driven by an
existing actuator (left_driver_joint <- fingers_actuator, right_driver_joint <- fingers_actuator);
refusing to double-actuate, because a second drive on the same joint fights the first. The
remaining 6 are positioned by the mechanism rather than by a ctrl of their own - a mimic joint or
a closed linkage member - so a position servo there would fight the constraint that already holds
them.
```

```
actuate_robot: 6 of the 6 hinge/slide joint(s) on robot 'wide' are already driven by an existing
actuator (w0 <- w0_act, w1 <- w1_act, w2 <- w2_act, w3 <- w3_act and 2 more); refusing to
double-actuate, because a second drive on the same joint fights the first. The robot is already
drivable through those actuators, so it needs no added ones.
```

The sibling refusals in the same method already answer at that resolution - the unknown-`kp` one lists the valid joints - so this was the one verdict there that named neither what it found nor what it left.

## Blast radius

Every loadable registry robot, `actuate_robot(kp=100.0)` on each:

| | count |
|---|---|
| verdict changed | **2** - `robotiq_2f85`, `robotiq_2f85_v4`, both `success` -> `error` |
| verdict unchanged | 60 (58 refused, 2 actuated) |
| refusal now names the driven joints | 56 |

The 2 that change are exactly the robots whose gripper the old `success` broke. The 2 that still actuate are the genuinely actuator-less ones, so the capability itself is untouched.

Only tendon transmissions were added to what the guard can see. A site or body transmission still does not refuse, and a tendon wrapping a **different** robot's joints still does not - the id-space collision the narrower gate was introduced for. `test_actuate_robot_is_not_refused_by_a_tendon_id_collision` covers that second case and passes unchanged, because the fix resolves the tendon to its real joints rather than comparing the raw id.

One capture script, run under each tree: add `robotiq_2f85`, call `actuate_robot`, then command the pre-existing `fingers_actuator` shut. Nothing is mocked - registry robot, shipped `actuate_robot`, shipped `send_action`, real physics.

robotiq_2f85 gripper travel, before and after

[tendon_guard.mp4](https://raw.githubusercontent.com/cagataycali/robots/media-actuate-robot-tendon-guard/tendon_guard.mp4)

Verified rather than eyeballed: the two panels' final frames differ (mean abs delta `7.02`, 15.9% of pixels), and the after panel's per-frame motion totals `18.38` against the before panel's `1.29` - the servos holding the fingers shut against the gripper actually closing.

## Tests

`tests/simulation/mujoco/test_actuate_robot_sees_a_tendon_drive.py`, hermetic - inline MJCF through `add_robot(urdf_path=...)`, nothing downloaded.

Against `upstream/main`: **7 failed, 6 passed**. The behaviour failures carry real numbers rather than a missing keyword:

```
test_a_robot_driven_only_through_a_tendon_is_refused
  assert 'success' == 'error'
  Robot 'grip' actuated: 2 position actuator(s) added (['finger1', 'finger2']), nu=3, ...

test_the_existing_tendon_drive_keeps_the_travel_it_had
  the tendon drive moved finger1 0.0139 rad before actuate_robot and 0.0041 rad after it
```

The travel test uses its own dynamics-grade fixture (explicit masses, real damping, a ctrlrange in tendon length units) and takes the same gripper's own pre-call travel as the reference, so the assertion carries no hand-written number. The sibling suite's `GRIPPER_XML` carries near-zero mass and drives straight through its joint limits, which is fine for the id assertions it was written for but not for reading travel off.

The 6 that pass on both trees are the ones that fail if the rule is widened too far: a robot carrying no actuator still actuates, a site transmission still does not block, `actuator_joint_id` still reports no joint for a tendon, a joint's own drive keeps priority over a tendon it shares with a neighbour, and `tendon_joint_ids` answers empty for an out-of-range id rather than reading past the wrap table.

`tests/simulation/mujoco/test_actuate_robot.py::test_double_actuation_refused` asserted the old phrase. It now asserts the joints and the actuators driving them, which is the part of the verdict a caller acts on - and it fails on `upstream/main`, so the rewrite strengthened it rather than relaxing it to fit.

## Verification

`tests/simulation tests/registry tests/policies tests/tools` plus the repo-wide docstring, unicode, ASCII, changelog, xref and dependency guards: **18392 passed, 189 skipped, 0 failed**, re-run after merging `main` (which landed #2380 into the same `scene_ops.py`; both sides' additions verified present by symbol, and `uv lock --check` clean). `ruff check` / `ruff format --check` and `mypy` over `strands_robots tests tests_integ` clean (19 pre-existing errors, identical to `upstream/main`'s own count, none in the touched files).


yinsong1986 (approved):
## Summary

Fixes a real physics-corruption bug: `actuate_robot`'s double-actuation guard resolved each existing actuator's transmission through `actuator_joint_id`, which correctly reports no joint for a tendon transmission, so a gripper driven only through a fixed tendon (`robotiq_2f85`, `robotiq_2f85_v4`) read as undriven - the call reported `success`, stacked a position servo per finger on top of the existing tendon drive, and the two drives fought (98% travel loss on the shipped gripper interface). The fix adds two shared helpers to `scene_ops` (`actuator_driven_joint_ids`, `tendon_joint_ids`), points the guard at the wider question, and deduplicates the wrap-list walk out of `_actuator_for_joint` so the two directions cannot disagree. `actuator_joint_id` is deliberately unchanged for the hold-pose seeding path, which needs the narrower answer.

## What's good

- The `success` -> `error` flip on the two affected robots is an explicit, well-worded refusal naming the driven joints and their actuators - not a silent behavior change, per the no-silent-defaults rule.
- Guard runs inside `self._lock` and keeps the error-dict contract; `tendon_joint_ids` bounds-checks the tendon id rather than reading past the wrap table.
- Hermetic regression suite (inline MJCF, nothing downloaded) pins both the fix and the boundaries a wider rule must not cross: the tendon id-space collision case, site transmissions, actuator-less robots, and direct-drive priority over a shared tendon. Verified locally at the head SHA: 32/32 pass across the new suite and the existing `test_actuate_robot.py`.
- Changelog fragment, docstring-completeness pins, and ASCII-only user-facing strings all in order.

PR #2387, opened 2026-08-17
fix(simulation/mujoco): export a scene whose asset references still resolve

cagataycali:
## Problem

`export_xml` is the sink for "save / hand off / reload this scene", and `describe()` calls it
"the read sibling of `replace_scene_mjcf`". The file it wrote could not be recompiled -- not by
`mujoco.MjModel.from_xml_path`, and not by strands' own `load_scene`:

```python
sim.create_world()
sim.add_robot(name="a", data_config="so101")
sim.add_robot(name="b", data_config="so100")
sim.add_object(name="crate", shape="box", position=[0.22, 0.26, 0.04], size=[0.05, 0.05, 0.04])

sim.export_xml(output_path="/tmp/handoff/scene.xml")
# status=success  "Model exported to /tmp/handoff/scene.xml"

other.load_scene(scene_path="/tmp/handoff/scene.xml")
# status=error  Failed to load scene: Error: Error opening file 'sts3215_03a_v1.stl'
```

The export named all 31 mesh assets by bare filename with no `meshdir`, so MuJoCo resolved them
against the directory the XML landed in. Nothing warned: the export reported success and the
scene it described was already unloadable.

## Root cause

MuJoCo resolves a `<mesh>` / `<texture>` / `<hfield>` `file=` against a base directory it keeps on
the spec (`modelfiledir`, plus `meshdir` / `texturedir`) and does **not** emit in `spec.to_xml()`.
`spec.attach()` does not carry that base onto the parent either. So the live spec kept enough
internal state to compile while serialising to text that had lost it:

| | result |
|---|---|
| `spec.compile()` on the live spec | OK, nbody=16 nmesh=31 |
| `spec.to_xml()` then `MjSpec.from_string(...).compile()` | `Error opening file 'motor_holder_so101_wrist_v1.stl'` |

The export was strictly less capable than the spec it came from. Two things follow:

- A multi-robot scene has no single `meshdir` that could paper over it. so101 resolves under
  `.../assets/robotstudio_so101/` and so100 under `.../assets/trs_so_arm100/`; each model
  contributes assets from its own root.
- `spec.to_zip()`, MuJoCo's own portable-bundle writer, is broken by the same state: it reads
  `spec.assets`, which is empty here, so it wrote a 1-entry zip containing only the XML and
  `from_zip` failed identically. Delegating to it would not have helped.

Three separate entry points produced the same dead artifact: multi-robot `add_robot`, a
texture-bearing model, and `load_scene` (which keeps the model's `meshdir` but not the directory
it is relative to).

## Change

Absolutize each file-backed asset reference at the one point where a spec is loaded from a file,
while it still knows its own base directory -- `SpecBuilder.from_file`. `attach_robot` now goes
through that helper instead of calling `mujoco.MjSpec.from_file` directly, so the repair cannot be
bypassed and every `to_xml()` consumer inherits it. No new bookkeeping is needed: the information
is already on the child spec and is only lost afterwards.

Meshes and height fields resolve against `meshdir`, textures against `texturedir`. MuJoCo honours
an absolute `file=` regardless of any `meshdir` still declared, so those attributes are left
untouched.

A reference that resolves to nothing is left exactly as authored. MuJoCo's own "Error opening
file" names what the model declares, which is more useful than an invented absolute path that was
never on disk.

Scope: this makes the export reload from any directory on the machine that produced it, which is
what the "read sibling" contract needs. It is deliberately not an asset-bundling feature -- the
new docs section states plainly that the XML alone does not carry assets to another machine.

## Verification

Structural parity on the reported scene, `nbody/njnt/nu/nq/nv/ngeom` identical live vs reloaded:

| Scene | live | reloaded before | reloaded after |
|---|---|---|---|
| so101 + so100 + crate | `(16,13,12,19,18,63)` | compile failed | `(16,13,12,19,18,63)` |
| anymal_c (file-backed texture) | `(14,13,12,19,18,95)` | compile failed | `(14,13,12,19,18,95)` |
| `load_scene(so101)` then export | `(8,6,6,6,6,30)` | compile failed | `(8,6,6,6,6,30)` |

Absolute asset refs in the export went 0/31 to 31/31.

New tests, `tests/simulation/mujoco/test_export_xml_asset_refs_reload.py`, build their models on
disk (a real tetrahedron STL and a real PNG under a `meshdir` / `texturedir`) so a reference
genuinely needs resolving and nothing is downloaded. Against `main`: **4 failed, 3 passed**, each
failure naming the asset MuJoCo could not open. After: 7 passed.

The 3 that pass on both branches are the controls -- a primitives-only scene still exports and
reloads unchanged, an unresolvable reference is left as authored, and an already-absolute
reference is preserved. The second is what fails if the helper ever starts inventing paths.

Gate: `ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` 19
pre-existing errors on this branch and on `main` (zero new); full `tests/` compared against a
`main` worktree with no new failures.

## Artifact

Scene exported into a directory holding no assets, then reloaded through `load_scene` and
rendered in MuJoCo (headless EGL).

export_xml reload

Before, there is nothing to render: `load_scene` refused the file `export_xml` had just reported
as a success. After, the reloaded scene renders, and its structure matches the live model while
the two renders differ by mean |d| = 0.003. The live scene is identical on both branches
(mean |d| = 0.0001) -- the change affects only how the export names its assets.

Fixes the `export_xml` round-trip reported for the MjSpec backend.

### DIFF
--- a/changelog.d/2387-export-xml-portable-asset-refs.md
+++ b/changelog.d/2387-export-xml-portable-asset-refs.md
@@ -1,0 +1,10 @@
+### Fixed
+
+`export_xml` now writes a scene that can actually be reloaded. MuJoCo resolves a `<mesh>` /
+`<texture>` / `<hfield>` `file=` against a base directory it keeps on the spec and does not emit in
+`spec.to_xml()`, and `spec.attach()` does not carry that base onto the parent - so an exported scene
+named its assets relative to wherever the XML happened to be written, and neither
+`mujoco.MjModel.from_xml_path` nor `load_scene` could recompile it. A multi-robot scene had no single
+`meshdir` that could cover it, because each model contributes assets from its own root. Asset
+references are now absolutized while the loaded spec still knows its own base directory, so the
+exported MJCF is self-describing; a reference that resolves to nothing is left as authored.
--- a/docs/simulation/overview.md
+++ b/docs/simulation/overview.md
@@ -20,7 +20,7 @@ For walkthroughs see [Simulation overview](../simulation/overview.md).
 | `reset` | - | State to t=0, keep model |
 | `get_state` | - | Sim time, joint positions, object poses |
 | `destroy` | - | Tear down model, data, executor |
-| `export_xml` | - | Serialise model to MJCF string |
+| `export_xml` | `output_path` | Serialise live scene to MJCF; reloadable via `load_scene` (assets referenced by absolute path) |
 
 ## Scene-MJCF
 
--- a/docs/simulation/world-building.md
+++ b/docs/simulation/world-building.md
@@ -507,6 +507,30 @@ pre-patch state, so a bad key or a non-finite component never leaves a
 half-applied scene. Use
 `replace_scene_mjcf(xml)` for MJCF elements this vocabulary does not cover.
 
+## Exporting a scene
+
+`export_xml(output_path=...)` serialises the live scene - including every runtime
+mutation - as MJCF. It is the read sibling of `replace_scene_mjcf`, so the file it
+writes is meant to be reloadable:
+
+```python
+sim.export_xml(output_path="/tmp/handoff.xml")
+other.load_scene(scene_path="/tmp/handoff.xml")   # same scene, same structure
+```
+
+Mesh, texture and height-field assets are referenced by ABSOLUTE path. MuJoCo
+resolves a relative `file=` against the model's own directory (plus `meshdir` /
+`texturedir`), and that directory is not part of the serialised XML - so a
+relative reference would resolve against wherever the export happened to be
+written. Absolute references keep the export reloadable from any location, and a
+scene composed from several models needs them: each model contributes assets from
+its own root, so no single `meshdir` could cover them all.
+
+The consequence is that an export names paths on the machine that produced it.
+Copying the XML alone to another machine will not carry the assets with it;
+copy the referenced asset trees too, or re-compose the scene there from the same
+`add_robot` calls.
+
 ## Cameras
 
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Fixes the `export_xml` round-trip for the MjSpec backend: MuJoCo resolves relative `<mesh>` / `<texture>` / `<hfield>` `file=` references against a base directory (`modelfiledir` + `meshdir` / `texturedir`) that `spec.to_xml()` does not emit and `spec.attach()` does not propagate, so an exported multi-robot scene named 31 assets by bare filename and could not be recompiled by `MjModel.from_xml_path` or `load_scene`. The PR absolutizes each file-backed asset reference at the single point where a spec is loaded from a file (`SpecBuilder.from_file`, which `attach_robot` now also routes through, closing the bypass), while the spec still knows its own base directory.

## What's good

- The repair is conservative in exactly the right places: a reference is rewritten only when it is relative **and** resolves to an existing file, so an unresolvable reference stays as authored and MuJoCo's own error keeps naming what the model declares - and both behaviours are pinned by control tests (`test_an_unresolvable_reference_is_left_as_authored`, `test_a_reference_already_absolute_is_preserved`).
- Tests build their fixture models on disk under `tmp_path` (real binary STL, real PNG under a `texturedir`), so the references genuinely need resolving, nothing is downloaded, and no host paths land in the tree (AGENTS.md > Review Learnings (#85) > Testing).
- The round-trip test shape matches the #85 learning: export -> reload through both `MjModel.from_xml_path` and strands' own `load_scene` -> assert structural parity, with a multi-root scene proving no single `meshdir` could substitute.
- `_absolutize_asset_paths` stays private and out of `__all__`; the mutation happens on a fresh spec before it is shared, so no thread-safety contract is touched.
- Changelog fragment under `changelog.d/`, docs updated in the same PR, and the docs state the one real limitation plainly (absolute paths are machine-local; the XML alone does not carry assets to another machine).

Issue #708, opened 2026-06-26
dataset_recorder: silent episode-boundary bug — 8 E2E runs collapsed 20 episodes → 1 (95% data loss)


cagataycali:
## Summary

`dataset_recorder` silently collapses multiple episodes into a single LeRobot episode when `next_episode()` is not explicitly called between rollouts. Across **8 consecutive E2E runs** (run-700 through run-1500), each declaring 20 episodes via `--episodes 20`, the on-disk `meta/info.json` shows **`total_episodes: 1`** — a 95% data loss event going undetected.

## Reproduction

```bash
# E2E harness reports 20/20 episodes via agent self-attestation:
python scripts/e2e_agent_test.py --policy molmoact2 --episodes 20 --output runs/20260626-150038
# → agent prints "Episode 1/20 ✓ ... Episode 20/20 ✓" + summary.txt "status=OK"

# But canonical SoT contradicts it:
cat runs/20260626-150038/dataset/meta/info.json
# → "total_episodes": 1, "total_frames": 1140
```

20 logical episodes × ~57 frames each were concatenated into a single LeRobot episode. No exception, no warning, summary.txt status=OK.

## Root Cause

`strands_robots.dataset_recorder` requires the caller to invoke `next_episode()` (or `start_episode()` / equivalent) between rollouts. When the e2e harness calls `env.reset()` without bumping the recorder's episode index, frames continue appending to episode 0. The recorder has no awareness of env resets and no guard against this footgun.

## Impact

- **Silent data corruption** at the dataset layer — downstream training jobs see 1 long episode instead of N short ones, breaking episode-keyed splits, returns-to-go, and any TD-learning signal that respects episode boundaries.
- **Trust collapse** between agent self-reports and ground truth. Agent says "20/20 ✅"; `meta/info.json` says 1.
- **8-run streak invalidated** (run-700…run-1500). Only run-1500 SOT-audited; earlier 7 runs assumed clean.

### P0 — `strands_robots.dataset_recorder` (this repo)
1. **Auto-bump episode on `env.reset()`** via a hook or explicit `recorder.observe_reset()` API.
2. **Fail-loud if `next_episode()` is missed** when a reset is detected mid-recording (raise `RecorderEpisodeBoundaryError`).
3. Add a `recorder.assert_episode_count(expected: int)` helper for end-of-run validation.

### P0 — `e2e_agent_test.py` (caller side, fix in same PR)
[...]

PR #2388, opened 2026-08-17
fix(dataset_recorder): un-count frames a discarded episode never wrote

cagataycali:
## Problem

`DatasetRecorder.add_frame` counts a frame the moment it is **buffered**, but frames only reach disk when `save_episode` flushes the buffer. `clear_episode_buffer` discards a buffered (aborted) episode and deliberately left the cumulative counter untouched:

```python
# Reset the per-episode frame counter regardless: the next episode
# reports frames from 0. frame_count (cumulative) is left untouched
# since those frames were really written to disk only on save_episode.
self.episode_frame_count = 0
```

The stated reason is inverted. *Because* frames reach disk only on `save_episode`, the frames just discarded **never became parquet rows**, so leaving them counted over-reports by exactly the aborted episode's length — permanently, and cumulatively across aborts.

Two surfaces define `frame_count` as the on-disk total, which is what makes this a correctness contract rather than a cosmetic one:

- **`resume()` seeds `frame_count` from `dataset.meta.total_frames`** — straight from disk truth. So a resumed recorder starts consistent and re-drifts on its first abort.
- **`RecordingMixin.stop_recording` refuses to finalize an empty dataset by asking `frame_count` whether anything was ever captured.** An inflated count blinds that refusal.

That second one is the reason this is more than a reporting bug. `stop_recording`'s own comment describes the regression it was added to prevent:

> Previously stop_recording reported success with "0 frames, 0 episode(s)", silently producing a dataset with only meta/info.json (no parquet/video).

Driving the documented flow on `upstream/main` reproduces exactly that, with a number attached:

```
start_recording -> run_policy(n_steps=12) -> discard the partial rollout -> stop_recording

stop_recording -> status: "success"
  "Episode saved to LeRobotDataset
   local/art_aborted -- 12 frames, 0 episode(s)"

actually on disk:  data/ empty, 0 parquet rows, meta/info.json total_frames = 0
```

"Episode saved ... 12 frames" for a dataset holding only `meta/info.json`. The `#708` parquet-truth gate below it reconciles `episode_count` against `meta.total_episodes` but never `frame_count`, so the inflated count flows into both the text and the JSON payload unchecked.

This is reachable through normal use, not a corner case: `run_multi_policy` calls `clear_episode_buffer()` from its `finally` when a rollout bails mid-episode (MuJoCo and Isaac), and `reset()`'s docstring recommends it to callers as the way to discard a partial rollout.

## Change

Un-count discarded frames from both counters, **conditional on the discard actually happening**:

- **Discard succeeded** — the frames are gone and will never be written, so `frame_count` drops by the buffered amount and `episode_frame_count` resets.
- **No clear surface available** — nothing was thrown away. The frames are still buffered and the warning tells the caller to drain them with `stop_recording()`/`save_episode()`, which writes them to disk. Neither counter changes.

The conditional matters. Decrementing unconditionally (the obvious one-liner) introduces the opposite defect: measured on the failure path, it reports **0 frames for 7 frames that reached disk**. Keeping the counters there also lets the recommended drain report the frames it really wrote, and lets a later successful clear still un-count them.

`clear_episode_buffer`'s docstring now states the counter contract, since it is caller-visible.

## Verification

Blast radius: 77 test files (everything referencing the counters, the recorder, or the recording lifecycle), same selection on both trees, `MUJOCO_GL=egl`.

|---|---|---|
| collected | 2911 | 2911 |
| failed | 1 | 15 |
| passed | 2908 | 2891 |

**Branch-only failing IDs: none.** The single branch failure (`test_every_on_frame_call_site_excludes_a_lost_recording_frame`, `KeyError`) fails identically on both trees under this subset and passes 7/7 in isolation on both — a selection artifact, not a regression. Of `main`'s 15, 13 are this PR's pre-fix failures and one is a pre-existing lerobot/draccus e2e-train env failure. `main` also module-skipped 5 isaac recording tests via a transient `has_lerobot_dataset()` probe under load; all 5 pass on both trees when run directly.

Pre-fix split, new + changed tests: **13 failed / 147 passed** on `upstream/main`, all 160 pass after. Representative pre-fix messages:

```
reported total_frames=15 but only 5 frames reached disk; the 10 discarded frames are still counted
nothing reached disk, but frame_count still reports 12 - stop_recording asks this counter
  whether anything was captured before refusing an empty dataset
stop_recording reported success for a dataset with no frames on disk (recorder.frame_count=12)
assert 10 == 6   # resume() seeded from disk, then drifted on one abort
assert 14 == 2   # drift accumulating over three aborts
```

Four of the new tests **pass on both trees** — the no-overreach controls: a saved episode reports what it wrote, frames accumulate across saved episodes, discarding an empty buffer changes nothing, and `stop_recording` still finalizes a dataset that really has frames on disk.

`tests/test_dataset_recorder.py` needed four assertion updates. One is the comment that stated the un-counted total as intended while never actually asserting `frame_count`. The other two paths asserted `episode_frame_count == 0` after a *failed* discard, justified as "does not carry over the discarded frames" — but nothing was discarded there, so the premise was false; they now assert the counters are preserved and that the recommended drain reports the frames it wrote.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` identical to `main` (19 errors on both, zero branch-only).

## Artifact

Same script run against both trees, MuJoCo sim headless. Top row aborts the rollout; bottom row is the control that does not. Each badge is graded from the measured report-vs-disk agreement, not from the wording.

recorder frame accounting

The control row is byte-identical on both trees — 12 frames written, 12 frames when reopened through `LeRobotDataset` — so the change only affects a rollout whose frames were discarded.

## Scope

Two things I checked and deliberately left alone:

- **`push_to_hub`'s empty-dataset guard is not defeated by this drift.** It is `frame_count == 0 or episode_count == 0`, and `episode_count` only advances on a successful `save_episode`, which LeRobot refuses on an empty buffer. So the `episode_count` half already holds. Its refusal *message* did quote the inflated frame count; that is fixed as a consequence.
- **`stop_recording` does not cross-check `frame_count` against the parquet** the way the `#708` gate cross-checks `episode_count`. Adding that is defence-in-depth for a different contract and belongs in its own change; this PR removes the drift at its source so the existing refusal works again.

### DIFF
--- a/changelog.d/2388-recorder-uncount-discarded-frames.md
+++ b/changelog.d/2388-recorder-uncount-discarded-frames.md
@@ -1,0 +1,12 @@
+### Fixed
+
+`DatasetRecorder.clear_episode_buffer` now un-counts the frames it discards.
+`add_frame` counts a frame when it is buffered, but frames only reach disk on
+`save_episode`, so a discarded (aborted) episode left `frame_count` reporting a
+total no parquet row backed. Because `stop_recording` asks that counter whether
+anything was ever captured before refusing an empty dataset, the inflated count
+blinded the refusal: discarding a partial rollout and then stopping reported
+success for a dataset holding only `meta/info.json`. The un-count applies only
+when the discard actually happened - if no clear surface is available the frames
+are still buffered and will be written by the recommended
+`stop_recording`/`save_episode` drain, so both counters are left in place.
--- a/strands_robots/dataset_recorder.py
+++ b/strands_robots/dataset_recorder.py
@@ -1634,6 +1634,15 @@ class DatasetRecorder:
           * leave the buffer in place and warn (caller must ``stop_recording`` /
             ``save_episode`` to drain it before recording again).
 
+        Discarded frames are un-counted from both ``frame_count`` and
+        ``episode_frame_count``, because ``add_frame`` counts at buffer time
+        while frames only reach disk on ``save_episode``. Both counters
+        therefore describe frames that reached, or will reach, disk - never
+        frames that were thrown away. When no clear surface is available the
+        frames are still buffered and will be written by the
+        ``stop_recording``/``save_episode`` drain the warning recommends, so
+        neither counter changes in that case.
+
         Returns:
             True if the buffer was actively cleared; False if no clear surface
             was available (a warning is logged in that case).
@@ -1650,10 +1659,25 @@ class DatasetRecorder:
             logger.warning("clear_episode_buffer failed: %s", e)
             cleared = False
 
-        # Reset the per-episode frame counter regardless: the next episode
-        # reports frames from 0. frame_count (cumulative) is left untouched
-        # since those frames were really written to disk only on save_episode.
-        self.episode_frame_count = 0
+        if cleared:
+            # The buffered frames are gone. They were counted optimistically by
+            # add_frame (which counts at buffer time) but never reached disk,
+            # since frames are only written by save_episode - so un-count them
+            # from the cumulative total as well as the per-episode one. Both
+            # counters describe frames that reached, or will reach, disk:
+            # resume() seeds frame_count from dataset.meta.total_frames, and
+            # stop_recording refuses an empty dataset by asking frame_count
+            # whether anything was ever captured. Leaving the discarded frames
+            # counted makes the recorder report a total no parquet row backs,
+            # and blinds that refusal.
+            self.frame_count -= self.episode_frame_count
+            self.episode_frame_count = 0
+        # When the discard did NOT happen the frames are still buffered, and the
+        # warning below tells the caller to drain them with
+        # stop_recording()/save_episode() - which writes them to disk. Nothing
+        # was thrown away, so both counters are left as they are: the eventual
+        # flush then reports the frames it really wrote, and a later successful
+        # clear can still un-count them.
 
         if not cleared:
             logger.warning(
--- a/tests/simulation/mujoco/test_stop_recording_finalize.py
+++ b/tests/simulation/mujoco/test_stop_recording_finalize.py
@@ -24,7 +24,12 @@ import pytest
 
 pytest.importorskip("mujoco")
 
+from strands_robots.dataset_recorder import DatasetRecorder  # noqa: E402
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Fixes a counter-drift bug in `DatasetRecorder`: `add_frame` counts a frame at buffer time, but frames only reach disk on `save_episode`, so `clear_episode_buffer` discarding an aborted episode left `frame_count` reporting frames no parquet row backs. Since `RecordingMixin.stop_recording` asks `frame_count` whether anything was captured before refusing an empty dataset, the inflated count let an abort-then-stop flow report success for a dataset holding only `meta/info.json`. The fix un-counts discarded frames from both counters, correctly conditional on the discard actually happening - when no clear surface is available the frames remain buffered and pending the recommended drain, so both counters are preserved (the unconditional decrement would have introduced the inverse defect, which the PR measures and the new `_WedgedDataset` tests pin). I verified the interaction with `stop_recording`'s three-case flow in `recording.py`: after a failed clear, `pending > 0` now routes to the flush path so the still-buffered frames are written and reported, which is strictly better than the old behavior (counter zeroed, buffer wedged, case-2 finalize). Reproduced the test results locally on the head SHA: the 11 new counter-contract tests, the 2 new mujoco finalize tests, and the 5 updated `clear_episode_buffer` tests all pass; the failures in `tests/test_dataset_recorder.py` on this branch are the identical pre-existing bucket-sync environment failures on `upstream/main`. `ruff check` / `ruff format --check` clean on the changed files.

## What's good

- The conditional un-count is the right call, and the PR documents why the obvious one-liner is wrong - with a measured counterexample.
- Tests assert both directions of the contract plus no-overreach controls (saved episodes, empty-buffer discard, repeated discards, resume-seeded consistency), per AGENTS.md's pin-every-reviewed-fix convention.
- New tests inject stub datasets directly, so they never touch the shared `$HF_LEROBOT_HOME` cache (AGENTS.md rule 15 is satisfied by construction).
- Docstring states the caller-visible counter contract; changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; the two premise-false assertions in the existing tests were corrected rather than papered over.

PR #2389, opened 2026-08-17
fix(dataset_recorder): un-count the frames an aborted episode discarded

cagataycali:
## Problem

`DatasetRecorder.add_frame` counts a frame into `frame_count` (cumulative) and `episode_frame_count` at **buffer** time; frames only reach disk when `save_episode` flushes the episode. `clear_episode_buffer` discarded the buffered frames but left `frame_count` still counting them, so after any aborted episode the recorder permanently reported frames that no parquet row accounts for.

That is not a corner case: it is the abort path. Both `run_multi_policy` implementations call it from their `finally` when a rollout bails mid-episode (`simulation/mujoco/simulation.py:5040`, `simulation/isaac/simulation.py:3924`).

The counter is on-disk truth for every consumer: `save_episode` returns it as `total_frames`, `push_to_hub` reports it as `frames` and refuses an empty dataset by asking it, and `resume` seeds it from `meta.total_frames`.

### The consequence that matters: a fail-loudly guard goes blind

`simulation/recording.py:640` is `stop_recording`'s case 3 - it asks `frame_count` whether anything was ever captured, and its own comment says it exists because *"Previously stop_recording reported success with '0 frames, 0 episode(s)', silently producing a dataset with only meta/info.json (no parquet/video)."*

An inflated `frame_count` makes `captured == 0` false, so that refusal never fires. Measured end to end with a real `LeRobotDataset` on disk (MuJoCo `so101`, 40 recorded frames, then the abort):

|---|---|---|
| `recorder.frame_count` | **40** | 0 |
| `meta/info.json` `total_frames` | 0 | 0 |
| `*.parquet` on disk | **0** | 0 |
| `stop_recording()` | **`success`** - "Episode saved to LeRobotDataset ... 40 frames" | `error` - "captured no frames - dataset would be empty" |

Both runs leave a directory holding nothing but `meta/info.json`. On `main` that is reported as a saved dataset of 40 frames.

## Change

Reconcile the counters with whichever way the discard actually went:

- **cleared** - the frames can never reach disk, so deduct them from the cumulative total and reset the per-episode counter.
- **not cleared** (no clear surface on this LeRobot version, or the dataset raised) - the frames are still queued and the warning already tells the caller to drain them with `save_episode`/`stop_recording`, which writes them. Both counters already describe that outcome, so they are left alone.

That second branch is load-bearing, not defensive. A plain unconditional `frame_count -= episode_frame_count` fixes the total but then **under**-reports the not-cleared path by exactly the frames it is about to flush. Measured against this PR's tests:

| variant | cumulative total vs disk | failed-discard boundary |
|---|---|---|
| `upstream/main` | 12 failed | - |
| unconditional deduct | fixed | **6 failed** |
| this PR (follows `cleared`) | fixed | fixed |

## Tests

`tests/test_recorder_counters_track_on_disk_frames.py` drives the real `add_frame`/`save_episode`/`clear_episode_buffer` against a dataset double that separates the open buffer from written frames, so a counter is only ever compared with what actually landed. The existing fakes in `test_dataset_recorder.py` hand-assign `frame_count` and define no `add_frame`/`save_episode` at all, which is why five passing tests never noticed.

Pre-fix split on `3aac69d7`: **12 failed / 147 passed**, post-fix **159 passed**. Headline failure text:

```
save_episode reported total_frames=15 but only 5 frame(s) reached disk;
the 10 frame(s) discarded by clear_episode_buffer are still counted

stop_recording reported success for a dataset with no frames on disk
(recorder.frame_count=12)
```

Controls that pass on **both** trees (they pin that the change does not overreach): counters on the normal no-abort path, `episode_frames`/`total_frames` across two saved episodes, clearing an untouched buffer, and `stop_recording` still finalizing a dataset that really has frames.

Two existing tests changed. `test_clear_episode_buffer_warns_when_no_surface` and `test_clear_episode_buffer_swallows_dataset_error` asserted the per-episode counter resets on the not-cleared path, and their comments described the frames as "discarded" - but that branch is precisely the one where nothing was discarded and the frames survive to be written.

## Verification

- Blast radius (66 test files touching `frame_count`/`clear_episode_buffer`/`stop_recording`/`save_episode`/`total_frames`, plus the repo-wide docstring/ASCII/changelog guards), run on this branch and on a clean `upstream/main` worktree with the same selection: **branch 2 failed / 2720 passed, base 10 failed / 2710 passed**. Branch-only failures: **none**. The 8 base-only failures are this PR's pre-fix failures; the 2 remaining are pre-existing on `main` (`test_recording_frame_loss_is_not_tolerated`, and `test_lerobot_e2e` failing on a local `draccus` version).
- `ruff check` / `ruff format --check` / `mypy` over `strands_robots tests tests_integ`: clean, no new mypy errors versus base.

## Artifact

Rollout in MuJoCo sim (headless EGL): the 40 frames `run_policy` recorded, the abort that discards them, and what each tree then reports against the dataset actually on disk.

recorder abort counters

The recorded rollout itself ([mp4](https://raw.githubusercontent.com/cagataycali/robots/media-recorder-abort-counters/recorded-rollout.mp4)): 21 frames, all 20 transitions moving (per-frame mean abs delta 2.34 to 8.27).

### DIFF
--- a/changelog.d/2389-recorder-uncount-discarded-frames.md
+++ b/changelog.d/2389-recorder-uncount-discarded-frames.md
@@ -1,0 +1,11 @@
+### Fixed
+
+`DatasetRecorder.clear_episode_buffer` no longer leaves the frames it discarded
+counted in the cumulative `frame_count`. `add_frame` counts a frame when it
+buffers it, so an aborted episode - the path `run_multi_policy` takes when a
+rollout bails mid-episode - used to inflate every later
+`save_episode()['total_frames']` and `push_to_hub()['frames']`, and blinded
+`stop_recording`'s "captured no frames" refusal into reporting success for a
+dataset holding only `meta/info.json`. When the buffer could not be discarded
+the frames are still queued for the next `save_episode`, so both counters are
+left describing them.
--- a/strands_robots/dataset_recorder.py
+++ b/strands_robots/dataset_recorder.py
@@ -1634,9 +1634,20 @@ class DatasetRecorder:
           * leave the buffer in place and warn (caller must ``stop_recording`` /
             ``save_episode`` to drain it before recording again).
 
+        The frame counters follow that outcome. ``add_frame`` counts a frame
+        into both ``frame_count`` and ``episode_frame_count`` when it buffers
+        the frame, not when ``save_episode`` writes it, so a successful discard
+        has to take the discarded frames back out of the cumulative
+        ``frame_count``. When the discard does NOT happen the frames are still
+        queued for the next ``save_episode``, so both counters are already right
+        and are left alone.
+
         Returns:
-            True if the buffer was actively cleared; False if no clear surface
-            was available (a warning is logged in that case).
+            True if the buffer was actively cleared - the discarded frames are
+            deducted from ``frame_count`` and ``episode_frame_count`` resets to
+            0. False if no clear surface was available or the dataset raised, in
+            which case the frames remain buffered, both counters are left
+            describing them, and a warning is logged.
         """
         cleared = False
         try:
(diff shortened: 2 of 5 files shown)


cagataycali:
Closing as a duplicate of #2388, which is already approved.

Both branches landed the same one-line correction in `clear_episode_buffer`:

```python
if cleared:
    self.frame_count -= self.episode_frame_count
    self.episode_frame_count = 0
```

`add_frame` bumps both counters at buffer time while frames only reach disk on
`save_episode`, so a successful discard has to take the discarded frames back
out of the cumulative total. The two diffs differ only in comment wording and
in test naming; #2388 additionally covers repeated discards and the
all-frames-discarded case, so nothing here is lost by closing this one.

One wording improvement from this branch is worth folding in if the docstring
is ever touched again: the `Returns:` section can state the counter outcome for
each branch rather than only whether the buffer was cleared.

PR #2390, opened 2026-08-17
fix(training): decode a text extra value the way lerobot's own CLI decodes it

cagataycali:
## Problem

`LerobotTrainer.build_config` assigns each `TrainSpec.extra` entry straight onto a field of lerobot's typed config tree. `build_command` renders the same entry as `--key=value` for lerobot's draccus CLI — the argv this module's own docstring describes the typed config as *corresponding to*. For every non-string field that correspondence did not hold, because the value was applied with a raw `setattr`:

```python
extra={"policy.freeze_vision_encoder": "false"}   # stores the string "false"
```

A non-empty string is truthy, so a SmolVLA run kept its vision encoder frozen while the identical `--policy.freeze_vision_encoder=false` unfroze it. Nothing raised, nothing warned. Measured on the same spec through both trees:

| | `cfg.policy.freeze_vision_encoder` | `bool(value)` |
|---|---|---|
| before | `'false'` (`str`) | `True` — still frozen |
| after | `False` (`bool`) | `False` — unfrozen |

SmolVLA ships `freeze_vision_encoder=True` and `train_expert_only=True` in its own config, so this is the one knob a caller must reach to train more than the action expert. `method` selects the tuning strategy, not lerobot's per-policy defaults.

text extra value decoding, before and after

## Change

A **text** value now travels the two stages lerobot's own parser uses — `cfgparsing.parse_string` (YAML scalar rules) then `draccus.decode` against the field's annotation — so the two paths honour the same spellings by construction rather than by a hand-copied list. Measured agreement, in-process vs feeding `build_command`'s argv to `draccus.parse`:

| spelling | CLI | in-process before | in-process after |
|---|---|---|---|
| `false` `False` `FALSE` `no` `off` | `False` | the string itself | `False` |
| `true` `yes` `on` | `True` | the string itself | `True` |
| `0` `1` `maybe` | refused | the string itself | refused |

**11/11 agree after; 8/11 differed before.** Note `0` and `1` are integers to draccus, not booleans, so a coercion that accepted them would have diverged from the CLI in the other direction.

Text that does not decode is refused naming the field, its declared type and a spelling that works, rather than stored with the wrong type:

```
extra['policy.freeze_vision_encoder']='0' does not decode to bool, the declared type of
'freeze_vision_encoder' on SmolVLAConfig. Values are read with lerobot's own CLI decoder,
so the accepted spellings are the ones '--policy.freeze_vision_encoder=0' accepts - for a
boolean: false/no/off or true/yes/on (any case); note 0 and 1 are not booleans. Pass a bool
value directly to skip decoding.
```

The annotation is resolved with `typing.get_type_hints`, **not** read off `dataclasses.fields()`. Six lerobot policy configs (`eo1`, `evo1`, `fastwam`, `molmoact2`, `vla_jepa`, `xvla`) are compiled with postponed annotations, so **45 of their boolean fields report the string `"bool"`** there — comparing that against `bool` is `False`, which would silently skip exactly those configs.

Deliberately unchanged: a value already of the field's own Python type is passed through untouched (so every existing caller is byte-identical), and a field that already admits text keeps the caller's text verbatim. Round-tripping a name through a boolean to retype it would corrupt e.g. `wandb.project="off"`; that is the one place the paths differ on purpose, and it is pinned by a test.

`draccus` is imported inside the function rather than declared: lerobot pins `draccus>=0.11.6,<0.12.0` and cannot function without it, and re-declaring that bound here would create a second owner for one pin. The API and the full spelling table were verified identical on both draccus 0.10.0 and 0.11.6.

## Tests

`tests/training/test_extra_value_decoding.py`, 30 cases. On `upstream/main`: **22 failed / 8 passed**; after: **30 passed**. The headline failure quotes the measurement rather than a wording diff:

```
extra['policy.freeze_vision_encoder']='false' reached the config as 'false' (str);
a non-empty string is truthy, so the encoder stays frozen while
--policy.freeze_vision_encoder=false unfreezes it
```

The parity cases use lerobot's own CLI parser as the oracle, so they pin agreement between the two paths rather than a copied list of spellings. The 8 that pass on both trees are the no-overreach controls: a typed value is passed through unchanged (bools and ints, both values), a string field keeps its text verbatim (3 cases), and an unknown key is still ignored with a warning.

## Verification

- `tests/training` plus the repo-wide guards (dependency audit, docstring xrefs, unicode dashes, ASCII logs, `Args:` completeness, changelog fragments, dangling doc refs): **2847 passed**, 3 failed. All 3 fail identically on `upstream/main` and are a local environment artifact — `TypeError: encode() takes 1 positional argument but 2 were given` from draccus 0.10.0 being below lerobot's own `draccus>=0.11.6` floor on this box.
- `ruff check` / `ruff format --check` clean; `mypy` reports no error in either changed module (the 4 it does report are pre-existing in `utils.py`, `simulation/ik.py`, `mesh/core.py`).

## Notes on scope

The reported case also raises whether `method="full"` should itself unfreeze a policy that freezes most of itself by default. That would either change what a shipped `method` value means for existing callers (a large increase in trainable parameters, and so in memory) or add a new value to the public `method` vocabulary — a contract choice, so it is not made here. What this change does is make the documented escape hatch reliable: the text form of the override now works, and the behaviour is written down in `TrainSpec.extra` and in `docs/training/overview.md` with the SmolVLA case spelled out.

### DIFF
--- a/changelog.d/2390-training-extra-text-decoding.md
+++ b/changelog.d/2390-training-extra-text-decoding.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **training**: a text value in `TrainSpec.extra` is now decoded with the same decoder lerobot's `--key=value` CLI uses, so one spec means one run whichever path the lerobot backend takes. `extra={"policy.freeze_vision_encoder": "false"}` previously stored the truthy string `"false"` and left the encoder frozen while the identical CLI flag unfroze it, with nothing raised or warned. `false`/`no`/`off` and `true`/`yes`/`on` (any case) are booleans; text that does not decode to the field's declared type is refused naming the field, its type and a spelling that works. A value already of the field's own Python type is passed through unchanged.
--- a/docs/training/overview.md
+++ b/docs/training/overview.md
@@ -131,6 +131,53 @@ TrainSpec(..., method="lora", lora_r=16, extra={"policy_type": "pi05"})
 # -> lerobot_train --peft.method_type=LORA --peft.r=16 --policy.type=pi05
 ```
 
+#### Passing lerobot's own flags through `extra`
+
+Any key in `extra` that names a real field of lerobot's config tree is applied to
+it, dotted for a sub-config (`policy.*`, `dataset.*`, `wandb.*`). A key that
+matches no field is ignored with a warning, so a typo can never become an
+arbitrary flag.
+
+A value may be written either as the field's own Python type or as text. Text is
+read with lerobot's own draccus decoder - the same one behind `--key=value` - so
+one spec means one run whether the backend builds the config in-process or shells
+out:
+
+```python
+# these two are the same request
+extra={"policy.freeze_vision_encoder": False}
+extra={"policy.freeze_vision_encoder": "false"}
+```
+
+For a boolean field that decoder accepts `false`/`no`/`off` and
+`true`/`yes`/`on` in any case. `0` and `1` are integers to it, not booleans, and
+are refused - as they are on the command line. Text that does not decode to the
+field's declared type is refused naming the field and its type, rather than
+stored as-is: a stored string is truthy, so `"false"` would otherwise read as
+"true" and silently invert the flag.
+
+Some policies freeze most of themselves by default, which `method="full"` does
+not override - it selects strands' tuning strategy, not lerobot's per-policy
+defaults. SmolVLA, for instance, ships `freeze_vision_encoder=True` and
+`train_expert_only=True`, so full-model finetuning means asking for both:
+
+```python
+TrainSpec(
+    dataset_repo_id="org/tictactoe",
+    base_model="lerobot/smolvla_base",
+    output_dir="/tmp/ft_out",
+    steps=20000,
+    method="full",
+    extra={"policy_type": "smolvla",
+           "policy.freeze_vision_encoder": False,
+           "policy.train_expert_only": False},
+)
+```
+
+Check what a policy freezes by default before assuming `method="full"` trains
+all of it: `dataclasses.fields()` on its lerobot config class lists every knob
+and its default.
+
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Fixes a silent type-corruption bug in `LerobotTrainer.build_config`: a text value in `TrainSpec.extra` (e.g. `"false"` for `policy.freeze_vision_encoder`) was assigned raw onto lerobot's typed config, where the truthy string inverted the flag's meaning relative to the identical `--key=value` CLI form. The fix routes text through the same two stages lerobot's own draccus parser uses (`cfgparsing.parse_string` then `draccus.decode` against the field's resolved annotation), so the two paths agree by construction; undecodable text is refused with an explicit `ValueError` naming the field, its declared type and a working spelling, rather than stored. Non-string values and string-typed fields are passed through byte-identical, so existing correct callers are unaffected.

## What's good

- The 30-test suite uses lerobot's own CLI parser as the oracle rather than a hand-copied spelling table, and includes no-overreach controls (typed values pass through, string fields keep text verbatim, unknown keys still warn-and-ignore). Verified locally: 30/30 pass on draccus 0.10.0, the older of the two versions the PR claims parity on.
- The annotation is resolved via `typing.get_type_hints` rather than `dataclasses.fields()`, correctly handling the six lerobot policy configs compiled with postponed annotations - and that subtlety is pinned by its own test.
- The broad `except` around the decode is a translate-and-chain re-raise (`from e`), not a swallow; the refusal is actionable and nothing degrades to a silent default.
- The behaviour change (refusing undecodable text) raises loudly and is documented in `TrainSpec.extra`, `docs/training/overview.md` and a changelog fragment - scope discipline on the `method="full"` contract question is explicitly deferred rather than smuggled in.

PR #2392, opened 2026-08-17
fix(robot): forward the spawn parameters add_robot accepts

cagataycali:
## Problem

`Robot(...)` is documented as a thin front door to `SimEngine.add_robot`, and it forwards
four of that method's six parameters. `orientation` and `keyframe` were not forwarded, so
they fell through to `**kwargs` — which the factory sends to the *backend constructor*, and
that constructor takes `**kwargs` of its own. Both were absorbed with no error and no
warning, and the factory reported success.

Measured on the built-in `panda`, through the factory only:

| Request | `add_robot` gives | `Robot(...)` gave |
|---|---|---|
| `keyframe="home"` spawn qpos | the model's `home` pose | **all zeros** (zero configuration) |
| `keyframe="home"` spawn ctrl | the command that holds that pose | **all zeros** |
| distance from `home` after 320 steps | 0.0066 rad | **1.5708 rad**, from step 0 on |
| `orientation=[0.7071, 0, 0, 0.7071]` | `[0.7071, 0, 0, 0.7071]` | **`[1, 0, 0, 0]`** (unrotated) |
| `keyframe="not_a_keyframe"` | error naming the available keyframes | **accepted** |
| `orientation=[0, 0, 1]` | `'orientation' must be a 4-element vector, got 3` | **accepted** |

The two halves compound. A keyframe spawn is how a caller reaches a pose a policy was
trained from — the zero configuration is out-of-distribution — and a MuJoCo `<key>` pairs
that pose with the actuator command that *holds* it, so dropping the parameter loses both.
The bypassed refusals are worse than the dropped values, because `add_robot` states the
keyframe one outright: "a hard error that names the available keyframes; **it never
silently falls back to zeros**". Through the factory it silently fell back to zeros.

## Why this is a forwarding fix and not a new contract

`orientation` and `keyframe` are declared by `SimEngine.add_robot` itself and by all three
backends (MuJoCo, Newton, Isaac), so they are part of the backend-agnostic contract rather
than a backend extra. And `position` — their sibling in the same call — already documents
exactly the terms they now travel on:

> Passed through to the backend's `add_robot` verbatim, so the backend's contract governs:
> omitting it spawns at the origin, and a wrong-length, non-numeric or non-finite vector is
> refused with an actionable message [...] instead of being replaced by the origin.

with the call site adding that passing the caller's value through "keeps ONE source of truth
for that default instead of a copy that can drift". That reasoning was settled once for
`position`; this applies it to the other two parameters of the same call. No default changes
and no value is interpreted by the factory.

## Change

- Forward `orientation` and `keyframe` in the factory's `add_robot` call.
- Declare them on the three `@overload`s and the implementation, **keyword-only** and
  appended after the existing parameters, so no existing positional call shifts meaning and
  the overloads stay positionally faithful to the implementation.
- Document them beside `position`, whose contract they share, and add them to the factory
  reference table and the spawn-pose docs.

## Verification

`tests/test_robot_factory.py::TestFactorySpawnParameterForwarding`, run against `main`
with only the test file applied: **9 failed, 1 passed** → 10 passed after.

The headline tests fail pre-fix with measured values, not with a missing keyword:

```
Robot() cannot express add_robot parameter(s) ['keyframe', 'orientation']; a caller passing
one gets it absorbed by the backend constructor's **kwargs and silently dropped.

orientation: factory gave [1.0, 0.0, 0.0, 0.0] but add_robot gives [0.7071068, 0.0, 0.0, 0.7071068]
```

The first of those is the root-cause guard: it compares the factory's parameters against
`SimEngine.add_robot`'s, so a parameter added to the ABC later cannot be silently dropped
by the front door again.

The one test that passes on both trees is the no-overreach control — omitting both
parameters still gives the zero joint configuration, its actuators commanded to zero, and no
rotation, so an existing caller is unaffected.

Also run: `ruff check` / `ruff format --check` and `mypy` over `strands_robots tests
tests_integ` clean with no new findings, and `tests/test_robot_factory.py` fully green
(119 passed).

One capture script, run against `main` and against this branch — same camera, same steps,
only `strands_robots` differs.

`Robot("panda", keyframe="home")` stepping 320 times. On the left the requested pose never
arrives; on the right it arrives and the keyed actuator command holds it (0.0066 rad of
settle over the whole rollout):

keyframe rollout

Both parameters, plus the two refusals now reaching the caller:

spawn parameter forwarding

[MP4 of the rollout](https://raw.githubusercontent.com/cagataycali/robots/media-factory-spawn-params/keyframe_rollout.mp4)

## Notes on scope

While measuring this I found that the factory reference table's `**kwargs` row claims
"Unknown kwargs raise `ValueError`", which does not hold in general — an arbitrary
`Robot("panda", total_nonsense=1)` is absorbed by the backend constructor too. Whether the
factory should reject unknown keywords is a behaviour change for existing callers rather
than a correction, so it is left out of this PR; with `orientation` and `keyframe` forwarded,
the two absorbed names that were real `add_robot` parameters are no longer among them.

### DIFF
--- a/changelog.d/2392-factory-forwards-spawn-parameters.md
+++ b/changelog.d/2392-factory-forwards-spawn-parameters.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **robot**: `Robot(...)` now forwards `orientation` and `keyframe` to the backend's
+  `add_robot`. Both are `SimEngine.add_robot` parameters, but the factory did not pass them
+  on, so they fell into `**kwargs` and were absorbed by the backend constructor: a requested
+  base rotation or `<keyframe>` spawn pose was silently dropped and the robot came up
+  unrotated in the zero configuration while the factory reported success. The refusals
+  `add_robot` documents for a malformed quaternion and for an unknown keyframe now reach the
+  caller too, instead of the unknown keyframe falling back to zeros. Both are keyword-only
+  and omitting them is unchanged.
--- a/docs/getting-started/robot-factory.md
+++ b/docs/getting-started/robot-factory.md
@@ -1,5 +1,5 @@
 ---
-description: Robot(name, mode, backend, urdf_path, cameras, position, data_config, mesh, peer_id, **kwargs) - the full signature with every kwarg explained.
+description: Robot(name, mode, backend, urdf_path, cameras, position, data_config, mesh, peer_id, orientation, keyframe, **kwargs) - the full signature with every kwarg explained.
 ---
 
 # Robot factory
@@ -27,6 +27,8 @@ robot = Robot("so100", mode="auto")  # probes USB, falls back to sim
 | `data_config` | str | `None` | GR00T data_config name. |
 | `mesh` | bool | `True` | Auto-join Zenoh mesh. |
 | `peer_id` | str | `None` | Stable mesh peer id. Auto-generated if omitted. |
+| `orientation` | list | `None` | Robot base orientation `[w, x, y, z]` in sim world. |
+| `keyframe` | str \| int | `None` | Spawn in a model `<keyframe>` pose (name or index) instead of the zero configuration. |
 | `**kwargs` | | | Forwarded to backend constructor. Unknown kwargs raise `ValueError`. |
 
 ## Name resolution
--- a/docs/simulation/world-building.md
+++ b/docs/simulation/world-building.md
@@ -66,6 +66,13 @@ out-of-distribution:
 sim.add_robot(name="panda", data_config="panda", keyframe="home")  # or keyframe=0
 ```
 
+The `Robot(...)` factory forwards `keyframe=` (and `orientation=`) to
+`add_robot`, so a one-line spawn reaches the same pose:
+
+```python
+robot = Robot("panda", keyframe="home")
+```
+
 The pose is applied to the robot's joints by name and is restored by `reset()`,
 so a keyframe spawn is sticky across episodes. A MuJoCo `<key>` pairs that pose
 with the actuator command that *holds* it, and both are applied and restored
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Forwards `orientation` and `keyframe` from the `Robot(...)` factory to the backend's `add_robot`, closing a silent-drop path where both fell into `**kwargs`, were absorbed by the backend constructor's own `**kwargs`, and the factory reported success while the robot spawned unrotated in the zero configuration. Both parameters are already declared by `SimEngine.add_robot` and all three backends (MuJoCo, Newton, Isaac), so this is a forwarding fix on the existing contract rather than new API surface, and the values travel on the same verbatim-passthrough terms `position` already documents. Verified locally: the new `TestFactorySpawnParameterForwarding` suite passes on the PR head (10/10), and the root-cause signature guard fails on `main` exactly as the description claims, so the class of bug (an ABC parameter unreachable through the front door) is pinned against recurrence.

## What's good

- The new parameters are keyword-only and appended after existing ones, so no positional call shifts meaning and the overloads stay faithful to the implementation.
- Real-mode behavior is unchanged in effect: previously both names were absorbed by `HardwareRobot.__init__`'s `**kwargs`; now the factory captures them - the caller-observable outcome is identical, matching the pre-existing `position` pattern.
- The signature-comparison test guards the root cause, not just the two symptoms - a parameter added to `SimEngine.add_robot` later cannot be silently dropped by the factory again.
- Changelog fragment, docs table, and spawn-pose docs all updated in the same PR; tests use `tmp_path` and `mesh=False`, no host paths, no shared-cache reads.
- Scope discipline: the stale `**kwargs`-raises-`ValueError` doc claim is noted and deliberately deferred as a behavior change rather than bundled in.

PR #2393, opened 2026-08-17
fix(docs): state the factory contract Robot() implements

cagataycali:
## Problem

`docs/getting-started/robot-factory.md` is the reference a caller reads before writing `Robot(...)`. Three of its statements describe a factory that does not exist.

**1. The `mesh` Default cell named the opposite state.** The row read `` `mesh` | bool | `True` ``. The parameter defaults to `None`, which consults `STRANDS_MESH` and leaves mesh **off**:

```python
>>> inspect.signature(Robot).parameters["mesh"].default
None
>>> Robot("so100", mode="sim").mesh
None
```

The harm compounds. Because the page said mesh was already on, the only knobs it offered were `Robot("so100", mesh=False)` and `STRANDS_MESH=false` -- both no-ops for the state the reader is actually in -- and the spelling that *does* enable it (`mesh=True`, `STRANDS_MESH=true`) appeared nowhere on the page.

**2. The Mesh snippet could not run.** It read `.mesh` attributes on that robot, so copied as written it raises:

```
sim = Robot("so100")
sim.mesh.peer_id   # 'so100_sim-a1b2c3d4'
    AttributeError: 'NoneType' object has no attribute 'peer_id'
```

**3. The `**kwargs` row promised a refusal that does not happen.** It read "Unknown kwargs raise `ValueError`". They are forwarded and ignored:

```python
>>> Robot("so100", mode="sim", urdf_path=..., definitely_not_a_forwardable_kwarg=1)
<MuJoCoSimEngine>          # no ValueError
```

`strands_robots/robot.py`'s own `Args:` entry is already correct here -- "Forwarded to the underlying backend constructor" -- so the table appended a guarantee the code never made. A reader who believes it ships the typo.

## Change

Correct the three statements, and grade the page against `Robot()` itself rather than against a hand-copied expectation:

- every Default cell must equal the signature default;
- every parameter must have a row;
- a row that promises an exception is verified **by raising it**, not by wording -- if the `**kwargs` row claims `ValueError`, the test constructs a robot with an unrecognized kwarg and requires it. A page that promises no refusal is graded as clean, so implementing one later and documenting it still passes;
- the Mesh section must name a spelling `_mesh_env_opt_in` accepts.

The row parser splits on *unescaped* pipes, so a union Type cell is graded rather than silently dropped. This is load-bearing today: #2392's `keyframe` row spells its type `str \| int`, and a naive split would drop that row and report the parameter as undocumented. 11 of the table's 12 rows are graded (`**kwargs` is variadic and has no default).

The precedence the Mesh section now documents was measured, not assumed:

| construction | `.mesh` |
|---|---|
| bare `Robot("so100")`, `STRANDS_MESH` unset | `None` |
| `Robot("so100", mesh=True)` | joined |
| `STRANDS_MESH=true`, bare `Robot(...)` | joined |
| `STRANDS_MESH=false` + `mesh=True` | `None` (kill switch) |

## Scope

Documentation and a guard only; no behaviour change. Whether the factory *should* refuse an unknown kwarg is a separate question -- it would change construction for existing callers -- so this states what it does today rather than changing it.

`tests/mesh/test_mesh_env_opt_in_documented_default.py` already grades this same opt-in contract wherever a table's first cell is the environment variable. This page states it as a constructor default instead, which that guard does not read, so the level it enforces at never reached here.

## Verification

Pre-fix, with the final test file copied onto an unmodified `main` tree: **3 failed, 4 passed**.

```
FAILED ...::test_no_row_states_a_default_the_signature_contradicts
  line 28: `mesh` documents `True` but omitting it yields None
FAILED ...::test_a_promised_unknown_kwarg_refusal_happens
  Failed: DID NOT RAISE <class 'ValueError'>
FAILED ...::test_the_section_names_an_enabling_spelling
  the STRANDS_MESH spellings it shows are ['false'], none of which opts in,
  and it never passes mesh=True. Copied as written it raises AttributeError on None.
```

Post-fix: **7 passed**. The 4 that pass on both trees are the controls -- the page ships, enough rows are graded for a clean result to mean something, every parameter has a row, and a bare `Robot(...)` really does leave mesh off.

Affected suite (`tests/test_docs_*`, `tests/test_robot_factory.py`, `tests/mesh/test_mesh_env_opt_in_documented_default.py`, `tests/mesh/test_mesh_wiring.py`, plus the repo-wide docstring / unicode / ASCII / dependency / changelog guards): **724 passed, 1 skipped**, against 711 passed + the 3 pre-fix failures on base -- the arithmetic reconciles with no regressions. `ruff check`, `ruff format --check` and `mypy` clean over `strands_robots tests tests_integ`.

## Artifact

The page is the input and the interpreter is the oracle: each panel extracts the `## Mesh` snippet from *that tree's own* `robot-factory.md` and runs it line by line. Package code is identical in both panels.

factory reference

Left, the snippet dies on its second line. Right, it runs to completion and the values match the comments beside them (`sim.mesh` -> `None`, `peer_id` -> `'so100_sim-af5c92ae'`, `alive` -> `True`).

### DIFF
--- a/changelog.d/2393-factory-reference-states-the-real-contract.md
+++ b/changelog.d/2393-factory-reference-states-the-real-contract.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **docs**: the robot factory reference stated three contracts `Robot()` does not implement: the `mesh` Default cell read `True` where the parameter defaults to `None` (mesh is off unless `STRANDS_MESH` opts in), so the page named a state a bare `Robot(...)` is never in and never named the spelling that enables it; its Mesh snippet read `.mesh.peer_id` on that robot and raised `AttributeError` when copied; and the `**kwargs` row promised unknown keywords raise `ValueError` where they are forwarded and ignored. A new guard grades the page against `Robot()` itself - Default cells against the signature, a documented refusal by raising it, and the Mesh section against the opt-in resolver.
--- a/docs/getting-started/robot-factory.md
+++ b/docs/getting-started/robot-factory.md
@@ -25,9 +25,9 @@ robot = Robot("so100", mode="auto")  # probes USB, falls back to sim
 | `cameras` | dict | `None` | Real-hardware camera config. **Rejected in `mode="sim"`** - raises `ValueError`. |
 | `position` | list | `None` | Robot position `[x, y, z]` in sim world. |
 | `data_config` | str | `None` | GR00T data_config name. |
-| `mesh` | bool | `True` | Auto-join Zenoh mesh. |
+| `mesh` | bool \| None | `None` | Join the Zenoh fleet mesh. `None` consults `STRANDS_MESH`, which leaves it **off** unless set to `true`/`1`/`yes` - pass `mesh=True` to opt in per robot. |
 | `peer_id` | str | `None` | Stable mesh peer id. Auto-generated if omitted. |
-| `**kwargs` | | | Forwarded to backend constructor. Unknown kwargs raise `ValueError`. |
+| `**kwargs` | | | Forwarded to the backend or driver constructor as given. A name it does not recognize is ignored, not refused, so check the spelling against the forwardable list below. |
 
 ## Name resolution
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Docs-only correction plus a self-verifying guard. The PR fixes three false statements in `docs/getting-started/robot-factory.md` -- the `mesh` Default cell (`True` where the signature default is `None`), a Mesh snippet that raised `AttributeError` when copied, and a `**kwargs` row promising a `ValueError` refusal the factory never makes -- and adds `tests/test_docs_robot_factory_reference.py`, which grades the page against `Robot()` itself rather than a hand-copied expectation. I verified each corrected claim against the code at this head: `mesh: bool | None = None` with `_mesh_env_opt_in()` accepting `true`/`1`/`yes` (`strands_robots/robot.py`), the `STRANDS_MESH=false` hard kill switch resolved in `init_mesh` (`strands_robots/mesh/core.py`), and unknown kwargs forwarded rather than refused. The new test file passes on this head (7 passed), premise tests guard against the table silently drifting out of reach, the refusal test verifies by raising rather than by wording, and the pipe-splitting handles escaped-union Type cells so the `keyframe` row from #2392 stays graded. Changelog fragment present, no behaviour change, no host paths, ASCII-clean, ruff/format clean.

## What's good

- The oracle is the interpreter: Default cells are compared to `inspect.signature(Robot)`, a promised refusal is verified with `pytest.raises`, and the Mesh section is checked against `_mesh_env_opt_in` itself -- none of it can drift silently.
- Premise tests (`_MINIMUM_GRADED_ROWS`, page-ships) make a clean run mean the page was actually read, per the repo's no-silent-skip convention.
- Scope discipline: the open question of whether `Robot()` *should* refuse unknown kwargs is explicitly deferred rather than smuggled in.

PR #2394, opened 2026-08-17
fix(mesh/iot): cite the method that exists in the offload threat model

cagataycali:
## Problem

`strands_robots/mesh/iot/camera_offload.py`'s module docstring carries a **Bucket-ownership threat model** section:

> The S3 PutObject path in :meth:`CameraOffloader._upload_frame` does not pass an ``ACL=`` kwarg.

There is no `CameraOffloader._upload_frame`. The `put_object` call the sentence is a claim *about* lives in the public `CameraOffloader.upload_frame`. So a reader auditing that security claim - "does this really omit `ACL=`?" - cannot follow the pointer to the code, and the target names a private spelling of a public method.

## Why the existing guard reports the tree clean

`tests/test_docstring_xref_roles_resolve.py` exists for exactly this rot, and on the unfixed tree it passes:

```
$ pytest tests/test_docstring_xref_roles_resolve.py -q      # upstream/main, dead pointer present
8 passed
```

Its scope is fully-qualified `strands_robots.*` targets only. Measured over the package, that leaves the majority spelling ungraded:

| role spelling | count in the package | graded before | graded after |
|---|---:|:---:|:---:|
| fully-qualified (`strands_robots.a.b.C.d`) | 709 | yes | yes |
| **short-form, decidable head** (`C.d`) | **262** | **no** | **yes** |
| short-form, undecidable head | 41 | no | no (nothing to resolve against) |
| bare name (`reset`) | 1875 | no | no (Sphinx current-module context) |

The short-form spelling is what most method cross-references actually use, so the guarantee was enforced on one spelling of the promise and not the other.

## Change

1. Cite `CameraOffloader.upload_frame`, the method that exists.
2. Widen the guard to grade a short-form role whose head is **decidable**: looked up in the citing module first, else against a package-wide index of class names, and only when exactly one class answers to that name. An ambiguous head is skipped rather than guessed.

A member resolves permissively - an attribute, a dataclass field, an annotated class attribute anywhere in the MRO, a `__slots__` entry, or a name the class assigns to `self`. Strict about what counts as a *claim*, permissive about what counts as *evidence*, so the sweep can only ever under-report. That asymmetry is load-bearing: a class-level `hasattr` alone reads three real, correctly-cited members in this package as missing, and they are pinned as controls -

| role | why `hasattr(cls, member)` is False | verdict |
|---|---|---|
| `Cosmos3Embodiment.camera_keys` | dataclass field, no class-level default | member |
| `BootstrappedAccount.skipped` | dataclass field, no class-level default | member |
| `PolicyServer.port` | only ever assigned as `self.port = port` | member |

The remaining scope exclusions are unchanged and stated in the module docstring: a bare unqualified role, a head the package does not define, and roles into third-party packages have no decidable target without an intersphinx inventory.

## Verification

Pre-fix, the new test file copied onto `upstream/main` (it imports only stdlib, so it collects there):

```
1 failed, 12 passed

FAILED test_short_form_xref_roles_resolve - a short-form docstring cross-reference
must name a real member of its head. [...] Offending docstrings:
{'strands_robots.mesh.iot.camera_offload::<module>': ['CameraOffloader._upload_frame']}
```

Only the repo-wide sweep is load-bearing on the fix. The other 12 pass on both trees and are what bound the change: the planted-dead-pointer non-vacuity check, the three false-positive controls above, and the undecidable-head control that keeps third-party roles out of scope.

Blast radius - the mesh camera-offload suites plus every repo-wide docstring/string/dependency guard (13 files):

| tree | result |
|---|---|
| this branch | **664 passed**, 1 skipped, 0 failed |
| `upstream/main` + the new test file | 1 failed, 663 passed, 1 skipped |

663 + 1 = 664, so the single base failure is the headline test and there are no branch-only failures. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 19 errors on both trees, identical sets, none in the touched files.

A guard whose sweep silently stops reaching the docstrings would report a clean tree it never inspected, so `test_short_form_xref_roles_resolve` asserts a floor on how many roles it graded before asserting they resolve.

No visual artifact: this changes a docstring and a test, and no policy, simulation behaviour, rendering, recording or robot asset.

### DIFF
--- a/changelog.d/2394-short-form-xref-roles-resolve.md
+++ b/changelog.d/2394-short-form-xref-roles-resolve.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **mesh/iot**: the `camera_offload` bucket-ownership threat model cited `CameraOffloader._upload_frame` for the S3 `PutObject` path that omits an `ACL=` kwarg; no such method exists, so a reader auditing that claim could not reach the code it is about. Cite the public `upload_frame`, and widen `test_docstring_xref_roles_resolve` to grade short-form (`Class.member`) roles - the spelling most method cross-references use, and previously outside the guard's fully-qualified-only scope.
--- a/strands_robots/mesh/iot/camera_offload.py
+++ b/strands_robots/mesh/iot/camera_offload.py
@@ -47,7 +47,7 @@ Configuration
 
 Bucket-ownership threat model
 -----------------------------
-The S3 PutObject path in :meth:`CameraOffloader._upload_frame` does
+The S3 PutObject path in :meth:`CameraOffloader.upload_frame` does
 not pass an ``ACL=`` kwarg. The contract for the offload bucket is
 that the operator configures it with object-ownership control
 ``BucketOwnerEnforced`` (and a bucket policy that denies public
--- a/tests/test_docstring_xref_roles_resolve.py
+++ b/tests/test_docstring_xref_roles_resolve.py
@@ -22,19 +22,41 @@ import - and neither does the whitespace-collapsed form a renderer would see.
 The pattern here therefore admits whitespace inside the target and reports such
 a role, rather than failing to match it and exempting it from the guard.
 
-Scope is deliberately conservative: only targets that start with
-``strands_robots.`` are checked. Unqualified roles (``:func:`reset```) and roles
-into third-party packages resolve against Sphinx's current-module context or an
-intersphinx inventory that is not available here, so flagging them would produce
-false positives without catching the rot this guard exists to prevent.
+Two spellings of the same promise are graded, because a target only has to
+resolve - it does not have to be fully qualified to be checkable:
+
+* A fully-qualified target (``strands_robots.a.b.C.d``) resolves on its own.
+* A SHORT-FORM target (``C.d``) resolves once its head is pinned. The head is
+  looked up in the citing module first, and then, if the module does not import
+  it, against a package-wide index of class names - used only when exactly one
+  class in the package answers to that name, so an ambiguous head is skipped
+  rather than guessed. This is the spelling most method cross-references
+  actually use, and the earlier scope excluded it, which left the rot it exists
+  to prevent free to accumulate in the majority form.
+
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

One-word docstring fix plus the guard that makes it stick: the `camera_offload` bucket-ownership threat model cited `:meth:`CameraOffloader._upload_frame`` for the S3 `PutObject` path that omits `ACL=`, but no such method exists — the call lives in the public `upload_frame` (camera_offload.py:160). The PR corrects the citation and widens `tests/test_docstring_xref_roles_resolve.py` to grade short-form (`Class.member`) roles, which the previous fully-qualified-only scope left entirely unchecked. Test-only + docstring-only change: no policy, wire-format, public-API, or robot-behaviour surface.

## What's good

- Verified independently: at head `1da53a6c`, `upload_frame` exists and `_upload_frame` does not; the new sweep run against the merge base (`8a9a060d`) fails naming exactly `{'strands_robots.mesh.iot.camera_offload::<module>': ['CameraOffloader._upload_frame']}`, so the non-vacuity claim in the description reproduces.
- The sweep asserts a graded-count floor (`>= 100`) before asserting cleanliness, so a regex or import regression that stops the sweep reaching docstrings fails loudly instead of reporting a tree it never inspected.
- The permissive-evidence design (dataclass fields, MRO annotations, `__slots__`, `self.`-assigned names) is pinned by three real-package false-positive controls, and undecidable heads return `None` rather than guessing — the sweep can only under-report, which is the right failure direction for a docs guard.
- Changelog fragment `changelog.d/2394-short-form-xref-roles-resolve.md` per AGENTS.md step 3; no host paths, no emojis, plain-ASCII messages.

## Verification suggestions

- `pytest tests/test_docstring_xref_roles_resolve.py -q` on an environment without the full extras still passes the new short-form tests; the one failure I saw locally (`test_qualified_strands_robots_xref_roles_resolve`, via `strands_robots.device_connect` needing the optional `device_connect_edge`) pre-exists on the merge base and is environmental, not attributable to this PR.

PR #160, opened 2026-05-16
docs: full MkDocs Material site and GitHub Pages CI

cagataycali:
## Summary

A complete **MkDocs Material** documentation site for `strands-robots`, written from scratch against the **current `main`**. Every page is grounded in the actual source (`strands_robots/...`) and registry JSON — no example or API claim is invented.

- **29 content pages** (~2,300 lines) organized into 13 nav sections
- `mkdocs build --strict` passes with **zero warnings** — no broken nav, no dead internal links, all images/media resolve
- Verified facts against code: **68 robots / 8 categories / 106 aliases**, **64 simulation actions**, **27 GR00T data_configs**, 4 policy providers
- **CI added**: builds on every PR, deploys to GitHub Pages from `main`

## How the docs are organized

The site is structured around the mental model a user moves through: *understand → install → pick a robot → simulate → choose a policy → go to hardware → record → train → reference*.

| Section | Pages | Purpose |
|---------|-------|---------|
| **Home** | `index.md` | The one-screen pitch: what it is, the sim↔real diagram, install one-liners. |
| **AI Agents** | `agents.md` | Hand a `Robot()` to a Strands `Agent` and drive it in plain English. The framing page — robots are tools. |
| **Getting Started** | `quickstart`, `installation`, `robot-factory` | Five-minute path to a moving robot; the full extras matrix + platform notes; every `Robot()` kwarg. |
| **Robots** | `index` + `arms`, `bimanual`, `hands`, `humanoids`, `mobile` | Catalog landing (counts + category cards) and one page per category with per-robot tables (name, description, joints, aliases) verified against `registry/robots.json`, plus sim renders. |
| **Simulation** | `overview`, `world-building`, `domain-randomization` | The MuJoCo backend: the 64-action vocabulary, building/loading scenes, and `randomize(...)` for sim2real. |
| **Multi-robot Mesh** | `mesh.md` | Zenoh auto-discovery, RPC, emergency-stop, teleop, published topics. |
| **Policy Providers** | `overview` + `groot`, `cosmos3`, `lerobot-local`, `custom-policies` | The `Policy` ABC and one page per provider (GR00T, Cosmos 3, LeRobot-local), plus how to implement/register your own. |
| **Real Hardware** | `robot-control`, `tools` | `HardwareRobot` async task lifecycle + the `@tool` helpers (calibrate, camera, teleop, serial). |
[...]

PR #2395, opened 2026-08-17
docs: state the opt-in that joining the Zenoh mesh requires

cagataycali:
## Problem

Joining the Zenoh mesh is opt-in. `Robot()` and `Simulation()` both default `mesh` to `None`, which consults `STRANDS_MESH`; unset leaves the mesh off, so `robot.mesh` is `None`. Measured on this tree:

```
signature default 'mesh' = None
_mesh_env_opt_in() with STRANDS_MESH unset : False
  STRANDS_MESH='true' -> True    STRANDS_MESH='false' -> False
  STRANDS_MESH='1'    -> True    STRANDS_MESH='on'    -> False
  STRANDS_MESH='yes'  -> True    STRANDS_MESH='enabled' -> False
```

The default flipped in #630. Three pages still describe the world before it, in three spellings that fail differently.

**Runnable snippets.** `docs/mesh.md` and `README.md` construct a robot without opting in and then read `.mesh` attributes - 11 reads across four fenced blocks. Executed exactly as written, the mesh guide's own two-process example dies on its fourth line:

```
AttributeError: 'NoneType' object has no attribute 'peers'
sim_a.mesh is NoneType        (nothing printed)
```

**Prose.** Four paragraphs say a bare construction joins by itself: `README.md:39` (the opening description of the library) and `README.md:976`, `docs/mesh.md:11` (its first sentence), and `docs/security.md:36` (the mesh-authentication section, so the page overstates what a default install exposes). `README.md` contradicted itself - its environment-variable table already reads `unset (mesh off)`, 657 lines below a parameter table documenting the default as `True`. `docs/mesh.md` did too: line 11 said every `Robot()` auto-joins, while line 152 said `Robot(name, mode="sim", mesh=True)` "is the normal path".

**Parameter table.** `README.md:440` said `| mesh | bool | True |`. That compounds the other two: the reader's only documented lever (`mesh=False`) is a no-op for the state they are actually in, while the lever that reaches the documented state was never shown anywhere on the page.

The claims are old rather than wrong-when-written - `docs/mesh.md`'s dates from #160 and README's opening line from #371, both before #630.

## Change

Correct the three pages, reusing the wording this contract already carries elsewhere in the tree (the environment-variable rows, and the factory reference page's `mesh` row and Mesh section):

- snippets pass `mesh=True` where they go on to use the mesh;
- prose states the opt-in and names the spellings that reach it;
- README's `mesh` row documents `None` and what `None` consults;
- `docs/mesh.md`'s "Disable" table becomes "Enable and disable" and lists the opt-in first.

Where a `Simulation` is built directly, the pages name assigning a started client rather than `mesh=True`, because that argument is not a boolean opt-in:

```
Simulation(mesh=True) -> TypeError: mesh=True is not a mesh client: mesh= takes an
already-started client exposing .stop() (got bool) ... To join a mesh use
Robot(name, mode='sim', mesh=True)
```

`tests/test_docs_mesh_join_is_opt_in.py` grades every `docs/**/*.md` page and `README.md` at the scope each spelling lives in: per fenced block for snippets, per paragraph for prose, per row for tables. Two sibling guards already grade this contract - one for a table keyed on the environment variable, one for the factory reference page - but neither reads any other page and neither grades a snippet. The snippet check is block-scoped because a reader copies one fence, and it grades a `Simulation` construction by the client assignment so it cannot bless a spelling the library refuses. A paragraph that names the enabling spelling is exempt, so a page stays free to write "a bare `Robot()` does not auto-join - pass `mesh=True`".

## Verification

New guard, before and after the pages are corrected:

| | before | after |
|---|---|---|
| `tests/test_docs_mesh_join_is_opt_in.py` | **3 failed**, 23 passed | **26 passed** |

The three failures name their offenders: 11 snippet reads, 4 prose paragraphs, 1 table row. The 23 that pass on both trees are the premises (page/fence/paragraph counts reached, some page really uses `.mesh`), the reality controls (a bare `Robot()` leaves `mesh` `None`; exactly `true`/`1`/`yes` opt in), and planted-regression checks in both directions - including that a paragraph earning the exemption only through a reassuring word about a different knob (`STRANDS_MESH_MULTICAST`) is still reported.

No regressions across every test that reads a documentation page, plus `tests/mesh/` and the repo-wide docstring and string guards (29 files + guards, 4233 collected, identical selection on both trees):

| | failed | passed |
|---|---|---|
| this branch | 21 | 4212 |
| `upstream/main` (with the new test file copied in) | 24 | 4209 |

Branch-only failures: none. The 3-failure difference is exactly the new guard's pre-fix failures, and 45 failing ids are shared and pre-existing (24 are `ModuleNotFoundError: No module named 'awsiot'`).

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`. `mypy` reports the same 19 errors in 8 files on both trees, none in a touched file. No non-ASCII character was added to any page, fence counts unchanged, no table left ragged.

## Artifact

The mesh guide's own snippet, extracted from each tree's `docs/mesh.md` and executed as written. `strands_robots` is identical in both panels; the only difference is the page the snippet is copied from.

Before and after: the mesh guide snippet raises AttributeError on upstream/main and runs on this branch

On the corrected page the block runs to completion, including the `tell()` on its last line, and prints a real peer list.


yinsong1986 (approved):
## Summary

Docs-only correction plus a guard: three pages (`README.md`, `docs/mesh.md`, `docs/security.md`) still described the pre-#630 world where a bare `Robot()` auto-joined the Zenoh mesh. The PR passes `mesh=True` in every snippet that goes on to read `.mesh`, rewrites the four auto-join prose paragraphs to state the opt-in, fixes README's parameter row (`bool | None`, default `None`), and turns `docs/mesh.md`'s "Disable" table into "Enable and disable" with the opt-in rows listed first. The new `tests/test_docs_mesh_join_is_opt_in.py` grades every `docs/**/*.md` page and `README.md` at the scope each spelling lives in (per fence / per paragraph / per table row), with premise checks, reality controls, and planted-regression tests in both directions.

I verified the factual claims against the code at the head SHA: `mesh` defaults to `None` and consults `_mesh_env_opt_in()` (`strands_robots/robot.py:339`), exactly `true`/`1`/`yes` opt in (`robot.py:141`), and `STRANDS_MESH=false` is a hard kill switch that overrides an explicit `mesh=True` (`robot.py:133-136`, `mesh/core.py:3517-3521`) - so the corrected README kill-switch sentence and the new mesh.md table rows are accurate, not just plausible. All 26 tests in the new guard pass on the head tree, `ruff check` is clean on the new file, no non-ASCII character is added anywhere in the diff, and the `Simulation`-vs-`Robot` asymmetry (`Simulation(mesh=True)` raising `TypeError`) is graded by client assignment so the guard cannot bless a spelling the library refuses.

## What's good

- Changelog fragment at `changelog.d/2395-mesh-join-is-opt-in.md` per AGENTS.md step 3, no direct `CHANGELOG.md` edit.
- The guard grades behaviour rather than a copied list: enabling spellings are checked by calling `_mesh_env_opt_in()`, and the table check reads the real default from `inspect.signature(Robot)`.
- Planted-regression tests cover both directions, including the `STRANDS_MESH_MULTICAST` false-exemption case and prose inside fences.
- Scope discipline: docs, one fragment, one test file; no production code touched.

PR #2396, opened 2026-08-17
fix(simulation/mujoco): refuse a non-string for a field the schema publishes as one

cagataycali:
## Problem

`_dispatch_action`'s docstring states the contract for its validation layer:

> * unknown top-level params are rejected with a friendly message,
> * missing required params produce a "requires parameter X" error **(no raw Python `TypeError`)**,
> * vector params have length + numeric dtype checked **before the value reaches numpy / MuJoCo**.

A scalar string field had no such layer. `tool_spec.json` publishes 32 fields as
`"type": "string"`, and a non-string value for any of them was carried into the
method body, which failed wherever it first assumed a string. Driving every
published action against every string field it accepts, with `7` / `["x"]` /
`{"k": 1}` (465 probes), **39 escaped as raw exceptions** across 13
`(action, field)` pairs on 8 actions:

| action | field | what the caller received |
|---|---|---|
| `add_robot` | `data_config` | `AttributeError: 'int' object has no attribute 'lower'` |
| `add_robot` | `urdf_path` | `TypeError: expected str, bytes or os.PathLike object, not int` |
| `load_scene` | `scene_path` | `TypeError: stat: path should be string, bytes, os.PathLike or integer, not list` |
| `replace_scene_mjcf` | `xml` | `TypeError: from_string(): incompatible function arguments` |
| `start_recording` | `repo_id`, `root` | `TypeError: argument of type 'int' is not iterable` |
| `save_state` | `name`, `checkpoint_name` | `TypeError: unhashable type: 'list'` |
| `set_geom_properties` | `geom_name` | `AttributeError: 'int' object has no attribute 'endswith'` |
| `export_xml` | `output_path` | `AttributeError: 'int' object has no attribute 'strip'` |
| `start_cameras_recording` | `name`, `checkpoint_name`, `output_dir` | `AttributeError: 'int' object has no attribute 'strip'` |

None of those names the field the caller sent, and an exception raised past this
boundary is not a worse message but no message at all: the result dict is the
only channel a bad input is reported on, so it takes the caller's `except`
clauses and any open recording with it.

**The silent form is worse.** A number is a perfectly good dict key, so
`save_state` stored it and reported success. Because every call at this boundary
carries the name as a string, nothing could address that checkpoint again:

```
save_state(checkpoint_name=1)     -> success   "State '1' saved ... Checkpoints: [1]"
load_state(checkpoint_name="1")   -> error     "Checkpoint '1' not found. Available: [1]"
```

The refusal lists the checkpoint it denies. That state is unreachable and
undeletable for the rest of the session.

`entity_name_error` already documents this exact harm for the creation sites it
guards -- *"`add_object(7, ...)` registered the key `7` and only then raised
`TypeError` ... `add_robot(7, ...)` did compile, under the int registry key `7` -
which the tool surface, where every name arrives as a JSON string, can never
address"* -- and it is why that guard exists. It just never covered the other 29
published string fields.

## Change

Add the missing layer to the dispatcher, next to the vector check:

- **`published_string_error(value, param, context)`** in `utils.py`, beside
  `entity_name_error` and for the reason that one records: these domains must not
  diverge, and a second copy of the rule would drift from the first.
- **`_PUBLISHED_STRING_PARAMS`** derived from `_TOOL_SPEC_SCHEMA` at import time,
  for the reason `_PUBLISHED_ACTIONS` already is -- *"a second list would be a
  second thing to keep true"*. A field published later is covered without another
  list to maintain. Enforcing it at one funnel rather than at each entry point is
  the reasoning `mj_name_to_id` records for entity lookups: *"a lookup added later
  is safe by construction."*
- The refusal **names the spelling the caller used**. `checkpoint_name` is an
  alias for `name`, and reporting the canonical parameter would send the caller
  looking for a key their payload never contained.
- **`keyframe`'s schema is corrected to `["string", "integer"]`.** The method
  annotates it `str | int`, its docstring says *"name `str` or index `int`"*, and
  both `docs/getting-started/robot-factory.md` and `docs/simulation/world-building.md`
  already document the index form. `tool_spec.json` was the sole outlier, so the
  index form was never discoverable -- and a rule derived from the old schema
  would have refused it.

Only the type is refused. `""` stays accepted (`render(camera_name="")` selects
the free camera, and an empty `instruction` is the documented rollout default), so
emptiness remains the creation site's call. A `str` subclass is accepted. `None`
still means unset. The Python path is unchanged and deliberately wider: calling a
method directly bypasses the dispatcher, so `entity_name_error` is still the guard
there, and a field the schema does not publish (`stop_recording`'s `bucket` /
`run_id`) is not type-enforced -- enforcing a published type on an unpublished
field would enforce a contract nobody was shown.

**Out of scope:** `set_geom_properties(geom_id="abc")` is the one remaining raise
(`TypeError: '<' not supported between 'str' and 'int'`), and it is an `integer`
field. `isinstance(x, str)` has no ambiguity; the numeric domain does -- the
vector layer already had to settle `bool`-is-an-`int`, integral floats, and NumPy
scalars, and answering those for 12 integer plus 19 number fields is a separate
change with real breakage risk. Doing it here would have bundled a contract
decision into a fix that needs none.

## Verification

Same script, same scene, same payloads against both trees; cube position read
from MuJoCo `d.xpos`. On `main` the save silently succeeds, there is nothing in
the reply to act on, and the restore never lands -- the cube is left **0.3842 m**
from the saved pose. With the fix the refusal names `checkpoint_name`, the retry a
caller can actually make succeeds, and the state comes back exactly (**0.0000 m**).

checkpoint round-trip

| | `main` | this branch |
|---|---|---|
| `save_state(checkpoint_name=1)` | `success`, `Checkpoints: [1]` | `error`, names `checkpoint_name` |
| field parsed from the reply | `None` | `'checkpoint_name'` |
| `load_state(checkpoint_name="1")` | `error` `"not found. Available: [1]"` | `success` |
| cube xy after restore | `[0.02, 0.30]` | `[0.26, 0.0]` |
| distance from saved pose | **0.3842 m** | **0.0000 m** |

Renders are `MUJOCO_GL=egl`, identical camera in both rows; mean per-pixel
difference between the initial and post-restore frames is 2.164 on `main` versus
0.2495 here (the residual is the arm settling -- the cube is exact).

**Regression test:** `tests/simulation/mujoco/test_router_refuses_a_nonstring_published_string_field.py`,
**7 failed / 8 passed** on `upstream/main`, **15 passed** here. The headline test
sweeps the published surface and reports every escaping site by name; the 8 that
pass on both trees bound the change -- the integer `keyframe` index, `""`, a `str`
subclass, `None`, unknown-parameter ordering, a string-named round trip, and both
creation-site guards on the Python path.

**Gate:** `ruff check` / `ruff format --check` clean over
`strands_robots tests tests_integ`; `mypy` adds no error (zero in the touched
files). Full `MUJOCO_GL=egl` suite on this branch and on `upstream/main` with the
new file copied in: 31478 collected on both, branch 5 failed / 31473 passed, base
10 failed / 31468 passed -- the difference is exactly the 5 pre-fix failures, and
an ID-level diff shows no branch-only regression (the two `test_zenoh_transport_security`
IDs that rotate between the runs pass 9/9 in isolation on both trees). A targeted
re-run after the final revision: 1 failed / 11661 passed, that one
(`newton/test_multi_camera.py::test_bad_position_shape_rejected`, a message-text
assertion) failing identically on `main`.


yinsong1986 (dismissed):
## Summary

Adds the missing scalar-string layer to `_dispatch_action`'s validation funnel: every field `tool_spec.json` publishes as `"type": "string"` (30 after the `keyframe` correction) is now refused at the dispatcher with a structured error naming the field, the value, and its type, instead of carrying a non-string into the method body where it either escaped as a raw `TypeError`/`AttributeError` or — the `save_state(checkpoint_name=1)` case — silently succeeded into a checkpoint no wire payload could ever address again. The enforced set is derived from the schema at import time (mirroring `_PUBLISHED_ACTIONS`), the refusal reports the alias spelling the caller actually used (`checkpoint_name`, not the canonical `name`), and `keyframe`'s schema is widened to `["string", "integer"]` to match the `str | int` annotation and documented index form before a rule derived from the old schema could refuse it. I independently verified the false-refusal surface: across all 76 (action, published-string-param) dispatch sites, every method annotation is `str`/`str | None` — `keyframe` was the sole wider one and the schema fix handles it, so the new check refuses nothing a dispatched method accepts. `None`-means-unset, empty strings, and `str` subclasses are all deliberately passed through, and the ordering against the unknown-parameter check is pinned.

## What's good

- The guard lives at the single funnel (`_validate_and_build_kwargs`) and its set is derived from the schema, so a field published later is covered with no second list to keep true — the same reasoning `_PUBLISHED_ACTIONS` and `mj_name_to_id` already record.
- The boundary is scoped precisely: type only (emptiness stays the creation site's call via `entity_name_error`), wire path only (direct Python calls keep the wider domain), published fields only (`stop_recording`'s unpublished `bucket`/`run_id` are not retroactively type-enforced).
- The regression test sweeps the published surface rather than pinning the 13 known-bad pairs, and separately pins every boundary of the fix: the integer `keyframe` index, `""`, a `str` subclass, `None`, alias-spelling in the refusal, unknown-parameter ordering, and both Python-path creation-site guards.
- `_validate_and_build_kwargs`'s new `received` parameter defaults to `remapped`, so the existing direct callers in `tests/` are unaffected.
- Changelog fragment present and follows the `### Fixed` fragment convention (AGENTS.md > PR Workflow step 3).

## Verification suggestions

- Spot-check the one behavioral widening: `sim(action="add_robot", name="r", urdf_path=<model with a keyframe>, keyframe=0)` succeeds on this branch and on `main` identically (the method always accepted the index; only the schema and the new guard's derivation needed to agree).
- `python -m pytest tests/simulation/mujoco/test_router_refuses_a_nonstring_published_string_field.py -x` on the merge result if any sibling PR touches `tool_spec.json` or `_FIELD_ALIASES` before this lands — the enforced set is computed from both at import time.

yinsong1986 (approved):
## Summary

Adds the missing scalar-string layer to `_dispatch_action`'s validation funnel: every field `tool_spec.json` publishes as `"type": "string"` is now refused at the dispatcher with a structured error naming the field, the value, and its type, instead of carrying a non-string into the method body where it escaped as a raw `TypeError`/`AttributeError` or - the `save_state(checkpoint_name=1)` case - silently succeeded into a checkpoint no wire payload could address again. The enforced set is derived from the schema at import time (`_published_string_params`, mirroring `_PUBLISHED_ACTIONS`), the refusal reports the alias spelling the caller actually sent via the new `received` parameter (which defaults to `remapped`, leaving existing direct callers unaffected), and `keyframe` is widened to `["string", "integer"]` to match the `str | int` its method, docstring and docs already carry - which also correctly excludes it from the derived set. I verified the head independently: the full 17-test regression file passes on `8921f735` under `MUJOCO_GL=egl`, the schema publishes 31 string fields all of whose dispatch targets accept `str`, the only other `tool_spec.json` type consumer (`harness_memory.get_valid_actions`) reads the action enum alone so the type-array form breaks nothing, and the one residual escape route (the pre-validation video fold popping `output_path`/`camera_name` on the four rollout actions) predates this PR byte-for-byte on `main`, so it is a pre-existing gap rather than a regression. Boundary scoping is deliberate and pinned: `""`, `str` subclasses, `None`-means-unset, unpublished fields (`stop_recording`'s `bucket`/`run_id`), and the Python path's `entity_name_error` are all unchanged. Changelog fragment present, `Args:` entries on every new parameter, no non-ASCII or host paths in the new test file. The only delta since the dismissed review is a test parametrization; no new findings.

PR #2397, opened 2026-08-17
fix(policies/lerobot_local): keep a best-effort Hub read out of the load path

cagataycali:
## Problem

Both Hub reads in the LeRobot-local processor bridge import `HfHubHTTPError` inside the same `try` whose handler names it:

```python
try:
    from huggingface_hub import hf_hub_download
    from huggingface_hub.errors import HfHubHTTPError

    path = hf_hub_download(...)
    return load_file(path)
except ImportError:
    return None
except (HfHubHTTPError, OSError, ValueError, SafetensorError) as exc:
    return None
```

Python evaluates an `except` clause's exception expression when the body raises, not when the `try` is entered. So that second handler is only evaluatable if the import it depends on already succeeded. An `ImportError` lands in the first handler and never reaches it, which is why the shape looks correct — but any *other* import failure (a partially installed distribution raising `OSError`, a module whose definition-time work raises `TypeError`) skips the `ImportError` handler and raises `UnboundLocalError` while evaluating the second one.

Two things go wrong at once. The original exception is replaced by one naming an unrelated local, and the handler written for that exception is skipped — `OSError` is in its own tuple.

Both helpers document the contract this breaks. `_load_checkpoint_state_dict`: *"stays best-effort and never raises into the load path"*. `_pipeline_step_keys`: *"Never raises into the load path."* Neither held.

The severity comes from the caller. `LerobotLocalPolicy._load_processor_bridge` wraps `ProcessorBridge.from_pretrained` in `except (FileNotFoundError, ValueError, ImportError)` and treats a missing hub as benign ("models work without it via the raw obs/action flow"). `UnboundLocalError` is a `NameError`, so it is not caught: the same root condition either degrades to passthrough or aborts the policy load outright, decided only by which exception type the hub import happened to raise.

Measured against `main`, driving both helpers with the hub made unimportable:

| import raises | `_load_checkpoint_state_dict` | `_pipeline_step_keys` |
|---|---|---|
| `ImportError` | returned `None` (handled) | returned `None` (handled) |
| `OSError` | `UnboundLocalError`, escapes the load path | `UnboundLocalError`, escapes |
| `TypeError` | `UnboundLocalError`, escapes the load path | `UnboundLocalError`, escapes |
| `RuntimeError` | `UnboundLocalError`, escapes the load path | `UnboundLocalError`, escapes |

with `__context__` holding the real `OSError`, and `issubclass(UnboundLocalError, (FileNotFoundError, ValueError, ImportError))` false.

## Change

The error class is imported in its own `try`, so it is bound before any handler naming it can be evaluated, and the call keeps its own handler unchanged. That import guard degrades on any failure rather than only `ImportError`, matching the guard already in `_load_in_model_normalization_fallback` ten lines below, which catches `Exception` with the same rationale recorded in a comment: an import in this path can fail at module-definition time without raising `ImportError`, and reconstruction is impossible either way, so it must degrade instead of crashing an ACT/diffusion load.

After the change all eight cells above return `None`. Both docstrings now enumerate the case, and both promises hold.

## The rule, not just the two sites

`tests/test_except_clauses_do_not_name_their_own_imports.py` refuses the shape across the package: no `except` clause may name a symbol its own `try` body imports.

It is worth a deterministic gate because nothing else refused it. `ruff` has no rule for it. `mypy` reports it under `possibly-undefined`, which is off by default and not enabled here. Code scanning has a rule for it, `py/uninitialized-local-variable`, and it did not report either of these two sites — all five of its open alerts are in test code, on the same shape. So this reached shipped source with no gate naming it.

Scope is imports only. A name bound by ordinary assignment in a `try` body and named in its handler is a different question with different answers; an import is the case where the binding statement and the failing statement are the same statement.

## Verification

Pre-fix, with the new test file copied onto `main`: **9 failed, 6 passed**. Post-fix: **15 passed**.

The six that pass on both trees are the controls — the rule's own non-vacuity (a planted violation is reported; the two-`try` remedy and an unrelated handler class are accepted), an absent hub still degrading to `None` (the path that always worked, unchanged), a failing download still handled, and a real `HfHubHTTPError` from the download still handled. The last two fail if splitting the import were to drop the call's handler.

Blast radius (`tests/policies`, the processor-bridge and partial-construction suites, and the repo-wide docstring / string / dependency / changelog guards), same selection on both trees:

| | failed | passed |
|---|---|---|
| this branch | 0 | 4697 |
| `main` + the new test file | 9 | 4688 |

`4688 + 9 = 4697`: the only difference is the nine pre-fix failures, no regressions. `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` reports the same 19 pre-existing errors on both trees and none in the changed files.

## Artifact

One capture script, run unmodified under each tree — only `strands_robots` differs, and nothing is stubbed but the Hub import. Left: `main`, two of eight cases degrade as documented. Right: this change, eight of eight.

hub read degradation

### DIFF
--- a/changelog.d/2397-hub-read-degrades-when-hub-unimportable.md
+++ b/changelog.d/2397-hub-read-degrades-when-hub-unimportable.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **policies/lerobot_local**: a best-effort Hub read in the processor bridge no longer aborts the policy load when `huggingface_hub` is installed but unimportable. Both reads imported `HfHubHTTPError` inside the same `try` whose handler names it, so any non-`ImportError` import failure raised `UnboundLocalError` while evaluating that handler - masking the real error and skipping the `OSError`/`ValueError` handler written for it. The error class is now imported in its own `try`, and a new package-wide guard refuses the shape.
--- a/strands_robots/policies/lerobot_local/processor.py
+++ b/strands_robots/policies/lerobot_local/processor.py
@@ -32,7 +32,8 @@ def _load_checkpoint_state_dict(pretrained_name_or_path: str, revision: str | No
 
     Resolves a local directory first, then the HF Hub (cached). Returns ``None``
     when no single-file ``model.safetensors`` is present (e.g. a sharded VLA
-    checkpoint) or safetensors is unavailable -- the OLD-FORMAT checkpoints this
+    checkpoint), or when safetensors or ``huggingface_hub`` is unavailable -- an
+    installed-but-unimportable hub included. The OLD-FORMAT checkpoints this
     supports (ACT / diffusion / tdmpc / vqbet zoo) are all single-file, so this
     stays best-effort and never raises into the load path.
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Fixes a real crash-class bug in the two best-effort Hub reads in `strands_robots/policies/lerobot_local/processor.py`: `HfHubHTTPError` was imported inside the same `try` whose second handler names it, so any non-`ImportError` import failure (e.g. a broken install raising `OSError`) raised `UnboundLocalError` while evaluating that handler — masking the real error, skipping the `OSError`/`ValueError` handler written for it, and escaping the load path's `(FileNotFoundError, ValueError, ImportError)` catch since `UnboundLocalError` is a `NameError`. The fix moves the error-class import into its own guarded `try` and keeps the call's handler unchanged; I verified the mechanism independently by running the PR's own AST rule against upstream `main` (both sites flagged) and against this head (clean). The broad `except Exception` on the import guard is justified: the contract is explicitly best-effort with a documented caller-side passthrough, it matches the existing `_load_in_model_normalization_fallback` precedent in the same file, and the rationale is recorded inline with a `noqa`.

## What's good

- The package-wide AST guard (`tests/test_except_clauses_do_not_name_their_own_imports.py`) closes the whole class, not just the two sites, with non-vacuity controls (planted violation reported, remedied form accepted, unrelated handler accepted) and a scan-scope floor so an empty scan fails loudly.
- End-to-end regression tests drive both sites with a meta-path finder raising `OSError`/`TypeError`/`RuntimeError`/`ValueError`, plus controls pinning that the split did not drop the call's `HfHubHTTPError`/`OSError` handling — per AGENTS.md's "pin every reviewed fix" rule.
- Changelog fragment under `changelog.d/` per step 3; docstrings updated to enumerate the new degradation case; scope disciplined to the fix.

PR #2398, opened 2026-08-17
test: bind an optional-dependency skip's names on every path to their use

### LINKED ISSUE #2359: test: clear six pre-existing py/uninitialized-local-variable alerts (try + pytest.skip shape)
## Summary

Six `py/uninitialized-local-variable` alerts are open on `main`, all in test code, all the same underlying shape: a name is bound inside a `try` whose handler only calls `pytest.skip`. `pytest.skip` raises, so the name is bound on every path that reaches its use - but that is a property of pytest's control flow, not of the function, so a liveness analysis scoped to one body reports the use as potentially uninitialized.

| alert | path | line |
|---|---|---|
| 3 | `tests/simulation/mujoco/test_concurrency.py` | 359 |
| 4 | `tests/benchmarks/libero/test_libero_adapter.py` | 3370 |
| 5, 6 | `tests_integ/benchmarks/libero/test_upstream_state_parity.py` | 725, 726 |
| 7, 8 | `tests_integ/benchmarks/libero/test_upstream_state_parity.py` | 732 |

These are the lowest-numbered alerts in the repo, so they predate the current security baseline. An AST sweep of `tests/` finds exactly three source sites of the shape, which is the full extent of it - this is a bounded cleanup, not an open-ended one.

## Why this is not one commit

The three sites need three different fixes, which is why I am filing this rather than widening #2356 (where the same shape was introduced and fixed):

**1. `test_concurrency.py:352`** - the `try` wraps an import *and* a constructor call, catching `(ImportError, RuntimeError)`:

```python
try:
    from lerobot.datasets.lerobot_dataset import LeRobotDataset
    ds = LeRobotDataset(repo_id="local/rt-test", root=ds_root)
except (ImportError, RuntimeError):
    pytest.skip("lerobot dataset API not available (torchcodec/ffmpeg missing)")
```

`pytest.importorskip` cannot express this: the `RuntimeError` comes from `LeRobotDataset.__init__` (torchcodec/ffmpeg missing at decode time), not from the import. Splitting the two concerns changes which failures skip and which fail, so it needs a deliberate decision rather than a mechanical rewrite.

**2. `test_libero_adapter.py:3366`** - `from libero.libero.envs.robots.mounted_panda import MountedPanda`, guarded because a future rename would move it. This one converts cleanly to `importorskip` on the module plus a `getattr`, and the file already calls `importorskip` twice immediately above.

**3. `test_upstream_state_parity.py:723`** - a bare `except Exception` around `NvidiaGr00tPolicy(...)` model loading. Not an import guard at all; the skip is for environmental model-load failure. Needs restructuring (or an explicit pre-binding), and it is under `tests_integ/`, which does not run in the default gate.

## Suggested resolution

One commit per site, in that order of increasing risk - or just (2), which is the mechanical one, if the appetite is small. The prior art is in #2356:

```python
module = pytest.importorskip(cfg["module"], reason=f"{provider} needs an optional dependency")
cls = getattr(module, cfg["class"])
```

`pytest.importorskip` is already the suite's idiom for gating on optional dependencies (890 call sites, 46 passing `reason=`), so where it fits, it both clears the alert and matches the surrounding style.

## Non-goals

Suppression comments and alert dismissal. The code is reachable-safe today, so dismissing as a false positive would be defensible for each one individually - but the shape is easy to reintroduce (it was, in #2356), and the `importorskip` form is shorter than the `try/except` it replaces wherever it applies.

cagataycali:
## Problem

Three test sites bind a name inside a `try` whose handler calls `pytest.skip`, then read it after the block:

```python
try:
    from libero.libero.envs.robots.mounted_panda import MountedPanda
except ImportError:
    pytest.skip(f"libero {libero.__version__} does not expose MountedPanda")
expected_home_qpos = np.asarray(MountedPanda().init_qpos, dtype=np.float64)
```

This runs correctly, because `pytest.skip` raises. But the binding's liveness is then a property of pytest's control flow rather than of the enclosing function, so `py/uninitialized-local-variable` reports the use. Five alerts, one per name:

| alert | site | name |
|---|---|---|
| 4 | `tests/benchmarks/libero/test_libero_adapter.py:3366` | `MountedPanda` |
| 5, 6, 7 | `tests_integ/benchmarks/libero/test_upstream_state_parity.py:690` | `NvidiaGr00tPolicy`, `EmbodimentTag`, `Gr00tSimPolicyWrapper` |
| 8 | `tests_integ/benchmarks/libero/test_upstream_state_parity.py:724` | `nvidia_policy` |

Fixes #2359. These are the lowest-numbered open alerts in the repository, and `github-advanced-security` opens a review thread per alert, so they surface as unresolved threads on whichever pull request happens to touch one of these files next.

## Change

Each site moves to the idiom the suite already uses at ~890 call sites.

**A missing module** is `importorskip`. **A missing attribute** on a module that does import needs its own skip, because `importorskip` does not cover it - and that is exactly what the libero guard exists for, per its own comment ("a future rename would move it"). A bare attribute read would convert that rename from a skip into an `AttributeError`, so `getattr(module, name, None)` plus an explicit skip preserves the behaviour:

```python
mounted_panda = pytest.importorskip(
    "libero.libero.envs.robots.mounted_panda",
    reason=f"libero {libero_version} does not expose the mounted_panda module",
)
MountedPanda = getattr(mounted_panda, "MountedPanda", None)
if MountedPanda is None:
    pytest.skip(f"libero {libero_version} does not expose MountedPanda")
```

**A value that has to be built** gets a helper that *returns* it, so the caller's binding is unconditional. `_attribute_or_skip` and `_load_upstream_policy_or_skip` follow `_read_first_frame_or_skip` in `tests/simulation/mujoco/test_concurrency.py`.

Those helpers **raise** `pytest.skip.Exception` rather than calling `pytest.skip`. The first revision of this PR called it, and code scanning caught that this only moves the unanalyzable branch one level down: a helper that both returns a value and calls `pytest.skip` mixes an explicit return with an implicit one, which is `py/mixed-returns` (alert 933 on the first push). `pytest.skip.Exception` is what `pytest.skip` raises, and is already used at three sites here, so the terminating path becomes a `raise` that needs no annotation to read. `__tracebackhide__` keeps the skip reported at the caller's line, as `pytest.skip` does; verified against a replica of both helpers that the reason text and the reported location are unchanged and the success paths still return their value.

That exchange is worth recording because it is the same lesson at two levels: the fix is not "use a helper", it is "leave no branch whose termination a reader has to infer".

### Equivalence of the rewrite

The replaced guard caught `ImportError`, which covers a missing package, a missing submodule and a missing attribute. Driving both forms over all four cases:

| case | replaced form | this PR |
|---|---|---|
| package + submodule + attribute | returns the class | returns the class |
| attribute renamed away | `Skipped: does not expose MountedPanda` | `Skipped: does not expose MountedPanda` |
| submodule absent | `Skipped: does not expose MountedPanda` | `Skipped: does not expose the mounted_panda module` |
| package absent | `Skipped: does not expose MountedPanda` | `Skipped: does not expose the mounted_panda module` |

Same skip-or-succeed outcome in every case, and a reason that names what was actually missing in the two where the module itself is gone.

## The regression check

`tests/test_optional_dependency_skips_bind_their_names.py` sweeps `tests/` and `tests_integ/` for the shape. It is here rather than left to code scanning because the shape has already been reintroduced once after being fixed, and because a check that runs in 7 seconds locally is a better place to learn this than an alert that arrives after the merge.

Scope is the reported shape and nothing wider. A handler ending in `return`, `raise` or `pytest.fail` leaves a name conditionally bound in exactly the same way, and none of those is reported - measured against `main`, the tree carries 20 `return`, 1 `raise` and 4 `pytest.fail` sites of the same structure with zero alerts between them. Those forms are also idiomatic here: a `pytest.fail` handler asserts that the `try` body *succeeded*, rather than gating on an environment. Widening the check to them would trade a bounded rule for a large mechanical rewrite that no gate asks for, so the module docstring records the boundary and the measurement behind it.

Beyond the sweep, the file carries six cases that pass on both trees and hold that boundary: the detector fires on a planted import guard and on a planted built value, and stays silent for `fail` / `return` / `raise` handlers, for a name pre-bound before the `try` (a valid fix, in use in the tree), and for a guard that binds nothing.

## Scope note: the broad guard on the model load stays

Alert 8's `try` wraps a checkpoint load under `except Exception`. This PR binds the name and leaves the guard's breadth alone. A checkpoint load reaches torch, the transformers stack and the Hub, any of which can fail for reasons about the host rather than about the code under test; narrowing the tuple without being able to measure what that loader raises would turn an unanticipated environmental failure into a red integration run that says nothing about strands-robots. The existing `# noqa: BLE001 - model load failure is environmental` already records that reasoning, and it is now on a helper whose docstring states it. I had suggested narrowing it on the issue; having read the load path, that suggestion was wrong for this guard and I have not carried it out here.

## Verification

Selection: `tests/benchmarks/libero/` plus every check that scans the test tree plus the repo-wide guards (44 paths, 2364 tests). Run identically on this branch and on a `upstream/main` worktree with the new file copied in, so the base run shows the pre-fix verdict:

|---|---|---|---|
| `upstream/main` + new file | **1** | 2283 | 80 |
| this branch | **0** | 2284 | 80 |

`2283 + 1 = 2284`: no other test changed verdict. The one base failure is the sweep, naming all five bindings:

```
tests/benchmarks/libero/test_libero_adapter.py:3366 binds 'MountedPanda'
tests_integ/benchmarks/libero/test_upstream_state_parity.py:690 binds 'EmbodimentTag'
tests_integ/benchmarks/libero/test_upstream_state_parity.py:690 binds 'Gr00tSimPolicyWrapper'
tests_integ/benchmarks/libero/test_upstream_state_parity.py:690 binds 'NvidiaGr00tPolicy'
tests_integ/benchmarks/libero/test_upstream_state_parity.py:724 binds 'nvidia_policy'
```

one per open alert, in the same three files. Both patched files collect and skip cleanly where their dependencies are absent (`tests/benchmarks/libero/test_libero_adapter.py`: 206 passed, 28 skipped).

`ruff check` / `ruff format --check` / `mypy` over `strands_robots tests tests_integ` are clean, with mypy at the same 19 pre-existing errors as `main` and none in the touched files. mypy is what rejected the first draft of the `gr00t` rewrite: an `absent`-list guard left the three names as `Any | None` at their use, which is the same "the checker cannot see the skip" problem one level up. `_attribute_or_skip` returns the attribute, so the narrowing is local and legible.

Code scanning on the branch: `Analyze (Python)` passes, and the five `py/uninitialized-local-variable` alerts this PR targets are the only ones it touches.

No visual artifact: this changes test-tree control flow only, and no policy, simulation, rendering, recording or asset behaviour.

### Round 1 - `514b9e9`: resolve the libero version once, outside the skip reason argument

Review feedback: the rewrite's `reason=f"libero {libero.__version__} ..."` reads `__version__` **eagerly**, so it runs on every host where `libero` imports rather than only on the skipping path - and the released distribution has no such attribute. That turned this test's happy path into an `AttributeError`.

Verified the premise against the sdist rather than from the report: `libero-0.1.1`'s top-level `libero/__init__.py` is **1 byte** (a bare newline), and `grep -rn __version__` over the entire distribution returns nothing - so the attribute is absent on every path, not just unset at the top level. 0.1.1 is also the only release satisfying the `benchmark-libero` extra's `libero>=0.1.0,<1.0.0`, so there is no resolvable version where the read succeeds.

Driving both forms against a libero-shaped stub (empty `__init__`, no `__version__`), plus a module that does publish one:

| form | outcome |
|---|---|
| pre-fix, libero importable | `AttributeError: module 'libero' has no attribute '__version__'` |
| post-fix, libero importable | `Skipped: libero unknown does not expose the mounted_panda module` |
| post-fix, `__version__` present | reports it verbatim (`9.9.9`) |

The failure was structurally invisible to CI: `importorskip("libero")` skips first wherever libero is absent, which is every host except the benchmark boxes this test exists for. So it could not be deferred to a follow-up.

**Fix.** One defaulting read after the `importorskip`, reused in both reasons:

```python
libero_version = getattr(libero, "__version__", "unknown")
```

Both sites move together: the `MountedPanda is None` reason carried the same read on its own rename branch. A libero that does publish `__version__` still reports it, so the diagnostic improves rather than degrading to a constant.

**Introduced, not inherited - but `main` is not clean either.** The replaced code only touched `__version__` inside the never-taken `except ImportError` handler, so the regression is this PR's. That handler nonetheless catches a missing *submodule* as well as a missing package, so on a libero that imports without `mounted_panda`, `main` raises the same `AttributeError` while building its skip reason. Unreachable today (0.1.1 does ship the submodule), which is why it never fired. The defaulting read closes both.

**Not gated, deliberately.** The sweep this PR adds is scoped to the name-binding shape; eager evaluation of a *call argument* is a different shape, and the module docstring already records that boundary and the measurement behind it. Widening it here would trade a bounded rule for an unbounded one mid-review. Offered as its own change if a gate is preferred to a review.

**Verification.** `tests/test_optional_dependency_skips_bind_their_names.py`: **8 passed** (its repo-wide scan confirms the edit reintroduces nothing). `tests/benchmarks/libero/test_libero_adapter.py`: **172 passed, 62 skipped**. `ruff check` / `ruff format --check` clean on the touched file. Test-only change; no source file differs.

## Update: the skip reason the guard is handed

Review found a crash this PR's own rewrite introduced, and it is worth stating because it is the same lesson one level further out.

`reason=` is an argument, so its f-string is built *before* the call receives it. Moving the version read into `pytest.importorskip(..., reason=f"libero {libero.__version__} ...")` therefore moved it off the never-taken rename branch and onto the path every run takes. The distribution the `benchmark-libero` extra resolves to has no `__version__` at all - 0.1.1 is the only release, its top-level `libero/__init__.py` is one byte, and `grep -rn '__version__'` over the whole sdist returns nothing - so the rewrite turned a passing test into an `AttributeError` on precisely the hosts that have the dependency, while CI could not see it: a host without libero skips at the first `importorskip`. The three forms against a libero built from that byte-identical `__init__.py`, each reaching the happy path:

| form | outcome |
|---|---|
| `main` (read inside the never-taken handler) | PASSED |
| this branch before the fix | **FAILED - `AttributeError: module 'libero' has no attribute '__version__'`** |
| version resolved into a local first | PASSED |

The version is now resolved once, before either reason is built, by `_installed_version`. It prefers installed distribution metadata and falls back to a module attribute, which is the order `strands_robots.doctor._resolve_version` documents for this exact case ("reading `module.__version__` yields a useless placeholder... the authoritative version lives in the installed distribution metadata"). That matters here rather than being tidiness: a defaulting attribute read cannot raise, but it reports `unknown` for *every* libero that exists, so the version the reason exists to carry would never be in it. Metadata reports `0.1.1`, and still degrades to `unknown` off a source tree that was never installed.

`_installed_version` is pinned behaviourally, because a check on the source shape cannot see whether the helper itself is safe: regressing it to a bare `module.__version__` read fails those tests with the same `AttributeError`.

### The sweep gains the sibling rule

That no gate could catch this is what makes a source check the only option, so the sweep now also holds: **a guard's reason must not be the first thing to read an attribute of the module the guard gates on.** An attribute already read on the way to the diagnostic is proven rather than assumed - `test_numba_coverage_clash_remedy.py` parses `coverage.__version__` to decide whether to skip and then quotes it, which cannot introduce a new failure - so only a read with no such predecessor is reported. Measured across `tests/` + `tests_integ/` (1094 files):

| tree | offenders |
|---|---|
| this branch before the fix | 2 - `test_libero_adapter.py:3373`, `:3377` |
| `main` | 1 - `test_libero_adapter.py:3369`, latent: the read sits on the rename branch |
| here | 0 |

The `main` entry is a real pre-existing latent read that this branch removes, and on `main` both of the file's checks now fire on the same site (`:3366 binds`, `:3369 reads`) - which is why the two rules live in one module: they are two obligations of one construct.

Verification for this iteration, over `tests/benchmarks/libero/` plus every check that scans the test tree plus the repo-wide guards (42 paths), both trees at current `main` and the new check file copied into the base run so it shows the pre-fix verdict:

|---|---|---|---|
| `main` + the new check | **2** | 2263 | 79 |
| this branch | **0** | 2268 | 79 |

`2263 + 2 + 3 = 2268`, the 3 being the new `_installed_version` tests, so no other test changed verdict. `ruff check` / `ruff format --check` clean; `mypy` at 14 errors on both trees, none in the touched files and nothing branch-only.


yinsong1986 (dismissed):
## Summary

Rewrites three test sites that bound a name inside a `try` whose handler calls `pytest.skip` and then read it after the block, clearing the five `py/uninitialized-local-variable` alerts tracked in #2359. Missing modules move to `pytest.importorskip`, a missing attribute keeps its own explicit skip (so an upstream rename stays a skip rather than becoming an `AttributeError`), and the checkpoint load moves into a `_load_upstream_policy_or_skip` helper that returns the value so the caller's binding is unconditional. A new AST sweep (`tests/test_optional_dependency_skips_bind_their_names.py`) pins the shape across `tests/` and `tests_integ/`, with non-vacuity floors and six self-tests covering both fire and no-fire cases.

## What's good

- Verified locally against the head SHA: the sweep passes on the branch, and the same detector run against `upstream/main`'s versions of the two patched files reports exactly the five `(line, name)` pairs the PR description claims — one per open alert.
- Skip-or-succeed equivalence is preserved: the old guards caught `ImportError` only, and `importorskip` plus the `getattr`-and-skip attribute read covers the same four cases with more precise skip reasons.
- The deliberately broad `except Exception` on the checkpoint load is left in place with its `# noqa: BLE001` rationale now stated in the helper's docstring, rather than narrowed speculatively — consistent with the environmental-failure reasoning.
- The sweep's scope boundary (skip handlers only; `return`/`raise`/`pytest.fail` handlers out of scope) is documented in the module docstring with the measurement behind it, and pinned by parametrized no-fire tests.
- Changelog fragment follows the `changelog.d/` convention; `Fixes #2359` is in the body, not the title; every helper parameter has its own `Args:` entry; no host paths or non-ASCII in the new files.

tests/benchmarks/libero/test_libero_adapter.py
```diff
@@ -3361,11 +3361,19 @@ def test_initial_joint_anchored_to_libero_home_pose_not_construction_state(self,
...
+        # Resolve the canonical LIBERO Panda home pose. Skip if this libero
+        # version does not expose MountedPanda (e.g. a future rename). The
+        # module and the attribute are two separate skips because
+        # ``importorskip`` covers only the module: reading the attribute with
+        # ``getattr`` and skipping on ``None`` keeps a rename a skip rather
+        # than turning it into an ``AttributeError``, and it binds the name on
+        # every path that reaches the use below.
+        mounted_panda = pytest.importorskip(
+            "libero.libero.envs.robots.mounted_panda",
+            reason=f"libero {libero.__version__} does not expose the mounted_panda module",
```
yinsong1986: [MUST FIX] `reason=f"libero {libero.__version__} ..."` evaluates `libero.__version__` eagerly - f-string arguments are computed before `importorskip` is called, on every run where `libero` imports - and the `libero` distribution the `benchmark-libero` extra resolves to (`libero>=0.1.0,<1.0.0`; only 0.1.1 exists on PyPI) ships an empty top-level `libero/__init__.py` with no `__version__`. Verified against that exact package: `import libero; libero.__version__` raises `AttributeError: module 'libero' has no attribute '__version__'`.

So this rewrite converts the test's *happy path* into an `AttributeError` on any host where libero is actually installed. The replaced code only touched `__version__` inside the never-taken rename branch; this line touches it unconditionally. It is a crash bug on the changed path that no gate can catch: CI has no libero, so `pytest.importorskip("libero")` at line 3361 skips before reaching this line, and the failure surfaces only on the benchmark/GPU boxes this test exists for - which is why it can't be deferred to a follow-up. (It also inverts the PR's own goal: a skip-preserving rewrite that turns a passing test into an error.)

Fix: compute the version defensively once and reuse it in both skip messages, e.g. `libero_version = getattr(libero, "__version__", "unknown")` after the `importorskip("libero")`, then interpolate `libero_version` here and in the `MountedPanda is None` skip at line 3377 (which carries the same latent read on its rename branch), or drop the version from the reason entirely.

yinsong1986 (commented):
## Summary

Rewrites the three test sites that bound a name inside a `try` whose handler calls `pytest.skip` and read it after the block, clearing the five `py/uninitialized-local-variable` alerts tracked in #2359: missing modules move to `pytest.importorskip`, the missing-attribute case keeps its own `getattr`-plus-skip so an upstream rename stays a skip, and the checkpoint load moves into `_load_upstream_policy_or_skip`, which returns the value so the caller's binding is unconditional. Since the prior (dismissed) review, the helpers now raise `pytest.skip.Exception` instead of calling `pytest.skip`, resolving the `py/mixed-returns` finding code scanning raised on the previous head. The new AST sweep is sound: run against `main`'s versions of the two patched files it reports exactly the five `(line, name)` pairs the description claims, it reports zero on the head versions and on live `main`'s `tests/`+`tests_integ/` trees, its non-vacuity floors hold (1000+ files, well over 200 `try` statements), and its six self-tests pin both the fire and no-fire cases. `pytest.importorskip`'s `reason` kwarg exists at the declared pytest floor (6.0), so no compat issue there. One must-fix is flagged inline: the adapter-test rewrite evaluates `libero.__version__` eagerly in the `importorskip` reason, and the `libero` package the `benchmark-libero` extra resolves to does not define `__version__`, so the test now errors on exactly the hosts that can run it - and CI cannot see it because CI skips before that line.

## What's good

- The detector's scope boundary (skip handlers only; `return`/`raise`/`pytest.fail` handlers out of scope) is documented with the measurement behind it and pinned by parametrized no-fire tests, including the pre-binding escape hatch.
- The deliberately broad `except Exception` on the checkpoint load is kept with its rationale promoted into the helper's docstring rather than narrowed speculatively.
- Changelog fragment follows the `changelog.d/` convention; `Fixes #2359` is in the body, not the title; every helper parameter has its own `Args:` entry; no host paths or non-ASCII in the new code.

## Verification suggestions

- On any box with the `benchmark-libero` extra installed, `python -c "import libero; libero.__version__"` reproduces the inline finding directly (`AttributeError`), and `pytest tests/benchmarks/libero/test_libero_adapter.py -k initial_joint_anchored` shows the changed test erroring rather than running.

tests/benchmarks/libero/test_libero_adapter.py
cagataycali: Confirmed and fixed in `514b9e9`.

Verified the premise independently against the sdist rather than taking it on report: `libero-0.1.1`'s top-level `libero/__init__.py` is **1 byte** (a bare newline), and `grep -rn __version__` over the whole distribution returns nothing - so there is no `__version__` on any path, not merely a missing one at the top level. And the mechanism is as described: `reason=` is an argument, so its f-string is built before the `importorskip` call it is passed to, on every run where `libero` imports.

Driving both forms against a libero-shaped stub (empty `__init__`, no `__version__`):

| form | outcome |
|---|---|
| pre-fix | `AttributeError: module 'libero' has no attribute '__version__'` |
| post-fix | `Skipped: libero unknown does not expose the mounted_panda module` |

So this was a crash on the happy path, and one CI structurally cannot see - `importorskip("libero")` skips first wherever libero is absent, which is everywhere except the benchmark boxes this test exists for. Agreed it can't be deferred.

Took the first suggested fix: one defaulting read after the `importorskip`, reused in both reasons. Both sites move together, since the `MountedPanda is None` reason carried the same read on its own rename branch.

One thing worth recording for the record, because it bears on the "introduced vs inherited" question: `main` is not entirely clean here either. Its single read sits in an `except ImportError` handler that catches a missing *submodule* as well as a missing package - so on a libero that imports but has no `mounted_panda`, `main` raises the same `AttributeError` while building its skip reason. That path is unreachable today (0.1.1 does ship the submodule), which is why it never fired. The regression this PR introduced is that the read moved onto the path that is always taken; the defaulting read closes both.

I did not widen the PR's sweep to catch this class - eager evaluation of a call argument is a different shape from the name-binding one the sweep is scoped to, and the module docstring already records that boundary deliberately. Happy to add a detector for it as its own change if you'd rather it were gated than reviewed.

cagataycali: Confirmed, and thank you - this was a crash on the changed path, exactly as described.

## The finding reproduces

Against the real distribution rather than the docs: `libero` has **one** release on PyPI (0.1.1), so `libero>=0.1.0,<1.0.0` can resolve nothing else, and in that sdist `libero/__init__.py` is **one byte** (a newline). `grep -rn '__version__'` over the whole distribution returns nothing at all.

Running the three guard forms verbatim against a `libero` whose top-level `__init__.py` is that byte-identical file, with the submodule importable so each test reaches the happy path:

| form | outcome |
|---|---|
| `main` (version read inside the never-taken handler) | PASSED |
| this branch at `c2f64ca4` | **FAILED - `AttributeError: module 'libero' has no attribute '__version__'`** |
| version resolved into a local first | PASSED |

So the rewrite moved the read from a branch that never runs into an argument that always does, and the ordering point is right too: `importorskip("libero")` at the top of the test skips first on a host without libero, so CI never reaches the line.

## What landed

The literal suggestion is in, plus three things it did not cover.

**The assignment landed twice.** Once before `import mujoco` and once above the guard; one was dead. Removed.

**`getattr(libero, "__version__", "unknown")` cannot raise, but it reports `"unknown"` for every libero that exists.** 0.1.1 is the only resolvable release and it defines no `__version__`, so the version the reason exists to carry would never be in it - the message degrades to `libero unknown does not expose MountedPanda` on every real host. The distribution metadata *does* record `0.1.1`. `_installed_version` reads metadata first and falls back to the module attribute, which is the order `strands_robots.doctor._resolve_version` already documents for this exact situation ("reading `module.__version__` yields a useless placeholder... the authoritative version lives in the installed distribution metadata"). Measured both ways:

| host | `getattr(..., "unknown")` | metadata-first |
|---|---|---|
| pip-installed libero | `libero unknown does not expose MountedPanda` | `libero 0.1.1 does not expose MountedPanda` |
| source tree, never installed | `unknown` | `unknown` (degrades, does not raise) |

**Nothing stopped the shape returning.** Your point that no gate can see this is what makes a source check the only option, so the sweep that already pins how these guards bind their names gains the sibling rule: a guard's `reason` must not be the first thing to read an attribute of the module the guard gates on. An attribute already read on the way to the diagnostic is proven rather than assumed, so `test_numba_coverage_clash_remedy.py` - which parses `coverage.__version__` to decide whether to skip and then quotes it - is deliberately not reported. Measured across `tests/` + `tests_integ/` (1094 files):

| tree | offenders |
|---|---|
| `c2f64ca4` (pre-fix) | 2 - `test_libero_adapter.py:3373`, `:3377` |
| `upstream/main` | 1 - `test_libero_adapter.py:3369` (latent: the read sits on the rename branch) |
| here | 0 |

To be precise about its scope: it does **not** fire on the already-pushed `getattr` fix - it would have caught the defect before the push, and it stops a reintroduction. Note it also names a genuinely latent read on `main`, which this branch now removes; on `main` both of the file's checks fire on the same site (`:3366 binds`, `:3369 reads`).

`_installed_version` is pinned behaviourally as well, because a check on the source shape cannot see whether the helper itself is safe: regressing it to a bare `module.__version__` read fails those tests with the same `AttributeError`. The premise assertion uses `"__version__" not in vars(module)` rather than `hasattr`, since the latter is the torch stand-in's discriminator and has one home.

## Verification

Selection: `tests/benchmarks/libero/` plus every check that scans the test tree plus the repo-wide guards (42 paths). Both trees at current `main`, run identically, with the new check file copied into the base run so it shows the pre-fix verdict:

|---|---|---|---|
| `upstream/main` + the new check | **2** | 2263 | 79 |
| this branch | **0** | 2268 | 79 |

`2263 + 2 + 3 = 2268`, the 3 being the new `_installed_version` tests, so no other test changed verdict. The two base failures are one per check, naming the five bindings and the one eager read.

`ruff check` / `ruff format --check` clean; `mypy` over `strands_robots tests tests_integ` at 14 errors on both trees with none in the touched files and nothing branch-only (all 14 are in `examples/isaac_gs/`, outside the lint targets). `main` merged in; no file overlap with it.

yinsong1986 (approved):
## Summary

Test-only change that rewrites the three sites binding a name inside a `try` whose handler calls `pytest.skip` and reading it after the block, clearing the five `py/uninitialized-local-variable` alerts tracked in #2359. Missing modules move to `pytest.importorskip`; the missing-attribute case keeps its own `getattr`-plus-skip so an upstream rename stays a skip rather than becoming an `AttributeError`; the checkpoint load moves into `_load_upstream_policy_or_skip`, which raises `pytest.skip.Exception` so every exit is explicit and preserves the deliberately-broad environmental guard with its `# noqa: BLE001` reasoning now on a documented helper. The new sweep (`tests/test_optional_dependency_skips_bind_their_names.py`) pins both the binding shape and its sibling rule (a skip reason must not be the first read of a gated module's attribute), carries non-vacuity floors so an empty scan cannot pass silently, and includes detector self-tests for both fire and no-fire cases. Verified at head: the sweep passes (15 tests), the modified files compile, and the `TestInstalledVersion` contract tests pass. The concerns from prior review rounds - the eager `libero.__version__` read in the `reason=` f-string and the redundant `libero_version` assignment CodeQL flagged - are both resolved at this head: the version is resolved once into a local via `_installed_version`, which prefers installed distribution metadata per `strands_robots.doctor._resolve_version`'s documented order.

## What's good

- Changelog fragment under `changelog.d/` per AGENTS.md step 3, no direct `CHANGELOG.md` edit.
- The sweep's scope boundary (skip handlers only, not `return`/`raise`/`pytest.fail` handlers) is measured against `main` and recorded in the module docstring rather than asserted.
- Detector under-reports rather than over-reports by design (nested scopes, pre-bound names), and the pre-binding remedy and the resolved-local remedy are both pinned as passing cases, so the check cannot reject its own prescribed fixes.
- No host paths, no non-ASCII in the new files.

PR #2399, opened 2026-08-17
fix(mesh/input): report a failed teleop-event read instead of publishing it as no signal

cagataycali:
## Problem

`InputPublisher` streams a leader's joint action and, alongside it, the operator's control
signals read from the teleoperator's `get_teleop_events()` - `terminate_episode` (the operator
asking to end the episode), `success`, `rerecord_episode`, `is_intervention`. The published
topic schema in the module docstring declares that field nullable, and `null` is the correct
value for a plain leader arm that exposes no event surface at all.

The read was wrapped in a bare `except Exception: pass`, so a failed read published exactly the
same thing. A teleoperator whose event surface stops answering - a keyboard listener thread that
died, a gamepad unplugged mid-session - published `events: None` on every frame while joint
commands kept flowing, and left no trace anywhere: `stats` reported a clean session and nothing
was logged. An operator pressing quit could not tell their signal was being dropped, and neither
could a subscriber watching the stream.

Measured through the real publish loop, 8 frames per case:

**Before** - the three rows that mean different things are byte-identical in every observable:

| case | wire `events` | action delivered | `stats["errors"]` | `stats["event_read_errors"]` | warnings |
|---|---|---|---|---|---|
| read raises (`RuntimeError`, listener died) | `None` | yes | 0 | *(absent)* | 0 |
| read raises (`OSError`, device unplugged) | `None` | yes | 0 | *(absent)* | 0 |
| no event surface (legitimate null) | `None` | yes | 0 | *(absent)* | 0 |
| read OK (operator pressed quit) | `{'terminate_episode': True, ...}` | yes | 0 | *(absent)* | 0 |

**After**:

| case | wire `events` | action delivered | `stats["errors"]` | `stats["event_read_errors"]` | warnings |
|---|---|---|---|---|---|
| read raises (`RuntimeError`, listener died) | `None` | yes | 0 | **8** | 5 |
| read raises (`OSError`, device unplugged) | `None` | yes | 0 | **8** | 5 |
| no event surface (legitimate null) | `None` | yes | 0 | 0 | 0 |
| read OK (operator pressed quit) | `{'terminate_episode': True, ...}` | yes | 0 | 0 | 0 |

```
[mesh] input teleop-event read failed (leader): keyboard listener thread is dead -
publishing events=None, so operator signals are not reaching subscribers
```

teleop-event read observability, before and after

Both panels are the same capture script run against `main` and against this branch, 8 frames
per case through the real `_publish_loop`. On `main` the three highlighted rows - two different
read failures and the legitimate "this leader has no event surface" - are byte-identical in every
observable a caller has. After the change a failed read is the only row that reports anything, and
the legitimate-null row is untouched.

## Change

The read stays best-effort, deliberately. The event surface is secondary to the joint stream, so
its failure must not drop the frame the follower is tracking - letting the exception reach the
loop handler 15 lines below would discard the whole frame, action included, and hold arm motion
hostage to a flaky event surface. What the failure must no longer do is disappear.

- `strands_robots/mesh/input.py`: a failed `get_teleop_events()` now increments a distinct
  `_event_read_error_count`, surfaced as `stats["event_read_errors"]` (and so through
  `get_teleop_status()`), and logs a warning naming the device and the cause. The handler keeps
  its breadth - it is a recovery path - and now carries the comment explaining why.
- The count is kept separate from `errors`, which counts *lost frames*; a failed event read loses
  no frame, so conflating them would misreport the loss. This mirrors how `InputReceiver` already
  breaks its losses down into `drops` / `rejected` / `rate_dropped` / `slew_rejected`.
- The log is rate-limited by `_MAX_LOGGED_LOOP_ERRORS`, a new module constant replacing the
  literal `5` the sibling handler in the same loop already used, so the two agree by construction
  rather than by coincidence. The loop runs at `hz`, so an uncapped log of a persistent fault
  would flood the console at the control rate; the counter stays exact (8 failures, 5 warnings).

## Scope

The wire value is unchanged. `events: null` remains within the documented schema, and widening
that schema to distinguish "not read" from "no event surface" is a protocol change that needs
subscriber coordination, so it is deliberately not part of this. What this fixes is that the
failure is now visible on the publisher, where the operator is.

## Tests

`tests/mesh/test_input_teleop_event_read_is_reported.py` drives the real `_publish_loop`
synchronously (a mesh double ends the loop after N frames, so the frame counts are exact rather
than thread-timing dependent).

Against `main`: **5 failed, 4 passed**. After: **9 passed**.

The 4 that pass on both trees are the controls, and each fails for a different wrong fix:

- a teleoperator with no event surface is still *not* reported as a failure, and a successful
  read is still silent - so the fix cannot over-report;
- every frame is still published with its action and a gapless `seq` - so the fix cannot be
  "let it propagate", which would drop the frame;
- `stats["errors"]` stays 0 - so the fix cannot bump the frame-loss counter instead.

## Verification

- `mypy strands_robots tests tests_integ`: 19 errors on this branch, 19 on `main` - unchanged,
  none in the touched files.
- Blast radius (all of `tests/mesh` plus every test file referencing `InputPublisher`,
  `teleop_mixin` or a docs page, 194 files) run on this branch and on `main`:
  branch 21 failed / 4358 passed / 24 errors, `main` 26 failed / 4353 passed / 24 errors.
  Branch-only failures: none. The 5 extra failures on `main` are exactly this PR's new tests
  before the fix, and `4358 - 4353 = +5` reconciles against them. The 45 shared failures are
  pre-existing (`awsiot` / zenoh backends not installed in this environment).

## Docs

`docs/mesh.md` mesh-teleop section now states that the events read is best-effort, that it never
stops the joint stream, and that a failed read is reported through `event_read_errors` rather
than only as `null` on the wire. The module docstring's topic schema records the same ambiguity.

### DIFF
--- a/changelog.d/2399-teleop-event-read-is-reported.md
+++ b/changelog.d/2399-teleop-event-read-is-reported.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- **mesh/input**: a failed `get_teleop_events()` read is no longer published as `events: None`
+  with no other trace. `null` is also what a leader arm with no event surface publishes, so a
+  teleoperator whose event surface stopped answering dropped the operator's `terminate_episode`
+  signal while joint commands kept flowing and `stats` reported a clean session. The read stays
+  best-effort and never drops the frame the follower is tracking, but a failure now increments
+  `InputPublisher.stats["event_read_errors"]` and logs a warning naming the device and the cause.
--- a/docs/mesh.md
+++ b/docs/mesh.md
@@ -139,6 +139,17 @@ follower.stop_teleop("leader")
 
 `get_teleop_status()` on either side inspects current teleop state.
 
+Each published frame carries the operator's control signals from the
+teleoperator's `get_teleop_events()` - `terminate_episode`, `success`,
+`rerecord_episode`, `is_intervention` - alongside the joint action. Reading them
+is best-effort: a teleoperator whose event surface stops answering (a keyboard
+listener thread that died, a gamepad unplugged mid-session) never stops the joint
+stream the follower is tracking. Because that field is also `null` for a leader
+arm with no event surface at all, a failed read is reported on the publisher
+rather than only on the wire - it increments `event_read_errors` in
+`get_teleop_status()` and logs a warning naming the device and the cause, so an
+operator whose signals are being dropped can see it.
+
 `source_peer_id` and `device_name` are single segments of the mesh key
 expression `strands/{peer_id}/input/{device_name}`, so both must be plain
 identifiers (`[A-Za-z0-9_.-]+`, at most 128 chars). A Zenoh wildcard (`*`,
--- a/strands_robots/mesh/input.py
+++ b/strands_robots/mesh/input.py
@@ -15,6 +15,11 @@ Topic schema for ``strands/{peer_id}/input/{device_name}``::
         "action": {"motor.pos": float, ...},
         "events": {"terminate_episode": bool, ...} | null
     }
+
+``events`` is ``null`` both when the teleoperator exposes no
+``get_teleop_events()`` surface and when reading it failed, so the publisher
+side reports a failed read through ``InputPublisher.stats``
+(``event_read_errors``) and a log line rather than only on the wire.
 """
 
 from __future__ import annotations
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Replaces the bare `except Exception: pass` around `get_teleop_events()` in `InputPublisher._publish_loop` with a counted, rate-limited-logged recovery path: a failed read still publishes the frame with `events: None` (so the joint stream the follower is tracking never stops), but now increments a new `stats["event_read_errors"]` counter and logs a warning naming the device and cause. The wire schema is deliberately unchanged; the failure becomes visible on the publisher side where the operator is. The literal `5` log cap in the sibling loop handler is lifted into a shared `_MAX_LOGGED_LOOP_ERRORS` constant so the two handlers agree by construction.

## What's good

- The kept `except Exception` breadth is a documented recovery path with the justification in a comment at the handler, matching AGENTS.md's exception-clause discipline, and `events` is re-initialized to `None` each iteration so a failed read can never publish a stale prior value.
- The new counter is kept distinct from `errors` (which counts lost frames), with a control test pinning that they are not conflated.
- Test premise verified: the new suite reports 5 failed / 4 passed on the pre-fix tree and 9 passed on the head, exactly as the description claims; the 4 controls each rule out a different wrong fix (over-reporting the legitimate null, letting the exception propagate, bumping the frame-loss counter).
- All 236 neighboring mesh input/teleop tests pass on the head; `ruff check` and `ruff format --check` are clean on the touched files.
- Changelog fragment, docs (`docs/mesh.md`), and the module docstring's schema note all land in the same PR; no host paths, no emojis, wire format untouched.

PR #2400, opened 2026-08-17
fix(tools/lerobot_teleoperate): keep the session record of a process that exists but cannot be inspected

cagataycali:
## Problem

`SessionManager._load_sessions` prunes finished teleop sessions **and writes the pruned store back to disk**. That store is the only place a detached teleoperation subprocess's pid is recorded - `stop` and `status` both look the session up there - so the prune's classification decides whether a session stays stoppable.

Two probes decide it. `psutil.pid_exists(pid)` answers existence; `psutil.Process(pid).is_running()` refines that (it also rules out pid reuse). Both ways the second one can raise were caught in one clause:

```python
except (psutil.NoSuchProcess, psutil.AccessDenied):
    pass
```

They mean opposite things. `NoSuchProcess` means the process was reaped between the two calls, so the record names nothing. `AccessDenied` means the process **exists** - `pid_exists` just said so - and this user may not inspect it; a session started under `sudo` to reach a serial port and later listed as the invoking user reads exactly this way. Treating that as death erased the record.

Measured against a real live subprocess and the real tool (pid exists throughout):

| | before | after |
|---|---|---|
| `list` | `**Active Teleoperation Sessions** (0)` | `(1)` |
| session store on disk after that `list` | `{}` - erased by a read-only query | `{"arm_teleop": ...}` |
| `stop` | `error` / `Session 'arm_teleop' not found` | `success` / `**Session Stopped**` |
| the process afterwards | **still running** | terminated (signal 15) |

The erasure is irreversible and happens on any load, including `list` and `status`; because `add_session`/`remove_session` load first, merely starting another session persists it. The teleoperation process keeps driving the arm with no supported way to stop it.

## Why this is a classification bug rather than a policy choice

* The loop's own comment says it cleans up **dead** sessions. `pid_exists` had already returned `True`.
* The sibling store, `lerobot_train.SessionManager._load_sessions`, reaches the same "cannot confirm running" verdict and **never writes it back**, so its record survives and its comment spells out the reason: *"only the running flag is derived from the PID."* This store conflated the record with the running flag and then persisted the conflation.
* Presence here is not the running claim in this module either - `list` and `status` each compute `is_running` from `pid_exists` at the moment of the call. So a retained record is reported running only while its pid really exists.

## Change

Split the handlers. `NoSuchProcess` still prunes, unchanged. `AccessDenied` keeps the record and reports it at `WARNING`, naming the pid - silence was the operator's only signal that a session had vanished. The docstring records which probe result means what, and `docs/hardware/tools.md` gains the verdict table.

## Tests

`tests/tools/test_teleop_session_store_keeps_a_live_pid.py`, **5 failed / 4 passed on `main`, 9 passed after**:

| test | on `main` |
|---|---|
| an uninspectable live session survives on disk | `assert None is not None` |
| a read-only query does not delete such a record | `assert 'arm_teleop' in {}` |
| adding a session does not erase an uninspectable sibling | `assert 'arm_teleop' in {'second': ...}` |
| `stop` can still reach a session it could not inspect | `assert 'error' == 'success'` |
| retaining the record is reported | no log record |
| a process reaped mid-probe is still pruned | passes |
| a pid that no longer exists is still pruned | passes |
| a pid that is not running is still pruned | passes |
| the training store's prune stays non-destructive | passes |

The four that pass on both trees are the bound: they fail if the retention grows into "never prune anything", and the last pins the sibling store this one is aligned with.

## Verification

* 35-file blast radius (everything referencing the session stores, these tools or `psutil`, plus the repo-wide guards): branch **2 failed / 2370 passed**, `main` with the new file copied in **7 failed / 2365 passed**; branch-only failures **none**, and `7 - 5 = 2` shared pre-existing failures unrelated to this change.
* `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` identical on both trees, none in the touched files.
* Docs/changelog/docstring/ASCII guards: 542 passed.

## Scope

Whether such a session should be *reported* by `lerobot_train`'s store is pinned the other way by its own test, with reasoning, and its prune is already non-destructive - so it is left alone and pinned here instead.

## Artifact

One capture script run against each tree: same live subprocess, same session record, same tool calls, only `strands_robots` differs.

before and after

### DIFF
--- a/changelog.d/2400-teleop-session-store-keeps-a-live-pid.md
+++ b/changelog.d/2400-teleop-session-store-keeps-a-live-pid.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **tools/lerobot_teleoperate**: a teleop session whose process exists but cannot be inspected (`psutil.AccessDenied`, e.g. a session started under `sudo` for serial-port access and later listed as the invoking user) is no longer erased from the session store. The prune is written back to disk and that store holds the only copy of the detached process's pid, so dropping the record left the session unlisted, unstoppable, and still driving the arm. A process reaped mid-probe (`psutil.NoSuchProcess`) is still pruned, and a retained record is now reported at `WARNING`.
--- a/docs/hardware/tools.md
+++ b/docs/hardware/tools.md
@@ -56,6 +56,25 @@ that means "stop at once", so treating it as unset would invert the request.
 Passing a value an action ignores is never an error: `action="start"` without a
 `dataset_repo_id` teleoperates and reads no `dataset_*` option.
 
+### A session is only forgotten once its process is gone
+
+Because the session runs detached, the on-disk session store is the only place
+its pid is recorded - `stop` and `status` both look the session up there. Every
+load prunes finished sessions and writes the pruned store back, so what counts
+as "finished" decides whether a session stays stoppable:
+
+| What the probe reports | Verdict |
+|------------------------|---------|
+| the pid no longer exists | finished - pruned |
+| `psutil.NoSuchProcess` (reaped between the existence check and the probe) | finished - pruned |
+| `is_running()` returns `False` (a zombie, or the pid was reused) | not this session - pruned |
+| `psutil.AccessDenied` (the pid exists, this user may not inspect it) | kept, and reported at `WARNING` |
+
+The last row is why a session started under `sudo` - a common way to reach a
+serial port - is still listed and still stoppable when the tool is later invoked
+as the unprivileged user. Being kept is not a claim that it is running: `list`
+and `status` each derive that from the pid's existence at the moment you ask.
+
 ### A raw servo write is bounded by the register it encodes into
 
 `serial_tool` writes Feetech registers by masking the value into fixed-width
--- a/strands_robots/tools/lerobot_teleoperate.py
+++ b/strands_robots/tools/lerobot_teleoperate.py
@@ -267,7 +267,30 @@ class SessionManager:
         self.sessions_file = SESSION_DIR / "active_sessions.json"
 
     def _load_sessions(self) -> dict[str, Any]:
-        """Load active sessions from disk."""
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Splits the single `except (psutil.NoSuchProcess, psutil.AccessDenied)` clause in `lerobot_teleoperate.SessionManager._load_sessions` into two handlers with opposite verdicts: `NoSuchProcess` (reaped mid-probe) still prunes, while `AccessDenied` (the pid exists — `pid_exists` returned `True` one line earlier — but this user may not inspect it) now retains the record and logs a `WARNING` naming the pid. The distinction is load-bearing because this store's prune is written back to disk and it holds the only copy of a detached session's pid, so the old behavior let a read-only `list` irreversibly erase a live session's record, leaving the process driving the arm with no supported `stop` path. Verified against the tree: the pre-fix clause and write-back are exactly as described, and the sibling `lerobot_train` store's non-destructive prune (which this change is aligned with) is pinned by the new test file. The retained record cannot go stale-forever: `list`/`status` derive `is_running` from `pid_exists` at call time, and once the pid is gone the ordinary prune path removes it.

## What's good

- Regression tests pin both directions: the retention (5 tests failing on `main`) and the bound that keeps it from growing into "never prune anything" (3 prune controls plus the `lerobot_train` sibling pin).
- Tests redirect `SESSION_DIR` for both stores to `tmp_path` via `monkeypatch.setattr`, so nothing touches the real cwd-relative store — consistent with the repo's test-isolation conventions.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; docs table in `docs/hardware/tools.md` records the four probe verdicts; log message is plain ASCII with lazy `%s` formatting.
- Scope discipline: `lerobot_train`'s reporting policy is deliberately left alone and pinned rather than changed alongside.

PR #2402, opened 2026-08-17
fix(simulation/mujoco): resolve a mesh reference the way MuJoCo resolves it

cagataycali:
## Problem

`MuJoCoSimEngine._ensure_meshes` answers one question before a model is compiled: are this model's mesh files on disk? A "no" triggers `download_robots(force=True)` - a full unconditional Menagerie re-fetch - and `add_robot` returns that download's error dict when the fetch cannot run.

It answered "no" for models whose meshes were all present, because it resolved a reference against the directory of whichever fragment declared it, joined with a `meshdir` attribute read from that same fragment. MuJoCo resolves neither part that way.

Measured on MuJoCo 3.5.0 and 3.10.0 by loading models whose mesh sits at exactly one candidate location:

| declared where | mesh on disk at | MuJoCo |
|---|---|---|
| `assetdir="bag"` in an included fragment | `<model dir>/bag/tet.stl` | accepted |
| `assetdir="bag"` in an included fragment | `<fragment dir>/bag/tet.stl` | **rejected** |
| no attribute, `file="meshes/tet.stl"` | `<model dir>/meshes/tet.stl` | accepted |
| no attribute, `file="meshes/tet.stl"` | `<fragment dir>/meshes/tet.stl` | accepted |
| `meshdir` + `assetdir` together | under `meshdir` | accepted (`meshdir` wins) |
| `<compiler>` in the top file, `<mesh>` in an include | - | accepted |
| `<compiler>` in an include, `<mesh>` in the top file | - | accepted |

So three things hold, and the old check got all three wrong:

- the mesh directory may be declared as **`assetdir`** (which sets the mesh and texture directories together); `meshdir` overrides it where both appear;
- `<compiler>` is **model-global** - the fragment naming the directory need not be the fragment naming the mesh;
- a reference resolves against the **main model file's** directory plus that subdirectory, and additionally relative to the declaring fragment's directory when that fragment sits under the model. `<fragment dir>` joined with the subdirectory - the only base the old code used - is a location MuJoCo rejects.

### Blast radius

Of the 62 loadable registry robots, **4 reported meshes missing while MuJoCo compiled the model without complaint**; the other 58 verdicts are unchanged.

| robot | why it was misread |
|---|---|
| `stretch3`, `stretch`, `skydio_x2` | declare `assetdir="assets"`, never `meshdir` |
| `ability_hand` | declares `meshdir="./assets"` in a fragment one directory below the model |

The user-visible effect, with the asset download unable to run:

| | `stretch3` | `skydio_x2` | `ability_hand` | `panda` |
|---|---|---|---|---|
| before | `error`, `robots=[]` | `error`, `robots=[]` | `error`, `robots=[]` | `success` |
| after | `success` | `success` | `success` | `success` |

The refusal reads `Auto-download failed for 'stretch3': network unreachable. Install robot_descriptions: pip install strands-robots[sim-mujoco]` - which names neither the real cause nor anything the caller can act on, since the meshes were already there.

## Change

- `_mjcf_mesh_subdir(*contents)` and `_mjcf_mesh_candidates(ref, model_dir, subdir, include_dir="")` in `strands_robots.assets.download` hold the rule, with the measurements above recorded in their docstrings.
- `_ensure_meshes` reads the subdirectory once for the whole model and checks both candidate locations.
- `_needs_download` asked the same question with its own copy of the same regex and the same `assetdir` blind spot; both now read the rule from one place, so they cannot disagree about a model.
- `docs/simulation/world-building.md` stated the resolution rule as "`meshdir` / `texturedir`"; it now names `assetdir` too.

Behaviour is unchanged where the old base happened to coincide with a MuJoCo-accepted one, which is every model whose meshes are declared in the model file itself or in a fragment beside it.

## Tests

`tests/simulation/mujoco/test_ensure_meshes.py`, +13. Against `upstream/main`: **10 failed / 10 passed**; after: **20 passed**. Same 20 on MuJoCo 3.5.0 and on the pinned 3.10.0.

Failing before, one per rule:

- `test_assetdir_is_honored_like_meshdir`
- `test_a_compiler_in_the_top_file_applies_to_an_included_mesh`
- `test_a_compiler_in_an_included_file_applies_to_a_top_level_mesh`
- `test_an_included_fragments_mesh_resolves_against_the_main_model_dir` (the `skydio_x2`/`stretch3` layout)
- `test_needs_download_honors_assetdir` (`assert True is False`)

Passing on **both** trees, so they bound the change rather than describe it:

- `test_an_included_fragments_mesh_may_also_sit_beside_that_fragment` - the `lekiwi` layout. An earlier revision resolved against the model directory only and broke exactly this; the test is why both candidates are checked.
- `test_a_genuinely_absent_mesh_is_still_reported` - widening where we look must not stop us noticing a real absence.
- `test_meshdir_overrides_assetdir`, plus the 7 pre-existing tests.
- `test_the_declaring_fragments_dir_is_not_a_candidate` pins that the location MuJoCo rejects stays rejected.

## Verification

- Registry sweep: spurious "missing" **4 -> 0** across 62 robots; `add_robot` download calls for the 4 go 1 -> 0.
- Blast radius, 220 test files on both trees, identical collection (5038): branch **5024 passed / 0 failed**, base 5014 / 10 - the 10 being exactly the pre-fix failures above, and **no branch-only failure**.
- `ruff check` / `ruff format --check` clean on `strands_robots tests tests_integ`; `mypy` 19 errors on both trees, none in the changed files.

## Artifact

`add_robot` for each robot with the asset download unavailable, rendered headless. Top row is `upstream/main`: three robots refused with nothing to render. `panda` is byte-identical across both rows (mean |px| = 0.000487), so the change alters only where a reference is looked for.

mesh reference resolution before and after

### DIFF
--- a/changelog.d/2402-mesh-reference-resolution.md
+++ b/changelog.d/2402-mesh-reference-resolution.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **simulation/mujoco**: a mesh-presence check now resolves a reference the way MuJoCo does - honouring `assetdir` as well as `meshdir`, reading the model-global `<compiler>` subdirectory once, and resolving against the main model file's directory. Four registry robots (`stretch3`, `stretch`, `skydio_x2`, `ability_hand`) reported their meshes missing while MuJoCo compiled them, which spent a `force=True` asset re-download on every `add_robot` and refused the robot outright when that download could not run. The download path shared the same blind spot and now reads the rule from the same place.
--- a/docs/simulation/world-building.md
+++ b/docs/simulation/world-building.md
@@ -533,7 +533,8 @@ other.load_scene(scene_path="/tmp/handoff.xml")   # same scene, same structure
 
 Mesh, texture and height-field assets are referenced by ABSOLUTE path. MuJoCo
 resolves a relative `file=` against the model's own directory (plus `meshdir` /
-`texturedir`), and that directory is not part of the serialised XML - so a
+`texturedir`, or the `assetdir` that sets both), and that directory is not part
+of the serialised XML - so a
 relative reference would resolve against wherever the export happened to be
 written. Absolute references keep the export reloadable from any location, and a
 scene composed from several models needs them: each model contributes assets from
--- a/strands_robots/assets/download.py
+++ b/strands_robots/assets/download.py
@@ -108,6 +108,63 @@ def _resolve_robot_descriptions_module(name: str, info: dict) -> str | None:
 get_user_assets_dir = get_assets_dir
 
 
+def _mjcf_mesh_subdir(*contents: str) -> str:
+    """Return the mesh search subdirectory declared by MJCF text.
+
+    MuJoCo's ``<compiler>`` element offers two attributes for this. ``meshdir``
+    names the mesh directory specifically; ``assetdir`` names the mesh AND
+    texture directories at once, and ``meshdir`` overrides it where both appear.
+    Both are model-global: they apply across ``<include>``, so the fragment
+    declaring the directory need not be the fragment declaring the mesh.
+
+    A reader that knows only ``meshdir`` resolves an ``assetdir`` model against
+    the model directory itself, so it reports a mesh that is present as absent.
+
+    Args:
+        *contents: MJCF text fragments making up one model, in any order.
+
+    Returns:
+        The declared subdirectory, or ``""`` when no fragment declares one.
+    """
+    for attr in ("meshdir", "assetdir"):
+        for content in contents:
+            if match := re.search(rf'{attr}="([^"]*)"', content):
+                return match.group(1)
+    return ""
+
+
+def _mjcf_mesh_candidates(mesh_ref: str, model_dir: str, mesh_subdir: str, include_dir: str = "") -> list[str]:
+    """Return the on-disk locations MuJoCo accepts for one mesh reference.
+
+    MuJoCo resolves a ``<mesh file=...>`` against the MAIN model file's
+    directory joined with the model's mesh subdirectory - never against the
+    directory of whichever ``<include>``d fragment happened to declare it. When
+    the declaring fragment lives in a subdirectory of the model, MuJoCo also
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Fixes a false "meshes missing" verdict in `MuJoCoSimEngine._ensure_meshes` (and the parallel check in `_needs_download`) by resolving mesh references the way MuJoCo actually resolves them: honouring `assetdir` as well as `meshdir`, treating `<compiler>` as model-global across `<include>` fragments, and resolving against the main model file's directory (plus the fragment-relative fallback MuJoCo also accepts). The resolution rule now has a single owner in `strands_robots.assets.download` (`_mjcf_mesh_subdir` / `_mjcf_mesh_candidates`), which both callers read, so the presence check and the download decision cannot drift apart. This eliminates the spurious `force=True` re-download on every `add_robot` for `stretch3`, `stretch`, `skydio_x2`, and `ability_hand`, and the outright refusal when the download cannot run.

## What's good

- New helpers are `_`-private with the measured MuJoCo behaviour recorded in their docstrings; no public API surface added.
- 13 new tests with clear split between failing-before (one per resolution rule) and passing-on-both-trees bounds (`lekiwi` layout, genuine-absence still reported, fragment-dir stays rejected); all 20 pass on the head, and the pre-existing asset-download suites (61 tests) are green.
- `except Exception` narrowed to `(OSError, UnicodeDecodeError)` with reasoning per AGENTS.md's narrow-exception rule; unreadable-file behaviour preserved and pinned by the pre-existing test.
- Changelog fragment under `changelog.d/` per step 3; docs updated to name `assetdir` alongside `meshdir`; no host paths or non-ASCII introduced.

## Verification suggestions

- Registry spot-check on a machine with Menagerie assets present and network blocked: `Robot("stretch3")` / `Robot("skydio_x2")` should succeed without any download attempt (the PR's sweep claims 4 -> 0 spurious downloads across 62 robots).

PR #2403, opened 2026-08-17
security(mesh): refuse control characters in a wire instruction

cagataycali:
## Summary

`validate_command` charset-gates every wire string it admits except the one its own comment names. The comment introducing the sim-targeted `execute` / `start` fields (`mesh/security.py`) states the purpose of the block:

> Validating them here keeps the wire schema honest end-to-end so a malicious peer cannot smuggle **control-byte instruction strings** in via `robot_name`

`robot_name`, `policy_host`, `server_address`, `turn_id`/`sender_id` and `override_code` are each checked against a CRLF/NUL/C0 gate. `instruction` - the field that sentence names - was bounded only by type, emptiness and length, so the bytes those five fields refuse travelled through it.

`instruction` is free-form natural language supplied by a remote peer (`Mesh._exec_cmd` reads it straight off the command payload). Any log record that echoes it then emits two records instead of one, and the second is fully attacker-shaped: it can claim an arbitrary level and logger name.

## Measured

Same payload bytes, offered through `instruction` and through the sibling field the comment names:

| bytes in the value | `instruction` | `robot_name` |
|---|---|---|
| `pick up the red cube` | admitted | admitted |
| `... \n WARNING:audit:E-STOP CLEARED` | **admitted** | refused |
| `... \r\n ...` | **admitted** | refused |
| `... \x00 ...` | **admitted** | refused |
| `... \x07 \x1b ...` | **admitted** | refused |
| `pick up the cafe cup` (non-ASCII `e`) | admitted | refused |

Feeding the admitted value to the shipped logging call it reaches:

```
raw emitted bytes: "DEBUG:...policy:VLA tokenized instruction: 'pick up cube\nWARNING:audit:E-STOP CLEARED...' -> (1, 48) tokens\n"

log records the reader sees: 2
  [0] DEBUG:strands_robots.policies.lerobot_local.policy:VLA tokenized instruction: 'pick up cube
  [1] WARNING:audit:E-STOP CLEARED...' -> (1, 48) tokens
```

instruction control-character gate: before and after

One capture script, run under each tree. The payload, the logging call and the handler are identical in both panels; only `strands_robots` differs. The right panel's control row is the point of the design: the legitimate non-ASCII instruction is still admitted.

Record `[1]` is a `WARNING` from a logger named `audit` that no code in the tree emitted. `docs/security.md` calls the safety audit trail "your forensic record of what the fleet was told to do", which a forgeable record undermines.

## Why the gate is not `_SAFE_PASSTHROUGH_RE`

The obvious fix is to reuse the sibling gate. It is wrong here, and the last row of the table above is why: `_SAFE_PASSTHROUGH_RE` is `^[\x20-\x7E]+$`, printable ASCII only. That is correct for an identifier such as `robot_name`, and it would refuse any instruction written in a language needing a non-ASCII letter.

A natural-language field bounds the **control range** and nothing else, so the new gate refuses C0 (`0x00-0x1F`, which is where NUL, CR and LF live), DEL (`0x7F`) and C1 (`0x80-0x9F`), and admits every printable codepoint. `test_accepts_printable_non_ascii` pins that and fails if the ASCII-only regex is ever substituted.

## The refusal

```
instruction contains a control character (U+000A at offset 12); send printable
text only (CRLF, NUL and other C0/C1 bytes are refused)
```

It names the codepoint and offset rather than echoing the instruction. A `ValidationError` is itself logged by the dispatcher, so interpolating a 2000-char attacker-controlled payload would carry the injection into the very record that reports it. The sibling fields can echo with `%r` because they are short and ASCII; a natural-language field is neither.

## Scope

Ordering is unchanged - type, then emptiness, then length, then charset - matching the discipline `_SAFE_PASSTHROUGH_RE` documents ("applied *after* type-check + length-bound"). A whitespace-only instruction is still refused as empty and an oversize one still reports its length; both are pinned.

This narrows only what a **remote peer** may put on the wire, which is the boundary `mesh/security.py` exists to defend ("everything here defends against an *authenticated* peer ... whose payload contents we still need to bound"). A local Python caller passing `run_policy(instruction=...)` is untouched. No multi-line or control-character instruction exists anywhere in `strands_robots/`, `tests/`, `docs/` or `examples/`.

The log call in `policies/lerobot_local` that echoes the instruction with `%s` is left as-is: closing the wire boundary removes the remote path entirely, and whether a provider should digest caller text rather than echo it is a separate contract (`policies/kimodo` already does, with its rationale in a comment). Fixing it here would make this two changes.

## Tests

Added to `tests/mesh/test_security_control_char_gates.py`, which is the guard for this contract - it pins the CRLF/NUL/C0 gate for the five other fields and used `instruction: "go"` only as a benign carrier.

- 9 control bytes x refused, x codepoint-and-offset named, x parity with `robot_name`
- the refusal does not echo the payload
- the gate covers `start` as well as `execute`
- 4 controls: a clean instruction is admitted, **printable non-ASCII is admitted**, a whitespace-only instruction is still refused as empty, the length bound still outranks the charset gate

Pre-fix / post-fix split on the file (34 new tests):

| tree | result |
|---|---|
| `upstream/main` + this test file | **30 failed**, 29 passed |
| this branch | **59 passed** |

The 4 tests that pass on both trees are the controls: they fail if the gate is widened to the ASCII-only regex, or if it displaces the emptiness or length verdicts.

## Verification

- `ruff check` / `ruff format --check` over `strands_robots tests tests_integ`: clean
- `mypy strands_robots tests tests_integ`: 19 errors in 8 files, identical set on this branch and `upstream/main`, none in the changed files
- 243-file blast radius (every test referencing `validate_command`, `MAX_INSTRUCTION_LEN` or `_SAFE_PASSTHROUGH_RE`, all of `tests/mesh`, every test reading a docs page, plus the repo-wide docstring/ASCII/changelog guards), same selection on both trees:

| tree | failed | passed | errors | failing node ids |
|---|---|---|---|---|
| this branch | 24 | 6281 | 24 | 48 |
| `upstream/main` + this test file | 54 | 6251 | 24 | 78 |

Both collected 6329. Regressions (`comm -23`): none. The 30 base-only ids are exactly this file's pre-fix failures; the 48 shared are pre-existing (`awsiot` absent and similar).

Docs: `docs/security.md` gains a "Payload bounds" bullet beside the audit-trail bullet it supports, stating what `validate_command` bounds and why a natural-language field bounds only the control range while an identifier field stays ASCII-only.


yinsong1986 (approved):
## Summary

Closes the one charset gap in `validate_command`: the wire `instruction` field for `execute`/`start` was bounded only by type, emptiness and length, so a remote peer could smuggle CR/LF/NUL and other control bytes through it into any log record that echoes it (log-record forgery). The fix adds a dedicated `_NATURAL_TEXT_CONTROL_RE` gate refusing C0/DEL/C1 while deliberately admitting printable non-ASCII - the right call for a natural-language field, and correctly not a reuse of the ASCII-only `_SAFE_PASSTHROUGH_RE` used for identifier fields. The refusal names the codepoint and offset without echoing the payload, so the `ValidationError` the dispatcher logs cannot itself carry the injection. Check ordering (type -> empty -> length -> charset) matches the sibling fields and is pinned by tests.

## What's good

- The new gate slots in after the existing length check, so pre-existing verdicts (`non-empty`, `exceeds N chars`) keep their wording - both pinned by control tests.
- 34 new tests verified passing on the head tree (59/59 in the file), including parity with `robot_name`, coverage of both `execute` and `start`, a no-echo assertion on the refusal message, and a printable-non-ASCII admit test that fails if the gate is ever widened to the ASCII-only regex.
- Tightening a security boundary is the reversible direction: refusal is loud (`ValidationError`), no in-tree caller sends a multi-line or control-character instruction, and a later widening breaks nobody.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; `docs/security.md` bullet lands beside the audit-trail bullet it supports; test file is ASCII-clean (non-ASCII test strings use `\u` escapes).

PR #2404, opened 2026-08-17
fix(tools): measure a published wait budget on a clock that cannot step

cagataycali:
## Problem

An agent-callable tool publishes its time budget as a tool parameter -- `use_ros(timeout=...)`, `use_rtps(timeout=...)`, `lerobot_camera(preview_duration=...)`, `serial_tool`'s five-second monitor window -- so the budget is part of the tool's contract with its caller. Six waits measured it on `time.time()`, which is not a clock: it is the current opinion about the date, and an NTP correction, a `date -s`, or a resume from suspend moves it by an arbitrary amount mid-wait. Forward, the wait ends early with the work still in flight, so a message that was arriving is reported as a timeout. Backward, it runs past the caller's budget by the size of the step. Neither is reported, because nothing raises.

`_RosBackend.spin_for` is the sharpest case, for two reasons.

**It is the only blocking primitive in the ROS 2 tool.** All 13 blocking waits -- every graph-introspection settle, `echo`, `service_call`, and each stage of `action_send_goal` -- go through it.

**One logical deadline was measured on two clocks.** `_action_send_goal` governs server discovery, goal acceptance and result delivery with a single `time.monotonic()` deadline and hands each remaining slice to `spin_for`, which re-based that slice on `time.time()`. The outer monotonic guard cannot correct the inner loop, because it is only consulted *between* `spin_for` calls.

Driving the real `spin_for` with a stub rclpy executor and a 1.0s budget:

| wall clock during the wait | seconds actually waited | executor pumps | verdict |
|---|---|---|---|
| no step (control) | 1.001 | 99 | honoured |
| steps **+30s** | **0.302** | 30 | abandoned at 30% of the budget, predicate still false |
| steps **+1h** | **0.303** | 30 | same |
| steps **-2s** | **3.006** | 298 | caller held 3x past its budget |

The step size does not have to be large to matter, and the `+1h` row shows why the truncation is not proportional to it -- what is lost is the remainder of the wait, whatever the step.

That also reaches the fail-safe this module documents: `_action_send_goal` cancels before surfacing a timeout so "a timed-out `use_ros` call never leaves a physical robot executing an orphaned goal", and waits two seconds on the cancel future. A cancel request only leaves the process while the executor is pumped, so a wait truncated by a step reports a cancel it may not have sent.

## Change

Every *duration* in the six sites now reads `time.monotonic()`. No behaviour changes absent a clock step; the control row above is unchanged.

Absolute *stamps* are deliberately left on the wall clock, and one function shows both halves: `serial_tool`'s monitor bounds its window on `time.monotonic()` while each returned record keeps a `time.time()` `timestamp`, because that names a point in time a reader correlates with other logs and seconds of local process uptime would not. That is the boundary the safety subsystem already settled three times -- `tests/mesh/test_replay_cache_monotonic.py`, `tests/mesh/test_bridge_dedup.py::TestDedupClock`, `tests/mesh/test_corroboration_clock_domain.py`, whose docstring calls monotonic "the same domain as every other local bookkeeping timestamp". It had never been written down, so it is now a convention entry in `AGENTS.md`, and `docs/ros2-integration.md` states it beside the `timeout` row it governs.

### Scope

The boundary is measured rather than asserted. Of the eight sites in the package where a wall-clock read decides whether to keep waiting, six are in a `@tool` module that *publishes* the budget, and all six are here. The other two -- `Robot._execute_task_async`'s task loop and `TeleopMixin._teleop_loop`'s session deadline -- are reached through the Python API rather than a tool schema, and each shares its stamp with a value it also *reports* (`get_task_status().duration`, `get_teleoperate_status().elapsed`), so they need a report/decide split rather than a clock swap. That is a separate change and not folded in here.

## Tests

`tests/tools/test_tool_wait_budgets_survive_a_clock_step.py`, **5 failed / 6 passed** on `main` -> **11 passed**.

Failing pre-fix, each on the contract rather than on a fixture detail:

- the scan, naming all six: `['gr00t_inference.py:1383', 'lerobot_camera.py:816', 'serial_tool.py:376', 'use_ros.py:308', 'use_rosbridge.py:217', 'use_rtps.py:292']`
- `these deadlines are not on the clock the callers measure their budget with: ['deadline = time.time() + timeout']`
- `a +30s wall-clock step cut a 0.6s wait to 0.101s`
- `a -2s wall-clock step stretched a 0.6s wait to 2.604s`
- `the cancel wait pumped the executor only 10 times in 0.051s of a 1.0s budget`

Passing on both trees, which is what bounds the change:

- with no step the wait is unchanged and still pumps the executor, and a satisfied predicate still short-circuits -- the fix is a no-op absent a step
- a *reported* wall-clock read is not treated as a wait decision, so the scan cannot be satisfied by moving stamps
- `serial_tool`'s record `timestamp` is still on the wall clock -- this fails if the module is ever swept clock-blind rather than per value
- a planted wall-clock wait is still detected, so a clean scan means the tools are right rather than that the scan sees nothing
- the stepping-clock double really takes the step it advertises

The premise assertions are stated *relatively* (`wall_moved - elapsed == step_by`), which matters: with a -2s step and a wait that overran to 2.6s the wall clock still moved forward on net, so an absolute "the clock went backwards" premise would have been a statement about the bug rather than about the clock.

`tests/tools/test_serial_tool.py::test_monitor_collects_chunks` scripted `time.time` by call order (`iter([0.0, 0.0, 1.0, 10.0, 10.0])`), so which reads consumed the script changed with the clock and it had to be re-scripted onto the clock that now bounds the window. It cannot pass on the pre-fix code either -- there the window is on a wall clock it no longer scripts, and with `sleep` stubbed the loop spins until `pytest-timeout` fires. Its assertions are strictly stronger than before: it now also pins that the record stamp comes from the wall clock, which nothing checked.

## Verification

- `mypy strands_robots tests tests_integ`: 19 errors, identical to `main`'s 19, none in the changed files.
- 45-file selection covering every test that touches the six modules plus the repository-wide guards: **2108 passed / 0 failed** on this branch; `main` with the new test file added is 5 failed / 2103 passed. Both collected 2116 and `2103 + 5 = 2108`, so the only verdicts that changed are the five the fix flips.
- No non-ASCII added in any changed file.

## What the fix looks like

The probe below is the same script run twice, against the same stub rclpy executor, differing only in which tree it imports. A pump is the only moment a subscription can deliver or a cancel request can leave the process.

spin_for honouring a 1.0s timeout while the host wall clock steps


yinsong1986 (approved):
## Summary

Moves six wait-budget measurements in agent-callable tools (`use_ros.spin_for`, `use_rosbridge` reconnect, `use_rtps` echo poll, `lerobot_camera` preview/FPS pacing, `serial_tool` monitor window, `gr00t_inference` service-readiness poll) from `time.time()` to `time.monotonic()`, so an NTP step / `date -s` / suspend-resume mid-wait can no longer truncate or stretch a published timeout. Absolute stamps are deliberately left on the wall clock (`serial_tool`'s per-record `timestamp` is unchanged and now pinned by a test), and the duration-vs-stamp boundary is written into `AGENTS.md` and `docs/ros2-integration.md`.

## What's good

- The new test file scans `strands_robots/tools` with no exemption list, includes a planted-offender meta-test so a clean scan is meaningful, and pins both halves of the boundary (monotonic waits *and* the wall-clock record stamp), so a future clock-blind sweep in either direction fails.
- Behaviour tests drive the real `spin_for` under forward and backward steps with relative premise assertions, and the control case pins that the fix is a no-op absent a step.
- Scope discipline: the two non-tool wall-clock waits (`Robot._execute_task_async`, `TeleopMixin._teleop_loop`) are explicitly measured and deferred with a stated reason rather than swept blindly.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; no non-ASCII or host paths in the changed files; tool-visible contracts (parameter names, record schema) are unchanged, so there is no one-way-door surface here.

PR #2405, opened 2026-08-17
fix(training/lerobot): report an absent lerobot call-time dependency at preflight

cagataycali:
## Problem

`LerobotTrainer.validate` is documented as a pure preflight whose empty list "means the spec is launchable", and `Trainer.validate` says an implementation SHOULD check that "any backend-required input is present". It already reports an absent lerobot, which `tests/training/test_lerobot.py` pins as surfacing the problem "rather than deferring the failure to the training launch".

Locating lerobot does not establish that lerobot can train. `lerobot.scripts.lerobot_train.train` calls `require_package(...)` for packages that are **not module-scope imports** of that module, so a spec lookup answers nothing about them — and the first such call is the opening statement of `train()`:

```python
require_package("accelerate", extra="training")
```

No `strands-robots[lerobot]` install supplies `accelerate`: that extra is exactly `lerobot[feetech,dataset]`. So `validate()` returned `[]` for a spec that could not train.

That is not only a late message. `train()` fails closed on a `validate()` problem precisely so nothing is touched before a knowably-bad run; with the problem unreported it instead ran on to the fresh-start hygiene that removes a checkpoint-less `output_dir` — deleting a directory for a run that never started.

Measured with `accelerate` hidden from `importlib`, `policy_type="smolvla"`:

|---|---|---|
| `validate()` | `[]` (launchable) | 1 problem naming `accelerate` + `pip install 'lerobot[training]'` |
| `output_dir` after `train()` | **removed** (`operator_notes.txt`, `launch.sh` gone) | intact |
| `result.message` | lerobot's `ImportError`, raised after the delete | `validation failed: 'accelerate' is required ...` |

## Change

`validate()` now reports the packages lerobot's `train()` will require, under the conditions this trainer itself determines:

- `accelerate` — always. It is required unconditionally, before `train()` branches on device and before it loads a dataset, so "no GPU" does not mean "no extra".
- `peft` — when `method == "lora"`, which is exactly when `build_config` sets `cfg.peft`, the flag lerobot gates `peft` on.

The probe is `importlib.util.find_spec`, the same lookup lerobot's own `is_package_available` uses, so `validate()` stays read-only and answers the very question the runtime gate will ask. Each problem names the lerobot extra that supplies the package, so the remedy is runnable.

Reported only once lerobot itself is present, so an install with neither gets one root cause rather than two.

### Out of scope

`diffusers` is the third `require_package` call site, guarded by `cfg.ema.enable`. Nothing here sets `cfg.ema`, so it is reachable only by naming `ema.enable` through the generic dotted `extra` passthrough; enumerating what an arbitrary passthrough key implies is a different question from what a spec's own fields ask for.

The extras themselves are unchanged. `docs/training/overview.md` documents the `lerobot[training]` requirement deliberately, so this makes the code agree with that document rather than moving the packaging boundary.

## Docs

The same paragraph claimed "the only `strands-robots` extra that declares `accelerate` is `cosmos3-diffusers`". `[kimodo]` declares `accelerate>=0.30.0` too, and unlike `cosmos3-diffusers` it **is** in `[all]` — so `[all]` has `accelerate` transitively while `[lerobot]` does not. Corrected, and the paragraph now states that `validate()` reports the absence.

## Tests

`tests/training/test_call_time_dependency_preflight.py`, 9 tests. Absence is simulated by making the package unlocatable to `find_spec`, which is how an uninstalled one presents to both this preflight and lerobot's own probe; asserting only where the package genuinely happens to be missing would leave the contract unverified in CI, where `[all]` supplies `accelerate` via `[kimodo]`.

Against `upstream/main`: **6 failed / 3 passed**. Headline:

```
AssertionError: validate() reported no problem for an absent 'accelerate', so it
called this spec launchable; lerobot's train() requires it as its first statement. Got: []

AssertionError: train() removed the output_dir for a spec whose dependency was
absent at preflight time; validate() must report it so train() fails closed instead.
```

4 of the 6 are behavioural; 2 exercise the new availability helper directly. The 3 that pass on both trees are the controls: a clean spec stays clean, `method="full"` does **not** demand `peft` (the conditional must not become unconditional), and an install with no lerobot at all still names lerobot rather than also the packages its `train()` would have gone on to require.

An autouse fixture restores `lerobot.utils.import_utils._require_package_cache`: `require_package` memoizes each answer process-globally, so hiding a package must not strand a `False` for the rest of the session.

## Verification

Blast radius = 133 test files referencing `LerobotTrainer` / `lerobot_local` / `TrainSpec` / `create_trainer`, plus the repo-wide docstring, ASCII, dependency-audit and changelog guards, run on both trees:

- branch **3 failed / 4802 passed / 56 skipped**, base **9 failed / 4796 passed / 56 skipped** — both collected 4861.
- Branch-only failures: **none**. Base-only: exactly the 6 pre-fix failures above (`9 - 3 = 6`, `4802 - 4796 = 6`).
- The 3 shared failures are a pre-existing local environment artifact (`draccus 0.10.0` against lerobot's `draccus>=0.11.6`, `TypeError: encode() takes 1 positional argument but 2 were given`); each reproduces on a pristine `upstream/main`.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 19 pre-existing errors on both trees, branch-only set empty, zero in the two files touched.

## Artifact

One capture script run under each tree with `accelerate` hidden — same script, same environment, only `strands_robots` differs.

call-time dependency preflight

### DIFF
--- a/changelog.d/2405-lerobot-call-time-dependency-preflight.md
+++ b/changelog.d/2405-lerobot-call-time-dependency-preflight.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **training/lerobot**: `LerobotTrainer.validate` now reports the packages lerobot's own `train()` requires at call time - `accelerate` always, and `peft` when `method == "lora"` - instead of returning no problems for a spec that cannot train. `lerobot.scripts.lerobot_train.train` reaches those through `require_package(...)` rather than a module-scope import, so locating lerobot said nothing about them, and no `strands-robots[lerobot]` install supplies `accelerate`. `train()` fails closed on a validate() problem, so an `output_dir` is no longer cleared on the way to a run that could never start. Each problem names the `lerobot[...]` extra that supplies the package.
--- a/docs/training/overview.md
+++ b/docs/training/overview.md
@@ -428,18 +428,22 @@ TrainSpec(..., num_gpus=8,
 as GPU.** LeRobot's `train()` calls
 `require_package("accelerate", extra="training")` *before* it branches on
 device, so "no GPU" does not mean "no extra" - and nothing on this path pulls
-`accelerate` in: the `[lerobot]` extra is exactly `lerobot[feetech,dataset]`, and
-the only `strands-robots` extra that declares `accelerate` is
-`cosmos3-diffusers`, a different provider.
+`accelerate` in: the `[lerobot]` extra is exactly `lerobot[feetech,dataset]`. The
+`strands-robots` extras that declare `accelerate` belong to other providers
+(`kimodo`, `cosmos3-diffusers`), so an install only has it by accident - `[all]`
+does, because it pulls `[kimodo]`; `[lerobot]` alone does not.
 
 ```bash
 pip install "lerobot[training]"
 ```
 
-`train()` reports the missing package in its `TrainResult` rather than raising,
-so check `result.status` and surface `result.message` - it carries LeRobot's own
-install remedy. An unchecked call hands whatever consumes `checkpoint_dir` a
-`None` instead.
+`validate()` reports an absent `accelerate` - and an absent `peft` when
+`method="lora"` - as a preflight problem, so `train()` fails closed on it and
+leaves `output_dir` untouched rather than clearing it on the way to a run that
+cannot start. Either way the cause arrives in the `TrainResult` rather than as a
+raise, so check `result.status` and surface `result.message` - it names the
+package and the `lerobot[...]` extra that supplies it. An unchecked call hands
+whatever consumes `checkpoint_dir` a `None` instead.
 
 The base `strands-robots[lerobot]` extra covers **recording, streaming, and
 loading a trained checkpoint**, and with `lerobot[training]` it covers **ACT /
--- a/strands_robots/training/lerobot.py
+++ b/strands_robots/training/lerobot.py
@@ -47,6 +47,7 @@ from __future__ import annotations
 
 import contextlib
 import dataclasses
+import importlib.util
 import json
 import logging
 import os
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes the gap between "lerobot is importable" and "lerobot can train": `LerobotTrainer.validate()` now reports the packages `lerobot.scripts.lerobot_train.train` demands via `require_package(...)` at call time - `accelerate` unconditionally, `peft` only when `method == "lora"` (the one case `build_config` sets `cfg.peft`). Because `train()` fails closed on a `validate()` problem, this also stops the fresh-start hygiene from deleting a checkpoint-less `output_dir` on the way to a run that could never start. I verified the transcribed call sites against lerobot v0.6.1 source (the version the `[lerobot]` extra pins): `require_package("accelerate", extra="training")` is the opening statement of `train()`, `peft` is gated on `cfg.peft is not None`, and both named extras (`lerobot[training]`, `lerobot[peft]`) exist in lerobot's pyproject and supply the package the remedy names. The docs correction also checks out: `[kimodo]` declares `accelerate>=0.30.0` and is in `[all]`, unlike `cosmos3-diffusers`.

## What's good

- The gate is ordered behind the lerobot-present check, so an install with neither reports one root cause; the control test pins that.
- `_module_available` stays read-only (`find_spec`, no import), catches narrowly (`ImportError`, `ValueError`), and fails toward reporting a problem rather than silently passing - consistent with the no-silent-defaults rule.
- `diffusers` deliberately excluded with the reasoning written down at the table, so the next reader knows it was a decision, not an omission.
- All 9 new tests pass; the autouse fixture restoring lerobot's process-global `_require_package_cache` prevents a hidden package from stranding a `False` for the rest of the session.
- `ruff check` / `ruff format --check` clean on both touched Python files; ASCII-clean; changelog fragment present per AGENTS.md step 3.

## Verification suggestions

- Independent re-derivation of the transcription: `grep -n 'require_package' src/lerobot/scripts/lerobot_train.py` on a lerobot v0.6.1 checkout - three call sites (`accelerate` unconditional, `peft` on `cfg.peft`, `diffusers` on `cfg.ema.enable`), matching the module table exactly.
- I ran the touched training suite (`tests/training/test_lerobot.py` + the new file) on both trees: branch-only failures none; the shared failures reproduce on pristine `upstream/main` and are the lerobot-version environment artifact the PR description already names.

PR #2406, opened 2026-08-17
fix(hardware): measure a control-loop time budget on a clock that cannot step

cagataycali:
## Problem

Two loops command physical actuators for a number of seconds the caller supplies:
`Robot._execute_task_async` (a policy rollout, bounded by `duration`) and
`TeleopMixin._teleop_loop` (a teleoperation session, bounded by the same parameter). Both
measured that budget against `time.time()`, which is not a clock but the current opinion
about the date: an NTP correction, a `date -s` or a resume from suspend moves it by an
arbitrary amount, and both loops moved with it.

That budget is the only thing bounding a rollout on hardware. `Robot._duration_error`
exists because a value the comparison cannot honor means "the loop commands the servo bus
indefinitely", and `tests/test_hardware_task_duration_guard.py` pins that such a value is
refused "instead of being spent on the arm". Both guard the *value*. A clock step defeats
the same guarantee from outside the value's domain, and neither direction is reported.

Driving the real loops with a stub driver, a 1.0s budget, and a clock that takes one step
mid-rollout:

| rollout, budget 1.0s | seconds commanding the arm | actions applied | reported duration | status |
|---|---|---|---|---|
| no step (control) | 0.980 | 49 | 1.001 | success |
| wall clock steps **+30s** | 0.000 | **1** | **30.021** | success |
| wall clock steps **+1h** | 0.000 | **1** | **3600.021** | success |
| wall clock steps **-2s** | **2.990** | **147** | 1.010 | success |

| teleop session, budget 1.0s | seconds driving the follower | frames applied | reported elapsed | status |
|---|---|---|---|---|
| no step (control) | 1.005 | 50 | 1.005 | success |
| wall clock steps **+30s** | 0.040 | **2** | **30.040** | success |
| wall clock steps **+1h** | 0.040 | **2** | **3600.040** | success |
| wall clock steps **-2s** | **3.016** | **150** | 1.016 | success |

Forward, the rollout is abandoned after its first action and the arm is left parked at that
pose mid-task, while the record claims the task ran for the size of the step. Backward, the
loop keeps commanding the servo bus for the budget *plus* the step - three times the
authorized window here - and the reported duration under-reports it by exactly the step, so
the overrun leaves no trace in the task record. The step's size does not have to be large to
matter, and the `+1h` row shows the truncation is not proportional to it: what is lost is the
remainder of the budget, whatever the step.

## Change

Every *duration* in both loops now reads `time.monotonic()`:

- the rollout bound (`while time.monotonic() - start_mono < duration`) and the `elapsed`
  it reports, which share the same base;
- the live duration `get_task_status()` computes for a running task;
- the teleop session deadline, and the `elapsed_s`/`hz_actual` both status surfaces derive
  from the same base;
- the control loop's `publish_step` telemetry throttle, which is a duration too.

The two bases are renamed for the clock they hold - `RobotTaskState.start_mono` and
`_teleop_start_mono` - because "start_time" is what invited `time.time()` in the first place.
Neither was ever read as a point in time: every reader subtracted it from a later reading to
get an elapsed duration, which is why this is a clock swap rather than a second field.
`_last_stream_pub` starts at `-inf` rather than `0.0`, since a monotonic reading is only
meaningful relative to another one and the first tick should be due wherever a platform's
monotonic epoch happens to sit.

Absent a clock step, nothing changes: the control rows above are unchanged, and two tests
assert exactly that.

This is the boundary the safety subsystem already settled three times -
`tests/mesh/test_replay_cache_monotonic.py`, `tests/mesh/test_bridge_dedup.py::TestDedupClock`,
`tests/mesh/test_corroboration_clock_domain.py`, whose docstring calls monotonic "the same
domain as every other local bookkeeping timestamp". Absolute stamps a reader correlates with
other logs stay on the wall clock and are untouched.

### Scope

`teleoperate(duration=...)` has a second, unrelated way to lose its budget: `_teleop_loop`
computes the deadline and *then* resolves a lazy `strands_robots.mesh.security` import, which
costs 2.17s in a cold process, so a short first session can end before polling the leader at
all. That is a different root cause with a different fix and is left out of this change; the
new tests resolve that import at module scope so it cannot confound the measurement.

## Visual

The real rollout loop, driven with a 1.0s budget while the wall clock is corrected back 2s.
The commands each run sent to the arm are replayed into MuJoCo (headless, `MUJOCO_GL=egl`) -
same script, same policy, same budget; only `strands_robots` differs between the panels.

rollout

[full-resolution MP4](https://raw.githubusercontent.com/cagataycali/robots/media-hw-budget-clock-32051900530/rollout.mp4)

Both panels are the same rollout up to the budget mark (panel-to-panel mean |pixel| 0.84, a
sub-command sampling offset between two real runs). Past it they separate: the BEFORE arm
keeps sweeping (motion after the budget, mean |pixel| vs the first post-budget frame, max
**8.147**) while the AFTER arm holds the pose the last authorized command left it in
(max **0.000**). BEFORE sent 147 commands over 2.998s and reported 1.019s; AFTER sent 49 over
0.986s and reported 1.007s.

## Tests

`tests/test_control_loop_budgets_survive_a_clock_step.py` drives each real loop through a
clock double that takes a known step, and asserts the loop spent the budget it was given,
applied a rollout's worth of commands, and reported the time that actually elapsed. The
contract is pinned on behavior, not on which clock is called.

**14 failed / 7 passed** on `main`, 21 pass after, with messages like:

```
a +30.0s wall-clock step ended the rollout after 0.000s of a 0.4s budget (1 commands)
a -2.0s wall-clock step kept the arm commanded for 2.395s on a 0.4s budget
a 3600.0s wall-clock step made a 0.007s rollout report itself as 3600.006s
a -2.0s wall-clock step made a 2.404s session report itself as 0.404s
a -2.0s wall-clock step stalled the stream for 2.053s
```

Thirteen of those fourteen are behavioral; the fourteenth just names the renamed field. The
seven that pass on both trees are the bound on the change: four assert the clock double
really hands out the step it advertises (so a clean run means something), and two assert the
no-step rollout and session are unchanged - which is what fails if the fix is ever mistaken
for a change in how long a loop runs. Correct code never reads the wall clock at all, so
"the double stepped" is deliberately not used as a precondition for any behavioral assertion.

Four existing tests were updated rather than added to:

- `tests/test_hardware_robot_lifecycle.py` drives the loop to exactly one iteration with a
  scripted clock the policy advances. It scripted `time.time()`, so after this change the
  script no longer reached the loop and the test ran the real budget (40 commands instead of
  2). It now scripts the clock the loop reads - by swapping the module reference the robot
  holds, not the attribute on `time`, because `time.monotonic()` is also asyncio's
  event-loop clock and freezing it would stop the `asyncio.sleep` between two commands from
  ever returning.
- `tests/test_hardware_task_duration_guard.py`, `tests/test_hardware_after_shutdown.py` and
  `tests/test_tool_result_contract.py` quote the comparison or the base by name.

`docs/hardware/robot-control.md` and `docs/hardware/teleoperation.md` state the clock beside
the `duration` they govern.

## Verification

- `mypy strands_robots tests tests_integ`: 19 errors, the same 19 as `main` (none in a
  touched file).
- Blast radius - every test file referencing either module, `RobotTaskState`, either base, or
  a touched docs page, plus all of `tests/mesh` and the repo-wide guards (219 files, 5134
  tests): **21 failed / 5084 passed** here vs **35 failed / 5070 passed** on `main`.
  Comparing failing test ids, **branch-only is empty** (no regressions); the 14 base-only ids
  are exactly this file's pre-fix failures. `5084 - 5070 = 14` and `21 + 14 = 35` reconcile.
  The 45 shared failures are pre-existing (`awsiot` / zenoh not installed here).

### DIFF
--- a/changelog.d/2406-control-loop-budget-monotonic-clock.md
+++ b/changelog.d/2406-control-loop-budget-monotonic-clock.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **hardware**: measure a control-loop time budget on a clock that cannot step. The policy
+  rollout bounded by `Robot.run_policy(duration=...)` / `start_task(duration=...)` and the
+  session bounded by `teleoperate(duration=...)` both compared elapsed wall-clock time
+  against the budget, so an NTP correction, a `date -s` or a resume from suspend moved the
+  bound with it: forward the loop ended early and left the arm parked mid-task, backward it
+  kept commanding the servo bus for the budget plus the step, and the reported duration was
+  wrong in both directions while the task still reported `success`. Every duration in both
+  loops - the bound, the deadline, the elapsed/Hz reports and the `publish_step` throttle -
+  now reads `time.monotonic()`. `RobotTaskState.start_time` is renamed `start_mono` for the
+  clock it holds.
--- a/docs/hardware/robot-control.md
+++ b/docs/hardware/robot-control.md
@@ -68,6 +68,13 @@ themselves apply, so a port the arm accepts is a port the provider can dial.
 of them can honor is reported by name, with the arm still disconnected and the
 command bus still free for a task that could run.
 
+`duration` is measured on a monotonic clock, not on the date. It is an elapsed
+time rather than a point in time, so an NTP correction or a resume from suspend
+cannot cut a rollout short or hold the servo bus past the budget - and the
+`duration` the task reports back is the time that actually elapsed. The same
+holds for `teleoperate(duration=...)`; see
+[Teleoperation](teleoperation.md#mixin-api).
+
 `cleanup()` (and `stop()`, which delegates to it) is terminal: it latches a
 shutdown, releases the task executor, tears down the mesh and ROS bridges, and
 disconnects the robot. It holds whatever state the robot is in - never
--- a/docs/hardware/teleoperation.md
+++ b/docs/hardware/teleoperation.md
@@ -111,6 +111,9 @@ Every hardware `Robot` and `Simulation` host exposes:
 - **`block`** - run inline until `duration` elapses / Ctrl+C (`True`) vs
   background thread (`False`, default).
 - **`duration`** - auto-stop after N seconds (`None` = until stopped).
+  Measured on a monotonic clock, so a wall-clock correction mid-session neither
+  ends it early nor keeps the follower driven past the budget, and the reported
+  `elapsed_s` is the time that actually elapsed.
 
 Each tick: poll every selected device's `get_action()` → apply its `map_fn` →
 **merge** (last-wins on key conflict, with a one-time warning) → check the
--- a/strands_robots/hardware_robot.py
+++ b/strands_robots/hardware_robot.py
@@ -414,7 +414,12 @@ class RobotTaskState:
 
     status: TaskStatus = TaskStatus.IDLE
     instruction: str = ""
-    start_time: float = 0.0
+    #: ``time.monotonic()`` reading taken when the task began. Every reader
+    #: subtracts it from a later reading of the same clock to get an elapsed
+    #: duration, and none of them reports it as a point in time - so it is
+    #: monotonic rather than wall-clock, which a step cannot move. Named for
+    #: its clock so a future reader does not reach for ``time.time()``.
+    start_mono: float = 0.0
     duration: float = 0.0
     step_count: int = 0
     error_message: str = ""
@@ -547,7 +552,11 @@ class Robot(TeleopMixin, AgentTool):
         # Task execution state
         self._task_state = RobotTaskState()
         # Stream-telemetry throttle (publish_step from the control loop).
-        self._last_stream_pub: float = 0.0
+        # A ``time.monotonic()`` reading, like every other elapsed-time base
+        # here. ``-inf`` rather than ``0.0`` because monotonic readings are
+        # only meaningful relative to each other: the first tick is due
+        # regardless of where this platform's monotonic epoch happens to sit.
+        self._last_stream_pub: float = float("-inf")
         # Imported here rather than at module top for the reason every other
         # mesh import in this file is: ``strands_robots.mesh`` pulls the
         # transport package, and a hardware Robot must construct without it.
@@ -1468,7 +1477,7 @@ class Robot(TeleopMixin, AgentTool):
             self._stop_requested.clear()
             self._task_state.status = TaskStatus.CONNECTING
             self._task_state.instruction = instruction
-            self._task_state.start_time = time.time()
+            self._task_state.start_mono = time.monotonic()
             self._task_state.duration = 0.0
             self._task_state.step_count = 0
             self._task_state.error_message = ""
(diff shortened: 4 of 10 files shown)


yinsong1986 (dismissed):
## Summary

Moves the two hardware time budgets — the policy-rollout bound in `Robot._execute_task_async` and the teleop session deadline in `TeleopMixin._teleop_loop` — plus every duration derived from the same bases (live `get_task_status()` duration, `elapsed_s`/`hz_actual`, the `publish_step` throttle) from `time.time()` to `time.monotonic()`, so an NTP correction, `date -s`, or suspend/resume can no longer end a rollout early (arm parked mid-task) or keep the servo bus commanded past the authorized window while the record under-reports it. The renamed bases (`RobotTaskState.start_mono`, `_teleop_start_mono`) have no consumers outside the two touched modules — `RobotTaskState` is not in `__all__`, and the mesh/device-connect readers of `_task_state` touch only `status`/`instruction`/`step_count` — so the rename is internal, not an API break.

## What's good

- The new suite pins behavior (budget spent, commands applied, elapsed reported) rather than which clock is called, and includes no-step controls plus faithfulness tests for the clock double — 21/21 pass on the head, and the four updated existing test files (164 tests) pass as well.
- Consistent clock domain: no `time.time()` reads remain in either touched module outside comments, so there is no mixed-clock comparison left behind.
- `_last_stream_pub` starting at `-inf` is the right base for a monotonic throttle, and the reasoning is documented at the assignment.
- Matches the boundary the mesh subsystem already settled (replay cache, dedup, corroboration clock-domain tests), with absolute wall-clock stamps left untouched.
- Changelog fragment, docs updates beside the `duration` parameters they govern, and the lifecycle test's module-reference clock swap (rather than patching `time.monotonic` globally, which would freeze asyncio's event-loop clock) are all careful, scoped work.

yinsong1986 (approved):
## Summary

Moves the two hardware time budgets - the rollout bound in `Robot._execute_task_async` and the teleop session deadline in `TeleopMixin._teleop_loop` - and every duration derived from the same bases (live `get_task_status()` duration, `elapsed_s`/`hz_actual`, the `publish_step` throttle) from `time.time()` onto `time.monotonic()`, so a wall-clock step can no longer end a rollout early or hold the servo bus past the authorized window while the record misreports the overrun. This is exactly the boundary AGENTS.md already prescribes ("a duration belongs on `time.monotonic()`"); absolute stamps are untouched. I verified the renames are internal: `RobotTaskState`, `start_mono`, `_teleop_start_mono` and `_last_stream_pub` have no readers outside the two touched modules and `RobotTaskState` appears in no `__all__`, so nothing here is a public-API or wire-format change. No `time.time()` read remains in either module outside comments, so no mixed-clock comparison is left behind. The `-inf` throttle base is sound (`now - (-inf)` is always past the period, so the first tick is due), and the head commit resolved the one prior CodeQL finding (unused import in the new test file). The new suite pins behavior across forward/backward steps with no-step controls and clock-double faithfulness checks, and the lifecycle test's module-reference clock swap correctly avoids freezing asyncio's event-loop clock. No blockers found.

PR #2407, opened 2026-08-17
fix(simulation): declare the is_static default that add_object delivers

cagataycali:
## Problem

`SimEngine.add_object` declared `is_static: bool = False`. The default backend does not deliver that value, and cannot. Measured against a real compiled model, one `add_object` per case:

| call | status | delivered `is_static` |
|---|---|---|
| `add_object(name="ground", shape="plane")` | `success` | **`True`** |
| `add_object(name="ground", shape="plane", is_static=False)` | **`error`** | nothing is built |
| `add_object(name="ground", shape="plane", is_static=True)` | `success` | `True` |

So a caller who omitted the argument got the opposite of the declared default, and a caller who read the base signature and wrote out its declared default got a hard refusal - for the one shape whose entire purpose is being static.

A plane is infinite and cannot carry a dynamic mass, so MuJoCo needs to tell "the caller did not specify" apart from "the caller said `False`", and its signature already did: `is_static: bool | None = None`. Only the base contract, Newton and Isaac declared `bool = False`.

## Why the default cannot be `False`

`False` is also what a caller passes to ask for a dynamic body. A backend reading a `bool` therefore has no way to distinguish the two, and cannot derive the answer from `shape` without overriding an explicit request. `None` is the only value that can mean "unspecified".

Three surfaces already stated this while the signature did not:

- `MuJoCoSimEngine.add_object`'s own comment: *"planes are infinite and must be static. Explicit is_static=False for a plane is an error; None or True both resolve to True."*
- `SimEngine.describe()` already advertised `is_static=None` - in the same file whose signature declared `False`.
- The README footgun list already says *"`add_object(shape="plane")` auto-sets `is_static=True`; passing `is_static=False` is a hard error."*

## Change

- `is_static` is declared `bool | None = None` on `SimEngine` and on all three backends, and the base docstring states what the sentinel means: a backend MAY derive the answer from `shape`, and one with no shape-derived rule resolves `None` to `False`.
- Newton and Isaac resolve the sentinel to `False` before any gate reads it, so the mass check still runs for a dynamic body and the recorded object state carries a real `bool` rather than the sentinel. Neither backend special-cases any shape - neither supports `shape="plane"` at all - so nothing about their behaviour changes.
- `tool_spec.json` published `is_static` as a bare `{"type": "boolean"}` with **no description at all**, so the shape-derived rule was unreachable from the agent surface even though the neighbouring `mass` description already refers to `is_static=true`. It now carries one.
- The `docs/simulation/overview.md` API table stated `is_static=False` in the same cell that said `plane` requires `True`.

## Scope: what does not change

No backend's behaviour changes. In particular a plane handed an explicit `is_static=False` is **still refused**, with `is_static=True` named as the remedy. Making the declared default deliverable by having a plane silently accept `False` would satisfy the headline test and lose the refusal - the caller would be told a dynamic plane was built and get a static one - so that is pinned as a test rather than left to reading.

## Tests

`tests/simulation/test_add_object_static_default_contract.py`, copied onto `upstream/main` for the pre-fix run: **7 failed / 13 passed**, then 20 passed. Every failure is a measurement, not a wording diff:

```
test_omitting_and_restating_the_default_agree[plane]
  AssertionError: shape='plane': omitting is_static gave 'success' but passing the
  base-declared default is_static=False gave 'error' (add_object: shape='plane'
  requires is_static=True (planes are infinite and cannot have dynamic mass).)

test_describe_advertises_the_default_the_signature_declares[base]
  AssertionError: base.describe() publishes a default the signature does not declare:
  add_object.is_static: describe() says 'None', signature says False

test_no_override_narrows_a_base_declared_default[mujoco]
  AssertionError: add_object(is_static=): base declares False, mujoco declares None
```

The 13 that pass on both trees are the controls: the plane refusal survives and still names the remedy (parsed back out of the message and applied), an unspecified plane is static while an unspecified box is not, the resolved flag is a `bool`, and the non-plane shapes agree on both spellings.

Two structural guards keep it that way. `describe()` and the signature both publish defaults for the same parameters and nothing compared them; the new guard grades every `describe()` string that states a *literal* default against the real signature, on all four surfaces, and skips a parameter documented by its domain (`state='open'|'close'`, `position=[x,y,z]`) because that is not a default claim. The second guard pins that no override changes a value the base already declared, while leaving a base-required parameter free to be widened.

`mypy` found the rest of the blast radius on its own: three minimal `SimEngine` stand-ins in the test tree also declared `is_static: bool`, which the widened base contract makes narrowing overrides. Widened with it.

## Verification

- `tests/simulation` plus every test file referencing `is_static` / `add_object` / `describe()` / `SimEngine`, plus the repo-wide docstring, ASCII and changelog guards: 262 files, identical selection on both trees. Branch **10755 passed / 6 failed**; base **10748 passed / 13 failed**. `comm` of the sorted failing ids gives an empty branch-only set; the base-only set is exactly the 7 pre-fix failures above, and the 6 remaining failures are pre-existing on both trees (all in `tests_integ/`, which is outside `testpaths`). The counts reconcile: `10755 + 6 == 10748 + 13 == 10761` collected, and `13 - 6 == 10755 - 10748 == 7`.
- `ruff check` / `ruff format --check` / `mypy` over `strands_robots tests tests_integ`: clean, and mypy's error set is byte-identical to `upstream/main` (19 = 19).

## Artifact

One script per tree. It reads the declared default from `inspect.signature(SimEngine.add_object)` and passes that value back in, so the panels differ only in what the contract declares. A shelf plane is added at z=0.50 and a 1 kg crate is dropped from z=0.86.

is_static declared default

Middle column, the caller restating the declared default:

| `upstream/main` (declared `False`) | this change (declared `None`) |
|---|---|
| refused | built |
| shelf refused, crate falls to the ground: `crate_z = 0.0699` | shelf built, crate rests on it: `crate_z = 0.5699` |

Panel deltas: on `main`, omitted vs restated is `mean|d| = 8.40181`; after the change it is `0.00003`. The same case rendered under the two trees differs by `0.00004` - the change is a no-op everywhere except the spelling that was refused.


yinsong1986 (approved):
## Summary

Aligns `add_object`'s declared `is_static` default with what the backends actually deliver: the base contract and the Newton/Isaac overrides move from `bool = False` to `bool | None = None` (the MuJoCo backend already declared the sentinel), with the sentinel's meaning stated on the ABC. Newton and Isaac resolve `None` to `False` before any gate reads it - verified that the resolution precedes Newton's mass check and Isaac's `_validate_mass` call, so the mass gate still runs for dynamic bodies and the recorded `SimObject.is_static` stays a real `bool`. The plane refusal on an explicit `is_static=False` is unchanged and pinned by a dedicated test, so no backend's behaviour changes; this is a contract alignment, not a behaviour change, and it is backward compatible for every existing caller (any explicit `bool` still means what it meant). The previously description-less `tool_spec.json` `is_static` entry now carries a description, closing the agent-surface gap AGENTS.md rule 13 exists for.

## What's good

- The obvious wrong fix (silently accepting `is_static=False` for a plane) is explicitly ruled out by `TestTheRefusalSurvives`, and the remedy in the refusal message is parsed back out and applied rather than string-matched alone.
- The two structural guards (`describe()` literal defaults vs. real signatures, and no override narrowing a base-declared default) close the drift class rather than the instance, with premise assertions (`graded >= 10`) so the extractors can't silently stop reaching anything.
- Changelog fragment, docs table, and the three test-tree `SimEngine` stand-ins all updated in the same PR; scope is tight and every file in the diff serves the one contract.

PR #1722, opened 2026-07-29
refactor(mesh): MobileBaseRobot + Transport - one drive contract for every transport

Vivek0712:
Follow-up to the design conversation on #1035 and #1110, in the order @cagataycali and I agreed: **port `RosBridgedRobot` + `RtpsRobot` onto the base first, with their existing tests as the acceptance bar**. #1035 and #1110 rebase onto this afterwards, as separate PRs.

## Why

Four robot classes, ~80% identical, and the duplication was not accidental - each was written by mirroring the last. From your table on #1110:

| Class | Transport | Kinematics |
|---|---|---|
| `RosBridgedRobot` | `use_ros` | diff-drive |
| `RtpsRobot` | `use_rtps` | diff-drive |
| `RosbridgeRobot` (#1110) | `use_rosbridge` | diff-drive |
| `AckermannRosRobot` (#1035) | `use_ros` | Ackermann |

The cost was not just line count. It is why `RosBridgedRobot` - the class most likely to be driving real ROS 2 hardware - shipped **without** the NaN guard, the duration bound and the trailing-zero stop that its newer siblings had. You flagged exactly this on #1110: *"a `RosBridgedRobot.drive(linear=float('nan'))` publishes NaN to `cmd_vel` today."* It does not anymore, and it cannot regress, because the guard is not in `RosBridgedRobot` at all.

## What

`strands_robots/mesh/_mobile_base.py`: `MobileBaseRobot` owns validation, the `drive`/`stop` contract and its safety semantics, the `init_services` handshake, `get_pose`/`get_scan`, and the `tools` property. Subclasses supply only what varies, through two seams - a `Transport` (how bytes move) and `_cmd_fields` (what the message looks like: the kinematics seam).

Both ports are now thin: `ros_bridge.py` 296 -> 244 lines (and most of what remains is the Nav2 surface, which is genuinely ROS 2-only), `rtps_robot.py` 156 -> 148. The *shape* is what shrank - `RtpsRobot` no longer contains a drive contract at all.

**The three amendments from my review comments are all in:**

1. **`Transport` requires `publish` + `echo` only.** `service_call` and `action_send_goal` are separate `ServiceCapable` / `ActionCapable` protocols. `use_rtps` has neither, and a protocol that forced it to declare them would make it lie. The base asks - `robot.supports("service_call")` - and reflects the answer: `init_services` on a service-less transport is refused **at construction**, not on the track.

[...]

PR #2408, opened 2026-08-17
fix(mesh): measure peer liveness and publish intervals on a clock that cannot step

cagataycali:
## Problem

The peer registry decides whether a robot is still on the mesh. `PeerInfo.last_seen` was
stamped from `time.time()`, and both readers subtracted it from a later reading of the same
clock: `PeerInfo.age`, and the `now - last_seen > timeout` in `prune_peers`. `time.time()` is
not a clock but the current opinion about the date - an NTP correction, a `date -s`, a resume
from suspend moves it by an arbitrary amount - so every peer's apparent age moved with it.

`Mesh._heartbeat_loop` calls `prune_peers()` on **every** tick at `HEARTBEAT_HZ = 2.0`, and
`PEER_TIMEOUT` is `10.0`. Driving the real `update_peer` / `prune_peers` / `PeerInfo.age`
against a clock double, with three peers that are all 5s into that 10s timeout and all
heartbeating normally:

| wall clock | peers surviving the prune | pruned | `arm-1` reported age |
|---|---|---|---|
| no step (control) | 3/3 | - | 5.0 |
| steps **+30s** | **0/3** | arm-1, arm-2, base-1 | gone |
| steps **+1h** | **0/3** | arm-1, arm-2, base-1 | gone |
| steps **-30s** | 3/3 | - | **-25.0** |

Forward, the next heartbeat tick deletes the entire fleet while every peer is still
broadcasting presence. The step does not have to be large - anything past `PEER_TIMEOUT` does
it - and the loss is not proportional to the step: `+30s` and `+1h` are the same outcome.

Backward is the direction that matters more, because it fails toward reporting a robot that
is gone as present. The same peer, silent for 15s - past the 10s timeout, so genuinely dead:

| wall clock | pruned? | reported age |
|---|---|---|
| no step (control) | pruned | - |
| steps **-30s** | **retained as live** | -15.0 |
| steps **-1h** | **retained as live** | -3585.0 |

`get_peers()` / `peer_count()` go on advertising it for the size of the step, and `age` - the
field `to_dict()` hands an operator and an agent - is reported negative.

The same step also inverts the phantom-peer flood guard. `update_peer` bounds a flood by
evicting the peer whose heartbeat is oldest (`min(..., key=last_seen)`); peers stamped either
side of a step are ordered by the step rather than by age:

| wall clock | evicted when a 4th peer arrives at `STRANDS_MESH_MAX_PEERS=3` | correct? |
|---|---|---|
| no step (control) | `oldest` | yes |
| steps **-30s** | **`newest`** | no |

Two further durations in the same subsystem read the same clock: the LiDAR **state** publish
interval in `SensorLoopsMixin._lidar_loop` (`now - last_state_publish >= state_period`), and
the `elapsed` behind `hz_actual` in `InputPublisher.stats` and `InputReceiver.stats`.

## Change

Every duration the mesh measures for itself now reads `time.monotonic()`:

- the peer age and the prune comparison, which share a base;
- the cap-eviction ordering;
- the LiDAR state publish interval;
- the `elapsed` both input `stats` surfaces derive `hz_actual` from.

The bases are renamed for the clock they hold - `PeerInfo.last_seen_mono`,
`InputPublisher._start_mono`, `InputReceiver._start_mono` - because a name that reads as a
point in time is what invited `time.time()`. None was ever read as a point in time: every
reader subtracted it from a later reading to get a duration, which is why this is a clock swap
rather than a second field. `PeerInfo` is internal (absent from `strands_robots/mesh/__init__.py`
and from the top-level `__all__`), and `to_dict()` reports `age` and never the base, so no
serialised or exported surface changes. The `_mono` suffix is the convention already in this
module (`InputReceiver._last_rate_gate_mono`) and in the safety subsystem (`_last_estop_mono`).

`last_state_publish_mono` starts at `-inf` rather than `0.0`, since a monotonic reading is only
meaningful relative to another one and the first tick should be due wherever a platform's
monotonic epoch happens to sit.

Absent a clock step nothing changes: every control row above is identical before and after, and
four tests assert exactly that.

**Absolute stamps are deliberately untouched.** A published envelope's `"t"` field
(`sensors.py`, nine sites) and the `_age = time.time() - float(_frame_t)` an input receiver
computes against a *remote* peer's `"t"` name a point in time that something off this process
correlates, and seconds of local process uptime is meaningless to that reader. After this
change every remaining `time.time()` in the three touched modules is one of those two.

This is the boundary the safety subsystem already settled three times -
`tests/mesh/test_replay_cache_monotonic.py`, `tests/mesh/test_bridge_dedup.py::TestDedupClock`,
`tests/mesh/test_corroboration_clock_domain.py`, whose docstring calls monotonic "the same
domain as every other local bookkeeping timestamp".

### Scope

`PeerInfo.last_seen_mono` keeps its `0.0` default. It means "never seen", no production path
reaches it (`update_peer` always stamps, and every construction in the tree passes the field
explicitly), and changing the sentinel is a different concern from the clock it is read on.

Wall-clock durations remain outside the mesh - `simulation/policy_runner.py` bounds a sim
rollout on `time.time()`, and `policies/lerobot_local/policy.py` reports inference latency the
same way. Those are a different subsystem with a different consequence (no servo bus, no peer
liveness) and are tracked separately rather than folded in here.

## Tests

`tests/mesh/test_mesh_durations_survive_a_clock_step.py` drives each real surface through a
clock double that takes a known step, and asserts on what the mesh *does*: which peers survive
a prune, what age is reported, which peer the cap evicts, when state is published, what rate is
reported. The contract is pinned on behaviour, not on which clock is called, so a future
refactor cannot satisfy it by renaming a call.

**17 failed / 10 passed** on `main`, 27 pass after:

```
a +30.0s wall-clock step pruned live peers: ['arm-1', 'arm-2', 'base-1']
a -30.0s wall-clock step kept a peer silent for 15.0s in the registry as live
a 30.0s wall-clock step misreported the age
a -30.0s wall-clock step evicted ['newest'] rather than the oldest peer
a -30.0s wall-clock step published lidar state at [0.0, 1.0] rather than [0.0, 1.0, 2.0]
a 30.0s wall-clock step misreported the achieved rate
```

The ten that pass on both trees are the bound on the change: five assert the clock double
really hands out the step it advertises and that monotonic does not move with it (so a clean
run means something), four are no-step controls across all three surfaces, and one guards the
fixture that stubs the publisher's loop body. The LiDAR assertions pin *when* state was
published rather than how often, because a count alone came out right for the wrong reason on
a forward step - three publishes, at `[0.0, 0.2, 1.2]` instead of `[0.0, 1.0, 2.0]`.

Three existing tests were updated rather than added to, all of which backdate a peer and so
have to do it on the clock the registry now measures against:
`tests/mesh/test_mesh_session.py`, `tests/mesh/test_deep_mesh.py`, and
`tests/test_repr_survives_partial_construction.py` (one line, the renamed kwarg).

`docs/mesh.md` states the split beside the fleet-clock guidance it sits next to, because that
section tells an operator to correct a robot's clock and a reader should not conclude that
doing so drops the fleet.

## Verification

- `ruff check` / `ruff format --check` over `strands_robots tests`: clean.
- `mypy strands_robots tests`: 57 errors, the same 57 as `main`, none in a touched file.
- Blast radius - all of `tests/mesh` plus every test file naming a touched symbol
  (178 files): **113 failed / 3196 passed** here vs **113 failed / 3169 passed** on `main`.
  Comparing failing test ids, **branch-only is empty** (no regressions) and the 139 shared
  failures are pre-existing (`awsiot` / zenoh not installed in this environment).
  `3196 - 3169 = 27` is exactly the new suite, and it reconciles.

Read as a delta rather than an absolute: the 139 pre-existing failures are an artefact of a
partial environment, not of this change.

## Review rounds

### Known overlap

#1722 also edits `tests/test_repr_survives_partial_construction.py` (my change there is the
one renamed kwarg in a parametrize list). #1722 is currently `CONFLICTING` and needs a rebase
regardless; whichever lands second takes the one-line resolution.


yinsong1986 (approved):
## Summary

Moves every mesh-local duration onto `time.monotonic()`: the peer registry's liveness decision (`PeerInfo.age`, the `prune_peers` timeout comparison, and the `STRANDS_MESH_MAX_PEERS` cap-eviction ordering), the LiDAR state publish interval, and the `elapsed` behind `hz_actual` in both input `stats` surfaces. Absolute stamps (`"t"` envelope fields, the receiver's age against a remote peer's `t`) are deliberately left on the wall clock, which is the correct side of the AGENTS.md clock split ("a duration is measured, a stamp is recorded"). I verified the rename is fully internal: no remaining `last_seen` or `_start_time` references anywhere in the tree, `PeerInfo` is absent from the mesh `__all__`, and `to_dict()` serialises only `age` - so no exported or wire surface changes. Reproduced the claimed test delta locally: the new suite reports 17 failed / 10 passed with the three touched modules reverted to `main`, and 114 passed across the new suite plus the updated `test_mesh_session.py` / `test_repr_survives_partial_construction.py` on the head. No must-fix concerns.

## What's good

- Regression pinned on behaviour (which peers survive, when state publishes, what rate is reported) rather than on which clock function is called, so a refactor cannot satisfy it by renaming a call.
- The clock double is itself tested (step moves only the wall clock), and no-step controls bound the change on every surface.
- Changelog fragment, `docs/mesh.md` note beside the fleet-clock guidance, and the `_mono` naming convention already used by the safety subsystem are all in place.
- `-inf` init for `last_state_publish_mono` is right: a `0.0` init on monotonic could delay the first publish on platforms whose monotonic epoch sits near boot.

PR #2410, opened 2026-08-17
fix(simulation): name the criterion, episode and step when an eval outcome criterion raises

cagataycali:
## Problem

Both eval routes call a caller-supplied **episode-outcome criterion** after every applied action:

- `eval_policy` / `PolicyRunner.evaluate` -> `success_fn(observation)`
- `evaluate_benchmark` / `PolicyRunner._evaluate_with_spec` -> `spec.is_success(sim)` / `spec.is_failure(sim)`

Every *other* per-step hook on those same paths already states what a raise means:

| per-step hook | a raise is... | stated in |
| --- | --- | --- |
| `on_frame` (generic) | best-effort telemetry: logged at WARN, the eval continues | `_fire_on_frame` |
| `on_frame` -> `RecordingFrameError` | data loss: propagates to the caller | `_fire_on_frame` |
| `spec.on_step` | `status="error"` naming the hook and the spec | `_evaluate_with_spec` |
| `run(stop_when=...)` | fatal, with a message naming the step | `_stop_when_fired` |
| **the outcome criterion** | **unspecified** | 7 bare call sites |

The one hook that decides the evaluation's headline `success_rate` was the only one with no policy, so it arrived as the caller's own bare exception from inside a nested loop. Measured on `so101` in MuJoCo over 3 episodes, with a criterion that indexes the engine result as if it were the payload - the mistake `_resolve_success_fn`'s own comment records having been bitten by ("the inline copy this replaces indexed the engine result as if it were the payload, so it never saw a real backend's envelope and scored every episode a failure"):

```
eval_policy(robot_name="so101", n_episodes=3, success_fn=displaced)
  -> KeyError: 'position'
```

Nothing names the criterion, the episode or the step. A criterion that only breaks on episode 7 of 10 therefore costs the whole evaluation and says nothing about why, and the caller cannot tell a broken criterion apart from a broken policy or a broken scene.

`stop_when` already argues the right posture for exactly this case: *"the caller asked for an early-return semantics the runner can no longer honor, and silently running to the step budget would misreport the rollout."* A `success_rate` averaged over episodes whose outcome was never determined misreports an evaluation the same way - so the criterion failure should be fatal, and it should say what broke.

## Change

One helper, `_criterion_verdict(check, subject, *, label, episode, step)`, mirroring `_stop_when_fired`: it coerces the verdict with `bool()` and converts a raise into a `RuntimeError` that names the criterion, the episode and the step and chains the original. All seven criterion invocations route through it (3 x `success_fn`, 2 x `spec.is_success`, 2 x `spec.is_failure`).

```
RuntimeError: success_fn raised at episode 0, step 1: KeyError('position'). The
episode-outcome criterion cannot be evaluated, so the evaluation is aborted
rather than reporting a success_rate over episodes whose outcome was never
determined.
```

Two deliberate scope boundaries:

- **The posture is unchanged - this is a message fix.** `evaluate` surfaces a rollout failure as a raise by design: a raising `get_actions` reaches the caller that way, and `test_recording_frame_loss_is_not_tolerated` pins a `RecordingFrameError` doing the same ("it surfaces as a raise rather than an error envelope because that is how `eval_policy` already surfaces every rollout failure"). Wrapping these loops in a terminal `except Exception -> status="error"` handler like `run`'s would change that pinned contract *and* swallow the recording-frame propagation with it, so it is not done here; a control test pins that propagation.
- **Verdicts are coerced, not type-checked.** `observation["x"] > 0.5` returns `numpy.bool_`, which is *not* an instance of `bool`, so an `isinstance` gate would reject the most ordinary predicate. `bool()` is what `stop_when` already does, so truthiness is unchanged - including the case where a criterion returns a non-bool without raising (`bool({"success": False})` is `True`). Narrowing *that* is a separate question about what a verdict may be, not about what a failure means.

## Verification

`tests/simulation/test_eval_outcome_criterion_exception_contract.py`, real MuJoCo sim, hermetic inline MJCF:

| | on `upstream/main` | with this change |
| --- | --- | --- |
| 5 contract tests | **fail** (`KeyError: 'cube'` / `ValueError: no body 'mug'` arrive unnamed) | pass |
| 8 control tests | pass | pass |

The 8 controls are the boundary: a `True` / `False` / `numpy.bool_` verdict still measures, an absent criterion still reports `success_measured=False`, `spec.on_step` keeps its own wording and its own envelope, a generic `on_frame` failure stays best-effort, a `CooperativeStop` still ends the eval cleanly, and **a `RecordingFrameError` still propagates** - the last one added after an earlier revision of this change was caught swallowing it.

Wider gate, `tests/simulation` + `tests/benchmarks` (11935 collected) on both trees:

| | failing ids | passed |
| --- | --- | --- |
| `upstream/main` (this test file copied in) | 7 = 2 pre-existing + the 5 contract tests | 11898 |
| with this change | **2 = exactly the same 2 pre-existing** | 11904 |

The two are environment-bound and unrelated (`newton/test_multi_camera.py::test_bad_position_shape_rejected` asserts older refusal wording; `libero/test_libero_adapter.py::TestInstallActionController` is order-dependent). `ruff check` / `ruff format` clean, and `mypy strands_robots tests tests_integ` reports the same 14 pre-existing errors on both trees.

## Artifact

One script, two trees, nothing else differing: a caller makes the envelope mistake, then tries to self-correct from whatever `eval_policy` raises. Both trees raise; what differs is whether the raise says anything.

eval outcome-criterion contract

On `main` the exception is an anonymous `KeyError: 'position'`, so there is nothing to act on and the evaluation is abandoned. With this change the caller reads `success_fn raised at episode 0, step 1: KeyError('position')`, fixes the indexing, and the re-run reports **1/3 = 0.3333 over avg 28/40 steps with `success_measured=true`**, writing 3 per-episode MP4s (23/23 inter-frame deltas non-zero). The strip is episode 1 of that evaluation - [the MP4](https://raw.githubusercontent.com/cagataycali/robots/media-eval-criterion-contract/eval-episode1.mp4).


yinsong1986 (approved):
## Summary

This PR gives the episode-outcome criterion the raise-policy it was missing: all seven criterion call sites (3x `success_fn`, 2x `spec.is_success`, 2x `spec.is_failure`) route through a new `_criterion_verdict` helper that coerces the verdict with `bool()` and converts a raise into a `RuntimeError` naming the criterion, the episode and the step, chaining the original. The current head deliberately does *not* add a terminal handler to `evaluate` / `_evaluate_with_spec` - the named failure propagates, matching how a raising `get_actions` or a `RecordingFrameError` already reaches the caller on those methods - and the agent-tool contract is still satisfied one layer up, at `MuJoCoSimEngine.stream`'s catch-all, which converts any escaping exception into `status="error"` before it reaches the model. I verified the mechanics against the head tree: the helper is the only change on the hot path, `CooperativeStop` (a `BaseException`) is untouched by the helper's `except Exception`, the documented direct-caller `ValueError` guards (`action_horizon`, `n_episodes`, `max_steps`) run before the loops and keep their surface, and the docs/changelog were rewritten in the same commit to state the propagate-by-design posture rather than the earlier envelope one.

## What's good

- `bool()` coercion rather than an `isinstance` gate keeps the `numpy.bool_` an ordinary predicate returns working, and the test pins the premise (`not isinstance(np.bool_(True), bool)`) rather than assuming it.
- The control tests pin the neighbouring policies precisely: `spec.on_step` keeps its own error envelope, a generic `on_frame` failure stays best-effort, `RecordingFrameError` still propagates (the test that justifies not wrapping the loop in a terminal handler), `CooperativeStop` still ends the eval cleanly, and an absent criterion still reports `success_measured=False`.
- Changelog fragment via `changelog.d/` per AGENTS.md step 3, docs updated in the same PR, and the helper stays private (not exported).

yinsong1986 (approved):
## Summary

This PR closes the one gap in the eval hooks' exception contract: the episode-outcome criterion (`eval_policy`'s `success_fn`, a benchmark spec's `is_success` / `is_failure`) previously surfaced a raise as the caller's own bare exception from inside a nested loop, naming neither the criterion, the episode nor the step. All seven criterion call sites (3x `success_fn`, 2x `spec.is_success`, 2x `spec.is_failure`) now route through a single `_criterion_verdict` helper that coerces the verdict with `bool()` and converts a raise into a `RuntimeError` naming all three and chaining the original. The helper mirrors `_stop_when_fired`'s existing pattern byte-for-byte in posture (wrap-and-rechain, not warn-and-continue; `CooperativeStop` is a `BaseException` and passes through untouched), and the propagate-vs-envelope surface of each eval method is deliberately unchanged. Verified independently on the head SHA: the full 13-test contract file passes against a real MuJoCo sim, `ruff check` / `ruff format` are clean on the changed files, and a grep confirms no bare criterion invocation remains in the eval loops.

## What's good

- Scope discipline: message fix only, with the two boundaries (no terminal handler, `bool()` coercion over `isinstance`) argued in the description and pinned by controls - including the `RecordingFrameError` propagation control added after an earlier revision was caught swallowing it.
- Contract tests use a hermetic inline MJCF under `tmp_path`, no dataset recording (so the shared-cache rule doesn't apply), no host paths, plain ASCII throughout.
- Changelog fragment present (`changelog.d/2410-...`) per AGENTS.md step 3, and `docs/simulation/overview.md` documents the new posture alongside its siblings.

PR #2411, opened 2026-08-17
docs(policies/protomotions): name the frame the reference-motion cache's velocity channels are in

cagataycali:
## Problem

`MotionPlayer` accepts a hand-built cache dict as a first-class input mode, so the frame of its `body_vel` / `body_ang_vel` rows is a contract a caller has to be able to read. Two things were wrong.

**A private helper named the wrong frame.** `_quat_finite_diff_ang_vel` documented its return as `local-frame angular velocities`. It left-multiplies the quaternion delta (`q1 * q0^-1`), which is the *world*-frame quantity. Driven on an analytic world-Z spin of a body tilted a quarter turn about X - a pose chosen so the two frames differ - it matches the world value exactly and misses the body-local one by `0.57 rad/s`:

```
analytic omega_world : [0. 0. 0.4]     err vs WORLD: 0.000000
analytic omega_local : [0. 0.4 0. ]    err vs LOCAL: 0.565685
bridge output        : [-0. 0. 0.4]
```

Every other surface in the package already names that channel world-frame, so the helper was the odd one out:

| surface | how it names the per-body angular-velocity array |
| --- | --- |
| `state_utils.compute_root_local_ang_vel` arg | `rigid_body_ang_vel: Shape [num_bodies, 3] (world frame)` |
| `policy.py` observation key | `body_ang_vel_world` |
| `bridge._quat_finite_diff_ang_vel` Returns | `local-frame angular velocities` |

`compute_root_local_ang_vel` exists precisely to rotate one row of that array into a body's local frame, so a producer that emits local-frame rows is rotated a second time rather than used as-is.

**The public contract named no frame at all.** Neither `qpos_to_motion_data`'s `Returns:` section nor the cache-format list in `motion_utils` said which frame `body_vel` / `body_ang_vel` are in, so a caller assembling a cache by hand had nothing to follow.

This is not a cosmetic distinction. Bridging a Kimodo G1 walk clip (120 frames @ 30 Hz) through `qpos_to_motion_data` and reading each row in both frames:

| quantity | peak world-vs-local gap |
| --- | --- |
| body `left_ankle_pitch_link` | **6.015 rad/s** |
| worst over all 30 bodies | 6.015 rad/s |
| pelvis alone | 0.329 rad/s |

## Change

Documentation and tests only - no behaviour changes.

- `bridge._quat_finite_diff_ang_vel`: `Returns:` now says world-frame, and the body explains *why* (left- vs right-multiplication) and points at `compute_root_local_ang_vel` for the local conversion.
- `bridge.qpos_to_motion_data`: `Returns:` names the frame of the pose and velocity channels it builds.
- `motion_utils` cache-format contract: annotates all four body channels and states what a hand-built cache must follow.
- `docs/policies/protomotions.md`: a "Cache velocities are world-frame" subsection, next to the existing anchor-vs-base discussion of the same class of wrong-frame error.

## Tests

`tests/policies/protomotions/test_cache_velocity_frames.py`, 10 tests in two complementary classes: one pins **which** frame the code produces, the other pins that **every claim names that frame**. Neither alone is sufficient, and together they close both drift directions.

Pre-fix split on `upstream/main`: **4 failed, 6 passed**. The failures name the offender and the measured frame:

```
the per-body angular-velocity array is measurably world-frame, but 1 documented
entr(y/ies) name another frame: bridge.py:_quat_finite_diff_ang_vel Returns:
'Shape ``[T, num_bodies, 3]`` local-frame angular velocities.'

qpos_to_motion_data builds the cache's velocity channels, so its Returns section
has to say which frame they are in; it names none
```

The 6 that pass on both trees are the behaviour controls, and they are load-bearing rather than decoration: flipping the implementation to genuinely right-multiply (`q0^-1 * q1`, i.e. actually local-frame) fails **7** of the 10, including all three analytic controls and both end-to-end angular-velocity checks. The root-cause guard is deliberately consistency-based, so it correctly stays green in that scenario while the behaviour class catches it - which is why both classes exist.

The end-to-end class drives the real `qpos_to_motion_data` on a hermetic MJCF generated in the test (a freejoint root plus 29 hinges, so `nq` is exactly 36), so it downloads nothing and skips cleanly without `mujoco`. It also covers the resample branch, since that path re-interpolates the velocity rows.

## Verification

- `mypy strands_robots tests tests_integ`: 19 pre-existing errors, identical set on this branch and on `upstream/main`; zero in the touched files.
- Blast radius - all `tests/policies/protomotions`, every test referencing the changed symbols, and the repo-wide docstring / xref / ASCII / docs-coverage / changelog guards (19 files): **branch 690 passed, 0 failed**; base with the new file copied in **4 failed, 686 passed**. Same 691 collected on both, no branch-only failures, and the base-only failures are exactly the 4 pre-fix ones.

## Artifact

The clip whose bridged cache the measurement above comes from, rendered headless in MuJoCo, beside the two frames of the same physical rotation for the worst-diverging body and the full claim-vs-measurement table.

reference-motion cache velocity frames

### DIFF
--- a/changelog.d/2411-protomotions-cache-velocity-frames.md
+++ b/changelog.d/2411-protomotions-cache-velocity-frames.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **policies/protomotions**: name the frame the reference-motion cache's velocity
+  channels are in. `_quat_finite_diff_ang_vel` documented its return as a
+  local-frame angular velocity while left-multiplying the quaternion delta, which
+  is the world-frame quantity that every other surface here already expects -
+  `compute_root_local_ang_vel` takes a world-frame `rigid_body_ang_vel` and the
+  policy reads a `body_ang_vel_world` observation key. Neither
+  `qpos_to_motion_data`'s `Returns:` section nor the `MotionPlayer` cache-format
+  contract named a frame at all, so a caller building a cache by hand had nothing
+  to follow; local-frame rows stored there are rotated a second time rather than
+  used as-is. On a bridged G1 walk clip the two frames differ by up to 6 rad/s.
--- a/docs/policies/protomotions.md
+++ b/docs/policies/protomotions.md
@@ -112,6 +112,18 @@ cache["num_frames"], cache["control_dt"]
 `MotionPlayer` accepts that dict, an `.npz` written by
 `MotionPlayer.save_cache_npz`, or a raw ProtoMotions `.pt`.
 
+### Cache velocities are world-frame
+
+`body_pos` and `body_rot` are world poses, and `body_vel` and `body_ang_vel`
+are **world-frame** velocities derived from them - the same convention a raw
+ProtoMotions motion library uses for `rigid_body_ang_vel`. That matters when
+you hand-build a cache instead of bridging one: the tracker's root input is a
+*local*-frame angular velocity, and `compute_root_local_ang_vel` produces it by
+rotating a world-frame row into the root's frame. Storing local-frame rows here
+gets them rotated a second time. The frames are not interchangeable - on a
+walking G1 clip they differ by whole rad/s, the same class of silent wrong-frame
+error as substituting the base for the anchor link above.
+
 ### Motion files are read with a restricted unpickler
 
 An `.npz` cache is read by NumPy and needs no torch. A raw `.pt` is read with
--- a/strands_robots/policies/protomotions/bridge.py
+++ b/strands_robots/policies/protomotions/bridge.py
@@ -105,12 +105,22 @@ def _quat_finite_diff_ang_vel(quats_xyzw: np.ndarray, dt: float) -> np.ndarray:
     then keeps the vector part. Output copies the last frame from the
     second-to-last so shapes match the input trajectory.
 
+    The delta is left-multiplied (``q_{t+1} * q_t^-1``), so the result is a
+    WORLD-frame angular velocity, matching the frame every other surface here
+    uses for a per-body angular-velocity array: the ``rigid_body_ang_vel``
+    argument of
+    :func:`~strands_robots.policies.protomotions.state_utils.compute_root_local_ang_vel`,
+    which exists to rotate one body's row of it into that body's local frame.
+    Right-multiplying instead (``q_t^-1 * q_{t+1}``) would give the local-frame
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Documentation-and-tests-only correction of a wrong frame claim in the ProtoMotions reference-motion cache contract. `_quat_finite_diff_ang_vel` left-multiplies the quaternion delta (`q1 * q0^-1`), which is the world-frame angular velocity, while its `Returns:` said local-frame; the fix flips that word, adds the why (left- vs right-multiplication) to the helper body, names the frame in `qpos_to_motion_data`'s `Returns:` and in the `MotionPlayer` cache-format contract, and adds a matching docs subsection. I verified the production diff is docstring-only (no executable lines change), and the math claim is correct - the left-multiplied delta of `q_{t+1} = exp(w_world*dt/2) * q_t` is the world-frame quantity, which `compute_root_local_ang_vel` then rotates into the root frame exactly as the pair-contract test asserts.

## What's good

- The pre-fix split reproduces exactly as described: the new test file on the base commit reports **4 failed, 6 passed**, and all 10 pass on the head; the full `tests/policies/protomotions` suite (44 tests) is green.
- The behaviour-class tests use a probe pose (tilted body, world-axis spin) deliberately chosen so the two frames are distinguishable, with an in-test premise assertion guarding that - so the tests cannot silently pass for either convention.
- The end-to-end class builds a hermetic 36-dof MJCF in `tmp_path`, downloads nothing, and skips cleanly without `mujoco`; it also covers the resample branch.
- Repo-wide guards pass on the head: docstring xref resolution, dangling-doc-refs, ASCII log strings, no-host-paths, and both changelog-fragment checks (466 tests). `ruff check` / `ruff format --check` clean on all touched files.
- Changelog fragment present and correctly named (`changelog.d/2411-protomotions-cache-velocity-frames.md`), per AGENTS.md PR workflow step 3.
- The new `:func:` cross-reference targets in `bridge.py` and `motion_utils.py` each stay on one line, per the one-line xref rule in Review Learnings (#86).

## Verification suggestions

- To reproduce the pre-fix split beyond CI: copy `tests/policies/protomotions/test_cache_velocity_frames.py` onto the merge base and run it - expect exactly the 4 docstring-claim failures and 6 behaviour-control passes.

PR #2412, opened 2026-08-17
fix(teleop): measure a session's duration budget from the end of setup

cagataycali:
## Problem

`TeleopMixin.teleoperate` stamps `self._teleop_start_mono`, and `_teleop_loop` builds its deadline as `_teleop_start_mono + duration`. The loop then resolved `strands_robots.mesh.security` -- the module holding the two helpers it judges every frame with -- on its own first line.

That import is lazy on purpose: the mesh package reaches `strands_robots.simulation`, which this mixin must not depend on. But it is not cheap, and it sat **inside** the window `duration` bounds:

```
import strands_robots            0.404s   350 modules
import strands_robots.mesh.security  2.110s  +841 modules  (pulls strands_robots.simulation)
```

So the first session in a process spent ~2.1s between the clock starting and the first poll. Driven through the public entry point on a cold process, `hz=50`, `duration=1.0`, a leader that returns a frame every call:

| | cold process | after `mesh.security` is already imported |
|---|---|---|
| `status` | `success` | `success` |
| report | `0 frames, 0 errors, 0.0Hz over 1.9s` | `50 frames, 0 errors, 49.6Hz over 1.0s` |
| leader polled | **0 times** | 50 |
| reached `send_action` | **nothing** | 50 frames |

`connect()` ran, `disconnect()` ran, and the session reported success having commanded nothing.

## Why this is a defect rather than a slow path

`positive_finite_number_error(duration, ...)` refuses `duration=0` precisely because the loop condition would be false on its first evaluation. `_teleop_stats` then derives the status from the counters rather than hardcoding `success`, and its comment says why:

> A leader whose every frame is refused ... otherwise reports 0 frames, 0 errors and "success": a silent no-op, which is the outcome this derivation exists to refuse.

Any positive `duration` under the import cost produced exactly that state by a route neither guard can see, and every longer session was silently shortened by the same ~2.1s (a 30s session teleoperated for ~27.9s).

## Fix

`teleoperate` resolves the module before it stamps the clock, alongside the rest of setup -- which already includes connecting every device. The loop's lazy import is unchanged and becomes a `sys.modules` lookup, so:

- the layering is untouched (nothing new at module scope; `test_the_mesh_package_is_not_imported_at_module_scope` and `test_the_slew_helper_is_imported_inside_the_loop` both still hold unmodified);
- a warm session is byte-identical;
- `duration` measures teleoperation instead of teleoperation plus setup.

Stated as `importlib.import_module(...)` rather than an unused `import`, so the statement is the effect.

## The other candidate fix does not work

Resolving the import at the top of `_teleop_loop`, above the `deadline = ...` line, looks equivalent and is not: `_teleop_start_mono` is stamped in `teleoperate`, before `_teleop_loop` is called at all, so moving statements around inside the loop cannot change how much time has already passed. Measured, same cold process:

| variant | leader polled | reported |
|---|---|---|
| `main` | 0 | `0 frames, 0.0Hz over 1.8s` |
| import moved above `deadline =` inside `_teleop_loop` | **0** | `0 frames, 0.0Hz over 1.8s` |
| resolved before the stamp in `teleoperate` (this PR) | 50 | `50 frames, 49.7Hz over 1.0s` |

Only the position relative to the **stamp** matters, which is why the ordering is pinned over the source here -- both statements read fine either way round.

## Scope

One site. An AST sweep for the shape -- a function that reads a clock and then resolves an in-repo package -- returns two other hits, both false positives: `mesh/core.py::send` stamps a `timestamp` field into a message payload, and `tools/lerobot_teleoperate.py` uses `int(time.time())` to name a session. Neither is a budget. (The sweep cannot see this case, because the stamp and the import are in different functions.)

`block=False` now returns once setup is done rather than immediately, which is the same trade: while it returned early, `get_teleoperate_status()` reported a running session at 0 frames and 0Hz for the first couple of seconds because the loop thread was still importing.

## Tests

`tests/test_teleop_session_budget_excludes_setup.py`, 7 cases. The cost is injected deterministically rather than relying on the real import being slow: the leaf module is dropped from `sys.modules` and a `sys.meta_path` finder sleeps while it is looked up (returning `None`, so the real finders still resolve it). Each behavioural case premise-asserts the finder actually ran, so it cannot pass vacuously against a warm `sys.modules`.

**5 failed / 2 passed on `main`, 7 pass here.** The pre-fix failures name measurements, not wording:

```
a 0.3s session polled the leader 0 times: the 0.8s resolve of
strands_robots.mesh.security was charged to the budget ...
reported elapsed 0.801s for a 0.3s session, which is at least the 0.8s resolve cost ...
teleoperate(block=False) returned in 0.000s while resolving ... costs 0.8s ...
```

The 2 that pass on both trees are the boundaries the fix must not move:

- a **warm** session is unchanged (the common case, and the obvious wrong fix would perturb it);
- a session refused for an unusable `duration` still resolves **nothing** -- the rate/duration guards run ahead of setup, so a refusal must not pay for an import. This fails if the resolve is ever hoisted above those guards.

## Verification

- 36-file blast radius (every test touching `teleop_mixin` / `teleoperate` / `attach_teleop`, plus the repo-wide docstring, ASCII, dash, dependency and changelog guards): **2183 passed, 0 failed** here; **2178 passed, 5 failed** on `upstream/main` with the same file present. `2178 + 5 == 2183`, both collected 2184, and the 5 base failures are exactly the pre-fix cases -- zero foreign, zero branch-only.
- `mypy strands_robots tests tests_integ`: 19 errors on both trees, identical sets, none in the changed files.

One capture script per tree, run in a genuinely cold process: the real `teleoperate()` loop drives a recording host, and the commanded frames are then replayed onto a MuJoCo `so101` follower on the session's own time axis. Same script, same leader sweep, same camera -- only `strands_robots` differs.

teleop session budget

[still](https://raw.githubusercontent.com/cagataycali/robots/media-teleop-session-budget/teleop-session-budget.png) | [mp4](https://raw.githubusercontent.com/cagataycali/robots/media-teleop-session-budget/teleop-session-budget.mp4)

The left panel is honestly static: its per-frame pixel delta is **0.0000 across all 20 transitions**, because no command ever reached the follower during the second the caller asked for. The right panel moves on every frame (3.08 to 12.61), delivering 50 commands, and the two panels differ by 14.39 mean absolute pixels at peak divergence (t = 0.25s). Both sessions reported `status=success`.


yinsong1986 (approved):
## Summary

Moves the one-time resolution of `strands_robots.mesh.security` out of the window that `teleoperate(duration=...)` bounds: `teleoperate` now imports the module during setup, before it stamps `_teleop_start_mono`, so a cold process's ~2s import cost is no longer charged to the session budget (previously a short first session polled the leader zero times and still reported `success`, and longer sessions were silently shortened). The placement is correct on both boundaries the diff must respect: the resolve sits after the `hz`/`duration` guards (a refused knob pays for no import) and before the clock stamp, and the loop's lazy `from strands_robots.mesh.security import ...` is unchanged, so the mixin-must-not-depend-on-simulation layering holds.

## What's good

- The regression test injects the import cost deterministically via a `sys.meta_path` finder and premise-asserts the finder ran, so it cannot pass vacuously against a warm `sys.modules`; the two boundary tests (warm session unchanged, refused `duration` resolves nothing) pin exactly the properties an alternate fix would break.
- The AST test pins the resolve-before-stamp ordering, which is the load-bearing part of the fix and invisible to a casual re-read.
- Changelog fragment (`changelog.d/2412-...`) per AGENTS.md step 3, docs and the `duration` docstring updated in the same PR, and the `block=False` behaviour change (returns after setup rather than immediately) is stated in both.

yinsong1986 (approved):
## Summary

Moves the one-time resolution of `strands_robots.mesh.security` out of the window `teleoperate(duration=...)` bounds. The loop resolved the module on its first line, after `teleoperate` had already stamped `_teleop_start_mono` (the base of the deadline), so a cold process charged the ~2s import to the session budget: a short first session polled the leader zero times and still reported `success`, and every longer session was silently shortened. The fix resolves the module during setup - after the hz/duration guards (a refused session pays for nothing) and before the clock stamp - leaving the loop's lazy import a `sys.modules` lookup, so the mixin/mesh layering and a warm session are both unchanged.

## What's good

- The regression tests inject the import cost deterministically (a `sys.meta_path` finder that sleeps), premise-assert the resolve was actually cold so they cannot pass vacuously, and pin the resolve-before-stamp ordering over the AST - the one property of this fix that is invisible in a source read.
- Both boundaries are pinned: a warm session is byte-identical, and a session refused for an unusable `duration` still resolves nothing (fails if the resolve is ever hoisted above the knob guards).
- The PR description rules out the tempting wrong fix (moving the import above `deadline =` inside `_teleop_loop`) with measurements showing it changes nothing.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; `docs/hardware/teleoperation.md` and the `duration` docstring both document the new `block=False` return timing.
- Scope discipline: one site, with an AST sweep documenting why the two other clock-then-import hits are false positives.

PR #2413, opened 2026-08-17
fix(simulation): measure a sim rollout's frame pacing and latency on a clock that cannot step

### LINKED ISSUE #2409: A sim rollout's frame pacer and duration bound are measured on a clock that can step (simulation/policy_runner.py)
A sim rollout paces its own frames and bounds its own duration against `time.time()`, which is not a clock but the current opinion about the date. This is the same defect class settled for the agent-callable tools in #2404, the hardware control loops in #2406, and the mesh in #2408; `strands_robots/simulation/policy_runner.py` is what is left of it, and the frame pacer is the part that decides rather than reports.

## The pacer

`strands_robots/simulation/policy_runner.py`:

```python
2267:  step_start = time.time()
2369:  sleep_time = frame_interval - (time.time() - step_start)
```

`sleep_time` is how long the rollout waits before the next frame, computed from two readings of a clock that can step between them:

- **Backward step of S** - `time.time() - step_start` goes negative, so `sleep_time` becomes `frame_interval + S`. The rollout stalls for the size of the step, mid-episode, with the policy already loaded and the sim mid-contact.
- **Forward step** - `sleep_time` goes negative, the frame pacing is skipped entirely and the loop runs unthrottled for the remainder of the rollout.

Neither is reported, because nothing raised. AGENTS.md names a frame pacer as one of the constructs that must be measured on `time.monotonic()` ("Clocks: a duration is measured, a stamp is recorded").

## The rollout bound and the reported duration

The same file measures how long the rollout ran on the same clock, so the record moves with the step too:

```python
1509:  start_time = time.time()
1917:  elapsed  = time.time() - start_time
2250:  start_time = time.time()
2373:  duration = time.time() - start_time
```

## Reporting-only, same class

`strands_robots/policies/lerobot_local/policy.py` derives inference latency the same way (`990`/`1054`, `1351`/`1393`, `2023`/`2026`). No decision hangs off these - they are reported - so they are lower consequence, but a latency that a clock correction makes negative or hour-long is not a useful measurement either.

## Out of scope

`strands_robots/policies/lerobot_async/policy.py:463` passes `timestamp=time.time()` into a record another reader correlates. That is an absolute stamp and correctly stays on the wall clock, as do the mesh envelope `"t"` fields left untouched by #2408.

## Shape of the fix

The same one the three landed changes used: swap the clock for every duration, rename each base for the clock it holds (`_mono`), leave absolute stamps alone, and pin the contract on behaviour with a clock double that takes a known step - `tests/tools/test_tool_wait_budgets_survive_a_clock_step.py`, `tests/test_control_loop_budgets_survive_a_clock_step.py` and `tests/mesh/test_mesh_durations_survive_a_clock_step.py` are the pattern. The pacer is worth a test that asserts the *achieved frame interval* across a step rather than the value of `sleep_time`, so the assertion cannot be satisfied by renaming a call.

Filed so the remainder of the class is tracked rather than recorded only in a pull request description.

cagataycali:
Closes #2409.

A sim rollout paced its frames and bounded its own reports against `time.time()`, which is not a clock but the current opinion about the date. This is the last member of the defect class settled for the agent-callable tools (#2404), the hardware control loops (#2406) and the mesh (#2408).

## The pacer decides, and it was the one on the wall clock

`replay()` read a base at the top of each frame and slept for whatever was left of `frame_interval`. Both readings were `time.time()`, so a step landing mid-episode changed the rate at which a recorded trajectory reached the robot. Measured by driving the real loop through a clock double (30 fps, 40 frames, ~5 ms of work per frame):

| clock event | achieved frame interval after it | reported `duration_s` |
| --- | --- | --- |
| no step (control) | 33.3 ms | 1.33s |
| wall clock steps +30s | 5.0 ms (pacing skipped) | 31.3s |
| wall clock steps +1h | 5.0 ms (pacing skipped) | 3601.3s |
| wall clock steps -2s | one 2033 ms stall | -0.67s |

Forward, `frame_interval - (time.time() - step_start)` goes negative, the sleep is skipped, and the remaining frames are pushed at whatever rate the loop can turn over - a position-servo arm is commanded through recorded targets ~6x faster than they were recorded. Backward, the sleep becomes `frame_interval + step` and the replay stalls with the arm parked on a recorded pose. Every row finished `Frames: N/N` with `status="success"`, and the `duration_s` in that result moved with the step too, so the record does not show it either. `run()`'s `elapsed_s` shared the clock.

## The RTC latency is not reporting-only

The issue filed `lerobot_local`'s inference latency as lower-consequence telemetry. Reading the consumer says otherwise: `inference_elapsed` is appended to `_rtc_latency_history`, and `_estimate_inference_delay` turns that window's p95 into `inference_delay` - the RTC paper's `d`, which lerobot's denoiser uses as `get_prefix_weights(start=d, ...)` to **freeze the first `d` actions of the new chunk** to the already-committed prefix, and which is also the offset the chunk-seam slice skips. On true-async hardware driven without a runner (no `set_rtc_observed_delay`) that estimate is the only source of `d`.

The persistence is measured, not assumed: a lone outlier **is** the p95 of a window holding 2..20 samples and drops out from 21 on, and `reset()` clears the window every episode - so an episode's first inferences are always in that range, which is also when a process idle between episodes is most likely to receive the correction. Driving the real `_predict_with_rtc` path (50-step chunks, 30 Hz, 117 ms true latency = a 3-step delay), with the step landing during the 2nd inference:

| clock event | usable actions delivered per inference |
| --- | --- |
| no step (control) | `[50, 47, 47, 47, 47, ...]` |
| wall clock steps +30s | `[50, 47, 1, 1, 1, ...]` (delay 903) |
| wall clock steps +1h | `[50, 47, 1, 1, 1, ...]` (delay 108003) |
| wall clock steps -2s | a negative latency recorded into the window |

At `d >= execution_horizon` the whole prefix is frozen: the freshly computed chunk is discarded at the seam and the arm keeps executing the stale one, for roughly the next 19 seams, then the estimate recovers by itself - which is what makes it hard to attribute later. Nothing raises; nothing is logged above debug.

## The change

Every duration in the two files reads `time.monotonic()`, and each base is renamed for the clock it holds (`start_mono`, `step_start_mono`, `inference_start_mono`) so a future reader does not reach for `time.time()` again. All six bases are function locals, so nothing serialised or public changes shape. Absolute stamps are untouched, including `lerobot_async/policy.py:463`'s `timestamp=time.time()`, which another reader correlates. Absent a clock step, behaviour is identical.

## Tests

Two files, each asserting on behaviour rather than on which clock is called, so neither can be satisfied by renaming a call:

- `tests/simulation/test_rollout_durations_survive_a_clock_step.py` - drives the real `replay()` loop through a clock double whose *wall* clock steps mid-episode and its monotonic clock does not, and asserts the **achieved** frame interval (the pacing the robot experiences), the reported duration, and that no single sleep exceeds one frame. No simulator, dataset or robot: the engine is an in-memory fake and the dataset is a list of dicts.
- `tests/policies/lerobot_local/test_rtc_latency_survives_a_clock_step.py` - drives the real `_predict_with_rtc` path and asserts on **what comes out of the seam** (the count of usable actions), that no impossible latency enters the window, and that the delay estimate is unmoved.

Both were run against the pre-fix source to confirm they fail for the stated reason: 9/15 and 7/12 fail with a step, and every no-step control row passes on both sides. `TRUE_LATENCY` is deliberately `0.117` rather than `0.1` - `0.117 * 30 = 3.51` sits away from an integer boundary, so `int()` returns 3 whether the subtraction happened on a ~1.78e9 wall base or a small monotonic one, and each test then fails only for the clock step it names. (A base of that magnitude loses enough precision to turn a 100 ms latency into 99.99999 ms and a 3-step delay into 2 - a second, smaller argument for the monotonic base, deliberately not asserted on.)

27 new tests pass; the neighbouring `tests/policies/**` and `tests/simulation/**` suites show no new failures (the 5 failures in this environment - a device-move test and 4 tokenizer tests - reproduce identically on an unmodified checkout, both caused by the absence of a real `torch`).

## Docs

- `docs/policies/lerobot-local.md` and `docs/policies/persistent-worker.md` described `load_time_s` / `policy_load_time_s` as **wall** seconds; both now name the clock they are actually measured on.
- `docs/simulation/overview.md` states the same for `run_policy`'s `elapsed_s`, matching the sentence `docs/hardware/teleoperation.md` already carries.
- `AGENTS.md` names a frame pacer as an example of the rule but its pin list did not include one; the two new tests are now named there.

## Deliberately not in this diff

- `_rtc_last_inference_time` is dead in production (assigned in `__init__` and `reset()`, read nowhere) and is a wall-clock-shaped name sitting in the code path this PR touches, so it is an invitation to reintroduce the defect. Removing it also requires editing an existing test that asserts on it, which is its own conversation - filed separately rather than smuggled in here.
- The same class remains in other subsystems (`tools/lerobot_camera.py`, `rendering/video.py`, `simulation/mujoco/rendering.py`, `simulation/isaac/simulation.py`). Out of scope for #2409 and filed separately so the remainder is tracked on the board rather than in this description.

## Changelog

**Round 1** - `Test and Lint` was red on a single mypy error in this PR's own new test, not on review feedback:

```
tests/simulation/test_rollout_durations_survive_a_clock_step.py:175: error:
Argument 1 to "PolicyRunner" has incompatible type "_FakeEngine"; expected "SimEngine"  [arg-type]
```

`_FakeEngine` was duck-typed, but `PolicyRunner.__init__` is annotated `sim: SimEngine`. Every other fake this suite hands to `PolicyRunner` (`test_policy_runner.py`, `test_policy_runner_paths.py`, `test_policy_runner_benchmark.py`, `test_get_world_point_math.py`) subclasses `SimEngine`; this one was the exception, so it now subclasses too and implements the abstract surface with the same trivial boilerplate those files use.

`send_action` mirrors the base declaration (`dict[str, Any] | Sequence[float]`) rather than dropping its annotations - the frame hook is the load-bearing method here, so it stays type-checked instead of silenced.

No production file changed, and the pacing the tests measure is unchanged: `replay()` reaches none of the added methods, and `SimEngine` declares no `_world` / `_is_recording` / `_active_recorder` class attribute for the runner's `getattr` probes to now resolve. Verified with the error reproduced pre-fix and absent post-fix under an otherwise identical local mypy run; `ruff check`, `ruff format --check` clean; 15 tests pass in this file and 27 across both files the PR adds.


yinsong1986 (dismissed):
## Summary

Moves every duration measurement in `simulation/policy_runner.py` and `policies/lerobot_local/policy.py` from `time.time()` to `time.monotonic()`: the `replay()` frame pacer and its reported `duration_s`, `run()`'s `elapsed_s`, the two model-load timings, and the RTC inference latency that feeds `_estimate_inference_delay`'s p95 - the one that decides how many actions of each new chunk are frozen at the seam, so it is a control input rather than telemetry. Each base is a function local renamed for the clock it holds (`start_mono`, `step_start_mono`, `inference_start_mono`); no serialised field names or shapes change, and absolute stamps (including `lerobot_async`'s wire `timestamp=time.time()`, which a remote reader correlates) are correctly left on the wall clock. The pre-existing `_rtc_last_log_time` throttle the new code sits beside was already monotonic, so no mixed-clock subtraction is introduced.

## What's good

- Both new test files assert on behaviour (achieved frame interval, usable actions delivered at the seam, reported duration) rather than on which clock is called, exactly as AGENTS.md > Review Learnings (#86) > 'Clocks' asks, and the clock doubles use disjoint epochs so reading the wrong clock cannot pass by luck.
- Verified locally at head `18b4b431`: all 27 new tests pass, and the neighbouring `test_policy_runner*.py` / `test_replay_episode_index_domain.py` suites (288 passed, 3 skipped) show no regressions.
- Changelog fragment per step 3, docs updated to name the clock, AGENTS.md pin list extended, and the two adjacent-but-separate concerns (`_rtc_last_inference_time` dead code, the same class in other subsystems) are explicitly deferred rather than smuggled in - good scope discipline.

yinsong1986 (approved):
## Summary

Moves every duration measurement in `simulation/policy_runner.py` and `policies/lerobot_local/policy.py` from `time.time()` to `time.monotonic()`: the `replay()` frame pacer and its reported `duration_s`, `run()`'s `elapsed_s`, the two model-load timings, and the RTC inference latency whose p95 becomes `_estimate_inference_delay`'s `d` - the value that decides how many actions of each new chunk are frozen to the committed prefix. Verified against the head: all six pre-fix `time.time()` call sites in the two files are converted with no leftover references to the old names (only comment mentions remain), all bases are function locals so nothing serialised or public changes shape, and absolute stamps are untouched. The change lands squarely on AGENTS.md's clock rule ("a duration belongs on `time.monotonic()`... a stamp stays on `time.time()`") and correctly identifies the RTC latency as a deciding duration rather than telemetry.

## What's good

- Both new test files assert on behaviour (achieved frame interval, usable actions at the seam, the delay estimate itself) rather than on which clock is called, so a rename cannot satisfy them - exactly the shape AGENTS.md's clock-rule pins ask for. The clock doubles use distinct, unrelated epochs so reading the wrong clock cannot pass by luck.
- The 27 new tests pass at head; the directly affected suites (`tests/policies/lerobot_local/`, the `policy_runner` test files) run clean, and the failures observable in a local environment reproduce identically on the unmodified merge base.
- Changelog fragment present per convention; docs that described `load_time_s` / `policy_load_time_s` / `elapsed_s` as wall seconds are corrected in the same PR; the AGENTS.md pin list is extended with the two new tests.
- Good scope discipline: the dead `_rtc_last_inference_time` and the same defect class in other subsystems are named and deferred rather than smuggled in, and the round-2 commit fixing the mypy `_FakeEngine` error touched only the test file.

## Verification suggestions

- To reproduce the failure mode this pins: check out the merge base's `strands_robots/` alongside this PR's tests and run the two new test files - the parametrised step rows fail while every `no-step-control` row passes on both sides.

cagataycali:
Verified `b5c8ba13` locally against the pinned build. Summary: the typing fix is the right one, and — more importantly — **the test step has never completed on this PR**, so I ran it.

## The lint blocker, and why the fix on the branch is the right shape

`18b4b431` failed `Test and Lint` on exactly one mypy error and nothing else:

```
tests/simulation/test_rollout_durations_survive_a_clock_step.py:175: error:
  Argument 1 to "PolicyRunner" has incompatible type "_FakeEngine"; expected "SimEngine"  [arg-type]
Found 1 error in 1 file (checked 1325 source files)
```

Worth recording why only this site was reported, since the suite hands a fake to `PolicyRunner` in ~25 other places: the construction sits in the one *annotated* body in the file (`_run(wall_step: float) -> tuple[...]`), and mypy leaves unannotated function bodies unchecked. So the neighbouring sites are not exempt by design — they are simply unchecked.

Declaring `_FakeEngine(SimEngine)` is the correct resolution rather than casting at the call site: **23 test files subclass `SimEngine`** for an engine double, including three of the four this suite's sibling files use (`test_policy_runner.py`, `test_policy_runner_benchmark.py`, `test_get_world_point_math.py`). A cast would have been the exception. Annotating `send_action` to mirror the base declaration rather than dropping its annotations is the right call too — it keeps the frame hook, which is the thing the tests measure, type-checked.

The check that matters for a change to a test double is whether it weakened the test. It did not. Copying the branch's test file onto unmodified `main` still fails for the defect:

| tree | `test_rollout_durations_survive_a_clock_step.py` |
| --- | --- |
| `main` (pre-fix source) | **9 failed, 6 passed** |
| this branch | **15 passed** (27 across both new files) |

Subclassing `SimEngine` did not divert the pacing path — every no-step control row still passes on both sides, and the failures on `main` are the stated ones (a single sleep of `30.03s` against a `0.033s` frame budget; `delay 903` and `delay 108003` from the RTC estimate; a negative latency entering the window).

## The test step has never run on this PR

Both `Test and Lint` runs on this branch to date:

| created | sha | `Run lint` | `Run tests` |
| --- | --- | --- | --- |
| 21:22:38Z | `18b4b431` | failure | **skipped** |
| 22:12:47Z | `b5c8ba13` | (in progress) | (in progress) |

The lint step precedes the test step, so the first run told us nothing about the diff's behaviour. The approving review therefore landed on a revision whose behaviour CI had not yet exercised, so rather than wait on the second attempt I ran the suite locally.

**Full `tests/`, this branch vs `main`, same interpreter, both trees in parallel:**

| --- | --- | --- | --- |
| branch | 31783 | 6 | 31889 |
| `main` | 31757 | 5 | 31862 |

Collected delta is **exactly 27**, the tests this PR adds — nothing else was gained or lost. The failure sets differ only inside `tests/mesh/test_zenoh_transport_security.py`, and they differ in *which* members fail (3 on one tree, 2 different ones on the other), which is contention rather than the diff: that file opens real localhost Zenoh sessions, and run on its own it is **9 passed / 2 skipped on both trees**. The three failures shared by both trees are pre-existing and environmental — two `tests/policies/vera/...` cases that need Docker, and one Newton camera case asserting on message wording.

`tests/simulation` + `tests/policies` re-run on `b5c8ba13` itself: **15378 passed, 3 failed**, those same three pre-existing cases. (The full-suite table above was produced on the same shipped source with an equivalent typing fix confined to that one test file — `git diff` between the two revisions under `strands_robots/` is empty, so the library-level result carries over unchanged.)

Lint on `b5c8ba13`, CI's exact scope: `ruff check` clean, `ruff format --check` clean (1328 files), `mypy strands_robots tests tests_integ` **0 errors in scope** (the 14 reported are in `examples/`, present identically on `main`, and outside the lint targets — which is why CI reported only the one).

## Two claims in the description I checked rather than took on trust

**The environmental caveat holds.** The description notes 5 failures attributed to the absence of a real `torch` (a device-move test and 4 tokenizer tests). On a checkout with a real `torch` build (2.10), none of those 5 appear in either tree — the attribution is correct.

**The RTC latency really is a decision input, not telemetry.** This is the part of the description that carries the second half of the diff, so it is worth substantiating against lerobot rather than reasoning about it. `get_prefix_weights(start, end, total)` in `lerobot/policies/rtc/modeling_rtc.py:251` sets `weights[:start] = 1.0` on the `ZEROS` schedule and applies `_add_leading_ones(weights, start, total)` on `LINEAR`/`EXP`. So passing the p95 estimate as `start` does pin that many leading actions to the already-committed prefix, and an estimate inflated by a clock step discards the freshly computed chunk at the seam. Measuring it on a monotonic clock is a correctness fix, not a cosmetic one.

Also confirmed for the #2415 deferral: `_rtc_last_inference_time` is assigned at `policies/lerobot_local/policy.py:607` and `:752` and read nowhere in `strands_robots/` — the only readers are `tests/policies/lerobot_local/test_policy.py:1881,1888`. Leaving it out of this diff is right; it cannot be removed without settling that test's intent.

Nothing blocking from me. Once `Test and Lint` reports on `b5c8ba13` this is ready.

cagataycali:
One note so the check state is not misread: attempt 1 of `Test and Lint` on `b5c8ba13` reports **cancelled**, but that is the job-timeout ceiling rather than a test failure.

| step | attempt 1 |
| --- | --- |
| `Run lint` | **success** |
| `Run tests` | cancelled mid-run |

The job ran `22:43:11Z -> 23:28:25Z` = **45m14s**, same `head_sha`, `run_attempt` 1, no superseding push — it hit the 45-minute limit while the tests were still executing. Dependency installation alone took ~17 minutes of the previous attempt's budget, so the margin over a full `tests/` run is thin on a slow runner.

The important part is that `Run lint` passed: the single `arg-type` error that blocked `18b4b431` is gone, and the test step is now the first one to actually execute on this PR rather than being skipped behind lint. I have re-run the failed jobs (now attempt 2).

The behaviour that attempt cannot finish is already covered by the local run in the previous comment — full `tests/` against `main` on the same source, collected delta exactly 27 with no new failures, and `tests/simulation` + `tests/policies` re-run on `b5c8ba13` itself at 15378 passed / 3 pre-existing failures.

PR #2416, opened 2026-08-17
fix(policies/protomotions): read a quaternion as the rotation it encodes

cagataycali:
## Problem

`policies/protomotions/state_utils.py` derives two rotations for the tracker: it
body-frames the root angular velocity (`quat_rotate_inverse`, via
`compute_root_local_ang_vel`) and extracts a yaw for heading alignment
(`extract_yaw_quat`, via `compute_yaw_offset`). Both formulas are the Rodrigues
form, valid only for a **unit** quaternion: each mixes a quadratic term in the
quaternion components with a constant, so scaling does not cancel. Both trusted
their input, so a quaternion scaled by `s` - which encodes exactly the same
rotation - was read as a *different* rotation.

Measured, for one rotation, against `R(q)^T v` built from first principles:

| `\|q\|` | `quat_rotate_inverse` error | `extract_yaw_quat` (true yaw 90 deg) |
|---|---|---|
| 1.000 | 0.0% | 90.000 deg |
| 1.010 | 3.0% of `\|v\|` | 91.129 deg |
| 1.100 | 31.5% | 99.846 deg |
| 2.000 | 450% | 126.870 deg |

An **all-zero** quaternion was worse than wrong. It collapses the form to
`v * (2*0 - 1)`, so the helper returned exactly `-v`: the angular velocity a
balance controller closes the loop on came back with its **sign flipped**.

## Why a non-unit quaternion arrives here

Three ordinary sources, none of them exotic:

* `ProtoMotionsPolicy._extract_root_local_ang_vel`'s own refusal invites a
  caller to hand-assemble `body_rot_xyzw`, and derives the tracker input from
  that batch.
* A real-robot IMU orientation drifts off unit.
* An orientation obtained by **linearly** interpolating two samples is short by
  up to ~8% - which is exactly why the sibling `motion_utils` normalises inside
  its slerp.

Driving the documented hand-assembly path with the lerp'd midpoint of two
samples 90 deg apart (`|q| = 0.92388`), before this change:

| input | read as | should be |
|---|---|---|
| lerp'd orientation | heading `38.825 deg` | `45.000 deg` |
| lerp'd orientation | 1.0 rad/s world yaw rate reaches the tracker as `0.707107` | `1.0` |
| all-zero orientation | `[0, 0, -1]` (sign flipped), no error | refused |

## The change

`quat_rotate_inverse` and `extract_yaw_quat` now normalise their input and
refuse a quaternion that cannot define a rotation (non-finite, or ~zero norm).
The dead soft-floor `_normalize_quat` that had been sitting unused in this
module is replaced by the helper that is now actually called; a soft floor would
not have helped the degenerate case anyway, since `0 / 1e-8` is still zero.
Input dtype is preserved, so the `float32` the ONNX session is typed for is
unchanged.

This brings the module in line with what the rest of the package already
settled. Every other world-to-body helper normalises internally, and the
degenerate case is refused by the same-layer sibling:

| helper | normalises | all-zero quaternion |
|---|---|---|
| `policies/wbc/control.quat_rotate_inverse` | yes | raises `ValueError` |
| `simulation/predicates._quat_rotate_inverse_wxyz` | yes | returns `vec` unchanged |
| `simulation/newton/simulation._quat_rotate_inverse_wxyz` | yes | returns `vec` unchanged |
| `policies/protomotions/state_utils.quat_rotate_inverse` (before) | **no** | **returns `-vec`** |

Scale-invariance is already a pinned contract for the `wxyz` pair
(`tests/simulation/newton/test_backend_helpers.py::TestQuatRotateInverseWxyz::test_unnormalized_quaternion_is_normalized_internally`,
whose docstring reads "the helper must normalise so callers can pass a raw
free-joint quat"). Refusing rather than answering follows the same-layer
`policies/wbc/control` sibling and AGENTS.md rules 5 and 6: a degenerate
orientation cannot be repaired by scaling, and standing in an arbitrary rotation
would feed the network a plausible-looking wrong frame.

`quat_mul` and `quat_conjugate` are deliberately left alone: a product of unit
quaternions is already unit, scaling an input scales the output uniformly (so it
still encodes the same rotation), and normalising a product would change what
"Hamilton product" means. A test pins that.

## Verification

* New `tests/policies/protomotions/test_state_utils_unit_quaternion_contract.py`:
  **22 failed / 7 passed** against `upstream/main`, **29 passed** after. Pre-fix
  failures carry the measurement, e.g. `the same robot orientation, scaled,
  aligned the clip to a different heading: 38.825 deg vs 45.000 deg` and `with
  a quaternion scaled by 4.0, these helpers disagreed with R(q)^T v =
  [0.5, -1.0, 2.0]: {'policies/protomotions/state_utils': [23.0, -46.0, 39.5]}`.
* The 7 that pass on both trees are the controls: a unit quaternion answers
  exactly as before, identity leaves the vector alone, `float32` stays
  `float32`, `quat_mul`/`quat_conjugate` stay plain, and all four helpers
  already agreed at `|q| = 1`.
* Full run over the 33 test files touching these symbols plus the repo-wide
  guards: branch **1257 passed / 0 failed**, base **22 failed / 1235 passed** -
  same collected total, `1235 + 22 = 1257`, no branch-only failure.
* `ruff check` + `ruff format --check` clean; `mypy strands_robots tests
  tests_integ` reports the same 19 pre-existing errors on both trees, none in
  the changed files.
* Re-verified after merging current `main` (which landed #2411 and #2412):
  1023 passed across `tests/policies/protomotions`, `tests/policies/wbc` and the
  doc/docstring guards.

A real Unitree G1 replaying the real clip `walk.qpos.npz` ("a person walking
forward with confident strides"), heading-aligned with `compute_yaw_offset` +
`apply_heading_offset`. The robot's anchor orientation is the lerp'd midpoint
above (`|q| = 0.9239`), and the goal is to walk the 4.53 m lane through the
green gate (half-width 0.42 m).

Both panels run the identical script; only `strands_robots` differs. The gate is
fixed by normalising the same orientation by hand - a value both trees compute
identically (44.616 deg), which is one of the control tests - so it is placed in
the same place in each panel.

heading alignment before and after

| | before (`main`) | after |
|---|---|---|
| clip heading aligned to | 38.816 deg | 44.991 deg |
| error vs the robot's 45 deg heading | **-6.184 deg** | -0.009 deg (float32) |
| lateral offset at the gate | **-0.487 m** | 0.000 m |
| result | **misses the gate** | through the gate |

The two panels are near-identical at the first frame (mean abs pixel delta
0.105 - the same clip, same start pose) and 3.361 apart at the last; 59 of 59
frame transitions move in both.
[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-protomotions-unit-quat/m/heading-align-unit-quat.mp4) -
[still](https://raw.githubusercontent.com/cagataycali/robots/media-protomotions-unit-quat/m/heading-align-unit-quat.png)

## Docs

`docs/policies/protomotions.md` gains "Orientations need not be exactly unit"
under the observation contract: that both helpers normalise, the interpolation
hazard with the measured numbers, and that a quaternion which is not a rotation
is refused.


yinsong1986 (approved):
## Summary

This PR makes `quat_rotate_inverse` and `extract_yaw_quat` in `policies/protomotions/state_utils.py` normalise their quaternion input and refuse one that cannot define a rotation (all-zero or non-finite), replacing the dead soft-floor `_normalize_quat` with a `_unit_quat` helper that is actually called. Both formulas are the Rodrigues form, valid only for unit quaternions, so the pre-fix behaviour silently read a scaled quaternion as a different rotation — and an all-zero quaternion returned the angular velocity sign-flipped. The diff matches what it claims: the fix brings this module in line with the three other world-to-body helpers in the package (all of which already normalise) and with the same-layer `policies/wbc/control` sibling's refusal of the degenerate case (rules 5/6 in AGENTS.md — raise on fatal errors, no silent defaults). I independently verified the new helper: scale invariance holds across 0.5x–4x, `float32` in stays `float32` out, all-zero/NaN/inf inputs raise `ValueError` naming the helper and a remedy, and the lerp'd-midpoint yaw reads 45.000 deg. Refusing where the code previously returned a wrong answer is the intended contract change, not a one-way door — the old behaviour was a measurable defect on the balance-control loop, and the pinned cross-helper agreement test (`test_all_four_helpers_match_first_principles`) keeps the four implementations from drifting apart again.

## What's good

- Changelog fragment follows the `changelog.d/<pr-number>-<slug>.md` convention (AGENTS.md step 3), no direct `CHANGELOG.md` edit.
- `quat_mul`/`quat_conjugate` deliberately left untouched, with a test pinning why — good scope discipline.
- Control tests pin that unit-quaternion behaviour, dtype, and the identity case are unchanged, and the refusal message is asserted to name the helper, the value, and the `mujoco_wxyz_to_xyzw` remedy.
- Docs snippet imports `extract_yaw_quat` from the package root, which is genuinely exported in `strands_robots/policies/protomotions/__init__.py.__all__`.
- Error message and test strings are plain ASCII; no host paths in the test file.

PR #2417, opened 2026-08-17
fix(policies/protomotions): fill the cache's body rows in the order the tracker reads them

cagataycali:
The tracker never looks a body up by name. `ProtoMotionsPolicy` reads its anchor orientation as `future["body_rot"][:, config.anchor_body_index]`, and `anchor_body_index`/`root_body_index` are offsets into `GTP_G1_BODY_NAMES` - the 33-name list pinned from the checkpoint's sidecar, whose comment in `config.py` states the order outright: *"Index 0 = pelvis (root), 16 = torso_link (anchor)"*.

`qpos_to_motion_data` filled those rows from `data.xpos[1:]` - the supplied MJCF's own body order - pairing the two lists positionally, with nothing checking they are the same list.

## The G1 family does not agree on that body set

Of the 44 loadable G1 MJCFs on a machine with mujoco-menagerie, `unitree_ros` and a GR00T checkout, **only 20 have the tracker's `qpos` width at all**, and the widely-shipped fingerless models expose **30** bodies rather than 33: they omit `head` and both `rubber_hand` placeholders. Both `mujoco_menagerie/unitree_g1/g1.xml` - what `Robot("unitree_g1")` resolves to - and the `g1_29dof.xml` shipped beside the tracker assets are in that group.

Handed one, every row after the gap shifts by one. Measured on a real 120-frame Kimodo walk clip:

| | value |
|---|---|
| row 16 resolves to | `left_shoulder_pitch_link` (tracker asked for `torso_link`) |
| anchor orientation error | **mean 20.80 deg, max 28.16 deg** |
| rows holding another body's rotation | 29 of 33 |
| reported status | success, 30 rows |

The physical models agree: `torso_link` sits at row 15 of that cache, and comparing row 15 against the tracker MJCF's torso gives **0.00 deg**. The entire error is the index.

anchor row before and after

Green is `torso_link`, the body the tracker names as its anchor; red is the row it was actually handed. Rendered headless from the worst frame of the clip.

## Change

Resolve each row by name against `GTP_G1_BODY_NAMES`, and refuse a model that cannot supply a body the tracker reads:

```text
ValueError: ProtoMotions G1 MJCF .../g1_29dof.xml is missing 3 of the 33 bodies the
tracker reads by row index: ['head', 'left_rubber_hand', 'right_rubber_hand']. ...
```

Refusing rather than dropping the row is the only non-silent option: the bridge cannot synthesise a pose for a body the model does not have, and a short row array would leave `anchor_body_index` pointing at whatever ended up there.

The model's `qpos` layout is validated the same way - a free root plus the 29 tracker joints. Previously a 50-wide model surfaced as `could not broadcast input array from shape (36,) into shape (50,)`, naming neither the model nor a width, and raising the same `ValueError` the docstring documents for a bad `qpos` **argument** - so a model-side mismatch read as a caller-side one. It now says which it is.

**For the tracker's own embodiment this is a no-op:** all six cache arrays are byte-identical to `main` (`body_pos`, `body_rot`, `body_vel`, `body_ang_vel`, `dof_pos`, `dof_vel`).

## Tests

`tests/policies/protomotions/test_cache_body_rows_match_the_tracker_order.py`, 16 tests, graded against an independent MuJoCo oracle that resolves each body **by name**, so a row is never confirmed by the same index arithmetic that produced it. Pre-fix on `main`: **10 failed / 6 passed**.

Failing pre-fix:

- the anchor row holds another body - `row 16 ... is 12.60 deg from 'torso_link' and matches 'left_shoulder_pitch_link' instead`
- 29 of 33 rows hold another body's rotation; body positions shift the same way
- a variant missing tracker bodies is refused, and the refusal names all three and says the rows are read by index (all three **DID NOT RAISE** before)
- a wider `qpos` layout is refused as a model problem, naming the model
- **a model with no free root is refused** - previously accepted silently, writing the root pose into hinge joints
- the `Returns:` section names the row order

Passing on both trees, so they bound the change rather than restate it: the root row is right in either order; the row count and the declared anchor/root indices stay in range; `dof_pos` is still read straight off `qpos` (joint order is untouched - it already matched for every 36-wide G1); a too-narrow `qpos` is still a `qpos` refusal; a padded `qpos` is still truncated rather than refused.

`test_cache_velocity_frames.py`'s end-to-end fixture built a generic 36-dof chain named `root`/`l0..l28`. It was only ever accepted because nothing checked, so it now carries the tracker's names via the shared builder; row 0 is still the root that every assertion there reads, and all of its assertions are unchanged.

## Verification

- `tests/policies/protomotions` 60 passed.
- Blast radius (26 files: every test touching protomotions/`MotionPlayer`/a docs page, plus the repo-wide docstring, xref, unicode, ASCII, dependency and changelog guards): branch **1034 passed / 0 failed**, base **1024 passed / 10 failed** - 1024 + 10 = 1034, same collected total, base failures exactly the 10 pre-fix ones, zero branch-only.
- `ruff check` + `ruff format --check` + `mypy` over `strands_robots tests tests_integ` clean; mypy 19 pre-existing errors on both trees, 0 in the touched files.

### DIFF
--- a/changelog.d/2417-protomotions-cache-body-row-order.md
+++ b/changelog.d/2417-protomotions-cache-body-row-order.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- **policies/protomotions**: `qpos_to_motion_data` now fills the reference-motion
+  cache's body rows by name against `GTP_G1_BODY_NAMES` - the order the tracker
+  indexes them by - and refuses an MJCF that cannot supply a body it reads or whose
+  `qpos` layout is not the tracker's. Read positionally, a fingerless 30-body G1
+  shifted every row after the missing `head`, so the tracker was handed
+  `left_shoulder_pitch_link` where it asked for its `torso_link` anchor.
--- a/docs/policies/protomotions.md
+++ b/docs/policies/protomotions.md
@@ -112,6 +112,32 @@ cache["num_frames"], cache["control_dt"]
 `MotionPlayer` accepts that dict, an `.npz` written by
 `MotionPlayer.save_cache_npz`, or a raw ProtoMotions `.pt`.
 
+### The MJCF has to be the tracker's own embodiment
+
+`proto_mjcf_path` is not just any G1. The tracker reads a body out of the cache
+by **row index**, never by name: `anchor_body_index` and `root_body_index` are
+offsets into `GTP_G1_BODY_NAMES`, the 33-name list pinned from the checkpoint's
+sidecar. `qpos_to_motion_data` fills those rows by name from the model you hand
+it, so the model has to carry all 33 - plus a free root and the 29
+`GTP_G1_JOINT_NAMES` joints, for a `qpos` width of 36.
+
+The G1 family does not agree on that body set. The widely-shipped fingerless
+models omit `head` and both `rubber_hand` placeholders and expose 30 bodies;
+`g1_29dof_with_hand` and `g1_with_hands` expose 44 and a `qpos` width of 50.
+Passing one of those is refused, with a message that names what is missing:
+
+```text
+ValueError: ProtoMotions G1 MJCF .../g1_29dof.xml is missing 3 of the 33 bodies
+the tracker reads by row index: ['head', 'left_rubber_hand', 'right_rubber_hand'].
+```
+
+The refusal is the point. Read positionally, a 30-body model shifts every row
+after the gap by one, so the tracker asks for `torso_link` at row 16 and is
+handed `left_shoulder_pitch_link` - on a walking clip, an anchor orientation
+some 20 degrees out, with nothing in the cache to say so. A model whose `qpos`
+layout differs is refused separately and says so, so a model-side mismatch is
+not read as a bad `qpos` argument.
+
 ### Cache velocities are world-frame
 
 `body_pos` and `body_rot` are world poses, and `body_vel` and `body_ang_vel`
--- a/strands_robots/policies/protomotions/bridge.py
+++ b/strands_robots/policies/protomotions/bridge.py
@@ -29,12 +29,20 @@ import numpy as np
 
 from strands_robots.utils import require_optional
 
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

`qpos_to_motion_data` previously filled the MotionPlayer cache's body rows from `data.xpos[1:]` in the supplied MJCF's own body order, while the tracker reads those rows by fixed offsets into `GTP_G1_BODY_NAMES` - so any G1 variant with a different body set (the widely-shipped fingerless 30-body models) silently handed the tracker the wrong link at every row after the gap, including its `torso_link` anchor. This PR pairs the rows by name via `mj_name2id`, and raises a `ValueError` naming the MJCF and the missing bodies (or the wrong `qpos` layout) when the model cannot serve the tracker's rows - converting a silent data-corruption path into a loud refusal, per the raise-on-fatal / no-silent-defaults conventions. For the tracker's own embodiment the output is unchanged.

## What's good

- Reproduced locally at head `e56e224e`: the new 16-test suite reports exactly the claimed **10 failed / 6 passed** with the base `bridge.py` swapped in, and **60/60** protomotions tests pass on the branch; `ruff check` and `ruff format --check` clean on the touched files.
- The oracle in the new tests resolves bodies by name independently of the code under test, so a row can't be confirmed by the same index arithmetic that produced it - and the shuffled-but-complete embodiment fixture isolates the ordering defect from the missing-body refusal.
- The refusal distinguishes a model-side layout mismatch from a bad `qpos` argument (previously both surfaced as the same `ValueError` class with a broadcast message naming neither), and the free-root check closes a genuinely silent path where a fixed-base model's root pose was written into hinge joints.
- Changelog fragment (`changelog.d/2417-...`), docs section, and the `Returns:` docstring update all land in the same PR; the docstring's row-order claim is itself pinned by a test.

## Verification suggestions

- Spot-check the no-op claim for the tracker's own embodiment: build a cache from the same qpos clip on `main` and on this branch against the real `g1_29dof_rev_1_0` MJCF and diff the six arrays byte-for-byte (`np.array_equal` per key).

PR #2418, opened 2026-08-18
fix(policies/protomotions): use the ground an MJCF already declares

cagataycali:
## Problem

`qpos_to_motion_data` runs forward kinematics on a caller-supplied G1 MJCF. FK needs a floor, and the ProtoMotions G1 MJCF names one from 15 `<contact><pair geom2="floor">` elements, so the bridge appends a plane geom called `floor` when the model does not already declare a ground. The name is fixed by those references, which makes "does this model already have a ground?" load-bearing: answer it wrongly and the appended geom is a second ground plane competing for the `floor` name, and MuJoCo refuses the whole model.

The check read the first `<worldbody>`'s direct `<geom>` children:

```python
worldbody = root.find("worldbody")
has_ground = any(... for g in worldbody.findall("geom"))
```

MuJoCo does three things that read misses. It **merges every `<worldbody>`** a file declares, it **splices `<include>`d content** in before compiling, and it **compiles geoms wherever they are nested** inside bodies. A ground declared any of those three ways is invisible, so the bridge appends a duplicate.

Measured over the six G1 MJCFs on a stock `robot_descriptions` checkout, **three are refused**:

| MJCF | plain MuJoCo | through the bridge (before) |
|---|---|---|
| `unitree_ros/g1_29dof.xml` | OK, nq=36, 74 geoms, planes `['floor']` | `ValueError: repeated name 'floor' in geom` |
| `unitree_ros/g1_23dof.xml` | OK, nq=30, 63 geoms, planes `['floor']` | `ValueError: repeated name 'floor' in geom` |
| `gear_sonic_deploy/g1_29dof.xml` | OK, nq=36, 76 geoms, planes `['floor']` | `ValueError: repeated name 'floor' in geom` |
| `menagerie/unitree_g1/g1.xml` | OK, no plane | OK (floor appended) |
| `unitree_mujoco/g1_29dof.xml` | OK, no plane | OK (floor appended) |
| `unitree_mujoco/scene_29dof.xml` | OK, plane in first worldbody | OK (used as-is) |

These are not exotic files. All three declare the robot tree in one `<worldbody>` and the scene's `floor` plane in a second, and `unitree_ros/g1_29dof.xml` is at the exact `nq=36` width the bridge documents as its input. The refusal blames a file MuJoCo loads without complaint, at a line number (`Element 'geom', line 286`) inside the temp copy the bridge wrote and has already deleted - so it reads as "your MJCF is malformed", and the caller's own file is one line shorter than the number quoted.

## Change

Ask MuJoCo's own parser for the flat geom list instead of re-deriving it from a partial ElementTree read:

```python
spec = mujoco.MjSpec.from_file(str(mjcf_path))
return any(... for geom in spec.geoms)
```

This is the rule `simulation.mujoco.spec_builder` already applies when it decides whether a world owns a ground plane (`g.type == mujoco.mjtGeom.mjGEOM_PLANE for g in scene_spec.geoms`). Parsing a spec resolves includes without compiling meshes - **2.1 ms** on a G1, against **172.6 ms** for a full compile - so it stays well inside the module's one-off build-time cost. It also works on the ProtoMotions MJCF itself, which does *not* compile unpatched (`geom 'floor' not found in collision 0`): the spec parses, reports no plane, and the append still fires.

**The predicate is unchanged** - a plane geom, or a geom whose name reads as a floor or ground - so only the set of geoms it is applied to moves.

## Verification

The three files that already worked are byte-identical after the change (same `nq` / `nbody` / `ngeom`), and the ProtoMotions MJCF still loads with its appended floor (`nq=36 nv=35 nbody=34 ngeom=75`, planes `['floor']`).

New tests, run against `main` and against this branch - **3 failed / 4 passed before, 7 passed after**:

| test | before | after |
|---|---|---|
| a second `<worldbody>` section's floor is not duplicated | fail | pass |
| a floor nested inside a body is not duplicated | fail | pass |
| an included floor is not duplicated | fail | pass |
| a model declaring no ground still gets a floor | pass | pass |
| the appended geom is named `floor` | pass | pass |
| a floor in the first `<worldbody>` is not duplicated | pass | pass |
| a geom named `floor` that is not a plane is not duplicated | pass | pass |

The four that pass on both trees are the bounds. "A model declaring no ground still gets a floor" is the one that matters most: the append is what lets the ProtoMotions MJCF compile at all, so it must keep firing. Each failing test reports MuJoCo's own complaint plus the fact that MuJoCo loads the same file on its own. Each asserts on the compiled model (`ngeom`, and the names of every plane geom) rather than only on "it loaded", so a fix that stopped appending a floor would fail them too.

Full local gate: `ruff check` + `ruff format --check` + `mypy` clean over `strands_robots tests tests_integ` (19 pre-existing errors, unchanged, none in the touched files). Blast radius - every test referencing `protomotions`, `spec_builder`, `qpos_to_motion_data`, a policy docs page, plus the repo-wide docstring/changelog/string guards, 34 files: **branch 1239 passed / 0 failed**, `main` with the same test file **1236 passed / 3 failed**, both collecting 1240. The three failures on `main` are exactly the three above; nothing else changed verdict.

Those gate numbers were measured at the first commit on this branch. `main` has since moved, so the branch was merged up: on the merged tree the whole `tests/policies/protomotions` suite is 96 passed, the pre-fix split against the new `main` is unchanged at 3 failed / 4 passed, and `ruff` / `mypy` are still clean.

## Reference motion from a real G1 MJCF

120 frames of a walking clip through `qpos_to_motion_data(fps=30)` against `unitree_ros/g1_29dof.xml`, captured by one script run under each tree - same MJCF, same clip, same renderer, only `strands_robots` differs. `main` cannot load the model, so there is nothing to render and the left panel is the refusal it produces; this branch resamples the clip into a **199-frame, 30-body** cache, replayed below.

reference motion bridged from a G1 MJCF that declares its own floor

[Full-resolution MP4](https://github.com/cagataycali/robots/raw/media-protomotions-ground-detect/ground-detect.mp4) - every one of the 66 frame transitions moves (per-frame mean |delta| 6.3 to 9.1).

## CI state: one superseded run exists at this head, and it is not a finding

Recorded because a cancelled run is easy to read as a test failure when scanning the checks list. Two runs of `Pull Request and Push Action` exist at `c506d306`:

| run | created | conclusion |
|---|---|---|
| [#5969](https://github.com/strands-labs/robots/actions/runs/32083642770) | 00:13:08Z | `CANCELLED` |
| [#5971](https://github.com/strands-labs/robots/actions/runs/32083916159) | 00:17:13Z | **`SUCCESS`** (completed 00:48:39Z) |

The cancelled one is the earlier of the two - it was superseded mid-flight when the head was re-evaluated to attach check suites, and concurrency cancelled it. The newest run of `call-test-lint / Test and Lint`, and of every other context at this head, is green, and branch protection agrees. If the aggregate rollup ever renders the cancelled run's conclusion, that is the summary counting both runs rather than taking the latest per context - the suite itself has not failed at this head.

No code change accompanies this note; the head commit is unchanged, so the approval above still refers to the tree that was reviewed.


yinsong1986 (dismissed):
## Summary

Fixes a real refusal class in the ProtoMotions reference-motion bridge: `_patch_and_load_mjcf` decided "does this MJCF already declare a ground?" from the first `<worldbody>`'s direct `<geom>` children, which misses grounds declared in a second `<worldbody>` section, nested inside a body, or spliced in via `<include>` - so the bridge appended a duplicate `floor` plane and MuJoCo refused files it loads fine on its own (three of the six stock G1 descriptions). The fix replaces the partial ElementTree read with `mujoco.MjSpec.from_file(...).geoms`, the same flat-geom-list rule `simulation.mujoco.spec_builder` already applies, with the predicate itself unchanged. I verified the mechanism independently (spec parse flattens included and nested geoms without compiling), ran the new test file (7/7 pass on the branch), and confirmed `ruff check` / `ruff format --check` are clean on both touched Python files. `MjSpec` and `spec.geoms` are already used in `spec_builder.py` against the project's `mujoco>=3.5.0` floor, so no new API-version exposure. The helper is private and not exported, the changelog lands as a `changelog.d/` fragment per AGENTS.md step 3, tests use `tmp_path` with no host paths or non-ASCII strings, and the docs page documents the new behaviour. No blockers found.

## What's good

- The three previously-failing cases each carry a premise assertion (`MuJoCo merges the second worldbody's floor in`) proving MuJoCo itself accepts the fixture, so the tests pin the divergence rather than an implementation detail.
- Assertions land on the compiled model (`ngeom`, plane-name list), so a regression that stopped appending the floor entirely would also fail - both directions are pinned.
- Reuses the existing `spec_builder` ground-detection rule instead of inventing a second, subtly different one.

yinsong1986 (approved):
## Summary

Fixes a real refusal class in the ProtoMotions reference-motion bridge: `_patch_and_load_mjcf` answered "does this MJCF already declare a ground?" from the first `<worldbody>`'s direct `<geom>` children, missing grounds declared in a second `<worldbody>`, nested inside a body, or spliced in via `<include>` - so the bridge appended a duplicate `floor` plane and MuJoCo refused files it loads fine on its own (three of six stock G1 descriptions). The fix asks `mujoco.MjSpec.from_file(...).geoms` for the flat geom list - the same rule `simulation.mujoco.spec_builder` already applies - with the predicate itself unchanged, so only the set of geoms it inspects widens. I verified the mechanism independently (a spec parse flattens included and nested geoms without compiling, confirmed on mujoco 3.8.0 against the project's `>=3.5.0` floor, where `MjSpec` is already load-bearing in `spec_builder.py`) and ran the suite at head `c506d306`: 7/7 new tests and 96/96 `tests/policies/protomotions` tests pass on the merged tree. The prior CodeQL finding on the test helper's mixed implicit/explicit returns was resolved in `a50ed542` (the `except` arm now raises, so every path is explicit). The helper is private and unexported, no public API / wire / persisted-schema surface moves, the changelog lands as a `changelog.d/` fragment per AGENTS.md step 3, and the tests use `tmp_path` with no host paths and ASCII-only strings.

## What's good

- Each previously-failing test carries a premise assertion (`MjModel.from_xml_path` on the raw fixture) proving MuJoCo itself accepts the file, so the tests pin the bridge/MuJoCo divergence rather than an implementation detail.
- Assertions land on the compiled model (`ngeom`, plane-name list), so a regression that stopped appending the floor entirely would also fail - both directions are pinned.
- Reuses the existing `spec_builder` ground-detection rule instead of inventing a second, subtly different one, and documents the cost (spec parse vs full compile) in the helper's docstring.

PR #2419, opened 2026-08-18
docs(annotation): say which camera the steerable labels are derived from

cagataycali:
## Problem

`docs/data/annotation.md` tells you how to run `lerobot-annotate` against a
`strands-robots` dataset, but not which camera the labels come from. On a
multi-camera dataset that turns out to be load-bearing, and the answer is
decided by a `strands-robots` argument the page never mentions.

`lerobot-annotate` does not label per camera. Its `plan` module (subtasks, plan,
memory) and its `interjections` module read **one** stream:

- `VideoFrameProvider.__post_init__` sets `self.camera_key = keys[0]`
  (`lerobot/annotations/steerable_pipeline/frames.py`), where `keys` is
  `LeRobotDatasetMetadata.video_keys` - the dataset's video features in
  insertion order.
- `modules/plan_subtasks_memory.py` and `modules/interjections_and_speech.py`
  call `frames_at(record, timestamps)` with no `camera_key`, so they inherit
  that default. Only `modules/general_vqa.py` passes one explicitly, which is
  why `vqa`/`trace` rows carry a `camera` field and every other style carries
  `camera=None`.

`start_recording` writes its video features in the caller's `cameras=` order, so
the first name in that list becomes `keys[0]`:

| how you recorded | first video key |
| --- | --- |
| no `cameras=` | `observation.images.default` - the implicit overview camera |
| `cameras=[...]` | the first name in your list |

Verified against a recorded dataset with lerobot's own class:

```
>>> VideoFrameProvider(root=<recorded dataset>).camera_key
'observation.images.cam_wrist'
```

The gap is that the recommended path leads to the bad case. The
`start_recording` warning already tells you to pass `cameras=` so the implicit
`default` overview view stays out of the dataset - and doing that replaces a
world-fixed first key with whichever sensor you happen to list first.

## Why a gripper-mounted first camera is the wrong stream

A camera added with `add_camera(..., parent_body="<arm>/gripper")` moves with
whatever the gripper is holding, so its image evidence about object motion runs
*opposite* to the scene. One recorded episode, two conditions differing only in
whether the manipuland is held, tracking the manipuland's centroid in each
recorded 256x256 video (90 frames), ground truth from `d.xpos`:

| recorded view | carried 0.3049 m | left on its stand (0.0198 m) |
| --- | --- | --- |
| world-fixed, front | 174.4 px | 13.7 px |
| world-fixed, overhead | 174.2 px | 16.3 px |
| gripper-mounted (wrist) | **2.0 px**, visible 90/90 | **140.0 px**, visible 46/90 |

A carried object is pinned in the wrist frame; a *stationary* object is what
sweeps across it and leaves. So a labeller reading that stream has the evidence
for "the object moved" exactly when the object did not move. The wrist view is
also the one where the manipuland is *largest* (8-10% of frame vs 1-2% for the
world-fixed views), so this is not a visibility or resolution problem - it is
which frame the camera is rigid to.

camera comparison

The same 0.3049 m transport, as the three recorded videos show it. The cube
travels across both world-fixed views and stays pinned in the wrist view:

rollout

## Change

Docs only, plus a test. A new "Which camera the labels come from" section states
the selection rule, both recording paths, the gripper-mounted hazard with the
measured numbers, and the two ways to fix it - `--vlm.camera_key` (already
supported by `VlmConfig`, previously unmentioned here) or listing a world-fixed
camera first when recording. It also adds `--vlm.camera_key` to the existing
"Common flags" list. No behaviour changes: the camera-selection default is
lerobot's, and the recording order deliberately follows the caller.

## Tests

`tests/test_docs_annotation_camera_selection.py`, two kinds of assertion:

- **Fails before this change** (3 tests): the guide must name the
  camera-selection field that lerobot's `VlmConfig` actually exposes - read out
  of that dataclass with `ast` rather than hard-coded, so a rename upstream
  fails the test instead of rotting the doc - and must mention `parent_body` and
  `cameras=`.
- **Passes before and after** (1 test): the behavioural control. A real
  recording with `cameras=["front", "wrist"]` must put `observation.images.front`
  first, because the guidance is only actionable while the recorded order
  follows the caller. Confirmed non-vacuous by inverting the caller's list,
  which fails with the expected ordering diff.

```
branch          4 passed
upstream/main   3 failed, 1 passed
```

## Verification

- `mypy tests/test_docs_annotation_camera_selection.py`: no issues.
- Doc-scanning and repo-wide guard selection (23 files, including
  `test_docs_*`, docstring, changelog and source-string guards):
  661 passed, 2 skipped.
- No added non-ASCII in the edited page; fence count even; every markdown table
  block has a consistent cell count.

### DIFF
--- a/changelog.d/2419-annotation-camera-selection.md
+++ b/changelog.d/2419-annotation-camera-selection.md
@@ -1,0 +1,9 @@
+### Documentation
+
+- **docs/data/annotation.md**: document which camera `lerobot-annotate` derives
+  its steerable labels from. The `plan` and `interjections` modules read only the
+  dataset's first video key, and `start_recording` writes video features in the
+  caller's `cameras=` order, so that list decides the view. Records the
+  gripper-mounted hazard - a camera parented to the gripper renders a carried
+  object as image-static and a stationary object as sweeping out of frame - and
+  names the `--vlm.camera_key` escape hatch.
--- a/docs/data/annotation.md
+++ b/docs/data/annotation.md
@@ -83,7 +83,64 @@ Common flags:
 - `--vlm.api_base=http://host:8000/v1` with `--vlm.auto_serve=false` - reuse a
   running vLLM server instead of spawning one.
 - `--only_episodes=0,1,2` - annotate a subset while iterating.
+- `--vlm.camera_key=observation.images.<name>` - choose which camera the
+  `plan` and `interjections` modules read; see [Which camera the labels come
+  from](#which-camera-the-labels-come-from).
 - toggle modules with `--plan.enabled`, `--interjections.enabled`, `--vqa.enabled`.
+
+## Which camera the labels come from
+
+A multi-camera dataset does not get one label per camera. The `plan` module
+(subtasks, plan, memory) and the `interjections` module read **one** stream: the
+dataset's *first* video key. `VideoFrameProvider` picks it in
+`lerobot/annotations/steerable_pipeline/frames.py` (`self.camera_key = keys[0]`)
+and those two modules call `frames_at()` without naming a camera, so they
+inherit that default. Only the `vqa` module iterates every camera, which is why
+`vqa`/`trace` rows carry a `camera` field and every other style carries
+`camera=None`.
+
+The first video key is the one **you** chose when you recorded:
+
+| how you recorded | first video key |
+| --- | --- |
+| `start_recording(...)` with no `cameras=` | `observation.images.default` - the implicit overview camera `create_world` adds |
+| `start_recording(..., cameras=[...])` | the **first name in your list** |
+
+So the `cameras=` list that scopes a dataset to its real sensors - the list the
+`start_recording` warning tells you to pass - also decides which view every
+subtask label is derived from.
+
+**Do not let that first key be a gripper-mounted camera.** A camera added with
+`add_camera(..., parent_body="<arm>/gripper")` travels with whatever the gripper
+is holding, so its image evidence about object motion is *inverted* with respect
+to the scene. Measured in simulation on one recorded episode, tracking the
+manipuland's centroid in each recorded video (256x256, 90 frames):
+
+| recorded view | object carried 0.3049 m | object left on its stand |
+| --- | --- | --- |
+| world-fixed, front | 174.4 px of image motion | 13.7 px |
+| world-fixed, overhead | 174.2 px | 16.3 px |
+| gripper-mounted (wrist) | **2.0 px** (visible 90/90) | **140.0 px** (visible 46/90) |
+
+A carried object is pinned in the wrist view; a *stationary* object is what
+sweeps across it and leaves the frame. A labeller reading that stream has the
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Docs-only change (+230/-0 across a doc section, a changelog fragment, and a guard test) documenting which camera `lerobot-annotate`'s `plan`/`interjections` modules derive labels from: the dataset's first video key, which `start_recording` writes in the caller's `cameras=` order. The new section states the selection rule, the gripper-mounted hazard with measured numbers, and both remedies (`--vlm.camera_key`, or listing a world-fixed camera first). No behaviour changes.

## What's good

- The doc-honesty tests read the `camera_key` field out of lerobot's `VlmConfig` via `ast` rather than hard-coding it, so an upstream rename fails the test instead of rotting the doc; both lerobot-dependent tests degrade to clean skips when the annotation pipeline or a GL context is absent (verified: 2 passed, 2 skipped headless; the behavioural recording test passes under EGL).
- The behavioural test complies with AGENTS.md rule 15 (recording test names its own dataset root under `tmp_path`), cleans up the `Simulation` in a `finally`, and asserts on the recorded `meta/info.json` rather than internal state.
- Changelog recorded as a `changelog.d/` fragment, not an append to `CHANGELOG.md`; the edited doc page is ASCII-clean with an even fence count and a correct anchor link; `ruff check`/`ruff format --check` clean on the new test.
- All simulation API usages in the test (`add_robot`, `add_camera(parent_body=...)`, `start_recording(cameras=..., root=..., overwrite=...)`, `run_policy`, `cleanup`) verified against the head tree's actual signatures.

yinsong1986 (approved):
## Summary

Docs-only change (+232/-0) that documents which camera `lerobot-annotate`'s `plan`/`interjections` modules derive steerable labels from: the dataset's first video key, which `start_recording` writes in the caller's `cameras=` order. The new section in `docs/data/annotation.md` states the selection rule, both recording paths, the gripper-mounted hazard (with measured pixel-motion numbers showing a wrist camera inverts the image evidence for object motion), and the two remedies (`--vlm.camera_key`, or listing a world-fixed camera first). No behaviour changes anywhere.

## What's good

- The guard test reads the `camera_key` field out of lerobot's `VlmConfig` via `ast` rather than hard-coding it, so an upstream rename fails the test instead of silently rotting the doc.
- The behavioural control (`test_first_video_key_is_the_callers_first_camera`) records a real dataset and passes `root=str(tmp_path / "ds")`, complying with AGENTS.md convention 15 (a recording test names its own dataset root); GL/lerobot/mujoco availability is skipped cleanly and `sim.cleanup()` runs in `finally`.
- Changelog fragment at `changelog.d/2419-annotation-camera-selection.md` per the PR workflow (no direct `CHANGELOG.md` edit).
- Scope discipline: the doc explicitly notes the camera-selection default is lerobot's and the recording order deliberately follows the caller, so nothing here becomes a compatibility surface this repo owns.

PR #2420, opened 2026-08-18
fix(simulation): carry a kinematically attached body through run_multi_policy

cagataycali:
## Problem

`attach_bodies(mode="kinematic")` documents its child as following its parent **"every physics step"**, applied by one hook the backend calls after `mj_step`. Four functions in the MuJoCo backend advance physics:

| stepping path | re-applies the follow |
|---|---|
| `simulation.py::step` (batch loop) | yes |
| `rendering.py::_apply_sim_action` (single-robot policy) | yes |
| `motion_primitives.py::_primitive_tick` | yes |
| `simulation.py::run_multi_policy` (synchronized multi-robot) | **no** |

So a body attached for a two-arm handover was simply not carried on the multi-robot loop. It fell to the floor while `attach_bodies` reported success, `run_multi_policy` reported success, and the attachment registry still named it as carried. `run_multi_policy` is the recommended path for recording concurrent multi-robot episodes, so those frames went into the dataset showing the arms working over an object lying on the ground.

Measured on two `so101` arms with a cube attached to the first arm's gripper, 300 synchronized steps at 50 Hz — the same rollout in both trees, differing only in `strands_robots`:

| path | `\|relpos - recorded\|` | cube z start -> end | dist. to gripper end | carried |
|---|---|---|---|---|
| `step()` | 0.000087 m | 0.3500 -> 0.2999 | — | yes |
| `run_policy` (1 robot) | 0.000822 m | 0.3500 -> 0.1846 | — | yes |
| `run_multi_policy` **before** | **0.301564 m** | 0.3000 -> **0.0220** (on the floor) | **0.2522 m** | **no** |
| `run_multi_policy` after | 0.002439 m | 0.3000 -> 0.2176 | 0.0865 m | yes |

After the change the cube stays within 1.6 mm of the 0.0870 m grasp offset for all 300 steps.

## Change

One line: re-apply the follow after the loop's single `mj_step`, mirroring the other policy-driven stepping path (`_apply_sim_action`), which calls the same hook and relies on its fast no-op when the registry is empty. Behaviour with no kinematic attachment registered is unchanged.

`_apply_kinematic_attachments`' docstring enumerated two of its callers; it now states the contract that every stepping path has to satisfy, which is what the new guard enforces. `attach_bodies` has no page under `docs/`, so its docstring is its documentation and already stated the promise correctly.

## Tests

`tests/simulation/mujoco/test_kinematic_attachment_follows_on_every_stepping_path.py` — hermetic (inline MJCF arm reusing the sibling suite's `_ROBOT_XML`, no assets, no GL).

**3 failed / 5 passed on `upstream/main` -> 8 passed.** The three that fail pre-fix:

- the contract: `run_multi_policy left the carried body 0.6459 m from the offset the attachment recorded (z 0.4494 -> 0.0145)`
- the user-visible harm, read through public `get_body_state`: `the carried body ended at z=0.0145 m, on the ground rather than with its parent 'alpha/link2', while run_multi_policy and attach_bodies both reported success`
- the root cause, as a source-level guard: `these MuJoCo backend functions call mj_step without re-applying the kinematic follow: ['simulation.py::run_multi_policy']`

The guard is a sweep rather than a per-path rollout, so a fifth stepping path added later fails there instead of silently dropping carried bodies. It carries a non-vacuity test (at least four `mj_step` callers must be found) so a refactor that hides them fails loudly. It is scoped to the MuJoCo backend, which owns the attachment registry; an `action_controller` that takes over stepping (LIBERO, WBC torque control) replaces the loop wholesale and is a separate contract.

The five that pass on both sides are the boundary: `step()` and single-robot `run_policy` still carry, a weld attachment is held by the solver and never used the hook, and a loop with no attachments leaves an unattached airborne body falling normally.

The carry tolerance is derived rather than invented — one integration step of latency is inherent to the follow, so the residual scales with parent speed times timestep; on this fixture the carrying paths land within 0.026 m while an unheld child reaches 0.65 m, and the bound sits about 2x above the former and 13x below the latter.

## Rollout

Two `so101` arms under `run_multi_policy`, cube attached to the first arm's gripper. Same script in both panels, only `strands_robots` differs. The panels are pixel-identical at step 0 (mean |delta| 0.0001) and diverge to 3.92 as the change takes effect; 74/74 frame transitions move.

kinematic carry through run_multi_policy

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-kinematic-carry-32086527241/kinematic-carry.mp4)

## Verification

- `mypy strands_robots tests tests_integ`: 19 errors, identical to pristine `upstream/main`, none in the changed files.
- 44-file selection (every test referencing `attach_bodies` / `kinematic_attachments` / `run_multi_policy` / `manipulation`, plus the repo-wide docstring, ASCII, dependency and changelog guards): branch **2 failed / 2420 passed**, base **5 failed / 2417 passed**; same 2422 collected on both, zero branch-only failures, the three base-only ids are exactly the pre-fix failures above. The two shared failures are a pre-existing `draccus` version mismatch in `tests/training/test_lerobot.py`, unrelated to this change.

### DIFF
--- a/changelog.d/2420-kinematic-attachment-multi-policy.md
+++ b/changelog.d/2420-kinematic-attachment-multi-policy.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **simulation/mujoco**: `run_multi_policy` now re-applies the kinematic-attachment follow after its physics step, so a body attached with `attach_bodies(mode="kinematic")` is carried through a synchronized multi-robot rollout instead of being left behind while every surface reports success.
--- a/strands_robots/simulation/mujoco/manipulation.py
+++ b/strands_robots/simulation/mujoco/manipulation.py
@@ -373,8 +373,12 @@ class ManipulationMixin:
     def _apply_kinematic_attachments(self) -> None:
         """Teleport every kinematically attached child to follow its parent.
 
-        Called after each ``mj_step`` (both the ``step()`` batch loop and the
-        policy-driven ``_apply_sim_action`` substep loop). Callers MUST hold
+        Called after EVERY ``mj_step`` the backend issues - the ``step()``
+        batch loop, the single-robot ``_apply_sim_action`` substep loop, the
+        motion-primitive tick, and the synchronized ``run_multi_policy`` loop -
+        because ``attach_bodies(mode="kinematic")`` promises the child follows
+        every physics step, and a stepping path that skips this leaves the
+        carried body behind while reporting success. Callers MUST hold
         ``self._lock``. Uses the parent's ``xpos``/``xquat`` from the step's
         forward pass (one integration step of latency at the physics timestep,
         matching the example's carry). Entries whose bodies or freejoint no
--- a/strands_robots/simulation/mujoco/simulation.py
+++ b/strands_robots/simulation/mujoco/simulation.py
@@ -5507,6 +5507,11 @@ class MuJoCoSimEngine(
                         pfx = robot.namespace or ""
                         self._apply_action_by_name(self._world._model, self._world._data, act, pfx, mj)
                     mj.mj_step(self._world._model, self._world._data)
+                    # Kinematic attachments (attach_bodies mode="kinematic")
+                    # follow their parent every physics step, on this
+                    # synchronized loop as much as on the single-robot policy
+                    # path. Fast no-op when none are registered.
+                    self._apply_kinematic_attachments()
                     self._world.sim_time = self._world._data.time
                     self._world.step_count += 1
                     if hasattr(self, "_viewer_handle") and self._viewer_handle is not None:
--- a/tests/simulation/mujoco/test_kinematic_attachment_follows_on_every_stepping_path.py
+++ b/tests/simulation/mujoco/test_kinematic_attachment_follows_on_every_stepping_path.py
@@ -1,0 +1,287 @@
+"""``attach_bodies(mode="kinematic")`` carries its child on EVERY stepping path.
+
+``attach_bodies`` documents the kinematic mode as "every physics step, the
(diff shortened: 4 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes the one MuJoCo stepping path (`run_multi_policy`) that advanced physics without re-applying the kinematic-attachment follow, so a body attached with `attach_bodies(mode="kinematic")` is now carried through synchronized multi-robot rollouts instead of silently falling to the floor while every surface reported success. The production change is a single hook call inserted immediately after the loop's `mj_step`, inside the existing `with self._lock:` block — which satisfies `_apply_kinematic_attachments`'s documented "callers MUST hold `self._lock`" contract and exactly mirrors the three paths that already call it (`step()`, `_apply_sim_action`, `_primitive_tick`). The hook is a fast no-op on an empty registry and defensively drops stale entries, so behavior with no attachment registered is unchanged, as the no-overreach control test confirms.

## What's good

- Minimal, correctly-scoped diff: one call site plus a docstring that now states the every-stepping-path contract rather than enumerating callers.
- The AST sweep test turns the root cause into a standing guard — a fifth `mj_step` caller added later fails loudly instead of silently dropping carried bodies — and carries its own non-vacuity check so a refactor that hides the call sites can't make it pass emptily.
- Boundary controls pin what must not change: `step()` and single-robot `run_policy` still carry, weld mode never used the hook, and an unattached body still falls.
- Changelog fragment follows the `changelog.d/` convention; test is hermetic (inline MJCF, no assets, no recording, no shared-cache reads); tolerance is derived from the measured latency floor rather than invented.

PR #2421, opened 2026-08-18
fix(simulation/mujoco): find a floating base on a robot that declares no joint

cagataycali:
## Problem

A floating base that is not a named entry in `robot.joint_names` is recovered by walking up the kinematic tree from a seed part of the machine (`_robot_base_free_joint`). That fallback could only be seeded one way — from a body carrying one of the robot's **declared joints**:

```python
for jnt_name in robot.joint_names:      # empty for an aerial robot
    ...                                  # -> loop body never runs
    body = int(model.jnt_bodyid[jnt_id]) # -> walk never starts
return -1
```

So the rung whose stated purpose is finding a base that is *absent from* `robot.joint_names` begins by iterating `robot.joint_names`. For a robot whose only joint **is** that unnamed base, the list is empty and both rungs no-op.

An aerial robot is exactly that shape, and it is the stock one. Each rotor is a force applied at a site on the airframe, so the transmission is `mjTRN_SITE` and the model declares no joint besides the unnamed `<freejoint>`:

```
skydio_x2:  njnt=1  jnt 0: name=None type=mjJNT_FREE body='skydio_x2/x2'
            nu=4    thrust1..4  trntype=mjTRN_SITE
            robot.joint_names = []          _robot_free_base_joint_id -> -1
```

Across the 63 registry sim robots, `mjTRN_SITE` is used by **exactly the two robots whose base was missed** — `crazyflie` and `skydio_x2`.

## Impact

The helper is the single owner for three surfaces, so all three went quiet rather than wrong:

| surface | before |
|---|---|
| `get_observation(robot_name=...)` | `{}` — no `base_pos`, no `base_quat`, no velocities, for a robot in the scene and moving |
| `get_robot_state` | no base to report |
| `start_recording` | declares **no `observation.state` column at all**, while still declaring the `action` columns |

Measured on a real rollout (`skydio_x2`, four rotors, climb-then-hold): the drone climbs from 0.101 m to **2.824 m** and the recorded LeRobotDataset's non-image features are

```
action, timestamp, frame_index, episode_index, index, task_index
```

`stop_recording` reports `success — 60 frames, 1 episode(s)`. The episode captures a demonstration with images and actions and nothing to condition a policy on; a state-conditioned policy (ACT, diffusion, SmolVLA) cannot be trained from it, and nothing in the result says so.

The repository already treats this as a defect: `tests/simulation/mujoco/test_recording_preserves_floating_base_state.py` was added because dropped base signals leave "a locomotion / velocity-tracking / whole-body-control policy trained on the dataset **base-blind**", and `test_mobile_base_observation.py` documents the mechanism as "walk up from an **actuated joint**". Both suites only ever build a robot that has one.

## Change

Seed the same walk from the bodies the robot's own actuators act on as well, resolved through a new `scene_ops.actuator_target_body_ids`. The seed is **additive** — a robot whose base already resolved answers through the joint seed first — and the shared ancestor walk is factored into `_ancestor_free_joint` so both seeds resolve identically.

`actuator_target_body_ids` is the third member of the transmission-reading family, and a genuinely different question from the two joint readers: `actuator_joint_id` and `actuator_driven_joint_ids` are both **empty** for a site, body or slider-crank transmission, which is precisely what an airframe uses. Pinned as a test:

```python
assert actuator_joint_id(model, 0, mj) == -1
assert actuator_driven_joint_ids(model, 0, mj) == frozenset()
```

Actuator ownership is resolved by the existing `robot_owned_actuator_ids` rather than re-derived, and the body the seed lands on is then required to carry the robot's namespace.

That second requirement is load-bearing, not defensive. `robot_owned_actuator_ids` recognizes an actuator two ways — a namespaced name, or **the joint id it drives** — and a stored id does not survive a recompile. `replace_scene_mjcf` compiles caller-supplied MJCF without rewriting the registry, so afterwards a robot's stored `joint_ids` address whatever those indices mean in the *new* model. Without the name check the seed claims a foreign actuator through that collision:

```
robot 'mob'  namespace='mob/'  joint_names=['shoulder']  joint_ids=[1]   (stale)
replaced model:  joints [(0, None, FREE), (1, 'elbow', HINGE)]   actuators ['elbow_act']
  robot_owned_actuator_ids -> ['elbow_act']     (actuator_joint_id == 1 in {1})
  seed body 'other_arm' -> walk up -> free joint on 'other_plate'
```

and reports an unrelated machine's base as this robot's. A name cannot be borrowed that way. `test_base_state_disappears_when_the_bare_name_misses_too` in `test_namespace_fallback_survives_scene_replace` is what pins this — it is that module's non-vacuity control, and it caught the first revision of this change. Removing the name check makes it fail with `base state reported for a robot with no resolvable joint`.

### Why not "any free joint under the robot's namespace"

That is the shorter fix and it is wrong. `spec.attach` prefixes every name it merges, so a task object shipped inside the robot's own MJCF arrives under the robot's namespace — which is how every Menagerie grasping scene is authored:

```
robotiq_2f85     freejoint 8  on body 'robotiq_2f85/object'
allegro_hand     freejoint 16 on body 'allegro_hand/object'
shadow_hand      freejoint 24 on body 'shadow_hand/object'
robotiq_2f85_v4  freejoint 6  on body 'robotiq_2f85_v4/object'
```

A namespace rule reports the grasp cube's pose as the gripper's base state on four stock robots. The actuator seed keeps the existing "only the robot's OWN base" guarantee for the same reason the joint seed did: no actuator of the robot acts on the cube. `test_a_fixed_base_gripper_shipped_with_a_task_object_has_no_base` is the control — with the namespace variant substituted it fails with `fixed-base gripper reported base state from a task object: ['base_pos', 'base_quat', 'base_lin_vel', 'base_ang_vel']`, while the drone tests still pass.

## Blast radius

Sweep over all 63 registry sim robots, comparing the resolved base joint and the observation key count before and after:

| | count |
|---|---|
| gain a base (`-1` -> resolved; observation `0` -> `4` base signals) | **2** — `crazyflie`, `skydio_x2` |
| ship a free-jointed task object, still correctly report no base | **4** — the grippers above |
| answer identically | **57** |

No robot changes which free joint it resolves, and no robot loses one.

## Tests

`tests/simulation/mujoco/test_free_base_found_without_a_declared_joint.py`, hermetic inline MJCF (no assets, no network). Against `upstream/main`: **8 failed / 5 passed**; with the change **13 passed**.

Failing pre-fix, with the message each produces:

- `floating base not surfaced: ['base_pos', 'base_quat', 'base_lin_vel', 'base_ang_vel'] absent from []`
- `recorded dataset declared no observation.state column at all` (`assert []`)
- `KeyError: 'observation.state'` reading the reopened dataset back
- the same with a task object present, so the base must be resolved rather than guessed
- four `actuator_target_body_ids` cases (the new helper)

Passing on both trees — the bounds of the change:

- a fixed-base gripper shipped with a free-jointed task object gains no base state
- a fixed-base arm still has none
- a named `floating_base_joint` (humanoid) and a declared-joint-seeded unnamed base (mobile) answer unchanged, asserted on the base position so a wrong body cannot pass
- neither joint reader answers for a site transmission

End-to-end: the recorded dataset is reopened through `LeRobotDataset` and frame 0's `observation.state[:3]` is asserted to be the airframe's spawn pose.

## Verification

- `mypy strands_robots tests tests_integ`: no error mentions a changed file; branch-only diff against `upstream/main` is empty.
- `changelog.d` fragment added; `scripts/assemble_changelog.py --check` passes.
- Full `tests` suite, this branch vs `upstream/main`, same interpreter, run concurrently so both see the same load: branch **8 failed / 31854 passed / 2 errors**, base **7 failed / 31842 passed / 2 errors** (1396.8 s vs 1397.5 s). `31854 - 31842 = 12 = 13 new tests - 1`, and that 1 is accounted for below.
- Comparing sorted failing ids: 7 are shared and pre-existing (two `tests/policies/vera` cases needing Docker, the GL probe's force-skip meta-test, a Newton message-text assertion, two `tests/mesh` node-id errors, one pytest-exit meta-test). The only differences are **3 branch-only and 2 base-only ids, all inside `tests/mesh/test_zenoh_transport_security.py`** - it opens real localhost Zenoh sessions, so two concurrent suites cross-talk and it rotates which subset fails. Run alone it is `9 passed, 2 skipped` on both trees, twice each. No file this change touches appears in either set.

## Artifact

Real `skydio_x2` rollout, MuJoCo headless. The same flight runs identically on both trees (`d.qpos[2]`: 0.101 m -> 2.824 m, byte-identical); only the recording differs. After the change the recorded `observation.state[base_pos.z]` tracks the true altitude to within 0.062 m — the one-step offset between when the observation is sampled and when the physics step lands.

floating base recorded for a thrust-actuated robot

[Flight video (MP4)](https://raw.githubusercontent.com/cagataycali/robots/media-free-base-seed/drone_climb.mp4)


yinsong1986 (dismissed):
## Summary

Seeds the floating-base kinematic-tree walk from the bodies a robot's own actuators act on, in addition to the existing declared-joint seed, so a robot whose only joint is its unnamed `<freejoint>` (aerial robots: `crazyflie`, `skydio_x2`, whose rotors are `mjTRN_SITE` transmissions and whose `robot.joint_names` is therefore empty) surfaces `base_pos`/`base_quat`/velocities in `get_observation`, `get_robot_state`, and the recorded `observation.state` schema. The seed resolution is factored through a new `scene_ops.actuator_target_body_ids`, which branches on `actuator_trntype` before reading `actuator_trnid` (with bounds checks per id space), and actuator ownership is reused from the existing `robot_owned_actuator_ids` rather than re-derived. The additive-seed design deliberately rejects the shorter namespace-membership fix, and the control test proves why: a Menagerie-style free-jointed task object shipped under the robot's namespace still reports no base.

## What's good

- The regression suite is hermetic (inline MJCF, no assets) and pins the pre-fix behaviour: verified locally that it fails 8/13 on `upstream/main` and passes 13/13 on this head, matching the description's claim exactly.
- Existing floating-base suites (`test_recording_preserves_floating_base_state.py`, `test_mobile_base_observation.py`) pass unchanged on this head, and both prior seeds are asserted on distinctive base positions so a wrong body cannot pass.
- Recording tests pass an explicit `root=` (AGENTS.md rule 15), the changelog fragment validates against `tests/test_changelog_fragments.py`, the new public helper is registered in `test_public_member_docstrings.py`, and `ruff check` / `ruff format --check` are clean on the changed files.
- The joint seed's existing behaviour (including the `break` after the first resolvable joint) is preserved byte-for-byte before the actuator seed is consulted, so the 57 unchanged registry robots answer through the same path they did before.

yinsong1986 (approved):
## Summary

Fixes the floating-base recovery walk so it can be seeded from the bodies a robot's own actuators act on, not only from a body carrying a declared joint. This makes the fallback reachable for the exact shape it was written for: an aerial robot whose rotors are `mjTRN_SITE` transmissions declares no joint besides its unnamed `<freejoint>`, so `robot.joint_names` is empty, the walk never started, and `get_observation` / `get_robot_state` / `start_recording` all went quiet (a recorded episode carried actions and images but no `observation.state` column, while reporting success). The new `scene_ops.actuator_target_body_ids` resolves each transmission's target body from the id space its own `trntype` selects, ownership is routed through the existing `robot_owned_actuator_ids`, and the seed body plus the resolved free joint's body are both required to carry the robot's namespace — which correctly closes the stale-joint-id borrow after `replace_scene_mjcf` and correctly refuses the free-jointed task object every Menagerie grasping scene ships under the robot's own namespace.

## What's good

- Verified locally at head `ffe3144a`: the new module is 13/13 green on the branch and, transplanted onto `upstream/main`, fails 8 / passes 5 — exactly matching the PR's non-vacuity claim, with the 5 passes being the bounds-of-change controls (fixed-base gripper with task object, fixed arm, named/unnamed bases asserted on position).
- Related existing suites (`test_recording_preserves_floating_base_state`, `test_mobile_base_observation`, `test_namespace_fallback_survives_scene_replace`, `test_actuator_ownership_by_transmission`, `test_tendon_actuator`, `test_public_member_docstrings`) all pass on the branch; `ruff check` / `ruff format --check` clean on the changed files.
- Recording tests name their own dataset root (`root=tempfile.mkdtemp(...)` at every `start_recording` site), per AGENTS.md rule 15.
- `actuator_target_body_ids` branches on `trntype` before indexing, so no cross-id-space collision (the same discipline `actuator_joint_id` documents), with bounds checks on every id read; the tendon rule delegates to `tendon_joint_ids` rather than re-deriving the wrap-table walk.
- The change is read-only on `model` (no `mjData`/model mutation), additive in seed order (joint seed answers first), and the 63-robot blast-radius sweep in the description matches what the code shape implies: only joint-less namespaced robots can take the new path.
- New public helper carries a full `Args:`-sectioned docstring and is registered in `test_public_member_docstrings.py`; changelog fragment present.

cagataycali:
Note on the check history, so the earlier states are not misread.

The first revision of this branch genuinely failed `Test and Lint` on one test, and that failure was correct: `test_namespace_fallback_survives_scene_replace.py::test_base_state_disappears_when_the_bare_name_misses_too`. It is that module's non-vacuity control, and it caught a real over-reach — the actuator seed was claiming a foreign machine's actuator through a stale joint id after `replace_scene_mjcf`. The current revision adds the namespace check on the seed body and that module is back to `9 passed` (the "Why not any free joint under the robot's namespace" section above has the detail).

The `cancelled` state on the current head is not a test result: editing this description re-triggered the `pull_request` workflows, and the concurrency group cancelled the in-flight job at the dependency-install step — `Run lint` and `Run tests` are both reported `skipped`, so nothing about the diff was evaluated. A re-run is queued.

The full `tests` suite was run locally on this branch and on `upstream/main` concurrently, and the comparison is in the Verification section: zero failing ids that are not shared with `upstream/main`, with the only differences confined to `tests/mesh/test_zenoh_transport_security.py`, which rotates which subset fails when two suites share a host and passes `9 passed, 2 skipped` alone on both trees.

PR #2422, opened 2026-08-18
fix(simulation/mujoco): check actuate_robot's posture flags instead of reading them by truthiness

cagataycali:
## Problem

`actuate_robot` takes six parameters. Four scale a quantity - `kp`, `damping`,
`armature` - and each is checked on a shared numeric domain that refuses a
string by name and refuses `bool` (an `int` subclass that would pass as a silent
`1`). The other two select a *posture*, and both were read by truthiness after a
`bool()` that laundered whatever arrived into a well-typed `True` before it
reached the spec.

Every non-empty string is truthy, so the spellings a caller reaches for when
opting out selected the branch it was opting *out* of. Measured on MuJoCo
through the agent-facing router, which publishes both parameters as
`"type": "boolean"` and binds them against the method signature without
checking the value:

| `disable_self_collision=` | verdict | `contype`/`conaffinity` on the robot's geoms |
| --- | --- | --- |
| `False` | success | `(1, 1)` - kept, as asked |
| `"no"` / `"false"` / `"off"` / `"0"` / `nan` | **success** | **`(0, 0)`** - every geom, opt-out inverted |
| `None` / `0.0` | success | `(1, 1)` - never a declared spelling of either posture |
| `True` | success | `(0, 0)` - disabled, as asked |
| `damping="2.0"` (same signature) | error | `'damping' must be a finite number >= 0, got '2.0'.` |

`gravity_compensation="no"` reached the spec the same way and set `gravcomp=1`
on the robot's bodies.

Two things make the `disable_self_collision` case the sharper one. Its own
docstring warns the branch "disables ALL collision on the robot's geoms (not
just link-vs-link)", so the robot stops colliding with the floor and with every
object in the scene, not only with itself. And the change lives on the spec, as
documented, so it outlives the recompile and undoing it needs a separate
`patch_scene_mjcf` call.

## Change

Both flags now go through `strands_robots.utils.boolean_flag_error`, ahead of
`_ensure_mujoco()` and ahead of anything written to the spec, so a rejected
call leaves the robot exactly as it found it - no actuators added, `nu` still 0.
The redundant `bool()` at the call site is gone: it was the laundering that let
the inverted value reach `actuate_robot_in_scene` well-typed, where nothing
downstream could tell it apart from a real `True`.

That is the domain the sibling `set_joint_positions`' `hold` flag in this same
backend already applies, and the one `start_recording`'s `overwrite` /
`push_to_hub` were moved onto after the same failure. Validation stays at the
one public entry point rather than being repeated in `actuate_robot_in_scene`,
whose only caller this is.

Scope: this is `actuate_robot`'s signature, where the numeric half was already
checked and the posture half was not. Other boolean parameters elsewhere in the
backend are a wider question about which of them are postures worth bounding,
and are deliberately left alone here.

## Tests

`tests/simulation/mujoco/test_actuate_robot_posture_flag_domain.py`, **47
failed / 16 passed before, 63 passed after**. The headline reads:

```
AssertionError: disable_self_collision='false' was accepted; the robot's geoms
are now [(0, 0), (0, 0), (0, 0)] (was [(1, 1), (1, 1), (1, 1)])
```

Each refusal test also asserts the compiled model is untouched (`nu == 0`,
collision and `body_gravcomp` unchanged), and one drives the agent-facing router
rather than the method, because that is the reachable route.

The 16 that pass on both trees are the controls, and they are what fails if the
guard reaches too far: a python or numpy boolean is still accepted for both
flags, `True` still disables collision while `False` still keeps it, `True`
still compensates gravity while `False` does not, the success text still names
both postures, and the four numeric siblings keep their own domain - including
their rejection of `bool`, since the two domains are inverses and must not be
conflated. One test is keyed on `boolean_flag_error` itself so the accepted
domain cannot drift here alone.

## Verification

- Blast radius: every test referencing `actuate_robot`, `boolean_flag_error`,
  `actuate_robot_in_scene`, `disable_self_collision` or
  `_apply_kinematic_attachments`, plus the repo-wide docstring / ASCII /
  changelog / tool-spec guards. Same selection on both trees at the pre-merge
  base: branch **1487 passed, 0 failed**; base **1440 passed, 47 failed**.
  `1440 + 47 == 1487`, both collected 1488, every base-only failure is in the
  new file, zero branch-only failures. After merging `main` (#2419 and #2420
  landed, and #2420 touches this same module) the selection is 33 files and the
  composed tree is **1797 passed, 0 failed**, with the pre-fix split re-measured
  against the new base and unchanged at 47 / 16.
- `ruff check` / `ruff format --check` and `mypy` over `strands_robots tests
  tests_integ` clean; mypy is 19 pre-existing errors on both trees, none in the
  touched files.

One capture script per tree, same URDF arm, same crate, same sweep. The caller
wants the arm's geoms to keep colliding and spells the opt-out `"no"`; where the
call is refused the script reads the refusal and supplies the boolean it asks
for. Only `strands_robots` differs between the panels.

actuate_robot posture flag

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-actuate-posture-flags/collision-flag.mp4)

Left: `"no"` accepted, `contype` `[0, 0, 0]`, the arm sweeps straight through the
crate, which travels **0.11 mm** over the whole sweep. Right: refused, retried
with `False`, `contype` `[1, 1, 1]`, the arm pushes the crate **54.58 mm**. The
panels are pixel-identical at frame 0 (mean abs delta 0.0000) and 5.97 apart at
the end, so they differ only where the flag bites. A rollout recorded against
the left scene contains no contacts at all.

The headlight is lifted in the capture script for legibility - this scene has no
MJCF `<light>` - which changes nothing physical and nothing measured above.

### DIFF
--- a/changelog.d/2422-actuate-robot-posture-flags.md
+++ b/changelog.d/2422-actuate-robot-posture-flags.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- **simulation/mujoco**: `actuate_robot` now checks `gravity_compensation` and
+  `disable_self_collision` on the shared boolean domain instead of reading them
+  by truthiness. Both are published as `"type": "boolean"` on the agent tool
+  surface, and both were laundered through `bool()` before reaching the spec, so
+  `disable_self_collision="false"` (also `"no"`, `"off"`, `"0"`, `nan`) zeroed
+  `contype`/`conaffinity` on every one of the robot's geoms - the branch the
+  caller was opting out of - while `None` and `0.0` took the other branch
+  without being a declared spelling of it. Their four numeric siblings in the
+  same signature have been on a shared domain all along.
--- a/strands_robots/simulation/mujoco/manipulation.py
+++ b/strands_robots/simulation/mujoco/manipulation.py
@@ -43,6 +43,7 @@ from strands_robots.simulation.mujoco.scene_ops import (
     add_weld_constraint,
     remove_equality_constraint,
 )
+from strands_robots.utils import boolean_flag_error
 
 if TYPE_CHECKING:
     from strands_robots.simulation.models import SimRobot
@@ -523,18 +524,23 @@ class ManipulationMixin:
             armature: Per-joint armature (rotor inertia) floor; existing
                 larger values are kept. Must be finite and >= 0.
             gravity_compensation: Apply ``gravcomp=1`` to the robot's bodies
-                so modest gains track tightly.
+                so modest gains track tightly. Must be a boolean - a truthy
+                string such as ``"no"`` is refused rather than read as True.
             disable_self_collision: Zero contype/conaffinity on the robot's
                 own geoms. Planners like cuRobo ignore adjacent-link contacts,
                 which otherwise block planned motion in MuJoCo. NOTE: this
                 disables ALL collision on the robot's geoms (not just
                 link-vs-link), so add contact geoms that must keep colliding
                 (e.g. fingertip pads) AFTER this call via ``patch_scene_mjcf``.
+                Must be a boolean, for the same reason and more sharply: read by
+                truthiness, ``"false"`` would disable every collision on the
+                robot for a caller spelling the opt-out.
 
         Returns:
             ``{status, content}`` tool result; ``status="error"`` when no world
             exists, a policy is running, the robot is unknown or already driven,
-            ``kp``/``damping``/``armature`` are invalid, or the recompile fails.
+            ``kp``/``damping``/``armature`` are invalid, either posture flag was
+            not supplied as a boolean, or the recompile fails.
             The already-driven refusal names the joints it found and the
             actuators driving them, because what a caller does next differs by
             whether the robot is driven throughout or only in part.
@@ -556,6 +562,19 @@ class ManipulationMixin:
                 "status": "error",
                 "content": [{"text": f"actuate_robot: 'armature' must be a finite number >= 0, got {armature!r}."}],
             }
+
+        # Read by truthiness these two would invert for exactly the spellings a
+        # caller opting out reaches for: "false", "no", "off" and "0" are all
+        # truthy, and disable_self_collision then zeroes contype/conaffinity on
+        # every one of the robot's geoms - on the spec, so it outlives the
+        # recompile - while the caller asked to keep them colliding. Their
+        # numeric siblings above have been on a shared domain all along.
+        for flag_name, flag_value in (
+            ("gravity_compensation", gravity_compensation),
+            ("disable_self_collision", disable_self_collision),
+        ):
+            if text := boolean_flag_error(flag_value, flag_name, "actuate_robot"):
+                return {"status": "error", "content": [{"text": text}]}
 
         mj = _ensure_mujoco()
         with self._lock:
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Moves `actuate_robot`'s two posture flags (`gravity_compensation`, `disable_self_collision`) onto the shared `boolean_flag_error` domain instead of laundering them through `bool()`, closing the inversion where opt-out spellings like `"false"`/`"no"`/`"off"`/`"0"` selected the branch being opted out of - for `disable_self_collision`, zeroing `contype`/`conaffinity` on every robot geom, persisted on the spec past recompiles. The guard sits after the existing numeric checks and before `_ensure_mujoco()` / the lock / any spec write, so a refused call leaves the model untouched (`nu == 0`, collision and `body_gravcomp` unchanged - asserted by the new tests). The only consumer, `actuate_robot_in_scene`, reads the flags by truthiness and now only ever receives python/numpy booleans, both of which it handles correctly. The stricter surface returns an explicit error naming the flag and method rather than silently changing behaviour, matching the precedent already set for `set_joint_positions`' `hold` and `start_recording`'s `overwrite`/`push_to_hub`.

## What's good

- Reuses the existing shared guard rather than adding a single-caller one (AGENTS.md convention 11), and one test is keyed on `boolean_flag_error` itself so the accepted domain cannot drift here alone.
- Refusal-path tests assert the compiled model is unchanged, not just the error status, and one test drives the agent-facing router - the reachable route.
- Controls pin both accepted values (python and numpy booleans, both branches) and the numeric siblings' inverse domain, so an over-reaching guard fails the suite.
- `Args:` entries updated for both parameters (convention 13), changelog fragment present and named per convention, validation ahead of the lock and of any mutation.

## Verification suggestions

The failing `Detect an untested overlap with the base branch` check is the advisory merge-base detector: `manipulation.py` also changed on `main` via #2420, but that change is a docstring-only edit to `_apply_kinematic_attachments` (lines 373-384), fully disjoint from `actuate_robot`. Merging `main` into the branch and pushing clears it; per the #1821 measurement in AGENTS.md a clean base merge does not cost the approval.

yinsong1986 (approved):
## Summary

Moves `actuate_robot`'s two posture flags (`gravity_compensation`, `disable_self_collision`) onto the shared `strands_robots.utils.boolean_flag_error` domain, removing the `bool()` laundering at the `actuate_robot_in_scene` call site that let truthy opt-out spellings (`"false"`, `"no"`, `"off"`, `"0"`, `nan`) select the branch being opted out of - for `disable_self_collision`, zeroing `contype`/`conaffinity` on every robot geom on the spec, where it outlives recompiles. The guard runs before `_ensure_mujoco()` and before anything is written to the spec, so a refused call leaves the model untouched (`nu == 0`, collision and `body_gravcomp` unchanged, asserted by the new tests), and it returns a structured error dict per the backend's error-handling contract rather than raising. Downstream `actuate_robot_in_scene` reads the flags by truthiness, which is correct for both `python` and numpy booleans the shared domain admits. The new test file covers refusal, state-preservation, the agent-facing router route, message-content keying on the shared helper, and controls pinning that real booleans still select both postures and that the four numeric siblings keep their inverse domain. Changelog fragment included; scope is disciplined to the one signature. No blocking concerns.

## What's good

- Reuses the existing shared `boolean_flag_error` guard (AGENTS.md rule 11: it already has multiple callers) rather than minting a local variant.
- Refusal path verified to be side-effect-free: tests assert the compiled model is byte-identical after a rejected call.
- Controls explicitly pin `np.bool_` acceptance and the numeric siblings' rejection of `bool`, so the two inverse domains cannot silently converge.

PR #2423, opened 2026-08-18
fix(simulation/mujoco): jitter a light around its authored position, not its last one

cagataycali:
## Problem

`randomize()` documents its lighting axis as **"light pos jittered ±0.5m"** — a bound on how far a light may sit from where the scene authored it. The sampler was:

```python
model.light_pos[i] += rng.uniform(-0.5, 0.5, size=3)
```

Every call started from wherever the previous call left the light, so the displacement is a **random walk**, not a bounded jitter. Each individual draw is inside the advertised half-width, so nothing ever reports the breach, while the total displacement grows without limit. `data.light_xpos` — the array the renderer actually reads — follows.

Measured on `so101` (lights authored 3.5 m above the scene), one `randomize(randomize_lighting=True)` call per episode:

| calls | max distance from the authored position | vs the documented 0.5 m bound |
|---|---|---|
| 1 | 0.25 m | inside |
| 3 | 0.53 m | **breached** |
| 10 | 2.58 m | 5.2x |
| 50 | **4.72 m** | **9.4x** |

Over 24 episodes, **23 of 24** ended with a light outside the bound. A key light metres away from the scene it is meant to illuminate is not domain randomization; it is a broken scene, and the per-episode loop that produces it is the documented sim2real workflow.

The same defect also made the axis unreproducible: `seed` is documented as giving "a reproducible stream", but because the result was an increment onto the current pose, replaying a seed depended on how many calls had already run.

## Why lighting specifically

`randomize()`'s other axes all measure from a fixed reference, and this function has already litigated exactly this hazard. Its own `position_noise` documentation says the offset is measured from each object's commanded pose

> so repeated calls draw independent offsets inside this bound instead of compounding into a random walk

`light_diffuse` is assigned, and colour is assigned, so both were bounded. `light_pos` was the one axis with no fixed reference. The Newton backend's lighting axis composes `base + jitter` around a constant base direction and is bounded by construction.

## Change

Offset each light from its **authored** position — the pose the scene spec declares. That is the same source a recompile regenerates `model.light_pos` from, so it stays correct for the life of the scene however many times the axis has run, and it needs no new state to cache. Verified against the compiled model on **62 of 62** loadable sim robots (`MjSpec.lights` is ordered as the `model.light_pos` rows are, including lights a robot spec contributes through `spec.attach`), and after `create_world`, `load_scene` and every scene op.

The reference is resolved during validation, **before the first mutation**, so a scene whose authored poses cannot be read is refused with nothing applied — resolving it inside the lock would leave the colour axis (which runs first) already written by the time lighting gave up. Falling back to the live position would silently reinstate the accumulating behaviour, so that case is refused with a message naming the axis and a way forward.

The `±0.5` half-width is now a named constant so the bound the docstring advertises and the bound the sampler draws cannot drift apart.

| axis | operator | bounded before? |
|---|---|---|
| `randomize_colors` | assign | yes |
| `randomize_lighting` (diffuse) | assign | yes |
| `randomize_lighting` (position) | **`+=`** | **no — this change** |
| `randomize_positions` | from the commanded pose | yes |

`randomize_physics` is deliberately untouched: `mass_range` / `friction_range` are documented as *multiplicative scales*, and `*=` is the literal reading of that. Whether a scale is relative to the authored value or to the current one is a semantics question rather than a bound that is not being honoured, so it is out of scope here.

## Verification

`3 failed / 7 passed` before, `10 passed` after. The three pre-fix failures name the numbers:

```
after 50 randomize(randomize_lighting=True) calls a light sits 4.7151 m from its
authored position, 9.4x the documented +/-0.5 m bound
```
```
the same seed produced a different light pose because seven unrelated calls ran
in between; max difference 0.4420 m
```
```
expected a refusal, got {'status': 'success', ...}
```

The seven tests that pass on both trees are the boundary: the axis still moves every light (the bound is not honoured by freezing it), a single call was always inside the bound, `light_diffuse` and the colour axis stay inside their ranges, an axis-off call writes nothing, a recompile still undoes the axis as it does the other three, and the success text is unchanged.

Gate: `ruff check` / `ruff format --check` clean on `strands_robots tests tests_integ`; `mypy` 19 pre-existing errors on both trees, none in the changed files. A 332-file selection (all of `tests/simulation`, every test naming `randomize` / `light_pos` / `set_obs_noise` / `_get_spec` / the docs page, plus the repo-wide guards) run on this branch and on `main`: **10776 passed / 1 failed** vs **10773 passed / 4 failed**, both collecting 10838. Branch-only failures: **none**. Base-only: exactly the three pre-fix failures above. The single shared failure (`test_recording_frame_loss_is_not_tolerated.py::test_every_on_frame_call_site_excludes_a_lost_recording_frame`) fails identically on both trees.

Measured at `2e466d81`; `main` advanced by two commits during the run, so the branch was merged forward and the affected suites re-run on the merged tree (145 passed).

## Artifact

One capture script, run against this branch and against `main` — same seeds, same scene, same camera; only `strands_robots` differs. 24 per-episode `randomize(randomize_lighting=True)` calls on `so101` in MuJoCo (headless EGL).

randomize_lighting stays inside its documented bound

The trajectory is the finding: before, the light wanders to 2.19 m with 23 of 24 episodes outside the bound; after, all 24 stay inside (max 0.4985 m). Episode 0 is the control — both trees start from the authored pose and produce an identical light position, so the panels diverge only as the accumulation takes hold. The render strip is secondary evidence: this scene's second light and headlight fill limit how much a displaced key light changes the frame (largest divergence 3.44/255 mean), which is why the quantitative panels lead.

## Docs

`docs/simulation/domain-randomization.md` gains the lighting counterpart of the paragraph that already explains the position axis's fixed reference, and its category table now states the bound and what the offset is measured from.


yinsong1986 (approved):
## Summary

Converts the MuJoCo `randomize_lighting` position sampler from an increment on the live `model.light_pos` (an unbounded random walk across repeated calls) to an absolute offset from each light's authored position read out of the tracked `MjSpec` - the same source a recompile regenerates the model from, so no new cached state is introduced. The authored reference is resolved before the first mutation, and a world whose spec is missing or disagrees with the compiled light count is refused with an error dict and nothing applied, rather than silently falling back to the accumulating behaviour. The half-width becomes a named constant, and the docs gain the lighting counterpart of the existing `position_noise` fixed-reference explanation.

## What's good

- Refusal path follows the #85 contract (error dict, no partial mutation - the colour axis is untouched on a refused call, and the new test pins that ordering).
- The spec read happens outside the lock in the same pattern the function already used for `model`/`data`, guarded by `_require_no_running_policy`; no new race class is introduced, and no circular import (`scene_ops` does not import `randomization`).
- Test suite covers both the fix (bound held over 50 calls, seed reproducibility) and its boundaries (axis still moves lights, diffuse/colour untouched, axis-off writes nothing, recompile still restores, refusal applies nothing) - verified passing on this head, and lint/format are clean on both changed Python files.
- Changelog fragment present per AGENTS.md step 3; `randomize_physics`'s multiplicative semantics correctly left out of scope.

PR #2424, opened 2026-08-18
fix(policies/wbc): complete an incomplete obs_scales map from the upstream defaults

cagataycali:
## Problem

`WBCConfig` documents three observation scales with upstream `g1_gear_wbc.yaml` defaults - `ang_vel` 0.5, `dof_pos` 1.0, `dof_vel` 0.05 - and `build_single_frame` / `build_gait_frame` read them on every tick.

A config that stated **no** scale got those defaults from the dataclass field. A config that stated **some** of them replaced the whole `obs_scales` map, leaving the rest to a second fallback number inside the builders: a bare `1.0`. So the effective scale of an unnamed key depended on which of its siblings the config happened to name.

The upstream YAML spells the scales as flat `ang_vel_scale` / `dof_pos_scale` / `dof_vel_scale` keys, which makes a partial config the ordinary case - a config that only wants to retune one scale states only that one.

| config states | `obs_scales` before | effective `dof_vel` |
|---|---|---|
| nothing about the scales | `{ang_vel: 0.5, dof_pos: 1.0, dof_vel: 0.05}` | 0.05 |
| `ang_vel_scale`, `dof_pos_scale` | `{ang_vel: 0.5, dof_pos: 1.0}` | **1.0 (20x)** |
| `dof_pos_scale` only | `{dof_pos: 1.0}` | **1.0 (20x)**, and `ang_vel` 1.0 (2x) |
| `obs_scales={"ang_vel": 0.5}` | `{ang_vel: 0.5}` | **1.0 (20x)** |
| all three flat keys | full map | 0.05 |

`dof_vel` scales the `dqj` block - 29 of the 86 entries of every frame - so a config that omitted it handed the network a frame 20x off on a third of its width, while the dataclass default, `docs/policies/wbc.md` ("joint velocities `[28:43]` (scaled by `dof_vel_scale=0.05`)") and the layout comment all said 0.05. Nothing raised and nothing warned: `config.obs_scales["dof_vel"]` raised `KeyError` rather than reporting the value the frame was built with.

Driving the GR00T SONIC checkpoint (`policy.onnx` + `walk_policy.onnx`) on the G1 with `target_velocity=[0.5, 0, 0]` for 400 ticks at 50 Hz:

| | effective `dof_vel` | pelvis min | travelled | status |
|---|---|---|---|---|
| config states two scales, omits `dof_vel_scale` | 1.0 | **0.193 m** (from 0.740) | 0.47 m | `success` |
| the same two scales, `dof_vel_scale: 0.05` stated | 0.05 | 0.740 m | 3.11 m | `success` |

## Change

`obs_scales` is completed once, on the config, from a single owner (`_DEFAULT_OBS_SCALES`) - for the reason `WBCPolicy` already fills the per-joint SONIC vectors on the config rather than in the policy: *"so the observation builder - which reads config.default_angles for the qj offset - and the controller see the same values"*. A stated scale always wins; only an omitted one is filled, so an explicit map still overrides the flat key it overlaps.

`config.obs_scales` is therefore always the complete map the frame is built with, whichever spelling the config used, and a reader taking `config.obs_scales["dof_vel"]` gets the value rather than a `KeyError`.

The builders keep a fallback but resolve it from the same table instead of a bare `1.0`, so no second number exists anywhere to drift from the first.

Out of scope: the map stays **open** - a key the layout does not read is kept, not refused, since `obs_scales` is documented as named scale factors and the gait variant may add to them. Pinned by a test.

## Rollout

Same checkpoint, same command, same config; only `strands_robots` differs. Left is `upstream/main`, right is this branch. The two panels are pixel-identical at t=0 (mean abs delta 0.0000) and diverge to 16.51 as the frame the network reads starts to matter.

obs_scales rollout

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-wbc-obs-scales/wbc-obs-scales.mp4) - [final frame](https://raw.githubusercontent.com/cagataycali/robots/media-wbc-obs-scales/wbc-obs-scales-final.png)

The free camera tracks forward progress at a fixed height, so a fall reads as the robot dropping in frame. In the final frame nothing on the left rises above the horizon (topmost structural row `-1`, the robot is entirely on the floor); on the right the head sits at row 105.

## Tests

`tests/policies/wbc/test_obs_scales_completed_from_upstream_defaults.py`, **10 failed / 10 passed** on `upstream/main`, 20 passed here. The expected values are read from a config that states no scales at all - the route whose defaults were never in question - rather than from a literal table, so the tests grade the agreement between the two routes.

Pre-fix, the headline reports:

```
a config that states ang_vel_scale and dof_pos_scale but leaves dof_vel_scale to
its documented default 0.05 scaled the joint velocity block by 20x what the same
scales state explicitly: [2. 2. 2.] vs [0.1 0.1 0.1]
```

and the root-cause test - parametrized over all 8 subsets of "which scales did this config state" - fails on the 6 partial ones, e.g.

```
stating ['ang_vel', 'dof_pos'] changed the effective scale of the unnamed
['dof_vel']: {'dof_vel': 1.0} instead of the documented [0.05]
```

The 10 that pass on both trees are the boundaries the fill must not cross: a stated scale is never replaced (flat and nested routes), an explicit map still wins over the flat key it overlaps, a config stating no scales still gets the upstream defaults, a scale the domain refuses is **still refused** (so completion cannot smuggle an unusable stated value past validation), and an unknown key survives.

## Verification

Measured at `404d924c`. Blast radius - the 48 test files referencing `obs_scales` / `WBCConfig` / the frame builders / `wbc`, plus `tests/policies/wbc/` and the repo-wide docstring, unicode, ASCII-log, args-docstring, dependency-audit and changelog guards:

- this branch: **1799 passed, 0 failed**, 1 skipped
- `upstream/main` with the new test file copied in: **1789 passed, 10 failed**, 1 skipped

`1789 + 10 == 1799` and both collected 1800, the 10 base failures are exactly the new file's pre-fix failures, and there are no branch-only failures.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` reports the same 19 pre-existing errors on both trees and none in the touched files.

### DIFF
--- a/changelog.d/2424-wbc-obs-scales-completed.md
+++ b/changelog.d/2424-wbc-obs-scales-completed.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **policies/wbc**: a `WBCConfig` that states some observation scales now keeps the documented upstream default for the rest. Stating `ang_vel_scale` and `dof_pos_scale` used to replace the whole `obs_scales` map, leaving `dof_vel` to a second fallback number inside the frame builders (a bare `1.0`) instead of its documented `0.05` - scaling the 29 joint-velocity entries of every observation frame by 20x. Driving the GR00T SONIC checkpoint with such a config walked the G1 0.37 m and dropped its pelvis from 0.74 m to 0.19 m while reporting success; it now walks 3.11 m as the same scales stated explicitly do. `obs_scales` is completed once on the config, so the observation builder, the gait builder and any other consumer read the same values.
--- a/docs/policies/wbc.md
+++ b/docs/policies/wbc.md
@@ -162,6 +162,14 @@ WBC reproduces the upstream `GearWbcController` loop (NVlabs/GR00T-WholeBodyCont
   - joint positions `[13:28]` (minus `default_angles`, scaled by `dof_pos_scale`)
   - joint velocities `[28:43]` (scaled by `dof_vel_scale=0.05`)
   - previous action `[43:58]`; indices `[58:86]` are a reserved (zero) tail.
+
+  The upstream YAML spells these three scales as flat `ang_vel_scale` /
+  `dof_pos_scale` / `dof_vel_scale` keys, and `WBCConfig` normalises them into
+  the nested `obs_scales` map. A config only has to state the scales it wants to
+  change: the rest keep the upstream defaults above, so naming one scale never
+  changes what an unnamed sibling is scaled by. `config.obs_scales` is therefore
+  always the complete map the frame is built with, whichever spelling the config
+  used.
 - **Action** - the network emits a 15-dim joint-position *offset*; the policy
   forms absolute targets `target_q = default_angles + action_scale * raw` and
   returns them keyed by actuator name. For torque-actuated MuJoCo, convert with
--- a/strands_robots/policies/wbc/config.py
+++ b/strands_robots/policies/wbc/config.py
@@ -84,6 +84,14 @@ _DEFAULT_NUM_ACTIONS = 15
 # Upstream g1_gear_wbc.yaml: obs_history_len=6 (num_obs = 86*6 = 516).
 _DEFAULT_OBS_HISTORY_LEN = 6
 _DEFAULT_COMMAND_DIM = 7
+# The observation scales from upstream g1_gear_wbc.yaml, and the single owner of
+# them. Every consumer resolves an omitted key from THIS table, so a config that
+# states some scales and leaves the rest to their documented default cannot end
+# up scaled differently from one that states none: ``dof_vel`` is 0.05 either
+# way. A second fallback number (a bare ``1.0``) would silently multiply the 29
+# joint-velocity entries of the frame by 20 for exactly the configs that name a
+# sibling key, which the network reads as a malformed observation.
+_DEFAULT_OBS_SCALES: dict[str, float] = {"ang_vel": 0.5, "dof_pos": 1.0, "dof_vel": 0.05}
 # Upstream cmd_scale applied to [vx, vy, omega] and the default base-height
 # command, from g1_gear_wbc.yaml.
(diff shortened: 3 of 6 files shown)


yinsong1986 (approved):
## Summary

Fixes a silent 20x mis-scaling of the 29-entry joint-velocity block for any `WBCConfig` that stated *some* observation scales: a partial config replaced the whole `obs_scales` map, and the omitted keys fell through to a second fallback (`1.0`) inside `build_single_frame` / `build_gait_frame` instead of the documented upstream defaults. The fix hoists the defaults into a single owner (`_DEFAULT_OBS_SCALES` in `config.py`), completes an incomplete map once in `__post_init__` (stated scales always win, validation still runs before the fill so a refused stated value stays refused), and repoints the builders' residual fallbacks at the same table so no second number exists to drift. Verified locally at the head SHA: the new 20-test file passes, the full `tests/policies/wbc/` suite passes (536/536), all `obs_scales` consumers in the tree now resolve from the one table, the caller's dict is not mutated by completion, and `from_dict` flat-key normalisation composes correctly with the fill. The behaviour change only affects configs that were getting the undocumented accidental `1.0` - the documented, dataclass-default and doc-comment value was always 0.05 - and it is recorded as a changelog fragment per AGENTS.md step 3.

## What's good

- Single-owner constant with both builders' fallbacks resolving from it - no second number left to drift, matching the stated design rationale.
- Validation-before-completion ordering means completion cannot smuggle an invalid stated scale past `finite_number_error`, and the test file pins that boundary explicitly.
- Tests grade agreement between the no-scales-stated route and the partial routes rather than restating literals, and cover both the flat-key (`from_dict`) and nested-map (constructor) spellings plus the open-map and precedence boundaries.
- Changelog fragment, docs update, and docstring updates all landed in the same PR; no `## [Unreleased]` edit.

Issue #2414, opened 2026-08-17
Three more frame pacers decide on time.time(): the MJPEG stream, the multi-camera recorder, and Isaac's idle-render gate


cagataycali:
#2409 settled `simulation/policy_runner.py` and `policies/lerobot_local/policy.py`, which completed the class as that issue scoped it. Reading the rest of the tree for the same shape - `frame_interval - (time.time() - t0)` feeding a `sleep` - turns up three more pacers that **decide** rather than report, all in rendering paths. Filed so the remainder is tracked rather than left in a pull request description.

The rule is AGENTS.md, "Clocks: a duration is measured, a stamp is recorded", which names a frame pacer explicitly.

## 1. MJPEG stream pacer - `strands_robots/rendering/video.py:375,394`

```python
375:  t0 = time.time()
      ...
394:  elapsed = time.time() - t0
      if elapsed < frame_dt:
          time.sleep(frame_dt - elapsed)
```

This paces the frames served to a connected HTTP client. Forward step, the pacing is skipped and the generator spins as fast as `frame_fn()` returns for the rest of the connection; backward step of S, one frame is held for `frame_dt + S` and the viewer stalls. The stream is long-lived, which is what makes it likely to be running when a correction lands.

## 2. Multi-camera recorder pacer - `strands_robots/simulation/mujoco/rendering.py:2131,2147`

```python
2131:  t0 = _time.time()
       # render every camera
2147:  lag = _time.time() - t0
       if lag < interval:
           _time.sleep(interval - lag)
```

Same shape, but the consequence is a dataset rather than a viewer: this thread fills `state["buffers"]` at a nominal `fps`, and a forward step makes it capture unthrottled into `max_frames` while a backward step drops the effective rate. The buffered frames carry no per-frame timestamp, so a recording paced by a stepped clock is indistinguishable afterwards from one paced correctly - the frames are simply not `1/fps` apart.

## 3. Isaac idle-render gate - `strands_robots/simulation/isaac/simulation.py:5769`

```python
5769:  now = time.time()
       do_render = (now - last_idle_render) >= self._idle_render_period
       self.pump(render=do_render)
       if do_render:
           last_idle_render = now
```

An interval comparison, so a backward step suppresses the idle render for the size of the step (the viewport freezes while the app is pumped) and a forward step renders every tick until the base catches up. `last_idle_render` is a duration base named as a point in time - the same invitation `#2408` renamed away in the mesh.

## Reporting-only, same class, lower consequence

[...]

PR #2425, opened 2026-08-18
fix(simulation/mujoco): pace the rendering loops on a clock that cannot step

cagataycali:
Two rendering loops pace themselves the same way: read a base, do the frame's work, sleep for whatever is left of the interval. Both bases were `time.time()` — the current opinion about the date, not a clock — and the sleep was computed from two readings of it, so a wall-clock step landing between them (an NTP correction, a `date -s`, a resume from suspend) changed the rate. Nothing raised; both calls return `status="success"`.

This is the rule AGENTS.md states as *"a duration is measured, a stamp is recorded"*, which names a frame pacer explicitly, and the boundary already settled for the agent-callable tools (#2404), the hardware control loops (#2406), the mesh (#2408) and the sim rollouts (#2413). #2414 enumerates the three pacers that class still had left after #2409; this change is the two rendering ones that issue groups as one coherent change, so it is *towards* #2414 rather than a close — the Isaac idle-render gate stays with the issue.

## The recorder: a dataset that cannot show what happened

`start_cameras_recording`'s thread fills the buffers `stop_cameras_recording` flushes to one MP4 per camera. Measured on a **real two-camera 320x240 MuJoCo recording of a sweeping arm**, with a −2 s step landing in the render window:

| | frames in the MP4 | duration the MP4 declares | duration the call reports | motion actually recorded |
|---|---|---|---|---|
| before | **11** | **1.1 s** | 1.9 s | 3.0 s |
| after | 31 | 3.1 s | 4.0 s | 3.0 s |

One gap of **2.1 s** (`interval + |step|`) where nothing was captured, against 0.100–0.102 s for every other gap. Because the buffers carry no per-frame timestamp, a recording paced by a stepped clock is **indistinguishable afterwards** from one paced correctly — the frames are simply not `1/fps` apart, and the MP4's own timebase claims 1.1 s for 3.0 s of motion. A forward step skips the pacing for that cycle instead.

before and after

The two recordings replayed on a shared real-time axis — the left panel stops advancing at 9 frames from t=0.9 s to t=2.8 s:

recording

Raw clips: [before](https://raw.githubusercontent.com/cagataycali/robots/media-rendering-pacer-clock/before__side.mp4) · [after](https://raw.githubusercontent.com/cagataycali/robots/media-rendering-pacer-clock/after__side.mp4) · [side-by-side MP4](https://raw.githubusercontent.com/cagataycali/robots/media-rendering-pacer-clock/recording.mp4)

### How often a step lands where it matters

The vulnerable window is between the base and the elapsed reading — i.e. the render itself — so its width as a fraction of the cycle is the probability a step changes the capture rate rather than only the reported elapsed:

| recorder configuration | render/cycle | interval | vulnerable |
|---|---|---|---|
| 2 cams @ 160x120, 10 fps | 6.4 ms | 100.0 ms | 6% |
| 2 cams @ 320x240, 30 fps | 17.3 ms | 33.3 ms | 52% |
| 4 cams @ 640x480, 30 fps | 97.8 ms | 33.3 ms | 100% |

`start_cameras_recording`'s own docstring sizes its memory cost for "a 2s / 4-cam / 320x240 / 15fps rollout", which is squarely in the range where the loop is render-bound and any step lands in the window.

## The MJPEG stream

`mjpeg_frames` paces the frames served to a connected `` client. At 12 fps (83.3 ms nominal), with the step inside one frame's window:

| clock event | achieved interval for that frame |
|---|---|
| no step (control) | 83.3 ms |
| wall clock steps +30 s | 4.0 ms (pacing skipped) |
| wall clock steps −2 s | 2083.3 ms (one stall) |
| wall clock steps +1 h | 4.0 ms (saturates) |

The stream is long-lived, which is what makes it likely to be running when a correction lands.

## The reported duration

Both recorders read one base at start and subtract it at stop, so this is corrupted by a step landing **anywhere** in the recording, not only in the render window. The same 3.01 s capture reported `Stopped 'before' after 1.9s`; a +1 h step reports `after 3604.0s` for a 4.0 s recording, and `get_cameras_recording_status` reads the same base. That base is now `started_mono`, matching the `_mono` convention the mesh (`_last_estop_mono`) and the hardware loops (`start_mono`) already use for a duration base.

## Tests

`tests/simulation/test_rendering_pacers_survive_a_clock_step.py` drives both real loops through a clock double whose *wall* clock takes a known step while its monotonic clock does not. Two properties make it robust:

- The double counts reads of **either** clock, so the step lands at the same position in the loop whichever clock the loop consults — one stimulus reaches the pre-fix and post-fix code identically.
- Each assertion is on the **achieved** interval — the pacing the viewer or the dataset actually got — rather than on the value of the computed sleep, so it cannot be satisfied by renaming a call.

**13 failed / 3 passed** on `main`, 16 pass here. The three that pass on both are the controls that bound the change: the nominal-rate reference for each loop, and a step taken *outside* the base..elapsed window, which was always harmless — the fix is scoped to the window that was actually vulnerable.

No GL context, dataset or robot is touched: `render` is faked on the instance, as in `test_daemon_camera_recording.py`. The second commit confines the double — the recording path lazily first-imports imageio's ffmpeg plugin, `subprocess` and `logging`, which would bind the double for the rest of the session, so those imports are warmed on the real clock first and a leak guard names any module that retains it.

## Verification

- `ruff check` / `ruff format --check` / `mypy` over `strands_robots tests tests_integ`: clean; mypy is 14 pre-existing errors on both trees with an empty branch-only diff.
- Blast radius (41 files — every test naming these surfaces, plus the repo-wide guards), same selection on both trees: branch **1850 passed / 0 failed / 29 skipped**, `main` with this test file copied in **1837 passed / 13 failed / 29 skipped**. `1837 + 13 = 1850` against 1879 collected on both, and the 13 are exactly this file's pre-fix failures — no branch-only failure anywhere.

## Scope

The Isaac idle-render gate (`simulation/isaac/simulation.py`) is the same shape in a different loop and is left for its own change, tracked on #2414; #2328 is currently editing that file. `tools/lerobot_camera.py` reads the wall clock for progress lines only — its recording loop is bounded by `target_frames`, not by elapsed time — so a step makes a report wrong rather than a capture wrongly paced.

## Review rounds

**Round 1** — `91b2e0a`. Code scanning flagged the new test file's module-level `_MEDIA_WARMED` flag. The flag did latch, but the alert named real redundancy underneath it: a hand-rolled memo around `_warm_media_stack`, plus a `tmp_path` threaded into `_install_time_double` from three call sites that only the **first** call ever used — the warmup wrote its throwaway MP4 into whichever test happened to run first, so one test's `tmp_path` held a file belonging to no test. `functools.cache` makes once-only the function's own property and a private `TemporaryDirectory` gives the warmup its own scratch, so the flag and the parameter both go. Test-only, and no stimulus changes: same 16 tests, same clock double, same 13/3 pre-fix split on `main`, and the confinement guard `_assert_double_did_not_leak` still names any module left holding the double.


yinsong1986 (dismissed):
## Summary

Moves the two rendering pacers - the MJPEG stream generator (`strands_robots/rendering/video.py:_mjpeg_stream`) and the multi-camera recorder thread (`strands_robots/simulation/mujoco/rendering.py`) - from `time.time()` to `time.monotonic()` for their frame-interval bases, and renames the recorders' duration base `started_at` -> `started_mono` so the key carries its clock, matching the `_mono` convention already used by the mesh and the hardware loops. This is the AGENTS.md "a duration is measured, a stamp is recorded" rule applied to the last two loops it names, and the diff is exactly scoped to it: the state dict is a private in-memory attribute (`_cams_rec_state`), never persisted or serialised, so the key rename is not a schema change, and every reader of the old key on the mujoco side is updated in this PR (the Isaac copy is explicitly left for its own change while #2328 edits that file).

## What's good

- `tests/simulation/test_rendering_pacers_survive_a_clock_step.py` pins the achieved interval rather than the computed sleep, so the pin cannot be satisfied by renaming a call; the clock double counts reads of either clock so one stimulus reaches pre- and post-fix code identically, and controls bound the fix to the window that was actually vulnerable.
- The `sys.modules["time"]` double is confined deliberately: the lazily-imported media stack (imageio ffmpeg plugin, subprocess, logging) is warmed on the real clock first, and a leak guard names any module left holding the double - closing the failure mode where an unrelated later test fails on unreadable MP4 metadata.
- Changelog fragment present per step 3; `test_both_recorders_name_the_clock_their_duration_base_holds` keeps the daemon and synchronous recorders on one spelling of the base.
- No non-ASCII, no host paths in the new test file; `monkeypatch` used throughout.

yinsong1986 (approved):
## Summary

Moves the two rendering pacers - the MJPEG stream generator (`strands_robots/rendering/video.py:_mjpeg_stream`) and the multi-camera recorder thread in `strands_robots/simulation/mujoco/rendering.py` - from `time.time()` to `time.monotonic()` for their frame-interval bases, and moves the recorders' reported-duration base onto the same clock, renaming the state key `started_at` -> `started_mono` so the key names the clock it holds. This is the AGENTS.md clocks rule ("a duration is measured, a stamp is recorded") applied to the last two pacers #2414 groups together, scoped deliberately away from the Isaac idle-render gate that #2328 is editing. The rename is confined to the private in-memory `_cams_rec_state` dict shared by the daemon and synchronous recorders - it is not a persisted schema, wire format, or public API, and the one test helper that builds the dict by hand is updated in the same change. The new test file pins both loops on the *achieved* interval through a clock double whose read counter is clock-agnostic, so the assertions cannot be satisfied by renaming a call, and the round-1 CodeQL finding (`_MEDIA_WARMED`) is resolved at this head via `functools.cache` plus a leak guard that names any module left holding the double.

## What's good

- Controls bound the fix on both sides: a nominal-rate reference per loop, and a step landing *outside* the vulnerable window that passes pre- and post-fix.
- The duration bases carry their clock in their name (`started_mono`, `frame_start_mono`), matching the `_mono` convention the mesh and hardware loops already use, and a dedicated test rejects `started_at` reappearing in either recorder's state.
- Changelog fragment in `changelog.d/` per step 3; no direct `CHANGELOG.md` edit. No host paths or non-ASCII in the new test file.
- The `_warm_media_stack` confinement (first-importing imageio's ffmpeg plugin on the real clock) plus `_assert_double_did_not_leak` keeps the `sys.modules` double from outliving its test - a failure there names the module rather than surfacing as unreadable MP4 metadata in an unrelated test.

PR #2426, opened 2026-08-18
fix(training/rl): honor the documented eval-mode freeze in EmpiricalNormalization.update

cagataycali:
## Problem

`EmpiricalNormalization` documents that its running statistics are frozen in eval
mode, "so an exported policy normalizes deterministically". That freeze has two
entry points and only one of them honored it:

| route | eval-mode fold | result |
|---|---|---|
| `forward(x, update=True)` | gated on `self.training` | frozen |
| `update(x)` (public) | ungated | **folds the batch** |

Measured on a normalizer trained to `count=2`, `mean=[2, 3, 4]`, then flipped to
`eval()` and handed an 8-row batch of `100.0`:

```
forward()  count 2 -> 2    mean [2.0, 3.0, 4.0] -> [2.0, 3.0, 4.0]     frozen
update()   count 2 -> 10   mean [2.0, 3.0, 4.0] -> [80.4, 80.6, 80.8]  NOT frozen
```

Nothing is raised and nothing is logged.

## Why it is not confined to the current step

`_mean`, `_var`, `_std` and `count` are registered buffers, so they are part of
`state_dict()`. `PpoTrainer` / `FastSacTrainer` persist the normalizer with
`state_dict()` and `BaseRLAlgo.load_checkpoint` reads it back, so a batch folded
in after `eval()` becomes the statistics the next checkpoint carries and the
whitening an exported policy applies to every observation. The same observation
whitened to `-0.990` before the fold and `-2.025` after it.

`BaseRLAlgo.evaluate` is the in-repo code that flips the normalizer to `eval()`
around a scoring rollout and restores the prior mode afterwards. Its docstring
promises it "never updates the policy, the normalizers, or the replay buffer ...
with ... observation normalization frozen, so the numbers it returns are exactly
what a deployed `policy.pt` would produce", and it currently gets that from a
second, independent mechanism (`update=False`) rather than from the mode.

## Fix

`update` returns early when the module is not training - the first thing the
rsl_rl implementation this was ported from does in its own `update`:

```python
if not self.training:
    return
if self.until is not None and int(self.count) >= self.until:
    return
```

`forward` already gates on the mode, so it never reaches the new branch and
every existing call site is byte-for-byte unaffected. Training-mode folds, the
`until` warmup freeze and the empty-batch no-op are unchanged. The class and
`update` docstrings now state which calls the freeze covers, and the
`docs/training/rl.md` component row says so too.

## Measurement on a real observation stream

One script, run once per tree, driving the `evaluate`-shaped sequence over a
so101 MuJoCo rollout (D=12: 6 joint positions + 6 velocities): train on 60
observations, snapshot `state_dict()`, `eval()`, fold 40 scoring-rollout
observations, restore the mode, read `state_dict()` again.

| | sample count | buffers that drifted | action delta of a fixed deterministic actor |
|---|---|---|---|
| before | 60 -> 100 | `_mean`, `_std`, `_var`, `count` | **0.1968** max-abs |
| after | 60 -> 60 | none | **0.0000** |

The action column feeds the same fixed actor the same observation, whitened by
each tree's next-checkpoint statistics; 0.1968 is ~10% of full scale on a
tanh-bounded action space.

EmpiricalNormalization eval-mode freeze

## Tests

Extended `tests/training/test_rl_normalization.py`, which already targets this
class's contract edges. Against `upstream/main`: **3 failed / 5 passed**; after
the fix **8 passed**.

Failing pre-fix:

- `test_eval_mode_freezes_the_statistics_through_either_entry_point[update]` - *"update() in eval mode advanced the sample count 2 -> 10"*
- `test_a_batch_folded_in_after_eval_does_not_reach_the_checkpoint` - *"buffers a checkpoint would carry drifted after eval(): ['_mean', '_std', '_var', 'count']"*
- `test_restoring_training_mode_resumes_updating` - `assert 6 == 2`

Passing on both trees, as the bounds on the change:

- the same test parametrized over `[forward]` - the route that was already correct, so a fix that reached further would fail here
- `test_update_still_folds_the_batch_in_training_mode` - fails if `update` is disabled outright rather than mode-scoped
- the three pre-existing edges (`center=False`, `until`, empty batch)

## Verification

- `mypy strands_robots tests tests_integ`: 19 pre-existing errors, identical set on this branch and on `upstream/main`, none in the changed files.
- 63 test files covering `training/rl`, every test referencing the normalizer, and the repo-wide docstring / string / changelog guards - run on this branch and on `upstream/main` (3146 collected, 5:01 each): **branch 3132 passed / 8 failed, base 3129 passed / 11 failed**. Branch-only failures: none. Base-only: exactly the 3 pre-fix failures above. The 8 shared failures are pre-existing on both trees (5 long-running `tests_integ` RL convergence tests, plus 3 that need `draccus>=0.11.6`).

### DIFF
--- a/changelog.d/2426-rl-normalizer-eval-mode-freeze.md
+++ b/changelog.d/2426-rl-normalizer-eval-mode-freeze.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **training/rl**: `EmpiricalNormalization.update()` is now a no-op in eval mode, so the documented eval-mode freeze holds for a direct `update()` call and not only for `forward(update=True)`. The running statistics are persistent buffers, so a batch folded in after `eval()` was written into the next checkpoint and changed how an exported policy whitened every observation, with nothing raised and nothing logged. Training-mode behavior, the `until` warmup freeze and the empty-batch no-op are unchanged.
--- a/docs/training/rl.md
+++ b/docs/training/rl.md
@@ -30,7 +30,7 @@ trainer = create_trainer("ppo")   # on-policy PPO, from-scratch RL
 | [`FastSacTrainer`](#fastsac) | Soft Actor-Critic (off-policy, replay buffer, twin Q critics, auto-tuned entropy). |
 | `SimpleReplayBuffer` | Off-policy transition store (fixed-capacity ring buffer). |
 | [`SimEnv`](#simenv) | Gym-style `reset -> step` adapter over a `SimEngine`. |
-| `EmpiricalNormalization` | Running observation normalizer (whitens inputs for stable training). |
+| `EmpiricalNormalization` | Running observation normalizer (whitens inputs for stable training). Statistics update only in training mode - `eval()` freezes them at both entry points (`forward` and `update`), so an exported policy whitens deterministically. |
 
 ## SimEnv
 
--- a/strands_robots/training/rl/normalization.py
+++ b/strands_robots/training/rl/normalization.py
@@ -22,9 +22,11 @@ from torch import nn
 class EmpiricalNormalization(nn.Module):
     """Normalize a tensor stream by its running (Welford) mean and std.
 
-    The running statistics update only while the module is in training mode and
-    ``update=True`` is passed to :meth:`forward`; in eval mode the learned
-    statistics are frozen, so an exported policy normalizes deterministically.
+    The running statistics update only while the module is in training mode:
+    :meth:`forward` additionally requires ``update=True``, and :meth:`update`
+    folds a batch whenever the module is training. In eval mode the learned
+    statistics are frozen at BOTH entry points, so an exported policy
+    normalizes deterministically.
 
     Args:
         shape: Per-feature shape of the observation (e.g. ``(num_obs,)``).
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes a real contract gap in `EmpiricalNormalization`: the class docstring promises the running statistics are frozen in eval mode, but only `forward(update=True)` honored it — a direct call to the public `update()` folded the batch regardless of mode. Because `_mean`/`_var`/`_std`/`count` are registered buffers, an eval-mode fold was persisted via `state_dict()` into the next checkpoint and changed the whitening an exported policy applies, with nothing raised or logged. The fix is a two-line mode gate at the top of `update()`, matching the upstream rsl_rl implementation this class was ported from. Verified against the tree: no in-repo caller invokes `update()` directly except `forward()`, which already gates on `self.training` before dispatching, so every existing call site is unaffected; `BaseRLAlgo.evaluate` keeps its independent `update=False` freeze and now gets the same guarantee from the mode flip it already performs. The class/`update` docstrings and the `docs/training/rl.md` component row are updated to state the two-entry-point coverage.

## What's good

- Regression tests pin both directions: the eval-mode freeze through each entry point, the checkpoint-drift consequence via `state_dict()` comparison, training-mode folding still working (guards against disabling `update` outright), and the train -> eval -> train round trip `BaseRLAlgo.evaluate` performs. The description reports 3 failed / 5 passed pre-fix, 8 passed post-fix.
- Changelog recorded as a fragment (`changelog.d/2426-rl-normalizer-eval-mode-freeze.md`) per AGENTS.md step 3, not appended to `CHANGELOG.md`.
- Scope discipline: the diff touches exactly the branch under fix plus its docs and tests; `until` warmup freeze and empty-batch no-op are untouched, and the parametrized `[forward]` case bounds the change from the other side.

PR #2427, opened 2026-08-18
test(docs): grade docstring cross-references in every tree, with one member rule

cagataycali:
## Problem

`tests/test_docstring_xref_roles_resolve.py` grades every fully-qualified
`strands_robots.*` Sphinx role against the real API, because such a role
"promises the reader an importable API object at that exact dotted path". It
scanned `strands_robots/` only.

The test trees carry a comparable population of the same promise, and two
pointers there had already rotted:

```
tests/mesh/test_mesh.py::<module>     -> strands_robots.mesh_session.get_session
tests/mesh/test_mesh_rpc.py::<module> -> strands_robots.mesh_session.get_session
```

`strands_robots/mesh_session.py` existed once and was folded into the `mesh`
package; there is no such module today. Both docstrings describe how the suite
injects a `MagicMock` Zenoh session, so this is the first pointer a maintainer
working on mesh sessions follows - and it is also the wrong seam. The fixtures
patch the name the *importing* module holds:

```python
# tests/mesh/test_mesh.py
with patch("strands_robots.mesh.core.get_session", return_value=fake_session):
# tests/mesh/test_mesh_rpc.py
patch.object(mesh_session, "get_session", ...)   # mesh_session = strands_robots.mesh.session
patch.object(mesh_core,    "get_session", ...)   # mesh_core    = strands_robots.mesh.core
```

`mesh_session` is a local import alias, which is where the bogus module path
came from.

## Widening the scan alone would have reported false positives

The guard already resolves a member permissively for short-form roles
(`_has_member`: an attribute, a dataclass field, an annotated class attribute
anywhere in the MRO, a `__slots__` entry, or a name the class assigns to
`self`). The qualified resolver walked members with a bare `hasattr`, so the
same member got opposite verdicts depending on which spelling a role used:

| target | qualified `_resolves` | short-form `_short_form_resolves` |
|---|---|---|
| `strands_robots.inference.server.PolicyServer.port` | `False` | `True` |
| `strands_robots.rendering.HybridCompositor.default_width` | `False` | `True` |
| `strands_robots.simulation.base.SimEngine.get_observation` | `True` | `True` |
| `strands_robots.simulation.predicates.base_below_z` | `False` | `False` |

Both disagreeing members exist; each is assigned in `__init__`. Widening the
scope with the old resolver reports them as dead pointers, and the only remedy
such a report offers is deleting a cross-reference that was correct:

```
graded=1173  offenders=2
   tests/_daemon_executor.py::<module> -> ['strands_robots.hardware_robot.Robot._executor']
   tests/policies/vera/test_vera_render_width_domain.py::<module>
       -> ['strands_robots.rendering.HybridCompositor.default_width']
```

So the shared member rule is a prerequisite for the widening, not a separate
cleanup. The package happens to contain no such citation today, which is why
nothing surfaced it.

## Change

- `_resolves` walks members through `_has_member`, the rule the short-form half
  already applied. One concept, one rule.
- The scan covers `strands_robots/`, `tests/` and `tests_integ/` (a missing tree
  is skipped, not assumed). Offender keys are repo-relative.
- The sweep reports how many targets it graded, with a floor, matching the
  sibling `_MINIMUM_GRADED_SHORT_FORM_ROLES` so a scan that stops reaching
  docstrings fails loudly instead of reporting a clean tree.
- The two docstrings now name the seams their fixtures patch.
- The module's own illustrative dead path is written as a literal rather than a
  role, so this guard is graded by itself instead of needing an exemption from
  it.

After: **1173 qualified targets graded across 1340 files** (220 package, 1072
`tests/`, 48 `tests_integ/`), 0 offenders.

## Tests

Pre-fix, with only the guard file taken from this branch so the two source
docstrings are still the ones on `main`:

```
tests/test_docstring_xref_roles_resolve.py::test_qualified_strands_robots_xref_roles_resolve FAILED
  Offending docstrings: {'tests/mesh/test_mesh.py::<module>': ['strands_robots.mesh_session.get_session'],
                         'tests/mesh/test_mesh_rpc.py::<module>': ['strands_robots.mesh_session.get_session']}
1 failed, 16 passed
```

Post-fix: 17 passed.

New cases:

- `test_both_spellings_agree_that_a_member_is_a_member` - parametrized over two
  `__init__`-assigned attributes, with premises asserting no class-level
  attribute exists and that the short form does resolve it. This is the
  root-cause pin; it fails on `main`'s resolver for both cases.
- `test_the_sweep_reaches_the_test_trees` - non-vacuity for the scope, so a
  revert to package-only cannot pass silently.
- `test_a_qualified_target_that_names_nothing_is_still_reported` - control:
  `_has_member` widens what counts as *evidence*, not what counts as a *claim*,
  so a path naming no object stays an offender. Passes on both trees.

## Verification

Measured at `f524d58e`. Blast radius is the 155 test files that read `AGENTS.md`,
walk a source tree, or inspect docstrings/changelog fragments, plus the two
edited mesh modules:

|---|---|---|---|
| this branch | 9 | 6471 | 53 |
| `upstream/main` | 9 | 6467 | 53 |

Identical failing-ID sets (branch-only and base-only both empty); `6471 - 6467`
is exactly the 4 new test items. The 9 shared failures are pre-existing: 8 are
`ModuleNotFoundError: No module named 'awsiot'`, 1 is an unrelated
`_safety_publishers_lock` `AttributeError`.

tests_integ`; `mypy` 19 pre-existing errors on both trees with an empty
branch-only diff.

No visual artifact: this changes a test guard, three docstrings and one
convention line, with no policy, simulation, rendering, recording or asset
behavior touched.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1356,6 +1356,7 @@ Corrections from code review that apply to all future contributions:
 - **`robot.py` is for the `Robot()` factory**, the user-facing entry point. Hardware-specific code lives in `hardware_robot.py`. Don't have two files both named "robot something" with different responsibilities.
 - **Reference module names, not filenames, in docstrings** - `strands_robots.hardware_robot` not `robot.py`. Filenames change; module paths are the public contract.
 - **Keep a cross-reference target on one line** - a `:class:`/`:func:`/`:meth:` path is only a dotted path while it is contiguous. Wrapping `:class:`~strands_robots.policies.protomotions.motion_utils.MotionPlayer`` over a line break leaves a token carrying a newline and the next line's indentation, which imports nowhere. Break the prose before the role and give the path its own line.
+- **A cross-reference in a test docstring is graded too** - the roles in `tests/` and `tests_integ/` are checked against the real API alongside the package's, because a test module's docstring is where a maintainer working on that subsystem starts reading. Name the seam a fixture actually patches (`strands_robots.mesh.core.get_session`), not a local import alias dressed up as a module path.
 
 ### Unicode & String Hygiene
 - **No emojis in user-facing strings** - this is a project rule. Tool result dicts (`{"content": [{"text": ...}]}`), log messages, error messages: plain ASCII only. Agents read these strings programmatically; emojis just add tokenizer noise.
--- a/changelog.d/2427-xref-guard-scope-and-member-rule.md
+++ b/changelog.d/2427-xref-guard-scope-and-member-rule.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **tests**: the docstring cross-reference guard now grades `tests/` and `tests_integ/` alongside the shipped package, and resolves a qualified target's members with the same permissive rule the short-form half already used. Two test-module docstrings cited `strands_robots.mesh_session.get_session`, a module that was folded into the `mesh` package; they now name the seam their fixtures patch.
--- a/tests/mesh/test_mesh.py
+++ b/tests/mesh/test_mesh.py
@@ -2,7 +2,7 @@
 
 All tests are 100% mocked.  No ``eclipse-zenoh`` install required: a
 ``MagicMock`` session is injected in place of the real Zenoh session by
-patching :func:`strands_robots.mesh_session.get_session`.
+patching :func:`strands_robots.mesh.core.get_session`.
 """
 
 from __future__ import annotations
--- a/tests/mesh/test_mesh_rpc.py
+++ b/tests/mesh/test_mesh_rpc.py
@@ -2,7 +2,8 @@
 
 All tests are 100% mocked.  No ``eclipse-zenoh`` install required: a
 ``MagicMock`` session is injected in place of the real Zenoh session by
-patching :func:`strands_robots.mesh_session.get_session`.
+patching :func:`strands_robots.mesh.session.get_session` and the reference
+:mod:`strands_robots.mesh.core` imported it under.
 """
 
 from __future__ import annotations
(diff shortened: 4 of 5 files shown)


yinsong1986 (approved):
## Summary

Widens the docstring cross-reference guard from the shipped package to `tests/` and `tests_integ/`, and unifies member resolution so the qualified and short-form spellings of the same target get the same verdict (`_resolves` now walks members through `_has_member` instead of a bare `hasattr`). Fixes the two rotted `strands_robots.mesh_session.get_session` pointers the widening surfaced, pointing them at the seams the fixtures actually patch — I verified `strands_robots.mesh.core.get_session`, `strands_robots.mesh.session.get_session`, and `strands_robots.mesh.core` all resolve on this head. The member-rule change moves the guard only in the permissive direction (under-report, never a false accusation), the ordering matters (shared rule first, then widen — otherwise two correct citations become false offenders), and the root cause, the scope, and the control case each get their own pin. Test-only change with no public API, wire-format, or persisted-schema surface.

## What's good

- The two `__init__`-assigned parametrized cases in `test_both_spellings_agree_that_a_member_is_a_member` carry premise assertions (`not hasattr`, short form resolves), so the pin fails on main's resolver for the stated reason rather than incidentally.
- `test_the_sweep_reaches_the_test_trees` plus the raised `_MINIMUM_GRADED_QUALIFIED_ROLES` floor make a silent revert to package-only scope impossible.
- The module's own illustrative dead path is now a literal instead of a role, so the guard grades its own file without an exemption.
- Changelog fragment follows the convention; the AGENTS.md addition is one ASCII line in the matching section.

## Verification suggestions

One pre-existing (not introduced here) sensitivity worth knowing when spot-checking locally: `_resolves` treats a module whose optional dependency is absent as unimportable, so in a partial install the qualified sweep reports `strands_robots.device_connect.*` citations as offenders — main's guard fails identically in the same environment (`utils.py::tcp_port_error`, `utils.py::boolean_flag_error`). Run `tests/test_docstring_xref_roles_resolve.py` under a full `[all,dev]` install to reproduce the PR's `1173 graded / 0 offenders` figure.

PR #2429, opened 2026-08-18
fix(training/rl): refuse a batch EmpiricalNormalization cannot fold, and commit the count after the fold

cagataycali:
## Problem

`EmpiricalNormalization` reads row 0 of whatever it is handed as the batch axis, and whitens by
broadcasting against buffers of shape `(1, *shape)`. Neither entry point checks that the tensor is
the `(batch, *shape)` its own docstring asks for, so a batch that is not that shape is reshaped
rather than refused.

Probing a normalizer built for per-feature shape `(4,)` after six well-formed folds:

| input | verdict on `main` | returned | `count` | statistics |
|---|---|---|---|---|
| `(4, 4)` | accepted | `(4, 4)` | +4 | correct |
| `(1, 4)` | accepted | `(1, 4)` | +1 | correct |
| `(0, 4)` | accepted, no-op | `(0, 4)` | +0 | untouched |
| `(4,)` unbatched | **accepted silently** | `(1, 4)` | **+4 for one observation** | pooled into one number |
| `(3, 1)` | **accepted silently** | `(3, 4)` | +3 | one feature written into all four |
| `(3, 3)` | raises out of torch | - | **+3 leaked** | untouched |
| `(2, 1, 4)` | raises out of torch | - | **+2 leaked** | untouched |
| `()` scalar | raises `IndexError` | - | +0 | untouched |

So the same caller error produces three different outcomes depending only on whether the arithmetic
happens to broadcast, and there is no shape for which a failure is clean. Two halves:

**Silent reshaping.** `(4,)` is the shape a single-environment rollout produces without `unsqueeze`.
Folding it reads the observation's four features as four samples of a one-feature stream, and
`forward` returns `(1, 4)` - a different shape than was passed in, which the method documents as
"same shape as `x`". All four persisted buffers drift, with nothing raised and nothing logged. The
same reshaping happens on the pure whitening path (`update=False` in eval mode), which is what
`BaseRLAlgo.evaluate` and an exported policy use.

**A leaked sample count.** `self.count += count_x` ran before the arithmetic that fails, so a batch
that raised had already advanced the count. That is not recoverable: `rate = count_x / count` shrinks
permanently, so the estimator under-weights every later batch. Feeding an identical stream whose
true mean is exactly `10.0`:

| mis-shaped batches rejected first | `count` after 5 clean folds | estimated mean |
|---|---|---|
| 0 | 20 | 10.000 |
| 1 | 23 | 8.696 |
| 3 | 29 | **6.897** |

It also brings the `until` warmup freeze forward - with `until=20`, three rejected batches reduce a
stream from 5 effective folds to 3 - and `count` is written into the next checkpoint.

## Why this is the class's own contract

`if count_x == 0: return` is not in the ported RSL-RL upstream; it was added here, and
`test_empty_batch_update_is_noop` pins it with the sentence *"A zero-row batch must not advance the
count or divide by zero."* That is exactly the rule this change completes: a batch that cannot be
folded must leave the estimator untouched. It held for zero rows and for no other shape.

This is not an upstream-parity fix. Both RSL-RL and the Holosoma port advance the count before the
arithmetic too, so the argument is the documented `x: Batched input, shape (batch, *shape)` and
`Returns: ... same shape as x` on this class.

## Change

One shape rule, `_batch_shape_error`, consulted by both `forward` and `update` - mirroring how the
eval-mode freeze is checked at both entry points - naming the received shape, the required shape and
the remedy:

```
EmpiricalNormalization.forward: expected a batched tensor of shape (batch, 12), got (12,);
add the leading batch dimension (x.unsqueeze(0) for a single observation).
```

A wrong feature count gets the other remedy ("the trailing dimensions must be the per-feature shape
`(4,)`"). The check runs before the mode check so both entry points answer identically in either
mode.

The fold arithmetic is hoisted above the first buffer write and `count` is committed last, so a
rejected batch changes nothing. This still matters with the shape check in place: an integer or
boolean observation tensor has an acceptable shape and still raises inside `torch.var_mean`, and now
does so without advancing the count. Restricting dtypes is deliberately out of scope - torch's own
error is kept, just made atomic.

The accepted path is unchanged. Over a 10-fold random stream both trees produce a bit-identical
mean/std/count trace (fingerprint `4e7724f883344d5d5235c758c17d379f`), and the empty-batch no-op is
the same rule at zero rows.

## Tests

`tests/training/test_rl_normalization.py` grows from four contract edges to five. **16 failed / 9
passed on `main`, 25 pass after**, every pre-fix failure behavioural:

- `test_a_batch_that_cannot_be_folded_leaves_the_estimator_untouched` - the root-cause property,
  parametrized over five unfoldable shapes x both entry points.
- `test_whitening_never_returns_a_shape_the_caller_did_not_pass` - the documented return contract,
  over the three inputs that come back a different shape (`(4,)` -> `(1,4)`, `(3,1)` -> `(3,4)`,
  `()` -> `(1,4)`).
- `test_the_deployed_eval_mode_path_refuses_an_unbatched_observation` - the `update=False` path an
  exported policy uses.
- `test_the_refusal_names_the_required_shape_and_its_remedy_works` - parses `x.unsqueeze(0)` out of
  the message and applies it, so the message is pinned as true rather than as a sentence.
- `test_a_fold_that_fails_on_dtype_does_not_advance_the_sample_count` - the half the shape check
  cannot screen.
- `test_two_folds_reproduce_the_population_statistics_of_their_union` - the accepted path, pinned
  against the exact population mean/std of the two batches taken together rather than a snapshot.

The nine that pass on both trees are the controls: `center=False`, the `until` freeze, the
empty-batch no-op, the eval-mode freeze through either entry point, the checkpoint round trip, and
`update` still folding in training mode. They fail for the obvious wrong fixes - making the guard
unconditional, or refusing an empty batch.

## Verification

57-file selection (every test touching the normalizer or `training/rl`, plus the repo-wide docstring,
docs, changelog and source-string guards), same selection on both trees:

|---|---|---|---|---|
| this branch | 3 | 2953 | 5 | 2961 |
| `upstream/main` (new test file copied in) | 19 | 2935 | 5 | 2959 |

Branch-only failing ids: none. Base-only: exactly the 16 pre-fix failures of the new test file. The
three shared failures are pre-existing (`draccus` resolves below the declared `>=0.11.6` floor in
this environment). The collected delta of 2 is `update` entering
`tests/test_args_docstring_completeness.py`, which it had escaped for want of an `Args:` block; both
new cases pass. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`;
`mypy` unchanged (19 pre-existing on both trees, none in the changed files).

## Rollout

`so101` in MuJoCo (headless) supplies a real observation stream - 6 joint positions + 6 velocities,
`D=12`. Sixty well-formed `(1, 12)` folds, then one `(12,)` observation, then a fixed actor is fed
the same observation whitened by whatever each tree's next checkpoint would carry.

EmpiricalNormalization batch-shape contract

Before: accepted, returned `(1, 12)`, `count` 60 -> 72 for a single observation, all four buffers
written to the next checkpoint, and the actor's action moves by **0.320** max-abs on a tanh-bounded
action space. After: refused by name, `count` 60 -> 60, no buffer touched, action delta **0.0000**.


yinsong1986 (approved):
## Summary

Closes two halves of one contract on `EmpiricalNormalization`: both entry points now validate that the input is a `(batch, *shape)` tensor for the per-feature shape the buffers were built for (refusing with the received shape, required shape, and remedy), and the fold arithmetic is hoisted above the first buffer write with `count` committed last, so a batch that cannot be folded — mis-shaped or wrong-dtype — leaves every persisted buffer untouched. The shared `_batch_shape_error` keeps `forward` and `update` agreeing, mirroring the existing eval-freeze pattern. I verified independently that the reordered Welford update (`rate = count_x / float(int(self.count) + count_x)` before the commit) is bit-identical to the old order over a random stream, that no in-tree caller passes shapes the new guard refuses (PPO/FastSAC feed `(N, D)` batches and `SimEnv._vector` unsqueezes at `strands_robots/training/rl/env.py:212`), and that the new test file passes 25/25 at head with `ruff check`/`ruff format --check` clean on both changed Python files. The behaviour change is loud (a `ValueError` naming the remedy), which is the right direction for a guard — relaxing later is cheap, tightening after silent broadcasts leak into checkpoints is not. Head commit `6db99505` is a zero-diff rename of the changelog fragment to the PR's own number.

## What's good

- The count-leak fix is independently valuable beyond the shape check: the dtype-failure test pins the half the shape guard cannot screen, so the atomicity claim is behavioural rather than asserted.
- `test_the_refusal_names_the_required_shape_and_its_remedy_works` parses `x.unsqueeze(0)` out of the message and applies it, pinning the message as true rather than as wording.
- `update` gained a proper `Args:` entry, bringing it under `tests/test_args_docstring_completeness.py` (AGENTS.md convention 13); error strings are plain ASCII; changelog fragment follows the `changelog.d/<pr-number>-<slug>.md` rule.

yinsong1986 (approved):
## Summary

Closes a two-part shape contract on `EmpiricalNormalization`: a new `_batch_shape_error` rule, consulted by both `forward` and `update` before the mode check, refuses any tensor that is not `(batch, *shape)` for the per-feature shape the buffers were built for — naming the received shape, the required shape, and the applicable remedy — and the Welford fold is reordered so `count` is committed only after `torch.var_mean` and the buffer writes succeed, making a failed fold (including the dtype failures the shape check cannot screen) atomic instead of leaking an unrecoverable count inflation into the next checkpoint. Verified locally: all 25 tests in `tests/training/test_rl_normalization.py` pass at this head; the accepted path is bit-identical between the two trees over a 10-fold random stream (matching mean/std/count fingerprints); and every in-tree caller already satisfies the guard (`SimEnv._vector` unsqueezes to `(1, num_obs)`, PPO/FastSAC feed vectorized `(N, num_obs)` batches, and `BaseRLAlgo.evaluate` whitens the `(1, num_obs)` observations SimEnv produces), so the refusal only ever fires on inputs that were previously corrupted silently or raised non-atomically out of torch.

## What's good

- The refusal direction matches AGENTS.md ("Raise on fatal errors", "No silent defaults on error"): the previously-accepted `(num_obs,)` input was silent data corruption, and the replacement is an explicit `ValueError` carrying its own remedy — pinned by a test that parses `x.unsqueeze(0)` out of the message and applies it.
- The guard stays local to the class rather than growing a `utils.py` helper for a single caller (AGENTS.md rule 11).
- The count-after-fold reorder is pinned two ways: a dtype-failure test for atomicity, and a union-statistics test that pins the accepted arithmetic against the exact population answer rather than a snapshot.
- Changelog fragment named for the PR, docs table row updated in the same PR, new `Args:` entries land both `_batch_shape_error` params in the docstring-completeness guard.

cagataycali:
Corroborating this from the other direction, in case it is useful for the record: the shape this refuses is the one **this package's own gymnasium adapter produces**, so the reachable caller is someone following our own integration path rather than making a mistake.

`GymSimEnv._to_numpy_obs` flattens the `(1, D)` actor observation to `(D,)` — that is the gymnasium contract, and its docstring points the reader at a running normalizer for exactly that observation (*"Observation bounds are `[-inf, inf]` (joint positions / velocities are unbounded a priori); SB3 handles this with a running normalizer"*). We ship one, `docs/training/rl.md` lists it, and before this change it mis-handled that shape.

Driving a real SO-101 through `SimEnv` -> `GymSimEnv` for 60 steps in MuJoCo and feeding each observation to `EmpiricalNormalization(env.num_actor_obs)`:

| | before #2429 | after |
|---|---|---|
| `observation_space` | `Box(-inf, inf, (6,), float32)` | same |
| `norm(obs)` for a `(6,)` input | `(1, 6)` | `ValueError` naming both shapes |
| `count` after 60 observations | **360** (6x) | 60 |
| distinct per-feature means | **1 of 6** (all `0.101093`) | 6 of 6 |

The last row is the part that matters: every joint was whitened by one pooled number, so the per-feature separation the estimator exists to maintain was gone — and since the statistics are persistent buffers it was in `state_dict()` and therefore in the next checkpoint. Six good `(1, 4)` batches of `[1, 2, 3, 4]` leave `_mean = [1, 2, 3, 4]` / `_std = [0, 0, 0, 0]`; one flat `(4,)` observation of the *same values* then left `_mean = [1.6, 2.2, 2.8, 3.4]` / `_std = [1.0198, 0.7483, 0.7483, 1.0198]`.

All of that is refused on merged `main`. Two small follow-ups this suggests, both optional:

- The class docstring explains the broadcast mechanism but does not name `GymSimEnv` as the surface that emits the refused shape. A pointer there would connect the refusal a caller hits to the adapter that handed them the observation.
- Separately, and not about this change: four `tests_integ/training/test_so101_pick_place_e2e.py` tests fail on `main` with `observation keys not produced by the engine: ['shoulder_pan', ...]; available: ['1', '1.vel', ...]`. The so101 MJCF names its joints `1`..`6`, so those `actor_obs_keys` cannot resolve. `tests_integ` is not in pytest `testpaths`, so CI never runs it and the drift is invisible.

PR #2430, opened 2026-08-18
fix(tests): an absent extra is an unknown, not a dead cross-reference

### LINKED ISSUE #2428: bug(tests): the xref guard reads an absent extra as a dead pointer, and a lazy import failure stops the sweep grading anything
## What happens

`tests/test_docstring_xref_roles_resolve.py::_resolves` imports the longest
importable prefix of a qualified target, falling back to a shorter one when an
import fails:

```python
for i in range(len(parts), 0, -1):
    try:
        module = importlib.import_module(".".join(parts[:i]))
        consumed = i
        break
    except Exception:
        continue
```

An `ImportError` has two causes here and they need opposite verdicts:

* the prefix **names nothing** - `strands_robots.mesh_session`, the rot this
  guard exists to find;
* the prefix **exists in the tree** and an optional third-party dependency of it
  is not installed - `strands_robots.tools.pose_tool` imports `serial`.

The second case falls back to the shorter `strands_robots.tools`, which does
import, then walks `pose_tool` as a *member* of it, fails, and reports a
citation that was correct as a dead pointer.

Measured on `b107aecf` with a `[dev]`-only install (no `torch`, `serial`,
`psutil`, `awsiot`, `lerobot`):

```
graded = 1173    offenders = 39
```

All 39 are correct citations. A sample, with the extra each one waits on:

| target | absent |
|---|---|
| `strands_robots.device_connect.reachy_mini_driver.ReachyMiniDriver` | `awsiot` |
| `strands_robots.policies.lerobot_local.processor.ProcessorBridge` | `lerobot` |
| `strands_robots.training.rl.env.SimEnv` | `torch` |
| `strands_robots.training.rl.ppo.PpoTrainer` | `torch` |

## A second, distinct failure: the sweep does not merely mis-grade, it stops

`_has_member` begins with `hasattr(owner, attr)`, and `strands_robots.tools`
imports its submodules from a module-level `__getattr__`. `hasattr` swallows
only `AttributeError`, so a `ModuleNotFoundError` raised inside that lazy import
propagates out of the sweep: the test errors with a traceback naming `serial`,
and no docstring is reported at all.

```
ModuleNotFoundError: No module named 'serial'
  tests/test_docstring_xref_roles_resolve.py:148  _offending_roles_in
  tests/test_docstring_xref_roles_resolve.py:112  _resolves
  tests/test_docstring_xref_roles_resolve.py:333  _has_member  ->  hasattr
  strands_robots/tools/__init__.py:39             __getattr__
```

Four targets reach it:

| citing file | target | absent |
|---|---|---|
| `strands_robots/tools/serial_tool.py` | `strands_robots.tools.pose_tool` | `serial` |
| `strands_robots/utils.py` | `strands_robots.tools.lerobot_teleoperate` | `psutil` |
| `tests/tools/test_public_member_docstrings.py` | `strands_robots.tools.lerobot_train.SessionManager` | `psutil` |
| `tests/tools/test_serial_tool_numeric_domain.py` | `strands_robots.tools.pose_tool` | `serial` |

This is the worse half: an over-report is at least legible as a list of
docstrings, whereas here the guard grades nothing and the traceback names a
dependency rather than a cross-reference, so it does not read as this guard's
finding at all.

## Why it is worth fixing

The only remedy such a report offers is deleting a cross-reference that was
correct, which is the expensive-in-the-diligent-direction shape: acting on it
looks like housekeeping and removes 39 working pointers.

The rule is already written down in the same file - in the *other* half. From
`_package_modules`, which feeds the short-form sweep:

> A module whose optional dependency is absent simply does not appear, so this
> guard grades the roles it can reach and never turns a missing extra into a
> dead pointer.

The qualified half does exactly what that sentence forbids, and the module
docstring's own claim that the guard "can only ever under-report" is false while
it does. This is the same asymmetry #2427 closed for *member* resolution, on the
import side: one concept, two rules.

## Why CI never showed it

CI installs `[all,dev]`, so no module is unimportable, nothing is
misclassified, and the sweep reports `1173 graded / 0 offenders`. The exposure
is entirely local, which is also where a contributor is most likely to act on
it. Widening the scan to `tests/` and `tests_integ/` in #2427 grew the affected
population - 22 of the 39 and 2 of the 4 are in the test trees.

## Proposed fix

Discriminate on the name the exception carries, not on its type. A
`ModuleNotFoundError` whose `name` is a `strands_robots` path is a dead pointer;
one naming anything else is a statement about the environment, so the target is
not decidable here and is skipped rather than accused - the tri-state
`_short_form_resolves` already returns for a target it cannot decide. Apply the
same guard to the member walk so a lazy import failure is skipped instead of
escaping.

A control matters as much as the fix: a genuinely absent `strands_robots` module
must stay an offender, or the change would launder real rot into an unknown.

cagataycali:
Closes #2428

## Problem

`_resolves` imports the longest importable prefix of a qualified target and falls
back to a shorter one whenever an import fails:

```python
try:
    module = importlib.import_module(".".join(parts[:i]))
    consumed = i
    break
except Exception:
    continue
```

A failed import has two causes there, and they need opposite verdicts:

* the prefix **names nothing** - `strands_robots.mesh_session`, the rot this guard
  exists to find;
* the prefix **exists in the tree** and an optional dependency of it is not
  installed - `strands_robots.device_connect.robot_driver` imports `awsiot`.

The second falls back to the shorter `strands_robots.device_connect`, which does
import, then looks for `robot_driver` as a *member* of it, does not find one, and
reports a citation that was correct as a dead pointer. The only remedy such a
report offers is deleting a cross-reference that worked.

The guard's module docstring already claims it "can only ever under-report", and
the rule is written down in the same file - in the *other* half. From
`_package_modules`, which feeds the short-form sweep:

> A module whose optional dependency is absent simply does not appear, so this
> guard grades the roles it can reach and never turns a missing extra into a dead
> pointer.

The qualified half did exactly what that sentence forbids. Same asymmetry #2427
closed for *member* resolution, on the import side: one concept, two rules.

## The sweep did not merely mis-grade - it stopped

`_has_member` begins with `hasattr(owner, attr)`, and `strands_robots.tools`
imports its submodules from a module-level `__getattr__`. `hasattr` swallows only
`AttributeError`, so a `ModuleNotFoundError` raised inside that lazy import
propagated out of the sweep:

```
FAILED tests/test_docstring_xref_roles_resolve.py::test_qualified_strands_robots_xref_roles_resolve
  ModuleNotFoundError: No module named 'serial'
```

No docstring was reported at all, and the traceback named a dependency rather
than a cross-reference, so it did not read as this guard's finding. This is the
worse half: an over-report is at least legible as a list of docstrings.

## Change

- `_absent_dependency(exc)` reads the name the exception carries. A
  `strands_robots` path means no such module exists; any other name means the
  cited module does exist and this environment cannot confirm it.
- `_resolves` returns `bool | None`, the tri-state `_short_form_resolves` already
  returns for a target it cannot decide. `None` is not an offender.
- The member walk catches the same case, so a lazy import failure is skipped
  rather than escaping.
- `_offending_roles_in` tests `is False` rather than truthiness, since `not None`
  is true and would read an undecided target as a rejected one.

`except Exception: continue` is deliberately left in place after the new clause:
it still covers a prefix that fails to import for a reason other than an absent
module, and narrowing it is a separate question this change does not settle.

The claim count is untouched - `graded` counts targets the pattern finds, not
resolution outcomes - so `_MINIMUM_GRADED_QUALIFIED_ROLES` still covers exactly
the same population and a scope regression still fails loudly.

## Tests

Measured on `4d0a1fc5` with a `[dev]`-only install, running the resolver over
every qualified claim in the tree under the suite's own conftest:

| | claims graded | reported dead | ended the sweep |
|---|---|---|---|
| `origin/main` | 1173 | **10** | **4** |
| this branch | 1173 | 0 | 0 |

Every one of the 10 is correct - all `strands_robots.device_connect.*`, waiting on
`awsiot`. The 4 are `strands_robots.tools.pose_tool` (x2, `serial`),
`strands_robots.tools.lerobot_teleoperate` and
`strands_robots.tools.lerobot_train.SessionManager` (`psutil`). Without the
conftest torch mock the same sweep reports 39 rather than 10, which is itself the
point: the verdict tracked the install, not the docstrings.

The guard file alone, same environment:

```
origin/main   1 failed, 16 passed
this branch   20 passed
```

New cases:

- `test_a_target_whose_module_needs_an_absent_extra_is_not_a_dead_pointer` - the
  root-cause pin, with a premise asserting the target resolves while its module
  imports. The absent extra is *staged* rather than sampled, because which extras
  are missing varies per install and is none of them in CI, so a test keyed on a
  real absence would assert nothing there.
- `test_a_lazy_import_failure_in_the_member_walk_does_not_end_the_sweep` - pins
  the crash, with a `pytest.raises` premise proving the exception really does
  escape `hasattr`.
- `test_a_module_the_package_does_not_have_is_still_a_dead_pointer` - the control
  that keeps the fix honest. Both cases raise `ModuleNotFoundError` and only the
  name separates them, so this is what stops the skip laundering real rot into an
  unknown; it asserts the discriminator directly on both sides and re-pins
  `strands_robots.mesh_session.get_session`.
- `test_guard_resolver_accepts_real_symbol_and_rejects_bogus` now asserts
  `is True` / `is False` instead of truthiness, for the reason above.

## Verification

Blast radius is this guard plus the sibling docstring, changelog-fragment and
host-path guards, measured on both trees in the same environment:

|---|---|---|---|
| this branch | 0 | 128 | 7 |
| `origin/main` | 1 | 124 | 7 |

`128 - 124` is exactly the 3 new items plus the one that fails on `main`.
`tests/tools/` cannot be collected in this environment at all - its conftest
imports `serial` - identically on both trees.

`ruff check` and `ruff format --check` clean on the changed file; no non-ASCII.

No behavior change in CI, where every extra is installed: nothing is
undecidable, so the sweep reports the same 1173 graded / 0 offenders as before.
Test-only change - no production code, no public API, wire-format or persisted
schema surface, and no visual artifact.

### DIFF
--- a/changelog.d/2428-xref-guard-absent-extra-is-not-rot.md
+++ b/changelog.d/2428-xref-guard-absent-extra-is-not-rot.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **tests**: the docstring cross-reference guard no longer reads a module whose optional dependency is absent as a dead pointer. Such a target is undecidable in that environment rather than wrong, so it is skipped - the rule the guard's short-form half already applied - while a path naming no `strands_robots` module stays reported. A lazy module-level `__getattr__` raising `ModuleNotFoundError` through `hasattr` no longer ends the sweep with nothing graded.
--- a/tests/test_docstring_xref_roles_resolve.py
+++ b/tests/test_docstring_xref_roles_resolve.py
@@ -58,6 +58,13 @@ unqualified role (``:func:`reset```), a short-form role whose head names nothing
 the package defines, and roles into third-party packages - all of which resolve
 against Sphinx's current-module context or an intersphinx inventory that is not
 available, so flagging them would produce false positives.
+
+One further target is undecidable for a reason that lives in the environment
+rather than in the docstring: a cited module whose own optional dependency is not
+installed. The module exists in the tree, so the citation is sound, but nothing
+here can confirm the object it names, and reporting it accuses a correct pointer
+whose only remedy would be deletion. Such a target is skipped - the rule the
+short-form half has always applied to the modules it indexes.
 """
 
 from __future__ import annotations
@@ -85,8 +92,24 @@ _REPO_ROOT = _PKG_ROOT.parent
 _ROLE_RE = re.compile(r":(?:mod|class|func|meth|attr|data|obj|exc):`~?([A-Za-z_][\w.\s]*)`")
 
 
-def _resolves(target: str) -> bool:
-    """True if ``target`` names a real ``strands_robots`` module or attribute.
+def _absent_dependency(exc: ImportError) -> str | None:
+    """The absent third-party module ``exc`` names, if that is what it reports.
+
+    Both causes of a failed import here raise the same exception type and need
+    opposite verdicts, and the name it carries is what separates them. A
+    ``strands_robots`` path means no such module exists - the rot being graded. A
+    name outside the package means the cited module does exist and could not be
+    imported because an extra is not installed, which is a fact about this
+    environment and not about the docstring.
+    """
+    name = getattr(exc, "name", None)
+    if not name or name == "strands_robots" or name.startswith("strands_robots."):
+        return None
+    return name
+
+
+def _resolves(target: str) -> bool | None:
+    """Grade one fully-qualified ``target`` against the real API.
 
     Imports the longest importable module prefix, then walks the remaining
     dotted components as members (so ``pkg.mod.Class.method`` resolves), using
@@ -94,6 +117,18 @@ def _resolves(target: str) -> bool:
     ``hasattr`` walk here would report a dataclass field with no class-level
     default, and an attribute only ever assigned in ``__init__``, as a dead
     pointer - while the identical member cited in the short form resolved.
+
+    The member walk is guarded for the same reason the imports are: a package
+    that imports its submodules from a module-level ``__getattr__`` raises
+    ``ModuleNotFoundError`` from inside ``hasattr``, which swallows only
+    ``AttributeError``. Left to propagate, one absent extra ends the sweep and
+    nothing is graded at all.
+
+    Returns:
+        ``True`` when every component resolves, ``False`` when the path names
+        nothing, and ``None`` when an optional dependency is absent, so this
+        environment cannot decide the target either way. The tri-state is the one
+        :func:`_short_form_resolves` already returns for an undecidable target.
     """
     parts = target.split(".")
     module = None
@@ -101,17 +136,26 @@ def _resolves(target: str) -> bool:
     for i in range(len(parts), 0, -1):
         try:
             module = importlib.import_module(".".join(parts[:i]))
-            consumed = i
-            break
+        except ImportError as exc:
+            if _absent_dependency(exc):
+                return None
+            continue
         except Exception:
             continue
+        consumed = i
+        break
     if module is None:
         return False
     obj = module
     for attr in parts[consumed:]:
-        if not _has_member(obj, attr):
-            return False
-        obj = getattr(obj, attr, obj)
+        try:
+            if not _has_member(obj, attr):
+                return False
+            obj = getattr(obj, attr, obj)
+        except ImportError as exc:
+            if _absent_dependency(exc):
+                return None
+            raise
     return True
 
 
@@ -145,7 +189,7 @@ def _offending_roles_in(doc: str) -> list[str]:
             continue
         if target != contiguous:
             offenders.append(f"{target!r} is wrapped over a line break; write it as {contiguous!r}")
-        elif not _resolves(target):
+        elif _resolves(target) is False:
             offenders.append(target)
     return offenders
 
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Test-only change to the qualified-xref docstring guard: a `ModuleNotFoundError` raised while importing a cited module is now discriminated by the name the exception carries. A `strands_robots.*` name stays a dead pointer (the rot the guard exists to find, re-pinned by the control test on `strands_robots.mesh_session.get_session`); any other name means the module exists but an optional extra is absent, so `_resolves` returns the same `None` tri-state `_short_form_resolves` already uses and the target is skipped rather than reported. The member walk gains the same guard, closing the sweep-ending crash where a lazy module-level `__getattr__` raised through `hasattr`. `_offending_roles_in` correctly tests `is False` so `None` is not read as an offender, and the updated resolver test asserts exact verdicts rather than truthiness for the same reason. Verified locally at `4d0a1fc5`: all 20 tests in the guard file pass, `ruff check`/`format --check` clean, no non-ASCII, no host paths.

## What's good

- The control test (`test_a_module_the_package_does_not_have_is_still_a_dead_pointer`) asserts the discriminator on both sides, which is what keeps the skip from laundering real rot into an unknown - the exact hazard this kind of fix usually introduces.
- The staged-failure approach (`_import_module_lacking`) keeps the tests meaningful in full-install CI where no extra is actually missing, and the sentinel module name is self-describing in a traceback.
- The deliberate `raise` in the member walk for a `strands_robots`-named import failure keeps an internal broken import loud rather than silently undecidable, and the diff says so explicitly.
- Changelog fragment follows the `changelog.d/` convention (issue number is explicitly allowed by `changelog.d/README.md`), closing keyword is in the PR body per AGENTS.md step 6, and the `graded` count is untouched so the `_MINIMUM_GRADED_QUALIFIED_ROLES` scope floor still binds.

PR #11, opened 2026-03-09
feat: 18 agent tools for robot control, simulation, training, and teleoperation

cagataycali:
## Summary

Adds **12 new agent tools** (bringing total to 18) that give AI agents complete control over robot hardware, simulation, training, and teleoperation through natural language.

### New Tools
| Tool | Description |
|------|-------------|
| `inference` | Universal VLA inference (any policy provider) |
| `isaac_sim` | NVIDIA Isaac Sim scene control |
| `lerobot_dataset` | LeRobot dataset management (record, upload, browse) |
| `marble_tool` | MARBLE visual grounding for manipulation |
| `newton_sim` | Newton differentiable sim control |
| `reachy_mini` | Pollen Reachy Mini expressive robot |
| `robot_mesh` | Zenoh mesh multi-robot coordination |
| `stereo_depth` | Stereo depth estimation |
| `stream` | Real-time telemetry streaming |
| `teleoperator` | Teleoperation (leader-follower, VR, keyboard) |
| `use_lerobot` | LeRobot training/evaluation orchestration |
| `use_unitree` | Unitree SDK integration (Go2, G1, H1) |

### Example
```python
from strands import Agent
from strands_robots.tools.inference import inference
from strands_robots.tools.use_lerobot import use_lerobot

agent = Agent(tools=[inference, use_lerobot])
agent('Train an ACT policy on the aloha_sim dataset for 10000 steps, then evaluate')
```

### Tests — 38 passed
Parametrized import + callable tests for all 18 tools.

PR #2431, opened 2026-08-18
fix(training/rl): refuse an observation shape EmpiricalNormalization can only broadcast

cagataycali:
## Problem

`EmpiricalNormalization` keeps one running mean/std **per feature** behind a leading batch axis. Both entry points document `(batch, *shape)` in and *"the same shape as `x`"* out. A shape that merely **broadcasts** against the `(1, D)` buffers satisfied neither and was accepted silently.

Measured on `shape=(4,)`, before this change:

| input | outcome |
|---|---|
| `(2, 4)` | accepted, correct |
| `(4,)` | **returned `(1, 4)`**, `count += 4` for one observation |
| `(3, 1)` | **returned `(3, 4)`**, one column's stats written into all four features |
| `(3, 3)` | `RuntimeError` from `var_mean` |
| `(2, 1, 4)` | `RuntimeError` from `var_mean` |
| `()` | `IndexError` |

So most mis-shaped inputs already failed loudly; only the two that happen to broadcast slipped through, and those raised nothing at all.

### The reachable case is the package's own gymnasium adapter

`GymSimEnv._to_numpy_obs` flattens the `(1, D)` actor observation to `(D,)` — that is the gymnasium contract, and its own docstring points the reader at a running normalizer for it (*"Observation bounds are `[-inf, inf]` ... SB3 handles this with a running normalizer"*). The library ships one, documented in `docs/training/rl.md`, and it mis-handles exactly that shape.

Driving a real SO-101 through `SimEnv` → `GymSimEnv` for 60 steps and feeding each observation to `EmpiricalNormalization(env.num_actor_obs)`:

| | before | after |
|---|---|---|
| `norm(obs)` for a `(6,)` input | `(1, 6)` | `ValueError` naming both shapes |
| `count` after 60 observations | **360** (6x) | 60 |
| distinct per-feature means | **1 of 6** | 6 of 6 |
| reported | nothing raised, nothing logged | refusal + the call that batches it |

The statistics are **persistent buffers**, so the pooled values reach `state_dict()` and change how an exported policy whitens every observation. Six good `(1, 4)` batches of `[1, 2, 3, 4]` leave `_mean = [1.0, 2.0, 3.0, 4.0]`, `_std = [0, 0, 0, 0]`; one flat `(4,)` observation of the *same values* then leaves `_mean = [1.6, 2.2, 2.8, 3.4]`, `_std = [1.0198, 0.7483, 0.7483, 1.0198]` — the per-feature identity the module exists to keep is gone, and it is not recoverable once saved.

### The raising path mutated state too

`count` was advanced by `x.shape[0]` *before* `var_mean` ran, so a batch that failed inside the statistics update had already moved it (`count` 1 → 4 on a `(3, 3)` that raised). The Welford step weights each batch by `count_x / count`, so that left every later batch folded in at the wrong rate — a permanent skew from a batch that was never folded at all.

## Change

A module-private tuple rule, `_batched_shape_error`, consulted at the top of **both** `forward` and `update`, before any buffer is touched:

- The refusal names the shape read (`(batch, 6)`), the shape received (`(6,)`), and — for a single unbatched observation — the call that batches it (`x.unsqueeze(0)`).
- Checked before `count` moves, so a refused batch leaves the running rate alone. This also makes the class's behaviour uniform: every mis-shaped input is now one `ValueError` instead of a mix of silence, `RuntimeError` and `IndexError`.
- `forward` is covered too, not only `update`. In eval mode nothing is folded, but the whitening arithmetic still broadcasts, so a deployed policy handed a `(D,)` observation got a `(1, D)` tensor back.

The rule lives beside the class rather than in `strands_robots.utils` because those shared domains are deliberately torch-free and describe scalars and plain numeric vectors, while this describes a tensor's rank and trailing extents. The feature shape is read off `_mean` rather than stored at construction, so there is one owner of the value and `load_state_dict` cannot desynchronize it.

## Tests

Added to `tests/training/test_rl_normalization.py`, whose docstring already scopes it to this class's contract edges and enumerated four; the input-shape contract is the fifth.

**8 failed / 12 passed** on `main` → **20 passed** after. Every new test is graded through the public surface, so the file collects on `main` and each failure is behavioural (`DID NOT RAISE ValueError`, and the raw `RuntimeError`/`IndexError` where the error type was not yet uniform).

The 12 that pass on both revisions are the bound: a correctly-batched vector, a multi-dimensional per-feature shape, a one-feature stream where `(batch, 1)` is genuinely a batch, the documented zero-row no-op, and the four pre-existing edges (`center=False`, the `until` freeze, the empty-batch no-op, the eval-mode freeze).

The guard-before-`count` ordering is load-bearing rather than incidental: moving the check below `self.count += count_x` fails 4 of the new tests on exact count assertions (`assert 4 == 0`, `assert 4 == 1`, `assert 3 == 1`).

## Verification

- `mypy strands_robots tests tests_integ`: 14 errors, all pre-existing in `examples/isaac_gs/`, **0** naming a changed file — identical on `main`.
- Blast radius (114 files: everything referencing the normalizer or `training/rl`, plus the repo-wide docs/docstring/changelog guards), same selection both trees: branch **6 failed / 4417 passed / 53 skipped**, `main` **14 / 4409 / 53**. Branch-only failures: **none**. Base-only: exactly the 8 pre-fix. `14 − 6 = 8 = 4417 − 4409`, same collected total. The 6 shared failures are pre-existing and unrelated (4 are a `tests_integ` so101 joint-naming mismatch, 1 a `pytest-timeout` under load, 1 a pytest-exit meta-test).
- The PPO / FastSAC convergence tests that exercise the normalizer end-to-end pass unchanged; every in-package caller already passes `(N, D)`.

## Artifact

The same integration script run against both revisions — only `strands_robots` differs, and the script reads the remedy out of whatever message it gets back and follows it.

shape contract

The rollout the numbers come from (SO-101 driven through `GymSimEnv`, MuJoCo headless): https://raw.githubusercontent.com/cagataycali/robots/media-rl-normalizer-shape-contract/so101-gymsimenv-rollout.mp4

## Scope

This refuses an ambiguous shape rather than picking a convenience semantic for it. Auto-batching a `(D,)` input would be a new contract — and a guess, since the current code reads that same shape as D samples of a one-feature stream — so the caller states the intent with `unsqueeze(0)`.


yinsong1986 (approved):
## Summary

Adds a shape guard to `EmpiricalNormalization` at both entry points (`forward` and `update`) that refuses any input not shaped `(batch, *shape)`, closing the two silent-broadcast cases (`(D,)` read as D one-feature samples, `(batch, 1)` pooled into every feature) that corrupted the persistent running-statistics buffers, and moving the check ahead of the `count` increment so a refused batch no longer skews the Welford update rate. The refusal is a loud `ValueError` that names both shapes and the `x.unsqueeze(0)` remedy, which satisfies the raise-on-fatal / no-silent-defaults conventions and makes the previously mixed failure modes (silence, `RuntimeError`, `IndexError`) uniform. In-package callers (PPO, FastSAC) already pass `(N, D)` batches, so the behaviour change only bites the caller the PR identifies as the reachable case (`GymSimEnv`'s flat observation), and it bites with a stated migration.

## What's good

- Guard-before-`count` ordering is pinned by exact count assertions, and the remedy string is verified by parsing it out of the message and executing it - the tests grade behaviour through the public surface.
- `_feature_shape` read off `_mean` rather than stored at construction keeps one owner of the value across `load_state_dict`, and the rule stays module-local per AGENTS.md convention #11 (one caller, torch-specific domain).
- Changelog fragment as its own file, docs table updated in the same PR, accepted-domain tests (empty batch, multi-dim feature, one-feature stream) pin the boundary so the guard can't over-refuse silently.

cagataycali:
## Closing: superseded by #2429, which landed the same rule and a strictly stronger count fix

This PR and #2429 are two implementations of one contract. #2429 merged as `ce3b6f1f`. Rather than rebase this branch onto it, I graded the two against each other, because a rebase here would not have produced a smaller diff -- it would have produced a second name for a rule that is already on `main`, and reverted part of it.

### The behaviour is already on `main`

This branch's own test file, copied verbatim onto `main` at `ce3b6f1f` and run unmodified:

```
3 failed, 17 passed
```

All three failures are `re.search` / `pytest.raises(match=...)` assertions on the exact sentence:

| test | why it fails |
|---|---|
| `test_an_unbatched_observation_is_refused_rather_than_broadcast[forward]` | regex `batch of shape \(batch, 4\)` |
| `test_an_unbatched_observation_is_refused_rather_than_broadcast[update]` | same regex |
| `test_the_refusal_names_the_call_that_batches_a_single_observation` | regex `normalized as (x\.unsqueeze\(0\))\.` |

Same input, same `ValueError`, same remedy named -- `main` says `add the leading batch dimension (x.unsqueeze(0) for a single observation)` where this branch says `A single unbatched observation is normalized as x.unsqueeze(0).` Nothing behavioural separates them. The other 17, including the `(3, 1)` broadcast case, the eval-mode `forward` path, the count-preservation parametrization and the accepted-domain controls, pass on `main` untouched.

### Composing this source onto `main` loses a fix

This branch's `normalization.py` against the merged test file (`main` tests, this branch's source):

```
12 failed, 425 passed
```

11 of the 12 are the same wording regexes in the other direction. Exactly one is behavioural:

```
FAILED test_a_fold_that_fails_on_dtype_does_not_advance_the_sample_count - assert 8 == 4
```

This branch guards the count by checking shape *before* `count` moves. That covers shape errors only. `main` guards it by committing `count` *after* the fold, which also covers the batch that has an acceptable shape and still raises inside `torch.var_mean` -- an integer or boolean observation tensor. This diff restores the pre-fix ordering:

```diff
+        self.count += count_x
+        rate = count_x / float(self.count)
         var_x, mean_x = torch.var_mean(x, dim=0, unbiased=False, keepdim=True)
-        rate = count_x / float(int(self.count) + count_x)
-        self.count += count_x
```

So the leaked count returns for the dtype case: 4 expected, 8 observed.

### And it silently narrows a docs guard

Removing the `Args:` block from `update` does not fail `tests/test_args_docstring_completeness.py` -- it removes `update` from the population that guard grades, which is worse than failing it:

| tree | collected in the args guard | items naming the normalizer |
|---|---|---|
| `main` at `ce3b6f1f` | 414 | 4 |
| this branch's source | 412 | 2 |

The `docs/training/rl.md` row is the same shape of loss: it drops the sentence recording that a refused batch leaves the estimator untouched, which is now a property `main` actually has.

### One thing here is genuinely additive

`test_a_correctly_batched_input_is_accepted_unchanged[multi-dim-feature]` builds `EmpiricalNormalization((2, 3))` and folds a `(5, 2, 3)` batch. `main` constructs only scalar widths (`1`, `2`, `3`, `4`) in this file, so a multi-dimensional per-feature shape is unpinned there. It passes on `main` today -- `got[1:] == expected` is generic in the shape rule -- so this is an unpinned control rather than a gap in behaviour. Recording it here rather than spending a review round on a test that already passes; it is a self-contained addition if anyone wants it later.

No reviewer time was spent on this branch beyond the approval, and the approval was correct on its own terms -- the rule is right, it is just the second copy of it. Closing rather than rebasing, so `main` keeps one owner for the shape contract.

cagataycali:
Superseded by #2429, which landed the same fix with strictly better coverage. Closing this.

I ran this branch's test file unchanged against merged `main` as a coverage check: **17 of 20 pass**. The 3 failures are only my regexes matching my own phrasing — `#2429`'s message carries the same information and the same remedy (`add the leading batch dimension (x.unsqueeze(0) for a single observation)`).

`#2429` is also better on an axis this branch does not cover. It fixes the premature `count` advance by **committing the count after the fold** rather than by validating before it, so a batch with a valid shape but an unusable dtype is caught too:

```
EmpiricalNormalization(4), count=1
update(torch.zeros(2, 4, dtype=torch.long))  -> RuntimeError from var_mean
count 1 -> 1   (preserved)
```

A shape guard alone cannot see that case, because the shape is correct. Its test suite is a superset here as well — it covers the `(3, 1)` too-few-features broadcast, the frozen eval path, the remedy round-trip and the dtype case, plus a Welford two-fold/union correctness test this branch does not have.

Nothing left to salvage. I have added the reachability evidence to #2429 for the record.

PR #2432, opened 2026-08-18
fix(simulation/isaac): measure the preview cadence and reported recording duration on a clock that cannot step

cagataycali:
## Problem

Two durations in the Isaac backend were measured by subtracting two readings of `time.time()`, which is not a clock but the current opinion about the date. An NTP correction, a `date -s` or a resume from suspend moves it by an arbitrary amount, so a step landing between the readings changed the answer, and nothing raised.

**1. The idle-render gate.** `run_pump_forever` decides once per iteration whether to refresh the live preview:

```python
now = time.time()
do_render = (now - last_idle_render) >= self._idle_render_period
self.pump(render=do_render)
if do_render:
    last_idle_render = now
```

Driving the real loop at a 0.2 s idle period through a clock double that steps once at iteration 12, with the loop's own `sleep(0.05)` setting the granularity:

| clock event | preview refreshes | gaps (s) | last refresh | max staleness |
|---|---|---|---|---|
| no step (control) | 10 | 0.20 | t = 1.80 | 0.15 s |
| wall clock steps +30s | 11 | 0.15, 0.20 | t = 1.95 | 0.15 s |
| wall clock steps -2s | **3** | 0.20 | **t = 0.40** | **1.55 s** |
| wall clock steps +1h | 11 | 0.15, 0.20 | t = 1.95 | 0.15 s |

Backward is the damaging direction: the preview stopped refreshing for the size of the step, 1.55 s of that 1.95 s run, while `pump()` kept draining the app. The viewport sat frozen on a stale frame with nothing reporting it. Forward fired the gate once early and then re-based onto the stepped clock, so it cost one spurious refresh rather than a run of them.

**2. The duration a camera recording reports.** `start_cameras_recording` stamped `"started_at": _time.time()` and `stop_cameras_recording` did `elapsed = _time.time() - state["started_at"]`, so a step anywhere in the recording, not only in some window, moved the answer. The same 3.0 s capture:

| clock event | reported |
|---|---|
| no step (control) | `Stopped 'cap' after 3.0s` |
| wall clock steps +30s | `Stopped 'cap' after 33.0s` |
| wall clock steps -2s | `Stopped 'cap' after 1.0s` |
| wall clock steps +1h | `Stopped 'cap' after 3603.0s` |

The buffers carry no per-frame timestamp, so that line is the only thing that says how much wall time the frames span.

## Why this scope

The MuJoCo backend's two recorders already hold this base on `time.monotonic()` under the name `started_mono`, and `tests/simulation/test_rendering_pacers_survive_a_clock_step.py` pins that they carry no `started_at` key, with the reason spelled out:

> a duration base named as a wall-clock stamp invites the next reader to subtract `time.time()` from it

Which is exactly what this backend's recorder did. It is a third recorder behind the same `start_cameras_recording` / `stop_cameras_recording` pair, so it holds the same contract. Both sites are durations in one module, which is the scope the equivalent MuJoCo change took: one pacer plus two `started_at` bases and their elapsed reports.

The rule is AGENTS.md, "Clocks: a duration is measured, a stamp is recorded", which names a frame pacer explicitly. This is the same boundary already settled for the agent-callable tools, the hardware control loops, the mesh, the sim rollouts and the rendering pacers.

## Change

`strands_robots/simulation/isaac/simulation.py`, 5 lines of logic:

- The gate reads `time.monotonic()` and its base is `last_idle_render_mono`.
- The recording state's base is `started_mono`, read back with `time.monotonic()`.
- The gate's "never refreshed" sentinel becomes `None` rather than `0.0`. `time.monotonic()`'s reference point is unspecified (seconds since boot on Linux), so a numeric sentinel would have suppressed the first preview refresh for the first `_idle_render_period` seconds of uptime; `0.0` only read as "long ago" because a wall clock's epoch is large. The `job()` path sets it to `None` for the same reason it set `0.0`: refresh on the next idle pass.

No signature, return shape or public name changes. Nothing outside this module read `started_at`.

## Tests

`tests/simulation/isaac/test_isaac_durations_survive_a_clock_step.py`, **8 failed / 4 passed on `main`, 12 pass after**. Every assertion is on the achieved refresh timeline or the reported duration rather than on which clock is called, so none can be satisfied by renaming a call. The clock double counts reads of either clock, so the step lands at the same loop position whichever clock the loop consults.

Fails before, passes after:

- `test_idle_refresh_cadence_survives_a_wall_clock_step[forward_30s|backward_2s|forward_1h]` - asserted against the unstepped timeline, so the contract is "a step changes nothing".
- `test_reported_recording_duration_survives_a_wall_clock_step[forward_30s|backward_2s|forward_1h]`
- `test_the_first_idle_iteration_refreshes_whatever_the_clock_epoch_reads` - the sentinel.
- `test_the_recorder_names_the_clock_its_duration_base_holds` - extends the MuJoCo naming contract to this recorder.

Passes on both trees, and each one fails for a different wrong fix:

- `test_idle_refresh_cadence_without_a_step_is_the_period` - fails if the gate is made unconditional. It is what makes the stepped cases a statement about the step rather than about the loop's granularity.
- `test_a_busy_pump_never_refreshes_the_preview` - the branch that consults no clock at all; asserts `clock.reads == 0`.
- `test_a_main_job_forces_the_next_idle_iteration_to_refresh` - fails if the job path's reset is dropped, which would leave the preview on the pre-job frame for up to one period after a record or plan finished.
- `test_reported_recording_duration_without_a_step_is_the_elapsed_time`

No Isaac Sim Kit runtime, GL context or MP4 encoder is touched: the engine is the skeleton `__new__` fixture shape `test_cameras_recording_preflight_guards.py` already uses, and `pump` is faked on the instance.

## Verification

At `df7a33ae` against `upstream/main` `b107aecf`, identical 112-file selection on both trees (all of `tests/simulation/isaac`, every test naming a recording-state key, `run_pump_forever` or AGENTS.md, the rendering-pacer contract, and the repo-wide docstring / changelog / string guards), 158.9 s each:

|---|---|---|---|
| this branch | 1 | 4216 | 29 |
| `upstream/main` + this test file | 9 | 4208 | 29 |

`4216 + 1 == 4208 + 9 == 4217` collected on both. Branch-only failures: **none**. Base-only: exactly the 8 pre-fix failures listed above. The one shared failure is pre-existing and unrelated (`tests/mesh/test_safety_timing_and_replay_defenses.py::TestResumeLockoutTimingOracleClosed::test_correct_override_code_clears_lockout`, `AttributeError: 'Mesh' object has no attribute '_safety_publishers_lock'`).

`ruff check` and `ruff format --check` over `strands_robots tests tests_integ` clean. `mypy` over the same: 19 errors, the same set on both trees, none in the changed files.

## What it looks like

Isaac preview cadence and reported recording duration across a wall-clock step

One capture script per tree, driving the real `run_pump_forever` and the real `start_cameras_recording` / `stop_cameras_recording` pair. The clock double, the faked `pump` and the stimulus are identical between the two runs; only `strands_robots` differs. Isaac Sim's renderer is not exercised, so the panels measure when the backend *decides* to refresh the preview and what it reports, which is what the change moves.

**A**: how stale the frame on the live preview is, through a run where the wall clock steps back 2 s. Before, it climbs to 1.55 s and stays there while the app keeps being pumped; after, it holds the 0.2 s sawtooth the gate promises.

**B**: the refresh ticks per clock event. Fixed, all four reproduce the control's timeline exactly, 10 refreshes 0.20 s apart.

**C**: the duration reported for the same 3.0 s capture, log scale.

## Scope

`run_pump_forever`'s idle gate and this recorder's duration report are the last two `time.time()` readings in the module; there are now none. Left alone deliberately: absolute stamps elsewhere in the tree that name a point in time another reader correlates, which is the other half of the AGENTS.md rule.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1330,7 +1330,13 @@ Corrections from code review that apply to all future contributions:
   rendering pacers - the MJPEG stream generator and the multi-camera recorder thread - are
   pinned on the same achieved-interval basis by
   `tests/simulation/test_rendering_pacers_survive_a_clock_step.py`, along with the duration
-  each recording reports.
+  each recording reports. The Isaac backend's own two - the idle gate that decides when
+  `run_pump_forever` refreshes the live preview, and the duration its camera recording
+  reports - are pinned by
+  `tests/simulation/isaac/test_isaac_durations_survive_a_clock_step.py`, which asserts the
+  achieved refresh timeline against the unstepped one rather than a tolerance. A duration
+  base also carries its clock in its name (`started_mono`, `last_idle_render_mono`), so a
+  later reader cannot mistake it for a stamp and subtract `time.time()` from it.
 
 ### Module-Level Side Effects
 - **If you must run code at import time, comment WHY it can't be lazy.** `MUJOCO_GL` is the canonical example: MuJoCo locks the GL backend at first `import mujoco`, so the env var must be set before any downstream import chain triggers it.
--- a/changelog.d/2432-isaac-durations-monotonic-clock.md
+++ b/changelog.d/2432-isaac-durations-monotonic-clock.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **simulation/isaac**: measure the live preview's refresh cadence and the duration a camera
+  recording reports on `time.monotonic()` instead of `time.time()`. Both subtracted two
+  readings of a wall clock, so a step landing between them changed the answer: a backward
+  step of 2 s stopped `run_pump_forever` refreshing the preview for 1.55 s of a 1.95 s run
+  while it kept draining the app, and a 3.0 s capture reported `after 33.0s`, `after 1.0s`
+  or `after 3603.0s`. The recording state's duration base is now spelled `started_mono`, the
+  name the MuJoCo backend's two recorders already carry.
--- a/strands_robots/simulation/isaac/simulation.py
+++ b/strands_robots/simulation/isaac/simulation.py
@@ -4966,7 +4966,10 @@ class IsaacSimulation(IsaacMotionPrimitivesMixin, IsaacRecordingMixin, SimEngine
                 "paths": paths,
                 "errors": dict.fromkeys(names, 0),
                 "output_dir": out_dir,
-                "started_at": _time.time(),
+                # A ``time.monotonic()`` reading: the only thing derived from
+                # this base is how long the recording has been running, so it
+                # is a duration base and carries its clock in its name.
+                "started_mono": _time.monotonic(),
                 "max_frames": max_frames_per_camera,
             }
             self._cams_rec_state = state
@@ -5050,7 +5053,7 @@ class IsaacSimulation(IsaacMotionPrimitivesMixin, IsaacRecordingMixin, SimEngine
 
         from strands_robots.rendering.video import encode_clip
 
-        elapsed = _time.time() - state["started_at"]
+        elapsed = _time.monotonic() - state["started_mono"]
         lines = [
             f"Stopped '{state['name']}' after {elapsed:.1f}s",
             f"   output_dir: {state['output_dir']}",
@@ -5745,7 +5748,12 @@ class IsaacSimulation(IsaacMotionPrimitivesMixin, IsaacRecordingMixin, SimEngine
         ``is_set()`` returning truthy ends the loop. ``None`` (default)
         loops until ``KeyboardInterrupt``.
         """
-        last_idle_render = 0.0
+        # ``None`` until the first idle refresh, rather than a sentinel that
+        # only reads as "long ago" because a wall clock's epoch is large: on
+        # ``time.monotonic()`` the epoch is unspecified (seconds since boot on
+        # Linux), so a numeric sentinel would suppress the first preview refresh
+        # for the first ``_idle_render_period`` seconds of uptime.
+        last_idle_render_mono: float | None = None
         self._pump_running = True
         try:
             while stop_event is None or not stop_event.is_set():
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Swaps the two remaining `time.time()` duration measurements in the Isaac backend onto `time.monotonic()`: the idle-render gate in `run_pump_forever` (base renamed `last_idle_render_mono`) and the camera-recording duration report (state key `started_at` -> `started_mono`, matching the spelling the MuJoCo recorders already carry on `main`). The logic change is five lines; the sentinel change from `0.0` to `None` is handled correctly - a numeric sentinel on a monotonic clock would have suppressed the first preview refresh for one period of uptime, and the `job()` reset path was updated in the same motion. I verified at the head SHA that nothing outside this module reads `started_at` (the one fixture in `test_cameras_recording_preflight_guards.py` is updated in this diff), that `_cams_rec_state` is transient in-process state rather than a persisted or wire schema, and that no signature or public name moves - so there is no one-way door here. This is squarely the AGENTS.md "Clocks: a duration is measured, a stamp is recorded" rule, applied at the same scope the equivalent MuJoCo change took.

## What's good

- The new test file asserts on the achieved refresh timeline and the reported duration rather than on which clock is called, so a rename cannot satisfy it; the busy-branch test (`clock.reads == 0`) and the unconditional-gate control pin the wrong fixes explicitly.
- The sentinel epoch hazard (`time.monotonic()` reading small early in uptime) is both fixed and pinned by its own test, not just mentioned in a comment.
- Changelog fragment present as `changelog.d/2432-isaac-durations-monotonic-clock.md` per step 3 of the PR workflow; no direct `CHANGELOG.md` edit.
- Test file is ASCII-clean with no host paths, and each duration base carries its clock in its name per the convention this PR also documents in AGENTS.md.

PR #2433, opened 2026-08-18
fix(policies/protomotions): settle a motion cache's frame count instead of trusting a declared num_frames

cagataycali:
## Problem

A `MotionPlayer` cache states its own length seven times over. The documented format gives every state channel a leading frame axis - `dof_pos` is `[num_frames, num_dofs]`, `body_rot` is `[num_frames, num_bodies, 4]`, and so on - alongside a scalar `num_frames`. Nothing made those statements agree.

The count is load-bearing rather than informational. `get_state_at_frame` clamps a query into `[0, num_frames - 1]` so an out-of-range frame is safe, and `get_future_references` - the call the ONNX tracker makes every control tick - asks for frames *ahead* of the playhead. That clamp is therefore the only thing standing between the future window and the end of the arrays, and it was calibrated against a number the channels need not support.

Both failure directions are reachable by editing a cache, which is a documented input mode (`MotionPlayer` accepts a hand-built dict, and an `.npz` that "gets downloaded, shared between machines and committed to dataset repos").

| cache | before | after |
|---|---|---|
| channels trimmed to 100, `num_frames` left at 120 | loads; claims 120; serves **92** then `IndexError: index 100 is out of bounds for axis 0 with size 100` | refused on load, both counts named |
| two 60-frame clips concatenated, `num_frames` left at 60 | loads; **60 frames silently unreachable**, nothing raised | refused on load |
| `save_cache_npz` on the above | writes `num_frames=60` beside 120 rows, so the truncation becomes durable | cannot emit a count its channels cannot serve |
| channels disagreeing with each other (`body_rot` short) | `IndexError` from that one channel | refused, naming `body_rot` and both counts |
| consistent cache, or `num_frames` omitted | works | unchanged |

The trimmed case is the sharpest: playback dies a **whole future window early** (at 92, not 100), mid-rollout, with a bare `IndexError` naming neither the cache nor the mismatch.

## Change

`_load_cache` already treated the channels' row count as the authority when `num_frames` is absent - it fell back to `dof_pos.shape[0]`. So the module already knew where the truth lives; it just still preferred a declared number that could contradict it. The number is now settled once, in `_cache_frame_count`, and a statement the channels cannot serve is refused with both counts and a remedy named:

```
MotionPlayer cache declares num_frames=199 but its channels carry 100 frames. Every
channel is documented as [num_frames, ...], and a frame index is clamped to
num_frames - 1, so a declared count the channels cannot serve reads out of range
before the clip ends while a smaller one hides the tail. Drop num_frames to take the
row count, or resize the channels to match it.
```

Channels that disagree with each other are refused too - the count is one number, not `dof_pos`'s number. `qpos_to_motion_data` sets the two consistently, so a consistent cache and one that omits `num_frames` are both unaffected; `ValueError` joins the `KeyError` `_load_cache` already raises for missing keys.

Scope: this settles the **frame axis** only. Channel widths (`body_rot`'s trailing 4, joint counts) are a separate contract and are left alone.

## Rollout

A real Kimodo clip - "a person throwing boxing punches", 120 frames at 30 Hz, resampled by `qpos_to_motion_data` to 199 at 50 Hz - with the six channels then trimmed to 100 and `num_frames` left at 199. One capture script, run against each tree; it reads the refusal from whichever tree it is on and applies the remedy that message gives, so nothing is hardcoded per side. Reference poses are replayed kinematically into MuJoCo (headless EGL).

MotionPlayer replays a hand-trimmed reference-motion cache

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-motionplayer-frame-count/motionplayer-frame-count.mp4) - [final frame](https://raw.githubusercontent.com/cagataycali/robots/media-motionplayer-frame-count/motionplayer-frame-count.png)

Before: the cache is accepted, playback claims 199 frames, stops at 92. After: refused on load, `del cache['num_frames']` as the message says, 100/100 frames served. The two panels are **pixel-identical (mean abs delta 0.0000) at frame 0 and still at frame 91** and 12.57 apart at the end - the same rollout until the change bites. Both panels animate (per-frame mean abs delta 1.53-7.29); the 8 static transitions in the left panel are the frozen tail after playback died.

## Verification

Regression test `tests/policies/protomotions/test_cache_frame_count_agreement.py`, public surface only so it collects on `main`: **6 failed / 7 passed** against pre-fix code, 13 passed after. The headline failure carries the measurement rather than a wording diff:

```
AssertionError: the player claims 120 frames but the future window overruns the
channels after 92
```

The 7 that pass on both trees are the boundary: a consistent cache's count, its full playback through the tracker's window, its `control_dt`, its channels bound verbatim, and the `.npz` round-trip. One of them - "the player never claims a frame it cannot serve" - is written as the consequence rather than as the mechanism, so it holds whether the cache is refused or simply consistent.

Local gate on the 89 test files touching `MotionPlayer` / `num_frames` / `motion_utils`, all of `tests/policies/protomotions`, and the repo-wide docstring, ASCII, dependency and changelog guards:

- branch **2891 passed, 52 skipped, 0 failed**
- `main` with the same file copied in: **2885 passed, 52 skipped, 6 failed**, all 6 in the new file, zero foreign
- `2885 + 6 == 2891`, same collected total both trees, no branch-only failure
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 19 pre-existing errors on both trees, none in the changed files

## Docs

The module docstring now states the agreement next to the cache-format table it belongs to, and `docs/policies/protomotions.md` gains the hand-edit case with the two-line remedy, since editing a cache is where both directions come from.


yinsong1986 (approved):
## Summary

This PR closes a real load-bearing gap in `MotionPlayer`: the six state channels and the scalar `num_frames` all state the clip's length, `get_state_at_frame` clamps against the declared count, and `get_future_references` reads ahead of the playhead — so a declared count the channels cannot serve overruns the arrays a whole future window early with a bare `IndexError`, and a count below the row count silently hides the tail (and `save_cache_npz` then makes the wrong count durable). The fix settles the count once in `_cache_frame_count`, refusing a contradiction on load with both numbers and a working remedy named, while leaving a consistent cache and the omitted-`num_frames` fallback byte-for-byte unchanged. I verified the new test file (13 tests) passes against the branch's `motion_utils.py` composed with the current tree, and the full `tests/policies/protomotions` suite stays green; both in-tree cache producers (`qpos_to_motion_data` and `_resample_raw`) construct the count from the rows, so only hand-edited caches — the documented failure source — are affected, and those now get an explicit `ValueError` rather than a mid-rollout crash, which is the sanctioned loud-rejection pattern rather than a silent behaviour change.

## What's good

- The refusal message names both counts and a remedy, and `test_the_refusal_offers_a_remedy_that_works` actually executes the remedy the message gives — the error text cannot drift into a dead end unnoticed.
- `test_the_player_never_claims_a_frame_it_cannot_serve` pins the consequence rather than the mechanism, so it survives future refactors of *how* the agreement is enforced.
- Scope discipline: frame axis only, channel widths explicitly left to a separate contract; `ValueError` joins the existing `KeyError` without disturbing the `.pt` resample path.
- Changelog fragment under `changelog.d/`, ASCII-clean strings, full `Args:`/`Returns:`/`Raises:` docstring on the new helper, docs updated next to the cache-format table where the contract lives.

PR #2434, opened 2026-08-18
fix(rtps): refuse a ROS 2 name this mangling cannot map

cagataycali:
## Problem

`strands_robots.rtps.mangling` maps ROS 2 graph names onto DDS names. Two of its
three functions returned names that cannot be what they claim to be.

### 1. `dds_type_name` returned a struct ROS 2 never generates

`_ROS_TYPE_RE` enumerated `msg|srv|action`, so a service or action was accepted:

```
dds_type_name("example_interfaces/srv/AddTwoInts")     -> 'example_interfaces::srv::dds_::AddTwoInts_'
dds_type_name("std_srvs/srv/Trigger")                  -> 'std_srvs::srv::dds_::Trigger_'
dds_type_name("control_msgs/action/FollowJointTrajectory")
                                                       -> 'control_msgs::action::dds_::FollowJointTrajectory_'
```

ROS 2 has no single DDS type for a service. `rosidl_generator_dds_idl`'s
`srv.idl.em` renders `msg.idl.em` once per constituent message, and `msg.idl.em`
emits `struct <Name>_` inside `module dds_` - so a service yields
`AddTwoInts_Request_` and `AddTwoInts_Response_` (plus the event message) and
never an `AddTwoInts_`. `rmw` agrees: `rmw_client.cpp` calls
`_create_type_name(request_members)` and `_create_type_name(response_members)`
against two separate *message* type supports, and puts them on the `rq`/`rr`
prefixes rather than `rt`. An action expands further, into
goal/result/feedback plus two nested services.

The consequence is a participant that misrepresents itself and nothing that says
so. Advertising the returned name and reading back DDS discovery
(`DCPSPublication.type_name`) on one topic:

```
topic  rq/add_two_intsRequest
advertised  example_interfaces::srv::dds_::AddTwoInts_Request_   <- what ROS 2 generates
advertised  example_interfaces::srv::dds_::AddTwoInts_           <- what this returned
```

Matching is by topic name, so the disagreement does not even surface as a
failure to connect: a reader created on the returned name still received all
three samples a writer published under the real type. So there is no error, no
log line, and no failed match - only a name that disagrees with every real ROS 2
participant's advertisement, and with any stack that does enforce type
consistency.

Three places already state the intended scope, and the pattern was the one thing
that did not:

- `strands_robots/rtps/__init__.py`: "Scope (v1): topics only ... Services and
  actions need the ROS 2 request/reply-over-DDS protocol and are deliberately
  deferred to a focused follow-up."
- `mangling.py`'s own docstring: "services use `rq`/`rr`, out of scope here".
- `docs/rtps-integration.md`: "validated against an allowlist before mangling
  (absolute `/`-rooted alnum/`_` names; **`pkg/msg/Name` types**)".

The refusal for a malformed name compounded it by advertising the two spellings
that are also refused: `expected pkg/msg/Name (or pkg/srv/Name, pkg/action/Name)`.

### 2. `ros_topic_name` handed back names `dds_topic_name` refuses

It is documented as "Inverse of `dds_topic_name`", and stripped the prefix
without checking what remained. A DDS graph carries topics no ROS 2 node
published, so a subscriber enumerating one gets back a "ROS 2 topic name" this
module itself rejects:

| `ros_topic_name(...)` | returned | `dds_topic_name` on that |
|---|---|---|
| `'rt/'` | `'/'` | refused |
| `'rt/bad name'` | `'/bad name'` | refused |
| `'rt/a/'` | `'/a/'` | refused |
| `'rt/x;y'` | `'/x;y'` | refused |

4 of 6 malformed names the module's own topic test already pins came back
through the inverse, so the failure landed at whatever the caller passed the name
to next rather than at the name it came from.

## Change

One property: every name this module returns is a name it would itself accept,
and a ROS 2 name it cannot map is refused rather than mangled.

- `_ROS_TYPE_RE` is message-only. A well-formed `srv`/`action` is matched
  separately and refused with the types ROS 2 does generate quoted, so the
  refusal is not itself a dead end - and it is not reported as "malformed",
  because it is a valid ROS 2 type this mangling simply cannot name:
  ```
  ROS 2 srv interface 'example_interfaces/srv/AddTwoInts' has no single DDS type: ROS 2
  generates one type per constituent message (example_interfaces::srv::dds_::AddTwoInts_Request_
  and example_interfaces::srv::dds_::AddTwoInts_Response_), exchanged over the rq/rr topic
  prefixes rather than rt. This package maps topics only, so pass a pkg/msg/Name type.
  ```
- The malformed-name refusal drops the `pkg/srv/Name` / `pkg/action/Name`
  suggestion it also rejects. An unrecognised kind (`pkg/badkind/Name`) still
  reports as malformed.
- `_require_ros_topic` is the one topic rule, applied by both directions, so
  `dds_topic_name` and `ros_topic_name` cannot come to disagree about what a
  valid ROS 2 topic name is. `ros_topic_name` names the DDS topic the rejected
  name was recovered from:
  ```
  invalid ROS 2 topic '/' recovered from 'rt/': expected an absolute name like /turtle1/cmd_vel
  ```

Message mangling is untouched. No production caller is affected: `use_rtps` and
the RTPS hardware bridge use `dds_topic_name` and resolve types through
`rtps.idl.get_type`, which is message-only by construction; `dds_type_name` has
no in-package caller.

## Tests

`tests/rtps/test_mangling.py` extended in place. Copying only that file onto
`e9240dc5` runs the new cases against the unfixed source: **8 failed, 21 passed**
-> 29 passed. Every failure carries the measured value, e.g.

```
dds_type_name('std_srvs/srv/Trigger') returned 'std_srvs::srv::dds_::Trigger_', but ROS 2
generates no such struct: the wire types are std_srvs::srv::dds_::Trigger_Request_ /
Trigger_Response_. ...

ros_topic_name('rt/bad name') handed back '/bad name', a name dds_topic_name itself refuses
(invalid ROS 2 topic '/bad name': ...), so the documented inverse does not hold
```

The 21 that pass on both trees bound the change:

- Every type in the bundled IDL registry still mangles to the name **cyclonedds
  itself** derives from the dataclass (`__idl_typename__`) - 9/9 locally. This
  grades the message path against the binding that carries it rather than a
  copied table. Skipped without the `[ros2]` extra.
- A valid topic still round-trips through both directions, so checking the
  recovered name does not narrow what a valid graph can carry.
- `pkg/badkind/Name` still reports as malformed, so the new branch is narrow.
- The pre-existing happy-path and rejection cases are unchanged; the malformed
  topic set is now a shared constant used by both directions.

The malformed-topic list is reused rather than restated, so the two directions
cannot drift.

## Verification

At `611ad465` against `e9240dc5`, on the 96 test files that reference these
symbols, the `rtps`/`mesh` suites, the pages that read the docs, and the
repo-wide docstring/ASCII/changelog guards:

|---|---|---|---|
| branch | 3171 | 0 | 3223 |
| base (+ the new test file) | 3163 | 8 | 3223 |

`3163 + 8 == 3171` and both collect 3223, so the only verdicts that moved are the
8 new cases; no branch-only failure, and no pre-existing failure in the
selection. `ruff check` / `ruff format --check` clean over
`strands_robots tests tests_integ`; `mypy` 19 pre-existing errors on both trees,
none in the changed files.

No visual artifact: this is a pure string transform with no policy, simulation,
rendering, recording or asset behaviour to show. The DDS discovery readback and
the receive counts above are the measured evidence, taken with real cyclonedds
participants on one domain.

### DIFF
--- a/changelog.d/2434-rtps-mangling-refuses-unmappable-names.md
+++ b/changelog.d/2434-rtps-mangling-refuses-unmappable-names.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **rtps**: `dds_type_name` now maps message interfaces only. ROS 2 generates one DDS type per constituent message of a service or an action, so there is no single type to return for one; a `pkg/srv/Name` or `pkg/action/Name` is refused with the types ROS 2 does generate quoted (`pkg::srv::dds_::Name_Request_` / `..._Response_`) instead of returning a `pkg::srv::dds_::Name_` the participant would then advertise for a struct that exists nowhere in the ROS 2 type system. `ros_topic_name` checks the name it recovers against the same rule `dds_topic_name` applies, so the documented inverse no longer hands back a ROS 2 topic name the module itself refuses.
--- a/docs/rtps-integration.md
+++ b/docs/rtps-integration.md
@@ -60,6 +60,15 @@ becomes `rt/turtle1/cmd_vel`, and a type `geometry_msgs/msg/Twist` becomes
 `geometry_msgs::msg::dds_::Twist_` - the conventions that make a bare DDS
 participant interoperable with real ROS 2 nodes.
 
+Message interfaces only. ROS 2 has no single DDS type for a service or an
+action: `rosidl` generates one type per constituent message, so
+`example_interfaces/srv/AddTwoInts` becomes
+`example_interfaces::srv::dds_::AddTwoInts_Request_` and
+`..._Response_` on the `rq`/`rr` prefixes rather than `rt`. A `pkg/srv/Name` or
+`pkg/action/Name` type is therefore refused, with the types ROS 2 does generate
+quoted - an invented `pkg::srv::dds_::AddTwoInts_` would match no participant,
+and DDS reports a type mismatch as silence rather than as an error.
+
 ## Examples
 
 ```python
--- a/strands_robots/rtps/mangling.py
+++ b/strands_robots/rtps/mangling.py
@@ -23,6 +23,19 @@ A ROS 2 type ``geometry_msgs/msg/Twist`` becomes the DDS type
 * the final segment gains a trailing underscore (``Twist`` -> ``Twist_``),
 * a ``dds_`` segment is inserted before the final segment.
 
+Message interfaces only. ROS 2 has no single DDS type for a service or an
+action: ``rosidl`` renders the message template once per constituent message, so
+``pkg/srv/Name`` becomes ``pkg::srv::dds_::Name_Request_`` and
+``pkg::srv::dds_::Name_Response_`` (an action adds goal/result/feedback and two
+nested services), exchanged over the ``rq``/``rr`` prefixes. There is no name
+this mangling could return for one, so it refuses them and says which types
+ROS 2 does generate. Returning ``pkg::srv::dds_::Name_`` instead would make the
+participant advertise, in DDS discovery, a type name ROS 2 generates no struct
+for, and nothing reports that: matching is by topic name, so a reader created
+with the wrong name can still receive, leaving a participant that misrepresents
+itself to every discovery tool and to any stack that does enforce type
+consistency.
+
 Both directions are pure string transforms with no ROS or DDS import, so they
 are trivially unit-testable with no middleware present.
 """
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Narrows `dds_type_name` to message interfaces only and makes `ros_topic_name` validate the name it recovers against the same rule `dds_topic_name` applies. Both changes replace silently-wrong outputs with loud `ValueError`s: the old `srv`/`action` path minted a DDS type name (`pkg::srv::dds_::Name_`) that ROS 2 generates no struct for — a participant advertising it misrepresents itself in discovery with nothing reporting the disagreement — and the old inverse handed back "ROS 2 topic names" the module itself refuses. The public-API behavior change is a correctness fix aligned with the package's already-documented topics-only scope, and no production caller is affected: `use_rtps` and the RTPS bridge use `dds_topic_name` plus `rtps.idl.get_type`, which is message-only by construction. Verified at `611ad465`: full `tests/rtps/` plus `tests/tools/test_use_rtps.py` pass (64 passed, 4 cyclonedds-gated skips), and the new refusals reproduce exactly as described, including the `recovered from 'rt/...'` attribution.

## What's good

- The srv/action refusal quotes the types ROS 2 does generate, so it is not a dead end, and the malformed-name message no longer suggests spellings that are also refused.
- `_require_ros_topic` is the single topic rule shared by both directions, and the malformed-topic test set is a shared constant, so the two directions cannot drift.
- The cyclonedds-gated registry test grades mangling against `__idl_typename__` — the binding that actually puts the name on the wire — rather than a copied table.
- Changelog fragment, per-parameter `Args:` entries, and ASCII-only messages all follow AGENTS.md conventions.
- The new test-file import of `strands_robots.rtps.idl` is safe without the `[ros2]` extra (REGISTRY stays empty; no import-time raise).

PR #2435, opened 2026-08-18
fix(simulation/mujoco): write a joint pose only where ctrl is a joint pose

cagataycali:
## Problem

`move_to`, `rotate_wrist` and `set_gripper` all drive a joint the same way: they write a joint coordinate into `data.ctrl` and step. That is the joint's target only on a position servo. A `<velocity kv>` drive reads the same number as a **rate**, a `<motor>` as a **torque** - so writing a joint angle there commands a different physical quantity that happens to be numerically equal to an angle. Nothing raises.

`scene_ops.joint_drive_map` already owns that split, and already guards the one other surface that writes a pose into `ctrl` (`set_joint_positions(hold=True)`) - its docstring states the rule outright: *"a joint angle written into a velocity drive's `ctrl` is commanded as a rate, which drives the joint away from the pose that was just written"*, and it names Menagerie's `pal_tiago` `tiago_velocity.xml` as a stock asset that reaches this path. The primitives resolved their actuators through `_joint_actuator_map` alone, which asks only whether the *transmission* is the joint - a test a `<velocity>` actuator passes.

Measured on that stock MJCF, `rotate_wrist(target_yaw=0.4)`:

| | before | after |
|---|---|---|
| status | `success`, `reached: True`, `final_yaw: 0.3815` | `error`, names `arm_velocity_7_servo` |
| `arm_7_joint` 200 steps later, nothing else commanded | **0.5413 rad** - still accelerating | 0.0000 rad (never commanded) |
| max drift of the 4 joints it promised to **hold** | **0.5033 rad** | 0.0001 rad |
| `torso_lift_joint` in the same call (the one position servo among them) | 0.0041 rad | 0.0041 rad |

The last two rows are one call: the servo channel held to 0.004 rad while the rate channels were driven up to 0.50 rad away from the pose being held. The convergence test was met by the joint sweeping *past* the set-point while accelerating, which is why it reported reached.

A wheel drive is the worst case, because a wheel angle accumulates - the "hold" grows with every turn the wheel has already made. On the scene in the new tests it diverges the integrator outright (`mj_step` reports `Nan, Inf or huge value in QACC` and the state resets mid-rollout, so the wrist never reaches either).

## Why not a blanket refusal

19 of the loadable registry robots have at least one joint-transmission actuator that is not a position servo, and they split two ways:

| | robots | outcome |
|---|---|---|
| the **targeted** drive is not a servo | `tiago_velocity`'s arm, `openarm` (2 servos beside 16 motors), `unitree_h1_2`, `jvrc`, the quadrupeds | the set-point cannot be commanded at all -> refuse, naming the actuator |
| only **other** joints are not servos | `lekiwi`, `stretch3`, `tiago_dual` - mobile manipulators whose arms are fully servo-driven, with `<velocity>` wheels | keep working; the wheel drives are left uncommanded and named |

Refusing on the second group would have regressed three stock mobile manipulators. Their `rotate_wrist` residuals are byte-identical before and after (`tiago_dual` 0.019787005230906984, `stretch3` 0.019465727775855313, `lekiwi` 0.018158874127248648); what changes is that the wheels are no longer commanded with a joint coordinate, and the payload now says so:

```
rotate_wrist: 'lekiwi' joint 'Wrist_Roll' reached 0.300 rad within 0.02 rad in 24 steps.
3 actuator(s) were left uncommanded because their ctrl is not a joint pose
(['base_back_wheel', 'base_left_wheel', 'base_right_wheel']); command those in
their own units with action='send_action'.
```

## Change

One shared `_pose_actuator_map` splits `_joint_actuator_map` through `joint_drive_map`, consulted at the three pose-write sites:

- **`rotate_wrist`** refuses when the resolved wrist joint's drive cannot take a pose; otherwise holds only the pose-writable joints and reports the rest in `uncommanded_drives`.
- **`move_to`** narrows `arm_jact` to the pose-writable half. No separate guard is needed for the solve: `arm_jact` *is* its `commanded_dofs` set, so narrowing it narrows the IK's decision space with it and keeps the residual a statement about the configuration `move_to` actually commands. On `tiago_velocity` it now reports the velocity arm joints among "the degrees of freedom move_to does not command" instead of writing rates into them.
- **`set_gripper`**'s driven-joint-range substitution gains a fourth "keeps refusing" shape. That substitution's premise is that `ctrl` is the joint target - exactly what `inheritrange="1"` would have compiled - so the drive, not just the transmission, has to supply it. Previously a velocity jaw reported `commanded open` while writing 0.04 m as 0.04 m/s, leaving the jaw 1% open and creeping.

**Deliberately out of scope**: a usable `ctrlrange` is authoritative and already in the drive's own units, so it is unchanged - only the *substitution* claims `ctrl` is a joint coordinate. Whether an open/close set-point is the right thing to ask of a rate command is a separate question about `set_gripper`'s vocabulary. A control test pins that boundary.

## Visual

`rotate_wrist(target_yaw=0.4)` on `tiago_velocity.xml`, same script under each tree, MuJoCo headless. The panels are **pixel-identical at frame 0** (mean |delta| 0.0000) and 30.70 apart at the end, so they differ only because of this change. The right panel is honestly static: the call refused, so nothing was commanded.

rotate_wrist on a velocity-driven arm

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-pose-drive-32128579456/pose-drive.mp4) - [still at maximum divergence](https://raw.githubusercontent.com/cagataycali/robots/media-pose-drive-32128579456/pose-drive.png)

## Tests

`tests/simulation/mujoco/test_primitives_write_a_pose_only_into_a_pose_drive.py`, hermetic MJCF fixtures, no downloads. **5 failed / 4 passed on `upstream/main`, 9 pass here.** Pre-fix messages name the cause, including `commanding a rate drive with a joint coordinate diverged the integrator`.

The 4 that pass on both trees are the boundary: a fully position-servo robot's envelope is unchanged (no `uncommanded_drives` key, historical text), every servo joint is still held, the driven-joint-range substitution still applies for a servo jaw, and a velocity drive with its own `ctrlrange` is still commanded.

## Verification

Measured at `93480fbe`, then merged `upstream/main` (`9ba5dc22`) and re-verified.

- Blast radius: 38 files touching these primitives, `joint_drive_map`, the shared result builders and the repo-wide guards - branch **1819 passed / 0 failed**, base **1814 passed / 5 failed**; `1814 + 5 = 1819`, same collected total, branch-only failures **empty**, base-only exactly the 5 pre-fix ids.
- The 2141 tests in every AGENTS.md-scanning guard pass.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` unchanged at its baseline with zero errors in the touched files.

## Docs

A new AGENTS.md convention section, *"Actuators: a joint pose goes only where `ctrl` IS a joint pose"*, recording the rule and both pinning tests. No `docs/` page currently documents these primitives.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1301,6 +1301,35 @@ Corrections from code review that apply to all future contributions:
 - **`except (ImportError, Exception)` is a bug** - `Exception` is a superclass of `ImportError`, so the tuple collapses to `except Exception`. Lint/review will catch this; don't write it.
 - **USB / hardware probing** - use `except (ImportError, OSError)`. `PermissionError` is an `OSError`, `FileNotFoundError` is an `OSError`, etc.
 
+### Actuators: a joint pose goes only where `ctrl` IS a joint pose
+- **`data.ctrl` is not a pose channel, it is whatever the actuator's force law reads.**
+  A `<position kp>` reads it as the joint target, a `<velocity kv>` as a rate, a `<motor>`
+  as a torque, an `<intvelocity>` integrates it as a rate. Writing a joint coordinate into
+  any of the latter commands a different physical quantity that happens to be numerically
+  equal to an angle, and nothing raises: the joint simply moves somewhere the caller did
+  not ask for.
+- **Anything that writes a pose asks
+  `strands_robots.simulation.mujoco.scene_ops.joint_drive_map` first.** Resolving the
+  transmission is not enough - a `<velocity>` actuator's transmission IS the joint, and
+  every stock tendon gripper measured clears the bias-type and position-gain terms a naive
+  servo check would look at. The classification is per actuator, not per robot: `openarm`
+  ships 2 position servos beside 16 motors, and 19 of the loadable registry robots have at
+  least one joint-transmission actuator that is not a servo.
+- **A drive that cannot take the pose is left uncommanded and named, not written to.** The
+  motion primitives hold the joints they can hold and report the rest, because writing a
+  live joint angle into a rate drive is what *moves* the joint the call meant to hold - and
+  a wheel angle accumulates, so that "hold" grows with every turn the wheel has already
+  made. Where the drive being targeted is the one that cannot take a pose, the primitive
+  refuses instead: a servo loop on a rate drive meets its convergence test when the joint
+  sweeps *past* the number while still accelerating, so it reports the set-point as reached
+  and the joint keeps going after the call returns.
+- Pinned by `tests/simulation/mujoco/test_pose_write_reports_whether_the_servos_hold_it.py`
+  for `set_joint_positions(hold=True)` and by
+  `tests/simulation/mujoco/test_primitives_write_a_pose_only_into_a_pose_drive.py` for
+  `move_to` / `rotate_wrist` / `set_gripper`, which also pins the boundary: a usable
+  `ctrlrange` is authoritative and already in the drive's own units, so only the
+  *substitution* of a driven joint's limits needs the drive to be a servo.
+
 ### Clocks: a duration is measured, a stamp is recorded
 - **A duration belongs on `time.monotonic()`.** Anything that decides *how long to keep
   waiting* - a timeout, a deadline, a TTL, a retry window, a rate window, a frame pacer -
--- a/changelog.d/2435-primitives-pose-drive.md
+++ b/changelog.d/2435-primitives-pose-drive.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **simulation/mujoco**: `move_to`, `rotate_wrist` and `set_gripper` now write a joint pose
+  only into an actuator whose `ctrl` IS a joint pose, the split
+  `scene_ops.joint_drive_map` already owns. A `<velocity>` drive reads the same number as a
+  rate and a `<motor>` as a torque, so the primitives previously commanded the wrong
+  physical quantity on any non-servo joint drive and reported success: `rotate_wrist` said
+  a set-point was reached while the joint kept accelerating past it, and the joints it
+  promised to hold were driven away from the pose being held. Where only other joints are
+  affected - the wheels of `lekiwi`, `stretch3` and `tiago_dual`, whose arms are fully
+  servo-driven - those drives are now left uncommanded and named in the payload; where the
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Routes the three motion primitives that write a joint pose into `data.ctrl` (`move_to`, `rotate_wrist`, `set_gripper`'s driven-joint-range substitution) through the pose/non-pose split `scene_ops.joint_drive_map` already owns, closing the gap where a `<velocity>` or `<motor>` joint drive was commanded with a joint coordinate that it reads as a rate or a torque. The split is applied per actuator, refusing only when the *targeted* drive cannot take a pose and otherwise leaving non-pose drives uncommanded and named in a new optional `uncommanded_drives` payload field (omitted when empty, so the envelope stays byte-identical for fully servo-driven robots). Verified locally on the head: the 9 new hermetic tests pass on this branch and reproduce the claimed 5-failed/4-passed split on the merge base (`9ba5dc22`), including the base-tree `QACC` divergence the docstring describes; the surrounding primitive suites (`test_motion_primitives`, `test_set_gripper_setpoint_range_sources`, `test_move_to_solves_over_commanded_joints_only`, `test_move_to_pose_convergence`, `test_pose_write_reports_whether_the_servos_hold_it`, numeric domains, resolution refusals) all pass, and `ruff check` / `ruff format --check` are clean on the touched files.

## What's good

- All new `joint_drive_map` reads happen under `self._lock`, honoring the mjModel/mjData thread-safety contract, and every refusal is a structured `_err` that names the offending actuator and quotes the `send_action` escape hatch - no raises, no silent defaults.
- The `jnt_id is None` (tendon) early return in `_gripper_setpoint_range` still precedes the new drive check, so the Franka-style tendon path is untouched; the Isaac caller of `_rotate_wrist_result` is unaffected by the defaulted new kwarg.
- Conservative failure direction throughout: a joint whose drive classification misses lands in `other_jact` and is declined-and-named rather than written to, which is the cheap error by the map's own analysis.
- Changelog fragment, AGENTS.md convention section with both pinning tests, ASCII-clean strings, no host paths, hermetic fixtures with no downloads or shared-cache reads.

PR #2437, opened 2026-08-18
fix(training): give the in-process run log a single writer

cagataycali:
## Problem

`capture_to_file` (the in-process run-log capture every training backend uses) opened the log **path twice**:

```python
self._stream = open(self.log_path, "w", encoding="utf-8")   # what the stdout/stderr tee writes to
self._fh = logging.FileHandler(self.log_path)               # a second file object over the same path
```

Two file objects over one path track **independent write offsets**. The tee's stream buffers and writes at its own offset; the handler appends at end-of-file. Once the buffer spills past the handler's bytes it overwrites them **in place**. Nothing raises, and `_Tee` swallows write errors by design, so there is no signal at all.

The module's own docstring promises the log holds "stdout/stderr + root-logger output". It held neither whole.

## Why it matters

lerobot reports training progress through the **root logger** (`MetricsTracker`), while the dataset/video layers `print`. That log is what `LerobotTrainer._parse_log` reads back for the RUNNING-vs-learning verdict (`latest_step`, `latest_epoch`, `latest_loss`, `learning`, `liveness_ok`).

Measured on a healthy run that emitted 12 metrics records ending at step 1200 (835 bytes of records):

| stdout written alongside | metrics records left in `run.log` | verdict |
|---|---|---|
| 0 B | 12 / 12 | `latest_step=1200, learning=True, liveness_ok=True` |
| 780 B (one print line per step) | **0 / 12** | **`{'liveness_ok': False}`** |
| 1 560 B | 0 / 12 | `{'liveness_ok': False}` |
| 8 192 B | 0 / 12 | `{'liveness_ok': False}` |
| 32 768 B | 0 / 12 | `{'liveness_ok': False}` |

The bar is one print line per step. A healthy 1200-step run is reported as having produced no metrics at all, with no `latest_step`, no loss and no `learning` key.

Below that threshold the failure is worse than a dropped line. With a short print *after* the records, the single close-flush lands over the **head** of the appended block and leaves the tail of a record behind as a line of its own:

```
training finished, saved checkpoint
ss:0.0100 grdn:0.512 lr:1.0e-04          <- all that is left of "step:100 ... loss:0.0100 ..."
step:200 smpl:800K ep:2 epch:2.00 loss:0.0050 grdn:0.512 lr:1.0e-04
```

That fragment is shorter than any record and still numeric enough to read like one.

## Change

Point the root-logger handler at the stream the tee already holds, so one file object - and one write offset - serves both halves:

```python
self._fh = logging.StreamHandler(self._stream)
```

`StreamHandler.close()` does not close the stream, so the existing teardown order (remove handler, close handler, close stream) is unchanged. Level and formatter are unchanged. A run with no stdout produces a byte-identical log (835 bytes before and after).

## Verification

`tests/training/test_inproc.py` gains `TestCaptureToFileIsTheOnlyWriter`. Pre-fix **6 failed / 18 passed**, post-fix **24 passed**:

| test | pre-fix |
|---|---|
| `test_every_record_survives_a_chatty_stdout` | `12 of 12 emitted records are absent from the log (first missing: 'step:100 ...')` |
| `test_no_record_is_left_half_overwritten` | `the log holds ['ss:0.1234 grdn:0.512'], which was never written` |
| `test_the_verdict_reads_a_chatty_run_as_learning` | `healthy run read as dead: {'liveness_ok': False}` |
| `test_stdout_and_records_keep_their_order` | `['BEFORE', 'AFTER']` - the record vanished |
| `test_the_handler_writes_the_stream_the_tee_holds` | the root cause: `mode='a'` object is not the `mode='w'` object |
| `test_removes_handler_and_closes_on_exit` (existing) | same root cause |

`test_a_quiet_run_is_unchanged` passes on both trees - the control that fails if pointing the handler at the tee's stream cost the records or their formatting.

Two notes on the test design. The existing `test_removes_handler_and_closes_on_exit` asserted `isinstance(added[0], logging.FileHandler)`, which pinned the handler *class* - and the class was the defect; it now pins the contract (`added[0].stream is cap._stream`). And this module's docstring already excludes assertions on which logging *records* reach the file, because record visibility depends on the ambient root-logger level and pytest's capture plugin. The new tests keep that exclusion: they hand records to the installed handler directly, so no logger level, propagation or capture plugin is in the path. What survives a write and what is filtered into one are different questions, and only the first is this module's.

Gate: `ruff check` / `ruff format --check` / `mypy` clean over `strands_robots tests tests_integ` (mypy 19 pre-existing errors, identical set on both trees, 0 in the changed files). A 55-file selection - every test referencing the capture primitives or the verdict parser, all of `tests/training`, and the repo-wide docstring/string/dependency/changelog guards - ran on this branch and on `upstream/main`:

- branch **3 failed / 2959 passed**, base **9 failed / 2953 passed**, both collecting 2962
- branch-only failures: **none**; base-only: exactly the 6 pre-fix failures above
- the 3 shared failures are pre-existing and environmental (`TypeError: encode() takes 1 positional argument but 2 were given` - draccus 0.8.0 locally against lerobot's `draccus>=0.11.6`)

## Artifact

Same capture script run against `upstream/main` and against this branch, driving the real `capture_to_file` and the real verdict parser; only `strands_robots` differs between the columns.

in-process run log, before and after

## Scope

One contract: the log has one writer. `_Tee`'s swallow-and-continue behaviour, the handler level and formatter, the teardown order and the `log_path=None` no-op are all untouched. The `# noqa: SIM115` on the stream open stays - the stream must outlive `__enter__`.

`AGENTS.md` gains a short convention section for the rule, in the format of the neighbouring ones, ending in the test that pins it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1367,6 +1367,24 @@ Corrections from code review that apply to all future contributions:
   base also carries its clock in its name (`started_mono`, `last_idle_render_mono`), so a
   later reader cannot mistake it for a stamp and subtract `time.time()` from it.
 
+### One writer per log file
+- **A file two writers share needs one file object, not one path.** Two file objects over
+  one path each track their own write offset, so a buffered writer flushes at its offset
+  and overwrites in place whatever the other appended there. The training backends' run
+  log tees stdout/stderr *and* root-logger records into a single file, so the logger
+  handler is pointed at the stream the tee already holds
+  (`logging.StreamHandler(stream)`) rather than opening the path a second time
+  (`logging.FileHandler(path)`).
+- **The failure is silent and shaped like data, not like an error.** Records vanish, and a
+  record straddling a flush boundary survives as a fragment that still reads like one - so
+  the loss surfaces as a wrong *value* somewhere downstream. The log a training backend
+  parses for its "RUNNING != learning" verdict reported a healthy 1200-step run as having
+  produced no metrics at all, with nothing raised and nothing logged.
+- Pinned by `tests/training/test_inproc.py::TestCaptureToFileIsTheOnlyWriter`, which
+  asserts the log holds exactly the lines that were written - so a dropped record and a
+  surviving fragment both fail - and hands records to the installed handler directly, so
+  the assertion does not depend on ambient logger levels or on pytest's capture plugin.
+
 ### Module-Level Side Effects
 - **If you must run code at import time, comment WHY it can't be lazy.** `MUJOCO_GL` is the canonical example: MuJoCo locks the GL backend at first `import mujoco`, so the env var must be set before any downstream import chain triggers it.
 - **Cheap-guard optional imports** - `if importlib.util.find_spec("mujoco") is not None:` before doing `from strands_robots.simulation.mujoco.backend import _configure_gl_backend`. Users without the `[sim-mujoco]` extra shouldn't pay an import-attempt cost on every `import strands_robots`.
--- a/changelog.d/2437-inproc-single-log-writer.md
+++ b/changelog.d/2437-inproc-single-log-writer.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **training**: the in-process run log now has a single writer, so a chatty stdout can no longer overwrite the root-logger records in it. The log holds stdout/stderr and logger output whole, and a trainer's "RUNNING != learning" verdict (`latest_step` / `learning` / `liveness_ok`) reads a healthy run correctly again.
--- a/strands_robots/training/_inproc.py
+++ b/strands_robots/training/_inproc.py
@@ -11,7 +11,8 @@ Two primitives:
 * :func:`call_callable` - run a Python callable in THIS interpreter with its
   stdout/stderr + root-logger output tee'd to a per-run log file (so the
   trainers can still parse a "RUNNING != learning" verdict, exactly as they did
-  from the old subprocess log).
+  from the old subprocess log). Both halves go through one file object, so
+  neither can overwrite the other's bytes.
 
 * :func:`elastic_launch_callable` - multi-GPU single-node launch via torch's
   programmatic :class:`torch.distributed.launcher.api.elastic_launch` (the same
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Fixes a silent data-corruption bug in `capture_to_file`: the run log was opened by two independent file objects (the tee's `open(..., "w")` stream and a `logging.FileHandler` over the same path), each tracking its own write offset, so a buffered tee flush overwrote the root-logger records in place - the exact log `LerobotTrainer._parse_log` reads back for its RUNNING-vs-learning verdict. The fix points the handler at the stream the tee already holds (`logging.StreamHandler(self._stream)`), giving the file a single writer and a single offset. Teardown is unchanged and remains correct: `StreamHandler.close()` does not close the underlying stream, so the existing remove-handler / close-handler / close-stream order still closes the file exactly once. Level, formatter, the `log_path=None` no-op and `_Tee`'s swallow-and-continue contract are all untouched, and the new `StreamHandler`-typed `_fh` annotation is compatible (`FileHandler` is a `StreamHandler` subclass). Verified locally on the PR head: `tests/training/test_inproc.py` reports 24 passed; with the pre-fix `_inproc.py` swapped in, the new `TestCaptureToFileIsTheOnlyWriter` class reports 5 failed / 1 passed - the one pass being the quiet-run control, exactly as the description claims.

## What's good

- The regression tests pin the contract (log holds exactly what was written - so both a dropped record and a surviving fragment fail) rather than the handler class, and the pre-existing `isinstance(added[0], logging.FileHandler)` assertion that pinned the defective class is corrected to pin the shared-stream contract instead.
- Handing records straight to the installed handler keeps the module's documented exclusion of record-visibility assertions, so the new tests are independent of ambient root-logger levels and pytest's capture plugin.
- Changelog fragment (`changelog.d/2437-inproc-single-log-writer.md`) per step 3 of the PR workflow, and the AGENTS.md convention section follows the neighbouring format, ending in the test that pins it.
- Tight scope: one contract changed, teardown/formatter/level/no-op paths untouched.

PR #2438, opened 2026-08-18
fix(mesh/transport): publish and release the Zenoh leg on the raw path

cagataycali:
## Problem

`strands_robots.mesh.session` exposes every Zenoh operation twice: a public,
**backend-aware** entry point that resolves whatever `STRANDS_MESH_BACKEND`
selects, and a private `_*_directly` helper that always takes the raw Zenoh path.
The module ships that bypass family for one stated reason -
`_get_zenoh_session_directly`'s docstring says a backend-aware call from inside a
`BridgeTransport` "would re-enter the factory's `_LOCK` ... causing a deadlock".

`ZenohTransport` used it for four of its six delegations and reached for the
backend-aware `session.put` and `session.release_session` for the other two.
Under `STRANDS_MESH_BACKEND=bridge` both resolve the `BridgeTransport` that owns
that very `ZenohTransport`, so each routed straight back into the caller.

Both re-entries are silent, which is what let them survive:

| | measured on `upstream/main` |
|---|---|
| `bridge.put('strands/arm-01/state')` | entered `BridgeTransport.put` **249** times, returned normally |
| payloads on the Zenoh session | **0** |
| publishes to MQTT | **249** copies of the same payload |
| the only trace | one DEBUG line, `[bridge] zenoh.put error ...: maximum recursion depth exceeded` |
| a LAN-only topic (`.../camera/wrist`) | **0** on the LAN, **0** on the WAN - lost entirely |
| `factory.release_transport()` | never returned (10 s bound); singleton not cleared, session not closed |

`RecursionError` is a `RuntimeError` subclass, so the bridge's own narrow
`except (RuntimeError, ConnectionError, OSError)` absorbed it - a `put` that
publishes nothing is indistinguishable from a successful fire-and-forget publish.
The release blocked on the transport factory's non-reentrant lock from the thread
already holding it.

`docs/security.md` describes this backend as one where "a `BridgeTransport` keeps
high-rate topics local while bridging presence, health, and safety to the cloud".
Both halves were untrue: high-rate LAN-only topics reached the LAN zero times, and
a bridged topic reached the cloud 249 times rather than once.

`IotMqttTransport.put` publishes through its own MQTT client, which is why only
the Zenoh leg could route back into its owner.

## Change

Completes the bypass family with `_put_zenoh_directly` and
`_release_zenoh_session_directly`, and points the two delegations at them. Both
are **extracted from the existing legacy branches** of `put` / `release_session`
rather than copied, so the raw publish and the legacy refcount each live in one
place and the zenoh-backend path is unchanged by construction.

`put` and `release_session` stay backend-aware for the callers that want routing
(`mesh.core` releases through the public one deliberately).

| `ZenohTransport` method | session entry point | raw path before |
|---|---|---|
| `connect` | `_get_zenoh_session_directly` | yes |
| `is_alive` | `_session_alive_directly` | yes |
| `raw_session` | `_current_zenoh_session_directly` | yes |
| `declare_subscriber` | via `raw_session` | yes |
| **`put`** | `put` -> `_put_zenoh_directly` | **no** |
| **`close`** | `release_session` -> `_release_zenoh_session_directly` | **no** |

The module and class docstrings named `session.get_session` / `session.put` /
`session.release_session` as the delegation targets; they now name the bypass
family and say why.

## Tests

`tests/mesh/test_zenoh_transport_bypasses_backend_routing.py`, driving a real
`ZenohTransport` over a fake `zenoh.Session` object - no broker, no `zenoh` wheel.
Every existing bridge test injects a *fake* Zenoh leg whose `put`/`close` record
instead of delegating, so no fixture in the suite could reach either path; the one
existing `ZenohTransport.put` test is a disconnected no-op whose comment names the
delegation and which only ever ran on the backend where it was correct.

**5 failed / 3 passed before, 8 pass after.** Pre-fix messages:

- `one publish entered BridgeTransport.put 241 times and delivered 0 payload(s) to the Zenoh session`
- `factory.release_transport() did not return within 10s: closing the bridge released its Zenoh leg through the backend-aware session.release_session, which re-enters the factory lock this thread already holds`
- `ZenohTransport delegates to backend-aware session entry point(s) ['put', 'release_session']`

The deadlock test runs the teardown on a worker thread with a bounded join, so a
pre-fix hang is a failed assertion rather than a suite that stops progressing, and
it swaps in a throwaway factory lock so the wedged thread blocks nothing else.

The 3 that pass on **both** sides are the controls, each failing for an obvious
wrong fix: the legacy zenoh backend still publishes to the session; `session.put`
still resolves the active transport (fails if the router is made to publish raw);
and `session.release_session` still delegates to the factory (fails if the backend
branch is dropped). A structural test asserts every `mesh.session` import in the
transport names a `_*_directly` helper, so a seventh delegation cannot
reintroduce either re-entry.

## Verification

Measured at `388a52f2` on a 227-file selection (everything referencing the changed
symbols, all of `tests/mesh`, and the repo-wide docstring / changelog / string
guards), same selection on both trees:

- branch **21 failed / 5624 passed / 24 errors**, base **26 failed / 5619 passed / 24 errors**
- `21+5624 == 26+5619 == 5645` collected on both; branch-only failing ids **empty**;
  base-only is exactly the 5 pre-fix ids with nothing foreign
- the 21 shared failures and 24 errors are pre-existing (absent `awsiot`)
- `mypy strands_robots tests tests_integ`: 19 on both trees, identical sets, none in the changed files
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`
- re-verified after merging the base forward: 826 passed over the new tests, the
  overlapping `tests/training/test_inproc.py`, the mesh transport suites and the guards

## Measurement

zenoh transport bypass measurement

One probe script, run under `PYTHONPATH` of an `upstream/main` worktree and of
this branch, publishing one bridged topic and one LAN-only topic and then
releasing the last reference. Every number in the figure is asserted against the
captured JSON before it is drawn.

## Docs

`AGENTS.md > Key Conventions` gains the rule as item 17 - a transport delegates to
the raw backend path, never to the router that resolves it - including why both
re-entries are silent and why a fixture that injects a fake leg into a composite
transport hides them.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -166,6 +166,26 @@ hatch run format            # ruff check --fix, ruff format
     audit dir to `tmp_path`, so there the whole log *is* the record set it
     means. Pinned by tests/test_examples_attest_only_what_they_report.py.
 
+17. **A transport delegates to the raw backend path, never to the router that
+    resolves it** - `strands_robots.mesh.session` exposes every Zenoh operation
+    twice: a public, backend-aware entry point (`get_session`, `put`,
+    `release_session`, `session_alive`) that resolves whatever
+    `STRANDS_MESH_BACKEND` selects, and a private `_*_directly` helper that
+    always takes the raw Zenoh path. A `MeshTransport` implementation must use
+    the second kind for *every* delegation, because under
+    `STRANDS_MESH_BACKEND=bridge` the router resolves the `BridgeTransport` that
+    owns that very transport - so a backend-aware call routes straight back into
+    the caller. Both re-entries are silent, which is what makes the rule worth
+    stating rather than leaving to review: a re-entrant `put` raises
+    `RecursionError`, which is a `RuntimeError` subclass and so is absorbed by
+    the narrow `except (RuntimeError, ConnectionError, OSError)` that idempotent
+    transport paths are required to use, and a re-entrant `close` blocks on the
+    factory's non-reentrant lock from the thread already holding it. Neither
+    reports anything a caller can act on. Every fixture that injects a *fake*
+    leg into a composite transport hides this by construction, so the raw path
+    has to be pinned structurally. Pinned by
+    tests/mesh/test_zenoh_transport_bypasses_backend_routing.py.
+
 ## PR Workflow
 
 1. Create the feature branch **on your fork**. Branch creation in the base
--- a/changelog.d/2438-zenoh-transport-bypasses-backend-routing.md
+++ b/changelog.d/2438-zenoh-transport-bypasses-backend-routing.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **mesh/transport**: `ZenohTransport` now publishes and releases through the raw
+  Zenoh path instead of the backend-aware `session.put` / `session.release_session`.
+  Under `STRANDS_MESH_BACKEND=bridge` those resolve the `BridgeTransport` that owns
+  the transport, so a publish re-entered the bridge until the stack was exhausted -
+  delivering nothing to the LAN, republishing a bridged topic to MQTT once per
+  re-entry, and returning normally - and a release blocked on the transport
+  factory's non-reentrant lock, so teardown never completed and the session was
+  never closed.
--- a/strands_robots/mesh/session.py
+++ b/strands_robots/mesh/session.py
@@ -923,13 +923,33 @@ def release_session() -> None:
     ``undeclare`` calls. The "session closed" INFO line is emitted only when
     the close actually completed.
     """
-    global _SESSION, _SESSION_REFS  # noqa: PLW0603
-
     if _is_transport_backend():
         from strands_robots.mesh.transport.factory import release_transport
 
         release_transport()
         return
+
+    _release_zenoh_session_directly()
+
+
+def _release_zenoh_session_directly() -> None:
+    """Release one reference to the raw Zenoh session, bypassing backend routing.
+
+    The teardown companion to :func:`_get_zenoh_session_directly`, and the
+    function :func:`release_session` itself calls once the backend branch is
+    not taken - so the legacy refcount and its close contract (documented on
+    :func:`release_session`) live in exactly one place.
+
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Fixes a silent re-entrancy pair in `ZenohTransport`: under `STRANDS_MESH_BACKEND=bridge` its `put` and `close` delegated to the backend-aware `session.put` / `session.release_session`, which resolve the `BridgeTransport` that owns the transport - so a publish recursed into the bridge until stack exhaustion (absorbed by the bridge's narrow `except (RuntimeError, ConnectionError, OSError)`, since `RecursionError` is a `RuntimeError` subclass) and a release deadlocked on the factory's non-reentrant lock. The fix completes the existing `_*_directly` bypass family with `_put_zenoh_directly` / `_release_zenoh_session_directly`, both extracted verbatim from the legacy branches of the public functions, and points the two delegations at them.

## What's good

- The extraction is byte-faithful: the legacy zenoh-backend bodies of `put` and `release_session` are unchanged, and the public routers keep their backend branch for callers that want routing (`mesh.core`).
- Lock ordering stays one-directional: the backend branches of every `session.*` router return before `_SESSION_LOCK` is taken, so the new factory-lock -> session-lock path introduced by `BridgeTransport.close` -> `_release_zenoh_session_directly` has no inverse anywhere.
- Verified the pre-fix claim independently: the new test file run against the merge base reports exactly **5 failed / 3 passed** with the advertised messages, and **8 passed** on the head; the surrounding transport suites (252 tests) pass on the head.
- The deadlock test's bounded worker-thread join and throwaway factory `_LOCK` mean a pre-fix hang fails cleanly instead of wedging the suite.
- The structural AST test (`_*_directly`-only imports from `mesh.session`) closes the class, not just the two instances.
- Conventions hold: changelog fragment present, `monkeypatch.setenv` throughout, no host paths, no non-ASCII, ruff check/format clean on the changed files.

PR #446, opened 2026-06-14
test(mesh): cover sensor reader branches (sensors.py 73%->87%)

cagataycali:
## Problem

The extended Mesh sensor publishers in `strands_robots/mesh/sensors.py` were only exercised indirectly through the threaded `_*_loop` methods. Those loop tests sleep on a real timer and assert that *something* gets published, so the branch-rich `_read_*` readers never had their outputs verified. The priority/fallback logic was effectively untested:

- the SE(3) 4x4-matrix pose path (translation extraction + quaternion decomposition, including the non-positive-trace branch),
- SLAM and odometry pose fallbacks,
- the inner-robot IMU observation path (`robot.robot.get_observation()` with ndarray->list conversion and truncation),
- multi-source health aggregation (dict vs scalar battery, temps),
- the uniform safety-event wire severity (issue #272: wire severity pinned to `info` so it can't be used as a content-channel oracle, true severity kept only in the audit record).

## Change

Add `tests/mesh/test_sensor_readers.py`: 27 focused unit tests that drive the readers directly through a minimal `SensorLoopsMixin` host, asserting on returned payloads rather than thread timing. No timer dependence, no real Zenoh transport.

Coverage for `strands_robots/mesh/sensors.py`: **73% -> 87%** (97 -> 47 missing lines), measured before/after by toggling the new test file in the `tests/mesh/` run.

## Tests

```
pytest tests/mesh/test_sensor_readers.py -q   # 27 passed
pytest tests/mesh/ -q                          # 1445 passed, 2 skipped
```

Lint/format/mypy clean. Test-only change; no library behavior modified.

PR #101, opened 2026-04-21
feat(mesh): Mesh, AWS IoT - session, presence, RPC, streams, wiring

cagataycali:
## Summary

Full Zenoh mesh implementation for `strands-labs/robots:main` — consolidates the original 6-PR roadmap into a single reviewable PR. Every `Robot()` and `Simulation()` constructed in a process is now automatically a peer on the local Zenoh mesh — no manual setup.

This PR has grown beyond the LAN-only mesh: it now also lands a **pluggable transport layer** with a complete **AWS IoT Core integration** sitting alongside Zenoh, behind the same `Mesh` API, end-to-end validated against a real AWS account.

```python
from strands_robots import Robot

# Default: Zenoh on the LAN (zero config)
sim_a = Robot("so100")
sim_b = Robot("so100")
sim_a.mesh.tell(sim_b.mesh.peer_id, "pick up the cube")
sim_a.mesh.emergency_stop()
```

```bash
# One env var → same code now runs over AWS IoT Core (mTLS)
export STRANDS_MESH_BACKEND=iot
export STRANDS_IOT_THING_NAME=so100-arm-01
export STRANDS_IOT_ENDPOINT=xxx.iot.us-west-2.amazonaws.com
```

```python
# OOBE: one call provisions the Thing + cert + policy
from strands_robots.mesh.iot import provision_robot
provision_robot("so100-arm-01")  # → ~/.strands_robots/iot/{cert.pem,key,CA}
# Then Robot()/Mesh() Just Works on AWS IoT.
```

## Review Status

- ✅ All **10/10 review threads resolved**
- ✅ CI green: ruff + mypy + pytest: **1678 passed**, 120 skipped
- ✅ @sundargthb: ZenohTransport deadlock — `_get_zenoh_session_directly()` bypass (dad059c)
- ✅ @sundargthb: IoT `handlers.pop()` → `.remove(self._handler)` by identity (dad059c)
- ✅ @sundargthb: peer_id validation — reject reserved names + MQTT-unsafe chars (dad059c)
- ✅ @sundargthb: README IoT docs — gated behind `dry_run=True` safety (dad059c)
- ✅ @sundargthb: `bootstrap_account()` safety — `confirm=True` required, `dry_run=True` default (dad059c)
- ✅ @sundargthb: E-stop Lambda IAM — `iot:Publish` scoped to `strands/*` (dad059c)
- ✅ @sundargthb: Operator shadow — attribute condition `strands-mesh-role=robot` (0097220)
[...]

PR #2439, opened 2026-08-18
fix(mesh): report the rotation a pose matrix encodes, not the identity past 120 degrees

cagataycali:
## The problem

`strands/{peer_id}/pose` decomposes a robot's 4x4 SE(3) pose provider into `x`/`y`/`z`, a planar `theta`, and a `quat`. The quaternion was built from the **trace branch of Shepperd's method alone**:

```python
trace = float(mat[0, 0] + mat[1, 1] + mat[2, 2])
if trace > 0:
    s = 0.5 / np.sqrt(trace + 1.0)
    ...
else:
    w, x, y, z = 1.0, 0.0, 0.0, 0.0   # the identity
```

That branch divides by `sqrt(trace + 1)`, which goes to zero as the trace approaches `-1`, so it cannot serve `trace <= 0` at all. A rotation's trace is `1 + 2*cos(angle)`, so **`trace <= 0` is exactly `angle >= 120 degrees` - 61% of SO(3) under the uniform measure** (measured over 20 000 uniformly drawn rotations: 12 190 of them). It is not a corner of the domain; it is where a robot that has turned to face back the way it came sits.

Substituting the identity there makes one payload carry two different headings, because `theta` is decomposed correctly from the same matrix:

| robot's true heading | SE(3) trace | `theta` published | `quat` published | the quat's angle |
|---|---|---|---|---|
| 80 deg | +1.347 | 80.0 deg | `[0.766, 0, 0, 0.643]` | 80.0 deg |
| 116 deg | +0.123 | 116.0 deg | `[0.508, 0, 0, 0.861]` | 116.0 deg |
| 140 deg | -0.532 | 140.0 deg | `[1, 0, 0, 0]` | **0.0 deg** |
| 180 deg | -1.000 | 180.0 deg | `[1, 0, 0, 0]` | **0.0 deg** |

Two things made it undetectable downstream. The identity **is** a valid unit quaternion, so no consumer had a reason to question it; and the error's magnitude equals the rotation itself, so the reading is not noisy - it is confidently wrong. A half turn about any axis (`trace = -1`) reported as "has not moved".

## Why it survived

The `else` branch dates from #101, the original mesh feature PR. The test covering it arrived with #446 - *"cover sensor reader branches (sensors.py 73%->87%)"* - a coverage sweep, which pinned whatever the code did:

```python
def test_read_pose_matrix_negative_trace_uses_identity_quat() -> None:
    # 180-degree rotation about X gives a non-positive trace branch.
    ...
    assert pose["quat"] == [1.0, 0.0, 0.0, 0.0]
```

Its comment names the trigger, not a justification. Its case is also a 180-degree **roll**, whose `theta` is `arctan2(0, 1) = 0` - so it was the one orientation where the payload's self-contradiction is invisible. This PR replaces it with a contract pin.

## The fix

Take all four branches (trace, then the largest diagonal term), normalize, and sign-canonicalize to `w >= 0`. Two properties come with it:

- **normalized** - a pose integrated from odometry or handed over by a SLAM stack is only approximately orthonormal, and the old code let that drift onto the wire as the quaternion's *length* (measured 1.0017 for a block scaled by 1.004), where a consumer reading it as a unit quaternion cannot see it.
- **`w >= 0`** - `q` and `-q` are the same rotation, so canonicalizing keeps repeated reads of an unchanged pose identical.

Rotations below the boundary are untouched: over the `trace > 0` half of that 20 000-rotation sample, old and new agree to **3.4e-06 deg**.

### Scope note

`simulation/isaac` and `simulation/newton` each hold this rule correctly, and this is deliberately a third copy rather than an import. `strands_robots.mesh` must not depend on `strands_robots.simulation`, and the module those layers do share (`strands_robots.utils`) is the refusal-guard module - it carries a scanned, tested invariant that no function in it converts a caller's value outside a `try`, which a matrix decomposition cannot honour without a meaningless wrapper. Giving all three one owner is a refactor of two already-correct implementations and does not belong in a fix to a wire payload; the reason is recorded in the helper's docstring so the next reader does not have to rediscover it.

## Tests

`tests/mesh/test_sensor_readers.py`, asserting through `_read_pose()` only - it imports nothing new, so every pre-fix failure is behavioural with a number in it. On `main`: **17 failed / 81 passed**; after: **98 passed**.

The headline is that one payload cannot carry two headings, and pre-fix it reads `assert 0.0 == 180.0`. Beside it: a half turn about each axis is a half turn; a round trip (recompose the published quaternion, compare to the input matrix) over an exact axis-angle set and over a 300-rotation uniform sample of SO(3), with the sample asserted to straddle the boundary so a lucky seed cannot pass vacuously.

The 81 that pass on both trees are the bound on the change. Two matter most: **`quat_unchanged_below_the_trace_boundary`** pins the exact quaternion for yaws the trace branch already served, so it fails if the branch selection is *changed* rather than *extended*; and **`quat_is_always_a_unit_quaternion`** passes pre-fix too - which is the point, since the wrong answer was unit length.

## Verification

Blast radius: 309 test files (all of `tests/mesh`, plus every test touching the changed symbols or the repo-wide docstring/string/changelog guards), run on this branch and on `upstream/main` with the new test file copied in.

| | failed | passed | collected |
|---|---|---|---|
| this branch | 21 | 14093 | 14114 |
| `upstream/main` + new tests | 38 | 14076 | 14114 |

Failed delta = passed delta = 17 = the pre-fix failure count, and the branch-only failure set is **empty**; the 21 shared failures are pre-existing (absent optional extras). `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 19 errors on both trees, none in the changed files.

One script per tree: sweep a robot's true heading 0 -> 180 degrees, read the pose payload through the real mesh reader of that tree, then place the robot at the orientation it was told - which is what a fleet consumer does with the topic.

pose quaternion

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-mesh-pose-quat-32150871764/pose-quaternion.mp4) - [final frame](https://raw.githubusercontent.com/cagataycali/robots/media-mesh-pose-quat-32150871764/pose-quaternion-final.png)

The two panels are **pixel-identical (mean |delta| 0.0000) for the whole 0-120 degree range** - the change is a no-op there - and diverge only past it, to 11.70. Left, the reported robot freezes at 124 degrees and snaps back to facing forward (31 of 45 frame transitions move, then nothing) while `theta` beside it keeps counting up to 180; right, it tracks every heading (45 of 45 moving). Worst `|reported - true|`: **180.0 deg before, 0.0 deg after**, over 46 headings.


yinsong1986 (approved):
## Summary

Completes Shepperd's method in the mesh `pose` topic's SE(3)-to-quaternion decomposition: the old code took only the trace branch and substituted the identity quaternion for `trace <= 0` (every rotation of 120 degrees or more), so a payload could carry `theta = pi` beside a `quat` meaning no rotation. The new `_quat_wxyz_from_rotmat` takes all four branches, normalizes, and sign-canonicalizes to `w >= 0`. Independently re-verified the implementation: over 50 000 uniformly drawn rotations the published quaternion round-trips to the input matrix at 1.3e-15 worst-case and matches scipy's reference conversion (scalar-first, `w >= 0`) to 3.3e-16; exact half-turns about principal and skew axes are all reported correctly; the 61% non-positive-trace measure reproduces. The `w >= 0` canonicalization cannot perturb previously-correct outputs, since the trace branch always yields `w > 0` - consistent with the below-boundary pin in the tests. Wire shape is unchanged (scalar-first list of four floats); only the previously-wrong values move, and the contract is now documented in `docs/mesh.md`.

## What's good

- The replaced coverage-sweep test (a 180-degree roll, the one orientation where the payload's self-contradiction was invisible) is superseded by contract pins: payload self-consistency, half-turns, round-trips over an exact axis-angle set and a 300-sample of SO(3) asserted to straddle the branch boundary, plus a below-boundary exact-value pin that fails if the branch selection is changed rather than extended.
- The deliberate third copy of the conversion (rather than a mesh -> simulation import or a `utils.py` home) is justified against the layering and refusal-guard-module constraints in the helper's docstring, so the next reader does not re-litigate it.
- Changelog fragment, docs update, and full `Args:` docstring all present; diff is ASCII-clean with no host paths.

PR #2440, opened 2026-08-18
fix(policies/moveit2): report every sidecar planning failure instead of exiting

cagataycali:
## Problem

`_plan` in the MoveIt2 ZMQ sidecar documents that it "Catches planner exceptions so the ZMQ loop can return a structured `{"success": False}` response **instead of crashing**". The `try` covers only `component.plan()`. Every other `moveit_py` interaction in the function is unguarded, so it raises instead of reporting:

| failure source | before | after |
| --- | --- | --- |
| unknown planning group | `unknown_planning_group:` | unchanged |
| no goal supplied | `missing_goal:` | unchanged |
| planner raises | `planner_exception:` | unchanged |
| planner returns empty | `planner_returned_empty` | unchanged |
| current-state read raises | **raises `RuntimeError`** | `start_state_error:` |
| goal names a joint the group lacks | **raises `RuntimeError`** | `invalid_goal:` |
| goal pose link unresolvable | **raises `RuntimeError`** | `invalid_goal:` |
| `target_pose` is not 7 values | **raises `ValueError`** | `invalid_goal:` |
| trajectory shape differs | **raises `AttributeError`** | `trajectory_error:` |
| joint position not numeric | **raises `ValueError`** | `trajectory_error:` |
| happy path | `success=True` | unchanged |

Six of eleven sources crashed, and the four that reported are exactly the four the existing test file pinned.

The first of those is reachable through a conformant client. `MoveIt2Policy` validates `target_joints` key *syntax* (`^[A-Za-z][A-Za-z0-9_-]*$`) but not whether the planning group has those joints, so an ordinary joint-name typo passes validation and reaches `component.set_goal_state(...)`.

REQ/REP is lockstep, so an escaping exception is not one lost answer. It unwinds the REP loop - which has no guard around dispatch - runs the `finally` that closes the socket with the in-flight reply never sent, and ends `main`. Driving the real loop with three requests where the second names a bad joint:

```
requests received : 3
replies sent      : 1     [{'status': 'ok'}]
main() returned   : None  (RuntimeError: Joint 'elbo' not found in group 'arm')
socket closed     : True
```

One bad request from one client stops planning for every client until the container is restarted.

## Change

Guard every `moveit_py` interaction in `_plan`, each reporting the stage that failed in `status`. `moveit_py` is a C++ binding whose exception types are not enumerable from Python, which is why the existing `component.plan()` guard is broad; the new guards follow it with the same `noqa` and rationale. The four statuses that already reported are byte-identical, so nothing an operator or fork matches on moves.

Guard dispatch in the REP loop so every request received gets exactly one reply. This is not redundant with the above: the module explicitly invites forks to edit these handlers ("Forks that need start-state override should plug their own `RobotState` builder here", "Forks: parameterise this", the `world_update` schema note), so an edit must not be able to turn into a process exit. Unexpected errors are answered with the `{"error": ...}` envelope the loop already uses for unknown endpoints and undecodable requests, which the client raises as `RuntimeError` - no client change.

`docs/policies/moveit2.md` gains a "Failure reporting" section under Wire protocol tabulating the `status` kinds and stating the one-reply-per-request guarantee, since the wire protocol is what a fork implements against.

The nearest sibling states the same posture: `policies/groot/server_wrapper.py` guards with `# noqa: BLE001 - degrade to non-strict, never kill the server`.

Reject a payload that is not a map before reading its endpoint. `msgpack.unpackb` decodes any valid msgpack value, so the undecodable-bytes guard did not leave `request` a mapping: the single byte `0x2a` is the integer 42, and a string, list, nil, bool or float decode just as cleanly. `request.get("endpoint")` on one of those raised *between* the two guards, where nothing answered - the same shape as the two above, at a third site. Measured on the real loop with a ping either side of the payload, 6 of the 7 malformed-payload shapes answered 1 of 3 requests and left the sidecar dead; only bytes that do not decode at all were answered. The check replies in the same `malformed_request:` class as undecodable bytes (both are the peer's bytes not being a request) and names the type that arrived, rather than blaming the sidecar with `internal_error:`. A non-map `data` field is unaffected - it is read inside the dispatch guard, which already answered it.

## Verification

One operator session, three tasks, one sidecar. Everything is shipped code - the real `zmq_node.main()` REP loop on a real ZMQ socket, the real `MoveIt2Policy`, a real MuJoCo panda driven by `run_policy`. Only `moveit_py` is doubled (it cannot be installed outside a ROS 2 environment). Task B names `elbo` instead of `elbow`.

| | before | after |
| --- | --- | --- |
| A: reach, valid goal | `success`, 60 commands | `success`, 60 commands |
| B: reach, joint typo | `error` after 1.21 s: `Resource temporarily unavailable` | `error` in 0.01 s: `invalid_goal:Joint 'elbo' not found in group 'arm'` |
| C: retract, valid goal | **`error`** - same socket error | `success`, 60 commands |
| sidecar at end | **exited** | running |
| commands applied | 60 | 120 |

Task C is a valid task that fails only because task B's typo killed the sidecar. The before column also reports a bare socket error naming nothing about the goal, after a full timeout; the after column names the joint and the group immediately.

sidecar failure reporting

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-moveit2-sidecar-failures/moveit2-sidecar-failures.mp4) | [still](https://raw.githubusercontent.com/cagataycali/robots/media-moveit2-sidecar-failures/moveit2-sidecar-failures-still.png)

The two panels are pixel-identical while task A runs (mean |delta| `0.0000` at the first frame, `0.0001` at the last frame of task A) and `11.24` apart at the end, so they diverge only where the change bites. Total motion after task A: `0.004` before (frozen) versus `47.844` after.

The third site has its own session, because the payload it needs is one the shipped
client cannot construct: a stray peer opens a raw REQ socket mid-session and sends
`b'\x2a'`. Any peer may do this - the reference sidecar accepts every client.

| | before | after |
| --- | --- | --- |
| A: reach, valid goal | `success`, 60 commands | `success`, 60 commands |
| stray peer sends `b'\x2a'` | **no reply** (`ContextTerminated` - the socket closed while it waited) | answered `malformed_request:expected a msgpack map, got int` |
| C: retract, valid goal | **`error`**, 0 commands | `success`, 60 commands |
| sidecar at end | **exited** | running |

Task C is a valid operator task that fails only because a stray peer sent one byte.

non-map payload halts the sidecar

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-moveit2-sidecar-failures/moveit2-non-map-payload.mp4) | [still](https://raw.githubusercontent.com/cagataycali/robots/media-moveit2-sidecar-failures/moveit2-non-map-payload-still.png)

Panels pixel-identical while task A runs (mean |delta| `0.0001` at the first frame),
`11.24` apart at the end; total motion after the stray payload `0.0044` before
(frozen) versus `47.84` after.

## Tests

Extended `tests/policies/moveit2/test_zmq_sidecar.py`, whose four existing failure-mode tests become controls. Against `upstream/main`: **18 failed, 14 passed**; after: 32 passed. Each behavioural failure names the escaping exception and its consequence, for example:

```
_plan raised RuntimeError: Joint 'elbo' not found in group 'arm' instead of
reporting the failure. The exception unwinds the REP loop, so the sidecar
closes its socket with the peer's reply never sent and stops serving every
other client.
```

The two loop tests fail pre-fix by the exception escaping `main` itself.

Two of the new tests pass on both trees and are the scope bound: the four already-guarded statuses keep their exact strings, and a plan that succeeds still serialises the same rows.

Gate on a 90-file selection covering `tests/policies/moveit2/`, every test referencing the moveit2 surface, and the repo-wide docs/changelog/docstring/source-string guards: **3011 passed / 0 failed** on the branch versus **2990 passed / 0 failed** on `upstream/main`, both with 52 skipped - the +21 is exactly the new tests, and there are no pre-existing failures in the selection. `ruff check`, `ruff format --check` and `mypy` clean over `strands_robots tests tests_integ`; mypy is 20 pre-existing errors on both trees with none in the changed files.


strands_robots/policies/moveit2/server/zmq_node.py
```diff
@@ -234,28 +270,40 @@ def main(argv: list[str] | None = None) -> int:
             endpoint = request.get("endpoint", "")
```
yinsong1986, line 288: [MUST FIX] `endpoint = request.get("endpoint", "")` (and `data = request.get("data") or {}` below it) sit *between* the two guards: after the `malformed_request` guard around `msgpack.unpackb`, and before the new dispatch `try`. A payload that is valid msgpack but not a map — the single byte `b'\x2a'` decodes to `42`, and a string, list, or nil behave the same — passes `unpackb` and then raises `AttributeError: 'int' object has no attribute 'get'` here. Nothing catches it, so it unwinds the REP loop, runs the `finally` that closes the socket with the reply never sent, and ends `main` — the exact one-bad-request-kills-every-client failure this PR exists to close, still reachable by any peer (the sidecar accepts any client, no auth).

This can't be deferred because the PR *documents* the closed class as a wire-protocol guarantee: the module docstring now states "every request that is received gets exactly one reply" (pinned by `test_loop_reports_the_invariant_it_holds`), and `docs/policies/moveit2.md` tells forks "an undecodable request ... comes back as `{"error": ...}`". Merging with this hole ships a contract that is false on day one for the audience the docs say implements against it.

Fix is one move: pull these two lines inside the dispatch `try` (the `internal_error:` envelope then answers it), or reply `malformed_request:` and `continue` when `not isinstance(request, dict)` — the latter keeps the error class consistent with the undecodable-bytes case.

yinsong1986 (commented):
## Summary

Guards every `moveit_py` interaction in the MoveIt2 sidecar's `_plan` (start-state read, goal set, trajectory serialisation) so each failure stage reports a structured `{"success": false, "status": ...}` instead of raising, and wraps REP-loop dispatch so a raising handler is answered with the existing `{"error": ...}` envelope rather than unwinding `main` and closing the socket with the in-flight reply never sent. The four previously-reporting statuses are pinned byte-identical, the wire-protocol docs gain a failure-reporting table, and the changelog fragment follows convention. One gap remains in the one-reply-per-request guarantee the PR now documents — flagged inline.

## What's good

- The already-guarded statuses are pinned as exact strings, so nothing operators or forks match on moves.
- The broad `except Exception` guards carry the same `noqa` + rationale as the `policies/groot/server_wrapper.py` precedent, and they are recovery paths (report structurally, keep serving), consistent with AGENTS.md's error-handling contract.
- Loop-level tests assert replies-sent == requests-received against the real `main()` loop, including a response msgpack cannot pack.
- Changelog fragment, docs, and docstring all updated in the same PR; docstring completeness is itself pinned by a test.

strands_robots/policies/moveit2/server/zmq_node.py
cagataycali: Fixed, and the case is wider than the integer: reading `endpoint` off a non-map was fatal for **6 of the 7** malformed-payload shapes. Driving the real loop with a `ping` either side of the payload, on the pre-fix tree:

| second request | replies | `main()` returned | sidecar |
| --- | --- | --- | --- |
| `b'\x2a'` -> `42` | 1 of 3 | `None` (`AttributeError: 'int' object has no attribute 'get'`) | exited |
| `"plan"` (str) | 1 of 3 | `None` (`'str' object has no attribute 'get'`) | exited |
| `["plan"]` (list) | 1 of 3 | `None` (`'list' ...`) | exited |
| nil | 1 of 3 | `None` (`'NoneType' ...`) | exited |
| `true` (bool) | 1 of 3 | `None` (`'bool' ...`) | exited |
| `1.5` (float) | 1 of 3 | `None` (`'float' ...`) | exited |
| `b'\xc1'` - does not decode | **3 of 3** | `0` | running |

I took your second option - reply `malformed_request:` and `continue` - for the reason you gave. It also avoids two things the first option would have shipped: `internal_error:` blames the sidecar for the peer's payload, and its detail (`AttributeError: 'int' object has no attribute 'get'`) is not something a client author can act on. The reply now names the type that arrived:

```
{"error": "malformed_request:expected a msgpack map, got int"}
```

One addition to your note: the guard is also what makes the `api_token` check the module's own fork notes suggest safe to add, since nothing reads a key off `request` until it is a map. I said so in the fork note itself.

Scope: the sibling `data = request.get("data") or {}` is left alone deliberately. `data` is only read *inside* the dispatch `try`, so a non-map value there was already answered - measured 3 of 3 replies with `main()` returning `0` - and pinning that is one of the two tests that pass on both sides of this commit, bounding the check to the one read that happens where no guard can answer it.

7 new tests, parametrized over the six shapes plus the shared-error-class case: **7 failed / 25 passed** against the previous commit, 32 passed after. `docs/policies/moveit2.md` no longer says only "an undecodable request", and there is a new before/after artifact in the description where a stray peer sends `b'\x2a'` mid-session and the operator's next task survives it.

yinsong1986 (approved):
## Summary

Guards every `moveit_py` interaction in the MoveIt2 sidecar's `_plan` (start-state read, goal set, trajectory serialisation) so each stage reports a structured `{"success": false, "status": ...}` instead of raising, wraps REP-loop dispatch so a raising handler is answered with the existing `{"error": ...}` envelope, and rejects a payload that decodes to a non-map with the same `malformed_request:` class as undecodable bytes - closing the gap the previous review flagged, where `request.get("endpoint")` sat between the two guards. Together these make the one-reply-per-request guarantee the module docstring and `docs/policies/moveit2.md` now state actually hold: a single bad request (or one stray byte from any peer) no longer unwinds `main` and takes planning down for every client.

## What's good

- The prior `[MUST FIX]` (non-map payload read between the guards) is resolved with an `isinstance(request, dict)` check that replies in the `malformed_request:` class rather than blaming the sidecar with `internal_error:`, pinned by parametrised tests over six non-map payload shapes plus a shared-error-class test.
- The four previously-reporting statuses are pinned byte-identical (`test_plan_already_guarded_statuses_are_unchanged`), so nothing operators or forks match on moves; all new wire-visible statuses are additive over paths that previously crashed the process.
- The broad `except Exception` guards carry the same `noqa: BLE001` + rationale as the `policies/groot/server_wrapper.py` precedent, and each is a recovery path that logs via `logger.exception` and reports the failing stage structurally - consistent with AGENTS.md's no-silent-defaults contract.
- Loop-level tests assert replies-sent == requests-received against the real `main()` loop, including a response msgpack cannot pack; docstring completeness of the status kinds is itself pinned by a test.
- Changelog fragment (`changelog.d/2440-...`) and wire-protocol docs updated in the same PR; `ruff check` / `ruff format --check` clean on the changed files, and the moveit2 test selection passes on the head.

PR #2441, opened 2026-08-18
fix(rendering): state the scale of an sRGB decode input instead of guessing it

cagataycali:
## Problem

`srgb_to_linear` is the entry point for every piece of light arithmetic in the hybrid-render pipeline (the shadow-catcher multiply, the optional linear seam blend, and the key-light estimate). It accepts **two scales** - byte codes in `[0, 255]` and unit floats in `[0, 1]` - and it decided between them from `dtype == np.uint8` alone:

```python
x = np.asarray(rgb)
if x.dtype == np.uint8:
    x = x.astype(np.float32) / 255.0
else:
    x = x.astype(np.float32)     # <- assumed to be [0, 1] already
x = np.clip(x, 0.0, 1.0)
```

So every *other* integer spelling of the same byte data fell into the `else` branch, was read as unit linear light, and was clipped. The parameter is annotated `npt.ArrayLike` and the docstring admits values "in `[0, 255]`", so the plainest spelling the annotation invites is already affected:

| input | decoded |
|---|---|
| `np.array([0, 64, 128, 192, 255], np.uint8)` | `[0.0, 0.0513, 0.2159, 0.5271, 1.0]` |
| the same codes as `uint16` / `int32` / `int16` / `int64` | `[0.0, 1.0, 1.0, 1.0, 1.0]` |
| `srgb_to_linear([0, 64, 128, 192, 255])` (a list is `int64`) | `[0.0, 1.0, 1.0, 1.0, 1.0]` |

Every code above 1 became pure white, with nothing raised, logged or returned to say so.

`derive_key_light` checks its environment map's *shape* and passes its *scale* straight into that decode, so the saturation reaches a user-visible result. On one map - a small bright warm window plus a broad dim cool wall glow - it stops reading the brightest region and instead reports the solid-angle centroid of everything that is not pure black:

| environment map spelled as | azimuth | elevation | colour |
|---|---|---|---|
| `uint8` | 14.8 deg | 56.0 deg | `(1.000, 0.746, 0.409)` |
| `uint16`, same codes | **70.2 deg** | **73.8 deg** | **`(1.000, 1.000, 1.000)`** |

A 55.4 deg azimuth error and the warm chromaticity gone, both reported as success. A 16-bit capture, an array a caller widened, or a hand-built map are all ordinary ways to arrive there.

## Change

Refuse an integer dtype other than `uint8` rather than resolving the ambiguity by assumption. A `uint16` array genuinely could be `[0, 255]` data or `[0, 65535]` data or already-unit values, and guessing is the present defect - so the message names both accepted spellings and the conversion for each:

```
srgb_to_linear: cannot tell what scale a uint16 array is on - sRGB input is either uint8 in
[0, 255] or float in [0, 1], and an integer array could be either. Pass .astype(np.uint8) for
byte-encoded values, or divide by the encoding maximum (255 for 8-bit, 65535 for 16-bit) and
pass float.
```

The guard belongs on this helper only, and that boundary is deliberate: `linear_to_srgb` and `relative_luminance` take unit linear light and nothing else, so an integer `0`/`1` is unambiguous for them and is still accepted. `srgb_to_linear` is the one helper with two scales, so it is the only one that can be handed a value whose scale it cannot know.

Both existing call paths already pass `uint8` (the compositor coerces `fg_rgb`/`bg_rgb` explicitly; `render_environment_map` returns `uint8`), so no shipped caller changes behaviour - the refusal is the only new outcome. `color.py`'s module docstring, which the package treats as the written-down statement of the colour-pipeline contract, now records the scale decision beside the space decision.

## Verification

Both trees on MuJoCo 3.5.0, `MUJOCO_GL=egl`.

Pre-fix split, new tests copied onto `upstream/main` (`74d078ab`): **6 failed / 29 passed**, every behavioural failure quoting the measured values (the tests use their own fixture map, so the numbers differ from the artifact's) -

```
the same codes as uint16 decoded to [0.0, 1.0, 1.0, 1.0, 1.0] instead of
[0.0, 0.0513, 0.2159, 0.5271, 1.0]: every code above 1 was read as unit linear
light and clipped to white.

the same map as uint16 derived azimuth=50.1 deg elevation=68.6 deg color=(1.0, 1.0, 1.0),
not azimuth=4.8 deg elevation=47.6 deg color=(1.0, 0.808, 0.036).
```

After: **35 passed**. Two of the 29 pre-fix passes are the controls that matter - both documented spellings (plus a `bool` mask) stay accepted, and the two unit-only helpers still take an integer `0`/`1`, which fails if the guard is ever widened to all three.

Blast radius - all of `tests/rendering`, every test naming the changed symbols, and the repo-wide docstring / ASCII / changelog / refusal guards (25 files):

|---|---|---|---|---|
| this branch | 0 | 1148 | 4 | 1152 |
| `upstream/main` + the new tests | 6 | 1142 | 4 | 1152 |

`1142 + 6 = 1148`, the base-only failures are exactly the 6 pre-fix ids, branch-only is empty, and the selection carries no pre-existing failures. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 20 errors on both trees, none in the changed files.

## Artifact

One capture script per tree, same environment map, same camera. It derives the key light from each spelling and lights the arm with it - and where the library refuses, it parses the remedy out of the refusal and applies it, so nothing is hard-coded per side.

sRGB input scale

Top row is `upstream/main`: two different lights from one map, both accepted. Bottom row is this change: the ambiguous spelling is refused and the refusal's own remedy recovers the correct light. The two renders differ by **16.38** mean absolute level before (64% of pixels by more than 8) and **0.0001** after, while the `uint8` render is unchanged by this commit (**0.00015** mean absolute level) - so the refusal is the only new behaviour.

### DIFF
--- a/changelog.d/2441-srgb-decode-states-its-scale.md
+++ b/changelog.d/2441-srgb-decode-states-its-scale.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **rendering**: `srgb_to_linear` no longer guesses the scale of an integer array. It accepts two scales - byte codes in `[0, 255]` (`uint8`) and unit floats in `[0, 1]` - and decided between them from `dtype == uint8` alone, so every other integer spelling of the same byte data (a `uint16` capture, an `int32` array, or the plain Python list its `ArrayLike` annotation invites) took the float branch and was clipped, turning every code above 1 into pure white with nothing reported. Such an array is now refused with a message naming both accepted spellings and the conversion for each. `derive_key_light`, which passes its caller's environment map straight into the decode, inherits the refusal instead of reporting a key light derived from a saturated map.
--- a/strands_robots/rendering/color.py
+++ b/strands_robots/rendering/color.py
@@ -31,6 +31,18 @@ go through the helpers here:
   shadow is a multiplicative ratio of received light, which is only a
   multiplication in linear space.
 
+**Which scale a value is on**
+
+:func:`srgb_to_linear` is the one helper here that accepts two scales -- bytes
+in ``[0, 255]`` or floats in ``[0, 1]`` -- so it is the only one that can be
+handed a value whose scale it cannot know. It reads ``uint8`` as bytes and
+floats as unit values, and *refuses* any other integer dtype rather than
+guessing: a ``uint16`` array (or a plain Python list of ints) could already be
+the unit domain or could be byte-encoded, and reading byte codes as unit floats
+clips every code above 1 to pure white. :func:`linear_to_srgb` and
+:func:`relative_luminance` take unit linear light only, so an integer ``0``/``1``
+there is unambiguous and is accepted.
+
 Transfer function: IEC 61966-2-1 (sRGB), the piecewise curve -- not a bare 2.2
 power -- so round trips are exact to float precision. Luminance weights are
 Rec. 709 / sRGB primaries.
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Replaces `srgb_to_linear`'s dtype-based scale guess with a loud refusal: an integer array of any dtype other than `uint8` now raises `ValueError` naming both accepted spellings (`uint8` in `[0, 255]`, float in `[0, 1]`) and the conversion for each, instead of being read as unit floats and clipped to white above code 1. `derive_key_light` inherits the refusal, so a widened/16-bit environment map can no longer silently produce a wrong key light. Both shipped call paths already pass `uint8` (`compositor.py:458` coerces `fg_rgb` explicitly; `render_environment_map` returns `uint8` at `ibl.py:181`; the `examples/isaac_gs/scene.py` map is `np.array(Image.open(...))`, i.e. `uint8`), so the refusal is the only new behaviour and it is loud with an actionable message -- exactly the AGENTS.md "raise on fatal errors / no silent defaults" pattern rather than a silent breaking change. Verified locally on the head SHA: `tests/rendering/` reports 364 passed, 1 skipped, including the new premise tests that quote measured values pre-fix.

## What's good

- Guard boundary is deliberate and pinned: the two unit-only helpers (`linear_to_srgb`, `relative_luminance`) still accept integer `0`/`1`, and a test fails if the guard is ever widened to them.
- Controls pin both documented spellings (plus `bool` masks) staying accepted, so the contract is not narrowed to one scale.
- Changelog fragment present (`changelog.d/2441-srgb-decode-states-its-scale.md`), module docstring records the scale decision beside the space decision, and the error message is plain ASCII.

PR #2443, opened 2026-08-18
fix(rendering): check derive_key_light's search-region flag instead of reading it by truthiness

cagataycali:
## Problem

`derive_key_light(env_map, brightest_fraction=..., upper_hemisphere=...)` estimates the dominant light direction and colour of an environment map, and `upper_hemisphere` selects **which region of the map is searched**. Its sibling in the same signature, `brightest_fraction`, is checked four ways (a number, not a `bool`, finite, inside `(0, 1]`). The flag had no check at all: it was read by truthiness at the region selection.

Two things followed from that.

**A value that reads as *off* selected the *on* posture.** Every spelling an operator reaches for when opting out is a truthy string, so `upper_hemisphere="false"` (and `"no"`, `"off"`, `"0"`) searched *above* the horizon, the region the value asks to leave out. Meanwhile an undeclared falsy value (`None`, `0`, `""`, `[]`) searched the full sphere, which is the footgun this argument's own docstring warns about and which `test_bright_floor_does_not_hijack_the_key_light` exists to keep callers out of: a bright surface under the bake point out-weighs the real light sources and aims the key light from underneath.

Measured on a captured-room map whose modest warm window (230/texel) sits above a broad sunlit counter (205/texel, far more solid angle), the two outcomes are **85.6 deg of elevation apart and disagree on colour as well** - and nothing distinguished them:

| `upper_hemisphere=` | before | region searched | elevation | colour |
|---|---|---|---|---|
| `True` (documented default) | accepted | above the horizon | `+35.7 deg` | `(1.00, 0.63, 0.23)` warm window |
| `False` (documented opt-out) | accepted | full sphere | `-49.9 deg` | `(0.77, 0.81, 1.00)` cool bounce |
| `"false"` `"False"` `"no"` `"off"` `"0"` | accepted | **above the horizon** | `+35.7 deg` | warm |
| `None` `0` `""` `[]` `0.0` | accepted | **full sphere** | `-49.9 deg` | cool |
| `1.0`, `nan` | accepted | above the horizon | `+35.7 deg` | warm |

**The empty-region refusal branched on the same truthiness.** On a map lit only by a bounce floor, a caller who wrote `upper_hemisphere="false"` was told the map is black *above the horizon* - a region they had asked not to search - and advised to `pass upper_hemisphere=False`, which is what they believed they had passed. A real `False` succeeded on the same map.

## Change

`upper_hemisphere` is now checked against `boolean_flag_error`, the shared domain for a flag that selects a posture rather than scaling a quantity, placed beside the existing `brightest_fraction` guards. Its docstring says exactly why: *"a truthy spelling of off, such as `'false'`, would otherwise select the opposite posture from the one it reads as"*.

The check is one line, and it closes both observables: with the flag validated, the empty-region branch that names the above-horizon region is reachable only when the search really was restricted, where its advice is actionable.

The shared domain rather than a local `isinstance(x, bool)` for two reasons. The spellings this refuses must not diverge from the ones its sibling call sites refuse - `actuate_robot`, `start_recording`, `lerobot_teleoperate`, `provision_robot` and `push_to_hub` all use it. And it accepts numpy booleans, which a hand-rolled `isinstance` turns away (`isinstance(np.True_, bool)` is `False`) - the four `np.bool_` control tests fail for that shortcut.

`docs/` has no page covering the rendering package, so the contract is stated in the function's own docstring and `Raises:` block, plus a new AGENTS.md section for the rule now shared by five call sites and previously written down nowhere. Its `Pinned by` list names five test files, each of which parametrizes over `boolean_flag_error` itself rather than a copied spelling list, so a spelling added to the shared domain is covered without an edit.

## Tests

`tests/rendering/test_key_light_posture_flag_domain.py`, following the naming of the four existing `*_flag_domain.py` suites. **15 failed / 9 passed against `upstream/main`, 24 passed after.** No new symbol is imported, so every pre-fix failure is behavioural and carries a number:

```
AssertionError: upper_hemisphere='false' reads as off, yet it searched above the
horizon (elevation 40.9 deg, the same region upper_hemisphere=True searches); the
full sphere it asks for derives -54.0 deg, 94.9 deg away
```

The 9 that pass on both trees are the controls, and each fails for a plausible wrong fix:

- `True`, `False`, `np.True_`, `np.False_` are all still accepted (fails for a hand-rolled `isinstance(x, bool)`, and for making the guard over-strict).
- The two declared spellings still select the two regions, and omitting the argument still equals `upper_hemisphere=True`.
- The full-sphere advice **still reaches a genuinely restricted search** (fails if the fix removes that branch rather than making it unreachable from a non-boolean).
- The `brightest_fraction` and `env_map` shape refusals are asserted byte-for-byte unchanged.

The root-cause test parametrizes over `boolean_flag_error` as the premise authority, so the spelling list cannot drift from the domain.

## Verification

Lint and types match CI scope: `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 20 errors on the branch and 20 on a pristine `upstream/main`, none in the changed files. `scripts/assemble_changelog.py --check` OK.

Blast radius of 126 files (all of `tests/rendering`, every test touching `derive_key_light` / `upper_hemisphere` / `boolean_flag_error` / a posture flag / AGENTS.md, plus the repo-wide docstring, string-hygiene, dependency and changelog guards): **branch 4 failed / 4512 passed, base 4497 passed / 19 failed, 4552 collected on both.** Branch-only failures: none. Base-only: exactly the 15 pre-fix failures, zero foreign. The 4 shared failures are pre-existing and environmental here (3 `ModuleNotFoundError: device_connect_agent_tools`, a declared dependency, plus one mesh `_safety_publishers_lock` case).

Re-measured after merging `upstream/main` (#2441 landed mid-review and touches `srgb_to_linear`, which this function calls): split still 15/9 -> 24, base elevation still `-49.9 deg`, and 848 tests pass across the overlapping suites.

## Artifact

A UR5e lit by the key light this function derives, from the environment map above. The capture script passes `upper_hemisphere=None` - what a caller writes when they mean "leave it at the default" - then reads whatever refusal its tree gives and applies **that message's own remedy**; nothing is hardcoded per side. Same map, same robot, same camera, same headlight exposure in both columns; only `strands_robots` differs.

Before, the undeclared value is accepted, the full sphere is searched, and the arm is lit from `-49.9 deg` by the cool counter bounce (frame mean 39.97). After, it is refused by name, the supplied default finds the window at `+35.7 deg` in warm light (frame mean 60.49). 75.6% of pixels differ.

derive_key_light posture flag


yinsong1986 (approved):
## Summary

Adds the missing domain check on `derive_key_light`'s `upper_hemisphere` flag: it is now validated against the shared `boolean_flag_error` guard (which accepts `np.bool_` via `is_boolean`) beside the existing `brightest_fraction` checks, instead of being read by truthiness at the region selection. This closes both observables the description documents - a truthy spelling of off (`"false"`, `"no"`, `"off"`, `"0"`) restricting the search it asks to widen, and an undeclared falsy value silently selecting the full-sphere search the docstring itself warns about - and makes the empty-region refusal's `pass upper_hemisphere=False` advice reachable only when the search really was restricted. Verified on the head SHA: the 24 new tests in `tests/rendering/test_key_light_posture_flag_domain.py` pass, the full `tests/rendering` suite is 388 passed / 1 skipped, and the pre-existing `brightest_fraction` / `env_map` refusals are pinned byte-for-byte unchanged by controls.

## What's good

- Uses the shared domain (already the pattern at `actuate_robot`, recording, teleoperate, provisioning) rather than a hand-rolled `isinstance(x, bool)`, so the accepted spellings cannot diverge across call sites and numpy booleans stay accepted - and the `np.True_`/`np.False_` controls pin exactly that shortcut.
- The new AGENTS.md section's five `Pinned by` test files all exist in the tree, and the tests parametrize over `boolean_flag_error` itself as the premise authority rather than a copied spelling list.
- Changelog fragment follows the `changelog.d/<pr-number>-<slug>.md` convention (step 3); docstring and `Raises:` updated in the same diff; plain-ASCII throughout the changed files.
- The new refusal raises an explicit `ValueError` naming the parameter, so previously-accepted junk values fail loudly rather than silently changing posture - the right direction for a behaviour change.

PR #2444, opened 2026-08-18
docs(ros2): document joint_limits with the key spelling the bound is matched against

cagataycali:
## Problem

`joint_limits` is the inbound-command safety bound on the one surface that drives a physical arm: if any commanded joint is outside its declared range the whole command is rejected. The bound is applied by looking the commanded joint's name up in the mapping (`RosTelemetryBase._command_action`), so the keys have to be spelled the way the names arrive on the wire - the `<motor>.pos` names both hardware bridges publish in `joint_states`.

Every prose declaration of the parameter instead wrote `{motor: (min, max)}`. A key spelled that way is not an error: `dict.get` returns `None`, the joint is treated as having no declared bound, and the command is forwarded to `send_action` unchanged. Measured against the shared parser with wire names `["shoulder_pan.pos", "elbow.pos"]` and positions `[99.0, 0.5]`:

| `joint_limits` as spelled by | outcome |
| --- | --- |
| the prose, `{"shoulder_pan": (-3.14, 3.14)}` | **forwarded** `{'shoulder_pan.pos': 99.0, 'elbow.pos': 0.5}` - no exception, no log line |
| the RTPS guide's runnable example, `{"shoulder_pan.pos": (-3.14, 3.14)}` | rejected whole, with the range warning |
| no limits at all (control) | forwarded, identically to row 1 |

So a reader who copied the documented spelling got behaviour indistinguishable from passing no limits - silent in the worst direction for a safety bound. `docs/rtps-integration.md` contradicted itself inside one section: the prose at line 207 and its own example 16 lines below. The code, the example and the rclpy bridge's own tests (`{"shoulder_pan.pos": ...}`, `{"j0.pos": ...}`) already agreed; only the prose disagreed.

## Change

Documentation only - no behaviour change. The 13 sites that declare the parameter (three guides, both bridge class docstrings, both `Robot` entry points, the shared base's validator and command parser, plus one inline comment and two test docstrings) now name the published vocabulary, `{"<motor>.pos": (min, max)}`, and state two things they did not: what the keys are matched against, and that a key naming no commanded joint constrains nothing.

The alternative - making the lookup also accept a bare motor name - was not taken. It would give one concept two spellings at a safety boundary, and the key the bound must use is already fixed by what the bridges publish and what `send_action` consumes.

## Verification

`tests/test_joint_limits_key_vocabulary.py` grades the documented spelling rather than asserting a literal. It collects every `joint_limits` mapping from `docs/**/*.md`, `README.md` and the package sources, resolves each for a real motor, and drives it through the real `_command_action` against wire names taken from the real publish path (`Robot._publish_ros_telemetry`), so the vocabulary it grades against is the one the bridges actually advertise.

Against `upstream/main`: **2 failed / 4 passed**; after: **6 passed**. Both failures name all 12 offending sites with the consequence, e.g.

```
docs/security.md:141: documented key 'motor' resolves to 'shoulder_pan',
  which does not bound shoulder_pan - the command is forwarded unchanged
```

The four tests that pass on both trees are the controls: a key that names no commanded joint constrains nothing, a joint with no declared bound stays unconstrained, a limit keyed on a published name still rejects the whole command, and the published names carry the motor they report (non-vacuity for the vocabulary helper itself). A `_MINIMUM_DOCUMENTED_SITES` premise makes a sweep that stops reaching the guides fail loudly instead of reporting clean.

Gate, 48 files covering both bridges, the shared base, the docs guards and the repo-wide string/changelog/docstring guards, same selection on both trees: branch **1888 passed, 0 failed**; `upstream/main` with the new file copied in **1886 passed, 2 failed** (`1886 + 2 = 1888`, same collected total, no branch-only failure). `ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` 20 errors on both trees, none in the changed files.

## Artifact

One capture script, one command sequence, one MuJoCo SO-101 standing in for the physical follower. Each side reads the `joint_limits` key spelling from **its own** `docs/security.md`, declares `shoulder_pan` bounded to +/-1.0 rad, and sweeps an inbound `joint_command` to 1.8 rad through the shared `RosTelemetryBase` command path. Only `strands_robots` differs between the panels.

joint_limits documented key spelling

| | documented key | commands accepted | peak delivered to the arm |
| --- | --- | --- | --- |
| before | `'shoulder_pan'` | 48 / 48 | **1.8000 rad** (0.8000 past the declared bound) |
| after | `'shoulder_pan.pos'` | 27 / 48 | 0.9957 rad (inside the bound) |

The panels are pixel-identical at frame 0 (mean abs 0.0000) and 30.83 apart at the end, so they diverge only where the declared bound starts to matter; 47/47 frame transitions move. [MP4](https://raw.githubusercontent.com/cagataycali/robots/media-joint-limits-key-vocabulary/joint_limits.mp4), [still](https://raw.githubusercontent.com/cagataycali/robots/media-joint-limits-key-vocabulary/joint_limits.png).


yinsong1986 (approved):
## Summary

Documentation-only correction of the `joint_limits` key vocabulary: the inbound-command clamp in `RosTelemetryBase._command_action` looks each commanded joint up by its wire name (`<motor>.pos`), so a mapping keyed on the bare motor name silently matches nothing and leaves the arm exactly as unbounded as passing no limits. The 13 prose sites (three guides, both bridge docstrings, both `Robot` entry points, the shared validator/parser) now state the published spelling, what the keys are matched against, and that an unmatched key constrains nothing - which agrees with the code path at `strands_robots/ros_telemetry.py` (`joint_limits.get(name)` against the message's own `name` entries) and with the pre-existing runnable example and bridge tests. No behavior change, no API or wire change.

## What's good

- Verified locally: the new `tests/test_joint_limits_key_vocabulary.py` fails 2/6 on `upstream/main` (naming all offending doc sites) and passes 6/6 on the branch - the fails-on-pre-fix regression-pin convention, applied to prose.
- The test derives the graded vocabulary from the real publish path (`Robot._publish_ros_telemetry`) and drives documented spellings through the real `_command_action`, so it cannot drift into asserting a literal; the `_MINIMUM_DOCUMENTED_SITES` premise keeps the sweep honest.
- Changelog fragment present (`changelog.d/2444-joint-limits-key-vocabulary.md`), no direct `CHANGELOG.md` edit; `ruff check` / `ruff format --check` clean on the changed files; no new non-ASCII or host paths in the diff hunks.
- The alternative of accepting bare motor names in the lookup was correctly declined - one spelling at a safety boundary, fixed by what the bridges publish.

PR #2445, opened 2026-08-18
fix(mesh): scope the drive guarantees each bridge actually keeps

cagataycali:
## Problem

`RosbridgeRobot.drive` opened its docstring with:

> **Fleet-standard contract, shared with `RosBridgedRobot.drive` and `RtpsRobot.drive`.** Inputs are validated before any side effect; velocities are then clamped to the constructor limits. Every timed or multi-message non-zero command is followed by a single zero Twist - even if the main publish failed - so a timed drive cannot leave the robot with a live velocity. A bare single-shot command latches until `stop()`, like any raw cmd_vel publish.

and `docs/rosbridge-integration.md` listed the same semantics under a **"Fleet drive contract"** heading. Four guarantees, all under one "shared with" label. Driving the three bridges through a recorder shows two of them are carried by `RosbridgeRobot` alone:

| guarantee | `RosBridgedRobot` | `RtpsRobot` | `RosbridgeRobot` |
|---|---|---|---|
| validated before any side effect | yes | yes | yes |
| bare single-shot latches | yes | yes | yes |
| velocities clamped to constructor limits | **no** | **no** | yes |
| hold beyond a ceiling refused | **no** | **no** | yes |
| timed command followed by a zero Twist | **no** | **no** | yes |

Only `RosbridgeRobot` accepts `max_linear` / `max_angular` / `max_duration`; the other two take `publish_rate` alone. Concretely, with all three built at `publish_rate=10.0`:

```
drive(linear=99.0, angular=-42.0, count=1)      # wire linear.x / angular.z
  RosBridgedRobot   99.0 / -42.0
  RtpsRobot         99.0 / -42.0
  RosbridgeRobot     2.0 /  -1.0      <- clamped

drive(linear=1.0, duration=5.0)                 # published bursts, last Twist
  RosBridgedRobot   [50]      last linear.x = 1.0
  RtpsRobot         [50]      last linear.x = 1.0
  RosbridgeRobot    [50, 1]   last linear.x = 0.0   <- trailing zero

drive(linear=1.0, duration=3600.0)              # a one-hour hold
  RosBridgedRobot   success, 36000 messages
  RtpsRobot         success, 36000 messages
  RosbridgeRobot    error, nothing published
```

The trailing zero is the costly one to mislabel, because it is the guarantee that *a timed drive cannot leave a live velocity behind*. A reader who takes it as fleet-standard expects `RtpsRobot("rover", "/cmd_vel").drive(linear=1.0, duration=5.0)` to self-stop; it publishes fifty Twists at 1.0 m/s and then simply stops publishing, leaving the last one latched in the robot's controller. `tests/mesh/test_drive_command_numeric_domains.py` names that hazard directly - "an agent that learns the contract from one bridge drives the others with it" - and `docs/rtps-integration.md` / `docs/ros2-integration.md` say nothing about latching, so nothing contradicts the label.

The split was already the intended one: that same suite is divided into a `# Cross-bridge parity` section holding the numeric domains, and a `# rosbridge-specific consequences` section holding exactly the clamp, the ceiling and the trailing zero. The prose is what collapsed the two.

## Change

Both surfaces now state each guarantee in the scope it holds in.

The docstring splits into a fleet-standard paragraph (validated inputs, single-shot latch) and a bridge-specific one, and the second names what the other two bridges do instead, so a reader who is told a guarantee is local also learns the alternative:

> `RosBridgedRobot.drive` and `RtpsRobot.drive` carry none of the three: they accept no velocity or duration ceiling, publish the requested burst unclamped, and stop publishing without a trailing zero, so a timed drive there leaves the last velocity latched in the robot's controller.

The guide's heading becomes `### Drive contract`, its intro states which two guarantees are fleet-standard, and each bridge-specific bullet is marked `(this bridge only)` with the alternative spelled out.

No behaviour changes - `drive`, `stop`, the clamps and the trailing zero are untouched.

### Out of scope

Whether the other two bridges *should* clamp, take a ceiling and append a trailing zero is a separate call: it adds three constructor parameters to two classes and changes what reaches the wire for existing callers. This PR makes the documented claim match the code rather than picking that.

## Tests

`tests/mesh/test_drive_contract_fleet_scope.py` measures all five guarantees on all three bridges and grades the prose against the measurement, so neither half is pinned to a hand-kept list. Against `upstream/main`: **12 failed, 4 passed**; after: **16 passed**.

The headline reads whatever prose is present, so it fails on the real text rather than on a missing marker:

```
AssertionError: this paragraph of RosbridgeRobot.drive claims a shared contract
and then states 'trailing zero Twist' ('zero Twist'), 'velocity clamp'
('clamped'), which no other bridge carries: Fleet-standard contract, shared
with :meth:`RosBridgedRobot.drive` and :meth:`RtpsRobot.drive`. ...
```

The 4 tests that pass on **both** trees are the controls, and they are the ones that show nothing behavioural moved: the two premises, the measured fleet-wide/bridge-specific split, and the wire-level consequence (`{RosBridgedRobot: 1.0, RtpsRobot: 1.0, RosbridgeRobot: 0.0}` as the last commanded velocity of a 5 s hold). The split assertion is deliberately two-directional - a bridge that later gains the trailing stop fails it, which is the signal that the guarantee became fleet-wide and the two places scoping it have to say so.

## Verification

Measured at `08f596d9` against `upstream/main` `f4675b15`.

- 264-file blast radius (all of `tests/mesh`, every test naming one of the three bridges or the rosbridge guide, and the repo-wide docs/docstring/changelog guards), run sequentially on both trees:
  - branch **21 failed / 6306 passed / 61 skipped / 24 errors**
  - base **33 failed / 6294 passed / 61 skipped / 24 errors**
  - `33 - 21 = 6306 - 6294 = 12`, both collected **6412**; branch-only failures **empty**; base-only exactly the 12 pre-fix ids. The 45 shared failing ids are pre-existing `awsiot`-absent `tests/mesh/test_iot_*`.
- `mypy strands_robots tests tests_integ`: **20 == 20**, branch-only empty, none in the changed files.

## Artifact

The Twists each bridge actually publishes for `drive(linear=1.0, duration=5.0)`, above the claim graded per surface. Both panels come from one capture script run against a worktree of `upstream/main` and against this branch; each side reads its **own** tree's docstring and guide, and the measurement row is identical in both, which is the control for "no behaviour changed". Wrong claims: 5 before (2 in the docstring, 3 in the guide), 0 after.

drive contract scope

### DIFF
--- a/changelog.d/2445-drive-contract-fleet-scope.md
+++ b/changelog.d/2445-drive-contract-fleet-scope.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **mesh**: `RosbridgeRobot.drive` and the rosbridge integration guide presented this bridge's velocity clamp, `max_duration` ceiling and trailing zero Twist as the fleet-standard drive contract shared with `RosBridgedRobot.drive` and `RtpsRobot.drive`. Only the numeric domains and the single-shot latch are shared; the other two bridges accept no ceilings and publish no trailing zero, so a timed drive there leaves the last velocity latched rather than self-stopping. Both surfaces now scope each guarantee to where it holds, and a guard grades the prose against all three bridges.
--- a/docs/rosbridge-integration.md
+++ b/docs/rosbridge-integration.md
@@ -119,9 +119,15 @@ robot = RosbridgeRobot(
 
 All parameters are optional except `node_name`, `cmd_vel_topic`, and `odom_topic`.
 
-### Fleet drive contract
+### Drive contract
 
-The `drive()` method follows the strands fleet-standard contract:
+`drive()` takes the fleet-standard `(linear, angular, duration, count)` shape,
+and two of its guarantees are fleet-standard too: every value is checked against
+the same numeric domains the [ROS 2](ros2-integration.md) and
+[RTPS](rtps-integration.md) bridges use, and a bare single-shot command latches.
+The velocity clamps, the `max_duration` ceiling and the trailing zero Twist are
+specific to this bridge - the other two accept no ceilings and publish no
+trailing zero, so a timed drive there leaves the last velocity latched.
 
 ```python
 # Direct, programmatic control:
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Documentation-scoping fix: `RosbridgeRobot.drive`'s docstring and the rosbridge integration guide presented four guarantees under one "fleet-standard, shared with `RosBridgedRobot.drive` / `RtpsRobot.drive`" label, when only input validation and the single-shot latch are actually shared; the velocity clamp, `max_duration` ceiling and trailing zero Twist are carried by `RosbridgeRobot` alone. Both prose surfaces now scope each guarantee, and a new suite (`tests/mesh/test_drive_contract_fleet_scope.py`) measures all five guarantees on all three bridges and grades the prose against the measurement in both directions. Verified independently: the production diff is docstring-only (the clamp, ceiling and trailing-zero code paths are untouched, and the other two bridges' `drive` indeed forward unclamped with no trailing zero), the 16 new tests pass when the diff is composed onto the current `main` tip (so no semantic conflict with commits landed since the PR base), `ruff check` / `ruff format --check` are clean on the changed files, and the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention.

## What's good

- The scope guard measures behaviour through the transport seam and grades whatever prose is present, rather than pinning either half to a hand-kept list - a bridge that later gains (or loses) a guarantee fails the split assertion, which is exactly the staleness signal wanted.
- Scope discipline: the PR corrects the claim instead of quietly widening two bridges' behaviour, and says so explicitly under "Out of scope".
- The bridge-specific paragraph names what the other two bridges do *instead* (last velocity stays latched), which is the operationally important half for anyone carrying a timed drive across bridges.

PR #2446, opened 2026-08-18
fix(rendering): check the environment-map resolution knobs before baking with them

cagataycali:
## Problem

`render_environment_map`, `bake_environment_map` and `environment_map_cache_path` take `face_size`, `equi_w` and `equi_h` and validated none of them. `derive_key_light`, in the same module, checks both of its own knobs (`brightest_fraction` four ways, `upper_hemisphere` on the shared flag domain) and validates the map's shape.

**A zero-sized grid produced a map rather than a refusal.** `equi_w=0` returned a `(64, 0, 3)` array of **0 pixels**, reporting success, after paying **six full background renders** - GPU-bound for a `GsplatBackground`. The only refusal the caller ever saw came from the consumer, and it blamed the scene:

```
derive_key_light: no light in the searched region (the map is black above the
horizon -- pass upper_hemisphere=False to search the full sphere, or keep the
caller's default key light).
```

Following that advice fails identically, because the map has no texels at all rather than a dark upper half:

```
derive_key_light: no light in the searched region (the map is black ; keep the
caller's default key light).
```

So the sole remedy on offer was a dead end, and `equi_w` - the knob that was actually wrong - was named nowhere. `environment_map_cache_path` would meanwhile happily name `room_env_0x64_f32_o+0.000_+0.000_+0.600.png` for an image that cannot exist.

**The values that did raise reported numpy internals**, naming neither the parameter nor the function, which for a call with five resolution knobs leaves the caller guessing: `equi_w=-8` -> `negative dimensions are not allowed`; `face_size=0` -> `index -1 is out of bounds for axis 0 with size 0`; `face_size=2.5` -> `arrays used as indices must be of integer (or boolean) type`; `equi_w=nan` -> `arange: cannot compute length`.

**An integral float, which the shared domain accepts, could not be used.** `positive_whole_number_error` deliberately accepts `30.0` "computed from a config float" and a NumPy integer, but `equi_w=32.0` raised a bare `TypeError` from `np.zeros` and `face_size=8.0` an `IndexError`. In the cache path the same value split the key: `2048` and `2048.0` spelled two different filenames for one set of pixels, so a bake already on disk was missed and paid for again.

## Change

All three functions check the three knobs against `positive_whole_number_error` - the shared domain for a pixel count, so what they refuse cannot diverge from the recorders' `width`/`height` or `encode_clip`'s rate - through one private helper, before any render or cache probe. Each then normalizes with `int()`, the way `HybridCompositor` normalizes its own `default_width`/`default_height` ("so a `np.int64` must not leak through").

| case | before | after |
|---|---|---|
| `equi_w=0` | `(64, 0, 3)` map, 6 renders, success | refused in **0** renders, names `equi_w` |
| `equi_h=0` | `(0, 64, 3)` map, 6 renders, success | refused in 0 renders, names `equi_h` |
| `face_size=0` / `-8` / `2.5` / `inf` | numpy internals | names the parameter and the function |
| `equi_w=True` | `an integer is required` | refused as a bool |
| `equi_w=32.0` | `TypeError` from `np.zeros` | accepted, normalized to `32` |
| cache key `2048` vs `2048.0` | two filenames | one filename |
| valid resolution | works | **byte-identical** |

`znear`/`zfar` are deliberately untouched: they are handed to the background, whose own contract governs them, and no consequence of a bad value is observable in this module.

Only the *domain* is checked, not the *quality* a resolution buys. A very coarse cube face is still a usable one and there is no boundary to pin a floor to: the reprojection scales by `face_size - 1`, so 1, 2 and 3 each resolve almost nothing within a face (measured on a background carrying a gradient inside every face: 4, 4 and 5 distinct colours in the whole map, growing smoothly from 4 upward). `0` cannot produce a map at all, so it is refused; `1` can, so it is accepted, and two tests pin that no floor was invented for `equi_w`/`equi_h` either.

## Tests

`tests/rendering/test_environment_map_resolution_domain.py`, parametrized over `positive_whole_number_error` itself rather than a copied value list, so a value added to the shared domain is covered without an edit.

Against `upstream/main`: **88 failed / 3 passed** -> **91 passed**. The headline failure reports the measured consequence rather than a missing raise:

```
AssertionError: equi_w=0 returned a map of shape (16, 0, 3) (0 pixels) after 6
background renders instead of naming equi_w. The caller's only refusal blames
the scene: "...the map is black above the horizon -- pass upper_hemisphere=False
..." -- and following it fails again: ...
```

The 3 that pass on both trees are the controls: a usable resolution still bakes (6 renders, correct shape, non-empty file), and a one-texel `equi_w` / `equi_h` axis is still accepted - the last two fail if anyone extends the refusal into a coarseness floor.

## Verification

Measured at `72d81fc6` against `upstream/main` `0640c9c9`, same interpreter, 104-file selection (all of `tests/rendering`, every test naming the changed symbols or `positive_whole_number_error`, plus the repo-wide docstring / string / changelog / dependency guards):

- branch **1 failed / 5212 passed / 33 skipped**, base **89 failed / 5124 passed / 33 skipped**
- `5124 + 89 == 5212 + 1 == 5213` collected on both; **branch-only failures: none**; base-only 88, all in the new file, zero foreign
- the single shared failure is pre-existing and unrelated (`tests/mesh/test_safety_timing_and_replay_defenses.py::TestResumeLockoutTimingOracleClosed::test_correct_override_code_clears_lockout`, `AttributeError: 'Mesh' object has no attribute '_safety_publishers_lock'`)
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 20 errors on both trees, none in the changed files

## Artifact

One capture script run against `upstream/main` and against this branch, same fake room background, only `strands_robots` differing. The left panel is the control: at a usable resolution both trees derive the identical key light (azimuth `+0.10 deg`, elevation `+10.72 deg`, colour `(1.0, 0.634, 0.222)`) and the MuJoCo render of a `ur5e` lit by it differs by a mean of `0.000123` levels.

environment-map resolution domain

The background is a stand-in for a `GsplatBackground` (`gsplat` ships in the `sim-gs` extra); everything else in the capture is shipped code.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1421,6 +1421,39 @@ Corrections from code review that apply to all future contributions:
   `boolean_flag_error` itself rather than a copied spelling list, so a spelling added to
   the shared domain is covered without an edit.
 
+### A resolution knob is validated before the work it sizes
+- **Check the knob that sizes an expensive result before producing the result.** A
+  resolution is caller input like any other, and the numeric domains
+  (`positive_whole_number_error` for a pixel or frame count) are what it is checked
+  against. Checking it late is not merely a worse message: `render_environment_map`
+  paid six full background renders - GPU-bound for a `GsplatBackground` - before
+  returning a `(H, 0, 3)` map for `equi_w=0`, and `bake_environment_map` probed and
+  wrote its cache file before anything looked at the size it was baking.
+- **A zero-sized grid is a resolution mistake, and the consumer will misdiagnose it
+  as a property of the scene.** An empty map is not distinguishable from a dark one
+  by the code that reads it, so `derive_key_light` blamed the background - "the map
+  is black above the horizon" - and advised a search flag. Following that advice
+  fails identically, because a map with no texels has no hemisphere to search, so
+  the only remedy on offer was a dead end and the knob the caller actually got
+  wrong was named nowhere. Refuse at the entry point and the accurate diagnosis is
+  the first one the caller sees.
+- **Check the domain, not the quality.** A resolution the module cannot use is a
+  refusal; a resolution that merely buys a poor result is the caller's call, and
+  the two are worth keeping apart. `face_size` is the example: the equirect
+  reprojection scales by `face_size - 1`, so 1, 2 and 3 each resolve almost
+  nothing within a cube face, and the detail a map carries grows smoothly from
+  there with no boundary to pin a floor to. `0` cannot produce a map at all, so
+  that is refused; `1` can, so it is accepted.
+- **Normalize after checking.** The numeric domains deliberately accept an integral
+  float and a NumPy integer, so the value has to be put through `int()` before it
+  indexes anything - `HybridCompositor` does this for its own `default_width` /
+  `default_height`. Where the value is formatted into a cache key that is not
+  cosmetic: `2048` and `2048.0` spelled two different environment-map filenames for
+  one set of pixels, so a bake already on disk was missed and paid for again.
+- Pinned by `tests/rendering/test_environment_map_resolution_domain.py`, which
+  parametrizes over `positive_whole_number_error` itself rather than a copied value
+  list, so a value added to the shared domain is covered without an edit.
+
 ### One writer per log file
 - **A file two writers share needs one file object, not one path.** Two file objects over
   one path each track their own write offset, so a buffered writer flushes at its offset
--- a/changelog.d/2446-environment-map-resolution-domain.md
+++ b/changelog.d/2446-environment-map-resolution-domain.md
@@ -1,0 +1,10 @@
+### Fixed
+- **rendering**: `render_environment_map`, `bake_environment_map` and
+  `environment_map_cache_path` check `face_size` / `equi_w` / `equi_h` against the
+  shared `positive_whole_number_error` domain before rendering anything, and
+  normalize them with `int()` afterwards. `equi_w=0` previously returned a
+  zero-pixel map after six background renders, leaving `derive_key_light` to
+  report the scene as black and advise a search flag that cannot fix it; an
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Adds up-front domain validation for the three resolution knobs (`face_size`, `equi_w`, `equi_h`) shared by `render_environment_map`, `bake_environment_map` and `environment_map_cache_path`, via one private `_resolution_error` helper that delegates to the shared `positive_whole_number_error` domain, then normalizes with `int()` so an integral float or NumPy integer (which the domain deliberately accepts) can index the grid and cannot spell a second cache filename for pixels already baked. The check sits before the six background renders and before the cache probe/write, so a zero-sized grid is refused in zero renders instead of producing an empty map that `derive_key_light` misdiagnoses as a dark scene. No public signatures change; previously-garbage inputs now raise a loud `ValueError` naming the knob and the function, which is the documented intent rather than a silent behavior change.

## What's good

- Uses the existing shared guard rather than inventing a local one, per AGENTS.md rule 11 (a second-plus caller of `positive_whole_number_error`), so the refusal cannot diverge from the recorders' `width`/`height`.
- Validation ordering verified in the diff: `bake_environment_map` checks before the `out.exists()` short-circuit, so an unusable resolution never writes a file the cache would then serve forever.
- Tests parametrize over the shared domain itself (out-of-domain values asserted as a premise), include controls pinning that no coarseness floor was invented (`face_size=1`, one-texel equirect axes still accepted), and the cache-key dedup test pins the normalization as load-bearing.
- Verified locally at the head SHA: the new file's 91 tests pass, the full `tests/rendering` directory passes (479 passed, 1 skipped), and `ruff check` / `ruff format --check` are clean on the changed files. Changelog fragment present and correctly named; no non-ASCII, no host paths.

PR #2447, opened 2026-08-18
fix(policies/cosmos3): name every column of the DROID midtrain action

cagataycali:
## Problem

A Cosmos 3 embodiment declares the width of the model's action (`raw_action_dim`) and the column names for the model's own unified action (`raw_action_layout` - a 9D effector pose `tx,ty,tz` + the 6D rotation `r0..r5`, plus a 1D `grasp`). `action_layouts` then names the columns per served action space.

`joint_pos` legitimately differs: it is the one space the RoboLab server post-processes, converting the effector pose into joint targets (DROID: 7 joints + gripper = 8). `midtrain` is the unified action served through un-converted, so its column names are the unified action's - which is what `umi`, `av` and `bridge` already declared:

| embodiment | `raw_action_dim` | `midtrain` width | `midtrain == raw_action_layout` |
|---|---|---|---|
| `umi` | 10 | 10 | yes |
| `av` | 9 | 9 | yes |
| `bridge` | 10 | 10 | yes |
| `droid` | 10 | **8** | **no** |

DROID's `midtrain` listed 8 quaternion-form columns (`ee_x,ee_y,ee_z,ee_qx..ee_qw,gripper`) for its 10-wide unified action, and its own attribute docstring said "`midtrain` (10D)". `_action_column_names` takes `layout[:width]` and pads any shortfall with synthesized `action_<i>` names, so the short layout did not shorten the row - it slid every name it carried onto the wrong column and left the surplus unnamed:

```
row      [0.11, 0.22, 0.33,  1.0, 0.0, 0.0, 0.0,  1.0,  0.0,  0.95]
                                                   ^r4         ^grasp
emitted  ee_x  ee_y  ee_z   ee_qx ee_qy ee_qz ee_qw gripper action_8 action_9
```

Column 7 (the rotation component `r4`) was emitted as `gripper`; the grasp left as `action_9`, a key no actuator answers to. `action_mapping` validation compounded it, in both directions:

| `action_mapping` | `droid` before | `bridge` (same 10-wide unified action) |
|---|---|---|
| `{"gripper": "finger_joint1"}` | accepted, drives the actuator with **1.0** (a rotation component) | refused |
| `{"grasp": "finger_joint1"}` | **refused** - the one column a gripper needs cannot be named | accepted, drives it with **0.95** (the grasp) |

## Change

`droid`'s `midtrain` layout is the unified action's columns, so it reuses the same `tx,ty,tz,r0..r5,grasp` list its `raw_action_layout` already used. `joint_pos` is untouched, and the attribute docstring now states that a `midtrain` entry names the same columns as `raw_action_layout` and what a narrower one does. `docs/policies/cosmos3.md` gains an "Action spaces" table naming the columns of each space and which mapping key reaches the gripper in each.

No behaviour changes for a caller who was already mapping a real column: a mapping key that names no column is still refused, now listing the correct ones.

## Rollout

Same unified action (fixed effector pose, `r4` pinned at 1.0, `grasp` ramping 1.0 -> 0.0), same panda, and `action_mapping` puts **each tree's own gripper column** on `panda/actuator8` - the capture script reads that column from the tree it runs in, so nothing is hardcoded per side.

midtrain gripper column

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-cosmos3-midtrain-layout/midtrain-gripper-column.mp4) - [final frame](https://raw.githubusercontent.com/cagataycali/robots/media-cosmos3-midtrain-layout/midtrain-gripper-column.png)

| | before | after |
|---|---|---|
| `midtrain` columns | 8 of 10 | 10 of 10 |
| unnamed columns | `action_8`, `action_9` | none |
| column mapped to `panda/actuator8` | `gripper` (= `r4`) | `grasp` |
| span of the commanded value over the ramp | **0.000** | 1.000 |
| finger gap, start -> end | 40.00 -> **40.00 mm** | 40.00 -> **0.35 mm** |
| frames in which the panel changes | **0 / 24** | 24 / 24 |

The gripper is commanded to close on the rod throughout. Before, the mapped column is `r4` held at 1.0, so the gripper never leaves 40 mm and the closing grasp is discarded as `action_9`; the left panel is genuinely static (max per-frame mean abs delta 0.0012). Both panels are pixel-identical at the first frame (mean abs delta 0.0000), so they differ only where the change bites.

## Tests

`tests/policies/cosmos3/test_midtrain_layout_is_the_unified_action.py` pins the invariant on every embodiment declaring a `midtrain` layout, plus the two consequences: every column of a unified-action row is named, and a mapped `grasp` carries the row's grasp value (both through `_unpack_actions` and through the public `get_actions_sync` path).

Against `main`: **4 failed / 17 passed**, all four naming `droid` - the three sibling embodiments pass, so they are built-in controls. Afterwards 21 pass.

```
test_a_midtrain_layout_names_the_unified_action_columns[droid]
  'droid' serves the unified action under 'midtrain' but names its columns
  ['ee_x', ..., 'gripper'] instead of the unified action's ['tx', ..., 'grasp']
test_every_column_of_a_unified_action_row_is_named[droid]
  'droid' emits 10 columns for its 10-wide unified action, 2 of them unnamed
  (['action_8', 'action_9']): the 'midtrain' layout [...] is 8 entries wide
test_a_mapped_grasp_column_carries_the_rows_grasp_value[droid]
test_the_served_chunk_reaches_the_gripper_through_get_actions[droid]
  ValueError: action_mapping keys ['grasp'] are not in the 'droid'
  'service'-backend action layout.
```

Three controls hold the boundary and pass on both trees: `joint_pos` stays the server's converted layout (deliberately narrower than `raw_action_dim`, so a guard that forced every layout to that width would fail here), `av` declares no `grasp` column, and an `action_mapping` key naming no column is still refused. The existing `test_unpack_unexpected_width_pads_names` covers `joint_pos` and stays green - the padding fallback itself is unchanged.

## Verification

Selection of 70 files (all of `tests/policies/cosmos3`, every test touching the changed symbols, and the repo-wide docs / docstring / changelog / source-string guards), same selection on both trees: branch **2547 passed, 0 failed**; `main` with the new file copied in **2543 passed, 4 failed**. `2543 + 4 == 2547` at the same collected total, no branch-only failure, and the four base-only ids are exactly the pre-fix ones. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` unchanged at 20 pre-existing errors with none in the changed files.


yinsong1986 (approved):
## Summary

Fixes a real column-misalignment bug in the Cosmos 3 DROID embodiment: `droid`'s `midtrain` layout listed 8 quaternion-form names for the model's 10-wide unified action, and since `_action_column_names` slices `layout[:width]` and pads the shortfall with `action_<i>`, every name landed on the wrong column - a rotation component (`r4`) was emitted as `gripper` while the actual grasp left as the unnamed `action_9`, and `action_mapping` validation accepted the misplaced `gripper` while refusing `grasp`. The fix reuses `_RAW_POSE9_GRASP` for `midtrain`, which is exactly what `umi`, `av` and `bridge` already declared, and leaves `joint_pos` (the one legitimately server-converted layout) untouched. I verified the mechanism against the code (`policy.py`'s `_action_column_names` / `_unpack_actions` / mapping validation behave as described), ran the new test file at the head commit (21 passed) and re-ran it with the base `embodiments.py` swapped in (4 failed, exactly the four `[droid]` ids claimed, with the three sibling embodiments passing as controls). The emitted key rename (`ee_x..ee_qw,gripper` -> `tx..r5,grasp`) is the fix itself, not a silent behaviour change: a caller mapping the old `gripper` key under `midtrain` now gets a loud `ValueError` listing the correct columns, and the old names described columns the row never carried.

## What's good

- The regression tests fail on pre-fix code with exactly the claimed failure set (AGENTS.md > Testing > "Pin every reviewed fix with a regression test") and are transport-free, so they run without a server or checkpoint.
- Deliberate boundary controls: `joint_pos` staying narrower than `raw_action_dim` guards against an over-broad width-equality fix, and the `av` no-grasp control keeps the invariant tied to the unified action rather than a hardcoded column list.
- No stale references remain to the removed `_DROID_MIDTRAIN` constant or the old `ee_q*` names anywhere in the tree; the full `tests/policies/cosmos3` directory passes (221 tests) and `ruff check` is clean on the changed files.
- Changelog fragment under `changelog.d/` named with the PR number, per AGENTS.md PR workflow step 3.

PR #2449, opened 2026-08-19
fix(policies/vera): carry the IDM tracker backend into the containerised server

cagataycali:
## Problem

`VeraConfig.server_env()` is the single statement of what the VERA policy server must be configured with. It has **exactly one consumer**: `VeraServerRunner.start` merges it into the child environment wholesale, so it cannot omit a value. `DockerServerRunner` never calls it - it re-enumerates the same vocabulary by hand as `docker run -e` flags - and that hand-written list omitted `VERA_TRACKER_BACKEND`.

Three of the four values the overlay declares do reach the container, each by an appropriate route. The tracker backend reached nothing:

| `server_env()` declares | subprocess mode | docker mode carries it by | carried |
|---|---|---|---|
| `VERA_CKPT_ROOT=/data/vera-ckpts` | env overlay | `-v /data/vera-ckpts:/ckpts:ro` + entrypoint default | yes |
| `VERA_WAN_CKPT_ROOT=/data/wan` | env overlay | `-e VERA_WAN_CKPT_ROOT=/wan` + bind mount | yes |
| `VERA_DYNAMICS_RUN_ID=37oa162u` | env overlay | `-e VERA_DYNAMICS_RUN_ID=37oa162u` | yes |
| `VERA_TRACKER_BACKEND=cotracker` | env overlay | - nothing - | **no** |

`server_env()` is `tracker_backend`'s **only sink anywhere in the package**, so under `server_mode="docker"` the field and its documented `VERA_TRACKER_BACKEND` override were inert, while the same config selected the requested tracker under `server_mode="subprocess"`:

```
tracker_backend='cotracker'  ->  which tracker does the server run?
   subprocess: 'cotracker'      docker: None   (server falls back to its default)
```

Nothing reports it. `docker run` has no unpassed environment variable to object to, so the server simply starts on the default the caller overrode - and the container ships more than one tracker (the image installs AllTracker at `VERA_ALLTRACKER_ROOT` for the eef-delta embodiments), so the override is meaningful precisely there.

`docs/policies/vera.md` tabulates `tracker_backend` -> `VERA_TRACKER_BACKEND` -> "IDM tracker" without qualifying either mode, and `make_server_runner` states that both runners expose the same interface onto the same server.

## Change

`DockerServerRunner._build_run_command` forwards the backend as `-e VERA_TRACKER_BACKEND=<value>`. A backend name is a scalar - it needs no host-to-container path translation - so it is forwarded verbatim, which puts it in the server process's environment exactly as the subprocess path's `env=` overlay does. The container is built to take such an override: the entrypoint's `_set_if_exists` fills a variable only when it finds it empty, so an explicit `-e` wins.

Also documents the carriage rule beside `server_mode`, and states on `DockerServerRunner` that every value `server_env()` declares reaches the container.

Out of scope, deliberately: `algo_config` is likewise absent from the `-e` list, but it holds a **host** filesystem path, so forwarding it verbatim would name a directory that does not exist inside the container. Making it work needs a mount-and-translate decision rather than a forward, unlike a backend name.

## Tests

`tests/policies/vera/test_server_env_reaches_both_server_modes.py`. The headline derives what must be carried **from `server_env()` itself** rather than from a copied list, so a fifth value added to the overlay is graded without touching the test. A value counts as carried when the container command either names it in an `-e` flag or bind-mounts the value it holds - the three real routes above.

Against `upstream/main`: **5 failed / 8 passed**; after: **13 passed**. The headline names the dropped value:

```
AssertionError: server_env() declares these for the server, and the docker run command
neither names them in an -e flag nor mounts their value: {'VERA_TRACKER_BACKEND': 'cotracker'}.
A dropped value means the server starts with a default the caller overrode.
```

The 8 that pass on both trees are the controls, and each fails for a shortcut fix:

- an unset `tracker_backend` is still not forwarded - no empty override is injected stating a choice the config never made;
- `wan_ckpt_root` is still forwarded as its **container** path `/wan`, not the host spelling a scalar fix invites;
- `ckpt_root` is still carried by its mount with no `-e`;
- the run id keeps its direct route;
- the subprocess launch argv is unchanged, and `tracker_backend` appears in neither its flags nor its values (it has never been a server flag - it travels by environment).

## Verification

Measured at `602d2920`. Blast radius: 44 files (all of `tests/policies/vera`, every test naming the changed symbols, and the repo-wide docstring / changelog / source-string / args-docstring guards).

| | branch | base |
|---|---|---|
| passed | 1896 | 1891 |
| failed | 0 | 5 |
| skipped | 1 | 1 |

`1891 + 5 == 1896`, both collected 1897, branch-only failures empty, base-only exactly the 5 pre-fix ids, and no pre-existing failure in the selection. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 with none in the changed files.

## Artifact

Both panels are real output of `DockerServerRunner._build_run_command()` - nothing is started, and the only stub is the `which docker` lookup the existing suite already stubs. The two trees differ only in `strands_robots`, and the subprocess launch argv is **byte-identical** between them.

VERA server_env carriage across both server modes

### DIFF
--- a/changelog.d/2449-vera-tracker-backend-reaches-the-container.md
+++ b/changelog.d/2449-vera-tracker-backend-reaches-the-container.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **policies/vera**: carry the IDM point-tracker backend into the containerised policy server. `VeraConfig.server_env()` is the single statement of what the server must be configured with, and only the subprocess runner reads it - the docker runner re-enumerated the same vocabulary by hand as `-e` flags and omitted `VERA_TRACKER_BACKEND`, whose only sink is that overlay. `server_mode="docker"` therefore ignored `tracker_backend` (and its documented `VERA_TRACKER_BACKEND` override) entirely: `docker run` has no unpassed variable to object to, so the server started on its default tracker while the same config selected the requested one in `server_mode="subprocess"`.
--- a/docs/policies/vera.md
+++ b/docs/policies/vera.md
@@ -126,6 +126,15 @@ server boots with **no network**. See
 `server_mode="subprocess"` launches a local `python -m vera.server...` when VERA
 is installed in the same env.
 
+Either mode configures the server from the same `VeraConfig`, so every value in
+the table below reaches it whichever one runs. The two modes get it there by
+different routes: the subprocess inherits the whole environment overlay, while
+the container takes a scalar (a backend name, a run id) as `-e` and a host path
+as a read-only bind mount forwarded under the container path it was mounted at
+(`ckpt_root` -> `/ckpts`, `wan_ckpt_root` -> `/wan`). A value the container
+command failed to pass would not be reported: `docker run` has nothing to
+object to, so the server would simply start on the default the caller overrode.
+
 ## Configuration
 
 `VeraConfig` maps 1:1 to VERA's server flags and is env-overridable (deploy/CI
--- a/strands_robots/policies/vera/server_runner.py
+++ b/strands_robots/policies/vera/server_runner.py
@@ -227,6 +227,12 @@ class DockerServerRunner:
         config: The :class:`~strands_robots.policies.vera.config.VeraConfig`
             (``docker_image``, ``docker_container_name``, ``docker_gpus``,
             ``ckpt_root``, ports, embodiment).
+
+    Every value :meth:`~strands_robots.policies.vera.config.VeraConfig.server_env`
+    declares reaches the container, so a server configured through the config
+    is configured the same way in either mode: a backend name is forwarded as
+    ``-e`` verbatim, and a host path is bind-mounted and forwarded as the
+    container path it was mounted at.
     """
 
     def __init__(self, config: VeraConfig) -> None:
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes the docker-mode gap where `VeraConfig.tracker_backend` (and its documented `VERA_TRACKER_BACKEND` override) was silently inert: `server_env()` is that field's only sink, `DockerServerRunner` never calls it, and the hand-written `-e` list omitted the variable, so the containerised server always started on its default tracker while the same config selected the requested one in subprocess mode. The fix is a three-line forward in `_build_run_command` that mirrors the existing `dynamics_run_id`/`text_prompt` pattern exactly (truthy guard, single `-e NAME=value` argv token, no shell string), plus a changelog fragment and a docs paragraph stating the carriage rule. No new public API, env-var name, or wire shape is introduced - `VERA_TRACKER_BACKEND` was already the documented vocabulary. The new test derives the must-carry set from `server_env()` itself rather than a copied list, so a future overlay value is graded without editing the test, and the controls pin the shortcut fixes (no empty override injected for an unset backend, host paths still travel by mount/translation, subprocess argv byte-unchanged). Verified locally at the head SHA: 13/13 in the new file, 746/746 across `tests/policies/vera`, ruff check and format clean on the changed files, no non-ASCII or host paths in the additions.

## What's good

- The headline assertion is self-updating: it reads `server_env()` at test time, so the class of bug (overlay value dropped by the docker re-enumeration) stays pinned even as the overlay grows.
- Scope discipline: `algo_config` is explicitly left out with the correct reasoning (it is a host path needing a mount-and-translate decision, not a scalar forward).
- Changelog fragment per AGENTS.md step 3; docstring cross-reference targets kept on one line.

PR #2451, opened 2026-08-19
fix(device_connect): publish each joint's own position, not the slot its id names

cagataycali:
## Problem

`SimulationDeviceDriver._publishState` emits an `observationUpdate` every 100 ms whose `joints` field is documented as `{joint name: position (radians)}`. It built that mapping by indexing `data.qpos` with each joint's MuJoCo **joint id**:

```python
for jname, jid in zip(joint_names, joint_ids):
    joints[jname] = float(data.qpos[jid])
```

A joint's position lives at its qpos **address** (`model.jnt_qposadr[jid]`). The two coincide only while every joint is single-DoF: a floating base occupies seven qpos slots for one joint id, so it shifts every later joint by six. The same loop also emitted a scalar for the free joint itself, which has no scalar position - its qpos is a position plus a quaternion.

Measured over the 62 loadable sim robots:

| | count |
|---|---|
| robots with at least one joint whose id and qpos address differ | **30** |
| mis-addressed joint readings across those robots | **724** |
| robots additionally publishing a free joint as a scalar | **15** |
| fixed-base arms unaffected | 32 |

On `unitree_g1` all 29 hinges were shifted, so the payload carried the pelvis **height** (`0.79160`) under `left_hip_roll_joint` and the base quaternion's **w** component (`0.99998`) under `left_hip_yaw_joint` - in a field whose documented unit is radians. Nothing raised and nothing logged; the values are plausible joint angles.

## Change

The simulation already answers this exact question, and `get_observation` owns three facts a second reader has to reproduce:

- it resolves each joint through `jnt_qposadr`;
- it excludes a free joint from the per-joint scalars, with a comment explaining that emitting one reports "the base x-coordinate as a joint angle ... a degenerate, misleading scalar";
- it takes the model lock "to prevent torn reads while a concurrent `mj_step` is mutating data arrays" - which the driver's own unlocked read did not, while a policy steps at 50 Hz.

The publisher now reads through that surface (`skip_images=True`: the 10 Hz loop wants joint state only, not a render of every camera). So there is one answer to "what are this robot's joint positions" instead of two, and the driver keeps its dependency surface - it still holds no MuJoCo import, where fixing the indexing in place would have put a third copy of the free-joint rule into a Device Connect adapter.

The hardware sibling already delegates this way: `RobotDeviceDriver.getState` reads the wrapped robot's own `get_observation`.

Out of scope: the payload shape is unchanged. `joints` stays per-joint scalars keyed by short name; `base_pos` / `base_quat` remain available on the simulation's observation for a subscriber that wants the base pose.

## Verification

A subscriber that renders the telemetry reconstructs the pose from `joints`. `unitree_g1` driven to a squat with both arms raised, then rendered from one publish cycle, MuJoCo headless:

observationUpdate joint addressing

| | before | after |
|---|---|---|
| keys published | 30 (incl. `floating_base_joint`) | 29 |
| joints disagreeing with the simulation | **15 of 29** | **0 of 29** |
| worst joint error | **2.2000 rad** (126 deg, `left_shoulder_pitch_joint`) | **0.0000 rad** |
| reconstructed render vs the true pose | mean abs 8.201, 9.91% of pixels off by >20 levels | mean abs 0.0001 (identical) |

The ground-truth panel is byte-identical between the two trees (mean abs 0.000067), so the two reconstructions differ only in the payload they were built from.

Per-robot agreement with `get_observation`, driven through the real publish path:

| robot | before | after |
|---|---|---|
| `so101` (fixed base) | 6 / 6 | 6 / 6 |
| `unitree_g1` | 0 / 29 | 29 / 29 |
| `unitree_go2` | 0 / 12 | 12 / 12 |

### Tests

`tests/test_device_connect_sim_joint_addressing.py` builds a hermetic MJCF - a floating base ahead of two hinges, so ids are `(0, 1, 2)` and addresses are `(0, 7, 8)` - with the base authored away from the origin and both hinges driven to distinct angles, so every candidate reading is a different number. Against `main`: **5 failed / 2 passed**, headline reporting

```
published {'floating_base': 0.0, 'j1': 0.25, 'j2': 0.5} for a robot whose joints
are {'j1': 0.3, 'j2': -0.6}; joint ids [0, 1, 2] address qpos slots [0, 7, 8]
```

i.e. `j2` published as `0.5`, the base height. All 7 pass after.

The 2 that pass on both trees are the controls: a fixed-base arm (ids and addresses coincide - the shape the publisher was already correct for) and a simulation exposing no observation surface (reports nothing rather than raising into the fleet-wide publish loop).

Two existing publisher tests drove the old reader and are repaired to drive the new one. One of them pinned the defect: its fixture was a `SimpleNamespace` carrying only `qpos`, which cannot express the difference between a joint id and a qpos address, so it asserted `qpos[jid]` as the contract. The repaired pair **fails against `main`** (2 failed / 5 passed on the `publish_state` selection), so the intent is preserved rather than relaxed.

### Gate

`ruff check` / `ruff format --check` / `mypy` clean over `strands_robots tests tests_integ` (mypy 24 errors, identical set to `upstream/main`, none in the changed files).

Blast radius - every test touching the driver, the publish loop, `get_observation` or `jnt_qposadr`, plus the repo-wide docs / docstring / changelog guards (197 files), run on this branch and on `upstream/main` with both test files copied into the base tree:

|---|---|---|
| failed | 4 | 11 |
| passed | 6695 | 6688 |
| skipped | 129 | 129 |
| collected | 6828 | 6828 |

Branch-only failures: **none**. The 7 base-only failures are exactly the 5 new tests plus the 2 repaired ones. The 4 shared failures are pre-existing and environmental here (2 need `device_connect_agent_tools`, which is a declared dependency; 2 are `test_doctor` distribution-metadata checks under an editable install).

### DIFF
--- a/changelog.d/2451-sim-driver-joint-observation-addressing.md
+++ b/changelog.d/2451-sim-driver-joint-observation-addressing.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **device_connect**: `SimulationDeviceDriver`'s 10Hz `observationUpdate` published each joint's position by indexing `qpos` with the joint's id rather than its qpos address, so on a robot with a floating base every later joint was shifted by six - a `unitree_g1` payload carried the pelvis height and the base quaternion under leg-joint names - and the free joint itself was emitted as a degenerate scalar. The publisher now reads through the simulation's own `get_observation`, which owns qpos addressing, the free-joint exclusion and the lock that keeps a read from being torn by a concurrent physics step.
--- a/docs/device-connect.md
+++ b/docs/device-connect.md
@@ -53,6 +53,28 @@ Each robot is wrapped as a Device Connect device by a `DeviceDriver` adapter:
 | `RobotDeviceDriver` | a hardware `Robot` | the same RPC surface, driving real servos |
 | `ReachyMiniDriver` | a Pollen Reachy Mini | device-native RPCs (`look`, `nod`, …) over Zenoh / WebSocket |
 
+### Published state events
+
+Besides answering RPCs, a driver publishes its own state on a 10 Hz loop while a
+policy is running. `SimulationDeviceDriver` emits two events:
+
+| Event | Payload |
+|-------|---------|
+| `stateUpdate` | `sim_time`, `step_count`, and `running_policies` (`{robot: {steps, instruction}}`) |
+| `observationUpdate` | `robot_name`, `sim_time`, `step_count`, and `joints` |
+
+`joints` is `{joint name: position}` in **radians**, and carries exactly the
+per-joint scalars the simulation's own
+[`get_observation`](simulation/overview.md) reports for that robot:
+
+- a value is read at the joint's own qpos address, so a robot with a floating
+  base reports its leg and arm angles rather than components of its base pose;
+- a floating base has no scalar joint position (its state is a position plus a
+  quaternion), so it is **not** a key in `joints` - subscribe to the simulation's
+  observation for `base_pos` / `base_quat` if you need the base pose;
+- only that robot's own joints appear, under their short names, whatever the
+  compiled model namespaces them to in a multi-robot scene.
+
 `init_device_connect()` / `init_device_connect_sync()` attach the right driver and start the runtime — `Robot().run()` calls these for you.
 
 ## Driving it from an agent
--- a/strands_robots/device_connect/sim_driver.py
+++ b/strands_robots/device_connect/sim_driver.py
@@ -219,6 +219,47 @@ class SimulationDeviceDriver(DeviceDriver):
 
     # ── Periodic state publishing ─────────────────────────────
 
+    def _joint_positions(self, robot_name: str, robot: Any) -> dict[str, float]:
+        """Per-joint positions for ``robot_name``, in radians.
+
+        Read through the simulation's own ``get_observation`` surface rather
+        than by indexing its state arrays here. That surface already answers
+        this exact question, and it owns three facts a second reader has to
+        reproduce and can silently get wrong:
+
+        * a joint's position lives at its qpos ADDRESS, not at its joint id.
+          The two coincide only while every joint is single-DoF: one floating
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Fixes the sim driver's 10 Hz `observationUpdate` publisher, which indexed `data.qpos` by joint **id** instead of qpos **address** — so any robot with a floating base published components of its base pose (height, quaternion terms) under joint names documented as radians, and emitted the free joint itself as a degenerate scalar. The fix delegates to the simulation's own `get_observation(robot_name, skip_images=True)`, which already owns `jnt_qposadr` addressing, the free-joint exclusion, short-name keying in multi-robot scenes, and the lock that serialises the read against a concurrent `mj_step` — the driver's old unlocked read had none of these, and the hardware sibling (`RobotDeviceDriver.getState`) already delegates the same way. Verified locally: the repaired publisher tests fail 2/5 against `main`'s driver and all 7 pass on this branch, matching the PR's claims; the new hermetic test file (`free-base` model where ids `(0,1,2)` diverge from addresses `(0,7,8)`) skips cleanly when `device_connect_edge` is absent.

## What's good

- The scalar filter in `_joint_positions` keeps the payload shape intact: non-scalar observation keys (`base_pos`, `base_quat`, image arrays when a recording override forces rendering) never reach `joints`, and the bool exclusion is correct.
- The one key removed from the payload (`floating_base_joint`) only ever carried a meaningless value (the base x-coordinate), so this is the fix rather than a wire-format break, and the docs now state the exclusion explicitly with a pointer to `base_pos`/`base_quat`.
- New tests pin the contract on a model where the two addressings actually differ, with premise assertions (`addresses != joint_ids`) so a fixture drift can't silently turn them into no-ops; the fixed-base and no-observation-surface controls pass on both trees.
- Changelog fragment, docs update, and the degraded-path guard (non-callable `get_observation` → `{}` rather than raising into the fleet-wide publish loop) all follow AGENTS.md conventions.

PR #2297, opened 2026-08-16
fix(rendering): key the baked-panorama cache on the geometry it was baked for

cagataycali:
## Problem

`bake_gsplat_panorama` short-circuits on a non-empty output file, and the default output path was `ply_path.stem + "_pano.jpg"` -- independent of `equi_w`, `equi_h` and `face_size`. Every bake of one scene therefore shared a single cache entry, so the second caller's geometry was accepted and then dropped:

```python
bake_gsplat_panorama(scene, equi_w=128,  equi_h=64)    # cheap preview
bake_gsplat_panorama(scene, equi_w=2048, equi_h=1024)  # -> returns the 128x64 preview
```

The second call returns the preview with **zero renders and no warning**, and `PanoramaBackground` then samples that 128x64 image as the backdrop for a full-resolution view. `examples/mujoco_gs/app.py` bakes with the defaults, so a later quality bump is silently a no-op.

`face_size` is dropped the same way, and it is the reason a size-aware cache check is not sufficient: `face_size` changes the sharpness of the cube faces the panorama is reprojected from, not the output dimensions, so a check that only compared the cached image's width/height would still serve the coarser bake.

## Change

Derive the default output path from the geometry that determines the pixels:

```
<stem>_pano_<equi_w>x<equi_h>_f<face_size>.jpg
```

An explicit `out_path` stays honored verbatim -- the caller named the file, so the caller owns what it holds -- and repeating one request still hits the cache, so the expensive six-render pass is not re-run per call.

## Evidence

Same call sequence on both trees, with the requested final bake at 2048x1024:

| | panorama actually used | cube faces rendered on the 2nd call | backdrop detail (mean abs gradient) |
|---|---|---|---|
| before | **128x64** | **0** | 0.997 |
| after | 2048x1024 | 6 | 1.363 |

baked panorama cache: requested resolution honored

Both panels are a real MuJoCo composite (headless EGL) of an so101 over a photoreal 3D-scene backdrop, produced by one script run against each tree so they differ only by this change (panel mean abs delta 6.33, max 215). The CUDA gaussian rasterizer is unavailable on the render host, so the splat render boundary is doubled exactly as `tests/rendering/test_backgrounds.py::TestBakeGsplatPanorama` does it -- each cube face is filled by sampling a real captured equirectangular panorama at that face's camera. Everything else is shipped code: the real upright transform (from real `.spz` splat means), the real cube-to-equirect reprojection, the real `PanoramaBackground` sampling and the real composite.

## Tests

[...]

PR #2452, opened 2026-08-19
fix(rendering): check the panorama bake's resolutions before it names a file or loads splats

### LINKED ISSUE #2450: rendering(backgrounds): bake_gsplat_panorama composes its cache key and loads the splats before anything checks the resolutions, so 640.0 re-bakes a warm 640 and equi_w=0 is reported as a missing torch
`bake_gsplat_panorama` takes the same three grid knobs `render_environment_map` does -- `face_size`, `equi_w`, `equi_h` -- and forwards them straight to it. #2446 gave that renderer, `bake_environment_map` and `environment_map_cache_path` one shared domain (`_resolution_error`, i.e. `positive_whole_number_error` over the three names) checked before any render and normalized with `int()` afterwards. `bake_gsplat_panorama` is the fourth caller of those knobs and got neither half.

Two reads happen before the renderer that owns the domain ever sees them.

**1. The default output path is composed from the raw values, and that name is the cache key.** #2297 made the default path `<stem>_pano_<equi_w>x<equi_h>_f<face_size>.jpg` precisely so a resolution could not be silently dropped. The shared domain accepts an integral float, so `face_size=640.0` -- a value a config float produces -- composes `scene_pano_2048x1024_f640.0.jpg` and misses the warm `scene_pano_2048x1024_f640.jpg` sitting beside it. Measured: a warm default-path cache plus `face_size=640.0` falls straight through the short-circuit into the splat load. That is the failure #2297 fixed, reintroduced by spelling. `environment_map_cache_path` normalizes for exactly this reason and says so.

**2. The splats load before the refusal.** An unusable resolution is refused only inside `render_environment_map`, which runs after `GsplatBackground._load()`. On a box with the `sim-gs` extra that costs the whole load. Without the extra it is worse than late -- it is *replaced*:

```
bake_gsplat_panorama(ply, equi_w=0)
-> ImportError: 'torch' is required for 3D Gaussian Splatting backgrounds (GsplatBackground)
   Install with:
     pip install 'strands-robots[sim-gs]'
```

The caller is told to install a dependency. Installing it does not help; `equi_w=0` is the problem and is named nowhere. Same shape as the refusal #2446 replaced, where a zero-width map was blamed on a black scene and the remedy offered was a dead end.

### Fix

Check the three knobs against `_resolution_error` and normalize with `int()` at the top of `bake_gsplat_panorama`, before the default path is composed and before the splats load. The domain belongs to `strands_robots.rendering.ibl` because that is the module this bake delegates its pixels to, so importing it keeps one authority rather than adding a fourth copy.

### Pinning

- an integral-float resolution hits a warm default-path cache instead of composing a second name for the same pixels
- an out-of-domain resolution raises `ValueError` naming the knob, with no `torch` import and no `GsplatBackground` construction attempted -- which is also what makes the test runnable without the `sim-gs` extra
- an explicit `out_path` is still honored verbatim and a good resolution still reaches the bake

cagataycali:
Closes #2450

## What

`bake_gsplat_panorama` forwards `face_size` / `equi_w` / `equi_h` to `render_environment_map`, which owns their domain, but two things read the values before that renderer is reached. Both are now answered at the top of the function: the three knobs are checked against the sibling module's `_resolution_error` and normalized with `int()`.

## Why

**The default output path is composed from them, and that name is the cache key.** #2297 made the default `<stem>_pano_<equi_w>x<equi_h>_f<face_size>.jpg` precisely so a resolution could not be silently dropped. The shared domain accepts an integral float, so `face_size=640.0` -- what a config float produces -- composed `scene_pano_2048x1024_f640.0.jpg` and fell straight through the short-circuit past the warm `scene_pano_2048x1024_f640.jpg` sitting beside it. Same silent no-op #2297 removed, reached by spelling rather than by an ignored knob. `environment_map_cache_path` normalizes for exactly this reason and says so.

**The splats load before the refusal.** An unusable resolution was refused only inside `render_environment_map`, i.e. after `GsplatBackground._load()`. With the `sim-gs` extra installed that costs the load. Without it, the load does not delay the refusal -- it replaces it:

```
bake_gsplat_panorama(ply, equi_w=0)
-> ImportError: 'torch' is required for 3D Gaussian Splatting backgrounds (GsplatBackground)
   Install with:
     pip install 'strands-robots[sim-gs]'
```

The caller is told to install a dependency. Installing it changes nothing; `equi_w=0` is the problem and is named nowhere. Same shape as the refusal #2446 replaced, where a zero-width map was blamed on a black scene and the only remedy offered was a dead end.

## Where the domain lives

In `strands_robots.rendering.ibl`, unchanged. `_resolution_error` is the one place the render, the bake and the cache path already agree on which resolutions are honorable, and this bake delegates its pixels to that same renderer -- so it imports the check rather than becoming a fourth copy of it. Its docstring's "one place" claim now names this caller, since the set it covers has grown past its own module.

## Tests

`tests/rendering/test_gsplat_panorama_resolution_domain.py`, 59 cases, parametrized over `positive_whole_number_error` itself so the case list cannot drift from the domain it claims to cover:

- each out-of-domain value x each knob raises `ValueError` naming the knob, **and the splat backend is never constructed** -- that second assertion is the ordering claim, not a detail
- the same, with an explicit `out_path`: that request shape bypasses the composed name, so the cache-key argument does not apply to it and the refusal still has to
- the refusal names `bake_gsplat_panorama`, so a message arriving from a six-render bake is attributable to it rather than to the renderer it delegates to
- `face_size=1` still reaches the bake -- the check refuses a domain, not a request, and where "too coarse" begins stays the caller's judgement
- an integral float for each knob hits the warm default-path cache, returns that file's bytes, and leaves no second name in the directory

Every test runs without `torch`, `gsplat` or a GPU, which is not a convenience: a refusal that arrives before the splat load is by definition reachable from a CPU-only environment, so the assertion and the property it pins are the same thing. `GsplatBackground` is doubled out to raise, marking the boundary between "checked the resolution" and "started paying for the pixels".

Measured against the parent commit: 58 of the 59 fail, the exception being `test_a_usable_resolution_reaches_the_bake`, which is the complement and must pass both ways.

`hatch run format`, `lint` (ruff + mypy) and `tests/rendering` clean. The 11 pre-existing `MockTensor` failures in `test_backgrounds.py` on a torch-less box are unchanged by this branch (429 passed before, 488 after, same 11 failures).

## Not in scope

`device` is the fourth argument this function takes and is unvalidated; it is a backend selector rather than a pixel count and shares no domain with these three, so it is left alone.

---

### §13 changelog

**Round 1** -- initial submission.

### DIFF
--- a/changelog.d/2450-gsplat-panorama-resolution-domain.md
+++ b/changelog.d/2450-gsplat-panorama-resolution-domain.md
@@ -1,0 +1,10 @@
+### Fixed
+- **rendering**: `bake_gsplat_panorama` checks `face_size` / `equi_w` / `equi_h`
+  against the same shared `positive_whole_number_error` domain its renderer uses,
+  and normalizes them with `int()`, before it composes its default output path or
+  loads the splats. That default path is the cache key, so `face_size=640.0` -
+  which the domain accepts - previously spelled `..._f640.0.jpg` and re-baked the
+  warm `..._f640.jpg` beside it. And an unusable resolution was refused only after
+  the splat load: without the `sim-gs` extra installed the load answered first,
+  so `equi_w=0` was reported as a missing `torch` and advised an install that
+  leaves the resolution exactly as unusable.
--- a/strands_robots/rendering/backgrounds.py
+++ b/strands_robots/rendering/backgrounds.py
@@ -800,7 +800,35 @@ def bake_gsplat_panorama(
 
     Returns the path to the written panorama ``.jpg`` (cached next to the ply
     unless ``out_path`` says otherwise).
+
+    Raises:
+        ValueError: if ``face_size``, ``equi_w`` or ``equi_h`` is not a positive
+            whole number. These knobs are forwarded to
+            :func:`~strands_robots.rendering.ibl.render_environment_map`, which
+            owns their domain, so this bake checks that same domain rather than
+            restating it -- and checks it *here* because two things read the
+            values before that renderer does (see below).
     """
+    # Both of those reads happen before ``render_environment_map`` is reached, so
+    # its own refusal arrives too late to be the only one.
+    #
+    # The default path is composed from these values and that name is the cache
+    # key, so an integral float must not spell a second file for pixels already
+    # baked: the shared domain accepts ``face_size=640.0``, which composed
+    # ``<stem>_pano_2048x1024_f640.0.jpg`` and re-baked a warm
+    # ``..._f640.jpg`` beside it -- the silent-no-op failure the geometry in this
+    # name exists to prevent, reintroduced by spelling. ``int()`` normalizes it
+    # away, as ``environment_map_cache_path`` does for the same reason.
+    #
+    # The splat load is what an unusable resolution would otherwise pay for
+    # first, and without the ``sim-gs`` extra installed the load does not merely
+    # delay the refusal, it replaces it: ``equi_w=0`` reported a missing ``torch``
+    # and advised installing it, which fixes nothing and names no resolution.
+    from .ibl import _resolution_error, render_environment_map
+
+    if text := _resolution_error("bake_gsplat_panorama", face_size=face_size, equi_w=equi_w, equi_h=equi_h):
+        raise ValueError(text)
+    face_size, equi_w, equi_h = int(face_size), int(equi_w), int(equi_h)
     ply_path = Path(ply_path)
     if out_path is not None:
         out = Path(out_path)
@@ -821,10 +849,9 @@ def bake_gsplat_panorama(
     base._transform = T
 
     # The six-cube-face render + equirect reprojection is shared with the
-    # world-frame environment-map bake (issue #2323). This bake keeps its own
-    # viewpoint (the scene centroid, in the unaligned upright frame above).
-    from .ibl import render_environment_map
-
+    # world-frame environment-map bake (issue #2323), imported above with the
+    # resolution domain it owns. This bake keeps its own viewpoint (the scene
+    # centroid, in the unaligned upright frame above).
     pano = render_environment_map(
         base,
         origin_world=(0.0, 0.0, 0.0),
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

`bake_gsplat_panorama` now validates `face_size` / `equi_w` / `equi_h` against the shared `_resolution_error` domain (backed by `positive_whole_number_error`) and normalizes them with `int()` at the top of the function, before the default output path is composed and before `GsplatBackground` is constructed. This closes two real defects: an integral float (`face_size=640.0`) previously spelled a second cache-key filename (`..._f640.0.jpg`) and silently re-baked pixels already warm on disk, and an out-of-domain resolution on a box without the `sim-gs` extra was reported as a missing `torch` install rather than as the bad resolution. The check is imported from `strands_robots.rendering.ibl` rather than duplicated, keeping one authority over the domain, and matches the `int()` normalization `render_environment_map` / `environment_map_cache_path` already do.

## What's good

- Verified locally at the head SHA: all 59 new tests pass, the full `tests/rendering/` suite is green (540 passed, 1 skipped), and with the parent commit's `backgrounds.py`/`ibl.py` restored the new suite fails 58/59 with exactly the complement case passing — matching the PR description's claim precisely.
- The tests parametrize over the shared domain itself (asserting each value is out-of-domain as a premise), so the case list cannot drift from the authority it covers; the ordering claim (backend never constructed on refusal) is asserted directly via the doubled-out `GsplatBackground`, and everything runs CPU-only.
- No public API signature change; the new `ValueError` on previously-mishandled inputs follows the repo's raise-on-fatal / no-silent-defaults rules and the #2446 precedent for the sibling entry points.
- Changelog fragment present (issue-number naming is explicitly allowed by `changelog.d/README.md`), ruff check/format clean on the changed files, no non-ASCII, no host paths, docstring cross-reference targets kept on single lines per AGENTS.md.
- No import cycle risk from the cross-module `_resolution_error` use: `ibl.py` does not import `backgrounds` at module level, and the import in `bake_gsplat_panorama` stays lazy alongside the existing `render_environment_map` one.

PR #2454, opened 2026-08-19
fix(policies/groot): report the determinism mode the server got, not the one it asked for

cagataycali:
## Problem

`server_wrapper.main()` prints a startup banner that records the determinism configuration the GR00T server is running under:

```
[srv_wrap] determinism: cudnn.deterministic=True, benchmark=False, strict=True
```

The `strict=` value came from `strict_det` - the flag read out of `STRANDS_GR00T_STRICT_DETERMINISTIC`, i.e. the **request**, never the outcome.

Enabling strict mode is best-effort by design. An op with no deterministic kernel makes `torch.use_deterministic_algorithms` refuse, and the wrapper deliberately degrades to non-strict rather than killing the server ("strict mode is a tightening, so losing it must degrade rather than kill"). On that path the banner still reported `strict=True`, one line after its own failure warning, so a single log both reported the applier failing and recorded the run as strict:

```
[srv_wrap] warning: use_deterministic_algorithms failed: linalg_cholesky has no deterministic implementation
[srv_wrap] determinism: cudnn.deterministic=True, benchmark=False, strict=True     <- strict is off
```

That banner is the only record of a container run's determinism configuration - it is what an operator reads back out of `docker logs` to see what the server enforces. A rollout that later failed to reproduce would therefore be investigated against a log that had already ruled out the one setting that was in fact missing.

Driving the real `main()` with `torch` doubled so its refusal is exact:

| scenario | torch applied strict | banner read | |
|---|---|---|---|
| no opt-in | False | `strict=False` | agrees |
| opt-in, torch accepts | True | `strict=True` | agrees |
| opt-in, torch refuses | **False** | **`strict=True`** | **contradicts** |

The same file already gets this right for the other thing it applies: a `reset` carrying a seed it cannot use prints `warning: <reason>; using <default>`, naming the dropped request and the value the episode actually ran with. Strict mode had no equivalent.

## Change

Track whether the applier succeeded rather than whether it was asked for, and report that. A request that was read and then refused says so, so off-because-refused stays distinguishable from never-asked-for:

```
[srv_wrap] determinism: cudnn.deterministic=True, benchmark=False, strict=False (requested, but use_deterministic_algorithms failed)
```

The success announcement moves into an `else:` clause, so "strict was applied" and "the `STRICT mode` line was printed" are the same condition by construction.

No behaviour change. A server without the opt-in and a server whose opt-in torch accepted print a byte-identical banner, and losing strict mode still degrades rather than kills - all six runs above reach the unmodified server entrypoint.

## Tests

Extends `TestStrictDeterminismIsOptIn`, whose three existing tests become controls. Against `upstream/main`: **3 failed / 55 passed**; after: **58 passed**.

Failing pre-fix:

- `test_a_refused_strict_request_is_not_recorded_as_a_strict_run` - the self-contradiction, quoting both lines: `warning: '[srv_wrap] warning: use_deterministic_algorithms failed: ...', banner verdict: 'True'`.
- `test_the_banner_states_the_mode_the_server_ended_up_in[requested-and-refused]` - the root-cause property, parametrized over all three states. The other two rows **pass on both trees**, so the parametrization carries its own controls.
- `test_the_banner_names_a_strict_request_it_could_not_honor` - off-because-refused vs never-asked-for.

Passing on both trees (they fail for the plausible wrong fixes):

- `test_the_strict_line_only_appears_when_torch_accepted_the_request` - fails if the announcement is printed unconditionally.
- `test_strict_mode_is_off_by_default` / `test_the_opt_in_enables_deterministic_algorithms` - fail if the opt-in stops reaching torch.
- `test_a_strict_mode_failure_does_not_stop_the_server` - fails if a refusal is made fatal instead of reported.

The verdict is read as the banner's leading token, so the check is about the verdict rather than about any reason appended after it.

## Verification

Measured at `913b7aa0` against `upstream/main` `ab2beb7b`, on a 128-file selection (all of `tests/policies/groot`, every test naming the wrapper / its env vars / the deterministic lifecycle, plus the repo-wide docstring, changelog, ASCII-string and dependency guards):

- branch **4 failed / 4501 passed / 54 skipped**, base **7 failed / 4498 passed / 54 skipped**, **4505 collected on both**. Failed delta == passed delta == 3 == the pre-fix failure count.
- branch-only failing IDs: **none**. Base-only: exactly the 3 pre-fix failures above.
- The 4 shared failures are environmental and unrelated: 2x `test_doctor.py::TestDoctorVersionResolution` (editable-install metadata) and 2x `test_lerobot.py::TestInProcessTrainerCorrectness` (`draccus` older than lerobot's declared floor locally).
- `ruff check` / `ruff format --check` over `strands_robots tests tests_integ`: clean. `mypy`: 24 pre-existing on both trees, none in the changed files.

## Artifact

The same three scenarios driven through the real `server_wrapper.main()` in each tree - `torch` is doubled so its refusal is exact, and nothing else differs between the two rows. Banner disagrees with what torch applied: **1 of 3 scenarios before, 0 of 3 after**.

determinism banner before and after

## Scope

Only the strict-mode verdict. `CUBLAS_WORKSPACE_CONFIG` is reported by reading the environment back, and the cuDNN flags are plain assignments that cannot partially apply, so neither can disagree with the banner. Whether `setdefault` should overwrite an operator's preset `CUBLAS_WORKSPACE_CONFIG` is a separate question with a test already pinning the current answer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1147,7 +1147,7 @@ touches ROS 2.
 | `STRANDS_GR00T_IMAGE` | Container image the `gr00t_inference` tool runs (must pass the image allowlist; agent cannot choose it) | `gr00t:latest` |
 | `STRANDS_GR00T_IMAGE_ALLOW` | Extra image-name patterns (trailing `*` = tag wildcard) added to the built-in allowlist (`gr00t:*`, `nvcr.io/nvidia/isaac-gr00t:*`) | built-in only |
 | `STRANDS_GR00T_SERVER_SEED` | Default seed the GR00T determinism wrapper applies at server start and on seedless `reset` calls (used with `gr00t_inference(..., deterministic=True)`; forwarded into the container) | `42` |
-| `STRANDS_GR00T_STRICT_DETERMINISTIC` | `1` makes the determinism wrapper additionally enable `torch.use_deterministic_algorithms(True, warn_only=True)` (slower kernels, strictest reproducibility; forwarded into the container) | `0` |
+| `STRANDS_GR00T_STRICT_DETERMINISTIC` | `1` makes the determinism wrapper additionally enable `torch.use_deterministic_algorithms(True, warn_only=True)` (slower kernels, strictest reproducibility; forwarded into the container). Best-effort: an op with no deterministic kernel makes torch refuse, degrading the server to non-strict rather than killing it, and the startup banner reports `strict=` as the mode the server ended up in | `0` |
 
 </details>
 
--- a/changelog.d/2454-groot-strict-determinism-banner.md
+++ b/changelog.d/2454-groot-strict-determinism-banner.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **policies/groot**: the determinism wrapper's startup banner now reports the mode the server
+  ended up in rather than the one it was asked for. Enabling strict mode is best-effort - an op
+  with no deterministic kernel makes `torch.use_deterministic_algorithms` refuse, which degrades
+  the server to non-strict instead of killing it - but the banner read the request, so such a run
+  logged `strict=True` one line after its own failure warning. That banner is the only record of
+  a container run's determinism configuration, so an investigation into a rollout that failed to
+  reproduce would read it and rule out the one setting that was in fact missing. A refused
+  request now reads as off and names itself.
--- a/strands_robots/policies/groot/server_wrapper.py
+++ b/strands_robots/policies/groot/server_wrapper.py
@@ -6,7 +6,8 @@ enforces:
 - ``cudnn.deterministic = True``
 - ``cudnn.benchmark = False``
 - ``torch.use_deterministic_algorithms(True, warn_only=True)`` (opt-in via
-  ``STRANDS_GR00T_STRICT_DETERMINISTIC=1``)
+  ``STRANDS_GR00T_STRICT_DETERMINISTIC=1``; a request torch refuses degrades to
+  non-strict, and the startup banner reports it as off)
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Narrow observability fix in `server_wrapper.main()`: the startup banner's `strict=` field now reports whether `torch.use_deterministic_algorithms(True, warn_only=True)` actually applied (`strict_applied`) rather than whether it was requested (`strict_requested`), so a request torch refuses reads as `strict=False (requested, but use_deterministic_algorithms failed)` instead of contradicting the failure warning printed one line above it. The `STRICT mode` success line moves into the `else:` of the same `try`, making "applied" and "announced" one condition by construction. Verified at head: the three new regression tests fail on `main`'s wrapper (3 failed / 55 passed) and the full file passes at `913b7aa0` (58 passed); the two common paths (never-requested, requested-and-accepted) print byte-identical banners to before, and the only in-tree consumer of the banner (`tests_integ/groot/test_deterministic_lifecycle.py`) matches a substring that does not extend past `strict=`, so nothing parses the changed tail. Degrade-rather-than-kill behaviour is unchanged. `ruff check`/`format` clean on the changed files; changelog fragment and README env-var row updated in the same PR per AGENTS.md.

## What's good

- The refused-request verdict stays machine-readable: the leading token is still `False`, with the reason appended after it, and the tests read the verdict as the leading token rather than string-matching the reason.
- Off-because-refused vs never-asked-for are kept distinguishable and pinned by a dedicated test, matching the file's existing seed-path convention of naming the dropped request beside the fallback.
- Parametrized test carries its own controls (the two agreeing rows pass on both trees), and the plausible wrong fixes each have a test that fails on them.

PR #2455, opened 2026-08-19
fix(training/rl): check the entropy temperature where it starts, not only where it moves

cagataycali:
## Problem

FastSAC does not store its entropy temperature. It stores the temperature's **logarithm**, so `init_alpha` reaches `torch.log` on both temperature branches of `setup`:

```python
self.log_alpha = torch.tensor(float(torch.log(torch.tensor(spec.init_alpha))), ...)
```

and `alpha = log_alpha.exp()` scales the entropy term in the critic's TD target and in the actor loss from there on. Only a positive finite value has a usable logarithm.

The temperature's *learning rate* is already checked against the shared positive-finite domain. The value that rate moves was not  - and that gate's own reasoning depends on this one, since it refuses `alpha_lr=0` on the grounds that *"the temperature stays at `init_alpha` for the whole run"*, which is only a usable statement if `init_alpha` is itself usable.

Measured on a 40-timestep so100 reach run, seed 0, driven through the real `validate()` -> `train()` path:

| `init_alpha` | preflight | outcome |
| --- | --- | --- |
| `1.0` (default) | accepted | success; alpha adapts `1.0 -> 0.9970` |
| `0` / `0.0` / `False` | **accepted** | **`status="success"`**; `log(0) = -inf` so alpha is exactly `0.0` at every update, the entropy term is absent from both losses, and the checkpoint holds `log_alpha = -inf` |
| `True` | accepted | a silent temperature of exactly `1.0` |
| `-1e-3`, `-1`, `nan`, `inf`, `-inf` | accepted | `ValueError` from `torch.distributions.Normal` about a 16x6 tensor of `nan` policy means, inside `train` |

Two distinct harms. `init_alpha=0` is silent and permanent: with the entropy term gone the run is no longer the maximum-entropy algorithm that was asked for, and because no finite gradient step moves an infinity the requested `autotune_alpha` cannot recover it  - `log_alpha` was still `-inf` after further updates. The non-finite value is then persisted, so it outlives the run that produced it. The other values raise, but the message names `Normal`'s `loc` parameter rather than the field or the value that produced it, and it arrives after the env and both networks are built  - past the point `Trainer.validate` documents itself as running before.

## Change

Route `init_alpha` through the same `positive_finite_number_error` domain its learning rate uses, via a shared `initial_temperature_problems` gate and a `BaseRLAlgo._initial_temperature_problems` wrapper, mirroring `temperature_learning_rate_problems`.

One deliberate difference from that sibling: this check is **not** conditioned on `autotune_alpha`. That gate guards an optimizer only the tuning branch constructs, so it is inert when tuning is off; `init_alpha` is read on *both* branches, and with tuning off it is the temperature for the entire run with nothing able to move it afterwards  - so the un-tuned case needs the check more, not less.

`target_entropy` is out of scope, measured rather than assumed: every value it cannot use (`nan`, `inf`, `-inf`, a `str`) already raises, and its domain differs  - a negative value is its default (`-num_actions`).

## Tests

`tests/training/test_initial_temperature_domain.py`, mirroring the sibling rate's domain file. Against `upstream/main`: **68 failed / 49 passed**; after: **117 passed**. Pre-fix failures are behavioural, e.g. `AssertionError: fast_sac accepted init_alpha=0` and `modules read spec.init_alpha without the shared gate: ['fast_sac.py']`.

The 49 that pass on both trees are the controls, and each fails for a specific wrong fix:

- the shipped `1.0` default and every usable temperature report nothing (fails if the domain is drawn too tight)
- `ppo` and `mock` stay silent about a field they never read, while their *own* `learning_rate` is still refused  - so the silence is scoping, not an empty preflight
- the verdict equals the shared domain's verdict for all 19 probe values, so the two cannot drift apart
- a usable temperature *does* move under the same optimizer loop that leaves a zero one pinned  - the control for the headline measurement
- the ownership sweep derives its reader set from the tree, detects a planted reader, and pins that no backend re-implements the comparison locally

## Verification

Measured at `c8dbb5d4` against `upstream/main` `6bf523d8`.

- 85-file blast radius (every test touching the changed symbols, all of `tests/training`, plus the repo-wide docs/docstring/changelog/source-string guards), both trees in parallel: branch **11 failed / 5440 passed / 6 skipped**, base **79 / 5372 / 6**, both collecting **5457**. Failed delta == passed delta == 68 == the pre-fix failure count. Branch-only failing set is **empty**; base-only is exactly the 68 from the new file, with zero foreign.
- The 11 shared failures are pre-existing and environmental: 8 in `tests/mesh/test_iot_connect_timeout_domain.py` (`awsiot` absent locally, a declared dependency) and 3 draccus-version cases in `tests/training/test_lerobot*.py`.
- `ruff check` and `ruff format --check` over `strands_robots tests tests_integ`: clean. `mypy` over the same: 24 errors on both trees, branch-only empty, none in the changed files.

## Artifact

One capture script per tree, driving the real `validate()` -> `train()` path and tracing the temperature each run actually uses.

FastSAC init_alpha domain

The bottom band is the control: for the shipped `init_alpha=1.0` the alpha trace and the checkpointed `log_alpha` (`-0.003004195401445031`) are byte-identical across the two trees, so only the refused spelling changes behaviour.


yinsong1986 (approved):
## Summary

Adds a `validate()`-time preflight for FastSAC's `init_alpha`, routing it through the same shared `positive_finite_number_error` domain its sibling `alpha_lr` already uses: a new `initial_temperature_problems` gate in `training/_validate.py`, a `Trainer._initial_temperature_problems` wrapper mirroring `_temperature_learning_rate_problems` (same lazy-import pattern to keep `base -> _validate` one-way), and one `problems.extend(...)` call in `FastSacTrainer.validate`. The deliberate scoping difference from the rate gate - not conditioned on `autotune_alpha`, because `setup` reads `spec.init_alpha` on both temperature branches (verified: two `torch.log(torch.tensor(spec.init_alpha))` sites) - is correct and well argued: with tuning off the value is the temperature for the whole run. The refusal is the right call per AGENTS.md's no-silent-defaults rule, since `init_alpha=0` previously produced a `status="success"` run with the entropy term silently absent and a `-inf` checkpoint, and the other unusable values raised from inside `torch.distributions.Normal` naming neither the field nor the value.

## What's good

- Verified locally at `c8dbb5d4`: the new file's 117 tests pass, the sibling `test_temperature_learning_rate_domain.py` / `test_rl_fast_sac.py` / `test_mock_trainer_validate.py` suites pass (149), and `ruff check` / `ruff format --check` are clean on all changed files.
- The domain-parity test pins the gate's verdict to the shared rule for all 19 probe values, and the AST ownership sweep derives the reader set from the tree (with a planted-reader control and a non-vacuity check), so a future off-policy backend that reads `init_alpha` without the gate fails loudly.
- Scope discipline: `ppo`/`mock` stay silent about a field they never read, matching the `gae_lambda_problems` scoping convention, with a control proving the silence is scoping rather than an empty preflight.
- Changelog fragment (`changelog.d/2455-...`), docs paragraph beside the existing `alpha_lr` one, and full `Args:` entries on both new functions - all AGENTS.md conventions followed. The one non-ASCII character `grep -nP '[^\x00-\x7F]'` finds in the touched modules (`base.py:279`) is pre-existing on `main`, not introduced here.

Issue #1800, opened 2026-07-31
A cancelled check rolls up as FAILURE, so main's merge history reads as broken when it is only batched


cagataycali:
`AGENTS.md` > PR Workflow > step 8 tells you to verify a PR's state by reading it back, and lists what to poll before merging. It says nothing about how to read `main` **after**, and the obvious reading is wrong in a way that invites a revert of a good change.

## The misread

`statusCheckRollup.state` on a merge commit returns `FAILURE` when a check was **cancelled**. Cancellation is not a verdict, but it aggregates as one, and the checks UI draws it the same red as a genuine failure. `main`'s test-lint workflow sets `cancel-in-progress`, so every merge cancels the run of the merge before it.

Four PRs merged in the 22 minutes from 03:03:44 to 03:25:25 today. Three consecutive commits on `main` each report rollup `FAILURE` whose only non-`SUCCESS` context is `call-test-lint / Test and Lint` = `CANCELLED`:

| commit | PR | committed | test-lint started | cancelled | elapsed when killed |
|---|---|---|---|---|---|
| `a648a00` | #1788 | 03:03:44 | 03:03:52 | 03:04:59 | 1m07s |
| `c753a76` | #1794 | 03:04:39 | 03:05:02 | 03:20:02 | 15m00s |
| `3107c94` | #1796 | 03:19:42 | 03:20:06 | 03:25:44 | 5m38s |
| `e567473` | #1793 | 03:25:25 | 03:25:47 | - | still running |

Every other context on all three is `SUCCESS`. Nothing failed. An agent or contributor polling `main` sees three red commits in a row and the natural conclusion - that one of those squashes broke the build - is available, cheap, and false. The remedy that conclusion suggests is a revert of a reviewed, individually-green change.

## The real cost underneath it

The same table carries a second fact that is not a misread. `c753a76`'s run reached 15m00s without finishing, so the suite outruns the interval between merges at this cadence. That means **only the tip is ever verified**: each intermediate merge commit's full run is killed before it can report, so no intermediate commit is attributable.

If the tip goes red, you cannot tell which of the four merges caused it without re-running each one. The batch is still a defensible thing to do - every PR had its own green `call-test-lint`, the `Detect an untested overlap with the base branch` gate passed on each, and their file sets were pairwise disjoint - but the coverage you end up with is tip-only, and that is worth knowing before you rely on an intermediate commit's green.

## Proposed change

[...]

PR #2456, opened 2026-08-19
ci: bound the apt steps so a hung mirror fails fast and names apt

### LINKED ISSUE #2442: ci(test-lint): the apt step retries a failure but cannot survive a hang - so a wedged mirror spends the whole 45-minute job budget and reports FAILURE with the test step skipped
## What happened

Twice inside one 80-minute window, `call-test-lint / Test and Lint` reported a rollup of `FAILURE` having **never run a test**. Both jobs were reaped by the job's own `timeout-minutes: 45` while still inside step 4, `Install system dependencies (OpenGL for MuJoCo)`:

| run | subject | window | duration | step 4 | `Run lint` / `Run tests` |
| [32163486141](https://github.com/strands-labs/robots/actions/runs/32163486141) | `main` @ `74d078ab` (push) | 17:03:05 -> 17:48:28Z | **45m23s** | `cancelled` | `skipped` |
| [32165565982](https://github.com/strands-labs/robots/actions/runs/32165565982) | #2440 @ `2c489f92` (pull request) | 17:36:09 -> 18:21:32Z | **45m23s** | `cancelled` | `skipped` |

Steps 1-3 (`Set up job`, `Checkout code`, `Set up Python`) succeeded in both. Steps 5-9 - including `Install dependencies`, `Run lint` and `Run tests` - are `skipped` in both. Two different runners, overlapping windows, the same wedge point, and the same duration to the second, which is the job timeout rather than anything the step decided.

## Why the existing mitigation does not cover it

The step already anticipates a bad apt mirror, and its comment names the failure mode it was built for:

> ... whose mirrors sometimes serve a corrupt InRelease file ("Clearsigned file isn't valid, got 'NOSPLIT'"), making a plain `apt-get update` **exit 100 and fail the whole build** ... so retry a few times with a short backoff

```bash
for attempt in 1 2 3; do
  if sudo apt-get update; then
    break
  fi
  echo "apt-get update failed (attempt ${attempt}/3), retrying in 5s..."
  sleep 5
done
sudo apt-get install -y libosmesa6-dev ffmpeg
```

The loop is driven by `apt-get update`'s **exit status**, so it rides out a mirror that *answers wrongly*. It cannot ride out a mirror that *does not answer*: on a hang the command never returns, the `if` never evaluates, the loop never reaches attempt 2, the backoff never runs, and the `echo` that would have named the problem in the log is never printed. Nothing between the step and the 45-minute job ceiling bounds the wait - neither `apt-get update` nor `apt-get install` carries a timeout, and the step carries no `timeout-minutes` of its own.

So the guard covers failure and not hang, and the two are the same event from the mirror's side.

## Why this costs more than the 45 minutes

The wasted runner time is the smaller half. The reported verdict is the expensive part, and it is wrong in the direction that looks like a code problem:

- **The rollup says `FAILURE` while carrying zero test information.** A cancelled context aggregates into `FAILURE` and the rollup carries no reason, which is the exact read AGENTS.md already warns about for #1800/#2304 - except there the producer was a concurrency-group cancel, so the fix keyed the group on `github.sha`. This is a *third* producer of the same false red, and it survives that fix untouched.
- **`main`'s tip has no test verdict, and cannot get one by itself.** Per AGENTS.md, "a commit on `main` is immutable and already merged, so unlike a branch it gets no next push to clear the answer." `74d078ab` is the current tip and its only substantive check is a cancelled one.
- **The failure is silent in the log.** Because the retry `echo` never runs, nothing in the step output says "apt is stuck"; the log simply stops mid-step. A maintainer reading the red arrives at `Run tests` being skipped with no stated cause, and the natural next move - re-run, or go looking in the diff - is the expensive one.
- **It is indistinguishable from a genuine test timeout** until you open the job and read which step holds the `cancelled`. Both present as `call-test-lint` = `FAILURE` at ~45 minutes.

## Suggested direction

Convert the hang into the non-zero exit the loop already handles, rather than adding a second mechanism beside it - `timeout(1)` returns 124, which the existing `if` reads as failure and retries, so the backoff and the diagnostic `echo` both start working for this case too:

```bash
for attempt in 1 2 3; do
  if sudo timeout 180 apt-get update; then
    break
  fi
  echo "apt-get update failed or hung (attempt ${attempt}/3), retrying in 5s..."
  sleep 5
done
sudo timeout 300 apt-get install -y libosmesa6-dev ffmpeg
```

That bounds the step at roughly 14 minutes worst case instead of consuming the whole job, and a stall that outlasts three bounded attempts then fails **fast, loudly and with a log line naming apt** - so the red means "the mirror is wedged", not "something in your diff". Worth pairing with an explicit `timeout-minutes` on the step so the bound is declared in one place rather than implied by the sum of the `timeout` calls.

A point worth deciding rather than assuming: whether an apt stall should be a *failure* at all, given the packages come from the standard Ubuntu archive while the flaky feed is a preconfigured third-party one. Failing fast is a strict improvement over the current behaviour either way, so it does not need to be settled first.

## Notes

- Not a concurrency-group cancel. The `main` run's 45m23s is wall-clock inside one job, not a supersede - #1800's `github.sha` keying is working as documented, and I checked that first because the presentation is identical.
- Not repo-wide at the time: [32163828751](https://github.com/strands-labs/robots/actions/runs/32163828751) (17:06) and [32156102998](https://github.com/strands-labs/robots/actions/runs/32156102998) (15:42) both completed `success`, so the stall is intermittent per-runner rather than a total outage. That is what makes a bounded retry the right shape - most attempts do succeed.
- A re-run of attempt 1 on #2440 was enough to get past it, which is evidence the stall is transient and *not* evidence the guard is adequate: the re-run is a manual step that a bounded retry would have performed automatically, and on `main` no such push exists to hang the re-run off.

Found while verifying #2440, whose own verdict this delayed by ~45 minutes.

cagataycali:
Fixes #2442

## Problem

`call-test-lint / Test and Lint` reported a rollup of `FAILURE` having **never run a test**, twice inside one 80-minute window. Read back from the attempt-1 job records (the API serves the latest attempt, and both runs were re-run to green afterwards):

| run | subject | job | step 4 `Install system dependencies` | `Run lint` / `Run tests` |
|---|---|---|---|---|
| 32163486141 | `main` @ `74d078ab` (push) | `cancelled` 45m23s | `cancelled` **45m12s** | `skipped` |
| 32165565982 | #2440 @ `2c489f92` (PR) | `cancelled` 45m23s | `cancelled` **45m11s** | `skipped` |

Steps 1-3 succeeded in both and steps 5-9 are `skipped` in both. Two runners, overlapping windows, the same wedge point, and the same duration to the second - which is the job's own `timeout-minutes: 45`, not a decision either step made.

The step already anticipates a bad mirror and retries it, but the loop is driven by `apt-get update`'s **exit status**:

```bash
for attempt in 1 2 3; do
  if sudo apt-get update; then break; fi
  echo "apt-get update failed (attempt ${attempt}/3), retrying in 5s..."
  sleep 5
done
```

That rides out a mirror that *answers wrongly* (exit 100 on a corrupt `InRelease`) but not one that *never answers*. On a hang the command never returns, the `if` never evaluates, attempt 2 is never reached, the backoff never runs, and the `echo` that would have named apt never prints - so the log simply stops mid-step. Nothing between the step and the 45-minute ceiling bounds the wait.

Two costs, and the reported verdict is the expensive one:

- **A job-level reap aggregates to `FAILURE` carrying no reason.** Same false red as #1800/#2304, from a third producer that their `github.sha` keying does not touch. A maintainer arrives at `Run tests` = `skipped` with no stated cause, so the red reads as "something in your diff".
- **`main`'s tip cannot clear its own answer.** `74d078ab` is merged and immutable, so unlike a branch it gets no next push.

## Change

Three parts, each covering a case the others cannot.

**1. `timeout` on every retried `apt-get update`** - converts a hang into exit 124, which is the failure the loop already handles, so the retry, the backoff and the diagnostic all start working for the hang case. A step-level bound cannot substitute here: it kills the step rather than the attempt, so the loop still never retries.

**2. `timeout-minutes` on both apt steps** (`test-lint.yml`, `agent-api-check.yml`) - so a stall is reaped on the step that caused it, which marks that step, instead of by the job, which names nothing. A per-command `timeout` cannot substitute for this either, per the sizing below.

**3. The retry loop in `agent-api-check.yml`, which previously had none** - `timeout` produces an exit status and nothing more, so a bound with nothing reading it turns a transient stall into a hard red. This one is here because CI demonstrated it on this PR; see round 1.

## Sizing (this is where the change departs from the issue)

#2442 suggested `timeout 180` for `update` and `timeout 300` for `install`. Measured against the real distribution, the `install` half of that would have reaped healthy runs, so the bounds here are sized from observation instead.

Apt-step duration over the last 31 **successful** `Test and Lint` jobs: p50 **33s**, p90 264s, max **455s** - with 7 runs above 180s, and 5 of those *outside* the incident window (455s at 00:22Z, 264s at 03:22Z). So the slow tail is a standing property, not the incident.

Reading the slowest run's log (job `95914546572`, step 455s) splits the cost:

| command | cost |
|---|---|
| `apt-get update` | **~5s** (finished 00:22:27, step began 00:22:21) |
| `apt-get install` | **~449s** - `Get:115`/`Get:116` at 00:25:41/00:26:07, setup done 00:29:53 |

So the two commands want different treatment:

- **`update` is bounded per attempt** (180s in `test-lint.yml`, 150s in `agent-api-check.yml`). It is the command the loop retries, and its honest cost is seconds.
- **`install` is bounded by the step, not by a per-command `timeout`.** Any bound tight enough to be useful sits inside the healthy distribution. The same tail appears in `agent-api-check.yml`, whose apt step installs no ffmpeg and still ran **474s** once over 30 runs against a p50 of 15s.

The step bound is 12 for `test-lint.yml`, and the band it has to sit in is narrow at both ends: it must clear the observed 455s maximum (7m35s), and the rest of the job costs up to ~32 min against a 45-minute job bound, so anything above ~13 would let a slow apt push the job into the very reap this prevents. `agent-api-check.yml` gets 9 inside its 10-minute job, which is also why its per-attempt bound is 150s rather than 180s: 3x(150+5) plus the install is ~8 min, so the worst case still lands inside the step's own bound.

## Tests

New `tests/test_apt_steps_are_bounded.py`, keyed on the class rather than on the one site - which is what makes the sibling workflow part of this diff rather than a follow-up:

1. every step running `apt-get` declares a literal `timeout-minutes`, strictly less than its job's;
2. every `apt-get` whose exit status drives a loop/branch condition carries a literal `timeout`;
3. each such `timeout` is smaller than its own step's bound, or it could never fire.

Against `upstream/main`: **9 failed / 21 passed / 1 skipped**. After: **33 passed**. The pre-fix failures are the two unbounded steps (both clauses, both workflows) and the unbounded looped `update`. The skip is clause 3's parametrization being empty on base - there are no command bounds there to check - and it populates once one exists.

Parser non-vacuity is asserted rather than assumed, since a structural guard that stops matching passes everything: `TestTheParserActuallySeesTheApt` cross-checks the parse against a raw text scan **per file**, and pins that at least one invocation reads as condition-driven so clause 2 is not vacuous. Quoted text is stripped before matching, so the loop's own `echo "apt-get update failed ..."` reads as the string it is rather than as a third invocation.

Parsing is line-based, matching the neighbouring workflow pins (`test_workflow_jobs_are_bounded.py`, `test_codeql_query_filters.py`): `tests/` is type-checked under `ignore_missing_imports = false` and `types-PyYAML` is not a dev dependency, so importing it would either fail `mypy` or force a dependency change plus a `uv.lock` relock.

## Verification

Measured at `919a1b3c` against `upstream/main` `aee7b961`:

- New test: **33 passed** on branch, **9 failed / 21 passed / 1 skipped** on base.
- The CI-contract and string-hygiene guards together - every test reading `.github/`, plus `test_changelog_fragments`, `test_changelog_format`, `test_log_strings_are_ascii`, `test_source_strings_no_unicode_dashes`: **387 passed**, including `test_workflow_jobs_are_bounded.py`, which owns the job-level half of this contract and is unchanged.
- Both edited workflows parse under `yaml.safe_load`, and the parsed values agree with the contract: `test-lint.yml` step 12 < job 45, `agent-api-check.yml` step 9 < job 10.
- `ruff check` / `ruff format --check` / `mypy` on the new file: clean.

## Scope

Only the bounding of apt. Whether an apt stall should be a *failure* at all - the packages come from the standard Ubuntu archive while the flaky feed is a preconfigured third-party one - is left open, as #2442 notes: failing fast is a strict improvement either way, so it does not need settling first. The `install` command deliberately keeps no per-command `timeout`, for the measured reason above; if one is ever wanted, clause 3 already requires it to be smaller than the step bound.

---

## Round 1 - `919a1b3c`

**CI reproduced the target defect against this PR's own first commit, and the fix was one part short.**

`Detect AgentTool API Breaking Changes` failed at exactly **180s** on `62cb0fa9`. Not a false red and not a mis-sized bound - the log shows a genuine stall, which is the event #2442 is about:

```
+30s  Ign:2 http://azure.archive.ubuntu.com/ubuntu noble InRelease   (x3, at +30/33/37s)
+38s  Hit:2 https://archive.ubuntu.com/ubuntu noble InRelease         <- fallback reached
+38s  Get:3/4/5 noble-updates, noble-backports, noble-security InRelease
      <-- 142 seconds of silence -->
+180s ##[error]Process completed with exit code 124
```

So the bound did its job: an invisible stall became a named failure at 180s instead of drifting toward the job ceiling. What was missing is that `timeout` **only** produces an exit status - `agent-api-check.yml` had no loop reading it, so a transient stall became a hard red rather than a retried attempt.

Changed: gave that step the same 3-attempt loop `test-lint.yml` has, at 150s per attempt so 3x(150+5) plus the install stays inside the step's 9-minute bound. The guard's clause 2 now covers both workflows (33 cases, up from 32), and the docstring records the pairing as the contract: a bounded apt-get belongs in a loop that reads its status, because the bound makes failure *fast* and only the loop makes it *recoverable*.

### DIFF
--- a/.github/workflows/agent-api-check.yml
+++ b/.github/workflows/agent-api-check.yml
@@ -59,7 +59,25 @@ jobs:
         # over the last 30 runs, whose p50 is 15s.
         timeout-minutes: 9
         run: |
-          sudo timeout 180 apt-get update
+          # Bounded and retried for the reason test-lint.yml's step is: the
+          # mirrors here stall as well as fail, and `timeout` only converts the
+          # stall into exit 124 - something has to act on that exit, or a
+          # transient stall becomes a hard red. Observed on this very step: the
+          # azure.archive mirror was ignored after 3 tries, the fallback to
+          # archive.ubuntu.com then went silent for 142s mid-InRelease, and the
+          # bound ended it at 180s with nothing left to retry it.
+          #
+          # 150s x 3 attempts, tighter than test-lint.yml's 180 because this
+          # job's budget is 10 minutes rather than 45: 3x(150+5) plus the
+          # install is ~8 min, so the worst case still lands inside the step's
+          # own 9 and the reap keeps naming this step.
+          for attempt in 1 2 3; do
+            if sudo timeout 150 apt-get update; then
+              break
+            fi
+            echo "apt-get update failed or hung (attempt ${attempt}/3), retrying in 5s..."
+            sleep 5
+          done
           sudo apt-get install -y libosmesa6-dev
 
       # Importing the sim backend needs the sim-mujoco extra. CPU-only backend,
--- a/.github/workflows/agent-api-check.yml
+++ b/.github/workflows/agent-api-check.yml
@@ -52,8 +52,14 @@ jobs:
           cache: 'pip'
 
       - name: Install system dependencies (OpenGL for MuJoCo)
+        # Same class as test-lint.yml's apt step (#2442): unbounded, an apt-get
+        # facing a wedged mirror spends this job's whole 10-minute budget and
+        # then reports a rollup carrying no reason. 9 keeps the reap on the step
+        # - which names it - and clears the observed maximum of 474s (7m54s)
+        # over the last 30 runs, whose p50 is 15s.
+        timeout-minutes: 9
         run: |
-          sudo apt-get update
+          sudo timeout 180 apt-get update
           sudo apt-get install -y libosmesa6-dev
 
       # Importing the sim backend needs the sim-mujoco extra. CPU-only backend,
--- a/.github/workflows/test-lint.yml
+++ b/.github/workflows/test-lint.yml
@@ -35,6 +35,15 @@ jobs:
           cache: 'pip'
 
       - name: Install system dependencies (OpenGL for MuJoCo)
+        # The step's own ceiling, so a stall is reaped here rather than at the
+        # job's 45 - only the step-level bound marks the step that stalled, and
+        # a job-level reap renders as a rollup FAILURE with no reason (#2442).
+        # Sized from the observed band, which is narrow at both ends: 31
+        # successful runs put this step at p50 33s and max 455s (7m35s), so 12
+        # clears the maximum by ~1.6x, while the rest of the job costs up to
+        # ~32 min, so anything above ~13 lets a slow apt push the job into the
+        # 45-minute reap it is meant to prevent.
+        timeout-minutes: 12
         run: |
           # The GitHub runner image preconfigures third-party apt repos (e.g.
           # the packages.microsoft.com azure-cli feed) whose mirrors sometimes
@@ -43,11 +52,24 @@ jobs:
           # whole build. The packages we actually need (libosmesa6-dev, ffmpeg)
           # come from the standard Ubuntu archive, so retry a few times with a
           # short backoff to ride out transient mirror hiccups before failing.
+          #
+          # The loop is driven by apt-get's *exit status*, so it rides out a
+          # mirror that answers wrongly but not one that never answers: on a
+          # hang the `if` never evaluates, attempt 2 is never reached, and the
+          # echo that would have named apt never prints. `timeout` turns the
+          # hang into exit 124 - the failure this loop already handles - so the
+          # backoff and the diagnostic start working for that case too.
+          #
+          # 180s bounds `update` only, which is the command the loop retries.
+          # It is not a bound on the download: `update` costs ~5s even on the
+          # slow runs (the 455s maximum above was `install` fetching ffmpeg's
+          # ~115 packages), so a bound tight enough to be useful here would
+          # reap a healthy `install`. `install` is bounded by the step instead.
           for attempt in 1 2 3; do
-            if sudo apt-get update; then
+            if sudo timeout 180 apt-get update; then
               break
             fi
-            echo "apt-get update failed (attempt ${attempt}/3), retrying in 5s..."
+            echo "apt-get update failed or hung (attempt ${attempt}/3), retrying in 5s..."
             sleep 5
           done
           sudo apt-get install -y libosmesa6-dev ffmpeg
--- a/changelog.d/2442-apt-steps-are-bounded.md
+++ b/changelog.d/2442-apt-steps-are-bounded.md
@@ -9,11 +9,13 @@ but not one that never answers - on a hang the `if` never evaluates, the next
 attempt and the backoff are unreachable, and the diagnostic `echo` never
 prints.
 
-The looped `apt-get update` is now wrapped in `timeout`, which turns a hang
+Each retried `apt-get update` is now wrapped in `timeout`, which turns a hang
 into the non-zero exit the loop already handles, and both apt steps
 (`test-lint.yml`, `agent-api-check.yml`) declare their own `timeout-minutes` so
 a stall is reaped on the step that caused it rather than aggregated into a
-rollup `FAILURE` carrying no reason. Bounds are sized from the observed
+rollup `FAILURE` carrying no reason. `agent-api-check.yml` gains the retry loop
+as well: `timeout` only converts a stall into an exit status, so without
+something reading that status a transient stall becomes a hard failure. Bounds are sized from the observed
 distribution rather than from the command's nominal cost: `update` finishes in
 ~5s even on slow runs, while `install` fetching ffmpeg's ~115 packages reached
 455s of a 45-minute budget, so `install` is bounded by the step and not by a
--- a/changelog.d/2442-apt-steps-are-bounded.md
+++ b/changelog.d/2442-apt-steps-are-bounded.md
@@ -1,0 +1,21 @@
+### Fixed: a CI apt step that hangs is bounded by the step, not by the job
+
+`Test and Lint` reported `FAILURE` having never run a test, twice inside one
+80-minute window: both jobs were reaped by the job's own `timeout-minutes: 45`
+while still inside `Install system dependencies`, with lint and tests `skipped`
+and nothing in the log naming apt. The step's retry loop is driven by
+`apt-get update`'s exit status, so it rides out a mirror that answers wrongly
+but not one that never answers - on a hang the `if` never evaluates, the next
+attempt and the backoff are unreachable, and the diagnostic `echo` never
+prints.
+
+The looped `apt-get update` is now wrapped in `timeout`, which turns a hang
+into the non-zero exit the loop already handles, and both apt steps
+(`test-lint.yml`, `agent-api-check.yml`) declare their own `timeout-minutes` so
+a stall is reaped on the step that caused it rather than aggregated into a
+rollup `FAILURE` carrying no reason. Bounds are sized from the observed
+distribution rather than from the command's nominal cost: `update` finishes in
+~5s even on slow runs, while `install` fetching ffmpeg's ~115 packages reached
+455s of a 45-minute budget, so `install` is bounded by the step and not by a
+per-command timeout that would reap healthy runs. Pinned by
+`tests/test_apt_steps_are_bounded.py`.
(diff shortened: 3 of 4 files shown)


yinsong1986 (dismissed):
## Summary

Bounds the two CI apt steps so a wedged mirror fails fast on the step that names it instead of being reaped by the job ceiling with no reason attached: the looped `apt-get update` in `test-lint.yml` gains a per-attempt `timeout 180` (converting a hang into the exit-124 failure the retry loop already handles), and both apt steps (`test-lint.yml`, `agent-api-check.yml`) gain a step-level `timeout-minutes` (12 and 9) sized from the measured duration distribution rather than the issue's suggested values, which would have reaped healthy runs in the observed 455s/474s tail. `install` deliberately keeps no per-command bound, with the measured reason inline. The change is CI-only - no public API, wire format, or persisted state is touched.

## What's good

- Verified at head: `tests/test_apt_steps_are_bounded.py` passes (32 passed), both workflows parse under `yaml.safe_load`, and the parsed bounds match the contract (step 12 < job 45; step 9 < job 10; 180s < both step bounds).
- The worst-case loop cost (3 x 180s + backoff, ~9.2 min) only approaches the 12-minute step bound when all three update attempts hang - a state where failing is the correct outcome; the healthy path (update ~5s + install max 455s) clears it comfortably.
- The pin is keyed on the class (every `apt-get` step in every workflow), asserts its own parser's non-vacuity per file, and follows the neighbouring line-based-parse convention (`test_workflow_jobs_are_bounded.py`) with the mypy/`types-PyYAML` reason documented.
- Changelog fragment follows `changelog.d/README.md` (issue number explicitly allowed); no non-ASCII, no host paths in the new files.

yinsong1986 (approved):
## Summary

Bounds the two CI apt steps so a wedged mirror fails fast on the step that names it, rather than being reaped by the job ceiling as an unexplained rollup `FAILURE` (#2442): each retried `apt-get update` gains a per-attempt `timeout` (180s in `test-lint.yml`, 150s in `agent-api-check.yml`) that converts a hang into the exit-124 failure the retry loop already handles, both apt steps declare a step-level `timeout-minutes` (12 inside the 45-minute job, 9 inside the 10-minute one) sized from the measured duration distribution rather than the issue's suggested values, and `agent-api-check.yml` gains the retry loop it previously lacked - closing the gap round 1's own CI demonstrated, where a bounded-but-unretried stall became a hard red. `install` deliberately keeps no per-command bound, with the measured justification (455s/474s healthy tail) recorded in the step comments.

## What's good

- The new guard (`tests/test_apt_steps_are_bounded.py`) pins the class rather than the two sites, and pins its own non-vacuity: the structural parse is cross-checked against a raw text scan per file, and clause 2 is asserted non-empty. Verified 33/33 passing at `919a1b3c`.
- Bound arithmetic checks out both ways: each step bound is strictly below its job's (12 < 45, 9 < 10), each command bound below its step's, and the worst-case retry path (3x(150+5) + install) stays inside `agent-api-check.yml`'s 9-minute step bound so the reap keeps naming the step.
- Changelog fragment follows the `changelog.d/<number>-<slug>.md` convention (issue number is explicitly allowed by `changelog.d/README.md`); all new strings are ASCII; no host paths; the neighbouring `test_workflow_jobs_are_bounded.py` contract is untouched and still passes.
- The retry loop's fall-through to `apt-get install` after three failed updates is the pre-existing pattern, unchanged in semantics - this PR only makes the hang case reach the loop at all.

PR #2458, opened 2026-08-19
docs(policies/cosmos3): name the receiver that accepts the safety-checker flag

cagataycali:
## Problem

The safety-checker callout in `docs/policies/cosmos3.md` tells the reader:

> Install `cosmos_guardrail` and pass `enable_safety_checker=True` to re-enable it.

No surface that page names accepts it. Every worked example on the page builds the
policy - `Cosmos3Policy(...)` or `create_policy("cosmos3", ...)` - and that
constructor declares no `enable_safety_checker` parameter and takes no `**kwargs`.
The flag belongs to `Cosmos3DiffusersBackend`, which the page never mentions.

Applying the instruction on every route a reader of that page has:

| route | before |
| --- | --- |
| `Cosmos3Policy(..., enable_safety_checker=True)` | `TypeError: Cosmos3Policy.__init__() got an unexpected keyword argument 'enable_safety_checker'` |
| `create_policy("cosmos3", ..., enable_safety_checker=True)` | same `TypeError` |
| the registry route (`build_policy_kwargs`) | accepted, then dropped - returns `['backend', 'embodiment', 'host', 'mode', 'port']`, so the run proceeds with the checker off |
| `Cosmos3Policy(diffusers_backend=...)` | carries the flag, but the page never names it |

So 0 of 4 routes deliver the setting. The third row is the quiet one: the
provider's `config_keys` filter keeps only the keys the constructor declares, so a
caller who "enabled" the checker through a config dict gets a successful run with
it disabled and nothing said.

The cause is that `Cosmos3Policy` forwards only `embodiment`, `model` and `mode`
of the backend's 13 parameters, so `resolution_tier`, `view_point`, `device`,
`dtype`, `num_inference_steps`, `guidance_scale`, `enable_sound` and
`enable_safety_checker` are all reachable on the backend object only.

## Change

Docs and a docstring; no behaviour change.

* The callout now names the class the flag belongs to and shows the route that
  carries it - build a `Cosmos3DiffusersBackend` with the flag and hand it to
  `Cosmos3Policy(diffusers_backend=...)` - and lists the other load and sampling
  knobs the same route carries, noting that the policy forwards only
  `embodiment`, `model` and `mode`.
* `Cosmos3Policy`'s `diffusers_backend` docstring said "(dependency injection for
  tests; skips the heavy import)". Injection is still how the tests drive this
  backend, but it is also the supported route to those knobs, so the docstring now
  says both.

The alternative - promoting the eight knobs onto `Cosmos3Policy` - changes a public
constructor and the provider's `config_keys`, and picking which subset belongs
there is a design call rather than a correction, so it is out of scope here.

## Tests

`tests/policies/cosmos3/test_documented_backend_knob_routes.py` grades every
keyword the page instructs against `inspect.signature` of the receivers the page
actually names, so a knob promoted onto the policy later, or a newly documented
one, is graded without editing the test. The page's Python fences are graded in
the other direction: every keyword they pass must be accepted by that call's own
receiver.

Against `upstream/main`: **3 failed, 5 passed**. Afterwards: **8 passed**.

The headline failure reports the offender rather than a wording mismatch:

```
cosmos3.md tells the reader to pass ['enable_safety_checker'], which no receiver
it names accepts. Named receivers: ['Cosmos3Policy', 'create_policy("cosmos3")'].
Name the class the keyword belongs to, or promote the keyword.
```

The five that pass on both trees are the controls:

* every keyword the page's Python fences pass is accepted by its own receiver
  (19 graded, so the examples themselves are pinned as correct);
* `model` and `mode` still reach the backend, which is the subset the policy carries;
* the registry route still returns exactly its declared `config_keys` - this is why
  the page names an object route rather than a config key;
* a knob list that named a parameter `Cosmos3DiffusersBackend` does not have would
  be reported, with a planted case so a clean result is not vacuous;
* a planted instruction is picked up by the extractor, so a clean sweep means the
  page is right rather than that the grader accepts anything.

## Verification

Measured at `c51da2ef` against `upstream/main` `ae57b3fa`.

* `mypy strands_robots tests tests_integ`: 24 pre-existing errors on both trees,
  none in the changed files.
* Blast radius (all of `tests/policies/cosmos3`, every test touching
  `Cosmos3Policy` / `config_keys` / `build_policy_kwargs` / a docs page, plus the
  repo-wide docs, docstring, source-string and changelog guards - 80 files):
  branch **1 failed / 2743 passed**, base **4 failed / 2740 passed**, 2744
  collected on both. Branch-only failures: none. Base-only: exactly the 3 pre-fix
  failures above. The single shared failure is
  `tests/training/test_lerobot_e2e.py::test_record_train_load_loop`, which fails
  the same way on both trees from a local `draccus` older than the `lerobot` pin
  (`TypeError: encode() takes 1 positional argument but 2 were given`).
* `scripts/assemble_changelog.py --check`: OK.

## Artifact

One capture script per tree, reading the instruction out of whichever page it runs
against and applying it on all four routes. The package code is identical in both
columns; only the page differs.

documented knob route

Routes delivering the flag: 0/4 before, 1/4 after.

### DIFF
--- a/changelog.d/2458-cosmos3-documented-backend-knob-route.md
+++ b/changelog.d/2458-cosmos3-documented-backend-knob-route.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **policies/cosmos3**: the safety-checker note in `docs/policies/cosmos3.md` told readers to
+  `pass enable_safety_checker=True`, which no surface the page names accepts - `Cosmos3Policy`
+  declares no such parameter and takes no `**kwargs`, so both documented routes raise `TypeError`,
+  and the registry route filters the key out and runs with the checker off. The note now names the
+  `Cosmos3DiffusersBackend` parameter it is, with a snippet handing that backend to
+  `Cosmos3Policy(diffusers_backend=...)`, and lists the other load and sampling knobs the same
+  route carries. `Cosmos3Policy`'s own `diffusers_backend` docstring no longer calls itself
+  test-only injection.
--- a/docs/policies/cosmos3.md
+++ b/docs/policies/cosmos3.md
@@ -111,10 +111,29 @@ The `cosmos3-diffusers` extra is native `diffusers` + `torch` + `transformers`
 > `cosmos_guardrail` package and otherwise raises `ImportError: cosmos_guardrail
 > is not installed`. The diffusers backend disables it by default
 > (`enable_safety_checker=False`, passed through to `from_pretrained`) so the
-> pipeline loads without that extra. Install `cosmos_guardrail` and pass
-> `enable_safety_checker=True` to re-enable it. Note Cosmos runs in `bfloat16`,
-> so the backend up-casts the half-precision action tensor to `float32` before
-> returning the chunk.
+> pipeline loads without that extra. To re-enable it, install `cosmos_guardrail`
+> and build the backend with `enable_safety_checker=True`, then hand that backend
+> to the policy - the flag is a `Cosmos3DiffusersBackend` parameter, not a
+> `Cosmos3Policy` one:
+>
+> ```python
+> from strands_robots.policies.cosmos3.embodiments import get_embodiment
+> from strands_robots.policies.cosmos3.policy import Cosmos3Policy
+> from strands_robots.policies.cosmos3.policy_diffusers import Cosmos3DiffusersBackend
+>
+> backend = Cosmos3DiffusersBackend(
+>     embodiment=get_embodiment("droid"),
+>     model="nvidia/Cosmos3-Nano",
+>     enable_safety_checker=True,   # needs cosmos_guardrail installed
+> )
+> policy = Cosmos3Policy(embodiment="droid", backend="diffusers", diffusers_backend=backend)
+> ```
+>
+> `Cosmos3Policy` forwards only `embodiment`, `model` and `mode` to the backend, so
+> the same route is how you reach its other load and sampling knobs
+> (`resolution_tier`, `view_point`, `device`, `dtype`, `num_inference_steps`,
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Docs-and-docstring correction with a self-updating guard test, no behaviour change. The safety-checker callout in `docs/policies/cosmos3.md` previously instructed `enable_safety_checker=True` against a page whose only named surfaces (`Cosmos3Policy` / `create_policy("cosmos3")`) do not accept it — two routes `TypeError`, and the registry route silently filters the key out and runs with the checker off. The page now names `Cosmos3DiffusersBackend` as the owner, shows the `diffusers_backend=` injection route, and lists the other backend-only knobs; the `diffusers_backend` docstring no longer calls itself test-only injection. I verified the new snippet's keywords against the real `Cosmos3DiffusersBackend.__init__` and `Cosmos3Policy.__init__` signatures — all match — and reproduced the test verdicts locally at `c51da2ef`: 8 passed on the branch, and with the base doc/docstring restored exactly the 3 claimed pre-fix failures (the headline one naming `enable_safety_checker` and the receivers that refuse it).

## What's good

- The guard test grades the page against `inspect.signature` of the receivers the page itself names, so a later promotion of a knob onto the policy — or a newly documented knob — is graded without editing the test.
- Both planted-case controls (`test_the_grader_reports_a_planted_instruction`, the planted knob list) keep a clean sweep from being vacuous, per AGENTS.md's pin-the-fix convention.
- Scope discipline: the promote-the-knobs alternative is correctly deferred as a design call rather than smuggled into a docs fix; changelog fragment present under `changelog.d/` per step 3.
- Test hygiene is clean: no host paths, ASCII-only, no shared-cache reads, stubs (`pipeline=`, `condition_cls=`) keep it runnable without the heavy `diffusers` extra.

PR #2468, opened 2026-08-19
fix(mesh): refuse an ACL file the JSON5 parser cannot descend

cagataycali:
## Problem

`_parse_json5` is the ACL loader's parse boundary and its docstring states the contract:

> Raises `ValueError` with operator-friendly diagnostics on **any** malformed input. The ACL loader treats this as a fail-closed boundary: a malformed file does NOT silently degrade to the permissive default.

`json5` is a recursive-descent parser, so a deeply nested document exhausts the interpreter stack and raises `RecursionError`. That is a `RuntimeError`, so it is outside the `ValueError` this boundary documents *and* outside the `(OSError, ValueError)` tuple both fail-closed callers narrow to:

- `snapshot_acl` reports an unloadable file as permissive precisely so the start-time gate refuses the wire.
- `Mesh.start`'s ACL gate does the same, and its comment reserves wider classes for "genuine bugs so they aren't masked at WARNING level".

An unreadable ACL file is not a bug in this library, so it was misclassified. The refusal escaped both handlers, and took the path, the reason and the remedy with it.

Measured against `48d58c02`, driving `_load_acl_file`, `snapshot_acl` and the gate:

| ACL file | size | `_parse_json5` | `snapshot_acl` | `Mesh.start` gate |
| a valid role-separated ACL | 480 B | loaded | `permissive=False` | wire may start |
| a malformed ACL (`{a:`) | 3 B | `ValueError` | `permissive=True` | refuses the wire |
| an ACL nested 200 deep | **400 B** | **`RecursionError`** | **escapes** | **escapes** |

### `ACL_FILE_MAX_BYTES` is not the bound that covers this

The module already bounds a hostile file, with the reason written next to it ("certainly an attacker probing for an OOM"), capping the read at 256 KiB. But **nesting depth, not size, is what the parser runs out of room for**: the shallowest document that trips it is 60 levels, i.e. **120 bytes - 2185x under that budget** - while both shipped operator templates (`mesh_acl_example.json5`, `mesh_acl_strict_per_peer.json5`) nest **4** levels.

### It is not a version to pin away

Every `json5` release the declared floor (`>=0.9.0`) admits raises it identically, so there is no floor that fixes it:

| json5 | deep arrays | deep objects | malformed |
| --- | --- | --- | --- |
| 0.9.0 / 0.9.14 / 0.9.28 / 0.10.0 / 0.12.0 / 0.14.0 / 0.15.0 | `RecursionError` | `RecursionError` | `ValueError` |

`RecursionError` is also the *only* non-`ValueError` class the parser raises: a battery over deep arrays/objects/unclosed-arrays, oversized escapes, a 60k-digit integer, `bytes`, `None`, NUL bytes, a BOM and a lone surrogate produced no other escape.

## Change

`_parse_json5` converts `RecursionError` into the `ValueError` it already documents, so the established fail-closed path handles it and every downstream narrow handler starts working. One handler, at the boundary that owns the contract.

The message does not reuse the neighbouring "not valid JSON5" wording: a deeply nested document *is* valid JSON5 that this parser cannot read, and reporting it as malformed syntax would send the operator looking for a typo that is not there. It names the file, the cause, the shape a real ACL has, and the template:

```
ACL file <path> is nested too deeply to parse: the JSON5 parser exhausted the
interpreter stack. A role-separated ACL nests about four levels -- see
examples/mesh/mesh_acl_example.json5.
```

Raising the recursion limit or pre-checking a maximum depth would each invent a new bound, so both are out of scope; this only restores the class the boundary already promises.

`docs/security.md`'s ACL bullet list documented the permissive default and the opt-out but never what happens to a file the loader cannot read, so it gains that bullet.

## Tests

Added to `tests/mesh/test_acl_config.py`, extending the sibling `TestJSON5MalformedFailsLoudly` - which already pins this contract for input the parser *can* descend, and whose four existing cases are untouched.

**6 failed / 67 passed before, 73 passed after.** Every pre-fix failure is behavioural, reporting the raw `RecursionError: maximum recursion depth exceeded`.

Fails pre-fix:
- the refusal arrives as a `ValueError`, parametrized over the three measured shapes (deep arrays, deep objects, deep unclosed arrays)
- the refusal names the file and points at the operator template
- the refusal does not report valid JSON5 as malformed
- `snapshot_acl` reports it permissive so the gate runs, instead of the exception escaping

Passes on both trees, i.e. the controls:
- a shallow malformed file keeps its own message byte for byte - fails if the two branches are collapsed
- a valid ACL still loads - fails if the new handler reaches past the parse call
- an oversize file is still refused on size - pins the ordering
- the payload is orders of magnitude inside `ACL_FILE_MAX_BYTES` - the severity gap as an assertion

## Verification

Measured at `4fb6562f` against `upstream/main` `48d58c02`, same interpreter, 190 test files (all of `tests/mesh`, every test touching the ACL loader or a docs page, and the repo-wide changelog/docstring/ASCII/dependency guards):

- branch **21 failed / 4260 passed / 5 skipped / 24 errors**; base **27 / 4254 / 5 / 24**; both collected **4310**
- `27 - 21 == 4260 - 4254 == 6` == exactly the pre-fix failure count
- **branch-only failures: none.** Base-only: exactly the 6 above. The 45 shared failures are all `awsiot`-absent `tests/mesh/test_iot_*`, which CI installs.
- `ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` 24 == 24 against a pristine `upstream/main` worktree, branch-only empty
- `scripts/assemble_changelog.py --check` OK

## Artifact

Both columns are the same probe script run against the two trees, driving the real `_load_acl_file`, `snapshot_acl` and `Mesh.start` gate. The first two rows are byte-identical in both trees - only the file the parser cannot descend changes behaviour.

ACL nesting refusal, before and after

### DIFF
--- a/changelog.d/2468-acl-file-nesting-refusal.md
+++ b/changelog.d/2468-acl-file-nesting-refusal.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **mesh**: an ACL file too deeply nested for the JSON5 parser to read is now refused as an unloadable ACL instead of escaping the loader's fail-closed boundary as a `RecursionError`, so the start-time gate still refuses to bring the wire up and the operator gets the path and a remedy.
--- a/docs/security.md
+++ b/docs/security.md
@@ -56,6 +56,7 @@ mTLS alone is not sufficient - pair it with an access-control list:
 - The built-in default ACL is permissive: any CA-signed peer may publish and subscribe on any key. If you forget to supply an ACL, the SDK warns on every session open.
 - Supply an operator ACL via `STRANDS_MESH_ACL_FILE` that enumerates each peer's certificate CN and the key expressions it may use. See [`examples/mesh/mesh_acl_example.json5`](https://github.com/strands-labs/robots/blob/main/examples/mesh/mesh_acl_example.json5) and [`examples/mesh/mesh_acl_strict_per_peer.json5`](https://github.com/strands-labs/robots/blob/main/examples/mesh/mesh_acl_strict_per_peer.json5).
 - `STRANDS_MESH_ACCEPT_PERMISSIVE_ACL=1` exists only to silence the permissive-ACL warning when you have deliberately accepted it (e.g. a closed lab). Do not set it in production - it does not make the mesh safer, it only quiets the reminder that it is not.
+- An ACL file the loader cannot read is refused, not ignored. A missing, oversize, non-UTF-8, malformed, or too-deeply-nested file is reported as an unloadable ACL, which the start-time gate treats as the permissive default and therefore refuses to bring the wire up. The error names the path and the reason, so a typo in the ACL stops the mesh rather than quietly widening it.
 
 > ⚠️ **WAN/cloud Zenoh routers MUST deploy a topic-level ACL.**
 > mTLS authenticates *who* a device is; it does **not** restrict *what* topics
--- a/strands_robots/mesh/_acl_config.py
+++ b/strands_robots/mesh/_acl_config.py
@@ -98,6 +98,13 @@ def _parse_json5(raw: str, path: Path) -> Any:
     malformed input. The ACL loader treats this as a fail-closed
     boundary: a malformed file does NOT silently degrade to the
     permissive default.
+
+    That includes a document the parser cannot descend. ``json5`` is a
+    recursive-descent parser, so a deeply nested file raises
+    :class:`RecursionError` rather than :class:`ValueError` -- a
+    :class:`RuntimeError`, outside both the class this boundary documents and
+    the ``(OSError, ValueError)`` tuple its two fail-closed callers narrow to.
+    It is converted here so the refusal reaches them.
     """
     # Use the project-standard
     # ``require_optional`` helper so the operator-facing import error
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

This PR closes a real gap in the ACL loader's fail-closed contract: `json5` is a recursive-descent parser, so a deeply nested ACL file (well under `ACL_FILE_MAX_BYTES` -- 60 levels is ~120 bytes) raises `RecursionError`, which as a `RuntimeError` escaped both the `ValueError` the `_parse_json5` docstring promises and the `(OSError, ValueError)` tuple that `snapshot_acl` and the `Mesh.start` gate narrow to. The fix converts `RecursionError` into the documented `ValueError` at the one boundary that owns the contract, wrapping only the `loads()` call so the handler cannot reach past the parse, with a distinct message that correctly avoids reporting valid-but-undescendable JSON5 as a syntax error. I reproduced the escape on base (`48d58c02`) through both `_load_acl_file` and `snapshot_acl` and confirmed the fail-closed path is what the conversion restores.

## What's good

- The escape is verified against the real callers, not just the parse function; the `snapshot_acl` test pins the consequence the fix exists for (gate runs instead of crashing).
- Controls pin the neighbouring branches: shallow-malformed keeps its own message, valid ACL still loads, size-cap ordering is preserved.
- Tests use `monkeypatch.setenv` and clear both cache tiers, per AGENTS.md testing conventions; error message is ASCII-only.
- Changelog fragment present; scope is tight -- one handler at the documented boundary, no new depth bound invented.

PR #2473, opened 2026-08-19
fix(hardware): check the ROS 2 command-surface posture flags instead of reading them by truthiness

cagataycali:
## Problem

`Robot(ros2_bridge=True)` can expose an inbound `/<robot>/joint_command` topic that drives a physical arm. `docs/security.md` names the flag that closes it as the ungated posture:

> **Telemetry-only is ungated.** `ros2_commands=False` is publish-only (no inbound surface) and needs no security config.

Both flags were read by truthiness (`self._ros2_bridge_enabled = bool(ros2_bridge)`, `self._enable_commands = bool(enable_commands) and robot is not None`). Every non-empty string is truthy, so `"false"`, `"no"`, `"off"` and `"0"` - the spellings a YAML/env deployment config yields for a disabled feature - each selected the *permissive* posture, and `None`, `0`, `""`, `[]` took the other branch without ever being a declared spelling of it. Nothing raised and nothing logged.

Measured through `Robot._init_ros_bridge` (the documented route) with the operator insecure opt-out set, which is the posture `docs/security.md` tells a telemetry-only caller they may use:

| request | main | this PR |
| --- | --- | --- |
| `ros2_commands="false"` (read-only) | subscribed `rt/so101/joint_command`, **90 inbound commands driven into `send_action`** | refused at construction, 0 commands |
| `ros2_bridge="false"` (no bridge) | `HardwareRtpsBridge` built and - because `ros2_commands` defaults to True - subscribed to `rt/so101/joint_command` | refused at construction |
| `ros2_commands=False` | 0 commands, no subscription | 0 commands, no subscription (unchanged) |
| `ros2_commands=True` | all 90 commands delivered | all 90 commands delivered (unchanged) |

Two further consequences fall out of the same read:

- **The DDS Security gate inherits the inversion.** `_require_secure_command_surface` branches on this flag, so without a security config a truthy non-boolean made it refuse a *read-only* request with "An enabled command bridge lets any DDS participant drive the physical arm" and advise `STRANDS_ROS2_BRIDGE_I_KNOW_THIS_IS_INSECURE=1` - the one remedy that turns that refusal into a silent open of the surface the caller asked to close.
- **The rclpy transport has no such gate** (it gets DDS Security from the RMW keystore, by design), so there the read-only request just became a live arm-driving subscription.

## Change

Route the three entry points that accept this posture through the shared `boolean_flag_error` domain - the one `AGENTS.md` already names for "a confirmation gate, a security opt-out, a preview mode, a search region", and whose own docstring calls out "a security opt-out that fails open":

- `Robot._init_ros_bridge` - `ros2_bridge` and `ros2_commands`, checked before the first attribute assignment and before either bridge is constructed.
- `HardwareRosBridge.__init__` - `enable_commands`, beside the existing `spin_period` guard.
- `HardwareRtpsBridge.__init__` - `enable_commands`, beside the existing `domain_id` / `poll_period` guards.

Both bridges are in `strands_robots.__all__`, so each checks the flag it receives rather than trusting its caller. Each guard lands ahead of its transport dependency and ahead of the process-wide `ROS_DOMAIN_ID` write - the placement `spin_period` and `domain_id` already document - so a refused flag reports identically with and without the `[ros2]` extra and leaves no bridge state, no environment write and no DDS participant behind.

`bool(...)` is kept *after* the guard, not instead of it: the shared domain accepts `np.bool_`, and the conversion is what makes an accepted value consumable.

## Not in scope

`SimEngine._init_ros_bridge`'s `ros2_bridge` is the same spelling on a different contract: `SimRosBridge` has no `joint_command` subscription at all (measured: zero occurrences), so that flag carries no command surface and belongs to its own change.

## Rollout

Left panel is `upstream/main`, right panel is this branch. One capture script per tree drives the shipped `Robot._init_ros_bridge` -> reader -> `_drive_from_command` -> `send_action` path with `ros2_commands="false"`; only cyclonedds itself is stood in for. so101 in MuJoCo then replays exactly the joint targets each bridge accepted.

A read-only bridge driving the arm on main, refused on this branch

[Full-resolution MP4](https://raw.githubusercontent.com/cagataycali/robots/media-ros2-command-surface-flag/ros2-command-surface-flag.mp4)

Both panels start from rest, pixel-identical (panel-vs-panel mean abs level **0.0000** on the first frame). The left panel then moves on **44 of 44** frame transitions (per-frame mean abs level 2.47 to 34.20) as the "read-only" bridge drives `shoulder_pan` to +1.55 rad / +89 deg. The right panel is **0.0000 on every transition** - the flag is refused at construction, so nothing is ever subscribed and the arm never moves.

## Tests

`tests/test_ros2_command_surface_flag_domain.py`, named and shaped like the five sibling flag-domain suites, parametrizing over `boolean_flag_error` itself rather than a copied spelling list so a spelling the shared domain grows is covered without an edit here.

Against `upstream/main`: **37 failed / 5 passed**. After: **42 passed**. The headline failure is a measurement, not a wording check:

```
AssertionError: ros2_commands='false' asks for a read-only telemetry bridge, yet an inbound
arm-driving subscription is live on 'rt/so101/joint_command' (enable_commands=True); any
participant on the DDS domain can now drive the arm
```

The 5 tests that pass on both trees are the no-overreach controls, and each one fails for a specific shortcut fix: `ros2_commands=False` stays read-only and `=True` still opens the surface (fails if the guard is made unconditional); `np.bool_(False)` still selects the posture it names (fails for a hand-rolled `isinstance(x, bool)`); `ros2_bridge=False` still builds nothing; and `ros2_domain` / `poll_period` / `spin_period` keep their own verdicts (fails if the new check displaces the guards it was added beside).

## Verification

Measured at `14324ddc` (this branch, with `upstream/main` merged in), on a 44-file selection covering every test that names the changed symbols, all five sibling flag-domain suites, and the repo-wide docstring / string / changelog / export guards:

- branch **1906 passed, 0 failed, 1 skipped**; `upstream/main` with the new test file copied in **37 failed, 1869 passed, 1 skipped**. `1869 + 37 = 1906` on the same collected total, the passed delta equals the failed delta equals the 37 pre-fix failures, and the branch-only failure set is empty.
- `mypy strands_robots tests tests_integ`: 24 on the branch, 24 on a pristine `upstream/main` worktree, branch-only set empty.

## Docs

- `docs/security.md` - the "Telemetry-only is ungated" bullet now states that the posture is only as good as how the flag is read, and that a non-boolean is refused.
- `docs/ros2-integration.md` and `docs/rtps-integration.md` - the `ros2_commands=False` paragraphs, the second noting that the DDS Security gate branches on the same flag.
- `AGENTS.md` - the existing "Posture flags are checked, never read by truthiness" convention gains this instance in its truthiness and refusal-inheritance bullets and the new suite in its `Pinned by` list.
- `Args` and `Raises` on all three entry points.


yinsong1986 (approved):
## Summary

Routes the three entry points that decide whether a hardware robot exposes an inbound `joint_command` surface (`Robot._init_ros_bridge`'s `ros2_bridge`/`ros2_commands`, and `enable_commands` on both exported bridge classes) through the shared `boolean_flag_error` domain, so truthy off-spellings (`"false"`, `"no"`, `"off"`, `"0"`) are refused at construction instead of silently opening the arm-driving surface they ask to close. I verified the guard placement against the base class: in both bridges the check lands ahead of the transport probe and the process-wide `ROS_DOMAIN_ID` write (`ros_telemetry.py` performs both in `super().__init__`), and in `_init_ros_bridge` it lands before the first attribute assignment and either bridge construction. `bool(...)` is correctly kept after the guard so `np.bool_` remains consumable, no in-tree caller passes a non-boolean, and the change is exactly the instance of the AGENTS.md "Posture flags are checked, never read by truthiness" convention it claims to be. New suite passes 42/42 locally; lint and mypy clean on the changed files.

## What's good

- Guard ordering is pinned by tests, not just comments: the RTPS test asserts zero DDS endpoints exist after a refusal, and the rclpy test asserts `ROS_DOMAIN_ID` is left unwritten - the exact placement contract the neighbouring `spin_period`/`domain_id` guards document.
- The five no-overreach controls each fail for a specific shortcut fix (unconditional refusal, hand-rolled `isinstance(x, bool)`, displaced numeric guards), which is the strongest shape for a guard-insertion PR.
- Suite parametrizes over `boolean_flag_error` itself per the convention, so a spelling the shared domain grows is covered without an edit here.
- Scope discipline: `SimEngine`'s same-named flag is explicitly measured as carrying no command surface and deferred, changelog is a fragment, docs updated in the same PR.

PR #2474, opened 2026-08-19
fix(examples/fleet): await the survivors' estop evidence instead of sampling the audit log once

### LINKED ISSUE #2469: bug(examples/fleet): 03 live fails deterministically - the resync reads the audit log inside the ~20ms window before the survivor's remote_estop_engaged row lands (in-process restart is instant)
# Symptom

`examples/fleet/03_failover_and_degraded_ops.py` (live mode) fails deterministically at its final assertion - 3/3 runs on a CPU box, zenoh 1.10.0, current `main`:

```
RuntimeError: re-sync did not recover the estop lockout from the audit log:
{'alive_peers': ['lekiwi-a1', 'so101-a1'], 'lockouts': {'lekiwi-a1': True}, ...}
```

The resync finds the issuer's lockout but not the survivor's, so `any(lockouts[robot] for robot in others)` is False.

# Root cause: the resync read races the survivor's asynchronous audit write, and the in-process "restart" makes the race deterministic

Timeline from an isolated run (`STRANDS_MESH_AUDIT_DIR` pointed at a fresh dir; row indices and `ts` deltas from the file):

```
13 t=+46.38s lekiwi-a1  emergency_stop            # issuer's own audit row
14 t=+46.38s lekiwi-a1  remote_estop_redundant
15 t=+46.40s so101-a1   remote_estop_engaged      # survivor's handler - LAST row in file
```

- The survivor's `remote_estop_engaged` row is written by `_on_safety_estop` on the zenoh callback thread, ~20 ms after the issuer's `emergency_stop()` returns.
- The example then "restarts" the orchestrator: `init_mesh(...)` in the SAME process gets the ref-counted shared session, and `peers_by_id` is backed by the process-shared presence registry that already contains the survivors - so `_wait_for(... presence discovery after the orchestrator restart)` returns immediately.
- `read_audit_log(since=run_start - 1.0)` therefore executes inside the ~20 ms window between row 13 and row 15: it sees the issuer's lockout row and not the survivor's. Read-side is exonerated: replaying the exact read post-hoc returns all 16 rows including `('so101-a1', 'remote_estop_engaged')`.

So the example asserts a durable property (lockout recoverable from the audit log) with a single point-in-time read taken milliseconds after the event that produces the evidence, and the in-process restart - instant by construction - guarantees the read wins the race. On a real fleet the restart takes seconds and the race hides; in the example it fails every time. The property is true; the observation method is wrong.

# Fix

Poll the resync predicate with a bounded deadline instead of reading once - which is also more faithful to what a real restarted orchestrator does:

```python
deadline = time.monotonic() + 10.0
while True:
    records = read_audit_log(since=run_start - 1.0)
    resync = resync_after_restart(orch_mesh2.peers, records)
    if any(resync["lockouts"].get(robot) for robot in others):
        break
    if time.monotonic() > deadline:
        raise RuntimeError(f"re-sync did not recover the estop lockout ...: {resync}")
    time.sleep(0.2)
```

(Deadline on `time.monotonic()` per the clocks convention.) Alternatively the example can wait for the survivors' own `remote_estop_engaged` rows to appear before restarting; the poll is simpler and keeps the drill's ordering honest.

# Repro

```
STRANDS_MESH_HITL_ACTIONS=none STRANDS_MESH_LOCAL_DEV=1 MUJOCO_GL=egl \
STRANDS_MESH_AUDIT_PSK=demo-psk STRANDS_MESH_OVERRIDE_CODE=demo-override \
STRANDS_MESH_AUDIT_DIR=$(mktemp -d) \
python examples/fleet/03_failover_and_degraded_ops.py
```

rc=1, ~50 s, identical resync dict every run. Examples 01, 02 and 05 pass live on the same box/env (5 s / 6 s / 3 s), so this is isolated to 03's final assertion.

# Adjacent finding (one line, separate concern)

With `eclipse-zenoh` not installed, `init_mesh` degrades to "mesh off" at DEBUG level only, so 02/03 fail as opaque 15 s presence-discovery timeouts with no hint at the cause. Worth a WARNING (or a raise in the examples, which require the mesh) - can be its own small issue if preferred.

Part of epic #2179 (example 03 = #2182).

cagataycali:
## Problem

`examples/fleet/03_failover_and_degraded_ops.py` part 2 kills the dispatcher, has a
surviving robot issue `emergency_stop`, restarts the orchestrator, and reconstructs the
outage from the two things that survived it - the presence registry and the signed audit
log. It then asserts the survivors' lockout is recoverable, and raises:

```
RuntimeError: re-sync did not recover the estop lockout from the audit log:
{'alive_peers': ['lekiwi-a1', 'so101-a1'], 'lockouts': {'lekiwi-a1': True}, ...}
```

The issuer's lockout is there; the survivor's is not.

A peer's own `remote_estop_engaged` row is written by *that peer's* `_on_safety_estop`, on
its callback thread, once the broadcast reaches it - after the issuing `emergency_stop()`
has already returned. `resync_after_restart` is a pure reduction over the records it is
handed, so it can only be as complete as the read that fed it. The drill read the log once,
immediately after the restart, and the restart is instant here: `init_mesh` returns the
ref-counted shared session and the presence registry already holds the survivors, so the
presence wait returns at once. The read then lands while the evidence is still arriving.

The property being asserted is true. The observation was a point-in-time sample.

## Measurements

Six live runs of the drill per side, each with `STRANDS_MESH_AUDIT_DIR` pointed at a fresh
directory, reading the resulting `mesh_audit.jsonl` afterwards:

| | part-2 assertion | audit rows the run produced |
|---|---|---|
| `main` (single read) | **5 of 6 fail** | 16 in 6/6 |
| this change (awaited) | **0 of 6 fail** | 17 in 6/6 |

Two details worth recording, because they constrain the fix:

- **It is a race, not a deterministic failure.** One of the six `main` runs passed. A fix
  may not rely on the read always losing.
- **The burst has no fixed ordering.** Across the 12 samples the survivor's row lands
  between **-14.4 ms and +30.6 ms** relative to the issuer's own row, and *before* it in 2
  of 12. So the survivor's row is not reliably "the last row", and no fixed sleep is
  correct.

The 16-vs-17 column is corroboration rather than the defect: sampling that early also means
the run ends before the survivor's own `command_rejected_lockout` row lands - the evidence
for the `execute` refusal the drill had already printed. Missing in 6/6 `main` runs, present
in 6/6 after.

re-sync race measurement

## Why the wait belongs on the observer

No issuer-side call can make a remote peer's write synchronous - that would require an ack
round-trip from every peer - so `emergency_stop()` cannot promise the trail is complete when
it returns. The observer is the only side that can wait.

The drill already does exactly that for everything else asynchronous, with the file's own
bounded `_wait_for`: presence discovery of all robots, the dispatcher dropping out of
presence, and presence discovery after the restart. The audit evidence was the one
asynchronous condition sampled instead of awaited.

## Change

`resync_until_lockout_recovered(...)` reads once, then polls until the reconstruction shows
the lockout, bounded on a monotonic clock:

- the **predicate and the `RuntimeError` text are unchanged**, so only the timing moves and
  an operator hitting a genuine failure still gets the reconstruction that fell short;
- a trail that is already complete returns on the first read, so the normal path costs no
  extra delay;
- `clock` and `sleep` are seams, matching `run_failover_drill`, so the wait is drivable in a
  test without sleeping;
- the deadline is taken from a monotonic clock, so a wall-clock step cannot extend or
  truncate the bound.

`_run_live` now reconstructs through it, which also removes the duplicated raise at the call
site. The dry-run path is untouched: it has no asynchronous estop evidence to wait for.

## Tests

`tests/test_fleet_failover_and_degraded_ops.py`: **5 failed / 8 passed** before, **13
passed** after.

The headline test drives the exact stimulus with a reader whose first call omits the
survivor's row and whose second includes it, and asserts both halves against the same trail:
reduced from the first read alone the survivor has no lockout, which is the drill's failure
mode; awaited, the reconstruction recovers it.

The four controls bound the change:

- an already-complete trail is reconstructed on the **first** read, with no waiting - so the
  fix cannot quietly add latency to the normal path;
- an unrecoverable lockout still raises with the reconstruction named in the message;
- the bound is measured on the injected clock, and an already-elapsed bound **still reads
  once**, so a caller never skips the evidence entirely;
- `_run_live` reconstructs through the awaiting core, so a regression to a single
  `read_audit_log` + `resync_after_restart` pair restores the race and fails.

The existing `test_restarted_orchestrator_resyncs_the_outage_from_presence_and_audit` keeps
passing unchanged; it calls `_on_safety_estop` synchronously in the test thread, so its
fixture cannot express a row that lands after the read.

## Notes on scope

I did not try to make the remote write synchronous or add a wait inside
`emergency_stop()` - it broadcasts to peers it cannot block on, and turning that into an
acked call would change the safety path's contract for every caller. The reduction in
`resync_after_restart` is also unchanged; it was already correct for the records it was
given.

Closes #2469

## Verification

- 6 live drill runs per side as above; `MUJOCO_GL=egl`, three MuJoCo worlds, real Zenoh
  session, `eclipse-zenoh` present.
- `ruff check` and `ruff format --check` over `strands_robots tests tests_integ`: clean; the
  changed example is clean under `ruff check` too.
- `mypy strands_robots tests tests_integ`: `Success: no issues found in 1362 source files`.
- 146-file blast radius (every test touching `examples/`, the fleet examples, READMEs,
  `changelog.d`, or walking the tree), same selection both sides: branch **4 failed / 7048
  passed / 83 skipped**, base `upstream/main` **4 failed / 7043 passed / 83 skipped**,
  identical failure set, branch-only failures empty, passed delta `+5` = exactly the new
  tests. The 4 shared failures are `rclpy`-related tests that assert an `ImportError` which
  cannot be raised where `rclpy` is installed.


yinsong1986 (approved):
## Summary

Converts the fleet-03 drill's part-2 re-sync from a single point-in-time `read_audit_log` + `resync_after_restart` pair into a bounded, polled wait (`resync_until_lockout_recovered`) that re-reads the audit log until the reconstruction recovers the survivors' lockout. This matches the diagnosis in #2469: a survivor's `remote_estop_engaged` row is written by that peer's own safety handler after the issuing `emergency_stop()` returns, so the old single read raced the evidence. The fix keeps the predicate and the `RuntimeError` text unchanged, moves only the timing, and confines the change to the example, its README paragraph, a changelog fragment, and tests.

## What's good

- The deadline is measured on `time.monotonic` with injectable `clock`/`sleep` seams, per AGENTS.md > "Clocks: a duration is measured, a stamp is recorded" - and a test pins that a wall-clock step cannot move the bound.
- An already-elapsed bound still performs one read, so a caller never skips the evidence entirely; an already-complete trail returns on the first read with no added latency. Both are pinned.
- `verify_audit_integrity(records)` in `main()` stays scoped to the records the run reports (AGENTS.md rule 16), and the awaited `records` now include the survivor rows the table shows.
- Change recorded as a `changelog.d/` fragment rather than an edit to `CHANGELOG.md` (AGENTS.md PR workflow step 3); `Closes #2469` is in the body, not the title.
- The new tests pass on the head commit (13 passed), and the `_run_live` seam test pins against a regression back to the single-read pair.

PR #2475, opened 2026-08-19
fix(dataset_recorder): check the creation API's posture flags instead of reading them by truthiness

cagataycali:
## Problem

`DatasetRecorder.create` takes three flags that select a *posture* rather than scaling a quantity - `use_videos`, `streaming_encoding` and `overwrite` - and `resume` takes the second. All four were read by truthiness. Every non-empty string is truthy, so `"false"` / `"no"` / `"off"` / `"0"` - the spellings a caller reaches for when opting *out* - selected the branch being opted out of, and `None` / `0` / `""` / `[]` took the other branch without ever being a declared spelling of it.

`overwrite` is a confirmation gate in front of a delete, so it is the expensive one. Read as True it reaches `_prepare_create_target`, which `shutil.rmtree`s the target. That function already refuses to clobber a non-empty *non*-dataset directory, so the one thing it removed without asking was a real LeRobotDataset.

Measured on `ba04dd21`, against a dataset recorded from a MuJoCo rollout (1 episode, 120 frames, two per-camera MP4s written by `start_recording` / `run_policy` / `stop_recording`):

| call | verdict | on disk after | reopen |
| --- | --- | --- | --- |
| `overwrite=True` | returned a recorder | 1 episode -> 0 (as asked) | - |
| `overwrite=False` | `FileExistsError` | 1 episode -> 1 (as asked) | len 120 |
| `overwrite="false"` | returned a recorder | 1 -> **0** ep, 120 -> **0** frames, 2 -> **0** MP4s | **404** |
| `overwrite="no"` / `"off"` / `"0"` | returned a recorder | 1 -> **0** episodes | **404** |
| `use_videos=False` | success | schema `dtype="image"` | - |
| `use_videos="false"` | success | schema `dtype="video"` | - |

`use_videos` writes its branch into `meta/info.json` as each camera's `dtype`, fixed for the life of the dataset, and the same unconverted string then reached LeRobot's own boolean parameter - as `streaming_encoding="false"` did on both entry points.

## Why this surface

Three of the four posture-flag surfaces in this module already check on `boolean_flag_error`:

| surface | flags | checked before |
| --- | --- | --- |
| `push_to_hub` | `private` | yes |
| `sync_dataset_to_bucket` | `create`, `private`, `delete` | yes |
| `start_recording` (all three backends) | `overwrite`, `push_to_hub` | yes |
| `DatasetRecorder.create` / `resume` | `use_videos`, `streaming_encoding`, `overwrite` | **no** |

So the facade and the method it forwards to disagreed about which values are usable. That is the disagreement the `fps` guard in the same block exists to prevent - its own comment says the domain is "the shared one every backend's `start_recording` applies to the `fps` it forwards here, so the facade and this method cannot disagree about which rates are usable" - and `dataset_recording_posture_error` names this exact mechanism, `shutil.rmtree` included, for the facade it already guards.

## Change

Check all four flags on `boolean_flag_error`, in the guard block that already validates the schema names, the frame shape and the rate, and ahead of the same two side effects: **before** the lerobot extra is probed, so one caller mistake reports identically on every install, and **before** the on-disk target is touched, so a refusal cannot arrive after the deletion it was refusing.

The domains are inverses - the numeric ones reject `bool` because an `int` subclass would pass as a silent `1`, this one requires the boolean they turn away - so both now hold in the same signature.

## Tests

`tests/test_dataset_recorder_posture_flag_domain.py`, 53 items. Against `ba04dd21`: **36 failed / 17 passed** -> 53 pass.

The 17 that pass on both trees are the controls: `overwrite=True` still replaces the target, `overwrite=False` still refuses with its own `FileExistsError` naming `overwrite=True` and `resume` (not displaced by a domain complaint about a valid boolean), both honorable `use_videos` values still declare the schema they name, an honorable `streaming_encoding` still reaches both datasets as the boolean given, `fps=True` is still refused by the inverse domain, and a call stating all four knobs honorably still builds a dataset.

`TestEveryDeclaredPostureFlagIsChecked` derives the flag set from the live signatures, so a fourth `bool` parameter added later fails before it can be missed, with a planted-source non-vacuity check. `TestTheDomainIsTheSharedOne` parametrizes over `boolean_flag_error` itself rather than a copied spelling list.

## Scope

Two neighbouring surfaces are deliberately out, and the test file records why:

- `DatasetRecorder.__init__(strict=)` selects raise-vs-drop on a malformed frame, so a truthy non-boolean fails toward *raising* - loud and recoverable, not the silent inversion the creation flags are.
- `sync_to_bucket` forwards to `sync_dataset_to_bucket`, which already checks all three of its flags.

## Verification

Measured at `444dde78` against `upstream/main` `d059c6c3` on the 164 test files touching the recorder, the recording facades, `boolean_flag_error` and the docs/changelog guards:

- branch **4 failed / 5749 passed / 30 skipped**, base **40 failed / 5713 passed / 30 skipped**; both collected 5753, and the failed/passed delta is 36 either way - exactly the pre-fix failure count.
- Branch-only failing ids: **none**. Base-only: exactly the 36 from the new file. The 4 shared failures are pre-existing and unrelated (`_safety_publishers_lock`, two `strands_robots.simulation.policy_runner` clock-double cases, and a `draccus` version mismatch in `test_lerobot_e2e`).
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 with no branch-only error and none in the changed files.

## Artifact

One capture script per tree: record an episode from a real MuJoCo rollout, then call `create(root=<that dataset>, overwrite="false")` and reopen what is left. Same script, same rollout; only `strands_robots` differs.

DatasetRecorder posture flag before and after

The two MP4s and the 120-frame parquet *are* the recorded episode. On `main` the call reported nothing and removed them, so the round trip falls through to the Hub and 404s; on this branch the call is refused by name and the dataset reopens at `len=120` with its image key intact.

The branch carries one merge of `upstream/main` to pick up `d059c6c3`. The only overlap was the posture-flag section of `AGENTS.md`, resolved additively - the RTPS gate's paragraph and this rule's bullet both stand, and both pinned test files are still named. Composition re-checked by symbol, and the pre-fix split is unchanged against the new base (36 failed / 17 passed). The numbers above are from the merged tree.


yinsong1986 (approved):
## Summary

Closes a real data-loss bug: `DatasetRecorder.create` read `overwrite` (and `use_videos` / `streaming_encoding`) by truthiness, so `overwrite="false"` / `"no"` / `"off"` / `"0"` reached `_prepare_create_target` as True and `shutil.rmtree`-d the caller's existing dataset while returning a working recorder. The fix routes all four flags (three on `create`, `streaming_encoding` on `resume`) through the shared `boolean_flag_error` domain, in the same guard block as the existing `fps` / schema checks, before the lerobot extra is probed and before the on-disk target is touched - verified in source: the posture guard precedes both `_get_lerobot_dataset_class()` and `_prepare_create_target(...)` in both entry points, and the new test file pins that ordering structurally via AST.

## What's good

- Uses the existing shared domain rather than a new local guard, closing the facade-vs-direct-API disagreement the module's own `fps` guard comment describes; the `start_recording` facades, `push_to_hub` and `sync_dataset_to_bucket` already checked, so this was the one surface left reading by truthiness.
- 53-test regression file passes at head; controls pin that `overwrite=True` / `False` and both honorable `use_videos` values keep their documented behaviour, that the refusal message is the shared wording verbatim, and that `fps=True` is still refused by the inverse numeric domain. `TestEveryDeclaredPostureFlagIsChecked` derives the flag set from the live signatures with a planted-source non-vacuity check.
- Every recording-test call site names its own `root=` under `tmp_path` (AGENTS.md > "A recording test names its own dataset root"), no host paths, ASCII-only additions.
- Scope discipline: `__init__(strict=)` and `sync_to_bucket` deliberately excluded with the reasoning pinned in tests, changelog fragment in `changelog.d/`, docs and AGENTS.md updated in the same PR.

The stricter refusal is a behaviour change for callers passing non-boolean values, but the previous behaviour was silent dataset deletion, so an explicit `ValueError` is exactly the migration path AGENTS.md prescribes ("Raise on fatal errors", posture-flag bullet).

PR #2478, opened 2026-08-19
fix(mesh): report an absent eclipse-zenoh at the level a consumer sees

### LINKED ISSUE #2477: bug(mesh): a missing eclipse-zenoh degrades to 'mesh off' at DEBUG only, so mesh-requiring examples fail as opaque presence timeouts
# Symptom

With `eclipse-zenoh` not installed, the mesh-requiring fleet examples (02, 03) fail as
opaque 15 s presence-discovery timeouts:

```
RuntimeError: timed out waiting for presence discovery of all robots
```

Nothing in the output names the actual cause - the mesh never came up at all, because the
zenoh import failed and the degradation was logged at DEBUG level only.

# Cause

Two sites, both DEBUG:

- `strands_robots/mesh/session.py` (`get_session`): `except ImportError:` ->
  `logger.debug("eclipse-zenoh not installed - mesh disabled")` -> `return None`.
- `strands_robots/mesh/core.py` (`Mesh.start`): `session is None` ->
  `logger.debug("[mesh] %s: zenoh unavailable, mesh off", self.peer_id)` -> `return`.

Default logging config shows neither, so the first observable symptom is whatever
downstream wait expires first - in the examples, a presence-discovery `_wait_for` whose
timeout message says nothing about a missing dependency.

# Constraint on the fix

Quiet-off is deliberate in the library core: the comment beside the port parsing in
`get_session` states the default behaviour is "to keep the mesh quietly off rather than
crash the host robot", and a robot that works fine without the mesh should not start
raising because an optional extra is absent. So the fix is not a raise in
`session.py`/`core.py`. Two compatible options:

1. **Promote the two log lines to WARNING.** A missing optional dependency that silently
   disables a subsystem the caller may be depending on is exactly the "warn on
   surprising degradation" case - same shape as the `STRANDS_ROBOT_MODE` typo rule in
   AGENTS.md ("silent typo'd env vars surprise users hours later").
2. **Raise in the examples that require the mesh.** `init_mesh` in the fleet examples can
   check `mesh.alive` (or that `get_session()` returned a session) and raise with an
   actionable message (`pip install 'strands-robots[mesh]'`), since for those examples
   the mesh is not optional.

Option 1 is one-line-per-site and helps every consumer; option 2 additionally turns the
examples' 15 s timeout into an immediate, named refusal. They are not exclusive.

# Repro

```
pip uninstall -y eclipse-zenoh
STRANDS_MESH_HITL_ACTIONS=none STRANDS_MESH_LOCAL_DEV=1 MUJOCO_GL=egl \
STRANDS_MESH_AUDIT_PSK=demo-psk STRANDS_MESH_OVERRIDE_CODE=demo-override \
python examples/fleet/03_failover_and_degraded_ops.py
# -> ~15 s, then an opaque presence-discovery timeout
```

Split out of #2469 (its "adjacent finding" - the race itself is fixed by #2474).
Part of epic #2179.

Board routing: please add to https://github.com/orgs/strands-labs/projects/2 with
Status=Todo, Priority=Low (filing account lacks org project write).

cagataycali:
## Problem

With `eclipse-zenoh` absent, a mesh-requiring workflow fails as an opaque presence-discovery timeout. Nothing in the output names the absent dependency, because the degradation is reported below the level a default-configured consumer prints.

Measured with `import zenoh` failing, root logger at its default `WARNING`:

| call | result | records naming the cause |
|---|---|---|
| `get_session()` | `None` | 0 |
| `_get_zenoh_session_directly()` | `None` | 0 |
| `Mesh.start()` | `mesh.alive` is `False` | 0 |

The only record a consumer *does* see is an unrelated e-stop advisory that describes fleet-wide e-stop semantics for a mesh that is not running at all.

## Why this is the odd one out

`get_session` reports every other way it can end with no session at `WARNING`, and even a bad `STRANDS_MESH_PORT` -- which still yields a **working** mesh -- warns:

| line | event | level | mesh comes up? |
|---|---|---|---|
| `Invalid STRANDS_MESH_PORT ... falling back to 7447` | recoverable misconfiguration | `WARNING` | yes |
| `Zenoh listener ... unavailable - trying client mode` | recoverable | `DEBUG` | yes (recovers) |
| `Zenoh session open failed (peer fallback)` | total | `WARNING` | no |
| `Zenoh session open failed` | total | `WARNING` | no |
| `eclipse-zenoh not installed` | total | **`DEBUG`** | **no** |

`_get_zenoh_session_directly` has the same shape. `Mesh.start` does too: the sibling ACL refusal a few lines above also leaves the mesh not-started and reports at `ERROR`.

So the most total outcome was the quietest one. The comment beside that port fallback is also where the quiet-off design is stated -- "falls back to the default and **warns once** - never raises (the default behaviour is to keep the mesh quietly off rather than crash the host robot)" -- i.e. quiet-off means *does not raise*, not *does not warn*.

## Change

`_report_zenoh_missing()` reports the absent dependency at `WARNING`, once per process, naming the extra that supplies it. Both openers re-attempt the import on every call (nothing is cached when it fails), so a fleet of N robots in one process would otherwise emit N copies of one static fact. The guard is a module-level set mutated via `.add` and never rebound -- the same shape as `_software_render_warned` in the MuJoCo backend, whose comment records why (`global` rebinding reads as unused to static analysis) -- and it is touched only under the existing `_SESSION_LOCK`, so it needs no new synchronisation. `Mesh.start`'s no-transport branch reports at `WARNING` and names the peer.

Return behaviour is untouched: the openers still return `None`, `Mesh.start` is still a clean no-op, and `init_mesh` still returns a `Mesh` rather than raising. A robot that does not need the mesh runs unchanged.

`docs/troubleshooting.md` claimed `init_mesh` **raises** when `eclipse-zenoh` is missing. It does not -- it returns a `Mesh` whose `alive` is `False`, which is what `docs/mesh.md` already documents ("Mesh failures are non-fatal"). The two pages contradicted each other, and the one a reader consults when the mesh is broken sent them looking for an exception that never comes. That row now names the observable a caller actually gets.

Out of scope: raising in the fleet examples that treat the mesh as mandatory. After this change the cause **is** named, but the downstream wait still expires -- turning it into an immediate refusal is a behaviour change in those examples and a separate contract. `factory.get_transport()` returning `None` is a *connect* failure (certs, broker), not an absent dependency, and is left alone.

## Console

Same script, same absent extra, only `strands_robots` differs. The capture drives the real `init_mesh` twice and then the presence wait, with `import zenoh` failing:

operator console with eclipse-zenoh absent

Both runs still end in the same timeout -- the mesh is optional, so it stays off rather than raising. What changed is that the cause is attributable: 0 records naming it before, 3 after (one cause + one per affected peer), and the remedy is runnable. `mesh.alive` is `False` in both.

## Tests

`tests/mesh/test_zenoh_absent_is_reported.py`, 10 items. Against `upstream/main`: **6 failed / 4 passed**; after: 10 passed.

Fails pre-fix:
- both openers name the dependency at `WARNING` (parametrized -- the private one is what a `BridgeTransport`'s zenoh leg uses and carried the same `DEBUG` report)
- `Mesh.start` names the peer whose mesh stayed off
- the offered extra is declared in `pyproject.toml` **and** supplies `eclipse-zenoh`, so following the message works (the remedy is parsed back out of the record, not restated)
- four session opens report it exactly once
- root cause: every `_report_zenoh_missing()` call sits inside an `ImportError` handler, so it cannot come to blame a missing dependency for some other failure

Passes on both trees -- the controls, which fail for a fix that raises instead of reporting:
- both openers still return `None` and leave `_SESSION` unset
- `Mesh.start` is still a clean no-op
- `init_mesh` still returns a `Mesh`

## Verification

Measured at `fdc407c2` against `upstream/main` `d059c6c3`.

- Blast radius: all of `tests/mesh` plus every test touching the changed symbols and the repo-wide doc/changelog/docstring guards -- 220 files, run sequentially. Branch **21 failed / 4959 passed / 5 skipped / 24 errors**; base with the new test file copied in **27 / 4953 / 5 / 24**. Both collected 5009; failed delta = passed delta = 6 = the pre-fix failure count; branch-only failures **empty**; base-only exactly the 6 above. The 45 shared failures are all `tests/mesh/test_iot_*` needing `awsiotsdk`, which is declared and installed in CI.
- `mypy strands_robots tests tests_integ`: 24 on both trees, branch-only empty, none in the changed files.

Closes #2477

### DIFF
--- a/changelog.d/2478-zenoh-absent-report-level.md
+++ b/changelog.d/2478-zenoh-absent-report-level.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- **mesh**: report an absent `eclipse-zenoh` at `WARNING` instead of `DEBUG`, so a
+  default-configured consumer learns why the mesh never came up. Both session-open
+  paths and `Mesh.start` left the degradation below the level anything prints, so the
+  first observable symptom was whichever downstream wait expired first. The
+  dependency is named once per process with the extra that supplies it; the mesh
+  still stays off rather than raising.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -72,7 +72,7 @@ description: Error → fix table for the most common gotchas across install, sim
 |---------|--------------|-----|
 | `mesh.peers` empty | Other peer not running | Wait ~1s; verify `mesh.alive == True` on both |
 | Port already bound | Another zenoh process | Mesh auto falls back to client mode; or set `STRANDS_MESH_PORT` |
-| `init_mesh` raises | `eclipse-zenoh` missing | `uv pip install "strands-robots[mesh]"` |
+| `mesh.alive` is `False`, `mesh.peers` stays empty | `eclipse-zenoh` missing (logged at WARNING: "eclipse-zenoh is not installed") | `uv pip install "strands-robots[mesh]"` |
 | Want mesh off | - | `STRANDS_MESH=false` or `Robot(..., mesh=False)` |
 
 ## Agent integration
--- a/strands_robots/mesh/core.py
+++ b/strands_robots/mesh/core.py
@@ -682,7 +682,17 @@ class Mesh(SensorLoopsMixin):
                 # state.
                 _acl_config._clear_thread_snapshot()
             if session is None:
-                logger.debug("[mesh] %s: zenoh unavailable, mesh off", self.peer_id)
+                # The sibling refusal above leaves the mesh not-started and says
+                # so at ERROR.  Arriving here leaves it not-started too, so the
+                # caller learns that ``mesh.alive`` is False from the same level
+                # rather than from whichever downstream wait expires first.  The
+                # cause is reported by the session/transport layer that returned
+                # None; this names the peer it applies to.
+                logger.warning(
+                    "[mesh] %s: no mesh transport - mesh off (mesh.alive is False, so this "
+                    "peer publishes no presence and discovers no peers)",
+                    self.peer_id,
+                )
                 return
 
             self._has_session_ref = True
--- a/strands_robots/mesh/session.py
+++ b/strands_robots/mesh/session.py
@@ -60,6 +60,15 @@ _SESSION: Any | None = None  # zenoh.Session when open, else None
 _SESSION_LOCK = threading.Lock()
 _SESSION_REFS: int = 0
 
+# One-shot guard so an absent ``eclipse-zenoh`` is reported at most once per
+# process.  Both session-open paths re-attempt the import on every call (nothing
+# is cached when it fails), so a fleet of N robots in one process would otherwise
+# emit N copies of an identical, static fact.  A set (mutated via .add, never
+# reassigned) avoids a `global` rebind so static analysis sees it as used -- the
+# same shape as ``_software_render_warned`` in the MuJoCo backend.  Mutated only
+# under ``_SESSION_LOCK``.
+_zenoh_missing_warned: set[str] = set()
+
 
 # Constants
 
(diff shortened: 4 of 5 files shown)


yinsong1986 (approved):
## Summary

Raises the report level for an absent `eclipse-zenoh` from `DEBUG` to `WARNING` in both session openers (`get_session` and `_get_zenoh_session_directly`), via a shared `_report_zenoh_missing()` helper with a once-per-process guard, and upgrades `Mesh.start`'s no-transport branch to `WARNING` naming the affected peer. The optional-dependency contract is unchanged: openers still return `None`, `Mesh.start` is still a no-op, `init_mesh` still returns a `Mesh` with `alive` False. The stale `docs/troubleshooting.md` row claiming `init_mesh` raises is corrected to the actual observable.

## What's good

- The once-guard is mutated only under the existing `_SESSION_LOCK` at both call sites (verified against the head tree), and the set-instead-of-global-rebind shape resolves the prior CodeQL unused-global finding.
- The new test file pins the level, the once-per-process guard, that the offered extra really declares `eclipse-zenoh` in `pyproject.toml`, and — via controls — that reporting did not become raising. All 10 pass at this head, along with the neighboring mesh suites (`test_mesh*`, `test_session*`, transport-routing pin: 224 passed).
- No other code references the old `"eclipse-zenoh not installed"` / `"zenoh unavailable, mesh off"` strings, so nothing asserts on the removed messages.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; messages are plain ASCII.

PR #2479, opened 2026-08-19
ci(test-lint): size the suite's job bound so a step spending its whole bound cannot reap it

### LINKED ISSUE #2457: ci(test-lint): the suite's healthy worst case is now 38.8 min against a 45-minute bound sized when the whole job was 26.5
## What I measured

While sizing the apt bounds for #2442 I sampled `call-test-lint / Test and Lint` over the 300 most recent workflow runs (31 **successful** jobs, window 2026-08-18T17:12Z to 2026-08-19T04:00Z). The job's own bound is `timeout-minutes: 45`.

| | n | min | p50 | p90 | max |
|---|---|---|---|---|---|
| `Run tests` step | 31 | 1311s (21.9m) | **1648s (27.5m)** | 1689s (28.2m) | 1723s (28.7m) |
| whole job | 31 | 1726s (28.8m) | **1954s (32.6m)** | 2123s (35.4m) | **2326s (38.8m)** |

The bound is 45, so the observed maximum is at **86%** of it and the median at 72%.

## Why this is worth recording

`tests/test_workflow_jobs_are_bounded.py` sized that bound, and its docstring states the band it was sized from:

> ```
> workflow / job                    n    p50 min   max min   bound
> Test and Lint / test-lint        20      25.2      26.5      45
> ```
> The suite's band is tight (19-27 min, and zero completed runs over the sampled 300 exceeded 45), so 45 is ~1.7x the observed ceiling

Against today's numbers that reads differently: `Run tests` **alone** now has a p50 of 27.5 min, which is above the old *whole-job* maximum of 26.5, and the whole job reaches 38.8. The margin the docstring describes as ~1.7x is now ~1.16x at the maximum. The assertion itself still passes (`_SUITE_FLOOR_MINUTES <= 45 <= _CEILING_MINUTES`), so nothing fails - the drift is in the reasoning the number rests on, not in the guard.

One concrete consequence, already visible: the two runs in #2442 reached the 45-minute reap only because ~45 min was spent in a *wedged apt step*. With the suite itself at 38.8 min worst case, a much smaller stall than 45 min is now enough to exhaust the budget - which is part of why #2456 has to size that step's bound against the job's remaining headroom (~13 min) rather than purely against apt's own distribution.

## Why this is filed rather than patched

The remedy is a choice, and the two answers have different costs:

- **Raise the bound** (e.g. 60, still inside the guard's `_CEILING_MINUTES`). Cheap, restores headroom, and buys nothing else - a genuinely hung job then holds a runner 15 minutes longer than today, which is the cost #2239 introduced the bound to cap.
- **Bring the suite back down.** `Run tests` is ~85% of the job and grew from a 26.5-minute whole-job maximum to a 28.7-minute step; whether that is test count, a slow subset, or lost parallelism is unmeasured here and worth knowing before choosing a number.
- **Leave it and re-sample.** Defensible if the growth is a level shift rather than a trend - my window is ~11 hours, which is too short to tell those apart.

Either of the first two also wants the docstring's measured table refreshed in the same change, since that table is what the next person will size against.

## Notes

- Sampling method: `GET /actions/runs?per_page=100` x3, then per-run `/jobs`, keeping jobs named `... Test and Lint` with `conclusion == success`, and differencing `started_at` / `completed_at` per step. The same script produced the apt figures in #2456.
- The 300-run window only reaches back ~11 hours, so a longer sample is the first thing to do if the trend question matters. n=31 is the successful-job subset of it.
- Not a duplicate of #2239 (a post-verdict hang, fixed by introducing the bound) and not #2442 (an unbounded apt step, fixed by #2456). This is the healthy-path duration approaching the bound both of those rely on.

cagataycali:
Closes #2457

## The defect

`Test and Lint` is the one required check every merge waits on. Its job bound (45 min) and the bound on its own apt step (12 min, #2456) were sized independently, and had stopped being compatible with each other.

Measured over **663 successful jobs in the 10 days to 2026-08-19**:

| | n | min | p50 | p90 | max |
|---|---|---|---|---|---|
| `Run tests` step | 663 | 13.8 | 22.6 | 27.2 | 29.1 |
| whole job | 663 | 17.3 | 27.1 | 32.0 | **44.4** |

The steps carrying no bound of their own (checkout, setup-python, install, lint) cost 4.4 min. So a run in which apt spends its **entire legal allowance** costs `29.1 + 12 + 4.4 = 45.5 min` and is reaped at 45 having done nothing wrong.

That is the worst possible presentation. A job-level reap renders as a rollup `FAILURE` naming no step -- which is precisely the diagnosis the apt step bound was added to provide -- and it is indistinguishable from the branch's own tests failing.

**The suite is not what put those runs near the bound.** The worst four successful jobs show the composition directly:

| job | apt step | `Run tests` |
|---|---|---|
| 44.4 min | **12.4** | 27.6 |
| 44.0 min | **11.9** | 27.7 |
| 42.0 min | **10.5** | 27.1 |
| 39.6 min | **9.1** | 26.4 |

`Run tests` never exceeded 29.1 min across all 663.

## What this changes

1. **Job bound 45 -> 60.** Sized against the arithmetic above rather than a multiple of an observed ceiling, because that ceiling is moving (below).
2. **A pin for the budget, not prose about it.** `tests/test_workflow_jobs_are_bounded.py` now asserts the job bound covers `sum(step bounds) + measured suite ceiling + unbounded overhead`. The two numbers live in different blocks, are edited by different changes for different reasons, and neither edit has any occasion to read the other -- which is exactly how they drifted apart. Verified to fail on pre-fix code:
   ```
   AssertionError: the suite bound is 45 min but a run in which every bounded step
   spends its whole allowance costs 47 min (steps 12 + suite 30 + overhead 5);
   a job-level reap names no step, so raise the job bound or lower a step bound
   ```
3. **Refreshed the stale measured table.** The docstring still carried the 20-run sample (`p50 25.2, max 26.5`) the old bound was sized from; that table is what the next person sizes against.
4. **Suite floor 30 -> 46.** The old floor sat *below* the observed maximum of 44.4, so it would have admitted a bound that reaps healthy runs.

## Why 60, and why it is the last raise here

60 is also `_CEILING_MINUTES` in the guard -- deliberately. This resample answers the question #2457 left open ("my window is ~11 hours, which is too short to tell a level shift from a trend"): it is a **trend**, and a clean one. Per-day p50 rose *monotonically on all ten days*, 23.0 -> 32.7 min, a +0.95 min/day fit at R^2 0.93, while `tests/` gained 3054 test functions (13770 -> 16824, +22%) across 297 commits.

So the next raise cannot be spent on this bound: it would have to raise the guard's ceiling too, and that edit is the forcing function to argue suite runtime instead. The apt step's own 12-minute bound is untouched -- `tests/test_apt_steps_are_bounded.py` passes unchanged.

## Verification

- The two new/retightened assertions fail at 45 and pass at 60 (both directions run).
- All **264 tests** across the ten `tests/*.py` files that read `test-lint.yml` pass, including `test_apt_steps_are_bounded.py`.
- `ruff check` + `ruff format --check` clean; workflow parses under `yaml.safe_load` to `timeout-minutes: 60`; non-ASCII sweep clean.
- Intake re-checked before push: `unique-claim`, 7 open PRs compared. Changelog fragment gate clean.

## Note for the reviewer

This edits `.github/workflows/**`, so `mergeStateStatus` read with the Actions `GITHUB_TOKEN` will report `BLOCKED` regardless of state (AGENTS.md step 8) -- read it with `PAT_TOKEN`. I am the pusher, so this needs an approver who is not me.

### Round 1
Initial submission.

### DIFF
--- a/.github/workflows/test-lint.yml
+++ b/.github/workflows/test-lint.yml
@@ -11,12 +11,23 @@ jobs:
   test-lint:
     name: Test and Lint
     runs-on: ubuntu-latest
-    # The suite is tightly distributed - 19-27 min over the last 20 successful
-    # runs, max 26.5 - so 45 is ~1.7x the observed ceiling and costs a healthy
-    # run nothing. Without it this job inherits the 6-hour default, which is how
-    # a test that had already printed its verdict at 18 min went on to hold the
+    # Sized so that every bounded step below can spend its ENTIRE allowance and
+    # the suite still finishes inside this bound - the two were sized
+    # independently and stopped being compatible. Over 663 successful runs in
+    # the 10 days to 2026-08-19: `Run tests` p50 22.6 / p90 27.2 / max 29.1 min,
+    # and the unbounded remainder (checkout, setup-python, install, lint) 4.4
+    # min. Against the apt step's own 12-minute bound that is 29.1 + 12 + 4.4 =
+    # 45.5 min of legal spend, so 45 could be reaped with nothing wrong and no
+    # step named - the exact outcome that step bound exists to prevent (#2456).
+    # 60 leaves ~14.5 min of margin. It is also `_CEILING_MINUTES` in
+    # tests/test_workflow_jobs_are_bounded.py, deliberately: the suite grew
+    # +0.95 min/day over that window (p50 23.0 -> 32.7, R^2 0.93, driven by
+    # +3054 test functions), so the next raise cannot be spent here and has to
+    # be argued as suite runtime instead (#2457).
+    # Without any bound this job inherits the 6-hour default, which is how a
+    # test that had already printed its verdict at 18 min went on to hold the
     # runner for another 82 (#2239).
-    timeout-minutes: 45
+    timeout-minutes: 60
     permissions:
       contents: read
     steps:
--- a/changelog.d/2457-suite-bound-clears-its-step-bounds.md
+++ b/changelog.d/2457-suite-bound-clears-its-step-bounds.md
@@ -1,0 +1,28 @@
+### Fixed: the suite's job bound now covers every step bound inside it
+
+`Test and Lint` is the one required check, and its 45-minute job bound had
+stopped being compatible with the 12-minute bound on its own apt step. The two
+were sized independently, months apart, and neither edit had occasion to read
+the other.
+
+Measured over 663 successful jobs in the 10 days to 2026-08-19: `Run tests`
+costs at most 29.1 min, the steps carrying no bound of their own cost 4.4 min,
+so a run in which apt spends its entire 12-minute allowance costs 45.5 min. That
+run was reaped at 45 having done nothing wrong -- and a job-level reap renders as
+a rollup `FAILURE` naming no step, which is exactly the diagnosis the step bound
+was added to provide. The worst four successful jobs in the window were
+44.4/44.0/42.0/39.6 min, and in each the excess is apt (12.4/11.9/10.5/9.1),
+never the suite.
+
+The job bound is now 60. `tests/test_workflow_jobs_are_bounded.py` asserts the
+budget rather than describing it: the job bound must cover the sum of its step
+bounds plus the measured suite ceiling and unbounded overhead, so a future edit
+to either number that breaks the relationship fails instead of silently
+narrowing the margin. The resampled band also replaces the stale 20-run table
+the old bound was sized from, and the suite floor rises from 30 to 46, which had
+sat below the observed maximum.
+
+60 is also the guard's own ceiling, deliberately. The suite grew +0.95 min/day
+across that window (p50 23.0 -> 32.7, R^2 0.93) as `tests/` gained 3054 test
+functions, so this is the last raise available here and the next one has to be
+argued as suite runtime instead.
--- a/tests/test_workflow_jobs_are_bounded.py
+++ b/tests/test_workflow_jobs_are_bounded.py
@@ -32,19 +32,40 @@ reviewer into waiting.  A job-level bound is the only mechanism that sits
 outside the interpreter, which is why it fixes the class rather than the one
 test that exposed it.
 
-The bounds themselves are sized from observed durations over the last 300
-workflow runs, not guessed::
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

Raises the `Test and Lint` job bound from 45 to 60 minutes so that the apt step's own 12-minute bound (#2456) can spend its entire allowance without pushing a healthy run into a job-level reap -- the failure mode that renders as a rollup `FAILURE` naming no step, which is exactly what the step bound was added to prevent. The change is backed by a 663-run resample, and the key move is converting the prose budget into an assertion: `test_the_suite_bound_survives_every_step_bound_spending_its_allowance` reads the step-level bounds out of `test-lint.yml` and requires the job bound to cover `sum(step bounds) + measured suite ceiling + unbounded overhead`, so the two independently-edited numbers can no longer drift apart silently. The suite floor also rises from 30 to 46, correcting a floor that sat below the observed maximum, and the bound deliberately lands on `_CEILING_MINUTES` so the next raise forces an argument about suite runtime instead.

## What's good

- Both assertion directions verified locally at the head SHA: the retightened floor and the new budget assertion fail at 45 (`... costs 47 min (steps 12 + suite 30 + overhead 5)`) and pass at 60; all 83 tests across `test_workflow_jobs_are_bounded.py` + `test_apt_steps_are_bounded.py` pass.
- The `_STEP_TIMEOUT_MINUTES` regex is correctly anchored at the 8-space step indent, so the job-level `timeout-minutes: 60` cannot self-count into the budget, and the `assert step_bounds` guard catches both the apt bound being removed and the indent contract changing.
- Workflow parses under `yaml.safe_load` to `timeout-minutes: 60`; non-ASCII sweep clean on all three files; changelog fragment follows the `changelog.d` convention (single `###` entry, issue-numbered filename is explicitly allowed by `changelog.d/README.md`).
- Scope discipline: the apt step's 12-minute bound is untouched, and the stale 20-run table the old bound was sized from is refreshed in the same place the next person will size against.

PR #2481, opened 2026-08-19
docs(simulation): name every exception class the on_frame posture exempts

cagataycali:
## Problem

Five surfaces document what happens when the `on_frame` hook raises, and each names one exempt exception class:

> A non-`CooperativeStop` hook exception is logged at WARN and never aborts the eval.

The rollout loops exempt **two**. All three `on_frame` call sites re-raise `RecordingFrameError` as well, deliberately - a lost dataset frame is data loss rather than telemetry, so tolerating it would write a short, re-timestamped episode under a successful rollout. So for that class the documented posture is wrong in every particular: it is not logged, it is not tolerated, and it aborts on the first occurrence.

That is the failure a caller following the hook's own documented use is most likely to meet. `eval_policy` says of it: *"Use it to record frames or stream telemetry synchronously on the eval thread"* - and then tells the reader a failed dataset write cannot abort the eval.

Measured on a one-robot sim, driving both routes with a hook that raises on every step:

| surface | hook raises | hook calls | WARN logs | outcome |
| `eval_policy` | `ValueError` | 12 (every step) | 12 | `status=success` - as documented |
| `eval_policy` | `RecordingFrameError` | **1** | **0** | **raised `RecordingFrameError`** out of the method |
| `PolicyRunner.run`, `max_onframe_failures=5` | `ValueError` | 5 | 5 | `status=error`, "failed 5 times in a row" - as documented |
| `PolicyRunner.run`, `max_onframe_failures=5` | `RecordingFrameError` | **1** | **0** | `status=error` on the first failure; the tolerance is not consulted |

The prose in `docs/simulation/overview.md` already gets this right - *"a `RecordingFrameError` from it is data loss and propagates"* - and the module's own note on `_MAX_CONSECUTIVE_ONFRAME_FAILURES` explains why. Only the docstrings a caller reads at the call site were narrower than the code.

## Change

Documentation only; no behaviour change. Each surface that describes the posture now names both exempt classes and says what the second one does:

- `SimEngine.run_policy` (`max_onframe_failures`) and `SimEngine.eval_policy`
- `SimEngine.evaluate_benchmark` (`on_frame`)
- `PolicyRunner.run` (`max_onframe_failures`) and `PolicyRunner.evaluate` (`on_frame`)
- the inline note in `evaluate`'s hook wrapper, which restated the same claim

`max_onframe_failures` reads as sizing the exemption, so its wording is explicit that the two classes are exempt *from the count* rather than tolerated by it, and that a lost frame aborts on the first occurrence whatever it is set to.

## Tests

`tests/simulation/test_recording_frame_loss_is_not_tolerated.py` already owns this contract behaviourally. Extended with the documentation half:

- `TestTheDocumentedPostureNamesEveryExemptClass` derives the exempt set **from the handlers themselves** - a handler whose whole body is a bare `raise` re-raises instead of tolerating - and requires every doc block that quantifies over the hook's exceptions to name each one. A third exempt class therefore cannot be added without these surfaces being updated with it. It carries a reach check (the five owners must all be graded, so a reformat fails loudly rather than reporting clean) and a planted-wording check (the replaced sentence must still be rejected).
- `TestAnExplicitToleranceDoesNotResizeTheExemption` pins the knob's own claim: against the same `max_onframe_failures=5`, a telemetry failure is called 5 times and a lost frame once.
- The innermost-`try` discovery the existing handler guard used is now a shared helper, so the handler guard and the docstring guard cannot disagree about which handlers define the posture.

Pre-fix split against `main`: **1 failed / 12 passed**, the failure naming all five surfaces and their exact wording. The 12 that pass on both trees are the evidence nothing behavioural moved - including both tolerance rows above.

## Verification

Measured at `db63300e` against `upstream/main` (`d059c6c3`).

- `ruff check` / `ruff format --check` clean on the CI scope; `mypy strands_robots tests tests_integ` 24 == 24, none in the changed files.
- 93-file selection (all of `tests/simulation`, every test naming the hook or either eval route, plus the repo-wide docstring/changelog guards): see the comment below for the final counts.

on_frame documented posture vs measured behaviour

Same package code in both panels - only the docstrings differ - and the measured rows are byte-identical across the two trees, which is asserted in the figure's own build step.


cagataycali:
Local gate at `db63300e`, 93-file selection (all of `tests/simulation`, every test naming the hook or either eval route, plus the repo-wide docstring / changelog / string-hygiene guards), run against this branch and against a worktree of `upstream/main` with the test file copied in:

| | passed | failed | skipped | collected |
| this branch | 12459 | **0** | 207 | 12666 |
| `upstream/main` | 12458 | 1 | 207 | 12666 |

Same collected total on both, and the passed delta equals the failed delta equals the one pre-fix failure. The single base failure is the headline guard, naming all five surfaces:

```
FAILED tests/simulation/test_recording_frame_loss_is_not_tolerated.py::
  TestTheDocumentedPostureNamesEveryExemptClass::test_every_graded_surface_names_every_exempt_class
  - base.run_policy describes the on_frame posture without naming
    ['CooperativeStop', 'RecordingFrameError']: ...
```

Zero branch-only failures, and no pre-existing failures anywhere in the selection. `mypy strands_robots tests tests_integ` is 24 on both trees with none in the changed files, measured against a pristine `upstream/main` worktree; `ruff check` and `ruff format --check` are clean on the CI scope.

yinsong1986 (approved):
## Summary

Documentation-only correction: five surfaces (`SimEngine.run_policy` / `eval_policy` / `evaluate_benchmark`, `PolicyRunner.run` / `evaluate`, plus the inline note in `evaluate`'s hook wrapper) claimed that any non-`CooperativeStop` exception from the `on_frame` hook is logged at WARN and tolerated, while all three call sites also re-raise `RecordingFrameError` on the first occurrence. The docstrings now name both exempt classes and state that a lost dataset frame aborts immediately regardless of `max_onframe_failures`. Verified locally at `db63300e`: the two package files are AST-identical to the merge base after stripping docstrings (genuinely no behaviour change), the documented posture matches the handlers (`except RecordingFrameError: raise` at all three `on_frame` guards), and the extended test file passes 13/13.

## What's good

- The doc guard derives the exempt set from the handler ASTs rather than a hard-coded list, so a third exempt class fails the docstring check the moment it is added - and the innermost-`try` discovery is now a shared helper, so the handler guard and the docstring guard cannot disagree.
- The guard carries its own reach check (>= 5 graded blocks, named owners) and a planted-wording non-vacuity check, so a reformat fails loudly instead of reporting clean.
- `TestAnExplicitToleranceDoesNotResizeTheExemption` pins the knob's corrected claim behaviourally (5 calls for telemetry, 1 for a lost frame against the same tolerance), without touching the shared dataset cache.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; no emojis, no host paths, no orphan combining marks in the diff.

PR #1746, opened 2026-07-29
fix(sim): refuse a dataset recording at a rate its frames are not captured at

cagataycali:
## Problem

`start_recording(fps=...)` fixes the frame rate written into the LeRobotDataset metadata, and LeRobot derives every timestamp from it **positionally** (`timestamp = frame_index / fps`). The dataset recorder is driven once per control step and **never decimates**, so the rate frames are really captured at is the rollout's `control_frequency`. A differing `fps` therefore cannot be honored, only mislabelled — and nothing compared the two.

The two library defaults are exactly such a pair (`fps=30` against `control_frequency=50.0`), which is what the documented record-then-rollout sequence in `docs/recording.md` used. Measured on a 40-step position-servo rollout in MuJoCo:

| | `fps` | `control_frequency` | true capture | timestamped | declared episode | actual episode |
|---|---|---|---|---|---|---|
| library defaults | 30 | 50.0 | 0.0200 s/frame | 0.0333 s/frame | **1.30 s** | 0.78 s |
| aligned | 50 | 50.0 | 0.0200 s/frame | 0.0200 s/frame | 0.78 s | 0.78 s |

`start_recording`, the rollout and `stop_recording` all returned `status="success"`, with no log line.

It is not a cosmetic label. That per-frame interval is the control period a policy trains on, and `replay_episode` derives its per-frame physics budget from the dataset rate — on its own stated invariant, in `policy_runner.py`:

> `# dt. The recorded control frequency IS the dataset fps, so derive the`
> `# physics substeps from it (same convention as run() and evaluate()).`

So a mislabelled episode also replays at the wrong speed. Record → replay round-tripped to **0.0000 rad** at matching rates and **0.0315 rad** at the two defaults.

Two further signs this was a known-but-unenforced rule rather than a design choice:

- `start_recording`'s own docstring already *advised* it — "Set it to the `run_policy` `control_frequency` so recorded timestamps match the cadence frames were captured at" — without enforcing it.
- Every existing test that exercises `replay_episode` had already hand-aligned the two rates (`fps=30` with `control_frequency=30`), while the tests that do not replay left them mismatched.

## Change

One shared guard, `SimEngine._validate_recording_rate`, comparing the open recording's rate against the rollout's `control_frequency`. Called from all five rollout entry points: `run_policy`, `eval_policy`, `evaluate_benchmark`, `start_policy`, `run_multi_policy`.

[...]

PR #2482, opened 2026-08-19
fix(examples): record at the rate the rollout captures, not a rate the recorder refuses

cagataycali:
## Problem

Three shipped examples cannot record a dataset. Each declares
`start_recording(fps=30)` and then runs a rollout that captures at 50 Hz, and the
recorder writes one frame per control step with **no decimation**, so it refuses
the mismatch (#1746) rather than writing frames at a distorted timestamp rate.
The episode lands with zero frames, and every one of the three then reports
success or fails somewhere unrelated:

| example | what it printed | what was on disk |
|---|---|---|
| `examples/03_record_dataset.py` | `stop_recording captured no frames - dataset would be empty (0 frames)`, exit 1 - against a docstring promising `Expected output: "Episode saved ... 100 frames, 1 episode(s)"` | `total_frames: 0` |
| `examples/07_post_tune_any_policy.py` | `Recorded LeRobotDataset -> /tmp/...`, then two steps later `RepositoryNotFoundError: 404 ... huggingface.co/api/datasets/local/refs` | `total_episodes: 0, total_frames: 0`, 0 parquet files |
| `examples/locomotion/vla_g1_workflow.py` | `Episode 1/1 recorded.` then `Dataset saved -> ...` | 0 frames, per episode |

`07` is the worst diagnostic: recording is refused, nothing checks the result, and
the failure surfaces three steps later as a Hub 404 for a local dataset, naming
neither the rate nor the recording. `03` at least fails loudly, because it already
checks `stop_recording`; it just never checks `run_policy`, which is where the
refusal actually happens.

The library is right at every step here. Both refusals are precise and name the
remedy - `run_policy` says "the active recording declares 30 fps but this rollout
captures at control_frequency=50 Hz ... Align the two rates", and
`stop_recording` says "captured no frames - dataset would be empty". The examples
discard both.

## Why they drift

`#1746` ("refuse a dataset recording at a rate its frames are not captured at")
landed 2026-07-29. `examples/03_record_dataset.py` was last touched 2026-06-13
and `examples/07_post_tune_any_policy.py` on 2026-07-02, so neither was ever
brought along. The rule is already known in-tree: notebook 05 pairs
`fps=30` with `control_frequency=30.0` and carries a comment explaining exactly
why, and `docs/training/overview.md` pairs `fps=50` with the 50 Hz default. Three
examples were simply left on the wrong side of it.

## Change

Each rollout now captures at the rate its recording declares, and each site
checks the result so the next mismatch is reported where it happens:

- `03_record_dataset.py`, `07_post_tune_any_policy.py`: `control_frequency=30.0`
  to match `fps=30`, plus a `run_policy` result check. `07` also checks
  `stop_recording`, since an empty dataset there is what became the Hub 404.
- `locomotion/vla_g1_workflow.py`: the recording moves to `fps=50` rather than the
  rollout slowing to 30 Hz, because that file's own comment fixes 50 Hz as the
  WBC controller's rate ("dt=0.005 x decimation 4"), and `docs/policies/wbc.md`
  uses 50.0 throughout - dropping the control rate would be a gait change, not a
  recording fix. This is the second remedy `run_policy`'s own message offers
  ("or record at the rollout's rate"). Both branches of that file were affected:
  the WBC branch states `control_frequency: 50.0` explicitly, the `MockPolicy`
  branch takes the same 50 Hz default.
- `07`'s `# 4. DEPLOY - load the freshly-trained checkpoint back as a Policy and
  run it.` now runs it. The loaded `policy` was read only inside a `print`, so the
  step announced a rollout it never performed.

## Verification

Each example was executed before and after, with `MUJOCO_GL=egl`:

| example | before | after |
|---|---|---|
| `03_record_dataset.py` | 0 frames, exit 1 | `Episode saved ... 100 frames, 1 episode(s)`; on disk `total_episodes: 1, total_frames: 100, fps: 30` - exactly the docstring's promise |
| `07_post_tune_any_policy.py` | Hub 404 at the train step | `Recorded ... -> ` (60 frames, fps 30) then `train status: success`, `exported artifact`, `loaded trained policy: LerobotLocalPolicy`, **`deployed: 30 steps in 1.277s`**, `Loop closed: record -> train -> export -> load -> run` |
| `locomotion/vla_g1_workflow.py` | `Episode 1/1 recorded.` over 0 frames | `Episode 1/1 recorded.` over `total_episodes: 1, total_frames: 12, fps: 50` |

## Test

`tests/test_examples_recording_rate_matches_rollout.py` compares every example's
declared `fps` against every rate its rollout can capture at - a stated
`control_frequency`, the signature default when omitted, or a `**kwargs` dict
resolved through its assignments, which is what the G1 workflow needs.

The default is read from `run_policy`'s own signature rather than restated, so a
change to that default cannot leave the test asserting a stale number. Values it
cannot resolve statically degrade to "not graded" rather than to a failure, and an
example declaring several different rates is left ungraded rather than guessed at.

Against pristine `main`: **1 failed, 7 passed**, the failure naming all three
files and the reason (`records at 30 fps but its rollout captures at 50 Hz - the
recorder refuses the mismatch and the episode lands with zero frames`). On this
branch: **8 passed**.

The 7 that pass either way are what bound it: a `_MINIMUM_GRADED_EXAMPLES`
premise so a refactor that stops reaching `examples/` fails loudly instead of
reporting a clean sweep over nothing, and six planted cases (omitted rate,
matching default, wrong explicit rate, right explicit rate, wrong splatted dict,
right splatted dict) so a clean result means the examples agree rather than that
the grader is blind.

## Gate

`ruff check` and `ruff format --check` clean over 1367 files;
`assemble_changelog.py --check` OK; `mypy strands_robots tests tests_integ` 19
errors, all pre-existing in `examples/isaac_gs/` and byte-identical to a pristine
`main` worktree, so branch-only is empty. Tests: the 430 files that read the
examples tree, pin the recording-rate contract, or act as repo-wide source/doc
guards, run on this branch and on pristine `main` and compared by failing ID.

### DIFF
--- a/changelog.d/2482-examples-recording-rate.md
+++ b/changelog.d/2482-examples-recording-rate.md
@@ -1,0 +1,13 @@
+### Fixed: three examples record at the rate their own rollout captures
+
+`examples/03_record_dataset.py`, `examples/07_post_tune_any_policy.py` and
+`examples/locomotion/vla_g1_workflow.py` each declared `start_recording(fps=30)`
+while their rollout captured at 50 Hz, so the recorder refused every frame and
+the episode landed empty: `03` exited reporting `0 frames` against a docstring
+promising `100 frames`, `07` printed `Recorded LeRobotDataset ->` and then died
+two steps later inside lerobot with an unrelated-looking Hub 404 for the empty
+local dataset, and the G1 workflow printed `Episode N/N recorded.` per episode
+for a dataset with nothing in it. Each rollout now captures at the rate its
+recording declares, and each checks the `run_policy` / `stop_recording` result so
+a future mismatch is reported where it happens. `07`'s DEPLOY step also runs the
+checkpoint it loads, which its own section comment already promised.
--- a/examples/03_record_dataset.py
+++ b/examples/03_record_dataset.py
@@ -34,12 +34,22 @@ if start["status"] != "success":
     raise SystemExit(1)
 
 # Run the policy - each step is automatically captured.
-sim.run_policy(
+# control_frequency must equal the recording's fps: the recorder writes one
+# frame per control step with no decimation, so the 50 Hz default rollout
+# against this 30 fps recording is refused rather than written at a distorted
+# timestamp rate, and the episode would land with zero frames.
+rollout = sim.run_policy(
     robot_name="so100",
     policy_object=MockPolicy(),
     instruction="pick up the red cube",
     n_steps=100,
+    control_frequency=30.0,
 )
+if rollout["status"] != "success":
+    # A refused rollout captures nothing, so stop_recording below would report
+    # an empty dataset instead of the reason. Surface it at its source.
+    print(f"run_policy failed: {rollout['content'][0]['text']}", file=sys.stderr)
+    raise SystemExit(1)
 
 # Finalize - writes parquet + video, ready for lerobot training scripts.
 # The response text reports the actual frame/episode count and on-disk path,
--- a/examples/07_post_tune_any_policy.py
+++ b/examples/07_post_tune_any_policy.py
@@ -38,13 +38,23 @@ sim.start_recording(
     task="pick up the red cube",
     overwrite=True,
 )
-sim.run_policy(
+# control_frequency must equal the recording's fps above: the recorder writes
+# one frame per control step with no decimation, so the 50 Hz default rollout
+# against a 30 fps recording is refused and the episode lands with zero frames.
+rollout = sim.run_policy(
     robot_name="so100",
     policy_object=MockPolicy(),
     instruction="pick up the red cube",
     n_steps=60,
+    control_frequency=30.0,
 )
-sim.stop_recording()
+if rollout["status"] != "success":
+    raise SystemExit(f"rollout failed: {rollout['content'][0]['text']}")
+stopped = sim.stop_recording()
+if stopped["status"] != "success":
+    # An empty dataset here surfaces two steps later as an unrelated-looking
+    # Hub 404 from lerobot, so report the recorder's own reason instead.
+    raise SystemExit(f"recording failed: {stopped['content'][0]['text']}")
 print(f"Recorded LeRobotDataset -> {DATASET_ROOT}")
 
 # 2. TRAIN - the trainer is selected by the SAME name as the inference policy.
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Fixes three shipped examples (`03_record_dataset.py`, `07_post_tune_any_policy.py`, `locomotion/vla_g1_workflow.py`) that declared `start_recording(fps=30)` while their rollouts captured at the 50 Hz `run_policy` default, so the recorder's rate guard (#1746) refused every frame and each example produced an empty dataset while reporting success or failing somewhere unrelated. Each rollout now captures at the rate its recording declares (03/07 pin `control_frequency=30.0`; the G1 workflow moves the recording to `fps=50` since 50 Hz is fixed by the WBC controller's own rate), and each site now checks the `run_policy` / `stop_recording` result so the next mismatch is reported where it happens. `07`'s DEPLOY step now actually runs the loaded checkpoint instead of only printing its type. The new `tests/test_examples_recording_rate_matches_rollout.py` pins the contract repo-wide via static AST reading, reads the default rate from `run_policy`'s own signature rather than restating it, and guards its own blindness with a minimum-graded-examples premise plus six planted positive/negative cases.

## What's good

- Verified independently at head `e66d7582`: the new test passes (8/8), all seven existing `examples/`-scanning guard suites pass (237 passed), the changelog-fragment gate passes, and `examples/03_record_dataset.py` executes end-to-end under `MUJOCO_GL=egl` producing exactly the docstring's promised `100 frames, 1 episode(s)`.
- The test degrades unresolvable rates to "not graded" rather than guessing, and the `_MINIMUM_GRADED_EXAMPLES` premise plus planted-case parametrization mean a clean sweep proves the examples agree rather than that the grader went blind - the pattern AGENTS.md's repo-wide guards already follow.
- Error-path handling in the examples surfaces the library's own refusal text at its source instead of letting it resurface as an unrelated Hub 404, matching the "no silent defaults on error" convention.
- The G1 fix correctly picks the second remedy (record at the rollout's rate) rather than silently changing the WBC gait rate, with the reasoning inline where the next reader needs it.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; no host paths, no non-ASCII in any changed file.

## Verification suggestions

- `MUJOCO_GL=egl python examples/03_record_dataset.py` completes in ~3s and prints `100 frames, 1 episode(s)` (verified in this review).
- `python -m pytest tests/test_examples_recording_rate_matches_rollout.py -q --no-cov` on a pristine `main` worktree should report the single mismatched-examples failure naming all three files, confirming the test fails on pre-fix code per AGENTS.md's regression-pin rule.

Issue #2459, opened 2026-08-19
feat(sim/isaac): LIBERO objects as real meshes - implement add_object mesh_path + SceneObject mesh passthrough (box proxies invalidate pixel-policy cross-backend evals)


yinsong1986:
# Problem

On the Isaac backend, LIBERO task objects are realized as box proxies, and for the mesh-typed objects LIBERO actually uses (bowls, plates), they are not even fitted boxes:

- `SceneObject` is documented as "a box-AABB approximation of the object's collision" (`strands_robots/simulation/isaac/loaders.py`, class docstring).
- `_geom_aabb` returns `None` for mesh geoms ("Mesh / plane / unknown geoms return None"), so a mesh-only body yields no analytic AABB at all, and `_extract_mjcf_shape` degrades a mesh geom to a hardcoded 0.05 m box (`loaders.py:328-330`).
- Consumption points: `IsaacSimulation.load_scene` -> `load_mjcf_scene_objects` (`isaac/simulation.py:2596`) and the LIBERO adapter (`benchmarks/libero/adapter.py:3435`).

This was a latent fidelity gap while nothing pixel-conditioned could run on Isaac. It is now an **eval-integrity** problem: #1802 (GR00T-on-Isaac blocked on `state.x`) is closed, so a GR00T-N1.7-LIBERO checkpoint - trained on MuJoCo-rendered LIBERO visuals with the real object meshes - can be evaluated on Isaac, where it observes gray boxes instead of the bowls/plates it was conditioned on. A cross-backend `success_rate` comparison for a pixel-conditioned policy then measures the visual domain gap as much as the policy, and nothing in the output says so. The backend matrix stays meaningful for physics and timing; the pixel-policy comparison should not be read at face value until this lands.

The robot path already proves the layered-USD approach end-to-end on Isaac (real Franka USD over the Omniverse CDN, RTX-rendered), so the fix is incremental, not architectural.

# Fix (three deliverables + one interim note)

1. **Implement `mesh_path` on Isaac `add_object`.** The API seam already exists and is an explicit refusal today: `add_object: mesh_path=/custom mesh objects are not supported on the Isaac backend yet` (`isaac/simulation.py:2072-2085`). Implementation: convert/reference the mesh as a USD prim (OBJ/STL -> USD via the asset converter, cached), placed at the requested pose. MuJoCo's `mesh_path` behavior is the parity contract.
[...]

PR #2483, opened 2026-08-19
feat(simulation): articulated-container pouring task - carton assets + particle-proxy pour predicates

### LINKED ISSUE #2465: feat(simulation): articulated-container pouring task - hinged carton assets + particle-proxy pour predicates
# Gap

Contact-rich container manipulation (open a carton, pour contents into a receptacle) is a common target task for VLA/WAM post-training, and the stack cannot currently model it: there are no articulated-container assets in the object path, and neither backend exposes liquid simulation through strands-robots (`add_object` is rigid primitives + meshes; Isaac's PhysX particle fluids are not surfaced).

# Proposal (v1 tractable, fluids deferred)

1. **Articulated-container assets**: a small set of registry-style task objects - carton/box with a hinged or sliding cap (MJCF for MuJoCo; USD for Isaac once #2459's mesh path lands) - loadable via the existing object/scene APIs.
2. **Pour-success predicates without fluid sim**: proxy the contents as a small set of rigid particles (spheres) inside the container, and score with the existing `DeclarativeBenchmark` DSL: `fraction_of_particles_in(receptacle) > X`, `spilled_particles < Y`, cap-opened predicate on the hinge joint angle. This makes "poured" measurable and trainable today with zero physics-engine changes.
3. **Example**: a task-authoring example composing these (spawn carton + receptacle, define the benchmark, run a policy, score) - the same pattern as example 11, applied to a contact-rich task.
4. **Follow-up (separate issue when needed)**: true fluid exposure on the Isaac backend (PhysX particle systems through `add_object`/scene APIs). Deliberately out of scope here - the particle-proxy version is what unblocks data collection and benchmarking now.

# Acceptance

- Assets load on MuJoCo (and Isaac when mesh support lands); hinge cap actuates
- The pour benchmark scores deterministically under seed (pinned smoke test)
- Example runs CPU-only in the standard example budget

# Related

- #2459 (Isaac mesh objects - prerequisite for the Isaac half of the assets)
- The predicate DSL (examples 10/11) this builds on

yinsong1986:
Closes #2465

## What changed

v1 of the articulated-container pouring task, fluids deferred exactly as the issue scopes it.

**Bundled task objects** (`strands_robots/simulation/task_objects/` - the first packaged MJCF assets in the tree):

| asset | joint | role |
|---|---|---|
| `hinged_carton` | `cap_hinge` (hinge, rad) | "open the lid" tasks, mounted upright |
| `sliding_carton` | `cap_slide` (slide, m) | lid-down pouring mounts |
| `open_tray` | none | the receptacle the pour is scored against |

Resolved by `task_object_path(name)` / `list_task_objects()` - unknown names are refused with the valid set, traversal is refused by `safe_join`. They load through the **existing** scene APIs (`add_robot(urdf_path=...)`), which is what makes the cap joints observable by the joint predicates (`get_observation` is robot-scoped; the DSL's scene-scope joint ladder resolves the bare `cap_hinge`/`cap_slide` names).

A physics finding drove the asset split, documented in both asset comments: a lid-down **hinge** creeps open under any resting load however large its `frictionloss` - MuJoCo dof friction is a soft constraint that saturates under persistent torque (measured: identical creep at `frictionloss` 0.05 and 2.0). A lid-down **slide** holds closed by geometry (the load is orthogonal to the slide axis), so the sliding carton is the pouring mount and the hinged carton documents its upright mounting. The smoke test pins the closed cap as a stable state over 1000 steps.

**Pour predicates** (closed registry, `strands_robots/simulation/predicates.py`) - contents proxied by rigid spheres, zero physics-engine changes:

- `particles_inside(particles, container, min_fraction, xy_tol, z_tol)` - success: fraction of particles inside the receptacle's containment box (same box semantics as `body_inside`, one shared owner)
- `particles_spilled(particles, containers, max_spilled, xy_tol, z_tol)` - failure: more than `max_spilled` particles inside none of the containers. Degradation is direction-aware: an unresolvable particle can lower a success fraction but can never fire a failure, and an unresolvable container degrades the whole predicate to `False`
- `particles_inside_fraction(particles, container, ...)` - dense reward, `weight * fraction`

Cap-opened is the existing `joint_above` on the cap joint - no new vocabulary needed there. The list-valued kwargs (`particles`, `containers`) get compile-time domain checks (shared `name_list_error`, empty selection refused as an authored constant predicate) and are collected element-wise by `stop_when_referenced_entities` so `stop_when` clauses probe every named body against the live sim, same as the singular kwargs.

**Example**: `examples/17_pour_task.py` - spawn tray + lid-down carton + 6 beads, author the benchmark as a spec dict, score the `mock` policy (0, as narrated), then a scripted pour flips the same predicates the benchmark scores with. CPU-only, `mesh=False`, mock policy, ~20 s - same budget conventions as examples 10/11.

## Test surface

- `tests/simulation/test_pour_predicates.py` (42 stub-level tests incl. task-objects file): counting semantics, every parameter domain, degradation direction, `stop_when` collection, `DeclarativeBenchmark` integration (bool/float kind enforcement)
- `tests/simulation/test_task_objects.py`: catalog resolution, refusals (unknown/traversal/non-name), asset shape pins via stdlib XML (no MuJoCo needed)
- `tests/simulation/mujoco/test_pour_task_smoke.py` (6 tests, 2.3 s): every asset attaches, **hinge cap actuates**, scripted pour flips `particles_inside` False->True with `particles_spilled` quiet, closed cap is stable, scripted pour is bit-deterministic across fresh builds, and the **seeded `evaluate_benchmark` reports identical metrics twice** (the deterministic-under-seed pin)
- `tests/simulation/test_predicate_kwarg_finiteness.py`: `list[str]` added to the annotation classification so the registry-wide drift guard covers the new factories

`hatch run format` + `hatch run lint` clean; full `hatch run test`: **32692 passed, 270 skipped, 0 failed** (95.6% coverage).

## Acceptance criteria from #2465

- [x] Assets load on MuJoCo; hinge cap actuates (`test_hinged_cap_actuates`, `test_every_bundled_task_object_attaches`)
- [x] The pour benchmark scores deterministically under seed (`test_pour_benchmark_scores_deterministically_under_seed`, plus the bit-identical scripted pin)
- [x] Example runs CPU-only in the standard example budget (`mock` policy, `mesh=False`, ~20 s)
- Isaac half of the assets: deferred to when #2459's mesh path lands, per the issue

## Out of scope / follow-ups

- True fluid exposure on Isaac (PhysX particle systems) - explicitly deferred by the issue; file when needed
- USD variants of the cartons for Isaac - blocked on #2459

Board: please move the #2465 card to "In review" on the Strands Labs - Robots project (write access to the org project is not available from this account).


cagataycali (approved):
Reviewed this on a CUDA host with MuJoCo 3.10.0 (the version `uv.lock` resolves), because the load-bearing claims here are physics claims and the interesting ones are not asserted by any test.

Approving. The design decision that shapes this PR checks out, the composition with current `main` is green, and the predicate semantics hold at the tolerances the example and the smoke test actually use. Two non-blocking findings at the end, both docstring/one-liner scale.

## The two-asset split is justified, and the effect is stronger than the comment claims

`hinged_carton.xml` tells users to mount that variant upright, on the grounds that a lid-down hinge "creeps open under any resting load however large the hinge's `frictionloss` is". That sentence is shipped documentation and the justification for bundling two cartons instead of one, and nothing tests it (`test_hinged_cap_actuates` mounts the hinge upright with no load; `test_the_closed_cap_is_a_stable_state` pins the slide). So I measured it: lid-down mount, 4 beads x 0.05 kg resting on the cap, 1000 steps, `frictionloss` substituted in a copy of the asset.

| asset | joint | frictionloss | start | max | final |
|---|---|---|---|---|---|
| hinged_carton | cap_hinge | 0.3 (as authored) | 0.00021 | 0.64903 | 0.64903 |
| hinged_carton | cap_hinge | 0.05 | 0.00022 | 0.84572 | 0.84572 |
| hinged_carton | cap_hinge | 2.0 | 0.00021 | 0.64903 | 0.64903 |
| hinged_carton | cap_hinge | 50.0 | 0.00021 | 0.64903 | 0.64903 |
| sliding_carton | cap_slide | 0.5 (as authored) | 0.00000 | 0.00000 | 0.00000 |

The hinge creeps to 0.649 rad (37 degrees) and the trace is bit-identical to five decimals from `frictionloss` 0.3 all the way to 50.0 - 166x the authored value changes nothing. So the saturation is real and "however large" holds further than the PR claims it does. The slide holds at exactly 0.00000, so the "held by the cavity geometry, not by joint friction" argument is confirmed too.

One precision note, since the sentence is shipped: 0.05 is **not** identical to 2.0 (0.84572 vs 0.64903) - it creeps *further*, because it sits below the saturation floor. The pair that is identical is 0.3 through 50.0. The parenthetical "(measured: identical creep at frictionloss 0.05 and 2.0)" is the one number in that comment I could not reproduce; the conclusion it supports is fine.

## Composition with current main

CI checked out the PR head, which is 63 files behind `main`. The merge is conflict-free with zero file overlap, and on the merged tree the PR's suites plus `test_benchmark_predicates.py` give **363 passed in 2.99 s**. The head alone is 249 passed in 2.58 s. `examples/17_pour_task.py` runs end to end: `mock` scores 0.0 exactly as narrated, then the scripted pour flips `cap_open` and `poured` to True.

## The pour is real physics, and the predicates are right at the tolerances used

Beads go 0.19 -> 0.0214 and 0.0449, i.e. two layers resting on the tray floor (top face 0.010, bead radius 0.012); final `|dxy|` from the tray centre is 0.020 against a 0.106 cavity limit; deepest geom-geom interpenetration over the whole pour is 0.56 mm, which is ordinary soft-contact depth rather than tunnelling.

pour

Closed cap, then t=100 / t=200 / t=560 after sliding it open. Left panel has 0 bead pixels because the contents are inside the carton; the others have 1092 / 1352 / 1352. Clip: https://raw.githubusercontent.com/cagataycali/robots/media-pour-task-2483/pour.mp4

Then I swept four beads through ten end-states (teleported, `mj_forward`, verdicts read off the live sim) at the example's `xy_tol=0.12, z_tol=0.08`. All ten agree with the geometry: every truly-in-tray state scores inside and not spilled, every on-the-ground-outside state scores not-inside and spilled. `xy_tol=0.12` is exactly the tray's outer footprint half-extent, and a bead's own 12 mm radius puts an adjacent-on-floor bead at `|dxy| = 0.132 > 0.12`, so the boundary lands in the right place. That is careful work and worth saying out loud.

## Finding 1 (non-blocking): the defaults are wider than the only receptacle this PR ships, and the two clauses fail in the same direction

`particles_inside` and `particles_spilled` default to `xy_tol=0.15`, while the bundled `open_tray` footprint is +/-0.120 - the default box is 25% wider than the tray. Because both clauses share `_within_box`, a bead outside the tray but inside the box is simultaneously counted as poured *and* as not spilled, so the failure clause cannot catch what the success clause over-counts. Same sweep, defaults alongside the example's values:

| bead state (resting on the ground, off tray centre) | inside @0.12/0.08 | spilled @0.12/0.30 | inside @default | spilled @default |
|---|---|---|---|---|
| 0.125 m off centre (outside the tray) | False | True | True | False |
| 0.135 m | False | True | True | False |
| 0.140 m | False | True | True | False |
| 0.149 m | False | True | True | False |
| 0.160 m | False | True | False | True |

So under the defaults a pour that dumps every bead on the table beside the tray reports `success_rate = 1.0` with zero spill. The same theme shows up in z: the example's `z_tol=0.08` exceeds the tray's 0.054 wall height, and I measured `particles_inside` going True at t=88 with beads at z=0.065, above the rim and still falling.

None of this contradicts a docstring - the box semantics are stated plainly and `body_inside` is explicitly "Approximate" - and both the example and the smoke test pick sound tolerances. What makes it worth raising is that `body_inside` is a query while these two are the *scoring* pair, so there is no second opinion on the box. Cheapest fix is a sentence in the two docstrings (and once in the example) saying to set `xy_tol`/`z_tol` from the receptacle's own half-extents rather than leaving the default, ideally quoting the bundled tray's 0.120 / 0.054 since that is the receptacle everyone will start from. Whether to also tighten the defaults is your call.

## Finding 2 (non-blocking, one word): the pre-arm collector's shape contract is narrower than the validator's

`name_list_error` accepts any `Sequence`, so `particles=("bead_0", "bead_1")` compiles; `stop_when_referenced_entities` gates on `isinstance(value, list)`, so a tuple's elements are never collected:

```
list  : make_predicate -> compiles | stop_when collects bodies = ['b0', 'b1', 'tray']
tuple : make_predicate -> compiles | stop_when collects bodies = ['tray']
```

A spec loaded from JSON or YAML always yields lists, so this is only reachable from a Python caller - but it is exactly the degradation `_BODY_NAME_KWARGS` exists to prevent, per its own comment ("a constant False at evaluation time ... indistinguishable from an honest miss"). `isinstance(value, (list, tuple))` closes it.

PR #2484, opened 2026-08-19
fix(doctor): report on the HuggingFace token file the Hub will read

cagataycali:
## Problem

`doctor`'s `check_hf_auth` reported whether a HuggingFace token was available by reading a hardcoded path:

```python
hf_token_path = Path.home() / ".cache" / "huggingface" / "token"
```

That is not the file `huggingface_hub` reads. The Hub resolves its token file from `HF_TOKEN_PATH`, else `<HF_HOME>/token`, where `HF_HOME` is `$HF_HOME`, else `<XDG_CACHE_HOME>/huggingface`, else `~/.cache/huggingface`. So on any host that relocated its cache - a shared model store, an XDG-configured workstation, a container with `HF_HOME` set - the check answered about a different file, and it was wrong in **both** directions.

Measured against `huggingface_hub.get_token()` on the same environment (`HOME` isolated so the result does not depend on the developer's own cache), `huggingface_hub` 1.12.2:

| environment | Hub resolves | doctor said | |
|---|---|---|---|
| `HF_HOME` points at a dir holding a token | `hf_relocatedToken` | `WARN  No HuggingFace token found` | disagrees |
| `HF_TOKEN_PATH` points at a token file | `hf_relocatedToken` | `WARN  No HuggingFace token found` | disagrees |
| `XDG_CACHE_HOME` relocated, token present | `hf_xdgToken` | `WARN  No HuggingFace token found` | disagrees |
| `HF_HOME` empty, default path holds an old login | `None` | `PASS  HuggingFace token found` | disagrees |
| `XDG_CACHE_HOME` empty, default path holds one | `None` | `PASS  HuggingFace token found` | disagrees |
| default path holds a token (control) | `hf_defaultToken` | `PASS` | agrees |
| default path holds none (control) | `None` | `WARN` | agrees |
| `HF_TOKEN` set (control) | `hf_envToken` | `PASS` | agrees |
| `HUGGING_FACE_HUB_TOKEN` set (control) | `hf_legacyToken` | `PASS` | agrees |

**5 disagreements, 4 controls agreeing.** The false-positive rows are the worse ones: every gated download and every dataset push on that host will fail with a 401, and the check whose stated job is to let users "fix problems before they hit cryptic errors at runtime" reports the setup as fine. The verdict also did not say which file it read, so neither direction was actionable.

Separately, the remedy in the same message - `Run: huggingface-cli login` - cannot be followed. `huggingface-cli` is not published as a console script at the declared `huggingface_hub>=1.5` floor (`console_scripts` there are `['hf', 'tiny-agents']`), and the 1.x releases that do install it exit with `Warning: huggingface-cli is deprecated and no longer works. Use hf instead.` The rest of the package already prescribes `hf auth login` (`dataset_recorder.py:158,214`, `README.md:375`); `doctor` and `docs/troubleshooting.md` were the two places still naming the dead one.

## Change

`strands_robots/doctor.py`

- New `_hf_token_path()` resolves the file the Hub reads. It prefers `huggingface_hub.constants.HF_TOKEN_PATH` - the constant the Hub computed for itself - so the two cannot drift.
- `huggingface_hub` ships only in the extras that pull a checkpoint (`[wbc]`, `[kimodo]`, `[protomotions]`), so a base install has no Hub to ask. The fallback transcribes the same rule and reads each variable by **presence** rather than truthiness, so an explicitly empty value resolves the way the Hub resolves it. A parity test pins the transcription against the Hub's own constant.
- `check_hf_auth` reads that path and names it in both verdicts, so a relocated host can see which file was consulted.
- The path is read with `is_file()` rather than `exists()`: an explicitly empty `HF_TOKEN_PATH` resolves to `Path(".")`, a directory that exists, and reading it would raise.
- The remedy is now `hf auth login`.

`docs/troubleshooting.md` - the one row giving the same login remedy for the same problem, which contradicted `dataset_recorder`.

Behaviour on an unrelocated host is unchanged; that is what the four control tests pin.

## Verification

Measured at `3a8e9cb6` against `upstream/main` `d059c6c3`.

New file `tests/test_doctor_hf_auth_answers_about_the_hub_token.py` (13 tests), covering **both** resolution branches - the Hub's own constant, and the no-Hub transcription.

Copied onto `upstream/main` together with the one repaired existing test: **10 failed / 4 passed** -> all pass after. Every pre-fix failure is behavioural and quotes the mismatch, e.g.

```
AssertionError: XDG_CACHE_HOME relocated the token, but doctor reported
'  WARN  No HuggingFace token found\n        Needed for dataset push + gated
models. Run: huggingface-cli login  # or export HF_TOKEN=hf_...'
```

The 4 tests that pass on both trees are the controls: `HF_TOKEN`, `HUGGING_FACE_HUB_TOKEN`, a token at the default path, and no token anywhere. They fail for the obvious wrong fixes - dropping the env-var precedence, or making the file check unconditional.

`tests/test_doctor.py::test_hf_auth_no_token_anywhere_warns` is repaired rather than left: it forced the warn branch with `monkeypatch.setattr(Path, "exists", lambda _self: False)`, which only worked while the check happened to consult that one method. It now relocates `HF_HOME` to an empty tree and blocks the Hub import, which states its premise ("regardless of the host's own cache") directly. Its assertion is unchanged.

Blast radius - 76 files (every test naming `doctor` / `check_hf_auth` / `HF_TOKEN` / `huggingface` / a docs page, plus the repo-wide docs, changelog, docstring and dependency guards), same selection on both trees:

|---|---|---|---|---|
| this branch | 2 | 2801 | 31 | 2803 |
| `upstream/main` (+ the new tests) | 12 | 2791 | 31 | 2803 |

Branch-only failures: **none**. Base-only: exactly the 10 pre-fix failures, no others. The 2 shared failures are pre-existing `TestDoctorVersionResolution` cases that need installed distribution metadata.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`. `mypy` 24 errors on this branch and 24 on a pristine `upstream/main`, none in the changed files.

## What the operator sees

Both columns run the same probe against the same environment; only `strands_robots` differs. The green line above each pair is what `huggingface_hub` itself resolves.

doctor HuggingFace verdict, before and after

Scenario 1 is an operator who moved the cache to shared storage and logged in there: before, `WARN No HuggingFace token found` for a host that is authenticated, and the prescribed recovery command does not run. Scenario 2 is the same operator before logging in, with an old token still at the default path: before, `PASS HuggingFace token found (~/.cache/huggingface/token)` for a host where the Hub resolves nothing.

## Notes on scope

`examples/libero/run_isaac.py`, `run_isaac_agent.py`, `run_mujoco.py` and `run_mujoco_agent.py` carry both problems - a hardcoded `~/.cache/huggingface/token` and the `huggingface-cli login` remedy. They are left alone here: they are on the LIBERO asset-download path with their own fallback semantics, and folding them in would make this a sweep rather than one check's contract.

### DIFF
--- a/changelog.d/2484-doctor-hf-token-path.md
+++ b/changelog.d/2484-doctor-hf-token-path.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **doctor**: `check_hf_auth` reports on the token file `huggingface_hub` will actually read, resolving it from `HF_TOKEN_PATH` / `HF_HOME` / `XDG_CACHE_HOME` as the Hub does instead of a hardcoded `~/.cache/huggingface/token`, and names that path in its verdict. A relocated cache previously read as "not logged in" while the Hub held a token, and a stale token at the default path read as "logged in" for an environment where the Hub resolved none. The login command offered is now `hf auth login`, matching the rest of the package: `huggingface-cli` is not published as a console script at the declared `huggingface_hub>=1.5` floor, and the later 1.x releases that install it refuse to run.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -64,7 +64,7 @@ description: Error → fix table for the most common gotchas across install, sim
 | `start_recording` fails: lerobot missing | `[lerobot]` not installed | `uv pip install "strands-robots[lerobot]"` |
 | Need MP4 without LeRobot | - | Use `start_cameras_recording` / `stop_cameras_recording` |
 | Empty MP4 files | Stopped before any frames | Check `get_recording_status()` frame count |
-| Push fails | Not logged into HF | `huggingface-cli login` |
+| Push fails | Not logged into HF | `hf auth login` (the `huggingface-cli` entry point is not published at the declared `huggingface_hub>=1.5` floor) |
 
 ## Mesh
 
--- a/strands_robots/doctor.py
+++ b/strands_robots/doctor.py
@@ -94,6 +94,49 @@ def _resolve_version(import_name: str, dist_name: str) -> str:
     return str(getattr(module, "__version__", "unknown"))
 
 
+def _hf_token_path() -> Path:
+    """Path a cached HuggingFace login token is read from.
+
+    ``huggingface_hub`` owns this resolution: ``HF_TOKEN_PATH`` when set, else
+    ``<HF_HOME>/token``, where ``HF_HOME`` is ``$HF_HOME`` when set, else
+    ``<XDG_CACHE_HOME>/huggingface``, else ``~/.cache/huggingface``. Answering
+    about a hardcoded ``~/.cache/huggingface/token`` instead describes a file the
+    Hub will not open on any host that relocated its cache, and it is wrong in
+    both directions: a relocated token reads as "not logged in", and a stale
+    token left at the default path reads as "logged in" for an environment where
+    every Hub call resolves no token at all.
+
+    Prefer the constant the Hub computed for itself, so the two cannot drift.
+    ``huggingface_hub`` ships only in the extras that pull a checkpoint
+    (``[wbc]``, ``[kimodo]``, ``[protomotions]``), so a base install has no Hub
+    to ask and the fallback transcribes the same rule. Each variable is read by
+    presence rather than truthiness, so an explicitly empty value resolves the
+    way the Hub resolves it rather than being treated as unset.
+
+    Returns:
+        The token path, with ``~`` and ``$VAR`` expanded.
+    """
+    try:
+        from huggingface_hub.constants import HF_TOKEN_PATH
+
+        return Path(HF_TOKEN_PATH)
+    except ImportError:
+        # No Hub installed, so nothing here can ask it where it would look;
+        # fall through to the transcription of its rule below.
+        pass
+
+    default_home = os.path.join(os.path.expanduser("~"), ".cache")
+    hf_home = os.path.expandvars(
+        os.path.expanduser(
+            os.environ.get(
+                "HF_HOME",
+                os.path.join(os.environ.get("XDG_CACHE_HOME", default_home), "huggingface"),
+            )
+        )
+    )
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Replaces `check_hf_auth`'s hardcoded `~/.cache/huggingface/token` with the resolution `huggingface_hub` itself uses (`HF_TOKEN_PATH`, else `<HF_HOME>/token`, with `HF_HOME` from `$HF_HOME` / `XDG_CACHE_HOME` / the default), preferring the Hub's own `HF_TOKEN_PATH` constant when installed and transcribing the same rule for base installs where the Hub is absent. The verdict now names the file consulted, and the dead `huggingface-cli login` remedy is replaced with `hf auth login` here and in the one matching `docs/troubleshooting.md` row. Verified independently: the fallback transcription matches the installed Hub's own constant computation exactly (presence-based reads, `expandvars(expanduser(...))`, empty-string handling via `is_file()`); the 13 new tests pass on this branch and 9 of them fail on `upstream/main` with behavioural messages, with the 4 controls passing on both trees as described.

## What's good

- Changelog fragment present (`changelog.d/2484-doctor-hf-token-path.md`), per AGENTS.md step 3.
- The parity test (`test_the_fallback_resolves_what_the_hub_resolves`) pins the transcription against the Hub's own constant in a subprocess, including the empty-string spellings, so the two statements of the rule cannot drift silently.
- The repaired `test_hf_auth_no_token_anywhere_warns` states its premise by relocating `HF_HOME` instead of patching `Path.exists`, which was coupled to the implementation.
- `except ImportError` is narrow, and the `is_file()` guard handles the `HF_TOKEN_PATH=""` -> `Path(".")` edge that `exists()` would trip on.
- Scope discipline: the `examples/libero/` instances of the same defect are named and deliberately left out rather than swept in.

## Verification suggestions

The `Fatal Python error: Aborted` that a full local `tests/test_doctor.py` run can hit is in `check_sim_smoke`'s MuJoCo GL path and reproduces identically on `upstream/main` on the same host - it is a host GL environment issue, not this diff.

PR #2485, opened 2026-08-19
fix(rendering): check GsplatBackground's alignment flags instead of reading them by truthiness

cagataycali:
## Problem

`GsplatBackground.__init__` takes four flags that each select an alignment or compositing branch, and all four were read by truthiness rather than checked:

| flag | read as | selects |
|---|---|---|
| `auto_backdrop` | `auto_backdrop and transform is None` | fit a backdrop `world_from_gs` |
| `skybox` | `skybox and transform is None` | fit the live-backdrop alignment, and pick the `bg_fill` default |
| `metric` | `bool(metric)` | keep the capture's own scale instead of scaling to `radius` |
| `own_floor` | `bool(own_floor)` | hide MuJoCo's grid ground so the photoreal floor shows |

Truthiness inverts exactly the spellings an operator reaches for. Measured through the constructor, on all four flags:

| value passed | branch selected |
|---|---|
| `True` | enabled |
| `"false"`, `"False"`, `"no"`, `"off"`, `"0"` | **enabled** - the branch the value asks to skip |
| `0`, `""`, `[]`, `None` | default, without being a declared spelling of it |

Nothing raised and nothing logged, so the wrong branch was indistinguishable from the right one. The two flags that compose with `transform is None` additionally *stored* the raw value, so `bg._skybox` was `None` or `[]` rather than a bool.

`metric` is the sharpest of the four, because it also decides whether `radius` is read at all - the fit is `s = 1.0 if metric else radius / r95`. On a real 500k-splat capture, asking for `radius=2.5` with `metric` spelled `"no"`:

| | stored `_metric` | fitted scale | room radius | ceiling |
|---|---|---|---|---|
| `metric="no"` | `True` | 1.0000 | **4.45 m** (1.78x the request) | 2.04 m |
| `metric=False` | `False` | 0.5662 | 2.54 m (honored) | 1.02 m |

So the requested size was dropped and the scene stood up at whatever size it happened to be captured at, behind a `status`-free successful construction.

This constructor already raises eagerly for a nonexistent or unreadable `ply_path`, and its own comment gives the reason: the heavy work is deferred to the first `render`, which "in app contexts sits inside a catch-all that demotes the photoreal background to a procedural fallback". A branch selected wrongly has the same problem - by the time it shows up it is a frame that looks slightly off, not an argument that can be reported. `HybridCompositor` in this package checks its own `blend_in_linear` explicitly for the same reason; its `Raises:` block records it as "every one of these was previously coerced or clamped into a plausible-but-different render". These four were the remaining posture flags in `strands_robots.rendering` that were not checked.

## Change

`strands_robots/rendering/backgrounds.py` - check all four on the shared `boolean_flag_error` domain, beside the path validation and before any state is stored, raising `ValueError`. The refusal names the flag and the value. The class docstring gains a `Raises:` block, which previously documented none of the constructor's three verdicts.

Nothing else moves: the flags keep their `and transform is None` composition, `radius` and the other numeric knobs keep their own float-coercing treatment (a knob that scales a quantity and a flag that selects a posture are deliberately inverse domains), and the path refusals still come first.

## Tests

`tests/rendering/test_gsplat_background_posture_flag_domain.py` - **42 failed / 18 passed** against `main`, 60 pass after. Dependency-free: the constructor defers every `gsplat`/`torch` import to the first render, and the scale consequence is measured through the pure-numpy skybox fit, so none of this needs the `sim-gs` extra.

The headline test fails on `main` with the measurement rather than a bare "did not raise":

```
GsplatBackground(metric='no') stored _metric=True, so the fit kept the capture's own
scale and placed the room at a 3.41 m radius instead of the requested 2.50 m.
```

(The test uses a synthetic room so it needs no asset; the 4.45 m figure above is the
same measurement on the real capture.)

The refusal tests parametrize over `boolean_flag_error` itself and premise-assert against it, so a spelling added to the shared domain is covered here without an edit.

The 18 that pass on both trees are the controls, and each fails for a specific wrong fix: a real `bool` still selects the branch it names; the defaults still leave every flag off; an explicit `transform` still gates the fitted modes off (so checking the flag has not changed the composition); the shipped `tabletop` skybox preset - which carries two of these flags - is still accepted verbatim; the path refusals still come first; and `radius` beside them is untouched.

## Verification

Blast radius: all of `tests/rendering` plus every test referencing the changed symbols, `boolean_flag_error`, `own_floor` or `AGENTS.md`, plus the repo-wide docstring / source-string / changelog / dependency guards - 102 files, run on this branch and on `main` with the new test file copied in.

|---|---|---|---|---|
| this branch | 1 | 3932 | 32 | 3965 |
| `main` | 43 | 3890 | 32 | 3965 |

Failed delta 42 == passed delta 42 == the pre-fix failure count, on the same collected total. Branch-only failures: none. The 42 base-only ids are all from the new file. The one shared failure is pre-existing on `main` (`test_safety_timing_and_replay_defenses.py::TestResumeLockoutTimingOracleClosed::test_correct_override_code_clears_lockout`, `AttributeError: 'Mesh' object has no attribute '_safety_publishers_lock'`).

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on both this branch and a pristine `main`, none in the changed files. All of the above was re-measured at `9e9cc09` (the amended head, which carries a `mypy`-only fix to the new test file's kwargs builder); the counts and the pre-fix split are unchanged from the first push.

## Artifact

Both panels come from one script run against each tree: the same shipped `.spz` reader, the same shipped skybox fit, the same real 500k-splat capture, both asking for `radius=2.5`. Only `strands_robots` differs - one tree accepts `metric="no"` and one refuses it, and the script applies whatever remedy the refusal it gets names. The so101 silhouette is drawn to scale from the compiled model (0.65 m across, 0.24 m tall) with a MuJoCo headless render as the reference.

GsplatBackground posture-flag scale evidence

## Scope

`PanoramaBackground` declares no posture flag, and the numeric knobs in this constructor are a separate domain, so this is the whole of the class's flag surface. `_fit_skybox_transform`'s own docstring claim about which asset formats are already metric is a question about the formats rather than about the flag, and is left alone.

## Round changelog

**Round 2 (`247c047`) - base merge only, no change to the reviewed diff.**

#2475 landed a sibling posture-flag bullet at the same `AGENTS.md` anchor (the "Corrections from code review" list), which put this branch in conflict. Merged `main` in and kept both bullets, the landed one first; each names its own pin inline, so the shared `Pinned by` trailer needed no edit.

The conflict resolution is the entire content of the merge: `git show --cc` is 15 lines, all of them the reinsertion of #2475's bullet. The diff under review is byte-identical to round 1 - `git diff main...HEAD --stat` is still the same 4 files, +293/-1. No source, test or changelog file was touched.

Noting for the record that the same single-anchor collision this file documents for `CHANGELOG.md` (step 3: "every branch inserts at the same anchor") now reproduces on `AGENTS.md`, and here it cost an approval - two independent learnings appended to one list conflicted on ordering alone. Raising that separately rather than in this PR.


yinsong1986 (dismissed):
## Summary

Adds a `boolean_flag_error` check for `GsplatBackground`'s four posture flags (`auto_backdrop`, `skybox`, `metric`, `own_floor`) in the constructor, beside the existing path refusals and before any state is stored, raising `ValueError` naming the flag and the value. This is the established posture-flag pattern from AGENTS.md ("Posture flags are checked, never read by truthiness"), applied to the remaining unchecked flags in `strands_robots.rendering`, matching `HybridCompositor`'s `blend_in_linear` and the recent ROS 2 command-surface fix. The behaviour change (previously-coerced non-bool values now refuse) is loud and explicit, which is exactly the sanctioned shape for this class of fix.

## What's good

- Reproduced the claimed test split independently at `9e9cc095`: 42 failed / 18 passed with `main`'s `backgrounds.py`, 60/60 pass at the head — the 18 controls pin the defaults, the real-bool spellings, the `transform is None` composition, the shipped `tabletop` preset, path-refusal ordering, and `radius`'s deliberately separate numeric treatment.
- Refusal tests parametrize over `boolean_flag_error` itself with premise assertions, so the shared domain stays the single authority — per the AGENTS.md pattern the other five posture-flag test files use.
- Validation sits before any attribute assignment after the path checks, so a refused construction stores nothing; changelog fragment present; AGENTS.md bullet added with its test pin; no host paths, no emojis, test is dependency-free (no `sim-gs` extra needed).

yinsong1986 (approved):
## Summary

Applies the established posture-flag convention (AGENTS.md > "Posture flags are checked, never read by truthiness") to `GsplatBackground`'s four alignment/compositing flags (`auto_backdrop`, `skybox`, `metric`, `own_floor`): each is now checked against the shared `boolean_flag_error` domain in the constructor, beside the existing path refusals and before any state is stored, raising `ValueError` naming the flag and the value. The class docstring gains a `Raises:` block covering all three constructor verdicts, a changelog fragment is present under `changelog.d/`, and the AGENTS.md learning bullet is added at the same anchor as its siblings. Round 2 is a base merge resolving the `AGENTS.md` anchor collision with #2475 only; the reviewed diff is unchanged from round 1.

## What's good

- Verified locally at `247c0479`: the new test file passes 60/60 on this branch and splits exactly 42 failed / 18 passed with the same file copied onto `main`, matching the PR description's pre-fix claim; all of `tests/rendering` passes (600 passed, 1 skipped) and `ruff check` / `ruff format --check` are clean on the changed files.
- The refusal tests parametrize over `boolean_flag_error` itself with premise assertions, so a spelling added to the shared domain is covered without an edit - the pattern the existing posture-flag pins use.
- The controls pin the composition invariants that a wrong fix would break: an explicit `transform` still gates the fitted modes off, the shipped `tabletop` skybox preset (which carries two of these flags) is still accepted, the path refusals still come first, and `radius` keeps its float-coercing numeric treatment (the deliberate inverse domain).
- The check sits before any state is stored, so a refused construction has no partial effect, and the docstring cross-references in the new test module resolve (repo docstring-xref guards pass).
- No semantic-conflict exposure: nothing that landed on `main` since the merge base touches `strands_robots/rendering/` or `strands_robots/utils.py`.

PR #2487, opened 2026-08-19
fix(simulation): check randomize()'s axis flags instead of reading them by truthiness

cagataycali:
## Problem

`randomize()` takes one flag per axis, and each selects a *posture* - run this axis or leave it alone - rather than scaling a quantity. Every non-empty string is truthy, so `"false"` / `"no"` / `"off"` / `"0"` - the spellings you reach for when turning an axis off - turned that axis **on**, and the call reported the axis applied.

Measured on MuJoCo, one 1 kg cube per colour on a floor, `seed=11`:

| `randomize_colors=` `randomize_physics=` `randomize_positions=` | status | colours re-sampled | mass/friction rescaled | objects displaced |
|---|---|---|---|---|
| `False` (a real bool) | success | no | no | no |
| `"false"` / `"no"` / `"off"` / `"0"` | **success** | **yes** | **yes** | **yes** |
| `0` / `""` / `None` / `[]` | success | no | no | no |

The two axes that ran there default to `False` precisely because they cannot be undone - the docstring says so ("Randomization IS destructive; recompile the scene to undo"). The physics axis rewrites `body_mass` / `body_inertia` / `geom_friction`, and the position axis rewrites `model.qpos0`, the pose a `reset` restores. So the misread outlives the call: in the capture below every object moves 44.1 mm and is **still 44.1 mm out after `reset()`**.

Newton has the same three declared flags and stores them in its randomization spec through `bool()`, so `randomize_colors="false"` was stored as `True` and the axis re-applied on **every later rebuild**, not just the call that misread it.

Newton's MuJoCo-parity refusal branches on `randomize_positions`, so it inherited the misread:

```
randomize(randomize_positions="false")
-> error: randomize_positions is not supported by the Newton backend yet.
          Supported axes: ... Use the MuJoCo backend for object-position randomization.
```

an unsupported-axis verdict for a caller who had asked *not* to randomize positions.

A misspelled axis **name** was already refused with the valid list, which the `**kwargs`-typed `SimEngine.randomize` base signature exists to require - "so a misspelled axis cannot report success while leaving that axis untouched". The value that name carries had no such check.

## Change

Each axis flag is checked on the shared `boolean_flag_error` domain before anything is written, on both backends that implement `randomize()`.

- **MuJoCo**: all four declared flags, checked with the other caller-value checks and ahead of the first model write - and ahead of the `randomize_lighting` refusal, which branches on its own flag and so would otherwise describe the axis the caller had asked to skip.
- **Newton**: the three declared flags plus `randomize_positions`, which is declared only by MuJoCo and so arrives through `**kwargs`. It is checked **before** the parity refusal, which is what stops that refusal blaming an unsupported axis for a misread value.

The numeric knobs in the same signature keep their own domain: a `mass_range` is a quantity and is still refused as a range. That half was promoted to a shared rule across both backends when the physics defect it caused was fixed; this brings the flag half along, so the accepted spellings cannot diverge between backends either.

## Sim capture

Same call on both trees, three renders each: the scene as built, after the call, and after `reset()`.

randomize axis flag domain

|  | status | colours | mass | max displacement | after `reset()` |
|---|---|---|---|---|---|
| before | `success` | red `[0.9, 0.15, 0.15]` -> `[0.216, 0.549, 0.641]` | `1.0` -> `1.2686` | **44.1 mm** | **44.1 mm** |
| after | `error` | unchanged | unchanged | 0.0 mm | 0.0 mm |

Panel 1 is **pixel-identical across the two trees** (mean abs level `0.00000`), so both start from the same scene; panel 2 differs by `20.834` and panel 3 by the same `20.834` - the displacement is not something the reset undoes.

## Tests

`tests/simulation/test_randomize_axis_flag_domain.py`, parametrized over `boolean_flag_error` itself rather than a copied spelling list, and over the axis flags read from the live signatures - so a spelling added to the shared domain, or an axis added to a backend, is graded without an edit.

Against `upstream/main` (0acbfc93): **96 failed / 21 passed**. Every one of the 21 is a control that passes on both trees:

- the six documented spellings keep working: MuJoCo `False` leaves every axis alone and `True` still applies one; Newton `False` stores exactly `False`; Newton still refuses `randomize_positions=True`; a misspelled axis *name* still names the parameter that was meant; and `mass_range=(0, 0)` is still refused as a **range**, not as a flag.
- the premise tests, which keep the refused set the shared domain's rather than a list copied into the file, plus a non-vacuity check that the signature scan found the flags it grades.

Local gate at `b3d89ebd`, 37 files covering every test that touches randomization, the flag domain or a docs page, plus the repo-wide docstring / changelog / string guards:

- branch **2535 passed, 0 failed**, 12 skipped
- base **2439 passed, 96 failed**, 12 skipped, the same 2535 collected
- branch-only failures: none. Every base failure is from the new file, and the selection has no pre-existing failures.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on both trees, none in the changed files.

### DIFF
--- a/changelog.d/2487-randomize-axis-flag-domain.md
+++ b/changelog.d/2487-randomize-axis-flag-domain.md
@@ -1,0 +1,19 @@
+### Fixed
+
+- **simulation**: `randomize()` checks each axis flag on the shared boolean-flag
+  domain instead of reading it for truthiness, on both backends that implement
+  it. `randomize_colors` / `randomize_lighting` / `randomize_physics` /
+  `randomize_positions` select a posture, so `"false"`, `"no"`, `"off"` and
+  `"0"` - every spelling an operator reaches for to turn an axis off - are
+  non-empty strings that turned that axis **on** while the call reported it
+  applied. Randomization is destructive and the physics and position axes
+  default to `False` for that reason: the misread rescaled mass, inertia and
+  friction and displaced every object by up to 44 mm, and because the position
+  axis writes `model.qpos0` the displacement survived `reset()`. On Newton the
+  flags are stored through `bool()`, so a misread persisted to every later
+  rebuild, and the MuJoCo-parity refusal - which branches on
+  `randomize_positions` - answered `randomize_positions="false"` with
+  "not supported by the Newton backend", an unsupported-axis verdict for a
+  caller who had asked *not* to randomize positions. A misspelled axis *name*
+  was already refused with the valid list; this is the same guarantee for the
+  value that name carries.
--- a/docs/simulation/domain-randomization.md
+++ b/docs/simulation/domain-randomization.md
@@ -33,6 +33,25 @@ sim.randomize(randomize_position=True)   # singular
 #                      'randomize_positions', 'seed']
 ```
 
+**An axis flag must be a boolean.** The four flags select a *posture* - run this
+axis or leave it alone - so each is checked rather than read for truthiness.
+Every non-empty string is truthy, so `"false"`, `"no"`, `"off"` and `"0"` - the
+spellings you reach for when turning an axis off - each turned that axis **on**,
+and the call reported it applied. That is the same guarantee as the misspelled
+key above, one level down: a value an axis cannot be read from can never look
+like a successful randomization either.
+
+```python
+sim.randomize(randomize_physics="false")
+# status=error: randomize: randomize_physics must be a boolean, got 'false'.
+#              It selects a posture, so it is not read for truthiness ...
+```
+
+Both backends that implement `randomize()` (MuJoCo and Newton) check the same
+domain, so an axis you can turn off on one is not left on by the other. The
+numeric knobs in the same signature keep their own domain: a `mass_range` is a
+quantity, and it is still refused as a range rather than as a flag.
+
 **Destructive** - writes into MuJoCo model arrays. To restore: `load_scene(...)` or recreate the sim.
 
 **Every axis survives `reset()`.** A reset restores the world's initial state,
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Extends the repo's established posture-flag rule (AGENTS.md > "Posture flags are checked, never read by truthiness") to `randomize()` on both backends that implement it: each axis flag (`randomize_colors` / `randomize_lighting` / `randomize_physics` / `randomize_positions`) is now checked on the shared `boolean_flag_error` domain before anything is written, so `"false"` / `"no"` / `"off"` / `"0"` refuse loudly instead of silently running a destructive axis. On MuJoCo the check sits ahead of the first model write and ahead of the `randomize_lighting` authored-light refusal (which branches on its own flag); on Newton the `**kwargs`-carried `randomize_positions` is checked before the MuJoCo-parity refusal, so that refusal can no longer report an unsupported axis to a caller who spelled "off" as a string. Verified locally at `b3d89ebd`: the new test file passes 117/117 on the branch and fails 96/21 against `upstream/main` (0acbfc93), matching the PR description exactly; `ruff check` / `ruff format --check` clean on the three changed Python files.

## What's good

- Tests parametrize over `boolean_flag_error` itself and over the live signatures rather than copied spelling/axis lists, so a spelling or axis added later is graded without an edit - the pattern every prior posture-flag PR here uses.
- Controls pin that the documented spellings, the Newton `bool()` storage of a real `False`, the parity refusal on a genuine `True`, the misspelled-name refusal, and the numeric `mass_range` domain are all unchanged.
- Changelog fragment (`changelog.d/2487-...`) and docs page updated in the same PR; error paths return structured error dicts, never raise; refusals happen before any model/spec write so a refused call has no partial effect to undo.

Issue #2472, opened 2026-08-19
decision(policies/cosmos3): eight backend knobs are reachable only by hand-building the backend, and the registry route drops them silently - promote a subset, or refuse the drop


cagataycali:
## Context

#2458 corrected `docs/policies/cosmos3.md`, which told the reader to pass
`enable_safety_checker=True` on a surface that does not accept it. The fix was
documentation: name `Cosmos3DiffusersBackend` and show the object route
(`Cosmos3Policy(diffusers_backend=...)`). It deliberately did **not** change the
constructor, because picking which knobs belong on the policy is a design call
rather than a correction. This issue is that design call.

## The gap, measured

`Cosmos3DiffusersBackend.__init__` (`strands_robots/policies/cosmos3/policy_diffusers.py:206`)
takes 13 parameters. When `Cosmos3Policy` builds the backend itself
(`policy.py:286`) it forwards exactly three:

```python
self._diffusers = _Backend(
    embodiment=self.embodiment,
    model=model or pretrained_name_or_path,
    mode=mode,
)
```

So eight tuning knobs are reachable only by constructing the backend yourself and
injecting it:

| knob | default |
| --- | --- |
| `resolution_tier` | `480` |
| `view_point` | `"ego_view"` |
| `device` | `None` (cuda if available) |
| `dtype` | `"bfloat16"` |
| `num_inference_steps` | `35` |
| `guidance_scale` | `6.0` |
| `enable_sound` | `False` |
| `enable_safety_checker` | `False` |

(The remaining two, `pipeline` and `condition_cls`, are test seams and should stay
off any public config surface.)

The registry route cannot reach them either. `providers/cosmos3.config_keys` in
`strands_robots/registry/policies.json` is:

```
['embodiment', 'host', 'port', 'action_space', 'observation_mapping',
 'action_mapping', 'robot', 'prompt', 'api_key', 'backend', 'mode', 'model']
```

None of the eight appear, and `build_policy_kwargs` keeps only declared keys, so a
caller who sets one in a config dict gets a successful run with the setting
silently dropped. #2458 pinned that drop as current behaviour; it did not make it
reachable.

## Why this is worth settling rather than leaving documented

Sibling providers already promote these exact knobs, so the tree is currently
inconsistent about where a diffusion knob lives:

| provider | promotes |
| --- | --- |
| `kimodo` (diffusion) | `guidance_scale`, `diffusion_steps`, `device`, `dtype` |
| `lerobot_local` | `device` |
[...]

PR #2488, opened 2026-08-19
feat(training): sagemaker trainer provider - submit a TrainSpec as a managed training job

### LINKED ISSUE #2463: feat(training): sagemaker trainer provider - submit a TrainSpec as a managed training job wrapping the containerized trainer path
# Gap

`create_trainer()` has four real providers (`lerobot_local` in-proc, `groot` docker, `cosmos3` recipe->`cosmos_framework` launch, plus RL trainers). All of them assume the training host is the local machine. There is no provider that submits the same `TrainSpec` as a managed cloud training job, so scaling a post-training run beyond the local GPUs means leaving the `TrainSpec` abstraction and hand-rolling job submission.

Prior art: #138 (closed) sketched an EC2/SageMaker fine-tuning tool in the pre-`TrainSpec` era; the trainer registry that exists now is the right seam for the same idea.

# Proposal

A `sagemaker` trainer provider that wraps an existing local provider rather than reimplementing any training logic:

1. **Containerize the wrapped path**: the `groot` provider already runs in docker, and the `cosmos3` provider drives `cosmos_framework` programmatically - either is directly packageable as a SageMaker training container.
2. **`TrainSpec` -> job mapping**: `dataset_root` (S3 or HF-bucket URI) -> input channel; `output_dir` -> model artifact path; `extra` -> hyperparameters; `validate()` runs locally BEFORE submission so a bad spec is refused without paying job-start latency (same fail-fast contract as the local providers).
3. **Result contract**: same `TrainResult` shape (status, checkpoint_dir pointing at the downloaded/S3 artifact), so `create_policy(result.checkpoint_dir)` works unchanged downstream.
4. **Optional dependency**: gated behind a `[sagemaker]` extra (`boto3`/`sagemaker` SDK) via `require_optional()`; zero import cost otherwise.

# Non-goals

- No orchestration/pipelines (Step Functions etc.) - one spec, one job
- No new training logic - the provider is transport, the wrapped trainer is behavior

# Acceptance

- Unit tests with a stubbed SageMaker client: spec->job mapping, validate-before-submit, error propagation (a failed job is a raised error with the job name, never a silent success)
- One `tests_integ/` smoke gated on AWS credentials, skipped otherwise
- README Configuration section documents the extra and required IAM surface

yinsong1986:
Closes #2463

## What

A `sagemaker` trainer provider: `create_trainer("sagemaker")` submits the same `TrainSpec` the local providers consume as **one managed SageMaker training job** wrapping a containerized trainer image. Transport only - the training behavior lives in the container (`image_uri`), which packages one of the local trainer paths; the provider is the `TrainSpec -> CreateTrainingJob` mapping and nothing else (per the issue's non-goals: no pipelines, no new training logic).

## How it maps

| `TrainSpec` | SageMaker job |
|---|---|
| `dataset_root` (must be `s3://`) | `training` input channel (`/opt/ml/input/data/training` in-container) |
| `output_dir` (must be `s3://`) | `OutputDataConfig.S3OutputPath`; success reports `checkpoint_dir` at the `model.tar.gz` artifact |
| `num_nodes` | `ResourceConfig.InstanceCount` |
| remaining spec fields + `extra` | string `HyperParameters` (strings verbatim, everything else JSON-encoded; one documented decoder for the container side) |

- **validate-before-submit**: `validate()` is local and pure (no boto3 import, no network) - the shared security gate first, then the shared numeric domains (`_run_size_problems`, `_learning_rate_problems`, `_launch_topology_problems`, `_seed_problems`, `_validation_episodes_problems`, `_lora_hyperparameter_problems`) for exactly the fields this provider serializes, then the hyperparameter mapping itself (collisions with forwarded field names, JSON-encodability, the API's 100-hyperparameter / 2500-char limits). A bad spec never pays job-start latency.
- **Error propagation**: a `Failed`/`Stopped` job is an error `TrainResult` naming the job and carrying SageMaker's `FailureReason` - never a silent success. (The trainer ABC's error contract is an error result rather than a raise - `train()` converts every failure to a result, same as the other providers - so "raised error" from the issue's acceptance is delivered as the never-silent error result the `Trainer` contract mandates.) A job outliving the local poll budget reports `running` and stays pollable.
- **`status(job_name)`**: real override - SageMaker jobs are the detached-job case the ABC documents; maps `InProgress -> running` and terminal states exactly as `train()` does.
- **Clock rule**: the poll deadline is a duration measured on `time.monotonic()`; the job-name stamp is a recorded point in time.
- **Optional dependency**: new `[sagemaker]` extra = `boto3>=1.34.0,<2.0.0` (same bounds as `[mesh-iot]`; deliberately NOT the heavy `sagemaker` SDK - the low-level API is all the transport speaks), gated via `require_optional()`. Zero import cost otherwise; `validate()` works without the extra installed. Registered through the factory's lazy loader so `list_trainers()` discovers it (there is no policy-side `"trainer"` block to hang it on - this is a training transport with no paired inference provider).
- **`hardware_floor`**: `{min_gpus: 0, min_vram_gb: 0, multinode: True}` - the submitting host needs no GPU.

## Test surface

- `tests/training/test_sagemaker.py` (30 tests, stubbed client via `sys.modules` + a swapped `require_optional` cache, same pattern as the other provider tests): factory wiring, validate problems (spec + transport config, parametrized), spec->job mapping, hyperparameter encoding, validate-before-submit (the fake client records zero calls for a bad spec), poll-until-terminal, failed/stopped/submission/polling error propagation, exhausted-poll-budget -> `running`, `status()` mapping, missing-boto3 install hint.
- `tests_integ/training/test_sagemaker_smoke.py`: skips cleanly without boto3/credentials; with credentials it exercises a real `DescribeTrainingJob` refusal (error result naming the job); the full submission half is gated on `STRANDS_SAGEMAKER_SMOKE_IMAGE_URI` / `_ROLE_ARN` / `_S3_PREFIX`.
- Updated the guards a fifth concrete trainer must satisfy: `test_trainer_provider_docstrings.py` module set, and the non-vacuity sets in `test_run_size_domain.py` / `test_learning_rate_domain.py` / `test_launch_topology_domain.py` (the new backend routes through all the shared gates, which those scans verify).
- `uv.lock` regenerated for the new extra (minimal diff: the extra + its `requires-dist` row).

## Acceptance criteria from #2463

- [x] Unit tests with a stubbed SageMaker client: spec->job mapping, validate-before-submit, error propagation (a failed job is an error naming the job, never a silent success)
- [x] One `tests_integ/` smoke gated on AWS credentials, skipped otherwise
- [x] README documents the extra (extras table row) and the required IAM surface (caller: `sagemaker:CreateTrainingJob`, `sagemaker:DescribeTrainingJob`, `iam:PassRole`; execution role: S3 read/write, ECR pull, CloudWatch Logs)

## Validation

`hatch run format` clean, `hatch run lint` clean (ruff + mypy), `hatch run test`: **32702 passed, 270 skipped, 0 failed** (full suite, run with EGL so the GL-dependent tests execute).

## Out of scope / follow-ups

- A ready-made container entry point that rebuilds a `TrainSpec` from `/opt/ml/input/config/hyperparameters.json` and dispatches to a wrapped local provider (the job contract is documented; shipping a reference `Dockerfile` + entry script is a natural follow-up).
- An `export()` that downloads/extracts the S3 `model.tar.gz` to a local directory for direct `create_policy()` consumption - today `checkpoint_dir` is the S3 artifact URI and the download is the caller's step.

Board: please move #2463 to **In review** on the [project board](https://github.com/orgs/strands-labs/projects/2) (filing account lacks org project write).


cagataycali (commented):
Read this end to end and exercised the submission path against a stubbed client rather than reading the check mark. The layering is the right one - a transport-only provider that reimplements no training logic - and the guarantees I could measure hold. **One substantive item**, with two measured instances of the same class, plus one drift guard I would like before this lands.

## What I verified

`tests/training/test_sagemaker.py` is **47 passed** locally. The 8 failures in the sibling domain files I also ran (`test_launch_topology_domain.py`, `test_run_size_domain.py`) are `ModuleNotFoundError: torch` on this host and pre-exist on the base - nothing attributable to this diff.

| claim | verdict |
|---|---|
| `validate` needs no boto3 and costs nothing to import | holds - every probe below ran in a venv where `import boto3` is `ModuleNotFoundError` |
| fail-closed: a spec with problems never touches the API | holds |
| a failed / stopped job is an error naming the job + `FailureReason` | holds, never a silent success |
| poll budget exhausted maps to `running`, not success | holds - the job is not assumed dead |
| `Completed` maps to the S3 artifact as `checkpoint_dir` | holds |
| every `TrainSpec` field reaches the container | holds - see below, and this is the one I most expected to find broken |

That last row is worth stating explicitly because it is the failure mode a `TrainSpec -> str/str hyperparameters` mapping usually has. It does not have it:

```
unforwarded, unhandled: []
forwarded but not a field: []
```

23 `TrainSpec` fields, 3 handled specially (`dataset_root`, `output_dir`, `extra`), and `_FORWARDED_FIELDS` is exactly the remaining 20. Nothing is silently dropped today.

Packaging is right too: boto3-only with the reason given (the provider speaks the low-level API, so the heavy `sagemaker` SDK would be dead weight), the same major-cap as `[mesh-iot]` per convention 2, and `all` updated - the extra is discoverable rather than orphaned.

## Substantive: `validate` implements two of the four CreateTrainingJob bounds it enumerates, and both gaps land as an API refusal

The module docstring's contract is that a bad spec is refused locally, and `hyperparameter_problems` states it directly: an oversized value "would be refused by CreateTrainingJob only after submission". Two limits in that same enumeration are not enforced, so for those inputs the refusal is exactly the one the file says it prevents.

### 1. `_BASE_JOB_NAME_RE` bounds the alphanumeric count, not the length

The pattern is `^[a-zA-Z0-9](-*[a-zA-Z0-9]){0,31}$`. The `{0,31}` caps how many alphanumeric characters appear; `-*` inside the group is unbounded, so the total length is not capped at all:

| `base_job_name` | length | `validate` | generated `TrainingJobName` | over 63 |
|---|---|---|---|---|
| `strands-robots` | 14 | 0 problems | 39 | no |
| `a` x 32 | 32 | 0 problems | 57 | no |
| `a` x 33 | 33 | **refused** | 58 | no |
| `a-` x 25 + `b` | 51 | **0 problems** | **76** | **yes** |
| `a` + `-` x 100 + `b` | 102 | **0 problems** | **127** | **yes** |

A 33-character name is correctly refused while a 51-character one passes, which is what makes this a defect rather than a deliberate relaxation. And end to end through a stubbed client, `validate` passing is not academic - the request is built and sent:

```
base_job_name len= 14  validate_problems=0  api_called=1  job_name_len=39  result=success
base_job_name len= 51  validate_problems=0  api_called=1  job_name_len=76  result=error
   message: SageMaker training job 'a-a-a-...-b-20260819-201253-182954df' submission failed: ValidationException...
```

The file already states the invariant it wants, one line above the regex: *"The base job name keeps 32 chars so the appended UTC stamp + entropy suffix (25 chars) always fits."* The regex does not implement that sentence, and the `validate` message - "must be 1-32 alphanumeric-or-hyphen characters" - describes the intent rather than what is enforced.

Worth naming the general cause, because it will recur wherever an AWS pattern is adopted: AWS expresses `TrainingJobName` as a pattern **plus a separate max-length constraint of 63**, and only the pattern was carried over. A pattern copied without its length constraint under-enforces by construction. `(?=.{1,32}$)` in front of the existing expression, or a `len()` check beside it, closes it.

### 2. The hyperparameter value cap is checked; the key cap is not

`_MAX_HYPERPARAMETER_VALUE_LEN = 2500` is correct and enforced. AWS also caps each hyperparameter **key** at 256 characters, and that one is absent:

```
extra value 2501 chars                       problems=1  ("...caps a hyperparameter value at 2500")
extra key 300 chars (AWS caps keys at 256)   problems=0
```

Same shape as the job name: `extra` is the agent-reachable escape hatch, so an over-long key is a reachable input, and the refusal moves to the API.

### Why I am raising these rather than filing them

Neither costs money or corrupts anything - `create_training_job` refuses before an instance is provisioned, and the code converts that into an error `TrainResult` naming the job, so the failure is legible. I am raising them because the entire justification for this provider having a 100-line `validate` is that the refusal happens before job-start latency, and these are the inputs where it does not. The fix for both is a bounds check each. Left alone, the docstring and the error message state a guarantee the code does not keep, which is more expensive than the missing check.

## Drift guard: nothing pins `_FORWARDED_FIELDS` to `TrainSpec`

Coverage is complete today, and there is no test anywhere that says so - `grep -rn _FORWARDED_FIELDS --include=*.py` returns only the five occurrences inside `sagemaker.py`. So the next field added to `TrainSpec` is silently dropped by this provider: the container never sees it, `validate` reports nothing, and the job trains with a setting the caller did not ask for while reporting success. That is the failure direction that cannot be caught by reading the diff that introduces it, because that diff will be in `base.py`.

This is the class #2472 is an open decision issue about - a registry route dropping knobs silently - and it is cheap to pin here, five lines:

```python
def test_every_trainspec_field_is_forwarded_or_deliberately_special() -> None:
    special = {"dataset_root", "output_dir", "extra"}  # channel, artifact path, spliced in
    fields = {f.name for f in dataclasses.fields(TrainSpec)}
    assert fields - special == set(_FORWARDED_FIELDS)
```

Asserting equality in both directions is what makes it work: it fails when a field is added to `TrainSpec` and not forwarded, and also when a name is removed from `TrainSpec` while `_FORWARDED_FIELDS` keeps it (which would make `_forwarded_items`' `getattr` raise `AttributeError` out of a pure mapping function). The `special` set names the three exemptions in the test, so a future reader sees why they are exempt rather than inferring it.

## Smaller notes

- `num_gpus` travels as a hyperparameter while the GPU count is really decided by `instance_type`, and nothing reconciles them - `num_gpus=8` on `ml.g5.xlarge` validates clean and fails inside the container after provisioning, which is the one failure mode here that bills. I am **not** asking for an instance catalog; the docstring's reason for not enumerating one is sound, and the live catalog would drift. One sentence on `num_gpus` in the constructor docstring saying it must agree with `instance_type` and is not cross-checked would put the constraint where the person setting it is looking.
- A `create_training_job` call that fails after the job was actually created (a response timeout, say) returns an error result, and a retry generates a fresh name and submits a second job. Rare and generic to the API rather than to this diff - noting it because two billing jobs is the outcome, and `status(job_name)` is the tool that resolves it.
- `_POLL_GRACE_S` being justified as the queueing/provisioning/upload time that bills *outside* `MaxRuntimeInSeconds` is the correct reasoning, and measuring the deadline on `time.monotonic()` while the job-name stamp uses `time.gmtime()` is the right split for the two different jobs those clocks do.
- `_TERMINAL_STATUSES` deliberately excluding `Stopping` so the loop keeps polling to a real `Stopped` is a good detail, and the comment says why.

Nothing else blocking. Happy to approve once the two bounds are checked and the field-coverage guard is in - the rest of this provider is ready, and the parts that are hard to get right here (fail-closed submission, terminal-status mapping, the timeout not masquerading as either success or death) are the parts that are already correct.

yinsong1986:
@cagataycali Thanks for exercising the submission path rather than reading the check mark - all four items are addressed in 29379f1d, point by point:

**1. `_BASE_JOB_NAME_RE` bounds the alphanumeric count, not the length** - fixed @ 29379f1d. Took your lookahead suggestion so the bound lives in the same expression the message describes:

```python
_BASE_JOB_NAME_RE = re.compile(r"^(?=.{1,32}$)[a-zA-Z0-9](-*[a-zA-Z0-9]){0,31}$")
```

The comment above it now states the general cause you named - AWS expresses `TrainingJobName` as a pattern **plus** a separate 63-char max length, and only the pattern was carried over - so the next AWS pattern adopted in this file has the warning sitting next to the precedent. Your two counterexamples are pinned as regression cases (`"a-" * 25 + "b"` and `"a" + "-" * 100 + "b"`, both refused now), with a comment explaining why a hyphen-padded name used to pass. A control test pins the other direction: a 32-char name still validates clean and `_job_name()` stays within the 63-char cap, so the fix can't silently over-tighten either.

**2. The hyperparameter key cap is absent** - fixed @ 29379f1d. `_MAX_HYPERPARAMETER_KEY_LEN = 256`, checked in the same `hyperparameter_problems` loop as the value cap, over the same candidate set (forwarded fields + `extra`), so the two caps can't drift apart in coverage. The `hyperparameter_problems` docstring now says "an oversized key, value or count". Regression case: a 300-char `extra` key is refused naming the cap.

Agreed on the framing for both: the cost was never the API refusal itself (legible, pre-provisioning, free) - it was `validate`'s docstring and message stating a guarantee the code didn't keep.

**3. Drift guard for `_FORWARDED_FIELDS`** - added @ 29379f1d, essentially your five lines: `fields(TrainSpec) - {dataset_root, output_dir, extra} == set(_FORWARDED_FIELDS)`, equality in both directions, with the `special` set commented inline (channel, artifact path, spliced in) and a docstring naming both failure directions - the silently-dropped new field, and the removed field that would make `_forwarded_items`'s `getattr` raise out of a pure mapping function. It lives in `tests/training/test_sagemaker.py::TestHyperparameterMapping`.

**4. Smaller notes:**
- `num_gpus` vs `instance_type`: one sentence added to the `instance_type` arg in the constructor docstring @ 29379f1d - `spec.num_gpus` must agree with the GPUs the type carries, travels as a hyperparameter, is NOT cross-checked against the instance catalog, and a mismatch fails inside the container only after the instance is provisioned (which bills). No catalog added, per your note.
- Retry-after-response-timeout submitting a second billing job: agreed it's real and generic to `CreateTrainingJob` rather than to this mapping. Leaving it as-is for this PR - the error result names the job that may exist, and `status(job_name)` is the documented resolution. A dedup submission token would need a `ListTrainingJobs`-by-prefix read, which is a different IAM surface than the three calls the module documents; happy to file a follow-up if you want it tracked.
- Poll-grace clock split and `Stopping` excluded from `_TERMINAL_STATUSES`: no change needed, thanks for confirming the reasoning reads back.

**Validation:** `hatch run lint` clean; `tests/training/test_sagemaker.py` is **54 passed** (your 47 + the 7 new: 2 job-name-length refusals, 3 length controls, 1 key-cap refusal, 1 field-coverage guard); the full `tests/training/` domain is 2614 passed with one pre-existing environmental error (`test_lerobot_e2e.py`, a torchcodec shared-library load failure on this host that reproduces identically with this diff stashed - same class as the torch-less failures you noted on yours).

cagataycali (approved):
Re-reviewed on the merged tree rather than the PR head, and re-ran the measurements from scratch instead of citing my own numbers from the last round. All four items are addressed, and more usefully, all three new guards are load-bearing: I mutated each one back and every mutation fails.

## The two bounds, re-measured end to end

Driving the real `train()` path against a stubbed client, so "validate passed" is graded on whether the API was actually touched:

| `base_job_name` | len | problems | api called | `TrainingJobName` | over 63 |
|---|---|---|---|---|---|
| `strands-robots` | 14 | 0 | 1 | 39 | no |
| `a` x 32 | 32 | 0 | 1 | 57 | no |
| `a` x 33 | 33 | **1** | **0** | - | - |
| `a-` x 25 + `b` | 51 | **1** | **0** | - | - |
| `a` + `-` x 100 + `b` | 102 | **1** | **0** | - | - |

Both counterexamples from the last round are refused before submission now, where they previously validated clean and generated 76- and 127-character names. `status=error` on all three refusals, `create_training_job` never reached.

The control direction matters as much and holds - the lookahead did not over-tighten. Sweeping hyphen-padded names across the boundary: 30, 31, 32 accepted; 33, 34 refused. So the bound lands exactly where the comment says it does, and the generated name at 32 is 57 characters.

Key cap boundary is also exact, which is the easy thing to get wrong by one:

| key length | verdict |
|---|---|
| 255 | accepted |
| 256 | accepted (AWS caps *at* 256) |
| 257 | **refused**, naming the cap |
| 300 | **refused** |

Two 300-character keys produce two problems, one per key, rather than short-circuiting on the first. The value cap is unchanged as a control: 2500 accepted, 2501 refused. And `validate` still needs no boto3 - with `sys.modules["boto3"] = None` it returns 0 problems rather than raising.

Coverage of the key cap over forwarded fields as well as `extra` reads back correctly: the candidate set is `_forwarded_items(spec) + extra.items()`, and the longest forwarded field name is 19 characters, so the shared loop is the right place for it and the two caps cannot drift apart in coverage.

## The guards are not vacuous

This is the part I care about most, because a guard that passes without being able to detect anything is worse than no guard. Reverting each thing it guards, in the merged tree, restoring after each:

| mutation | result |
|---|---|
| drop the `(?=.{1,32}$)` lookahead (back to main's pattern) | **2 failed** - exactly `overrides6` and `overrides7`, the two counterexamples |
| `if False and len(key) > _MAX_HYPERPARAMETER_KEY_LEN` | **1 failed** - `expected a problem containing 'caps a hyperparameter key'` |
| add `gradient_accumulation: int = 1` to `TrainSpec` | **1 failed** - the drift guard, printing the forwarded set |

That third one is the one worth stating plainly: it is exactly the failure mode the guard exists for - a field added in `base.py`, by a diff that never touches this file, silently dropped by this provider. It now fails at the point the field is added. Baseline is 54 passed, so each mutation is a real signal rather than noise.

Field coverage computed live on the merged tree: 23 `TrainSpec` fields, 3 special, `_FORWARDED_FIELDS` exactly the remaining 20, `unforwarded unhandled: []` and `forwarded but not a field: []`.

## Merged-tree gate

The PR head was 3 commits behind main, so I built the merge locally rather than reading the check mark against the head. Clean merge, zero conflicts, and `comm -12` of the two changed-file lists is empty - the 11 files main moved and the 12 this PR changes do not intersect at all, so there is nothing for the composition to get wrong. Verified both sides survive by symbol anyway (lookahead present, key cap present, drift guard present; main's audit poison record, gsplat posture-flag test and robot_mesh readonly-audit test all present).

| | collected | passed | skipped | failed |
|---|---|---|---|---|
| merged | 2619 | 2615 | 4 | **0** |
| pristine main | 2514 | 2510 | 4 | **0** |
| delta | **+105** | **+105** | 0 | 0 |

Collected delta equals passed delta, and it itemizes exactly: 54 new in `test_sagemaker.py`, +20 `test_learning_rate_domain`, +30 `test_run_size_domain`, +1 `test_factory` (the registry-enumeration test picking up the provider). `list_trainers()` gains exactly `sagemaker` and nothing else. `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1381 files). mypy is 19 errors on the merged tree and 19 on pristine main, branch-only diff empty, none naming a file this PR touches. `tests_integ/training/test_sagemaker_smoke.py` is 2 skipped without credentials, which is the right shape for a smoke file.

## Approving

Everything I asked for is in, correct at the boundaries, and provably detecting the regression it guards. The parts that are hard to get right in a transport provider - fail-closed submission, terminal-status mapping, a timeout that masquerades as neither success nor death - were already right last round and are unchanged.

One observation I am explicitly **not** blocking on, because it is not this PR's to fix. Python's `$` matches before a trailing newline, so a `base_job_name` ending in `\n` slips past both the pattern and the new length bound:

```
base_job_name="strands-robots\n"   problems=0  api=1  TrainingJobName='strands-robots\n-20260820-...'
base_job_name="a"*32 + "\n"  (33)  problems=0  api=1  submitted
```

The newline is invisible to the length half because the lookahead's `$` stops before it. That is the same pattern-under-enforces-its-own-intent class as the length bound, and `\Z` closes it. I am not asking for it here: `^...$` is the universal idiom in this codebase - 41 compiled patterns across `strands_robots/` use it and **zero** use `\Z` - so changing it in one file would be inconsistent rather than better, and the outcome is the same legible pre-provisioning API refusal the two bounds produced before this PR. If it is worth doing it is worth doing as its own sweep across all 41, and I am happy to file that separately. Mentioning it only because your new comment above the regex is now the best explanation of this class anywhere in the tree, and the same reasoning covers the anchor as well as the length.

PR #2489, opened 2026-08-19
fix(simulation/terrain): check the resolution that sizes the heightfield before generating it

cagataycali:
## Problem

`generate_heightfield` is the exported, backend-independent heightfield generator behind
`create_world(terrain=...)`. Its `resolution` is the grid-cell count its documented
`resolution * resolution` output length is squared from - and it was read with a bare
`int(resolution)`, while `difficulty`, the other caller-supplied knob in the same module, is
measured against the shared continuous domain by `validate_difficulty`.

That docstring already names, axis by axis, why the hand-rolled form was wrong. All three axes
were still open for `resolution`, measured on `upstream/main`:

| `resolution=` | outcome on main | why it matters |
|---|---|---|
| `39.7` | **accepted**, 1521 floats | the documented length is `resolution * resolution`; 39.7 squared is 1576.09 |
| `2.5` | **accepted**, a 2x2 field | a count computed by division silently collapses to the minimum |
| `"40"` | **accepted**, 40x40 | a resolution read from a config file or argv is refused nowhere |
| `None`, `[]` | `TypeError` from `int()` | outside the `ValueError` every terrain refusal here uses |
| `inf` | `OverflowError` from `int()` | same |
| `nan`, `'abc'` | `ValueError`, but `int()`'s own text | names neither the parameter nor the surface |

The truncation is not caught downstream either - it is only *renamed*. A caller who sizes an
`<hfield>` from the number they passed gets MuJoCo's `elevation data length must match nrow*ncol`,
which names neither `resolution` nor the truncation, so the one value they got wrong is reported
nowhere. That is the same misdiagnosis `validate_difficulty` was moved onto the shared domain to
avoid.

## Change

`resolution` goes through `positive_whole_number_error` before anything is generated, bound the
same way `validate_difficulty` binds the continuous domain (helper as the predicate, this module's
own message on the raise). Every value in the table above is now one actionable `ValueError` naming
the parameter.

The `>= 2` floor stays a **separate** check with its message unchanged. The shared domain answers
whether the value is a usable count at all; refusing a 1x1 field - which MuJoCo does compile - is
this module's own decision, and `1` still reaches it. Ordering is domain-then-floor, and the kind
is still checked first, so a caller who got both wrong is told about the kind.

`AGENTS.md` already states this rule under *A resolution knob is validated before the work it
sizes*; this is the terrain generator complying with it, so no convention text changed.

## Verification

**Regression tests** extend `tests/simulation/test_terrain.py`, so its 27 existing tests are
controls. Split against `upstream/main` (the file copied in): **16 failed / 54 passed -> 70 passed**.
Every pre-fix failure is behavioural and quotes the value and the consequence, e.g.

```
resolution=39.7 was accepted and returned 1521 floats, so the documented
resolution * resolution length (1576.09) does not hold

resolution=inf raised OverflowError, which a ValueError-only caller does not
see as a refusal: cannot convert float infinity to integer
```

The expected verdicts are read from `positive_whole_number_error` itself rather than a copied value
list, with a premise assert per case, so a value the shared domain starts accepting or refusing is
covered without an edit.

**Controls that pass on both trees** - each fails for a specific shortcut fix:

- `4`, `4.0` and the module default still build that grid, and `4.0` gives the identical field `4`
  does (the domain deliberately accepts an integral float).
- `1` still hits the `>= 2` floor with its own message - fails if the domain is treated as the whole
  check and the floor is dropped or folded in.
- An unknown kind keeps its actionable message even with a bad resolution.

**Local gate** at `c1554b7d`, against `upstream/main` `0acbfc93`: 50 test files touching terrain,
the shared domain or the repo-wide docstring/changelog/string guards - branch **3727 passed, 0
failed**; base with the new test file copied in **16 failed, 3711 passed**. `3711 + 16 == 3727`, the
same collected total, no branch-only failures, and no pre-existing failures in the selection.
`ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` 24 errors on both
trees with none in the changed files.

## The grid a caller did not ask for

terrain resolution

One capture script, run once per tree with `sys.path` pointed at that tree; only `strands_robots`
differs. It calls `generate_heightfield` directly, feeds the result to an `<hfield>` exactly as the
MuJoCo spec builder does, and drops 25 balls so the facet size reads. Relief is scaled to
`difficulty=6.0` (0.48 m peak) for print legibility.

- **Left, `resolution=40` (control)**: the same world in both trees - identical settled ball spread
  to 17 significant figures (0.22609517258814468 m), mean abs pixel delta 6.25e-06.
- **Middle, `resolution=8.9` on `upstream/main`**: an 8x8 = 64-cell field, accepted, nothing raised.
  Its large blocky facets differ from the requested-scale 40x40 field over **41.2%** of the frame.
- **Right, `resolution=8.9` on this branch**: refused up front, naming the parameter, so no
  heightfield and no world are built from a number the module cannot honour.

### DIFF
--- a/changelog.d/2489-terrain-resolution-shared-domain.md
+++ b/changelog.d/2489-terrain-resolution-shared-domain.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **simulation/terrain**: `generate_heightfield` measures `resolution` against the shared
+  `positive_whole_number_error` domain before generating anything, so the grid-cell count that
+  sizes the returned field is checked the way `difficulty` already is. A bare `int(resolution)`
+  truncated a fractional count silently - `39.7` returned 1521 floats for a number whose square is
+  1576.09 - accepted a string outright, and let `None`/`[]`/`inf` raise `TypeError`/`OverflowError`
+  from inside `int()`, outside the `ValueError` every other terrain refusal uses. The `>= 2` floor
+  is unchanged.
--- a/strands_robots/simulation/terrain.py
+++ b/strands_robots/simulation/terrain.py
@@ -37,7 +37,7 @@ from __future__ import annotations
 
 import random
 
-from strands_robots.utils import positive_finite_number_error
+from strands_robots.utils import positive_finite_number_error, positive_whole_number_error
 
 # Supported terrain kinds. ``"rough"`` is smoothed value-noise bumps; ``"stairs"``
 # is a flight of discrete parallel step plateaus rising along +x; ``"pyramid"`` is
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Moves `generate_heightfield`'s `resolution` onto the shared `positive_whole_number_error` domain before any grid is built, closing the three axes a bare `int(resolution)` left open (silent truncation of a fractional count, string acceptance, and `TypeError`/`OverflowError` escaping the module's `ValueError` contract). The binding mirrors `validate_difficulty` exactly - shared helper as the predicate, this module's own message on the raise - and the `>= 2` floor stays a separate, unchanged check with `1` still reaching it, so ordering (kind, then domain, then floor) is preserved. Verified locally at head: the pre-fix defect reproduces on main (`resolution=39.7` returns 1521 floats, `"40"` is accepted), all 70 tests in `tests/simulation/test_terrain.py` pass on the branch, and `ruff check`/`ruff format --check` are clean on both changed files.

## What's good

- Complies with AGENTS.md > "A resolution knob is validated before the work it sizes", including the normalize-after-checking detail (`int(resolution)` still runs after the domain check, so an integral float like `4.0` builds the identical grid `4` does - pinned by a control test).
- Reuses the existing shared guard rather than adding a new single-caller one (AGENTS.md convention 11), and parametrizes expected verdicts over the domain function itself rather than a copied value list.
- The behavior change is loud, not silent: previously-accepted spellings now raise one actionable `ValueError` naming the parameter, inside the `ValueError` contract `create_world` implementations already narrow to.
- Changelog fragment present (`changelog.d/2489-terrain-resolution-shared-domain.md`); scope is tight at 3 files.

Issue #2462, opened 2026-08-19
test(policies/cosmos3): tests_integ real-inference coverage - 13 unit files, zero integration tests (rule-8 gap)


yinsong1986:
# Gap

AGENTS.md rule 8: "Integration tests required - each policy needs `tests_integ/` tests with real inference." The cosmos3 provider has 13 unit-test files (`tests/policies/cosmos3/`: wire format, action decode, embodiments, IK solve loop, diffusers, registry) and **zero entries in `tests_integ/`** - no test anywhere runs real Cosmos 3 inference.

Every other inference-capable provider convention holds or is tracked; cosmos3 is the outlier, and it is the provider whose integration spans the most moving parts (external `cosmos_framework` policy server, msgpack+NumPy wire protocol, diffusers pipeline, IK decode).

# Scope

Two integration tests, mirroring how the other providers structure theirs:

1. **Service mode**: start the RoboLab policy server (per the module docstring quickstart: `cosmos_framework.scripts.action_policy_server_robolab` with `nvidia/Cosmos3-Nano-Policy-DROID`), connect `create_policy("cosmos3", ...)`, run a short MuJoCo rollout, assert actions are finite/in-range and the episode completes. Skips cleanly when the server/checkpoint/GPU is unavailable.
2. **In-proc diffusers mode**: load the pipeline directly, one `get_action` round-trip on a real observation, same assertions.

Both assert on outputs (behavior), not internal state, per the testing conventions.

# Why now

The provider is being recommended as the integrated WAM-class model precisely because of its complete integration path - an integration nobody executes in CI-adjacent form is a claim, not a fact. The unit suite cannot catch a wire-format or checkpoint-schema drift against the real upstream.

# Related

- Sibling: #2461 (OpenArm embodiment for cosmos3) - its acceptance path reuses these tests' scaffolding

Board routing: please add to https://github.com/orgs/strands-labs/projects/2 with Status=Todo, Priority=Medium (filing account lacks org project write).

yinsong1986:
Decomposition plan (this issue proved too large for a single implementation session; the children below carry the work and each is single-session sized).

[...]

PR #2492, opened 2026-08-19
feat(policies/cosmos3): OpenArm embodiment - post-training path for openarm

### LINKED ISSUE #2461: feat(policies/cosmos3): OpenArm embodiment - Cosmos3Embodiment entry + post-training path for openarm/bi_openarm
# Gap

The cosmos3 provider's embodiment table (`strands_robots/policies/cosmos3/embodiments.py`) configures `droid`, `umi`, `av`, and `bridge`. There is no entry for OpenArm, although the robot is fully present everywhere else in the stack: registry entries `openarm`, `bi_openarm`, `openarm_follower`, `openarm_bimanual` (sim MJCF via `openarm_v1_mj_description`), and lerobot >= 0.6.1 ships `openarm_follower` / `bi_openarm_follower` hardware drivers.

So the record -> post-train -> deploy loop is closed for OpenArm with every provider except cosmos3, where the embodiment mapping is the one missing piece.

# Scope

1. Add a `Cosmos3Embodiment` entry for OpenArm: name, conditioning `domain_name`, `raw_action_dim`, and the output action-column mapping (single-arm first; `bi_openarm` as a follow-up once the single-arm mapping is validated).
2. Wire it through the same paths the existing four embodiments use (service mode and in-proc diffusers; `sim_ik.py` decode if the domain emits eef-space actions).
3. Document the post-training expectation: a new domain is only as good as its SFT data, so the entry ships with a docstring note that it requires post-training on OpenArm episodes via the existing `cosmos3` trainer (recipe TOML -> `cosmos_framework` launch) rather than promising zero-shot behavior.

# Acceptance

- Unit tests in `tests/policies/cosmos3/test_embodiments.py` cover the new entry (action-dim/mapping round-trip), matching the existing embodiment tests
- A MuJoCo rollout smoke path (mock server or recorded fixtures) exercises the OpenArm mapping end-to-end
- No silent fallback: an unknown embodiment name keeps failing loudly as today

# Related

- Sibling: #2462 (cosmos3 tests_integ gap)

yinsong1986:
# feat(policies/cosmos3): OpenArm embodiment - post-training path for openarm

Closes #2461

## What changed

The cosmos3 provider's embodiment table (`strands_robots/policies/cosmos3/embodiments.py`) gains an `openarm` entry, closing the record -> post-train -> deploy loop for the Enactic OpenArm with cosmos3 the way `droid`/`umi`/`av`/`bridge` already close it:

- **`Cosmos3Embodiment("openarm")`**: conditioning domain `openarm_lerobot`, `raw_action_dim=10` (the standard single-arm unified action: 9D EE pose + 1D grasp), `midtrain` action space, front + wrist camera keys (mirroring the recorded OpenArm episode set / `openarm_real` lerobot embodiment). Aliases `openarm_lerobot`, `openarm_follower`, `enactic_openarm` resolve to it.
- **Wired through the same paths as the existing four** - the embodiment table is data-driven, so service mode, the in-proc diffusers backend, and the `sim_ik.py` decode (the domain emits eef-space actions) pick the entry up through `get_embodiment` with no per-backend special cases. Deliberately **no `joint_pos` layout**: the RoboLab server's joint conversion is DROID-only, so promising one here would fabricate a server post-process that does not exist (pinned by a test).
- **Post-training expectation documented on the entry itself**: no released Cosmos 3 checkpoint ships this domain, so the entry maps a checkpoint post-trained on OpenArm episodes via the existing `cosmos3` trainer (SFT recipe TOML -> `cosmos_framework` launch) rather than promising zero-shot behavior. Consistently, **no action stats are bundled** for `openarm_lerobot`: `load_action_stats` keeps refusing the unbundled domain by name, and the post-trained domain's own quantiles go in via the existing `stats=` + `stats_domain="openarm_lerobot"` route.
- Docstrings/messages that enumerate the embodiments or the shared 10-column width (`policy.py`, `__init__.py`, `sim_ik.py`, `registry/policies.json` description) updated to stay truthful.

## Test surface

- `tests/policies/cosmos3/test_embodiments.py` (`TestOpenArmEmbodiment`): spec, action-dim/mapping round-trip (every named layout matches its width; `midtrain` IS the raw layout, ending in `grasp`), no fabricated `joint_pos`, aliases, and the bundled-stats refusal naming the domain + the safe explicit-stats remedy.
- `tests/policies/cosmos3/test_sim_ik_openarm.py` (new): MuJoCo rollout smoke path on the single-arm OpenArm MJCF (`robot_descriptions.openarm_v1_mj_description`, the same module the robot registry's `openarm` entry declares) - mock policy server -> `Cosmos3Policy` per-step column dicts -> reassembled chunk -> de-normalize + IK bridge -> joint targets within the existing tracking-error bar (mean <= 12 mm / max <= 45 mm) and inside the model's own joint limits; plus the no-stats loud failure and the full-camera-set client-side guard. Skips cleanly (`importorskip`) without the `cosmos3-sim` extra, mirroring `test_sim_ik.py`.

`hatch run format` / `hatch run lint` clean; full `hatch run test`: **32617 passed, 270 skipped, 0 failed** (run with `MUJOCO_GL=egl` - the default GL backend has no display on the dev runner, which is environmental).

## Acceptance criteria (from #2461)

- [x] Unit tests in `tests/policies/cosmos3/test_embodiments.py` cover the new entry (action-dim/mapping round-trip), matching the existing embodiment tests
- [x] A MuJoCo rollout smoke path (mock server) exercises the OpenArm mapping end-to-end
- [x] No silent fallback: an unknown embodiment name keeps failing loudly (pinned), and the unbundled `openarm_lerobot` stats are refused by name rather than substituted

## Out of scope / follow-ups

- `bi_openarm` (16-DOF bimanual) embodiment - deliberate follow-up per the issue, once this single-arm mapping is validated on post-trained weights (the bimanual MJCF is already in `openarm_v1_mj_description`).
- GPU integration tests against a real post-trained checkpoint - sibling issue #2462 (cosmos3 `tests_integ` gap).

Board: please move #2461 to "In review" on the [Strands Labs - Robots board](https://github.com/orgs/strands-labs/projects/2) (filing account lacks org project write).

### DIFF
--- a/README.md
+++ b/README.md
@@ -600,7 +600,7 @@ create_policy("lerobot/act_aloha_sim_transfer_cube")   # local HF inference
 |----------|---------|-------|
 | `mock` | none | Sinusoidal trajectories; `requires_images=False` (~10x faster) |
 | `groot` | NVIDIA GR00T N1.5/N1.6/N1.7 | Service mode (ZMQ to a Docker container) or local in-process (`model_path=`) |
-| `cosmos3` | NVIDIA Cosmos 3 omnimodal VLA | Service mode (WebSocket to a Cosmos Framework RoboLab policy server); embodiments: `droid`, `umi`, `av`, `bridge` |
+| `cosmos3` | NVIDIA Cosmos 3 omnimodal VLA | Service mode (WebSocket to a Cosmos Framework RoboLab policy server); embodiments: `droid`, `umi`, `av`, `bridge`, `openarm` |
 | `lerobot_local` | HuggingFace | Direct ACT / Pi0 / SmolVLA / Diffusion inference, no server |
 | `lerobot_async` | HuggingFace via gRPC | Offload a LeRobot policy to a remote `PolicyServer` over lerobot's native async-inference gRPC transport (edge/light robot host) |
 | `remote` | any policy, over WebSocket | Drop-in client that forwards observations to a remote `PolicyServer` and returns its action chunk: `create_policy("remote", endpoint="ws://gpu-box:8765")` (or the smart string `create_policy("ws://gpu-box:8765")`). For a light robot host with a GPU box elsewhere; mirrors the server policy's RTC support |
@@ -728,7 +728,8 @@ arm, so use the `franka` (or `panda`) sim asset:
 MUJOCO_GL=egl python examples/vla/cosmos3_sim_rollout.py --record /tmp/c3.mp4
 ```
 
-Embodiments: `droid` (10D, chunk 32, 15 fps), `umi`, `av`, `bridge`. If the
+Embodiments: `droid` (10D, chunk 32, 15 fps), `umi`, `av`, `bridge`, `openarm`
+(post-training only). If the
 server is not running, the policy raises a `ConnectionError` with the exact
 command to start it.
 
--- a/changelog.d/2461-cosmos3-openarm-embodiment.md
+++ b/changelog.d/2461-cosmos3-openarm-embodiment.md
@@ -1,0 +1,18 @@
+### Added: cosmos3 OpenArm embodiment (post-training path)
+
+The cosmos3 provider's embodiment table gains an `openarm` entry
+(`Cosmos3Embodiment`: domain `openarm_lerobot`, 10D unified action = 9D EE pose
++ 1D grasp, `midtrain` action space, front + wrist camera keys), closing the
+record -> post-train -> deploy loop for the Enactic OpenArm with cosmos3 the
+way `droid`/`umi`/`av`/`bridge` already close it. The entry is data-driven
+through the same service-mode, in-proc diffusers and `sim_ik` decode paths as
+the existing four; aliases `openarm_lerobot`, `openarm_follower` and
+`enactic_openarm` resolve to it. It documents the post-training expectation
+explicitly: no released Cosmos 3 checkpoint ships this domain, so it maps a
+checkpoint post-trained on OpenArm episodes via the existing `cosmos3` trainer
+(SFT recipe TOML -> `cosmos_framework` launch) rather than promising zero-shot
+behavior, and no action stats are bundled - the decode refuses the unbundled
+domain by name and accepts the post-trained domain's own quantiles via
+`stats=` + `stats_domain="openarm_lerobot"`. Unknown embodiment names keep
+failing loudly. `bi_openarm` (bimanual) is a deliberate follow-up once the
+single-arm mapping is validated.
--- a/docs/policies/cosmos3.md
+++ b/docs/policies/cosmos3.md
@@ -1,5 +1,5 @@
 ---
-description: NVIDIA Cosmos 3 omnimodal VLA - WebSocket service, droid/umi/av/bridge embodiments, MuJoCo rollout.
+description: NVIDIA Cosmos 3 omnimodal VLA - WebSocket service, droid/umi/av/bridge/openarm embodiments, MuJoCo rollout.
 ---
 
 # Cosmos 3
@@ -27,7 +27,7 @@ python -m cosmos_framework.scripts.action_policy_server_robolab \
 
 ```python
 Cosmos3Policy(
-    embodiment="droid",          # droid | umi | av | bridge
+    embodiment="droid",          # droid | umi | av | bridge | openarm
     host="localhost",
     port=8000,
     action_space=None,
@@ -52,6 +52,7 @@ Cosmos3Policy(
 | `umi` | UMI gripper | - |
 | `av` | Autonomous vehicle cameras | - |
 | `bridge` | Bridge dataset robots | - |
+| `openarm` | Enactic OpenArm (7-DOF + gripper) | `"openarm"` |
 
 ### Action spaces
 
@@ -244,7 +245,7 @@ concern, not an IK one.)
 #### De-normalization stats are per domain
 
 The de-normalize step needs that domain's own `q01`/`q99` quantiles. Two domains
-ship them bundled; the other two registered embodiments do not:
+ship them bundled; the other three registered embodiments do not:
 
 | embodiment | domain | raw dim | bundled stats |
 |---|---|---|---|
@@ -252,9 +253,10 @@ ship them bundled; the other two registered embodiments do not:
 | `bridge` | `bridge_orig_lerobot` | 10 | yes |
 | `umi` | `umi` | 10 | no |
 | `av` | `av` | 9 | no |
+| `openarm` | `openarm_lerobot` | 10 | no |
 
 `nvidia/Cosmos3-Edge` documents its forward-dynamics example on `umi` and its
-inverse-dynamics example on `av` — exactly the two without bundled quantiles — so
+inverse-dynamics example on `av` — both without bundled quantiles — so
 driving the sim bridge from Edge means supplying that domain's stats yourself:
 
 ```python
@@ -266,8 +268,9 @@ out = decode_cosmos_chunk_to_targets(
 ```
 
 `stats_domain` is required whenever `stats` is passed, and must match the
-embodiment's domain. It is not bookkeeping: `umi`, `droid_lerobot` and
-`bridge_orig_lerobot` are all 10 columns, so the width check cannot tell one
+embodiment's domain. It is not bookkeeping: `umi`, `droid_lerobot`,
+`bridge_orig_lerobot` and `openarm_lerobot` are all 10 columns, so the width
+check cannot tell one
 domain's quantiles from another's, and the two bundled domains disagree by up to
 **2.77x** on the physical translation they decode from the same normalized
 action. Substituting another domain's stats would rescale every commanded pose
--- a/strands_robots/policies/cosmos3/__init__.py
+++ b/strands_robots/policies/cosmos3/__init__.py
@@ -36,8 +36,10 @@ In MuJoCo (the ``droid`` embodiment drives a Franka/DROID-class arm - use the
                    control_frequency=15.0)
 
 See ``examples/vla/cosmos3_sim_rollout.py`` for a complete, runnable rollout +
-recording. Available embodiments: droid, umi, av, bridge
-(see :mod:`strands_robots.policies.cosmos3.embodiments`).
+recording. Available embodiments: droid, umi, av, bridge, openarm
+(see :mod:`strands_robots.policies.cosmos3.embodiments`; ``openarm`` requires
+post-training on OpenArm episodes via the ``cosmos3`` trainer - it is not a
+released zero-shot checkpoint).
 """
 
 from .action_decode import decode_pose_trajectory, denormalize_quantile, load_action_stats
--- a/strands_robots/policies/cosmos3/embodiments.py
+++ b/strands_robots/policies/cosmos3/embodiments.py
@@ -154,6 +154,44 @@ EMBODIMENTS: dict[str, Cosmos3Embodiment] = {
         raw_action_layout=_RAW_POSE9_GRASP,
         default_action_space="midtrain",
     ),
+    # Enactic OpenArm (7-DOF + gripper; registry ``openarm``, lerobot
+    # ``openarm_follower``). **Requires post-training** - no released Cosmos 3
+    # checkpoint ships this domain, so the entry promises no zero-shot
+    # behavior: it is the deploy-side mapping for a checkpoint post-trained on
+    # OpenArm episodes via the existing ``cosmos3`` trainer
+    # (:class:`strands_robots.training.cosmos3.Cosmos3Trainer`: SFT recipe TOML
+    # -> ``cosmos_framework`` launch). The unified action is the standard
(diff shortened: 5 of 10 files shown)


cagataycali (commented):
Reviewed this on a machine with the sim extras, so the parts CI skips actually ran. Almost all of it holds up under measurement - one gap, and it is documentation-only.

## What verifies

I ran the whole cosmos3 suite on `mujoco 3.10.0` (the lockfile pin) with `mink` and `robot_descriptions` present, `MUJOCO_GL=egl`:

```
tests/policies/cosmos3/  ->  243 passed in 2.41s
  test_sim_ik_openarm.py  ->  4 passed   (real MuJoCo, not skipped)
```

So the new sim-IK smoke path is not merely importable - it solves on the real `openarm_v1_mj_description` MJCF and lands inside the existing tracking bar.

Spot-checking the claims rather than taking them:

- **`midtrain` really is the raw layout.** `test_midtrain_layout_is_the_unified_action.py` derives its parametrization from `EMBODIMENTS.items()`, so it auto-enrolled the new entry: **21 -> 26 cases, 5 of them `[openarm]`, all passing**. That guard exists because a narrower `midtrain` entry does not shorten the served row, it shifts every name onto the wrong column - and the entry is clean on it.
- **The camera keys match what they claim to mirror.** `openarm_real` in `lerobot_local/embodiments.json` maps `front -> observation.images.image` and `wrist -> observation.images.wrist_image`; the entry declares the OpenPI-namespace form of exactly those two. Verified, not assumed.
- **The registry and driver-type claims hold.** `robots.json`'s `openarm` declares `robot_descriptions_module: openarm_v1_mj_description` (the module the new test loads), and `openarm_follower` is a real lerobot robot type (`lerobot/robots/openarm_follower/`), so that alias is not invented.
- **No `joint_pos` layout** is the right call, and pinning it is better than a comment.
- **`raw_action_dim=10` is consistent** with `_RAW_POSE9_GRASP` and with the three other 10-column domains.

## The gap: six stale enumerations

`docs/policies/cosmos3.md` and `README.md` enumerate the accepted embodiment set by hand, and this PR updates the code-internal copies (`__init__.py`, `policy.py`, `sim_ik.py`, the registry description) but not the two most user-facing ones. On this branch:

| Surface | Says |
|---|---|
| `docs/policies/cosmos3.md:2` front matter | `droid/umi/av/bridge` |
| `docs/policies/cosmos3.md:30` first example comment | `# droid \| umi \| av \| bridge` |
| `## Embodiments` table | 4 rows |
| domain / raw dim / bundled-stats table | 4 rows |
| the sentence above it | "the other **two** registered embodiments do not" - now three |
| `README.md:603` provider row | `` `droid`, `umi`, `av`, `bridge` `` |

Two of these are sharper than ordinary staleness:

1. **The docs now contradict this PR's own runtime message.** You correctly extended the shared-width sentence in `sim_ik.py` in both places - the docstring at :141 and the `ValueError` at :189 - to name four domains. `docs/policies/cosmos3.md:269` still carries the same sentence with three. A user reading the page learns three domains share a 10-column width; the refusal they hit names four.
2. **The domain table is the one a caller acts on.** It is what says whether `stats=` + `stats_domain=` are required, so a missing `openarm` row reads as "no stats needed" - the opposite of the post-training story the entry documents, and the one place where the omission could cost someone a silently mis-scaled decode.

Concretely:

```diff
-description: NVIDIA Cosmos 3 omnimodal VLA - WebSocket service, droid/umi/av/bridge embodiments, MuJoCo rollout.
+description: NVIDIA Cosmos 3 omnimodal VLA - WebSocket service, droid/umi/av/bridge/openarm embodiments, MuJoCo rollout.

-    embodiment="droid",          # droid | umi | av | bridge
+    embodiment="droid",          # droid | umi | av | bridge | openarm

 | `bridge` | Bridge dataset robots | - |
+| `openarm` | Enactic OpenArm (7-DOF + gripper) | `"openarm"` |

-ship them bundled; the other two registered embodiments do not:
+ship them bundled; the other three registered embodiments do not:

 | `av` | `av` | 9 | no |
+| `openarm` | `openarm_lerobot` | 10 | no |

-`bridge_orig_lerobot` are all 10 columns, so the width check cannot tell one
+`bridge_orig_lerobot` and `openarm_lerobot` are all 10 columns, so the width
+check cannot tell one
```

and in `README.md:603`, `` embodiments: `droid`, `umi`, `av`, `bridge`, `openarm` ``.

Worth saying plainly: **this is not really your miss.** Nothing graded any of those six surfaces. `tests/test_docs_policy_coverage.py` ties the overview table to `policies.json`, but it grades *providers*, so an embodiment added to an existing provider was graded by nothing - which is exactly why a PR this careful about truthfulness (the post-training framing, the deliberate absence of `joint_pos`, the by-name stats refusal) could still leave the page behind. I have opened #2496 to close that: it ties all six enumerations, plus the tables' stated `domain` / `raw dim` / bundled-stats facts, to `EMBODIMENTS`. Against this branch it reports precisely those six surfaces by name, so once it lands CI will answer this instead of a reviewer.

Also worth noting for whoever merges: the earlier red on this PR was infrastructure, not the diff. `Install system dependencies (OpenGL for MuJoCo)` was reaped at its 12-minute bound with `Run lint` and `Run tests` never reached - the download is ffmpeg's 116-package closure and the mirror was crawling (`libcodec2` alone took 4m28s for 9 MB). I have re-run it and the job is green through to the suite. #2493 cuts that download 30.5%.

Only the docs list is blocking, and it is mechanical - happy to approve as soon as those rows land (or if you would rather wait for #2496 and let CI enumerate them for you, that is fine by me too). Nice piece of work: the double-verification of the mapping (round-trip plus a real MuJoCo solve) and the refusal to fabricate a `joint_pos` layout are exactly right.

yinsong1986:
@cagataycali Thanks for the measured review - running the sim-IK path on real MuJoCo rather than taking the claims is exactly the verification this entry needed. Point by point:

**1. The verification section** (243-passing suite, `midtrain` auto-enrollment 21 -> 26 cases, camera-key mirror, registry/driver-type aliases, the deliberate absence of `joint_pos`, `raw_action_dim=10`) - nothing to change; glad the spot-checks held under measurement.

**2. The six stale enumerations** - all fixed in 06bb6b97f5b60d1348a3aade8f9743a9facd2718, exactly per your diff:

- `docs/policies/cosmos3.md` front matter: `droid/umi/av/bridge/openarm`
- the first example comment: `# droid | umi | av | bridge | openarm`
- `## Embodiments` table: `| openarm | Enactic OpenArm (7-DOF + gripper) | "openarm" |`
- the domain / raw dim / bundled-stats table: `| openarm | openarm_lerobot | 10 | no |` - agreed this was the sharpest of the six, since a missing row there reads as "no stats needed", the opposite of the post-training story
- "the other **two** registered embodiments" -> **three**
- the shared-width sentence now names `openarm_lerobot` as the fourth 10-column domain, so the page and the `ValueError` in `sim_ik.py` say the same thing again
- `README.md` provider row: `` `droid`, `umi`, `av`, `bridge`, `openarm` ``

Two adjacent fixes in the same class, called out here so they don't read as scope creep:

- **A seventh surface your list didn't name**: the README Cosmos 3 quickstart's rollout section also enumerates the set by hand (`Embodiments: droid (10D, chunk 32, 15 fps), umi, av, bridge.`). Now `..., bridge, openarm (post-training only).` Same class of gap, so it seemed wrong to fix six and leave the seventh for #2496 to report.
- **The table change falsified a neighboring sentence**: "`umi` ... `av` - exactly the two without bundled quantiles" is no longer "exactly the two" once `openarm` joins the no-bundled-stats set. Scoped it to "both without bundled quantiles", which stays true (the sentence is about the Cosmos3-Edge examples specifically, not an exhaustive enumeration).

**3. #2496** - thanks for filing and for tying the tables' `domain` / `raw dim` / bundled-stats facts to `EMBODIMENTS` rather than just the name lists; that's what would have caught the falsified neighboring sentence too. I fixed the rows now rather than waiting, so the new test should go green against this branch when it lands.

**4. The infra red + re-run and #2493** - appreciated; nothing for this branch to do.

Validation on the new head: `hatch run lint` clean, `hatch run test` 32617 passed / 270 skipped (`MUJOCO_GL=egl`).

cagataycali (approved):
Re-reviewed against the new head `06bb6b97`. All six enumerations I named are fixed, the seventh you found yourself is fixed, and the falsified neighbouring sentence is scoped correctly. Approving.

I want to record the verification rather than just say "looks good", because two of the checks are ones this branch could not have run for itself.

## The claim you made about #2496, verified three ways

You wrote that the new guard "should go green against this branch when it lands". I ran #2496's grader against three trees, since a guard that passes on the fixed tree proves nothing unless it also fails on the broken one:

| tree | #2496 grader |
|---|---|
| `upstream/main` `0acbfc93` (openarm not yet registered) | 27 passed |
| `2bf73fa2` (this PR's commit 1: openarm registered, docs stale) | **7 failed / 20 passed** |
| `06bb6b97` (this PR's commit 2: docs updated) | 27 passed |

The middle row is the important one. It names exactly seven surfaces: front matter, the `## Embodiments` table, the first example's inline comment, the domain / raw-dim / bundled-stats table, the README provider row, the README quickstart paragraph, and the bundled-count sentence (`sentence says 'two' (2) unbundled, registry has 3`). So the set really was seven, not six - your extra find was the README quickstart, and the count sentence is the one my original list did not separate out. All seven are green on the new head, and the grader is demonstrably not vacuous.

## Independent fact-check, not routed through that grader

Read straight off `EMBODIMENTS` plus the stats directory:

| embodiment | domain | raw dim | bundled stats | `midtrain == raw_action_layout` |
|---|---|---|---|---|
| av | `av` | 9 | no | yes |
| bridge | `bridge_orig_lerobot` | 10 | yes | yes |
| droid | `droid_lerobot` | 10 | yes | yes |
| openarm | `openarm_lerobot` | 10 | no | yes |
| umi | `umi` | 10 | no | yes |

- unbundled set is `['av', 'openarm', 'umi']`, count 3 - so "the other three registered embodiments do not" is now correct
- the domain row `| openarm | openarm_lerobot | 10 | no |` agrees with the entry on all three facts
- the neighbouring sentence now reads "both without bundled quantiles", which stays true
- `sim_ik.py` is the change I most wanted: the docstring **and** the `ValueError` both name `openarm_lerobot` as the fourth 10-column domain, so the page and the runtime message say the same thing again
- guard auto-enrollment holds: `tests/policies/cosmos3/` goes 256 -> 270 collected, `test_embodiments.py` 6 -> 11, with 10 parametrized `openarm` cases

## Real geometry on the pinned build

`mujoco 3.10.0`, which is what `uv.lock` resolves, with `mink 1.2.0` / `robot_descriptions 1.23.0`:

- `tests/policies/cosmos3/` - **270 passed**, including the 4 new OpenArm IK tests. Worth stating explicitly that they *ran* rather than skipped: `openarm.xml` is shipped next to `MJCF_PATH`, so the `pytest.skip` in the model fixture did not fire, and the geometry claim is real.
- drove `decode_cosmos_chunk_to_targets` on the real OpenArm MJCF (16-step chunk, declared stats, re-anchored): tracking **mean 0.233 mm, max 0.484 mm** against the test's 12.0 / 45.0 mm bars, EE travel 11.84 mm over the chunk, gripper commands inside the `[0, 0.044]` finger range, all joint targets finite. The bridge's own `tracking_error` and my independent recompute from `qpos` + `poses` agree to four decimals, which is a nice cross-check on the bridge's reporting.

Gate on this head: `ruff check` clean, `ruff format --check` 1368 files already formatted, `mypy strands_robots tests tests_integ` 19 errors on both this branch and `main` with the branch-only diff empty and none naming a changed file, `assemble_changelog.py --check` OK.

## One non-blocking note, not a change request

The model fixture's comment says the registry's `openarm` entry "resolves the single-arm `openarm.xml`". That is true of `assets.resolve_model_path("openarm")`, but not of the simulation path:

- `simulation.resolve_model("openarm")` -> `.../enactic_openarm/scene.xml`
- `sim.add_robot(name="openarm")` compiles `nq=18 njnt=18 nu=18` - the bimanual model, with `openarm/openarm_left_link7` and `openarm/openarm_right_link7` and **no** bare `openarm_link7`

So the geometry the smoke test validates (`nq=9`, `ee_frame_name="openarm_link7"`, a length-9 `q_init`) is not the geometry a reader gets from `add_robot("openarm")`. Someone copying the test as the OpenArm sim recipe lands on `InvalidFrame` - which does list the available bodies, so it is actionable rather than silent - and needs a length-18 seed and a left/right frame choice.

None of that is this PR's doing: `robots.json` is untouched here and the `openarm` robot entry predates the branch, so this is a comment-accuracy point rather than a defect. Either scoping the sentence to `resolve_model_path`, or naming the bimanual split so the reader knows which frame to pick, would close it. Fine as a follow-up.

Nice work on the seventh surface and on catching that the table change falsified the sentence above it - that second one is the failure mode the guard was written for, and you found it before the guard could.

PR #2493, opened 2026-08-19
ci: decline apt Recommends so the MuJoCo dependency step stops fetching 30% it cannot use

cagataycali:
## Problem

`Install system dependencies (OpenGL for MuJoCo)` is bounded `timeout-minutes: 12`, sized when 31
successful runs put it at p50 33s and max 455s (7m35s) — "clears the maximum by ~1.6x". That band no
longer holds. On 2026-08-19 it reaped four runs across three unrelated pull requests, same
message, nothing wrong in any diff:

| PR | job | reaped |
|---|---|---|
| #2488 | `96100298786` | 14:27:49Z |
| #2489 | `96107703838` | 15:07:23Z |
| #2488 (rerun) | `96119408569` | 15:32:50Z |
| #2499 | `96162011539` | 18:14:04Z |

#2489 is a `simulation/terrain.py` validation fix; #2488 is a SageMaker trainer provider. Neither
touches CI, MuJoCo or video. The `Get:` timestamps show the 12 minutes went to multi-minute **stalls**
landing after individual files, not to a uniform slowdown:

```
14:56:45  Get:109  mesa-common-dev      3133 kB
15:00:54  Get:110  mesa-va-drivers         6784 B   <- 4m09s gap
15:01:25  Get:115  libosmesa6           3555 kB
15:06:09  Get:116  libosmesa6-dev          9220 B   <- 4m44s gap
15:06:11  Get:117  pocketsphinx-en-us  27400 kB
15:07:23           step reaped, 72s into that fetch
```

The existing retry loop cannot help: it wraps `apt-get update`, not the `install`, and the step's own
comment says so — *"the 455s maximum above was `install` fetching ffmpeg's ~115 packages"*.

## Change

`--no-install-recommends` on both CI apt installs. Measured with `apt-get install --print-uris` on
`ubuntu-24.04`, summing the sizes apt reports:

| request | packages | download |
|---|---|---|
| `libosmesa6-dev ffmpeg` | 116 | 98.1 MiB |
| `libosmesa6-dev ffmpeg` (no recs) | **104** | **68.2 MiB** |
| `libosmesa6-dev` | 17 | 8.33 MiB |
| `libosmesa6-dev` (no recs) | 17 | 8.33 MiB |

**−29.9 MiB and −12 fetches, 30.5% of the transfer.** The 12 dropped packages are reachable only as
Recommends and none is load-bearing here:

- `i965-va-driver`, `intel-media-va-driver`, `mesa-va-drivers`, `mesa-vdpau-drivers`,
  `va-driver-all`, `vdpau-driver-all`, `libigdgmm12` — VA-API/VDPAU **hardware** video acceleration.
  No GPU on this runner; the point of `libosmesa6-dev` is that MuJoCo renders through OSMesa in software.
- `libaacs0`, `libbdplus0` — Blu-ray DRM.
- `libdecor-0-plugin-1-gtk`, `librsvg2-common` — Wayland decorations, SVG pixbuf loader.
- `pocketsphinx-en-us` — **27,433,580 bytes of speech-recognition acoustic models**, the largest single
  file in the set and the one being fetched when #2489 died. `ffmpeg → libavfilter9 → libpocketsphinx3`
  is a `Depends` chain; `libpocketsphinx3` then `Recommends: pocketsphinx-en-us`. It backs
  `libavfilter`'s ASR filter, which nothing here calls.

**What deliberately stays:** `ffmpeg` and every `libav*` library — they are `Depends`, they carry the
`h264`/`hevc`/`libsvtav1` software encoders `strands_robots/video.py` names in `SUPPORTED_CODECS`, and
`torchcodec` links the system FFmpeg *libraries* rather than the binary
(`tests/test_package_lazy_imports.py`). Dropping `ffmpeg` outright would cut 89.8 of the 98.1 MiB and
was **not** done: that is a behavioural change this measurement does not cover. (`libflite1`, 13.6 MB
of TTS voices, is a hard `Depends` of `libavfilter9` and is therefore *not* droppable this way.)

### Why the payload rather than the bound

#2457 measures the job's legal spend as already `29.1 + 12 + 4.4 = 45.5` min against a 45-minute job
bound, so every minute added to this step must be added to the job bound too. **`timeout-minutes: 12`
is unchanged** and this PR asserts nothing about it — `tests/test_workflow_jobs_are_bounded.py` owns that.

### Relationship to #2479

Complementary. #2479 raises the *job* bound 45 → 60 so a step spending its whole allowance cannot
cause an unnamed job-level reap; this reduces what the step spends. Different hunks of `test-lint.yml`
(#2479 lines 11–23, the job header; this the step at 38–95), so they compose textually. Either order.

### Why the flag is applied where it currently changes nothing

`agent-api-check.yml` installs `libosmesa6-dev` only — byte-identical either way, a measured no-op
today. Applied anyway, and the guard takes no exemption, because the no-op is a property of *that
argument list*, not of the workflow: the day a package with heavy Recommends is added there the 29.9
MiB arrives silently, and the reap it produces names a step rather than a dependency class.

## Live result — and a correction to the framing above

This PR's own run exercises the patched step. Under the **same degraded mirror** that produced the
three reaps, both steps passed:

| job | payload | apt step |
|---|---|---|
| `call-test-lint / Test and Lint` | 68.2 MiB | **487s** SUCCESS |
| `Detect AgentTool API Breaking Changes` | 8.33 MiB | **474s** SUCCESS |

Green under the bound with ~233s of margin. But those two numbers do not support the reading that
30.5% fewer bytes buys 30.5% less time, and I would rather state that than let it be inferred: the
8.33 MiB job took only 13s less than the 68.2 MiB one. Phase split from the completed
`agent-api-check` log:

```
15:55:55 -> 16:01:43   apt-get update, 105 metadata fetches   ~348s   (73%)
16:01:43 -> 16:03:45   .deb download, 17 packages / 8.33 MiB  ~122s   (26%)
```

So **under today's conditions the dominant cost of this step is `apt-get update`'s metadata refresh,
which `--no-install-recommends` does not touch at all.** The honest claim for this diff is therefore
narrower than "30% off the step":

1. It removes **12 fetches**, and the reap mechanism is a per-fetch stall (4m09s and 4m44s in the
   #2489 log), so it removes 12 stall opportunities — including the largest file in the set. That is a
   reduction in *stall surface*, which is probabilistic, not a throughput saving.
2. It removes 29.9 MiB of packages that are waste on a headless GPU-less runner on their own merits,
   and the guard stops them regrowing.

It is **not** sufficient on its own, and the third row of the incident table already showed that:
#2488's rerun reached only `Get:9` of 116 inside the full 12 minutes. No payload cut survives a mirror
that has stopped serving.

### The follow-up this measurement actually points at

Two levers, both outside this diff and both now with a number attached:

- **`apt-get update` is 73% of the step and fetches 105 metadata files**, most for third-party repos
  the runner image preconfigures and this job never installs from. `-o Acquire::Languages=none` (drops
  the `Translation-*` fetches) and scoping the sources are the lever on the dominant term.
- **Per-connection bounding**: `-o Acquire::http::Timeout` with `Acquire::Retries`, turning a stalled
  socket into a fast retry instead of a hang — the same move `timeout 180` already makes for a hung
  `apt-get update`.

Neither is in this diff on purpose. Their effect only shows against a degraded mirror, which I cannot
reproduce on demand, and every number above is one I can show; shipping an unmeasured mechanism beside
a measured one would make the measured one unfalsifiable.

## Verification

**Split against `upstream/main` `0acbfc93`**, new test file copied into the base tree:

|---|---|---|---|
| this branch | **9** | **0** | 9 |
| `upstream/main` + test file | 5 | 4 | 9 |

`5 + 4 == 9`, same collected total, no branch-only failures. The 4 base failures are behavioural and
each quotes the offending line:

```
test_the_install_declines_recommends[test-lint.yml:75]
  - test-lint.yml:75 installs without --no-install-recommends, so apt also
    fetches the Recommends closure (29.9 MiB / 12 packages for
    `libosmesa6-dev ffmpeg`, 26.2 MiB of it pocketsphinx-en-us speech models):
    sudo apt-get install -y libosmesa6-dev ffmpeg
test_the_install_declines_recommends[agent-api-check.yml:81]
test_the_comment_names_the_dominant_package
test_the_comment_names_the_dependency_class
```

The 5 passing on both trees are non-vacuity controls, each failing for a specific shortcut: call sites
are found at all, both known call sites are still covered, an install missing the flag is caught, a
commented-out install is not counted, and `apt-get update` is not treated as a call site (it takes no
package list, so Recommends cannot apply).

**Guard is repo-wide** over `.github/workflows/*.yml` rather than a hardcoded pair, so a third call
site added later is covered by construction.

**Adjacent guards, run to confirm no disturbance** — `test_workflow_jobs_are_bounded`,
`test_merge_gate_viewer_scope`, `test_pull_request_trigger_types`, `test_codeql_query_filters`,
`test_lockfile_parity_gate`, `test_ref_creation_ruleset_scope`, `test_dependabot_config_location`,
`test_changelog_format`: **112 passed**.

**Workflow integrity:** both files parse under `yaml.safe_load`; the rendered `run:` block passes
`bash -n`. `ruff check` / `ruff format --check` clean; `mypy` clean on the new file.

Parsing is line-based rather than via `yaml`, matching `test_workflow_jobs_are_bounded.py`:
`types-PyYAML` is not a dev dependency, and the command lives in a `run: |` block scalar a YAML parser
hands back as the same string anyway.

## Fourth reap, 18:14:04Z on #2499 — and it died inside the payload this diff drops

Recorded after the sections above were written, so it is a prediction tested rather than an example
chosen. #2499 is a `docs(mesh)` change: it touches no CI, no MuJoCo and no video, and its other twelve
contexts are green. `call-test-lint / Test and Lint` (`96162011539`) died with the same message, and
the fetch log names the package it was on:

```
18:12:58  Get:88  ffmpeg                     1879 kB
18:13:13  Get:89  libigdgmm12                 145 kB
18:13:14  Get:90  intel-media-va-driver      3163 kB
18:14:04          step reaped, 50s into that fetch
```

`intel-media-va-driver` and `libigdgmm12` are **both on the 12-package list this diff removes** — the
VA-API hardware-video group, on a runner with no GPU, for a job whose whole reason to want
`libosmesa6-dev` is that MuJoCo renders in software. The step got to `Get:90` of 116 and was reaped
inside a package it had no reason to fetch.

Consistent with the claim this PR actually makes, and I want to keep it to that: this removes a stall
opportunity, it does not prove this run would have passed. The reap landed 50s into one file, and
`apt-get update` is still 73% of the step's cost by the phase split above — `--no-install-recommends`
does not touch that term. What the incident does establish is frequency: four reaps in one day across
three pull requests whose diffs have nothing in common, which puts the follow-up levers
(`Acquire::Languages=none`, per-connection `Acquire::http::Timeout` + `Acquire::Retries`) on the
critical path rather than in the nice-to-have column.

The failed job was re-run rather than pushed to, so no approval was consumed and #2499's diff is
untouched.


yinsong1986 (approved):
## Summary

Adds `--no-install-recommends` to the two CI `apt-get install` call sites (`test-lint.yml`, `agent-api-check.yml`), cutting the MuJoCo dependency step's transfer by 29.9 MiB / 12 packages (30.5%) — all of it Recommends-only payload with no consumer on a headless GPU-less runner (VA-API/VDPAU drivers, Blu-ray DRM, GTK/SVG assets, and the 26.2 MiB `pocketsphinx-en-us` model set that #2489 died fetching). `ffmpeg` and every `libav*` library stay as `Depends`, preserving the `h264`/`hevc`/`libsvtav1` encoders `video.py` names and the libraries `torchcodec` links. The change is pinned by a repo-wide guard (`tests/test_apt_install_declines_recommends.py`) that covers any future third call site by construction, with non-vacuity controls and rationale-pinning tests; the split verification shows the 4 behavioural tests fail on base and pass on the branch. `timeout-minutes: 12` is deliberately unchanged, keeping the job's #2457 budget arithmetic intact and leaving the bound to `test_workflow_jobs_are_bounded.py`.

## What's good

- Changelog lands as a fragment (`changelog.d/2493-apt-declines-recommends.md`) per AGENTS.md step 3, not an `## [Unreleased]` append.
- The description corrects its own framing rather than overclaiming: the phase split shows `apt-get update` metadata is 73% of the step and untouched by this flag, so the claim is scoped to stall-surface reduction, with the follow-up levers (`Acquire::Languages=none`, per-connection timeouts) named but deliberately excluded as unmeasured.
- The flag is applied at the no-op call site (`agent-api-check.yml`) with the reason documented, so the guard takes no exemption and the invariant stays one line.
- Both patched steps ran green on this PR's own CI under the same degraded mirror that produced the reaps (487s / 474s, inside the 12-minute bound).

PR #2494, opened 2026-08-19
docs(examples/fleet): name the write path the read-only dashboard keeps

cagataycali:
## Problem

`restrict_to_subscribe_only` in `examples/fleet/dashboard.py` turns the dashboard's mesh peer into a subscribe-only surface, and both its docstrings describe that as replacing **"every command-capable method on the peer"**, enumerating `send` / `tell` / `broadcast` / `emergency_stop` / `publish_step`, plus confining raw `publish` to the peer's own namespace.

`publish_safety_event` is a sixth write method on `Mesh` and is in neither list. Driven on a real started peer, restricted by the shipped function:

| `Mesh` write method | verdict on the restricted peer |
| --- | --- |
| `send` | REFUSED |
| `tell` | REFUSED |
| `broadcast` | REFUSED |
| `emergency_stop` | REFUSED |
| `publish_step` | REFUSED |
| `publish` (another peer's topic) | REFUSED |
| `publish_safety_event` | **ACCEPTED** |

It publishes `strands/{peer_id}/safety/event` - inside the confined namespace, so the confinement permits it by design - and then calls `log_safety_event`, appending a **signed record to the audit trail this dashboard itself tails and renders** as its event timeline. One call on the restricted mesh put this on the forensic record:

```
event    = 'collision_imminent'
peer_id  = 'fleet-dashboard'
payload  = {'note': ..., 'severity': 'critical'}
```

That `publish_step` *is* refused is what pins the intended boundary. It also writes only the peer's own namespace (`strands/{peer_id}/stream`), so the namespace confinement alone would have permitted it, and it is refused anyway. The line the module draws is "a write that fabricates fleet state", not "another peer's topic" - and an audit-trail append is on the far side of it. A reader auditing "can this peer write?" is currently told it writes nothing but its own presence.

## Why the method cannot simply join the refusal list

The mesh's own receiver-side safety handlers call it to record *this* peer's lockout transitions:

| handler | `publish_safety_event` calls | outside a `try` |
| --- | --- | --- |
| `Mesh._on_safety_estop` | 5 | `remote_estop_engaged` |
| `Mesh._on_safety_resume` | 4 | `remote_resume_applied` |

The guarded calls narrow to `(TypeError, ValueError, OSError)` - none catches `RuntimeError`. So adding the method to the refusal list raises inside a Zenoh subscription callback on the safety path: a break, not a guard. The exemption is necessary, so this PR documents it instead of closing it. Scoping the write itself is a mesh-side contract question and is deliberately not decided here.

## Change

Documentation and tests only - **no behaviour change**, and no file under `strands_robots/` is touched.

- The module docstring and `restrict_to_subscribe_only`'s docstring now name the two write paths the restricted peer keeps (`publish` under its own namespace, and `publish_safety_event`), why the second one stays, and its bound: it can only ever write this peer's own `safety/event` topic and audit records carrying this peer's own `peer_id`, so it cannot speak for another peer.
- The `_COMMAND_METHODS` comment records why each of the five is safe to remove (their only internal callers are each other: `tell` delegates to `send`, `emergency_stop` to `broadcast`) and why the two write paths are absent.

## Tests

`tests/test_fleet_cross_zone_transport.py`, **2 failed / 26 passed** on `upstream/main`, 28 pass here.

Fails pre-fix:

- `test_every_public_mesh_method_is_refused_confined_or_documented` - every public `Mesh` method must be refused, confined, classified read-only/lifecycle, or named in `restrict_to_subscribe_only`'s docstring. Pre-fix it reports `['publish_safety_event'] can write to the mesh from a peer advertised as read-only, and are neither refused ... nor named ...`. `Mesh` grows independently of this example, so a method added there now fails this instead of silently landing on the read-only peer.
- `test_the_module_docstring_names_every_write_path_the_peer_keeps` - the read-only claim must name what it keeps, not only what it removes.

Passes on both trees, and each fails for a specific wrong fix:

- `test_the_safety_event_exemption_is_load_bearing_on_the_estop_path` - parses the two safety handlers and asserts at least one call is unguarded and no guarded one catches `RuntimeError`. This is the test that stops the naive "just add it to `_COMMAND_METHODS`" change, and it fails if the handlers ever guard every call - which is the point at which refusing the method becomes an option worth reconsidering.
- `test_the_exempt_safety_write_is_scoped_to_the_peers_own_id` - pins that the wire topic and the audit record are both derived from `self.peer_id`, so the exemption stays narrow enough to document rather than close.
- `test_the_read_only_classification_names_only_real_mesh_methods` - non-vacuity: a rename on `Mesh` must not leave a stale exemption behind.
- `test_every_command_method_refuses_on_the_restricted_peer` now iterates the module's own `_COMMAND_METHODS` instead of a copied literal, and the mesh fake gained `publish_safety_event` - without it the surface tests could pass by the method simply being absent from the fake.

## Verification

Measured at `c303578a` against `upstream/main` `0acbfc93`, on an 86-file selection (every guard that scans `examples/` or walks the tests tree, all fleet tests, the mesh safety suites, and the repo-wide docs/changelog/docstring guards):

- branch **2159 passed, 0 failed**; base **2 failed, 2157 passed**; 2213 collected on both, so `2157 + 2 == 2159`.
- branch-only failures: none. Base-only: exactly the 2 pre-fix failures above. No pre-existing failures in the selection.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on both trees with none in the changed files.

## Artifact

The same surface probe run against both trees. The measured column is byte-identical (this branch changes no code) and the audit record is the same; what changes is whether the docs account for the accepted row.

dashboard write surface vs documented surface


yinsong1986 (approved):
## Summary

Docs-and-tests-only PR that closes a documentation gap in `examples/fleet/dashboard.py`: `restrict_to_subscribe_only` advertised the peer as holding no command-capable method, but `publish_safety_event` (inherited from `SensorLoopsMixin`) remains callable on the restricted peer and appends signed records to the very audit trail the dashboard renders. The PR documents that exemption and its bound (peer-scoped topic and `peer_id`) rather than refusing the method, and the reason checks out against the mesh source on `main`: `_on_safety_estop` / `_on_safety_resume` make 5 and 4 `publish_safety_event` calls respectively, one unguarded in each handler, with the guarded ones catching only `(TypeError, ValueError, OSError)` — so a `RuntimeError`-raising refusal would break the Zenoh safety callback rather than guard anything. Reproduced the test claims exactly: the changed suite is 28 passed on this head and 2 failed / 26 passed with the base `dashboard.py` (the two new classification/docstring tests, as described), `ruff check` and `ruff format --check` clean on both changed files, no non-ASCII or host paths in the diff, changelog fragment present under `changelog.d/` per AGENTS.md step 3, and no file under `strands_robots/` is touched.

## What's good

- The classification test (`test_every_public_mesh_method_is_refused_confined_or_documented`) turns a one-off doc fix into a standing gate: a write method added to `Mesh` now fails a test instead of silently landing on the read-only peer.
- Each wrong-fix path is pinned: the AST test on the safety handlers fails exactly when refusing the method becomes viable, and the non-vacuity test keeps the exemption list from going stale on a `Mesh` rename.
- `test_every_command_method_refuses_on_the_restricted_peer` now iterates the module's own `_COMMAND_METHODS`, and the fake gained `publish_safety_event` so the surface tests cannot pass by absence.

PR #2496, opened 2026-08-19
test(policies/cosmos3): tie the docs embodiment enumeration to the registry

cagataycali:
## Problem

`create_policy("cosmos3", embodiment=...)` selects the conditioning domain, the
raw action width and the column layout, so that name is a public API surface in
its own right. The accepted set is enumerated **nine times by hand**, and nothing
tied any of them to
`strands_robots.policies.cosmos3.embodiments.EMBODIMENTS`.

Six are on the pages a reader consults to discover it:

| # | Surface | Form on `main` |
|---|---------|----------------|
| 1 | `docs/policies/cosmos3.md` front-matter `description:` | `... droid/umi/av/bridge embodiments ...` |
| 2 | `## Embodiments` table | 4 rows |
| 3 | first worked example's comment | `embodiment="droid",  # droid \| umi \| av \| bridge` |
| 4 | domain / raw dim / bundled-stats table | 4 rows |
| 5 | the sentence above table 4 | "Two domains ship them bundled; the other **two** registered embodiments do not" |
| 6 | `README.md` provider row | ``embodiments: `droid`, `umi`, `av`, `bridge` `` |

Three more are not on the provider page at all, which is why the first round of
this guard missed them:

| # | Surface | Form on `main` |
|---|---------|----------------|
| 7 | `README.md` quickstart rollout paragraph | ``Embodiments: `droid` (10D, chunk 32, 15 fps), `umi`, `av`, `bridge`.`` |
| 8 | `strands_robots.policies.cosmos3` docstring | `Available embodiments: droid, umi, av, bridge` |
| 9 | `Cosmos3Policy`'s `embodiment:` `Args:` entry | ``(``"droid"``, ``"umi"``, ``"av"``, ``"bridge"``)`` |

All nine are exhaustive claims, so all nine drift the same way.

The provider-level catalogue *is* guarded - `tests/test_docs_policy_coverage.py`
ties the overview table to `policies.json` for the same stated reason ("the docs
silently drift and users - and agent LLMs reading the docs - never discover the
provider"). But that guard grades **providers**, so an embodiment added to an
existing provider is graded by nothing, and the level below is where the drift
lives.

The failure is silent in the direction that matters: the code accepts the new
name, so no call breaks. Only the documentation disagrees, and a reader is told
the accepted set is smaller than it is. Surface 4 is the sharpest - it is what
tells a caller whether it must pass `stats=` and `stats_domain=` to
`decode_cosmos_chunk_to_targets`, so an omitted row reads as "no stats needed",
the opposite of the truth for an unbundled domain. Surfaces 5 and 7 state hard
numbers that go false the moment the registry moves.

## Change

One test module, `tests/policies/cosmos3/test_docs_embodiment_enumeration_matches_registry.py`.

This is the rule the sibling guard in the same directory already applies to the
same page's **keywords**:
`tests/policies/cosmos3/test_documented_backend_knob_routes.py` grades them
against `inspect.signature` "rather than against a copied list, so a knob
promoted onto the policy later, or a newly documented one, is graded without
touching this file". This applies that rule to the embodiment names.

It also grades the **facts** these surfaces state, rather than restating them:

- `domain` and `raw dim` are read from the entry.
- the `bundled stats` column is read from the shipped
  `strands_robots/policies/cosmos3/stats/<domain>_stats.json` - the file
  `load_action_stats()` resolves and refuses the domain without - so adding or
  removing a domain's quantiles cannot leave the column stale.
- surface 5's spelled count is compared against the computed bundled/unbundled
  split, and only *when that sentence shape is present*: rewording it to carry
  no count is allowed, stating a count the registry contradicts is not.
- surface 7's parenthetical width / chunk / fps are compared against the entry.
  A parenthetical carrying none of those patterns states no fact and is graded
  on nothing, which is what lets an annotation like `(post-training only)`
  coexist with the graded numbers.

Surfaces 8 and 9 are graded through `__doc__` rather than by reading source
text, so a reflow or a moved definition cannot disarm them and the graded string
is exactly what a reader is shown. Surface 7 is located by its paragraph's
opening word and collapsed before matching, so a re-wrap cannot either.

Both directions are graded on the tables: a registered embodiment missing from a
surface fails, and so does a documented embodiment the registry does not accept
(which would send a reader into a loud unknown-embodiment failure). On the two
prose surfaces (6 and 7) only the missing direction is graded, because they carry
backtick-quoted identifiers that are not embodiments (`ConnectionError`), so an
"extra" check there would report the prose rather than a drift.

**One nearby enumeration is deliberately not graded**, and the module docstring
says so: the README's "other embodiments such as `umi`/`av`/`bridge` need only
`observation/image`" sentence. "such as" disclaims exhaustiveness, so pinning it
would convert a hedge into a maintained list - and the claim it does make is
about `Cosmos3Embodiment.camera_keys`, not about the accepted set, so it belongs
to a camera-key guard rather than this one.

No source or documentation is changed. `main` already agrees with itself, so
this PR is the contract, not a correction.

## Verification

**The graders are not vacuous.** Every grader is a pure function of
`(surface text, registry)`, so eleven meta-tests hand them a registry carrying an
embodiment no surface can name and assert each one reports it. That is what makes
a guard which passes today load-bearing tomorrow. Four are adversarial in the
other direction: one rewrites the `av` row's stated width to `7`, one rewrites
the quickstart's `chunk 32` to `chunk 31`, and one drops `av` from the registry
and requires both table graders to report it as documented-but-unregistered.

Writing those meta-tests immediately paid for itself: they caught that surface
5's sentence **wraps across a line break** in the page, so the first regex never
matched and that check was passing vacuously. The graders now collapse
whitespace before matching, and a future re-wrap cannot disarm them.

**Premise tests.** Tests assert each graded table, enumeration, paragraph and
docstring anchor is still found, and that the stats directory is readable, so a
reformat or a moved page fails loudly rather than making the graders report a
clean set. A further test asserts a missing surface raises rather than grading
the empty set.

**It catches the live case, and composes with it.** Measured against the open
OpenArm embodiment branch (#2492), which registers a fifth embodiment:

| tree | result |
|------|--------|
| #2492's registry change, every documentation surface reverted to `main` | **9 failed, 18 passed** - all nine surfaces named |
| #2492 unmodified (its own hand-fixes in place) | **27 passed** |
| this branch on `main` | **27 passed** |

The first row is the guard doing its job; the second is the guard agreeing with a
branch that fixed the same nine surfaces by hand. Each failure names the surface
and the remedy, so CI answers the question instead of a review round-trip:

```
the README cosmos3 quickstart's 'Embodiments: ...' paragraph omits ['openarm'].
  It sits directly under the runnable rollout command, so it is the enumeration
  a reader who never opens the provider page relies on.
the 'Available embodiments:' sentence in strands_robots.policies.cosmos3's
  docstring omits ['openarm']. It is one import from the registry and is what
  help() on the package prints.
Cosmos3Policy's 'embodiment:' Args entry omits ['openarm']. That parenthetical
  is the accepted-value list for the parameter, so a caller reading it is told
  the registry accepts less than it does.
```

**Gate.** `ruff check` and `ruff format --check` clean over
`strands_robots tests tests_integ` (1371 files). No `mypy` error in the changed
file. `tests/policies/cosmos3/` is **223 passed** on this branch against **196
passed** on pristine `main`, with a byte-identical 3-failure set in both columns
(pre-existing `test_policy_diffusers.py` `bfloat16` failures from this host's
torch mock, present on `main` too). `223 - 196 = 27`, exactly the new test items,
so no existing test changed verdict.

## Scope

Deliberately narrow. This does not touch the page, the README or any source: it
records the enumerations `main` already gets right so the next embodiment cannot
quietly shrink them. Extending the same treatment to other providers is a
follow-up once this shape has settled.

## Round 2 changes

Review feedback on #2492 named a **seventh** hand-maintained enumeration this
guard did not cover - the README quickstart's `Embodiments: ...` paragraph.
Sweeping for the class rather than for that one line found two more (surfaces 8
and 9 above), so the count in this description and in the module docstring was
wrong, not merely incomplete.

- Added graders and tests for surfaces 7, 8 and 9, with a vacuity meta-test each,
  plus a fact grader for the quickstart's stated width / chunk / fps and an
  adversarial test for it. 19 tests -> 27.
- Surfaces 8 and 9 are read from `__doc__` rather than from source text, so the
  graded string is what a reader is shown and a reflow cannot disarm the guard.
- Recorded the one enumeration deliberately left ungraded (the "such as" camera-key
  sentence) and why, so a future reader does not read the omission as an oversight.
- Corrected the "six times by hand" premise to nine in the module docstring, the
  changelog fragment and this description.

Fixing six of nine would have left the same silent failure reachable by two paths
and left this file's own stated premise false.


yinsong1986 (approved):
## Summary

Test-only PR adding `tests/policies/cosmos3/test_docs_embodiment_enumeration_matches_registry.py` (plus a changelog fragment), which ties all nine hand-maintained enumerations of the cosmos3 embodiment set - six on `docs/policies/cosmos3.md` / the README provider row, plus the README quickstart paragraph and the two docstring surfaces read through `__doc__` - to `strands_robots.policies.cosmos3.embodiments.EMBODIMENTS`. Per-embodiment facts (domain, raw dim, bundled-stats presence via the shipped `stats/<domain>_stats.json`, and the quickstart's width/chunk/fps parentheticals) are read from the entry and the stats directory rather than restated. No source or documentation is changed; `main` already agrees with itself, so the module is a contract against future drift.

## What's good

- The graders are pure functions of `(surface text, registry)`, and the vacuity meta-tests plant an embodiment no surface names and assert every grader reports it - including adversarial rewrites (`raw dim 9 -> 7`, `chunk 32 -> 31`) and the documented-but-unregistered direction. This is what keeps a guard that passes today load-bearing tomorrow.
- Premise tests make a moved page, renamed table header or reworded anchor sentence fail loudly instead of grading an empty set (`test_a_missing_surface_is_reported_rather_than_skipped`), and the whitespace-collapse fix for the wrapped count sentence is pinned.
- Follows the sibling guard's stated rule (`test_documented_backend_knob_routes.py`): grade against the source of truth rather than a copied list, so a new embodiment is graded without touching this file.
- The one deliberately ungraded enumeration (the "such as" camera-key sentence) is recorded in the module docstring with the reasoning, so the omission cannot be read as an oversight.
- Changelog fragment per AGENTS.md step 3; no host paths, no non-ASCII; verified 27/27 pass on the head SHA.

PR #2497, opened 2026-08-19
refactor(examples/libero): merge run_mujoco.py and run_isaac.py into run.py with backend subcommands

### LINKED ISSUE #2460: refactor(examples/libero): merge run_mujoco.py and run_isaac.py into run.py with backend subcommands
# Motivation

The LIBERO drivers promise "switch backend", but the backend switch is a different script. Measured on current `main`:

| | `run_mujoco.py` (515 LOC) | `run_isaac.py` (989 LOC) |
|---|---|---|
| Core driver | `load_libero_suite` -> `create_simulation("mujoco")` -> `add_robot` -> `evaluate_benchmark(...)` | Identical shape - its own docstring says "same `evaluate_benchmark(...)` driver" |
| Shared helpers | `_date_dir`, `_suite_for_task`, `_default_checkpoint_dir`, `_explain_lifecycle_failure`, `_configure_gr00t_image` | Copy-pasted; diffs are docstring-only, and the Isaac copy of `_suite_for_task` literally says "Identical to `run_mujoco._suite_for_task`" |
| Genuinely backend-specific | ~30 lines (`add_robot(data_config="panda")` + load-race workaround) | ~350 lines (CDN Franka-USD resolution, `--robot-usd`/`--robot-urdf`, adapter kwargs, exit-code epilogue for SimulationApp's exception-swallowing fast-shutdown) |
| Asymmetric feature | - | `_orchestrate_groot_server` exists only here; the MuJoCo path makes the user start the server manually for no backend reason |

Five identical functions in two files is the two-copies signal this repo treats as the evidence for unification, and the drift has already begun (docstrings diverged).

# Proposal

`examples/libero/run.py` with backend subcommands:

- **Subcommands for scoped flags**: `run.py mujoco ...` / `run.py isaac ...` - Isaac-only flags simply do not exist on the mujoco subcommand.
- **One shared core**: parser base, the five helpers, task resolution, GR00T orchestration (the MuJoCo path gains `_orchestrate_groot_server` in the merge), episode loop, report format.
- **Two thin setup shims**: `setup_mujoco(args) -> sim` (~30 lines) and `setup_isaac(args) -> sim` (~120 lines, CDN probe + warmup); the Isaac exit-code epilogue wraps `main` only for that subcommand.
- **Import hygiene**: import the chosen backend only after parsing. Already safe - the library defers `SimulationApp` boot into `create_world`, and one backend per process invocation sidesteps the both-backends-in-one-process concern that made `libero_backend_matrix.py` subprocess-based.

# Constraints the PR must respect

1. **`libero_backend_matrix.py` subprocess-parses driver stdout** - the output-line format is a de-facto contract (both scripts already emit the same lines). Update the matrix to invoke `run.py <backend>` in the same PR and pin the line format with a test.
2. **Old filenames are referenced** (README, matrix, prior issues): delete and update all references atomically per the no-dead-code rule, or keep 2-line forwarding shims for one release - decide in review, but do not leave two diverging copies.

# Plan

- PR 1: merge `run_mujoco.py` + `run_isaac.py` -> `run.py`; update matrix + README; pin output format.
- PR 2 (follow-up): same treatment for `run_mujoco_agent.py` / `run_isaac_agent.py`.

# Related

- #2459 (LIBERO meshes on Isaac - adjacent LIBERO/Isaac work; the merged driver is where its interim cross-backend caveat prints)
- #2122 / #2123 (the Isaac parity landings that shrank the honest differences between these scripts)

yinsong1986:
Closes #2460

## What changed

`examples/libero/run_mujoco.py` (515 LOC) and `run_isaac.py` (989 LOC) are merged into one driver, `examples/libero/run.py`, with backend subcommands:

```bash
python examples/libero/run.py mujoco --policy mock --n-episodes 5
python examples/libero/run.py isaac  --policy groot --port 8000 --n-episodes 50
```

**One shared core** (per the issue's proposal):
- parser base (shared flags on an argparse parent parser)
- the five previously copy-pasted helpers: `_date_dir`, `_suite_for_task`, `_default_checkpoint_dir`, `_explain_lifecycle_failure`, `_configure_gr00t_image`
- task resolution (`_resolve_task`, with the Isaac `adapter_kwargs` hook)
- GR00T orchestration: **the mujoco path gains `_orchestrate_groot_server`** (and `HF_TOKEN` / `HUGGING_FACE_HUB_TOKEN` env-var support via `_resolve_hf_token`, previously Isaac-only); **the isaac path gains `--deterministic`** (previously mujoco-only) since the flag configures the backend-agnostic server container
- the eval-and-report loop (`_evaluate_and_report`) and a single result-line formatter (`_format_result_lines`)

**Two thin setup shims**: `_setup_mujoco` (procedural Panda + LIBERO scene pre-warm) and `_setup_isaac` (CDN Franka-USD resolution, explicit RTX cameras + warm-up, synchronous `on_frame` recorder with the retry wrapper and the zero-frame fail-loud check). Each returns `(sim, plan)` where the plan carries the per-backend recording arm/check hooks; each destroys the sim on partial setup failure before re-raising.

**Scoped flags**: `--robot-usd` / `--robot-urdf` / `--eef-body-name` exist only on the `isaac` subcommand - the `mujoco` subcommand rejects them at parse time.

**Import hygiene**: the chosen backend is imported only after parsing (pinned by a test that asserts no `mujoco`/`omni`/`isaacsim` module lands in `sys.modules` on import of `run.py`).

**Exit-code epilogue**: the `os._exit(1)` guard against SimulationApp's exception-swallowing fast-shutdown wraps `main` only for the `isaac` subcommand, as proposed.

## Issue constraints

1. **Matrix + line-format pin (same PR)**: `libero_backend_matrix.py` rows now invoke `run.py <backend>` (the `isaac-4096` row still points at the future `run_isaac_fleet.py` and keeps printing `unavailable`). The output-line contract is pinned in `tests/test_examples_libero_drivers.py` **against the driver's own formatter** (`_format_result_lines` x matrix `_RE_RESULT`), for both backends, so either side drifting goes red.
2. **Old filenames**: deleted with every reference updated atomically (READMEs, `docs/simulation/isaac.md`, `examples/isaac_gs/scene.py`, `examples/mujoco_gs/README.md`, library docstrings in `benchmarks/libero/{suite,adapter}.py`, test/tests_integ docstrings, and the sibling agent drivers' cross-references). I chose **deletion over forwarding shims** per the repo's no-dead-code rule (rule 10) - a 2-line shim has no in-tree caller the moment the matrix moves. Happy to add grace-period shims instead if review prefers; `CHANGELOG.md` / `changelog.d` historical entries were left untouched as records of their time.

### Output-line note (reviewable behavior change)

Both subcommands now emit the unified line, i.e. the mujoco line gains `resolved_task=` and `backend=mujoco` (previously Isaac-only fields). The matrix parser already tolerated extra fields between anchors (pinned pre-existing behavior), and `task=` now uniformly echoes the CLI-requested task for replayability (Isaac's semantics; previously the mujoco driver echoed the resolved fallback). Recording filenames gain `--backend=<backend>` on mujoco, still matching the matrix discovery glob.

## Test surface

- `hatch run lint`: clean (ruff + ruff format + mypy), plus `ruff check`/`format --check` over `examples/libero/`.
- `hatch run test`: **32602 passed, 269 skipped, 0 failed** (95.6% coverage) with `MUJOCO_GL=egl`.
- `tests/test_examples_libero_drivers.py` rewritten: 11 tests - import-cleanliness without any backend, helper surface, result-line format pinned for both backends via the driver's formatter, matrix rows point at `run.py` subcommands, Isaac-only flag scoping (rejected on `mujoco`, accepted on `isaac`, shared base identical), private-API hygiene, library-surface pins (now also pinning `load_libero_suite(adapter_kwargs=...)`).
- GPU verification on this branch (4x L4 box):
  - `run.py mujoco --policy mock --n-episodes 2` -> `success_rate=0.00  wall_time=31.1s`, MP4 written.
  - `run.py isaac --policy mock --n-episodes 1` -> RTX cameras warmed, **720 frames** recorded, `success_rate=0.00  wall_time=40.9s  ... backend=isaac`, clean SimulationApp shutdown.
  - `libero_backend_matrix.py --backends mujoco,isaac-4096` -> mujoco row `ok` with parsed cells; fleet row `unavailable (no run_isaac_fleet.py)`.

## Out of scope (per the issue's plan)

- PR 2 follow-up: same treatment for `run_mujoco_agent.py` / `run_isaac_agent.py` (their cross-references to the merged driver are already updated here).

## Board

Please move #2460 to "In review" on the project board (filing account lacks org project write).


examples/libero/run.py
```diff
@@ -0,0 +1,1223 @@
...
+        "Franka USD). Default 'panda_hand' resolves on the bundled Franka; "
+        "override when --robot-usd / --robot-urdf loads an asset with "
+        "different link names.",
+    )
+    return parser
+
+
+def main(args: argparse.Namespace) -> None:
+    if args.container is None:
+        args.container = f"gr00t-libero-{args.backend}"
```
cagataycali, line 1166: [MUST FIX] This is the only behavioral mechanism the refactor changed, and it is the only one no test pins.

`--container`'s default moved from two hardcoded strings to one derived f-string. That is the right call - but `tests/test_examples_libero_drivers.py` contains **zero** occurrences of `--container` and **zero** of `gr00t-libero-`, so nothing in the suite fails if this f-string or the subparser's `dest="backend"` drifts.

The failure mode is quiet and remote from the cause: both backends would resolve to the same container name, which is precisely what this flag's own help text promises cannot happen ("so they don't clobber each other's containers when run side-by-side on the same host"). It would surface as a confusing GR00T lifecycle collision on a GPU host - a machine CI does not have - rather than as a red test.

Every other flag is covered by the surface parity above, because those are unchanged literal strings; a comparison against the old files is a sufficient pin for them. This one is *computed* from the subcommand, so the parity argument does not reach it, and it is the single line where a rename anywhere else in the file changes user-visible behaviour.

Cheap pin - parametrize `("mujoco", "gr00t-libero-mujoco")` and `("isaac", "gr00t-libero-isaac")`, `parse_args([backend])` through whatever entry point applies this resolution, and assert `args.container == expected`; plus one case asserting an explicit `--container my-name` survives resolution unchanged (the `is None` guard, which is the half a naive `or` rewrite would break).

That test also documents the equivalence claim the changelog fragment makes, so a future reader can see the two old defaults were preserved deliberately rather than by accident.

cagataycali (changes requested):
The consolidation itself verifies clean. I checked the property that a merge-two-drivers-into-one refactor can most easily break silently - the CLI surface - and it holds exactly.

**No flag was dropped or invented.** Parsed the argparse surface of all three files with `ast`, at the merge base `e757aca3` and at head `16f9d924`:

| file | arguments |
|---|---|
| `run_mujoco.py` (deleted) | 11 |
| `run_isaac.py` (deleted) | 13 |
| `run.py` (new) | 14 |

old union minus new: **0**. new minus old union: **0**. The 14 is the union of 11 and 13 with the shared flags counted once, so the consolidation is exactly a merge, with nothing smuggled in or quietly retired - which is what a refactor changelog entry should be able to claim.

**One default changed shape, and it is equivalent.** `--container` was hardcoded per file (`gr00t-libero-mujoco` / `gr00t-libero-isaac`); it is now `default=None`, resolved at `run.py:1152-1153` as `f"gr00t-libero-{args.backend}"`. With `dest="backend"` on the required subparser that reproduces both old values exactly, and derives them instead of repeating them - strictly better, and it keeps the help text's "so they don't clobber each other's containers when run side-by-side" true. See the inline note; this is the one thing I want pinned.

**No stale path references survive** to `run_mujoco.py` / `run_isaac.py` anywhere in `strands_robots/`, `docs/`, `examples/`, `tests/` or `tests_integ/`. Test surface is real rather than nominal: 8 tests, both subcommands exercised, backend-matrix rows asserted to invoke the subcommands, and isaac-only flags asserted **absent** from the mujoco subcommand.

Two items, both cheap, one commit.

## [MUST FIX] The retarget sweep stopped short of `changelog.d/`, whose fragments assemble verbatim

This PR correctly retargets prose in `docs/simulation/isaac.md`, `examples/README.md`, `examples/libero/README.md` and `examples/mujoco_gs/README.md`. Five **unreleased** fragments still name the deleted files:

- `changelog.d/1790-gr00t-deterministic-lifecycle-flag.md:19` - "`run_mujoco.py --policy groot` grows a `--deterministic` passthrough"
- `changelog.d/1798-isaac-pip-6x-deferred-physics-warmup.md:16` - "(e.g. `run_isaac.py`'s `resolved_task=`)"
- `changelog.d/1802-isaac-get-body-state-eef-source.md:24` - "`run_isaac.py` plumbs the measured Isaac Franka calibration"
- `changelog.d/1812-isaac-delta-eef-action-controller.md:36` - "`run_isaac.py` passes `control_frequency=20.0`"
- `changelog.d/1820-isaac-load-scene-frozen-timeline.md:39` - "Verified end-to-end on Isaac Sim 6.0 / L4: `run_isaac.py --policy mock --n-episodes 2`"

Why this is blocking rather than a nit: `scripts/assemble_changelog.py:178` renders fragments as `ENTRY_SEPARATOR.join(fragment.body.strip("\n") ...)` - **verbatim**, with no rewriting anywhere in the path. So at the next release these five sentences become release notes that name paths the same release deleted. All five are present tense or imperative about the shipped tree, not narration of a past state, and the last one is a reproduce command a reader cannot run.

It is also now-or-never. Nobody re-reads 688 pending fragments at assembly time, and once assembled the text is frozen in `CHANGELOG.md` under the convention that released entries are immutable - so the window to make these true closes when this PR merges, and the cost afterwards is permanent. Five path edits (`run.py mujoco` / `run.py isaac`) close it.

Explicitly **leave `CHANGELOG.md:1167,1170` alone.** Those are released history and correctly describe what `run_mujoco.py` was when it shipped; rewriting them would be the opposite error. I am flagging the boundary because a grep-and-replace over the whole tree would cross it.

Two further hits are narration of the bug being fixed (`1802:3` "`examples/libero/run_isaac.py --policy groot` failed every inference call", and `1812:3`), where the old name was accurate at the time of the failure being described. Your call - I would retarget them for greppability, but I am not asking for it.

examples/libero/run.py
yinsong1986: Pinned in 1f6ede54. The resolution is extracted into `_resolve_container_name()` (called by `main`, semantics unchanged) so the test drives the real code path rather than restating the f-string. `tests/test_examples_libero_drivers.py` now parametrizes `("mujoco", "gr00t-libero-mujoco")` / `("isaac", "gr00t-libero-isaac")` through `_build_parser().parse_args([backend])` - so drift in either the f-string or the subparser's `dest="backend"` goes red - and a third case asserts an explicit `--container my-name` survives resolution verbatim, pinning the `is None` guard a naive `or` rewrite would break. The changelog fragment now also records that the two pre-merge hardcoded defaults were preserved deliberately. `hatch run lint` clean; the new tests plus the full drivers file pass (14 passed).

yinsong1986:
@cagataycali Point-by-point on the review body:

**[MUST FIX] changelog.d retarget** — fixed in de4a49bb. All five unreleased fragments now spell the subcommand form:

- `1790-gr00t-deterministic-lifecycle-flag.md` -> `run.py mujoco --policy groot`
- `1798-isaac-pip-6x-deferred-physics-warmup.md` -> `run.py isaac`'s `resolved_task=`
- `1802-isaac-get-body-state-eef-source.md` -> `run.py isaac` plumbs the calibration
- `1812-isaac-delta-eef-action-controller.md` -> `run.py isaac` passes `control_frequency=20.0`
- `1820-isaac-load-scene-frozen-timeline.md` -> reproduce command is now `run.py isaac --policy mock --n-episodes 2`

Verified via `scripts/assemble_changelog.py --check` (687 pending fragments OK) and `--print`: the only `run_mujoco.py`/`run_isaac.py` spellings surviving in assembled output are in this PR's own fragment (`2460-libero-run-backend-subcommands.md`), whose subject *is* the deleted files — left as-is deliberately, same reasoning as the boundary you flagged.

**Boundary respected** — `CHANGELOG.md:1167,1170` untouched; released history stays describing what `run_mujoco.py` was when it shipped. The sweep was scoped to `changelog.d/`, not a tree-wide replace.

**The two optional narration hits (`1802:3`, `1812:3`)** — took your stated preference and retargeted both for greppability, in the same commit. The subcommand invocation is argument-for-argument what the narrated failure ran, so nothing about the described behavior changes; if you'd rather keep the historical spelling there, it's a two-line revert.

**The `--container` default pin (your inline note)** — addressed earlier in 1f6ede54: the resolution is extracted into `_resolve_container_name()` and `tests/test_examples_libero_drivers.py` parametrizes both backends against the exact old hardcoded values, plus asserts an explicit `--container` wins. Details in the inline thread.

Validation: `hatch run lint` clean; `tests/test_examples_libero_drivers.py` (40 passed) and the three changelog gates (63 passed) green; fragment assembly `--check` passes. The full `hatch run test` sweep aborts in an unrelated MuJoCo rendering test on this box (GL environment) — that test passes in isolation on the unmodified tree too, so it is environmental, not this diff (which is markdown-only under `changelog.d/`).

cagataycali (approved):
Approving - the MUST FIX is addressed, and better than I asked: extracting `_resolve_container_name()` means the test drives the real resolution rather than restating the f-string, and asserting the parse-time `args.container is None` sentinel documents why the default cannot just be `default=f"..."`.

I ran the break table rather than reading it, on `de4a49bb`, to check the pin is load-bearing:

| break | new test fires |
| --- | --- |
| `args.container = f"gr00t-libero-{args.backend}"` -> a constant `"gr00t-libero-mujoco"` | YES - `[isaac-gr00t-libero-isaac]`, `assert 'gr00t-libero-mujoco' == 'gr00t-libero-isaac'` |
| `dest="backend"` renamed (derived name follows it) | YES - via `test_isaac_only_flags_do_not_exist_on_the_mujoco_subcommand` |
| `if args.container is None:` -> `if not args.container:` | no - 14 passed |
| control | 0 of 14 |

So the collision the flag's help text promises cannot happen is genuinely pinned now, which was the whole point. Good to merge.

One residual, non-blocking, purely because the test's own docstring makes a claim one line stronger than the test checks:

> The resolver guards on `is None`, not truthiness - the half a naive `args.container or f"..."` rewrite would break.

A truthiness rewrite passes all 14. `"my-name"` is truthy, so the two guards agree on it; measured, they differ for exactly one input:

| `--container` | `is None` (shipped) | truthiness | differ |
| --- | --- | --- | --- |
| omitted | `'gr00t-libero-mujoco'` | `'gr00t-libero-mujoco'` | no |
| `'my-name'` | `'my-name'` | `'my-name'` | no |
| `''` | `''` | `'gr00t-libero-mujoco'` | YES |
| `' '` | `' '` | `' '` | no |

Worth one extra `pytest.param("")` asserting `args.container == ""` if you want the docstring's claim to be true, because the shipped guard is the right one and the reason is downstream: `container_name` reaches `gr00t_inference` unvalidated and lands in a `docker exec <container_name> ...` argv (unlike `image_name`, which has an allowlist). Preserving `""` surfaces the operator's mistake as a docker error naming the empty string; substituting the derived default would silently run against a container the caller did not name, which is the same shape of quiet substitution the `is None` sentinel exists to avoid. Either add the case or soften the docstring - I would not hold the PR for it, since the mechanism the MUST FIX was about is covered.

Issue #1802, opened 2026-07-31
feat(sim/isaac): EEF state source for LiberoAdapter + IsaacSimulation.get_body_state - GR00T-on-Isaac blocked on 'State key state.x must be in observation'


yinsong1986:
## Summary

With the #1798 fixes in place, `examples/libero/run_isaac.py --policy groot` gets past scene-load and the `wrist_image` camera, then fails on every inference call with:

```
RuntimeError: Server error: State key 'state.x' must be in observation
```

Root cause: `LiberoAdapter.augment_observation` - the hook that injects the `state.x/y/z/roll/pitch/yaw/gripper` EEF pose the `libero_panda` GR00T data-config requires - has **no working pose source on the Isaac backend**. Both of its paths are MuJoCo-shaped (verified on `main` @ HEAD, 2026-07-31, live on a 4x L4 host with pip `isaacsim` 6.0.0.1):

1. **Direct read** (`_read_eef_pose`, `strands_robots/benchmarks/libero/adapter.py:1619+`): reads `sim._world._model` / `._data` (`data.site_xpos[gripper0_grip_site]` for position, `data.xquat[robot0_right_hand]` for orientation - the #168 RoboSuite split). On Isaac, `sim._world` is the Isaac `World` handle with no `_model`/`_data`, so both lookups fail.
2. **Fallback** (`adapter.py:~1726`): `getattr(sim, "get_body_state", None)` - **`IsaacSimulation` does not implement `get_body_state` at all** (it exists only on the MuJoCo backend, `simulation/mujoco/physics.py:851`), so the adapter logs "sim has no get_body_state(); skipping EEF state injection" and injects nothing. The GR00T service then correctly rejects the observation.

## Wider impact than LIBERO

`get_body_state` is also the read primitive under the predicate DSL (`simulation/predicates.py` `_body_position`/`_body_quaternion`), so **declarative benchmarks using body predicates (`body_inside`, `body_on`, `distance_less_than`, ...) cannot evaluate on Isaac either**. Implementing `IsaacSimulation.get_body_state` fixes both consumers with one surface.

## Proposed fix (two parts)

**Part 1 - `IsaacSimulation.get_body_state(body_name)` (the fallback path, and the predicate DSL).**
[...]

PR #2498, opened 2026-08-19
feat(sim/isaac): LIBERO objects render as real meshes instead of box proxies

### LINKED ISSUE #2459: feat(sim/isaac): LIBERO objects as real meshes - implement add_object mesh_path + SceneObject mesh passthrough (box proxies invalidate pixel-policy cross-backend evals)

yinsong1986:
Closes #2459

## What changed

On the Isaac backend, LIBERO task objects were realized as gray box proxies -- for the mesh-typed objects LIBERO actually uses (bowls, plates), not even fitted boxes (`_extract_mjcf_shape` degraded a mesh geom to a hardcoded 0.05 m box). With #1802 closed, a GR00T-LIBERO checkpoint trained on MuJoCo-rendered visuals can now run on Isaac, so this became an eval-integrity problem: a cross-backend `success_rate` comparison for a pixel-conditioned policy measured the visual domain gap as much as the policy.

This PR implements all four deliverables from the issue:

1. **`add_object(shape="mesh", mesh_path=...)` on Isaac** -- the previous explicit refusal is replaced by an implementation: the asset is converted to USD once (content-addressed cache under `$STRANDS_BASE_DIR/asset_cache/usd_meshes/`, same pattern as the LIBERO scene cache) and referenced onto the stage with convex-hull collision. MuJoCo's mesh contract is matched: `size` is ignored (the asset defines the extent, reported back from the compiled geometry), collision is the convex hull.
2. **`SceneObject` + `load_mjcf_scene_objects` carry the mesh asset** -- new `mesh_path` / `mesh_scale` / `mesh_pos` / `mesh_quat` fields, resolved from the MJCF `<asset>` registry (with `meshdir` handling). `load_scene` renders the mesh as the **visual** and keeps the collision-AABB box as the **invisible physics proxy**, so physics behavior is unchanged. A mesh-only body gets the mesh's computed bounds as its proxy instead of the 0.05 m fallback.
3. **Asset resolution + conversion** -- new `strands_robots/simulation/isaac/mesh_assets.py`: pure-stdlib OBJ/STL/**MSH** parsing, mesh AABB, and USD authoring via `usd-core` (no Kit session needed). **MSH support turned out to be load-bearing, not a nicety**: LIBERO's robosuite-compiled scenes declare the bowl/plate *visual* meshes as legacy MuJoCo binary `.msh` (88 such files in the LIBERO asset tree), so without it the exact objects the issue names would have stayed box proxies.
4. **Interim caveat** -- while any object in a loaded scene still renders as a proxy, the `load_scene` report (text + json `visual_caveat`) states that pixel-conditioned policy scores on that scene are not cross-backend comparable. The sentence disappears when every object carries its mesh.

Fail-loud contract preserved: a mesh asset that is declared in `<asset>` but missing on disk raises `ValueError` at parse time -- never a silent box. An unconvertible format (e.g. COLLADA `.dae`) keeps the visible proxy and is named in the report.

## Verification against real LIBERO scenes

Ran `load_mjcf_scene_objects` over all 7 robosuite-compiled LIBERO scenes in the local scene cache: **every bowl / plate / mug / ramekin / cabinet / stove resolves its real visual mesh** (e.g. `akita_black_bowl_vis.msh`, `model_vis.msh` for plates, `burnerplate.stl`); the only remaining proxy is the analytic `table`, which is boxes in MuJoCo too. Converted `akita_black_bowl_vis.msh` (127,566 vertices / 42,522 faces) end-to-end to a referenceable `UsdGeom.Mesh` and validated the authored USD with `usd-core`. All 141 LIBERO OBJ assets also parse and convert cleanly.

## Test surface

- `tests/simulation/isaac/test_mesh_assets.py` (new, 20 tests): OBJ (slash syntax, n-gons, negative indices, out-of-range), binary/ASCII STL, binary MSH (round-trip, truncation, out-of-range index), AABB + scale, USD conversion (referenceable output, content-addressed caching, USD passthrough, no torn cache entry on failure), extension-vocabulary disjointness.
- `tests/simulation/isaac/test_scene_object_meshes.py` (new, 12 tests): mesh passthrough with collision AABB unchanged, visual-group preference over collision meshes, mesh-bounds fallback for mesh-only bodies, missing-file `ValueError`, unregistered/unconvertible degrade paths, `meshdir` resolution, pre-existing analytic-scene contract unchanged, `load_scene` dispatch/report/caveat pins, failed realization is a loud skip.
- `tests/simulation/isaac/test_isaac_backend.py`: the explicit-refusal pin is replaced by behavior tests (mesh_path on a primitive shape refused, `shape='mesh'` without path refused, missing file refused by name, unsupported format refused with the supported set named, and a valid asset passing the mesh validation through to the ordinary lifecycle gates).
- `tests_integ/simulation/test_isaac_mesh_objects_gpu.py` (new, GPU-gated): `add_object(shape='mesh')` lands a real `UsdGeom.Mesh` prim with the asset-derived extent; `load_scene` realizes a mesh-carrying object with a visible mesh visual + invisible AABB collision cube; physics steps.

Full suite: **32623 passed, 288 skipped**, `hatch run lint` clean, changelog fragment included. The 5 pxr-gated conversion tests were additionally run with `usd-core` installed (skipped in the default env, which matches CI: `[all]` does not include `[sim-isaac]`).

## Acceptance criteria from the issue

- [x] `add_object(mesh_path=...)` on Isaac places a visible mesh -- implemented; CPU-verifiable halves pinned in `tests/`, live-stage half in `tests_integ/` (GPU-gated). The explicit-refusal test is replaced by behavior tests.
- [x] A LIBERO task loaded on Isaac shows the real object meshes, collision unchanged (AABB proxy) -- verified structurally against all 7 cached compiled LIBERO scenes; the invisible-proxy/visible-mesh split is pinned in the GPU test. **A/B image caveat**: this dev box has no Isaac Sim (Kit) install, so an RTX-rendered A/B frame could not be produced here; the GPU integration test asserts the visibility split (`ComputeVisibility()` on proxy vs. visual) and is ready to run on an Isaac Sim host.
- [x] MuJoCo path unchanged; loaders' fail-loud contract preserved (missing mesh file -> `ValueError`, never a silent box).
- [x] Cross-backend eval docs updated (`docs/simulation/isaac.md`): the caveat is stated as conditional -- it appears in the `load_scene` report exactly while any object renders as a proxy and disappears when the visuals match.

## Out of scope / follow-ups

- Switching *collision* from the AABB proxy to the mesh's convex hull for scene objects stays an explicit follow-up, per the issue's "collision unchanged unless explicitly switched" contract.
- Textures/materials for the referenced meshes (the LIBERO assets ship `texture.png` + MTL): the mesh geometry is what lands on the stage; material passthrough would be its own change.
- Running the GPU integration test + capturing the A/B frame requires an Isaac Sim install; tracked by the test itself (`STRANDS_GPU_TEST=1`).

## Board

Issue #2459 asks for board routing (Status=Todo -> In review, Priority=Medium); the filing account lacks org-project write, so please move it to "In review" manually.


yinsong1986:
Automated check-fix could not resolve **call-test-lint / Test and Lint**: the sole failure is `tests/training/test_lerobot_e2e.py::test_record_train_load_loop`, which failed with `URLError: [Errno 104] Connection reset by peer` while lerobot's ACT policy creation downloaded the `ResNet18_Weights.IMAGENET1K_V1` backbone over the network on the runner (33818 other tests passed). This is a runner-side network flake unrelated to this PR's diff (Isaac-only: loaders, mesh_assets, simulation, Isaac tests, docs). The same suite passed on this branch's three previous heads (most recently `0694ad9b`, ~6h before), and the current head `c3a7468` only merges an all-green `main`. A re-run of the failed job should clear it, but the automation token lacks `actions: write` to trigger one. Needs manual attention: please re-run the failed job.

cagataycali (approved):
Approving. I went looking for the two ways this change could be quietly worthless - a mesh path that silently degrades back to a box, and a docs claim that outruns the code - and neither is present.

**The no-silent-fallback doctrine is real, not aspirational.** `mesh_assets.py:71` promises "never a silent fallback" and `_require_mesh_file` delivers it: `FileNotFoundError` on a missing path, `ValueError` on an extension outside `MESH_EXTENSIONS`, with `USD_EXTENSIONS` named in the message so a `.usdz` caller is redirected rather than refused blankly. The parsers are equally strict - out-of-range OBJ face index, truncated MSH, non-numeric vertex, a facet with the wrong vertex count, and an empty mesh are each a raise with file and line number, not a shrug. `test_missing_mesh_file_is_an_error_never_a_silent_box` pins exactly the failure this feature exists to prevent, which is the right test to have written first.

**The deliberate fallbacks are separated from the accidental ones.** `test_unregistered_mesh_reference_degrades_to_the_box_fallback` and `test_unconvertible_format_stays_a_box_proxy` pin the cases where a proxy is still correct, and `test_mesh_only_body_gets_mesh_bounds_not_the_5cm_fallback` pins that a mesh body stops inheriting the hardcoded 5 cm box. That trio is what makes the box proxy a documented outcome for known-unrealizable assets rather than an undetectable one. `test_report_carries_the_caveat_while_proxies_remain` / `test_caveat_absent_when_every_object_has_a_mesh_visual` then make partial realization self-reporting, and `test_failed_mesh_realization_is_a_loud_skip` keeps a broken asset loud. Cache discipline is covered too - `test_conversion_is_cached_by_content` and `test_no_torn_cache_entry_on_failure`, the second being the one people forget.

**The `size` semantics are honest, which is the subtle part.** For `shape="mesh"`, `size` is documented as ignored - the asset's units define the extent - and the docstring goes further than most would: it names MuJoCo as sharing that contract, names Newton as consuming `size` as a scale instead (#2300), and states that the result reports the extent computed from the asset "rather than echoing a request no component of which was consumed." That last clause is what keeps this out of the trap `test_mesh_size_docs_match_backend_divergence.py` exists to document, where "both calls report success" while behaviour differs: here the returned payload disagrees with the request, so the ignore is observable by the caller rather than invisible. The edit to that test's prose is correspondingly accurate - Isaac genuinely does sit on MuJoCo's side of the divergence now, and replacing "Isaac refuses mesh objects outright" was required, not optional.

### One hardening, not a condition of merge

The claim that gate now leans on - Isaac ignores a mesh `size` - is the one claim in this PR that nothing pins.

The GPU journey test asserts `payload["size"] == approx([0.1, 0.1, 0.1])` for the tetra asset, but it calls `add_object` **without** a `size=`, so it pins "the extent comes from the asset when none was requested". It does not pin "a requested `size` is discarded and the asset extent reported anyway", which is the actual divergence assertion. Nothing fails today if Isaac later starts consuming `size` as a scale the way Newton does - and at that point `test_mesh_size_docs_match_backend_divergence.py` would itself be asserting something false about a third backend, which is a worse outcome than the divergence it was written to describe.

The fix is one argument on a call that already exists, so it costs no new test and no extra Kit boot: pass `size=[2.0, 3.0, 4.0]` in journey 1 and keep the existing `approx([0.1, 0.1, 0.1])` assertion. A value that is wrong on every axis makes the discard unambiguous, and it is the same `[2, 3, 4]` the divergence test uses as its example, so the two read as a matched pair.

I am not blocking on it, because nothing in this PR is false or broken - the behaviour, the docstring and the gate prose all agree with each other today, and this is hardening against a future regression rather than a defect. Given the PR has been waiting four days, the feature is worth more landed than held for a test argument. Filing it as a follow-up so it survives the merge; if you would rather add the one argument before merging, that closes it outright.

PR #372, opened 2026-06-08
Harden mesh actuation, telemetry, and inference control surfaces

cagataycali:
## Summary

Proactive security hardening of the mesh networking layer and the GR00T inference tool. Tightens five control surfaces so an untrusted agent context (prompt injection, manipulated tool output, a compromised mesh peer) cannot drive physical actuation, leak cross-peer telemetry, or reach the host filesystem.

All changes are safe-by-default and consumer-configurable via env vars.

### 1. Zenoh ACL example templates
- Rewrote `examples/mesh_acl_example.json5` to scope `declare_subscriber` by **role**: the robot role gets a narrow subscribe rule (presence, broadcast, safety, own cmd/response) instead of a blanket `**`; the operator role keeps broad observe as an explicit, documented grant.
- Added `examples/mesh_acl_strict_per_peer.json5` for operators who need hard per-peer subscribe isolation (one rule + subject per robot CN), documenting the Zenoh 1.x literal-CN limitation honestly.

### 2. `policy_host` defense in depth
- Added an `is_safe_policy_host()` re-check inside `Mesh._dispatch` so an off-allowlist VLA policy host is rejected even if a caller reaches `_dispatch` without passing through `validate_command`.

### 3. Configurable human-in-the-loop interrupt set (`robot_mesh`)
- New `STRANDS_MESH_HITL_ACTIONS` env (`all` / `none` / explicit subset).
- Default now gates **every physical-actuation action** (`emergency_stop`, `broadcast`, `tell`, `send`, `stop`) -- previously only the two fleet-wide ones.
- Command bodies for `tell` / `send` / `broadcast` are validated **before** the interrupt fires, so an operator never approves a payload the validator then rejects. Malformed env values fail loud.

### 4. subscribe/inbox telemetry-leak hardening (`robot_mesh`)
- Added a tool-layer topic allowlist (`STRANDS_MESH_SUBSCRIBE_ALLOW`, defaults to low-impact shared classes) so the tool cannot subscribe to another peer's control/sensor streams.
- Every `inbox` read is now audited.

### 5. `gr00t_inference` container hardening
[...]

PR #370, opened 2026-06-08
feat: add Device Connect integration (on top of main)

kavya-chennoju:
## Summary

Re-applies the Device Connect integration onto `main`'s architecture, as an alternative to #52 (which targets `dev`). `main` and `dev` diverged structurally — `main` carries the productionized `mesh/` subsystem and a `Robot()` factory in `robot.py` — so this branch **ports** the integration onto `main` rather than merging the `dev`-based branch wholesale (a direct merge produced a 367-file diff with broken `robot.py`).

## What's included

- **`strands_robots/device_connect/`** — `DeviceDriver` adapters wrapping a `Simulation` / `HardwareRobot` (`sim_driver`, `robot_driver`, `reachy_*`), plus `init_device_connect[_sync]`, `GUIDE.md`, `setup.sh`, and a control-loop smoke test. The drivers target `main`'s `Simulation` API unchanged (`start_policy`, `get_features`, `get_state`, `step`, `reset`, `_world` / `SimRobot` fields).
- **`robot.py`** — `Robot("…").run()` server hook attached by the factory. `run()` stops the auto-started mesh and brings the device online via Device Connect (the primary networking layer in server mode).
- **`tools/robot_mesh.py`** — Device Connect dispatch layered **underneath** the existing safety gates. `robot_mesh()` still runs its rate limit, HITL approval (`emergency_stop` / `broadcast`), command validation, and audit first; only then does it dispatch via Device Connect, falling back to the built-in mesh when DC is unavailable or has no devices. DC dispatch re-applies the per-action validation/audit so it inherits the same safety. Gated by `STRANDS_ROBOT_MESH_DC` (default on; disabled in the test suite so mesh unit tests stay hermetic).
- **`__init__.py` / `pyproject.toml`** — lazy DC exports + `device-connect-edge` / `device-connect-agent-tools` core deps.

## Behavior change

`emergency_stop` / `broadcast` are HITL-gated and **fail closed** when called as a bare script (no operator to approve); they run from within a Strands agent loop on approval. `GUIDE.md` documents this.

## Testing

- **1236 passed / 3 skipped** — the Device Connect tests together with `main`'s `robot_mesh` safety, deep-mesh, and factory tests.
- GUIDE D2D demo (`peers` / `tell` / `emergency_stop`) verified end-to-end against the latest `arm/device-connect` packages.

## Notes

[...]

PR #2499, opened 2026-08-19
docs(mesh): name rpc in every enumeration of the HITL-gated action set

cagataycali:
## Problem

`rpc` - the Device Connect device-native call - is gated by default by the `robot_mesh` human-in-the-loop interrupt. Every human-facing enumeration of the gated set omitted it.

The gate's two constants own the set: `_GATEABLE_ACTIONS` (what `STRANDS_MESH_HITL_ACTIONS` accepts, 8 actions) and `_DEFAULT_INTERRUPT_ACTIONS` (gated when the operator configures nothing, 6 actions). Both contain `rpc`. Measured against them, eight surfaces listed the set and named five or seven of it:

| surface | names | omits |
|---|---|---|
| README `STRANDS_MESH_HITL_ACTIONS` row | 7 of 8 gateable | `rpc` |
| `docs/security.md` - "every physical-actuation action is gated" | 5 of 6 | `rpc` |
| `docs/security.md` - "Audit trail. Every ..." | 5 of 6 | `rpc` |
| the warning logged when the gate is disabled | 5 of 6 | `rpc` |
| the dispatcher's interrupt-set comment | 5 of 6 | `rpc` |
| the tool docstring's audit list | 5 of 6 | `rpc` |
| `examples/lerobot/hub_to_hardware.py` operator note | 5 of 6 | `rpc` |
| the HITL config suite's contract docstring | 5 of 6 | `rpc` |

Two consequences, both measured by driving the real resolver.

**The documented vocabulary cannot express the shipped default gate.** `_DEFAULT_INTERRUPT_ACTIONS - documented tokens == {'rpc'}`. An operator keeping the default minus one action writes the rest of it from the documented list:

```
STRANDS_MESH_HITL_ACTIONS=emergency_stop,broadcast,tell,send
-> accepted; gated set = broadcast, emergency_stop, send, tell
-> meant to drop: stop
-> also ungated:  rpc
```

The value parses, so the `_InterruptConfigError` that exists so "a typo fails loud rather than silently degrading the gate" never fires. A second actuation action leaves the gate with no notice.

**The warning shown when the gate is disabled under-reported what it disabled**, naming five of the six actions it had just left unguarded - and that warning is the operator's only notice at that point.

`rpc` is also audited (`if action == "rpc"` reaches `_audit_tool_action`), so the audit-trail claim omitted a member of the set it quantifies over.

## Why this is drift rather than a design choice

`rpc` joined both constants when Device Connect landed in #370 (2026-06-13), one day after the README row was written in #372 (2026-06-12). The row was correct when written. #370 updated its own documentation - `docs/device-connect.md` and `strands_robots/device_connect/GUIDE.md` both name all six actuation actions - and the shared mesh surfaces did not follow. AGENTS.md makes the README's env-var list "the single source of truth for users".

## Change

Name `rpc` in all eight enumerations. No behaviour changes: the constants, the resolver and the gate are untouched, so the accepted vocabulary and the gated set are identical before and after.

`tests/mesh/test_hitl_gate_enumerations_name_every_gated_action.py` derives the required set from the constants and grades each surface, so an action added to the gate later fails until every enumeration names it. It also keeps the two Device Connect guides graded, so the asymmetry cannot reappear from the other side.

While here, the HITL config suite's docstring said "the five actuation actions"; the count is dropped rather than corrected so it cannot go stale again, and `test_resolver_default_is_actuation_set` now asserts `rpc` alongside its siblings.

Out of scope, stated for the reviewer: README's `robot_mesh` action summary (`peers`, `status`, `tell`, ...) enumerates the tool's whole action surface rather than the gated set, and omits `rpc` and `unsubscribe`; that is a different set and a different question.

## What an operator sees

HITL gate enumerations before and after

One capture script per tree, each reading its **own** tree's docs to decide what an operator would write, then driving the real `_parse_interrupt_actions` and the real warning. Package code is identical in both columns; only what the operator is told differs. Left: the warning names 5 of 6, the narrowing silently ungates `rpc`, 8 of 9 surfaces incomplete. Right: 6 of 6, only `stop` dropped, 0 incomplete.

## Tests

Pre-fix (this test file on `upstream/main`): **9 failed / 12 passed**, each failure naming `rpc` and quoting the real text, e.g.

```
the README configuration row (README.md) lists the gated actions but omits ['rpc']:
'| `STRANDS_MESH_HITL_ACTIONS` | ... subset of
 `emergency_stop,broadcast,tell,send,stop,subscribe,watch` | actuation default |'.
An operator can only narrow the gate to actions they have been shown, and an
omitted action is ungated without a refusal.
```

Post-fix: **21 passed** (13 new + the 8 existing resolver tests). The 12 both-tree passes are the controls: `subscribe`/`watch` stay out of the default gate (so widening the default is not a way to make an enumeration true), the Device Connect guides stay complete, `rpc` really is gated and audited, and every graded anchor is still present so a reflow fails loudly instead of reporting clean.

## Verification

Measured at `012703ca` against `upstream/main` `0acbfc9`.

- 240-file blast radius (every test touching the gate, its env var, `robot_mesh`, a docs page, plus the repo-wide docs/changelog/docstring guards), run on both trees: branch **23 failed / 6243 passed / 24 errors**, base **32 / 6234 / 24**, **6295 collected on both**. Branch-only failing ids: **none**. Base-only: exactly the 9 pre-fix failures, no foreign ids. The 47 shared pre-existing are `awsiot` / `device_connect_agent_tools` absent locally (both declared dependencies).
- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 against a pristine `upstream/main` worktree, none in the changed files.
- No non-ASCII added to any changed file; `scripts/assemble_changelog.py --check` passes.


yinsong1986 (approved):
## Summary

Documentation-consistency fix: `rpc` (the Device Connect device-native call) has been in both `_GATEABLE_ACTIONS` and `_DEFAULT_INTERRUPT_ACTIONS` since Device Connect landed, but eight human-facing enumerations of the HITL-gated set omitted it — including the README's `STRANDS_MESH_HITL_ACTIONS` token vocabulary (so an operator narrowing the gate from the documented list silently ungated `rpc` with no `_InterruptConfigError`) and the one-time warning shown when the gate is disabled. The diff names `rpc` in all eight surfaces and adds a test suite that derives the required set from the two constants and grades every enumeration, so a future action added to the gate fails until each surface names it. No behavior changes: the constants, resolver and gate are untouched; the only production-code edits are string/comment/docstring content in `robot_mesh.py`.

## What's good

- The new test derives the required set from `_GATEABLE_ACTIONS` / `_DEFAULT_INTERRUPT_ACTIONS` rather than restating it, and `test_every_graded_surface_is_still_present` makes a reflowed anchor fail loudly instead of reporting clean.
- Controls pin the boundaries: `subscribe`/`watch` stay out of the default gate (so widening the default is not a way to make an enumeration true), and the `rpc` audit premise is verified via AST against the real dispatcher.
- Changelog fragment present (`changelog.d/2499-...`), all-ASCII additions, and the stale count in the HITL config suite docstring is dropped rather than re-pinned.

Verified locally at head `c0e01512`: the two touched test files report 21 passed.

PR #411, opened 2026-06-13
fix(robot_mesh): close decline side-channel + audit read-only actions; AGENTS docs (#394)

cagataycali:
## Problem

A declined request in `robot_mesh` could still leak state through a side-channel, and read-only actions were not audited, so they were not observable in the audit log.

## Change

Close the decline side-channel so a declined request leaks no state, and add audit logging for read-only actions. Document the behavior in AGENTS.md.

## Tests

`test_robot_mesh_security.py` covers the decline path and read-only action auditing (24 cases). `hatch run lint` clean.

Resolves #394.

PR #2500, opened 2026-08-19
fix(robot_mesh): audit the read-only actions on the Device Connect backend too

cagataycali:
## Problem

`robot_mesh` renders each action onto one of two backends: an agent-side Device Connect connection when one is reachable and has devices, and the built-in Zenoh mesh otherwise. The audit contract is a property of the **action** - "the audit log is a complete record of agent mesh access" - which is why the read-only observation actions are recorded alongside the actuation ones rather than only when something moved.

That half of the contract reached the mesh rendering of `peers` / `status` and not the Device Connect rendering. Device Connect is tried **first** whenever it has devices, so the audited implementations were the fallback and the unaudited ones the default.

Driving the real `_device_connect_dispatch` against a two-device fleet, one agent session, reading the audit log back:

| action | reply | device ids returned | function names returned | audit row |
|---|---|---|---|---|
| `peers` | 153 chars | 2 of 2 | **5 of 5** | **none** |
| `status` | 75 chars | 2 of 2 | - | **none** |
| `tell` | 61 chars | 1 | - | yes |
| `rpc` | 68 chars | 1 | 1 | yes |

Four calls, two rows. `peers` is the richest read the tool offers: it returns every device id **plus every function name the fleet exposes**, which is the callable surface a later `rpc` would use - and it was the one call that left no forensic trail. Every other action this backend answers already audited its success path, so these two were the exception rather than the rule.

`git log -S` dates it: the Device Connect branch arrived in #370, and #411 closed the read-only audit gap on the mesh branch a while later without bringing the sibling backend along. Nothing pinned the read-only half of the contract - `TestAudit` covers `emergency_stop` and a declined call only - which is why it went unnoticed.

## Change

- `_device_connect_dispatch` writes a row for the `peers` / `status` branch, recording the size of the fleet that was read (`devices=N`) to match the mesh rendering's `local=N remote=M`. A marker with no magnitude is not a usable record of what was read.
- `docs/security.md`'s audit-trail bullet enumerated only the five actuation actions, so it understated the trail the tool already keeps. It now names the read-only half and says the row is written on whichever backend served the action.
- AGENTS.md's *Audit completeness* rule gains the backend dimension and a `Pinned by` bullet.

Replies are untouched: the reply for every action is byte-identical either side (153 / 75 / 61 / 68 chars), and both pre-existing actuation rows are unchanged. The change adds two rows and alters none.

### Scope

Only the success path of actions this backend **answers**. `subscribe` / `watch` / `inbox` / `unsubscribe` are mesh-only and still defer (a control test pins that widening the audit here did not turn into this backend answering an action it does not implement). Whether a request refused by a required-parameter guard - one that never reached a device - belongs in the audit log is a separate question from whether a completed read does, and is left alone; a control test pins that boundary too.

## Tests

`tests/mesh/test_robot_mesh_readonly_audit_parity.py` (24 items). The graded set is discovered by **calling the dispatcher** rather than restating an action list, so an action added to this backend is covered without editing the test.

**5 failed / 19 passed against `upstream/main`; 24 pass after.** Every pre-fix failure is behavioural and names the gap, e.g.

```
AssertionError: Device Connect answered 'peers' with success and wrote no audit row,
so this access is absent from the forensic record the mesh rendering of the same
action keeps.
```

The 19 that pass on both trees are the controls, and each fails for a specific wrong fix: every actuation action still audits its success (fails if auditing is made conditional), the observation reply still carries the fleet listing (fails if the row replaces part of the reply), a required-parameter refusal stays a refusal recorded as no success, and the four mesh-only actions still defer and are not recorded here.

## Verification

Re-measured at `83f5749` after merging `upstream/main` = `9913887` (below).

- Blast radius: 106 files (every test touching `robot_mesh` / the audit helper / Device Connect / a docs page, plus the repo-wide docstring, ASCII, changelog and dependency guards). Branch **2 failed / 3976 passed / 9 skipped**, base **7 failed / 3971 passed / 9 skipped**, both collecting 3987. Failed delta == passed delta == 5 == the pre-fix count; **branch-only failure set empty**; base-only is exactly the five new tests. The two shared failures need `device_connect_agent_tools`, a declared dependency absent from this local environment.
- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ` clean. `mypy` 24 == 24 against a pristine `upstream/main` worktree, none in the changed files.

## Artifact

One agent session over Device Connect, run under each tree - same package code either side except the audit call - with the audit log read back as an operator would see it.

read-only audit parity

### Rebased onto a moved base

`docs/security.md`'s audit-trail bullet was edited on both sides: #2499 named `rpc` in the
actuation enumeration, this branch added the read-only actions and the backend that served them.
Both are kept - the enumeration is graded against the default gated set by
`tests/mesh/test_hitl_gate_enumerations_name_every_gated_action.py`, and that guard passes on the
merged bullet. `strands_robots/tools/robot_mesh.py` merged without conflict, and both sides' symbols
are present: the seven `rpc` references from #2499, and this branch's audit call.

Re-verified on the merged tree, because a moved base invalidates numbers measured against the old one:

- Blast radius widened to 273 files (all of `tests/mesh`, plus every test touching `robot_mesh`, the
  audit helper, Device Connect or a docs page, plus the repo-wide guards). Branch
  **23 failed / 7406 passed / 64 skipped / 24 errors**, base **28 failed / 7401 passed**, both
  collecting **7429**. Failed delta == passed delta == 5 == the pre-fix count; **branch-only failure
  set empty**; base-only is exactly the five new tests. The 47 shared failures need `awsiotsdk`
  (45, all `tests/mesh/test_iot_*`) or `device_connect_agent_tools` (2), both declared
  dependencies absent from this local environment.
- The pre-fix split is unchanged against the new base: **5 failed / 19 passed**, all 24 pass after.
- `ruff check` and `ruff format --check` clean; `mypy` 24 == 24 against a pristine `upstream/main`
  worktree, none in the changed files; changelog fragments validate.


yinsong1986 (dismissed):
## Summary

Closes the read-only audit gap on the Device Connect backend of `robot_mesh`: `peers` / `status` answered by `_device_connect_dispatch` now write an audit row (`devices=N`) matching the mesh rendering's `local=N remote=M`, so the backend tried *first* no longer skips the forensic record the fallback keeps. The one production change is a single `_audit_tool_action(...)` call on the existing success path of the `peers`/`status` branch, using the same best-effort helper every other action in the file already uses (never raises, wide catch documented at the helper), so no reply bytes change and no new failure mode is introduced. Docs (`docs/security.md`) and AGENTS.md's *Audit completeness* rule are updated to state the backend dimension of the contract, and a changelog fragment is present per the fragment rule. Verified locally at `4d126ca8`: the new test module passes 24/24 on the branch, and with `strands_robots/tools/robot_mesh.py` reverted to `main` (`0acbfc9`) it fails exactly the 5 behavioural items the description claims, with the 19 controls (actuation rows unchanged, reply payload intact, required-parameter refusal not recorded as success, mesh-only actions still deferring) passing on both trees.

## What's good

- The graded action set is discovered by driving the dispatcher rather than restating a list, so a new Device Connect action is covered without editing the test.
- Controls pin the scope boundaries explicitly: the audit row is additive to the reply, refusals stay unrecorded successes, and the four mesh-only actions still defer - so the fix cannot silently widen into the backend answering actions it does not implement.
- The audit detail records a magnitude (`devices=N`) rather than a bare marker, matching the mesh side's convention.
- AGENTS.md rule lands with a `Pinned by` bullet, and the changelog fragment follows `changelog.d/<pr>-<slug>.md`.

yinsong1986 (approved):
## Summary

Closes the read-only audit gap on the Device Connect backend of `robot_mesh`: the `peers` / `status` branch of `_device_connect_dispatch` now writes an audit row (`devices=N`) matching the mesh rendering's `local=N remote=M`, so the backend tried first when it has devices no longer skips the forensic record the fallback keeps. The single production change is one call to the existing best-effort `_audit_tool_action` helper (which cannot raise into the dispatch path by design), placed on the success path after the reply text is built - replies are untouched, both pre-existing actuation rows are unchanged, and the mesh-only actions still defer. `docs/security.md`'s audit-trail bullet now names the read-only half it previously understated, and I verified the claim against the code: all six read-only actions (`peers`, `status`, `subscribe`, `watch`, `inbox`, `unsubscribe`) do write rows on the backend that serves them.

## What's good

- Reproduced the claimed pre-fix split exactly: reverting only `strands_robots/tools/robot_mesh.py` to `upstream/main` under the new test file gives **5 failed / 19 passed**; the PR head passes all 24. Every pre-fix failure names the gap behaviourally.
- The graded action set is discovered by calling the dispatcher rather than restating a list, and the controls pin the scope boundaries (actuation rows unconditional, reply unchanged, required-parameter refusals not recorded as success, mesh-only actions still deferred and unrecorded here) - per AGENTS.md > Review Learnings (PR-7) > Audit completeness, now extended with the backend dimension and a `Pinned by` bullet.
- No public API, wire format, or persisted-schema change: the row reuses the existing `log_safety_event("llm_tool_action", ...)` shape, and `devices=N` is a detail string consistent with the sibling branches' conventions.
- Changelog fragment present (step 3), `ruff check` / `ruff format --check` clean on both changed Python files, no non-ASCII or host paths in the additions.

PR #2501, opened 2026-08-19
fix(mesh/audit): poison a record the JSON encoder cannot represent instead of dropping it

cagataycali:
## Problem

`log_safety_event` has four degraded paths, and each one writes a record carrying a discriminating `sig` rather than returning early, so a `verify_audit_integrity` walker can attribute a stream anomaly to a failure class:

| trigger | `sig` written |
| --- | --- |
| the seq lockfile is a symlink | `SEQ_LOCK_DEGRADED` |
| the seq sidecar is unreadable | `NEXT_SEQ_DEGRADED` |
| the PSK flipped mid-run | `PSK_DEGRADED` |
| signing raised | `SIGN_FAILED` |
| **the payload is not representable** | **nothing - the record was dropped** |

The serialisation branch kept only the fail-soft half of the contract. That matters because of the ordering: `_next_seq` consumes **and persists** the peer's sequence number before the record is encoded, so the `return` left the sequence number spent with nothing to name it. The module header documents that state as *"sequence gaps (records were deleted)"* - an ordinary payload mistake wearing the forensic signature of tampering, on the one degraded path a walker could not classify.

Measured on the shipped path (`Mesh.publish_safety_event`, one arm, one shift: estop -> zone -> collision -> clear, where the collision payload carries the pose it is about):

```
wire:  4 safety events published to the fleet
audit: 3 records          seq 1, 2, ... 4
verify_audit_integrity(): ok=False  sequence_gaps=[(2, 4)]
log:   WARNING [audit] could not serialise record ... -- record dropped
```

The fleet was told four times; the forensic record keeps three, and the missing one is the collision.

The trigger is ordinary, not exotic. Of nine payload values a robot safety event plausibly carries, eight reach this branch - a numpy pose or joint vector, `np.float32`, raw bytes, a set of zone ids, a `datetime`, a `Path`, a `Decimal`, a bare object. With a PSK configured the code even *classifies* it (`_sign_record` raises the same `TypeError`, so `sig="SIGN_FAILED"` is set) and then discards the classification one line later.

## Change

On a serialisation failure the record is now poisoned rather than dropped: the envelope (`ts` / `event` / `peer_id` / `seq`) is kept, the unrepresentable payload is replaced by `{"unrepresentable": True}` plus a bounded `payload_error` diagnostic, and `sig` is set to `SERIALISE_FAILED`. The only remaining drop is an envelope that is itself unrepresentable - a non-string `event_type` - where there is nothing left to poison with, and that path keeps the original WARNING.

```
wire:  4 safety events published to the fleet
audit: 4 records          seq 1, 2, 3, 4
       seq 3  collision_imminent  SERIALISE_FAILED  {'unrepresentable': True}
verify_audit_integrity(): ok=False  sequence_gaps=[(2, 4)]
log:   ERROR [audit] SERIALISE_FAILED for peer_id='arm-7': ... -- writing poison record
```

Both sides report `ok=False`, and that is correct: a degraded audit is never ok. What changes is whether the anomaly can be attributed. Driven between a clean record and a clean record, the new class behaves identically to the two existing poison classes that also occupy a real sequence number:

| case | records | `sig` | `seq` | `sequence_gaps` | `ok` |
| seq lockfile is a symlink | 3 | `SEQ_LOCK_DEGRADED` | 0 | `[]` | False |
| seq sidecar unreadable | 3 | `NEXT_SEQ_DEGRADED` | 0 | `[]` | False |
| PSK flipped mid-run | 3 | `PSK_DEGRADED` | 2 | `[(1, 3)]` | False |
| signing raised | 3 | `SIGN_FAILED` | 2 | `[(1, 3)]` | False |
| payload not representable (before) | 2 | *no record* | - | `[(1, 3)]` | False |
| payload not representable (after) | 3 | `SERIALISE_FAILED` | 2 | `[(1, 3)]` | False |

## Scope

`tests/mesh/test_audit_serialise_safety.py` already owned this contract - it exists to prove the function does not raise - and its docstring justified the drop as "matching the existing fail-soft discipline for signing failures". The signing discipline it cites does not drop; it poisons. Its module docstring now states both halves, and `test_warning_logged_on_non_serialisable` becomes `test_the_failure_is_reported_under_its_discriminator`: the stated intent was that the failure is reported rather than silent, and it is now reported at ERROR under its discriminator, the level and shape the other four poison paths use.

The `sig` key description in the module header said a signature is "present only when `STRANDS_MESH_AUDIT_PSK` is configured", which the four existing poison classes already contradicted and this branch makes reachable from an ordinary payload mistake; it now names the discriminators.

## Tests

`tests/mesh/test_audit_serialise_safety.py`: **10 failed / 12 passed** against `upstream/main`, 22 passed after.

The 10 failures are behavioural - the headline reports that the unrepresentable payload left no record, so the sequence number `_next_seq` already consumed names no failure class. The 12 that pass on both trees are the controls, and each fails for a specific wrong fix:

- `TestEveryDegradedPathWritesARecord` drives all five degraded branches from one table, so the four already-compliant rows pass unchanged and a path added later is graded rather than quietly becoming the next silent drop.
- the verifier still fails closed on a poisoned stream - the poison record must not read as a pass.
- a representable payload carries no `sig` and no `payload_error`: the write path a caller normally takes is byte-identical.
- an unrepresentable *envelope* still raises nothing and still logs `record dropped`.
- the three pre-existing "does not raise" cases and the nan/inf case are untouched.

## Verification

At `d0fa014`, against `upstream/main` `0acbfc93`.

- 254-file blast radius (all of `tests/mesh`, every test touching the audit surface, and the repo-wide docstring / string / changelog / dependency guards), run sequentially on both trees.
- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 with zero branch-only errors and none in the changed files.
- No added non-ASCII in any changed file.

audit stream after a collision event whose payload carries a pose

The figure is one capture script run against each tree with the same package code either side; only `strands_robots` differs. The `wire: 4` line is identical in both panels, which is the control - the change touches only what the audit keeps.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1737,9 +1737,21 @@ apply to all future work on `strands_robots/mesh/{core,audit,security}.py`.
 ### Audit poison-record symmetry
 - **Every degraded audit path writes a poison record, never a silent drop.** The
   poison `sig` discriminators (`PSK_DEGRADED`, `SIGN_FAILED`, `SEQ_LOCK_DEGRADED`,
-  `NEXT_SEQ_DEGRADED`) let a `verify_audit_integrity` walker attribute a stream
-  gap to a specific failure class. When you add a new `_next_seq`/sign/persist
-  failure branch, add the matching poison `sig` instead of returning early.
+  `NEXT_SEQ_DEGRADED`, `SERIALISE_FAILED`) let a `verify_audit_integrity` walker
+  attribute a stream gap to a specific failure class. When you add a new
+  `_next_seq`/sign/serialise/persist failure branch, add the matching poison
+  `sig` instead of returning early.
+- **`_next_seq` runs before serialisation, so an early `return` is a deletion.**
+  The sequence number is consumed and persisted before the record is encoded, so
+  a branch that gives up after that point leaves the signature this module's
+  header documents as "records were deleted". A payload the JSON encoder cannot
+  represent keeps the envelope (`ts` / `event` / `peer_id` / `seq`) and swaps the
+  payload for a bounded diagnostic; the only remaining drop is an envelope that
+  is itself unrepresentable, where there is nothing left to poison with.
+- Pinned by `tests/mesh/test_audit_serialise_safety.py`, whose
+  `TestEveryDegradedPathWritesARecord` drives every degraded branch from one
+  table so a path added later is graded rather than quietly becoming the next
+  silent drop.
 
 ### Replay-cache eviction
 - **TTL purge runs unconditionally, not only when the cache is full.** On a
--- a/changelog.d/2501-audit-serialise-poison-record.md
+++ b/changelog.d/2501-audit-serialise-poison-record.md
@@ -1,0 +1,2 @@
+### Fixed
+- **mesh/audit**: a safety event whose payload the JSON encoder cannot represent now leaves an attributable poison record (`sig="SERIALISE_FAILED"`) instead of being dropped. `_next_seq` consumes and persists the peer's sequence number before the record is encoded, so the drop left the signature the module header documents as "records were deleted" - the one degraded path a `verify_audit_integrity` walker could not attribute to a failure class. The envelope (`ts` / `event` / `peer_id` / `seq`) is kept and the payload is replaced by a bounded diagnostic.
--- a/strands_robots/mesh/audit.py
+++ b/strands_robots/mesh/audit.py
@@ -23,9 +23,13 @@ Each line is one JSON object with these keys:
 * ``payload`` -- free-form dict with event-specific fields
 * ``seq`` -- process-monotonic sequence number. Useful for detecting
   truncation: gaps within a single peer's stream indicate missing events.
-* ``sig`` -- HMAC-SHA256 hex over the rest of the record. Present only when
+* ``sig`` -- HMAC-SHA256 hex over the rest of the record, present when
   ``STRANDS_MESH_AUDIT_PSK`` is configured. Verifies that the record
-  content has not been edited after write.
+  content has not been edited after write. The field also carries one of the
+  poison discriminators (``PSK_DEGRADED``, ``SIGN_FAILED``,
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes the one degraded path in `log_safety_event` that dropped a record instead of poisoning it: a payload the JSON encoder cannot represent now keeps its envelope (`ts` / `event` / `peer_id` / `seq`) and writes `sig="SERIALISE_FAILED"` with a bounded `payload_error` diagnostic, so the sequence number `_next_seq` already consumed and persisted no longer reads as a deletion. The change is symmetric with the four existing poison classes and the verifier fails closed on the new record in both postures: with a PSK the discriminator fails the HMAC compare (`bad_signature`, cursor held, gap attributed), and without one it counts as `unverifiable_signed` -- both confirmed by reading the walker at head, and the seq-seeding walk skips the poison record under a PSK via the same failed compare while correctly trusting its genuine `seq` in the dev posture. The nested retry's only remaining drop is an envelope that is itself unrepresentable, preserving the original WARNING there.

## What's good

- `TestEveryDegradedPathWritesARecord` drives all five degraded branches from one table, so a sixth branch added later is graded rather than becoming the next silent drop -- exactly the shape AGENTS.md > "Audit poison-record symmetry" asks for, and the AGENTS.md bullet is updated in the same PR.
- The diagnostic is bounded (`_MAX_POISON_REASON_CHARS`) and carries only the encoder's exception type and message, not payload values.
- Changelog fragment under `changelog.d/` instead of a direct `CHANGELOG.md` edit; ruff clean; no non-ASCII in changed files.

## Verification suggestions

- `pytest tests/mesh/test_audit_serialise_safety.py tests/mesh/test_audit_integrity.py tests/mesh/test_audit_psk_poison_record.py tests/mesh/test_safety_audit_poison_records.py tests/mesh/test_audit_seq_psk_verified_seed.py tests/mesh/test_mesh_audit.py --no-cov` -- 72 passed at `d0fa0142` when run here.

Issue #2305, opened 2026-08-16
decision(ci): docs.yml cancels a per-commit verdict and a shared-resource deploy under one group key - fixing the false red needs the two split, not a new key


cagataycali:
Follow-up to #2304, which is about `pr-and-push.yml`. Its last section asks to check the other workflows that subscribe to `push`; this is the one that does not have the same answer.

## The overlap with #2304

`docs.yml` keys its group on the ref and cancels in progress:

```yaml
concurrency:
  group: docs-${{ github.ref }}
  cancel-in-progress: true
```

On a push to `main` that is one group for the whole branch, exactly as #2304 describes, so each merge cancels the previous merge's docs run. Three of the eleven red commits in #2304's table are red partly or wholly for this reason rather than for `call-test-lint`:

| commit | PR | non-SUCCESS contexts |
|---|---|---|
| `c505c697` | #2291 | `call-test-lint`, **`deploy`** CANCELLED |
| `977a3d52` | #2292 | `call-test-lint`, **`deploy`** CANCELLED |
| `43e4987f` | #2267 | **`build`**, `call-test-lint`, **`deploy`** CANCELLED |

So #2304's one-line change settles `call-test-lint` on all eleven, and leaves these three still reading red on a `build` / `deploy` context.

## Why the same one-line change is the wrong fix here

The two jobs answer different kinds of question, and the workflow-level group covers both:

- **`build`** runs `mkdocs build --strict`. That is a per-commit verdict about a tree, the same category as `call-test-lint`, and cancelling it has the same defect: the commit it was testing keeps a permanently unfinished answer, and on `main` nothing ever re-establishes it.
- **`deploy`** publishes to GitHub Pages. A superseded run here is not a superseded verdict but a superseded *deployment*, and the newest commit's site is the one that should be live. Cancelling an older in-flight publish is the wanted behaviour, and keying the group per commit would let two publishes of different trees race to the same resource.

Keying `docs-${{ github.event.pull_request.number || github.sha }}` would fix `build` and break `deploy`. This is why #2304's PR pins `docs.yml` as an explicit exemption with this issue cited, rather than treating it as an oversight.

## The shape a fix probably has

GitHub's own Pages starter workflow separates the two concerns, and the separation is what is missing here rather than a key:

- workflow-level group per commit (`|| github.sha`), so `build` runs to completion per commit;
- a job-level group on `deploy` alone, `group: pages`, `cancel-in-progress: false`, so deployments serialize instead of cancelling.

[...]

Issue #2216, opened 2026-08-13
the closing-reference report cancels its own in-flight run, so a head that satisfies it keeps a permanent red context - the exemption's safety premise checked the required check's concurrency group, not its own


cagataycali:
## Summary

`closing-reference.yml` is the one workflow exempted from the sha-invariant-trigger rule, so it is the one workflow that can be started twice on a single head sha. It also carries `cancel-in-progress: true` on a group keyed on the pull request **number**, so when those two runs overlap one is cancelled -- and a cancelled check run is permanent, attaches to the *live* head, and aggregates into `statusCheckRollup.state == FAILURE`.

That is the same producer #1914 removed from the required check, reintroduced on the exempt workflow's own check. The exemption was admitted with an explicit safety argument, and the argument is sound about the thing it checked: it asks whether an `edited` event can reach **`pr-and-push.yml`'s** concurrency group, and the answer is no. It never asks whether the event can reach **`closing-reference.yml`'s own** group, where the answer is yes, because that run shares the workflow name and pull request number with the run already in flight.

## Measured, open population, 2026-08-12 ~23:30 UTC

18 open pull requests. Five heads ran `Closing Reference Check` twice on one sha; two of the five raced into a cancellation:

| pull request | head | runs on that head | contexts named `Refuse a closing keyword that only appears in the title` |
|---|---|---|---|
| #1722 | `741f4057` | #251 att 2, #252 att 2 | **`SUCCESS` + `CANCELLED`** |
| #2205 | `8ca67d34` | #323, #324 | **`SUCCESS` + `CANCELLED`** |
| #2204 | `9e626421` | #325, #326 | `SUCCESS` + `SUCCESS` |
| #2207 | `fe856c49` | #327, #328 | `SUCCESS` + `SUCCESS` |
| #2213 | `f958666b` | #337, #338 | `SUCCESS` + `SUCCESS` |

On #1722 and #2205 every other context is `SUCCESS`/`SKIPPED`/`NEUTRAL`, the required `call-test-lint / Test and Lint` included. The cancelled duplicate is the only non-`SUCCESS` context on either head, and each is masked by a same-named `SUCCESS`: GitHub does not dedupe rollup contexts by name, so the pass does not displace it.

[...]

PR #2503, opened 2026-08-19
ci: give the remaining pull_request workflows a concurrency group

### LINKED ISSUE #2502: ci: five of the eleven pull_request workflows declare no concurrency group, so a superseded head keeps running ahead of the current one
## What

A concurrency group is per workflow. `pr-and-push.yml` keys its own on `github.workflow`, so cancelling a superseded run of the required check says nothing about the other ten workflows a pull request starts. Each of those declares its own group or declares none, and five declare none:

| workflow | group |
|---|---|
| `codeql.yml` | none |
| `dependency-review.yml` | none |
| `breaking-change-check.yml` | none |
| `agent-api-check.yml` | none |
| `llm-input-safety.yml` | none |
| `pr-and-push.yml` | `${{ github.workflow }}-${{ github.event.pull_request.number \|\| github.sha }}` |
| `changelog-fragment.yml` | `${{ github.workflow }}-${{ github.event.pull_request.number }}` |
| `lockfile-parity.yml` | same |
| `merge-base-overlap.yml` | same |
| `last-push-approval.yml` | same, plus `-${{ github.event_name }}` |
| `docs.yml` | `docs-${{ github.ref }}` |
| `closing-reference.yml` | none, **deliberately** (#2216) |

A push to a pull request replaces the head sha those jobs read, so a run still going over the previous sha can only recompute a verdict about a commit no reviewer will open - on a runner the current head is queued behind.

## Measurement

Last 100 `pull_request` runs of each, read at 2026-08-19 ~19:30Z. Runner time is summed from job-level `started_at` / `completed_at`, not from the run object's `updated_at`: the latter is bumped after the jobs finish and reported 478 minutes where the jobs account for 67. A run counts as superseded when the next run on the same branch was created before it finished, which is the condition `cancel-in-progress` acts on.

| workflow | sampled window | superseded | runner-min after supersede |
|---|---|---|---|
| CodeQL | 08-18 .. 08-19 | 13 | 48.2 |
| Detect AgentTool API Breaking Changes | 08-16 .. 08-19 | 21 | 16.4 |
| Detect Breaking Changes | 08-17 .. 08-19 | 4 | 1.2 |
| dependency-review | 08-18 .. 08-19 | 6 | 0.8 |
| LLM Input Safety | 08-17 .. 08-19 | 1 | 0.1 |
| **total** | | **45** | **66.7** |

Three of the five account for ~2 minutes between them, so for those this is not a minutes argument and is not offered as one. How much a workflow wastes depends on how often a second push lands while its run is still going - a property of the review loop and of the job's duration, not of the workflow - so grading it per workflow would make the fleet's behaviour depend on which workflow happened to be sampled in a busy week.

The queue is what turns those minutes into latency. At 19:15Z the same day, 35 runs were queued against 1 in progress, and the in-progress run of the required check had waited 53 minutes between `created_at` 17:28Z and `run_started_at` 18:21Z.

## `closing-reference.yml` must keep declaring none

It is the one workflow subscribed to an activity type that cannot change the head sha (`edited`), so it is the only one that can have two runs on one head, and #2216 measured what cancelling one does: the head carries the same check as `SUCCESS` and `CANCELLED` together, and the roll-up read `SUCCESS`, then `FAILURE`, then `SUCCESS` across three reads of one unchanged sha. `tests/test_closing_reference_gate.py::test_the_gate_does_not_cancel_its_own_run` pins the absence of the block; any fleet-wide sweep has to record that as an exemption rather than tidy it away.

## Remedy

Give the five the group the other five pull-request workflows already use, and grade the rule so the next `pull_request` workflow inherits it instead of rediscovering it. `codeql.yml` also runs on `push` and `schedule`, so its group has to key the push side on the commit - the defect `tests/test_push_concurrency_group.py` exists for - and separate the events, so the Monday scan cannot cancel a pushed commit's analysis.

That module's own docstring currently names `codeql.yml` as the workflow declaring no group, which stops being true with this change and needs correcting in the same diff.

cagataycali:
## Problem

A concurrency group is per workflow. `pr-and-push.yml` keys its own on `github.workflow`, so cancelling a superseded run of the required check says nothing about the other ten workflows a pull request starts. Each of those declares its own group or declares none, and five declare none: `codeql.yml`, `dependency-review.yml`, `breaking-change-check.yml`, `agent-api-check.yml`, `llm-input-safety.yml`. Six already had one.

What a missing group costs is not a wrong answer but a run of the *right* answer to a question nobody asked. A push to a pull request replaces the head sha those jobs read, so a run still going over the previous sha can only recompute a verdict about a commit no reviewer will open - and it recomputes it on a runner the current head is queued behind.

## Measurement

Last 100 `pull_request` runs of each, read 2026-08-19 ~19:30Z. Runner time is summed from job-level `started_at` / `completed_at` rather than the run object's `updated_at`: the latter is bumped after the jobs finish and reported **478** minutes where the jobs account for **67**. That correction is worth stating because the first number is the one the obvious query returns. A run counts as superseded when the next run on the same branch was created before it finished, which is exactly the condition `cancel-in-progress` acts on.

| workflow | sampled window | superseded | runner-min after supersede |
|---|---|---|---|
| CodeQL | 08-18 .. 08-19 | 13 | 48.2 |
| Detect AgentTool API Breaking Changes | 08-16 .. 08-19 | 21 | 16.4 |
| Detect Breaking Changes | 08-17 .. 08-19 | 4 | 1.2 |
| dependency-review | 08-18 .. 08-19 | 6 | 0.8 |
| LLM Input Safety | 08-17 .. 08-19 | 1 | 0.1 |
| **total** | | **45** | **66.7** |

**Three of the five account for ~2 minutes between them, so for those this is not a minutes argument and is not offered as one.** How much a workflow wastes depends on how often a second push lands while its run is still going - a property of the review loop and of the job's own duration, not of the workflow - so grading it per workflow would make the fleet's behaviour depend on which workflow happened to be sampled in a busy week. The rule is the one the other six already follow: a `pull_request` run whose sha has been replaced should not still be running.

The queue is what turns those minutes into latency rather than cost. At 19:15Z the same day, 35 runs were queued against 1 in progress, and the in-progress run of the required check had waited 53 minutes between `created_at` 17:28Z and `run_started_at` 18:21Z. An obsolete run does not merely spend a minute; it spends it ahead of the head everybody is waiting on.

## Change

Four of the five get the block the other pull-request workflows already carry, bare and identically spelled, because the reasoning belongs in the pin rather than in five copies of a comment:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

`codeql.yml` is the one that needs a different key, and it carries the only comment: it also runs on `push` and on the Monday `schedule`. Keying the push side on the ref is the defect `tests/test_push_concurrency_group.py` was written for - each merge cancels the previous merge's run and the commit keeps a permanent `CANCELLED` context that reads as `FAILURE` and gets no next push to clear it - so the group keys on the commit, and on the event as well, so the weekly scan cannot land on the same key as a commit's own analysis:

```yaml
group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}-${{ github.event_name }}
```

That has a second effect worth naming: `codeql.yml` was previously *skipped* by that module's push sweep, because a workflow with no block cancels nothing and the question does not arise for it. It is now inside the sweep and satisfies it per commit, so the push side is graded rather than merely absent. Its docstring named `codeql.yml` as the workflow in that position, which stops being true here, so it is corrected in the same diff and now names `closing-reference.yml`.

### `closing-reference.yml` keeps declaring none

It is the one workflow subscribed to an activity type that cannot change the head sha (`edited`), so it is the only one that can have two runs on one head, and #2216 measured what cancelling one does: the head carried the same check as `SUCCESS` and `CANCELLED` together, and the roll-up read `SUCCESS`, then `FAILURE`, then `SUCCESS` across three reads of one unchanged sha. A cancelled context on a head that satisfies the check needs a push to clear, which defeats the self-clearing property that gate exists to have. A fleet-wide sweep is exactly the kind of change that would tidy that away for uniformity, so the exemption is a table entry that has to cite where the question is settled, and four tests keep it honest: the entry's workflow must still exist, must still be startable by a pull request, must still declare no block, and must still subscribe to a sha-invariant type - otherwise the entry is documenting a fault that no longer exists.

## Verification

**The guard bites.** With the five workflow edits reverted and the test kept, `test_every_pull_request_workflow_supersedes_its_own_previous_run` fails naming all five and prescribing the block. Four vacuity/adversarial tests hand the grader a synthesized fleet, because the sweep passes on this branch and a grader that reported nothing would satisfy it: no block, `cancel-in-progress: false`, a group two pull requests share (`${{ github.workflow }}` alone), and the plausible near-miss - a group keyed on `github.event.pull_request.head.sha`, which reads like a refinement, gives every push its own bucket and supersedes nothing. A fifth asserts the exempt file *would* be reported without its exemption, so a table that stopped being consulted shows up as a failure rather than as a silently passing sweep.

**The parsing helpers are imported from the sibling module rather than re-derived.** It grades the same `concurrency:` blocks from the push side, and two readers of one block can disagree - which would let both pins pass while describing different files. The stated reasons for the text-assertion style (`pyyaml` is optional; a pin that skips when a dep is missing is not a pin) are reasons not to parse YAML, not reasons to keep two copies of a text reader.

**Gate.** Every workflow-config pin in the tree, run against this branch: `382 passed` across `test_pull_request_concurrency_group`, `test_push_concurrency_group`, `test_pull_request_trigger_types`, `test_closing_reference_gate`, `test_codeql_query_filters`, `test_workflow_jobs_are_bounded`, `test_apt_steps_are_bounded`, `test_last_push_approval`, `test_merge_base_overlap`, `test_lockfile_parity_gate`, `test_changelog_fragment_gate`, `test_merge_gate_viewer_scope`, `test_dependabot_config_location`, `test_ref_creation_ruleset_scope`. Changelog and string-hygiene guards: `39 passed`, and `scripts/assemble_changelog.py --check` reports `changelog fragments OK`. `ruff check` and `ruff format --check` clean on the changed files; `mypy` reports no issue in the new module. The five YAML edits are 4 lines each (7 for CodeQL, including the comment), so the reviewable surface is the pin.

## Scope

No job, step, permission or trigger changes: the only thing added to each workflow is when its own previous run stops. `docs.yml` is untouched - its group is already per pull request via `github.ref`, and the split its push side needs is tracked in the sibling module's exemption table as #2305.

Closes #2502

## Round 2 - merge-base overlap cleared

The merge-base overlap gate reported `.github/workflows/agent-api-check.yml` as changed on `main` after this branch diverged at `0acbfc93`. For a branch whose whole subject is workflow configuration this is the overlap worth reading closely rather than clearing reflexively, so both directions were checked:

- The base-side edit is #2493 (`ci: decline apt Recommends...`), which adds `--no-install-recommends` to the `libosmesa6-dev` step. This branch adds only a top-level `concurrency:` block to that file, so the two edits touch disjoint regions and the merge is textually clean.
- The sweep in `test_pull_request_concurrency_group.py` *enumerates* the `pull_request` workflows rather than checking a fixed list, so a workflow newly added on `main` would be newly graded and could fail the branch without either side changing. `git diff --name-status` over `.github/workflows/` across the divergence shows two modifications (`agent-api-check.yml`, `test-lint.yml`) and no additions, so the graded set is unchanged.
- #2493 also landed `tests/test_apt_install_declines_recommends.py`, a new base-side guard that reads the same workflow files this branch edits. Running it against the merged tree is the specific thing neither branch's own checks had done.

`main` is merged in at `bf1b8b34` (which also carries #2496).

Re-run on the merged tree before pushing:

- This branch's pins plus the new base-side guard - `test_pull_request_concurrency_group.py`, `test_push_concurrency_group.py`, `test_apt_install_declines_recommends.py`: **27 passed**.
- Every test in the tree that reads `.github/workflows`, which is the real blast radius for a CI-config change (the 14 modules listed under Verification plus `test_apt_install_declines_recommends`): **392 passed**, 0 failed.

No behaviour change in this round - the diff against the previous head is the merge commit only.


yinsong1986 (approved):
## Summary

Adds a `concurrency` block to the five `pull_request` workflows that declared none (`agent-api-check.yml`, `breaking-change-check.yml`, `dependency-review.yml`, `llm-input-safety.yml`, `codeql.yml`), so a push that replaces a PR's head sha cancels the now-obsolete runs instead of letting them finish ahead of the queue. Four get the bare PR-number-keyed group the other six workflows already carry; `codeql.yml` — the only one of the five that also runs on `push` and `schedule` — keys on `pull_request.number || github.sha` plus the event name, which correctly avoids re-introducing the ref-keyed cancellation defect `tests/test_push_concurrency_group.py` pins (each pushed commit keeps its own group; the weekly scan cannot collide with a commit's own analysis). The new sweep in `tests/test_pull_request_concurrency_group.py` enumerates the fleet rather than a fixed list, distinguishes all four ways a group can fail to supersede (no block, no cancel, over-broad key, head-sha-keyed near-miss), and records the `closing-reference.yml` exemption with four pins that make the exemption self-expiring.

## What's good

- Verified locally at the head sha: the config-pin suites pass (96 tests across the four related modules), and the sweep bites — reverting `dependency-review.yml`'s block alone fails `test_every_pull_request_workflow_supersedes_its_own_previous_run` naming the file and prescribing the fix.
- The four PR-only workflows genuinely have no other trigger, so the bare `pull_request.number` key cannot render empty on an event that reaches it; the one workflow where that would have been a defect (`codeql.yml`) got the event-aware key instead.
- Parsing helpers are imported from the sibling push-side module rather than re-derived, so the two graders cannot disagree about what a `concurrency:` block says; the `_render` model raises on an unmodeled operand rather than silently rendering a constant.
- Both exemption citations (`test_the_gate_does_not_cancel_its_own_run`, `test_an_exempt_workflow_cannot_cancel_its_own_run`) resolve to real test functions at this head.
- Scope discipline: the workflow diffs add only the blocks — no trigger, permission, job or step changes — and the changelog fragment rule (AGENTS.md PR workflow step 3) is followed.

## Verification suggestions

- `git checkout <pre-PR sha> -- .github/workflows/codeql.yml && pytest tests/test_pull_request_concurrency_group.py -q` to see the guard name the file (restore afterwards).
- After merge, spot-check that a superseded CodeQL run on the next multi-push PR actually reads `cancelled` while the merged commit's own push-side analysis runs to completion.

Issue #265, opened 2026-05-29
mesh(perf): cache os.getenv tunables at handler entry to avoid repeated syscalls under lock


cagataycali:
## Origin

PR #225 review thread on `strands_robots/mesh/core.py:1396` (2026-05-29).

## Problem

`_resume_forward_skew_s()` and `_resume_freshness_window_s()` each parse `os.getenv` + run `_parse_positive_float_env` validation on every call. Within `_on_safety_estop` alone they're called 5-6 times per envelope. Two issues:

1. **Internal inconsistency window**: values can change during a single handler invocation if an operator modifies env vars concurrently.
2. **Critical-section duration**: lines 1547-1551 read both env vars while holding `_estop_replay_lock`, extending lock duration by a syscall + regex per env var. The 0.2s corroboration window (R8 fix) is timing-sensitive.

## Proposed fix

Read both into locals at handler entry (around line 1383) and use those for the rest of the method. Same fix applies to `_on_safety_resume`. Cheap, locks down internal consistency, doesn't break the operator-tunable contract (next envelope picks up new value).

## Pin tests required

- `test_tunables_read_once_per_handler_invocation`: monkeypatch os.getenv to count calls; assert at most 1 read per tunable per handler.
- `test_env_change_mid_handler_does_not_affect_current_envelope`: change env after handler entry; assert envelope uses original value.
- `test_lock_held_duration_does_not_include_env_reads`: structural assertion that `_estop_replay_lock` acquisition does not contain `os.getenv` or `_parse_positive_float_env` calls.

## References

- PR #225 (mesh core: replay caches + safety topic handlers, 6/9 of #195 split)
- #195 (parent tracking issue)
- AGENTS.md > Review Learnings (#85) > 'Don't create executors in hot loops' (same principle applied to repeated env reads)

PR #2504, opened 2026-08-19
fix(mesh): resolve the replay-cache tunables before the lock, not inside it

cagataycali:
## Problem

`Mesh` keeps three replay caches, each behind its own lock: e-stop envelopes, resume envelopes, and inbound-command dedup. Every eviction call needs three numbers - a freshness window, a forward-skew tolerance and a size bound - and all three come from lazy resolvers (`_resume_freshness_window_s`, `_resume_forward_skew_s`, `_resume_replay_cache_max`). A resolver is one `os.getenv` plus a validating parse, and when the operator value is unusable it logs as well, so its cost is not constant.

Issue #265 established that this work does not belong in a critical section and hoisted the two float tunables out of the e-stop handler. The hoist did not reach everything:

| lock | holder | resolvers called while held |
|---|---|---|
| `_cmd_replay_lock` | `_exec_cmd` | all three |
| `_estop_replay_lock` | `_on_safety_estop` | the size bound |
| `_resume_replay_lock` | `_on_safety_resume` | the size bound, twice |

The command-dedup cache is the newest of the three and its own comment says it mirrors the other two; it inherited none of the hoist. The two safety handlers comply for the two tunables the rule named and resolve the size bound inside the lock they had just been cleared of.

Measured through 200 real `_exec_cmd` dispatches against a warm 64-entry cache, same script against both trees:

| env | lock held per command | env parses inside the lock |
|---|---|---|
| clean | 6.65 us -> **3.77 us** | 3 -> **0** |
| one operator typo | 35.15 us -> **3.79 us** | 3 -> **0** |

The second row is the point. Before, the duration of the critical section tracked the operator's environment: a single unusable value made every inbound command hold the lock 5.29x longer, because the three refusals were logged from inside it. After, the spread between a clean environment and a typo is 1.01x - the lock hold no longer depends on the environment at all.

lock-hold measurement

## Change

Each site resolves its tunables into locals before taking its lock, and reads the locals inside it:

- `_exec_cmd` resolves the TTL and the bound above `with self._cmd_replay_lock`.
- `_on_safety_estop` already resolved the bound outside the lock to derive `per_issuer_cap`; the eviction call reuses that local instead of resolving it again inside.
- `_on_safety_resume` resolves the bound before its lock and uses it for both the eviction bound and `per_issuer_cap`, which matches how the e-stop site is ordered.

No value changes and no call is dropped - each envelope still resolves each tunable once, so the knobs stay operator-tunable without a restart. What changes is where.

`_resume_freshness_window_s` documented itself as "called only on safety-handler entry, which is bounded by the transport rate cap". Four paths resolve it - presence, command dispatch and both safety handlers - and there is no rate cap on the command path, so the docstring now names its real callers and states the resolve-before-the-lock contract instead.

## Tests

`tests/mesh/test_safety_tunables_cached_at_handler_entry.py` owns this contract. Its static half graded one lock, one method and two tunables - the exact cells the defect avoided. It now derives the lock set from `Mesh` itself and bans every resolver, so a fourth replay cache is graded when it lands rather than inheriting the same gap, plus two behavioural tests that instrument a lock and assert nothing parses the environment while it is held.

Against `upstream/main`: **3 failed / 3 passed**, each failure naming its own sites -

```
_exec_cmd holds ['_cmd_replay_lock'] and calls _resume_freshness_window_s() (line 1139)
_exec_cmd holds ['_cmd_replay_lock'] and calls _resume_forward_skew_s() (line 1139)
_exec_cmd holds ['_cmd_replay_lock'] and calls _resume_replay_cache_max() (line 1142)
_on_safety_estop holds ['_estop_replay_lock'] and calls _resume_replay_cache_max() (line 1989)
_on_safety_resume holds ['_resume_replay_lock'] and calls _resume_replay_cache_max() (line 2360)
_on_safety_resume holds ['_resume_replay_lock'] and calls _resume_replay_cache_max() (line 2374)
```

The 3 that pass on both trees are the controls: the two original issue #265 behaviour tests, and a non-vacuity test asserting the scan really found at least three locks and three methods - without it, collapsing the discovery to an empty set would report a clean tree while grading nothing.

One existing test needed repair. `test_resume_and_estop_per_issuer_cap_use_same_expression` pinned the symmetry of the two fairness bounds by matching a literal copy of the assignment, which this change respells in both handlers. Its stated intent is that the two must not diverge, so it now extracts the assignment each handler carries and compares them to each other. It passes on both trees, and still fails - naming both spellings - when the two are made to diverge.

## Verification

Measured at `1c07700` against `upstream/main` at `0acbfc9`.

- 230-file blast radius (all of `tests/mesh`, every test touching the resolvers, the evictor, `_exec_cmd` or `AGENTS.md`, plus the repo-wide guards), run on both trees: branch **21 failed / 5694 passed**, base **24 failed / 5691 passed**, both collecting 5715. Branch-only failures: none. Base-only: exactly the 3 pre-fix failures above. All 45 shared failures are `tests/mesh/test_iot_*` skipping on an absent `awsiotsdk`.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on both trees with none branch-only and none in the changed files.

## Round 2 - merge-base overlap cleared

The merge-base overlap gate reported `strands_robots/mesh/core.py` as changed on `main` after this branch diverged at `0acbfc93`, so every green check on the branch had been computed against a base predating it. The overlap is #2478 (`fix(mesh): report an absent eclipse-zenoh at the level a consumer sees`), which edits the absent-transport report inside `Mesh.start`. That is a different region of the file from the three replay-cache sites this branch touches (`_exec_cmd`, `_on_safety_estop`, `_on_safety_resume`) and shares no symbol with them, so there is no semantic interaction - only the same-file overlap the gate is defined to flag.

`main` is merged in at `bf1b8b34` (which also carries #2496). The merge is textually clean.

Re-run on the merged tree before pushing, as the gate prescribes:

- `test_safety_tunables_cached_at_handler_entry.py` + `test_resume_replay.py` + the overlap's own `test_zenoh_absent_is_reported.py`: **33 passed**.
- All of `tests/mesh` on the merged tree: **21 failed / 3358 passed**. Every failure is a `tests/mesh/test_iot_*` case on absent `awsiotsdk`, the same environment gap recorded above. No branch-attributable failure.

No behaviour change in this round - the diff against the previous head is the merge commit only.


yinsong1986 (approved):
## Summary

Hoists the three lazy env resolvers (`_resume_freshness_window_s`, `_resume_forward_skew_s`, `_resume_replay_cache_max`) out of the three replay-cache critical sections (`_cmd_replay_lock` in `_exec_cmd`, `_estop_replay_lock` in `_on_safety_estop`, `_resume_replay_lock` in `_on_safety_resume`), completing the issue #265 hoist that had only covered the two float tunables in the e-stop handler. Each site now resolves into a local before its `with` and reads the local inside, so the lock-hold duration no longer depends on the operator's environment (the pathological case being an unusable value that logs three refusals from inside the lock). No behaviour change: each envelope still resolves each tunable exactly once, so the knobs remain operator-tunable without a restart, and a value stays consistent within one envelope's critical section. Verified locally at head `d240731a`: `tests/mesh/test_safety_tunables_cached_at_handler_entry.py` + `tests/mesh/test_resume_replay.py` pass (23 tests), and the three hot-path sites in `core.py` match the description exactly.

## What's good

- The static pin was generalized the right way: the lock set and the methods taking them are derived from the `Mesh` class rather than listed, with a non-vacuity test so an empty discovery cannot report a clean tree - a fourth replay cache is graded on arrival.
- The repaired `test_resume_and_estop_per_issuer_cap_use_same_expression` preserves its stated intent (the two fairness bounds must agree) instead of pinning a literal spelling, and keeps the `// 4` quarter-of-cache floor assertion.
- Two behavioural tests instrument a real lock and assert nothing parses the environment while it is held, so the contract does not rest on the AST scan alone.
- Conventions followed: changelog fragment under `changelog.d/` (not `CHANGELOG.md`), AGENTS.md bullet added beside the rule it extends, ASCII-only strings, no host paths, `monkeypatch.setattr` throughout.
- Round-2 merge-base overlap (#2478) was diagnosed, merged in, and re-verified on the merged tree per the AGENTS.md step-8 prescription.

Issue #2181, opened 2026-08-12
feat(examples/fleet): 02_cross_zone_transport + read-only Rerun fleet dashboard


yinsong1986:
Part of epic #2179 (PR2 of 5). Depends on #2180 (scaffold + `capabilities.py`).

# Scope

`examples/fleet/02_cross_zone_transport.py`: two zone orchestrator agents plus one fleet agent, each an in-process mesh peer (`init_mesh(..., peer_id="zone-...")` - the in-process multi-peer pattern from `notebooks/06_fleet_orchestration.ipynb`). One shared `transport` skill definition is executed in both zones by different robot types. The fleet agent decomposes a cross-zone request into a handoff dispatched via `mesh.send` behind the `robot_mesh` HITL gate.

Plus dashboard v1 - `examples/fleet/dashboard.py`: a READ-ONLY mesh peer subscribing presence/health/safety topics and tailing the signed audit log, rendered in Rerun (already a transitive dependency via lerobot; `require_optional`-gated with a clean install hint):

- fleet table: peer id, zone, presence age, current task, safety state
- event timeline: dispatch / estop / resume / HITL decisions

It attaches to the mesh, not the simulator, so the same dashboard serves every example in the suite and any backend (epic D9). A write-capable UI is an explicit epic non-goal.

# Acceptance

- Suite criteria (CPU-only, seeded, air-gapped, smoke test)
- The same skill artifact executes in both zones with zero per-zone forks of skill code
- Dashboard runs as a separate process/peer, degrades gracefully when rerun is absent, and cannot publish commands (subscribe-only surface)

Board routing: please add to https://github.com/orgs/strands-labs/projects/2 with Status=Todo, Priority=Medium (filing account lacks org project write).

Issue #2179, opened 2026-08-12
epic(examples/fleet): fleet orchestration example suite - skill dispatch, cross-zone transport, failover, emergency evacuation, work-order ingress


yinsong1986:
# Motivation

`examples/` covers single-robot workflows thoroughly (core scripts 01-16, notebooks 05/06), but nothing composes the pieces that already exist - multi-robot worlds (`add_robot` / `run_multi_policy`), the Zenoh mesh with its safety protocol (estop propagation, lockout, HMAC resume), the `robot_mesh` HITL tool, the registry's embodiment metadata, and the `DeclarativeBenchmark` predicate DSL - into fleet-level orchestration examples. The closest artifact is `notebooks/06_fleet_orchestration.ipynb` (NL goal decomposition + mesh dispatch); there is no example of capability-based dispatch, cross-zone coordination, failover, deterministic emergency coordination, or structured task ingress.

# Deliverables

New `examples/fleet/` suite:

| # | Example | Issue | PR |
|---|---|---|---|
| 01 | `skill_dispatch_multi_vendor` | #2180 | PR1 (+ scaffold + README) |
| 02 | `cross_zone_transport` + dashboard v1 | #2181 | PR2 |
| 03 | `failover_and_degraded_ops` | #2182 | PR3 |
| 04 | `emergency_evacuation` | #2183 | PR4 |
| 05 | `work_order_dispatch` | #2185 | PR5 |

Plus: `examples/fleet/README.md`, `dashboard.py` (read-only Rerun fleet dashboard, D9), shared `capabilities.py` schema (consumed by 01 and 05), and an independent docs fix #2184.

# Design decisions (locked during planning)

- **D1 air-gap acceptance**: every example passes with the network fully off, given a pre-populated asset cache (`HF_HUB_OFFLINE=1` + cached robot assets). First-run asset download from GitHub is expected and documented.
- **D2 motion honesty**: 01/02/03/05 use MockPolicy with predicate/timer-based task completion (honest framing). 04's clear-path phase uses scripted deterministic joint/base setpoints - deliberately: the safety layer is deterministic pre-validated code; learned policies are for normal operation only.
- **D3 robot cast**: `so101` (arm) + `lekiwi` (wheeled) + `unitree_go2` (quadruped) - maximally heterogeneous; one-line swap if desired.
[...]

PR #2506, opened 2026-08-19
feat(examples/fleet): dashboard --serve-web serves a live Rerun web viewer for headless hosts

### LINKED ISSUE #2471: feat(examples/fleet): dashboard --serve-web - live Rerun web viewer for headless/remote hosts, loopback-bound by default
# Gap

`examples/fleet/dashboard.py` renders through Rerun only by spawning a **native viewer process** (`RerunRenderer.__init__` -> `rr.init("strands-fleet-dashboard", spawn=spawn)`; flags: `--no-spawn`, `--no-rerun`). On a headless or remote host - which is where a fleet dashboard naturally runs - there is no way to watch the live Rerun view at all. The workarounds verified end-to-end on a headless box:

- terminal renderer (`--no-rerun`): works, live, but tables only;
- post-hoc replay: example 04's `--rrd` artifact served via the CLI (`rerun --serve-web --web-viewer-port ... file.rrd`) works in a browser over an SSH tunnel - but it is replay, not live.

The missing piece is live remote viewing: the mesh half is proven (the read-only peer discovered `fleet-coordinator`/`zone-a`/`zone-b` and tailed the signed audit trail live during example 02), only the render transport is display-bound.

# Proposal

Add `--serve-web` to `dashboard.py`: instead of spawning a native viewer, serve the Rerun web viewer + log stream from the dashboard process (rerun-sdk 0.26 ships the serve API; pick `rr.serve_web()` or the `serve_grpc` + `serve_web_viewer` pair per what 0.26.x actually exposes - verify at implementation, the CLI equivalent is already proven on this data).

Details:

- `--web-port` / `--grpc-port` with sane defaults; print the ready-to-open URL on startup (the `?url=rerun%2Bhttp...` form) plus the one-line SSH tunnel recipe.
- **Bind `127.0.0.1` by default** per the repo's network-exposure convention ("bind to 127.0.0.1 by default, not 0.0.0.0 - users explicitly opt into network exposure"); an explicit `--bind` opts into wider exposure. Note the rerun CLI defaults to 0.0.0.0, so the flag must set the address deliberately rather than inherit it.
- Read-only enforcement is untouched: `restrict_to_subscribe_only` acts on the mesh peer and is orthogonal to the render transport. The web viewer is a *view*; it must not grow any command surface (epic non-goal).
- Graceful degradation unchanged: no rerun-sdk -> terminal renderer with the existing hint.
- README dashboard section gains the headless recipe (`--serve-web` + tunnel), replacing the implicit assumption of a local display.

# Acceptance

- On a headless host: `dashboard.py --serve-web --duration 30` + a fleet example on the same mesh -> the browser view (via tunnel) shows the fleet table updating and audit events arriving live
- Default binding is loopback; `--bind 0.0.0.0` required for anything wider, and says so
- `--serve-web` with rerun-sdk absent fails with the actionable install hint (not a silent fall-through to the terminal renderer, since the user explicitly asked for the web viewer)

Part of epic #2179 (dashboard deliverable #2181). Evidence for the verified workarounds: live terminal capture + served `.rrd` replay from an end-to-end suite run on 2026-08-19 (01/02/04/05 pass live; 03's unrelated failure is #2469).

yinsong1986:
Adds `--serve-web` to `examples/fleet/dashboard.py`: instead of spawning a native viewer window, the dashboard serves the Rerun web viewer + live log stream from its own process, so a browser - locally or over an SSH tunnel - watches the fleet table and audit timeline update live on a headless or remote host.

Closes #2471

## What changed

- **`--serve-web`** starts the rerun-cli web server as a child of the dashboard and streams the render data into it via `rr.connect_grpc`. Startup prints the ready-to-open URL (the `?url=rerun%2Bhttp...` form) and, when loopback-bound, the one-line SSH tunnel recipe (both ports, because the browser fetches the viewer from one and dials the log stream on the other).
- **`--bind` / `--web-port` / `--grpc-port`** with defaults `127.0.0.1` / `9090` / `9876`. Loopback by default per the repo convention; a wider bind prints an explicit exposure warning naming both ports. Passing any of the three without `--serve-web` is refused rather than silently dropped; ports go through the shared `tcp_port_error` domain and must differ.
- **No silent fall-through**: `--serve-web` with rerun-sdk absent raises the `pip install rerun-sdk` hint instead of degrading to the terminal renderer (the explicit ask for a web viewer must not turn into tables nobody is watching). `--no-rerun` + `--serve-web` is refused as contradictory.
- **Lifecycle**: `close()` detaches the SDK stream first (flush against a live server), then terminates the child; the renderer is constructed before the mesh peer so a refused `--serve-web` never leaves a peer announced; `main()` releases both the mesh and the renderer on every path.
- **README** (`examples/fleet/README.md`): headless recipe added to the dashboard section.
- Changelog fragment `changelog.d/2471-dashboard-serve-web.md`.

## Decisions the issue asked to verify at implementation (each measured on rerun-sdk 0.26.2)

- **CLI child, not the SDK serve pair.** `rr.serve_web` is deprecated in 0.26, and `rr.serve_grpc` + `rr.serve_web_viewer` expose **no bind address and bind `0.0.0.0`** (`ss -ltn` shows both listeners on `0.0.0.0` while the returned URI says `127.0.0.1`). The CLI's `--bind` is the only surface that can honour loopback-by-default - and its own default is also `0.0.0.0`, so the address is always set deliberately, exactly as the issue predicted.
- **The child is the native `rerun_cli` binary, not `python -m rerun`.** The wrapper runs the binary via `subprocess.call`, so terminating the wrapper orphans the server: measured, the dashboard exited cleanly and both ports stayed bound by a leaked `rerun_cli/rerun` process.
- **Readiness is the child's own log line, not a TCP probe.** With the gRPC port already taken, a connect probe reaches the squatter and reports ready exactly when the server is not (measured: false pass, dashboard hung streaming into a socket that speaks no gRPC). The 0.26 CLI also does not exit on that bind failure - it logs an ERROR and keeps serving HTML - so the gate recognises both the ready line and the error line, and falls back to a bounded (monotonic) timeout that quotes the child's log.

## Acceptance criteria

- [x] Headless host, live view: verified end-to-end on this (displayless) dev box - `dashboard.py --serve-web --duration 90` + `02_cross_zone_transport.py` on the same mesh (multicast, shared audit dir, HITL auto-approve). A late `rerun rerun+http://127.0.0.1:9876/proxy --save` tap - a headless stand-in for the browser - replayed 11 fleet-table renders and **9 live audit timeline rows** (`handoff_dispatch` -> `handoff_leg_done` -> `handoff_custody` -> ... -> `handoff_refused`), and `curl http://127.0.0.1:9090/` returns the viewer (HTTP 200).
- [x] Default binding is loopback (`ss -ltn`: both listeners on `127.0.0.1`); `--bind 0.0.0.0` required for anything wider and the startup output says so with an explicit warning.
- [x] `--serve-web` with rerun-sdk absent raises the actionable install hint; pinned by `test_serve_web_with_rerun_absent_raises_the_install_hint`.

## Test surface

`tests/test_fleet_cross_zone_transport.py` (+13 tests, 44 total in the file): loopback bind is argv-deliberate and the child is the binary not the wrapper; binary resolution honours/refuses `RERUN_CLI_PATH`; URL + tunnel recipe for loopback and the exposure warning for wider binds; the serve path inits with `spawn=False`, dials the served gRPC port, and `close()` detaches the stream **before** terminating the child (ordering journal); the whole flag domain (`--no-rerun` conflict, options without `--serve-web`, out-of-range ports, equal ports, empty bind) is refused before a renderer, subprocess or mesh peer exists.

Full suite: `32783 passed, 283 skipped` with `MUJOCO_GL=egl` (plus one fragment-format failure fixed before commit); `hatch run lint` clean. Without EGL the run aborts in `tests/benchmarks/libero/test_libero_camera_config_domain.py` - environmental (no GLX on the dev box), passes under EGL, untouched by this diff.

## Out of scope / follow-ups

- The web viewer is a view only (epic #2179 non-goal): no command surface, `restrict_to_subscribe_only` untouched.
- If rerun's log wording ever changes, the readiness gate degrades to the bounded timeout and quotes the child's log - no silent behaviour change; a version bump PR would update the two pinned lines.

Part of epic #2179 (dashboard deliverable #2181). Board: please move #2471 to "In review" on the [Strands Labs - Robots board](https://github.com/orgs/strands-labs/projects/2) (filing account lacks org project write).


cagataycali (approved):
Verified this end to end on a Linux host with `rerun-sdk` 0.26.2 / Python 3.12, because the
load-bearing claim here is a network-exposure default and the suite proves it with a pure argv
builder (`test_serve_web_command_sets_the_loopback_bind_deliberately` asserts the `--bind` flag
string is present) rather than with a socket. The flag being passed and the resulting listener
actually being loopback-only are different facts, so I measured the second one. It holds.

## Sockets, measured (`/proc/net/tcp{,6}`, LISTEN only)

| step | 9090 (web) | 9876 (gRPC) |
|---|---|---|
| before | - | - |
| `RerunRenderer(serve_web=True)`, all defaults | `127.0.0.1` | `127.0.0.1` |
| after `close()` | - | - |

No `0.0.0.0`, no v6 wildcard, on either port. Startup took **0.48s**, so the readiness gate is not
paying anything like its 15s ceiling. `GET http://127.0.0.1:9090/` returned **200** with real
viewer HTML, so the printed `?url=rerun%2Bhttp...` link is openable rather than merely
well-formed. The child is the native binary, not the wrapper:

```
.../rerun_sdk/rerun_cli/rerun --serve-web --bind 127.0.0.1 --port 9876 --web-viewer-port 9090
```

and after `close()` both ports are free and the child is reaped (`/proc/<pid>` gone) - the leak
`rerun_binary()`'s docstring describes does not happen on this path.

## The gate's statement ordering is load-bearing, and I can show it

`_start_web_server` checks `proc.poll()` **before** it looks for the ready line, and the comment
says a ready line found after the child died must not return a dead server. That reads like
defensive paranoia until you occupy the *web* port and leave gRPC free. The child's own log:

```
INFO  re_grpc_server] Listening for gRPC connections on 127.0.0.1:9876 ...
Error: Failed to create server at address 127.0.0.1:9090: Address already in use (os error 98)
```

The ready line **is** there, and the process is already gone. With the two checks swapped this
returns a dead server as ready and the dashboard streams into nothing. As written it refuses in
**0.34s**. The case the gate was actually written for - gRPC port taken - refuses in **0.10s** via
`message proxy server crashed`, with the child still alive serving viewer HTML exactly as
documented, so the timeout never has to catch it. Both directions are covered; only one of them
is the one you'd think of.

## The design rationale reproduces

The `serve_web_command` docstring justifies shelling out by claiming the in-process API binds
`0.0.0.0` while reporting loopback. Measured on 0.26.2:

```
rr.serve_grpc(grpc_port=9392) + rr.serve_web_viewer(web_port=9391)
  listeners -> 0.0.0.0:9391, 0.0.0.0:9392
  returned uri -> rerun+http://127.0.0.1:9392/proxy
```

So the obvious implementation would have exposed the dashboard on every interface while printing
a `127.0.0.1` URL. That is worth having in the docstring, and it is right.

`--bind 0.0.0.0` does bind `0.0.0.0` on both ports and does print the exposure WARNING, so the
opt-in is real in both directions.

## End to end

`dashboard.py --serve-web --duration 6 --interval 2 --web-port 9490 --grpc-port 9491` printed the
URL and tunnel recipe, reported `dashboard peer fleet-dashboard attached (subscribe-only)`, ran
the loop, exited **0**, left **both ports free** and **zero** leaked `rerun_cli/rerun` processes.
Three server lifecycles across my probes and the port table came back empty every time.

## Gate

- PR's own file: **44 passed**.
- Blast radius (143 files: everything touching `examples/`, the tests tree, the changelog/doc
  guards, plus all of `tests/mesh`): **4 failed / 6628 passed / 56 skipped**. Those same 4 fail
  byte-identically on `main` (`tests/test_dds_domain_id_domain.py`,
  `tests/test_dependency_audit.py::test_the_rclpy_refusals_name_the_step_that_supplies_it`,
  2x `tests/test_wait_budget_domain.py`) - they are `rclpy`-present-locally artifacts, so zero
  regressions.
- `ruff check` and `ruff format --check` clean on both changed files. Worth saying explicitly:
  CI's lint scope is `strands_robots tests tests_integ`, so nothing in CI lints `examples/` - I
  ran it here so that is covered rather than assumed.
- `assemble_changelog.py --check` OK.
- `git diff --name-only $MB..HEAD -- strands_robots/` is **empty**, so there is no runtime surface
  to regress in the first place.
- Branch is on current `main` (merge-base == `main`), so the head is the merged tree.

## Two non-blocking notes

1. `mypy examples/fleet/dashboard.py` gains 4 errors (`"object" has no attribute
   "connect_grpc" / "disconnect" / "init" / "__file__"`). They are the same
   `require_optional`-returns-`object` shape as the **6** already on `main`'s version of this
   file, and `examples/` is outside the mypy scope above, so this is the file's existing norm and
   not something this PR introduced the kind of. Mentioning it only so it is on the record.

2. `--bind` is validated for non-emptiness only, so an IPv6 literal produces a URL that needs
   brackets per RFC 3986:

   ```
   --bind ::1  ->  http://::1:9090/?url=rerun%2Bhttp%3A%2F%2F%3A%3A1%3A9876%2Fproxy
   urllib.parse.urlsplit(...) -> ValueError: Port could not be cast to integer value as ':1:9090'
   ```

   `--bind 127.0.0.1 / 0.0.0.0 / 192.0.2.7` all parse fine. Wrapping the host in `[...]` when it
   contains `:` inside `serve_connect_host`/`web_viewer_lines` would cover it, and `::1` also
   takes the "beyond loopback" WARNING branch because the comparison is against the literal
   `"127.0.0.1"`. Both are cosmetic and both fail in the safe direction (unopenable URL,
   over-warning), so I would not hold the PR for either - your call whether it is worth a line.

Approving: the exposure default is loopback-only in fact and not just in flags, both port-conflict
directions refuse promptly with the child's own error, the server is reaped, and the mesh peer
stays subscribe-only.

PR #2510, opened 2026-08-19
fix(examples/fleet): refuse a mesh that did not start instead of awaiting presence

### LINKED ISSUE #2505: feat(examples/fleet): mesh-requiring examples refuse immediately when mesh.alive is False instead of spending the 15s presence timeout
# Symptom

With `eclipse-zenoh` absent, the mesh-requiring fleet examples (02, 03) still spend the
full 15 s presence-discovery window before failing:

```
RuntimeError: timed out after 15s waiting for presence discovery of all robots
```

After #2478 the *cause* is now named - `get_session()` reports the absent dependency at
WARNING and `Mesh.start` names each peer whose mesh stayed off - but the examples
themselves still run to the timeout, and the timeout message still says nothing about a
dependency. #2478 deliberately scoped this half out ("turning it into an immediate
refusal is a behaviour change in those examples and a separate contract"); this issue is
that separate contract, i.e. option 2 of #2477.

# What the examples already check, and what they miss

Both live paths guard the `STRANDS_MESH=0` case but not the zenoh-missing one:

- `examples/fleet/03_failover_and_degraded_ops.py` (`_build_live_fleet`): after each
  `init_mesh(...)` it raises on `mesh is None` ("mesh is disabled (STRANDS_MESH=0);
  rerun with --dry-run") - but `init_mesh` with zenoh absent returns a `Mesh` whose
  `alive` is `False`, which passes that guard and dies 15 s later in `_wait_for`.
- `examples/fleet/02_cross_zone_transport.py` (`_build_live_zones`): same shape.

`mesh.alive` is the documented observable for this state (docs/troubleshooting.md as of
#2478: "`mesh.alive` is `False`, `mesh.peers` stays empty").

# Acceptance criteria

- In both examples' live-fleet builders, after `init_mesh(...)` returns, check
  `mesh.alive`; when `False`, raise immediately with an actionable message naming the
  remedy (`pip install 'strands-robots[mesh]'`) and the `--dry-run` alternative, instead
  of proceeding to the presence wait.
- The existing `mesh is None` guard (mesh disabled on purpose) keeps its own distinct
  message - a deliberate opt-out and a missing dependency need different advice.
- The cleanup path still runs on the refusal (both builders already wrap construction in
  `try/except BaseException: cleanup(); raise`, so raising inside the loop is safe -
  keep it that way).
- Dry-run behaviour unchanged.

# Non-goals

- No change to `strands_robots/mesh/{session,core}.py` - the library-side reporting is
  done (#2478), and quiet-off remains the library contract.
- Examples 04/05 are out of scope unless they share the same live-fleet builder shape;
  check and extend only if the guard is a copy anyway.

Split out of #2477 (option 2 of its two compatible fixes; option 1 shipped in #2478).
Part of epic #2179.

Board routing: please add to https://github.com/orgs/strands-labs/projects/2 with
Status=Ready (filing account lacks org project write).

cagataycali:
Closes #2505.

## Problem

`init_mesh` returns `None` **only** when the mesh is switched off on purpose. When the
mesh is enabled but no session opens - `eclipse-zenoh` absent, or any of the other paths
on which `get_session` yields nothing - it returns a `Mesh` whose `alive` is `False`.
That peer publishes no presence and discovers none, so it is not a slow peer: every wait
below it can only expire.

The four live-fleet builders checked only `is None`, so a dead peer passed the guard.
`mesh.alive` is the observable `docs/troubleshooting.md` documents for this state, and
`examples/fleet/dashboard.py` already refuses on it with the remedy named - one of five
peer starts in `examples/fleet/` got it right.

## Measured

With `import zenoh` failing and nothing else changed. Both columns come from one capture
script run against each tree, so the numbers are comparable; a cold process adds ~2.5 s
of world construction to both sides.

| example | `upstream/main` | this branch |
|---|---|---|
| 02 cross-zone | presence wait, **15.84 s**: `timed out after 15s waiting for presence discovery of both zone peers` | builder, **0.49 s**: names peer `'zone-a'`, `mesh.alive is False`, the extra, `--dry-run` |
| 03 failover | presence wait, **15.49 s**: `timed out after 15s waiting for presence discovery of all robots` | builder, **0.34 s**: names peer `'go2-a1'` + remedy |
| 04 evacuation | presence wait, **15.97 s**: `timed out after 15s waiting for presence discovery of the fleet sim peer` | builder, 15.84 s: names peer `'evac-fleet-sim'` + remedy |
| 05 work orders | **not stopped - exit 0** after 0.55 s | builder, 0.63 s: names peer `'fleet-sites'` + remedy |

None of the three timeout messages mentions a dependency.

**05 is the worst case, not the mildest.** It has no presence wait, so every RPC came back
`{'status': 'error', 'error': 'mesh not running'}`, the whole queue ran, and it wrote four
events - three `work_order_failed`, one `work_order_nacked` - whose first reason is
`code='dispatch_failed' robot='so101-a1'`. It then printed a plausible audit
reconstruction, reported `audit integrity: ok=True`, and **exited 0**. An operator reads
that as a fleet that refused every order; automation reads a successful run.

## Change

Each builder checks `mesh.alive` after `init_mesh` returns and refuses, naming the peer,
`mesh.alive`, `pip install "strands-robots[mesh]"` and the `--dry-run` alternative. The
deliberate `STRANDS_MESH=0` guard keeps its own distinct message: an install remedy cannot
fix a switch, so the two states need different advice.

Cleanup is preserved on the new path - 02/03/04 raise inside the existing
`try/except BaseException: cleanup(); raise`, and 05 (which has no such wrapper) repeats
the explicit teardown its neighbouring guards already do.

The test helper that drives those builders returns the refusal itself rather than handing
back whatever it caught, so its own leak-catching clause ends in
`raise AssertionError(...) from exc` - the shape the tree's two other leak-catching test
helpers use, and the shape those `cleanup(); raise` wrappers use. Every refusal here is a
`RuntimeError`, on the `mesh.alive` path and on the opt-out alike, so a builder that raises
anything else is reported where the diagnosis is built instead of being handed back for an
`isinstance` assertion to classify.

## Scope

- **Examples 04 and 05 are included** because the guard was a verbatim copy: the string
  `"mesh is disabled (STRANDS_MESH=0); rerun with --dry-run"` appeared in all four files.
- **04's wall time is essentially unchanged** (15.97 s -> 15.84 s) and the table says so.
  Its refusal is raised ~1.3 s in, but the builder starts five policies before it touches
  the mesh, so tearing those down costs about what the presence wait did. What changes for
  04 is the diagnosis, not the duration.
- **03's mid-run orchestrator restart is guarded too.** It is not only reachable via an
  absent extra: `get_session` has six paths that yield no session, so a restart can return
  a dead peer on a host where zenoh is installed, and the wait below it is the same 15 s
  presence wait. Its message omits `--dry-run` (you cannot switch mid-run), matching the
  distinct message that site already had.
- No library change. The reporting side is already done and quiet-off remains the library
  contract.

## Tests

`tests/test_examples_refuse_a_mesh_that_did_not_start.py`, **9 failed / 7 passed against
`upstream/main`, 16 pass here**.

Behavioural: each real builder is driven with `get_session` returning `None` and must
refuse actionably. World construction is stubbed (only the mesh is under test), which is
why 04's base failure reports `raised AttributeError (...) instead of refusing` rather
than a returned fleet - the end-to-end measurement above covers 04's timeout, and the
structural rule below names it directly.

Structural, derived from the scripts rather than hand-listed so a sixth peer start is
graded on arrival: every function that starts a peer and guards it for `None` must also
refuse on `alive`. On `upstream/main` it reports all five offenders
(`02:_build_live_zones`, `03:_build_live_fleet`, `03:_run_live`, `04:_build_live_world`,
`05:_build_live_transport`); here it reports none. It ships with a non-vacuity test, a
planted `None`-only guard, and a clean control that is the shape `dashboard.py` already
uses.

The 7 tests that pass on both trees are the controls: the `STRANDS_MESH=0` refusal still
names the switch and must not offer an install remedy.

## Verification

78-file selection (every test that reads `examples/`, walks the test tree, names the fleet
examples, `init_mesh` or `BaseException`, plus the repo-wide docs/changelog/docstring/ASCII
guards), re-measured at `e0de1cbc`:

- branch **2275 passed, 0 failed**, 54 skipped
- base `14f49f41` with the new test file copied in: 2266 passed, **9 failed**, 54 skipped
- `2266 + 9 == 2275`, 2329 collected on both, no branch-only failure, and no pre-existing
  failure in the selection
- the same selection on the previously reviewed head reports 2275 passed / 0 failed too, so
  the helper's rewrite changes no verdict anywhere
- `mypy strands_robots tests tests_integ`: 24 == 24 against `main`, none in the changed files
- `ruff check` / `ruff format --check` clean; `scripts/assemble_changelog.py --check` OK

Based on `14f49f41` so the diff is exactly these seven files - `main` is one CI-only
commit ahead.

## Artifact

fleet examples refuse a mesh that did not start

Both columns are real program output from the capture script above; the figure asserts its
own headline claims against the captured JSON.

### DIFF
--- a/changelog.d/2510-fleet-examples-refuse-a-dead-mesh.md
+++ b/changelog.d/2510-fleet-examples-refuse-a-dead-mesh.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **examples/fleet**: refuse a mesh that did not start instead of proceeding to the
+  presence wait. `init_mesh` returns `None` only when the mesh is switched off on
+  purpose; when it is enabled but no session opens it returns a `Mesh` whose `alive`
+  is `False`, which passed the four live-fleet builders' `is None` guard. Examples 02,
+  03 and 04 then spent their 15 s presence-discovery timeout on a message naming no
+  dependency, and 05 - which has no presence wait - ran its whole queue, attributed
+  every order to a per-robot `dispatch_failed`, and exited 0. Each builder now checks
+  `mesh.alive` (the observable `dashboard.py` already refuses on) and names the peer,
+  the remedy and the `--dry-run` alternative. The deliberate `STRANDS_MESH=0` opt-out
+  keeps its own distinct advice.
--- a/examples/fleet/02_cross_zone_transport.py
+++ b/examples/fleet/02_cross_zone_transport.py
@@ -374,11 +374,24 @@ def _build_live_zones() -> tuple[Any, Callable[[], None]]:
             mesh = init_mesh(sim, peer_id=zone_id, peer_type="sim")
             if mesh is None:
                 raise RuntimeError("mesh is disabled (STRANDS_MESH=0); rerun with --dry-run")
+            # A peer whose mesh did not start is not a slow peer: it publishes no
+            # presence and discovers none, so the presence wait below can only
+            # expire.  Refuse here, where the cause is still known.
+            if not mesh.alive:
+                raise RuntimeError(
+                    f"mesh did not start for peer {zone_id!r} (mesh.alive is False): install the mesh "
+                    'extra with pip install "strands-robots[mesh]", or rerun with --dry-run'
+                )
             sim.mesh, sim.peer_id = mesh, mesh.peer_id
             meshes.append(mesh)
         coordinator = init_mesh(_Coordinator(), peer_id=COORDINATOR_ID)
         if coordinator is None:
             raise RuntimeError("mesh is disabled (STRANDS_MESH=0); rerun with --dry-run")
+        if not coordinator.alive:
+            raise RuntimeError(
+                f"mesh did not start for peer {COORDINATOR_ID!r} (mesh.alive is False): install the mesh "
+                'extra with pip install "strands-robots[mesh]", or rerun with --dry-run'
+            )
     except BaseException:
         cleanup()
         raise
--- a/examples/fleet/03_failover_and_degraded_ops.py
+++ b/examples/fleet/03_failover_and_degraded_ops.py
@@ -487,11 +487,24 @@ def _build_live_fleet() -> tuple[dict[str, Any], Any, Callable[[], None]]:
             mesh = init_mesh(sim, peer_id=robot_id, peer_type="sim")
             if mesh is None:
                 raise RuntimeError("mesh is disabled (STRANDS_MESH=0); rerun with --dry-run")
+            # A peer whose mesh did not start publishes no presence and discovers
+            # none, so the presence wait below can only expire.  Refuse here,
+            # where the cause is still known.
+            if not mesh.alive:
+                raise RuntimeError(
+                    f"mesh did not start for peer {robot_id!r} (mesh.alive is False): install the mesh "
+                    'extra with pip install "strands-robots[mesh]", or rerun with --dry-run'
+                )
             sim.mesh, sim.peer_id = mesh, mesh.peer_id
             meshes[robot_id] = mesh
         orch_mesh = init_mesh(_Orchestrator(), peer_id=ORCHESTRATOR_ID)
         if orch_mesh is None:
             raise RuntimeError("mesh is disabled (STRANDS_MESH=0); rerun with --dry-run")
+        if not orch_mesh.alive:
+            raise RuntimeError(
+                f"mesh did not start for peer {ORCHESTRATOR_ID!r} (mesh.alive is False): install the mesh "
+                'extra with pip install "strands-robots[mesh]", or rerun with --dry-run'
+            )
     except BaseException:
         cleanup()
         raise
@@ -576,6 +589,11 @@ def _run_live(n_steps: int) -> tuple[dict[str, Any], dict[str, Any], list[dict[s
         orch_mesh2 = init_mesh(_Orchestrator(), peer_id=ORCHESTRATOR_ID)
         if orch_mesh2 is None:
             raise RuntimeError("mesh is disabled (STRANDS_MESH=0)")
+        if not orch_mesh2.alive:
+            raise RuntimeError(
+                f"the restarted {ORCHESTRATOR_ID!r} peer did not join the mesh (mesh.alive is False), "
+                "so the presence wait below cannot succeed"
+            )
         try:
             _wait_for(
                 lambda: all(robot in orch_mesh2.peers_by_id for robot in survivors),
--- a/examples/fleet/04_emergency_evacuation.py
+++ b/examples/fleet/04_emergency_evacuation.py
@@ -846,6 +846,15 @@ def _build_live_world() -> tuple[MujocoEvacuationWorld, Any, Callable[[], None]]
         coordinator = init_mesh(_Coordinator(), peer_id=COORDINATOR_ID)
         if sim_mesh is None or coordinator is None:
             raise RuntimeError("mesh is disabled (STRANDS_MESH=0); rerun with --dry-run")
+        # A peer whose mesh did not start publishes no presence and discovers
+        # none, so the presence wait below can only expire.  Refuse here, where
+        # the cause is still known.
+        for peer_id, peer in ((FLEET_PEER_ID, sim_mesh), (COORDINATOR_ID, coordinator)):
+            if not peer.alive:
+                raise RuntimeError(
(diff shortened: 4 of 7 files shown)


yinsong1986 (dismissed):
## Summary

Adds a `mesh.alive` refusal after every `is None` guard in the four live-fleet builders (02/03/04/05, plus 03's mid-run orchestrator restart), so a mesh that was enabled but never opened a session is refused at start-up - naming the peer, the observable, the `strands-robots[mesh]` remedy and the `--dry-run` alternative - instead of expiring a 15 s presence wait (02/03/04) or running the whole queue and exiting 0 (05). The deliberate `STRANDS_MESH=0` opt-out keeps its distinct message. I verified the semantics against `Mesh.start` in `strands_robots/mesh/core.py`: on the no-session path it returns before `_has_session_ref` is set, so a dead mesh holds nothing to stop, and every cleanup path here correctly guards `stop()` on `alive` - 02/03/04 raise inside the existing `try/except BaseException: cleanup(); raise` wrappers and 05 repeats the explicit teardown its neighbouring guards already do. The new test file passes 16/16 against the PR tree, and the diff is exactly the seven claimed files off `14f49f41`.

## What's good

- The structural test derives its offender set from the scripts (`init_mesh` callers with a `None`-only guard) rather than hand-listing them, and ships a non-vacuity check, a planted offender, and a clean control matching `dashboard.py`'s existing shape.
- The refusal messages distinguish the deliberate opt-out (switch advice) from the failed start (install advice), and 03's mid-run restart correctly omits `--dry-run`.
- Changelog fragment present (`changelog.d/2510-*`), README env table updated in the same PR, all new strings ASCII, no host paths.

## Verification suggestions

- `python -m pytest tests/test_examples_refuse_a_mesh_that_did_not_start.py -q` on the branch (16 pass) and on `main` with the test file copied in (9 fail, 7 control passes) reproduces the claimed split.
- With `eclipse-zenoh` uninstalled, `python examples/fleet/05_work_order_dispatch.py` should now exit non-zero naming `'fleet-sites'` rather than writing `dispatch_failed` events and exiting 0.

yinsong1986 (approved):
## Summary

Examples-only fix: the four live-fleet builders (02/03/04/05, plus 03's mid-run orchestrator restart) gain a `mesh.alive` refusal after their existing `is None` guard, so a mesh that was enabled but never opened a session is refused at start-up - naming the peer, the observable, the `strands-robots[mesh]` remedy and the `--dry-run` alternative - instead of expiring a 15 s presence wait (02/03/04) or, worst, running the whole queue and exiting 0 with every order misattributed to a per-robot `dispatch_failed` (05). The premise checks out against `strands_robots/mesh/core.py`: `init_mesh` returns `None` only on the deliberate opt-out and otherwise returns a `Mesh` whose `alive` can be `False`, and the new guards mirror the shape `dashboard.py` already ships. No library change, no public API or wire-format surface touched.

## What's good

- Cleanup is preserved on every new refusal path: 02/03/04 raise inside the existing `try/except BaseException: cleanup(); raise` wrappers (whose `cleanup()` correctly skips `stop()` on a non-alive mesh), and 05 repeats the explicit `sim.destroy()` / `sim_mesh.stop()` teardown its neighbouring guards use - consistent with AGENTS.md > Review Learnings (#86) > Resource Cleanup on Partial Failure.
- The new test file is both behavioural (drives the real builders with `get_session` patched to `None`) and structural (AST-derived over every peer-starting function, with a planted-offender non-vacuity check and a clean control matching the `dashboard.py` shape), so a future peer start is graded on arrival. Ran locally at `e0de1cbc`: 16/16 pass.
- The deliberate `STRANDS_MESH=0` opt-out keeps its distinct message, and the controls assert the install remedy never appears there - the two states need different advice and the tests pin that.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; README env-var table updated in the same PR; all new strings are plain ASCII.

## Verification suggestions

- `python -m pytest tests/test_examples_refuse_a_mesh_that_did_not_start.py -q` at the head - 16 pass; copying the file onto `main` reproduces the 9 claimed base failures.
- To reproduce the end-to-end behaviour: in an env without `eclipse-zenoh`, `python examples/fleet/05_work_order_dispatch.py` should now refuse at start-up naming `'fleet-sites'` instead of exiting 0 after dispatch-failing the whole queue.

PR #1763, opened 2026-07-30
fix(sim): add_robot no longer rewinds the scene it is added to

cagataycali:
## Problem

`add_robot` owes two things at once: the robot it just merged in needs a defined starting configuration (the recompile leaves its new joints wherever the compiler put them), and the scene it joined must be left alone. Its own docstring promises the second:

> Instead of replacing the entire world model, this method merges the robot's bodies, actuators, assets, and sensors into the existing scene XML. **This preserves previously-created world state (gravity, objects, cameras, other robots).**

Both were met with a single `mj_resetData` over the whole world, so the first cancelled the second. Measured on a live scene (one Panda parked with `send_action`, a 0.5 kg crate settled on the floor, 901 steps elapsed), then adding a second Panda:

| | before `add_robot("b")` | after, on `main` | after, this PR |
|---|---|---|---|
| arm `a` joint1..3 | `0.900, -0.347, 0.451` | `0.000, 0.000, 0.000` | `0.900, -0.347, 0.451` |
| arm `a` servo setpoints | holding the pose | released | holding the pose |
| crate z | `0.0699 m` (at rest) | `0.5500 m` (back at its spawn) | `0.0699 m` |
| clock | `t=1.802 s, step 901` | `t=0.000 s, step 0` | `t=1.802 s, step 901` |
| `add_robot` status | | `success` | `success` |

The pose loss is not momentary: the setpoints go with it, so the arm then collapses under gravity over the following steps (1.74 rad of travel on one joint in 800 steps, uncommanded). A latched `apply_force` wrench is dropped the same way, and a checkpoint's worth of elapsed time disappears from the clock every recorded timestamp is taken from.

The world-wide reset is also what forced the home-pose re-apply that used to follow it, and the comment above that call says so — a partial repair that only covered robots spawned with `keyframe=`, and that *overwrote* wherever such a robot had since been driven (measured: an arm spawned at `stow` and then driven to `0.900` came back at `1.200`).

This is reachable from the documented way to compose a scene — `create_world` → `add_robot` → `add_object` → `add_robot` — and from adding a robot mid-session after a rollout.

## Change

Scope the reset to the robot being added. A new `_reset_robot_to_reference` walks that robot's own joints and actuators: `qpos` slices come from `model.qpos0` (the reference configuration `mj_resetData` writes, read per joint so nothing else is touched), `qvel` and `ctrl` go to zero, and an actuator carrying its own activation state has that cleared too. Nothing outside the robot is written.

[...]

PR #2511, opened 2026-08-19
fix(simulation): refuse a negative predicate tolerance instead of compiling an unsatisfiable clause

cagataycali:
## Problem

Every tolerance in the predicate DSL is compared against a distance, an absolute difference or a squared magnitude - quantities that are never negative - so a negative tolerance is not a *looser* bound, it is an *unsatisfiable* one.

`body_upright` and `base_tipped` already refused `tol < 0` inside their factories. The other four tolerance params the registry declares did not, and each compiled clean at a negated value and then evaluated `False` at every pose:

| predicate | param | positive value | negated |
|---|---|---|---|
| `body_upright` | `tol` | `True` | refused |
| `base_tipped` | `tol` | `False` | refused |
| `distance_less_than` | `threshold` | `True` | **accepted -> permanently `False`** |
| `body_inside` | `xy_tol` | `True` | **accepted -> permanently `False`** |
| `body_inside` | `z_tol` | `True` | **accepted -> permanently `False`** |
| `body_on` | `xy_tol` | `True` | **accepted -> permanently `False`** |

Measured through the real caller route, `run_policy(stop_when=...)`, with a clause that *is* satisfied at `threshold=+0.35` (it stops the rollout at step 3):

```
stop_when={"predicate": "distance_less_than", "body_a": "so101/gripper",
           "body_b": "cube", "threshold": -0.35}

before  ->  status="success"   Policy complete on 'so101' | 120 steps | sim_t=4.320s
after   ->  status="error"     predicate 'distance_less_than': threshold must be >= 0, got -0.35. ...
```

That `120 steps` is the whole budget spent on a clause that could never fire. It is the same permanently-`False` outcome a `nan` threshold produces - the failure mode the finiteness guard in `make_predicate` already exists to prevent - reached by a route that was not checked. Unlike `nan` it also reads as a *wider* bound, so nothing about the spec looks wrong.

## Change

Holds the non-negative domain in `make_predicate`, the one choke point every clause passes through - including the per-stage `reward` / `advance_when` calls a `staged_reward` compiles by calling back into it, which a guard in the spec compiler would miss.

The domain is read from the param **name**, for the same reason the existing finiteness domain is read from the **annotation**: a predicate added later is covered by naming its tolerance the way every shipped one already does, with no per-predicate table to drift out of step with the registry. Matched on the whole name (`tol`, `threshold`) or a `_tol` suffix (`xy_tol`, `z_tol`), never as a substring.

Signed params keep both signs, because their sign is part of the value - `body_on`'s `z_offset` is an offset a caller lowers below zero to accept a body resting slightly *under* the reference, `base_velocity`'s `vx` is a velocity component whose sign is a direction, and `body_below_z`'s `z` is a coordinate. A guard over every numeric kwarg would have broken all three, so each is pinned.

The sign check reads `float(value)`, which made `_kwarg_domain_error` match the discovery signature in `tests/simulation/test_input_validators_refuse_a_boolean.py` for the first time - that scan finds every validator under `simulation/` that returns a *reason* and coerces with a bare `float()`, and this function had no coercion before. The coercion is safe: it compares a value `finite_number_error` has already accepted, and that shared domain refuses `bool` and `numpy.bool_` explicitly, so a boolean is answered with its own reason and never reaches it. It is recorded there as an exemption, alongside the three entries of the same shape already present (`dataset_rate_mismatch_reason`, `rollout_rate_mismatch_reason`, `_workspace_sanity_error`), rather than adding a second boolean check that `finite_number_error` already performs and which would be unreachable.

Those exemptions are prose that nothing verifies, so this one does not rest on prose. Thirty cases assert that every boolean spelling of a numeric predicate kwarg is refused, and they are measured against the thing they claim: deleting the `finite_number_error` call makes 26 of them report `DID NOT RAISE` - `tol=True` accepted as a tolerance of `1.0`, since `float(True)` is not an error - while the four `still accepted` cases keep passing. They deliberately claim no *ordering*: moving the sign check above the finiteness guard changes no verdict, because the sign check only refuses a negative value and a boolean is never negative.

`0` stays accepted, which is the boundary `body_upright` / `base_tipped` have admitted since they were written (`tol < 0`, not `<= 0`). Whether a zero-width tolerance is itself worth refusing is a question about those two shipped guards, not this one, and is left alone.

## Verification

Measured at `ee3e452` against base `b498fe71`.

**Pre-fix split** (both test files copied onto `upstream/main`): **11 failed / 307 passed** -> 318 pass.

Nine failures are behavioural (`DID NOT RAISE` for the four unguarded params, on both the refusal and the message assertions, plus a later-registered predicate); one names the new helper, in the drift test that keeps the module's rule and the test's restatement of it in step. The eleventh is the enumeration's own staleness check, which reports `exemptions no longer match a discovered validator: ['_kwarg_domain_error']` on base - the function carries no coercion there, so it is not discovered and the exemption would be stale. That is the exemption invalidating itself if the coercion ever goes away, which is the property that makes recording it safe.

The 307 that pass on both trees are the no-overreach half: every signed param still accepts a negative, `z_offset=-0.05` still evaluates `True` on a real scene, `0` is still accepted, a `nan` tolerance still reports *finiteness* rather than being displaced by the sign check, a param whose name merely contains the letters (`tolerance`, `threshold_scale`, `control`) keeps its signed domain, and every boolean spelling of a numeric kwarg is refused with the finiteness reason rather than the sign one.

**Blast radius** - 88 test files (every test touching `make_predicate`, `PREDICATE_REGISTRY`, `compile_stop_when`, `stop_when`, `is_boolean` or the shared numeric domains, plus the repo-wide docs/docstring/changelog guards), run on both trees:

```
branch    8 failed / 6350 passed / 22 skipped
base     19 failed / 6339 passed / 22 skipped
8 + 6350 == 19 + 6339 == 6380 collected on both
branch-only failing ids: none;  base-only: exactly the 11 pre-fix ids above
```

The 8 failures shared by both trees are `tests/mesh/test_iot_connect_timeout_domain.py` needing `awsiot`, which is a declared dependency and so present in CI.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 with none in the changed files.

## Rollout

Same MuJoCo rollout in both trees - the gripper-cube distance trace is **bit-identical** across them (`max|delta| = 0.0`), so the only difference is whether the clause compiles. The shaded half-plane is the region a distance can never enter, which is why the negative threshold can never be crossed.

predicate tolerance sign domain

---

## S13 - Review Round Changelog

| Round | Concern | Fix Commit | Pin Test |
|---|---|---|---|
| R1 | `"str" in annotation` substring match also matches `list[str]`, handing a bare string to a `name_list_error` factory (semantic composition conflict with #2483 pour predicates) | `c2969d4` | `tests/simulation/test_predicate_tolerance_sign_domain.py` - `test_every_probe_value_has_the_shape_its_annotation_declares` and `test_every_graded_predicate_can_actually_be_built` |
| R1 | Merge upstream/main to absorb #2483 and verify composition is green | `1bea0b6` | same file, `test_zero_stays_accepted[particles_*]` (6 cases) |
| R1 | Stale probe-table comment referencing counts the registry moved | `adb2cb4` | -- |
| R1 | Head was behind `main` again after #2515/#2518/#2519 landed, so CI was not grading the composition the squash produces (head-checkout, AGENTS.md > PR Workflow step 8) | `0a69634` (base merge only, `git show --cc` empty, PR diff unchanged at 5 files) | same file - 629 passed across the predicate blast radius on the composition, including `tests/simulation/test_pour_predicates.py` |


yinsong1986 (dismissed):
## Summary

Extends the existing compile-time numeric-domain guard in `make_predicate` (`_kwarg_domain_error`) so that a kwarg naming a tolerance (`tol`, `threshold`, or a `_tol` suffix) is additionally refused when negative, closing the four registry params (`distance_less_than.threshold`, `body_inside.xy_tol`/`z_tol`, `body_on.xy_tol`) that previously compiled clean at a negated value and evaluated permanently `False` — the same unsatisfiable-clause outcome the finiteness guard already prevents for `nan`, reached by a route that read as a *looser* bound. The guard sits after `finite_number_error`, which already rejects bools, non-Reals and values beyond float64 range, so the `float(value) < 0.0` comparison is safe on everything that reaches it. The refusal is an explicit `ValueError` naming the predicate, parameter and value at authoring time, so nothing about older specs is silently invalidated.

## What's good

- The name-based match is whole-name-or-suffix, never substring, and the test file pins that boundary from both directions (`tolerance`/`threshold_scale`/`control` keep their signed domain; every signed param in the registry is asserted unclassified).
- The registry-derived sweep plus the `_MINIMUM_TOLERANCE_PARAMS` reach assertion means the parametrized cases cannot go vacuous if the registry shifts, and `test_module_rule_matches_the_pinned_rule` keeps the test's restated rule and the module's `_is_tolerance_param` from drifting — the drift-guard shape AGENTS.md asks for.
- `0` stays accepted, matching the boundary the two pre-existing factory guards (`body_upright`/`base_tipped`, `tol < 0`) have admitted since they were written, so the whole family lands on one domain.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; docs updated in the same PR; no host paths, no emojis, guard kept local to its single caller per AGENTS.md rule 11.

tests/simulation/test_predicate_tolerance_sign_domain.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    annotations = getattr(factory, "__annotations__", {})
+    kwargs: dict[str, Any] = {}
+    for p in inspect.signature(factory).parameters.values():
+        if p.default is not inspect.Parameter.empty:
+            continue
+        annotation = str(annotations.get(p.name, ""))
+        if annotation in _NUMERIC_ANNOTATIONS:
+            kwargs[p.name] = 0.25
+        elif "str" in annotation:
```
yinsong1986: [MUST FIX] `elif "str" in annotation:` is a substring match, so it also matches `list[str]` and supplies a bare string where a list is required - and that combination is already live on `main`. #2483 (merged ~17 minutes after this head commit, so absent from this branch's base) registered `particles_inside`, `particles_spilled` and `particles_inside_fraction`, each declaring `xy_tol`/`z_tol` plus a required `particles: list[str]` param guarded by `name_list_error`. The registry-derived sweep correctly picks up the six new tolerance cases, but this helper hands the factory `"probe_particles"`, and the factory refuses it ("must be a list of names, not a single string"). Measured on the composition (`git merge upstream/main` - textually clean, 0 conflicts): **6 failed / 104 passed** in this file, all `test_zero_stays_accepted[particles_*]`. The negative-refusal cases pass only because `_kwarg_domain_error` runs before the factory; `test_zero_stays_accepted` must reach the factory and cannot.

Why this can't be deferred: `pr-and-push.yml` checks out the PR **head** commit, not the merge ref (AGENTS.md > PR Workflow step 8), so CI on this branch stays green and the failure surfaces only on `main`'s required `call-test-lint` check after the squash - the exact #1763/#1766 semantic-conflict class AGENTS.md documents for this same file, and a red `main` blocks every subsequent merge until fixed.

Fix: handle the list annotation before the substring match, e.g.
```python
if annotation in _NUMERIC_ANNOTATIONS:
    kwargs[p.name] = 0.25
elif annotation == "list[str]":
    kwargs[p.name] = [f"probe_{p.name}"]
elif "str" in annotation:
    kwargs[p.name] = f"probe_{p.name}"
```
then merge `main` into the branch and re-run this file to confirm the composition is green (the production guard needs no change - verified the new pour-predicate tolerances are refused/accepted correctly through it).

yinsong1986 (commented):
## Summary

Extends the compile-time numeric-domain guard in `make_predicate` (`_kwarg_domain_error`) so a kwarg naming a tolerance (`tol`, `threshold`, or a `_tol` suffix, matched whole-name-or-suffix only) is refused when negative, closing the four registry params that previously compiled clean at a negated value and evaluated permanently `False` while reading as a *wider* bound. The guard sits at the single choke point every clause passes through (including nested `staged_reward` calls), after `finite_number_error`, so the `float(value) < 0.0` read is safe. The production change is sound and the registry-derived test sweep is the right drift-guard shape - it correctly picks up new tolerance params on arrival, which is exactly how I found the one blocker: composed with current `main` (which gained the #2483 pour predicates after this branch's base), the sweep reaches the new `xy_tol`/`z_tol` params but the test's kwarg-builder helper mis-handles their required `list[str]` param, and 6 tests fail on a textually clean merge. Details inline.

## Verification suggestions

Reproduce the composition failure without merging: `git merge upstream/main --no-commit --no-ff` on the branch (merges clean, 0 conflicts), then `pytest tests/simulation/test_predicate_tolerance_sign_domain.py -q` -> 6 failed (`test_zero_stays_accepted[particles_*]`), 104 passed. On the branch's own tree the same file is 68/68 green, and the production guard itself composes correctly - a negative `xy_tol` on the new pour predicates is refused as intended.

tests/simulation/test_predicate_tolerance_sign_domain.py
cagataycali: Confirmed and fixed in `c2969d4`. Reproduced your numbers exactly on the composition: **6 failed / 104 passed**, all `test_zero_stays_accepted[particles_*]`, each on `particles must be a list of names, not a single string, got 'probe_particles'`.

I took the whole-annotation match rather than the ordered `elif`, because the substring test is wrong in a second way the ordered version leaves in place. The required-param annotations the shipped registry declares are `str` (21), `float` (12), `list[str]` (4), `list[float]` (2) and `list[Any]` (1) - so the old helper handed 4 params a wrong value *and supplied nothing at all* for the other 3. Those 3 are unreached today only because `inside_region` and `staged_reward` carry no tolerance; a tolerance added to either would have produced a bare `TypeError: missing a required argument` instead. The placeholder is now keyed on the whole annotation:

```python
_PROBE_VALUES: dict[str, Any] = {
    "float": lambda param: 0.25,
    "int": lambda param: 1,
    "str": lambda param: f"probe_{param}",
    "list[str]": lambda param: [f"probe_{param}"],
    "list[float]": lambda param: [0.0, 0.0, 0.0],
}
```

and the two halves are asserted in `TestTheRuleIsMatchedOnAWholeNameNotASubstring`, which already pinned exactly this rule for the param *name* - the annotation axis is the same rule, one level over:

- `test_every_probe_value_has_the_shape_its_annotation_declares` - a `list[...]` annotation gets a `list`. Against the previous head's helper it fails naming all four offenders (`particles_inside.particles`, `particles_inside_fraction.particles`, `particles_spilled.particles`, `particles_spilled.containers`).
- `test_every_graded_predicate_can_actually_be_built` - every predicate a tolerance case points at is fully supplied. This one *passes* on the previous head (the substring helper did supply a value, just the wrong one), so it is the no-overreach half; deleting `list[str]` from the table makes it report `no probe value for ['particles']`, which is how I checked it is load-bearing.

Merged `main` in - textually clean, 0 conflicts, `predicates.py` the one file both sides touched, and both halves compose. The file is **112 passed** on the composition, and the pre-fix count for the PR as a whole went 11 -> 23 with no test change, because the registry-derived sweep picked up your six new tolerance params on its own. As you found, the production guard needed nothing: 18 of 18 cases correct across the three new predicates (refused at `-0.2`, accepted at `0` and `0.15`).

Blast radius re-run on the composition, 64 paths on both trees: branch `0 failed / 12988 passed`, base `23 failed / 12965 passed`, `0 + 12988 == 23 + 12965 == 13168` collected on both, no branch-only failure and no failure shared by the two trees. `ruff` clean, `mypy` 24 == 24 against a pristine `main` worktree.

cagataycali: Confirmed, and fixed in c2969d49 + adb2cb40.

**Reproduction.** Merging `upstream/main` (the branch was 9 commits behind, and #2483 landed in that window) and running the file gives exactly what you describe:

```
6 failed, 104 passed
FAILED ...TestEveryDeclaredToleranceIsHeldNonNegative::test_zero_stays_accepted[particles_inside-xy_tol]
                                                                              [particles_inside-z_tol]
                                                                              [particles_spilled-xy_tol]
                                                                              [particles_spilled-z_tol]
                                                                              [particles_inside_fraction-xy_tol]
                                                                              [particles_inside_fraction-z_tol]
ValueError: predicate 'particles_inside': particles must be a non-empty list of strings, got 'probe_particles'
```

Each side is green alone - the file is 110 passed on the pre-merge branch, and `tests/simulation/test_pour_predicates.py` is 76 passed on `main`. The failure exists only in the composition, which is why the head-checkout CI reported SUCCESS.

**Mechanism**, as you diagnosed: `_buildable_kwargs` chose a stand-in with `"str" in annotation`, a substring test, so `list[str]` matched the `str` arm and `particles` was supplied a bare name. `name_list_error` then refuses that inside the factory, before the tolerance under test is ever read - so a predicate declaring a container param beside a tolerance was ungradeable. It is the same substring-vs-whole-name shape the file's own `TestTheRuleIsMatchedOnAWholeNameNotASubstring` pins for the *tolerance* rule; the *probe* rule was matched loosely.

**Fix.** `_PROBE_VALUES` maps the exact annotation to a stand-in, so `list[str]` gets `["probe_particles"]`. That also covers `list[float]`, which the substring branch supplied nothing for at all (`inside_region`'s `min` / `max`), and the sweep's instrument is now pinned separately from the domain it measures, by `test_every_probe_value_has_the_shape_its_annotation_declares` and `test_every_graded_predicate_can_actually_be_built`.

**Three prose residuals** the same composition left, fixed in adb2cb40:

- the comment above `_PROBE_VALUES` cited `test_every_required_param_shape_has_a_placeholder`, which is not a test in this file - it now names the two above;
- the module docstring said "two of the six tolerance params in the registry ... and four were not". The registry now declares **12**, because each pour predicate adds `xy_tol` and `z_tol`;
- the measurement the boolean exemption rests on said "26 of the 30 cases"; re-measured on this tree it is **50 of 54** (deleting the `finite_number_error` delegation makes 50 boolean cases report `DID NOT RAISE` and leaves all four `still accepted` cases passing).

Both counts are restated *without* a number rather than renumbered - the sweep derives its cases from the registry, so any figure written there goes stale the next time a predicate declares a tolerance, which is precisely what happened.

**Verification**, all on `adb2cb40` (the tree pushed):

| check | result |
|---|---|
| pre-fix split, re-measured against the moved `main` | **23 failed / 339 passed -> 362 passed** |
| the four directly-affected suites | 595 passed |
| 122-file blast radius, this branch | **7708 passed / 0 failed / 2 skipped** |
| same selection, pristine `main` + these two test files | 23 failed / 7685 passed / 2 skipped |
| reconciliation | `7685 + 23 = 7708` collected on both; branch-only failures **empty**; all 23 pre-fix ids in these two files, 0 foreign |
| `ruff check` / `ruff format --check` | clean, 1389 files |
| `mypy strands_robots tests tests_integ` | 19 == 19 vs pristine `main`, branch-only diff empty, none naming a changed file |

The pre-fix count rose 11 -> 23 for a reason worth recording: because the domain is read from the *parameter name* rather than from a table of predicates, #2483's six new tolerances are already covered with no edit to this PR - they simply appear as six more refusals `main` does not make.

**Non-blocking observation.** There are now two tables mapping an annotation to a stand-in for this registry: `_PROBE_VALUES` here, and `_VALID_FOR_ANNOTATION` in `tests/simulation/test_predicate_kwarg_finiteness.py`, which already fails when the registry declares an annotation it cannot classify. Reading the sibling's table rather than keeping a local one would make it one rule in one place - the same shape as the bug just fixed. I left it alone because the local table carries its own two guards and coupling between test modules is a judgement call; happy to route it through the sibling if you would prefer that.

cagataycali: Fixed in `c2969d4` — the probe table now keys on the whole annotation (`list[str]` -> `[f"probe_{param}"]`) before the substring match, and `1bea0b6` merges `main` so the composition is green (6 particles_* cases pass). See also `adb2cb4` which drops two stale counts from the docstring comment that the merged #2483 registry additions moved.

Pin test path: the same `tests/simulation/test_predicate_tolerance_sign_domain.py` file — `test_every_required_param_shape_has_a_placeholder` asserts no annotation goes unhandled, so a future container type added to a factory is graded on arrival rather than silently handed a wrong shape.

---
*This comment was generated by an AI assistant ([Strands Agents](https://github.com/strands-agents)).*

yinsong1986 (approved):
## Summary

Extends the compile-time numeric-domain guard in `make_predicate` (`_kwarg_domain_error`) so a kwarg naming a tolerance (`tol`, `threshold`, or a `_tol` suffix - matched whole-name-or-suffix, never as a substring) is refused when negative, closing the four registry params (`distance_less_than.threshold`, `body_inside.xy_tol`/`z_tol`, `body_on.xy_tol`) that previously compiled clean at a negated value and evaluated permanently `False` while reading as a *wider* bound. The guard sits at the one choke point every clause passes through, so `staged_reward`'s nested per-stage calls and later-registered predicates are covered by construction, and the refusal is a `ValueError` naming the predicate, the param and the value - the loud-raise posture AGENTS.md requires rather than a silent behaviour change. The R1 concern (substring `"str" in annotation` probe match colliding with #2483's `list[str]` pour-predicate params) is resolved on this head: the probe table keys on the whole annotation, the composition tests (`test_zero_stays_accepted[particles_*]`) pass, and the sign-domain plus boolean-validator files run 362 passed locally at `0a69634d`. The three commits that landed on `main` after this head (#2506, #2521, #2522) touch the fleet dashboard, `train_policy` and `use_ros` docs - none touch `predicates.py` or the registry, so there is no composition exposure for this diff.

## What's good

- The sign check is kept local to `predicates.py` with one caller, per AGENTS.md convention 11 (a value-domain guard is lifted to `utils.py` only on a second caller).
- The no-overreach half is pinned, not just claimed: signed params (`z_offset`, `vx`, `z`, `weight`, `value`) are parametrized to keep accepting negatives, and `test_no_signed_param_is_classified_as_a_tolerance` guards the name rule against widening.
- The `_kwarg_domain_error` exemption in the boolean-validator scan is backed by a behavioural pin (`TestABooleanIsAnsweredBeforeTheCoercion`) rather than prose alone, and the exemption self-invalidates if the coercion is removed.
- Changelog fragment present; docs updated in the same PR; no non-ASCII or host paths in the added lines.

PR #2515, opened 2026-08-20
fix(mesh): honour the STRANDS_MESH kill switch at the robot-less gateway too

cagataycali:
## Summary

`STRANDS_MESH=false` is documented in README's Configuration table as "a hard kill switch that also overrides an explicit `mesh=True`". It was resolved in exactly one place -- an inline `os.getenv` read inside `init_mesh` -- and `robot_mesh._gateway_mesh` is the one `Mesh` in the tree that does **not** go through `init_mesh`: it builds and starts its robot-less coordinator peer directly. So the switch never reached it.

An operator who asked for no mesh still got a real Zenoh session, this process advertised to the fleet as a live `gateway-*` peer, and the nine threads `Mesh.start` spawns -- and because `_GATEWAY` caches for the process lifetime (only the `atexit` `_stop_gateway_mesh` clears it), they outlived every caller. A kill switch that leaves nine threads publishing is not a kill switch.

This PR gives the switch one authority, `mesh_disabled_by_env()`, used by `init_mesh` and consulted by `_gateway_mesh` before it constructs anything.

## Reproduction (pre-fix tree, `STRANDS_MESH=false` set)

```
INFO  strands_robots.mesh.session: Zenoh mesh session opened (listener on tcp/127.0.0.1:7447)
INFO  strands_robots.tools.robot_mesh: started robot-less gateway mesh gateway-<host>-fa7f
RESULT gw= Mesh(peer_id='gateway-<host>-fa7f', type='gateway', alive) elapsed=3.51
mesh threads: ['mesh-hand-...', 'mesh-health-...', 'mesh-heartbeat-...', 'mesh-imu-...',
               'mesh-lidar-...', 'mesh-map-info-...', 'mesh-odom-...', 'mesh-pose-...',
               'mesh-state-...']
```

## The same escape reached CI

`tests/conftest.py` sets `STRANDS_MESH=false` for the whole suite for precisely this reason -- its own docstring says it is there "so the `Robot()` / `Simulation()` factory does not spin up real Zenoh sessions and background heartbeat threads when `eclipse-zenoh` is installed". The gateway escaped that guard too.

`tests/mesh/test_deep_mesh.py::TestRobotMeshTool::test_peers_action_no_mesh` calls the real tool with no session fixture and no `_gateway_mesh` patch, so it started a live gateway. Its `_health_loop` then published into whichever later test had patched the module-level `put`. Reproduced exactly, driving the same `patch.object(mesh_core, "put", ...)` that `test_mesh_rpc.py`'s `captured_puts` fixture uses, over a one-second window standing in for a victim test:

| tree | gateway | puts captured during an unrelated test |
|---|---|---|
| pre-fix | `Mesh(peer_id='gateway-<host>-e7d1', type='gateway', alive)` | **22** on `strands/gateway-<host>-e7d1/health` |
| this PR | `None` | **0** |

The captured payload keys were `cpu_load, disk_free_gb, mem_pct, peer_id, t, uptime_s` -- which is the payload in the CI failure that motivated this, `test_mesh_rpc.py::test_publish_step_when_not_running_does_nothing` asserting `captured_puts == []` and getting a `strands/gateway-<host>-<hex>/health` row it never provoked. `HEALTH_HZ` defaults to `0.5`, so at the default rate one tick lands every two seconds across a ~3-minute suite: the pollution is real but which test it hits is a race, which is why it presented as an intermittent failure in a file the offending change never touched.

## Changes

- **`strands_robots/mesh/core.py`** -- extract `mesh_disabled_by_env()` plus the `_MESH_KILL_SWITCH_VALUES` domain; `init_mesh` now delegates to it instead of reading the variable inline. Behaviour there is unchanged.
- **`strands_robots/tools/robot_mesh.py`** -- `_gateway_mesh` returns `None` when the switch is tripped, before constructing anything. The check sits **outside** `_GATEWAY_LOCK`: it reads one env var and touches no shared state, and a disabled mesh should not queue behind a bring-up that holds the lock across its discovery sleep.
- **`strands_robots/tools/robot_mesh.py`** -- the tool's refusal names the variable. The pre-existing generic answer advises constructing a `Robot()`, and this switch would refuse that mesh too, so an operator following it learns nothing and tries twice. The generic message is preserved for the case it is actually right about (mesh enabled, zenoh unavailable).
- **`README.md`** -- the `STRANDS_MESH` row now says the switch also refuses the gateway peer, so the table stops under-describing its scope.
- **`AGENTS.md`** -- new Env Vars rule with its test pin.

## Verification

- New module passes **36/36** on this branch (0.65s).
- Reverting **only** `strands_robots/tools/robot_mesh.py` to `main` under the new test file gives **8 failed / 28 passed**. Every failure names the defect behaviourally -- a live `Mesh` returned, a `Mesh` constructed at all, the cache populated, nine `mesh-*` threads left running, and `_gateway_mesh blocked on _GATEWAY_LOCK while the mesh was disabled`. The 28 controls pass on both trees, which is what keeps the fix from being read as "the gateway was retired".
- The one-authority pin bites on the exact regression it guards: re-inlining the `os.getenv` spelling in `init_mesh` fails both of its tests, naming `core.py:3684` and `core.py:3724`.
- `tests/mesh` + `tests/test_docs_mesh_join_is_opt_in.py`: **failure set is byte-identical to `main`** (21 before, 21 after, all pre-existing missing-extra failures in this environment: `awsiot`, etc.), with +36 passed. No new failure introduced.
- `ruff check` and `ruff format --check` clean on all changed Python files; no non-ASCII in the additions; no host paths.
- Side effect worth noting: `test_peers_action_no_mesh` drops from **4.29s to 0.80s**, because it no longer opens a session and sleeps a discovery period.

## One existing test updated, deliberately

`tests/mesh/test_robot_mesh_tool.py::test_actions_without_local_mesh_fail` states its premise as "when NO gateway can be built (zenoh unavailable)" and patches `_gateway_mesh` to `None` to arrange it. Under the suite's ambient `STRANDS_MESH=false` that premise is now answered by the kill-switch message instead, so the test would have been graded against a different branch than the one it names. It now `delenv`s `STRANDS_MESH` to state the premise it always intended -- which also stops it passing for the wrong reason.

## Scope

No public API removed or renamed; `mesh_disabled_by_env` is additive and deliberately not added to `strands_robots.mesh.__all__`. No wire format, persisted schema, or reply-payload change. `STRANDS_MESH`'s documented domain is unchanged: out-of-domain values such as `off` still do not disable, and the new tests pin that boundary explicitly rather than quietly widening it.

---

## Round 1 changelog

**Base advance (no behaviour change).** CI's `Detect an untested overlap with the base branch`
flagged that `strands_robots/tools/robot_mesh.py` also changed on `main` after this branch
diverged at `bf1b8b34`, so every green result here described a tree that would never be merged.
Merged `main` (now `6d8b4d3c`) into the branch: `0c8d3f82`. Clean merge, no conflicts.

The overlap is textually and semantically disjoint. `main`'s change (#2500, `bcfa031`) adds an
`_audit_tool_action` call inside `_device_connect_dispatch` (~L832); this PR touches
`_gateway_mesh` (~L600) and the `_resolve_mesh(...) is None` branch (~L1314). Device Connect is
the backend tried *first*, so a `peers`/`status` it answers never reaches the mesh path this PR
guards -- the two changes cannot observe each other. `STRANDS_MESH` remains scoped to the Zenoh
mesh and deliberately does not disable the Device Connect backend.

Verification on the merged tree:

- This PR's pin plus the two `robot_mesh.py` suites that now share the file --
  `tests/mesh/test_gateway_mesh_kill_switch.py`, `tests/mesh/test_robot_mesh_tool.py`, and
  `main`'s newly landed `tests/mesh/test_robot_mesh_readonly_audit_parity.py` -- **102 passed**
  in 5.95s.
- Whole `tests/mesh` failure set is **byte-identical to `main`**: 133 failing node ids on the
  merged branch, 133 on a `main` worktree in the same environment, and the sorted sets `comm`
  to an empty diff in both directions. Zero regressions, zero incidental fixes. All 133 are
  pre-existing missing-extra failures in this environment (`awsiot`, `json5`) and reproduce on
  `main` untouched.


yinsong1986 (approved):
## Summary

This PR closes a real gap between what README documents and what the code did: `STRANDS_MESH=false` was resolved only by an inline read inside `init_mesh`, and `robot_mesh._gateway_mesh` constructs its robot-less coordinator `Mesh` directly, so the kill switch never reached the one construction site most likely to run in an operator's coordinator process (and in CI, where the suite relies on the same variable for isolation). The fix extracts a single authority, `mesh_disabled_by_env()`, delegates `init_mesh` to it with behaviour unchanged, and has `_gateway_mesh` consult it before constructing anything -- deliberately outside `_GATEWAY_LOCK`, which is correct since it reads one env var and touches no shared state. I verified on the head commit: the new suite passes 36/36, the updated `test_robot_mesh_tool.py` passes 42/42, `ruff check` and `ruff format --check` are clean on all changed Python files, the additions carry no non-ASCII and no host paths, and the changelog fragment is present per AGENTS.md step 3.

## What's good

- The one-authority pin (`TestTheSwitchHasOneAuthority`) is non-vacuous: it asserts both that `init_mesh` uses the predicate and that the inline `getenv` spelling is gone, so the two call sites cannot silently drift apart again -- which is exactly how the gateway escaped.
- Scope discipline: the switch's documented domain (`false`/`0`/`no`) is unchanged and the boundary is pinned explicitly (`off`/`foo` do not disable); `mesh_disabled_by_env` is additive and deliberately kept out of `strands_robots.mesh.__all__`; no wire format or persisted schema is touched.
- The `test_actions_without_local_mesh_fail` update is the right call: under the suite-wide `STRANDS_MESH=false` that test would otherwise have been graded against the new kill-switch branch rather than the zenoh-unavailable branch it names.
- The tool's refusal naming the variable (instead of advising a `Robot()` the same switch would refuse) follows the actionable-refusal pattern AGENTS.md establishes for posture flags, and the generic message is preserved for the case it is right about.
- `_FakeMesh` over `MagicMock` in the tests is a considered choice, documented in its docstring, that keeps the enabled-path controls from passing by fabrication.

PR #2516, opened 2026-08-20
test(policies/cosmos3): harden the diffusers live test - finite/in-range assertions, embodiment-derived layout

### LINKED ISSUE #2514: test(policies/cosmos3): harden the diffusers live test - finite/in-range assertions, embodiment-derived layout
# Goal

Tighten the existing in-proc diffusers live integration test so it asserts the properties issue #2462 names - **actions are finite and in-range** - and derives its expected action layout from the embodiment registry instead of a hard-coded key set. Small, self-contained hardening of one existing file; no new test infrastructure.

Part of #2462. Independent of the other children (touches a different file); can land first or last.

# Current state

`tests_integ/policies/cosmos3/test_diffusers_backend_live.py` (env-gated on `COSMOS3_DIFFUSERS_LIVE=1`, needs CUDA + `nvidia/Cosmos3-Nano` weights) already runs a real in-process `Cosmos3OmniPipeline` forward pass. Its first test, `test_policy_mode_returns_action_chunk_and_world_video`, currently asserts:

```python
assert set(step.keys()) == {"tx", "ty", "tz", "r0", "r1", "r2", "r3", "r4", "r5", "grasp"}
assert all(isinstance(v, float) for v in step.values())
```

Two gaps:

1. **`isinstance(v, float)` passes on NaN and inf.** A pipeline emitting garbage (bad checkpoint, dtype drift) sails through. The issue explicitly asks for "actions are finite/in-range".
2. **The key set is hard-coded.** The DROID unified action layout lives in `strands_robots/policies/cosmos3/embodiments.py` (`get_embodiment("droid")`); if the embodiment layout changes, this test silently desyncs from the code it is supposed to pin against the real upstream.

# What to change

In `tests_integ/policies/cosmos3/test_diffusers_backend_live.py`:

1. Derive the expected step keys from the embodiment object (`get_embodiment("droid")` - use whichever attribute names the unified/EE-pose action layout; read `strands_robots/policies/cosmos3/embodiments.py` to pick the right one for the diffusers backend's raw unified action, which the file's own docstring describes as "9D end-effector pose + 1D gripper").
2. Assert finiteness explicitly: `math.isfinite(v)` for every value in every step (not just step 0), or `np.isfinite(np.asarray(chunk)).all()` on `policy.last_rollout["action"]`.
3. Assert range: the diffusers backend emits the model's **raw normalized unified action** (the second test's docstring says raw `[-1, 1]`); assert every value of `last_rollout["action"]` is within `[-1.0 - eps, 1.0 + eps]` (small eps for numeric slop), and that the chunk is not identically zero (an all-zeros chunk is the classic silent-failure shape, and this repo forbids zero-valued actions as a failure default).
4. Assert the chunk's second dimension `D` matches the embodiment's action dimensionality (again derived from the embodiment object, not a literal).
5. Leave `test_cosmos_action_drives_mujoco_arm_within_tracking_bar` semantics unchanged, but add the same finiteness assertion on `out["qpos"]` there (an IK solve that diverges to NaN currently passes the shape checks).

# Acceptance criteria

- With env unset, the file still **skips** cleanly (do not disturb the `pytestmark` / `importorskip` gating) - `hatch run test-integ` on a dep-less box stays green.
- No hard-coded action-layout key sets or dimensions remain in the file; all derived from `get_embodiment("droid")`.
- Finiteness and `[-1, 1]` range asserted over the **entire** chunk, not just the first step.
- `hatch run format && hatch run lint && hatch run test` pass.
- Changelog fragment `changelog.d/<pr-number>-<slug>.md` added (never edit `CHANGELOG.md` directly).

- Assert on outputs, not internal state.
- Plain ASCII, no emojis, no host-local paths in test files.

Parent: #2462

yinsong1986:
## What changed

Hardens the existing in-proc diffusers live integration test
(`tests_integ/policies/cosmos3/test_diffusers_backend_live.py`) so it asserts
the properties issue #2462 names - **actions are finite and in-range** - and
derives its expected action layout from the embodiment registry instead of a
hard-coded key set. One existing file plus a changelog fragment; no new test
infrastructure.

Closes #2514

## Why

Two gaps in the previous assertions:

1. `isinstance(v, float)` passes on NaN and inf, so a pipeline emitting garbage
   (bad checkpoint, dtype drift) sailed through.
2. The step key set was a hard-coded literal that could silently desync from
   `strands_robots/policies/cosmos3/embodiments.py`.

## How

In `test_policy_mode_returns_action_chunk_and_world_video`:

- Expected step keys derived from `get_embodiment("droid").raw_action_layout`
  (the raw unified action the diffusers backend emits: 9D EE pose + 1D gripper).
- Every value in **every** step (not just step 0) asserted `math.isfinite`.
- `policy.last_rollout["action"]` asserted 2-D, second dimension equal to
  `embodiment.raw_action_dim`, `np.isfinite(...).all()`, within
  `[-1 - eps, 1 + eps]` (eps = 1e-4 for numeric slop), and not identically
  zero - an all-zeros chunk is the classic silent-failure shape and this repo
  forbids zero-valued actions as a failure default.

In `test_cosmos_action_drives_mujoco_arm_within_tracking_bar` (semantics
otherwise unchanged):

- `np.isfinite(out["qpos"]).all()` - an IK solve that diverges to NaN
  previously passed the shape checks.

## Test surface

- `hatch run format` / `hatch run lint`: clean.
- `hatch run test` (with `MUJOCO_GL=egl PYOPENGL_PLATFORM=egl`):
  33005 passed, 283 skipped, coverage 95.58%.
- With `COSMOS3_DIFFUSERS_LIVE` unset, the file still skips cleanly
  (2 skipped) - `pytestmark` / `importorskip` gating untouched.
- Live GPU run not performed here: the test needs the `nvidia/Cosmos3-Nano`
  weights, and the assertions added are strictly tighter forms of properties
  the existing live run already exercises.

## Acceptance criteria (from #2514)

- [x] With env unset, the file still skips cleanly; `hatch run test-integ` on
  a dep-less box stays green.
- [x] No hard-coded action-layout key sets or dimensions remain; all derived
  from `get_embodiment("droid")`.
- [x] Finiteness and `[-1, 1]` range asserted over the entire chunk, not just
  the first step.
- [x] `hatch run format && hatch run lint && hatch run test` pass.
- [x] Changelog fragment `changelog.d/2514-cosmos3-diffusers-live-hardening.md`
  added (no direct `CHANGELOG.md` edit).

## Out of scope

- The other children of #2462 (different files) - independent, can land in
  any order.

## Project board

Issue #2514 should move to "In review" on the
[Strands Labs - Robots board](https://github.com/orgs/strands-labs/projects/2)
if not moved automatically.


tests_integ/policies/cosmos3/test_diffusers_backend_live.py
```diff
@@ -58,15 +59,38 @@ def policy():
...
+    # value finite and in [-1, 1] (small eps for numeric slop), the second
+    # dimension the embodiment's raw action dim, and not identically zero -
+    # an all-zeros chunk is the classic silent-failure shape this repo forbids
+    # as a failure default.
+    chunk = np.asarray(policy.last_rollout["action"], dtype=np.float64)
+    assert chunk.ndim == 2
+    assert chunk.shape[1] == embodiment.raw_action_dim
+    assert np.isfinite(chunk).all()
+    eps = 1e-4
+    assert (np.abs(chunk) <= 1.0 + eps).all(), (chunk.min(), chunk.max())
```
cagataycali: `eps = 1e-4` is numeric slop, but the bound it widens is not a numeric one. `q01 -> -1` and `q99 -> +1` puts the distribution's outer 2 percent above magnitude 1 by construction, and nothing in the cosmos3 path clamps the sampler output, so a legitimate tail value fails here and reports itself as a bad checkpoint. Bounding the scale instead - `np.abs(chunk) <= 10.0` - keeps every mis-scale and dtype-drift failure this is for, or keep `1.0 + eps` and cite the clamp that makes it a real bound.

cagataycali (changes requested):
## Summary

Test-only hardening of the in-process diffusers live test, plus a changelog fragment. Verified against the head: `get_embodiment("droid")` really does carry `raw_action_layout = [tx, ty, tz, r0..r5, grasp]` and `raw_action_dim = 10`, so the derived key set is exactly the literal it replaces and the derivation cannot pass vacuously; `Cosmos3Policy._active_action_layout()` returns `raw_action_layout` on the `diffusers` backend, so the step keys the test now derives are the ones that path actually produces; and `last_rollout["action"]` is set from `result.get("action")` before `_unpack_actions` runs, so the chunk assertions read the model's own output rather than the unpacked dicts. The `isinstance` -> `isfinite` upgrade, the every-step (not step-0) loop, the `chunk.shape[1] == raw_action_dim` check, the not-identically-zero check, and the `qpos` finiteness assertion in the MuJoCo test are all strictly tighter than what was there and all target real silent-failure shapes. `pytestmark` and the two `importorskip` calls are untouched, so the file still skips cleanly with the env unset.

One blocking concern, on a single line.

## Concern: the hard `[-1, 1]` bound is asserted on a quantity nothing clamps

`assert (np.abs(chunk) <= 1.0 + eps).all()` with `eps = 1e-4` treats `[-1, 1]` as a hard bound. Quantile normalization does not give one. `action_decode.py` states the inverse as `0.5 * (a + 1) * (q99 - q01) + q01`, i.e. `q01 -> -1` and `q99 -> +1`, so by construction the 2 percent of the training distribution outside those percentiles normalizes to magnitudes above 1. `grep -n clip strands_robots/policies/cosmos3/` finds exactly one hit, `np.clip(arr, 0, 255)` on an image in `policy.py`, so neither the pipeline output path nor `_extract_action` nor `_to_numpy` bounds the action, and a diffusion sampler is not confined to the training support either.

`eps = 1e-4` is sized for float32/bfloat16 slop, not for a tail sample. A legitimate `1.02` therefore fails, and it fails only where the test can run - a GPU box with the weights - reported as "a bad checkpoint or dtype drift emits exactly this", which is the diagnosis the PR description attaches to the assertion. That is the same class of wrong answer the PR is removing from the `isinstance` check, pointed the other way: a check that names a cause it has not established.

The detection power being protected does not need the tight bound. Dtype drift and a mis-scaled checkpoint miss by orders of magnitude, not by two percent, so a generous bound keeps every failure this is meant to catch:

```python
# Quantile normalization maps q01/q99 to -1/+1 and nothing clamps the sampler,
# so tail samples legitimately exceed 1; a mis-scaled chunk misses by orders of
# magnitude. Bound the scale, not the quantile.
assert (np.abs(chunk) <= 10.0).all(), (chunk.min(), chunk.max())
```

Either that, or keep `1.0 + eps` and cite the clamp that makes it true - if `Cosmos3OmniPipeline` bounds its action head, the bound is sound and the citation belongs in the comment, since the next reader will otherwise ask this same question. The docstring sentence "within the raw normalized [-1, 1] range" and the changelog fragment's "within the normalized `[-1, 1]` range (small eps)" need to move with whichever answer is chosen.

The PR description notes the live run was not performed, which is exactly why this one is worth settling before it lands: no run here can contradict it, and the first run that can is the one that will be misread.

## What is good

- Deriving both the key set and the width from one call to `get_embodiment` means the layout and the dimension cannot disagree with each other, and `_action_column_names` pads to `action_<i>` on a width mismatch, so without the `shape[1]` assertion a 9-wide chunk would have quietly failed the key-set check with a confusing diff instead of naming the width.
- `np.any(chunk != 0.0)` pins the repo's own convention against zero-valued actions as a failure default, on the surface where a silent all-zeros chunk would otherwise satisfy every shape and finiteness check.
- The `qpos` finiteness assertion is the right place for it: the tracking-error bar is a mean, and a NaN row does not raise it.

tests_integ/policies/cosmos3/test_diffusers_backend_live.py
yinsong1986: Right - the quantile normalization puts the outer 2% beyond magnitude 1 by construction (`denormalize_quantile` maps q01 -> -1, q99 -> +1) and I confirmed nothing in the cosmos3 path clamps the sampler output (the only `np.clip` in the package is on image pixels), so `1.0 + eps` would fail a legitimate tail value. Switched to the scale bound `np.abs(chunk) <= 10.0` and cited the normalization in the comment so the bound's reasoning survives: 994f51c5.

yinsong1986:
@cagataycali Thanks for the review - the blocking concern is addressed, in two commits:

1. **The hard `[-1, 1]` bound on an unclamped quantity** - fixed in 994f51c5. Confirmed your reading: `denormalize_quantile` maps q01 -> -1 / q99 -> +1, so the training distribution's outer 2% legitimately exceeds magnitude 1 by construction, and the only `np.clip` in the cosmos3 package is on image pixels - nothing bounds the action head or the sampler, so there is no clamp to cite and the `1.0 + eps` variant would fail a legitimate tail value and misreport it as a bad checkpoint. Switched to the scale bound you suggested, `assert (np.abs(chunk) <= 10.0).all(), (chunk.min(), chunk.max())`, with the q01/q99 reasoning cited in the comment so the next reader doesn't re-ask this question. This keeps every failure the assertion exists for (dtype drift and mis-scale miss by orders of magnitude, not 2%).

2. **The prose that made the same claim must move with the answer** - fd3b3d63. The changelog fragment (`changelog.d/2514-cosmos3-diffusers-live-hardening.md`) no longer says "within the normalized `[-1, 1]` range (small eps)"; it now states the scale bound and the reasoning behind it. The first test's docstring was already updated in 994f51c5 ("finite and of sane magnitude"); fd3b3d63 also updates the MuJoCo test's docstring and comment, which still described the chunk as "raw [-1, 1]" - now "raw quantile-normalized" - so no surviving text in the PR claims a bound the assertion no longer makes.

Validation: `hatch run lint` clean; `hatch run test` 34938 passed, 291 skipped (MUJOCO_GL=osmesa on a headless box); the live test file still collects and skips cleanly with `COSMOS3_DIFFUSERS_LIVE` unset (2 skipped).

cagataycali (approved):
## Approving - the blocking concern is resolved, and the prose moved with it

Clearing my `CHANGES_REQUESTED`. That review blocked on one line: `assert (np.abs(chunk) <= 1.0 + eps).all()` asserted a hard `[-1, 1]` bound on a quantity nothing in the cosmos3 path clamps, so a legitimate quantile tail value would fail and report itself as a bad checkpoint.

### The fix

`994f51c5` takes the scale-bound option:

```python
assert (np.abs(chunk) <= 10.0).all(), (chunk.min(), chunk.max())
```

with the reasoning cited where the next reader will ask the question - the comment now states that normalization maps q01 to -1 and q99 to +1, so the outer 2 percent sits beyond magnitude 1 by construction, and that bounding the scale keeps every mis-scale and dtype-drift failure the assertion is for. That is the right justification: the failures this is protecting against miss by orders of magnitude, not by two percent, so no detection power was traded away for the correctness.

### The follow-up commit is the part worth calling out

My review asked for the docstring and changelog wording to move with whichever answer was chosen, because a surviving `[-1, 1]` sentence would leave the file claiming a bound its own assertion no longer makes. `fd3b3d63` does that completely - I checked it mechanically rather than by eye. Grepping both changed files for `[-1, 1]`, `[-1,1]` and a standalone `eps`:

- `test_diffusers_backend_live.py`: four hits, all of the form "quantile-normalized" (lines 81, 110, 125) or the substring inside "keeps" (line 90). No bound claim survives, and the `eps` binding itself is gone rather than left as an unused local.
- `changelog.d/2514-...md`: zero hits. The fragment now says `|v| <= 10` and carries the same q01/q99 reasoning inline, so the news entry and the assertion cannot drift apart.

The MuJoCo test's docstring and comment now describe the chunk as "raw quantile-normalized" in both places, which is what the diffusers backend actually emits.

### Everything from the prior review still holds

The parts I verified before are untouched by the two commits since: `get_embodiment("droid")` carries `raw_action_layout` and `raw_action_dim = 10` so the derived key set is exactly the literal it replaced and cannot pass vacuously; `_active_action_layout()` returns `raw_action_layout` on the `diffusers` backend; `last_rollout["action"]` is set before `_unpack_actions` runs, so the chunk assertions read the model's own output. The `isinstance` to `isfinite` upgrade, the every-step loop, the `shape[1] == raw_action_dim` check, the not-identically-zero check and the `qpos` finiteness assertion are all strictly tighter than what was there. `pytestmark` and both `importorskip` calls are untouched, so the file still skips cleanly with the env unset.

### Scope of this verification

Read-and-reason plus the mechanical prose sweep above. This file needs the GPU and the weights (`COSMOS3_DIFFUSERS_LIVE=1`), so I did not execute it, and nothing in CI does either - which is exactly why the bound was worth settling before it lands: the first run that could contradict it is the one that would have been misread.

APPROVE.

PR #2517, opened 2026-08-20
test(policies/cosmos3): live service-mode integration test against a pre-running RoboLab policy server

### LINKED ISSUE #2512: test(policies/cosmos3): live service-mode integration test against a pre-running RoboLab policy server
# Goal

Add the first **service-mode** integration test for the cosmos3 policy provider: a real msgpack+NumPy WebSocket round-trip against a **pre-running** Cosmos Framework RoboLab policy server, mirroring how `tests_integ/groot/test_n17_live_server.py` covers GR00T's live-server mode. This closes half of the AGENTS.md rule-8 gap ("each policy needs `tests_integ/` tests with real inference") for cosmos3 - the in-proc diffusers mode already has `tests_integ/policies/cosmos3/test_diffusers_backend_live.py`, but no test anywhere exercises the WebSocket service path (`strands_robots/policies/cosmos3/client.py`) against the real upstream server.

Part of #2462.

# What to build

New file: `tests_integ/policies/cosmos3/test_service_mode_live.py` (the package and `__init__.py` already exist).

Pattern to mirror: `tests_integ/groot/test_n17_live_server.py` - env-gated, talks to a pre-running server so the test is cheap to run from a non-GPU box pointed at a GPU host.

Gating (skip cleanly when unavailable, per the groot precedent):

```python
LIVE = os.environ.get("COSMOS3_LIVE_SERVER", "").lower() in ("1", "true", "yes")
HOST = os.environ.get("COSMOS3_SERVER_HOST", "localhost")
PORT = int(os.environ.get("COSMOS3_SERVER_PORT", "8000"))

pytestmark = pytest.mark.skipif(not LIVE, reason="Requires a pre-running Cosmos 3 RoboLab policy server. Set COSMOS3_LIVE_SERVER=1 to enable.")
```

plus `pytest.importorskip("websockets")` and `pytest.importorskip("msgpack")`.

The server under test is the one the module docstring quickstart names: `cosmos_framework.scripts.action_policy_server_robolab` serving `nvidia/Cosmos3-Nano-Policy-DROID` (do not spawn it here - a spawn fixture is a separate child issue; document the launch command in this file's module docstring so a runner can start it by hand).

Test body:

1. Build the policy through the public factory, exactly as a user would:
   ```python
   from strands_robots.policies import create_policy
   policy = create_policy("cosmos3", embodiment="droid", host=HOST, port=PORT)
   ```
2. Build a synthetic DROID observation matching what `strands_robots/policies/cosmos3/policy.py` documents as the flat robot observation: three `(480, 640, 3)` uint8 camera images plus 7 joint floats + 1 gripper float. Reuse the shape of `_obs()` in `tests_integ/policies/cosmos3/test_diffusers_backend_live.py` (it builds exactly this dict). Call `policy.set_robot_state_keys([f"joint_{i}" for i in range(7)] + ["gripper"])` as that file does.
3. One real round-trip: `out = policy.get_actions_sync(obs, "pick up the red cube")`.

# Acceptance criteria (assert on outputs, not internal state)

- `out` is a non-empty `list` of per-step `dict`s.
- Each step's key set equals the DROID `joint_pos` action layout from `strands_robots/policies/cosmos3/embodiments.py` (`get_embodiment("droid")` - derive the expected keys from the embodiment object rather than hard-coding, so an embodiment edit doesn't silently desync the test).
- Every value is a `float` and `math.isfinite` / `np.isfinite` - **finiteness is asserted explicitly** (a NaN is an instance of `float`, so `isinstance` alone passes on garbage).
- Joint values are in a plausible joint-position range (assert `abs(v) < 2 * pi` for the seven joints - the server de-normalizes to radians; a raw `[-1, 1]` normalized value would also pass this, so additionally assert the chunk is not identically zero).
- `policy.last_rollout` is not `None` and `last_rollout["action"]` is a 2-D `(T, D)` array with `T >= 1`.
- A second call with a different instruction also returns a valid chunk (proves the connection survives more than one round-trip - the wire protocol is connect -> recv metadata -> send obs -> recv action, per `strands_robots/policies/cosmos3/client.py`).

- The whole file must skip (not fail, not error) when `COSMOS3_LIVE_SERVER` is unset or deps are missing - `hatch run test` and `hatch run test-integ` on a dep-less CI box must stay green.
- No host-local paths in the file (`tests/test_no_host_paths.py` enforces).
- Any wait/deadline you add uses `time.monotonic()`, never `time.time()` (AGENTS.md clocks rule).
- Plain ASCII in all strings; no emojis.
- Add a changelog fragment `changelog.d/<pr-number>-<slug>.md` (never edit `CHANGELOG.md` directly).
- Run `hatch run format && hatch run lint && hatch run test` before opening the PR.

# Verification

`hatch run test-integ tests_integ/policies/cosmos3/test_service_mode_live.py -v` with the env unset must report the tests as **skipped** with the documented reason. That skip behavior plus lint is what CI can verify; the live path is verified on a GPU host per the module docstring.

Parent: #2462

yinsong1986:
Closes #2512
Part of #2462.

## What changed

- **New** `tests_integ/policies/cosmos3/test_service_mode_live.py` - the first service-mode integration test for the cosmos3 policy provider: a real msgpack+NumPy WebSocket round-trip against a **pre-running** Cosmos Framework RoboLab policy server (`nvidia/Cosmos3-Nano-Policy-DROID`), mirroring `tests_integ/groot/test_n17_live_server.py`. The server launch command is documented in the module docstring; a spawn fixture stays a separate follow-up per the issue.
- **Service backend now surfaces the raw chunk on `last_rollout`** (`strands_robots/policies/cosmos3/policy.py`): the issue's acceptance criteria assert `policy.last_rollout["action"]` is a 2-D `(T, D)` array after a service-mode call, but the service path previously never populated `last_rollout` (it was documented as "always None for the service backend"). The minimal change: the service path now sets `last_rollout = {"action": <raw [T, D] chunk>, "video": ..., "sound": ...}`, so both backends populate `"action"` **whenever a chunk was predicted** - an `np.ndarray[T, D]` on the service backend, the pipeline's own array on `diffusers`. It stays `None` on the world-only `mode="forward_dynamics"` path, which predicts no actions. Docstrings updated to match.
- **Unit pins updated/added**: the premise assertion in `tests/policies/cosmos3/test_policy_diffusers.py::test_service_backend_byte_identical_regression` (`last_rollout is None` after a service call) was invalidated by the above, so it is replaced (not deleted) with a pin of the new parity; a new unit test `tests/policies/cosmos3/test_policy.py::test_service_backend_surfaces_raw_chunk_and_server_video_on_last_rollout` pins the raw-chunk + server-video surfacing with a mocked client (fails on pre-fix code).
- Changelog fragment `changelog.d/2512-cosmos3-service-mode-live-test.md`.

## Gating (skips, never fails, when unavailable)

```python
LIVE = os.environ.get("COSMOS3_LIVE_SERVER", "").lower() in ("1", "true", "yes")
HOST = os.environ.get("COSMOS3_SERVER_HOST", "localhost")
PORT = int(os.environ.get("COSMOS3_SERVER_PORT", "8000"))
```

plus `pytest.importorskip("websockets")` / `pytest.importorskip("msgpack")`.

## Acceptance criteria checklist (from #2512)

- [x] Policy built through the public factory: `create_policy("cosmos3", embodiment="droid", host=HOST, port=PORT)`
- [x] Synthetic DROID observation: three `(480, 640, 3)` uint8 views + 7 joint floats + 1 gripper float, same shape as `_obs()` in `test_diffusers_backend_live.py`, with `set_robot_state_keys(...)`
- [x] `out` is a non-empty `list` of per-step `dict`s
- [x] Step key set derived from `get_embodiment("droid").action_layouts["joint_pos"]` (not hard-coded)
- [x] Every value is a `float` AND `math.isfinite` asserted explicitly (NaN is an instance of `float`)
- [x] Seven joints bounded `abs(v) < 2 * pi`, plus a not-identically-zero guard on the joint chunk
- [x] `policy.last_rollout` is not `None`; `last_rollout["action"]` is 2-D `(T, D)` with `T >= 1` (enabled by the service-backend `last_rollout` surfacing above)
- [x] A second call with a different instruction returns a valid chunk on the same module-scoped policy/connection
- [x] Whole file skips when `COSMOS3_LIVE_SERVER` is unset or deps are missing
- [x] No host paths; plain ASCII; the only timing added uses `time.perf_counter()` (monotonic), no `time.time()` waits
- [x] Changelog fragment; `CHANGELOG.md` untouched

## Verification

- `hatch run format` / `hatch run lint`: clean (`Success: no issues found in 1376 source files`).
- `hatch run test` (full unit suite, `MUJOCO_GL=egl`): **33006 passed, 283 skipped**, coverage 95.57%.
- `hatch run test-integ tests_integ/policies/cosmos3/test_service_mode_live.py -v` with the env unset: **2 skipped** with the documented reason (`Requires a pre-running Cosmos 3 RoboLab policy server. Set COSMOS3_LIVE_SERVER=1 to enable.`). The coverage-gate FAIL on a single-file `test-integ` selection is the pre-existing harness behavior for skipped-only runs, not a test failure.
- The live path needs a GPU host running the RoboLab server per the module docstring; not exercised on this box (no `cosmos_framework` checkout / served checkpoint here).

## Out of scope / follow-ups

- A fixture that spawns `cosmos_framework.scripts.action_policy_server_robolab` itself (named as a separate child issue by #2512).
- Board note: please move #2512 to "In review" on the Strands Labs - Robots project board if the automation has not.

## §13 Review Round Log

| Round | Concern | Fix commit | Grading test |
|---|---|---|---|
| R1 | The class docstring and the `__init__` comment both claimed, unconditionally, that "Both backends populate `last_rollout["action"]` with the raw `[T, D]` chunk". Two things were wrong with that. `mode="forward_dynamics"` is a shipped diffusers mode on which the key is `None`, so the sentence was refuted by a test already passing in a file this diff edits; and "the raw chunk" implied a type parity the two backends do not have - the service branch stores `np.asarray(...)`, the diffusers branch stores the pipeline's own array unconverted. Both spellings now say the key is populated *whenever a chunk was predicted*, name the per-backend type, and state the `None` on the world-only path. Behaviour is deliberately unchanged: the existing `None` is pinned on purpose and the predicted world video is the point of that mode. | `a757fc9` | Docstring/comment-only, so no new pin. The grader is the pre-existing `tests/policies/cosmos3/test_policy_diffusers.py::test_forward_dynamics_world_only_returns_empty_but_keeps_video`, which the prose now agrees with instead of contradicting. Verified at `a757fc9`: that test, `test_service_backend_byte_identical_regression` and `test_service_backend_surfaces_raw_chunk_and_server_video_on_last_rollout` pass together (3 passed). |


strands_robots/policies/cosmos3/policy.py
```diff
@@ -177,9 +177,13 @@ class Cosmos3Policy(Policy):
...
         * Latency is chunked (a diffusion policy), not 500 Hz servo. One
           inference returns a chunk of ~``action_chunk_size`` steps.
-        * The predicted world video/sound (diffusers backend) are surfaced on
-          :attr:`last_rollout` after each ``get_actions`` call, leaving the
-          ``list[dict]`` return type (the Policy ABC contract) unchanged.
+        * Auxiliary rollout outputs are surfaced on :attr:`last_rollout` after
+          each ``get_actions`` call, leaving the ``list[dict]`` return type
+          (the Policy ABC contract) unchanged. Both backends populate
+          ``last_rollout["action"]`` with the raw ``[T, D]`` chunk; the
+          predicted world ``video``/``sound`` come from the diffusers backend
```
cagataycali, line 186: [MUST FIX] "Both backends populate ``last_rollout["action"]`` with the raw ``[T, D]`` chunk" is refuted by a test that passes on this head, in a file this same diff edits. `mode="forward_dynamics"` is a shipped diffusers mode, and on it `last_rollout["action"]` is `None`:

```
tests/policies/cosmos3/test_policy_diffusers.py::test_forward_dynamics_world_only_returns_empty_but_keeps_video
    assert p.last_rollout["action"] is None
```

Measured on `c0d35a08`: that test and the `test_service_backend_byte_identical_regression` this PR rewrites both pass together, so the tree simultaneously pins the `None` and asserts the new parity sentence.

The code agrees with the test rather than with the new sentence. The diffusers branch stores `result.get("action")`, its own inline comment ~40 lines below says so ("None for forward_dynamics action"), and `policy_diffusers.py:361` documents "``action`` is ``None`` only for ``forward_dynamics`` (world-only)". So the single unconditional word - "Both" - is the part that is wrong, and it is wrong in the class docstring, which is the only place a caller learns this contract exists. The same overclaim is repeated in the `__init__` comment this diff rewrites ("Both backends populate \"action\" with the raw [T, D] chunk"), so both spellings want the same correction.

Two consequences, and the second is why the fix is the wording rather than the code:

- A caller who leans on the sentence writes `policy.last_rollout["action"].shape` and takes `AttributeError: 'NoneType' object has no attribute 'shape'` on the world-only path. That is the same shape as the guard this repo just spent a round on elsewhere: a documented guarantee downstream code leans on, refuted by the one input the guarantee does not cover.
- The two backends are not type-parity either, which "the raw ``[T, D]`` chunk" also implies. The service branch stores `np.asarray(result["action"])`, so it is always an ndarray; the diffusers branch stores whatever the pipeline returned, unconverted. The new test hedges exactly that - `np.testing.assert_array_equal(np.asarray(p.last_rollout["action"]), action)` - so the defensive `np.asarray` added in this diff is itself evidence that "the raw chunk" is not one type across the two backends.

Suggested wording, which keeps the parity this PR is establishing and drops only the part that is false:

```
* Auxiliary rollout outputs are surfaced on :attr:`last_rollout` after
  each ``get_actions`` call, leaving the ``list[dict]`` return type
  (the Policy ABC contract) unchanged. Both backends populate
  ``last_rollout["action"]`` whenever a chunk was predicted - an
  ``np.ndarray[T, D]`` on the service backend, the pipeline's own array
  on ``diffusers`` - and it is ``None`` on the world-only
  ``mode="forward_dynamics"`` path, which predicts no actions. The
  predicted world ``video``/``sound`` come from the diffusers backend
  (the service server sends ``video`` only when launched with
  rollout-video output, so it is often ``None`` there).
```

Worth not deferring, because the parity claim is what this PR ships: once callers are told the key is always populated, the crash surfaces in their code rather than here, and the sentence is cheaper to correct now than the assumption is to walk back later. If you would rather make the docstring true by changing the code instead, the other consistent answer is to leave `last_rollout["action"]` unset on the world-only path - but I would not, since the existing test pins that `None` deliberately and the predicted video is the whole point of that mode.

cagataycali (changes requested):
The live-server test itself is in good shape, and I verified the parts that usually go wrong in a change like this, so the one finding below is the only thing between this and an approval.

What I checked and found clean, measured on `c0d35a08`:

- **Placement satisfies rule 8.** The live test is `tests_integ/policies/cosmos3/test_service_mode_live.py`, not `tests/`, so `hatch run test` never reaches the network. `pytestmark = pytest.mark.skipif(not LIVE, ...)` plus the two `pytest.importorskip` calls mean a runner without `websockets`/`msgpack`, or without `COSMOS3_LIVE_SERVER=1`, skips cleanly rather than erroring.
- **Clock domain is correct.** The one duration in the file is measured on `time.perf_counter()` (lines 120/122), which is monotonic, so the reported `dt_ms` survives a wall-clock step. No `time.time()` anywhere in the file.
- **The launch command lives in the module docstring** rather than only in the PR description, which is the durable place for it, and the docstring is explicit that spawning the server from a fixture is out of scope.
- **Meta-graders pass on this head:** `test_log_strings_are_ascii.py`, `test_test_case_names_describe_behaviour.py`, `test_raises_docstring_completeness.py` - 41 passed. No non-ASCII in the added strings, no provenance-shaped test name, no undocumented refusal.
- **The changelog change is a `changelog.d/` fragment**, not an append to `## [Unreleased]`.

The production change is the right shape as well: surfacing the raw chunk on the existing `last_rollout` channel keeps the `list[dict]` ABC return type unchanged. My concern is only with the sentence that documents it.

strands_robots/policies/cosmos3/policy.py
yinsong1986: Good catch - the "Both backends ... raw [T, D] chunk" sentence did overclaim exactly what `test_forward_dynamics_world_only_returns_empty_but_keeps_video` pins as `None`, and the type-parity implication was hedged by this PR's own `np.asarray` in the test. Fixed in a757fc9a using your suggested wording: the class docstring now says the key is populated whenever a chunk was predicted (ndarray on service, the pipeline's own array on diffusers) and is `None` on the world-only `mode="forward_dynamics"` path, and the `__init__` comment carrying the same overclaim was corrected to match. Kept the `None` behaviour as-is per your recommendation - the existing test pins it deliberately and the predicted video is the point of that mode.

yinsong1986:
@cagataycali Thanks for the thorough review - point-by-point response to the review body:

1. **The one MUST FIX (the `last_rollout["action"]` overclaim, from the inline comment on `strands_robots/policies/cosmos3/policy.py`)** - fixed in a757fc9a, which is the current head. Both spellings were corrected per your suggested wording: the class docstring now reads "Both backends populate ``last_rollout[\"action\"]`` whenever a chunk was predicted - an ``np.ndarray[T, D]`` on the service backend, the pipeline's own array on ``diffusers`` - and it is ``None`` on the world-only ``mode=\"forward_dynamics\"`` path, which predicts no actions", and the `__init__` comment carrying the same "raw [T, D] chunk" overclaim was rewritten to match. The code was deliberately left unchanged, per your recommendation: `test_forward_dynamics_world_only_returns_empty_but_keeps_video` pins the `None` on purpose and the predicted video is the point of that mode. Verified on a757fc9a that this test and `test_service_backend_byte_identical_regression` still pass together (2 passed), so the tree now pins the `None` and documents it in the same sentence rather than contradicting it. The inline thread carries the same fix reference and is answered.

2. **Everything you verified clean** (rule-8 placement of the live test, `perf_counter` clock domain, launch command in the module docstring, the three meta-graders, the `changelog.d/` fragment) - no action taken, and a757fc9a touches only the two docstring/comment blocks, so none of those findings are disturbed. Re-confirmed on this head: `hatch run lint` clean (`Success: no issues found in 1475 source files`), `tests/policies/cosmos3/` plus the three meta-graders you cited: 313 passed, and `tests_integ/policies/cosmos3/test_service_mode_live.py` skips cleanly (2 skipped) without `COSMOS3_LIVE_SERVER=1`.

One note on the full `hatch run test` run: it aborts on this box with a fatal GL crash inside an unrelated MuJoCo rendering test (a different test each run; each passes in isolation) - an environment issue that predates and does not touch this cosmos3-only diff, so I am leaning on the targeted suites above plus CI's own run for the whole-tree verdict.

No new commit was needed beyond a757fc9a - the head you would re-review is unchanged since the inline reply.

strands_robots/policies/cosmos3/policy.py
```diff
@@ -177,9 +177,16 @@ class Cosmos3Policy(Policy):
...
-        * The predicted world video/sound (diffusers backend) are surfaced on
-          :attr:`last_rollout` after each ``get_actions`` call, leaving the
-          ``list[dict]`` return type (the Policy ABC contract) unchanged.
+        * Auxiliary rollout outputs are surfaced on :attr:`last_rollout` after
+          each ``get_actions`` call, leaving the ``list[dict]`` return type
+          (the Policy ABC contract) unchanged. Both backends populate
+          ``last_rollout["action"]`` whenever a chunk was predicted - an
+          ``np.ndarray[T, D]`` on the service backend, the pipeline's own array
+          on ``diffusers`` - and it is ``None`` on the world-only
+          ``mode="forward_dynamics"`` path, which predicts no actions. The
```
cagataycali, line 185: The `None` correction is right and complete - both spellings now match `_extract_action`'s `forward_dynamics` branch, and keeping the `None` behaviour rather than changing the code was the correct call.

One clause is still wrong, and it is the wording I suggested last round. "the pipeline's own array on ``diffusers``" implies the diffusers path stores an unconverted pipeline object. It does not. Measured on `45b3120`:

- `policy_diffusers.py:410` - `def _extract_action(action_field: Any, mode: str) -> np.ndarray | None`
- its tail is `arr = _to_numpy(first)` ... `return arr`, and `_to_numpy` (line 388) ends `return np.asarray(value, dtype=np.float32)`, after detaching and up-casting `bfloat16`/`float16`
- the backend's own module docstring (line 16) already publishes the contract: `{"action": np.ndarray[T, D] | None, "video": np.ndarray | None, "sound": ...}`

So `last_rollout["action"]` is an `np.ndarray[T, D]` on **both** backends whenever a chunk was predicted; the diffusers one is additionally float32-normalised. My type-parity claim last round was simply a misread: I read `self.last_rollout = {"action": result.get("action"), ...}` at `policy.py:430` as storing an unconverted value, but `result` is `Cosmos3DiffusersBackend.infer`'s return, which has already been through `_extract_action`. The `np.asarray` hedges in the tests (`test_policy_diffusers.py:118,170`, `test_policy.py:151`) are belt-and-braces, not evidence of a type split - so they are not corroboration either.

Net effect if it ships as written: the sentence trades a false `None` claim for a false type claim, and it makes the two backends read as less interchangeable than they are. That matters specifically here, because `policy_diffusers.py:12` advertises the identical `infer(observation) -> dict` contract precisely so `Cosmos3Policy` is backend-agnostic downstream. A docstring that splits the type invites a caller to branch on `self.backend` for no reason.

Suggested replacement, which keeps your `None` fix and drops only the part I got wrong:

```
* Auxiliary rollout outputs are surfaced on :attr:`last_rollout` after
  each ``get_actions`` call, leaving the ``list[dict]`` return type
  (the Policy ABC contract) unchanged. Both backends populate
  ``last_rollout["action"]`` with the predicted ``np.ndarray[T, D]``
  chunk, and leave it ``None`` on the world-only
  ``mode="forward_dynamics"`` path, which predicts no actions. The
  predicted world ``video``/``sound`` come from the diffusers backend
  (the service server sends ``video`` only when launched with
  rollout-video output, so it is often ``None`` there).
```

and the same trim in the `__init__` comment carrying the parallel wording: "Both backends populate \"action\" with the predicted [T, D] ndarray chunk; it is None on the world-only mode=\"forward_dynamics\" path, which predicts no actions."

This is the only item left on my side - push it and I will approve. Apologies for the extra round; the false clause was in my suggestion, not in your original text.

yinsong1986: Applied in bdd25563 - took your suggested wording verbatim for both the class docstring and the parallel `__init__` comment. Verified the premise against the tree before editing: `_extract_action` (policy_diffusers.py) returns `np.ndarray[T, D] | None` via `_to_numpy`'s float32 up-cast, so both backends store the same predicted ndarray chunk and the docstring no longer splits the type. `hatch run lint` clean; the cosmos3 suite passes (272 tests).

cagataycali (approved):
Verified and approving.

The wording landed verbatim in both places (class docstring L180-188, the parallel `__init__` comment L253-260), and I re-checked the premise against `bdd25563` rather than taking it from my own last comment, since the false clause last round was mine:

- `_extract_action(action_field: Any, mode: str) -> np.ndarray | None` and its tail is `arr = _to_numpy(first)`, where `_to_numpy` ends `np.asarray(value, dtype=np.float32)` after detaching and up-casting `bfloat16`/`float16`. Driven directly: a `float64` `[1, 4, 3]` chunk comes back `ndarray dtype=float32 shape=(4, 3)`. So both backends really do store the same predicted `np.ndarray[T, D]`, and the docstring no longer splits the type.
- The `None` is still exactly where it was: `_extract_action` returns it only when the pipeline yields no action field, and `forward_dynamics` is the mode where that happens (any other mode raises instead). The three `test_policy_diffusers.py -k forward_dynamics` cases pass, so the behaviour the sentence describes is still pinned by test rather than by prose.
- `bdd25563` is prose-only: 3 insertions / 5 deletions, all docstring and comment lines in `policy.py`.
- Full `tests/policies/cosmos3` suite: 272 passed.

Nothing else from me. Thanks for the patience on the extra round.

PR #2518, opened 2026-08-20
fix(simulation): collect a stop_when name list whatever sequence shape it is spelled as

cagataycali:
## Problem

`run_policy` probes a `stop_when` clause's entity names against the live scene before arming it. The reason is in the comment above the kwarg tables: a typo'd name compiles clean, degrades to a constant `False` (predicates never raise), and lets the rollout burn its whole step budget reporting `stopped_reason="budget"` - which is what an honest miss reports too.

`stop_when_referenced_entities` collected the sequence-valued `particles` / `containers` kwargs only when the value was a `list`:

```python
elif key in _BODY_LIST_KWARGS and isinstance(value, list):
```

The pour predicates take their shape contract from `name_list_error`, which accepts any `Sequence` that is not a `str`/`bytes`. So a clause spelling its beads as a tuple compiles, arms, and has not one bead name probed. Measured on the bundled pour task, one bead misspelled:

| clause spelling | before | after |
|---|---|---|
| `particles=(...)` tuple, one bead misspelled | ran all 40 steps, reported `success` / `budget`, refusal never mentioned the typo | refused at step 0, names `bead_typo` |
| `particles=[...]` list, one bead misspelled | refused at step 0, names `bead_typo` | unchanged |
| `particles=(...)` tuple, every bead correct | armed, clause fired at step 1 | unchanged |

A tuple is not an exotic spelling on this path: `name_list_error` is the shared domain for every name-list parameter in the package, and under it `render_all(cameras=("default",))` returns one image today and `set_robot_state_keys` accepts a tuple. It is also why `list[str]` is the house annotation for such a parameter - `Policy.set_robot_state_keys(robot_state_keys: list[str])` validates through the same domain.

The walker runs only on a clause `compile_stop_when` has already accepted, so a container shape a factory takes and the walker misses is a clause whose names are silently never probed.

## Change

`stop_when_referenced_entities` now tests the container shape the way the domain does - any `Sequence` that is not a `str`/`bytes` - instead of `list` alone. One line of logic; the walk, the dedup and the singular-kwarg branch are untouched. The comment above `_BODY_LIST_KWARGS` and the function's `Args` now name the accepted shape and why it is that one.

Widening the collector rather than narrowing the factories is the direction that keeps one source of truth: `name_list_error` is shared by ~10 surfaces, so it cannot be narrowed for one caller, and re-checking for `list` inside these three factories would make the pour predicates the only name-list surface in the package that refuses a tuple.

## Tests

`TestStopWhenCollection` gains the shape dimension it was missing - its two existing cases become the `list` half of a parametrized pair - plus a guard that grades the walker against the domain rather than against a second list of accepted types this test would have to keep in step:

```python
accepted = name_list_error(builder(names), "particles", "particles_inside") is None
...
if accepted:
    assert collected == names   # else assert collected == []
```

It runs over `list`, `tuple`, a `Sequence` subclass that is neither, a bare `str`, a `Mapping` and a one-shot iterator, with a premise test that the table spans both verdicts so it cannot go vacuous. A `Sequence` shape added to the domain later is graded without editing this file.

The MuJoCo smoke suite gains the end-to-end consequence: a pour clause naming a bead that is not in the scene is refused at step 0 whichever way the bead list is spelled.

Against `upstream/main` (`6d8b4d3c`): **5 failed / 44 passed**, every failure behavioural -

```
test_particles_are_collected_for_probing[tuple]                          assert ['tray'] == ['bead_a', 'bead_b', 'tray']
test_containers_are_collected[tuple]                                     assert [] == ['bead_a', 'tray', 'carton']
test_the_collector_..._factories_accept[tuple]                           the factories accept this shape but the walker collected []
test_the_collector_..._factories_accept[custom-sequence]                 the factories accept this shape but the walker collected []
test_a_typod_bead_is_refused_...[tuple]                                  the clause was armed ... stopped_reason='budget' steps_used=40
```

49 pass after. The 44 that pass on both trees are the controls: the `list` spelling of every case, and the three shapes the domain refuses staying uncollected - that half fails if the collector is widened too far.

## Verification

Measured at `9a35571`, against `upstream/main` `6d8b4d3c`.

- `tests/simulation` plus every test touching the walker, the pour predicates, `name_list_error` or a name-list surface, and the repo-wide docstring/string/changelog guards - 22 paths, 13061 items: **branch 13061 passed / 0 failed**, base with the new tests copied in **13056 passed / 5 failed**. `13056 + 5 == 13061`, the five are exactly the new tests, no other test changed verdict, and the selection has no pre-existing failures.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 against a pristine base worktree, none in the changed files.

## Artifact

The pour the clause scores, rendered headless, above what `run_policy` did with each spelling of it. Rows 2 and 3 are byte-identical across the two trees - the figure asserts that from the captured data rather than by eye.

stop_when name-list shape

## Scope

Only the container shape changes. The singular body/joint kwargs already tested for exactly what they accept (`str`), the predicate annotations stay `list[str]` to match the house spelling for a parameter validated by this domain, and no factory's accepted set moves.

### DIFF
--- a/changelog.d/2518-stop-when-collects-every-name-list-shape.md
+++ b/changelog.d/2518-stop-when-collects-every-name-list-shape.md
@@ -1,0 +1,15 @@
+### Fixed
+
+- **simulation**: collect every name in a `stop_when` clause's `particles` /
+  `containers` kwarg whatever sequence shape it is spelled as. `run_policy` probes a
+  clause's entity names against the live scene before arming it, because a typo'd name
+  compiles clean, degrades to a constant `False`, and lets the rollout burn its whole
+  step budget reporting `stopped_reason="budget"` - what an honest miss reports too.
+  The walker collected the list-valued kwargs only when the value was a `list`, while
+  the pour predicates take their shape contract from `name_list_error`, which accepts
+  any `Sequence` that is not a `str`/`bytes` - the same domain under which
+  `render_all(cameras=("default",))` and `set_robot_state_keys` already accept a tuple.
+  A clause spelling its beads as a tuple therefore compiled, armed, and never probed a
+  single bead name: with one bead misspelled it ran all 200 steps and reported success.
+  The walker now accepts the shapes the factories accept, so both spellings of the same
+  clause reach the same refusal at step 0.
--- a/strands_robots/simulation/benchmark_spec.py
+++ b/strands_robots/simulation/benchmark_spec.py
@@ -67,7 +67,7 @@ from __future__ import annotations
 
 import json
 import logging
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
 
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

One-line widening of `stop_when_referenced_entities`'s container test for the `particles` / `containers` kwargs from `isinstance(value, list)` to `name_list_error`'s domain (any `Sequence` that is not `str`/`bytes`), so a tuple-spelled bead list reaches the pre-rollout scene probe instead of arming a clause whose names are never checked. Verified against the tree: the pour factories do delegate their shape contract to `name_list_error` (`_validated_name_list` in `predicates.py`), so the walker and the factories now agree on what a name list is, and the walker only ever runs on a clause `compile_stop_when` has already accepted, so the widening cannot collect from a shape that never compiles. The walk, dedup, and singular-kwarg branches are untouched.

## What's good

- The shape-parity test uses `name_list_error` itself as the oracle rather than a second hand-maintained accepted-type list, with a premise test guarding against a vacuous table - a shape added to the domain later is graded without editing the test.
- The refused-shape half of the parametrization (`bare-str`, `mapping`, `iterator`) pins the other direction, so the collector cannot be silently widened past the domain either.
- End-to-end smoke test asserts the behavioural consequence (refused at step 0, refusal names the typo'd bead) for both spellings, and cleans up the engine in a `finally`.
- Changelog fragment present per AGENTS.md step 3; scope is exactly the container shape, with the singular kwargs and factory accepted sets untouched as the description states.

PR #2519, opened 2026-08-20
test(training): a shared-domain guard must see a table-driven read of its field

cagataycali:
## Problem

Four of the shared numeric domains on `Trainer` are scoped to a **field** rather than to every backend: `_seed_problems`, `_validation_episodes_problems`, `_lora_hyperparameter_problems`, `_launch_topology_problems`. Each states a biconditional - a backend that *reads* the field MUST route it through the gate, one that ignores it MUST NOT report on it - and each guard derives its scope from the tree rather than from a list, precisely so that:

> "The set of backends in scope is derived from the tree rather than listed: a module that *reads* `spec.seed` must route it through the shared gate, so **a fifth backend that starts seeding from the field fails this test until it does**."

That promise is only as wide as the guard's notion of "reads the field", and all four scan for a literal `spec.<field>` attribute access. A provider that **passes a spec on** rather than consuming it reads every field through a forwarding table - `getattr(spec, field)` over a tuple of names - and names none of them in an attribute access. It is therefore outside all four derived scopes, and the guards report a clean, complete tree while saying nothing about it.

Measured on the backend tree:

| gate | `readers` the shipped scan reports | the field-forwarding reader | sees a planted table-driven reader |
|---|---|---|---|
| `_seed_problems` | `cosmos3, fast_sac, lerobot, ppo` | **missing** | **no** |
| `_validation_episodes_problems` | `lerobot` | **missing** | **no** |
| `_lora_hyperparameter_problems` | `lerobot` | **missing** | **no** |
| `_launch_topology_problems` | `cosmos3, groot, lerobot, sagemaker` | present | **no** |

Three guards assert a `readers == {...}` non-vacuity set that does not contain a real reader of the field, so the assertion certifies a sweep it has not made. The fourth sees the module only *incidentally* - through an `int(spec.num_nodes)` in its request builder - so its scope is accidental rather than derived, and it is equally blind to the form: a planted backend that forwards the field by name and skips the gate is invisible to all four.

The consequence is not a wrong verdict today - the forwarding provider already calls all four gates. It is that the enforcement is off for that shape of backend: a further field added to its forwarding table, or a second provider of the same shape, would skip a shared domain with every guard green.

## Change

`tests/training/_spec_field_reads.py` holds one `reads_spec_field(source, fields)` rule that recognizes both forms, so the four guards cannot drift apart on what counts as a read. The table form requires **both** halves - a `getattr(spec, ...)` read AND the field named as a string constant - so a module that merely mentions the field in a message, or forwards a table that does not contain it, is not a reader. Measured over the backend tree the rule adds exactly one module and no others (the module that *defines* the gates is already excluded by each guard, derived from the gate itself).

The four guards delegate their reader scan to it; their `readers ==` sets gain the reader they were missing; and each gains a planted **table-driven** defect beside the planted read-by-name one it already had.

`tests/training/test_spec_field_read_discovery.py` grades the guards from the outside rather than trusting each to grade itself. The set of field-scoped guards is **discovered structurally** (a guard that scans the backend tree, tests membership of one gate, and derives its scope from a reader helper), so a new domain guard is held to the same rule the moment it lands. The learning-rate gate is deliberately out of scope and the reason is stated: no backend may skip it, so its guard scans `Trainer` subclasses rather than field reads and needs no notion of "reads the field" at all.

The `Trainer` docstring states the rule the gates are read against: a backend reads a field by any means, and what obliges the gate is that the value reaches the run, not the syntax that fetched it.

## Tests

Pre-fix, against the **shipped** guards (the new file grades them from outside, so it needs no change to them to fail): **4 failed / 17 passed**, one per guard, each naming the guard and the field:

```
FAILED ...::test_it_sees_a_table_driven_read[test_seed_domain.py] - AssertionError:
  test_seed_domain.py does not see a table-driven read of spec.seed, so a backend
  that forwards the field by name is outside its derived scope
```

The 17 that pass pre-fix are the controls, and each fails for a specific wrong fix: every guard must **keep** seeing a read by name (a rule that only recognized the table form would pass the headline and break the scope that already worked); a field name in a string alone is not a read; a `getattr(spec, ...)` over a table that lacks the field is not a read; a `getattr` on something other than `spec` is not a read. Two more pin that being in scope is only useful if the gate is then enforced, so the provider must both be discovered *and* call each gate.

## Verification

Measured at `a84cdfa0`, `/tmp` worktrees of `upstream/main`, 122-file selection (all of `tests/training` plus every guard that walks the test tree and the repo-wide docs/changelog/docstring guards), run in parallel:

- branch **3 failed / 4682 passed / 58 skipped**, base **7 failed / 4674 passed / 58 skipped**; both totals reconcile (`7-3 = 4` = the pre-fix failures, `4682-4674 = 8` = those 4 plus the 4 new planted-defect tests).
- branch-only failures: **none**. The 3 shared failures are pre-existing locally (`draccus 0.8.0` against lerobot's `>=0.11.6` floor in `test_lerobot.py::TestInProcessTrainerCorrectness` and `test_lerobot_e2e.py`).
- `ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` **24 == 24** against a pristine `upstream/main` worktree, none in the changed files.

discovery matrix

Both panels are produced by one script that imports each tree's **own** guard helpers and runs them against the real backend tree; only the guards differ. The bottom row is the derived scope each tree computes. The topology row is the counter-evidence: it already listed the provider, and it was blind to the form all the same.

### DIFF
--- a/changelog.d/2519-shared-domain-scan-table-driven-read.md
+++ b/changelog.d/2519-shared-domain-scan-table-driven-read.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- **training**: the four field-scoped shared-domain guards (`_seed_problems`,
+  `_validation_episodes_problems`, `_lora_hyperparameter_problems`,
+  `_launch_topology_problems`) derive their scope from the tree, and now
+  recognize a table-driven read (`getattr(spec, field)` over a tuple of field
+  names) as well as a read by name. A provider that forwards a spec on names no
+  field in an attribute access, so it sat outside all four derived scopes: three
+  guards certified a `readers == {...}` sweep that omitted a real reader, and all
+  four were blind to a backend that forwards a gated field and skips the gate.
+  One shared rule now defines both forms for every guard.
--- a/strands_robots/training/base.py
+++ b/strands_robots/training/base.py
@@ -239,6 +239,15 @@ class Trainer(ABC):
     NOT reimplement training and do NOT shell out to a subprocess. Multi-GPU is
     driven via torch's programmatic ``elastic_launch`` (the engine behind
     ``torchrun``), still in-process.
+
+    The field-scoped shared domains below (:meth:`_seed_problems` and its
+    siblings) are each obliged of "a backend that reads the field", and a
+    backend reads a field by any means: naming it (``spec.seed``) or forwarding
+    it through a table (``getattr(spec, field)`` over a tuple of field names,
+    which is how a provider that passes a spec on serializes every field
+    without naming any). The second form is a read for this rule exactly as the
+    first is - what obliges the gate is that the value reaches the run, not the
+    syntax that fetched it.
     """
 
     @property
--- a/tests/training/_spec_field_reads.py
+++ b/tests/training/_spec_field_reads.py
@@ -1,0 +1,84 @@
+"""Discovery rule shared by the field-scoped shared-domain guards.
+
+Each shared numeric domain on :class:`~strands_robots.training.base.Trainer`
+(:meth:`~strands_robots.training.base.Trainer._seed_problems` and its siblings)
+documents a biconditional: a backend that *reads* the field MUST route it
+through the gate, and one that ignores the field MUST NOT report on it. The
+guards that pin the first half derive their scope from the tree rather than
+from a list, so that a new backend which starts reading the field fails them
+until it routes through the gate.
+
+That derivation is only as wide as its notion of "reads the field". A backend
+may read a spec field two ways:
+
+* by name, ``spec.seed`` - an :class:`ast.Attribute` on ``spec``; or
+* through a forwarding table, ``getattr(spec, field)`` over a tuple of field
+  names - which is how a transport-only provider serializes every field it
+  passes on without naming any of them in an attribute access.
+
+A rule that recognizes only the first form reports a complete sweep while a
+table-driven reader sits outside it, so the guards' promise ("a new backend
+that starts reading the field fails this test") does not hold for that backend
+and the biconditional goes unenforced for it. :func:`reads_spec_field`
+recognizes both, and is the one place the two forms are defined so the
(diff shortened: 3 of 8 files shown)


yinsong1986 (approved):
## Summary

Test-only hardening of the four field-scoped shared-domain guards (`_seed_problems`, `_validation_episodes_problems`, `_lora_hyperparameter_problems`, `_launch_topology_problems`): their reader scans previously recognized only a literal `spec.<field>` attribute access, so a provider that forwards spec fields through a table (`getattr(spec, field)` over a tuple of names, which is exactly what `sagemaker.py`'s `_FORWARDED_FIELDS` loop does) sat outside every derived scope. The PR adds one shared `reads_spec_field` rule recognizing both forms, delegates all four guards to it, updates their `readers ==` non-vacuity sets to include the forwarding provider, plants a table-driven defect beside each existing planted defect, and adds an outer discovery test that structurally finds the field-scoped guards and grades each against both read forms. The only production change is a docstring addition to `Trainer` stating the rule. Verified locally at the head SHA: the changed guard suites plus the discovery test pass (558 passed), `ruff check` / `ruff format --check` are clean on the changed files, the docstring cross-reference guards pass, `sagemaker.py` on main does forward all four gated field sets and calls all four gates, and the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention.

## What's good

- The shared rule requires both halves of the table form (a `getattr(spec, ...)` read AND the field named as a string constant), with negative controls pinning each wrong-fix direction, so the widened scan cannot silently regress the by-name scope or over-reach on a string mention alone.
- The outer discovery test derives the set of field-scoped guards structurally rather than listing them, and documents why the learning-rate gate is out of scope - the same derived-not-listed discipline the guards themselves follow.
- Scope discipline: no behavior change anywhere in `strands_robots/` beyond a docstring; the enforcement gap is closed entirely in the test tree.

PR #2521, opened 2026-08-20
fix(tools): name the step that fits a train result, not a load of an absent artifact

cagataycali:
## Problem

`train_policy(action="train")` rendered the completed-run report for **every** non-error
result, so a run with no artifact was reported as a loadable one.

`TrainResult` documents three statuses (`"success" | "running" | "error"`), and two of them
arrive without a checkpoint:

* `running` is the managed-job backend reporting that its job outlived the submitting
  process. `SagemakerTrainer.train` returns it when the local poll budget expires - pinned
  by `test_exhausted_poll_budget_reports_running_not_success`.
* even `success` carries whatever `Trainer.latest_checkpoint` found, which is `None` when the
  run wrote no discoverable checkpoint tree. `LerobotTrainer.train` passes that value
  straight through (`checkpoint_dir=self.latest_checkpoint(spec.output_dir)`), so the default
  provider reaches it too.

Both were reported with the completed-run load instruction, which interpolates the `None`
sentinel:

```
[sagemaker] SageMaker training job 'strands-train-7f3c1a' still running after the 600s local poll budget
job_id: strands-train-7f3c1a
checkpoint_dir: None
metrics: {'sagemaker_status': 'InProgress', 'secondary_status': 'Training'}
Load the result with: create_policy('None')
```

Following that instruction is a measured dead end:

```
>>> create_policy("None")
ValueError: Unknown policy provider: 'None'. Available: ['cosmos3', 'curobo', 'groot', ...]
```

This is the failure `docs/training/overview.md` already warns about - "An unchecked call
hands whatever consumes `checkpoint_dir` a `None` instead" - and the tool was the in-tree
consumer doing it. Meanwhile the result carries the `job_id` this same tool's `status`
action reads, and that step was never named, so the one line an agent acts on pointed at a
provider that does not exist instead of at the poll that does.

The tool already describes an unfinished run correctly on its other path: `action="status"`
emits no load instruction for the same `running` result. The two paths of one tool disagreed
about what a running job means.

## Change

`train` now names the step that fits the result it got:

| result | next step named |
|---|---|
| `success` with a checkpoint | `Load the result with: create_policy('<checkpoint_dir>')` (unchanged) |
| `running` | `The run has not finished ... Poll it with: train_policy(action='status', provider='<p>', job_id='<id>')` |
| `success`, no checkpoint | `The run finished but reported no checkpoint path, so there is nothing to load yet.` |

A private `_next_step_after_train` holds the rule so the three renderings cannot drift, and
its docstring records why the load line is conditional.

**Out of scope, deliberately:** the tool-level `status` key still reports `success` for a
submitted-but-unfinished run. The action is documented as "validate + **launch** training",
and the sibling `action="status"` path likewise maps a running job onto a successful poll
(`"success" if res.status != "error"`). Changing that vocabulary is a separate contract; a
control test pins that this change does not.

## Verification

Measured at `592ee784`.

Pre-fix split - the new tests copied onto `upstream/main` (`db955ff0`): **4 failed / 17
passed**, every failure behavioural and naming the offender:

```
AssertionError: a run that has not finished has no artifact, but the report offered one:
  ... Load the result with: create_policy('None')
AssertionError: no runnable poll step in: ... Load the result with: create_policy('None')
AssertionError: no checkpoint was written, but a load was offered: ... create_policy('None')
```

After the fix: **21 passed**.

Four of the tests pass on both sides and are the boundary:

* the value an artifact-less result interpolates really is a provider `create_policy` refuses
  (the premise that makes this a defect rather than a wording preference);
* a completed run with a checkpoint still names `create_policy('<dir>')` verbatim;
* the run's own status and `checkpoint_dir` still travel verbatim in the `{"json": ...}`
  block, and the tool-level status is still `success`;
* the offered poll step is followed verbatim and answered by this same tool - so the new
  remedy is not a second dead end.

Local gate, 115 test files (everything naming `train_policy` / `TrainResult` /
`checkpoint_dir` / a trainer factory, all of `tests/training` and `tests/tools`, plus the
repo-wide docstring / xref / ASCII / changelog guards), both trees run in lockstep
(217.6 s each):

| | failed | passed | collected |
|---|---|---|---|
| this branch | 3 | 5617 | 5620 |
| `upstream/main` + the new tests | 7 | 5613 | 5620 |

Branch-only failures: **none**. The 4 base-only ids are exactly the pre-fix failures above.
The 3 shared failures are pre-existing and environmental (`tests/training/test_lerobot*.py`
needs `draccus>=0.11.6`; this box resolves 0.8.0).

`ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` = 24
pre-existing errors on both trees, none in the changed files.

## Report, before and after

Same request, same still-running job, same `checkpoint_dir: None`. The capture script parses
whatever next step each tree offers out of the report and **follows it verbatim**, so the
bottom line of each panel is what an agent obeying that report actually gets.

train_policy next-step report

The completed-run report is byte-identical between the two trees - only the artifact-less
case changes.

### DIFF
--- a/changelog.d/2521-train-report-next-step.md
+++ b/changelog.d/2521-train-report-next-step.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **tools/train_policy**: `action="train"` names the step that fits the result it got instead of always naming a checkpoint load. A run with no artifact - a `running` result from the managed-job backend, or a `success` result whose `latest_checkpoint` found nothing - was reported with `Load the result with: create_policy('None')`, which raises `Unknown policy provider: 'None'`. An unfinished run now names the `status` poll for its `job_id`; the tool-level status and the run status in the `{"json": ...}` block are unchanged.
--- a/docs/training/overview.md
+++ b/docs/training/overview.md
@@ -122,6 +122,16 @@ agent("Record 50 cube-pick episodes, then post-tune lerobot ACT on the dataset "
 
 `train_policy` actions: `train`, `validate`, `status`, `export`, `list`.
 
+`train` reports the step that fits the run it got. A finished run names the
+checkpoint to load; a run that has not finished - the managed-job backend
+whose job outlives the submitting process, which returns `running` once its
+local poll budget expires - names the `status` poll for its `job_id` instead,
+and a run that wrote no discoverable checkpoint says so. The tool never offers
+`create_policy(<checkpoint_dir>)` for a result whose `checkpoint_dir` is
+`None`, because that renders as `create_policy('None')` and raises
+`Unknown policy provider: 'None'`. The run's own status always travels verbatim
+in the result's `{"json": ...}` block.
+
 ## Provider-specific knobs
 
 ### LeRobot (`lerobot_local`)
--- a/strands_robots/tools/train_policy.py
+++ b/strands_robots/tools/train_policy.py
@@ -44,6 +44,37 @@ def _err(text: str) -> dict[str, Any]:
     return {"status": "error", "content": [{"text": text}]}
 
 
+def _next_step_after_train(provider: str, res: Any) -> str:
+    """Name the step that fits a train result, never a load of an absent artifact.
+
+    A non-error ``train`` result does not always carry a loadable checkpoint.
+    ``status="running"`` is the backend whose job outlives the submitting
+    process reporting that the job is still going (the SageMaker provider
+    returns it when its local poll budget expires, pollable by ``job_id``), and
+    even ``status="success"`` carries whatever
+    :meth:`~strands_robots.training.base.Trainer.latest_checkpoint` found -
+    which is ``None`` when the run wrote no discoverable checkpoint tree.
+
+    Interpolating that value into a load instruction prints the ``None``
+    sentinel and sends the caller to ``create_policy('None')``, which raises
+    ``Unknown policy provider: 'None'`` - the dead end
+    ``docs/training/overview.md`` warns about when it notes that an unchecked
+    call "hands whatever consumes ``checkpoint_dir`` a ``None`` instead". So the
+    load instruction is named only when there is an artifact to load; a run that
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Makes `train_policy(action="train")` report the next step that actually fits the result instead of unconditionally offering `create_policy('<checkpoint_dir>')` - which, for a `running` result from a managed-job backend or a `success` result whose `latest_checkpoint` found nothing, interpolated the `None` sentinel and sent the caller to `create_policy('None')` (a confirmed `ValueError: Unknown policy provider`). The rule lives in one private helper (`_next_step_after_train`) so the three renderings cannot drift; an unfinished run now names the `status` poll for its `job_id`, and an artifact-less success says so plainly. The tool-level `status` and the `{"json": ...}` block are deliberately unchanged and pinned by control tests.

## What's good

- Verified at head `9dcc2b22`: all 21 tests in `tests/tools/test_train_policy.py` pass, the docstring xref guards resolve the new `:meth:` role (`Trainer.latest_checkpoint` exists at `strands_robots/training/base.py:806`), and `ruff check` / `ruff format --check` are clean on the changed files.
- The offered remedy is itself tested end-to-end: `test_the_poll_step_a_running_run_names_is_one_this_tool_answers` parses the poll instruction out of the report and follows it verbatim, so the fix cannot substitute one dead end for another.
- Scope discipline: the wire-facing pieces (tool-level `status`, json-block fields) are explicitly out of scope and pinned unchanged, so no contract moves; changelog fragment present per AGENTS.md step 3; `docs/training/overview.md` quote ("hands whatever consumes `checkpoint_dir` a `None` instead", line 446) checks out verbatim.
- The premise test (`create_policy(str(None))` raises `Unknown policy provider`) passes on both sides of the fix, which is what makes this a behavioural defect rather than a wording preference.

PR #2522, opened 2026-08-20
docs(use_ros): state the reach a command pre-approval actually has

cagataycali:
## Problem

`STRANDS_ROS2_COMMAND_ALLOW` pre-approves safety-critical ROS 2 command surfaces for headless runs, and it is matched by `_match_blocklist`, which compares **base names** as well as exact names. That breadth is deliberate and load-bearing for the blocklist, where it is fail-safe: one bare `/cmd_vel` entry has to catch `/robot_b/cmd_vel` and every other namespaced drive topic in the graph, which is why the default set is spelled bare.

The same matcher serves the operator's pre-approval list, where the identical breadth points the other way. Measured from the running gate:

| `STRANDS_ROS2_COMMAND_ALLOW` | `/cmd_vel` | `/robot_a/cmd_vel` | `/robot_b/cmd_vel` | `/mobile_base/cmd_vel` |
|---|---|---|---|---|
| `/cmd_vel` | pre-approved | pre-approved | pre-approved | pre-approved |
| `/robot_a/cmd_vel` | still gated | pre-approved | still gated | still gated |

One entry lifts the gate on every robot's drive topic, not the one being driven. Two tests pin that breadth - `test_allowlist_namespaced` and `test_programmatic_drive_and_stop_run_under_the_headless_allowlist[cmd_vel]`, the latter a namespaced `RosBridgedRobot` - so it is the intended behaviour. Sizing that exemption is the operator's job, and every surface they read for it understates the reach:

- `README.md` Configuration row: "Everything else on the safety-critical blocklist stays gated"
- `docs/ros2-integration.md`: "pre-approves those surfaces, everything else stays gated"
- `docs/security.md`: "Pre-approve **individual** surfaces"
- `_gate_command` docstring: "STRANDS_ROS2_COMMAND_ALLOW names it -> allow silently"

An operator following them writes `/cmd_vel` for the one robot that must halt unattended and silently pre-approves the whole fleet's drive topics. `/emergency_stop` and `/joint_trajectory` behave the same way.

## Change

Documentation only. The gate, the shared matcher and both lists are untouched.

- All four surfaces now state the base-name reach and the remedy that scopes an approval to one robot: name the namespace (`/turtle1/cmd_vel`), which is already what the worked turtlesim example does.
- `docs/ros2-integration.md` gains a paragraph next to the "three ways through the gate" table contrasting the two directions of the one matcher - fail-safe on the blocklist, fail-open on the pre-approval list - and repeating that case is never folded.
- `_match_blocklist`'s own docstring records the asymmetry, so a maintainer changing the matcher sees which direction each arm serves. It gains the `Args:`/`Returns:` block it was missing.

Deliberately out of scope: narrowing the allowlist arm to exact matching. It would lose no documented capability, but it changes the meaning of a shipped safety variable and breaks the two tests that pin the breadth today, so it is a contract decision rather than a docs correction.

## Tests

`TestTheDocumentedAllowlistReachIsTheRealReach` in `tests/test_use_ros_command_blocklist.py`, extending the file's existing pattern of grading documented claims against the running gate (`TestTheDocumentedExemptionsAreTheRealOnes`).

The reach is **measured**, not restated: the guard sets the variable to one bare entry, asks the gate about namespaced siblings with `tool_context=None` (so a still-gated surface comes back as the fail-closed error and nothing reaches a transport), and grades every surface naming the variable against the answer. It is graded in both directions, so a gate narrowed to exact matching fails on a leftover namespaced-reach claim instead of passing silently, and the surface set is discovered rather than listed, so a fourth document naming the variable is graded on arrival.

Split against `upstream/main` (test file copied onto the base tree): **1 failed / 107 passed -> 108 passed**.

```
AssertionError: a single STRANDS_ROS2_COMMAND_ALLOW=/cmd_vel entry pre-approves
['/robot_a/cmd_vel', '/robot_b/cmd_vel', '/mobile_base/cmd_vel'] - surfaces it never
names - so the gate's reach is 'base', which these do not state:
['README.md (and calls the pre-approval exclusive)',
 'docs/ros2-integration.md (and calls the pre-approval exclusive)',
 'docs/security.md (and calls the pre-approval exclusive)',
 '_gate_command docstring']
```

The 107 both-tree passes are the evidence that no behaviour moved, and three of the new tests are among them:

- `test_one_bare_entry_pre_approves_every_robots_drive_topic` - the measured fact the documentation has to carry, so a future exact-match gate reports "update the documented reach" rather than leaving the docs stale.
- `test_a_namespaced_entry_scopes_the_approval_to_one_robot` - the remedy the corrected text points at, pinned rather than left to prose.
- `test_the_sweep_reaches_every_operator_facing_surface` - non-vacuity, so a doc reflow that hides the variable fails loudly.

The file's existing guards on the same row also pass unchanged: the halt is still documented as gated and measured as gated, reads are still exempt, and the row's own pre-approval example still lifts the gate.

## Verification

Measured at `d46a7946`; the head has since moved to `f859b526`, which adds the changelog fragment only (`git diff d46a7946..f859b526 -- '*.py' '*.md' ':!changelog.d'` is empty).

- Blast radius: 46 files - every test naming the variable, the gate, the matcher, `use_ros` or a touched docs page, plus the repo-wide docstring / source-string / changelog guards. Branch **1628 passed / 0 failed / 8 skipped**; base with the new test file copied in **1 failed / 1627 passed / 8 skipped**. `1627 + 1 == 1628`, same collected total, no branch-only failure, and no pre-existing failure in the selection.
- `ruff check` / `ruff format --check` and `mypy strands_robots tests tests_integ` clean: **24 == 24** against a pristine `upstream/main` worktree, none in the changed files.
- Docs edits add no non-ASCII, leave the fence count unchanged, and leave every markdown table's cell count consistent.

## Measured reach and the documented claim

The top band is produced by one probe script run against each tree, and the `measured` half is asserted identical in both before the figure is drawn - that equality is the no-behaviour-change control. Only the claim rows differ.

documented reach vs measured reach


yinsong1986 (approved):
## Summary

Docs-only correction plus a doc-vs-gate regression guard: `STRANDS_ROS2_COMMAND_ALLOW` is matched by the same base-name matcher as the blocklist, so a bare `/cmd_vel` entry pre-approves every namespaced `cmd_vel` in the graph, and all four operator/maintainer surfaces (README Configuration row, `docs/ros2-integration.md`, `docs/security.md`, the `_gate_command` docstring) previously described the pre-approval as covering only the listed surfaces. The PR states the real reach and the namespacing remedy on each surface, records the blocklist-fail-safe / allowlist-fail-open asymmetry in `_match_blocklist`'s docstring, and adds `TestTheDocumentedAllowlistReachIsTheRealReach`, which measures the reach from the running gate (with `tool_context=None`, so nothing reaches a transport) and grades every surface naming the variable in both directions. Verified locally: `strands_robots/tools/use_ros.py` is AST-identical to base modulo docstrings, the touched test file passes on the head tree (108 passed) and reproduces the claimed 1 failed / 107 passed when copied onto the base tree, `ruff check` / `ruff format --check` are clean on the changed files, the added lines are ASCII-only, and the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention.

## What's good

- The scope discipline is right: narrowing the allowlist arm to exact matching is correctly identified as a contract change (it would break `test_allowlist_namespaced` and the headless-drive pin) and deliberately left out of a docs correction.
- The reach is measured against the running gate rather than restated, and graded in both directions, so a future exact-match gate fails on a leftover namespaced-reach claim instead of leaving the docs stale.
- `test_the_sweep_reaches_every_operator_facing_surface` pins non-vacuity, so a doc reflow that hides the variable cannot silently turn the sweep into a no-op.
- The remedy the corrected prose points at (`/robot_a/cmd_vel` scopes to one robot) is pinned by a test, not left to prose.

PR #2523, opened 2026-08-20
docs(training): describe both Trainer execution models, and grade the prose against the providers

cagataycali:
## Problem

The `Trainer` ABC documents a single execution model. Four surfaces assert it:

| Surface | Claim |
|---|---|
| `training/base.py` module docstring | "All three run **in-process** (imported and called as libraries, no subprocess)" |
| `Trainer` class docstring | "Concrete trainers are thin adapters that **import the backend package and call its own training function in-process** ... do NOT shell out to a subprocess" |
| `Trainer.train` | "Run the backend's training in-process"; "returns a terminal `TrainResult` (`success`/`error`) ... **there is no detached job to poll**" |
| `Trainer.status` | "Because `train` is synchronous ... this is **only** meaningful for a job launched OUT of band" |

Six of the seven shipped providers work exactly that way. `sagemaker` does not:
it imports no training library, submits the spec as a managed job, and its
`train()` can return `status="running"` with a `job_id` and no `checkpoint_dir`,
because the job outlives the submitting process. That is not incidental - it is
already pinned, by `test_exhausted_poll_budget_reports_running_not_success`
("A job outliving the local budget is 'running' (pollable), never a verdict"),
and `status()` there is a real `DescribeTrainingJob` read of a job `train`
itself submitted.

So the ABC contradicts `TrainResult` in the same file, which declares
`status: "success" | "running" | "error"`, and it contradicts the one provider
whose result a caller most needs to handle. Measured, by reading the statuses
each provider's `train` can reach out of its own source:

| provider | statuses `train()` can return | shape |
|---|---|---|
| `cosmos3`, `fast_sac`, `groot`, `lerobot_local`, `mock`, `ppo` | `error`, `success` | local (blocks) |
| `sagemaker` | `error`, **`running`**, `success` | transport (submits) |

A caller written against the documented contract reads "not `error`" as
finished. #2521 fixed exactly that in the `train_policy` tool, where a
`running` result rendered the completed-run report and offered
`create_policy('None')`. This corrects the contract that told it so.

## Change

Describe both shapes and the return contract each implies, in the four surfaces
above. A **local** trainer imports the backend and calls it in-process, so the
call blocks and the result is terminal. A **transport** trainer imports no
training library and submits a run that outlives the process, so its result may
be `running` with a `job_id` that `status()` polls. `status()` now documents
both the out-of-band job and the one its own `train` handed back. Behaviour is
untouched; every claim here was already true of the code.

`docs/training/overview.md` gains the `sagemaker` row its provider table was
missing and states the one difference a caller has to handle. That page already
said "whether the backend builds the config in-process **or shells out**", and
the README already calls the provider "transport, not behavior" - the ABC was
the surface left behind.

The prose is now **graded rather than trusted**. The statuses each shipped
provider's `train` can return are read out of the code, and the ABC must name
them. The grading is positive on purpose - it asks the prose to cover what the
code does, so it cannot be satisfied by rewording, and a provider that starts
returning a further status is graded on arrival. It keys on the ``literal`` form
this package writes a status value in, because a grader keyed on the bare word
was satisfied by the unrelated phrase "a still-running job" and passed on prose
that never mentioned the return at all.

The scan is scoped to the providers the package ships. The registry is open, and
a sibling test registers a throwaway trainer whose class lives in the test tree;
grading it made the verdict depend on which tests ran first. That is pinned by
`test_a_caller_registered_trainer_does_not_enter_the_grading`.

Finally, `test_trainer_base_contract.py`'s own premise was inverted. It said
"Every shipped backend (mock / lerobot_local / groot / cosmos3 / rl) overrides
`Trainer.status`", so the default was a fallback for some future backend.
Measured: only `mock` and `sagemaker` override it - the default is the live
behaviour of five of seven providers, and each is now driven through it.

## Verification

Measured at `b5a22611`, base `8504f52c`.

Pre-fix split, this test file against `upstream/main`: **3 failed / 13 passed**
-> 16 passed. The headline names the provider and the status it can return:

```
AssertionError: Trainer.train's docstring does not name the status(es) these
providers' train() can return: {'sagemaker': ['running']}. A caller branching
on the documented contract cannot handle a result the ABC never mentions.
```

The 13 both-tree passes are the controls, and they are the proof no behaviour
moved: the four inherited-default tests, the premise that the scan sees both
shapes, the planted-claim scan, the value-vs-word distinction, the
caller-registered-trainer scoping, and the five parametrized providers driven
through the inherited `status()`.

Ordering independence is proven both ways: with the scoping reverted, the
polluted order fails (`mock_done: train() reaches no terminal status at all`)
and the new ordering test fails with it; with it in place that same order is
green (37 passed).

Blast radius (177 files: all of `tests/training`, every test that reads a docs
page or inspects docstrings, `tests/tools/test_train_policy.py`, and the
repo-wide guards): branch **7818 passed / 3 failed**, base **7815 passed /
6 failed**, 7879 collected on both - the 3 extra on base are exactly the pre-fix
failures above (branch-only failures: none), and the
3 shared are the pre-existing `draccus 0.8.0` vs lerobot's `>=0.11.6` local
environment cases in `TestInProcessTrainerCorrectness` / `test_lerobot_e2e`.
`ruff check` + `ruff format --check` clean on `strands_robots tests tests_integ`;
`mypy` 24 == 24 against a pristine base, none in the changed files.

## Artifact

Both columns run the same package code; only the ABC prose differs. The measured
provider table is identical in both (the control), and the claim column goes
0/4 -> 4/4.

Trainer execution model: documented claims vs the registered providers

## Out of scope

`docs/training/overview.md`'s dependency table still says GR00T is "launched as
a subprocess" and cosmos3 is "torchrun-driven; same subprocess-interpreter
rule", while both now run in-process via `call_callable` /
`elastic_launch_callable` and say so in their own module docstrings. That is a
different claim (which interpreter runs the backend, and the venv advice that
follows from it), so it is left alone here rather than folded in.


yinsong1986 (approved):
## Summary

Documentation-only correction to the `Trainer` ABC plus the tests that keep it honest: the module docstring, class docstring, `train()` and `status()` prose previously asserted a single in-process execution model ("terminal result, no detached job to poll"), which is false for the `sagemaker` provider, whose `train()` submits a managed job and can hand back `status="running"` with a `job_id` and no `checkpoint_dir` (verified against `SagemakerTrainer._wait_for_job` on `main`). The rewrite describes both shapes (local vs transport) and tells callers to branch on all three `TrainResult.status` values; `docs/training/overview.md` gains the missing `sagemaker` row. No package behavior changes - the only non-docstring diff lines are in tests and docs.

## What's good

- The new grading tests read the statuses each shipped provider's `train` can reach out of its own source (AST walk following `self._helper` calls), so the docstring cannot silently drift from the providers again - a provider returning a new status is graded on arrival.
- The scan is deliberately scoped to `strands_robots.training.*` classes, with a dedicated ordering-independence test pinning that a runtime-registered trainer from a sibling test cannot change the verdict.
- The `_names_the_status_value` literal-form check closes the loophole where prose containing the English word "running" would satisfy a bare-word grader without documenting the return value.
- Changelog fragment per AGENTS.md step 3; corrected inverted premise in `test_trainer_base_contract.py`'s module docstring (only `mock`/`sagemaker` override `status`, so the default is live behavior for five providers, now each driven through it).
- New Sphinx cross-references (`:class:`~strands_robots.training.sagemaker.SagemakerTrainer``, `:attr:`TrainResult.status``) are contiguous on one line and name real API, per AGENTS.md > Naming & Module Organization.

PR #2524, opened 2026-08-20
fix(examples/fleet): classify a --serve-web bind by address class, not by equality with the default

cagataycali:
## Problem

`--serve-web` decides its network posture with an exact comparison against one
spelling of loopback:

```python
if bind == DEFAULT_BIND:   # DEFAULT_BIND == "127.0.0.1"
    ... tunnel recipe ...
else:
    ... "WARNING: bound to {bind} (network exposure beyond loopback); ..."
```

`127.0.0.0/8` is loopback in its entirety, and the Rerun CLI binds it. Measured
on rerun-sdk 0.26.2, driving the dashboard's own `_start_web_server` readiness
gate with `--bind 127.0.0.2`:

| measured | value |
| --- | --- |
| readiness gate | served, child alive |
| listeners (`ss -ltn`) | `127.0.0.2:43010`, `127.0.0.2:43011` |
| connect from loopback | reachable (both ports) |
| connect from this host's LAN address | **refused (ConnectionRefusedError), both ports** |
| GET the printed URL | HTTP 200, serves the Rerun viewer |
| what the dashboard printed | `WARNING: bound to 127.0.0.2 (network exposure beyond loopback); anyone who can reach ports 43010 and 43011 can watch this dashboard.` |

So the message makes a claim about who can reach the dashboard that is
measurably false, and - because the recipe is gated on the same comparison -
withholds the SSH tunnel lines. On a headless remote host, which is the case
`--serve-web` exists for, that recipe is the actionable half of the output: the
operator is told they are exposed and not told how to reach the viewer.

Two spellings that look affected are not, and both stay out of scope for a
measured reason: `--bind localhost` is refused by the CLI itself
(`error: invalid value 'localhost' for '--bind <BIND>': invalid IP address
syntax`), and no IPv6 bind serves on 0.26.2 (the gRPC listener binds `[::1]`
and the process then exits `Could not parse address`, so the readiness gate
refuses it).

## Change

`is_loopback_bind()` asks the address instead of comparing it to the default, so
the whole loopback class is recognised, and the loopback branch now forwards to
the address the server is actually on - a forward to `127.0.0.1` would reach
nothing on a bind that is not it.

A hostname is deliberately not classified. `--bind` takes an IP literal only, so
parsing the address keeps this classification and the server's own accepted
domain in step; a fixed set of loopback spellings (`localhost`, `127.0.0.1`,
`::1`) would instead print a tunnel recipe for a bind the server refuses.

`url_host()` brackets an IPv6 literal wherever a port follows it. Unbracketed,
`http://::1:9090` parses as neither address nor port, so a classified-loopback
`::1` would have been handed an unopenable URL and an unparseable `ssh -L`
forward. No IPv6 bind is reachable today, so this is not a reachable fix - it is
what keeps the classification and the spelling consistent, and it also removes
the malformed `http://:::9090` the wildcard produced.

The default bind is byte-identical, so the posture that already worked reads
exactly as before.

The README gains the same statement it now makes in output: any loopback
address gets the recipe, and the warning is for a bind that really is wider.

## Tests

`tests/test_fleet_cross_zone_transport.py`, in the existing `--serve-web`
section. Against `upstream/main`: **15 failed / 46 passed** -> 61 passed.

Five failures are behavioural, quoting the measurement:

```
E  AssertionError: bind=127.0.0.2 is loopback, but the startup message calls it
   network exposure: WARNING: bound to 127.0.0.2 (network exposure beyond
   loopback); anyone who can reach ports 9090 and 9876 can watch this dashboard.
E  AssertionError: assert 'http://[::1]:9090/' in 'Rerun web viewer:
   http://::1:9090/?url=rerun%2Bhttp%3A%2F%2F%3A%3A1%3A9876%2Fproxy'
E  AssertionError: no tunnel recipe was printed for a loopback bind: [...]
```

The remaining ten are the classifier's truth table, which names the new helper
and so fails pre-fix on the attribute rather than on behaviour - a root-cause pin
rather than regression evidence.

The 46 both-tree passes are the controls, and each fails for a specific wrong fix:

- `test_the_default_bind_message_is_unchanged` pins all four lines of the default
  output, so widening the classification cannot change the posture that worked.
- `test_a_hostname_is_not_classified_because_the_server_refuses_one` fails for the
  fixed-spelling-set fix, which would classify `localhost`.
- `test_web_viewer_lines_say_so_when_bound_wider_than_loopback` (existing) keeps
  `0.0.0.0` on the warning branch.

## Verification

- 233-file selection (all `tests/test_fleet_*`, every test touching the dashboard
  or reading `examples/`, plus the repo-wide guards), same selection both trees:
  branch **8 failed / 9493 passed / 101 skipped**, base **23 / 9478 / 101**, both
  collecting 9602. Failed delta == passed delta == 15 == the pre-fix count, and
  the branch-only failure set is empty. The 8 shared failures are
  `tests/mesh/test_iot_connect_timeout_domain.py` needing `awsiotsdk`.
- `ruff check` and `ruff format --check` over `strands_robots tests tests_integ`
  clean; `mypy` 24 == 24 against a pristine `upstream/main`, none in the changed
  files.
- Measured at `ee4b6b2`; the head has since moved to `2a0ff9e`, which adds
  only the changelog fragment (no `.py` or docs change), so the numbers above
  still describe the tree.

## Artifact

The real Rerun web server on `--bind 127.0.0.2`, started through the dashboard's
own readiness gate, under both trees. Every measured field - the listeners, the
loopback and off-host probes, the HTTP status of the printed URL - is identical
in the two runs; only the message differs, which is the control.

--serve-web loopback bind classification

### DIFF
--- a/changelog.d/2524-dashboard-loopback-bind-class.md
+++ b/changelog.d/2524-dashboard-loopback-bind-class.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **examples/fleet**: the `--serve-web` startup message classifies the bind
+  address by class rather than by equality with the default, so any loopback
+  address (`127.0.0.0/8`, not only `127.0.0.1`) gets the SSH tunnel recipe
+  instead of a network-exposure warning, and the recipe forwards to the address
+  the server is actually on. A hostname stays unclassified, matching the Rerun
+  CLI's own IP-literal-only `--bind` domain, and an IPv6 literal is bracketed
+  wherever a port follows it. The default bind is byte-identical.
--- a/examples/fleet/README.md
+++ b/examples/fleet/README.md
@@ -206,6 +206,12 @@ STRANDS_MESH_MULTICAST=true python examples/fleet/dashboard.py --serve-web
 ssh -N -L 9090:127.0.0.1:9090 -L 9876:127.0.0.1:9876 user@remote-box
 ```
 
+Any loopback address gets that tunnel recipe, not just the default one:
+`--bind` takes an IP literal, and the whole `127.0.0.0/8` block is reachable
+only from the host, so `--bind 127.0.0.2` (isolating the viewer on its own
+loopback address) prints the recipe forwarded to `127.0.0.2`. The
+network-exposure warning is for a bind that really is wider.
+
 The web viewer is a view, not a surface: it changes only the render
 transport, and the mesh peer stays subscribe-only exactly as above. With
 rerun-sdk absent, `--serve-web` fails with the install hint rather than
--- a/examples/fleet/dashboard.py
+++ b/examples/fleet/dashboard.py
@@ -57,6 +57,7 @@ import os
 os.environ.setdefault("STRANDS_MESH_LOCAL_DEV", "1")
 
 import argparse
+import ipaddress
 import subprocess
 import sys
 import tempfile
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Replaces the `--serve-web` startup message's equality check against the default bind (`bind == "127.0.0.1"`) with class membership via `ipaddress.ip_address(bind).is_loopback`, so any `127.0.0.0/8` (or `::1`) bind gets the SSH tunnel recipe instead of a false network-exposure warning, and the recipe now forwards to the address the server is actually on. `url_host()` brackets IPv6 literals so the printed URL and `ssh -L` spec stay parseable. The change is confined to an example's message rendering plus tests, README, and a changelog fragment - no public API, wire format, or subprocess input surface changes (the bind was already passed to argv-style `serve_web_command` unchanged). Verified locally at head `2a0ff9ee`: all 21 tests in the `--serve-web` message section pass, including the control pinning the default-bind output byte-identical.

## What's good

- The deliberate non-classification of hostnames (parse the address rather than match a spelling set) keeps the classifier and the Rerun CLI's accepted `--bind` domain from diverging, and a control test pins that choice.
- `test_the_default_bind_message_is_unchanged` pins all four lines of the previously-working posture, so the widened classification cannot regress it.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; measured claims (rerun-sdk 0.26.2 bind behaviour, `localhost` refusal, IPv6 non-serving) are quoted in docstrings where the next reader needs them.

PR #628, opened 2026-06-24
feat(training): Trainer abstraction — one interface over LeRobot / GR00T / Cosmos3 post-tuning

cagataycali:
## What

Adds a `Trainer` abstraction — the **training-side peer of `Policy`** — that post-tunes a policy through one provider-agnostic interface and produces a checkpoint that loads straight back via `create_policy`. The same provider name selects both the trainer and the inference policy, so the **record → train → export → deploy** loop closes with a single string swap.

```python
from strands_robots.training import create_trainer, TrainSpec

trainer = create_trainer("lerobot_local")        # same name as create_policy(...)
spec = TrainSpec(dataset_root="/tmp/ds", base_model="lerobot/smolvla_base",
                 output_dir="/tmp/ft", steps=20000, extra={"policy_type": "smolvla"})
result = trainer.train(spec)
policy = create_policy(result.checkpoint_dir)     # loop closed
```

One interface wraps three genuinely different upstream pipelines, each driven **as a native Python library, in-process** (no `subprocess`, no `torchrun` binary, no `python_executable`, no argv):

| Provider | Native pipeline driven (library call) |
|---|---|
| `lerobot_local` | `lerobot.scripts.lerobot_train.train(cfg)` with a typed `TrainPipelineConfig` |
| `groot` | `gr00t.experiment.experiment.run(config)` (real `FinetuneConfig` → `Config`) |
| `cosmos3` | `convert_model_to_dcp(Args)` → `train.launch(config, args)` → `export_model(Args)` |
| `mock` | in-proc, dep-free (for tests) |

Multi-GPU single-node uses `torch.distributed.launcher.api.elastic_launch` (the programmatic engine behind `torchrun`) calling a Python worker callable. A new `training/_inproc.py` holds the shared in-process + elastic-launch primitives. Replaces the old `tools/lerobot_train.py` with the provider-agnostic `tools/train_policy.py` agent tool (`train`/`validate`/`status`/`export`/`list`).

## Interface correctness (from review)

- **`latest_checkpoint()` is a public ABC method** (default `None`), implemented by all four backends — `export` uses it instead of duck-typing a private name. lerobot splits the loadable-dir (`latest_checkpoint`) from the resume-config-file (`_resume_config_path`).
- **`num_nodes > 1` rejected** in lerobot and GR00T `validate()`; cosmos3 honors it via HSDP.
- **synchronous `train()` contract**: blocks and returns the full metrics verdict; docstrings + default `status()` are honest about there being no detached job to poll.
[...]

PR #2525, opened 2026-08-20
docs(training): describe the interpreter a trainer actually imports its backend in

cagataycali:
## Problem

`docs/training/overview.md` still describes the execution model the branch that
introduced the page refactored away before it merged. Inside that branch the
trainers were moved to driving their backends as libraries - `7b315ce2`
("run LeRobot training in-process (no subprocess)"), `ca53727a` ("run GR00T +
Cosmos3 in-process too (drop subprocess/torchrun)") - and `b0d24aaf` then
"drop[ped] python_executable + runpy". The whole branch squashed into #628, so
the shipped code has never had either. The page kept the pre-refactor guidance:

| Surface | Says | Reality |
|---|---|---|
| `groot` dependency row | "Isaac-GR00T checkout + **its own venv**"; "launched as a subprocess, so it uses GR00T's interpreter, not ours" | `gr00t` is `importlib.import_module`'d in the calling interpreter |
| `cosmos3` dependency row | "torchrun-driven; same subprocess-interpreter rule" | `cosmos_framework` likewise; multi-GPU uses torch's programmatic `elastic_launch` |
| interpreter callout | "`LerobotTrainer` / `Gr00tTrainer` / `Cosmos3Trainer` accept a `python_executable=` argument (defaults to `sys.executable`)" | no trainer declares it, and `python_executable` appears nowhere in `strands_robots/training/` in the entire history |

The argument is not merely absent - it is **absorbed**. Every trainer takes
`**kwargs`, so the constructor call a reader would make succeeds and the value is
dropped with no raise and no log line:

```
Gr00tTrainer(python_executable="/venv/bin/python")  -> constructed OK, attribute ABSENT
```

Following the page end to end, with the checkout in its own venv as instructed:

```
validate() -> []                                  # clean preflight, no mention
train()    -> status="error"
              "Isaac-GR00T is not importable from this interpreter. Install it
               from source ... into the *same* Python that imports strands_robots"
```

So the page sends an operator to a second interpreter, the runtime asks for the
same one, and the knob the page offers to bridge them is silently a no-op.

Five surfaces in the package already say the opposite, which is why this is a
documentation correction rather than a design question:

- `Gr00tTrainer.__init__` docstring: "`GR00T_ROOT` is for runtime config resolution, **not the interpreter path**"
- `groot._INSTALL_HINT` / `cosmos3._INSTALL_HINT`: "into the ***same*** Python that imports strands_robots"
- `groot.py` module docstring: "same interpreter, no argv, no nested `python`"
- `lerobot.py` module docstring: "**directly in this interpreter**"
- `training/base.py`: "NOT shell out to a subprocess"

Adding the parameter instead would resurrect an execution model that branch
deliberately removed, so the code stays as it is.

## Change

`docs/training/overview.md` only.

- The `groot` and `cosmos3` dependency rows now state that the backend is imported
  in the calling interpreter and has to be installed into the same environment,
  and that `GR00T_ROOT` / `COSMOS_ROOT` resolve the checkout for relative configs
  rather than selecting an interpreter.
- The **Subprocess interpreter** callout becomes **One interpreter**: there is no
  second interpreter to point at and no `python_executable=` argument, the
  absorption is named so a reader who tried it knows why nothing happened, and the
  refusal `train()` actually reports is quoted.

The capability table higher up the page is deliberately untouched: its columns
describe each *upstream* project ("Upstream entry point", "Launcher"), and the
prose directly beneath it already states that the three local backends drive
their libraries in-process. The new guard is scoped to the dependency table by
that table's own header so it does not grade the upstream one as install
guidance.

## Guard

`tests/training/test_documented_trainer_interpreter_contract.py` grades the prose
against the code, in two directions, both derived rather than listed:

1. **A documented `name=` argument is an explicit parameter of the class the prose
   names.** `**kwargs` is excluded from the declared set on purpose - absorbing a
   keyword is not accepting it, and counting it is what let an argument no
   constructor declares read as supported. Class names are resolved over the
   containing paragraph, because the subject precedes the verb
   ("A / B / C accept a `x=` argument").
2. **A provider whose module can select no interpreter says so in its install
   row.** "Can select no interpreter" is measured from the AST - the module
   references neither `sys.executable` nor anything in `subprocess` - not from
   other prose. Stated positively so the row has to carry the requirement, and so
   a correctly qualified mention of the alternative does not trip it.

Both walk the trainer registry, so a provider added later is graded on arrival
with no edit here.

## Verification

Measured at `2a4acb26`, `upstream/main` = `227faf14`.

Pre-fix (both files copied onto `upstream/main`): **3 failed / 7 passed**. The
headline names all three classes:

```
documentation promises trainer constructor arguments that are not declared
(each is absorbed by **kwargs and dropped):
{'docs/training/overview.md': ['python_executable= on Cosmos3Trainer, Gr00tTrainer, LerobotTrainer']}
```

Post-fix: **10 passed**. The seven that pass on both trees are the controls, and
each one fails for a specific wrong fix:

- the absorption premise (`groot`, `cosmos3`) - fails if a constructor starts
  refusing unknown keywords, which would make `**kwargs` exclusion the wrong rule;
- no trainer reaches for `sys.executable` or `subprocess` - fails if a trainer can
  select an interpreter again, which is when the same-environment guidance needs
  revisiting;
- the runtime install hints still ask for the same interpreter - the message the
  guidance has to agree with;
- a planted acceptance claim is reported, a denial of the same argument is not, and
  a *declared* argument (`groot_root=`) is not - so the rule cannot be satisfied by
  rewording, and the corrected page's own explicit denial is correctly not a claim;
- non-vacuity: the graded corpus really contains the page and the providers.

Blast radius (160 files: all of `tests/training`, every test naming the trainer
classes / the registry helpers / a docs page, plus the repository-wide docstring,
source-string and changelog guards), the identical selection on both trees:

|---|---|---|---|---|
| this branch | 3 | 6202 | 33 | 6205 |
| `upstream/main` + both files | 6 | 6199 | 33 | 6205 |

Branch-only failures: **none**. Base-only: exactly the 3 above. The 3 shared are
pre-existing on `upstream/main` (`TestInProcessTrainerCorrectness` x2 plus
`test_lerobot_e2e`, all the locally installed draccus predating lerobot's
`>=0.11.6` floor).

`ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ`
= 24 errors on both trees, none in the changed file.

## What it looks like

One capture script, run against each tree, reading **that tree's own page**,
extracting the guidance, following it, and recording the outcome. The package code
is identical in both columns - `interpreter consulted`, `validate()` and the
`train()` refusal are byte-identical, which is the control - so the only variable
is what the page told the operator to do.

trainer interpreter guidance, before and after

## Scope

Refusing unknown constructor keywords across the seven trainers would turn this
class of mistake into an error rather than a silent drop, but it changes behaviour
for existing callers (`_groot_worker` constructs `Gr00tTrainer(groot_root=...)`)
and is a decision for all seven at once, so it is left alone here and the guard
pins the absorption as a measured premise instead.

### DIFF
--- a/changelog.d/2525-trainer-interpreter-docs.md
+++ b/changelog.d/2525-trainer-interpreter-docs.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **docs/training**: the dependency table and interpreter callout described a subprocess execution model the trainers do not have, and advertised a `python_executable=` constructor argument none of them declares - every trainer's `**kwargs` absorbed it silently, so following the guidance installed the backend into an interpreter the trainer never consults. The `groot` and `cosmos3` rows now state that the backend is imported in the calling interpreter, and a new guard grades a documented `name=` argument against the class's real parameters.
--- a/docs/training/overview.md
+++ b/docs/training/overview.md
@@ -476,8 +476,8 @@ on an L40S GPU:
 | `lerobot_local` + ACT / diffusion | `pip install 'strands-robots[lerobot]' 'lerobot[training]'` | `[lerobot]` supplies torch + torchcodec + datasets; it does **not** supply `accelerate` |
 | `lerobot_local` + `smolvla` | `pip install 'strands-robots[lerobot]' 'lerobot[training]' 'lerobot[smolvla]'` | lerobot 0.6's `[smolvla]` extra layers `transformers>=5.4.0,<5.6.0` + num2words on top. Do **not** pin `transformers==5.3.0` - it conflicts with lerobot 0.6's transformers floor. |
 | `lerobot_local` + `pi0` / `pi05` | `pip install 'strands-robots[lerobot]' 'lerobot[training]' 'lerobot[pi]'` | lerobot 0.6's `[pi]` extra (same `transformers>=5.4.0,<5.6.0` range + scipy) |
-| `groot` | Isaac-GR00T checkout + its own venv (`omegaconf`, `tyro`, …); point `extra["groot_root"]` / `GR00T_ROOT` at it | launched as a subprocess, so it uses GR00T's interpreter, not ours |
-| `cosmos3` | cosmos-framework checkout (`uv sync --group=cu130-train`); point `extra["cosmos_root"]` / `COSMOS_ROOT` at it | torchrun-driven; same subprocess-interpreter rule |
+| `groot` | Isaac-GR00T checkout, installed with `pip install -e` into the **same** environment as `strands_robots` (it pulls `omegaconf`, `tyro`, …); point `extra["groot_root"]` / `GR00T_ROOT` at the checkout | `gr00t` is imported in the calling interpreter, so it has to be importable there; `GR00T_ROOT` resolves relative configs, not the interpreter |
+| `cosmos3` | cosmos-framework checkout (`uv sync --group=cu130-train`), installed into the **same** environment as `strands_robots`; point `extra["cosmos_root"]` / `COSMOS_ROOT` at the checkout | `cosmos_framework` is imported in the calling interpreter; multi-GPU goes through torch's programmatic `elastic_launch`, not a `torchrun` binary |
 
 > **torchcodec / torch ABI:** the lerobot training dataloader decodes video via
 > `torchcodec`, whose compiled `.so` must match the **exact** installed torch
@@ -487,10 +487,15 @@ on an L40S GPU:
 > training fails with a generic non-zero exit. Pin `torch` + `torchcodec`
 > together (verified-good combo: `torch==2.10.0+cu128` + `torchcodec==0.10.0`).
 
-> **Subprocess interpreter:** `LerobotTrainer` / `Gr00tTrainer` / `Cosmos3Trainer`
-> accept a `python_executable=` argument (defaults to `sys.executable`). Set it
-> to a venv that has the provider's deps if your agent process runs in a
-> different environment — the training pipeline runs in that interpreter.
+> **One interpreter:** `LerobotTrainer` / `Gr00tTrainer` / `Cosmos3Trainer` call their
+> backend as a library in the **same** interpreter that imports `strands_robots`, so
+> there is no second interpreter to point them at. There is no `python_executable=`
+> argument either, and because each constructor absorbs unknown keywords, passing one
+> is silently a no-op rather than an error. Install the provider's deps into the
+> environment your agent process runs in; otherwise `train()` reports
+> `<package> is not importable from this interpreter` in `TrainResult.message`.
+> `GR00T_ROOT` / `COSMOS_ROOT` (and `extra["groot_root"]` / `extra["cosmos_root"]`)
+> resolve the checkout so relative configs load - they are not interpreter paths.
 
 ## See also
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Docs-only correction plus a derived guard. `docs/training/overview.md` still described the pre-#628 execution model (per-provider venvs, subprocess/torchrun, a `python_executable=` constructor argument) that the shipped code never had; the diff rewrites the `groot`/`cosmos3` dependency rows and the interpreter callout to match the in-process reality, and the corrected prose agrees with the code (`importlib.import_module` in the calling interpreter, programmatic `elastic_launch`, `_INSTALL_HINT` asking for the same interpreter — all verified against `strands_robots/training/{groot,cosmos3}.py` at this head). The new `tests/training/test_documented_trainer_interpreter_contract.py` grades the prose in both directions, walks the trainer registry rather than a hardcoded list, and correctly excludes `**kwargs` absorption from the declared-parameter set.

## What's good

- Verification claims reproduce exactly: copying the two files onto the base (`227faf14`) yields 3 failed / 7 passed with the headline naming all three classes; at head, 10 passed. `ruff check` / `ruff format --check` clean on the new file.
- The guard is derived, not listed (registry walk + `inspect` + AST), so a later provider is graded on arrival, and the planted-claim/denial/declared-argument controls pin the regex against both false positives and rewording.
- Scope discipline: the code's `**kwargs` absorption is measured as a premise rather than changed, with the seven-trainer refusal decision explicitly deferred — consistent with the PR description.
- The dependency-table guard is scoped by the table's own header so the upstream capability table is not misgraded as install guidance.

yinsong1986 (approved):
## Summary

Docs-only correction plus a derived regression guard. `docs/training/overview.md` described a pre-#628 execution model the shipped code never had: per-provider venvs for `groot`/`cosmos3`, subprocess/torchrun launching, and a `python_executable=` constructor argument that no trainer declares (every trainer's `**kwargs` absorbs it silently). The diff rewrites the two dependency rows and the interpreter callout to match the in-process reality, and the new test grades the prose against the code in both directions - a documented `name=` argument must be an explicit parameter of the class the paragraph names (`**kwargs` deliberately excluded), and a provider whose module references neither `sys.executable` nor `subprocess` must have its dependency row say the backend is imported in the calling interpreter. Verified against the head: `python_executable` appears nowhere under `strands_robots/training/` (only in the unrelated vera policy, which genuinely spawns a server subprocess), both trainer constructors are `(root, **kwargs)` sinks, the `_INSTALL_HINT` strings ask for the same interpreter, and `cosmos3` multi-GPU goes through the programmatic `elastic_launch` as the row now states. The new test file passes 10/10 on the head, is ruff-clean, ASCII-only, contains no host paths, and both assertions walk the trainer registry rather than a hardcoded list, so a provider added later is graded on arrival. Changelog fragment present per AGENTS.md step 3. No behavioural change to the package; safe to merge.

## What's good

- The guard's controls are unusually thorough: the absorption premise is pinned separately (so `**kwargs` exclusion is revisited if a constructor starts refusing unknown keywords), a planted claim is reported while a denial and a declared argument are not, and non-vacuity asserts the corpus actually contains the page and the providers.
- Scope discipline: refusing unknown constructor keywords across all seven trainers was correctly identified as a behaviour change for existing callers and left out, with the absorption pinned as a measured premise instead.
- The dependency-table read is anchored to that table's own header so the upstream capability table is not graded as install guidance.

PR #2526, opened 2026-08-20
fix(training): resolve a runtime-registered trainer through the public resolver

cagataycali:
## Problem

`list_trainers()` advertises seven providers. `create_trainer` resolves all seven.
`import_trainer_class` - the public answer for a caller that wants the `Trainer`
class without paying for construction - refuses four of them:

```
provider        declared in         create_trainer   import_trainer_class
cosmos3         policies.json       resolves         resolves
fast_sac        register_trainer    resolves         ValueError: No trainer registered for provider 'fast_sac'
groot           policies.json       resolves         resolves
lerobot_local   policies.json       resolves         resolves
mock            policies.json       resolves         resolves
ppo             register_trainer    resolves         ValueError: No trainer registered for provider 'ppo'
sagemaker       register_trainer    resolves         resolves
```

`import_trainer_class` consults the `"trainer"` block in `policies.json` and then
auto-discovery on `strands_robots.training.<provider>`. Neither rung reaches the
runtime registry, so every provider declared with `register_trainer` is refused -
including two the package registers itself in `training/__init__.py`. `ppo` and
`fast_sac` live in the `training.rl` subpackage, so auto-discovery on
`strands_robots.training.ppo` finds nothing. `sagemaker` resolves only because its
module happens to sit at `strands_robots.training.sagemaker`, which the registration
comment already notes ("Auto-discovery would resolve `create_trainer("sagemaker")`
from the module name alone").

The same gap hits the documented user route. `register_trainer`'s own docstring
example registers `"my_vla"` with the alias `"mv"`; neither spelling resolves
through `import_trainer_class`.

The refusal then over-promises, because it builds its list from `list_trainers()`:

```
ValueError: No trainer registered for provider 'fast_sac'.
Available trainers: ['cosmos3', 'fast_sac', 'groot', 'lerobot_local', 'mock', 'ppo', 'sagemaker']
```

The message enumerates the provider it has just rejected, so the one list a caller
has to recover from the failure names four providers this resolver cannot serve.

## Change

`import_trainer_class` gains the runtime rung as step 1 - alias resolved first, and
ahead of the `policies.json` block so re-registering a name still overrides a
shipped `trainer` block rather than being ignored by it. `create_trainer` then
delegates wholly to it, dropping its own copy of the same lookup.

The runtime lookup now exists once. That is what the module docstring already
promised (`"Mirrors strands_robots.policies.factory exactly"`) and how the policy
side is built: `create_policy` adds nothing but the constructor call and the
trust-remote-code gate on top of a single resolver whose first rung is the runtime
registry.

`create_trainer`'s behaviour is unchanged - it consulted the runtime registry first
before, and the shared resolver keeps that precedence. Verified: it returns the
identical class for all seven providers on both sides.

Docstrings updated in the same change: `import_trainer_class`'s stated resolution
order listed two rungs where the resolver behind `create_trainer` had three, and the
module docstring described only the `policies.json` route. No prose page documents
this resolver, so the docstrings are its documentation surface.

## Evidence

trainer resolver, before and after

One probe script run against `upstream/main` and against this branch. Package code
is the only variable; the probe registers `my_vla`/`mv` exactly as the factory's own
docstring example does.

|---|---|---|
| listed providers `import_trainer_class` refuses | `fast_sac`, `ppo`, `my_vla`, `mv` | none |
| names the refusal advertises but cannot resolve | `fast_sac`, `ppo`, `my_vla`, `mv` | none |
| `create_trainer` class per provider | identical | identical |

The last row is the control: this branch changes which rungs
`import_trainer_class` consults, not what any provider resolves to. There is no
simulation behaviour in the diff, so there is no rollout to film.

## Tests

Extended `tests/training/test_factory.py`. Against `upstream/main`: **6 failed,
32 passed**; after: **38 passed**, and unchanged under random ordering.

The failures are behavioural and name the providers:

```
test_every_listed_trainer_resolves
  list_trainers() advertises providers import_trainer_class refuses:
  {'custom_x': ..., 'cx': ..., 'fast_sac': ..., 'ppo': ...}

test_the_refusal_advertises_only_names_it_can_serve
  the refusal offers providers it cannot resolve: ['custom_x', 'cx', 'fast_sac', 'ppo']

test_the_two_entry_points_agree_on_every_listed_name
  (name, import_trainer_class, create_trainer) disagree:
  [('custom_x', False, True), ('cx', False, True), ('fast_sac', False, True), ('ppo', False, True)]
```

The properties are derived from `list_trainers()` at run time rather than from a
literal, so a provider added later is graded without editing the test.
`test_at_least_one_listed_trainer_is_runtime_only` keeps that from going vacuous -
were every provider declared in `policies.json`, the properties would hold for a
resolver that consults only the JSON rung.

Two tests pass on both sides and are the no-overreach controls:
`test_an_unknown_provider_is_still_refused` (the fix must not make the resolver
permissive) and `test_at_least_one_listed_trainer_is_runtime_only`.

`test_a_runtime_registration_overrides_a_json_trainer_block` pins the rung
*ordering*, which the two shapes of a wrong fix are measured against:

- rung removed entirely -> 9 failed (the three pre-existing tests that reach
  `ppo`/`fast_sac`/a runtime alias through `create_trainer` fail too, which is what
  shows the delegation is real)
- rung merely demoted below the JSON rung -> exactly 1 failed, that test

`TestAutoDiscoveryFallback`'s docstring named its rung by number
("Resolution-order step 2"); it is now the last rung, and the docstring says so
instead of counting.

## Verification

Measured at `44cbd63e` on `/tmp` worktrees of `upstream/main` (`4e2e89f1`).

- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ`: clean
  `upstream/main` worktree, none in the changed files
- 71-file selection (every test naming the factory surface, all of
  `tests/training`, `tests/tools/test_train_policy.py`, and the repo-wide
  docstring / changelog / source-string guards), both trees, same selection:

  |---|---|---|---|---|
  | this branch | 3 | 3547 | 5 | 3550 |
  | `upstream/main` + this test file | 9 | 3541 | 5 | 3550 |

  Branch-only failures: none. The six base-only failures are exactly the pre-fix
  ones above. The three shared failures are pre-existing and environmental -
  `test_lerobot_e2e.py::test_record_train_load_loop` and two
  `TestInProcessTrainerCorrectness` cases need `draccus>=0.11.6`, and this
  machine resolves 0.8.0.

### DIFF
--- a/changelog.d/2526-trainer-resolver-runtime-registry.md
+++ b/changelog.d/2526-trainer-resolver-runtime-registry.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- **training**: `import_trainer_class` now resolves a provider registered with
+  `register_trainer`, so it serves every name `list_trainers()` advertises.
+  It consulted only the `policies.json` `"trainer"` block and auto-discovery on
+  `strands_robots.training.<provider>`, which refused the runtime-registered
+  providers - including `ppo` and `fast_sac`, whose modules live in the
+  `training.rl` subpackage - while `create_trainer` resolved them; the refusal
+  it raised also listed those providers as available. The runtime lookup now
+  lives once in `import_trainer_class`, which `create_trainer` delegates to,
+  and keeps its precedence over a shipped `trainer` block.
--- a/strands_robots/training/factory.py
+++ b/strands_robots/training/factory.py
@@ -17,6 +17,13 @@ Mirrors ``strands_robots.policies.factory`` exactly, but resolves the
 so a single provider name owns BOTH the inference class
 (``create_policy("lerobot_local")``) and the training class
 (``create_trainer("lerobot_local")``).
+
+A provider can also be declared at runtime with :func:`register_trainer`, which
+is how the ``training.rl`` backends (``ppo``, ``fast_sac``) and the ``sagemaker``
+transport are wired without a paired inference provider to hang a JSON block
+off. Both routes resolve through :func:`import_trainer_class`, so
+:func:`list_trainers`, :func:`import_trainer_class` and :func:`create_trainer`
+answer for the same set of names.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Consolidates the trainer resolution logic into a single resolver: `import_trainer_class` gains the runtime-registry rung (alias-first, ahead of the `policies.json` `"trainer"` block, matching `create_trainer`'s pre-existing precedence), and `create_trainer` delegates wholly to it. This closes the gap where `list_trainers()` advertised runtime-registered providers (`ppo`, `fast_sac`, and anything registered via the documented `register_trainer` route) that `import_trainer_class` refused - with a refusal message that enumerated the very names it could not serve. `create_trainer`'s observable behaviour is unchanged; the only widening is that `import_trainer_class` now serves the full advertised set.

## What's good

- Verified locally at head `b9faa37a`: `tests/training/test_factory.py` passes 38/38, stable under random ordering; `ruff check` + `ruff format --check` clean on both changed files.
- The rung-ordering test (`test_a_runtime_registration_overrides_a_json_trainer_block`) pins the precedence that distinguishes this fix from the two plausible wrong versions (rung removed / rung demoted), and cleans up its `mock` override in a `finally`.
- The agreement properties are derived from `list_trainers()` at run time rather than a literal, with `test_at_least_one_listed_trainer_is_runtime_only` guarding against vacuity - a provider added later is graded without editing the test.
- Changelog fragment present per AGENTS.md step 3; no non-ASCII in any changed file; no new untrusted-input surface (the runtime rung only invokes loaders the caller registered in-process).

PR #2527, opened 2026-08-20
fix(episode-labels): grade an episode index that arrives inside a collection

cagataycali:
## Problem

`tests/simulation/test_replay_episode_index_domain.py` pins one contract: every public
surface that resolves an episode index applies the shared
`non_negative_whole_number_error` rule. `episode` selects *which* recorded trajectory a
call is about, so a value outside that domain is not a slow answer - `int(True)` is `1`
and every episode table is indexable by 1.

The sweep that keeps that true discovered a surface only when a parameter was literally
named `episode`. An index also arrives inside a **collection**:

- an entry of the `episodes` list `record_deterministic_verdicts` records (`episodes[i]["episode"]`),
- a key of the `human_labels` mapping `measure_agreement` calibrates against.

Both apply the shared rule today, and both were in neither the discovered set nor the
registered one, so:

| | upstream/main | this branch |
|---|---|---|
| episode-index surfaces in the tree | 11 | 11 |
| graded by the sweep | **9** | **11** |
| ungraded | `record_deterministic_verdicts`, `measure_agreement` | none |
| both guards deleted -> sweep says | **203 passed** (clean tree) | **16 failed**, naming both |

`test_the_surface_set_is_exactly_the_pinned_one` - whose own docstring is *"a new surface
joins the sweep or fails this test"* - reported a clean tree, and
`test_every_surface_validates_or_forwards` never ran on either surface. Deleting both
guards is invisible to CI, and the behaviour they hold degrades exactly as the table at
the top of that file describes:

```
record_deterministic_verdicts(root, [{"episode": True, "success": False}])
  -> sidecar keys ['1']        a flag labelled EPISODE 1; episode 0 has no verdict at all

measure_agreement(root, {True: {"quality": "high"}})
  -> quality_agreement 1.0     calibrated against episode 1, not the episode keyed
```

The second is the number that decides whether a judge is trusted to filter training data.

`measure_agreement` also reported only `human_labels key`, naming the container slot
rather than the episode index it holds. That is not only a message gap: the sweep's
recogniser requires the shared rule to be applied *to the episode*, so it could not
recognise the guard that was already there.

## Change

- Discovery is derived from each surface's own `Args:` entry: a parameter named `episode`,
  or one whose documented description says it carries an episode index. That keeps the
  rule a property of the tree rather than a list to maintain, and holds the collection
  spellings to the same contract. `filter_episodes` stays out on that same rule - it
  *returns* indices rather than resolving one, so requiring a guard there would demand
  one on a value it never receives.
- The two surfaces join `_REPLAY_EPISODE_SURFACES`.
- `measure_agreement`'s refusal names the quantity (`human_labels episode index`), and its
  `Args:` entry states the domain.
- `docs/data/episode-labels.md` gains an `episode_index` row in the field-domain table,
  stating that the domain holds in each spelling the index arrives in.

`_validates_or_forwards` is unchanged: the recogniser still requires the shared rule to be
applied to the episode, which is what makes naming the quantity the fix rather than
loosening the check.

## Tests

`tests/simulation/test_replay_episode_index_domain.py`, **2 failed / 220 passed -> 222 passed**:

- `test_every_surface_validates_or_forwards[measure_agreement]` - the newly graded surface,
  pre-fix `neither validates nor forwards 'episode'`.
- `test_both_refusals_name_the_episode_index` - pre-fix
  `measure_agreement: human_labels key must be a non-negative whole number, got True.`

The 220 that pass on both trees are the no-behaviour-change evidence. New alongside them:

- `test_discovery_finds_an_index_that_arrives_inside_a_collection` - a planted
  collection-valued surface is discovered, and an unguarded one does not pass.
- `test_discovery_leaves_a_surface_that_only_returns_indices_out` - the boundary, on the
  same derived rule.
- `TestTheCollectionSpellingsRefuseAnUnusableIndex` - drives the real functions over
  `True / False / 2.5 / -1 / "0" / None / [0]`, asserts a refused verdict writes no
  sidecar, and keeps a control that an index inside the domain still resolves that episode.

Load-bearing: deleting the guard at both collection surfaces fails the widened sweep both
structurally (both surfaces named) and behaviourally, where it left the shipped sweep at
203 passed.

## Verification

Measured at `6b1e7e2e`. Blast radius - every test naming the episode-label surfaces, the
shared rule, the replay spellings, plus the repo-wide docs/docstring/changelog guards
(58 files): branch **2604 passed / 0 failed / 2 skipped**, `upstream/main` with the new
guard file copied in **2602 passed / 2 failed / 2 skipped**; 2606 collected on both, the
failed and passed deltas are both 2 (the pre-fix count), no branch-only failure, and no
failure shared by the two trees. `ruff check` / `ruff format --check` clean over
`strands_robots tests tests_integ`; `mypy` 24 == 24 against a pristine `upstream/main`
worktree with nothing branch-only.

episode-index sweep coverage

Both columns come from one capture script run once per tree: it loads that tree's own
discovery and recogniser, enumerates the real surfaces, drives the two refusals, and
deletes the guards to see whether that tree's sweep notices.
`record_deterministic_verdicts`' refusal is byte-identical in both columns - it already
named the episode index - so the only refusal that changes is the one that did not.

### DIFF
--- a/changelog.d/2527-episode-index-collection-spelling.md
+++ b/changelog.d/2527-episode-index-collection-spelling.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **episode-labels**: the structural sweep that pins the shared non-negative-whole-number domain for an episode index now grades an index that arrives inside a collection - an entry of the `episodes` list `record_deterministic_verdicts` records, a key of the `human_labels` mapping `measure_agreement` calibrates against. Discovery used to admit a surface only when a parameter was literally named `episode`, so those two were graded by nothing: the completeness test reported a clean tree over 9 of 11 surfaces and deleting both guards left the file passing. It is now derived from each surface's own `Args:` entry, and `measure_agreement`'s refusal names the episode index rather than only `human_labels key`.
--- a/docs/data/episode-labels.md
+++ b/docs/data/episode-labels.md
@@ -67,6 +67,7 @@ Field domains:
 
 | field | domain |
 |---|---|
+| `episode_index` | a non-negative whole number, on the shared domain every surface applies. Holds in each spelling the index arrives in: the `episode` argument of `deterministic_verdict` / `annotate_episode` and the judge tools, an `episodes[i]["episode"]` entry handed to `record_deterministic_verdicts`, and a key of `measure_agreement`'s holdout mapping. A value outside it selects a *different* episode rather than failing slowly - `True` is `1` to an index - so it is refused and named |
 | `quality` | `low` / `medium` / `high` (ordered; filters compare rank). An *execution* grade, orthogonal to the outcome - see below |
 | `failure_mode` | `null` or one of `jerky_motion`, `near_miss`, `camera_occlusion`, `wrong_but_lucky`, `drift`, `collision`, `incomplete`, `other` |
 | `success_opinion` | `null` (no opinion) or a boolean |
--- a/strands_robots/episode_labels.py
+++ b/strands_robots/episode_labels.py
@@ -466,7 +466,11 @@ def measure_agreement(root: str | Path, human_labels: dict[int, dict[str, Any]])
         root: Dataset root directory.
         human_labels: Mapping of episode index to the human's labels for the
             holdout, each a dict with ``quality`` (required, one of
-            :data:`QUALITY_GRADES`) and optionally ``failure_mode``.
+            :data:`QUALITY_GRADES`) and optionally ``failure_mode``. Each key
+            is an episode index on the shared non-negative-whole-number
+            domain, checked like every other spelling of that quantity: a key
+            that is not one selects the wrong episode to calibrate against,
+            not a slower calibration.
 
     Returns:
         Dict with ``episodes_compared``, ``quality_agreement`` and
@@ -491,7 +495,7 @@ def measure_agreement(root: str | Path, human_labels: dict[int, dict[str, Any]])
     mode_hits = 0
     disagreements: list[dict[str, Any]] = []
     for index, human in human_labels.items():
-        if msg := non_negative_whole_number_error(index, "human_labels key", "measure_agreement"):
+        if msg := non_negative_whole_number_error(index, "human_labels episode index", "measure_agreement"):
             raise ValueError(msg)
         if not isinstance(human, dict) or human.get("quality") not in QUALITY_GRADES:
             raise ValueError(
--- a/tests/simulation/test_replay_episode_index_domain.py
+++ b/tests/simulation/test_replay_episode_index_domain.py
@@ -41,6 +41,29 @@ The fix applies ``non_negative_whole_number_error`` - the shared rule whose own
 docstring already names ``replay_episode``, the same quantity on the
 neighbouring teleop surface - so the refusal is identical across both spellings
 rather than merely equivalent in verdict.
+
+The structural sweep below is what keeps that true as surfaces are added, so
+its *discovery* has to be as wide as the quantity. It used to admit a surface
+only when a parameter was literally named ``episode``, which is one spelling of
+two: an index also arrives inside a **collection** - an entry of the
+``episodes`` list ``record_deterministic_verdicts`` records, a key of the
+``human_labels`` mapping ``measure_agreement`` calibrates against. Both apply
+the shared rule, and both were invisible to the sweep, so
+``test_the_surface_set_is_exactly_the_pinned_one`` reported a clean tree while
+grading 9 of 11 surfaces and ``test_every_surface_validates_or_forwards`` never
+ran on either. Deleting both guards left the whole file passing (203 passed),
+while the behaviour they hold degraded exactly as the table above:
+``{"episode": True}`` recorded a verdict against **episode 1** and left episode
+0 unlabeled, and ``measure_agreement({True: ...})`` reported
+``quality_agreement 1.0`` calibrated against episode 1 - the number that
+decides whether a judge is trusted to filter training data.
+
+Discovery is therefore derived from each surface's own ``Args:`` entry: a
+parameter named ``episode``, or one whose documented description says it
+carries an episode index. That keeps the rule a property of the tree rather
+than a list to maintain, and it holds the collection spellings to the same
+contract - ``filter_episodes``, which *returns* indices rather than resolving
+one, stays out on the same rule.
 """
 
 from __future__ import annotations
@@ -49,12 +72,14 @@ import ast
 import inspect
 import math
 import pathlib
+import re
 import sys
 import types
 
 import numpy as np
 import pytest
 
+from strands_robots import episode_labels
 from strands_robots.utils import non_negative_whole_number_error
 
 # ── the fake dataset ────────────────────────────────────────────────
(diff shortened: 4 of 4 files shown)


yinsong1986 (approved):
## Summary

This PR widens the structural sweep in `tests/simulation/test_replay_episode_index_domain.py` so an episode index arriving inside a collection (an `episodes[i]["episode"]` entry of `record_deterministic_verdicts`, a key of `measure_agreement`'s `human_labels` mapping) is discovered and graded on the same shared `non_negative_whole_number_error` contract as a bare `episode` parameter. Discovery is now derived from each surface's own `Args:` entry rather than a literal parameter name, the two surfaces join `_REPLAY_EPISODE_SURFACES`, and `measure_agreement`'s refusal is renamed from `human_labels key` to `human_labels episode index` so the recogniser can see the guard that was already there. Production behaviour is otherwise unchanged - both guards pre-existed; only the refusal wording and docstrings move.

## What's good

- The load-bearing claim reproduces: at head, the changed test file passes 222/222, and disabling the `measure_agreement` guard fails 7 tests naming that surface (both structurally via the widened sweep and behaviourally via `TestTheCollectionSpellingsRefuseAnUnusableIndex`), where the pre-PR sweep was blind to it.
- No other test pinned the old `human_labels key` wording, so the message rename breaks nothing.
- The boundary is principled rather than an exemption list: `filter_episodes` stays out because no parameter of it is documented as carrying an index, and `test_discovery_leaves_a_surface_that_only_returns_indices_out` pins that.
- Changelog fragment present (`changelog.d/2527-episode-index-collection-spelling.md`), docs table updated, `ruff check` / `ruff format --check` clean on the changed files, tests use `tmp_path` throughout with no shared-cache or host-path exposure.

cagataycali:
`call-test-lint` here is red at step 7 `Install dependencies`, with `Run lint` and `Run tests`
**skipped** - so it reports nothing about this diff, which touches no packaging file. The same
job fails identically on `main` itself (`a7170f26`, #2526), so it is not specific to this PR.

Root cause is an upstream publish, not a change in this repo: **`mujoco 3.12.0` was uploaded to
PyPI at 2026-08-20T11:01:09Z as an sdist with no wheels for any interpreter** (3.9.0 / 3.10.0 /
3.11.0 all ship cp310-cp314). `mujoco>=3.5.0,<4.0.0` therefore resolves 3.12.0, pip falls back to
a source build, and that needs a MuJoCo tree:

```
Downloading mujoco-3.12.0.tar.gz (1.2 MB)
Building wheel for mujoco (pyproject.toml): finished with status 'error'
    RuntimeError: MUJOCO_PATH environment variable is not set
error: failed-wheel-build-for-install
```

The timeline matches exactly: `d08fcea4` (10:26Z) green, `a7170f26` (11:07Z) red, this PR (11:11Z)
red, and PR heads from 07:56Z / 09:55Z still green. Two attempts here failed identically.

Wheels for a fresh MuJoCo release usually follow the sdist within hours, so this may clear on a
re-run without any change; whether to add a temporary ceiling in the meantime is a call for the
maintainers rather than something to slip into this PR.

Local verification for this diff stands as described above (58-file blast radius, 2604 passed / 0
failed on the branch against 2602 passed / 2 failed on `main` with the new guard file copied in;
ruff and mypy clean).

yinsong1986 (approved):
## Summary

This PR closes a discovery gap in the structural sweep of `tests/simulation/test_replay_episode_index_domain.py`: the sweep previously admitted a surface only when a parameter was literally named `episode`, so the two collection-valued spellings of the same quantity - an `episodes[i]["episode"]` entry handed to `record_deterministic_verdicts` and a key of `measure_agreement`'s `human_labels` mapping - were graded by nothing, and deleting their guards left the file green. Discovery is now derived from each surface's own `Args:` entry, both surfaces join `_REPLAY_EPISODE_SURFACES` (9 -> 11 graded), and `measure_agreement`'s refusal names the episode index rather than the container slot, which is also what makes the sweep's recogniser able to see the guard. The production diff is deliberately minimal - a docstring extension and a refusal-message rename, no signature, schema, or wire-format change - with the bulk of the change in the test file, a docs field-domain row, and a `changelog.d/` fragment. Verified at the head SHA: the full sweep file passes (222 passed), `ruff check` / `ruff format --check` are clean on the changed files, and no non-ASCII characters were introduced.

## What's good

- The widened discovery is load-bearing both ways: a planted unguarded collection surface fails it, and `filter_episodes` stays out on the same derived rule rather than an exemption list.
- Behavioural tests drive the real functions over the full unusable-value set (`True` is the sharp case - `int(True)` is 1 and every episode table is indexable by 1), assert a refused verdict writes no sidecar, and keep an in-domain control.
- Changelog fragment in `changelog.d/` per AGENTS.md step 3; docs table row states the domain in each spelling the index arrives in.

cagataycali:
The failing check here is not this diff. `call-test-lint` reports `Run lint: success` and `Run tests: failure` at 16m51s, and the one failure is:

```
FAILED tests/simulation/test_add_robot_seats_on_terrain.py::test_floating_base_spawns_seated_on_terrain[named_base] - assert 0.0 >= (0.16 - 1e-06)
```

`main` @ `a7170f26` fails on the same test with the same assertion, so it is inherited. Root cause is `mujoco` 3.12.0, published at 11:01:09Z, which CI began resolving inside the declared `>=3.5.0,<4.0.0` range; #2528 is the fix and is approved. Verified locally at that PR's head: 3 failed on 3.12.0 for `main`, 6 passed for `d6cc7e98`, both 6 passed on 3.10.0.

This branch touches `strands_robots/episode_labels.py`, one replay-index test and docs, none of which reach the seating path. Once #2528 lands a re-run here should be green.

yinsong1986 (approved):
## Summary

This PR closes a discovery gap in the episode-index structural sweep: `tests/simulation/test_replay_episode_index_domain.py` previously admitted a surface only when a parameter was literally named `episode`, so the two collection-valued spellings of the quantity - `episodes[i]["episode"]` in `record_deterministic_verdicts` and the keys of `measure_agreement`'s `human_labels` mapping - were guarded in the tree but graded by nothing. Discovery is now derived from each surface's own `Args:` entry, the two surfaces join the pinned set (9 -> 11), `measure_agreement`'s refusal names the episode index instead of the container slot (which is also what lets the recogniser see the guard), and the docs field-domain table gains an `episode_index` row. The production diff is deliberately minimal: a docstring addition and a refusal-message rename; both guards already existed and their behaviour is unchanged.

## What's good

- The load-bearing claim reproduces at head: the full file passes 222, and deleting both collection guards fails the sweep 16 ways, naming both surfaces structurally and behaviourally - exactly the table in the description.
- The completeness test (`test_the_surface_set_is_exactly_the_pinned_one`) still pins discovered == registered over the whole `strands_robots/` tree, so the wider docstring-derived recogniser introduces no false positives, and the `filter_episodes` boundary is asserted on the same derived rule rather than as an exemption.
- Behavioural tests drive the real functions over the shared domain's failure values and assert a refused verdict writes no sidecar - refusal lands before `_write_document`, so there is no partial write to undo.
- AGENTS.md compliance: changelog fragment (step 3), docs updated in the same PR, `tmp_path`-rooted test datasets, plain-ASCII messages, ruff clean on the touched files.

## Verification suggestions

- Spot-check the load-bearing property in one command: delete the two `non_negative_whole_number_error` guards in `strands_robots/episode_labels.py` and run `pytest tests/simulation/test_replay_episode_index_domain.py -q --no-cov` - expect 16 failures naming `record_deterministic_verdicts` and `measure_agreement`, then restore.

PR #2528, opened 2026-08-20
fix(simulation): match a MuJoCo enum by value, not by operand order

cagataycali:
## Problem

`mjModel` / `mjData` expose their type fields as numpy integer arrays
(`model.geom_type[i]` is an `np.int32`), while the matching vocabulary is a
pybind11 enum (`mujoco.mjtGeom.mjGEOM_HFIELD`). Whether the two compare equal
depends on which side the enum is on, and that changed inside the declared
`mujoco>=3.5.0,<4.0.0` range:

| comparison, on a numpy model field | 3.9.0 / 3.10.0 / 3.11.0 | 3.12.0 |
| --- | --- | --- |
| `np.int32(1) == mjGEOM_HFIELD` (numpy left) | True | True |
| `mjGEOM_HFIELD == np.int32(1)` (enum left) | True | **False** |
| `geom_type[i] in (HFIELD, PLANE)` | True | **False** |
| `int(geom_type[i]) in (int(HFIELD), int(PLANE))` | True | True |

The enum-on-the-left form is not usually written by hand: it is what `in`
produces. CPython's `tuple.__contains__` compares `element == needle`, so
`model.geom_type[i] in (mjGEOM_HFIELD, mjGEOM_PLANE)` puts the enum on the left
and the membership test simply answers `False` for every element - nothing
raised, nothing logged. A `set` of enum members degrades identically: the hashes
still collide and the equality that confirms the hit is the failing direction.

Two measured consequences on 3.12.0:

- `tests/simulation/test_add_robot_seats_on_terrain.py` - the heightfield ground
  geom stopped being recognised as ground, so the helper that computes "the
  lowest collidable robot geom" returned the ground's own `z=0.0`. Three
  `create_world(terrain=...)` regression tests failed with `assert 0.0 >= 0.16`
  (**6 passed on 3.10.0, 3 failed on 3.12.0**, same tree).
- `examples/mujoco_gs/scene.py::_erect_arm` and its `examples/so101_curobo`
  twin - the home-pose helper matched **0 of 4** hinge/slide joints and wrote
  neither `qpos` nor `ctrl`, leaving the arm in its compiled zero pose while
  reporting success and logging that the home pose had been set.

## Change

Every comparison that put a `mjt*` enum on the left now reads `int()` on both
sides - the convention the MuJoCo backend already uses at its own type checks
(`int(model.actuator_trntype[act_id]) != int(mj.mjtTrn.mjTRN_TENDON)`, six such
sites today). Seven call sites across `strands_robots`, `tests`, `tests_integ`
and `examples`.

The rule is applied uniformly rather than with an exemption. The one shipped
site, `spec_builder.attach_robot`'s ground detection, was measured **not**
broken - `MjsGeom.type` really is an enum, so `in` works there - but a reader
cannot tell an `MjsGeom` attribute from an `MjModel` array element at the call
site, so it is written by value too. Comparisons with the enum on the *right*
(`geom_type[i] == mjGEOM_PLANE`) still match on every build measured and are
left alone.

`tests/test_mujoco_enum_comparisons_are_value_based.py` closes the class: it
grades all four trees for a `mjt*` enum on the left of `==`, whether that comes
from a literal collection, a named collection of enum members, or a
hand-written reversed comparison. It is structural rather than behavioural on
purpose - the defect only manifests from 3.12.0 on, so a test that merely
exercised the installed build would pass on an older one and pin nothing.

## Verification

Measured at `d6cc7e9`, mujoco 3.12.0 unless stated.

Pre-fix, the new guard against `upstream/main` sources: **1 failed / 13
passed**, naming all seven offenders. Pre-fix, the terrain-seat tests on
`upstream/main`: **6 passed on mujoco 3.10.0**, **3 failed on 3.12.0**.

Post-fix the new guard and the terrain-seat tests are **20 passed on mujoco
3.10.0 and 20 passed on 3.12.0** - the fix is a no-op on the builds where the
old form already worked.

Blast radius (all of `tests/simulation`, every test that scans a source tree or
`AGENTS.md`, and the repo-wide guards - 179 paths, 17275 collected), branch vs
an `upstream/main` worktree, same interpreter:

| measure | branch | upstream/main |
| --- | --- | --- |
| passed | 16956 | 16951 |
| failed | 35 | 40 |
| errors | 126 | 126 |

Branch-only failing ids: none. The two ids that appear branch-only are in
`tests/mesh/test_zenoh_transport_security.py`, which opens real localhost Zenoh
sessions; base fails three *different* ids in that same file, and the file is
**9 passed on both trees** when run alone. Base-only ids are exactly the three
terrain-seat tests and the new guard. The 126 errors and the remaining shared
failures are the same on both trees (optional deps absent from this
interpreter).

tests_integ`; `mypy` 24 errors on the branch and 24 on a pristine

## Artifact

mujoco 3.12.0 in both columns, same scene, each column executing that tree's
own `_erect_arm` verbatim. Left: 0 of 4 joints matched, no `qpos` and no `ctrl`
written, arm still in its zero pose. Right: 4 of 4 matched, arm in its `home`
pose.

enum matched by value

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1350,6 +1350,32 @@ Corrections from code review that apply to all future contributions:
   `ctrlrange` is authoritative and already in the drive's own units, so only the
   *substitution* of a driven joint's limits needs the drive to be a servo.
 
+### MuJoCo enums are matched by value, never by operand order
+
+`mjModel` / `mjData` expose their type fields as numpy integer arrays, while the
+matching vocabulary is a pybind11 enum. Whether the two compare equal depends on
+which side the enum is on.
+
+- **Compare `int()` to `int()`.** `int(model.geom_type[i]) in (int(mjGEOM_PLANE),
+  int(mjGEOM_HFIELD))`, `int(model.actuator_trntype[a]) != int(mjTRN_TENDON)`.
+- **`x in (enum, ...)` is the trap, not a hand-written reversed `==`.** CPython
+  compares `element == needle`, so membership puts the ENUM on the left. That is
+  `True` on mujoco 3.9.0 / 3.10.0 / 3.11.0 and `False` on 3.12.0 - all inside the
+  declared `mujoco>=3.5.0,<4.0.0` range. A `set` of enum members degrades the
+  same way: the hashes still collide and the confirming equality is the failing
+  direction.
+- **The failure is silent.** The membership test simply answers `False` for every
+  element. Measured: a heightfield ground geom stopped being recognised as
+  ground, so a "lowest robot geom" scan returned the ground's own `z=0.0`; and an
+  example's home-pose helper matched 0 of 3 joints, writing no `qpos` and no
+  `ctrl` while logging that the pose had been set.
+- **The rule is uniform, with no exemption for a spec-side value.** `MjsGeom.type`
+  really is an enum, so `in` works there - but a reader cannot tell an `MjsGeom`
+  attribute from an `MjModel` array element at the call site, so both are written
+  by value.
+- Pinned by `tests/test_mujoco_enum_comparisons_are_value_based.py`, which grades
+  `strands_robots`, `tests`, `tests_integ` and `examples`.
+
 ### Clocks: a duration is measured, a stamp is recorded
 - **A duration belongs on `time.monotonic()`.** Anything that decides *how long to keep
   waiting* - a timeout, a deadline, a TTL, a retry window, a rate window, a frame pacer -
--- a/changelog.d/2528-mujoco-enum-value-comparison.md
+++ b/changelog.d/2528-mujoco-enum-value-comparison.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **simulation**: match a MuJoCo `mjt*` enum by value rather than by operand order. A model field is a numpy integer and `x in (enum, ...)` puts the enum on the *left* of `==`, where a pybind11 enum stops matching a numpy integer on mujoco 3.12.0 - inside the declared `mujoco>=3.5.0,<4.0.0` range. The membership test then answers `False` for every element with nothing raised: a heightfield ground geom was no longer recognised as ground, and an example's home-pose helper matched 0 of 3 joints and wrote neither `qpos` nor `ctrl` while logging that the pose had been set. Every such comparison now reads `int(...)` on both sides, and a new guard grades `strands_robots`, `tests`, `tests_integ` and `examples` so a new site fails CI instead of degrading silently.
--- a/examples/mujoco_gs/scene.py
+++ b/examples/mujoco_gs/scene.py
@@ -210,17 +210,20 @@ def _erect_arm(sim, robot_name: str = "arm") -> bool:
 
         kq = m.key_qpos[key_id]
         ns = f"{robot_name}/"
-        hinge_slide = (mujoco.mjtJoint.mjJNT_HINGE, mujoco.mjtJoint.mjJNT_SLIDE)
+        # Joint types are compared BY VALUE: ``m.jnt_type[j]`` is a numpy integer and
+        # ``x in (enum, ...)`` puts the enum on the left of ``==``, where a pybind11
+        # enum does not match a numpy integer on every mujoco build.
+        hinge_slide = (int(mujoco.mjtJoint.mjJNT_HINGE), int(mujoco.mjtJoint.mjJNT_SLIDE))
         # 1) qpos for the arm's own actuated joints.
         for j in range(m.njnt):
             jn = mujoco.mj_id2name(m, mujoco.mjtObj.mjOBJ_JOINT, j) or ""
-            if jn.startswith(ns) and m.jnt_type[j] in hinge_slide:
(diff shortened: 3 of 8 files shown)


yinsong1986 (approved):
## Summary

Fixes a silent, version-dependent MuJoCo enum-comparison defect: `model.geom_type[i]` / `jnt_type[j]` are numpy integers, and `x in (mjt_enum, ...)` puts the pybind11 enum on the left of `==`, where it stops matching a numpy integer on mujoco 3.12.0 (inside the declared `>=3.5.0,<4.0.0` range). The membership test then answers `False` for every element with nothing raised - measured as a heightfield ground geom no longer recognised as ground and an example home-pose helper writing neither `qpos` nor `ctrl` while logging success. The fix converts all seven enum-on-the-left sites to `int()`-on-both-sides (the convention the backend already uses), applied uniformly including the one `MjsGeom` site that was measured unbroken. A new structural AST guard (`tests/test_mujoco_enum_comparisons_are_value_based.py`) grades all four trees for the reversed shapes (literal collections, named collections, hand-written reversed `==`/`!=`), pins its own reach (>=40 files, all four trees), and includes planted-source positive/negative cases plus a behavioural check on the installed build. Verified locally at `d6cc7e98`: the guard is 14/14 and the three terrain-seat regression tests pass on mujoco 3.8.0, confirming the fix is a no-op on builds where the old form worked. Changelog fragment present (`changelog.d/2528-mujoco-enum-value-comparison.md`), AGENTS.md section added and pinned by the guard, no public API / wire / schema changes, no host paths or non-ASCII in the new files.

## What's good

- The guard is structural by design and says why: a behavioural test would pass on any pre-3.12.0 build and pin nothing.
- The scan grades its own reach (`test_the_scan_reaches_every_tree`), so a refactor that stops finding sources fails instead of reporting clean - the same discipline AGENTS.md asks of tree-scanning tests.
- The uniform-rule call (converting the unbroken `MjsGeom` site too) is argued in the description rather than left as an unexplained exemption, and the guard enforces it.
- Measured verification in both directions: pre-fix failure counts on both mujoco versions, post-fix parity, and a branch-vs-base blast-radius table with the branch-only deltas accounted for.

cagataycali:
Independently reproduced this on a separate machine at the current head `d6cc7e98`, and it is worth flagging that `main` has gone red on exactly this failure since the PR was opened.

## `main` is currently red on this test

| time (UTC) | event |
| --- | --- |
| 11:01:09 | `mujoco` **3.12.0** published to PyPI (24 wheels) |
| 10:26:21 | `main` @ `d08fcea4` -- `Pull Request and Push Action` **success** |
| 11:07:08 | `main` @ `a7170f26` -- **failure**, six minutes after the upload |

`main`'s failing job reports one line:

```
FAILED tests/simulation/test_add_robot_seats_on_terrain.py::test_floating_base_spawns_seated_on_terrain[named_base] - assert 0.0 >= (0.16 - 1e-06)
```

Nothing in `a7170f26` touches the seating path; the only variable is that CI now resolves 3.12.0 (`Downloading mujoco-3.12.0-cp312-cp312-manylinux_...whl`) where the previous run resolved an older build inside the same `mujoco>=3.5.0,<4.0.0` range.

## Version boundary, with numpy held fixed

The one thing the table in the description does not state is that numpy is not the variable. Installing each mujoco with `--no-deps` so numpy stayed at 2.2.6 throughout:

| mujoco | `np.int32(v) in (mjGEOM_PLANE,)` | `mjGEOM_PLANE == np.int32(v)` | `int(x) in (int(enum),)` |
| --- | --- | --- | --- |
| 3.10.0 | True | True | True |
| 3.11.0 | True | True | True |
| **3.12.0** | **False** | **False** | True |

3.12.0 is the exact boundary, and the asymmetry is the whole defect: on 3.12.0 `np.int32(v) == mjGEOM_PLANE` is still `True` while `mjGEOM_PLANE == np.int32(v)` is `False`, which is the direction `tuple.__contains__` uses. Same result for `np.int64`.

## Fix verification

Same worktree pair, same command, only `PYTHONPATH` selecting the mujoco build:

| tree | mujoco 3.10.0 | mujoco 3.12.0 |
| --- | --- | --- |
| `upstream/main` | 6 passed | **3 failed, 3 passed** |
| `d6cc7e98` | 6 passed | **6 passed** |

CI surfaces only one of the three because it runs `-x`; the other two are `[unnamed_base]` and `test_terrain_seat_survives_reset`, both `assert 0.0 >= (0.16 - 1e-06)`. The 3.10.0 column is the no-op control.

## The `spec_builder` caveat holds

Confirming the description's own claim rather than taking it on trust, since it is the only shipped site: on 3.12.0 the spec-level predicate answers `True` under both spellings for a scene whose only geom is a heightfield (`types seen: [('terr', 1)]`), because `MjsGeom.type` really is a plain integer rather than a numpy array element. So that site is consistency, not a live fix, exactly as written. The live failure is `model.geom_type[gid]`, which is the numpy side.

One practical note: any open PR whose CI re-runs after 11:01Z inherits this. #2527 is already red on this same test with an unrelated diff.

PR #2529, opened 2026-08-20
fix(simulation): refuse a model whose actuators outnumber their control slots

cagataycali:
## Problem

`add_robot(urdf_path=...)` with a caller MJCF that mujoco 3.12 compiles fails with a numpy index error and leaves the robot's bodies in the scene:

```
add_robot -> error | Failed to load: index 2 is out of bounds for axis 0 with size 2
list_robots()      -> []
live model          nbody 1 -> 3, nu 0 -> 3, bodies ['world', 'arm/link1', 'arm/link2']
```

Every actuator surface in the MuJoCo backend addresses an actuator **by its control index**: `SimRobot.actuator_ids` holds indices into `data.ctrl`, and the `actuator_*` model arrays are read at that same index. Through mujoco 3.11 that identity was free -- every actuator owned exactly one control slot, so `model.nu` was both the control width and the actuator count.

mujoco 3.12 separated them. `<pid>` takes two controls by default (`input="pos vel"`) and `<orientation>` up to four. Measured on 3.12.0 for a two-joint arm:

| actuators | `nu` | `nactuator` | `actuator_ctrladr` | `actuator_ctrlnum` |
|---|---|---|---|---|
| `position`, `position` | 2 | 2 | `[0, 1]` | `[1, 1]` |
| `position`, `pid` | 3 | 2 | `[0, 1]` | `[1, 2]` |
| `pid`, `position` | 3 | 2 | `[0, 2]` | `[2, 1]` |

`robot_owned_actuator_ids` loops `range(model.nu)` and hands each index to `actuator_joint_id`, which reads `actuator_trntype[act_id]` -- an actuator-indexed array now shorter than `nu`. The raise escapes `_recompile_preserving_state`'s refusal path (it catches only around `spec.recompile`) and `inject_robot_into_scene`'s spec restore (it catches only around `attach_robot`), so the model is already installed: an orphan subtree no registry entry owns is left in the live scene. A corrected model under the same name then collides -- `Robot attach failed for 'arm': Error: repeated name 'arm/a1' in actuator` -- reported as the generic `Failed to inject robot 'arm' into scene.`

**Iterating `range(nactuator)` instead is not the fix.** On the `pid, position` row the position servo's setpoint is `ctrl[2]` while such a loop writes `ctrl[1]` -- the pid's *velocity* setpoint -- so a pose would be applied as a rate and the servo never commanded at all. Choosing which of an actuator's controls a pose is written to, and what `robot_action_keys` reports for the rest, is a control layout this surface does not have.

No shipped registry robot declares a multi-control actuator, so this is reachable through a caller MJCF (`add_robot(urdf_path=)`, `replace_scene_mjcf`, `patch_scene_mjcf`, `load_scene`).

## Change

A model this backend cannot address is now refused **where the compiler's own refusal is raised** -- after `spec.recompile` succeeds and before `install_compiled_model` -- so the world keeps the model it had and a caller's rollback finds the scene exactly as it was. The reason names each wide actuator and the control slots it owns, plus a remedy:

```
Failed to load: model declares 2 actuator(s) spanning 3 control slots ('arm/a2' owns control
slots [1, 2]). This backend addresses an actuator by its control index, so a multi-control
actuator (mujoco >= 3.12 <pid>, <orientation>) has no single index to drive it by. Replace it
with a single-control actuator (<position>, <motor>, <velocity>, <intvelocity>, <general>), or
drive that joint outside this backend.
```

`inject_robot_into_scene` now carries that reason out the way `inject_object_into_scene` already does, instead of folding it into a bare `False`. That injector's own comment records why: folded into a `False` a refusal "became 'spec recompile refused.' with the actionable text left in the log - the same dead end the unsupported-shape path was fixed to stop producing". The robot injector was the sibling that never adopted it, which is why the reason above reaches the caller at all.

The check is capability-probed on `nactuator`, which mujoco 3.5 (the declared floor) does not expose -- there `nu` **is** the actuator count, so the reason is always `None` and the path is byte-identical.

## Artifact

Headless EGL on mujoco 3.12.0. Same script against each tree; the renders are the live scene **after** each call. Row 2 is a caller who corrects the MJCF and retries under the same robot name.

multi-control actuator refusal

|---|---|---|
| refused add | `index 2 is out of bounds for axis 0 with size 2` | names `'arm/a2'`, slots `[1, 2]`, and a remedy |
| scene after the refusal | `nbody=3 nu=3`, orphans `['arm/link1', 'arm/link2']`, `list_robots()==[]` | `nbody=1 nu=0`, no orphans -- untouched |
| retry, corrected MJCF, same name | `error` (`repeated name 'arm/a1'`) | `success`, `robot_action_keys('arm') == ['a1', 'a2']` |
| remedy offered | none | `<position>, <motor>, <velocity>, <intvelocity>, <general>` |

Both trees start from the identical scene (`nbody=1 nu=0 robots=[]`), and `upstream/main`'s two renders are pixel-for-pixel the same subject count (25669 px) -- the retry added nothing.

## Tests

`tests/simulation/mujoco/test_multi_control_actuator_is_refused.py`, in two halves:

* The reason is graded against **stub models**, so the rule is pinned on every supported build: one control per actuator and a build with no `nactuator` are addressable; a trailing wide actuator, a leading one, an unnamed one, and a build that reports the counts but no slot layout are each refused with what it can name.
* The end-to-end half drives `add_robot` and is gated on a build that compiles a multi-control actuator, covering both `pid`-first and `pid`-second: the refusal names the actuator rather than an index, the scene keeps its sizes and gains no `arm/` body, later mutations still compile, and the remedy is parsed out of the message and applied -- reaching `success`.

Pre-fix on `upstream/main` (mujoco 3.12.0): **13 failed / 1 passed**. Six are behavioural -- `Failed to load: index 2 is out of bounds for axis 0 with size 2`, `assert (3, 3, 2) == (1, 0, 0)` for the orphan subtree, and `assert 'error' == 'success'` for the bricked retry; the other seven pin the root cause by naming the new helper. The one pass is the control: a single-control robot still adds.

On mujoco 3.5.0 the file is 8 passed / 6 skipped -- the stub half and the single-control control still run.

## Verification

Rebased onto `63e0a034` by merging `main` in, which picks up #2528 -- the terrain-seat
failure this branch's base predated is gone from both trees, so the earlier red `Run tests`
is resolved without touching this diff.

Measured at `3f6906e9`, both trees on mujoco 3.12.0 (what CI resolves under `mujoco>=3.5.0,<4.0.0`).

* Blast radius, 46 paths (all of `tests/simulation`, every test naming the injectors or the actuator-slot arrays, every docs-reading test, and the repo-wide guards): branch **2 failed / 12596 passed / 226 skipped**, base **15 / 12583 / 226**. Both collected 12598; `15 - 2 == 12596 - 12583 == 13`, exactly the pre-fix failures. Branch-only failures: **none**, and all 13 base-only ids come from this PR's own test file -- no foreign test changes verdict.
* Pre-fix split re-measured against the moved base: **13 failed / 1 passed**, unchanged.
* `tests/simulation/test_add_robot_seats_on_terrain.py` now passes **6/6 on both trees**.
* The 2 remaining shared failures are pre-existing on `upstream/main` and are local dependency artifacts rather than CI ones: `test_declared_qp_backends_are_real_qpsolvers_extras` needs `qpsolvers`, which this venv lacks and CI installs via `.[all,dev]`, and `test_a_numpy_array_pose_is_accepted_by_both` is a numpy-version artifact.
* `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 against a pristine `upstream/main` worktree, none in the changed files.
* `scripts/assemble_changelog.py --check` OK.
* `tests/test_mujoco_enum_comparisons_are_value_based.py`, the AST guard #2528 added, passes against this diff: the new helper reads `mjOBJ_ACTUATOR` as an argument to `mj_id2name`, never as an operand of a comparison.


yinsong1986 (approved):
## Summary

Guards the MuJoCo backend against models whose actuators own more than one control slot (mujoco >= 3.12 `<pid>` / `<orientation>`), where the backend's actuator-by-control-index addressing breaks in both directions. The new `_unaddressable_actuator_reason` probe runs after `spec.recompile` and before `install_compiled_model`, so a refused model leaves the world's model/data and spec untouched, and `inject_robot_into_scene` adopts the same raise-with-rollback shape `inject_object_into_scene` already uses so the actionable reason (offending actuator, its slots, a remedy) reaches the caller instead of a bare `False`. I traced the new raise path: the injector's sole caller is `add_robot`, whose `except Exception` pops the registry entry and folds the reason into a structured error dict, so the #85 "return error dicts, never raise" and "clean up on failure" contracts both hold. Callers on the default `raise_on_refusal=False` get a `False` they already handle. Deliberately refusing rather than guessing a control layout (the `pid, position` row where a count-based loop writes a pose into a velocity setpoint) is the right call for a surface with no multi-control model.

## What's good

- The capability probe reproduces as claimed: on a build with no `nactuator` (verified locally on 3.8.0, where the field and the `actuator_ctrladr`/`actuator_ctrlnum` arrays are all absent) the reason is always `None` and the path is byte-identical; the stub half of the test file pins the 3.12 behaviour without needing that build.
- New test file runs green here (8 passed / 6 skipped on a pre-3.12 build), and the adjacent `test_add_robot_failure_cleanup.py` / `test_add_robot_preserves_scene_state.py` suites still pass against the branch, so the injector contract change did not disturb the existing failure-path pins.
- Hygiene is clean across the diff: ASCII-only, no host paths, changelog fragment named per convention, ruff check/format clean on the changed files, and the `:meth:` cross-reference target (`MuJoCoSimEngine.robot_action_keys`) exists and sits on one line.
- The refusal message names the element (`<pid>`), the actuator, its slots, and a remedy, and the end-to-end test applies the remedy verbatim and reaches `success` under the same robot name -- the orphan-subtree and name-collision halves of the bug are both pinned.

## Verification suggestions

- On a mujoco >= 3.12 environment, run the end-to-end half of `tests/simulation/mujoco/test_multi_control_actuator_is_refused.py` (the `_needs_multi_control`-gated tests) to confirm the six behavioural pins fail pre-fix and pass post-fix as the description reports; CI's resolved mujoco should cover this already.
- `replace_scene_mjcf` / `patch_scene_mjcf` install a compiled model without passing through `_recompile_preserving_state`, so a wide-actuator scene loaded whole is refused only at the next mutation that recompiles (e.g. `add_robot`, `actuate_robot`) rather than at load. That is consistent with the fix's seam and nothing crashes in between (no robots are registered, so no control-index read happens), but a follow-up issue for refusing at those install sites too would close the class at every door.

yinsong1986 (approved):
## Summary

Refuses a recompiled model whose actuators span more than one control slot (mujoco >= 3.12 `<pid>` / `<orientation>`), placing the check on the same rung as the compiler's own refusal -- after `spec.recompile`, before `install_compiled_model` -- so the previously-crashing `add_robot` path can no longer install an orphan subtree the registry has already dropped, and `inject_robot_into_scene` now carries the refusal's reason out via `raise_on_refusal=True` the way `inject_object_into_scene` already does. I independently traced the new raise: the injector's only caller is `add_robot`, whose `except Exception` pops the registry entry and folds the reason into a structured error dict, and the injector restores the pre-attach spec before re-raising, so a refused add leaves the scene exactly as found. The new test file passes on a pre-3.12 build (8 passed / 6 skipped, matching the description), the adjacent add_robot failure-path suites are undisturbed by the injector contract change, and ruff check/format are clean on the changed files. Refusing rather than guessing a control layout is the right call for the `pid, position` ordering, where a count-based loop would silently write a pose into a velocity setpoint. No public API, wire-format, or persisted-schema change; changelog fragment present. No blockers found.

Issue #2507, opened 2026-08-19
process: two PRs appending to the same AGENTS.md learnings list conflict on ordering alone, and it costs an approval - but the collision is per-list-anchor, not per-file


cagataycali:
## What happened

#2475 and #2485 both appended a bullet to the **same** AGENTS.md list - `Corrections from code review that apply to all future contributions:` - at adjacent lines. #2475 landed first, which put #2485 in conflict on a file it shares with no behavioural change.

Resolving it cost an approval. The conflict resolution was 15 lines (`git show --cc`), entirely the reinsertion of #2475's bullet, and the diff under review was byte-identical before and after (`4 files, +293/-1`). #2485 went `APPROVED` -> `REVIEW_REQUIRED` for a push that changed nothing a reviewer had approved.

This is the mechanism AGENTS.md already documents for `CHANGELOG.md` in PR Workflow step 3 - "every branch inserts at the same anchor, so two PRs open at once conflict on ordering alone, and because stale approvals are dismissed on push each resolution costs a re-approval round that reviews no changed behaviour." That was solved for the changelog by `changelog.d/` news fragments (#1692). The same sentence is now true of AGENTS.md, and nothing addresses it there.

## The measurement that narrows it - please read before proposing a fix

My first hypothesis was that this is a *file*-level problem, because **4 of the 15 currently open PRs touch AGENTS.md** (#2485, #2500, #2501, #2504) and 13 of the last 100 commits on `main` do. That hypothesis is **false**, and the negative result is the useful part of this report.

Those four insert at four different section anchors:

| PR | inserts at | section |
|---|---|---|
| #2485 | 1432 | `Corrections from code review ...` |
| #2504 | 1724 | mesh `{core,audit,security}.py` learnings |
| #2501 | 1737 | mesh `{core,audit,security}.py` learnings |
| #2500 | 1765 | `robot_mesh` HITL review trail |

Simulating every landing order - land one, then re-test the other three against the new `main` with `git merge-tree` - gives **12 of 12 clean**:

```
--- after landing #2501 ---   #2504 CLEAN  #2500 CLEAN  #2485 CLEAN
--- after landing #2504 ---   #2501 CLEAN  #2500 CLEAN  #2485 CLEAN
--- after landing #2500 ---   #2501 CLEAN  #2504 CLEAN  #2485 CLEAN
--- after landing #2485 ---   #2501 CLEAN  #2504 CLEAN  #2500 CLEAN
```

[...]

PR #2532, opened 2026-08-20
fix(assets): refuse an empty robot-name selection instead of downloading every robot

### LINKED ISSUE #2531: download_robots(names=[]) downloads every robot in the registry, and the download_assets tool reaches that value from a non-empty argument
## What happens

`names` on `download_robots` selects a SUBSET of the sim robots the registry lists. The docstring documents `None` = all sim robots, so an explicitly empty selection is the *opposite* of that. It is read by truthiness:

```python
if names:
    robots = {...}          # the named subset
elif category:
    robots = {...}          # the whole category
else:
    robots = dict(all_sim)  # every sim robot
```

So `names=[]` falls through to the branches that never read it.

## Measured

Against the shipped registry (63 sim robots), with the download calls stubbed and counted:

| call | robots downloaded |
|---|---|
| `names=None` (documented "all") | 56 (+7 failed) |
| `names=[]` (asks for nothing) | **56 (+7 failed)** |
| `names=[]`, `category="humanoid"` | **13 (+4 failed)** |
| `names=["g1"]` (a real subset) | 1 |

The empty selection is byte-identical to the record-all branch, and with a `category` it is silently reinterpreted as that whole category. Either way the result dict reports the count as the caller's own request.

## Why this is reachable without anyone writing `[]`

The `download_assets` agent tool builds the list from a comma-separated string with its own blank filter:

```python
robot_names = [r.strip() for r in robots.split(",") if r.strip()] if robots else None
```

`robots` is non-empty and truthy, but every field is dropped, so the parse yields `[]`:

| `robots=` | parses to |
|---|---|
| `","` | `[]` |
| `" "` | `[]` |
| `",,,"` | `[]` |
| `"  ,  "` | `[]` |

So an agent emitting `robots=","` triggers a clone of the entire registry and is told `status: success`. This is the same shape AGENTS.md already records for `teleoperate(names=[])` -- "an empty selection is what a filter that matched nothing produces" -- with a network side effect instead of a hardware one.

The cost is not theoretical: a regression test exercising `names=[]` against the unfixed function takes **191s** (real clones) where the fixed one takes 0.48s.

## Suggested repair

This needs no contract decision -- AGENTS.md's "A subset selector is read by membership, never by truthiness" already rules on it, and `teleoperate` is the pinned precedent: read `names is None` for the all-branch and refuse the empty selection rather than widening it, ahead of `get_user_assets_dir()` (which performs the cache `mkdir`) so a refused selection leaves nothing behind.

One deliberate narrowing: the shape should NOT be routed through the shared `name_list_error` domain. This surface resolves each name by membership into a dict, so a repeat resolves to its first occurrence, and a mapping and a one-shot iterator are each read exactly once -- all three are honored as written today (measured). That is the same carve-out AGENTS.md records for the WBC and MotionBricks providers. Only the emptiness verdict belongs here.

The tool should refuse in its own vocabulary rather than lean on the library refusal, whose remedy names `names=` -- an argument that surface's caller never passed.

## Out of scope (separate, lower severity)

`names="g1"` as a bare string is iterable per character, so it resolves as the two robots `"g"` and `"1"` and reports `No matching robots found.` -- a mistyped parameter reported as an empty registry match. It does not widen anything, so it is a distinct defect from this one.

cagataycali:
Fixes #2531

## The defect

`names` on `download_robots` selects a SUBSET of the sim robots the registry lists, and the docstring documents `None` as "all sim robots". It was read by truthiness:

```python
if names:                   # [] is falsy -> falls through
    robots = {...}          # the named subset
elif category:
    robots = {...}          # the whole category
else:
    robots = dict(all_sim)  # every sim robot
```

An empty selection asks for *no* robot, which is the opposite of "all of them", not a spelling of it.

## Measured, against the shipped registry (63 sim robots)

| call | robots downloaded |
|---|---|
| `names=None` (documented "all") | 56 (+7 failed) |
| `names=[]` (asks for nothing) | **56 (+7 failed)** |
| `names=[]`, `category="humanoid"` | **13 (+4 failed)** |
| `names=["g1"]` (a real subset) | 1 |

The empty selection is indistinguishable from the record-all branch, and with a `category` it is silently reinterpreted as that entire category. Both report the count as the caller's own request.

## Nobody has to write `[]` to reach it

The `download_assets` agent tool builds the list from a comma-separated string through its own blank filter, so a **non-empty** argument parses to zero names:

| `robots=` | parses to |
|---|---|
| `","` | `[]` |
| `" "` | `[]` |
| `",,,"` | `[]` |
| `"  ,  "` | `[]` |

So an agent emitting `robots=","` clones the entire registry and is told `status: success`. Same shape AGENTS.md already records for `teleoperate(names=[])` -- "what a filter that matched nothing produces" -- with a network side effect rather than a hardware one.

The cost is measurable in the test run itself: the new tests take **191s against the unfixed source** (real clones) and **0.48s** against the fix.

## The change

No new contract. AGENTS.md's "A subset selector is read by membership, never by truthiness" already rules on this, and `teleoperate` is the pinned precedent this mirrors.

1. **`download_robots`** reads `names is None` for the all-branch and refuses the empty selection, raising `ValueError` (already this module's convention -- see the clone-URL guard). Refused **ahead of `get_user_assets_dir()`**, which performs the cache `mkdir`, so a refused selection leaves nothing behind to undo.
2. **`download_assets`** refuses in its own vocabulary, naming `robots=` rather than the `names=` its caller never passed -- a refusal whose remedy names an argument the caller did not use is the failure mode AGENTS.md records under the posture-flag rule. An absent *or empty* `robots=` still means "all": for a single string argument unset and empty genuinely coincide, and only a value that carries content while naming nothing is a mistake.

### What is deliberately *not* changed

The shape is **not** routed through the shared `name_list_error` domain, unlike `teleoperate`. This surface resolves each name by membership into a dict, so -- measured -- a repeat resolves to its first occurrence, and a mapping and a one-shot iterator are each read exactly once. All three are honored as written today, so routing them through the shared domain would reject working calls. That is the same carve-out AGENTS.md records for the WBC and MotionBricks providers ("resolving these names by membership rather than by position ... so they are not callers of this function"). Only the emptiness verdict is taken locally, and the three tolerated spellings are **pinned by tests** so the narrowing stays deliberate rather than incidental.

A bare string (`names="g1"`, read as the two robots `"g"` and `"1"`, reported as `No matching robots found.`) is a real but separate and lower-severity defect -- it does not widen anything. Recorded in #2531 rather than folded in here.

## Tests

`tests/test_asset_download_selection_domain.py` (21 tests), named after the `test_teleop_device_selection_domain.py` precedent.

- 4 refusals: `names=[]`; `names=[]` + `category` (the quieter widening branch); nothing downloaded; **the asset cache directory is not created** -- checked with the real directory helper, since a guard placed after it would leave the dir behind.
- 5 tool refusals, parametrized over `","`, `" "`, `",,,"`, `" , "`, `", ,"`, asserting the message names `robots=` and never `names=`, and that `download_robots` is not called.
- 12 controls pinning every honored spelling unchanged: `None` -> all, `None`+`category` -> that category, a real subset, the three tolerated shapes, and the tool's six forwarding cases.

**9 of the 21 fail against the unfixed source** (4 library refusals + 5 tool refusals), verified by stashing the source change and re-running with the test file in place.

```
115 passed, 1 failed  # tests/test_asset_download*.py, tools/test_download_assets.py,
                      # test_ensure_meshes.py, assets docstrings, changelog gates
```

The single failure is `test_robot_descriptions_available_true_when_importable`, which fails identically on unmodified `main` in this environment (`robot_descriptions` not installed) -- verified by stashing the change and re-running.

`ruff check`, `ruff format --check` and `mypy` are clean on all three source files.

## Commits

| commit | concern |
|---|---|
| `7c539c3` | the fix + the pinned regression tests |
| `4a22ed1` | `changelog.d/` news fragment (per the fragment convention, not a `CHANGELOG.md` append) |
| `c68ceab` | AGENTS.md: record the membership-resolver carve-out on the existing subset-selector rule, and add this surface to its `Pinned by` list |

The AGENTS.md bullet is additive at a new anchor rather than an edit to the existing `Pinned by` line, to avoid the per-list-anchor conflict class described in #2507. Both test paths it names were verified to resolve on disk.

## How this was found

An AST sweep of the tree for optional-collection parameters read by truthiness (54 candidates), triaged against the documented rule. Worth recording that the sweep's most suspicious group was a **false positive**: `start_recording(cameras=...)` reads `cameras` by truthiness only to gate a validation call, and all three backends (MuJoCo, Newton, Isaac) resolve it `is not None`, so an empty camera selection already records no camera. That group is correct as written and is not touched here.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1503,6 +1503,23 @@ which side the enum is on.
 - **A selection widened to everything can also reach past the call.** `detach_teleop` stops
   the local loop once nothing is left to drive, so a detach widened from one stream to all
   of them ended a running session as a side effect of the misread.
+- **A surface that resolves the names by membership takes only the emptiness verdict.**
+  Where the selector is resolved into a dict rather than bound by position, a repeat
+  resolves to its first occurrence and a mapping and a one-shot iterator are each read
+  exactly once, so routing the shape through `name_list_error` would refuse calls that
+  are honored as written today - the same carve-out that keeps the WBC and MotionBricks
+  providers out of that domain. `download_robots(names=...)` is that case, and the
+  membership read is still owed: read by truthiness, `names=[]` downloaded 56 robots on
+  the shipped registry, and 13 - a whole category - when a `category` was also passed,
+  reporting either count as the caller's own request. Nothing had to write `[]` to get
+  there, which is what made it reachable rather than theoretical: the `download_assets`
+  tool builds the list from a comma-separated string through its own blank-field filter,
+  so the NON-empty `robots=","`, `" "` and `",,,"` each parse to zero names. Refuse the
+  empty selection ahead of the call that creates the asset cache directory, so a refused
+  selection leaves nothing behind - and refuse it again in the tool's own vocabulary, so
+  the remedy names the `robots=` that caller passed rather than the `names=` it never
+  did. Pinned by `tests/test_asset_download_selection_domain.py`, whose controls pin the
+  three tolerated spellings so the narrowing stays deliberate rather than incidental.
 - Pinned by `tests/test_teleop_device_selection_domain.py`, whose controls assert that the
   documented spellings (`names=None`, a real subset, `detach_teleop(None)`) are unchanged,
   and by the render path's `cameras` resolution, which has read the same kind of selector
--- a/changelog.d/2532-download-robots-empty-name-selection.md
+++ b/changelog.d/2532-download-robots-empty-name-selection.md
@@ -1,0 +1,27 @@
+### Fixed: an empty robot-name selection is refused, not read as "every robot"
+
+`names` on `download_robots` selects a subset of the sim robots the registry
+lists, and `None` is the documented "all of them". It was read by truthiness, so
+`names=[]` -- a selection that asks for no robot -- fell through to the branches
+that never read it. Measured on the shipped registry it downloaded 56 robots on
+its own, and 13 (the whole `humanoid` category) when a `category` was also
+passed, reporting either count as the caller's own request.
+
+Nothing had to write `[]` to reach that. The `download_assets` tool builds the
+list from a comma-separated string through its own blank-field filter, so the
+non-empty `robots=","`, `" "`, `",,,"` and `"  ,  "` each parsed to zero names --
+and an agent that emitted one of them cloned the entire registry and was told
+`status: success`.
+
+The selector is now read `is None`, and an empty selection is refused ahead of
+`get_user_assets_dir()` -- the call that creates the asset cache directory -- so
+a refused selection leaves nothing behind. The tool refuses in its own
+vocabulary, naming `robots=` rather than the `names=` its caller never passed; an
+absent or empty `robots=` still means "all", because for a single string argument
+unset and empty genuinely coincide.
+
+Only the emptiness verdict is taken locally. The shape is deliberately not routed
+through the shared `name_list_error` domain: this surface resolves each name by
+membership into a dict, so a repeat resolves to its first occurrence and a mapping
+and a one-shot iterator are each read exactly once. All three are honored as
+written today, and the tests pin them so that stays deliberate.
--- a/strands_robots/assets/download.py
+++ b/strands_robots/assets/download.py
@@ -469,21 +469,60 @@ def download_robots(
       3. Custom GitHub repos for non-Menagerie robots.
 
     Args:
-        names: Robot names to download (``None`` = all sim robots).
-        category: Filter by category (arm, humanoid, mobile, ...).
+        names: Robot names to download - a SUBSET of the sim robots the registry
+            lists. ``None`` selects all of them; an empty list selects none and
+            is refused rather than widened to all (see :exc:`ValueError` below).
+        category: Filter by category (arm, humanoid, mobile, ...). Applied only
+            when ``names`` is ``None``.
         force: Re-download even if present.
 
     Returns:
         Dict with downloaded/skipped/failed counts, names, and details.
+
+    Raises:
+        ValueError: If ``names`` is an empty selection, which asks for no robot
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Fixes the truthiness read of the `names` subset selector on `download_robots`: `names=[]` (and, via the `download_assets` tool's blank-field filter, non-empty strings like `robots=","`) previously fell through to the all-robots / whole-category branches and cloned the entire registry while reporting success. The selector is now read `is None`, an empty selection raises `ValueError` ahead of `get_user_assets_dir()` (verified: that helper performs the `mkdir`, so the ordering means a refused call leaves nothing behind), and the tool refuses in its own vocabulary naming `robots=` rather than `names=`. This is the same class AGENTS.md already rules on ("A subset selector is read by membership, never by truthiness") with `teleoperate` as the pinned precedent, and the deliberate narrowing to the emptiness verdict only (repeat / mapping / one-shot iterator stay honored, since resolution here is by dict membership) is pinned by controls rather than left implicit.

## What's good

- All 21 new tests pass on the head commit (verified independently in a clean worktree); the refusal tests assert on behavior (nothing downloaded, cache dir not created, `download_robots` not called), not implementation.
- The tool-side refusal returns an error dict rather than raising, per the AgentTool error-contract convention, and the `robots:` docstring `Args:` entry was updated so the schema the model reads documents the refusal.
- Changelog lands as a `changelog.d/` fragment; the AGENTS.md addition is anchored additively to avoid the per-list-anchor conflict class; the bare-string `names="g1"` defect is correctly scoped out to #2531 rather than folded in.
- The breaking-ish behavior change (`names=[]` now raises) is explicit via `ValueError` with an actionable remedy naming both honorable alternatives, which is exactly the shape the conventions ask for.

PR #2533, opened 2026-08-20
docs(device-connect): document the two variables that secure the Reachy daemon link

cagataycali:
## Problem

`docs/device-connect.md` is where an operator configures Device Connect. Its reference documented `REACHY_DAEMON_TOKEN` - filed under `#### Other` as "Auth token for the Reachy Mini daemon, if it requires one" - and documented neither of the two variables that secure the link that token travels on. Both were absent from the page, and from every other page: `grep -rn REACHY_DAEMON_TLS README.md docs/` returned nothing.

`REACHY_DAEMON_TLS` is the only knob that encrypts the daemon channel. Its own resolver says why that matters:

> a bearer token (see `_daemon_auth_token`) only authenticates the caller -- over plaintext `ws://` / `http://` the token and every actuator command still travel in cleartext and can be sniffed or replayed by anyone on the network segment.

So the page listed the credential and omitted the thing that protects it. Measured by following the page exactly - set the one documented variable, change nothing else:

| followed | REST URL | WebSocket URL | `Authorization` | encrypted |
|---|---|---|---|---|
| `REACHY_DAEMON_TOKEN=s3cr3t` | `http://reachy.local:8080/pose` | `ws://reachy.local:8080/ws/sdk` | `Bearer s3cr3t` | no |
| `+ REACHY_DAEMON_TLS=1` | `https://...` | `wss://...` | `Bearer s3cr3t` | yes |

The reader who followed the page got the first row: authenticated, and readable by anyone on the segment. `REACHY_DAEMON_TLS_INSECURE`, which keeps the encryption but stops verifying the daemon's certificate, was undocumented too - a weakened posture with no entry in the reference that describes it.

Of the five environment variables the Device Connect surface reads, three were documented and the two missing ones were the security-relevant pair.

## Change

Documentation only; no package code is touched.

The three variables that secure the daemon link are documented together under `#### Transport security`, with a note that `ReachyMiniDriver` reaches its robot over a second link the messaging variables above do not cover, that the link is plaintext by default, and that the token authenticates while only `REACHY_DAEMON_TLS` encrypts. Each row states its accepted spellings (`1`/`true`/`yes`/`on`, any case) and its default. The token row moves out of `#### Other` into that table: one link, one table, because a reader who finds only one of the three configures half a posture.

`REACHY_DAEMON_TOKEN` is documented on this page and not in README - the established home for Device Connect variables - so the new rows follow it there.

## Guard

`tests/test_docs_device_connect_env_reference.py` grades two properties, both derived from the package rather than from a list kept beside it, so a variable added later is graded on arrival:

- **Every variable the surface reads is documented.** The set comes from an AST walk of `strands_robots/device_connect/` for literal `os.getenv` / `os.environ.get` / `os.environ[...]` keys.
- **The variables of one link are documented under one heading.** Scoped to the module that owns the daemon link, so `MESSAGING_BACKEND` and `DEVICE_CONNECT_ALLOW_INSECURE` - a different channel, read elsewhere - are graded for presence, not placement. A control pins that scoping.

A third test parses the accepted spellings back out of the `REACHY_DAEMON_TLS` row and applies each one, so the row cannot advertise a value that leaves the link plaintext.

The plaintext default is asserted directly as well. Those assertions hold on both sides of this change - they are facts about the code, not the page, and they are the reason the omission mattered.

## Verification

Measured at `479041a2`.

Pre-fix split, this test file copied onto `upstream/main`: **3 failed / 7 passed**, each failure naming the offender.

```
test_every_device_connect_env_var_is_documented
  -> documents no row for environment variables the Device Connect surface reads:
     ['REACHY_DAEMON_TLS', 'REACHY_DAEMON_TLS_INSECURE']
test_the_daemon_link_variables_are_documented_in_one_table
  -> documented under different headings (empty = not documented at all):
     {'REACHY_DAEMON_TOKEN': 'Other', 'REACHY_DAEMON_TLS': '', 'REACHY_DAEMON_TLS_INSECURE': ''}
test_the_documented_tls_spellings_really_upgrade_the_link
  -> REACHY_DAEMON_TLS' row advertises no accepted spelling
```

All 10 pass here. The 7 that pass on both trees are the controls: the plaintext default, the closed spelling vocabulary, the verification default, the messaging variables' exemption from the placement rule, a non-vacuity check that the scan reaches the package, and two plants - a page missing a variable and a page splitting the link across two tables are both reported.

Local gate over the 71 test files that read a docs page or touch the Device Connect surface, plus the repo-wide guards:

|---|---|---|---|---|
| this branch | 2 | 3727 | 8 | 3737 |
| `upstream/main` | 5 | 3724 | 8 | 3737 |

Branch-only failures: none. The base-only three are the pre-fix failures above. The two failures shared by both trees are `tests/test_device_connect_hardening.py::{test_broadcast_dispatch_uses_validated_command,test_tell_invoke_carries_identity}`, which need `device_connect_agent_tools` - a declared dependency absent from this environment.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on this branch and 24 on `upstream/main`, none in the changed files.

## The link, configured from the documentation

Both columns run identical package code against the same probe. The only variable is what each tree's own copy of the page tells the operator to set: the capture reads that page, derives which daemon-link variables it documents and which value it says encrypts the channel, applies them, and reports the resulting URLs and header.

The Reachy Mini daemon link, configured from the documentation


yinsong1986 (approved):
## Summary

Documentation-only fix: `docs/device-connect.md` listed `REACHY_DAEMON_TOKEN` under `#### Other` but documented neither `REACHY_DAEMON_TLS` (the only knob that encrypts the Reachy Mini daemon link) nor `REACHY_DAEMON_TLS_INSECURE`, so an operator following the page shipped the bearer token over plaintext `http://`/`ws://`. The PR moves all three daemon-link variables into one table under transport security, states each row's accepted spellings and plaintext default, and adds a guard test that derives the required variable set from an AST walk of `strands_robots/device_connect/` rather than a hand-kept list, so a variable added later is graded on arrival. No package code is touched.

## What's good

- The documented rows match the code exactly: the `1`/`true`/`yes`/`on` (any case) vocabulary, the plaintext default, and the verify-by-default posture all check out against `reachy_transport.py`'s resolvers, and the spelling test round-trips each advertised value through `_daemon_use_tls()` so the row cannot drift from the code.
- The one-table placement rule is scoped to the module that owns the daemon link, with a control pinning that `MESSAGING_BACKEND` / `DEVICE_CONNECT_ALLOW_INSECURE` are graded for presence only - the guard won't misfire on the messaging channel's variables.
- Both plants (a page missing a variable, a page splitting the link across headings) are exercised, and env mutation uses `monkeypatch` / `MonkeyPatch.context()` per AGENTS.md testing conventions. Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention.

## Verification suggestions

- With the `[all,dev]` extras installed (the new tests import `strands_robots.device_connect`, which needs `device-connect-edge`): `pytest tests/test_docs_device_connect_env_reference.py -q` - 10 pass at head; copied onto `main` the three coverage/placement/spelling tests fail naming the two undocumented variables, matching the PR description's pre-fix split.

Issue #2466, opened 2026-08-19
feat(datasets): episode augmentation as a dataset transform - LeRobotDataset in/out, provenance-marked, pixel-verdict re-validated (Cosmos-Transfer-style as first backend)


yinsong1986:
# Gap

strands-robots integrates Cosmos 3's **action/policy surface** (VLA inference + SFT post-training) but has no **generation** surface: no video2video / world-model episode synthesis anywhere in the tree (verified: zero generation/predict/transfer APIs in `strands_robots/policies/cosmos3/`). The three data-diversity mechanisms that exist all *perturb within* the simulator's visual distribution:

| Mechanism | Where | When it acts |
|---|---|---|
| Domain randomization + obs noise (`randomize()`, `set_obs_noise()`) | `simulation/base.py:4239/4251`, example 12 | collection time |
| GS photoreal backdrop compositing | `mujoco_gs`/`isaac_gs` + `strands_robots/rendering` | render time |
| `TrainSpec.augmentation` -> GR00T `--color_jitter_params` etc. | `training/base.py:194`, `training/groot.py:230+` | train time |

Generative augmentation - record N episodes, synthesize visually diverse variants, train on many times N - is the one mechanism that escapes the sim's visual distribution, and the one that multiplies data without robot or sim time.

# Proposal: a dataset transform surface

This is not a policy and not a trainer; it is a third provider shape - a **dataset transform**:

- **Contract**: LeRobotDataset in -> augmented LeRobotDataset out. Video observation streams are transformed (Cosmos-Transfer-style domain/style transfer); action/state/timestamp columns pass through byte-identical.
- **Provenance is mandatory**: every generated episode carries `synthetic=true` (+ source episode id, transform name/version) in episode metadata, so training filters and the judge agent treat generated pixels honestly. Silent mixing of generated and recorded data is the failure mode this field exists to prevent.
- **Re-validation as the acceptance gate**: after transformation, deterministic predicates (and the judge, when it lands) re-score each generated episode; a generated episode that flips its deterministic verdict is discarded and counted, measured not assumed. Same eval-integrity doctrine as the Isaac box-proxy issue.
- **Optional extra**, `require_optional()`-gated; the transform *surface* is the deliverable and Cosmos Transfer is its first backend. Honest caveat: depends on upstream generation-model availability and licensing - to be verified before implementation, and the surface must not assume any single vendor's model.

# Acceptance

[...]

PR #2534, opened 2026-08-20
test(transforms): real-weights Cosmos-Transfer2.5 run through the pipeline seam

### LINKED ISSUE #2530: tests_integ(transforms): one real-weights Cosmos-Transfer run through the pipeline seam before the generative path is relied on
## What

Run `CosmosTransferTransform` once end-to-end with **real Cosmos-Transfer weights bound** to the pipeline seam, and pin the shape as a `tests_integ/` item so the generative path has a real-inference test the way every policy backend does (AGENTS.md convention 8).

## Why

PR #2480's verification (review [4983459931](https://github.com/strands-labs/robots/pull/2480#pullrequestreview-4983459931)) graded the pipeline seam through the public API - dotted-path binding, every refusal, the determinism key, round trip and verdict gate with a fake pipeline - but deliberately stopped at the injection seam:

> I did **not** exercise a real Cosmos-Transfer generation: the actual video2video weights would not co-fit alongside what this box already had resident, so the generative path is verified only up to the injection seam. Worth someone running once with real weights bound before this is relied on for a dataset that matters.

So the one unverified claim is that a real generation stack, wrapped per the documented adapter, produces a dataset the round-trip and provenance guarantees hold on. That is a GPU + weights availability question, not a surface question, which is why it is a follow-up rather than part of #2480.

## How

The diffusers-hosted checkpoints are the lowest-friction binding (no source install), via the adapter documented in `strands_robots.transforms.cosmos_transfer`'s module docstring and `_PIPELINE_HELP`:

- `Cosmos2_5_TransferPipeline` (`nvidia/Cosmos-Transfer2.5-2B`) - `__call__`-shaped, `.frames`-returning, `generator=`-seeded, conditions on `controls=` (edge/depth/seg maps derived from the source pixels), so the adapter owns that derivation (e.g. per-frame Canny for the edge controlnet variant, per the upstream example).
- Alternatively `Cosmos2_5_PredictBasePipeline` in Video2World mode (`video=` conditioning) as the simpler first smoke, though Transfer is the namesake.

Acceptance:

- [ ] `tests_integ/transforms/` test: record (or reuse) a small real LeRobot dataset -> `CosmosTransferTransform(pipeline=<real adapter>).transform(spec)` -> reopen the output; assert schema parity, action/state byte-equality, provenance rows naming the real pipeline `version`, pixels changed.
- [ ] Same run twice with the same `seed`: byte-identical outputs (or document precisely where the real pipeline breaks determinism, e.g. nondeterministic CUDA kernels, so the provenance `seed` field's meaning is honest for real backends).
- [ ] Licensing note verified for the deployment running the test (NVIDIA Open Model License), per the caveat `validate()` already states.

## Out of scope

- Shipping a first-party wrapper class for any single vendor's pipeline - the seam is deliberately vendor-neutral; whether a maintained wrapper is worth it can be decided after this run shows what the adapter has to own in practice.

Refs #2466, #2480.

yinsong1986:
## What

One real-weights Cosmos-Transfer run through `CosmosTransferTransform`'s pipeline seam, pinned as a `tests_integ/` item so the generative path has a real-inference test the way every policy backend does (AGENTS.md convention 8).

New: `tests_integ/transforms/test_cosmos_transfer_real_weights.py` (7 tests, `gpu` marker) plus a changelog fragment.

Closes #2530

> **Stacked on #2480.** The transform surface this exercises only exists on #2480's branch (`feat/dataset-transform-augmentation`), which is still open, so this branch is based on that head (`4674448c`) and the diff below includes #2480's commits until it merges. The commits belonging to *this* PR are the ones on top; once #2480 lands and this branch absorbs `main`, the diff collapses to the two new files.

## Why

PR #2480's verification (review 4983459931) graded the seam end to end with a fake pipeline and deliberately stopped at the injection seam - the real video2video weights would not co-fit on the box that ran it. The one unverified claim was that a real generation stack, wrapped per the documented adapter protocol, produces a dataset the round-trip and provenance guarantees hold on. This PR is that run, executed on real hardware, plus the test that keeps it pinned.

## The real run (executed, not just authored)

Binding: diffusers-hosted `nvidia/Cosmos-Transfer2.5-2B` via `Cosmos2_5_TransferPipeline` (diffusers 0.40.0), revision `diffusers/general` + edge controlnet revision `diffusers/controlnet/general/edge`, bf16, `enable_model_cpu_offload`, **real Cosmos Guardrail enabled** (`cosmos_guardrail` 0.3.1 - the pipeline refuses to run without it). The adapter lives in the test module and owns exactly what #2530 asked to be measured:

- **control derivation** - per-frame Canny edge maps from the source pixels (the upstream edge-controlnet recipe);
- **resolution round-trip** - pipeline needs height/width divisible by 16, so source frames are upscaled to a 256x256 work resolution and generated frames resized back to the source shape (the transform contract requires shape/dtype parity);
- **seeding** - the seam's derived per-variant seed becomes a CPU `torch.Generator`, the pipeline's only noise source;
- **version** - the exact checkpoints + diffusers version, recorded into every provenance row.

Measured on an NVIDIA L4 (24 GB): `7 passed in 143.81s` cold / `108.57s` warm, peak ~19 GB GPU memory, ~40 GB of HF cache across model + guardrail weights, ~45 s per generation (17 frames, 10 steps).

## Acceptance criteria from #2530

- [x] `tests_integ/transforms/` test: record a small real LeRobot dataset -> `CosmosTransferTransform(pipeline=<real adapter>).transform(spec)` -> reopen the output; schema parity, action/state byte-equality, provenance rows naming the real pipeline `version` (asserts the model id + controlnet revision appear, never `"unversioned"`), pixels changed (mean shift ~64/255 vs source).
- [x] Same run twice with the same `seed`: **byte-identical outputs** - measured, so the test asserts byte-equality outright. On this stack (L4, torch 2.11 + CUDA 13, diffusers 0.40.0, bf16, cpu-offload) no nondeterministic CUDA kernel surfaced at these shapes; the module docstring documents the stack and says the honest fix if a future kernel/hardware change trips this is to document the divergence, not silently loosen the assertion. So the provenance `seed` field's meaning is honest for this real backend.
- [x] Licensing note verified for the deployment that ran this: both `nvidia/Cosmos-Transfer2.5-2B` and `nvidia/Cosmos-1.0-Guardrail` are auto-gated behind the **NVIDIA Open Model License**, which the account running the test accepted on Hugging Face for both repos; the guardrail (safety checker) was installed and enabled for the run rather than stubbed out, per the license expectation the diffusers pipeline enforces. The test's module docstring carries the same caveat `validate()` states: verify availability and licensing per deployment.

## Test surface

- `TestRealWeightsRoundTrip` (5 tests): counts, schema parity, action/state byte-equality (with a non-vacuous-on-zeros guard), pixels changed, provenance naming the real pipeline.
- `TestRealWeightsDeterminism` (2 tests): decoded image column byte-identical across two identically-seeded `transform()` runs; provenance seeds match.
- Skips cleanly (module-level) without lerobot / torch / CUDA / diffusers>=0.40, and skips at the fixture on gated-weights access failure, so `hatch run test-integ` on a box without the stack is unaffected.
- Exactly 2 pipeline generations for the whole module (module-scoped fixtures; the round-trip run doubles as determinism run A).

`hatch run lint` clean; unit suite (`hatch run test`) green apart from this dev box's pre-existing environmental failures; the new files are integ-only and touch no package code.

## Out of scope / follow-ups

- No first-party vendor wrapper class is shipped - per #2530's out-of-scope note, the seam stays vendor-neutral; what the adapter had to own is now measured and documented in the test for whenever that decision is revisited.
- One practical note for future runners: `cosmos_guardrail`'s blocklist loads NLTK data straight out of the HF cache snapshot, and NLTK >= 3.10's hardened opener refuses the cache's symlinked files; dereferencing the symlinks under the guardrail snapshot's `blocklist/` (copy in place) is the workaround. Upstream packaging issue, documented here for the next runner rather than worked around in the test.

## Project board

Issue #2530 should move to "In review" on the Strands Labs - Robots board.

Refs #2466, #2480.


yinsong1986:
Fixed the failing **call-test-lint / Test and Lint** check: the failure was `tests/simulation/test_replay_episode_index_domain.py::TestNoEpisodeIndexSurfaceDrifts::test_the_surface_set_is_exactly_the_pinned_one` discovering the transform surfaces (`transform_frames` / `derive_variant_seed`) via their `Args:` wording, plus the coverage shortfall that follows from pytest stopping there before `tests/transforms/` ran. `main` already carries the remedy - #2540 scopes the sweep to surfaces that can resolve an index against a dataset, and its planted-surface test is exactly the transform-backend shape - so the fix is a clean, conflict-free merge of `main` (combined diff: 0 lines; the PR's own diff is unchanged). Validated locally under CI's environment (`MUJOCO_GL=osmesa`): the sweep file passes, lint clean, coverage 95.44%, and the only local failures are this box's known X11/GL environmental ones in files this PR does not touch. CI re-running on `4083ba96`.

yinsong1986:
Merged `main` in and resolved the add/add conflicts in `strands_robots/transforms/base.py`, `strands_robots/transforms/cosmos_transfer.py`, `strands_robots/transforms/mock.py` and `tests/transforms/test_cosmos_transfer.py` — this branch carried #2480's pre-squash commits, and the squashed #2480 on `main` is a strict superset (adds the `source_episode` domain guards and the `__call__`-shaped adapter docs/tests), so the resolution takes `main`'s side; the branch's own content (`tests_integ/transforms/test_cosmos_transfer_real_weights.py` + changelog fragment) is untouched. Merge commit 7e8519d0. Lint clean; full unit suite green (35235 passed, 292 skipped, coverage 95.57%). Ready for re-review.

cagataycali (approved):
Approving. This closes the rule-8 gap for the generative path with a real-inference test, and it is built the way this repo asks for. What I verified on 23f5a8d8, rather than taking the description's word for it:

**Placement and gating.** It lives in tests_integ/transforms/, so `hatch run test` never loads a 2B checkpoint. The guard stack is layered correctly and each layer skips rather than errors: importorskip for lerobot/torch/diffusers, an explicit Cosmos2_5_TransferPipeline attribute check with the required diffusers version in the skip reason, a torch.cuda.is_available() check, and pytest.skip.Exception on gated-repo or missing-weights failures. The gpu marker plus timeout(1800) keeps it off a default selection and bounds two real generations.

**The dataset root is the test's own.** DatasetRecorder.create with root= derived from tmp_path_factory.mktemp(), so nothing reads or writes $HF_LEROBOT_HOME. tests_integ/ is out of scope for that convention since it records real datasets, so passing root= here is stricter than required and I would keep it.

**The assertions are non-vacuous, which is the part that usually rots.** _record_source_dataset deliberately writes per-frame distinct non-zero actions and states, and the byte-equality test re-asserts np.any(src_col != 0.0) beside it, so the pass-through check cannot succeed on a buffer of zeros. The source frames are a moving square precisely so the edge controlnet has real edges to condition on. The provenance test pins both the model id and the controlnet revision inside transform_version, so an unversioned provenance row fails rather than passing quietly.

**test_pixels_changed is sound as written.** I considered whether the 0.01 floor could be satisfied by video re-encoding alone, which would make it pass on a transform that silently no-opped. It cannot, and the reason is that it compares the mean over every frame and pixel: codec noise is approximately zero-mean, so it does not shift a full-column mean by that much, whereas a wholesale regeneration does. Aggregating before differencing is what makes the loose-looking floor a real discriminator.

**Names and provenance are on the right side of the line.** TestRealWeightsRoundTrip and TestRealWeightsDeterminism describe behaviour; the issue references sit in docstrings, which is where they stay useful.

**Graders pass on this head:** test_no_host_paths.py, test_log_strings_are_ascii.py, test_test_case_names_describe_behaviour.py, test_recording_root_is_not_the_shared_cache.py - 28 passed. The changelog change is a changelog.d fragment.

One note, not a request. The broad OSError beside GatedRepoError and LocalEntryNotFoundError reads at first like a tuple whose named members contribute no scope, but those are third-party classes that can be re-parented between releases, so naming both them and the builtin superclass is the sanctioned hedge rather than the collapse the convention warns about. Leaving it alone is correct.

On the byte-identical determinism assertion: asserting it outright rather than within a tolerance is the right call, and the module docstring already records the exact stack it was measured on and says the remedy for a future divergence is to document it here rather than to loosen the assertion. That is the tripwire behaving as intended.

Not blocking on anything. The required check is still running, so this needs no further action from the author - it can merge once call-test-lint reports.

PR #2535, opened 2026-08-20
fix(tools): record the operator's HITL reply at every gate, not only the mesh one

cagataycali:
## Problem

Three tools stop and ask a human before an agent-issued command reaches a robot or a training run:

- `robot_mesh` gates the physical-actuation actions (`emergency_stop`, `broadcast`, `tell`, `send`, `stop`, `rpc`)
- `use_ros` gates a `publish` / `service_call` / `action_send_goal` aimed at a safety-critical graph surface (`/cmd_vel`, `/emergency_stop`, `/joint_trajectory`, ...)
- `lerobot_train` gates the `extra_flags` that control output paths, telemetry and code loading (`output_dir`, `wandb.api_key`, `policy.pretrained_path`, ...)

Each gate owes the operator's reply two things that pull in opposite directions. It must **not** reach the model - a flat sentinel goes back instead, so an agent that authors the approval reason cannot make the operator's typed answer carry data into the context. And it must reach the **local audit log**, because every gate accepts a canonical affirmative only (`y` / `yes` / `approve` / `approved`): a reply that carries a *reason* is therefore always a decline, and that row is the only place the reason survives.

The echo half was honoured at all three gates. The record half was honoured at one.

Driving each gate through its real entry point with a reasoned decline and then an approval, and reading the audit log back with `read_audit_log()`:

| gate | verdict | reply reached the model | audit rows | reply in the row |
|---|---|---|---|---|
| `use_ros` decline | halted | no | **0** | no |
| `use_ros` approve | proceeded | no | **0** | no |
| `lerobot_train` decline | halted | no | **0** | no |
| `lerobot_train` approve | proceeded | no | **0** | no |
| `robot_mesh` decline | halted | no | 1 | yes |
| `robot_mesh` approve | proceeded | no | 1 | yes |

`use_ros` and `lerobot_train` wrote nothing on either outcome - and not just no audit row: capturing every log record emitted during a declined `/cmd_vel` publish returns **zero records**. So an operator who declines an agent's request to drive a robot, or to redirect a training run's output path, leaves no trace that a gate fired at all, let alone why they refused. The approved path is equally silent, so nothing records that a human authorised the agent to reach that surface - which is the first question an incident review asks.

## Change

One owner writes the row: `strands_robots/tools/_hitl_audit.py::log_operator_response`. Each of the three gates calls it on **both** outcomes.

A single owner rather than a row each gate spells itself: a forensic reader greps one phrasing, and a gate that formatted the row locally could drift to `operator refused` and become invisible to a search for `operator declined`, with no way to tell that apart from a gate that wrote nothing. `robot_mesh` therefore routes through it too, and its rows are unchanged - asserted byte-identical against the payload it wrote before.

The row carries the verb, the target, the verdict and the operator's literal reply, `repr`-quoted (so a reply containing a newline cannot forge a second record) and bounded to 500 characters, matching the audit-detail bound the mesh tool already applied. The audit module is imported inside the call, as the mesh tool already does, so a tool that gates a ROS graph or a training run does not pay for the transport stack on import. An audit write that fails is logged at DEBUG and swallowed: the record is owed, but an unwritable audit log must not turn a decline into an exception the gate never accounted for.

`docs/ros2-integration.md` already stated the echo half ("the reply text is never echoed back into the agent's context") and was silent on the record; it now states both, and why a reasoned reply is always a decline.

## Tests

`tests/tools/test_hitl_operator_response_audit.py` grades the **observable** - a row exists, carrying the reply - rather than which function writes it, and derives the set of gates from the `interrupt()` call sites so a fourth gate is graded on arrival instead of inheriting the silence.

**12 failed / 12 passed** against `upstream/main`; 24 pass after. The failures are behavioural, e.g. `use_ros wrote 0 operator rows for a decline, want 1: []` and `the declined publish left no readable audit row: []`.

The 12 that pass on both trees are the controls, and each fails for a specific wrong fix:

- the reply is still absent from every gate's model-visible text (fails if recording it starts echoing it)
- `robot_mesh`'s recorded payload is byte-identical (fails if the shared owner changes the row shape)
- every gate reaches the same verdict when the audit write raises `OSError` (fails if the record becomes load-bearing for safety)
- the `interrupt()`-site scan matches the graded table (fails if the discovery rule breaks, so a clean result cannot be vacuous)

One assertion is worth naming: the row must name the verb and the target, and `emergency_stop`'s target is legitimately **empty** - it is fleet-wide, so no single peer is named. The expected target is pinned per gate rather than asserted non-empty, which keeps that deliberate.

## Verification

- `ruff check` / `ruff format --check` over `strands_robots tests tests_integ`: clean
- `mypy strands_robots tests tests_integ`: **24 errors on this branch, 24 on a pristine `upstream/main`**, none in the changed files
- 233 test files covering the three tools, every test touching the audit helpers or the operator-response path, and the repo-wide docstring/changelog/source-string guards, measured at `123bbd4`:

|---|---|---|---|---|
| this branch | 12 | 11268 | 90 | 11370 |
| `upstream/main`, with the new test file copied in | 24 | 11255 | 90 | 11369 |

Branch-only failing ids: **empty**. Base-only: exactly the 12 pre-fix failures. The `+1` collected is the new module enrolling in the repo-wide `test_no_teardown_method_imports_a_stdlib_module[tools/_hitl_audit.py]` case, which passes - so the passed delta of 13 is those 12 plus that one.

The 12 failures shared by both trees are dependency-absent locally and green on CI: 8 in `tests/mesh/test_iot_connect_timeout_domain.py` (`awsiotsdk`), 2 in `tests/test_device_connect_hardening.py` (`device_connect_agent_tools`), and 2 `TestInProcessTrainerCorrectness` cases (an older `draccus` than the declared floor). None is in a file this branch touches.

## Artifact

One capture script per tree drives all three gates in a single operator session against a real audit directory, then reads the log back with `read_audit_log()`. Same script, same package code elsewhere; the only variable is the branch.

operator-verdict audit trail

The gate verdicts are identical in both trees (refused / refused / allowed) and the reply reaches the model in neither. What changes is that the log goes from **1 of 3** prompts recorded to **3 of 3**, and the two declines carry the operator's reason (`'n - not while the cell door is open'`, `'n - that path is the shared archive'`) instead of leaving a gap.


yinsong1986 (approved):
## Summary

Closes the record half of the HITL operator-reply contract: `use_ros` and `lerobot_train` gated safety-critical commands but wrote no audit row on either outcome, so a declined `/cmd_vel` publish or `output_dir` override left no trace that a gate fired or why an operator refused. The fix routes all three gates through a single owner (`strands_robots/tools/_hitl_audit.py::log_operator_response`) that writes the row on both verdicts, repr-quoted and bounded to the same 500-char detail limit the mesh tool already applied, while the reply continues to never reach the model. Verified locally at head `6a2136db`: the new suite is 24/24 on the branch and exactly 12 failed / 12 passed with the test file copied onto `upstream/main` (matching the PR's claim); `ruff check` and `ruff format --check` are clean on all changed files; the one failure in the wider `tests/tools/` run (`test_lerobot_train_size_knob_domain.py`) reproduces identically on `upstream/main` and is environment-local, not from this diff.

## What's good

- The `except Exception` in the new module is the one sanctioned shape: a documented, tested recovery path matching `_audit_tool_action`'s existing pattern, with the AGENTS.md citation inline and an `OSError`-injection control asserting the gate's verdict is unchanged when the audit write fails.
- `robot_mesh`'s recorded payload is pinned byte-identical, so the consolidation cannot silently drift the row shape a forensic reader greps for.
- The graded gate set is derived from the `interrupt()` call sites via an AST scan, so a fourth gate is graded on arrival rather than inheriting the silence - and a control pins the scan against the table so a clean result cannot be vacuous.
- Compliance is complete: changelog fragment (step 3), AGENTS.md pin entry, `STRANDS_MESH_AUDIT_DIR` redirected to `tmp_path` in the on-disk test, no host paths, ASCII-only strings, lazy import of the transport stack documented at the import site.

PR #2536, opened 2026-08-20
fix(simulation): report a refused default_robot instead of scoring a robotless episode

cagataycali:
## Problem

`BenchmarkProtocol.on_episode_start` auto-loads `default_robot` when the sim is empty, and discarded the envelope `add_robot` returns:

```python
robots = sim.list_robots()
if not robots:
    sim.add_robot(name="robot", data_config=self.default_robot)
    return
```

`add_robot` reports refusal in its return value rather than by raising, and it refuses for reasons a spec author hits: a `default_robot` naming one of the nine hardware-only registry entries, an asset that is not on disk, a model the backend cannot load. With the envelope dropped the hook returned normally, and the episode ran against a scene with no robot in it.

The sim can be empty at that point even though `evaluate_benchmark`'s up-front `"No robots in sim. Add one first."` guard passed: a `DeclarativeBenchmark` that declares a `scene` loads it first, and `load_scene` rebuilds the world, so the robot registry is empty by the time the base hook looks.

Measured end to end through `evaluate_benchmark`, with a spec declaring `default_robot: reachy2` (a real registry name, hardware-only):

| | upstream/main | this branch |
|---|---|---|
| tool `status` | `success` | `error` |
| `success_rate` | `0.0` | absent |
| `success_measured` | `True` | absent |
| `avg_steps` | `5/5` | absent |
| `sim.list_robots()` after the eval | `[]` | `[]` |
| reason reported | none | names `reachy2`, why there is no model, and the two remedies |

So a benchmark that could not load its robot reported a plausible 0% success rate, with `success_measured: True` - the field a consumer reads to decide whether the number is trustworthy - and the actionable reason discarded.

## Change

Read the refusal out of the envelope and raise, the way the sibling setup call in the same hook chain already does. `DeclarativeBenchmark.on_episode_start` checks its own `load_scene` envelope and raises a `RuntimeError` with the backend's reason attached; the base impl's `add_robot` was the half that did not. `PolicyRunner.evaluate` already converts an exception from this hook into a structured error naming the spec, so no new reporting channel is needed:

```
on_episode_start failed in DeclarativeBenchmark: DeclarativeBenchmark:
add_robot(data_config='reachy2') failed: Robot 'reachy2' is registered for real
hardware only (LeRobot type 'reachy2'): its registry entry declares no
simulation asset, so there is no model to load. [...] drive it as hardware with
Robot('reachy2', mode='real'), or pass urdf_path= to supply a model of your own.
```

The check is `isinstance(result, dict) and result.get("status") == "error"` - the sibling's exact shape - so a duck-typed sim that returns nothing from `add_robot` keeps working. Nothing changes on the accepted path.

This is the consequence the sibling required attribute is already validated to prevent. `max_steps`' own `Attributes:` entry says it is checked where it is read "rather than an evaluation over episodes of zero length that still reports a 0% success rate", and `LiberoAdapter` documents the same distinction for its own controller install: a failure in *setup* must not read as a *policy* that scored zero on an exit-0 run.

## Tests

`tests/simulation/test_benchmark.py::TestRefusedDefaultRobotIsASetupFailure` pins the hook contract, and `tests/simulation/test_policy_runner_benchmark.py::TestARefusedDefaultRobotIsReportedNotScored` pins it end to end through `PolicyRunner.evaluate`.

Copied onto `upstream/main` (`ec74c5d1`): **3 failed / 90 passed**, the end-to-end one reporting `assert 'success' == 'error'`. 93 pass here.

The six that pass on both trees are the controls, and each fails for a specific wrong fix:

- an accepted add still returns, and the eval still reports a rate for it
- `None`, a dict with no `status` key, a string and `0` are all left alone - only an error envelope is a refusal
- both setup calls in the chain read a refusal out of their envelope, so a change that drops either check is caught

## Verification

Measured at `fbf9688`, MuJoCo 3.12.0.

- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `upstream/main` worktree, branch-only empty.
- Blast radius - every test naming the benchmark hook, protocol, `default_robot`, `evaluate_benchmark` or `DeclarativeBenchmark`, all of `tests/benchmarks/libero`, plus the repo-wide docs/docstring/changelog guards (102 files): branch **1 failed / 3263 passed / 66 skipped**, base **4 failed / 3260 passed / 66 skipped**, both collecting 3264. Failed delta == passed delta == 3 == the pre-fix failure count; branch-only failures empty. The one shared failure is `test_declared_qp_backends_are_real_qpsolvers_extras`, which needs `qpsolvers` (installed by `.[all,dev]` in CI, absent from this local venv).

## Artifact

Rendered headless with `MUJOCO_GL=egl`. Panel 1 is the scene the rollout actually measured; panel 2 is what `evaluate_benchmark` reported for it. Both renders and the whole control row are byte-identical across the two trees - what changes is the verdict for a refused robot, not the scene.

benchmark refused robot

Closes nothing; found while auditing which callers of a `{"status": ...}`-returning function check what it returns.


yinsong1986 (approved):
## Summary

Narrow correctness fix: `BenchmarkProtocol.on_episode_start`'s auto-load path now reads the envelope `sim.add_robot` returns and raises `RuntimeError` with the backend's reason when the add is refused, instead of returning normally and letting the episode score a robotless scene as `success_rate: 0.0` with `success_measured: True`. The check (`isinstance(result, dict) and result.get("status") == "error"`) is byte-for-byte the shape the sibling `load_scene` check in `DeclarativeBenchmark.on_episode_start` already uses, so the two setup calls in the hook chain now agree, and `PolicyRunner.evaluate`'s existing broad `except` converts the raise into the structured error naming the spec - no new reporting channel. This is the AGENTS.md "raise on fatal errors / no silent defaults on error" convention applied to the one caller in the chain that was dropping its envelope; not a one-way door, since the only behaviour that changes is a run that was previously reporting a fabricated score.

## What's good

- Verified locally at the head SHA: the two new test classes fail 3/9 with the fix reverted to `ec74c5d1` (exactly the pre-fix failure count the description claims) and all 93 tests in the two touched files pass on the branch.
- Controls pin the scoping: an accepted add still returns and still scores; `None`, a status-less dict, a string and `0` are all left alone, so a duck-typed sim that returns nothing keeps working.
- Changelog fragment present (`changelog.d/2536-benchmark-refused-default-robot.md`), docstring gains a `Raises:` section, ruff check/format clean on the three touched files, no non-ASCII or host paths in the diff.
- Scope discipline: one production hunk, everything else is tests and the fragment.

PR #2537, opened 2026-08-20
fix(tools): report a GR00T restart whose teardown could not free the port

cagataycali:
## Problem

`gr00t_inference(action="restart")` frees a port and rebinds it for a new checkpoint - the action's documented reason for existing is "e.g., to swap checkpoints".

`_stop_service` returns an error exactly when something still holds the port after SIGTERM *and* SIGKILL, and its message says why that matters in as many words:

> GR00T service on port 5555 in container gr00t-jetson survived SIGTERM and SIGKILL: pid(s) 999 still match. **The port is still held, so a restart will fail to bind it.** Check the container with 'docker exec gr00t-jetson ps -ef'.

`restart` was the one caller that performs that bind, and it discarded the verdict.

The rebind cannot detect the collision on its own:

- the launch is a **detached** `docker exec -d`, so `subprocess.run(..., check=True)` returns 0 as soon as the daemon accepts the exec - it says nothing about whether the server inside bound the port;
- the readiness wait is `_is_service_running(port)`, a bare `socket.connect_ex(("localhost", port))`, which the **surviving** server answers on the first poll.

Measured through the real dispatch against a stubbed docker/lsof/kill process table, one pid surviving SIGKILL:

| | teardown fails | control: teardown frees the port |
|---|---|---|
| `status` | `"success"` | `"success"` |
| `checkpoint_path` | `/data/checkpoints/policy-v2` | `/data/checkpoints/policy-v2` |
| `message` | `GR00T ZMQ service started on port 5555 (server: n1.5)` | *(identical)* |
| port held by | `['999']` | none |
| actually serving | **`policy-v1` (the old checkpoint)** | `policy-v2` |

The two responses are byte-identical. A restart that could not swap the checkpoint is indistinguishable from one that did, and every subsequent inference request reaches the old policy.

## Change

`restart` reads the teardown verdict and refuses the rebind when it is an error, naming the port, the checkpoint that was **not** started, and carrying the teardown's own reason:

```
Restart aborted: the service on port 5555 could not be stopped, so
'/data/checkpoints/policy-v2' was not started and the previous checkpoint is
still serving. GR00T service on port 5555 in container gr00t-jetson survived
SIGTERM and SIGKILL: pid(s) 999 still match. ...
```

The sibling call site - the `stop` action at `gr00t_inference.py:802` - already returned that verdict unchanged; a control test pins that its response has not moved. The action's docstring now states the refusal.

An AST pin asserts no caller of `_stop_service` discards the result, so a third caller cannot drop it.

`test_restart_stops_then_starts_in_order` stubbed `_stop_service` with `side_effect=lambda port: calls.append("stop")` - `list.append` returns `None`, so the fake could express neither verdict, which is why the discard was invisible there. It now answers with a real envelope (matching the `_start_service` stub two lines below) and still passes against pre-fix source, so its stated intent is preserved rather than relaxed.

## Not in scope

`_start_container` discards `_remove_container` on its `force=True` path (`gr00t_inference.py:1804`), while the other call site (`lifecycle="teardown"`) checks it. Measured: when `docker rm -f` fails, the following `docker run --name` fails too, so the caller gets `status="error"` either way - `'docker run failed: docker: Conflict. The container name "/gr00t" is already in use.'` - identical to the control where the removal succeeded. That is a misdiagnosis (it blames the run, not the removal), not a silent success, so it is a different failure mode and left alone here.

Making `_is_service_running` verify server *identity* rather than port occupancy would close the gap from the other side, but it needs a protocol-level probe and is a design choice; this change only requires the caller to read a verdict that is already computed.

## Verification

Measured at `f49fa11f`.

Pre-fix split (both test files copied onto `upstream/main`): **6 failed / 18 passed** in the teardown module, every failure behavioural -

```
test_a_restart_that_cannot_free_the_port_is_refused
  AssertionError: reported 'success' ('GR00T ZMQ service started on port 5555
  (server: n1.5)') while pid 999 still holds port 5555, so the new checkpoint
  cannot have bound it. The response is indistinguishable from a restart that
  really did swap the checkpoint.

test_every_stop_service_call_site_consults_the_verdict
  AssertionError: _stop_service result discarded at line(s) [941].
```

Six controls pass on both trees: the ordinary restart starts the new checkpoint, a restart onto a free port proceeds, the missing-checkpoint guard still runs before any teardown, and the `stop` action's response is byte-identical to calling `_stop_service` directly.

Gate - 68 files (every test naming the service/container lifecycle helpers, all of `tests/tools`, and the repo-wide docstring / changelog / string-hygiene guards), same selection on both trees:

|---|---|---|
| passed | 3046 | 3040 |
| failed | 0 | 6 |
| collected | 3054 | 3054 |

Branch-only failures: none. Base-only: exactly the six above. No pre-existing failures in the selection. `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 against a pristine `upstream/main` worktree, none in the changed files.

## Artifact

Both columns run the same capture script against the same stubbed process table; only the restart branch differs. Ground truth is port ownership - pid 999 survives SIGKILL, so the new server cannot have bound 5555.

gr00t restart teardown verdict

### DIFF
--- a/changelog.d/2537-gr00t-restart-honours-the-teardown-verdict.md
+++ b/changelog.d/2537-gr00t-restart-honours-the-teardown-verdict.md
@@ -1,0 +1,2 @@
+### Fixed
+- **tools**: `gr00t_inference(action="restart")` now reports a teardown that could not free the port instead of reporting the new checkpoint as serving. The rebind cannot detect the collision itself - the launch is a detached `docker exec -d` and the readiness wait is a bare TCP connect that the surviving server answers - so a restart whose stop failed returned `status="success"` naming the new `checkpoint_path` while every inference request still reached the old one.
--- a/strands_robots/tools/gr00t_inference.py
+++ b/strands_robots/tools/gr00t_inference.py
@@ -489,7 +489,10 @@ def gr00t_inference(
         - ``stop``: Terminate a running service on the specified ``port``.
         - ``status``: Check whether a service is running on the specified ``port``.
         - ``list``: Discover all running services across common ports (5555-5558, 8000-8003).
-        - ``restart``: Stop and re-start a service (e.g., to swap checkpoints). Requires ``checkpoint_path``.
+        - ``restart``: Stop and re-start a service (e.g., to swap checkpoints). Requires
+          ``checkpoint_path``. Refused when the running service cannot be stopped: the
+          port is still held, so the new checkpoint cannot bind it and the previous one
+          keeps serving.
         - ``find_containers``: List available Isaac-GR00T Docker containers.
         - ``build_image``: Clone Isaac-GR00T at ``repo_tag`` and run ``bash docker/build.sh``.
           Idempotent - skips the build when ``image_name`` already exists in the local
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

`gr00t_inference(action="restart")` previously discarded `_stop_service`'s verdict and rebound the port regardless; because the launch is a detached `docker exec -d` and the readiness wait is a bare TCP connect that a surviving server answers, a restart whose teardown failed reported `status="success"` naming the new `checkpoint_path` while the old checkpoint kept serving. The fix has `restart` read the teardown result and refuse the rebind on error with a structured error dict naming the port, the checkpoint that was not started, and the teardown's own reason. Verified locally at `f49fa11f`: the pre-fix discard is confirmed on `upstream/main`, `_stop_service` returns a dict with `status` in every branch (including its catch-all) so the new read cannot KeyError, the full teardown + gr00t_inference test modules pass (102/102), ruff check/format are clean on the changed files, and no non-ASCII was added to user-facing strings.

## What's good

- The refusal follows the tool contract (error dict, never raise) and carries the teardown's diagnostic instead of dropping it, so the caller learns which pid still holds the port.
- The AST pin (`test_every_stop_service_call_site_consults_the_verdict`) closes the class, not just the instance - a third caller cannot silently discard the verdict.
- Controls pin both untouched paths: the `stop` action's response is asserted byte-identical to `_stop_service`, and the free-port / ordinary-restart paths still start the new checkpoint.
- The `_stop_service` stub in `test_restart_stops_then_starts_in_order` was upgraded to return a real envelope rather than relaxed, and the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention (AGENTS.md PR workflow step 3).
- Scope discipline: the adjacent `_start_container` discard is measured, shown to be a different (misdiagnosis, not silent-success) failure mode, and explicitly left out.

PR #2542, opened 2026-08-20
fix(mesh): report a mistyped STRANDS_MESH_BACKEND from the resolver that sees it

cagataycali:
## Problem

`STRANDS_MESH_BACKEND` is resolved by two readers, and they disagreed about the one case that matters.

`session._backend_choice` runs **first**, on every session and publish path, and its verdict is what `session._is_transport_backend()` gates on. `transport.factory` resolves the variable only once that verdict is `iot` or `bridge` - `get_transport()` has exactly one caller, inside that gate. The factory reported an unrecognized value; the gate in front of it did not. So a typo resolved to `zenoh` at the gate, the factory was never consulted, and **the report could not fire for the input class it named**.

Measured against `upstream/main`, driving the real `mesh.core.init_mesh` and then 500 publish-gate calls:

| exported | resolved | factory consulted | traffic carried by | console names the variable |
|---|---|---|---|---|
| `iot` | `iot` | yes | IoT transport | - |
| `zenoh` | `zenoh` | no | `zenoh.Session` | - |
| **`iott`** (meant `iot`) | **`zenoh`** | **no** | **`zenoh.Session`** | **0 of 2 lines** |

The typo row is indistinguishable from an explicit `STRANDS_MESH_BACKEND=zenoh`: same resolved backend, same transport, same console. `mesh.iot.provision.env_vars()` writes `STRANDS_MESH_BACKEND=iot` into an operator's environment file, so a mistyped value is the ordinary way this happens, and the cost is a fleet talking over local Zenoh instead of the cloud broker it was pointed at - peers on other networks silently never find each other, with nothing saying why.

AGENTS.md already asks for this ("**Warn on unrecognized values** - ... must `logger.warning(...)`, not silently fall through"), and the factory's own test states the intent verbatim: *"Typos must NOT crash the host; warn and default to zenoh."* That test only ever exercised the resolver a typo cannot reach.

## Change

Move the accepted values, the fallback and the report into `strands_robots.mesh._backend_select`, which **both** readers ask, so they cannot diverge again. `session._backend_choice`'s docstring previously claimed it "matches `strands_robots.mesh.transport.factory`" while re-reading the variable independently; that claim is now structural.

Two properties are deliberate and each is pinned:

- **The report is keyed by the offending value, not emitted per call.** The gate is consulted per published message: measured, an unguarded delegation emits 500 warnings for 10 s of 50 Hz telemetry. Keyed by value, a second *distinct* typo still gets its own line.
- **The owner imports nothing from the mesh package.** The factory reaches `session` through `transport.zenoh_transport`, so a resolver either reader imported from the other would close that cycle; and a call-time import inside `session` would raise in the documented no-zenoh case, where `get_session` must return `None` rather than propagate (`test_mesh_session.py::TestSessionLifecycle::test_returns_none_when_zenoh_missing` pins that, and caught exactly that mistake in an earlier revision of this change).

Every value resolves to the same backend as before, reaches the same transport, and produces the same session type. Only the report is new.

## Verification

Measured at `eb5ea531`, base `22316477`. The head is `ab27e629`, which adds only the changelog fragment (no `.py` file differs between the two), so every number below still describes the tree under review.

**Pre-fix split** (this test file copied onto `upstream/main`): **6 failed / 32 passed -> 38 passed**. Five failures are behavioural, quoting what an operator sees, e.g.

```
STRANDS_MESH_BACKEND=iott resolved to 'zenoh' and said nothing. The operator asked for
a transport that does not exist and got the default, indistinguishable from an explicit
STRANDS_MESH_BACKEND=zenoh.
```

The sixth (`test_both_resolvers_share_one_owner`) is the root-cause pin and fails on the symbol this change adds.

**Both new properties are load-bearing.** Dropping the per-value key so it reports every call: 1 failure, `500 reports for 500 publish-path calls`. Restoring the independent re-read in `session`: the same 5 behavioural failures.

**Controls** (pass on both trees, 32 of them): every accepted spelling including `IOT` / `" zenoh "` / `Bridge` is never reported; an unset variable is never reported; all eight `(value -> resolved backend, transport?)` pairs are unchanged; the factory is still never consulted for an unknown value; and the resolver still works with every import raising.

**Gate** - 234 files (all of `tests/mesh`, every test touching the resolvers or `AGENTS.md`, plus the repo-wide docstring / changelog / string guards), run sequentially on both trees:

| | failed | passed | skipped | errors |
|---|---|---|---|---|
| this branch | 21 | 5830 | 33 | 24 |
| `upstream/main` | 21 | 5807 | 33 | 24 |

Failing-ID sets **identical both ways** (branch-only empty, base-only empty). `+23` passed reconciles exactly: 22 new cases in the two touched test files, plus one new parametrized case in `test_discovery_posture_prose.py` that the new module enrols in and passes. The 45 shared failures are pre-existing (`awsiotsdk` / `device_connect_agent_tools` absent locally; both are declared dependencies).

`ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` 24 errors on both trees, none in the changed files.

## Artifact

One capture script per tree: export the mistyped value, run the real `init_mesh`, then 500 publish-gate calls, and record the operator console at the default level. Same package code in both columns; only `strands_robots` differs. The two surrounding console lines are identical in both trees - the fix adds exactly one line, and changes no resolved value.

Operator console for a mistyped STRANDS_MESH_BACKEND

## Scope

`STRANDS_MESH_BACKEND` has no README row while five sibling `STRANDS_MESH_*` variables do. That is one of ~48 undocumented environment variables and a documentation-breadth call rather than part of this contract; the report names the variable and its accepted values, so it stands on its own.


yinsong1986 (approved):
## Summary

Moves the `STRANDS_MESH_BACKEND` vocabulary, fallback and unrecognized-value report into a single owner (`strands_robots.mesh._backend_select.select_backend`) that both resolvers — `session._backend_choice` (the gate that runs first, on every session/publish path) and `transport.factory.get_transport` — now delegate to. This makes the existing warn-on-typo report actually reachable for the input class it names: previously an unknown value resolved to `zenoh` at the session gate, the factory was never consulted, and its warning could not fire. Resolution semantics are unchanged for every value (verified: all eight `(value -> backend, transport?)` pairs pinned as controls); only the report is new, keyed once per distinct offending value so the per-message publish gate does not flood the operator log.

## What's good

- The new module deliberately imports nothing from the mesh package, avoiding the `factory -> zenoh_transport -> session` cycle, and the no-zenoh degradation contract (`get_session` returns `None` rather than propagating `ImportError`) is explicitly pinned by `test_the_resolver_survives_a_world_where_every_import_raises` — `tests/mesh/test_mesh_session.py` still passes (41/41) at the head.
- Both load-bearing properties (single owner, warn-once-per-value) have root-cause pins, and the controls cover the accepted spellings including case/whitespace normalisation and the unset case.
- AGENTS.md > Env Vars gets the generalised lesson, and the changelog fragment (`changelog.d/2542-mesh-backend-typo-report.md`) landed in the follow-up commit rather than an edit to `CHANGELOG.md`, per step 3.
- No behaviour change on any resolved value; the only remaining `_select_backend` reference in the tree is historical prose in a test docstring.

Both touched test files pass at the head (55/55), and `ruff check` / `ruff format --check` are clean on all changed files.

yinsong1986 (approved):
## Summary

Consolidates the `STRANDS_MESH_BACKEND` vocabulary, fallback and unrecognized-value report into a single private owner (`strands_robots.mesh._backend_select.select_backend`) that both resolvers - `session._backend_choice` (the gate that runs first) and `transport.factory.get_transport` - delegate to, making the existing warn-on-typo report reachable for the input class it names. I verified resolution parity empirically on the head tree for all eight documented `(value -> backend, transport?)` pairs plus the unset case; the touched test files (55 tests) and the full `tests/mesh` suite pass locally, with the only error being the pre-existing `awsiotsdk` environment gap the PR description already accounts for. The deleted private `factory._select_backend` has no consumers outside the factory and its own tests, both updated in this diff; the new module imports only stdlib, so the import-cycle and no-zenoh-degradation constraints stated in its docstring hold and are pinned. `_UNKNOWN_WARNED`'s check-then-add is unlocked, but the worst case is one duplicate warning line, which is benign.

## What's good

- The per-value warn-once key is the right call for a resolver sitting on a per-message path, and both load-bearing properties (keying, import-time binding) have dedicated regression tests rather than resting on the description.
- AGENTS.md conventions followed end to end: changelog fragment instead of a direct `CHANGELOG.md` edit, the new Env Vars bullet carries its own pin, warning text is plain ASCII, no host paths.
- Scope discipline: resolution is byte-for-byte unchanged for every value; only the report is new.

PR #2543, opened 2026-08-20
fix(policies/lerobot_local): refuse a norm_tag the checkpoint's stats do not declare

cagataycali:
## Problem

A LeRobot checkpoint that ships only `norm_stats.json` (the MolmoAct2 SO-100/101 family) gets its normalization from `norm_stats.py`, keyed by an embodiment *tag*. `norm_tag` is a documented `config_keys` knob, so the tag is a free-form caller string and a misspelling is the ordinary mistake.

That misspelling was absorbed. `build_norm_stats_processors` answered a requested-but-undeclared tag with the same `(None, None)` it uses for "this checkpoint ships no stats", so the bridge became a full passthrough while usable stats sat in the payload - `observation.state` reaching the policy un-normalized and predicted actions reaching the motors un-unnormalized, the exact failure that module exists to prevent. Measured against the bundled SO-100/101 fixture, a mid-range action collapses from **150.88 degrees to the raw 0.5** the model emitted.

The load report then named the wrong cause. With the pipelines absent it emitted the generic message, so the operator was told `(no policy_postprocessor.json)` and advised to *"Provide the checkpoint's postprocessor"* - a postprocessor this checkpoint was never going to ship, while the one-character remedy appeared nowhere.

The reference implementation refuses the same input: `_RobotStats.validate_tag` in `lerobot.policies.molmoact2.molmoact2_hf_model.modeling_molmoact2` raises `ValueError` naming the allowed tags rather than proceeding un-normalized.

| requested `norm_tag` | pipelines | action for an emitted `0.5` | operator is told |
|---|---|---|---|
| declared tag | built | `150.88` (robot units) | nothing (correct) |
| omitted (single-tag file) | built | `150.88` | nothing (correct) |
| **`'so101'` (a typo)** | **none** | **`0.5`** | **"no policy_postprocessor.json"** |
| no stats file at all | none | `0.5` | "no policy_postprocessor.json" (correct) |

## Change

`build_norm_stats_processors` raises `UnknownNormTagError` when `norm_tag` names a tag the payload does not declare, quoting the tag, the declared tags and the consequence. It subclasses `ValueError`, so a caller narrowing on the loader's existing exception contract still catches it.

Propagating alone would not have reached the operator: `LerobotLocalPolicy._load_processor_bridge` narrows on `ValueError` to treat an absent bridge as benign, so a raise degrades to the same passthrough with its reason at debug. `ProcessorBridge` therefore records the reason, and the load report names it in place of the generic message - the same accurate-cause rule the embodiment-config failure already follows (*"already warned with the accurate cause, so we do not emit the misleading `no policy_postprocessor.json` message for that case"*). `get_info()` reports it too.

Scoped to the caller error. A payload with several tags and no requested one keeps its quiet `(None, None)` verdict - nothing to blame the caller for - and the public `select_norm_tag` keeps its documented resolve-or-`None` contract.

## Sim artifact

MuJoCo SO-101, headless. The same policy output is driven through the real `ProcessorBridge` twice; the only difference is the requested tag. Peak per-frame motion **0.164 (undeclared) vs 20.728 (declared)** - a 127x difference in what reaches the arm, from one character. The consoles below are the real load reports from each tree.

norm_tag not declared

The commanded values are byte-identical in both trees (asserted in the capture): this change alters what is reported, not what is computed.

## Tests

`tests/policies/lerobot_local/test_unknown_norm_tag_is_reported.py` - **5 failed / 8 passed** against `upstream/main`, 13 pass after. Four of the five failures are behavioural (the refusal is not raised, the reason is absent, the report is misleading); the fifth pins the new symbol.

The eight both-tree passes are the controls, and each fails for a specific wrong fix: a declared tag and an omitted tag still build both pipelines; a checkpoint with no stats keeps a `None` reason and the existing message; the ambiguous multi-tag case still degrades quietly; `select_norm_tag` still returns `None`; and an empty tag mapping does not raise. One control parses a declared tag back out of the refusal and applies it, so the remedy the message offers is not a second dead end.

## Verification

Measured at `815c3cca` with `MUJOCO_GL=egl`, lerobot 0.6.2, mujoco 3.12.0.

- 100-file blast radius (every test naming the changed symbols, all of `tests/policies/lerobot_local`, the docs and repo-wide guards): branch **1 failed / 2392 passed / 3 skipped**, base **6 failed / 2387 passed / 3 skipped**, 2396 collected on both. Branch-only failures: none. Base-only: exactly the 5 pre-fix cases.
- The one shared failure, `test_vla_jepa.py::test_vla_jepa_processor_steps_register_via_strands_path`, reproduces identically on a pristine `upstream/main` in isolation - a lerobot step-registration artifact of this environment, untouched by this diff.
- `ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` 24 errors on both trees, none in the changed files.

### DIFF
--- a/changelog.d/2543-unknown-norm-tag-is-reported.md
+++ b/changelog.d/2543-unknown-norm-tag-is-reported.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **policies/lerobot_local**: a `norm_tag` the checkpoint's `norm_stats.json` does not declare is now refused instead of absorbed. It previously shared the `(None, None)` verdict used for "this checkpoint ships no stats", so the processor bridge became a full passthrough while usable stats sat in the payload - state reaching the policy un-normalized and actions reaching the motors un-unnormalized - and the load report blamed a missing `policy_postprocessor.json` the checkpoint was never going to ship. The refusal names the requested tag, the declared tags and the consequence, and that reason reaches the load report in place of the generic message.
--- a/docs/policies/lerobot-local.md
+++ b/docs/policies/lerobot-local.md
@@ -294,6 +294,14 @@ action_unnorm = (clip(action, -1, 1) + 1) * (q99 - q01) / 2 + q01
 When a stats file declares multiple embodiment tags, pass `norm_tag=` to select
 one; a single-tag file is auto-detected.
 
+A `norm_tag` the stats file does not declare is refused rather than absorbed. The
+tag is a free-form string, so a misspelling would otherwise skip normalization
+entirely - state reaching the policy un-normalized and actions reaching the robot
+un-unnormalized, with usable stats sitting unused in the payload. The load report
+names the tag that was asked for and the tags the checkpoint declares, instead of
+blaming a missing `policy_postprocessor.json` the checkpoint was never going to
+ship.
+
 ### Device-pinned checkpoints
 
 A checkpoint trained on GPU bakes `device_processor.device = "cuda"` into its
--- a/strands_robots/policies/lerobot_local/norm_stats.py
+++ b/strands_robots/policies/lerobot_local/norm_stats.py
@@ -54,6 +54,22 @@ OBS_STATE = "observation.state"
 _EPS = 1e-6
 
 
+class UnknownNormTagError(ValueError):
+    """An explicitly requested ``norm_tag`` the checkpoint's stats do not carry.
+
+    Distinct from the two benign reasons a norm-stats fallback yields no
+    pipelines - no recognized stats file, or a multi-tag file the caller left
+    unresolved. Here the stats ARE present and usable and the caller named a tag
+    that is not among them, which is a caller error: the reference implementation
+    (``_RobotStats.validate_tag`` in
+    ``lerobot.policies.molmoact2.molmoact2_hf_model.modeling_molmoact2``) raises
+    for exactly this input rather than proceeding un-normalized.
+
+    Subclasses :class:`ValueError` so a caller narrowing on the loader's existing
+    exception contract keeps catching it.
+    """
+
+
 def _to_array(value: Any) -> np.ndarray | None:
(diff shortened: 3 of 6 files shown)


yinsong1986 (approved):
## Summary

Turns a silently-absorbed caller error into a named refusal: `build_norm_stats_processors` now raises `UnknownNormTagError` (a `ValueError` subclass, so the loader's existing exception contract is preserved) when an explicitly requested `norm_tag` is not among the tags the checkpoint's `norm_stats.json` declares, instead of returning the same `(None, None)` used for "no stats at all" and letting the bridge become a full passthrough. Because `_load_processor_bridge` deliberately narrows on `ValueError`, the bridge records the cause as `inert_reason` and the load report names the misspelled tag and the declared alternatives in place of the misleading `no policy_postprocessor.json` message. The boundary is carefully scoped: multi-tag files with no requested tag, empty tag mappings, checkpoints with no stats, and the public `select_norm_tag` resolve-or-`None` contract all keep their prior behaviour, each pinned by a control test.

## What's good

- The refusal fires before `select_norm_tag` and only when the payload actually declares tags, so every benign absence keeps its quiet verdict - the diff changes exactly the one row of the truth table that was wrong.
- Accurate-cause reporting follows the existing embodiment-config precedent in `_load_processor_bridge` rather than inventing a new pattern, and `_processor_inert_reason` is reset at the top of the method alongside `_embodiment_config_failed`.
- New test module verified locally at head `4910578d`: all 13 pass, the full `tests/policies/lerobot_local/` suite is green (986 passed / 2 skipped), and `ruff check` / `ruff format --check` are clean on the changed files. Controls include parsing a declared tag back out of the refusal message and applying it, so the offered remedy is itself exercised.
- Changelog fragment (`changelog.d/2543-unknown-norm-tag-is-reported.md`) and docs updated in the same PR; new warning text is plain ASCII; no host paths in the test file.

PR #2544, opened 2026-08-20
fix(simulation): refuse to remove a reserved free-view camera name

cagataycali:
## Problem

`create_world()` bakes a `default` camera into the MJCF, and `render`/`get_frame` route
`camera_name in (None, "", "default", "free")` to the free camera by an explicit token check.
Because of that token check, `add_camera(name="default")` is refused as reserved -- a camera
created under the name could never be rendered from.

`remove_camera("default")` had no such check, so the one name the scene ships with could be
deleted but never restored. Measured on the MuJoCo backend:

| step | upstream/main |
|---|---|
| `create_world(); add_robot("so101")` | compiled MJCF cameras `['default']`, `list_cameras()` `['default']` |
| `remove_camera("default")` | **`success`**, "Camera 'default' removed." |
| after removal | compiled MJCF cameras **`[]`**, `list_cameras()` still **`['default']`** |
| `start_recording(cameras=["default"])` | **`error`**: "unknown camera(s) `['default']` ... Available scene cameras: `[]`" |
| `add_camera("default", ...)` to restore | **`error`**: reserved |

Three surfaces disagree after that one call: the compiled model has no camera, `list_cameras()`
and `describe()` still advertise the name, and `get_camera_params("default")` still answers from
the free view. Newton is the same shape (its `remove_camera` also deletes the registry entry the
scene starts with, `list_cameras()` `['default']` -> `[]`).

The remedy the recording refusal offers -- "Add them with `add_camera(...)` before recording" --
is itself refused for this name, so the state is not recoverable through the public surface. A
`Simulation` that has lost it can only be rebuilt.

## Change

Both backends now check `reserved_camera_name_error` in `remove_camera`, the same owner their
`add_camera` already consults, so the rule is applied at both ends of the name's life instead of
only at creation. The refusal carries the reason (the `render` token check) and the actionable
consequence, matching the creation-side message.

`add_camera` and `remove_camera` are inverse operations over one namespace; a name creation
cannot mint is a name removal cannot surrender.

## Tests

`tests/simulation/test_reserved_camera_name_at_removal.py` (new, 15 cases across both backends):

- the removal is refused for every reserved token (`default`, `free`, and the case variants the
  owner accepts), and the refusal names the parameter and the reason
- the compiled MuJoCo camera and the Newton registry entry both survive the refused call
- after a refused removal, `start_recording(cameras=["default"])` still succeeds and records
- a non-reserved camera can still be added and removed (the fix does not widen)
- an unknown name still reports "not found" rather than the reserved message, so the two
  refusals stay distinguishable
- the parameter label in the message is `remove_camera`, not the creation-side context

`tests/simulation/test_reserved_camera_name_at_creation.py` is repaired: it removed the shipped
`default` camera to set up its own premise, and that setup was the defect. It now asserts the
compiled camera survives, which is a stronger statement of what it was written to pin.

Pre-fix split on `upstream/main`: **8 failed / 7 passed**. Post-fix: **15 passed**.

## Verification

- 32-file selection (every test naming `remove_camera` / `list_cameras` /
  `reserved_camera_name_error`, plus the repo-wide changelog / docstring / ASCII guards), run on
  this branch and on a pristine `upstream/main` worktree: **1 failed / 1826 passed on both**, the
  same single pre-existing failure
  (`newton/test_multi_camera.py::TestNamedCameraRegistration::test_bad_position_shape_rejected`,
  a message-text assertion unrelated to this diff). Branch-only failures: none.
- `mypy strands_robots tests tests_integ`: no issues.
- Rollout artifact below produced by one script run under each tree with only `strands_robots`
  differing.

## Scope

`IsaacSimulation.remove_camera` refuses any name not in its `_cameras` registry, and its
`default` is a render-time alias rather than a created prim, so the deletion fixed here is not
reachable there by the same route. Its `add_camera` validates `entity_name_error` but not
`reserved_camera_name_error` (`simulation/isaac/simulation.py:4653`), and the comment above that
line records the same `(None, "", "default", "free")` token routing -- worth a look, but it is a
creation-side question on a backend I could not exercise, so it is left out of this change.

## Artifact

Identical script, both trees, MuJoCo backend headless. Left: the shipped camera is deleted and
the recording of it becomes impossible. Right: the removal is refused, the compiled camera
survives, and a 30-frame dataset with `observation.images.default` records normally.

remove_camera reserved name

### DIFF
--- a/changelog.d/2543-reserved-camera-name-at-removal.md
+++ b/changelog.d/2543-reserved-camera-name-at-removal.md
@@ -1,3 +1,0 @@
-### Fixed
-
-- **simulation**: `remove_camera` refuses a reserved free-view routing token (`default`, `free`, `""`) instead of deleting it. `create_world` both registers `default` and bakes a compiled MJCF camera of that name, while the render path resolves the token to the built-in free view without consulting either, so removing it reported success, permanently deleted a compiled camera the scene ships with, and left `list_cameras` still advertising the name. `start_recording(cameras=["default"])` then refused, and its remedy - `add_camera("default", ...)` - is itself refused as reserved, so nothing could restore it. Both backends now read the rule from the same `reserved_camera_name_error` owner that already guards creation.
--- a/changelog.d/2543-reserved-camera-name-at-removal.md
+++ b/changelog.d/2543-reserved-camera-name-at-removal.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **simulation**: `remove_camera` refuses a reserved free-view routing token (`default`, `free`, `""`) instead of deleting it. `create_world` both registers `default` and bakes a compiled MJCF camera of that name, while the render path resolves the token to the built-in free view without consulting either, so removing it reported success, permanently deleted a compiled camera the scene ships with, and left `list_cameras` still advertising the name. `start_recording(cameras=["default"])` then refused, and its remedy - `add_camera("default", ...)` - is itself refused as reserved, so nothing could restore it. Both backends now read the rule from the same `reserved_camera_name_error` owner that already guards creation.
--- a/changelog.d/2544-reserved-camera-name-at-removal.md
+++ b/changelog.d/2544-reserved-camera-name-at-removal.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- **simulation**: `remove_camera` on the MuJoCo and Newton backends now refuses a reserved
+  free-view camera name (`default`, `free`) on the same shared domain their `add_camera` already
+  consults. Removing the `default` camera `create_world` bakes in used to succeed, after which the
+  name could not be re-added (creation refuses it as reserved) and `start_recording(cameras=["default"])`
+  reported it as an unknown camera, while `list_cameras()` and `describe()` still advertised it.
--- a/strands_robots/simulation/mujoco/simulation.py
+++ b/strands_robots/simulation/mujoco/simulation.py
@@ -4055,12 +4055,24 @@ class MuJoCoSimEngine(
 
         Returns:
             A ``{status, content}`` tool result. ``status`` is ``"error"`` when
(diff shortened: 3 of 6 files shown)


yinsong1986 (approved):
## Summary

Closes the removal side of the free-view token reservation: `remove_camera` on the MuJoCo and Newton backends now consults `reserved_camera_name_error` - the same shared domain their `add_camera` already reads - so the `default`/`free` routing tokens can no longer be surrendered into an unrecoverable state (deleted compiled camera, `list_cameras()` still advertising the name, `start_recording` refusing it, and `add_camera` refusing the restore). The guard precedes the existence test for the same ordering reason as the creation-side check, returns an error dict rather than raising, and requires no new imports (both backends already import the domain for `add_camera`). The repaired creation-side test is a strict strengthening: it previously used the defect (`remove_camera("default")` succeeding) as setup and now asserts the compiled camera survives. Verified the new test module passes on the PR head, re-runs cleanly (the fixed `/tmp` recording root resumes rather than colliding, and it satisfies the recording-root call-site pin), and the only other in-tree reference to removing a reserved token is the historical account in the `reserved_camera_name_error` docstring.

## What's good

- The rule is read from one owner (`reserved_camera_name_error`) at both lifecycle ends rather than a second spelling of the token list, so the router and the refusals cannot drift apart - exactly the shared-domain convention AGENTS.md prescribes.
- Regression tests pin the ordering property (reserved beats "not found"), the survival of the compiled camera and the recordable alias, and the controls (`default_cam` still removable, unknown names still report "not found"), across both token-routing backends.
- Scope discipline: the Isaac backend's differing semantics (no token routing in `get_frame`) are documented as deliberately out of scope rather than speculatively guarded.
- Changelog fragment present under the correct PR number; error strings ASCII; handlers return error dicts, never raise.

PR #2545, opened 2026-08-20
fix(robot): report the sim-only parameters mode='real' cannot honour, and forward the one it can

cagataycali:
## Problem

`Robot()` is one front door onto two destinations, so its signature is the union of what they accept. Two parameters already handle the mismatch, in opposite directions:

- a real-only `cameras` supplied in `mode="sim"` is **refused** with the route that does work;
- a sim-only `backend` supplied in `mode="real"` is **reported** at debug level - the hardware path has no backend, so ignoring it is right, and saying so is what keeps it from looking honoured.

The five remaining parameters the sim path hands to `add_robot` had neither. They are bound by name, so they never reach the hardware constructor's `**kwargs`, and the real branch never read them. Measured on `Robot("so100", mode="real", ...)` with the hardware class doubled:

| supplied in `mode="real"` | reaches the hardware class | reported |
|---|---|---|
| `keyframe="home"` | no | **no** |
| `keyframe=0` | no | **no** |
| `position=[1.0, 2.0, 0.5]` | no | **no** |
| `orientation=[0.0, 1.0, 0.0, 0.0]` | no | **no** |
| `urdf_path="/tmp/arm.xml"` | no | **no** |
| `data_config="so100_dualcam"` | no | **no** |
| `backend="isaac"` | no | yes |

A caller asking a physical arm to come up at its `home` keyframe got a working robot that discarded the request with nothing emitted at any log level.

`data_config` was the one with a consequence beyond silence. `hardware_robot.Robot` **declares** it and carries it into the `policy_config` a policy is built with (`hardware_robot.py`, `policy_config["data_config"] = self.data_config`), so the schema selector its own docstring tells a multi-camera user to "specify explicitly" was dropped between the two, and the policy came up on its default embodiment. The forwarded kwargs were `['cameras', 'robot', 'tool_name']`.

## Change

The contract is the union of both existing behaviours: a parameter the sim path forwards must, in `mode="real"`, either have its value reach the hardware class or be named in a debug record. The mechanism per parameter is decided by its destination, not chosen here.

- **`data_config` is forwarded**, verbatim. `None` is the hardware class's own default, so a caller who names no config keeps today's behaviour - deliberately *not* defaulted to the canonical name the way the sim path does, which would change what every existing hardware caller's policy is built with.
- **`urdf_path`, `position`, `orientation`, `keyframe` are reported**, in one debug record naming each supplied value, using `backend`'s existing mechanism. They describe a pose in a simulated world and a physical arm is already where it is, so there is nowhere to forward them.

Supply is read by `is not None`, not by truthiness: `keyframe=0` is a valid keyframe index and an empty `position`/`orientation` was still supplied, so a truthiness test would drop exactly the values a caller is most likely to be surprised by.

Each of the five parameters now also states its `mode="real"` scope in its own `Args:` entry and in the factory reference table, where `backend` ("Ignored when `mode="real"`") and `cameras` ("Rejected in `mode="sim"`") already did and the other five did not. A reader deciding whether `keyframe` means anything on hardware reads the parameter's entry, not a debug stream they would have to enable first.

Nothing is refused that was not refused before. `cameras` keeps its `ValueError` in the opposite direction, because sim cameras exist and are added another way, so pointing the caller at the route that works is the better answer there.

## Verification

Measured at `2e05babf`, against `upstream/main` at `6af38bff`.

**Pre-fix split** (new tests copied onto `upstream/main`): **13 failed / 4 passed** -> 17 pass. The headline names all five:

```
AssertionError: mode='real' neither carried nor reported these parameters, so a caller
who supplied one got a working robot that discarded it in silence:
{'urdf_path': 'carried=False reported=False', 'data_config': ..., 'position': ...,
 'orientation': ..., 'keyframe': ...}
```

The graded set is read from the sim path's own `add_robot` call rather than listed in the test, so a sixth spawn parameter is held to the contract on arrival.

The 4 tests that pass on both trees are the controls, and each fails for a specific wrong fix:

- `test_supplying_none_of_them_reports_nothing` - a bare hardware build stays quiet.
- `test_data_config_is_not_reported_as_ignored` - fails if `data_config` is reported instead of forwarded.
- `test_the_sibling_backend_report_is_unchanged` - the parameter that already had this mechanism keeps it.
- `test_a_real_only_parameter_is_still_refused_in_sim` - the opposite direction is untouched.

**Gate**: 140-file selection (every test naming `data_config` / `hardware_robot` / `strands_robots.robot` / `mode="real"`, plus the repo-wide docstring, changelog, ASCII, dependency and docs guards).

|---|---|---|---|---|
| this branch | 5 | 5803 | 81 | 5808 |
| `upstream/main` + the new tests | 18 | 5790 | 81 | 5808 |

`18 - 5 = 13` and `5803 - 5790 = 13`, both equal to the 13 pre-fix failures, over the same 5808 collected on each tree. The branch-only failing-id set is empty, and the 13 base-only ids are exactly the new tests with nothing foreign among them. The 5 shared failures are pre-existing `KeyError: 'strands_robots.simulation.policy_runner'` cases in `test_recording_frame_loss_is_not_tolerated.py` and `test_rendering_pacers_survive_a_clock_step.py`, which depend on module import order and reference none of the changed files.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on both this branch and a pristine `upstream/main` worktree, none in the changed files.

## Behaviour

Both columns are the same capture script run against each tree, with the hardware class doubled so the forwarded kwargs and the emitted records are both observable. Every number in the figure is asserted from the captured JSON by the figure script itself.

Robot(mode='real') parameter accounting

## Scope

`mesh` and `peer_id` are read by both branches and are out of scope. Whether the hardware path should also apply the sim path's canonical-name default to `data_config` is a separate question about what an unnamed config means on hardware, and changing it would move what every existing hardware caller's policy is built with, so this forwards the caller's value and leaves the default where the hardware class puts it.


yinsong1986 (approved):
## Summary

Closes the accounting gap for the five sim-only spawn parameters `Robot(mode="real")` used to discard in silence. `data_config` is now forwarded verbatim to `strands_robots.hardware_robot.Robot` (which declares it and carries it into `policy_config`, so a multi-camera schema named at the factory actually selects the embodiment), while `urdf_path`, `position`, `orientation` and `keyframe` are named in a single debug record using the mechanism `backend` already had. Verified on the head: the forwarding cannot collide with `**kwargs` (`data_config` is a bound factory parameter, so it can never also arrive in `kwargs`), `None` forwarded is the hardware class's own default so no existing caller's behaviour moves, `mode="auto"` resolves to `real`/`sim` before the branch so it inherits the fix, and supply is read by `is not None` so `keyframe=0` is reported rather than dropped.

## What's good

- The graded set in the headline test is derived from the sim path's own `add_robot` call via AST rather than a second hand-maintained list, so a sixth spawn parameter is held to the contract on arrival.
- Controls pin the specific wrong fixes: `data_config` must not be reported as ignored, the sibling `backend` report is unchanged, and the opposite-direction `cameras` refusal in sim is untouched.
- Changelog fragment, docstring `Args:` entries and the factory reference table all updated in the same PR; deliberately not applying the sim path's canonical-name default to hardware `data_config` is the right scope call and is documented in both the code comment and the description.
- All 17 new tests pass on the head, the full `tests/test_robot_factory.py` passes (136), and `ruff check` / `ruff format --check` are clean on the changed files; the new strings are plain ASCII with lazy `%`-formatting on the debug record.

PR #2546, opened 2026-08-20
fix(mesh): stop exporting the private local-robot registry from the mesh package

cagataycali:
## Problem

`strands_robots/mesh/__init__.py` listed two private names in `__all__`:

```python
    # Safety
    "log_safety_event",
    # Private (exposed for test patching only)
    "_LOCAL_ROBOTS",
    "_LOCAL_ROBOTS_LOCK",
]
```

`AGENTS.md` states the rule this breaks directly, under *API Consistency*:
"`_`-prefixed names must never appear in `__all__`". Across the package this was
the only violation - 72 modules declare a literal `__all__`, and one had drifted.

**The stated reason does not hold.** `__all__` governs star-imports; it has no
bearing on attribute access, which is how both test files that touch the registry
actually reach it - and one of them already reached the module that defines the
two objects:

| test file | how it reaches the registry | affected by `__all__`? |
| --- | --- | --- |
| `tests/mesh/test_deep_mesh.py:122` | `mesh_core._LOCAL_ROBOTS_LOCK`, via `from strands_robots.mesh import core as mesh_core` | no |
| `tests/mesh/test_mesh.py:71` | `mesh_mod._LOCAL_ROBOTS_LOCK`, via `from strands_robots import mesh as mesh_mod` | no |

**What the entries did govern is measurable.** A star-import skips underscore
names *unless* `__all__` lists them, so listing them was the sole reason
`from strands_robots.mesh import *` bound a mutable registry dict and its lock
into the importer's namespace. Measured three ways on a throwaway module, so the
mechanism is not inferred:

| module's `__all__` | what `from ... import *` binds |
| --- | --- |
| `["y"]` | `y` |
| `["y", "_x"]` | `_x`, `y` |
| absent entirely | `y` |

On `strands_robots.mesh` itself the star-import bound **20** names before and
**18** after, and the two it stopped binding are exactly `_LOCAL_ROBOTS` and
`_LOCAL_ROBOTS_LOCK`.

**The entries were load-bearing for one thing, and it was not test patching.**
They silenced `F401` on an import whose only purpose was to place the names on
the package namespace. Removing only the `__all__` entries makes ruff say so, in
its own remedy text:

```
F401 `strands_robots.mesh.core._LOCAL_ROBOTS` imported but unused;
     consider removing, adding to `__all__`, or using a redundant alias
```

So the import goes with them. The public accessor for the same state,
`get_local_robots()`, was already exported; and with the two names gone,
`test_mesh.py` no longer needed the package-namespace import at all - its only
use of it was those four lines.

## Change

- `strands_robots/mesh/__init__.py` drops both `__all__` entries and the import
  that fed them. The module docstring now says where the registry lives:
  `get_local_robots()` for a snapshot, `strands_robots.mesh.core` for code that
  needs to mutate it.
- `tests/mesh/test_mesh.py` reaches the registry in the module that defines it -
  the spelling `test_deep_mesh.py` already used. `get_local_robots()` cannot
  serve this fixture, which has to *clear* the registry, and that returns a copy;
  reaching the private name is right, reaching it through the package was not.
- `tests/test_all_exports_are_statically_defined.py` grades the second promise
  `__all__` makes about a name. It had graded only the first.

## Why the existing guard did not see it

`__all__` makes two promises about each name it lists: that it is **defined**,
and that it is **public**. The guard covered the first, so it reported this tree
clean with the offender present:

```
$ git stash          # unfixed tree, shipped guard
$ pytest tests/test_all_exports_are_statically_defined.py -q
15 passed
```

Both halves now run over the same population, so a module cannot satisfy one and
quietly fail the other. A leading underscore is the whole test; dunders are
excluded, because a module re-exporting `__version__` is naming a documented
attribute rather than reaching into a private one.

## Tests

Copied both updated test files onto `upstream/main` and ran them:

**1 failed / 65 passed pre-fix; 66 pass after.**

The single failure is the contract, and it names both offenders:

```
AssertionError: __all__ lists names that are not public. A star-import skips
underscore names unless __all__ overrides that, so listing one publishes it;
and an entry added to silence F401 on an import that exists only to place a
name on the package namespace should drop the import too. Export a public
accessor instead, or reach the private name in the module that defines it:
{'mesh/__init__.py': ['_LOCAL_ROBOTS', '_LOCAL_ROBOTS_LOCK']}
```

The 65 that pass on both trees are the controls, and each fails for a specific
wrong fix:

- **`tests/mesh/test_mesh.py` - all 46 - passes on `upstream/main` too.** The
  repair is spelling-only: `mesh.core._LOCAL_ROBOTS` resolves in both trees, so
  the fixture was not weakened to accommodate the removal.
- `test_a_public_only_all_is_not_reported` fails if the underscore test is
  inverted or the scan starts reporting everything.
- `test_a_dunder_export_is_not_reported` fails if `__version__` gets caught.
- `test_a_private_name_is_reported_even_when_it_is_defined` asserts both halves
  on one source: definedness is satisfied and only the public check has anything
  to say. That is the gap being closed, stated as a test.
- `test_a_private_export_is_reported` is the planted-defect direction, so a clean
  tree means the tree is clean rather than that the detector accepts everything.
- The definedness half and its six detector tests are untouched and still pass.

## Evidence

Each column is read from its own tree by one script, so the surface and the
verdict are that tree's own answers rather than a transcription. `mesh.core`
resolving both objects in both columns is the control: nothing moved, the package
just stopped re-publishing them.

__all__ public-export evidence

## Verification

Measured at `2551d807`, against `upstream/main` at `4fdb6eb8`.

**312-file blast radius** - every test that walks the test tree, grades exports,
reads `AGENTS.md`, or touches the mesh package namespace, plus all of
`tests/mesh` and the repo-wide guards:

| | failed | passed | skipped | errors | collected |
| this branch | 21 | 10824 | 85 | 24 | 10930 |
| `upstream/main` + both updated test files | 22 | 10823 | 85 | 24 | 10930 |

`22 - 21 == 10824 - 10823 == 1`, which is the pre-fix failure count. Branch-only
failures: none. Base-only: exactly
`TestNoExportIsPrivate::test_no_module_exports_a_private_name`. The 21 shared
failures and all 24 errors are `tests/mesh/test_iot_*` needing `awsiotsdk`, which
is a declared dependency and installed in CI; none is in a file this touches.

- `ruff check` / `ruff format --check` clean over `strands_robots tests
  tests_integ`.
- `mypy strands_robots tests tests_integ`: 24 errors on the branch, 24 on a
  pristine `upstream/main` worktree, zero of them branch-only.
- No behaviour changes in the package beyond the removed re-export: the star
  import binds 18 names instead of 20, `strands_robots.mesh.get_local_robots()`
  is unchanged, and `strands_robots.mesh.core._LOCAL_ROBOTS` is where it always
  was.


yinsong1986 (approved):
## Summary

Removes the two private names (`_LOCAL_ROBOTS`, `_LOCAL_ROBOTS_LOCK`) from `strands_robots/mesh/__init__.py`'s `__all__` along with the import that existed only to place them on the package namespace, closing the one violation of the AGENTS.md API Consistency rule ("`_`-prefixed names must never appear in `__all__`") in the tree. The only consumer that reached the registry through the package namespace, `tests/mesh/test_mesh.py`, is repaired to reach `strands_robots.mesh.core` directly - the same spelling `test_deep_mesh.py` already used - so the fixture's clear-the-registry behaviour is unchanged (`get_local_robots()` returns a copy and could not serve it). A new `TestNoExportIsPrivate` half is added to `tests/test_all_exports_are_statically_defined.py`, grading the public-surface promise over the same 72-module population as the existing definedness half, with planted-defect and control tests (dunder exports and public-only `__all__` excluded correctly). Verified against the tree: no other module declares a private name in a literal `__all__`, and nothing else accesses `strands_robots.mesh._LOCAL_ROBOTS*`, so the removal breaks no in-tree consumer; the removed names are private, so this is not a compatibility door. Changelog fragment present per AGENTS.md step 3.

## What's good

- The guard closes the gap that let this drift ship, rather than only fixing the instance - and its error message names the F401-silencing mechanism that produces the next one.
- The test repair is spelling-only (`mesh.core._LOCAL_ROBOTS` resolves in both trees), so the fixture was not weakened to accommodate the removal.
- AGENTS.md update, changelog fragment, and pre/post-fix evidence all in one PR, per repo conventions.

PR #2547, opened 2026-08-21
fix(simulation): hold a patch op's claimed name to the shared entity-name domain

cagataycali:
## Problem

`patch_scene_mjcf` is a published action, and three of its ops create an entity in the live `MjSpec`: `add_body`, `add_geom` and `add_site`. Each checked only that the supplied name was *truthy*, while `add_object` / `add_camera` / `add_robot` route the same question through the shared `entity_name_error` domain. That domain exists because a name which cannot address the entity it creates leaves something unreachable through the very API that made it, and its docstring measures the values on MuJoCo. The patch door is a fourth creation site on the same spec, so those values were accepted there.

Measured on one `create_world`, MuJoCo 3.12.0:

| op | name | before | after |
|---|---|---|---|
| `add_body` | `"a\x00b"` | **success**, body compiled as `a` | error, names the NUL |
| `add_geom` | `"a\x00b"` | **success**, geom compiled as `a` | error |
| `add_site` | `"a\x00b"` | **success**, site compiled as `a` | error |
| `add_geom` | `""` | **success**, geom compiled **unnamed** | error |
| `add_body` / `add_site` | `""` | error (`requires 'name'`) | error, now says why |
| `add_body` | `7`, `["x"]` | error, but a C++ overload table naming neither the op nor the field | error, names the op, the field and the value |
| `add_object` | any of the above | error, with the reason | unchanged |

Two consequences follow, and neither is visible to the caller:

- **Two names, one entity.** MuJoCo compares names only up to the first NUL. `add_body(name="a\x00b")` reported success, `list_bodies()` returned `['world', 'a']`, and a later op that genuinely asked for `a` was refused `Error: repeated name 'a' in body` - blaming a collision with a name the caller never used. So the door silently reserved a name nobody requested and then rejected the request for it.
- **An entity no name-based surface can reach.** `add_geom(name="")` compiled the geom with `None` for a name, and `set_geom_properties(geom_name="")` answered `Geom 'None' not found. Available geoms: ['ground']`. The caller claimed a name and got something the only name-addressed geom surface cannot touch.

## Change

`_apply_patch_op` routes the name of the three ops that **claim** one through `entity_name_error`, so the refusal names the op, the parameter, the value and the reason, and arrives before the spec is written.

Scope follows the domain's own reasoning rather than a judgement call:

- The three ops that **look a name up** (`set_body_pos`, `set_body_quat`, `delete_body`) are unchanged. A name that addresses nothing is honestly *absent* at a lookup, which is what their own `body '<name>' not found` message already reports.
- **Omitting** `add_geom`'s `name` is unchanged and still produces an unnamed geom, which is legal MJCF. Only a *supplied* name is a claim, so `""` is refused and an absent key is not.
- An absent `name` key keeps its exact `<op> requires 'name'` message, so the four existing tests that pin it are untouched.

`entity_name_error`'s docstring and `patch_scene_mjcf`'s now state that the patch ops are a fourth door onto the same domain, beside the numeric-domain paragraph they parallel.

## Tests

Added to `tests/simulation/test_entity_name_domain_at_creation.py`, the file that already owns this contract for the three methods and whose docstring says the domain lives in one shared helper so the creation sites cannot drift.

Against `upstream/main`: **59 failed / 296 passed** -> 355 pass. 56 of the failures are behavioural, e.g.

```
test_a_nul_name_does_not_reserve_its_truncation
  AssertionError: assert 'success' == 'error'

test_an_empty_geom_name_compiles_no_unnamed_geom
  AssertionError: assert None not in ['ground', None]
```

The remaining 3 are the root-cause guard: `TestEveryClaimingOpRoutesThroughTheDomain` derives the op set from `_PATCH_OP_KEYS`, so a seventh op cannot arrive unclassified - it has to be a claiming op that consults the domain or a lookup op that does not.

The 296 that pass on both trees are the controls, and each fails for a specific wrong fix:

- omitting `add_geom`'s name still yields an unnamed geom (fails if the check is made unconditional);
- `add_body` / `add_site` keep their `requires 'name'` message for an absent key (fails if the two reports are collapsed);
- all three lookup ops still report absence (fails if the domain is applied to them);
- every name in `_GOOD_NAMES` - namespaced, dotted, spaced, non-ASCII, 200 chars - still round-trips through all three ops and `get_body_state`;
- the patch door and `add_object` produce the same sentence for the same value.

## Verification

Measured at `31994f92`, against `upstream/main` `40514971`.

- `mypy strands_robots tests tests_integ`: 24 == 24 against a pristine `upstream/main` worktree, none in the changed files.
- 424-file blast radius (every test naming the changed symbols, all of `tests/simulation`, and the repo-wide docstring / string / changelog guards), same selection on both trees:

|---|---|---|---|---|
| this branch | 5 | 13524 | 81 | 13610 |
| `upstream/main` (with the new test file copied in) | 64 | 13465 | 81 | 13610 |

  Branch-only failures: **none**. The 59 base-only ids are exactly the pre-fix failures, all in the new test file. The 5 shared failures (4 in `test_recording_frame_loss_is_not_tolerated.py`, 1 in `test_rendering_pacers_survive_a_clock_step.py[forward_30s]`) are pre-existing and import-order dependent within a selection this wide - all 29 tests in those two files pass in isolation on both trees.

## Artifact

An agent places three markers with three separate `patch_scene_mjcf` calls; the second name has picked up a NUL from an untrimmed upstream string. The yellow pad marks where the third marker, `"probe"`, was asked for. Same script, same scene, MuJoCo headless (`MUJOCO_GL=egl`) - only `strands_robots` differs.

Before, the ask naming `"probe\x00tmp"` reported success and compiled the body as `probe` **0.28 m** from where `probe` was requested, so the ask that did name `probe` was refused for colliding with it. After, the NUL is refused where it is written and `probe` lands on the pad (**0.00 m**).

The green marker (1682 px) and the pad (2121 px) are pixel-identical with the same centroids in both panels, so the red marker's 167 px move is the only difference.

patch-op entity-name domain, before and after

### DIFF
--- a/changelog.d/2547-patch-op-entity-name-domain.md
+++ b/changelog.d/2547-patch-op-entity-name-domain.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **simulation/mujoco**: the `patch_scene_mjcf` ops that claim a name (`add_body`, `add_geom`, `add_site`) now hold it to the same `entity_name_error` domain `add_object` / `add_camera` / `add_robot` apply. A name carrying a NUL used to report success and compile the entity under the truncated name, so a later op that genuinely asked for that shorter name was refused as a repeated name the caller never used; an empty `add_geom` name compiled an unnamed geom that `set_geom_properties(geom_name=...)` could not reach. Omitting `add_geom`'s name still produces an unnamed geom, and the ops that look a name up are unchanged.
--- a/strands_robots/simulation/mujoco/scene_ops.py
+++ b/strands_robots/simulation/mujoco/scene_ops.py
@@ -43,7 +43,7 @@ from typing import Any
 from strands_robots.simulation.models import SimCamera, SimObject, SimRobot, SimWorld
 from strands_robots.simulation.mujoco.backend import _ensure_mujoco, filter_mujoco_attach_noise, mj_name_to_id
 from strands_robots.simulation.mujoco.spec_builder import SpecBuilder
-from strands_robots.utils import coerce_rgba, finite_vector_error, pose_vector_error
+from strands_robots.utils import coerce_rgba, entity_name_error, finite_vector_error, pose_vector_error
 
 logger = logging.getLogger(__name__)
 
@@ -2302,6 +2302,12 @@ def _apply_patch_op(spec: Any, op: dict[str, Any], new_bodies: dict[str, Any]) -
     MuJoCo compile errors surface on the enclosing ``recompile`` call.
     ``new_bodies`` is a batch-local cache of body handles added earlier
     in the same patch (see ``_find_body`` for why this is needed).
+
+    An op that CLAIMS a name holds it to the shared
+    :func:`~strands_robots.utils.entity_name_error` domain, so this door cannot
+    create an entity a name-based surface is unable to address. An op that
+    LOOKS a name UP does not: there a name addressing nothing is honestly
+    absent, which is what its own "not found" message already reports.
     """
     if not isinstance(op, dict):
         raise ValueError(f"each op must be a dict, got {type(op).__name__}")
@@ -2321,8 +2327,10 @@ def _apply_patch_op(spec: Any, op: dict[str, Any], new_bodies: dict[str, Any]) -
     if kind == "add_body":
         parent = op.get("parent", "world")
         name = op.get("name")
-        if not name:
+        if name is None:
             raise ValueError("add_body requires 'name'")
+        if (name_err := entity_name_error("add_body", "name", name)) is not None:
+            raise ValueError(name_err)
         pos = op.get("pos", [0.0, 0.0, 0.0])
         quat = op.get("quat", [1.0, 0.0, 0.0, 0.0])
         parent_body = _find_body(spec, parent, new_bodies)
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Closes the fourth creation door onto the MuJoCo spec: the `patch_scene_mjcf` ops that claim a name (`add_body` / `add_geom` / `add_site`) now route through the shared `entity_name_error` domain that `add_object` / `add_camera` / `add_robot` already apply, so a NUL-carrying name is refused instead of silently compiling under its truncation (and then blaming a later legitimate caller for a "repeated name"), and an empty `add_geom` name is refused instead of compiling a geom that `set_geom_properties` can never address. The scope split is principled and pinned: supplied names are claims and are checked, omitted `add_geom` names stay legal unnamed MJCF, and the three lookup ops keep their own "not found" reporting.

## What's good

- Reuses the existing shared guard rather than adding a local copy — exactly the AGENTS.md convention-11 shape (a second-plus caller lifts into the shared domain; this is the fourth).
- Regression pins are real: with only the test file applied to `main`, 59/355 fail (56 behavioural + 3 root-cause structural); on the branch all 355 pass (verified locally, headless MuJoCo). The `TestEveryClaimingOpRoutesThroughTheDomain` class derives the op set from `_PATCH_OP_KEYS`, so a future op cannot arrive unclassified.
- Controls guard against every plausible wrong fix (unconditional check, collapsed messages, domain applied to lookups), and the refused-batch test pins that rollback leaves no partial spec writes.
- Error-dict contract preserved (`ValueError` surfaces through `patch_scene_mjcf`'s structured result), changelog fragment used, docstrings updated on both the domain and the published action, cross-reference roles kept on one line, no NUL bytes in source (escape sequences only).

Issue #389, opened 2026-06-13
[v0.4.0-D] hardware_robot + robot_factory: lerobot 0.5.x compat, walker robustness, exception-narrowing, thor deps


cagataycali:
> Part of the **v0.4.0 issue-consolidation** effort (roadmap #94). This umbrella replaces the individual issues listed below — each is being closed with a pointer here. One surgical PR will close this umbrella.

**Component(s):** `strands_robots/hardware_robot.py`
**Consolidates 16 issues:** #275, #281, #283, #285, #287, #288, #289, #290, #291, #292, #293, #294, #297, #298, #377, #378

---

## Sub-tasks

- [ ] **#275** — [hardware_robot] real-mode SO-101/SO-100 broken on lerobot 0.5.1: SOFollowerConfig has no `id` field; should use SOFollowerRobotConfig
  - `Robot("so101", mode="real", port=...)` (and any other `so100`/`so101` real-mode call) fails on `lerobot==0.5.1` because `strands_robots/hardware_robot.py::_create_minimal_config` constructs `SOFollowerConfig(id=...)`, but `SOFollowerConfig` in lerobot 0.5.1 has **no `id` field** — only its registered subclass `SOFollowerRobotConfig` does.
- [ ] **#281** — [hardware_robot] narrow `except Exception` at _create_minimal_config:393 to (TypeError, ValueError) — AGENTS.md exception-narrowing convention
  - Follow-up from PR #276 R5 review feedback.
- [ ] **#283** — [hardware_robot] `forwardable` allowlist undermines 'zero strands_robots changes for new lerobot robots' claim — accept dataclass-declared kwargs without allowlist gating
  - Follow-up from PR #276 R5 review feedback (post-budget concern).
- [ ] **#285** — test(robot_factory): hard-coded lerobot built-in robot list creates upstream-coupled flake risk on `pip install -U lerobot`
  - Follow-up from PR #276 R5 review feedback.
- [ ] **#287** — [hardware_robot] split logger.warning into debug (lerobot fully missing) vs warning (partial install) for noise discipline
  - Follow-up from PR #276 R8 review thread (`PRRT_kwDORUMiZs6F75fG`), filed to keep the merge-ready PR focused.
- [ ] **#288** — [hardware_robot] document or trim _FORWARDABLE_KWARGS post-R5 — only role left is cross-robot polymorphism silent-drop allowlist
[...]

Issue #388, opened 2026-06-13
[v0.4.0-C] mesh(iot): v0.4.1 hardening + operator-facing docs bundle


cagataycali:
> Part of the **v0.4.0 issue-consolidation** effort (roadmap #94). This umbrella replaces the individual issues listed below — each is being closed with a pointer here. One surgical PR will close this umbrella.

**Component(s):** `strands_robots/mesh/iot/*, README env matrix`
**Consolidates 5 issues:** #254, #311, #312, #316, #368

---

## Sub-tasks

- [ ] **#254** — mesh(iot): IoT comment accuracy, multi-pin rotation test, #251 cross-ref, Subscribe/Receive doc (from #228 R5)
  - From PR #228 R5 review feedback (2026-05-29). Four small documentation/test improvements grouped as one coherent diff.
- [ ] **#311** — [mesh/iot] _ensure_ca break-glass marker has create-then-chmod TOCTOU window
  - Discovered during PR #228 R10 review (thread `provision.py:855`).
- [ ] **#312** — [mesh/iot] _ensure_ca symlink-rejection error message is generic; should mirror verify_ca_pin's explicit branch
  - Discovered during PR #228 R10 review (thread `provision.py:744`).
- [ ] **#316** — mesh(iot): v0.4.1 operator-facing docs bundle — cert-rotation cadence, thing-name `:` migration, teardown_thing residue (from #228 R10)
  - PR #228 (8/9 of the #195 split) merged the AWS IoT provisioning hardening with CA pinning, strict thing-name regex, scoped `iot:Receive`, per-recv timeout, and 60s default presigned-URL TTL. Round 10 review (yinsong1986, 2026-06-03) flagged three operator-facing **documentation gaps** that were explicitly tracked as `[FOLLOW-UP]` rather than blocki
- [ ] **#368** — docs(mesh): add STRANDS_MESH_BRIDGE_TOPICS + _PREFIX rows to README env-var matrix (deferred from #244/#354)
  - Deferred from #244 / PR #354 (the `_resolve_bridge_prefix_filter` regression pin).

---

## Acceptance criteria

- [ ] Every sub-task addressed with a surgical, default-preserving fix
- [ ] Regression test per fix (fails pre-fix, passes post-fix)
- [ ] `ruff check . && ruff format --check . && mypy strands_robots` clean
- [ ] Targeted pytest suite green; no regressions
- [ ] PR body maps each sub-task → fix → why; closes this umbrella

[...]

PR #276, opened 2026-05-30
fix(hardware_robot): real-mode factory uses lerobot ChoiceRegistry discovery (closes #275)

cagataycali:
Closes #275.

## TL;DR

Replace the hand-rolled `config_mapping` in `HardwareRobot._create_minimal_config` with discovery via `lerobot.robots.config.RobotConfig.get_choice_class()` — the same lookup `make_robot_from_config` performs internally. Single change closes both real-mode regressions in the issue **and** unblocks every robot in lerobot's registry (not just the 7 in the old hand-rolled list).

```python
# before:  ❌  ValueError: SOFollowerConfig.__init__() got an unexpected keyword argument 'id'
# after :  ✅  Robot loaded, mesh up, joints streaming
Robot("so101", mode="real", port="/dev/ttyACM0", id="left_arm")

# before:  ❌  ValueError: Unsupported robot type: unitree_g1
# after :  ✅  reaches lerobot's UnitreeG1 driver (which then probes DDS, etc.)
Robot("unitree_g1", mode="real", robot_ip="192.168.123.164",
      kp=[100.0]*29, kd=[2.0]*29, default_positions=[0.0]*29,
      is_simulation=False)
```

### 1. SO-100 / SO-101 real-mode broken on lerobot >= 0.5.0

`SOFollowerConfig` in lerobot 0.5.x is the bare leaf-fields struct (no `id`); the registered constructible class is `SOFollowerRobotConfig`, which inherits both `RobotConfig` and `SOFollowerConfig`. The old `config_mapping` pointed at the wrong one, so:

```
Failed to create SOFollowerConfig for robot type 'so101_follower':
  SOFollowerConfig.__init__() got an unexpected keyword argument 'id'.
```

Discovery picks the registered subclass automatically — same lookup `make_robot_from_config` already uses.

### 2. `unitree_g1` advertised but unreachable

The registry advertises `has_real=True` for `unitree_g1`, but it had no entry in the hand-rolled `config_mapping`. Users got a misleading `ValueError: Unsupported robot type: unitree_g1` despite the registry, lerobot, and the robot itself all supporting real mode. Discovery sees it for free because lerobot registers it via `@RobotConfig.register_subclass("unitree_g1")`.

### 3. Cleanup `AttributeError` on init failure

[...]

PR #228, opened 2026-05-25
mesh(iot): AWS IoT provisioning hardening (CA pin + thing-name regex + scoped policy) (8/9 of #195 split)

cagataycali:
Part 8 / 9 of the split of #195 — tracked by #219.

**Drafted** until PR-2 (#223) and PR-4 (#221) land.

The IoT path is an alternative wire transport with its own threat model: cloud-side certs, IoT policy wildcards, CA-substitution MITM, camera URL capture. This PR closes those vectors.

## What's in this PR (after R2 scope correction)

- `strands_robots/mesh/iot/provision.py` (+350/-10) — **CA pinning** (defeats CA-substitution MITM), **strict thing-name regex** (anchored, not just `match`), per-recv timeout bound, **IoT policy scope tightening** (no `Resource: '*'` wildcards, per-thing topic prefixes only).
- `strands_robots/mesh/iot/camera_offload.py` — short default presigned-URL TTL (60s, was 3600s), 1-hour cap, kwarg-vs-env precedence fixed for explicit `presign_ttl=0`. Bucket-ownership threat model documented.
- `strands_robots/mesh/iot/shadow.py` (+2/-2), `iot/__init__.py` (+5/-5).
- 6 test files (~850 LOC).

## Reviewer focus

- **CA-substitution MITM defence** — CA pinning verifies the AWS IoT chain is the one the operator pinned, not whatever the broker presents.
- **Thing-name regex anchored** — `^[a-zA-Z0-9_-]{1,128}$` (strict subset of AWS's charset; `-`, `_` only as separators -- no `:` due to NTFS/classic-Mac semantics), applied symmetrically across `provision_robot`, `provision_operator`, `teardown_thing`.
- **Per-recv timeout** — prevents a malicious broker from holding a connection open indefinitely.
- **IoT policy scope** — explicit per-thing prefix in `Resource`, never `*`.
- **Camera presigned-URL TTL** — default cut from 1 hour to 60 seconds; 1-hour ceiling; explicit `presign_ttl=0` is now treated as an operator value (clamped to 1) rather than silently falling back to the env default.

## Carries review fixes from #195

iot-CA (CA-substitution MITM defence), iot-policy-scope (no wildcards). The R22 camera-side privacy kill-switch and S3 PutObject ACL hardenings originally claimed in this slice were not actually implemented and are deferred to #249 (see R2 changelog below).

## Stacking note

[...]

PR #2550, opened 2026-08-21
test: name every test case for its behaviour, not for the release that added it

cagataycali:
## Problem

Two test classes were named for the release and review round that produced them rather than for the behaviour they verify:

| | |
|---|---|
| `tests/test_robot_factory.py:1576` | `class TestHardwareConfigV040Followups` - *"v0.4.0 hardware_robot follow-up bundle (#389) - PR #276 review trail."* |
| `tests/mesh/test_iot_ca_pin.py:362` | `class TestEnsureCaV040Followups` - *"v0.4.0 mesh-iot follow-ups (#388) from the #228 review trail."* |

A test's name is the first thing a maintainer reads, so finding the test that covers a behaviour means opening every plausible file. And once the release is long past the name actively misleads: `V040` reads as historical, which is an invitation to skip it.

This is not the tree's habit - it is the exception. Measured over both test trees at the merge base:

| | total | named for a release / review round |
|---|---|---|
| test classes | 3294 | **2** |
| test functions | 17508 | **0** |

## Change

**The names.** The hardware class was three unrelated review items (a dropped-kwarg debug signal, an exception-clause narrowness check, a packaging-extra assertion), so no single behaviour name covers it - a bundle named for a review round is usually a bundle. It is split into one class per behaviour:

- `TestForwardableKwargDropIsReported`
- `TestThirdPartyPluginGuardIsNarrow`
- `TestLerobotExtraShipsAnAarch64VideoDecoder`

The CA class does share one subject - `_ensure_ca`'s own filesystem writes - so it is renamed to `TestEnsureCaFilesystemDiscipline`. No assertion changes in either file; the `#NNN` issue references stay in the per-test docstrings, which is where a reader wants provenance.

**The guard.** `tests/test_test_case_names_describe_behaviour.py` refuses two tokens in a `class Test*` or `def test_*` name:

- a **release token** - three or more digits after a `v` (`V040` is `0.4.0` with the dots dropped);
- a **review-round token** - `Followup` / `Followups`.

Both offenders carried both tokens, so both rules are load-bearing on real data rather than only on planted input.

**What is deliberately still accepted:** a one- or two-digit version token, because there it names a *data format* that genuinely is the behaviour under test. The nine such names in the tree are pinned as controls, e.g. `test_load_v3_parses_every_field` (the `.spz` v3 layout), `test_a_v21_root_is_refused_and_pointed_at_lerobots_converter` (a LeRobot v2.1 dataset), `test_current_transform_version_is_v4`. That boundary is what keeps the rule from becoming "no digits in test names".

`AGENTS.md` gains the convention under `### Testing`, in that section's existing style of ending a bullet with the test that enforces it.

## Verification

Measured at `8b234dfb`, against `upstream/main` at `11acaa7e`.

**Pre-fix split.** Copying only the new guard into an `upstream/main` worktree: **1 failed / 14 passed**, the failure naming both offenders and both tokens:

```
test case(s) named for their provenance rather than the behaviour they verify:
  tests/mesh/test_iot_ca_pin.py:362 TestEnsureCaV040Followups (release token 'V040', review-round token 'Followups')
  tests/test_robot_factory.py:1576 TestHardwareConfigV040Followups (release token 'V040', review-round token 'Followups')
Name the behaviour and keep the issue reference in the docstring.
```

15 passed after. The 14 both-tree passes are the controls: 7 format-version names that must keep working, 6 planted provenance-named cases proving each rule detects (so a clean sweep is a real result, not an inert grader), and a scope test that the scan reaches both `tests/` and `tests_integ/`. A `>= 15000` floor on the number of names scanned means a discovery that goes blind fails loudly instead of reporting clean.

**The renames change no behaviour.** `tests/test_robot_factory.py` + `tests/mesh/test_iot_ca_pin.py`: **162 passed** before and after.

**Blast radius.** The 154 test files that walk a test tree, read `AGENTS.md` or grade names, plus both edited files and the repo-wide guards, run against both trees:

|---|---|---|---|---|
| branch | 8 | 6664 | 55 | 6727 |
| `upstream/main` + the guard | 9 | 6663 | 55 | 6727 |

Same collected total, and the failed and passed deltas are both 1 - the one pre-fix failure. The branch-only failure set is empty; the base-only set is exactly that failure. The 8 shared failures are `tests/mesh/test_iot_*` needing `awsiotsdk`, which is declared and installed in CI.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on both trees with none in the changed files.

**No visual artifact:** this touches no policy, simulation behaviour, rendering, recording or robot asset - it renames test cases and adds a naming guard.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1388,6 +1388,7 @@ Corrections from code review that apply to all future contributions:
 - **Round-trip tests for recording** - Any recording feature needs: start -> write -> stop -> reopen -> assert non-empty. Schema-only tests miss silent data loss.
 - **Pin regression tests for reviewed fixes** - Every review fix gets a test that fails on pre-fix code. Otherwise the next refactor silently reintroduces the bug.
 - **No host paths in test files** - Never commit `/Users/<name>/` or `/home/<name>/` paths. CI test `test_no_host_paths.py` enforces this.
+- **Name a test for its behaviour, not for its provenance** - a test class or function must not carry the release (`TestHardwareConfigV040Followups`) or review round (`...Followups`) that produced it. The name is the first thing a maintainer reads, so it has to say what is verified; and a name tied to a shipped release reads as historical, which invites skipping it. A bundle named for a review round is usually a bundle of unrelated checks - split it into one behaviour per class rather than inventing a name that covers all of them. Provenance belongs in the docstring, where the `#NNN` reference stays useful. A version token of one or two digits is fine: it names a *data format* under test (`test_load_v3_parses_every_field`), not a release. `Pinned by` `tests/test_test_case_names_describe_behaviour.py`.
 
 ### Performance
 - **Don't create executors in hot loops** - Reuse a single `ThreadPoolExecutor` instance instead of creating one per call at 50Hz.
--- a/changelog.d/2550-test-names-describe-behaviour.md
+++ b/changelog.d/2550-test-names-describe-behaviour.md
@@ -1,0 +1,18 @@
+### Changed
+
+- **tests**: no test class or function is named for the release or review round that
+  produced it. `TestHardwareConfigV040Followups` and `TestEnsureCaV040Followups` named
+  a shipped release and a review trail rather than the behaviour they verify, which is
+  what a maintainer scanning for coverage of a behaviour actually reads - and `V040`
+  reads as historical once the release is past, which invites skipping the class. The
+  hardware class was three unrelated review items, so no single behaviour name covers
+  it: it is split into `TestForwardableKwargDropIsReported`,
+  `TestThirdPartyPluginGuardIsNarrow` and `TestLerobotExtraShipsAnAarch64VideoDecoder`.
+  The CA class shares one subject and is renamed to
+  `TestEnsureCaFilesystemDiscipline`. No assertion changed, and the `#NNN` issue
+  references stay in the per-test docstrings. A repo-wide guard now refuses a release
+  token (three or more digits after a `v`) and a review-round token
+  (`Followup`/`Followups`) in a test-case name, while deliberately accepting a one- or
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Renames two test classes that were named for the release/review round that produced them (`TestHardwareConfigV040Followups`, `TestEnsureCaV040Followups`) to behaviour-describing names, splitting the hardware bundle into three single-behaviour classes, and adds a repo-wide guard (`tests/test_test_case_names_describe_behaviour.py`) that refuses a release token (`[Vv]\d{3,}`) or a review-round token (`Followup(s)`) in any `class Test*` / `def test_*` name across both test trees. No assertion changes; `#NNN` provenance stays in the per-test docstrings. Verified at `8b234dfb`: both edited test files pass unchanged (161 passed, 1 skipped), the new guard passes 15/15, no stale references to the old class names remain anywhere in the tree, and `ruff check` / `ruff format --check` are clean on all three touched test files.

## What's good

- The guard is self-validating in both directions: planted-offender parametrizations prove each token rule detects, a `>= 15000` scanned-names floor catches a blind discovery, and a tree-reach test pins that `tests_integ/` stays in scope - so a clean sweep is a real result.
- The format-version allowance (one/two-digit `v` tokens) is pinned with real names from the tree, which keeps the rule from drifting into "no digits in test names".
- AGENTS.md convention added in the existing `### Testing` section style, with the enforcing test named; changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention rather than editing `CHANGELOG.md`.
- Splitting the three-unrelated-checks bundle into one class per behaviour rather than inventing an umbrella name is the right call, and the class docstrings now describe the contract under test.

Issue #2551, opened 2026-08-21
test(simulation): the total-lookup guard scans three directories non-recursively, and two same-shape lookups sit outside it - no live defect, and nothing pins why


cagataycali:
## No live defect

Stating that first so this is not read as a bug report. Every agent-callable route into the two lookups below refuses an unhashable name before reaching them. What is missing is the *pin* on that fact.

## What the guard covers

`tests/simulation/test_unhashable_entity_name_is_reported.py::test_every_backend_lookup_uses_the_shared_rule` is the check that keeps a registry lookup added later inside the total-lookup rule. Its scope, read from `_backend_dirs()` and the scan loop:

```python
return [simulation_dir / "mujoco", simulation_dir / "newton", simulation_dir / "isaac"]
...
for path in sorted(directory.glob("*.py")):
```

Three directories, `glob` rather than `rglob`, matching attribute names from `_REGISTRY_ATTRS`:

```
_action_controllers, _cam_out_size, _cameras, _objects, _policy_threads, _robots, cameras, objects, robots
```

It works: it is what caught #2548's `self._world.robots.get(robot_name)` in `_resolve_joint_write_targets` at review time rather than in production.

## Measured: what sits outside that scope

Running the guard's own `_raw_lookups` helper unchanged over `strands_robots/**/*.py`, excluding the three scanned directories -- **203 files scanned, 4 shape matches**, none of them a live defect:

| site | why it is not a gap |
|---|---|
| `benchmarks/libero/adapter.py:2363` in `_register_default_robot()` — `if "robot" in world.robots` | the name is a string **literal**; it cannot be unhashable |
| `simulation/benchmark.py:218` in `on_episode_start()` — `world.robots.get(rname)` | `rname` iterates `sim.list_robots()`, i.e. names the code produced, never a caller's |
| `policies/wbc/sim_control.py:156` in `from_sim()` — `world.robots.get(robot_name)` | reachable only behind an upstream refusal (below) |
| `policies/wbc/sim_control.py:408` in `wbc_uses_position_servo()` — `world.robots.get(robot_name)` | same |

The first two are closed by construction. The last two are closed only by something upstream, which I measured rather than assumed. `wbc_uses_position_servo` and `from_sim` are reached through `MujocoSimulation._maybe_install_wbc_torque_control(policy, robot_name)`, which is called from `simulation/base.py:2735` on the `run_policy` / `start_policy` path. Against a one-robot world with `robot_name=["alice"]`:

```
[...]

PR #2552, opened 2026-08-21
test(simulation): grade the total-lookup rule over the package, not three directories

cagataycali:
## Problem

`tests/simulation/test_unhashable_entity_name_is_reported.py::test_every_backend_lookup_uses_the_shared_rule` is the check that keeps a registry lookup added later inside the total-lookup rule - a bare `name in registry` / `registry.get(name)` raises `TypeError: unhashable type` for a name that cannot be a key, so the unknown-entity error path the lookup guards is never reached and the exception escapes the agent-tool dict the method documents as its only failure channel.

Its scope was three backend directories, globbed non-recursively:

```python
return [simulation_dir / "mujoco", simulation_dir / "newton", simulation_dir / "isaac"]
...
for path in sorted(directory.glob("*.py")):
```

That is **26 of the package's 229 modules**, and a registry read is not confined to a backend. Running the check's own `_raw_lookups` helper unchanged over the other 203 finds the shape in three more places:

| site | why it is not a live defect |
|---|---|
| `benchmarks/libero/adapter.py:2363` in `_register_default_robot()` - `if "robot" in world.robots` | the name is a string **literal** |
| `simulation/benchmark.py:218` in `on_episode_start()` - `world.robots.get(rname)` | `rname` iterates `sim.list_robots()`, a name the code produced |
| `policies/wbc/sim_control.py:156` / `:408` in `from_sim()` / `wbc_uses_position_servo()` | reachable only behind an upstream refusal |

None of them is a live defect, so nothing here is a bug fix. What is missing is the scope: **the check reported clean over 89% of the files it reads as covering**, and the coupling the last row depends on was asserted nowhere.

That coupling is real. `_maybe_install_wbc_torque_control` is the only route into those two helpers, and unlike the policy binding two lines above it, it is not wrapped in a `try`:

```python
try:
    policy.set_robot_state_keys(self.robot_action_keys(robot_name))
    self.bind_policy_sim_context(policy, robot_name)
except Exception as exc:  # noqa: BLE001 - non-fatal policy configuration
    logger.debug(...)

controller_cleanup = (
    self._maybe_install_wbc_torque_control(policy, robot_name) if wbc_install_torque_control else None
)
```

Handed a name directly it does raise - `_maybe_install_wbc_torque_control(policy, ["alice"])` -> `TypeError: unhashable type: 'list'`. What keeps that unreachable is `run_policy` / `eval_policy` / `start_policy` refusing a `robot_name` that is not in `list_robots()` first, which is a *list* membership test and therefore total for any type. Nothing asserted that ordering.

## Change

Test-only; no production code changes, because none of the sites above is a defect.

- **Widen the scan** from `_backend_dirs()` to `_scanned_modules()` - every module in the package, resolved from a symbol rather than a path literal. 26 -> 229 modules, 203 of them outside the backend directories.
- **Classify the two safe shapes structurally**, so they need no exemption name that can go stale: a lookup whose name is an `ast.Constant`, and one whose name is a local bound by iterating the engine's own `list_robots()` / `list_objects()` / `list_cameras()` - directly, or one hop through a local the listing call filled, which is the spelling `BenchmarkProtocol.on_episode_start` uses. A `registry.get()` written with no argument is still reported: the shape is matched on the registry, and a lookup whose name cannot be read has not been shown to be total.
- **Keep one name-based exemption** for the two WBC helpers, and **assert the refusal it rests on**: for every name that cannot be a key, `run_policy` and `eval_policy` report an error envelope and the WBC hook is never reached, with the converse pinned so skipping it is a real skip. `bind_policy_sim_context` - the other internal hook that reaches the same registry - is pinned as total, since it asks through `registered`.
- **Assert the scan as a floor**, so a `_scanned_modules` that stopped resolving reports no scope rather than no offenders.
- `_package_lookups` is `lru_cache`d: parsing the package costs 0.542 s and five checks read it. The file runs in 5.13 s rather than 9.96 s.

## Verification

Measured at `8ab44cb2` on `mujoco 3.12.0`, the version the declared `mujoco>=3.5.0,<4.0.0` range resolves.

**The widening is load-bearing.** A raw `world.robots.get(robot_name)` planted in `simulation/policy_runner.py` - outside the three directories, a file the package imports on every rollout:

| guard | scope | verdict on the planted tree |
|---|---|---|
| `upstream/main` | 26 of 229 modules | **reports clean** - `1 passed` |
| this branch | 229 of 229 modules | **reports it** - `policy_runner.py:3571 in _planted_raw_lookup()` |

**Every exemption is load-bearing**, each naming exactly the site it is written for when dropped - there is no dead exemption:

| exemption removed | what the check then reports |
|---|---|
| literal name | `adapter.py:2363 in _register_default_robot()` |
| name from the listing call | `benchmark.py:218 in on_episode_start()` |
| `_RESOLVED_UPSTREAM` | `sim_control.py:156 in from_sim()`, `sim_control.py:408 in wbc_uses_position_servo()` |

**The coupling pin is load-bearing.** Deleting `run_policy`'s `if robot_name not in self.list_robots()` refusal makes it fail with `a list robot_name reached the WBC hook via [['crate']]`, for each of the four unhashable builtins.

`tests/simulation/test_unhashable_entity_name_is_reported.py`: 51 -> 67 tests, all passing in 5.31 s.

Blast radius of 490 files - all of `tests/simulation`, every guard that walks a source or test tree, and the repo-wide docstring / changelog / string checks - run on this branch and on `upstream/main`:

|---|---|---|---|---|
| this branch | 43 | 16609 | 299 | 126 |
| `upstream/main` | 43 | 16593 | 299 | 126 |

**169 failing ids, the same set on both trees** - the branch-only and base-only differences are both empty, and none of the 169 is in the changed file. `+16 passed` is exactly the 16 tests this adds. The 169 are all local dependency artifacts (`device_connect_edge`, `awsiot`), both declared extras that CI installs.

`ruff check` and `ruff format --check` clean. `mypy strands_robots tests tests_integ`: 24 errors on both this branch and a pristine `upstream/main` worktree, none of them in the changed file.

## Artifact

Scope, the planted-lookup verdict from each tree's own guard, and what the widened check reports when each exemption is dropped:

the total-lookup rule now grades the package

Refs #2551.

### DIFF
--- a/changelog.d/2552-total-lookup-rule-package-scope.md
+++ b/changelog.d/2552-total-lookup-rule-package-scope.md
@@ -1,0 +1,3 @@
+### Changed
+
+- **tests/simulation**: the check that keeps a registry lookup inside the total-lookup rule now grades every module in the package instead of three backend directories, so the same shape added outside a backend is reported rather than passing unseen. The two safe shapes are classified structurally - a literal name, and a name read out of the engine's own `list_robots()` - and the one remaining exemption now has the refusal it rests on asserted beside it.
--- a/tests/simulation/test_unhashable_entity_name_is_reported.py
+++ b/tests/simulation/test_unhashable_entity_name_is_reported.py
@@ -17,6 +17,7 @@ check at the end is what keeps a lookup added later inside that rule.
 from __future__ import annotations
 
 import ast
+import functools
 import inspect
 import threading
 from collections.abc import Callable
@@ -477,11 +478,43 @@ _REGISTRY_ATTRS = frozenset(
 # this exemption cannot outlive them.
 _CREATION_FUNCTIONS = frozenset({"add_robot", "add_object", "add_camera"})
 
+# Two WBC helpers read the robot registry directly, and neither is reachable
+# with a name a caller supplied: both are internal, and the only route into
+# them is ``MuJoCoSimEngine._maybe_install_wbc_torque_control``, which
+# ``run_policy`` / ``eval_policy`` / ``start_policy`` reach only after refusing
+# a ``robot_name`` that is not in ``list_robots()`` - a *list* membership test,
+# total for any type. That refusal is the whole reason this exemption is safe,
+# so :class:`TestTheRefusalTheUpstreamExemptionRestsOn` asserts it rather than
+# leaving the coupling to be remembered.
+_RESOLVED_UPSTREAM = frozenset({"from_sim", "wbc_uses_position_servo"})
 
-def _backend_dirs() -> list[Path]:
-    """The backend packages, located from a symbol rather than a path literal."""
-    simulation_dir = Path(inspect.getfile(registered)).parent
-    return [simulation_dir / "mujoco", simulation_dir / "newton", simulation_dir / "isaac"]
+# A floor on the scan, not an exact count: a ``_scanned_modules`` that stopped
+# resolving would report an empty offender list and read as a clean package.
+_MINIMUM_SCANNED_MODULES = 150
+
+
+# A lookup can only be handed a name that cannot be a key if the name is a
+# caller's. Two shapes never are, and both are decidable from the AST, so they
+# need no exemption list that could go stale: a literal, and a name the engine
+# read out of its own entity listing. ``_LISTING_METHODS`` is that second
+# source - whatever ``list_robots()`` and its siblings returned are names the
+# code produced, so a lookup keyed on one is total by construction.
+_LISTING_METHODS = frozenset({"list_robots", "list_objects", "list_cameras"})
+
+
+def _scanned_modules() -> list[Path]:
+    """Every module in the package, located from a symbol rather than a path literal.
+
+    Scoped to the whole package rather than the three backend directories,
+    because a registry read is not confined to a backend: the shape already
+    appears in :mod:`strands_robots.simulation.benchmark`,
+    :mod:`strands_robots.policies.wbc.sim_control` and
+    :mod:`strands_robots.benchmarks.libero.adapter`, none of which a
+    backend-directory scan looks at. A check that reports clean over 11% of the
+    files it is read as covering is the failure mode this widening removes.
+    """
+    package_dir = Path(inspect.getfile(registered)).parents[1]
+    return sorted(package_dir.rglob("*.py"))
 
 
 def _is_registry(node: ast.AST) -> bool:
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Test-only widening of the total-lookup AST guard in `tests/simulation/test_unhashable_entity_name_is_reported.py`: the scan grows from the three backend directories (26 modules) to the whole package (229 modules), the two provably-safe lookup shapes (literal names, names iterated from the engine's own `list_*()` calls) are classified structurally rather than via a stale-able exemption list, and the one remaining name-based exemption (`_RESOLVED_UPSTREAM` for the WBC helpers) ships with `TestTheRefusalTheUpstreamExemptionRestsOn`, which pins the `run_policy`/`eval_policy` refusal ordering the exemption depends on - plus the converse (the hook does run for a resolvable name), so skipping it is a real skip. No production code changes, and none are owed: none of the three newly-visible sites is a live defect.

## What's good

- The scan floor (`_MINIMUM_SCANNED_MODULES`) means a `_scanned_modules()` that stops resolving fails loudly instead of reporting a clean package - the same no-silent-defaults discipline AGENTS.md requires of production code, applied to the guard itself.
- Every exemption is verified load-bearing in the PR description (each names exactly the site it exists for when dropped), and `TestTheExemptionsStillDescribeRealCode` keeps that true going forward.
- The `registry.get()` no-argument edge stays reported rather than skipped - an unreadable name is not a proven-total lookup.
- Changelog fragment present (`changelog.d/2552-total-lookup-rule-package-scope.md`), matching the step-3 convention.
- The `lru_cache` on `_package_lookups` returns a tuple, so the shared result cannot be mutated by a consumer.

yinsong1986 (approved):
## Summary

Test-only change that widens the total-lookup AST guard in `tests/simulation/test_unhashable_entity_name_is_reported.py` from the three backend directories (26 modules) to the entire package (229 modules, resolved from the `registered` symbol rather than a path literal). The two provably-safe lookup shapes are now classified structurally instead of by name — a literal key, and a name iterated out of the engine's own `list_robots()`/`list_objects()`/`list_cameras()` (directly or one local-assignment hop away) — while an argless `registry.get()` deliberately stays reported. The one remaining name-based exemption (`_RESOLVED_UPSTREAM` for the two WBC helpers) gets the ordering it depends on pinned: `run_policy`/`eval_policy` refuse an unresolvable `robot_name` before `_maybe_install_wbc_torque_control` is reached, with the converse (the hook does run for a resolvable name) asserted so the skip stays a real skip. The scan itself is floored at 150 modules so a resolution failure reads as lost scope rather than a clean package, and the package parse is `lru_cache`d behind an immutable tuple. No production code changes, and none of the three newly-in-scope sites is a live defect.

## What's good

- Every exemption is pinned as load-bearing (`TestTheExemptionsStillDescribeRealCode`), and the safe-shape classifiers each have both a positive and a negative control (`test_a_local_filled_from_anywhere_else_is_still_a_caller_value`, `test_a_lookup_with_no_name_to_read_is_still_reported`), so the widening cannot silently degrade into matching nothing.
- The coupling the `_RESOLVED_UPSTREAM` exemption rests on is asserted rather than remembered, parametrized over the shared `UNHASHABLE` table, and the sibling hook `bind_policy_sim_context` is pinned as total so the two internal routes cannot diverge.
- AGENTS.md compliance: changelog fragment (`changelog.d/2552-total-lookup-rule-package-scope.md`) instead of a direct `CHANGELOG.md` edit, paths located from symbols rather than literals, no host paths, no emojis.
- The PR description demonstrates each guard is load-bearing by deletion (planted lookup, each exemption dropped, refusal removed) and reports a blast-radius run with identical failure sets on both trees.

Issue #191, opened 2026-05-22
start_cameras_recording daemon-thread artifacts under multi-threaded eval (Strands `Agent` tool dispatch)


yinsong1986:
# `start_cameras_recording` daemon-thread artifacts under multi-threaded eval (Strands `Agent` tool dispatch)

## Summary

`Simulation.start_cameras_recording(...)` spawns a free-running daemon thread that polls cameras at 30 fps via `self.render(camera_name=...)`. The thread shares `mjData` with whatever thread runs the eval. When the eval runs **on the same thread** as the recorder owner (the typical `evaluate_benchmark` from a Python `main()`), this works fine — `mjData` mutations and renderer reads interleave under GIL pressure but don't produce visible artifacts.

When the eval runs **on a different thread** — specifically the Strands `Agent` worker thread that dispatches `libero_sim.evaluate_benchmark` from inside an asyncio event loop — the recorder thread sees concurrent `mjData` mutations mid-render. Result: **2-3% frame capture rate plus visible "greenish" GL clear-colour gradient frames** in the output MP4.

This blocks PR [`strands-labs/robots-sim#26`](https://github.com/strands-labs/robots-sim/pull/26)'s `examples/libero/run_mujoco_agent.py --policy=groot` from producing matrix-quality rollout videos. The eval itself reaches `success_rate=1.00 (5/5)`; only the video-recording side path is affected.

## Reproduction

`examples/libero/run_mujoco_agent.py` from PR `strands-labs/robots-sim#26` at commit `66005d4`:

```bash
unset DISPLAY
MUJOCO_GL=egl PYOPENGL_PLATFORM=egl python3 examples/libero/run_mujoco_agent.py \
    --policy groot --no-auto-server --port 8000 \
    --task libero-10-LIVING_ROOM_SCENE5_put_the_white_mug_on_the_left_plate_and_put_the_yellow_and_white_mug_on_the_right_plate \
    --n-episodes 5 --seed 42
```

Side-by-side (same task, seed, n_episodes):

| Driver | Eval thread | Wall | Recording duration | Recording quality |
|---|---|---|---|---|
| `examples/libero/run_mujoco.py` (programmatic) | main | 53.5 s | full eval span | clean |
| `examples/libero/run_mujoco_agent.py` (Strands `Agent`) | Strands worker | 175.4 s | **4.0 s / 120 frames @ 30 fps** | greenish artifacts at boundaries |

[...]

PR #2553, opened 2026-08-21
fix(simulation): serialize render_depth's mjData read like its two sibling readers

cagataycali:
## Problem

`Renderer.update_scene` copies `xpos`/`xquat`/`xmat` and the geom poses out of mjData, and `Renderer.render` dereferences `data.contact`. Both race a concurrent `mj_step`, and `render()` already carries the comment saying so:

> Reading mjData ... races a concurrent `mj_step` from a policy worker or the `step()` loop, producing torn frames in recorded MP4s (upstream MuJoCo #191) and risking a native crash. The recorder daemon calls `render()` on its own thread, so this path is NOT covered by the blanket dispatch lock; serialize the mjData read + frame copy under `self._lock`.

Three public methods reach that pair. `render()` and `get_frame()` serialize it. `render_depth()` did not - the identical `update_scene` + `render` sequence on the same buffers, outside the lock - while its own docstring says its content "mirrors `render`" and that the depth map is "pixel-aligned with `render` for the same camera".

Measured with a writer thread holding `sim._lock`:

| reader | completes while a writer holds the lock |
| --- | --- |
| `render` | no - blocks |
| `get_frame` | no - blocks |
| `render_depth` | **yes** - returns anyway, having read mjData |

And with a real policy worker stepping on its own thread (60 reads each, `mock` provider at 50 Hz):

| reader | reads landing inside a physics step |
| --- | --- |
| `render` | 0 / 60 |
| `render_depth` | **1-3 / 60** across runs |

The blanket dispatch lock is not the guarantee: it covers the tool surface only, so a direct Python call, the camera-recorder daemon (which renders on its own thread) and a `PolicyRunner` worker all hold nothing.

## Change

`render_depth` now holds `self._lock` across the mjData read and the depth render, and `.copy()`s the buffer inside it, so the sanitize/linearize and the PNG encoding stay unlocked - the shape `render()` already uses. The one-time `ARB_clip_control` stderr capture stays inside the critical section with the render it wraps, because that notice is a C-level write to fd 2 emitted by the render itself.

What `render_depth` reports is unchanged. Captured through the shipped surface on each tree in turn, at the same camera and step count:

| | value | identical |
| --- | --- | --- |
| depth PNG sha256 | `bed43b7f0f9dfd50` | yes |
| `depth_min` / `depth_max` | 0.179472 m / 0.854201 m | yes |
| summary text | `Depth 560x420 from 'cam' ...` | yes |
| RGB PNG sha256 | `0a6eb45ce6e84653` | yes |

## Verification

Measured at `3992f230` (base `9786372c`, MuJoCo 3.12.0, headless EGL).

**Regression test** - `tests/simulation/mujoco/test_frame_readers_serialize_the_mjdata_read.py`, **2 failed / 10 passed** against `upstream/main`, 12 passed after:

```
FAILED ...::test_a_reader_cannot_complete_while_a_writer_holds_the_lock[render_depth]
  render_depth finished while another thread held sim._lock, so its update_scene read
  mjData a concurrent mj_step was free to mutate (update_scene calls during the window: 1)
FAILED ...::test_no_public_reader_reads_mjdata_outside_the_lock
  Reads outside the lock: {'render_depth': [1236, 1238]}
```

The behavioural half needs no wall-clock: the writer holds the real lock and never sleeps, so the serialised verdict is an `Event` that times out. A reader blocked on the lock cannot signal completion while the writer is inside its critical section, whatever the host's load.

The structural half derives the graded set from the module's own AST rather than a hand-written list, so a fourth public reader is held to the rule when it is added. The one private helper that reads mjData the same way (`_get_sim_observation`) is graded through its caller instead - the assertion is that `get_observation` holds the lock across the delegation, which pins the arrangement rather than exempting it.

The **10 tests that pass on both trees are the controls**, and each fails for a specific wrong fix: `render` and `get_frame` already block (so this is an extension of their contract, not a change to it); all three readers still return once the writer releases (no deadlock, and the read still happens); the scan reaches every known reader (non-vacuity); the depth bounds, the grayscale PNG, the unknown-camera refusal and the cached one-time warning are all unchanged.

**Gate** - 353 files touching the readers, the lock discipline, `tests/simulation/mujoco`, and the repo-wide guards, run on both trees:

| | branch | base (with the new test copied in) |
| --- | --- | --- |
| passed | 10047 | 10045 |
| failed | 9 | 11 |
| collected | 10056 | 10056 |

Branch-only failures: **none**. Base-only: exactly the 2 pre-fix failures above. The 9 shared are pre-existing here and unrelated: 8 in `tests/mesh/test_iot_connect_timeout_domain.py` (`awsiotsdk` absent locally; it is a declared extra) and one order-dependent case in `tests/simulation/test_rendering_pacers_survive_a_clock_step.py` that passes in isolation on an untouched tree.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on both this branch and an untouched `upstream/main` worktree, none in the changed files.

## Docs

- `AGENTS.md` **Thread Safety** gains the reader side of the rule it already states for writes: a read that copies model/data *out* races a concurrent `mj_step` exactly as a write does, and the corrupt half is the copy. The bullet carries a `Pinned by` for the new guard and for the existing read-only-query audit.
- `docs/simulation/overview.md` documents the concurrency contract next to the render surface it describes, so a caller capturing frames alongside a policy knows the frame is a consistent snapshot.

## Artifact

render_depth mjData-read serialisation

Both columns are the same script run against each tree in turn on so101 with a 4.4 cm cube. The bottom row is the real MuJoCo pair from the branch - the RGB scene and the depth map the fix protects - and both are byte-identical to the ones the base tree produces, which is what makes this a change to *when* the read happens rather than to what it returns.


yinsong1986 (approved):
## Summary

Closes the one gap in the frame-reader lock discipline: `render_depth` drove the same `Renderer.update_scene` + `render` pair as `render` and `get_frame` — a read that copies `xpos`/`xquat`/`xmat` and dereferences `data.contact` out of mjData — without holding `self._lock`, so a depth capture raced a concurrent `mj_step` from a policy worker, the `step()` loop, or the recorder daemon (none covered by the blanket dispatch lock). The fix holds the lock across the mjData read and the depth render, `.copy()`s the buffer inside it so the sanitize/linearize and PNG encoding stay unlocked (the exact shape `render()` already uses), and keeps the one-time `ARB_clip_control` fd-2 capture inside the critical section with the render that emits it. `self._lock` is an RLock, so re-entrant callers already holding the dispatch lock cannot deadlock on the new critical section. What `render_depth` reports is unchanged, backed by byte-identical PNG/bounds measurements in the description.

## What's good

- The regression test pins both halves: a behavioural check (a reader cannot complete while a writer holds the real lock, verdict via an `Event` rather than wall-clock) and a structural AST scan that derives the graded reader set from the module itself, so a fourth public reader is held to the rule on arrival — per AGENTS.md > Testing > 'Pin every reviewed fix with a regression test'.
- The private helper (`_get_sim_observation`) is graded through its caller's lock rather than exempted, which pins the arrangement.
- Ten control tests document what each wrong fix would break (no deadlock, read still happens, scan non-vacuous, depth answer unchanged).
- Changelog lands as a `changelog.d/` fragment, AGENTS.md gains the reader-side statement of the existing write-lock rule with `Pinned by` citations, and the docs note the concurrency contract next to the render surface. No emojis, no host paths, ASCII-clean.

PR #2554, opened 2026-08-21
docs(hardware): name only actions the tools dispatch

cagataycali:
## Problem

`docs/hardware/tools.md` names nine `action` values no tool dispatches - seven in
its "Key actions" table and two in its runnable examples. Every `@tool` in
`strands_robots.tools` takes a single `action` string and dispatches on its
value, so that value is an enumerated vocabulary rather than free text, and a
name outside it is refused on the first line a reader runs.

Driving each claim through the real tool:

| Documented | Refusal a reader gets |
|---|---|
| `lerobot_calibrate(action="info")` | `**Unknown action:** `info`` - then lists all eight real actions |
| `lerobot_calibrate(action="compare")` | `**Unknown action:** `compare`` - same, recoverable in one step |
| `lerobot_camera(action="stream")` | `Unknown action: stream` - and names no valid action at all |
| `pose_tool(action="fk")` / `"ik"` / `"set_gripper")` | `Unknown action: fk` - the row also promised kinematics this tool has none of |
| `serial_tool(action="list")` | `Port parameter required for this action` - not reported as an unknown action at all |

Two rows are worse than a plain refusal:

- **`serial_tool`** is the quietest. `action="list"` is not diagnosed as unknown;
  the call asks for a port, so a reader supplies one and gets
  `Serial error: could not open port /dev/ttyACM0` next - two steps away from
  `list_ports`, which needs no port. The page already names the real spelling in
  its own prose 33 lines above the example that uses the wrong one
  (``action="list_ports"`` reads none of these options).
- **`pose_tool`**'s "What" column read *Forward/inverse kinematics, gripper
  control*. That tool has no kinematics: it stores and replays named servo poses
  and moves one motor at a time. A reader who needs Cartesian IK is sent to a
  tool that cannot do it, and away from `Simulation.move_to`, which can.

`README.md` describes the same four tools correctly (`view`, "Store, recall, and
execute named robot poses", "discovery, capture, record"), and every other page
that names an action - `device-connect.md`, `policies/groot.md`,
`ros2-integration.md`, `rosbridge-integration.md`, `rtps-integration.md`,
`security.md`, `troubleshooting.md` - is accurate. The drift is confined to this
one page.

The sibling guard `test_docs_python_examples_are_callable` grades the other half
of the same example - whether the *keywords* a documented call passes are
accepted - and a wrong action value passes it, because `action` is a real
parameter of all of these tools. So `pose_tool(action="fk", robot_id=..., port=...)`
spends three accepted keywords and is still refused. On the unfixed tree that
guard reports `4 passed`.

## Change

The four affected rows and both examples now name only actions the tools
dispatch, and `pose_tool`'s description says what it does (joint-space poses and
single-motor moves) and where Cartesian IK actually lives.

`tests/test_docs_tool_action_values_are_dispatched.py` grades the value half:
every action `docs/**/*.md` and `README.md` attribute to a `@tool`, in a
`<tool>(action="...")` call or a tools-table row, must be one that tool
dispatches. The vocabulary is read from the module by AST, so a renamed action
fails the sweep rather than waiting for a reader; a new tool row is graded on
arrival.

The rule is one-directional. The column is "Key actions", so naming a subset is
correct and omission is never reported - only an action no tool dispatches is.
`use_ros` documents 5 of its 10 and is deliberately kept as the control for that.

Driving every action the corrected page names: 38 claims, **0** answered
`unknown action`. Every remaining refusal is an honest per-action requirement
(`view` needs a device id, `record` needs a camera id, `send` needs a port).

## Tests

Copying the new file onto `upstream/main`: **1 failed / 18 passed**. Exactly one
assertion flips - the headline, naming all nine offenders with each tool's real
vocabulary - and the 18 that pass on both trees are the grader's own controls:

- both claim shapes (table row and fenced call) are reached, and >= 40 claims are
  graded, so a page reflow that hid the tables could not report a clean sweep;
- every tool the prose claims dispatches a readable vocabulary, so a short read
  cannot silently accept a claim it should report;
- a documented strict subset is not reported (`use_ros`);
- a planted undispatched action is reported, a planted real one is not, a planted
  table row is graded, and a same-shaped call to something the package does not
  ship degrades to "not graded" rather than to a failure.

## Verification

Measured at `80ba1fa`, `MUJOCO_GL=egl`, mujoco 3.12.0.

| Selection: 76 paths (every test that reads a docs page or walks the test tree, `tests/tools`, and the repo-wide guards), 4553 collected | branch | `upstream/main` + the new file |
|---|---|---|
| result | **4492 passed / 0 failed** / 61 skipped, 190s | 1 failed / 4491 passed / 61 skipped, 191s |

`4491 + 1 == 4492`, so both trees collected the same total. Branch-only failures:
**none**. The single base failure is this PR's own headline. No test in the
selection fails on both trees.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`;
`mypy` identical to a pristine `upstream/main` worktree (5 errors, none in the
changed files).

## Artifact

The same probe run against each tree's own page - identical package code in both
columns, so what differs is only which actions the page names. Each tool's
dispatched vocabulary is byte-identical across the two runs
(`lerobot_calibrate` 8, `lerobot_camera` 8, `pose_tool` 13, `serial_tool` 8).

hardware-tools action vocabulary, before and after

### DIFF
--- a/changelog.d/2554-docs-tool-action-vocabulary.md
+++ b/changelog.d/2554-docs-tool-action-vocabulary.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **docs**: the hardware-tools page named nine `action` values no tool dispatches, in its "Key actions" table and in two of its runnable examples. `pose_tool`'s row advertised forward/inverse kinematics and gripper control, none of which that tool has - it is joint-space only, and Cartesian IK is `Simulation.move_to`. `serial_tool(action="list")` was the quietest: it is not reported as an unknown action at all, so a reader supplies the port the refusal asks for and fails to open it next, two steps away from `list_ports`, which needs no port. The rows and examples now name only actions the tools dispatch, and a new sweep grades every action value `docs/**/*.md` and `README.md` attribute to a `@tool` against the vocabulary that tool really dispatches.
--- a/docs/hardware/tools.md
+++ b/docs/hardware/tools.md
@@ -20,12 +20,12 @@ from strands_robots.tools import (
 
 | Tool | Key actions | What |
 |------|-------------|------|
-| `lerobot_calibrate` | `"list"`, `"info"`, `"search"`, `"compare"`, `"backup"` | Manage existing calibration JSONs under `~/.cache/huggingface/lerobot/calibration/` (this tool inspects/organizes - actual calibration is run via the LeRobot CLI) |
-| `lerobot_camera` | `"list"`, `"test"`, `"stream"` | Enumerate, test, stream connected cameras |
+| `lerobot_calibrate` | `"list"`, `"view"`, `"search"`, `"backup"`, `"restore"` | Manage existing calibration JSONs under `~/.cache/huggingface/lerobot/calibration/` (this tool inspects/organizes - actual calibration is run via the LeRobot CLI) |
+| `lerobot_camera` | `"list"`, `"test"`, `"capture"`, `"record"` | Enumerate, test, capture from, and record connected cameras |
 | `lerobot_teleoperate` | `"start"`, `"stop"`, `"status"`, `"replay"`, `"dagger"` | Leader-follower teleop session, episode replay, and DAgger correction collection |
 | `lerobot_train` | `"start"`, `"status"`, `"stop"`, `"list"` | Fine-tune a policy on a local dataset via `lerobot-train` |
-| `pose_tool` | `"fk"`, `"ik"`, `"set_gripper"` | Forward/inverse kinematics, gripper control |
-| `serial_tool` | `"list"`, `"send"` | Enumerate serial ports, send raw commands |
+| `pose_tool` | `"store_pose"`, `"load_pose"`, `"read_all"`, `"move_motor"` | Store, recall and replay named servo poses on a real bus, and read or move one motor at a time. This tool is joint-space only - Cartesian IK is `Simulation.move_to` |
+| `serial_tool` | `"list_ports"`, `"send"` | Enumerate serial ports, send raw commands |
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Corrects `docs/hardware/tools.md` so its "Key actions" table and two runnable examples name only `action` values the tools actually dispatch (four rows: `lerobot_calibrate`, `lerobot_camera`, `pose_tool`, `serial_tool`), fixes `pose_tool`'s description to stop advertising kinematics it does not have (pointing Cartesian IK at `Simulation.move_to`, which exists in both backends' motion primitives), and adds `tests/test_docs_tool_action_values_are_dispatched.py` to grade every documented action value against the vocabulary read from each tool module by AST. I verified independently: every action the corrected table and examples name is a real dispatch branch in the corresponding tool source; the new test reports 19 passed on the PR tree and exactly 1 failed / 18 passed with the old docs restored (the headline assertion, naming the stale actions), matching the PR's claim; the compare read is `ahead_by=1, behind_by=0` so the tested tree is the merge result; `ruff check` / `ruff format --check` are clean on the new file; no host paths or non-ASCII introduced; changelog fragment present per AGENTS.md step 3. The grader's controls (minimum-claim floor, both claim shapes reached, planted-offender/planted-real/planted-row checks, `use_ros` subset control) make a silent short-read loud rather than trusted.

## What's good

- The new sweep is one-directional by design (subset documentation stays legal), with the `use_ros` control pinning that deliberately.
- The table-row regex reads only the "Key actions" column (non-greedy to the next `|`), so backticked identifiers in the "What" column are not misgraded.

PR #2555, opened 2026-08-21
fix(tools): hold gr00t_inference's selector options to a domain, like the numerics beside them

cagataycali:
## Problem

`gr00t_inference` builds one `docker exec` argv and runs it **detached** (`-d`). Nine
agent-supplied options ride that argv. Three of them - `port`, `denoising_steps`,
`timeout` - are held to a domain at the dispatch boundary, and
`_numeric_option_error` records exactly why:

> `port` and `denoising_steps` are interpolated into a `docker exec` command line that
> runs detached, so a value the server's own argument parser rejects surfaces minutes
> later inside the container's log rather than as this call's result. Refusing them
> here - before any container is built, any checkpoint is downloaded, and any socket is
> opened - is what makes the value's rejection a property of the request rather than of
> how far the orchestration got.

The five **selector strings** beside them in that same argv had no domain:
`data_config`, `embodiment_tag`, `vit_dtype`, `llm_dtype`, `dit_dtype`. Measured
through the tool, with `_start_service` stood in so nothing touches docker:

| request | before | after |
|---|---|---|
| `data_config="So100"` (ordinary typo) | `success`, reaches the starter | refused, names `data_config` |
| `data_config="so100; rm -rf /"` | `success`, reaches the starter | refused |
| `embodiment_tag="gr1\nInjected"` | `success`, reaches the starter | refused |
| `vit_dtype="not a dtype"` (`use_tensorrt=True`) | `success`, reaches the starter | refused |
| `data_config="so100_dualcam"` (**control**) | `success`, reaches the starter | `success`, reaches the starter |

Against the real argv builder the value lands as an argument:

```
argv[14] = 'so100; rm -rf /'   (preceded by '--data-config')
argv[16] = 'gr1 && curl evil'  (preceded by '--embodiment-tag')
argv[23] = 'not_a_dtype'       (preceded by '--vit-dtype')
argv[25] = ''                  (preceded by '--llm-dtype')
```

The argv is argv-style, so this is not shell injection - it is the failure mode the
numeric guard already names: the caller is told the service "failed to start", or is
told nothing at all, while the reason sits in a container log. `AGENTS.md` names these
exact fields (`data_config`, `embodiment_tag`, dtype strings) as surfaces that must be
allowlisted.

## Change

`_enumerable_option_error`, the enum-valued sibling of `_numeric_option_error`, called
from the same boundary immediately after it. It mirrors that helper's structure rather
than introducing a second shape:

- **Keyed on the action.** `_ACTION_ENUMERABLE_OPTIONS` covers `start` / `restart` /
  `lifecycle:full` - derived as the set that builds an inference command line, which is
  the same set the numeric table records for `denoising_steps`. A caller is never
  refused for a value the requested action ignores.
- **Two effectiveness conditions, both read off `_build_inference_command`.** The N1.7
  entrypoint takes no `--data-config` flag, so under `protocol="n1.7"` that value is
  genuinely inert and is left alone - the same carve-out `denoising_steps` already has.
  The three dtype flags are only emitted when `use_tensorrt` is set.
- **The domain is a lowercase `[a-z][a-z0-9_]+` token**, the class the repository
  already requires for these fields. The predicate is total over any input, like
  `_is_allowed_image`, so a non-string is refused rather than carried into `re`.
- **Returns a reason for the dispatcher to fold into its error dict**, so
  "Return error dicts, never raise" still holds at the tool surface.

The refusal names the parameter and where an unreadable value would otherwise have
surfaced, and echoes the value with `!r` so a newline cannot break up the message.

**Non-breaking, and it is derived rather than asserted.** The admitted set is read from
the shipped `strands_robots/policies/groot/data_configs.json` rather than copied, so a
config added there is admitted by construction: all **30** catalogue names and aliases
pass, as do `fp8`, `nvfp4`, `fp16`, `bf16`, `fp32`, `int8`.

`AGENTS.md`'s `### LLM Input Safety` prescribed a single `validate_inputs(...)` helper
that raises `ValueError`, and offered `gr00t_inference.validate_inputs()` as the
canonical example. That symbol exists nowhere in the package, and the module cited is
the one with the most dispersed validation in the tree - eight per-concern predicates
returning `str | None`, no single entry point - so the citation pointed at a
counter-example of its own rule. The bullet now describes the shipped shape and names
the pair that implements it.

## Tests

`tests/tools/test_gr00t_enumerable_option_guards.py`, a sibling of the numeric file,
reusing its `no_side_effects` fixture so every refusal is asserted to reach neither
`subprocess.run` nor a socket.

Against `upstream/main`: **69 failed, 15 passed**. The sharpest single line is a
refused `embodiment_tag="gr1 && curl evil"` returning
`{'message': 'stub', 'status': 'success'}` - the value reached the service starter and
the tool called it a success. After: **84 passed**.

The 15 that pass on both trees are the controls, and each fails for a specific wrong
fix:

- every shipped catalogue `data_config` still reaches the starter carrying the
  caller's value (fails if the domain is narrowed past the catalogue)
- six real TensorRT precisions are admitted
- `protocol="n1.7"` still ignores `data_config`; a dtype is still inert without
  `use_tensorrt`; `status` / `stop` / `list` / `find_containers` read no selector
  (all fail if the guard stops being action-keyed)

Two more pin the root cause structurally: the selector action table equals the set of
actions the numeric table records as reading `denoising_steps`, and every option the
table covers is a real `_build_inference_command` parameter - so a sixth selector
cannot be added to that argv and left ungraded.

## Verification

Measured at `c0d9acd`, blast radius of 132 files (`tests/tools`, every test naming the
tool or the option tables or `AGENTS.md`, plus the repo-wide guards), both trees in
parallel and in lockstep (118.95s vs 118.96s):

- branch **5379 passed, 0 failed, 36 skipped**
- base **69 failed, 5310 passed, 36 skipped** - `5310 + 69 = 5379`, same collected
  total on both
- branch-only failing ids: **empty**; base-only: exactly the new file, zero foreign;
  no pre-existing failures in the selection
  `upstream/main` worktree, none in the changed files
- `ruff check` / `ruff format --check` clean

No sim behaviour, policy, rendering, recording or robot asset changes, so there is
nothing to render; the figure below is the measured before/after of what the caller is
told and where the value stopped.

the same five requests through both trees

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1721,14 +1721,24 @@ Corrections from code review that apply to all future contributions:
   into `subprocess.run`, `subprocess.Popen`, MJCF / XML interpolation, or filesystem
   path construction MUST be validated up front via regex allowlist, enum match, or
   range check. Argv-style subprocess does not exempt you - defense-in-depth.
-- **Centralise validation in one function** - pattern: a `validate_inputs(...)` helper
-  at the top of the tool module that takes every user-supplied param as a keyword arg
-  and raises `ValueError` with a clear message on any rejection. Single entry-point
-  is independently testable. PR #90's `gr00t_inference.validate_inputs()` is the
-  canonical example.
+- **Group the check by what consumes the value, and run it at the dispatch
+  boundary** - one `..._option_error(action, *, <params>) -> str | None` helper per
+  *kind* of value, called before any expensive work starts, returning the reason for
+  the dispatcher to fold into its error dict. Keying it on the action means a caller
+  is never refused for a value the requested action ignores.
+  `gr00t_inference._numeric_option_error` and its enum-valued sibling
+  `_enumerable_option_error` are the shipped pair: both cover options that are
+  interpolated into a `docker exec` command line run **detached**, where a value the
+  server's own flag parser rejects surfaces minutes later in the container log
+  instead of as the call's result. Note the shape returns rather than raises, so
+  **Return error dicts, never raise** above still holds at the tool surface.
 - **Allowlist enumerable values** - `data_config`, `embodiment_tag`, dtype strings,
   container names: all match `^[a-z][a-z0-9_]+$` or an explicit `{"fp16", "fp8", ...}`
-  set. Never accept arbitrary strings into enumerable surfaces.
+  set. Never accept arbitrary strings into enumerable surfaces. Derive the
+  "refuses nothing real" half from a shipped catalogue rather than a copied list -
+  `data_configs.json` names every valid `data_config`, so a config added there is
+  admitted by construction. Pinned by
+  `tests/tools/test_gr00t_enumerable_option_guards.py`.
 - **Reject shell metacharacters in paths** - `;`, `|`, `$`, backticks, `>`, `<`,
   `\n`, `\r`, `\x00`. Also reject `..` path traversal components. Apply even when
   using argv-style subprocess.
--- a/changelog.d/2555-gr00t-enumerable-option-domain.md
+++ b/changelog.d/2555-gr00t-enumerable-option-domain.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **tools/gr00t_inference**: `data_config`, `embodiment_tag` and the three TensorRT dtype flags are now held to a lowercase selector-token domain at the dispatch boundary, like the numeric options beside them in the same detached `docker exec` argv. An unusable value is refused up front, naming the parameter, instead of reporting a service that "failed to start" with the reason left in the container log. The admitted set is derived from the shipped `data_configs.json`, and options the requested action ignores (`data_config` under `protocol="n1.7"`, the dtypes without `use_tensorrt`) are left alone.
--- a/strands_robots/tools/gr00t_inference.py
+++ b/strands_robots/tools/gr00t_inference.py
@@ -442,6 +442,105 @@ def _numeric_option_error(
     return None
 
 
+_ENUMERABLE_OPTION_RE = re.compile(r"^[a-z][a-z0-9_]+$")
+
+#: Per-action set of enumerable string options the action's command line reads.
+#: The keys mirror :data:`_ACTION_NUMERIC_OPTIONS` -- these five options are
+#: consumed by exactly the actions that build an inference command line, which is
+#: the same set that reads ``denoising_steps``.
+_ACTION_ENUMERABLE_OPTIONS: dict[str, tuple[str, ...]] = {
+    "start": ("data_config", "embodiment_tag", "vit_dtype", "llm_dtype", "dit_dtype"),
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

This PR closes the validation gap between the numeric options and the five selector strings (`data_config`, `embodiment_tag`, `vit_dtype`, `llm_dtype`, `dit_dtype`) that ride the same detached `docker exec` argv in `gr00t_inference`. A new `_enumerable_option_error` mirrors the shipped `_numeric_option_error` shape exactly: action-keyed (`start` / `restart` / `lifecycle:full`), effectiveness-gated (`data_config` skipped under `protocol="n1.7"`, dtypes only checked when `use_tensorrt` is set), returning a reason string the dispatcher folds into an error dict rather than raising. I verified the effectiveness conditions against `_build_inference_command` directly - the N1.7 branch really does omit `--data-config`, both branches emit `--embodiment-tag`, and the dtype flags are only extended under `use_tensorrt`, so the guard refuses nothing the requested action ignores. All 30 shipped `data_configs.json` names/aliases and all four signature defaults (`fourier_gr1_arms_only`, `gr1`, `fp8`, `nvfp4`) pass the `^[a-z][a-z0-9_]+$` domain, so the change is non-breaking as claimed. This is exactly the allowlist AGENTS.md > Review Learnings (PR #92) > 'LLM Input Safety' requires for these fields, and the AGENTS.md edit correcting the stale `validate_inputs()` citation is accurate - that symbol exists nowhere in the package.

## What's good

- The 84-test suite passes on the head, reuses the numeric sibling's no-side-effects discipline (a refused call reaches neither `subprocess.run` nor a socket), and asserts the accepted path through a `_start_service` spy rather than only the refusal path.
- Two structural tests pin the root cause: the selector action table is derived from the numeric table's `denoising_steps` entry, and every covered option is a real `_build_inference_command` parameter - so a sixth selector added to that argv cannot go ungraded.
- The admitted set is read from the shipped `data_configs.json` rather than copied, so a catalogue addition is admitted by construction.
- Changelog fragment present, no non-ASCII additions, no host paths, `ruff` clean, and the docstring `Args:` entries for the guarded parameters keep the tool-schema parameter-description rule satisfied (`test_agent_tool_parameter_descriptions.py` passes).

## Verification suggestions

- `python3 -m pytest tests/tools/test_gr00t_enumerable_option_guards.py tests/tools/test_gr00t_numeric_option_guards.py -q` - 84 + 113 pass on the head.
- Spot-check the non-breaking claim: `python3 -c "import json,re; c=json.load(open('strands_robots/policies/groot/data_configs.json')); rx=re.compile(r'^[a-z][a-z0-9_]+$'); print([v for v in list(c['configs'])+list(c.get('aliases',{})) if not rx.match(v)])"` prints `[]`.

PR #2556, opened 2026-08-21
fix(simulation): refuse a joint write whose robot_name no robot carries

### LINKED ISSUE #2549: decision(simulation): a dict-form joint write accepts a robot_name no robot carries, and the list form refuses it - now that robot_name selects the resolution namespace, one call has two answers
## What

In the dict form of `set_joint_positions` / `set_joint_velocities`, a `robot_name` that no robot in the world carries is accepted and silently does nothing. The list form refuses the same value:

```python
sim.set_joint_positions([0.1, 0.2], robot_name="nobody")
# -> error: unknown robot 'nobody' (via _unknown_robot_msg)

sim.set_joint_positions({"j1": 0.1}, robot_name="nobody")
# -> success | "Set 1/1 joint positions, FK updated"
```

So one call has two answers for the same bad argument, and in the dict form the answer is a success.

## Why it matters now

Before #2548 this was invisible, because the dict form ignored `robot_name` outright -- there was nothing for a wrong value to be wrong about. #2548 makes `robot_name` load-bearing for dict resolution: it selects the namespace an unqualified key resolves in. A misspelled robot name therefore now means "scope this write to a namespace that does not exist", which falls back to the pre-existing cross-robot first-match lookup and writes *some* robot's joint while the caller believes they addressed a specific one.

That is the same class of defect #2453 reported (a write reported against a robot that did not receive it), reached by a different route: not an unscoped name, but an unchecked scope.

## Why this is filed rather than folded into #2548

Refusing it changes the set of accepted calls -- today's silently-tolerated `robot_name` typo becomes a `status="error"` -- which is a different kind of change from #2548, where the namespace attempt is tried first and falls through so that nothing that worked is refused. Keeping them apart keeps #2548's blast-radius claim ("no call that worked is now refused") true and separately reviewable. #2548 pins the current fallback behaviour in `test_unknown_robot_name_falls_back_rather_than_raising`, so whichever way this is decided, the pin is the thing to update and the decision will be explicit rather than incidental.

## Options

- **Refuse it**, reusing `_unknown_robot_msg`, matching the list form exactly. Loudest and symmetric; rejects calls that succeed today, so it wants a changelog note as a behavioural change.
- **Refuse it only when it would have changed resolution** -- i.e. when at least one key is unqualified. A fully qualified mapping does not consult the namespace, so an unknown `robot_name` is inert there. Narrower, but the rule is harder to state.
- **Keep tolerating it** and document that `robot_name` is advisory in the dict form. Cheapest, but leaves one argument with two meanings across the two accepted forms.

My inclination is the first: `robot_name` is now genuinely load-bearing in both forms, so the two forms disagreeing about whether it must name a real robot is the anomaly.

## Scope

`_resolve_joint_write_targets` already looks the robot up (`self._world.robots.get(robot_name)`) and currently treats a miss as "no namespace", so the check has a natural home and needs no new lookup. Both setters share the resolver, so one change covers both. Update `test_unknown_robot_name_falls_back_rather_than_raising` in `tests/simulation/mujoco/test_joint_write_scopes_a_bare_name_to_robot_name.py`.

Refs #2453, #2548.

cagataycali:
Closes #2549

## What

The dict form of `set_joint_positions` / `set_joint_velocities` accepted a `robot_name` that names no robot in the world and reported success. The list form refused the same value:

```python
sim.set_joint_positions([0.1, 0.2], robot_name="nobody")
# -> error: Robot 'nobody' not found. Available robots: [...]

sim.set_joint_positions({"j1": 0.1}, robot_name="nobody")
# before -> success | "Set 1/1 joint positions, FK updated"
# now    -> error: Robot 'nobody' not found. Available robots: [...]
```

One argument had two answers, and in the dict form the answer was a successful write. This implements option 1 of the three the issue set out (refuse, reusing the list form's `_unknown_robot_msg`), which is the option the issue and its follow-up both favoured.

## Why it had to change

Tolerating the value was inert while the dict form ignored `robot_name` outright -- there was nothing for a wrong value to be wrong about. Scoping an unqualified key to `robot_name`'s namespace made the argument load-bearing: it now selects the namespace a bare joint name resolves in. So a misspelled robot name means "scope this write to a namespace that does not exist", which fell through to the pre-existing cross-robot first-match lookup and wrote **whichever robot declared the name first** -- while the success text reported the requested count for a robot that never moved.

That is the same wrong-robot write already fixed for an unqualified name, reached through an unchecked *scope* rather than an unqualified *name*.

## How

One branch in the shared joint-write resolver, so it covers both setters and both call forms from one place:

- It reuses the list form's own message rather than writing a second one, so the two forms now answer a bad `robot_name` **identically** -- the divergence is deleted, not papered over.
- The lookup already routes through the total `registry_entry`, so the same branch also refuses a name that cannot be a registry key at all (a list, a dict, a set -- what wrapping a name in brackets or passing a half-built kwargs mapping produces) instead of raising `TypeError: unhashable type` past the `{"status", "content"}` contract.
- The list form validates `robot_name` before normalising to a dict, so for that form the new branch is unreachable and behaviour is unchanged.

Production change is 4 lines of logic; the rest is the docstrings that state the contract and the pins.

## Blast radius

This refuses calls that previously succeeded -- which is the point: each one was writing a joint on a robot the caller did not name. **Nothing that resolved correctly changes.** All still reach the same lookup as before, and are pinned:

| call | behaviour |
|---|---|
| bare key, no `robot_name` | unchanged (first-match lookup) |
| qualified `<robot>/<joint>` key | unchanged, not re-scoped |
| bare key naming no joint of `robot_name` but one elsewhere | unchanged (cross-robot fallback) |
| known `robot_name` | unchanged |

The one in-tree caller that passes `robot_name` -- the LIBERO adapter's init-state arm-qpos apply -- takes it from `list_robots()`, i.e. a name the code produced, so it cannot trip the refusal.

**Refusal precedence is pinned too.** A real `robot_name` with an unresolvable joint key must still be diagnosed as the joint problem it is; a robot lookup firing too eagerly would replace the joint hint (which names the key and the joints the model does have) with "Robot 'bob' not found", sending the caller after the wrong argument.

## Pins

The three cases that recorded the tolerated behaviour are flipped **deliberately** rather than incidentally, as the issue asked:

- `test_unknown_robot_name_falls_back_rather_than_raising` -> `test_unknown_robot_name_is_refused` (renamed for the behaviour it now asserts)
- `test_dict_form_reports_a_robot_name_that_cannot_be_a_key` -> `test_dict_form_refuses_...` (4 parametrised cases; the envelope check stays, since that is what separates a refusal from a `TypeError`)
- `test_velocities_share_the_total_lookup` (4 parametrised cases)

Added:

- `test_the_two_forms_agree_on_an_unknown_robot_name` -- asserts the two refusals are the *same string*, not merely both errors. That is the actual claim of the issue, and it is what a future change cannot satisfy by inventing a second, differently-worded refusal.
- `test_a_known_robot_name_is_still_accepted` -- non-vacuity: the refusal must key on the name, not refuse every name.
- `test_velocities_refuse_an_unknown_robot_name` -- the sibling setter for a plain misspelling.
- refusal-precedence and no-leak-into-`qpos`/`qvel` assertions on the refused paths.

No references to the two renamed cases survive anywhere in the tree (checked across `*.py`, `*.md`, `*.ipynb`).

## Verification

```
tests/simulation/mujoco/test_joint_write_scopes_a_bare_name_to_robot_name.py
  27 passed

with the production change reverted:
  11 failed, 16 passed      <- the pins grade the fix; the 16 are the
                               blast-radius cases, which pass either way

guards + neighbours (624 + 643 selected):
  tests/test_changelog_{format,fragments,fragment_gate}.py
  tests/test_test_case_names_describe_behaviour.py
  tests/test_docstring_xref_roles_resolve.py
  tests/test_docstrings_no_dangling_doc_refs.py
  tests/test_args_docstring_completeness.py
  tests/test_source_no_implicit_string_concatenation_in_lists.py
  tests/simulation/test_unhashable_entity_name_is_reported.py
  tests/simulation/mujoco/test_joint_setter_unknown_names_rejected.py
  tests/simulation/test_sim_engine_describe_discovery.py
  -> all pass except two pre-existing environment failures, verified
     identical on a clean `main` checkout in the same runner:
       test_sim_engine_describe_discovery: 'so100' in []        (asset absent)
       test_joint_discovery_hint_is_followable: 7 failed        (so101 absent)
     Same test IDs, same cause (model file not on disk); this branch changes
     only the message text on an already-red path.

ruff check / ruff format --check: clean
mypy strands_robots/simulation/mujoco/physics.py: no issues
python scripts/assemble_changelog.py --check: OK
scripts/check_duplicate_claim.py --issue 2549: unique-claim
```

## Changelog

`changelog.d/2549-joint-write-refuses-an-unknown-robot-name.md` -- a behavioural change, so it is recorded as `Fixed` with the blast radius spelled out.

## §13 Review rounds

- **round 0** (initial): the refusal in the shared resolver, three pins flipped, three added, the precedence pin, public docstrings updated, changelog fragment.

### DIFF
--- a/changelog.d/2549-joint-write-refuses-an-unknown-robot-name.md
+++ b/changelog.d/2549-joint-write-refuses-an-unknown-robot-name.md
@@ -1,0 +1,43 @@
+### Fixed: a dict-form joint write refuses a `robot_name` no robot carries
+
+`set_joint_positions` / `set_joint_velocities` now refuse a `robot_name` that
+names no robot in the world in the dict form, with the message the ordered
+(list) form already used. Previously the two accepted forms answered the same
+bad argument differently, and the dict form's answer was success:
+
+```python
+sim.set_joint_positions([0.1, 0.2], robot_name="nobody")
+# -> error: Robot 'nobody' not found. Available robots: [...]
+
+sim.set_joint_positions({"j1": 0.1}, robot_name="nobody")
+# -> success: "Set 1/1 joint positions, FK updated"   (before)
+```
+
+Tolerating it was harmless while the dict form ignored `robot_name` outright --
+there was nothing for a wrong value to be wrong about. Scoping an unqualified
+key to `robot_name`'s namespace made the argument load-bearing: it now selects
+the namespace a bare joint name resolves in, so a misspelled robot name means
+"scope this write to a namespace that does not exist". That fell through to the
+pre-existing cross-robot first-match lookup and wrote *some* robot's joint --
+whichever declared the name first -- while the caller believed they had
+addressed a specific one, and the success text reported the requested count for
+a robot that never moved. It is the wrong-robot write already fixed for an
+unqualified name, reached instead through an unchecked scope.
+
+The refusal lives in the shared joint-write resolver, so it covers both setters
+and both call forms from one place, and it covers two kinds of bad name at
+once: one no robot carries, and one that cannot be a registry key at all (a
+list, a dict, a set -- what wrapping a name in brackets or passing a half-built
+kwargs mapping produces). The lookup is routed through the total
+`registry_entry`, so the second kind resolves to "no such robot" and is
+reported rather than raising `TypeError: unhashable type` past the
+`{"status", "content"}` contract.
+
+Blast radius: this refuses calls that previously succeeded, which is the point
+-- each one was writing a joint on a robot the caller did not name. Nothing
+that resolved correctly changes. A bare key with no `robot_name` at all, a
+fully qualified `<robot>/<joint>` key, and a bare key that names no joint of
+`robot_name` but does name one elsewhere in the scene all reach the same lookup
+as before. The one in-tree caller that passes `robot_name` (the LIBERO
+adapter's init-state arm-qpos apply) takes it from `list_robots()`, so it
+cannot trip the new refusal.
--- a/strands_robots/simulation/mujoco/physics.py
+++ b/strands_robots/simulation/mujoco/physics.py
@@ -1323,8 +1323,9 @@ class PhysicsMixin:
             method: Calling method name, used in error text.
             robot_name: The robot an unqualified key belongs to. Its namespace is
                 tried first, so a bare name reaches that robot's joint even when
-                a sibling robot declares the same short name. ``None`` (or a name
-                no robot in the world carries) leaves resolution as it was.
+                a sibling robot declares the same short name. ``None`` leaves
+                resolution as it was; a name no robot in the world carries is
+                refused, on the same terms as the list form.
 
         Returns:
             ``({joint_name: joint_id}, None)`` when every name resolves, else
@@ -1355,11 +1356,28 @@ class PhysicsMixin:
         # lands on a robot the caller never addressed (#2453). Scoping first
         # keeps that fallback for everything it already resolved: a bare name
         # that is not a joint of this robot still reaches the shared lookup.
+        #
+        # A ``robot_name`` no robot carries is refused rather than read as "no
+        # namespace". Once the namespace is load-bearing, a misspelling means
+        # "scope this write to a namespace that does not exist", which falls
+        # through to the cross-robot first-match lookup and writes some other
+        # robot's joint while the caller believes they addressed a specific one -
+        # the defect #2453 reported, reached through an unchecked scope instead of
+        # an unqualified name. The list form already refuses the same value here,
+        # so this deletes a divergence rather than adding a rule: one argument had
+        # two answers, and in the dict form the answer was success (#2549).
+        # ``registry_entry`` keeps the lookup total, so a name that cannot be a
+        # registry key at all is refused by this same branch rather than raising
+        # ``TypeError`` past the structured-error contract.
         ns = ""
         if robot_name is not None and self._world is not None:
             robot = registry_entry(self._world.robots, robot_name)
-            if robot is not None:
-                ns = robot.namespace or ""
+            if robot is None:
+                return {}, {
+                    "status": "error",
+                    "content": [{"text": self._unknown_robot_msg(robot_name)}],
+                }
+            ns = robot.namespace or ""
 
         resolved: dict[str, int] = {}
         unresolved: list[str] = []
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

The dict form of `set_joint_positions` / `set_joint_velocities` silently tolerated a `robot_name` naming no robot in the world: the namespace lookup missed, `ns` stayed empty, and the write fell through to the cross-robot first-match lookup -- writing whichever robot declared the bare joint name first while reporting success for the robot the caller named. This PR adds one guard in the shared `_resolve_joint_write_targets` so that case is refused with the exact string the list form already used (`_unknown_robot_msg`), covering both setters and both call forms from one branch. The lookup routes through the total `registry_entry`, so an unhashable name (list/dict/set/bytearray) also resolves to the same refusal instead of raising `TypeError` past the `{"status", "content"}` envelope. I verified at head `959b2386`: all 27 pins in the touched test file pass, 515 neighbour guard tests (changelog fragments, Args completeness, unhashable-entity reporting, joint-setter rejection) pass, refusal precedence holds (a real `robot_name` with a bad joint key still gets the joint diagnosis), and the message-equality pin holds because both forms call the one `_unknown_robot_msg`. The behaviour change refuses only calls that were writing a robot the caller never addressed, and the refusal is an explicit structured error rather than a silent default change.

## What's good

- The two-forms-agree pin asserts string equality of the refusals, not merely two errors -- that is the actual claim of #2549 and cannot be satisfied by a second, differently-worded message.
- Refusal precedence is pinned (`test_unresolvable_name_is_still_all_or_nothing`), so the new robot refusal cannot later pre-empt the more actionable joint hint.
- No-leak assertions on every refused path (`qpos`/`qvel` compared before/after) keep the all-or-nothing contract observable rather than assumed.
- Every in-tree caller passing `robot_name` (LIBERO adapter via `list_robots()`, `examples/17_pour_task.py`'s `carton` added by the same script, the curobo collector's scene-config names) uses a name the code itself produced, so the blast-radius claim in the description checks out.
- Changelog fragment present and correctly shaped; renamed tests carry no dangling references.

PR #2557, opened 2026-08-21
fix(strings): render tool-result text in ASCII, and grade the surface the rule names first

cagataycali:
## Problem

AGENTS.md's Unicode rule names three surfaces:

> Tool result dicts (`{"content": [{"text": ...}]}`), log messages, error messages: plain ASCII only. Agents read these strings programmatically.

`tests/test_log_strings_are_ascii.py` grades two of them. Its own first sentence repeats all three ("AGENTS.md mandates plain ASCII in logs, error messages, **and tool-result text**"), and its scan walks `logger.<level>(...)` / `warnings.warn(...)` arguments and `raise` messages. The text a tool result carries -- the surface the rule names first, and the one whose entire audience is programmatic -- is graded by nothing.

Ten glyphs sat in it. Seven **published** sim actions rendered one, measured through the tool surface on a real MuJoCo scene:

| action | glyph | agent-visible text |
| --- | --- | --- |
| `get_contact_forces` | `U+2194` | `ground U+2194 cube_geom: normal=1.226N, dist=-0.0004m` |
| `set_geom_properties` | `U+2192` | `Geom 'cube': color U+2192 [1.0, 0.0, 0.0, 1.0]` |
| `set_body_properties` | `U+2192` | `Body 'cube': mass: 0.100 U+2192 0.500` |
| `get_mass_matrix` | `U+00D7` | `Mass matrix: 6U+00D76, rank=6, cond=1.85e+00` |
| `apply_force` | `U+00B7` | `Torque: [0.0, 0.0, 0.0] NU+00B7m` |
| `randomize` | `U+00B1` | `Positions: 1 dynamic objects perturbed by U+00B10.02m` |
| `list_benchmarks` | `U+2022` | `  U+2022 demo_pick: _Demo (robots=['so101'], ...)` |

The eighth site is `gr00t_inference`'s checkpoint-download `message`.

`get_contact_forces` is the sharpest: `simulation.mujoco.rendering` already lists the identical contact pair as `geom1 <-> geom2`, so one package rendered the same content two ways depending on which module answered.

## Change

**Source.** Each of the ten becomes the spelling the package already uses for the same content: `<->` (`simulation.mujoco.rendering`), `->`, `+/-` (8 existing sites), `x` (`rendering.ibl`, `rendering.backgrounds`), `N*m` (`simulation.isaac.simulation`), `  - ` for a bullet (four existing sites in `simulation.mujoco.simulation`). No message keeps a count or a prefix an existing test matches on, so `"No active contacts."` and `"Positions: N dynamic objects perturbed"` are untouched.

**Guard.** `test_log_strings_are_ascii.py` gains the third rung. It needs a flow step the other two do not: a handler rarely spells its report inline -- the dominant idiom builds it up in a local and joins that into the returned dict, so grading only the expressions written inside the `return` sees no literal at all (eight of the ten offenders are written exactly that way). `_tool_result_strings` follows the local names that reach the `content` / `message` value and grades the strings assigned or appended to them.

It stays narrower than "every literal in a function that returns a result dict": on this tree the flow-following scan grades **4947** strings and that blanket alternative **10002**, and both report the same ten offenders -- so nothing here is caught by breadth alone. The `json` block is machine-readable payload, not prose, and is not graded. Semantic Unicode in a docstring stays untouched, as the two byte-level sibling scans already intend.

**Docs.** The AGENTS.md rule gains the flow-step note, the ASCII renderings, and a `Pinned by`.

## Verification

Measured at `a12d3cbc`, against `upstream/main` at `f9065e60`. Base moved twice during the work (#2555 and #2556, each touching a file this PR also touches); both auto-merges were clean, both sides' markers were verified present by symbol afterwards, and the numbers below are from the final merged tree.

**The shipped guard reports clean on the unfixed tree** -- `3 passed` with all ten glyphs present. That is the gap.

Pre-fix split, the widened guard against unfixed source: **1 failed / 7 passed -> 8 passed**. Exactly one assertion flips, and the seven that hold on both trees are the point: two of them are the log/`raise` surface, unchanged, and four are new controls.

Each control is load-bearing, proven by breaking the detector:

| break | control that fires |
| --- | --- |
| drop the flow step | `test_the_scan_follows_a_report_built_up_in_a_local` -- `flow step missed the appended line: ['text', '']` |
| widen to every literal in a handler | `test_a_string_that_never_reaches_the_result_is_left_alone` and `test_semantic_unicode_in_a_handler_docstring_is_left_alone`, each naming the string it protects |

Zero dead controls.

Gate: 174-file selection (every test naming the changed result surfaces, every guard that walks the package sources or grades strings/docstrings, every test reading AGENTS.md, plus the repo-wide docs and changelog guards), run against both trees:

| --- | --- | --- | --- |
| this branch | 0 | 6676 | 90 |
| `upstream/main` + this guard file | 1 | 6675 | 90 |

`6675 + 1 == 6676` on both, so the two runs collected the same tests; the failed delta equals the passed delta equals the one pre-fix failure; the branch-only failing set is empty and the base-only set is exactly `test_tool_result_strings_are_ascii`. No failure is shared by the two trees, so the selection carries no pre-existing red.

`ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` 24 == 24 against a pristine worktree, branch-only empty.

## Behaviour is unchanged

Every action's `status` and every JSON payload block is byte-identical across the two trees (7/7 matching sha256). This changes how a report is spelled, not what it reports.

tool-result text before and after

## Scope

Only text a tool result renders as prose. The 93 non-ASCII literals in package docstrings are deliberate and stay -- the rule is about what a caller reads back, not about how the source documents itself.


yinsong1986 (approved):
## Summary

Swaps ten Unicode glyphs in agent-facing tool-result text for the ASCII spellings the package already uses elsewhere for the same content (`<->`, `->`, `+/-`, `x`, `N*m`, `  - `), and closes the guard gap that let them sit there: `tests/test_log_strings_are_ascii.py` gains a third surface (`content` / `message` values of returned result dicts) with a flow-following AST scan that grades strings accumulated in locals before being joined into the return value. Verified locally at `a6976b47` (the head then moved only by `3753143a`, which adds the `changelog.d/2557-tool-result-text-ascii.md` fragment AGENTS.md step 3 requires — well-formed, single `### Fixed` entry): all 8 guard tests pass, the changed files are ruff-clean, no test or consumer matches on the old glyph spellings (all remaining `U+2194`/`U+00B7` instances in the tree are docstrings, comments, or docs — deliberately out of scope), the modified test file itself stays pure ASCII by spelling its fixture glyphs as `\uXXXX` escapes, and the four new detector controls exercise both the recall (flow step) and precision (untainted literal, docstring) boundaries. `status` values and `json` payload blocks are untouched, so no machine-matched surface changes. No blocking concerns.

## What's good

- The guard is pinned in AGENTS.md (`Pinned by ...`), matching the repo's documented-and-enforced pattern rather than leaving the rule aspirational again.
- Each fixed spelling is sourced from an existing site in the package rather than invented, resolving the `geom1 <-> geom2` vs `geom1 U+2194 geom2` inconsistency in the same act.
- The `_flow_into` fixed point is provably terminating (tainted set only grows, bounded by identifier count), and the scan errs wide (over-taint grades more strings) rather than narrow, which is the safe direction for this class of guard.

yinsong1986 (approved):
## Summary

Swaps ten Unicode glyphs for ASCII spellings in tool-result prose across seven sim actions plus the GR00T checkpoint-download message, and closes the guard gap: `tests/test_log_strings_are_ascii.py` gains a third surface (`content` / `message` values of returned result dicts) via a flow-following AST scan that traces string literals through the local-accumulator idiom (`lines.append(...)` then join) that eight of the ten offenders used. Verified locally at `e8cb61e8` and re-verified at `a12d3cbc` (a zero-conflict base merge of #2556, which also touches `physics.py`; the PR's own diff is byte-identical across the two heads): the widened guard fails on unfixed source naming all seven `physics.py` glyph sites and passes on the branch (8/8 in the guard file); no test in the tree asserts on the old glyph spellings in any changed string (the remaining `U+2194`/`U+00B7` hits are docstrings and comments in unrelated tests, deliberately out of scope); no package or example code parses the changed prose, and each action's `status` and `json` payload blocks are untouched, so nothing downstream keys on the old text. The chosen ASCII renderings match spellings the package already uses for the same content (`<->` in `simulation.mujoco.rendering`, `+/-`, `x`, `N*m`, `  - `), the `_flow_into` fixed-point provably terminates (tainted set only grows, bounded by identifier count), the three detector controls each protect a distinct precision property, a changelog fragment is present per AGENTS.md step 3, and the AGENTS.md rule gains the flow-step note with a `Pinned by`. `ruff check` and `ruff format --check` are clean on the changed files, and 2393 tests across the touched simulation/tools surfaces pass.

yinsong1986 (approved):
## Summary

Swaps ten Unicode glyphs in agent-facing tool-result prose for the ASCII spellings the package already uses for the same content (`<->`, `->`, `+/-`, `x`, `N*m`, `  - `) across seven sim actions plus the GR00T checkpoint-download `message`, and closes the guard gap by adding the third surface (`content` / `message` values of returned result dicts) to `tests/test_log_strings_are_ascii.py` via a flow-following AST scan that traces string literals through the local-accumulator idiom into the returned dict. Verified independently: the widened guard run against the unfixed base tree fails naming exactly the ten claimed sites (`base.py:4137`, `physics.py:810/1168/1913/2150/2155/2178/2240`, `randomization.py:405`, `gr00t_inference.py:1836`), and the branch passes 8/8; no test in `tests/` or `tests_integ/` asserts on any of the changed prose strings (the remaining `U+2194`/`U+00B7` hits are docstrings, which the guard deliberately leaves alone and the three precision controls pin); every `status` and `json` payload block is untouched, so nothing structured changes shape.

## What's good

- The flow step (`_flow_into`) is what makes the guard real: eight of the ten offenders live in `lines.append(...)` locals a return-expression scan cannot see, and `test_the_scan_follows_a_report_built_up_in_a_local` pins that step against removal.
- Precision controls (`test_a_string_that_never_reaches_the_result_is_left_alone`, `test_semantic_unicode_in_a_handler_docstring_is_left_alone`) keep the scan from drifting into the blanket-literal alternative the docstring argues against.
- ASCII renderings match existing package spellings rather than inventing new ones - `get_contact_forces` now agrees with `simulation.mujoco.rendering`'s `<->` for the identical contact pair.
- Changelog fragment per AGENTS.md step 3; AGENTS.md rule updated with the flow-step note and a `Pinned by` reference in the same PR.

PR #2559, opened 2026-08-21
fix(tools): hold lerobot_camera's documented vocabularies to a domain, like the numerics beside them

cagataycali:
## Problem

`lerobot_camera` holds seven numeric options to a shared domain before a camera is
opened, and `_numeric_option_error` records why: *"every value here is
agent-supplied, so each is checked against the shared domain for its kind before a
camera is opened"*, and *"validating here rather than letting the camera driver
object is what makes the refusal a property of the request instead of the device"*.

Two documented enumerated options sat in the same signature unchecked, and unlike a
frame rate neither is ever refused downstream. Each names a closed vocabulary that
`_create_camera` resolved by lookup, and an unrecognised spelling selected a
**plausible neighbour** instead of failing:

```python
color_mode_enum = ColorMode.RGB if color_mode.upper() == "RGB" else ColorMode.BGR
rotation_enum   = rotation_map.get(rotation.upper(), Cv2Rotation.NO_ROTATION)
```

So anything that was not exactly `RGB` selected `BGR`, and any rotation the map did
not name selected no rotation. Both substitutions reported success, and neither
appears anywhere in the tool result.

The `color_mode` half is the expensive one. The driver delivers frames in the
channel order its config asks for, and the capture path converts what it is handed
with `COLOR_RGB2BGR` unconditionally, so the saved image has its red and blue
channels transposed. Measured on a MuJoCo render of an arm and a red target, driven
through the real tool (`action="capture"`) with a camera that honours its config:

| `color_mode` | resolved to | status | reported | saved image |
|---|---|---|---|---|
| `"RGB"` | `ColorMode.RGB` | `success` | Image Capture Success! | red 6736 px / blue 120605 px |
| `"rgb "` (trailing space) | **`ColorMode.BGR`** | **`success`** | Image Capture Success! | **red 120605 px / blue 6736 px** |

The two counts trade places exactly: the red target is written blue. A trailing
space or a transposed pair of letters (`"RBG"`) was enough to reach it.

`rotation` is the quieter half - `"ROTATE_90_DEG"` and `"90"` both resolved to
`NO_ROTATION`, so the caller receives a correct image of the wrong orientation,
again under `status="success"`.

## Change

Both vocabularies become module-level maps, and they are the single owner of the
accepted set: `_create_camera` resolves a spelling through them and
`_vocabulary_option_error` refuses a spelling that resolves to nothing, so the
accepted set and the enforced set cannot diverge.

`_vocabulary_option_error` is the enum-valued counterpart to
`_numeric_option_error`, folded by the dispatcher the same way and scoped the same
way - `_ACTION_VOCABULARY_OPTIONS` names the actions that actually apply a
selector, so `discover` and `list`, which probe devices without applying either,
are never refused for one. It is total over any input: a non-string is refused
rather than reaching `.upper()`, where it would raise past the tool's structured
result.

Comparison stays case-insensitive, matching how `_create_camera` already resolved
these, so **every value that worked before still works** - this refuses only the
spellings that used to be silently replaced. The refusal names the action, the
option, the accepted vocabulary and the value:

```
capture: color_mode must be one of RGB, BGR (compared case-insensitively), got
'rgb '. An unrecognised value used to select a different one silently.
```

The two public `Args:` entries now state the domain rather than only listing the
values.

## Verified in simulation

A MuJoCo render of the arm and a red target is fed through the real tool in both
trees. The left column is the same scene captured with a spelling one space off.

color_mode transposition

The declared spelling is byte-identical in both columns (red 6736 / blue 120605
px), which is the control: this changes what is refused, not what is captured.

## Tests

`tests/tools/test_lerobot_camera_vocabulary_options.py`, the enum-valued sibling of
the existing numeric-domain file. Against `upstream/main`: **106 failed / 16
passed**; after: **122 passed**. 98 of those failures are behavioural - the
headline one reports

```
assert result["status"] == "error"
E  AssertionError: {'content': [{'text': ' **Image Capture Success!**
E  assert 'success' == 'error'
```

for a request that used to transpose the image. The other 8 are the drift guards
naming the new vocabulary maps.

The 16 that pass on both trees are the controls, and each fails for a specific
wrong fix: every declared spelling is accepted in any case (so the fix cannot
narrow to exact-match), `discover`/`list` are not refused for an inert selector,
the numeric domain beside it still refuses `fps=0`, and the saved pixels for the
declared spelling are still the scene that was photographed.

## Scope

`format` is deliberately left alone and a test pins that. Unlike the two selectors
an unlisted `format` is *honored* rather than replaced - it becomes the saved
file's extension and OpenCV writes it if it can - so narrowing it to the three
names the docstring lists would refuse working requests, and its failure mode is
loud rather than silent. Which extensions should be accepted is a separate
question.

## Verification

Measured at `a932e291`. The gate numbers above were measured at `838311d5`; the only
change since is the changelog fragment's filename, so the diff under test is unchanged.

- 86-file selection (every test naming the changed symbols, all of `tests/tools`,
  and the repo-wide docs / changelog / string / docstring guards), both trees:
  branch **3764 passed / 0 failed / 8 skipped**, base **3658 passed / 106 failed /
  8 skipped**. `3658 + 106 == 3764` collected on both, failed delta == passed
  delta == 106, **branch-only failures empty**, every base-only id from the new
  file, and no failure shared by the two trees.
- `ruff check` / `ruff format --check` clean over `strands_robots tests
  tests_integ`; `mypy` 24 == 24 against a pristine `upstream/main` worktree, none
  in the changed files.


yinsong1986 (approved):
## Summary

Closes the silent-substitution gap on `lerobot_camera`'s two enumerated selectors: an unrecognised `color_mode` used to resolve to `BGR` (transposing red/blue in the saved image under an "Image Capture Success!" summary) and an unrecognised `rotation` to `NO_ROTATION`, both reporting success. The fix hoists both vocabularies into module-level maps that are the single owner of the accepted set, adds `_vocabulary_option_error` as the enum-valued sibling of the existing `_numeric_option_error` gate (same dispatch fold, same per-action scoping), and updates the two `Args:` entries to state the domain. Verified locally at `a932e291`: the gate is scoped to exactly the six actions whose handlers call `_create_camera` (all six call sites confirmed), it runs before any camera is opened, it is total over non-string inputs (refused ahead of `.upper()`, so nothing raises past the structured result), declared spellings remain case-insensitive so no previously-honored request is refused, the refusal text is ASCII-only and names action/option/vocabulary/value, and `discover`/`list` are not refused for inert selectors. The new test file plus both existing camera test files pass (252 passed), `ruff check`/`ruff format --check` are clean on the changed files, and the changelog fragment is in `changelog.d/` per AGENTS.md step 3. The deliberate exclusion of `format` (honored rather than replaced, loud failure mode) is reasoned and pinned by a test.

## What's good

- Single-owner vocabularies: `_create_camera` resolves through the same maps the gate validates against, with drift guards pinning the equivalence, so the accepted and enforced sets cannot diverge.
- Follows AGENTS.md's "Reject invalid modes loudly" / no-silent-defaults conventions, and matches the established numeric-domain pattern in the same file rather than inventing a second shape.
- Controls pin the non-regressions that a wrong fix would break: any-case acceptance of declared spellings, defaults never refused, `discover`/`list` untouched, the numeric gate beside it unchanged, and the declared-spelling capture still writes the scene photographed.

PR #2560, opened 2026-08-21
fix(tools): hold lerobot_train's device to torch's own device-string domain

### LINKED ISSUE #2558: fix(tools): lerobot_train's device is the one unchecked token in its detached argv, so an unusable value reports a started run
## The gap, measured

`lerobot_train(action="start")` builds an argv in `build_train_command`
(`strands_robots/tools/lerobot_train.py:371`) and runs it with
`subprocess.Popen(..., start_new_session=True)` (line 784), writing stdout to a
session log. The tool then returns `status="success"` with a PID, a log path and
the command line.

Every other caller-supplied value in that argv is already held to a domain
*before* the dispatch:

| value | guard | where |
| --- | --- | --- |
| `steps`, `batch_size` | `positive_count_error` | line 435 |
| `lora_r`, `lora_alpha` | `positive_count_error` | line 523 |
| `val_episodes` | `positive_count_error` + split checks | line 539 |
| `dtype` | gated on the resolved lerobot policy config's fields | line 493 |
| `gradient_checkpointing` | same | line 502 |
| `policy_type` (under `train_expert_only`) | live lerobot registry | line 422 |
| `num_gpus` | `>= 1` | line 427 |
| **`device`** | **none** | **line 469** |

`device` is interpolated straight into `--policy.device={device}` and is the only
value in the fresh-run argv carried through unchecked. The comment beside the
size guards states the reason those exist:

> An unusable value here is not merely rejected downstream: it is written into
> the argv of a DETACHED process, so the caller is told the run started and only
> the training log records that it could not be honored.

That reasoning applies unchanged to `device`, and the parameter's own docstring
already names a vocabulary rather than a free string (line 651):

```
device: Torch device (cuda, cuda:0, cpu, mps).
```

## What a caller sees today

`device="gpu"` is the ordinary mistake - it is the word the rest of the world
uses, and it is not a torch device type. Measured against torch 2.13:

```
>>> torch.device("gpu")
RuntimeError: Expected one of cpu, cuda, ipu, xpu, mkldnn, opengl, opencl,
ideep, hip, ve, fpga, maia, xla, lazy, vulkan, mps, meta, hpu, mtia,
privateuseone device type at start of device string: gpu
```

So the run aborts - but it aborts inside the detached process. The tool call has
already returned `**Training Session Started**` with a PID, so the agent driving
it believes training is running, and the reason is in a log file the agent was
not told to read. `device="CUDA"` and `device="cuda:x"` fail the same way.

This is the same shape #2555 fixed for `gr00t_inference`'s five selector strings
("reported a service that failed to start minutes later from the container log,
or reported success") and the same shape #2537 fixed for a GR00T restart whose
teardown could not free the port.

## Scope

* A local guard in `lerobot_train.py` refuses a `device` no torch build can
  parse, naming the parameter, before the argv is dispatched. Per AGENTS.md #11
  this stays local and the domain goes in the field's docstring - `device` is one
  field with one caller, so nothing is lifted into `utils.py`.
* The admitted domain is derived by handing the value to `torch.device` rather
  than by comparing against a copied list of device types, mirroring how
  `_expert_only_policy_types` and `_policy_config_field_names` in the same file
  source their domains live from lerobot. A torch build that gains a backend is
  admitted with no change here, and torch's own exception already enumerates the
  types it accepts, so the refusal text is derived too.
* **Spelling only, never availability.** `torch.device("cuda")` constructs fine
  on a CPU-only box, and it must keep doing so: a queued or containerised run
  legitimately names a device the dispatching machine does not currently have,
  and a machine's device inventory is not a property of the argv. The guard must
  therefore reject a non-`str` before consulting torch - `torch.device(0)` *does*
  consult the accelerator inventory (`RuntimeError: Cannot access accelerator
  device when none is available`), which would make the refusal
  machine-dependent.
* Scoped to the fresh-run path. The resume branch returns early with only
  `--config_path` + `--resume=true` and never emits `--policy.device`, so a
  resumed run must not be refused for a value it does not carry - the same
  "never refuse a value this action ignores" contract #2555 encodes.
* When torch is not importable the domain is unknown and the value passes
  through unguarded, exactly as `_policy_config_field_names` documents for its
  own field set.

## Not in scope

`strands_robots/training/lerobot.py` carries the same unvalidated string into
`--policy.device=` (line 1096) and `--reward_model.device=` (line 1154) from
`LerobotTrainer.__init__`'s `device` argument, which is `device or
_auto_device()` with no check. That is a second surface with a different entry
point and a different owner (`training/_validate.py`'s spec gates, where `device`
is not a `TrainSpec` field at all), so it is filed separately rather than folded
in here. `training/reward.py`'s `device` is in-process and raises where the
caller sees it, so it is not part of this class.

## Acceptance

* `build_train_command(..., device="gpu")` raises `ValueError` naming `device`,
  and the tool returns `status="error"` without a `Popen`.
* `device="cuda"`, `"cuda:0"`, `"cpu"`, `"mps"`, `"cpu:0"` still build an argv on
  a CPU-only test machine.
* A resume-path build with an unusable `device` is not refused.
* Pinned by a new `tests/tools/` test, asserted through the dispatch boundary
  (no `Popen`), not by calling the guard directly.

cagataycali:
Fixes #2558

## What

`build_train_command` interpolates `device` into `--policy.device=` in the argv of
a process `lerobot_train(action="start")` launches **detached**, then returns
`status="success"` with a pid and a log path. `device` was the one
caller-supplied token in that fresh-run argv with no domain.

Everything beside it is already held to one before the dispatch: `steps`,
`batch_size`, `lora_r`, `lora_alpha` and `val_episodes` against the shared
positive-count domain; `dtype`, `gradient_checkpointing` and `policy_type`
against a live lerobot registry; `num_gpus` against `>= 1`. The comment beside
the size guards already states the reason those exist:

> An unusable value here is not merely rejected downstream: it is written into
> the argv of a DETACHED process, so the caller is told the run started and only
> the training log records that it could not be honored.

`device="gpu"` is the ordinary mistake - it is the word the rest of the world
uses and it is not a torch device type - so the run aborted before its first step
while the tool reported it started. `"CUDA"` and `"cuda:x"` fail the same way.

## How

A local `_torch_device_error` refuses the value at the dispatch boundary, naming
the parameter. Per AGENTS.md #11 this stays local and the domain is stated in the
field's docstring - one field, one caller, so nothing is lifted into `utils.py`.

The admitted domain is **torch's own**, read by handing the value to
`torch.device` rather than by comparing against a copied list of device types.
That mirrors how `_expert_only_policy_types` and `_policy_config_field_names` in
the same file source their domains live from lerobot: a torch build that gains a
backend is admitted here with no change, and torch's own exception already
enumerates the types it accepts, so the refusal quotes it instead of restating
the set.

## Three scoping decisions worth reviewing

**Only the spelling is graded, never availability.** `torch.device("cuda")`
constructs on a CPU-only box and must keep building an argv there - a queued or
containerised run legitimately names a device the dispatching machine does not
currently have, and a machine's device inventory is not a property of the argv.
Pinned by `test_cuda_builds_on_a_machine_without_cuda`, plus four
valid-but-essentially-never-present device types (`xla`, `vulkan`, `hpu`, `meta`)
in the admitted probe set.

**A non-`str` is refused before torch is consulted.** `torch.device(0)` reads the
accelerator inventory (`RuntimeError: Cannot access accelerator device when none
is available`), so routing an int through torch would build one request on a GPU
box and refuse it on a CPU box. The type gate keeps the verdict
machine-independent.

**Scoped to the fresh-run path.** The resume branch has already returned with
only `--config_path` + `--resume=true` and never emits `--policy.device`, so a
resumed run is not refused for a value its argv does not carry - the same "never
refuse a value this action ignores" contract #2555 encodes. Pinned both ways: the
refusal does not fire on resume, and a non-vacuity test asserts the resume argv
really carries no device flag.

When torch is not importable the domain is unknown and the value passes through
unguarded, as `_policy_config_field_names` documents for its own field set.

## Tests

`tests/tools/test_lerobot_train_device_domain.py`, 68 cases, mirroring the shape
of its sibling `test_lerobot_train_size_knob_domain.py`:

* 20 unusable spellings refused and never reaching the argv, including the
  injection shapes and the non-strings.
* 10 parseable spellings still reaching it.
* The refusal names the parameter, the tool, the value, **and a spelling the
  caller can use instead**.
* The domain is torch's verdict, not a local list: every probe in both halves is
  asserted against `torch.device` directly, so the probe sets cannot drift from
  the guard.
* The refusal reaches the caller as an error envelope and launches no process
  (asserted through a `Popen` spy, so a guard that ran after the dispatch could
  not pass).
* The other argv guards are asserted untouched.

## Not in scope

`strands_robots/training/lerobot.py` carries the same unvalidated string into
`--policy.device=` and `--reward_model.device=` from `LerobotTrainer.__init__`.
Different entry point, different validation owner (`training/_validate.py`'s spec
gates, where `device` is not a `TrainSpec` field at all), so it is left for a
separate change rather than folded in here. `training/reward.py`'s `device` is
in-process and raises where the caller sees it, so it is not part of this class.

## Verification

`ruff check` and `ruff format --check` clean on both files; `mypy` reports
nothing in the changed file. The new test file passes 67/68 locally - the single
failure is `test_the_tool_reports_an_error_envelope_rather_than_raising`, which
needs `lerobot` importable for the tool's `start` action to get past its first
check. The pre-existing sibling test of identical shape fails identically in the
same minimal environment, and CI installs `features = ["all"]`.

### DIFF
--- a/changelog.d/2560-lerobot-train-device-domain.md
+++ b/changelog.d/2560-lerobot-train-device-domain.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **tools/lerobot_train**: `device` is now held to torch's own device-string domain at the dispatch boundary, like the run-size numerics beside it in the same detached argv. `device="gpu"` (or `"CUDA"`, or `"cuda:x"`) is refused up front, naming the parameter and a spelling to use instead, rather than aborting the detached run after the tool has already reported it started with a pid. The admitted set is torch's own - the value is handed to `torch.device` rather than compared against a copied list - and only the spelling is graded, so naming a device the dispatching machine does not currently have still builds an argv. A resumed run, whose argv carries no `--policy.device`, is not refused for one.
--- a/strands_robots/tools/lerobot_train.py
+++ b/strands_robots/tools/lerobot_train.py
@@ -368,6 +368,55 @@ def _has_resumable_checkpoint(output_dir: str) -> Path | None:
     return last if last.exists() else None
 
 
+def _torch_device_error(device: Any) -> str | None:
+    """Error text for a ``device`` no torch build can parse, or ``None``.
+
+    ``device`` is interpolated into ``--policy.device=`` in the argv of a
+    DETACHED process, so an unusable value is not refused where the caller can
+    see it: the tool reports a started run with a PID and a log path, and only
+    the training log records that lerobot aborted before the first step. That is
+    the reason the run-size numerics in the same argv are refused up front, and
+    ``device`` is the one token beside them that was carried through unchecked.
+
+    The admitted domain is torch's own, read by handing the value to
+    ``torch.device`` rather than by comparing against a copied list of device
+    types - the same "source the domain live" shape as
+    :func:`_expert_only_policy_types` and :func:`_policy_config_field_names`
+    above. A torch build that gains a backend is admitted here with no change,
+    and torch's own exception enumerates the types it accepts, so the refusal
+    names the admitted set without restating it.
+
+    Only the spelling is graded, never availability: ``torch.device("cuda")``
+    constructs on a CPU-only box and must keep building an argv there, because a
+    queued or containerised run legitimately names a device the dispatching
+    machine does not currently have. A non-``str`` is refused before torch is
+    consulted for that reason - ``torch.device(0)`` reads the accelerator
+    inventory (``Cannot access accelerator device when none is available``),
+    which would make the same request build here and refuse there.
+
+    When torch is not importable the domain is unknown and the value passes
+    through unguarded, as :func:`_policy_config_field_names` documents for its
+    own field set.
+    """
+    if not isinstance(device, str):
+        return (
+            f"lerobot_train: device must be a torch device string, got {type(device).__name__}. "
+            "Pass a device type, optionally with an index (e.g. 'cuda', 'cuda:0', 'cpu', 'mps')."
+        )
+    try:
+        import torch
+    except Exception:  # noqa: BLE001 - torch missing -> domain unknown, pass through
+        return None
+    try:
+        torch.device(device)
+    except (RuntimeError, ValueError) as e:
+        return (
+            f"lerobot_train: device={device!r} is not a torch device string ({e}). "
+            "Pass a device type, optionally with an index (e.g. 'cuda', 'cuda:0', 'cpu', 'mps')."
+        )
+    return None
+
+
 def build_train_command(
     dataset_root: str,
     policy_type: str = "act",
@@ -412,7 +461,8 @@ def build_train_command(
             freeze the VLM and are mutually exclusive), if ``train_expert_only``
             is requested for a non-expert policy, if ``num_gpus < 1``, or if a
             supplied ``steps`` / ``batch_size`` - or, under ``lora``, a supplied
-            ``lora_r`` / ``lora_alpha`` - is not a positive integer.
+            ``lora_r`` / ``lora_alpha`` - is not a positive integer, or if a
+            fresh run's ``device`` is not a device string torch can parse.
     """
     if lora and train_expert_only:
         raise ValueError(
@@ -459,6 +509,14 @@ def build_train_command(
         # only honors --config_path + --resume. Other flags are ignored on resume.
         cmd.extend([f"--config_path={resume_config}", "--resume=true"])
         return cmd
+
+    # The one selector string in the fresh-run argv. Checked here rather than
+    # beside the size guards above so a resumed run - which returned already, with
+    # only --config_path + --resume and no --policy.device - is never refused for a
+    # value its argv does not carry.
+    device_error = _torch_device_error(device)
+    if device_error:
+        raise ValueError(device_error)
 
     # Fresh-run flags.
     cmd.extend(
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes the one unvalidated caller-supplied token in `build_train_command`'s fresh-run argv: `device` is now checked against torch's own parser (`torch.device`) at the dispatch boundary, so `device="gpu"` / `"CUDA"` / `"cuda:x"` are refused with the parameter, tool, value and a usable spelling named, instead of the tool reporting a started run whose detached process aborted before its first step. The refusal is raised inside `build_train_command` and lands in the tool's existing `except (ValueError, FileNotFoundError)` envelope, so the error-dict contract holds and no process is spawned (pinned via a `Popen` spy). Scoping is careful in both directions: only the spelling is graded, never availability (`cuda` still builds an argv on a CPU-only box, plus exotic-but-valid types), non-`str` values are refused before torch is consulted so the verdict cannot vary by machine inventory, and the resume path — whose argv carries no `--policy.device` — returns before the guard, with a non-vacuity test asserting that argv really emits no device flag.

## What's good

- AGENTS.md #11 followed to the letter: single-caller guard stays local with the domain stated in the field's docstring, nothing lifted into `utils.py`.
- The domain is sourced live from torch rather than a copied device-type list, matching `_expert_only_policy_types` / `_policy_config_field_names` in the same file, and the test file's non-vacuity classes assert both probe halves against `torch.device` directly so the probes cannot drift from the guard.
- Changelog fragment present (`changelog.d/2560-lerobot-train-device-domain.md`), tool docstring's `Args:` entry for `device` updated (AGENTS.md #13), and the torch-unimportable pass-through mirrors the documented `_policy_config_field_names` contract — an absent source of truth does not become a refusal.
- Scope discipline: the parallel unvalidated `device` in `strands_robots/training/lerobot.py` is explicitly named as out of scope with its different validation owner, rather than silently folded in.

Issue #2561, opened 2026-08-21
check_merge_base_overlap.py cannot see a whole-tree grader, whose coupling is to files it never names


cagataycali:
## What

`scripts/check_merge_base_overlap.py` decides that two pull requests are worth composing by intersecting their changed-**path** sets. A test that grades the whole package shares no path with the files it grades, so its intersection with every sibling is empty and the sweep reports clean - while the grader is in fact coupled to all of them.

AGENTS.md step 8 already names the highest-value case it *does* catch:

> A file carrying a `strict=True` xfail is the highest-value overlap candidate there is, because its whole purpose is to fail when a sibling change lands.

That one is catchable because the xfail lives in a file a sibling also edits. A whole-tree grader is the same hazard with the path evidence removed.

## Measurement

Three pull requests merged as a batch on 2026-08-21, all `behind_by: 0`, each individually green:

| PR | changed paths |
|---|---|
| #2557 | `AGENTS.md`, `changelog.d/2557-*.md`, `strands_robots/simulation/base.py`, `strands_robots/simulation/mujoco/physics.py`, `strands_robots/simulation/mujoco/randomization.py`, `strands_robots/tools/gr00t_inference.py`, **`tests/test_log_strings_are_ascii.py`** |
| #2559 | `changelog.d/2559-*.md`, `strands_robots/tools/lerobot_camera.py`, `tests/tools/test_lerobot_camera_vocabulary_options.py` |
| #2560 | `changelog.d/2560-*.md`, `strands_robots/tools/lerobot_train.py`, `tests/tools/test_lerobot_train_device_domain.py` |

Path intersections:

```
2557 n 2559 = []
2557 n 2560 = []
```

But #2557's `test_tool_result_strings_are_ascii` resolves its population from `_python_sources()`, a walk of the package directory. That set **includes** `strands_robots/tools/lerobot_camera.py` and `strands_robots/tools/lerobot_train.py` - the two files the siblings modify. Both siblings add new tool-result prose, which is precisely the surface the new grader scores.

So the composition was never compiled by any of the three CI runs, and no signal available at merge time reported a reason to compose it: each PR read `MERGEABLE` / `CLEAN`, each passed `Detect an untested overlap with the base branch`, and the pairwise path intersection is empty.

The batch was safe, but only because it was checked by hand:

```
git diff of added lines, PRs 2559 and 2560, matching [^\x00-\x7F]  ->  0
```

[...]

PR #2562, opened 2026-08-21
fix: state the scope an except tuple actually has

cagataycali:
## Problem

An `except` tuple that names a class another member already covers does not catch two things. `except (FileNotFoundError, OSError)` is `except OSError` -- and with it every other `OSError` subclass: `PermissionError`, `IsADirectoryError`, `TimeoutError`, the whole `ConnectionError` family. The covered name contributes no scope. What it contributes is the impression of a small superset over a handler that has a large one, which is the opposite of what AGENTS.md > "Exception Clauses Must Be Narrow" asks for, and that section already names the shape -- `except (ImportError, Exception)` -- as a bug.

Where the prose on such a handler enumerates the tuple, the impression becomes a claim. `BridgeTransport.put` carried `except (RuntimeError, ConnectionError, OSError)` under a docstring reading *"(RuntimeError from closed session, ConnectionError from broker drop, OSError from socket-level write) are absorbed; everything else propagates"*. It named three classes and caught twenty-one. Driven through the real `put` with one leg raising:

| leg raises | named by the tuple or its prose | `put()` |
|---|---|---|
| `NotImplementedError` | no | absorbed, returned normally |
| `PermissionError` | no | absorbed, returned normally |
| `RecursionError` | no | absorbed, returned normally |
| `TimeoutError` | no | absorbed, returned normally |
| `ConnectionResetError` | no | absorbed, returned normally |
| `ValueError` | no | propagated (control) |

`NotImplementedError` matters on this path in particular. `strands_robots.mesh.sensors` re-raises it ahead of its own broad handler at every one of its seven sensor loops, with the reason in a comment -- "surface immediately rather than silently dropping every sensor tick" -- and the bridge is the transport those publishes traverse.

Nothing refuses the shape. `ruff` has `B014`, but it reports a *duplicate* member only; run over `except (FileNotFoundError, OSError)` and `except (ImportError, Exception)`, `B014` reports neither, and neither does the full catalogue (`--select ALL`). So AGENTS.md's "Lint/review will catch this" described no mechanism, and 25 sites accumulated across 12 modules.

## Change

Drops the covered member at each of the 25 sites. This is a strict no-op: the absorbed set is unchanged (21 classes before and after on `BridgeTransport.put`, and every probe verdict above is identical in both columns of the artifact). One import existed only to name a class in such a tuple (`benchmarks.libero.suite`'s `BDDLParseError`, beside `ValueError`) and goes with it -- the handler's own log line already carries the BDDL context.

Corrects the prose that enumerated a tuple: both `BridgeTransport` docstrings, the `mesh.audit` comment that named `json.JSONDecodeError`, and the `inference.client` comment that named `TimeoutError`. The bridge docstrings now say plainly that `RuntimeError` and `OSError` are class trees, so the absorbed surface is wider than two names.

Adds `tests/test_except_tuples_state_their_real_scope.py`, which refuses the shape over `strands_robots`, `tests`, `tests_integ` and `examples`, and updates the AGENTS.md bullet: the collapse generalises beyond `Exception`, no linter reports it, and the rule is pinned.

Scope is redundancy only. Whether a handler's *effective* surface should be narrower -- whether `except OSError` around a socket write ought to be `except ConnectionError` -- is a behaviour change, one per site, and is out of scope. A third-party exception tree is left alone on purpose: `huggingface_hub`'s `HfHubHTTPError` is an `OSError` only by way of `httpx.HTTPError`, a hierarchy this repository does not control, so a handler naming both is hedging and the rule must not refuse it. Two such sites in `policies/lerobot_local/processor.py` are untouched and pinned as a control.

## Tests

`tests/test_except_tuples_state_their_real_scope.py`, 14 items. Against `upstream/main`: **1 failed / 13 passed**, the failure naming all 26 redundancies across the 12 modules. After: 14 passed.

The 13 that pass on both trees are the controls, and each fails for a specific wrong fix: a covered member is reported whichever order it is written in; a tuple of genuine siblings (`ValueError, OSError`; `UnicodeDecodeError, json.JSONDecodeError`; `RuntimeError, AttributeError, OSError`) is left alone; a stdlib member is graded; a third-party hierarchy is not; an unresolvable name is skipped rather than reported; and the scan is held to a floor of 200 `except` tuples so a clean result cannot be vacuous.

`tests/mesh/test_bridge_transport.py::TestBridgeCloseFailSoft` already parametrizes over `[RuntimeError, ConnectionError, OSError]` and asserts each is swallowed. It passes unchanged, which is the behaviour-level proof that dropping `ConnectionError` from the tuple changed nothing.

## Verification

Measured at `a6be5d39`, full `tests/` on both trees with the new file present in each:

| tree | failed | passed | skipped | errors | collected |
|---|---|---|---|---|---|
| this branch | 29 | 34036 | 295 | 24 | 34384 |
| `upstream/main` | 30 | 34035 | 295 | 24 | 34384 |

Branch-only failing ids: none. Base-only: exactly the one headline test. The 29 shared failures and all 24 errors are `awsiot`-absent `tests/mesh/test_iot_*` in this environment; `[mesh-iot]` is a declared extra.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors, byte-identical to a pristine `upstream/main` worktree, none in a changed file.

stated vs real scope

Each column runs one capture script against that tree, reading the tuple out of its own source and the claim out of its own docstring, then driving the real `put`. The absorbed count and every verdict are identical across the columns; only the claim moves.

### Rebased onto a moved base

`main` advanced by three commits while this was in flight (#2557, #2559, #2560), overlapping this branch on `AGENTS.md` and `strands_robots/tools/gr00t_inference.py`. Merged at `227b497c`; both files auto-merged with no conflict, and composition was checked by symbol rather than by absence of conflict - this PR's AGENTS.md pin and its three `gr00t_inference` handlers are present alongside the merged PRs' own additions and all three of their changelog fragments.

Re-verified on the merged tree: the pre-fix split is unchanged (1 failed / 13 passed against the new `upstream/main`), `ruff check` and `ruff format --check` clean, `mypy` still 24, and the overlapping selection - the new guard plus `tests/mesh/test_bridge_*`, `tests/inference`, `tests/registry`, `tests/tools` and the repo-wide string, docstring and changelog guards - is 3702 passed / 8 skipped. The full-suite table above was measured at `a6be5d39`, before the merge.

### Composition against the open queue

This branch *adds* a whole-tree grader, which is the shape #2561 records as invisible to a pairwise path check: the new test resolves its population from a walk of `strands_robots`, `tests`, `tests_integ` and `examples`, so it shares no path with the files it grades, and `Detect an untested overlap with the base branch` reads `SUCCESS` on compositions no CI run has compiled. Checked by asking the grader instead (that issue's option 2 -- run the new test against the sibling's tree), over the whole open population:

| composition | merge | grader |
|---|---|---|
| this branch alone | - | 14 passed |
| + #2480 / #2497 / #2498 / #2508 / #2516 / #2517 / #2534 | clean, each | 14 passed, each |
| #1035, #1722 | `CONFLICTING`, cannot compose | added `except` tuples: 0 |

Four siblings do add an `except` tuple, and each survives the rule for a reason a control already pins: `(ImportError, AttributeError)` in #2480 and #2534, and `(RuntimeError, ValueError, OSError, AttributeError, TypeError, ImportError)` in #2498, name unrelated trees; `(KeyboardInterrupt, Exception)` in #2497 is not redundant, because `KeyboardInterrupt` descends from `BaseException` rather than `Exception`; and #2534's `(GatedRepoError, LocalEntryNotFoundError, ImportError, OSError)` is the third-party hierarchy the rule declines to grade. So this guard does not turn `main` red behind any sibling, and no merge order between them matters.

Seven compositions, about 10 s each once the sibling refs are fetched. Measurement recorded on #2561, since the cost is the argument for putting it in the sweep rather than running it by hand.


yinsong1986 (approved):
## Summary

Drops the covered member from 25 `except` tuples across 12 modules (`except (FileNotFoundError, OSError)` -> `except OSError`, `(RuntimeError, ConnectionError, OSError)` -> `(RuntimeError, OSError)`, etc.), corrects the prose that enumerated those tuples as if the names were the scope, and pins the rule with `tests/test_except_tuples_state_their_real_scope.py`. I verified every subclass relationship the diff relies on, including the one non-builtin case: `BDDLParseError` is declared as `class BDDLParseError(ValueError)` in `strands_robots/benchmarks/libero/bddl_parser.py:59`, so dropping it (and its now-unused import) beside `ValueError` in `benchmarks/libero/suite.py` changes nothing. `TimeoutError`, `ConnectionRefusedError`, `ConnectionError`, `FileNotFoundError`, `PermissionError` < `OSError`; `json.JSONDecodeError` < `ValueError`; `NotImplementedError`, `RecursionError` < `RuntimeError` - all hold on Python 3.12, so every edit is a strict no-op on the absorbed set, exactly as the description claims. The bridge docstring corrections also keep AGENTS.md rule 17's `RecursionError`-absorption analysis intact, since `RecursionError` remains covered via `RuntimeError`.

## What's good

- The one behavioural-looking edit (`BDDLParseError` removal) is backed by the class actually subclassing `ValueError`, and the handler's log line retains the BDDL context.
- Scope discipline: redundancy-only, with the effective-surface question (e.g. `OSError` vs `ConnectionError` around a socket write) explicitly deferred, and third-party hierarchies (`HfHubHTTPError`) deliberately excluded with a pinned control - matching the AGENTS.md preference for no-behaviour-change rules enforced by a gate rather than prose.
- The new test carries a 200-tuple vacuity floor and skips unresolvable names rather than reporting them, so a clean pass cannot be an empty scan and optional-dep-gated modules cannot produce false positives.
- Changelog fragment present (`changelog.d/2562-except-tuple-real-scope.md`), and `tests/mesh/test_bridge_dedup.py`'s guidance string was updated in the same PR so its remedy text cannot re-prescribe the dropped member.

## Verification suggestions

- `python -c "print(issubclass(__import__('strands_robots.benchmarks.libero.bddl_parser', fromlist=['BDDLParseError']).BDDLParseError, ValueError))"` - the single edit whose no-op status depends on this repository's own hierarchy rather than builtins.
- `pytest tests/mesh/test_bridge_transport.py::TestBridgeCloseFailSoft tests/test_except_tuples_state_their_real_scope.py` - the behaviour-level proof plus the new gate in one run.

yinsong1986 (approved):
## Summary

Drops the covered member from 25 `except` tuples across 12 modules (e.g. `except (FileNotFoundError, OSError)` -> `except OSError`, `(RuntimeError, ConnectionError, OSError)` -> `(RuntimeError, OSError)`), corrects the prose that enumerated those tuples as if the names were the scope, and pins the rule with `tests/test_except_tuples_state_their_real_scope.py`. I independently verified all nine subclass relations the drops rely on (`issubclass` true for every pair, including `BDDLParseError < ValueError` and `TimeoutError < OSError` on 3.12), so each removal is a strict no-op on the absorbed set; the `mesh.sensors` `NotImplementedError` framing in the description is accurate given the bridge docstring previously claimed "everything else propagates". The new guard passes at head (14/14), its controls cover order-independence, genuine-sibling tuples, third-party hierarchies, unresolvable names and a 200-tuple vacuity floor, and the behaviour-level proof holds: `tests/mesh/test_bridge_transport.py` (including the parametrized `[RuntimeError, ConnectionError, OSError]` swallow test) and `tests/mesh/test_bridge_dedup.py` pass unchanged (131 passed), as do the audit, registry and episode-judge suites for the other touched handlers.

## What's good

- Scope discipline: redundancy-only, with the effective-surface narrowing question explicitly deferred per site, and the third-party hedge (`HfHubHTTPError` beside `OSError`) deliberately excluded and pinned as a control.
- The AGENTS.md bullet update replaces a "lint will catch this" claim that described no mechanism (`B014` reports duplicates only) with an actual gate.
- Changelog fragment per step 3; composition against the open queue checked via the grader itself rather than path overlap, per #2561.

PR #2563, opened 2026-08-21
fix(training): hold LerobotTrainer's device to torch's own device-string domain

cagataycali:
## Problem

`Trainer.validate` is documented as a pure preflight, and its contract is explicit about what an empty list means:

> Return a list of human-readable problems; **an empty list means the spec is launchable.** [...] it powers a `plan` advisor that runs *before* anything expensive starts.

`LerobotTrainer.validate` returned `[]` for a `device` no torch build can parse.

`device` is the trainer's one constructor knob, and it reaches lerobot twice:

- `build_command` interpolates it into `--policy.device=` (and `--reward_model.device=` on the reward-model path),
- `build_config` assigns it onto `policy_cfg.device`.

Neither refuses an unusable spelling. The argv is built verbatim and the config carries the string forward, so the run aborts before its first step - after the preflight said it was launchable.

Measured on this tree (`torch.device` is the oracle):

| `device` | `validate()` | `build_command` emits | `build_config` sets | `torch.device()` |
|---|---|---|---|---|
| `'cuda'` | `[]` | `--policy.device=cuda` | `'cuda'` | OK |
| `'cpu'` | `[]` | `--policy.device=cpu` | `'cpu'` | OK |
| `'cuda:0'` | `[]` | `--policy.device=cuda:0` | `'cuda:0'` | OK |
| `'gpu'` | **`[]`** | `--policy.device=gpu` | `'gpu'` | `RuntimeError: Expected one of cpu, cuda, ipu, xpu, ...` |
| `'CUDA'` | **`[]`** | `--policy.device=CUDA` | `'CUDA'` | `RuntimeError: Expected one of ...` |
| `'nvidia'` | **`[]`** | `--policy.device=nvidia` | `'nvidia'` | `RuntimeError: Expected one of ...` |
| `'cuda:abc'` | **`[]`** | `--policy.device=cuda:abc` | `'cuda:abc'` | `RuntimeError: Invalid device string` |
| `'  '` | **`[]`** | `--policy.device=  ` | `'  '` | `RuntimeError: Invalid device string` |

`'gpu'` is the ordinary mistake: it is the word the rest of the world uses, and it is not a torch device type.

The knob is reachable through the documented route - `create_trainer("lerobot_local", device="gpu").device` is `'gpu'`.

## Why this is the one knob without a domain

Every other caller-supplied value this same preflight sees is already held to one:

| value | gate |
|---|---|
| `steps`, `global_batch_size`, `lora_r`, `lora_alpha` | `_run_size_problems` |
| `learning_rate` | `_learning_rate_problems` |
| `seed` | `_seed_problems` |
| `num_gpus`, `num_nodes` | `_launch_topology_problems` |
| `val_episodes` | `_validation_episodes_problems` |
| `dataset_repo_id` | `_HUB_REPO_ID_RE` |
| paths / `extra` keys | `_security_problems` |
| **`device`** | **none** |

## Change

`validate` now extends its problems with `_device_problems()`, beside the other knob gates.

The admitted domain is torch's own, read by handing the value to `torch.device` rather than by comparing against a copied list of device types. A torch build that gains a backend is admitted here with no change, and torch's own exception enumerates the types it accepts, so the refusal names the admitted set without restating it. A test asserts no device-type vocabulary is copied into this module.

Two things are deliberately **not** graded, each with a control test:

- **availability.** `torch.device("cuda")` constructs on a CPU-only box, and a spec legitimately names a device the machine writing it does not have - a queued or containerised run is dispatched from one host and executed on another. A non-`str` is refused *before* torch is consulted for the same reason: `torch.device(0)` reads the accelerator inventory, which would make one spec validate on a GPU box and fail on a CPU box.
- **the falsy case.** `device or _auto_device()` predates this gate and still owns `None` and `""`, so neither reaches the domain check.

When torch is not importable the domain is unknown and the value passes through unguarded, which is the posture the reward-model field set already takes when its registry cannot be read.

The `Args:` entry for `device` now states the domain, as its numeric siblings' entries do.

### On stating the rule twice

`strands_robots.tools.lerobot_train` holds the same value to the same domain for its own detached argv. The two live in layers that cannot import each other - `strands_robots.tools` depends on the library, not the reverse - so a shared owner would have to be `strands_robots.utils`, whose refusal-guard scan is asserted against a frozen inventory (`found == KNOWN_MESSAGE_READS`); where a torch-domain rule belongs is a call for whoever owns that module's scope, not something to decide inside this fix.

Rather than an alias, the duplication is legitimised by a parity test: the trainer and the tool must reach the same verdict for every probe value. Pre-fix that test reports `device='gpu': the tool surface refuses it and the trainer admits it`.

## Tests

`tests/training/test_lerobot_trainer_device_domain.py`, **20 failed / 14 passed** against `upstream/main` with the file copied in, 34 pass after. 14 of the 20 are behavioural; 6 name the new private gate directly and are a root-cause pin rather than regression evidence.

The headline reports the whole chain in one message:

```
AssertionError: validate() returned [] for device='gpu', which torch cannot parse;
build_command still emits ['--policy.device=gpu'] and build_config assigns the same
string to policy_cfg.device, so the run aborts after the preflight called it launchable
```

The 14 that pass on both trees are the controls, and each fails for a specific over-tight fix:

- every usable spelling (`cpu`, `cuda`, `cuda:0`, `mps`) still validates clean - fails for an exact-match allowlist;
- `cuda` and `cuda:7` validate clean with `torch.cuda.is_available()` stubbed `False` - fails for a fix that grades availability;
- `None`, `""` and `0` still resolve through `_auto_device()` - fails for a fix that refuses a falsy device;
- `--policy.device=cpu` appears exactly once in the argv, unchanged - the no-overreach control;
- with `torch` unimportable the value passes through - fails for a fix that refuses every device on a dispatch-only host.

## Verification

Measured at `f17de5cf`.

- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `upstream/main` worktree, branch-only set empty, none in the changed files.
- Blast radius - all of `tests/training`, every test naming the trainer or the device gate, and the repo-wide docstring / changelog / string guards - run on both trees; numbers in the comment below.

No visual artifact: this changes a preflight verdict and adds no policy, simulation, rendering, recording or asset behaviour. The argv and the built config for every usable device are byte-identical, which is what the no-overreach control asserts.


cagataycali:
Blast-radius gate, both trees, same selection, run in parallel and in lockstep (134.27s vs 134.26s).

Selection: all of `tests/training`, every test naming `LerobotTrainer` / `create_trainer` / `_auto_device` / `policy.device` / the tool's device gate, and the repo-wide docstring, xref, string, dependency and changelog guards - 68 files.

```
branch    3 failed, 3539 passed, 6 skipped
base     23 failed, 3519 passed, 6 skipped
```

- collected identical on both: `3 + 3539 == 23 + 3519 == 3542`
- failed delta == passed delta == **20** == the pre-fix failure count
- **branch-only failing ids: empty**
- base-only: 20, every one from `tests/training/test_lerobot_trainer_device_domain.py`, zero foreign

The 3 shared failures are pre-existing and environmental, not related to this change - a `draccus` older than lerobot's `>=0.11.6` floor on the machine that ran the gate:

```
tests/training/test_lerobot.py::TestInProcessTrainerCorrectness::test_base_model_warm_start_reads_checkpoint_config
tests/training/test_lerobot.py::TestInProcessTrainerCorrectness::test_lora_base_model_does_not_preset_use_peft
tests/training/test_lerobot_e2e.py::test_record_train_load_loop
    TypeError: encode() takes 1 positional argument but 2 were given
```

They fail identically on both trees, which is what places them outside the diff.

Also re-run on the committed tree after the changelog fragment landed: `tests/test_changelog_fragments.py` + `tests/test_changelog_fragment_gate.py`, 85 passed, and `scripts/assemble_changelog.py --check` reports `changelog fragments OK (612 pending)`.

yinsong1986 (approved):
## Summary

Adds a `_device_problems()` gate to `LerobotTrainer.validate` so a `device` no torch build can parse (`'gpu'`, `'CUDA'`, `'cuda:abc'`, whitespace) is reported by the preflight instead of passing `validate()` and aborting the run after `build_command`/`build_config` carry the string into lerobot verbatim. The gate is a deliberate mirror of the tool-side `_torch_device_error` that already merged in #2560 (currently the tip of `main`): same non-`str` refusal before torch is consulted, same live `torch.device(...)` domain read with no copied device-type vocabulary, same pass-through posture when torch is unimportable, and a parity test pins the two layers to the same verdict on every probe value. Only the spelling is graded, never availability, with controls for each over-tight variant (availability stubbing, falsy pass-through via `_auto_device()`, unchanged argv for usable devices, torch-absent pass-through).

## What's good

- All 34 new tests pass on the PR head; the rest of `tests/training` is green apart from one pre-existing sim-recording e2e fixture error that reproduces identically on `upstream/main` (environment-dependent, unrelated to this diff).
- Changelog fragment present (`changelog.d/2563-lerobot-trainer-device-domain.md`), per AGENTS.md step 3; no host paths or non-ASCII in the diff; ruff check/format clean on the changed files.
- Scope discipline: the argv and built config for every usable device are byte-identical, pinned by the no-overreach control.
- The stated-twice duplication with `tools.lerobot_train` is legitimised by a cross-layer parity test rather than an alias that would invert the tool/library layering, and the docstring `Args:` entry for `device` now states the domain.

PR #2564, opened 2026-08-21
fix(policies): one owner for resolving a policy coroutine in a sync context

cagataycali:
## Problem

`strands_robots._async_utils` owns "resolve a policy coroutine in a sync context". With no running loop it is `asyncio.run`; inside one, the coroutine is offloaded to a single module-level worker that is reused across calls, for the reason its own comment gives: *"Module-level executor reused across calls to avoid creating threads at high frequency."* Every in-process rollout path resolves through it - `PolicyRunner` at three sites, and both backends' `run_multi_policy`.

Two things had drifted from that owner.

**The owner's branch guard covered the offload as well as the loop check.** `asyncio.get_running_loop()` reports "no running loop" *by raising `RuntimeError`*, so that class is the branch verdict. With the offload inside the same `try`, a `RuntimeError` raised by the awaited coroutine was read as that verdict: the fallback then called `asyncio.run` a second time, from inside the running loop and on an already-consumed coroutine. The caller received an unrelated asyncio message in place of what the policy said.

Measured over every builtin exception class a coroutine can raise (64 constructible), resolved from inside a running loop:

| raised by the policy | reaches the caller as |
|---|---|
| `ValueError`, `TimeoutError`, `KeyError`, ... (61 classes) | itself, faithfully |
| `RuntimeError("...")` | `RuntimeError: asyncio.run() cannot be called from a running event loop` |
| `NotImplementedError("...")` | same replacement |
| `RecursionError("...")` | same replacement |

The affected set is exactly the builtin `RuntimeError` family. `ValueError` - the class the neighbouring propagation test raises - was never affected, which is how the family could reach a caller renamed. `NotImplementedError` is the one that matters most: it is the MRO-contract violation a caller has to be able to see, and elsewhere in the package it is deliberately re-raised ahead of a broad handler for exactly that reason.

**`Policy.get_actions_sync` re-derived the same three-branch rule** with a private `ThreadPoolExecutor(max_workers=1)` per call. The wrapper is documented as safe to call from an event loop or a notebook, and a notebook cell body always runs inside a running loop - so that is the branch a documented caller lands in, at control rate. A `with` block joins its worker before returning, so the live thread *count* never grows and nothing watching thread counts can see the churn:

| 24 warm calls inside a running loop | OS threads started |
|---|---|
| via `get_actions_sync` | 24 |
| via the owner | 0 |

Interleaved wall time over 7 reps of 100 calls: median 179.1 us/call through the private executor vs 99.0 us/call through the owner.

## Change

`_resolve_coroutine`: the guarded call moves after the `except`, so the handler can only ever see the operation it is classifying. No behaviour changes for any class that already propagated.

`Policy.get_actions_sync`: delegates to the owner. `asyncio` and `concurrent.futures` are no longer imported by `policies/base.py`; `_async_utils` is stdlib-only, is already in `sys.modules` after a bare `import strands_robots`, and its executor spawns workers lazily, so the import costs nothing and starts no thread.

**The two halves are one change.** The wrapper's own report was already faithful - its `try` wrapped only the loop check - so delegating to the owner *without* narrowing the guard would have regressed it. That is how the guard defect surfaced: the wrapper's new exception-propagation test failed.

Also updated: the two AGENTS.md sections that state these rules (`### Exception Clauses Must Be Narrow` gains the try-scope bullet; `### Performance` names the existing owner and how to observe the churn), and the two test modules whose docstrings described the construct being replaced.

## Tests

`tests/test_async_utils.py::TestTheOffloadReportsWhatTheCoroutineRaised` and `tests/policies/test_base.py::TestSyncWrapperResolvesThroughTheSharedOwner`.

Against `upstream/main` with both test files copied in: **6 failed / 59 passed**, every failure behavioural and quoting the measured value -

- `RuntimeError raised inside the offloaded coroutine reached the caller as RuntimeError: asyncio.run() cannot be called from a running event loop -- the branch guard classified the coroutine's own failure as its 'no running loop' verdict` (and the same for `NotImplementedError`, `RecursionError`)
- `24 calls to get_actions_sync inside a running loop started 24 OS threads; the shared resolver in strands_robots._async_utils reuses one worker, so a warm wrapper must start none`
- `get_actions_sync did not resolve through strands_robots._async_utils._resolve_coroutine`
- `get_actions_sync constructs its own executor`

65 pass after.

The controls pass on both trees, and each fails for a specific wrong fix: an unrelated exception class still propagates (the guard was narrowed, not widened); the no-loop branch is still selected by the same verdict and still reports its own failures; the wrapper and the runner's resolver agree on the same policy in both loop contexts; and the observation, instruction and kwargs still reach the policy verbatim on both branches.

One existing test was named for the implementation being replaced (`..._uses_threadpool`); its stated intent - offload rather than raise "already in a loop" - is unchanged, so only the name and the module docstrings moved.

## Verification

Measured at `510f6559` (the changelog fragment is `17da5610`), on lerobot 0.6.2 / mujoco 3.12.0.

- Blast radius: 146 files (every test naming the wrapper, the owner, `policies.base` or AGENTS.md, plus `tests/policies/` and the repo-wide docs/string/changelog guards), run on both trees: branch **1 failed / 4648 passed / 29 skipped**, base **7 failed / 4642 passed / 29 skipped**, 4678 collected on both. The branch-only failure set is empty; base-only is exactly the 6 pre-fix failures above. The one shared failure (`test_rendering_pacers_survive_a_clock_step.py::...[forward_30s]`) is pre-existing and order-dependent: it passes 16/16 in isolation and fails identically on a pristine `upstream/main` worktree with these test files removed.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 against a pristine `upstream/main` worktree, none in the changed files.

## Artifact

A real MuJoCo so101 rollout driven through `Policy.get_actions_sync` from inside a running event loop - the state a notebook cell body is always in - with a policy that sweeps the arm for 40 ticks and then raises `NotImplementedError`. Same script, same scene; only `strands_robots` differs.

get_actions_sync resolution: report and thread churn

The rollout is **bit-identical in both trees** (per-frame mean `|pixel|` `0.00000` over all 5 sampled frames, 40 ticks applied either way), so what the panels differ on is only what is reported and what is spent: the runner's resolver goes from an unrelated asyncio message to the policy's own `NotImplementedError`, and the 46 wrapper calls go from 41 OS threads to 1 (the owner's worker, created once).


yinsong1986 (dismissed):
## Summary

This PR consolidates "resolve a policy coroutine in a sync context" into its single owner, `strands_robots._async_utils._resolve_coroutine`, and fixes a branch-guard defect in that owner: the offload call sat inside the same `try` as `asyncio.get_running_loop()`, so any `RuntimeError`-family exception raised by the awaited coroutine (`RuntimeError`, `NotImplementedError`, `RecursionError`) was misread as the "no running loop" verdict and replaced with `asyncio.run() cannot be called from a running event loop`. The guarded call now moves after the `except`, and `Policy.get_actions_sync` delegates to the owner instead of re-deriving the three-branch rule with a private per-call `ThreadPoolExecutor`. Verified locally at the head SHA: the 6 new tests fail behaviorally against the merge-base code with exactly the failure messages the description quotes, all 65 pass after; `ruff check` / `ruff format --check` clean on the four changed files.

## What's good

- The two halves genuinely are one change: delegating the wrapper without narrowing the guard would have regressed the wrapper's exception fidelity, and the tests pin both.
- Controls are thorough: unrelated exception classes still propagate (guard narrowed, not widened), the no-loop branch is unchanged, kwargs/observation forwarding is pinned verbatim on both branches, and the thread-churn pin counts `Thread.start` rather than live thread count for exactly the reason AGENTS.md now documents.
- AGENTS.md updates land in the right sections (`Exception Clauses Must Be Narrow`, `Performance`) with the tests named as pins; changelog fragment present per step 3; no new imports of `asyncio`/`concurrent.futures` in `policies/base.py`, and `_async_utils` is stdlib-only so no import-cost concern.

yinsong1986 (approved):
## Summary

This PR consolidates "resolve a policy coroutine in a sync context" into its single owner, `strands_robots._async_utils._resolve_coroutine`, and fixes a branch-guard scope defect in that owner: with the executor offload inside the same `try` as `asyncio.get_running_loop()`, any `RuntimeError`-family exception raised by the awaited coroutine (`RuntimeError`, `NotImplementedError`, `RecursionError`) was misread as the guard's "no running loop" verdict, and the caller received `asyncio.run() cannot be called from a running event loop` in place of the policy's own exception. The guarded call now sits after the `except`, so the handler classifies only the loop check. `Policy.get_actions_sync` drops its private per-call `ThreadPoolExecutor(max_workers=1)` and delegates to the owner, so the offload branch a notebook or async host lands in reuses the module-level worker instead of starting and joining one OS thread per call. Public API signature and documented contract of `get_actions_sync` are unchanged; both dispatch branches behave identically for every exception class that already propagated.

## What's good

- The two halves are correctly coupled: delegating without narrowing the guard would have regressed the wrapper's previously-faithful exception reporting, and the tests pin both directions (the `RuntimeError` family propagates unchanged, an unrelated class still propagates, the no-loop branch is still selected by the same verdict).
- The thread-churn pin counts `Thread.start` rather than the live thread count, which is the only observable that catches a `with ThreadPoolExecutor(...)` block joining before return.
- Changelog fragment present per AGENTS.md step 3; AGENTS.md's `### Performance` and `### Exception Clauses Must Be Narrow` sections updated in the same PR to name the owner and the try-scope rule.
- The prior `github-advanced-security` finding on the `except BaseException` handler was addressed at this head: the handler now re-raises lexically (`raise AssertionError(...) from unexpected`), the cleanup-and-reraise shape.
- Verified locally at `3d69bb27`: both changed test modules pass (65/65) and `ruff check` / `ruff format --check` are clean on all four changed Python files.
- No realistic nesting hazard from sharing the single-worker executor: the in-package `get_actions_sync` callers (`inference/server.py`, `PersistentPolicy`) run in sync contexts, and `PolicyRunner` already resolved through the same owner, so the wrapper inherits no exposure the runner did not already have.

PR #2565, opened 2026-08-21
fix(ci): report what the overlap sweep measured, not what it cannot see

cagataycali:
## What

`scripts/check_merge_base_overlap.py --all-open` intersects changed **paths**. A test that resolves its population from a filesystem walk is coupled to files it never names, so its intersection with the sibling it grades is empty and the sweep reports clean while the composition is in fact untested.

A clean run nonetheless printed:

```
No untested composition in the open set. Nothing here needs a merge-order decision.
```

That is a claim about **compositions**, which is strictly more than the relation measured, and a reader who trusts it stops looking. This PR scopes the report to its evidence and names the class it cannot describe.

## Measurement

#2557 added `tests/test_log_strings_are_ascii.py`, whose population is a walk of the package, and merged in a batch with #2559 and #2560. Both siblings added exactly the tool-result prose that grader scores. Pairwise path intersection:

```
2557 n 2559 = []
2557 n 2560 = []
```

Every signal read green and `Detect an untested overlap with the base branch` was `SUCCESS` on all three. The batch was safe, but only because it was checked by hand.

## Why not a new heuristic

The obvious fix -- widen the path set to the *walked root* -- was measured on the live open set of 9 pull requests and rejected:

```
rows emitted                     14
distinct pairs                   11  of 36
already reported by path overlap  2  (#1035+#1722, #2480+#2534)
new pairs added                   9
defects among them                0
```

The reason is structural. 125 of the 1136 test files in the tree resolve a population from a walk, and the ones that reach furthest are rooted at `strands_robots` entire, so they intersect nearly every open branch -- the two whole-package graders alone produce all 9 new pairs. A narrowly-rooted grader (`strands_robots/mesh/`) selects only the pair the path intersection already reports. A relation that fires on a third of the queue and names no defect is the `awaiting-first-review` failure mode from AGENTS.md step 8 in a different field.

What *does* separate them is composing the two branches and running the grader, which needs no model of what the grader reads. Run by hand over the open set for #2562's whole-tree grader it cost ~10 s per composition, and it correctly left alone three siblings whose new `except` tuples a path-or-keyword heuristic would have flagged -- including `(KeyboardInterrupt, Exception)`, which is correct code.

That relation cannot live in this mode. The sweep reads the open set from the API and **no checkout at all**, which is what lets a health report run it without a clone, and every sweep test pins that by running from a directory that is not a repository. So the change here is scope, not capability, and the composition run stays the manual step AGENTS.md documents.

## Changes

- `scripts/check_merge_base_overlap.py` -- the clean line now states the two intersections it actually ran; every report ends with an unconditional paragraph naming the composition class no relation in it can describe. New docstring section `What the path relation cannot see` carries the measurement.
- `tests/test_merge_base_overlap.py` -- the clean-report pin is retargeted off the over-claim, plus two new pins: a clean sweep may not claim there is no untested composition, and the caveat survives a populated report (it qualifies both).
- `AGENTS.md` -- the measurement recorded in the step that owns the sweep, so the next reader does not re-derive it from the issue thread.
- `changelog.d/2564-overlap-sweep-reports-what-it-measured.md`

## Verification

```
tests/test_merge_base_overlap.py                    49 passed
tests/test_composition_run_precondition.py
tests/test_merge_gate_viewer_scope.py
tests/test_except_tuples_state_their_real_scope.py
tests/test_pull_request_trigger_types.py            62 passed
ruff check / ruff format --check                     clean
non-ASCII in changed files                           none
```

The sweep was also re-run against the live open set with this change: 2 blocking pairs, 1 prose-only pair, 2 unevaluated, and the new caveat renders last.

## Scope

Report wording and its pins only. No relation is added, removed, or reweighted, and the exit status is unchanged -- a clean sweep still exits 0 and a pair still exits 1.

Refs #2561

## Round changelog

- **Round 0** -- initial submission.


yinsong1986 (approved):
## Summary

Scopes the `--all-open` sweep's report to its evidence: the clean line now states the two path intersections the sweep actually ran instead of claiming "no untested composition", and every report ends with an unconditional paragraph naming the composition class no path relation can see (a test whose population is a filesystem walk shares no path with the siblings it grades). The rejected alternative - widening the path set to the walked root - is documented with its measurement (11 of 36 pairs selected, none a defect) in the script docstring and AGENTS.md, so the next reader does not re-derive it. Exit statuses and relations are unchanged; the two new pins assert the clean report no longer over-claims and that the caveat survives a populated report. Verified on the head commit: `tests/test_merge_base_overlap.py` 49 passed, ruff check/format clean, no non-ASCII in the changed files, and the changelog fragment is correctly numbered `2565-...` per step 3.

## What's good

- The retargeted pin (`"shares a changed path"`) plus the negative assertion (`"No untested composition" not in report`) make the over-claim unable to return silently.
- The caveat is appended after `render_sweep`'s unconditional remedy paragraph, so it genuinely renders on both clean and populated reports - and the populated-report pin proves it.
- Scope discipline: wording and pins only, no relation added or reweighted, exit codes untouched, and the measurement that justifies rejecting the walked-root heuristic is recorded where the sweep is documented rather than only in the issue thread.

yinsong1986 (approved):
## Summary

Scopes the `--all-open` overlap sweep's report to the relations it actually computed: the clean line now names the two path intersections it ran instead of claiming "no untested composition", and an unconditional closing paragraph names the composition class no path relation can see (a walk-population grader shares no path with the siblings it grades - the #2557/#2559/#2560 batch). The rejected alternative (widening the path set to the walked root) is documented with its measurement in the script docstring and AGENTS.md, so the next reader does not re-derive it. Exit semantics are untouched - the diff changes only rendered strings inside `render_sweep`, and the caveat paragraph sits after the remedy on the single return path, so it reaches both clean and populated reports as the two new pins assert.

## What's good

- The retargeted pin (`"shares a changed path"`) plus the negative assertion (`"No untested composition" not in report`) mean the over-claim cannot quietly return.
- The old clean-line string has no other consumer in the tree (no workflow or script greps for it), so the wording change breaks nothing downstream.
- Changelog fragment present, non-ASCII sweep clean, ruff clean; `tests/test_merge_base_overlap.py` passes at head (49 passed).
- Scope discipline: no relation added, removed, or reweighted, and the measurement justifying the rejected heuristic is recorded where the sweep is owned (AGENTS.md step 8).

PR #2566, opened 2026-08-21
fix(simulation): name a refused field by a spelling the schema publishes

cagataycali:
## Problem

`_dispatch_action` rewrites `_FIELD_ALIASES` to method parameter names **before** validating, so every refusal downstream of that rewrite reports the parameter it validated rather than the field the caller wrote. One of those canonical names has no property of its own: `apply_force`'s torque is published as `torque_vec` and validated as `torque`, and `tool_spec.json` has no `torque` property at all.

Four refusals reached the caller naming it. Measured on a live world holding one body:

| sent | refusal on `main` | names a published property |
| --- | --- | --- |
| `torque_vec="up"` | `Parameter 'torque' must be a list of 3 numbers, got 8.` | no |
| `torque_vec=[1, 2]` | `Parameter 'torque' must be a list of 3 numbers, got 2.` | no |
| `torque_vec=[1, 2, "a"]` | `Parameter 'torque'[2] must be numeric, got str.` | no |
| an unknown field on `apply_force` | `Valid: ['body_name', 'force', 'point', 'torque']` | no |

A model constrained to this schema cannot emit `torque`, so it was told to fix a field it is not permitted to send, and the spelling that works went unmentioned. Its only route back was to guess that `torque` and `torque_vec` are one field.

The string-type check thirty lines above the vector one, in the same function, already resolved the spelling back to what the caller sent - its comment records why ("reporting the canonical parameter would send them looking for a key their payload does not contain"). `tests/simulation/mujoco/test_tool_spec.py` independently keeps a `_schema_name` helper for the same mapping, in order to grade the very table these refusals report from. What was missing was one owner for the rule, consulted by every site that names a field.

## Change

`_reported_param_name(param, field_aliases, received)` is that owner. It prefers, most specific first: the spelling the payload carries, then the alias for it that the payload carries, then the one the schema publishes. All five refusal sites in `_validate_and_build_kwargs` now go through it - the unknown-parameter `Valid:` list, the string-type check (which loses its inline copy of the same logic), the vector arity check, the per-component check, and the missing-required check.

`_PUBLISHED_PARAMS` is derived from `tool_spec.json`, for the same reason `_PUBLISHED_ACTIONS` and `_PUBLISHED_STRING_PARAMS` are: a second list would be a second thing to keep true.

**Scope.** The dispatcher is deliberately wider than the schema for Python callers - `stop_recording` accepts `bucket` / `run_id` without publishing either - so an unpublished name is not by itself a defect, and the rule leaves such a parameter under its own name. Only a parameter the schema reaches under **another** name is substituted. Sweeping all 77 actions that publish a `Valid:` list, **1 changed**:

```
apply_force: ['torque'] -> ['torque_vec']
stop_recording: ['bucket', 'output_path', 'push_to_hub', 'run_id']  (unchanged)
```

Narrowing the reporting rule to "name only schema properties" would have been the obvious wrong fix: it drops `bucket` / `run_id` from the list of what `stop_recording` accepts. `test_a_parameter_with_no_published_spelling_is_still_advertised` fails for it.

## Tests

`tests/simulation/mujoco/test_router_refusals_name_a_published_spelling.py`, 115 items. The sweeps are derived from `_FIELD_ALIASES` and from the schema, so an alias added later is graded without a second list.

| tree | failed | passed |
| --- | --- | --- |
| `upstream/main` + this test file | 5 | 110 |
| this branch | 0 | 115 |

The three behavioural failures name the offending spelling in the assertion (`AssertionError: Parameter 'torque' must be a list of 3 numbers, got 2.`); the fourth is the root-cause pin (`a refusal still formats the raw name '{vparam}'`) and the fifth is the helper's own unit test. The 110 both-tree passes are the controls: a vector with no alias keeps its own name, a Python caller who writes `torque` is still named by it, the string refusal keeps the behaviour it already had, the deliberate Python-only width survives, and no action advertises a substituted field it should not.

The bare-dispatch sweep is a special case. `sim(action=X)` with no other key is a refusal probe only for an action whose signature requires something: when every parameter carries a default the payload **binds**, and the method runs against the live world. Measured over the published enum, 53 of 77 bound and 36 reached a success status - `render`, `save_state`, `export_xml` among them, and two with effects outside the test process:

| action | dispatched bare | why it is not visible on CI |
| --- | --- | --- |
| `start_recording` | resolves `repo_id="local/sim_recording"` with no `root` and **creates** `$HF_LEROBOT_HOME/local/sim_recording/meta/info.json` (rule 15) | no dataset exists there yet on a fresh runner |
| `open_viewer` | reaches `mujoco.viewer.launch_passive` wherever `_is_headless()` is false - unconditionally on macOS/Windows, and on Linux whenever `DISPLAY` is set | runners set no `DISPLAY`, so `mujoco.viewer` is unimportable and the call is a clean error dict |

So it is parametrized over `_REQUIRED_PARAM_ACTIONS`, derived from the signatures the way `_VAR_KEYWORD_ACTIONS` already is - an action with a named parameter carrying no default. All 24 are refused at the router with a named required parameter and zero writes, and both skip branches become unreachable, so the sweep now asserts the refusal instead of skipping past it:

| sweep parametrized over | HF root after the run | pytest | wall |
| --- | --- | --- | --- |
| sweep over the published enum | `local/sim_recording/meta/info.json` created | 110 passed, **53 skipped** | 15.06 s |
| sweep over `_REQUIRED_PARAM_ACTIONS` | empty | **115 passed, 0 skipped** | 2.40 s |

`TestTheBareDispatchSweepCannotExecuteAnAction` keeps that true, and its two halves are complementary - neither alone catches both ways of losing it:

| regression | bind pin | `start_recording` pin | `open_viewer` pin | parametrize pin |
| sweep widened back to the enum (the set untouched) | pass | pass | pass | **fail** |
| derivation stops reading `p.default` | **fail** | **fail** | pass | pass |

`open_viewer` takes no parameters at all, so it stays outside the set even under a broken derivation - which is why the parametrize pin, read off the sweep's own `pytestmark`, is what protects it.

## Verification

Measured at `ac13fdbf`, blast radius = 280 files (all of `tests/simulation/mujoco`, every guard that walks a source or test tree, every test naming a recording root or `$HF_LEROBOT_HOME`, and the repo-wide docstring / string / changelog guards). Both trees run with `MUJOCO_GL=egl`, in parallel and in lockstep at 443 s.

| --- | --- | --- | --- |
| this branch | 0 | 8414 | 73 |
| the previous head `d92e403a` | 0 | 8409 | 126 |

The branch-only failure set is empty and the selection has no pre-existing failures. The counts reconcile exactly: 8487 collected against 8535, i.e. 53 sweep parameters dropped less 5 new guard items; passed `+5` is those items, and skipped `-53` is the bare-executing actions that used to be skipped past.

Regression evidence for the reporting rule itself, `upstream/main` + this test file: **5 failed / 110 passed**, the same five as before (the new guards grade the sweep's own shape, so they hold on `main` too). `ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` 24 == 24 against a pristine `upstream/main` worktree, nothing branch-only and nothing in the changed file.

## Rollout

Same scene, same request, same 676-step headless MuJoCo rollout in both columns - only which field the refusal names differs. A caller that follows the refusal it is given recovers on the right and is stuck on the left; the bar's yaw travel is **0.00 deg** against **36.77 deg**. Frame 0 is pixel-identical across trees (mean |px| = 0.000), so the two runs are the same rollout until the refusal is acted on.

refusal names a published spelling

### DIFF
--- a/tests/simulation/mujoco/test_router_refusals_name_a_published_spelling.py
+++ b/tests/simulation/mujoco/test_router_refusals_name_a_published_spelling.py
@@ -75,6 +75,31 @@ _VAR_KEYWORD_ACTIONS: frozenset[str] = frozenset(
 )
 
 
+def _required_parameters(action: str) -> list[str]:
+    """Named parameters of *action*'s method that carry no default."""
+    method = getattr(MuJoCoSimEngine, MuJoCoSimEngine._ACTION_ALIASES.get(action, action), None)
+    if method is None:  # pragma: no cover - every published action resolves today
+        return []
+    return [
+        name
+        for name, p in inspect.signature(method).parameters.items()
+        if name != "self"
+        and p.default is inspect.Parameter.empty
+        and p.kind not in (inspect.Parameter.VAR_KEYWORD, inspect.Parameter.VAR_POSITIONAL)
+    ]
+
+
+# Actions the router refuses before the method body runs, because the payload is
+# missing a parameter the signature requires. Only these may be dispatched bare:
+# an action whose parameters all carry defaults BINDS, so ``sim(action=X)``
+# executes it on the live world instead of producing the refusal under test --
+# which for ``start_recording`` means creating a dataset in the shared cache
+# (rule 15) and for ``open_viewer`` means reaching ``launch_passive``.
+_REQUIRED_PARAM_ACTIONS: frozenset[str] = frozenset(
+    action for action in _SPEC["properties"]["action"]["enum"] if _required_parameters(action)
+)
+
+
 @pytest.fixture
 def sim() -> Generator[Simulation, None, None]:
     """A world holding one body, so a refusal is the only thing under test."""
@@ -187,6 +212,77 @@ class TestAVectorRefusalNamesTheFieldTheCallerSent:
         assert "Parameter 'torque' must be a list of 3 numbers, got 2." in _text(result)
 
 
+class TestTheBareDispatchSweepCannotExecuteAnAction:
+    """A sweep that dispatches bare may only name actions the router refuses.
+
+    ``sim(action=X)`` with no other key is a refusal probe, but it is only a
+    probe for an action whose signature requires something: when every parameter
+    carries a default the payload BINDS, and the method runs against the live
+    world. Two of those reach outside the test process -- ``start_recording``
+    resolves ``repo_id="local/sim_recording"`` with no ``root`` and creates it
+    under ``$HF_LEROBOT_HOME`` (AGENTS.md rule 15, whose measured cost is one
+    planted dataset turning 133 passes into 22 ``FileExistsError`` failures in
+    unrelated modules), and ``open_viewer`` reaches
+    ``mujoco.viewer.launch_passive`` on any host that is not headless. Neither is
+    visible on CI, where ``DISPLAY`` is unset and no dataset exists yet.
+
+    So the filter is a property of the sweep, not a convenience, and these tests
+    fail if it is widened back to the published enum.
+    """
+
+    def test_no_graded_action_can_be_called_with_no_arguments(self) -> None:
+        """The mechanical form of "the router refuses it before the body runs".
+
+        Binding the signature with nothing but ``self`` is what
+        ``_dispatch_action`` effectively attempts for a bare payload; a method
+        that binds is a method that executes.
+        """
+        bindable = []
+        for action in sorted(_REQUIRED_PARAM_ACTIONS):
+            method = getattr(MuJoCoSimEngine, MuJoCoSimEngine._ACTION_ALIASES.get(action, action))
+            try:
+                inspect.signature(method).bind(object())
+            except TypeError:
+                continue
+            bindable.append(action)
+        assert not bindable, f"bare dispatch would execute these on the live world: {bindable}"
+
+    @pytest.mark.parametrize(
+        ("action", "effect"),
+        [
+            ("start_recording", "creates a dataset under $HF_LEROBOT_HOME (rule 15)"),
+            ("open_viewer", "reaches mujoco.viewer.launch_passive on a host with a display"),
+        ],
+    )
+    def test_an_action_with_a_side_effect_outside_the_process_is_not_swept(self, action: str, effect: str) -> None:
+        """Named individually, because these are the two the filter exists for."""
+        assert action in _SPEC["properties"]["action"]["enum"], f"premise: {action} is published"
+        assert action not in _REQUIRED_PARAM_ACTIONS, (
+            f"{action} would be dispatched bare by the sweep above, which {effect}"
+        )
+
+    def test_the_bare_dispatch_sweep_is_parametrized_over_the_filtered_set(self) -> None:
+        """Pins the decorator, not just the set it is derived from.
+
+        The two tests above grade what :data:`_REQUIRED_PARAM_ACTIONS` contains,
+        which stays true if the sweep is widened back to the published enum
+        beside them. This reads the sweep's own parametrization, so that edit is
+        what fails rather than the next developer's cache.
+        """
+        suite = TestARefusalAdvertisesOnlyPublishedSpellings
+        sweep = suite.test_a_missing_required_parameter_is_named_by_a_published_spelling
+        marks = [m for m in getattr(sweep, "pytestmark", []) if m.name == "parametrize"]
+        assert len(marks) == 1, marks
+        argnames, argvalues = marks[0].args
+        assert argnames == "action", argnames
+        assert sorted(argvalues) == sorted(_REQUIRED_PARAM_ACTIONS), sorted(argvalues)
+
+    def test_the_filter_still_grades_most_of_the_published_surface(self) -> None:
+        """Non-vacuity: a derivation that collapsed to nothing would report clean."""
+        assert len(_REQUIRED_PARAM_ACTIONS) >= 20, sorted(_REQUIRED_PARAM_ACTIONS)
+        assert _REQUIRED_PARAM_ACTIONS <= set(_SPEC["properties"]["action"]["enum"])
+
+
 class TestARefusalAdvertisesOnlyPublishedSpellings:
     """Whatever a refusal offers as a remedy, a model has to be able to emit it."""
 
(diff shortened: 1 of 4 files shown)


tests/simulation/mujoco/test_router_refusals_name_a_published_spelling.py
```diff
@@ -0,0 +1,291 @@
...
+
+    @pytest.mark.parametrize("action", sorted(_SPEC["properties"]["action"]["enum"]))
+    def test_a_missing_required_parameter_is_named_by_a_published_spelling(self, sim: Simulation, action: str) -> None:
+        """The required-parameter refusal reports a signature name too.
+
+        No required parameter is aliased away today, so this sweep passes either
+        way - it is here so that the day one is, the refusal names a field the
+        caller can supply rather than the day the report is read.
+        """
+        result = sim(action=action)
```
yinsong1986, line 337: [MUST FIX] `result = sim(action=action)` - this sweep is parametrized over the full published enum and dispatches every action bare, so the 53 actions whose parameters all carry defaults **execute on the live world** instead of being refused at the router. Two of them have effects outside the test process; both reproduce on this branch:

1. **`open_viewer` can kill the whole pytest run.** It reaches `mujoco.viewer.launch_passive`; on a host where `mujoco.viewer` imports but GLFW cannot initialize (e.g. `DISPLAY` set to a stale X forward - an ordinary ssh dev box), GLFW's failure path exits the interpreter with code 1. The suite dies mid-run (~76% through this file) with **no failure summary at all**, and `open_viewer`'s own `except Exception` never runs because the process is gone. On a host with a working display, a unit test opens a real interactive GL window instead. CI is green only because runners have no `DISPLAY`, which makes `mujoco.viewer` unimportable and the call a clean error dict - so the crash is invisible to every signal on this PR.

2. **`start_recording` writes to the developer's shared cache.** Bare dispatch executes it with its defaults (`repo_id="local/sim_recording"`, `root=None`), which resolves to and **creates** `$HF_LEROBOT_HOME/local/sim_recording` (`~/.cache/huggingface/lerobot/...`) - verified: the `meta/` directory appears after one run. This is exactly what AGENTS.md rule 15 ("A recording test names its own dataset root") forbids, including for call sites refused before the root is resolved - and this site is not even refused, it writes. Rule 15's own measurement is the cost being risked for the rest of the suite: one dataset planted in the shared cache turned 133 passes into 22 `FileExistsError` failures in unrelated tests. The AST pin in `tests/test_recording_root_is_not_the_shared_cache.py` cannot see this call site because the dispatch is dynamic (`sim(action=action)` names no `start_recording` attribute), so nothing downstream catches it.

Why this can't be deferred: once merged, every run of this file on a dev box with `lerobot` installed plants a dataset in the shared cache, and any box with a broken `DISPLAY` loses the entire test run with no report to bisect - a silent, environment-dependent suite killer shipped inside a test whose docstring says it "passes either way".

Resolution: restrict the parametrization to actions that actually carry a required parameter, derived from the method signatures exactly the way `_VAR_KEYWORD_ACTIONS` is derived above (any named param with `p.default is inspect.Parameter.empty`). Then `sim(action=action)` is always refused by the router before the method body runs, both side effects disappear, and the `status != "error"` skip branch becomes unnecessary. (Monkeypatching the viewer and passing a `tmp_path` root would also work, but the signature filter closes the whole class of bare-execution side effects, including any action added later.)

yinsong1986 (commented):
## Summary

This PR gives the router's refusal messages one owner for the field-spelling rule: `_reported_param_name` prefers the spelling the payload carries, then the alias the payload carries, then the schema's published spelling, and all five refusal sites in `_validate_and_build_kwargs` (unknown-parameter `Valid:` list, string-type, vector arity, per-component, missing-required) now consult it. The concrete defect it fixes is `apply_force`'s torque - published as `torque_vec`, validated as `torque`, with no `torque` property in the schema - so four refusals named a field a schema-constrained model cannot emit. The production change is correct and well-scoped: the helper is pure, `_PUBLISHED_PARAMS` is derived from `tool_spec.json` rather than restated, and the fallback deliberately preserves the dispatcher's Python-only width (`stop_recording`'s `bucket`/`run_id` stay advertised, verified by a dedicated control test). One blocker in the new test file is flagged inline.

## What's good

- The reporting rule has one owner now; the string-type check's inline copy of the same logic is deleted rather than left as a second spelling of the rule.
- Sweeps are derived from `_FIELD_ALIASES` and the schema, so a future alias is graded without a second list to keep true - and `TestTheAliasMapIsWhatMakesThisReachable` pins the premise so the sweeps cannot go silently vacuous.
- Controls cover both directions: a Python caller who writes `torque` is still named by it, and a vector with no alias keeps its own name.
- Changelog fragment named per convention; README updated in the same PR.

tests/simulation/mujoco/test_router_refusals_name_a_published_spelling.py
cagataycali: Confirmed and fixed in `ac13fdbf`, by the signature filter you suggested.

Both effects reproduce, and the counts match yours: over the published enum **53 of 77 bind** and **36 reach a success status** - `render`, `save_state`, `export_xml` among them. Driving the sweep's exact shape (a fresh fixture per action) with `HF_LEROBOT_HOME` pointed at an empty directory:

| | HF root after the run | pytest | wall |
| --- | --- | --- | --- |
| sweep over the published enum | `local/sim_recording/meta/info.json` **created** | 110 passed, 53 skipped | 15.06 s |
| sweep over `_REQUIRED_PARAM_ACTIONS` | empty | 115 passed, **0 skipped** | 2.40 s |

The `open_viewer` path is a little wider than the ssh case: `_is_headless()` returns `False` unconditionally on macOS and Windows, so the viewer import gate is open there regardless of `DISPLAY`. Verified the gate directly rather than calling `launch_passive` - with `DISPLAY=:99`, `_is_headless()` is `False` and `_mujoco_viewer is None` is `False`, i.e. `launch_passive` is reached; with `DISPLAY` unset both invert, which is the CI shape that hid it.

`_REQUIRED_PARAM_ACTIONS` is derived exactly as `_VAR_KEYWORD_ACTIONS` is. All **24** survivors are refused at the router with a named required parameter and zero writes, so both skip branches became unreachable and the sweep now asserts the refusal instead of skipping past it - which also means a future widening fails loudly rather than silently.

One thing your suggestion made me add. Grading only what the derived set *contains* leaves the sweep free to be widened back beside it, so `TestTheBareDispatchSweepCannotExecuteAnAction` also reads the sweep's own `pytestmark`. The two halves turn out to be complementary - neither alone catches both regressions:

| regression | bind pin | `start_recording` pin | `open_viewer` pin | parametrize pin |
| sweep widened back to the enum (set untouched) | pass | pass | pass | **fail** |
| derivation stops reading `p.default` | **fail** | **fail** | pass | pass |

`open_viewer` takes no parameters at all, so it stays outside the set even under a broken derivation; the parametrize pin is what protects it.

Blast radius re-run at 280 files against the previous head `d92e403a`: 0 failed either side, branch-only empty, and the counts reconcile exactly (8487 collected against 8535 = 53 parameters dropped less 5 new guard items; passed +5, skipped -53).

yinsong1986 (approved):
## Summary

This PR gives the router's refusal messages a single owner for which field spelling they name: `_reported_param_name` prefers the spelling the payload carries, then the alias the payload carries, then the schema's published spelling, falling back to the parameter's own name so the dispatcher's deliberate Python-only width (`stop_recording`'s `bucket`/`run_id`) is untouched. All five refusal sites in `_validate_and_build_kwargs` now consult it, fixing the one canonical name with no published property (`apply_force`'s `torque`, published as `torque_vec`) that four refusals previously echoed to schema-constrained callers who cannot emit it. `_PUBLISHED_PARAMS` is derived from `tool_spec.json` like its siblings, so there is no second list to keep true.

## What's good

- The concern from the prior review round (the bare-dispatch sweep binding and executing 53 defaulted actions, including `start_recording`'s shared-cache write and `open_viewer`'s `launch_passive` reach) is resolved at this head: the sweep is parametrized over `_REQUIRED_PARAM_ACTIONS`, derived from the signatures, and `TestTheBareDispatchSweepCannotExecuteAnAction` pins both the set's derivation and the sweep's own `pytestmark` so neither half can silently regress. Verified locally: all 115 items pass at `ac13fdbf` with `HF_LEROBOT_HOME` pointed at an empty directory, and the directory is still empty afterwards.
- The sweeps are derived from `_FIELD_ALIASES` and the schema rather than restated, so a future alias is graded without editing the test file, and the premise test (`test_at_least_one_canonical_parameter_name_is_unpublished`) is the one that flags when the sweeps go vacuous.
- The scope guard is exactly right: only a parameter the schema reaches under another name is substituted, and `test_a_parameter_with_no_published_spelling_is_still_advertised` fails for the obvious wrong fix that would have dropped `bucket`/`run_id` from `stop_recording`'s `Valid:` list.
- Changelog fragment present per AGENTS.md step 3; refusal text stays plain ASCII; the `_reported_param_name` docstring documents each argument under `Args:` (rule 13-adjacent hygiene even for a private helper).

PR #2568, opened 2026-08-21
fix(tools): confine a caller-named file to the directory that was validated

### LINKED ISSUE #2567: validate_save_path checks the directory, but the file is composed after it - so a traversal reaches the write through filename or format
## Summary

`validate_save_path` is the guard every filesystem-writing tool here runs on the directory it was told to write into. It refuses a `..` component explicitly. But the *file* a tool writes is composed afterwards, from separate caller-supplied parameters, and joined onto the directory the guard returned:

```python
save_path = validate_save_path(save_path, label="save_path")   # checked
file_path = os.path.join(save_path, f"{filename}.{format}")    # not checked
```

`os.path.join` walks back out of its first argument whenever the second asks it to. So the traversal `validate_save_path` refuses in `save_path` is reachable through `filename` or `format` instead -- the halves of the path nothing looks at. Both are agent-supplied `@tool` parameters with no domain of their own.

## Measurement

Composition only, on the resolved directory a passing `validate_save_path` returns:

```
validated save_path: /tmp/robots_probe/captures

 filename='shot' format='jpg'
   realpath : /tmp/robots_probe/captures/shot.jpg          escapes: False

 filename='../../../../tmp/robots_probe/escaped_via_filename' format='jpg'
   realpath : /tmp/robots_probe/escaped_via_filename.jpg   escapes: True

 filename='shot' format='jpg/../../../../tmp/robots_probe/escaped_via_format.jpg'
   realpath : /tmp/robots_probe/escaped_via_format.jpg     escapes: True
```

The second row is the ordinary shape. The third is the one worth noting: the stem is legitimate, and the escape is carried entirely by the extension -- a parameter whose documented role is to pick an image format.

## The sites

Every place a caller-supplied component is concatenated onto an already-validated directory:

| site | components that reach the path |
| --- | --- |
| `lerobot_camera._capture_single_image` | `filename`, `format` |
| `lerobot_camera._capture_batch_images` | `filename`, `format` |
| `lerobot_camera._record_video_sequence` | `filename` (the `.mp4` extension is fixed) |
| `pose_tool.PoseManager.__init__` | `robot_id` |

Two sites are deliberately **not** in that list, and both are already correct:

* `lerobot_camera._configure_camera_settings` composes its name from `camera_type`, which `_create_camera` refuses before the name is built, and from a `camera_id` whose separators are already stripped.
* `lerobot_calibrate.get_calibration_path` already guards this exact class, with a function-local allowlist and the comment "Security hardening: restrict path components to prevent traversal". So the rule was known here; it was applied at one of the three places that needed it.

## What the fix is not

The tempting fix is to reuse `lerobot_calibrate`'s character allowlist (`^[a-zA-Z0-9][a-zA-Z0-9._-]*$`). That would work, but it decides more than is needed and it collides with a decision already on record.

#2559 held `color_mode` and `rotation` to closed vocabularies and **deliberately left `format` out**, pinned in `test_lerobot_camera_vocabulary_options.py`:

> `format` is deliberately outside this domain. Unlike the two selectors, an unlisted `format` is *honored* rather than replaced -- it becomes the saved file's extension and OpenCV writes it if it can -- so narrowing it to the three names the docstring lists would refuse working requests.

That reasoning is sound and it is exactly what makes this issue real: `format` reaching the path is treated as the feature, and nothing confines it to a single component. An allowlist would also refuse `my capture.jpg`, which works today.

So the property to assert is narrower and exactly statable: **the resolved write target must lie inside the resolved directory.** That is orthogonal to which extensions are allowed, so it closes the traversal without reopening #2559's decision, and it needs no guess about which characters are dangerous on which platform. Resolving with `realpath` also means a symlink inside the directory cannot be used to step outside it.

## Suggested shape

A second helper beside `validate_save_path` in `strands_robots/tools/_path_validation.py` -- that module's stated job is the path guards every tool imports, and it currently owns only the directory half of a write target. One helper, four call sites, plus pins that a traversal is refused and that every value which worked before still works (`tiff` and a filename with a space included).

cagataycali:
## What

`validate_save_path` guards the directory a tool writes into. The *file* is composed afterwards, from separate caller-supplied parameters, and joined onto the directory that guard returned:

```python
save_path = validate_save_path(save_path, label="save_path")   # checked
file_path = os.path.join(save_path, f"{filename}.{format}")    # not checked
```

`os.path.join` walks back out of its first argument whenever the second asks it to, so the `..` traversal that guard refuses in `save_path` was reachable through `filename` or `format` instead -- the halves of the path nothing looked at -- and the write reported `status="success"` at whatever location it had reached.

## Measurement

Composition only, on the resolved directory a *passing* `validate_save_path` returns:

```
validated save_path: /tmp/robots_probe/captures

 filename='shot' format='jpg'
   realpath : /tmp/robots_probe/captures/shot.jpg          escapes: False

 filename='../../../../tmp/robots_probe/escaped_via_filename' format='jpg'
   realpath : /tmp/robots_probe/escaped_via_filename.jpg   escapes: True

 filename='shot' format='jpg/../../../../tmp/robots_probe/escaped_via_format.jpg'
   realpath : /tmp/robots_probe/escaped_via_format.jpg     escapes: True
```

The third row is the one worth noting: the stem is legitimate and the escape is carried entirely by the extension -- a parameter whose documented role is to pick an image format.

## Why containment, and not the allowlist already in the tree

`lerobot_calibrate.get_calibration_path` already guards this exact class with a function-local `^[a-zA-Z0-9][a-zA-Z0-9._-]*$` and the comment "Security hardening: restrict path components to prevent traversal". Reusing it would have been the smaller-looking change, and it was measured and rejected, because it decides more than is needed and collides with a decision already on record.

#2559 held `color_mode` and `rotation` to closed vocabularies and **deliberately left `format` out**, pinned in `test_lerobot_camera_vocabulary_options.py`:

> `format` is deliberately outside this domain. Unlike the two selectors, an unlisted `format` is *honored* rather than replaced -- it becomes the saved file's extension and OpenCV writes it if it can -- so narrowing it to the three names the docstring lists would refuse working requests.

That reasoning is sound, and it is what makes this defect possible: `format` reaching the path is the feature, and nothing confined it to a single component. An allowlist would also have refused `my capture.jpg`, which works today.

So what is asserted is narrower and exactly statable -- **the resolved write target lies inside the resolved directory** -- checked after resolution, on the composed path. That is orthogonal to which extensions are allowed, so it closes the traversal without reopening #2559's decision, and it needs no guess about which characters are dangerous on which platform. Resolving with `realpath` also means a symlink inside the directory cannot be used to step outside it.

A name that stays inside is accepted whatever it is spelled with, including one naming a subdirectory (`a/b.jpg`). That is not an oversight: it is contained, and if the subdirectory does not exist the write fails loudly rather than landing somewhere else.

## Changes

- `strands_robots/tools/_path_validation.py` -- new `resolve_output_path(directory, name, *, label)`. Added here because this module's stated job is the path guards every tool imports, and it owned only the directory half of a write target. The out-of-bounds and names-the-directory-itself cases are reported separately, because the first message would otherwise have to say a path inside the directory is outside it.
- `strands_robots/tools/lerobot_camera.py` -- the three composition sites (`capture`, `capture_batch`, `record`) resolve through it; `filename` and `format` docstrings state the refusal.
- `strands_robots/tools/pose_tool.py` -- `PoseManager` resolves its pose file through it. `PoseManager` is constructed *outside* the tool's own `try`, so the refusal is caught explicitly there -- otherwise it would leave the tool as an exception instead of the error envelope every other refusal in it uses.
- `tests/tools/test_output_path_is_confined_to_its_directory.py`
- `changelog.d/2567-output-path-confined-to-validated-directory.md`

## Scope

Two sites are deliberately **not** touched, both already correct:

- `_configure_camera_settings` composes its name from `camera_type`, which `_create_camera` refuses before the name is built, and a `camera_id` whose separators are already stripped.
- `lerobot_calibrate.get_calibration_path` is not defective, so its local allowlist is left alone rather than migrated for symmetry.

No vocabulary is narrowed, added or reweighted, and no accepted value becomes refused except one that resolves outside its directory.

## Verification

```
tests/tools/test_output_path_is_confined_to_its_directory.py    32 passed
tests/tools/test_path_validation.py
tests/tools/test_lerobot_camera.py
tests/tools/test_lerobot_camera_vocabulary_options.py
tests/tools/test_lerobot_camera_numeric_options.py
tests/tools/test_lerobot_camera_async_read_budget.py
tests/tools/test_pose_tool*.py
tests/tools/test_lerobot_calibrate.py                         652 passed
tests/tools/test_agent_tool_parameter_descriptions.py
tests/tools/test_public_member_docstrings.py
tests/tools/test_docstring_module_xrefs.py
tests/test_except_tuples_state_their_real_scope.py              77 passed
ruff check / ruff format --check                              clean
mypy on the changed files, tests included                    no errors
non-ASCII in changed files                                   none
```

The local environment could not resolve lerobot 0.6's CUDA wheel set, so `lerobot.cameras` was supplied by a stub. Five tests in `test_lerobot_camera.py` depend on the real package (`OpenCVCamera.find_cameras`, the RealSense config dataclass) and fail under it. They fail **identically on a pristine `main` with the same stub**, so they are an artifact of that environment rather than of this change; CI exercises them against the real dependency.

Both halves are pinned: a traversal is refused at each site and nothing lands outside the directory (asserted on the filesystem, not only on the message), and every value that worked before still works -- `jpg`, `png`, `bmp`, `tiff`, and a filename containing a space.

Closes #2567

## Round changelog

- **Round 0** -- initial submission. The first push failed `hatch run lint`: mypy rejected a `cv2.imwrite` stand-in written as `lambda path, img: written.append(path) or True`, because `list.append` returns `None`. Replaced with a named, typed helper, and mypy re-run over the test file as well as the source -- which is what the table above now reports, and what the local run had checked only for the source. That gate runs before the test step, so no test result had been produced yet. No behaviour in the change itself was affected, and the branch is still a single commit.

### DIFF
--- a/changelog.d/2567-output-path-confined-to-validated-directory.md
+++ b/changelog.d/2567-output-path-confined-to-validated-directory.md
@@ -1,0 +1,16 @@
+### Fixed: a caller-named capture or pose file cannot be written outside the directory that was validated
+
+`validate_save_path` guards the directory a tool writes into, but the file was
+composed afterwards from separate caller-supplied parameters and joined onto it,
+and `os.path.join` walks back out of its first argument whenever the second asks
+it to. So the `..` traversal that guard refuses in `save_path` was reachable
+through `lerobot_camera`'s `filename` or `format`, or `pose_tool`'s `robot_id` --
+the halves of the path nothing checked -- and the write reported success at
+whatever location it had reached.
+
+A new `resolve_output_path` helper beside `validate_save_path` resolves the
+composed name and refuses one that does not land inside the resolved directory,
+which also means a symlink inside it cannot be used to step outside. It is a
+containment check rather than a character allowlist, so every name that worked
+before still works: an unlisted extension such as `tiff` stays honored, per the
+decision recorded in #2559, and so does a filename containing a space.
--- a/strands_robots/tools/_path_validation.py
+++ b/strands_robots/tools/_path_validation.py
@@ -1,7 +1,15 @@
 """Shared path validation utilities for tools that write to the filesystem.
 
-Provides a consistent ``validate_save_path`` helper that all tool modules
-can import to reject dangerous path values before any I/O occurs.
+Provides two helpers all tool modules can import to reject dangerous path values
+before any I/O occurs, one per half of a write target:
+
+* :func:`validate_save_path` validates the *directory* a tool writes into.
+* :func:`resolve_output_path` resolves the *file name* a caller asks for inside
+  that directory, and refuses a name that leaves it.
+
+Both are needed because a tool composes its write target from the two, and
+validating only the directory leaves the composition unchecked - the defect
+:func:`resolve_output_path` exists to close.
 
 Cross-platform: blocks sensitive directories on Linux, macOS, and Windows.
 """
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Closes the path-traversal gap between `validate_save_path` (which guards the directory) and the file name composed afterwards from separate caller-supplied parameters (`filename`/`format` in `lerobot_camera`, `robot_id` in `pose_tool`). The new `resolve_output_path` helper asserts the one property actually required — the resolved write target lies inside the resolved directory, checked after `realpath` resolution on the composed path — rather than a character allowlist, which keeps #2559's decision intact (`tiff` and a filename with a space still work, pinned by tests). Verified locally at the head SHA: the new suite passes 32/32, and `test_pose_tool.py` + `test_path_validation.py` pass 107/107.

## What's good

- Both halves are pinned: escaping names refused at every composition site (including the filesystem-level assertion that nothing lands outside), and every previously-working value still works.
- All three camera composition sites sit inside `try/except` blocks that return the standard error envelope, so the `ValueError` refusal surfaces as `status="error"` rather than an exception; the `pose_tool` construction site is wrapped explicitly with a narrow `except ValueError` for the same reason.
- `commonpath` over `startswith` (sibling-prefix case tested), symlink escape tested, absolute-name discard by `os.path.join` tested.
- No public API one-way door: the helper lives in the `_`-private module, tool parameter names are unchanged, and the only newly-refused values are the escaping ones. Changelog fragment present; `Args:` entries updated for every affected tool parameter per AGENTS.md rule 13.
- Scope discipline: `_configure_camera_settings` and `lerobot_calibrate.get_calibration_path` correctly left alone, with the reasoning recorded.

PR #2570, opened 2026-08-21
fix(ci): refuse a local-checkout flag in the mode that reads no checkout

### LINKED ISSUE #2569: ci(gates): --repo means owner/name in one gate script and a local path in its sibling, so the overlap sweep reported a clean open set for the wrong repository
## Summary

Two gate scripts in `scripts/` take a flag spelled `--repo` and mean different things by it. One means `owner/name` and *refuses* to infer it; the other means a local directory and silently infers the target repository from `$GITHUB_REPOSITORY`.

```
scripts/check_duplicate_claim.py     --repo strands-labs/robots     # owner/name, inference refused
scripts/check_merge_base_overlap.py --repo <path>                  # "repository root (default: current directory)"
```

So the same flag, spelled the same way, passed the same value, is correct for one and a silent no-op for the other.

## How it presented

Hit while running AGENTS.md step 12's health sweep from a scheduled agent, whose checkout is *not* this repository:

```
$ python3 scripts/check_merge_base_overlap.py --repo strands-labs/robots --all-open

`cagataycali/strands-gtc-nvidia`, base `main`: 1 open non-draft pull request(s), 0 pair(s) compared.

No pair in the open set shares a changed path [...] Nothing here needs a merge-order decision.
```

Exit `0`. The value was accepted as a filesystem path, the API read went to `$GITHUB_REPOSITORY`, and the report described a repository nobody asked about. Re-run with `GITHUB_REPOSITORY=strands-labs/robots`, the same command reports **11 open pull requests, 55 pairs, 2 blocking pairs, 1 prose-only pair, 2 unevaluated**.

So the failure mode is a clean report for the wrong repository -- a wrong answer shaped exactly like the right one. The report does name the repository it read on its first line, which is what made it noticeable, but only to a reader who was already suspicious of the `1 open pull request` count.

## Why this is the same defect AGENTS.md already documents

Step 1 records exactly this hazard for the *other* script, and it is why that one refuses inference:

> Name the repository rather than leaving it to be inferred. `$GITHUB_REPOSITORY` is where the command is *running*, which for a scheduled agent need not be a checkout of this one, and an intake check that reads a different repository's open pull requests reports `unique-claim` and exits `0` -- a wrong answer shaped exactly like the right one.

The remedy was applied to `check_duplicate_claim.py` and the reasoning generalises to every mode that reads the *open set* from the API rather than one pull request. `--all-open` is such a mode, and AGENTS.md step 12 requires it of a scheduled agent -- the one caller for which the inference is least likely to be right.

## Why it is an issue and not a patch

The fix is a naming decision, and there are two defensible ones with different blast radii:

1. **Rename the local-path flag** (`--repo-root`, say) and give `--repo` the `owner/name` meaning it has in the sibling, refusing inference in `--all-open` mode. Consistent, and it makes the two scripts teach the same vocabulary -- but it is a breaking flag change for any existing caller, and the workflows that invoke it would need auditing.
2. **Leave the flag alone and refuse inference in `--all-open` only**, requiring the repository to be named explicitly through whatever spelling that mode ends up with. Narrower, no rename, but it leaves two sibling scripts disagreeing about what `--repo` means, which is the thing that produced the wrong report.

Either way the `--all-open` mode should refuse an inferred repository rather than report on it, matching the sibling. Worth noting alongside: the single-pull-request mode's default is *correct* to infer, for the reason step 1 gives -- a workflow reviewing a pull request runs where that pull request lives -- so this is specifically about the open-set mode.

## Related

#2561 is about a different limitation of the same script (what the path relation cannot see). This one is about which repository it reads, and the two are independent.

cagataycali:
`--all-open` sweeps the open set from the API. Every value it reads is handed to `_run_sweep`: `--github-repo`, `--base-ref`, `--token`. `--head` and `--repo` name local git inputs it reads none of, and the script already refuses `--head` for exactly that reason:

> Mutually exclusive rather than ignored: `--head` names one commit and the sweep reads none, so honouring both would answer a question the caller did not ask while looking like it had.

`--repo` was silently accepted. It is also the flag a caller reaches for, because the four sibling gate scripts spell `owner/name` `--repo` (`check_duplicate_claim`, `check_pr_head_is_current`, `check_last_push_approval`, `check_closing_reference`). Accepted here as a filesystem path it named nothing: the sweep read `$GITHUB_REPOSITORY` instead, and the report described whatever repository the command happened to be running in.

Measured on `d464c04`, run from a checkout of a fork:

| command | exit | report |
| --- | --- | --- |
| `--all-open --repo strands-labs/robots` | **0** | `cagataycali/robots`, base `main`: **0 open** non-draft pull request(s), 0 pair(s) compared |
| `--all-open --github-repo strands-labs/robots` | 1 | `strands-labs/robots`, base `main`: **10 open** non-draft pull request(s), 45 pair(s) compared |

A clean report for a repository nobody asked about, exiting `0` - which is the harm AGENTS.md step 1 records for the sibling intake check ("a wrong answer shaped exactly like the right one"), and the reason that one refuses an inferred repository. `--github-repo`'s own docstring already anticipates the ambiguity ("one flag that means a filesystem path in one mode and a slug in the other is the kind of ambiguity a caller discovers by getting a wrong answer") - the flag was separated, the refusal was not added.

## Change

Extend the existing mutual exclusion to `--repo`, and name `--github-repo` in the refusal: a caller who passed `--repo` wanted the repository named, and that is the flag that names it.

```
error: --all-open sweeps the open set from the API and reads no local checkout; --repo names one.
       To name the repository the sweep reads, pass --github-repo owner/name.
```

No rename, and no change to what either flag means, so the single-branch path and the CI invocation (`--base-ref ... --head ...`, no `--repo`) are untouched. The `--repo` docstring entry now says it is single-branch only.

The test derives the partition from the source rather than listing it - every value-bearing flag outside `_run_sweep`'s arguments must be refused - so a third local-only flag is graded on arrival instead of silently ignored. Derived today: the sweep reads `{base_ref, github_repo, token}` and is handed none of `{head, repo}`.

## Verification

Measured at `cd46bb1`.

Pre-fix split, the new tests against `d464c04`: **2 failed / 54 passed** -> 56 passed. Both failures name what happened rather than a missing raise:

```
AssertionError: argparse accepted --all-open --repo strands-labs/robots --github-repo owner/name
--token t and the run went ahead (exit 1): the sweep is handed none of that flag, so the value
named nothing and the report describes whatever $GITHUB_REPOSITORY names instead
```

The five tests that pass on both trees are the boundary, and each fails for a specific wrong fix: `--base-ref` is handed to the sweep so naming it is **not** refused; the single-branch path still reads `--repo`; `--head`'s refusal stays byte-identical (collapsing the two messages would lose the `--github-repo` pointer); and the derivation is pinned non-vacuously in both directions.

Blast radius, 117 files - every test naming a gate script or `AGENTS.md`, every guard that walks the test tree or grades sources, plus the repo-wide guards:

| this branch | 0 | 4138 | 82 | 4220 |
| `d464c04` + the new tests | 2 | 4136 | 82 | 4220 |

Branch-only failures: none. Base-only: exactly the two pre-fix failures above, no others. No failure is shared by the two trees. `mypy strands_robots tests tests_integ` is 24 errors on both, none in the changed files; `ruff check` and `ruff format --check` clean.

## The report the two spellings produce

Same command, same environment, both trees. The control is the second block of each panel: the documented spelling is byte-identical on both trees, so what changes is only which spelling is allowed to describe a run that is not happening.

--all-open refuses a local-checkout flag

Closes #2569.

### Scope

The issue also asked whether `--repo` should be renamed so the two scripts teach one vocabulary. That question is already answered in this script and against the rename: `--github-repo` exists precisely so the slug meaning has its own spelling, with the reasoning in the docstring. Renaming `--repo` would break every existing caller of the single-branch path to restate a decision already made, so this change adds the refusal the separation was missing and leaves both flags' meanings alone.

### DIFF
--- a/changelog.d/2570-all-open-refuses-a-local-checkout-flag.md
+++ b/changelog.d/2570-all-open-refuses-a-local-checkout-flag.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- **ci**: `check_merge_base_overlap.py --all-open` now refuses `--repo` instead of
+  silently accepting it. The sweep reads the open set from the API and is handed
+  none of `--repo`, so a caller who named the repository that way - the spelling
+  the four sibling gate scripts use for `owner/name` - had it read as a filesystem
+  path while the sweep read `$GITHUB_REPOSITORY`, reporting cleanly and exiting `0`
+  for a repository nobody asked about. The refusal names `--github-repo`, which is
+  the flag that does name the repository the sweep reads. `--head` was already
+  refused for the same reason; neither flag's meaning changes, so the
+  single-branch path and the CI invocation are untouched.
--- a/scripts/check_merge_base_overlap.py
+++ b/scripts/check_merge_base_overlap.py
@@ -156,10 +156,17 @@ Usage
                 landed does not carry that gate's script (issue #1791). Sound
                 because every input below is read from the object database and
                 never from the working tree.
-``--repo``      repository root (default: the current working directory).
+``--repo``      repository root (default: the current working directory). One
+                branch only: the sweep reads no checkout, so ``--all-open``
+                refuses it rather than ignoring it. The sibling gate scripts
+                spell ``owner/name`` ``--repo``, so a caller who reaches for it
+                here names a path, the sweep reads ``$GITHUB_REPOSITORY``
+                instead, and the report describes a repository nobody asked
+                about (issue #2569).
 
 ``--all-open``  Sweep the open set instead of one branch (see above). Mutually
-                exclusive with ``--head``, which names one commit.
+                exclusive with ``--head`` and ``--repo``, which name a local
+                commit and a local checkout.
 ``--github-repo``
                 ``owner/name`` for ``--all-open`` (default:
                 ``$GITHUB_REPOSITORY``). Deliberately not ``--repo``: in this
@@ -818,11 +825,21 @@ def main(argv: Sequence[str] | None = None) -> int:
     )
     args = parser.parse_args(argv)
 
-    # Mutually exclusive rather than ignored: --head names one commit and the
-    # sweep reads none, so honouring both would answer a question the caller did
-    # not ask while looking like it had.
+    # Mutually exclusive rather than ignored: --head names one commit and --repo
+    # one checkout, and the sweep reads neither, so honouring either would answer
+    # a question the caller did not ask while looking like it had. --repo is the
+    # one a caller is most likely to pass by mistake, because the sibling gate
+    # scripts spell owner/name --repo: accepted as a path, it leaves the sweep
+    # reading $GITHUB_REPOSITORY and reporting on a repository nobody asked about
+    # (issue #2569). Every value-bearing flag the sweep does read is passed to
+    # _run_sweep below, which is what pins this partition rather than a list.
     if args.all_open and args.head is not None:
         parser.error("--all-open sweeps the open set and reads no local commit; --head names one branch")
+    if args.all_open and args.repo is not None:
+        parser.error(
+            "--all-open sweeps the open set from the API and reads no local checkout; --repo names one. "
+            "To name the repository the sweep reads, pass --github-repo owner/name."
+        )
     if args.all_open:
         if not args.github_repo:
             parser.error("--all-open needs --github-repo owner/name (or $GITHUB_REPOSITORY)")
--- a/tests/test_merge_base_overlap.py
+++ b/tests/test_merge_base_overlap.py
@@ -25,7 +25,9 @@ CI, pinned by ``test_a_merge_commit_head_defeats_the_check``.
 
 from __future__ import annotations
 
+import ast
 import importlib.util
+import inspect
 import subprocess
 import sys
 from collections.abc import Callable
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

Extends the existing `--all-open` / `--head` mutual exclusion in `scripts/check_merge_base_overlap.py` to `--repo`, closing the gap where a caller reaching for the sibling gate scripts' `--repo owner/name` spelling got a clean exit-0 report about whatever repository `$GITHUB_REPOSITORY` happened to name -- the same wrong-answer-shaped-like-the-right-one class AGENTS.md step 1 documents for the intake check. The refusal points at `--github-repo`, which is the flag the mistaken caller actually wanted. `--repo` defaults to `None`, so the new guard fires only when the flag is explicitly passed; the single-branch path (`repo = Path(args.repo) if args.repo is not None else None`) and the CI invocation in `merge-base-overlap.yml` (`--base-ref ... --head ...`, no `--repo`) are untouched. This is a tightening of a previously-meaningless flag combination into an explicit error, consistent with AGENTS.md's no-silent-defaults rule, not a behaviour change for any documented caller. The follow-up commit adds the `changelog.d/2570-*` fragment per AGENTS.md step 3.

## What's good

- The regression test derives the local-vs-sweep flag partition from `main`'s own source (`add_argument` calls vs `_run_sweep`'s arguments), so a third local-only flag added later is graded on arrival rather than listed by hand -- and non-vacuity of the derivation is pinned in both directions.
- Boundary tests pin the specific wrong fixes: `--base-ref` (read by both modes) stays accepted, the single-branch `--repo` path still works, and the `--head` refusal message stays byte-identical so the two refusals keep their distinct remedies.
- `_refusal` asserts only the text after `error:`, deliberately avoiding the usage line that names every flag -- the assertion cannot pass vacuously.
- Verified locally at the code head (`aaa9f4c1`; `edc8a7f9` adds only the changelog fragment): all 56 tests in `tests/test_merge_base_overlap.py` pass.

yinsong1986 (approved):
## Summary

Extends `check_merge_base_overlap.py`'s existing `--all-open` / `--head` mutual exclusion to `--repo`, so a caller who reaches for the sibling gate scripts' `--repo owner/name` spelling gets a loud argparse refusal naming `--github-repo` instead of a clean exit-0 report about whatever repository `$GITHUB_REPOSITORY` happens to name. Verified independently at head `edc8a7f9`: `--repo` defaults to `None`, so the refusal fires only when the flag is explicitly passed and the sweep's default invocation is unchanged; the CI invocation in `.github/workflows/merge-base-overlap.yml` (`--base-ref ... --head ...`, no `--all-open`, no `--repo`) is untouched; the full test file passes 56/56 on this branch, and swapping in the `d464c04` script reproduces exactly the two claimed pre-fix failures (both naming `--repo`) with no others; ruff check and format are clean on both changed files. The partition-derivation test pins `_run_sweep`'s inputs from the source with a non-vacuity check on both sides, so a future local-only flag is graded on arrival. Changelog fragment present, no flag renamed, single-branch path behaviour unchanged. No blocking concerns.

## What's good

- The refusal converts a silent wrong answer (exit 0 for a repository nobody asked about) into a loud error that names the correct flag - the AGENTS.md "no silent defaults on error" shape, applied to the same wrong-answer class step 1 records for the sibling intake check.
- The two refusal messages stay distinct, and `test_the_head_refusal_is_unchanged` pins the `--head` message byte-for-byte so the `--github-repo` pointer cannot be lost to a later message merge.
- Boundary tests fail for specific wrong fixes: `--base-ref` (handed to the sweep) is not refused, and the single-branch path still reads `--repo`.

Issue #1919, opened 2026-08-03
process(codeql): py/catch-base-exception is the one alert class none of the three documented dispositions can clear - it is open on main untouched since 2026-07-07 and simultaneously blocks #1899 on two threads deferred to a human


cagataycali:
AGENTS.md > Review Learnings (PR #92) > CI Security Baseline says an alert has three clearing tools: **fix it**, **dismiss it with a reason** when the construct is *deliberate and test-only*, or **filter the rule** when *every* instance in the tree is an obliged idiom. It also warns that rewriting the flagged code to satisfy the query is the tempting fourth option and the one that costs (#1879).

`py/catch-base-exception` falls through all three, and the result is not a stalled judgment call in the abstract - it is #1899 parked for 16 hours on two threads whose text is `please resolve this thread if you agree - or say the word and I'll narrow it`.

## The same construct is advisory on main and a merge gate in a PR

| where | alert | path | created | effect |
|---|---|---|---|---|
| `refs/heads/main` | #691 | `strands_robots/simulation/isaac/simulation.py:5125` | 2026-07-07 | open, note severity, gates nothing, 27 days untouched |
| #1899 branch | #853 | `examples/libero/run_isaac_agent.py:867` | 2026-08-03 | opened a review thread -> blocks the merge |
| #1899 branch | #854 | `tests/simulation/isaac/test_isaac_backend.py:522` | 2026-08-03 | opened a review thread -> blocks the merge |

Alert #691 is `IsaacSimulation.run_on_main`'s cross-thread exception box, in the *same file* #1899 edits, and it already carries `# noqa: BLE001 - surfaced to caller below`. #853 and #854 are the identical idiom. So the tree ships the construct on `main` indefinitely and blocks a PR for introducing it, with nothing recording why the two answers differ.

## The rule's whole alert surface in this tree is one construct

Eight `except BaseException` handlers exist. Exactly one is flagged, and the correlation is total: the query accepts a handler that re-raises lexically.

| site | handler ends in | flagged |
|---|---|---|
| `strands_robots/robot.py:368` | `sim.destroy()`, bare `raise` | no |
| `strands_robots/policies/persistent.py:193` | `handoff.abandon()`, bare `raise` | no |
| `strands_robots/simulation/safe_output.py:185` | `os.unlink(tmp)`, bare `raise` | no |
| `strands_robots/hardware_robot.py:1865` | `self._release_task()`, bare `raise` | no |
[...]

PR #2572, opened 2026-08-21
docs(agents): derive the except-BaseException census from the tree

cagataycali:
## Problem

`AGENTS.md`'s `py/catch-base-exception` passage argues from a census of the tree: every
`except BaseException` handler re-raises *lexically*, which the query accepts, **except** the
cross-thread exception-marshal box. That claim is what makes the section's disposition
exhaustive rather than a judgment call - dismiss the box, because `concurrent.futures` cannot
target an already-running foreign thread; delegate everything else - and #1919 records two
rounds spent arguing the idiom before anyone wrote it down.

The claim is still true. The census it rests on was not:

| | passage said | tree says |
| --- | --- | --- |
| handlers in the tree | seven | **16** |
| handlers in `strands_robots/` listed | 4 of 5 | **5 of 6** - `episode_labels._write_document` omitted |
| cited sites that resolve | - | **2 of 7** |

Five of the seven rows cited a line the handler had moved away from, including the flagged one -
the site the entire section is about - by 745 lines (`simulation/isaac/simulation.py:5125`, really
`:5870`), and again in the alert-691 paragraph below the table.

## Why nothing caught it

`tests/test_codeql_query_filters.py::TestTheRulesFileSettlesTheCrossThreadMarshalClass` exists to
keep this passage honest, and grades which **concepts** it names: the rule id,
`concurrent.futures`, `run_on_main`, `SystemExit`. Every one of those survives the census rotting
underneath it. Its own docstring repeated `seven` too. Run the shipped file against the unfixed
tree:

```
$ pytest tests/test_codeql_query_filters.py -q     # on upstream/main
16 passed
```

## Change

The census is derived instead of transcribed.

- The table names each handler as `path::function`. The function that owns a handler survives
  edits above it; a line number does not, which is exactly how five of seven citations drifted.
  It now lists `strands_robots/` completely, and a following sentence covers the `tests/`,
  `examples/` and `scripts/` handlers, several of which use `raise AssertionError(...) from exc`
  rather than a bare `raise` - the query accepts either.
- The count is **gone** from the prose rather than corrected, in favour of the property that
  carries the argument. A total is incidental and would rot again; "every handler does so but
  one" is what the disposition needs, and it is now measured.
- `TestTheMarshalBoxCensusIsDerivedFromTheTree` asserts that against the tree: exactly one
  handler does not re-raise lexically, it is `run_on_main`'s box, every row names a handler that
  exists, the table is the whole package census, and the flagged column marks the one the tree
  says is flagged. Plus a floor, so a scan that stops reading the tree fails instead of
  reporting clean.

## Tests

`3 failed / 19 passed` against `upstream/main`, `22 passed` here.

| test | on main |
| --- | --- |
| `test_every_row_names_a_handler_that_exists` | **fails** - names all 7 stale citations |
| `test_the_table_is_the_whole_package_census` | **fails** - names the omitted `episode_labels._write_document` |
| `test_the_flagged_column_marks_exactly_the_one_that_is_flagged` | **fails** |
| `test_exactly_one_handler_does_not_reraise_lexically` | passes |
| `test_the_one_that_does_not_is_the_cross_thread_marshal_box` | passes |

The last two pass on **both** trees, and that is the point: the argument was correct throughout
and is not what this changes. They are also the ones with teeth going forward - a second
non-re-raising handler is a new alert of a rule whose review thread gates the merge, arriving
with no recorded disposition while the passage still reads as if it had one. It now fails loudly,
naming the site.

No package code changes, so there is no behaviour to render.

## Verification

Measured at `8247dda2`. 113-file blast radius - every test that reads `AGENTS.md`, walks a
source tree, or grades docstrings, strings and changelog fragments - run on this branch and on
`upstream/main` with the same selection:

| this branch | **0** | 3880 | 85 | 3965 |
| `upstream/main` + this test file | 3 | 3877 | 85 | 3965 |

`3877 + 3 == 3880`, the failed and passed deltas are both 3, the branch-only failure set is
**empty**, the base-only set is exactly the three pre-fix failures above, and the two trees share
no failure at all.

`ruff check` / `ruff format --check` and `mypy strands_robots tests tests_integ` clean over CI's
scope, with the mypy count unchanged against a pristine `upstream/main` worktree.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1829,18 +1829,26 @@ Corrections from code review that apply to all future contributions:
 - **One alert class clears under none of the three, and the question that settles
   it is which thread you marshal onto.** `py/catch-base-exception` never fires on
   cleanup-and-reraise: the query accepts a handler that re-raises *lexically*, and
-  six of the tree's seven `except BaseException` handlers do, so they have never
-  been flagged.
+  every `except BaseException` handler in this tree does so but one, which is why
+  none of them has ever been flagged. A count would rot here, so the census below
+  is stated as the property that matters and is derived from the tree by
+  `tests/test_codeql_query_filters.py` rather than copied into this file.
+
+  Every handler in `strands_robots/`, named by the function that owns it because a
+  line number is the part that goes stale:
 
   | handler | ends in | flagged |
   |---|---|---|
-  | `robot.py:368` | `sim.destroy()`, bare `raise` | no |
-  | `policies/persistent.py:193` | `handoff.abandon()`, bare `raise` | no |
-  | `simulation/safe_output.py:185` | `os.unlink(tmp)`, bare `raise` | no |
-  | `hardware_robot.py:1865` | `self._release_task()`, bare `raise` | no |
-  | `tests/policies/lerobot_local/test_list_policy_types.py:70` | `raise AssertionError(...) from exc` | no |
-  | `tests/policies/lerobot_local/test_vla_jepa.py:164` | `raise AssertionError(...) from exc` | no |
-  | `simulation/isaac/simulation.py:5125` | `box["exc"] = exc`, no lexical raise | **yes** |
+  | `strands_robots/episode_labels.py::_write_document` | `os.unlink(tmp_name)`, bare `raise` | no |
+  | `strands_robots/hardware_robot.py::start_task` | `self._release_task()`, bare `raise` | no |
+  | `strands_robots/policies/persistent.py::get_actions` | `handoff.abandon()`, bare `raise` | no |
+  | `strands_robots/robot.py::Robot` | `sim.destroy()`, bare `raise` | no |
+  | `strands_robots/simulation/safe_output.py::atomic_write_bytes` | `os.unlink(tmp)`, bare `raise` | no |
+  | `strands_robots/simulation/isaac/simulation.py::_job` | `box["exc"] = exc`, no lexical raise | **yes** |
+
+  The handlers under `tests/`, `examples/` and `scripts/` re-raise lexically too,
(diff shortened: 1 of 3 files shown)


yinsong1986 (approved):
## Summary

This PR replaces the hand-transcribed `py/catch-base-exception` handler census in AGENTS.md with one derived from the tree: the table now names each site as `path::function` (which survives edits above it, unlike the line numbers the old table cited — five of seven had drifted, the flagged one by 745 lines), lists `strands_robots/` completely (the old table omitted `episode_labels.py::_write_document`), and drops the rot-prone handler count in favour of the property the section's disposition rests on. The new `TestTheMarshalBoxCensusIsDerivedFromTheTree` class in `tests/test_codeql_query_filters.py` measures that property with an AST scan across all five source trees: exactly one handler does not re-raise lexically, it is the cross-thread marshal box by identity, and the AGENTS.md table's rows and flagged column are cross-checked against the scan — so a second non-re-raising handler (a new merge-gating alert) now fails loudly instead of arriving under a passage that reads as if it had a disposition for it.

## What's good

- Verified at head: all 22 tests in the module pass, ruff check/format clean, and the census matches the tree (6 handlers in `strands_robots/`, marshal box at `simulation/isaac/simulation.py::_job`).
- The scan guards its own vacuity (`_MIN_HANDLERS` floor), so a scan that stops reading the tree fails rather than passing every set-comparison trivially.
- Table parsing anchors on the unique header and fails loudly (`ValueError`) if the table is removed, which is the correct failure mode for a pin.
- Changelog fragment present and named per convention (`changelog.d/2572-base-exception-census-derived-from-the-tree.md`); no host paths, no non-ASCII in the diff.

PR #2574, opened 2026-08-21
ci(actions): pin every workflow uses: to a commit SHA, and grade it

### LINKED ISSUE #2573: ci(actions): docs.yml names four moving tags where the other 13 workflows pin all 31 uses: to a SHA, and no guard grades action pinning
`AGENTS.md` > **Action Pinning** states the rule and its reason:

> **All `uses:` references in workflows pin to a full 40-character commit SHA**, with the version
> tag preserved as a trailing comment [...] **Especially `pypa/gh-action-pypi-publish`** - it uses a
> moving `release/v1` branch, which is exactly the supply-chain pattern that the
> `tj-actions/changed-files` incident exploited. This pin is non-negotiable.

13 of the 14 workflows follow it: 31 of the 35 `uses:` references pin a 40-character SHA with the
tag as a trailing comment. `docs.yml` is the exception, and all four of its references name a
moving tag:

| file | line | reference |
| --- | --- | --- |
| `.github/workflows/docs.yml` | 35 | `actions/checkout@v4` |
| `.github/workflows/docs.yml` | 36 | `actions/setup-python@v5` |
| `.github/workflows/docs.yml` | 47 | `actions/upload-pages-artifact@v3` |
| `.github/workflows/docs.yml` | 65 | `actions/deploy-pages@v4` |

A tag is mutable, so each of these resolves to whatever the tag points at when the job starts.
That is the property the rule exists to remove, and `docs.yml` is the workflow that holds
`id-token: write` and `pages: write` for the Pages deployment, so it is not the cheapest place in
the tree to leave it.

Nothing detects this. There is no guard over `uses:` pinning - `tests/` reads
`.github/workflows/` in seven places (apt recommends, the overlap gate, the CodeQL config, the
ruleset scope, the viewer scope, the dependabot location, the lockfile gate), and none of them
grades action references. So the file has been outside the rule with every signal green.

### Suggested fix

Two of the four are decision-free: the repository already pins the same actions elsewhere, 13
times and 11 times respectively, so `docs.yml` can use the SHA its siblings use rather than a new
one.

```yaml
- uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
- uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
```

The other two appear only here. Resolving the tag each one currently names, so behaviour is
unchanged and only the mutability goes away:

```yaml
- uses: actions/upload-pages-artifact@56afc609e74202658d3ffba0e8f6dda462b719fa  # v3.0.1
- uses: actions/deploy-pages@d6db90164ac5ed86f2b6aed7e0febac5b3c0c03e  # v4.0.5
```

Both were read from the tag refs today; worth re-resolving at the time of the change rather than
trusting these, since that is the whole point of the rule. Dependabot's `github-actions` ecosystem
entry then keeps all four fresh, which is how the other 31 stay current.

A guard belongs with it, or the next workflow can be added the same way: every `uses:` that is not
a local `./`-relative reference must pin a 40-character SHA and carry a trailing comment. It would
have to land together with the four pins, since it fails on them.

Filing rather than opening a pull request because a change under `.github/workflows/` needs a token
carrying `workflow` scope, which is a separate permission from the one that can push the rest of
the tree.

cagataycali:
Fixes #2573

## Problem

`AGENTS.md` > **Action Pinning** calls the rule non-negotiable, and names the incident it exists to prevent:

> **All `uses:` references in workflows pin to a full 40-character commit SHA**, with the version tag preserved as a trailing comment [...] which is exactly the supply-chain pattern that the `tj-actions/changed-files` incident exploited.

`docs.yml` was outside it, and all four of its references named a moving tag:

| file | line | reference |
| --- | --- | --- |
| `.github/workflows/docs.yml` | 35 | `actions/checkout@v4` |
| `.github/workflows/docs.yml` | 36 | `actions/setup-python@v5` |
| `.github/workflows/docs.yml` | 47 | `actions/upload-pages-artifact@v3` |
| `.github/workflows/docs.yml` | 65 | `actions/deploy-pages@v4` |

A tag resolves to whatever it points at when the job starts. `docs.yml` is also the workflow holding `id-token: write` and `pages: write` for the Pages deployment, so it was the most expensive file in the tree to leave mutable.

The census, derived rather than transcribed. The linked issue said 31 of 35 references were pinned; that counts the 2 `./`-relative reusable-workflow calls as pins, and they name no version by design:

| | count |
| --- | --- |
| workflow files | 14 |
| `uses:` references | 35 |
| &nbsp;&nbsp;local `./` reusable-workflow calls | 2 |
| &nbsp;&nbsp;remote references | 33 |
| &nbsp;&nbsp;&nbsp;&nbsp;pinned before this change | **29** |

## Why nothing caught it

`tests/` reads `.github/workflows/` in seven places - apt recommends, the overlap gate, the CodeQL config, the ruleset scope, the viewer scope, the dependabot location, the lockfile gate - and not one of them reads a `uses:` line. So the file sat outside a non-negotiable rule with every signal green.

## Change

The four references are pinned, and the rule is graded from the tree.

**On which SHA.** The tags move, and that is measured rather than assumed: `actions/checkout@v4` resolved to `11bd7190` (v4.2.2) when the sibling workflows were pinned, and resolving the same tag for this change returned `11d5960a` - **v4.4.0**, two minor versions along, under a reference that reads as fixed.

So `checkout` and `setup-python` take the SHA their 12 and 9 siblings already carry, not the tag's current target. Pinning to today's `v4` would have removed the mutability *and* silently upgraded the docs build's checkout while the other 12 workflows stayed on v4.2.2 - an unrequested behaviour change inside a supply-chain fix, and a second version for Dependabot to reconcile separately. `upload-pages-artifact` and `deploy-pages` appear only here and are pinned to the exact commit the tag they named resolves to, so the build is unchanged.

**The guard.** `tests/test_workflow_action_refs_pin_a_sha.py` grades four properties, each derived from the tree rather than from a list:

1. a remote reference pins 40 hex characters;
2. it carries the trailing `# <tag>` comment, because a bare SHA names no version a reviewer can recognise;
3. an unpinned reference is a `./` call that resolves to a workflow file that exists - the exemption is structural, not discretionary;
4. an action resolves to one SHA tree-wide.

Property 4 is what makes the sibling SHA the only legal answer instead of a preference stated in a commit message, and keeps each Dependabot bump one atomic edit. Plus floors on the workflow and reference counts, and a per-file check that every line *mentioning* `uses:` was actually read - a line the parser cannot see is a reference nothing grades, and would otherwise report clean.

## Tests

`8 failed / 79 passed` against `upstream/main`, `87 passed` here.

| test | on main |
| --- | --- |
| `test_a_forty_character_sha_is_named` | **fails 4x** - names each of the four tags |
| `test_a_trailing_comment_names_the_version` | **fails 4x** - same four sites |
| `test_no_action_is_pinned_to_two_different_shas` | passes |
| `test_every_uses_line_is_read` | passes (14 files) |

The two halves are complementary - neither alone catches both regressions:

| regression | SHA pin | comment pin | one-SHA pin |
| --- | --- | --- | --- |
| a new workflow copies `@v4` | **fail** | **fail** | pass |
| the four pinned to the tag's current target | pass | pass | **fail** |

The second row is the plausible alternative fix, and it was run rather than reasoned about. With all four pinned but `checkout` at `11d5960a`, property 4 fails and names both SHAs with all 13 call sites:

```
an action is pinned to more than one SHA:
  actions/checkout:
    11bd7190 (v4.2.2) - agent-api-check.yml:47, breaking-change-check.yml:42, ... test-lint.yml:35
    11d5960a (v4.4.0) - docs.yml:35
```

## Verification

Measured at `71744545`. 19-file blast radius - every test that reads `.github/workflows/` or `changelog.d/`, including the six that reach the directory by path join rather than by literal string - run on this branch and on `upstream/main` with the same selection:

| --- | --- | --- | --- |
| this branch | 2 | 586 | 1 |
| `upstream/main` | 2 | 499 | 1 |

`499 + 87 == 586`, the branch-only failure set is **empty**, and the 2 failures are identical on both trees and environmental (`tests/test_dependency_audit.py`, missing `strands` and `qpsolvers` wheels in this sandbox - neither reads a workflow).

`docs.yml` re-parsed with `yaml.safe_load` after the edit: both jobs intact, all four steps resolving to the intended commits. `ruff check`, `ruff format --check` and `mypy` clean on the new file. No package code changes, so there is no behaviour to render.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,8 +32,8 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
         with:
           python-version: "3.12"
           cache: pip
@@ -44,7 +44,7 @@ jobs:
         run: mkdocs build --strict
       - name: Upload Pages artifact
         if: github.ref == 'refs/heads/main' && github.event_name != 'pull_request'
-        uses: actions/upload-pages-artifact@v3
+        uses: actions/upload-pages-artifact@56afc609e74202658d3ffba0e8f6dda462b719fa  # v3.0.1
         with:
           path: site
 
@@ -62,4 +62,4 @@ jobs:
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@v4
+        uses: actions/deploy-pages@d6db90164ac5ed86f2b6aed7e0febac5b3c0c03e  # v4.0.5
--- a/changelog.d/2573-workflow-action-refs-pin-a-sha.md
+++ b/changelog.d/2573-workflow-action-refs-pin-a-sha.md
@@ -1,0 +1,43 @@
+### Security: every workflow `uses:` reference pins a commit SHA, and a guard keeps it that way
+
+`AGENTS.md` > Action Pinning requires every `uses:` reference to name a full
+40-character commit SHA with the version tag as a trailing comment, because a tag
+is mutable and resolves to whatever it points at when the job starts -- the
+pattern the `tj-actions/changed-files` incident exploited. `docs.yml` was outside
+the rule, and all four of its references named a moving tag:
+
+| file | line | was | now |
+| --- | --- | --- | --- |
+| `.github/workflows/docs.yml` | 35 | `actions/checkout@v4` | `@11bd7190`  `# v4.2.2` |
+| `.github/workflows/docs.yml` | 36 | `actions/setup-python@v5` | `@a26af69b`  `# v5.6.0` |
+| `.github/workflows/docs.yml` | 47 | `actions/upload-pages-artifact@v3` | `@56afc609`  `# v3.0.1` |
+| `.github/workflows/docs.yml` | 65 | `actions/deploy-pages@v4` | `@d6db9016`  `# v4.0.5` |
+
+Of 35 `uses:` references across the 14 workflows, 2 are `./`-relative calls to
+this repository's own reusable workflows and name no version by design; of the 33
+remaining, 29 were pinned and these 4 were not. `docs.yml` is also the workflow
+holding `id-token: write` and `pages: write` for the Pages deployment, so it was
+the most expensive file in the tree to leave mutable.
+
+That the tags move is measured, not hypothetical. `actions/checkout@v4` resolved
+to `11bd7190` (v4.2.2) when the sibling workflows were pinned; resolving the same
+tag for this change returned `11d5960a` (**v4.4.0**), two minor versions along,
+under a reference that reads as fixed. So `checkout` and `setup-python` take the
+SHA their 12 and 9 siblings already carry rather than the tag's current target:
+pinning to today's `v4` would have removed the mutability and silently upgraded
+the docs build's checkout while every other workflow stayed on v4.2.2 -- an
+unrequested behaviour change inside a supply-chain fix. The other two actions
+appear only here, and are pinned to the exact commit the tag they named resolves
+to, so the build is unchanged.
+
+Nothing detected any of this. `tests/` reads `.github/workflows/` in seven places
+and none of them reads a `uses:` line, so the file sat outside a non-negotiable
+rule with every signal green. `tests/test_workflow_action_refs_pin_a_sha.py`
+grades four properties against the tree: a remote reference pins 40 hex
+characters, it carries the trailing tag comment, an unpinned reference is a `./`
+call resolving to a workflow that exists, and an action resolves to one SHA
+tree-wide. The last is what makes the sibling SHA the only legal answer and keeps
+each Dependabot bump one atomic edit. 8 of its 87 checks fail on the previous
+`docs.yml`, and the one-SHA check fails on the plausible alternative fix.
+
+No package code changes.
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Pins the four remaining mutable `uses:` references in `docs.yml` to 40-character commit SHAs with trailing version-tag comments, and adds `tests/test_workflow_action_refs_pin_a_sha.py` to grade the AGENTS.md > Action Pinning rule tree-wide. Verified independently: all four SHAs dereference to exactly the tags their comments claim (`actions/checkout@v4.2.2` -> `11bd7190...`, `actions/setup-python@v5.6.0` -> `a26af69b...`, `actions/upload-pages-artifact@v3.0.1` -> `56afc609...`, `actions/deploy-pages@v4.0.5` -> `d6db9016...`), the `checkout`/`setup-python` pins are byte-identical to their 12 and 9 sibling call sites so the one-SHA-tree-wide property holds and no build behaviour changes, the new suite passes 87/87 on this head and reproduces the claimed 8 failures when main's `docs.yml` is substituted back in, `docs.yml` re-parses with both jobs intact, and ruff/format are clean on the new file. The guard's failure direction is fail-safe (an unparseable `uses:` line fails the mention/parse parity check rather than going ungraded), the `./`-local exemption is structural, and the existing Dependabot `github-actions` ecosystem entry keeps the pins fresh.

## What's good

- SHA choice discipline: taking the sibling SHA rather than the tag's current target avoids smuggling a v4.2.2 -> v4.4.0 checkout upgrade into a supply-chain fix, and the one-SHA-per-action test makes that the only answer the suite accepts.
- The census floors and per-file mention/parse parity mean a broken walk or a regex gone blind fails loudly instead of reporting a clean pinning record.
- Changelog fragment per `changelog.d/README.md`; no package code touched; scope matches the title exactly.

PR #2575, opened 2026-08-21
fix(sim/isaac): refuse a render whose camera_name no camera carries

cagataycali:
## Problem

`IsaacSimulation.render` documents four blank-frame fallback paths, and one of them named **two causes with one answer**:

> `Rendered (no camera)` -- `camera_name` is unknown to `self._cameras`. Caller probably forgot to call `add_camera` (or typo'd the name).

The typo half was answered with success. Measured with one camera named `wrist` registered and `render_mode="rtx_realtime"`:

```
render("wrst") -> status='success'
                  text='Rendered (no camera): 640x480'
                  json={'pixel_mean': 0.0, 'pixel_variance': 0.0, 'camera': 'wrst'}
                  image=PNG block present
```

Three things follow from that envelope:

- `pixel_mean` `0.0` is reported as a **measurement**, and `camera` names a camera that does not exist.
- the frame is `640x480` - the config default - rather than any camera's resolution.
- the PNG block above that json is emitted precisely so the shared `PolicyRunner._extract_frame_ndarray` can pull frames for video recording (`_rgb_png_block`'s own docstring says so). A rollout recording a mistyped camera therefore **writes an all-black video and reports success**.

The same name through this same class's `get_frame` raised `KeyError: Camera 'wrst' not found. Available: ['wrist']` - one registry, one lock, one question, two answers. And the sibling backends refuse it too, so Isaac's `render` was the one cell of six that succeeded:

| backend | `render(<name nothing carries>)` | `get_frame(<same name>)` |
| --- | --- | --- |
| MuJoCo | `status=error`, `Camera 'x' not found. Available: [...]` | `KeyError`, same sentence |
| Newton | `status=error`, same sentence | `KeyError`, same sentence |
| **Isaac** | **`status=success`**, black frame, `pixel_mean 0.0` | `KeyError`, same sentence |

## Change

Split the branch by the docstring's own two causes.

`FREE_CAMERA_TOKENS` is the shared spelling of "no camera was *named*" - `None`, `""`, `"default"` and `"free"`. A `camera_name` outside that set **names** a camera the scene does not carry: that is a caller mistake this method can name, and it is now refused with the verdict `get_frame` already gives. A token that names no camera keeps the blank frame **byte for byte**: Isaac has no free camera to fall back to, unlike the two backends whose render entry points resolve those tokens to one, and `"default"` is this method's own signature default - so broadening the refusal to every registry miss would turn a bare `render()` into an error.

The token test is membership in a *tuple*, whose `==`-per-element comparison stays total for a name that is not hashable at all, matching `registered`'s own totality contract.

Scope notes:

- `render_mode="headless"` answers before the camera lookup and is untouched: RTX produces no frames there at all, which is a degradation rather than a mistake.
- `reserved_camera_name_error` is deliberately **not** applied on this backend, and that stays true: its docstring records that Isaac routes no tokens, so `"default"` is an ordinary addressable name here. The token set is consulted only on a registry *miss*, so `add_camera("default", ...)` is still accepted and `render("default")` still renders it.
- the documented usage example in `docs/simulation/isaac.md` registers `front` via `add_camera` before rendering it, so it is unaffected.

`docs/simulation/isaac.md` already carried the paragraph enumerating the lookup surfaces and how each reports a miss - naming `get_frame` / `get_camera_params` but not `render`. It now names `render` too.

## Verification

Pre-fix split, the new test file copied onto `upstream/main`: **23 failed / 13 passed**, 36 passed after. Every behavioural failure carries the measurement, e.g.

```
AssertionError: render('wrst') reported 'success' for a camera this scene does not carry,
  with text 'Rendered (no camera): 640x480' and json {'pixel_mean': 0.0, 'pixel_variance': 0.0, 'camera': 'wrst'}
```

The 13 that pass on both trees are the controls, and each fails for a specific wrong fix: all four `FREE_CAMERA_TOKENS` members keep the blank frame byte for byte; the signature default is one of them; both `headless` cases answer first as documented; a carried camera with no RTX handle keeps its Phase-1 frame; a carried camera still renders its own frame; a refused name costs no handle read; and the MuJoCo and Newton parity rows were already green - only the Isaac row of that parity test flips.

Blast radius, 79 files (all of `tests/simulation/isaac` plus every test naming the shared camera domains or reading `AGENTS.md` / a docs page, plus the repo-wide guards), same selection on both trees:

| this branch | 0 | 4101 | 29 | 4130 |
| `upstream/main` + the new test file | 23 | 4078 | 29 | 4130 |

Branch-only failing ids: empty. Base-only: exactly the 23 from the new file, none foreign. No failure is shared by the two trees. `mypy strands_robots tests tests_integ`: 24 errors on a pristine `upstream/main` worktree and 24 on the branch, none in the changed files. `ruff check` / `ruff format --check` clean.

## Artifact

Isaac Sim is not needed to see this: the camera lookup runs before any RTX handle is read, and the handle is duck-typed. Both columns below run the **same** capture script against a stand-in engine whose handle hands back a real 6-frame MuJoCo rollout - only `strands_robots` differs.

Isaac render refuses an uncarried camera name

The `wrist` strip is the control: identical in both columns, byte for byte (mean `110.92003942418981`). The mistyped `wrst` name is where they diverge - on `upstream/main` six all-black frames (`mean 0.00`) each carrying a PNG block, so six frames reach a recording under success; on this branch the call is refused and nothing is recorded.

### DIFF
--- a/changelog.d/2575-isaac-render-refuses-an-uncarried-camera-name.md
+++ b/changelog.d/2575-isaac-render-refuses-an-uncarried-camera-name.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **sim/isaac**: `render` refuses a `camera_name` that names a camera the scene does not carry, with the same `Camera '<name>' not found. Available: [...]` verdict its own `get_frame` raises and the MuJoCo and Newton `render` give. It previously reported `status="success"` with an all-black frame sized from the config default, `pixel_mean` `0.0` as a measurement, the missing name in the `camera` field, and the PNG block the shared frame extractor reads - so a rollout recording a mistyped camera wrote an all-black video and reported success. A `camera_name` that names no camera at all (`None`, `""`, `"default"`, `"free"`) keeps that blank frame unchanged: this backend has no free camera to fall back to, and `"default"` is `render`'s own signature default.
--- a/docs/simulation/isaac.md
+++ b/docs/simulation/isaac.md
@@ -237,6 +237,22 @@ membership test itself raised `TypeError: unhashable type` for a list or dict
 name, so the miss escaped the envelope those methods document as their only
 failure channel - reachable with no entities registered at all.
 
+`render` is the one lookup that cannot answer with a frame, so it reports the
+same verdict as its raw sibling. Given a `camera_name` that *names* a camera the
+scene does not carry it returns `{"status": "error"}` with
+`Camera '<name>' not found. Available: [...]` - the message `get_frame` raises
+for the identical name, and the one the MuJoCo and Newton `render` give. It used
+to report `status="success"` with an all-black frame instead, tagged
+`Rendered (no camera)`, plus `pixel_mean` `0.0` as a measurement and the missing
+name in the `camera` field; because that envelope carries the PNG block the
+shared frame extractor reads, a rollout recording a mistyped camera wrote an
+all-black video and reported success. A `camera_name` that names *no* camera -
+`None`, `""`, `"default"` (the signature default) or `"free"` - still gets that
+blank frame: Isaac has no free camera to fall back to, unlike the two backends
+whose render entry points resolve those tokens to one, so for them it is a
+degradation rather than a mistake. Registering a camera under one of those names
+is accepted here and renders normally, since nothing on this backend routes them.
+
 ## Fleet (IsaacLab-style) preview
 
 ```python
--- a/strands_robots/simulation/isaac/simulation.py
+++ b/strands_robots/simulation/isaac/simulation.py
@@ -45,6 +45,7 @@ from strands_robots.simulation.isaac.recording import IsaacRecordingMixin
 from strands_robots.simulation.models import registered, registry_entry
 from strands_robots.simulation.terrain import validate_difficulty
 from strands_robots.utils import (
+    FREE_CAMERA_TOKENS,
     camera_fov_error,
     coerce_pose_vector,
     coerce_rgba,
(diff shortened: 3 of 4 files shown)


yinsong1986 (dismissed):
## Summary

Splits Isaac's `_render_frame` registry-miss branch by the docstring's own two causes: a `camera_name` outside `FREE_CAMERA_TOKENS` that names a camera the scene does not carry is now refused with the exact sentence `get_frame` (and the MuJoCo/Newton `render`) already give, while the four token spellings of "no camera was named" (including the `"default"` signature default) keep the documented blank-frame degradation byte for byte. This closes the one cell of the 3-backend x 2-surface matrix that answered a typo'd camera with `status="success"`, black pixels, `pixel_mean 0.0` reported as a measurement, and a PNG block that fed an all-black video into the shared recording pipeline.

## What's good

- The refusal is ordered correctly: inside `self._lock`, after the headless early-return (so headless stays a degradation as documented), and before any RTX handle read - the `test_a_refused_name_costs_no_render` control pins that.
- Both internal `_render_frame` consumers already handle the `(None, None, meta)` refusal shape: `get_frame` guards the registry before calling, and the recording `on_frame` increments the camera's error counter on `rgb is None` rather than crashing, so the recording hot path degrades to counted errors instead of silent black frames - which is the point of the fix.
- The verdict sentence is shared with `get_frame` (same f-string, same `sorted(self._cameras)`), and the parity/one-verdict tests pin the two surfaces against re-drifting.
- Scope discipline: `reserved_camera_name_error` stays un-applied on Isaac (its own docstring records why), so `add_camera("default")` + `render("default")` keep working; the token check is only consulted on a registry miss.
- Changelog fragment recorded per AGENTS.md step 3 (`changelog.d/2575-...md`, `### Fixed` first line).
- 36 tests pass on the head; 13 controls pin every documented fallback byte for byte, including the signature default. `ruff check` / `ruff format --check` clean on the changed files; no non-ASCII, no host paths.

## Verification suggestions

- `pytest tests/simulation/isaac/test_render_refuses_a_camera_the_scene_does_not_carry.py -q` - 36 pass on the head, and copying the file onto `main` reproduces the 23-failed pre-fix split the description reports.

yinsong1986 (approved):
## Summary

Splits Isaac `_render_frame`'s registry-miss branch by the docstring's own two causes: a `camera_name` outside `FREE_CAMERA_TOKENS` that names a camera the scene does not carry is now refused with the exact `Camera '<name>' not found. Available: [...]` sentence this backend's `get_frame` raises and the MuJoCo/Newton `render` already give, while the four token spellings of "no camera was named" (including the `"default"` signature default) keep the documented blank-frame degradation byte for byte. This closes the one cell of the 3-backend x 2-surface matrix that answered a typo with `status="success"`, a config-default-sized black frame, `pixel_mean` 0.0 as a measurement, and a PNG block the shared `PolicyRunner._extract_frame_ndarray` would record as an all-black video. The refusal sits before any RTX handle read, `render` already routes `meta["error"]` into the error envelope, and the one internal `_render_frame` caller that could see the new `None` return (the camera-recording `on_frame` hook) already handles `rgb is None` by counting an error rather than raising. The membership test on the shared tuple stays total for non-string and unhashable names, matching `registered`'s contract.

## What's good

- Verified locally at head `5b63cf4c`: the new file's 36 tests pass, the full `tests/simulation/isaac` selection is 934 passed / 0 failed, and `hatch run lint` is clean.
- Parity is pinned structurally (`TestEveryBackendGivesTheSameVerdict`) and the over-reach direction is controlled: all four `FREE_CAMERA_TOKENS` members, the signature default, both `headless` orderings, and the Phase-1 no-handle fallback are asserted unchanged, so the fix cannot silently widen into refusing a bare `render()`.
- The prior CodeQL "unnecessary lambda" finding on `bed97956` is addressed at head (`ids=str` on the parametrize).
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; docstring, docs page and code comment all state the same split; error strings are plain ASCII.

Issue #250, opened 2026-05-28
mesh(iot): CA pin rotation runbook (signed manifest or documented runbook)


cagataycali:
## Background

Review feedback on PR #228 (thread on `strands_robots/mesh/iot/provision.py:79`) flagged that the current Amazon Root CA1 pinning strategy makes operational rotation awkward:

- Pin tuple is hard-coded in `provision._AMAZON_ROOT_CA1_PINS`.
- Optional `STRANDS_MESH_CA_PINS` env var lets operators add pins out-of-band, but that's not a fleet-wide rotation strategy.
- An AWS root CA rotation requires a tagged release of `strands_robots`, then waiting for fleets to upgrade. On-call at 3 AM needs better than 'derive the recompute command from the docstring.'

## Scope of this issue

Deliver one of (preferred order):

### Option A: Signed pins.json manifest

- Library at startup fetches `pins.json` from a known URL (e.g., S3 bucket the user controls) over HTTPS with the existing pinned CA chain.
- Manifest is signed (Sigstore / cosign / Ed25519); signature verified before pins are accepted.
- Manifest format includes both the new pin and the previous pin during the rotation grace period; library accepts any pin in the manifest.
- Document the operator-side workflow: 'how to publish a new manifest before AWS rotates' + grace-period guidance.

### Option B: Documented rotation runbook (minimum viable)

If Option A is too heavy for now, ship at minimum:

- README.md section 'CA pin rotation' with:
  - The recompute one-liner (already in `provision.py` docstring; promote to README).
  - Grace-period strategy: how to land the new pin while the old one is still valid (likely: ship a release with both pins in `_AMAZON_ROOT_CA1_PINS`, wait for fleet uptake, then a follow-up release that drops the old pin).
  - How operators monitor for upcoming AWS root CA rotations (subscribing to AWS security bulletins; AWS publishes deprecation timelines).
  - Emergency override: temporary use of `STRANDS_MESH_CA_PINS` if a rotation lands faster than a code release can ship.
- AGENTS.md > Review Learnings entry capturing 'security pins must come with a rotation runbook, not just a recompute command.'

## Acceptance criteria

- [ ] Either signed-manifest fetch path lands (Option A) or rotation runbook is published in README.md (Option B).
- [ ] AGENTS.md learnings entry pinned for the operational-fragility-of-static-pins lesson.
[...]

PR #2576, opened 2026-08-21
docs(mesh/iot): publish the CA pin rotation runbook the pin's own artifacts cite

cagataycali:
## Problem

The Amazon Root CA1 pin ships three artifacts. Two of them cite a third.

`AGENTS.md > "Operational Runbooks for Security Pins"` states the rule and names where the procedure lives:

> A static security pin must ship with a rotation runbook, not just a recompute command. [...] For the Amazon Root CA1 pin (`provision._AMAZON_ROOT_CA1_PINS`) the runbook lives in README.md > "CA Pin Rotation Runbook"

`tests/mesh/test_iot_ca_pin_rotation.py` opens by saying the same thing, and explains that its own assertions exist to hold up that document:

> Issue #250 (Option B) ships a documented rotation runbook for the Amazon Root CA1 pin [...] The operational guarantee **the runbook depends on** is that the accepted-pin set is a collection

`"CA Pin Rotation Runbook"` appeared in exactly one place in the tree: the AGENTS.md sentence citing it. README carried no such heading, and no page under `docs/` covers the pin either, so the ordered procedure existed only in a contributor file and a test docstring. Issue #250's first acceptance criterion was the runbook; the other two (the AGENTS.md entry, the dual-pin test) landed and this one did not.

## What an operator meets during a rotation

Driving both pin paths against a root that matches no accepted pin:

| path | refuses | names a procedure | env var it does name |
| --- | --- | --- | --- |
| cached PEM re-used | yes | no | `STRANDS_MESH_DISABLE_CA_PIN` (to say it does not apply here) |
| re-download of the same bytes | yes | no | none |

Both refusals are correct and neither is actionable. The on-disk warning advises `Delete the file to force re-download`, which leads to the download path, which refuses the same bytes for the same reason - so following the one piece of advice on offer returns to where it started. The only env var either message mentions is the break-glass that turns verification off.

The capability was never missing: staging a verified digest in `STRANDS_MESH_CA_PINS` admits the certificate immediately (`admitted=True, accepted pins=2`), because the accepted set is a collection. Only the procedure was undiscoverable.

## Change

Publishes the runbook under the heading the citation names, transcribing the strategy already decided in the AGENTS.md bullet and in #250 - verify out of band, ship a release carrying both pins, wait for fleet uptake, drop the old pin in a follow-up, and use `STRANDS_MESH_CA_PINS` as the emergency overlap. It states explicitly that `STRANDS_MESH_DISABLE_CA_PIN` is not part of the procedure: it accepts whatever the URL serves rather than widening the accepted set, and a rotation has a verified pin to stage. Nothing new is decided here.

All three pin refusals now name the runbook, through one module-level owner so they cannot drift to three different places or away from the cited heading. The pointer is appended - `failed pin check`, `accepted pins:`, `does NOT match pinned SHA-256` and `refusing to write rogue CA` are unchanged, since operators and forks match on them. The on-disk warning also now says why re-downloading will not help if the root has rotated.

The mismatch path deliberately does **not** name `STRANDS_MESH_CA_PINS`. It cannot tell a rotation from a MITM, so naming the add-a-pin knob in a message that just called those bytes rogue would read as "stage what you got"; it points at a procedure whose first step is out-of-band verification instead. That boundary has its own test.

## Tests

`tests/mesh/test_iot_ca_pin_rotation.py` - the file #250 names - gains a runbook half beside its dual-pin half. Everything graded is derived from the code rather than restated: the heading comes from AGENTS.md's own citation (so a rename on one side fails here instead of becoming a dead pointer), the env vars from the module's own `os.getenv` calls, and the CA URL from `_AMAZON_ROOT_CA1_URL`.

Against `upstream/main`: **6 failed / 6 passed**, 12 pass after. The headline:

```
AssertionError: AGENTS.md cites README.md > 'CA Pin Rotation Runbook' as where the pin
rotation runbook lives, and README.md has no such heading - the rotation procedure is
published nowhere an operator reads
```

and the refusal test fails on the message itself:

```
AssertionError: the on-disk re-use refusal names no rotation procedure: AmazonRootCA1 at
<path> failed pin check; accepted pins: 2c43952e...
```

The 6 that pass on both trees are the controls, and each fails for a specific wrong fix: a rejected download digest is not offered as something to stage; the four existing refusal phrases are preserved; a staged pin still admits the certificate it covers; and the three pre-existing dual-pin assertions show the mechanism is untouched.

## Verification

Measured at `b7238a0`. Selection of 121 files - both CA-pin suites, every test that reads README.md or AGENTS.md, everything naming the provisioner, plus the repo-wide docs/changelog/docstring/string guards - run on this branch and on a worktree of `upstream/main` carrying the same new test file:

| --- | --- | --- | --- |
| this branch | 0 | 3698 | 29 |
| `upstream/main` | 6 | 3692 | 29 |

3727 collected on both, `3692 + 6 == 3698`, branch-only failures empty, and the 6 base failures are exactly the new tests with nothing foreign. `mypy strands_robots tests tests_integ` is 24 errors on both trees with none in the changed files; ruff check and format clean.

## Artifact

Both columns run the same capture script against the same rotated root; only `strands_robots`, README and AGENTS.md differ. The refusals, the digest and the accepted-pin set are identical - the figure script asserts that - and what changes is whether the cited procedure exists and whether the failure can reach it.

CA pin rotation runbook: same refusal, different reachability

## Scope

The signed-manifest path (#250 Option A) stays out of scope - this is Option B, which #250 describes as the minimum viable alternative to it. `STRANDS_MESH_CA_PINS` and `STRANDS_MESH_DISABLE_CA_PIN` already had rows in the Configuration matrix; what they lacked was a procedure telling an operator which of the two a rotation calls for.


yinsong1986 (approved):
## Summary

Publishes the CA pin rotation runbook under the exact README heading that `AGENTS.md > "Operational Runbooks for Security Pins"` already cited (`README.md > "CA Pin Rotation Runbook"`), closing the gap where the ordered procedure existed only in a contributor file and a test docstring. The three pin refusals in `strands_robots/mesh/iot/provision.py` (on-disk re-use warning, on-disk re-use error, download mismatch error) now name the runbook through a single module-level owner (`_CA_ROTATION_RUNBOOK`), with the pre-existing operator-matched phrases (`failed pin check`, `accepted pins:`, `does NOT match pinned SHA-256`, `refusing to write rogue CA`) preserved verbatim and pinned by test. The runbook content matches the code it describes: `STRANDS_MESH_CA_PINS` is documented as additive with invalid entries skipped (which is what `_resolve_ca_pins` does), and `STRANDS_MESH_DISABLE_CA_PIN` is explicitly kept out of the rotation steps - and the mismatch refusal deliberately does not name the add-a-pin knob, a boundary with its own test. New tests derive the graded heading from AGENTS.md's own citation and the env vars from the module's `os.getenv` calls rather than restating them, so a rename on either side fails loudly instead of becoming a dead pointer.

## What's good

- Verified on the head SHA: all 12 new tests pass, the full CA-pin/provision-related suite (176 tests) passes, and `ruff check` is clean on both changed Python files.
- Changelog fragment present (`changelog.d/2576-ca-pin-rotation-runbook.md`) per AGENTS.md step 3; no `CHANGELOG.md` edit.
- The refusal pointer is appended, never substituted, and the security boundary (no `STRANDS_MESH_CA_PINS` in the rogue-bytes message) is both stated and test-pinned - this is the right call, since the mismatch path cannot distinguish a rotation from a MITM.
- No emojis / non-ASCII in the diff, no host paths, lazy `%s` logging with matching args in the modified warning.

PR #2577, opened 2026-08-21
docs(sim): name what ends a randomization, not just what preserves it

cagataycali:
## Problem

`MuJoCoSimEngine.randomize`'s docstring told a caller how long a randomization lasts:

> Randomization IS destructive; recompile the scene to undo. **Every axis persists for the rest of the scene's life**, including across `reset()`

Neither half survives measurement. Nothing named `recompile` is among the 77 actions the tool surface publishes, and the persistence claim is false for every axis the moment any scene mutation happens - which is well inside "the scene's life". Measured over the eight `model` arrays the four axes write, on a scene with a robot, an object and a camera:

| operation | axes lost (of 5) |
|---|---|
| `reset` | 0 |
| `step` | 0 |
| `set_geom_properties` / `set_body_properties` | 1 (the entity the caller named) |
| `add_object`, `remove_object`, `add_camera`, `remove_camera`, `add_robot`, `remove_robot`, `patch_scene_mjcf` | 4-5 |

The compiled model is derived state, and every scene mutation recompiles the scene spec over it. So the documented undo is unreachable by that name while seven ordinary published actions perform it as a side effect.

The reachable cost is an ordering the API invites. `BenchmarkProtocol.on_episode_start` documents itself as the place to "layer on per-episode randomization, goal sampling, or **procedural scene generation**", and the two orderings disagree while both calls report success:

| per-episode sequence | randomized at the policy's first observation? |
|---|---|
| `randomize()` then add this episode's objects | **no** |
| add this episode's objects then `randomize()` | yes |

A caller told the axis persists for the scene's life has no reason to prefer one.

## Change

The behaviour is right and is unchanged. The scene spec carries the authored values deliberately: `randomize_lighting` measures its offset from "the pose the scene spec declares, which is also what a recompile restores", and `randomize_positions` from each object's commanded pose, and that fixed reference is what keeps a per-episode loop inside the advertised bound instead of walking. Writing the perturbed values into the spec would move the reference and bring the random walk back. The suite already pins this for one axis - `test_a_recompile_still_undoes_the_position_axis`, whose own comment reads "Any scene mutation recompiles the model" - and its module docstring already says the axes are "undone only by a recompile". Only the API prose disagreed.

So both surfaces that make the claim now name the boundary:

- `randomize`'s docstring keeps the reset-survival half and the reasoning that depends on it (a rollout entry point resets before an episode's first step, so an axis a reset undid would never reach a rollout), then states that the compiled model is derived state, names the seven operations measured to restore the authored scene, says there is no `recompile` action to reach for, and tells the caller to randomize after the episode's scene is built.
- `docs/simulation/domain-randomization.md` completes its restore list, which named only `load_scene(...)` and "recreate the sim", and adds the ordering note to the eval-loop example.

## Tests

`tests/simulation/mujoco/test_randomize_persistence_boundary.py` measures the boundary from the running engine and then grades the prose against it, so the two cannot drift again. The classification counts distinct **axes** rather than arrays, because a mass change legitimately moves `body_mass` and `body_inertia` together; the measured separation is wide (a targeted setter loses one axis, a scene mutation four or five). The name check is deliberately permissive about *where* a name appears, so it can under-report a surface and never over-report one.

Against `upstream/main` (`4c33345`): **3 failed / 11 passed**, each failure naming the exact operations the surface omits, e.g.

```
randomize()'s docstring tells the caller how long a randomization lasts but never names
these operations, each measured to restore the authored scene: ['add_camera', 'add_robot',
'patch_scene_mjcf', 'remove_camera', 'remove_object', 'remove_robot']
```

All 14 pass after. The 11 that pass on both trees are the point - they are the behaviour, asserted as the control: the two keeping operations, the targeted setters, all seven discarding routes, and the ordering above.

## Verification

Measured at `d04a896` on mujoco 3.12.0 (the version the declared range resolves), MUJOCO_GL=egl.

|---|---|---|---|
| this branch | 8 | 8977 | 93 |
| `upstream/main` (`549ebef`) | 8 | 8963 | 93 |

186-file selection (every test naming `randomize` or the persistence helpers, the MuJoCo randomization suites, every docs-reading test, and every repo-wide guard that walks a source tree or grades docstrings, strings or changelog fragments). The failing ID sets are **identical** - `comm` is empty in both directions - and the `+14` passed is exactly this PR's new test file. The 8 shared are pre-existing locally and all sit in `tests/mesh/test_iot_connect_timeout_domain.py`: `awsiotsdk` is absent from this environment and is declared by `[mesh-iot]`, which `[all]` pulls in, so they pass on CI.

Over that same selection the round below moves exactly one verdict - the previous head `b966ce6` measured 9 failed / 8976 passed, the difference being `test_qualified_strands_robots_xref_roles_resolve`. The prose split quoted under **Tests** was re-measured against this base and is unchanged.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on both trees, none in the changed files.

## Artifact

One capture script per tree. It reads `randomize`'s docstring **from the tree it runs against**, takes the call order that docstring prescribes, and renders what the policy's first observation would see - so nothing is hardcoded per column and only the prose differs.

randomize persistence boundary

The left column is the control: the authored scene renders **byte-identically in both trees** (`np.array_equal` true). Following `upstream/main`'s guidance leaves the coloured-cube pixel count at **20805 - exactly the authored value** and a mean absolute difference of 0.192; following this branch's leaves 242042 and 19.601. Camera picked by measurement over six candidates (20805 cube pixels against 4562 for the widest).

### Round 1 - `d04a8969`

One concern, one commit. A cross-reference in the new test module's docstring named `strands_robots.simulation.base.BenchmarkProtocol.on_episode_start`, but the class is defined in `strands_robots.simulation.benchmark` and `simulation.base` does not re-export it (`hasattr(base, 'BenchmarkProtocol')` is `False`). Because the repo's xref guard grades test-module docstrings alongside the package's, and that guard runs inside `call-test-lint` - the sole required context - this was a merge blocker rather than docs polish, and it is what put the previous head red.

```diff
-everything: :meth:`~strands_robots.simulation.base.BenchmarkProtocol.on_episode_start`
+everything: :meth:`~strands_robots.simulation.benchmark.BenchmarkProtocol.on_episode_start`
```

Re-derived from the branch head with the guard's own resolver, so the check and the claim cannot disagree:

| target | `_resolves()` |
|---|---|
| `...simulation.base.BenchmarkProtocol.on_episode_start` | `False` |
| `...simulation.benchmark.BenchmarkProtocol.on_episode_start` | `True` |

`_offending_roles_in()` over this module's docstring now returns `[]`, where it previously returned exactly that one target. Both modules import in the environment measured in, so the optional-dep skip carve-out was not laundering the verdict.

Two checks beyond the one token, since a wrong module path can travel:

- **The quoted prose is faithful at the new path.** The "per-episode randomization, goal sampling, or procedural scene generation" line this module quotes is `on_episode_start`'s own docstring at `strands_robots/simulation/benchmark.py:171`, so the citation points at the text it paraphrases rather than merely at something importable.
- **No other copy of the wrong path exists.** `grep -rn 'simulation\.base\.BenchmarkProtocol' --include=*.py --include=*.md` over the tree returns nothing, so this was a single occurrence and not one instance of a pattern.

The target stays contiguous on one line at 91 chars (limit 120), so the wrapped-target rule the same guard enforces still holds - the five extra characters did not push it into a break.

No behaviour, no test logic and no shipped prose changed in this round: the diff is one dotted path in one docstring.


tests/simulation/mujoco/test_randomize_persistence_boundary.py
```diff
@@ -0,0 +1,311 @@
...
+deliberately, because the lighting and position axes measure their bounded
+offsets from that reference. So a scene mutation restores the authored scene and
+undoes every axis.
+
+``randomize()``'s docstring claimed the opposite -- "Every axis persists for the
+rest of the scene's life" -- and named no operation that ends it, while the one
+undo it did name (``recompile the scene``) is not an action the tool surface
+publishes at all. Seven published scene actions perform that undo incidentally,
+and the sequence a per-episode loop reaches for first is the one that loses
+everything: :meth:`~strands_robots.simulation.base.BenchmarkProtocol.on_episode_start`
```
yinsong1986: [MUST FIX] This cross-reference names a module that does not hold the class: `BenchmarkProtocol` is defined in `strands_robots.simulation.benchmark` (`benchmark.py:72`), not `strands_robots.simulation.base`, and `strands_robots.simulation.base` does not re-export it (`hasattr(base, 'BenchmarkProtocol')` is `False`).

This is a merge blocker rather than docs polish because the repo's xref guard grades test-module docstrings too (AGENTS.md > Review Learnings (#86) > Public API Hygiene > "A cross-reference in a test docstring is graded too"), and that guard runs inside the one required check. Verified on this head:

```
tests/test_docstring_xref_roles_resolve.py::test_qualified_strands_robots_xref_roles_resolve
FAILED — offenders: {'tests/simulation/mujoco/test_randomize_persistence_boundary.py::<module>':
  ['strands_robots.simulation.base.BenchmarkProtocol.on_episode_start']}
```

It passes on `upstream/main` (549ebef0), so the failure is branch-only, and CI installs `.[all,dev]` so the optional-dep skip carve-out cannot save it there. `call-test-lint / Test and Lint` — the sole required context — will go red on this head.

Fix is one token: `:meth:`~strands_robots.simulation.benchmark.BenchmarkProtocol.on_episode_start`` (verified to resolve, and its docstring carries the "procedural scene generation" line this module quotes).

yinsong1986 (commented):
## Summary

Docs-only behaviour surface: `randomize()`'s docstring and `docs/simulation/domain-randomization.md` claimed every axis "persists for the rest of the scene's life" and offered a `recompile` undo that no published action provides; this PR rewrites both to name the real boundary (survives `reset()`/`step`, undone by any of the seven scene-mutation actions) and adds `tests/simulation/mujoco/test_randomize_persistence_boundary.py`, which measures the boundary from the running engine and grades both prose surfaces against it so they cannot drift again. No behaviour change — the spec-carries-authored-values design is asserted as the control. One blocker found and verified locally, flagged inline; it is a one-token fix.

## What's good

- The grader measures the boundary from the engine rather than restating a list, so a new scene mutation is graded on arrival — same shape as the repo's other derive-the-population guards.
- Boundary claims verified independently on this tree: all 14 new tests pass on the branch, and the three prose-grading tests fail on `upstream/main` exactly as the description claims.
- Changelog fragment present and passing the fragment guards; lint/format clean; no host paths; the `_ok` raise-not-assert rationale is documented.

## Verification suggestions

After the inline fix lands: `python -m pytest tests/test_docstring_xref_roles_resolve.py tests/simulation/mujoco/test_randomize_persistence_boundary.py --no-cov -q` — the first is the guard the current head fails, the second is the PR's own suite.

tests/simulation/mujoco/test_randomize_persistence_boundary.py
cagataycali: Confirmed and fixed in `d04a8969`.

```diff
-everything: :meth:`~strands_robots.simulation.base.BenchmarkProtocol.on_episode_start`
+everything: :meth:`~strands_robots.simulation.benchmark.BenchmarkProtocol.on_episode_start`
```

I re-derived the verdict from the branch head rather than from this thread, using the guard's own resolver so the check and the claim cannot disagree:

| target | `_resolves()` |
|---|---|
| `strands_robots.simulation.base.BenchmarkProtocol.on_episode_start` | `False` |
| `strands_robots.simulation.benchmark.BenchmarkProtocol.on_episode_start` | `True` |

and `_offending_roles_in()` over this module's docstring now returns `[]`, where it previously returned exactly the one target you named. `hasattr(base, 'BenchmarkProtocol')` is `False` as you measured; the class is at `benchmark.py:72`. Both modules import in the environment I measured in, so the optional-dep skip carve-out was not laundering the verdict here either.

Two things I checked beyond the one token, since a wrong module path can travel:

- **The quoted prose is still faithful at the new path.** The `"per-episode randomization, goal sampling, or procedural scene generation"` line this module quotes is `on_episode_start`'s own docstring, `strands_robots/simulation/benchmark.py:171` - so the citation now points at the text it paraphrases, not merely at something importable.
- **No other copy of the wrong path exists.** `grep -rn 'simulation\.base\.BenchmarkProtocol' --include=*.py --include=*.md` over the tree returns nothing, so this was the single occurrence rather than one instance of a pattern.

The target stays contiguous on one line at 91 chars (limit 120), so the wrapped-target rule the guard also enforces still holds - the five extra characters did not push it into a break.

cagataycali: Fixed in `d04a896` - the cross-reference now names `strands_robots.simulation.benchmark.BenchmarkProtocol.on_episode_start` (verified to resolve).

---
<sub>This comment was generated by an AI contributor. The technical approach was validated against the repository's test suite.</sub>

yinsong1986 (approved):
## Summary

Docs-and-tests change with no behaviour delta: `randomize()`'s docstring and `docs/simulation/domain-randomization.md` claimed every axis "persists for the rest of the scene's life" and offered a `recompile the scene` undo that no published action provides; both surfaces now name the real boundary (survives `reset()`/`step`, restored by any of the seven scene-mutation actions) and the ordering it implies for per-episode loops. The new `tests/simulation/mujoco/test_randomize_persistence_boundary.py` measures the boundary from the running engine and grades both prose surfaces against the measurement, so the claim cannot drift from the behaviour again. The one blocker from the prior review round (the `simulation.base` vs `simulation.benchmark` cross-reference) is fixed at this head — the xref guard suite passes.

## What's good

- Verified on this head: all 14 new tests pass, the xref guard (`tests/test_docstring_xref_roles_resolve.py` plus the simulation module-xref guard) passes, and the changelog-fragment and no-host-path guards pass.
- The grader derives the wholesale-undo set from the engine (`_wholesale(measured)`) and checks it against `_PUBLISHED_ACTIONS` rather than restating a hand-copied list, so a future scene mutation is graded on arrival — the same derive-the-population shape the repo's other guards use.
- Scope discipline: the only shipped-code edit is the docstring; the spec-carries-authored-values design is asserted as the control (11 of the 14 tests pass on both trees) rather than changed.
- Changelog fragment present under `changelog.d/2577-randomize-persistence-boundary.md` per AGENTS.md step 3; no new non-ASCII in user-facing strings; the `_ok` raise-not-assert rationale is documented in the helper itself.

Issue #1988, opened 2026-08-07
A head commit created through the API with the Actions token spawns zero check suites, so an approved PR holds at BLOCKED forever and reads as reviewer bandwidth


cagataycali:
## Summary

AGENTS.md step 8 says `mergeStateStatus` is the field to trust, and describes its values as
`BLOCKED` **while the required check runs**, then `UNSTABLE` or `CLEAN`. That sentence assumes
the required check will run. When the head commit was created through the GitHub API with the
Actions `GITHUB_TOKEN`, it never does, and `BLOCKED` is terminal rather than transient.

GitHub suppresses workflow triggers for events attributed to `GITHUB_TOKEN`, to stop a workflow
re-triggering itself. A commit written with `createCommitOnBranch` or `PUT .../contents/{path}`
under that token therefore produces **no `push`, no `synchronize`, and no check suite at all** -
not a queued one, not a failing one. The repository's one required context,
`call-test-lint / Test and Lint`, has nothing to report, so the ruleset is never satisfied.

## Measured on #1987

Every commit pushed from a clone got a full suite. The one written through the API got none:

| commit | committer | `checkSuites.totalCount` |
|---|---|---|
| `b8a50a54` | clone | 10 |
| `e8afa76a` | clone | 12 |
| `7b0b53e7` | clone | 12 |
| `b10f4dce` | clone | 13 |
| `a3f3e3f6` | API | **0** |

At `a3f3e3f6` the PR read `reviewDecision: APPROVED` (a review by a *different* account,
post-dating the head, so `require_last_push_approval` was satisfied), `mergeable: MERGEABLE`,
zero unresolved threads - and `mergeStateStatus: BLOCKED`, with `statusCheckRollup: null`.

Nothing in that payload names a cause. It is indistinguishable from a PR whose required check is
still queued, which is why the honest reading of it two cycles running was "waiting on CI" - a
gate that could not move being polled as though it would.

## Why it is expensive

The two obvious remedies both cost more than the bug:

- **Re-push the tree with `PAT_TOKEN`.** This triggers `synchronize`, but
  `dismiss_stale_reviews_on_push` drops the existing approval in the same motion, buying a full
  re-approval round that reviews no changed behaviour.
- **Re-run the checks.** There is no suite to re-run - `totalCount` is `0`, not a red suite - and
  `pr-and-push.yml` has no `workflow_dispatch`, so there is no run to dispatch either.

## The cheap remedy

[...]

Issue #1917, opened 2026-08-03
process: the documented merge gate is viewer-scoped - the Actions token reads BLOCKED forever on any PR that edits .github/workflows/**, and it presents as reviewer bandwidth


cagataycali:
`AGENTS.md` > PR Workflow > step 8 prescribes the merge gate as: poll the required contexts' own conclusions and `mergeStateStatus == CLEAN` together, and says of `mergeStateStatus` that it "is the field that already accounts for the required set, which is why it is the one to trust".

That instruction is correct about the field and silent about the reader. `mergeStateStatus` answers *can the viewer merge this pull request*, so it is scoped to the token that asks - and this agent asks with the Actions `GITHUB_TOKEN` by default. On a pull request that edits `.github/workflows/**` that token can never read anything but `BLOCKED`, however ready the PR is, because an installation token is refused writes to workflow files and therefore genuinely cannot perform the merge.

## Measured

A natural control pair from today, both approved, both with zero unresolved threads and `call-test-lint / Test and Lint` = `SUCCESS`, read minutes apart:

| PR | edits `.github/workflows/**` | `GITHUB_TOKEN` reads | `PAT_TOKEN` reads | truth |
|---|---|---|---|---|
| #1915 | yes (`pr-and-push.yml`) | `BLOCKED` | `CLEAN` | merged clean as `f4dfde6` |
| #1902 | no | `CLEAN` | `CLEAN` | merged clean as `6cf0470` |

On #1915 the two reads were issued back to back at the same instant, and the bot's `BLOCKED` was stable across three polls spanning several minutes, ~45 minutes after the approval landed. `mergeable` agreed (`MERGEABLE`) for both tokens on both PRs - it is only the *state* that diverges, which fits: text-conflict is viewer-independent, mergeability-by-you is not.

## Mechanism, isolated to one variable

Same token, same scratch branch, same instant, via `PUT /repos/{owner}/{repo}/contents/{path}`:

| path written | `GITHUB_TOKEN` | `PAT_TOKEN` |
|---|---|---|
| `zz_probe.txt` | created | - |
| `.github/workflows/zz_probe.yml` | `Resource not accessible by integration` | created |

The token that can write an ordinary file in the same commit series is refused the workflow path. So `BLOCKED` is not a bug and not staleness - it is the honest answer to the question the field asks, addressed to a viewer who cannot merge that particular diff.

## Two ways this misleads, both silent

[...]

PR #2579, opened 2026-08-21
ci: name the ruleset rule a blocked pull request has left unsatisfied

### LINKED ISSUE #2578: process: `mergeStateStatus: BLOCKED` names none of the six rules it stands for, so an author-clearable obligation reads as reviewer bandwidth
## Summary

`mergeStateStatus: BLOCKED` (and REST's `mergeable_state: blocked`) is one word for at least six unrelated situations and names none of them. The rule that is actually unsatisfied is what decides *who owes the next action*, so collapsing them loses the only fact a triage pass needs.

This repository has now recorded four separate causes of the same presentation, three of them under their own issues, and each was diagnosed by hand:

| cause | issue | who clears it |
|---|---|---|
| `require_last_push_approval` satisfied by nobody but the pusher | #1905 | a reviewer other than the pusher |
| the Actions token cannot merge a workflow edit, so it can never read anything but `BLOCKED` | #1917 | re-read with `PAT_TOKEN` |
| a head written through the API spawns zero check suites, so the required context never reports | #1988 | a maintainer, by reopening |
| an unresolved review thread on an approved, green pull request | this issue | **the author** |
| the cached state is simply stale | this issue | **anyone, by retrying** |

The last two are the ones with no detection at all, and both are cheap to get wrong in the expensive direction: they look like waiting on a reviewer, so a scheduled author-side pass correctly declines to merge and reports the pull request as blocked on someone else.

## Evidence, measured 2026-08-21

Both approved, both with every check green, both idle after approval on an obligation that was not a reviewer's:

| PR | approved | merged | idle | actually unsatisfied | who could clear it |
|---|---|---|---|---|---|
| #2566 | 16:39 | 17:10 | 31 min | one unresolved review thread | the author |
| #2574 | 16:25 | 17:10 | 45 min | nothing - a stale computation | anyone, by retrying |

**#2566.** `GET /repos/strands-labs/robots/rules/branches/main` carries `pull_request.required_review_thread_resolution: true`. The pull request read `blocked` purely because one thread was open - a thread whose fix had already landed, whose last comment was the author's, and which the reviewer had approved *past*. Resolving it flipped `blocked` to `clean` on the next read. Nothing on the pull request said the thread was the blocker.

**#2574.** APPROVED, 12 checks SUCCESS, **zero** review threads, and `strict_required_status_checks_policy: false` so being behind the base is irrelevant. Both GraphQL and REST reported `blocked`. `mergePullRequest` refused with `Pull Request is not mergeable`, which names nothing. `PUT /repos/{owner}/{repo}/pulls/2574/merge` then succeeded on the first attempt with no state having changed in between:

```
{"sha": "207f4e0d08d5f6e2465f91f5983308c35f72291c", "merged": true,
 "message": "Pull Request successfully merged"}
```

Two things follow that are worth keeping regardless of what gets built: **a merge attempt is cheap and self-verifying**, so `BLOCKED` with no unsatisfied rule is not a reason to wait; and **REST's refusal names the unmet requirement while the GraphQL mutation's does not**, so REST is the right surface to attempt on.

## Why this is not fixed by what already exists

`scripts/check_last_push_approval.py` answers exactly one rule, deliberately, and neither of the two cases above is that rule - so it reports `satisfied` on both. It is the right shape and the wrong scope.

## Proposed remedy

One missing read, no new policy and no new gate: **read the ruleset and name which of its rules is unsatisfied**, rather than inferring. That distinguishes "waiting on a reviewer" from "waiting on the author" from "waiting on nothing", which is the fact both cases above needed and neither had.

Specifically, a check that reports, per pull request, each unsatisfied rule together with the party who can clear it:

- a conflict, an unresolved thread (with its count), or a failing required check -> the author
- a missing approval -> any reviewer
- an approval only its own pusher supplied -> a different reviewer (defer to #1905's check, do not restate it)
- a required check absent because a fork run is held at `action_required` -> a maintainer
- a required check still running -> nobody
- no unsatisfied rule at all -> anyone, by attempting the merge

A conflict or a draft should be reported as *gating*: the rules behind it cannot be assessed, so an approval there is necessary but not sufficient. #1905 records that distinction being lost as its own failure ("an approving review is necessary but no longer sufficient").

Reading the ruleset rather than hardcoding it also keeps the report honest as settings change, and lets a rule the branch carries but the check cannot answer be named as such - `require_code_owner_review` (vacuous here, there is no CODEOWNERS file) and `require_extra_approval_for_unattributed_changes` are both set on `main` today and neither is answerable from pull request fields.

Not a gate, and not in the required set: the remedy for most of these is somebody else's action, and a gate whose remedy is "find another human" sits red on a branch that has done nothing wrong. Reporting is the useful part.

Related: #1905, #1917, #1988, #2566, #2574.

cagataycali:
## What this changes

Adds `scripts/check_merge_blockers.py`, which reads the branch ruleset and names every rule a pull request leaves unsatisfied, together with the party who can clear it. Nothing is gated and no policy changes.

Closes #2578.

## Why

`mergeStateStatus: BLOCKED` is one word for at least six unrelated situations and names none of them. The rule that is actually unsatisfied is what decides *who owes the next action*, so collapsing them loses the only fact a triage pass needs. Two pull requests measured 2026-08-21, both approved and green, both idle after approval on obligations that were not a reviewer's:

| PR | approved | merged | idle | actually unsatisfied | who could clear it |
|---|---|---|---|---|---|
| #2566 | 16:39 | 17:10 | 31 min | one unresolved review thread | the author |
| #2574 | 16:25 | 17:10 | 45 min | nothing - a stale computation | anyone, by retrying |

#2566 sat on `required_review_thread_resolution`, on a thread whose fix had already landed and which the reviewer had approved *past*. #2574 had zero threads and twelve green checks; `mergePullRequest` refused with `Pull Request is not mergeable`, naming nothing, and the REST merge then succeeded on the first attempt with no state having changed in between.

This repository has now recorded four causes of the same presentation - #1905 (pusher-only approval), #1917 (viewer-scoped token), #1988 (zero check suites), and these two. Each was diagnosed by hand. `check_last_push_approval.py` answers exactly one of them, deliberately, and reports `satisfied` on both cases above: right shape, wrong scope.

## Verdicts on the live queue

Run against all open pull requests, which is the discrimination the change exists for:

```
| pull request | unsatisfied rule(s)                      | owed by      |
| #1035        | merge-conflict, pusher-only-approval     | the author   |
| #1722        | merge-conflict, missing-approval          | the author   |
| #2480        | required-check-pending, missing-approval  | any reviewer |
| #2497        | missing-approval                         | any reviewer |
| #2534        | missing-approval                         | any reviewer |
```

This matches the independent measurement recorded on #1905 on 2026-08-21 (seven awaiting first review, two `DIRTY`), and separates the two rows that no amount of reviewing clears from the seven that a single reviewer clears each.

## Design notes

**The ruleset is read, not hardcoded.** A rule changed in repository settings cannot drift from the report, a rule the branch does not carry is never named as a blocker, and a rule the branch *does* carry but this check cannot answer is named as unevaluated rather than silently dropped. `require_code_owner_review` (vacuous here - there is no CODEOWNERS file) and `require_extra_approval_for_unattributed_changes` are both set on `main` today and neither is answerable from pull request fields.

**Precedence is explicit.** A conflict or a draft is *gating*: the rules behind it cannot be assessed, so the report names one next action instead of a set the reader has to order. #1905 records losing that distinction as its own failure ("an approving review is necessary but no longer sufficient"). Conversely a blocker owed by nobody, such as a check still running, never masks one owed by a person - #2480 is answerable by a reviewer now, and reporting it as owed by nobody would park it.

**`require_last_push_approval` is not re-derived.** That rule, and the semantics of what counts as a current approval (per author, their most recent review that expresses a position, so `COMMENTED` is not a retraction), belong to `check_last_push_approval.py`. This check needs the same primitive to count approvals at all, so it imports it rather than restating it - the repository's own rule for a guard that acquires a second caller (Key Conventions 11). A test asserts the functions are defined in that module, so a future copy-paste fails rather than drifting.

**Not a gate, and not in the required set.** The remedy for most outcomes is somebody else's action, and a gate whose remedy is "find another human" sits red on a branch that has done nothing wrong. No workflow is added either: a report-only check in the same rollup as the required set makes a branch read as broken, which #1905 already complains about. This is a scan-time caller, the same shape as `--all-open` on the sibling.

**The stale-state outcome carries its own caveat.** An Actions installation token reads `blocked` on any pull request touching `.github/workflows/**` whatever the rules say (#1917), so `no-unsatisfied-rule` prints the instruction to re-read with a personal access token before concluding the state is stale. Without that, this check would manufacture the very false reading it exists to prevent.

## Testing

`tests/test_merge_blockers.py`, 59 tests. Fixtures are the real observations, not invented ones: the measured ruleset payload verbatim, and the three-case control where one varied input each produces three different owners. Also pinned - the required check's failing/pending/absent split by who clears it, an outdated thread not being counted, a GraphQL error raising rather than reporting zero threads (which would be a silent pass on the exact rule #2566 was blocked by), a re-run that succeeded not retiring a failing sibling, ASCII-only report strings, and the exit-status contract.

Two bugs were caught by these tests while writing them and are fixed here: the pusher's own approval was discounted even when the branch does not carry `require_last_push_approval`, which invented a blocker; and the sweep applied precedence in one renderer and not the other.

`check_merge_blockers.py` is registered in `_NAMES_REPOSITORY` in `tests/test_duplicate_claim_check.py`, since it reads `$GITHUB_REPOSITORY` and so must declare the flag that names one.

```
pytest tests/test_merge_blockers.py                                             59 passed
pytest <all AGENTS.md-reading process graders>                                 388 passed
pytest tests/test_{merge_blockers,last_push_approval,duplicate_claim_check}.py 171 passed
mypy scripts/check_merge_blockers.py                                           no issues
ruff check / ruff format --check                                               clean
python3 scripts/assemble_changelog.py --check                                  OK
```

The remaining suite failures in this environment are missing optional heavy dependencies (`lerobot`, `mujoco`, `psutil`, `serial`). Verified pre-existing by running the same subset on a clean checkout of `main`: 37 failed / 127 passed both with and without this change.

AGENTS.md records the two new causes and the invocation in the PR Workflow step that already enumerates the others, since a decision recorded only in a comment is not durable.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -948,6 +948,52 @@ hatch run format            # ruff check --fix, ruff format
    still `APPROVED`, nine suites queued including the required one. #1988 has the
    full account.
 
+   **Two further causes need no second token, and one needs nothing at all.**
+   `mergeStateStatus` is one word for at least six situations and names none of
+   them, so the rule that is actually unsatisfied has to be read separately.
+   Measured 2026-08-21, both approved and green, both idle after approval because
+   a scheduled pass read `BLOCKED` and concluded the next move was a reviewer's:
+
+   | PR | approved | merged | idle | actually unsatisfied | who could clear it |
+   |---|---|---|---|---|---|
+   | #2566 | 16:39 | 17:10 | 31 min | one unresolved review thread | the author |
+   | #2574 | 16:25 | 17:10 | 45 min | nothing - a stale computation | anyone, by retrying |
+
+   #2566 sat on `required_review_thread_resolution`, which the ruleset carries:
+   one thread was open, its fix had already landed, and the reviewer had approved
+   *past* it. Resolving it flipped `blocked` to `clean` on the next read. #2574 had
+   zero threads and twelve green checks, and `blocked` was simply stale -
+   `mergePullRequest` refused with `Pull Request is not mergeable`, naming nothing,
+   and `PUT /repos/{owner}/{repo}/pulls/{n}/merge` then succeeded on the first
+   attempt with no state having changed in between.
+
+   Two consequences worth keeping. **A merge attempt is cheap and self-verifying**,
+   so `BLOCKED` on a pull request with no unsatisfied rule is not a reason to wait;
+   and **prefer REST for the attempt**, because its refusal names the requirement
+   that is unmet while the GraphQL mutation's does not.
+
+   Rather than infer which of these is operating, read it:
+
+   ```
+   python3 scripts/check_merge_blockers.py --repo strands-labs/robots --pr <N>
+   python3 scripts/check_merge_blockers.py --repo strands-labs/robots --all-open
+   ```
+
+   It reads the branch ruleset - so a rule that is changed in settings cannot
+   drift from this file - and names every rule the pull request leaves
+   unsatisfied together with the party who can clear it: a conflict or an
+   unresolved thread or a failing check (the author), a missing approval (any
+   reviewer), an approval only its own pusher supplied (a different reviewer,
+   per #1905), a required check absent because a fork run is held at
+   `action_required` (a maintainer), a check still running (nobody), or no
+   unsatisfied rule at all, which is the #2574 case and the one worth saying out
+   loud. A conflict or a draft is reported as *gating*: the rules behind it
+   cannot be assessed, so an approval there is necessary but not sufficient.
+
+   It composes `check_last_push_approval.py` rather than restating it, so what
+   counts as a current approval has one owner. Neither script gates a merge.
+   Pinned by tests/test_merge_blockers.py.
+
    This is worth the words because the failure mode is silent and expensive in the
    opposite direction from the usual one. Treating an advisory red as a merge
    blocker does not look like a mistake; it looks like diligence, and it costs a
--- a/changelog.d/2578-name-the-unsatisfied-merge-rule.md
+++ b/changelog.d/2578-name-the-unsatisfied-merge-rule.md
@@ -1,0 +1,15 @@
+### Added: a blocked pull request's report names the rule it left unsatisfied
+
+`scripts/check_merge_blockers.py` reads the branch ruleset and names every rule a
+pull request leaves unsatisfied, together with the party who can clear it -- a
+conflict, an unresolved review thread or a failing check (the author), a missing
+approval (any reviewer), an approval only its own pusher supplied (a different
+reviewer), a required check held at `action_required` (a maintainer), a check
+still running (nobody), or no unsatisfied rule at all, which means the state is
+stale and the merge is worth attempting.
+
+`mergeStateStatus: BLOCKED` is one word for all of these and names none of them,
+so two approved and green pull requests sat idle after approval on obligations
+their own authors could have cleared. Takes `--pr` for one pull request or
+`--all-open` to sweep the queue; composes `check_last_push_approval.py` so what
+counts as a current approval keeps one owner, and gates nothing.
--- a/scripts/check_merge_blockers.py
+++ b/scripts/check_merge_blockers.py
@@ -1,0 +1,882 @@
+#!/usr/bin/env python3
+"""Name the branch-ruleset rule a blocked pull request has left unsatisfied.
+
+Why this exists
+---------------
+``mergeStateStatus: BLOCKED`` is one word for at least six unrelated
+situations, and it names none of them. The rule that is actually unsatisfied
+decides who owes the next action, so collapsing them loses the only fact a
+triage pass needs. Three cases measured in this repository on 2026-08-21, all
+reading ``blocked``, all needing different people:
+
+===========  ==========================================  ====================
+pull request what was actually unsatisfied               who could clear it
+===========  ==========================================  ====================
+#2566        one unresolved review thread, whose fix     **the author**
+             had already landed and which the reviewer
+             had approved past
+#2574        nothing at all -- the ``blocked`` was a      anyone, by retrying
+             stale computation, and the merge succeeded
+             on the first attempt with no state change
+#2497        no approving review yet                     any reviewer
+===========  ==========================================  ====================
+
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Adds `scripts/check_merge_blockers.py`, a report-only diagnostic that reads the branch ruleset (`GET /repos/{owner}/{repo}/rules/branches/{ref}`) and names every rule an open pull request leaves unsatisfied, together with the party who can clear it — separating the author-clearable states (conflict, draft, failing check, unresolved thread) and the stale-computation state (#2574) from the ordinary waiting-on-review state, all of which read identically as `mergeStateStatus: BLOCKED`. The evaluation core is pure and pinned by 59 tests whose fixtures are the measured ruleset payload and the live pull requests cited in the description; the transport-touching resolvers are exercised through monkeypatched fakes. The design follows the repository's own rules where it matters: approval semantics are imported from `check_last_push_approval.py` rather than restated (Key Conventions 11, with a test that fails on a future copy), the script is registered in `_NAMES_REPOSITORY` for the `$GITHUB_REPOSITORY` inference hazard, report strings are pinned ASCII-only, a lookup failure exits 0 rather than accusing a branch, and nothing is gated. AGENTS.md and the changelog fragment record the two new measured causes durably. No security, crash, data-loss, or one-way-door concerns found in the diff.

## What's good

- The `no-unsatisfied-rule` outcome prints the #1917 token caveat inline, so the check cannot manufacture the false stale-state reading it exists to prevent — and a test pins that.
- Rules the branch carries but the check cannot answer (`require_code_owner_review`, `require_extra_approval_for_unattributed_changes`) are named as unevaluated rather than silently passed.
- A GraphQL error on the thread query raises instead of reporting zero unresolved threads, which would have been a silent pass on exactly the rule #2566 was blocked by.
- Precedence (gating conflict/draft vs. parallel blockers) routes both renderers and both warning paths through one `primary()`, closing the drift the author reports catching while writing the tests.

Issue #218, opened 2026-05-25
security(mesh): close ACL file TOCTOU window between is_default_acl_in_use and resolve_acl


cagataycali:
**Context:** Surfaced in the 2026-05-25 06:42 review on PR #195 (`PRRT_kwDORUMiZs6EeH_B`).

The R18-C identity-tuple cache in `_acl_config.py` is documented as a TOCTOU close, but it only delivers one-shot consistency on stable inputs. If an attacker rewrites the ACL file between `is_default_acl_in_use()` and `resolve_acl()` -- exactly when the TOCTOU matters -- the second `_load_acl_cached` call computes a fresh identity tuple, misses the cache, and re-reads the (now-malicious) file.

The docstring at `_acl_config.py:354-362` says "both functions take the same snapshot; if the file changes mid-flight the next call refreshes" -- the second clause is the bug surface, not the fix.

## Fix options

Reviewer offered two paths:

### Option A (architectural, preferred): thread a single resolved dict through `Mesh.start`

```python
# in Mesh.start, before _refuse_under_permissive_default_acl
resolved_acl = _acl_config.resolve_acl(namespace)
if resolve_auth_mode() == "mtls" and _acl_config._is_permissive_acl_shape(resolved_acl):
    ...
```

One read at the top of `Mesh.start`, dict travels through the gate and into the `acl_block` insertion. Closes the TOCTOU window completely (the attacker would need to rewrite within a single `_load_acl_file` call).

### Option B (minimal): soften the docstring claim

Drop the "if the file changes mid-flight the next call refreshes" sentence. The cache delivers one-shot consistency on stable inputs only; an attacker rewriting between calls still wins. State the actual guarantee.

## Acceptance criteria

- One read of `STRANDS_MESH_ACL_FILE` per `Mesh.start()` call (option A) **OR**
- Docstring on `_acl_config.py:354-362` no longer claims TOCTOU close behaviour (option B).
- Pin test in `tests/mesh/test_default_acl_warning.py` that asserts the chosen behaviour (a counter on `_load_acl_file` calls per `Mesh.start` for option A; a docstring grep for option B).
- Reviewer's recipe in `Mesh.start` is preserved verbatim if option A is chosen (it threads the dict cleanly through the existing call sites).

## Why a separate PR

PR #195 is at R25 already (round-budget invoked). Folding this in would either:
- expand R25 to 5 fixes instead of 4 (T2 has architectural choice that warrants its own review round); or
[...]

PR #2580, opened 2026-08-21
docs(mesh): describe the ACL cache that ships, not the one it replaced

cagataycali:
## Problem

`strands_robots/mesh/_acl_config.py::_load_acl_cached` is the single cache the ACL
file is read through, and its docstring is where a reader auditing that path learns
three things. All three had drifted from the code, and two of them in the direction
that overstates the guarantee.

**1. The caller census named a superseded pair.** The docstring said *"Two callers in
the same `Mesh.start` flow (the gate check and the config builder)"* - the pre-#218
`is_default_acl_in_use` + `resolve_acl` pattern. Measured from the module's own AST,
three functions call the cache, and the two the docstring named are reached from
nowhere outside the module:

| in-module reader of the cache | call sites elsewhere in `strands_robots/` |
|---|---|
| `snapshot_acl` | `mesh/core.py:543`, `mesh/session.py:574` |
| `resolve_acl` | none |
| `is_default_acl_in_use` | none |

So the caller the `Mesh.start` flow actually goes through is `snapshot_acl`. The
module comment directly above the cache already describes that pair in the past
tense (*"In an earlier revision `Mesh.start` called ..."*), and
`tests/mesh/test_acl_snapshot_toctou.py` calls it *"the previous two-call pattern"* -
this docstring was the one place still presenting it as current.

**2. Callers were said to "get the same dict object".** Every return path
deep-copies, and deliberately: one of the deep-copy comments in the body records
that returning the parsed dict directly is exactly what it replaced. Two reads of
one file:

```
same object   : False
equal contents: True
```

Callers share the file's *contents*; identity is precisely what they do not share,
so that a caller mutating what it got can poison neither the cache nor another
caller's dict.

**3. It claimed to close "the prior TOCTOU surface".** `snapshot_acl` documents this
identity-keyed cache as its **second** tier, whose reload on a rewritten file is a
*"by-design refresh window"*, and attributes the defence to its thread-local
snapshot. Two docstrings in one module disagreed about which mechanism holds the
line, and the weaker one claimed it.

## Change

One docstring, rewritten to describe the cache that ships: three callers named, only
`snapshot_acl` marked as the one wired from outside the module, contents-not-identity
stated with a pointer to the deep-copy comments, and the TOCTOU window attributed to
`snapshot_acl`'s thread-local snapshot with this tier named as the one that refreshes.

No behaviour change - the cache, its lock, its cap and its deep-copy contract are
untouched.

## Tests

`tests/mesh/test_acl_cache_docstring_matches_the_cache.py` pins the docstring
against the code rather than against a copy of it. The caller census and the caller
names are derived from the module's own AST, so a fourth reader of the cache is
graded the moment it is added; the identity claim is checked by reading one ACL file
twice.

Against `upstream/main` with only the new test file added:

| tree | failed | passed |
|---|---|---|
| `upstream/main` | 4 | 9 |
| this branch | 0 | 13 |

The four that flip are the three drifted claims plus the caller census. The nine that
pass on both sides are the controls: the two deep-copy behaviour tests (the
measurement the docstring has to agree with, not a change in what the cache does),
the premise that only `snapshot_acl` is wired from outside the module, and the
parser/scan non-vacuity cases - including a planted-count case, so a census parser
that silently matched nothing could not look like a clean tree.

## Verification

Measured at `73298f10`, on the same 193-file selection in both trees (every test naming
the ACL readers or `_acl_config`, all of `tests/mesh`, and the repo-wide docstring /
string / changelog / export guards). The selection is identical on both sides, so any
difference would be caused by the source change alone:

|---|---|---|---|---|
| `upstream/main` | 21 | 4100 | 5 | 24 |
| this branch | 21 | 4100 | 5 | 24 |

45 failing ids, the same set both ways (`comm` empty in both directions). All 45 are
`tests/mesh/test_iot_*` needing `awsiotsdk`, which is a declared extra and installed in
CI. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`;
`mypy` 24 errors on both trees with an empty branch-only set, none in the changed files.

## Artifact

Every row is read from the tree it runs in: one capture script per tree extracts that
tree's own docstring, grades each claim against the code, and dumps the pair. The
measured (code-side) half is byte-identical in both trees - asserted inside the
figure script - so nothing about the cache changed, only whether its docstring
describes it. Wrong claims 4 -> 0.

ACL cache docstring claim matrix

### DIFF
--- a/changelog.d/2580-acl-cache-docstring-matches-the-cache.md
+++ b/changelog.d/2580-acl-cache-docstring-matches-the-cache.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- **mesh**: `_load_acl_cached`'s docstring now describes the ACL cache that ships. It
+  named the superseded pre-#218 `is_default_acl_in_use` + `resolve_acl` pair as its two
+  callers (three functions call it, and only `snapshot_acl` is wired from outside the
+  module), said callers "get the same dict object" (every return path deep-copies, so
+  they share contents and not identity), and claimed to close the `Mesh.start` TOCTOU
+  window that `snapshot_acl` attributes to its thread-local snapshot while calling this
+  identity-keyed tier a by-design refresh window. The caller census and caller names are
+  now graded against the module's own AST, so a fourth reader of the cache is reported
+  when it is added.
--- a/strands_robots/mesh/_acl_config.py
+++ b/strands_robots/mesh/_acl_config.py
@@ -526,12 +526,27 @@ def _file_identity(path: Path) -> tuple | None:
 
 
 def _load_acl_cached(path: Path) -> dict[str, Any]:
-    """Load + cache an ACL file, keyed on its identity tuple.
+    """Load an ACL file once per file identity, keyed on its identity tuple.
 
-    Two callers in the same ``Mesh.start`` flow (the gate check and the
-    config builder) get the same dict object instead of two independent
-    reads -- closing the prior TOCTOU surface. If the file changes a
-    later call computes a fresh identity tuple and re-loads.
+    Three callers read the file through here: :func:`snapshot_acl`,
+    :func:`resolve_acl` and :func:`is_default_acl_in_use`. Only
+    :func:`snapshot_acl` is reached from outside this module - by the
+    ``Mesh.start`` shape gate and by the wire-config builder - so it is the
+    one whose two reads per flow this cache collapses. The other two are the
+    single-read spellings of the superseded two-call pattern the comment above
+    describes, which no caller outside this module wires any more.
+
+    Callers share the file's *contents*, not one object: every return path
+    deep-copies, so a caller mutating what it got can poison neither the cache
+    nor another caller's dict. The identity is deliberately not shared - the
+    deep-copy comments below say why, and one of them records that returning
+    the parsed dict directly is what this replaced.
+
+    This is the identity-keyed tier, not the TOCTOU defence. A file rewritten
+    between two reads computes a fresh identity tuple and re-loads, which is
+    the by-design refresh window :func:`snapshot_acl` documents; what closes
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Docs-only correction of `_load_acl_cached`'s docstring in `strands_robots/mesh/_acl_config.py`, plus a regression test and a changelog fragment. The old docstring made three claims that had drifted from the code, two of them overstating the guarantee: it named the superseded pre-#218 `is_default_acl_in_use` + `resolve_acl` pair as the cache's two callers, promised callers "the same dict object" when every return path deliberately deep-copies, and claimed to close the `Mesh.start` TOCTOU window that `snapshot_acl` attributes to its thread-local snapshot. I verified each corrected claim against the tree at the head SHA: three in-module functions call the cache; only `snapshot_acl` is called from outside the module (`mesh/core.py:543`, `mesh/session.py:574`); all three return paths deep-copy; and `_set_thread_snapshot` / the "by-design refresh window" wording in `snapshot_acl` exist as the new docstring says. The cache body (lock, 4-entry cap, deep-copy contract) is byte-untouched. All 13 tests in the new file pass on the PR head.

## What's good

- The test pins the docstring against the code rather than a copy of it: the caller census and names are derived from the module's own AST, so a fourth reader of the cache fails the census the moment it is added.
- Non-vacuity controls (planted-count census parses, the scan finds the real readers) prevent a parser that silently matches nothing from reading as a clean tree.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention (AGENTS.md step 3); no host paths, no non-ASCII, `monkeypatch`-free fixture correctly clears the module cache around each test.
- The `python -O` comment on the hoisted cache-populating call is a nice detail: side effects are kept out of `assert` expressions.

Issue #2520, opened 2026-08-20
process: a review thread got four answers to one question, because each scheduled run re-derived the fix from the thread instead of from the branch head


cagataycali:
## What happened

One `[MUST FIX]` thread on #2511 (`tests/simulation/test_predicate_tolerance_sign_domain.py:87`, the `"str" in annotation` substring match) received **four author replies** inside 27 minutes, each independently reproducing the same 6 failures, each announcing the same fix:

| reply | time (UTC) | what it claimed |
|---|---|---|
| 1 | 04:19:23 | fixed in `c2969d4`, whole-annotation match, 6 failed / 104 passed reproduced |
| 2 | 04:31:30 | fixed in `c2969d4` + `adb2cb40`, same 6 failures reproduced again |
| 3 | 04:46:38 | fixed in `c2969d4`, `1bea0b6` merges main |
| 4 | (this run, not sent) | would have been a fourth reproduction plus a **redundant fifth commit** |

The fix itself was correct on the first attempt. Everything after 04:19 was re-work.

## Mechanism

A scheduled run reads the thread, sees `isResolved: false` with an unanswered `[MUST FIX]`, and treats that as "this needs fixing". It is not. The thread is a statement about a **commit**, and the only thing that answers "is this still true" is the **branch head** - which by 04:26 already contained `c2969d49` and `adb2cb40` fixing it.

The cost compounds because a review thread is append-only: a duplicate commit can be dropped locally at zero cost (this run dropped one), but a duplicate *reply* is permanent and has to be read by every subsequent reviewer and every subsequent run - which is itself an input that makes the next run more likely to add a fifth.

This is the same class as #2507 (two runs appending to one AGENTS.md list) but on a different surface: there the collision is textual and `git` reports it, here the collision is semantic and nothing reports it at all.

## What the ordering should be

Cheap, and entirely on the read side - no coordination or locking:

1. `git fetch` the PR head **before** deciding a thread is actionable, and diff the head against the thread's `originalCommit`. If the head already changes the lines the thread names, the thread is answered; verify that and stop.
2. Re-read `reviewThreads.comments` in full, not just the first comment. A thread whose last comment is an author reply naming a commit that exists in the head is closed business, whatever `isResolved` says.
3. `isResolved: false` is not a work signal on its own. Resolution is a *reviewer* action, so an unresolved-but-answered thread is the normal steady state between a push and the next review pass.

## Evidence

[...]

PR #2581, opened 2026-08-21
ci: report which review threads are still owed an author reply

cagataycali:
## Problem

A review thread is append-only, so a run that rebuilds its context each cycle cannot tell
*unanswered* from *answered, and the answer is in the payload I just read*. The reviewer's
question sits there verbatim either way. Nothing in the serialised thread says "answered".

So the same thread gets answered repeatedly:

| pull request | thread | author replies | what the replies after the first added |
|---|---|---|---|
| #2511 | `PRRT_kwDORUMiZs6arq8z` | 3 | nothing - same fix, same commit (`c2969d4`), the same 6 failures reproduced three more times, over 27 minutes |
| #2577 | `PRRT_kwDORUMiZs6bQMZz` | 2 | nothing - `d04a8969` was already named by the first reply, 10 minutes earlier |

The cost is not the wasted run. A duplicate reply is **permanent**: every later reviewer and
every later run has to read it, and reading it is what makes the next run add one more.
#2520 records the loop.

## Why the rule that already exists did not stop it

AGENTS.md is not silent here - it already says a resolved/outdated thread is terminal, and
that a thread whose last non-bot comment is yours has been answered. #2577's thread was
**both** `isResolved: true` and `isOutdated: true` when it took its third comment at
19:28:20Z, and its second and third comments were both the author's. Every field needed to
refuse that reply was in the payload, and had been fetched.

That is the argument for a command rather than another paragraph: the rule asks a reader to
re-derive one boolean from a payload that also contains a reviewer's question addressed to
them. The same reasoning put `check_duplicate_claim.py` in `scripts/` after prose alone
failed to prevent three duplicate claims.

## Change

`scripts/check_thread_is_answered.py` - one outcome per thread, one exit code per pull
request, no clone:

```
python3 scripts/check_thread_is_answered.py --repo strands-labs/robots --pr 2577
python3 scripts/check_thread_is_answered.py --repo strands-labs/robots --all-open
```

| outcome | meaning |
|---|---|
| `settled` | `isResolved`. A reviewer closed the business. |
| `answered` | Open, last non-bot comment is the author's. Next move is the reviewer's. |
| `awaiting-the-author` | Open, last non-bot comment is someone else's (or only bots). **The only outcome that is work, and the only one that exits 1.** |

It reports and does not gate, wired to no workflow, like its three agent-invoked siblings:
what an author should do next is not something a branch can turn green.

## One rule decides it; two tempting fields are reported instead

**Whoever spoke last owes the next move, unless a reviewer resolved the thread.**

**`isOutdated` is not consulted**, and both directions of its unreliability are measured
here. It describes the diff, not the conversation:

- It is `false` on threads that *are* answered. #2480's still reads `isOutdated: false`
  after `e83cf51` fixed the finding, because the fix added a method rather than rewriting
  the commented line.
- A thread keeps taking comments after it flips to `true`. #2577's is `isOutdated: true` and
  took two more comments anyway. Both happened to be the author's, but nothing about the
  flag stops a reviewer's arriving the same way - and a rule reading outdated as terminal
  would file that demand as settled.

Authorship subsumes what the proxy was reaching for. The AGENTS.md bullet calling outdated
terminal is corrected in place rather than contradicted by a second rule.

**Whether the head moved is reported, not decided on.** An author reply that explains rather
than changes is a complete answer that leaves the head exactly where the thread was written;
folding that into the verdict would call it unanswered and license the duplicate reply this
exists to prevent.

It is reported because of a measurement that kills the obvious implementation:
`originalCommit` is a property of the **thread**, not of the comment. All four comments on
#2511 - one reviewer comment and three author replies spanning 04:19 to 04:46 - report the
same `originalCommit` `ee3e4526`, while the head by the last reply was `0a69634d`:

| comment | author | `originalCommit` |
|---|---|---|
| 1 (04:19, reviewer) | reviewer | `ee3e4526` |
| 2 (04:19, reply) | author | `ee3e4526` |
| 3 (04:31, reply) | author | `ee3e4526` |
| 4 (04:46, reply) | author | `ee3e4526` |

So a reply carries no record of the commit it was verified against, and `headRefOid` is the
only commit a thread can be keyed to - which is what #2520 asks for, to tell "already fixed
at `<oid>`" apart from "not yet fixed".

## Tests

`tests/test_thread_is_answered.py`, **41 passed**. Every payload is the measured shape of a
real thread with the resolution flag set to what it was in the window the check runs in.

Coverage comes from replayed incidents rather than a historical sweep, deliberately - see
Verification: on a finished pull request `settled` is the only reachable outcome, so a
historical sweep *cannot* contain the two discriminating ones.

The non-vacuity pin is `test_flipping_only_the_last_comments_author_flips_the_outcome`: same
thread, same head, same commit, same comment count, only the final login differs. A check
that always agreed would not move there. Also pinned: a reviewer demand on an outdated
thread is owed; a resolved thread is settled whoever spoke last; a bot comment after the
author's reply does not reopen it (`github-advanced-security` posts into threads on #2480
and all four of #1722); a bot-only thread is owed; an answer with no commit after it is
still answered; `head_moved` changes no verdict; the thread commit is read from the thread
and not from a reply; unread threads past the ceiling are named rather than cleared; a
lookup failure is never a finding.

## Verification

**Driven against live data, not only fixtures.**

`--pr 2577` -> `nothing-owed`, exit 0, naming the thread `settled`, head `d04a8969`, thread
commit `b966ce64`, head moved `yes`.

`--all-open` over all 11 open pull requests -> `nothing-owed` throughout, exit 0.

**Steady state, and why the sweep is quiet.** Over the 150 most recently updated closed pull
requests: 26 review threads found, **all 26 `isResolved`** and therefore `settled`, all 150
`nothing-owed`. That is the expected result rather than a broken check - threads here do get
resolved, so the two discriminating outcomes exist only in the window between a reviewer's
comment and its resolution. That window is exactly when a scheduled run reads, and exactly
when the wrong answer costs a permanent duplicate reply.

**The incidents are attributable on timestamped fields alone**, which is why authorship
carries the rule: thread resolution has no timestamp in the API, so "was it resolved yet" is
not answerable after the fact. But at 19:28:20Z on #2577 the last non-bot comment was the
author's own reply from 19:17:50Z, and at 04:31 and 04:46 on #2511 it was the author's reply
from 04:19. The authorship test returns `answered` for all three with no appeal to a field
whose history cannot be read back.

**Gate.** `ruff check` and `ruff format --check` clean on both new files. Repo-wide graders
that walk the source or test tree: **530 passed, 1 failed**, and the failure is
`test_docstring_xref_roles_resolve.py::test_qualified_strands_robots_xref_roles_resolve`
with `ImportError: No module named 'cv2'` - reproduced identically on a pristine
`origin/main` worktree, so it is this box lacking an optional dep rather than a branch
finding. CI installs `.[all,dev]`.

**Duplication check before writing any of it.** `check_merge_blockers.py` is the only
existing reader of `reviewThreads` and it collapses them to a scalar count of live threads
(`resolve_unresolved_threads`), discarding every per-thread field. `originalCommit`, thread
comments and comment authors are read **nowhere** in the repository.

## Scope - why #2520 stays open

#2520 has two halves. This lands the read: a thread's state as a verdict instead of a
re-derivation. It does **not** address the other half its second comment records - two
parallel runs each independently deriving the same fix (#2566), where both would correctly
see `awaiting-the-author` and both would still do the work. That needs something about
claiming a thread rather than reading one, so the issue stays open and this only
cross-references it.

Towards #2520.

<sub>An earlier revision of this description phrased that heading with a closing keyword in
front of the issue number, which GitHub linked as a closing reference - it does not read the
surrounding negation. Reworded so merging this cannot auto-close an issue it only half
answers.</sub>

## Round 1

Two graders on `main` pinned premises this branch invalidates, and both failed on `acc21ea`.
Found by running them against the branch tree rather than by reading the rollup -
`call-test-lint` was still in flight, so the branch was never green.

| grader | what it pinned | why this branch broke it |
|---|---|---|
| `tests/test_review_thread_reply_gate.py` | `` `isResolved` or `isOutdated`** -> do not reply `` as a single directive | step 5 now separates the two fields, so that literal is absent |
| `tests/test_duplicate_claim_check.py` | derived `_INFERS_REPOSITORY` == declared `_NAMES_REPOSITORY` | the new script reads `$GITHUB_REPOSITORY`, so it joined the derived set only |

- `1b5c05e` replaces the invalidated premise rather than deleting it. The terminal directive
  is pinned on `isResolved` alone, and the `isOutdated` correction gets its own test
  asserting the two halves that must travel together: the explicit denial, and the
  disposition that replaces it. A denial that removes the rule without supplying one leaves
  an outdated thread undecided - the failure the correction exists to prevent. All three
  literals are absent from the pre-correction prose, so none is vacuous and re-merging the
  fields fails here.
- `d5cf250` declares `check_thread_is_answered.py` as `("--repo", None)`: both modes reach
  the API, so both owe the flag and there is no argv marker to scope it to. Verified
  load-bearing rather than decorative - dropping `--repo` from the documented sweep
  invocation fails naming that invocation.

Regression sweep: the 33 top-level AGENTS.md graders plus this branch's own tests ->
**892 passed**. The only failures need `serial` / `torch` / `strands` optional deps this box
lacks and reproduce identically on a pristine `origin/main`, per the base-first comparison
AGENTS.md requires.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -291,9 +291,15 @@ hatch run format            # ruff check --fix, ruff format
    - Thread's last non-bot comment is **yours** -> do not reply. You have
      already said it. If there is code to push, push it; the push is the
      message.
-   - Thread is **`isResolved` or `isOutdated`** -> do not reply. Resolution is
-     terminal. Reopening it to restate a landed fix reads as noise, not
-     diligence.
+   - Thread is **`isResolved`** -> do not reply. Resolution is a reviewer action
+     and terminal. Reopening it to restate a landed fix reads as noise, not
+     diligence. `isOutdated` is **not** terminal, measured: it describes the diff
+     rather than the conversation, and a thread keeps accepting comments after it
+     flips - #2577's took two more after `d04a8969` moved its lines. It is also
+     `false` on threads that *are* answered, when the fix adds lines elsewhere
+     (#2480, still `false` after `e83cf51` fixed it). Decide an outdated thread on
+     authorship like any other, or a reviewer's new demand on a moved line is
+     filed as settled.
    - Last comment is **someone else's** and your existing replies do not answer
      it -> reply once, then resolve.
 
@@ -301,6 +307,26 @@ hatch run format            # ruff check --fix, ruff format
    those twelve comments on its own: no semantic comparison, just the author of
    the last comment. Both `isResolved` and the comment authors are already in
    the context payload - they were fetched and not read.
+
+   **Ask for the verdict rather than re-deriving it.** "Fetched and not read" is
+   the whole failure, and it recurred after this rule was written: #2511 took
+   four author replies to one question and #2577 took two, with every field
+   needed to refuse them present in the payload each time. Re-deriving one
+   boolean from a payload that also contains a reviewer's question addressed to
+   you is the part that does not survive a context rebuild, so ask a command:
+
+   ```
+   python3 scripts/check_thread_is_answered.py --repo strands-labs/robots --pr <N>
+   python3 scripts/check_thread_is_answered.py --repo strands-labs/robots --all-open
+   ```
+
+   `settled` and `answered` are not work. `awaiting-the-author` is, and it is the
+   only outcome that exits 1 - so the sweep answers "which of my open pull
+   requests actually need me" in one read. The report also names the commit each
+   thread was written against beside the pull request's head, which is what tells
+   "already fixed at `<oid>`" apart from "not yet fixed" without re-deriving the
+   fix (#2520). It reports and does not gate: what an author should do next is not
+   something a branch can turn green.
 
    What makes this worth writing down is that the previous rule was *satisfied*
    by all twelve. "Address all review comments", and "reply when a thread asks
--- a/changelog.d/2581-thread-answered-check.md
+++ b/changelog.d/2581-thread-answered-check.md
@@ -1,0 +1,32 @@
+### Added: a command that says which review threads are still owed an author reply
+
+`scripts/check_thread_is_answered.py` reports, per review thread on a pull request,
+whether anything is owed: `settled` (a reviewer resolved it), `answered` (the last
+non-bot comment is the author's, so the next move is the reviewer's), or
+`awaiting-the-author` - the only outcome that is work, and the only one that exits
+`1`. `--all-open` sweeps every open non-draft pull request, answering "which of
+these actually need me" in one read.
+
+One rule decides it: whoever spoke last owes the next move, unless a reviewer
+resolved the thread. A thread is append-only, so a reviewer's question stays in it
+verbatim after it is answered and its presence is not evidence that anything is
+outstanding - #2511 collected four author replies to one question in 27 minutes and
+#2577 collected two, each announcing a commit the previous reply had already named.
+
+`isOutdated` is reported but not consulted, because it describes the diff rather
+than the conversation and is unreliable in both directions: it stays `false` on an
+answered thread whose fix added lines elsewhere (#2480 after `e83cf51`), and a
+thread goes on accepting comments after it flips to `true` (#2577 took two more),
+so reading it as terminal files a reviewer's new demand on a moved line as settled.
+The AGENTS.md rule that called it terminal is corrected to match.
+
+Whether the pull request's head has moved past the commit a thread was written
+against is reported beside each thread but decides nothing - an author reply that
+explains rather than changes is a complete answer. It is reported because
+`originalCommit` belongs to the thread and not to the comment (identical on all
+four comments of #2511, across two pushes), so `headRefOid` is the only commit a
+thread can be keyed to when telling "already fixed at `<oid>`" from "not yet
+fixed".
+
+Reports and does not gate, and is wired to no workflow: what an author should do
+next is not something a branch can turn green.
--- a/scripts/check_thread_is_answered.py
+++ b/scripts/check_thread_is_answered.py
@@ -1,0 +1,648 @@
+#!/usr/bin/env python3
+"""Report which of a pull request's review threads are still owed an author reply.
+
+Why this exists
+---------------
+A review thread is append-only, and an agent that rebuilds its context each run
+cannot tell "this question is unanswered" from "this question is answered and the
+answer is sitting in the payload I just read". The reviewer's question is still
+there verbatim either way. Nothing in the serialised thread says *answered*.
+
+So the same thread gets answered repeatedly. Measured on two threads in this
+repository:
+
+============  ==========================  ===============  =========================
+pull request  thread                       author replies   what the replies after
+                                                            the first one added
+============  ==========================  ===============  =========================
+#2511         ``PRRT_kwDORUMiZs6arq8z``    3                nothing -- the same fix,
+                                                            the same commit
+                                                            (``c2969d4``), the same
+                                                            6 failures reproduced
+                                                            three more times, over
+                                                            27 minutes
+#2577         ``PRRT_kwDORUMiZs6bQMZz``    2                nothing -- ``d04a8969``
+                                                            was already named by the
+                                                            first reply, 10 minutes
+                                                            earlier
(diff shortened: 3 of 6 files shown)


yinsong1986 (dismissed):
## Summary

Adds `scripts/check_thread_is_answered.py`, a report-only sibling of `check_pr_head_is_current.py` / `check_last_push_approval.py` that classifies each review thread on a PR as `settled` / `answered` / `awaiting-the-author`, with exit 1 only for the last, and corrects the AGENTS.md bullet that treated `isOutdated` as terminal (with measurement in both directions). The decision rule is deliberately narrow — resolution outranks everything, otherwise last non-bot author decides — and the two tempting proxies (`isOutdated`, head movement) are reported but not consulted, each exclusion pinned by a test.

## What's good

- Conventions match the existing sibling scripts exactly: same flag names (`--repo` / `--pr` / `--token` / `--all-open`), same env-var defaults, same fail-open-on-lookup-failure semantics ("a lookup failure is never a finding"), same mutual-exclusion refusal for `--pr` + `--all-open`.
- No injection surface: the repo string is validated (`owner/name`) and passed as GraphQL variables rather than interpolated; the API host is fixed; no subprocess.
- Read ceilings are directional on purpose — comments from the tail (only the last non-bot comment decides), threads from the head with the shortfall *named* rather than silently cleared, and both properties are pinned.
- Changelog fragment present per AGENTS.md step 3; issue reference worded as `Towards #2520` so the merge cannot auto-close a half-answered issue.
- Test suite (41 passed locally) replays the measured incident payloads and includes a real non-vacuity pin (flipping only the final comment's login flips the outcome).

## Verification suggestions

- Live spot-check beyond fixtures: `python3 scripts/check_thread_is_answered.py --repo strands-labs/robots --pr 2577` should report `nothing-owed` with the thread `settled`, and `--all-open` should exit 0 while every open PR's threads are settled or answered.

yinsong1986 (approved):
## Summary

Adds `scripts/check_thread_is_answered.py`, a report-only sibling of `check_pr_head_is_current.py` / `check_last_push_approval.py` / `check_merge_blockers.py` that classifies each review thread on a pull request as `settled` / `answered` / `awaiting-the-author` (only the last is work, and only it exits 1), with an `--all-open` sweep. The decision rule is deliberately narrow -- resolution outranks everything, otherwise the last non-bot author decides -- and the two tempting proxies (`isOutdated`, head movement) are reported but not consulted, each exclusion measured and pinned. The AGENTS.md step-5 bullet that read `isOutdated` as terminal is corrected in place, and the two graders whose premises that invalidated are updated in the same push (the Round 1 delta since the dismissed review is exactly those two test files).

## What's good

- Conventions match the sibling scripts exactly: same flag names and env-var defaults, same fail-open-on-lookup-failure semantics (a lookup failure is never a finding), same mutual-exclusion refusal for `--pr` + `--all-open`, and the `("--repo", None)` declaration in `test_duplicate_claim_check.py` keeps the derived/declared sets equal.
- No injection surface: repo string validated as `owner/name`, values passed as GraphQL variables, fixed API host, no subprocess.
- Read ceilings are directional with the failure mode in mind -- comments from the tail (only the last non-bot comment decides), threads from the head with the shortfall named rather than silently cleared, both pinned.
- The non-vacuity pin (`test_flipping_only_the_last_comments_author_flips_the_outcome`) and the AGENTS.md correction pin (denial + replacement disposition travel together) are real tests of behaviour, not restatements.
- Changelog fragment present per step 3; the issue reference is worded `Towards #2520` so the merge cannot auto-close a half-answered issue.
- Verified locally at `d5cf250c`: the three touched test files pass (113), the neighbouring workflow graders pass (306), and `ruff check` / `ruff format --check` are clean on both new files.

## Verification suggestions

- Live spot-check beyond fixtures: `python3 scripts/check_thread_is_answered.py --repo strands-labs/robots --pr 2577` should report `nothing-owed` with the thread `settled`, and `--all-open` should exit 0 while every open PR's threads are settled or answered.

PR #2582, opened 2026-08-21
docs(policies): the documented spelling recipe is the registry, not the factory

cagataycali:
## Problem

`list_providers()` and `list_aliases()` report the two policy registries. Three surfaces told a
caller their union was *every spelling `create_policy` accepts*, and offered it as a recipe -
`list_aliases`' own docstring, the `docs/api-reference.md` row, and the `docs/policies/overview.md`
paragraph:

```python
accepted = set(list_providers()) | set(list_aliases())   # 28 spellings
```

`import_policy_class` also falls back to auto-discovery, so a module under `strands_robots.policies`
that exports a `Policy` subclass resolves under its own module name with no registry entry. Two ship,
and the union omits both. Measured on `aa1c5b8e`:

| spelling | in the documented union | `import_policy_class` resolves | `create_policy` builds |
|---|---|---|---|
| `random` (a registry alias) | yes | yes | yes |
| `composite` | **no** | yes -> `CompositePolicy` | **yes** |
| `persistent` | **no** | yes -> `PersistentPolicy` | no (see below) |
| `base`, `factory` | no | no - `Unknown policy provider` | no |

So a caller who validates a provider name against the documented recipe - which is what these
surfaces told them to do - rejects `composite`, and `create_policy("composite", lower=..., upper=...)`
builds a `CompositePolicy`. Both wrappers are documented in their own right
(`docs/policies/persistent-worker.md` is a whole page; `docs/policies/wbc.md` has a composite
section), so the spelling is reachable and the recipe is what hides it.

## Why the prose, and not the discovery surfaces

Reporting the two as providers would be wrong rather than merely wider:

- **`persistent` cannot be built through the factory at all.** `PersistentPolicy.__init__`'s first
  parameter is named `provider`, which `create_policy` has already bound, so no keyword can reach it -
  `create_policy("persistent")` raises `TypeError: missing 1 required positional argument: 'provider'`.
  Listing it would advertise a provider the factory refuses.
- **The package already records the property the recipe denied.** `mesh/security.py`, explaining why
  its policy-provider allowlist stays a literal: *"A provider this package can build but does not list
  here is refused on the mesh `tell()` path"*. Buildable is already a wider set than registry-listed.
- `docs/policies/overview.md`'s provider table is pinned to exactly `policies.json`'s canonical keys by
  `tests/test_docs_policy_coverage.py`, whose docstring calls it *"the human-facing catalogue of what
  `create_policy("<name>")` accepts"*.

Code, tests and the registry agree; the three prose surfaces did not. So the surfaces keep reporting
the registry, and the prose names the second route.

## Change

- `list_aliases`' docstring: the union is every **registered** spelling (the recipe's variable is
  renamed `accepted` -> `registered`), then names the auto-discovery route and both spellings it
  reaches, including why `persistent` is constructed directly.
- `docs/api-reference.md`: same correction in the `list_aliases()` row.
- `docs/policies/overview.md`: the enumeration paragraph now says *registered*, and a new
  **Beyond the registry** subsection documents the two wrappers with links to the pages that already
  cover them. Both anchors verified to resolve.

No behaviour change: `strands_robots/policies/factory.py` is docstring-only, and no other file under
`strands_robots/` is touched.

## The guard could not see this

`tests/policies/test_provider_alias_discovery.py` owns this contract - its module docstring opens
*"Every provider spelling `create_policy` accepts is discoverable"* - but its `_accepted_spellings()`
reads `policies.json`, so it grades the registry only. **On the unfixed tree the shipped guard reports
`10 passed`** with `composite` accepted and reported by nothing.

It now derives the unregistered spellings from the resolution path (probing only module names the
surfaces do *not* already report, so it needs no optional dependency) and grades every surface that
teaches enumeration for naming them. A third such module is held to the same rule the day it lands.
The match is whole-token: a dotted path like `strands_robots.policies.persistent.PersistentPolicy`
identifies the class to construct, which is a different fact from `persistent` being a spelling the
factory resolves.

The existing headline guard is unchanged - the registry is still the set the discovery surfaces own.

## Tests

`tests/policies/test_provider_alias_discovery.py`, **1 failed / 17 passed** pre-fix -> 18 passed. The
one failure names every surface and both spellings:

```
AssertionError: create_policy resolves ['composite', 'persistent'] with no registry entry, and these
surfaces teach a caller to enumerate accepted spellings without naming them:
{'strands_robots/policies/factory.py::list_aliases': ['composite', 'persistent'],
 'docs/api-reference.md: | `list_aliases()` | Every provider alias and the canoni...': [...],
 "docs/policies/overview.md: `create_policy` also accepts each provider's declared al...": [...]}.
A caller validating against the documented union refuses a spelling the factory accepts.
```

A 1-failed split is the point: one assertion flips, and the 17 that pass on both trees are the
evidence nothing else moved. Every new guard is load-bearing - each break is detected by exactly one,
naming exactly its own site:

| regression | detected by | reports |
|---|---|---|
| the derivation stops probing auto-discovery | premise | `no module ... resolves outside the registry` |
| `overview.md` stops stating the recipe | premise | `overview.md no longer states the enumeration recipe` |
| the factory docstring drops `persistent` (dotted path left in place) | headline | `{...::list_aliases: ['persistent']}` |
| the api-reference row drops both spellings | headline | `{docs/api-reference.md...: ['composite', 'persistent']}` |

Controls, all passing on both trees: `composite` really does build through the factory and reports
`provider_name == "composite"`; `persistent` resolves but raises on construction (pinning *why* it is
documented differently); `base`/`factory` do not resolve, so the derived set is not "every module in
the package"; and a dotted path alone does not satisfy the token match.

## Verification

Measured at `c33b4b8e`.

- 255-file blast radius (everything naming the factory/registry surfaces or `policies.json`, every
  guard that reads a docs page or walks a source tree, all of `tests/policies` and `tests/registry`,
  plus the repo-wide docstring/string/changelog guards), two trees in parallel and in lockstep
  (397.71s vs 397.34s): branch **11 failed / 11199 passed / 81 skipped**, `upstream/main` **11 failed
  / 11191 passed / 81 skipped**. Failing-ID sets **identical**, `comm` empty both ways; the `+8`
  passed is exactly the 8 new items. The 11 shared failures are dependency-absent locally
  (`awsiotsdk`, and a `draccus` floor below lerobot's) - both declared extras CI installs.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `upstream/main` worktree,
  branch-only empty, none in the changed files.
- No non-ASCII added to any changed file; fence counts unchanged; no ragged table rows.

## Artifact

Same package code in both columns - only the documentation differs. The capture script reads the
recipe out of the tree it runs against, checks whether that tree's prose names any spelling beyond the
union, builds the allowlist a caller would build, then validates and dispatches. `random` (a registry
alias) is admitted, resolves and builds identically in both columns; on `upstream/main` `composite` is
refused by the caller's own allowlist while the factory builds it.

A caller following the documented recipe refuses composite on main and admits it after

### DIFF
--- a/changelog.d/2582-enumeration-claim-names-what-the-factory-resolves.md
+++ b/changelog.d/2582-enumeration-claim-names-what-the-factory-resolves.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **policies**: `list_providers()` and `list_aliases()` are documented as every *registered* provider spelling rather than every spelling `create_policy` accepts. `import_policy_class` also resolves a module under `strands_robots.policies` that exports a `Policy` subclass, so the union omitted `composite` - which `create_policy` builds - and `persistent`, which resolves but is constructed directly because its first parameter is named `provider`. Both are now named where enumeration is taught, and a guard derives them from the resolution path so a third such module is held to the same rule.
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -122,7 +122,7 @@ from strands_robots.policies.cosmos3 import Cosmos3Policy
 | `create_policy(provider, **kw)` | Resolve + construct. Accepts `zmq://`, `cosmos3://`, HF `org/model`. |
 | `register_policy(name, loader, aliases)` | Runtime registration. |
 | `list_providers()` | Sorted canonical names of every JSON-registered provider, plus any runtime `register_policy` names and their aliases. Canonical names only for the JSON registry: pair it with `list_aliases()` for the rest. |
-| `list_aliases()` | Every provider alias and the canonical name it resolves to, across both registries. Together with `list_providers()` they are every spelling `create_policy` accepts. |
+| `list_aliases()` | Every provider alias and the canonical name it resolves to, across both registries. Together with `list_providers()` they are every *registered* spelling - not every spelling `create_policy` resolves. A module under `strands_robots.policies` exporting a `Policy` subclass also resolves under its own module name: `composite` (builds through the factory) and `persistent` (resolves; constructed directly). |
 | `list_policy_types()` | `policy_type` strings the installed lerobot resolves; `[]` without lerobot. Discovery peer of `list_providers`. |
 | `UntrustedRemoteCodeError` | Raised when `STRANDS_TRUST_REMOTE_CODE` is required but unset. |
 | `Gr00tPolicy` | GR00T N1.5/N1.6/N1.7 via ZMQ (service) or in-process. |
--- a/docs/policies/overview.md
+++ b/docs/policies/overview.md
@@ -14,7 +14,9 @@ python -c 'from strands_robots.policies import list_providers; print(list_provid
 
 `create_policy` also accepts each provider's declared aliases and shorthands,
 which `list_providers()` does not repeat. `list_aliases()` reports those, so
-the two together are every spelling `create_policy` accepts:
+the two together are every *registered* spelling. They are not every spelling
+`create_policy` resolves: `composite` and `persistent` resolve with no registry
+entry, as described under [Beyond the registry](#beyond-the-registry) below.
 
 ```bash
 python -c 'from strands_robots.policies import list_aliases; print(list_aliases())'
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Documentation-and-guard correction: three surfaces (`list_aliases`' docstring, the `docs/api-reference.md` row, and `docs/policies/overview.md`) claimed that `set(list_providers()) | set(list_aliases())` is every spelling `create_policy` accepts, when `import_policy_class`'s auto-discovery fallback also resolves `composite` (buildable through the factory) and `persistent` (resolves, but unbuildable there since its first parameter collides with the already-bound `provider`). The prose now says *registered* and names both auto-discovered spellings, and a new guard in `tests/policies/test_provider_alias_discovery.py` derives the unregistered spellings from the resolution path itself so a third such module is graded on arrival. Verified against the head SHA: the `strands_robots/policies/factory.py` change is docstring-only (AST-equivalent after stripping docstrings), the changed test file reports 18 passed, no non-ASCII in any added line, ruff check/format clean on the changed Python files, and both markdown anchors (`wbc.md#composing-an-upper-body-manipulation-on-top-of-wbc`, `persistent-worker.md`) resolve. Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention. No behaviour change, no public API/wire/schema surface touched.

## What's good

- The guard derives the unregistered spellings by probing `import_policy_class` rather than hard-coding `['composite', 'persistent']`, with premise tests that keep it from going vacuous and controls pinning the token-boundary match and the `base`/`factory` negative cases.
- Keeps the discovery surfaces reporting the registry only, consistent with `tests/test_docs_policy_coverage.py`'s pin and the `mesh/security.py` precedent, instead of widening `list_providers()` to include a spelling the factory cannot build.
- Changelog fragment instead of a direct `CHANGELOG.md` edit (AGENTS.md PR workflow step 3).

PR #1374, opened 2026-07-14
docs(readme): correct simulation action count 65 -> 67

sundargthb:
## What

The README describes the MuJoCo simulation AgentTool as a "65-action" tool in five places, but the tool_spec enum now has **67 actions**. Two actions were added since the count was last updated:
- `list_cameras` (via #1352, the unknown-camera error now hints it)
- `register_builtin_benchmarks`

This updates the five references to 67.

## Verification

```
strands_robots/simulation/mujoco/tool_spec.json -> action enum length = 67
```

Lines updated: features list, `Robot()` factory note, hardware-mode note, Simulation section intro, and the repo-tree comment.

## Not touched

The `STRANDS_MESH_CA_PINS` row ("comma-separated **64-char hex**") is unrelated to the action count and is intentionally left as-is.

Docs only, no code changes.

PR #2583, opened 2026-08-21
docs(architecture): derive the orientation page's counts from the code

cagataycali:
## Problem

`docs/architecture.md` is the map a contributor reads to size each subsystem, and its "ABCs"
paragraph is read as the contract a new implementation conforms to. Both restate the code in
prose. One row of that table is graded - `test_docs_robot_catalog_coverage.py` ties the registry
row to `robots.json`, and it is exact. Nothing graded the rest, and the rest drifted.

Measured against the code on `upstream/main`:

| claim | page states | code says |
|---|---|---|
| policies row: providers | `4` | **14** (`registry/policies.json`) |
| tools row: `@tool` helpers | `8` | **20** (AST walk of `strands_robots/tools/`) |
| README: Simulation actions (x2 sites) | `67` | **77** (published `tool_spec.json` enum) |
| `Policy` para: declaration seams named | 1 of 3 | 3 of 3 |
| `Policy` para: implementations named | 4 of 15, by name | 15 |
| registry row: robots / aliases / categories | 72 / 119 / 8 | 72 / 119 / 8 - **already graded, already exact** |

Every module-table count and the ABCs paragraph date from the original docs site (#160,
2026-06-11) and never moved while the package grew. README's action count is the exception that
proves the point: #1374 corrected it `65 -> 67` by hand on 2026-07-14, which is how a number
becomes authoritative, and it drifted again by the same mechanism - because nothing ties it to
the enum.

## The seam half is the one with teeth

The paragraph enumerated the `Policy` contract as five members, naming `requires_images` and
omitting `required_bodies` and `children`. Those are not incidental: `Policy.children`'s own
docstring calls all three "the same `policy declares, runtime supplies` contract", so the code
marks a three-member family and the page named one of it.

Driven in MuJoCo on a `unitree_g1`:

- A policy implementing **exactly the five members the paragraph named** receives `base_quat`
  and **no body pose at all** - `status="success"`, nothing reports it. The runtime supplies a
  body pose only for a body `required_bodies` names.
- Declaring the seam adds `body.torso_link.{pos,quat,lin_vel,ang_vel}`.
- `base_quat` is the *pelvis*. At `waist_yaw = 0.6 rad` it diverges from `torso_link` by
  **34.38 deg**, so a whole-body tracker written against the documented contract feeds its
  network the wrong frame whenever the waist is not neutral.

And the enumeration was self-consistently incomplete: **0 of the 4 implementations it named
override either seam**, while **all 3 that do** (`CompositePolicy`, `PersistentPolicy`,
`ProtoMotionsPolicy`) were absent. It named the subset for which the omission is invisible.

## Change

`docs/architecture.md` and `README.md`, plus a guard that derives each number from the code so a
count cannot be right the day it is written and wrong a month later:

- module-table counts from `registry/policies.json` and an AST walk of `strands_robots/tools/`;
- the Simulation action count from the published `tool_spec.json` enum, graded wherever either
  document states one rather than at fixed lines, so a fresh claim elsewhere in the file is held
  to the same enum;
- the `Policy` paragraph against `Policy.__abstractmethods__` (what an implementation must
  supply) and against the declaration-seam family, itself derived from the base class's own
  docstrings - so a fourth seam is graded the day it is added;
- the paragraph now states the implementation count and points at the provider table instead of
  closing a list of four.

The provider count reads `policies.json` rather than `list_providers()` on purpose: the row
describes what the package *contains*, and `list_providers()` also reports runtime registrations,
which in a test session includes every throwaway `register_policy` a sibling module left behind
(23 mid-suite here).

## Verification

Measured at `641d113b`, `upstream/main` at `b96412ad`, mujoco 3.12.0.

**Pre-fix split** - the new guard on `upstream/main`'s pages: **6 failed / 6 passed**, each
failure naming the claim and the code count, e.g.

```
docs/architecture.md's Policy paragraph does not name ['children', 'required_bodies'];
the base class marks ['children', 'required_bodies', 'requires_images'] as the
'policy declares, runtime supplies' family, and a policy that skips one is not told it exists
```

After: **12 passed**. The 6 that pass on both trees are the controls - the three abstract members
were already named; `docs/architecture.md`'s own `60+ actions` was not a false claim, so the
action grader correctly did not flag it (it is now the exact number for consistency); the seam
family derivation is pinned non-vacuously; and two planted-prose cases prove the rules fire on an
omission rather than only on the real page.

**Every guard is load-bearing** - reverting one claim at a time fails exactly one test, and only
its own:

| break | tests failed |
|---|---|
| revert provider count to 4 | `..._number_of_providers_that_ship` |
| revert tool count to 8 | `..._number_of_tools_that_ship` |
| revert README action count to 67 | `..._is_the_published_enum[README.md]` |
| drop the `children` seam | `..._names_every_declaration_seam` |
| drop an abstract member | `..._names_every_abstract_member` |
| re-enumerate four implementations | `..._does_not_enumerate_a_strict_subset...` |

**Zero regressions** - 77 test files (every guard that reads `docs/architecture.md`, README or a
docs page, walks a source tree, grades docstrings/exports/changelog fragments, plus
`tests/policies/`), identical selection on both trees, run in parallel:

|---|---|---|---|
| this branch | 0 | 1908 | 1 |
| `upstream/main` | 0 | 1908 | 1 |

Failing-ID sets identical (empty both ways). `ruff check` + `ruff format --check` clean over
`strands_robots tests tests_integ`; `mypy` 24 == 24 against a pristine `upstream/main` worktree
with none in the changed files.

## Artifact

Both columns run the same package - only `docs/architecture.md` and `README.md` differ, and each
row is read from that column's own page. The code-side measurement and the sim divergence are
asserted identical across trees inside the figure script, so the picture cannot drift from the
claim: **5 of 6 claims disagree with the code before, 0 after.** The strip below is the same
`unitree_g1` at `waist_yaw` 0.0 and 0.6 rad (13.3% of pixels differ), with the observation a
documented-contract-only policy actually receives.

architecture counts graded against the code

## Scope

Left alone, with the measurement:

- The `SimEngine` line's "30+ abstract actions" - 14 `__abstractmethods__` against 38 public
  members, so which reading it takes is genuinely ambiguous and "30+" is true for the surface one.
- README's "70+ robots" and the `docs/robots/` and `docs/policies/` counts, all measured correct.
- The registry row, which `test_docs_robot_catalog_coverage.py` owns. Two guards on one claim can
  disagree about the wording.

### DIFF
--- a/README.md
+++ b/README.md
@@ -540,7 +540,7 @@ AgentTool returning `{"status", "content"}`.
 | `start` | `instruction`, `policy_port`, `duration` | Non-blocking async start |
 | `status` | - | Current task status |
 | `stop` | - | Interrupt running task (emergency stop) |
-In sim mode the same tool exposes the 67 Simulation actions - see Simulation (MuJoCo).
+In sim mode the same tool exposes the 77 Simulation actions - see Simulation (MuJoCo).
 </details>
 
 <details>
@@ -924,7 +924,7 @@ the library.
 ## Simulation (MuJoCo)
 
 `Robot("so100")` (sim mode) returns a `Simulation` - a MuJoCo-backed AgentTool
-exposing **67 actions** for world composition, physics, rendering, policy
+exposing **77 actions** for world composition, physics, rendering, policy
 execution, and dataset recording. Build it directly when you want full control:
 
 ```python
--- a/changelog.d/2583-architecture-counts-derived-from-the-code.md
+++ b/changelog.d/2583-architecture-counts-derived-from-the-code.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **docs**: `docs/architecture.md`'s module table and `Policy` contract paragraph, plus README's Simulation action count, are now derived from the code by a guard rather than restated by hand. The page had claimed 4 policy providers for 14, 8 `@tool` helpers for 20, and four `Policy` implementations of fifteen; README quoted 67 published actions for 77. The `Policy` paragraph also named only one of the three "policy declares, runtime supplies" seams, so a policy conforming to the documented contract received no body pose at all - `base_quat` is the pelvis, which diverges from `torso_link` by 34.4 degrees at 0.6 rad of waist yaw.
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -79,17 +79,17 @@ graph TB
 |--------|--------------|-----------|
 | `strands_robots/robot.py` | Factory `Robot(name, mode, backend, **kwargs)`. Name resolution, sim/real dispatch, mesh attach. | `Robot()` function |
 | `strands_robots/registry/` | 72 robots, 119 aliases, 8 categories. `robots.json` is source of truth. | `list_robots()`, `resolve_name()`, `get_robot()` |
-| `strands_robots/simulation/` | MuJoCo `AgentTool` - 60+ actions. | `Simulation`, `SimWorld`, `SimRobot`, `SimObject`, `SimCamera` |
+| `strands_robots/simulation/` | MuJoCo `AgentTool` - 77 actions. | `Simulation`, `SimWorld`, `SimRobot`, `SimObject`, `SimCamera` |
 | `strands_robots/simulation/base.py` | Backend ABC for future Isaac/Newton backends. | `SimEngine` |
 | `strands_robots/hardware_robot.py` | Real-servo path. Async task execution + status. | `Robot` (class), `TaskStatus`, `RobotTaskState` |
-| `strands_robots/policies/` | ABC + 4 providers + factory + JSON registry. | `Policy`, `create_policy()` |
+| `strands_robots/policies/` | ABC + 14 providers + factory + JSON registry. | `Policy`, `create_policy()` |
 | `strands_robots/dataset_recorder.py` | LeRobot v3 writer. | `DatasetRecorder` |
-| `strands_robots/tools/` | 8 `@tool`-decorated helpers. | `lerobot_calibrate`, `serial_tool`, etc. |
+| `strands_robots/tools/` | 20 `@tool`-decorated helpers. | `lerobot_calibrate`, `serial_tool`, etc. |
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Docs-and-guard PR: corrects five drifted counts across `docs/architecture.md` and `README.md` (4 -> 14 policy providers, 8 -> 20 `@tool` helpers, 67 -> 77 published Simulation actions at both README sites) and rewrites the `Policy` ABC paragraph to name all three abstract members plus all three "policy declares, runtime supplies" seams (`requires_images`, `required_bodies`, `children`), replacing the stale closed list of four implementations with the count and a pointer to the provider table. A new test module derives every stated number from the code itself - `registry/policies.json`, an AST walk of `strands_robots/tools/`, the published `tool_spec.json` enum, `Policy.__abstractmethods__`, and the seam family read from the base class's own docstrings - so the page cannot drift again silently. Verified at head: all three counts match the code exactly, the 12 new tests pass, the seam-family derivation is pinned non-vacuously against phrasing drift, and two planted-prose cases prove the graders fire on omissions rather than only on the current page.

## What's good

- Every guard is falsifiable and was demonstrated load-bearing (per-claim revert -> exactly one failing test), and the non-vacuity pin on the derived seam family closes the usual "grader silently grades an empty set" failure mode.
- Deriving the provider count from `registry/policies.json` rather than `list_providers()` is the right call and the docstring records why - runtime `register_policy` leakage from sibling test modules would make the count session-dependent.
- `_policy_implementations()` walks the AST instead of importing, so a missing optional dependency cannot shrink the set and let a stale number pass.
- Scope discipline: the already-graded registry row and the genuinely ambiguous `SimEngine` "30+" line are left alone, with the measurement recorded in the description.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; no non-ASCII, no host paths, and the `Policy` paragraph's `policies/overview.md` link target exists.

PR #2584, opened 2026-08-21
fix(simulation): report the shared policy fields at every episode count

cagataycali:
## Problem

`run_policy`'s docstring names the fields an agent reads out of the `json` block and tells it which to gate on:

> **IMPORTANT** - `status` is not the rollout verdict. [...] A rollout that drove only a SUBSET of the robot's actuators is deliberately `success` [...] **Gate on `partial_action_failure_rate` and the binding-degradation flags below to decide whether a rollout is worth anything.**

Two of those fields are documented for a **loop** specifically, so the multi-episode payload is the one that has to carry them:

- `policy_load_cache_hit` - "`False` on episode 2+ of a loop is a smell that the caller rebuilt the policy instead of reusing `policy_object=`"
- `policy_resident_rss_mb` - "across a loop, that it stays resident instead of oscillating as it would on a per-episode reload"

The `n_episodes > 1` aggregate is built separately from the single-rollout payload, and every field it did not restate was absent from it. `_episodes_result`'s own docstring claimed the payload "keeps ONE shape across episode counts" and justified that by naming the two fields it did carry (`stopped_reason` / `steps_used`); measured, the shape went from 30 json fields to 10.

Because a caller reads a health flag with `.get(field, <healthy default>)`, an absent field is not an error - it is a clean bill of health. Same policy, same robot, same 12-step horizon, only `n_episodes` differs:

| call | json fields | `positional_fallback_used` | `generic_state_keys_used` | `policy_load_cache_hit` | documented gate |
|---|---|---|---|---|---|
| `n_episodes=1` | 30 | `True` | `True` | `True` | **degraded** |
| `n_episodes=3` | 10 | `<absent>` -> `False` | `<absent>` -> `False` | `<absent>` -> `False` | **healthy** |

Both `status="success"`. The policy drove 1 of the robot's 6 actuators on positionally-routed inputs, and the multi-episode shape is the one that collects a dataset - so those episodes recorded a robot moving on meaningless inputs while the documented gate said the run was fine. The data was never lost: each record in `episodes` reported `partial_action_failure_rate = 0.8333` the whole time. It was just unreachable from the top level, and no docstring said so.

## Change

`_episodes_result` now carries every field whose aggregate value the call already knows:

- **Identity** - `robot_name`, `policy`, `instruction`. These are the call's own inputs, constant across the episodes rather than something to aggregate.
- **Policy binding** - `positional_fallback_used`, `generic_state_keys_used`, `missing_state_keys_used`.
- **Policy load** - `policy_load_time_s`, `policy_load_cache_hit`, `policy_resident_rss_mb`.

The six policy fields are read off the ONE policy object every episode ran with, after the last episode. That is not a new aggregation: `eval_policy` and `evaluate_benchmark` already report exactly these six that way for their own multi-episode aggregates, so the semantics were already settled in-tree. The three identity values need no semantics at all. `_episodes_result` takes them as required keyword arguments, so all four call sites supply them and an error-path return carries them too.

**Deliberately not aggregated:** per-episode action health (`action_errors`, `action_resolution_rate`, `partial_action_failure_rate`) and the per-episode horizon/video fields. Summarising a per-step rate across episodes is a choice - mean, max, or step-weighted - with no in-tree precedent, and each record in `episodes` already reports its own. Both docstring and docs now point there and show the read:

```python
worst = max(e["partial_action_failure_rate"] for e in report["episodes"])
```

`test_the_aggregate_reports_no_summarised_action_rate` pins that boundary, so a future summary has to be decided rather than added.

## Docs

- `_episodes_result`'s docstring replaces the "ONE shape across episode counts" claim with the three groups it carries, why each is knowable, and where the per-episode health lives.
- `run_policy`'s "Fields in the json block" gains an "Across episodes" paragraph, so the field list is accurate at both shapes.
- `docs/simulation/overview.md`'s `run_policy` report section gains the same, next to the paragraph that enumerates the single-episode fields.

## Verification

Measured at `8878e15`, base `9e4e067`.

**Regression split** - the new test file copied onto `upstream/main`: **4 failed / 6 passed**, 10 passed after. The headline failure is the finding:

```
the same policy on the same robot reads 'degraded' at n_episodes=1 and
'healthy' at n_episodes=3; the gate run_policy documents flips with the
episode count. Missing from the aggregate: [...24 fields...]
```

The 6 both-tree passes are the controls: the AST scan's non-vacuity premise, per-episode records still carrying the action health, no summarised rate on the aggregate, and the single-episode shape keeping its own action health and carrying no aggregate-only field.

**The root-cause guard is self-maintaining.** `_policy_derived_payload_fields()` derives the graded set from the runner's own payload build - keys whose value expression reads `policy` or the process RSS - rather than listing it, so a seventh such field added to the single-rollout payload is graded on arrival.

**Each guard is load-bearing.** One break per guard, each failing exactly and only its own:

| break | fails |
|---|---|
| drop the 6 policy fields from the aggregate | gate parity, binding-flag values, policy-derived carry |
| drop the 3 identity fields | identity constancy |
| cripple the AST derivation | the non-vacuity premise (`{'policy_resident_rss_mb'}` left) |
| invent a summarised `partial_action_failure_rate` | the no-summary boundary |

**Gate** - 151 test files covering the aggregator, every field it now carries, the multi-episode surface and the repo-wide docs/docstring/string/changelog guards, run on both trees with the identical selection (the new file excluded, so any difference is the source change alone): **7 failed / 4850 passed / 82 skipped on both**, failing-ID sets identical and `comm` empty both ways. Those 7 are import-order artifacts of a wide `tests/simulation` selection rather than anything in this diff - the two files they live in pass 137/137 when run on their own against an unmodified base. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 with an empty branch-only set.

## Rollout artifact

The rollout is **pixel-identical in both trees** (mean |px| 3e-06) - what changes is what the report says about it. The single-episode payload is unchanged apart from `policy_resident_rss_mb`, which is a live process-RSS reading and legitimately differs between two processes.

run_policy health gate across episode counts


yinsong1986 (approved):
## Summary

Closes a real reporting gap: `run_policy`'s `n_episodes > 1` aggregate was built separately from the single-rollout payload and omitted the policy-binding flags and policy-load telemetry, so the health gate the docstring tells callers to read (`.get(field, <healthy default>)`) flipped from degraded to healthy purely on the episode count — on exactly the shape used to collect datasets. The fix carries the six policy-derived fields plus the three identity fields into the aggregate, with value expressions byte-identical to the settled single-rollout payload in `policy_runner.py` (same `getattr` defaults, same rounding, same `process_rss_mb()` read), and deliberately does not summarise per-episode action rates — the `episodes` records keep those, and both docstring and docs now say so.

## What's good

- Independently verified the test's AST derivation against `policy_runner.py`: it yields exactly the six claimed fields (`positional_fallback_used`, `generic_state_keys_used`, `missing_state_keys_used`, `policy_load_time_s`, `policy_load_cache_hit`, `policy_resident_rss_mb`), all now carried by the aggregate — so a seventh policy-derived field is graded on arrival rather than requiring a test edit.
- `_episodes_result` is private with all four call sites in `base.py` updated (including the three error paths), no subclass overrides, and no external callers — the required-keyword signature change is contained.
- The additive payload fields reuse the semantics `eval_policy`/`evaluate_benchmark` already established for multi-episode aggregates, so no new aggregation choice was invented; the no-summary boundary on `partial_action_failure_rate` is pinned by its own test.
- Changelog fragment in the current format, docs updated in the same PR, `ruff check`/`ruff format --check` clean on both changed Python files, no non-ASCII or host paths in the diff, and the new test file follows the in-tree `importorskip` → `MUJOCO_GL` ordering convention.

## Verification suggestions

The GL-dependent tests in the new file need a real EGL/OSMesa context (they fail on a headless box without one, as does the pre-existing `test_run_policy_multi_episode.py`); the non-rendering AST-scan test passes standalone. On a GL-capable machine: `pytest tests/simulation/test_multi_episode_payload_shape_parity.py tests/simulation/test_run_policy_multi_episode.py -q` to confirm both the new parity pins and the existing multi-episode surface together.

Issue #2017, opened 2026-08-07
process: nothing reports two open PRs claiming the same issue - 3 of the last 100 PRs were duplicate pairs, and all three burned an approval


cagataycali:
## The gap

`.github/workflows/closing-reference.yml` reads one pull request in isolation: it compares the issue numbers a *title* claims against that PR's own `closingIssuesReferences`. It has no cross-PR view, so a pull request that claims an issue **another open pull request already claims** passes every check in the repository. Nothing on either side reports it until a human notices, and the notice has consistently arrived after review.

## Measured

Over the last 100 pull requests (#1867 through #2016), three pairs claimed one issue each -- six PRs, three of them wasted:

| issue | pair | opened apart | abandoned | had reached |
|---|---|---|---|---|
| #1942 | #1944, #1946 | 8m 32s | #1944 | APPROVED |
| #1994 | #1995, #1996 | 32m 0s | #1995 | APPROVED |
| #2007 | #2015, #2016 | 4m 9s | #2016 | APPROVED |

Three properties of that table are the actual finding:

1. **Every abandoned PR was already approved when it was closed.** The cost is not the wasted authoring, it is a review approval spent on a change that could never ship -- and review is the scarcest resource here (#1905 measures the same scarcity from the other direction).
2. **Two of the three pairs could not both land regardless.** The closing comments on #1944 and #1995 each record a real `git merge-tree` content conflict (`strands_robots/simulation/mujoco/motion_primitives.py` and `strands_robots/policies/wbc/policy.py`). So in those cases the duplication was not merely redundant, it was a merge failure waiting on whichever PR lost the race.
3. **Every pair was opened inside one ~35-minute window.** The collision is an intake failure -- the second PR is opened before anything has observed the first -- not a drift that accumulates over days.

Each resolution also had to be adjudicated by hand. The closing comments on #1995 and #1944 are among the longest in the repository, because deciding which of two approved, conflicting implementations survives requires re-measuring both.

### A second, quieter symptom

Duplicate pairs also duplicate changelog fragments. #1995 added `changelog.d/1994-wbc-goal-override-domain.md` while #1996 added `changelog.d/1996-wbc-command-override-domains.md`, both for one fix; `scripts/assemble_changelog.py --check` validates names and headings and cannot know that two fragments describe the same change, so landing both would assemble two `###` entries for one defect. The #2007 pair split the other way -- #2015 carried a fragment and #2016 argued none was needed -- which is the same ambiguity from the opposite side.

[...]

PR #2586, opened 2026-08-21
fix(scripts): name an uncomputed mergeability instead of reading it as clean

### LINKED ISSUE #2585: check_merge_blockers.py reads an uncomputed mergeability as "no conflict", so a DIRTY pull request is reported as owed by a reviewer
`scripts/check_merge_blockers.py` models mergeability as `bool | None` and classifies only two of those three states. A `None` -- which is what GitHub returns while it recomputes -- falls through the conflict branch and is reported as if the branch merges cleanly. The consequence is the exact misfiling #1905 exists to prevent, with the sign flipped: a `DIRTY` pull request is reported as **owed by a reviewer**.

## Measured today

Two pull requests merged at 23:06 UTC (#2577, #2583), so `main` moved twice. The `--all-open` sweep ran about a minute later and split two equally-conflicted branches into two different outcomes:

| pull request | sweep said | measured `mergeable` / `mergeStateStatus` |
|---|---|---|
| #1722 | `merge-conflict, missing-approval` -- owed by **the author** | `CONFLICTING` / `DIRTY` |
| #1035 | `pusher-only-approval` -- owed by **a reviewer other than the pusher** | `CONFLICTING` / `DIRTY` |

Both are conflicted. Only one was reported as conflicted.

The differentiator is not the branches, it is REST's on-demand computation. At sweep time `GET /pulls/1035` had not finished recomputing against the new base and returned `mergeable: null`; #1722's was already cached as `false`. Read again minutes later, both agree:

```
GET /repos/strands-labs/robots/pulls/1035  ->  mergeable=False  mergeable_state='dirty'
GET /repos/strands-labs/robots/pulls/1722  ->  mergeable=False  mergeable_state='dirty'
```

Same script, same run, same true state, two different owners -- decided by which value happened to be warm in a cache.

## Mechanism

`PullRequestState.mergeable` is declared `bool | None`, and `evaluate` tests one of the three:

```python
if state.mergeable is False:
    found.append(Blocker(MERGE_CONFLICT, ...))
```

With `None` no blocker is appended, so for #1035 the conflict simply is not in the report and the classification reduces to the approval rule underneath it. The third state is modelled and then not handled.

## Why this is not "just guess it is a conflict"

It is not, and `tests/test_merge_blockers.py::test_an_unknown_mergeability_is_not_reported_as_a_conflict` is right to pin that -- *"Guessing turns a wait into an accusation."* Reporting `None` as `merge-conflict` would be wrong.

The gap is that it is not reported **at all**. Silence here is not neutral: it is indistinguishable from `mergeable: true`, and the reader is told a reviewer owes the next action.

The script already has the vocabulary for this, one field over. A required check whose conclusion is `None` gets its own outcome -- `required-check-pending`, owed by `NOBODY`, documented as *"Nobody. The answer is not in yet."* An unknown mergeability is the same shape and gets nothing.

## Why the window is systematic, not rare

Every merge into `main` invalidates the cached mergeability of every open pull request, and a repository-health sweep is most often run **right after a merge**. So the degraded read is not an unlucky coincidence; it is correlated with the moment the sweep is most likely to run. All 9 open pull requests read `mergeable: UNKNOWN` in the sibling head sweep in the same window, which is the same cache in the same state.

## Cost

The report routes a reviewer to approve #1035, and #1905 already records what that approval buys: nothing today, because the branch needs an author-side rebase first, and the approval is then **dismissed by the resolving push**. That is a burned approval -- the cost #2017 tracks -- spent because a null read as clean.

## Proposed remedy

Name the third state instead of collapsing it:

- a `merge-state-unknown` outcome, owed by **nobody** (the answer is not in yet), carrying the remedy: re-read, the value is computed on demand and settles on a subsequent read;
- **gating**, for the same reason a conflict is gating -- until it resolves the rules behind it cannot be assessed, so the approval rule is reported as necessary but not sufficient rather than as the next action;
- **not** a finding, so the exit status keeps its current meaning (matching `required-check-pending`, which is also not a finding).

Reporting rather than polling is deliberate and matches the script's existing idiom: it prints `Attempt the merge` for the stale-state case rather than attempting it, and `Neither script gates a merge`. A bounded re-poll would put a sleep inside a sweep over N pull requests to answer a question the caller can answer by reading again.

cagataycali:
`scripts/check_merge_blockers.py` models `mergeable` as `bool | None` and classifies only two of those three states. GitHub computes mergeability on demand and returns null while it works, so the third state fell through the conflict branch and was reported as a clean merge -- letting the rules underneath it own the next action.

Closes #2585

## How this was found

Not by reading the code. By running the sweep AGENTS.md mandates for repository health, about a minute after merging #2577 and #2583 -- which is what made `main` move and invalidated the cached value for every open pull request.

The sweep split two equally-conflicted branches into two different owners:

| pull request | sweep said | true state |
|---|---|---|
| #1722 | `merge-conflict, missing-approval` -- owed by **the author** | `CONFLICTING` / `DIRTY` |
| #1035 | `pusher-only-approval` -- owed by **a reviewer other than the pusher** | `CONFLICTING` / `DIRTY` |

The only difference was which value happened to be warm in REST's cache. Read again minutes later, both agree, and the patched sweep on the same live data now reports #1035 as `merge-conflict, pusher-only-approval` owed by the author.

## Two manifestations, not one

1. **A conflicted branch is handed to a reviewer.** #1035 above. #1905 records what that approval buys: nothing today, because the branch needs an author-side rebase first, and the approval is then dismissed by the resolving push -- a burned approval, the cost #2017 tracks.
2. **An otherwise satisfied branch is told to merge.** With every other rule satisfied and mergeability uncomputed, the outcome was `no-unsatisfied-rule`, owed by `anyone, by attempting the merge`, and the report printed the `Attempt the merge` remedy. So the tool advised attempting a merge on a branch whose conflict status was still being computed. Measured directly against the pre-fix module:

```
PRE-FIX outcome: ['no-unsatisfied-rule']
PRE-FIX owed by: anyone, by attempting the merge
```

## This is not "guess it is a conflict"

`test_an_unknown_mergeability_is_not_reported_as_a_conflict` is right, and it still passes unchanged -- *"Guessing turns a wait into an accusation."* A null is never reported as a conflict here.

The gap was that it was not reported **at all**, and silence is not neutral: it reads identically to `mergeable: true`.

The script already had the vocabulary one field over. A required check whose conclusion is `None` gets its own outcome -- `required-check-pending`, owed by `NOBODY`, *"the answer is not in yet."* An unknown mergeability is the same shape and had none.

## The change

- `merge-state-unknown`, owed by **nobody**, detail carrying the re-read that resolves it.
- **Gating**, so the approval rule underneath reads as *necessary but not sufficient* rather than as the next action. The comment says why it gates on a weaker basis than a conflict does: the cost is asymmetric -- a wrong reviewer hand-off burns an approval, a re-read costs one round trip.
- **Not a finding**, so the exit status keeps its current meaning, matching `required-check-pending`.

## Why report and not poll

Deliberate, and it matches the script's existing idiom rather than introducing a new one: it prints `Attempt the merge` for the stale-state case instead of attempting it, and `Neither script gates a merge`. A bounded re-poll would put a sleep inside a sweep over N pull requests to answer a question the caller resolves by reading again.

## Verification

- The 4 new tests **fail on pre-fix code** and pass after: `4 failed, 59 passed` before, `63 passed` after. One fails behaviourally rather than on the missing symbol, which is the `Attempt the merge` case above.
- 334 passed across the merge-blocker, sibling-check, changelog and string-hygiene suites.
- `ruff check` clean, `ruff format --check` clean, `mypy scripts/check_merge_blockers.py` clean, `assemble_changelog.py --check` OK.
- End-to-end `--all-open` against live data: 10 pull requests, normal path unchanged.

## Doc drift

AGENTS.md described the outcome list and said *"A conflict or a draft is reported as gating"*, which this change makes incomplete. Updated in the same diff, with the reason the null is correlated with the moment a health sweep runs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1011,10 +1011,21 @@ hatch run format            # ruff check --fix, ruff format
    unresolved thread or a failing check (the author), a missing approval (any
    reviewer), an approval only its own pusher supplied (a different reviewer,
    per #1905), a required check absent because a fork run is held at
-   `action_required` (a maintainer), a check still running (nobody), or no
-   unsatisfied rule at all, which is the #2574 case and the one worth saying out
-   loud. A conflict or a draft is reported as *gating*: the rules behind it
-   cannot be assessed, so an approval there is necessary but not sufficient.
+   `action_required` (a maintainer), a check still running (nobody), a
+   mergeability GitHub has not finished computing (nobody, until a re-read), or
+   no unsatisfied rule at all, which is the #2574 case and the one worth saying
+   out loud. A conflict, a draft, or an uncomputed mergeability is reported as
+   *gating*: the rules behind it cannot be assessed, so an approval there is
+   necessary but not sufficient.
+
+   That last one is why the sweep is worth re-reading rather than trusting once.
+   `mergeable` is `bool | None`, and a merge into `main` invalidates the cached
+   value for **every** open pull request -- so a sweep run just after a merge is
+   exactly when the null appears, which is also when a health pass is most likely
+   to run. #1035 was measured reading `pusher-only-approval` (owed by a reviewer)
+   while it was in fact `CONFLICTING`/`DIRTY`, and an otherwise-satisfied pull
+   request in the same state read `no-unsatisfied-rule`, whose printed remedy is
+   to attempt the merge. Both are now `merge-state-unknown`. See #2585.
 
    It composes `check_last_push_approval.py` rather than restating it, so what
    counts as a current approval has one owner. Neither script gates a merge.
--- a/changelog.d/2585-merge-state-unknown-is-not-clean.md
+++ b/changelog.d/2585-merge-state-unknown-is-not-clean.md
@@ -1,0 +1,18 @@
+### Fixed: an uncomputed mergeability is no longer reported as a clean merge
+
+`scripts/check_merge_blockers.py` modelled `mergeable` as `bool | None` but
+classified only two of those three states. GitHub computes mergeability on
+demand and returns null while it works, and a merge into the base invalidates
+the cached value for every open pull request -- so a sweep run just after a
+merge is exactly when the null appears. Falling through the conflict branch
+reported it as if the branch merged cleanly, and the rules underneath it then
+owned the next action: #1035 was measured reporting `pusher-only-approval`,
+owed by a reviewer, while it was in fact `CONFLICTING`/`DIRTY`. An otherwise
+satisfied pull request in the same state reported `no-unsatisfied-rule`, whose
+printed remedy is to attempt the merge.
+
+The third state is now named: a `merge-state-unknown` outcome, owed by nobody,
+carrying the re-read that resolves it. It is reported as *gating*, so an
+approval underneath it reads as necessary but not sufficient rather than as the
+next action, and it is deliberately not a finding, so the exit status keeps its
+meaning. A null is still never reported as a conflict.
--- a/scripts/check_merge_blockers.py
+++ b/scripts/check_merge_blockers.py
@@ -78,6 +78,16 @@ is*, so the outcomes group that way rather than by severity:
 ``required-check-pending``
     Nobody. The answer is not in yet.
 
+``merge-state-unknown``
+    Nobody, for now, and the same shape as the entry above one field over.
+    ``mergeable`` is ``bool | None``: GitHub computes it on demand and returns
+    null while it works, which is neither "conflicts" nor "merges cleanly". A
+    merge into the base invalidates it for every open pull request, so a sweep
+    run just after a merge is precisely when it is null. Reported as *gating*,
+    because reading the null as clean is how a conflicted branch is reported as
+    owed by a reviewer -- measured on #1035, which read
+    ``pusher-only-approval`` while it was ``DIRTY``. Re-read to resolve it.
+
 ``required-check-absent``
     A **maintainer**, by authorising or re-running the workflow. A fork pull
     request whose runs are held at ``action_required`` reports ``completed``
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes the third-state gap in `scripts/check_merge_blockers.py`: `mergeable` is `bool | None`, and the `None` GitHub returns while recomputing mergeability previously fell through the conflict branch and was reported as a clean merge, letting the rules underneath own the next action (measured: #1035 read `pusher-only-approval` while `CONFLICTING`/`DIRTY`, and an otherwise-satisfied branch read `no-unsatisfied-rule` with the `Attempt the merge` remedy). The fix names the state as a new `merge-state-unknown` outcome owed by nobody, adds it to `_GATING` so downstream rules render as necessary-but-not-sufficient, and deliberately keeps it out of `_FINDINGS` so the exit-status contract is unchanged -- consistent with `required-check-pending`, the existing outcome of the same shape. The prior `test_an_unknown_mergeability_is_not_reported_as_a_conflict` pin still holds (a null is never called a conflict), so both halves of the invariant are now pinned.

## What's good

- Verified locally: the 4 new tests fail on pre-fix code (`4 failed, 59 passed`) and the full suite passes post-fix (`63 passed`); `ruff check` and `ruff format --check` clean on both changed Python files.
- The report-not-poll decision matches the script's existing idiom (`Neither script gates a merge`), and the asymmetric-cost rationale for gating on a weaker basis than a conflict is written down at the `_GATING` definition rather than left to be rediscovered.
- Changelog fragment follows `changelog.d/<number>-<slug>.md` (issue number is explicitly permitted by `changelog.d/README.md`), and the AGENTS.md outcome-list description is updated in the same diff so the doc does not drift from the code.
- Scope discipline: no transport, exit-status, or CLI-surface changes; the diff is one classification branch plus its documentation and pins.

PR #2587, opened 2026-08-22
docs(policies): document the fields a config dataclass declares, and grade it

cagataycali:
## Problem

An `Attributes:` block is where a caller learns which keys a config dict may carry.
`EmbodimentMap` is exported from `strands_robots.policies.lerobot_local.embodiment`
and is built straight from a caller-supplied dict, so that block is its API - and it
documented **five of its ten fields**.

The five it omitted are the unit-conversion group. `state_units` / `action_units`
decide whether the map converts at all, and `gripper_index` / `gripper_joint_range` /
`joint_mids` parameterise the conversion. Their defaults are `"native"` / `-1` /
empty, so a caller writing an embodiment from that block gets a map that converts
nothing - and `validate()` accepts it:

```
spec built from the documented fields -> state_units='native' action_units='native' gripper_index=-1
validate() -> ACCEPTED (no refusal)
```

The consequence is the one `LerobotLocalPolicy` already names in a comment at the
conversion site: *"feeding the raw degree values straight in saturates the radian
joint limits"*. Driving a MuJoCo SO-101 with a degrees-trained SO-arm checkpoint's
action stream through each spec:

| | 5 documented fields | 10 documented fields |
| --- | --- | --- |
| largest value commanded | **70.0000 rad** | 1.2217 rad |
| joint 1 range | +/-1.9199 rad | +/-1.9199 rad |
| final pose vs the one requested | **-0.2782 vs +1.2217** (off by 1.4999) | +1.0222 vs +1.2217 (off by 0.1995) |
| mean tracking error over the sweep | 1.1171 rad | 0.4115 rad |

The arm does not stall - it is driven to its limit and unwinds, ending 1.50 rad from
the requested pose and on the wrong side of zero.

Two more surfaces had the same shape:

- `PackStateProcessorStep`, the registered pipeline step that **receives** four of
  those five fields (`processor.py` passes `state_units=`, `gripper_index=`,
  `gripper_joint_range=`, `joint_mids=` into it), documented three of its seven.
  The same field group was undocumented at both ends of the handoff.
- `ProtoMotionsConfig` omitted `onnx_in_names`, `onnx_out_names` and `metadata` -
  while `ProtoMotionsPolicy._run_session`'s own docstring cites
  `ProtoMotionsConfig.onnx_in_names` as where its shapes are documented.

In every case the semantics were already written down as inline comments beside the
field declarations, just not where a caller reads.

## Change

Each of the three blocks now names every field its class declares, including what the
default means and - for the unit group - what happens when it is left at that default.
No code changes: the diff is docstrings plus a new guard.

`tests/test_args_docstring_completeness.py` pins exactly this comparison for `Args:`
blocks, and its scope is a function signature, so an attribute list was graded by
nothing. `tests/test_attributes_docstring_completeness.py` closes that: for every
dataclass whose docstring **already** has an `Attributes:` block, every field it
declares must have an entry. A class with no such block is out of scope, which is the
same boundary the `Args:` guard draws.

Two conventions the shipped tree already uses are honoured, so the rule needs no
exemptions:

- a field documented on a **base class** counts, so a subclass may document only what
  it adds (`RLTrainSpec` relies on `TrainSpec` for `learning_rate`);
- one entry may name **several fields** that share a description
  (`TrainSpec`'s `lora_r / lora_alpha / lora_target_modules:`,
  `RLTrainSpec`'s `actor_obs_keys / critic_obs_keys:`), matching the `Args:` guard's
  own label handling.

The scan is AST-only, so a config dataclass behind an optional backend is still graded
with that backend absent.

## Scope

21 dataclasses carry an `Attributes:` block and declare 192 fields between them.
Three were incomplete; the other 18 were already exact, and the guard reports them
clean. The converse direction - an entry naming something that is not a field - was
measured across all 189 pre-existing entries and had no offenders, so it is not part
of this change.

## Verification

Measured at `284677f3` against `upstream/main` at `578e3674`.

Pre-fix, the new guard against unmodified sources: **3 failed / 34 passed**, one
failure per class naming its omitted fields, e.g.

```
EmbodimentMap has an Attributes: block that omits
['state_units', 'action_units', 'gripper_index', 'gripper_joint_range', 'joint_mids']
```

After: 37 passed. Every control is load-bearing - each of six planted regressions
fails exactly and only its own guard:

| planted regression | headline | combined labels | base class | dataclass only | ClassVar | non-vacuity |
| drop the five `EmbodimentMap` entries | FAIL | | | | | |
| stop splitting `a / b:` labels | FAIL | FAIL | | | | |
| stop crediting a base class's entries | FAIL | | FAIL | | | |
| treat every class as a dataclass | FAIL | | | FAIL | | |
| drop the `ClassVar` exclusion | | | | | FAIL | |
| make the scan reach nothing | | FAIL | | | | FAIL |

Gate: `ruff check` / `ruff format --check` clean over `strands_robots tests
tests_integ`; `mypy` 24 == 24 against a pristine `upstream/main` worktree with no
branch-only error.

Blast radius - every test that walks a source tree, inspects a docstring, reads
`AGENTS.md` or names the changed modules (319 files), run on this branch and on
`upstream/main` with the new test file excluded so the two selections are identical:
**21 failed / 12474 passed / 56 skipped on both**, with the failing-ID sets matching
and nothing on either side alone. All 21 are dependency-absent in this environment and
green on CI: 17 `tests/mesh/test_iot_*` needing `awsiotsdk`, 3 in `tests/training`
from an older `draccus`, and one `lerobot` lazy-registration case.

## Artifact

The capture script reads whichever tree it runs against for `EmbodimentMap`'s own
`Attributes:` block, builds an embodiment from exactly the fields that block names,
and drives a real MuJoCo SO-101 headless with the checkpoint's action stream. Package
code is identical in both rows - only the docstring differs. The first frames of the
two rollouts agree to 0.027 mean abs level (same starting pose); they end 25.76 apart.

Attributes-block completeness

### DIFF
--- a/changelog.d/2587-attributes-block-names-every-declared-field.md
+++ b/changelog.d/2587-attributes-block-names-every-declared-field.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- **policies**: a config dataclass's `Attributes:` block now names every field it
+  declares. `EmbodimentMap` documented five of ten, omitting the unit-conversion
+  group (`state_units`, `action_units`, `gripper_index`, `gripper_joint_range`,
+  `joint_mids`), so an embodiment written from that block converted nothing and a
+  degrees-trained SO-arm checkpoint's values reached the sim unconverted;
+  `PackStateProcessorStep`, which receives four of them, and `ProtoMotionsConfig`
+  had the same gap. Graded for every dataclass with such a block by
+  `tests/test_attributes_docstring_completeness.py`, the attribute-list counterpart
+  to the existing `Args:` completeness guard.
--- a/strands_robots/policies/lerobot_local/embodiment.py
+++ b/strands_robots/policies/lerobot_local/embodiment.py
@@ -626,6 +626,21 @@ def register_pack_state_step() -> type | None:
             state_keys: Ordered robot/sim scalar keys composing the state vector.
             expected_dim: Model's declared ``observation.state`` dimension.
             dim_policy: ``"strict"`` | ``"pad"`` | ``"truncate"``.
+            state_units: ``"native"`` (the default - pack the sim's own values)
+                or ``"degrees"`` (convert the sim's radian joints to the model's
+                training units before packing). Mirrors
+                :attr:`EmbodimentMap.state_units`, which is where this step's
+                value comes from.
+            gripper_index: Column of the gripper inside ``state_keys``, which
+                speaks ``RANGE_0_100`` rather than degrees. ``-1`` (the default)
+                = no distinct gripper column.
+            gripper_joint_range: The sim gripper joint's ``[min, max]`` radians,
+                used to map that column onto 0..100. Empty (the default) =
+                convert the gripper like an arm joint.
+            joint_mids: Per-joint calibration mid-points in degrees, aligned to
+                ``state_keys``, subtracted from the arm columns so the packed
+                state is mid-centered like LeRobot's ``DEGREES`` mode. Empty
+                (the default) = mid 0.
         """
 
         state_keys: list[str] = field(default_factory=list)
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Docs-plus-guard PR with no behavior changes: completes the `Attributes:` blocks on three config dataclasses (`EmbodimentMap`, `PackStateProcessorStep`, `ProtoMotionsConfig`) so every declared field is documented — closing the gap where the unit-conversion group (`state_units` / `action_units` / `gripper_index` / `gripper_joint_range` / `joint_mids`) was invisible to a caller building an embodiment from the block, with the measured saturate-the-radian-limits consequence — and adds `tests/test_attributes_docstring_completeness.py`, an AST-only completeness guard mirroring the existing `Args:` guard's scope boundary (only classes that already carry the block are graded), label conventions, and base-class credit.

## What's good

- The new docstring entries were spot-checked against the conversion code and match: the empty-`gripper_joint_range` fallback (convert like an arm joint), the `state_units == "degrees"` gate in `PackStateProcessorStep.observation`, and the by-name output pairing the `onnx_out_names` entry describes are all what the code does.
- The guard is non-vacuous by construction: minimum-graded-surface floors, a by-name pin on the fields whose omission motivated the change, and planted-defect controls for each parser behavior (combined labels, base-class credit, dataclass-only scope, `ClassVar` exclusion, section bounding).
- Package root derived from the imported symbol rather than a path literal, so a moved package fails loudly instead of scanning nothing.
- Changelog fragment follows the `changelog.d/<pr-number>-<slug>.md` convention; no host paths, no emojis, ASCII-clean.

Verified locally at head `a3568aaa`: the new guard is 37 passed, the existing `Args:` guard 416 passed, and `ruff check` / `ruff format --check` are clean over the changed files.

PR #168, opened 2026-05-17
fix(libero): snapshot-and-restore canonical qpos for procedurally-generated MJCFs (#166)

yinsong1986:
## Summary

Implements #166's canonical fix per [the reviewer's verification](https://github.com/strands-labs/robots/pull/168#issuecomment-4468703705) and [recommendation](https://github.com/strands-labs/robots/pull/168#issuecomment-4468709540). Initial scope was keyframe-only; verification showed that's inert on the actual codepath because `_generate_scene_from_bddl`-emitted MJCFs have `model.nkey == 0`. Updated to a two-branch helper with snapshot-and-restore as the fallback that fires on the procedural codepath.

## What's verified by review (the gap this PR closes)

- ep1 t=0 renders the LIBERO living-room scene correctly. **ep2-5 t=0 render the degraded mid-ep1 view byte-for-byte**, post-`reset()`.
- The procedurally-generated MJCFs from PR #165 have **zero `<keyframe>` elements** (verified via `grep` + `mujoco.MjModel.from_xml_path` returning `nkey == 0`).
- Initial keyframe-only PR was therefore inert: the helper's `if nkey <= 0: skip` guard fired unconditionally.
- The small RGB Δ=120 vs Δ=149 swing observed across branches is attributable to `init_jitter=0.02` differing (because #167 was closed), NOT keyframe restoration.

## What this PR ships

`LiberoAdapter._apply_canonical_state` now has **two branches**:

| Branch | Fires when | Mechanism |
|---|---|---|
| **Keyframe** (preferred) | `model.nkey > 0` | `mj_resetDataKeyframe(model, data, scene_keyframe_index)` + `mj_forward`. LIBERO-authored hand-written scenes (upstream `libero/libero/assets/scenes/`) ship a `<keyframe>` element. |
| **Snapshot-and-restore** (fallback) | `model.nkey == 0` | First episode captures `data.qpos.copy()` / `data.qvel.copy()` after `super()` + `_install_libero_cameras` have run. Every subsequent episode does `np.copyto` + `mj_forward`. **Procedurally-generated MJCFs from #165 hit this branch** — the actual #166 fix on the codepath `examples/libero_mujoco.py` exercises. |

Both branches end with `mj_forward` so `xpos` / `xquat` / sensor data is current before the next `get_observation` call.

### Snapshot guarantees

- Per-instance state (`self._canonical_qpos` / `self._canonical_qvel`); no leak across `evaluate_benchmark` calls.
[...]

PR #2589, opened 2026-08-22
fix(benchmarks/libero): an orientation that encodes no rotation is not the identity

### LINKED ISSUE #2588: LiberoAdapter refuses a non-rotation quaternion from config and accepts one from the backend, where `or 1.0` reads the all-zero spelling as identity
## What

`LiberoAdapter` validates a config-supplied end-effector orientation and refuses one that is not a unit quaternion. Four lines of the same constructor comment away, the orientation it reads *from the backend* is checked for shape only, and the helper that consumes it converts a quaternion of zero norm into the identity rotation rather than refusing it.

The two values feed the same observation. `strands_robots/benchmarks/libero/adapter.py`, on `146ace1`:

```python
# #1802 - EEF source corrections for backends without the RoboSuite
# grip site (Isaac). Validated eagerly: a malformed offset silently
# feeding GR00T garbage state is exactly the failure class #168
# spent 30 rounds bisecting.
self._eef_quat_offset = _validated_float_vector(eef_quat_offset, 4, "eef_quat_offset")
if self._eef_quat_offset is not None:
    norm = float(np.linalg.norm(self._eef_quat_offset))
    if not (0.99 < norm < 1.01):
        raise ValueError(f"eef_quat_offset must be a unit quaternion (wxyz); got norm={norm:.4f}")
```

and the payload side, `_extract_pose`:

```python
raw_quat = json_block.get("quaternion")
if isinstance(raw_quat, list) and len(raw_quat) == 4 and all(isinstance(c, (int, float)) for c in raw_quat):
    quat = [float(c) for c in raw_quat]
```

Four numbers is the whole test. `[0.0, 0.0, 0.0, 0.0]` and `[nan, 0.0, 0.0, 0.0]` both pass it.

## Measurement

`_quat_wxyz_rotate_vec` normalises with `float(np.linalg.norm(q)) or 1.0`, so a zero-norm quaternion divides by 1.0 and the rotation is computed from `w=x=y=z=0`. Against the `[0, 0, 0.097]` hand offset that #1802 added for the Isaac wrist body:

```
identity  [1, 0, 0, 0]  rotate([0, 0, 0.097]) -> [0.0, 0.0, 0.097]
all-zero  [0, 0, 0, 0]  rotate([0, 0, 0.097]) -> [0.0, 0.0, 0.097]   <- indistinguishable
NaN       [nan,0, 0, 0] rotate([0, 0, 0.097]) -> [nan, nan, nan]
```

So an orientation that was never written is read as "no rotation", the body-frame offset is added along world Z whatever the wrist is actually doing, and `_read_eef_state` returns the result as a corrected position. The reported orientation goes out the other side of `_quat_wxyz_multiply([0,0,0,0], offset)` as `[0, 0, 0, 0]`, which is not a rotation either.

Both land in the observation dict a policy consumes. That is the failure class the comment quoted above names by number: state that is silently wrong rather than absent.

## Why the existing guards do not catch it

- The config check is on the *offset*, not on the body orientation the offset is applied in. It cannot see the payload.
- `_extract_pose` already drops a quaternion of the wrong length (`test_malformed_quaternion_length_is_dropped`), and the caller already has an honest branch for a missing orientation - it logs `"reported no orientation; cannot express the body-frame offset - using the uncorrected position"`. A degenerate quaternion is the same unreadable orientation with the length right, and it takes the other path.
- `or 1.0` is what makes the difference invisible. Without it the division would at least produce NaN and be noticed.

## The rest of the tree already settled this contract

`tests/policies/protomotions/test_state_utils_unit_quaternion_contract.py` states it directly, including this exact value:

> an all-zero orientation - the spelling of one that was never written - being read as a rotation that *negates* the angular velocity it is supposed to body-frame

and names where the rule is applied:

> Three other world-to-body helpers (`policies/wbc/control.quat_rotate_inverse`, `simulation/predicates._quat_rotate_inverse_wxyz` and the Newton backend's copy) all normalise internally, and the degenerate case is refused by the same-layer `policies/wbc/control` sibling rather than answered with a made-up rotation.

`quat_rotate_inverse` is the refusal to copy:

```python
norm = float(np.linalg.norm(q))
if norm < 1e-8:
    raise ValueError("quat_rotate_inverse: quaternion has ~zero norm; cannot normalise")
```

`_quat_wxyz_rotate_vec` is a fourth helper of the same shape, and the only one that answers.

## Proposed

1. `_extract_pose` drops a `quaternion` whose components are not all finite or whose norm is ~zero, exactly as it already drops a wrong-length one. The caller's existing "reported no orientation" branch then reports the uncorrected position instead of a confidently wrong one, and no eval aborts - which is the constraint that put the `except Exception` around `get_body_state` in the first place.
2. `_quat_wxyz_rotate_vec` refuses a non-rotation instead of substituting `1.0`, matching `quat_rotate_inverse`'s threshold and message shape, so the made-up rotation cannot return through a future caller.

## Deliberately out of scope

The same `or 1.0` appears in three matrix-to-quaternion converters (`simulation/isaac/simulation.py:168`, `mesh/sensors.py:100`, `benchmarks/libero/adapter.py`'s own `_mat_to_quat` siblings). Those build the quaternion themselves from a rotation matrix, and no finite matrix reaches them with a zero-norm result - every branch sets one component to `0.25 * s` with `s > 0`, or divides by `s == 0` and produces a non-finite value rather than a zero. `or 1.0` guards nothing reachable there, and removing it is a separate question from this one.

Measured against `main` at `146ace1`.

cagataycali:
Fixes #2588

## The asymmetry

`LiberoAdapter` validates a config-supplied end-effector orientation and refuses one that is not a unit quaternion. The orientation it reads back *from the backend* feeds the same observation and was checked for shape only - four numbers, any four. The constructor comment beside the config check already states what that costs:

> `#1802` - EEF source corrections for backends without the RoboSuite grip site (Isaac). Validated eagerly: **a malformed offset silently feeding GR00T garbage state is exactly the failure class #168 spent 30 rounds bisecting.**

`_extract_pose` had no such check, and `_quat_wxyz_rotate_vec` normalized with `float(np.linalg.norm(q)) or 1.0`.

## What the `or 1.0` did

A zero-norm quaternion divided by `1.0`, so the rotation was computed from `w = x = y = z = 0`. Against the `[0, 0, 0.097]` wrist offset #1802 adds for the Isaac hand body, on `146ace1`:

```
identity  [1, 0, 0, 0]    rotate([0, 0, 0.097]) -> [0.0, 0.0, 0.097]
all-zero  [0, 0, 0, 0]    rotate([0, 0, 0.097]) -> [0.0, 0.0, 0.097]   <- indistinguishable
NaN       [nan, 0, 0, 0]  rotate([0, 0, 0.097]) -> [nan, nan, nan]
```

So an orientation that was never written was read as "no rotation": the body-frame offset was added along world Z whatever the wrist was actually doing, and `_read_eef_state` returned that as a *corrected* position. `_quat_wxyz_multiply([0,0,0,0], offset)` then reported `[0, 0, 0, 0]` as the orientation, which is not a rotation either. Both land in the observation dict a policy consumes - state that is silently wrong rather than absent.

## Why this is a fix and not a new rule

The tree had already settled the contract. `tests/policies/protomotions/test_state_utils_unit_quaternion_contract.py` names this exact value - "an all-zero orientation - the spelling of one that was never written" - and records where the rule is applied:

> Three other world-to-body helpers (`policies/wbc/control.quat_rotate_inverse`, `simulation/predicates._quat_rotate_inverse_wxyz` and the Newton backend's copy) all normalise internally, and the degenerate case is refused by the same-layer `policies/wbc/control` sibling rather than answered with a made-up rotation.

`_quat_wxyz_rotate_vec` was a fourth helper of the same shape and the only one that answered.

## The change

One rule, applied at the two places a quaternion arrives, plus the predicate they share:

| site | behaviour | why that one |
|---|---|---|
| `_extract_pose` | **drops** the orientation | The caller wraps `get_body_state` in `except Exception` precisely so a state read cannot abort an eval, and it already has an honest branch for a missing orientation (`"reported no orientation; cannot express the body-frame offset - using the uncorrected position"`). A degenerate quaternion is the same unreadable orientation with the length right, so it takes the path the wrong-length one already took. |
| `_quat_wxyz_rotate_vec` | **raises** | So the made-up rotation cannot return through a caller that has not gone through `_extract_pose`. Threshold and message shape match `quat_rotate_inverse`. |

A large-magnitude quaternion is *not* refused - it encodes an ordinary rotation. `math.hypot` is used rather than `np.linalg.norm` for that: the latter squares before summing, so `[1e200, 1e200, 0, 0]` overflows to `inf` and would be refused despite normalizing to a plain 45-degree rotation. That also keeps the suite free of a numpy overflow warning.

## Deliberately out of scope

The same `or 1.0` appears in three matrix-to-quaternion converters (`simulation/isaac/simulation.py`, `mesh/sensors.py`, and this module's sibling). Those build the quaternion themselves from a rotation matrix, and no finite matrix reaches them with a zero-norm result - every branch either sets one component to `0.25 * s` with `s > 0`, or divides by `s == 0` and produces a non-finite value rather than a zero. `or 1.0` guards nothing reachable there, and removing it is a separate question.

## Overlap with #2497, already checked

`check_merge_base_overlap.py --all-open` reports `#2497 + #2589` sharing `strands_robots/benchmarks/libero/adapter.py`, so the composition is answered here rather than left to be derived.

#2497's entire change to that file is **one comment line**, a docs-drift fix at line ~70:

```
-# (``examples/libero/run_isaac.py``) loads Isaac Sim's bundled Franka USD,
+# (``examples/libero/run.py isaac``) loads Isaac Sim's bundled Franka USD,
```

This branch's hunks are at lines 4059 and 4102-4152. No shared line, no behavioural interaction - a comment naming an example script cannot reach the quaternion domain - and the merge is clean:

```
$ git merge-tree --write-tree --name-only pr2497 HEAD
exit 0, zero conflicted paths
```

## Verification

`tests/benchmarks/libero/test_eef_quaternion_encodes_a_rotation.py` - 56 cases. Both halves of the rule and the shared predicate, each parameterized over the spellings that passed the old shape check, and over rotations that must still be accepted:

- `test_the_all_zero_spelling_no_longer_answers_like_the_identity` is the measurement above as an assertion.
- `test_a_rotation_is_normalized_rather_than_trusted` compares `q` against `3q`; the formula is quadratic in the components and does not cancel, so this only holds if the helper normalizes internally.
- `test_a_rotation_still_rotates` pins 180 degrees about x mapping +z to -z, so a guard that refused everything would not pass.
- `test_the_position_beside_it_still_survives` pins that the drop is per-field, which is what makes the caller's existing branch reachable.

```
tests/benchmarks/libero/                      562 passed, 79 skipped
tests/test_refusal_messages_never_raise.py
tests/test_container_refusals_render_elementwise.py
tests/test_conversion_escape_is_closed.py
tests/simulation/test_input_validators_refuse_a_boolean.py     589 passed
ruff check / ruff format --check / mypy       clean
scripts/assemble_changelog.py --check        OK (631 pending)
```

The 5 failures in `tests/benchmarks/libero/test_libero_prewarm_init_state.py` are pre-existing and reproduce identically on `main` at `146ace1` in the same environment (`ModuleNotFoundError: No module named 'mujoco'`); they are untouched by this branch.

## §13 changelog

### DIFF
--- a/changelog.d/2588-eef-quaternion-encodes-a-rotation.md
+++ b/changelog.d/2588-eef-quaternion-encodes-a-rotation.md
@@ -1,0 +1,15 @@
+### Fixed: a LIBERO end-effector orientation that encodes no rotation is dropped, not read as the identity
+
+`LiberoAdapter` refused a config-supplied `eef_quat_offset` that was not a unit
+quaternion but accepted any four numbers as the orientation read back from the
+backend, and `_quat_wxyz_rotate_vec` normalised with `norm or 1.0` - so an
+all-zero quaternion, the spelling of an orientation that was never written,
+rotated a vector exactly as the identity does. The #1802 wrist offset was then
+added along a world axis whatever the body was doing and returned as a corrected
+position, and `[0, 0, 0, 0]` went out as the reported orientation; a NaN
+component propagated through both. `_extract_pose` now drops an orientation that
+encodes no rotation, exactly as it already dropped a wrong-length one, so the
+caller takes its existing "reported no orientation" path; `_quat_wxyz_rotate_vec`
+refuses one instead of substituting a norm, matching the settled
+`policies/wbc/control.quat_rotate_inverse` sibling. A large-magnitude quaternion
+is still accepted and normalised - it encodes an ordinary rotation.
--- a/strands_robots/benchmarks/libero/adapter.py
+++ b/strands_robots/benchmarks/libero/adapter.py
@@ -4056,7 +4056,15 @@ def _extract_pose(state: dict[str, Any] | None) -> tuple[list[float] | None, lis
             pos = [float(c) for c in raw_pos]
         raw_quat = json_block.get("quaternion")
         if isinstance(raw_quat, list) and len(raw_quat) == 4 and all(isinstance(c, (int, float)) for c in raw_quat):
-            quat = [float(c) for c in raw_quat]
+            if _rotation_quat_norm(raw_quat) is None:
+                # Four numbers that encode no rotation are an unreadable
+                # orientation, not the identity one - dropped exactly like a
+                # wrong-length quaternion above, so the caller takes its
+                # existing "reported no orientation" path instead of adding a
+                # body-frame offset in a rotation the body does not have.
+                logger.debug("LiberoAdapter: body-state quaternion encodes no rotation; dropped: %r", raw_quat)
+            else:
+                quat = [float(c) for c in raw_quat]
     return (pos, quat)
 
 
@@ -4091,10 +4099,57 @@ def _quat_wxyz_multiply(a: list[float], b: list[float]) -> list[float]:
     ]
 
 
+#: Below this norm a quaternion's direction is numerically unrecoverable, so it
+#: encodes no rotation. Same boundary as the settled sibling
+#: ``policies/wbc/control.quat_rotate_inverse``, so the two agree on where the
+#: domain ends rather than merely on the idea that it ends somewhere.
+_MIN_ROTATION_QUAT_NORM = 1e-8
+
+
+def _rotation_quat_norm(quat_wxyz: list[float]) -> float | None:
+    """Return the norm of ``quat_wxyz`` when it encodes a rotation, else ``None``.
+
+    ``None`` means the four numbers are not an orientation: a non-finite
+    component, or a norm too small for the direction to be recovered. Both are
+    spellings of an orientation that was never written, and neither is the
+    identity - #2588 measured the all-zero one being read as exactly that.
+
+    A large magnitude is *not* out of domain: a quaternion and its scaling encode
+    one rotation, so ``[1e200, 1e200, 0, 0]`` normalizes to a perfectly ordinary
+    45-degree rotation. ``math.hypot`` is used rather than ``np.linalg.norm`` for
+    exactly that case - the latter squares before summing, so it overflows to
+    ``inf`` and would refuse a rotation it could have recovered.
+    """
+    try:
+        q = [float(c) for c in quat_wxyz]
+    except (TypeError, ValueError):
+        return None
+    if len(q) != 4 or not all(math.isfinite(c) for c in q):
+        return None
+    norm = math.hypot(*q)
+    if not math.isfinite(norm) or norm <= _MIN_ROTATION_QUAT_NORM:
+        return None
+    return norm
+
+
 def _quat_wxyz_rotate_vec(quat_wxyz: list[float], vec: list[float]) -> list[float]:
-    """Rotate ``vec`` by the (normalized) wxyz quaternion: ``R(q) @ vec``."""
+    """Rotate ``vec`` by the wxyz quaternion: ``R(q) @ vec``.
+
+    ``quat_wxyz`` need not be exactly unit - it is normalized here - but it must
+    encode a rotation. A quaternion that does not is refused rather than
+    normalized by a substituted ``1.0``: that substitution makes an orientation
+    that was never written return the vector unrotated, which is what the
+    identity rotation returns, so the caller adds a body-frame offset along a
+    world axis and reports the result as corrected (#2588). The refusal matches
+    the same-shaped sibling ``policies/wbc/control.quat_rotate_inverse``.
+
+    Raises:
+        ValueError: If ``quat_wxyz`` encodes no rotation.
+    """
+    norm = _rotation_quat_norm(quat_wxyz)
+    if norm is None:
+        raise ValueError(f"_quat_wxyz_rotate_vec: quaternion encodes no rotation; got {quat_wxyz!r}")
     q = np.asarray(quat_wxyz, dtype=np.float64)
-    norm = float(np.linalg.norm(q)) or 1.0
     w, x, y, z = q / norm
     u = np.array([x, y, z], dtype=np.float64)
     v = np.asarray(vec, dtype=np.float64)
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes the asymmetry where a backend-reported EEF quaternion was shape-checked only while the config-supplied one was validated: `_quat_wxyz_rotate_vec`'s `norm or 1.0` made an all-zero (or NaN) orientation rotate exactly like the identity, so the #1802 body-frame wrist offset was applied along a world axis and returned as a corrected position. The fix applies the tree's already-settled contract (`policies/wbc/control.quat_rotate_inverse`) at both arrival points: `_extract_pose` drops a degenerate orientation onto the caller's existing "reported no orientation" branch, and `_quat_wxyz_rotate_vec` raises as a backstop. I verified the raise is unreachable in production: the helper's only non-test caller (`adapter.py:1978`) receives its quaternion exclusively from `_extract_pose`, which now gates it, so no eval path can newly abort. `math.hypot` over `np.linalg.norm` for overflow-safety on large-magnitude (still-valid) quaternions is correct and tested. Both helpers are `_`-private, so no public API or wire shape moves.

## What's good

- Ran the new test file plus the adjacent `test_isaac_eef_state_source.py` at head: 74 passed; ruff clean on both changed Python files; `assemble_changelog.py --check` passes.
- The full `tests/benchmarks/libero/` abort I hit locally reproduces identically on `main` at `146ace1c` in the same environment - pre-existing, not this branch.
- The #2497 overlap is answered in the description with a clean `merge-tree` rather than left to the reviewer to derive, per AGENTS.md step 8.
- The drop path logs at debug rather than silently substituting, and the test suite pins the per-field drop, the normalization (q vs 3q), and a positive rotation control - so a refuse-everything guard could not pass.

One non-blocking nit: the comment on `_MIN_ROTATION_QUAT_NORM` says "same boundary as the settled sibling", but the sibling refuses `norm < 1e-8` (strict) while this refuses `norm <= 1e-8` (inclusive). The boundary point itself differs; immaterial in float64 practice.

PR #1614, opened 2026-07-25
fix(lerobot-compat): harmonize version-floor text and drop stale/misleading guards

cagataycali:
## Summary
A small sweep of correctness/clarity drift in the lerobot-compatibility layer. Each change is text/comment/guard only (no policy, sim, rendering, or recording behavior changes) and carries a pinned regression test that fails on pre-fix code.

### streaming_dataset.py
- `repo_type`: no released lerobot's `StreamingLeRobotDataset.__init__` accepts a `repo_type` parameter (the versioned-dataset vs bucket storage split is not upstream), so the old `requires lerobot>=0.6.1` message was misleading -- 0.6.1 does not have it either. Reworded to `not supported by any released lerobot`; docstring and inline comment updated to match. The fail-fast on a non-default `repo_type` is unchanged (silently opening the wrong storage system stays forbidden).
- `return_uint8`: when the installed `StreamingLeRobotDataset` lacks the parameter it is dropped and frames stream as float32 (~4x the bandwidth of uint8). That is a real cost, not a semantics-preserving no-op, so it now logs a warning when dropped instead of doing so silently. Policies still normalize correctly either way.

### training/lerobot.py, training/reward.py
- Harmonized the sample-weighting / reward-model version-floor text to `lerobot >= 0.6.0`. `lerobot.rewards` and `lerobot.utils.sample_weighting` exist only at 0.6.x, and the actually-raised errors already said 0.6.0; the surrounding docstrings/comments said 0.5.2. Now consistent.

### tools/use_lerobot.py
- Dropped the stale `lerobot.common.datasets.push` blocklist prefix. That module was removed upstream (`lerobot.common` is now `control_utils` / `train_utils` / `wandb_utils` only), so the prefix matched nothing. Hub push/upload stays blocked via the `push_to_hub` / `upload_folder` method blocklist (added a note).
- Stopped advertising `scripts.lerobot_train.train` in the discovery `entry_points`: it is refused by the `lerobot.scripts` blocklist, so listing it as a call target was self-contradictory.

### tools/lerobot_teleoperate.py
- DAgger shells `python -m lerobot.scripts.lerobot_rollout`, which exists only in lerobot>=0.6.0. Added an `importlib.util.find_spec` preflight so an older install fails fast with an actionable upgrade hint instead of an opaque `No module named` subprocess error.

### policies/lerobot_local/embodiment.py
[...]

PR #732, opened 2026-06-27
fix(teleoperate): port record/replay branches to lerobot 0.5 nested CLI

cagataycali:
## Problem

`lerobot_teleoperate`'s command builder shells out to `lerobot.scripts.lerobot_record` and `lerobot.scripts.lerobot_replay` using the **pre-0.5 flat CLI**:

```
--robot-path  --robot-port  --repo-id  --num-episodes  --single-task
--episode-time-s  --reset-time-s  --no-video  --push-to-hub        (record)
--robot-path  --policy-path  --episode  --robot-port               (replay)
--camera-config name=type:path:fps:WxH                             (cameras)
```

lerobot 0.5's `lerobot_record` / `lerobot_replay` are **draccus-nested** (`RecordConfig`/`ReplayConfig`, `DatasetRecordConfig`/`DatasetReplayConfig`). The flat flags no longer exist, so the scripts exit with `unrecognized arguments` (exit code 2) and **data collection never runs** through the tool. The teleop branch of the same builder already used the correct nested `--robot.type` / `--teleop.type` form; the record and replay branches were simply stale.

Verified against lerobot 0.5.2:

```
# OLD flat args -> rejected
$ python -m lerobot.scripts.lerobot_record --robot-path so101_follower --robot-port /dev/null \
    --repo-id u/d --num-episodes 1 --single-task x --no-video
lerobot_record.py: error: unrecognized arguments: --robot-path so101_follower --robot-port /dev/null \
    --repo-id u/d --num-episodes 1 --single-task x --no-video        # exit 2
```

## Change

Port the record and replay branches to lerobot 0.5's nested schema:

- `--robot.type / --robot.port / --robot.id / --robot.left_arm_port / --robot.right_arm_port`
- `--teleop.type / --teleop.port / --teleop.id / --teleop.{left,right}_arm_port`
- `--dataset.repo_id / --dataset.num_episodes / --dataset.single_task / --dataset.fps / --dataset.episode_time_s / --dataset.reset_time_s / --dataset.root / --dataset.push_to_hub / --dataset.video` (record)
- `--dataset.repo_id / --dataset.episode / --dataset.root` (replay)
- cameras as the nested `--robot.cameras={name: {type: opencv, index_or_path: 0, width: 640, height: 480, fps: 30}}` dict
- booleans as explicit `true`/`false` values
- drop `--display_data` for replay (`ReplayConfig` has no such field; emitting it would error)

[...]

PR #2590, opened 2026-08-22
docs(tools): name the refusal a caller cannot otherwise handle

cagataycali:
## Problem

`build_lerobot_command`'s `Raises:` block documents `ValueError` alone:

```
Raises:
    ValueError: If ``action`` is unknown, ``replay`` is requested without
        ``dataset_repo_id``, or a numeric knob ... cannot be honored.
```

It also raises `RuntimeError`, from the DAgger preflight added 28 days later:

```python
if importlib.util.find_spec("lerobot.scripts.lerobot_rollout") is None:
    raise RuntimeError("dagger requires lerobot>=0.6.0: ...")
```

A `Raises:` block is the only place a caller learns which `except` clause to
write, so a caller who wrote exactly the handler the docstring licensed took an
escaped exception. Measured through the real builder, with the handler generated
from the docstring rather than hard-coded:

| call | before | after |
| --- | --- | --- |
| `dagger` on a lerobot with no rollout entry point | **ESCAPED `RuntimeError`** | handled `RuntimeError` |
| `dagger` with a bad `dagger_input_device` | handled `ValueError` | handled `ValueError` |
| `dagger` on a conforming install | 31-token argv | 31-token argv |

caller handler

Both columns run identical package code; only the `try/except` differs, because
each is generated from that column's own docstring. Rows 2-3 are byte-identical
across columns, which is the control.

`git log -S` dates it: the block was written in #732 (2026-06-27) and the
preflight landed in #1614 (2026-07-25). The class itself is right - a missing
module is an environment problem, not a bad argument - and
`test_build_dagger_command_preflights_missing_rollout_module` already pins it, so
code and tests agreed and only the prose did not.

## Change

`strands_robots/tools/lerobot_teleoperate.py` gains the `RuntimeError` entry,
naming the condition, why it is not a `ValueError`, and that it likewise precedes
the argv.

## Root cause

Nothing compared a `Raises:` block against the classes its function raises.
`tests/test_raises_docstring_completeness.py` adds that comparison for every
function in the package whose docstring **already** has a `Raises:` block - a
block that exists is checked for completeness rather than demanded, exactly as
its `Args:` sibling scopes itself. **258 surfaces graded, 1 uncovered.**

The population is module-level functions as well as methods, which matters here:
the `Args:` guard grades public methods of public classes, and the only surface
this finds is not a method, so a population borrowed unchanged from that guard
would not have seen it.

The check is deliberately one-directional. A documented class raised by a helper
the function delegates to is legitimate and common - 80 surfaces document a class
their own body never constructs, because a validator, a
`subprocess.run(check=True)` or an optional-dependency shim raises it one frame
down - so only the forward direction is a rule and the converse is left alone
rather than exempted case by case.

Three shapes are not refusals the function chose. A raw sweep reports 7; these
filters reduce it to the 1 above, and each is calibrated against the shipped
shape that needs it:

| shape | shipped example | why it is not required |
| --- | --- | --- |
| `raise error_holder[0]` | `device_connect.init_device_connect_sync` | re-raises an exception captured elsewhere; no class name at the raise site |
| `raise _factory_import_error(exc)` | `molmoact2.build_policy`, `registry.policies.import_policy_class` | a same-module factory `-> ImportError`; the annotation is what the block names, and does |
| raised and caught in one function | `IsaacSimulation.run_multi_policy` (`CooperativeStop`) | the caller never sees it, so documenting it would be the opposite error |

Evidence is read permissively and claims strictly - a class named anywhere in the
block credits that class - so the scan can only ever under-report.

## Tests

`3 failed / 74 passed` against `upstream/main`, `77 passed` after. The three:

```
test_every_refusal_a_function_makes_is_in_its_own_raises_block
  strands_robots/tools/lerobot_teleoperate.py:444 build_lerobot_command
  raises ['RuntimeError (line 658)'] with no Raises: entry

test_the_teleoperate_builder_documents_both_classes_it_raises
  documents only ['ValueError']

test_dagger_preflight_is_a_class_the_docstring_licenses
  dagger on a lerobot without the rollout entry point raises RuntimeError,
  which the Raises: block does not name (it names ['ValueError']);
  a caller handling only the documented classes takes it unhandled
```

The 74 that pass on both trees are the controls. Each filter and each
non-vacuity assertion was proven load-bearing by breaking it and checking that
exactly its own guard fires alongside the sweep:

| break | sweep reports | own control |
| --- | --- | --- |
| drop the caught-in-scope filter | `run_multi_policy raises CooperativeStop` | fails |
| drop the factory annotation | `molmoact2.build_policy raises _factory_import_error` | fails |
| accept any raise target | `init_device_connect_sync raises error_holder[0]` | fails |
| descend into nested defs | `LiberoAdapter._try_install_isaac_action_controller` | fails |
| re-narrow the scan | - | both non-vacuity tests fail (`only 5 surfaces scanned`) |
| drop the permissive evidence pass | - | fails |

## Verification

Measured at `8d5c9a6c` against `upstream/main` `8c6720aa`. 330 test files - every
test that walks a source tree, grades docstrings or reads `AGENTS.md`, plus all
of `tests/tools` and the repo-wide guards - run on both trees with the identical
selection:

| --- | --- | --- | --- |
| this branch | 17 | 15190 | 92 |
| `upstream/main` | 17 | 15189 | 92 |

The failing ID sets are identical (`comm` empty both ways); the `+1` passed is
the one new behavioural test in the modified file. All 17 are
`tests/mesh/test_iot_*` needing `awsiotsdk`, a declared extra CI installs.
`ruff check` / `ruff format --check` clean over the CI scope, and `mypy` reports
the same 24 pre-existing errors on both trees with none in the changed files.

No behaviour changed: the refusal, its class and its message are untouched, and
`AGENTS.md` records the convention under `API Consistency` beside the
"Match docstrings to semantics" rule it specialises.


yinsong1986 (approved):
## Summary

Docs-and-tests PR with no behaviour change: `build_lerobot_command`'s `Raises:` block gains the `RuntimeError` entry for the DAgger rollout preflight (#1614) that the `ValueError`-only block written in #732 never learned about, so a caller writing exactly the handler the docstring licenses no longer takes an escaped exception. The root cause is closed by a new repo-wide guard, `tests/test_raises_docstring_completeness.py`, which compares every existing `Raises:` block against the classes the function's own scope raises (one-directional, with three calibrated filters for re-raised captures, exception factories, and classes caught in scope), plus a behavioural test in `tests/tools/test_lerobot_teleoperate.py` that derives its handler from the block rather than hard-coding it. The refusal, its class and its message are untouched.

## What's good

- Verified on the head: the new sweep (18 tests) and the full `tests/tools/test_lerobot_teleoperate.py` (59 tests) pass; reverting only the docstring to the `upstream/main` version makes the sweep fail on exactly `build_lerobot_command`'s undocumented `RuntimeError`, so the guard is load-bearing and not vacuous.
- The sweep scopes itself to blocks that already exist (mirroring the `Args:` guard), reads evidence permissively and claims strictly, and carries non-vacuity tests (`>= 150` surfaces, module-level population, planted-defect detection) so a mis-rooted scan fails rather than passing silently.
- AGENTS.md convention entry added beside the "Match docstrings to semantics" rule it specialises; changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention (step 3); `ruff check` / `ruff format --check` clean on the changed files; no host paths, no non-ASCII in the new files.

PR #2591, opened 2026-08-22
docs(tools): drop the Actions: entry pose_tool dispatches nowhere

cagataycali:
## Problem

`pose_tool`'s `Actions:` block advertised

```
        - "calibrate_motor": Interactive motor calibration
```

alongside thirteen real entries, in the same quoted-and-colon format, under the
same `Motor Control:` sub-heading. `action == "calibrate_motor"` has never
existed in any commit: the entry and the tool arrived together in the original
commit (`10518bb0`, 2025-11-10) and the dispatch never followed.

The docstring is not merely documentation on this surface. `@tool` publishes it
as `tool_spec["description"]`, so the entry sat in the 1190-character string
handed to a model, which is the only place a caller learns which verbs the tool
answers.

Measured, driving the shipped tool:

| call | status | text |
| --- | --- | --- |
| `pose_tool(action="calibrate_motor")` | `error` | `Unknown action: calibrate_motor` / `Available actions: store_pose, ... emergency_stop` (13 names, this one absent) |
| `pose_tool(action="bogus_typo")` | `error` | byte-identical to the row above, modulo the echoed name |
| `pose_tool(action="list_poses")` | `success` | `No poses stored for robot probe` |

So the tool answered its own documented verb exactly as it answers a
misspelling, and the `Available actions:` list it printed contradicted the block
a model had just read. The tool described itself twice and the two disagreed -
with the half read first being the wrong one.

## Change

Drops the entry, and names where calibration does live. `pose_tool` has no
`input()` or `stdin` anywhere, so an "interactive" action is not merely
unimplemented on this surface - an agent tool has no channel to be interactive
on. Calibration already has two homes, and `lerobot_teleoperate` sets the
convention for pointing at them: its `Calibration Management:` section says "For
calibration management (list, view, backup, etc.), use the separate
lerobot_calibrate tool". The new `Calibration:` section follows it, naming
`lerobot_calibrate` for stored calibrations and `lerobot_teleoperate`'s
`auto_accept_calibration` for the prompt LeRobot shows when a device has none.

Nothing that worked is refused: the thirteen dispatched verbs, their
descriptions and the refusal text are untouched.

## Guard

`tests/tools/test_tool_actions_block_names_the_dispatched_vocabulary.py`
compares every `@tool` module's `Actions:` block against the vocabulary the
module dispatches, in both directions. Ten surfaces are graded and **nine were
already exact**, so this pins a convention the package already keeps.

The parser is calibrated against the three delimiter styles that ship - bare
(`status - ...`, `use_rosbridge`), double-quoted (`- "store_pose": ...`,
`pose_tool`) and double-backticked (`` - ``start``: ... ``, `gr00t_inference`) -
and against the two shapes that are not entries: a capitalised sub-heading, and
a wrapped continuation. Two separate rules exclude a continuation and both are
load-bearing, so both are pinned: a name must be followed *directly* by `:` or
`-` (which is what keeps `summary (JSON object: strategy...)` inside
`harness_memory`'s `save_trace` description from reading as a verb), and a wrap
that *does* begin `word:` is separated only by its deeper indent.

The existing guards on this contract sit at the published-schema boundary of the
MuJoCo `Simulation` tool and grade an enum
(`tests/simulation/mujoco/test_agent_boundary_refuses_unpublished_action.py`),
so a prose block was graded by nothing.

`pose_tool` states its vocabulary twice, and both halves are now graded: the
refusal's `Available actions:` list was already exact while the block was not,
which is how the tool came to contradict itself, so grading one half would have
left the other reachable.

Every docstring is read from source, which is also what makes the grading
independent of import order. `strands_robots.tools` exports each tool lazily
through module `__getattr__`, and the import system rebinds that package
attribute to the **submodule** as soon as anything imports the submodule
directly - `from strands_robots.tools.pose_tool import pose_tool`, which five
sibling test modules do, all sorting before this file. The cached binding is then
a module, `__getattr__` is never consulted again, and a read of
`tools.pose_tool` returns the *module* docstring, which carries none of the
tool's prose. The calibration-pointer test read the package attribute and so
answered correctly alone and wrongly in a suite run:

| run | reading the package attribute | reading from source |
| --- | --- | --- |
| this file alone | 18 passed | 21 passed |
| with `test_pose_tool.py` | **1 failed** / 79 passed | 83 passed |
| all of `tests/tools` | **1 failed** / 2565 passed | 2569 passed, 8 skipped |

The lazy-import guard in `tests/tools/test_tools_lazy_import.py` drops the cached
binding before each read for exactly this reason; reading from source needs no
such dance. A failure visible only in a suite run cannot be pinned by a
behavioural test - running two files in one order is not something the suite
guarantees - so `TestTheGuardReadsDocstringsFromSource` pins it structurally: no
tool is resolved through the package attribute without first dropping the cache,
in either spelling (`tools.pose_tool` and `getattr(tools, "pose_tool")` reach the
same rebound attribute). That check fails on the offending line with the file
running **alone**, deterministically. Nothing else in the tree takes that route:
of the 21 lazily exported names, 16 are also submodules and so shadowable, and
across `tests/`, `strands_robots/`, `examples/` and `scripts/` this was the only
package-attribute read of one.

## Verification

Pre-fix split, the guard copied onto pristine `52b2460f`: **2 failed / 19
passed** -> 21 passed. The headline names the surface and the verb:

```
AssertionError: an Actions: block advertises a verb the module dispatches
nowhere, so the tool answers its own documented action with its Unknown-action
refusal: {'pose_tool.py::pose_tool': ['calibrate_motor']}
```

Every guard is load-bearing - twelve injected regressions, each failing exactly
and only its own:

| --- | --- |
| re-add the phantom entry | headline (phantom direction) |
| accept every candidate line (drop the indent rule) | continuation control |
| drop the double-backtick delimiter | converse direction (`gr00t_inference`, 10 verbs) + backtick style + non-vacuity |
| let a capitalised sub-heading be an entry | headline (`Motor Control`, `Pose Management`, `System`) + sub-heading control |
| re-narrow the scan to one module | both non-vacuity guards |
| drop a verb from the refusal list | `pose_tool` self-consistency |
| remove the calibration pointer | pointer control |
| restore the package-attribute read | read-from-source guard, **with the file alone** (`line 259: tools_package.pose_tool`) |
| drop the cache pop from the parity test | read-from-source guard, `getattr` spelling (`line 436`) |
| read the *module* docstring instead of the function's | pointer control + source/lazy parity + missing-function guard |
| let a missing function read as an empty docstring | missing-function guard |
| stop detecting the cache pop | read-from-source guard, on the read the pop legitimately permits |

Gate: 449 test files - every test touching `pose_tool`, an `Actions:` block, a
tool docstring or the tools package surface, every guard that walks a source tree
or grades docstrings, all of `tests/tools`, plus the repo-wide
docs/changelog/string guards. Run over the branch and over pristine `52b2460f`
**with the new file excluded**, so any difference would be the source change
alone:

| tree | result |
| --- | --- |
| this branch | 17 failed, 21467 passed, 93 skipped |
| pristine `52b2460f` | 17 failed, 21467 passed, 93 skipped |

Failing-ID sets are byte-identical, `comm` empty both ways. Twelve of the 17 are
dependency-absent locally and pass in CI, which installs them: eight in
`tests/mesh/test_iot_connect_timeout_domain.py` need `awsiotsdk`, two in
`tests/test_device_connect_hardening.py` need `device_connect_agent_tools`, two in
`tests/training/test_lerobot.py::TestInProcessTrainerCorrectness` need
`draccus>=0.11.6` (0.10.0 locally). The remaining five are import-order artifacts
of a wide `tests/simulation` selection and give 29 passed run alone on pristine
`52b2460f`.

The base moved during this work (`dd63c3c7` -> `52b2460f`); `upstream/main` is
merged in, the true file overlap was empty, and every number above is measured at
`73ce932a`.

tests_integ`; `mypy` is 24 == 24 against a pristine worktree, none in the
changed files.

## Artifact

One script per column: read that tree's own `tool_spec["description"]`, parse
its `Actions:` block, and drive exactly the verbs it names - only verbs the tool
does *not* dispatch are driven, so nothing in the capture can act. The block
parser is loaded from the guard itself, so the picture and the test cannot
disagree about what a block names. Package code is identical in both columns.

The dispatched vocabulary and the refusal text are byte-identical across the two
columns (asserted in the capture, not eyeballed): what changes is which verbs
the description offers. The capture loads `named_actions` from the guard, and
that parser is byte-identical to the reviewed revision, so the figure still
measures what the tests do.

pose_tool's Actions: block against the vocabulary it dispatches


yinsong1986 (commented):
## Summary

Removes the phantom `"calibrate_motor"` entry from `pose_tool`'s `Actions:` block (advertised since the original commit, dispatched nowhere, so the tool refused its own documented verb), adds a `Calibration:` section pointing at `lerobot_calibrate` / `lerobot_teleoperate`'s `auto_accept_calibration`, and adds a guard that compares every tool module's `Actions:` block against its dispatched vocabulary in both directions. Verified the docstring change reaches the published `tool_spec["description"]` (phantom gone, calibration pointer present), the AST parser/dispatch collection logic is sound, and lint/format are clean. One blocker: the new self-consistency test is order-dependent and fails deterministically in the full-suite run (see inline comment).

## What's good

- The guard scans by AST, so it needs no optional hardware deps, and pins its own non-vacuity (`_MINIMUM_SURFACES`, pose_tool-in-scope, no-empty-block).
- Parser calibration tests cover all three shipped delimiter styles plus both non-entry shapes (sub-heading, wrapped continuation), so a parser regression fails loudly rather than grading nothing.
- Scope discipline: the thirteen dispatched verbs, their descriptions and the refusal text are untouched, matching the PR description exactly.

tests/tools/test_tool_actions_block_names_the_dispatched_vocabulary.py
```diff
@@ -0,0 +1,324 @@
...
+            f"pose_tool's refusal lists {sorted(advertised)} but dispatches {sorted(dispatched)}"
+        )
+
+    def test_calibration_is_pointed_at_the_tools_that_own_it(self) -> None:
+        """Dropping the entry must not leave calibration unfindable.
+
+        ``lerobot_teleoperate`` sets the convention: a tool that does not own
+        calibration names the one that does.
+        """
+        doc = inspect.getdoc(getattr(tools_package.pose_tool, "__wrapped__", tools_package.pose_tool)) or ""
```
yinsong1986: [MUST FIX] This test fails deterministically in the full-suite run, which is what the required check `call-test-lint / Test and Lint` executes, so the PR lands (or holds) that check red.

`strands_robots/tools/__init__.py` exports `pose_tool` lazily via module `__getattr__`. But any *direct* submodule import - `from strands_robots.tools.pose_tool import pose_tool`, which `tests/tools/test_agent_tool_parameter_descriptions.py` and `tests/tools/test_pose_tool*.py` all perform, and which pytest collects alphabetically **before** this file - makes the import system bind the **submodule** onto the package attribute. `globals()["pose_tool"]` is then a module, `__getattr__` is never consulted again, and this line reads the *module* docstring ("Robot Pose Management Tool..."), which contains no `lerobot_calibrate`. Reproduced on this head:

```
pytest tests/tools/test_tool_actions_block_names_the_dispatched_vocabulary.py  ->  18 passed
pytest tests/tools/                                                            ->  1 failed, 2565 passed
pytest tests/tools/test_pose_tool.py tests/tools/test_tool_actions_...py       ->  1 failed, 79 passed  (minimal pair)
```

The fix is one line - resolve the function without going through the package attribute, e.g.:

```python
from strands_robots.tools.pose_tool import pose_tool as _pose_tool
doc = inspect.getdoc(getattr(_pose_tool, "__wrapped__", _pose_tool)) or ""
```

or, more in keeping with the rest of this file, read the function docstring by AST (`ast.get_docstring` of the `pose_tool` `FunctionDef` in `pose_tool.py`), which also keeps the whole module free of import-order coupling. This can't be deferred: a test the PR itself adds that reds the one required status check blocks the merge outright, and if it slipped through on a lucky ordering it would make `main`'s full runs order-dependently red.

cagataycali: Fixed in f4ba1bf8. Confirmed exactly as diagnosed - 18 passed alone, and the
minimal pair reproduced `1 failed / 79 passed` with the assertion showing the
module docstring (`'Robot Pose Management Tool\n\nThis tool provides...'`) where
the tool's should have been. In a fresh interpreter the package attribute is a
`DecoratedFunctionTool`; after `from strands_robots.tools.pose_tool import
pose_tool` it `is` the submodule.

Took the AST route you suggested, which is what every other graded surface in the
file already does: a `tool_docstring(module, function)` helper reads the
`FunctionDef` docstring from source. It is a faithful substitute rather than
merely a safe one - 3382 == 3382 characters, identical to the lazy read, pinned
by `test_the_source_read_matches_what_the_lazy_export_resolves_to`, which drops
the cached binding first so it compares the two readings rather than whatever a
prior import left behind. A rename now fails loudly instead of grading an empty
docstring.

Since the failure is invisible to the file's own run, I added a structural guard
rather than relying on a behavioural one: no tool may be resolved through the
package attribute without first dropping the cache, in either spelling -
`tools.pose_tool` and `getattr(tools, "pose_tool")` reach the same rebound
attribute. Restoring the old read makes it fail **with the file running alone**,
deterministically. The pop-first exemption is not a carve-out for a file: it is
what makes such a read correct, which is why
`tests/tools/test_tools_lazy_import.py` has to do it, and stubbing out the
pop-detector makes the guard flag that legitimate read too.

Swept for other instances - of the 21 lazily exported names 16 are also
submodules and so shadowable, and this was the only package-attribute read of one
anywhere in `tests/`, `strands_robots/`, `examples/` or `scripts/`.

| run | before | after |
| --- | --- | --- |
| this file alone | 18 passed | 21 passed |
| minimal pair with `test_pose_tool.py` | 1 failed / 79 passed | 83 passed |
| all five direct-import siblings + the lazy guard | - | 333 passed |
| all of `tests/tools` | 1 failed / 2565 passed | 2569 passed, 8 skipped |

`upstream/main` is also merged in (the base moved to `52b2460f`; true file
overlap was empty). Full gate re-run at `73ce932a` over 449 files with the new
file excluded: `17 failed, 21467 passed, 93 skipped` on both the branch and
pristine `52b2460f`, failing-ID sets byte-identical. The five injected
regressions for the new guards are in the PR description's break table.

yinsong1986 (approved):
## Summary

Removes the phantom `"calibrate_motor"` entry from `pose_tool`'s `Actions:` block - a verb advertised since the tool's original commit but dispatched nowhere, so the published `tool_spec["description"]` promised an action the tool refused byte-identically to a misspelling - and replaces it with a `Calibration:` section pointing at the two tools that actually own calibration (`lerobot_calibrate` for stored calibrations, `lerobot_teleoperate`'s `auto_accept_calibration` for the first-run prompt), following the convention `lerobot_teleoperate` already sets. A new guard (`tests/tools/test_tool_actions_block_names_the_dispatched_vocabulary.py`) compares every tool module's `Actions:` block against its dispatched vocabulary in both directions, plus `pose_tool`'s refusal-list half of the same contract. The one prior `[MUST FIX]` on this PR (the package-attribute docstring read that failed deterministically in suite runs) is resolved at this head: every graded docstring is now read from source by AST, and `TestTheGuardReadsDocstringsFromSource` pins that structurally so the regression fails with the file running alone. Verified at `73ce932a`: the guard passes alone (21) and alongside `test_pose_tool.py` (83, where the pre-fix commit gave 1 failed / 79 passed), all of `tests/tools` is green (2570 passed, 8 skipped), the repo-wide docstring/xref/ASCII/host-path guards pass, and `ruff check` / `ruff format --check` are clean on both changed files. The `pose_tool.py` diff touches only the docstring - the thirteen dispatched verbs and the refusal text are unchanged - and the changelog lands as a `changelog.d/` fragment per AGENTS.md step 3.

## What's good

- The guard grades both halves of `pose_tool`'s self-description, so the exact failure mode (block and refusal disagreeing) cannot recur on either side.
- The parser's calibration tests pin all three shipped delimiter styles and both non-entry shapes, so a parser that stops recognising a style fails loudly instead of grading nothing.
- The import-order fix is pinned structurally (AST check on this file's own reads), which is the only form that can fail deterministically for a defect visible only in suite ordering.
- Non-vacuity guards (`_MINIMUM_SURFACES`, pose_tool-in-scope, no-empty-block) keep a re-narrowed or mis-rooted scan from passing quietly.

PR #2593, opened 2026-08-22
fix(mesh): STRANDS_MESH=off silently starts the mesh it was meant to kill

### LINKED ISSUE #2592: fix(mesh): STRANDS_MESH=off silently starts the mesh it was meant to kill - the vocabulary is split across two modules, so neither reader can report a typo
## What

`STRANDS_MESH` is documented in README's Configuration table as a hard kill switch:

> `false`/`0`/`no` is a hard kill switch that also overrides an explicit `mesh=True`, and refuses the robot-less gateway peer the `robot_mesh` tool would otherwise start in a coordinator process

An operator who spells that intent any other way gets a live mesh, and nothing says so. Measured through the shipped resolvers:

| `STRANDS_MESH` | `robot._mesh_env_opt_in()` | `core.mesh_disabled_by_env()` | reported |
|---|---|---|---|
| `true` / `1` / `yes` | True | False | SILENT |
| `false` / `0` / `no` | False | True | SILENT |
| unset / `""` | False | False | SILENT |
| **`off`** | False | **False** | **SILENT** |
| **`on`** | False | False | **SILENT** |
| **`ture`** | False | False | **SILENT** |
| **`disabled`** | False | **False** | **SILENT** |

The consequence, with an explicit `mesh=True`:

```
STRANDS_MESH='false'    mesh=True -> None (gate refused)
STRANDS_MESH='off'      mesh=True -> Mesh CONSTRUCTED (gate passed)
STRANDS_MESH='disabled' mesh=True -> Mesh CONSTRUCTED (gate passed)
```

So the operator who asked for no mesh gets a real Zenoh session, a `gateway-*` peer advertised to the fleet and the threads `Mesh.start` spawns - the exact outcome #2515 removed for the gateway path, arriving through the vocabulary instead of through a missed call site.

## Why this violates a rule already in AGENTS.md

> **Warn on unrecognized values** - `STRANDS_ROBOT_MODE=foo` (typo) must `logger.warning(...)`, not silently fall through. Silent typo'd env vars surprise users hours later.

`STRANDS_MESH` has no such report on either reader.

## `off` is not a strawman - the package teaches it

Two other variables in this tree accept the spellings `STRANDS_MESH` silently ignores:

| variable | vocabulary | site |
|---|---|---|
| `STRANDS_ROBOT_MESH_DC` | `("off", "0", "false", "no")` | `tools/robot_mesh.py:798` |
| `REACHY_DAEMON_TLS` | `("1", "true", "yes", "on")` | `device_connect/reachy_transport.py:69` |

An operator who learned `off` from the first one and writes it here is silently overruled.

## Root cause: the vocabulary is split, so neither reader can report

The two halves live in different modules and each correctly treats the other's spellings as not its business:

- `robot._mesh_env_opt_in` reads the affirmative half `("true", "1", "yes")` to decide whether a bare `Robot()` opts in. To it, `false` is simply not an opt-in - not a mistake.
- `mesh.core.mesh_disabled_by_env` reads `_MESH_KILL_SWITCH_VALUES = ("false", "0", "no")`. To it, `true` is simply not a kill - not a mistake.

Neither is in a position to call a third value wrong, because each would have to know the other's half to tell a typo from a legitimate spelling it does not own. That is a structural reason for the silence rather than an omission at either site, and it is why the fix is a single owner holding both halves rather than a warning added twice.

The precedent is in the tree: `mesh/_backend_select.py` is the "Sole owner of the `STRANDS_MESH_BACKEND` vocabulary", written for the same failure (`STRANDS_MESH_BACKEND=iott` produced a plain Zenoh session indistinguishable from an explicit `zenoh`) and reporting once per distinct offending value.

## Suggested direction

One owner returning a tristate - affirmative / negative / said-nothing - with the report on the third case, keyed by the offending value so a per-construction-site call does not put one line per call in the log.

It cannot live in the mesh package: `robot.py` reaches `strands_robots.mesh` only lazily, inside the function that starts one, because `mesh/__init__.py` eagerly imports the Zenoh-backed session and core. A resolver there would make that import eager and break the documented no-zenoh-installed path.

**Report only, not a wider vocabulary.** Making `off` a synonym for `false` would be a behaviour change on a safety switch and is a separate call for whoever owns it; naming the accepted spellings in the warning tells the operator who wrote `off` what to write instead, which is what the rule above asks for.

cagataycali:
Closes #2592

## Problem

`STRANDS_MESH` is documented in README's Configuration table as a hard kill switch:

> `false`/`0`/`no` is a hard kill switch that also overrides an explicit `mesh=True`, and refuses the robot-less gateway peer the `robot_mesh` tool would otherwise start in a coordinator process

An operator who spells that intent any other way gets a live mesh, and nothing says so. Measured through the two shipped resolvers on `upstream/main`:

| `STRANDS_MESH` | `_mesh_env_opt_in()` | `mesh_disabled_by_env()` | reported |
|---|---|---|---|
| `true` / `1` / `yes` | True | False | silent |
| `false` / `0` / `no` | False | True | silent |
| unset / empty | False | False | silent |
| **`off`** | False | **False** | **SILENT** |
| **`on`** | False | False | **SILENT** |
| **`ture`** | False | False | **SILENT** |
| **`disabled`** | False | **False** | **SILENT** |

The consequence, with an explicit `mesh=True`:

```
STRANDS_MESH='false'    mesh=True -> None (gate refused)
STRANDS_MESH='off'      mesh=True -> Mesh CONSTRUCTED (gate passed)
STRANDS_MESH='disabled' mesh=True -> Mesh CONSTRUCTED (gate passed)
```

So the operator who asked for no mesh gets a real Zenoh session, a `gateway-*` peer advertised to the fleet and the threads `Mesh.start` spawns - the outcome #2515 removed for the gateway call site, arriving through the vocabulary instead of through a missed one.

AGENTS.md already has the rule this breaks:

> **Warn on unrecognized values** - `STRANDS_ROBOT_MODE=foo` (typo) must `logger.warning(...)`, not silently fall through. Silent typo'd env vars surprise users hours later.

`off` is not an invented spelling. Two other variables in this tree accept the words this one ignores, so an operator can arrive at either without guessing:

| variable | vocabulary | site |
|---|---|---|
| `STRANDS_ROBOT_MESH_DC` | `("off", "0", "false", "no")` | `tools/robot_mesh.py:798` |
| `REACHY_DAEMON_TLS` | `("1", "true", "yes", "on")` | `device_connect/reachy_transport.py:69` |

## Root cause

The silence is structural, not an omission at either call site. The vocabulary was split across two modules and each half legitimately owns only its own words:

- `robot._mesh_env_opt_in` reads `("true", "1", "yes")`. To it, `false` is simply **not an opt-in** - not a mistake.
- `mesh.core.mesh_disabled_by_env` reads `("false", "0", "no")`. To it, `true` is simply **not a kill** - not a mistake.

Neither is in a position to call a *third* value wrong, because doing so requires knowing both halves to tell a typo from a legitimate spelling the reader does not own. That is why the fix is one owner rather than a warning added twice: a warning at either site would have to re-list the other site's vocabulary to avoid crying wolf on it, which re-creates the split it was meant to close.

## Change

`strands_robots/_mesh_switch.py` becomes the sole owner of the vocabulary and returns a tristate - affirmative / negative / said-nothing - with the report on the third case. Both readers delegate: `_mesh_env_opt_in` is `mesh_env_request() is True`, `mesh_disabled_by_env` is `mesh_env_request() is False`.

The tristate is what lets one resolver serve two readers. `None` for "said nothing" preserves the property both docstrings already state - that an unset variable answers False to both - which collapsing it into either bool would break.

It is modelled on `mesh/_backend_select.py`, the existing "Sole owner of the `STRANDS_MESH_BACKEND` vocabulary", written for the same failure (`STRANDS_MESH_BACKEND=iott` produced a plain Zenoh session indistinguishable from an explicit `zenoh`) and likewise reporting once per distinct offending value - once per value rather than once per call because every construction site of a `Mesh` consults the kill switch, so per-call would put one line per bring-up in the operator's log.

It lives at package top level rather than under `mesh/`, and imports nothing from `strands_robots`: `robot.py` reaches the mesh package only lazily, inside the function that starts a mesh, because `mesh/__init__.py` eagerly imports the Zenoh-backed session and core. A resolver under `mesh` would make that import eager at `Robot` import time and break the documented no-zenoh-installed path. A test pins that the owner imports nothing from the package.

## Scope

**Report only.** Making `off` *mean* `false` is a behaviour change on a safety switch and is deliberately not decided here - the warning names the spellings that work instead, which is what the rule above asks for. The consequence table above is therefore unchanged after this PR by design; what changes is that the third row class is no longer silent. Widening the vocabulary is a call for whoever owns the switch, and #2592 is left describing it.

## Tests

`tests/mesh/test_mesh_env_vocabulary_has_one_owner.py`, 34 tests:

- the four unrecognized values each warn exactly once, and the message names every working spelling
- reached through the two real resolvers, not the owner directly - the property that was missing
- once per distinct value across 25 calls; a second distinct typo is still news
- **controls**: every affirmative spelling, every negative spelling, unset, empty, whitespace-only and mixed case keep their verdicts *and* stay silent
- the tristate really has three outcomes
- **a guard against re-splitting**: an AST check that neither reader reads `STRANDS_MESH` itself, plus a check that `mesh.core` keeps no second copy of the negative half. Re-adding an inline spelling restores the exact condition that made the typo unreportable, and would do so without failing any behavioural test above
- non-vacuity: the unrecognized sample is disjoint from the vocabulary, `off` really is spelled by `robot_mesh.py`, and the two readers really do disagree about which half they own

## Verification

Measured at `8f791131` against `upstream/main` `52b2460f`, identical environment and identical selection - all of `tests/mesh`, plus `test_robot_factory.py`, `test_customer_workflow_regressions.py` and `test_all_exports_are_statically_defined.py`:

| this branch | 112 | 3497 | 81 | 24 |
| `upstream/main` | 112 | 3463 | 81 | 24 |

The failing ID sets are **identical** (`diff` empty); the `+34` passed are exactly the new tests. All 112 failures and 24 errors are missing optional extras in this sandbox (`awsiot`, `zenoh`, `torch`, `lerobot`), which CI installs.

The prose graders were run as their own selection (`tests/test_docs_*`, `test_changelog_*`, `test_log_strings_are_ascii`, `test_docstring*`). That reading was taken in a sandbox missing several optional extras and reported `8 failed / 238 passed` on both trees, attributing every failure to the absent extras. Re-run in an environment that carries them, the same selection is **`1 failed / 237 passed`** on this branch against **`238 passed`** on `upstream/main` - and the one failure is `test_qualified_strands_robots_xref_roles_resolve`, this branch's own offender rather than an environment artifact. A count-level comparison of two equal-sized failure sets cannot separate the two, which is how the red check reached review; the corrected numbers are below.

`ruff check` and `ruff format --check` clean over `strands_robots/` and `tests/mesh/`. `mypy` on the three changed source files reports the same **2 pre-existing** `mesh/core.py` errors as `main` (`_dispatch_sim_policy` `extra` arg-type, at lines 1850/1864 on main and 1851/1865 here - the shift is my one added import line) and none in the changed lines.

No recognized input changed behaviour: only the previously silent class gains a log line.

---

## Round 1 changelog

Three fixes from CI feedback, one concern each:

1. **`docs(readme)`** - `test_the_env_table_names_no_variable_it_never_lists` refused my README row: it cited `STRANDS_ROBOT_MESH_DC` to attribute the `off` spelling, but that variable has no row of its own, so the citation sent a reader looking for one that does not exist. The row now names the ignored spellings directly. The cross-reference stays in the resolver's module docstring and in #2592, which that rule does not govern - adding a row for `STRANDS_ROBOT_MESH_DC` would mean documenting a second variable inside a fix for this one.
2. **`refactor(mesh)`** - CodeQL flagged `_MESH_KILL_SWITCH_VALUES` as an unused global, and it was right: once `mesh_disabled_by_env` delegated, the name was an alias no package code read, which is also dead code under convention 10. Removed, and the two tests that parametrized over it now read the vocabulary from the module that owns it. The new guard asserts `mesh.core` keeps no second copy, because an unread copy is what a later edit could start reading again.
3. **Verification re-baselined.** The first table was measured before `pyserial` was installed in the sandbox, which made one `test_robot_factory.py` case look like a code difference when it was an environment one. Both trees were re-run in a single environment; the numbers above are the corrected pair, and the `mypy` claim is now stated against `main` rather than asserted clean.

---

## Round 2 changelog

Two prose fixes, one concern each. No source behaviour, no test, and no public
name changed; both commits touch comments inside `_mesh_switch.py` only.

1. **`docs(mesh)`** - `test_qualified_strands_robots_xref_roles_resolve` was the
   sole failing check. The `:func:` role for `mesh_disabled_by_env` sat
   mid-sentence in the module docstring and wrapped across the line break, and
   the grader refuses a wrapped path on its own terms: once the newline
   collapses to a space `strands_robots.mesh.core. mesh_disabled_by_env` imports
   nothing, and the path as written is what a reader copies. The role now sits
   on its own line, where re-flowing the surrounding sentence cannot push it
   back over the margin.

2. **`docs(mesh)`** - self-review caught stale prose that round 1 created.
   `NEGATIVE`'s comment still described itself as re-exported under
   `mesh.core._MESH_KILL_SWITCH_VALUES`, which is the name round 1 deleted.
   Because `test_core_keeps_no_second_copy_of_the_negative_half` now asserts
   that name is absent from `mesh.core`, the comment pointed a reader at a name
   this PR's own guard forbids from existing - worse than merely out of date,
   since it reads as an invitation to restore the alias. It now says where the
   tests read the tuple from, and that the absence in `core` is deliberate.

### Verification at `c4e5f17`

Fix 1 is checked against the grader's own helper rather than by re-running the
suite, because this sandbox lacks the extras the sweep imports (`cv2`), which is
an environment failure unrelated to the docstring:

| check | result |
| --- | --- |
| `_offending_roles_in(_mesh_switch.__doc__)` | `[]` |
| same helper on the pre-fix wrap (control) | reproduces the exact CI message |
| all four roles in the docstring | extracted contiguous and resolving |

Fix 2 cannot change behaviour, and the guards that read this file are AST-based,
not textual - re-checked directly: the owner still imports nothing from
`strands_robots` (`set()`), and `mesh.core` still contains no
`_MESH_KILL_SWITCH_VALUES`. The full tristate table was re-derived from the
pushed module - all three affirmative spellings `True`, all three negative
`False`, `off`/`on`/`ture`/`disabled`/empty/unset `None`, one warning per
distinct unrecognized value naming every working spelling - identical to the
table at the top of this description.

tests_integ`.

The suite has since been re-run in an environment carrying the extras the
earlier sandbox lacked, over 381 files - every guard that reads a docstring or
walks the package sources, plus all of `tests/mesh`:

| this branch (`c4e5f17`) | 21 | 14714 | 60 | 24 |
| reviewed head (`8f79113`) | 22 | 14713 | 60 | 24 |

Both collect 14735. The failed and passed deltas are each exactly `1`, the
branch-only failure set is **empty**, and the single reviewed-head-only failure
is `test_qualified_strands_robots_xref_roles_resolve` - the check that was red.
The 45 shared failures and all 24 errors are `tests/mesh/test_iot_*` needing
`awsiot`, a declared extra CI installs. `mypy` over `strands_robots tests
tests_integ` reports the same 24 pre-existing errors as before, none in the
changed file, which the diff makes structural: it contains no code lines at all.

Both commits land on top of the reviewed tree rather than rewriting it, so the
approval that predates them refers to a diff these two prose changes only add
to.


yinsong1986 (dismissed):
## Summary

Report-only fix for the silent third class of `STRANDS_MESH` values: the vocabulary that was split across `robot._mesh_env_opt_in` (affirmative half) and `mesh.core.mesh_disabled_by_env` (kill-switch half) now lives in one private owner, `strands_robots._mesh_switch.mesh_env_request()`, which returns a tristate and warns once per distinct unrecognized value, naming the spellings that work. Both readers delegate via `is True` / `is False`, so every recognized spelling, unset and empty keep their exact prior verdicts - the diff deliberately does not make `off` mean `false`, leaving that behaviour call to #2592. Verified against the head: `os` remains used in both modified files, `tests/conftest.py` sets the recognized `false` so the suite gains no warnings, `_MESH_KILL_SWITCH_VALUES` keeps the name the existing kill-switch regression test parametrizes over (now bound to the owner's `NEGATIVE`), and the new module imports nothing from `strands_robots`, preserving the lazy no-zenoh import path.

## What's good

- Follows the established `mesh/_backend_select.py` pattern exactly (once-per-value ledger, `.add`-only mutation, warning wording that names the valid vocabulary), per AGENTS.md > Env Vars > "The report belongs to the resolver that runs first".
- Scope discipline: no recognized input changes behaviour; the safety-switch semantics question is explicitly deferred to the issue rather than smuggled in.
- README Configuration row and changelog fragment (`changelog.d/2593-...`) updated in the same PR, per convention.
- The test module pins the delegation structurally (AST guard that neither reader reads the variable itself) plus non-vacuity checks, so a re-split fails a test even though it would pass every behavioural one.

yinsong1986 (dismissed):
## Summary

Report-only fix for the silent third class of `STRANDS_MESH` values. The vocabulary was split across `robot._mesh_env_opt_in` (affirmative half) and `mesh.core.mesh_disabled_by_env` (kill-switch half), so a value belonging to neither -- `off`, `on`, `disabled`, a transposition -- fell through both in silence and let an explicit `mesh=True` construct the mesh the operator meant to kill. The new private owner `strands_robots._mesh_switch.mesh_env_request()` holds both halves, returns a tristate, and warns once per distinct unrecognized value naming the spellings that work; both readers now delegate (`is True` / `is False`), so no recognized input, unset, or empty value changes behaviour. Verified at head: the two touched test suites (70 tests, including the 34 new ones) pass, `mesh_disabled_by_env` and `_mesh_env_opt_in` are provably equivalent to the pre-change predicates on every recognized/unset/empty input, `tests/conftest.py`'s `STRANDS_MESH=false` stays in the recognized-and-silent class, and the owner module imports only stdlib so the lazy no-zenoh import path is preserved (and pinned by an AST test). Scope discipline is right for a safety switch: whether `off` should *mean* `false` is explicitly deferred to #2592 rather than widened here.

## What's good

- The re-split guard (AST check that neither reader reads the variable itself, plus the no-second-copy check on `mesh.core`) pins the structural cause, not just the symptom -- re-adding an inline spelling would fail no behavioural test, and the guard catches exactly that.
- Once-per-distinct-value reporting follows the existing `mesh/_backend_select.py` pattern for the same per-bring-up-path reason, per AGENTS.md > Env Vars > "key the report by the offending value".
- Controls assert every recognized spelling, unset, empty, whitespace and mixed case keep their verdicts *and* stay silent, so the fix cannot have narrowed the vocabulary.
- Changelog fragment, README env-table row, and the AGENTS.md learning all land in the same PR, per the repo's documentation conventions.

One non-blocking note, mentioned here only because it needs no thread: the `#:` comment on `NEGATIVE` in `_mesh_switch.py` still says the tuple is "Re-exported as `strands_robots.mesh.core._MESH_KILL_SWITCH_VALUES`", but the head commit removed that re-export (the kill-switch test now imports `NEGATIVE` directly). Stale comment, zero behavioural impact -- fine to fix in any later pass.

yinsong1986 (approved):
## Summary

Report-only fix for the silent third class of `STRANDS_MESH` values. The vocabulary was split across two readers that each own one half -- `robot._mesh_env_opt_in` (affirmative) and `mesh.core.mesh_disabled_by_env` (kill switch) -- so a value belonging to neither (`off`, `on`, `disabled`, a transposition) fell through both in silence and let an explicit `mesh=True` construct the mesh the operator meant to kill. The new private owner `strands_robots._mesh_switch.mesh_env_request()` holds both halves, returns a tristate (`True`/`False`/`None`), and warns once per distinct unrecognized value naming every working spelling; both readers delegate via identity checks. Scope discipline is right: no recognized input changes behaviour, and whether `off` should *mean* `false` is correctly left to #2592.

## What's good

- Verified by direct execution at `c4e5f174`: all three affirmative spellings return `True`, all three negative return `False`, `off`/`on`/`ture`/`disabled`/empty/unset return `None`; the warning fires once per distinct value across repeated calls and names all six working spellings.
- The owner module imports only stdlib (`logging`, `os`), so the documented no-zenoh lazy-import path of `Robot` is preserved -- and the new test suite pins that structurally, along with an AST guard that neither reader re-reads the variable inline (the exact split that caused the silence).
- Both `tests/mesh/test_mesh_env_vocabulary_has_one_owner.py` (34 passed) and the updated `tests/mesh/test_gateway_mesh_kill_switch.py` (36 passed) pass locally; `ruff check` and `ruff format --check` clean on the changed files.
- AGENTS.md 'Warn on unrecognized values' / one-vocabulary-owner convention followed, README env-var row updated in the same PR, changelog fragment present per step 3, prior CodeQL finding (unused `_MESH_KILL_SWITCH_VALUES`) resolved by deletion rather than suppression.
- `_mesh_switch` is an underscore-private module and the env var name and semantics of every recognized spelling are unchanged, so there is no new public surface to regret later.

PR #663, opened 2026-06-25
feat(policies): add VERA video-to-action policy provider (#78)

cagataycali:
## Add VERA video-to-action policy provider (#78)

Adds **VERA** ([Video-to-Embodied Robot Action](https://github.com/sizhe-li/VERA),
MIT/CSAIL) as a first-class `strands_robots` policy provider — a two-stage
**DFoT/WAN video planner + Jacobian IDM**, served from an NVIDIA container over a
WebSocket protocol, mirroring the `cosmos3` service pattern.

> *One video planner, many IDMs* — the planner dreams future frames
> (embodiment-agnostic); a small per-embodiment Jacobian IDM translates the dream
> into robot actions.

### Rollouts (both PASS an automated Cosmos 3 QA critic)

| PushT (planar, task-space) | MimicGen → Panda (real arm via IK) |
|---|---|
| PushT | MimicGen Panda |

Both validated end-to-end on an L40S: the containerized DFoT/WAN planner + Jacobian
IDM produce action chunks, which `run_policy` applies in MuJoCo and records to MP4.

### Architecture — no VERA install on the host

```
host venv (numpy>=2)                 strands-vera-server container (torch 2.6 / CUDA 12.8)
  VeraPolicy ─ VeraWebsocketClient ─ws─▶ vera.server.start_vera_server
  (websockets + msgpack only)            DFoT/WAN planner + Jacobian IDM + AllTracker + /ckpts
```

The provider is an import-light msgpack+WebSocket client; VERA's heavy/conflicting
GPU stack (PyTorch 2.6, VGGT, DFoT, AllTracker) lives in the container. The host
needs only `websockets` + `msgpack`.

### What's included

- **`strands_robots/policies/vera/`**
  - `provider.py` — `VeraPolicy(Policy)`: rolling context window + action-chunk
    queue, resolution-aware multi-view frames, gripper binarization, and an
    **IK bridge** for `eef_delta`/`cartesian_delta` arms (auto-discovers the
    end-effector frame; seeds mink in full model qpos; `ik_smoothing` EMA).
  - `client.py` + `_msgpack_numpy.py` — self-contained wire client (no `vera` import).
  - `server_runner.py` — `VeraServerRunner` (subprocess) **and** `DockerServerRunner`
    (manages the container; list-arg `docker run`, no shell strings — per PR #621).
  - `config.py` — env-overridable `VeraConfig` (server flags 1:1, per-embodiment
    ports + render widths, `subprocess`|`docker` modes).
[...]

PR #317, opened 2026-06-04
feat(policy): NVIDIA Cosmos 3 omnimodal VLA policy provider

cagataycali:
Adds **Cosmos 3** as a first-class policy provider in `strands_robots.policies`,
so `create_policy("cosmos3", ...)` returns a `Policy` that maps robot
observations → action chunks using NVIDIA's `Cosmos3-Nano-Policy-DROID`
vision-language robot policy - served locally via **Cosmos Framework** (pure
Python).

> Cosmos 3 is an omnimodal world model. Its **Generator _action_ surface** (the
> `policy` mode: `image + instruction → action chunk + rollout video`) is, in
> effect, a VLA policy — a 1:1 fit for the robots `Policy` contract.
---

## 🧩 Where it fits

```mermaid
flowchart LR
    subgraph robots["strands_robots.policies"]
        direction TB
        BASE["Policy (ABC)"]:::base
        MOCK["MockPolicy"]:::existing
        GROOT["Gr00tPolicy<br/>(ZMQ service)"]:::existing
        LEROBOT["LerobotLocalPolicy<br/>(HF local)"]:::existing
        C3["Cosmos3Policy<br/>★ NEW (WS service)"]:::new
    end
    BASE --> MOCK
    BASE --> GROOT
    BASE --> LEROBOT
    BASE --> C3
    classDef base fill:#1f2937,stroke:#9ca3af,color:#fff;
    classDef existing fill:#374151,stroke:#6b7280,color:#e5e7eb;
    classDef new fill:#065f46,stroke:#10b981,color:#ecfdf5,stroke-width:3px;
```

`Cosmos3Policy` mirrors `Gr00tPolicy`'s **service mode** exactly: the GPU model
lives in a server process; the policy is a thin client. For Cosmos 3 that
server is the framework's ready-made **RoboLab OpenPI WebSocket policy server**.

---

## 🔌 Runtime architecture

```mermaid
flowchart TB
    subgraph sim["Simulation / Robot (CPU)"]
        RUNNER["PolicyRunner<br/>obs → act → step loop"]:::cpu
        POLICY["Cosmos3Policy"]:::new
        CLIENT["Cosmos3WebsocketClient<br/>(OpenPI msgpack+NumPy)"]:::new
    end
    subgraph gpu["Cosmos Framework server (GPU)"]
        SERVER["action_policy_server_robolab<br/>OpenPI WebsocketPolicyServer"]:::srv
[...]

PR #54, opened 2026-03-17
feat: add policy registry system with JSON-driven plugin resolution

cagataycali:
## Summary

Introduces a unified registry for policy providers and robot definitions, replacing hardcoded if/elif chains with a JSON-driven plugin system.

This is **PR 1 of 4** in the dev → main migration. All subsequent PRs (GR00T upgrade, LeRobot local, LeRobot async) depend on this registry.

### Registry System (`strands_robots/registry/`)
- **`loader.py`** — JSON loader with mtime-based hot-reload + duplicate alias/shorthand/URL validation
- **`policies.json`** — 2 providers: mock, groot
- **`robots.json`** — 36 robots across 7 categories (arm, bimanual, hand, humanoid, expressive, mobile, mobile_manip) with aliases
- **`policies.py`** — Resolve smart strings (HF model IDs, server URLs, shorthands) → provider + kwargs
- **`robots.py`** — Resolve robot names/aliases, list by category, sim/real filtering

### Updated Policy Abstraction (`strands_robots/policies/__init__.py`)
- Registry-based `create_policy()` — accepts provider names, HuggingFace IDs, or server URLs
- `register_policy()` for runtime custom providers
- `get_actions_sync()` convenience wrapper on Policy ABC
- Smooth sinusoidal `MockPolicy` (deterministic, better for testing)

### Shared Utilities (`strands_robots/policies/_utils.py`)
- `extract_pil_image()` — camera image extraction from observation dicts
- `detect_device()` — CUDA → MPS → CPU auto-detection
- `parse_numbers_from_text()` — VLM text output → numpy arrays
- `check_trust_remote_code()` — opt-in security gate

### Policy Resolver (`strands_robots/policy_resolver.py`)
- Thin wrapper for `resolve_policy_string()` — convenience import

## Tests

**34 tests passing** covering:
- Policy ABC contract
- MockPolicy (init, state keys, async/sync actions, sinusoidal determinism, auto key generation)
- Registry (provider config lookup, aliases, unknown handling, list providers)
- Smart resolution (shorthands, HF model IDs, server addresses, extra kwargs)
- Custom provider registration

## Design Decisions
- **JSON over Python** — Robot/policy definitions in JSON files for easy editing without code changes
- **mtime hot-reload** — Registry files re-read on modification without restart
- **No breaking changes** — Existing `create_policy('groot', ...)` still works
- **Plugin-ready** — New providers added by editing JSON + implementing the Policy ABC

cc @awsarron

PR #2594, opened 2026-08-22
docs(registry): the documented resolution ladder names the URL forms the registry ships

cagataycali:
## Problem

`resolve_policy`'s docstring opens with a five-stage `Resolution order:` block whose first rung enumerates the URL forms stage 1 recognises:

```
1. URL patterns (ws://, zmq://, grpc://, host:port)
```

Stage 1 has no vocabulary of its own. It matches the `url_patterns` entries providers declare in `policies.json`, so the forms it recognises are exactly the ones the shipped registry carries. That list is wrong in both directions:

- **`host:port` is not one of them.** All five shipped patterns require a scheme (`^zmq://`, `^grpc://`, `^cosmos3://`, `^vera://`, `^wss?://`), so a scheme-less address matches nothing at stage 1 and falls through all five stages to the fallback, where it is forwarded to `lerobot_local` as a checkpoint id.
- **`cosmos3://` and `vera://` are missing.** Both are shipped patterns the rung never names, so two transports a caller could have used are undiscoverable from the ladder.

The rung dates from #54 (2026-03-17). `cosmos3://` arrived in #317 (2026-06-08) and `vera://` in #663 (2026-06-25); neither was added to the list, and `host:port` has never been true of the shipped registry.

`policy_provider_error` carries the same claim, listing "a `host:port` pair" among the spellings it treats as resolvable and promising that "only a name no spelling can reach yields a reason".

## What a caller following the rung gets

Measured on a MuJoCo `so101` sim, with `STRANDS_TRUST_REMOTE_CODE=1` set so the trust gate is not the first thing hit:

| step | result |
| --- | --- |
| `policy_provider_error("localhost:5555")` | `None` - the preflight reports no reason |
| `run_policy(policy_provider="localhost:5555")` | **escapes** the documented `status=error` envelope |
| the raise | `ValueError: Could not determine policy type from 'localhost:5555'. config.json had no usable 'type', 'model_type', or recognized 'auto_map' entry` |

A checkpoint diagnosis for a network address. With the trust gate closed instead, the same string reports that `lerobot_local` "loads HuggingFace models with `trust_remote_code=True`" and offers `export STRANDS_TRUST_REMOTE_CODE=1` - so the remedy on offer for a hostname is to enable remote code execution.

Escaping the envelope is the specific harm `policy_provider_error` exists to prevent; its own docstring says a raise out of `run_policy` "escapes the `status=error` envelope those tools are documented to return". It returns `None` here because the string *does* resolve - as a checkpoint id.

## Change

Docstrings only; no code path moves.

- **`resolve_policy`** - the stage-1 rung now names the shipped schemes (`ws://`, `wss://`, `zmq://`, `grpc://`, `cosmos3://`, `vera://`). A new paragraph below the ladder states that stage 1's vocabulary *is* the registry's, and that a scheme-less `host:port` is matched only by a provider declaring a scheme-less pattern - none of the shipped ones does, so such a string reaches stage 5 and is forwarded as a checkpoint id rather than dialled as an address.
- **`policy_provider_error`** - drops `host:port` from the enumerating aside and says why it is not among them.
- **the generic scheme-less branch** keeps its behaviour and gains a comment: it is reached only when a provider declares a scheme-less `url_patterns` entry, so against the shipped registry it is an extension point rather than a live path.

Routing a bare `host:port` to a transport provider is deliberately **not** attempted: which provider owns the scheme-less form is a contract choice (`server_address` suggests gRPC, but zmq and ws are equally plausible), and the branch itself does not decide - it returns whichever provider declared the matching pattern.

The branch is a tested extension point, not dead code: `tests/registry/test_resolve_policy_url_schemes.py` exercises it by injecting a provider that declares `^[^/]+:[0-9]+$`. That file's module docstring also carried a stale census - "only declares `zmq://` and `cosmos3://`" against five shipped patterns, and listing `ws(s)://` and `grpc://` as lacking an exercising input when both are now shipped - corrected here.

## Guard

`tests/registry/test_resolution_order_names_the_shipped_url_forms.py` grades the rung against the registry in **both** directions using one relation - does some shipped `url_patterns` entry match a probe of this form:

- every form the rung names must be matched by a shipped pattern (soundness)
- every shipped pattern must match a form the rung names (completeness)

The same relation grades `policy_provider_error`'s aside, so a third surface cannot drift either. Only the enumerating constructs are graded - the numbered rung and the `--` aside - never the prose around them, since that prose exists to explain the forms stage 1 does *not* match; grading it would make an accurate docstring fail.

## Verification

Pre-fix split, the new guard copied onto `upstream/main`: **3 failed / 5 passed** -> 8 passed. Each failure names its offender:

```
resolve_policy's stage-1 rung names ['host:port'], which no shipped url_patterns
entry matches; such a string falls through to stage 5 and is forwarded to
lerobot_local as a checkpoint id.

the shipped registry declares url_patterns that resolve_policy's stage-1 rung
does not name: ['cosmos3 (^cosmos3://)', 'vera (^vera://)']

policy_provider_error's docstring names ['host:port'] among the spellings that
resolve, but no shipped url_patterns entry matches it
```

Every guard is load-bearing - six injected regressions, each failing exactly and only its own:

| --- | --- |
| rung names `host:port` again | soundness + completeness |
| rung omits `cosmos3://`, `vera://` | completeness |
| preflight aside names `host:port` again | preflight |
| generic scheme-less branch deleted | the extension-point control |
| the form parse matches nothing | completeness + non-vacuity |
| aside scoping widened to the whole docstring | preflight, on the accurate qualification |

The fifth row is why the non-vacuity test exists: with nothing parsed, soundness and the preflight rule pass vacuously.

Gate over 87 files (every test touching the resolver, the registry, or a docstring/string/changelog guard), the identical selection run on both trees with the new guard excluded, so any difference is the source change alone:

| tree | result |
| --- | --- |
| this branch (`47cd3d1c`) | 3 failed, 3245 passed, 1 skipped |
| pristine `upstream/main` (`bdfe49a`) | 3 failed, 3245 passed, 1 skipped |

The failing-ID sets are byte-identical - `comm` empty in both directions. All three are environment-dependent locally and unrelated to the diff: two are
`tests/training/test_lerobot.py::TestInProcessTrainerCorrectness` on a `draccus` older than lerobot's declared `>=0.11.6` floor, and one is `test_vla_jepa.py`'s lerobot lazy step registration, which fails identically on a pristine tree in isolation.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 against a pristine worktree with no branch-only error.

## Artifact

Each column reads **its own tree's** rung, extracts the forms it names, and follows them. Package code is identical in both columns - only the docstring differs - so the console block at the bottom, which is what a caller who writes the scheme-less form actually gets, is byte-identical in both trees. Asserted in the figure script alongside every headline number.

documented stage-1 URL vocabulary vs the shipped registry

### DIFF
--- a/changelog.d/2594-resolution-order-names-shipped-url-forms.md
+++ b/changelog.d/2594-resolution-order-names-shipped-url-forms.md
@@ -1,0 +1,17 @@
+### Fixed
+
+- **registry**: `resolve_policy`'s documented stage-1 rung now names the URL forms
+  the shipped registry actually matches. Stage 1 has no vocabulary of its own - it
+  matches the `url_patterns` providers declare in `policies.json` - and the rung was
+  wrong in both directions: it listed a scheme-less `host:port`, which no shipped
+  pattern matches (all five require a scheme), and omitted `cosmos3://` and
+  `vera://`, which they do. A caller following the rung with `localhost:5555` got no
+  reason from `policy_provider_error`, then a raise out of `run_policy` diagnosing a
+  broken HuggingFace checkpoint - `config.json had no usable 'type'` - for a network
+  address, having reached stage 5 and been forwarded to `lerobot_local` as a
+  checkpoint id. `policy_provider_error`'s own enumeration of resolvable spellings
+  carried the same claim and is corrected too. The generic scheme-less branch keeps
+  its behaviour: it is a tested extension point, reached when a provider declares a
+  scheme-less pattern, and which provider should own the form is a contract choice
+  left open. A new bidirectional guard grades both surfaces against the registry, so
+  a scheme added to one cannot drift from the other.
--- a/strands_robots/policies/factory.py
+++ b/strands_robots/policies/factory.py
@@ -313,9 +313,11 @@ def policy_provider_error(provider: str, **kwargs) -> str | None:
 
     Probes the SAME resolution path :func:`create_policy` uses, without
     instantiating anything, so every spelling that provider accepts -- a
-    registered name, a HuggingFace model ID, a ``zmq://`` / ``ws://`` URL, a
-    ``host:port`` pair -- resolves here too. Only a name no spelling can reach
-    yields a reason.
+    registered name, a HuggingFace model ID, a ``zmq://`` / ``ws://`` URL --
+    resolves here too. Only a name no spelling can reach yields a reason. A
+    scheme-less ``host:port`` is not among them: no shipped provider declares a
+    scheme-less ``url_patterns`` entry, so such a string is resolvable only as a
+    checkpoint id and this preflight reports no reason for it.
 
     This is the agent-tool companion to :func:`preflight_policy`, which
     deliberately swallows resolution failures on the stated grounds that
--- a/strands_robots/registry/policies.py
+++ b/strands_robots/registry/policies.py
@@ -131,11 +131,21 @@ def resolve_policy(policy: str, **extra_kwargs) -> tuple[str, dict[str, Any]]:
     and returns the canonical provider + ready-to-use kwargs.
 
     Resolution order:
-        1. URL patterns (ws://, zmq://, grpc://, host:port)
+        1. URL patterns declared in ``policies.json`` (ws://, wss://, zmq://,
+           grpc://, cosmos3://, vera://)
         2. Shorthand names (mock, groot, lerobot_local, ...)
         3. HuggingFace model IDs (org/model)
         4. Registered provider name
         5. Fallback to lerobot_local
+
+    Stage 1 recognises exactly the forms the registry declares: the
+    ``url_patterns`` entries providers carry are the whole vocabulary, so a
+    scheme this package does not ship is not a URL here. A scheme-less
+    ``host:port`` address is matched only by a provider that declares a
+    scheme-less pattern -- the generic ``server_address`` branch below exists
+    for that -- and none of the shipped providers declares one, so with the
+    shipped registry such a string reaches stage 5 and is forwarded to
+    ``lerobot_local`` as a checkpoint id rather than dialled as an address.
 
     Every stage matches case-insensitively. A URL scheme is folded per RFC 3986
     section 3.1 (``ZMQ://gpu:5555`` resolves exactly as ``zmq://gpu:5555``, and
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Docs-only correction of `resolve_policy`'s stage-1 rung and `policy_provider_error`'s enumerating aside, plus a bidirectional guard test that grades both docstrings against the shipped `url_patterns` in `policies.json`. Verified against the head: the shipped registry declares exactly the five patterns the PR claims (`^zmq://`, `^grpc://`, `^cosmos3://`, `^vera://`, `^wss?://`), so the old rung was wrong in both directions (`host:port` never shipped; `cosmos3://` and `vera://` were omitted). No behavior moves — the only change under `strands_robots/` is docstring text and a comment on the untouched generic scheme-less branch, and the diff introduces no non-ASCII, no host paths, and no new API surface. The follow-up commit adds the `changelog.d/2594-*.md` fragment per the AGENTS.md step-3 convention, and it validates against the assembler (`assemble_changelog.py --check` passes). The new test file (36 tests across the two registry test modules) passes on the head, `ruff check` and `ruff format --check` are clean on all four changed files, and the guard is scoped to the enumerating constructs only, so accurate qualifying prose cannot fail it.

## What's good

- The guard reads the rung and the registry through one relation (pattern-matches-probe), so soundness and completeness cannot disagree, and it covers the third surface (`policy_provider_error`'s aside) with the same relation.
- The stale census in `test_resolve_policy_url_schemes.py`'s module docstring was corrected in the same pass rather than left to drift further.
- The scheme-less branch is documented as an extension point and kept exercised via an injected provider, rather than deleted or silently rerouted — routing a bare `host:port` to a transport would have been a contract choice this PR correctly declines to make.

yinsong1986 (approved):
## Summary

Docs-only correction plus a bidirectional guard: `resolve_policy`'s stage-1 rung now enumerates the six scheme forms the shipped `url_patterns` actually match (`ws://`, `wss://`, `zmq://`, `grpc://`, `cosmos3://`, `vera://`), dropping the never-true `host:port` and adding the two schemes that arrived after the rung was written; `policy_provider_error`'s enumerating aside is corrected on the same claim; the generic scheme-less branch gains a comment marking it an extension point; and `tests/registry/test_resolution_order_names_the_shipped_url_forms.py` grades both docstrings against the registry in both directions so the three surfaces cannot drift again. Verified against the head: `policies.json` declares exactly the five patterns claimed (`^zmq://`, `^grpc://`, `^cosmos3://`, `^vera://`, `^wss?://`), all scheme-anchored, so the corrected rung is accurate in both directions; no code path moves (the source-file hunks are docstring and comment only, confirmed by inspection); the pre-existing `Examples::` block in the same docstring stays consistent with the corrected rung (`grpc://gpu-box:8080`, with scheme, is what routes to `lerobot_async`); the new guard and the whole `tests/registry/` selection pass on the head (795 passed); `ruff check` and `ruff format --check` are clean on all four touched Python files; the changelog fragment follows the `changelog.d/<pr-number>-<slug>.md` convention from AGENTS.md step 3; and the new content is ASCII-only with no host paths.

## What's good

- The guard grades the enumerating constructs only (the numbered rung and the `--` aside), never the surrounding prose, so an accurate qualification like "a scheme-less `host:port` is not among them" cannot trip it - and the non-vacuity test keeps the parse from passing by matching nothing.
- The stale census in `test_resolve_policy_url_schemes.py`'s module docstring was corrected in the same pass rather than left to drift a fourth surface.
- Routing a bare `host:port` to a transport was deliberately not attempted, keeping the change zero-risk to behaviour while documenting why the extension point stays open.

PR #2595, opened 2026-08-22
fix(policies): read one observation as one state vector in every provider

cagataycali:
## Problem

A `Policy` whose caller declared no `robot_state_keys` has to infer this step's
joint-state ordering, and the only ordering available is the observation's own
insertion order. Every sim backend writes a velocity companion beside each joint
position - `simulation/mujoco/rendering.py` writes `obs[jnt_name] = qpos` then
`obs[f"{jnt_name}.vel"] = qvel`, calling the second an "Additive key ... existing
position-only" consumer contract - so that order alternates
`[pos0, vel0, pos1, vel1, ...]`.

The LeRobot provider dropped the paired velocity siblings before using that
ordering, and said why: taking the keys in observation order "puts a velocity in
every other slot of `observation.state` and, once truncated to the model's state
dim, drops the trailing joints entirely - a wrong state vector with no error
raised". The Cosmos 3 provider inferred its ordering the same way and did not
filter, so on a real 7-DOF arm rollout its `joint_pos` request carried:

| reading | vector |
| --- | --- |
| the arm's actual joints | `[+0.0677, +0.1379, +0.3685, -0.0598, +0.2481, +0.4527, +0.0668]` |
| the request, before | `[+0.0677, +1.7042, +0.1379, +3.1892, +0.3685, +2.2692, -0.0598]` |

Three of the seven slots hold rad/s readings, joints 5-7 are never sent, worst
joint error **3.2491 rad (186 deg)**, and the request is reported as sent.

## Change

The rule now has one owner outside both providers
(`strands_robots/policies/_state_keys.py`, stdlib-only, the shape
`policies/_rng.py` already uses for a contract two providers must agree on), and
both consume it. Two providers conforming to the same `Policy` contract can no
longer read one observation as two different state vectors.

What did not change: a declared `robot_state_keys` ordering is still never
filtered - naming `elbow.vel` there states the model's input - and pairing is
decided per key, so an unpaired `.vel` survives (LeKiwi declares body-frame base
velocities `x.vel` / `y.vel` / `theta.vel` with no position companion).

`docs/policies/cosmos3.md` now states the rule where `docs/policies/lerobot-local.md`
already did, since it is one rule.

## Verification

Rollout in MuJoCo (headless). One capture script per tree drives the same
scripted panda sweep, takes the tick of peak joint speed, asks the real
`_attach_joint_state` for the request with no declared ordering, then writes those
seven values back through `set_joint_positions` and renders - so each panel is the
pose the request describes.

one observation, one state vector

Panels 1 and 3 differ by **0.00014** mean abs level (the request now describes the
arm); the middle panel differs from the arm by **10.506**, with 13.5% of pixels
moved. The control rows are byte-equal across trees: the rollout
(`0.00015`) and a **declared** ordering, which was correct before and after.

Pre-fix split, both test files copied onto pristine `upstream/main`:
**10 failed / 36 passed** -> 46 passed. Five are behavioural or structural, each
naming the finding:

- the rule is defined once and in the shared leaf -> `['policies/lerobot_local/policy.py']`
- every inferred ordering routes through the rule -> `{'policies/cosmos3/policy.py:509': '[k for k, v in robot_obs.items() if np.isscalar(v) or np.ndim(v) == 0]'}`
- cosmos3 infers positions only -> `[0.1, -100.0, 0.2, -100.0, 0.3, -100.0, ...]`
- both providers read one observation the same way -> the two orderings, side by side
- the end-to-end `joint_pos` request test in the cosmos3 suite

The other five can only report the shared module's absence pre-fix; they pin the
rule's own contract rather than the regression.

Each guard was checked by injecting the regression it exists for:

| --- | --- |
| revert the cosmos3 fallback filter | routing + cosmos3 behavioural + parity + end-to-end (4) |
| grade every `robot_state_keys or ...`, not just an inferred ordering | routing + the membership-is-not-an-ordering control (2) |
| match the rule name as a substring | the provider-local-copy control (1) |
| filter `.vel` by suffix alone | unpaired-kept + declared-ordering controls (2) |
| re-derive the rule inside the provider | single-owner + routing (2) |
| make the structural scan reach nothing | single-owner + non-vacuity (2) |

Gate: 414 test files touching the changed symbols, both policy suites, and the
repo-wide docs/docstring/string guards, run on this branch and on pristine
`upstream/main` with the new file excluded from both. Both report **22 failed / 16623 (branch) and 16621 (pristine) passed / 87 skipped**, and the failing-ID sets are byte-identical, so the branch-only set is empty. The `+2` collected reconciles exactly: the new end-to-end test, and the repo-wide teardown-import guard that auto-enrolls the new module. The 22 are pre-existing here and dependency-absent or known: 17 `tests/mesh/test_iot_*` (`awsiot`), 4 import-order cases in `test_recording_frame_loss_is_not_tolerated.py`, and one lerobot lazy processor-step registration in `test_vla_jepa.py`.
`ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ`
24 == 24 against pristine, none in the changed files.


yinsong1986 (approved):
## Summary

Unifies the observation-derived joint-state ordering rule across policy providers: the `.vel`-sibling filter the LeRobot provider already applied to an inferred ordering is lifted into a shared stdlib-only leaf (`strands_robots/policies/_state_keys.py`) and the Cosmos 3 provider now routes its inferred fallback through it, closing a real correctness bug where a sim-shaped observation put rad/s readings in joint slots of a 7-DOF `joint_pos` request and silently truncated the trailing joints. Declared `robot_state_keys` orderings are never filtered (unchanged), and pairing is per-key so unpaired velocity state (LeKiwi base velocities) survives - both invariants are pinned by tests.

## What's good

- Follows AGENTS.md convention #11 exactly: the guard was lifted to a shared module only once it had a second caller, and the old provider-local copy is fully removed (no dangling `_VELOCITY_SUFFIX` / `_drop_velocity_siblings` references anywhere in the tree).
- The new module is underscore-private and not exported in `__all__`, so no new public API surface is committed to.
- The structural test derives its graded set from the AST (`robot_state_keys or <inferred ordering>` fallbacks) rather than a hardcoded list, so a third provider written the same way is held to the rule on arrival - the same pattern the repo's other structural guards use.
- Regression coverage is behavioural (end-to-end `joint_pos` request assertion with sentinel velocity values) plus controls pinning the tolerated spellings (unpaired `.vel` kept, declared ordering untouched).
- Changelog fragment in `changelog.d/`, ASCII-only added lines, docs updated in both providers' pages for the one shared rule.

PR #1726, opened 2026-07-29
fix(policies): the state fallback does not interleave velocity siblings

cagataycali:
## Problem

`LerobotLocalPolicy._resolve_state_order` falls back to the observation's own scalar keys when none of the configured `robot_state_keys` are present in it - the generic `joint_0..joint_N` vs named-joint mismatch. It took those keys in the observation's **insertion order**, and the MuJoCo backend emits a velocity sibling beside every joint position (`strands_robots/simulation/mujoco/rendering.py`):

```python
obs[jnt_name] = float(data.qpos[model.jnt_qposadr[jnt_id]])
# Additive key (``<name>.vel``) - existing position-only
# consumers (dataset recording, arm policies) are unaffected;
obs[f"{jnt_name}.vel"] = float(data.qvel[model.jnt_dofadr[jnt_id]])
```

So the resolved order alternates `[pos0, vel0, pos1, vel1, ...]`: twice the DOF count, then truncated to the model's declared `observation.state` dim. Half the slots held velocities and the trailing joints were dropped entirely.

Measured on a 6-DoF arm driven to distinct known values, 6-dim model:

```
resolved order        ['shoulder_pan', 'shoulder_pan.vel', 'shoulder_lift', 'shoulder_lift.vel', 'elbow_flex', 'elbow_flex.vel']
fed to the model      [0.2981, -0.4700, -0.1973,  0.6561,  0.4978, -0.5439]
CORRECT positions     [0.2981, -0.1973,  0.4978,  0.0987, -0.3974,  0.2474]

wrong slots           5 of 6
joints absent entirely wrist_flex, wrist_roll, gripper
```

Nothing raised and nothing logged about it: the policy ran on a wrong state vector while reporting success. The producer's own comment above documents `<name>.vel` as an **additive** key that position-only consumers are unaffected by - this consumer was the exception to a contract stated in the code.

## Change

An ordering **derived from the observation** now drops each `<joint>.vel` whose `<joint>` position companion is present, via a new `_drop_velocity_siblings` helper applied at the two points in `_resolve_state_order` that return an observation-derived ordering (the all-missing fallback, and the no-`robot_state_keys` early return).

Three things the rule deliberately does **not** do:

[...]

PR #1099, opened 2026-07-05
fix(lerobot_local): keep observation.state index-aligned on a partial robot_state_keys mismatch

cagataycali:
## Problem

The `lerobot_local` provider builds `observation.state` by resolving `robot_state_keys` against the per-step observation. `_resolve_state_order` already makes the *all*-missing case loud (generic `joint_0..N` vs named joints, #897), but the **partial**-missing case was still silent: both state-build paths (`_to_lerobot_observation` and `_build_batch_from_strands_format`) iterated the resolved key order and appended **only** the keys found in the observation. An absent key was therefore *dropped*, shifting every following joint value up one index before the trailing zero-pad.

The canonical trigger is a robot whose `robot_state_keys` are keyed by actuator names, where a mimic/tendon gripper's actuator (`left/gripper` / `right/gripper` on aloha) is **not** among the observation's finger-joint names while the arm joints are. The policy then receives a garbage `observation.state` while the run reports `success`.

### Reproduced on a real MuJoCo aloha sim

`sim.robot_action_keys('aloha')` = 14 actuator names with the gripper actuators interspersed at indices 6 and 13, but `get_observation()` exposes `left/left_finger`, `left/right_finger`, ... (finger *joints*) and **not** `left/gripper` / `right/gripper`. Building the 14-dim state from that observation (joints perturbed to distinct values so the shift is visible):

```
idx  key                    PRE (drop+shift)   POST (in-place)
 0  left/waist                 0.100              0.100
 1  left/shoulder              0.110              0.110
 ...
 5  left/wrist_rotate          0.150              0.150
 6  left/gripper               0.180  <-- r-arm    0.000   (gripper slot, missing -> 0)
 7  right/waist                0.190  <-- shifted   0.180
 8  right/shoulder             0.200  <-- shifted   0.190
 9  right/elbow                0.210  <-- shifted   0.200
10  right/forearm_roll         0.220  <-- shifted   0.210
11  right/wrist_angle          0.230  <-- shifted   0.220
12  right/wrist_rotate         0.000  <-- dropped   0.230
13  right/gripper              0.000              0.000
```

Pre-fix, **8 of the 14 state dims are wrong** — the entire right arm slides into the wrong slots and both gripper dims land wrong — silently.

## Change

A shared `_collect_state_values(observation_dict, order)` helper (used by both state-build paths) now zero-fills a missing resolved key **in place** so present joints keep their model index, and surfaces the degradation instead of leaving it silent:

[...]

PR #897, opened 2026-06-30
fix(lerobot_local): make state-key mismatch loud instead of a silent open-loop rollout

cagataycali:
## Problem

`LerobotLocalPolicy` auto-fills `robot_state_keys` with generic names (`joint_0..N`, derived from the model action dim) when no `embodiment=` and no `set_robot_state_keys()` are given. A sim/robot instead reports **named** joints (`shoulder_pan`, `shoulder_lift`, ...), so none of the configured keys match the observation. Both observation-to-batch paths handled that mismatch silently:

- `_to_lerobot_observation` (preprocessor/VLA path, e.g. MolmoAct2) fell back to the observation's scalar keys with **no warning**, and
- `_build_batch_from_strands_format` (strands-native ACT/diffusion path) **dropped `observation.state` entirely**.

The model then ran conditioned on a zero/missing state - effectively open-loop - while reporting `action_errors=0`, `success_rate=0`, and **no error log**. The robot moves but never accomplishes the task, with nothing pointing at the misconfiguration.

## Change

Both state-building paths now route through a shared `_resolve_state_order(observation_dict, scalar_keys)` that honors the existing `strict_keys` knob:

- **`strict_keys=True`** -> raise a `ValueError` that names the actual observation keys and points at the two remedies (`embodiment='<name>'` / `set_robot_state_keys([...])`).
- **`strict_keys=False`** (default) -> log a single `WARNING` with the same guidance, then fall back to the observation's own scalar keys, so the state is **populated** rather than silently zeroed or dropped.

This closes the remaining silent surface: the strands-native path used to drop `observation.state` with no fallback at all, and the VLA path's fallback was completely silent.

Example message (both paths):

```
None of the configured robot_state_keys ['joint_0', 'joint_1', ...] are present
in the observation. Observed joint/state keys: ['shoulder_pan', 'shoulder_lift',
'elbow_flex', 'wrist_flex', 'wrist_roll', 'gripper']. This usually means generic
auto-generated keys (joint_0..joint_N) were paired with a robot/sim that reports
named joints. Pass embodiment='<name>' (e.g. embodiment='so101') or call
set_robot_state_keys([...]) with the robot's actual joint names.
```

## Tests

New `tests/policies/lerobot_local/test_state_key_mismatch.py` pins both paths in both modes (the 5 mismatch cases fail on pre-fix code, verified by reverting `policy.py`):

[...]

PR #614, opened 2026-06-23
feat(policies): add strict_keys to Gr00tPolicy + LerobotLocalPolicy

cagataycali:
## Problem

When a policy's declared keys don't exactly match the robot/sim keys, `Gr00tPolicy` and `LerobotLocalPolicy` silently fall back to positional `zip` matching with only a log line. On a multi-camera / multi-DOF rig this can map the wrong camera to a feature or the wrong action column to a joint -- a silent, hard-to-debug correctness bug. For production rollouts an opt-in strict mode that raises is safer than guessing.

## Change

Add a `strict_keys: bool = False` constructor kwarg to both policies, threaded through their mapping/camera-resolution helpers:

- `strict_keys=False` (default): unchanged positional-fallback-with-warning -- preserves zero-config ergonomics, byte-identical behavior.
- `strict_keys=True`: when exact-name matching leaves any key unresolved, raise a `ValueError` instead of positional-guessing. The message lists the unmatched robot keys vs the available model keys, mirroring the existing `ObservationMapping.validate` / `ActionMapping.validate` errors.

**GR00T** (`policies/groot/policy.py`): threads `strict_keys` through `_match_keys` (observation video/state) and `_auto_infer_action_mapping`, wired from `Gr00tPolicy.__init__` via `_init_mappings`. Ignored when an explicit `observation_mapping`/`action_mapping` is supplied.

**LeRobot Local** (`policies/lerobot_local/policy.py`): applies on both unresolved-camera paths -- `_resolve_camera_targets` (the documented camera-routing fallback) and the legacy `_to_lerobot_observation` remap, which had the same positional fill. `camera_key_map` and exact-name matches still satisfy strict mode.

`strict_keys` is added to `config_keys` for both providers in `registry/policies.json`, so `create_policy("groot", strict_keys=True)` / `create_policy("lerobot_local", strict_keys=True)` resolve.

Example error:
```
strict_keys=True: cannot resolve video keys by exact name. Unmatched robot keys: ['webcam']; available model keys: ['ego_view_bg_crop_pad_res256_freq20']. Provide an explicit mapping (observation_mapping/action_mapping) or set strict_keys=False to allow positional fallback.
```

## Tests

New `TestStrictKeys` (groot) and `TestStrictKeysCameraRouting` (lerobot_local), 13 cases total, covering:
- strict raises on a deliberate name mismatch, with unmatched-vs-available keys in the message;
- non-strict default still maps positionally + warns (behavior unchanged);
[...]

PR #2596, opened 2026-08-22
docs(policies): describe strict_keys as the whole key binding it gates

cagataycali:
## Problem

`strict_keys` is a public constructor parameter of `LerobotLocalPolicy`. Its
`Args:` entry defined it as a camera concern:

> When True, raise (instead of warning + positional fallback) if any **camera name**
> cannot be matched to a declared policy image key by exact name and no
> `camera_key_map` covers it.

The flag gates **four** refusals, and only two are on the camera path:

| gate | subject | refuses when | its docstring named the flag |
| --- | --- | --- | --- |
| `_resolve_state_order` | `robot_state_keys` | no configured key matches the observation | yes |
| `_collect_state_values` | `robot_state_keys` | some configured keys are missing | yes |
| `_to_lerobot_observation` | `camera_key_map` | a camera cannot be bound by name | no |
| `_resolve_camera_targets` | `camera_key_map` | a camera cannot be bound by name | no |

The entry was accurate when written. `strict_keys` shipped in #614 (2026-06-23)
with exactly two gates, both `unmatched_imgs and free_feats` / `unmatched and
free` - camera only. A joint-state refusal was gated on the same flag seven days
later (#897, 2026-06-30) and a second twelve days after that (#1099,
2026-07-05). The docs page caught up in #1726; the entry never did.

**Measured** on a real MuJoCo aloha observation (16 named joints, 13 cameras),
with `robot_state_keys` generically auto-filled - the case #897 gated:

| flag | cameras (documented) | joint state (was not) |
| --- | --- | --- |
| `strict_keys=False` | bound by name | bound, 16 values (1 warning) |
| `strict_keys=True` | bound by name | **REFUSED** - `ValueError` |

So a caller who read the entry, saw every camera bind by exact name - where the
flag is genuinely a no-op - and set `strict_keys=True` turned a joint-state
warning into a hard failure, from a parameter whose own definition excluded the
joint-state path.

## Why the entry is what is wrong

The sibling that got the flag in the same pull request is the counter-evidence.
`Gr00tPolicy`'s entry says "if auto-inferred **observation/action** keys cannot be
matched", naming the whole key surface, and is still accurate. Both joint-state
methods document their own `strict_keys` condition - `_resolve_state_order` ships
an explicit two-bullet list of the True and False branches - and
`docs/policies/lerobot-local.md` documents the joint-state raise in its body
("warns (or raises under `strict_keys=True`)"). The code, the methods, that page's
body and the sibling class all agree the flag governs both halves; only the flag's
own definition and that page's one-line gloss did not, and the page therefore
contradicted itself.

## Change

Documentation only.

- The `strict_keys` entry now names both halves, and qualifies its no-op claim as
  applying to a robot whose cameras *and joints* all bind by name.
- `_resolve_camera_targets` described its third routing rung as falling back with
  a WARN, which is what happens only when the flag is unset, and its `Raises:`
  enumerated two of the three conditions it raises `ValueError` for. Both now
  include the flag.
- `_to_lerobot_observation` described an unconditional positional fill; it now
  says the flag raises there instead.
- The docs page's one-line gloss matches its own body.

`docs/policies/camera-naming.md` is left alone deliberately: its sentence
("set `strict_keys=True` to turn an unresolved camera name into a hard error") is
a true camera-scoped statement on a camera-scoped page, not a definition of the
flag.

No behaviour change: no refusal, exception class, message or gate condition is
touched.

## Guard

Both rules derive what they grade from the tree, so neither carries a list:

- **every gate documents the flag** - a method whose behaviour the flag changes
  must name it in its own docstring, because that docstring is where a caller
  learns what the method does.
- **the flag's entry names every surface it governs** - for each gating method,
  the public attribute it reads alongside the flag is the surface that gate is
  about (`camera_key_map` for the camera gates, `robot_state_keys` for the
  joint-state ones), and every such attribute must appear in the entry.

The class set is derived as well: any class assigning `self.strict_keys` is
graded, so a second class taking the flag is held to the same rule on arrival.
The constructor's own `self.strict_keys = strict_keys` is a write, not a gate, and
is excluded by requiring a `Load` context.

## Tests

Pre-fix split, the new file copied onto `upstream/main`: **5 failed / 8 passed**,
every failure semantic and quoting the real entry text -

```
_resolve_camera_targets behaves differently under strict_keys (it reads
self.strict_keys and self.camera_key_map) but its docstring never names the flag
_resolve_state_order refuses on ['robot_state_keys'] under strict_keys, but the
strict_keys Args: entry never names ['robot_state_keys']
```

Each guard was verified load-bearing by injecting the regression it exists for:

| --- | --- |
| revert the entry to camera-only | entry rule for both state gates + the both-halves pin |
| remove the flag from `_resolve_camera_targets` | gate rule for that method |
| remove the flag from `_to_lerobot_observation` | gate rule for that method |
| gate scan counts a `Store` (so the constructor reads as a gate) | gate rule for `__init__` + entry rule x5 |
| carrier scan reaches no file | non-vacuity |
| no-op claim made without qualifying the state gates | the no-op qualification pin |

## Verification

Measured at `ba849425`, with `upstream/main` merged in (the base moved to
`71735507`; the true file overlap was `policy.py` alone, from #2595, and the
auto-merge was verified by symbol - all five of this change's markers present,
and #2595's `drop_velocity_siblings` promotion intact). Selection: 288 files - all of
`tests/policies/lerobot_local`, every test naming the flag, `robot_state_keys`,
`camera_key_map`, `LerobotLocalPolicy` or the docs page, and every guard that
grades docstrings, source strings or changelog fragments. Run with the new file
**excluded from both trees**, so any difference is caused by the source change
alone:

| tree | result |
| --- | --- |
| this branch | 5 failed, 9733 passed, 31 skipped in 453.25s |
| pristine `upstream/main` | 5 failed, 9733 passed, 31 skipped in 453.15s |

Failing-ID sets byte-identical; `comm` empty in both directions. The 5 are
pre-existing and were confirmed on the pristine tree: the four in
`test_recording_frame_loss_is_not_tolerated.py` pass 17/17 when that file runs
alone (import-order artifacts of a wide `tests/simulation` selection), and
`test_vla_jepa.py::test_vla_jepa_processor_steps_register_via_strands_path` fails
in isolation on pristine `upstream/main` too (a lerobot processor-step lazy
registration artifact of this environment).

tests_integ`; `mypy` 24 == 24 against a third pristine worktree at the same
commit, with an empty branch-only set and none in the changed files. The new file
alone: 13 passed, and the pre-fix split re-measured against the moved base is
unchanged at 5 failed / 8 passed.

## Artifact

Same package code in both columns - only the flag's own entry differs. The
capture script reads its own tree's `strict_keys` entry, derives the surfaces a
caller can conclude the flag governs, and then drives both refusals against a
real MuJoCo aloha observation (headless, EGL). The measured half is
**byte-identical across trees**, which is the control; the render differs by a
mean absolute 1.5e-05 (the same scene).

strict_keys entry vs the surfaces it gates


yinsong1986 (approved):
## Summary

Documentation-only correction: `LerobotLocalPolicy`'s `strict_keys` `Args:` entry described the flag as camera-scoped, while the flag has gated two joint-state refusals since #897/#1099. The entry now names both halves (`camera_key_map` and `robot_state_keys` surfaces), the two gating methods that never named the flag (`_resolve_camera_targets`, `_to_lerobot_observation`) now do, and the docs page's one-line gloss matches its own body. I verified every new docstring claim against the code at this head: `_resolve_state_order` (all-missing raise), `_collect_state_values` (partial-missing raise), and both camera gates (`strict_keys and unmatched and free`) behave exactly as the updated prose says, and no changed line in `policy.py` is outside a docstring - no refusal, exception class, message or gate condition is touched. The new guard test derives both its rules and the graded class set from the AST rather than a list (a `Load`-context read of `self.strict_keys` defines a gate, excluding the constructor's write), so a future gate or a second carrier class is graded on arrival; it runs 13/13 green at this head, ruff check/format are clean on both changed Python files, all additions are ASCII, no host paths, and the changelog fragment is correctly named for the PR per AGENTS.md step 3.

## What's good

- The no-op claim in the entry is correctly qualified ("cameras *and joints* all bind by name"), and the test pins that qualification separately from the both-halves rule, so the two cannot regress together silently.
- `docs/policies/camera-naming.md` was deliberately left alone with the reasoning stated - a true camera-scoped sentence on a camera-scoped page is not a definition of the flag. Good scope discipline.
- The guard's non-vacuity class (`TestTheScanIsNonVacuous`) means a refactor that empties the scan fails loudly instead of passing on nothing.

yinsong1986 (approved):
## Summary

Documentation-only correction of `LerobotLocalPolicy.strict_keys`'s `Args:` entry, which described the flag as camera-scoped while it has gated two joint-state refusals since #897/#1099. The entry now names both halves of the key binding (`camera_key_map` and `robot_state_keys` surfaces), the two gating methods that never mentioned the flag (`_resolve_camera_targets`, `_to_lerobot_observation`) now document its effect, and the docs page's one-line gloss now agrees with its own body. Verified against the head commit that all four methods reading `self.strict_keys` name the flag in their docstrings and that no refusal, exception class, or gate condition changed. The new guard derives both rules (every gate documents the flag; the entry names every gated surface) and the carrier class set from the AST rather than a hard-coded list, includes non-vacuity checks, and the PR description documents a load-bearing verification for each guard. Changelog fragment present per AGENTS.md step 3; no host paths or non-ASCII in the new files; test names describe behaviour, not provenance. No blocking concerns.

PR #2597, opened 2026-08-22
fix(mesh): report a state probe that fails instead of swallowing it

cagataycali:
## Problem

`Mesh._read_state` probes a robot driver defensively so that a flaky read cannot kill the state thread. All four probes ended in a bare `except Exception: pass`, so a probe that raised left **no trace at any log level**.

That is not a missing line in a log nobody reads. `_read_state` returns `None` when nothing but `peer_id` and `t` survived, and the state loop publishes nothing for a `None` -- so a peer whose joint probe raises **stops publishing state entirely**, while its presence broadcast still advertises it as `connected`. On the wire and in the log it is indistinguishable from a healthy peer that simply has no joints to report.

The loop's own report cannot cover this. `_state_loop` wraps the call in `except Exception` and logs at debug, but `_read_state`'s top-level statement kinds are `Assign, AnnAssign, Try, Try, Try, Return` -- every statement that can raise is already inside a probe `try`, so **`logger.debug("state tick error")` is unreachable for a probe failure** and only ever sees a failure of `publish` itself.

`STATE_HZ` is `10.0`, so a persistent fault is ten silent decisions per second, and `mesh/iot/shadow.py` mirrors this topic to an AWS IoT named shadow.

Measured on `upstream/main`, driving the real `_read_state`:

| case | published snapshot keys | log records |
|---|---|---|
| healthy arm, 2 joints | `['joints', 'peer_id', 't']` | 0 |
| joint probe raises `RuntimeError` | **`None`** | **0** |
| joint probe raises `ConnectionError` | **`None`** | **0** |
| task probe raises, joints fine | `['joints', 'peer_id', 't']` (task silently absent) | **0** |
| both probes raise | **`None`** | **0** |

## Change

One private helper, `Mesh._warn_read_state_once(category, exc)`, and the four probe handlers routed through it.

Each probe **keeps `except Exception`** on purpose: a flaky read must not kill the state thread, so this is a recovery path and the breadth is correct. The defect was the silence, not the surface -- narrowing any of the four is a separate behaviour change, one per site.

- The **first** failure of each category (`hw_joints`, `task_state`, `sim_world`, `sim_joints`) is logged at WARNING naming the category, the peer and the exception's `repr`. The type is in the line because it selects the operator's next move: a `ConnectionError` from a contended serial port and a `RuntimeError` from an uncalibrated arm are different jobs.
- Later failures of that category drop to DEBUG, because the loop retries at `STATE_HZ` and would otherwise emit ten warnings a second.
- The category bookkeeping is read with a `getattr` default, so a `Mesh` built through `__new__` (several tests, and the sim paths) cannot make the handler that reports a failure raise one.

**What is published is unchanged** -- every row of the table above keeps its exact snapshot keys. This changes the report, not the snapshot.

`docs/troubleshooting.md`'s `## Mesh` table gains the symptom, next to the existing row that already cites a logged WARNING line for a missing `eclipse-zenoh`.

## Tests

`tests/mesh/test_read_state_probe_failures_are_reported.py`, **10 failed / 2 passed** on `upstream/main`, 12 passed after.

The headline failures are behavioural and quote the defect:

```
_read_state swallows 4 probe failure(s) with a bare pass
(['Exception', 'Exception', 'Exception', 'Exception']); the state loop's own
debug handler cannot see them, because every statement that can raise is
already inside a probe try block
```

The **2 tests that pass on both trees are the no-overreach controls** -- a readable arm logs nothing, and an armless gateway peer logs nothing. Both fail for the obvious wrong fix (reporting an absent section rather than a thrown exception).

`TestNoProbeSwallowsItsFailure` pins the root cause structurally and derives its expectations from the source, so a fifth probe added later is held to the same rule instead of being able to end in `pass` again.

## Verification

Measured at `3a4bcd2a` against `upstream/main` `71735507`:

- `mypy strands_robots tests tests_integ`: **24 == 24**, branch-only set empty (measured on a pristine `upstream/main` worktree).
- Blast radius, 436 test files (every test naming `_read_state` / `_state_loop` / `STATE_HZ`, all of `tests/mesh`, every test that reads a docs page or walks a source tree, plus the repo-wide guards). The **identical selection, excluding the new test file, was run on both trees**: branch and a pristine `upstream/main` worktree both reported `21 failed, 16869 passed, 60 skipped, 24 errors`, with **identical failing-ID sets** (the diff is empty in both directions). All 45 are `tests/mesh/test_iot_*` needing `awsiotsdk`, a declared extra that CI installs.
- The new test file on its own: 12 passed on the branch, **10 failed / 2 passed** on `upstream/main`.
- A CodeQL AST pre-flight over the changed files reports no new `py/empty-except`, `py/mixed-returns`, `py/side-effect-in-assert`, `py/file-not-closed` or `py/unnecessary-lambda` shape. The same file on `upstream/main` carries **15** bare-`pass` handlers and this branch **11**: the change clears four of them and introduces none.

### DIFF
--- a/changelog.d/2597-read-state-probe-failures-are-reported.md
+++ b/changelog.d/2597-read-state-probe-failures-are-reported.md
@@ -1,0 +1,33 @@
+### Fixed
+
+A `_read_state` probe that fails is now reported instead of swallowed, once per
+category. `Mesh._read_state` probes a robot driver defensively so that a flaky
+read cannot kill the state thread, and all four probes ended in a bare
+`except Exception: pass` -- so a probe that raised left no trace at any log level.
+
+The consequence was not a missing log line. `_read_state` returns `None` when
+nothing but `peer_id` and `t` survived and the state loop publishes nothing for a
+`None`, so a peer whose joint probe raised stopped publishing state entirely
+while its presence broadcast still advertised it as `connected`: on the wire and
+in the log it was indistinguishable from a healthy peer with no joints to report.
+Measured against the real method, a joint probe raising `RuntimeError` or
+`ConnectionError` produced a `None` snapshot and zero log records, and a failing
+task probe dropped the `task` section out of an otherwise healthy snapshot in
+silence.
+
+The state loop's own report could not cover it. `_read_state`'s top-level
+statements are `Assign, AnnAssign, Try, Try, Try, Return`, so every statement
+that can raise is already inside a probe `try` and `_state_loop`'s
+`logger.debug("state tick error")` is unreachable for a probe failure -- it only
+ever sees a failure of `publish` itself.
+
+Each probe keeps `except Exception`: a flaky read must not kill the state thread,
+so this is a recovery path and the breadth is correct. The defect was the
+silence. The first failure of each category (`hw_joints`, `task_state`,
+`sim_world`, `sim_joints`) is now logged at WARNING naming the category, the peer
+and the exception's `repr` -- the type is in the line because it selects the
+operator's next move, a contended serial port being a different job from an
+uncalibrated arm. Later failures of that category drop to DEBUG, because the loop
+retries at `STATE_HZ` (10.0) and would otherwise emit ten warnings a second.
+
+What is published is unchanged; this changes the report, not the snapshot.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -75,6 +75,7 @@ description: Error → fix table for the most common gotchas across install, sim
 | Port already bound | Another zenoh process | Mesh auto falls back to client mode; or set `STRANDS_MESH_PORT` |
 | `mesh.alive` is `False`, `mesh.peers` stays empty | `eclipse-zenoh` missing (logged at WARNING: "eclipse-zenoh is not installed") | `uv pip install "strands-robots[mesh]"` |
 | Want mesh off | - | `STRANDS_MESH=false` or `Robot(..., mesh=False)` |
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

This PR replaces four bare `except Exception: pass` handlers in `Mesh._read_state` with a single reporting helper, `_warn_read_state_once`, that logs the first failure of each probe category (`hw_joints`, `task_state`, `sim_world`, `sim_joints`) at WARNING and subsequent failures at DEBUG. The defensive breadth of each handler is deliberately kept (a flaky read must not kill the state thread), and the published snapshot is byte-for-byte unchanged -- only the report changes. The failure mode being fixed is real and well-argued: a peer whose joint probe raises returns `None`, publishes no state at 10 Hz, and remains advertised as `connected`, indistinguishable from a healthy armless peer.

## What's good

- The warn-once-then-debug design correctly accounts for `STATE_HZ` retry pressure, and the `getattr` fallback keeps the reporting handler itself unable to raise on a `__new__`-built `Mesh`.
- `tests/mesh/test_read_state_probe_failures_are_reported.py` pins the contract behaviourally (once per category, independent categories, healthy/armless peers stay silent) and structurally (`TestNoProbeSwallowsItsFailure` derives its expectations from the AST, so a fifth probe added later is held to the rule).
- Log strings are plain ASCII, the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention, and the `docs/troubleshooting.md` row quotes the actual logged phrase so it is greppable.
- `_read_state` is reachable from both the state loop and the command dispatcher, but the only shared mutation is a GIL-atomic `set.add` whose worst-case race is a duplicate warning line -- acceptable for a reporting path.

PR #2598, opened 2026-08-22
fix(mesh): size the teleop input envelope in the unit the frames carry

cagataycali:
## Problem

`validate_input_frame` bounds how far one teleop command may reach
(`DEFAULT_INPUT_VALUE_ABS`) and `input_frame_slew_violation` bounds how fast a
joint may be commanded to travel (`DEFAULT_INPUT_SLEW_ABS`). Both are pure
numbers compared against whatever the leader driver puts on the wire, so the
unit of that wire value is part of the contract even though no signature states
it. Both were sized for radians.

lerobot's SO leader/follower default to `use_degrees=True`
(`SOFollowerRobotConfig.use_degrees` field default, read off the dataclass), and
their gripper is `MotorNormMode.RANGE_0_100` whatever `use_degrees` says. So a
shipped SO-100 class arm streams degrees and a 0-100 gripper, and there is no
configuration of it in which the radian-shaped default admits a full command.

Driving the real validator on `main`:

| frame | verdict |
| --- | --- |
| arm at rest (all joints near zero) | accepted |
| 45 deg of shoulder pan | **refused** - `\|45.0\| > 12.566` |
| gripper half open (`gripper.pos = 50`) | **refused** - `\|50.0\| > 12.566` |
| one joint at 20 deg | **refused** |
| one joint at 12 deg | accepted |

The usable reach is ~12.6 units on joints whose own travel is measured in
hundreds, and the gripper cannot be commanded past 12.57% open.

## The slew bound's own docstring is the proof

It derives its margin explicitly: *"roughly 4x the no-load speed of the Feetech
STS3215 servos on an SO-100 class arm (~6.5 rad/s at 12 V): a physical leader
arm cannot reach it, so only a synthetic stream trips it."*

`25.133 / 6.5 = 3.87x`, so that reasoning is self-consistent **only if the
frames are radians**. In the unit the frames actually carry, the same servo
produces `6.5 * 180/pi = 372.4 deg/s`, and the claim inverts:

| leader motion | commanded speed | verdict on `main` |
| --- | --- | --- |
| servo no-load sweep at 50 Hz | 372.4 units/s | **refused** - `372.4 > 25.1327` |
| half that (a brisk human sweep) | 186.2 units/s | **refused** - `186.2 > 25.1327` |

The property `tests/mesh/test_input_slew_bound.py` says the design rests on -
*"a physical leader arm at full speed is never refused"* - does not hold, and its
own assertion cannot see it because it compares against `6.5`, the rad/s figure:

```python
def test_default_is_above_leader_servo_speed(self):
    # The bound must sit above what the leader hardware itself can produce,
    # or legitimate teleoperation trips it.
    assert DEFAULT_INPUT_SLEW_ABS > 6.5
```

## Change

Both defaults are **converted**, preserving the policy they encode rather than
retuning it:

| bound | before | after | policy, unchanged |
| --- | --- | --- | --- |
| reach | `12.566` (`4*pi`) | `720.0` | two full turns |
| slew | `25.133` (`8*pi`) | `1440.0` | that envelope traversed once per second |

`1440 / 372.4 = 3.87x`, reproducing the docstring's own "roughly 4x" margin in
the right unit. The docstrings now name the unit and the driver defaults they
are sized for, and say that a fleet on a *smaller* unit (radians, or normalized
-1..1) narrows via `STRANDS_MESH_INPUT_VALUE_ABS` / `STRANDS_MESH_INPUT_SLEW_ABS`
rather than widening.

`tests/mesh/test_input_slew_bound.py` chose every magnitude relative to the old
bound, so those magnitudes are converted by the same factor via one named
`FRAME_UNITS_PER_RADIAN`. Because the bound and the stimuli move together by the
same ratio, every verdict in that file is preserved exactly; the file is not
independently meaningful against the old bound, which is why the conversion and
the constants are one change. `LEADER_MAX_STEP` and
`test_default_is_above_leader_servo_speed` are corrected to the frame unit,
which is what makes them regression evidence rather than restatements.

## Tests

`tests/mesh/test_input_envelope_units.py` pins the unit rather than a magnitude,
so a future retune must keep the relation instead of silently re-entering the
mismatch. Against `main`, that file alone is **8 failed / 5 passed**:

- failing: a 45 deg pan, a full joint sweep, a multi-turn wrist and a fully open
  gripper are all refused; a no-load sweep and a half-speed sweep are refused;
  reach is not two full turns; slew does not clear the servo speed.
- passing on both trees (the controls): the unit is read off lerobot's own
  dataclass default rather than restated; `slew == 2 * reach`; a runaway reach is
  still refused; a full-scale reversal is still refused; an operator can still
  narrow the envelope.

The repaired file was checked for weakening in both directions: it passes on
this branch (63 passed with the new file), and neutering
`input_frame_slew_violation` to always return "no violation" on this branch
fails **26** tests including `test_a_full_scale_reversal_is_still_refused`, so
the bound is still detected as load-bearing.

## Scope

`mesh/input.py` carries a one-line comment correction only (it cited the reach
bound as `(4pi)`). Widening for the real unit deliberately does not touch the
rate cap, the freshness window, or the charset guard - each is judged in its own
domain and none of them is unit-bearing.

## Verification

Measured at `c9cfadd0` on lerobot 0.6.2 / mujoco 3.5.0.

A 193-file selection (all of `tests/mesh` plus every test referencing the two
bounds or the validators they gate, plus the repo-wide docstring / string /
changelog guards), run on this branch and on `main` with the two files this PR
adds or converts excluded, so the run measures collateral damage only:

| tree | result |
| --- | --- |
| this branch | 21 failed, 4207 passed, 4 skipped, 24 errors |
| `main` | 21 failed, 4207 passed, 4 skipped, 24 errors |

45 failing ids on each, identical sets - `comm` empty in both directions. All 45
are `tests/mesh/test_iot_*` needing `awsiot`, which is a declared extra.

tests_integ`; `mypy` over the same scope is unchanged at its baseline with none
of the errors in the changed files.

`tests/mesh/test_mesh_safety_hardening.py` also wrote its out-of-envelope
stimulus as a literal sized to the old bound (`31.0`, with the comment `4*pi
default ~= 12.57`); it is now derived from the constant, and that test passes on
both trees - it is a generalisation, not a regression test.

No visual artifact applies: this changes the numeric domain of a validator, and
the rollout it gates is unchanged. The evidence is the measured accept/refuse
tables above, taken by driving `validate_input_frame` and
`input_frame_slew_violation` directly on each tree.

### Round 1 - `c9cfadd`

`tests/mesh/test_mesh_safety_hardening.py::TestH2TeleopBoundAndRate::test_default_bound_rejects_large_slew` asserted a literal `31.0`, sized to the retired `4pi` bound (`# 4*pi default ~= 12.57; 31 rad must reject`). Under `720.0` that frame is admitted, so the test the envelope relies on to refuse an out-of-range reach would have failed - turning the one required check red. The stimulus is now derived from `DEFAULT_INPUT_VALUE_ABS` rather than restated as a magnitude, the same conversion this PR already applied to `test_input_slew_bound.py`, so it stays outside the envelope through any future retune.

Verified: that class passes 5/5, and 166 passed across the envelope, validation and local-slew suites together.

### Round 2 - `78ca149`

The stale-magnitude class reached three sites outside the tests, and the first is operator-facing:

| site | stale claim | corrected to |
| --- | --- | --- |
| `README.md` (config table) | `STRANDS_MESH_INPUT_VALUE_ABS` = `12.566` (4pi), unit `(radians)` | `720` (two full turns), in frame units |
| `README.md` (config table) | `STRANDS_MESH_INPUT_SLEW_ABS` = `25.133` (8pi) | `1440` (the value envelope traversed once per second) |
| `strands_robots/teleop_mixin.py` | the mesh path's "8*pi default is radian-scoped and already requires explicit widening for driver-unit devices" | why the local constant stands on its own terms; at 1440 units/s the mesh bound is the **looser** of the two |
| `changelog.d/2363-*.md` (unreleased) | "vs the mesh path's 8*pi for radian streams" | the mesh receive path keeps its own, looser bound |

README is the file this project calls the single source of truth for operators, so those two rows were two orders of magnitude off the bound in force - and they pointed the wrong way to tune, since under a frame-unit default a radian or normalized-`-1..1` fleet *narrows* rather than widens. The `teleop_mixin` comment is worse than stale: it justified the local default by a comparison that is now inverted.

Pinned by `tests/mesh/test_readme_documents_the_envelope_in_force.py`, which reads the numbers back out of the table and compares them to the constants, and asserts each row names the unit - `720` is a plausible bound in degrees and an implausible one in radians, so the magnitude alone does not imply it. Nothing graded that pair before, which is why this drift was invisible to every check on the change that introduced it.

Verified: 3 of its 4 cases fail against the pre-fix README and all 4 pass after; 166 passed across the affected suites; the ASCII, test-name and host-path graders pass over the new file; `ruff check` and `ruff format --check` clean.


yinsong1986 (commented):
## Summary

Converts the teleop input envelope defaults (`DEFAULT_INPUT_VALUE_ABS`, `DEFAULT_INPUT_SLEW_ABS`) from radian-sized values (4pi / 8pi) to the unit the SO leader actually puts on the wire (degrees + a 0-100 gripper), preserving the encoded policy (two full turns of reach, traversed once per second) rather than retuning it. The diagnosis is well-evidenced: the slew docstring's own "roughly 4x the STS3215 no-load speed" margin only holds in radians, and the new `tests/mesh/test_input_envelope_units.py` pins the unit relationally (reading `use_degrees` off lerobot's own dataclass) rather than restating a magnitude. The existing slew-bound test file is converted by one named `FRAME_UNITS_PER_RADIAN` factor so every verdict is preserved, and the runaway/reversal refusals are re-asserted at the new bound. One consumer of the old default was missed, though: a test outside the converted files still pins the 4pi magnitude and fails on this branch (inline comment).

## What's good

- Changelog fragment present (`changelog.d/2598-teleop-envelope-frame-units.md`), per AGENTS.md step 3.
- The new test file pins the *relation* (unit read off lerobot, `slew == 2 * reach`, servo-speed clearance) rather than magnitudes, so a future retune cannot silently re-enter the mismatch.
- Runaway protection is re-verified at the widened bound (`test_a_runaway_reach_is_still_refused`, `test_a_full_scale_reversal_is_still_refused`), and the narrowing path for radian/normalized fleets is documented and tested.
- Scope discipline: the rate cap, freshness window and charset guard are correctly left alone as non-unit-bearing.

## Verification suggestions

- `python -m pytest tests/mesh/test_mesh_safety_hardening.py -q` reproduces the one branch-introduced failure (passes 20/20 on `main`).
- `python -m pytest tests/mesh/test_input_envelope_units.py tests/mesh/test_input_slew_bound.py tests/test_teleop_slew_bound.py tests/mesh/test_input_validation.py -q` — all pass on this branch.

strands_robots/mesh/security.py
```diff
@@ -157,13 +157,21 @@ def _env_pos_float(env_var: str, default: float) -> float:
...
+#: streams degrees and a 0-100 gripper -- never radians. The bound is two full
+#: turns, which clears a multi-turn wrist with headroom while still refusing a
+#: runaway three orders of magnitude out. Operators whose actuators use a
+#: smaller unit (radians, or normalized -1..1) can narrow it via
 #: ``STRANDS_MESH_INPUT_VALUE_ABS``. The module-level constant is captured
 #: at import for backward compat; the hot path calls
 #: :func:`_input_value_abs` so an operator-set env var takes effect
 #: without a process restart.
-DEFAULT_INPUT_VALUE_ABS: float = 12.566370614359172  # 4 * pi
+DEFAULT_INPUT_VALUE_ABS: float = 720.0  # two full turns, in frame units
```
yinsong1986, line 174: [MUST FIX] Widening this default breaks a test this PR did not convert: `tests/mesh/test_mesh_safety_hardening.py::TestH2TeleopBoundAndRate::test_default_bound_rejects_large_slew` (line 74) asserts `validate_input_frame({"j0": 31.0})` raises against the old 4pi default — `# 4*pi default ~= 12.57; 31 rad must reject` — and under `720.0` the frame is admitted, so the test fails `DID NOT RAISE ValidationError`. Verified: it fails on this branch and passes 20/20 on `main`, so the failure is introduced by this PR and will turn the required `call-test-lint / Test and Lint` check red, which blocks the merge outright (it is the one required context per AGENTS.md step 8). The fix is the same conversion this PR already applied to `test_input_slew_bound.py`: restate that test's stimulus relative to the bound in force (e.g. `DEFAULT_INPUT_VALUE_ABS * 1.1`, or a frame-unit magnitude above 720) so it pins "an out-of-envelope reach is refused" rather than the retired radian magnitude — exactly the stale-magnitude class the PR description identifies in the other files.

cagataycali: Confirmed and fixed in `c9cfadd`, which converts that test the same way the PR already converted `test_input_slew_bound.py`:

```python
with pytest.raises(_sec.ValidationError):
    _sec.validate_input_frame({"j0": _sec.DEFAULT_INPUT_VALUE_ABS * 2.5})
```

Derived from the bound rather than restated as a magnitude, so the stimulus stays outside the envelope if the default is ever retuned again - which is the property whose absence caused this.

Verified: `TestH2TeleopBoundAndRate` passes 5/5, and 166 passed across the envelope, validation and local-slew suites together.

The stale-magnitude class you identified went further than the test, and the rest is fixed in `78ca149`. README's configuration table still documented this knob as `12.566` (4pi) "(radians)" and the slew knob as `25.133` (8pi), so the operator-facing numbers were two orders of magnitude off the bound in force - and they implied the wrong direction to tune, since under a frame-unit default a radian or normalized fleet narrows rather than widens. `teleop_mixin` also justified its own local default by calling the mesh bound "radian-scoped", which is now doubly wrong: at 1440 units/s the mesh bound is the looser of the two. Both are corrected, and `tests/mesh/test_readme_documents_the_envelope_in_force.py` now reads the table's numbers back against the constants so the pair cannot drift again - 3 of its 4 cases fail against the pre-fix README.

cagataycali: Already addressed in `c9cfadd` — the stimulus at line 76 now reads `_sec.DEFAULT_INPUT_VALUE_ABS * 2.5` (derived from the bound, not the retired `31.0` literal). The `31.0` magnitude is gone from this file entirely.

`pytest tests/mesh/test_mesh_safety_hardening.py::TestH2TeleopBoundAndRate::test_default_bound_rejects_large_slew` passes on this branch.

---
<sub>This comment was generated by an AI agent ([strands-agents](https://github.com/strands-agents)).</sub>

yinsong1986 (approved):
## Summary

Converts the mesh teleop input envelope defaults (`DEFAULT_INPUT_VALUE_ABS`: 4pi -> 720, `DEFAULT_INPUT_SLEW_ABS`: 8pi -> 1440) from radian-sized magnitudes to the unit the SO leader actually puts on the wire (degrees, plus a 0-100 gripper), preserving the encoded policy (two full turns of reach; that envelope traversed once per second) rather than retuning it. The diagnosis is convincing: the old slew docstring's own "roughly 4x the STS3215 no-load speed" arithmetic only holds in radians, and in the frame unit the bound sat an order of magnitude below the leader's own servo speed, so legitimate teleop was refused. The env-var names and their semantics are unchanged (operators on smaller units narrow via the existing knobs), the default change is announced in a changelog fragment and README's config table, and the new `test_readme_documents_the_envelope_in_force.py` grades the table against the constants so the documentation cannot drift again.

## What's good

- The unit premise is read off lerobot's own dataclass default (`use_degrees=True`) rather than restated, so the test fails if the upstream driver changes.
- `test_input_slew_bound.py` stimuli are converted by one named `FRAME_UNITS_PER_RADIAN` factor, preserving every verdict; the runaway-reach and full-scale-reversal refusals are still pinned, so the widened envelope is verified to still protect.
- The prior review's stale-literal finding (`31.0` in `test_mesh_safety_hardening.py`) is resolved at `c9cfadd` by deriving the stimulus from the constant; verified passing at head.
- Verified at head: the four touched test files pass (87 tests), and the full `tests/mesh` selection is green except one pre-existing headless-GL rendering failure that reproduces identically on `main`.
- Changelog fragment per AGENTS.md step 3; ASCII, host-path and test-name graders pass over the new files.

PR #2599, opened 2026-08-22
fix(mesh): one reader at a time on the motor bus, with self-healing

cagataycali:
## Problem

A serial motor bus is a single conversation: the host writes a sync-read packet addressed to a set of servo ids, then reads the replies off the wire. Two threads doing that at once interleave their packets, and the feetech/dynamixel SDKs refuse outright:

```
ConnectionError("Failed to sync read 'Present_Position' on ids=[1, 2, 3, 4, 5, 6]
                after 3 tries. [TxRxResult] Port is in use!")
```

A mesh peer does this to itself. Four threads in one process reach for the same device:

| thread | why it touches the bus |
|---|---|
| state probe (`_read_state`, `STATE_HZ`) | reads joint positions for the fleet snapshot |
| hardware camera publisher (`_publish_hardware_cameras`, `STRANDS_MESH_CAMERA_HZ`) | lerobot's `get_observation()` sync-reads the MOTORS before it grabs any frame |
| sensors probe (`_read_imu`) | reads the observation for imu keys |
| IoT camera offload | same `get_observation()` |

...and teleop writes to it besides. Nothing serialised them, so on real hardware the reads collided continuously. The SDK's three retries do not help: all three land inside the same collision.

The joint probe's failure is not partial. `_read_state` returns `snapshot if len(snapshot) > 2 else None`, so when the joint read raises, the state loop's `if state:` is False and the peer publishes **no state message at all** - the fleet sees a peer that has gone quiet, not one with missing joints.

Two further failures in the same family:

* **A camera can hide an arm's joints.** `get_observation()` sync-reads the motors FIRST, then loops the cameras. A camera that raises discards the joint positions *already in hand*, so a single dead USB camera erased an arm's entire joint telemetry while its presence stayed healthy. Joints and frames are independent facts and must fail independently.
* **A stranded in-use flag never recovers.** The vendored SDK sets `port.is_using = True` in `txPacket` and clears it with a bare assignment AFTER the call, with no `try`/`finally`. An exception between those two lines leaves the flag set for the life of the process, and every later read fails before the port is touched.

## Measured

Four *real* strands readers (`Mesh._read_state`, `Mesh._publish_hardware_cameras`, `Mesh._read_imu`, `InputReceiver._default_apply`) driven against one fake bus that implements the vendored feetech protocol - `txPacket` refuses when `port.is_using` is set, otherwise sets it. Barrier-synchronised, 25 rounds x 3 repeats:

| metric | upstream/main | this branch |
|---|---|---|
| state snapshots published with joints | **0 / 75** | **75 / 75** |
| imu reads that returned a reading | **0 / 75** | **75 / 75** |
| teleop writes accepted (control) | 75 / 75 | 75 / 75 |

The stranded-flag recovery that used to sit beside this is no longer part of the PR (see Round 3): both columns behave identically there, which the figure states.

bus access measurement

## What ships

`strands_robots/bus_access.py` (new) plus the four call sites, because without them the module is dead code:

| file | change |
|---|---|
| `strands_robots/bus_access.py` | new, +219: `bus_lock`, `read_observation`, `write_action`, `read_joints` |
| `strands_robots/mesh/core.py` | the state probe reads joints through the lock; the camera publisher reads through the lock |
| `strands_robots/mesh/input.py` | `_default_apply` writes through the same lock the readers take |
| `strands_robots/mesh/sensors.py` | the imu probe reads through the lock |
| `strands_robots/mesh/iot/camera_offload.py` | the offload reads through the lock |

## Design invariants

* **The lock lives on the DEVICE** (`_strands_bus_lock`), not on any caller, because the device is what is shared: the mesh modules and the teleop rail hold different wrappers around one lerobot robot.
* **It is an `RLock`**, so lerobot's own `get_observation()` may call back into a locked read.
* **The module imports nothing from lerobot or the mesh**, which is the only way every reader can share one lock without an import cycle.

## Scope

Deliberately out of scope, though the same files carry them elsewhere in the work this was extracted from:

* loop **pacing** (`sensors.py`, `input.py` and the state loop all pace with `_stop_event.wait`) - a separate change with its own measurements.
* `_read_state` **degradation reporting** (naming which probe failed in the snapshot) - separate.
* the camera-degraded **connect/rollback** contract in `hardware_robot.py` - separate; despite appearances that file needs no change here.

The camera publisher already has its own fallback (it re-reads `inner.cameras` when `get_observation()` raises), so its *frames* survive a collision. It is included because the motor read it performs first is what harms the other readers.

## Verification

Measured at `b2e5759`:

* `mypy strands_robots tests tests_integ`: **24 errors on this branch, 24 on `upstream/main`**, identical sets, none in the changed files.
* `pytest tests/test_bus_access_serializes_motor_reads.py tests/test_bus_read_joints.py`: **21 passed**.
* Full suite, this branch vs `upstream/main`, same interpreter, run sequentially: **29 failed / 34923 passed** on `upstream/main` and the same **29** failing ids here, identical sets both ways (`comm` empty in both directions). 24 of them are `ModuleNotFoundError: No module named 'awsiot'`, a declared extra this environment does not install.
  An earlier revision of this branch had 4 branch-only failures - one existing state-probe test and three in the base-exception census guard - both fixed in the rounds below and re-verified green.
* Pre-fix: the two pinned files cannot collect on `upstream/main` at all (the module they pin does not exist there), so the regression evidence is the behavioural measurement above rather than a failing-assert split.

### Round 1 - `bf0bf48`

The parallel-probe helper hand-rolled an exception-marshal box - `try` / `except BaseException` per callable, appending to a shared list - which CodeQL reports as `py/catch-base-exception`, and whose review thread blocks the merge under `required_review_thread_resolution`.

AGENTS.md splits that rule by direction, and this site is the half not obliged to keep the box: the helper **creates** the threads it collects from, so `ThreadPoolExecutor` is the pattern. `_WorkItem.run` catches `BaseException` and `Future.exception()` hands it back with identity preserved, so the handler now belongs to CPython and the alert, the handler and the blocking thread go together.

Narrowing to `Exception` was the alternative and it would have cost the measurement: a worker raising `SystemExit` would be collected as nothing at all, since `threading.excepthook` ignores that class specifically and writes no traceback - a silent hole in the one file whose job is to notice failures.

`max_workers` is the whole callable set deliberately. A callable left queued for a free worker never overlaps the others, and overlap is the condition under test: the guard-rail case asserts the fake bus still answers `[TxRxResult] Port is in use!` when the lock is bypassed, so a helper that serialised its own workers would leave the other three tests proving nothing.

Verified: 30 passed across the two pinned files, and the concurrency file run **20 consecutive times with no failure**, so the overlap is not schedule-dependent. `ruff check` / `ruff format --check` clean on the changed file.

### Round 2 - `be76f11`

A real regression in this branch, found by composing it with the two siblings that share its files rather than by its own suite.

`read_joints` took the bus path on any device exposing `bus.sync_read` and handed the result back unshaped when it was not a `dict`. `Mesh._read_state` then iterates that value (`for key, value in obs.items()`), so a non-mapping raised `TypeError` there, the probe's `except Exception: pass` swallowed it, and `_read_state` returned `None` on its `len(snapshot) > 2` check - **the peer published no state message at all**. That is the exact failure this PR exists to remove, reintroduced one layer up.

The bus is being *detected* here, not trusted, and the detection was half done: `bus` is an ordinary attribute name a wrapper or a proxy can also use, and its presence says nothing about what `sync_read` answers with. The fallback that already serves a driver exposing no bus now also serves one whose bus answers with a non-mapping. It stays a route to the reading rather than a way to swallow a failure - where the fallback observation raises too, the driver's own error surfaces.

`tests/test_bus_read_joints.py` had a case pinning the old passthrough (`test_a_bus_returning_something_odd_is_passed_through_unshaped`), which asserted the defect; it is converted into the two the contract actually has.

| measurement | before | after |
| --- | --- | --- |
| `tests/mesh/test_mesh.py::TestReadState::test_publishes_joints_filtered_against_cameras` | **failed** | passed |
| failing-id set over `tests/mesh` + both bus files vs `origin/main` | 1 regression | **identical, 147 each** (`comm -23` empty; all from absent declared extras) |

Attribution matters here: the regression is this branch **alone**, not the composition, so it would have turned the required check red on its own. The earlier verification ran the two new pinned files instead of the existing tests that the `core.py` call site changes, which is what let it through.

## Merge order with the sibling branches

`scripts/check_merge_base_overlap.py --all-open` reports this branch in two pairs, and neither branch's CI can compile the other's tree:

| pair | shared path | composed result |
| --- | --- | --- |
| #2598 + #2599 | `strands_robots/mesh/input.py` | auto-merges, no conflict |
| #2597 + #2599 | `strands_robots/mesh/core.py` | **content conflict** |

Both read `MERGEABLE` against `main` because each is mergeable against a base containing neither. The conflict is a real one and its resolution keeps both changes: #2597 replaces this probe's `except Exception: pass` with `self._warn_read_state_once("hw_joints", exc)`, while this branch adds the `bus_recoveries` count inside the same `try`. Composed correctly, the recovery count sits under `if joints:` and the reporting handler replaces the swallow - measured green, and the composition of all three carries zero regressions against `main` once the fix above is in.

Suggested order: **#2597 first**, then this branch absorbs `main` and takes that one-line resolution. Whichever merges second must do so - resolving it now would be resolving against a tree that has not landed.

### Round 3 - `b2e5759` (scope reduction)

Review feedback established that the stranded-flag recovery rested on a proof this tree cannot make, and `port_busy_action`'s own docstring had stated that proof: clearing the SDK's `port.is_using` is safe *"only under a proof, never on a hunch"*, the proof being that **every** read and write happens inside `bus_lock`, so a port still claiming to be in use is "either stale ... or held by a reader that bypasses the lock".

Verified: that second case is the primary hardware path, not a bug case. `strands_robots/hardware_robot.py` drives the same device outside `bus_access` and does not import it:

| line | call | thread |
| --- | --- | --- |
| 890 | `self.robot.get_observation()` | ROS telemetry publish |
| 1426 / 1597 | `await asyncio.to_thread(self.robot.get_observation)` | task preflight, policy rollout |
| 1630 | `await asyncio.to_thread(self.robot.send_action, action_dict)` | policy rollout |
| 2734 | `self.robot.send_action(action)` | `send_action` facade |

So a policy commanding the arm while the mesh reads state is the normal operating mode, and the flag a probe would clear can be one a live exchange legitimately holds. The direction of the trade is what settles it: on `main` that collision is a **safe refusal** (the SDK declines without touching the wire; the probe publishes nothing that cycle), while the recovery re-transmitted onto a half-duplex UART mid-reply - able to garble a goal-position write for an arm under power, with the "REAL owner" report firing only *after* the first interleave, and not at all when the retry lands after the other exchange finishes. A silent no-data cycle traded for a silent corrupted write is the wrong direction, and this diff introduced it rather than inheriting it.

Removed: `port_busy_action`, `clear_stale_port_busy`, `write_refusal`, `recovery_count`, the retry wrapper, and the `bus_recoveries` field published from it - `bus_access.py` **372 -> 219 lines**, with the tests that pinned them.

What ships is the two halves whose proof needs no premise about other callers, both measurements unchanged:

1. one `RLock` on the **device**, serialising every reader and writer that goes through this module (0/75 -> 75/75 state snapshots with joints);
2. a joints read that a failing camera cannot discard.

The alternative offered was converting `hardware_robot.py`'s call sites inside this PR. Declined deliberately: that is the primary hardware path and deserves its own concurrency measurements rather than being appended at a third review round - and it is also the prerequisite that would make the recovery sound, so the ordering is real. Filed as **#2602** with the call-site table and the sequencing, so the recovery can return on top of a tree where the lock is the proof it claimed to be.

The `bus_recoveries` mention is out of the news fragment, and the description's earlier stranded-flag table describes behaviour that is no longer in this branch - it is kept above as the record of why #2602 exists.


tests/test_bus_access_serializes_motor_reads.py
```diff
@@ -0,0 +1,210 @@
...
+def _run_all(fns: list) -> list[BaseException]:
+    """Run callables in parallel threads, collecting whatever they raise."""
+    errors: list[BaseException] = []
+    lock = threading.Lock()
+
+    def wrap(fn):
+        def inner():
+            try:
+                fn()
+            except BaseException as exc:  # noqa: BLE001 - collected for the test
```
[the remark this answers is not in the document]
cagataycali: Fixed in `bf0bf48` by removing the handler rather than dismissing the alert.

AGENTS.md splits this rule by direction, and this site is the half that is not obliged to hand-roll the box: the helper creates the threads it collects from, so `ThreadPoolExecutor` *is* the pattern. `_WorkItem.run` catches `BaseException` and `Future.exception()` hands it back with identity preserved, so the handler now belongs to CPython and the alert, the handler and this blocking thread go together.

```python
with futures.ThreadPoolExecutor(max_workers=len(fns)) as pool:
    submitted = [pool.submit(fn) for fn in fns]
return [exc for fut in submitted if (exc := fut.exception()) is not None]
```

Narrowing to `Exception` was the other option and it would have cost the measurement: a worker raising `SystemExit` would be collected as nothing at all, since `threading.excepthook` ignores that class specifically and writes no traceback - a silent hole in the file whose whole job is to notice failures.

`max_workers` is the whole callable set deliberately. A callable left queued for a free worker never overlaps the others, and overlap is the condition under test: the guard-rail case asserts the fake bus still answers `[TxRxResult] Port is in use!` when the lock is bypassed, so a helper that serialised its own workers would leave the other three tests proving nothing. Verified 30 passed, and the concurrency file run 20 consecutive times with no failure, so that overlap is not schedule-dependent.

cagataycali:
## `tests/mesh/test_mesh.py::TestReadState::test_publishes_joints_filtered_against_cameras` fails on this branch and passes on `main`

Measured with one identical 89-file selection (the two new test files excluded, since `main` does not have them) run on both trees:

| tree | result |
|---|---|
| `upstream/main` `71735507` | 2952 passed, 0 failed |
| `extract/bus-access` `6eecc68a` | **1 failed**, 2951 passed |

```
tests/mesh/test_mesh.py::TestReadState::test_publishes_joints_filtered_against_cameras
    m = Mesh(_FakeRobot(with_inner_robot=True), peer_id="hw")
    s = m._read_state()
>   assert s is not None
E   assert None is not None
```

It reproduces isolated on `6eecc68a` and on `bf0bf48d` (`1 failed in 1.78s`), so it is not a selection-ordering artifact.

### Root cause

`_FakeRobot(with_inner_robot=True)` builds `inner = MagicMock()`. `_read_state` now reads through `read_joints(inner)`, whose first branch is

```python
bus = getattr(device, "bus", None)
if bus is None or not hasattr(bus, "sync_read"):
    return read_observation(device)
```

On a `MagicMock`, `.bus` auto-creates a child mock that *does* have `sync_read`. So the fake takes the `bus.sync_read("Present_Position")` fast path instead of the `get_observation()` fallback it was built to model; `sync_read` answers with a `Mock`, `isinstance(raw, dict)` is False, the joints never materialise, and `_read_state()` returns `None`.

The conversion is right and the fake is what is now inaccurate: `MagicMock` silently claims a capability the driver it stands for does not have.

### The fix is one line, and the branch this was extracted from already carries it

```diff
             inner = MagicMock()
             inner.is_connected = True
+            # A driver whose ONLY reader is get_observation (no separate motor
+            # bus), which is the fallback path in read_joints. MagicMock would
+            # otherwise answer .bus.sync_read() with a Mock and this fake would
+            # stop modelling anything real.
+            inner.bus = None
             inner.name = "so100"
```

Verified locally: `1 passed` for that case, `46 passed` for the whole file, with no other file in the 89-file selection changing verdict.

Worth saying explicitly, because two other repairs are tempting and both are worse:

- relaxing the assertion would hide a genuine `None` from a probe, which is the failure `_read_state` reporting exists to make visible;
- making `read_joints` treat a `Mock` bus as absent would move test-double detection into the production path.

### Why CI has not shown this yet

`call-test-lint / Test and Lint` on `6eecc68a` reads `Run lint: success` / `Run tests: cancelled`, 08:01:42Z -> 08:09:41Z. It was cancelled by the `bf0bf48d` push rather than failing, so the branch has not yet completed a test job.

cagataycali:
### Triage of the 33 failures a full-suite run reports on this branch

Following up on the previous comment, because a batch that size invites being dismissed wholesale and exactly one of them is a code regression. All 33 accounted for:

| n | family | cause | real? |
|---|---|---|---|
| 21 | `tests/mesh/test_iot_*` | `ModuleNotFoundError: No module named 'awsiot'` | no - declared extra, installed in CI |
| 3 | `tests/test_codeql_query_filters.py::TestTheMarshalBoxCensusIsDerivedFromTheTree` | the hand-rolled `except BaseException` marshal in `_run_all` | **already fixed by `bf0bf48d`** |
| 3 | `tests/training/test_lerobot*.py` | `encode() takes 1 positional argument but 2 were given` - draccus below lerobot's `>=0.11.6` floor | no - environment |
| 2 | `tests/test_device_connect_hardening.py` | `device_connect_agent_tools` absent | no - declared extra |
| 2 | `tests/test_doctor.py::TestDoctorVersionResolution` | `PackageNotFoundError: strands-robots` - worktree not pip-installed | no - environment |
| 1 | `tests/policies/lerobot_local/test_vla_jepa.py` | `vla_jepa_clip_actions` not in the processor registry (lazy registration) | no - environment |
| 1 | `tests/mesh/test_mesh.py::TestReadState::test_publishes_joints_filtered_against_cameras` | `assert None is not None` | **yes** |

Two things worth knowing:

**The census failures are already cured, and the run that reported them was measuring the previous head.** Verified:

| head | `tests/test_codeql_query_filters.py` |
|---|---|
| `6eecc68a` | 3 failed, 19 passed |
| `bf0bf48d` | **22 passed** |

`_run_all`'s hand-rolled marshal was a second `except BaseException` that did not end in a lexical raise, which is what that guard measures; delegating to `ThreadPoolExecutor` removed the handler rather than annotating it, so the census is back to one construct. Nothing further needed there.

**The `_read_state` failure is the only one that survives `bf0bf48d`**: `1 failed, 45 passed` in that file at the current head. That is the `inner.bus = None` fake gap from the previous comment.

So the branch is one line away from a clean suite here.

strands_robots/bus_access.py
```diff
@@ -0,0 +1,359 @@
...
+
+    Deliberately narrow: it touches one boolean on the port handler and nothing else - no reopen, no
+    reconnect, no write to a motor. A bus that does not expose a port handler (a stub, a sim, a
+    future driver) is left alone and reported as unrecovered, so the caller re-raises the original
+    error rather than claiming a repair it did not make.
+    """
+    handler = getattr(bus, "port_handler", None)
+    if handler is None or not getattr(handler, "is_using", False):
+        return False
+    handler.is_using = False
```
yinsong1986: [MUST FIX] `handler.is_using = False` can clear a flag a *live* in-process exchange is legitimately holding, because the safety proof stated in `port_busy_action` -- "every read and write here happens inside `bus_lock`" -- is not true of this tree at this head. `strands_robots/hardware_robot.py` still drives the same `self.robot` directly, outside `bus_access`: the policy rollout loop (`await asyncio.to_thread(self.robot.get_observation)` and `await asyncio.to_thread(self.robot.send_action, action_dict)` in the task loop), the ROS telemetry publish (`observation = self.robot.get_observation()`), and the `send_action` facade. Those threads run concurrently with the mesh probes in the same process -- the rollout loop even publishes per-step mesh telemetry -- so "policy driving the arm while the mesh reads state" is the *normal* operating mode, not a bug case.

In that mode the sequence is: the policy's exchange sets `is_using`; a mesh probe (holding only the mesh's lock, which the policy does not take) gets `Port is in use!`; `_read_recovering_a_stale_flag` clears the flag and immediately re-transmits -- interleaving a second conversation onto a half-duplex servo UART while the first is mid-reply. The "REAL owner" report only fires on the *second* busy failure; the first clear has already interleaved by then, and if the retry happens to land after the policy's exchange finishes (the common case, since exchanges take milliseconds), no report ever fires and the corruption of the policy's in-flight read/write is silent. On `main` this same collision is a safe refusal (the SDK returns port-busy without touching the wire; the probe publishes nothing that cycle).

This is must-fix because it is a race/data-corruption regression introduced by the diff on the primary hardware path, with physical-motion stakes: the interleave can garble a `send_action` goal-position exchange for an arm a policy is actively commanding. Two ways to resolve within this PR: (a) route `hardware_robot.py`'s direct `self.robot.get_observation()` / `self.robot.send_action(...)` call sites through `bus_access.read_observation` / `write_action` (the same one-line change the mesh call sites got, which makes the lock actually the proof the docstring claims), or (b) keep the flag-clear out of this PR until those call sites are converted, shipping the serialisation and `read_joints` halves as-is. The out-of-scope note in the description covers `hardware_robot.py`'s connect/rollback contract, not its bus access, so this isn't a deliberately deferred gap.

yinsong1986 (commented):
## Summary

Adds `strands_robots.bus_access`, a device-scoped `RLock` serialising every mesh reader (state probe, hardware camera publisher, IMU probe, IoT camera offload) and the teleop writer on one motor bus, plus two targeted fixes in the same family: `read_joints` reads `Present_Position` directly off the bus so a dead camera inside lerobot's `get_observation()` can no longer discard joints already in hand, and a stranded SDK `is_using` flag is now cleared once, under the lock, with a per-port `bus_recoveries` counter published in the peer's state so physical degradation stays visible. The diff matches the description, the measurements are convincing, and the mesh-side call sites are complete.

## What's good

- The lock lives on the device object rather than any caller, so wrappers in different modules genuinely share it; the creation race and the `__slots__`/frozen fallback are both handled and tested.
- Read/write asymmetry on recovery (a read may clear the flag, a write refuses and names the read that heals it) is well argued and pinned.
- Recovery is counted per port and published rather than silently healed, per the no-silent-defaults rule.
- Tests include a guard-rail test proving the fake bus still reproduces the live bug, and the prior `BaseException`-marshalling finding was resolved by delegating to `ThreadPoolExecutor` per AGENTS.md's direction split.

strands_robots/bus_access.py
cagataycali: Confirmed, and taken as written. Resolved by option (b) in `b2e5759`: the flag recovery comes out of this PR.

I verified the premise before acting on it - `hardware_robot.py` drives the same device at six call sites and does not import `bus_access` at all (890, 1426, 1597, 1630, 2734 plus the facade), so "held by a reader that bypasses the lock" is the primary hardware path rather than the bug case `port_busy_action`'s docstring assumed. That docstring named its own proof ("every read and write here happens inside `bus_lock`"), and this tree cannot make it, so the code was correct only against a repository it was extracted from.

The asymmetry you name is the deciding half. On `main` the collision is a *safe* refusal - the SDK declines without touching the wire and the probe publishes nothing that cycle - and the recovery traded that for a re-transmit onto a half-duplex UART mid-reply, with the "REAL owner" report firing only after the first interleave and not at all when the retry lands after the other exchange completes. A silent no-data cycle is a worse outcome than nothing; a silent corrupted goal-position write to an arm under power is a different category, and this diff introduced it rather than inheriting it.

Removed: `port_busy_action`, `clear_stale_port_busy`, `write_refusal`, `recovery_count`, the retry wrapper, and the `bus_recoveries` field the state probe published from it - `bus_access.py` 372 -> 219 lines, and the tests that pinned them go with them.

What remains is the two halves whose proof needs no premise about other callers: one `RLock` on the device serialising every reader and writer that goes through this module, and a joints read a failing camera cannot discard. Both measurements are unchanged.

I chose (b) over (a) deliberately. (a) would put the primary hardware path into a PR at its third review round, and the conversion deserves its own concurrency measurements rather than being appended here - and it is also the prerequisite that makes the recovery sound, so the ordering is real rather than cosmetic. Filed as #2602 with the call-site table and the sequencing, so the recovery can return on top of a tree where the lock is the proof.

Verified: 25 passed across the two bus files and `TestReadState`; over `tests/mesh` plus both bus files the failing set is identical to `origin/main`'s at 147 ids (`comm -23` empty, all from absent declared extras); `ruff check` / `ruff format --check` clean.

yinsong1986 (approved):
## Summary

Adds `strands_robots.bus_access`, a device-scoped `RLock` that serialises every mesh reader (state probe, hardware camera publisher, IMU probe, IoT camera offload) and the teleop writer on one motor bus, plus `read_joints`, a joints-only read so a camera raising inside lerobot's `get_observation()` can no longer discard joint positions already in hand. The module imports nothing from lerobot or the mesh, the lock lives on the device so wrappers in different modules share it, and lock creation is double-checked under a module guard with a shared-fallback for `__slots__`/frozen/proxy devices that refuse the attribute. All three prior review rounds are resolved at this head: the hand-rolled `BaseException` marshal box is gone (delegated to `ThreadPoolExecutor`), a non-mapping `sync_read` answer now takes the observation fallback instead of reaching the state probe unshaped, and the in-use flag recovery -- whose safety proof `hardware_robot.py`'s unlocked call sites defeat -- is removed entirely and sequenced behind #2602.

## What's good

- Lock identity verified end to end: hardware teleop hosts pass the inner lerobot driver as the receiver's target (`teleop_mixin.py`), so `write_action` locks the same object the probes lock -- the serialisation the PR claims is the one that ships.
- The Round-3 scope reduction is the right trade: on a tree where `hardware_robot.py` drives the device outside the lock, a stranded-flag clear can re-transmit mid-reply on a half-duplex UART; a safe refusal beats a silent corrupted write, and #2602 records the prerequisite ordering.
- Test discipline per AGENTS.md: the guard-rail test pins that the fake bus still refuses overlap when the lock is bypassed, the RLock re-entrancy and error-release contracts are pinned, lock-held-during-wire is probed from another thread (asking on the reading thread would prove nothing against an RLock), and the old passthrough test was converted into the two the contract actually has rather than deleted.
- Verified at `b2e57594`: both pinned files pass (21), the Round-2 regression file `tests/mesh/test_mesh.py` passes (46), and the sensor/input/offload suites the call sites touch pass (265). Ruff clean, changelog fragment present, no host paths, ASCII-only strings.

## Verification suggestions

- The merge-order note in the description is real: `strands_robots/mesh/core.py` conflicts with #2597 (`_read_state`'s `except Exception: pass` vs the `bus_recoveries`-era `try` body). Whichever merges second should absorb `main` and re-run `tests/mesh/test_mesh.py` plus both pinned bus files on the composition, per AGENTS.md's semantic-conflict guidance.
- On real hardware, a quick spot-check that two arms stay independent: `bus_lock(left) is not bus_lock(right)` holds by construction except for devices that refuse attributes, which all share one fallback lock -- acceptable over-serialisation, but worth knowing if a proxy-wrapped fleet ever runs hot.

PR #2600, opened 2026-08-22
fix(training): a single-node elastic launch rendezvouses on an address a peer can reach

cagataycali:
## The defect

`elastic_launch_callable` left two rendezvous values for torch to resolve, and both of
torch's fallbacks are addresses no peer can dial.

**MASTER_ADDR.** `RendezvousStoreInfo.build` resolves the address itself when
`local_addr` is None. In torch 2.11 that line reads, verbatim:

```python
addr = local_addr or socket.getfqdn()
```

On a host whose `getfqdn()` answers with a reverse-DNS PTR name - the name *of an
address* rather than a hostname - that name has no forward lookup, so the agent
publishes an address the worker store's client can never resolve. Measured here, a
PTR name is unresolvable while a loopback literal is not:

| name | `gethostbyname` |
| --- | --- |
| `1.0.0.0...ip6.arpa` | `gaierror` |
| `127.0.0.1` | resolves |

**The endpoint.** `rdzv_endpoint` fell back to the literal `localhost:0`, and neither
half of that is usable. Port 0 is not an address a client can dial (measured:
`connect(("127.0.0.1", 0))` is refused). Whether the c10d backend hosts the store or
dials it is decided by `_matches_machine_hostname(host)`, and when it dials,
`_create_tcp_store` retries with a `read_timeout` defaulting to 60 seconds. `localhost`
also resolves to both `::1` and `127.0.0.1` wherever both stacks are configured, which
lets the store server and its client miss each other.

Both resolutions happen inside libtorch's C++ socket code, so a wrong value there is a
wait no Python-level timeout, `pytest-timeout` signal or rendezvous budget can end: the
run parks on "Rendezvous'ing worker group" with no error at all.

Only one `rdzv_configs` key was set, by `LaunchConfig.__post_init__`, leaving the two
the c10d path actually reads at their defaults.

## What changes

`rendezvous_endpoint` derives the endpoint: an explicit one always wins, a single-node
launch gets a concrete free port on `127.0.0.1`, and a multi-node launch with no
endpoint is **refused with the reason** rather than falling back to a loopback address
that can only rendezvous with itself. `launch_local_addr` pins a single-node launch to
`127.0.0.1` and keeps torch's own resolution for multi-node (the address really must be
reachable from the other nodes), warning when the resolved name is a reverse-DNS
artifact instead of leaving the operator to watch a silent hang. `rdzv_timeout_s` bounds
each phase in the unit each backend reads, with `STRANDS_TRAIN_RDZV_TIMEOUT_S` and
`STRANDS_TRAIN_LOCAL_ADDR` as operator overrides.

The three `rdzv_configs` keys are the ones consumed on these paths, not a guess:
`read_timeout` at `c10d_rendezvous_backend.py:148`, `join_timeout` via
`create_handler`'s `_get_timeout(params, "join")`, and `timeout` for the static backend.

Captured from the real `LaunchConfig` on each tree:

| | before | after |
| --- | --- | --- |
| `rdzv_endpoint` | `localhost:0` | `127.0.0.1:39911` |
| endpoint port dialable | no | yes |
| `local_addr` | `None` | `127.0.0.1` |
| MASTER_ADDR published | `socket.getfqdn()` | `127.0.0.1` |
| `rdzv_configs` | `{timeout: 900}` | `{timeout: 120, read_timeout: 120, join_timeout: 120}` |

rendezvous address and bound

## Scope, and what was not measured

The hang itself does **not** reproduce on the runner this was developed on: its
`getfqdn()` answers `cagatay`, which forward-resolves, and the two real single-node
spawns in `tests/training/test_inproc.py` pass there in 4.56s before the change. So the
evidence offered is the torch source line that makes `local_addr=None` resolve to
`getfqdn()`, the measured unresolvability of a PTR name, the measured undialability of
port 0, and the captured `LaunchConfig` above - not a reproduced hang. The tests inject
`fqdn` and `port_picker` for exactly this reason: no assertion depends on the runner's
own DNS or on a real bind.

Multi-node behaviour is deliberately unchanged apart from the refusal: guessing an
address reachable from another node is not something this function can do.

## An env var that could not fall back

Reading `STRANDS_TRAIN_RDZV_TIMEOUT_S` is meant to fall back for every unusable
spelling, so that a typo cannot restore the unbounded wait. Converting with
`int(float(raw))` under `except (TypeError, ValueError)` does not achieve that:
`int(float("inf"))` raises **OverflowError**, which escapes that handler, so `inf`,
`-inf` and `1e999` would each turn a training launch into a traceback. Reading the
value as a float and rejecting non-finite and non-positive values covers all of them:

| value | read as |
| --- | --- |
| `900` | 900 |
| `soon`, `0`, `-5` | 120 (default) |
| `inf`, `-inf`, `1e999`, `nan` | 120 (default) |

## Tests

`tests/training/test_inproc_rendezvous.py`, 35 pure tests in 0.21s. Four classes: the
endpoint, the bound, the published address, and - because a helper returning the right
value is worth nothing if the config is still built from torch's defaults - a class that
captures the real `LaunchConfig` and reads it.

Pre-fix, against `upstream/main`: the file as shipped cannot collect, because every
helper is new. Keeping only the wiring class and stubbing the one new constant it needs
gives behavioural evidence instead - **4 failed / 1 passed**:

```
assert 'localhost' == '127.0.0.1'                                    (endpoint host)
assert None == '127.0.0.1'                                           (local_addr)
TypeError: elastic_launch_callable() got an unexpected keyword argument 'local_addr'
AssertionError: ('timeout', {'timeout': 900})  assert 900 == 120     (only one phase bound)
```

The one pass is the control: an explicit `rdzv_endpoint` already reached the config
before this change, and still does.

## Verification

Measured at `278c34b6`.

75 test files - all of `tests/training`, everything naming the changed symbols, and the
repo-wide docstring, string-hygiene, exception-clause and changelog guards - run on this
branch and on a pristine `upstream/main` worktree, with the new file excluded from both
so the comparison isolates the source change:

| --- | --- | --- | --- |
| this branch | 3 | 3683 | 5 |
| `upstream/main` | 3 | 3683 | 5 |

Failing-ID sets identical, `comm` empty in both directions. The 3 are pre-existing and
dependency-related, not reached by this diff: two `TestInProcessTrainerCorrectness`
cases and `test_lerobot_e2e` fail on `encode() takes 1 positional argument but 2 were
given`, a draccus version below lerobot's declared floor in this environment.

`tests/training/test_inproc.py` is 24 passed on both trees, including the two spawns
that drive a real single-node rendezvous - the endpoint and the address change, the
launch does not. `ruff check` and `ruff format --check` clean over
`strands_robots tests tests_integ`; mypy 24 errors on both trees, identical sets, none
in either changed file.

### DIFF
--- a/changelog.d/2600-single-node-rendezvous-address.md
+++ b/changelog.d/2600-single-node-rendezvous-address.md
@@ -1,0 +1,32 @@
+### Fixed: a single-node elastic launch rendezvouses on an address a peer can reach
+
+`elastic_launch_callable` left two rendezvous values for torch to resolve, and both of
+torch's fallbacks are addresses no peer can dial.
+
+`RendezvousStoreInfo.build` resolves `MASTER_ADDR` itself when `local_addr` is None, as
+`addr = local_addr or socket.getfqdn()`. On a host whose `getfqdn()` answers with a
+reverse-DNS PTR name - the name of an address rather than a hostname - that name has no
+forward lookup, so the agent publishes an address the worker store's client can never
+resolve. The endpoint fell back to the literal `localhost:0`, whose port is not an
+address a client can dial and whose host resolves to two different stacks wherever both
+are configured. Both resolutions happen inside libtorch's C++ socket code, so a wrong
+value there is a wait that no Python timeout, `pytest-timeout` signal or rendezvous
+budget can end: the run parks on "Rendezvous'ing worker group" with no error at all.
+
+Both are derived now. A single-node launch pins `127.0.0.1` and picks a concrete free
+port; a multi-node launch with no endpoint is refused with the reason rather than falling
+back to a loopback address that can only rendezvous with itself; a multi-node launch
+keeps torch's own resolution, because the address really must be reachable from the other
+nodes, but says so when the resolved name is a reverse-DNS artifact instead of leaving
+the operator to watch a silent hang. `STRANDS_TRAIN_LOCAL_ADDR` and
+`STRANDS_TRAIN_RDZV_TIMEOUT_S` are the operator overrides.
+
+Each rendezvous phase now carries a bound in the unit the backend reading it expects:
+`read_timeout` for the c10d store, `join_timeout` for its handler, and `timeout` for the
+static backend. Previously only the last of the three was set, leaving the two the c10d
+path reads at torch's own 60 and 600 second defaults.
+
+Every unusable spelling of the timeout env var falls back to the default rather than
+raising, including the non-finite `inf` and `1e999`: converting with `int(float(raw))`
+raises `OverflowError`, which a `(TypeError, ValueError)` handler does not catch, so an
+operator's typo would have turned a training launch into a traceback.
--- a/strands_robots/training/_inproc.py
+++ b/strands_robots/training/_inproc.py
@@ -34,8 +34,11 @@ from __future__ import annotations
 import contextlib
 import io
 import logging
+import math
+import os
+import socket
 import sys
-from collections.abc import Callable, Iterator
+from collections.abc import Callable, Iterator, Mapping
 from typing import Any
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Replaces torch's two undialable rendezvous fallbacks for a single-node elastic launch: the endpoint moves from the literal `localhost:0` to a concrete free port on `127.0.0.1`, and the published `MASTER_ADDR` is pinned to `127.0.0.1` instead of being left to `socket.getfqdn()` (which can resolve to a reverse-DNS PTR name with no forward lookup). A multi-node launch with no explicit endpoint is now refused with the reason rather than silently rendezvousing with itself, and all three rendezvous phase timeouts (`timeout`, `read_timeout`, `join_timeout`) are bounded instead of only the static-backend one. Two operator env overrides (`STRANDS_TRAIN_RDZV_TIMEOUT_S`, `STRANDS_TRAIN_LOCAL_ADDR`) fall back to defaults on every unusable spelling, including the non-finite values that would escape a naive `int(float(raw))` conversion.

## What's good

- Verified against the declared torch floor: torch 2.0's `LaunchConfig` already carries `local_addr` and forwards arbitrary `rdzv_configs` keys via `**config.rdzv_configs` into `RendezvousParameters`, so the new kwargs are compatible across `torch>=2.0`, not just the 2.11 the PR was developed on.
- The one caller that can reach the new multi-node refusal (`cosmos3.py`, `rdzv_endpoint` from `spec.extra`) already converts the raised `ValueError` into an error `TrainResult`, so the refusal surfaces as a result rather than an unhandled traceback - and it replaces a silent hang, which is the correct direction per the raise-on-fatal-errors convention.
- Env-var reads follow the repo's fallback-never-raise pattern for tunables (same shape as `STRANDS_MESH_INPUT_MAX_HZ`), with the `OverflowError` escape from `int(float("inf"))` handled by reading as float and rejecting non-finite values.
- `fqdn` and `port_picker` are injected throughout the tests, so no assertion depends on the runner's DNS or a real bind; the wiring class captures the real `LaunchConfig` rather than trusting the helpers alone. All 35 new tests and the 24 existing `test_inproc.py` tests (including the two real single-node spawns) pass on the head; ruff and the ASCII string scan are clean on both changed Python files; the changelog fragment is present.

## Verification suggestions

Beyond CI: `python -c "from strands_robots.training._inproc import rendezvous_endpoint; print(rendezvous_endpoint('', 1))"` on a host whose `python -c "import socket; print(socket.getfqdn())"` answers a PTR-style name is the environment the fix targets; the two spawn tests in `tests/training/test_inproc.py` drive a real rendezvous through the derived endpoint in a few seconds.

PR #2601, opened 2026-08-22
fix(mesh): pace every publish loop on a deadline, not a delay

cagataycali:
## The defect

Thirteen publish loops in the mesh paced themselves with `self._stop_event.wait(period)`. The shape was chosen for a good reason -- it sleeps for the period **and** returns early when the loop is stopped, so shutdown never waits out a tick -- but a `wait(period)` is a **delay** where a rate needs a **deadline**. The time a tick spends reading a bus or grabbing a frame is added to the period instead of subtracted from it, so a loop achieves `1 / (period + work)` and every counter reports that as the rate the hardware managed.

Measured on Linux against an idle epoll timer, so this is the floor of the problem rather than a platform quirk:

| requested | tick body | `wait(period)` | `Ticker` |
|---|---|---|---|
| 10 Hz | 0 ms | 9.97 Hz | 9.98 Hz |
| 10 Hz | 40 ms | **7.12 Hz** | 10.00 Hz |
| 30 Hz | 10 ms | **22.97 Hz** | 29.99 Hz |
| 50 Hz | 10 ms | **31.92 Hz** | 49.96 Hz |
| 50 Hz | 15 ms | **28.33 Hz** | 49.97 Hz |

A second, platform-specific cause stacks on top. In a process tree that inherits background-QoS timer coalescing -- any robot started by a daemon, a launchd agent or a supervising process -- macOS inflates every `nanosleep`-family wait by ~145 ms, so a nominal 100 ms `Event.wait` costs ~247 ms while a kqueue timer costs ~102 ms, and even a 10 Hz loop with an *empty* body reaches 4.33 Hz. That half is stated as inherited measurement, not re-measured here: this host does not coalesce (`sleep_penalty_s()` reads **0.07 ms**), which is exactly why the Linux table above is the useful evidence for a Linux CI.

## The change

`strands_robots/mesh/pacing.py` adds `Ticker`, which waits on the selector timer:

- **The period is a deadline.** `wait()` sleeps until the next multiple of the period from the loop's start, so a tick's work is subtracted rather than added, and a slow tick does not push every later tick late.
- **Missed deadlines are dropped, not chased.** Catching up would fire several ticks back to back; on a publish loop that means a burst of frames with near-identical timestamps, which is a worse lie about a camera than a gap.
- **A stop is honoured within a slice, not within a period.** The wait is broken into 10 ms selector waits and `wake()` interrupts one immediately, so a 30 Hz loop does not take a third of a second to notice the robot going down. `wait()` keeps `Event.wait`'s sense -- `True` means stop -- so converting a loop cannot invert a shutdown test by accident.
- **`sleep_penalty_s()`** measures what this process tree adds to a plain `sleep`, so a test can calibrate against the host it runs on instead of loosening a number until it passes.

`Ticker` is entered with `with` at all seven call sites rather than a hand-rolled `try` / `finally: ticker.close()`, since it implements the context-manager protocol; and its two bounds go through the package's shared positive-finite numeric domain rather than a hand-rolled `x != x` NaN check, which also refuses a `bool` -- an `int` subclass that would otherwise read as a one-second period -- and refuses it *before* the `float()` coercion that would turn `True` into `1.0`. Apart from that one helper the module imports nothing but the standard library, and nothing at all from the mesh, so it is unit-testable without a session, a robot or a network.

Converted, and the split matters:

| loop | gains rate on Linux | why |
|---|---|---|
| `Mesh._state_loop`, `_heartbeat_loop`, `_camera_loop` | yes | plain `wait(period)`: the read/grab was added to the period |
| the seven `SensorLoopsMixin` loops | yes | same, now through one shared `_paced` generator |
| `HardwareRtpsBridge._poll_loop` | yes | plain `_stop.wait(period)`, no arithmetic at all |
| `InputPublisher._publish_loop` | no | it **already** subtracted its own body (measured 9.99 / 29.94 / 49.84 Hz on Linux) |
| the teleop apply loop | no | same |

For the last two the change is single-ownership of the deadline arithmetic -- one owner instead of a copy per loop that can drift -- plus the coalescing penalty, which no amount of correct arithmetic avoids because it is the wait itself.

The seven sensor loops are converted as one generator rather than seven edits on purpose: they differ only in what they read, so the pacing belongs in a single place where its ownership rules (one ticker per loop, closed even when the body raises) cannot be got right in six and wrong in the seventh.

### `core.py` is in scope, and the inventory test is why

The inventory test scans the whole package for the *shape* rather than for one spelling, and on `upstream/main` it names every pacer:

```
hardware_rtps_bridge.py:375, mesh/core.py:1016, 1093, 1200, mesh/input.py:311,
mesh/sensors.py:180, 253, 339, 394, 442, 488, 525, teleop_mixin.py:860
```

That is 13 sites across 5 files, and the two `test_the_converted_loop_no_longer_paces_on_the_stop_event` / `_camera_loop` cases are parametrized over `core.py` directly -- so `core.py` cannot be left out without the test it ships failing. Only the pacing hunks are taken: `git diff` on `sensors.py`, `input.py`, `teleop_mixin.py` and `core.py` contains no other change, and `teleop_mixin.py` is provably indentation-only apart from 12 added lines (7 comment, 5 ticker) and the 5 removed arithmetic lines.

Two waits are deliberately **not** converted and are listed in the test with the reason each rests on: the one-shot shutdown wait in `Mesh.stop`, and the `except`-path backoff in `HardwareRosBridge._spin_loop` where waiting *longer* than `spin_period` is the intent (verified: the happy path has no wait at all).

## Verification

Measured at `8d8a45bc`, base `9f68f452`. The branch was rebuilt on a moved base after those runs: `mesh/core.py` is the one file the base and this branch both touch, and #2597's `_read_state` probe reporting (`_warn_read_state_once`, four call sites) and the three loop conversions compose cleanly -- verified by resolving both sides' symbols in the merged file, not by the absence of a conflict. The selection was re-run on that base: **21 failed / 6196 passed / 5 skipped / 24 errors** against **21 / 6195 / 5 / 24** pristine, failing ID sets identical (45 each, `comm` empty both ways), `mypy` 24 == 24. The pre-CodeQL-round numbers below were measured at `3fe862d1`; the round's own are at the end of this section.

**Pre-fix split** -- the two new test files plus `pacing.py` copied onto `upstream/main` with no conversions: **12 failed / 25 passed / 1 skipped**, becoming **37 passed / 1 skipped**. The skip is `sleep_penalty_s` doing its job: `sleeps are accurate here (penalty 0.1ms), so Event.wait pacing is not degraded and there is nothing to out-perform` -- the one test that needs a coalescing host declines to assert on a host that is not one, which is what keeps these timing tests non-flaky on Linux CI.

**Every guard is load-bearing.** Each break was applied to the finished tree in isolation:

| break | `test_wait_budget_domain` | inventory |
|---|---|---|
| revert the RTPS poll to `_stop.wait(period)` | 1 failed | 1 failed |
| revert one sensor loop (`_imu_loop`) | -- | 3 failed |
| drop `ticker.close()` from the RTPS loop | 1 failed | -- |
| narrow the inventory regex to `_stop_event` **and** revert the teleop pacer | -- | **17 passed -- missed** |
| keep the wide regex, revert the teleop pacer | -- | 1 failed -- caught |
| control (finished tree) | 98 passed | 135 passed / 1 skipped |

The fourth row is the reason the scan matches a shape rather than a name: the teleop loop's event is `_teleop_stop_event`, so a `_stop_event`-only regex misses it entirely while its eleven siblings look fixed -- which is harder to notice than all twelve being slow.

**Four sibling tests were repaired, and each kept its stated intent.** All four drove a loop by scripting `_stop_event.wait` return values, which is a statement about the pacing call rather than about the guarantee under test; a ticker does not call `wait` on the stop event, and a bare `MagicMock` stop event answers `is_set()` truthily, so the loops ran one tick.

- `tests/mesh/test_camera_loop_resilience.py` -- now uses a real `threading.Event` set by the tick body, so both its guarantees (a transient tick error does not kill the loop; a stop is honoured) are asserted without naming a primitive. The period is read off the ticker the loop builds.
- `tests/mesh/test_mesh_durations_survive_a_clock_step.py` -- the scripted stop event becomes a scripted **ticker**, sitting exactly where the old `wait(period)` sat, so the cadence it drives is unchanged. Its host now borrows the real `_paced` generator, so the test additionally covers the shipped ownership rules.
- `tests/test_inbound_command_costs_one_message.py` -- the loop is bounded by **reader polls** instead of by calls to `is_set()`. A consuming `is_set()` was only safe while exactly one caller asked once per iteration; the ticker asks too, so half the iterations went missing. `is_set()` is now idempotent, as a real event's is.
All three ticker doubles gained `__enter__` / `__exit__` when the call sites moved to `with`, which is the point of a double matching the real interface rather than a subset of it.

- `tests/test_wait_budget_domain.py` -- the RTPS case asserts the stored period is what the loop's ticker is *built* with, that the stop event reaches it, that it paces every iteration and that it is closed. The rclpy half is untouched; its module docstring now names both budget consumers.

**Gate** -- 242 test files touching mesh, teleop, RTPS, pacing and the repo-wide docstring/string/changelog guards, run on the branch and on a pristine `upstream/main` worktree with the identical selection (the two new files excluded from both, so any difference is the source change alone):

| tree | failed | passed | skipped | errors | collected |
|---|---|---|---|---|---|
| branch | 21 | 6184 | 5 | 24 | 6234 |
| pristine `upstream/main` | 21 | 6183 | 5 | 24 | 6233 |

The failing ID sets are **identical** (`comm` empty both ways). All 45 are `tests/mesh/test_iot_*` needing `awsiot`, a declared extra this host lacks. The `+1` reconciles exactly: `tests/mesh/test_discovery_posture_prose.py` is parametrized over every mesh module, so the new `pacing.py` enrols in it automatically and passes.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`. `mypy` **24 == 24** against a pristine worktree, branch-only set empty, none in a changed file.

**Both CodeQL shapes are pinned, not just fixed.** `py/should-use-with` fired at all seven call sites and `py/comparison-missing-self` twice on the hand-rolled `x != x` NaN test (already dead: `not period > 0` refuses NaN). `tests/test_mesh_pacing_ticker.py` now carries the regression cover, so neither shape can return. Against `f1d3d4fa` -- the head CodeQL flagged -- that file splits **17 failed / 54 passed / 1 skipped**, becoming **71 passed / 1 skipped**:

- The release guard scans the package for a `Ticker` built outside a `with`, and pre-fix it names exactly the seven sites CodeQL did: `hardware_rtps_bridge.py:379`, `mesh/core.py:1021/1111/1236`, `mesh/input.py:269`, `mesh/sensors.py:190`, `teleop_mixin.py:796`. An eighth loop cannot hand-roll the release and be caught only by a scanner. A companion case pins the class docstring's example, since that example taught the flagged shape and is how one module arrived at seven copies of it; both directions of the detector are pinned by planted sources, and a non-vacuity case asserts the scan reaches all seven loops rather than reporting a clean sweep from nowhere.
- The domain cases cover the five values the hand-rolled check was narrower than. `True` and `"0.05"` were **accepted** -- a loop built with `Ticker(True)` paces at **1.00 Hz**, measured -- and `None`, a list and `10**400` raised a bare `TypeError` / `OverflowError` out of the conversion rather than the `ValueError` the constructor documents. The set is deliberately the one `test_wait_budget_domain.py` already holds for the bridges' `poll_period`, because it is the same quantity, and one case asserts the equivalence directly so the two cannot come to disagree. The NumPy case pins why the `float()` conversion has to follow the guard rather than precede it.

**This round's gate** -- the same selection (221 files; the four changed test files excluded from both trees, so any difference is the source change alone) run on `bb0e00ba` and on `f1d3d4fa`:

|---|---|---|---|---|
| `bb0e00ba` | 21 | 6851 | 5 | 24 |
| `f1d3d4fa` | 21 | 6851 | 5 | 24 |

Failing ID sets **identical**, 45 each, `comm` empty both ways, all `tests/mesh/test_iot_*` needing `awsiot`. The four changed test files together: **198 passed / 1 skipped**. Each of the four repairs was re-run against the pre-fix source and still passes there, so the doubles gained a protocol rather than losing an assertion.

**The doorbell is a socketpair, not a pipe -- a Windows crash regression, caught in review.** `selectors.DefaultSelector` is `SelectSelector` on Windows and its WinSock `select()` accepts sockets only, so the `os.pipe()` descriptor raised `OSError` (WSAENOTSOCK, 10038) out of the first `Ticker.wait()`. That call sits *outside* the `try` in every paced loop -- the `try` wraps the tick body, so a raise there is logged and the loop continues, while a raise from the pacing unwinds the thread. Each publish thread would have died on its first tick while the mesh still looked up: presence, state, camera and all seven sensor streams stopping on a robot that had joined the fleet.

Every premise was checked rather than taken: `pyproject.toml:27` ships `Operating System :: OS Independent`; `mesh/audit.py` already codes an explicit Windows fallback for `fcntl` at four sites; `test-lint.yml` is `runs-on: ubuntu-latest` with no matrix; and `asyncio.selector_events.BaseSelectorEventLoop._make_self_pipe` is literally `socket.socketpair()`. That last one is why this module's claim to wait on the same primitive as `asyncio` was false with a pipe and is now true.

The `sys.platform` fallback was **not** taken: it would put the one platform with a 145 ms `nanosleep` inflation *and* a coalescing timer back on the primitive this PR replaces, so Windows would keep the achieved-rate bug while needing its own pacing tests to stay honest. The socketpair costs nothing on POSIX and removes the platform split.

The WinSock refusal cannot be reproduced on a POSIX host, where `select()` takes any descriptor, so three cases pin the mechanism rather than the symptom: both doorbell ends are sockets; an AST check refuses an `os.pipe()` call in the module, because the POSIX-only shape passes every behavioural test in the file; and one registers the doorbell with `SelectSelector` directly, the class Windows resolves to. The first two fail against `3fe862d1` -- the second naming the pipe at line 114 -- while the other 73 in that file pass there, so they are specific to this change.

Re-gated after that round on the merged base: the same 221-file selection on `8d8a45bc` and on `d1698859`, the four changed test files excluded from both -- **21 failed / 6863 passed / 5 skipped / 24 errors on each**, failing ID sets identical (`comm` empty both ways, all 45 `awsiot`). The four changed test files: **202 passed / 1 skipped**. `mypy` 24 == 24, branch-only set empty.

## Artifact

`Mesh._camera_loop` asked for 30 Hz with a 10 ms frame grab, driven for one wall-clock second with the transport mocked. The strips are a MuJoCo SO-101 rendered from the poses each loop actually sampled, laid out at their true timestamps -- so the sparser row is literally the motion the loop did not see. Same request, same grab cost, same window in both rows.

| | requested | achieved | frames in 1.0 s | median gap |
|---|---|---|---|---|
| `upstream/main` | 30 Hz | 22.98 Hz (76.6%) | 23 | 43.5 ms = period **+** the 10 ms grab |
| this branch | 30 Hz | 29.99 Hz (100.0%) | 30 | 33.34 ms = the period |

camera loop pacing

## 13. Round changelog

**Round 4 (`dc63cc9`) -- base moved under the branch; one import-anchor conflict resolved.**

#2610, #2598 and #2599 merged to `main` while this PR sat approved-pending. #2598 and #2599 both touch `mesh/input.py`, and #2599 added `from strands_robots.bus_access import write_action` at the same import anchor this branch uses for `from strands_robots.mesh.pacing import Ticker` -- a textual adjacency, not a semantic one. `main` was merged in and both imports kept, in isort order.

The merge was verified by resolving both sides' symbols in the merged file rather than by the absence of further conflicts: `write_action` is live at `mesh/input.py:679` and `:681` (#2599's bus serialization) and `Ticker` at `:270` (this branch's pacing), so neither merged concern was dropped. `mesh/core.py`, `mesh/sensors.py` and `teleop_mixin.py` auto-merged with no conflict.

Re-gated on the moved base. The 10 pinned regression files across both concerns -- this PR's pacing set plus #2598's envelope/slew and #2599's bus-access set -- are **325 passed / 2 skipped**. The wider mesh + teleop + bus selection on this branch is **120 failed / 3650 passed / 19 skipped / 24 errors** against **120 / 3649 / 19 / 24** on `main` at `8c562d3`: the failure count is identical on both trees, so every one of them is pre-existing in this environment rather than attributable to the merge. They sit entirely in `tests/mesh/test_acl_*` and `test_iot_*`, and the 88 collection errors among them name a missing `awsiot` (with `lerobot` missing for the hardware cases) -- declared extras this host lacks. The `+1` passed is this branch's added pacing case. No regression is attributable to the merge.


yinsong1986 (dismissed):
## Summary

Converts all thirteen mesh publish loops from `stop_event.wait(period)` (a delay, which adds each tick's work to the period) to a new dependency-free `mesh.pacing.Ticker` that treats the period as a deadline on the selector timer, drops missed deadlines rather than chasing them, and still honours a stop within a 10 ms slice. The three `Mesh` loops, the seven `SensorLoopsMixin` loops (via one shared `_paced` generator), the RTPS poll loop, and the two loops that already did the subtraction by hand (input publisher, teleop apply) all pace through the one owner; four sibling tests were repaired to assert behaviour rather than the old pacing primitive. I verified the claims independently on the head tree: `Ticker`'s deadline/overrun arithmetic is sound, every converted loop closes its ticker in a `finally` (including the generator path), shipped call paths use each ticker from a single thread so there is no live race, the shutdown joins (2.0 s / 5 s) dwarf the added <=10 ms stop-slice latency, `strands_robots.mesh.pacing` and `hardware_rtps_bridge` import cleanly with mujoco/cv2/zenoh absent (no circular import from the bridge's new module-scope mesh import), and the new + repaired test files plus the repo-wide guards (Raises-docstring completeness, ASCII tool strings, host paths, discovery-posture prose which auto-enrols `pacing.py`) all pass, with `ruff check` clean on the changed files.

## What's good

- The inventory test scans for the shape (`self._*stop*.wait(...)`) rather than one spelling, with the two deliberately-unconverted waits allowlisted with reasons -- exactly the failure mode that hid the teleop pacer is what it pins.
- `Ticker.wait()` keeps `Event.wait`'s return sense, so no conversion can invert a shutdown test; the refused-period domain (zero/nan/inf busy-spin) is checked at construction, ahead of the hardware it would hammer.
- The four repaired tests each kept their stated intent and shed their dependence on the pacing primitive (real `threading.Event` set from the tick body; reader-poll bounds instead of a consuming `is_set()`).
- Timing assertions calibrate against `sleep_penalty_s()` on the host rather than loosening numbers until they pass, and the one test needing a coalescing host skips honestly on Linux.
- Lazy `Ticker` import in `teleop_mixin` keeps that module off the mesh package at import time, with the reason documented, per AGENTS.md's imports-at-top exception.
- Changelog fragment (`changelog.d/2601-mesh-pacing-deadline.md`) present per AGENTS.md > PR Workflow step 3.

## Verification suggestions

- `pytest tests/test_mesh_pacing_ticker.py tests/test_mesh_state_loop_rate.py -q` on a loaded CI runner a few times to confirm the calibrated floors (75% of nominal, 21 fps of 30) hold under contention, since these are the only wall-clock-sensitive tests added to the default suite.
- Spot-check the fd budget claim: `python -c "import resource; from strands_robots.mesh.pacing import Ticker; t=[Ticker(0.1) for _ in range(13)]; [x.close() for x in t]"` -- 3 fds per live ticker (pipe pair + epoll), all released on close.

strands_robots/mesh/pacing.py
```diff
@@ -0,0 +1,253 @@
...
+        period = float(period)
+        slice_s = float(slice_s)
+        self.period = period
+        self.slice_s = min(slice_s, period)
+        self._stop_event = stop_event
+        self._selector = selectors.DefaultSelector()
+        # A self-pipe gives two things: a registered fd (so the selector always
+        # has something to watch, whatever the platform's implementation) and a
+        # way for the stopping thread to interrupt a wait immediately.
+        self._wake_r, self._wake_w = os.pipe()
```
yinsong1986: [MUST FIX] `os.pipe()` registered with `selectors.DefaultSelector` does not work on Windows, and the failure mode is every converted loop's thread dying on its first `wait()`.

On Windows `DefaultSelector` resolves to `SelectSelector`, and WinSock `select()` accepts only sockets -- passing this pipe fd raises `OSError` (WSAENOTSOCK, 10038) on the first `self._selector.select(...)` call in `wait()`. That exception sits *outside* the `try`/`except` in every converted loop (`_state_loop`, `_heartbeat_loop`, `_camera_loop`, all seven `_paced` sensor loops, the input publisher, the teleop apply loop, the RTPS poll), so each thread dies immediately while the mesh itself appears up: presence, state, camera and sensor streams all silently stop.

Why this can't be deferred: it is a crash regression on a platform the package declares support for -- `pyproject.toml` ships the `Operating System :: OS Independent` classifier, and this same subsystem codes an explicit Windows fallback (`mesh/audit.py`: "fcntl is POSIX-only. Windows deployments fall back to in-process...") -- and the CI matrix is `ubuntu-latest` only, so nothing automated will ever surface it. The primitive being replaced (`Event.wait`) works everywhere; after this ships, mesh-on-Windows failures present as a robot that joins the fleet and publishes nothing, which is the hardest shape of this bug to attribute.

What would resolve it: use `socket.socketpair()` for the doorbell instead of `os.pipe()` (this is exactly why asyncio's selector event loop uses a socketpair for its self-pipe on Windows -- the module docstring's "same primitive asyncio sleeps on" claim only holds with the socketpair), with `setblocking(False)` on the read end and `send`/`recv` in `wake()`/`_drain()`. Alternatively, gate on `sys.platform == "win32"` and fall back to sliced `Event.wait` pacing there -- less accurate, but it degrades to today's behaviour instead of killing the thread.

yinsong1986 (commented):
## Summary

Converts all thirteen mesh publish loops from `stop_event.wait(period)` -- a delay that adds each tick's work to the period -- to a new stdlib-only `mesh.pacing.Ticker` that treats the period as a deadline on the selector timer, drops missed deadlines rather than firing catch-up bursts, and honours a stop within a 10 ms slice. The three `Mesh` loops, the seven `SensorLoopsMixin` loops (through one shared `_paced` generator), `InputPublisher`, the teleop apply loop and the RTPS command poll are converted; two waits are deliberately left alone with recorded reasons. Verified locally at `3fe862d1` (merge base `71735507`, diff matches the description exactly at 13 files / +1290 / -174): `tests/test_mesh_pacing_ticker.py` (20 passed / 1 calibrated skip) and the four repaired sibling suites plus the two new rate/inventory suites (167 passed) are green, and the new files are clean on the ASCII and host-path guards. The `Ticker` bounds go through the shared `positive_finite_number_error` domain, every call site enters the ticker with `with`, and the prior CodeQL findings from `f1d3d4fa` (nan self-comparison, close-in-finally) are all resolved at this head.

## What's good

- The inventory test scans for the *shape* (`.wait(...)` on any stop-flag-named attribute) rather than one spelling, with an explicit allowlist carrying reasons -- the teleop `_teleop_stop_event` near-miss it guards against is documented in the test itself.
- Timing tests calibrate against `sleep_penalty_s()` on the host rather than loosening numbers until they pass, and the one comparison that needs a coalescing host skips honestly elsewhere.
- The four repaired sibling tests each kept their stated intent while dropping the pacing-primitive coupling, and every test double carries the full context-manager protocol of the real `Ticker`.
- Changelog fragment present per AGENTS.md step 3; period validation refuses `bool` before `float()` coercion per the shared numeric-domain convention.

cagataycali:
All nine alerts resolved by writing the code they asked for, in `3fe862d1` (the change) and `bb0e00ba` (the regression cover). Neither is a dismissal.

**`py/should-use-with` x7** -- `Ticker` already defined `__enter__` / `__exit__`, and every one of the seven loops released it in a hand-rolled `try` / `finally: ticker.close()`. Entered with `with` at all seven. The root cause was upstream of the call sites: the class docstring *taught* the hand-rolled form, which is how one module arrived at seven copies of it at once, so the example moved too.

**`py/comparison-missing-self` x2** -- the flagged expression was the hand-rolled `period != period` / `slice_s != slice_s` NaN test, and it was already dead: `not period > 0` refuses NaN, because `nan > 0` is False. Rather than respell it, both bounds now go through `positive_finite_number_error`, the domain the surfaces that hand a period down already use for this quantity -- `HardwareRtpsBridge.poll_period`, a loop's `1 / hz`. That was the more interesting half, because the check written here instead was narrower than the values that reach it:

| value | before | after |
|---|---|---|
| `True` | **accepted**, `period=1.0` -- a loop built with it paces at **1.00 Hz** (measured) | refused, naming the value |
| `"0.05"` | **accepted** | refused |
| `None`, `[0.05]` | bare `TypeError` from `float()` | refused as `ValueError`, which is what the constructor documents |
| `10**400` | bare `OverflowError` | refused, and told it is past the float64 range |
| `0`, `-1`, `nan`, `inf`, `-inf`, `False` | refused | refused (unchanged) |

`True` is the one worth naming: it is an `int` subclass, so a bare positivity test admits it, and this branch's own `test_wait_budget_domain.py` already calls that case `test_a_boolean_period_is_a_silent_one_second_cadence` at the bridge constructor -- while the ticker one frame down re-derived a check that accepted it. Two tests of one quantity is how they come to disagree, so there is now one. The `float()` coercion moved *after* the guard, since the domain accepts a NumPy scalar the selector cannot be handed.

**Pinned rather than just fixed.** Against `f1d3d4fa` -- the head the alerts were raised on -- `tests/test_mesh_pacing_ticker.py` splits **17 failed / 54 passed**, becoming 71 passed. The release guard scans the package for a `Ticker` built outside a `with` and pre-fix names *exactly* the seven sites CodeQL did (`hardware_rtps_bridge.py:379`, `mesh/core.py:1021/1111/1236`, `mesh/input.py:269`, `mesh/sensors.py:190`, `teleop_mixin.py:796`), so an eighth loop cannot hand-roll the release and be caught only by a scanner; both directions of the detector are pinned by planted sources, and a non-vacuity case asserts the scan reaches all seven rather than reporting a clean sweep from nowhere.

**Four ticker doubles gained `__enter__` / `__exit__`**, because the loops now acquire with `with` and a double that models only the methods it is asked for is not reached at all. Each was re-run against the pre-fix source and still passes there, so they gained a protocol rather than losing an assertion -- and `_ScriptedTicker` / `_RecordingTicker` keep their `closed` assertion, which now proves the block releases the ticker. Only the full blast radius found two of the four: a scoped run of the file I was editing was green.

**Gate** -- 221 files, the same selection on `bb0e00ba` and on `f1d3d4fa` with the four changed test files excluded from both, so any difference is the source change alone: **21 failed / 6851 passed / 5 skipped / 24 errors on each**, failing ID sets identical (`comm` empty both ways, all 45 `tests/mesh/test_iot_*` needing `awsiot`, a declared extra this host lacks). The four changed test files together: 198 passed / 1 skipped. `ruff check` / `ruff format --check` clean; `mypy` 24 == 24 against a pristine worktree with the branch-only set empty.

strands_robots/mesh/pacing.py
cagataycali: Confirmed, and taken as written. Fixed in `8d8a45bc` with the socketpair -- your first option, because it keeps the accuracy the module exists for rather than trading it away on one platform.

I verified each premise before acting on it rather than after:

- `pyproject.toml:27` ships `Operating System :: OS Independent`, so Windows is declared rather than incidental.
- `mesh/audit.py` already codes the fallback for `fcntl` at four sites, so this subsystem has met the platform before.
- `test-lint.yml` is `runs-on: ubuntu-latest` with no matrix, so nothing automated reaches it.
- `asyncio.selector_events.BaseSelectorEventLoop._make_self_pipe` is literally `self._ssock, self._csock = socket.socketpair()`, comment included ("A self-socket, really. :-)"). That is the sharpest part of your point: the module docstring claimed to wait on the same primitive asyncio sleeps on, and with `os.pipe()` that claim was false in exactly the way that mattered. It is now true, and the docstring says why.
- `ticker.wait()` is outside the `try` at every converted site -- the `try` wraps the tick body, so a raise there is logged and the loop continues, while a raise from the pacing unwinds the thread. Your failure mode is right: the mesh stays up, the streams stop.

The change is `socket.socketpair()` with both ends non-blocking, `send` / `recv` for the doorbell and `close()` on both. `_drain`'s post-condition is unchanged, since `recv` on a drained non-blocking socket raises `BlockingIOError` exactly as the pipe read did, and a closed socket still lands in the `OSError` arm.

I did not take the `sys.platform` fallback. It would put the one platform with a 145 ms `nanosleep` inflation *and* a coalescing timer back on the primitive this module replaces, so Windows would keep the achieved-rate bug the PR is for, and the branch would then need its own pacing tests to stay honest. The socketpair costs nothing on POSIX and removes the platform split entirely.

**On testing it.** The WinSock refusal cannot be reproduced on a POSIX host -- `select()` there takes any descriptor -- so I pinned the mechanism rather than asserting the symptom, and said so in the test docstring:

- `test_both_ends_of_the_doorbell_are_sockets` -- what makes WinSock `select()` accept it.
- `test_the_module_does_not_build_the_doorbell_from_a_pipe` -- an AST check, because the POSIX-only shape passes every behavioural test in the file, so nothing else would notice a revert.
- `test_the_doorbell_is_accepted_by_the_selector_windows_resolves_to` -- registers the doorbell with `SelectSelector` directly rather than `DefaultSelector` (epoll here), and asserts a `wake()` is visible through it.

Both of the first two fail against `3fe862d1`, the second naming the pipe at line 114, while the other 73 in that file pass there -- so the guards are specific to this change rather than to the branch.

Verified on `8d8a45bc`, which also merges `upstream/main`: the four changed test files 202 passed / 1 skipped; the 221-file mesh/teleop/RTPS/pacing selection run on `8d8a45bc` and on `d1698859` with those four excluded gives **21 failed / 6863 passed / 5 skipped / 24 errors on each**, failing ID sets identical (`comm` empty both ways, all 45 `tests/mesh/test_iot_*` needing `awsiot`). `ruff check` / `ruff format --check` clean; `mypy` 24 == 24 with the branch-only set empty.

yinsong1986 (approved):
## Summary

Converts all thirteen mesh publish loops from `stop_event.wait(period)` -- a delay that adds each tick's work to the period -- to `mesh.pacing.Ticker`, a stdlib-only pacer that treats the period as a deadline on the selector timer, drops missed deadlines rather than firing catch-up bursts, and honours a stop within a 10 ms slice while keeping `Event.wait`'s return sense. The three `Mesh` loops, the seven `SensorLoopsMixin` loops (via one shared `_paced` generator), the RTPS command poll, the input publisher and the teleop apply loop are all converted; the two waits deliberately left alone (one-shot shutdown wait, exception-path backoff) are listed in the inventory test with reasons. I verified at this head that the concerns raised in prior rounds are resolved: the doorbell is a `socket.socketpair()` (with an AST guard refusing `os.pipe()`), every `Ticker` is acquired with `with` (release guard scans the package), and the period/slice bounds go through the shared `positive_finite_number_error` domain ahead of the `float()` coercion, with parity to `test_wait_budget_domain.py` asserted as an equivalence.

## What's good

- The inventory test matches the shape (`.wait(...)` on any stop-flag-named attribute) rather than one spelling, which is what caught the `_teleop_stop_event` pacer -- and the two exempted waits carry their reasons inline.
- Timing assertions calibrate against `sleep_penalty_s()` on the host rather than hard-coding numbers, and the one test that needs a coalescing host skips honestly on an accurate one.
- The four repaired sibling tests kept their stated intent while gaining the ticker's context-manager protocol; ticker doubles match the real interface.
- `Ticker` uses `time.perf_counter()` throughout, consistent with the AGENTS.md clocks rule (a duration is measured on a monotonic clock).
- Changelog fragment present; repo-wide guards (ASCII tool strings, host paths, `Raises:` completeness, test naming) pass over the changed files; imports verified cycle-free (`hardware_rtps_bridge` -> `mesh.pacing` triggers no circular import, and zenoh stays lazy).

## Verification suggestions

- `python -m pytest tests/test_mesh_pacing_ticker.py tests/test_mesh_state_loop_rate.py tests/test_wait_budget_domain.py tests/mesh/test_camera_loop_resilience.py tests/mesh/test_mesh_durations_survive_a_clock_step.py tests/test_inbound_command_costs_one_message.py -q` -- 242 passed / 1 skipped locally at this head.
- Spot-check the round-4 merge kept both sides in `mesh/input.py`: `write_action` at the bus-serialization call sites and `Ticker` at the pacing import both resolve.

PR #2603, opened 2026-08-22
fix(sim): the symlink refusal names where the link points

cagataycali:
## Problem

`validate_output_path` refuses to follow a symlink planted at an output target, which is right: it is an arbitrary-write vector. But the refusal named no way forward:

```
output_dir '/tmp' is a symlink - refusing to follow
```

On macOS the single most reachable scratch path IS a symlink - `/tmp` points at `/private/tmp` - so a caller that asked for the most ordinary directory on the machine got a dead end that read like an attack had been detected. The reachable surface is the two `label="output_dir"` sinks (`start_cameras_recording` / `..._synchronous`), where the target is a directory rather than a file inside one.

Measured through the real sinks (a directory reached through a symlink, guards-only mode, which is the recording default):

| sink | before |
| --- | --- |
| `render(output_path=)` | refused, no path named, 0 bytes written |
| `start_cameras_recording(output_dir=)` | refused, no path named, 0 MP4 |

## Change

The refusal names the destination, and when this same call would accept that destination it is offered as the path to pass. The link itself is still never followed - `resolve()` is a read.

Two properties the hint has to hold, both of which a naive version of this fix gets wrong.

**It must not raise.** Naming the destination means resolving it while a refusal is being built, and `Path.resolve(strict=False)` is not total. Measured on both supported interpreters:

| link | CPython 3.12.3 | CPython 3.13.12 |
| --- | --- | --- |
| dangling target | returns the target, no raise | returns the target, no raise |
| cycle | raises `RuntimeError("Symlink loop from ...")` | returns the link unresolved, no raise |

`requires-python = ">=3.12"` admits both, so a handler naming only `OSError` catches neither, and on 3.12 a `RuntimeError` escapes a function documented to raise `ValueError` - past the `except ValueError` every sink maps to a structured tool error. The handler names `(OSError, RuntimeError)`, and the reasoning is in the helper's docstring rather than left to be re-derived.

**It must not advertise a path this same call would refuse.** Under active confinement (the `render` sink's default) a link inside the sandbox pointing outside resolves to a path the confinement check then rejects, so offering it would send the caller straight into a second refusal. That case names the destination as a diagnosis and points back into the sandbox instead. Naming a resolved path in a refusal is already this function's convention - the confinement refusal one branch below does the same.

After, through the same two sinks, following the remedy the message gives:

| sink | mode | after |
| --- | --- | --- |
| `render(output_path=)` | confined | offered `.../shots/frame.png`, retry `success`, PNG 46089 bytes |
| `start_cameras_recording(output_dir=)` | guards-only | offered the real directory, retry `success`, 1 MP4 / 40513 bytes / 16 frames |

Also measured and left as-is: a dangling link does get a hint, because `resolve(strict=False)` returns its target and a not-yet-existing path is exactly what an output sink is about to create - verified accepted on retry.

## Tests

Extends `tests/simulation/test_output_path_sandbox.py` beside the existing `test_rejects_symlink_target`. Split against `upstream/main` with only the test file copied in: **5 failed / 54 passed** -> 59 passed.

The value of the other four is what they do to a naive form of this fix (catch `OSError` only, always offer the destination). Applied verbatim to `upstream/main` and run against these tests:

| tree | failing | which |
| --- | --- | --- |
| `upstream/main` | 5 | the five that need a destination named |
| naive form | 3 | `..._cannot_resolve_is_still_a_value_error[RuntimeError]`, `..._real_symlink_loop_is_refused_without_a_hint`, `..._pointing_outside_is_named_but_not_offered` |
| this branch | 0 | - |

`test_a_symlink_the_call_cannot_resolve_is_still_a_value_error` drives `RuntimeError` and `OSError` through a patched `resolve`, so it holds on either interpreter rather than only the one running it. `test_the_path_the_symlink_refusal_offers_is_one_this_call_accepts` parses the destination back out of the message and passes it, which is what a caller acting on the refusal does.

## Verification

Measured at `aad8f2d9`. The head has since moved to `df1674e`, which adds two assertions to the headline test (that `refusing to follow` still appears, so naming a way forward is an addition to the verdict rather than a replacement) and changes no source line; the file still reports 59 passed and the pre-fix split is unchanged at 5 failed / 54 passed.

Blast radius: 108 test files (every test touching the sandbox helpers or the two sinks, plus the repo-wide docstring / string / changelog guards), the changed test file excluded from both trees so any difference is the source change alone.

| tree | result |
| --- | --- |
| this branch | 4 failed, 4812 passed, 4 skipped |
| `upstream/main` (pristine) | 4 failed, 4812 passed, 4 skipped |

Failing ID sets identical, `comm` empty in both directions. All 4 are pre-existing and environmental: `test_rendering_pacers_survive_a_clock_step[forward_30s]` passes 16/16 when that file runs alone on the pristine tree (order-dependent), and the three `tests/training` failures are `draccus 0.10.0` against `lerobot 0.6.2`'s `draccus>=0.11.6` floor.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`. `mypy` 24 errors on this branch and 24 on a pristine `upstream/main`, identical sets, none in the changed files.

## Artifact

Both columns run the same capture script against the same two real sinks; only `strands_robots` differs. Each column follows whatever remedy its own tree's message gives, so the right-hand frame is the one a caller obtains by acting on the refusal, and the left-hand panel is what the call produced instead.

symlink refusal names its destination

### DIFF
--- a/changelog.d/2603-symlink-refusal-names-destination.md
+++ b/changelog.d/2603-symlink-refusal-names-destination.md
@@ -1,0 +1,30 @@
+### Fixed
+
+The symlink refusal in `validate_output_path` now names where the link points,
+and hands over that path when this same call would accept it. Refusing to follow
+a link planted at an output target is right -- it is an arbitrary-write vector --
+but the refusal named no way forward, and on macOS the single most reachable
+scratch path IS a symlink: `/tmp` points at `/private/tmp`. A caller that asked
+for the most ordinary directory on the machine got a dead end that read like an
+attack had been detected. The reachable surface is the two `label="output_dir"`
+sinks, where the target is a directory rather than a file inside one: measured
+through `start_cameras_recording(output_dir=<a directory reached through a
+symlink>)`, the call refused and wrote nothing, and following the corrected
+message writes the recording (1 MP4, 16 frames).
+
+The link itself is still never followed; `resolve()` is a read.
+
+Two properties the hint holds. It does not raise: naming the destination means
+resolving it while a refusal is being built, and `Path.resolve(strict=False)` is
+not total -- CPython 3.12 raises `RuntimeError("Symlink loop from ...")` for a
+cycle while 3.13 returns the link unresolved, and `requires-python = ">=3.12"`
+admits both. A handler naming only `OSError` would catch neither, letting 3.12's
+`RuntimeError` escape a function documented to raise `ValueError`, past the
+`except ValueError` every sink maps to a structured tool error. And it does not
+advertise a path this same call would refuse: under active confinement a link
+pointing outside the sandbox has its destination named as a diagnosis rather than
+offered, so a caller who follows the message cannot land in a second refusal.
+
+A dangling link still gets a hint. `resolve(strict=False)` returns its target
+without raising, and a not-yet-existing path is exactly what an output sink is
+about to create.
--- a/strands_robots/simulation/safe_output.py
+++ b/strands_robots/simulation/safe_output.py
@@ -128,6 +128,60 @@ def _confinement_refusal(
     )
 
 
+def _symlink_destination_hint(raw: Path, *, sandbox_root: Path | None, allow_abs: bool) -> str:
+    """Return a suffix naming where a refused symlink points, or ``""``.
+
+    Refusing to follow a link planted at the target is right - it is an
+    arbitrary-write vector - but the refusal has to leave the caller somewhere
+    to go, and on macOS the single most reachable scratch path IS a symlink:
+    ``/tmp`` points at ``/private/tmp``. A caller (often an LLM) that asked for
+    the most ordinary directory on the machine got a refusal that read like an
+    attack had been detected and named no way forward. Naming the destination is
+    the missing step; the link itself is still never followed.
+
+    Two things this must not do.
+
+    It must not raise. Building a refusal is part of answering the caller, and
+    ``Path.resolve(strict=False)`` is not total: on CPython 3.12 a symlink cycle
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Message-only fix to `validate_output_path`'s symlink refusal: the link is still never followed, but the refusal now names where it points, and offers that destination as the path to pass only when this same call would accept it (under active confinement an outside-sandbox destination is named as a diagnosis and the caller is pointed back into the sandbox instead). The hint is built inside a `try` scoped to the single `Path.resolve(strict=False)` call, with `except (OSError, RuntimeError)` covering the measured 3.12/3.13 divergence on symlink cycles, so the function's documented `ValueError` contract - which every sink maps to a structured tool error - is preserved on both supported interpreters. No API, wire, or default-behaviour change; the diff is the helper, a message interpolation, six behaviour-named tests, and a changelog fragment.

## What's good

- The exception-class analysis is measured rather than assumed, and the cycle case is pinned twice: once through a patched `resolve` driving both classes (so the guard holds on either interpreter), and once with a real on-disk loop.
- The offered-path test parses the destination back out of the message and retries it, which is exactly what a caller acting on the refusal does - so "never offer a path this call would refuse" is asserted as behaviour, not as prose.
- `except (OSError, RuntimeError)` has no covered member (per AGENTS.md > Exception Clauses Must Be Narrow), and the `try` covers only the operation whose exception it classifies.
- Changelog fragment in `changelog.d/`, ASCII-only user-facing strings, no host paths, `tmp_path` roots throughout, and naming a resolved path in a refusal follows the pre-existing `_confinement_refusal` convention.

Issue #384, opened 2026-06-12
tech-debt(gr00t-volume-safety): v0.4.1 follow-up bundle - /home blocklist granularity + realpath symlink resolution (deferred from #372)


cagataycali:
Two **non-blocking** findings raised during the approving review of #372. Both were explicitly deferred to follow-up by the approver so the security-hardening PR could merge clean within its round budget (#372 reached R7 / 21 commits; the author's own round-budget note recommends lifting any further variant into a fresh PR rather than adding an R8). Tracking here in the v0.4.1 bundle pattern (cf. #383, #378, #367).

## 1. `/home` wholesale block may reject legitimate `hf_local_dir` on Linux

`gr00t_inference.py` (`_check_volume_safety` + `_BLOCKED_VOLUME_HOST_PATHS`) blocks the entire `/home` prefix. On Linux `$HOME=/home/<user>`, so an explicit operator-supplied `hf_local_dir="~/checkpoints"` / `"/home/ubuntu/ckpts"` is rejected. The `None` default path is not validated, so the happy path (auto-derived `~/.cache/huggingface`) still works.

**Question to resolve:** is blocking all of `/home` intentional (defence-in-depth: assume any home dir may hold credentials), or should the block be narrowed to the actual sensitive subpaths (`~/.ssh`, `~/.aws`, `~/.config`, other users' homes) while permitting an operator's explicit checkpoint dir under their own `$HOME`?

**Acceptance criteria:**
- Decision documented inline (NOTE comment near `_BLOCKED_VOLUME_HOST_PATHS`) and in README Configuration.
- If narrowed: pin tests for `~/.ssh`/`~/.aws`/`~/.config` (still blocked) and `~/checkpoints` (now allowed); if kept wholesale: a pin test + doc line stating operators must place checkpoints outside `/home`.

## 2. Volume guard uses `normpath`, not `realpath` (symlink resolution gap)

`_normalize_host_path` canonicalises leading slashes and runs `os.path.normpath`, but does not resolve symlinks. A pre-existing symlink on the host pointing at `/etc` (or another protected dir) passes `_check_volume_safety` while docker mounts the resolved target.

**Exploitability: low** - the gr00t tool gives the agent no primitive to create the symlink; this requires a pre-existing malicious/misconfigured symlink on the host. Flagged for completeness of the defence-in-depth claim.

**Acceptance criteria:**
[...]

PR #2604, opened 2026-08-22
fix(tools): refuse a protected directory under every name that reaches it

cagataycali:
## Problem

`_check_volume_safety` resolves the candidate mount through symlinks (#384 item 2) and compared it against a **literal** blocklist, so it compared a *directory* against a *name*. Docker mounts the directory, so a protected directory reachable under a second name was refused under one spelling and admitted under the other.

Measured on a host whose protected directory is reached through a symlink whose target escapes the blocklist (`<host>/etc -> <host>/private/etc`, blocklist naming `<host>/etc`) - the shape macOS ships as `/etc -> /private/etc`, and that a Linux server with a separate data volume ships as `/home -> /mnt/home`:

| mount | before | after |
| --- | --- | --- |
| `<host>/etc` | refused | refused |
| `<host>/etc/shadow` | refused | refused |
| `<host>/private/etc` | **allowed** | refused |
| `<host>/private/etc/shadow` | **allowed** | refused |
| `<host>/safe` | allowed | allowed |

The admitted spelling is the same directory, so row 4 is the file the blocklist exists to protect, reached under the name it does not check.

`_check_hf_local_dir_safety` delegates to this helper, and `hf_local_dir` is an **agent-supplied** string that reaches a host write with no docker mediation (`_download_checkpoint` writes the snapshot there directly). Driven through that path on the same layout, the resolved spelling and its children were admitted before and are refused after, so the reachable surface was an agent-named write into a protected directory rather than only an operator bind mount.

## Change

Resolve both sides of the comparison:

```python
blocked_dirs = _with_resolved(_BLOCKED_VOLUME_HOST_PATHS)
blocked_exact = _with_resolved(_BLOCKED_VOLUME_EXACT)
```

`_with_resolved` contributes each protected path **and** the path its symlinks resolve to. A path that is not a symlink contributes one entry, so on a host whose protected paths are real directories the returned set is the blocklist and every verdict is the one it was.

Measured on this Linux host: 15 literal entries become 18, the three added being `/usr/bin`, `/usr/sbin`, `/usr/lib` from `/bin`, `/sbin`, `/lib` - **each already covered by the literal `/usr`** - while `/var/run -> /run` adds nothing because `/run` is already an entry, and the exact-match set is unchanged. So the effective coverage here is identical; what changes is a host where a protected path resolves outside every prefix.

## Notes on scope

The issue this came from reported the headline as a mount of `/etc` being allowed because it resolves to `/private/etc`. That is **not** what happens: `candidates` holds both `norm` and `resolved`, so the symlink spelling matches the blocklist by name and is refused (row 1 above, before and after). The defect is the *target* spelling, which the issue mentions only in passing, and its children.

The same issue proposed a second piece - carving the per-user temp root out of the blocklist, because macOS resolves `tempfile.gettempdir()` to `/private/var/folders/<...>/T`, i.e. under `/var`. That is deliberately **not** here. It is a widening of a security blocklist rather than a fix to a one-sided comparison, it is a no-op on a host where `gettempdir()` is `/tmp` (not under any protected prefix), and `gettempdir()` reads `TMPDIR`/`TMP`/`TEMP` and honours any *writable* candidate: measured, `TMPDIR=/dev/shm`, `TMPDIR=/var/tmp` and `TMPDIR=/run/user/<uid>` are all honoured on this host, each of which would unblock a subtree of a protected prefix (`/dev`, `/var`, `/run`) through an environment variable that is not the documented opt-out. Whether an operator-facing carve-out should exist, and what should name it, is a contract choice rather than a defect - worth its own change with its own justification.

## Tests

New `TestTheBlocklistIsResolvedOnBothSides` in the existing hardening file, so the `#384 item 2` tests beside it stay as controls.

**4 failed / 99 passed** against `upstream/main`, **103 passed / 1 skipped** after. The four are behavioural, each naming the admitted spelling:

- `test_the_resolved_spelling_of_a_protected_directory_is_refused` - `... is the directory ... resolves to, and docker mounts that directory, but the blocklist admitted it: None`
- `test_a_child_of_the_resolved_spelling_is_refused`
- `test_an_agent_supplied_hf_local_dir_cannot_name_the_resolved_spelling`
- `test_an_exact_match_entry_is_refused_under_its_resolved_name`

Five pass on both trees and are the controls, each failing for a specific wrong fix:

- `test_a_blocklist_of_real_directories_changes_no_verdict` - the no-op guarantee; fails if resolution ever adds an entry where nothing is a symlink.
- `test_a_symlink_into_a_safe_directory_is_still_allowed` - fails if resolving the blocklist starts refusing an unprotected target.
- `test_every_shipped_blocklist_entry_refuses_its_own_resolved_path` - parametrized over the shipped blocklist, so an entry added later is covered without anyone remembering this. It already held on a host whose symlinks resolve inside another blocked prefix, which is this host; it is the invariant statement rather than regression evidence here.
- `test_the_layout_under_test_really_escapes_the_blocklist` - premise, so the headline cannot pass vacuously.
- The pre-existing `test_check_volume_safety_allows_symlink_into_safe_dir` and `..._resolves_symlink_into_protected_dir`.

## Artifact

The agent-supplied sink, driven end to end on the synthetic host. Both columns run the same capture script against the same layout, with `huggingface_hub.snapshot_download` stood in by a function that writes into `local_dir` exactly as the real one does - `local_dir.mkdir(parents=True, exist_ok=True)` and the guard call above it are shipped code, and only `strands_robots` differs between the columns.

the blocklist compared a directory against a name

Before, `hf_local_dir=<host>/private/etc` returns `status="success"` with `Downloaded nvidia/GR00T-N1.5-3B -> <host>/private/etc`, and `model.safetensors` lands **inside the protected directory, beside `shadow`**. After, it is refused and the directory is untouched. Row 1 is the control: the symlink spelling was refused in both columns and wrote nothing.

## Verification

Measured at `df9a436`.

Blast radius - every test naming the volume/`hf_local_dir` guards or the blocklist, all of `tests/tools`, and the repo-wide docstring, string, changelog and dependency guards. 68 files, run on this branch and on a pristine `upstream/main` with the changed test file excluded from both, so any difference is caused by the source change alone:

| tree | passed | failed | skipped |
| --- | --- | --- | --- |
| this branch | 3324 | 0 | 9 |
| `upstream/main` | 3324 | 0 | 9 |

Identical, and zero failures on either side.

`tests/tools/test_gr00t_container_hardening.py` separately: **4 failed / 99 passed / 1 skipped** on `upstream/main` with this file copied in, **103 passed / 1 skipped** here.

`mypy strands_robots tests tests_integ`: 24 pre-existing on both trees, branch-only empty, none in the changed files. `ruff check` and `ruff format --check` clean over the CI scope.

### DIFF
--- a/changelog.d/2604-blocklist-resolved-on-both-sides.md
+++ b/changelog.d/2604-blocklist-resolved-on-both-sides.md
@@ -1,0 +1,25 @@
+### Fixed
+
+`gr00t_inference` now refuses a protected host directory under every name that
+reaches it. `_check_volume_safety` resolved the candidate mount through symlinks
+and compared it against a literal blocklist, so it compared a directory against
+a name. Docker mounts the directory, so a protected directory reachable under a
+second name was refused under one spelling and admitted under the other. On a
+host whose protected directory is reached through a symlink whose target escapes
+the blocklist - macOS ships `/etc -> /private/etc`, and a Linux server with a
+separate data volume may ship `/home -> /mnt/home` - the symlink spelling was
+refused and the target spelling, along with every file beneath it, was admitted.
+
+The admitted spelling reached two sinks, and the second is the one that matters:
+`_check_hf_local_dir_safety` delegates to the same helper, and `hf_local_dir` is
+an agent-supplied string that `_download_checkpoint` writes to on the host with
+no docker mediation. So the reachable surface was an agent-named write into a
+protected directory, not only an operator bind mount.
+
+Both sides of the comparison are resolved now. A protected path that is not a
+symlink contributes one entry, so on a host whose protected paths are real
+directories the set is the blocklist and every verdict is the one it was; where
+`/bin`, `/sbin` and `/lib` are symlinks into `/usr`, the three entries they add
+were already covered by `/usr`. The invariant is stated for every entry in the
+shipped blocklist, parametrized, so an entry added later is covered without
+anyone remembering this.
--- a/strands_robots/tools/gr00t_inference.py
+++ b/strands_robots/tools/gr00t_inference.py
@@ -293,6 +293,28 @@ def _resolve_host_path(host_path: str) -> str:
     return os.path.realpath(os.path.expanduser(host_path))
 
 
+def _with_resolved(paths: tuple[str, ...]) -> set[str]:
+    """Return each protected path AND the path its symlinks resolve to.
+
+    The blocklist names a path, but docker mounts the directory that path
+    resolves to, so a protected directory reachable under two names has to be
+    blocked under both. A host that reaches a protected directory through a
+    symlink whose target escapes the blocklist is the case this covers: macOS
+    ships ``/etc -> /private/etc``, and a Linux server with a separate data
+    volume may ship ``/home -> /mnt/home``. Resolving only the candidate mount
+    (#384 item 2) refuses the symlink spelling and admits the target spelling,
+    which is the same directory.
+
+    A path that is not a symlink contributes one entry, so on a host whose
+    protected paths are all real directories the returned set is the blocklist.
+    """
+    out: set[str] = set()
+    for raw in paths:
+        out.add(os.path.normpath(raw))
+        out.add(os.path.realpath(raw))
+    return out
+
+
 def _check_volume_safety(volumes: dict[str, str] | None) -> str | None:
     """Return None if all bind-mount host paths are safe, else a reason.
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes the one-sided comparison in `_check_volume_safety`: the candidate mount was resolved through symlinks (#384 item 2) but compared against a *literal* blocklist, so a protected directory reachable under a second name (`/etc -> /private/etc` on macOS, `/home -> /mnt/home` on a split-volume Linux host) was refused under the symlink spelling and admitted under the target spelling - and the target spelling is the same directory docker actually mounts. The fix resolves the blocklist side too via `_with_resolved`, which always retains the literal `normpath` entry, so the refusal set can only widen - no fail-open direction exists in this change, and an entry resolving to `/` falls into the existing exact-match-only root branch rather than refusing every absolute path. The exposure through `_check_hf_local_dir_safety` (agent-supplied `hf_local_dir` reaching a direct host write in `_download_checkpoint`) is covered by the same fix and pinned by its own test.

## What's good

- Regression pins verified independently: the four behavioural tests fail on `upstream/main` (4 failed / 99 passed) and the full hardening file passes on the head (103 passed / 1 skipped), matching the PR description exactly.
- Controls are precise: the no-op guarantee on a symlink-free blocklist, the safe-symlink-still-allowed case, and a premise test that keeps the headline from passing vacuously. The per-entry parametrized invariant covers blocklist entries added later without anyone remembering this fix.
- Scope discipline: the temp-dir carve-out proposed alongside the original issue is correctly excluded as a blocklist *widening* with its own `TMPDIR`-override hazard, and the PR description documents the measurement behind that call.
- Changelog fragment present per AGENTS.md step 3; tests use `tmp_path`/`monkeypatch.setattr`, no host paths, ASCII-only additions; lint and format clean on the changed files.

## Verification suggestions

- Spot-check the before/after in one command from a checkout: `git stash-free` equivalent - check out `upstream/main -- strands_robots/tools/gr00t_inference.py` on the PR branch and run `python -m pytest tests/tools/test_gr00t_container_hardening.py -q --no-cov`; expect exactly the four named behavioural failures, then restore and expect green.
- On a macOS box (the layout the fix is motivated by), confirm `gi._check_volume_safety({"/private/etc": "/x"})` is non-None against the shipped blocklist.

PR #2605, opened 2026-08-22
fix(sim/mujoco): a render REFUSAL is not a camera that is merely "cold"

cagataycali:
## Problem

Two reporting defects on the same path: the renderer could not build a GL context.

**1. The advice was one sentence for every host.** `render`, `render_depth` and `get_frame` each carried their own copy of *"Install EGL or OSMesa for offscreen rendering: apt-get install libosmesa6-dev"*. macOS has neither EGL nor OSMesa - MuJoCo renders through CGL there, and there is no apt - so a Mac operator was sent to install a package that does not exist for their machine, and while they chased it the real cause kept its cover. Measured by asking the same tree twice with `sys.platform` set each way:

| `sys.platform` | text `render` returns on `upstream/main` |
|---|---|
| `linux` | `Rendering unavailable (no OpenGL context). Install EGL or OSMesa ... apt-get install libosmesa6-dev` |
| `darwin` | byte-identical |

The platform is never consulted. The real macOS cause is that a CGL context needs a window-server session, which a process started by launchd, cron or a bare ssh login does not have; and a context that worked EARLIER in the same process was lost rather than missing, which a fresh process fixes and an install does not.

**2. A refusal was reported as coldness.** The recorder thread warms each camera before the timing loop and had one verdict for every camera that never produced usable output. A render can instead come back as a structured error RESULT rather than a frame - and because that is a *returned dict* and not a raised exception, the warmup loop's `except` branch never sees it, so the reason was dropped at every log level. Measured end to end by making `_get_renderer` return `None` for one real recording:

|---|---|---|
| warmup verdict | `1 cameras still cold after 30 attempts ... First captured frames may show gradient artifact.` | `rendering REFUSED for 1 of 1 cameras ... This is not a cold context that settles - every captured frame will be empty. First reason: Rendering unavailable (no OpenGL context) ...` |
| artifact keys | `camera, errors, frames, path, size_kb` | the same **plus `render_refused`** |
| reason recorded anywhere | no | log and artifact |

No number of attempts fixes a missing GL context, so "cold" is not a slower version of the same story, and its stated consequence - a gradient artifact on the first frames - understates a recording in which *every* frame is missing. The recording reported `status="success"` with 0 frames and no MP4, and the artifact said only `errors: 12`, which does not say why.

## Change

`no_gl_context_message(*, depth=False, platform=None)` becomes the one owner of the sentence and answers for the host reading it: darwin gets CGL, the window-server session and the lost-context case; every other platform keeps the package advice, which on Linux is exactly right. The three inline copies in `render`, `render_depth` and `get_frame` now call it, so `apt-get install libosmesa6-dev` appears once in the module.

The warmup loop captures the refusal per camera in a `_refusals` map, and the verdict splits: refused cameras are reported as refused, genuinely cold ones keep the message they always had, and a batch containing both gets both lines. The reason travels into `state["refusals"]` and out through the artifact as `render_refused` plus a line in the result text, so `frames: 0` arrives with its cause attached.

`platform=` is injected rather than read from the host, so both halves are graded on every runner - nothing in the new tests skips on a Linux CI box.

**What this does not change:** a host that cannot render still cannot record. `status` is still `success` and `frames` is still `0` in both columns. What changes is that the operator learns why, instead of being told to wait for a context that will never settle.

## Deviations from the source of the two commits

Three, each deliberate:

- The helper's docstring stated the mechanism in terms of a specific deployment; rewritten to state the mechanism itself, with `Args:`/`Returns:`.
- The refusal-text read was wrapped in `except Exception`. AGENTS.md forbids that for a non-recovery path and asks for the smallest superset, so I enumerated what the read can actually raise - a missing key, an empty content list, and a non-subscriptable content or block - and narrowed it to `(IndexError, KeyError, TypeError)`.
- One test named an internal tracking id; renamed for the behaviour it pins.

`import sys as _sys` stays at function scope because that is the one pre-existing spelling in this same module (`rendering.py:1273` on `main`); hoisting it would leave the file with two conventions.

## Scope

`backend.py` names `libosmesa6-dev` twice more and is deliberately untouched: both sites sit behind `_is_headless()`, which returns `False` on macOS by design - its own docstring says the EGL/OSMesa fallback is Linux-specific - so neither is reachable on the host this fixes.

`render_refused` is added to the artifact without a docstring entry because its two sibling optional keys, `frames_skipped_size_mismatch` and `flush_error`, are equally undocumented there; documenting one of three would be the odd one out.

## Tests

`tests/simulation/mujoco/test_recorder_warmup_refusal_is_not_coldness.py` is new - there was no test for the refusal path anywhere. It drives the real daemon recorder: `start_cameras_recording` waits on `state["ready"]`, which `_loop` sets only after warmup finishes, so the call returning IS the synchronisation point. No sleeps and no thread patching.

`TestTheFixIsPlatformCorrect` joins the existing `test_render_no_gl_context_message.py`, whose module docstring already pins the four renderer consumers and their channels.

Pre-fix, both files copied onto `upstream/main`: **12 failed / 10 passed**. Six failures are behavioural in the refusal file, each quoting the real message:

```
AssertionError: a refused render was reported as a camera that is merely cold:
["recorder thread warmup: 1 cameras still cold after 30 attempts: ['cam_a'].
  First captured frames may show gradient artifact."]
```

A seventh is behavioural in the platform file (`the Linux advice lives in no_gl_context_message and nowhere else` - it counts two copies on `main`). The remaining five can only report that the helper does not exist yet, so they pin the contract rather than the regression.

The 10 that pass on both trees are the controls. Three matter most, and each fails for a specific wrong fix:

- a genuinely cold camera is still reported as cold, with `gradient artifact` and no `REFUSED` - fails if the two paths are ever collapsed into one
- a camera that warms reports neither
- the synchronous recorder, which has no warmup loop and so never carries the key, still flushes - pins that the shared flush reads `refusals` as absent rather than assuming it

## Verification

Measured at `18bc0cf6`, against `upstream/main` at `71735507`, on one host with `MUJOCO_GL=egl`.

**Control - a recording on a host that CAN render**, driven with real GL through the same code: 19 frames captured, 45.7 KB MP4, and **19 frames decoded back out of the file** on both trees. Normal recording is untouched.

**Gate.** 268 test files - the mujoco suite, every test naming the changed symbols, and the repo-wide docstring / string / changelog guards - run on both trees with the identical selection, the new file excluded so any difference is the source change alone:

|---|---|---|---|
| this branch | 0 | 7405 | 52 |
| `upstream/main` | 0 | 7399 | 52 |

Branch-only failures: none. The `+6` is exactly the six tests in `TestTheFixIsPlatformCorrect`; the new refusal file's own 9 tests pass separately.

`ruff check` and `ruff format --check` clean. `mypy strands_robots tests tests_integ`: 24 errors on the branch and 24 on `upstream/main`, identical sets, none in the changed files. The two `py/empty-except` sites an AST pre-flight reports in `rendering.py` are byte-identical on `main` and none of the eight hunks here is near them.

`rendering.py::no_gl_context_message` is registered in the mujoco public-member inventory, which is what a new public module-level function is required to do.

no-GL render refusal vs coldness

The figure is three measurements from one script run against each tree: the control recording with its decoded frames, the same recording on a GL-free host, and the advice each tree gives a Mac. Both columns run identical package code except `rendering.py`.

### DIFF
--- a/changelog.d/2605-no-gl-refusal-is-not-coldness.md
+++ b/changelog.d/2605-no-gl-refusal-is-not-coldness.md
@@ -1,0 +1,37 @@
+### Fixed
+
+The no-OpenGL-context error now answers for the host reading it, and the camera
+recorder no longer reports a render REFUSAL as a camera that is merely "cold".
+
+`render`, `render_depth` and `get_frame` each carried their own copy of one
+sentence for every host: "Install EGL or OSMesa for offscreen rendering:
+apt-get install libosmesa6-dev". macOS has neither EGL nor OSMesa -- MuJoCo
+renders through CGL there, and there is no apt -- so a Mac operator was sent to
+install a package that does not exist for their machine while the real cause
+kept its cover. Measured on one tree with `sys.platform` set each way, the text
+`render` returns is byte-identical: the platform is never consulted.
+`no_gl_context_message` now owns the sentence and names the two macOS causes,
+which need opposite actions -- a CGL context needs a window-server session,
+which a process started by launchd, cron or a bare ssh login does not have; and
+a context that worked earlier in the same process was lost rather than missing,
+which a fresh process fixes and no install does. Linux keeps the packages, where
+they are exactly right, and the advice now lives in one place instead of three.
+
+The recorder thread's warmup had one verdict for every camera that never
+produced usable output: "still cold after 30 attempts ... first captured frames
+may show gradient artifact". A render can instead come back as a structured
+error RESULT rather than a frame, and because that is a returned dict and not a
+raised exception the warmup loop's `except` branch never saw it -- so the reason
+was dropped at every log level and the recording finished as `status="success"`
+with zero frames, no MP4, and an artifact whose only hint was an error count.
+No number of attempts fixes a missing GL context, so "cold" is not a slower
+version of the same story and its stated consequence understates a recording in
+which every frame is missing. Warmup now records the refusal per camera, reports
+refused cameras apart from genuinely cold ones, and carries the reason out
+through the artifact as `render_refused` and a line in the result text, so
+`frames: 0` arrives with its cause attached.
+
+A host that cannot render still cannot record: `status` and `frames` are
+unchanged. What changes is that the operator learns why. A genuinely cold camera
+keeps the message it always had, and a batch containing both kinds gets both
+lines.
--- a/strands_robots/simulation/mujoco/rendering.py
+++ b/strands_robots/simulation/mujoco/rendering.py
@@ -45,6 +45,45 @@ logger = logging.getLogger(__name__)
 _DEFAULT_MAX_RENDER_BYTES = 50 * 1024 * 1024  # 50 MB
 
 
+def no_gl_context_message(*, depth: bool = False, platform: str | None = None) -> str:
+    """The one sentence every renderer consumer says when there is no GL context.
+
+    The advice was a Linux package install on every host. macOS has neither EGL
+    nor OSMesa - MuJoCo renders through CGL there - so on a Mac the reader was
+    sent to install a package that does not exist for their machine while the
+    real cause kept its cover: a CGL context needs a window-server session,
+    which a process started by launchd, cron or a bare ssh login does not have.
+    The Linux advice is kept, because on Linux it is exactly right.
+
+    Args:
+        depth: ``True`` when the caller is the depth renderer, so the sentence
+            names depth rendering rather than rendering.
+        platform: Platform to answer for, spelled as ``sys.platform`` spells it.
+            Defaults to the running platform; a caller passes it explicitly to
+            answer for a host it is not running on.
+
+    Returns:
+        One stripped sentence naming the failure and a fix that exists on
+        ``platform``.
+    """
+    import sys as _sys
+
+    system = platform or _sys.platform
+    head = (
+        "Depth rendering unavailable (no OpenGL context). " if depth else "Rendering unavailable (no OpenGL context). "
+    )
+    if system == "darwin":
+        return head + (
+            "macOS has no EGL or OSMesa - MuJoCo renders through CGL - so installing "
+            "Linux GL packages cannot help here. A CGL context needs a window-server "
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Two reporting fixes on the no-GL-context path, exactly as described. `no_gl_context_message(*, depth=False, platform=None)` becomes the single owner of the advice sentence and answers per-platform (darwin gets the CGL / window-server / lost-context causes; everything else keeps the Linux package advice), replacing the three inline copies in `render`, `render_depth` and `get_frame`. The recorder thread's warmup loop now captures a structured error *result* (which the `except` branch could never see, since it is a returned dict) into a per-camera `_refusals` map, splits the verdict between refused and genuinely cold cameras, and carries the reason out through `state["refusals"]` into the artifact as `render_refused` plus a result-text line - so `frames: 0` beside `status: success` arrives with its cause attached. Behaviour on a host that can render is untouched.

## What's good

- The refusal-text read uses a narrow `except (IndexError, KeyError, TypeError)` with a placeholder fallback, per AGENTS.md's exception-clause discipline - the fact of the refusal survives even a malformed envelope, with no silent downgrade back to "cold".
- The flush reads `state.get("refusals") or {}` defensively, and a control test pins that the synchronous recorder (which never carries the key) still flushes - the shared path assumes absence, not presence.
- `platform=` is injected rather than read from the host, so both message halves are graded on a Linux CI runner; `test_every_consumer_goes_through_the_helper` pins that the apt-get advice cannot drift back into a second copy.
- Sequencing is sound: `_refusals` is written before `state["ready"].set()`, and `stop_cameras_recording` joins the thread before the flush reads it.
- Changelog fragment, public-member docstring inventory entry, ASCII-only strings, and behaviour-named tests all in place. Verified locally at the head SHA: all 25 tests in the three touched/added test files pass with real MuJoCo, and the repo-wide ASCII string guard passes.
- The `backend.py` scope carve-out is correctly reasoned - both remaining `libosmesa6-dev` sites sit behind `_is_headless()`, which is `False` on macOS by design.

PR #2606, opened 2026-08-22
fix(mesh/iot): the fan-out's routing attribute is not a caller label

cagataycali:
## Problem

`provision_robot` / `provision_operator` inject a `strands-mesh-role` Thing
attribute for their own fleet routing, and merged a caller entry of the same
name:

```python
attributes = dict(attributes) if attributes else {}
attributes.setdefault("strands-mesh-role", "robot")   # the caller's value wins
```

That attribute is not a label. The deployed e-stop fan-out
(`bootstrap._ESTOP_LAMBDA_SOURCE`) enumerates every Thing and publishes
`{"action": "stop"}` only to those whose value is exactly `"robot"`:

```python
attrs = thing.get("attributes") or {}
if attrs.get("strands-mesh-role") != "robot":
    continue
```

So a robot provisioned with any other value is **silently excluded from every
fleet-wide emergency stop**. Nothing reports it: `provision_robot` returns a
`ProvisionedThing`, the Thing exists, its certificate works, it connects to the
mesh and obeys `cmd` - it just never receives the fleet stop. The fan-out
reports a successful invocation too, because a skipped Thing is not an error
there. The docstring never said the key was reserved, so a caller re-provisioning
a Thing from `describe_thing` output has no way to learn it.

The same merge on `provision_operator` is the mirror: an operator console
relabelled `robot` is enumerated by the fan-out and sent a stop on its own
`cmd` inbox.

## Measured

Driven through the shipped code - `provision_robot` against a boto3 double,
then `bootstrap._ESTOP_LAMBDA_SOURCE` executed verbatim over the Things that
resulted. Every row reported `status`-equivalent success on `main`.

| caller passes | AWS receives | role | fleet e-stop |
| --- | --- | --- | --- |
| nothing | 1 key | `robot` | delivered |
| `{"hw": ..., "site": ...}` | 3 keys | `robot` | delivered |
| `{"strands-mesh-role": "so101-arm"}` | 1 key | `so101-arm` | **NOT delivered** |
| three keys | **4 keys** | `robot` | delivered |

`published: 3` of 4 provisioned Things, and the one skipped is the one whose
routing label the caller set.

strands-mesh-role reservation

## Change

`strands-mesh-role` is refused rather than merged, on the same rung as the two
refusals already there - before `_require_boto3()`, so the verdict is the same
with and without boto3 installed and the docstring's "checked before any AWS
call, so a refused call provisions nothing" covers all three inputs. The
injection is now an assignment through a named constant rather than
`setdefault`, so the module's own value cannot lose.

Refusing rather than overwriting follows the module's established posture: it
already refuses a `thing_name` outside its charset and a non-boolean
`allow_estop_publish` before touching AWS, and silently discarding a value the
caller deliberately passed would leave them no signal at all.

Two docstring claims are corrected in the same contract, because both are
about the budget this key spends:

- `attributes` documented `<=3 keys` as the *caller's* budget. AWS allows three
  attributes per Thing and this function spends one of them, so the caller's
  budget is two. The third caller key silently made a fourth attribute.
- It read `<=800 chars total`. Per botocore's own service model for
  `CreateThing`, `AttributeValue` has `max: 800` and `AttributeName` has
  `max: 128` - 800 bounds one value, not the dict.

Out of scope, deliberately: enforcing AWS's per-Thing attribute quota or its
`[a-zA-Z0-9_.,@/:#-]` charset locally. Those are breadth decisions about how
much of the service contract this module should restate, and neither is what
made a robot unreachable.

## Tests

`tests/mesh/test_mesh_role_attribute_is_reserved.py` is new. It measures the
consumer premise rather than asserting it: `_fanout_targets` executes
`bootstrap._ESTOP_LAMBDA_SOURCE` verbatim over a Things page, so the routing
rule under test is the one that is deployed.

That driver installs its boto3 double in `sys.modules`, because that is where
the source resolves the module. `_ESTOP_LAMBDA_SOURCE` carries its own
`import boto3`, so a patch on a module-level binding this test file holds is
only reached while nothing has replaced that entry - and a sibling that removes
`boto3` from `sys.modules` leaves the patch on an orphaned object the source
never consults. The fan-out then runs against the real SDK: a live signed AWS
request from a unit test. CI measured it as `1 failed, 2766 passed` with
`botocore.exceptions.NoRegionError`; forcing the order locally
(`pytest tests/mesh/test_iot_camera_offload.py tests/mesh/test_mesh_role_attribute_is_reserved.py`)
gives 3 failed / 68 passed with `NoCredentialsError` and the Lambda's own
fail-open log line, against 74 passed after. `test_iot_provisioning_hook.py`
already drives the sibling hook source in this same module through
`sys.modules`, with a comment naming the reason ("which would hit AWS"); this
is the same seam for the other source.

`TestTheDoubleIsInstalledWhereTheSourceLooks` pins it, because the binding
shape passes all 22 behavioural tests in this file when the file runs alone -
so it is dormant on whichever collation order happens to put a remover after
it, and only a structural check notices a revert. Each pin has a break that
fires it and nothing else:

| break | forced order | pins firing, file alone |
| --- | --- | --- |
| revert to the binding patch | 6 failed | all 3 |
| keep the seam, re-add the module-level import | 1 failed | `test_this_module_holds_no_boto3_binding_to_patch` |
| behaviour intact, seam spelled so the check cannot see it | 1 failed | `test_the_double_goes_through_sys_modules` |
| half-migration: seam added, old patch left behind | - | `test_no_module_level_boto3_binding_is_patched` |
| revert the seam **and** delete the pin class | 3 failed | none - 22 passed |

The last row is why the pins are there. The binding is graded as an import
statement rather than a substring, so naming it in prose binds nothing.

Out of scope: `test_estop_fanout_turn_id_unique.py` carries the same binding
shape. It sorts before the only remover in the tree, so it is dormant rather
than broken, and it is not what this branch is about.

`tests/mesh/test_iot_provision.py::test_user_can_override_role_attribute`
asserted the caller's value was kept, with the docstring "If user explicitly
sets strands-mesh-role, their value is kept." That is this defect written down
as intent, by a test that predates the fan-out consumer. It is now
`test_user_cannot_override_role_attribute` and asserts the refusal, with a
docstring saying which consumer made the old claim unsafe.

Pre-fix, against `upstream/main` with both test files copied in:
**16 failed / 45 passed**, of which 8 are behavioural and 8 report the absence
of the new private helper. The three seam pins pass there too - they grade this
test file's own shape rather than `provision.py`. The behavioural eight:

| test | pre-fix |
| --- | --- |
| `provision_robot` refuses the reserved key | DID NOT RAISE `ValueError` |
| `provision_operator` refuses the reserved key | DID NOT RAISE `ValueError` |
| the refusal lands before boto3 is resolved (x2) | boto3 was resolved first |
| the offered remedy is accepted | DID NOT RAISE `ValueError` |
| the documented budget is the caller budget | quotes `'<=3 keys, <=800 chars total)'` |
| the reserved key is named in both docstrings | never names it |
| `test_user_cannot_override_role_attribute` | DID NOT RAISE `ValueError` |

The 42 that pass on both trees are the controls, and the load-bearing ones are:

- `TestTheFanoutRoutesOnThisAttribute` - the premise. The consumer is untouched
  by this branch, so these pass either way; they are what makes the exclusion a
  measurement rather than a claim.
- `TestTheModuleStillOwnsTheLabel` - a robot is still labelled `robot`, an
  operator `operator`, and caller attributes still travel alongside. Nothing on
  the paths that already worked moves.
- `TestTheHelperIsTotal` - a non-mapping `attributes` is left alone, so
  `dict(attributes)` keeps owning that refusal.

## Verification

Measured at `cb9b66a5`. Merge base `a5007b66`; `upstream/main` has since moved
to `604285d6` and the overlap gate is green, so no base merge is carried.

- **Gate**: 214 files (all of `tests/mesh`, everything naming boto3 /
  `sys.modules` / provisioning, plus the repo-wide docstring, string,
  dependency and changelog guards), the identical selection on both trees:
  branch **33 failed / 8420 passed / 38 skipped / 24 errors**, the PR's previous
  head `f75a9775` **36 / 8414 / 38 / 24**. Branch-only failing ids **empty**;
  base-only is exactly the three the fix repairs. `+6 passed` reconciles as the
  3 repaired plus the 3 new pins, and `+3` collected as the pins.
- The 57 shared ids are identical on both trees and all environmental here: 45
  `tests/mesh/test_iot_*` (`awsiot`), 2 `test_device_connect_hardening`
  (`device_connect_agent_tools`), 2 `tests/training/test_lerobot.py`
  (`draccus` floor), 2 `test_doctor` (worktree not pip-installed), and 6
  import-order-dependent (`test_recording_frame_loss_is_not_tolerated`,
  `test_rendering_pacers_survive_a_clock_step[forward_30s]`, `test_vla_jepa`).
  The first four are declared extras CI installs.
- `mypy` **24 == 24** against a pristine `upstream/main` worktree, branch-only
  set empty, none in the changed files.
- The `except`-block census in `provision.py` is unchanged at 6 sites, all
  pre-existing.

No new visual artifact for this round: it changes where a test installs its own
boto3 double, so nothing renders. The render above still stands for the
provisioning change, which is untouched.


yinsong1986 (dismissed):
## Summary

Reserves the `strands-mesh-role` Thing attribute in `provision_robot` / `provision_operator`: a caller-supplied entry is now refused with a `ValueError` (before `_require_boto3()`, on the same rung as the `thing_name` charset and `allow_estop_publish` checks, so a refused call provisions nothing), and the module's injection is an unconditional assignment through a named constant instead of `setdefault`. This closes a real safety hole - the deployed e-stop fan-out (`bootstrap._ESTOP_LAMBDA_SOURCE`) publishes `{"action": "stop"}` only to Things whose attribute is exactly `"robot"`, so a caller-overridden value silently excluded that robot from every fleet-wide stop while provisioning reported success. I verified the premise against the shipped fan-out source, confirmed the refusal ordering in both entry points, and found no internal caller passing the reserved key. The behavior change is loud (explicit `ValueError` naming the key, the consequence, and the remedy) with a changelog fragment, which is the migration pattern AGENTS.md requires. The new test module measures the consumer premise by executing the deployed Lambda source verbatim rather than restating its rule, and the flipped `test_user_cannot_override_role_attribute` documents why the old assertion was the defect written down as intent. All 58 tests in the two touched test files pass at this head; ruff check/format clean on the changed files.

## What's good

- The premise tests (`TestTheFanoutRoutesOnThisAttribute`) drive `bootstrap._ESTOP_LAMBDA_SOURCE` verbatim, so the routing rule under test is the one that is deployed - if the fan-out's key or match value ever changes, these tests fail rather than the reservation silently guarding a key nobody reads.
- Refusing rather than silently overwriting follows the module's established posture (no-silent-defaults), and the assignment-through-constant means the module's value wins even for inputs the guard deliberately leaves to `dict()` downstream.
- The docstring budget correction (two caller keys, 800 chars per value) is pinned by tests, and the `Raises:` blocks were updated in the same change per AGENTS.md's completeness rule.

One non-blocking observation for a possible follow-up: `_ensure_thing` returns an existing Thing's ARN without updating attributes, so a Thing already mis-provisioned on current `main` with an overridden role is not corrected by re-running `provision_robot` - the refusal prevents new exclusions but does not remediate existing ones. That path is pre-existing and untouched by this diff.

cagataycali:
The red `Test and Lint` here is not this diff. `Run lint` passed and `Run tests` failed on one
test in this PR's own file:

```
FAILED tests/mesh/test_mesh_role_attribute_is_reserved.py::TestTheFanoutRoutesOnThisAttribute::test_only_a_robot_labelled_thing_is_stopped
  - botocore.exceptions.NoRegionError: You must specify a region.
```

The cause is a test-isolation defect on `main`, not the fan-out test. `tests/mesh/test_iot_camera_offload.py::test_boto3_missing_returns_none`
ends with a bare `sys.modules.pop("boto3", None)`, and it collects ahead of this file. A removal
does not undo an import - it orphans every reference already bound to that module, so this file's
module-level `import boto3` keeps an object no later `import boto3` returns. Its
`monkeypatch.setattr(boto3, "client", ...)` double is installed there while the `exec`d Lambda
source re-imports the real SDK, which is what reaches for a region.

Measured, region unset, on this branch:

| ordering | result |
| --- | --- |
| this file alone | 22 passed |
| `test_iot_camera_offload.py` first (CI's collection order) | 3 failed, all `NoRegionError` |
| the same order with the pop removed | **71 passed** |

The third row is only the one-line removal in `test_iot_camera_offload.py`; nothing in this PR
changes. `tests/mesh/test_estop_fanout_turn_id_unique.py` has the same exposure and escapes today
only because it sorts ahead of the pop.

Fixed at the root in #2608, which drops the removal (it is redundant - the `builtins.__import__`
block above it already makes the import raise `ImportError`, which is what that test asserts) and
adds a guard deriving the rule from the tree. This PR needs no change and goes green once that
lands.

yinsong1986 (approved):
## Summary

This PR reserves the `strands-mesh-role` Thing attribute in `provision_robot` / `provision_operator`: a caller-supplied entry is refused with a `ValueError` before `_require_boto3()` is resolved (on the same rung as the `thing_name` charset and `allow_estop_publish` refusals, so a refused call provisions nothing), and the module's own injection becomes an unconditional assignment through the new `_MESH_ROLE_ATTRIBUTE` constant instead of `setdefault`. The defect it closes is real and silent: the deployed e-stop fan-out (`bootstrap._ESTOP_LAMBDA_SOURCE`) publishes `{"action": "stop"}` only to Things whose attribute is exactly `"robot"`, so a caller-won value excluded that robot from every fleet-wide stop while provisioning, connectivity and `cmd` handling all reported success. I verified the premise against the shipped fan-out source at head, confirmed the refusal ordering in both entry points, confirmed no internal caller or agent tool passes the reserved key, and found no blockers: the behaviour change is loud (an explicit `ValueError` naming the key, the consequence and the remedy) with a changelog fragment, which is the migration pattern AGENTS.md requires for a previously-accepted input. The head commit's `sys.modules` seam for the boto3 double is the correct fix for a Lambda source that does its own `import boto3`, and it is pinned structurally so a binding patch cannot quietly regress into reaching the real SDK.

## What's good

- The premise is measured, not asserted: `_fanout_targets` executes `bootstrap._ESTOP_LAMBDA_SOURCE` verbatim, so if the fan-out's key or match value ever changes, the reservation's justification fails a test rather than silently guarding a key nobody reads.
- The flipped `test_user_cannot_override_role_attribute` documents why the old assertion was the defect written down as intent, with a pointer to the consumer test that proves it.
- Refusing rather than silently overwriting follows the module's established posture (no-silent-defaults), and the assignment-through-constant means the module's value wins even for input shapes the guard deliberately leaves to `dict()` downstream.
- Docstring corrections (caller budget is two of AWS's three attributes; 800 chars bounds one value, not the dict) are pinned by tests rather than left as prose.

## Verification suggestions

- `python -m pytest tests/mesh/test_mesh_role_attribute_is_reserved.py tests/mesh/test_iot_provision.py --no-cov -q` - 61 passed at this head.
- The repo-wide guards that grade the new surface (`tests/test_raises_docstring_completeness.py`, `tests/test_docstring_xref_roles_resolve.py`, `tests/test_log_strings_are_ascii.py`, `tests/test_test_case_names_describe_behaviour.py`) all pass at this head, and `ruff check` is clean on the three changed Python files.
- One residual worth a follow-up issue (not this PR): `_ensure_thing` returns unchanged when the Thing already exists, so a Thing provisioned pre-fix with a wrong role keeps it across a re-provision that reports success - remediation for existing fleets is out of this PR's scope but worth tracking on the board.

PR #442, opened 2026-06-14
refactor(curobo): migrate to cuRobo main API (MotionPlanner / DeviceCfg / GoalToolPose) (closes #421)

yinsong1986:
# Migrate CuroboPolicy to cuRobo main API (closes #421)

## What & Why

PR #306 introduced `CuroboPolicy` against cuRobo's `0.7.x` API surface.
@cagataycali's [Thor validation report](https://github.com/strands-labs/robots/pull/306#issuecomment-4700020819)
(sm_110 / CUDA 13) confirmed the stub-based 42 unit tests pass cleanly
and the chunked-trajectory architecture is solid, but the live code
paths in `_build_motion_gen` / `_build_start_state` / `_build_goal_pose` /
`_build_goal_joint_state` / `_apply_world_update` reference import
paths that no longer exist on a fresh `NVlabs/curobo` checkout.

This PR ports those internal helpers to the restructured `main` API
while keeping the public surface (`get_actions`, `reset`,
`set_robot_state_keys`, the `target_pose` / `target_joints` /
`world_update` kwargs) unchanged.

## API mapping (issue #421 verbatim)

| 0.7.x (PR #306) | `main` (this PR) |
|---|---|
| `curobo.types.base.TensorDeviceType` | `curobo.types.DeviceCfg` |
| `curobo.types.state.JointState` | `curobo.types.JointState` |
| `curobo.types.math.Pose` | `curobo.types.Pose` (planning needs `GoalToolPose`) |
| `curobo.wrap.reacher.motion_gen.MotionGen` | `curobo.motion_planner.MotionPlanner` |
| `curobo.wrap.reacher.motion_gen.MotionGenConfig` | `curobo.motion_planner.MotionPlannerCfg` |
| `curobo.geom.types.WorldConfig` | Removed - scene flows via `MotionPlannerCfg.create(scene_model=...)` |

Shape changes:

- `MotionGen.plan_single(start, goal_pose)` -> `MotionPlanner.plan_pose(GoalToolPose, JointState)`
- `GoalToolPose` requires 5D tensors `[B,H,L,G,3]` + explicit `tool_frames`
- `MotionGenConfig.load_from_robot_config(...)` -> `MotionPlannerCfg.create(robot=...)`
- `TensorDeviceType()` -> `DeviceCfg(device=torch.device('cuda:0'))`

## Per-helper changes

`strands_robots/policies/curobo/policy.py`:

- **`_build_motion_gen`** - now uses `MotionPlannerCfg.create(robot=,
[...]

PR #2607, opened 2026-08-22
docs(policies/curobo): number every rung the joint-space dispatch probes

cagataycali:
## Problem

`CuroboPolicy._plan_and_cache` takes the first planner entry point a planner exposes, because cuRobo renamed those entry points between its legacy `MotionGen` surface and the `main` `MotionPlanner` surface. For a joint-space goal it probes three of them:

```python
plan_js_fn = getattr(self._motion_planner, "plan_js", None) or getattr(
    self._motion_planner, "plan_single_js", None
)
if plan_js_fn is not None:
    result = plan_js_fn(start_state, goal_js)
else:
    result = self._motion_planner.plan_single(start_state, goal_js)
```

Its docstring published that ladder as **two** numbered rungs, and folded the other two probes into the first rung's parenthetical:

```
1. New API: ``MotionPlanner.plan_js(JointState, JointState)``
   (or any other goal-aware joint-space entry point cuRobo
   lands; we probe ``plan_js`` first, then ``plan_single_js``,
   then ``plan_single`` as a last resort).
2. Legacy / stub API: ``plan_single_js(JointState, JointState)``.
```

So the list disagreed with itself and with the code in two ways. `plan_single_js` was named both as rung one's second probe and as rung two's subject, which collapses the ladder into itself - the count of rungs stopped matching the count of things tried. And `plan_single`, the rung the code actually ends on, was numbered nowhere: a reader sizing a stub planner from the numbered list is told a planner exposing only `plan_single` has no joint-space path at all, when it is in fact the rung that plans the goal.

Three pieces of in-file evidence say the code is the correct half:

- The **Cartesian** ladder eight lines above numbers exactly the two entry points it probes (`plan_pose`, then `plan_single`) - the same rule, already met by the sibling ladder in the same docstring.
- The code comment inside the joint-space branch already names all three and states the fallback condition correctly ("Fall back to `plan_single` only if neither joint-space entry point is exposed").
- The dispatch is pinned by tests at rungs one and two.

Measured through the real dispatch, one planner shape per rung:

| goal | planner exposes | lands on |
| --- | --- | --- |
| joint-space | `plan_js` + `plan_single_js` + `plan_single` | `plan_js` |
| joint-space | `plan_single_js` + `plan_single` | `plan_single_js` |
| joint-space | `plan_single` only | `plan_single` |
| Cartesian | `plan_pose` + `plan_single` | `plan_pose` |
| Cartesian | `plan_single` only | `plan_single` |

That third row had no test. `_NewApiStubMotionPlanner` removes the legacy pair to pin rung one, and `_StubMotionGen` exposes both legacy entry points to pin rung two, so no stub reached a planner exposing `plan_single` without `plan_single_js` - which is how the last resort came to be missing from the numbered list.

Both the numbered list and the parenthetical were written in the same commit (#442, 2026-06-15), so the list has never matched the ladder.

## Change

The joint-space ladder now numbers a rung per entry point, in probe order, with the forward-looking note kept as prose rather than as a competing order inside rung one:

```
1. New API: ``MotionPlanner.plan_js(JointState, JointState)``
2. Legacy / stub API: ``plan_single_js(JointState, JointState)``
3. Last resort: ``plan_single(JointState, JointState)`` - a planner
   exposing neither joint-space entry point still plans a joint-space
   goal through its Cartesian-shaped entry point.
```

No behaviour changes: the probe order and every dispatch outcome are identical before and after.

`tests/policies/curobo/test_dispatch_ladder_is_documented_in_probe_order.py` grades both ladders against the code rather than against other prose, so neither can drift again:

- **Same rungs, same order.** The numbered rungs must equal the entry points the branch probes, derived from the AST (both the `getattr(self._motion_planner, "name", None)` probe and the `self._motion_planner.name(...)` call). A fourth entry point is graded on arrival, with no list to update.
- **One rung, one entry point.** A rung may not name the probes belonging to the rungs below it.
- Each rung of the joint-space ladder is driven through the real dispatch, which pins the last resort for the first time.
- A rung floor and a probe floor keep a docstring reflow that hides the numbered lists from satisfying every rule by grading nothing.

Both rules are parametrized over the two ladders, so the Cartesian half is the control: it passes on both trees and shows the rule is the file's own convention rather than a new one.

## Verification

Measured at `ff8673c`, on the branch and on a pristine `upstream/main`.

Pre-fix split, with the new file copied onto `upstream/main`: **2 failed / 8 passed** (10 passed after). Both failures are semantic and quote the sequences:

```
'Dispatch order for joint-space goals:' numbers 2 rung(s) ['plan_js', 'plan_single_js']
for a ladder of 3: the code probes ['plan_js', 'plan_single_js', 'plan_single'].

'Dispatch order for joint-space goals:' has a rung naming several entry points, so the
same name reads as both this rung's probe and another rung's subject:
{'1. New API: ``MotionPlanner.plan_js(...)`` (or any other ... we probe ``plan_js``
first, then ``plan_single_js``, then ``plan_single`` as a last resort).':
 ['plan_js', 'plan_single_js', 'plan_single']}
```

The eight that pass on both trees are the controls: both Cartesian cases, the three behavioural rungs, and the three non-vacuity checks.

Blast radius - 342 test files (everything naming the curobo policy or a planner entry point, plus every guard that walks docstrings or sources), the new file excluded so the two runs are comparable:

| --- | --- | --- | --- |
| branch | 17 | 15897 | 85 |
| pristine `upstream/main` | 17 | 15897 | 85 |

Failing-ID sets identical in both directions (`comm` empty each way). All 17 are `tests/mesh/test_iot_*` needing `awsiotsdk`, a declared extra CI installs.

`tests/policies/curobo/` 172 passed. `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on the branch and 24 on a pristine worktree, none in the changed files.

## Artifact

The ladder each tree documents beside the ladder the code probes, both columns produced by one script that reads its own tree's docstring and then drives the real dispatch. The dispatch table is the control: identical in both trees, because no behaviour changed.

curobo joint-space dispatch ladder, documented vs probed

### DIFF
--- a/changelog.d/2607-curobo-joint-space-ladder.md
+++ b/changelog.d/2607-curobo-joint-space-ladder.md
@@ -1,0 +1,15 @@
+### Fixed
+
+- **policies/curobo**: the documented joint-space dispatch ladder now numbers a rung
+  per planner entry point, in the order `_plan_and_cache` probes them. It numbered two
+  rungs for a ladder of three and folded the other probes into the first rung's
+  parenthetical, so `plan_single_js` read as both rung one's second probe and rung
+  two's subject, and `plan_single` - the entry point the code actually ends on - was
+  numbered nowhere. A reader sizing a stub planner from the list was told a planner
+  exposing only `plan_single` has no joint-space path, when it is the rung that plans
+  the goal. The Cartesian ladder in the same docstring already numbered exactly the two
+  entry points it probes, and the joint-space branch's own comment already stated the
+  fallback condition correctly, so no behaviour changes: the probe order and every
+  dispatch outcome are identical. Both ladders are now graded against the AST rather
+  than against other prose - a rung per entry point in probe order, one entry point per
+  rung - and the last resort is driven through the real dispatch for the first time.
--- a/strands_robots/policies/curobo/policy.py
+++ b/strands_robots/policies/curobo/policy.py
@@ -685,10 +685,14 @@ class CuroboPolicy(Policy):
         Dispatch order for joint-space goals:
 
         1. New API: ``MotionPlanner.plan_js(JointState, JointState)``
-           (or any other goal-aware joint-space entry point cuRobo
-           lands; we probe ``plan_js`` first, then ``plan_single_js``,
-           then ``plan_single`` as a last resort).
-        2. Legacy / stub API: ``plan_single_js(JointState, JointState)``.
+        2. Legacy / stub API: ``plan_single_js(JointState, JointState)``
+        3. Last resort: ``plan_single(JointState, JointState)`` - a planner
+           exposing neither joint-space entry point still plans a joint-space
+           goal through its Cartesian-shaped entry point.
+
+        Both joint-space names exist because cuRobo renamed that entry point
+        between its legacy and ``main`` surfaces, so a planner is free to
+        expose either one and the first it has is the one used.
         """
         # Refresh the collision scene if requested.
         if world_update is not None:
--- a/tests/policies/curobo/test_dispatch_ladder_is_documented_in_probe_order.py
+++ b/tests/policies/curobo/test_dispatch_ladder_is_documented_in_probe_order.py
@@ -1,0 +1,241 @@
+"""The documented dispatch ladders name the entry points, in the order probed.
+
+``CuroboPolicy._plan_and_cache`` probes several planner entry points per goal
+type and takes the first one the planner exposes, because cuRobo renamed them
+between the legacy ``MotionGen`` surface and the ``main`` ``MotionPlanner``
+surface. Its docstring publishes that ladder as a numbered list, and a reader
+sizing a stub planner - or a maintainer adding a fourth entry point - works
+from the list rather than from the ``getattr`` chain.
+
+Two properties keep the list usable, and both are graded against the code
+rather than against other prose:
+
+* **Same rungs, same order.** The numbered list is the probe order, so a rung
+  the code tries is a rung the list numbers.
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

Docs-only correction plus a guard test: the joint-space dispatch ladder in `CuroboPolicy._plan_and_cache`'s docstring now numbers one rung per planner entry point in the order the code probes them (`plan_js` -> `plan_single_js` -> `plan_single`), replacing the two-rung list that folded all three probes into rung one's parenthetical and named `plan_single_js` twice. Verified against `main`'s dispatch code that the new list matches the `getattr` chain exactly and that no behaviour changes. The new test file reproduces the PR's claims: 2 failed / 8 passed with the test on the pre-fix docstring (both failures the documented semantic ones), 10 passed post-fix, and `tests/policies/curobo/` at 172 passed. Lint and format are clean on both changed Python files, the diff is ASCII throughout, the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention, and the test class/function names describe behaviour rather than provenance per AGENTS.md.

## What's good

- Both ladders are graded against the AST rather than against other prose, so a fourth entry point is caught on arrival with no list to update; the Cartesian ladder serves as a passing control.
- The previously untested last-resort rung (a planner exposing only `plan_single`) is now driven through the real dispatch, and the rung floor / probe floor keep the docstring scan from going vacuous under a reflow.
- Tight scope: one docstring, one fragment, one test file; probe order and every dispatch outcome are byte-identical before and after.

PR #2608, opened 2026-08-22
test(mesh): a module removal must not orphan a reference a sibling patches

cagataycali:
`tests/mesh/test_iot_camera_offload.py::test_boto3_missing_returns_none` ended with a bare
`sys.modules.pop("boto3", None)`. A removal does not undo an import - it **orphans** every
reference already bound to that module. `tests/mesh/test_estop_fanout_turn_id_unique.py` does
`import boto3` at collection time and then `monkeypatch.setattr(boto3, "client", fakes[name])`,
so once the entry is gone its double is installed on an object no `import boto3` returns, and
the deployed Lambda source it `exec`s re-imports the real SDK.

It is dormant on `main` today only by sort order: `test_estop_...` collects ahead of
`test_iot_camera_offload.py`, so it runs before the pop. Any later test that patches `boto3`
through a module-level binding is reached.

## Measured

Same two files on `upstream/main`, only the order and the ambient region varying.

| order | `AWS_DEFAULT_REGION` | result |
| --- | --- | --- |
| natural (`estop` then `camera_offload`) | unset | 51 passed - the pop happens last, nothing observes it |
| `camera_offload` then `estop` | unset | **2 failed**, `botocore.exceptions.NoRegionError: You must specify a region` |
| `camera_offload` then `estop` | `us-west-2` | **2 failed**, `botocore.exceptions.NoCredentialsError: Unable to locate credentials` |

The third row is the reason this matters more than a fixture nit: with a region configured the
unit test builds a real client and attempts a **signed AWS request**. The exec'd fan-out source
logs its own fail-open path against it - `estop dedup check failed (failing open): Unable to
locate credentials` - so on a host with credentials that is a live `dynamodb:PutItem` /
`iot:ListThings` from `pytest`. The failure also depends on ambient AWS configuration, so it
reproduces differently on a developer machine than in CI.

## Change

**The removal.** Dropped. It was redundant: `monkeypatch.setattr(builtins, "__import__",
blocked)` on the line above already makes the function-scope `import boto3` in
`CameraOffloader._client()` raise `ImportError`, which is the whole of what the test asserts,
and monkeypatch undoes it. The test's assertions are untouched. The now-unused local
`import sys` goes with it, which is the evidence the pop was its only purpose.

`monkeypatch.setitem(sys.modules, name, None)` is the idiom for "make `import name` raise
`ImportError`" - same effect, and it restores. This file already uses it for `cv2` 338 lines
later, so the correct spelling was in the same file.

**The guard.** `tests/test_sys_modules_removal_leaves_no_orphan.py` derives the rule from the
tree rather than listing it:

- **Protected** is every module a test file binds with a module-level `import` *and* then
  patches an attribute on. Those are exactly the references a removal can orphan; a file that
  only *reads* `mod.attr` is unaffected by getting a fresh copy. Today that resolves to 31
  entries, including `boto3`, `serial`, `imageio` and a set of `strands_robots.*` submodules.
- **Reported** is a removal of a protected module with no restoration in the same function.

Of the 3 unrestored literal removals in the test tree, it reports exactly **1** - this one.

The rule is one-directional and under-reports rather than over-reports, and the module
docstring says where: only a literal key is graded (a loop purging `mujoco*` by variable is out
of reach of a static read); any `finally`, `patch.dict`, `monkeypatch.setitem` or re-assignment
counts as restoring without checking that it restores the same key; and every name bound to
`sys` is followed, so an aliased `import sys as _sys` is graded on both sides.

**AGENTS.md** gains the rule beside the `os.environ` one it specialises, in `### Testing
Patterns` - the same contract for the other process-global dict.

## Scope

Purging a module **nothing patches** stays legal, because it cannot orphan a patch. That is a
deliberate cache-invalidation idiom here, and the two other unrestored removals are both of
that kind: `tests/policies/lerobot_local/test_resolution.py` drops `lerobot.*` to force
re-registration, and `tests/simulation/test_policy_runner.py` drops the runner to measure its
import graph. Neither is reported, and
`test_purging_a_module_no_test_patches_stays_legal` pins that.

A blanket "no raw `sys.modules` mutation" rule was not shipped: the tree has 99 of them and
nearly all are legitimate (module-scope injection of a script under test, a `finally`-restored
purge, the torch mock installer). Widening the rule to those is a breadth decision, not this
change.

No file under `strands_robots/` is touched, so there is no shipped behaviour to film and no
visual artifact - the change is a test-isolation fix, a guard, and one documentation bullet.

## Break table

Every guard is load-bearing; each break fires exactly its own.

| break | reports the offender | purge stays legal | dynamic key unclaimed | restore credited | scan non-vacuous | aliased `sys` |
| restore the removal (`upstream/main`'s file) | FAIL | | | | | |
| protect every module-level import | FAIL | FAIL | | | | |
| grade a non-literal key | FAIL | | FAIL | | | |
| stop crediting a restore | | | | FAIL | | |
| point the scan at no tree | | | | | FAIL | |
| drop the `sys`-alias resolution | | | | | | FAIL |
| control | 12 passed | | | | | |

## Verification

Measured at `566e34bf`.

- **Pre-fix, the guard against the unfixed tree:** 1 failed / 10 passed, the failure naming
  `tests/mesh/test_iot_camera_offload.py:321 in test_boto3_missing_returns_none() removes
  'boto3', which is patched by tests/mesh/test_estop_fanout_turn_id_unique.py`. 12 passed after.
- **The ordering:** the forced order above goes from 2 failed to **51 passed**.
- **The repaired test:** `tests/mesh/test_iot_camera_offload.py` 49 passed, with and without an
  ambient region.
- **Collateral:** the identical 514-file selection (every test that touches `sys.modules`, walks a source tree or inspects docstrings, plus all of `tests/mesh`) run on this branch and on a pristine `upstream/main` worktree, with the new guard excluded from both, gives **33 failed / 20527 passed / 98 skipped / 24 errors on each** - byte-identical summary lines, 57 failing ids on each, `comm` empty in both directions. All 57 are local-environment artifacts of declared extras: 45 `tests/mesh/test_iot_*` (`awsiot`), 2 `tests/test_device_connect_hardening.py` (`device_connect_agent_tools`), 2 `tests/training/test_lerobot.py` (a `draccus` floor), 2 `tests/test_doctor.py` (installed-distribution metadata), and 5 order-dependent cases that pass in isolation on a pristine tree.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy`
  24 == 24 with the branch-only set empty.


yinsong1986 (approved):
## Summary

Test-isolation fix plus a tree-derived guard. `test_boto3_missing_returns_none` ended with a bare `sys.modules.pop("boto3", None)`, which orphans the module-level `import boto3` binding that `tests/mesh/test_estop_fanout_turn_id_unique.py` patches `client` on - so in the reversed collection order the e-stop fan-out tests install their double on an object no later `import boto3` returns and the exec'd fan-out source builds a real client, attempting a signed AWS request from a unit test. The fix drops the pop (redundant beside the `builtins.__import__` block, which is what the test actually asserts and which monkeypatch restores), and `tests/test_sys_modules_removal_leaves_no_orphan.py` derives the rule from the tree: a module a test binds at module level and attribute-patches may not be removed without restoration, while purging an unpatched module to force a re-import stays legal. I ran the guard against the current (unfixed) tree and it reports exactly the one offender, naming the correct patcher file; the repaired `test_iot_camera_offload.py` passes 49/49 with boto3 importable, confirming the `__import__` block alone carries the assertion. No `strands_robots/` code is touched; changelog fragment and AGENTS.md Testing Patterns bullet are present.

## What's good

- The guard's scope discipline matches the PR's claims: literal keys only, restore credited permissively, purge-without-patch left legal - all one-directional under-reporting, documented in the module docstring and pinned by planted-source tests.
- The break table's controls (aliased `sys`, `del` statement, dynamic key, read-only module) are each independently pinned, so a clean result means the scan looked.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; new strings are ASCII-clean.

## Verification suggestions

- One docstring nit worth a glance before merge (not blocking): the guard's module docstring attributes the measured failures to `tests/mesh/test_mesh_role_attribute_is_reserved.py`, a file that does not exist on `main` - the PR description and the guard's own runtime output both name `tests/mesh/test_estop_fanout_turn_id_unique.py`, which is the only module-level boto3 patcher in the tree.
- To reproduce the ordering evidence: `pytest tests/mesh/test_iot_camera_offload.py tests/mesh/test_estop_fanout_turn_id_unique.py -p no:randomly` on the pre-fix tree fails 2 with `NoRegionError`/`NoCredentialsError` depending on ambient AWS config; on this branch it passes.

PR #2610, opened 2026-08-22
fix(transforms): resolve a runtime-registered transform through the public resolver

cagataycali:
## Problem

`strands_robots.transforms.factory` has two public entry points onto one question - which
`DatasetTransform` subclass a provider name means - and they disagreed.
`create_transform` consulted the runtime registry first; `import_transform_class` never
consulted it at all, resolving only a `policies.json` `"transform"` block and then
auto-discovery of `strands_robots.transforms.<provider>`. Each function kept its own copy
of the lookup, so one had the rung and the other did not.

`register_transform`'s own docstring example, and the "Custom backends" section of
`docs/data/transforms.md`, both prescribe the route that breaks:

```python
register_transform("my_v2v", lambda: MyTransform, aliases=["mv"])

create_transform("my_v2v")          # -> MyTransform
import_transform_class("my_v2v")    # -> ValueError: No transform registered for provider
                                    #    'my_v2v'. Available transforms:
                                    #    ['cosmos_transfer', 'mock', 'mv', 'my_v2v']
```

The refusal builds its available list from `list_transforms()`, so it advertised the name
it had just rejected - the provider name and its alias alike.

Measured on `upstream/main`, with a provider registered exactly as documented:

| name | registered in | `import_transform_class` | `create_transform` |
|---|---|---|---|
| `cosmos_transfer` | runtime registry | resolves | resolves |
| `mock` | runtime registry | resolves | resolves |
| `mv` (alias) | runtime registry | **ValueError** | resolves `MyTransform` |
| `my_v2v` | runtime registry | **ValueError** | resolves `MyTransform` |

No provider in `policies.json` declares a `"transform"` block, so **every** name this
factory can serve today arrives through `register_transform` - and the two shipped
transforms resolved only because their modules happen to sit under
`strands_robots.transforms`. `cosmos_transfer` does not even match the
`<Provider>Transform` naming rung - that rung looks for
`"cosmos_transfer".capitalize() + "Transform"`, i.e. `Cosmos_transferTransform`, and the
class is `CosmosTransferTransform` - so it falls through to the "first `DatasetTransform`
subclass in `dir(mod)`" scan. A provider registered the documented way ships no such module and
therefore had no rung at all.

## Why this shape is settled

The module docstring states that this factory "Mirrors
`strands_robots.training.factory`". That sibling's resolver states the contract this PR
restores:

> The one resolver behind `create_trainer`, which adds nothing but the constructor call -
> so a name either function can resolve is a name both can.

and its first rung is the runtime registry, "so re-registering a name overrides a shipped
`trainer` block rather than being ignored by it". The fix is that rung, in that position.

## Change

`strands_robots/transforms/factory.py`:

- `import_transform_class` gains the runtime rung as rung 1, alias-resolved first, ahead
  of the `policies.json` block and auto-discovery. Its docstring now numbers all three
  rungs and states why rung 1 is not optional for the shipped transforms either.
- `create_transform` delegates wholly - `import_transform_class(provider)(**kwargs)` -
  instead of keeping a second copy of the runtime lookup. One resolution order, not two.
- The module docstring states that both registration routes resolve through
  `import_transform_class`, so `list_transforms`, `import_transform_class` and
  `create_transform` answer for the same set of names.

`create_transform`'s behaviour is unchanged: it consulted the runtime registry first
before, and the shared resolver keeps that precedence, so a re-registered name still
overrides a shipped `"transform"` block. The advertised set is identical in both trees and
every listed name resolves to the same class through `create_transform` - nothing about
what a transform *does* changes.

## Tests

`tests/transforms/test_factory.py`, mirroring the sibling module's own resolver class:

- `TestBothResolversServeEveryListedTransform` - every name `list_transforms()` advertises
  resolves; the refusal advertises only names it can serve; the two entry points agree per
  name; the documented `register_transform` route (and its alias) reaches the public
  resolver; the built-ins are reachable through it.
- `TestTheRuntimeRungKeepsItsPrecedence` - a runtime registration shadows a `"transform"`
  block (injected, since none is shipped) through both entry points and stops shadowing
  once removed; an unknown provider is still refused by both; `create_transform`'s kwargs
  still reach the constructor; and `create_transform` keeps no second copy of the rung
  ladder.

Properties are derived from `list_transforms()` **at run time**, so a provider added later
is graded without editing the test. A class fixture registers a runtime-**only** provider
and removes it again, rather than depending on a sibling test having leaked one: without
it the three sweeps grade only names whose modules sit under `strands_robots.transforms`
and pass on a resolver that never consults the runtime registry.

`test_auto_discovery_import` is repaired rather than left in place. It asserted the
built-ins resolve, and once the resolver consults the runtime registry they are answered
by rung 1 - so the test would have stopped covering the rung it is named for. It now
exercises a provider that exists *only* as a module.

**Pre-fix split: 6 failed / 13 passed** against `upstream/main` (`bd6e6333`), all six
behavioural, e.g.

```
list_transforms() advertises providers import_transform_class refuses:
  {'sp': "ValueError: No transform registered for provider 'sp'. Available transforms:
   ['cosmos_transfer', 'mock', 'sp', 'sweep_probe']", 'sweep_probe': ...}
the refusal offers providers it cannot resolve: ['sp', 'sweep_probe']
(name, import_transform_class, create_transform) disagree: [('sp', False, True), ...]
```

The same six fail with the two new classes run **alone**, so the evidence does not depend
on collection order.

Each guard is load-bearing against a distinct regression:

| break | failing guards |
|---|---|
| revert `factory.py` to `upstream/main` | all 6 |
| keep the rung but demote it below the `policies.json` rung | precedence only |
| keep the rung, restore `create_transform`'s own copy | the single-owner pin only |
| the rung ignores aliases | 4, incl. the file's pre-existing `test_runtime_register_and_alias` |
| drop the fixture's runtime-only registration (on base) | the 3 sweeps stop firing - it is what makes them non-vacuous |

## Verification

Measured at `3519b577`; the head is now `03376d03`, which differs only by the
changelog fragment's filename (`2609-` -> `2610-` once the PR number was known),
so every number above still describes this tree.

- `tests/transforms/test_factory.py` **19 passed**; whole `tests/transforms` **140 passed**.
- Order independence: `tests/transforms` + `tests/training/test_factory.py` under
  `pytest-randomly` **178 passed**.
- 211-file selection (everything naming the transform / trainer factory surface, the
  `tests/transforms` suite, and the repo-wide docs / docstring / changelog / source-string
  guards), run on this branch and on a pristine `upstream/main` worktree with the changed
  test file excluded from both so any difference is the source change alone:
  **11 failed, 9132 passed, 57 skipped on each**, 359.23s vs 358.66s, failing ID sets
  identical (`comm` empty both ways). The 11 are absent optional dependencies:
  8 `tests/mesh/test_iot_connect_timeout_domain.py` (`awsiot`) and 3 `tests/training`
  (`draccus` below `lerobot`'s `>=0.11.6` floor).
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`;
  `mypy` **24 == 24** against pristine with the branch-only set empty and none in the
  changed files.

## Artifact

No policy, simulation behaviour, rendering, recording or robot asset changes here - this
is name resolution in a factory, and the transform behaviour is byte-identical either way.
The figure is the resolver matrix, captured by running one script against each tree and
driving the documented `register_transform` route: same advertised set in both panels,
`create_transform` returning the same class for every listed name, and the real refusal a
caller of that route reads on `upstream/main`.

transform factory resolver matrix

### DIFF
--- a/changelog.d/2610-transform-resolver-runtime-registry.md
+++ b/changelog.d/2610-transform-resolver-runtime-registry.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- **transforms**: `import_transform_class` now resolves a provider registered
+  with `register_transform`, so it serves every name `list_transforms()`
+  advertises. It consulted only the `policies.json` `"transform"` block and
+  auto-discovery on `strands_robots.transforms.<provider>`, which refused every
+  runtime-registered provider while `create_transform` resolved it - and the
+  refusal it raised built its available list from `list_transforms()`, so it
+  advertised the name it had just rejected, provider and alias alike. No
+  provider in `policies.json` declares a `"transform"` block, so today every
+  name this factory can serve arrives through `register_transform`; the two
+  shipped transforms resolved only because their modules happen to sit under
+  `strands_robots.transforms`, and a provider registered the documented way
+  ships no such module. The runtime lookup now lives once in
+  `import_transform_class`, which `create_transform` delegates to, and keeps
+  its precedence over a shipped `"transform"` block.
--- a/strands_robots/transforms/factory.py
+++ b/strands_robots/transforms/factory.py
@@ -9,6 +9,10 @@ exactly as trainers use a ``"trainer"`` block; the built-in transforms
 (``mock``, ``cosmos_transfer``) are registered at package import through the
 runtime registry instead, because ``cosmos_transfer`` is a generation model
 with no policy identity to hang a JSON block on.
+
+Both routes resolve through :func:`import_transform_class`, so
+:func:`list_transforms`, :func:`import_transform_class` and
+:func:`create_transform` answer for the same set of names.
 """
 
 from __future__ import annotations
@@ -77,16 +81,33 @@ def list_transforms() -> list[str]:
 def import_transform_class(provider: str) -> type[DatasetTransform]:
     """Import and return the :class:`DatasetTransform` subclass for a provider.
 
-    Resolution order:
-      1. The provider's ``"transform"`` block in policies.json.
-      2. Auto-discovery fallback: ``strands_robots.transforms.<provider>`` with
+    The one resolver behind :func:`create_transform`, which adds nothing but
+    the constructor call - so a name either function can resolve is a name both
+    can. Resolution order mirrors
+    :func:`~strands_robots.training.factory.import_trainer_class`:
+
+      1. The runtime registry (:func:`register_transform`), resolving an alias
+         to its provider first. This rung comes first so re-registering a name
+         overrides a shipped ``transform`` block rather than being ignored by
+         it.
+      2. The provider's ``"transform"`` block in policies.json.
+      3. Auto-discovery fallback: ``strands_robots.transforms.<provider>`` with
          a class named ``<Provider>Transform`` or the first
          :class:`DatasetTransform` subclass.
 
+    Rung 1 is not optional for the shipped transforms either: ``mock`` and
+    ``cosmos_transfer`` register at runtime, so without it they resolve only
+    because their modules happen to be auto-discoverable - and a provider
+    registered through the documented :func:`register_transform` route, which
+    ships no module under ``strands_robots.transforms``, has no other rung at
+    all.
+
     Raises:
-        ValueError: If no transform can be resolved for the provider - no
-            ``transform`` block in policies.json and no
-            ``strands_robots.transforms.<provider>`` module exists.
+        ValueError: If no transform can be resolved for the provider - not in
+            the runtime registry, no ``transform`` block in policies.json and
+            no ``strands_robots.transforms.<provider>`` module exists. The
+            message names the providers this resolver can serve, which is the
+            same set :func:`list_transforms` advertises.
         ImportError: If a module that DOES exist can't be imported: the
             declared policies.json module, or the auto-discovered provider
             module whose own dependency is missing. "Your dependency is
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Moves the runtime-registry lookup (alias-resolved first) from `create_transform` into `import_transform_class` as rung 1, and has `create_transform` delegate wholly to the shared resolver - so both public entry points answer the same question with one resolution order, and a provider registered via the documented `register_transform` route (which ships no module under `strands_robots.transforms` and no `policies.json` block) is now servable by `import_transform_class` instead of being refused by an error message that advertises the very name it rejected. The shape is the settled one from `strands_robots.training.factory.import_trainer_class`, which this module's docstring already claims to mirror, and the runtime rung keeps its precedence over a shipped `"transform"` block, so `create_transform`'s observable behaviour is unchanged.

## What's good

- Verified locally at `03376d03`: `tests/transforms/test_factory.py` 19 passed, whole `tests/transforms` 140 passed.
- The new sweeps derive their graded set from `list_transforms()` at run time, and the autouse fixture registers (and removes) a runtime-only provider so the sweeps cannot pass vacuously on a resolver that skips the runtime rung - the non-vacuity reasoning is stated in the fixture docstring.
- `test_auto_discovery_import` is repaired rather than left silently covering the wrong rung: it now exercises a module-only provider, since the built-ins are answered by rung 1 after this change.
- The `Raises:` block is updated to name the runtime rung in the ValueError condition (AGENTS.md > API Consistency > "A `Raises:` block names every refusal the function itself makes").
- Single-owner pin (`test_create_transform_adds_nothing_but_the_constructor_call`) guards against the exact divergence mechanism that produced the bug.
- Changelog fragment present under `changelog.d/2610-...` per step 3 of the PR workflow; no host paths, no non-ASCII in any changed surface.

PR #2611, opened 2026-08-22
fix(training): the unit suite's merge gate no longer fetches ImageNet weights

### LINKED ISSUE #2609: ci: the sole required check downloads ImageNet ResNet-18 from an external CDN, so any pull request can go red for a reason the merge sweep bills to its author
`tests/training/test_lerobot_e2e.py::test_record_train_load_loop` trains a real ACT policy in the **unit** suite, and lerobot's ACT config initialises its `resnet18` vision backbone from pretrained ImageNet weights. That is an unconditional fetch of `https://download.pytorch.org/models/resnet18-f37072fd.pth` on every run of `call-test-lint / Test and Lint` -- which is the one required status check in the `default` ruleset. So the merge gate for every pull request in this repository depends on an external CDN answering.

Measured on #2498 at head `c3a74688`, whose diff is an Isaac-sim mesh change and touches nothing in `training/`:

```
Downloading: "https://download.pytorch.org/models/resnet18-f37072fd.pth" to /home/runner/.cache/torch/hub/checkpoints/resnet18-f37072fd.pth
            'vision_backbone': 'resnet18'},
ERROR lerobot.py:1683 LerobotTrainer in-process train failed: <urlopen error [Errno 104] Connection reset by peer>
FAILED tests/training/test_lerobot_e2e.py::test_record_train_load_loop - AssertionError: lerobot train raised URLError: <urlopen error [Errno 104] Connection reset by peer>
==== 1 failed, 33818 passed, 285 skipped, 2 warnings in 1409.14s (0:23:29) ====
```

One failure in 33819, on a connection reset, 23 minutes in.

### Why this is worth a fix rather than a re-run

The re-run is the remedy today and it worked here, but the cost is not the retry -- it is the **misattribution**, which is silent and points at the wrong person. `scripts/check_merge_blockers.py --all-open` reports:

```
| #2498 | required-check-failing, missing-approval | the author |
```

`required-check-failing` is owed by the author by construction, because a failing required check normally means the branch broke something. Here the branch broke nothing, so the sweep's advice -- correct for every other cause of that outcome -- sends the author to debug a CDN. This is the same shape as the causes catalogued in AGENTS.md step 8: a pull request that reads exactly like it is waiting on its author and is not. It is also indistinguishable, in every field a sweep reads, from a genuine regression in `training/`.

The exposure is the whole open queue rather than one branch, and it does not decay: a pull request idle in review never re-runs, so a branch that happened to be evaluated during a CDN blip stays red until someone pushes to it or re-runs it by hand.

### Why it recurs rather than being cached away

`.github/workflows/test-lint.yml` sets `cache: 'pip'` and nothing else. There is no cache entry for `~/.cache/torch/hub`, `TORCH_HOME` or `HF_HOME` in any workflow in the tree, so the weights are re-downloaded on every job on every pull request, not fetched once and reused.

### The test does not want these weights

The spec is already written for the from-scratch path, and says so:

```python
base_model="",  # ACT from scratch - smallest CPU path
```

and the module docstring states the intent: "Runs CPU-only, 2 steps - just enough to prove the record->train->load loop." Two steps of ACT on a 20-frame dataset proves the record/train/load seam whether or not the backbone carries ImageNet priors, so the download buys the assertion nothing. It is a default arriving from lerobot rather than a choice this test made -- which is why it reads as intentional in the source and is not.

### Remedies

1. **Turn the pretrained backbone off in the spec** (preferred). lerobot's ACT config exposes a pretrained-backbone-weights knob; setting it to `None` makes the test hermetic, faster, and truer to the `from scratch` comment it already carries. The exact field name should be confirmed against the pinned `lerobot>=0.5.0,<0.6.0` rather than taken from this issue, and it has to reach the policy config through `TrainSpec.extra` alongside `policy_type`.
2. **Move the test to `tests_integ/`.** Consistent with convention 8, which is where real-weights work belongs -- but it removes the record/train/load seam from the suite that gates merges, so 1 is better if the knob exists.

Prefer 1 and keep the coverage. Whichever lands, pin it: a test asserting that this path performs no network access would fail on pre-fix code and is what stops the default drifting back in on a lerobot bump.

### Population

This is the only in-suite test that trains a policy with a vision backbone. The other `trainer.train(spec)` call sites in `tests/` are the SAC and PPO reinforcement-learning tests, which build no image encoder, and the `tests/training/test_lerobot.py` cases, which do not reach a real backbone download. So the fix is one test, not a class -- but the *exposure* it creates is shared by every pull request, which is what makes it worth tracking rather than absorbing.

Found during a scheduled health sweep; #2498's failed job has been re-run and the branch is not at fault.

cagataycali:
Closes #2609.

## Problem

`tests/training/test_lerobot_e2e.py::test_record_train_load_loop` trains a real ACT policy inside the **unit** suite, and `call-test-lint / Test and Lint` is the one required status check in the `default` ruleset. lerobot's `ACTConfig` defaults `pretrained_backbone_weights` to `ResNet18_Weights.IMAGENET1K_V1`, so every run of the merge gate for every pull request fetched **46,830,571 bytes** of `resnet18-f37072fd.pth` from `download.pytorch.org`.

The spec already asks for ACT from scratch:

```python
base_model="",  # ACT from scratch - smallest CPU path
```

and two steps on a 20-frame dataset prove the record -> train -> load seam whether or not the backbone carries ImageNet priors. So the download bought the assertions nothing; it only made the gate depend on a third party answering. That failure is also **indistinguishable from a regression in `training/`** in every field a merge sweep reads, so `required-check-failing` bills it to an author whose branch changed nothing here.

Measured on this host, fresh `TORCH_HOME` per run, with a socket spy recording non-loopback peers:

| measured | `upstream/main` | this branch |
|---|---|---|
| external hosts reached | **1** - `('2600:9000:2801:6c00:d:607e:4540:93a1', 443)` | **0** |
| fresh `TORCH_HOME` after the run | `resnet18-f37072fd.pth`, **46,830,571 B** | none, **0 B** |
| `result.status` | `success` | `success` |
| `model.safetensors` written | yes | yes |
| train wall time | 13.93 s | 12.78 s |

There is no cache entry for `~/.cache/torch/hub` or `TORCH_HOME` in any workflow, so this recurs on every job rather than being fetched once.

## Change

**The spec clears the knob, and the loop asserts it reaches no external host.** `_record_outbound_connects` records non-loopback peers rather than blocking them, so a failure names the host that was contacted instead of surfacing as whatever error the caller makes of a refusal.

**Clearing it needs a `None` in `extra`, and `None` was the one value the two lerobot paths disagreed on.** `build_command` rendered every entry with an f-string, and Python's text for `None` is not YAML's: draccus reads `None` as a plain scalar, so a `str | None` field given `--key=None` was assigned the **string** `"None"` while `build_config` assigned `None`. Measured on the pinned `lerobot[feetech,dataset]>=0.6.1,<0.7.0`:

| `extra` value | in-process | CLI decodes to |
|---|---|---|
| `None` | `None` | `'None'` |
| `"null"` | `'null'` | `None` |
| `"None"` | `'None'` | `'None'` |

No spelling meant the same thing on both paths. And the divergence is not cosmetic - `"None"` reaches torchvision as a weights *name*:

```
>>> torchvision.models.resnet18(weights="None")
KeyError: 'None'
```

so the shelled-out path crashed on the spec the in-process path ran. `_render_extra_value` now renders the null literal, and only the null literal - a rule keyed on truthiness would clear a field for `0` or `False`.

Three surfaces already stated the correspondence this restores: `_decode_extra_value`'s docstring ("the same spelling means the same thing here as on the `--flag=value` CLI"), `tests/training/test_extra_value_decoding.py`'s module docstring ("must mean the same thing on both lerobot paths"), and `docs/training/overview.md` ("one spec means one run whether the backend builds the config in-process or shells out").

`pretrained_backbone_weights` is the only `str | None` field in lerobot's policy configs carrying a non-`None` default, which is why passing `None` had no reachable caller until now.

## Scope

`build_command` at `training/lerobot.py` is the only place a lerobot `extra` entry is rendered as `--key=value`. `training/groot.py` renders for GR00T's own CLI, not draccus, so a YAML null literal would not be its spelling; `tools/lerobot_train.py` carries a separate gated `extra_flags` surface. Neither is touched.

The issue that prompted this cited the pin as `lerobot>=0.5.0,<0.6.0`; it is `lerobot[feetech,dataset]>=0.6.1,<0.7.0`, and the knob was confirmed against that rather than taken from the issue.

## Tests

Pre-fix, with both test files copied onto `upstream/main` and the e2e spec restored to its pre-fix form: **3 failed / 34 passed**, all behavioural.

```
FAILED ...::test_a_none_agrees_with_lerobots_own_cli_parser
  AssertionError: the CLI resolves --policy.pretrained_backbone_weights to 'None' but build_config produced None
FAILED ...::test_the_rendered_token_is_the_yaml_null_not_pythons
  AssertionError: ['--policy.pretrained_backbone_weights=None']
FAILED tests/training/test_lerobot_e2e.py::test_record_train_load_loop
  AssertionError: the record->train->load loop reached 1 external host(s)
  [('2600:9000:2801:1000:d:607e:4540:93a1', 443, 0, 0)]; this suite is the required
  check for every pull request, so a fetch here makes the merge gate depend on that
  host answering
```

The parity tests extend `TestATypedValueMeansTheSameOnBothPaths`, the direction that file's title claims and its cases did not cover: it graded **text** values, while a value already of the field's type was graded on the in-process half alone. `_cli_value` - the file's existing draccus oracle - now derives its registry import from the spec's `policy_type`, so it serves ACT unchanged.

Each guard was checked against a break that should fire it, and only it:

| break | fires |
|---|---|
| revert `build_command` to `upstream/main` | the 2 parity tests |
| drop the knob from the e2e spec | the e2e network assertion |
| also translate the **string** `"None"` | `test_a_string_field_still_carries_the_literal_text_none` only |
| key the rule on truthiness rather than `is None` | `..._still_renders_as_its_own_text[num_workers-0]` and `[policy.freeze_vision_encoder-False]` only |

The last row is why those two falsy cases exist: without them a truthiness-keyed rule cleared a field and nothing noticed.

## Verification

Measured at `02c6ca59` (the code commit; `4718cfc2` adds only the changelog fragment), on the
pinned dependency set (`lerobot 0.6.2`, `draccus>=0.11.6`).

- The two changed test files: `38 passed` (parity) and `1 passed` (e2e), the e2e loop reaching **0 external hosts** and leaving a fresh `TORCH_HOME` empty.
- 117-file blast radius - all of `tests/training`, every test naming `build_command` / `build_config` / `TrainSpec`, every test reading a docs page, and the repo-wide docstring / string / changelog guards - run with the identical selection on this branch and on a pristine `upstream/main`, with the two changed test files excluded so it measures collateral damage only: **`4727 passed, 5 skipped` on both**, byte-identical, zero failures on either side.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 with the branch-only set empty.

## Artifact

The loop records a real MuJoCo episode and trains on it, so the strip is the control: **byte-identical on both trees** (mean|delta| = 0.0000 over all 20 frames) while genuinely moving (per-frame mean|delta| up to 3.094). Only what the run fetched differs.

the unit suite's merge gate fetches no ImageNet weights


yinsong1986 (approved):
## Summary

This PR removes a 47 MB ImageNet-weights fetch from the unit suite's required merge gate by clearing ACT's `pretrained_backbone_weights` in the e2e spec, and fixes the rendering divergence that clearing exposed: `build_command` stringified a Python `None` into the token `None`, which draccus decodes to the *string* `"None"` (a nonexistent torchvision weights name, `KeyError: 'None'`), while the in-process path assigns `None`. The new `_render_extra_value` translates only the null literal - keyed on `is None`, not truthiness - so `0`, `False`, and the literal string `"None"` all survive unchanged, and each of those boundaries is pinned by a test the PR description shows firing on exactly one break. The e2e loop now asserts it reaches no non-loopback host, recording rather than blocking so a failure names the dependency contacted.

## What's good

- The mutation-style verification table (each guard checked against the one break that fires it, including the truthiness-keyed and string-`"None"` traps) is exactly the regression-pinning discipline AGENTS.md asks for.
- Scope discipline: `training/groot.py` and `tools/lerobot_train.py` are explicitly considered and correctly left alone since neither renders for draccus.
- Changelog fragment (`changelog.d/2611-e2e-no-imagenet-fetch.md`), docs update, and the recording fixture's own `root=` all follow repo conventions; all strings ASCII, no host paths.
- The pin cited by the prompting issue (`lerobot>=0.5.0,<0.6.0`) was re-verified against the actual manifest (`lerobot[feetech,dataset]>=0.6.1,<0.7.0`) rather than taken on faith.

PR #2612, opened 2026-08-22
fix(transforms): the re-validation probe records every read a verdict can make

cagataycali:
## Problem

`DatasetTransform.transform` may report a run as gated only if the deterministic verdict it ran actually read an `observation.images.*` column - contract item 1 holds every other column byte-identical between a source episode and its variants, so a verdict that read none of them returns the same answer on all of them and cannot flip. It answers that question by wrapping the source episode in `_KeyRecordingEpisode` and asking afterwards which keys the verdict consulted.

The probe intercepted **6 of `dict`'s 12 value-bearing read paths**. A verdict that copied the mapping before reading it - the ordinary defensive copy a caller makes when it will not mutate a mapping it was handed - received every pixel value and was recorded as having read none:

| how the verdict reads the column | pixel values received | recorded (before) |
|---|---|---|
| `episode[key]`, `get`, `items`, `values`, `==`, `!=` | yes | yes |
| `dict(episode)` | yes | **no** |
| `{**episode}` | yes | **no** |
| `episode.copy()` | yes | **no** |
| `episode \| {}`, `{} \| episode` | yes | **no** |
| `pop`, `popitem`, `setdefault` | yes | **no** |

Driving the real transform with one predicate - `mean pixel < 50` - spelled five ways over the same recorded dataset, with a shift that flips it:

| verdict spelling | `revalidated` | `episodes_discarded` |
|---|---|---|
| `episode["observation.images.cam"]` | `True` | 1 |
| `dict(episode)[...]` | **`False`** | 1 |
| `{**episode}[...]` | **`False`** | 1 |
| `episode.copy()[...]` | **`False`** | 1 |
| `(episode \| {})[...]` | **`False`** | 1 |
| `episode["observation.state"]` (genuinely pixel-blind) | `False` | 0 |

The four middle rows contradict themselves in one payload. `episodes_discarded=1` is a measurement that the verdict answered differently on a generated variant than on its source, and it is printed beside:

> `discarded 1 on the re-validation gate (revalidate consulted no observation.images.* column on source episode(s) [0] - every non-image column passes through byte-identical, so this verdict cannot flip and the output was NOT gated)`

So the gate fired, and the run told the caller it could not.

## Change

Recording now covers every `dict` member that can hand a stored value to a caller, on the rule the class already stated for `items()` / `values()` / `==`:

- a **bulk** read (`items`, `values`, `==`, `!=`, `copy`, `|`) records every key - the verdict received every column's value, so accusing it of reading no pixels would be the false refusal;
- a **keyed** read (`[]`, `get`, `pop`, `popitem`, `setdefault`) records that key;
- a read that reaches **no value** (`keys()`, iterating keys, `in`, `len()`) records nothing, so a genuinely pixel-blind verdict is still reported as ungated. A false record would claim a gate the verdict never applied, which is the over-claiming direction.

The four-times-repeated bulk rule becomes one named `_record_every_key`, since it would otherwise appear seven times.

`dict(episode)` and `{**episode}` are reached by overriding `__iter__`. It records nothing itself - iteration yields keys, never values - but CPython performs those two merges by reading the C table directly, invisibly to `__getitem__`, and takes that fast path only while `tp_iter` is `dict.__iter__`. Overriding it routes both spellings through `keys()` and `__getitem__` per key instead. That is an interpreter detail, so it is pinned behaviourally rather than assumed: `copy()` / `|` are also overridden explicitly, which leaves only those two spellings depending on the merge route, and if a future interpreter merges some other way the totality tests fail loudly instead of the probe quietly returning to under-reporting. Verified on CPython 3.12.3 and 3.13.12 (`requires-python = ">=3.12"`).

## Scope

Only the measurement changes. Which episodes are generated, discarded and written is untouched - the discard comparison already ran on the raw generated mapping - and the probe stays a transparent stand-in for the episode dict: still a `dict`, still comparing by episode contents, still unhashable - asserted as parity with `dict` through `collections.abc.Hashable` rather than by calling `hash()`, which states the contract directly and keeps the deliberate refusal out of CodeQL's `py/hash-unhashable-value` - and `copy()` / `|` still return a plain `dict` exactly as they do for any other `dict` subclass.

`__repr__` is deliberately **not** recorded, with the reason stated where the classification lives: it is a formatted rendering rather than a value handed to the caller, and recording on it would let an unrelated debug log of the episode claim the gate.

## Tests

`tests/transforms/test_verdict_probe_records_every_read_path.py` grades `dict` itself, partitioned exhaustively against `vars(dict)` (35 members: 12 value-bearing, 23 with a stated reason they cannot hand a value out), so a member a future interpreter adds is unclassified and fails there rather than becoming a silent hole. `tests/transforms/test_verdict_flip_discard.py` gains the end-to-end half in the class that already owns this contract.

**Pre-fix split** (both test files against `upstream/main`, source unmodified): **29 failed / 48 passed** -> 77 passed. The sharpest failure quotes the contradiction directly:

> `dict(episode): discarded 1 on the gate and still reports it ungated - mock: read 1 episode(s), wrote 0 generated episode(s), discarded 1 on the re-validation gate (revalidate consulted no observation.images.* column ...)`

The 48 that pass on both trees are the controls: every keys-only read records nothing, the probe remains a `dict` that compares and copies like one, and the two existing "not accused" tests for `items()` and `==` are undisturbed.

**Each guard is load-bearing.** Reverting one piece of the fix at a time, the failures are disjoint:

| break | failures |
|---|---|
| exact pre-fix `base.py` | 29 - the whole defect |
| drop `__iter__` | 7 - only the `dict()` / `{**}` spellings, their end-to-end cases, and the mechanism pin |
| drop `copy` / `__or__` / `__ror__` | 3 - the explicit-override pins only; the behavioural tests still pass through the merge route, which is exactly what those three overrides remove the reliance on |
| drop `pop` / `popitem` / `setdefault` | 8 - those three |
| bulk rule records nothing | 22, including the two pre-existing `items()` / `==` tests, so the shared helper preserved them |
| `__iter__` records every key | 2 - the keys-only false-record controls |
| control (restored) | 0 of 207 |

## Verification

Measured at `3df44b8f`. 72-file selection - all of `tests/transforms`, every test naming the transform surface or reading `docs/data/`, and the repo-wide docs / docstring / string / changelog guards - run identically on the branch and on a pristine `upstream/main` worktree, with the new test file excluded from both:

| tree | result |
|---|---|
| branch | 2624 passed, 3 skipped, **0 failed** |
| pristine `upstream/main` | 2616 passed, 3 skipped, **0 failed** |

Zero failures on either side; `+8` passed is exactly the 8 cases added to the existing file. `tests/transforms` alone: 207 passed. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 against the pristine worktree with the branch-only set empty.

## Artifact

One recorded 3-episode dataset, one verdict spelled `dict(episode)["observation.images.cam"]`, run end to end through the real transform on both trees. Both produce the **same** augmented dataset - reads 3 episodes, writes 2, discards 1 on the gate, one MP4, provenance `synthetic=true` with `source_episode_index=[0, 2]` - and the frames read back out compare equal byte for byte. Only what the caller is told about the gate differs.

re-validation gate report before and after

The generated pixels, read back out of the written dataset (source left, generated variant right; mean 26.9 -> 120.3):

round trip of the written dataset


yinsong1986 (dismissed):
## Summary

This PR closes an under-reporting hole in `_KeyRecordingEpisode`, the probe `DatasetTransform.transform` uses to decide whether a re-validation run may honestly be reported as gated. Previously a verdict that read pixels through a defensive copy (`dict(episode)`, `{**episode}`, `episode.copy()`, `episode | {}`) or through `pop`/`popitem`/`setdefault` received every value while being recorded as having read none, producing the self-contradictory payload of `revalidated=False` ("cannot flip") beside `episodes_discarded=1` on that same gate. The fix extends recording to every value-bearing `dict` member on the class's stated rule (bulk read records every key, keyed read records its key, keys-only read records nothing), and the `__iter__` override is the load-bearing piece: I verified empirically on CPython 3.12 that without it, `dict(probe)` takes the dict-merge fast path and records zero keys, and with it, every key is recorded through `keys()` + `__getitem__`. The probe copies the episode's entries at construction, so the newly intercepted mutating reads (`pop`/`popitem`/`setdefault`) touch the probe, not the source episode the variants are generated from. Which episodes are generated, discarded or written is unchanged.

## What's good

- The `vars(dict)`-exhaustive classification test means a member a future interpreter adds fails loudly as unclassified rather than becoming a silent recording hole, and the `CANNOT_HAND_OUT_A_VALUE` table states its reasons so the partition is checkable rather than trusted.
- The bulk-copy tests pin the interpreter mechanism (the fast-path bypass), not just this module, so an interpreter change degrades to a loud failure instead of silent under-reporting.
- Full compliance sweep is clean: 77 transform tests pass, ruff check/format clean, ASCII-only strings, no host paths, changelog fragment present and correctly named, behavioral test names, docs updated in the same PR.

yinsong1986 (approved):
## Summary

This PR closes an under-reporting hole in `_KeyRecordingEpisode`, the probe `DatasetTransform.transform` uses to decide whether a re-validation run may honestly report itself as gated. Before, a verdict that read pixels through a defensive copy (`dict(episode)`, `{**episode}`, `episode.copy()`, `episode | {}`) or through `pop`/`popitem`/`setdefault` received every value while being recorded as having consulted none, producing the self-contradicting payload of `episodes_discarded=1` beside a message claiming the verdict cannot flip. The fix intercepts every value-bearing `dict` member on the rule the class already stated (bulk reads record every key, keyed reads record their key, keys-only reads record nothing), and overrides `__iter__` to disable CPython's dict-merge fast path so `dict(probe)` / `{**probe}` route through `keys()` + `__getitem__`. Scope is measurement-only: which episodes are generated, discarded and written is untouched, and the probe stays a transparent `dict` stand-in (copies return plain `dict`, equality ignores instrumentation, hashing stays refused via the `Hashable` parity assertion the head commit added in response to the CodeQL thread).

## What's good

- The new test file partitions `vars(dict)` exhaustively (12 value-bearing, 23 with a stated reason), so a member a future interpreter adds fails loudly instead of becoming a silent hole - and the `__iter__` mechanism is pinned behaviourally rather than assumed.
- Verified locally at `3df44b8f`: both transform test files pass (77 tests), `ruff check` / `ruff format --check` clean on the changed files, ASCII-clean, no host paths; mutation isolation holds (a verdict popping keys off the probe leaves the wrapped source episode intact), and `dict.update(probe)` - a defensive-copy spelling not listed in the tables - is also recorded through the same slow-path mechanism.
- Changelog fragment per AGENTS.md step 3, docs updated in the same PR, test names describe behaviour, and the keys-only controls pin the over-claiming direction (a pixel-blind verdict is still reported as ungated).

PR #483, opened 2026-06-16
feat(policies): WBCPolicy provider for GR00T Whole-Body-Control (SONIC, ONNX)

rmncardoso:
## Summary

Adds **`WBCPolicy`** — a new non-VLA policy provider wrapping NVIDIA's GR00T
Whole-Body-Control (SONIC / decoupled-WBC) ONNX controllers for deploy-grade
**Unitree G1 locomotion**. Clean-room implementation against the upstream
reference (`NVlabs/GR00T-WholeBodyControl`,
`decoupled_wbc/sim2mujoco/scripts/run_mujoco_gear_wbc.py` +
`resources/robots/g1/g1_gear_wbc.yaml`). Closes #466.

Validated end-to-end against the **real shipped weights**
(`GR00T-WholeBodyControl-{Balance,Walk}.onnx`): through the torque-control deploy
harness the G1 produces a **stable forward walk** (~0.38 m/s for a 0.5 m/s
command, height held, no fall), and holds balance in place on a zero command.

## What's included

- **`strands_robots/policies/wbc/`** — `WBCPolicy` (provider `wbc`, shorthand
  `sonic`), `WBCConfig`, the observation builder, and pure-NumPy control/quat
  helpers.
  - `requires_images = False`; reads the goal from the well-known kwargs
    (`target_velocity = [vx, vy, omega]`, optional `target_orientation` for base
    RPY, optional `height`). Drives the **15 leg+waist DOFs**; the 14 arm joints
    are held at defaults (upper-body composition is left to a future
    `CompositePolicy`, #468).
  - Faithful to the reference contract: 86-dim observation
    (`command[7] + base_ang_vel[3] + projected_gravity[3] + qj[29] + dqj[29] +
    prev_action[15]`) stacked over `obs_history_len=6` → 516-wide network input;
    **whole-body** qj/dqj (all 29 joints, not just the 15 controlled); two ONNX
    sessions (`policy` + `walk_policy`) selected by `norm(raw velocity) <= 0.05`;
    PD→torque law via `compute_torques(...)`; zero-warm-started history deque;
    quaternion math numerically identical to upstream (verified to 1e-10).
  - `WBCConfig` loads the upstream `g1_gear_wbc.yaml` directly (JSON or YAML;
    flat `*_scale` keys normalised). Joints resolved **by name** (handles the
    G1's leading `floating_base_joint` + interleaved arms).
  - No weights bundled — fetched/pointed-at at runtime under the NVIDIA Open
    Model License.
[...]

PR #2613, opened 2026-08-22
fix(mesh): forward the locomotion goal a tell() peer needs

cagataycali:
## Problem

`SimEngine.run_policy` documents four issue #300 well-known goal keys and offers the mesh as the remote analogue of the local call:

> Carries the well-known #300 goal keys (`target_pose` / `target_joints` / `target_velocity` / `world_update`) to non-VLA providers (cuRobo, MoveIt2, WBC) that read their goal from kwargs rather than the instruction. **This is the local-sim analogue of the mesh `tell()` path, which already forwards these keys.**

The mesh carried three of them. `target_velocity` was absent from `validate_command`'s field allowlist and from `Mesh._SIM_WELL_KNOWN_POLICY_KWARGS`, so a `tell()` that named it lost it at two layers, and neither says a word: `validate_command` builds its output key by key, so an unlisted field is dropped rather than refused, and the dispatcher then builds `policy_kwargs` from its own tuple.

| documented #300 goal key | admitted by the wire | in the dispatcher's forward set |
|---|---|---|
| `target_pose` | yes | yes |
| `target_joints` | yes | yes |
| `target_velocity` | **no** | **no** |
| `world_update` | yes | yes |

It is the one goal key that is a whole command rather than a refinement. WBC and `wbc_gait` read `[vx, vy, omega]` and nothing else, and with no goal they hold a standing balance - so asking a humanoid peer to walk produced a rollout that reported `status="success"` and did not go anywhere.

Both providers are reachable over the mesh by construction: the policy-provider allowlist is derived from the registry, and the comment on that set already names WBC as the provider a hand-maintained copy had omitted once. `docs/policies/wbc.md` promises the route in as many words - "sharing the non-VLA goal vocabulary so a command can flow through `run_policy` / mesh `tell()` without coupling to a backend".

Dated: the three-key forward set is #304 (2026-06-06), the first `target_velocity` consumer is #483 twelve days later, and the docstring claiming mesh parity is 2026-06-29. Correct when written, never brought along.

## Change

`validate_command` admits `target_velocity` on `target_pose`'s per-component domain - `_coerce_float` per element, so non-finite is refused, out-of-range is refused, and a `bool` is refused by name rather than read as a 1 m/s command. `_SIM_WELL_KNOWN_POLICY_KWARGS` gains it, so it lands in `policy_kwargs` and reaches `get_actions(obs, instruction, **policy_kwargs)`.

The component count is bounded by a new `MAX_TARGET_VELOCITY_COMPONENTS = 16`, purely as DoS defence - the role `MAX_TARGET_JOINTS` plays. The arity is deliberately **not** fixed, because the receivers disagree: WBC and `wbc_gait` require at least three components and read the first three, MotionBricks reads a planar `[vx, vy]`. A fixed length would refuse a shape a shipped receiver accepts, and each one already names its own requirement, so a two-component velocity crosses the wire and WBC answers `target_velocity must have at least 3 elements [vx, vy, omega], got 2`.

Three stale enumerations in the same code paths are corrected: two comments still said the goal travels via `policy_config` (it has travelled in `policy_kwargs` since #2312), and both named only the planner providers.

One invariant needed widening rather than extending. The dispatcher has two sinks and no way to ask a provider which it reads, and that choice was pinned as "no provider names a goal key on its constructor" - true of cuRobo and MoveIt2, and not of WBC, whose constructor `target_velocity` is a documented static default. What makes `policy_kwargs` the correct sink there is precedence: a per-call value overrides the constructor one. Were that reversed, a peer told a new direction would answer with the direction it was built with, so the precedence is now pinned directly.

## Rollout

`mesh tell(peer, "walk forward", policy_provider="wbc", target_velocity=[0.5, 0.0, 0.0])` against a Unitree G1 sim peer, with real GR00T-WholeBodyControl (SONIC) ONNX weights, MuJoCo headless (EGL), 300 ticks at 50 Hz. Same command, same weights, same scene; only `strands_robots` differs between the panels.

mesh forwards the locomotion goal

[MP4](https://raw.githubusercontent.com/cagataycali/robots/media-mesh-locomotion-goal/mesh-locomotion-goal.mp4) | [final frame](https://raw.githubusercontent.com/cagataycali/robots/media-mesh-locomotion-goal/mesh-locomotion-goal.png)

| measured | before | after |
|---|---|---|
| goal keys that survived the wire | `(nothing)` | `target_velocity` |
| dispatcher forward set | 3 of 4 documented | 4 of 4 |
| reported status | `success` | `success` |
| pelvis x after 6 s | `-0.064 m` | `+2.295 m` |
| travelled | **0.064 m** | **2.298 m** |
| frames with motion | 34 of 59 (balance sway) | 59 of 59 |

The rollout reports `success` in both columns, which is the point: nothing in the result distinguishes a peer that was given its goal from one that was not.

The capture binds the checkpoint by registering a `WBCPolicy` subclass and admitting its name through `STRANDS_MESH_POLICY_TYPE_ALLOW` - the documented operator extension point - because the dispatcher's constructor-extras allowlist (`model_path` / `server_address` / `policy_type` / `pretrained_name_or_path`) does not include `checkpoint`. That is orthogonal to this change and left alone; the wire validator, the dispatcher, the policy and the physics are all shipped code.

## Tests

The set both layers carry is graded against `run_policy`'s own docstring rather than restated, in both directions, so a fifth goal key is held to the same parity on arrival and an undocumented `target_`-shaped key is still dropped.

Pre-fix, with both test files copied onto `upstream/main`: **12 failed / 46 passed**. Eleven are behavioural; one names the new constant. The headline reads

```
AssertionError: _SIM_WELL_KNOWN_POLICY_KWARGS omits documented #300 goal keys
['target_velocity']; it carries ['target_pose', 'target_joints', 'world_update']
```

Every new guard fires for exactly its own regression, and nothing else:

| injected regression | guards that fail |
|---|---|
| revert both source files (exact pre-fix) | 12 |
| revert only the wire validator | 10 (wire admission + the component domain) |
| revert only the dispatcher's forward set | 2 (forward set + the dispatch behaviour) |
| the docstring-derived key set parses nothing | 1 (non-vacuity) |
| the wire admits any `target_`-prefixed key | 1 (undocumented-key control) |
| reverse WBC's per-call precedence | 1 (the constructor-sink invariant) |
| none | 0 of 58 |

The 46 both-tree passes are the existing contracts in those two files, unchanged - including the extended constructor-sink invariant, which passes on `upstream/main` too because it states a property of WBC rather than of this change.

## Verification

Measured at `2cb209cd`, `/tmp/lr6env` (mujoco 3.5.0, lerobot 0.6.2).

- `mypy strands_robots tests tests_integ`: **24 errors on the branch, 24 on a pristine `upstream/main` worktree**, branch-only set empty, none in the changed files.
- Blast radius: 237 files - everything naming `validate_command` / `_SIM_WELL_KNOWN_POLICY_KWARGS` / `_dispatch_sim_policy` / `target_velocity` / `policy_kwargs`, all of `tests/mesh`, plus the repo-wide docs, docstring, string and changelog guards. Run with the identical selection on both trees, **excluding the two test files this PR modifies**, so any difference is caused by the source change alone: branch `23 failed / 5390 passed`, base `24 failed / 5389 passed`.
- 45 of those failures are shared and pre-existing here: `tests/mesh/test_iot_*` needing `awsiotsdk`, a declared extra CI installs. The IDs that differ between the two runs are all in `tests/mesh/test_zenoh_transport_security.py`, which opens real localhost Zenoh sessions - branch failed 2 of them, base failed 3 *different* ones, and the file passes **9 passed / 2 skipped on both trees** when run alone. Contention between the two parallel suites, not a regression.
- CodeQL pre-flight over the changed files: identical hit sets on the branch and on the pristine base (line-shifted only); both test files clean.

## Scope

`target_orientation` and `height` are WBC's own per-call kwargs rather than part of the shared #300 vocabulary that `run_policy` documents, so they are out of scope here and `docs/policies/wbc.md` now says which route each takes. Widening the dispatcher's constructor-extras allowlist so a checkpoint can travel is a separate contract and is untouched.

### DIFF
--- a/changelog.d/2613-mesh-forwards-locomotion-goal.md
+++ b/changelog.d/2613-mesh-forwards-locomotion-goal.md
@@ -1,0 +1,51 @@
+### Fixed: the mesh forwards the locomotion goal it documents
+
+`SimEngine.run_policy` documents four issue #300 well-known goal keys -
+`target_pose` / `target_joints` / `target_velocity` / `world_update` - and offers
+the mesh as the remote analogue: "the local-sim analogue of the mesh `tell()`
+path, which already forwards these keys". The mesh carried three of them.
+`target_velocity` was absent from the wire validator's field allowlist and from
+`Mesh._SIM_WELL_KNOWN_POLICY_KWARGS`, so telling a peer to walk left the goal
+behind at two layers.
+
+It is the one goal key that is a whole command rather than a refinement. WBC and
+`wbc_gait` read `[vx, vy, omega]` and nothing else; with no goal they hold a
+standing balance, so a fleet operator asking a humanoid to walk got a rollout
+that reported `status="success"` and did not go anywhere. Neither layer says a
+word about it: `validate_command` builds its output key by key from an allowlist,
+so an unlisted field is dropped rather than refused, and the dispatcher then
+builds `policy_kwargs` from its own tuple.
+
+Both providers are reachable over the mesh by construction - the policy-provider
+allowlist is derived from the registry, and the comment on that set names WBC as
+the provider a hand-maintained copy had already omitted once. `docs/policies/wbc.md`
+promises the route in as many words: "sharing the non-VLA goal vocabulary so a
+command can flow through `run_policy` / mesh `tell()` without coupling to a
+backend".
+
+Dated: the three-key forward set is from #304 (2026-06-06), the first
+`target_velocity` consumer from #483 twelve days later, and the docstring that
+claims mesh parity from 2026-06-29. Correct when written, never brought along.
+
+The wire now validates `target_velocity` on `target_pose`'s per-component domain -
+finite, in range, a bool refused by name rather than read as a 1 m/s command - and
+bounds the component count purely as DoS defence, the role `MAX_TARGET_JOINTS`
+plays. It does not fix the arity, because the receivers disagree: WBC needs at
+least three components and reads the first three, MotionBricks reads a planar
+`[vx, vy]`. A fixed length here would refuse a shape a shipped receiver accepts,
+and each one already names its own requirement, so a two-component velocity
+crosses the wire and WBC answers "target_velocity must have at least 3 elements
+[vx, vy, omega], got 2".
+
+The set both layers carry is now graded against `run_policy`'s own docstring
+rather than restated, in both directions, so a fifth goal key is held to the same
+parity on arrival and an undocumented `target_`-shaped key is still dropped.
+
+One invariant needed widening rather than extending. The dispatcher has two sinks
+and no way to ask a provider which it reads, and that choice was pinned as "no
+provider names a goal key on its constructor" - true of cuRobo and MoveIt2, and
+not of WBC, whose constructor `target_velocity` is a documented static default.
+What makes `policy_kwargs` the correct sink there is precedence: a per-call value
+overrides the constructor one. Were that reversed, a peer told a new direction
+would answer with the direction it was built with, so the precedence is now pinned
+directly.
--- a/docs/policies/wbc.md
+++ b/docs/policies/wbc.md
@@ -131,6 +131,14 @@ command at all the controller holds a standing balance (zero velocity, default
 height + level orientation). Omitting a key (or passing `None`) selects the next
 source in the precedence chain, so `None` is how a kwarg spells "not supplied".
 
+`target_velocity` is one of the issue #300 well-known goal keys, so the mesh
+path forwards it the same way it forwards a planner's goal:
+`mesh.tell(peer, "walk forward", policy_provider="wbc", target_velocity=[0.5, 0.0, 0.0])`.
+It travels as a per-call kwarg, which is why it overrides the constructor
+default rather than being read as one. `target_orientation` and `height` are
+WBC's own kwargs rather than part of that shared vocabulary - set them on a
+local `run_policy(policy_kwargs=...)` call, or as constructor defaults.
+
 Each key's accepted domain is the one `WBCConfig` enforces for the field it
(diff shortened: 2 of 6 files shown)


yinsong1986 (approved):
## Summary

Closes the `target_velocity` gap between the mesh and the goal set `SimEngine.run_policy` documents: the wire validator (`validate_command`) now admits the key on `target_pose`'s per-component domain (finite, bounded, bool refused by name via the shared `_coerce_float`) with a component-count cap (`MAX_TARGET_VELOCITY_COMPONENTS`) that is DoS defence only, and the dispatcher's `_SIM_WELL_KNOWN_POLICY_KWARGS` forwards it into `policy_kwargs` where every provider reads its goal. The arity verdict is deliberately left to the receiving policy since shipped receivers disagree (WBC needs >= 3, MotionBricks reads 2), and the invariant behind the `policy_kwargs`-not-`policy_config` sink choice is widened from "no provider names a goal key on its constructor" (false for WBC) to the precedence fact that actually matters - a per-call kwarg overrides the constructor default - and pinned as such.

## What's good

- The documented goal set is derived from `run_policy`'s own docstring and graded against both layers in both directions, with a non-vacuity guard so a docstring reflow cannot silently empty the parsed set.
- The negative control (`target_acceleration` still dropped at both layers) pins that the allowlist was extended, not widened.
- Wire change is purely additive on a per-key allowlist - an old receiver drops the new key exactly as it did pre-fix - so no compatibility cliff.
- Changelog fragment per AGENTS.md step 3; validator messages are plain ASCII; ruff check/format clean; all 58 tests in the two touched test files pass against the head commit.

PR #2614, opened 2026-08-22
fix(hardware): a rollout shares the motor bus with the mesh's readers

### LINKED ISSUE #2602: hardware_robot.py drives the motor bus outside bus_access, so bus_lock does not serialize a rollout against the mesh's readers
## What is unserialised

`strands_robots/bus_access.py` (#2599) puts one `RLock` on the device so every reader and writer that goes *through it* shares one conversation on the serial motor bus. The mesh call sites were converted. `strands_robots/hardware_robot.py` was not, and it does not import `bus_access` at all:

| line | call | thread |
| --- | --- | --- |
| 890 | `self.robot.get_observation()` | ROS telemetry publish |
| 1426 | `await asyncio.to_thread(self.robot.get_observation)` | task preflight |
| 1597 | `await asyncio.to_thread(self.robot.get_observation)` | policy rollout loop |
| 1630 | `await asyncio.to_thread(self.robot.send_action, action_dict)` | policy rollout loop |
| 2734 | `self.robot.send_action(action)` | `send_action` facade |

These run concurrently with the mesh probes in the same process - the rollout loop even publishes per-step mesh telemetry - so "a policy commanding the arm while the mesh reads its state" is the normal operating mode, not an edge case. Every such overlap is a `[TxRxResult] Port is in use!` refusal: the SDK declines without touching the wire, so it is safe, but the probe publishes nothing that cycle and the rollout's own read can be the one refused.

## Why it is filed rather than fixed in #2599

#2599 originally also cleared the SDK's stranded `port.is_using` flag and counted the recoveries. That is only sound under the proof its own docstring stated - that *every* read and write happens inside `bus_lock` - and these five call sites are the counter-example. Under them, the flag a probe would clear can be one a live exchange legitimately holds, and clearing it re-transmits onto a half-duplex UART mid-reply, which can garble a goal-position write for an arm under power. The recovery was removed from #2599 for that reason (`b2e5759`), leaving the two halves whose proof holds with no premise about other callers.

So this is the prerequisite, and the ordering is the point: it is a change to the primary hardware path, and the flag recovery can only return on top of it.

## What to do

1. Route the five call sites above through `bus_access.read_observation` / `bus_access.write_action`. The lock is an `RLock` held on the device, so a caller already inside it can read again; `hardware_robot` and the mesh hold different wrappers around one lerobot robot, which is exactly why the lock lives on the device rather than on a caller.
2. Measure it the way #2599 measured the mesh side: real readers against a fake bus implementing the vendored feetech `txPacket` refusal, asserting the rollout's reads and writes and the mesh probes all complete rather than colliding.
3. Then, as its own change, reconsider the stranded-flag recovery with the proof now available - including whether the "report a REAL owner" branch can be made to fire before an interleave rather than after one.

## Note on scope

#2599's description lists `hardware_robot.py`'s camera-degraded connect/rollback contract as deliberately out of scope. That is a different concern in the same file; its bus *access* was not a deliberately deferred gap, which is what this issue records.

cagataycali:
Fixes #2602

## Problem

`strands_robots/bus_access.py` puts one `RLock` on the device so every reader and writer that goes *through it* takes the serial motor bus in turn. The mesh modules were converted to it. `strands_robots/hardware_robot.py` was not, and imported it nowhere -- it drove the same device directly at five call sites:

| line | call | thread |
| --- | --- | --- |
| 890 | `self.robot.get_observation()` | `publish_ros_observation` |
| 1426 | `await asyncio.to_thread(self.robot.get_observation)` | policy preflight (`_initialize_policy`) |
| 1597 | `await asyncio.to_thread(self.robot.get_observation)` | rollout loop read |
| 1630 | `await asyncio.to_thread(self.robot.send_action, ...)` | rollout loop write |
| 2734 | `self.robot.send_action(action)` | `send_action` facade (teleop + inbound ROS 2) |

So the lock serialised the mesh's readers against *each other*, which is what it was built and tested for, and not against a policy rollout -- while one process holds both. `hardware_robot` has its own `threading.Lock` (`_task_admission`) that admits one rollout at a time, but that is admission control over *tasks*: the two locks do not know about each other, and neither one stops a mesh sensor loop and a rollout from reaching for the bus at the same instant.

**The failure is asymmetric, and the caller that loses is the one holding nothing.** Measured against the real `run_policy` loop with a mesh reader polling the same device, and a bus double that refuses overlap exactly as the feetech SDK does:

| | status | steps | commands on the wire | bus refusals | mesh reads |
| before | `error` | 0 of 20 | **0** | 1 | 54 (unharmed) |
| after | `success` | 20 of 20 | **20** | 0 | 85 |
| control: rollout alone, no mesh reader | `success` | 20 of 20 | 20 | 0 | n/a |

One refused read is enough, and either of the rollout's two reads will do it. The preflight read is its first touch of the bus and `_initialize_policy`'s `except Exception` turns a refusal into `return False`, so the caller is told

```
Policy rollout error: 0 steps in 0.0s (P on arm)
Error: Failed to initialize policy
```

-- a **policy** fault, for a run that never commanded the arm once. Lose the race one step later instead and the loop's read propagates, so the rollout ends with the SDK's own `Port is in use!` after a step or two. Which one loses varies with the timing; that the rollout ends does not. The mesh reader, holding the lock, logs nothing at all.

Also affected on the same bus: the `send_action` facade is where teleop (`teleop_mixin`) and inbound ROS 2 `joint_command`s (`ros_telemetry`) both arrive, and with three writes racing three mesh reads **3 of 3 arm commands were refused and dropped**, each reported as `status="error"`.

## Change

Route the five sites through `bus_access.read_observation` / `bus_access.write_action`. The lock is an `RLock` held on the device, so a caller already inside it can read again, and `hardware_robot` and the mesh hold different wrappers around one lerobot robot -- which is exactly why the lock lives on the device rather than on a caller. `bus_access` imports only the standard library, so this adds no cycle.

The root cause is then pinned as a rule over the source rather than one test per call site: **no caller reaches an operation `bus_access` owns through a device it holds.** Both halves of that rule are derived, so neither is a list that can fall out of step:

- the guarded operations come from what `bus_access` itself calls on the device it is handed (`get_observation`, `send_action`, `sync_read`) -- so an operation it learns to serialise later is graded without touching the test;
- the population is every module that assigns `self.robot`, i.e. holds a device. That is `hardware_robot` plus the mesh's `core` and `input`, **and the mesh two already pass** -- so the rule needs no list of exempt files, and a simulation backend's identically named `get_observation` is never in scope because it drives no serial port.

The rule matches an attribute *access*, not only a call, because three of the five sites pass the method to `asyncio.to_thread` as a reference. A call-only rule sees 2 of 5 (measured in the break table below).

`bus_access`'s module docstring now names the rollout as a participant and records the asymmetry, and the `send_action` facade's docstring names what it delegates through.

## Scope

- **`connect`/`disconnect` stay direct.** `bus_access` wraps reads and writes; opening the port is not its call, and the derived operation set does not contain `connect` -- so the boundary falls out of the derivation rather than being an exemption. Pinned by `test_connect_is_not_claimed_by_this_rule`.
- **Teleop and inbound ROS 2 are fixed transitively, not separately.** `teleop_mixin` and `ros_telemetry` call the *wrapper*'s `send_action`, which is site 2734.
- No stranded-flag or in-use-flag recovery is introduced or assumed: none exists in the tree, and this change makes no claim that depends on one.

## Sim rollout

The commanded joint targets that actually reached the wire, from each side's real `run_policy` rollout, replayed into a MuJoCo so101 (headless EGL) and rendered. Same policy, same 24-step sweep, same scene; the only difference is whether the rollout takes the device's bus lock. Both panels start from the same rest pose -- mean absolute pixel delta between them at frame 0 is **0.0001** -- and the left panel is honestly static because nothing was ever commanded: per-frame mean absolute delta **0.0000 across all 23 transitions**, against 23 of 23 moving on the right.

before and after: the commanded sweep reaching the arm

<details>
<summary>Final frame as a still</summary>

final frame

</details>

| | status | steps | commands on the wire | bus refusals | mesh reads | replay: frames moving |
| before | `error` | 0 of 24 | **0** | 1 | 73 | **0 of 23** |
| after | `success` | 24 of 24 | **24** | 0 | 96 | 23 of 23 |

## Verification

Pre-fix split, the new file against `upstream/main` (`0b358aaf`): **5 failed / 6 passed -> 11 passed**. Every failure is behavioural or structural with a real message -- no import or attribute error:

- the rollout: `assert 'error' == 'success'`, with `Failed to initialize policy` in the payload;
- the facade: `assert ['error', 'error', 'error'] == ['success', 'success', 'success']`;
- the telemetry publish: the SDK's `Port is in use!` propagating;
- the preflight: `the preflight read was refused, so the rollout reports a policy fault`;
- the rule: naming all five sites, `strands_robots/hardware_robot.py:890, 1426, 1597, 1630, 2734`.

The 6 that pass on both trees are the controls: a rollout alone is unchanged, the mesh's own readers still take turns (the contract `bus_access` already had), `connect` is not claimed by the rule, a real device failure still reaches the caller, and the two non-vacuity checks on the derivation.

Break table -- each break fires exactly and only its own guards, so no guard is decorative:

| break | fails |
| --- | --- |
| revert the source fix entirely | 5 (the pre-fix split) |
| revert only the `send_action` facade | facade test + the rule |
| revert only the policy preflight read | rollout + preflight + the rule |
| derivation returns no operations | the derivation non-vacuity check only |
| module scan reaches nothing | the scan non-vacuity check only |
| revert the three `to_thread` sites, full matcher | rollout + preflight + **the rule catches them** |
| revert the same three, call-only matcher | rollout + preflight -- **the rule reports clean** |

The last pair is why the rule matches an attribute access rather than only a call: a call-only matcher sees 2 of the 5 sites and reports a clean tree while three are reverted.

Gate: 368 test files (every test naming the changed symbols, all `tests/test_hardware_*`, every guard that walks the package tree or reads `AGENTS.md`, plus the repo-wide docs/docstring/string/changelog guards), the identical selection run on this branch and on a pristine `upstream/main` worktree with the new file excluded from both -- **`17 failed, 16750 passed, 86 skipped` byte-identical on both**, failing-ID sets identical, `comm` empty in both directions. All 17 are `tests/mesh/test_iot_*` on `No module named 'awsiot'`, a declared extra CI installs. `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on both trees with none in the changed files.

One test-scaffolding note: this module builds a `Robot._executor` and a test can abandon a work item (`future.result(timeout=...)`), which is the shape `tests/_daemon_executor.DaemonThreadExecutor` exists for -- a `ThreadPoolExecutor` worker is not a daemon and the interpreter joins it at exit, so a wedged item would deliver a failing test as a hung job. The fixture uses the helper, and the module is added to the non-vacuity set in `test_abandoned_work_item_cannot_hang_exit.py` that lists the modules which abandon work.

No serial port is opened and no arm is commanded anywhere in the tests. The bus double is the one the `bus_access` tests already use, subclassed to add the connect surface a rollout drives, so both files agree on what an overlap costs.


yinsong1986 (approved):
## Summary

Routes `hardware_robot.py`'s five direct motor-bus touches (the ROS 2 telemetry publish, the policy preflight read, the rollout loop's read and write, and the `send_action` facade that teleop and inbound ROS 2 commands share) through `bus_access.read_observation` / `write_action`, so a policy rollout takes the same per-device `RLock` the mesh's readers already hold. This closes the asymmetric race the description measures: the unconverted caller was the one that lost, and one refused preflight read was misreported as `Failed to initialize policy` for a run that never commanded the arm. The fix is the minimal conversion (no signature, wire, or config changes; `bus_access` was already the shipped public seam), exception semantics through the wrappers are unchanged, and only one lock is involved on these paths so no lock-ordering hazard is introduced. Teleop and inbound ROS 2 commands are covered transitively through the facade, matching what `mesh/input.py`'s `_default_apply` already does. I verified the new suite against the head tree: all 38 tests in `tests/test_hardware_bus_is_shared_with_the_mesh.py`, `tests/test_bus_access_serializes_motor_reads.py`, and the updated `tests/test_abandoned_work_item_cannot_hang_exit.py` pass, with no serial port opened.

## What's good

- The regression is pinned twice, per AGENTS.md > Review Learnings (#85) > Testing: once as behaviour against the real `run_policy` loop with a bus double that refuses overlap like the feetech SDK, and once as a source-level rule (`TestNoCallerDrivesAHeldDeviceOutsideBusAccess`) whose guarded-operation set and graded population are both derived rather than listed, so a sixth call site or a new `bus_access` operation is graded without editing the test.
- The rule matches attribute *access*, not only calls, which is what catches the three `asyncio.to_thread(self.robot.get_observation)` method-reference sites a call-only scan would miss.
- Scope discipline: `connect`/`disconnect` stay direct and the boundary falls out of the derivation (`test_connect_is_not_claimed_by_this_rule`), and `test_a_real_bus_failure_still_reaches_the_caller` pins that taking the lock does not swallow what the device reports.
- The concurrency contract is documented at both ends (`bus_access` module docstring and the `send_action` facade docstring), per AGENTS.md > Thread Safety > 'Document the concurrency contract'.

## Verification suggestions

- `python -m pytest tests/test_hardware_bus_is_shared_with_the_mesh.py -q` -- runs in ~90s, no hardware; the before/after numbers in the module docstring are reproducible by reverting the five `hardware_robot.py` hunks and re-running.
- The head moved twice during this review (`18abdd79` -> `5606f3e1` -> `70c782e5`); the deltas are confined to the test file (rollout executor swapped to `tests._daemon_executor.DaemonThreadExecutor` so an abandoned work item cannot hang interpreter exit, plus the matching expected-set entry in `test_abandoned_work_item_cannot_hang_exit.py`) and the addition of the `changelog.d/2614-hardware-shares-the-motor-bus.md` fragment. Production code under `strands_robots/` is byte-identical across all three heads; the test run above was against the `5606f3e1` tree, whose graded sources are unchanged at `70c782e5`.

yinsong1986 (approved):
## Summary

Routes the five sites in `strands_robots/hardware_robot.py` that drove the lerobot device directly (the ROS 2 telemetry publish at 891, the policy preflight read at 1427, the rollout loop's read/write at 1598/1631, and the `send_action` facade at 2737) through `bus_access.read_observation` / `write_action`, so a policy rollout, teleop, and inbound ROS 2 commands all take the same per-device `RLock` the mesh's readers already hold. This closes the asymmetric race where the unconverted caller was the one that lost -- one refused preflight read ended a rollout as a misreported policy fault. The root cause is pinned as a derived source-level rule (`TestNoCallerDrivesAHeldDeviceOutsideBusAccess`) whose guarded operations come from what `bus_access` itself drives and whose population is every module assigning `self.robot`, so neither half is a maintainable list.

## What's good

- Behaviour-preserving delegation: `bus_access` re-raises driver exceptions unchanged, so the facade's error-dict contract and the rollout's existing error paths are intact; the remaining direct `self.robot.connect` / `is_connected` / `config` accesses correctly fall outside the derived operation set.
- The AST rule matches attribute *access* rather than only calls, which is what makes the three `asyncio.to_thread(read_observation, ...)` sites gradable -- and the break table demonstrates a call-only matcher would miss 3 of 5.
- AGENTS.md convention compliance across the board: changelog fragment (not a `CHANGELOG.md` append), behaviour-named test classes, `DaemonThreadExecutor` for the abandoning fixture plus the non-vacuity set update, no host paths, ASCII-only tool/log strings, and the AGENTS.md bullet cites its pinning test.
- Teleop and inbound ROS 2 `joint_command` are fixed transitively through the facade rather than with parallel conversions, keeping one seam.

PR #2615, opened 2026-08-22
fix(policies): the ABC documents the locomotion goal key providers already read

cagataycali:
## Problem

`Policy.get_actions` is where the issue #300 well-known `**kwargs` goal vocabulary is *defined*. It is the docstring a provider author reads to learn which keys a caller may pass without coupling to a backend, and `strands_robots.policies`'s own docstring points at it by name for the list: "the well-known `**kwargs` keys (...) **documented on** `Policy.get_actions`".

It listed three keys. `target_velocity` was the fourth everywhere else:

| site | role | keys |
|---|---|---|
| **`Policy.get_actions`** | **the definition** | **3** |
| **`strands_robots.policies` docstring** | **copy that cites the ABC** | **3** |
| **`Cosmos3Policy.get_actions`** | **restates the list to disclaim all of it** | **3** |
| `SimEngine.run_policy` | consumer | 4 |
| `SimEngine.eval_policy` | consumer | 4 |
| `SimEngine.run_benchmark` | consumer | 4 |
| `PolicyRunner.run` | consumer | 4 |
| mesh wire validator + dispatcher | transport | 4 (since #2613) |
| `docs/policies/wbc.md` | reference | 4 |

`docs/policies/wbc.md` says it in as many words - "`target_velocity` is one of the issue #300 well-known goal keys" - and two independent provider families read it: `wbc` (policy + its `gait` variant) and `motionbricks`. So the contract's own definition was the only place the key was missing, and it is the place a new provider author looks first.

Nothing executes a docstring. That is the mechanism: the one list in the set that is not a consumer is the one a new key can be forgotten on while every consumer keeps working, and no test noticed because the guard that existed compared the vocabulary against *another docstring*. #2613 graded the mesh wire and dispatcher against `run_policy`'s docstring - a consumer - so the tree ended up holding two competing definitions with the downstream guards anchored to the wrong one.

`tests/policies/test_base.py::test_well_known_kwargs_are_accepted_by_contract` was the closest thing to a guard and it hard-coded the same three keys, so it agreed with the omission rather than catching it.

## Change

Source is docstring-only, 15 lines across the three enumerations that each claim to be the *present, whole* vocabulary:

- **`policies/base.py`** - the ABC gains the `target_velocity` bullet, in `run_policy`'s documented order. It also records why this key alone carries no fixed arity: shipped receivers disagree (whole-body controllers require `[vx, vy, omega]`, planar drivers read `[vx, vy]`), so each states its own requirement and the key crosses a transport on the per-component domain alone. That is #2613's finding, stated where the vocabulary is defined rather than only where it is validated.
- **`policies/__init__.py`** - the copy that cites the ABC now matches it.
- **`cosmos3/policy.py`** - the disclaimer restates the ABC's list in order to say none of it is read; the restatement is brought current so the sentence keeps covering every key it names.

### Deliberately not touched

Two other kinds of three-key enumeration are **correct** and are left alone, because holding either to the current vocabulary would demand a false statement:

- **provider-scoped** - `curobo/__init__.py` says cuRobo reads pose / joints / `world_update`, which is exactly what cuRobo reads.
- **historical** - "the `Policy` ABC contract for non-VLA providers landed in #300 (well-known ... kwargs)" is true of #300 as shipped; `target_velocity` arrived later.

The discriminator for grading a restatement is therefore that it claims to be the present whole vocabulary, and that rule is written into the test module.

## Tests

The guards deliberately **do not** compare the ABC against another docstring - that is how the tree came to hold two definitions. They derive what the vocabulary must contain from **shipped provider code**, statically scanning every provider package for the keys actually read out of `**kwargs`, and hold the ABC to it in both directions.

The discriminator between shared vocabulary and a provider-private kwarg is the second-caller rule AGENTS.md convention 11 already codifies for value-domain guards: a second independent caller is the evidence a shared name is right, one caller is evidence of nothing. A "provider family" is the provider package, so `wbc/policy.py` and its `wbc/gait.py` variant count once between them - otherwise any provider shipping two entry points would look like a consensus.

Measured on this tree the rule separates the vocabulary from the private kwargs with **no special cases at all**:

```
read by >= 2 families : target_joints, target_pose, target_velocity, world_update
read by exactly 1     : target_orientation, target_heading, target_heading_angle,
                        height, gait_frequency, style, mode, locomotion_style,
                        speed_scale, replan, scene_model, planning_group, seed,
                        text_prompt, guidance_scale, diffusion_steps, video,
                        raw_actions, enable_safety_checker, execution_horizon,
                        prev_chunk_left_over, motion, dof_pos, dof_vel,
                        root_ang_vel_local, anchor_rot_xyzw          (26 keys)
```

which is why the assertion is a set **equality** rather than a containment: the vocabulary is neither missing a key two families share nor carrying one no two families do.

Every guard fires for exactly its own regression, and nothing else:

| injected regression | guards that fail |
|---|---|
| revert all three source files (exact pre-fix) | 2 |
| revert only the ABC | 4 |
| revert only the package docstring | 1 |
| revert only the cosmos3 restatement | 1 |
| falsely promote a single-family kwarg (`target_orientation`) onto the ABC | 6 |
| a second family starts reading `target_orientation` | 2 |
| reword the ABC block sentinel (parse goes vacuous) | 6 |
| none | 0 of 62 |

The pre-fix row failing **2** rather than 4 is the defect's own signature: with the package docstring and cosmos3 reverted too, all three copies agree with each other and only the two guards anchored outside the prose - the provider-code scan and the ABC/`run_policy` equivalence - can still see the omission. That is precisely why the copies were not evidence.

Both halves carry a non-vacuity guard, since a bounding-sentinel parse that silently returns nothing would turn every comparison green. `target_orientation` / `target_heading` / `height` are pinned as single-family negative controls: they are `target_`-shaped and sit in the same `get_actions` bodies as the real vocabulary, so a guard keyed on name shape rather than on the second-caller rule would promote them - pinning them is what makes the equality evidence the allowlist was extended by one key rather than opened to anything goal-shaped.

The ABC/`run_policy` equivalence guard is what makes #2613's wire and dispatcher pins transitively anchored to the ABC that owns the vocabulary, instead of to a second copy that can drift from it independently.

`test_base.py`'s round trip now reads the key set off the contract and fails on a documented key it has no sample value for, so it can no longer agree with an omission.

## Verification

Measured at `bb045f4`, `/tmp/tenv` (python 3.13.15, pytest 9.1.1).

- `ruff check` over `strands_robots tests`: clean. `ruff format --check`: clean (1423 files).
- Blast radius: `tests/policies`, `tests/mesh`, `tests/simulation`, plus the repo-wide `Args:`/attribute docstring-completeness, static-export, docs-architecture, docs-callable-example and changelog-fragment guards. Run with the identical selection on this branch and on a pristine `fc4dba37` worktree, **excluding the two test files this PR adds or modifies**, so any difference is caused by the source change alone: **branch `329 failed / 13328 passed / 731 errors`, base `329 failed / 13328 passed / 731 errors`** - branch-only failure set **empty**, base-only set **empty**, all 329 shared.
- Those 329 failures and 731 collection errors are pre-existing in this environment and identical on both trees: `mujoco`, `lerobot` and other declared extras CI installs are absent from the minimal venv.

## Scope

`target_orientation`, `height` and `gait_frequency` are WBC's own per-call kwargs rather than part of the shared #300 vocabulary, so they stay off the ABC and are pinned as such. Widening the vocabulary itself, or the mesh dispatcher's constructor-extras allowlist, are separate contracts and are untouched.

### DIFF
--- a/changelog.d/2615-abc-documents-locomotion-goal-key.md
+++ b/changelog.d/2615-abc-documents-locomotion-goal-key.md
@@ -1,0 +1,45 @@
+### Fixed: the policy contract documents the locomotion goal key its providers read
+
+`Policy.get_actions` is where the issue #300 well-known `**kwargs` goal
+vocabulary is defined - the docstring a provider author reads to learn which
+keys a caller may pass without coupling to a backend, and the one
+`strands_robots.policies` points at by name for the list. It documented three
+keys.
+
+`target_velocity` was the fourth everywhere else: read by two independent
+provider families (`wbc`, including its `gait` variant, and `motionbricks`),
+forwarded by `run_policy` / `eval_policy` / `run_benchmark`, admitted by the
+mesh wire validator and forwarded by the mesh dispatcher since #2613, and named
+by `docs/policies/wbc.md` as "one of the issue #300 well-known goal keys". The
+contract's own definition was the only place it was missing, and it is the place
+a new provider author looks first.
+
+Nothing executes a docstring, which is the mechanism: the one list in the set
+that is not a consumer is the one a new key can be forgotten on while every
+consumer keeps working. No guard caught it because the guards compared the
+vocabulary against *another docstring* - #2613 graded the wire and dispatcher
+against `run_policy`, a consumer - so the tree held two competing definitions
+with the downstream pins anchored to the wrong one. The nearest thing to a
+contract test hard-coded the same three keys and therefore agreed with the
+omission.
+
+Three enumerations that each claim to be the present, whole vocabulary are
+corrected: the ABC, the package docstring that cites it, and `Cosmos3Policy`'s
+disclaimer, which restates the list in order to say none of it is read and so
+needs to keep covering every key it names. Provider-scoped lists (cuRobo reads
+pose / joints / `world_update`, which is what cuRobo reads) and historical ones
+("the contract landed in #300 (...)", true of #300 as shipped) are deliberately
+left alone, because holding either to the current vocabulary would demand a
+false statement.
+
+The new guards do not compare the ABC against another docstring. They derive
+what the vocabulary must contain from shipped provider code - the keys providers
+actually read out of `**kwargs` - and hold the ABC to it in both directions. The
+discriminator between shared vocabulary and a provider-private kwarg is the
+second-caller rule AGENTS.md convention 11 already codifies, with a provider
+family being the provider package so a provider's own variants count once
+between them. On this tree that rule separates the four vocabulary keys from 26
+private ones with no special cases, so the assertion is a set equality rather
+than a containment, and `target_orientation` / `target_heading` / `height` are
+pinned as single-family controls: they are `target_`-shaped and sit in the same
+`get_actions` bodies, so a guard keyed on name shape would have promoted them.
--- a/strands_robots/policies/__init__.py
+++ b/strands_robots/policies/__init__.py
@@ -8,7 +8,7 @@ produced, so the same interface fits VLA-style providers (consume images +
 instruction) and non-VLA providers (cuRobo, MoveIt2, MPC, pure-IK / scripted
 trajectories).  Non-VLA providers typically set ``requires_images=False`` and
 read their goal from the well-known ``**kwargs`` keys (``target_pose``,
-``target_joints``, ``world_update``) documented on
+``target_joints``, ``target_velocity``, ``world_update``) documented on
 :meth:`Policy.get_actions`.
 
 Built-in providers (see policies.json for full list):
--- a/strands_robots/policies/base.py
+++ b/strands_robots/policies/base.py
@@ -170,6 +170,15 @@ class Policy(ABC):
                 - ``target_joints: dict[str, float]`` - joint-space goal
                   keyed by joint name; values are in radians (revolute) or
                   metres (prismatic).
+                - ``target_velocity: list[float]`` - base-velocity goal as
+                  ``[vx, vy, omega]`` (linear m/s, angular rad/s, in the robot
+                  base frame), read by locomotion providers.  Unlike the keys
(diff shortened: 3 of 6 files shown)


yinsong1986 (approved):
## Summary

Docstring-only source change (+12/-3 across `policies/base.py`, `policies/__init__.py`, `cosmos3/policy.py`) adding the `target_velocity` bullet to the `Policy.get_actions` well-known goal vocabulary and bringing the two restatements that claim to be the whole list (the package docstring that cites the ABC, and `Cosmos3Policy`'s disclaimer) into agreement with it. The new guard module derives the vocabulary from shipped provider code via an AST scan of `**kwargs` reads, applies the AGENTS.md convention-11 second-caller rule at provider-package granularity, and pins the ABC to that set in both directions with non-vacuity guards on every parse; `test_base.py`'s round trip now reads the key set off the contract instead of hard-coding it. Verified independently: the scan on the head tree yields exactly `{target_joints, target_pose, target_velocity, world_update}` (with `target_velocity` read by `wbc` and `motionbricks`), matching the ABC and `run_policy` docstrings, and the single-family controls (`target_orientation`/`wbc`, `target_heading`/`motionbricks`, `height`/`wbc`) hold; all 62 tests in the two touched test files pass at head. No API, wire, or persisted-schema surface is touched, changed strings are ASCII, no host paths, changelog fragment present.

## What's good

- The guards are anchored to shipped provider code rather than a second docstring, which is exactly the mechanism the PR identifies as having let the omission survive #2613's pins.
- Scope discipline: provider-scoped and historical three-key enumerations are deliberately left alone, with the discriminator written into the test module.
- Every parse (ABC block, `run_policy` block, both restatements, the provider scan) carries its own non-vacuity assertion, so a prose reflow fails loudly instead of turning the parity guards green.

PR #2616, opened 2026-08-22
fix(transforms): a provenance record that cannot answer is refused, not read as recorded

cagataycali:
## Problem

The provenance sidecar exists so generated and recorded episodes stay
distinguishable. Its module docstring, the `docs/data/transforms.md` contract
bullet and the test file's own docstring all name the same failure mode - *silent
mixing of generated and recorded data* - and `synthetic_episode_indices()` is the
one call that separates them. The documented consumer is:

```python
synthetic = synthetic_episode_indices("/data/augmented")
# everything in `synthetic` was generated; everything outside it was recorded
```

It read the verdict with `r.get("synthetic") is True`, so a record whose field
was present but not a boolean landed **outside** the set - i.e. was reported as
recorded, with nothing raised. `write_provenance()` is public and documents its
record shape in full, and required only that the key be *present*, so the whole
path is reachable through public API:

| record a caller supplies | `write_provenance` | `synthetic_episode_indices` |
|---|---|---|
| `synthetic=True` | accepted | `{0}` |
| `synthetic=1` | **accepted** | **`set()`** - reported recorded |
| `synthetic=1.0` | **accepted** | **`set()`** - reported recorded |
| `synthetic="true"` | **accepted** | **`set()`** - reported recorded |
| `synthetic=np.True_` | `TypeError: Object of type bool is not JSON serializable` | - |
| `episode_index=None` | accepted | `TypeError` from `int()` |
| `episode_index="abc"` | accepted | `ValueError: invalid literal for int()` |

Two further consequences of the same gap. A verdict computed with NumPy is a
`numpy.bool_`, which is not `True` and is not JSON-serialisable, so the most
natural way to *produce* a verdict could not be stored at all. And an
`episode_index` that cannot name an episode surfaced as `int()`'s own complaint -
naming neither the file nor the field, and as a `TypeError`, outside anything
this module documents.

`load_provenance()`'s own `Raises:` block already promises this cannot happen:
*"A present-but-unreadable honesty record is corruption, not absence, so it is
never read as 'no synthetic episodes'."* That held for a JSON parse error and for
the top-level shape, and not for a record.

## Change

One rule, consulted by both ends, holding the two keys a reader turns into a
verdict to the shared domains:

- `episode_index` -> `non_negative_whole_number_error` (the domain
  `PolicyRunner.replay` and `record_deterministic_verdicts` already use for an
  episode index)
- `synthetic` -> `boolean_flag_error`
- `transform` keeps its presence-only check

`synthetic` is refused rather than coerced. Every non-empty string and every
non-zero number is truthy, so guessing which of them meant "generated" is how a
generated episode gets counted as recorded; and `synthetic=false` stays a
readable answer that is simply outside the set.

`write_provenance` also stores the two verdict keys in the type the schema
documents (`bool()` / `int()`, on a copy - the caller's dict is untouched), so
the file a reader loads holds the same answer the writer graded. That is what
lets a NumPy verdict round-trip as JSON `true`.

That split - validate the keys a reader turns into a verdict, carry the
descriptive keys through untouched - is the one
`strands_robots.episode_labels.record_deterministic_verdicts` already makes for
the other per-episode sidecar: it validates `episode` on
`non_negative_whole_number_error` and `success` / `failure` on
`boolean_flag_error`, then carries `steps` / `cumulative_reward` / `seed`
through. The two sidecars now agree on what an episode index and a boolean
verdict are.

**Narrowing:** an `episode_index` spelled `"0"` used to be accepted and now
refuses, because that is the same domain's verdict for the same field on every
other episode-index surface in the tree.

## Verification

Measured at `627c794f`.

Real end-to-end run of the shipped transform, then the same records supplied
through the public writer and planted on disk:

provenance record refusal

The control is the whole top half: the shipped transform reads 2 episodes, writes
2, discards 0, and the produced dataset is **byte-identical on both trees** -
`meta/provenance.json` bytes and all 12 generated frames read back out
(`np.array_equal` True). Three rows reported *recorded* before, none after; the
NumPy verdict now stores instead of failing to serialise.

Regression tests (`tests/transforms/test_provenance.py`, extending the file whose
docstring already states this contract): **15 failed / 16 passed** against
`upstream/main`, all 31 pass after. Every pre-fix failure is behavioural and
carries the measurement, e.g.

```
Failed: write_provenance accepted synthetic=1, and synthetic_episode_indices
        then reports set() - the generated episode reads as recorded
Failed: a record carrying synthetic='true' was read as set() - the episode it
        declares generated is reported as recorded, with nothing raised
AssertionError: Regex pattern did not match.
  Expected regex: 'episode_index'
  Actual message: "invalid literal for int() with base 10: 'abc'"
```

Each guard is load-bearing, one break each:

| break | failed | which guards fire |
|---|---|---|
| revert the source | 15 | the whole new set (the pre-fix split) |
| drop the reader half only | 14 | the 5 reader refusals + 6 writer/reader parity cases |
| drop the writer half only | 10 | the 4 writer refusals + 5 parity cases + the **pre-existing** `test_write_refuses_record_missing_mandatory_keys` |
| store the caller's value verbatim | 1 | the NumPy round-trip |
| coerce truthiness instead of refusing | 8 | both non-boolean refusal families |
| also validate the descriptive keys | 1 | the descriptive-key-carried-through control |
| control | 0 of 31 | - |

The writer-half break firing the *pre-existing* writer test is the evidence the
shared rule preserved the original refusal rather than replacing it.

Blast radius: 132 files (every test naming the provenance surface or the shared
domains, all of `tests/transforms`, plus the docs / docstring / string /
changelog guards), the identical selection on both trees with the changed test
file excluded from each:

| tree | result |
|---|---|
| this branch | `2 failed, 5632 passed, 83 skipped in 294.53s` |
| `upstream/main` | `2 failed, 5632 passed, 83 skipped in 294.58s` |

Byte-identical, with the failing-ID sets identical in both directions. The two
shared failures are a local environment artifact, not the diff:
`tests/training/test_lerobot.py::TestInProcessTrainerCorrectness::{test_base_model_warm_start_reads_checkpoint_config,
test_lora_base_model_does_not_preset_use_peft}` fail with `encode() takes 1
positional argument but 2 were given` on draccus 0.10.0, below the
`draccus>=0.11.6` floor lerobot 0.6.2 declares.

`ruff check` and `ruff format --check` clean on the CI scope; `mypy strands_robots
tests tests_integ` reports the same 24 pre-existing errors as a pristine
`upstream/main` worktree, none of them in the changed files. The PR is 4 files:
the module, its test file, the docs contract bullet and a changelog fragment.

Measured at `e24cd82a` (the fragment added the PR number after the PR was
opened; the source and test diff is unchanged from `627c794f`).

### DIFF
--- a/changelog.d/2616-provenance-record-that-cannot-answer.md
+++ b/changelog.d/2616-provenance-record-that-cannot-answer.md
@@ -1,0 +1,25 @@
+### Fixed: a provenance record that cannot answer is refused, not read as recorded
+
+The provenance sidecar exists so generated and recorded episodes stay
+distinguishable, and `synthetic_episode_indices()` is the one call that separates
+them: everything it returns was generated, everything outside it was recorded. It
+read the verdict with `synthetic is True`, so a record whose field was present but
+not a boolean landed outside the set - i.e. was reported as recorded, with nothing
+raised. `write_provenance()` is public and documents its record shape in full, and
+required only that the key be present, so `synthetic=1` or `synthetic="true"` was
+stored and then read as a recorded episode.
+
+The two keys a reader turns into a verdict are now held to the shared domains -
+`episode_index` on `non_negative_whole_number_error`, `synthetic` on
+`boolean_flag_error` - by one rule both `write_provenance()` and
+`load_provenance()` consult, so a record the writer refuses is a record the reader
+refuses. `synthetic` is refused rather than coerced, because every non-empty string
+and every non-zero number is truthy; `synthetic=false` remains a readable answer
+that is simply outside the set. The writer also stores both keys in the type the
+schema documents, which lets a NumPy verdict round-trip as JSON `true` instead of
+failing to serialise, and an `episode_index` that cannot name an episode is now
+named in the refusal rather than surfacing as `int()`'s own complaint.
+
+That split - validate the keys a reader turns into a verdict, carry the descriptive
+keys through untouched - is the one `record_deterministic_verdicts` already makes
+for the other per-episode sidecar.
--- a/docs/data/transforms.md
+++ b/docs/data/transforms.md
@@ -28,7 +28,15 @@ record (strands-robots)  ->  transform (this page)  ->  train (create_trainer)
    episode index, and the transform's name and version. Training filters and
    evaluation read it via `load_provenance()` / `synthetic_episode_indices()`,
    so generated pixels are treated honestly - silent mixing of generated and
-   recorded data is the failure mode this field exists to prevent.
+   recorded data is the failure mode this field exists to prevent. A record that
+   cannot answer what it is read for is refused rather than stored or trusted:
+   `episode_index` must be a non-negative whole number and `synthetic` must be a
+   boolean, checked by one rule that both `write_provenance()` and
+   `load_provenance()` consult. `synthetic` is not coerced, because every
+   non-empty string and every non-zero number is truthy - guessing which of them
+   meant "generated" is how a generated episode ends up counted as recorded. The
+   descriptive keys (`transform_version`, `prompt`, `seed`) are carried through
+   untouched; nothing reads them as a verdict.
 3. **Re-validation is the acceptance gate.** Supply a deterministic verdict
    function and every generated episode is re-scored against its source
    episode's verdict; a generated episode that flips the verdict is discarded
@@ -93,6 +101,12 @@ from strands_robots.transforms import synthetic_episode_indices
 synthetic = synthetic_episode_indices("/data/augmented")
 # everything in `synthetic` was generated; everything outside it was recorded
 ```
+
+An empty set is that statement, not a shrug: a dataset with no
+`meta/provenance.json` declares no synthetic episodes (the ordinary state of a
+recorded dataset), while a file that is present but unreadable raises. Absence
+and corruption are different verdicts, so "outside the set" always means
+recorded.
 
 ## Backends
 
--- a/strands_robots/transforms/provenance.py
+++ b/strands_robots/transforms/provenance.py
@@ -23,10 +23,58 @@ import json
 from pathlib import Path
 from typing import Any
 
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes the `synthetic is True` read gap in the provenance sidecar: a record whose `synthetic` field was present but not a boolean (`1`, `1.0`, `"true"`) previously landed outside `synthetic_episode_indices()` and was silently reported as recorded - exactly the mixing the sidecar's own docstring names as the failure mode. The fix holds the two verdict keys to the shared domains (`episode_index` -> `non_negative_whole_number_error`, `synthetic` -> `boolean_flag_error`) through a single `_record_problem` rule consulted by both `write_provenance()` and `load_provenance()`, stores the verdict keys in the schema-documented types (which lets a `numpy.bool_` verdict round-trip as JSON `true` instead of failing to serialise), and keeps the descriptive keys pass-through. The one narrowing (`episode_index="0"` now refused) is an explicit `ValueError` matching the same domain's verdict on every other episode-index surface, and is disclosed in the description. Verified locally at the head SHA: all 31 provenance tests pass, the full `tests/transforms` suite (231 tests) passes, the ASCII-string and `Raises:`-completeness guards pass, ruff check/format are clean, and the changed files carry no non-ASCII or host paths.

## What's good

- The writer/reader parity test class (`TestBothEndsApplyOneRule`) pins the invariant the fix depends on, so the two ends cannot drift apart in a later edit.
- Refusing rather than coercing `synthetic` follows the `boolean_flag_error` domain rationale in AGENTS.md > "Posture flags are checked, never read by truthiness", and the validate-verdict-keys / carry-descriptive-keys split matches the existing `record_deterministic_verdicts` precedent.
- Validation completes before any file is written, so a refused record leaves nothing on disk; the caller's dicts are copied, not mutated.
- Scope discipline: 4 files - the module, its test file, the docs contract bullet, and a correctly named changelog fragment (no direct `CHANGELOG.md` edit).

PR #2617, opened 2026-08-22
fix(transforms): every input the determinism key is spread from is held to its domain

cagataycali:
## Problem

`derive_variant_seed` spreads the triple `(seed, source_episode, variant)` through one
`numpy.random.SeedSequence`, so all three inputs name the stream a generated variant is
rendered from. Only one of the three was checked.

The consequence is the one the function's own `Raises:` block already names for
`source_episode` - "unrefused it derives episode 1's seed, so a variant of episode `True`
silently collides with a variant of episode 1" - reached through the other two inputs.
Measured on mid-grey frames through `MockTransform`, and on the raw key:

| input | pre-fix outcome | collided with |
| --- | --- | --- |
| `variant=True` | pixel 110, no refusal | `variant=1` (pixel 110) |
| `variant=False` | pixel 121, no refusal | `variant=0` (pixel 121) |
| `variant="1"` | pixel 110, no refusal | `variant=1` |
| `seed=True` | key `1835504127`, no refusal | `seed=1` |
| `seed="1"` | key `1835504127`, no refusal | `seed=1` |
| `variant=-1` | `ValueError: expected non-negative integer` (NumPy) | - |
| `variant=2.5` | `TypeError: seed must be integer` (NumPy) | - |
| `variant=None` | `TypeError: object of type 'NoneType' has no len()` | - |

`np.array_equal` on the two renderings is `True` for the collision rows: two "distinct"
variants of one episode were written into the augmented dataset as two episodes whose
pixels are byte-identical, under provenance records naming different `variant` counters -
so the data-multiplication factor the transform surface exists for was silently 1 for that
pair. NumPy coerces a str spelling of a whole number, which is why `"1"` behaves as `1`.

The three rows NumPy refuses on its own are the other two classes the neighbouring
episode-index sweep records: the message named neither the parameter nor the surface, and
two of them are `TypeError`, which is not the `ValueError` these surfaces document as their
refusal channel.

At the backend seams the same counter degraded twice over:

| surface | pre-fix, unusable `variant` |
| --- | --- |
| `mock.transform_frames(pixel_shift=3)` | accepted every value, shifted the pixels - that path derives no key, so the counter was never read at all |
| `cosmos_transfer.transform_frames`, no pipeline | `RuntimeError: no video2video pipeline is bound` - a wiring diagnosis for a bad counter, naming the seam the caller had got right |
| `cosmos_transfer.transform_frames`, pipeline bound | `True` / `"1"` accepted and handed to `generate`; `-1` / `2.5` / `None` raised NumPy's internals |

## Change

Apply the shared `non_negative_whole_number_error` rule to all three inputs of the key in
`derive_variant_seed`, keeping `seed=None` as the documented opt-out from determinism
rather than a stream name. The `int()` coercion moves *after* the guard, never before,
which is the ordering that shared rule's own docstring asks for ("the caller coerces the
accepted value with `int()`; that coercion is safe *because* this guard has already
performed it and compared the result back").

Both backend seams refuse the counter too, for the reason their `source_episode` guards
already exist:

- a backend need never reach `derive_variant_seed` - `mock`'s explicit `pixel_shift` mode
  derives no key - so a guard only in the shared helper leaves that path unchecked;
- the refusal should name the counter rather than whatever the pipeline seam complains
  about first, which is what `test_the_cosmos_backend_refuses_before_the_pipeline_binding_check`
  already pins for the episode index.

Docs: `docs/data/transforms.md` "Custom backends" now states the one obligation
`transform_frames` owns beyond the pixels, with the two-line loop to copy, and says why it
is needed in the backend as well as in the shared helper.

## Scope

`transform()` is unaffected: it derives `variant` from `range(int(spec.variants_per_episode))`
and `_spec_problems` already holds `variants_per_episode`, `episodes[]` and `seed` to their
domains, so every value reaching the key through the orchestration was already clean. The
reachable surface is a direct caller of the public `derive_variant_seed` or of
`transform_frames` - which is exactly the reachability the shipped `source_episode` guard
is justified by ("`transform_frames` is the documented seam a backend author calls
directly, so it refuses the same domain rather than disagreeing with the orchestration
that validated `spec.episodes` upstream").

The structural sweep in `tests/simulation/test_replay_episode_index_domain.py` is untouched:
its scope is the episode-index quantity, and a variant counter is a different one.

## Tests

`tests/transforms/test_frame_surface_episode_domain.py`, extended from the
`source_episode` half to the whole key, with the measurement table in its module docstring.

Copied onto `upstream/main`: **50 failed / 47 passed**, all behavioural - the file imports
only pre-existing public symbols, so nothing fails on a missing name. Sharpest lines:
`DID NOT RAISE <class 'ValueError'>` for the collision rows, and
`TypeError: seed must be integer` / `object of type 'NoneType' has no len()` escaping a
surface documented to raise `ValueError`.

The 47 that pass on both trees are the controls, and each fails for a specific wrong fix:

| control | fails if |
| --- | --- |
| `test_an_accepted_triple_derives_the_key_it_derived_before` | the coercion moves before the guard, or what is spread into the sequence changes |
| `test_none_seed_is_still_the_documented_opt_out` | `seed=None` is read as a value rather than as absence |
| `test_accepted_variants_still_name_distinct_streams` | the guard narrows the variance the key exists for |
| `test_two_accepted_variants_of_one_episode_still_differ` | as above, at the pixel level |
| `test_an_integral_float_is_honored_on_every_input` | the new inputs are held to a narrower domain than `source_episode` |
| `test_a_bound_pipeline_still_sees_an_accepted_variants_pixels` | the guard refuses a legitimate call, or stops forwarding the derived key |

Break table on the fixed tree - each break fires exactly and only its own guards:

| break | fails | which |
| --- | --- | --- |
| revert all three files to `upstream/main` | 50 | the whole regression set |
| drop the `variant` guard from `derive_variant_seed` | 16 | the key + derived-shift mock cases |
| drop the `seed` guard from `derive_variant_seed` | 8 | only the two `seed` tests |
| drop the inline `variant` guard from `mock.transform_frames` | 6 | only `test_the_explicit_shift_mode_refuses_too` |
| drop the inline `variant` guard from `cosmos_transfer.transform_frames` | 6 | only `test_the_cosmos_backend_refuses_before_the_pipeline_binding_check` |
| coerce with `int()` before the guard instead of after | 19 | the shared-verdict and accepted-key controls |
| control (fixed tree) | 0 of 97 | - |

The last two rows of the first block are the reason both inline guards exist rather than
only the shared one. Dropping the mock guard fails **only** the explicit-shift class - the
derived-shift path still reaches `derive_variant_seed` and is still covered - and dropping
the cosmos guard fails **only** the unbound-pipeline class, where the `RuntimeError` would
otherwise answer first. They are complementary, not redundant.

## Artifact

The change is to which values the pixel-producing seam accepts, so the artifact is the
renderings themselves: `MockTransform` on a uniform mid-grey source at `seed=29`,
`source_episode=0`, one tile per variant counter (the mock adds a per-variant constant, so
each tile is that variant's whole frame).

per-variant renderings before and after

Before, `variant=False` renders pixel 71 - variant 0's frame - and `variant=True` and
`variant="1"` both render pixel 192, variant 1's. Each is an extra generated episode
duplicating one already written. After, those three are refused and variants 0/1/2 are
unmoved at 71/192/134.

The console below the strip is the control: the same real end-to-end run - a recorded
LeRobotDataset in, an augmented LeRobotDataset out - on both trees. `status=success`,
`episodes_read=2`, `episodes_written=4`, `episodes_discarded=0`, the four
`meta/provenance.json` records identical field for field including their derived seeds, and
the first pixel of every written episode read back out of the produced dataset identical at
`[51, 39, 102, 80]`. The figure script asserts each of those from the captured JSON.

## Verification

Measured at `cc9ba885`.

- `tests/transforms/test_frame_surface_episode_domain.py`: **97 passed**, up from 37 collected.
- Blast radius, the identical 85-file selection (every test naming the transform surfaces,
  the shared rule, `variants_per_episode`, or reading the docs tree) run on this branch and
  on a pristine `upstream/main` worktree, with this file excluded from both so the
  comparison measures the source change alone: **3276 passed / 55 skipped / 0 failed,
  byte-identical on both**, 200.66s vs 202.33s.
- Every triple accepted before derives the identical key after: `(7,0,0) 2083679832`,
  `(7,0,1) 2028854884`, `(7,0,2) 3835300069`, `(7,2,1) 1060258595`, `(11,2,1) 2549836301`,
  `(None,0,0) None`, and `np.int64` unchanged. The one behaviour change on the accepted
  side is that an integral float is now honored on `variant`/`seed` exactly as it already
  was on `source_episode`.
- The mock's derived shift on accepted variants is unchanged in both trees:
  `variant 0/1/2 -> pixel 121/110/187`.
  `mypy`: 24 == 24 against the pristine base, branch-only empty.

### DIFF
--- a/changelog.d/2617-determinism-key-input-domain.md
+++ b/changelog.d/2617-determinism-key-input-domain.md
@@ -1,0 +1,41 @@
+### Fixed: every input the determinism key is spread from is held to its domain
+
+`derive_variant_seed` spreads the triple `(seed, source_episode, variant)` through one
+`numpy.random.SeedSequence`, so all three name the stream a generated variant is rendered
+from. Only `source_episode` was checked, and the other two degraded exactly the way that
+function's own `Raises:` block already warns about for the one that was.
+
+Measured on mid-grey frames through `MockTransform`: `variant=True` derived pixel 110,
+byte-identical to `variant=1`; `variant="1"` the same, because NumPy coerces a str spelling
+of a whole number; `variant=False` matched `variant=0`; and `seed=True` and `seed="1"` both
+derived `seed=1`'s key. Two "distinct" variants of one episode were therefore written into
+the augmented dataset as two episodes whose pixels are identical, under provenance records
+naming different `variant` counters - so the data-multiplication factor the transform
+surface exists for was silently 1 for that pair, with no refusal anywhere.
+
+The values NumPy refuses on its own were no better placed: `-1`, `2.5` and `None` arrived as
+its internal `expected non-negative integer` / `seed must be integer` /
+`object of type 'NoneType' has no len()`, naming neither the parameter nor the surface, and
+two of those are `TypeError` rather than the `ValueError` these surfaces document as their
+refusal channel.
+
+All three inputs now go through the shared `non_negative_whole_number_error` rule, with
+`seed=None` kept as the documented opt-out from determinism rather than a stream name. The
+`int()` coercion moves after the guard, never before, which is the ordering that rule's own
+docstring asks for. Every triple accepted before derives the identical key; the only change
+on the accepted side is that an integral float is honored on `variant` and `seed` exactly as
+it already was on `source_episode`.
+
+Both backend seams refuse the counter too, for the reason their `source_episode` guards
+already exist. A backend need never reach `derive_variant_seed` at all - `mock`'s explicit
+`pixel_shift` mode derives no key, so it shifted the pixels without reading the counter once
+for any value above - and the refusal should name the counter rather than whatever the
+pipeline seam reaches first: unbound, `cosmos_transfer` reported an unusable counter as "no
+video2video pipeline is bound", a wiring diagnosis for the one thing the caller had got
+right.
+
+`transform()` is unaffected either way: it derives `variant` from
+`range(int(spec.variants_per_episode))` and the shared spec preflight already holds
+`variants_per_episode`, `episodes[]` and `seed` to their domains, so the reachable surface is
+a direct caller of the public helper or of the backend seam - the same reachability the
+shipped `source_episode` guard is justified by.
--- a/docs/data/transforms.md
+++ b/docs/data/transforms.md
@@ -147,3 +147,24 @@ register_transform("my_v2v", lambda: MyTransform)
 
 The base class owns the dataset plumbing, pass-through, provenance and
 re-validation gate, so a backend cannot accidentally weaken them.
+
+`transform_frames` is called with the determinism key's two per-call inputs,
+`source_episode` and `variant`, and owns one obligation of its own: refuse a
+value either is not. Both are non-negative whole numbers - together with
+`spec.seed` they are spread through one `SeedSequence` by `derive_variant_seed`,
+so an unusable value on any of the three names a stream another variant already
+owns rather than failing:
+
+```python
+from strands_robots.utils import non_negative_whole_number_error
+
+for name, value in (("source_episode", source_episode), ("variant", variant)):
+    if text := non_negative_whole_number_error(value, name, "my_v2v.transform_frames"):
+        raise ValueError(text)
+```
+
+`derive_variant_seed` applies the same rule to all three, so a backend that
+always derives a key inherits it - but refuse in `transform_frames` too, because
+a backend need not derive one at all (`mock`'s explicit `pixel_shift` mode reads
+no key), and because the refusal should name the counter rather than whatever
+the pipeline seam happens to complain about first.
--- a/strands_robots/transforms/base.py
+++ b/strands_robots/transforms/base.py
@@ -196,26 +196,42 @@ def derive_variant_seed(seed: int | None, source_episode: int, variant: int) -> 
     same spec always reproduces the same output.
 
     Args:
-        seed: The spec's base seed (:attr:`TransformSpec.seed`).
+        seed: The spec's base seed (:attr:`TransformSpec.seed`), a non-negative
+            whole number or ``None`` to opt out of determinism.
         source_episode: Source episode index the variant is generated from.
         variant: Variant counter within that episode
-            (``0 .. variants_per_episode - 1``).
+            (``0 .. variants_per_episode - 1``), a non-negative whole number.
 
     Returns:
         A 32-bit seed, or ``None`` when ``seed`` is ``None``.
 
     Raises:
-        ValueError: ``source_episode`` is outside the non-negative
-            whole-number domain every episode-resolving surface shares.
-            ``True`` is the value worth naming: unrefused it derives episode
-            1's seed, so a variant of episode ``True`` silently collides with
-            a variant of episode 1.
+        ValueError: Any of the three inputs is outside the non-negative
+            whole-number domain the key needs (``seed=None`` excepted - that
+            spelling opts out of determinism rather than naming a stream).
+            Each is checked because each spreads into the same
+            :class:`~numpy.random.SeedSequence`, so an unusable value on any
+            of them yields a key some other triple already owns: ``True`` is
+            the value worth naming, since unrefused it is ``1`` to NumPy, and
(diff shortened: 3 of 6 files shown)


yinsong1986 (approved):
## Summary

Closes the determinism-key domain gap in `derive_variant_seed`: the triple `(seed, source_episode, variant)` is spread through one `SeedSequence`, and only `source_episode` was guarded, so `variant=True`/`"1"` and `seed=True`/`"1"` silently derived another triple's stream (two "distinct" variants written with byte-identical pixels), while `-1`/`2.5`/`None` escaped as NumPy-internal `TypeError`s on a surface documented to raise `ValueError`. The fix applies the shared `non_negative_whole_number_error` rule to all three inputs (keeping `seed=None` as the documented determinism opt-out), coerces with `int()` only after the guard per that rule's own contract, and adds the same inline `variant` guard at both backend seams — justified independently, since `mock`'s explicit `pixel_shift` mode never reaches the shared helper and `cosmos_transfer` otherwise reports an unusable counter as a pipeline-wiring problem. Verified locally at the head: the extended test file passes 97/97, the whole `tests/transforms/` + episode-index sweep selection passes 493/493, every accepted triple derives the identical pre-fix key (including `np.int64`, integral floats, and an arbitrarily large int through `SeedSequence`), all refusals name the parameter and surface, and the changed files are ASCII-clean, host-path-free, and `ruff` clean.

## What's good

- The break table in the description holds up: the two inline backend guards are each pinned by a test class only that guard can satisfy (explicit-shift mode for `mock`, refusal-outranks-`RuntimeError` for `cosmos_transfer`), so neither is redundant with the shared helper.
- Accepted-side stability is pinned with pre-fix key values, so a later reordering of guard and coercion fails loudly.
- `Raises:` blocks updated on all three surfaces in the same change (AGENTS.md > Review Learnings #85 > "A `Raises:` block names every refusal the function itself makes"), changelog fragment present under `changelog.d/`, docs state the backend-author obligation with the copyable loop.
- The previously-accepted collision spellings now fail with an explicit `ValueError` naming parameter and surface — a loud refusal, not a silent default change.

yinsong1986 (approved):
## Summary

Closes the remaining two-thirds of the determinism-key domain gap: `derive_variant_seed` spreads `(seed, source_episode, variant)` through one `SeedSequence`, and only `source_episode` was guarded, so `variant=True`/`"1"` and `seed=True`/`"1"` silently derived another triple's stream (two "distinct" variants written with byte-identical pixels), while `-1`/`2.5`/`None` escaped as NumPy-internal `TypeError`/`ValueError` naming neither parameter nor surface. The fix routes all three inputs through the shared `non_negative_whole_number_error` rule with `seed=None` preserved as the documented determinism opt-out, coerces with `int()` only after the guard (the ordering the shared rule's docstring requires), and adds the matching inline `variant` guards at both backend seams for the same reachability reason the shipped `source_episode` guards exist - `mock`'s explicit `pixel_shift` path derives no key at all, and `cosmos_transfer`'s unbound-pipeline `RuntimeError` would otherwise misdiagnose a bad counter as a wiring problem. Verified locally at `cc9ba885`: the extended test file passes 97/97, every previously-accepted triple derives the identical key (including `np.int64` inputs and the `None` opt-out), the bool/str collision rows now refuse with the shared verdict, and all changed files are ASCII-clean. `Raises:` blocks on all three surfaces were updated in the same change, the changelog fragment and docs obligation for custom backends are present, and the orchestration path (`transform()`) was already clean via `_spec_problems`, so the tightening only reaches direct callers of the public helper or the backend seam - which is the surface the fix exists for.

## What's good

- Guard-before-coercion ordering matches the shared rule's own contract, and the pre-fix key table pins that no accepted stream moved.
- The break table demonstrates each of the four guards (shared x2, inline x2) is independently load-bearing rather than redundant.
- `Raises:` completeness, changelog fragment, and docs updated in the same PR per AGENTS.md.

PR #2618, opened 2026-08-22
fix(transforms): the pipeline seam reports a failed probe instead of raising it

cagataycali:
## Problem

`CosmosTransferTransform` binds a caller-supplied video2video pipeline, so resolving the seam runs the caller's code at three points: the module import and attribute lookup, the zero-arg construction of a class / factory target, and the read of the object's `generate` surface.

`DatasetTransform.validate` is documented as a *"Pure, side-effect-free preflight. Return a list of human-readable problems"*, and `transform()` documents exactly one raise (a shape / dtype `ValueError` from `transform_frames`), `status="error"` otherwise. `_pipeline_problems`' own docstring says *"return problems instead of raising"*.

It reported three classes: `ImportError` and `AttributeError` around the import, `TypeError` around the construction. Driving every builtin exception class through a raising factory, **51 of 52 escaped `validate()` - and therefore `transform()`.** Only `TypeError` was reported.

The classes that escaped are the ones a real generation stack raises, because constructing one loads weights and touches a device:

| probe point | raised | `validate()` | `transform()` |
| --- | --- | --- | --- |
| module import + attribute lookup | `ImportError` | 1 problem | `status="error"` |
| zero-arg construction | `ImportError` (a factory body's `import torch`) | **RAISED** | **RAISED** |
| zero-arg construction | `RuntimeError` (no NVIDIA driver) | **RAISED** | **RAISED** |
| zero-arg construction | `OSError` (weights absent) | **RAISED** | **RAISED** |
| `generate()` surface read | `RuntimeError` (lazy handle, CUDA context) | **RAISED** | **RAISED** |

Row 1 is the control - the import point already handled the one class its handler named. Row 2 is **the same class**, raised one frame lower: a factory whose body imports an optional dependency raises `ModuleNotFoundError`, an `ImportError` subclass, straight past a handler written for exactly that case, because the handler wrapped only the module import and not the factory call.

Two further consequences worth naming:

- The `ValueError` case is indistinguishable from the shape / dtype `ValueError` that `transform()` documents as its only raise, so a malformed checkpoint config reads as a backend bug.
- `getattr(obj, "generate", None)`'s default absorbs only `AttributeError`, so a lazy pipeline handle whose `generate` is a property raised at the third point - reachable with no import path at all, by injecting the object directly.

The package's own module docstring teaches the shape where this lands: `problems = transform.validate(spec)` / `if not problems: result = transform.transform(spec)`.

## Change

Each of the three probe points now reports, and the cause is named accurately.

- The import and attribute lookup catch `Exception`: a module body runs arbitrary caller code on first import, and a module-level `__getattr__` (PEP 562, which this package uses in its own `__init__.py`) runs it again on the lookup.
- The zero-arg construction keeps `except TypeError` with its existing wording, and adds a sibling handler for everything else. A factory that raised **is** constructible zero-arg, so folding it into that wording would name the wrong remedy - "no driver" is not fixed by a different factory signature.
- The `generate` read moves into one guarded owner, `_generate_surface`, used at both the injected-object and resolved-candidate sites. It returns `(surface, problem)` and distinguishes "absent" from "present but `None`" with a sentinel, so the construct decision is byte-equivalent to the `hasattr` it replaces.

`Exception`, not `BaseException`: an operator's Ctrl-C while a multi-gigabyte pipeline loads is not a bad spec, and `KeyboardInterrupt` / `SystemExit` / `GeneratorExit` still propagate.

Every spelling that already reached a verdict keeps it, wording included.

The in-repo precedent for resolving a caller-supplied dotted path is a wide handler: `tools/use_lerobot.py::_import_from_lerobot` catches `Exception` with the comment *"non-Import errors during import are real too"*, and `policies/lerobot_local/resolution.py` pairs `except ImportError` with `except (AttributeError, RuntimeError, TypeError)` so a failure means *"this strategy cannot resolve the class here"* rather than *"leaking lerobot's internal exception type to the caller"*.

`docs/data/transforms.md` documented only the missing-pipeline case; it now states the same contract for a pipeline that is named but cannot be loaded.

## Out of scope

`_pipeline_problems` resolves and constructs *inside* `validate()`, which the ABC says MUST NOT allocate GPUs. Whether the resolution should move out of the preflight (and what `transform_frames`' existing direct-caller `RuntimeError` then means) is a design decision, not a reporting one, so it is left alone here.

## Tests

`tests/transforms/test_cosmos_transfer.py`, extended in place. Pre-fix (the new tests against `upstream/main` source): **27 failed / 33 passed**. Of the 27, **24 fail by the third-party exception escaping** (`RuntimeError` 8, `ValueError` 3, `OSError` 3, `KeyError` 3, `FileNotFoundError` 3, `ModuleNotFoundError` 2, `ImportError` 2) and 3 by assertion.

Each break below was applied to the fixed tree and fires exactly and only its own guards:

| break | failed of 60 | guards fired |
| --- | --- | --- |
| control | 0 | - |
| `git checkout upstream/main -- cosmos_transfer.py` | 27 | the headline, the documented run path, the accurate-cause test, the injected-object test, both structural |
| widen the handlers to `BaseException` | 11 | **only** the operator-interrupt controls + the handler pin |
| fold construction failures into the zero-arg wording | 1 | **only** the accurate-cause test |
| re-derive the `generate` read inline | 1 | **only** the single-owner pin |
| revert just the import handler | 7 | **only** the import-point cases + the handler pin |
| move `import_module` outside its `try` | 3 | the guarded-import pin, and the **pre-existing** `test_unresolvable_dotted_path_is_a_problem_not_a_raise` |

The last row is the one worth reading: the existing test whose name already states this contract (*"is a problem not a raise"*) still holds, because it exercised the module import - the one path that already worked. The fix extends that contract to the other two points rather than replacing it.

## Verification

Measured at `b6cff7f7`, on `mujoco 3.5.0` / `lerobot 0.6.2`. The branch was merged with `main` after #2617 landed in the two files this touches; the merge was clean, and every number below was re-measured at that head (the pre-fix split is unchanged).

- `ruff check` and `ruff format --check` over `strands_robots tests tests_integ`: clean.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `upstream/main` worktree, branch-only set empty.
- 77-file blast radius (all of `tests/transforms`, every test naming the transform surface, and the repo-wide docs / docstring / string / changelog guards), the identical selection run on both trees with the changed test file excluded from both: **2303 passed, 1 skipped, 0 failed on each**, failing-ID sets empty in both directions.
- `tests/transforms/test_cosmos_transfer.py`: 60 passed.

## Artifact

Produced by one capture script run against each tree, with only `strands_robots` differing. Section 1 is the control: a working pipeline, a real recorded `LeRobotDataset` in and a provenance-marked one out - 2 episodes read, 2 written, 1 MP4, and the written pixels read back **out** of the produced dataset. That whole section is byte-identical across the two trees (asserted in the figure script, along with every number quoted here), so the only thing that changes is what a caller is handed when the seam fails.

pipeline seam probe reporting

### DIFF
--- a/changelog.d/2618-pipeline-seam-probe-reports.md
+++ b/changelog.d/2618-pipeline-seam-probe-reports.md
@@ -1,0 +1,22 @@
+### Fixed: a failed pipeline resolution is reported by the transform preflight, not raised at the caller
+
+`CosmosTransferTransform` binds a caller-supplied video2video pipeline, so resolving the seam runs the
+caller's code at three points: the module import and attribute lookup, the zero-arg construction of a
+class or factory target, and the read of the object's `generate` surface. `validate()` is documented to
+return a list of problems and `transform()` to return `status="error"` for anything but a backend
+shape/dtype bug, but only `ImportError`, `AttributeError` and `TypeError` were reported - 51 of 52 builtin
+exception classes raised out of both surfaces.
+
+The classes that escaped are the ones a real generation stack raises, because constructing one loads
+weights and touches a device: `ImportError` from an optional dependency imported inside a factory body
+(so `ModuleNotFoundError` escaped a handler written for exactly that case, which wrapped only the module
+import), `RuntimeError` with no driver, `OSError` with the weights absent, and `ValueError` on a malformed
+config - the last indistinguishable from the shape/dtype `ValueError` that `transform()` documents as its
+only raise. A lazy handle whose `generate` is a property raised at the third point, where `getattr`'s
+default absorbs only `AttributeError`.
+
+Each probe point now reports, with the cause named accurately: a factory that raised is constructible, so
+it is not folded into the "not constructible zero-arg" wording, which would name the wrong remedy. The
+`generate` read has one guarded owner. `Exception`, not `BaseException` - an operator's Ctrl-C during a
+multi-gigabyte load is not a bad spec and still propagates. Every spelling that already reached a verdict
+keeps it, wording included.
--- a/docs/data/transforms.md
+++ b/docs/data/transforms.md
@@ -148,6 +148,17 @@ spec = TransformSpec(
 Without a pipeline, `validate()` names the missing seam (and the licensing
 caveat) instead of crashing; nothing is read or written.
 
+The same holds for a pipeline that is named but cannot be loaded. Resolving the
+seam runs your code at three points - the module import and attribute lookup,
+the zero-arg construction of a class or factory target, and the read of the
+object's `generate` surface - and constructing a real generation pipeline loads
+weights and touches a device. So a missing optional dependency imported inside
+a factory body, an absent driver, absent weights or a malformed config are
+reported by `validate()` as problems, and by `transform()` as
+`status="error"`, each naming the class and message the pipeline raised. An
+operator interrupt (`KeyboardInterrupt`, `SystemExit`) is not a spec problem
+and still propagates.
+
 ## Custom backends
 
 Subclass `DatasetTransform`, implement `provider_name`, `validate` (call
--- a/strands_robots/transforms/cosmos_transfer.py
+++ b/strands_robots/transforms/cosmos_transfer.py
@@ -89,6 +89,38 @@ _PIPELINE_HELP = (
     "wiring it in."
 )
 
+#: Sentinel for "the pipeline exposes no ``generate`` attribute at all", kept
+#: distinct from a ``generate`` that is present but ``None`` so the probe below
+#: classifies both exactly as ``hasattr`` / ``getattr`` did.
+_UNSET: Any = object()
+
+
+def _generate_surface(obj: Any, subject: str) -> tuple[Any, str | None]:
+    """Read ``obj.generate`` for the probe, or report why the read failed.
+
+    The pipeline is caller-supplied, so reading an attribute off it runs the
+    caller's code: a lazy handle that initialises a CUDA context on first
+    access raises here, and ``getattr``'s default only absorbs
+    ``AttributeError``. Reporting keeps this probe inside
+    :meth:`~strands_robots.transforms.base.DatasetTransform.validate`'s
+    "return problems" contract.
+
+    Args:
+        obj: The candidate pipeline - an injected object or a resolved import
+            target.
+        subject: How the caller named it, for the problem text.
+
+    Returns:
+        ``(surface, None)`` where ``surface`` is the attribute or :data:`_UNSET`
+        when absent, or ``(_UNSET, problem)`` when the read itself raised.
+    """
+    try:
+        return getattr(obj, "generate", _UNSET), None
+    except Exception as exc:  # noqa: BLE001 - never fatal while probing a caller's object
+        return _UNSET, (
+            f"{subject} raised {type(exc).__name__} while its generate() surface was read ({exc}) - {_PIPELINE_HELP}"
+        )
+
 
 class CosmosTransferTransform(DatasetTransform):
     """Video2video episode augmentation behind a vendor-neutral pipeline seam.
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

This PR closes a documented-contract gap in `CosmosTransferTransform._pipeline_problems`: `validate()` is documented as a pure preflight that returns problems, but three probe points on the caller-supplied pipeline seam (module import + attribute lookup, zero-arg construction, `generate` surface read) let 51 of 52 builtin exception classes escape both `validate()` and `transform()`. The fix widens the import handler to `Exception` (with `BaseException` interrupts deliberately still propagating), adds a sibling construction handler that names the accurate cause instead of folding a raising factory into the "not constructible zero-arg" wording, and centralises the `generate` read into one guarded owner (`_generate_surface`) with an `_UNSET` sentinel keeping the construct decision byte-equivalent to the `hasattr` it replaces. The `except Exception` handlers sit squarely inside the "return problems, never raise" contract (the report-not-raise pattern from AGENTS.md's error-handling learnings), carry `noqa: BLE001` with explanatory comments, and match the in-repo precedent for resolving caller-supplied dotted paths (`tools/use_lerobot.py`, `policies/lerobot_local/resolution.py`). No public API, wire format, or persisted schema changes — `_UNSET` and `_generate_surface` are module-private. Changelog fragment present, docs updated for the new contract, diff is ASCII-clean, and the changed test file passes 60/60 on the head tree.

## What's good

- Operator-interrupt controls (`KeyboardInterrupt` / `SystemExit` / `GeneratorExit` still raise) pin the `Exception`-not-`BaseException` boundary at every probe point, so a later widening fails loudly.
- The verdict-preservation parametrization pins that every previously accepted spelling keeps its exact verdict and wording.
- Structural AST tests (single guarded owner for the `generate` read, `import_module` never outside a `try`) hold a fourth probe point added later to the same rule rather than covering only today's three.
- The construction handler keeps `except TypeError` with its existing wording and reports everything else with an accurate subject — the "no driver" case is not misdiagnosed as a signature problem.
- Changelog fragment (`changelog.d/2618-pipeline-seam-probe-reports.md`) per step 3 of the PR workflow; `docs/data/transforms.md` states the new contract alongside the existing missing-pipeline case.

yinsong1986 (approved):
## Summary

This PR makes `CosmosTransferTransform._pipeline_problems` honour the documented `validate()` contract ("return a list of problems, never raise") at all three points where resolving a caller-supplied pipeline runs the caller's code: the module import + attribute lookup, the zero-arg construction, and the `generate` surface read. Previously only `ImportError`/`AttributeError` (import point) and `TypeError` (construction) were reported; every other class - including the ones a real generation stack raises (`RuntimeError` with no driver, `OSError` with weights absent, `ModuleNotFoundError` from a factory body) - escaped both `validate()` and `transform()`. The handlers widen to `except Exception` (not `BaseException`, so `KeyboardInterrupt`/`SystemExit`/`GeneratorExit` still propagate - pinned by dedicated tests), the construction failure gets its own accurately-worded report rather than being folded into the "not constructible zero-arg" wording, and the `generate` read moves into a single guarded owner (`_generate_surface`) with a sentinel keeping the absent-vs-`None` distinction byte-equivalent to the `hasattr`/`getattr` it replaces.

## What's good

- The wide catch is on a report-don't-raise preflight path with in-repo precedent cited (`tools/use_lerobot.py`, `policies/lerobot_local/resolution.py`), each site carries a `# noqa: BLE001` with a reason, and the operator-interrupt propagation is behaviourally pinned per probe point.
- Every previously-reached verdict keeps its wording, and a parametrized test pins that no accepted spelling changed - so this is purely additive on the refusal surface, with no one-way door.
- Structural AST tests pin the rule itself (single guarded owner of the `generate` read, `import_module` never outside a `try`), so a fourth probe point added later is held to the same contract.
- Changelog fragment (`changelog.d/2618-...md`) and the `docs/data/transforms.md` contract statement ship in the same PR; test names describe behaviour, not provenance; problem strings are plain ASCII.

PR #2619, opened 2026-08-22
fix(policies/protomotions): a body index is refused unless it addresses a body

cagataycali:
## Problem

`ProtoMotionsConfig.anchor_body_index` and `root_body_index` are offsets into
`body_names` - the tracker reads a body out of the reference-motion cache by row
index, never by name. `load_config_from_yaml` documents its result as "validated
for consistent dimensions" and promises `ValueError` for "an inconsistent
dimension"; the module docstring promises such an error "surfaces at policy build
time with a clean message". It checked two dimensions - the `stiffness` and
`damping` lengths against the joint count - and neither index.

The two ways an index misses are not symmetric, and the silent one is a valid
Python lookup:

* A **negative** index wraps. `body_names[-1]` is `right_rubber_hand`, a real
  link, so nothing raises anywhere and the tracker anchors on the hand. The whole
  chain agrees: `required_bodies` declares that link, the runtime resolves it once
  per rollout and merges `body.right_rubber_hand.quat` into every observation,
  `_extract_anchor_rot` finds exactly that key, and the future-reference window
  slices the same row out of the cache. Every stage is consistent, and every stage
  is wrong.
* A **positive out-of-range** index deferred `IndexError: tuple index out of
  range` to whichever property read the name first - naming neither the field, the
  value, the valid range, nor the sidecar it came from.

Measured on the tracker's own G1 embodiment (`g1_holo_compat.xml`, whose 33 bodies
are `GTP_G1_BODY_NAMES` in order), the anchor orientation a `-1` sidecar feeds the
network, against the one index 16 declares:

| pose | resolved anchor | degrees from `torso_link` |
| --- | --- | --- |
| waist 0.0 | `right_rubber_hand` | 0.01 |
| waist 0.3, arm raised | `right_rubber_hand` | 10.37 |
| waist 0.6, arm raised | `right_rubber_hand` | 20.20 |

The `int()` coercion in the loader widened the same hole: it ran *before* anything
looked at the value, so a yaml `anchor_body_index: true` became row 1 (`head`,
34.4 degrees out at waist 0.6) and a `2.7` became row 2
(`left_hip_pitch_link`). Both loaded, reported nothing, and disagreed with a
hand-built `ProtoMotionsConfig(anchor_body_index=True)`, which took the same
value through a different path.

This is the config-side mirror of a hole the package already documents and
refuses on the model side: `docs/policies/protomotions.md` explains that a
30-body G1 shifts every row after the gap, so "the tracker asks for `torso_link`
at row 16 and is handed `left_shoulder_pitch_link` - on a walking clip, an anchor
orientation some 20 degrees out, with nothing in the cache to say so." The model
that produces that shift is refused. The sidecar that names the same wrong row
was not.

## Change

`ProtoMotionsConfig.__post_init__` now holds both indices to the shared
`non_negative_whole_number_error` domain and then to the row count of the body
list they index, raising `ValueError` either way:

```text
ValueError: ProtoMotionsConfig.anchor_body_index must be a row of body_names
(0..32), got 99 for 33 bodies. The index is an offset into body_names, so one
that misses cannot resolve the body it names.
```

Validating in `__post_init__` rather than in the loader makes it one owner for
both routes - a sidecar and a hand-built config now report the same value the same
way - and matches the four sibling policy configs (`kimodo`, `motionbricks`,
`vera`, `wbc`) that already validate there; `protomotions` was the one that did
not. The domain runs **before** the `int()` normalisation, which is the ordering
`MotionBricksConfig.__post_init__` states the reason for in as many words
("Before the `float()` below, not after: that conversion is what used to launder
..."). The loader therefore hands the raw yaml value through instead of coercing
it first, and an integral float such as `16.0` still addresses a row and is kept,
normalised to the row number both consumers index with.

`anchor_body_name` and `root_body_name` documented `Raises: IndexError`. That is
now unreachable, so both docstrings say so instead.

**Scope.** Only the two indices, because they are the only fields this config
resolves a body NAME from and the only unvalidated inputs whose wrongness is
silent. Measured and deliberately left alone: `future_step_indices` accepts `[]`,
`[0]` and `[-1, 2]`; `decimation` is documented as `control_dt / physics_dt` and a
sidecar may disagree with its own timing; `control_dt = 0.0` surfaces as
`ZeroDivisionError` from the load log line and `-0.02` is accepted;
`action_ema_alpha` is unbounded. Each of those is a value-domain or
derived-relation question rather than "an index that cannot address a body", and
each wants its own decision. The sidecar also carries `robot.anchor_body_name`
beside the index, which the loader ignores; cross-checking the two would refuse a
disagreement, but which of them wins is a contract choice, not a range.

## Visual

Same sidecar slip, same embodiment, same three poses, run through each tree's own
`ProtoMotionsConfig`. The red frame marks the link the config resolved as the
anchor - the orientation the tracker consumes - and the green frame marks the
declared anchor, `torso_link`. On the left they separate as the waist turns; on
the right the sidecar is refused, so the corrected config anchors where it
declares. The shipped index 16 resolves `torso_link` in both columns, which is the
control.

anchor body index

## Tests

`tests/policies/protomotions/test_body_index_addresses_a_body.py`, 50 cases.

| tree | failed | passed |
| --- | --- | --- |
| `upstream/main` | 36 | 14 |
| this branch | 0 | 50 |

All 36 pre-fix failures are in the regression class; both control classes are
green on `upstream/main` and here. The headline two fail with the wrong answer
rather than only "did not raise":

```text
Failed: ProtoMotionsConfig accepted anchor_body_index=-1 and resolved
'right_rubber_hand' as the anchor, where the config's own index 16 is 'torso_link'.
Failed: ProtoMotionsConfig accepted anchor_body_index=True and resolved 'head'
as the anchor.
```

Each guard was broken one at a time; every break fires a distinct set and none is
redundant:

| break | failed | fires |
| --- | --- | --- |
| revert the fix | 36 | every regression case |
| upper bound only (drop the shared domain) | 24 | both wrapping negatives, `True`, `2.7`, `"16"`, `None`, nan, inf, and the headline |
| range against `GTP_G1_BODY_NAMES` instead of the config's own list | 2 | trimmed embodiment, empty body list |
| loader coerces with `int()` first | 4 | the yaml route for `true` / `2.7` only |
| drop the `object.__setattr__` normalisation | 2 | the integral-float row-number cases |
| guard `anchor_body_index` only | 16 | every `root_body_index` case |
| control (as shipped here) | 0 of 50 | - |

## Verification

Measured at `25bcf10e` on `upstream/main` @ `1ab3af1d`.

* Identical 114-file selection on both trees, with the new test file excluded from
  each so it measures collateral only (every test naming the changed symbols, all
  of `tests/policies/protomotions`, and the repo-wide docstring / string /
  changelog / source guards): **3986 passed, 56 skipped, 0 failed on both**,
  219.4s vs 216.1s.
* `tests/policies/protomotions` alone: 159 passed.
* The shipped `unified_pipeline.yaml` from
  `cagataydev/protomotions-gtp-unitree-g1` loads unchanged - `anchor_body_index`
  16 -> `torso_link`, `root_body_index` 0 -> `pelvis`, 33 bodies, 29 joints - and
  both values are plain ints there, so the domain is a no-op on the shipped
  artifact.
* `ruff check` / `ruff format --check` clean over `strands_robots tests
  tests_integ`; `mypy` 24 errors on both this branch and a pristine

### DIFF
--- a/changelog.d/2619-body-index-addresses-a-body.md
+++ b/changelog.d/2619-body-index-addresses-a-body.md
@@ -1,0 +1,42 @@
+### Fixed: a tracker body index is refused unless it addresses a body
+
+`ProtoMotionsConfig.anchor_body_index` and `root_body_index` are offsets into
+`body_names` - the tracker reads a body out of the reference-motion cache by row
+index, never by name. `load_config_from_yaml` documents its result as "validated for
+consistent dimensions" and promises `ValueError` for "an inconsistent dimension", and
+the module docstring promises such an error "surfaces at policy build time with a
+clean message". It checked the `stiffness` and `damping` lengths against the joint
+count, and neither index.
+
+A negative index is a valid tuple lookup, so it was the silent case. `body_names[-1]`
+is `right_rubber_hand`, a real link: an `anchor_body_index: -1` sidecar loaded, and
+the whole chain then agreed on the hand - `required_bodies` declared it, the runtime
+resolved it once per rollout and merged `body.right_rubber_hand.quat` into every
+observation, `_extract_anchor_rot` found exactly that key, and the future-reference
+window sliced the same row. Measured on the tracker's own G1 embodiment, that link's
+world orientation is 0.01, 10.37 and 20.20 degrees from `torso_link` across a
+waist-turn pose set - an anchor orientation the network consumes every tick, with
+nothing anywhere reporting it. An index past the end instead deferred `IndexError:
+tuple index out of range` to whichever property read the name first, naming neither
+the field, the value, the range, nor the sidecar.
+
+The loader's `int()` coercion widened the same hole by running before anything looked
+at the value: a yaml `anchor_body_index: true` became row 1 (`head`, 34.4 degrees out)
+and a `2.7` became row 2 (`left_hip_pitch_link`), so the sidecar route and a
+hand-built `ProtoMotionsConfig(anchor_body_index=True)` disagreed about the same value.
+
+Both indices now go through the shared `non_negative_whole_number_error` rule and then
+the row count of the body list they index, in `ProtoMotionsConfig.__post_init__` so a
+sidecar and a hand-built config report the same value the same way. The domain runs
+before the `int()` normalisation, never after, which is the ordering
+`MotionBricksConfig.__post_init__` already states the reason for; the four sibling
+policy configs validate in `__post_init__` too, and `protomotions` was the one that did
+not. An integral float such as `16.0` still addresses a row and is kept, normalised to
+the row number both consumers index with, and the refusal quotes the range it was
+checked against. `anchor_body_name` and `root_body_name` documented
+`Raises: IndexError`, which is now unreachable, so both say so instead.
+
+This is the config-side mirror of a hole the package already refuses on the model side:
+a G1 MJCF missing bodies is rejected precisely because reading positionally hands the
+tracker `left_shoulder_pitch_link` where it asked for `torso_link`. The sidecar that
+named the same wrong row was not.
--- a/docs/policies/protomotions.md
+++ b/docs/policies/protomotions.md
@@ -244,6 +244,34 @@ sidecar; omit it to use the defaults, which match the shipped export.
 | `control_dt` | `0.02` | Seconds per control tick (50 Hz) |
 | `future_step_indices` | `(1, 2, 4, 8)` | Lookahead offsets, in control steps |
 
+### A body index has to address a body
+
+The two body indices are offsets into `body_names`, so an index that misses is
+the config-side mirror of the model-side row shift above - and it is refused the
+same way, when the config is built rather than when something reads it:
+
+```text
+ValueError: ProtoMotionsConfig.anchor_body_index must be a row of body_names
+(0..32), got 99 for 33 bodies. The index is an offset into body_names, so one
+that misses cannot resolve the body it names.
+```
+
+Both directions are checked, because they fail differently. A negative index is
+a valid tuple lookup: `body_names[-1]` is `right_rubber_hand`, so an
+`anchor_body_index: -1` sidecar resolves a real link and the tracker anchors on
+it consistently - `required_bodies` declares that link, the runtime supplies its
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes the config-side half of a hole the package already refuses on the model side: `ProtoMotionsConfig.anchor_body_index` / `root_body_index` are row offsets into `body_names`, and a negative index was a valid tuple lookup that silently anchored the tracker on a real but wrong link (`-1` -> `right_rubber_hand`, up to ~20 degrees of anchor-orientation error), while a positive out-of-range index deferred a bare `IndexError` to whichever property read the name first. The fix routes both indices through the shared `non_negative_whole_number_error` domain and a range check against the config's own `body_names` in `__post_init__`, so a yaml sidecar and a hand-built config report the same value the same way; the loader stops pre-coercing with `int()`, which used to launder `true` into row 1 and `2.7` into row 2 before the domain could see them. Integral floats are still accepted and normalised to a plain `int`, which both consumers (the tuple lookup and the numpy future-reference slice) index with.

## What's good

- Reuses the existing shared domain rather than minting a new single-caller guard (AGENTS.md rule 11), and the domain-before-`int()` ordering matches the four sibling policy configs.
- The now-refused inputs previously either misled silently or raised an explicit error; every new refusal is an explicit `ValueError` naming the field, value and range, so no older config is silently invalidated.
- Stale `Raises: IndexError` blocks on the two name properties are corrected in the same PR, keeping `tests/test_raises_docstring_completeness.py` honest.
- Verified at head `5b3ba55b`: `tests/policies/protomotions` (159 passed), the new 50-case suite, plus the repo-wide docstring/xref/test-naming/ASCII/changelog-fragment guard tests all green; `ruff check` and `ruff format --check` clean on the changed files.
- Test negatives assert what the wrong answer *was* (which body a wrapped index resolved), and the bound is checked against the config's own list rather than the G1 constant.
- Changelog fragment follows `changelog.d/<pr>-<slug>.md`; no direct `CHANGELOG.md` edit.

PR #2620, opened 2026-08-22
fix(mesh): a hardware stop that fails answers, instead of raising past the e-stop accounting

cagataycali:
## Problem

`Mesh._dispatch`'s `action="stop"` branch has three sub-paths that can raise. Two are wrapped, with the contract stated in the handler itself:

```python
except Exception as exc:  # noqa: BLE001 - stop must answer, not raise
    return {"ok": False, "error": f"stop_policy failed: {exc}"}
```

The hardware one, twenty lines above, was not:

```python
if hasattr(r, "stop_task"):
    return dict(r.stop_task())
```

An exception escaping it reaches `_exec_cmd`, which publishes `{"type": "error", "error": "dispatch error"}`. That envelope carries no `result`, so `_peers_that_did_not_stop` takes its bare-result fallback, reads the envelope itself, finds neither `ok` nor `status`, and `emergency_stop` counts the peer among its acknowledgements.

**The arm whose stop failed is reported as halted** - what `tests/mesh/test_estop_stop_honesty.py` calls "the worst available failure mode" on a safety path.

## Measured

Real envelopes from the real `_exec_cmd`, through the real `_on_response` (3 of 3 delivered), into the real `emergency_stop`:

| peer | stop outcome | wire reply | before | after |
| `arm-ok` | stopped | `type=response`, `status=success` | acknowledged | acknowledged |
| `arm-nostop` | no `stop_task` | `type=response`, `ok=False` | **flagged** | **flagged** |
| `arm-a1` | `stop_task` raised | `type=error`, `error="dispatch error"` | **acknowledged** | **flagged** |

```
before:  responses_received = 3   peers_not_stopped = ['arm-nostop']
after:   responses_received = 3   peers_not_stopped = ['arm-a1', 'arm-nostop']
```

Two of three peers did not stop; the safety envelope named one.

**Raising is the only way a hardware stop can report failure.** `Robot.stop_task` has exactly two `return` statements and both answer `{"status": "success", ...}` - so the shape the classifier's docstring named for a failed hardware stop was unreachable, and the shape that actually occurs was unread. The `{"status": "error", ...}` the docstring names is real, but it comes from the sim `stop_policy` path.

## Change

`strands_robots/mesh/core.py`, two edits:

1. The hardware `stop_task` call answers `{"ok": False, "error": f"stop_task failed: {exc}"}` and logs at ERROR with the `[safety]` prefix, mirroring the sim branches and the `no stop_task` fallback beside it. The classifier's existing `ok is False` rung then reads it - no change to `_peers_that_did_not_stop` itself.
2. The classifier's docstring census now names all three reachable shapes instead of two, one of which was unreachable.

The reason reaches the wire (`stop_task failed: <exc>`) because the `stop` action's three other failure answers already do - `stop_policy failed: {exc}` twice and `peer exposes no stop_task; nothing was stopped` - and on a safety path the requesting operator needs to know why. The **generic** dispatch-error path keeps its sanitised `"dispatch error"`, pinned by a test here. If you would rather this one were generic too it is a one-word change and the accounting fix is unaffected.

## Scope

Deliberately **not** in this PR, with the measurement:

- **Widening the classifier to read `type == "error"`** would also flag a validation rejection, a non-dict envelope and a lockout refusal. `_dispatch` raises a generic `LockoutError("command rejected")` precisely so a remote caller cannot map the lockout window, so a consumer cannot single it out - and a peer already in lockout was stopped by the e-stop that engaged it. Reading every error envelope as "did not stop" would report an already-halted peer as possibly moving, the false alarm `_peers_that_did_not_stop` is written to avoid. That trade is a contract decision; `test_an_error_envelope_is_still_not_flagged_by_the_classifier` pins the boundary and fails if it is widened without one.
- **A top-level `ok` / `code` on the reply envelope.** Additive wire change, and the fix above makes the only unambiguous case unreachable for `stop`.

Two measured findings from the same surface, reported rather than fixed here because each is its own contract:

- `validate_command` does not carry `robot_name`, so a wire `stop` to a sim peer can never be scoped to one robot - the `if robot_name:` sub-path is reachable only by a direct call. The test that covers that shape says so and drives `_dispatch` directly.
- The multi-robot sim stop answers `{"ok": True, "stopped": [...], "results": {name: {...}}}` even when an inner per-robot `stop_policy` returned `{"status": "error"}`, so a refusal nested one level deeper reads as an acknowledgement.

## Tests

`tests/mesh/test_stop_answers_instead_of_raising.py`, 13 tests. Pre-fix on a pristine `upstream/main`: **6 failed / 7 passed**; all six are behavioural, e.g.

```
Failed: a hardware stop_task that raised produced {"type": "error", "responder_id":
"arm-fault", "turn_id": "t-arm-fault", "error": "dispatch error", "timestamp": ...},
which the e-stop accounting reads as an acknowledgement -- the arm is reported as halted

AssertionError: two of three peers did not stop; the safety envelope named ['arm-nostop']

AssertionError: these stop calls can raise out of _dispatch, so _exec_cmd answers with
a type='error' envelope the e-stop accounting cannot read: ['r.stop_task()']
```

The structural guard derives the stop calls from `_dispatch`'s own AST, so a fourth raising sub-path is graded on arrival rather than added to a list.

Each guard is load-bearing:

| regression | tests that fail |
| --- | --- |
| revert the source | headline, end-to-end, reason, log, structural, census (6) |
| answer a shape the classifier does not read (`ok` dropped) | headline, end-to-end, reason (3) |
| revert only the docstring census | census (1) |
| drop the local ERROR log | log (1) |
| AST scan reaches nothing | structural + non-vacuity (2) |
| widen the classifier to every `type="error"` | scope-boundary control (1) |
| control | 0 of 13 |

## Verification

Blast radius: **294 files** (every test naming the changed surface, all of `tests/mesh`, and the repo-wide guards), the identical selection on this branch and on a pristine `upstream/main`, with the new file excluded from both.

| tree | result |
| --- | --- |
| this branch | `23 failed, 9697 passed, 22 skipped, 24 errors` in 357.9s |
| pristine `upstream/main` | `23 failed, 9697 passed, 22 skipped, 24 errors` in 361.2s |

Byte-identical summaries; 47 failing ids, `comm` empty in **both** directions. All 47 are dependency-absent on the machine this was measured on and are declared extras CI installs: 45 under `tests/mesh/test_iot_*` (`No module named 'awsiot'`, 88 occurrences; `awsiotsdk` is declared) and 2 in `tests/test_device_connect_hardening.py` (`device_connect_agent_tools`, declared).

`ruff check` / `ruff format --check` clean over the CI scope. `mypy strands_robots tests tests_integ` **24 == 24** against a pristine worktree, identical sets, branch-only empty, none in the changed files.

Those numbers were measured at `83ba49d1`. The head has since been amended to `80ce776e`, whose only delta is inside the new test file's AST helper: `pytest.fail(...)` became `raise AssertionError(...)`, because `fail()` is only terminal at runtime and a helper that also returns a value therefore mixes an explicit return with an implicit one. The pre-fix split is unchanged at 6 failed / 7 passed and the file still passes 13 of 13.

## Artifact

An SO-101 in MuJoCo, with a peer whose `stop_task` raises a serial fault. The rollout is one capture shown in both columns; the arm is still sweeping in every frame after the e-stop fires (9 of 9, per-frame mean |px| up to 39.3). What changes is the reply the failed stop puts on the wire, and therefore what the operator is told: before, one CRITICAL line and `peers_not_stopped = []`; after, the `did NOT stop ... use a hardware cutoff` warning. The peer that did stop replies identically in both columns.

e-stop accounting for a failed hardware stop


yinsong1986 (approved):
## Summary

Wraps the one bare stop sub-path in `Mesh._dispatch`'s `action="stop"` branch -- the hardware `r.stop_task()` call -- in the same answer-not-raise contract its two sim siblings already carry, so a hardware stop that raises now answers `{"ok": False, "error": "stop_task failed: ..."}` instead of escaping to `_exec_cmd`'s generic `{"type": "error"}` envelope, which `_peers_that_did_not_stop` cannot read and `emergency_stop` counted as an acknowledgement. The classifier itself is unchanged (its existing `ok is False` rung reads the new shape), the docstring census is corrected to name the three reachable shapes, and the deliberately-narrow classifier boundary (a bare `type="error"` envelope is still not flagged, so a lockout refusal cannot masquerade as a moving robot) is pinned by a test. Verified end-to-end against the head commit: the new suite (13 tests) and the existing `test_estop_stop_honesty.py` (13 tests) all pass, and the fix's mechanism checks out against `_exec_cmd`'s response nesting (`result` key) and the classifier's fallback read.

## What's good

- The exception-reason-on-the-wire question is anticipated and answered in-tree: the stop branch's two pre-existing siblings already answer `stop_policy failed: {exc}`, and the generic dispatch path's sanitised `"dispatch error"` contract is pinned by `test_a_non_stop_action_still_sanitises_its_dispatch_error` so this fix cannot be misread as licensing detail leaks elsewhere.
- The structural AST guard derives the graded stop-call set from `_dispatch`'s own source with a non-vacuity check, so a fourth raising sub-path is caught on arrival rather than appended to a list.
- Scope discipline: the two adjacent findings (unscopeable wire `robot_name`, nested multi-robot sim refusal reading as an acknowledgement) are measured and reported rather than bundled in.
- Changelog fragment present; log strings ASCII; `[safety]` ERROR log mirrors the `no stop_task` sibling.

yinsong1986 (approved):
## Summary

Wraps the one bare stop sub-path in `Mesh._dispatch`'s `action="stop"` branch (the hardware `r.stop_task()` call) in the same answer-not-raise contract its two sim siblings already carry, so a raising hardware stop now answers `{"ok": False, "error": "stop_task failed: ..."}` -- a shape the existing `ok is False` rung of `_peers_that_did_not_stop` already reads -- instead of escaping to `_exec_cmd`'s bare `{"type": "error"}` envelope, which the e-stop accounting counted as an acknowledgement. The fix is minimal (one try/except plus a docstring census correction; the classifier itself is untouched), the scope boundaries are deliberate and pinned (the classifier is not widened to `type="error"`, so a lockout refusal on an already-halted peer cannot fire a false "did not stop"; the generic dispatch-error path keeps its sanitised wording), and the exception reason reaching the wire matches the three sibling failure answers on the same action. The prior CodeQL finding on the test helper's mixed returns was resolved in the head amendment (`pytest.fail` replaced with `raise AssertionError`).

## What's good

- Changelog fragment present per AGENTS.md step 3; test names describe behaviour, not provenance; no host paths, no non-ASCII in user-facing strings.
- The `# noqa: BLE001 - stop must answer, not raise` clause mirrors the established sibling contract on a recovery path, and the failure is loud locally (ERROR + `[safety]` prefix + `exc_info`), consistent with the `no stop_task` fallback beside it.
- The AST-derived structural guard grades a future fourth raising stop sub-path on arrival instead of maintaining a list, with a non-vacuity check so an empty scan cannot pass.
- Regression tests drive the real `_exec_cmd` receive path and the real `emergency_stop` aggregation, with a pre-fix split (6 failed / 7 passed) demonstrating each guard is load-bearing.

PR #2621, opened 2026-08-22
fix(mesh): a degraded state probe names itself in the published snapshot

cagataycali:
## Problem

Every section of a state snapshot is optional, because a robot may be hardware, sim, both or neither. So an absent section is ambiguous on its own: a robot with no joints and a robot whose joint read just raised publish the same thing.

A previous change made a failing probe report itself in the peer's own log, once per category. That closed half of the problem and could not close the other half, because the observer that has to explain the absence is on another machine. Its own test file said so, and scoped the wire out explicitly:

> what is *published* is byte-for-byte what it was before -- this changes the report, not the snapshot.

With nothing on the wire, a fleet view could only tell the two cases apart by reading that peer's log -- so one grew a regex over `mesh.core`'s log lines and used it as an API.

## Measured

A hardware peer whose only section is `joints`, driven through the real `_read_state` with four faults an operator answers differently:

| joint read raises | publishes | `joints` | reason on the wire | log lines |
| (nothing -- control) | yes | yes | n/a | 0 |
| `ConnectionError("Port is in use!")` | **no** | no | none | 1 |
| `RuntimeError("no calibration registered")` | **no** | no | none | 1 |
| `TimeoutError("no response from motors")` | **no** | no | none | 1 |

The peer does not merely omit a section -- it publishes **nothing at all**. `_read_state` returns `None` when only `peer_id` and `t` survived and the state loop publishes nothing for a `None`, so the peer goes silent on the state topic for as long as its bus is contended, while its presence heartbeat keeps advertising it. `_presence_loop` does not read `_read_state`, so the two disagree.

That is why a log regex was the only way to explain it: there was no message to inspect. A peer that also runs a task keeps publishing, and its snapshot then carries `{peer_id, t, task}` with `joints` silently absent and no reason -- indistinguishable from an arm with no joints.

## Change

A probe that fails names itself in `degraded`, keyed by category:

```json
"degraded": {
  "hw_joints": {
    "reason": "ConnectionError",
    "detail": "Port is in use!",
    "failures": 37,
    "for_seconds": 3.7
  }
}
```

`reason` is the exception's **type name**, which is not an invented vocabulary: `_warn_read_state_once`'s own docstring already names it as the discriminator that selects the operator's next move ("a `ConnectionError` from a contended serial port and a `RuntimeError` from an uncalibrated arm need different actions"). The same value the log carries is now the value the wire carries. `detail` is that exception's message, bounded by a new `MAX_DEGRADED_DETAIL_LEN` because the text comes from a driver and the topic publishes ten times a second -- the same reason `MAX_INSTRUCTION_LEN` and `MAX_MODEL_PATH_LEN` exist beside it.

The entry is removed on the tick the probe answers again, so the block describes the current state rather than the worst thing that ever happened. The clear happens where the read returned, not where the probe was skipped, so a sim peer with no motor bus cannot read as an `hw_joints` recovery and an arm that has gone away keeps its last diagnosis.

`for_seconds` is a duration, so it is measured on `time.monotonic()`; the `since_mono` stamp it is derived from stays off the wire, because seconds of local process uptime mean nothing to the machine reading the snapshot. `t` is an absolute stamp another machine correlates and stays on `time.time()`.

And because a diagnosis is something to report, the empty-snapshot guard stops suppressing it: the silenced peer above publishes its fault instead of vanishing.

**The log is deliberately unchanged.** `_read_state_warned` arms once per category for the life of the peer, so a probe flapping at `STATE_HZ` costs the same one warning it always did while the wire tracks every transition. Clearing that gate on recovery would have re-armed the warning and traded ten silent ticks for a warning per tick -- exactly the noise the once-per-category gate exists to prevent. The two structures answer different questions and are kept separate for that reason.

## Tests

Pre-fix, against `upstream/main`: **22 failed, 10 passed.** Twenty-one fail by measuring the pre-fix behaviour or source -- eleven with `assert None is not None` (the peer went silent), two with a `KeyError` on the absent block, and the rest on the structural pins -- and one on the new constant's absence.

Five existing assertions are inverted rather than repaired, because each pinned the behaviour this changes. Four are in the probe-failure file and asserted `out is None` with the message "a snapshot with no readable section is still not published"; every warning assertion in those tests survives untouched, so what moved is the scope boundary the earlier change drew, not the contract it pinned. That file's module docstring said the snapshot was unchanged, and it no longer is, so it says what it says now.

The fifth is `test_mesh.py::TestTelemetryRobustness::test_read_state_swallows_every_attribute_error`, and only the wider run found it -- the two probe files alone were green. Its contract is that nothing propagates from a hostile driver, which is unchanged and still asserted. What it yields is no longer nothing: a robot whose every probed attribute raises now names all three of `hw_joints`, `task_state` and `sim_world`, each with `reason` `RuntimeError`, so a peer whose every read fails is distinguishable from one with nothing to report. The mutating call also moves out of its `assert`, which leaves that file with one such site where it had two.

Every guard was verified load-bearing by breaking the fix one way at a time:

| break | failing | fires |
| --- | --- | --- |
| revert the source | 22 | the full pre-fix split |
| never clear the record | 2 | both recovery tests |
| `for_seconds` on the wall clock | 1 | the monotonic-clock test |
| keep the first reason forever | 1 | the changed-failure-mode test |
| unbounded `detail` | 1 | the bound test |
| recovery re-arms the log gate | 1 | the flapping test |
| attach the block even when empty | 4 | three healthy-peer tests plus a pre-existing one |
| clear where the probe was skipped | 1 | the skipped-probe test |
| leak `since_mono` onto the wire | 1 | the off-the-wire test |
| one category reported but never cleared | 1 | the structural recovery-parity test |
| attach the block inside a probe's `try` | 21 | the attach-unconditionally pin plus twenty more |
| control | 0 of 32 | |

The last two are derived from `_read_state`'s source rather than listed, so a fifth probe is held to both halves of the contract instead of being able to report a failure it never clears.

## Artifact

degraded state probes on the wire

Twelve ticks of one hardware peer, driven through the real `_read_state` and the state loop's own `if state:` gate on each tree in turn. The bus is contended for six of them, then the port is freed. Same script, same package code; only `strands_robots` differs between the columns.

`upstream/main` publishes **6 of 12** ticks, with **six** fault ticks carrying no message at all. This branch publishes **12 of 12**, with those six carrying the reason. Both emit exactly **one** log warning, which is the no-change control for the report, and a healthy tick's snapshot is identical in both columns -- `{peer_id, t, joints}`, no key added. Every number in the figure is asserted from the captured JSON by the script that draws it.

## Scope

`reason` is the exception type, not a code vocabulary. Mapping types onto codes such as `PORT_IN_USE` or `NO_RESPONSE` would mean matching a driver's message **text** -- `"Port is in use!"` is a lerobot string -- and this module has no basis for that classification. No such vocabulary exists in the tree today.

A recovery counter is out of scope: `bus_access` does not keep one, so there is nothing to surface. `failures` and `for_seconds` describe the current fault, and a fault that clears and returns starts a fresh count, which is pinned.

The `except Exception` breadth on each probe stays. A flaky read must not kill the state thread; the defect was the silence, not the breadth.

## Verification

Measured at `7d3c3780` on `upstream/main` `a12120ca`.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors, the same 24 as a pristine worktree of the base, none in a changed file.

Blast radius: 390 test files -- everything naming the probe machinery, all of `tests/mesh`, and the repo-wide docs, docstring, string and changelog guards -- run with the identical selection on this branch and on a pristine `upstream/main` worktree, with the three test files this PR changes excluded from both.

| this branch | 24 | 13527 | 86 | 24 |
| pristine base | 23 | 13528 | 86 | 24 |

Failing ids: 48 against 47. The single difference was `test_mesh.py`'s hostile-robot test, which that run measured before it was repaired; excluding that file the sets are **47 and 47 with an empty diff in both directions**, and the file itself is 46 passed on each tree afterwards. The 47 shared are local environment gaps, not code: 45 in `tests/mesh/test_iot_*` on `No module named 'awsiot'` (a declared extra CI installs) and two in `tests/test_doctor.py` on `PackageNotFoundError`, because the worktree is not pip-installed.

### DIFF
--- a/strands_robots/mesh/core.py
+++ b/strands_robots/mesh/core.py
@@ -190,6 +190,13 @@ RESUME_FORWARD_SKEW_S: float = _parse_positive_float_env("STRANDS_MESH_RESUME_FO
 #: See :data:`RESUME_FRESHNESS_WINDOW_S` for lazy-resolution note.
 RESUME_REPLAY_CACHE_MAX: int = _parse_positive_int_env("STRANDS_MESH_RESUME_REPLAY_CACHE_MAX", "4096")
 
+#: Longest ``detail`` string a degraded-probe record carries onto the state
+#: topic. The text is a third-party driver's exception message, the state topic
+#: publishes at ``STATE_HZ``, and a driver that puts a whole device dump in that
+#: message would otherwise put ten copies of it on the wire every second. The
+#: ``reason`` beside it is the exception's type name, which is bounded already.
+MAX_DEGRADED_DETAIL_LEN: int = 256
+
 
 def _resume_freshness_window_s() -> float:
     """Lazy resolver for ``STRANDS_MESH_RESUME_FRESHNESS_S``.
@@ -412,6 +419,14 @@ class Mesh(SensorLoopsMixin):
         # The state loop retries at STATE_HZ, so a persistent fault is reported
         # once and then kept at debug level rather than once per tick.
         self._read_state_warned: set[str] = set()
+
+        # Which _read_state probes are degraded RIGHT NOW, keyed by category.
+        # Separate from _read_state_warned, which answers a different question:
+        # this one is cleared when a probe answers again, so the published
+        # snapshot stops claiming a fault that has cleared, while the log gate
+        # above stays armed for the life of the peer so a probe that flaps at
+        # STATE_HZ cannot re-arm the warning ten times a second.
+        self._read_state_degraded: dict[str, dict[str, Any]] = {}
 
         # RPC correlation state.
         #
@@ -1132,7 +1147,8 @@ class Mesh(SensorLoopsMixin):
                 survives even when the message is empty, because the type is
                 what selects the operator's next move: a ``ConnectionError``
                 from a contended serial port and a ``RuntimeError`` from an
-                uncalibrated arm need different actions.
+                uncalibrated arm need different actions. That same type name is
+                what the published record carries as its ``reason``.
 
         Each probe stays wrapped in ``except Exception`` on purpose -- a flaky
         read must not kill the state thread -- so this exists to keep that
@@ -1140,12 +1156,41 @@ class Mesh(SensorLoopsMixin):
         is warned about; the rest drop to debug, because the loop retries at
         ``STATE_HZ`` and a persistent fault would otherwise emit ten warnings a
         second.
+
+        The log gate and the published record are deliberately separate. The
+        gate arms once and stays armed for the life of the peer, so a probe that
+        flaps cannot emit a warning per tick; the record is cleared by
+        :meth:`_note_read_state_ok` as soon as the probe answers, so the
+        snapshot stops claiming a fault the moment it clears. Every call updates
+        the record's failure count, reason and detail, because a probe whose
+        failure mode CHANGES -- a contended port that becomes an uncalibrated
+        arm -- would otherwise keep publishing the first reason forever.
         """
-        warned = getattr(self, "_read_state_warned", None)
-        if warned is None:
+        records = getattr(self, "_read_state_degraded", None)
+        if records is None:
             # A Mesh built through __new__ (several tests, and the sim paths)
             # never ran __init__. Bookkeeping must not be able to raise inside
             # the handler that exists to report a failure.
+            records = {}
+            self._read_state_degraded = records
+        record = records.get(category)
+        if record is None:
+            records[category] = {
+                "reason": type(exc).__name__,
+                "detail": str(exc)[:MAX_DEGRADED_DETAIL_LEN],
+                "failures": 1,
+                # Monotonic: for_seconds answers "how long has this been
+                # failing", a duration, and a wall-clock stamp would make that
+                # duration jump by the size of any NTP correction mid-fault.
+                "since_mono": time.monotonic(),
+            }
+        else:
+            record["failures"] = int(record.get("failures", 0)) + 1
+            record["reason"] = type(exc).__name__
+            record["detail"] = str(exc)[:MAX_DEGRADED_DETAIL_LEN]
+
+        warned = getattr(self, "_read_state_warned", None)
+        if warned is None:
             warned = set()
             self._read_state_warned = warned
         if category in warned:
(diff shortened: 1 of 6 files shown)


strands_robots/mesh/core.py
```diff
@@ -1242,11 +1377,20 @@ def _read_state(self) -> dict[str, Any] | None:
...
+                self._note_read_state_ok("sim_world")
         except Exception as exc:
             self._warn_read_state_once("sim_world", exc)
 
+        # Unconditional: a probe that is failing says so on the wire for as long
+        # as it fails, which is also what keeps a hardware-only peer publishing
+        # at all when its one section is the one that raised.
+        degraded = self._degraded_probes()
+        if degraded is not None:
+            snapshot["degraded"] = degraded
```
yinsong1986, line 1392: [MUST FIX] This unconditional attach changes `_read_state`'s contract for a robot whose every probe raises -- it now returns `{peer_id, t, degraded: {...}}` instead of `None` -- but one pre-existing pin of the old contract was not updated: `tests/mesh/test_mesh.py::TestTelemetryRobustness::test_read_state_swallows_every_attribute_error` asserts `m._read_state() is None` for `_HostileRobot`, and now receives a degraded-only snapshot. This is the exact test failing the required `call-test-lint / Test and Lint` check at head `6fa17321` (CI short summary: `1 failed, 2664 passed`, this test alone). Verified locally: it passes on `upstream/main` and fails on this branch, and it is the only PR-introduced failure in the mesh suite.

This blocks merge mechanically: `call-test-lint / Test and Lint` is the one context in the ruleset's required set (AGENTS.md > PR Workflow step 4, "All tests must pass"), so the PR cannot go `CLEAN` until the suite agrees with itself. The PR inverted the four `out is None` assertions in `test_read_state_probe_failures_are_reported.py` but missed this fifth one in `test_mesh.py`.

Resolution: update that test to assert the new contract -- a hostile robot yields a degraded-only snapshot naming all three probes, and the "must not raise" half of its docstring survives untouched -- mirroring the inversions already made in `test_read_state_probe_failures_are_reported.py`.

yinsong1986 (commented):
## Summary

This PR puts a failing `_read_state` probe on the wire: a `degraded` block keyed by category (`hw_joints`, `task_state`, `sim_world`, `sim_joints`) carrying the exception's type name as `reason`, its message bounded by the new `MAX_DEGRADED_DETAIL_LEN`, a `failures` count and a monotonic-derived `for_seconds` -- and, because a diagnosis is content, a hardware-only peer whose one section raised now publishes its fault instead of going silent while its presence heartbeat keeps advertising it. The record clears where the read returned (not where the probe was skipped), the once-per-category log gate is deliberately unchanged, and a healthy peer's snapshot is byte-for-byte what it was. One blocker, flagged inline: a fifth pre-existing pin of the old `out is None` contract in `tests/mesh/test_mesh.py` was not updated alongside the four in `test_read_state_probe_failures_are_reported.py`, and it is the exact test failing the required `call-test-lint / Test and Lint` check at this head.

## What's good

- Clock discipline follows AGENTS.md exactly: `for_seconds` is a duration on `time.monotonic()`, the `since_mono` stamp stays off the wire, `t` stays on `time.time()` -- and both halves are pinned.
- The wire addition is additive and optional (no key when healthy), `detail` is bounded for the same reason `MAX_INSTRUCTION_LEN` exists, and the shape is documented in `docs/mesh.md` with a changelog fragment.
- The two structural tests derive the category set and the unconditional attach from `_read_state`'s AST, so a fifth probe is held to both halves of the contract.
- All added strings are ASCII; the clear-on-answer vs. arm-for-life split between `_read_state_degraded` and `_read_state_warned` is well reasoned and pinned against flapping.

strands_robots/mesh/core.py
cagataycali: Fixed at `7d3c3780` (`test(mesh): the hostile-robot probe test names the failures it now reports`), which is the current head.

That commit updates the fifth pin exactly as you describe: `test_read_state_swallows_every_attribute_error` now asserts the new contract -- `out is not None`, no telemetry survived (`"joints" not in out and "task" not in out`), `set(out["degraded"]) == {"hw_joints", "task_state", "sim_world"}` and every reason is `RuntimeError` -- while the "must not raise" half survives untouched as the `m._read_state()  # must not raise` call plus the docstring sentence naming it. The inversion mirrors the four already made in `test_read_state_probe_failures_are_reported.py`.

Verified at head on this box: the named test plus both probe-failure suites are `33 passed`.

```
tests/mesh/test_mesh.py::TestTelemetryRobustness::test_read_state_swallows_every_attribute_error .
tests/mesh/test_read_state_probe_failures_are_reported.py ............
tests/mesh/test_state_degraded_probes_are_published.py ....................
33 passed in 1.98s
```

Thanks for catching it -- the grep that would have found it is the contract phrase rather than the symbol (`_read_state() is None` across `tests/mesh/`), not the four assertions I did invert.

yinsong1986 (approved):
## Summary

This PR closes the observer-side half of the degraded-probe problem: a `_read_state` probe that raises now names itself in a `degraded` block on the state topic, keyed by category (`hw_joints`, `task_state`, `sim_world`, `sim_joints`), carrying the exception's type name as `reason` (the same discriminator the existing log warning documents), its message bounded by the new `MAX_DEGRADED_DETAIL_LEN`, a per-fault `failures` count and a monotonic-derived `for_seconds`. Because a diagnosis is content, the empty-snapshot guard no longer silences a hardware-only peer whose one section is the one that raised - previously it published nothing at all while its presence heartbeat kept advertising it. The entry clears only where a read actually returned, so a skipped probe is not a recovery; the once-per-category log gate is deliberately unchanged, so a flapping probe still costs one warning. The wire-schema addition is additive and absent when healthy (a healthy snapshot is byte-for-byte unchanged, pinned by test), the `for_seconds`/`since_mono` split follows the clock rule in AGENTS.md exactly, and the `degraded`-block attach and clear are both derived structurally from `_read_state`'s source so a fifth probe is held to the full contract. The one prior [MUST FIX] (the stale `is None` pin in `test_mesh.py`) is resolved at this head. Ran the three touched test files at `7d3c3780`: 78 passed; `ruff check` and `ruff format --check` clean on the changed files.

## What's good

- Changelog fragment and `docs/mesh.md` updated in the same PR, with the categories and the bounded-`detail` rationale documented where consumers will read them.
- The break-the-fix table in the description is verifiable and the structural AST pins (attach-unconditionally, report-implies-clear) make the contract hold for probes added later.
- Scope discipline: no code vocabulary invented for `reason`, no recovery counter without a source for one, and the `except Exception` breadth on each probe correctly left alone.

PR #2622, opened 2026-08-22
fix(mesh): broadcast reports the cmd-topic size cap it shares with send

cagataycali:
## Problem

The transport's `low_pass_filter` caps two key expressions with **one** rule:

```json
{"id": "strands_cmd_size_cap", "key_exprs": ["**/cmd", "**/broadcast"],
 "flows": ["ingress", "egress"], "size_limit": cmd_bytes}
```

`DEFAULT_MAX_CMD_BYTES` is 16 KiB. The command validator admits a `world_update` up to
`MAX_WORLD_UPDATE_BYTES`, which is **64 KiB** -  so a validator-valid command can be four
times the size the transport will carry, and the filter drops it with no diagnostics.

`Mesh.send` already pre-checks that gap and answers with a structured error naming the cap.
`Mesh.broadcast` publishes on the *other* key expression of the *same* rule and did not, so an
over-cap broadcast was handed to the filter and returned `[]` -  which is exactly what a
broadcast nobody answered returns. `broadcast` is a fleet-wide action the `robot_mesh` tool
exposes and one of `_DEFAULT_INTERRUPT_ACTIONS`.

Measured on `a12120ca`, driving both publishers with the same validator-valid command
(`{"action": "execute", "instruction": "go", "policy_provider": "mock", "world_update": {"blob": "x"*20000}}`,
20257 bytes encoded, cap 16384):

| publisher | result | reached the transport |
| --- | --- | --- |
| `send` | `{"status": "error", "error": "command message is 20257 bytes; the transport drops cmd messages over 16384 bytes (STRANDS_MESH_MAX_CMD_BYTES). ..."}` | no |
| `broadcast` | `[]` | **yes -  `('strands/broadcast', 20257)`** |
| `broadcast`, in-cap (control) | `[]` | yes -  `('strands/broadcast', 130)` |

Rows 2 and 3 are byte-identical to the caller. That is the "the sender only ever sees a
timeout" failure mode, one method over from where it was already fixed.

## Change

`Mesh._cmd_topic_size_problem(msg)` is now the one owner of the cmd-topic cap, and both
publishers on that filter rule ask it. A second inline check is how two key expressions of one
rule come to disagree about the cap they share, so `send`'s inline block moved into the helper
rather than being copied.

`broadcast` reports the reason the way it **already** reports a client-side rejection -  its
return type is `list[dict]`, so a rejection has no structured slot, and the existing
`ValidationError` path logs and returns `[]`. The size rejection reuses that shape and wording
(`"broadcast rejected client-side: ..."`), so the method has one rejection shape rather than two.

`send`'s error text is unchanged, byte for byte -  pinned by a control test. The helper keeps
`send`'s documented `ImportError` behaviour: if the config helper cannot be imported there is no
cap to compare against, so the payload is published as it was before the check existed rather
than every command failing on a missing optional module.

## Scope

Only the cmd-topic rule. The filter has two other rules with **different** caps, and neither is
the same contract:

- `**/camera/**` is capped at 1 MiB **ingress only** -  the comment says "publisher trusts own
  frames", so a publisher-side check is not what that rule asks for.
- `**/safety/**` is capped at 4 KiB ingress and egress. Its publishers have no pre-check either,
  but a safety envelope cannot simply be refused (the e-stop must proceed), so what to do instead
 -  truncate, split, or report locally -  is a design decision rather than a mirror of an existing
  one. Reported separately with the measurement rather than guessed at here.

## Tests

`tests/mesh/test_cmd_topic_size_cap_is_reported.py`, 19 cases. Against `a12120ca`:
**8 failed / 11 passed** -> 19 passed. Three of the eight are behavioural and all name
`broadcast`:

```
FAILED ...::test_an_over_cap_command_does_not_reach_the_transport[broadcast]
  AssertionError: broadcast() handed a 20099-byte command to the transport, whose cmd-topic
  rule caps it at 16384 bytes and drops it silently: [('strands/broadcast', 20257)]
FAILED ...::test_the_reason_names_the_actual_size_the_cap_and_the_env_var[broadcast]
  AssertionError: broadcast() did not name '16384': []
FAILED ...::test_the_over_cap_broadcast_is_logged_as_a_client_side_rejection
```

Three more are the single-owner pins and two report the helper's absence; those pin the
contract's shape rather than the regression, and I have labelled them that way.

The eleven that pass on both trees are the controls -  including the whole `[send]` half of the
parametrized headline, which is what makes "`send` is unchanged" a measurement.

Each guard is load-bearing (6 breaks, 6 distinct firing sets, zero dead guards):

| break | tests that fire | n |
| --- | --- | --- |
| revert the whole change | all of them | 8 |
| revert `broadcast`'s check only | the 3 behavioural + `broadcast`'s consult pin | 4 |
| `broadcast` checks but does not log | reason[broadcast] + the log-shape test | 2 |
| the reason omits the env var | reason[send] + reason[broadcast] + the helper pin | 3 |
| `send` re-derives the cap inline | reason[send] + the wording control + both `send` pins | 4 |
| the helper refuses a payload that fits | both in-cap controls | 2 |
| unmodified | -  | 0 of 19 |

## Artifact

The same validator-valid command driven through both publishers on each tree. Row 2 is the
change; rows 3 and 4 are in-cap controls, and row 1 is `send`, whose result is byte-identical
across the two columns.

over-cap broadcast reached the filter that drops it

No robot behaves differently either way: on `main` the over-cap broadcast is eaten by the filter
and on this branch it is refused before the wire, so in both cases no peer receives it. What
changes is whether the caller is told, which is the whole point of the pre-check `send` already
had.

## Verification

Measured at `a43d33ac`, against `upstream/main` `a12120ca`.

- `tests/mesh/test_cmd_topic_size_cap_is_reported.py`: 19 passed (8 failed / 11 passed on `a12120ca`).
- Blast radius: the 207 test files naming the changed symbols, plus all of `tests/mesh` and the
  repo-wide docstring / string / changelog guards, run as the identical selection on both trees
  with the new file excluded from each. Both report `23 failed, 5463 passed, 5 skipped, 24 errors`
  as byte-identical summary lines, and the 47 failing ids are the same set in both directions
  (`comm` empty each way). All 47 need optional extras this environment lacks: 45 want `awsiot`
  and 2 want `device_connect_agent_tools`, both of which CI installs.
- `mypy strands_robots tests tests_integ`: 24 errors on the branch and 24 on a pristine
  `upstream/main` worktree, identical sets, none in the changed files.

### DIFF
--- a/changelog.d/2622-broadcast-cmd-size-cap.md
+++ b/changelog.d/2622-broadcast-cmd-size-cap.md
@@ -1,0 +1,14 @@
+### Fixed: broadcast reports the cmd-topic size cap it shares with send
+
+The transport's `low_pass_filter` caps `**/cmd` and `**/broadcast` with one rule
+(`strands_cmd_size_cap`, 16 KiB, ingress and egress) while the command validator admits a
+`world_update` up to `MAX_WORLD_UPDATE_BYTES` (64 KiB). `Mesh.send` pre-checked that gap and
+answered with a structured error naming the cap and the env var that raises it.
+`Mesh.broadcast` publishes on the other key expression of the same rule and did not, so a
+validator-valid command four times the wire limit was handed to the filter and returned the
+empty list a broadcast nobody answered also returns.
+
+`Mesh._cmd_topic_size_problem` is now the one owner of that cap and both publishers ask it, so
+the two key expressions of one filter rule cannot come to disagree about the limit they share.
+`broadcast` reports the reason the way it already reports a client-side rejection: log it and
+return no responses. `send`'s error wording is unchanged.
--- a/strands_robots/mesh/core.py
+++ b/strands_robots/mesh/core.py
@@ -2972,6 +2972,43 @@ class Mesh(SensorLoopsMixin):
                 )
 
     # RPC -- outgoing
+    def _cmd_topic_size_problem(self, msg: dict[str, Any]) -> str | None:
+        """Report why *msg* would be dropped by the cmd-topic byte cap.
+
+        The transport's ``low_pass_filter`` caps ``**/cmd`` AND
+        ``**/broadcast`` with one rule (``strands_cmd_size_cap``, default
+        16 KiB, ingress and egress) while the command validator admits a
+        ``world_update`` up to :data:`~strands_robots.mesh.security.MAX_WORLD_UPDATE_BYTES`
+        (64 KiB). A valid command in that gap is silently dropped, so both
+        publishers on that rule ask this before handing the payload over.
+
+        One owner rather than a copy per publisher: a second inline check is
+        how the two topics come to disagree about the cap they share.
+
+        Returns:
+            A reason naming the encoded size, the cap and the env var that
+            raises it, or ``None`` when the message fits -- and also ``None``
+            when the cap cannot be read at all. The check is a diagnostic, not
+            part of publishing: it turns a silent transport drop into a
+            reported one. If the config helper cannot be imported there is no
+            cap to compare against, so the caller publishes and an over-cap
+            message behaves as it did before this check existed. Failing every
+            command because an optional module is missing would be worse.
+        """
+        try:
+            from strands_robots.mesh._zenoh_config import cmd_bytes_cap
+        except ImportError:
+            return None
+        encoded_len = len(json.dumps(msg).encode("utf-8"))
+        cap = cmd_bytes_cap()
+        if encoded_len <= cap:
+            return None
+        return (
+            f"command message is {encoded_len} bytes; the transport drops "
+            f"cmd messages over {cap} bytes (STRANDS_MESH_MAX_CMD_BYTES). "
+            "Shrink world_update/instruction or raise the cap on BOTH peers."
+        )
+
     def send(self, target: str, cmd: dict[str, Any], timeout: float = 30.0) -> dict[str, Any]:
         """Send a command to a single peer and return the first response.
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes the diagnostics gap between the command validator's 64 KiB `world_update` bound and the transport's 16 KiB `strands_cmd_size_cap` rule on the `**/broadcast` key expression: `Mesh.broadcast` now pre-checks the encoded message size the same way `Mesh.send` already did, so an over-cap broadcast is logged as a client-side rejection instead of being silently dropped by the low-pass filter and returning the same `[]` a broadcast nobody answered returns. The check moves into a single `_cmd_topic_size_problem` helper that both publishers on the shared filter rule consult, so the two key expressions of one rule cannot drift apart on the cap they share. `send`'s error wording and its documented `ImportError` fail-open behaviour are unchanged and pinned by controls; the pending-RPC state cleanup on the new `broadcast` rejection path holds `_rpc_lock` and mirrors `send`'s existing pattern. Verified locally at `58e1aec0` (the head before the changelog-fragment-only commit `a43d33ac`; the code and test files are byte-identical between the two): all 19 new tests pass, 249 related mesh tests pass, `ruff check` / `ruff format --check` clean on both changed files.

## What's good

- Single-owner refactor is pinned structurally (`TestOneOwnerForTheSharedCap` walks the AST of both publishers), so an inline re-derivation of the cap fails a test rather than a review.
- `send`'s error text is pinned byte-for-byte by a control, making "send is unchanged" a measurement rather than a claim.
- The `broadcast` rejection reuses the method's existing `ValidationError` shape (warning log + `[]`) rather than inventing a second rejection shape, and the log line is plain ASCII.
- The `sys.modules` ImportError fixture uses `monkeypatch.setitem(..., None)`, the sanctioned pattern from AGENTS.md that restores the entry and orphans nothing.
- Scope discipline: the camera and safety filter rules are explicitly left alone with the reasoning stated, rather than mirrored blindly.
- The changelog fragment (`changelog.d/2622-broadcast-cmd-size-cap.md`) follows the fragment convention with a valid `### Fixed:` heading.

yinsong1986 (approved):
## Summary

Closes the diagnostics gap on the `**/broadcast` half of the transport's `strands_cmd_size_cap` rule: a validator-valid command (up to 64 KiB `world_update`) over the 16 KiB cmd-topic cap was silently dropped by the low-pass filter, so `Mesh.broadcast` returned the same empty list a broadcast nobody answered returns. The fix hoists `send`'s existing inline pre-check into a single owner, `Mesh._cmd_topic_size_problem`, and has both publishers on the shared filter rule consult it; `broadcast` reports the refusal through its existing client-side-rejection shape (warning log + `[]`), and `send`'s error wording is byte-identical (pinned by a control test). Verified locally at `a43d33ac`: the new 19-test file passes, and the 127 existing mesh RPC/broadcast/validation tests pass unchanged. No public API, wire format, env var, or error-wording change; the pending-turn state is cleaned up under `_rpc_lock` on the refusal path in both publishers; the `ImportError` fallback preserves pre-check-era behaviour rather than failing every command on a missing optional module. Scope discipline is right - the camera and safety filter rules have different contracts and are correctly left to a separate report.

## What's good

- Single-owner refactor instead of a copied inline check, with structural pins (`ast`-based) so a re-derived cap fails a test rather than drifting silently.
- Changelog fragment per AGENTS.md step 3; `monkeypatch.setitem(sys.modules, ..., None)` for the ImportError case per the sys.modules learning; ASCII-only log/error strings.
- The in-cap controls and the byte-identical `send` wording pin make "nothing else changed" a measurement rather than a claim.

PR #2623, opened 2026-08-22
feat(refusals): a continuable refusal carries a stable code and its subject

cagataycali:
## Problem

Some refusals are *continuable*: the request was well formed, and an operator who accepts the risk can grant something that makes the identical request succeed. An untrusted policy provider, a repo or host or policy type outside a mesh allowlist, a teleop frame past the value envelope -- each has an operator answer behind it. Anything that offers that answer (a UI consent card, an approval endpoint, a supervising agent) first has to recognise *which* refusal it is looking at and *what* it is about.

The only thing on offer was the message text. So recognition means prose matching: an env-var name appearing somewhere in the sentence, the subject pulled back out with an anchored regex. **Two of the refusals name no env var at all** -- `input frame value for ... out of range` and its slew sibling -- so they can only be recognised by their own words.

The measured consequence is that this package cannot tell when it breaks such a consumer. Rewording the teleop refusals from `out of range` to `exceeds the safety envelope` -- a plain improvement, and nothing in `tests/` asserts on either phrase:

| after that one reword | result |
| --- | --- |
| this package's own mesh suite (`tests/mesh/test_*input* test_*slew* test_*security*`) | **270 passed** |
| a consumer's classification of the value refusal | recognised -> **`None`** |
| a consumer's classification of the slew refusal | recognised -> **`None`** |

The wording change is silently breaking and the suite stays green, because prose is the only contract on offer.

## Change

Continuable refusals now carry a stable machine-readable `code` and the `subject` they are about, structurally:

```python
from strands_robots import refusal_codes
from strands_robots.mesh.security import ValidationError, validate_command

try:
    validate_command(cmd)
except ValidationError as refusal:
    if refusal.code == refusal_codes.HF_REPO_NOT_ALLOWED:
        offer_to_allowlist(refusal.subject)                  # the repo, already parsed out
        os.environ[refusal_codes.REFUSAL_GRANTS[refusal.code]] = ...   # STRANDS_MESH_HF_REPO_ALLOW
```

`strands_robots/refusal_codes.py` is a new stdlib-only leaf holding five codes, the closed `REFUSAL_CODES` vocabulary, and `REFUSAL_GRANTS` -- the operator grant that lifts each refusal, which is what a consumer otherwise hard-codes. `SecurityError` (so `ValidationError` and `LockoutError`) and `UntrustedRemoteCodeError` gained keyword-only `code` / `subject`, defaulting to `None`.

This follows the shape `episode_labels.FAILURE_MODES` already uses for the same reason, in its own words: *"Fixed so downstream filters match on identity rather than parsing prose; the free-text goes in `note`."*

### The six sites, measured end to end

| refusal | code | subject |
| --- | --- | --- |
| `_check_trust_remote_code` | `TRUST_REMOTE_CODE_REQUIRED` | `lerobot_local` |
| `pretrained_name_or_path` not allowlisted | `HF_REPO_NOT_ALLOWED` | `sketchy-org/model` |
| `policy_type` not allowlisted | `POLICY_TYPE_NOT_ALLOWED` | `mystery_net` |
| `policy_provider` not allowlisted | `POLICY_TYPE_NOT_ALLOWED` | `mystery_prov` |
| `policy_host` not allowlisted | `POLICY_HOST_NOT_ALLOWED` | `10.9.9.9` |
| `server_address` host not allowlisted | `POLICY_HOST_NOT_ALLOWED` | `tcp://10.9.9.9:5555` |
| teleop frame past the value envelope | `TELEOP_VALUE_OUT_OF_RANGE` | `shoulder.pos` |
| *`instruction exceeds 2000 chars`* | *`None`* | *`None`* |

### Additive, measured

Driving all eight refusals on this branch and on `upstream/main`: **messages byte-identical, codes 0 -> 7**. No message was touched, and no request is refused that was not refused before.

And the property the whole change exists to provide -- applying the same reword *to this branch*:

| reworded refusal | `code` | `subject` | prose matching |
| --- | --- | --- | --- |
| `policy_host` not allowlisted | `POLICY_HOST_NOT_ALLOWED` | `10.9.9.9` | `None` |
| teleop value envelope | `TELEOP_VALUE_OUT_OF_RANGE` | `shoulder.pos` | `None` |

The code and the subject survive; only the prose matching breaks. That is the contract.

## The boundary: whether an operator can grant something

A rejection with no grant behind it stays code-less on purpose. A schema or bounds failure gives a consumer nothing to offer, so there is nothing to recognise, and `code is None` means "show the message and stop" rather than "unknown code".

`test_every_refusal_naming_an_operator_grant_carries_a_code` derives the sites it must find from `REFUSAL_GRANTS` rather than a hand-written list, so a seventh grant-bearing refusal added later is graded on arrival. It finds exactly 6 today, all coded, with no exemption -- which is why the trust-remote-code raise passes `code=` explicitly rather than relying on a class default the scan could not see.

## Scope

Deliberately out, each with the measurement:

- **A `code` field on the wire.** `mesh/core.py:1908` returns `{"error": ...}` -- a dict, not an exception -- and the command-response envelope is its own contract. This change is exception-side only.
- **`input_frame_slew_violation`.** It *returns* `str | None`, consumed as a log line by `mesh/input.py:633` and `teleop_mixin.py:842`; giving it a code needs either a return-type change or a code embedded in the message text, and the latter is exactly the prose coupling this removes.
- **The mtls and calibration refusals** (`mesh/_zenoh_config.py:286`, `hardware_robot.py:1405`). Neither is continuable in the sense above, and no measured consumer classifies them.
- **The other ~67 `ValidationError` sites.** They are schema and bounds failures with no operator grant, so they stay code-less by the boundary rule rather than by omission.

## Verification

Measured at `207afce5`.

- **Pre-fix split**: 19 failed / 2 passed, of which **11 fail behaviourally** (`AttributeError: 'ValidationError' object has no attribute 'code'` x10, plus `TypeError: ValidationError() takes no keyword arguments`). The other 8 can only report the new registry module's absence; they pin the vocabulary rather than the regression. 21 passed after. The per-site tests spell their codes as literals precisely so they fail behaviourally, with one drift test tying those literals to `REFUSAL_CODES`.
- **Break table** -- 7 breaks, 7 distinct firing sets, no dead guards:

| break | fails | guards fired |
| --- | --- | --- |
| B1 revert both source files | 19 | all |
| B2 drop the teleop `code=` | 2 | that site + every-declared-code-is-raised |
| B3 drop the trust-remote-code `code=` | 3 | that site + declared-code-is-raised + **the AST grant guard** |
| B4 drop the teleop `subject=` | 1 | that site only |
| B5 claim a code on `instruction exceeds ...` | 1 | **the boundary control only** |
| B6 point the package scan at an empty subtree | 2 | non-vacuity + declared-code-is-raised |
| B7 declare a code no site raises | 4 | registry closure |

- **Blast radius**: the identical 316-file selection on both trees (every test naming the changed
  symbols, every guard that walks the package or reads a docs page, plus the changelog guards), with the
  new test file excluded from both:

| tree | result |
| --- | --- |
| this branch | 19 failed, 14788 passed, 60 skipped |
| pristine `upstream/main` | 17 failed, 14789 passed, 60 skipped |

  The 17 shared failures are all `ModuleNotFoundError` in `tests/mesh/test_iot_*` (9 +
  8), which need the `awsiot` extra CI installs. The 2 branch-only ids are both in
  `tests/mesh/test_zenoh_transport_security.py`, which opens real localhost Zenoh sessions and
  cross-talks when two suites run in parallel: that file alone is **9 passed, 2 skipped on both
  trees**. Collected differs by exactly 1 -
  `test_no_teardown_method_imports_a_stdlib_module[refusal_codes.py]`, the package-wide guard
  enrolling the new module, which passes.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` **24 == 24** against a pristine `upstream/main` worktree, branch-only empty.
- Import cost: `mesh.security` already pulls +1118 modules via the package `__init__`; with this change it pulls **+1119** -- the one new leaf, no new transitive dependency.

## Artifact

There is no simulation behaviour to film: no message, no signature and no accepted-input set changed, and the figure below is the measurement instead -- each column produced by running one capture script against that tree.

refusal codes are the stable contract

### DIFF
--- a/changelog.d/2623-refusal-codes-stable-contract.md
+++ b/changelog.d/2623-refusal-codes-stable-contract.md
@@ -1,0 +1,24 @@
+### Added: a stable code and subject on the refusals an operator can answer
+
+Some refusals are *continuable*: the request was well formed, and an operator who
+accepts the risk can grant something that makes the identical request succeed.
+Anything that offers that choice - a consent card, an approval endpoint, a
+supervising agent - has to recognise which refusal it is and what it is about,
+and the only thing on offer was the message text. So recognition meant prose
+matching: an env-var name appearing somewhere in the sentence, the subject pulled
+back out with an anchored regex. Two of the refusals name no env var at all.
+
+That made every wording improvement silently breaking, and this package could
+not detect it. Rewording the teleop refusals from "out of range" to "exceeds the
+safety envelope" - a plain improvement, and nothing in `tests/` asserts on either
+phrase - leaves the mesh suite at 270 passed while a consumer's classification of
+both refusals drops to `None`.
+
+Continuable refusals now carry a stable machine-readable `code` from the new
+`strands_robots.refusal_codes` and the `subject` they are about, so a consumer
+switches on identity and the message stays free to improve.
+`REFUSAL_GRANTS` names the environment variable that lifts each refusal, which a
+consumer otherwise hard-codes. The codes are additive: all eight measured refusal
+messages are byte-identical to before, and a rejection with no operator grant
+behind it - a schema or bounds failure - stays code-less on purpose, so
+`code is None` means "show the message and stop" rather than "unknown code".
--- a/docs/security.md
+++ b/docs/security.md
@@ -97,6 +97,32 @@ What this looks like in practice, and how to configure it:
 
 Reference: `strands_robots.tools.robot_mesh`.
 
+## Refusal codes are the stable contract; prose is not
+
+Some refusals are *continuable*: the request was well formed, and an operator who accepts the risk can grant something that makes the identical request succeed. An untrusted policy provider, a repo or host or policy type outside a mesh allowlist, a teleop frame past the value envelope - each of these has an operator answer behind it. Anything that offers that answer (a UI consent card, an approval endpoint, a supervising agent) first has to recognise *which* refusal it is looking at and *what* it is about.
+
+Recognise it by its `code`, never by its message text:
+
+```python
+from strands_robots import refusal_codes
+from strands_robots.mesh.security import ValidationError, validate_command
+
+try:
+    validate_command(cmd)
+except ValidationError as refusal:
+    if refusal.code == refusal_codes.HF_REPO_NOT_ALLOWED:
+        offer_to_allowlist(refusal.subject)          # the repo, already parsed out
+        print(refusal_codes.REFUSAL_GRANTS[refusal.code])   # STRANDS_MESH_HF_REPO_ALLOW
+```
+
+- **`code` is stable; the message is not.** `code` is a member of `refusal_codes.REFUSAL_CODES`, a closed vocabulary you may switch on. The message is an operator-facing sentence and may be reworded at any time to explain the refusal better. Matching on prose - looking for an env-var name in the sentence, or pulling the subject back out with a regex - couples you to wording that is free to change, and neither this package's tests nor yours will notice when it does.
+- **`subject` is what the refusal is about**, so you do not have to parse it back out: the repo id, the host or whole `server_address`, the policy type or provider name, the joint key, the refused provider.
+- **`REFUSAL_GRANTS` names the environment variable that lifts each refusal.** Read it from there rather than hard-coding it, so your consumer and this package cannot drift apart.
+- **A refusal with no code is not continuable.** `code` is `None` for rejections an operator cannot grant their way past - a schema failure, an over-long instruction, a lockout. There is nothing to offer, so there is nothing to recognise. Treat `code is None` as "show the message and stop", not as an unknown code.
+- **Codes are additive.** They were introduced without changing a single refusal message, and new codes may be added for refusals that become continuable later. Switch on the codes you know and fall through to the message for the rest.
+
+Reference: `strands_robots.refusal_codes`; `strands_robots.mesh.security.SecurityError`; `strands_robots.policies.factory.UntrustedRemoteCodeError`.
+
 ## HuggingFace policy code execution (`trust_remote_code`)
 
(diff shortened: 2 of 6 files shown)


yinsong1986 (approved):
## Summary

Adds a stable, machine-readable contract to the six continuable refusals: `SecurityError` (so `ValidationError`/`LockoutError`) and `UntrustedRemoteCodeError` gain keyword-only `code`/`subject` attributes defaulting to `None`, a new stdlib-only leaf `strands_robots.refusal_codes` holds the closed five-code vocabulary plus `REFUSAL_GRANTS` (the env var that lifts each refusal), and the six raise sites pass their code and subject. Verified locally at `207afce5`: no circular import in either direction (`import strands_robots` first and `import strands_robots.mesh.security` first both resolve, since `policies/__init__` eagerly pulls `factory` which now imports the registry through the partially-initialized package); an AST scan of `mesh/security.py` confirms no call site constructs these exceptions with more than one positional arg, so the keyword-only widening breaks nothing; refusal messages are unchanged in the diff; every `REFUSAL_GRANTS` env var already exists and is documented in README. The new test file passes (21) and the affected mesh suites pass (312 passed, 2 skipped across `security`/`input`/`slew` selections).

## What's good

- The additive-by-construction shape: keyword-only params with `None` defaults, byte-identical messages, and the boundary rule (no grant -> no code) stated and pinned rather than left implicit.
- The grant-derived AST guard (`test_every_refusal_naming_an_operator_grant_carries_a_code`) grades future grant-bearing refusals on arrival instead of keeping a hand-written site list, with a non-vacuity check on the scan itself.
- `REFUSAL_GRANTS` closes over `REFUSAL_CODES` exactly (pinned), and every named env var is an existing, README-documented one - no new env surface.
- Changelog fragment, `docs/security.md` consumer guidance, and scope discipline (wire envelope and the `str | None`-returning slew check deliberately excluded, with the reasoning recorded).

PR #2624, opened 2026-08-22
fix(mesh): a refused teleop frame is counted under the guard that refused it

cagataycali:
## Problem

`InputReceiver._on_input` refuses a teleop frame for six distinct reasons. Two of them keep their
own counter -- `rate_dropped` (the apply-rate ceiling) and `slew_rejected` (the per-joint slew
bound). The other four shared `rejected`:

| refusal | counter before |
|---|---|
| arrived during an E-stop lockout | `rejected` |
| frame `t` missing or non-numeric | `rejected` |
| frame `t` stale or too far in the future | `rejected` |
| `validate_input_frame` refused the shape or a value | `rejected` |
| over the apply-rate ceiling | `rate_dropped` |
| over the per-joint slew bound | `slew_rejected` |

Two things followed from the sharing, and the second is what makes the first unrecoverable rather
than merely awkward.

**A report cannot say which guard refused the stream.** `rejected: 8` is the whole answer, and the
reason exists only in the follower's log line.

**Each guard's log line is rate-limited against the shared counter**, so a stream refused for one
reason spends the budget and the *first* refusal of a different reason is never written. Measured on
one receiver, same frame both times:

| the stream | `rejected` | lines written | is the value refusal stated? |
|---|---|---|---|
| the out-of-range frame arrives first | 1 | 1 | yes -- `value for '1' out of range: \|1500.0\| > 720.0` |
| five stale-`t` frames, then the same frame | 6 | 5 | **no** |

Both rows refuse the same frame for the same reason. Only the second is undiagnosable, and the
difference is how many *unrelated* frames preceded it -- an ordinary clock-skewed start.

The `stats` docstring also enumerated causes that did not match the code. It named "ACL checks"
(the string `acl` appears nowhere else in the module, and neither `_on_input` nor
`validate_input_frame` makes such a check) and omitted the frame-validation refusal, which is the
one that names an offending value and its bound.

## Change

`_refuse(cause, message, *args)` becomes the single place a refusal on this path is accounted for.
It counts the frame under its cause, adds it to the `rejected` total, and spends
`_REFUSAL_LOG_BUDGET` **per cause**. `stats` reports the breakdown beside the total, derived from
the declared causes so a cause added later is reported without a second list to update:

```
rejected            8      <- unchanged: still the total
rejected_lockout    0
rejected_freshness  5
rejected_invalid    3
```

`rejected` always equals their sum, so nothing that reads it today changes. The two guards that
already had their own counter are untouched, and no refusal decision moves: both trees refuse the
same eight frames in the measurement above.

The docstring now names the three causes it reports and drops the ACL claim.

## Not in scope

* **A stable code vocabulary.** The breakdown keys are counter names, not an error-code taxonomy;
  `_REJECTION_CAUSES` stays module-private so it cannot become one by accident.
* **`last_refusal` / `last_frame_t` / publisher timestamps.** Each adds a new reported field with a
  freshness contract of its own; this change only makes the counters already reported carry which
  guard produced them.
* **Narrowing any of the four guards.** Their refusal conditions are unchanged.
* The reported shape is documented at its owner, `InputReceiver.stats`, rather than duplicated into
  `TeleopMixin.get_teleop_status`, which forwards `rcv.stats` verbatim.

## Tests

`tests/mesh/test_input_refusal_causes.py`, 15 tests. Against `upstream/main`: **11 failed / 4
passed**. The headline reads

```
AssertionError: the value refusal was counted and never stated: rejected=6 with no line naming
the value or the bound, so the budget spent by the replay-freshness guard silenced a different
guard's only report
```

Nine of the eleven fail behaviourally or structurally with the offending state in the message (the
full `stats` dict, the four call sites, the ACL claim); two can only report that the new symbols are
absent, and pin the root cause rather than the regression.

Every guard is load-bearing -- each break fires exactly its own:

| break | fires |
|---|---|
| revert `input.py` | all 11 |
| log budget measured against the shared total again | headline + the budget's structural pin |
| docstring names ACL again | the ACL pin |
| docstring stops naming the reported keys | the key-naming pin |
| one site counts the total directly again | lockout attribution + single-owner + cause coverage |
| a declared cause is never used (both freshness sites mislabelled) | freshness attribution + the sum + headline + cause coverage |
| breakdown absorbs `slew_rejected` | cause coverage + the vocabulary pin + the separate-counter control |
| control (restored) | 0 of 15 |

Four tests pass on both trees and are the no-overreach controls: a rate-capped frame is still not
`rejected`, the breakdown does not absorb the two separately-counted causes, one cause still goes
quiet once *its own* budget is spent (so the fix is not "log every refusal" in a 50 Hz loop), and
`rejected == 1` for a single lockout refusal -- the contract
`test_input_stream_lifecycle.py::test_estop_lockout_rejects_frame` already pins.

## Verification

Measured at `9e01c20c`; the head is now `ebe9c6d2`, which adds only the changelog fragment (`git diff 9e01c20c..ebe9c6d2 -- '*.py'` is empty), so the numbers below still stand.

Identical 446-file selection on both trees, with the new test file excluded from each so the
comparison measures the source change alone:

|---|---|---|---|---|
| branch | 50 | 14794 | 149 | 24 |
| `upstream/main` | 48 | 14796 | 149 | 24 |

70 failing ids are shared and all dep-absent locally: 45 `tests/mesh/test_iot_*`
(`No module named 'awsiot'`), 23 across `test_robot_mesh_dc_dispatch_errors.py` and
`test_device_connect_hardening.py` (`device_connect_agent_tools`), 2 in `tests/training/test_lerobot.py`
(installed draccus 0.10.0 against lerobot 0.6.2's `draccus>=0.11.6` floor). All three are declared
extras that CI installs.

The six ids that differ are contention, proven by isolation on both trees:
`tests/mesh/test_zenoh_transport_security.py` fails with a *different* subset on each tree (3 on the
branch, 2 different ones on `upstream/main`) because it opens real localhost Zenoh sessions while
several suites run in parallel -- run alone it is `9 passed, 2 skipped` on **both** trees. And
`test_eval_async_rtc_masks_inference_latency` is wall-clock sensitive: 3/3 isolated runs pass on both
trees. Neither file is touched by this change.

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24
against a pristine `upstream/main` worktree with no branch-only error.

## Artifact

The shipped receiver driven against a real MuJoCo SO-101 (`MUJOCO_GL=egl`) with the stream above --
five stale-`t` frames, then three the leader reports in degrees into a 720-unit envelope. Both trees
refuse all eight (`frames_received: 0`, the follower never moves, and the render is identical to
within `5.4e-05` mean level). What changes is what a consumer reading the counters and the log can
conclude: on `upstream/main` the refusal is unrecoverable and the verdict stops at "every frame is
being refused"; after the change the same consumer reaches "the envelope is 720 units, and this
leader reported joint 1 at 1500 -- the arm reports DEGREES while the envelope assumes radians".

teleop refusal causes

### DIFF
--- a/changelog.d/2624-teleop-refusal-cause-accounting.md
+++ b/changelog.d/2624-teleop-refusal-cause-accounting.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- **A refused teleop frame is now counted under the guard that refused it, and each guard's
+  rate-limited log line is spent on its own budget.**
+  `InputReceiver` refuses a frame for six reasons; the apply-rate ceiling and the per-joint slew bound
+  each reported themselves in `rate_dropped` and `slew_rejected`, while an E-stop lockout, a stale or
+  missing frame timestamp, and a frame `validate_input_frame` refused all shared `rejected`. Beyond
+  leaving a report unable to say which guard refused a stream, the warning each guard emits was
+  rate-limited against that shared counter, so a follower that had already refused a few frames for
+  one reason -- an ordinary clock-skewed start -- refused the next reason in complete silence, and the
+  log line is the only place a refusal names the value it refused and the bound it exceeded. `stats`
+  now reports `rejected_lockout`, `rejected_freshness` and `rejected_invalid` beside the unchanged
+  `rejected` total, which always equals their sum, and the budget is spent per cause. No refusal
+  condition changes: the same frames are refused, and the two causes that already had their own
+  counter keep it. The `stats` docstring also enumerated an ACL check this path never makes while
+  omitting the frame-validation refusal, and now names the causes it reports.
--- a/strands_robots/mesh/input.py
+++ b/strands_robots/mesh/input.py
@@ -73,6 +73,19 @@ _MAX_LOGGED_LOOP_ERRORS = 5
 INPUT_MAX_HZ_DEFAULT = 100.0
 
 
+#: How many refused frames of ONE cause are logged before that cause goes quiet.
+#: Spent per cause rather than per receiver: a cause that keeps refusing must not
+#: silence the FIRST refusal of a different one, because the log line is the only
+#: place a refusal states which value it refused and against which bound.
+_REFUSAL_LOG_BUDGET = 5
+
+#: The refusal causes that share the ``rejected`` total, in the order
+#: :meth:`InputReceiver._on_input` checks them. Each is reported as
+#: ``rejected_<cause>`` in :attr:`InputReceiver.stats` and spends its own share of
+#: :data:`_REFUSAL_LOG_BUDGET`.
+_REJECTION_CAUSES: tuple[str, ...] = ("lockout", "freshness", "invalid")
+
+
 def _input_max_hz() -> float:
     """Resolve ``STRANDS_MESH_INPUT_MAX_HZ`` (lazy, restart-free).
 
@@ -401,6 +414,10 @@ class InputReceiver:
         self._last_seq = -1
         self._drops = 0
         self._rejected = 0
+        # Per-cause breakdown of ``_rejected``. Every cause that shares the total
+        # keeps its own count, so a report can say WHICH guard refused the stream
+        # and each cause spends its own log budget.
+        self._rejected_by_cause: dict[str, int] = dict.fromkeys(_REJECTION_CAUSES, 0)
         self._rate_dropped = 0
         self._slew_rejected = 0
         self._last_rate_gate_mono = 0.0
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Gives `InputReceiver` per-cause accounting for the four refusal reasons that previously shared one `rejected` counter, routed through a single `_refuse(cause, message, *args)` owner. The real fix is the log budget: it is now spent per cause instead of against the shared total, so a stream refused for one reason (e.g. a clock-skewed start burning the budget on stale-`t` frames) can no longer silence the first — and only — log line naming a different guard's refused value and bound. `rejected` stays the total and always equals the breakdown's sum, the two guards that already had their own counters (`rate_dropped`, `slew_rejected`) are untouched, and no refusal condition changes. The new `rejected_<cause>` stats keys are additive and derived from the declared `_REJECTION_CAUSES` vocabulary (kept module-private, per the stated non-goal of not minting an error-code taxonomy), so nothing reading today's counters changes. The stale ACL claim in the `stats` docstring is corrected in the same pass.

## What's good

- Verified locally at `ebe9c6d2`: the 15 new tests pass, the neighbouring `test_input_stream_lifecycle.py` / `test_input_validation.py` suites (68 tests) pass unchanged, `ruff check` / `ruff format --check` are clean on both touched Python files, and all three changed files are ASCII-clean.
- The structural pins are well-aimed: every `self._rejected` increment must go through `_refuse`, the budget gate must not read the shared total, and every call site must name a declared cause — so the fix cannot be undone one refusal site at a time.
- Good no-overreach controls: a rate-capped frame is still not `rejected`, the breakdown does not absorb the two separately-counted causes, one cause still goes quiet after its own budget (so this is not "log every refusal" in a 50 Hz loop), and the pre-existing `rejected == 1` lockout contract holds.
- Changelog fragment follows the `changelog.d/<pr-number>-<slug>.md` convention (AGENTS.md PR workflow step 3); the `stats` docstring names every reported key and is itself pinned against drift.

PR #2625, opened 2026-08-22
fix(mesh): a peer does not report the age this process observed of it

cagataycali:
## Problem

`PeerInfo.to_dict()` spread the peer's own presence payload **after** the four fields this
process decides about the peer, so a payload carrying `age`, `peer_id`, `type` or `hostname`
replaced the local reading:

```python
return {
    "peer_id": self.peer_id,
    "type": self.peer_type,
    "hostname": self.hostname,
    "age": round(self.age, 1),
    **self.caps,          # <- the peer's presence payload, spread last
}
```

`caps` is the parsed presence JSON, handed to `update_peer(..., caps=data)` straight from
`_on_presence`. Its purpose is capabilities - `tool_name`, `connected`, `cameras` - and
nothing validated the rest of its key space.

`age` is the field that matters most, because it is not a capability. `last_seen_mono`'s own
docstring calls it *"a local observation, never a stamp the peer sent, and it is not
serialised: `to_dict` reports `age`"*, and `docs/mesh.md` names it among the things the mesh
decides from a duration on `time.monotonic()` *"which no NTP correction, `date -s` or resume
from suspend can move"*. A peer publishing `"age": 0.0` set it anyway.

`peer_id` is the second: it is the key the registry files a peer under, the key
`Mesh.peers_by_id` and `Mesh.get_peer` look it up by, and the key `Mesh.peers` filters *self*
out by.

## Measured, through the real `Mesh._on_presence`

Two peers admitted; the second's payload also carries `age`, `peer_id` and `type`.

| tree  | registry holds | `Mesh.peers` reports | `peers_by_id` | `get_peer("evil")` |
| before | `['evil', 'good']` | `peer_id='good'` **twice** | `['good']` | `None` |
| after | `['evil', 'good']` | `good`, `evil` | `['evil', 'good']` | the peer |

Before, `evil`'s payload answered a lookup for `good` (`type='operator'`, `age=0.0`,
`hostname='spoofed'`) and `evil` itself was unreachable. `age` read `0.0` for a peer this
process had last heard from 0.5s earlier; after, both read `0.5`.

A peer whose payload names **our own** `peer_id` is a hiding primitive - `Mesh.peers` filters
self by that key:

| tree  | registry holds | `Mesh.peers` reports |
| --- | --- | --- |
| before | `['ghost']` | `[]` |
| after | `['ghost']` | `['ghost']` |

## Change

`caps` merges **first**, so the four locally decided fields win a name collision. The
spread's purpose is unchanged.

**No honest payload changes.** Of the twelve keys `_build_presence` emits, `hostname` is the
only one that collides with a local name, and `update_peer` derives the local `hostname` from
that same wire value - so the reorder is a value no-op for it. Measured on both trees for a
full honest payload: 15 keys, and `{"age": 1.0, "hostname": "cagatay", "peer_id": "g1",
"type": "sim"}` identical.

## Why now

`**self.caps` dates from #101 (2026-05-21), the original mesh PR. **88 days later**, #2408
(2026-08-17) moved peer liveness onto `time.monotonic()`, wrote the `last_seen_mono` docstring
quoted above and the `docs/mesh.md` paragraph, and stated in its own commit message:

> `to_dict()` reports age and never the base, so no serialised surface changes.

It made `age` immune to a wall-clock step and left it settable by the peer being measured,
which is the stronger override of the two.

## Scope

The fix is the field ordering. Deliberately **not** here:

- **A `stale: bool` verdict or an `on_peer_lost` callback.** The SDK already has a verdict on
  absence - `prune_peers()` runs on every heartbeat tick and drops a peer past
  `PEER_TIMEOUT` (10s), returning the ids it dropped - and `age` is already reported. Adding
  an intermediate state or a callback is new public surface.
- **Making `PEER_TIMEOUT` operator-tunable.** Its sibling bound in the same module
  (`STRANDS_MESH_MAX_PEERS`) is, via a lazy `_max_peers()`; this one is a module constant, and
  because it is bound as `prune_peers(timeout=PEER_TIMEOUT)` the default is fixed at import,
  so reassigning the module attribute does not change it. Both are worth a decision, neither
  is this fix.

## Verification

`tests/mesh/test_peer_age_is_a_local_observation.py`, 17 items. Against `upstream/main`:
**12 failed / 5 passed**; after, 17 passed. The 5 that pass on both are the no-overreach
controls - capability keys still merge, an honest payload reports every key, `age` is still a
rounded non-negative duration.

Every guard is load-bearing (each break applied to the fixed tree):

| break | tests it fires |
| --- | --- |
| revert the order (`caps` last) | 12 - the full regression set |
| `caps` first but drop local `age` | 4 - both age tests, the rounding control, the structural pin |
| `caps` first but drop local `peer_id` | 8 - every masquerade test, the honest-payload control, the structural pin |
| drop the `caps` merge entirely | 3 - both capability controls, the structural pin |
| control (fixed tree) | 0 of 17 |

The last row of the third and fourth breaks is why the capability controls exist: they fire
for the tempting wrong fix - dropping the spread - and for nothing else.

`TestTheOrderIsTheContract` pins the ordering structurally off the AST, so a later edit cannot
reintroduce the override by moving the spread back.

Gate: the 421 test files that name the peer registry, its serialised view, `tests/mesh`, or a
repo-wide doc/docstring/changelog guard, run as the **identical selection on both trees** with
this PR's new file excluded from each:

| tree | result |
| --- | --- |
| this branch | `21 failed, 14328 passed, 86 skipped, 24 errors` |
| pristine `upstream/main` | `21 failed, 14328 passed, 86 skipped, 24 errors` |

Byte-identical, 45 failing ids each, and the id sets are equal in both directions - so nothing
else in that selection changed verdict. All 45 are `tests/mesh/test_iot_*` on
`No module named 'awsiot'`, the declared `[mesh-iot]` extra (`awsiotsdk>=1.21.0`) absent on this
box and installed in CI. `mypy strands_robots tests tests_integ` reports the same 24 errors on
both trees, none in the changed files; `ruff check` / `ruff format --check` clean.

## Fleet view

The same three heartbeats through the real `Mesh._on_presence`: two honest `so101` robots and
one whose payload also carries `age` / `peer_id` / `type`. `so101-b` stopped heartbeating 6s
ago, past a 5.0s threshold.

fleet view before and after

`so101-b` reads `STALE` in **both** columns, and that is the point worth being precise about:
the defect is not that an honest peer's age is misjudged, it is that a peer chooses what is
judged. Before, the view carries three rows under two ids - `so101-a` appears twice, once as
`robot` and once as `operator` - `rogue-1` is in the registry while `get_peer("rogue-1")`
returns `None`, and the rogue peer reads `0.0s / fresh`. After, three rows, three ids, all
three reachable.

### DIFF
--- a/changelog.d/2625-peer-age-is-a-local-observation.md
+++ b/changelog.d/2625-peer-age-is-a-local-observation.md
@@ -1,0 +1,29 @@
+### Fixed
+
+A peer's presence payload no longer decides what this process reports having
+observed about it. `PeerInfo.to_dict()` spread the payload (`caps`) after the
+four fields the local process decides -- `peer_id`, `type`, `hostname` and
+`age` -- so a heartbeat carrying any of those names replaced the local reading.
+
+`age` is the one that is not a capability at all: `last_seen_mono`'s docstring
+calls it "a local observation, never a stamp the peer sent", and `docs/mesh.md`
+names it among the things the mesh decides from a duration on `time.monotonic()`
+"which no NTP correction, `date -s` or resume from suspend can move". A peer
+publishing `"age": 0.0` in its heartbeat reported itself perpetually fresh, so
+every staleness verdict read from that field was the sender's to choose rather
+than the reader's to measure.
+
+`peer_id` is the second, and it costs a lookup rather than a verdict: it is the
+key the registry files a peer under, the key `Mesh.peers_by_id` and
+`Mesh.get_peer` resolve by, and the key `Mesh.peers` filters *self* out by.
+Measured through `Mesh._on_presence` with two peers admitted, the second naming
+the first in its payload: the registry held both, `Mesh.peers` reported one id
+twice, `peers_by_id` resolved only one of them, `get_peer` returned `None` for a
+peer that was present, and the second peer's capabilities answered a lookup for
+the first. A payload naming the *reader's* own id disappeared from `Mesh.peers`
+entirely while staying in the registry.
+
+`caps` now merges first, so those four win a name collision. The spread's purpose
+is unchanged and no honest payload changes: of the twelve keys `_build_presence`
+emits, `hostname` is the only one that collides, and `update_peer` derives the
+local `hostname` from that same wire value.
--- a/docs/mesh.md
+++ b/docs/mesh.md
@@ -97,6 +97,15 @@ that peer has timed out (10s), which peer the registry evicts when it hits
 move. So bringing a robot's clock into sync to satisfy the forward bound above
 will not make the next heartbeat tick drop every peer it can still hear.
 
+**Nor are those durations the measured peer's to report.** `age` is *your*
+process's reading of when it last heard from a peer, and the `peer_id` a peer is
+filed under is the one its topic and certificate bind - not a field inside the
+payload. A presence payload is merged into what you read about a peer so you get
+its capabilities (`tool_name`, `connected`, `cameras`, ...), and those four
+locally decided keys - `peer_id`, `type`, `hostname`, `age` - win a name
+collision with it. A peer heartbeating `"age": 0` does not report itself fresh,
+and one naming another peer's id does not answer a lookup for that peer.
+
 Repeated wrong codes arm a brute-force cooldown
 (`STRANDS_MESH_RESUME_MAX_FAILS`, `STRANDS_MESH_RESUME_BACKOFF_S`): during the
 cooldown even the correct code is refused, so wait it out rather than retrying in
--- a/strands_robots/mesh/session.py
+++ b/strands_robots/mesh/session.py
@@ -317,13 +317,26 @@ class PeerInfo:
         return time.monotonic() - self.last_seen_mono
 
     def to_dict(self) -> dict[str, Any]:
-        """Serialise to a plain dict (JSON-friendly)."""
+        """Serialise to a plain dict (JSON-friendly).
+
+        ``caps`` merges *first* so the four locally-decided fields win a name
+        collision. ``caps`` is the peer's own presence payload, and every field
+        below it is something this process decided about the peer rather than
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

This PR closes a peer-spoofing hole in `PeerInfo.to_dict()`: the presence payload (`caps`) was spread *after* the four locally-decided fields, so a heartbeat carrying `age`, `peer_id`, `type` or `hostname` replaced the reader's own observation -- letting a peer report itself perpetually fresh (`"age": 0.0`), answer lookups for another peer's id, or hide itself from `Mesh.peers` by naming the reader's own `peer_id`. The fix is the minimal one: `caps` merges first, so the local fields win a name collision, and the spread's capability-merging purpose is unchanged. The key set of the serialised view is identical and honest payloads are a value no-op (`hostname` is the only honest collision, and `update_peer` derives the local value from the same wire field), so nothing consuming `get_peers()` / `get_peer()` changes shape.

## What's good

- The fix direction matches the codebase's own contracts: `last_seen_mono`'s docstring and `docs/mesh.md` both already state `age` is a local observation, so this is enforcing documented semantics, not changing them.
- The regression suite is thorough without over-reaching: unit-level precedence tests, end-to-end masquerade tests through the real `Mesh._on_presence`, no-overreach controls pinning that capability keys still merge, and an AST structural pin (`TestTheOrderIsTheContract`) so the spread cannot quietly move back.
- Scope discipline: the description explicitly declines the tempting adjacent work (`stale` verdict, tunable `PEER_TIMEOUT`) with reasons.
- Changelog fragment present per AGENTS.md step 3; docs updated in the same PR; added lines are plain ASCII with no host paths.

Verified locally on the PR head: the new file passes 17/17, and the wider `tests/mesh` peer/session selection passes with only the pre-existing `awsiot`-environment errors the PR description already accounts for.

PR #2626, opened 2026-08-22
fix(mesh): a subscription never silently fails to exist

cagataycali:
## Problem

`Mesh.subscribe` has three `None` returns. One of them - a raising `declare_subscriber` - has always logged a WARNING naming the topic, in line with the six other client-side refusals in `strands_robots/mesh/core.py` (`send`, `broadcast`, the invalid-cmd and lockout rejections, the startup declare failure, `declare_publisher`). The other two answered `None` with no record at all:

```python
if not self._running:
    return None
session = current_session()
if session is None:
    return None
```

Those two are the pair a caller meets on the only rejoin the class offers. `stop()` drops every subscription `subscribe` records and clears `inbox`; `start()` re-declares only the peer's own built-in topics, and the `(topic, callback)` pairs behind a user subscription are not retained anywhere, so a consumer that leaves and rejoins re-subscribes. If the mesh has not come back up, that re-subscribe answers `None` and says nothing - and `stop()` never said the subscription had gone in the first place.

Measured end to end against a live Zenoh session (`STRANDS_MESH_LOCAL_DEV=1`), one peer, one topic:

| phase | `upstream/main` | this branch |
| --- | --- | --- |
| join, subscribe, publish | `subscribe -> "state"`, 1 callback fire | identical |
| `stop()` | `Zenoh mesh session closed` / `off mesh` - **no mention of the subscription** | + `dropped 1 subscription(s) on leaving the mesh (state); re-declare after start()` |
| state after `stop()` | `user_subs=[]  inbox=[]` | identical |
| re-subscribe before the rejoin finishes | `None`, **0 log records** | `None`, `WARNING ... subscribe(strands/*/state) refused: peer is not on the mesh (start() first)` |
| rejoin + re-declare | `subscribe -> "state"`, 1 fire, `peer_id` kept, lockout still engaged (`last_estop_ts 1755900000.0`) | identical |

## Change

The two silent refusals now say why at WARNING, naming the topic and which refusal it was, so all three `None` returns report themselves. `stop()` reports at INFO how many subscriptions it dropped and their names, only when there were any - a rejoining caller is told what it has to re-declare rather than inferring it from an empty `_user_subs`.

`subscribe`'s docstring gains a `Returns:` block accounting for all three refusals and states that a subscription does not survive `stop()`; `stop`'s states what it drops and why `start()` cannot restore it. `docs/mesh.md` gains a **Rejoining the mesh** subsection saying the same thing to a consumer: what the round trip keeps (`peer_id`, an engaged lockout) and what it does not.

No behaviour changes. The accepted path, the pre-existing declare-failure WARNING, the `peer_id` and an engaged e-stop lockout are all unchanged across `stop()`, and each is a control test.

## Scope

A public `Mesh.reconnect()` verb is **not** in this PR, and measuring the round trip is why. Of what such a verb would have to guarantee, the peer identity and the safety lockout already survive `stop()`/`start()` unchanged, and presence resumes; the one thing that does not is the user's own subscriptions. Restoring those needs `subscribe` to retain the `(topic, callback)` pair it currently discards, and a default for whether a reconnect preserves or drops them - a new public surface plus a contract choice, not a report. Worth noting too that the raw-session consumers that hand-roll this dance today call `get_session()` / `release_session()` and declare their own subscribers directly, so a `Mesh`-level verb would not serve them; a session-level one is a separate question again.

What this PR does is make the existing round trip honest, which is the part that needs no decision.

## Tests

`tests/mesh/test_subscribe_publish_step_resilience.py` already pinned this surface - including the `None` *return* for two of the three refusal paths - and said nothing about the report, which is why the two that answered without a word were never noticed. Extended there rather than in a new file, so its ten existing cases are controls.

Pre-fix (this file on `upstream/main` source): **5 failed / 14 passed**, four of them behavioural -

```
subscribe answered None and said nothing
a re-subscribe after stop() answered None and said nothing
stop() discarded two subscriptions and said nothing
subscribe documents no return
```

Post-fix: 19 passed. Each break fires exactly and only its own guard:

| break | fires |
| --- | --- |
| control (unmodified) | 0 of 19 |
| revert `core.py` to `upstream/main` | all 5 |
| drop only the two `subscribe` reports | 3 (before start / after stop / no session) |
| drop only the `stop()` report | 1 |
| make the `stop()` report unconditional | 1 - the no-noise control |
| drop only the `Returns:` block | 1 |
| report at DEBUG instead of WARNING | 3 |

## Verification

Measured at `80e69b73`, base `58b9a793`. Head is now `fa7bbe6a`, which adds two test-only commits: the `subscribe()` calls hoisted out of their asserts (the call is the premise, and `python -O` strips an assert wholesale), and the stand-in session bound to a local as the three pre-existing cases in that file already do - `lambda: MagicMock()` was four `py/unnecessary-lambda` notes and the odd one out. `strands_robots/` is untouched by both; the pre-fix split and the break table re-measured unchanged at the new head.

- Gate: 234 files (all of `tests/mesh`, every test naming `subscribe` / `_user_subs` / `docs/mesh.md`, plus the repo-wide docs, changelog, ASCII, docstring-xref and dependency guards), the identical selection on both trees with the changed test file excluded from each: **`21 failed, 6162 passed, 8 skipped, 24 errors` byte-identical**, 45 failing ids each, `comm` empty in both directions. All 45 are `tests/mesh/test_iot_*` on `No module named 'awsiot'` (88 occurrences), a declared extra CI installs.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `upstream/main` worktree, none in the changed files.
- `ruff check` / `ruff format --check` clean over the CI scope.

## Artifact

The rejoin driven end to end against a live Zenoh session, once per tree - same script, same topic, same timings, only `strands_robots` differs. Every measured value is identical in the two columns; what changes is what the peer says.

rejoin

### DIFF
--- a/changelog.d/2626-a-subscription-never-silently-fails-to-exist.md
+++ b/changelog.d/2626-a-subscription-never-silently-fails-to-exist.md
@@ -1,0 +1,33 @@
+### Fixed: a mesh subscription never silently fails to exist
+
+`Mesh.subscribe` has three `None` returns. One - a raising `declare_subscriber` -
+has always logged a WARNING naming the topic, in line with the six other
+client-side refusals in `strands_robots/mesh/core.py`: `send`, `broadcast`, the
+invalid-command and lockout rejections, the startup declare failure, and
+`declare_publisher`. The other two, the peer not being on the mesh and there
+being no session, answered `None` with no record at all.
+
+Those two are the pair a caller meets on the only rejoin the class offers.
+`stop()` drops every subscription `subscribe` records and clears `inbox`;
+`start()` re-declares only the peer's own built-in topics, and the
+`(topic, callback)` pairs behind a user subscription are not retained anywhere,
+so a consumer that leaves and rejoins after a config change or a hub restart
+re-subscribes. If the mesh has not come back up, that re-subscribe answered
+`None` and said nothing - and `stop()` had not said the subscription was gone in
+the first place, so an empty `_user_subs` was the only evidence either had
+happened.
+
+Both refusals now say why at WARNING, naming the topic and which of the three it
+was, and `stop()` reports at INFO how many subscriptions it dropped and their
+names when there were any, so a rejoining caller is told what it has to
+re-declare. `subscribe`'s docstring gains a `Returns:` block accounting for all
+three refusals and states that a subscription does not survive `stop()`;
+`stop`'s states what it drops and why `start()` cannot restore it, and
+`docs/mesh.md` says the same to a consumer under **Rejoining the mesh**.
+
+Nothing about the round trip changes. Measured end to end against a live Zenoh
+session, the subscription attaches and fires once, `stop()` empties `user_subs`
+and `inbox`, the early re-subscribe answers `None`, and the rejoin restores the
+`peer_id` and leaves an engaged e-stop lockout engaged - identically before and
+after. A network blip was never a way to forget a stop; what it was is a way to
+lose a subscription without being told.
--- a/docs/mesh.md
+++ b/docs/mesh.md
@@ -129,6 +129,30 @@ always agree: both describe the full rotation for every orientation, including
 the half of SO(3) past 120 degrees that a robot turning back the way it came
 lands in.
 
+### Rejoining the mesh
+
+`stop()` then `start()` is how a peer leaves and rejoins - after a config
+change, or a hub that went away and came back. The peer keeps its identity
+across it: the `peer_id` is unchanged, and an engaged e-stop lockout stays
+engaged, so a network blip is not a way to forget a stop.
+
+What does not survive is your own `subscribe()` topics. `start()` re-declares
+the peer's built-in topics from the table above; the subscribers `subscribe()`
+returned are undeclared with the session reference and their callbacks are not
+retained, so a rejoining consumer re-declares its own. `stop()` reports at INFO
+how many it dropped and their names, and `subscribe()` says at WARNING when it
+refuses - naming the topic and whether the peer is off the mesh, has no session,
+or had the declare itself fail - so a rejoin that has not finished is visible
+rather than a silent `None`.
+
+```python
+sim.mesh.stop()
+sim.mesh.start()
+name = sim.mesh.subscribe("strands/*/state", callback=on_state)
+if name is None:
+    ...  # the WARNING says which of the three refusals it was
+```
+
 ## Agent-driven mesh
 
 ```python
--- a/strands_robots/mesh/core.py
+++ b/strands_robots/mesh/core.py
@@ -825,7 +825,15 @@ class Mesh(SensorLoopsMixin):
             logger.info("[mesh] %s on mesh (%s)", self.peer_id, self.peer_type)
 
     def stop(self) -> None:
-        """Stop all loops and release the session reference."""
+        """Stop all loops and release the session reference.
+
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Makes the `stop()`/`start()` rejoin round trip honest about user subscriptions without changing any behaviour: the two previously silent `None` returns in `Mesh.subscribe` (peer not running, no session) now log a WARNING naming the topic and the refusal, matching the pre-existing `declare_subscriber` failure path and the six other client-side refusals in `mesh/core.py`; `stop()` reports at INFO how many `subscribe()` subscriptions it dropped and their names, only when there were any. The docstrings and `docs/mesh.md` document the return contract and the rejoin semantics. Verified against the current head (`fa7bbe6a`, which supersedes `dda2befb` and addresses the CodeQL lambda comments): the extended test file passes 19/19, the repo-wide ASCII/log-string, docstring-xref and changelog-fragment guards pass, log strings are plain ASCII with lazy `%` formatting, and the `user_sub_names` snapshot is taken inside `_subs_lock` and logged outside it, keeping the critical section minimal per the mesh lock discipline. The stop-report is gated on non-emptiness (with a no-noise control test), the accepted path stays quiet, and the peer-id/e-stop-lockout survival across `stop()` is pinned as a control. No public API shape, wire format, or default behaviour changes.

## What's good

- Extended the existing pinning test file rather than adding a new one, so its ten prior cases act as controls; break-matrix in the description shows each new guard fires exactly on its own regression.
- Changelog fragment (`changelog.d/2626-...md`) instead of a direct `CHANGELOG.md` edit, per AGENTS.md step 3.
- Scope discipline: explicitly defers the `reconnect()` verb (a new public surface plus a contract choice) and only makes the existing round trip report itself.
- Refusal log wording names the remedy (`start() first`, `re-declare after start()`), and all three refusals are distinguishable in the message text.

yinsong1986 (approved):
## Summary

This PR makes the `stop()`/`start()` rejoin round trip report itself without changing any behaviour. The two previously silent `None` returns in `Mesh.subscribe` (peer not running, no session) now log a WARNING naming the topic and the specific refusal, matching the pre-existing `declare_subscriber` failure and the other client-side refusals in `mesh/core.py`; `stop()` reports at INFO how many user subscriptions it dropped and their names, only when there were any. The docstrings and `docs/mesh.md` gain the matching contract statements. Verified at head: the diff touches only logging, docstrings, docs, a changelog fragment and tests; the `stop()` report captures `user_sub_names` inside `_subs_lock` and logs after the lock is released, all new log strings are plain ASCII, and the full modified test file passes (19/19) at `fa7bbe6a`. The earlier CodeQL `py/unnecessary-lambda` notes from the `dda2befb` review are resolved at this head - the stand-in session is bound to a local, matching the file's pre-existing pattern.

## What's good

- Changelog fragment under `changelog.d/` per AGENTS.md step 3, no direct `CHANGELOG.md` edit.
- The `stop()` notice is conditional on there being subscriptions to drop, with a dedicated no-noise control test, so the INFO line carries signal.
- Controls pin what must not change: the accepted path stays quiet, the pre-existing declare-failure WARNING is unchanged, and `peer_id` plus an engaged e-stop lockout survive `stop()` - the reporting cannot quietly grow into a reset.
- Scope discipline: a `reconnect()` verb is explicitly deferred with the measurement that justifies deferring it.

PR #2628, opened 2026-08-22
fix(hardware): a leader arm is named as a teleoperator, not listed among the followers

cagataycali:
## Problem

`Robot()`'s registry guard refuses an **unregistered** `*_leader` name and names `Teleoperator()` instead, deliberately *without* listing the registry. Its own comment says why:

> Answering it with the generic registry listing invites the caller to retry with the follower name on the leader's port - the exact mistake that torque-enables the arm a human is holding.

That guard sits behind `get_robot(canonical) is None and not (has_sim(canonical) or has_hardware(canonical)) and not is_discoverable(canonical)`. Registering the leader arm makes every one of those false, so the guard cannot fire.

Registering it is not misuse. `hardware={"lerobot_type": "so101_leader", ...}` is the honest registration for a leader arm, and it satisfies `tests/test_leader_arm_is_not_a_robot.py::test_no_leader_name_resolves_to_a_follower_entry` (a leader name may name a leader device; only naming the *follower* it drives is forbidden). Measured on a registered SO-101 leader:

```
Robot('so101_rig_leader', mode='real')
ValueError: Unsupported robot type: 'so101_leader'. Known lerobot robot types:
  ['bi_openarm_follower', 'bi_rebot_b601_follower', 'bi_so_follower',
   'earthrover_mini_plus', 'hope_jr_arm', 'hope_jr_hand', 'koch_follower',
   'lekiwi', 'lekiwi_client', 'omx_follower', 'openarm_follower', 'reachy2',
   'rebot_b601_follower', 'so100_follower', 'so101_follower', 'unitree_g1']
```

All 16 names are followers. That is the listing the first guard exists to remove, offered by the refusal a real rig actually reaches.

lerobot is not confused about the device: `so101_leader` is one of 9 `*_leader` entries in `TeleoperatorConfig.get_known_choices()` (20 choices). The refusal simply never asked the other registry.

## Consequence

Following each refusal to the driver it directs the caller to build, for the same registered leader on the same physical port:

| tree | offers | a caller who follows it builds | on the leader's port |
|---|---|---|---|
| `upstream/main` | the 16 robot types | `SOFollowerRobotConfig` | torque-drives it |
| this branch | `Teleoperator('so101_leader', port=...)` | `SOLeaderTeleopConfig` | reads it |

## Change

`teleoperator._other_lerobot_kind_refusal(requested, *, wanted=)` is the one owner of the rule: lerobot keeps robots and teleoperators in two `ChoiceRegistries`, a leader and its follower share a servo bus and a USB-serial shape, so a request naming a device of the other kind is a **wrong entry point, not a typo** -  and the names of the kind it is not answer the wrong question. Both refusal sites consult it before listing their own choices:

- `hardware_robot._create_minimal_config` -  a known teleoperator is named as one, pointing at `Teleoperator(...)`, and no robot types are listed (matching the decision `Robot()`'s guard already made).
- `teleoperator._build_teleop_config` -  the mirror image: a known robot is named as one, pointing at `Robot(...)`. Not a safety case (driving a follower is the normal thing) but the same wrong-entry-point answer, so it gets the same treatment.

A name neither registry knows is a genuine typo, and it keeps its listing byte-identical. `wanted` outside `{"robot", "teleoperator"}` raises rather than reporting "not the other kind".

The helper is private and lives in `teleoperator.py` because that module is stdlib-only at import scope; the robot registry is reached by a function-scope import, on a refusal path only. No public surface changes; no accepted call changes.

## Tests

`tests/test_leader_arm_is_not_a_robot.py` is the existing home for this contract -  its module docstring already states the harm -  so the reachability half is pinned beside the registry invariant it slips past. Pre-fix, with the file copied onto `upstream/main`: **10 failed / 21 passed**, 9 of them behavioural (each failure prints the follower list). Post-fix **31 passed**.

The 21 that pass on both trees are the controls, and each fails for a specific wrong fix: an unknown robot type keeps its listing (fails if the cross-kind branch widens into every refusal); an unknown teleoperator type keeps its listing; a valid follower type still builds a config with its `id` intact (the accepted path is untouched); and the pre-existing unregistered-leader and registry-invariant tests are unchanged.

## Verification

Measured at `2ad1abd6` on Linux/py3.12, lerobot 0.6.2.

- **Gate** -  65 test files (every test naming the changed symbols, the hardware/teleoperator/registry suites, and the repo-wide docstring/string/changelog guards), the identical selection run on this branch and on a pristine `upstream/main` worktree with the changed test file excluded from both: **2463 passed, 1 skipped, 0 failed on each**, failing-ID sets identical, 81.1s vs 82.4s.
- `mypy strands_robots tests tests_integ`: **24 errors on this branch and 24 on pristine `upstream/main`**, branch-only set empty, none in the changed files.

## Scope

Issue context: this is the name-based half. A physics-based check (reading each servo's `Present_Voltage` to tell a 7.4V leader pack from a 12V follower supply) is a separate change with open contract questions -  whether a probe given a port string may share a bus with a live robot, and what a voltage verdict's schema is -  so it is deliberately not here.

leader arm named as a teleoperator

### DIFF
--- a/changelog.d/2628-leader-named-as-a-teleoperator.md
+++ b/changelog.d/2628-leader-named-as-a-teleoperator.md
@@ -1,0 +1,26 @@
+### Fixed: a leader arm is named as a teleoperator, not listed among the followers
+
+`Robot()`'s registry guard refuses an unregistered `*_leader` name and names
+`Teleoperator()` instead, deliberately without listing the registry - its own
+comment says a listing "invites the caller to retry with the follower name on the
+leader's port - the exact mistake that torque-enables the arm a human is holding".
+That guard sits behind `get_robot(canonical) is None and not has_hardware(canonical)`,
+so registering the leader arm makes it unreachable.
+
+Registering it is not misuse: `hardware={"lerobot_type": "so101_leader"}` is the
+honest registration for a leader arm, and it satisfies the invariant that a leader
+name may never name the follower it drives. The request fell through to lerobot's
+`RobotConfig` lookup instead, which answered `Unsupported robot type: 'so101_leader'.
+Known lerobot robot types: [...]` - sixteen names, every one a follower. That is the
+listing the first guard exists to remove, offered by the refusal a registered rig
+actually reaches. Following it and retrying with `so101_follower` on the leader's
+port builds an `SOFollowerRobotConfig` bound to the arm a human is holding.
+
+lerobot is not confused about the device: `so101_leader` is one of nine `*_leader`
+entries in `TeleoperatorConfig`. The refusal never asked the other registry. Both
+config sites now consult one shared rule first, so a name that is a device of the
+other kind is named as that kind together with the entry point that builds it -
+`Teleoperator(...)` for a leader handed to `Robot()`, and `Robot(...)` for a follower
+handed to `Teleoperator()`. Neither cross-kind refusal lists the choices of the kind
+that was asked for, because a caller whose name is a known device has not made a
+typo. A name neither registry knows is a typo, and keeps its listing unchanged.
--- a/strands_robots/hardware_robot.py
+++ b/strands_robots/hardware_robot.py
@@ -1079,6 +1079,17 @@ class Robot(TeleopMixin, AgentTool):
         try:
             ConfigClass = RobotConfig.get_choice_class(robot_type)
         except KeyError:
+            # A leader arm is a teleoperator, not a robot, and lerobot keeps the
+            # two in separate registries. Listing the robot types for a leader
+            # name is the retry that torque-enables the arm a human is holding,
+            # so name the kind it really is instead. ``Robot()``'s own registry
+            # guard does this for an unregistered ``*_leader`` name; a leader
+            # that IS registered -- as itself, which is the honest thing to
+            # register -- reaches this site instead.
+            from strands_robots.teleoperator import _other_lerobot_kind_refusal
+
+            if other := _other_lerobot_kind_refusal(robot_type, wanted="robot"):
+                raise ValueError(other) from None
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes the reachability gap in the leader-arm safety refusal: `Robot()`'s existing guard only fires for an *unregistered* `*_leader` name, so an operator who honestly registers their leader (`hardware.lerobot_type="so101_leader"`) fell through to lerobot's `RobotConfig` lookup and got back the full follower listing - exactly the retry the guard exists to suppress. The fix adds one private helper, `teleoperator._other_lerobot_kind_refusal`, consulted by both cross-kind refusal sites (`hardware_robot._create_minimal_config` and `teleoperator._build_teleop_config`) before either lists its own registry: a name the other registry knows is named as that kind with the correct entry point, and a name neither registry knows keeps its listing byte-identical. No accepted call path changes; the helper is private so no public surface is added.

## What's good

- The cross-module imports are both function-scope on refusal paths with documented reasons, so `teleoperator.py` stays stdlib-only at import scope and no module-level cycle is introduced.
- The helper carries a complete `Raises:` block, raises on an unknown `wanted` rather than silently reporting "not the other kind", and both refusal messages are plain ASCII.
- Controls pin the boundary in both directions: unknown names on either side keep their listings, the accepted follower path is untouched, and the pre-existing unregistered-leader tests are unchanged. Verified locally at the PR head: 31/31 in the changed test file and 165 passed / 1 skipped across the teleoperator/hardware-config/robot-factory suites (lerobot 0.6.1).
- Changelog fragment present per AGENTS.md step 3; test names describe behaviour; the `isolated_registry` fixture keeps the probe registration out of the developer's `~/.strands_robots`.

PR #2629, opened 2026-08-22
fix(hardware): the health probe names which camera is down

cagataycali:
## Problem

`Robot.get_status()` is the operator's health probe -- the module's own test says so ("`get_status` is the operator's health probe; a raising serial/USB backend must never propagate out of it"). Every field in it is a fact read off the device at call time, except one. `cameras` was built from `robot.config.cameras` -- the names the device was *asked* for -- while the measured reading sat one attribute away, on `robot.cameras[name].is_connected`.

That matters because of what `is_connected` is on a lerobot arm:

```python
# lerobot src/lerobot/robots/so_follower/so_follower.py:89
return self.bus.is_connected and all(cam.is_connected for cam in self.cameras.values())
```

It collapses N+1 independent facts into one boolean, and 15 lerobot robot classes build it that way. With only the configured enumeration beside it, four physically distinct faults produced **one identical report**. Measured by driving `get_status()` on a two-camera arm shaped like `SOFollower`:

| fault | `is_connected` | `cameras` | `cameras_connected` (before) |
| --- | --- | --- | --- |
| `front_cam` dropped | `False` | `['wrist_cam', 'front_cam']` | field absent |
| both cameras dropped | `False` | `['wrist_cam', 'front_cam']` | field absent |
| motor bus down, cameras fine | `False` | `['wrist_cam', 'front_cam']` | field absent |
| `front_cam` probe raises | `False` | `['wrist_cam', 'front_cam']` | field absent |

**4 distinct faults, 1 distinct report.** An operator reading it learns that something in the set fell, and the field that enumerates the set answers a different question, so it is byte-identical whether every camera is streaming or every one is dead.

## Change

Report each live camera's own `is_connected` reading under `cameras_connected`. After: **4 faults, 4 reports.**

| fault | `is_connected` | `cameras_connected` (after) |
| --- | --- | --- |
| `front_cam` dropped | `False` | `{'wrist_cam': True, 'front_cam': False}` |
| both cameras dropped | `False` | `{'wrist_cam': False, 'front_cam': False}` |
| motor bus down, cameras fine | `False` | `{'wrist_cam': True, 'front_cam': True}` |
| `front_cam` probe raises | `False` | `{'wrist_cam': True}` |

Row 3 is the one worth noting: the probe has no direct reading for the motor bus, so every camera reading `True` under a `False` aggregate isolates the bus by elimination.

Three deliberate boundaries:

- **`cameras` is unchanged, and unfiltered.** It keeps answering "which streams exist" -- a camera that is down is still a stream the device is configured for. The existing `test_get_status_lists_the_devices_configured_cameras` passes untouched, and filtering it would destroy the set difference below.
- **A camera whose probe raises is omitted, not reported as `False`.** Reporting `False` would claim a measurement the device never made and make an unreadable camera indistinguishable from a disconnected one. Omission mirrors the observation path, which already says "a camera whose RTX product hasn't warmed up is omitted rather than failing the whole observation". Because `cameras` stays unfiltered, `set(cameras) - set(cameras_connected)` is exactly the cameras whose state could not be read.
- **The per-camera probe is guarded per camera**, so one bad camera costs its own entry rather than the whole status. The breadth is the camera backend's, not ours: the default `OpenCVCamera` answers through `cv2.VideoCapture.isOpened()`, and `cv2.error`'s MRO is `['error', 'Exception', 'BaseException', 'object']` -- a direct `Exception` subclass -- so the narrower tuple the observation path uses (`RuntimeError, ValueError, AttributeError, TypeError, IndexError`) would let the shipped camera through. The handler mirrors `mujoco/rendering.py`'s per-camera best-effort (`# noqa: BLE001`) and the reason is stated at the site.

## Scope

This is the existing probe reporting the fact it already reaches for; it adds no new method and changes no existing field's shape or meaning. A `readiness()`/`wait_ready()` lifecycle API and a fatal-exit log convention are separate surfaces and separate decisions, not touched here.

## Tests

`tests/test_hardware_robot_lifecycle.py::TestCameraHealthIsAttributable`, 8 cases. Pre-fix on `upstream/main`: **6 failed / 71 passed**; after: **77 passed**.

Two failures are behavioural and name the consequence:

```
Failed: the health probe answered is_connected=False with cameras=['wrist_cam', 'front_cam']
and no per-camera reading, so which of the motor bus and the two cameras fell is not in the report

Failed: one dropped camera and two dropped cameras produce the same report:
is_connected=False, cameras=['wrist_cam', 'front_cam'], cameras_connected=None
```

The other four pin the field's contract shape rather than the regression. Two cases pass on both trees and are the controls: the probe survives a camera that cannot answer, and the configured enumeration still names every camera.

Six breaks, six distinct firing sets:

| break | fails | which |
| --- | --- | --- |
| revert the source | 6 | the two measurements plus the four contract pins |
| probe the config instead of the live camera objects | 3 | partial and total failure report identically again |
| drop the per-camera guard | 3 | the fail-soft control fires -- one bad camera blanks the whole probe (`KeyError: 'cameras'`) |
| a raising probe reports `False` instead of being omitted | 1 | only the omission case |
| filter `cameras` to the connected ones | 4 | including the **pre-existing** `test_get_status_lists_the_devices_configured_cameras` |
| drop the non-mapping guard | 1 | only the unfamiliar-shape case |
| control (restored) | 0 of 77 | |

The fifth row is the one that shows the original field's contract survived rather than being replaced.

## Verification

Measured at `52e82502`, against `upstream/main` `58b9a793`.

- `mypy strands_robots tests tests_integ`: **24 errors on both trees**, none in the changed files, branch-only set empty.
- Blast radius: the 87 test files naming the status surface, `hardware_robot`, `is_connected`, or a repo-wide guard, run with the **identical** selection on both trees with the file this PR modifies excluded from each: **2814 passed, 5 skipped, 0 failed** on the branch and byte-identically on `upstream/main` (171.7s vs 173.4s). No pre-existing failures in the selection.

get_status camera health, before and after

The columns are one capture script run against each tree: the same two-camera arm, the same four faults, and only `strands_robots` different between them. Asserted from the captured payloads rather than read off the picture: the device aggregate `is_connected`, the configured `cameras` list, the healthy and no-camera rows, and the probe's survival of a camera that cannot answer are **identical in both columns** -- what changes is only whether the report names the camera.

### DIFF
--- a/changelog.d/2629-health-probe-names-the-camera.md
+++ b/changelog.d/2629-health-probe-names-the-camera.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- **`Robot.get_status()` now reports each camera's own connection reading, so an unhealthy arm names
+  which stream is down.**
+  The probe's `cameras` field was built from `robot.config.cameras` -- the camera names the device was
+  asked for -- while the measured reading sat one attribute away on `robot.cameras[name].is_connected`.
+  Because `is_connected` on a lerobot arm is `bus and all(cameras)`, it collapses N+1 independent facts
+  into one boolean, and with only the configured enumeration beside it four physically distinct faults
+  produced one identical report: a single dropped camera, every camera dropped, the motor bus down, and
+  a camera whose probe raises all answered `is_connected=False` with the same `cameras` list. A new
+  `cameras_connected` maps each live camera to its own reading, so a camera reading `False` is the
+  culprit and every camera reading `True` leaves the motor bus as the one by elimination. `cameras`
+  keeps answering which streams the device is configured for, unfiltered, so the difference between the
+  two names exactly the cameras whose state could not be read: a camera whose probe raises is omitted
+  rather than reported as `False` -- which would claim a measurement the device never made -- mirroring
+  the observation path, and the rest of the status still arrives.
--- a/strands_robots/hardware_robot.py
+++ b/strands_robots/hardware_robot.py
@@ -2655,17 +2655,64 @@ class Robot(TeleopMixin, AgentTool):
             pass  # Ignore errors in destructor
 
     async def get_status(self) -> dict[str, Any]:
-        """Get robot status including connection and task state."""
+        """Report the device's measured state plus this robot's task state.
+
+        Two fields answer different questions about cameras, and the pair is
+        what makes an unhealthy device attributable:
+
+        * ``cameras`` enumerates the camera names the device is *configured*
+          for - which image streams exist at all.
+        * ``cameras_connected`` maps each live camera to its own
+          ``is_connected`` reading - which of those streams are actually up.
+
+        ``is_connected`` on a lerobot arm is ``bus and all(cameras)``, so a
+        single dropped camera pulls it to ``False`` without naming which of the
+        N+1 facts fell. Reported beside the per-camera readings, a ``False``
+        becomes attributable: a camera reading ``False`` is the culprit, and
+        every camera reading ``True`` leaves the motor bus as the one by
+        elimination.
+
+        Best-effort per camera, mirroring the observation path: a camera whose
+        ``is_connected`` probe raises is omitted from ``cameras_connected``
+        rather than failing the whole probe, so a name present in ``cameras``
+        and absent from ``cameras_connected`` is one whose state could not be
+        read. A device exposing no live camera objects reports an empty map.
+
+        Returns:
+            Status dict of measured facts. An unexpected failure anywhere in
+            the probe degrades to ``{"robot_name", "error", "is_connected":
+            False, "task_status": "error"}`` rather than propagating, so a
+            supervising agent reads a verdict instead of taking an exception.
+        """
         try:
             # Get robot connection status
             is_connected = self.robot.is_connected if hasattr(self.robot, "is_connected") else False
             is_calibrated = self.robot.is_calibrated if hasattr(self.robot, "is_calibrated") else True
 
-            # Get camera status
+            # The configured enumeration: which image streams the device was
+            # asked for. Says nothing about whether any of them came up.
             camera_status = []
             if hasattr(self.robot, "config") and hasattr(self.robot.config, "cameras"):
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Adds a `cameras_connected` field to `Robot.get_status()` that maps each live camera to its own measured `is_connected` reading, making an aggregate `is_connected=False` attributable to a specific camera (or, by elimination, the motor bus) instead of leaving the operator with an undifferentiated set to guess from. The existing `cameras` field keeps its "which streams are configured" contract unfiltered, so the set difference between the two names exactly the cameras whose state could not be read. The per-camera probe is guarded per camera with the breadth justified at the site (`cv2.error` is a direct `Exception` subclass, mirroring the observation path's per-camera best-effort), so one unreadable camera costs its own entry rather than the whole probe.

## What's good

- The eight new tests in `TestCameraHealthIsAttributable` pin behaviour (partial vs total failure distinguishable, bus isolation by elimination, omission-not-guessing for an unreadable probe) with behaviour-named cases, and the mutation table in the description shows each guard has its own firing set.
- The `except Exception` on the per-camera read is a recovery path with the reason stated at the site and a `# noqa: BLE001`, consistent with the `mujoco/rendering.py` per-camera precedent AGENTS.md cites; the failure is logged at debug rather than silently dropped, and the omission is observable via the unfiltered `cameras` enumeration.
- Changelog fragment present (`changelog.d/2629-health-probe-names-the-camera.md`), all strings ASCII, no host paths; verified locally at `52e82502`: the modified lifecycle suite passes (77 passed) and the repo-wide guards (`test_log_strings_are_ascii`, `test_no_host_paths`, `test_test_case_names_describe_behaviour`, `test_except_tuples_state_their_real_scope`) are green.
- Scope discipline: additive field only, no existing field's shape or meaning changes, and the deliberate boundaries (unfiltered `cameras`, omission over a fabricated `False`) are documented in the docstring where the next reader will look.

Issue #322, opened 2026-06-04
robot_mesh: declined-approval response echoed to LLM (side-channel) + read-only action audit gap (inbox/unsubscribe/peers/status)


cagataycali:
## Context

Follow-up bundle from PR #227 (`mesh(tools): robot_mesh HITL via tool_context.interrupt + per-action rate limit`) review feedback. Two related v0.4.1 hardenings to the `robot_mesh` tool that share the same surface area (`strands_robots/tools/robot_mesh.py`) and audit-symmetry theme; bundling so a single PR can land both with one regression-test sweep.

### Issue

When `tool_context.interrupt(...)` returns a non-affirmative response on `emergency_stop` / `broadcast`, the tool's error return currently includes the operator's raw response string in the LLM-facing payload. Audit capture of the literal is the right call (forensic trail), but the LLM-facing return value should not surface it.

### Why it matters

The operator's literal input becomes a side-channel from the operator (out-of-band SDK actor) into the LLM context. A malicious or compromised approval UI could embed prompt-injection text in the decline string, and that text would land in the agent's tool-result and influence the next reasoning step. The HITL gate's whole point is to keep operator authority out-of-band of the LLM; echoing the literal back inverts that.

### Fix shape

In the decline branch (`robot_mesh.py:~389`), keep the audit row carrying the literal response as today, but change the LLM-facing return to a flat sentinel:

```python
_audit_tool_action(action, target, False, f"declined: response={response!r}")  # audit: keep literal
return _err("declined by operator")  # LLM-facing: no echo
```

Pin test: assert that for `response="<inject>...</inject>"`, `_err(...)` payload does NOT contain the literal substring, while `_audit_tool_action` was called with the literal in its message.

### Issue

The `peers`, `status`, `inbox` (~L583-594), and `unsubscribe` (~L597-602) actions do not call `_audit_tool_action`. The omission is symmetric across the four read-only / housekeeping branches and was intentional in PR #227's scope ("safety-significant only"). Reviewer feedback observes that for incident reconstruction ("what did the LLM do at 03:42?") a uniform audit trail across all dispatch branches makes forensic walks meaningfully easier.

### Why it matters

[...]

PR #2630, opened 2026-08-22
fix(mesh): a human's approval is recorded even when the gate then refuses the action

cagataycali:
## Problem

The mesh tool re-checks its rate limit **under the lock after** the operator approves, because the pre-interrupt check deliberately does not consume a slot. So a concurrent invocation can take the last slot while the human is deciding, and the re-check then refuses an action the operator **approved**.

The operator row was written per branch -- `approved=False` in the decline branch, `approved=True` after that re-check -- so the raced refusal returned *between* them and the verdict went unrecorded:

| measured | operator rows | what the audit log carries |
|---|---|---|
| plain approval | 1 | `success=True  "operator approved: 'y'"` |
| approval, slot taken while deciding | **0** | `success=False "rate_limit_race: rate limit exceeded for action 'tell' between check and record ..."` |

A human authorised a physical actuation and the trail has no row saying so. The one surviving row attributes the refusal to a rate limit and says nothing about who authorised it, so an incident audit asking "did a human approve this?" finds no answer -- on the one gate that reaches physical actuation (`tell`, `send`, `stop`, `broadcast`, `emergency_stop`, `rpc`).

`log_operator_response`'s own docstring already asks for the opposite:

> Call this on BOTH outcomes. A decline is the row that carries the operator's reason; an approval is the record that a human authorised an agent to reach a physical surface, **which is the question an audit of an incident asks first.**

And the two sibling gates already had the right shape: `use_ros._gate_command` and `lerobot_train._gate_extra_flags` both compute the verdict and record it **unconditionally**, then branch. `robot_mesh` was the one gate of three that recorded per branch.

## Change

Record the verdict once, as soon as it is known, before any later refusal can return -- the shape the other two gates already used:

```python
approved = _interrupt_approves(response)
log_operator_response(_AUDIT_SOURCE, action, target, approved=approved, response=response)
if not approved:
    return _err(f"action '{action}' was declined by the operator interrupt.")
rl_race_err = _rate_limit_check_and_record(action)
if rl_race_err is not None:
    _audit_tool_action(action, target, False, f"rate_limit_race: {rl_race_err}")
    return _err(rl_race_err)
```

The raced path now leaves **two rows, two facts**: `operator approved: 'y'` (a human authorised it) and `rate_limit_race: ...` (why it was refused anyway).

`log_operator_response`'s docstring gains the requirement the fix satisfies -- one site, reached before any later refusal can return -- so a fourth gate is told the shape rather than inferring it from a sibling.

Nothing else moves: the grammar, the model-visible text, and the "a decline does not consume a rate-limit slot" contract are untouched.

## Tests

`2 failed / 29 passed` against `upstream/main`, both failures behavioural or structural:

```
FAILED ...::test_the_human_verdict_is_recorded_even_though_the_action_was_refused
  a human approved a physical actuation and the audit log has no row saying so;
  rows a forensic reader would find: [(... 'success': False,
  'detail': "rate_limit_race: rate limit exceeded for action 'tell' ...")]
FAILED ...::test_each_gate_records_from_one_site_no_later_refusal_can_return_past[robot_mesh]
  robot_mesh records the operator's verdict at 2 sites [252, 241]; want one that every
  path reaches once the verdict is known, so a later refusal cannot return past all of them
```

The behavioural test drives the real race deterministically without threads: the stand-in `interrupt()` drains the bucket while the gate is blocked on the human, which is exactly the concurrency the re-check exists for. A premise assert fails loudly if no slot was taken.

The structural test is the root cause, and it is derived rather than listed -- the graded set comes from the tree's `interrupt()` call sites, so a fourth gate is held to the rule on arrival. Two clauses, measured complementary:

| break | fires |
|---|---|
| control (this branch) | 0 of 31 |
| exact pre-fix source | **2** -- the behavioural race test + the `one site` clause |
| one site, but placed after the race check | **6** -- incl. **4 pre-existing** tests (the decline row, its verb/target, the echo contract, the byte-identical payload) + the `no return between` clause |
| same, with the `no return between` clause disabled | **5** -- the structural test now passes, so both clauses carry weight |

The third row is the no-overreach evidence: four tests that already pinned the decline row pass on this branch and fire the moment it is lost, so the reorder preserved the contract they pin. The `use_ros` and `lerobot_train` parameters of the structural rule pass on both trees -- they already had the shape.

## Scope

The **grammar** is deliberately unchanged. A non-string reply (including `True` and `{"approve": True}`) still fails closed, because that is a tested, documented posture rather than drift:

- `_interrupt_approves`' docstring: "We accept the canonical short forms only - defence in depth against accidental approval from a typo."
- `tests/mesh/test_robot_mesh_hitl_approval_contract.py::test_non_string_payloads_never_approve` names `{"approve": True}` and `True` explicitly as cases that must not approve, and a dispatch-level test pins that they return a `declined` error without reaching the mesh.
- `docs/security.md`: "Only `y` / `yes` / `approve` / `approved` count as approval; anything else (including an empty response) is treated as a decline."

Widening it would invert that safety test, which is a contract decision and not this PR's. This PR only fixes the accounting on the path where the operator *did* approve in the accepted grammar.

## Verification

Measured at `4bfaf7b0`, on the identical 274-file selection (every test naming the changed symbols, the three HITL
gates, or a repo-wide guard) run on both trees with the changed test file excluded from each:

| tree | result |
|---|---|
| this branch | `25 failed, 8929 passed, 14 skipped, 24 errors` |
| `upstream/main` | `25 failed, 8922 passed, 14 skipped, 24 errors` |

49 failing ids, the **same set on both trees** (`comm` empty in both directions), and `+7 passed` is exactly the
seven items this PR adds (four tests plus the structural rule's three gate parameters). Every one of the 49 is a
dependency absent from this machine and installed by CI's `.[all,dev]`: 45 `tests/mesh/test_iot_*` on
`No module named 'awsiot'` (the `[mesh-iot]` extra), 2 `tests/test_device_connect_hardening.py` on
`device_connect_agent_tools`, and 2 `tests/training/test_lerobot.py::TestInProcessTrainerCorrectness` on
`draccus 0.10.0` against lerobot 0.6.2's `draccus>=0.11.6` floor (`encode() takes 1 positional argument but 2 were
given`).

`ruff check` and `ruff format --check` are clean over `strands_robots tests tests_integ`; `mypy` reports the same
24 pre-existing errors on both trees, none in the changed files.

## Artifact

The change is to what an auditor reads, so the artifact is the audit log itself. One script per tree drives the
real gate twice against a real `STRANDS_MESH_AUDIT_DIR` -- once with nothing racing, once with a concurrent call
taking the last slot while the operator decides -- and reads the rows back with `read_audit_log`:

HITL approval accounting, before and after

Both columns run the same session and reach the same verdict: `status="error"`, the byte-identical rate-limit text,
the action dispatched 0 times, and 30 slots taken while the operator was deciding. The unraced approval records the
identical row in both trees. The only difference is that the raced path now carries the row saying a human
authorised it.

## Review rounds

**Round 1** - `c39b713`, one concern: a unit test in this PR reached the real audit writer.

`test_the_model_is_told_the_rate_limit_refused_it_not_the_operator` took neither the `audit_rows`
fixture nor an audit directory, so it drove the real gate onto the developer's own trail. Reproduced
under a scratch `HOME` before fixing: one run appended `operator approved: 'y'` for action `tell` with
`success: true` to `~/.strands_robots/mesh_audit.jsonl`, advanced the signed `mesh_audit.seq.json`
sidecar and left a lockfile. A fabricated approval row is indistinguishable from a genuine human
authorisation to exactly the reader this PR exists to serve, and it accumulates per run.

Fixed by naming the audit directory **autouse at module scope** rather than in the one offending test.
The hole is structural: capturing `log_safety_event` is a choice each test makes, so a gate-driving test
added later re-grows the contamination with nothing failing. An AST sweep found two tests taking
neither -- this one, and `test_the_table_covers_every_interrupt_site`, a pure `interrupt()` call-site
scan that drives no gate. `test_a_declined_use_ros_publish_lands_on_disk` is unaffected; its own
`setenv` runs after the fixture and wins.

Pinned both directions by `TestTheAuditTrailATestWritesIsItsOwn`, which fails without the redirect and
names the real trail when it does. Read as a delta, since this box lacks several optional deps:

| tree | failed | passed | files written under `HOME` |
|---|---|---|---|
| `4bfaf7b0` | 62 | 2053 | 3, including the fabricated approval row |
| `c39b713` | 62 | 2055 | none |

The 62 are pre-existing and identical on both sides; the whole effect is +2 (the new pins) and the
pollution gone. Lint clean on the file.

This round also recorded a dependency nothing was checking: AGENTS.md item 16 exempts `tests/` from
whole-log attestation *because* a test redirects the audit dir, so that exemption rests on a property
no test asserted. The new class states it in its docstring.

### DIFF
--- a/tests/tools/test_hitl_operator_response_audit.py
+++ b/tests/tools/test_hitl_operator_response_audit.py
@@ -33,7 +33,7 @@ import strands_robots
 import strands_robots.tools.lerobot_train as train_mod
 import strands_robots.tools.robot_mesh as mesh_mod
 import strands_robots.tools.use_ros as ros_mod
-from strands_robots.mesh.audit import read_audit_log
+from strands_robots.mesh.audit import audit_log_path, read_audit_log
 
 # A reply that carries a reason. Every gate accepts a canonical affirmative only,
 # so this is always a decline - which is exactly why the audit row is the only
@@ -143,6 +143,29 @@ def audit_rows(monkeypatch: pytest.MonkeyPatch) -> list[tuple[str, str, dict[str
 
     monkeypatch.setattr("strands_robots.mesh.audit.log_safety_event", _record)
     return recorded
+
+
+@pytest.fixture(autouse=True)
+def audit_dir_is_this_tests_own(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> None:
+    """Keep every gate driven in this file off the developer's real audit trail.
+
+    ``audit_log_path()`` falls back to ``~/.strands_robots`` when
+    ``STRANDS_MESH_AUDIT_DIR`` is unset, so a test that drives a real gate
+    reaches the real writer. Measured before this fixture existed: one run of
+    the rate-limit-race control appended an ``operator approved: 'y'`` row for
+    action ``tell`` with ``success: true`` to the developer's own
+    ``mesh_audit.jsonl`` and advanced the signed ``mesh_audit.seq.json``
+    sidecar beside it. That row is indistinguishable from a genuine human
+    authorisation to exactly the incident reader this file exists to serve, and
+    it accumulates on every local run.
+
+    Autouse rather than one line in the offending test, because the hole is
+    structural: capturing ``log_safety_event`` is a choice each test makes, so a
+    gate-driving test added later re-grows the same contamination in silence and
+    nothing fails. A test that wants a directory it can read back still names one
+    itself - its own ``setenv`` runs after this one and wins.
+    """
+    monkeypatch.setenv("STRANDS_MESH_AUDIT_DIR", str(tmp_path / "audit"))
 
 
 class TestEveryGateRecordsTheOperatorReply:
@@ -463,3 +486,33 @@ class TestAnApprovedActionRefusedForAnotherReasonStillRecordsTheHuman:
                 },
             )
         ]
+
+
+class TestTheAuditTrailATestWritesIsItsOwn:
+    """A gate driven by this suite must not write where an incident audit reads.
+
+    AGENTS.md item 16 exempts ``tests/`` from whole-log attestation *because* a
+    test redirects the audit directory. That exemption is only sound while the
+    redirect actually holds, so it is pinned here rather than assumed.
+    """
+
+    def test_the_resolved_audit_log_is_not_the_real_home(self, tmp_path: Path) -> None:
+        """The fallback in ``audit_log_path()`` is never the path under test."""
+        resolved = audit_log_path()
+
+        assert resolved.parent != Path.home() / ".strands_robots", (
+            "a gate driven by this suite would write to the developer's real audit trail at "
+            f"{resolved}, where a fabricated 'operator approved' row is indistinguishable "
+            "from a genuine human authorisation"
+        )
+        assert tmp_path in resolved.parents, f"the audit log is not this test's own: {resolved}"
+
+    def test_driving_a_real_gate_writes_only_under_this_tests_directory(self, tmp_path: Path) -> None:
+        """The row the rate-limit-race control used to plant lands in tmp_path."""
+        _drive_robot_mesh_racing_the_rate_limit(APPROVAL)
+
+        details = [str(rec.get("payload", {}).get("detail", "")) for rec in read_audit_log(since=0)]
+        assert any(detail.startswith("operator approved") for detail in details), (
+            f"premise: driving the gate left no operator row to locate: {details}"
+        )
+        assert tmp_path in audit_log_path().parents, "the operator row was written outside this test"
--- a/tests/tools/test_hitl_operator_response_audit.py
+++ b/tests/tools/test_hitl_operator_response_audit.py
@@ -272,6 +272,50 @@ class TestTheGatedSetIsDerivedFromTheInterruptSites:
             f"{gate.label}'s gate does not record the operator's reply through the shared owner"
         )
 
+    @pytest.mark.parametrize("gate", _GATES, ids=_GATE_IDS)
+    def test_each_gate_records_from_one_site_no_later_refusal_can_return_past(self, gate: _Gate) -> None:
+        """One site, reached as soon as the verdict is known.
+
+        Two per-branch sites are what let a third exit skip both: the mesh tool
+        recorded ``approved=False`` in its decline branch and ``approved=True``
+        after its post-approval rate-limit re-check, so the raced refusal between
+        them returned with the operator's verdict unrecorded. A single site the
+        gate reaches before it can return again cannot be bypassed that way, and
+        it is the shape the other two gates already had.
+        """
+        fn_src = textwrap.dedent(inspect.getsource(getattr(gate.module, gate.function)))
+        fn = next(
+            node
+            for node in ast.walk(ast.parse(fn_src))
+            if isinstance(node, ast.FunctionDef | ast.AsyncFunctionDef) and node.name == gate.function
+        )
+        calls = self._own_calls(fn)
+
+        sites = [call.lineno for call in calls if ast.unparse(call.func) == "log_operator_response"]
+        assert len(sites) == 1, (
+            f"{gate.label} records the operator's verdict at {len(sites)} sites {sites}; want one that every "
+            "path reaches once the verdict is known, so a later refusal cannot return past all of them"
+        )
+
+        verdicts = [
+            call
+            for call in calls
+            if "approve" in ast.unparse(call.func) and ast.unparse(call.func) != "log_operator_response"
+        ]
+        assert len(verdicts) == 1, (
+            f"{gate.label} computes the operator verdict at {len(verdicts)} sites; the rule below reads the first"
+        )
+
+        escapes = [
+            node.lineno
+            for node in ast.walk(fn)
+            if isinstance(node, ast.Return) and verdicts[0].lineno < node.lineno < sites[0]
+        ]
+        assert not escapes, (
+            f"{gate.label} can return at lines {escapes} between computing the operator's verdict and recording "
+            "it, so that path leaves no row saying whether a human authorised the action"
+        )
+
 
 class TestTheMeshToolRowIsUnchanged:
     """Control: the tool that already recorded keeps writing the same row."""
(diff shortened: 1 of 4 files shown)


yinsong1986 (dismissed):
## Summary

Moves the mesh tool's `log_operator_response` call to a single unconditional site immediately after the operator verdict is computed, so the raced path (a concurrent invocation consumes the last rate-limit slot while the human is deciding) no longer returns between the two per-branch recording sites with the approval unrecorded. The raced refusal now leaves two audit rows -- the operator's approval and the `rate_limit_race` refusal -- matching the shape `use_ros._gate_command` and `lerobot_train._gate_extra_flags` already had, and `_hitl_audit.log_operator_response`'s docstring now states the one-site-before-any-refusal requirement so a fourth gate is told the shape.

## What's good

- The safety contracts around the moved call are preserved and verified: a decline still consumes no rate-limit slot, the operator's literal reply still never reaches the model (the #322 side-channel comment moved with the code), and the decline row is byte-identical -- the pre-existing contract tests pass on this branch and the PR description shows they fire if the reorder loses any of them.
- The behavioural regression test drives the actual race deterministically (draining the bucket from inside the `interrupt()` stand-in) with a premise assert that fails loudly if nothing raced, and the structural rule derives its graded set from the tree's `interrupt()` call sites per AGENTS.md's derived-population convention, so a future gate is held to the rule on arrival.
- Scope discipline: the approval grammar, model-visible text, and rate-limit semantics are untouched; the diff is exactly the accounting fix plus its pins, and the follow-up commit adds the `changelog.d/2630-*` fragment per AGENTS.md step 3 (validated clean against `assemble_changelog.validate_fragment`). Ran the changed test file (31 passed), the HITL approval-contract suite, and the robot_mesh tool/config/validation/audit-parity suites (87 passed) against this head; ruff check and format are clean on all three files.

tests/tools/test_hitl_operator_response_audit.py
```diff
@@ -312,3 +356,110 @@ def _boom(*_a: object, **_k: object) -> None:
...
+        self, audit_rows: list[tuple[str, str, dict[str, Any]]]
+    ) -> None:
+        """Two rows, two facts: a human authorised it, and it was refused anyway."""
+        _drive_robot_mesh_racing_the_rate_limit(APPROVAL)
+
+        raced = [p for _e, _s, p in audit_rows if "rate_limit_race" in str(p.get("detail", ""))]
+        assert len(raced) == 1, f"the refusal reason is not recorded: {audit_rows}"
+        assert raced[0]["success"] is False
+
+    def test_the_model_is_told_the_rate_limit_refused_it_not_the_operator(self) -> None:
```
yinsong1986, line 462: [MUST FIX] `test_the_model_is_told_the_rate_limit_refused_it_not_the_operator` is the one test in this file that neither takes the `audit_rows` fixture (which captures `log_safety_event`) nor sets `STRANDS_MESH_AUDIT_DIR`, and it drives the real gate -- so `log_operator_response` and `_audit_tool_action` reach the real writer. Measured on this head: one run of this test appends two rows to the developer's real `~/.strands_robots/mesh_audit.jsonl` (`"detail": "operator approved: 'y'"` for action `tell`, plus the `rate_limit_race` row) and advances the signed `mesh_audit.seq.json` sidecar.

That is not just test-isolation hygiene: this PR's whole point is that the audit log is the forensic record of whether a human authorised a physical actuation, and this test permanently plants a fabricated `operator approved` actuation record in the real trail on every local run of the unit suite -- exactly the artifact an incident audit would later read as a genuine approval (AGENTS.md item 16 documents that `tests/` is exempt from whole-log attestation *because* a test redirects the audit dir; this one doesn't). Every sibling test in this file already handles it: `audit_rows` captures the call, `TestAnUnwritableAuditLogDoesNotChangeTheVerdict` patches it, and `test_a_declined_use_ros_publish_lands_on_disk` sets `STRANDS_MESH_AUDIT_DIR` to `tmp_path`.

Fix is one line: take the `audit_rows` fixture (its capture is a side effect; the unused-argument name can be `_audit_rows`-style or asserted on), or `monkeypatch.setenv("STRANDS_MESH_AUDIT_DIR", str(tmp_path))` as the on-disk sibling does. Can't be deferred because it contaminates a real security audit artifact from the first local run after merge, and the pollution is silent and cumulative.

yinsong1986 (commented):
## Summary

Moves the mesh tool's operator-verdict recording (`log_operator_response`) to a single unconditional site immediately after `_interrupt_approves` computes the verdict, so the raced path -- a concurrent invocation takes the last rate-limit slot while the human is deciding, and the post-approval re-check refuses an action the operator approved -- can no longer return between the two per-branch recording sites with the approval unrecorded. This is the shape `use_ros._gate_command` and `lerobot_train._gate_extra_flags` already had, and the change is behaviour-preserving on every other path: the decline row is byte-identical, a decline still consumes no rate-limit slot, `log_operator_response` swallows audit-write failures so the moved call cannot change a gate's verdict, and the #322 no-echo contract moved with the code. The behavioural race test and the derived structural rule are well built. One must-fix in the new test file: a test that drives the real gate without capturing or redirecting the audit writer appends forged approval rows to the developer's real `~/.strands_robots/mesh_audit.jsonl` on every run (measured; inline comment).

tests/tools/test_hitl_operator_response_audit.py
cagataycali: Confirmed and fixed at `c39b713`, which is the current head.

Reproduced first, because the fix depends on which artifacts are touched. Running that one test under a scratch `HOME` planted exactly what you describe:

```
/tmp/fakehome/.strands_robots/mesh_audit.jsonl
/tmp/fakehome/.strands_robots/mesh_audit.seq.json
/tmp/fakehome/.strands_robots/mesh_audit.seq.lock

1 | llm_tool_action | {"action": "tell", "detail": "operator approved: 'y'", "success": true, "target": "peer-a"}
2 | llm_tool_action | {"action": "tell", "detail": "rate_limit_race: ..."}
```

So it is the `success: true` / `operator approved` pair, plus the signed sidecar and a lockfile.

I took your second option -- naming the directory -- but applied it **autouse at module scope** rather than in the one test, and that is the only place I departed from what you asked. The reason is that the hole is structural rather than local: capturing `log_safety_event` is a choice each test makes, so the invariant "nothing in this file reaches the real writer" is re-established per test and a gate-driving test added later re-grows the contamination with nothing failing. An AST sweep of the file found two tests taking neither the fixture nor the variable -- this one, and `test_the_table_covers_every_interrupt_site`, which is a pure `interrupt()` call-site scan and drives no gate, so your count of one gate-driving offender is exact. `test_a_declined_use_ros_publish_lands_on_disk` is unaffected: its own `setenv` runs after the fixture and wins, which is what keeps a test that wants to read the log back working.

Pinned both directions, since a redirect that silently stops holding is the failure mode. `TestTheAuditTrailATestWritesIsItsOwn` asserts the resolved path is under `tmp_path` and is not `~/.strands_robots`, and drives the racing gate to locate the row it used to plant. With the fixture disabled both fail, naming the real trail:

```
E  AssertionError: a gate driven by this suite would write to the developer's real audit trail at
   /tmp/fh2/.strands_robots/mesh_audit.jsonl, where a fabricated 'operator approved' row is
   indistinguishable from a genuine human authorisation
2 failed
```

Verified as a delta rather than an absolute, since this box lacks several optional deps:

| tree | failed | passed | files written under `HOME` |
|---|---|---|---|
| `4bfaf7b0` | 62 | 2053 | 3, including the fabricated approval row |
| `c39b713` | 62 | 2055 | none |

The 62 are pre-existing collection/environment failures present on both sides; the whole effect is +2 (the new pins) and the pollution gone. `ruff check` and `ruff format --check` clean on the file.

One thing your note taught me that I have not seen written down: AGENTS.md item 16 exempts `tests/` from whole-log attestation *because* a test redirects the audit dir, so that exemption is load-bearing on a property nothing was checking. The new class says so in its docstring, so the next reader finds the dependency rather than rediscovering it.

yinsong1986 (approved):
## Summary

Moves the mesh tool's operator-verdict recording to a single unconditional `log_operator_response` site immediately after `_interrupt_approves`, so the post-approval rate-limit re-check (which can refuse an action the operator approved, when a concurrent invocation takes the last slot during the interrupt) can no longer return between the two per-branch recording sites with the approval unrecorded. The raced path now leaves two rows for two facts -- `operator approved: 'y'` and `rate_limit_race: ...` -- matching the shape `use_ros._gate_command` and `lerobot_train._gate_extra_flags` already had. Verified locally at `c39b7137`: all 33 tests in the changed file pass, and the new behavioural race test plus the structural one-site rule fail on pre-fix `upstream/main` exactly as the description claims, so neither is vacuous. The round-1 concern (a gate-driving test reaching the developer's real audit trail) is resolved at this head by the module-scope autouse `STRANDS_MESH_AUDIT_DIR` redirect, and a run under a scratch `HOME` confirms nothing is written outside the test's own directory on either tree.

## What's good

- The structural guard derives its graded set from the tree's `interrupt()` call sites, so a fourth gate added later is held to the one-site rule on arrival -- the AGENTS.md "derive, don't list" pattern.
- Scope discipline: the approval grammar, the model-visible refusal text, the no-echo contract (#322), and the "a decline does not consume a rate-limit slot" contract are all provably untouched, each pinned by a control test that passes on both trees.
- Changelog fragment per AGENTS.md step 3; the `log_operator_response` docstring now states the one-site requirement so the shape is documented rather than inferred from a sibling gate.
- `TestTheAuditTrailATestWritesIsItsOwn` turns AGENTS.md item 16's tests-are-exempt-because-they-redirect assumption into an asserted property, closing the structural hole rather than the one offending test.

PR #2631, opened 2026-08-22
fix(hardware): a missing required parameter is refused by name, with the bus

cagataycali:
## Problem

Eight of lerobot's sixteen robot types declare `port` with no default, so forgetting it is the commonest caller mistake on the hardware path. `_create_minimal_config` let the resolved dataclass raise and wrapped whatever it said:

```
ValueError: Failed to construct SOFollowerRobotConfig for robot type 'so101_follower':
SOFollowerRobotConfig.__init__() missing 1 required positional argument: 'port'.
Config: {'id': 'so101', 'cameras': {}}
```

That names `SOFollowerRobotConfig.__init__()`, a lerobot internal. It does not name `port=` as the thing the caller supplies, and it does not name a single device on the host - even though `mode="auto"` had enumerated the candidate ports one frame up, logged them, and returned only the string `"real"`. `Robot('so101', mode='auto')` and `Robot('so101', mode='real')` produced the identical message, so the probe's work was invisible.

The mirror-image mistake was already reported properly. A kwarg the dataclass does not declare is refused by name with the accepted field list, and `_create_minimal_config`'s own docstring says why: it "catches typos like `prot=` (instead of `port=`) at config-build time rather than as a delayed connection failure with no kwarg in sight." A required kwarg the caller omitted is the same class of mistake and got the third-party internal.

`tests/test_robot_factory.py::test_config_construction_failure_wrapped_as_value_error` already used a `_RequiresArg` config with a no-default field, and its docstring asks for a refusal "not a raw TypeError leaking the dataclass internals" - while the wrapper it pins interpolates exactly that TypeError.

## Change

A required field no forwarded kwarg supplied is detected before construction and refused by name. The prefix and the `Config: {...}` tail are unchanged, so the wrapper's existing contract (name the class, name the assembled values) still holds:

```
ValueError: Failed to construct SOFollowerRobotConfig for robot type 'so101_follower':
missing required parameter(s) ['port'], which the caller supplies -- e.g.
Robot('so101', mode='real', port=...). Candidate servo-bus device(s) on this host:
/dev/ttyACM0 (usb id 5AB0181806), /dev/ttyACM3 (usb id 5AB0158428). A port path is a
position on the bus, not an identity: it can change when the device is replugged,
while the usb id does not. Config: {'id': 'so101', 'cameras': {}}
```

The USB serial number is named beside the path because the path is what moves: unplug an arm and plug it back in and the kernel may hand it a different number, and plug a second arm in first and it may take the number the first one had. `pyserial` already reports the serial number; nothing was reading it.

The serial scan answers a missing `port` only. `lekiwi_client` requires `remote_ip`, and naming this host's serial devices there would point a network robot's caller at the wrong bus entirely.

Which serial device is a robot's motor bus is now decided once, in `strands_robots/_serial_discovery.py`. `_auto_detect_mode` had that rule inline - the keyword list, the WCH CH34x / FTDI vendor ids, the exclusions - and the refusal needs the same answer, so a second copy would let the two disagree about what a robot is. Enumeration there is best-effort: `pyserial` is not a declared dependency of this package (it arrives with lerobot), so an absent import or a libusb hub glitch reports no devices rather than replacing a precise refusal with an `ImportError`.

## Scope

Refusing by name is decision-free: the same values reach the same dataclass either way, so this changes which sentence a refused call gets, not which calls are refused.

Deliberately not included, because each is a public-surface decision this does not need to make: a public `list_serial_devices()`, a `Robot(stable_id=)` parameter that resolves a serial number to its current port at connect time, and any on-disk `stable_id -> last_port` persistence. The rule and the identity are now in one place should any of those be wanted later.

## Verification

Both trees run the identical selection with this PR's new test file excluded, so any difference would be the source change alone.

| --- | --- | --- | --- |
| this branch | 0 | 1288 | 1 |
| `upstream/main` | 0 | 1287 | 1 |

The selection is 33 files: every test naming `_create_minimal_config`, `_auto_detect_mode`, the port enumeration or the construct-failure text, plus the repo-wide guards. Zero failures on either side, and the `comm` of the failing-id sets is empty in both directions. The one extra passing item on this branch is `test_attributes_docstring_completeness.py::test_every_field_has_an_attributes_entry[strands_robots/_serial_discovery.py::SerialCandidate]`, which that guard parametrized over the new dataclass on its own.

Pre-fix split of the new file against `upstream/main`: **27 failed / 8 passed** -> 35 passed. Eleven of the 27 are behavioural or structural (nine drive the real refusal, two pin that both call sites consult the owner); the other sixteen report the new module's absence, which is weaker evidence and is labelled as such. The 8 that pass on `upstream/main` are the controls: a supplied port still builds, a `__post_init__` refusal still routes to the generic wrapper, the refusal still names the config class and the assembled values, the serial scan still does not answer a missing `remote_ip`, and all four `_auto_detect_mode` verdicts are unchanged.

Every guard is load-bearing - each break fires exactly its own set:

| break | tests that fail |
| --- | --- |
| revert the hardware refusal | 9 (the refusal tests) |
| drop the port-shaped gate, always append the bus | 1 (only the `remote_ip` control) |
| revert `robot.py` to its own inline vendor-id table | 2 (one-owner guard, mode-probe pin) |
| make the scan raise instead of reporting nothing | 3 (the best-effort tests) |
| make `matches_servo_bus` recognise everything | 5 (rule tests, uart verdict, "not a servo bus") |
| drop the `vid:pid` identity fallback | 2 (the fallback tests) |
| control | 0 of 35 |

`mypy strands_robots tests tests_integ`: 24 errors on both trees, none branch-only. `ruff check` and `ruff format --check` clean on CI's scope. 753 repo-wide guard tests pass (docstring xrefs, ASCII, `Args:`/`Attributes:`/`Raises:` completeness, exception clauses, exports, dependency audit, changelog).

## Artifact

`Robot('so101', mode='real')` with two SO-101 arms attached, run against each tree. The capture script reads whichever refusal its own tree produces, extracts the device paths that refusal names, and retries with the first one - so the bottom row is what a caller who obeys the message actually gets, with nothing hardcoded per column.

the refusal a caller who forgot port= is given, per tree

On `upstream/main` the refusal names no device, so there is nothing to retry with. On this branch it names `/dev/ttyACM0` and `/dev/ttyACM3` with their USB serial numbers `5AB0181806` and `5AB0158428`, and the retry builds the config. The controls are identical on both trees: `mode="auto"` reaches `real` either way, and a call that supplies `port=` builds either way.

### DIFF
--- a/changelog.d/2631-missing-hardware-parameter-names-the-bus.md
+++ b/changelog.d/2631-missing-hardware-parameter-names-the-bus.md
@@ -1,0 +1,30 @@
+### Fixed: a missing required hardware parameter is refused by name, with the bus
+
+Eight of lerobot's sixteen robot types declare `port` with no default, so
+forgetting it is the commonest caller mistake on the hardware path.
+`_create_minimal_config` let the resolved dataclass raise and wrapped whatever it
+said, so the refusal read `SOFollowerRobotConfig.__init__() missing 1 required
+positional argument: 'port'` -- a lerobot internal, naming neither the parameter
+the caller supplies nor a single device on the host. `mode="auto"` had already
+enumerated the candidate ports one frame up, logged them, and returned only the
+string `"real"`, so `mode="auto"` and `mode="real"` produced the identical message
+and the probe's work was invisible.
+
+The mirror-image mistake was already reported properly: a kwarg the dataclass
+does not declare is refused with the accepted field names, and the docstring says
+why -- it catches a typo like `prot=` at config-build time rather than as a
+delayed connection failure with no kwarg in sight. Both halves of "the caller got
+a kwarg wrong" now report the same way. A missing port additionally names this
+host's servo-bus candidates and their USB serial numbers, because a port path is
+a position on the bus that the kernel may renumber across a replug while the
+serial number does not move. The scan answers a missing `port` only: naming
+serial devices for a missing `remote_ip` would point a network robot's caller at
+the wrong bus entirely.
+
+Which serial device is a robot's motor bus is now decided once, in
+`strands_robots._serial_discovery`. That rule was inline in `_auto_detect_mode`
+-- the keyword list, the WCH CH34x / FTDI vendor ids, the exclusions -- and the
+refusal needs the same answer, so a second copy would let the two disagree about
+what a robot is. Enumeration there is best-effort, because pyserial is not a
+declared dependency of this package: an absent import or a libusb hub glitch
+reports no devices rather than replacing a precise refusal with an `ImportError`.
--- a/strands_robots/_serial_discovery.py
+++ b/strands_robots/_serial_discovery.py
@@ -1,0 +1,197 @@
+"""Sole owner of the servo-bus recognition rule.
+
+A USB port path -- ``/dev/ttyACM0`` on Linux, ``/dev/cu.usbmodem58FA0826181`` on
+macOS -- is a POSITION on the bus, not an identity. Unplug an arm and plug it
+back in and the kernel may hand it a different number; plug a second arm in
+first and it may take the number the first one had. The USB serial number does
+not move, so a report that names a port should name the serial number beside it.
+
+Two callers in different layers need the same answer to "which of this host's
+serial devices look like a robot's motor bus":
+
+- :func:`strands_robots.robot._auto_detect_mode`, deciding sim vs real for
+  ``mode="auto"``.
+- :meth:`strands_robots.hardware_robot.Robot._create_minimal_config`, naming the
+  candidates when the caller did not supply the ``port`` its robot requires.
+
+:mod:`strands_robots.robot` defers importing :mod:`strands_robots.hardware_robot`
+(it is the heavy hardware layer, and a sim-only caller must not pay for it), so
+neither can own the rule for the other. It lives here, once, as
+:data:`SERVO_BUS_VIDS` plus :func:`matches_servo_bus`.
+
+``pyserial`` is not a declared dependency of this package -- it arrives with
+lerobot -- so enumeration is best-effort throughout: :func:`scan_serial_devices`
+answers with an empty list rather than raising, and every caller treats "no
+devices" and "cannot enumerate" the same way.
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Replaces the leaked lerobot-internal `TypeError` (`SOFollowerRobotConfig.__init__() missing 1 required positional argument: 'port'`) with a pre-construction refusal that names the missing parameter, the call that supplies it, and -- for a port-shaped parameter only -- the host's servo-bus candidates with their USB serial numbers. The servo-bus recognition rule (vendor ids, keywords, exclusions) moves out of `_auto_detect_mode` into a single private owner, `strands_robots/_serial_discovery.py`, which both call sites consult; enumeration stays best-effort (pyserial arrives via lerobot, so an absent import or hub glitch reports no devices rather than raising). The gate is decision-free -- I verified against installed lerobot 0.6.1 that every no-default field on a registered `RobotConfig` subclass is `init=True`, so the same calls are refused either way and only the sentence changes. Spot-verified in an isolated worktree at the head SHA: the 35 new tests pass, and the 211 pre-existing tests touching `_create_minimal_config` / `_auto_detect_mode` (`test_robot_factory.py`, `test_hardware_servo_safety_clamp.py`, `test_hardware_robot_camera_config.py`, `test_hardware_robot_config.py`) pass unchanged.

## What's good

- The `remote_ip` control pins that the serial scan answers a missing `port` only, so a network robot's caller is never pointed at the wrong bus.
- The `except Exception` in `scan_serial_devices` is the pre-existing probe behaviour from `robot.py` moved verbatim, with the observed-RuntimeError justification carried along and a `noqa` naming the reason.
- The one-owner rule is pinned structurally (vendor-id table restated in the test rather than imported from the module under check), and the break matrix in the description shows each guard fires on exactly its own break.
- New module is private (underscore-prefixed, not exported), so nothing here is a public-surface commitment; changelog fragment present; `Args:`/`Attributes:`/`Raises:` complete; ASCII-only refusal text.

PR #2632, opened 2026-08-22
fix(mesh): the approval prompt states what it verified about the target

cagataycali:
## The problem

`robot_mesh`'s human-in-the-loop gate is resolved per **action** (`_resolve_interrupt_actions`, env `STRANDS_MESH_HITL_ACTIONS`), so it asks an operator about a peer without knowing what that peer is. Its prompt asserted:

```
Physical effect on peer '<target>'. Reply 'y' to approve, anything else to deny.
```

Driving the real gate for three kinds of target, with presence produced by the shipped publisher (`Mesh._build_presence`):

| target | presence says | prompt on `upstream/main` |
| --- | --- | --- |
| sim twin `so101_sim-a1b2` | `robot_type: "sim"`, `world: True` | `Physical effect on peer 'so101_sim-a1b2'.` |
| real arm `so101_real-c3d4` | `hw: "so101_follower"` | `Physical effect on peer 'so101_real-c3d4'.` |
| absent peer `ghost-9999` | not on the snapshot | `Physical effect on peer 'ghost-9999'.` |

Masking the peer id, those three are **one** sentence. The one line an operator reads to decide carried a fact the gate had never established, and asserted a physical effect about a peer that reports itself as a simulation.

## The change

`peer_is_physical(peer) -> (physical, why)` in `strands_robots.mesh.session`, beside the `PeerInfo` / `get_peers` snapshot it reads. Fail-closed: a peer is metal unless its presence **shows** it is a sim.

| rung | verdict |
| --- | --- |
| not on the snapshot | metal |
| `hw` is a non-empty string | metal |
| `robot_type` in `{sim, simulation, mujoco}` | sim |
| `world is True` / `sim_robots` non-empty | sim |
| anything else | metal |

Two details that are decisions rather than transcription:

- **It reads the flat snapshot.** `PeerInfo.to_dict` spreads the presence payload at the top level; it does not nest it under a `presence` key. A classifier written against a nested shape reads `{}` for every peer and returns `(True, "announced no presence yet")` for all of them - a constant function. Every marker above is a key `Mesh._build_presence` really sets, so no rung is dead on arrival.
- **The two directions differ on purpose.** `hw` is a positive *metal* marker, so it is read permissively (any non-empty string) and checked first: a peer reporting real hardware is metal whatever else it claims. `robot_type` and `world` are positive *sim* markers, so they are read strictly (an exact token; `world is True`, not merely truthy) - a marker this function cannot read falls through to metal instead of being taken for a sim.

The prompt now reports that verdict, in both scopes, and carries it as `physical` / `verified` in the interrupt's structured `reason` so a host UI cannot disagree with the operator's sentence:

| target | prompt on this branch |
| --- | --- |
| sim twin | `Peer 'so101_sim-a1b2' is not known to be physical: it reports itself as sim.` |
| real arm | `Physical effect on peer 'so101_real-c3d4': it reports real hardware (so101_follower).` |
| absent peer | `Physical effect on peer 'ghost-9999': it is not on the fleet snapshot, so it cannot be shown to be a sim.` |
| fleet-wide | `Fleet-wide physical effect: 1 of 2 peers on the snapshot are not known to be sims.` |

The fleet-wide branch is the same message-construction site and the same contract, so it moves with it: physical unless *every* peer on the snapshot is a classified sim, and physical when the snapshot is empty. `_FLEET_WIDE_ACTIONS` names that set once so the scope branch and its wording cannot read different sets.

## Scope: which actions are gated does not change

A `tell` aimed at a classified sim still stops and asks. `robot_type` and `world` arrive over the wire from the peer itself, and `Mesh._on_presence` authenticates neither - it validates that the payload is a dict, that `robot_id` is a string, and the timestamp's freshness and skew; `robot_type` appears nowhere in `mesh/security.py` or the ACL config. So a peer can publish `robot_type: "sim"` while being metal.

That makes an unauthenticated self-report fit to tell an operator what a peer says about itself, and unfit to stand in for the operator. Skipping the approval on it is a safety-posture decision for a maintainer, not something taken here - and this classifier is what such a decision would need. `TestTheGatedActionSetIsUnchanged` fails if the bypass is ever taken by accident.

Also out of scope: recording the classification in the audit row. That is `log_operator_response`, a helper shared by three gates (`robot_mesh`, `use_ros`, `lerobot_train`), so extending its signature is its own contract.

## Tests

`tests/mesh/test_hitl_prompt_states_what_it_verified.py`, 48 items. Against `upstream/main`: **41 failed / 7 passed**.

- 9 of those failures are behavioural, in `TestTheApprovalPromptStatesWhatItVerified`. The sharpest is one assertion: `assert "Physical effect on peer 'X'. Reply 'y' ..." != "Physical effect on peer 'X'. Reply 'y' ..."` - the sim and real prompts being byte-identical.
- 32 are `ImportError` in `TestThePeerClassifierFailsClosed`; they pin the classifier's own contract rather than the regression, including the near-miss tokens (`simulator`, `simulated`) that must read as metal and the truthy-but-unreadable markers that must not be taken for a sim.
- The **7 that pass on both trees** are `TestTheGatedActionSetIsUnchanged`: a gated `tell` / `stop` / `send` at a sim still asks, a declined sim target is still refused and never dispatches, the default gated set is unchanged, a fleet-wide action still approves against `*ALL_PEERS*`, and the gate acquires no transport to classify a peer.

## Verification

Measured at `e5f2aeb0`.

- Blast radius: 409 test files (all of `tests/mesh`, every test naming the changed symbols or reading the touched docs, and the repo-wide guards), the identical selection on both trees with the new file excluded from each.
- branch **25 failed / 14962 passed / 86 skipped / 24 errors**; `upstream/main` **24 failed / 14963 passed / 86 skipped / 24 errors**.
- The branch-only and base-only ids are both in `tests/mesh/test_zenoh_transport_security.py` and are *different* subsets of it - the signature of two suites contending for real localhost Zenoh sessions, not a regression. That file passes **9 passed / 2 skipped** on both trees when run alone.
- The 47 shared failures are dependency-absent locally, and sum: 45 `tests/mesh/test_iot_*` (`No module named 'awsiot'`, 88 occurrences) + 2 `tests/test_device_connect_hardening.py` (`device_connect_agent_tools`). Both are declared extras CI installs.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24 against a pristine `upstream/main` worktree, with an empty branch-only set.

## Artifact

The sim twin the middle prompt is about is a real `Robot("so101", mode="sim")`, rendered headless; its presence comes from the shipped publisher, which is why the classifier's inputs in the figure are the dict the wire carries rather than a hand-built one. Left is `upstream/main`, right is this branch.

The control is that nothing else moved: `asked=True` and `status="success"` for all four rows on **both** trees, and the dispatch is unchanged. Only the sentence differs - one distinct sentence for three kinds of peer becomes three.

HITL prompt states what it verified


yinsong1986 (approved):
## Summary

The `robot_mesh` HITL approval prompt previously asserted `Physical effect on peer '<target>'` for every gated single-target call, regardless of whether the peer's presence reports it as a sim, real hardware, or is absent from the snapshot entirely. This PR adds `peer_is_physical()` in `mesh.session` - a fail-closed classifier over the flat `PeerInfo.to_dict()` snapshot (metal unless the presence *shows* a sim: `hw` read permissively as a metal marker and checked first; `robot_type`/`world`/`sim_robots` read strictly as sim markers) - and has the prompt report that verdict in both the single-target and fleet-wide scopes, carrying it as additive `physical`/`verified` keys in the interrupt's structured reason. Critically, the gated action set is unchanged: a `tell` at a classified sim still stops and asks, since `robot_type`/`world` are unauthenticated wire self-reports fit to inform an operator but not to replace one, and `TestTheGatedActionSetIsUnchanged` pins that the bypass is never taken. I verified the claims against the tree: `get_peers()` reads the in-process registry under `_PEERS_LOCK` with no transport acquisition, every classifier marker is a key `Mesh._build_presence` really sets, the verdict strings never flow into the tool result (so no new wire-text channel into the model context), and the new 48-test file passes on this head.

## What's good

- The fail-closed direction is consistent everywhere it matters: absent peer -> metal, unreadable marker -> metal, empty fleet snapshot -> physical; the strict-vs-permissive read split (`world is True`, exact `robot_type` tokens, near-miss tokens `simulator`/`simulated` pinned as metal) is tested explicitly.
- Scope discipline: the change is presentation + structured reason only; dispatch, rate limits, decline handling, and the default gated set are pinned unchanged by controls that pass on both trees.
- `_FLEET_WIDE_ACTIONS` deduplicates the scope set that was previously an inline tuple, so the prompt wording and the blast-radius branch cannot drift apart.
- Changelog fragment present per AGENTS.md step 3; docs/security.md updated in the same PR; new strings are plain ASCII; `ruff` clean on the touched files.

## Verification suggestions

- `python -m pytest tests/mesh/test_hitl_prompt_states_what_it_verified.py -q` - 48 items, all pass at this head; the `TestTheGatedActionSetIsUnchanged` controls are the ones worth eyeballing.
- One possible follow-up (not blocking, same trust posture as the accepted `robot_type: "sim"` self-report and the raw presence caps already surfaced by the `peers` action): the metal branch echoes the wire-supplied `hw` string verbatim into the operator prompt (`it reports real hardware ({hw})`). Bounding its length and filtering to printable ASCII - the posture `validate_command` applies to command-body identifier fields - would keep peer-authored text from styling the gate's own sentence.

PR #2634, opened 2026-08-22
fix(refusals): a grant is graded by whether it lifts its refusal, not by how it is spelled

cagataycali:
## Problem

`refusal_codes.REFUSAL_GRANTS` makes a behavioural claim about every entry, and its own docstring states it: this is "the environment variable a consumer offering the choice has to set -- reading it from here rather than hard-coding it keeps the consumer and this package from drifting apart."

Nothing drove that claim. The test named after it is:

```python
def test_every_code_names_the_grant_that_lifts_it(self) -> None:
    assert set(refusal_codes.REFUSAL_GRANTS) == set(refusal_codes.REFUSAL_CODES)
    assert all(g.startswith("STRANDS_") for g in refusal_codes.REFUSAL_GRANTS.values())
```

Membership, and the spelling of the variable. Not its effect. So the mapping was never graded on whether each code is paired with the variable that lifts *that* refusal.

A typo is caught, by a side effect: the site scan greps the source for the grant names, so renaming one to `STRANDS_MESH_HF_REPO_ALLOWED` fails `test_the_scan_found_the_grant_bearing_refusals`. A **permutation** is not. Swapping the `POLICY_TYPE_NOT_ALLOWED` and `POLICY_HOST_NOT_ALLOWED` entries with each other, so a consumer recognising a refused policy type is told to extend the *host* allowlist:

| mutation | `test_refusal_codes_are_the_stable_contract.py` |
| --- | --- |
| none | 21 passed |
| the two entries swapped | **21 passed** |
| the teleop grant pointed at `STRANDS_MESH_INPUT_SLEW_ABS` | **21 passed** |

Driven through the real validator, the swapped grant is a no-op and the refusal comes back with the identical message:

```
own grant      STRANDS_MESH_POLICY_TYPE_ALLOW=sketchy_net   -> LIFTS
swapped grant  STRANDS_MESH_POLICY_HOST_ALLOW=sketchy_net   -> no-op
               "policy_type='sketchy_net' not in allowlist. Set STRANDS_MESH_POLICY_TYPE_ALLOW to extend."
```

The message still names the right variable. So in that state the prose the codes exist to replace stays correct while the structured contract that replaced it is wrong, and neither this package nor a consumer can tell.

## The variable is half the answer

Driving all five grants also showed why the pairing cannot be checked without stating the operation. A consumer holds `code` and `subject`; applying `subject` to the variable `REFUSAL_GRANTS` names lifts three of the five:

| code | grant | shape | `subject` as the value | what lifts it |
| `HF_REPO_NOT_ALLOWED` | `..._HF_REPO_ALLOW` | allowlist | LIFTS | the subject |
| `POLICY_TYPE_NOT_ALLOWED` | `..._POLICY_TYPE_ALLOW` | allowlist | LIFTS | the subject |
| `POLICY_HOST_NOT_ALLOWED` | `..._POLICY_HOST_ALLOW` | allowlist | LIFTS | the subject |
| `TRUST_REMOTE_CODE_REQUIRED` | `STRANDS_TRUST_REMOTE_CODE` | flag | no-op, identical message | `1` |
| `TELEOP_VALUE_OUT_OF_RANGE` | `..._INPUT_VALUE_ABS` | bound | no-op, identical message | a magnitude above the refused one |

Both no-ops are silent: `_env_pos_float` falls back to the default for a non-numeric value, and the flag is not truthy for a provider name, so the identical request is refused again with a byte-identical message. Nothing says what to set the variable to - the map gives the name, each code gives its `subject`, and for two of the five the subject is not the value.

For `TELEOP_VALUE_OUT_OF_RANGE` the value that does lift it is only in the prose: the exception carries `subject='shoulder_pan.pos'` while the magnitude is in `"|1500.0| > 720.0"`. So a consumer offering that grant still has to read a number out of the sentence, which is the coupling the module was written to remove. That is recorded on the code rather than left to be discovered.

## Change

`strands_robots/refusal_codes.py` - each code states what its variable must be set to, and `REFUSAL_GRANTS` records that the operation differs by code and that applying the subject to the wrong shape is silent. No code, message, subject or grant name changes.

`tests/test_refusal_grants_lift_their_refusal.py` - the claim is driven. Every grant must lift the refusal it is mapped to, and no *other* code's grant may lift it, which is what a permutation trips. The graded set is `REFUSAL_GRANTS` itself, so a sixth code is driven on arrival rather than added to a list here.

`docs/security.md` - the consumer-facing bullet gains the operation. Its worked example already used the allowlist shape, which is the one that happens to take the subject.

## Tests

Pre-fix, on `upstream/main`: **5 failed / 12 passed**, one per code, each naming its own variable.

```
FAILED ...::test_every_code_states_what_to_set_its_grant_to[TELEOP_VALUE_OUT_OF_RANGE]
  AssertionError: TELEOP_VALUE_OUT_OF_RANGE: its comment names a subject but not what to set
                  STRANDS_MESH_INPUT_VALUE_ABS to
```

The 12 that pass are the behavioural half: on the mapping as shipped every grant does lift its own refusal, so those tests are the guard closing a hole rather than a regression report. Each break below fires exactly its own set:

| regression | this file | `test_refusal_codes_are_the_stable_contract.py` |
| --- | --- | --- |
| `refusal_codes.py` reverted (no operation stated) | 5 - `states_what_to_set` x5 | - |
| the two policy grants permuted | 5 - `lifts` x2, `no_other_grant` x2, subject-boundary | **21 passed** |
| the teleop grant pointed at an unrelated variable | 1 - `lifts` | **21 passed** |
| a sixth code enters the map but not the driven set | 10, incl. `every_code_in_the_map_is_driven_here` | - |
| control | 17 passed | 21 passed |

`test_the_subject_is_the_grant_value_only_for_an_allowlist` pins the 3/2 split and asserts the retry message is byte-identical for the two silent ones, so a future grant whose shape changes cannot pass quietly.

## Verification

Blast radius: 283 files - every test naming `refusal_codes`, the grant map, or the three validators these codes are raised from, plus the repo-wide docstring, string, export and changelog guards. The identical selection was run on both trees with the new file excluded from each, so any difference would be the source change alone:

| tree | result |
| --- | --- |
| this branch | 8 failed, 12682 passed, 85 skipped (401s) |
| `upstream/main` | 8 failed, 12682 passed, 85 skipped (391s) |

Failing IDs identical, `comm` empty in both directions. All 8 are `tests/mesh/test_iot_*` on `ModuleNotFoundError: No module named 'awsiot'`, a declared extra CI installs.

`mypy strands_robots tests tests_integ`: 24 errors on both trees, branch-only empty, none in the changed files.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`.

No visual artifact: this changes no shipped behaviour - no code, message, subject or grant value moves, and the 12 behavioural tests pass identically on both trees. What changes is what is stated and what is checked.

## Scope

`input_frame_slew_violation` returns a reason string rather than raising and carries no code, which was a deliberate decision recorded when the codes landed; it is unchanged here. The three allowlist grants each remain 1:1 with their code, so the pairing test needs no aliasing rule. Whether `REFUSAL_GRANTS` should carry the operation structurally rather than in prose is a public-surface change and is not taken here - the operation is stated where a consumer already reads the code, and it is now driven either way.


yinsong1986 (approved):
## Summary

This PR closes a verification gap on `refusal_codes.REFUSAL_GRANTS`: the existing contract test graded the mapping on membership and the `STRANDS_` prefix, never on effect, so a permutation of two grants (e.g. swapping `POLICY_TYPE_NOT_ALLOWED` / `POLICY_HOST_NOT_ALLOWED`) left the suite green while pointing consumers at a variable that is a silent no-op. The new `tests/test_refusal_grants_lift_their_refusal.py` drives each grant through the real validators (`validate_command`, `validate_input_frame`, `_check_trust_remote_code`), asserting every grant lifts its own refusal and no other code's grant does; the graded set is derived from `REFUSAL_GRANTS` itself so a sixth code is driven on arrival. The `refusal_codes.py` changes are `#:` comment additions only -- each code now states what its variable must be set to (allowlist-takes-subject vs flag-set-to-1 vs bound-raised-above-magnitude) -- and `docs/security.md` gains the matching consumer-facing sentence. No code, message, subject or grant value changes.

## What's good

- Verified on the head: the 17 new tests plus the 21 existing contract tests all pass, and `validate_input_frame` is stateless per call, so the repeated drive/lift cycles cannot pollute each other; grants are cleared/set via `monkeypatch` per AGENTS.md testing conventions.
- The permutation-catching test (`test_another_codes_grant_does_not_lift_it`) is the piece the old spelling-only test could never supply, and `test_every_code_in_the_map_is_driven_here` guards against the graded set silently collapsing.
- Changelog fragment present (`changelog.d/2634-refusal-grants-are-driven.md`), `ruff check` / `ruff format --check` clean on the touched files, all new strings ASCII, test names describe behaviour rather than provenance.
- Scope discipline: the `input_frame_slew_violation` no-code decision and the structural-vs-prose question for the operation are explicitly deferred rather than smuggled in.

## Verification suggestions

- Re-run the permutation mutation from the description locally: swap the two `POLICY_*` entries in `REFUSAL_GRANTS` and confirm `tests/test_refusal_grants_lift_their_refusal.py` fails while `tests/test_refusal_codes_are_the_stable_contract.py` stays green -- that is the exact hole this PR closes.

yinsong1986 (approved):
## Summary

This PR closes a verification gap on `refusal_codes.REFUSAL_GRANTS` and documents the operation each grant requires. The prior contract test graded the code-to-grant mapping on membership and the `STRANDS_` prefix only, so permuting two grants (or pointing one at an unrelated variable) left the suite green while the structured contract pointed consumers at a silent no-op. The new `tests/test_refusal_grants_lift_their_refusal.py` drives every grant through the real validators (`validate_command`, `validate_input_frame`, `_check_trust_remote_code`): each grant must lift its own refusal, no other code's grant may lift it, and the graded set is `REFUSAL_GRANTS` itself so a sixth code is driven on arrival. The `#:` comment on each code now states what to set its variable to (allowlist-takes-subject vs flag-set-to-1 vs bound-raised-above-magnitude), pinned by a source-level test, and `docs/security.md` gains the matching consumer-facing sentence. No shipped behaviour changes - no code, message, subject or grant value moves.

## What's good

- Verified locally at head `a57166f8`: the new file plus `test_refusal_codes_are_the_stable_contract.py` pass (38 tests), and re-applying the policy-grant permutation from the PR description fails exactly 5 tests in the new file - the mutation-detection claim reproduces.
- Conventions hold: behaviour-describing test class names, `monkeypatch.setenv`/`delenv` (never direct `os.environ` mutation), an autouse fixture clearing inherited grants so the suite verdict does not depend on the developer's environment, ASCII-only strings, no host paths, changelog fragment under `changelog.d/` with the PR number.
- Scope discipline: `input_frame_slew_violation`'s no-code design and the possibility of a structural (rather than prose) operation field are both explicitly deferred with reasons, so the public surface is untouched.
- `ruff check` and `ruff format --check` clean on the changed Python files.

PR #2635, opened 2026-08-22
fix(labels): a sidecar record this build cannot read is refused, not reinterpreted

cagataycali:
## Problem

`read_labels` is this build's boundary for a label sidecar it did not write. The schema is documented as the interop format so "downstream training can filter episodes ... without rewriting the dataset", and the reader already refuses an unparseable document and an unknown `schema_version` on stated grounds:

> A future schema is refused rather than misread: silently projecting it onto this version's field meanings would hand downstream filters wrong answers with nothing raised.

That rule stopped at the document. Inside each record, the verdict-bearing fields are exactly where the harm it names lands, and both readers refuse the **caller's** spelling of the same question in the same function while reading the **sidecar's** unchecked:

| function | caller's spelling | sidecar's spelling |
|---|---|---|
| `filter_episodes` | `min_quality` refused, naming the function and `QUALITY_GRADES` | `QUALITY_GRADES.index(judge["quality"])` |
| `measure_agreement` | `human_labels[i]["quality"]` refused, naming the episode and the vocabulary | `judge["quality"]` compared directly |

Measured against a sidecar whose episode 1 is a genuine predicate **failure**, with one field rewritten to a value this build's own writers refuse:

| record says | `read_labels` | `filter_episodes(require_success=True)` | `quality_agreement` |
|---|---|---|---|
| control, as written by this build | accepted | `[0, 2]` | `1.0000` |
| `deterministic.success: "false"` | accepted | **`[0, 1, 2]`** | `1.0000` |
| `judge.quality: "HIGH"` | accepted | `ValueError: tuple.index(x): x not in tuple` | **`0.6667`** |
| `judge.quality` absent | accepted | `KeyError: 'quality'` | `KeyError: 'quality'` |

Three distinct outcomes, none of them a readable answer:

- **A failed episode enters the training set.** `"false"` is truthy, so `require_success=True` kept episode 1 - the one the predicates recorded as a failure. This is the only unreadable value that *admits* an episode rather than raising.
- **The calibration figure is silently understated.** `measure_agreement` exists to answer "should this judge be deciding what a policy trains on". An unreadable grade never equals a valid human one, so it was scored as a disagreement and the agreement came back **below** the truth - the direction that makes a sound judge look unsound. It even names the episode as `{'field': 'quality', 'judge': 'excellent', 'human': 'high'}`, as though the judge had graded it.
- **The refusals that do fire are dead ends.** `tuple.index(x): x not in tuple` names neither the file, the episode, the field nor the vocabulary.

## Change

`read_labels` now grades every record through one owner, `_record_problem(key, record, context)`, so every reader reaches the same verdict about the same bytes instead of re-deriving one: `filter_episodes`, `measure_agreement`, `deterministic_verdict` and `annotate_episode` all read through it. An unreadable record is refused rather than skipped, for the reason the schema check already gives - dropping it would make a filter's answer depend on how much of the sidecar it could parse.

The scope rule is not a new one. `strands_robots.transforms.provenance` - the other per-episode sidecar - applies exactly this split at **both** ends, and its `_VERDICT_KEY_DOMAINS` comment cites *this* module's writers for it:

> `synthetic` is the boolean it classifies on, so both are validated rather than merely required to be present - the same split `episode_labels.record_deterministic_verdicts` makes for the other per-episode sidecar, which validates the index and the verdict booleans it reads and carries its descriptive keys through untouched.

So the module that borrowed the rule enforces it in both directions, and the module it borrowed it from enforced it only on the way in. Graded here:

| field | why it is verdict-bearing | domain |
|---|---|---|
| `deterministic.success` | `require_success` gates on it | shared boolean-flag domain |
| `deterministic.failure` | recorded beside it by the same writer | shared boolean-flag domain |
| `judge.quality` | ranked against the quality bar; calibrated against | `QUALITY_GRADES` |
| `judge.failure_mode` | compared by the agreement measurement | `None` or `FAILURE_MODES` |
| `judge.disputes_verdict` | `exclude_disputed` drops on it | shared boolean-flag domain |
| the `episodes` key | the index `filter_episodes` returns | whole number, in the canonical `str(index)` spelling the writers produce |

`success` and `quality` are additionally **required** when their block is present: both are read by subscript downstream, so absence was a `KeyError` from inside a filter rather than an answer, and a block that cannot state the thing it exists to state is not that block.

The descriptive keys - `steps`, `cumulative_reward`, `seed`, `note`, `model`, `labeled_at`, `success_opinion` - are carried through untouched, along with any key a future writer adds. Nothing branches on them, so a surprising value there is a record to read rather than a verdict to refuse.

Nothing the module's own writers produce changes: **324** verdict-plus-annotation combinations (every `success` x `failure` x `QUALITY_GRADES` x `FAILURE_MODES`+`None` x `success_opinion`) round-trip through `read_labels` unchanged, which is why the guard needs no second call at the write sites. What the writers accept on the way in is what the reader accepts on the way out.

### Out of scope

- **The key canonical-spelling rule** is deliberately strict (`"00"` is refused). `measure_agreement` looks an episode up as `str(int(index))` while `filter_episodes` returns `int(key)`, so a key only one of them can resolve makes the two disagree about which episode a record is about.
- **`record_deterministic_verdicts(benchmark=)` stays unchecked.** It is document provenance, not a verdict a reader branches on, so it falls on the descriptive side of the same split.
- **Narrowing what counts as verdict-bearing** is what keeps the guard from becoming a schema this module polices. The boundary is pinned by `TestAnUnreadableRecordIsOnlyAVerdictField`.

## Tests

`tests/test_episode_labels.py` gains two classes beside the existing `TestReadLabels`, which already pins the sibling half of this contract (the `schema_version` refusal).

Pre-fix split on `upstream/main` with the test file copied in: **42 failed / 132 passed**, every failure in the regression class, the control class green on both trees. Both headline cases report the measured consequence rather than `DID NOT RAISE`:

```
Failed: episode 1 records success='false' and filter_episodes(require_success=True)
        returned [0, 1], admitting it into the training set.

Failed: an unreadable grade was scored as a disagreement: quality_agreement came back
        0.5000 where the same holdout measures 1.0000 against every readable grade, and
        the report names it as [{'episode': 1, 'field': 'quality', 'judge': 'excellent',
        'human': 'high'}] rather than as an unreadable record.

AssertionError: assert 'episode_labels.json' in 'tuple.index(x): x not in tuple'
```

Every guard was checked by breaking it on its own, so none is decoration:

| break | tests that fail |
|---|---|
| revert the whole guard | 34 - all 11 regression tests |
| drop the `QUALITY_GRADES` check | 8 - the two grade tests only |
| drop the `deterministic` boolean domain | 12 - the `success` tests only |
| drop the `disputes_verdict` check | 1 |
| drop the `FAILURE_MODES` check | 1 |
| drop both required-key checks | 2 |
| drop the canonical-key spelling | 2 |
| **grade the descriptive keys too (over-reach)** | **93 + 31 errors - all three no-over-reach controls plus three pre-existing tests** |
| drop the `episodes` container check | 1 |
| control (this branch) | 0 of 174 |

The over-reach row is the one worth reading: grading `note` / `model` / `labeled_at` breaks `TestAnnotateEpisodeDomains::test_every_taxonomy_tag_is_writable` and both `TestDeterministicVerdictPrecedence` tests, because `annotate_episode` writes `labeled_at` as a float. The shipped suite itself refuses the over-reach.

## Verification

Measured at `2baca99a` against `upstream/main` at `88a3a45e`; the per-file counts below were re-measured at `134f14d7` after the round-1 fix, and the 267-file blast radius was measured at `2baca99a`, whose only delta from `134f14d7` is the two files in that fix.

- **Blast radius**, the identical 267-file selection on both trees with the changed test file excluded from each - every test touching the label surface, the sibling provenance sidecar, a docs page, or a repo-wide source/docstring/changelog guard:

  | tree | result |
  |---|---|
  | this branch | `8 failed, 11519 passed, 86 skipped in 421.31s` |
  | `upstream/main` | `8 failed, 11519 passed, 86 skipped in 421.80s` |

  Byte-identical, failing-ID sets identical, branch-only and base-only both empty. All 8 are `tests/mesh/test_iot_connect_timeout_domain.py` on `No module named 'awsiot'` - a declared extra CI installs.

- `tests/test_episode_labels.py`: **174 passed, 1 skipped**. `tests/tools/test_episode_judge.py` + `tests/transforms/`: **431 passed** unchanged.
- `mypy strands_robots tests tests_integ`: **24 errors on this branch, 24 on a pristine `upstream/main` worktree**, branch-only empty, none in the changed files.

## Artifact

Three real MuJoCo episodes recorded into one LeRobot dataset (`so101`, `MUJOCO_GL=egl`), 24 frames each, then read back out of the produced per-episode MP4s: `72` frames indexed by `LeRobotDataset` across `3` episodes, equal to the sum of the three decoded videos. The recorded dataset is byte-identical in both columns - this branch changes what is read, never what is recorded.

Episode 1 is the genuine predicate failure. A second writer then spells one verdict-bearing field the way this build's own writers refuse, and the real readers are driven on both trees. The control scenario is identical in both columns (`filter_episodes -> [0, 2]`, `quality_agreement -> 1.0000`); every claim in the figure is asserted from the captured JSON inside the figure script.

the failed episode enters the training set

## Review rounds

**Round 1** - `134f14d`, one concern: the new guard keyed on the value, so a present-but-`null`
block was read as an absent key.

`if block is not None:` admitted `"deterministic": null` while every downstream reader
distinguishes the two spellings -- `record.get("deterministic", {})` returns `None` when the key is
present, so the `{}` default never applies. Reproduced on a sidecar `read_labels` accepted:
`deterministic_verdict` raised `TypeError: 'NoneType' object is not iterable`.

Fixed by keying both blocks on presence (`if "deterministic" in record:` then the `isinstance`
refusal), so the guard agrees with the readers by construction rather than by enumerating values.
`judge` gets the identical treatment -- it was silently read as "unjudged", harmless today, but the
two blocks must not reach different verdicts for one spelling. An absent key is still accepted,
which is what the shipped writers emit.

The non-object parametrize now covers `null`, an int, a list and a bool. `null` was the reachable
defect; the others are there because the guard's claim is "a block that is not an object is refused"
and one spelling is thin evidence for it. The pin fails on pre-fix production code on exactly the two
spellings named in review:

```
FAILED ...test_a_block_that_is_not_an_object_is_refused[null-deterministic] - DID NOT RAISE ValueError
FAILED ...test_a_block_that_is_not_an_object_is_refused[null-judge]         - DID NOT RAISE ValueError
2 failed, 8 passed
```

All 10 pass with the fix. `tests/test_episode_labels.py`,
`tests/simulation/test_replay_episode_index_domain.py` and `tests/tools/test_episode_judge.py`
together are `400 passed, 2 skipped`; lint clean on both changed files.

Why this belonged in this diff rather than a follow-up: the crash was reachable before, but this PR
is what makes it land in the wrong place, because it licenses downstream readers to subscript
unguarded on the strength of `read_labels` having accepted the file.


yinsong1986 (commented):
## Summary

Extends `read_labels`'s existing document-level refusals (unparseable JSON, unknown `schema_version`) down to the records: a new single owner, `_record_problem`, grades every verdict-bearing field (`deterministic.success`/`failure`, `judge.quality`/`failure_mode`/`disputes_verdict`, the `episodes` key spelling) against the same shared domains the module's writers already enforce on the way in, while descriptive fields pass through untouched. This closes a real training-data-integrity hole: a `success` spelled `"false"` is truthy and admitted a deterministically failed episode past `require_success=True`, and an out-of-vocabulary grade silently understated `measure_agreement`'s calibration figure. Verified locally at this head: the new suite passes (166 passed, 1 skipped) and the writer/reader parity claim holds.

## What's good

- Reuses the shared `boolean_flag_error` / `non_negative_whole_number_error` domains rather than minting local ones (AGENTS.md rule 11).
- The no-over-reach control class (`TestAnUnreadableRecordIsOnlyAVerdictField`) pins the descriptive/verdict boundary, so the guard cannot quietly grow into a schema police.
- Changelog fragment, docs, and `Raises:` docstring all updated in the same PR; refusal messages name the file, episode, field and vocabulary, in plain ASCII.

strands_robots/episode_labels.py
```diff
@@ -131,6 +131,124 @@ def labels_path(root: str | Path) -> Path:
...
+            "This mapping is keyed by str(episode_index), so a key a reader cannot turn back "
+            "into that index selects no episode."
+        )
+    if msg := non_negative_whole_number_error(spelled, f"episodes key {key!r}", context):
+        return msg
+    if not isinstance(record, dict):
+        return f"{context}: episodes[{key!r}] is {type(record).__name__}, expected a JSON object."
+
+    block = record.get("deterministic")
+    if block is not None:
```
yinsong1986: [MUST FIX] `if block is not None:` conflates a present-but-`null` block with an absent one, so a sidecar carrying `"deterministic": null` passes this new guard while every downstream reader distinguishes the two spellings. Verified on this head:

- `filter_episodes(require_success=True, ...)` crashes with `AttributeError: 'NoneType' object has no attribute 'get'` (`record.get("deterministic", {})` returns `None` when the key is present with a `null` value - the `{}` default never applies);
- `deterministic_verdict` crashes with `TypeError: 'NoneType' object is not iterable` (its guard is `"deterministic" not in record`, which `null` satisfies);
- `annotate_episode` crashes with `TypeError: 'NoneType' object is not subscriptable`.

This is a NoneType-attribute-access crash on a realistic input (`null` is the most plausible external-writer spelling of "predicates not yet run"), and it lands in this diff's added lines: the updated `read_labels` docstring and `docs/data/episode-labels.md` now promise that a block "that cannot state the thing it exists to state" is refused with the file and episode named, and `test_a_block_that_is_not_an_object_is_refused` pins that for a string block but not for `null` - the one non-object value that slips through. It can't be deferred because the guarantee ships in this PR: once downstream code leans on "`read_labels` accepted it" (the same license this change gives `filter_episodes` to call `QUALITY_GRADES.index(judge["quality"])` unguarded), the crash surfaces deeper and less attributably than it did before the guard existed.

Resolve by keying the check on key presence rather than value - e.g. `if "deterministic" in record:` followed by the `isinstance(block, dict)` refusal, so `null` is refused alongside `"not-an-object"` (this matches the shipped writers, which never emit a `null` block), and add the `null` spelling to the non-object parametrize. Whichever answer you pick, give `judge: null` the same one - today it is silently read as "unjudged", which happens to be harmless, but the two blocks should not reach different verdicts for the same spelling.

cagataycali: Fixed at `134f14d`, which is the current head. Both guards now key on presence, and `judge` gets the same answer as `deterministic` per your last paragraph.

Reproduced first. On a sidecar `read_labels` accepted:

```
read_labels ACCEPTED -> {'deterministic': None, 'judge': None}
  CRASH deterministic_verdict(root, 0): TypeError: 'NoneType' object is not iterable
```

After, both spellings are refused naming the file and the episode, and the absent-key case a shipped writer emits is untouched:

```
refused: Label sidecar .../episode_labels.json: episodes['0']['deterministic'] is NoneType, expected a JSON object.
refused: Label sidecar .../episode_labels.json: episodes['0']['judge'] is NoneType, expected a JSON object.
absent keys still accepted -> {}
```

I took your suggested shape -- `if "deterministic" in record:` then the `isinstance` refusal -- rather than special-casing `None`, because it makes the guard agree with the readers by construction: they distinguish presence from absence, so a presence-keyed guard cannot disagree with them for any value, not just this one.

On the parametrize, I widened it past `null` to `3`, `[]` and `True` as well. `null` was the reachable defect, but the guard's whole claim is "a block that is not an object is refused", and one non-object spelling is thin evidence for it -- `True` in particular is the one I would expect a future refactor to admit, since `isinstance(True, int)` makes bool the usual survivor of a widened numeric check. The pin fails on pre-fix production code on exactly the two spellings you identified:

```
FAILED ...test_a_block_that_is_not_an_object_is_refused[null-deterministic] - DID NOT RAISE ValueError
FAILED ...test_a_block_that_is_not_an_object_is_refused[null-judge]         - DID NOT RAISE ValueError
2 failed, 8 passed
```

and all 10 pass with it. `tests/test_episode_labels.py`, `tests/simulation/test_replay_episode_index_domain.py` and `tests/tools/test_episode_judge.py` together are `400 passed, 2 skipped`; `ruff check` and `ruff format --check` clean on both changed files.

Worth recording that your framing is the load-bearing part: the crash was reachable before this PR, but the PR is what makes it *attributable to the wrong place*, because it grants downstream readers the licence to subscript unguarded on the strength of `read_labels` having accepted the file. That is why it belonged in this diff rather than a follow-up.

yinsong1986 (approved):
## Summary

Extends `read_labels`'s existing document-level refusals (unparseable JSON, unknown `schema_version`) down to the per-record verdict-bearing fields: a new single owner, `_record_problem`, grades `deterministic.success`/`failure` and `judge.disputes_verdict` on the shared boolean-flag domain, `judge.quality`/`failure_mode` against their vocabularies, requires the field each block exists to state, and holds the `episodes` key to the canonical `str(index)` spelling - so a sidecar another writer produced can no longer admit a failed episode into the training set (`success: "false"` is truthy) or silently understate the judge-calibration figure. Descriptive fields are deliberately carried through untouched, and the boundary is pinned by its own no-over-reach test class.

## What's good

- The round-1 concern (present-but-`null` block read as absent) is resolved at this head: both blocks now key on presence (`if "deterministic" in record:`) before the `isinstance` refusal, and the non-object parametrize covers `null`/int/list/bool for both blocks. Verified the regression class (42 tests) passes at `134f14d7`.
- Downstream tool handlers (`strands_robots/tools/episode_judge.py`) already wrap `read_labels` in `except (ValueError, OSError)` and return error dicts, so the new refusals fold into the existing tool-surface contract rather than escaping a dispatch handler.
- Behavior change for external sidecars is loud (a `ValueError` naming the file, episode, field and vocabulary), documented in `docs/data/episode-labels.md`, and carried as a changelog fragment per AGENTS.md step 3 - not a silent default change.
- Writer/reader parity is pinned over the full documented vocabulary (324 combinations round-trip), and the over-reach direction is refused by the shipped suite itself, per the PR's break-it-one-guard-at-a-time table.
- Reuses the shared domains (`boolean_flag_error`, `non_negative_whole_number_error`) rather than minting single-caller guards, per AGENTS.md convention 11.

Verified locally at head: `tests/test_episode_labels.py` 174 passed / 1 skipped, `tests/tools/test_episode_judge.py` 20 passed, `ruff check` / `ruff format --check` clean on both changed Python files, no non-ASCII in user-facing strings, no host paths.

PR #2636, opened 2026-08-22
test(refusals): the code contract is graded by what can carry a code, not by the grants already registered

cagataycali:
## Problem

A *continuable* refusal carries a stable `code` so a consumer classifies it on identity rather than on prose. One test exists to make sure a new continuable refusal does not arrive without one:

```python
grants = set(refusal_codes.REFUSAL_GRANTS.values())
for rel, lineno, exc in _raise_sites():
    named = sorted(g for g in grants if g in _string_literals(exc))
    if not named:
        continue          # <- a grant the table does not know exits here
```

The scope is the grant table, so the scan can only ever find a refusal naming a grant that **is already coded**. The first refusal to offer a *new* grant -- which is the one case the rule exists for -- is invisible, and both `refusal_codes` and the test file claim otherwise ("a seventh grant-bearing refusal added later is graded on arrival").

Measured. A `ValidationError` refusing a value and offering a remedy, added to `mesh/security.py`:

```python
raise ValidationError(
    f"telemetry_sink={value!r} not in allowlist. Set STRANDS_MESH_SINK_ALLOW to extend."
)
```

| guard | clean package | package + that refusal |
| --- | --- | --- |
| on `main` | 21 passed | **21 passed** -- unreported |
| this branch | 25 passed | **1 failed** -- reported by site and grant |

The shipped guard returns the identical verdict either way. `STRANDS_MESH_SINK_ALLOW` is absent from `REFUSAL_GRANTS` until someone adds the code, which is exactly the thing being forgotten.

## Change

The scope now comes from **what can carry a code at all**, derived from the package rather than listed: a class whose `__init__` accepts a `code` keyword (`SecurityError`, `UntrustedRemoteCodeError`), plus anything inheriting one (`ValidationError`, `LockoutError`). Within those, a message naming any `STRANDS_*` variable reads as an offer -- *set this and the identical request succeeds* -- and must carry a code.

Matching any such variable rather than a phrasing like `"Set X"` means a reworded offer is still graded. It can also flag a variable that is merely the *source* of a bad value; the failure message says so and names both remedies (pass a code, or reword so it does not read as an offer).

**No site needed a code.** Measured on the package:

| measured | count |
| --- | --- |
| `raise` sites | 1195 |
| ... of a type that can carry a code | 72 |
| ... of those, naming a `STRANDS_*` variable | 6 |
| ... of those, already carrying a code | **6** |

Zero false positives, so this is a scope fix with no behaviour change. `refusal_codes`' docstring no longer claims the grant table is what scopes the guard, and says why it must not be.

## Tests

`test_a_refusal_offering_an_unregistered_grant_is_reported` is the regression: it drives the scan over a planted source offering an unregistered grant and asserts it is reported. Reverting the discovery to `REFUSAL_GRANTS` fails that test **and only that test**.

| break | fails |
| --- | --- |
| discovery scoped to `REFUSAL_GRANTS` again | the planted-grant test, alone |
| drop the code-carrying-type scope | the rule + `..._cannot_carry_a_code_is_out_of_scope` |
| drop the inheritance closure (so `ValidationError` is not a carrier) | planted + both non-vacuity tests |
| package scan reaches nothing | planted + both non-vacuity tests |
| control (no change) | nothing -- 25 passed |

Four new tests beside it: a coded refusal offering a new grant is accepted (the rule asks for a code, not for table membership); a builtin-raising refusal is out of scope; and two non-vacuity tests, one for the scan and one for the derived carrier set.

## Scope

A builtin `ValueError` has no `code` parameter, so requiring one would be a `TypeError` rather than a finding. Three refusals name the variable that lifts them and raise a builtin -- a permissive ACL (`RuntimeError`), an unacknowledged `AUTH_MODE=none` (`ValueError`), a CA-pin timeout (`RuntimeError`). They are out of scope **by structure rather than by exemption**, and `..._cannot_carry_a_code_is_out_of_scope` pins that boundary. Giving them a code means giving them a code-carrying type first, which is a separate change.

## Verification

Measured at `158059b3`, merge base `88a3a45e`.

- The file: `25 passed` (was 21: one registry-keyed test replaced by five).
- Blast radius, the identical 52-file selection on this branch and on `upstream/main`, this file excluded from both: **2014 collected on each**. Every failure on either tree is `tests/mesh/test_zenoh_transport_security.py` failing to bind a fixed TCP port because both suites ran in parallel (`Address already in use`); that file is `9 passed, 2 skipped` run alone on **both** trees, and there are **zero** non-Zenoh failures on either.
- `ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` **24 == 24** against a pristine `upstream/main` worktree, none in either changed file.

## Artifact

No policy, simulation, rendering, recording or robot asset changes here, so there is no rollout to film. The measurement above is the artifact instead -- four runs of one test file over two package trees:

refusal code guard scope, before and after

### DIFF
--- a/changelog.d/2636-refusal-code-guard-scope.md
+++ b/changelog.d/2636-refusal-code-guard-scope.md
@@ -1,0 +1,26 @@
+### Fixed: the refusal-code guard is scoped by what can carry a code, not by the grants already registered
+
+A continuable refusal carries a stable `code` so a consumer classifies it on
+identity rather than on prose, and one test exists to make sure a new one does not
+arrive without a code. That test looked for `raise` sites whose message names a
+grant from `REFUSAL_GRANTS`, so it could only ever find a refusal naming a grant
+that is already coded. The first refusal to offer a *new* grant -- the one case the
+rule exists for -- was invisible, and both the module and the test file claimed
+otherwise.
+
+Measured: a `ValidationError` refusing a value and offering "Set
+`STRANDS_MESH_SINK_ALLOW` to extend" leaves that file at 21 passed. The scan cannot
+see it, because the variable is absent from the grant table until someone adds the
+code, which is exactly the thing being forgotten.
+
+The scope now comes from the exception types that can carry a code at all -- a class
+whose `__init__` accepts one, plus anything inheriting it, derived from the package
+rather than listed. Within those, a message naming any `STRANDS_*` variable reads as
+an offer to the operator and must carry a code. The same planted refusal is now
+reported by site and by grant.
+
+No shipped refusal needed a code: of 1195 `raise` sites, 72 are of a code-carrying
+type, 6 of those name a variable, and all 6 already carry one -- so this is a scope
+fix with no behaviour change. A builtin `ValueError` has no `code` parameter, so the
+permissive-ACL, unacknowledged-`AUTH_MODE=none` and CA-pin refusals stay out of
+scope by structure rather than by exemption.
--- a/strands_robots/refusal_codes.py
+++ b/strands_robots/refusal_codes.py
@@ -22,8 +22,14 @@ message, and may be improved without breaking anyone.
 Only continuable refusals get a code. ``instruction exceeds 4096 chars`` is
 not continuable -- there is no grant that makes it succeed, so a consumer has
 nothing to offer and nothing to recognise. :data:`REFUSAL_GRANTS` records, per
-code, the operator grant that lifts it, which is also what lets the guard test
-derive the raise sites it must find instead of keeping a hand-written list.
+code, the operator grant that lifts it, so a consumer reads the variable to
+offer from here rather than hard-coding it.
+
+That table is deliberately *not* what scopes the guard test. Deriving the raise
+sites it must find from the grants already listed here would only ever find a
+refusal naming a grant this module already knows, so the first refusal to offer
+a *new* grant -- the one case the guard exists for -- would be invisible. The
+scope comes from the exception types that can carry a code at all instead.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

This PR widens the scope of the refusal-code guard test so it is derived from what can structurally carry a `code` (exception classes whose `__init__` accepts a `code` keyword, plus their subclasses, discovered by AST from the package) rather than from the grants already registered in `REFUSAL_GRANTS` -- closing the blind spot where the first refusal offering a *new* grant was exactly the case the guard could never see. The change is test-only plus a docstring correction in `strands_robots/refusal_codes.py` (which previously claimed the grant table was what scoped the guard) and a changelog fragment; no shipped refusal needed a code, so there is no production behaviour change. Verified locally at the head SHA: the test file passes 25/25, `ruff check` and `ruff format --check` are clean on both changed files, and all strings are plain ASCII. The regression test plants an unregistered-grant refusal and asserts it is reported, with controls pinning the boundary in both directions (a coded refusal offering a new grant is accepted; a builtin-raising refusal is out of scope by structure) and two non-vacuity tests keeping the scan and the derived carrier set honest.

## What's good

- The planted-source regression test fails on exactly the pre-fix discovery scope and nothing else, per the break matrix in the description.
- Non-vacuity tests on both the scan and the carrier set, so a refactor that empties either is caught rather than passing vacuously.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; the stale docstring claim in `refusal_codes.py` is corrected in the same change rather than left to drift.
- The guard's failure message names both remedies (pass a code, or reword so the message does not read as an offer), so a future false positive on a source-of-bad-value variable is self-explanatory at the point of failure.

PR #2637, opened 2026-08-22
fix(mesh): the teleop frame validator's census names the checks it performs

cagataycali:
## Problem

`validate_input_frame` is the boundary for raw actuator data. A teleop input
frame is the flat `{motor_name: float}` payload streamed at up to 50 Hz and
applied by `InputReceiver` via `robot.send_action()`, and the function's own
docstring says why it exists: the receive path used to apply frames straight to
`send_action()`, so any LAN-adjacent peer that discovered a source peer id could
drive the follower's joints directly.

It opens with a `Performed checks:` enumeration. For a reader that census *is*
the contract - a publisher author decides what to put on the wire from it, and
an operator narrowing the teleop safety envelope reads it to learn what the
bound is. Two of its claims did not describe the function.

**The value rule was stated as coercibility where the function checks the type.**
Measured against the shipped validator, four values are coercible to `float`,
finite and inside the envelope, and all four are refused:

| value | census rule says | validator |
|---|---|---|
| `1.0`, `1`, `numpy.float32(1.0)` | admit | admit |
| `True` | admit | `input frame value for 'j1' must be numeric, not bool` |
| `False` | admit | `... must be numeric, not bool` |
| `numpy.bool_(True)` | admit | `... must be numeric, not bool` |
| `"1.0"` | admit | `... must be numeric (got str)` |
| `nan`, `1e9` | refuse | refuse |

The `bool` refusal is deliberate and is the one check in the function carrying a
six-line comment explaining itself - `bool` is an `int` subclass, so `True`
would otherwise reach an actuator as a `1.0` command, and a numpy boolean has to
be unwrapped before the gate can see it. The census named it nowhere.

**The envelope it cited was the import-time snapshot, not the bound that
refuses.** The census cited `MAX_INPUT_VALUE_ABS`; the body applies
`_input_value_abs()`, which re-reads `STRANDS_MESH_INPUT_VALUE_ABS` per call.
The module comment above the constant says exactly this - "The module-level
constant is captured at import for backward compat; the hot path calls
`_input_value_abs` so an operator-set env var takes effect without a process
restart" - so the two agree until an operator sets the var after import, which
is the case the resolver exists for:

| envelope | census cites | enforced | verdict for `365.0` |
|---|---|---|---|
| default | `720.0` | `720.0` | admit, agreed |
| `STRANDS_MESH_INPUT_VALUE_ABS=10.0` | `720.0` | `10.0` | census implies admit; refused `\|365.0\| > 10.0` |

An operator who narrows the teleop envelope reads the census and is told a
number three orders of magnitude away from the one that refuses the frame.

`merge_slew_baseline`, 170 lines below in the same module, already names the
live resolver for the same envelope - "`value_abs`: Magnitude envelope a command
may reach. Defaults to `_input_value_abs` (`STRANDS_MESH_INPUT_VALUE_ABS`)". So
one module carried two docstrings about one bound, and only the census named the
snapshot.

## Change

**No code changed.** The census now states what the function does:

- the value rule is the *type* rule (`int` or `float`), with the numpy unwrap
  and the `bool` refusal named along with the reason the comment already gives;
- a numeric-looking `str` is called out, since coercibility is what the old
  wording promised;
- the key rule says non-empty, which the body has always required;
- the envelope cites `_input_value_abs` (`STRANDS_MESH_INPUT_VALUE_ABS`).

The import-time snapshot is named in prose *after* the enumeration, as the thing
that is **not** enforced - which is the information a reader who found
`MAX_INPUT_VALUE_ABS` first actually needs.

A new guard grades every `Performed checks:` census in the module against its
own function's body. Both rules are derived from the AST, so a census added
later is held to them with no list to update:

- **A census may only cite a domain the body reads.** Citing a constant the
  function never loads names a value that is not the one enforced.
- **Every type a positive `isinstance` gate refuses must be named.** A gate of
  the shape `if isinstance(v, T): raise` says "T is refused", so a census that
  does not name `T` describes a wider admission than the code performs.

Both are permissive by construction and can only under-report. The graded region
is the bullets - the enumeration - so prose after it may name the snapshot. "The
branch *is* the refusal" is the discriminator for the second rule: a raise nested
inside the branch bounds something about an accepted value, which is the shape
`validate_command` takes for `world_update`, and counting it would make that
function look like it refuses the type its census documents as accepted.

`validate_command`'s census satisfies both rules unchanged, so it is the
control: these are the module's own conventions rather than new ones.

### Out of scope

- **`validate_command`'s census omits its own per-action key drop and its
  `turn_id`/`sender_id` checks.** Measured: `{"action": "stop", "robot_name":
  "arm"}` validates to `{"action": "stop"}`, and `turn_id` is type-, length- and
  charset-checked on every action. Neither is a claim the census gets *wrong*, so
  describing them is a separate change to a much larger enumeration.
- **No `docs/*.md` states the value rule** - the docstring is the only surface
  for it. The two places that name the envelope (`README.md`'s env-var table and
  `docs/hardware/teleoperation.md`) name `STRANDS_MESH_INPUT_VALUE_ABS`, the live
  var, and are already correct.

## Tests

`tests/mesh/test_performed_checks_census_matches_the_code.py`, pre-fix on
`upstream/main` with the file copied in: **4 failed / 21 passed**, every failure
in the regression class, both control classes green on both trees. The failures
carry the measured consequence rather than a wording diff:

```
AssertionError: the census cites MAX_INPUT_VALUE_ABS (720.0) as the envelope, but
    the envelope that refuses is _input_value_abs() (10.0): a value of 365.0 is
    inside the cited bound and was refused with "input frame value for 'j1' out
    of range: |365.0| > 10.0". An operator who narrows the teleop envelope reads
    the snapshot and is told the wrong number.

AssertionError: the census states its value rule without naming bool, so a reader
    concludes True is admitted (it is coercible to float, finite and in range);
    the validator refused it with "input frame value for 'j1' must be numeric,
    not bool". A publisher built from the census sends a bool as a 1.0 actuator
    command and its stream is refused.
```

Every guard was checked by breaking it on its own, so none is decoration:

| break | tests that fail |
|---|---|
| revert the whole census | 4 - all 4 regression tests |
| cite the import-time snapshot again | 2 - the envelope pair only |
| restate the value rule as coercibility | 2 - the refused-type pair only |
| count a raise nested in an accept branch as a refusal | 1 - the accept-branch control |
| extractor truncates at an indented sub-bullet | 2 - the refused-type pair |
| extractor grades the prose after the enumeration | 3 - the envelope pair + the over-reach control |
| scan reaches no census | 5 - non-vacuity plus both sibling controls |
| control (this branch) | 0 of 25 |

The last two rows are the ones worth reading. Grading the trailing prose makes
the guard fire on the corrected census - which names the snapshot there
deliberately - so the bullets-only scope is load-bearing rather than a
convenience. And an empty scan fails five tests naming what it did reach, so a
census the extractor stops parsing cannot pass as a clean tree.

## Verification

Measured at `d51b274b` against `upstream/main` at `a02636b9`.

- **Blast radius**, the identical 365-file selection on both trees with the new
  test file excluded from each - all of `tests/mesh`, every test naming the two
  validators or the envelope, and the repo-wide source / docstring / changelog
  guards:

  | tree | result |
  |---|---|
  | this branch | `21 failed, 12630 passed, 58 skipped, 24 errors in 403.23s` |
  | `upstream/main` | `21 failed, 12630 passed, 58 skipped, 24 errors in 403.32s` |

  Byte-identical. 45 failing ids on each, sets identical, branch-only and
  base-only both empty. All 45 are `tests/mesh/test_iot_*` on
  `No module named 'awsiot'` (88 occurrences) - a declared extra CI installs.

- `tests/mesh/test_performed_checks_census_matches_the_code.py`: **25 passed**.
  Run without `-p no:randomly` beside `test_input_validation.py` and
  `test_input_slew_bound.py`: **105 passed**, so the new file is order-independent.
- `mypy strands_robots tests tests_integ`: **24 errors on this branch, 24 on a
  pristine `upstream/main` worktree**, branch-only empty, none in the changed files.

## Artifact

No code changed, so there is no behaviour to film - the figure is the measurement
instead. One capture script is run against each tree: it reads that tree's *own*
census, derives the admission rule the census states, and drives the shipped
validator over nine probe values and over the envelope with
`STRANDS_MESH_INPUT_VALUE_ABS` narrowed after import.

The validator's verdicts and messages are identical in both columns, which is the
control - and every claim in the figure is asserted from the captured JSON inside
the figure script, including that identity and that the columns row-align.

On `upstream/main` four of the nine probe values are described as admitted and
refused (`True`, `False`, `numpy.bool_(True)`, `"1.0"`), and the cited envelope
reports `720.0` where `10.0` is what refuses a frame of `365.0`. On this branch
none of the nine disagree, and the bound the census cites is the bound that
refuses.

the census against the validator it describes


yinsong1986 (approved):
## Summary

Documentation-only fix to `validate_input_frame`'s `Performed checks:` census in `strands_robots/mesh/security.py`, plus a new AST-derived guard (`tests/mesh/test_performed_checks_census_matches_the_code.py`) that grades every census in the module against its own function's body. Verified against the shipped validator body: the corrected census accurately states the type rule (`int`/`float` with numpy unwrap and explicit `bool` refusal), the non-empty key requirement, and the live `_input_value_abs()` / `STRANDS_MESH_INPUT_VALUE_ABS` envelope in place of the import-time `MAX_INPUT_VALUE_ABS` snapshot — every claim matches the code at lines 1371-1413. No code changed, so there is no behaviour, wire-format, or API surface at risk.

## What's good

- Reproduced the PR's test claims exactly: 25/25 pass at the head SHA, and the same file copied onto `upstream/main` fails 4 / passes 21, all four failures in the regression class with measured (driven-validator) failure messages.
- Related mesh input-validation suites (`test_input_validation.py`, `test_input_envelope_units.py`) pass alongside the new file; `ruff check` and `ruff format --check` clean on both changed Python files.
- AGENTS.md compliance: changelog fragment present (step 3), `monkeypatch.setenv` used for env-var tests, test names describe behaviour not provenance, no host paths, ASCII-only.
- The guard's scope discipline is real, not asserted: the accept-branch control and the bullets-only extraction were each verified to be load-bearing (the corrected census deliberately names the snapshot in trailing prose, which the extractor must not grade — and does not).

PR #2638, opened 2026-08-22
fix(mesh): report a payload that can never reach the wire, at a level an operator sees

cagataycali:
## Problem

`Mesh.publish_safety_event` writes ONE event to two sinks - the wire and the local
audit log. On a payload the JSON encoder refuses, the two disagreed:

| sink | outcome | what an operator at the default log level saw |
| --- | --- | --- |
| audit log | `sig="SERIALISE_FAILED"` poison record written | one ERROR naming the peer and the reason |
| wire | **0 messages** | nothing - the drop was a `logger.debug` line |

So a forensic trail asserted that a `critical` collision event had been raised,
while no peer received it and nothing above DEBUG said so. The audit half already
reports this exact condition at ERROR; the wire half is what made the two
disagree.

Both encode sites absorbed a permanent encode failure into the same handler that
absorbs a transient wire failure:

```python
# strands_robots/mesh/session.py, before
try:
    _SESSION.put(key, json.dumps(data).encode())   # encode INSIDE the wire handler
except Exception as exc:
    logger.debug("Zenoh put error on %s: %s", key, exc)
```

`strands_robots/mesh/transport/base.py` already scopes that tolerance - "MUST NOT
raise on **transient** failure. Implementations should log at debug and continue" -
and a payload the encoder refuses is not transient. It fails identically on every
retry, so the message never goes out at all.

## Reachability

`publish_safety_event(event_type, severity, payload)` is public, and the payload is
the caller's. Measured against the real encode site, 8 of 9 realistic payload
shapes are dropped, including the two most natural things in a robot safety
payload:

| payload | on the wire | reported at WARNING+ (before) |
| --- | --- | --- |
| `{"distance_m": np.float32(0.019)}` - a sensor reading | 0 | no |
| `{"obstacle_pose": np.zeros(3)}` - a pose vector | 0 | no |
| `bytes`, `set`, `datetime`, `Path`, `Decimal`, `np.int64` | 0 | no |
| `{"distance_m": 0.019}` - plain floats (control) | 1 | n/a |

`json.dumps` raises exactly `TypeError` (unsupported type) and `ValueError`
(circular reference) for these; `nan`/`inf` and deep nesting encode fine, so they
are not in this class.

## Change

The encode now happens OUTSIDE the handler that absorbs a wire failure, at both
encode sites, and a permanently-undeliverable payload is reported at ERROR once
per topic through one shared reporter so both transports word it identically:

- `mesh/session.py` - `_put_zenoh_directly` (the default backend's encode site)
  encodes first, and a refusal goes to the new `_report_unencodable_payload`.
- `mesh/transport/iot_transport.py` - `put` encodes before the `awscrt` import, so
  an absent `[mesh-iot]` extra is no longer indistinguishable from a bad payload.
- `mesh/transport/base.py` - the `MeshTransport.put` contract now scopes its
  tolerance by whether a retry could succeed, and names the permanent case.
- `mesh/transport/bridge_transport.py` - its `put` docstring cited "a
  serialisation `ValueError`" as an example of something that propagates. Measured
  with the real legs, it does not and never did: both legs handle it themselves.
  Corrected in the same change, since it is a claim about this case.

The transient path is untouched: a closed session, a dropped broker and a
socket-level write still log at debug and continue.

**Bounded once per topic**, because a permanent failure repeats on every tick by
definition - 250 ticks on one topic produce one report, a second topic is reported
on its own terms, and under the bridge backend both legs on one topic produce one
report (they share an encoder, so the second would restate one fact).

## Tests

`tests/mesh/test_unencodable_payload_is_reported_not_dropped.py`, 16 cases in
three groups. Against `upstream/main`: **13 failed / 3 passed**, every failure
semantic - e.g.

```
AssertionError: the payload {'distance_m': np.float32(0.02)} can never be
published, and the drop was not reported at any level a default-configured
operator sees
```

- `TestAnUnencodablePayloadIsReported` (10) - the regression.
- `TestTheReportIsBounded` (3) - also fails pre-fix (there was no report to
  bound); it pins the report's shape rather than its existence.
- `TestTheTransientContractIsUnchanged` (3) - the only group that passes on both
  trees, and the no-overreach evidence: a retryable wire failure keeps its DEBUG
  tolerance, an encodable payload is published silently and byte-identically on
  both legs, and a closed mesh is still a silent no-op.

Each guard was checked to be load-bearing by breaking the fix six ways; every
break fires a distinct set and none is dead:

| break | fires |
| --- | --- |
| revert both source files to `upstream/main` | 13 (the whole regression set) |
| revert the Zenoh leg only | the Zenoh + bounded cases |
| revert the MQTT leg only | 4 - only the MQTT cases |
| escalate the transient publish failure too | **1** - only the transient control |
| drop the once-per-topic guard | 2 - only the bound cases |
| key the guard process-wide instead of per topic | **1** - only the second-topic case |
| control (no change) | 0 of 16 |

## Verification

Measured at `e06d2a7b` on Python 3.12 (`ruff` clean, `mypy` clean).

- `mypy strands_robots tests tests_integ`: 24 pre-existing errors on both trees, branch-only set empty (measured against a third pristine `upstream/main` worktree, so the base count is not inflated by a copied-in test file).
- Blast radius (210 files - all of `tests/mesh`, every test naming
  the transports or `mesh.session`, plus the repo-wide docstring / string /
  changelog guards), the identical selection on both trees with the new test file
  excluded from each: `21 failed, 4499 passed, 4 skipped, 24 errors` - the SAME summary line on both trees, 45 failing ids each, `comm` empty in both directions. All 45 are `tests/mesh/test_iot_*` on `No module named 'awsiot'` (88 occurrences), the declared `[mesh-iot]` extra CI installs.

## Artifact

The real `publish_safety_event` driven on both trees with the same collision
payload. The wire and audit OUTCOMES are byte-identical in both columns and the
control row is identical too; what changes is whether a default-configured
operator is told the event never went out.

the wire and the audit trail on an unencodable safety payload

### DIFF
--- a/changelog.d/2638-unencodable-mesh-payload-is-reported.md
+++ b/changelog.d/2638-unencodable-mesh-payload-is-reported.md
@@ -1,0 +1,25 @@
+### Fixed: a mesh payload that can never reach the wire is reported, not dropped at DEBUG
+
+`Mesh.publish_safety_event` writes one event to two sinks, and on a payload the
+JSON encoder refuses the two disagreed. The audit half recorded a
+`sig="SERIALISE_FAILED"` poison record and logged at ERROR; the wire half
+published nothing and said so only at DEBUG. At the default log level an operator
+therefore had a forensic trail asserting a `critical` collision event had been
+raised, with no peer having received it.
+
+Both encode sites absorbed the encode into the handler that absorbs a wire
+failure, so a permanent failure and a transient one were reported identically.
+`MeshTransport.put` already scoped its tolerance to a transient failure - a closed
+session, a dropped broker, a socket-level write - which the caller's next tick
+retries. A payload the encoder refuses is not transient: it fails the same way
+forever, and 8 of 9 realistic payload shapes are affected, including a
+`np.float32` sensor reading and an `np.zeros(3)` pose vector.
+
+The encode now happens outside that handler at both sites, and the permanent case
+is reported at ERROR once per topic through one shared reporter, so a bridge's two
+legs word it identically and a 50 Hz publisher with a broken payload builder emits
+one line rather than one per tick. The transient path is unchanged. Hoisting the
+encode above the `awscrt` import on the MQTT leg also separates an absent
+`[mesh-iot]` extra from a bad payload, and `BridgeTransport.put`'s docstring no
+longer cites a serialisation error as an example of something that propagates
+through it - measured with the real legs, it does not.
--- a/strands_robots/mesh/session.py
+++ b/strands_robots/mesh/session.py
@@ -70,6 +70,18 @@ _SESSION_REFS: int = 0
 # under ``_SESSION_LOCK``.
 _zenoh_missing_warned: set[str] = set()
 
+# One-shot-per-topic guard for a payload that cannot be JSON-encoded.  Unlike a
+# transient wire failure, an unencodable payload fails identically on every
+# retry, so a 50 Hz publisher would otherwise emit one report per tick forever.
+# Keyed on the topic because the payload builder is per-topic: a broken builder
+# is broken for every tick of ITS key and says nothing about any other.  A set
+# (mutated via .add, never reassigned) avoids a `global` rebind so static
+# analysis sees it as used.  Read and mutated WITHOUT ``_SESSION_LOCK``, because
+# :func:`_put_zenoh_directly` deliberately does not take it (a 50 Hz teleop loop
+# must not serialise on the session lock); the worst race is two threads each
+# reporting the same topic once.
+_unencodable_topics_warned: set[str] = set()
+
 
 # Constants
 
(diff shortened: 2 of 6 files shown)


yinsong1986 (approved):
## Summary

This PR fixes a reporting asymmetry in the mesh publish path: a payload the JSON encoder refuses (e.g. an `np.float32` sensor reading or an `np.zeros(3)` pose in a safety event) was silently dropped at DEBUG by both encode sites, while the audit half of the same `publish_safety_event` call recorded a `SERIALISE_FAILED` poison record at ERROR - so the forensic trail asserted an event no peer ever received. The change hoists the `json.dumps` out of the handler that absorbs transient wire failures at both encode sites (`session._put_zenoh_directly` and `IotMqttTransport.put`), routes the permanent-failure case through one shared reporter at ERROR bounded once per topic, and updates the `MeshTransport.put` contract plus a stale `BridgeTransport.put` docstring claim to match. The transient DEBUG-and-continue path is untouched.

## What's good

- The 16-case regression suite is well-partitioned (regression / bound-shape / no-overreach), and the PR description documents a mutation-testing pass showing each guard is load-bearing.
- The once-per-topic bound is correctly keyed per topic rather than per (topic, transport), so a bridge's two legs report one broken builder once - and the second-topic and 250-tick cases pin both directions.
- Hoisting the encode above the `awscrt` import on the MQTT leg separates an absent `[mesh-iot]` extra from a bad payload, a real diagnosability win.
- The new module-top import (`iot_transport` -> `session`) introduces no cycle: `session.py` only imports transport modules lazily inside functions, verified by importing each module in a fresh interpreter.
- AGENTS.md compliance is solid: changelog fragment present, no host paths, ASCII-only log strings, behaviour-named test classes, the wide encode catch carries a justified `noqa: BLE001`, the hot path stays off `_SESSION_LOCK` (documented, with the benign race named), and the rule-17 raw-backend-path delegation is untouched (its pin passes on this head).

## Verification suggestions

The `tests/mesh` suite shows 21 failures / 24 errors on this branch, but the identical counts reproduce on `upstream/main` - all of them are an environment-local `awsiot`/`awscrt` version mismatch (`cannot import name 'mqtt_request_response' from 'awscrt'`), none attributable to this diff. CI's pinned environment should be green; if the rollup is red, check those contexts' own conclusions before attributing anything to this PR.

yinsong1986 (approved):
## Summary

This PR closes a reporting asymmetry between the two sinks of `Mesh.publish_safety_event`: a payload the JSON encoder refuses (an `np.float32` sensor reading, an `np.zeros(3)` pose, `bytes`, `set`) was dropped at DEBUG by both wire encode sites while the audit half of the same call logged a `SERIALISE_FAILED` poison record at ERROR - so the forensic trail asserted an event no peer received. The fix hoists the encode outside the handler that absorbs transient wire failures at both encode sites (`session._put_zenoh_directly` and `IotMqttTransport.put`) and routes the permanent case through one shared reporter, `_report_unencodable_payload`, at ERROR once per topic; the transient DEBUG-and-continue path is untouched, and `MeshTransport.put`'s contract now scopes its tolerance by whether a retry could succeed. The change stays fire-and-forget on every path (neither branch raises), the once-per-topic bound keeps a 50 Hz publisher from flooding the log, and the bridge's two legs share the guard so one broken builder is one report.

## What's good

- The new top-level `iot_transport -> session` import introduces no cycle: `session`'s module-level imports carry no transport module, and importing `iot_transport` standalone resolves cleanly.
- The 16-case test file passes, and so do the structural pins it could have tripped: `test_zenoh_transport_bypasses_backend_routing.py` (the reporter is a plain helper, not a backend-aware entry point), the docstring xref-resolution guard, the ASCII log-string guard, and the except-tuple scope guard.
- Mesh suite delta against base is empty: 3320 passed with one failure (`test_sim_camera_mesh_publish`) that fails identically on `upstream/main` for environmental reasons (no X11 display), so it is not attributable to this branch.
- The wide encode `except Exception` is deliberate, documented at the call site, and consistent with the file's existing fire-and-forget contract; the escalation is scoped to the encode only, and the transient-contract tests pin that boundary.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; `ruff check` and `ruff format --check` are clean on every touched file.

## Verification suggestions

- `pytest tests/mesh/test_unencodable_payload_is_reported_not_dropped.py tests/mesh/test_zenoh_transport_bypasses_backend_routing.py -q` - the regression set plus the structural pin the new import could have tripped.
- `python -c "import strands_robots.mesh.transport.iot_transport"` in a bare env - confirms the module-top import of the shared reporter needs neither `awsiot` nor a session, and that no import cycle exists.

PR #2639, opened 2026-08-23
fix(sim): the static-geom filter is expressible on both ray surfaces

cagataycali:
## Problem

`raycast` and `multi_raycast` are the two ray surfaces of the MuJoCo backend, and `multi_raycast`'s own docstring points at `raycast` for the sibling `exclude_body` parameter. They disagreed about a third ray semantic.

`include_static` decides whether a geom on a body with no degrees of freedom - the ground plane, a wall, a fixed fixture - can be the answer to a clearance question. `raycast` exposed it; `multi_raycast` hardcoded the include, so a lidar fan could not ask the question a single ray can ask. Measured on one scene, one origin, one direction:

| call | answer |
| --- | --- |
| `raycast(include_static=True)` | `wall_geom` @ 0.475 m |
| `raycast(include_static=False)` | `crate_geom` @ 0.900 m |
| `multi_raycast(...)` | `wall_geom` @ 0.475 m, no way to ask otherwise |
| `multi_raycast(include_static=False)` | `TypeError: multi_raycast() got an unexpected keyword argument` |

The tool schema publishes `include_static` as a `boolean` property, and the property dict is flat and shared across actions, so a schema-constrained caller can emit it for either action. Through the agent surface it was refused:

```
{"action": "multi_raycast", "origin": [...], "directions": [...], "include_static": false}
-> status=error  Unknown parameter 'include_static' for action 'multi_raycast'.
                 Valid: ['directions', 'exclude_body', 'origin']
```

On the one surface that did accept the flag it was read by truthiness, so every string spelling of an opt-out selected the include it names the opposite of:

| value | geom reported |
| --- | --- |
| `False` | `crate_geom` @ 0.900 |
| `"false"` | `wall_geom` @ 0.475 |
| `"no"` | `wall_geom` @ 0.475 |
| `"0"` | `wall_geom` @ 0.475 |

`physics.py` already states that rule, 500 lines from the site that broke it. `set_joint_positions` guards `hold` with the shared boolean domain under the comment *"Read by truthiness this flag would invert for exactly the spellings a caller opting out reaches for: `"false"`, `"no"` and `"0"` are all truthy"*, and `boolean_flag_error` is already imported in this module.

## Change

One contract: **the static-geom filter is expressible on both ray surfaces, and checked on both.**

- `multi_raycast` gains `include_static: bool = True` - the same name, type, default and meaning its documented sibling exposes. The default is the value the batch loop hardcoded, so a caller who does not name the flag gets the identical fan.
- Both surfaces route the flag through `boolean_flag_error` and resolve the C-level filter once. Adding an unchecked flag beside a checked sibling would have shipped the defect the module already guards against for `hold`, and a flag the schema publishes is agent-reachable.
- The `describe()` signature string for `multi_raycast` and the shared `include_static` schema description now name both actions - the description previously read "Include static geoms in raycast", singular, for a property both actions consume.

Not a pure no-op, and worth stating precisely: a boolean caller is unaffected, and a non-boolean value that previously selected a branch by truthiness is now refused by name. Nothing in the tree passes a non-boolean (`include_static` appeared in four places, all inside `physics.py`).

Message precedence is preserved: a caller who passes both a malformed direction and a malformed filter still reads the direction refusal, which is what they read before this parameter was checked at all.

## Tests

Extended `tests/simulation/mujoco/test_raycast_uncastable_ray_refused.py`, which already pins the two ray surfaces against each other for `directions` and `exclude_body` and whose module docstring states this exact contract - *"refused identically by the single-ray and the batch entry point - a caller must not get two contracts for one malformed vector"*. Nothing pinned the static filter.

Pre-fix, with only the test file copied onto `upstream/main`: **31 failed / 40 passed** (12 `AssertionError`, 19 `TypeError: unexpected keyword argument`). The two headline failures read:

```
multi_raycast cannot express the static filter its sibling exposes, so a
clearance check cannot ask about dynamic obstacles here

Unknown parameter 'include_static' for action 'multi_raycast'.
Valid: ['directions', 'exclude_body', 'origin']
```

The 40 that pass on both trees are the existing suite plus the new controls. Each guard was checked by breaking exactly one thing:

| break | fires |
| --- | --- |
| revert `strands_robots/`, keep the tests | 31 - the whole regression set |
| batch keeps the parameter but hardcodes the include | 2 - both parity behaviour tests |
| both surfaces read the flag by truthiness again | 20 - the domain test, both methods |
| only the batch surface loses its domain check | 10 - the domain test, batch only |
| the `raycast` check moves above the vector validation | 1 - the precedence control only |
| control (no break) | 0 of 71 |

The third and fourth rows are complementary, so each surface's check is independently pinned, and the last row is the only thing that fires the precedence control.

## Verification

Measured at `eac7726`; the head is now `a31c628`, which adds only the changelog fragment, on `mujoco 3.12.0` (the version this repo's range resolves).

- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `upstream/main` worktree, none in the changed files.
- 250-file selection (every test naming the ray surfaces, the shared boolean domain, `tool_spec` or `describe()`, plus all of `tests/simulation/mujoco/` and the repo-wide docs/docstring/string guards), run with the changed test file excluded from **both** trees so the comparison measures collateral damage only:

  | tree | result |
  | --- | --- |
  | this branch | 6221 passed, 125 skipped, 0 failed |
  | pristine `upstream/main` | 6221 passed, 125 skipped, 0 failed |

  Byte-identical summary lines, zero failures on either side, so there is no passed-delta arithmetic to reconcile.
- The changed test file itself: 71 passed.
- `changelog.d/2639-...`: `scripts/assemble_changelog.py --check` OK, and the 85 fragment-gate tests pass.

## Artifact

A 72-ray fan from one origin, MuJoCo headless. A static wall rings the robot; one movable crate sits inside it. The caller asks the batch surface for the movable scene only.

static filter on both ray surfaces

- `upstream/main`: the parameter does not exist on that surface, so no fan is cast.
- this branch: **6 of 72** bearings answered, all by the movable crate.
- Control, top-down: the same fan with the filter omitted answers **72 of 72** - the wall on every bearing - and the 72 distances are byte-identical on both trees.
- Control: `raycast(include_static=False)` returns `crate_geom @ 0.6278648857607881` on both trees. The single-ray surface is unchanged, and that is the answer the fan could not produce.

Both scene renders and every control are asserted from the captured JSON inside the figure script, so the picture cannot drift from these numbers.

### DIFF
--- a/changelog.d/2639-static-filter-on-both-ray-surfaces.md
+++ b/changelog.d/2639-static-filter-on-both-ray-surfaces.md
@@ -1,0 +1,5 @@
+### Fixed: the static-geom filter is expressible on both ray surfaces, and checked on both
+
+`include_static` decides whether a geom on a body with no degrees of freedom - the ground plane, a wall, a fixed fixture - can be the answer to a clearance question. `raycast` exposed it; `multi_raycast` hardcoded the include, so a lidar fan could not ask the question a single ray can ask, and passing the flag refused the call as an unknown parameter even though the tool schema publishes `include_static` for both actions. A 72-ray fan around a walled scene holding one movable crate answered 72 of 72 bearings with the wall where the caller had asked for the movable scene only; it now answers the 6 the crate occupies.
+
+Both surfaces now hold the flag to the shared boolean domain the module already applies to `set_joint_positions`' `hold`. Read by truthiness the filter inverted for exactly the spellings a caller opting out reaches for, since `"false"`, `"no"` and `"0"` are all truthy: each of them selected the include it names the opposite of. A boolean caller is unaffected and a caller who omits the flag gets the identical fan; a non-boolean value is now refused by name.
--- a/strands_robots/simulation/mujoco/physics.py
+++ b/strands_robots/simulation/mujoco/physics.py
@@ -928,7 +928,14 @@ class PhysicsMixin:
                 model defines - an id outside ``[0, model.nbody)`` matches no
                 body, so the geoms the caller asked to skip would be included
                 and could be reported as the obstacle.
-            include_static: Whether to include static geoms.
+            include_static: Whether the ray sees static geoms (a geom on a body
+                with no degrees of freedom - the ground plane, a wall, a fixed
+                fixture). ``False`` casts against the movable scene only, so a
+                clearance check can ask about dynamic obstacles without the
+                static world answering first. Must be a boolean: read by
+                truthiness this flag inverts for exactly the spellings a caller
+                opting out reaches for, since ``"false"``, ``"no"`` and ``"0"``
+                are all truthy.
         """
         if self._world is None or self._world._model is None or self._world._data is None:
             return {"status": "error", "content": [{"text": _NO_WORLD_MSG}]}
@@ -974,6 +981,15 @@ class PhysicsMixin:
         if err is not None:
             return err
 
+        # The static filter is a posture flag, so it is held to the shared boolean
+        # domain rather than read by truthiness (the rule ``set_joint_positions``
+        # states for ``hold``): "false", "no" and "0" are all truthy, so every
+        # string spelling of an opt-out would select the include it asks to skip.
+        # Resolved once here so the batch entry point can apply the same value.
+        if text := boolean_flag_error(include_static, "include_static", "raycast"):
+            return {"status": "error", "content": [{"text": text}]}
+        static_filter = 1 if include_static else 0
+
         pnt = np.array(origin_f, dtype=np.float64)
         vec = np.array(direction_f, dtype=np.float64)
         # Normalize direction
@@ -1002,7 +1018,7 @@ class PhysicsMixin:
                 pnt,
                 vec,
                 None,  # geom group filter (None = all)
-                1 if include_static else 0,
+                static_filter,
                 exclusion,
                 geomid,
             )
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Brings `multi_raycast` to parity with `raycast` on the third ray semantic: `include_static` becomes a parameter of the batch surface with the same name, type, meaning and a default (`True`) that reproduces the previously hardcoded include, so an unchanged caller gets a byte-identical fan. Both surfaces now hold the flag to the shared `boolean_flag_error` domain the module already applies to `set_joint_positions`' `hold`, closing the truthiness inversion where `"false"`/`"no"`/`"0"` selected the include they name the opposite of. Since the dispatch layer derives valid parameters from the method signature, the already-published `include_static` tool-spec property becomes accepted on the `multi_raycast` action with no router change. Verified locally at the head SHA: the extended test file passes 71/71 (including the agent-surface and precedence controls) on mujoco 3.8.0, ruff check/format clean on the changed files, and the changed strings are ASCII-only with no host paths.

## What's good

- The narrowing is loud, not silent: a non-boolean that previously selected a branch by truthiness is now refused by name via an error dict, matching the AGENTS.md posture-flag rule and the facade/method-parity guidance (the same guard runs on both the direct method and the dispatched action).
- Changelog fragment present (`changelog.d/2639-static-filter-on-both-ray-surfaces.md`), correctly named for the PR.
- The regression tests parametrize the non-boolean spellings and pin both surfaces independently (the fault-injection table in the description shows each guard fires in isolation), and the precedence control pins that the vector refusal still outranks the filter refusal for existing callers.
- Lock discipline untouched: the flag is resolved before the sim lock, and the `mj_kinematics` + batch cast stay under it.

PR #2640, opened 2026-08-23
fix(sim): a DOF-indexed payload names the joint each column belongs to

cagataycali:
## Problem

`get_jacobian` and `get_mass_matrix` report arrays indexed by MuJoCo's DOF ordering - a Jacobian's columns, the mass matrix's diagonal - and reported only the width (`nv`, `shape`). A caller cannot reconstruct that ordering from the width, for two reasons that both arise in ordinary scenes:

- **A free or ball joint owns several consecutive DOFs**, so `nv` exceeds the joint count. Measured over the registry: **36 of the 62 sim robots that load here have `nv != njnt`**, and the gap reaches +20 (`lekiwi` njnt=13 nv=33; `stretch3` 23/38; `unitree_go2` 13/18).
- **`nv` spans the whole compiled model**, so a scene holding two robots reports one width covering both and one robot's columns are an interior slice.

The third DOF-indexed query in the same mixin, `inverse_dynamics`, already answers this - it returns `qfrc_inverse` as a mapping keyed by joint name. The other two did not, so the same payload was self-describing on one query and not on the other two.

## Change

Both payloads now carry `dof_joint_names`: a list of length `nv` naming the joint that owns each entry, in DOF order, with `None` where a joint is unnamed (MJCF permits a bare `<freejoint/>`, and dropping those columns would break the index alignment the labels exist to give). A private `_dof_joint_names` helper reads `model.dof_jntid`, which is MuJoCo's own DOF-to-joint map.

No number changes: `jacp`/`jacr` and the mass diagonal are byte-identical, and the pre-existing keys are untouched. `inverse_dynamics` is left exactly as it was - it already named its entries.

Docstrings, the `describe()` signature strings and the `docs/simulation/overview.md` argument table now state that the columns are whole-model DOFs and point at the new key, because the previous text let a reader take `nv` for the robot's joint count.

## Measured

Two 2-joint arms in one scene, so `nv = 4` spans both, and `get_jacobian(site_name="arm1/tip")`:

| measured | before | after |
| --- | --- | --- |
| payload keys | `jacp`, `jacr`, `nv` | + `dof_joint_names` |
| `nv` | 4 | 4 |
| non-zero columns (arm1's own) | `[2, 3]` | `[2, 3]` |
| `robot_joint_names("arm1")` | `['j1', 'j2']` | `['j1', 'j2']` |
| the columns those names address | not stated | `arm1/j1`, `arm1/j2` |

A caller pairing the robot's two joint names with columns `[0, 1]` takes arm0's columns and is told nothing.

## Artifact

The consequence is a real Jacobian IK step. One capture script runs the identical loop against each tree and selects columns from whatever that tree's payload offers, so nothing is hardcoded per side: with `dof_joint_names` the caller picks the robot's own columns; without it the only thing stated is the width, so it pairs the robot's joint names with the leading ones.

DOF-indexed Jacobian columns named by joint

| run | route | tip error | joint travel | rendered subject travel |
| arm1, before | leading columns | 0.1476 -> **0.1476 m** | **0.0000 rad** | **0.0 px** |
| arm1, after | columns named by `dof_joint_names` | 0.1476 -> **0.0098 m** | 2.3719 rad | 48.9 px |
| arm0 (control), before | leading columns | 0.1476 -> 0.0098 m | 2.3719 rad | 53.9 px |
| arm0 (control), after | columns named | 0.1476 -> 0.0098 m | 2.3719 rad | 53.9 px |

The same loop converges on the first robot and moves the second **not at all**: arm0's columns are the leading ones, so the naive mapping happens to be right there, while for arm1 the leading columns are arm0's, which are zero in arm1's Jacobian, so `pinv` returns a zero step. The arm0 rows are identical across trees field for field (tip error, joint travel and final joint values), which is the no-behaviour-change control.

Before ([mp4](https://raw.githubusercontent.com/cagataycali/robots/media-dof-joint-names-32609231541/arm1-before.mp4)) and after ([mp4](https://raw.githubusercontent.com/cagataycali/robots/media-dof-joint-names-32609231541/arm1-after.mp4)) for the arm1 run.

## Tests

`tests/simulation/mujoco/test_dof_indexed_payloads_name_their_columns.py`. Against `upstream/main`: **5 failed / 5 passed**, four of the five behavioural with a measured message, e.g.

```
get_jacobian reported a 3x4 Jacobian for 'arm1/tip' with only nv=4. Its non-zero
columns are [2, 3], which belong to arm1, while robot_joint_names('arm1') is
['j1', 'j2'] - so a caller pairing those names with columns [0, 1] reads arm0's
Jacobian and is told nothing.
```

The five that pass on both trees are the controls: `jacp`/`jacr` still equal a direct `mj_jacSite`, the mass matrix keeps its existing keys, `inverse_dynamics` still returns a name-keyed mapping, and a single-robot scene still reports its own width.

`TestEveryReadOnlyQuerysDofIndexedPayloadIsLabelled` derives its population from the mixin - every public `get_*`/`inverse_*`/`forward_*` with no required parameter - and requires any DOF-indexed value (a flat list of `nv` numbers, or rows each `nv` wide) to be accompanied by the labels, so a query added later is held to the same rule.

Each guard was checked by breaking it, and each break fires a distinct set:

| break | fires |
| --- | --- |
| exact pre-fix (`git checkout upstream/main -- physics.py`) | all 5 |
| drop the key from `get_jacobian` only | the 3 Jacobian tests + the sweep |
| drop the key from `get_mass_matrix` only | the diagonal test + the sweep |
| helper skips unnamed joints | the unnamed-column test only |
| one name per joint instead of per DOF | free-joint + unnamed |
| labels emitted in the wrong order | headline + free-joint + diagonal |
| sweep narrowed to flat lists (+ Jacobian reverted) | the sweep, via its non-vacuity clause |
| the same, with non-vacuity relaxed | the sweep now **passes** - the matrix branch is what catches a matrix-shaped offender |
| the query scan reaches nothing | the sweep + its scan-reach pin |
| control | none of 10 |

## Verification

Measured at `a4894d6` (`upstream/main` merged in; the merge is clean and both sides were verified by symbol). mujoco 3.12.0, headless EGL.

- Identical 237-file selection on both trees, this PR's new test file excluded from each: `5 failed, 4974 passed, 147 skipped` **byte-identical**, failing-ID sets identical, `comm` empty in both directions. The 5 shared are local dependency artifacts, not code: 4 in `test_docs_device_connect_env_reference.py` (`ModuleNotFoundError: No module named 'device_connect_edge'`, a declared extra CI installs) and `test_pose_vector_rule_parity.py::test_a_numpy_array_pose_is_accepted_by_both`, which fails on this box's numpy on both trees.
- `tests/simulation/mujoco` on the merged tree: `1 failed, 3848 passed` - that same numpy case.
- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors, identical to the pristine base, none in the changed files.

## Scope

Only the two DOF-indexed payloads in `PhysicsMixin` that stated a width and nothing else. `inverse_dynamics` already names its entries and is untouched; `get_contact_forces`, `forward_kinematics` and `get_sensor_data` are keyed by name already or are not DOF-indexed, and the sweep confirms it rather than assuming it.

Isaac and Newton are out of scope: neither implements `get_jacobian` or `get_mass_matrix`.


yinsong1986 (approved):
## Summary

Adds a `dof_joint_names` key to the `json` payloads of `get_jacobian` and `get_mass_matrix`, a list of length `nv` naming the joint that owns each DOF column (with `None` preserved for unnamed joints so index alignment survives), bringing the two queries onto the same self-describing terms as `inverse_dynamics`. The change is purely additive - all pre-existing keys and numbers are untouched, which the new controls pin against a direct `mj_jacSite` - and the helper reads only `mjModel` fields (`dof_jntid` via `mj_id2name`), so it introduces no new `mjData` read outside the existing `self._lock` sections. Docstrings, `describe()` strings, the docs table and a changelog fragment all landed in the same PR, and the new test file derives its sweep population from the mixin so a future DOF-indexed query is held to the same rule.

## What's good

- The sweep (`TestEveryReadOnlyQuerysDofIndexedPayloadIsLabelled`) is derived from the mixin rather than a hardcoded list, with a non-vacuity clause, matching the repo's derive-the-population convention.
- Changelog fragment per AGENTS.md step 3; no direct `CHANGELOG.md` edit.
- The two `\u00d7` glyphs in the touched docstrings were replaced with ASCII `x` while passing through.
- Scope discipline: `inverse_dynamics` (already name-keyed) untouched; Isaac/Newton correctly out of scope since neither implements these queries.

PR #2641, opened 2026-08-23
test(refusals): a code is graded in every spelling that reaches it

cagataycali:
## Problem

`test_every_code_passed_at_a_raise_site_is_a_declared_code` is the only thing that
holds `SecurityError.code` to `refusal_codes.REFUSAL_CODES`. It found the codes to
grade with

```python
if keyword.arg == "code" and isinstance(keyword.value, ast.Attribute):
```

so it read the value in one spelling: an attribute of the codes module. Two
asymmetries inside that same file:

- its sibling `_refusals_naming_an_env_var` reads the **presence** of the very same
  keyword spelling-independently (`any(k.arg == "code" for k in exc.keywords)`), and
  `test_a_coded_refusal_offering_a_new_grant_is_accepted` deliberately accepts a code
  the module does not even declare -- *because grading the value belongs to the rule
  above*;
- `test_every_code_is_exported_under_its_own_name` pins that every code is exported
  under its own name **and listed in `__all__`**, i.e. the file asserts that importing
  a code directly is a first-class way to reach it, while the helper that validates
  its use can only see it reached through the module object.

Nothing validates the value at runtime and nothing should: `SecurityError.__init__`
assigns `code` as given, and raising from an exception constructor would replace a
security refusal with a constructor error. That scan is the only thing between a
mistyped code and a consumer.

Measured. Adding two plausible gates to `main` -- one spelling the code as a
from-imported name, one as a string literal with a typo -- and running the shipped
guard:

| tree | shipped guard, whole file | shipped guard, the rule |
| --- | --- | --- |
| `main` + those two gates | 25 passed | 1 passed |

The typo'd code ships. Driving the consumer pattern from `docs/security.md` verbatim:

| refusal | `code` | the documented consumer |
| --- | --- | --- |
| repo outside the allowlist (control) | `'HF_REPO_NOT_ALLOWED'` | `offer_to_allowlist('sketchy/model')` via `STRANDS_MESH_HF_REPO_ALLOW` |
| the planted gate | `'POLICY_MODE_NOT_ALOWED'` | raises `KeyError: 'POLICY_MODE_NOT_ALOWED'` |

`REFUSAL_GRANTS[refusal.code]` is the lookup the docs page shows, and an undeclared
code turns it into a `KeyError`. The refusal is also invisible to a `code ==` switch,
so the offer the code exists to enable disappears and the consumer is back to reading
prose.

What the old scope did and did not catch, measured one change at a time:

| change to a shipped site | shipped guard |
| --- | --- |
| an attribute-form typo (`refusal_codes.HF_REPO_NOT_ALOWED`) | 3 failed |
| the **sole** site of a code -> from-import form | 1 failed, but from `test_every_declared_code_is_actually_raised_somewhere` reporting the code as never raised -- a spelling change misdiagnosed as a missing code |
| **one of two** sites of a code -> from-import form | 25 passed |
| a new gate, from-import form | 25 passed |
| a new gate, string literal with a typo | 25 passed |

## Change

The scan reads a `code=` value in every spelling that reaches a declared code, and
reports one it cannot read instead of skipping it:

- a `refusal_codes` attribute (as before);
- a name imported from the codes module, resolved through that module's own import
  list so `import HF_REPO_NOT_ALLOWED as REPO_CODE` resolves to the code it aliases
  rather than to `REPO_CODE`, which this package does not declare;
- the literal string. `test_every_code_is_exported_under_its_own_name` is what makes
  these one channel: each code is exported under its own name, so all three spellings
  are that same name.

A value that is not statically readable -- a parameter forwarded by a shared helper, a
table lookup -- is reported rather than skipped, because a code nothing can check is a
code that may not be in the vocabulary at all. The message names both remedies.
There are no such sites today.

The rule itself is now driven over sources (`_undeclared_codes_in(_package_sources())`)
the way its sibling `_uncoded_refusals_naming_an_env_var` already is, so it can be
planted against. Its assertion is unchanged and it now names the offending value.

All **7** shipped sites use the attribute form, so the widening grades exactly the same
7 sites and changes no verdict on `main`. `strands_robots/refusal_codes.py` gains a
paragraph on the second dimension of the guard's scope, next to the paragraph that
already explains why the scope is not the grants table; `docs/security.md` states what
backs the closed-vocabulary promise a consumer switches on.

No visual artifact: the package diff is the module docstring only -- verified
byte-identical below it -- so no policy, sim behaviour, rendering, recording or asset
changes.

## Tests

The fix is the guard, so the regression evidence is the plant above: the shipped guard
reports clean on a tree carrying an undeclared code, the new one fails naming it.

| tree | new guard, whole file | new guard, the rule |
| --- | --- | --- |
| `main` + those two gates | 3 failed, 32 passed | 1 failed |

The failure names the site: `strands_robots/mesh/security.py:1619 passes
code='POLICY_MODE_NOT_ALOWED' -> 'POLICY_MODE_NOT_ALOWED'`.

Every new guard is load-bearing, one break at a time:

| break | fires |
| --- | --- |
| control (unbroken branch) | 0 of 35 |
| `git checkout upstream/main -- <the guard>` | 25 passed: nothing grades a spelling |
| drop the `ast.Name` branch | 3 -- the imported-name read, its typo, and the alias test (which needs that branch) |
| drop the string-literal branch | 2 -- the literal read and its typo |
| an alias resolves to the bound name, not the code | **1** -- the alias test alone |
| an unreadable spelling is skipped, not reported | **1** -- the unreadable test alone |
| the scan reaches nothing | 11, including the non-vacuity pin and the pre-existing `test_every_declared_code_is_actually_raised_somewhere` -- the widened helper still feeds its original consumer |

## Verification

`tests/test_refusal_codes_are_the_stable_contract.py`: **25 -> 35 passed**
(`upstream/main` vs this branch). 154 passed alongside the two sibling refusal
files with random ordering enabled, so the new class is order-independent.

The identical 290-file selection on both trees -- every test naming
`refusal_codes` / `REFUSAL_CODES` / `REFUSAL_GRANTS` / `UntrustedRemoteCodeError`,
every guard that walks a source tree or grades docstrings and strings, and the
repo-wide docs and changelog guards -- with the changed guard excluded from both
so the run measures collateral damage only:

| tree | result |
| --- | --- |
| this branch | 17 failed, 13819 passed, 58 skipped in 473.81s |
| `upstream/main` | 17 failed, 13819 passed, 58 skipped in 474.03s |

Byte-identical, and the failing-id sets match with `comm` empty in both
directions. All 17 are `tests/mesh/test_iot_*` on
`ModuleNotFoundError: No module named 'awsiot'` (34 occurrences) -- the
`[mesh-iot]` extra, which declares `awsiotsdk>=1.21.0,<2.0.0`, so they pass where
it is installed.

tests_integ`; `mypy` over the same reports 24 errors on this branch and 24 on
`upstream/main`, with the branch-only set empty and none in a changed file.
Measured at `4a96f11d`.


yinsong1986 (approved):
## Summary

Widens the static guard in `tests/test_refusal_codes_are_the_stable_contract.py` so a `code=` argument at a raise site is graded in every spelling that can reach a declared refusal code -- a `refusal_codes` module attribute (the old scope), a name imported directly from the codes module (with `as` aliases resolved to the code they alias), and the literal string -- and so a value the scan cannot read statically is reported instead of silently skipped. The package diff is confined to the `refusal_codes` module docstring, `docs/security.md` gains the closed-vocabulary consumer promise, and a changelog fragment is included per AGENTS.md step 3. No runtime behaviour changes: all seven shipped sites use the attribute form, so the widened scan grades the same set and changes no verdict on `main`.

## What's good

- The regression is demonstrated, not asserted: the PR plants a from-imported and a string-literal typo'd code against the shipped guard (25 passed) and the new one (fails naming the site), and each new branch of the scan is shown load-bearing one break at a time.
- The unreadable-spelling case fails loudly (`test_a_code_this_scan_cannot_read_is_reported_not_skipped`) rather than skipping, which is the correct failure direction for a guard whose only enforcement is static -- a heuristic miss surfaces as a test failure, never as a shipped undeclared code.
- Verified locally at `4a96f11d`: the changed file reports 35 passed, and `test_the_shipped_sites_are_all_read` pins that nothing in the package is unreadable today, so the widening is a no-op on the shipped tree.
- Changelog fragment, ASCII-only strings, and behaviour-describing test names all follow AGENTS.md conventions.

Issue #2642, opened 2026-08-23
two merged examples still promise the fleet dashboard "once it lands" - it landed in #2198, and no gate reads issue state to notice


cagataycali:
## The claim

Prose that says "X has not shipped yet" or "once it lands" is the only kind of documentation whose truth changes **without the file being touched**. Every drift check in the tree compares two things that are both *in* the tree - a docstring against an enum (#2138), a lockfile against `requires-dist` (#2041), prose against a default (#2192). This claim's other half is the state of an issue, which lives outside the tree, so nothing can currently see it go stale.

Measured on `main` at `12a7d06b`. Four forward-looking claims exist in merged prose. **Two are false.**

## The two false ones

Both promise the read-only Rerun fleet dashboard as future work:

```
examples/fleet/03_failover_and_degraded_ops.py:34
  The read-only Rerun fleet dashboard (#2181) attaches to
  the same mesh and shows the peer loss and the recovery live once it lands.

examples/fleet/04_emergency_evacuation.py:48
  the read-only Rerun fleet dashboard (#2181) attaches to the same mesh and
  shows the safety events live once it lands.
```

It landed:

| fact | value |
|---|---|
| artifact | `examples/fleet/dashboard.py` - "Read-only fleet dashboard: a subscribe-only mesh peer rendered in Rerun" |
| added by | `ae57b3fa` - `feat(examples/fleet): 02_cross_zone_transport + read-only Rerun fleet dashboard (#2198)` |
| issue #2181 | **CLOSED** 2026-08-19T04:48:25Z, title matches the PR exactly |

So two examples in the same directory as `dashboard.py` tell the reader it does not exist yet. The reader most likely to hit this is the one working through the fleet suite in order - they are in `examples/fleet/`, the file is sitting next to the one they are reading, and the docstring tells them to wait for it.

## The two that are still true, which is the point

A check keyed on this would not be noisy. The other two forward-looking claims are accurate, and for the same reason the false ones are false - the referenced work has not landed:

```
examples/libero/README.md:81   "(run_isaac_fleet.py) ... until it lands"
  -> examples/libero/run_isaac_fleet.py: absent. TRUE.

examples/fleet/04_emergency_evacuation.py:46   "the Isaac adapter (#2123) drops in later"
  -> #2123: OPEN. TRUE.
```

2 of 4 false, 2 of 4 true, and the discriminator is mechanical in both directions. That is a gate with precision, not a lint that has to be suppressed.

[...]

Issue #2123, opened 2026-08-11
feat(simulation/isaac): motion primitives parity - move_to / set_gripper / rotate_wrist


yinsong1986:
# Gap

The motion primitives - `move_to`, `set_gripper`, `rotate_wrist` - live in `MotionPrimitivesMixin` (`strands_robots/simulation/mujoco/motion_primitives.py:124`; `move_to` at `:457`, `set_gripper` at `:759`, `rotate_wrist` at `:895`) and are mixed into the MuJoCo simulation only. On Isaac the only motion path is raw `set_joint_positions` (`strands_robots/simulation/isaac/simulation.py:5330`): no blocking joint-space move with timeout and abort reasons, no registry-metadata-driven gripper semantics, no wrist primitive.

These primitives were introduced in #1645 / #1654 together with the shared `MinkIKBridge` (`strands_robots/simulation/ik.py`), which was deliberately built as a shared component - the kinematics layer is already backend-neutral.

# Proposal

Split the mixin into:

1. **Backend-agnostic core** (candidate home: `strands_robots/simulation/motion_primitives_base.py`): robot/target resolution, gripper setpoint ranges from registry metadata, the tick/timeout/abort-reason contract, input validation. This is the majority of the current mixin's logic and none of it touches `mjModel`/`mjData` directly.
2. **Per-backend actuation adapter**: MuJoCo keeps its current actuator-map path; the Isaac adapter drives the articulation controller (position targets + convergence polling on the Kit loop).

Public API and docstring semantics must be identical across backends (AGENTS.md "Match docstrings to semantics"). If the primitives are surfaced through agent tools, every parameter needs its own `Args:` entry (AGENTS.md rule 13).

**Verify, don't assume**: `MinkIKBridge` operates on the MJCF model, and Isaac registry robots carry MJCF sources, so IK-backed end-effector targets may transfer nearly free - but the Isaac articulation's joint ordering/namespacing must be reconciled with the MJCF-side solution before claiming parity.

# Dependencies / relations

- Sibling: #2122 (run_multi_policy Isaac parity) - independent implementations, one shared motivation.
- #1895 (`reset()` kills articulation handles): not a blocker for the primitives themselves, but parity tests that reset between cases must account for it.
- Newton has the same gap; scope stays Isaac, Newton noted as follow-up.

# Acceptance

- Same-named tests under `tests/simulation/isaac/` mirroring the MuJoCo primitive tests (GPU-skipped in unit CI).
[...]

PR #2643, opened 2026-08-23
docs(examples/fleet): the fleet dashboard is named, not promised

cagataycali:
Towards #2642 (the immediate-fix half; the gate stays open there, and see "Why there is no test" for why it cannot be offline).

## The defect

Two examples in the fleet suite describe the read-only Rerun fleet dashboard as future work:

```
examples/fleet/03_failover_and_degraded_ops.py:34
  The read-only Rerun fleet dashboard (#2181) attaches to
  the same mesh and shows the peer loss and the recovery live once it lands.

examples/fleet/04_emergency_evacuation.py:48
  the read-only Rerun fleet dashboard (#2181) attaches to the same mesh and
  shows the safety events live once it lands.
```

It landed four days ago:

| fact | value |
|---|---|
| artifact | `examples/fleet/dashboard.py` - "Read-only fleet dashboard: a subscribe-only mesh peer rendered in Rerun" |
| added by | `ae57b3fa` - `feat(examples/fleet): 02_cross_zone_transport + read-only Rerun fleet dashboard (#2198)` |
| issue #2181 | CLOSED 2026-08-19T04:48:25Z, title matching the PR exactly |

The reader who hits this is the one working through the fleet suite in order: they are inside `examples/fleet/`, `dashboard.py` is sitting next to the file they are reading, and its docstring tells them to wait for it.

## The change

Both docstrings name the artifact in the present tense instead of promising a ticket:

> The read-only Rerun fleet dashboard (``examples/fleet/dashboard.py``) attaches to the same mesh and shows the safety events live.

That is also the spelling that cannot go stale again, which is the reason to prefer it over merely flipping the tense: prose that names a path is checkable against the tree, while prose that names a ticket depends on state no in-tree check can read.

`04_emergency_evacuation.py`'s other forward reference - "the Isaac adapter (#2123) drops in later" - is deliberately **unchanged**: #2123 is still open, so that sentence is still true. This PR corrects what is false and leaves what is accurate alone.

## Why there is no test

Docstring-only, and I can state that precisely rather than assert it. For both files, with the module docstring removed the AST is identical to `main`, and every changed byte lies inside the docstring:

```
examples/fleet/03_failover_and_degraded_ops.py
  AST identical with module docstring removed : True
  every changed byte lies inside the docstring: True
examples/fleet/04_emergency_evacuation.py
  AST identical with module docstring removed : True
  every changed byte lies inside the docstring: True
```

I did try to ship a pin with this, and I am reporting the result because it changes the design on #2642 rather than being something I quietly dropped. I wrote a corpus-wide offline check - find forward-looking prose, resolve the path it names, fail if that path already exists - and ran it against the tree. It produced **two false positives and could not see the actual defect**:

1. `examples/libero/README.md` - the sentence names two paths, `libero_backend_matrix.py` (exists; it is the thing doing the printing) and `run_isaac_fleet.py` (absent; the thing actually promised). Proximity picks the wrong one.
2. `04_emergency_evacuation.py` - "drops in later" refers to the Isaac adapter **#2123**, while the nearest path token belongs to a different clause across a semicolon.
3. Decisive: the drifted sentences named **#2181 - a ticket, not a path**. A path-resolving check is structurally incapable of catching the shape that motivated it.

So the discriminating signal for this defect class is issue state, and a check for it cannot be a unit test. Shipping the offline version would have added a grader whose corpus is clean and which cannot detect the bug it was written for. Recorded on #2642 so the gate is scoped correctly there.

## Gates

- `ruff check` and `ruff format --check` on both files: clean.
- `scripts/assemble_changelog.py --check`: `changelog fragments OK`.
- Changelog recorded as a fragment (`changelog.d/2642-...md`), not appended to `CHANGELOG.md`.
- No runtime surface: `git diff --name-only main...HEAD -- strands_robots/ tests/` is empty.

## Round changelog

### DIFF
--- a/changelog.d/2642-fleet-examples-name-the-shipped-dashboard.md
+++ b/changelog.d/2642-fleet-examples-name-the-shipped-dashboard.md
@@ -1,0 +1,11 @@
+### Fixed: two fleet examples no longer promise a dashboard that already ships
+
+`examples/fleet/03_failover_and_degraded_ops.py` and
+`04_emergency_evacuation.py` described the read-only Rerun fleet dashboard as
+future work ("shows the safety events live once it lands", referencing #2181),
+but it landed in `examples/fleet/dashboard.py`. A reader working through the
+fleet suite in order was told to wait for a file sitting in the same directory.
+Both docstrings now name the artifact in the present tense, which is also the
+spelling that cannot go stale again: prose naming a path stops depending on
+issue state. The still-accurate forward reference to the Isaac adapter (#2123,
+open) is unchanged. (#2642)
--- a/examples/fleet/03_failover_and_degraded_ops.py
+++ b/examples/fleet/03_failover_and_degraded_ops.py
@@ -30,8 +30,9 @@ Note: The live estop drill leaves the surviving robots in safety lockout by
       Set STRANDS_MESH_LOCAL_DEV=1 (defaulted below) to skip TLS locally.
 
 Part of the fleet suite (epic #2179); the shared manifest schema lives in
-``capabilities.py``. The read-only Rerun fleet dashboard (#2181) attaches to
-the same mesh and shows the peer loss and the recovery live once it lands.
+``capabilities.py``. The read-only Rerun fleet dashboard
+(``examples/fleet/dashboard.py``) attaches to the same mesh and shows the peer
+loss and the recovery live.
 """
 
 from __future__ import annotations
--- a/examples/fleet/04_emergency_evacuation.py
+++ b/examples/fleet/04_emergency_evacuation.py
@@ -44,8 +44,8 @@ Note: The live lockout drill needs `STRANDS_MESH_OVERRIDE_CODE`; when unset,
 
 Part of the fleet suite (epic #2179). The retreat sits behind the small
 ``EvacuationWorld`` seam so the Isaac adapter (#2123) drops in later; the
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

Docstring-only correction in two fleet examples: both described the read-only Rerun fleet dashboard as future work ("once it lands", referencing #2181), but the dashboard shipped as `examples/fleet/dashboard.py` four days ago. The PR rewrites both sentences in the present tense and names the in-tree path instead of the closed ticket, which is the spelling that can be checked against the tree and so cannot silently go stale again. Verified against the head SHA: `examples/fleet/dashboard.py` exists on `main`, every changed byte in both `.py` files lies inside the module docstring (no runtime surface), the still-open Isaac adapter reference (#2123) is correctly left untouched, and the changelog is recorded as a `changelog.d/` fragment per AGENTS.md step 3 rather than appended to `CHANGELOG.md`.

## What's good

- Tight scope discipline: corrects exactly the two false sentences and leaves the still-accurate #2123 forward reference alone.
- The description reports a rejected test design (path-resolving offline check) with measured false positives rather than shipping a grader that cannot see the defect class it was written for, and records the finding on #2642 where the gate lives.
- Changelog fragment convention followed; prose in the changed docstrings is plain ASCII.

PR #2645, opened 2026-08-23
docs(mesh): the command census names every action it validates a field for

cagataycali:
## Problem

`validate_command` documents its per-action work under a `Performed checks:` census. Six action
branches validate at least one field out of `cmd`; the census named four of them, and for one of
those four it described the check inaccurately.

Derived from the body's own `if action == "x"` chain and its `cmd` reads:

| action branch | fields it validates | census names the action | names every field | names the shared validator |
| `execute` | 18 | yes | yes | n/a |
| `start` | 18 | yes | yes | n/a |
| `step` | 1 | yes | yes | n/a |
| `teleop_receive` | 2 | yes | **no** (`device_name` absent) | **no** |
| `teleop_stop` | 1 | **no** | **no** | n/a |
| `resume` | 1 | **no** | **no** | n/a |

Three of the six branches were incompletely named. The inaccuracy in the fourth is the sharpest of
the four findings: the census said `source_peer_id` must be "a non-empty str", where the body calls
`validate_mesh_identifier`. Measured, that validator refuses the Zenoh wildcards `*` and `**`, `/`,
`?`, `#`, `$`, control characters, and anything over `MAX_PEER_ID_LEN` -- all of which the census's
own words admit. So a reader who sizes an allowlist entry or a fixture from the census picks a value
the validator rejects, and the rejection names a charset the census never mentioned.

`resume.override_code` is the omission with the most weight behind it. It is the operator's second
factor for clearing an e-stop lockout, bounded at 256 printable-ASCII characters, and the census
named neither the action nor the field.

## Change

Extend the census to name all six branches, every field each one validates, and the shared validator
the two teleop identifiers are decided by. `source_peer_id` and `device_name` now say they are
checked by `validate_mesh_identifier`, so the admitted charset is discoverable from the census rather
than by reading the body.

`validate_command`'s docstring is the entire package diff. With that docstring removed from both
versions the two files are byte-identical, so no verdict, message or accepted input moves.

## Tests

`tests/mesh/test_performed_checks_census_matches_the_code.py` is the existing guard for this census
(it already grades the value-domain bullets against the validator). It gains the per-action
completeness rule, with both halves derived rather than listed:

- the action set comes from the body's own `if action == "x"` / `in (...)` chain;
- the field set per branch comes from every way the body reads `cmd` -- `.get`, `.pop`, `cmd["x"]`
  and `"x" in cmd`;
- a field handed to a `validate_*` helper as `"<action>.<field>"` records that helper as its
  delegate, read off the AST argument rather than the unparsed source.

A seventh action branch, or a new field on an existing one, is therefore graded on arrival.

Pre-fix split, running the updated guard against `upstream/main`: **3 failed / 54 passed**. All three
failures are semantic and name their own finding -- the missing actions, the missing field, and the
undocumented delegate.

Break table, each row the whole file:

| break | failures | which |
| --- | --- | --- |
| control (this branch) | 0 of 57 | - |
| exact pre-fix census | 3 | action + field + delegate rules |
| drop the `resume` / `teleop_stop` bullets | 2 | action + field |
| name `resume` but not `override_code` | 1 | field rule only |
| describe the identifiers without naming the validator | 1 | delegate rule only |
| field scan loses the `"x" in cmd` spelling | 1 | the read-spelling control only |
| delegate read off the unparsed source, not the AST | 2 | non-vacuity + the AST control (the rule itself passes vacuously) |
| the action scan reaches nothing | 5 | 3 non-vacuity + both derivation controls |

Seven breaks, seven distinct firing sets, no dead guards. The three single-firing rows are what show
each rule is independently pinned; the unparsed-source row is the real bug this scan hit while it was
being written, and it is caught by the non-vacuity clause rather than by the rule, which is why that
clause is not decoration.

## Verification

Measured at `a95ad651`, on `lerobot 0.6.2` / `mujoco 3.5.0`.

- `mypy strands_robots tests tests_integ`: 24 errors on this branch and 24 on a pristine
  `upstream/main` worktree, with the branch-only set empty.
- Blast radius: 410 test files (all of `tests/mesh`, every test naming `validate_command`,
  `validate_mesh_identifier`, `MAX_OVERRIDE_CODE_LEN`, `override_code`, `source_peer_id` or
  `device_name`, plus the repo-wide docstring and docs guards). The identical selection was run on
  this branch and on a pristine `upstream/main`, with the changed test file excluded from both:
  `23 failed, 13490 passed, 86 skipped, 24 errors` on each, in 404.07s and 400.58s. The summary
  lines are byte-identical and the 47 failing ids are the same set, with the branch-only and
  base-only sets both empty. All 47 are dependency-absent here and pass on CI: 45 under
  `tests/mesh/test_iot_*` on `ModuleNotFoundError: No module named 'awsiot'` (88 occurrences; the
  declared `[mesh-iot]` extra pins `awsiotsdk>=1.21.0,<2.0.0`), and the two
  `tests/test_doctor.py::TestDoctorVersionResolution` cases on
  `PackageNotFoundError: No package metadata was found for strands-robots`, which reproduce in
  isolation on a pristine `upstream/main` worktree because a worktree is not pip-installed.
- The updated guard on its own: 57 passed.

## Scope

The census's `Raises` and value-domain bullets are unchanged; only the per-action enumeration and the
`source_peer_id` / `device_name` descriptions move.

Two things measured and deliberately left alone. `execute` and `start` share one branch and one
18-field set, and the census already names both, so nothing there is per-action. And the census still
does not enumerate the field allowlist itself -- the set of keys `validate_command` will carry through
at all -- which is a much larger enumeration with its own drift question, and not this change.

## Artifact

The census is documentation, so there is no rendered behaviour to film. What the figure shows instead
is the completeness matrix above beside the verdicts the validator really returns for those fields,
captured by running both trees: the verdicts are identical in both columns, which is the control.

per-action census completeness

### DIFF
--- a/changelog.d/2645-command-census-per-action.md
+++ b/changelog.d/2645-command-census-per-action.md
@@ -1,0 +1,28 @@
+### Fixed: the command census names every action it validates a field for
+
+`validate_command` documents its per-action work under a `Performed checks:` census. Six
+action branches validate at least one field out of `cmd`; the census named four of them, and
+for one of those four it described the check inaccurately.
+
+`resume` and `teleop_stop` were absent entirely, so nothing in the census said that
+`resume.override_code` -- the operator's second factor for clearing an e-stop lockout -- is
+bounded at 256 printable-ASCII characters, or that `teleop_stop.device_name` is checked at
+all. `teleop_receive` was named but only for one of its two fields.
+
+The inaccuracy is the sharpest of the four. The census said `source_peer_id` must be "a
+non-empty str", where the body hands it to `validate_mesh_identifier`. Measured, that
+validator refuses the Zenoh wildcards `*` and `**`, `/`, `?`, `#`, `$`, control characters and
+anything over `MAX_PEER_ID_LEN` -- every one of which the census's own words admit. A reader
+sizing an allowlist entry or a fixture from the census therefore picks a value the validator
+rejects, and the rejection names a charset the census never mentioned.
+
+The census now names all six branches, every field each one validates, and the shared
+validator the two teleop identifiers are decided by, so the admitted charset is discoverable
+from the census rather than by reading the body.
+
+`validate_command`'s docstring is the whole package diff: with that docstring removed from
+both versions the two files are byte-identical, so no verdict, message or accepted input
+moves. The existing census guard gains the per-action completeness rule, with the action set
+derived from the body's own `if action == "x"` chain and the field set from every way the body
+reads `cmd` -- `.get`, `.pop`, `cmd["x"]` and `"x" in cmd` -- so a seventh branch, or a new
+field on an existing one, is graded on arrival.
--- a/strands_robots/mesh/security.py
+++ b/strands_robots/mesh/security.py
@@ -947,7 +947,19 @@ def validate_command(cmd: dict[str, Any]) -> dict[str, Any]:
         - ``fast_mode`` (optional): boolean.
         - ``n_steps`` (optional): integer in ``[1, 10_000_000]``.
     * ``step``: ``steps`` integer in ``[1, 10_000]``, defaults to 1.
-    * ``teleop_receive``: ``source_peer_id`` non-empty str.
+    * ``teleop_receive``: both identifiers are checked by
+      :func:`validate_mesh_identifier`, so the admitted charset is
+      ``[A-Za-z0-9_.-]+`` rather than any non-empty string -- a Zenoh
+      wildcard is refused instead of silently widening the subscription
+      the follower builds out of them.
+        - ``source_peer_id``: REQUIRED.
+        - ``device_name`` (optional): defaults to ``"leader"`` downstream.
+    * ``teleop_stop``: ``device_name`` (optional) must be a str or null.
+    * ``resume``: ``override_code`` (optional, defaults to ``""``): str of at
+      most :data:`MAX_OVERRIDE_CODE_LEN` characters, printable ASCII only
+      (no C0/DEL/CRLF). The operator's second factor for clearing an e-stop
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Docs-plus-guard change: `validate_command`'s `Performed checks:` census gains bullets for the two action branches it never named (`resume`, `teleop_stop`), the missing `teleop_receive.device_name` field, and it now names `validate_mesh_identifier` as the delegate for the two teleop identifiers instead of the inaccurate "non-empty str" description. Verified against the body at head: the new bullets match the code exactly - `_PEER_ID_RE` is `[A-Za-z0-9_.-]+`, `MAX_OVERRIDE_CODE_LEN` is 256, `_SAFE_PASSTHROUGH_RE` is `^[\x20-\x7E]+$` (so "printable ASCII only, no C0/DEL/CRLF" is accurate), and `teleop_stop.device_name` is str-or-null. The package diff is confined to the docstring, so no verdict, message or accepted input moves. The extended census guard derives both the action set and the per-branch field set from the AST (all three `cmd` read spellings, plus the dotted `"action.field"` delegate label read off the AST rather than unparsed source), with non-vacuity and over-reach controls; all 57 tests pass at head, checked out and run locally.

## What's good

- Both halves of the guard are derived from the function body rather than listed, so a new action branch or field is graded on arrival - consistent with the existing conventions in this test module.
- Changelog fragment present per AGENTS.md step 3; no `CHANGELOG.md` edit.
- Behavioural tests (`TestTheActionRulesTheCensusNowStates`) pin the census's new claims against the validator's real verdicts, not just the prose.
- The description's break table documents seven distinct firing sets, so none of the new rules is a dead guard.

PR #2649, opened 2026-08-23
fix(tests): a module-scope MUJOCO_GL default must not name a windowed GL backend

cagataycali:
## Problem

MuJoCo has two *windowed* GL backends, `cgl` (macOS) and `glfw` (needs a window server), and two offscreen ones, `egl` and `osmesa`. Thirteen module-scope `MUJOCO_GL` defaults named `glfw`.

The two windowed backends fail differently, and `glfw` fails worse. `cgl` is refused at `import mujoco` with a message naming the variable. `glfw` is a valid value everywhere, so the import succeeds; the render probe then fails and the backend warns that rendering is unavailable, quoting `GLFWError: X11: The DISPLAY environment variable is missing` -- it names the missing display, but not the `MUJOCO_GL` value that asked for a window -- and camera observations are skipped from there on. The *failure* the caller finally gets is whatever it was doing with those frames.

Driving the shipped multi-robot recording path on a headless host with `MUJOCO_GL` unexported, using the default each tree's own test module sets:

| measured | `setdefault("MUJOCO_GL", "glfw")` | `"cgl" if darwin else "egl"` |
| --- | --- | --- |
| resolved backend | `glfw` | `egl` |
| `start_recording` | success, declares 4 camera features | success, declares the same 4 |
| `run_policy("alice")` / `("bob")` | error / error | success / success |
| `stop_recording` | error | success |
| `meta/info.json` total_frames | **0** | **12** |
| total_episodes | 0 | 1 |
| MP4 files written | **0** | **4** |
| frames decoded back out | **0** | **12** |

The refusal the caller gets is `RecordingFrameError: dataset add_frame failed after 0 frame(s) written ... Feature mismatch in \`frame\` dictionary: Missing features: {observation.images.top, ...default, ...alice__wrist_cam, ...bob__wrist_cam}` -- a dataset feature mismatch, several frames from the setting that caused it.

`tests_integ/simulation/test_multi_robot_tasks.py` is that path, and it went from failing to passing on that host.

## Why scope matters here

`MUJOCO_GL` is read once, at `import mujoco`. A **module-scope** default runs at import and therefore selects the backend for the whole file; one inside a test function runs *after* the module already imported mujoco and cannot change anything. So the module-scope sites are the ones that decide, and they are what this PR touches.

Which tree they sit in decides whether the wrong choice is visible:

* `tests/` is run by CI with `MUJOCO_GL: osmesa` exported (`test-lint.yml`), so the ten sites there are no-ops in CI. Run headless without exporting, their GL assertions skip rather than fail (measured: `6 passed, 4 skipped`).
* `tests_integ/` is not in `testpaths` and no workflow runs it, so nothing exports `MUJOCO_GL` for it. The three sites there are the ones that break.

Across the tree, 28 module-scope defaults already name a backend that works headless and 13 named `glfw`; all five notebooks and every guarded example already use the platform-guarded form. This adopts it at the remaining 13.

## Change

Each of the 13 module-scope sites becomes the form the notebooks and guarded examples already use:

```python
os.environ.setdefault("MUJOCO_GL", "cgl" if sys.platform == "darwin" else "egl")
```

That is the only change that fixes headless **without regressing macOS**: a bare `"egl"` is not in MuJoCo's valid set on Darwin and would raise there, where `glfw` works today. So this is not "replace glfw with egl", it is "name the platform".

`tests/test_examples_mujoco_gl.py` gains the second rule and now scans `tests_integ/` too. The existing line-scoped `cgl` rule is **untouched**. The new rule is AST-scoped to module level, which excludes by construction the sites where a backend name is the value *under test* -- a `monkeypatch.setenv`, or an assertion about what the resolver did (`test_backend.py` has both).

Out of scope, stated: the rule does not grade `egl`/`osmesa` defaults. An unguarded `"egl"` raises on macOS, but that is a macOS-hostile default rather than a windowed one, and converging the tree's offscreen spellings is a separate question.

## Evidence

**Pre-fix.** The new rule against `upstream/main` sources fails naming all thirteen with line numbers:

```
FAILED tests/test_examples_mujoco_gl.py::test_no_module_scope_windowed_gl_default
  a module-scope MUJOCO_GL default names a windowed GL backend (cgl, glfw), which cannot
  render on a headless host. Offending sites: {
    'tests/simulation/mujoco/test_cameras_recording_preflight_guards.py': ["line 46: 'glfw'"],
    'tests/simulation/mujoco/test_error_paths.py': ["line 25: 'glfw'"], ... 13 entries ...
    'tests_integ/simulation/test_multi_robot_tasks.py': ["line 27: 'glfw'"]}
1 failed, 10 passed
```

**Break table** -- each break fires exactly its own set, so no guard is decoration:

| break | fires |
| --- | --- |
| control (this branch) | 0 of 11 |
| revert the 13 sites | the rule, naming all 13 |
| descend into function bodies | the rule + `test_a_default_inside_a_test_function_is_out_of_scope` |
| stop honouring the platform guard | the rule + `test_the_guarded_form_is_accepted` |
| **revert only the 3 `tests_integ` sites AND narrow the scope back** | **nothing: 11 passed while three files are broken** |
| make the scan reach nothing | `test_scan_reaches_the_module_scope_defaults` only (the rule passes vacuously) |
| drop the subscript-assignment branch | `test_a_subscript_assignment_is_a_default_too` only |

The fourth row is why the scope widening is part of the fix rather than tidying: with the old scope the guard reports a clean tree while three files cannot render. The fifth is why the non-vacuity test is not decoration.

## Artifact

Same headless host, same shipped recording path, same four declared camera features; only the default the test module sets differs. Left: the recording that call produced. Right: `observation.images.top` decoded back out of the dataset it wrote.

MUJOCO_GL module-scope default

## Verification

Measured at `6838207`, `MUJOCO_GL=egl`, identical 426-file selection on both trees (every guard that walks a
source or test tree, reads `AGENTS.md`, or grades names and strings, plus `tests/simulation/**`):

| tree | result |
| --- | --- |
| this branch | `5 failed, 13681 passed, 82 skipped` in 730.89s |
| `upstream/main` | `5 failed, 13672 passed, 82 skipped` in 721.78s |

Failing ID sets identical, `comm` empty in both directions. `+9 passed` is exactly the nine tests this PR
adds. The five shared failures are import-order artifacts of a wide `tests/simulation` selection, not of this
change: four in `test_recording_frame_loss_is_not_tolerated.py` raising
`KeyError: 'strands_robots.simulation.policy_runner'` and one clock-double binding assertion in
`test_rendering_pacers_survive_a_clock_step.py`. Those two files run `29 passed` in isolation on pristine
`upstream/main`.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 errors on
both trees, none in the changed files. `tests/test_examples_mujoco_gl.py` is `11 passed` on this branch and
`1 failed, 10 passed` against `upstream/main` sources.


yinsong1986 (approved):
## Summary

Replaces the thirteen module-scope `os.environ.setdefault("MUJOCO_GL", "glfw")` defaults (ten in `tests/`, three in `tests_integ/`) with the platform-guarded form the notebooks and guarded examples already use, and extends `tests/test_examples_mujoco_gl.py` with an AST-scoped rule that refuses any module-scope windowed-backend default and now scans `tests_integ/` as well. The diff matches the description exactly: no production code, wire format, env-var name, or public API is touched; the change only alters which backend a test module selects when the operator exported nothing, and a user-exported `MUJOCO_GL` still wins via `setdefault`. Verified at head: the new scanner suite is 11/11, the rule reports a planted reverted site (`line N: 'glfw'`), the pre-existing line-scoped `cgl` rule is untouched, the diff is all-ASCII, and the changelog fragment (`changelog.d/2649-...md`) satisfies the step-3 rule. CI behaviour for `tests/` is unchanged since `test-lint.yml` exports `MUJOCO_GL=osmesa`, so the fix lands where it matters (`tests_integ/`, which nothing exports for) without regressing macOS, where a bare `"egl"` would raise.

## What's good

- The rule is scoped to what actually selects the backend (module level at import time), which excludes `monkeypatch.setenv` and resolver-assertion sites by construction rather than by an exemption list - consistent with how other tree-graders in this repo are built.
- Non-vacuity is pinned (`test_scan_reaches_the_module_scope_defaults`), and the break table in the description shows each guard fires exactly its own set - including the row demonstrating why the `tests_integ/` scope widening is part of the fix rather than tidying.
- The subscript-assignment spelling (`os.environ["MUJOCO_GL"] = V`) and module-level conditionals are both covered, and unparseable notebook cells fall back to a line scan instead of being silently skipped.
- AGENTS.md gains the second half of the existing `MUJOCO_GL` bullet with the pinning test named beside it, and the changelog fragment is in `changelog.d/` per step 3.

PR #2650, opened 2026-08-23
test(sim/isaac): a requested mesh size is discarded for the asset's extent

### LINKED ISSUE #2648: test(sim/isaac): nothing pins "mesh size is ignored" - the one claim the cross-backend divergence gate now leans on
## What is unpinned

`tests/simulation/test_mesh_size_docs_match_backend_divergence.py` documents a cross-backend divergence: `size=[2, 3, 4]` on a mesh object **scales the asset** on the Newton backend and is **ignored** on MuJoCo, and *both calls report success* - which is what makes the divergence dangerous rather than merely inconsistent.

#2498 moved Isaac onto MuJoCo's side of that divergence and updated the gate's prose to say so:

> Isaac sits on MuJoCo's side of this: its mesh `add_object` (#2459) documents `size` as ignored - the asset's own units define the extent - so the divergence measured here stays a two-party one (Newton vs. that shared read).

That is accurate against the code today. `IsaacSimulation.add_object`'s docstring states, for `shape="mesh"`, that `size` is ignored and that "the result reports the extent computed from the asset rather than echoing a request no component of which was consumed" - which is the honest design, because it makes the ignore observable to the caller instead of invisible.

**Nothing in the tree pins it.**

- `tests_integ/simulation/test_isaac_mesh_objects_gpu.py::test_mesh_add_object_and_scene_visuals_end_to_end` asserts `payload["size"] == pytest.approx([0.1, 0.1, 0.1])` for the tetra asset - but it calls `add_object` **without** a `size=` argument. So it pins "the reported extent comes from the asset when none was requested", not "a requested `size` is discarded and the asset extent reported anyway".
- The unit-level additions in `tests/simulation/isaac/test_isaac_backend.py` for the mesh path are all refusal cases (`mesh_path` without `shape='mesh'`, asset not found, unsupported `.dae`, no world), so none of them reach the success path where `size` would be consumed or dropped.
- `tests/simulation/isaac/test_scene_object_meshes.py` asserts `obj.size` against mesh bounds, but on the MJCF `load_scene` path, where the caller never supplies a `size` at all.

## Why it matters

If Isaac later starts consuming `size` as a scale - the Newton reading, and the more intuitive one for anyone who has just written Newton code - no test fails. The regression is silent by construction, because the whole point of the divergence is that both spellings report success.

The second-order cost is worse than the first. `test_mesh_size_docs_match_backend_divergence.py` would then itself be asserting something false about a third backend, so the artifact the repo relies on to *describe* this hazard becomes a source of it. A gate that documents a divergence it no longer measures correctly is worse than no gate, because it is trusted.

## Fix

One argument on a call that already exists - no new test, no extra Kit boot:

```python
result = sim.add_object(
    "widget",
    shape="mesh",
    mesh_path=str(asset),
    position=[0.4, 0.0, 0.2],
    size=[2.0, 3.0, 4.0],   # ignored for shape="mesh"; asset units define the extent
    is_static=True,
)
...
assert payload["size"] == pytest.approx([0.1, 0.1, 0.1])  # asset extent, not the request
```

A value wrong on every axis makes the discard unambiguous, and `[2, 3, 4]` is the example the divergence test itself uses, so the two read as a matched pair for anyone tracing the claim.

## Acceptance criteria

- [ ] The GPU journey test passes an explicit `size=` on the `shape="mesh"` `add_object` call and asserts the reported extent is the asset's, not the request
- [ ] The assertion is reachable from `test_mesh_size_docs_match_backend_divergence.py`'s prose (a comment naming that file, so the claim and its pin are greppable from each other)
- [ ] No change to `add_object` behaviour - this is a pin on existing, already-correct behaviour

## Notes

Raised during review of #2498 and deliberately not blocking it: nothing in that PR is false or broken, the behaviour and both docstrings agree with the gate prose today, and the feature was already four days old. This is hardening against a future silent regression, tracked so it survives the merge.

If #2498 picks up the one-line change before it merges, close this as done.

cagataycali:
## What is unpinned

`tests/simulation/test_mesh_size_docs_match_backend_divergence.py` is the gate that measures a cross-backend divergence in `add_object(shape="mesh", ...)`: **Newton consumes `size`** as a per-axis scale, **MuJoCo does not**, and *both calls report success* - which is what makes it expensive to discover rather than merely inconsistent.

#2498 put Isaac on MuJoCo's side of that split and updated the gate's prose to say so:

> Isaac sits on MuJoCo's side of this: its mesh `add_object` (#2459) documents `size` as ignored - the asset's own units define the extent - so the divergence measured here stays a two-party one.

That is accurate against the code. **Nothing measured it.**

| surface | what it pins | reaches a consumed/dropped `size`? |
| --- | --- | --- |
| `tests_integ/.../test_isaac_mesh_objects_gpu.py` | the extent comes from the asset when **no** `size` is requested (the call passes none) | no |
| `tests/simulation/isaac/test_isaac_backend.py` mesh cases | 4 refusals - `mesh_path` without `shape='mesh'`, asset missing, `.dae`, no world | no, all pre-success |
| `tests/simulation/isaac/test_scene_object_meshes.py` | `obj.size` against mesh bounds on the `load_scene` path, where the caller supplies no `size` | no |

So a change that started reading `size` as a scale - the Newton reading, and the intuitive one for anyone arriving from Newton code - would break no test. The second-order cost is worse: the gate above would then be asserting something false about a third backend, so the artifact the repo relies on to *describe* this hazard becomes a source of it.

## Measured

Driven through the real `add_object`, on the tree as it stands. The asset is a `0.1 x 0.2 x 0.3` OBJ box; the request is the `[2, 3, 4]` the divergence gate itself uses.

| call | reported `size` |
| --- | --- |
| `shape="mesh"`, `size=[2, 3, 4]` | `[0.1, 0.2, 0.3]` - the asset's extent |
| `shape="mesh"`, no `size` | `[0.1, 0.2, 0.3]` - identical |
| `shape="box"`, `size=[0.21, 0.31, 0.41]` | `[0.21, 0.31, 0.41]` - **consumed** |

and no component of `[2, 3, 4]` reaches the prim construction at all: the `SingleGeometryPrim` wrapper is handed `collision`, `name`, `orientation`, `position` and nothing else.

## Change

Pins that contract where CI runs it - `tests/`, not `tests_integ/`, which no workflow runs and which needs a GPU and a Kit boot. The mesh success path turns out to be reachable without either: the Kit-only leaves (the stage reference and the prim wrappers) are stood in following `test_scene_object_meshes`, and `mesh_aabb` - the function whose output the assertions are about - runs for real. Only the USD *conversion* is stubbed, because the extent is parsed from the source asset rather than from the converted stage.

Both halves are asserted, because the payload alone does not cover the class:

* the **report** is the asset's extent whatever was requested, and is identical to the same call with no `size` - "ignored" in its strongest form;
* **no component of the request reaches the prim construction**. Newton's reading scales the *geometry*, while `resolved_size` here is parsed from the *unscaled* asset file - so a scale that reached the prim would leave the payload honest and put a wrong-sized object on the stage, invisible to any assertion on the report. Searched numerically, so a scale arriving as a `numpy` array - which is how a real Isaac call spells it - is not missed.

Plus a structural assertion that `_create_mesh_prim` declares no `size` parameter. A scale-consuming regression arrives in two steps - the parameter, then the behaviour - and the two above only see the second.

The request is imported from the gate rather than restated, so the claim and its pin are one edit apart in either direction, and the gate's Isaac paragraph now names the pin. The GPU journey test passes the same value explicitly, so the real-stage half carries the assertion too.

`strands_robots/` is untouched - `git diff --numstat upstream/main..HEAD` is three paths, all under `tests/` and `tests_integ/`. No policy, sim behaviour, rendering, recording or robot asset changes, so there is nothing to film; the behaviour this pins is #2498's, already shipped.

## Each guard is load-bearing

The behaviour is already correct, so there is no pre-fix failure to show. Instead, each plausible regression toward the Newton reading was applied to `simulation.py` and the pin re-run.

| break applied to `_create_shape_prim` / `_create_mesh_prim` | fails |
| --- | --- |
| *(control - the tree as it stands)* | 0 of 6 |
| full Newton reading: `size` parameter + forwarded + extent multiplied by it | 3 - both report assertions + the signature |
| **`size` parameter added but unused** (first step of a two-step regression) | **1 - the signature alone** |
| `size` forwarded and applied to the prim constructor as a scale | 2 - the construction assertion + the signature |
| the payload echoes the request, no signature change | 2 - both report assertions |
| **the primitive path stops consuming `size`** (the tempting "drop the extent earlier" shortcut) | **1 - the no-over-reach control alone** |
| **a scale applied to the handle after `_create_mesh_prim` returns**, no signature change | **1 - the construction assertion alone** |
| the gate's `MESH_SIZE` changes to a value the asset happens to be | **1 - the premise alone** |

Six of the seven fire a different set, and each of the six tests fires uniquely somewhere - including the premise, which guards the fixture rather than the code and reports when the request stops being distinguishable from the rest of the call.

## Verification

Measured at `b2dd4ba8`; the head has since moved to `9998729a`, which adds only the changelog fragment (`git diff b2dd4ba8..9998729a --stat` is that one path), so the numbers below stand.

* `mypy strands_robots tests tests_integ`: **24** on this branch and **24** on a pristine `upstream/main` worktree, identical sets - none in the changed files.
* The new file: **6 passed**.
* 283-file selection - every test touching the mesh `size` surface, all of `tests/simulation/isaac/`, and the repo-wide docstring / string / changelog guards - run with the **identical** list on this branch and on a pristine `upstream/main` worktree, with the new file excluded from both so the comparison measures only the two edited ones:

  | tree | result |
  | --- | --- |
  | branch | `4 failed, 12344 passed, 182 skipped` in 571s |
  | `upstream/main` | `4 failed, 12344 passed, 182 skipped` in 575s |

  Byte-identical summaries, identical failing-ID sets, `comm` empty in both directions. All four are `tests/simulation/test_recording_frame_loss_is_not_tolerated.py` failing on `KeyError: 'strands_robots.simulation.policy_runner'`; that file is **13 passed** run in isolation on pristine `upstream/main`, so they are the collection-order artifact of the wide selection rather than anything in this diff.
* The new file run with random ordering enabled beside its two siblings: **24 passed** - order-independent.

Closes #2648.

### DIFF
--- a/changelog.d/2650-isaac-mesh-size-discarded-for-asset-extent.md
+++ b/changelog.d/2650-isaac-mesh-size-discarded-for-asset-extent.md
@@ -1,0 +1,46 @@
+### Quality: the Isaac mesh path's discard of a requested `size` is now driven
+
+`tests/simulation/test_mesh_size_docs_match_backend_divergence.py` measures a
+cross-backend divergence in `add_object(shape="mesh", ...)`: Newton consumes
+`size` as a per-axis scale, MuJoCo does not, and both calls report success. #2498
+placed Isaac on MuJoCo's side and said so in that gate's prose, and nothing
+measured it.
+
+The three surfaces that reach Isaac's mesh path all stop short of the property.
+The GPU journey test passes no `size` at all, so it pins "the extent comes from
+the asset when none was requested"; the four unit-level mesh cases are refusals
+that never reach the success path; and the `load_scene` mesh cases read
+`obj.size` on a path where the caller supplies none. A change that started
+reading `size` as a scale -- the Newton reading, and the intuitive one for anyone
+arriving from Newton code -- would break no test, and the gate above would then
+be asserting something false about a third backend, so the artifact the repo
+relies on to describe the hazard becomes a source of it.
+
+Six cases now drive the mesh success path in `tests/`, where CI runs, rather than
+in `tests_integ/`, which no workflow runs. Only the Kit-only leaves are stood in,
+following `test_scene_object_meshes`; `mesh_aabb`, whose output the assertions are
+about, runs for real against a `0.1 x 0.2 x 0.3` OBJ box. Both halves of
+"ignored" are pinned, because the payload alone does not cover the class: the
+report is the asset's extent whatever was requested and is identical to the same
+call with no `size`; and no component of the request reaches the prim
+construction, searched numerically so a scale arriving as a `numpy` array is not
+missed. That second half matters because `resolved_size` is parsed from the
+unscaled asset file, so a scale that reached the prim would leave the report
+honest and put a wrong-sized object on the stage. A structural assertion that
+`_create_mesh_prim` declares no `size` parameter catches the first step of the
+two-step regression the others only see the second of.
+
+Seven plausible regressions toward the Newton reading were applied and the pin
+re-run. Six fire a different set, and each of the six cases fires uniquely
+somewhere: the unused-parameter step is caught by the signature assertion alone,
+a scale applied to the handle after `_create_mesh_prim` returns by the
+construction assertion alone, the tempting "drop the extent earlier" shortcut by
+the no-over-reach control alone, and a request that stops being distinguishable
+from the rest of the call by the premise alone.
+
+The request is imported from the divergence gate rather than restated, so the
+claim and its pin are one edit apart in either direction, and that gate's Isaac
+paragraph now names the pin. The GPU journey test passes the same value
+explicitly.
+
+No library behaviour changes.
--- a/tests/simulation/isaac/test_mesh_size_is_discarded_for_the_asset_extent.py
+++ b/tests/simulation/isaac/test_mesh_size_is_discarded_for_the_asset_extent.py
@@ -1,0 +1,371 @@
+"""Isaac's mesh ``size`` is discarded and the asset's own extent reported (#2459).
+
+``add_object(shape="mesh", ...)`` reads ``size`` differently per backend, and
+:mod:`tests.simulation.test_mesh_size_docs_match_backend_divergence` is the gate
+that measures the disagreement: **Newton consumes it** as a per-axis scale,
+**MuJoCo does not**, and *both calls report success* - which is what makes the
+divergence expensive to discover rather than merely inconsistent.
+
+#2498 put Isaac on MuJoCo's side of that split, and that gate's prose now says
+so. Nothing measured it. The claim was reachable only through the GPU journey
+test, which calls the mesh ``add_object`` **without** a ``size=`` argument - so
+it pinned "the reported extent comes from the asset when none was requested",
+not "a requested extent is discarded and the asset's reported anyway". The
+remaining unit-level mesh cases are all refusals (``mesh_path`` without
+``shape='mesh'``, asset missing, ``.dae``, no world), none of which reach the
+success path where a ``size`` would be consumed or dropped.
+
+So a future change that starts reading ``size`` as a scale - the Newton reading,
+and the intuitive one for anyone arriving from Newton code - would break no
+test, and the gate above would then be documenting a third backend's behaviour
+falsely. A gate trusted to *describe* a hazard is worse than none once it is
+wrong about it.
+
+Both halves are asserted, because the payload alone does not cover the class:
+
+* the **report** is the asset's extent whatever was requested, and is
+  byte-identical to the same call with no ``size`` at all - "ignored" in its
+  strongest form;
+* **no component of the request reaches the prim construction**. Newton's
+  reading scales the *geometry*, while ``resolved_size`` here is computed from
+  the *unscaled* asset file - so a scale that reached the prim would leave the
+  payload honest and the object the wrong size on the stage, invisible to any
+  assertion on the report. Every argument the construction leaves are handed is
+  searched numerically, so a scale arriving as a ``numpy`` array - which is how
+  a real Isaac call would spell it - is not missed.
+
+The request ``[2, 3, 4]`` is imported from that gate rather than restated, so
+the claim and its pin are one edit apart in either direction, and the asset's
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Test-only PR that pins Isaac's mesh `add_object` contract where CI actually runs it: a requested `size` on `shape="mesh"` is discarded and the asset's own extent reported, which #2498 claimed in the divergence gate's prose but nothing measured. The new `tests/simulation/isaac/test_mesh_size_is_discarded_for_the_asset_extent.py` drives the real `add_object` success path with only the Kit-only leaves stood in (`mesh_aabb` runs real; only the USD conversion is stubbed, and `_create_mesh_prim`'s call-time `from ... import` means that monkeypatch actually lands), asserting both the report and — numerically, so a numpy-spelled scale is not missed — that no component of the request reaches the prim construction, plus a structural guard on `_create_mesh_prim`'s signature that catches the first step of a two-step regression. `MESH_SIZE` is imported from the divergence gate rather than restated, and the gate's prose now names the pin, so the claim and its measurement move together. `strands_robots/` is untouched; the diff against the merge base is exactly the four declared files.

## What's good

- The new file runs green in isolation (6 passed, verified locally at the head SHA) and its premise test guards the fixture's own discriminating power rather than assuming it.
- AGENTS.md compliance throughout: `monkeypatch.setitem(sys.modules, ...)` for the fake `isaacsim` tree (restores, no orphaned references), behaviour-describing test names, contiguous `:mod:` cross-references naming real modules, ASCII-only strings, no host paths, changelog fragment under `changelog.d/` instead of a `CHANGELOG.md` edit.
- The mutation table in the description (seven plausible regressions, six distinct failure sets, each guard firing uniquely somewhere) is exactly the load-bearing-guard evidence the repo's conventions ask for on a behaviour that is already correct.
- The GPU journey test now passes the same `[2, 3, 4]` explicitly, so the real-stage half carries the assertion too.

PR #2651, opened 2026-08-23
test(policies/cosmos3): spawned-server MuJoCo rollout integration test (service mode, closed loop)

### LINKED ISSUE #2513: test(policies/cosmos3): spawned-server MuJoCo rollout integration test (service mode, closed loop)
# Goal

Extend the cosmos3 **service-mode** integration coverage with (a) an opt-in fixture that **spawns** the RoboLab policy server itself (so a single GPU box can run the test with one command, no hand-started server), and (b) a short **MuJoCo rollout** driven by real server actions - the "run a short MuJoCo rollout, assert actions are finite/in-range and the episode completes" half of the rule-8 gap. This is the cosmos3 analogue of `tests_integ/groot/test_groot_integration.py`, which spins up a local GR00T server and then consumes it, versus the pre-running-server test added by the sibling issue.

Depends on #2512 - if that PR hasn't merged yet, post a comment on this issue and stop.

Part of #2462.

# What to build

Extend/add alongside `tests_integ/policies/cosmos3/test_service_mode_live.py` (created by the dependency). Suggested new file: `tests_integ/policies/cosmos3/test_service_mode_rollout.py`.

## 1. Server-spawn fixture (module-scoped)

Mirror the `groot_server` fixture pattern in `tests_integ/groot/test_groot_integration.py`:

- Gate on `COSMOS3_SPAWN_SERVER=1` (skip otherwise); also `pytest.importorskip` on `websockets`/`msgpack`, and skip if `importlib.util.find_spec("cosmos_framework")` is None (the server package is external and not a dependency of this repo).
- Launch via `subprocess.Popen([sys.executable, "-m", "cosmos_framework.scripts.action_policy_server_robolab", ...])` with the checkpoint `nvidia/Cosmos3-Nano-Policy-DROID` (allow override via `COSMOS3_MODEL`), argv-style only, on `127.0.0.1` and a port from `COSMOS3_SERVER_PORT` (default 8000). Check the actual CLI flags of the installed script before hard-coding them - read `python -m cosmos_framework.scripts.action_policy_server_robolab --help` in the docstring's quickstart or the module docstring of `strands_robots/policies/cosmos3/policy.py`.
- Poll for readiness by attempting the client handshake (`create_policy("cosmos3", ...)` connect or a raw socket connect) with a deadline measured on `time.monotonic()` (AGENTS.md clocks rule - never `time.time()` for a wait budget). Model load can take minutes; make the deadline generous (e.g. 600 s) and env-overridable.
- Teardown must terminate the process even on test failure (`try/finally` or fixture finalizer: `proc.terminate()`, wait, `proc.kill()` fallback). Leaking a GPU-holding server is silent damage (AGENTS.md resource-cleanup rule).
- If the spawn exits before ready, **fail with the server's captured stderr tail** in the message, not a bare timeout - a missing checkpoint must be diagnosable from the pytest output.

## 2. Short MuJoCo closed-loop rollout

- `pytest.importorskip("mujoco")`; load the Panda model via `robot_descriptions.panda_mj_description.MJCF_PATH` exactly as `tests_integ/policies/cosmos3/test_diffusers_backend_live.py::test_cosmos_action_drives_mujoco_arm_within_tracking_bar` already does (reuse its setup shape).
- Loop for a small number of chunks (2-3): render/synthesize the three DROID camera observations plus the current 7 joint positions + gripper from MuJoCo state, call `policy.get_actions_sync(obs, instruction)`, apply each step's joint targets to the sim (`data.ctrl` via position actuators, or `qpos` + `mj_forward` if the model's actuators are not position servos - respect the AGENTS.md actuator rule: only write a pose where the drive takes a pose; the Panda MJCF from robot_descriptions ships position actuators), `mj_step` a fixed number of substeps per action step.
- Use `create_policy("cosmos3", embodiment="droid", host=..., port=..., robot="panda")` so the built-in DROID->Panda actuator mapping documented in `strands_robots/policies/cosmos3/policy.py` is exercised (per-step dicts keyed by real Panda actuator names).

# Acceptance criteria

- Every action value across the whole rollout is a finite float (`np.isfinite` on the stacked chunk) and within the corresponding actuator's `ctrlrange` when one is defined (read from the MjModel, don't hard-code numbers).
- The episode completes: the loop runs its full planned number of chunks x steps without an exception and the final `mj_step` leaves `data.qpos` finite.
- The policy is asked at least twice, so the test proves the connection survives repeated inference (chunked closed loop, not a single round-trip).
- With `COSMOS3_SPAWN_SERVER` unset, the entire file **skips** cleanly - `hatch run test-integ` on a dep-less box stays green.
- Fixture teardown leaves no orphan server process (assert `proc.poll() is not None` after teardown in the fixture finalizer itself).

- Assert on outputs/behavior, not internal state.
- No host-local paths in test files; plain ASCII; no emojis.
- All wait budgets on `time.monotonic()`.
- Changelog fragment `changelog.d/<pr-number>-<slug>.md`; never edit `CHANGELOG.md` directly.
- `hatch run format && hatch run lint && hatch run test` must pass before the PR.

Parent: #2462

yinsong1986:
Closes #2513
Part of #2462

## What changed

Adds `tests_integ/policies/cosmos3/test_service_mode_rollout.py` - the **spawned-server** half of the cosmos3 service-mode integration coverage. The sibling file added by #2517 (`test_service_mode_live.py`) consumes a pre-running RoboLab policy server; this one is the cosmos3 analogue of `tests_integ/groot/test_groot_integration.py`: a module-scoped fixture spawns the server itself, so a single GPU box runs the whole thing with one command, and a short closed-loop MuJoCo Panda rollout is driven by real server actions.

Plus the changelog fragment `changelog.d/2513-cosmos3-spawned-server-rollout.md`.

## How it works

**Server-spawn fixture (module-scoped, opt-in via `COSMOS3_SPAWN_SERVER=1`):**

- Launches `python -m cosmos_framework.scripts.action_policy_server_robolab --checkpoint-path <COSMOS3_MODEL> --port <COSMOS3_SERVER_PORT>` argv-style (no shell), defaults `nvidia/Cosmos3-Nano-Policy-DROID` / port 8000. `--checkpoint-path` + `--port` are the flags this repo documents in the `strands_robots.policies.cosmos3` quickstart and `client._server_hint()`; `cosmos_framework` is not installed on the dev box, so the installed script's `--help` could not be read directly (noted below).
- Server output goes to a file rather than a pipe, so a chatty model load cannot fill the pipe buffer and stall the server, and the failure paths can report the log tail while the process is alive.
- Readiness = a completed WebSocket connect + msgpack metadata handshake through the shipped `Cosmos3WebsocketClient` (not a bare TCP accept), polled on a `time.monotonic()` deadline (default 600 s, `COSMOS3_SERVER_TIMEOUT`). A spawn that exits before serving fails with the captured log tail, so a missing checkpoint is diagnosable from the pytest output rather than reading as a bare timeout.
- Teardown runs in a `finally`: SIGTERM to the process group, wait, SIGKILL fallback - and the finalizer itself asserts `proc.poll() is not None`, so a leaked GPU-holding server fails the run rather than outliving it.

**Closed-loop rollout:**

- 3 chunks x 10 `mj_step` substeps per action step against the Panda MJCF from `robot_descriptions.panda_mj_description.MJCF_PATH` (same setup shape as `test_diffusers_backend_live.py`). Each chunk's observation reads the 7 joint positions + gripper from the live sim state (gripper normalized by the joint's own range from the model), so each inference is conditioned on the state the previous chunk produced. The three RoboArena camera views are synthesized, keeping the test off the GL stack; the assertions are about the action path.
- `create_policy("cosmos3", embodiment="droid", host=..., port=..., robot="panda")` exercises the built-in DROID -> Panda actuator mapping (per-step dicts keyed `joint1..joint7`, `finger_joint1`).
- Actuator rule respected per actuator via the shared `scene_ops.joint_drive_map`: poses are written only into position servos; the tendon-driven Panda gripper (`actuator8`, one ctrl in tendon units coupling two finger joints) is left uncommanded and named, and the test asserts the uncommanded set is exactly that.

## Acceptance criteria (from #2513)

- [x] Every action value across the whole rollout is a finite float (`np.isfinite` on the stacked raw chunks) and within the corresponding actuator's `ctrlrange` when one is defined - read from the `MjModel`, checked before the write so an out-of-range action never reaches the sim.
- [x] The episode completes: full planned chunks x steps, final `data.qpos` finite.
- [x] The policy is asked at least twice (3 chunks; asserted).
- [x] With `COSMOS3_SPAWN_SERVER` unset the entire file skips cleanly (verified: 1 module skip, dep-less path also covered by `importorskip` + the `cosmos_framework` find_spec gate).
- [x] Fixture teardown leaves no orphan server (`assert proc.poll() is not None` in the finalizer itself).

## Test surface / validation

- `hatch run format` / `hatch run lint` (ruff + mypy): clean.
- `hatch run test` (`MUJOCO_GL=egl`): **36369 passed, 279 skipped, 0 failed**, coverage 95.53%.
- Skip path verified with and without `COSMOS3_SPAWN_SERVER=1` on a box without `cosmos_framework`: clean module-level skip both ways.
- The full spawn -> readiness-poll -> 3-chunk closed loop -> ctrlrange/finite assertions -> teardown/no-orphan path was exercised end-to-end locally against a stand-in `cosmos_framework.scripts.action_policy_server_robolab` speaking the real wire protocol (msgpack metadata handshake + per-request `{"action": [T, 8]}` chunks, simulated slow start): 1 passed, no orphan process after teardown. The real-weights run needs the external `cosmos_framework` package + checkpoint on a GPU box.

## Out of scope / notes

- The launch flags are the two this repo documents (`--checkpoint-path`, `--port`). If the installed server script's CLI diverges, the fixture fails fast with the server's own argparse error in the reported log tail - which is exactly the diagnosable-failure shape the issue asks for.
- Project board: please move #2513 to "In review" (or it will auto-move on merge).


tests_integ/policies/cosmos3/test_service_mode_rollout.py
```diff
@@ -0,0 +1,307 @@
...
+        sys.executable,
+        "-m",
+        "cosmos_framework.scripts.action_policy_server_robolab",
+        "--checkpoint-path",
+        MODEL,
+        "--port",
+        str(PORT),
+    ]
+    with open(log_path, "wb") as log_file:
+        proc = subprocess.Popen(
```
cagataycali, line 211: The fixture never establishes that the server answering the readiness probe is the one it spawned.

`_wait_until_serving` checks `proc.poll()` at the top of iteration 1 -- microseconds after this `Popen`, and before the child can reach its own `bind()` -- then probes the port. If a foreign server already holds `HOST:PORT`, the probe succeeds, the fixture yields, and this child dies of `EADDRINUSE` unobserved. The `assert proc.poll() is not None` in the `finally` block then passes *because* the child crashed, so the leak guard is satisfied by the exact condition it should refuse.

Reachable via the sibling `test_service_mode_live.py`, which reads the same `COSMOS3_SERVER_PORT` with the same default `8000` against a server it expects to be already running; `HOST` is pinned to `127.0.0.1` here, so nothing else separates the two.

A `socket.bind((HOST, PORT))` probe before this line, failing with `pytest.fail`, turns the collision into a named refusal. Full reasoning, the mirrored-control-flow demonstration and an alternative positive-assertion fix are in the review summary.

cagataycali (changes requested):
The test itself is solid, and the parts that are usually wrong in a spawned-server fixture are right here: readiness is a real client handshake rather than a TCP accept, the deadline is on `time.monotonic()`, output goes to a file so a chatty model load cannot fill a pipe buffer, teardown kills the process *group* and runs on the failure path, and `ctrlrange` is read from the `MjModel` rather than hard-coded. I checked the four API claims it rests on and all four hold: `last_rollout` is public and documented (`policies/cosmos3/policy.py:180`), `joint_drive_map(model, mj)` matches the call, the `gpu` marker is registered, and `--timeout=300` in the `test-integ` script is what makes the `pytest.mark.timeout(READY_TIMEOUT_S + 300)` override necessary rather than decorative. The fragment name is fine too -- `changelog.d/README.md` says `<number>` is "your PR (or issue) number", so `2513-` is legal.

One change requested, on the fixture's identification of its own subject. Reviewed at `01be0e36`.

## The spawn can be silently replaced by a foreign server

`_wait_until_serving` proves *a* server answers on the port; nothing proves it is the one the fixture spawned. Two facts already in the tree make this reachable rather than theoretical: the sibling `test_service_mode_live.py` reads the **same** `COSMOS3_SERVER_PORT` env var with the **same** default `8000`, and its documented setup is a server already running on it. So "run the spawn gate on a box where the live server is up" is one env var away, and `HOST` is hardcoded to `127.0.0.1`, so there is no host difference to separate them.

In that case the spawned process dies of `EADDRINUSE` and the rollout is driven by the foreign server. The sharp part is that the assertion written to catch a broken spawn is exactly the one a crashed spawn satisfies: `assert proc.poll() is not None` passes *because* the process is gone, and it cannot tell "we killed it" from "it never bound".

I mirrored the control flow verbatim, substituting a bare TCP connect for the websocket handshake -- the ordering of `poll()` and the probe is what is under test, not what the probe speaks:

```
foreign server listening on 127.0.0.1:8765
readiness verdict          : READY (fixture yields, rollout proceeds)
probe answered by          : metadata-from-whoever-answered
spawned proc returncode    : 1  (nonzero => it crashed, EADDRINUSE)
leak-guard assert          : PASSED
```

Green run, every assertion satisfied, zero server actions from the process under test. The race is not tight: iteration 1 evaluates `proc.poll()` microseconds after `Popen`, long before the child can reach its own `bind()`, so the `None` there is guaranteed rather than lucky -- the crash is never observable at the only place the loop looks for it.

This is the same shape as key convention 6 (no silent default on error). The test does not report a weaker result; it reports a *passing* one about the wrong subject.

## Suggested fix

Refuse an occupied port before spawning, so the collision is named instead of absorbed:

```python
import socket

def _refuse_occupied_port() -> None:
    """Fail before spawning if anything already answers on the port.

    Readiness below cannot tell the spawned server from a pre-existing one,
    and the sibling live-server test reads the same COSMOS3_SERVER_PORT with
    the same default -- so an occupied port would silently redirect this
    rollout onto a process this fixture does not own or control.
    """
    probe = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    try:
        probe.bind((HOST, PORT))
    except OSError as exc:
        pytest.fail(
            f"{HOST}:{PORT} is already in use ({exc}). This file spawns its own "
            f"server; a pre-running one (see test_service_mode_live.py, same "
            f"COSMOS3_SERVER_PORT) would answer the readiness probe and the "
            f"rollout would measure a process this fixture did not spawn. "
            f"Set COSMOS3_SERVER_PORT to a free port, or stop the running server."
        )
    finally:
        probe.close()
```

called at the top of the fixture, before `Popen`. `SO_REUSEADDR` is deliberately **not** set on that probe: the point is to observe the contention the child will hit, not to work around it. There is a benign close/bind gap between the probe and the child's own `bind()`, which is acceptable because the child still fails loudly on `EADDRINUSE` in that window -- the guard removes the silent-wrong-subject case and does not claim to be atomic.

If you would rather assert positively than pre-check, an equally good answer is to have readiness confirm identity: require that the metadata handshake succeed *and* `proc.poll() is None` immediately afterwards, then re-assert `proc.poll() is None` right before `yield`. That closes the same hole from the other side and additionally catches a spawn that dies for a reason unrelated to the port. Either mechanism is fine by me; the current state is what I am asking to change.

One small consequence worth folding in: with the guard in place, the module docstring's `COSMOS3_SERVER_PORT` line can say the port must be free, which is the one thing an operator reading only the docstring cannot currently infer.

tests_integ/policies/cosmos3/test_service_mode_rollout.py
yinsong1986: Good catch - the readiness probe was checking the port, not the child, so a foreign server on the same `HOST:PORT` (exactly what the sibling `test_service_mode_live.py` sets up, same `COSMOS3_SERVER_PORT`, same default) would answer the handshake while the spawned child died of `EADDRINUSE` unobserved, and the teardown leak assert would pass *because* of the crash.

Fixed in `fdab295d` with the `socket.bind((HOST, PORT))` preflight you suggested: the fixture now runs `_fail_unless_port_is_free()` before the `Popen` and `pytest.fail`s with a named refusal (stop the other server, or set `COSMOS3_SERVER_PORT` to a free port). The probe sets `SO_REUSEADDR` so a TIME_WAIT remnant from a previous run does not refuse a port the child could in fact bind - the refusal fires only for a live holder, which is the case the readiness probe would otherwise misread as ready.

yinsong1986:
@cagataycali Thanks for the review - point by point:

**1. The spawn can be silently replaced by a foreign server -> fixed in `fdab295d` (first of your two suggested mechanisms).** The fixture now runs `_fail_unless_port_is_free()` before the `Popen`: a `socket.bind((HOST, PORT))` preflight that `pytest.fail`s with a named refusal pointing at the `test_service_mode_live.py` collision and the two remedies (stop the other server, or set `COSMOS3_SERVER_PORT` to a free port). With the port guaranteed unowned at spawn time, `_wait_until_serving`'s per-iteration `proc.poll()` also recovers its meaning: nothing else can answer the handshake, so a child that dies for *any* reason is caught on the next iteration and reported with the captured log tail - which covers the extra case your second mechanism (identity-confirming readiness) would have caught.

**2. One deliberate deviation from your snippet: the probe *does* set `SO_REUSEADDR`.** I read your stated goal - "observe the contention the child will hit, not work around it" - as the spec, and matching the child's posture is what serves it: the child binds through asyncio's `create_server` (which `websockets.serve` delegates to), and that enables `SO_REUSEADDR` by default on POSIX, so a `TIME_WAIT` remnant from a previous run is contention the child will *not* hit. A probe without the flag would refuse a port the child can in fact bind - a false refusal on exactly the run-this-file-repeatedly workflow. The guard is not weakened by the flag: on Linux `SO_REUSEADDR` permits binding over `TIME_WAIT` only, never over an active LISTEN socket, so a live foreign server still refuses the probe's bind with `EADDRINUSE`. And in the one corner where the postures could differ (child without `SO_REUSEADDR`, port in `TIME_WAIT`), the outcome is the child's own loud `EADDRINUSE`, caught by `_wait_until_serving`'s poll with the log tail - a named failure, not the silent-wrong-subject case. If you'd rather have the letter of the snippet, dropping the `setsockopt` line is a one-liner - say the word.

**3. Docstring consequence -> fixed in `a3d9d576`.** The `COSMOS3_SERVER_PORT` line now says the port must be **free**, names the pre-spawn refusal, and points at the sibling live-server file that shares the variable and default - the one thing an operator reading only the docstring previously could not infer.

Validation: `hatch run format` and `hatch run lint` clean. `hatch run test`: the full suite aborts on this box inside an unrelated libero MuJoCo GL test (`test_libero_camera_config_domain.py`), and the abort reproduces byte-for-byte on an unmodified checkout of `main` here (headless GL, order-dependent - passes in isolation and as a directory run), so it is environmental and pre-existing; the diff on this branch touches only `tests_integ/` (not collected by the unit run) and CI on the head is the authoritative verdict. The inline thread from this review is resolved pointing at `fdab295d`.

cagataycali (approved):
Verified `fdab295d` closes the finding, and the preflight is measurably correct in both directions rather than only in the one the review named.

## The preflight, driven against real sockets

`_fail_unless_port_is_free` extracted verbatim and driven with a live holder on a scratch port (the module cannot be imported here - `cosmos_framework` is absent, so the module-level skip fires):

| holder of `HOST:PORT` | verdict |
| --- | --- |
| nothing holds the port | PASSES |
| live listener on `127.0.0.1:PORT` | REFUSES |
| live listener on `0.0.0.0:PORT` (wildcard) | REFUSES |
| server-side TIME_WAIT remnant | PASSES |
| live listener that also set `SO_REUSEPORT` | REFUSES |

Three things worth naming:

**The wildcard row is the one that matters.** A server launched with `--port 8000` binds `0.0.0.0`, not the loopback address this file pins - so the realistic collision is a holder on a *different* local address than the probe's. `bind(("127.0.0.1", PORT))` still refuses it, so the refusal covers the case the review was actually about rather than only an exact-address match.

**`SO_REUSEADDR` is load-bearing, not decoration.** Against that same TIME_WAIT remnant a probe *without* it returns `EADDRINUSE` - so without the option the preflight would refuse a port the child could in fact bind, exactly as the docstring says. The option is scoped precisely: it lets a remnant through and still refuses every live holder.

**The refusal is slightly broader than the docstring claims.** A holder that set `SO_REUSEPORT` is also refused, because the probe does not set it and therefore cannot join the group. No change needed - just noting the bound is the safe side of what is documented.

Ordering is right: the preflight is line 199, `Popen` line 211, so nothing is spawned before the port is claimed. And with the collision now a named refusal, the `assert proc.poll() is not None` leak guard can no longer be satisfied *by* the crash it should refuse - a child that dies for any other reason (missing checkpoint, bad flag) reaches `_wait_until_serving`'s `pytest.fail` with the captured log tail instead.

## Rollout half

Checked the facts that would otherwise only surface on a GPU box:

- `joint_drive_map(model, mj)` - signature matches the call.
- `panda_mj_description` reports `nkey=1`, so `mj_resetDataKeyframe(model, data, 0)` is valid.
- Servo classification is exactly as the comment claims: all seven arm joints have position servos; `finger_joint1` does not (actuator 7 is the tendon coupling both fingers). So `uncommanded <= {finger_joint1}` is precise, and leaving the gripper target uncommanded is the correct reading of the actuator rule rather than a workaround.
- All eight actuators report `ctrllimited=1`, so the `ctrlrange` assertion is live for every arm joint rather than vacuously skipped.

`ruff check` and `ruff format --check` clean; `mypy` reports no error in this file.

One non-blocking note for the record: a window remains between `probe.close()` and the child's own `bind()`, so the preflight narrows the race rather than closing it. Closing it entirely would mean handing the bound descriptor to the child, which is not worth the coupling for an opt-in single-box fixture - the named refusal is the right trade.

Approving.

PR #2652, opened 2026-08-23
fix(examples): ask the Hub where its token lives, and offer a login that runs

cagataycali:
## The defect

The three `examples/libero` drivers refuse `--policy groot` up front when no
HuggingFace token is available, rather than letting the gated Cosmos-Reason2-2B
download fail later inside `gr00t_inference`. That refusal *is* the preflight, so
it carries two obligations: resolve the token the Hub resolves, and name a login
that runs. It did neither.

`huggingface_hub` resolves its cached login from `HF_TOKEN_PATH`, else
`<HF_HOME>/token`, else `<XDG_CACHE_HOME>/huggingface/token`, else
`~/.cache/huggingface/token`. The preflight read the last of those directly, so
on any host that relocated its cache it names a file the Hub will not open - and
a box that *is* logged in gets refused. Measured on a host with `HF_HOME`
relocated and a token cached where the Hub looks for it:

| tree | relocated cache, logged in | `HF_TOKEN` set, no cached login | login offered |
| --- | --- | --- | --- |
| `upstream/main` | **REFUSED**, 3 of 3 drivers | refused by `run_mujoco_agent.py` | `huggingface-cli login` |
| this branch | accepted, 3 of 3 | accepted, 3 of 3 | `hf auth login` |

The remedy has its own failure mode, and it is worse than deprecated:

| offered command | published at the declared `huggingface_hub>=1.5` floor | running it |
| --- | --- | --- |
| `huggingface-cli login` | **no** - 1.5.0's `console_scripts` are `hf`, `tiny-agents` | `Warning: huggingface-cli is deprecated and no longer works. Use hf instead.` (exit 1) |
| `hf auth login` | yes | the live entry point |

So on a conforming minimal install the command does not exist at all, and on a
later 1.x that does ship it, it refuses to do anything. The advice is a dead end
across the whole declared range.

`run_mujoco_agent.py` carried the check inline in its lifecycle block rather than
as a resolver, and never read `HF_TOKEN` at all - so an `HF_TOKEN`-only host, the
environment the drivers' own prose calls preferred for CI, was refused outright.

## What changes

Resolve through `huggingface_hub.get_token()` and quote `HF_TOKEN_PATH` in the
refusal, so a relocated cache is actionable rather than a mystery. Prescribe
`hf auth login` - what `doctor`, `dataset_recorder`, the README,
`docs/troubleshooting.md` and `huggingface_hub.get_token`'s own docstring already
name. `run_mujoco_agent.py` gets the same extracted resolver as its two siblings,
so the environment path reaches the download there too.

The env-var branch is unchanged and still preferred: a container that injects
`HF_TOKEN` never pays for a Hub import.

## Scope

Six prescription sites across the three drivers, plus the one line in
`examples/libero/README.md`'s env-var list that now names the cached-login
alternative. `docs/troubleshooting.md` and the existing changelog fragments
already name `huggingface-cli` only to explain that it does not work, which the
new rule permits - it keys on the two-token prescription, not the mention.

## Tests

`tests/test_examples_hf_token_preflight.py` (22 tests). Pre-fix, against
`upstream/main`: **15 failed / 7 passed**. Nine are behavioural or structural -
the logged-in host refused, the refusal naming no runnable login, the refusal not
naming the path it looked in, and the prose rule reporting all six sites:

```
{'examples/libero/run.py': [46, 270, 284],
 'examples/libero/run_isaac_agent.py': [319, 333],
 'examples/libero/run_mujoco_agent.py': [247]}
```

The other six report that `run_mujoco_agent.py` had no extracted preflight at
all; they pin the shape rather than the behaviour. Its behavioural defect is
covered separately by `TestTheEnvironmentPathReachesTheDownload`, which drives
the lifecycle function with the download stubbed and fails pre-fix with the token
`RuntimeError`.

Each guard fires for its own regression and nothing else:

| break | failed of 22 | fires |
| --- | --- | --- |
| control | 0 | - |
| exact pre-fix (`git checkout upstream/main -- examples/libero/`) | 15 | everything |
| Hub resolution reverted, live command kept | 6 | the 3 cached-login and 3 path tests only |
| live command reverted, Hub resolution kept | 5 | the 3 command tests, the prose rule, the naming control |
| `run_mujoco_agent.py` reverted only | 9 | its 7 cases, the prose rule, the naming control |
| prose scan reaches nothing | 2 | the non-vacuity test only - the rule itself passes *vacuously* |
| rule keys on bare `huggingface-cli` instead of the two-token phrase | 2 | the naming control, plus the rule over-reporting |
| refusal drops the resolved path | 3 | the 3 path tests only |

The middle two are complementary, which is what shows the resolution half and the
remedy half are independently pinned. The prose rule passing vacuously under an
empty scan is why its non-vacuity clause is not decoration.

## Verification

Gate: 182 test files - every test that touches `examples/`, reads a docs page,
names an HF token, or walks the source tree - run with the identical selection on
both trees, this branch's new file excluded from each. Byte-identical:

| tree | result |
| --- | --- |
| this branch | `12 failed, 8086 passed, 86 skipped` in 354.72s |
| `upstream/main` | `12 failed, 8086 passed, 86 skipped` in 354.61s |

Two of the twelve differ *by id* between the trees, and both are in
`tests/mesh/test_zenoh_transport_security.py` failing with `zenoh.ZError`
(`Address already in use`) - two suites in parallel on a fixed localhost port.
That file is `9 passed, 2 skipped` run alone on **both** trees. The other ten are
dependency-absent locally: eight in `tests/mesh/test_iot_connect_timeout_domain.py`
on `awsiot` (the `[mesh-iot]` extra declares `awsiotsdk>=1.21.0,<2.0.0`, which CI
installs) and two in `tests/test_doctor.py` on `PackageNotFoundError`, since a
git worktree is not pip-installed.

`ruff check` and `ruff format --check` clean. `mypy strands_robots tests
tests_integ`: 24 errors on this branch and 24 on a pristine `upstream/main`
worktree, none in the changed files. Removing the hardcoded path left
`run_mujoco_agent.py`'s function-scope `from pathlib import Path` unused, which
`ruff` flagged - the import existed only to build that path.

## Artifact

One capture script per tree, run against the same host and the same
`huggingface_hub`: it relocates the HF cache, writes a token where the Hub looks
for it, drives each driver's own preflight, and then asks each tree's refusal
which login it names.

HF token preflight, before and after

The lower table is measured once and is identical for both trees - the figure
script asserts that, along with the premise that the Hub really does resolve the
relocated token, so the picture cannot drift from the claim. What differs between
the columns is only which token the preflight finds and which login it names.

### DIFF
--- a/changelog.d/2652-libero-hf-token-resolution.md
+++ b/changelog.d/2652-libero-hf-token-resolution.md
@@ -1,0 +1,24 @@
+### Fixed: a LIBERO driver's HF-token preflight resolves what the Hub resolves
+
+The three `examples/libero` drivers refuse `--policy groot` up front when no
+HuggingFace token is available, rather than letting the gated Cosmos-Reason2-2B
+download fail later inside `gr00t_inference`. That refusal read
+`~/.cache/huggingface/token` directly, so on any host that relocated its HF cache
+it named a file the Hub will not open: a box that was logged in got refused, three
+drivers out of three. It now asks `huggingface_hub.get_token()` and quotes
+`HF_TOKEN_PATH` in the refusal, so a relocated `HF_HOME` / `XDG_CACHE_HOME` /
+`HF_TOKEN_PATH` is honoured and, when there really is no token, the message names
+the path it looked in.
+
+The remedy it offered could not be run either. `huggingface-cli` is not published
+as a console script at the declared `huggingface_hub>=1.5` floor at all - 1.5.0's
+`console_scripts` are `hf` and `tiny-agents` - and the later 1.x releases that do
+install it exit "deprecated and no longer works". The drivers now prescribe
+`hf auth login`, which is what `doctor`, `dataset_recorder`, the README,
+`docs/troubleshooting.md` and `huggingface_hub.get_token`'s own docstring already
+name.
+
+`run_mujoco_agent.py` carried the check inline in its lifecycle block rather than
+as a resolver, and never read `HF_TOKEN` at all, so an `HF_TOKEN`-only host - the
+environment the drivers' own prose calls preferred for CI - was refused outright.
+It now shares the same resolver as its two siblings.
--- a/examples/libero/README.md
+++ b/examples/libero/README.md
@@ -94,7 +94,10 @@ evaluates the `mujoco` and Isaac single-env (`isaac-1`) rows.
 - `STRANDS_GR00T_IMAGE` / `STRANDS_GR00T_IMAGE_ALLOW` - operator-configured
   GR00T docker image + allowlist (`--image` sets these).
 - `HF_TOKEN` (or `HUGGING_FACE_HUB_TOKEN`) - HuggingFace token for the gated
-  GR00T checkpoint download (`--policy groot`).
+  GR00T checkpoint download (`--policy groot`). With neither set, the drivers
+  fall back to the Hub's cached login (`hf auth login`), asking
+  `huggingface_hub` where that lives - so a relocated `HF_HOME` /
+  `XDG_CACHE_HOME` / `HF_TOKEN_PATH` is honoured.
 - `STRANDS_GR00T_SERVER_SEED` / `STRANDS_GR00T_STRICT_DETERMINISTIC` - consumed
   by the packaged determinism wrapper
   (`strands_robots/policies/groot/server_wrapper.py`) inside the GR00T
--- a/examples/libero/run.py
+++ b/examples/libero/run.py
@@ -42,9 +42,9 @@ Usage
     #    starts the inference server before the eval, then tears down on
     #    exit. Each step is idempotent so re-runs are cheap.
     #
-    #    Pre-condition: an HF token (the HF_TOKEN env var, or
-    #    `~/.cache/huggingface/token` from `huggingface-cli login`) with
-    #    access to `nvidia/Cosmos-Reason2-2B` (the gated VLM backbone) +
+    #    Pre-condition: an HF token (the HF_TOKEN env var, or the Hub's
+    #    cached login from `hf auth login`) with access to
+    #    `nvidia/Cosmos-Reason2-2B` (the gated VLM backbone) +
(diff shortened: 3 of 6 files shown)


yinsong1986 (approved):
## Summary

Examples-only fix to the three `examples/libero` drivers' `--policy groot` HF-token preflight: token resolution now goes through `huggingface_hub.get_token()` instead of a hardcoded `~/.cache/huggingface/token` read (so a relocated `HF_HOME` / `XDG_CACHE_HOME` / `HF_TOKEN_PATH` is honoured), the refusal names the path the Hub actually resolves plus the live `hf auth login` entry point instead of the `huggingface-cli login` command that is not published at the declared `huggingface_hub>=1.5` floor, and `run_mujoco_agent.py` gains the same extracted resolver as its siblings so an `HF_TOKEN`-only host reaches the download. No public API, wire format, or persisted schema is touched; the env-var branch stays first so containers never pay the Hub import. Verified locally at the head SHA: the new 22-test file passes, the pre-existing `tests/test_examples_libero_drivers.py` / `tests/test_examples_mujoco_gl.py` still pass, the removed function-scope `pathlib` import has no remaining users in `run_mujoco_agent.py`, the added lines are all ASCII, and the tree-wide prose rule has no remaining `huggingface-cli login` prescriptions to report.

## What's good

- Refusal messages name the resolved `HF_TOKEN_PATH` and a command that actually runs, per the no-dead-end-remedy expectations in AGENTS.md.
- `Raises:` blocks added to every resolver docstring, matching the docstring-completeness convention.
- Changelog fragment under `changelog.d/` rather than a direct `CHANGELOG.md` edit; README env-var list updated in the same PR.
- The prose-rule test keys on the two-token prescription so troubleshooting docs may still explain why `huggingface-cli` is dead, and it carries its own non-vacuity clause.
- Tests isolate the Hub's token resolution onto `tmp_path` via `monkeypatch` and restore the `sys.modules` entries they add.

PR #2653, opened 2026-08-23
fix(simulation/mujoco): an add_geom op asking for a mesh is refused, not compiled

### LINKED ISSUE #2310: decision(simulation/mujoco): `_normalize_size`'s mesh branch returns `[0, 0, 0]` while its docstring says `[]`, and the build path never calls it
Found while measuring the mesh `size` divergence for #2300 and deliberately left out of #2309, so a docs fix does not smuggle a code change. Two facts that cannot both be intended.

## Measured

`_normalize_size` in `strands_robots/simulation/mujoco/spec_builder.py` has a mesh branch:

```python
if shape == "mesh":
    return [0.0, 0.0, 0.0]
```

**1. Its own docstring documents a different return.** The conventions block above it reads:

```
* ``mesh``:      ``[]``            (mesh asset dictates extent; size ignored)
```

Measured: `_normalize_size("mesh", [2.0, 3.0, 4.0])` returns `[0.0, 0.0, 0.0]`, not `[]`. #2300's own write-up repeated the docstring's `[]` as the measured behaviour, which is how far that claim travels.

**2. `SpecBuilder.build` never routes a mesh through it.** The sole call site there is guarded the other way:

```python
if obj.shape == "mesh":
    geom_kwargs["meshname"] = f"mesh_{obj.name}"
else:
    geom_kwargs["size"] = _normalize_size(obj.shape, list(obj.size))
```

So on the build path the mesh geom carries no `size` key at all - which is the behaviour both the docstring and the docs describe, reached by never calling the function rather than by its return value.

The other call site is `strands_robots/simulation/mujoco/scene_ops.py:1730`, which calls it unconditionally on a caller-supplied `shape`:

```python
"size": _normalize_size(shape, op.get("size", [0.1, 0.1, 0.1])),
```

## What needs deciding

Whether that branch is reachable at all, and if so what it should emit:

* **Reachable via `scene_ops`** - then `[0.0, 0.0, 0.0]` is a live value being written as a mesh geom's `size`, and the question is whether MuJoCo treats an explicit all-zero size on a mesh geom identically to an absent one (worth measuring against a compiled model, not assumed), and the docstring is simply wrong.
* **Not reachable** - then it is dead code under convention 10, and deleting it makes the `raise ValueError(f"Cannot normalize size for shape {shape!r}.")` fall-through the correct answer for a mesh: nothing should be asking.

Either way the docstring's `[]` should stop being the third statement of this contract. Cheap to settle - one guard on the `scene_ops` path answers the reachability question - and worth settling before #2300 picks a mesh `size` meaning, since option 1 there (MuJoCo honours `size` as a mesh scale) would make this the function that has to carry it.

## Not blocking

No user-visible defect is claimed here: on the documented `add_object` path the mesh geom gets its extent from the asset, which is what `docs/simulation/world-building.md` says. This is an internal contract stated three ways, one of which is measurably false.

per #2300, per #2309

cagataycali:
Closes #2310

## What #2310 asked

#2310 named two facts about `_normalize_size`'s mesh branch that cannot both be intended - it returns `[0.0, 0.0, 0.0]` where its own docstring documents `[]`, and `SpecBuilder.build` never routes a mesh through it - and left one question open: **is the branch reachable at all?** It named the one candidate call site, `scene_ops`'s `add_geom` patch op, and asked for a guard on it.

It is reachable. And reaching it is worse than a wrong return value.

## Measured

`add_geom` resolves its `type` through the full shape vocabulary, `"mesh"` included, while `_PATCH_OP_KEYS` gives the op no key that could name a mesh asset. So a well-formed op - every key accepted, every numeric field in domain - adds a mesh geom with no meshid, and MuJoCo refuses the batch at recompile. On a default world, before this change:

```
sim.patch_scene_mjcf([{"op": "add_geom", "body": "world", "type": "mesh", "name": "part"}])
-> error: MJCF patch failed: Error: mesh geom 'part' (id = 1) must have valid meshid

sim.patch_scene_mjcf([{"op": "add_body", "name": "crate", "pos": [0, 0, 1]}])
-> error: MJCF patch failed: Error: mesh geom 'part' (id = 1) must have valid meshid

sim.patch_scene_mjcf([{"op": "add_body", "name": "crate2", "pos": [0, 0, 2]}])
-> error: MJCF patch failed: Error: mesh geom 'part' (id = 1) must have valid meshid

sim.add_object("ball", shape="sphere", size=[0.1], position=[0, 0, 1])
-> error: Failed to inject 'ball' into live scene: Error: mesh geom 'part' (id = 1) must have valid meshid
```

Two things are wrong with that, and only the first is the one #2310 set out to fix:

1. **The refusal names a MuJoCo element id, not the op or the key that is missing.** Nothing in it says `add_geom` cannot reference a mesh asset, and nothing points at the surface that can.
2. **The world never recovers.** `patch_scene_mjcf` snapshots the spec and restores it if any *op* raises, but `spec.recompile(...)` runs **outside** that `try`. A compile-time refusal therefore discards the snapshot and leaves the mutated spec installed, so every later mutation of that world re-fails on the leftover geom. `docs/simulation/world-building.md` and the function's own docstring both promise atomicity; this path is where that promise was false.

The nearest neighbour is *not* affected, which is what makes the mesh value specific rather than an instance of a wider class: a duplicate geom name raises inside `MjSpec.add_geom`, i.e. inside the op loop, so it rolls back and the world stays usable (`patch op #2 failed: Error: repeated name 'g' in geom`, followed by a successful patch). A `plane` geom in a non-world body and a `delete_body` of a body added in the same batch both compile. `type="mesh"` is the only op field value measured to clear every op-level check and then fail the recompile.

## The change

**`add_geom` refuses `type="mesh"` before it touches the spec.**

```
sim.patch_scene_mjcf([{"op": "add_geom", "body": "world", "type": "mesh", "name": "part"}])
-> error: patch op #1 failed: add_geom: 'type' cannot be 'mesh' - this op has no key that
   names a mesh asset, so the geom it would add has no mesh to take its extent from and
   MuJoCo refuses the whole scene at recompile. Add a mesh with
   add_object(shape="mesh", mesh_path=...), which registers the asset alongside the body.
   Accepted 'type' values: box, capsule, cylinder, ellipsoid, plane, sphere.
```

After it, the world is untouched: the two follow-up patches and the `add_object` above all return `success`.

The accepted list is built from the per-shape table rather than written as a literal, so it cannot drift from the vocabulary `_geom_type` actually resolves - and the new module pins that by compiling one geom of every shape the message offers, plus that the two tables share one vocabulary. A shape outside the vocabulary altogether is deliberately left to `_geom_type`, whose message already names it and which raises inside the loop where the rollback applies.

**`_normalize_size`'s mesh branch is deleted, not corrected.** With the patch op refusing a mesh, no call site normalizes one: `SpecBuilder.build` passes `meshname` instead of calling the function. Convention 10 applies, and #2310's second option is the one the measurement supports - a mesh now falls through to the existing `raise ValueError(f"Cannot normalize size for shape {shape!r}.")`, which is what "nothing should be asking" looks like. The docstring's `[]` stops being a third statement of the contract.

Also measured while settling this, against a compiled model rather than assumed (#2310 asked for exactly this): **MuJoCo overwrites a mesh geom's `size` with the asset's own extent whatever was supplied.** A tetrahedron mesh geom compiled with `size=[0, 0, 0]`, with no `size` key, and with `size=[2, 3, 4]` all report the same `geom_size`. So the branch's return value was never observable even in principle - which is why nothing noticed it disagreed with its docstring.

**`add_object(shape="mesh", mesh_path=...)` is unchanged** and now has a test asserting the extent still comes from the asset, so the deletion cannot quietly reach the documented route.

## Deliberately not in this PR

The atomicity gap - a refused **recompile** discarding the snapshot - is the general form of finding (2) and it survives this change: `patch_scene_mjcf` still recompiles outside its `try`, and moving it in needs the snapshot-restore path re-measured against `install_compiled_model`. This PR closes the one route by which a fully validated op reached it; the residual is filed separately so a narrow refusal does not smuggle a change to the batch's failure semantics.

## Verification

- `tests/simulation/mujoco/test_patch_op_geom_shape_domain.py` - 8 new tests: the refusal names the value and the working route and does not leak MuJoCo's message; it identifies which op failed; the world survives it (two patches + an `add_object`); every shape the message offers compiles; the two shape tables agree; an unknown shape is still `_geom_type`'s to refuse; `_normalize_size` refuses a mesh; the documented mesh route still takes its extent from the asset.
- `tests/simulation/mujoco/test_spec_builder.py` - `test_mesh_returns_zeros` becomes `test_mesh_is_refused_because_no_call_site_normalizes_one`.
- `tests/simulation/mujoco/` full run, mujoco 3.12.0, headless with no GL context: no new failures against `main` (the pre-existing failures in that environment are OpenGL- and torch-dependent; the set is identical minus one order-sensitive flake).
- `tests/simulation/test_mesh_size_docs_match_backend_divergence.py`, `test_object_size_domain_across_backends.py`, the changelog-fragment gates, and the patch-op suite: 313 passed.
- `ruff check`, `ruff format --check`, `mypy` clean on the touched files.

## Changelog

`changelog.d/2310-patch-op-mesh-geom-refused.md`.

---

### Round 1

Initial submission.


yinsong1986 (approved):
## Summary

Closes a world-poisoning path in `patch_scene_mjcf`: an `add_geom` op with `type="mesh"` cleared every op-level check but had no key to name a mesh asset, so MuJoCo refused the batch at recompile - outside the rollback `try` - leaving the mutated spec installed and every later mutation of that world re-failing on the leftover geom. The op now refuses `"mesh"` in `_apply_patch_op` before touching the spec, so the raise lands inside the op loop where the snapshot restore applies and the refusal names the value, the failing op, and the working route (`add_object(shape="mesh", mesh_path=...)`). `_normalize_size`'s unreachable mesh branch (which returned `[0.0, 0.0, 0.0]` against a docstring saying `[]`) is deleted rather than corrected, and `git grep` at head confirms its only two call sites are `SpecBuilder.build` (which passes `meshname` for meshes) and the now-guarded patch op - so nothing is stranded. The accepted-shape list in the message is derived from `_SIZE_LAYOUT` rather than a literal, and the new test module pins the table/`_geom_type` agreement, compiles every offered shape through the op, and asserts the world survives the refusal. The deliberately deferred residual (recompile still outside the rollback `try`) is scoped out explicitly and filed separately, which is the right call for a narrow fix.

## What's good

- The refusal raises inside the op loop, so the documented atomicity is actually true for this path - verified against the head source, not just the diff.
- Deriving the accepted list from `_SIZE_LAYOUT` and pinning the two-table agreement means the message cannot drift from the vocabulary the op resolves.
- Changelog fragment present, tool_spec.json change is description-only (no parameter/schema change), diff is ASCII-clean, test names describe behaviour, `pytest.importorskip("mujoco")` and fixture cleanup in place.
- A shape outside the vocabulary is deliberately left to `_geom_type`, whose raise already sits where the rollback applies - the narrow guard does not widen into a second vocabulary owner.

PR #2655, opened 2026-08-23
fix(sim/mujoco): route a patch batch's recompile through the shared helper

### LINKED ISSUE #2654: patch_scene_mjcf recompiles outside its own rollback, so a batch refused at compile time leaves the mutated spec installed
`patch_scene_mjcf` documents its batch as atomic - "if any op raises, the whole patch is rejected and the world is left in its original state" - and `docs/simulation/world-building.md` repeats it: "The batch is atomic: if any op is rejected the world is rolled back to its pre-patch state." Both claims are scoped narrower than they read.

## Measured

The snapshot restore covers the op loop only:

```python
backup_spec = _snapshot_spec(spec, context="patch_scene_mjcf")
...
try:
    for op in ops:
        _apply_patch_op(spec, op, new_bodies)
        applied += 1
except Exception as err:
    world._backend_state["spec"] = backup_spec
    raise ValueError(f"patch op #{applied + 1} failed: {err}") from err

# One recompile for the whole batch - preserves qpos/qvel for unchanged joints.
with filter_mujoco_attach_noise():
    new_model, new_data = spec.recompile(world._model, world._data)
```

`spec.recompile` is outside the `try`. So a batch whose ops all apply and whose *compile* is refused takes neither branch: the exception propagates, `backup_spec` is dropped on the floor, and `world._backend_state["spec"]` keeps the mutated spec. That spec is what every later mutation recompiles from, so the failure is permanent rather than per-call.

Measured before #2653 closed the one route that reached it (`{"op": "add_geom", "type": "mesh"}`), on a default world:

| call after the refused batch | result |
|---|---|
| a valid `patch_scene_mjcf` (`add_body`) | `error: ... mesh geom 'part' (id = 1) must have valid meshid` |
| a second valid `patch_scene_mjcf` | same error |
| an unrelated `add_object("ball", shape="sphere", ...)` | `Failed to inject 'ball' into live scene: ... must have valid meshid` |

The world was unusable for the rest of its life, and nothing in any of those three messages says why - each names the leftover geom from a batch the caller was told had failed.

## Why it stayed invisible

`tests/simulation/mujoco/test_patch_scene_mjcf_rejected_batch_recovery.py` exists precisely to pin that a rejected batch leaves a clean live spec, and its module docstring states the assumption that exempts this path:

> `tests/simulation/mujoco/test_patch_scene_mjcf.py` pins that the *compiled model* is unchanged after a rejected batch. That assertion is nearly free -- **a rejected batch never reaches the recompile** [...]

That sentence is true of every rejection the suite constructs and false of the class as a whole. It is the sentence to fix alongside the code.

## Current reachability

After #2653 there is no *measured* op value that clears every op-level check and then fails the recompile. Probed on a default world:

| batch | where it fails | world afterwards |
|---|---|---|
| duplicate geom `name` | inside `MjSpec.add_geom`, i.e. inside the loop | rolled back, next patch succeeds |
| `plane` geom in a non-world body | compiles | fine |
| `delete_body` of a body added in the same batch | compiles | fine |
| `add_geom` with `type="mesh"` | refused at the door as of #2653 | fine |

So the class is currently unwitnessed. That is an argument for closing it structurally rather than for leaving it: the op vocabulary is deliberately narrow but not frozen, `_apply_patch_op` is where new ops land, and the next one that can produce a compile-only failure inherits a permanently poisoned world with no test in either direction. The recompile is also reached by `inject_object_into_scene` / `inject_camera_into_scene` / `eject_body_from_scene`, which have their own rollback paths - whether they share this shape is part of the read.

## What needs deciding

Moving the recompile inside the `try` is one line and is not obviously sufficient, which is why this is an issue rather than a PR:

1. **What "restore" means at that point.** `install_compiled_model` has not run, so `world._model` / `world._data` still hold the previously compiled pair and are already consistent. Only `_backend_state["spec"]` is wrong. Whether putting `backup_spec` back is enough - or whether the discarded mutated spec can still be referenced from anywhere, including the cached XML `_sync_cached_xml` maintains - wants measuring rather than assuming.
2. **What the message says.** `patch op #N failed: ...` is wrong for a compile refusal: no single op is at fault, and the count would be `len(ops)`. A compile failure needs its own wording that says the batch compiled to an invalid model and names MuJoCo's reason, without claiming an op index the caller cannot use.
3. **Whether the docs' unconditional "atomic" claim is the target or the description.** If (1) lands, both prose sites become true as written; if it does not, both need the scope stated.

A guard test keyed on the structure - the recompile inside the rollback's scope, asserted behaviourally by a batch that compiles to an invalid model - is what makes this non-recurring. Constructing that batch is itself part of the work now that the mesh route is closed; `replace_scene_mjcf` with invalid XML is the neighbouring surface worth reading for how it does it.

## Not blocking

No currently reachable defect is claimed. This is the general form of the second finding in #2653, filed separately so a narrow op-level refusal does not smuggle a change to the batch's failure semantics.

per #2653, per #2310

cagataycali:
Closes #2654

## Problem

`patch_scene_mjcf` was the one scene mutation that recompiled the spec itself instead of through `_recompile_preserving_state`:

```python
with filter_mujoco_attach_noise():
    new_model, new_data = spec.recompile(world._model, world._data)
install_compiled_model(world, new_model, new_data)
```

That helper has two jobs, and this path inherited neither.

### 1. It kept none of the state the helper carries

`apply_force`'s docstring names exactly two things that end a latch: the next `apply_force` on that body, or a `reset()`. `spec.recompile` returns `xfrc_applied` zeroed, so a patch ended it too. Measured with a 1 kg crate held against gravity, patched with a batch that only adds an unrelated body:

| tree | patch | `xfrc_applied` after | crate over 0.6 s |
| --- | --- | --- | --- |
| `upstream/main` | `success` | `[0.0, 0.0, 0.0]` | 0.620 m -> 0.150 m, **fell 470 mm** |
| this branch | `success` | `[0.0, 0.0, 9.81]` | 0.620 m -> 0.620 m, **0 mm** |

The sibling suite already enumerates *"every operation that rebuilds the model"* for this contract - and listed five ops, omitting this sixth. That omission is why the gap stayed invisible: the enumeration read as complete without it.

### 2. Its refusal landed past the batch's own rollback

The ops all apply to the spec, then MuJoCo refuses the model they add up to. The raw recompile raised from *outside* the `try` that restores the snapshot, so the rejection took neither branch and the whole batch stayed on the live spec. Every later mutation recompiles from that spec, so one refused batch cost the world rather than the batch:

| tree | `patch_scene_mjcf([{"op": "add_site", "size": [0.0]}])` | then `add_object("probe")` |
| --- | --- | --- |
| `upstream/main` | `error: Error: size 0 must be positive in geom / Element name 'unbuildable'` | **`error: Failed to inject 'probe' into live scene: Error: size 0 must be positive in geom / Element name 'unbuildable'`** |
| this branch | same message | `success` |

A zero `size` is finite, so it clears the op-level `size` domain and reaches the compiler - which is what makes this reachable through the published action alone.

#2653 measured the same thing one commit earlier, from the other direction. Its comment justifying `_UNSUPPORTED_GEOM_SHAPES` reads: a mesh geom's refusal *"fires OUTSIDE the try/except that rolls a rejected batch back, so the mutated spec stays installed and every later mutation of that world - a valid patch, an `add_object` - re-fails on the same leftover geom. Measured on a default world: three successive mutations after one such patch all returned the meshid error."* That refusal is one compiler complaint among many; refusing the single value at the door made it unreachable without making the others survivable.

## Change

One call, through the helper, inside the rollback - mirroring `inject_robot_into_scene`, whose own comment records that folding a refusal into a bool "became 'spec recompile refused.' with the actionable text left in the log - the same dead end the unsupported-shape path was fixed to stop producing".

```python
try:
    _recompile_preserving_state(world, spec, raise_on_refusal=True)
except (ValueError, RuntimeError):
    world._backend_state["spec"] = backup_spec
    raise
```

Measured on mujoco 3.5.0 (the declared floor) and 3.12.0 (what CI resolves), identically: **restoring the snapshot is the whole rollback.** After a refused batch the live spec carries none of it, the compiled model and data objects are the same objects, and the cached XML is byte-identical - `_sync_cached_xml` only runs after a successful install, so it never saw the refused spec. Two further patches and an `add_object` then succeed.

The refusal message is unchanged and deliberately claims no op index: every op applied, so there is no single op at fault.

## Verification

Pre-fix split, both test files against `upstream/main`: **6 failed / 24 passed**, all six semantic.

```
the compiler-refused batch survived on the live spec
an unrelated add_object inherited the refused batch: {'status': 'error', ...
  Failed to inject 'crate' into live scene: Error: size 0 must be positive in geom
a scene_ops function recompiles the spec itself instead of through the shared
  helper: ['_recompile_preserving_state', 'patch_scene_mjcf']
patch_scene_mjcf's recompile is not inside a try, so a refused compile leaves
  the batch on the live spec
patch_scene_mjcf dropped the latched wrench: the puck fell 0.1982 m
assert [0.0, 0.0, 0.0, 0.0, 0.0, 0.0] == approx([0.0, 0.0, 4.905, ...])
```

The 24 that pass on both trees are the controls: the op-level rejection path in full, the five rebuild ops that already went through the helper, the "latch still ends where it is documented to" set, and - notably - `test_the_compiled_model_and_cached_xml_are_untouched` and `test_the_world_still_steps`, which held pre-fix too. Only the spec was ever wrong.

Every break fires a distinct set, so no guard is decoration:

| break | collected | result | fires |
| --- | --- | --- | --- |
| control (this branch) | 30 | 30 passed | - |
| exact pre-fix source | 30 | 6 failed | all six above |
| helper called *outside* the try | 30 | 3 failed | both rollback behaviours + "not inside a try" |
| try kept, handler drops the restore | 30 | 3 failed | same two + "the handler does not put the pre-patch spec back" |
| re-derive the recompile inline | 30 | 4 failed | "only the shared helper recompiles" + the try + **both wrench cases** |
| witness stops being a compile refusal | 30 | 5 failed | **all five premise asserts** |
| enumeration loses its sixth member | **28** | 28 passed | coverage vanishes silently |

The last two are the ones worth reading. A future op-level `size` domain on `add_site` would quietly turn the compile-refusal tests into op-level tests, so each asserts the refusal carried no `patch op #` - the fifth break shows all five firing. And the seventh shows the hand-listed enumeration is not self-maintaining, which is why `test_only_the_shared_helper_recompiles_a_spec` is derived from the AST: a future path that re-derives a recompile is caught by that one regardless.

Gate at `8b8dfa9b`: identical 55-file selection on both trees, the two changed test
files excluded from each, `upstream/main` restored pristine. Byte-identical summary
lines - `1 failed, 2497 passed, 1 skipped` on both, and both comm sets empty. The one
shared failure is `test_declared_qp_backends_are_real_qpsolvers_extras`, which needs
`qpsolvers` (declared in `sim-mujoco` and `cosmos3-sim`, absent from this
`--system-site-packages` interpreter). The two changed files then run **30 passed on
mujoco 3.12.0 and 30 passed on 3.5.0**, and 82 passed with random ordering beside four
sibling scene files. mypy `24 == 24` against a third pristine worktree, branch-only
empty; ruff clean; the CodeQL AST preflight reports the same single pre-existing
`_snapshot_spec` hit on both trees.

### Rebased onto a moved base

`upstream/main` gained #2652 and #2653 during this work, and #2653 touches both files
this PR touches. Merged; the auto-merge was clean and both sides are present by symbol.
Its own eight tests pass unchanged here, because the door refusal fires before the
compile either way. One consequence: its comment described this defect in the present
tense, so the second commit corrects that sentence and keeps its actual argument - the
compiler names a MuJoCo id, the door names the op. Re-measured at the merged head: the
pre-fix split is unchanged at 6 failed / 24 passed, the gate is unchanged and still
byte-identical, and mypy is `24 == 24`.

## Artifact

The crate is held up by nothing but the latched wrench, so "is it still latched" is answered by where it is. Frame 0 is **pixel-identical between the two columns** (mean abs level `0.0`); the left column's own motion over the window is `14.09` and the right column's is `0.0498`.

patch_scene_mjcf recompile

---

### Round 2

Description only: added the `Closes #2654` link this PR was missing, matching the
convention #2653 used for #2310. No code, test, docs, or changelog change; head is
unchanged at `8b8dfa9b`.


yinsong1986 (dismissed):
## Summary

Routes `patch_scene_mjcf`'s single batch recompile through the shared `_recompile_preserving_state` helper and moves it inside the batch's own rollback scope, fixing two real defects: a successful patch zeroing the latched `apply_force` wrench (the helper snapshots and re-latches both applied-force buffers, defines the positional-transfer tail, re-discovers per-robot ids, and runs the forward pass - all of which the removed inline `spec.recompile` path skipped), and a compiler-refused batch surviving on the live spec and poisoning every later mutation. The `except (ValueError, RuntimeError)` handler restores the pre-patch snapshot and re-raises, mirroring `inject_robot_into_scene`; the helper only re-syncs the cached XML after a successful install, so the new test asserting the model, data and cached XML are untouched after a refusal holds by construction. All symbols the removed inline path used remain live elsewhere in `scene_ops.py`, so nothing is orphaned.

## What's good

- The root cause is pinned structurally: `test_only_the_shared_helper_recompiles_a_spec` derives the recompile-call census from the AST, so a future path that re-derives an inline `spec.recompile` is caught regardless of the hand-listed enumeration - which the PR also demonstrates is not self-maintaining (the sixth-member omission).
- Every compile-refusal test asserts the refusal carried no `patch op #`, so a future op-level `size` domain on `add_site` turns these into loud failures instead of quietly retargeting them at the op-level path.
- The exception tuple is narrow with no covered member, the rollback handler matches the sibling `inject_robot_into_scene` pattern, and the refusal message is unchanged (deliberately claiming no op index, since every op applied).
- Changelog fragment (`changelog.d/2655-...md`), docs update, and the `_REBUILD_OPS` parametrization extension all land in the same PR; test names describe behaviour, not provenance.

yinsong1986 (approved):
## Summary

Routes `patch_scene_mjcf`'s single batch recompile through `_recompile_preserving_state` and places the call inside the batch's own rollback scope, which fixes the two defects the description measures: a successful patch zeroing a latched `apply_force` wrench (the helper snapshots and re-latches both applied-force buffers, defines the positional-transfer tail, re-discovers per-robot ids, runs the forward pass, and re-syncs the cached XML - all of which the removed inline `spec.recompile` path skipped), and a compiler-refused batch surviving on the live spec so every later mutation re-failed on it. I verified the substitution is complete against the helper's body at this head: every refusal path in the helper (`spec.recompile` caught as `(ValueError, RuntimeError)`, plus the unaddressable-actuator gate's `ValueError`) fires before `install_compiled_model`, so restoring the spec snapshot really is the whole rollback and the new test asserting model/data/cached-XML are untouched after a refusal holds by construction rather than by luck. The `except (ValueError, RuntimeError)` tuple matches the helper's documented raise surface with no covered member, and the handler mirrors the sibling `inject_robot_into_scene` pattern exactly.

## What's good

- `test_only_the_shared_helper_recompiles_a_spec` derives the recompile-call census from the AST rather than a hand-list, so the next path that re-derives an inline `spec.recompile` fails structurally - which matters, since this PR also demonstrates the hand-listed `_REBUILD_OPS` enumeration was not self-maintaining (the omitted sixth member).
- Every compile-refusal test asserts the message carries no `patch op #`, so a future op-level `size` domain on `add_site` turns these into loud premise failures rather than quietly retargeting them at the op-level path.
- The `_UNSUPPORTED_GEOM_SHAPES` comment is corrected in the same PR to stop describing the fixed defect in the present tense while keeping its surviving argument (the compiler names a MuJoCo id, the door names the op).
- Changelog fragment, docs update, and the wrench-suite extension land together; test names describe behaviour, not provenance; scope is disciplined at 5 files.

PR #2656, opened 2026-08-23
fix(isaac): read an MJCF's mesh registry from the whole model, not just the top file

cagataycali:
## Problem

MuJoCo treats `<include file=...>` as a textual splice, so `<compiler>` and `<asset>` are **model-global**: the fragment declaring the mesh search directory need not be the fragment declaring the mesh, and neither need be the top file. `assets/download._mjcf_mesh_subdir` states exactly this for the robot-asset path and names the harm of a narrower reader, that it "reports a mesh that is present as absent", and `mujoco/simulation._ensure_meshes` already follows includes for the same question.

The MJCF scene loader read only the top file's direct children (`root.findall("compiler")` / `root.findall("asset")`). Both directions were wrong, on models real MuJoCo compiles, and they failed differently:

| model | MuJoCo | loader before | loader after |
| --- | --- | --- | --- |
| `<compiler>` + `<asset>` in the top file (control) | compiles, nmesh=1 | `meshes/widget.obj`, size (0.4, 0.3, 0.2) | unchanged |
| `<asset>` in an `<include>`d fragment | compiles, nmesh=1 | **`mesh_path=None`, size (0.05, 0.05, 0.05)** | `meshes/widget.obj`, size (0.4, 0.3, 0.2) |
| `<compiler meshdir>` in an `<include>`d fragment | compiles, nmesh=1 | **the whole scene load raises** | `meshes/widget.obj`, size (0.4, 0.3, 0.2) |
| `<compiler assetdir>` in an `<include>`d fragment | compiles, nmesh=1 | **the whole scene load raises** | `assets/widget.obj`, size (1.0, 1.0, 1.0) |
| fragment in a subdirectory declaring a relative `meshdir` | compiles, nmesh=1 | **raises** | resolved against the model directory, as MuJoCo does |
| nested `<include>` | compiles, nmesh=1 | **raises** | resolved |

An `<asset>` block in an included fragment left the registry empty, so the body fell back to the hardcoded 0.05 m box proxy that #2459 removed. That path is **silent**: `load_scene` reports a successful scene, and a pixel-conditioned policy is evaluated against a 5 cm cube standing in for a 0.4 m object.

A `<compiler meshdir>` in an included fragment is louder and wider. The mesh resolved against the model directory instead, so the existence check refused a file that is present:

```
MJCF scene loader: mesh asset '<model dir>/widget.obj' for body 'widget' in <scene>.xml
is missing on disk - refusing to degrade the object to a silent box proxy
(#2459 fail-loud contract)
```

That raise leaves the scene with **no objects at all**, not one bad object, and it blames a missing asset for a mesh sitting at `meshes/widget.obj`.

The loader carried a comment asserting the first case was impossible: "A mesh reference with no `<asset>` entry cannot occur in a robosuite-COMPILED scene (MuJoCo refuses to compile it)". MuJoCo compiles it (measured above, `nmesh=1` on both 3.5.0 and 3.12.0). That false premise is what made the gap look unreachable, and it is corrected here. The scene MJCFs this loader was written for do use includes: `benchmarks/libero/adapter`'s own module docstring says "LIBERO's scene MJCFs `<include>` Panda geometry".

## Change

`_mjcf_model_toplevel` returns the model's top-level elements with `<include>` spliced in document order, resolving each include path against the *including* file's directory so nested includes chain. `_parse_mjcf_mesh_assets` reads `<compiler>` and `<asset>` from that sequence instead of from the top file alone. A missing, unreadable, malformed or cyclic include contributes nothing rather than failing the scene, matching the reasoning already in `_ensure_meshes`: MuJoCo names the offending file itself on the load that follows, which is a better report than anything this reader could invent.

Object selection is untouched: `<worldbody>` is still read from the top file only.

## Why the search directory is not delegated to `_mjcf_mesh_subdir`

`_mjcf_mesh_subdir` is the single owner of this rule on the robot-asset path, so delegating looks right. Measured against MuJoCo, it answers one case differently. MuJoCo's precedence is: within one `<compiler>` element `meshdir` overrides `assetdir`, but a later element overrides an earlier one.

| model | MuJoCo loads | `_mjcf_mesh_subdir` | loader (before and after) |
| --- | --- | --- | --- |
| one element, `meshdir` + `assetdir` | `meshes/` (4 verts) | `meshes` | `meshes/` |
| `meshdir` then `assetdir` | `assets/` (5 verts) | `meshes` | `assets/` |
| `assetdir` then `meshdir` | `meshes/` (4 verts) | `meshes` | `meshes/` |

MuJoCo's own vertex count identifies the directory it searched. The loader already matched all three, so this PR widens *what* it reads and leaves *how* it resolves alone. `TestDocumentOrderDecidesTheSearchDirectory` pins all three against MuJoCo, and a delegation to `_mjcf_mesh_subdir` fires exactly that control (break b4 below).

The divergence is latent rather than live for `_mjcf_mesh_subdir`'s own callers: it reads robot model files, and a Menagerie model carries a single `<compiler>`. Changing it is a separate contract and is not attempted here.

## Scope

Walking an included `<worldbody>` is out of scope. Today a top file whose only `<worldbody>` lives in a fragment raises "has no `<worldbody>`", and splicing bodies would change *which objects a scene reports* rather than how a declared mesh resolves. For LIBERO it would newly expose the included Panda's bodies to object selection, which is a different question with its own filtering contract.

## Tests

`tests/simulation/isaac/test_mjcf_mesh_assets_are_model_global.py`, 23 tests. The file asserts only through the public `load_mjcf_scene_objects`, so it collects on `main` and every pre-fix failure is behavioural.

Pre-fix: **6 failed / 17 passed**. One reports the silent degradation ("the mesh reference did not resolve, so the object fell back to the (0.05, 0.05, 0.05) box proxy #2459 removed") and five report the refusal ("the loader refused a model MuJoCo compiles ... A present mesh reported as absent loses every object in the scene, not just this one").

The two mesh directories hold *different* meshes, so the reported `size` proves which one was searched: `meshes/` is a tetrahedron with extent (0.4, 0.3, 0.2), `assets/` a pyramid with extent (1.0, 1.0, 1.0). `TestMuJoCoCompilesEveryFixture` pins that every fixture is a model MuJoCo accepts and agrees with on the directory, which is what makes a loader disagreement a defect rather than invalid scaffolding.

| break | fires | count |
| --- | --- | --- |
| control, as shipped | nothing | 0 of 23 |
| b1 exact pre-fix source | the 6 regression tests | 6 |
| b2 `<asset>` read from the top file only | the `<asset>`-in-a-fragment case alone | 1 |
| b3 `<compiler>` read from the top file only | the 4 compiler cases and the cycle | 5 |
| b4 delegate the search directory to `_mjcf_mesh_subdir` | `test_a_later_element_overrides_an_earlier_one` alone | 1 |
| b5 resolve a fragment's relative `meshdir` against the fragment directory | the subdirectory and nested cases | 2 |
| b6 drop the include cycle guard | `test_an_include_cycle_terminates` alone | 1 |
| b7 resolve a nested include against the model directory | the nested case alone | 1 |

b2 and b3 are complementary (1 + 5 = 6 = b1), so the `<asset>` half and the `<compiler>` half are independently pinned. Four breaks fire exactly one test each.

## Verification

Measured at `73a04539`.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`. `mypy` 24 pre-existing errors on this branch and on a pristine `upstream/main` worktree, none in the changed files.

71-file blast radius (every test naming the loader, `_mjcf_mesh_subdir`, `mesh_aabb`, `load_mesh_geometry`, the box proxy or `_ensure_meshes`, plus all of `tests/simulation/isaac` and `tests/benchmarks/libero` and the repo-wide guards), the identical selection on both trees with the new file excluded from each:

| tree | result |
| --- | --- |
| this branch | 2418 passed, 43 skipped, 0 failed (41.44s) |
| pristine `upstream/main` | 2418 passed, 43 skipped, 0 failed (41.69s) |

Byte-identical summary lines, both comm sets empty, no failures on either side.

The new file: 23 passed on mujoco 3.12.0 and on 3.5.0 (the declared floor), and 71 passed with random ordering enabled beside its `tests/simulation/isaac` siblings.

## Artifact

`load_mjcf_scene_objects` answers `(mesh_path, size)` per body, so the artifact renders exactly that: a MuJoCo scene built from the loader's own answer, in both trees, with a fixed 10 cm reference cube beside it. The declared mesh is a hexagonal prism with extent (0.30, 0.225, 0.10) sitting at `meshes/widget.obj`, and MuJoCo compiles the model in both cases.

what the MJCF scene loader hands the realization

Before, the object is the 5 cm box proxy (2795 object pixels) and `load_scene` reports success. After, it is the declared mesh (26413 pixels), a 9.5x difference. The control row is the same model with the `<asset>` block in the top file: those two renders are identical across the trees, mean absolute level difference `0.000000`.


yinsong1986 (approved):
## Summary

The Isaac MJCF scene loader read `<compiler>` and `<asset>` only from the top file's direct children, while MuJoCo splices `<include file=...>` and treats both as model-global. This PR adds `_mjcf_model_toplevel`, which returns the model's top-level elements with includes spliced in document order (each path resolved against the including file's directory so nested includes chain), and points `_parse_mjcf_mesh_assets` at that sequence. That closes both failure directions on models MuJoCo compiles: an `<asset>` in a fragment no longer silently degrades the body to the 0.05 m box proxy #2459 removed, and a `<compiler meshdir>` in a fragment no longer makes the existence check refuse a mesh that is present (which lost the whole scene, not one object). Precedence is deliberately unchanged and now pinned against MuJoCo's own choice (within one element `meshdir` beats `assetdir`; across elements the last in document order wins), which is also the measured reason the search directory is not delegated to `_mjcf_mesh_subdir`. An unusable include contributes nothing rather than failing the scene, with the rationale documented and matching `_ensure_meshes`; the false "MuJoCo refuses to compile it" comment is corrected. Object selection (`<worldbody>` from the top file only) is untouched, and the changed function is private, so no public surface moves.

## What's good

- I reproduced the regression suite locally against this head: 23/23 pass with the fix, and the same file against the pre-fix loader reports exactly the claimed 6 failed / 17 passed, all through the public `load_mjcf_scene_objects` surface.
- `TestMuJoCoCompilesEveryFixture` pins that every fixture is a model real MuJoCo accepts and agrees with on the searched directory (vertex count discriminates the two meshes), so a loader disagreement is a defect by construction rather than scaffolding.
- The include-follower's failure handling is narrow (`except (ET.ParseError, OSError)`) and its degrade-quietly choice is justified in the docstring by the better report MuJoCo itself produces on the load that follows - consistent with the reasoning already shipped in `mujoco/simulation._ensure_meshes`.
- Changelog fragment present per AGENTS.md step 3; test names describe behaviour; no host paths or non-ASCII in user-facing strings.

## Verification suggestions

- One nit worth a glance on the author's next push (not blocking): in `test_an_included_declaration_resolves_the_mesh`, `extent = _resolved_extent(scene)` shadows the parametrized `extent`, so the final `assert extent == pytest.approx(extent)` compares the resolved value to itself. The pre-fix failures all fire through `_resolved_extent`'s degradation checks (verified locally), and the directional precedence cases have their own properly-asserted tests, so the fix is still pinned - but renaming the local (e.g. `resolved = _resolved_extent(scene)`) would restore the which-directory discrimination the parametrized fixtures were designed to carry (e.g. the `assetdir`-in-a-fragment case would then fail if the loader searched `meshes/`).

PR #2657, opened 2026-08-23
fix(simulation): name an empty model source instead of diagnosing the robot name

cagataycali:
## Problem

`MuJoCoSimEngine.add_robot` read its model source by truthiness:

```python
resolved_path = urdf_path
if not resolved_path and data_config:
    ...
elif not resolved_path and name:
    # DEPRECATED name-as-registry-key fallback
```

So a model source the caller **supplied but left empty** was indistinguishable from one
they **omitted**. `urdf_path=""` took the absent branch, ran the deprecated
name-as-registry-key fallback for a call that had asked for a file, and the refusal then
diagnosed the *name*:

| call | message |
| --- | --- |
| `add_robot(name="probe", urdf_path="")` | `No model found for 'probe'. Did you mean: ur5e, spot, rby1? Use action='list_urdfs' ... Or pass data_config=<registered model> or urdf_path=<file> ...` |
| `add_robot(name="probe", data_config="")` | byte-identical to the above |
| `add_robot(name="probe")` (nothing supplied) | **byte-identical to the above** |
| `add_robot(urdf_path="")` | `Either urdf_path or data_config is required.` |
| `add_robot()` | **byte-identical to the above** |
| `add_robot(name="probe", urdf_path=" ")` | `File not found:  ` |

Three things are wrong at once, and the last row is the sharpest: **one whitespace
character away the diagnosis was already correct.**

1. It advises the caller to `pass data_config=<registered model> or urdf_path=<file>` -
   the kwarg they had just passed. `_unknown_model_msg`'s own docstring exists to avoid
   "a dead-end 'supply a model source'".
2. It offers close-match suggestions (`ur5e, spot, rby1`) for a name the caller never
   asked to resolve - they asked to load a file.
3. It is **byte-identical to supplying no source at all**, so the report cannot tell
   "you passed an empty path" from "you passed no path".

`add_robot` is in the published action enum and `urdf_path`/`data_config` are published
`string` properties, so a schema-constrained agent can emit either.

## Why this is the odd site out

Everything else in the family already reads the source by absence:

| surface | read | verdict |
| --- | --- | --- |
| Newton `add_robot` | `if urdf_path is not None:` | correct |
| Isaac `add_robot` | `usd_path is None` / `is not None` | correct |
| `register_urdf` (MuJoCo **and** Newton) | `if not urdf_path: "'urdf_path' must be a non-empty string."` | correct, refuses by name |
| MuJoCo `add_robot` | `if not resolved_path` | **this PR** |

And the rule is stated three times inside the same file:

- `add_robot`'s own `name` paragraph: *"another falsy value (`0`, `[]`) would take the
  same branch and report success under a label that was never asked for"*.
- 50 lines below the defect: *"`is None` means 'omitted' -> the documented default. `or`
  would additionally ... read an empty vector as omitted"* (for `position`/`orientation`).
- `register_urdf`, on this same parameter.

`_unknown_model_msg`'s docstring says **"Three conditions reach this message"** and
enumerates three; a supplied-empty source was a fourth. This PR restores that census to
three rather than needing a new claim.

## Change

Refuse a supplied-but-empty model source naming that parameter, mirroring
`register_urdf`'s wording:

```
add_robot: 'urdf_path' must be a non-empty string. Pass a model source, or omit
urdf_path= to resolve the model from the registry instead.
```

The guard sits **after** the existing name-taken check, so no settled precedence moves: a
caller who both reuses a label and empties the path is still told about the label. `" "`
stays a path (`File not found`), matching `register_urdf`'s `if not urdf_path` boundary
exactly - this refuses the empty value, not an unusable one.

`AGENTS.md` gets the rule, written as a **refinement** of the existing selector section.
That section says a scalar path may be read by truthiness *"because empty and absent
genuinely coincide there - the value is derived either way"*, which read literally
licenses this defect. It holds where absence derives the same kind of value
(`examples/wbc/motionbricks_g1_mujoco.py` declares `--scene-xml` with a `""` default and
derives it from `--result-dir`, so `""` IS its sentinel and truthiness is right there). It
does not hold where absence selects a different **resolution path**, because then the two
spellings pick a different thing to blame when the call fails.

## Tests

`tests/simulation/mujoco/test_add_robot_unknown_model_message.py` already owns this
contract (its docstring covers "an instance label given *without* a model source"). Nine
new items; **9 failed / 9 passed** against `upstream/main`, every failure behavioural with
the real message quoted, e.g.:

```
assert 'No model found for' not in 'No model found for 'probe'. Did you mean: ur5e, spot, rby1? ...'
assert "'urdf_path'" in 'Either urdf_path or data_config is required.'
```

Break table - 5 breaks, 5 distinct firing sets, zero dead guards:

| break | fires |
| --- | --- |
| control (this branch) | 0 of 18 |
| exact pre-fix (`git checkout upstream/main -- simulation.py`) | 9 - the whole regression set |
| guard covers only `urdf_path` | **4** - the three `data_config` cases + the signature-completeness pin |
| drop `is not None` (over-reach onto an *omitted* source) | **11, including 5 pre-existing tests** |
| guard moved *before* the name-taken check | **1** - the precedence control alone |
| guard strips whitespace | **1** - the whitespace boundary control alone |

The over-reach row is the one worth reading: the shipped suite itself refuses it
(`test_positional_typo_is_actionable`, `test_typo_offers_close_match_suggestion`,
`test_instance_label_without_model_points_at_both_options`,
`test_no_args_keeps_generic_model_source_message` all fire), which is what bounds the rule
to a *supplied* empty value. The two single-firing rows are what make the precedence and
whitespace controls load-bearing rather than decoration.

The completeness pin derives the graded set from `inspect.signature(add_robot)`, so a
third model source added later is refused-when-empty too or the test fails until it is
graded.

## Verification

Measured on `/tmp` worktrees, mujoco 3.12.0, `MUJOCO_GL=egl`.

| gate | result |
| --- | --- |
| 384-file selection, **identical on both trees** with the changed test file excluded from each | `5 failed, 10348 passed, 201 skipped` **byte-identical** (556.61s vs 557.29s, run in parallel) |
| failing-ID sets | identical - `comm` empty in **both** directions |
| the changed test file | 18 passed on this branch; **9 failed / 9 passed** on `upstream/main` |
| `mypy strands_robots tests tests_integ` | 24 == 24 against a pristine worktree, branch-only empty |
| `ruff check` + `ruff format --check` over the CI scope | clean |

The 5 shared failures are import-order artifacts of the wide selection, not regressions:
four in `test_recording_frame_loss_is_not_tolerated.py` and
`test_rendering_pacers_survive_a_clock_step.py[forward_30s]`, all
`KeyError: 'strands_robots.simulation.policy_runner'` or the clock-double binding assertion.
Those two files are **29 passed in isolation on the pristine base**.

A CodeQL AST pre-flight over both changed files reports the same hits on both trees for
`simulation.py` (two `py/empty-except` sites, line-shifted by exactly the +41 lines added)
and nothing for the test file.

## Artifact

One caller, one mistake (`urdf_path=""` from an unexpanded shell variable), on both trees.
After the refusal the capture script does exactly one blind thing: intersect the names the
refusal **quotes** with the parameters it passed empty, and retry the one it finds. Nothing
is hardcoded per column.

add_robot empty model source

| measured | `upstream/main` | this branch |
| --- | --- | --- |
| names quoted in the refusal | `'arm'`, `'list_urdfs'` | `'urdf_path'` |
| a parameter the caller can fix | **none** | `'urdf_path'` |
| retry | not possible | `urdf_path=<the real scene.xml>` |
| outcome | **dead end**, `robots=[]`, nothing rendered | `success`, `robots=['arm']`, rendered |

Controls, asserted inside the figure script from the captured JSON:

- Passing the real path from the start returns a **byte-identical** success message on both
  trees, and the two renders agree to `0.000137` mean level.
- Following this branch's remedy lands the caller `0.000091` from having passed the real
  path first time, so the remedy is not merely actionable - it is equivalent to getting it
  right.

Neither spelling ever loaded a wrong model: both error either way. What changed is which
parameter the caller is sent to fix.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1738,6 +1738,32 @@ which side the enum is on.
   and by the render path's `cameras` resolution, which has read the same kind of selector
   `is None` all along - an empty camera selection there resolves to no camera rather than
   to every one.
+### A model source is read by absence when absence selects another resolution path
+- **The selector rule above allows a scalar path to be read by truthiness "because empty
+  and absent genuinely coincide there - the value is derived either way". That holds where
+  absence derives the SAME kind of value.** `examples/wbc/motionbricks_g1_mujoco.py`
+  declares `--scene-xml` with a `""` default and derives the scene from `--result-dir`, so
+  `""` IS its sentinel there and truthiness is correct. It does not hold where absence
+  selects a DIFFERENT RESOLUTION PATH: then the two spellings do not pick the same value
+  by another route, they pick a different thing to blame when the call fails.
+- **`add_robot`'s model source is that case.** Absent means "resolve from `data_config`, or
+  from the instance label via the deprecated registry short form"; supplied means "load
+  this file". Read by truthiness, `urdf_path=""` took the absent branch, so the refusal
+  diagnosed the NAME - close-match suggestions for a name the caller never asked to
+  resolve, and "Or pass data_config=<registered model> or urdf_path=<file>" to a caller who
+  had just passed exactly that. It was byte-identical to what supplying no source returns.
+- **The tell is the REPORT, not the value.** Both spellings error either way, so nothing
+  loads a wrong model; what diverges is which parameter the caller is sent to fix. One
+  whitespace character away (`urdf_path=" "`) the diagnosis was already correct ("File not
+  found"), which is the shape to match - refuse the empty value, not the unusable one.
+- **Check the whole family, and note that most of it was already right.** Newton reads
+  `urdf_path is not None`, Isaac reads `usd_path is None`, and `register_urdf` refuses an
+  empty `urdf_path` by name in both backends that expose it; MuJoCo's `add_robot` was the
+  one site reading the source by truthiness. Refuse behind any existing precedence (the
+  name-collision report still wins) so no settled message moves.
+- Pinned by `tests/simulation/mujoco/test_add_robot_unknown_model_message.py`, whose
+  controls pin the boundary (whitespace stays a path), the untouched short forms (`name=""`
+  still derives a label, an omitted source still diagnoses the name) and the precedence.
 ### A resolution knob is validated before the work it sizes
 - **Check the knob that sizes an expensive result before producing the result.** A
   resolution is caller input like any other, and the numeric domains
--- a/changelog.d/2657-add-robot-empty-model-source.md
+++ b/changelog.d/2657-add-robot-empty-model-source.md
@@ -1,0 +1,19 @@
+### Fixed: an empty `urdf_path`/`data_config` names that parameter instead of diagnosing the robot name
+
+`MuJoCoSimEngine.add_robot` read its model source by truthiness, so a source the caller
+supplied but left empty was indistinguishable from one they omitted. `urdf_path=""` took
+the absent branch, ran the deprecated name-as-registry-key fallback for a call that had
+asked for a file, and the refusal then diagnosed the name: it offered close-match
+suggestions for a name the caller never asked to resolve, and advised them to "pass
+data_config=<registered model> or urdf_path=<file>" - the kwarg they had just passed. The
+message was byte-identical to what supplying no source at all returns, so the report could
+not tell the two apart, while one whitespace character away (`urdf_path=" "`) the diagnosis
+was already correct ("File not found").
+
+A supplied-but-empty model source is now refused naming that parameter, the way
+`register_urdf` already refuses an empty `urdf_path` and the way the Newton
+(`urdf_path is not None`) and Isaac (`usd_path is None`) backends already read their own
+model source - MuJoCo's `add_robot` was the one site in the family reading it by
+truthiness. The guard sits after the existing name-taken check, so a caller who both
+reuses a label and empties the path is still told about the label, and whitespace stays a
+path rather than an empty value.
--- a/strands_robots/simulation/mujoco/simulation.py
+++ b/strands_robots/simulation/mujoco/simulation.py
@@ -1677,6 +1677,17 @@ class MuJoCoSimEngine(
         model source". The bare model-source message is kept only when no
         ``name`` was supplied at all.
 
+        A model source that is SUPPLIED but empty (``urdf_path=""`` /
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Narrow error-reporting fix in `MuJoCoSimEngine.add_robot`: a model source the caller supplied but left empty (`urdf_path=""` / `data_config=""`) is now refused by name instead of falling through the truthiness read into the deprecated name-as-registry-key fallback and returning a name diagnosis byte-identical to the omitted-source case. I verified the guard in the head tree: it uses the membership read (`supplied is not None and not supplied`), sits after the name-taken check so no settled precedence moves, and keeps whitespace as a path (`File not found`), matching `register_urdf`'s boundary. The dispatch layer (`_dispatch_action`) only forwards params the caller actually sent (`remapped = {k: v for k, v in d.items()...}`), so an omitted `urdf_path` on the tool surface never reaches the new guard - no regression for schema-constrained agents resolving from the registry. The `Robot()` factory passes `data_config or canonical` and forwards `urdf_path` verbatim, so the only behavior change is for a caller who explicitly passed the empty string, which is the mistake class the fix targets. The changed test file runs 18/18 green on the head (mujoco 3.8.0, `MUJOCO_GL=egl`).

## What's good

- The guard matches the family (Newton `is not None`, Isaac `is None`, `register_urdf`'s refusal wording) rather than inventing a new spelling, and the refusal names the parameter plus the omit-it remedy.
- Changelog fragment present (`changelog.d/2657-add-robot-empty-model-source.md`) per AGENTS.md step 3; no direct `CHANGELOG.md` edit.
- The AGENTS.md addition is written as a refinement of the existing selector rule (which read literally licenses this defect), with the `--scene-xml` counter-example that bounds it - so the rule does not over-reach onto sites where `""` genuinely is the sentinel.
- Tests pin both directions: the refusal, the distinguishability from omission, the signature-derived completeness set, and the controls (whitespace boundary, name-collision precedence, `name=""` short form, omitted-source name diagnosis).
- Error strings are plain ASCII; test class names describe behavior, not provenance.

PR #2658, opened 2026-08-23
fix(examples/libero): refuse a zero-frame rollout on both subcommands, not just isaac

cagataycali:
## Problem

`examples/libero/run.py` merged the two LIBERO drivers behind `mujoco` / `isaac`
subcommands sharing one eval-and-report loop. That loop prints the grep-stable
result line for whichever backend ran, and the line always carries a
`videos=<path>`. But the check that the rollout actually recorded something was
an optional per-backend hook: `_setup_isaac` supplied one, `_setup_mujoco`
supplied none, and `_EvalPlan` documented `check_recording` as *"Isaac's
zero-frame fail-loud check; `None` skips the check."*

The Isaac hook's own comment states the harm, and every clause of it is
backend-agnostic:

> `stop_cameras_recording` always returns status="success" (it's best-effort and
> idempotent), and imageio silently drops a 0-frame mp4 -- so without this check
> a blank rollout would ship a "success" line pointing at a file that never
> landed on disk.

So on the `mujoco` subcommand a rollout that recorded nothing printed a
`success_rate` line naming a file that does not exist, and exited 0.
`libero_backend_matrix.py` subprocess-runs this driver and parses those lines,
so it records such a run as a completed cell with a video the discovery glob
will never find.

## Measured

A real MuJoCo recording on this host, twice: once with a working offscreen
context and once with a GL context the host cannot create (a `MUJOCO_GL` naming
a windowed backend on a headless box - a value valid on every platform, so the
import succeeds and only the renders fail). Both payloads were then driven
through the driver's own shared `_evaluate_and_report` in the `mujoco` plan
shape.

| real recording | stop status | frames | render errors | file on disk | `upstream/main` | this branch |
| healthy (`MUJOCO_GL=egl`) | `success` | 39 | 0 | yes | `success_rate=0.40`, exit 0 | identical, exit 0 |
| no GL context | `success` | 0 | 44 | **no** | `success_rate=0.40`, exit 0 | refuses, exit 1 |

On `upstream/main` the two console outputs are identical apart from the run
timestamp in the `videos=` path - the operator cannot tell the two apart. The
recorder knew: its payload carries `render_refused` with the actionable remedy
(`Install EGL or OSMesa ...`), which the driver discarded.

## Change

One shared owner. `_require_recorded_frames(stop, camera)` is called
unconditionally by `_evaluate_and_report`, so the verdict holds for both
subcommands and cannot be opted out of by a plan that omits a hook. It also
reports the recorder's own `render_refused` reason, so the refusal names the
remedy instead of only the symptom.

`check_recording` keeps its remaining job - backend-specific recorder reporting,
which today is Isaac's `on_frame` retry counters - and still runs first, so those
diagnostics are on stdout before the refusal raises. Isaac's duplicate zero-frame
raise is removed: two owners of one verdict is how the two backends drifted apart
in the first place.

Scope: nothing else in the loop moves. An artifact missing for the named camera
counts as zero frames, because the result line names that camera's file either
way.

## Tests

`tests/test_examples_libero_drivers.py` is the file the formatter's docstring
already names as its pin, so the new cases extend it. Against `upstream/main`
with only the test file copied in: **6 failed / 16 passed**, of which 3 are
behavioural (`DID NOT RAISE` for a zero-frame rollout, on *both* backend plan
shapes), 2 report the new helper's absence, and 1 is the structural pin.

The two `frames > 0` cases pass on both trees - they are the controls that a
real recording is untouched and the backend hook still runs.

| break | fires | which |
| --- | --- | --- |
| control (no break) | 0 of 22 | - |
| `git checkout upstream/main -- examples/libero/run.py` | 6 | the full regression set |
| verdict guarded on `plan.check_recording is None` | 3 | the isaac shape, the ordering test, the unconditional pin |
| hook called after the verdict instead of before | 1 | the ordering test |
| `render_refused` dropped from the message | 1 | the reason test |
| artifact lookup ignores the camera name | 1 | the missing-artifact test |

The third row is the tempting minimal fix, and it still lets the isaac shape
through: pre-fix that backend's verdict came from its own hook, so guarding on
the hook's absence keeps the verdict dependent on a hook supplying it. Three
breaks fire exactly one test each.

## Verification

Measured at `ccddf76c` on Python 3.12 / mujoco 3.12.0 / lerobot 0.6.2.

- `tests/test_examples_libero_drivers.py`: 22 passed (14 pre-existing + 8 new).
- 166-file blast radius (every test reading `examples/`, walking a source or
  test tree, or naming the LIBERO surface, plus the repo-wide docs / changelog /
  docstring / source-string guards), identical selection on both trees with the
  changed test file excluded from each: branch `10 failed, 7882 passed,
  57 skipped`; `upstream/main` `11 failed, 7881 passed, 57 skipped`; both
  collected 7949.
- 8 of those failures are shared and are `tests/mesh/test_iot_*` on
  `ModuleNotFoundError: No module named 'awsiot'` - the `[mesh-iot]` extra
  (`awsiotsdk>=1.21.0,<2.0.0`), which CI installs. Every remaining difference is
  in `tests/mesh/test_zenoh_transport_security.py`, with a *different subset* per
  tree (2 on the branch, 3 on `upstream/main`): it opens real localhost Zenoh
  sessions, so two parallel suites contend for a fixed port. Run alone that file
  is `9 passed, 2 skipped` on both trees.
- `ruff check` + `ruff format --check` clean over `strands_robots tests
  tests_integ`; `mypy` over the same scope 24 == 24 against a pristine
  `upstream/main` worktree, branch-only empty, none in either changed file.

## Artifact

Both real recordings, and what the driver does with each. The frame strip is the
healthy recording decoded back out of the mp4 it wrote (39 frames, per-frame mean
|delta| 0.165-4.153, so it genuinely animates); it is byte-identical in both
trees and is the control.

zero-frame rollout refused on both subcommands


yinsong1986 (approved):
## Summary

Consolidates the zero-frame rollout verdict for `examples/libero/run.py` into the shared eval-and-report loop: `_require_recorded_frames(stop, camera)` now runs unconditionally in `_evaluate_and_report`, so a blank rollout is refused (exit 1) on both the `mujoco` and `isaac` subcommands instead of only where a per-backend `check_recording` hook happened to supply the check. The Isaac hook keeps its remaining job (reporting `on_frame` retry counters) and still runs before the shared verdict, so backend diagnostics precede the refusal; the refusal message additionally surfaces the recorder's own `render_refused` reason, which is where the actionable remedy lives. The diff matches the description, and the new helper's parsing matches the real `artifacts` payload shape both backends emit (including the missing-artifact-counts-as-zero case, since the result line names that camera's file either way).

## What's good

- Root cause addressed rather than symptom: one owner of the verdict, with an AST pin (`test_the_zero_frame_verdict_is_applied_unconditionally`) that fails if the call is ever re-guarded on `plan.check_recording is None`.
- Controls both ways: `frames > 0` cases pin that a healthy recording is untouched and the backend hook still runs; the missing-artifact and `render_refused` cases each get their own test.
- AGENTS.md compliance: changelog fragment shipped (`changelog.d/2658-libero-zero-frame-rollout-refused.md`), `Raises:` block on the new helper documents its one refusal, tests are behaviour-named, no host paths, no new non-ASCII in user-facing strings, and the change removes a silent-default path rather than adding one.
- 22/22 pass on `tests/test_examples_libero_drivers.py` at the head, reproduced locally.

yinsong1986 (approved):
## Summary

Moves the zero-frame rollout verdict for `examples/libero/run.py` out of the optional per-backend `check_recording` hook and into the shared eval-and-report loop: `_require_recorded_frames(stop, camera)` now runs unconditionally in `_evaluate_and_report`, so a rollout that recorded nothing is refused (exit 1) on both the `mujoco` and `isaac` subcommands instead of only where a backend happened to supply the check. Isaac's duplicate raise is removed and its hook keeps only backend-specific reporting (the `on_frame` retry counters), which still runs before the refusal so its diagnostics precede the raise. The refusal also surfaces the recorder's own `render_refused` reason, which is where the actionable remedy (EGL/OSMesa) lives.

## What's good

- The artifact fields the new helper reads (`camera` / `frames` / `errors` / `render_refused`) match exactly what `stop_cameras_recording` in `strands_robots/simulation/mujoco/rendering.py` emits, including the missing-artifact-counts-as-zero-frames case.
- Verified locally at `ccddf76c`: the extended test file is 22 passed on the branch, and swapping in only `upstream/main`'s `run.py` yields 6 failed / 16 passed - matching the PR description's break matrix, with the two `frames > 0` controls passing on both trees.
- `libero_backend_matrix.py` already handles a non-zero driver exit by recording the row as `skip (exit=N)`, so the new refusal degrades a matrix cell honestly instead of crashing the sweep.
- The AST pin against re-guarding the call on `plan.check_recording is None` closes the exact minimal-fix regression that would restore the per-backend split.
- Changelog fragment present per AGENTS.md step 3; the raise path is covered by the caller's `finally: sim.destroy()`; new user-facing strings are ASCII and the helper's `Raises:` block names its one refusal.

PR #2659, opened 2026-08-23
fix(isaac): read an OBJ keyword off the split, not as a "v " prefix

cagataycali:
## Problem

`_parse_obj` recognised an OBJ keyword by matching a `"v "` / `"f "` prefix. OBJ is a
whitespace-delimited format, so a keyword may be followed by a tab as legitimately as by a
space, and MuJoCo - the reference reader for every asset this module is pointed at - reads
both forms identically:

| the same tetrahedron | MuJoCo | `load_mesh_geometry` on `main` |
| --- | --- | --- |
| `v 0 0 0` (spaces) | 4 vertices, 4 faces | 4 vertices, 4 faces |
| `v\t0 0 0` (tabs) | 4 vertices, 4 faces | `has no triangle geometry (empty vertices/faces)` |

A skipped vertex line surfaced three ways, and none of them named the vertex:

| the file | outcome on `main` |
| --- | --- |
| every `v`/`f` tab-separated | `mesh <path> has no triangle geometry (empty vertices/faces)` - for a file declaring 4 vertices and 4 faces |
| tab vertices, space faces | `:5: face index 1 out of range (0 vertices declared)` - blames line 5's face; the vertices are on lines 1-4 |
| one writer emitting both | `:5: face index 3 out of range (2 vertices declared)` - 2 of 4 vertices silently dropped, so every later 1-based reference lands on the wrong vertex |
| a tab vertex no face references | **accepted**: 3 of 4 vertices, and `mesh_aabb` reports a z extent of `0.0` for a 9-unit-tall asset |

The last row raises nothing at all. `mesh_aabb` is what the MJCF scene loader turns into a
mesh-only body's collision proxy, so a silently dropped vertex is a silently wrong proxy under
`status="success"` - the eval-integrity failure mode `#2459` removed the default box proxy for.

The ASCII-STL parser in this same module has always read its keyword off the split fields, so
one module answered "is a tab a separator" two different ways.

## Change

`_parse_obj` reads the keyword off `line.split()` instead of matching a prefix. `vn` / `vt`
fall out naturally - they are different keywords, not the vertex keyword followed by a
separator - and an empty line is skipped explicitly.

## Scope

Only the OBJ keyword recognition moves. Every refusal keeps its exact wording and line number
(vertex arity, non-numeric component, face arity, non-integer index, out-of-range index,
`no triangle geometry`), negative face indices and `i/t/n` slash refs are untouched, and the
binary STL / ASCII STL / legacy MSH parsers are not touched at all.

**All 277 `.obj` / `.stl` / `.msh` assets in the real LIBERO tree parse byte-identically** -
same vertex and face counts, same geometry hash, same `mesh_aabb` centre and extent on both
trees (0 differing rows of 277).

## Verification

Measured at `70c0d1cd` against `upstream/main` `b717e91c`; the head has since moved to `3b1b7fc8`, which adds only the changelog fragment (`git diff 70c0d1cd..3b1b7fc8 -- '*.py'` is empty), so every number below still describes the tree under review.

Pre-fix split, new test class copied onto a pristine base: **3 failed / 18 passed**, all three
behavioural:

- `the same tetrahedron written with a tab after each v/f keyword was refused (mesh ... has no triangle geometry (empty vertices/faces)), and the refusal names the face rather than the vertex lines the parser did not recognise`
- `the tab-separated vertex was dropped: [(0.0, 0.0, 0.0), (1.0, 0.0, 0.0), (0.0, 1.0, 0.0)]`
- `OBJ ...:5: face index 3 out of range (2 vertices declared)`

The three remaining tests in the class pass on both trees: `vn`/`vt` are still not vertices,
blank and comment lines are still ignored, and the ASCII-STL sibling already tolerated a tab.

Break table - each break fires a distinct set, so no guard is decoration:

| break | fires |
| --- | --- |
| control (nothing broken) | 0 of 21 |
| `git checkout upstream/main -- mesh_assets.py` | 3 - the full regression set |
| widen the vertex rule to `keyword.startswith("v")` | 2 - the `vn`/`vt` control **and the pre-existing `test_obj_slash_syntax_and_quads`**, so the shipped suite itself refuses the over-reach |
| fix `v` but leave the face rule as the `"f "` prefix | 1 - the headline alone |
| remove the empty-line guard | 1 - the blank-line control, `IndexError: list index out of range` |

Gate: 149 files (every test naming the mesh-asset surface, all of `tests/simulation/isaac`, and
the repo-wide docstring / string / changelog guards), the changed test file excluded from both
trees so the run measures collateral damage only. **`6641 passed, 62 skipped` byte-identical on
both trees** (313.90s vs 314.29s), zero failures on either side. `tests/simulation/isaac/test_mesh_assets.py`
itself: 21 passed / 5 skipped (the 5 need `usd-core`).

`ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` 24 == 24
against a pristine `upstream/main` worktree with none in the changed files.

## Artifact

The real LIBERO `white_yellow_mug.obj` (4227 vertices), re-emitted with a tab after each `v`/`f`
keyword. Each thumbnail is a headless MuJoCo render (`MUJOCO_GL=egl`) of the geometry that tree
parsed, re-emitted verbatim as a canonical OBJ - so the picture is what `load_mesh_geometry`
returned, not what MuJoCo read off the file.

OBJ keyword whitespace

`main` refuses the tab form, so there is nothing to realize on the stage; this branch reads the
same 4227 vertices / 8454 faces and the same `0.1382 x 0.1025 x 0.1048 m` extent MuJoCo reports.
The space-separated control is identical on both trees and its render is byte-identical across
them (mean |delta| 0.000000).

### DIFF
--- a/changelog.d/2659-obj-keyword-whitespace.md
+++ b/changelog.d/2659-obj-keyword-whitespace.md
@@ -1,0 +1,29 @@
+### Fixed: a tab-separated OBJ asset parses as the space-separated one does
+
+`_parse_obj` recognised an OBJ keyword by matching a `"v "` / `"f "` prefix.
+OBJ is whitespace-delimited, so a keyword may be followed by a tab as
+legitimately as by a space, and MuJoCo - the reference reader for every
+asset this module is pointed at - reads both forms identically. Matching the
+prefix skipped a tab-separated vertex line, and the resulting failure never
+named the vertex: an all-tab tetrahedron was refused as `has no triangle
+geometry (empty vertices/faces)` despite declaring four vertices and four
+faces, tab vertices beside space faces were reported as `face index 1 out of
+range (0 vertices declared)` against the face on line 5, and one writer
+emitting both separators silently dropped half the vertices so that every
+later 1-based face reference landed on the wrong one.
+
+The quiet case is the one that reaches a scene. A tab-separated vertex that
+no face references was dropped with nothing raised, so `mesh_aabb` reported
+bounds the file never declared - a 9-unit-tall asset came back with a zero z
+extent - and `mesh_aabb` is what the MJCF scene loader turns into a
+mesh-only body's collision proxy. That is the eval-integrity failure mode
+the default box proxy was removed for, arriving under `status="success"`.
+
+The keyword is now read off `line.split()`. `vn` / `vt` fall out naturally,
+being different keywords rather than the vertex keyword followed by a
+separator, and the ASCII-STL parser in the same module had always read its
+keyword this way - so one module no longer answers "is a tab a separator"
+two different ways. Every refusal keeps its exact wording and line number,
+and all 277 `.obj` / `.stl` / `.msh` assets in the LIBERO tree parse
+byte-identically: same vertex and face counts, same geometry, same
+`mesh_aabb` centre and extent.
--- a/strands_robots/simulation/isaac/mesh_assets.py
+++ b/strands_robots/simulation/isaac/mesh_assets.py
@@ -118,23 +118,37 @@ def load_mesh_geometry(
 
 
 def _parse_obj(mesh_path: str) -> tuple[list[tuple[float, float, float]], list[int], list[int]]:
-    """Parse a Wavefront OBJ: ``v`` positions + ``f`` faces (n-gons kept)."""
+    """Parse a Wavefront OBJ: ``v`` positions + ``f`` faces (n-gons kept).
+
+    OBJ is whitespace-delimited, so a keyword may be followed by a tab as
+    legitimately as by a space (MuJoCo reads either form identically). The
+    keyword is therefore read off the split fields rather than matched as a
+    ``"v "`` prefix: matching the prefix skipped a tab-separated ``v`` line,
+    which either dropped the vertex silently - leaving a mesh whose bounds
+    are the ones the file did not declare - or made a later face reference
+    fall out of range and blamed the face for a vertex the parser never
+    recognised. ``vn`` / ``vt`` still fall out naturally: they are different
+    keywords, not the vertex keyword followed by a separator.
+    """
     points: list[tuple[float, float, float]] = []
     counts: list[int] = []
     indices: list[int] = []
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Narrow, well-verified parser fix: `_parse_obj` now recognises OBJ keywords by reading `parts[0]` off `line.split()` instead of matching a `"v "` / `"f "` prefix, so a tab after the keyword - legal in a whitespace-delimited format, and accepted by MuJoCo, the reference reader - no longer drops the line. This closes the quiet failure the description leads with: a tab-separated vertex no face references was dropped with nothing raised, leaving `mesh_aabb` (and the collision proxy the MJCF scene loader derives from it) reporting bounds the file never declared. The exact-match `== "v"` / `== "f"` comparison keeps `vn`/`vt` out by construction, the empty-line guard replaces the implicit skip the prefix match provided, and all existing refusal wordings and line numbers are preserved. One incidental behaviour change is strictly in the fail-loud direction: a bare `v` or `f` keyword line with no fields was previously skipped silently and now raises the existing arity refusal, consistent with the module's documented "nothing degrades to a silent default" contract.

## What's good

- Changelog fragment at `changelog.d/2659-obj-keyword-whitespace.md` per AGENTS.md step 3; no direct `CHANGELOG.md` edit.
- Test class named for the behaviour (`TestObjKeywordsAreWhitespaceDelimited`), not provenance, with controls pinning both directions - `vn`/`vt` are still not vertices, blank/comment lines still skip, and the ASCII-STL sibling's existing tolerance is pinned so the two parsers stay in agreement.
- Verified locally at the head SHA: the full mesh-assets file passes (26 passed), `ruff check` / `ruff format --check` clean on the changed files, no non-ASCII in any changed file, no host paths in tests.
- Scope discipline: only the private `_parse_obj` keyword recognition moves; STL/MSH parsers, refusal wordings, and the slash-ref/negative-index handling are untouched, and the byte-identical parse of the 277 real LIBERO assets bounds the real-world blast radius at zero.

PR #2660, opened 2026-08-23
fix(sim/isaac): a fromto collision geom carries its length and its offset

cagataycali:
## Problem

MJCF spells a capsule or a cylinder two ways. Either `pos` plus
`size="radius half-length"`, or `fromto="x1 y1 z1 x2 y2 z2"` plus
`size="radius"`, where the two endpoints carry both the placement and the
axis extent and the compiler re-derives `geom_size` from them.

`load_mjcf_scene_objects` read only the first spelling. `_geom_aabb` took
`pos` and `size`, so for a `fromto` geom `size` has a single component (the
radius), which fell through to the one-size `(r, r, r)` fallback, and `pos`
was absent so the box landed on the body origin.

Measured on one 0.60 m `fromto` capsule (radius 0.025, body origin at one end):

| measured | upstream/main | this branch | MuJoCo's own `geom_aabb` |
| --- | --- | --- | --- |
| reported size (m) | `(0.05, 0.05, 0.05)` | `(0.65, 0.05, 0.05)` | `(0.65, 0.05, 0.05)` |
| reported offset (m) | `(0.00, 0.00, 0.00)` | `(0.30, 0.00, 0.00)` | `(0.30, 0.00, 0.00)` |
| long axis covered | 7.7% | 100% | - |
| proxy volume | 125 cm3 | 1625 cm3 | object is 1625 cm3 |
| placement error | 0.30 m | 0.00 m | - |
| status | success | success | - |

Two things are wrong and they compound. The box is a ball of the geom's
radius rather than the bar, so it covers 7.7% of the object it is the physics
proxy for. And the offset is dropped, so the box sits 0.30 m from the object's
centre - the same `offset` the LIBERO pose applier adds when it places the prim
(`benchmarks/libero/adapter.py`, "``offset`` recorded by
`load_mjcf_scene_objects`"), so the placement inherits the error too.

The old fallback was answering for a shape that cannot exist: one `size`
component and no `fromto` is refused by MuJoCo with `size 1 must be positive
in geom`, so a confident `(r, r, r)` box was the one output no compilable
input could produce.

## Change

`_parse_fromto` reads the endpoint pair, and `_segment_aabb` returns the
`(center, half_extent)` of the segment - the midpoint, and the exact
half-extent `|d|/2 + r` along each axis for a capsule (`+ r` on the
cross-section only for a cylinder, whose flat caps do not extend past the end
discs). `_geom_aabb` consults `fromto` before `pos`, for `capsule` and
`cylinder`. The one-size-no-`fromto` fallback is dropped, because that spelling
does not compile.

This is the Isaac sibling of a rule the MuJoCo backend already owns:
`simulation.mujoco.scene_ops.fromto_fixed_size_components` documents that
"the axis extent comes from the endpoints" and that those components are
re-derived on every compile.

## Scope

`fromto` on a box or an ellipsoid additionally squares the cross-section and
is not covered here: those keep the existing `pos` + `size` reading, pinned by
a control test. A `fromto` whose endpoints coincide is refused by MuJoCo
("fromto points too close") and is left to the caller's fallback.

The sibling robot loader has the same gap and is deliberately not in this PR:
`_extract_mjcf_shape` also ignores `fromto`, so a 0.40 m `fromto` link reports
`shape_size=(0.03, 0.05)` - a 0.10 m stub. That is a different consumer (a
`ProceduralRobot` body shape rather than a scene object's collision AABB) with
different downstream semantics, so it wants its own change.

On reachability: `load_mjcf_scene_objects` is reached by `load_scene` on the
Isaac backend and by the LIBERO adapter. The 214 LIBERO scene files on this
machine carry no `fromto` geom, so the live surface is a caller-supplied or
third-party scene rather than the shipped tasks - `load_scene` accepts an
arbitrary MJCF.

## Tests

`tests/simulation/isaac/test_scene_object_fromto_geometry.py`, 27 cases.
Against `upstream/main`: **16 failed / 11 passed**. Every expected value is
computed from MuJoCo's own `geom_aabb` for the same geom rather than restated,
so the tests grade the loader against the compiler.

The pre-existing `test_capsule_extends_by_radius_ellipsoid_and_short_cylinder`
pinned the old fallback in as many words - "a single-size cylinder degrades to
a radius-cube" - for the spelling MuJoCo refuses. Its assertion now states
that there is no segment to measure, and the repaired test **fails against
pre-fix source**, so it is part of the regression evidence rather than a
relaxation.

Each guard was checked by breaking the thing it grades:

| break | failures | fires |
| --- | --- | --- |
| control | 0 of 27 | - |
| exact pre-fix (`git checkout upstream/main -- loaders.py`) | 16 | the full regression set |
| cylinder uses the capsule formula (caps extend by `r`) | 6 | the cylinder rows, the cap distinction, the diagonal |
| centre is `pos`, not the segment midpoint | 11 | every offset assertion, world position, the nested union |
| `_parse_fromto` accepts `>= 6` numbers | 1 | seven-numbers, alone |
| keep the one-`size` `(r, r, r)` fallback | 3 | the uncompilable-segment cases |
| route box / ellipsoid through the segment formula | 2 | the out-of-scope control, alone |
| drop the zero-length guard | 1 | zero-length, alone |

## Verification

Measured at `480882b4`, on mujoco 3.12.0. The head has since moved to
`d7df9bf6`, which adds only the changelog fragment (`git diff 480882b4..d7df9bf6 -- '*.py'` is empty).

- Blast radius: 394 test files (every test naming the changed symbols, all of
  `tests/simulation/isaac`, the LIBERO suites, and the repo-wide guards), the
  identical selection on this branch and on `upstream/main` with the new file
  excluded from both. **167 failing ids on each, `comm` empty in both
  directions.** All 167 are locally dep-absent (`reachy2_sdk`,
  `device_connect_edge`, `awsiot`, `qpsolvers` - each declared in a real extra
  CI installs). The one branch-only id was the pinning test above, reported
  separately: 32 passed here, 1 failed against pre-fix source.
- `mypy strands_robots tests tests_integ`: 24 == 24 against a pristine
  worktree, none in the changed files.
- `ruff check` / `ruff format --check` clean over the CI scope.
- New file: 27 passed on mujoco 3.12.0 and 27 on mujoco 3.5.0, the declared
  floor. 117 passed with random ordering enabled beside its sibling geometry
  and scene-object files.

## Artifact

The object is one 0.60 m `fromto` capsule on a table. Red is the collision
proxy `load_mjcf_scene_objects` hands the Isaac realization, drawn back into
the scene at the size and offset it reported; the tan bar is the object MuJoCo
simulates. Isaac Sim is not installed on this machine, so the artifact renders
what the loader hands the realization rather than the realized stage.

Control: the object itself is byte-identical in both columns (mean absolute
pixel difference 0.000000) - only the proxy moves.

fromto collision proxy

### DIFF
--- a/changelog.d/2660-isaac-scene-fromto-collision-geometry.md
+++ b/changelog.d/2660-isaac-scene-fromto-collision-geometry.md
@@ -1,0 +1,24 @@
+### Fixed: an Isaac scene object's `fromto` collision geom carries its length and its offset
+
+MJCF spells a capsule or cylinder two ways: `pos` plus
+`size="radius half-length"`, or `fromto` plus `size="radius"`, where the two
+endpoints carry both the placement and the axis extent. The Isaac scene-object
+loader read only the first spelling, so a `fromto` geom's single `size`
+component fell through to the one-size `(r, r, r)` fallback and the absent
+`pos` put the box on the body origin. A 0.60 m `fromto` capsule was reported as
+a 0.05 m cube at the origin - 7.7% of the object's long axis, 125 cm3 against
+1625 cm3, and 0.30 m from the object's centre - as `status="success"`. The
+offset is the same value the LIBERO pose applier adds when it places the prim,
+so the placement inherited the error too. `_geom_aabb` now consults `fromto`
+first for a capsule or cylinder, taking the segment midpoint as the centre and
+`|d|/2 + r` as the half-extent along each axis, with the `+ r` applied to the
+cross-section only for a cylinder, whose flat caps do not extend past the end
+discs. This matches what MuJoCo's own `geom_aabb` reports for the same geom,
+and it mirrors a rule the MuJoCo backend already owns in
+`scene_ops.fromto_fixed_size_components`.
+
+The one-`size`-no-`fromto` fallback is dropped rather than kept: that spelling
+is not a shape MuJoCo compiles (`size 1 must be positive in geom`), so the old
+`(r, r, r)` box was the one answer no compilable input could produce. `fromto`
+on a box or an ellipsoid additionally squares the cross-section and keeps the
+existing `pos` + `size` reading.
--- a/docs/simulation/isaac.md
+++ b/docs/simulation/isaac.md
@@ -220,7 +220,11 @@ LIBERO scenes get the same treatment for their *visuals*: `load_scene`
 renders each task object with its real mesh (bowls, plates - the assets a
 pixel-conditioned policy was trained on) while keeping the validated
 collision-AABB box as the invisible physics proxy, so switching backends does
-not also switch what the cameras see. An object whose mesh cannot be resolved
+not also switch what the cameras see. That box covers both MJCF spellings of a
+capsule or cylinder - `pos` plus `size="radius half-length"`, and `fromto` plus
+`size="radius"`, where the two endpoints carry the placement and the axis
+extent - so a `fromto` bar is proxied by its full length at its midpoint rather
+than by a ball of its radius at the body origin. An object whose mesh cannot be resolved
 keeps a visible box proxy, and the `load_scene` report then carries an
 explicit caveat that pixel-conditioned policy scores on that scene are not
 comparable across backends; when every object renders its mesh, the caveat
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -36,6 +36,7 @@ description file is configured. The loaders here layer on top.
 
 from __future__ import annotations
 
+import math
 import os
 import xml.etree.ElementTree as ET
 from dataclasses import dataclass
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Fixes the Isaac scene-object loader's collision AABB for the `fromto` spelling of a capsule / cylinder: `_geom_aabb` previously read only `pos` + `size`, so a `fromto` geom's single `size` component fell into the one-size `(r, r, r)` fallback and the absent `pos` placed the box at the body origin - a 0.60 m bar proxied by a radius-sized ball 0.30 m off-centre, reported as success. The new `_parse_fromto` / `_segment_aabb` pair takes the segment midpoint as the centre and the exact per-axis bound (`|d|/2 + r` for a capsule; the cross-section-only `+ r*sqrt(1 - u_i^2)` for a cylinder's flat caps), matching MuJoCo's own resolved geometry. The dropped one-size fallback only ever answered for a spelling MuJoCo refuses to compile (`size 1 must be positive in geom`), so no compilable scene loses behaviour, and `fromto` on box / ellipsoid deliberately keeps falling back rather than asserting an unrotated bound.

## What's good

- Independently reproduced the regression evidence: the new test file reports exactly 16 failed / 11 passed against the pre-fix `loaders.py` and 27 passed on the branch (verified here on mujoco 3.8.0, between the PR's declared 3.5.0 floor and 3.12.0 measurements).
- The oracle tests grade the loader against MuJoCo's own compiled `geom_aabb` / `geom_pos` / `geom_quat` rather than restated constants, and the two premises (one-size-no-fromto refused; `pos` + `fromto` together refused) are pinned against MuJoCo directly, so the dropped fallback's justification is itself under test.
- The out-of-scope boundary (box / ellipsoid `fromto`, zero-length segments, malformed `fromto`) is pinned by control tests, and the repaired pre-existing test fails on pre-fix source rather than merely relaxing.
- Changelog fragment present, ruff / format clean on the changed files, no host paths or non-ASCII in the new test or the tool-facing strings, and the `:func:` cross-reference to `scene_ops.fromto_fixed_size_components` resolves to a real symbol on one line.

PR #2662, opened 2026-08-23
fix(isaac): refuse an ASCII-STL facet that carries no endfacet

### LINKED ISSUE #2661: an ASCII-STL facet with no 'endfacet' is dropped silently, while the two binary parsers beside it refuse a truncated file
## Summary

`_parse_stl_ascii` flushes a facet only when it reads an `endfacet` line, and nothing tracks that a facet is open. A file whose **last** facet carries no `endfacet` therefore parses successfully, one triangle short, with that triangle's three vertices still in `points` - so `mesh_aabb` (and the USD `extent` the converter authors from the same list) reports bounds for geometry the returned mesh has no face for.

The two binary parsers in the same module both refuse a truncated file, because each reconciles its declared counts against the byte length. Only the ASCII reader accepts one, and it accepts it silently.

## Measured

At `21fa19a`, four ASCII STL files built from the same two triangles - a unit-square corner and a 9-unit-tall one - differing only in which `endfacet` lines are present:

| the file | outcome on `main` |
| --- | --- |
| both facets closed (control) | 4 vertices, 2 faces, extent `1.0 x 1.0 x 9.0` |
| the **last** `endfacet` missing | **accepted**: 4 vertices, **1** face, extent still `1.0 x 1.0 x 9.0` - the dropped triangle's vertices are in `points`, so the bounds span a triangle the caller did not get |
| the **only** facet unterminated | `mesh <path> has no triangle geometry (empty vertices/faces)` - for a file declaring one facet with three vertices |
| a **mid-file** facet unterminated | `:14: facet with 6 vertices (expected 3)` - blames line 14's `endfacet`; the unclosed facet is the one before it, and no facet in the file declares six vertices |

The siblings, for comparison, on the same geometry:

| the file | outcome on `main` |
| --- | --- |
| binary STL, truncated by one 50-byte record | `neither a well-formed binary STL nor an ASCII one (no 'solid' header)` |
| legacy MSH, truncated by 4 bytes | `declared counts need 64 bytes but the file holds 60 - truncated, or not a legacy MuJoCo binary mesh` |

## Why the first row is the one that costs something

Row 2 raises nothing. `mesh_aabb` is what the MJCF scene loader turns into a mesh-only body's collision proxy, and `convert_mesh_to_usd` writes the same all-points bounds into the USD `extent`, so an unterminated facet leaves a proxy describing geometry that is not in the mesh - under `status="success"`. That is the eval-integrity failure mode #2459 removed the default box proxy for, and #2659 fixed one instance of for OBJ vertices.

Rows 3 and 4 are loud but misdiagnosed: one reports an empty asset for a file that declares a facet, the other blames a line whose `endfacet` is the only correct thing about it, and reports an arity no line in the file has.

The module docstring states the contract this violates: "Nothing here degrades to a silent default."

## Fix shape

Track the open facet - the line its first vertex is on - and refuse when it is still open at the end of the file, and when a new `facet` keyword arrives before its `endfacet`. Naming that line fixes all three defective rows at once: row 2 stops being silent, and rows 3 and 4 stop naming the wrong thing. The `facet with N vertices (expected 3)` refusal stays as-is for a facet that really does declare the wrong number between its own `facet` and `endfacet` lines.

MuJoCo is not the reference reader here the way it is for OBJ: its mesh compiler reads binary STL, so the contract argument is this module's own fail-loud rule and the two binary siblings that already keep it, not a cross-reader comparison.

Measured with the parser imported directly (it is pure stdlib); no Isaac Sim or `usd-core` needed to reproduce.

cagataycali:
Closes #2661.

## Problem

`_parse_stl_ascii` flushes a facet when it reads an `endfacet` line and tracks nothing about one still being open. A file whose **last** facet is unterminated therefore parses successfully, one triangle short, with that triangle's three vertices still in `points` - and so in the bounds `mesh_aabb` reports and the `extent` `convert_mesh_to_usd` authors.

Four ASCII STL files built from the same two triangles - a unit-square corner and a 9-unit-tall one - differing only in which `endfacet` lines are present:

| the file | outcome on `main` |
| --- | --- |
| both facets closed (control) | 4 vertices, 2 faces, extent `1.0 x 1.0 x 9.0` |
| the **last** `endfacet` missing | **accepted**: 4 vertices, **1** face, extent still `1.0 x 1.0 x 9.0` |
| the **only** facet unterminated | `mesh <path> has no triangle geometry (empty vertices/faces)` - for a file declaring a facet with three vertices |
| a **mid-file** facet unterminated | `:14: facet with 6 vertices (expected 3)` - blames line 14's `endfacet`; the unclosed facet is the one before it, and no facet in the file declares six vertices |

Row 2 is the one that costs something: it raises nothing. `mesh_aabb` is what the MJCF scene loader turns into a mesh-only body's collision proxy, so the asset gets a proxy spanning a triangle its own mesh does not carry, under `status="success"` - the eval-integrity failure mode `#2459` removed the default box proxy for, and `#2659` fixed one instance of for OBJ vertices. Rows 3 and 4 are loud but misdiagnosed: one reports an empty asset for a file that declares a facet, the other reports an arity no line in the file has and blames the one line that is correct.

The two binary parsers in this same module both refuse a truncated file, because each reconciles its declared counts against the byte length:

| the file | outcome on `main` |
| --- | --- |
| binary STL, cut by one 50-byte record | `neither a well-formed binary STL nor an ASCII one (no 'solid' header)` |
| legacy MSH, cut by 4 bytes | `declared counts need 64 bytes but the file holds 60 - truncated, or not a legacy MuJoCo binary mesh` |

So one module answered "is a truncated asset an error" two different ways, and the module docstring states which one is the contract: "Nothing here degrades to a silent default."

## Change

The open facet is tracked by the line its first vertex is on, and refused at both ends - a following `facet` keyword, and the end of the file - through one shared wording (`_unterminated_facet_error`) so the two sites cannot drift. An ASCII STL declares no triangle count anywhere, so the open facet is the only thing there is to reconcile; that is what makes this the binary siblings' refusal rather than a new rule.

The regression test reports the mesh that parsed from an `else` branch rather than from inside `pytest.raises`. `pytest.raises.__exit__` suppresses, so putting the "it parsed instead" diagnosis at the end of the `with` body works at runtime - but a static reader cannot know a context manager swallows, so a `with` body ending in an unconditional `raise` makes the assertions after it unreachable, and CodeQL read them that way. The OBJ tab test this class's docstring points at already shapes it the other way round: `try` the call, put the diagnosis on the branch that means the parser got it wrong, assert afterwards. There the failure is a refusal so the diagnosis sits in `except`; here it is acceptance, so it sits in `else`. The diagnosis text and the wording the `match=` regex pinned are both unchanged - the latter is now an explicit assertion, so a different `ValueError` shows its whole message instead of a regex mismatch.

## Scope

Only the ASCII-STL facet state moves. `facet with N vertices (expected 3)` is unchanged for a facet that really does declare the wrong number between its own `facet` and `endfacet` lines; the vertex-arity and non-numeric-component refusals keep their exact wording and line numbers; vertex deduplication is untouched; and the binary STL, legacy MSH and OBJ parsers are not touched at all.

## Verification

Measured at `f2b8ed6` against `upstream/main` `671aae5`. `mesh_assets.py` is byte-identical at this branch's merge base and at current `upstream/main`, so the pre-fix numbers below are measured against the tree as it stands.

The new class copied onto a pristine base: **3 failed / 24 passed**, all three behavioural - the silent row plus the two misdiagnosed ones. The three controls in the class pass on both trees: a well-formed two-facet file still parses with a 9-unit z extent, a facet genuinely declaring four vertices still reports its arity, and the binary sibling already refuses a file cut mid-record.

Break table - each break fires a distinct set, so no guard is decoration:

| break | fires |
| --- | --- |
| control (nothing broken) | 0 of 32 |
| `git checkout upstream/main -- mesh_assets.py` | 3 - the full regression set |
| drop the end-of-file check only | 2 - the silent row and the empty-asset misdiagnosis |
| drop the following-`facet`-keyword branch only | 1 - the mid-file misdiagnosis |
| refuse a `facet` keyword whether or not one is open | 7 - including **two shipped tests** (`test_ascii_stl`, `test_the_ascii_stl_parser_already_tolerated_a_tab`), so the existing suite refuses the over-reach |
| name the current line instead of the open facet's first vertex | 3 - every refusal that pins which line identifies the facet |

Gate: 126 files (all of `tests/simulation/isaac`, plus every guard that walks the test tree or grades sources, docstrings, strings or changelog fragments - the ones a change to a test file can reach), the changed test file excluded from both trees so the run measures collateral damage only. **`4524 passed, 54 skipped` byte-identical on both trees, zero failures on either side** (286.36s vs 293.25s). `tests/simulation/isaac/test_mesh_assets.py` itself: 27 passed / 5 skipped (the 5 need `usd-core`); 32 collected either way, so the refactor moved no case.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1502 files). `mypy strands_robots tests tests_integ`: **24 errors on this branch and 24 on a pristine `upstream/main` worktree, with the branch-only set empty** and none in either changed file.

An AST sweep for the shape the alert names - a statement following a `with` whose body ends in `raise` or `return` - reports **0 sites** across `strands_robots` and `tests`, where before this round it reported exactly this one, tree-wide.

### DIFF
--- a/changelog.d/2662-ascii-stl-unterminated-facet.md
+++ b/changelog.d/2662-ascii-stl-unterminated-facet.md
@@ -1,0 +1,28 @@
+### Fixed: an ASCII-STL facet with no `endfacet` is refused instead of dropped
+
+`_parse_stl_ascii` flushed a facet when it read an `endfacet` line and kept no
+state about one still being open. A file whose last facet was unterminated
+therefore parsed successfully, one triangle short, with that triangle's three
+vertices still in the returned points - and so in the bounds `mesh_aabb`
+reports and the `extent` `convert_mesh_to_usd` authors. Nothing was raised.
+The MJCF scene loader turns those bounds into a mesh-only body's collision
+proxy, so such an asset got a proxy spanning geometry its own mesh does not
+carry: the eval-integrity failure mode the default box proxy was removed for,
+arriving under `status="success"`.
+
+Two other spellings of the same missing state were loud but misdiagnosed. A
+file whose only facet was unterminated was refused as `has no triangle
+geometry (empty vertices/faces)`, for a file that declares a facet with three
+vertices. A `facet` keyword arriving before the previous `endfacet`
+accumulated both facets' vertices into one, so the file was refused for
+`facet with 6 vertices (expected 3)` - an arity no facet in it declares -
+against the line whose `endfacet` was the only correct thing about it.
+
+The open facet is now tracked by the line its first vertex is on, and refused
+at both ends, the following `facet` keyword and the end of the file, through
+one shared wording. The binary STL and legacy MSH parsers in the same module
+already refuse a truncated file by reconciling their declared counts against
+the byte length; an ASCII STL declares no counts, so the open facet is the
+only thing there is to reconcile. `facet with N vertices (expected 3)` is
+unchanged for a facet that really does declare the wrong number between its
+own `facet` and `endfacet` lines, and no other parser is touched.
--- a/strands_robots/simulation/isaac/mesh_assets.py
+++ b/strands_robots/simulation/isaac/mesh_assets.py
@@ -209,7 +209,33 @@ def _parse_stl_binary(data: bytes, tri_count: int) -> tuple[list[tuple[float, fl
     return points, counts, indices
 
 
+def _unterminated_facet_error(mesh_path: str, lineno: int, vertex_count: int, before: str) -> ValueError:
+    """The refusal for an ASCII-STL facet still open when something else ends it.
+
+    One wording for both sites - a following ``facet`` keyword, and the end
+    of the file - so the two cannot drift, and each names the line the open
+    facet's first vertex is on, which is the only line that identifies it.
+    """
+    return ValueError(
+        f"STL {mesh_path}:{lineno}: facet left unterminated - {vertex_count} vertices read and no 'endfacet' "
+        f"before {before}, so the triangle they form would be dropped while their coordinates stay in the bounds"
+    )
+
+
 def _parse_stl_ascii(mesh_path: str, data: bytes) -> tuple[list[tuple[float, float, float]], list[int], list[int]]:
+    """Parse an ASCII STL, deduplicating shared vertices.
+
+    A facet is flushed on its ``endfacet``, so a facet still open when the
+    file ends - or when another ``facet`` keyword arrives - is refused rather
+    than left pending. Its vertices are already in ``points``, and therefore
+    in the bounds :func:`mesh_aabb` reports and the ``extent``
+    :func:`convert_mesh_to_usd` authors, while the triangle they form is in
+    no face: accepting one returns bounds for geometry the mesh does not
+    carry, under no error at all. The two binary parsers in this module
+    already refuse a truncated file by reconciling their declared counts
+    against the byte length; this is the same refusal for the format whose
+    triangle count is not declared anywhere but in its structure.
+    """
     text = data.decode("utf-8", errors="replace")
     if not text.lstrip().lower().startswith("solid"):
         raise ValueError(f"STL {mesh_path}: neither a well-formed binary STL nor an ASCII one (no 'solid' header)")
@@ -218,9 +244,11 @@ def _parse_stl_ascii(mesh_path: str, data: bytes) -> tuple[list[tuple[float, flo
     counts: list[int] = []
     indices: list[int] = []
     facet: list[int] = []
+    facet_lineno = 0
     for lineno, raw in enumerate(text.splitlines(), start=1):
         line = raw.strip()
-        if line.lower().startswith("vertex"):
+        lowered = line.lower()
+        if lowered.startswith("vertex"):
             parts = line.split()
             if len(parts) < 4:
                 raise ValueError(f"STL {mesh_path}:{lineno}: vertex with fewer than 3 components: {line!r}")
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes the last hole in `_parse_stl_ascii`'s truncation handling: a facet whose `endfacet` never arrives is now refused instead of being silently dropped from the face list while its vertices remain in `points` (and therefore in the `mesh_aabb` bounds and the USD `extent`). The open facet is tracked by the line of its first vertex and refused at both possible ends - a following `facet` keyword and end-of-file - through one shared `_unterminated_facet_error` wording, which also replaces the two previously misdiagnosed refusals (`no triangle geometry` for a single unterminated facet, and a phantom `6 vertices (expected 3)` blaming the wrong line for a mid-file one). This matches the module's stated contract ("Nothing here degrades to a silent default") and brings the ASCII path in line with the binary STL and legacy MSH parsers, which already reconcile declared counts against byte length. Scope is tight: the pre-existing arity refusal for a genuinely 4-vertex facet is pinned unchanged, and no other parser is touched.

## What's good

- Verified locally: the three behavioural regression tests fail with the test class copied onto pristine `main` (`3 failed, 29 passed`) and the full file passes at head (`32 passed`) - the tests measure the defect, not the implementation.
- The prior CodeQL unreachable-code finding (at `b99ed1a5`) is resolved at this head: the acceptance diagnosis now sits in the `else` branch of a `try`/`except`, matching the OBJ tab test's shape, so every assertion is statically reachable.
- Refusal messages name the open facet's first-vertex line (the only line that identifies it), are plain ASCII, and the shared-wording helper keeps the two refusal sites from drifting.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; guard suites (`Raises:` docstring completeness, ASCII tool strings, behaviour-named test classes) all pass at head.

Issue #2663, opened 2026-08-23
decision(sim): flat video keys are accepted for 1 of the 4 actions whose payload folds them


cagataycali:
## What

`MuJoCoSimEngine._dispatch_action` folds the flat video keys the schema publishes
(`output_path`, `fps`, `camera_name`) into the structured `video=` dict that four
actions take: `run_policy`, `start_policy`, `eval_policy`, `evaluate_benchmark`.

The fold cannot always honour a flat key. A bare `camera_name` has no path to
attach itself to, and a flat key sent alongside an explicit `video=` is not folded
at all. Whatever is left reaches the router's unknown-key check, and **only
`run_policy` accepts that residue** — so the same payload is accepted on
`run_policy` and refused on its three siblings.

## Measured

Driven through `_dispatch_action` on a live engine (a bogus `robot_name` /
`benchmark_name` so the method refuses fast and only the router's verdict shows):

| payload | refused by |
| --- | --- |
| `output_path` alone | 0/4 — the fold consumes it |
| `fps` alone | 0/4 — the fold consumes it |
| `camera_name` alone | **3/4** — `start_policy`, `eval_policy`, `evaluate_benchmark` |
| `video=` + `output_path` | **3/4** — same three |
| `video=` + `fps` | **3/4** — same three |

Refusal text on the three siblings, e.g.

```
Unknown parameter 'camera_name' for action 'eval_policy'. Valid: [...]
```

And what `run_policy` does with the same residue, measured end to end:

```
run_policy(video={"path": ".../v.mp4", "fps": 30}, output_path=".../o.mp4")
  -> status="success"; ".../v.mp4" written; ".../o.mp4" NOT written
```

So neither behaviour states what happened:

- `run_policy` **silently drops** the caller's `output_path` and writes to the
  `video` path instead. A caller who names both gets a rollout MP4 at a
  destination they did not ask for, under `status="success"`.
- the three siblings refuse with `Unknown parameter 'output_path'` — but
  `output_path` **is** a published property whose own schema description names
  `eval_policy` as a consumer, so the message denies a documented parameter.

## Why this is a decision, not a defect fix

The two candidate fixes both make something worse:

- **Widen** the acceptance to all four actions: propagates the silent
  wrong-destination drop to three more actions.
- **Narrow** it (drop the `run_policy` clause): makes `run_policy` report
  `Unknown parameter 'output_path'` for a parameter its own schema publishes.

The option that needs no precedence choice is a **third** one: have the fold
[...]

Issue #244, opened 2026-05-27
mesh(transport): pin STRANDS_MESH_BRIDGE_TOPICS_PREFIX reader + default-prefix tuple against the README env-var matrix


cagataycali:
## Context

Follow-up filed from PR #226 R4 review feedback.

The PR #226 README documents `STRANDS_MESH_BRIDGE_TOPICS_PREFIX` as an env var with a default of `response`, governing the prefix-vs-exact match split for the bridge-transport allowlist. On the `pr9/docs-mesh-readme-changelog` branch:

```
grep -rE 'STRANDS_MESH_BRIDGE_TOPICS_PREFIX|BRIDGE_TOPICS_PREFIX' --include='*.py' strands_robots/ tests/
# returns zero hits
```

The actual prefix-vs-exact match behaviour is implemented inline in `strands_robots/mesh/transport/bridge_transport.py:135-151` (`_should_bridge` walks `suffix_parts` from full path down to `/`-delimited prefixes), with no env-var hook on that branch.

Sibling to #241 (which pins `DEFAULT_BRIDGE_SUFFIXES`); this issue covers the env-var-reader side that #241 does not.

## Two possibilities to verify on PR #222 land

1. **Most likely**: the env-var reader is added in PR #222 (mesh/transport bridge dedup) but on a code path the current `_should_bridge` doesn't yet expose. In which case the PR #222 reviewer should verify the env-var name and default match the README row before merge.
2. **Less likely**: the env var was specced in the README but never implemented. In which case the README row is wrong and the prefix-vs-exact split needs either (a) an env-var hook added to `_should_bridge`, or (b) the README row deleted.

## Acceptance criteria

- [ ] After PR #222 lands, verify `grep -rE 'STRANDS_MESH_BRIDGE_TOPICS_PREFIX' --include='*.py' strands_robots/` returns at least one reader site.
- [ ] Add a unit test in the test module that lands alongside the reader (`tests/mesh/test_bridge_dedup.py` or successor) that:
  - Asserts the env-var name (`STRANDS_MESH_BRIDGE_TOPICS_PREFIX`) parses correctly.
  - Asserts the unset-default produces the documented `response` prefix-match behaviour.
  - Failure message points at the README env-var matrix row so a future edit gets a self-correcting hint.
- [ ] If possibility 2 turns out to be the case, file a docs follow-up to delete the row from the README `Mesh security configuration` section and the CHANGELOG `New env vars` table.

## Why a separate issue

[...]

PR #2664, opened 2026-08-23
test(sim): make the acceptance mirror track the payload-dependent video fold

cagataycali:
## What

`MuJoCoSimEngine._dispatch_action` folds the flat video keys the tool schema
publishes (`output_path`, `fps`, `camera_name`) into the structured `video=` dict
that four actions take: `run_policy`, `start_policy`, `eval_policy`,
`evaluate_benchmark`.

The fold cannot always honour a flat key. A bare `camera_name` has no path to
attach itself to, and a flat key sent alongside an explicit `video=` is not folded
at all. Whatever is left reaches the router's unknown-key check, and only
`run_policy` accepts that residue.

`tests/test_examples_dispatch_action_params.py` scans every example for a
`_dispatch_action` payload the router would refuse. Its own class docstring says
"the derived set must match what the live router actually does". It modelled the
three flat keys as accepted for all four folding actions:

```python
if action in _VIDEO_FOLDING_ACTIONS:
    accepted |= _VIDEO_FLAT_KEYS
```

So for three of those four actions the mirror blessed payloads the router refuses,
and its verdict table only ever probed the one action that agreed.

## Measured

Driven through `_dispatch_action` on a live engine, with each tree's own mirror
graded against its own router over 4 folding actions x 6 payload shapes:

| tree | mirror disagrees with the router |
| --- | --- |
| `upstream/main` | **9 of 24 cells** |
| this branch | 0 of 24 |

The 9 are `camera_name` alone, `video=` + `output_path`, and `video=` + `fps`, on
each of `start_policy`, `eval_policy` and `evaluate_benchmark`. Refusal text, e.g.

```
Unknown parameter 'camera_name' for action 'eval_policy'. Valid: [...]
```

The consequence is the defect the guard exists to prevent. An example written the
plausible way for an eval rollout:

```python
sim._dispatch_action("eval_policy", {"robot_name": "arm", "camera_name": "wrist"})
```

| grade the same example with | `upstream/main` guard | this branch's guard |
| --- | --- | --- |
| verdict | **11 passed** (reports clean) | FAILS, naming the line |
| message | (none) | `action 'eval_policy' does not accept 'camera_name'` |

and the router's own answer for that payload:

```
status=error  Unknown parameter 'camera_name' for action 'eval_policy'.
```

So an example that no-ops at runtime passed this guard. That is the dropped-payload
defect the file was written for (#244), reached through the guard's own blind spot.

## Change

Extract the fold into `MuJoCoSimEngine._fold_flat_video_keys`, the single owner of
"which flat video key is legitimate for which action", and have the mirror RUN it
on a probe payload rather than restate its rule:

```python
probe = dict.fromkeys(param_names, "probe")
MuJoCoSimEngine._fold_flat_video_keys(action, probe)
accepted |= {key for key in param_names if key not in probe}
if action == "run_policy":
    accepted |= set(MuJoCoSimEngine._RUN_POLICY_RESIDUAL_VIDEO_KEYS)
```

Acceptance is payload-dependent, so `_accepted_params` takes the payload's key set
and the scanner passes it. The residue `run_policy` alone accepts is named
(`_RUN_POLICY_RESIDUAL_VIDEO_KEYS`) instead of being buried in an
`action == "run_policy"` comparison, so the asymmetry is visible; #2663 asks
whether it should exist at all.

The extraction is behaviour-identical: the 4x6 accept/refuse matrix and
`run_policy`'s video-wins-over-`output_path` outcome are byte-for-byte unchanged
(the artifact asserts the router half of every cell is equal across the two trees).

Tests go 11 -> 48: a router-as-oracle test over 4 actions x 7 payload shapes that
requires the mirror to reach the router's verdict, the planted-example regression,
and a drift test pinning the table's restated action set to the engine's own owner.

## Break table

Each break fires a distinct set, so no guard is decoration.

| break | failures | fires |
| --- | --- | --- |
| control (no break) | 0 of 48 | - |
| mirror hand-lists the flat keys again (exact pre-fix) | 13 | oracle + verdict table + planted |
| mirror ignores the payload | 15 | oracle + verdict table + planted |
| fold drops the `camera_name` pairing rule | 4 | verdict table + planted only |
| `run_policy` stops accepting the residue | 3 | oracle only |
| engine narrows `_VIDEO_FOLDING_ACTIONS` | 5 | verdict table + drift only |
| the table's restated action set drifts | 1 | drift only |
| fold moved after `_validate_and_build_kwargs` | 12 | oracle only |

The fourth row is why the hand-written verdict table stays: when the source rule
changes, the mirror follows it, so the oracle still agrees and only the table
notices. The sixth row also collects 27 instead of 48, which is how a hand-listed
enumeration shrinks silently; the drift test is what reports it.

## Verification

Measured at `b8314ac`, `/tmp` worktrees, `MUJOCO_GL=egl`. The head has since
moved to `caab1d9`, which adds only the changelog fragment
(`git diff b8314ac..caab1d9 -- '*.py'` is empty), so every number below still
describes the tree under review.

- Pre-fix split (this file copied onto `upstream/main`): **3 failed / 8 passed**,
  each failure semantic and naming the offending cell.
- Gate: identical 310-file selection on both trees with this changed test file
  excluded from each, run in parallel. Byte-identical summary lines:
  `9 failed, 13031 passed, 120 skipped` (533.12s vs 537.67s). 9 failing ids on
  each, `comm` empty in both directions.
- The 9 shared: 8 in `tests/mesh/test_iot_connect_timeout_domain.py` on
  `No module named 'awsiot'` (the declared `[mesh-iot]` extra,
  `awsiotsdk>=1.21.0,<2.0.0`), and
  `tests/simulation/test_rendering_pacers_survive_a_clock_step.py`, which is
  **16 passed in isolation on the pristine base** (import-order dependent).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a third pristine
  worktree, none in the changed files. `ruff check` + `ruff format --check` clean
  over the CI scope.
- The changed file: 48 passed, and 379 passed / 51 skipped with random ordering
  enabled beside its 13 sibling `tests/test_examples*` files.

## Artifact

Each column is that tree's own mirror graded against that tree's own router, from
one capture script. The two dark consoles are the same planted example put through
each tree's guard.

mirror vs router

Control, asserted inside the figure script: the router's verdict is identical in
both columns for all 24 cells, so the extraction changed no behaviour, only what
the mirror models.

## Scope

`run_policy` accepting a residue its three siblings refuse is left exactly as it
is. Widening it propagates a silent wrong-destination drop
(`video=` + `output_path` writes the `video` path and drops `output_path` under
`status="success"`); narrowing it makes `run_policy` refuse a parameter its own
schema publishes. The third option, reporting what the fold could not honour, adds
a refusal where `run_policy` currently succeeds. Which is wanted is a maintainer
call, so it is filed as #2663 and named here rather than guessed.


yinsong1986 (approved):
## Summary

This PR fixes the example-scanner's acceptance mirror in `tests/test_examples_dispatch_action_params.py`, which restated the flat-video-key fold rule and modelled `output_path`/`fps`/`camera_name` as accepted for all four folding actions while the live router accepts the unfolded residue on `run_policy` alone. The fold is extracted into a single owner, `MuJoCoSimEngine._fold_flat_video_keys`, and the mirror now runs it on a probe payload instead of re-deriving the rule, making acceptance payload-dependent as the router's actually is. The extraction is behaviour-identical - I diffed it against the inline fold on `main` (same guard condition, same pops, same `flat.get("path")` checks, same call site before `_validate_and_build_kwargs`) - and all new names (`_VIDEO_FOLDING_ACTIONS`, `_RUN_POLICY_RESIDUAL_VIDEO_KEYS`, the classmethod) are private, so no public API, wire format, or persisted schema moves. The `run_policy` residue asymmetry is correctly left as-is and filed as #2663 rather than guessed at.

## What's good

- The mirror runs the router's own code rather than restating it, which is the drift-proof shape; the hand-written verdict table and the restated action set are each pinned to the engine's owner by dedicated tests, and the break table shows each guard fires a distinct failure set.
- `_router_refuses` answers the routing question on a bare engine with no world, GL context, or fixture - the unknown-key check runs before the method body, and the folding actions all refuse a `None` world cleanly, so the oracle test is cheap and hermetic.
- Verified locally at `caab1d90`: the changed test file reports 48 passed under `MUJOCO_GL=egl`.
- Changelog fragment present and correctly named; changed files are ASCII-clean; test names describe behaviour, not provenance; the new classmethod carries a full `Args:` block.

PR #2665, opened 2026-08-23
fix(isaac): read a fromto link's extent from its endpoints

cagataycali:
## What

MJCF spells a capsule or cylinder two ways:

- `pos` + `size="radius half-length"` - the extent is in `size`, along the geom's local z.
- `fromto="x1 y1 z1 x2 y2 z2"` + `size="radius"` - the endpoints carry both the placement and the axis extent, and `size` carries only the radius.

Both compile. `_extract_mjcf_shape` (the MJCF -> `ProceduralRobot` link reader behind `load_mjcf`) read only the first, so a `fromto` link's single `size` component fell to the radius-only branch and the half-length was padded to the module's 0.05 m default - however long the segment actually was.

This reads the endpoints when they are present, and routes both `fromto` consumers in the module through one `_segment_length` helper.

## Measured

A three-link arm plus one `pos`+`size` link as a control, through `load_mjcf`:

| link | spelling | reported half-length | true half-length | verdict |
|---|---|---|---|---|
| `upper` | `fromto="0 0 0 0 0 0.40"` `size="0.035"` | 0.05 m | 0.20 m | **4.0x too short** |
| `fore` | `fromto="0 0 0 0 0 0.28"` `size="0.030"` | 0.05 m | 0.14 m | **2.8x too short** |
| `shin` | `fromto="0 0 0 0 0 0.50"` `size="0.040"` | 0.05 m | 0.25 m | **5.0x too short** |
| `wrist_control` | `pos="0 0 0.09"` `size="0.025 0.09"` | 0.09 m | 0.09 m | matches (control) |

Nothing raises and nothing warns: `load_mjcf` returns a `ProceduralRobot` whose links are stubs, and that object *is* this function's product - it is what a caller builds an articulation from.

## Why this rule, and why here

The scene-object side of this same module already consults `_parse_fromto` before reading `size`, because a `fromto` geom's `size` does not carry the axis extent - `_geom_aabb` has done that since #2660. The MuJoCo backend states the same rule in `scene_ops.fromto_fixed_size_components`: *"the axis extent comes from the endpoints"*. One module was reading the two spellings two ways depending on which consumer asked.

Sharing `_segment_length` is what keeps that one answer. The helper deliberately keeps `_segment_aabb`'s exact `sum(d * d ...)` expression rather than the equivalent-looking `x ** 2`: the two disagree in the last ULP, so respelling it would move the scene-object AABBs it now shares with. Verified bit-identical over 8000 random segments (`x ** 2` differs on 2 of them), and a test pins the expression with that reason.

## Scope

- **`fromto` on a box or ellipsoid keeps its existing reading.** That case squares the cross-section and needs the rotated-box bound, which this module deliberately does not compute - the same exclusion `_geom_aabb` documents. Reading the segment there would publish an approximation nothing derived. Pinned as a control.
- **A degenerate or malformed `fromto` keeps the `size` reading.** MuJoCo refuses coincident endpoints outright, so there is no length to report, and a zero-length link would be worse than the default.
- **Reachability, stated honestly.** The `fromto` spelling is common in third-party MJCF - on this machine `mujoco_menagerie/shadow_dexee/shadow_dexee.xml` types 31 of 31 of its `fromto` geoms explicitly and `unitree_h1/h1.xml` 14 of 17. In those two files the *first* geom of each body is a visual mesh, so this function reports its default box for an unrelated reason (its first-geom/mesh rule) and those links are not reached by this change. The surface this fixes is an MJCF whose body's first geom is the typed `fromto` capsule - a hand-authored, procedurally generated, or robosuite-compiled arm, which is what `load_mjcf` exists to read.
- The scene-object AABBs are unchanged (bit-identical, above).

## Tests

`tests/simulation/isaac/test_description_loader_geometry.py`, extending the file that already covers this loader's shape fallbacks.

Pre-fix split against `upstream/main`: **7 failed / 36 passed** -> 43 passed. Six failures are behavioural; the seventh is the single-owner pin. The headline reads:

```
'thigh' spells a 0.40 m capsule with fromto and the loader reported
(capsule, (0.03, 0.05)) - the endpoints were not read, so the link is
published with the 0.05 m default half-length instead of 0.20 m
```

Break table (43 passed on the branch, so every row below is a regression the suite catches):

| break | failures | which |
|---|---|---|
| `git checkout upstream/main -- loaders.py` | 7 | the full regression set |
| drop the `length > 0.0` guard | **1** | the degenerate-`fromto` control alone |
| re-derive the length inside `_extract_mjcf_shape` | **1** | single-owner pin (`_extract_mjcf_shape` clause) |
| extend the `fromto` branch to box/ellipsoid | **1** | the documented-exclusion control alone |
| helper respelled `x ** 2` | **1** | single-owner pin (the ULP clause) |
| `_segment_aabb` re-derives its own length | **1** | single-owner pin (`_segment_aabb` clause) |

The three rows that fire the single-owner pin fire *different clauses* of it, and the four single-failure rows are what show each control is load-bearing rather than decorative.

## Verification

- 132-file blast radius (all of `tests/simulation/isaac`, everything naming the loader surface, and the repo-wide guards that walk source trees or grade docstrings), run with the changed test file excluded from both trees: **`4836 passed, 58 skipped` byte-identical on `upstream/main` and on this branch, zero failures on either side.**
- `ruff check` + `ruff format --check` clean over the CI scope; `mypy strands_robots tests tests_integ` **24 == 24** against a pristine `upstream/main` worktree, none in the changed files.
- The changed test file: 43 passed on mujoco **3.12.0** and on **3.5.0** (the declared floor).

## Artifact

Rendered headless on the MuJoCo backend (`MUJOCO_GL=egl`) from exactly the `(shape, shape_size)` each tree returned - orange is the shape the loader handed the caller, the grey bar beside each is the extent the MJCF declares. Same MJCF, same camera, same scene builder; only `strands_robots` differs.

fromto link extent: reported vs declared

The grey reference bars measure **866 px in both columns** (they come from the MJCF, not the loader) and the `pos`+`size` control link is byte-identical; the reported shapes go from 7691 to 19693 orange px.

Measured at `14f55012cdd8` (the source and tests commit). The head is now `fee551fc046a`, which adds only the changelog fragment and touches no `.py` file (`git diff 14f55012..fee551fc -- '*.py'` is empty).


yinsong1986 (approved):
## Summary

Fixes `_extract_mjcf_shape` (the MJCF -> `ProceduralRobot` link reader behind `load_mjcf`) so a capsule/cylinder spelled with `fromto` reports its half-length from the endpoints instead of falling to the radius-only branch and the 0.05 m default. The endpoint read is guarded (`_parse_fromto` returns `None` for anything that is not six parseable numbers, and a zero-length segment falls through to the existing `size` reading), so malformed and degenerate inputs keep their prior behaviour. Both `fromto` consumers in the module now route through one `_segment_length` helper that preserves the exact `sum(d * d ...)` expression, so the scene-object AABBs `_segment_aabb` produces are unchanged. `fromto` on a box/ellipsoid deliberately keeps the default-box reading, matching the exclusion `_geom_aabb` already documents.

## What's good

- Changelog fragment in `changelog.d/` per AGENTS.md step 3; no direct `CHANGELOG.md` edit.
- Only private helpers touched - no public API, wire-format, or schema surface changes.
- The regression tests pin both directions: the fix itself, the unchanged `pos`+`size` spelling, the degenerate/malformed fall-throughs, the box/ellipsoid exclusion, and the single-owner helper (including the ULP-sensitive expression, with the reason recorded).
- Test class/method names describe behaviour, not provenance; error message in the headline test names the wrong value and the expected one.
- Verified locally at the head SHA: the changed test file passes (42 passed, 1 environment-dependent skip) and `ruff check` is clean on both changed files.

PR #2666, opened 2026-08-23
fix(device-connect): read the state RPC's joints through the shared motor bus

cagataycali:
## Problem

`RobotDeviceDriver.getState` is the "what is this arm doing" RPC an operator or an agent calls at will. It reached the wrapped lerobot device directly:

```python
inner = getattr(self._robot, "robot", None)
obs = await asyncio.to_thread(inner.get_observation)
```

so it took no bus lock and read the **full** observation. The handler around that read logs at debug and returns what it has, so two independent failures both surfaced as a **successful RPC whose response simply has no `joints` key**:

- it barged in on a reader already holding the bus, and the feetech SDK answered `[TxRxResult] Port is in use!`;
- lerobot's `get_observation` sync-reads the motors **first** and only then loops the cameras, so one dead USB camera threw away the joint positions already in hand -- the eleven-hour incident `bus_access.read_joints` exists to prevent (its docstring: *"a single dead USB camera silently disarmed the whole arm"*).

Measured against a lerobot-shaped follower, same six positions on the wire in every row:

| scenario | upstream/main | this branch |
| --- | --- | --- |
| healthy arm (control) | 6 joints | 6 joints, byte-identical |
| one dead USB camera | **no `joints` key**, no error | 6 joints |
| another reader holds the bus | **no `joints` key**, device refused 1 transaction | 6 joints, waited 0.25 s for its turn |

## Change

Route the read through `bus_access.read_joints`, which owns both rules -- one lock per device, and joints read from the motor bus rather than from an observation a camera can spoil. `getState`'s frame filter is kept, because the function documents a `read_observation` fallback for a driver exposing no readable bus.

## The guard could not see the site

#2614 shipped `test_no_caller_drives_a_held_device_outside_bus_access`, a rule over the source rather than one call site. It grades an owned bus operation reached through a device the module holds, matched as `self.robot` or `self.robot.*` -- and this device arrives as `getattr(self._robot, "robot", None)` bound to a local, so **the shipped guard reports a clean tree**:

| tree | shipped guard | widened guard |
| --- | --- | --- |
| unfixed tree | **3 passed** (clean) | fails: `robot_driver.py:155 (inner.get_observation)` |
| fixed tree | 3 passed | 11 passed |

Its sibling half already follows exactly that binding form on `bus_access` itself (`bus = getattr(device, "bus", None)`); `_device_names_in` now follows it on the graded population too, so a device parked in a local is on the same bus whichever spelling reaches it. The `.robot` restriction stays load-bearing: relaxing it to any local binding reports `doctor.py:353 (sim.get_observation)`, a simulation engine that drives no serial port.

## Tests

`tests/test_device_connect_state_rpc_shares_the_bus.py`, reusing the fakes from the module that pins the camera incident so the RPC is graded against the same arm state. **4 failed / 4 passed** on `upstream/main`, every failure behavioural:

```
the RPC reported no joints for an arm whose motors answered: keys=[]
  (get_observation raised, and the joints were already in hand)
the device refused 1 overlapping transaction(s)
the wire was driven with the device lock free, so a concurrent reader can collide with this RPC
```

The four `TestNothingElseChanges` cases pass on both trees.

Breaks, each firing its own set:

| break | fires |
| --- | --- |
| control | 0 of 19 |
| exact pre-fix source, guard widened | 5 -- all four behavioural plus the guard naming the site |
| `read_observation` instead of `read_joints` (lock taken, full observation) | 3 -- the camera half only; the contention test passes |
| drop the frame filter | 1 -- the no-bus fallback control alone |
| guard stops following the `getattr` binding, source fixed | 0 -- invisible on a clean tree; pair with the row below |
| source reverted **and** guard narrow (what shipped) | 4 behavioural, and the guard reports **clean** |
| guard treats any local binding as a device | 1 -- it reports `doctor.py:353`, a sim engine |

## Verification

Measured at `1a2a15be`; the head has since moved to `c5618856`, which adds only the changelog fragment (`git diff 1a2a15be..c5618856 -- '*.py'` is empty). 129 files -- everything reaching a bus read, a device driver, or a tree-walking guard -- run with the identical selection on both trees, the two changed test files excluded from each:

| tree | result |
| --- | --- |
| this branch | `2 failed, 5930 passed, 60 skipped` in 317.11s |
| `upstream/main` | `2 failed, 5930 passed, 60 skipped` in 316.11s |

Failing-ID sets identical, `comm` empty in both directions. Both shared failures are `tests/test_device_connect_hardening.py` on absent `device-connect-agent-tools`, declared in `[device-connect]` and installed by `[all]`. `ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` is 24 errors on both trees with none in the changed files. The two changed test files: 19 passed.

## Sim artifact

The pose a Device Connect consumer reconstructs from `getState`, rendered headless in MuJoCo (`MUJOCO_GL=egl`) from the reported joints. The arm never moved -- what differs is whether its positions reached the response. The ground-truth render is identical across trees (mean level difference 0.000451) and the healthy row's joints are byte-identical, so the only difference is the two rows that had nothing to reconstruct.

state RPC reports joints through the shared bus

### DIFF
--- a/changelog.d/2666-state-rpc-shares-the-motor-bus.md
+++ b/changelog.d/2666-state-rpc-shares-the-motor-bus.md
@@ -1,0 +1,7 @@
+### Fixed: the Device Connect state RPC reads an arm's joints through the shared motor bus
+
+`RobotDeviceDriver.getState` reached the wrapped lerobot device directly -- `inner = getattr(self._robot, "robot", None)`, then `asyncio.to_thread(inner.get_observation)` -- so it took no bus lock and read the full observation. The handler around that read logs at debug and returns what it has, so two independent failures both surfaced as a successful RPC whose response simply has no `joints` key: barging in on a reader already holding the bus drew `[TxRxResult] Port is in use!`, and lerobot's `get_observation` sync-reads the motors first and only then loops the cameras, so one dead USB camera threw away the joint positions already in hand.
+
+It now routes through `bus_access.read_joints`, which owns both rules. Measured against a lerobot-shaped follower: a dead camera and a held bus each went from no joints at all to the arm's six positions, and the contended read waits for its turn instead of colliding. A healthy read is byte-identical.
+
+The structural guard from #2614 could not see the site: it matches a device held as `self.robot` or `self.robot.*`, and this one arrives as `getattr(self._robot, "robot", None)` bound to a local, so the shipped rule reported a clean tree. Its sibling half already follows exactly that binding form on `bus_access` itself, and the graded population now does too.
--- a/strands_robots/device_connect/robot_driver.py
+++ b/strands_robots/device_connect/robot_driver.py
@@ -18,6 +18,7 @@ from device_connect_edge.drivers import (
 )
 from device_connect_edge.types import DeviceIdentity, DeviceStatus
 
+from strands_robots.bus_access import read_joints
 from strands_robots.device_connect._authz import authz_error, is_authorized_caller
 from strands_robots.mesh.security import is_safe_policy_provider
 
@@ -140,6 +141,11 @@ class RobotDeviceDriver(DeviceDriver):
         """Get current robot state (joints, task info).
 
         Returns joint positions and task state if a task is running.
+
+        The joints are read through the shared motor-bus lock, so this RPC
+        waits its turn behind an in-flight rollout, teleop write or mesh probe
+        instead of colliding with it, and reports the joints even when a
+        camera on the same driver is failing.
         """
         result = {}
         task = getattr(self._robot, "_task_state", None)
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Routes `RobotDeviceDriver.getState`'s joint read through `bus_access.read_joints` instead of calling the wrapped device's `get_observation` directly, so the state RPC takes the per-device bus lock (no more `Port is in use!` collisions with a rollout/teleop/mesh probe) and reads joints from the motor bus rather than from a full observation a dead camera can spoil. The RPC's response shape is unchanged -- the frame filter is kept for the documented no-bus fallback where `read_joints` returns the full observation -- and the widened structural guard in `test_hardware_bus_is_shared_with_the_mesh.py` now follows a device parked in a local via `getattr(..., "robot", ...)`, closing the blind spot that let this site go ungraded.

## What's good

- Independently verified the behavioural claim: with the pre-fix `robot_driver.py` restored under the new tests, exactly the four behavioural cases fail plus the widened guard names `robot_driver.py` as the site; on the fixed tree all 19 pass, and the full device-connect selection (1311 tests) is green.
- The guard widening stays deliberately narrow (`.robot` bindings only), and the local-binding form mirrors the one `_owned_bus_operations` already uses on `bus_access` itself -- relaxing it would misfire on the sim engine in `doctor.py`, and the restriction keeps that out of scope by construction.
- Test fakes are reused from `tests/test_bus_read_joints.py` (the module pinning the original camera incident), so the RPC is graded against the same arm state; the `_Wrapper` docstring explains why a `MagicMock` cannot stand in.
- Conventions all satisfied: changelog fragment present, `ruff check`/`format` clean on the changed files, ASCII-only strings, no host paths, behaviour-describing test names, and the docstring xref roles resolve.

## Verification suggestions

- `git checkout <base> -- strands_robots/device_connect/robot_driver.py && pytest tests/test_device_connect_state_rpc_shares_the_bus.py tests/test_hardware_bus_is_shared_with_the_mesh.py -q` reproduces the 5 expected failures (4 behavioural + the guard naming the site); restore and re-run to confirm all green.

Issue #1961, opened 2026-08-05
Board coverage is measured over open items only - 18 of the last 30 merged PRs have neither a board item nor a closing link


cagataycali:
## Summary

The daily grooming record has reported "board coverage 100%, 0 orphaned PRs" for six consecutive days. That claim is **scoped to open items only**. Measured over merged PRs, coverage is not close to complete: **24 of the last 30 pull requests in this repo have no item on project board #2**, and **18 of those 24 also carry no closing issue reference** — so the work they delivered has no representation on the board at all, before or after merge.

This is a metric-definition defect plus a missing automation, not a backlog of missing clicks. Filing it because the same ask ("auto-add PRs on open") has been repeated in the grooming thread on 07-31, 08-01, 08-02, 08-03, 08-04 and 08-05 without a durable home.

## Evidence (2026-08-05, verified from the project side)

Board `PVT_kwDOD151Fs4BSRJP` fully paginated: **154 active items** (105 Done / 34 Backlog / 8 In review / 4 In progress / 3 Ready). Cross-referenced against the last 30 PRs:

| Class | Count | PRs |
|---|---:|---|
| On board | 6 | #1931, #1932, #1933, #1934, #1948, #1960 |
| Off board, **but** closing issue is on the board | 6 | #1930 (→#1506/#1516), #1938 (→#1937), #1944 (→#1942), #1946 (→#1942), #1953 (→#1952), #1954 (→#1949) |
| **Off board and no closing link — invisible work** | **18** | #1926, #1927, #1928, #1929, #1935, #1936, #1939, #1940, #1941, #1943, #1945, #1947, #1950, #1951, #1955, #1956, #1957, #1958 |

The 6 on-board PRs are precisely the ones that happened to be open at the moment a grooming pass ran. Coverage is therefore **a function of grooming timing**, not of a rule.

## Why this matters

1. **The `Done` column is not a countable delivery record.** It mixes issues and PRs inconsistently — some units of work appear once as an issue, some once as a PR, some twice (issue *and* PR), and 18 recent ones not at all. Any "items delivered" figure read off the board is wrong by an unknown margin. Velocity must keep coming from `search(...).issueCount`, never from board counts.
2. **"0 orphaned PRs" was never measured over the population it implies.** The honest statement is "0 orphaned *open* PRs".
[...]

PR #2667, opened 2026-08-23
fix(isaac): read an MJCF's bodies from the whole model, not just the top file

### LINKED ISSUE #2668: an MJCF's bodies are read from the top file only, so 75 of 227 registry models are refused and 12 more load with their arm missing
MuJoCo treats `<include file=...>` as a textual splice and merges every `<worldbody>` the spliced model carries, exactly as it treats `<compiler>` and `<asset>` as model-global. `tests/simulation/isaac/test_mjcf_mesh_assets_are_model_global.py` already states that rule for the mesh registry and the search directory.

Both loaders in `simulation/isaac/loaders.py` answer the bodies from `root.find("worldbody")` alone, so whether a body is visible depends on which **file** declared it. That fails in two directions, and they fail differently.

### 1. A model whose `<worldbody>` lives in a fragment reads as having none

`load_mjcf` refuses a model MuJoCo compiles. `aloha/scene.xml` is `<include file="aloha.xml"/>` plus a table:

| loaded file | MuJoCo compiles | `load_mjcf` today |
| --- | --- | --- |
| `aloha/aloha.xml` | nbody=21, njnt=16 | 21 bodies, 16 joints |
| `aloha/scene.xml` | nbody=21, njnt=16 | `ValueError: ... <worldbody> has no <body> children (phantom robot guard)` |

### 2. A model split across the top file and an include silently drops the rest

The bodies it cannot see, their subtrees and their joints are dropped under a **successful** load. `franka_emika_panda/mjx_single_cube.xml` keeps the cube and the mocap target locally and includes the arm:

| reported | MuJoCo | `load_mjcf` today |
| --- | --- | --- |
| bodies | 13 | **2** |
| joints | 10 | **0** |

Body and joint topology is what `load_mjcf` produces, so this is the worse of the two shapes: the whole Panda arm (`link0` .. `link7`, `hand`, both fingers) is absent from a `ProceduralRobot` the caller was told it received, and only `box` and `mocap_target` -- the two bodies the top file happens to declare locally -- come back. No error is raised.

### Measured scope

Over the 227 downloaded registry MJCFs: **75** refusals of models MuJoCo compiles, and **12** silent body/joint drops.

### Why this is a bug and not a limitation

The module already reads `<compiler>` and `<asset>` through a splice helper (`_mjcf_model_toplevel`), and a shipped test states that mesh assets are model-global. `<worldbody>` is the one model-global element still read per-file, so one module answers "which file may declare this" two different ways for facts MuJoCo treats identically.

Shape of the fix: read `<worldbody>` through the same splice the other two already use, preserving document order (MuJoCo assigns body indices in read order) and keeping "declares a `<worldbody>`" as a fact separate from "has bodies", since `load_mjcf` reports `has no <worldbody>` and the phantom-robot guard as different causes.

An `<include>` nested *inside* a `<worldbody>` is a separate question -- legal MJCF, spliced by MuJoCo, but used by 0 of the 227 shipped registry models -- and should stay out of scope until something needs it.

cagataycali:
Closes #2668.

## Problem

MuJoCo treats `<include file=...>` as a textual splice and MERGES every `<worldbody>` the spliced model carries, exactly as it treats `<compiler>` and `<asset>` as model-global. `tests/simulation/isaac/test_mjcf_mesh_assets_are_model_global.py` already states that rule for the mesh registry and the search directory.

Both loaders in `simulation/isaac/loaders.py` still answered the bodies from `root.find("worldbody")` alone, so whether a body was visible depended on which FILE declared it. That failed in two directions, and they failed differently.

**A model whose `<worldbody>` lives in a fragment read as having none**, so `load_mjcf` refused a model MuJoCo compiles. `aloha/scene.xml` is `<include file="aloha.xml"/>` plus a table:

| loaded file | MuJoCo compiles | `load_mjcf` on `upstream/main` |
| --- | --- | --- |
| `aloha/aloha.xml` | nbody=21, njnt=16 | 21 bodies, 16 joints |
| `aloha/scene.xml` | nbody=21, njnt=16 | `ValueError: ... <worldbody> has no <body> children (phantom robot guard)` |

**A model keeping some bodies in the top file and including the rest read only the ones it could see** and silently dropped the others, their subtrees and their joints - under a successful load. `franka_emika_panda/mjx_single_cube.xml` keeps the cube and the mocap target locally and includes the arm:

| reported | MuJoCo | `upstream/main` | this branch |
| --- | --- | --- | --- |
| bodies | 13 | **2** | 13 |
| joints | 10 | **0** | 9 |

The branch matches MuJoCo's bodies exactly. The one joint it does not report is the cube's `<freejoint>`, which the loader represents as a floating base rather than an articulation joint - that part is unchanged.

Body and joint topology is what `load_mjcf` produces, so this is the worse of the two shapes: on `upstream/main` the whole Panda arm (`link0` .. `link7`, `hand`, both fingers) is absent from a `ProceduralRobot` the caller was told it received, and only `box` and `mocap_target` - the two bodies the top file happens to declare locally - come back.

Measured over the 227 downloaded registry MJCFs: **75** refusals of models MuJoCo compiles, and **12** silent body/joint drops.

## Change

`_mjcf_model_worldbody_bodies` reads both facts through the module's existing `_mjcf_model_toplevel` splice, so `<worldbody>` is now answered the way `<compiler>` and `<asset>` already are. Both loaders call it.

Two properties the helper has to keep, each pinned:

- **Document order.** MuJoCo assigns body indices in the order it reads them, so a body from an `<include>` that PRECEDES the local `<worldbody>` comes first.
- **The "declares one" flag, separate from "has bodies".** `load_mjcf` reports `has no <worldbody>` and `<worldbody> has no <body> children (phantom robot guard)` as different causes; deriving the flag from the body list would collapse them.

An `<include>` nested INSIDE a `<worldbody>` is deliberately out of scope. It is legal MJCF and MuJoCo splices it, but no shipped registry `<worldbody>` uses one (0 of 227), and it is a question about splicing inside a `<worldbody>` rather than about which `<worldbody>` elements a model has. `TestAnIncludeInsideAWorldbodyIsOutOfScope` pins the boundary, including the premise that MuJoCo does read that body, so widening it later is a decision rather than an accident.

## Scope

| registry MJCFs (227 downloaded) | before | after |
| --- | --- | --- |
| refused a model MuJoCo compiles | 75 | 0 |
| loaded, silently missing bodies/joints | 12 | 0 |
| now report the articulation refusal their own included model file already reports | - | 3 |
| unchanged | 137 | 137 |

The 3 changed refusals (`unitree_g1`, `unitree_h1`, `unitree_h1_2` scene files) now report `has no joints (rigid body, not an articulation)` - which is what `load_mjcf` already reported for those robots' own model files, so the scene and the model it includes finally agree.

The 106 LIBERO asset scenes - the only in-package consumer of `load_mjcf_scene_objects`, and they use no `<include>` - return byte-identical object lists before and after.

## Tests

`tests/simulation/isaac/test_mjcf_bodies_are_model_global.py`, 31 cases. `TestMuJoCoAgreesOnTheModelsBodies` compares the loader's top-level body names against MuJoCo's own (the bodies whose parent is the world), so the expectations are derived from the compiler rather than restated, and every fixture is a model MuJoCo compiles.

Pre-fix split against `upstream/main`: **19 failed / 12 passed**. All 19 failures are in the two regression classes; the three control classes pass on both trees.

| break | fails | which |
| --- | --- | --- |
| control (no break) | 0 of 31 | - |
| `git checkout upstream/main -- loaders.py` | 19 | the full regression set |
| `load_mjcf` reads the top file only | 13 | the `load_mjcf` half + the guard-verdict test + the oracle |
| `load_mjcf_scene_objects` reads the top file only | **6** | the scene-loader half only |
| only the FIRST `<worldbody>` is read | 7 | the two-worldbody cases only |
| the flag derived from the bodies, not the `<worldbody>` elements | **2** | both phantom-guard tests, including the top-file-only control |
| document order not preserved | 3 | the order-sensitive cases only |
| the `<include>` cycle guard removed | **1** | the cycle test only (5.07s vs 0.29s: recursion) |

The two loader breaks are complementary (13 + 6 = 19), which is what shows each loader is independently pinned. The flag break firing a control that passes on both trees is what shows the flag keeps the two refusals distinguishable.

## Verification

Measured at `c3987ed6`, on `mujoco 3.12.0` (the version CI resolves under `mujoco>=3.5.0,<4.0.0`). The head has since moved to `235b6167`, which adds only the changelog fragment (`git diff c3987ed6..235b6167 -- '*.py'` is empty).

- Identical 187-file selection on both trees with the new test file excluded from each (`tests/simulation`, `tests/simulation/isaac`, everything naming the loaders, plus the repo-wide source/docstring guards): **`7630 passed, 66 skipped` byte-identical**, 330.70s vs 330.69s, zero failures on either side.
- New file: 31 passed on `mujoco 3.12.0` and on `3.5.0` (the declared floor).
- `ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ` 24 == 24 against a pristine `upstream/main` worktree, none in the changed files.

## Artifact

Each panel is what `load_mjcf` RETURNED, realized: one proxy box per reported `BodyDef`, at its reported pose, nested per its reported parent -> child edges. Same three registry assets in both columns; only `strands_robots` differs.

loader bodies before and after

Controls, asserted from the captured data inside the figure script: `aloha/aloha.xml` is identical in both columns (21 bodies, 16 joints, renders agreeing to 0.000000 mean level), and on this branch `aloha/scene.xml` renders BYTE-IDENTICALLY to `aloha/aloha.xml` - one robot, two spellings, one answer.

---

### Round 2 - description only

No code change: the diff and the head commit are untouched, so the approval stands on exactly the tree it was given for. Added the closing link to #2668, the defect issue this fixes, which was missing - without it this PR would have merged carrying neither a closing link nor a board item, which is the coverage gap #1961 tracks.

### DIFF
--- a/changelog.d/2667-isaac-mjcf-bodies-model-global.md
+++ b/changelog.d/2667-isaac-mjcf-bodies-model-global.md
@@ -1,0 +1,31 @@
+### Fixed: an MJCF's bodies are read from the whole model, not just the top file
+
+MuJoCo splices `<include file=...>` textually and merges every `<worldbody>` the
+spliced model carries, the same way it treats `<compiler>` and `<asset>` as
+model-global. Both loaders in `strands_robots.simulation.isaac.loaders` still
+answered that question from the top file's own `<worldbody>`, so whether a body
+was visible depended on which file declared it.
+
+That failed in two directions. A model whose `<worldbody>` lives entirely in a
+fragment read as having none, so `load_mjcf` refused a model MuJoCo compiles -
+`aloha/scene.xml` is an `<include>` plus a table, and it was rejected as a
+"phantom robot" while `aloha/aloha.xml` returned all 21 bodies. A model keeping
+some bodies locally and including the rest read only the ones it could see and
+silently dropped the others, their subtrees and their joints:
+`franka_emika_panda/mjx_single_cube.xml` returned 2 bodies and zero joints for a
+model with 13 and 10, so the whole Panda arm was absent from a
+`ProceduralRobot` the caller was told it received.
+
+Both loaders now read the bodies through the module's existing splice, in
+document order (which is the order MuJoCo assigns body indices in), and keep "no
+`<worldbody>` anywhere" distinguishable from "a `<worldbody>` with no bodies" so
+both refusals still report their own cause. Measured over the 227 downloaded
+registry MJCFs: 75 refusals of models MuJoCo compiles are fixed, 12 silent
+body/joint drops are fixed, and 3 scenes now report the articulation refusal
+their own included model file already reported. The 106 LIBERO asset scenes -
+the only in-package consumer of the scene loader, and they use no `<include>` -
+return byte-identical object lists.
+
+An `<include>` nested inside a `<worldbody>` stays out of scope: no shipped
+registry `<worldbody>` uses one, and it is a question about splicing inside a
+`<worldbody>` rather than about which `<worldbody>` elements a model has.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -403,8 +403,8 @@ def load_mjcf(path: str) -> ProceduralRobot:
 
     model_name = root.get("model", os.path.splitext(os.path.basename(path))[0])
 
-    worldbody = root.find("worldbody")
-    if worldbody is None:
+    declares_worldbody, top_bodies = _mjcf_model_worldbody_bodies(root, os.path.dirname(os.path.abspath(path)))
+    if not declares_worldbody:
         raise ValueError(f"MJCF loader: {path} has no <worldbody>")
 
     bodies: list[BodyDef] = []
@@ -490,7 +490,6 @@ def load_mjcf(path: str) -> ProceduralRobot:
         for child in body_el.findall("body"):
             _walk(child, body_idx)
 
-    top_bodies = list(worldbody.findall("body"))
     if not top_bodies:
         raise ValueError(f"MJCF loader: {path} <worldbody> has no <body> children (phantom robot guard)")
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Both Isaac MJCF loaders answered "what are the model's bodies" from `root.find("worldbody")` alone, while MuJoCo splices `<include file=...>` and merges every `<worldbody>` the spliced model carries. This PR routes that question through the module's existing `_mjcf_model_toplevel` splice via a new private helper, `_mjcf_model_worldbody_bodies`, which preserves document order (MuJoCo's body-index order) and keeps "no `<worldbody>` anywhere" distinguishable from "a `<worldbody>` with no bodies" so both refusals still report their own cause. That closes both failure directions on models MuJoCo compiles: a fragment-only `<worldbody>` no longer reads as a phantom robot (`aloha/scene.xml`), and a top file that keeps some bodies locally no longer silently drops the included ones with their subtrees and joints (`franka_emika_panda/mjx_single_cube.xml` returned 2 of 13 bodies and 0 of 10 joints under a successful load). No public surface moves: the helper is private, both loader signatures are unchanged, and the change only widens acceptance toward what MuJoCo itself compiles.

## What's good

- I reproduced the claims locally at this head: the new file passes 31/31, and with `loaders.py` checked out from `upstream/main` the split is exactly the claimed 19 failed / 12 passed - both regression classes fail, all three control classes pass.
- The rest of `tests/simulation/isaac/` is delta-zero on this branch (the 3 failures in my environment reproduce identically on a clean `upstream/main` worktree, so they are environmental, not this PR).
- `TestMuJoCoAgreesOnTheModelsBodies` derives expectations from the compiler (`body_parentid == 0`) rather than restating them, and `test_mujoco_does_splice_it` pins the premise of the deliberately out-of-scope worldbody-nested `<include>`, so widening that boundary later is a decision rather than an accident.
- The unusable-include semantics (missing / malformed / directory / cycle) are inherited from the already-reviewed `_mjcf_model_toplevel` and re-pinned here through the new call path rather than assumed.
- Changelog fragment follows `changelog.d/<pr>-<slug>.md`; docstrings carry full `Args:`/`Returns:`; ASCII-only throughout; test names describe behaviour, not provenance.

PR #2669, opened 2026-08-23
fix(isaac): read a geom's shape from the <default> class it inherits

cagataycali:
## Problem

MJCF lets a `<geom>` spell almost nothing itself. `<default class="X">` supplies every attribute the element omits, `<body childclass="X">` names that class for a whole subtree, and the classes form an inheritance tree rooted at the unnamed top-level `<default>`. So `type`, `size` and `fromto` - the three attributes that decide what shape a geom is - need not be on the geom at all.

Both MJCF readers in `simulation/isaac/loaders.py` asked the element directly, so they saw only the half a geom spells itself. `asimov_v0`, a shipped registry robot, declares its leg links this way:

```xml
<default class="body_capsule"><geom type="capsule" size="0.05"/></default>
...
<geom class="body_capsule" fromto="0 0 0 0 0 -0.25"/>
```

Read as the geom's own attributes alone, that geom has no `type` and a one-component `size`, so `_extract_mjcf_shape` fell through to its `box (0.05, 0.05, 0.05)` fallback: a 10 cm cube for a 25 cm shin, on six links, under `load_mjcf` reporting success. It also made the endpoint reading that same function performs unreachable for exactly those links, because that branch is only taken once the geom is known to be a capsule or a cylinder.

The `abh_*` hands are the extreme form. Every geom there is `<geom class="X"/>` - attribute-free apart from the class - so `_find_body_mesh` saw a four-geom body declaring no mesh at all, and nine finger segments of 7 mm radius reported as 10 cm cubes.

| link | MuJoCo's compiled geom | reported before |
| --- | --- | --- |
| `left_knee_link` | capsule, half-length 0.125 m | box, 0.05 m half-extent |
| `left_hip_pitch_link` | capsule, half-length 0.035355 m | box, 0.05 m half-extent |
| `left_hip_yaw_link` | capsule, half-length 0.02 m | box, 0.05 m half-extent |

## Change

`_mjcf_geom_defaults` flattens the class tree, and `_geom_attrs` answers "what did this geom declare" once, for every reader: the geom's own `class`, else the nearest enclosing `childclass`, else the root class, overridden by whatever the element spells itself. All twelve geom attribute reads in the module now go through it, so no reader is left asking the element.

Three properties are inherited from MJCF rather than invented:

- **The geom wins.** An attribute the element spells itself overrides the class, so resolving a class never displaces what a geom says.
- **The classes are model-global.** `<default>` is a top-level element, so a spliced `<include>` fragment declares classes for the whole model - the same splice `<compiler>` and `<asset>` already get here.
- **One root.** MuJoCo refuses a second top-level `<default>` ("top-level default class 'main' cannot be renamed"), so the tree is one recursion from one root.

An undeclared class contributes nothing rather than failing the load: MuJoCo refuses such a model itself and names the offending class, so that report is its to make.

## Blast radius

Measured over 660 MJCFs (the downloaded registry assets plus every LIBERO scene), with MuJoCo's own compiled geoms as the oracle:

| reader | files changed | what changed |
| --- | --- | --- |
| `load_mjcf` | 12 | all box -> capsule/cylinder corrections; no verdict changes |
| `load_mjcf_scene_objects` | 66 | 58 collision-proxy corrections, 8 that now refuse |
| LIBERO subset (109 files) | 0 | unchanged on both readers |

The 8 that now refuse are all `ability_hand_api` models, and **MuJoCo itself refuses all 8** - their `assets/` directory is absent (`Error opening file 'assets/FB_palm_ref.STL'`). Resolving the class-declared mesh name is what lets the fail-loud contract from #2459 see a missing asset there instead of silently emitting a box proxy, so the loader now agrees with MuJoCo that those files are broken. That is a genuine verdict change and the only one in the sweep.

## Visual

default-class geometry

Each panel renders the proxy geometry `load_mjcf` **returned** for the same three `asimov_v0` links, at the reported positions, headless with `MUJOCO_GL=egl`. The right panel is built from `model.geom_size` on the compiled model.

This branch's render is **byte-identical** to MuJoCo's (`np.array_equal` is `True`, mean absolute level difference `0.000000`). `upstream/main` is 13.62 mean levels away, with 15190 proxy pixels against 6597 - 2.3x the area.

## Tests

`tests/simulation/isaac/test_mjcf_default_class_geometry.py`, 13 cases, every expected shape read back off a compiled `MjModel` rather than restated. Pre-fix split against `upstream/main`: **11 failed / 2 passed**, all 11 in the regression and rule classes and none in the controls.

| break | fires |
| --- | --- |
| `git checkout upstream/main -- loaders.py` | 11, the full regression set |
| `childclass` ignored | 2, both `childclass` cases |
| a nested class does not inherit its parent | 1 |
| `<default>` read from the top file only | 1, the `<include>` case |
| the class wins over the geom's own attributes | 1, the precedence control alone |
| only the shape readers resolve; mesh and group left raw | 3, the mesh case and both rule tests |
| the AST scan reaches nothing | 1, the rule's own non-vacuity clause |
| control | 0 of 13 |

The two controls hold on `upstream/main` too - they pin that resolving a class does not displace what the geom spells itself, and that a model with no `<default>` at all is read exactly as before.

While writing the break table, the break aimed at the root class did not fire, which showed the pre-seed it broke was dead code: with one top-level `<default>` guaranteed, the recursion already populates the root class. It is removed, and the corpus sweep is byte-identical without it.

## Verification

Measured at `6b634f6`, mujoco 3.12.0. The head has since moved to `1863b46`, which adds only the changelog fragment (`git diff 6b634f6..1863b46 -- '*.py'` is empty).

| tree | result |
| --- | --- |
| this branch | `1 failed, 3106 passed, 64 skipped` |
| `upstream/main` | `1 failed, 3106 passed, 64 skipped` |

90-file selection (everything naming the loaders, the scene-object and robot readers, the LIBERO adapter, plus the repo-wide guards), identical on both trees with the new test file excluded from each. Failing-ID sets identical, branch-only and base-only both empty. The one shared failure is `test_declared_qp_backends_are_real_qpsolvers_extras`: `qpsolvers` is absent from this venv and is declared in `[sim-mujoco]` and `[cosmos3-sim]`, so CI installs it.

The new file: 13 passed on mujoco 3.12.0 and on 3.5.0, the declared floor. `ruff check` and `ruff format --check` clean; `mypy strands_robots tests tests_integ` 24 errors on both trees, branch-only empty.

## Rebased onto a moved base

`main` gained #2667 in the same file, so the head is now `e79093d`. Both new functions land at the same insertion point and both are kept. #2667 replaced the single `<worldbody>` element with a flattened body list, so the `childclass` read becomes per-body - a spliced model may carry several `<worldbody>` elements and they need not name the same class, which the previous single-element read could not express.

Re-measured on the merged tree: the pre-fix split against the new base is unchanged at **11 failed / 2 passed**, all 11 in the regression and rule classes; `tests/simulation/isaac/` is **1059 passed, 35 skipped**, covering this change, #2667's and #2662's together; `ruff check` and `ruff format --check` clean.

### DIFF
--- a/changelog.d/2669-isaac-mjcf-default-class-geometry.md
+++ b/changelog.d/2669-isaac-mjcf-default-class-geometry.md
@@ -1,0 +1,9 @@
+### Fixed: a geom's shape is read from the `<default>` class it inherits
+
+MJCF lets a `<geom>` spell almost nothing itself. `<default class="X">` supplies every attribute the element omits, `<body childclass="X">` names that class for a whole subtree, and the classes form an inheritance tree rooted at the unnamed top-level `<default>`. So `type`, `size` and `fromto` - the three attributes that decide what shape a geom is - need not be on the geom at all.
+
+Both MJCF readers in the Isaac loaders asked the element directly, so they saw only the half a geom spells itself. `asimov_v0`'s six leg links reported the 0.05 m fallback box for capsules whose `<default class="body_capsule">` declares `type="capsule" size="0.05"` over a 0.25 m `fromto` segment: a 10 cm cube for a 25 cm shin, on a shipped registry robot, under `load_mjcf` reporting success. The `abh_*` hands are the extreme form, where every geom is `<geom class="X"/>` and the class carries the mesh name, so the reader saw a four-geom body declaring no mesh at all.
+
+One rule now answers what a geom declared, for every reader: the geom's own class, else the nearest enclosing `childclass`, else the root class, overridden by whatever the element spells itself. Classes are model-global through `<include>`, the same splice `<compiler>` and `<asset>` already get.
+
+Measured over 660 MJCFs with MuJoCo's own compiled geoms as the oracle: 12 files change on the robot reader, all box to capsule or cylinder corrections; 66 change on the scene-object reader, 58 of them collision-proxy corrections and 8 that now refuse. Those 8 are `ability_hand_api` models MuJoCo itself refuses, where resolving the class-declared mesh name lets the fail-loud contract see a missing asset instead of silently emitting a box proxy. LIBERO is unchanged on both readers.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -407,6 +407,8 @@ def load_mjcf(path: str) -> ProceduralRobot:
     if worldbody is None:
         raise ValueError(f"MJCF loader: {path} has no <worldbody>")
 
+    geom_defaults = _mjcf_geom_defaults(root, os.path.dirname(os.path.abspath(path)))
+
     bodies: list[BodyDef] = []
     joints: list[JointDef] = []
 
@@ -422,7 +424,11 @@ def load_mjcf(path: str) -> ProceduralRobot:
         )
     )
 
-    def _walk(body_el: ET.Element, parent_idx: int) -> None:
+    def _walk(body_el: ET.Element, parent_idx: int, childclass: str) -> None:
+        # ``childclass`` is MJCF's per-subtree default class: it applies to every
+        # descendant that does not name a class of its own, and a nested body
+        # overrides it for its own subtree.
+        childclass = body_el.get("childclass") or childclass
         body_name = body_el.get("name") or f"body_{len(bodies)}"
         position = _parse_xyz(body_el.get("pos"))
 
@@ -434,7 +440,7 @@ def load_mjcf(path: str) -> ProceduralRobot:
             mass = _safe_float(inertial.get("mass"), 1.0)
 
         # Geometry - first <geom> primitive type.
-        shape, shape_size = _extract_mjcf_shape(body_el)
+        shape, shape_size = _extract_mjcf_shape(body_el, geom_defaults, childclass)
 
         bodies.append(
             BodyDef(
@@ -488,21 +494,26 @@ def load_mjcf(path: str) -> ProceduralRobot:
             )
 
         for child in body_el.findall("body"):
-            _walk(child, body_idx)
+            _walk(child, body_idx, childclass)
 
     top_bodies = list(worldbody.findall("body"))
     if not top_bodies:
         raise ValueError(f"MJCF loader: {path} <worldbody> has no <body> children (phantom robot guard)")
 
+    world_childclass = worldbody.get("childclass") or ""
     for body_el in top_bodies:
-        _walk(body_el, parent_idx=0)
+        _walk(body_el, parent_idx=0, childclass=world_childclass)
 
     robot = ProceduralRobot(name=model_name, bodies=bodies, joints=joints)
     _validate_kinematic_tree(robot)
     return robot
 
 
-def _extract_mjcf_shape(body_el: ET.Element) -> tuple[str, tuple[float, ...]]:
+def _extract_mjcf_shape(
+    body_el: ET.Element,
+    defaults: dict[str, dict[str, str]],
+    childclass: str,
+) -> tuple[str, tuple[float, ...]]:
     """Best-effort MJCF body -> (shape, shape_size) extraction from first <geom>.
 
     A capsule or cylinder may spell its axis extent with ``fromto`` instead of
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Both MJCF readers in `simulation/isaac/loaders.py` asked the `<geom>` element directly for `type`/`size`/`fromto`/`mesh`/`group`, so attributes supplied by `<default>` class inheritance (geom `class`, enclosing `childclass`, or the unnamed root class) were invisible - shipping registry robots like `asimov_v0` reported 0.05 m fallback boxes for 0.25 m capsules. This PR adds `_mjcf_geom_defaults` (flattens the class tree, model-global through the same `<include>` splice `_mjcf_model_toplevel` already resolves for `<compiler>`/`<asset>`) and `_geom_attrs` (class attrs overridden by the element's own), and routes all twelve geom attribute reads in the module through it. All touched helpers are private; `load_mjcf` and `load_mjcf_scene_objects` keep their public signatures, so there is no API surface change. The one verdict change (8 `ability_hand_api` scene files now refuse) matches MuJoCo's own refusal of those files and the fail-loud contract from #2459, and is documented rather than silent.

The head at `e79093d1` also carries a real merge resolution with #2667 (the model-global worldbody read that landed on `main`): the single-worldbody `childclass` read becomes a per-body map keyed by element identity. I reviewed the combined diff: the `id()`-keyed map cannot match bodies from re-parsed `<include>` fragments (a second `_mjcf_model_toplevel` call yields fresh Element objects), but the fallback is `""` and that is the only value a compilable model can have - MuJoCo refuses any attribute on `<worldbody>` outright ("World body cannot have attributes"), so the miss is inert and the read is defensive-only, exactly as it was pre-merge.

## What's good

- Independently reproduced: the 13 new tests pass on mujoco 3.8.0 in a fresh checkout (the PR reports 3.5.0 and 3.12.0, so this adds a third point in the declared range), and the full `tests/simulation/isaac/` suite shows no branch-caused failures at either head - the 3 failures in my environment reproduce identically on the merge base. The post-merge composition with #2667's `test_mjcf_bodies_are_model_global.py` also passes (44/44 across both files).
- MuJoCo as oracle: every expected shape is read back off a compiled `MjModel`, with `premise:` asserts separating fixture rot from loader regressions, and enum comparisons are value-based per AGENTS.md > "MuJoCo enums are matched by value".
- The AST rule test (`TestOneRuleAnswersWhatAGeomDeclared`) holds future readers to the resolver and pins its own non-vacuity, so a fifth geom reader added later is graded on arrival.
- Controls pin both directions: the geom's own attribute still wins over its class, and a model with no `<default>` is read byte-identically to before.
- Changelog fragment present (`changelog.d/2669-...md`), no non-ASCII in any touched file, no host paths in tests.

One non-blocking observation, for a possible follow-up rather than this PR: MuJoCo's implicit root class is addressable by name as `class="main"`, and the resolver stores the root only under `""`, so `<geom class="main"/>` still falls to the fallback box. That is byte-identical to the pre-PR behaviour (not a regression), the reader is documented best-effort, and no registry asset appears to spell it - noting it here only so the class of gap is on record.

## Verification suggestions

- `pytest tests/simulation/isaac/test_mjcf_default_class_geometry.py tests/simulation/isaac/test_mjcf_bodies_are_model_global.py -q` on the oldest mujoco you have handy - the composition with #2667 is the part CI ran for the first time on this head.
- Spot-check the headline case against a downloaded `asimov_v0`: `load_mjcf` on its MJCF should now report capsules for the six leg links, with `(radius, half-length)` matching `model.geom_size` off the compiled model.

yinsong1986 (approved):
## Summary

Both MJCF readers in `simulation/isaac/loaders.py` previously read `type`/`size`/`fromto`/`mesh`/`group`/`pos`/`quat` off the `<geom>` element directly, so anything supplied by `<default>` class inheritance was invisible - shipped registry robots (`asimov_v0`, the `abh_*` hands) reported fallback boxes for class-declared capsules and meshes. This PR adds `_mjcf_geom_defaults` (flattens the class tree across `<include>` fragments, root class under `""`) and `_geom_attrs` (class attrs overridden by the element's own, class resolved as own `class` -> nearest enclosing `childclass` -> root), and routes all twelve geom attribute reads in the module through it. The precedence matches MJCF's own semantics, an undeclared class deliberately contributes nothing (MuJoCo refuses such a model itself and names the class), and the only verdict change in the 660-file corpus sweep is 8 `ability_hand_api` files that MuJoCo itself refuses - the fail-loud contract from #2459 now sees their missing assets instead of emitting silent box proxies. Public API signatures (`load_mjcf`, `load_mjcf_scene_objects`) are unchanged; every reworked function is module-private.

## What's good

- Verified locally at `e79093d1` (mujoco 3.8.0): the new file's 13 tests pass, and the full `tests/simulation/isaac/` suite shows an identical failure set on this branch and on the base commit `0f2c417c` (3 pre-existing environment failures, none in the loaders), so the change introduces no regression here.
- Every expected shape in the new tests is read back off a compiled `MjModel` rather than restated, and the enum-to-name map uses `int()` keys on both sides per AGENTS.md's value-based MuJoCo enum comparison rule.
- The AST rule test (`test_every_geom_attribute_read_goes_through_the_resolver`) carries its own non-vacuity guards (`scanned >= 20`, plus a reachability check on the four graded readers), so a future reader that asks the element directly is caught on arrival.
- Precedence controls pin both tempting shortcuts: a geom's own attribute still wins over its class, and a model with no `<default>` at all reads exactly as before.
- Changelog fragment present per AGENTS.md step 3; strings are ASCII-clean; no host paths in the test file.

## Verification suggestions

- `pytest tests/simulation/isaac/test_mjcf_default_class_geometry.py -q` on the declared mujoco floor (3.5.0) as well as current - the PR reports both green.
- Spot-check the headline regression against the shipped asset: `load_mjcf` on a downloaded `asimov_v0` MJCF should now report capsules with half-lengths matching `model.geom_size` on the compiled model for the six leg links.

PR #2671, opened 2026-08-23
fix(isaac): refuse a body reached by more than one joint

cagataycali:
## What

`_validate_kinematic_tree` is the fail-fast topology guard every procedural builder and every URDF / MJCF / USD loader runs before handing a `ProceduralRobot` back, and its docstring states the invariant it enforces:

> Every joint must reference valid body indices, and each non-root body must have exactly one inbound joint (tree topology, as USD/PhysX articulations require).

It counted duplicate `(parent_body, child_body)` **edges**. That catches only the case where the two joints share a parent -- an MJCF body carrying two `<joint>` children, i.e. a 2-DOF compound joint. URDF is a flat list of joints each naming `<parent>` and `<child>` explicitly, so two joints can name the same child with *different* parents. Those are two distinct edges, so the per-edge count sees nothing while the link still has two parents.

A joint whose `parent_body` is its own `child_body` loaded too: one inbound joint by count.

## Measured through the public `load_urdf`

A three-link arm whose `forearm` is the child of both `base` and `shoulder`:

| shape | `upstream/main` | this branch |
| --- | --- | --- |
| `forearm` reached by 2 joints with different parents | **accepted**, `num_joints=3` | refused, names `forearm` + both joints |
| a 3-cycle giving one body 2 parents | **accepted**, `num_joints=3` | refused |
| `parent_body == child_body` | **accepted**, `num_joints=2` | refused, names the joint |
| two joints sharing one edge (compound joint) | refused | refused, wording unchanged |
| well-formed chain | accepted, `num_joints=2` | accepted, `num_joints=2` |

`num_joints=3` for an arm that can carry two is the reported DOF count a caller sizes an action vector from. Nothing raises: `load_urdf` already refuses a duplicate `<link>`, an unknown parent or child link, a missing `<parent>` / `<child>` and an unknown joint type, so the topology guard is the one that should have caught this.

No articulation can instantiate a link with two parents, so a realizer has to pick one -- and **the two choices are different robots**. The artifact renders both from the one file.

## Change

Count inbound joints per **child** rather than per edge, and refuse a self-parented joint on its own terms.

The duplicate-edge branch keeps its wording and its remedy. Splitting with an intermediate massless link is the fix for a compound joint and the wrong advice for a link with two parents, which carries one joint too many -- so the two shapes stay distinguishable to the author who hit one of them. The refusal names the body, not just its index, because a URDF author wrote link names.

## Scope

Multiple roots stay legitimate: a body with no inbound joint is a root, the shipped `unitree_g1` builder has two, and so does an MJCF whose `<worldbody>` declares several top-level bodies. Connectivity and general acyclicity therefore remain out of scope -- the guard answers how many joints reach a body, not whether every body is reached -- and a 2-cycle in which each body is reached exactly once is still accepted. Both boundaries are pinned by tests either way, so widening them later is a decision rather than an accident.

## Blast radius

**638** registry MJCF / URDF assets loaded through `load_mjcf` / `load_urdf` on both trees: **0 changed verdicts** (415 load, 223 refuse, identically). All three shipped procedural builders (`so100`, `panda`, `unitree_g1`) build unchanged.

## Tests

`tests/simulation/isaac/test_one_inbound_joint_per_body.py`, 14 cases. Pre-fix split against `upstream/main`: **6 failed / 8 passed**, every failure in the regression class and none in the two control classes. Four fail behaviourally, e.g.

```
AssertionError: load_urdf accepted a robot no articulation can instantiate: bodies
reached by more than one joint {'forearm': ['elbow_from_shoulder', 'elbow_from_base']},
and it reports num_joints=3 with joint_names=['shoulder_pan', 'elbow_from_shoulder',
'elbow_from_base'].
```

Break table -- each break fires a distinct set, and none is dead:

| break | fires |
| --- | --- |
| exact pre-fix (`git checkout upstream/main -- procedural.py`) | 6 (the full regression set) |
| drop the self-parent check, keep the per-child count | **1** (the self-parent case alone) |
| drop the per-child count, keep the self-parent check | **5** |
| collapse the two branches into one message | **1** (the compound-joint control alone) |
| the label drops the body name | 3 |
| enforce exactly one root (over-reach) | 6, incl. 3 controls **and the shipped `unitree_g1` builder** |
| control (unmodified) | 0 of 14 |

Rows 2 and 3 are complementary (1 + 5 = 6), so the two halves are independently pinned. Row 4 shows the compound-joint remedy survived rather than being replaced. Row 6 is the over-reach: enforcing a single root refuses a robot this package ships.

## Verification

Measured at `50fa63ac`. The head has since moved to `aed5fe76`, which adds only the changelog fragment (`git diff 50fa63a..HEAD --name-only` is that one file).

- Gate: 194-file selection (everything naming the loaders / `ProceduralRobot` / the procedural registry, all of `tests/simulation/isaac`, plus the repo-wide source-walking guards), the identical selection on both trees with the new test file excluded from each -> **`9895 passed, 89 skipped`, byte-identical, zero failures on either side** (410.08s vs 409.42s).
- The pre-existing `tests/simulation/isaac/test_description_loader_geometry.py` passes unchanged, including its duplicate-edge case.
- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` **24 == 24** against a pristine `upstream/main` worktree, branch-only empty.
- New file: 14 passed, and 111 passed with random ordering alongside three sibling loader suites in `tests/simulation/isaac`.

## Artifact

Both columns run one capture script; only `strands_robots` differs. Each render is a MuJoCo realization of the `ProceduralRobot` the loader returned -- one proxy capsule per reported body, nested per the reported parent -> child edges. Colour and link length are for legibility.

`upstream/main` accepts the file, so the panel shows **both** realizations: honouring `elbow_from_shoulder` gives a 3-link chain (forearm 1356 px), honouring `elbow_from_base` puts the forearm straight out from the base (2885 px). One file, two robots.

The control row is the same arm written as a well-formed chain: accepted on both trees with `num_joints=2`, and the two renders are byte-identical (`np.array_equal` is `True`).

one inbound joint per body


yinsong1986 (approved):
## Summary

Tightens `_validate_kinematic_tree` from counting duplicate `(parent_body, child_body)` edges to counting inbound joints per child body, plus an explicit refusal for a self-parented joint -- closing the gap where a URDF link named as the `<child>` of two joints with different parents (or a self-loop) loaded and reported an inflated `num_joints`. The change enforces the invariant the guard's docstring already stated, so this is a bug fix rather than a contract change: the refusal is a loud `ValueError` naming the offending body and joints (per AGENTS.md, raise on fatal errors / no silent defaults), the duplicate-edge branch keeps its distinct wording and remedy so a compound-joint author is not sent the wrong advice, and the boundaries that must stay open (multiple roots, the 2-cycle acyclicity boundary) are pinned by controls rather than left to chance. I verified the guard's behaviour standalone against the head-SHA file: two-parent, self-parent and duplicate-edge shapes are refused with distinct messages; chain, multi-root and 2-cycle controls are accepted; and all three shipped builders (`so100` 6, `panda` 7, `unitree_g1` 21 joints) build unchanged. Refusal strings are ASCII-only, the test file carries no host paths, and the change ships a `changelog.d/` fragment rather than editing `CHANGELOG.md`.

## What's good

- The break table in the description shows each half of the fix (self-parent check, per-child count) is independently pinned, and the over-reach row demonstrates why single-root enforcement was correctly kept out of scope (it would refuse the shipped `unitree_g1`).
- Blast radius measured over 638 registry assets with 0 changed verdicts, so the stricter guard provably refuses only shapes no articulation could instantiate.
- `_load_or_refusal` reports the accepted topology on failure instead of a bare `DID NOT RAISE`, and test names describe behaviour, not provenance.
- `_body_label` bounds-checks the index rather than letting a malformed robot turn the refusal itself into an `IndexError`, and the fallback is pinned by its own test.

PR #2672, opened 2026-08-23
fix(isaac): read the root <default> class under both of its spellings

### LINKED ISSUE #2670: an MJCF root <default class="main"> is read as a different class from an unnamed one, so a model whose geoms name no class loses the root entirely
## What

`_mjcf_geom_defaults` (landed in #2669) keys the root default class on **the spelling the file happened to use**. MJCF gives that one class two spellings, and MuJoCo treats them as the same class:

```xml
<default>                  <!-- unnamed  -->
<default class="main">     <!-- named, same class -->
```

MuJoCo names the root `main` and refuses to let a model rename it, so those are the only two spellings a file has:

```
<default class="base">  ->  XML Error: top-level default class 'main' cannot be renamed
```

A geom reaches that class three ways - its own `class="main"`, an enclosing body's `childclass="main"`, or by naming no class at all. The resolver registers it under one key, so every geom arriving by the other spelling resolves to `{}` and falls through to `_extract_mjcf_shape`'s `box (0.05, 0.05, 0.05)` fallback.

## Two directions, and the second is the whole model

**a. Unnamed root, geom spells the name.** Noted as a possible follow-up in #2669's review: `<default>` plus `<geom class="main"/>` misses. Byte-identical to pre-#2669 behaviour, so not a regression.

**b. Named root, geom names no class.** Every geom in the model misses, because `classes[""]` is never written - and this is what shipped assets do. MuJoCo Menagerie's `pal_tiago_dual/tiago_dual.xml`:

```xml
<default class="main">
  <geom contype="0" conaffinity="1" group="1" type="mesh" mass="0"/>
```

| measured on `tiago_dual.xml` | value |
| --- | --- |
| geoms in the model | 46 |
| geoms carrying a `class` attribute | 0 |
| geoms with no `class` and no own `type`/`fromto` | **34** |
| what `_mjcf_geom_defaults` returns for `""` | `{}` |
| what it returns for `"main"` | `{contype, conaffinity, group, type: mesh, mass}` |
| shapes the loader can name after #2669 | 0 of 46 |

So 34 links report a 10 cm cube for the `type="mesh"` their own root class declares, under `load_mjcf` reporting success - the same failure mode #2669 was written to close, for a spelling it does not cover. `group` is read through the same resolver, so collision/visual group filtering is wrong for all 46.

## Why an alias is the fix and not a guess

Neither name can refer to a class other than the root, so registering the root under both cannot shadow anything. MuJoCo refuses both alternatives:

```
<default><default class="main">...   ->  XML Error: repeated default class name
<default><default>...               ->  XML Error: empty class name
```

## Separately

While measuring the above: #2669's code comment and PR body both state that MuJoCo "refuses a second top-level `<default>`", attributing the `cannot be renamed` error to that. Measured, two top-level `<default>` elements in one file are **accepted and merged** - that error only fires when a top-level `<default>` is given a name other than `main`. The comment is wrong about what it cites; the merge behaviour it implies the resolver can ignore is a distinct concern and gets its own issue.

## Verification

mujoco 3.12.0. Every claim above is a compiled `MjModel` read-back or a MuJoCo refusal string, not a restatement.

cagataycali:
## Problem

MJCF's root default class has two spellings and one identity. MuJoCo names the root `main` and refuses to let a model rename it, so those two are the only spellings a file has:

```xml
<default>                  <!-- unnamed -->
<default class="main">     <!-- named, the same class -->
```
```
<default class="base">  ->  XML Error: top-level default class 'main' cannot be renamed
```

A geom reaches that class three ways: its own `class="main"`, an enclosing body's `childclass="main"`, or by naming no class at all. `_mjcf_geom_defaults` (#2669) writes `classes[default_el.get("class", "")]`, so it registers the root under **whichever spelling the file happened to use** and every geom arriving by the other one resolves to `{}` and falls through to `_extract_mjcf_shape`'s `box (0.05, 0.05, 0.05)` fallback.

The two directions fail differently, and the second is the whole model.

**a. Unnamed root, geom spells the name.** `<default>` plus `<geom class="main"/>` misses. Byte-identical to pre-#2669 behaviour, so not a regression - this was noted as a possible follow-up when #2669 was reviewed.

**b. Named root, geom names no class.** `classes[""]` is never written, so *every* geom in the model misses. This is what shipped assets do. Menagerie's `pal_tiago_dual/tiago_dual.xml`:

```xml
<default class="main">
  <geom contype="0" conaffinity="1" group="1" type="mesh" mass="0"/>
```

| measured on `tiago_dual.xml` | before | after |
| --- | --- | --- |
| geoms in the model | 46 | 46 |
| geoms carrying a `class` attribute | 0 | 0 |
| resolver's entry for `""` | `{}` | `{contype, conaffinity, group, type: mesh, mass}` |
| resolver's entry for `"main"` | `{contype, ..., type: mesh, ...}` | same |
| geoms whose shape the loader cannot name | **34** | **0** |

So 34 links reported a 10 cm cube for the `type="mesh"` their own root class declares, under `load_mjcf` reporting success - the same failure mode #2669 closed, for a spelling it did not cover. `group` is read through the same resolver, so collision/visual group filtering was wrong for all 46.

## Change

Register the root class under both keys. `_flatten` returns the merged attributes it just stored, and the top-level loop writes them to `""` and to `"main"`.

That is an alias rather than a guess, because neither name is available to any other class - MuJoCo refuses both alternatives:

```
<default><default class="main">...   ->  XML Error: repeated default class name
<default><default>...               ->  XML Error: empty class name
```

So aliasing cannot shadow a distinct class. Both refusals are asserted in the tests (`TestWhyAliasingTheRootIsSafe`), so if a future MuJoCo permits either, the justification fails loudly instead of the alias silently merging two classes.

The fix is in the resolver, so all three arrival paths are covered at once: `_geom_attrs` looks up `geom.get("class") or childclass`, and both `""` and `"main"` now resolve.

### One correction folded in, one concern kept out

The code comment this replaces attributed the `cannot be renamed` error to MuJoCo refusing a *second* top-level `<default>`. Measured, two top-level `<default>` elements in one file are **accepted and merged**; that error only fires when a top-level `<default>` is given a name other than `main`. The comment is removed rather than restated.

What that merge implies for this resolver is a real second bug - the loop restarts from `{}` per element, so the last top-level `<default>` replaces the others and a `type` the first declared is dropped. It is a distinct concern with a distinct fix, so it is filed as #2673 with its own measurement rather than widened into this diff.

## Tests

`tests/simulation/isaac/test_mjcf_root_default_class_spellings.py`, 16 cases. MuJoCo is the oracle throughout: every expected shape is read back off a compiled `MjModel` rather than restated.

Pre-fix split against `main` (`39a6447`, i.e. with #2669 landed): **7 failed / 9 passed**.

| class | pre-fix | what it pins |
| --- | --- | --- |
| `TestTheRootClassAnswersToBothOfItsSpellings` | 4 fail of 6 | the regression, on both readers and all three arrival paths |
| `TestTheResolverPublishesBothSpellings` | 3 fail of 3 | the alias at the point it is made, with a non-vacuity clause |
| `TestWhyAliasingTheRootIsSafe` | 0 of 3 | MuJoCo's refusals - premises, expected to hold on both trees |
| `TestWhatTheAliasMustNotChange` | 0 of 4 | controls: the geom's own attribute still wins, an unnamed root still supplies an unclassed geom, an undeclared class still contributes nothing |

The failure messages name the symptom rather than the assertion, e.g.

```
the root class is spelled class='main' and the geom names no class, so the loader
reported 'box' with size (0.05, 0.05, 0.05) - the fallback box - for a capsule
```

Two notes on test design:

- `test_both_spellings_are_read_as_the_same_model` checks the two files agree **and** that they agree with MuJoCo. Agreement alone is satisfied by the bug - two readings that are both the fallback box are also equal - so it asserts against the oracle as well, and consequently fails pre-fix.
- `test_a_class_a_model_never_declares_still_contributes_nothing` exists because `"main"` is now always a key: it pins that making it unconditional did not turn the mapping into one that answers for every name.

## Verification

mujoco 3.12.0, re-measured at the head `e4d935a`.

| suite | result |
| --- | --- |
| the new file | 16 passed |
| the new file on `main` | 7 failed, 9 passed |
| `test_mjcf_default_class_geometry.py` + `test_mjcf_bodies_are_model_global.py` (#2669's and #2667's pins) | 44 passed |
| `tests/simulation/isaac/` | 1015 passed, 38 skipped, 0 failed |

The 30 collection errors in `tests/simulation/isaac/` are all one file (`test_add_camera_numeric_validation.py`) and all `ModuleNotFoundError: No module named 'strands'` - the Strands SDK is absent from the venv these were measured in, and they reproduce identically on `main`. `ruff check` and `ruff format --check` clean on both touched files; no non-ASCII in either or in the fragment; no host paths in the tests.

## Commits

| commit | concern |
| --- | --- |
| `e317950` | the fix and its pins |
| `da6fafa` | changelog fragment |
| `e4d935a` | the same docstring's opening still called the root class "unnamed"; prose only |

Fixes #2670

### DIFF
--- a/changelog.d/2672-isaac-mjcf-root-default-class-spellings.md
+++ b/changelog.d/2672-isaac-mjcf-root-default-class-spellings.md
@@ -1,0 +1,7 @@
+### Fixed: the root `<default>` class is read under both of its spellings
+
+MJCF's root default class has two spellings and one identity: MuJoCo names the root `main` and refuses to let a model rename it, so a file may leave its top-level `<default>` unnamed or write `class="main"` and mean the same class. A geom reaches that class by either name, or by naming no class at all.
+
+The Isaac loaders' default-class resolver keyed the root on whichever spelling the file used, so every geom arriving by the other one resolved to nothing and fell through to the `box (0.05, 0.05, 0.05)` fallback. In one direction that is an unnamed root plus `<geom class="main"/>`. In the other it is the whole model, and it is what shipped assets do: Menagerie's `pal_tiago_dual` writes `<default class="main">` declaring `type="mesh" group="1"` and gives none of its 46 geoms a class, so 34 of them - the ones carrying no `type` or `fromto` of their own - reported a 10 cm cube for a mesh under `load_mjcf` reporting success. `group` is read through the same resolver, so group filtering was wrong for all 46.
+
+The root is now registered under both keys, which cannot shadow a distinct class: MuJoCo refuses a nested `class="main"` ("repeated default class name") and a nested unnamed `<default>` ("empty class name"). Both refusals are asserted alongside the fix, so the justification for aliasing is pinned rather than remembered.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -949,15 +949,16 @@ _MJCF_ROOT_DEFAULT_CLASS = "main"
 def _mjcf_geom_defaults(root: ET.Element, base_dir: str) -> dict[str, dict[str, str]]:
     """Every ``<default>`` class's effective ``<geom>`` attributes, flattened.
 
-    MJCF's ``<default>`` elements form a tree: the unnamed top-level element is
-    the root class, a ``<default class="X">`` inherits its enclosing element's
-    attributes, and a ``<geom>`` takes its class's attributes for every
-    attribute it does not spell itself. So ``type``, ``size`` and ``fromto`` -
-    the three attributes that decide what shape a geom is - need not appear on
-    the geom at all. Read as the geom's own attributes alone, a link whose class
-    declares ``type="capsule" size="0.05"`` reports the default box however long
-    its ``fromto`` segment is, which also makes the endpoint reading
-    :func:`_extract_mjcf_shape` performs unreachable for it.
+    MJCF's ``<default>`` elements form a tree: the top-level element is the
+    root class whether or not it names itself, a nested ``<default class="X">``
+    inherits its enclosing element's attributes, and a ``<geom>`` takes its
+    class's attributes for every attribute it does not spell itself. So
+    ``type``, ``size`` and ``fromto`` - the three attributes that decide what
+    shape a geom is - need not appear on the geom at all. Read as the geom's
+    own attributes alone, a link whose class declares ``type="capsule"
+    size="0.05"`` reports the default box however long its ``fromto`` segment
+    is, which also makes the endpoint reading :func:`_extract_mjcf_shape`
+    performs unreachable for it.
 
     ``<default>`` is a top-level element, so it is model-global: the fragment
     declaring a class need not be the fragment declaring the geom, and neither
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -940,6 +940,12 @@ def _mjcf_model_toplevel(root: ET.Element, base_dir: str, _seen: frozenset[str] 
     return out
 
 
+# MJCF's implicit root default class. MuJoCo names it and will not let a model
+# rename it, so this is the one spelling a file can use for the root besides
+# leaving the top-level ``<default>`` unnamed.
+_MJCF_ROOT_DEFAULT_CLASS = "main"
+
+
 def _mjcf_geom_defaults(root: ET.Element, base_dir: str) -> dict[str, dict[str, str]]:
     """Every ``<default>`` class's effective ``<geom>`` attributes, flattened.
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Registers the MJCF root `<default>` class under both of its spellings (`""` and `"main"`) in the Isaac loaders' default-class resolver, so a geom reaching the root by the spelling the file did not use no longer falls through to the `box (0.05, 0.05, 0.05)` fallback. The named-root direction is the one shipped assets hit (Menagerie's `pal_tiago_dual`: 34 of 46 geoms misreported), and the fix lands in the resolver itself so all three arrival paths - `class="main"`, `childclass="main"`, and no class at all - resolve at once. The alias is justified by MuJoCo's own refusals (a nested class cannot be named `main` and cannot be unnamed), and both refusals are asserted in the tests so the justification fails loudly if a future MuJoCo relaxes it. Change is confined to a private function; no public API, wire format, or persisted schema is touched.

## What's good

- Tests use a compiled `MjModel` as the oracle rather than restated expectations, with the pre-fix failure split (7/16) documented per class; the same-model test correctly guards against the vacuous both-fallback agreement.
- Enum comparisons are value-based (`int(...)` on both sides) per AGENTS.md's MuJoCo-enum rule; test class names describe behaviour, not provenance; no non-ASCII, no host paths.
- Scope discipline: the distinct multi-top-level-`<default>` merge bug uncovered while removing the stale comment is filed as #2673 with its own measurement rather than widened into this diff.
- Changelog fragment follows `changelog.d/<pr>-<slug>.md`; `Fixes #2670` is in the body, not the title.

PR #2674, opened 2026-08-23
fix(isaac): merge a model's several top-level <default> elements

### LINKED ISSUE #2673: MuJoCo merges a model's several top-level <default> elements; _mjcf_geom_defaults lets the last one replace the rest, dropping the type the first declared
## What

`_mjcf_geom_defaults` recurses from each top-level `<default>` independently:

```python
for el in _mjcf_model_toplevel(root, base_dir):
    if el.tag == "default":
        root_attrs = _flatten(el, {})
        classes[""] = root_attrs
```

Each call starts from `{}` and overwrites `classes[""]`, so the last top-level `<default>` **replaces** the others rather than merging with them. MuJoCo merges them, exactly as it merges `<compiler>`, `<asset>` and `<worldbody>` - the model-global treatment `_mjcf_model_toplevel` exists to reproduce.

## Measured

```xml
<mujoco model="two-roots">
  <default><geom type="capsule" size="0.25 0.4"/></default>
  <default><geom size="0.05 0.3"/></default>
  <worldbody><body name="b"><geom name="g"/></body></worldbody>
</mujoco>
```

| reader | result |
| --- | --- |
| MuJoCo (compiled `MjModel`) | `capsule`, size `(0.05, 0.3)` - `type` from the first, `size` from the second |
| `_mjcf_geom_defaults` entry for `""` | `{'size': '0.05 0.3'}` |
| `_geom_attrs` for the geom | `{'size': '0.05 0.3', 'name': 'g'}` |
| `load_mjcf` | no `type` resolved, so the `box (0.05, 0.05, 0.05)` fallback |

The first element's `type="capsule"` is dropped, and the failure is the familiar silent one: a capsule reported as a 10 cm cube under a successful load.

A nested class keeps inheriting from its own root correctly (`class="leg"` resolves to `{'type': 'capsule', 'size': '0.06 0.2'}` in the same model), so the loss is confined to the root class - which is the class every geom that names none resolves against.

## Why this is reachable in practice

`<default>` is a top-level element and `_mjcf_model_toplevel` splices `<include>` fragments, so the several top-level `<default>` elements need not be in one file: a scene that includes a robot contributes one each, and the resolver keeps only whichever the splice yields last. Two `<default>` elements in a single file is the minimal reproduction above, not the only shape.

## Provenance

Surfaced while measuring #2670. Both #2669's code comment and its PR body stated that MuJoCo "refuses a second top-level `<default>`", citing `top-level default class 'main' cannot be renamed`. That is not what that error means - it fires when a top-level `<default>` is given a name other than `main`. Two unnamed top-level `<default>` elements compile fine and are merged. The incorrect comment is removed in #2672; this issue is the behaviour it was standing in for.

## Shape of the fix

Thread the accumulated root attributes through the loop instead of restarting from `{}`, so successive top-level elements merge in document order the way MuJoCo does. Worth pinning the fragment case as well as the single-file one, since the splice is what makes it more than a curiosity. mujoco 3.12.0 for all measurements above.

cagataycali:
## Problem

`<default>` is a top-level MJCF element, and MuJoCo merges every one a model carries into the single root class - the same model-global treatment it gives `<compiler>`, `<asset>` and `<worldbody>`, which is the treatment `_mjcf_model_toplevel` exists to reproduce.

`_mjcf_geom_defaults` recursed from each one independently:

```python
for el in _mjcf_model_toplevel(root, base_dir):
    if el.tag == "default":
        root_attrs = _flatten(el, {})     # restarts from {} per element
```

So the last top-level `<default>` **replaced** the others instead of merging with them.

```xml
<default><geom type="capsule" size="0.25 0.4"/></default>
<default><geom size="0.05 0.3"/></default>
```

| reader | result |
| --- | --- |
| MuJoCo (compiled `MjModel`) | `capsule`, size `(0.05, 0.3)` - `type` from the first, `size` from the second |
| resolver's entry for `""` | `{'size': '0.05 0.3'}` |
| `load_mjcf` | no `type` resolved, so the `box (0.05, 0.05, 0.05)` fallback |

The first element's `type="capsule"` is dropped, and the failure is the familiar silent one: a capsule reported as a 10 cm cube under a load reporting success. `group` is read through the same resolver, so collision/visual filtering is wrong for the same geoms.

## This is reachable in shipped assets, through `<include>`

No Menagerie file writes two top-level `<default>` elements itself. `<include>` is a textual splice, so it does not have to: a scene including a robot contributes one element each, and the resolver kept only whichever the splice yielded last.

Measured over all 261 Menagerie MJCFs - **7** carry two top-level `<default>` elements once spliced, and **4 change verdict**:

| model | resolver's root class before | after |
| --- | --- | --- |
| `pal_tiago/tiago_position.xml` | `{}` | `{contype, conaffinity, group, mass, type: mesh}` |
| `pal_tiago/scene_position.xml` | `{}` | same |
| `pal_tiago_dual/tiago_dual_position.xml` | `{}` | same |
| `pal_tiago_dual/scene_position.xml` | `{}` | same |

The whole root class was lost, so every geom naming no class resolved to nothing:

| model | geoms with no resolvable `group` | with no resolvable `type` |
| --- | --- | --- |
| `pal_tiago/tiago_position.xml` | **15** of 35 -> 0 | 14 -> 0 |
| `pal_tiago_dual/tiago_dual_position.xml` | **11** of 43 -> 0 | 10 -> 0 |

Being precise about the harm on these four: the reported *shape* does not change, because their root class declares `type="mesh"` and `_extract_mjcf_shape` maps a mesh to the same fallback box either way. What was wrong is the `group` read - the visual/collision split `_find_body_mesh` and `_geom_aabb` both branch on. The shape harm is the one the minimal reproduction above shows, and it needs only a root class declaring an analytic `type`.

The remaining 3 of the 7 (`aloha/scene.xml`, `pal_talos/*`) do not change: neither of their spliced elements carries a direct `<geom>`, so the root class is empty on both trees. That is the non-vacuity control - two elements alone are not sufficient, the loss needs a `<geom>` in one of them.

Blast radius overall: **261** models loaded through `load_mjcf` on both trees, 201 load / 60 refuse **identically**, 4 changed verdicts, all 4 listed above, no model changed between loading and refusing.

## Change

Accumulate the root attributes across the loop rather than restarting at each element - one line, plus the reasoning.

```diff
+   root_attrs: dict[str, str] = {}
    for el in _mjcf_model_toplevel(root, base_dir):
        if el.tag == "default":
-           root_attrs = _flatten(el, {})
+           root_attrs = _flatten(el, root_attrs)
```

### The boundary, which is not the obvious one

Threading the root must **not** become "merge every element into every class". A nested class is snapshotted where it appears, and MuJoCo says so - measured on mujoco 3.12.0:

```xml
<default><default class="leg"><geom size="0.06 0.2"/></default></default>
<default><geom type="capsule" size="0.5 0.6"/></default>
```

MuJoCo compiles `class="leg"` as a **sphere** - its own default - not the `capsule` the *following* top-level element declares. `_flatten` recurses into nested classes at the moment it processes each element, so it reproduces that ordering for free; the fix is correct because of where the recursion happens, not in spite of it. That boundary is pinned, so widening the merge later is a decision rather than an accident.

## Tests

`tests/simulation/isaac/test_mjcf_merged_toplevel_defaults.py`, 16 cases. MuJoCo is the oracle throughout: every expected shape is read back off a compiled `MjModel` rather than restated.

Pre-fix split against `main` (`2f18d9f5`, i.e. with #2669 and #2672 landed): **10 failed / 6 passed**.

| class | pre-fix | what it pins |
| --- | --- | --- |
| `TestSeveralTopLevelDefaultsMergeIntoOneRootClass` | **7 fail of 7** | the regression: two and three elements, both root spellings in either order, all three arrival paths, and the `<include>` splice |
| `TestTheResolverMergesRatherThanReplaces` | **2 fail of 2** | the merge at the point it is made, and that the `"main"` alias publishes the merged class |
| `TestNestedClassesFollowDocumentOrder` | 1 fail of 3 | the boundary above, plus MuJoCo's `repeated default class name` refusal as a premise |
| `TestWhatTheMergeMustNotChange` | 0 of 4 | controls: one element unchanged, the geom's own attribute still wins, two bare elements contribute nothing, an undeclared class contributes nothing |

Failure messages name the symptom rather than the assertion, e.g.

```
a second top-level <default> overriding only size dropped the first one's type:
MuJoCo reads 'capsule', and the loader reported 'box' with size (0.05, 0.05, 0.05)
```

### Break table - each break fires a distinct set, and none is dead

| break | fires |
| --- | --- |
| control (unmodified) | **0** of 16 |
| exact pre-fix (`_flatten(el, {})`) | 10 (the full regression set) |
| over-merge: give every nested class the whole model's root merge, ignoring document order | **1** (the scoping boundary alone) |
| earlier declaration wins instead of later | 9 |
| the `"main"` alias publishes one element instead of the merge | 3 |
| nested classes not registered at all (over-reach) | 3, incl. the bare-elements control |

Rows 2 and 3 are disjoint and complementary: the plausible-but-wrong reading of "MuJoCo merges them" - merge everything into everything - passes all 10 regression cases and is caught by exactly one test, which is the reason that test exists.

## Verification

mujoco 3.12.0, measured at the head `3c49c6aa`.

| suite | result |
| --- | --- |
| the new file | 16 passed |
| the new file on `main` | 10 failed, 6 passed |
| #2667's, #2669's and #2672's pins + `test_description_loader_geometry.py` | 126 passed |
| `tests/simulation/isaac/` | **1045 passed, 38 skipped, 0 failed** |
| the new file with 3 sibling MJCF suites, one selection | 76 passed |

The 30 collection errors in `tests/simulation/isaac/` are all one file (`test_add_camera_numeric_validation.py`) and all `ModuleNotFoundError: No module named 'strands'` - the Strands SDK is absent from the venv these were measured in, and they reproduce identically on `main` (`10 failed, 1035 passed, 38 skipped, 30 errors`, the 10 being this branch's new pins).

`ruff check` and `ruff format --check` clean on both touched files. `mypy` **63 == 63** against `main` for the same file, branch-only empty. No non-ASCII in either file or in the fragment; no host paths in the tests - every model is written under `tmp_path`.

## Commits

| commit | concern |
| --- | --- |
| `4fefea90` | the fix and its pins |
| `3c49c6aa` | changelog fragment |

Fixes #2673

### DIFF
--- a/changelog.d/2673-isaac-mjcf-merge-toplevel-defaults.md
+++ b/changelog.d/2673-isaac-mjcf-merge-toplevel-defaults.md
@@ -1,0 +1,9 @@
+### Fixed: a model's several top-level `<default>` elements merge into one root class
+
+`<default>` is a top-level MJCF element, and MuJoCo merges every one a model carries into the single root class - the same model-global treatment it gives `<compiler>`, `<asset>` and `<worldbody>`. The merge is per attribute and in document order, so a later element overriding `size` does not discard a `type` an earlier one declared.
+
+The Isaac loaders' default-class resolver read each top-level element independently, restarting from nothing, so the last one **replaced** the others. The dropped attribute then failed the familiar silent way: a geom resolving against the root class found no `type`, so `load_mjcf` reported the `box (0.05, 0.05, 0.05)` fallback for a capsule under a successful load. `group` is read through the same resolver, so collision/visual filtering was wrong for the same geoms.
+
+The shape does not require a file that writes two `<default>` elements itself, which is what makes it more than a curiosity: `<include>` is a textual splice, so a scene including a robot contributes one element each. Measured on Menagerie, 7 models carry two top-level `<default>` elements once spliced, and `pal_tiago` and `pal_tiago_dual` lost their whole root class to the replace - 15 of 35 and 11 of 43 geoms respectively had no resolvable `group`.
+
+The root class is now accumulated across the elements rather than restarted at each. A nested class is still flattened where it appears, which is what MuJoCo does: it inherits the root as accumulated up to its own position, so a `type` declared by a top-level element *after* the one enclosing it does not reach it. That boundary is pinned, so widening the merge later is a decision rather than an accident.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -965,6 +965,20 @@ def _mjcf_geom_defaults(root: ET.Element, base_dir: str) -> dict[str, dict[str, 
     need be the top file - the same splice :func:`_mjcf_model_toplevel` resolves
     for ``<compiler>`` and ``<asset>``.
 
+    A model may carry SEVERAL top-level ``<default>`` elements, and MuJoCo
+    merges them into the one root class in document order - the same treatment
+    it gives ``<compiler>``, ``<asset>`` and ``<worldbody>``. They are therefore
+    accumulated here, not read independently. The merge is per attribute, so a
+    later element overriding ``size`` does not discard a ``type`` an earlier one
+    declared.
+
+    A nested class is flattened where it appears, which is what MuJoCo does: it
+    inherits the root class as accumulated *up to its own position*, so a
+    ``type`` declared by a top-level element AFTER the one enclosing it does not
+    reach it. Measured on mujoco 3.12.0 - a nested class in the first of two
+    top-level elements compiles with MuJoCo's own sphere default, not the
+    ``type="capsule"`` the second element declares.
+
     Returns a mapping from class name to that class's merged ``<geom>``
     attributes, with the root class under both of the spellings that reach it -
     ``""`` and ``"main"``. A class a geom names but no ``<default>`` declares
@@ -987,6 +1001,12 @@ def _mjcf_geom_defaults(root: ET.Element, base_dir: str) -> dict[str, dict[str, 
             _flatten(nested, merged)
         return merged
 
+    # The root class accumulates across the model's top-level ``<default>``
+    # elements rather than restarting at each one, because MuJoCo merges them.
+    # Starting from ``{}`` per element lets the last one REPLACE the others, so a
+    # ``type`` only the first declares is dropped and every geom resolving
+    # against the root reports the fallback box under a successful load.
+    root_attrs: dict[str, str] = {}
     for el in _mjcf_model_toplevel(root, base_dir):
         if el.tag == "default":
             # A top-level ``<default>`` is the root class under either of two
@@ -1005,7 +1025,7 @@ def _mjcf_geom_defaults(root: ET.Element, base_dir: str) -> dict[str, dict[str, 
             # because neither name is available to one: MuJoCo refuses a nested
             # ``class="main"`` ("repeated default class name") and a nested
             # unnamed ``<default>`` ("empty class name").
-            root_attrs = _flatten(el, {})
+            root_attrs = _flatten(el, root_attrs)
             classes[""] = root_attrs
             classes[_MJCF_ROOT_DEFAULT_CLASS] = root_attrs
     return classes
--- a/tests/simulation/isaac/test_mjcf_merged_toplevel_defaults.py
+++ b/tests/simulation/isaac/test_mjcf_merged_toplevel_defaults.py
@@ -1,0 +1,399 @@
+"""A model's several top-level ``<default>`` elements are one root class, merged.
+
+``<default>`` is a top-level MJCF element, and MuJoCo merges every one a model
+carries into the single root class - the same model-global treatment it gives
+``<compiler>``, ``<asset>`` and ``<worldbody>``. The merge is per attribute and
+in document order, so a later element overriding ``size`` does not discard a
+``type`` an earlier one declared.
+
+Read independently, each element restarting from nothing, the last one REPLACES
+the others. The dropped attribute then fails the familiar silent way: a geom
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

Fixes the Isaac loaders' MJCF default-class resolver so a model's several top-level `<default>` elements merge into one root class in document order (per attribute, as MuJoCo does) instead of the last element replacing the others. The fix is a single line - thread `root_attrs` through the loop rather than restarting from `{}` per element - and it correctly preserves MuJoCo's non-obvious nested-class scoping (a nested class inherits the root only as accumulated up to its own position), which falls out of where `_flatten` recurses. `_mjcf_geom_defaults` is a private resolver, so there is no public API, wire-format, or schema surface in play.

## What's good

- Verified independently: the 16 new tests pass at the head, revert the one line and exactly 10 fail / 6 pass, matching the PR description's pre-fix split; the sibling MJCF suites (default-class geometry, root-class spellings, model-global bodies/meshes, description-loader geometry) all pass on the branch.
- The tests also hold on mujoco 3.8.0, not just the 3.12.0 the PR measured on - the MuJoCo-as-oracle design pays off across versions.
- The scoping-boundary test (`TestNestedClassesFollowDocumentOrder`) pins the one wrong-but-plausible implementation (merge everything into everything) that passes all 10 regression cases, per the PR's break table.
- AGENTS.md compliance throughout: `int()` enum comparisons with the reason commented, behavioral test names, `tmp_path`-rooted fixtures, ASCII-only, changelog fragment (issue-numbered, which `changelog.d/README.md` explicitly permits), ruff clean on both touched files.

PR #2675, opened 2026-08-23
fix(isaac): read every joint attribute through its MJCF default class

cagataycali:
## Problem

MJCF lets a `<default class="...">` supply a `<joint>`'s attributes, so a joint element need not spell its own `type`, `axis`, `range`, `damping` or `armature`. The isaac MJCF robot loader read every one of them off the element:

```python
jtype = "prismatic" if joint.get("type") == "slide" else "revolute"
axis = _parse_vec3(joint.get("axis"), (0.0, 0.0, 1.0))
rng = joint.get("range")
```

so a class-declared joint fell to the loader's own defaults. aloha's gripper finger is the clean case - its element is `<joint name="left/left_finger" class="left_finger"/>` and nothing else:

| field | reported | MuJoCo's own model |
| --- | --- | --- |
| type | `revolute` | `prismatic` |
| axis | `(0.0, 0.0, 1.0)` | `(0.0, 0.0, -1.0)` |
| range | `(-3.14159, 3.14159)` | `(0.0, 0.041)` |
| damping | `0.1` | `40.0` (declared) |
| armature | `0.01` | `0.243` (declared) |

Five of five wrong, under `load_mjcf` reporting success. `type="slide"` comes from class `finger`, `range="0 0.041"` and `axis="0 0 -1"` from the nested `left_finger`, and the loader saw neither. `num_joints` and every `JointDef` are the function's product, so a caller sizing an articulation reads a 41 mm prismatic finger as a revolute joint free to turn 360 degrees about the wrong axis.

Graded against MuJoCo's compiled model over the 1504 joints in the 97 registry assets that declare a `<joint>` inside a `<default>`:

| tree | agrees with MuJoCo | disagrees |
| --- | --- | --- |
| upstream/main | 1093 | **411** |
| this branch | **1504** | 0 |

Every change moves toward MuJoCo's answer and none away. 52 of those 97 assets change, including panda (6 of 9 joints), ur5e and ur10e (16 of 16), iiwa14 (7 of 7), xarm7 (10 of 13), go2 (12 of 12), g1_mjx (29 of 29), shadow_hand (24 of 24), allegro and leap_hand (16 of 16), so_arm100 (6 of 6), aloha (14 of 16), robotiq_2f85 (8 of 8); four grippers (piper, arx_l5, yam, panda hand) go from revolute to prismatic.

The same reader already resolves a geom's class (#2669). Its resolver was named `_geom_attrs` and `_mjcf_geom_defaults` and read `<default>/<geom>` only, so one file carried two rules for one MJCF feature.

## Change

`_mjcf_class_defaults(root, model_dir, tag)` and `_class_attrs(element, defaults, inherited)` replace the geom-only pair and take the element tag, so one rule answers "what did this element declare" for both `<geom>` and `<joint>`. `_extract_mjcf_joint` resolves its five attributes through it. Nested classes, `childclass`, an `<include>`d fragment and element-over-class precedence all come from the existing rule rather than a second implementation; the two tag namespaces stay separate, because a class carrying both a shape and a degree of freedom must not leak one into the other.

## Verification

Measured at `15d7f6dc`. The sub-sections below record what moved since the first measurement: **two** base merges and one review round. Each table names the sha it was measured at.

### Rebased onto a moved base

#2671 and #2672 merged while this was open, and #2672 touched this same resolver: it taught the geom collector that MJCF's root default class answers to both `""` and `"main"`, because MuJoCo names it `main` and refuses to let a model rename it. Composed with the tag generalisation here, that aliasing now covers **joint** classes too - a joint reaching the root class by `class="main"` was a second instance of the same miss, and neither side alone fixed it.

#2672 also corrected the resolver docstring's description of the root as "unnamed"; the merged docstring keeps that correction. Its test file named the pre-rename symbol at four sites, so those follow the rename and are otherwise unchanged - `tests/simulation/isaac/` is **1109 passed, 35 skipped** with all three changes together. The pre-fix split, `ruff` and the mypy baseline (24 == 24) were re-measured at the merged head and are unchanged.

Pre-fix split at `caaeec68`, the new file copied onto the then-`upstream/main`: **11 failed / 22 passed**, all 11 in the two regression classes and none in the three control classes. Post-fix 33 passed. The table below was measured at that sha; the second base merge moved the basis, and the re-measurement is further down.

| break | fails | fires |
| --- | --- | --- |
| control | 0 of 33 | - |
| exact pre-fix `loaders.py` | 14 | the 11 regressions, plus 3 geometry tests naming the renamed helpers |
| joints resolved against the geom defaults | 10 | 8 regressions + the no-merge test + the arm oracle |
| `childclass` no longer reaches a joint | **1** | the `childclass` test alone |
| classes read from the top file only | **2** | the `<include>` test in each file (shared rule) |
| a nested class no longer inherits | 17 | 10 joint + 7 geom (shared rule) |
| the class overrides the element | **3** | both joint precedence controls + the geometry file's own precedence test |
| the structural scan reaches nothing | **1** | the non-vacuity clause alone |

Seven breaks, seven distinct firing sets, no dead guards. The over-reach row is refused by the shipped geometry suite, not only by my controls.

Gate: 132 test files touching the changed symbols, the isaac suite and every guard that walks a source tree or grades docstrings, run with the identical selection on both trees and the new file excluded from each - `40 failed, 4684 passed, 90 skipped, 126 errors` byte-identical (284.79s vs 304.89s), 166 failing ids, `comm` empty in both directions. All 166 are `device_connect_edge` absent locally (326 occurrences); it is declared in the `[device-connect]` extra, which `[all]` pulls in, so CI installs it.

The new file passes on mujoco 3.12.0 and on 3.5.0, the declared floor. `ruff check` and `ruff format --check` clean; `mypy strands_robots tests tests_integ` 24 == 24 against a pristine worktree with nothing branch-only.

#### Second base merge: the merged root class

#2674 then landed the accumulator that merges a model's several top-level `<default>` elements into the one root class, while this branch renames the same resolver and parameterises it by tag. The code auto-merged - both the accumulator and the `find(tag)` read survive - and the only conflict was the docstring, resolved by keeping #2674's two paragraphs and their mujoco-3.12.0 measurement verbatim and generalising the `size`/`type` example to "one attribute `tag` selects".

**The composition is better than either input**, and none of the three trees is a matter of opinion - MuJoCo compiles the fixture, so it is the oracle. A joint whose kind comes from one top-level element and whose travel comes from another:

| tree | reported | agrees with MuJoCo |
| --- | --- | --- |
| `upstream/main` (accumulator, element-only joint read) | `revolute` `(-3.14159, 3.14159)` | no |
| this branch before the merge (tag resolver, no accumulator) | `revolute` `(-0.04, 0.04)` | no |
| the merge | `prismatic` `(-0.04, 0.04)` | **yes** |

Main reads the element only, so it reports both fallbacks; this branch alone took the last element and kept the travel while dropping the kind and the axis. That half was pinned by nothing - disabling the accumulator fired ten tests, **all of them geom** - so `test_a_joint_sees_every_top_level_default_element_merged` now states it, with every expected value premise-asserted to differ from the loader's own fallback so the assertions cannot hold vacuously. The axis matters here: the loader's fallback axis is `(0, 0, 1)`, so a fixture that used it would pass whether or not the merge was read.

The rename broke #2674's new test file, which the auto-merge could not see: it imported the pre-rename symbol at two sites, so all 399 lines failed at collection with an `ImportError`. Both sites now call `_mjcf_class_defaults(root, base_dir, "geom")`, and its 16 tests pass.

Measured at `15d7f6dc`:

| break | fails | fires |
| --- | --- | --- |
| control | 0 of 66 | - |
| the accumulator disabled | 11 | 10 of #2674's own tests **plus the new joint test** |
| the fixture set to the loader's own fallback | **1** | the new test's premise (non-vacuity) |
| #2674's test file un-repaired | 1 error | its collection |

Pre-fix split re-measured against the new base, my three test files copied onto `upstream/main`: **19 failed / 31 passed** over 50 tests. Fifteen are behavioural or structural and each names the defect - `assert 'revolute' == 'prismatic'` five times, `assert 0.1 == 4.2` for damping, `assert (-3.14159, 3.14159) == approx((-1.85005, 1.25664))` for the range, and the new test's `the first top-level <default>'s type did not reach the joint: MuJoCo reads 'prismatic', loader reported 'revolute'`. The other four report only that the resolver is renamed.

Gate re-run at the merged head: 305 test files - the isaac suite, every guard that walks a source tree or grades docstrings, sources, strings or changelog fragments, and everything naming the changed symbols - identical selection on both trees with the four changed test files excluded from each. **`17 failed, 14375 passed, 64 skipped` byte-identical** (495.13s vs 497.00s), `comm` empty in both directions. All 17 are `tests/mesh/test_iot_*` on absent `awsiot`, declared as `awsiotsdk>=1.21.0,<2.0.0` in the `[mesh-iot]` extra, so CI installs it.

The four changed test files are **66 passed on mujoco 3.5.0, the declared floor, and on 3.12.0**, what CI resolves. `tests/simulation/isaac` is **1156 passed, 5 skipped** under random ordering. `ruff check` and `ruff format --check` clean; `mypy strands_robots tests tests_integ` **24 == 24** against a pristine worktree, identical message sets, nothing branch-only, none in the changed files.

### Reading the fallback the file names

CodeQL flagged `_FALLBACK_LIMIT` as declared and never read. It was right, and it was pointing at two missing assertions rather than at a stray constant: that constant was the only module-level one in the file with zero load-reads, while each of its five siblings had at least one.

`test_a_class_declared_range_bounds_the_joint` asserted only that the compiled range reached the joint. Its axis sibling and its damping/armature sibling each first assert that the fixture differs from the loader's fallback, because otherwise a fixture whose value happened to be the fallback would pass whether or not the class was read at all. `test_an_undeclared_class_contributes_nothing` pinned two of the three fallbacks a joint gets when no class declares otherwise and left the range unasserted, so a change to that fallback went unreported on the one path whose whole subject is the fallback.

The split and the break table above are unchanged: both new assertions hold on the pre-fix source too, because they grade the fixture's intent and the fallback constant rather than the resolver. Measured against the three MJCF default-class files (49 tests):

| break | result | fires |
| --- | --- | --- |
| range fixture set to the fallback `-3.14159 3.14159` | 1 failed | the new premise only |
| same fixture, premise removed | **49 passed** | nothing - the test measures nothing and reports success |
| loader's MJCF joint fallback range changed | 1 failed | `test_an_undeclared_class_contributes_nothing` only |

Those first two rows together are the hole. Re-gated at this head over 189 test files - every `tests/simulation/isaac` file plus every guard that walks a source tree or grades docstrings, strings or changelog fragments - with the identical selection on both heads: `9651 passed, 59 skipped` byte-identical, zero failures either side (425.37s vs 425.25s). mypy 24 == 24 with identical message sets and nothing in the changed file; 1139 passed across the isaac package under random ordering.

## Artifact

aloha's finger driven from its reported lower limit to its reported upper limit, same file and camera in every row. The third row is MuJoCo's own compiled model, and the branch's five frames are identical to it pixel for pixel (`np.array_equal`); upstream/main's are up to 8.90 mean levels away. Tip travel 0.0 mm against 41.0 mm.

class-declared joint attributes

## Scope

`<default>` also carries `<mesh>`, `<material>` and `<site>` children. The joint and geom readers are the two that read attributes off an element, so those three are out of scope here; a class-declared `<site>` or `<material>` is a separate consumer with its own semantics. Element-over-class precedence is unchanged, and `TestTheJointsOwnAttributesWin` pins it.


strands_robots/simulation/isaac/loaders.py
```diff
@@ -940,88 +946,77 @@ def _mjcf_model_toplevel(root: ET.Element, base_dir: str, _seen: frozenset[str]
...
-        if geom_el is None:
+    def _attrs_of(default_el: ET.Element) -> dict[str, str]:
+        child = default_el.find(tag)
+        if child is None:
             return {}
-        return {k: v for k, v in geom_el.attrib.items() if k != "class"}
+        return {k: v for k, v in child.attrib.items() if k != "class"}
 
-    classes: dict[str, dict[str, str]] = {"": {}, _MJCF_ROOT_DEFAULT_CLASS: {}}
+    classes: dict[str, dict[str, str]] = {"": {}}
```
yinsong1986: [MUST FIX] This branch predates #2672 (`c0cfc766`, on `main`, `behind_by: 2`), and `_mjcf_class_defaults` was derived from the pre-#2672 `_mjcf_geom_defaults` — so this initializer is `{"": {}}` where `main`'s resolver now registers the root class under **both** of its spellings (`""` and `"main"` / `_MJCF_ROOT_DEFAULT_CLASS`), with `_flatten` returning the root's merged attributes to alias them. As written, merging this PR reverts that shipped fix (Menagerie's `pal_tiago_dual` writes `<default class="main">` and gives none of its 46 geoms a class; 34 of them fall back to the default box), and the new joint side inherits the same gap: a joint reaching a `class="main"` root by naming no class resolves to `{}`.

The conflict is not hypothetical — the PR reads `mergeable: false` / `dirty` right now because #2672 rewrote this same region, so a base merge is owed before anything can land. Two things must survive that resolution:

1. Carry #2672's dual-spelling root-class registration into the unified `_mjcf_class_defaults`, applied to **both** tags (the alias comment in `main`'s version explains why registering both names cannot shadow a distinct class).
2. `main`'s `tests/simulation/isaac/test_mjcf_root_default_class_spellings.py` calls `mod._mjcf_geom_defaults(...)` at three sites — a name this PR removes. Update those call sites to `_mjcf_class_defaults(root, ".", "geom")` (and keep its `defaults[""] == defaults["main"]` assertions passing).

Why this can't be deferred: this is the semantic-composition class AGENTS.md > PR Workflow step 8 documents for #1763/#1766 — the branch's green is a statement about a tree that does not contain #2672, and a conflict resolution that keeps this function as-is regresses a fix already on `main` (wrong geometry for shipped assets, silently, under `load_mjcf` reporting success). The merge is mechanically blocked until this region is reconciled, so the reconciliation itself is the thing under review.

yinsong1986 (commented):
## Summary

The MJCF robot loader read a `<joint>`'s `type`, `axis`, `range`, `damping` and `armature` off the element alone, so a class-declared joint (aloha's finger: `<joint name="left/left_finger" class="left_finger"/>`) fell to the loader's own defaults — five of five fields wrong under a successful load. This PR generalises the existing geom-side default-class resolver (`_geom_attrs` / `_mjcf_geom_defaults` → `_class_attrs` / `_mjcf_class_defaults(root, dir, tag)`) so one rule answers "what did this element declare" for both tags, with the two tag namespaces kept separate, and routes all five joint reads through it. Graded against MuJoCo's compiled model over 1504 joints in 97 registry assets, disagreement goes from 411 to 0.

## What's good

- The regression suite uses MuJoCo's own compiled model (`jnt_type` / `jnt_axis` / `jnt_range`) as the oracle rather than restating expectations, with premise asserts that keep a fixture from quietly pinning the wrong answer.
- The AST scan (`TestOneRuleAnswersWhatAJointDeclared`) derives its graded set from the module and pins its own non-vacuity, so a second MJCF joint reader is held to the rule on arrival — mirroring the shipped geom-side scan.
- Out-of-scope contracts (`compiler angle` units, `limited="false"`) are pinned as contracts of their own instead of silently drifting.
- Changelog fragment present per AGENTS.md step 3; renamed helpers are private (not in `__all__`); test names describe behaviour, not provenance.

## Verification suggestions

After the base refresh, compose against `main`'s `tests/simulation/isaac/test_mjcf_root_default_class_spellings.py` and run it together with the two suites this PR touches — that file is the pin for the regression flagged inline, and the branch's own CI cannot see it until the branch absorbs `main`.

strands_robots/simulation/isaac/loaders.py
cagataycali: Both requirements are already satisfied on the branch head, by the base merge in `6895aed3` -- which landed about two minutes after this comment was written, so the state it quotes (`mergeable: false` / `dirty` / `behind_by: 2`) is no longer the branch's. `upstream/main` (`2f18d9f5`, which contains `c0cfc766`) is now an ancestor of the head, and the PR reads `MERGEABLE`.

Verified by symbol rather than by inspection:

**1. The dual-spelling root-class registration survived, and it now covers joints too.** `_mjcf_class_defaults` initializes `{"": {}, _MJCF_ROOT_DEFAULT_CLASS: {}}` -- not `{"": {}}` -- and the top-level loop assigns `_flatten`'s merged root attributes to both keys. #2672's alias comment (the `pal_tiago_dual` measurement and the "cannot shadow a different class" argument) is carried verbatim. Because the unified resolver is parameterised by `tag`, that aliasing applies to `joint` as well as `geom`, so the gap you name -- a joint reaching a `class="main"` root by naming no class -- resolves to the root's merged attributes rather than `{}`. That is strictly more than either change shipped alone: #2672 aliased the root for geoms, this PR generalises the resolver by tag, and only the composition aliases it for joints.

**2. The call sites moved.** `_mjcf_geom_defaults` has **0 occurrences** across `strands_robots/` and `tests/`. `test_mjcf_root_default_class_spellings.py` calls `_mjcf_class_defaults(root, ".", "geom")` at all four sites, and its `defaults[""] == defaults["main"]` / `defaults["main"] == {}` assertions are unchanged and passing.

**The composition is load-bearing, not merely present.** Reverting only this PR's resolver (`git checkout upstream/main -- strands_robots/simulation/isaac/loaders.py`) fails 18 tests, and 4 of them are #2672's own:

```
test_mjcf_root_default_class_spellings.py::TestTheResolverPublishesBothSpellings::test_the_root_class_is_registered_under_both_names[<default>]
test_mjcf_root_default_class_spellings.py::TestTheResolverPublishesBothSpellings::test_the_root_class_is_registered_under_both_names[<default class="main">]
test_mjcf_root_default_class_spellings.py::TestTheResolverPublishesBothSpellings::test_a_model_with_no_default_reports_both_names_as_empty
test_mjcf_root_default_class_spellings.py::TestWhatTheAliasMustNotChange::test_a_class_a_model_never_declares_still_contributes_nothing
```

plus 3 in `test_mjcf_default_class_geometry.py` and 11 in the joint file. A resolution that had kept the pre-#2672 function would have failed #2672's tests rather than passing them silently, which is the property you were asking for.

The three files run green together on both the version CI resolves and the declared floor: 49 passed on mujoco 3.12.0 and 49 passed on 3.5.0. The whole `tests/simulation/isaac` package is 1139 passed under random ordering.

tests/simulation/isaac/test_mjcf_default_class_joints.py
```diff
@@ -0,0 +1,504 @@
...
+
+# MuJoCo's own numbering for mjtJoint, and the loader's mapping onto the
+# articulation vocabulary ``JointDef`` speaks.
+_MJ_JOINT_KIND = {0: "free", 1: "ball", 2: "slide", 3: "hinge"}
+_KIND_TO_JOINT_TYPE = {"hinge": "revolute", "slide": "prismatic", "ball": "fixed", "free": "fixed"}
+
+# The loader's own fallbacks - what a joint gets when nothing declares otherwise.
+_FALLBACK_TYPE = "revolute"
+_FALLBACK_AXIS = (0.0, 0.0, 1.0)
+_FALLBACK_LIMIT = (-3.14159, 3.14159)
```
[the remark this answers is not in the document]
cagataycali: Fixed in `caaeec68`. The alert was right, and it was pointing at two missing assertions rather than at a stray constant: `_FALLBACK_LIMIT` was the only module-level constant in the file with zero load-reads, while each of its five siblings had at least one.

Both sites where it belongs now read it.

`test_a_class_declared_range_bounds_the_joint` asserted only that the compiled range reached the joint. Its axis sibling and its damping/armature sibling each first assert that the fixture differs from the loader's fallback, because without that a fixture whose value happened to be the fallback would pass whether or not the class was read at all -- the test could not tell "the class-declared range reached the joint" from "the loader fell back". The fixture's `-1.85005 1.25664` was chosen to differ and nothing pinned that intent, so the premise assertion is now there alongside `limits is not None`.

`test_an_undeclared_class_contributes_nothing` pinned two of the three fallbacks a joint gets when no class declares otherwise (`_FALLBACK_TYPE`, `_FALLBACK_AXIS`); the range went unasserted, so a change to that fallback was unreported on the one path whose whole subject is the fallback.

Both are load-bearing, measured against the three MJCF default-class files (49 tests):

| break | result | fires |
| --- | --- | --- |
| control | 49 passed | -- |
| range fixture set to the fallback `-3.14159 3.14159` | 1 failed | the new premise only, reporting `assert (-3.14159, 3.14159) != (-3.14159, 3.14159)` |
| same fixture, **premise removed** | **49 passed** | nothing -- the test measures nothing and reports success |
| loader's MJCF joint fallback range changed to `-1.0, 1.0` | 1 failed | `test_an_undeclared_class_contributes_nothing` only |
| revert the source fix entirely | 18 failed | the PR's whole regression set |

The middle two rows together are the hole: with the fallback-equal fixture the premise fires, and without the premise that same fixture passes silently.

`ruff check` and `ruff format --check` are clean over `strands_robots tests tests_integ`; mypy is 24 errors on both this head and the previous one, with the message sets identical and none in the changed file. The identical 189-file selection (every `tests/simulation/isaac` file plus every guard that walks the test tree or grades sources, docstrings, strings or changelog fragments) is byte-identical on both heads: `9651 passed, 59 skipped`, zero failures on either side.

strands_robots/simulation/isaac/loaders.py
cagataycali: A second base merge has landed since this thread, and it was the same shape you flagged here, so it seemed worth recording that the check was run for the pattern rather than only the instance.

`#2674` merged the accumulator that folds a model's several top-level `<default>` elements into the one root class. This branch renames and re-parameterises that same resolver, so the risk was identical to the `#2672` one: the rename could have dropped the accumulator, and nothing would have said so.

It did not, and the composition turns out to be better than either input. MuJoCo compiles the fixture, so it is the oracle rather than a matter of opinion:

| tree | reported | agrees with MuJoCo |
| --- | --- | --- |
| `upstream/main` | `revolute` `(-3.14159, 3.14159)` | no |
| this branch before the merge | `revolute` `(-0.04, 0.04)` | no |
| the merge | `prismatic` `(-0.04, 0.04)` | **yes** |

Two things worth flagging from checking it:

The joint half of that accumulator was pinned by nothing. Disabling it fired ten tests, **all of them geom** - so a regression that read the root as whichever element came last would have been caught for geoms and not for joints. `test_a_joint_sees_every_top_level_default_element_merged` now states it, with every expected value premise-asserted to differ from the loader's own fallback so it cannot pass vacuously. Setting the fixture to the fallback fires that premise alone.

The rename also broke `#2674`'s new test file in a way the auto-merge could not see: it imported the pre-rename symbol at two sites, so all 399 lines failed at collection with an `ImportError`. Reverting that repair fires its collection; both sites now call the renamed resolver and its 16 tests pass.

Gate re-run at `15d7f6dc` over 305 files: `17 failed, 14375 passed, 64 skipped` byte-identical on both trees, `comm` empty both ways, the 17 all `awsiot`-absent locally. `mypy` 24 == 24 with nothing branch-only. Full break table and the re-measured pre-fix split are in the description.

yinsong1986 (approved):
## Summary

The isaac MJCF loader read a `<joint>`'s `type`, `axis`, `range`, `damping` and `armature` off the element alone, so a class-declared joint (aloha's finger, panda, the UR arms, and 49 other registry assets) silently fell to the loader's own fallbacks under a successful `load_mjcf`. This PR generalises the existing geom-side default-class resolver by element tag (`_mjcf_geom_defaults`/`_geom_attrs` -> `_mjcf_class_defaults`/`_class_attrs`) and routes the five joint attributes through it, so nested classes, `childclass`, `<include>`d fragments, the merged multi-element root class (#2674), the `""`/`"main"` root aliasing (#2672) and element-over-class precedence all come from the one shipped rule rather than a second implementation - with the geom and joint attribute namespaces kept separate so a shared attribute name like `type` cannot leak across kinds. All touched symbols are module-private, so there is no public-API or wire-format surface in the change, and the behaviour change moves strictly toward MuJoCo's compiled model, which the new tests use as the oracle. Both prior review concerns (the pre-#2672 root-class aliasing in the resolver initializer, and the unread `_FALLBACK_LIMIT` CodeQL alert) are resolved on this head.

## What's good

- The new test file uses `MjModel.from_xml_path` as the oracle rather than restating expectations, and premise-asserts every fixture value differs from the loader's fallback so the assertions cannot hold vacuously.
- The AST scan (`TestOneRuleAnswersWhatAJointDeclared`) derives its graded set from the module and pins its own non-vacuity, so a future joint reader is held to the rule on arrival - mirroring the shipped geom-side scan.
- Scope discipline: `<mesh>`/`<material>`/`<site>` defaults, angle units, and `limited` are named out of scope with tests pinning the boundary rather than leaving it implicit.
- Changelog fragment present per AGENTS.md step 3; docstring cross-references updated with the rename.

## Verification

Re-ran locally at `15d7f6dc`: the four changed test files are 66 passed (mujoco 3.8.0, inside the declared `>=3.5.0` range); the full `tests/simulation/isaac` suite shows zero delta against the merge base `bde956f0` (the same 3 environment-dependent failures on both trees); `ruff check` and `ruff format --check` clean on every changed file; no non-ASCII in user-facing strings and no host paths in the new files.

PR #2676, opened 2026-08-24
test(isaac): grade the .msh reader against MuJoCo instead of its own layout assumption

cagataycali:
## Change

Grade the legacy binary mesh (`.msh`) reader in
`strands_robots/simulation/isaac/mesh_assets.py` against MuJoCo, the format's
only owner, instead of against a fixture that shares the reader's own layout
assumption. Test-only: no library file changes.

## Why this pin, now

`_binary_msh` - the helper every existing `.msh` test builds its input with -
writes the four header fields in the order `load_mesh_geometry` reads them and
lays the vertex / normal / texcoord / face blocks out in the order it walks
them. So a round trip through it cannot tell a correct layout from a mirrored
one: a reader that swapped two header fields would be graded by a fixture that
swapped them too, and both halves would move together. The legacy binary mesh
has no published specification beyond MuJoCo's own `LoadMSH`, so there was
nothing on the other side of the comparison.

Two of the three `.msh` fixtures also declared **three** vertices. MuJoCo
refuses `nvertex < 4` (`Error: invalid sizes in MSH file`), so those two graded
the reader against a file the format's owner calls invalid - measured, not
inferred.

This matters for LIBERO specifically: it ships 88 `.msh` assets, and
`mesh_aabb` turns what this reader returns into a mesh-only body's collision
proxy.

## What MuJoCo requires, measured

| fixture | MuJoCo | why it matters here |
| --- | --- | --- |
| 4-vertex tetra, `nnormal=ntexcoord=4` | accepted | the shared happy-path fixture |
| 4-vertex tetra, `nnormal=ntexcoord=0` | accepted | block order: faces sit directly after the vertices |
| 4-vertex tetra, `nnormal=0, ntexcoord=4` | accepted | header field order: normals are 3 floats/vertex, texcoords 2 |
| 3-vertex | refused, `invalid sizes in MSH file` | why every fixture now carries four |
| 4 coplanar vertices | refused, no convex hull | why the comparisons need no extent premise |

MuJoCo does not store the authored frame - it recentres a mesh on the mesh
frame and rotates it onto its principal inertia axes - so the comparison
recovers the file frame as `mesh_pos + R(mesh_quat) @ mesh_vert`, and matches
vertices on distance rather than on sort order (float32 storage leaves ~1e-6,
enough to flip two vertices whose leading coordinate ties).

## The reader is already correct on the real corpus

Every `.msh` asset LIBERO ships, reader vs MuJoCo in the authored frame:

| real `.msh` assets | compared | count mismatches | worst AABB disagreement |
| --- | --- | --- | --- |
| 88 | 88 | 0 | 6.54e-06 |

So this pins correct behaviour rather than fixing a defect. What it removes is
the possibility of that agreement drifting silently.

## Break table

Each row applied to a scratch tree; the parser breaks touch
`mesh_assets.py`, the fixture breaks touch the test file.

| break | fires |
| --- | --- |
| control (no break) | 0 of 33 |
| header fields `nnormal` / `ntexcoord` swapped in the parser | **1** - the asymmetric fixture alone |
| faces read immediately after the vertices | 3, incl. the **pre-existing** `test_binary_msh` |
| texcoords sized as 3 floats instead of 2 | 3, incl. the **pre-existing** `test_binary_msh` |
| shared fixture dropped below MuJoCo's 4-vertex floor | 5, incl. the non-vacuity test (`assert 3 >= 4`) |
| the coplanar fixture made non-coplanar | **1** - the coplanar test alone (`DID NOT RAISE`) |

Two rows fire exactly one test, and two fire a pre-existing test, so the new
pins compose with the contract that was already there rather than replacing it.

A sixth break - asserting the fixture has extent on every axis - showed that
assertion was **unreachable**: MuJoCo runs qhull over the vertices and refuses
a coplanar mesh before any comparison is made. The dead premise was removed and
replaced by `test_mujoco_refuses_a_coplanar_msh`, which states the guarantee the
oracle already provides.

## Verification

Measured at `bd55980`. The head has since moved to `1907aca`, which adds
only the changelog fragment (`git diff bd55980..1907aca -- '*.py'` is empty).

| check | result |
| --- | --- |
| changed file, mujoco 3.12.0 (what CI resolves) | 33 passed, 5 skipped |
| changed file, mujoco 3.5.0 (the declared floor) | 33 passed, 5 skipped |
| new class alone | 6 passed (none skipped) |
| `ruff check` / `ruff format --check`, CI scope | clean |
| `mypy strands_robots tests tests_integ` | 24 errors, byte-identical to `upstream/main` |

Gate: the isaac package plus every guard that walks the test tree or grades
sources, docstrings, strings or changelog fragments - 128 files, the identical
selection on both trees with the changed file excluded from each:

| tree | result |
| --- | --- |
| this branch | 4600 passed, 54 skipped, 0 failed (280.86s) |
| `upstream/main` | 4600 passed, 54 skipped, 0 failed (296.46s) |

Byte-identical, and the branch-only and base-only failing-id sets are both
empty.

## Scope

Only `tests/simulation/isaac/test_mesh_assets.py` changes (`git diff --numstat`
names one path), so there is no library behaviour to film. The 5 skips are the
pre-existing `usd-core`-absent conversion tests.

Deliberately out of scope: the OBJ and STL siblings in the same module can be
graded against the same oracle, but each is a separate contract.

### DIFF
--- a/changelog.d/2676-msh-reader-graded-against-mujoco.md
+++ b/changelog.d/2676-msh-reader-graded-against-mujoco.md
@@ -1,0 +1,30 @@
+### Quality: the .msh reader is graded against MuJoCo, not against its own layout assumption
+
+The legacy binary mesh (`.msh`) tests built every input with a helper that
+writes the four header fields in the order `load_mesh_geometry` reads them and
+lays the vertex, normal, texcoord and face blocks out in the order it walks
+them. A round trip through that helper cannot tell a correct layout from a
+mirrored one: a reader that swapped two header fields would be graded by a
+fixture that swapped them too, and both halves would move together. The format
+has no published specification beyond MuJoCo's own `LoadMSH`, so there was
+nothing on the other side of the comparison.
+
+The reader is now handed the same bytes as MuJoCo and required to agree with it,
+over three block layouts chosen so that a swap is visible: normals and
+texcoords both present, both absent, and texcoords present with normals absent -
+the asymmetric case, where a reader that transposed those two header fields
+computes a different total length for the same file. MuJoCo does not store the
+authored frame, so the comparison recovers it as
+`mesh_pos + R(mesh_quat) @ mesh_vert` and matches vertices on distance rather
+than on sort order.
+
+Two fixtures also declared three vertices, which `LoadMSH` refuses
+(`nvertex < 4`), so they graded the reader against a file the format's owner
+calls invalid; both now use a four-vertex tetrahedron. The refusal is pinned, as
+is MuJoCo's refusal of a coplanar mesh - that second refusal is why the
+comparisons need no separate extent premise.
+
+Every `.msh` asset LIBERO ships already agrees with MuJoCo to 6.54e-06 in the
+authored frame across all 88 files, so this pins correct behaviour rather than
+fixing a defect; what it removes is the possibility of that agreement drifting
+silently. No library behaviour changes.
--- a/tests/simulation/isaac/test_mesh_assets.py
+++ b/tests/simulation/isaac/test_mesh_assets.py
@@ -36,6 +36,14 @@ def _binary_stl(tri_vertices: list[tuple[tuple[float, float, float], ...]]) -> b
     return blob
 
 
+# MuJoCo's ``LoadMSH`` refuses ``nvertex < 4``, so every fixture below carries
+# at least four - a three-vertex .msh is a file the format's owner calls
+# invalid, and a reader graded only against such a file is graded against
+# nothing MuJoCo would ever hand it.
+_MSH_TETRA_VERTS = [(0.0, 0.0, 0.0), (1.0, 0.0, 0.0), (0.0, 2.0, 0.0), (0.0, 0.0, 3.0)]
+_MSH_TETRA_FACES = [(0, 1, 2), (0, 1, 3), (0, 2, 3), (1, 2, 3)]
+
+
 def _binary_msh(
     vertices: list[tuple[float, float, float]],
     faces: list[tuple[int, int, int]],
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Test-only change that re-grades the legacy binary `.msh` reader in `strands_robots/simulation/isaac/mesh_assets.py` against MuJoCo's own `LoadMSH` instead of against `_binary_msh`, a fixture helper that shares the reader's layout assumptions and therefore cannot detect a mirrored header or block order. The new `TestTheMshLayoutIsGradedAgainstMujoco` class hands identical bytes to both readers across three block layouts chosen so a header-field swap is visible, recovers MuJoCo's authored frame via `mesh_pos + R(mesh_quat) @ mesh_vert`, and matches vertices on nearest-neighbour distance to absorb float32 storage error. It also repairs two pre-existing fixtures that declared three vertices - a file MuJoCo itself refuses (`nvertex < 4`) - and pins both that refusal and the coplanar-mesh refusal that makes a separate extent premise unnecessary. No library code changes (`git diff --numstat` against the merge base names exactly the test file and the changelog fragment), so there is no public API, wire format, or persisted schema exposure.

## What's good

- The oracle-based break table in the description is backed up by what the tests actually do: I ran the changed file against mujoco 3.8.0 (a version between the author's tested 3.5.0 floor and 3.12.0 CI resolution) - all 38 tests pass, including all 6 new ones.
- The two `pytest.raises(Exception)` sites carry `# noqa: B017` with the reason (the binding's own error type) and each narrows the match on the message, with both message spellings across the declared mujoco range accounted for.
- Changelog fragment present per AGENTS.md step 3; no host paths, no non-ASCII in the diff; test names describe behaviour rather than provenance; the repo's own guard tests (`test_test_case_names_describe_behaviour`, `test_no_host_paths`) pass on the branch.
- Scope discipline: the OBJ/STL siblings are explicitly deferred rather than bundled in.

## Verification suggestions

- `python -m pytest tests/simulation/isaac/test_mesh_assets.py -q` on any mujoco in the declared `>=3.5.0,<4.0.0` range - the new class's oracle comparisons are version-sensitive by design, so a third data point beyond the author's 3.5.0/3.12.0 pair is cheap confidence (3.8.0 passed here).

PR #2677, opened 2026-08-24
fix(examples): report a refused LIBERO setup step where it refuses

cagataycali:
## The gap

Every `Simulation` / `IsaacSimulation` setup call returns a `{"status": ...}` envelope. The two LIBERO driver setup shims in `examples/libero/run.py` read those envelopes differently: `_setup_isaac` checked all five of its calls with inline `if r.get("status") == "error": raise` guards, and `_setup_mujoco` checked none of its five.

One driver, one merged eval loop, two rules for the same question.

## Measured

Driving `_setup_mujoco` with each step refused in turn, on the real `Simulation`:

| refused step | upstream/main | this branch |
| --- | --- | --- |
| `create_world` | proceeds, prints a `success_rate=` line | refuses, naming `create_world` |
| `add_robot` | proceeds, prints a `success_rate=` line | refuses, naming `add_robot` |
| `load_scene` | proceeds, prints a `success_rate=` line | refuses, naming `load_scene` |
| `start_cameras_recording` | reported as a *rendering* failure, later | refuses, naming `start_cameras_recording` |

The first three are silent: the run continues on a world that was never set up and prints a result line for a rollout whose setup failed. The fourth is louder but misdiagnosed - it surfaces as the zero-frame recording check, so a refused recorder start reads as a camera that produced no frames.

The refusal that is discarded is the one that says *what* went wrong, and it is discarded at the point where that is still knowable.

## Change

`_require_ok(envelope, step)` becomes the single owner of the rule: it returns the envelope untouched on success and raises `RuntimeError("<step> failed: <envelope>")` otherwise. Both shims route every setup call through it - `_setup_mujoco`'s five, and `_setup_isaac`'s five, which previously repeated the check inline.

`_setup_isaac`'s operator-visible wording is unchanged; the Isaac recorder still chains its `on_frame` handle off the returned envelope, which is why the owner returns it rather than swallowing it.

Deliberately left unchecked, and pinned as such: `sim.destroy()` on the cleanup path (a cleanup that refuses must not mask the failure being cleaned up after) and the per-episode `sim.step()` in the retry loop (its verdict is the loop's own, not a setup step's).

## Tests

`tests/test_examples_libero_drivers.py` gains a class that refuses one setup step at a time and asserts the driver names it, a structural rule that both shims route every setup call through the one owner, a consistency rule that a step checked in one shim is checked in the other, and controls for the owner's success behaviour and for the two deliberately-unchecked calls.

Pre-fix: **9 failed / 25 passed**. After: **34 passed**.

Break table - each break fires a distinct set, so no guard is decorative:

| break | fires |
| --- | --- |
| control (this branch) | 0 of 34 |
| exact pre-fix source | 9 - the full regression set |
| `_setup_mujoco` reverted only | 6 - the four behavioural cases, the fallback, and consistency |
| `_setup_isaac` reverted to inline guards | 2 - consistency and single-owner only |
| only the recording arm's check dropped | 2 - that case and consistency |
| the predicate loses its quote normalisation, on pre-fix source | 7 - **consistency and single-owner both go quiet** |
| over-reach: `sim.destroy()` checked too | 1 - the boundary control alone |
| the owner drops the envelope it returns | 1 - the returns-untouched control alone |
| the scan reaches no shim | 2 - non-vacuity and the boundary control |

The sixth row is why the structural rules read the source through `ast.unparse` rather than matching text: with a text predicate they silently stop finding the calls they grade, and a reader cannot tell that from a clean tree.

## Verification

Measured at `70b44215` on one host, mujoco 3.12.0 (what CI resolves) unless stated. The head has since moved to `d930dfdc`, which adds only the changelog fragment (`git diff 70b44215..d930dfdc` is that one file).

- Gate: 151 test files that read `examples/`, walk the test tree, or touch the LIBERO adapter, run identically on this branch and on a pristine `upstream/main` worktree with the changed test file excluded from both: **`2 failed, 4990 passed, 84 skipped`** on each, at 298.58s and 299.03s. The two failures are in `tests/mesh/test_zenoh_transport_security.py` and are a *different* subset on each tree - that file opens real localhost Zenoh sessions on fixed ports, so two parallel runs collide. Run alone it is `9 passed, 2 skipped` on **both** trees. Excluding that one contended file, **0 failing ids on either tree**.
- `mypy strands_robots tests tests_integ`: 24 pre-existing errors on both this branch and a pristine worktree, identical message sets, none in the changed files.
- `ruff check` / `ruff format --check` clean over the CI scope and over the changed example.
- The changed test file: **34 passed** on mujoco 3.12.0 and on 3.5.0, the declared floor.
- **387 passed, 51 skipped** across all 13 `tests/test_examples*.py` files with random ordering enabled.

## Artifact

The same real MuJoCo `Simulation` refusing for two different reasons, and the happy path unchanged. Only the task resolver, the benchmark lookup and `evaluate_benchmark` are stood in - they need the `benchmark-libero` extra and a live policy server. Every refusal shown is the `Simulation`'s own.

the driver reporting a refused setup step

Row 2 is the point: on `upstream/main` a missing scene file and a camera the world does not have produce a **byte-identical** message about zero recorded frames, so the `load_scene` refusal is reported as a camera problem. On this branch each names its own step, and the scene-file case refuses during setup rather than after a rollout. The control row's printed result line and its render are identical on both trees - the render to a mean absolute difference of 0.000001 levels.

### DIFF
--- a/changelog.d/2677-libero-driver-reports-a-refused-setup-step.md
+++ b/changelog.d/2677-libero-driver-reports-a-refused-setup-step.md
@@ -1,0 +1,23 @@
+### Fixed: a refused LIBERO setup step is reported where it refuses
+
+The LIBERO driver's two setup shims read the same `{"status": ...}` envelope
+differently: the Isaac shim checked all five of its `Simulation` calls with
+inline guards, the MuJoCo shim checked none of its five. A refused
+`create_world`, `add_robot` or `load_scene` was therefore discarded, the run
+continued on a world that was never set up, and it printed a `success_rate=`
+line for a rollout whose setup had failed; a refused
+`start_cameras_recording` surfaced later as the zero-frame recording check,
+so the recorder never starting read as a camera that produced no frames.
+Measured on the real `Simulation`, a missing scene file and a camera the
+world does not have produced a byte-identical message, leaving the two
+causes indistinguishable.
+
+`_require_ok(envelope, step)` is now the single owner of the rule - it
+returns the envelope untouched on success and raises
+`RuntimeError("<step> failed: <envelope>")` otherwise - and both shims route
+every setup call through it. The Isaac shim's operator-visible wording is
+unchanged, and it still chains its `on_frame` handle off the returned
+envelope. `sim.destroy()` on the cleanup path and the per-episode
+`sim.step()` in the retry loop stay unchecked on purpose, pinned by controls:
+a cleanup that refuses must not mask the failure being cleaned up after, and
+the step's verdict belongs to the retry loop rather than to setup.
--- a/examples/libero/run.py
+++ b/examples/libero/run.py
@@ -513,6 +513,46 @@ class _EvalPlan:
     eval_kwargs: dict = field(default_factory=dict)
 
 
+def _require_ok(result: dict, what: str) -> dict:
+    """Refuse a setup step the backend declined, where it declined it.
+
+    Every ``Simulation`` / ``IsaacSimulation`` setup call answers a tool
+    envelope rather than raising, so a discarded return leaves a refusal
+    invisible. Both backend shims below assemble the same rollout out of the
+    same envelope-returning calls, and this is the single place either reads
+    the verdict - so they cannot drift back into two rules for it.
+
+    Measured on the ``mujoco`` shim before this existed, isolating one refused
+    step at a time: a refused ``create_world``, ``add_robot`` or ``load_scene``
+    printed a full ``success_rate=`` result line, which
+    ``libero_backend_matrix.py`` parses as a completed cell; and a refused
+    ``start_cameras_recording`` surfaced only after the whole rollout, as
+    ``rollout recorded 0 frames for camera 'image' (0 per-frame render
+    errors)`` - a render diagnosis for a recorder that never started, naming
+    neither the step nor the available-camera list the refusal carried.
+    ``stop_cameras_recording`` answers ``status="success"`` with
+    ``"Was not recording cameras."`` and no json block, so the ``finally``
+    arm gave no signal either.
+
+    Args:
+        result: The tool envelope the setup call answered.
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes the asymmetry between the two LIBERO driver setup shims: `_setup_isaac` checked every `Simulation` envelope inline while `_setup_mujoco` discarded all five of its, so a refused `create_world` / `add_robot` / `load_scene` ran the rollout anyway and printed a `success_rate=` line, and a refused `start_cameras_recording` was misdiagnosed later as a zero-frame camera. The fix introduces `_require_ok(envelope, step)` as the single owner of the status check, routes all ten setup calls in both shims through it (the Isaac shim's operator-visible wording unchanged, still chaining `on_frame` off the returned envelope), and deliberately leaves `sim.destroy()` and the RTX-warmup `sim.step()` unchecked with controls pinning that boundary.

## What's good

- Verified locally at the head SHA: the new test file is 34/34 green, and swapping in the pre-fix `run.py` reproduces exactly the claimed 9 failed / 25 passed - the regression tests genuinely fail on pre-fix code, per AGENTS.md's pin-every-reviewed-fix rule.
- The structural rules read the source via `ast.unparse` with quote normalisation, so a drifted shim or a silently-matching-nothing predicate both fail loudly (the break table's sixth row).
- Scope discipline: changelog fragment in `changelog.d/`, no `CHANGELOG.md` edit, no host paths, ASCII-only added strings, `_require_ok` carries complete `Args:` / `Returns:` / `Raises:` sections, and the tests use `monkeypatch.setitem(sys.modules, ...)` per the restore convention.
- The recording-arm refusal path leaks nothing: `main()` destroys the sim in its `finally`, and both shims keep their `except BaseException` cleanup-and-reraise for setup-time refusals.
- Ruff check and format are clean on both changed Python files.

PR #2678, opened 2026-08-24
fix(sim): name the joints whose rate drive is commanded elsewhere on a velocity write

cagataycali:
## Problem

`set_joint_velocities` writes `qvel` directly. On a joint driven by a `<velocity>`
actuator that `ctrl` **is the joint's own rate**, so the drive is already commanding a
velocity of its own and the next step resolves the disagreement in the drive's favour.
Measured through the public surface on a one-hinge bar, gravity off, `kv=2`:

| drive `ctrl` | written | `qvel` right after the write | angle after 72 steps | report |
| `-3.0` rad/s | `+2.0` | `+2.0000` (as asked) | **`-0.2532` rad** | `Set 1/1 joint velocities` |
| `+2.0` rad/s | `+2.0` | `+2.0000` | **`+0.2391` rad** | `Set 1/1 joint velocities` |

Two opposite outcomes -- the joint turns the other way -- and one byte-identical report.
The write lands, so `status` is `success` and `qvel` really does read back `+2.0`; what the
caller is not told is that something else is commanding that same quantity.

The sibling `set_joint_positions` already names exactly this class for a position servo:

> `1 of them are driven by a position servo still commanded to a different setpoint (arm/j),
> so the next step drives the pose back toward that setpoint; pass hold=True to move the
> setpoint too`

So the shape of the answer was settled one method over. Only the velocity half had no
equivalent, and `hold=True` is not the answer here -- a rate drive's own `ctrl` is the rate,
so the way to hold a velocity is to command the drive.

## Change

`scene_ops.joint_rate_drive_map` is the velocity counterpart of the existing
`joint_drive_map`. Its terms mirror that classifier and are read off the same compiled
fields: an affine bias, no position feedback but negative velocity feedback
(`biasprm = [0, 0, -kv]`), a stateless `dyntype`, and a joint transmission resolved through
`actuator_joint_id`. `set_joint_velocities` consults it and names the joints whose drive is
commanding a different rate, pointing at `send_action`, which writes that `ctrl`.

Measured against every MuJoCo actuator shortcut, `<velocity>` is the only one that clears
all of the terms:

| shortcut | `biastype` | `biasprm` | `dyntype` | `ctrl` is a joint rate |
| `<velocity kv=2>` | affine | `[0, 0, -2]` | none | **yes** |
| `<position kp=10>` | affine | `[0, -10, 0]` | none | no (a pose) |
| `<motor>` | none | `[0, 0, 0]` | none | no (a torque) |
| `<intvelocity>` | affine | `[0, -1, 0]` | integrator | no (integrated to a pose) |
| `<damper kv=2>` | none | `[0, 0, 0]` | none | no (`ctrl` scales damping) |
| `<cylinder>` | none | `[0, 0, 0]` | filter | no |
| `<adhesion>` / `<general>` | none | `[0, 0, 0]` | none | no |

A `<general>` spelling `<velocity>` longhand (`biastype="affine" biasprm="0 0 -2"`)
classifies identically, which is the point of reading the compiled fields rather than the
element. Slot 2 has to be **negative** rather than merely non-zero: positive velocity
feedback is anti-damping, and measured, a written rate against it diverges (`mj_step`
reports `Nan, Inf or huge value in QACC`) instead of settling on any commanded value.

A joint driven by a torque motor, a position servo, a stateful drive or a tendon is not
reported: its `ctrl` is not a velocity in the joint's units, so there is no rate for the
write to disagree with.

## Scope

Report only -- no verdict, no accepted input and no written value moves, and a drive that
happens to be commanding the written rate keeps the plain report byte-for-byte. A `hold=`
parameter for `set_joint_velocities` (write the drive's `ctrl` as well as `qvel`, the way
`set_joint_positions(hold=True)` does) would be a new behaviour on a public surface and is
deliberately not part of this; `TestTheReportIsNotWidened` pins the current shape either way.

## Tests

`tests/simulation/mujoco/test_joint_velocity_write_names_a_conflicting_rate_drive.py`,
19 cases. Against `upstream/main`: **6 failed / 13 passed**, of which **4 are behavioural** --

- `AssertionError: Set 1/1 joint velocities` (the report does not name `arm/j_vel`)
- `AssertionError: a drive commanding the written rate and one commanding another rate are opposite outcomes and must not share one report: 'Set 1/1 joint velocities'`
- `j_vel named in the report: True expected, False measured -- 'Set 1/1 joint velocities'`
- `j_gen_vel named in the report: True expected, False measured -- 'Set 1/1 joint velocities'`

-- and 2 report only that `joint_rate_drive_map` does not exist yet. All four
`TestTheReportIsNotWidened` cases and every non-rate drive in the parametrized sweep pass on
both trees.

### Break table

| break | fires |
| --- | --- |
| control | 0 of 19 |
| exact pre-fix source | 6 (4 behavioural + 2 symbol) |
| the comparison dropped, or keyed on `0.0` rather than the written rate | 2 (the two-outcomes test + the plain-report control) |
| slot 2 merely non-zero rather than negative | 2 (the anti-damping fixture + the classifier count) |
| transmission target read raw rather than through `actuator_joint_id` | **1** (the tendon case alone) |
| the joint is named but the remedy omitted | **1** (the remedy assertion alone) |
| `_RATE_TAG` pointed at the pose servo | 3 (the premise names the cause + 2 behavioural) |
| the same, with the premise removed | 2, and **three of the five behavioural cases pass** while driving a drive whose `ctrl` is not a rate |
| `_RATE_TAG` pointed at a tag the fixture does not declare | 6 (the premise names the drives it does declare) |

The behavioural cases read the rate-drive tag rather than spelling it out, so the joint they
measure the physical outcome on and the tag the sweep expects the report to name cannot drift
apart; the last three rows are that premise. Its own value is that the drift is otherwise
*quiet*: with the premise removed and the tag pointed at the pose servo, three of the five
behavioural cases still pass, and the two that fail say only `Set 1/1 joint velocities` -- a
report that is correctly silent, which reads as the report having regressed.

Given the shortcut corpus above, `biasprm[2] < 0` is the term that discriminates on its own;
the affine-bias and stateless-`dyntype` terms are conjunctive and matter for a hand-written
`<general>`, which is why the longhand fixtures exist.

## Artifact

One program, run against both trees; only `strands_robots` differs. The two frame strips are
byte-identical across the trees (`np.array_equal` is `True` for the conflicting rollout) --
this changes what is reported, not what is simulated -- so the picture is the consequence the
report now names. Rendered headless on MuJoCo 3.12.0 (`MUJOCO_GL=egl`), gravity off so the
motion is the drive alone, `mj_step` reporting no bad-qacc warning in either rollout.

velocity write on a rate-driven joint

## Verification

Measured at `4f47eba2`. The head has since moved to `0b070a05`, which adds the changelog
fragment and, in the test file alone, the rate-drive tag read plus its premise case:
`git diff 4f47eba2..0b070a05 -- 'strands_robots/'` is **empty**, so the source under test is
byte-identical to the tree the blast-radius and `mypy` runs below were measured on.

- Blast radius, the identical 374-file selection on both trees with the new test file excluded
  from each: **`44 failed, 10496 passed, 243 skipped, 126 errors` byte-identical** (649.82s vs
  649.86s), **170 failing ids on each tree with `comm` empty in both directions**. Every one is
  a locally absent optional dependency -- 326 `No module named 'device_connect_edge'`, declared
  as `device-connect-edge>=0.2.0` in the `device-connect` extra which `all` pulls in, plus
  `qpsolvers` from `sim-mujoco` and one numpy artifact.
- The new file: **19 passed** on MuJoCo **3.12.0** and on **3.5.0**, the declared floor.
- `ruff check` / `ruff format --check` clean; `mypy strands_robots tests tests_integ`
  **24 == 24** against a pristine worktree, none in the changed files.

## 13. Review rounds

**Round 1** -- `0b070a05`

- Static analysis flagged `_RATE_TAG` as an unused global. It was a leftover, and deleting it
  would have missed why it was there: the behavioural classes spelled `a_vel` / `arm/j_vel` out
  at each call site while the parametrized sweep graded against `_RATE_TAGS`, so nothing tied
  the joint whose physical outcome is measured to the tag the sweep expects the report to name.
  The behavioural cases now read the constant (`_RATE_ACTUATOR`, `_RATE_JOINT`) and
  `TestTheBehaviouralFixtureIsARateDrive` pins the tie. The drift it prevents is otherwise
  quiet: with the tag pointed at the pose servo, three of the five behavioural cases still pass
  and the two that fail report only a correctly-silent `Set 1/1 joint velocities`, which reads
  as the report having regressed rather than as the fixture no longer driving a rate.
  18 cases -> 19, and three rows added to the break table.
- `Verification` re-stated against the new head: `strands_robots/` is unchanged since the
  measured commit, so the blast radius still holds as reported; the case count is corrected to
  19, re-run on MuJoCo 3.12.0 and on the 3.5.0 floor; `ruff check`, `ruff format --check` and
  `mypy` are clean on the changed test file.

### DIFF
--- a/changelog.d/2678-velocity-write-names-rate-drive.md
+++ b/changelog.d/2678-velocity-write-names-rate-drive.md
@@ -1,0 +1,41 @@
+### Fixed
+
+`set_joint_velocities` now names the joints whose actuator is commanding a
+different rate. The write lands and `qvel` reads back what was asked for, so
+`status` was `success` and the report was the plain `Set 1/1 joint velocities` --
+but on a joint driven by a `<velocity>` actuator that `ctrl` **is** the joint's
+own rate, so the drive was already commanding a velocity of its own and the next
+step resolved the disagreement in the drive's favour. Measured on a one-hinge bar
+with gravity off and `kv=2`: writing `+2.0` rad/s while the drive commanded
+`-3.0` left the joint at `-0.2532` rad after 72 steps, and writing the same
+`+2.0` while the drive commanded `+2.0` left it at `+0.2391`. Two opposite
+outcomes, and one byte-identical report.
+
+The sibling `set_joint_positions` had already settled the shape of that answer
+for a position servo, naming the joint and the setpoint that overrides it. Only
+the velocity half had no equivalent, and its remedy is not the same one:
+`hold=True` moves a servo's setpoint, while a rate drive's `ctrl` already is the
+rate, so the report points at `send_action`.
+
+`scene_ops.joint_rate_drive_map` is the velocity counterpart of the existing
+`joint_drive_map`, reading the same compiled fields: an affine bias with no
+position feedback but negative velocity feedback (`biasprm = [0, 0, -kv]`), a
+stateless `dyntype`, and a joint transmission resolved through
+`actuator_joint_id`. Measured against every MuJoCo actuator shortcut, `<velocity>`
+is the only one that clears all of them -- a torque motor, a position servo,
+`<intvelocity>` (whose `ctrl` is integrated into a pose), `<damper>` (whose `ctrl`
+scales damping rather than commanding a rate) and `<cylinder>` are all excluded,
+and a `<general>` spelling `<velocity>` longhand classifies identically, which is
+the point of reading the compiled fields rather than the element. Slot 2 has to
+be negative rather than merely non-zero: positive velocity feedback is
+anti-damping, and a written rate against it diverges rather than settling on any
+commanded value.
+
+A tendon rate drive is not reported. Its `ctrl` moves a tendon length rate, not
+the joint's, so there is no rate in the joint's units for the write to disagree
+with -- which is why the transmission is resolved through `actuator_joint_id`
+rather than read raw.
+
+Report only: no verdict, no accepted input and no written value moves, and a
+drive that happens to be commanding the written rate keeps the plain report
+byte-for-byte.
--- a/strands_robots/simulation/mujoco/physics.py
+++ b/strands_robots/simulation/mujoco/physics.py
@@ -33,6 +33,7 @@ from strands_robots.simulation.mujoco.backend import (
 from strands_robots.simulation.mujoco.scene_ops import (
     fromto_fixed_size_components,
     joint_drive_map,
+    joint_rate_drive_map,
     persist_body_mass,
     persist_geom_properties,
     refresh_body_inertial_from_geometry,
@@ -1744,6 +1745,20 @@ class PhysicsMixin:
         resolved inside ``robot_name``'s namespace before the cross-robot
         fallback, so a bare name reaches the robot the caller addressed, and a
         ``robot_name`` no robot carries is refused in both call forms.
+
+        The write is a state write, so on a joint driven by a *rate* actuator --
+        one whose ``ctrl`` is the joint's own velocity, per
+        :func:`~strands_robots.simulation.mujoco.scene_ops.joint_rate_drive_map`
+        -- it survives exactly as long as nothing steps: the drive is still
+        commanding whatever rate it held, and the next step resolves that
+        disagreement in the drive's favour. The success report therefore names
+        the written joints whose rate drive is commanded elsewhere, on the same
+        terms :meth:`set_joint_positions` reports a position servo holding a
+        different setpoint, and points at ``send_action`` -- which writes that
+        ``ctrl`` -- as the way to command the rate rather than only the state. A
+        joint driven by a torque motor, a stateful drive or a tendon is not
+        reported: its ``ctrl`` is not a velocity in the joint's units, so there
+        is no rate for it to disagree with.
         """
         if self._world is None or self._world._model is None or self._world._data is None:
             return {"status": "error", "content": [{"text": _NO_WORLD_MSG}]}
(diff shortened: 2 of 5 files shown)


yinsong1986 (dismissed):
## Summary

This PR closes the velocity half of a reporting gap `set_joint_positions` already handles for position servos: `set_joint_velocities` writes `qvel` directly, and on a joint driven by a `<velocity>` actuator the drive's own `ctrl` is the joint's rate, so the next step overrides the write while the tool reports plain success. The change adds `scene_ops.joint_rate_drive_map` -- a classifier mirroring the existing `joint_drive_map`, read off the same compiled fields (`biastype`/`biasprm`/`dyntype`, transmission resolved through `actuator_joint_id` so tendon drives are excluded) -- and has the success report name the written joints whose rate drive is commanded elsewhere, pointing at `send_action` as the remedy. Report-only: the plain report stays byte-identical when no drive disagrees, pinned by `TestTheReportIsNotWidened`.

## What's good

- The classifier scan and the `data.ctrl` reads sit inside `self._lock`, honouring the mjData thread-safety contract, and every MuJoCo enum comparison is value-based (`int()` on both sides) per AGENTS.md.
- The discriminator is measured against the full actuator-shortcut corpus, including the `<general>` longhand and the anti-damping (positive `biasprm[2]`) counterexample, with a parametrized sweep asserting every non-rate drive stays unreported.
- Conventions all in place: changelog fragment, `test_public_member_docstrings.py` entry, ASCII-only tool-result strings, behaviour-named test classes, no host paths.
- Verified locally: the 18-case test file passes on MuJoCo 3.8.0 (inside the declared `>=3.5.0` range), and `ruff check` / `ruff format --check` are clean on the changed files.

yinsong1986 (approved):
## Summary

This PR closes the velocity half of a reporting gap that `set_joint_positions` already handles for position servos: `set_joint_velocities` writes `qvel` directly, and on a joint driven by a `<velocity>` actuator the drive's own `ctrl` is the joint's rate, so the next step overrides the write while the tool reports plain success. The change adds `scene_ops.joint_rate_drive_map` -- a classifier mirroring the existing `joint_drive_map`, read off the same compiled fields (`biastype`, `biasprm`, `dyntype`, transmission via `actuator_joint_id`) -- and has the velocity write's success report name the written joints whose rate drive is commanded elsewhere, pointing at `send_action` as the way to command the rate. It is report-only: no verdict, accepted input or written value moves, no public signature changes, and a drive agreeing with the written rate keeps the plain report byte-for-byte, so there is no one-way door here.

## What's good

- Enum comparisons are value-based (`int()` on both sides) per AGENTS.md > MuJoCo enums, and the transmission is resolved through `actuator_joint_id` rather than read raw, which the tendon fixture pins.
- The classifier scan and the `data.ctrl` read both sit inside `self._lock`, alongside the `qvel` write they qualify.
- The classifier's exclusions are argued from the compiled fields with a fixture per shortcut, including the `<general>` longhand and the anti-damping (`biasprm[2] > 0`) counterexample, and the plain-report control pins that the report is not widened.
- Changelog fragment named for the PR, docstring registered in `test_public_member_docstrings.py`, all report strings ASCII, test names describe behaviour. Verified locally at `0b070a05`: the new file is 19/19 green (MuJoCo 3.8.0, `MUJOCO_GL=egl`), `ruff check` / `ruff format --check` clean on the changed files, and the docstring/naming/ASCII convention suites pass.
- The prior CodeQL finding (`_RATE_TAG` unused at `33f8bcf4`) is resolved at this head: the behavioural classes now read the constant and `TestTheBehaviouralFixtureIsARateDrive` pins the tie between the measured joint and the tag the sweep grades.

PR #2646, opened 2026-08-23
docs(notebooks): notebook 7 section 4 runs the transform surface on CPU via the mock provider

yinsong1986:
Closes #2644

> **Stacked on #2508** (draft until it merges). The notebook this PR edits only exists on #2508's branch, and #2644 is explicitly "blocked on #2508 merging (it owns the current section 4 text)" - so this branch is based on #2508's head and the diff currently shows both changes. Once #2508 squash-merges, this branch will be rebased so the diff collapses to this PR's own delta (the last commit, `+189/-32` across 3 files), and the draft flag comes off. Review of the delta alone is possible now via the head commit.

## What changed

Section 4 of `examples/notebooks/07_data_augmentation.ipynb` was markdown-only because the generative `cosmos_transfer` backend needs a GPU. Per cagataycali's review call on #2508 (filed as #2644), it is now a runnable CPU-only demonstration of the same `strands_robots.transforms` surface via the `mock` provider:

- **Runnable transform**: `create_transform("mock")` + `TransformSpec(source_root=ROOT, seed=SEED, variants_per_episode=3)` against the `local/nb7_demo` dataset section 1 records - `validate()` preflight, then `transform()`. One recorded episode becomes three provenance-marked variants.
- **Provenance round trip**: the transform writes `meta/provenance.json` via `write_provenance`; the notebook reads it back with `load_provenance` (every record `synthetic=true`, naming the source episode, transform name/version, variant and derived seed) and `synthetic_episode_indices` - including the empty set on the recorded source dataset, which is the "absence means recorded" contract.
- **Pixels measurably change**: the mock's deterministic per-variant brightness shift is measured against the source frame (validated run: variant frame-0 means 142.5 / 109.6 / 59.2 vs source 115.8; mean shift on variant 0 was +26.8) and rendered side by side.
- **Intro + framing**: the mechanism table gains row 4 (post-hoc dataset transforms), "three mechanisms" becomes four, and the "markdown-only" framing is replaced by "runs on CPU via the mock provider; `cosmos_transfer` needs a GPU and stays in prose".
- **README row**: `examples/notebooks/README.md` row 7 updated to match.
- Changelog fragment `changelog.d/2644-nb7-runnable-mock-transform.md`.

## What deliberately did not change

- `cosmos_transfer` stays out of the code cells (GPU-only); the prose keeps pointing at `docs/data/transforms.md`, including the re-validation gate and its pixel-blind-verdict honesty rule.
[...]

PR #2679, opened 2026-08-24
ci: refuse a fix derived on a checkout the branch has moved past

cagataycali:
## Problem

`check_pr_head_is_current.py` opens by saying a pull request has **two** answers to
"what is the head commit". It has three, and the count is not pedantry -- the third is the
one every checkout reaches for and the only one carrying no signal at all.

| answer | what moves it out of date |
| --- | --- |
| `refs/pull/N/head` | a mirror ref GitHub refreshes on its own schedule, so it trails a push to the fork branch |
| `pullRequest { headRefOid }` | normally reconciled within a second of a push; #2508 sat five hours behind |
| tip of the head repository's branch | nothing -- this is the value that exists, and what the other two are *for* |

Measured on #2678. A run checked the pull request out the obvious way and got the mirror,
while the API was already serving the commit that existed:

```
git fetch origin pull/2678/head   ->  33f8bcf4   one commit behind
pullRequest { headRefOid }        ->  0b070a05   the tip
git merge-base --is-ancestor 33f8bcf4 0b070a05   ->  true, a pure lag
```

The cost is not a wasted fetch. The run believed it *had* checked out the branch, grepped
the symbol its review thread named, found it genuinely unused **on that tree**, and derived
a correct fix for source that no longer existed. What surfaced the drift was `git push`
being refused as non-fast-forward -- after the work. A fix touching a different file would
have pushed cleanly.

### And the answer was worse than a duplicate

This is the part that makes re-reading the thread non-optional rather than tidy. The thread
was a CodeQL *unused global `_RATE_TAG`*.

| derived against | what the finding means | fix | suite |
| --- | --- | --- | --- |
| the stale mirror | dead code | delete the constant | 18 passed, `ruff` clean |
| the actual tip | a **missing call site** | wire the constant up, pin the tie | 19 passed |

The constant was unused because the behavioural tests spelled its value out at each call
site, so nothing tied the joint under measurement to the tag the parametrized sweep graded.
Deletion would have **passed CI while removing the invariant the symbol existed to carry**.
An unused symbol in a test file is often a missing call site rather than dead code, and the
two fixes are indistinguishable by test outcome.

#2520 records four instances of this class -- #2511 (one thread, four author replies), #2566
and #2577 (two runs deriving one fix), and #2678 above. In three of the four the only thing
that prevented a duplicate landing was the non-fast-forward rejection, which is load-bearing
purely by accident: it fires because the mandated push shape is a plain `git push`, and both
nearby shapes defeat it -- `pull --rebase` lands an empty-diff commit on top, and
`--amend --force-with-lease` deletes the commit that already answered the thread.

## Change

`scripts/check_checkout_is_pr_head.py` compares the commit a local clone is on against the
tip of the branch in the head repository, and exits 1 when the branch has commits the clone
does not, so it can sit in front of the work in a `&&` chain.

The tip is resolved from the head repository's own ref, for the same reason its sibling does
it: the value under suspicion cannot be its own witness. Here **both** cheap answers are
suspect -- the mirror produced the #2678 lag, `headRefOid` produced the #2508 one.

### Ancestry, not equality

The comparison cannot be equality, and this is the design point rather than a detail. A run
that has committed its own work sits at a commit the tip does not have yet -- the ordinary
state between a commit and its push -- so an equality check would fire **exactly when the
tool is being used correctly**, and a check that cries wolf gets ignored.

| checkout vs tip | outcome | finding |
| --- | --- | --- |
| equal | `current` | no |
| tip is an ancestor of the checkout | `ahead` (own unpushed work) | no |
| otherwise | `stale-checkout` | **yes** |
| tip unreadable (deleted fork/branch) | `unresolvable-head` | no |
| local `HEAD` unreadable | `unknown-checkout` | no |

A tip **absent from the local object database** is `stale-checkout`, not indeterminate: a
clone that never fetched a commit cannot contain it. `tip_is_ancestor_of_checkout` still
distinguishes that (`None`) from a genuine divergence (`False`) so the two causes stay
legible, but both are stale. Folding an unreadable reading into a pass is the one failure
mode that would turn this into a check agreeing with everything, which is why
`unresolvable-head` is its own outcome too.

`AGENTS.md` is corrected in the same diff -- the "two answers" sentence, plus a sibling
bullet for the third answer, the `FETCH_HEAD == headRefOid` assertion, and the
missing-call-site lesson. The doc's own framing is part of why the mistake feels safe.

## Scope

Additive. No existing script, test or workflow changes behaviour; `classify()` in the
sibling check is untouched and its 23 pinned tests pass unmodified. No workflow wires this
in, matching both existing agent-invoked checks (`check_pr_head_is_current.py` and
`check_thread_is_answered.py` have none): a checkout is a property of one clone, so there is
no `--all-open` population to sweep and nothing for a `pull_request` event to fire on.

It reports and does not gate. The remedy is to re-fetch and re-derive, which is not
something a branch clears by pushing.

## Tests

`tests/test_checkout_is_pr_head.py`, **35 cases**, using the two #2678 commits verbatim as
fixtures -- the pattern `test_pr_head_is_current.py` uses for #2508. The git-touching helper
is pinned against a two-commit repository built in `tmp_path`, deliberately not this
repository's history: the #2678 commits live on a fork branch that has since been deleted,
so a test reading them from the object database would pass or fail on what the clone
happened to fetch.

### Break table

| break | fires |
| --- | --- |
| control | 0 of 35 |
| ancestry dropped, equality only (`ahead` branch removed) | **4** |
| absent tip object folded into a pass (`None` -> `ahead`) | 1 |
| unreadable tip folded into `current` | 2 |
| remedy re-fetches the mirror instead of the branch by name | 1 |
| `evaluate` consults ancestry even when checkout == tip | 1 |
| tip taken from the record (`headRefOid`) instead of the fork ref | 2 |
| finding summary drops the "means something else at the tip" clause | 1 |
| sibling `headRefOid` note dropped | 1 |

The first row is the one worth naming: it is the mutation a reviewer would most plausibly
suggest as a simplification, and it breaks the tool for every mid-cycle run.

## Verification

Measured at `9830553`.

- **Live, on an open pull request** (#2646, `yinsong1986/robots`), the two directions:

| `--checkout` | outcome | exit |
| --- | --- | --- |
| `b036c956` (the tip) | `current` | 0 |
| `e07967f1` (its parent) | `stale-checkout` | **1**, with the `::warning` annotation |

- **Dogfooded on this pull request.** From the clone this branch was written in, with no
  argument but `--pr 2679`, the check resolved `cagataycali/robots`
  `ci/refuse-a-fix-derived-on-a-stale-checkout` and reported `current` at `98305533`, exit 0.
- **Settled-state control:** all three other open pull requests' `refs/pull/N/head` mirrors
  agreed with their `headRefOid` (`b036c956`, `741f4057`, `8d6a4c42`), so agreement is the
  ordinary case and the lag is the anomaly -- the check is silent on the steady state.
- **The lag did not reproduce on demand, which is the argument for a check rather than a
  habit.** Sampling `refs/pull/2679/head` against `headRefOid` every 3s for 15s from this
  pull request's creation, the mirror was already correct at the first sample (`98305533`,
  5/5 agreeing). A window that is usually absent and occasionally seconds-to-hours wide is
  one you cannot time your way around, and #2508 is the same phenomenon on the other API
  answer lasting over five hours. So the finding is rare, silent, and cheap to test for --
  which is why this exits 1 in front of the work rather than asking anyone to remember.
- `tests/test_checkout_is_pr_head.py`: **35 passed**.
- `tests/test_pr_head_is_current.py`: **23 passed**, unmodified.
- Convention sweeps that grade the whole tree, including the new files:
  `test_log_strings_are_ascii`, `test_except_tuples_state_their_real_scope`,
  `test_graphql_node_id_targeting`, `test_composition_compare_ref_form`,
  `test_composition_run_precondition`, `test_codeql_query_filters`,
  `test_all_exports_are_statically_defined` -- **139 passed, 1 failed**, the failure being
  `ModuleNotFoundError: No module named 'strands'`, reproduced identically on a stashed
  worktree, so it is a locally absent dependency and not this branch.
- `scripts/assemble_changelog.py --check`: `changelog fragments OK`.
  `tests/test_changelog_fragments.py` + `test_changelog_format.py`: 30 passed.
- `ruff check` / `ruff format --check` clean; `mypy` clean on both new files. File mode 644,
  matching the sibling scripts.

One correction to the record while measuring: my own note on #2520 said the mirror was "two
commits" behind. `git rev-list --count 33f8bcf4..0b070a05` is **1**. The oids and the
ancestry hold; the distance did not, and the PR says 1.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -775,10 +775,12 @@ hatch run format            # ruff check --fix, ruff format
      `require_last_push_approval` then disqualifies the pushing account from
      re-supplying it, turning a one-approval merge into one that needs a second
      reviewer.
-   - *And that the head it names is the branch's tip.* A pull request has two
-     answers to "what is the head commit" and they can disagree for hours.
-     `headRefOid` is the value the pull request *records*; the tip of the branch
-     in the head repository is the value that exists. GitHub normally reconciles
+   - *And that the head it names is the branch's tip.* A pull request has three
+     answers to "what is the head commit" and they can disagree for hours. Two
+     of them are the API's, and are the pair this bullet compares: `headRefOid`
+     is the value the pull request *records*; the tip of the branch in the head
+     repository is the value that exists. The third is the commit your clone is
+     actually on, which is the next bullet. GitHub normally reconciles
      them within a second of a push, and when it does not, nothing on the pull
      request says so - because every gate this step tells you to poll resolves
      the head *through the pull request's own view of it*, so they all agree
@@ -862,6 +864,65 @@ hatch run format            # ruff check --fix, ruff format
 
      It agreed with `git ls-remote` on all 10 open pull requests, so it needs no
      clone. Pinned by tests/test_pr_head_is_current.py. See #2538.
+   - *And that the tree you are deriving from is that tip.* The third answer is
+     `refs/pull/N/head`, and it is the one every checkout reaches for and the
+     only one with no signal at all. It is a mirror ref GitHub refreshes on its
+     own schedule, so it trails a push to the fork branch. Measured on #2678:
+
+     ```
+     git fetch origin pull/2678/head   ->  33f8bcf4   one commit behind
+     pullRequest { headRefOid }        ->  0b070a05   the tip
+     git merge-base --is-ancestor 33f8bcf4 0b070a05   ->  true, a pure lag
+     ```
+
+     What that costs is not a wasted fetch. The run believed it *had* checked out
+     the branch, grepped the symbol its review thread named, found it genuinely
+     unused on that tree, and derived a correct fix for source that no longer
+     existed. Only `git push` refusing as non-fast-forward surfaced the drift -
+     after the work. A fix touching a different file would have pushed cleanly.
+
+     And the answer was worse than a duplicate, which is why re-reading the
+     thread against the tip is not optional. The thread was a CodeQL "unused
+     global `_RATE_TAG`". Against the stale tree the fix is deletion and the
+     suite agrees - 18 passed, `ruff` clean. Against the tip the same finding
+     means the opposite: the constant was unused because the behavioural tests
+     spelled its value out at each call site, so nothing tied the joint under
+     measurement to the tag the parametrized sweep graded, and the fix is to
+     wire it up. Deletion would have passed CI while removing the invariant the
+     symbol carried. **An unused symbol in a test file is often a missing call
+     site rather than dead code, and the two fixes are indistinguishable by test
+     outcome.**
+
+     So resolve the head from the API and fetch the branch *by name*. Never
+     `refs/pull/N/head`, and re-fetching that ref can return the same stale
+     commit again:
+
+     ```
+     git fetch https://github.com/$HEAD_REPO.git $HEAD_REF
+     git rev-parse FETCH_HEAD      # must equal headRefOid from the API
+     ```
+
+     That assertion is the cheap part and catches the whole class: if
+     `FETCH_HEAD != headRefOid`, no fix derived on the local tree is
+     trustworthy. Or run the check, which resolves the head repository's ref
+     itself and exits 1 on a stale tree, so it can sit in front of the work:
+
+     ```
+     python3 scripts/check_checkout_is_pr_head.py --repo <owner/name> --pr <N>
+     ```
+
+     It compares by **ancestry, not equality**: a clone sitting at its own
+     unpushed commit contains the tip and reads `ahead`, which is the ordinary
+     state between a commit and its push and is not a finding. A tip missing
+     from the local object database is `stale-checkout` rather than
+     indeterminate - a clone that never fetched a commit cannot contain it.
+     Pinned by tests/test_checkout_is_pr_head.py. See #2520, which records four
+     instances: #2511 (one thread, four author replies), #2566 and #2577 (two
+     runs deriving one fix, the duplicate discarded only because a plain push
+     was refused) and #2678 above. That refusal is load-bearing by accident -
+     `git pull --rebase` then push lands an empty-diff commit on top, and
+     `--amend --force-with-lease` deletes the commit that already answered the
+     thread.
    And before merging, `reviewDecision: APPROVED` alone is not the gate: poll
    the **required** contexts' own conclusions and `mergeStateStatus == CLEAN`
    together, since `reviewDecision` flips before the checks finish.
--- a/changelog.d/2520-checkout-is-pr-head.md
+++ b/changelog.d/2520-checkout-is-pr-head.md
@@ -1,0 +1,37 @@
+### Added: a check for a checkout the pull request's branch has moved past
+
+`scripts/check_checkout_is_pr_head.py` compares the commit a local clone is on
+against the tip of the branch in the head repository, and exits 1 when the branch
+has commits the clone does not, so it can sit in front of the work.
+
+A pull request has three answers to "what is the head commit", not the two
+`check_pr_head_is_current.py` compares. The third is `refs/pull/N/head`, which is
+what a checkout reaches for and the only one carrying no signal: it is a mirror
+ref GitHub refreshes on its own schedule, so it trails a push to the fork
+branch. Measured on #2678, where the mirror served `33f8bcf4` while the API
+already served the tip `0b070a05`, one commit further on and with the mirror as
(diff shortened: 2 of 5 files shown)


yinsong1986 (dismissed):
## Summary

Adds `scripts/check_checkout_is_pr_head.py`, an advisory (report-not-gate) check that compares a local clone's checkout against the tip of a pull request's branch in the head repository and exits 1 when the branch has commits the clone lacks — closing the third head-answer gap (`refs/pull/N/head` mirror lag) that #2678 measured, alongside the two `check_pr_head_is_current.py` already compares. The design centre is sound and pinned: the comparison is ancestry rather than equality so a mid-cycle clone at its own unpushed commit reads `ahead` (not a finding), a tip absent from the local object database reads `stale-checkout` rather than indeterminate, and an unreadable tip or `HEAD` gets its own non-finding outcome instead of folding into a pass. `AGENTS.md` and a changelog fragment are corrected/added in the same diff.

## What's good

- 35 tests using the #2678 commits verbatim as fixtures, with the git-touching helper pinned against a hermetic two-commit repo in `tmp_path` rather than this repository's history; the sibling's 23 tests pass unmodified.
- House conventions followed precisely: exception tuples, fail-open-on-missing-input shape, exit semantics, stderr wording, and remedy rendering all match the six sibling `scripts/check_*.py` checks; `WHAT_CLEARS_THIS` is named once so the report and docstring cannot drift.
- Scope discipline: purely additive, no workflow wiring (consistent with the other agent-invoked checks), `classify()` in the sibling untouched.
- Verified locally at `9830553`: new suite 35 passed, `ruff check`/`ruff format --check` clean, ASCII sweep clean, fragment naming matches `changelog.d/README.md` (issue number is explicitly allowed).

## Verification suggestions

- Live spot-check both directions on any open PR: `python3 scripts/check_checkout_is_pr_head.py --repo strands-labs/robots --pr <N> --checkout <tip>` (expect `current`, exit 0) and `--checkout <tip's parent>` (expect `stale-checkout`, exit 1 with the `::warning` annotation).
- Confirm the squash tree matches the head tree per AGENTS.md step 8 (`behind_by` was 0 at review time).

yinsong1986 (approved):
## Summary

Adds `scripts/check_checkout_is_pr_head.py`, an advisory check that compares the commit a local clone is on against the tip of the pull request's branch in the head repository, exiting 1 only when the branch carries commits the clone does not (`stale-checkout`) -- closing the third head-answer gap (`refs/pull/N/head` mirror lag, measured on #2678) that neither of the two answers `check_pr_head_is_current.py` compares can see. The design centre holds up under scrutiny: ancestry rather than equality keeps a mid-cycle clone at its own unpushed commit out of the finding (`ahead`), a tip absent from the local object database is correctly stale rather than indeterminate, and unreadable tip/`HEAD` each get their own non-finding outcome instead of folding into a pass. `AGENTS.md` is corrected in the same diff, a changelog fragment is present, and `tests/test_duplicate_claim_check.py` gains the `_NAMES_REPOSITORY` entry the new script's `GITHUB_REPOSITORY` inference requires (the second commit).

## What's good

- Verified at `f2c0fef7` on a clean worktree: the new 35-test suite passes, the sibling's 23 pass unmodified, and the tree-wide convention sweeps (`test_except_tuples_state_their_real_scope`, `test_log_strings_are_ascii`, `test_changelog_fragments`) pass with the new files in scope. `ruff check` clean; both new files are ASCII-clean.
- House conventions followed exactly: fail-open on missing `--repo`/`--pr`/`--token` and on an unreachable API (matching all six sibling checks), fixed-argv `subprocess` with no shell, GraphQL inputs passed as variables rather than interpolated, token only in the request header, `WHAT_CLEARS_THIS` named once so the remedy cannot drift between report and docstring.
- The break table's first row is the right thing to have pinned: the equality-only simplification a reviewer would plausibly suggest is the one mutation that breaks the tool for every mid-cycle run, and four tests fire on it.
- Scope discipline: purely additive, no workflow wiring (consistent with the other agent-invoked checks), sibling `classify()` untouched.

## Verification suggestions

- Live spot-check both directions on any open PR: `python3 scripts/check_checkout_is_pr_head.py --repo strands-labs/robots --pr <N> --checkout <tip>` (expect `current`, exit 0) and `--checkout <tip's parent>` (expect `stale-checkout`, exit 1 with the `::warning` annotation).

PR #2680, opened 2026-08-24
fix(isaac): read every MJCF orientation spelling, not just quat

cagataycali:
## Problem

MJCF gives a body or a geom **five mutually exclusive ways to state one rotation**: `quat`, `euler`, `axisangle`, `xyaxes` and `zaxis`. The Isaac scene loader read only `quat`, so an object a model rotates with any of the other four was reported unrotated and the load reported success.

Measured against `mujoco.MjModel` on a one-body scene, one declaration per spelling:

| declared | MuJoCo compiles | `upstream/main` reports | verdict |
| --- | --- | --- | --- |
| `quat="0.707 0 0 0.707"` | `(0.7071, 0, 0, 0.7071)` | `(0.7071, 0, 0, 0.7071)` | read |
| `euler="30 -40 55"` | `(0.8232, 0.0868, -0.3830, 0.4114)` | `(1, 0, 0, 0)` | identity |
| `axisangle="0 1 0 -70"` | `(0.8192, 0, -0.5736, 0)` | `(1, 0, 0, 0)` | identity |
| `xyaxes="0 1 0 -1 0 0"` | `(0.7071, 0, 0, 0.7071)` | `(1, 0, 0, 0)` | identity |
| `zaxis="1 0 1"` | `(0.9239, 0, 0.3827, 0)` | `(1, 0, 0, 0)` | identity |

Four of the five, on both the body and the mesh geom. `euler` is the common one: across the downloaded `robot_descriptions` corpus it accounts for 1120 of the 1218 non-`quat` orientation declarations.

This is the shape of the `<compiler meshdir>`, `<worldbody>` and `<default>` gaps before it - a format feature the reader never asks about, where the fallback is a plausible value rather than a refusal. Identity is a valid orientation, so nothing downstream can tell a body that was never rotated from one whose rotation was dropped.

## Change

`_parse_orientation` reads all five spellings and is consulted for bodies and for mesh geoms.

Two of the inputs are **model-global**, so they are resolved across `<include>` the way `meshdir` already is (`_mjcf_model_toplevel`):

- **`<compiler angle>`** - MJCF's default is degrees, and a model saying `angle="radian"` means `euler="0 0 -1.5708"` is a quarter turn rather than a degree. `lekiwi` is a shipped example.
- **`<compiler eulerseq>`** - a lower-case sequence rotates about the **fixed** axes and an upper-case one about the **moving** axes, which is the same three rotations composed in the opposite order. Both readings and all six axis permutations are graded against the compiler.

A body or geom declaring **two** spellings at once is refused rather than resolved. MuJoCo refuses that model outright (`... multiple orientation specifiers`), so any rotation the reader picked would be a guess; the refusal names both spellings and says they are mutually exclusive.

A malformed value still reads as identity, which is the historical reading for a malformed `quat`, and a zero-length `zaxis`/`xyaxes` cannot define a frame so it does too.

#### The default-class boundary

The exclusivity refusal above is scoped to the spellings **one element declares itself**, which is the only place MuJoCo applies it. A `<default>` class and the element may state the orientation differently and MuJoCo compiles that model, so `_class_attrs`' merged mapping is not the mapping to count.

MuJoCo does not simply let the element win. It keeps the four non-quat spellings in **one slot**, separate from `quat`, and resolves that slot first:

| declared | MuJoCo compiles | why |
| --- | --- | --- |
| class `quat`, element `euler` | the element's `euler` | the alternative slot is resolved before `quat` |
| class `euler`, element `quat` | **the class's `euler`** | same rule - the slot wins whichever level filled it |
| class `zaxis`, element `euler` | the element's `euler` | an inner alternative *replaces* an outer one |
| class `euler`, element `zaxis` | the element's `zaxis` | same slot, inner wins |

So the merge is per-slot: an inner alternative spelling replaces an outer one instead of joining it, `quat` merges by name, and which of the two is in effect is `_parse_orientation`'s decision. Reading the merged mapping flat lost both rules - a class `euler` beside an element `quat` looked like the mutually exclusive case and refused the scene.

## Graded against MuJoCo's own compiler

No expected quaternion in this PR is written by hand. Every one is `model.body_quat` from a compiled fixture.

- **1115 cases** across 5 compiler configurations (degrees, radians, and `eulerseq` `xyz`/`ZYX`/`zyx`), spanning the five spellings on bodies and on mesh geoms: **0 mismatches** on this branch. On `upstream/main` the same sweep reports 4 of 5 spellings as identity.
- **570 real MJCF assets** (the shipped registry plus the downloaded `robot_descriptions` corpus) through `load_mjcf_scene_objects` on both trees: **3 changed, 0 verdict changes.** All three are mesh-geom rotations previously reported as identity - `lekiwi` (`euler="0 0 -1.5708"` in radians, now a quarter turn about Z rather than 0.9 degrees) and `agility_cassie` in two scene files. Every other asset is byte-identical.

A mesh geom's compiled `geom_quat` is **not** the authored frame - MuJoCo folds a mesh's principal-inertia alignment into it - so the geom half is graded against the same declaration on a body, which is itself graded against the compiler.

## Tests

`tests/simulation/isaac/test_mjcf_orientation_spellings.py`, 34 cases.

| tree | failed | passed |
| --- | --- | --- |
| `upstream/main` | 20 | 14 |
| this branch | 0 | 34 |

All 20 pre-fix failures are in the three regression classes and none is in `TestTheReadingIsNotWidened`, whose 8 cases pass on both trees. A sample:

```
body axisangle="0 1 0 -70" was reported as (1.0, 0.0, 0.0, 0.0), but MuJoCo
compiles that model with (0.819152, 0.0, -0.573576, 0.0) (|delta| 5.736e-01)

eulerseq='ZYX' composes the three rotations in an order this reader did not
follow: reported (0.834, 0.037, -0.297, 0.464), MuJoCo compiles
(0.766, 0.319, -0.156, 0.541) (|delta| 2.826e-01)
```

Break table - each row is a plausible regression, applied to the fixed tree:

| break | fires |
| --- | --- |
| `git checkout upstream/main -- loaders.py` | 20 (the full regression set) |
| read the spellings but ignore `<compiler angle>` | the radian cases, incl. the `<include>` one |
| read `angle` but ignore `eulerseq` | the 6 sequence permutations |
| resolve a multiply-declared element instead of refusing | the 2 refusal cases |
| lower-case `eulerseq` composed as moving axes | the 3 lower-case permutations |
| accept a zero-length `zaxis` | 1, the boundary control alone |

## Artifact

A scene declaring one rotation as `euler` under `angle="radian"`. MuJoCo compiles a quarter turn about Z, so the bar lies along Y; reading only `quat` leaves it along X, and the load reports success either way.

MJCF orientation spellings

The right panel is **byte-identical to the compiler's own render** (`np.array_equal` True, `|delta|` 0.0); `upstream/main` is 5.434 mean levels away with the bar's bounding box 365x48 px instead of 49x365. The middle and right panels read the same source file, and the compiler's answer is identical in both columns - the only variable is which spelling the loader reads.

## Verification

Measured at `6a3b791`, on the mujoco build CI resolves (3.12.0) and on the declared floor (3.5.0). The head has since moved to `47118fc`, which adds only the changelog fragment (`git diff 6a3b791..47118fc -- '*.py'` is empty).

- **Gate** - 193 files (every test touching the Isaac loaders, plus the guards that walk the package sources, docstrings and changelog), the identical selection on both trees with the new test file excluded from each: `40 failed, 8569 passed, 121 skipped, 126 errors` **byte-identical**, 166 failing ids each, `comm` empty in both directions. All 166 are `device_connect_edge` absent locally (326 occurrences); it is declared `device-connect-edge>=0.2.0` in `[device-connect]`, which `[all]` pulls in, so CI installs it.
- **mypy** `strands_robots tests tests_integ`: 24 == 24 against a pristine `upstream/main` worktree, none in the changed files.
- `ruff check` / `ruff format --check` clean; the new file passes on **both** mujoco builds.
- No `pxr`/`isaacsim` on this host, so the artifact renders the orientation the loader **reports** through MuJoCo rather than realizing a USD prim. Nothing in the loader is stood in for - the reported value is what a realizer receives.

## Scope

`<default>` also carries `<mesh>`, `<material>` and `<site>` children, and a class-declared `<site>` is a different consumer with its own semantics, so it is not touched here.

### DIFF
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -897,9 +897,45 @@ def _parse_quat(quat_str: str | None) -> tuple[float, float, float, float]:
     return (parts[0], parts[1], parts[2], parts[3])
 
 
+#: MJCF's orientation spellings other than ``quat``. MuJoCo keeps these four in
+#: one "alternative orientation" slot carrying which of them was used, separate
+#: from the ``quat`` attribute, and resolves that slot in preference to ``quat``
+#: when it is set. Two consequences, both measured on mujoco 3.5.0 and 3.12.0:
+#: a ``<default>`` class supplying any of these beats a ``quat`` the element
+#: itself declares, and an inner declaration of one of these replaces an outer
+#: declaration of another rather than sitting beside it.
+_ORIENTATION_ALT_SPELLINGS = ("euler", "axisangle", "xyaxes", "zaxis")
+
 #: MJCF's mutually exclusive orientation spellings, in the order this module
-#: reports them when a model declares more than one (MuJoCo refuses that).
-_ORIENTATION_SPELLINGS = ("quat", "euler", "axisangle", "xyaxes", "zaxis")
+#: reports them when ONE element declares more than one (MuJoCo refuses that).
+_ORIENTATION_SPELLINGS = ("quat", *_ORIENTATION_ALT_SPELLINGS)
+
+
+def _merge_mjcf_attrs(outer: Mapping[str, str], inner: Mapping[str, str]) -> dict[str, str]:
+    """``inner``'s attributes over ``outer``'s, with MJCF's orientation precedence.
+
+    Every attribute merges by name -- an inner declaration replaces an outer one
+    -- except that the orientation spellings do not share a name. MuJoCo keeps
+    :data:`_ORIENTATION_ALT_SPELLINGS` in one slot, so an inner ``euler``
+    REPLACES an outer ``zaxis`` instead of joining it, and reading both would
+    leave the effective rotation ambiguous. ``quat`` is a plain attribute and
+    merges by name; whether it or a surviving alternative spelling is in effect
+    is :func:`_parse_orientation`'s decision, not this one's.
+
+    Args:
+        outer: The enclosing level's attributes -- a ``<default>`` class's, or
+            the parent class's when flattening a nested one.
+        inner: The narrower level's own attributes.
+
+    Returns:
+        The merged mapping, carrying at most one of the alternative spellings.
+    """
+    merged = {**outer, **inner}
+    if any(inner.get(spelling) for spelling in _ORIENTATION_ALT_SPELLINGS):
+        for spelling in _ORIENTATION_ALT_SPELLINGS:
+            if not inner.get(spelling):
+                merged.pop(spelling, None)
+    return merged
 
 
 def _mjcf_angle_units(elements: list[ET.Element]) -> tuple[float, str]:
@@ -992,6 +1028,7 @@ def _quat_from_basis(
 def _parse_orientation(
     attrs: Mapping[str, str],
     *,
+    own: Mapping[str, str] | None = None,
     angle_scale: float = math.pi / 180.0,
     eulerseq: str = "xyz",
 ) -> tuple[float, float, float, float]:
@@ -1013,6 +1050,15 @@ def _parse_orientation(
       orthogonalized against it, then z as their cross product.
     * ``zaxis="x y z"`` -- the minimal rotation taking ``+z`` onto that axis.
 
+    They are mutually exclusive on ONE element, but not across MJCF's
+    ``<default>`` precedence: a class may supply one spelling and the element
+    declare another, and MuJoCo compiles that model. It resolves it by keeping
+    :data:`_ORIENTATION_ALT_SPELLINGS` in a slot separate from ``quat`` and
+    preferring that slot, so an alternative spelling from ANY level beats a
+    ``quat`` from any level -- including a class ``euler`` over the element's own
+    ``quat``. :func:`_merge_mjcf_attrs` has already resolved which alternative
+    spelling survives, so at most one reaches here.
+
     A ``quat`` is reported as written rather than normalized. MuJoCo normalizes
     it, but no ``quat`` in the shipped asset corpus is unnormalized, so
     normalizing here would only move output for input no shipped model
@@ -1022,6 +1068,11 @@ def _parse_orientation(
         attrs: The element's effective attributes -- ``el.attrib``, or
             :func:`_class_attrs`' result where a ``<default>`` class may supply
             the orientation.
+        own: The element's OWN attributes when ``attrs`` merges a class's in.
+            Only the spellings an element declares itself are mutually
+            exclusive, so this is what the refusal below is measured against;
+            defaults to ``attrs``, which is correct for a ``<body>`` (MJCF has
+            no ``<body>`` default class, so nothing is merged in).
         angle_scale: Factor converting a declared angle to radians, from
             :func:`_mjcf_angle_units`.
         eulerseq: Euler axis sequence with its case intact, from
@@ -1033,19 +1084,27 @@ def _parse_orientation(
         reading for a malformed ``quat``).
 
     Raises:
-        ValueError: If the element declares more than one orientation spelling.
+        ValueError: If ``own`` declares more than one orientation spelling.
             MuJoCo refuses such a model outright ("multiple orientation
-            specifiers are not allowed"), so there is no rotation to pick.
+            specifiers are not allowed"), so there is no rotation to pick. A
+            class supplying one spelling while the element declares another is
+            NOT that case -- MuJoCo compiles it, and it is resolved above.
     """
-    declared = [s for s in _ORIENTATION_SPELLINGS if attrs.get(s)]
-    if not declared:
-        return (1.0, 0.0, 0.0, 0.0)
-    if len(declared) > 1:
+    declared_here = [s for s in _ORIENTATION_SPELLINGS if (attrs if own is None else own).get(s)]
+    if len(declared_here) > 1:
         raise ValueError(
-            f"MJCF orientation: {declared} were all declared on one element, but they are mutually "
+            f"MJCF orientation: {declared_here} were all declared on one element, but they are mutually "
             "exclusive - MuJoCo refuses a model with multiple orientation specifiers. Keep one."
         )
-    spelling = declared[0]
+    # An alternative spelling beats ``quat`` whichever level declared it, and
+    # ``_merge_mjcf_attrs`` leaves at most one of them, so the first is the one.
+    alternatives = [s for s in _ORIENTATION_ALT_SPELLINGS if attrs.get(s)]
+    if alternatives:
+        spelling = alternatives[0]
+    elif attrs.get("quat"):
+        spelling = "quat"
+    else:
+        return (1.0, 0.0, 0.0, 0.0)
     if spelling == "quat":
         return _parse_quat(attrs.get("quat"))
     try:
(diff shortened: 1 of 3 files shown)


strands_robots/simulation/isaac/loaders.py
```diff
@@ -896,6 +897,215 @@ def _parse_quat(quat_str: str | None) -> tuple[float, float, float, float]:
...
+    Raises:
+        ValueError: If the element declares more than one orientation spelling.
+            MuJoCo refuses such a model outright ("multiple orientation
+            specifiers are not allowed"), so there is no rotation to pick.
+    """
+    declared = [s for s in _ORIENTATION_SPELLINGS if attrs.get(s)]
+    if not declared:
+        return (1.0, 0.0, 0.0, 0.0)
+    if len(declared) > 1:
+        raise ValueError(
```
yinsong1986, line 1095: [MUST FIX] This refusal fires on valid MJCF when the mapping is `_class_attrs`' merged result, and the raise escapes `load_mjcf_scene_objects`.

On the geom path (`_find_body_mesh`, line 1386) `attrs` is `{**defaults.get(cls, {}), **el.attrib}`, so a `<default>` class supplying one spelling and the element declaring another leaves *both* keys in the mapping. MuJoCo compiles that model - the two specifiers are never on one XML element - and the element's spelling wins. Measured against mujoco 3.8.0:

- class `<geom quat="0.707 0.707 0 0"/>` + element `euler="0 0 90"` on a mesh geom: MuJoCo compiles (element's euler wins, `geom_quat` = quarter turn about Z); this branch raises `ValueError: MJCF orientation: ['quat', 'euler'] were all declared on one element...` out of `load_mjcf_scene_objects`.
- The reverse (class `euler`, element `quat`) also raises - and that one is a **regression**: `upstream/main` read the element's `quat` from the merged attrs and loaded the model correctly.

Why this can't be deferred: it is an unhandled exception on the scene-load path for a model MuJoCo accepts - the exact "loader refuses a model MuJoCo compiles" failure mode this module's own docstrings (`_mjcf_model_worldbody_bodies`) name as a break - and the error message misattributes it to MuJoCo semantics ("MuJoCo refuses a model with multiple orientation specifiers"), which is false for this input, so the caller is pointed at a fix their file does not need.

What would resolve it: apply the element-wins rule before counting - if the element's *own* `attrib` declares any orientation spelling, drop the class-supplied spellings from consideration (that is MuJoCo's resolution, measured above), and reserve the `ValueError` for two spellings on the element's own `attrib`, which MuJoCo does refuse. The body path (`body_el.attrib`, line 1703) is unaffected since it never merges defaults.

yinsong1986 (commented):
## Summary

The Isaac scene loader's `_parse_orientation` now reads all five MJCF orientation spellings (`quat`, `euler`, `axisangle`, `xyaxes`, `zaxis`) for bodies and mesh geoms, resolving `<compiler angle>` and `<compiler eulerseq>` across `<include>` the same way `meshdir` already is. Expectations are graded against `mujoco.MjModel` rather than hand-written quaternions, and the multiply-declared case is refused. One must-fix inline: the refusal is evaluated on the `_class_attrs`-merged mapping on the geom path, so a `<default>` class supplying one spelling plus an element declaring another - a model MuJoCo compiles, with the element winning - now raises `ValueError` out of `load_mjcf_scene_objects`, including a case `upstream/main` handled correctly.

## What's good

- Every expected quaternion is derived from the compiler (`model.body_quat` on a compiled fixture), not restated by hand - the mesh-geom oracle problem (principal-inertia folding into `geom_quat`) is correctly routed around.
- The 570-asset corpus sweep with a break table is exactly the right evidence shape for a silent-plausible-fallback fix.
- Changelog fragment per AGENTS.md step 3; stdlib-only module boundary preserved with the dependency direction documented on `_quat_from_basis`.

strands_robots/simulation/isaac/loaders.py
cagataycali: Fixed in `0fd0f4e3`. Your diagnosis of the refusal is exactly right, and both of your cases now load. One measured refinement to the resolution rule, because it changes what the second case should report.

MuJoCo does not let the element win. It keeps the four non-quat spellings in **one slot**, separate from `quat`, and resolves that slot first - so an alternative spelling from *any* level beats a `quat` from any level. Your two cases, graded by the compiler (`geom_quat`, mujoco 3.12.0, same on the 3.5.0 floor):

| declared | MuJoCo compiles | element alone would be |
| --- | --- | --- |
| class `quat="0.707 0.707 0 0"`, element `euler="0 0 90"` | `[0.707107 0 0 0.707107]` | `[0.707107 0 0 0.707107]` |
| class `euler="0 0 90"`, element `quat="0.707 0.707 0 0"` | `[0.707107 0 0 0.707107]` | `[0.707107 0.707107 0 0]` |

The second row is the refinement: MuJoCo compiles **the class's `euler`**, not the element's `quat`. So `upstream/main` was not reading that model correctly - it read the element's `quat` (a quarter turn about X) where the compiler produces the class's `euler` (a quarter turn about Z). It loaded, but with the wrong rotation, silently. Dropping the class-supplied spellings whenever the element declares one would reproduce that.

What landed instead:

- `_merge_mjcf_attrs` merges a class's attributes into the element's **per slot** - an inner alternative spelling *replaces* an outer one rather than joining it, and `quat` merges by name.
- `_parse_orientation` takes an `own=` mapping and applies the exclusivity check only to the spellings the element declares itself, which is where MuJoCo applies it. Both call sites pass `el.attrib`.
- The nested-class chain uses the same merge, so an inner `<default>` replacing an outer one's spelling resolves the way the compiler does.

Graded against the compiler rather than against a restated table: **215 combinations** - 25 class/element pairs over all five spellings, 180 nested chains, 10 class-only - agree on mujoco 3.12.0 and 3.5.0. The previous reading disagreed with **180 of the 215**, every one by refusing a model MuJoCo compiles.

Break table (control 64 passed):

| break | fires |
| --- | --- |
| exact pre-fix `loaders.py` | 25 |
| exclusivity counted on the merged mapping instead of the element's own | 12 |
| merge not per-slot (an outer alternative survives beside an inner one) | 7 |
| class chain merged outer-over-inner | 1 (the nested sweep alone) |

Blast radius: the **554** downloaded MJCF assets report byte-identical orientations and verdicts. The pair does not occur in the curated corpus, which is why it survived; a scene written by hand reaches it, as yours did.

Gate: the same 120-file selection on this branch and on the reviewed head is byte-identical (`28 failed, 3892 passed, 89 skipped, 126 errors`, 154 failing ids, `comm` empty both ways; all of them `device_connect_edge`, the declared `[device-connect]` extra). `mypy` 24 == 24, none in the changed files.

yinsong1986 (approved):
## Summary

The Isaac scene loader's orientation reading is widened from `quat` alone to all five mutually exclusive MJCF spellings (`quat`, `euler`, `axisangle`, `xyaxes`, `zaxis`) for top-level bodies and mesh geoms, with the model-global `<compiler angle>` and `<compiler eulerseq>` resolved across `<include>` the same way `meshdir` already is. A single element declaring two spellings is refused with a `ValueError` naming both - a model MuJoCo itself refuses, so no compilable asset can hit the raise - while a class/element pair spelling the orientation differently is resolved per MuJoCo's alternative-orientation slot (`_merge_mjcf_attrs` + the `own=` split in `_parse_orientation`), which addresses the concern raised in the prior review round. All new helpers are private and `load_mjcf_scene_objects`' signature is unchanged, so there is no API surface at stake; malformed values keep the historical identity fallback, so the only behaviour that moves is rotations that were previously dropped.

## What's good

- Every expected quaternion is read from `mujoco.MjModel` rather than hand-written, so the tests grade against the compiler itself; I re-ran the new file on this head (all pass, mujoco 3.8.0) and against `main`'s loader (45 of 64 fail, the regression classes), confirming the pin.
- The other `_class_attrs` consumers (joint attrs, geom size/type/fromto readers) are unaffected by the merge change, which only prunes the orientation alternatives.
- The `Raises:` block on `_parse_orientation` names the new refusal, per AGENTS.md's docstring-completeness rule; the changelog fragment follows the `changelog.d/` process; new strings are plain ASCII.
- Deliberate scope boundaries (mesh-geom `geom_quat` not being the authored frame, mixed-case `eulerseq`, `<default>` sites) are stated rather than silently assumed.

PR #2681, opened 2026-08-24
fix(sim): resolve a robot plane's orientation through all five MJCF spellings before stripping it as the ground

cagataycali:
## What

`SpecBuilder.attach_robot` deletes a plane geom from a robot MJCF when it reads as the world's z=0 ground, so a scene does not end up with two overlapping floors. That check asked the geom for `quat` alone. MJCF spells one rotation five ways -- `quat`, `euler`, `axisangle`, `xyaxes`, `zaxis` -- and MuJoCo keeps the four alternatives in a slot separate from `quat`. So a plane rotated by any of those four read as unrotated, and an authored wall, ramp or angled panel sitting at z=0 was silently deleted on attach.

`zaxis` is the idiomatic spelling for a plane, because a plane's normal *is* its local z.

## The defect

`_is_z0_ground_plane` resolved the orientation as:

```python
quat = geom.get("quat", "").split()
if quat and len(quat) == 4 and abs(float(quat[0]) - 1.0) > 1e-6:
    return False   # rotated -> not the ground plane
```

An `euler` / `axisangle` / `xyaxes` / `zaxis` declaration leaves `quat` absent, so `quat` is empty, the guard does not fire, and the geom is treated as flat.

## Measured

A robot MJCF declaring one wall at z=0, once per spelling, attached into a world with `attach_robot` (all five describe the same quarter turn about x, so the plane is a vertical wall in every row):

| spelling on the wall geom | upstream/main | this branch |
| --- | --- | --- |
| `quat="0.7071 0.7071 0 0"` | kept | kept |
| `euler="90 0 0"` | **DELETED** | kept |
| `axisangle="1 0 0 90"` | **DELETED** | kept |
| `zaxis="0 1 0"` | **DELETED** | kept |
| `xyaxes="1 0 0 0 0 1"` | **DELETED** | kept |

4 of the 5 spellings lost the geom. Nothing is reported: `attach_robot` returns normally and the compiled scene simply has one geom fewer, so the robot arrives with a wall missing.

Two further readings the same resolution needed, both measured against `mujoco.MjModel`:

- `<compiler angle>` defaults to **degrees**, so `euler="0 0 -1.5708"` is 0.9 degrees, not a quarter turn. It is model-global and survives `<include>`.
- `zaxis` need not be a unit vector. `zaxis="0 0 5"` is flat and was read as flat only by luck; `zaxis="0 5 0"` is a wall.

## Change

`spec_builder.py` resolves the geom's orientation through all five spellings before deciding, with the same `<compiler angle>` and `<compiler eulerseq>` reading the module already applies elsewhere, and normalises `zaxis` / `axisangle` axes. `_is_z0_ground_plane` keeps its existing position and size rules untouched; only the flatness test changed.

The refusal to strip is unchanged in shape -- a rotated plane is still simply not the ground plane -- so no message moves and no caller sees a new error.

## Tests

`tests/simulation/mujoco/test_spec_builder.py`, 92 tests. Pre-fix split at `9adfefb1`: **9 failed / 83 passed**, every failure behavioural, e.g.

```
AssertionError: a wall declared with zaxis was stripped as the ground plane:
planes after attach == ['ground'], expected ['ground', 'arm1/wall']
```

Three classes are regression evidence (the five spellings, the degree/radian reading, the axis normalisation); `TestTheStripStillStrips` and `TestTheReadingIsNotWidened` pass on both trees and hold the boundary -- a genuinely flat plane in every spelling is still removed, and a second flat plane is still removed too.

## Break table

Each row is a plausible regression applied to the fixed tree; the control is the unmodified branch.

| break | fires | notes |
| --- | --- | --- |
| control | 0 of 92 | |
| exact pre-fix (`git checkout upstream/main -- spec_builder.py`) | 9 | the full regression set |
| only `euler` handled | 7 | the three other alternatives, plus the end-to-end attach cases |
| an unused slot read as "not flat" | 14 | includes `test_a_second_flat_plane_is_still_stripped` -- the strip stops working at all |
| `zaxis` not normalised | 1 | the scaled-flat-`zaxis` case alone |
| `axisangle` checks the axis instead of the angle | 2 | the two `axisangle` cases alone |
| the alternative slot read without the absent-attribute guard | 4 | **three pre-existing `TestIsZ0GroundPlane` tests** plus the stand-in control |

The last row is the useful one: it fires tests that were already in the file, so the fix extends the `quat` path rather than replacing it. The two single-firing rows show the normalisation and the angle-vs-axis clauses are each load-bearing.

## Artifact

authored wall stripped as a duplicate ground plane

One robot MJCF that ships its own z=0 floor **and** an authored wall declared with `zaxis`, attached into a world by the same `attach_robot` call in all three panels, rendered headless with `MUJOCO_GL=egl`. The third panel is MuJoCo's own compiled model of that robot file, i.e. the asset as authored.

| panel | planes after attach | wall pixels | arm pixels |
| --- | --- | --- | --- |
| upstream/main | `['ground']` | **0** | 6868 |
| this branch | `['ground', 'arm1/wall']` | **35833** | 6856 |
| the asset as authored | `['floor', 'wall']` | 35833 | 6856 |

Pixels per geom come from MuJoCo's segmentation render, so the count is exact rather than a colour threshold. The branch renders the wall and the arm identically to the authored asset. In both trees the robot's own duplicate z=0 floor is still stripped and the world `ground` survives, which is what the strip exists to do; the 12-pixel arm difference on main is antialiasing along the silhouette where the wall would be behind it.

## Verification

Measured at `3eb12232`, on `mujoco 3.12.0` (what CI resolves) unless noted.

- 323-file blast radius, the identical selection on this branch and on `upstream/main`, with the changed test file excluded from both: **41 failed, 8024 passed, 116 skipped, 126 errors** on each, **167 failing ids on each, `comm` empty in both directions**. All 167 are `device_connect_edge` absent on this box (326 occurrences), declared `device-connect-edge>=0.2.0` in the `[device-connect]` extra which `[all]` pulls in, so CI installs it.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `upstream/main` worktree, identical message sets, none in the changed files.
- `ruff check` and `ruff format --check` clean over the CI scope.
- The changed test file: **92 passed on mujoco 3.12.0 and on 3.5.0**, the declared floor.

## Scope

`_is_z0_ground_plane`'s position and size rules are untouched -- only the flatness test read one spelling of five. The same five spellings apply to `<camera>`, `<site>` and `<light>` elsewhere in the file, and to the robot-link reader in the Isaac loader, which reads no orientation at all; each is a different consumer with its own semantics and wants its own change.

### DIFF
--- a/changelog.d/2681-mjcf-ground-plane-orientation-spellings.md
+++ b/changelog.d/2681-mjcf-ground-plane-orientation-spellings.md
@@ -1,0 +1,17 @@
+### Fixed: a robot's authored plane is no longer stripped as the ground when its rotation is spelled with euler, axisangle, zaxis or xyaxes
+
+`SpecBuilder.attach_robot` deletes a plane geom from a robot MJCF when it reads as
+the world's z=0 ground, so a scene does not end up with two overlapping floors.
+That flatness test asked the geom for `quat` alone. MJCF spells one rotation five
+ways and MuJoCo keeps `euler`, `axisangle`, `xyaxes` and `zaxis` in a slot separate
+from `quat`, so a plane rotated by any of those four read as unrotated and an
+authored wall, ramp or angled panel sitting at z=0 was silently deleted on attach.
+`zaxis` is the idiomatic spelling for a plane, because a plane's normal is its
+local z.
+
+The orientation is now resolved through all five spellings, honouring the
+model-global `<compiler angle>` (which defaults to degrees) and `<compiler
+eulerseq>` readings the module already applies elsewhere, and normalising `zaxis`
+and `axisangle` axes. The position and size rules are unchanged, a genuinely flat
+plane is still stripped in every spelling, and no message or caller-visible error
+moves.
--- a/strands_robots/simulation/mujoco/spec_builder.py
+++ b/strands_robots/simulation/mujoco/spec_builder.py
@@ -1342,18 +1342,86 @@ _ADOPTED_OPTION_FIELDS: tuple[str, ...] = (
 )
 
 
+_ORIENTATION_TOL: Final[float] = 1e-6
+
+
+def _alt_orientation_is_identity(alt: Any) -> bool | None:
+    """Whether the alternative-orientation slot carries an identity rotation.
+
+    MJCF spells a rotation five ways - ``quat`` plus the four alternatives
+    ``euler`` / ``axisangle`` / ``xyaxes`` / ``zaxis`` - and MuJoCo keeps the
+    four alternatives in one slot (``elem.alt``) behind a discriminator,
+    leaving ``elem.quat`` at identity. Reading ``quat`` alone therefore reports
+    every alternative spelling as unrotated, and ``zaxis`` is the idiomatic
+    spelling for a plane, whose normal *is* that vector.
+
+    Only identity is decided here, never the rotation itself: a plane's normal
+    stays +Z exactly when the slot carrying the rotation is identity, and for
+    each spelling that is a direct test on the declared numbers. It needs no
+    angle unit, because zero is zero in degrees and in radians, so the model's
+    ``<compiler angle>`` does not enter into it.
+
+    Args:
+        alt: the element's ``alt`` slot.
+
+    Returns:
+        ``True``/``False`` when this slot carries the element's rotation, or
+        ``None`` when its discriminator says the rotation lives in ``quat``
+        instead, which the caller should then read.
+    """
+    mujoco = _ensure_mujoco()
+    kinds = mujoco.mjtOrientation
+    try:
+        kind = int(alt.type)
+    except (AttributeError, TypeError, ValueError):
+        return None
+    if kind == int(kinds.mjORIENTATION_EULER):
+        return bool(np.allclose(np.asarray(alt.euler, dtype=float), 0.0, atol=_ORIENTATION_TOL))
+    if kind == int(kinds.mjORIENTATION_AXISANGLE):
+        # Any axis, provided the rotation about it is zero.
+        return abs(float(alt.axisangle[3])) <= _ORIENTATION_TOL
+    if kind == int(kinds.mjORIENTATION_ZAXIS):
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

`_is_z0_ground_plane` decided flatness from `geom.quat` alone, so a robot plane rotated via any of MJCF's four alternative spellings (`euler`, `axisangle`, `zaxis`, `xyaxes`) read as unrotated and was silently deleted by `attach_robot`'s duplicate-floor strip. The fix adds `_alt_orientation_is_identity`, which reads the spec element's `alt` slot behind its `mjtOrientation` discriminator and answers identity per spelling; the discriminator falls through to the existing `quat` path when the rotation lives there, so the pre-existing stand-in-object tests keep passing unchanged. The asymmetry is the right one and is documented in the docstring: an unrecognisable orientation means "kept", so the worst residual case (e.g. a scaled non-unit `xyaxes` that compiles to identity but is not byte-recognised) leaves a redundant second floor rather than deleting an authored surface. Verified locally at `407157d6`: the changed test file passes 92/92 on mujoco 3.8.0, and reverting `spec_builder.py` to the pre-fix tree reproduces the claimed 9 failed / 83 passed split with behavioural failure messages.

## What's good

- Enum comparisons follow AGENTS.md's value-based rule (`int(kind) == int(kinds.mjORIENTATION_*)`), no `in (enum, ...)` membership anywhere in the new code.
- The `except (AttributeError, TypeError, ValueError)` tuple has no member covered by another, per the except-tuple scope rule.
- Changelog fragment under `changelog.d/`, no direct `CHANGELOG.md` edit; test names describe behaviour, not provenance; no host paths; all-ASCII; ruff check/format clean on the changed files.
- Tests derive expectations from the compiled `geom_quat` oracle rather than a hand-written table, and pin the boundary in both directions (walls survive, genuinely flat duplicates are still stripped, elevated planes unchanged).
- Scope discipline: only the flatness clause of the predicate changed; position and size rules untouched, no caller-visible message moves.

PR #2683, opened 2026-08-24
fix(examples): report a LIBERO matrix row whose result line could not be read

cagataycali:
## Problem

`examples/libero/libero_backend_matrix.py` records a row as `ok` whenever the
per-backend driver exits 0, without checking that it actually read the row's
numbers:

```python
sr, wt = _parse_driver_output(completed.stdout)
return RowResult(label, sr, wt, "ok")
```

`ok` is therefore a verdict about the exit code alone. A driver that exits 0 but
prints no result line - or one whose line has drifted out of `_RE_RESULT`'s
shape - reports the same status as a row that measured cleanly, differing only
in two `--` cells. A reader of the `status` column cannot tell them apart, and
the module's own docstring says the table is meant to be parsed ("The output is
stable enough that a follow-up CI job can parse it for regression tracking").

Measured by driving the real `_run_one` against stub drivers (`sr`/`wt` are the
values it reported, identical on both trees - only the status moves):

| driver behaviour | success_rate | wall_time | `upstream/main` status | this branch |
| exits 0, prints the real result line | 0.4 | 12.5 | `ok` | `ok` |
| exits 0, prints no result line | None | None | **`ok`** | `incomplete (driver exited 0 but printed no parseable result line)` |
| exits 0, line does not match `_RE_RESULT` | None | None | **`ok`** | `incomplete (driver exited 0 but printed no parseable result line)` |
| exits non-zero | None | None | `skip (Isaac Sim is not available on this host)` | unchanged |

## Change

A row whose `success_rate` / `wall_time` could not be read now records
`_UNPARSED_RESULT_STATUS` instead of `ok`.

It is deliberately **not** a `skip (...)`. `skip` means the driver refused to
run - the documented normal case on a host without Isaac Sim - whereas this
driver ran to completion, so the reader is pointed at a driver/parser format
disagreement rather than at a missing backend.

`_parse_driver_output`'s docstring justified the old `ok` with a driver that
"exited early before the eval started (e.g., HF token missing for
`--policy=groot`)". That path no longer reaches it: `run.py` raises on a refused
setup step and on an eval it cannot complete, and its `main` has **no return
statement at all** (measured), so such a run exits non-zero and is recorded by
exit code as a `skip`. The docstring now states the reachable cause, and that
reasoning is pinned by `TestTheDriverHasNoZeroExitPathThatSkipsItsResultLine`
rather than left as prose.

## Artifact

The table the script prints, from the same four stub drivers on each tree. Three
rows read `ok` on `upstream/main` and only one of them carries a measurement:

`upstream/main`:

```
=== libero_backend_matrix ===
Task: libero-spatial-pick_up_the_red_cube  (10 episodes, seed=42)
backend       success_rate   wall_time  status
----------------------------------------------------------------
mujoco                0.40       12.5s  ok
isaac-1                 --          --  ok
isaac-4096              --          --  ok
fleet                   --          --  skip (Isaac Sim is not available on this host: no isaacsim)
=== /libero_backend_matrix ===
```

This branch:

```
=== libero_backend_matrix ===
Task: libero-spatial-pick_up_the_red_cube  (10 episodes, seed=42)
backend       success_rate   wall_time  status
----------------------------------------------------------------
mujoco                0.40       12.5s  ok
isaac-1                 --          --  incomplete (driver exited 0 but printed no parseable result line)
isaac-4096              --          --  incomplete (driver exited 0 but printed no parseable result line)
fleet                   --          --  skip (Isaac Sim is not available on this host: no isaacsim)
=== /libero_backend_matrix ===
```

## Tests

12 new tests in `tests/test_examples_libero_drivers.py`, which already owns this
module (11 functions, one parametrized over both backends; the file goes 34 -> 46). Pre-fix split: **4 failed / 42 passed**, all four failures in
`TestARowWithNoMeasurementIsNotReportedAsOk` and none in the two control
classes. Branch: **46 passed**.

The regression tests drive the real `_run_one` (a real subprocess against a stub
driver), and the happy-path control builds its result line with `run.py`'s own
`_format_result_lines`, so it stays tied to the single formatter both
subcommands print through instead of restating the format.

### Break table

| break | fails | which |
| --- | --- | --- |
| control (branch as pushed) | 0 / 46 | - |
| exact pre-fix (`git checkout upstream/main -- <file>`) | 4 | the whole regression class |
| unparsed row classed as `skip (...)` | 1 | `test_the_status_names_the_cause_and_is_not_a_skip`, on its "must not be a skip" clause |
| status stops naming what was unreadable (`"incomplete"`) | 1 | the same test, on its "names the cause" clause |
| an `incomplete` row forces a non-zero matrix exit | 1 | `test_the_matrix_still_exits_zero_so_the_table_stays_the_verdict` (the boundary control) |
| `sr is None or wt is None` -> `and` | **0** | does not fire, and deliberately so - see below |

The last row is worth stating rather than hiding: `_parse_driver_output` returns
both floats or `(None, None)`, so a half-read row cannot occur and `or` is
today equivalent to `and`. That invariant was pinned by nothing, so
`test_the_parser_reports_both_numbers_or_neither` now asserts it over six
stdout shapes; the defensive `or` is what still catches a future parser that
reports one number without the other.

## Scope

Deliberately not taken: making an `incomplete` row a non-zero exit for the
matrix process. `main` returns 0 after printing the table however the rows
turned out, and an `unavailable` Isaac row on a CPU-only host is the documented
normal case - so the exit code is a separate contract for the matrix's callers.
The boundary control fails if that is changed without deciding it.

No library behaviour changes: the diff touches only
`examples/libero/libero_backend_matrix.py` and its test file, so no policy,
simulation, rendering, recording or asset path is involved.

## Verification

- `ruff check` and `ruff format --check` clean on the CI scope
  (`strands_robots tests tests_integ`, 1508 files) and on the changed example.
- `mypy strands_robots tests tests_integ`: **24 errors == 24** on a pristine
- Gate: a 150-file selection of every test that walks the examples tree or
  grades example sources, run identically on both trees with the changed test
  file excluded from each -> **`35 failed, 7393 passed, 61 skipped, 126 errors`
  byte-identical**, 35 failing ids each, `comm` empty both ways. All are
  dependency-absent locally: `device_connect_edge` (declared
  `device-connect-edge>=0.2.0` in the `device-connect` extra, which `all`
  pulls in), `awsiot` (`awsiotsdk>=1.21.0,<2.0.0` in `mesh-iot`), plus two
  `tests/test_lerobot_import_surface.py` failures that reproduce on the pristine
  base in isolation: that environment resolves lerobot 0.5.1, which predates
  `lerobot.policies.rtc.reanchor_relative_rtc_prefix` exactly as the call site's
  own comment describes, so they are an environment artifact rather than drift
  and are unrelated to this change.


yinsong1986 (approved):
## Summary

This PR stops `examples/libero/libero_backend_matrix.py` from recording a matrix row as `ok` when the per-backend driver exits 0 but its result line is absent or does not match `_RE_RESULT` -- such a row now records a distinct `incomplete (driver exited 0 but printed no parseable result line)` status, deliberately kept apart from `skip (...)` (driver refused to run) so a reader is pointed at a driver/parser format disagreement rather than a missing backend. The change is confined to the example script and its existing test file, plus a changelog fragment. I verified the description's key claims against the source at this head: `run.py`'s `main` has no return statement, `_evaluate_and_report` raises rather than returning on a failed eval, `_format_result_lines` exists with exactly the kwargs the test control passes, and the test's `_run_one` invocation matches the real signature. The new status string is ASCII, the happy-path `ok` spelling downstream parsers grep for is unchanged, and the boundary control pins that an `incomplete` row does not change the matrix's exit-code contract.

## What's good

- The regression tests drive the real `_run_one` against real subprocesses, and the happy-path control builds its line via `run.py`'s own formatter rather than restating the format -- so the test cannot drift from the single source of the line's shape.
- The `or`-vs-`and` guard question is pinned by `test_the_parser_reports_both_numbers_or_neither` instead of left as prose, and the docstring's reachability claim is pinned by an AST-level test class rather than asserted in a comment.
- Test names describe behaviour, not provenance (AGENTS.md > Review Learnings #85 > Testing), no host paths, changelog fragment present, and the deliberately-not-taken exit-code change is documented and guarded by its own control.

yinsong1986 (approved):
## Summary

This PR makes `examples/libero/libero_backend_matrix.py` stop recording a row as `ok` when the per-backend driver exits 0 but its result line is absent or fails to match `_RE_RESULT`: such a row now records a distinct `incomplete (driver exited 0 but printed no parseable result line)` status, deliberately kept apart from `skip (...)` (driver refused to run) so a reader is pointed at a driver/parser format disagreement rather than a missing backend. The `ok`, `skip (...)` and `unavailable (...)` spellings for the other outcomes are unchanged, the matrix's zero-exit contract is untouched and pinned by a boundary control, and the stale `_parse_driver_output` docstring justification (a zero-exit early bail in `run.py` that no longer exists) is replaced and pinned by AST tests over `run.py`'s `main` / `_evaluate_and_report`. Verified independently: the 12 new tests fail 4/42 with the pre-fix example file restored and pass 46/46 on the head, `ruff check` / `ruff format --check` are clean on both changed files, the new status string is plain ASCII, and a changelog fragment is present per AGENTS.md step 3.

## What's good

- The regression tests drive the real `_run_one` against a real subprocess stub rather than mocking the parse, and the happy-path control builds its result line with `run.py`'s own `_format_result_lines`, so the format has one source.
- The `or`-vs-`and` equivalence in the new guard is pinned explicitly (`test_the_parser_reports_both_numbers_or_neither`) instead of left as an unstated invariant.
- Scope discipline: the diff touches only the example and its owning test file; the exit-code contract for matrix callers is explicitly left as a separate decision with a control test that fires if it changes undecided.

PR #2684, opened 2026-08-24
fix(mesh): record every resource the IoT bootstrap ensures in its ledger

cagataycali:
## Problem

`BootstrappedAccount.created` + `.skipped` are the whole record an operator has of what a provisioning run touched - the closing log line counts them, and the dataclass docstring calls itself the record of "every resource `bootstrap_account` ensured". Thirteen module-level helpers mutate the account. Twelve recorded what they did; three did not.

`_grant_iot_invoke_lambda` recorded neither the `lambda:InvokeFunction` grant it creates for the E-stop fan-out rule nor the one it finds already present. Its sibling `_grant_iot_invoke_provisioning_hook` - the other `add_permission` helper, forty lines away, with the same `try` / `ResourceConflictException` shape - recorded both. So the one resource missing from the audit trail was a permission grant.

`_ensure_iot_action_role` and `_ensure_provisioning_role` recorded the role they create but not the existing role they reuse, unlike `_ensure_lambda_role` and `_ensure_provisioning_hook_role`, which record both.

## Measurement

The real helpers driven against fake AWS clients, in the three states an account can be in:

| account state | `created` / `skipped` before | after |
| --- | --- | --- |
| fresh (nothing exists) | 12 / 0 | **13** / 0 |
| fully provisioned | 0 / 10 | 0 / **11** |
| resumed (roles + Lambdas kept, rules + template lost) | 3 / 7 | 3 / **10** |

The entry absent from the fresh-account record is exactly `lambda-permission:estop-invoke`. The resumed row is the realistic re-run - an earlier attempt died after the roles and Lambdas - and it is the one that reaches both role helpers on their reuse path; it under-reported three of the ten resources it reused.

## Change

The three missing ledger writes, each in the vocabulary its own sibling already uses. Both permission statements now record under a single-sourced constant (`ESTOP_INVOKE_PERMISSION`, `PROVISIONING_HOOK_INVOKE_PERMISSION`) because a resource-policy statement has no ARN of its own to return, so its ledger name is its only handle. The values are byte-identical to the literal the provisioning-hook helper used, so the existing assertions in `tests/mesh/test_iot_provisioning_hook.py` still hold unchanged.

`BootstrappedAccount`'s docstring now states the invariant the helpers are held to: every ensured resource appears in exactly one of the two lists.

Scope left alone deliberately, each with a control test: a `dry_run` preview ensures nothing so it records nothing, and a `force_update` replacement stays in `created` under the existing `(updated)` spelling rather than becoming a reuse.

## Tests

`tests/mesh/test_iot_bootstrap_ledger.py` (new, 20 tests). Nothing previously ran the real `_ensure_*` / `_grant_*` helpers together - the existing full-bootstrap tests stub every one of them, so the ledger they build was unobservable, and `TestGrantIotInvokeLambda` asserted the `add_permission` call without looking at the record. The new file supplies fake clients so the helpers run unmodified, then grades the ledger.

Pre-fix split, with only the three ledger writes reverted: **8 failed / 12 passed**, all eight in the three regression classes and zero in `TestTheLedgerIsNotWidened`. (Reverting the source file wholesale instead fails at collection, since the two ledger-name constants do not exist there.)

| break | fails | notable |
| --- | --- | --- |
| control (branch) | 0 of 20 | |
| all three writes reverted | 8 | the pre-fix behaviour |
| grant helper only | 6 | |
| `_ensure_iot_action_role` only | 3 | complementary with the row below |
| `_ensure_provisioning_role` only | 3 | shares only the resumed + structural tests |
| an already-present grant recorded as `created` | 4 | over-reach: a reuse reported as a change |
| sibling stops using the constant | 1 | the single-sourcing test alone |
| reuse path records but returns the wrong ARN shape | 1 | the role test alone |

`TestTheLedgerContractHoldsForEveryHelper` grades the module by AST rather than by scenario, so a helper added later inherits the contract; it carries its own non-vacuity assertion (the scan must reach at least thirteen helpers), because an empty scan would satisfy the rule while measuring nothing.

## Gate

`ruff check` + `ruff format --check` on `strands_robots tests tests_integ` clean. mypy **24 = 24** against a pristine `upstream/main` worktree, identical message sets, none in the changed files.

A 329-file selection - all of `tests/mesh`, every test naming the bootstrap, and every test that scans package sources - run identically on both trees with the new file excluded from each: **`73 failed, 10642 passed, 65 skipped, 150 errors`** on both (427.45s vs 427.83s), 73 failing ids each, set difference empty in both directions. Every shared failure is a declared optional dependency absent from this environment - `device_connect_edge` (302 occurrences) and `device_connect_agent_tools` (40), declared in `[device-connect]`, and `awsiot` (88), declared as `awsiotsdk` in `[mesh-iot]`, all of which `[all]` installs. The new file plus the three existing bootstrap test files: 83 passed under random ordering.

No visual artifact: this changes no policy, simulation, rendering, recording or robot asset - it changes what a provisioning run reports about itself.

### DIFF
--- a/changelog.d/2684-iot-bootstrap-ledger-completeness.md
+++ b/changelog.d/2684-iot-bootstrap-ledger-completeness.md
@@ -1,0 +1,20 @@
+### Fixed
+
+- **mesh**: every resource the AWS IoT bootstrap ensures is now recorded in
+  `BootstrappedAccount.created` / `.skipped`. Those two lists are the whole record an
+  operator has of what a provisioning run touched, and the closing log line counts them, but
+  three of the thirteen helpers that mutate the account ensured a resource without recording
+  it. `_grant_iot_invoke_lambda` recorded neither the `lambda:InvokeFunction` grant it creates
+  for the E-stop fan-out rule nor the one it finds already present, while the other
+  `add_permission` helper forty lines away recorded both - so the one resource absent from the
+  audit trail was a permission grant. `_ensure_iot_action_role` and `_ensure_provisioning_role`
+  recorded the role they create but not the existing role they reuse, unlike the two sibling
+  role helpers. Measured with the real helpers driven against fake clients: a fresh account
+  creates thirteen resources and reported twelve, a fully provisioned one reused eleven and
+  reported ten, and a resumed run - roles and Lambdas left by an earlier attempt, rules and
+  template lost - reused ten and reported seven. Both permission statements now record under a
+  single-sourced ledger name, because a resource-policy statement has no ARN of its own to
+  return; the values are unchanged, so existing assertions on the provisioning-hook entry hold.
+  No resource, ARN, IAM policy or API call changes - only what the run reports about itself.
+  A new structural guard derives the helper set from the module, so a helper added later is
+  held to the same contract on arrival.
--- a/strands_robots/mesh/iot/bootstrap.py
+++ b/strands_robots/mesh/iot/bootstrap.py
@@ -73,6 +73,13 @@ PROVISIONING_HOOK_ROLE = "strands-mesh-provisioning-hook-role"
 _PROVISIONING_HOOK_VERSION = 1
 LOG_GROUP_NAME = "/aws/iot/strands-mesh"
 
+#: Ledger names for the two Lambda resource policy statements the bootstrap
+#: grants. A permission statement has no ARN of its own, so these are the
+#: only handle :class:`BootstrappedAccount` can record it under - and they are
+#: single-sourced here so the record and any reader agree on the spelling.
+ESTOP_INVOKE_PERMISSION = "lambda-permission:estop-invoke"
+PROVISIONING_HOOK_INVOKE_PERMISSION = "lambda-permission:provisioning-hook-invoke"
+
 
 # Lambda source for the E-stop fan-out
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Completes the `BootstrappedAccount.created` / `.skipped` audit trail: `_grant_iot_invoke_lambda` now records the E-stop `lambda:InvokeFunction` grant on both its create and reuse paths, and `_ensure_iot_action_role` / `_ensure_provisioning_role` now record the role they reuse, matching the vocabulary of their sibling helpers. The two permission-statement ledger names are lifted into single-sourced module constants (values byte-identical to the prior literal, so existing membership assertions hold), and the dataclass docstring now states the exactly-one-of-two-lists invariant. Verified against `main`: each helper runs once per `bootstrap_account` call, so no duplicate entries; nothing outside the closing log line consumes the two lists, so the additions are strictly informative. No resource, ARN, IAM policy, wire format, or API call changes.

## What's good

- The new ledger test file drives the real helpers against fake clients rather than stubbing them, which is what made the missing entries observable in the first place; the AST-based `TestTheLedgerContractHoldsForEveryHelper` carries its own non-vacuity floor (>= 13 helpers) per the pattern AGENTS.md pins elsewhere.
- Scope discipline: `dry_run` and `force_update` boundaries are deliberately left alone, each with a control test, and the changelog lands as a `changelog.d/` fragment per step 3 of the PR workflow.
- Pre-fix failure split (8/20 with only the ledger writes reverted, complementary per-helper splits) demonstrates the regression tests actually pin the fix.

Issue #2167, opened 2026-08-12
so101 gripper tool site sits inside its own collision hull, making a two-finger grasp unreachable in MuJoCo


sundargthb:
## Summary

On the shipped `so101` model, the gripper's advertised tool site (`arm/gripper`)
lies **inside** the gripper's own collision geometry. An object placed at the
grasp point is 15 mm penetrated at `t=0`, before any motion. The three collidable
geoms on the gripper chain are all **unnamed convex hull meshes**, and the hull of
the static jaw overlaps the hull of the moving jaw at *every* jaw angle, so there
is no open gap between the fingertips for an object to occupy.

This makes a two-finger friction grasp unreachable in MuJoCo without the user
patching the model, and the failure is silent: the sim reports success, the object
is simply shoved out of the jaw.

## Reproduction

Public API only, no user patches:

```python
import mujoco, numpy as np
from strands_robots.simulation.mujoco.simulation import MuJoCoSimEngine

sim = MuJoCoSimEngine()
sim.create_world(timestep=0.002, gravity=[0, 0, -9.81], ground_plane=True)
sim.add_robot(name="arm", data_config="so101", position=[0, 0, 0])
m, d = sim.mj_model, sim.mj_data

for body in ("arm/gripper", "arm/moving_jaw_so101_v1"):
    b = mujoco.mj_name2id(m, mujoco.mjtObj.mjOBJ_BODY, body)
    for g in range(m.ngeom):
        if m.geom_bodyid[g] == b and m.geom_contype[g] != 0:
            name = mujoco.mj_id2name(m, mujoco.mjtObj.mjOBJ_GEOM, g)
            print(body, "gid", g, "name", name, "type", int(m.geom_type[g]))
```

Output - every collidable geom is an unnamed mesh (`type 7`):

```
arm/gripper               gid 26 name None type 7
arm/gripper               gid 28 name None type 7
arm/moving_jaw_so101_v1   gid 30 name None type 7
```

### The hulls overlap at every jaw angle

```python
jid = mujoco.mj_name2id(m, mujoco.mjtObj.mjOBJ_JOINT, "arm/6")
qa = m.jnt_qposadr[jid]
for q in (float(m.jnt_range[jid][1]), float(m.jnt_range[jid][0])):   # open, closed
    d.qpos[:] = 0
    d.qpos[qa] = q
    mujoco.mj_forward(m, d)
    print(q, mujoco.mj_geomDistance(m, d, 28, 30, 1.0, None))
```

[...]

Issue #2145, opened 2026-08-11
sim: no route to a working SO-101 pick - grasp quality has three conditions and the library exposes none of them


sundargthb:
## Summary

`strands-robots` ships an SO-101 model, a MuJoCo sim backend, and RL/policy
plumbing, but there is **no route in the library to a working SO-101 pick in
simulation**. Users who follow the sim path to its natural conclusion get an arm
that reaches, closes on the cube, and does not lift it.

Measured over two SAC runs (12 parallel envs, L4 GPU, 6-dim action space
including both wrist joints):

| run | steps | success | clean pinch | lift (goal 8cm) |
|---|---|---|---|---|
| reward = levelness + gated lift | 1.5M | 0.0% | 78% | 0.18-1.14cm |
| reward = levelness x face-align, upright term, clean-gated lift | 320k | 0.0% | 6.0% | 0.20cm |

## Three separate geometric conditions, none of them expressible in the library today

Each was found by rewarding the previous one and measuring what the policy
actually did with it.

1. **Level pad axis.** A wrist probe optimising "fingertips lowest" returns a
   roll that holds one pad above the other, so the high pad meets the cube's top
   face and it rolls off the jaw corner.
2. **Face-aligned pad axis.** `levelness` scores **0.994-1.000 at every wrist
   roll from -1.5 to +1.5 rad** while face alignment over the same sweep runs
   **0.372 to 0.997**. A levelness reward therefore cannot distinguish a face
   pinch from a 45deg corner grab, and the 1.5M-step run optimised into the
   corner: 78% "clean" contact, no lift.
3. **Cube upright.** A pinch that is level 1.000 and aligned 1.000 at contact
   still arrives at the 8cm goal at **33deg** of cube roll, because the cube
   pivots about the pad line inside a grip that never slips. Lift speed is
   irrelevant (33deg at full or quarter speed). A pad-contact test reports this
   as a clean grasp throughout.

## Why this is a library gap and not a tuning exercise

- **No grasp-quality primitive.** There is no `levelness` / `face_align` /
  `upright`-equivalent anywhere in `strands_robots`, so every user reinvents
  them, and the obvious first attempt (level, or fingertip height) is the one
  that trains into a corner grab.
- **No reference pick task.** `docs/assets/*_demo.mp4` shows reach-style
  policies. Nothing in `examples/` gets an SO-101 from a cube on a table to a
  cube in the air.
[...]

PR #2685, opened 2026-08-24
examples: a reference SO-101 pick that lifts the cube

cagataycali:
## What

Adds `examples/18_so101_pick_and_lift.py` - a minimal, always-works **reference pick** that gets an SO-101 from a cube on the table to a cube in the air, using only the public simulation surface. Closes the gap in #2145 ("Nothing in `examples/` gets an SO-101 from a cube on a table to a cube in the air") and demonstrates the route past the grasp reachability problem in #2167.

SO-101 reference pick

*Rollout in MuJoCo (headless, `MUJOCO_GL=egl`). Cube lifts ~150 mm.*

## Why a grasp-assist weld, not a friction pinch

A two-finger **friction** grasp does not hold the cube on the shipped `so101` model - the gripper reaches and closes on the cube, but it is not lifted. Measured three independent ways at `upstream/main` (`5c878128`, mujoco 3.10.0):

| path | result |
|---|---|
| scripted pinch, cube placed in the pad channel, jaw closed | 4 gripper-cube contacts, **0.0 mm lift** (3 cube sizes) |
| `examples/so101_curobo` scripted physical-grasp path, registry `so101` | `success=False`, cube displacement **6 mm** (3 episodes) |
| RL study, 6-dim action, reward = levelness x face-align | **0% success** over 1.5M steps (#2145) |

The cause is the gripper's convex-hull collision geometry (#2167): the advertised tool site `so101/gripper` sits ~2 mm inside the static jaw's hull boundary and the free channel between the pads is offset from it, so no reward tuning of the pinch makes the cube lift.

The **kinematics are not the blocker** - 5-DOF DLS-IK on `so101/gripper` reaches the grasp pose to < 0.01 mm. What was missing is a reference composing the public primitives into a lift. This is it:

```
move_to(above) -> set_gripper(open) -> move_to(cube) -> set_gripper(close)
  -> attach_bodies(mode="weld") -> move_to(lift) -> detach_bodies()
```

`attach_bodies(mode="weld")` is the library's supported grasp-assist (an equality constraint holding the cube's pose relative to the gripper). The module docstring states the data-honesty caveat verbatim from `attach_bodies`: a welded carry is **not** a physical grasp (no contact force holds the cube), so datasets recorded during one contain idealized transport segments and should be labelled/gated accordingly.

## Test

`tests/test_examples_so101_pick_lifts.py` loads the example by path (leading-digit module) and asserts the composed sequence lifts the cube >= 80 mm. It is headless/CPU-only (`run_pick()` renders nothing unless a video path is passed).

**Non-vacuous:** with the `attach_bodies` weld removed, the identical `move_to`/`set_gripper` sequence lifts **-0.7 mm** (cube stays on the table) vs **+150.6 mm** with it - so the test catches a "reference pick that does not pick" regression, not just an import.

## Local gate

- `ruff check` / `ruff format --check` clean on `strands_robots tests tests_integ` (CI scope; `examples/` is out of scope but the file is ruff-clean and formatted anyway).
- `mypy` clean on the new test.
- `MUJOCO_GL=egl pytest tests/test_examples_so101_pick_lifts.py` -> 1 passed.

A changelog fragment is added as a follow-up commit once this PR has a number.

### DIFF
--- a/changelog.d/2685-so101-reference-pick.md
+++ b/changelog.d/2685-so101-reference-pick.md
@@ -1,0 +1,17 @@
+### Added: `examples/18_so101_pick_and_lift.py` - a reference SO-101 pick that lifts the cube
+
+There was no example that gets an SO-101 from a cube on the table to a cube in
+the air. A two-finger friction grasp does not hold the cube on the shipped
+`so101` model - the gripper reaches and closes on it but does not lift it,
+because of the gripper's convex-hull collision geometry (the advertised tool
+site `so101/gripper` sits ~2 mm inside the static jaw's hull and the free
+channel between the pads is offset from it). The kinematics are not the blocker
+(5-DOF IK reaches the grasp pose to <0.01 mm); a reference composing the public
+primitives into a lift was.
+
+The new example does that with the public surface only - `move_to`,
+`set_gripper`, `attach_bodies(mode="weld")` (the supported grasp-assist),
+`detach_bodies` - and lifts the cube ~150 mm. It states the data-honesty caveat
+(a welded carry is not a physical grasp) and its regression test asserts the
+composed sequence lifts the cube, so a change that silently stopped lifting
+fails rather than shipping a "reference pick" that does not pick.
--- a/examples/18_so101_pick_and_lift.py
+++ b/examples/18_so101_pick_and_lift.py
@@ -1,0 +1,171 @@
+#!/usr/bin/env python3
+"""Get an SO-101 from a cube on the table to a cube in the air - no hardware, no RL.
+
+Goal: a minimal, always-works *reference pick* for the SO-101 in MuJoCo, using
+only the public simulation surface. It reaches the cube with :meth:`move_to`,
+closes the fingers with :meth:`set_gripper`, holds the cube with the supported
+grasp-assist primitive :meth:`attach_bodies` (``mode="weld"``), lifts it, and
+places it down again with :meth:`detach_bodies`.
+
+Why grasp-assist and not a friction pinch
+-----------------------------------------
+A two-finger *friction* grasp does not hold the cube on the shipped ``so101``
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Adds `examples/18_so101_pick_and_lift.py`, a scripted reference pick for the SO-101 in MuJoCo that composes only public simulation primitives (`move_to` -> `set_gripper` -> `attach_bodies(mode="weld")` -> `move_to` -> `detach_bodies`) to lift a cube ~150 mm, plus a headless regression test and a changelog fragment. I verified the example against the current simulation surface: `add_object`'s `size` is documented as full extents (so the 25 mm cube resting at z = 0.0125 is physically correct), `so101/gripper` resolves as a body for `attach_bodies`, the `get_body_state` / `render` result shapes match the indexing used, `move_to(tol=)` / `set_gripper(state=)` / `add_camera` / `step` signatures all match, and `Robot("so101", mesh=False)` in default sim mode already creates the world and adds the arm, so the example correctly does not repeat that. The weld-not-a-grasp data-honesty caveat from the `attach_bodies` docstring is restated in the module docstring, and the test is non-vacuous (the same sequence without the weld lifts ~0 mm, per the PR's measurement). No public API, wire-format, or persisted-schema changes; nothing security-sensitive.

## What's good

- The regression test asserts behaviour (>= 80 mm lift), not implementation, and loads the leading-digit module by path with `pytest.importorskip("mujoco")` - headless and CPU-only.
- Changelog lands as a `changelog.d/2685-*.md` fragment per AGENTS.md step 3, not an edit to `CHANGELOG.md`.
- ASCII-only user-facing strings, no host paths, no module-scope windowed GL default.
- The friction-grasp-doesn't-hold claim is backed by three independent measurements and cross-referenced to #2145 / #2167 rather than asserted.

PR #2686, opened 2026-08-24
fix(isaac): read a robot link's rotation from its MJCF orientation spelling

cagataycali:
## Problem

`load_mjcf` reads each `<body>`'s `pos` into `BodyDef.position` but read **no rotation at all**, so `BodyDef.orientation` kept its identity default for every link it reported. A robot whose links the model rotates was reported upright, with the load reporting success.

Identity is a valid orientation, so nothing downstream can tell a link the model never rotates from one whose rotation was dropped. Half of one pose was read and half was silently discarded.

The sibling reader in the same module, `load_mjcf_scene_objects`, already resolves all five of MJCF's mutually exclusive spellings -- `quat`, `euler`, `axisangle`, `xyaxes`, `zaxis` -- through `_parse_orientation`, under the model-global `<compiler angle>` and `<compiler eulerseq>`. One module, one format feature, two readings; the scene reader's is the one MuJoCo compiles.

Measured on a five-spelling fixture against `mujoco.MjModel.body_quat`, `load_mjcf` disagreed on **6 of 6** bodies -- `quat` included, because the reader never asked about orientation at all:

| body declares | `load_mjcf` reported | MuJoCo compiles |
| --- | --- | --- |
| `quat="0.7071 0 0 0.7071"` | `(1, 0, 0, 0)` | `(0.707107, 0, 0, 0.707107)` |
| `euler="0 0 90"` | `(1, 0, 0, 0)` | `(0.707107, 0, 0, 0.707107)` |
| `axisangle="0 0 1 90"` | `(1, 0, 0, 0)` | `(0.707107, 0, 0, 0.707107)` |
| `zaxis="1 0 0"` | `(1, 0, 0, 0)` | `(0.707107, 0, 0.707107, 0)` |
| `xyaxes="0 1 0 -1 0 0"` | `(1, 0, 0, 0)` | `(0.707107, 0, 0, 0.707107)` |
| nested `euler="0 90 0"` | `(1, 0, 0, 0)` | `(0.707107, 0, 0.707107, 0)` |

## Change

The robot-link walk resolves the rotation through `_parse_orientation`, in the parent's frame -- the frame `position` is already reported in, and the frame `body_quat` stores. `<compiler angle>` and `<compiler eulerseq>` are read from the spliced model, so a robot inheriting `angle="radian"` from an `<include>` is not read as degrees (that misreading is a rotation wrong by a factor of 57). 13 lines of source.

A body declaring two spellings is now refused, as MuJoCo refuses such a model outright: there is no rotation to pick.

## Blast radius, graded against the compiler

Every shipped MJCF asset, both readings compared to `mujoco.MjModel.body_quat`:

| reading | links graded | agrees with MuJoCo | disagrees |
| --- | --- | --- | --- |
| `upstream/main` | 8502 | 5474 | 3028 |
| this branch | 8502 | **7692** | 810 |

**0 verdict changes** (765 load / 302 refuse on both readings) and **0 links moved away from the compiler**. The 3030 changed values span 307 assets; `reachy_mini` alone had 16 of its 17 links misreported.

All **810 of 810** remaining disagreements are one separate cause: a non-unit `quat` such as `quat="1 -1 0 0"`, which MuJoCo normalizes and `_parse_orientation` reports as written. That choice is shared with the scene reader (already reporting 18 non-unit quats on `main`) rather than local to this reader, so reading the spelling and normalizing the result are separate questions with separate blast radii. This change stops at the first, and `TestTheNormalizationBoundaryIsUnchanged` pins the boundary so it cannot move silently.

## Artifact

Each panel renders **only what the loader returned**: one proxy geom per reported `BodyDef`, nested per the model's own hierarchy, at the reported pose. Geometry and camera are identical across panels, so every visible difference is the per-body rotation. The third panel is built from `model.body_quat` -- the oracle itself.

load_mjcf reports a robot link's rotation

The branch's render is **pixel-identical to MuJoCo's own** (`np.array_equal` True, mean |delta| `0.000000`); `main` sits 16.45 mean levels away. In-figure control: subject pixels 28072 / 35107 / 35107 -- branch and oracle equal, `main` different.

## Tests

`tests/simulation/isaac/test_mjcf_robot_body_orientation.py`, 38 tests. Every expectation is derived from `mujoco.MjModel`, so no expected quaternion is restated by hand, and each fixture premise-asserts that its rotation is far from identity -- a reader that fell back to identity cannot pass by coincidence.

Pre-fix split with the source reverted: **28 failed / 10 passed**, class-clean -- all 28 in the regression and boundary classes, **0** in `TestTheReportedRotationIsNotWidened`.

Break table, 7 breaks and 7 distinct firing sets:

| break | tests firing |
| --- | --- |
| drop the `BodyDef` field (rotation computed, not reported) | 27 |
| read only `quat` | 24 |
| assume degrees, ignoring `<compiler angle>` | 8 |
| read `<compiler>` from the top file only, not the spliced model | **1** (the `<include>` case alone) |
| over-reach: normalize the reported quaternion | **2** (the boundary class alone) |
| over-reach: report the absolute frame, not the parent-relative one | 6 |
| skip the mutual-exclusivity check | **1** (the refusal case alone) |

The two over-reach rows are what make the scope defensible: normalizing, or composing the parent's rotation in, each breaks a test that states why this reader does neither.

## Gate

- `ruff check` / `ruff format --check` clean on `strands_robots tests tests_integ`.
- `mypy`: **24 errors before, 24 after**, message sets byte-identical, none in the changed files.
- Regression selection of 130 files touching these loaders, run on both readings with the new file excluded from each: **`5964 passed, 71 skipped`, zero failures either side**, no branch-only or base-only failure.
- The new file: **38 passed on mujoco 3.12.0 and on 3.5.0**, the declared floor.

### DIFF
--- a/changelog.d/2686-mjcf-robot-link-orientation.md
+++ b/changelog.d/2686-mjcf-robot-link-orientation.md
@@ -1,0 +1,26 @@
+### Fixed: a robot link's rotation is read from whichever MJCF spelling declares it, instead of being reported as identity
+
+`load_mjcf` read each `<body>`'s `pos` into `BodyDef.position` but read no rotation at
+all, so `BodyDef.orientation` kept its identity default for every link it reported. A
+robot whose links the model rotates was reported upright with the load reporting
+success, and identity is a valid orientation, so no caller could tell a link the model
+never rotates from one whose rotation was dropped -- half of one pose was read and half
+was silently discarded.
+
+The sibling reader in the same module, `load_mjcf_scene_objects`, already resolves all
+five of MJCF's mutually exclusive spellings (`quat`, `euler`, `axisangle`, `xyaxes` and
+`zaxis`) through `_parse_orientation`, under the model-global `<compiler angle>` and
+`<compiler eulerseq>`. The robot-link walk now reads the rotation the same way, in the
+parent's frame -- the frame `position` is already reported in, and the frame MuJoCo's
+`body_quat` stores, so nested links resolve too. `<compiler>` is read from the spliced
+model, so a robot inheriting `angle="radian"` from an `<include>` is not read as
+degrees. A body declaring two spellings is refused, as MuJoCo refuses such a model
+outright.
+
+Graded against `mujoco.MjModel.body_quat` across the shipped asset corpus, agreement
+rises from 5474 of 8502 links to 7692, with no verdict changes, no link moving away
+from the compiler, and 3030 corrected values spanning 307 assets. The 810 links that
+still differ are all non-unit `quat` declarations, which MuJoCo normalizes and
+`_parse_orientation` reports as written; that choice is shared with the scene reader
+rather than local to this one, so it is left unchanged here and pinned by a test so it
+cannot move silently.
--- a/docs/simulation/isaac.md
+++ b/docs/simulation/isaac.md
@@ -211,7 +211,10 @@ open a window.
   camera prims are parented to the stage camera scope rather than to an
   articulation link.
 - **Loaders** - `load_urdf` / `load_mjcf` / `load_usd` resolve to a
-  `ProceduralRobot` dataclass.
+  `ProceduralRobot` dataclass. `load_mjcf` reports each link's pose in its
+  parent's frame, reading the rotation from whichever of MJCF's five spellings
+  the body uses - `quat`, `euler`, `axisangle`, `xyaxes` or `zaxis` - under the
+  model-global `<compiler angle>` and `<compiler eulerseq>`.
 
 Because the joint-name and observation contract matches the MuJoCo backend,
 policies and observation mappings transfer unchanged between backends.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -378,6 +378,14 @@ def load_mjcf(path: str) -> ProceduralRobot:
     parent. Useful for LIBERO scenes (the matrix's main consumer ships
     MJCF).
 
+    Each body's pose is reported in its parent's frame, as MJCF declares it:
+    ``position`` from ``pos``, and ``orientation`` from whichever of MJCF's five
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

`load_mjcf`'s robot-link walk read each `<body>`'s `pos` but never its rotation, so `BodyDef.orientation` stayed at its identity default for every reported link. This PR routes the rotation through the module's existing `_parse_orientation` -- the same resolver `load_mjcf_scene_objects` already uses for all five of MJCF's mutually exclusive spellings -- with `<compiler angle>`/`<compiler eulerseq>` read from the spliced model via `_mjcf_angle_units(_mjcf_model_toplevel(...))`, so an `angle="radian"` inherited from an `<include>` is honoured. The source delta is 13 lines that delegate to already-shipped helpers; `BodyDef.orientation` is a pre-existing dataclass field, so no public API shape changes. The one behaviour change beyond the fix -- refusing a body that declares two orientation spellings -- mirrors MuJoCo's own compile-time refusal, raises a documented `ValueError` (the `Raises:` block is updated), and the corpus grading reports zero load/refuse verdict changes.

## What's good

- Every test expectation is derived from `mujoco.MjModel.body_quat` rather than hand-restated quaternions, and each fixture premise-asserts its rotation is far from identity, so a fall-back-to-identity reader cannot pass by coincidence.
- `TestTheNormalizationBoundaryIsUnchanged` and `TestTheReportedRotationIsNotWidened` pin the scope in both directions -- the non-unit-quat boundary stays shared with the scene reader instead of moving silently here.
- Changelog fragment under `changelog.d/`, docstring `Raises:` updated for the new refusal, behaviour-named test classes, ASCII-only strings, no host paths -- AGENTS.md conventions all followed.
- Independently verified: the 38 new tests pass against the head tree on mujoco 3.8.0 (inside the declared 3.5.0-3.12.0 range), complementing the PR's own runs at both ends.

PR #2687, opened 2026-08-24
fix(scripts): an answered review thread owes its resolve

### LINKED ISSUE #2682: An answered review thread is still a merge blocker until it is resolved, and the sweep that answers "which of my PRs need me" reports nothing owed
## What happened

A scheduled cycle read #2680 as waiting on a reviewer and stopped. It was blocked on the author. The whole cycle was spent reaching that wrong conclusion, and the next one had to reach the right one.

The pull request's one review thread carried a `[MUST FIX]`, the fix had landed in `0fd0f4e3`, the reply explaining it was posted, and CI was green. So the authorship gate in AGENTS.md step 5 fired correctly - the thread's last comment was the author's, therefore do not reply - and that gate is about *replying*. Nothing carried it on to the thread's other terminal state, which is the resolve.

## The two sweeps disagree, and the reassuring one is the one step 5 sends you to

Both run against the same head, `0fd0f4e3`, minutes apart:

| sweep | outcome for #2680 | exit |
|---|---|---|
| `check_thread_is_answered.py --all-open` | `nothing-owed`, 1 answered, 0 awaiting | **0** |
| `check_merge_blockers.py --all-open` | `unresolved-threads`, **owed by the author** | 1 |

AGENTS.md step 5 introduces the first as the way to "ask for the verdict rather than re-deriving it", and says of it: "`settled` and `answered` are not work. `awaiting-the-author` is, and it is the only outcome that exits 1 - so the sweep answers 'which of my open pull requests actually need me' in one read."

On this pull request that sentence is false, and false in the reassuring direction. `answered` was accurate - the thread *was* answered - and work was owed anyway, because `required_review_thread_resolution` is a ruleset rule and an answered thread is not a resolved one. A sweep whose stated contract is "which of my pull requests need me" returned exit 0 over a pull request that could not merge until its author acted.

Resolving it moved the merge blocker exactly as step 8 documents for #1974 and #2566:

| | before the resolve | after |
|---|---|---|
| unsatisfied rule(s) | `unresolved-threads`, `missing-approval` | `missing-approval` |
| owed by | **the author** | any reviewer |
| `reviewDecision` | `null` | `REVIEW_REQUIRED` |

The `null` is the #2328 half rather than the #1974 half - every review on #2680 is `COMMENTED`, so the resolve was necessary and not sufficient, and a first approval is still owed. That part is the ordinary state and is not this issue.

## Why it recurs rather than being one bad cycle

The two facts an agent needs live in two different sweeps, and the cheaper, more prominent one is scoped to replying. AGENTS.md already says "Resolve threads as you fix them" under Reviewing & Iteration, and step 8 already says "Resolve rather than push" - but both are written for the cycle that *lands* the fix. A later cycle rebuilding its context sees an answered thread and a green branch, asks the sweep step 5 names, is told nothing is owed, and correctly declines to reply. Nothing in that path ever asks whether the thread is resolved.

That is the same shape as the twelve-reply incident the authorship gate was written for, inverted: there the gate was silent on a case that recurs every scheduled cycle and the agent did too much; here it is silent on the adjacent case and the agent does too little. Same cost centre either way - the pull request sits, and the sweep reports it as reviewer bandwidth, which is the presentation #1905 and #1917 each record for their own cause.

## What would resolve it

One of, in rough order of preference:

1. **Have `check_thread_is_answered.py` report the resolve as owed.** An answered-but-unresolved thread on a pull request whose author is the viewer is author-owed work, so it belongs in the outcome that exits 1 - or in a new one beside it. This puts the answer in the sweep whose contract already claims to give it, and needs no reader to remember a second command.
2. **Say so in step 5.** The authorship gate's three bullets decide whether to *reply*; a fourth line noting that a thread you have answered still gates the merge until it is resolved, with `check_merge_blockers.py` as the read that sees it, closes the gap in prose.
3. Both. They are cheap and they fail differently: the sweep change helps an agent that runs the command, the prose helps one that reasons from the file.

What is not sufficient is restating "resolve threads as you fix them" more firmly. The cycle that missed this was not the cycle that fixed anything - it was a later one, reading a thread it had not written, and the rule as written does not address it.

## Evidence

- #2680, thread `PRRT_kwDORUMiZs6bmZqs`, head `0fd0f4e3`, `statusCheckRollup` `SUCCESS` throughout.
- Both sweeps above run from `scripts/` at `main`, same token, minutes apart.
- The cycle that read it as reviewer-owed left no artifact by design, which is part of why this is worth pinning: a wrong "nothing owed" is indistinguishable from a correct one afterwards.

cagataycali:
Fixes #2682.

## The defect

`check_thread_is_answered.py` classifies a thread whose last non-bot comment is the author's as `answered`, and reports it as not work. That is true of the *reply* and false of the *merge*.

`ANSWERED` is chosen only after `isResolved` has been ruled out, so **every thread carrying it is answered *and* unresolved** - and `required_review_thread_resolution` is a branch ruleset rule. So the one outcome the sweep called "nothing to do" is precisely the state that blocks the merge and that only the author can clear.

Measured on #2680 at head `0fd0f4e3`, both sweeps run minutes apart from `scripts/` with the same token:

| sweep | outcome for #2680 | exit |
|---|---|---|
| `check_thread_is_answered.py --all-open` | `nothing-owed`, 1 answered, 0 awaiting | **0** |
| `check_merge_blockers.py --all-open` | `unresolved-threads`, owed by the author | **1** |

A sweep whose stated contract is "which of my open pull requests actually need me" cleared a pull request that could not merge until its author acted. The cost was a whole scheduled cycle spent reaching the wrong conclusion, and the next one had to reach the right one.

## The fix

Replaying the #2680 payload through both versions of `evaluate()`:

| | thread outcome | pull request outcome | `is_finding` |
|---|---|---|---|
| before | `answered` | `nothing-owed` | `False` |
| after | `answered` | `author-owes-a-resolve` | `True` |

The thread taxonomy already isolated the right state, so nothing about *classification* changes. What changes is whether that state is work:

- `Thread.owes_a_resolve` sits beside `Thread.is_owed`. Kept separate on purpose: the remedy for a thread awaiting a reply is words, and the remedy for an answered thread is emphatically not.
- Per-pull-request outcome gains `author-owes-a-resolve`. `author-owes-a-reply` outranks it, because a thread still owed an answer is not yet ready to be resolved - but **both exit 1**, because both are the author's.
- `settled` is now the only outcome that is not work.

This is issue option 1 **and** option 2 ("Both. They are cheap and they fail differently"): AGENTS.md step 5 gains the matching fourth bullet and drops the sentence that was false in the reassuring direction, so an agent reasoning from the file reaches the same answer as one running the command.

## The risk this change carries, and what holds it down

Making an answered thread "work" is one step from re-licensing the twelve-reply incident the check was written to prevent. So the remedy text says both things at once - resolve it, **and do not reply to it** - and that wording is pinned rather than left to prose:

```
test_the_remedy_for_an_answered_thread_is_the_resolve_and_forbids_a_reply
```

## Verification

- `tests/test_thread_is_answered.py`: **46 passed**. Four existing tests encoded the old contract and flip deliberately (the remedy gate, the single-report outcome, the sweep's finding set, and the exit-status matrix); the `answered` incident pins for #2577/#2511 keep asserting `is_owed is False` and now also assert `owes_a_resolve is True`, so the reply is still refused at exactly the sites that measured it.
- New pins: the #2680 regression, non-vacuity of the reply/resolve split, reply-outranks-resolve, attribution to another author's pull request, and the remedy wording above.
- `ruff check` / `ruff format --check` / `mypy` clean on both changed files.
- Live `--all-open` against this repository, before and after: `nothing-owed` on all 4 open pull requests, exit 0. **This fix adds no finding to today's state** - 4 settled threads on #1722, none elsewhere - which is why coverage comes from replayed payloads, as the module docstring already argues it must.

## Scope

Four files: the script, its pins, the AGENTS.md step it contradicted, and a changelog fragment. No behaviour outside this sweep changes, and neither script gates CI.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -291,8 +291,8 @@ hatch run format            # ruff check --fix, ruff format
    - Thread's last non-bot comment is **yours** -> do not reply. You have
      already said it. If there is code to push, push it; the push is the
      message.
-   - Thread is **`isResolved`** -> do not reply. Resolution is a reviewer action
-     and terminal. Reopening it to restate a landed fix reads as noise, not
+   - Thread is **`isResolved`** -> do not reply. Resolution is terminal, whoever
+     performed it. Reopening it to restate a landed fix reads as noise, not
      diligence. `isOutdated` is **not** terminal, measured: it describes the diff
      rather than the conversation, and a thread keeps accepting comments after it
      flips - #2577's took two more after `d04a8969` moved its lines. It is also
@@ -302,6 +302,14 @@ hatch run format            # ruff check --fix, ruff format
      filed as settled.
    - Last comment is **someone else's** and your existing replies do not answer
      it -> reply once, then resolve.
+   - Thread you have **answered** and nobody has resolved -> resolve it, and still
+     do not reply. The three bullets above decide whether to *speak*, and none of
+     them reaches the thread's other terminal state.
+     `required_review_thread_resolution` is a branch ruleset rule, so a thread you
+     have answered goes on gating the merge until it is resolved, and no reviewer
+     can clear that for you. #2682 measured the cost: a cycle read #2680 as waiting
+     on a reviewer and stopped, when the fix had landed in `0fd0f4e3`, the reply was
+     posted, CI was green, and the only thing outstanding was the resolve.
 
    The authorship check is the cheap one, and it would have prevented ten of
    those twelve comments on its own: no semantic comparison, just the author of
@@ -320,13 +328,17 @@ hatch run format            # ruff check --fix, ruff format
    python3 scripts/check_thread_is_answered.py --repo strands-labs/robots --all-open
    ```
 
-   `settled` and `answered` are not work. `awaiting-the-author` is, and it is the
-   only outcome that exits 1 - so the sweep answers "which of my open pull
-   requests actually need me" in one read. The report also names the commit each
-   thread was written against beside the pull request's head, which is what tells
-   "already fixed at `<oid>`" apart from "not yet fixed" without re-deriving the
-   fix (#2520). It reports and does not gate: what an author should do next is not
-   something a branch can turn green.
+   `settled` is not work. `awaiting-the-author` is a reply, `answered` is the
+   resolve that answer still owes, and **both** exit 1 - so the sweep answers
+   "which of my open pull requests actually need me" in one read. It did not
+   always, and the gap was in the reassuring direction: #2682 measured
+   `nothing-owed` and exit 0 over #2680, at the same head where
+   `check_merge_blockers.py` read `unresolved-threads` owed by the author.
+
+   The report also names the commit each thread was written against beside the
+   pull request's head, which is what tells "already fixed at `<oid>`" apart from
+   "not yet fixed" without re-deriving the fix (#2520). It reports and does not
+   gate: what an author should do next is not something a branch can turn green.
 
    What makes this worth writing down is that the previous rule was *satisfied*
    by all twelve. "Address all review comments", and "reply when a thread asks
--- a/changelog.d/2682-answered-thread-owes-the-resolve.md
+++ b/changelog.d/2682-answered-thread-owes-the-resolve.md
@@ -1,0 +1,22 @@
+### Fixed: an answered review thread is now reported as owing its resolve
+
+`check_thread_is_answered.py` classified a thread whose last non-bot comment is the
+author's as `answered` and called it "not work". That is true of the *reply* and
+false of the merge: `answered` is chosen only after `isResolved` has been ruled out,
+so every thread carrying it is answered **and** unresolved, and
+`required_review_thread_resolution` is a branch ruleset rule. Measured on #2680 at
+head `0fd0f4e3`, minutes apart with the same token: this sweep reported
+`nothing-owed` and exited 0, while `check_merge_blockers.py` reported
+`unresolved-threads` owed by the author. A sweep whose stated contract is "which of
+my open pull requests actually need me" cleared a pull request that could not merge
+until its author acted, and the cost was a whole scheduled cycle spent reaching the
+wrong conclusion.
+
+An answered-but-unresolved thread is now author-owed work: it exits 1, and the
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

This PR closes the gap #2682 measured: `check_thread_is_answered.py` treated an `answered` thread as "not work", but `answered` is by construction answered-and-unresolved, and `required_review_thread_resolution` is a branch ruleset rule -- so the sweep's `nothing-owed` / exit 0 cleared a pull request (#2680) that could not merge until its author resolved the thread. The fix keeps the thread taxonomy untouched and changes only what counts as work: `Thread.owes_a_resolve` sits beside `is_owed` (deliberately separate so the two remedies cannot collapse), the per-PR outcome gains `author-owes-a-resolve` with `author-owes-a-reply` outranking it, and both exit 1. The obvious regression risk -- re-licensing the twelve-reply loop the check exists to prevent -- is handled directly: the remedy text says "resolve it, and do **not** reply to it" and that wording is pinned by `test_the_remedy_for_an_answered_thread_is_the_resolve_and_forbids_a_reply`. Verified locally against the head files: 46 tests pass, ruff clean, no non-ASCII in user-facing strings, and no workflow consumes this script's exit code, so the exit-status change gates nothing in CI.

## What's good

- The #2680 regression is pinned with a replayed payload, matching the module's own argument for why coverage cannot come from a historical sweep.
- The four deliberately-flipped tests (remedy gate, single-report outcome, sweep detail set, exit-status matrix) each document the flip with a #2682 reference, so the old contract cannot be restored by accident.
- AGENTS.md step 5 gains the matching fourth bullet and drops the sentence that was false in the reassuring direction, keeping the file and the command in agreement -- the same documented-and-enforced pairing the repo requires elsewhere.
- Changelog fragment present (`changelog.d/2682-answered-thread-owes-the-resolve.md`), closing keyword in the PR body (links #2682), scope disciplined to the sweep and its documentation.
- The sweep table's `answered` column now reads `len(row.unresolved)`, which is value-identical to the old `sum(outcome == ANSWERED)` -- no silent meaning change in the rendered report.

PR #2142, opened 2026-08-11
fix(mesh): expose a stop tool on the ROS 2 bridge

cagataycali:
## Problem

A ROS 2 velocity command latches. `RosBridgedRobot.drive()` with no `duration`
publishes a **single** `cmd_vel` message and the base holds it until something
else is published — the contract `RosbridgeRobot`'s own drive tool description
states as *"without duration the last command latches until stop"*.

All three mobile-base bridges carry a public `stop()` that publishes a zero
`Twist`. Two of the three also exposed it as a named agent tool. The ROS 2
bridge did not:

| transport | `stop()` method | `stop_<node>` tool |
|---|---|---|
| `RosBridgedRobot` (rclpy) | yes | **missing** |
| `RosbridgeRobot` (websocket) | yes | yes |
| `RtpsRobot` (cyclonedds) | yes | yes |

So an agent handed `RosBridgedRobot.tools` could start motion and had no tool to
end it. Measured on one session — `drive_rover(linear=0.5)` publishes one
message at `linear.x = 0.5` and the surface offers `{drive_rover,
get_pose_rover}`. The only halt available was `drive_rover(linear=0, angular=0)`,
i.e. the caller having to know that a zero-velocity drive is the halt idiom,
which is exactly what `stop()` exists to name.

`docs/ros2-integration.md` already lists `stop()` as a first-class capability
and its example comment enumerated the tools as `drive_turtlesim,
get_pose_turtlesim, ...`.

## Change

`tools` now builds `stop_<node>` unconditionally, next to `drive`, so the three
transports agree on the capability an agent is given. The property docstring
records why the pair is unconditional (unlike `get_scan` and `navigate`, which
stay gated on their optional topic/action).

Four lines of behaviour; the rest is the test that would have caught it.

## Why it was invisible

`test_tools_are_named_per_instance` asserted the **exact** set
`{drive_turtlesim, get_pose_turtlesim}`, so the absence was recorded as intent.
It now includes the halt and says so in its docstring.

Underneath that, of the ten forwarding closures these three bridges construct,
**four had never been invoked through the tool** (`get_pose`/`get_scan` on the
ROS 2 bridge, `stop`/`get_scan` on the rosbridge one) — nothing observed which
capabilities the surface actually reached. The new module drives all of them.

## Artifact

One agent session, recorded off the forwarded `use_ros` call (no ROS 2 present).
The drive is byte-identical on both trees; the difference is only what the agent
can reach next.

ROS 2 bridge stop tool

[...]

PR #2689, opened 2026-08-24
feat(mesh): AckermannRosRobot - Ackermann ROS 2 cars (AWS DeepRacer) as strands robots

cagataycali:
## Ackermann ROS 2 cars (AWS DeepRacer) as strands robots, brought current with `main`

Continues @Vivek0712's #1035, whose commits are ancestors of this branch. That
PR is a complete, reviewed feature that has been unmergeable on a textual
conflict since 2026-08-04. This branch carries it forward: `main` merged in,
the conflict resolved, and the two defects that its stale base was hiding
fixed.

### The feature

`RosBridgedRobot` covers differential-drive bases (a `Twist` on `cmd_vel` plus
odometry). An Ackermann car exposes neither. `AckermannRosRobot` keeps the same
`drive(linear, angular, duration)` agent contract and converts through a
bicycle model to normalized servo commands (`ServoCtrlMsg` on the DeepRacer),
runs a declarative `init_services` handshake once before the first command,
clamps speed, refuses over-long holds loudly, and always trails a timed command
with a zero servo message so a timed drive cannot leave the car driving. It
validates through the same `finite_number_error`,
`positive_finite_number_error` and `positive_whole_number_error` domains the
three differential-drive bridges call, so the fourth transport refuses an
unusable velocity, hold or count with byte-identical text.

### The conflict

One file, `docs/ros2-integration.md`, where both sides appended a new section at
the same anchor: this branch a new `##` "Ackermann robots (AWS DeepRacer)"
section, `main` a `###` "Safety-critical command surfaces need operator
approval" subsection. Both are kept. `main`'s block goes first because it is an
`h3` that continues the pre-existing `## Safety` section - putting the new `h2`
first would nest `main`'s subsection under "Ackermann robots". Resulting order
is `## Safety` -> `### Safety-critical ...` -> `## Ackermann robots ...` ->
`## Sim bridge ...`.

### Two defects the stale base hid

The branch's checks were green, and correctly so - both guards that fail it
landed on `main` *after* its base (`c1acf21d`), and neither guard's file is one
the branch touches, so there is no file overlap to notice:

| Guard | Landed on `main` | What it found |
|---|---|---|
| `tests/test_repr_survives_partial_construction.py` | #2408, 2026-08-17 | `AckermannRosRobot.__repr__` raises on a half-built instance |
| `tests/mesh/test_bridge_stop_tool_parity.py` | #2142, 2026-08-11 | the drive/stop survey inventory was not triaged for a fourth bridge |

**1. `__repr__` on a half-built instance.** `__init__` validates `node_name`
through `_check_topic` before assigning it, so a refused construction leaves the
instance the raising frame keeps alive rendering as

```
[AttributeError("'AckermannRosRobot' object has no attribute 'node_name'") raised in repr()]
```

which sends the reader after `node_name` when the real failure is the
`ValueError: invalid node_name: 'bad name!'` they already caused. Now wrapped in
`try/except AttributeError` delegating to `strands_robots.utils.partial_construction_repr`,
exactly as `RosBridgedRobot`, `RosbridgeRobot` and `RtpsRobot` do, so the phrase
a reader learns to recognise in a traceback has one owner. Half-built now renders
`AckermannRosRobot(partially constructed, id=0x...)`. The class is triaged into
both halves of that module's survey - the half-built sweep and the
documented-refusal cases - for parity with the other three bridges.

**2. drive/stop survey inventory.** `test_the_survey_finds_the_bridges_it_is_meant_to_cover`
is the non-vacuity half of the drive/stop guard: it pins that the scan finds the
bridges it is meant to cover, so a mis-rooted or empty survey cannot read as
compliant. A fourth drive/stop bridge has to be triaged there. The rule itself
already covered the class - the survey walks the mesh package - and renaming its
`stop_` tool fails `test_every_bridge_owning_a_drive_stop_pair_builds_a_stop_tool`.
It stays out of `BRIDGES` deliberately, with the reason recorded next to the
inventory: those cases assert a zero `geometry_msgs/msg/Twist` on `/cmd_vel`, and
an Ackermann car halts with a zero `ServoCtrlMsg` (`angle`/`throttle`) on its
servo topic, which `tests/mesh/test_ackermann_robot.py` already pins.

### Verification

Regression split, on the merged tree, over the two guard files:

| tree state | result |
|---|---|
| at the merge commit, before both fixes | **3 failed**, 60 passed |
| after | **65 passed** |

The three are the two `repr` cases plus the inventory case; the delta to 65 is
those three plus the two new parametrized cases the class now contributes.

Each fix is load-bearing, checked by mutation:

| mutation | fails |
|---|---|
| drop the `try/except` from `__repr__` | the half-built case, and the single-owner count (`tolerant == len(EXPECTED_REPR_CLASSES)`) |
| return an inline string instead of `partial_construction_repr` | the new documented-refusal case, and the single-owner check |
| rename the Ackermann `stop_` tool to `halt_` | `test_every_bridge_owning_a_drive_stop_pair_builds_a_stop_tool` |

Zero regressions, over 296 tree-walking guard files plus all of `tests/mesh`,
run identically on this branch and on a pristine `upstream/main` worktree:

| tree | result |
|---|---|
| this branch | 5 failed, **17887** passed, 61 skipped |
| pristine `upstream/main` | 5 failed, **17840** passed, 60 skipped |

The five failing ids are identical on both trees and all environment-bound
(four `DID NOT RAISE ImportError` where `rclpy` is importable, and one
`auth_mode` case sensitive to a mesh env var another test in the selection
sets). The +47 passed is this branch's own suite plus the two new cases.

tests_integ`. `mypy` over the same scope: 19 errors on this branch and 19 on
pristine `main`, with the branch-only difference empty. `assemble_changelog.py
--check` OK.

No visual artifact: the diff adds a ROS 2 transport bridge and a `__repr__`
fallback, and changes no simulation, rendering or policy behaviour.

Supersedes #1035. Its changelog fragment is kept and extended with the `repr`
fix rather than replaced.

---

### Round 2 changes

Two review concerns, both on safety paths in the changed code, one commit each,
plus the rule the second belongs to recorded where the next bridge will read it.

**1. The trailing safety stop's verdict was discarded** (`fcb5ed4`). `drive` sent
its trailing zero servo message from a `finally` and dropped the result, so when
the main publish succeeded and the halt then failed - node or daemon dying
mid-hold - the caller saw the drive's `success` for a car still holding the
commanded throttle, on a tool whose description promises that a command with a
duration stops itself. The halt's result is now captured into a local from the
`finally` (so it still goes out when the main publish raises) and, when the drive
succeeded but the halt did not, returned as an error naming the published
command, the possibly-live throttle, `stop` as the next action, and the halt's
own cause. A failed drive stays the reported cause; a latching single-shot sends
no halt, so there is no verdict to read.

Three cases in `tests/mesh/test_ackermann_robot.py`, next to
`test_drive_trailing_zero_sent_even_when_publish_errors` which covers the inverse
direction: the failed-halt case, the both-failed precedence, and a control that a
single-shot success is not reinterpreted. Restoring the discard fails the first.

**2. The bridge's commands were outside the `use_ros` operator gate** (`43ee3d5`).
Both halves were missing. The gate only fires on a blocklisted surface, and the
shipped DeepRacer wiring matched no entry (`/webserver_pkg/manual_drive`, and the
`/ctrl_pkg` mode services that arm the vehicle), so every servo publish and
handshake call went out with no prompt, no allowlist check and no audit row -
while the functionally identical `/motor_enable` was blocked. And nothing here
carried an operator context, so fixing coverage alone would have made every
command, `stop` included, a fail-closed refusal.

| half | change |
|---|---|
| blocklist coverage | `/manual_drive`, `/vehicle_state`, `/enable_state` are default entries, spelled bare so the base-name rule covers the namespaced spellings |
| context threading | `drive`, `stop`, `enable` take a `tool_context` and thread it into every `use_ros` call (the trailing halt included); `drive_<node>` / `stop_<node>` are `@tool(context=True)` and forward the injected context, as `RosBridgedRobot` does. `get_scan` stays contextless - `echo` is never gated |

Pinned by `tests/mesh/test_ackermann_command_gate.py` (18 cases), which keeps the
real `use_ros` and doubles the rclpy transport instead - the boundary
`test_ackermann_robot.py` patches over, which is why neither half was visible
there. It covers the prompt order over all three surfaces, a decline at the
handshake and at the servo topic, the halt reaching the operator, a timed drive's
trailing halt, both allowlist spellings, the headless refusal naming both
variables, and structural guards on `context=True` and on every command call
site. `TestEveryCommandingMeshBridgeHasAGateSuite` derives the mesh modules that
send a command through `use_ros` from the tree and compares them against an
inventory of gate suites, so a fifth bridge is graded on arrival rather than
shipping ungated the way this one did.

| mutation | fails |
|---|---|
| remove the three blocklist entries | 7 cases (5 behavioural + both coverage guards) |
| drop `context=True` from the drive tool | 2 cases (structural + input-schema) |
| drop one `tool_context` forward in `_publish_servo` | 5 cases |
| restore the discarded halt verdict | 1 case |

Doc drift closed in the same commits, at every surface an operator reads to size
a pre-approval: the blocklist table and the final-path-segment example list in
`docs/ros2-integration.md`, a per-method gated-surface table plus the
`STRANDS_ROS2_COMMAND_ALLOW` line in its Ackermann section, the surface
enumeration and the context-forwarding sentence in `docs/security.md`, the
example's SAFETY block, and the class and module docstrings. The changelog
fragment carries both fixes.

**3. The rule, recorded** (`d44be59`). AGENTS.md convention 18: a new bridge
ships un-gated because the gate is keyed on surface names it does not know, and a
bridge suite that patches the module's `use_ros` symbol cannot see it - so naming
the surfaces and threading the context are one change, either alone being a
different defect. It names the pin above rather than restating the rule, since
that pin's inventory is derived from the tree.

**Round 2 verification.** Same selection on this branch and on a pristine
`main` worktree (`tests/mesh`, `tests/tools/test_use_ros.py`,
`tests/test_use_ros_command_blocklist.py`,
`tests/test_repr_survives_partial_construction.py`,
`tests/test_args_docstring_completeness.py`,
`tests/test_docstring_xref_roles_resolve.py`):

| tree | result |
|---|---|
| this branch | 112 failed, **4412** passed, 18 skipped, 24 errors |
| pristine `main` | 112 failed, **4347** passed, 18 skipped, 24 errors |

The 136 failing/erroring ids are byte-identical between the two trees and all
bound to optional dependencies absent from this runner (90 `json5`, 44 `awsiot`,
1 `lerobot`, and one pre-existing `test_iot_camera_ref_delivery` assertion). The
+65 passed is this branch's new gate suite (18) plus its own earlier cases. The
968-test sweep of every guard that reads AGENTS.md passes (one `lerobot`-bound
failure, present before the edit). `ruff check` and `ruff format --check` clean
over `strands_robots tests`; `mypy` reports zero errors in either changed source
file, and its count on `use_ros.py` is 11 on both trees; `assemble_changelog.py
--check` OK.

---

### Round 3 changes

No review feedback outstanding; this round is a base merge forced by a sibling
change, plus the absorption its sequencing note called for.

`main` landed the ROS 2 command gate as a shared owner,
`strands_robots/tools/_command_gate.py`, so that `use_rtps` and `use_rosbridge`
consult the same blocklist as `use_ros` - a surface refused on one transport can
no longer be sent on another. Round 2 above added three DeepRacer surfaces to
that blocklist and extended the two `docs/security.md` sentences describing it,
while it still lived in `use_ros.py`. Both sides therefore edited the same text,
and this branch went `CONFLICTING` the moment that change merged.

| conflict | resolution |
|---|---|
| `strands_robots/tools/use_ros.py` | main's version wholesale. This branch's only edit to it was the blocklist and its comment, and both now live in the shared module. |
| `docs/security.md`, both sentences | main's widened subject (all three transports, one shared blocklist) plus what this branch added to them. |

The three surfaces move to `_command_gate.COMMAND_BLOCKLIST` unchanged, which
**widens** what round 2 achieved: `/manual_drive`, `/vehicle_state` and
`/enable_state` are now gated on all three transports rather than only on
in-process rclpy. Their justification moves with them, minus the
bare-entry/final-segment rule the shared comment already states. The two
`docs/security.md` sentences regain the Ackermann servo topic and arming
services in the enumeration, and `AckermannRosRobot` beside `RosBridgedRobot` as
a context forwarder. `docs/ros2-integration.md` merged with no conflict and
carries both the transport axis and this branch's surface table.

Three symbols this branch referenced moved with the gate, so its own suite and
the rule recorded in round 2 name them at the new owner:

| was | is |
|---|---|
| `use_ros._DEFAULT_COMMAND_BLOCKLIST` | `_command_gate.COMMAND_BLOCKLIST` |
| `use_ros._is_command_blocked` | `_command_gate.command_block_message` |
| `use_ros._COMMAND_ALLOW_ENV` | `_command_gate.COMMAND_ALLOW_ENV` |

A tree-wide sweep confirms no reference to any of the old names survives, in
source, tests or AGENTS.md.

**Round 3 verification.** The 12 suites covering the gate, the Ackermann bridge
and both sibling transports: **545 passed, 1 skipped**. The whole-tree graders
that read the edited docs and AGENTS.md: **111 passed**.

Full `tests/` on the composition and on a pristine `main` worktree, same
interpreter and same environment:

| tree | result |
|---|---|
| this branch merged with `main` | 627 failed, **27511** passed, 1871 skipped, 981 errors |
| pristine `main` | 627 failed, **27443** passed, 1870 skipped, 981 errors |

Read as a delta, per the rule that an absolute count in a partial environment
invents regressions: the failing-and-erroring id sets are compared directly and
the **branch-only set is empty**, with identical 627/981 totals on both trees.
Every one is an absent optional dependency here (`torch`, `mujoco`, `lerobot`,
`rclpy`, `device_connect_edge`, `json5`, `awsiot`). The +68 passed is this
branch's own tests.

`ruff check` and `ruff format --check` clean over `strands_robots tests`
(1469 files already formatted).


strands_robots/mesh/ackermann_robot.py
```diff
@@ -0,0 +1,369 @@
...
+
+        Tools are bound to this instance and suffixed with ``node_name`` so
+        multiple robots coexist in one ``Agent(tools=[...])`` call. The drive
+        tool's description states the Ackermann kinematic limit (minimum
+        turning radius) so the agent can plan paths the platform can follow.
+        """
+        suffix = self.node_name.strip("/").replace("/", "_")
+        min_radius = self.wheelbase_m / math.tan(self.max_steering_rad)
+
+        @tool(
```
yinsong1986, line 409: [MUST FIX] This bridge's motion commands are entirely outside the `use_ros` operator-approval gate, and these tools cannot carry the context that gate needs.

Two halves, both missing:

1. **No blocklist coverage.** `_gate_command` in `strands_robots/tools/use_ros.py` only fires for surfaces matching `_DEFAULT_COMMAND_BLOCKLIST` by exact or base name. The DeepRacer wiring's surfaces - `/webserver_pkg/manual_drive` (base `/manual_drive`), `/ctrl_pkg/vehicle_state` (base `/vehicle_state`), `/ctrl_pkg/enable_state` (base `/enable_state`) - match no entry, so every publish and handshake service call this bridge issues proceeds with no prompt, no allowlist check, and no audit row. The gate's own comment says the motor-enable services are "its most dangerous entries"; `/motor_enable` and `/enable_motor` are blocked while the functionally identical `/enable_state` is not.
2. **No context threading.** `drive`/`stop` here are plain `@tool(...)` with no `context=True`, and `drive()`/`stop()`/`enable()` call `use_ros` without `tool_context` - so even after a blocklist fix, these tools would inherit the fail-closed path (headless refusal) instead of prompting. Compare `RosBridgedRobot.tools`, which declares `context=True` on `drive`/`stop`/`navigate` and forwards the context precisely so "its operator-approval gate prompts rather than failing closed".

Why this can't be deferred: it is a bypass of the repository's documented security control for LLM-initiated physical motion. `docs/ros2-integration.md` - in the "Safety-critical command surfaces need operator approval" section this PR's merge places directly above the new Ackermann section - states "Anything that wraps `use_ros` has to forward that context or it inherits the fail-closed path for every command it sends." This class wraps `use_ros`, forwards nothing, and ships with an example (`examples/ros2/deepracer_agent.py`) that has an agent drive a physical car's throttle with zero human approval on the same transport where the sibling bridge's `/cmd_vel` prompts. Once released, operators deploying this bridge get an ungated actuation surface by default; tightening it afterwards is the harder change (see AGENTS.md > Review Learnings (PR-7 - robot_mesh HITL patterns)).

Resolution: (a) add the DeepRacer motion surfaces to `_DEFAULT_COMMAND_BLOCKLIST` (base-name matching means `/manual_drive`, `/vehicle_state`, `/enable_state` entries cover namespaced instances); (b) declare the `drive_`/`stop_` tools `@tool(context=True)` and thread `tool_context: ToolContext | None = None` through `drive()`, `stop()` and `enable()` into every `use_ros` call, exactly as `RosBridgedRobot` does. Programmatic/headless use then works via `STRANDS_ROS2_COMMAND_ALLOW`, matching the documented posture for the other `use_ros` bridge.

yinsong1986 (commented):
## Summary

Carries #1035 forward onto current `main`: a fourth mobile-base transport, `AckermannRosRobot`, presenting Ackermann-steering ROS 2 cars (AWS DeepRacer reference wiring in `from_deepracer()`) with the fleet-standard `drive(linear, angular, duration)` contract, a bicycle-model conversion to normalized `ServoCtrlMsg` servo pairs, a declarative `init_services` enable handshake, and a trailing zero-servo message after timed/multi-message commands so a hold cannot leave the throttle latched. The merge also fixes the two guard failures the stale base was hiding (`__repr__` on a half-built instance, and the drive/stop survey inventory), triages the class into the shared numeric-domain parity suite, and resolves the one documentation conflict keeping both sides. The two inline comments are both on the new bridge's command path: its motion surfaces sit entirely outside the `use_ros` operator-approval gate the adjacent doc section documents, and the trailing safety stop's outcome is discarded.

## What's good

- Validation goes through the shared `finite_number_error` / `positive_finite_number_error` / `positive_whole_number_error` domains, and the parity suite (`test_drive_command_numeric_domains.py`) was extended to compare the fourth transport against the other three rather than adding a hand-rolled copy.
- The `__repr__` fix delegates to `partial_construction_repr` (single owner of the wording) and the class is triaged into both halves of that survey, per AGENTS.md rule 14.
- The deliberate exclusion from `BRIDGES` in `test_bridge_stop_tool_parity.py` is recorded next to the inventory with its reason, and the zero-servo halt is pinned in the bridge's own test file.
- Changelog fragment under `changelog.d/`, docs conflict resolution keeps `main`'s safety subsection in the right heading hierarchy, and the atan2 quadrant handling for reverse motion is both explained and tested.

strands_robots/mesh/ackermann_robot.py
```diff
@@ -0,0 +1,369 @@
...
+            wheelbase_m=self.wheelbase_m,
+            max_speed=self.max_speed,
+            max_steering_rad=self.max_steering_rad,
+        )
+        n = max(1, round(duration * self.publish_rate)) if duration is not None else count
+        try:
+            return self._publish_servo(angle, throttle, count=n)
+        finally:
+            if (duration is not None or n > 1) and (angle or throttle):
+                self._publish_servo(0.0, 0.0, count=1)
```
yinsong1986: [MUST FIX] The trailing safety stop's result is discarded, so `drive()` can report success while the car is left at live throttle.

```python
finally:
    if (duration is not None or n > 1) and (angle or throttle):
        self._publish_servo(0.0, 0.0, count=1)
```

`_publish_servo` returns a structured `{"status": ...}` dict (`use_ros` returns error dicts rather than raising), and this call's verdict is dropped. If the main publish succeeded (car driving for `duration` seconds) and the trailing zero then fails - node/daemon dying mid-hold, transport error - the `try` block's success result is what the caller sees. That is exactly the hazard this line exists to close: the tool description promises "A command with duration stops automatically afterwards", so an agent reading `success` will not issue `stop_<node>`, and the servo command latches per the platform semantics this module documents.

This is the "No silent defaults on error" rule plus AGENTS.md > Review Learnings (PR #86) > "Don't discard return values - if a step returns `{"status": ...}`, check it", on a physical-safety path in the changed code - which is why it can't wait for a follow-up: the failure mode is a moving vehicle reported as stopped.

Resolution: capture the trailing publish's result outside the `finally` (or inside it into a local), and when the main publish succeeded but the trailing zero did not, return an error dict stating the drive was published but the halt failed and the car may still be moving - so the agent's next action is `stop`. The existing `test_drive_trailing_zero_sent_even_when_publish_errors` covers the inverse case (main fails, zero sent); this direction is the one where the discard is load-bearing.

strands_robots/mesh/ackermann_robot.py
cagataycali: Both halves are fixed in `43ee3d5`.

**Blocklist coverage.** `/manual_drive`, `/vehicle_state` and `/enable_state` are now default entries, spelled bare so the base-name rule covers `/webserver_pkg/manual_drive` and the two `/ctrl_pkg/...` services. The comment above the set records why the arming pair belongs next to `/motor_enable`: it is what makes the car act on a servo message at all.

**Context threading.** `drive`, `stop` and `enable` take a `tool_context` and thread it into every `use_ros` call (including the trailing halt), and `drive_<node>` / `stop_<node>` are declared `@tool(context=True)` and forward the injected context, exactly as `RosBridgedRobot` does. `get_scan` stays contextless because `echo` is never gated.

Order of prompts on a first drive is now the handshake's two services then the servo publish, pinned as a list rather than a count.

Pinned by `tests/mesh/test_ackermann_command_gate.py` (18 cases), which keeps the real `use_ros` and doubles the rclpy transport instead - the boundary `tests/mesh/test_ackermann_robot.py` patches over, which is why neither half was visible there. It covers the prompt order, a decline at the handshake (nothing published, stops at the first refusal) and at the servo topic, the halt reaching the operator, the timed drive's trailing halt, both allowlist spellings (bare and namespaced), the headless refusal naming both variables, and structural guards on `context=True` plus every command call site.

On the recurrence question: `TestEveryCommandingMeshBridgeHasAGateSuite` derives the set of mesh modules that send a command through `use_ros` from the tree by AST and compares it against an inventory of gate suites, so a fifth bridge fails until it is triaged rather than shipping ungated the way this one did.

Measured: removing the three blocklist entries fails 7 cases; dropping `context=True` from the drive tool fails 2; dropping one `tool_context` forward fails 5.

The posture is stated where an operator sizes a pre-approval: the blocklist table and the final-path-segment example list in `docs/ros2-integration.md`, a per-method gated-surface table in its Ackermann section with the `STRANDS_ROS2_COMMAND_ALLOW` line for a headless run, the surface enumeration and the context-forwarding sentence in `docs/security.md`, and the example's SAFETY block.

strands_robots/mesh/ackermann_robot.py
cagataycali: Fixed in `fcb5ed4`. The halt's result is captured into a local from the `finally` (so it still goes out when the main publish raises), and when the drive succeeded but the halt did not, `drive` returns an error naming the published command, the still-live throttle, the instruction to call `stop`, and the halt's own cause text. A failed drive stays the reported cause - a halt that also failed does not replace it - and a latching single-shot command sends no halt, so there is no verdict to read.

Three cases sit next to `test_drive_trailing_zero_sent_even_when_publish_errors` in `tests/mesh/test_ackermann_robot.py`: the failed-halt direction, the both-failed precedence, and a control that a single-shot success is not reinterpreted from a publish the call never makes. Restoring the discard fails the first.

`docs/ros2-integration.md`'s Ackermann section now states the reported-failure half beside the always-trails-a-zero half, so the promise the tool description makes is the one the docs make.

yinsong1986 (approved):
## Summary

Carries #1035 forward onto current `main`: a fourth mobile-base transport, `AckermannRosRobot`, presenting Ackermann-steering ROS 2 cars (stock AWS DeepRacer wiring in `from_deepracer()`) with the fleet-standard `drive(linear, angular, duration)` contract, a bicycle-model conversion to normalized `ServoCtrlMsg` servo pairs, a run-once `init_services` arming handshake, and a trailing zero-servo halt on every timed command. Both must-fix concerns from the earlier review round are resolved on this head: the trailing halt's verdict is captured from the `finally` and a drive whose halt fails now returns an error naming the possibly-live throttle and `stop` as the next action, and the bridge's command surfaces are gated - `/manual_drive`, `/vehicle_state` and `/enable_state` are default `_command_gate.COMMAND_BLOCKLIST` entries (so round 3's merge widens the gate to all three transports), `drive`/`stop`/`enable` thread `tool_context` into every `use_ros` call including the trailing halt, and the command tools are `@tool(context=True)` while `get_scan` stays contextless. The round-3 base merge correctly hands `use_ros.py` back to `main`'s shared-gate refactor and moves the three entries and their justification to the new owner.

## What's good

- The gate suite keeps the real `use_ros` and doubles the rclpy transport - exactly the boundary the bridge suite patches over, which is why the original gap was invisible there - and `TestEveryCommandingMeshBridgeHasAGateSuite` derives the commanding-bridge inventory from the tree, so a fifth bridge is graded on arrival.
- Validation goes through the shared numeric domains, and `tests/mesh/test_drive_command_numeric_domains.py` is extended to compare all four transports against each other, with the Ackermann wire contract (servo pair) pinned separately from the shared caller-facing domain.
- The half-built-`repr` and drive/stop-survey triage both land with parity entries in the whole-tree guards (`test_repr_survives_partial_construction.py`, `test_bridge_stop_tool_parity.py`), with the reason `AckermannRosRobot` stays out of the Twist-asserting `BRIDGES` cases recorded next to the inventory.
- Operator-facing posture is stated everywhere it is read: the blocklist table and final-segment examples in `docs/ros2-integration.md`, the per-method gated-surface table, the `STRANDS_ROS2_COMMAND_ALLOW` line, `docs/security.md`, and the example's SAFETY block, with both allowlist spellings pinned by tests.
- All new/changed strings are plain ASCII; changelog fragment present and `assemble_changelog.py --check` passes.

## Verification suggestions

Beyond CI: `pytest tests/mesh/test_ackermann_robot.py tests/mesh/test_ackermann_command_gate.py tests/mesh/test_bridge_stop_tool_parity.py tests/mesh/test_drive_command_numeric_domains.py tests/test_repr_survives_partial_construction.py` (212 passed, 1 skipped on this head locally), plus `tests/test_use_ros_command_blocklist.py tests/tools/test_use_ros.py tests/mesh/test_ros_bridge_command_gate.py` (173 passed) to confirm the widened blocklist regresses nothing on the sibling transports.

PR #2690, opened 2026-08-24
fix(sim): hold every orientation parameter to one quaternion domain

cagataycali:
## Problem

Four finite components make a readable vector, not a rotation. A wxyz value whose norm rounds to zero has no direction to recover, and nothing downstream reports that. MuJoCo refuses it through its XML door outright:

```
quat="0 0 0 0"   -> XML Error: zero quaternion is not allowed
```

but the spec-attribute and `qpos` doors this package writes through accept it, substituting identity. Measured on mujoco 3.12.0: `body.quat = [0, 0, 0, 0]` compiles to `body_quat = [0, 0, 0, 0]` and `mj_forward` reads it as identity, while `quat="2 0 0 0"` is normalized to identity and `quat="1e-9 0 0 0"` is stored verbatim.

`move_to` had always refused such a value, with a hand-rolled norm check sitting directly after the shared pose guard. Its scene-construction siblings held the same quaternion to the position contract - four finite components - so the verdict depended on which entry point received it. That is the drift `tests/simulation/mujoco/test_pose_vector_rule_parity.py` exists to prevent; it pinned the position contract on both sides and the quaternion contract on one.

Measured through the public API on `upstream/main` at `fe5ced57`, on a body already holding a quarter turn about z:

| call | verdict on main | body afterwards |
| --- | --- | --- |
| `move_to(orientation=[0,0,0,0])` | error, "~zero norm" | unchanged |
| `move_object(orientation=[0,0,0,0])` | success, text names `[0.0, 0.0, 0.0, 0.0]` | `[1, 0, 0, 0]` |
| `add_object(orientation=[0,0,0,0])` | success | `[1, 0, 0, 0]` |
| `add_robot(orientation=[0,0,0,0])` | success | identity |
| `patch_scene_mjcf` `set_body_quat` / `add_body` | success | identity |

The `move_object` row is the sharp one: the requested quaternion, the quaternion the success text names and the quaternion `get_body_state` reports are three different values, and the quarter turn the body did have is destroyed.

## Change

The norm rule moves into the shared guard beside `coerce_pose_vector`:

- `coerce_orientation_quaternion(method, param, quat)` - `coerce_pose_vector`'s four-component read plus the direction rule, for the `orientation=` keyword parameters.
- `orientation_quaternion_error(method, param, quat)` - the error-only form, for the structured scene ops that pass the caller's value straight to the spec write.
- `MIN_QUATERNION_NORM = 1e-8` - `move_to`'s shipped bound, unchanged.

All eleven orientation call sites route through it (`add_object`, `move_object`, `add_robot` and `set_robot_pose` across the MuJoCo, Isaac and Newton engines; `move_to`; the `quat` entry in the MuJoCo scene ops' field-domain table). `move_to` loses its private copy, so there is one definition rather than two.

Magnitude is still not part of the contract: `[0, 2, 0, 0]` and `[0, 1, 0, 0]` are the same half turn and both are accepted, normalized by the consumer as before. Only a norm with no direction is refused.

The two wrappers differ on `None` in exactly the way `pose_vector_error` and `coerce_pose_vector` already do: a keyword argument left at `None` was not supplied, while a key PRESENT in an op dict carries a value, so `{"op": "set_body_quat", "quat": None}` stays a refusal rather than becoming an omission that applies identity.

## Artifact

The same call on both trees, MuJoCo headless (`MUJOCO_GL=egl`). Left: the starting pose, byte-identical across trees. Middle: `upstream/main` reports success and the quarter turn is gone (5633 px of the render changed). Right: this branch refuses and the pose the body has is the pose it had.

one quaternion domain

The branch's before/after renders differ by at most one level on a handful of pixels (0 px above a 20-level threshold) - EGL noise, not a pose change. Both premises are asserted inside the figure script.

## Tests

`tests/test_orientation_quaternion_domain.py` (new) covers the shared domain and the rule that keeps the two contracts together: no `orientation` is held to the position guard, and no pose guard is asked for a width of 4 anywhere in the package. `tests/simulation/mujoco/test_pose_vector_rule_parity.py` gains the behavioural half - each entry point refusing the value, the verdict matching `move_to`, MuJoCo's own XML refusal as the oracle, and two controls that a non-unit quaternion is still applied.

Pre-fix (the domain present, no call site using it): **8 failed / 308 passed**, all 8 in the two regression classes, 0 in the control classes.

| break | fires |
| --- | --- |
| control (as shipped) | 0 of 316 |
| the norm bound never fires | 12, incl. the two pre-existing `move_to` zero-norm tests |
| refuse any non-unit quaternion (over-reach) | 8, incl. the pre-existing `test_a_non_unit_quaternion_is_normalized` and `test_a_different_quaternion_changes_the_solve` |
| an omitted orientation read as a value | 44 failed + 36 errors |
| largest component instead of the norm | 1 (`test_the_bound_is_a_norm_not_a_component`) |
| revert the scene-op `quat` domain alone | 3 |
| revert the MuJoCo engine call sites alone | 6 |
| error-only wrapper reads a present `None` as an omission | 2, incl. the pre-existing `test_a_none_pose_is_refused[quat]` |

The two single-surface reverts fire complementary sets sharing only the structural rule, so each surface is independently pinned; the over-reach row is refused by the shipped suite rather than only by the new controls.

## Gate

`ruff check` and `ruff format --check` clean on `strands_robots tests tests_integ`. mypy: **5 errors on this branch, 5 on a pristine `upstream/main` worktree**, identical message sets, none in the changed files.

A 379-file selection (every test naming a pose guard, an orientation, `move_to`, the scene-construction calls or `patch_scene_mjcf`), run on both trees with the two changed test files excluded from each: **`14 failed, 12043 passed, 240 skipped, 45 errors` on both**, 532s vs 541s, with the branch-only and base-only failure sets both empty. The 14 are pre-existing (9 `tests_integ` needing live infra or model weights, 4 recording/pacer tests, 1 mesh integration); the 45 errors are `device_connect_edge` absent locally, declared in the `[device-connect]` extra that `[all]` pulls in.

The changed test files pass on mujoco 3.12.0 and on 3.5.0, the floor `pyproject.toml` declares: **139 passed** on both.

### DIFF
--- a/changelog.d/2690-one-quaternion-domain.md
+++ b/changelog.d/2690-one-quaternion-domain.md
@@ -1,0 +1,22 @@
+### Fixed
+
+- **sim**: every entry point that takes a wxyz `orientation` now holds it to one quaternion
+  domain. Four finite components make a readable vector, not a rotation: a value whose norm
+  rounds to zero has no direction to recover, and nothing downstream reported that. MuJoCo
+  refuses `quat="0 0 0 0"` through its XML door outright ("zero quaternion is not allowed") but
+  accepts it through the spec-attribute and `qpos` doors this package writes through,
+  substituting identity. `move_to` had always refused such a value, with a hand-rolled norm
+  check sitting directly after the shared pose guard, while its scene-construction siblings held
+  the same quaternion to the position contract - so the verdict depended on which entry point
+  received it. Measured on a body already holding a quarter turn about z,
+  `move_object(orientation=[0, 0, 0, 0])` reported success, echoed that quaternion back in its
+  text and left the body at identity: the requested value, the reported value and the actual
+  value all different, and the rotation the body did have destroyed. `add_object`, `add_robot`,
+  `set_robot_pose` and the structured `set_body_quat` / `add_body` scene ops accepted it the same
+  way. The check moves into the shared guard as `coerce_orientation_quaternion` /
+  `orientation_quaternion_error` and all eleven orientation call sites route through it, across
+  the MuJoCo, Isaac and Newton engines; `move_to` loses its private copy, so `MIN_QUATERNION_NORM`
+  has one definition rather than two. Magnitude is still not part of the contract - a non-unit
+  quaternion is accepted and normalized by the consumer as before, and only a norm with no
+  direction is refused. A structural guard derives the orientation call sites from the package,
+  so one added later is held to the same domain on arrival.
--- a/strands_robots/simulation/isaac/simulation.py
+++ b/strands_robots/simulation/isaac/simulation.py
@@ -47,6 +47,7 @@ from strands_robots.simulation.terrain import validate_difficulty
 from strands_robots.utils import (
     FREE_CAMERA_TOKENS,
     camera_fov_error,
+    coerce_orientation_quaternion,
     coerce_pose_vector,
     coerce_rgba,
     coerce_size_vector,
@@ -1660,7 +1661,7 @@ class IsaacSimulation(IsaacMotionPrimitivesMixin, IsaacRecordingMixin, SimEngine
         position, _perr = coerce_pose_vector("add_robot", "position", position, 3)
         if _perr is not None:
             return {"status": "error", "content": [{"text": _perr}]}
-        orientation, _oerr = coerce_pose_vector("add_robot", "orientation", orientation, 4)
+        orientation, _oerr = coerce_orientation_quaternion("add_robot", "orientation", orientation)
         if _oerr is not None:
             return {"status": "error", "content": [{"text": _oerr}]}
         # The default None means identity; only a non-identity quaternion is
@@ -2197,7 +2198,7 @@ class IsaacSimulation(IsaacMotionPrimitivesMixin, IsaacRecordingMixin, SimEngine
             position, _perr = coerce_pose_vector("add_object", "position", position, 3)
             if _perr is not None:
                 return {"status": "error", "content": [{"text": _perr}]}
-            orientation, _oerr = coerce_pose_vector("add_object", "orientation", orientation, 4)
+            orientation, _oerr = coerce_orientation_quaternion("add_object", "orientation", orientation)
             if _oerr is not None:
                 return {"status": "error", "content": [{"text": _oerr}]}
             # Same shared domain for the colour, whose accepted counts the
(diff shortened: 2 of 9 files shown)


yinsong1986 (approved):
## Summary

Moves `move_to`'s hand-rolled zero-norm quaternion refusal into the shared pose-guard layer (`coerce_orientation_quaternion` / `orientation_quaternion_error` / `MIN_QUATERNION_NORM` in `strands_robots/utils.py`) and routes all eleven `orientation`/`quat` call sites through it across the MuJoCo, Isaac and Newton engines and the structured scene ops. This closes a real silent-corruption path: on main, `move_object(orientation=[0,0,0,0])` reported success, echoed the zero quaternion in its text, and left the body at identity — destroying the rotation the body previously held. The change is behaviour-tightening in the loud direction (a structured error dict where a silent identity substitution used to be), which matches the repo's no-silent-defaults rule. The error message and the `1e-8` bound are carried over from `move_to` unchanged, so the one pre-existing refusal surface keeps its exact text; the `None`-handling split between the keyword-argument wrapper (omission) and the op-dict wrapper (refusal) mirrors the existing `coerce_pose_vector` / `pose_vector_error` pair.

## What's good

- Single definition of the contract with a structural AST guard (`tests/test_orientation_quaternion_domain.py`) so a future orientation call site held to the position contract fails on arrival — the same drift-prevention shape the repo's shared-guard convention (AGENTS.md key convention 11) asks for, and the second-caller threshold is well past met here.
- Behavioural parity tests extend the existing `test_pose_vector_rule_parity.py` with MuJoCo's own XML refusal as the oracle, plus two controls pinning that non-unit magnitude is still accepted (no over-reach).
- Changelog fragment present (`changelog.d/2690-one-quaternion-domain.md`); docstrings updated at the surfaces whose accepted domain changed; break-detection table in the description shows each surface is independently pinned.
- Scope discipline: only norm-with-no-direction is refused; magnitude normalization semantics are untouched.

yinsong1986 (approved):
## Summary

Moves the zero-norm quaternion refusal that `move_to` had always carried privately into the shared pose-guard layer (`coerce_orientation_quaternion` / `orientation_quaternion_error` / `MIN_QUATERNION_NORM` in `strands_robots/utils.py`) and routes all eleven `orientation`/`quat` call sites through it across the MuJoCo, Isaac and Newton engines and the structured scene ops. This closes a real silent-corruption path: `move_object(orientation=[0,0,0,0])` previously reported success, echoed the requested quaternion in its text, and left the body at identity - three different answers for one call, with the body's prior rotation destroyed. The change tightens validation only (a refusal can be widened later without breaking anyone), keeps the documented `None` split between the coercing and error-only wrappers exactly as the existing `coerce_pose_vector` / `pose_vector_error` pair defines it, and leaves magnitude out of the contract so non-unit quaternions are still accepted and normalized by the consumer.

## What's good

- Error-dict contract preserved at every touched surface; the guard returns rather than raises, per the tool-surface rule.
- The `MIN_QUATERNION_NORM` bound moved unchanged, so `move_to`'s shipped behaviour is byte-identical while gaining a single owner.
- The structural test derives the orientation call sites from the package AST, so a twelfth call site added later is held to the domain on arrival - the same pinning shape AGENTS.md prescribes for shared guards.
- Changelog fragment present (`changelog.d/2690-one-quaternion-domain.md`), correctly named for the PR.
- Mutation table in the description shows each surface independently pinned (the two single-surface reverts fire complementary test sets).

PR #2691, opened 2026-08-24
fix(isaac): read a mesh asset's scale through its MJCF default class

cagataycali:
## Problem

`_parse_mjcf_mesh_assets` in `strands_robots/simulation/isaac/loaders.py` read each
`<asset><mesh>`'s `scale` off the element alone:

```python
scale = _parse_axis(mesh_el.get("scale"), default=(1.0, 1.0, 1.0))
```

MJCF lets a `<default>` class supply that attribute, exactly as it supplies a `<geom>`'s
`type` and a `<joint>`'s `axis` -- both of which this module already resolves through a
class:

```xml
<default class="right_hand"><mesh scale="0.001 0.001 0.001"/></default>
<asset><mesh class="right_hand" file="palm.obj"/></asset>
```

Every mesh written that way was reported at unit scale. Unit scale is a plausible value,
so nothing downstream can tell an asset authored at unit scale from one whose declared
scale was never read -- the load reports success and the asset is reported up to a
thousand times too large.

The reported scale is not cosmetic. It rides onto the visual prim's xform as
`SceneObject.mesh_scale`, and `mesh_aabb` measures the object's collision proxy `size`
through it, so a body whose only geometry is a mesh reports a proxy scaled by the same
factor. Through the public `load_mjcf_scene_objects`, on Menagerie's `flybody` head mesh
(the model's root class declares `<mesh scale="0.1 0.1 0.1"/>`, and each `<asset><mesh>`
names no scale):

| reading | reported `mesh_scale` | reported `size` |
| --- | --- | --- |
| `upstream/main` | `(1.0, 1.0, 1.0)` | 0.5037 x 0.7748 x 0.6438 m |
| this change | `(0.1, 0.1, 0.1)` | 0.0504 x 0.0775 x 0.0644 m |
| `mujoco.MjSpec`, resolving the class itself | `(0.1, 0.1, 0.1)` | -- |

A fruit fly's head reported as 0.77 m across, under a successful scene load.

## Change

`scale` now resolves through `_mjcf_class_defaults(root, mjcf_dir, "mesh")` and the shared
`_class_attrs` precedence -- the same two helpers the `<geom>` and `<joint>` readers in
this module already use, so the rule is stated once rather than per reader. That gives, for
free and by construction: the element's own value wins over its class's; the root class is
reachable as `""`, as `class="main"`, and by naming no class at all; nested classes are
flattened, so an inner class overrides an enclosing one and a nested class declaring no
`<mesh>` inherits it; and `<default>` is read from the spliced model, so a class declared
in an `<include>`d fragment resolves.

Two boundaries are deliberate and tested:

- The mesh class map is collected separately from the geom one, because one `<default>`
  class carries a separate attribute set per element kind. MuJoCo enforces this: neither
  `scale` on a `<geom>` nor `size` on a `<mesh>` is a legal declaration, so resolving a
  mesh against the geom map could only ever lose the scale.
- `file` and `name` stay element-only reads. That is the format's rule rather than a
  simplification -- MuJoCo's schema refuses either attribute inside a `<default><mesh>`
  ("unrecognized attribute"), so a class cannot name an asset or its file.

## Corpus grading

`mujoco.MjSpec` exposes each mesh's own resolved `scale`, which is the very attribute this
reader reports, so it grades the reader directly. Swept over the shipped asset corpus (537
models parsed, 336 graded -- the rest are include-fragments MuJoCo refuses standalone,
identically on both trees):

| reading | mesh assets agreeing with the compiler | disagreeing |
| --- | --- | --- |
| `upstream/main` | 9864 | 456 |
| this change | 10320 | 0 |

No asset moved away from the compiler. The 456 corrected values span 18 shipped Menagerie
models -- `flybody` (85 each in `fruitfly.xml` and `scene.xml`), `robotis_op3` (48),
`anybotics_anymal_b` (46), `shadow_hand` (13 right / 10 left), `trossen_wx250s` (10),
`robotiq_2f85` (8), `stanford_tidybot` (7), `skydio_x2` (1), and each of their scenes.
Every one was reported at unit scale where the model declares 0.001 (284 meshes), 0.1 (170)
or 0.01 (2).

One measurement worth recording, because it silently under-reports this class of defect: a
mesh declaring no `name` carries an *empty* name in a parsed `MjSpec` until `compile()`
runs -- the compiler is what derives the name from the file. An oracle that reads names
before compiling collapses every such mesh into one entry and compares nothing. Most of the
affected assets are written exactly that way, so the oracle here compiles first, and
`test_an_unnamed_mesh_is_keyed_by_its_basename_and_still_scaled` pins that shape.

## Tests

New: `tests/simulation/isaac/test_mjcf_mesh_asset_scale_defaults.py`, 23 tests in three
classes. Every expectation is derived from MuJoCo rather than restated, and every fixture
is a model MuJoCo compiles.

Pre-fix split (reader reverted to `upstream/main`, tests kept): **14 failed / 9 passed**,
all 14 in the two regression classes and **0** in `TestTheElementAndClassPrecedenceIsUnchanged`.

Break table -- each row a single mutation, with the tests it fires:

| mutation | fails | fires |
| --- | --- | --- |
| (control) branch as shipped | 0 | -- |
| exact pre-fix revert of the reader | 14 | every regression + oracle test |
| resolve against the **geom** class defaults instead of the mesh ones | 14 | same set: the geom map can never carry a `scale` |
| let the class override the element instead of the reverse | 1 | `test_the_elements_own_scale_wins_over_its_class` |
| read only the class, ignoring the element | 3 | the three element-precedence tests |
| always resolve against the root class, ignoring the named one | 10 | every named-class case; the root-class cases still pass |
| resolve `name=` through the class too | 0 | deliberate: MuJoCo refuses `name` inside a `<default><mesh>`, so no valid model can distinguish the two readings -- pinned by `test_mujoco_refuses_an_asset_attribute_inside_a_default` |
| oracle reads mesh names without compiling first | 1 | `test_an_unnamed_mesh_is_keyed_by_its_basename_and_still_scaled` alone |
| fixture scale set **to** the unit fallback, premises present | 6 | the premises, naming the cause |
| the same vacuous fixture with the premises removed | 0 | 23 tests pass while measuring nothing -- which is what the premises exist to catch |

## Artifact

`head_body.obj` from Menagerie's `flybody`, rendered headless in MuJoCo (EGL) at the scale
each column reports, centred on its own reported AABB, with an identical camera. The grey
bar is a 0.10 m box declared in the MJCF rather than by the loader, so it is an in-image
control.

reported vs compiled mesh asset scale

Measured on the renders with a segmentation pass (exact geom-id pixel counts, not a colour
threshold): subject pixels **41553** (`upstream/main`) / **359** (this change) / **359**
(MuJoCo resolving the class itself); subject bbox 205x259 / 18x25 / 18x25 px; control bar
**427 px in all three**. This change reproduces MuJoCo's own render exactly --
`np.array_equal` is `True`, mean |delta| `0.000000` -- while `upstream/main` sits 26.69 mean
levels away.

## Local gate

- `ruff check` and `ruff format --check` over `strands_robots tests tests_integ`: clean.
- `mypy strands_robots tests tests_integ`: **24 errors on the branch and 24 on a pristine
  `upstream/main` worktree, byte-identical message sets**, 0 in the changed files.
- Selection of every test touching the loaders / mesh assets / scene objects (197 files,
  identical selection on both trees, the new file excluded from each):
  **`7587 passed, 73 skipped`, zero failures either side** (213.70s branch / 214.05s base).
- The new file: **23 passed on mujoco 3.12.0 and on 3.5.0**, the floor the manifest declares.

### DIFF
--- a/changelog.d/2691-mjcf-mesh-asset-scale-default-class.md
+++ b/changelog.d/2691-mjcf-mesh-asset-scale-default-class.md
@@ -1,0 +1,34 @@
+### Fixed: a mesh asset's `scale` is read from its MJCF `<default>` class, instead of falling back to unit scale
+
+`_parse_mjcf_mesh_assets` read each `<asset><mesh>`'s `scale` off the element alone, but
+MJCF lets a `<default>` class supply it -- `<default class="right_hand"><mesh
+scale="0.001 0.001 0.001"/>` plus `<mesh class="right_hand" file="palm.obj"/>` declares a
+millimetre asset with no `scale` on the element at all. Every such mesh was reported at
+unit scale. That is a plausible value, so nothing downstream could tell an asset authored
+at unit scale from one whose declared scale was never read: the load reported success and
+the asset was reported up to a thousand times too large.
+
+The reported scale is not cosmetic. It rides onto the visual prim's xform as
+`SceneObject.mesh_scale`, and `mesh_aabb` measures the object's collision proxy `size`
+through it, so a body whose only geometry is a mesh reported a proxy scaled by the same
+factor -- Menagerie's `flybody` head, declared at `scale="0.1 0.1 0.1"` by the model's
+root class, was reported 0.50 x 0.77 x 0.64 m instead of 0.050 x 0.077 x 0.064 m.
+
+The reader now resolves `scale` through `_mjcf_class_defaults(root, mjcf_dir, "mesh")` and
+the shared `_class_attrs` precedence, which is how the sibling `<geom>` and `<joint>`
+readers in the same module already resolve their attributes: the element's own value wins,
+then its named class, then the root class (reachable as `""`, as `class="main"`, or by
+naming no class), with nested classes flattened and `<default>` read from the spliced
+model so a class declared in an `<include>`d fragment resolves. The mesh class map is
+collected separately from the geom one because one class carries a separate attribute set
+per element kind. `file` and `name` stay element-only reads, which is the format's rule
+rather than a simplification -- MuJoCo's schema refuses either attribute inside a
+`<default><mesh>`.
+
+Graded against `mujoco.MjSpec`'s own resolved `scale` -- the very attribute this reader
+reports -- across the shipped asset corpus: agreement rises from 9864 of 10320 mesh assets
+to 10320 of 10320, with no asset moving away from the compiler. The 456 corrected values
+span 18 shipped Menagerie models (`flybody`, `robotis_op3`, `anybotics_anymal_b`,
+`shadow_hand`, `trossen_wx250s`, `robotiq_2f85`, `stanford_tidybot`, `skydio_x2` and their
+scenes), and every one of them was reported at unit scale where the model declares 0.001,
+0.1 or 0.01.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -1347,6 +1347,11 @@ def _class_attrs(el: ET.Element, defaults: dict[str, dict[str, str]], childclass
     ``defaults`` must be the map :func:`_mjcf_class_defaults` collected for
     ``el``'s own tag: a joint resolved against geom defaults would inherit a
     geom's ``type``, which names a shape rather than a degree of freedom.
+
+    An ``<asset><mesh>`` reaches this with an empty ``childclass``, which is the
+    format's own rule rather than a simplification: ``<asset>`` is a top-level
+    element, so no body encloses it and its only class is the one it names -
+    else the root class.
     """
     cls = el.get("class") or childclass
     return _merge_mjcf_attrs(defaults.get(cls, {}), el.attrib)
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

A one-line behavioral fix in `_parse_mjcf_mesh_assets`: a mesh asset's `scale` now resolves through its MJCF `<default>` class via the module's existing `_mjcf_class_defaults` / `_class_attrs` helpers, instead of being read off the element alone and silently falling back to unit scale. The fix reuses the same precedence machinery the `<geom>` and `<joint>` readers already use (so no new resolution rule is introduced), keeps `file`/`name` element-only per MuJoCo's own schema, and collects the mesh class map separately from the geom one so the two kinds' attribute sets cannot alias. Corrected values flow through `SceneObject.mesh_scale` and the `mesh_aabb`-measured proxy `size`, both of which the new tests measure through the public `load_mjcf_scene_objects`.

## What's good

- The 23 new tests derive their expectations from `mujoco.MjSpec`'s own resolved `scale` (the exact attribute the reader reports) rather than restating them, and the compile-first oracle pins the empty-name collapse hazard explicitly.
- Precedence controls pin what the fix must not disturb: element-over-class, no-defaults models, a geom class sharing the name, an undeclared class contributing nothing, and MuJoCo's refusal of `file`/`name` inside a `<default><mesh>`.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; no host paths; ASCII-only strings; test names describe behaviour; cross-reference targets (`SceneObject.mesh_scale`, `mesh_assets.mesh_aabb`) resolve and stay on one line.
- The break table in the description (each mutation mapped to the tests it fires) and the corpus grading (456 -> 0 disagreements with the compiler, no asset moving away) make the fix's scope directly checkable.

## Verification suggestions

- Spot-check one shipped model end-to-end: load Menagerie's `flybody` `scene.xml` through `load_mjcf_scene_objects` and confirm the head mesh reports `mesh_scale == (0.1, 0.1, 0.1)` and a sub-0.1 m proxy `size`.
- Re-run the pre-fix split locally (revert the two reader lines, keep the tests) and confirm 14 failures land in the two regression classes and 0 in `TestTheElementAndClassPrecedenceIsUnchanged`, matching the description's table.

PR #2692, opened 2026-08-24
fix(isaac): report an MJCF quat normalized, as the compiler reads it

cagataycali:
## Problem

A quaternion describes a rotation only at unit norm, and MJCF authors routinely spell one non-unit: `quat="1 -1 0 0"` is the idiomatic quarter turn. MuJoCo's compiler normalizes what it reads. `_parse_quat` reported the four components as written, so what the loaders handed back was not a rotation.

What that costs a caller, measured through MuJoCo's own converter:

| quaternion | norm | `mju_quat2Mat` det(R) | length of R applied to a unit axis |
| --- | --- | --- | --- |
| `1 -1 0 0` as written | 1.414214 | 8.0 | 2.0 |
| the same value normalized | 1.0 | 1.0 | 1.0 |

`mju_quat2Mat` builds the matrix from the components as given, so a frame reconstructed from the reported value is the intended quarter turn composed with a uniform scale of `|q|**2`.

The module docstring justified reporting the components as written with "no `quat` in the shipped asset corpus is unnormalized". The corpus refutes that.

## Change

`_parse_quat` normalizes. It is the one helper both loaders resolve a `quat` through, so `load_mjcf`, `load_mjcf_scene_objects` and the mesh-geom frame move together rather than one at a time.

Three things deliberately do not move:

- The four alternative spellings (`euler`, `axisangle`, `xyaxes`, `zaxis`) are constructed at unit norm already and are untouched.
- A malformed `quat` still reads as identity.
- A zero quaternion has no direction to normalize onto, so it joins the malformed readings instead of being scaled. Previously it was reported as four zeros. MuJoCo refuses such a model outright ("zero quaternion is not allowed"), and it is the one orientation `coerce_orientation_quaternion` refuses on the write side, so reporting it verbatim was a reader handing out a value this library's own writers reject.

Magnitude is still not part of the *write* contract: a non-unit `orientation` passed into a setter is accepted and normalized by the consumer, exactly as #2690 left it. This is about what a *read* reports, where the consumer is unknown and the oracle is the compiler.

## Measured against the compiler

Every robot link in the downloaded asset corpus, graded against the `body_quat` MuJoCo's compiler stored for the same body (compared up to sign, since `q` and `-q` are one rotation):

| reading | links graded | agree | disagree | reconciled by normalizing | unexplained |
| `upstream/main` | 8398 | 7589 | 809 | 809 | 0 |
| this branch | 8398 | 8398 | 0 | - | - |

Not one link moves away from the compiler, and the 809 that move toward it are explained in full: normalizing the reported value reconciles every one.

815 links across 140 files report a quaternion whose norm differs from 1 by more than `1e-6` (the six beyond the 809 round to the compiled value anyway). Affected shipped registry robots include `pal_tiago_dual` (15 links, in each of its 7 model and scene files), `shadow_hand`, `wonik_allegro` and `robotiq_2f85_v4`. On the scene-object side, 18 of 619 reported objects were non-unit, in `shadow_hand`, `wonik_allegro` and `robotiq_2f85_v4`.

## Artifact

`pal_tiago_dual`, `arm_left_2_link`, which declares `quat="1 1 0 0"`. Each panel draws the frame the reported orientation describes - three capsules from the origin along the columns of the matrix `mju_quat2Mat` builds from that value, 0.24 m per unit of length. The grey bar is declared in the MJCF rather than derived from the loader, which makes it an in-image control.

reported orientation, upstream vs this branch vs the compiler

| panel | reported wxyz | norm | det(R) | axis length | axis pixels (x, y, z) | reference bar |
| `upstream/main` | `1, 1, 0, 0` | 1.414214 | 8.0 | 0.48 m | 2348, 2320, 2119 | 960 px |
| this branch | `0.707107, 0.707107, 0, 0` | 1.000000 | 1.0 | 0.24 m | 990, 1044, 847 | 960 px |
| MuJoCo's `body_quat` | `0.707107, 0.707107, 0, 0` | 1.000000 | 1.0 | 0.24 m | 990, 1044, 847 | 960 px |

The right two panels are byte-identical (`np.array_equal`), so the branch reports the frame the compiler stored; main is 2.70 mean levels away. Pixel counts are exact geom-id segmentation renders, not colour masks. Rendered headless (`MUJOCO_GL=egl`, mujoco 3.12.0).

## Tests

`tests/simulation/isaac/test_mjcf_quat_normalization.py`, 48 cases, every expectation derived from `mujoco.MjModel` rather than restated by hand. It covers both public loaders, the mesh-geom frame, a nested link (the reported frame is parent-relative and both levels are normalized), the frame the reported value builds, and the zero quaternion offered to this library's own write-side guard. A non-vacuity class asserts each fixture is what it claims: the non-unit declarations really differ from unit norm, and MuJoCo really compiles them.

Two existing expectations were superseded and are updated in the same change: the boundary class in `test_mjcf_robot_body_orientation.py` that pinned "reported as written", and a fixture in `test_description_loader_geometry.py` whose rounded `quat="0.707 0 0 0.707"` now reads as the quarter turn it means.

Pre-fix, with the source reverted and the tests kept: **22 failed, 106 passed** across the three files, all 22 in the four regression classes plus the superseded fixture, **0 in either control class**. Post-fix: **128 passed**, on mujoco 3.12.0 and on the declared 3.5.0 floor.

### What each break costs

Failures over the same three files (128 tests):

| mutation | failures | notes |
| --- | --- | --- |
| control, no mutation | 0 | |
| exact pre-fix source | 22 | the split above |
| infinity norm instead of Euclidean | 20 | also fires the pre-existing spelling suite and a control: a wrong norm breaks already-unit input too |
| no zero-norm guard | 2 | `ZeroDivisionError` on the zero case alone |
| raise on a zero quat instead of identity | 2 | the same two: the zero reading is pinned independently of how it fails |
| near-unit values left as written | 30 | the rounded `0.707` spelling is pinned separately from the plainly scaled ones |
| sign flipped (same rotation, other sign) | 17 | the comparisons grade sign against the compiler, not just the rotation |
| normalized in the robot reader only | 10 | scene-reader classes alone, which is why the fix belongs in the shared helper |

Eight mutations, seven distinct firing sets (the two zero-quaternion mutations coincide).

## Gate

- `ruff check` / `ruff format --check` over `strands_robots tests tests_integ`: clean, 1514 files already formatted.
- `mypy strands_robots tests tests_integ`: 24 errors on this branch and 24 on a pristine `upstream/main` worktree, with byte-identical message sets and none in the changed files.
- A 242-file selection (every test naming the loaders, a reported pose, a scene object or the Isaac backend), with the three changed test files excluded from both sides so the collected set is identical: **9705 passed, 141 skipped, 45 errors on both trees** (201.70s vs 203.76s), 45 failing ids each, and both `comm` directions empty. The 45 errors are `device_connect_edge` missing from this local environment; it is declared as `device-connect-edge>=0.2.0` in the `[device-connect]` extra, which `[all]` pulls in.
- CodeQL AST preflight over the changed source file: identical hit sets on this branch and on the pristine base.

#### After merging main

`main` landed #2691 in this same file after this branch diverged, so the combination was verified rather than assumed:

- Both changes are present: the normalization in `_parse_quat` once, and #2691's mesh-asset class resolution at its call site.
- Both PRs' test files together: 151 passed. The whole `tests/simulation/isaac/` suite: 1334 passed, 5 skipped, under random ordering.
- A break that reverts #2691's mesh-asset class resolution on the merged tree fires 14 tests, all of them in its own three classes - so the resolution did not quietly drop it.
- The pre-fix split, re-measured against the new base: unchanged at 22 failed, 106 passed, with the same class distribution.
- The corpus grading, re-run on the merged tree: still 8398 of 8398 links agreeing with the compiler.
- The selection above (243 files at the new base): **9728 passed, 141 skipped, 45 errors on both trees** (194.33s vs 194.82s), both `comm` directions empty. `mypy` 24 == 24 with byte-identical message sets; `ruff` clean, 1515 files formatted.

### DIFF
--- a/changelog.d/2692-mjcf-quat-normalization.md
+++ b/changelog.d/2692-mjcf-quat-normalization.md
@@ -1,0 +1,23 @@
+### Fixed
+
+- **isaac**: an MJCF `quat` is reported normalized, the way MuJoCo's compiler reads it. A
+  quaternion describes a rotation only at unit norm, and MJCF authors routinely spell one
+  non-unit - `quat="1 -1 0 0"` is the idiomatic quarter turn. `_parse_quat` reported the four
+  components as written, so what the loaders handed back was not a rotation: MuJoCo's own
+  `mju_quat2Mat` builds the matrix from the components as given, and for that declaration it
+  yields the intended turn composed with a uniform scale of `|q|**2` - `det(R)` 8.0 instead of
+  1.0, a unit axis coming back twice as long. Graded against the `body_quat` MuJoCo's compiler
+  stored, 809 of 8398 robot links in the downloaded asset corpus disagreed with the compiler and
+  every one of the 809 is reconciled by normalizing; the reading now agrees on 8398 of 8398, with
+  no link moving away. 815 links across 140 files report a non-unit quaternion, including the
+  shipped registry robots `pal_tiago_dual` (15 links), `shadow_hand`, `wonik_allegro` and
+  `robotiq_2f85_v4`, plus 18 of 619 reported scene objects. `_parse_quat` is the one helper both
+  loaders resolve a `quat` through, so `load_mjcf`, `load_mjcf_scene_objects` and the mesh-geom
+  frame move together; the four alternative spellings (`euler`, `axisangle`, `xyaxes`, `zaxis`)
+  are constructed at unit norm already and are untouched. A zero quaternion has no direction to
+  normalize onto and joins the malformed readings (identity) rather than being reported as four
+  zeros - MuJoCo refuses such a model outright, and it is the one orientation
+  `coerce_orientation_quaternion` refuses on the write side, so reporting it was a reader handing
+  out a value this library's own writers reject. Magnitude remains outside the write contract: a
+  non-unit `orientation` passed into a setter is still accepted and normalized by the consumer.
+  This is what a read reports, where the consumer is unknown and the compiler is the oracle.
--- a/docs/simulation/isaac.md
+++ b/docs/simulation/isaac.md
@@ -214,7 +214,11 @@ open a window.
   `ProceduralRobot` dataclass. `load_mjcf` reports each link's pose in its
   parent's frame, reading the rotation from whichever of MJCF's five spellings
   the body uses - `quat`, `euler`, `axisangle`, `xyaxes` or `zaxis` - under the
-  model-global `<compiler angle>` and `<compiler eulerseq>`.
+  model-global `<compiler angle>` and `<compiler eulerseq>`. The reported
+  orientation is always a unit quaternion, the one MuJoCo's compiler stores: a
+  non-unit spelling such as `quat="1 -1 0 0"` (the idiomatic quarter turn) is
+  reported as the quarter turn it means, not as the components as written, which
+  applied as a rotation would scale the frame by `|q|^2` as well as turning it.
 
 Because the joint-name and observation contract matches the MuJoCo backend,
 policies and observation mappings transfer unchanged between backends.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -908,7 +908,21 @@ class SceneObject:
 
 
 def _parse_quat(quat_str: str | None) -> tuple[float, float, float, float]:
-    """Parse an MJCF ``quat="w x y z"`` string. Identity on failure."""
+    """Parse an MJCF ``quat="w x y z"`` string, normalized. Identity on failure.
+
+    A quaternion describes a rotation only at unit norm, and MuJoCo normalizes
+    every ``quat`` its compiler reads. Reporting the four components as written
+    hands the caller a value that is not a rotation: MuJoCo's own
+    ``mju_quat2Mat`` builds the matrix from the components as given, so
+    ``quat="1 -1 0 0"`` -- the idiomatic spelling of a quarter turn, which the
+    shipped asset corpus uses on hundreds of robot links -- yields that quarter
+    turn composed with a uniform scale of ``|q|**2``, twice size.
+
+    A zero quaternion has no direction to normalize onto, so it is read as
+    malformed: identity, this function's reading for every ``quat`` it cannot
+    use. MuJoCo refuses such a model outright ("zero quaternion is not
+    allowed"), so no model that compiles reaches here with one.
+    """
     if not quat_str:
         return (1.0, 0.0, 0.0, 0.0)
     try:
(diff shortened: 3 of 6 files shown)


yinsong1986 (approved):
## Summary

Normalizes the quaternion `_parse_quat` reports so both public loaders (`load_mjcf`, `load_mjcf_scene_objects`) and the mesh-geom frame hand back the unit quaternion MuJoCo's compiler stores, instead of the components as written in the MJCF - which, applied through `mju_quat2Mat`, rotate *and* scale by `|q|**2`. The fix lands in the one shared helper, the zero quaternion is guarded (it joins the malformed readings as identity rather than dividing by zero), and the two superseded expectations - the boundary class that pinned "reported as written" and the rounded-`0.707` fixture - are replaced in the same change rather than deleted, with the reasoning carried forward. Verified locally at head `7597a573`: the three touched test files pass (127 passed, 1 unrelated `pxr` skip) on mujoco 3.8.0, inside the declared `>=3.5.0` range and distinct from the 3.5.0/3.12.0 pair the description reports; `_parse_quat` returns identity for `"0 0 0 0"`, for a subnormal-underflow input, and for malformed strings without raising. `import math` was already present in `loaders.py`, the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention, and `docs/simulation/isaac.md` documents the new read contract.

## What's good

- Every expected quaternion in the new suite is derived from `mujoco.MjModel` rather than restated by hand, with non-vacuity classes pinning that the fixtures are what they claim.
- The invalidated boundary test was replaced with the new contract statement in the same change, preserving the reasoning (AGENTS.md > "If a premise test is invalidated by a fix landing, replace it rather than deleting it").
- Scope discipline: the four alternative orientation spellings, the malformed-input reading, and the write-side contract are all explicitly left unmoved and pinned as such.

yinsong1986 (approved):
## Summary

Normalizes the quaternion `_parse_quat` reports, so `load_mjcf`, `load_mjcf_scene_objects` and the mesh-geom frame all hand back the unit quaternion MuJoCo's compiler stores rather than the components as written - which, applied through `mju_quat2Mat`, rotate and scale by `|q|**2`. The fix is three lines in the one shared helper (norm, zero guard, divide); the zero quaternion joins the malformed readings (identity) instead of being reported as four zeros, which is consistent with MuJoCo refusing such a model and with `coerce_orientation_quaternion` refusing it on the write side. The previously-pinned "reported as written" boundary is superseded deliberately, with the boundary test replaced rather than deleted and the reasoning updated in the same change - per AGENTS.md, a premise test invalidated by a fix is replaced so the conclusion's new reason survives.

## What's good

- Verified locally against the head: the three affected test files pass post-fix (127 passed, 1 skipped on mujoco 3.8.0), and reverting `loaders.py` to `upstream/main` with the tests kept reproduces the claimed pre-fix split exactly (22 failed, 105 passed) - so the regression pins are load-bearing, not decorative.
- Every expectation in the new suite is derived from `mujoco.MjModel` rather than restated by hand, with non-vacuity classes asserting the fixtures are what they claim and control classes pinning what must not move (unit input, the four alternative spellings, malformed readings, the other link fields).
- Changelog fragment (`changelog.d/2692-mjcf-quat-normalization.md`), docs update, and the corpus-scale measurement (8398/8398 links agreeing with the compiler, 809 reconciled, none moving away) make the behaviour change auditable rather than silent.
- ASCII-clean throughout; the `import math` the change relies on was already present in `loaders.py`.

PR #2693, opened 2026-08-24
security(tools): gate every ROS 2 transport's commands, not just use_ros

cagataycali:
## The operator-approval gate covered one transport, and three reach the graph

Three agent-callable tools reach a ROS 2 graph, all exported from
`strands_robots.tools`, and every one of them can carry a command to a physical
robot:

| tool | transport | `_gate_command` call sites before this change |
|---|---|---|
| `use_ros` | in-process rclpy | 4 |
| `use_rtps` | raw RTPS (cyclonedds) | **0** |
| `use_rosbridge` | rosbridge WebSocket | **0** |

`/cmd_vel` is in `_DEFAULT_COMMAND_BLOCKLIST`, and only `use_ros` was
`@tool(context=True)`. Measured against the same blocklisted surface, with
`BYPASS_TOOL_CONSENT` and `STRANDS_ROS2_COMMAND_ALLOW` unset and the operator
declining:

```
use_ros    prompted=True   status='error'    use_ros: publish to '/probe_only/cmd_vel' was declined by the operator.
use_rtps   accepts tool_context: False
use_rtps                   status='success'  published 1 message(s) to /probe_only/cmd_vel (geometry_msgs/msg/Twist)
```

An agent declined at a drive topic could re-issue the identical command under
another tool name and it went out with no prompt, no allowlist check and no audit
row. `use_rtps` and `use_rosbridge` had no `tool_context` parameter at all, so
there was nothing to decline with - the tool name was the whole difference. This
is the same defect as #2313 ("gate every verb that can command a robot, not just
publish") one level up: one tool gated, three tools reachable.

## The fix: one owner, consulted from every transport

The blocklist, the env knobs and the approval decision move to
`strands_robots.tools._command_gate`, which all three tools consult. That is the
shape `_numeric_options` already uses for the same three tools, and its own
docstring states the argument for this one - it is the single owner of its rule
because the rule "cannot differ between two transports onto the same graph, where
the same publish would otherwise be refused by one and silently mis-paced by the
other". A blocklist is a statement about a physical surface, not about a wire
format.

Which verbs carry a command stays per transport at each call site, mirroring how
`_numeric_options` leaves each tool its own per-action table: only `use_ros`
speaks all three protocols, `use_rtps` has just `publish`, `use_rosbridge` has
`publish` and `service_call`.

`use_ros` keeps `_gate_command` as its transport-specific wrapper, so its three
call sites and its `use_ros:`-prefixed error text are unchanged, and the audit
event source stays `use_ros_tool`.

### Two placement decisions

- **After the backend probe, before the transport lock.** A transport that cannot
  publish never prompts, and a human deciding does not hold the process-wide DDS
  lock or the WebSocket connection. `use_rtps`'s own comment already states this
  posture for its numeric guard - "a refusal happens before a writer joins the
  graph".
- **After argument validation.** The condition mirrors each publish branch's own
  `requires topic and type`, so an incomplete call is reported without asking an
  operator about it - the contract `TestGateRunsAfterArgumentValidation` already
  pins for `use_ros`.

Reading stays ungated on every transport, and so does `use_rtps`'s `advertise`,
which creates a publisher without writing a sample. Both boundaries are pinned so
a later widening cannot prompt an operator for something that moves nothing.

## Verification

New suite `tests/test_every_ros_transport_consults_the_command_gate.py`, run
against a pristine `upstream/main` worktree and against this branch:

| tree | result |
|---|---|
| pristine `upstream/main` | **21 failed**, 10 passed |
| this branch | **31 passed** |

It imports only the three public tools, so it collects on `main` and every
failure there is behavioural or structural rather than a missing import. The
headline row states the bypass in its own assertion text:

```
FAILED ...::test_a_blocked_publish_is_refused_without_passing_any_context[use_rtps]
  AssertionError: use_rtps published to /cmd_vel with no operator approval:
                  published 1 message(s) to /cmd_vel (geometry_msgs/msg/Twist)
```

The 10 that pass on both trees are the `use_ros` cells plus the non-vacuity
check - the controls showing the working transport is untouched.

The structural half derives its inventory from the tree (a module qualifies when
its dispatch compares `action` against a command verb) rather than listing the
three tools, so a fourth transport is graded on arrival instead of shipping
un-gated the way these two did. On this tree the scan finds exactly `use_ros`,
`use_rtps`, `use_rosbridge` and no false positives.

Each half is load-bearing, checked by mutation (baseline: 0 failed):

| mutation | fails |
|---|---|
| `use_rtps` stops consulting the gate | 4 |
| `use_rosbridge` stops consulting the gate | 4 |
| `use_rtps` drops `context=True` | 1 |
| the gate runs before argument validation | 1 |
| a second blocklist copy appears in `use_ros` | 1 |

### Zero regressions

384 files - every suite touching these tools, the gate, the docs they change, or
that walks the tree - run on this branch and on a pristine `upstream/main`
worktree (383 files there; the delta is the new file):

| tree | result |
|---|---|
| this branch | 6 failed, **13197** passed, 90 skipped |
| pristine `upstream/main` | 8 failed, **13163** passed, 90 skipped |

The set of failing ids **branch-only is empty**: all 6 remaining failures are
present on `main` and environment-bound - 2 need Docker, and 4 assert
`DID NOT RAISE ImportError` on a runner where `rclpy` happens to be importable.
`main`'s extra 2 are `tests/mesh/test_zenoh_transport_security.py`, which rotates
which case fails under load.

tests_integ`. `mypy` over the same scope: **Success: no issues found in 1511
source files**.

### Suites that needed triage, and why

Three existing suites aimed publish plumbing at `/cmd_vel`, which is now
correctly gated. Their subject is field-building, message counts and the
advertise/unadvertise lifecycle, not the gate, so they move to a
`_PLUMBING_TOPIC` constant that records why the topic is not a drive surface.
Two neighbours deliberately keep `/cmd_vel` and pass -
`test_ros1_two_segment_type_enforced` and
`test_publish_requires_topic_and_type` - because the gate runs after validation;
they are free ordering controls.

`tests/tools/test_hitl_operator_response_audit.py` derives every `interrupt()`
call site and requires each to be graded for its audit row. Its `_GATES` entry
now names the shared owner, which is what that guard prefers: one interrupt site
and one audit row for the ROS 2 command gate, whichever transport asked, instead
of one per tool.

## Docs

`docs/security.md` scoped both of its gate bullets to "the `use_ros` tool"; each
now names all three transports and states that one shared blocklist serves them,
so a surface refused on one cannot be sent on another. The
"safety-critical command surfaces" section of `docs/ros2-integration.md` gains
the transport axis beside the verb axis it already documented.

No visual artifact: the diff moves a tool-side approval gate and changes no
simulation, rendering or policy behaviour. The before/after measurement above is
the evidence.

## Sequencing note

#2689 adds three entries to `_DEFAULT_COMMAND_BLOCKLIST` and extends the same two
`docs/security.md` sentences. The two changes are complementary - it extends the
enumerations, this one widens each sentence's subject - so whichever lands second
takes both edits; the blocklist entries move to `_command_gate.COMMAND_BLOCKLIST`
unchanged.


yinsong1986 (approved):
## Summary

This PR closes a real gate bypass: `/cmd_vel` and the other safety-critical surfaces were held behind an operator interrupt only on `use_ros`, while `use_rtps` and `use_rosbridge` could publish the identical command with no prompt, no allowlist check and no audit row. The fix extracts the blocklist, both env knobs and the approval decision into a single owner (`strands_robots.tools._command_gate`) consulted from all three transports, makes the two siblings `@tool(context=True)`, and keeps `use_ros`'s wrapper so its interrupt id (`use_ros-command-approval`), audit source (`use_ros_tool`) and `use_ros:`-prefixed refusals are byte-identical. I verified the extraction is faithful: the blocklist set is unchanged, and a differential run of the old `_match_blocklist` against the new `match_blocklist` over exact, unrooted, trailing-slash, namespaced, case-variant, substring and empty spellings diverges on exactly one input - a whitespace-padded `"/cmd_vel "` is now blocked where it previously slipped through - a strictly fail-safe widening. Gate placement in both new call sites mirrors each branch's own `requires topic and type` condition (verified against the rosbridge `service_call` branch, which does require `type`), sits after the backend probe and before the DDS lock / WebSocket dial, and leaves reads and `use_rtps`'s sample-less `advertise` ungated, with both boundaries pinned. No dangling references to the removed `use_ros` private helpers remain anywhere in the tree.

## What's good

- Ran the new suite plus every touched suite locally at this head: 139 passed (gate suites), 208 passed (rtps/rosbridge/numeric-guard/HITL-audit/mesh-robot suites), 227 passed in `tests/tools -k ros`, 60 passed (ros_bridge + flag-domain), and the tree-wide pins (`test_agent_tool_parameter_descriptions`, `test_raises_docstring_completeness`, `test_log_strings_are_ascii`, `test_except_tuples_state_their_real_scope`, `test_test_case_names_describe_behaviour`) all green - the new `tool_context` params carry proper `Args:` entries per AGENTS.md rule 13.
- The structural half derives the commanding-transport inventory from the tree (AST scan for `action` compared against command verbs) and pins single ownership of the blocklist, so a fourth transport or a second blocklist copy fails on arrival - exactly the shape that would have caught this defect when `use_rtps` shipped.
- The HITL audit contract moves with the interrupt site: `test_hitl_operator_response_audit._GATES` now names the shared owner, so one interrupt site and one audit row serve all three transports, keyed by `tool` so the row still says which transport reached the robot.
- One deliberate downstream consequence worth the human reviewer's awareness (not a blocker): `RtpsRobot.drive()`/`stop()` and `RosbridgeRobot.drive()` forward through the now-gated tools with no `tool_context`, so a headless programmatic drive at a blocklisted topic now fails closed with an explicit error naming `STRANDS_ROS2_COMMAND_ALLOW` / `BYPASS_TOOL_CONSENT` - the same documented posture `RosBridgedRobot` has always had over rclpy, loud rather than silent, and fail-closed in the safe direction.
- Changelog fragment present per step 3; both doc surfaces widened to name all three transports.

## Verification suggestions

- Spot-check the extraction differential: `git show upstream/main:strands_robots/tools/use_ros.py` and compare `_match_blocklist`/`_canonical_command_name` behaviour against `strands_robots.tools._command_gate.match_blocklist` on the spellings above - the only divergence should be whitespace-padded names, now blocked.
- `pytest tests/test_every_ros_transport_consults_the_command_gate.py tests/test_use_ros_command_blocklist.py tests/tools/test_hitl_operator_response_audit.py -q` reproduces the headline coverage in ~15s with no ROS install.

PR #2694, opened 2026-08-24
fix(isaac): read a URDF link's pose from the joint that reaches it

cagataycali:
## Problem

`load_urdf` gave every link `position=(0.0, 0.0, 0.0)` and left `BodyDef.orientation` at its identity default, discarding the `<joint><origin>` that is the only place URDF states where a link sits. The comment justifying the discard said the absolute pose would be "computed by joint chain at instantiation time" - but `JointDef` carries no origin, so the offsets were held nowhere in the returned `ProceduralRobot`. A seven-link arm and a seven-link pile at the origin were the same report, and the load succeeded either way.

The sibling reader in the same module, `load_mjcf`, reports both halves of a link's pose in its parent's frame (`pos` into `position`, and the rotation from whichever of MJCF's five spellings the body uses). This is the same quantity read by the same module two different ways, and the MJCF reader's answer is the one MuJoCo compiles.

Measured on the shipped asset corpus: **572 of 601 joints across all 27 URDFs** declare a non-zero `<origin>`, so the discard affected every URDF the registry ships.

## Change

Each link's pose now comes from the `<origin>` of the joint that reaches it - `xyz` into `position`, `rpy` into `orientation` - because URDF places a link on that joint rather than on the `<link>` element. A root link, reached by no joint, keeps the identity pose, which is its placement in the model frame.

`rpy` needs none of the lookups the MJCF reader does: URDF has no `<compiler angle>`, so the triple is always radians, and it always names rotations about the **fixed** axes applied roll-then-pitch-then-yaw, so the composed rotation is `Rz(yaw) * Ry(pitch) * Rx(roll)`. An absent or malformed triple reads as identity, matching `_parse_xyz`'s tolerant reading of the sibling `xyz` attribute on the same element.

The origins are applied **after** `_validate_kinematic_tree`. That guard establishes that at most one joint reaches each body, which is what makes "the joint that reaches this link" a single origin rather than a choice between two - so no origin has to be picked for a link with two parents; that model is refused as before.

## Grading against the compiler

MuJoCo parses URDF, and `body_pos` / `body_quat` are parent-relative - the frame `BodyDef.position` reports - so the compiler is a direct oracle. No expected pose in the new suite is restated by hand. Every shipped URDF, for the 14 files MuJoCo can compile (the other 12 reference mesh assets it cannot decode):

| reading | links agreeing with `mujoco.MjModel` | of |
| --- | --- | --- |
| `upstream/main` | 29 | 127 |
| this change | **127** | 127 |

0 links moved away, and the verdicts are unchanged (26 load, 1 refused, identically on both trees).

The 127 split into two groups, both reconciled with **0 unexplained**:

| group | count | agreement |
| --- | --- | --- |
| loader parent == compiler parent, compared directly | 98 | 98 / 98 |
| compiler re-parents the body (root link merged into `world`, or a `fixed`-joint link welded into its parent), reconciled by composing the loader's chain | 29 | 29 / 29 |

`upstream/main` scores 29 / 98 and 0 / 29 on the same two groups. The re-parenting difference is deliberate - the loader keeps every `<link>` as its own body - and is pinned by `TestAWeldedFixedLinkComposesToTheCompilersFrame`.

## Tests

`tests/simulation/isaac/test_urdf_link_pose.py`, 27 tests. Pre-fix (this file against `upstream/main`'s `loaders.py`): **16 failed / 11 passed**, all 16 in the four regression classes and **0 in `TestTheReadingIsNotWidened` or `TestTheFixtureIsNotVacuous`**.

| mutation | fires | notes |
| --- | --- | --- |
| exact pre-fix `loaders.py` | 16 | 9 + 1 + 5 + 1 across the regression classes |
| only the `xyz` half read | 10 | every rotation assertion, none of the translation-only ones |
| only the `rpy` half read | 9 | complementary: the translation assertions plus the malformed-`rpy` control's `xyz` half |
| `rpy` composed about the moving axes | 6 | only the non-commuting triples, plus the convention test |
| origin applied to the joint's parent, not its child | 17 | includes a control - the wrong body is placed |
| over-reach: a `<visual><origin>` read as the link's pose | 1 | `test_an_origin_on_a_visual_or_collision_is_not_the_links_pose`, alone |
| over-reach: a malformed `rpy` refused instead of read as identity | 2 | the two wrong-length cases, alone |
| vacuity: a fixture rotation set to identity | 1 | the premise fires and names the cause |
| vacuity: same fixture, **premises removed** | 0 | 20 pass while two of them measure nothing - the hole the premise class closes |

The malformed-`rpy` row is only reachable because the length guard sits **outside** the `try`: only the float conversion can raise, so only it is guarded. A wider `try` would swallow a refusal raised by the checks above it, which is what made that mutation a no-op on the first draft.

## Local gate

- `mypy strands_robots tests tests_integ`: **24 errors on this branch, 24 on a pristine `upstream/main` worktree, message sets byte-identical**; 0 in the changed files.
- Full run over the 258-file selection touching the loaders, `ProceduralRobot` or URDF handling, identical on both trees with the new file excluded from each: **`8613 passed, 124 skipped`, zero failures either side** (265.68s vs 258.93s); the branch-only and base-only failing-id sets are both empty.
- The new file: **27 passed on mujoco 3.12.0 and on 3.5.0**, the declared floor.

## Artifact

Proxy render of the poses each reader *reports* for the shipped `so101_new_calib.urdf` (7 links, 6 joints): one box per link at its reported position and orientation, nested per the reported parent. Identical proxy geometry and one fixed camera in all three panels, so the only variable is the per-link pose; the grey 100 mm bar is declared in the proxy file rather than derived from any reader, and measures **786 px in all three**. Rendered headless in MuJoCo with `MUJOCO_GL=egl`.

`upstream/main` collapses all seven links onto the origin (603 px of link proxy); this change renders **5956 px**, and so does a third panel built from the compiler's own `body_pos` / `body_quat` - the two are **pixel-identical** (`np.array_equal` true, mean |delta| 0.000000), while `upstream/main` sits 2.1286 mean levels away.

load_urdf link pose graded against MuJoCo

---

### Round 2 changes

No review feedback outstanding. This round is a base merge: `main` landed an
MJCF quat normalization in `strands_robots/simulation/isaac/loaders.py`, the
same file this branch edits, which took this branch to `CONFLICTING`.

The **source auto-merged** - the two changes touch different functions,
`_parse_quat` and the URDF link-pose read - and only `docs/simulation/isaac.md`
conflicted, where both sides rewrote the one paragraph describing what the
loaders report. Resolved as a union, with the normalization sentence kept
attached to `load_mjcf`: normalizing applies to MJCF's `quat` spelling, while
URDF states a rotation as `rpy`, which is constructed at unit norm and never
needed it.

A clean text merge is not evidence when both changes alter what a link's
reported orientation *is*, so the composition was verified rather than assumed.

**Both changes are present and live.** `_parse_quat` was probed directly on the
merged tree rather than through a suite that skips without `mujoco`:

| input | reported | norm |
|---|---|---|
| `1 -1 0 0` | `0.707107, -0.707107, 0, 0` | 1.000000 |
| `0.707 0 0 0.707` | `0.707107, 0, 0, 0.707107` | 1.000000 |
| `0 0 0 0` | `1, 0, 0, 0` (identity, as documented) | 1.000000 |

**`tests/simulation/isaac/` with `mujoco` 3.12.0 installed**, so the compiler
oracle cases actually run instead of skipping - without it 125 of them skip,
including every case that grades this composition:

| tree | result |
|---|---|
| this branch merged with `main` | **1301** passed, 38 skipped |
| pristine `main` | **1274** passed, 38 skipped |

Zero failures on either; the +27 is this branch's own tests.

**Wider selection** (all of `tests/simulation` plus the ASCII, test-name and
docstring-xref tree graders): an identical **456 failed, 34 errors** on both
trees, every one an absent optional dependency in this environment, with +27
passed on the composition and the branch-only failing set empty.

`ruff check` and `ruff format --check` clean over `strands_robots tests`.

The previous approval was dismissed by this push, as a conflict resolution is
text belonging to neither parent. The re-review surface is the one resolved
documentation paragraph; the source merged without conflict.


yinsong1986 (approved):
## Summary

`load_urdf` previously discarded every `<joint><origin>`, leaving each `BodyDef` at `position=(0,0,0)` / identity orientation with a comment promising a joint-chain computation that `JointDef` (which carries no origin) could never perform. This PR reads both halves of the reaching joint's `<origin>` into the child link's `BodyDef` - `xyz` into `position`, `rpy` (fixed-axis roll-pitch-yaw, radians, composed `Rz*Ry*Rx`) into `orientation` - matching the parent-relative frame the sibling `load_mjcf` reader already reports. The `_parse_urdf_rpy` quaternion composition is correct for URDF's extrinsic convention, the origin write happens after parent/child name validation (no `KeyError` path), and poses are applied only after `_validate_kinematic_tree`, so a link reached by two joints is refused before any origin is picked. The malformed-`rpy` -> identity fallback mirrors the module's established `_parse_xyz` tolerance and is pinned deliberately. No other consumer of `ProceduralRobot` interprets `BodyDef.position` on the URDF path (Isaac instantiation goes through the Kit importer), so there is no double-application risk.

## What's good

- Compiler-as-oracle test design: every expected pose is derived from `mujoco.MjModel.body_pos`/`body_quat` rather than restated by hand, quaternions compared up to sign, and the vacuity/premise classes close the silent-pass hole.
- Reproduced locally at the head SHA: the new suite is 27/27 green, and against the pre-fix `loaders.py` it reports exactly the 16 failed / 11 passed the description claims - the regression classes fire and the controls hold.
- `ruff check` / `ruff format --check` clean on both changed Python files; test names describe behaviour, no host paths, ASCII-only strings, MuJoCo enums passed by value - all per AGENTS.md testing conventions.
- Scope discipline: 4 files including a valid `changelog.d/2694-urdf-link-pose-from-joint-origin.md` fragment (per AGENTS.md step 3), and `docs/simulation/isaac.md` updated in the same PR so the loader contract documentation does not drift.

yinsong1986 (dismissed):
## Summary

`load_urdf` discarded every `<joint><origin>`, leaving each `BodyDef` at `position=(0,0,0)` / identity orientation on the promise of a joint-chain computation that `JointDef` (carrying no origin) could never perform. This PR reads both halves of the reaching joint's `<origin>` into the child link's `BodyDef` -- `xyz` into `position`, `rpy` (fixed-axis roll-pitch-yaw, always radians, composed `Rz*Ry*Rx`) into `orientation` -- applied after `_validate_kinematic_tree`, whose multi-parent guard is exactly what makes "the joint that reaches this link" a single origin. Root links keep the identity pose. The change is scoped to the URDF reader plus a changelog fragment and a docs paragraph; no public signature, wire format, or persisted schema moves -- `BodyDef.orientation` already existed with the identity default the loader had been leaving in place.

## What's good

- Expected poses in the new suite are derived from `mujoco.MjModel`'s `body_pos`/`body_quat` rather than restated by hand, with quaternions compared up to sign and the fixture's own non-vacuity pinned (`TestTheFixtureIsNotVacuous`), so the oracle cannot silently degrade.
- The malformed-`rpy` identity fallback is deliberate, documented on the function, and consistent with the module's existing tolerant readers (`_parse_xyz`, `_parse_axis`, `_safe_float`); the narrow `try` around only the float conversion is commented and mutation-tested per the PR description.
- Controls pin both directions: the reading is not widened to `<visual>`/`<collision>` origins, the joint's other fields are undisturbed, and the two-parent refusal still fires before any origin is picked.
- Changelog fragment and `docs/simulation/isaac.md` land in the same PR, per AGENTS.md step 3.

## Verification suggestions

I reproduced both halves locally: the new suite is 27/27 green against the head `loaders.py` + `procedural.py` on mujoco 3.8.0 (a version between the PR's measured 3.5.0 floor and 3.12.0), and the same file against the pre-fix loader fails in the regression classes only -- consistent with the PR's claimed pre-fix split.

yinsong1986 (approved):
## Summary

`load_urdf` discarded every `<joint><origin>` and reported each link at `position=(0,0,0)` with an identity orientation, on the strength of a comment promising a joint-chain computation `JointDef` (which carries no origin) could never perform. This PR reads both halves of the reaching joint's `<origin>` into the child link's `BodyDef` -- `xyz` into `position`, `rpy` (fixed-axis roll-pitch-yaw, always radians, composed `Rz(yaw)*Ry(pitch)*Rx(roll)`) into `orientation` -- applied after `_validate_kinematic_tree`, whose at-most-one-inbound-joint guard is exactly what makes "the joint that reaches this link" a single origin. Root links keep the identity pose. No public signature, wire format, or persisted schema moves: `BodyDef.orientation` already existed with the identity default the loader had been leaving in place, so this is a behavior fix that brings the URDF reader into agreement with the sibling `load_mjcf` and with MuJoCo's own `body_pos`/`body_quat`.

This re-review covers the round-2 base merge that dismissed the prior approval. The source auto-merged -- the base's `_parse_quat` normalization and this branch's `_parse_urdf_rpy` touch different functions, and both are present and live at the head -- and the one resolved conflict is the `docs/simulation/isaac.md` paragraph, whose union resolution correctly keeps the normalization sentence attached to `load_mjcf` (URDF's `rpy` is constructed at unit norm and never needed it).

## What's good

- Expected poses in the new suite are derived from `mujoco.MjModel` rather than restated by hand, quaternions are compared up to sign, and the fixtures' own non-vacuity is pinned (`TestTheFixtureIsNotVacuous`), so the oracle cannot silently degrade.
- Controls pin both directions: the reading is not widened to `<visual>`/`<collision>` origins, a joint's other fields are undisturbed, and the two-parent model is still refused before any origin is picked.
- The malformed-`rpy` identity fallback matches the module's existing tolerant readers (`_parse_xyz`, `_parse_axis`), with the narrow `try` around only the float conversion commented and mutation-tested per the PR description.
- Changelog fragment and docs paragraph land in the same PR, per AGENTS.md step 3.

cagataycali:
## Conflict resolution at `e4084065` verified

`main` moved under this branch (#2692 landed in the same module, #2693 alongside), and the resolution at `e4084065` is now the head. Verifying it by inspection is not enough here: both sides changed `strands_robots/simulation/isaac/loaders.py`, so a resolution can look clean and still have quietly dropped one side's behaviour, with that side's tests the only thing that would notice. Every number below was measured at `e4084065`.

### Both features survive, proven by break rather than by reading the diff

Reverting each side's change **on the merged head** and reading which tests fire:

| break applied to `e4084065` | fires | where |
|---|---|---|
| #2692's `_parse_quat` normalization reverted | 22 failed, 133 passed | 20 in `test_mjcf_quat_normalization.py` (`TestARobotLinksQuatIsNormalized` 12, `TestASceneObjectsQuatIsNormalized` 6, `TestAZeroQuatIsReadAsMalformed` 2), 1 `test_mjcf_robot_body_orientation.py::TestANonUnitQuatIsNormalized`, 1 `test_description_loader_geometry.py::TestSceneObjectExtraction`; **0** in `test_urdf_link_pose.py` |
| this PR's URDF pose read reverted | 16 failed, 139 passed | all 16 in `test_urdf_link_pose.py` (`TestALinkPoseIsReadFromTheReachingJointsOrigin` 9, `TestOnlyTheJointsOwnOriginIsRead` 5, `TestAWeldedFixedLinkComposesToTheCompilersFrame` 1, `TestTheRpyConventionIsStated` 1); **0** in #2692's files |

The two firing sets are disjoint, so both behaviours are live on the merged head and each is pinned independently of the other. The pre-fix split for this PR is unchanged against the new base: 16 failed / 11 passed, all 16 in the four regression classes and 0 in `TestTheFixtureIsNotVacuous` / `TestTheReadingIsNotWidened`.

### The oracle sweep re-measured against the new base

`MjModel.body_pos` / `body_quat` from MuJoCo's own URDF parse, compared per link (quaternions up to sign, since `q` and `-q` are one rotation). Where MuJoCo re-parented a link - it merges the root link into `world` and welds a `fixed`-joint link into its parent - the loader's chain is composed up to the ancestor MuJoCo kept, so those links are graded rather than skipped.

| tree | direct-parent links agreeing | re-parented links agreeing | total |
|---|---|---|---|
| `c1efac8d` (base) | 54 / 247 | 0 / 33 | 54 of 280 |
| `e4084065` (head) | 247 / 247 | 33 / 33 | **280 of 280** |

Worst residual among agreeing links on the head is 1.15e-16. 24 files are read by both MuJoCo and the loader and 46 are skipped, identically on both trees, so no file changed verdict.

One consequence of the base merge worth naming: #2692 stated the guarantee that every reported orientation is a unit quaternion, at a point when `load_mjcf` was the only loader reporting one. It now holds for both - the URDF path composes its rotation from unit factors, measured at worst `|norm - 1|` of 3.3e-16 over 20006 `rpy` triples including the malformed and absent readings - so the guarantee is a property of the loader pair rather than of one loader.

### Gate at `e4084065`

- 269-file selection (every test naming a loader, `ProceduralRobot`, `BodyDef`/`JointDef`, the Isaac backend, the docs pages or the changelog convention): **8912 passed, 124 skipped, 0 failures**.
- The same selection with this PR's test file excluded, on pristine `c1efac8d`: **8885 passed, 124 skipped, 0 failures**. 8912 - 8885 = 27, exactly this file's tests, so the merge moves no other test.
- `tests/simulation/isaac/test_urdf_link_pose.py`: 27 passed on mujoco 3.12.0 and on 3.5.0, the declared floor.
- mypy: 5 errors on the head and 5 on pristine `c1efac8d`, byte-identical message sets, 0 in the changed files. `ruff check` clean, 1518 files already formatted.

PR #2695, opened 2026-08-24
fix(rendering): hold encode_clip's macro_block_size to the shared whole-number domain

cagataycali:
## Problem

`encode_clip` has three encoder knobs. Two are held to a domain: `fps` resolves through the shared `positive_whole_number_error`, and `quality` through `_clip_quality_error`. The third, `macro_block_size`, was handed to the writer unchecked - and the accepted set was wrong in both directions.

**Nonsense in.** `imageio`'s ffmpeg plugin normalizes a falsy value (`macro_block_size = macro_block_size or 1`) and only rounds when the value is `> 1`, so `0`, `-4`, `None`, `False` and `True` were all read as "no rounding" and the caller was told the clip had been written. Measured on 60x40 frames, a request for `macro_block_size=0` produced a clip **byte-identical** to `macro_block_size=1` (md5 `dffa78600e76`); nothing reported the knob had been dropped. `True` is the same `bool` substitution `_clip_quality_error` already rejects for this module's other quality knob.

**The dependency's check is a bare assert.** `imageio-ffmpeg` guards the knob with `assert isinstance(macro_block_size, int), "macro_block_size must be int"`, so the refusal was an `AssertionError` - absent from the documented `Raises:` set - and `python -O` strips it. On an optimized interpreter `nan` produced that same byte-identical clip and reported success, `2.5` and `inf` surfaced only as "the encoder wrote no clip", and `"8"` leaked a raw `TypeError` out of the writer's `macro_block_size > 1` comparison. **The verdict for one call depended on an interpreter flag.** This is exactly the reasoning already recorded above `_MIN_CLIP_QUALITY` for why `quality` is not left to the writer's own assert; it applies verbatim to the knob 25 lines below.

**Usable out.** `8.0`, `np.int64(8)` and `np.float64(8.0)` each name the same block size as the `8` that encodes, and the writer's `isinstance(..., int)` gate refused all three - the mirror of the `float(quality)` conversion this module already documents as load-bearing.

| request | upstream/main `f224db1c` | this branch |
| --- | --- | --- |
| `macro_block_size=1` | encoded 60x40, md5 `dffa78600e76` | encoded 60x40, md5 `dffa78600e76` |
| `macro_block_size=0` | encoded 60x40, md5 `dffa78600e76` (the `1` clip) | `ValueError: macro_block_size must be a positive whole number, got 0` |
| `nan`, assertions on | `AssertionError: macro_block_size must be int` | `ValueError` naming the parameter |
| `nan`, `python -O` | encoded 60x40, md5 `dffa78600e76` (the `1` clip) | the same `ValueError` |
| `np.int64(8)` | `AssertionError: macro_block_size must be int` | encoded 64x40, md5 `c06938be9bd9` (the `8` clip) |
| `macro_block_size=8` | encoded 64x40, md5 `c06938be9bd9` | encoded 64x40, md5 `c06938be9bd9` |

## Change

`encode_clip` resolves `macro_block_size` through `positive_whole_number_error` beside its existing `fps` guard - the same shared media domain, so a value refused when it names a frame rate cannot be accepted when it names the rounding of that same clip - and passes `int(macro_block_size)` to the writer, exactly as `int(fps)` and `float(quality)` beside it already do and for the same reason. The guard runs before `require_optional`, so the same mistake reports identically whether or not `imageio` is installed, and it applies to both containers: GIF has no macro blocks, but a call must not become valid by changing an extension.

`positive_whole_number_error`'s own docstring enumerates its callers, so `macro_block_size` is added there.

### Scope boundary

A **legitimate** block size the codec refuses stays the post-encode `RuntimeError` it already was: `7` rounds a 48x32 frame to 49x35, which libx264 will not encode, and `10**6` rounds it past what the encoder accepts. This function cannot know which rounded sizes a codec takes, so that verdict belongs to the encoder. The domain therefore applies no ceiling of its own, and `TestTheCodecRefusalBoundaryIsUnchanged` pins that either way, so a later ceiling is a decision rather than an accident.

## Tests

`tests/rendering/test_encode_clip_macro_block_size_domain.py`, 74 tests. Pre-fix (source reverted, test file kept): **60 failed / 14 passed**, all 60 in the four regression classes and **0 in the control class**.

| mutation | fires |
| --- | --- |
| control, no mutation | 0 of 74 |
| both source files reverted (exact pre-fix) | 60 |
| guard removed, `int()` coercion kept | 54 |
| `int(macro_block_size)` coercion removed, guard kept | 6 |
| guard moved after `require_optional` | 5 (the probe tests alone) |
| guard reports the wrong parameter name | 54 (coincides with the guard-removed row: both leave the parameter unnamed to the caller) |
| over-reach: `non_negative` domain, so `0` is accepted | 21 |
| over-reach: an upper bound of 8 added | 3, incl. the row needing `16` |
| over-reach: refuse a size the frame is not a multiple of | 5, incl. `TestTheCodecRefusalBoundaryIsUnchanged` |

The two halves are exactly complementary - 54 (guard) + 6 (coercion) = 60 (both reverted) - so each is independently pinned. The last row is the boundary control firing, which is what makes that boundary load-bearing rather than asserted.

`python -O` is exercised in a child interpreter, with a premise test that measures the dependency's own behaviour there directly (bypassing `encode_clip`), so the `-O` test cannot pass vacuously.

## Gate

- `mypy strands_robots tests tests_integ`: **24 errors on both** this branch and a pristine `upstream/main` worktree, identical message sets - the single diff is a pre-existing `utils.py` error line-shifted by the two docstring lines added.
- Broad selection of 359 test files (every test naming `encode_clip`, `positive_whole_number_error`, rendering, video, mjpeg or recording), identical on both trees with the new file excluded from each: `49 failed, 13734 passed, 132 skipped, 136 errors` on **both**, 49 failing ids each, set difference empty in both directions (589.43s vs 590.90s). Every failure is a declared optional dependency absent locally - `device_connect_edge` (`[device-connect]`) and `awsiot` (`[mesh-iot]`).
- `tests/rendering/`: 666 passed, 9 skipped, under both fixed and random ordering.

## Artifact

Decoded first frame of the clip `encode_clip` returned, per request, on each tree. Source frames are 60x40 (divisible by neither 8 nor 16) so the rounding is visible in the encoded size. Rows 1 and 5 are byte-identical on both trees, so only the disputed values move; every md5 and the two cross-tree equalities are asserted inside the script that draws the figure.

encode_clip macro_block_size domain

### DIFF
--- a/changelog.d/2695-encode-clip-macro-block-size-domain.md
+++ b/changelog.d/2695-encode-clip-macro-block-size-domain.md
@@ -1,0 +1,33 @@
+### Fixed: `encode_clip`'s `macro_block_size` is held to the shared whole-number domain
+
+`encode_clip` has three encoder knobs and only two were checked: `fps` resolved through
+`positive_whole_number_error` and `quality` through `_clip_quality_error`, while
+`macro_block_size` was handed to the writer unchecked. The accepted set was wrong in both
+directions.
+
+`imageio`'s ffmpeg plugin normalizes a falsy value (`macro_block_size = macro_block_size or 1`)
+and only rounds when the value is `> 1`, so `0`, `-4`, `None`, `False` and `True` were all read
+as "no rounding" while the caller was told the clip had been written. On 60x40 frames a request
+for `macro_block_size=0` produced a clip byte-identical to `macro_block_size=1`, so nothing
+could tell a dropped request from an honored one. And `imageio-ffmpeg` enforces the knob's type
+with a bare `assert isinstance(macro_block_size, int)`, which `python -O` strips: on an
+optimized interpreter `nan` produced that same byte-identical clip and reported success, `2.5`
+and `inf` surfaced only as "the encoder wrote no clip", and `"8"` leaked a raw `TypeError` out
+of the writer's comparison. The verdict for one call depended on an interpreter flag - the same
+reason `quality` is not left to the writer's own assert either.
+
+In the other direction, `8.0`, `np.int64(8)` and `np.float64(8.0)` each name the same block size
+as the `8` that encodes, and the writer's `isinstance(..., int)` gate refused all three.
+
+The knob now resolves through the shared `positive_whole_number_error` beside the existing `fps`
+guard, so a value refused when it names a frame rate cannot be accepted when it names the
+rounding of that same clip, and it is passed to the writer as `int(macro_block_size)` exactly as
+`int(fps)` and `float(quality)` beside it already are. The guard runs before the optional-encoder
+probe, so the same mistake reports identically whether or not `imageio` is installed, and it
+applies to both containers: GIF has no macro blocks, but a call must not become valid by changing
+an extension.
+
+A block size that is a positive whole number and that the codec nonetheless refuses - `7` rounds
+a 48x32 frame to 49x35, which libx264 will not encode - stays the post-encode `RuntimeError` it
+already was. This function cannot know which rounded sizes a codec takes, so the domain applies
+no ceiling of its own and that boundary is pinned either way.
--- a/strands_robots/rendering/video.py
+++ b/strands_robots/rendering/video.py
@@ -104,9 +104,16 @@ def encode_clip(
             but the domain applies to both containers so that one call does not
             become valid by changing the output extension. A NumPy real is
             accepted and converted for the writer.
-        macro_block_size: libx264 macro-block rounding; ``1`` preserves the
-            exact frame size (the recorders' convention), ``8``/``16`` lets
-            ffmpeg pad to codec-friendly sizes. Ignored for GIF.
+        macro_block_size: libx264 macro-block rounding, a positive whole number
+            of pixels on the same shared media domain as ``fps`` (see
+            :func:`~strands_robots.utils.positive_whole_number_error`); ``1``
+            preserves the exact frame size (the recorders' convention),
+            ``8``/``16`` lets ffmpeg pad to codec-friendly sizes. Read only by
+            the MP4 writer, since GIF has no macro blocks, but the domain
+            applies to both containers for the same reason ``quality``'s does.
+            A NumPy integer or an integral float is accepted and converted for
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Adds the missing domain guard for `encode_clip`'s third encoder knob: `macro_block_size` now resolves through the shared `positive_whole_number_error` beside the existing `fps` guard, and is passed to the writer as `int(macro_block_size)` for the same reason `float(quality)` is already load-bearing there. This closes both failure directions the description documents: falsy/bool/non-finite values that `imageio`'s plugin silently read as "no rounding" (with the dependency's own `assert` stripped under `python -O`, so the verdict depended on an interpreter flag), and usable NumPy/integral-float spellings the writer's `isinstance(..., int)` gate refused. The default (`1`) and every in-tree caller (`8`) sit inside the accepted set, so no legitimate call changes behavior; previously-accepted nonsense now fails with a loud `ValueError` naming the parameter, which is the repo's documented pattern rather than a silent behavior change. The guard runs before `require_optional` so the refusal is independent of whether `imageio` is installed, and the codec-refusal boundary (a legitimate size like `7` that libx264 rejects post-rounding) is deliberately left as the existing `RuntimeError` and pinned either way.

## What's good

- Follows the module's own precedent exactly (`quality`/`_MIN_CLIP_QUALITY` reasoning applied to the knob 25 lines below), so the fix is a consistency restoration rather than a new policy.
- `Raises:` block updated in the same change (AGENTS.md > "A `Raises:` block names every refusal the function itself makes"), and the shared guard's caller enumeration in `strands_robots/utils.py` updated alongside.
- Changelog lands as a `changelog.d/` fragment, not an edit to `CHANGELOG.md`.
- Test suite compares refusal text against the helper itself rather than a restated string, exercises `-O` in a child interpreter with a premise test that measures the dependency directly, and pins the no-ceiling boundary so a later ceiling is a decision rather than an accident. Mutation table in the description shows the two halves (guard + coercion) are independently pinned.
- All changed files are plain ASCII; test names describe behavior, not provenance.

PR #2696, opened 2026-08-24
refactor(mesh): MobileBaseRobot + Transport - one drive contract for every transport

cagataycali:
Continuation of #1722 by @Vivek0712, carried forward over a base that has moved 421 commits. Their commits are preserved as ancestors of this branch (`741f4057` is an ancestor of the head here); the only new work is the base merge, the conflict resolution, and the guard updates the merge exposed.

#1722 is approved twice on its merits and green at its tip. Its one remaining blocker is the conflict, which `scripts/check_merge_blockers.py` reports as owed by the author. This carries that over the line.

## Why the branch is worth landing

Four robot classes were roughly 80% identical, and the duplication is why `RosBridgedRobot` - the class most likely to be driving real hardware - shipped without the NaN guard, the duration bound and the trailing-zero stop its newer siblings had. `MobileBaseRobot` now owns validation, the drive/stop contract, the `init_services` handshake and the `tools` property; a robot class supplies only a `Transport` (how bytes move) and, when the platform is not differential-drive, a `_cmd_fields` override (what the message looks like).

The strongest evidence is that main's own guard suites become the acceptance bar: `tests/mesh/test_drive_command_guards.py` + `test_navigation_goal_guards.py`, written against the pre-refactor classes, pass unmodified against the consolidated ones.

## The conflict was not mechanical

Main landed two hardenings directly on `RosBridgedRobot` while this sat, and both are on the surface the branch consolidates:

- **#2142** gave the ROS 2 bridge the `stop_<node>` agent tool it lacked.
- **#2313** gated every commanding verb behind `use_ros`'s operator approval: `drive` / `stop` / `navigate` became `@tool(context=True)` and forward the injected context into `use_ros`.

Taking the branch side for the conflicting regions reverts the second one. The branch deletes `RosBridgedRobot.drive` and moves it to a base written before that gate existed, so the tools would have shipped without `context=True` and the bridge's whole command surface would have published to `/cmd_vel` with no operator decision reachable.

Main's guard catches it. `tests/mesh/test_ros_bridge_command_gate.py` on the naive resolution:

```
8 failed, 6 passed
TypeError: MobileBaseRobot.tools.<locals>.drive() got an unexpected keyword argument 'tool_context'
TypeError: MobileBaseRobot.tools.<locals>.stop() got an unexpected keyword argument 'tool_context'
TypeError: _UseRosTransport.action_send_goal() got an unexpected keyword argument 'tool_context'
AssertionError: assert {'navigate'} == {'drive', 'navigate', 'stop'}
AssertionError: use_ros(action='publish') at ros_bridge.py:115 does not forward tool_context
```

Five behavioural, three structural.

## Resolved by carrying the gate into the base

Which is this branch's own thesis applied to the newest safety fix - the gate lands once for every transport instead of once per bridge:

- `Transport.publish` and the two optional command capabilities (`ServiceCapable.service_call`, `ActionCapable.action_send_goal`) take `tool_context`.
- `drive` / `stop` / `enable` thread it. The trailing zero carries the same context as the command it undoes - a trailing zero that could not reach the gate would leave the robot latched at speed.
- The shared `tools` property declares the command tools `@tool(context=True)`; the read tools stay contextless because a read is never gated.
- The transport is the single place the context is handed to the underlying tool, because the gate is the tool's and not the base's.

`_UseRtpsTransport.publish` forwards `tool_context` to `use_rtps`, which gates its command surface through `_command_gate.gate_command` (landed in #2693). Both shipped transports now gate, which is the correct security posture.

The structural half of main's guard now reads the classes in the bridge's MRO instead of one file, so it follows the tools to wherever they are declared. Scanning only `ros_bridge.py` stopped covering `drive` and `stop` the moment they moved to the base, which is precisely how a command tool could have lost `context=True` with nothing noticing. The membership assertion on the discovered file set is the non-vacuity check.

## Three guards main grew for this surface

The full suite caught six regressions the affected-suite runs could not - all three guards landed after this branch's base, and none lives in a file the branch touches, so the base-overlap gate had nothing to flag.

**`test_drive_contract_fleet_scope.py`** measures which drive guarantees hold on all three bridges and grades the prose against the measurement. Its docstring says a bridge gaining the trailing stop "makes the scope assertion fail, which is the intended signal: the guarantee has become fleet-wide and the two places that scope it have to say so." That is what happened - measured on this branch, a timed drive ends with a zero on all three:

| bridge | final `linear.x` after `drive(linear=1.0, duration=5.0)`, before | after |
|---|---|---|
| `RosBridgedRobot` | 1.0 (latched) | 0.0 |
| `RtpsRobot` | 1.0 (latched) | 0.0 |
| `RosbridgeRobot` | 0.0 | 0.0 |

So the trailing zero moves into the fleet-wide half, and the two prose surfaces that scoped it to one bridge now say so. The velocity clamp and the `max_duration` ceiling stay this bridge's alone *among the three Twist bridges this guard compares*, which keeps that guard non-vacuous: neither ported class knows the ceilings of the third-party robot it drives, and defaulting one would silently cap an existing caller. The bridge-specific paragraph said the other two "leave the last velocity latched", which is no longer true; it now says what they actually do instead, and the assertion that pinned the stale word is updated with it rather than the prose being kept false to satisfy a test. `test_a_timed_drive_self_stops_on_one_bridge_and_latches_on_the_other_two` is renamed - the count is no longer one and two.

**`test_bridge_stop_tool_parity.py`** pins the set of classes owning a drive/stop pair. Updated the way this branch already handled the same shape in `test_drive_command_numeric_domains.py`: assert the new set and assert the *absence* of the two that now inherit, so the consolidation itself is pinned rather than merely tolerated.

**`test_use_ros_command_blocklist.py`** refuses an operator-facing document that claims the halt is exempt from a gate that gates it. The docs sentence added here opened "stop() is never gated on the enable handshake", which its clause scanner reads as an exemption claim - correctly, since a reader who believes it leaves `cmd_vel` out of the allowlist and finds `stop()` refused in the field. Reworded to state the gating first and the two things the halt genuinely does not depend on second.

## Verification

On the composed tree (current `main` merged in, not the head alone):

- Affected suites, 14 files: **753 passed, 1 skipped, 0 failed**. Includes main's ROS 2 gate suite, the new `test_ackermann_command_gate.py` and `test_every_ros_transport_consults_the_command_gate.py` from #2693/#2689, its 191 pre-refactor drive and navigation guard tests unmodified, and the repr guard.
- `tests/mesh/test_mobile_base.py` alone: **105 passed** (was 101 before R3).
- Whole `tests/mesh` directory against a pristine `origin/main` worktree in the same environment, identical selection: **135 failing node IDs on each tree, sets byte-identical** - zero branch-only, zero base-only. All are missing optional deps (`awsiot`, the lerobot family). `tests/mesh/test_zenoh_transport_security.py` is excluded from that pair because two Zenoh sessions on one host cross-talk; this branch touches no zenoh file.
- `mypy strands_robots tests tests_integ`: **36 == 36** against the pristine base worktree, identical message sets, zero in the changed files.

## Note for @Vivek0712

If you would rather land this yourself, the resolved tree is on `cagataycali:carry/mobile-base-abstraction` and cherry-picks cleanly onto your branch - pushing it from your own account also clears the last-push-approval condition that #1722 has been stuck behind. Happy either way; close whichever of the two is redundant.

---

## Section 13 - Review Round Log

| Round | Concern | Fix Commit | Pin Test |
|-------|---------|------------|----------|
| R1 | `_UseRtpsTransport.publish` drops `tool_context` after #2693 gave `use_rtps` its operator gate | `b683611c` | `tests/mesh/test_mobile_base.py::test_a_transport_forwards_the_context_exactly_when_its_tool_gates` |
| R2 | Base merge owed: `main` moved 4 commits and the PR was `CONFLICTING`. All four conflicts were set-membership - #2689 added `AckermannRosRobot`, a fourth drive-owning class - resolved as unions with the consolidation semantics kept | `87ca88e2` | `tests/mesh/test_bridge_stop_tool_parity.py`, `test_drive_command_numeric_domains.py::test_the_drive_owning_bridges_are_the_expected_ones`, `tests/test_repr_survives_partial_construction.py` |
| R2 | R1's edit dropped the blank lines around `_check` and before both `class` statements in `rtps_robot.py`, leaving the one file out of 1528 that failed `ruff format --check` | `52776e64` | `ruff format --check` |
| R3 | R1's forwarding fix was pinned only by a source scan for a `tool_context=` keyword, which never inspects what is passed - so `tool_context=None` reintroduced the reported defect with **124 passed** | `eb4be204` | `tests/mesh/test_rtps_robot_command_gate.py` - 9 outcome cases; **4 fail** on that mutation |
| R3 | `test_the_shipped_transports_cover_both_sides_of_the_gate_split` had decayed to a non-empty assertion with an unreachable `if ungated:` branch, because every shipped tool now gates | `d6f6fc6d` | `test_the_forwarding_rule_reports_exactly_the_two_wrong_pairings` (4 constructed pairings), `test_every_shipped_transport_onto_a_ros2_graph_gates_its_commands` |
| R3 | `docs/security.md` enumerated two forwarding robots when three forward; the `docs/ros2-integration.md` shared-contract bullet and `Transport.publish`'s docstring both implied a has-no-gate exemption no shipped transport is in | `e4d55f57` | prose only; the three doc pins in `test_hitl_gate_enumerations_name_every_gated_action.py` key on other sentences and stay green |
| R3 | Base merge owed again: `main` moved 2 commits (#2698, #2699 - both simulation recorder, textually and semantically disjoint from `mesh/`) | `5d2e9d52` | 259 passed on the composed drive/gate suites; clean merge, no conflicts |
| R4 | Code scanning: `_TWIST` in the new RTPS gate suite was a module-level constant defined once and never read | `63ba04da` | dropped the constant; `tests/mesh` failing-node-ID sets byte-identical either way, 135 == 135 |
| R5 | The constant marked a premise the suite never wired up: the IDL double answered every type string, so `RtpsRobot.cmd_vel_type` was not an input to any assertion and a robot declaring `geometry_msgs/msg/Point` still read as having driven | `d794a7a6` | `tests/mesh/test_rtps_robot_command_gate.py::TestTheDeclaredInterfaceReachesTheGatedPublish` - 5 cases; the wrong declaration fires **7 of 14** where it fired **0 of 9** |

**R2 note:** `AckermannRosRobot` keeps its own drive/stop pair - it is not layered on `MobileBaseRobot` - so it joins the *owning* sets, while `RosBridgedRobot`/`RtpsRobot` stay asserted *absent*, which is what pins the consolidation rather than merely tolerating it. Consolidating that fourth class is deliberately out of scope here.

**R3 note - what changed and why it was worth a round.** R1 fixed the forwarding, but the security guarantee rested entirely on a structural check a one-token edit defeats. Measured on `52776e64`:

```
# strands_robots/mesh/rtps_robot.py, the single forward
-            tool_context=tool_context,
+            tool_context=None,

$ pytest tests/mesh/test_mobile_base.py tests/mesh/test_rtps_robot.py \
         tests/mesh/test_ros_bridge_command_gate.py -q
124 passed
```

That mutation *is* the reported defect: the operator becomes unreachable and every `RtpsRobot` command at a blocklisted surface fails closed with `No tool_context available for operator approval`, whose only offered remedy is the blanket bypass. Nothing asserted an outcome.

`tests/mesh/test_rtps_robot_command_gate.py` is the sibling of the existing `test_ros_bridge_command_gate.py` for the other transport onto a ROS 2 graph, so both gate suites present the same thing to their tool: keep the real `use_rtps`, double the DDS backend beneath it (the boundary `tests/tools/test_use_rtps.py` already doubles). No DDS domain needed. The operator's answer is now an input to the outcome:

| operator | prompted about | status | samples written |
|---|---|---|---|
| approves (`y`) | `/turtle1/cmd_vel` | success | 1 |
| declines (`n`) | `/turtle1/cmd_vel` | error - declined | 0 |
| none present | never asked | error - names both env vars | 0 |

Before R1 all three rows were byte-identical to row 3. The headless row is unchanged by the fix and is kept as the boundary control. Against the mutation above: **4 failed** rather than 124 passed.

The non-vacuity rework is the same lesson one level up. Reading the gating/ungated split off the shipped tools decays silently the moment the last ungated tool grows a gate - which is exactly what #2693 did. The rule is now one predicate (`gate_forwarding_violations`, returning the offending verbs) graded on four constructed pairings, so both directions stay pinned however the shipped set moves. Breaking the predicate so it never reports a violation gives **2 failed, 103 passed** on the pairings while the shipped matrix stays green - which is why the shipped matrix could not be the only grader. The split count is replaced by the security claim stated positively: no shipped transport reaches a ROS 2 graph through a tool with no operator approval, so an agent declined on one transport cannot re-issue the command through another.

The merge also surfaced a defect that is **not** in this diff, because it is neither caused nor fixed by the consolidation: #2689 shipped `max_speed`, `max_steering_rad` and `max_duration` on `AckermannRosRobot` while `docs/rosbridge-integration.md` still marks "Velocity clamps" and "Loud duration rejection" as *(this bridge only)*. That is false on `main` today, and `test_drive_contract_fleet_scope.py` - the guard written to catch exactly this drift - passes **16/16 on pristine `main`** because its `_BRIDGES` list is the three Twist bridges, so the fourth class falls outside what it measures. Filed separately.

**R4/R5 note - the constant marked a premise, and dropping it closed the finding without closing the hole.** The alert was correct: `_TWIST` was defined once and never read, and `63ba04da` removed it. Re-reading what it was *for* showed the finding was load-bearing rather than cosmetic. `idl_mod.get_type` was doubled as `lambda ros_type: _Twist`, discarding its argument and handing back the fake velocity dataclass whatever the caller asked for, so the interface the robot declares was not an input to any assertion in the file. Measured with only `RtpsRobot`'s declared `cmd_vel_type` changed and the suite otherwise untouched:

| declared `cmd_vel_type` | the real stack | gate suite at `63ba04da` | gate suite at `d794a7a6` |
|---|---|---|---|
| `geometry_msgs/msg/Twist` | `success`, 1 sample written | 9 passed | 14 passed |
| `geometry_msgs/msg/Point` - in the bundle, no velocity field | `error` - `unknown field 'linear' for Point (have: ['x', 'y', 'z'])`, 0 written | **9 passed** | **7 failed** |
| `sensor_msgs/msg/JointState` | `error` - `unknown field 'linear'`, 0 written | **9 passed** | **8 failed** |

Row 2 is the shape that matters: a declaration that is well-formed for the tool's `pkg/msg/Name` check, resolvable, and unable to carry `linear.x`. The real `use_rtps` reports it and publishes nothing; under a resolver that answers everything, all nine cases passed - `assert sample.linear.x == 1.0` included, because the sample was built from the double rather than from the declaration.

So the resolver double now carries a bundle keyed by the same type strings the real one uses (`_BUNDLE`: `_TWIST` to `_Twist`, `_POINT` to `_Point`) and refuses the rest the way `idl.get_type` does, and the writer double records the `(topic, type)` surface it was opened for. That is what makes the declaration reach the nine cases that were already here; five new cases in `TestTheDeclaredInterfaceReachesTheGatedPublish` assert it directly, and the tool's own success report - which names the resolved type - is now matched in full rather than truncated one character before it.

The readability hazard R4 cites is addressed by the same change rather than by removal: what was ambiguous was a *dead* `_TWIST` string beside a live `_Twist` dataclass. Both are live now and `_BUNDLE` writes down the relationship between them, so "the Twist one" has a single referent.

Break table, all against `tests/mesh/test_rtps_robot_command_gate.py` (14 cases):

| mutation | fires |
|---|---|
| the resolver double answers every type string (the pre-R5 double) | 2 - both interface refusal cases |
| the doubles record the assumed constant instead of the argument they were given | 2 - same pair |
| the base publishes the transport's own `twist_type` instead of the robot's declaration | 2 - same pair |
| the tool's report stops naming the resolved type | 1 - `test_drive_tool_prompts_the_operator_and_publishes_on_approval` |
| declared `cmd_vel_type` -> `geometry_msgs/msg/Point` | **7**, five of them pre-existing gate cases |
| declared `cmd_vel_type` -> `sensor_msgs/msg/JointState` | **8** |
| the doubled bundle carries a type no real bundle has | 1 - the bundle-fidelity premise |
| the type used to provoke the refusal is one the real bundle resolves | 1 - the same premise |

The first three coincide, and the reason is worth stating: no type in the shipped bundle other than `Twist` can carry a velocity, so on a publish that succeeds the writer pair can only ever hold `_TWIST`. The two refusal cases - which declare a non-default type - are what pin the declaration reaching the resolver at all. The two premise rows keep the double honest in both directions: it must resolve only types the real bundle has, and the type used to provoke a refusal must be one the real bundle would also refuse. That premise skips rather than passes when `cyclonedds` is absent, which is the case in CI, since the real `REGISTRY` is empty without it.

Verification of this round, on the composed tree (`63ba04da` + current `main` + the change):

- `tests/mesh/test_rtps_robot_command_gate.py`: **14 passed**.
- `tests/mesh` + `tests/tools` + every test that walks the tree with `ast`/`rglob`/`inspect.getsource`, 442 files, identical selection against a pristine worktree of the previous head with the changed file excluded from each: **66 failed, 17503 passed, 98 skipped, 150 errors on both**, 216 failing node IDs each, `comm` empty in both directions. All missing optional deps: 326 `device_connect_edge` (the `[device-connect]` extra), 88 `awsiot` (`awsiotsdk` in `[mesh-iot]`).
- `tests/mesh/test_zenoh_transport_security.py` run alone on each tree, since two Zenoh sessions on one host cross-talk in a paired run: **9 passed, 2 skipped on both**.
- `mypy strands_robots tests tests_integ`: **24 == 24**, normalized message sets byte-identical, zero in the changed file.
- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1528 files).
- The changelog fragment gains one sentence naming the guarantee; its three graders are **202 passed** with the merged-in `main` suites.


strands_robots/mesh/rtps_robot.py
```diff
@@ -57,7 +58,50 @@ def _check(label: str, value: str, pattern: re.Pattern[str]) -> str:
...
+        count: int,
+        rate: float,
+        tool_context: ToolContext | None = None,
+    ) -> dict[str, Any]:
+        # Accepted to satisfy the Transport protocol and deliberately not
+        # forwarded: ``use_rtps`` exposes no operator-approval gate, so there is
+        # no parameter to hand it and passing one would raise. This is the single
+        # line that changes if it grows one - the base already carries the
+        # context this far for every transport.
+        return use_rtps(action="publish", topic=topic, type=type, fields=fields, count=count, rate=rate)
```
yinsong1986: [MUST FIX] `use_rtps` gained the operator-approval gate on `main` after this branch's base: #2693 (`c1efac8d`, merged ~35 minutes before this head was pushed) made `use_rtps` `@tool(context=True)` and routes every `publish` through `strands_robots.tools._command_gate.gate_command`. This line - and the comment above it saying "passing one would raise" - is written against the pre-#2693 tool, so on the composed tree this transport drops the operator context the gate now requires.

Measured on the composition (current `main` + this branch; the merge is textually clean, so no conflict marker flags it):

- `tests/mesh/test_mobile_base.py`: **2 failed, 99 passed** - `test_a_transport_forwards_the_context_exactly_when_its_tool_gates[_UseRtpsTransport-use_rtps]` (`"_UseRtpsTransport.publish drops the operator context but its tool gates"`) and `test_the_shipped_transports_cover_both_sides_of_the_gate_split` (the ungated side is now empty). On this branch against its own base the same file is 101 passed.
- Behaviourally, every `RtpsRobot` command at a blocklisted surface fails closed with `"No tool_context available for operator approval. Set STRANDS_ROS2_COMMAND_ALLOW or BYPASS_TOOL_CONSENT=true"` - the prompt is unreachable even when an operator is present. `/cmd_vel` is on `COMMAND_BLOCKLIST` and matches every namespaced variant, so this covers the module's own usage example, the `stop_<suffix>` halt tool, and the trailing zero. The refusal's only offered remedy is the blanket bypass, which is the shape AGENTS.md flags for the RTPS bridge's own security gate ("advise the insecure opt-out").

Why this can't be deferred: branch CI checks out the head commit, not the merge ref, and the base-overlap gate ran against a base two commits stale - so this merges green and lands red on `main`, the #1763/#1766 composition shape this PR's own description narrates for #2313. The fix is the one this comment already names as "the single line that changes if it grows one": forward `tool_context` here, update the two now-stale comments (this one and the class docstring's "exposes no operator-approval gate" paragraph), rework `test_the_shipped_transports_cover_both_sides_of_the_gate_split` - its premise that `use_rtps` does not gate no longer holds, so the non-vacuity check needs a new ungated exemplar or a different form - then merge `main` into the branch and re-run `tests/mesh/test_mobile_base.py` on the composition.

yinsong1986 (commented):
## Summary

Carries #1722's four-way consolidation over a 417-commit base move: `MobileBaseRobot` + a `Transport` protocol (with `ServiceCapable` / `ActionCapable` as separate optional capabilities) now own validation, the drive/stop contract, the trailing-zero stop, the `init_services` handshake and the `tools` property, and `RosBridgedRobot` / `RtpsRobot` shrink to a transport plus platform-specific seams. The conflict resolution correctly threads #2313's operator gate through the base instead of reverting it, and the guard updates (MRO-derived scan in `test_ros_bridge_command_gate.py`, the fleet-scope re-measurement, the inherit-don't-redefine assertions) are the right shape. The one blocker is that `main` moved again after this branch's base: #2693 (merged ~35 minutes before this head was pushed) gave `use_rtps` the very gate `_UseRtpsTransport` is written to assume it lacks - details inline. The merge is textually clean and branch CI tests the head tree, so nothing on GitHub will flag it; it needs one more base refresh before merging.

## What's good

- The context-forwarding rule is derived from the tool's own signature rather than a maintained table - which is exactly why the composition break below is a *failing test* on the merged tree instead of a silent gate drop. The design caught its own hazard.
- Main's pre-refactor guard suites (drive/navigation domains, gate suite, stop-tool parity) pass unmodified against the consolidated classes, and the consolidation itself is pinned (`INHERIT_THE_DRIVE_STOP_PAIR` absence assertions) rather than merely tolerated.
- Prose that the refactor invalidated (`rosbridge-integration.md` latch claims, the fleet-scope docstring) is updated with its pinning assertions rather than kept false to satisfy a test.
- Changelog fragment, no `_`-prefixed names in `__all__`, ASCII-clean user-facing strings, `__repr__` survives partial construction on both ported classes.

## Verification suggestions

- After the fix: `git merge upstream/main` into the branch, then `pytest tests/mesh/test_mobile_base.py tests/test_every_ros_transport_consults_the_command_gate.py tests/mesh/test_ros_bridge_command_gate.py` on the result. Measured on today's composition, everything outside `test_mobile_base.py` (291 tests across the gate/blocklist/fleet-scope/parity suites) already passes - the break is isolated to the two RTPS context-split tests.

strands_robots/mesh/rtps_robot.py
cagataycali: Done, and the base merge this asked for is now in: `87ca88e2` (merge) + `52776e64`.

Measured on the composed tree, not on the head alone:

- `tests/mesh/test_mobile_base.py`: **101 passed** (was 2 failed / 99 passed before the forwarding fix, exactly as reported).
- 14 affected files including both new gate suites: **753 passed, 1 skipped, 0 failed**.
- `tests/mesh` whole directory against a pristine `origin/main` worktree, identical selection: **135 failing node IDs on each, sets byte-identical** - zero branch-only, zero base-only (all missing optional deps: `awsiot`, lerobot family).
- `mypy strands_robots tests tests_integ`: **36 == 36**, message sets identical, zero in the changed files. `ruff check` + `ruff format --check` clean over 1528 files.

On the three sub-points:

1. **Forwarding** - `tool_context` now reaches `use_rtps`. 
2. **The two stale comments** - both rewritten. The class docstring no longer says the tool "exposes no operator-approval gate"; it names `_command_gate` and states why dropping the context would be the same class of defect. 
3. **The gate-split non-vacuity check** - the premise that `use_rtps` does not gate is gone, so the assertion that both shipped transports sit on opposite sides of the split could not be repaired, only replaced. It now asserts the property that actually matters and survives a third transport: *a transport forwards the context exactly when its underlying tool gates*, parametrized over the shipped transports, with the ungated side supplied by an in-test exemplar instead of by whichever real tool happens not to gate yet. That is what keeps it non-vacuous now that both real ones gate.

One thing the merge surfaced that is **not** in this diff, deliberately. `main` gained a fourth drive-owning class in #2689, so the four conflicts were all set-membership and resolved as unions - `AckermannRosRobot` keeps its own drive/stop pair and joins the *owning* sets, while `RosBridgedRobot`/`RtpsRobot` stay asserted *absent*, which is what pins the consolidation. But #2689 also shipped `max_speed`, `max_steering_rad` and `max_duration` on that class while `docs/rosbridge-integration.md` still marks "Velocity clamps" and "Loud duration rejection" as **(this bridge only)**. That is false on `main` today, independent of this branch, and `tests/mesh/test_drive_contract_fleet_scope.py` - the guard written to catch exactly this - passes **16/16 on pristine `main`** because its `_BRIDGES` list is three Twist bridges, so the fourth class is outside what it measures. Filed separately rather than widened into this PR, since it is neither caused nor fixed by the consolidation.

cagataycali:
## Review of `52776e64`: the forwarding fix is correct, but pinned only by a source scan

The `_UseRtpsTransport.publish` fix is right and the four conflict resolutions
union correctly (`AckermannRosRobot` commands a steering angle and a throttle
rather than a body twist, so it belongs with the drive/stop owners; the two
layered bridges belong in the inherit sets). `tests/mesh/test_mobile_base.py` is
101 passed at this head.

Three follow-ups, each measured on `52776e64`.

### 1. [MUST FIX] A one-token change restores the defect and nothing fails

`test_a_transport_forwards_the_context_exactly_when_its_tool_gates` decides
forwarding with `_forwards_a_context`, which walks the method source for any call
carrying a `tool_context=` keyword. It does not look at what is passed. So
forwarding a literal is indistinguishable from forwarding the parameter:

```
# strands_robots/mesh/rtps_robot.py, the single forward
-            tool_context=tool_context,
+            tool_context=None,
```

```
$ pytest tests/mesh/test_mobile_base.py -q
101 passed
```

That mutation is the reported defect exactly: the operator becomes unreachable
again and every `RtpsRobot` command at a blocklisted surface fails closed with
`No tool_context available for operator approval`, whose only offered remedy is
the blanket bypass. The whole fix currently rests on a check a rename or a
default defeats.

The gap is that nothing in the file asserts an *outcome*. An end-to-end case
closes it, and it needs no DDS domain - keep the real `use_rtps` and double the
backend beneath it, the boundary `tests/tools/test_use_rtps.py` already doubles
(`_backend.available`, `_backend.writer`, `time.sleep`, `rtps.idl.get_type`).
That is the same shape `tests/mesh/test_ros_bridge_command_gate.py` uses for
`use_ros`, so both gate suites would present the same thing to the tool.

What such a section measures today, driving `RtpsRobot("tb", "/cmd_vel")`:

| operator | prompted about | status | reported | samples written |
|---|---|---|---|---|
| approves (`y`) | `['/cmd_vel']` | success | `published 1 message(s) to /cmd_vel` | 1 |
| declines (`n`) | `['/cmd_vel']` | error | `publish to '/cmd_vel' was declined by the operator.` | 0 |
| none present | never asked | error | `No tool_context available for operator approval` | 0 |

Before the forwarding fix all three rows were byte-identical to row 3 - the
operator's answer was not an input to the outcome, which is what made the
severity worth the fix. Four cases (prompt, approval, refusal, and the trailing
zero of a multi-message drive reaching the same operator) catch the mutation
above: 3 failed rather than 101 passed. The headless row is unchanged by the fix
and is worth keeping as the boundary control.

### 2. The non-vacuity check no longer exercises the branch it exists for

`test_the_shipped_transports_cover_both_sides_of_the_gate_split` now asserts only
that the gating set is non-empty, and the docstring's closing claim ("this test
verifies that at least one transport is in the gating set so the rule is not
vacuous") is a weaker guarantee than its name promises. Measured at this head:

```
gating : ['_UseRosTransport', '_UseRtpsTransport']
ungated: []
```

So the `is gates` rule only ever evaluates the `True` branch, and the trailing

```python
    if ungated:
        # Non-vacuity: both branches are exercised.
        pass
```

is unreachable. `ruff check` passes on it, so CI will not report it either.

The underlying problem is that reading the split off the shipped tools makes the
check decay silently the moment the last ungated tool grows a gate - which is
what just happened. Grading constructed pairings instead keeps both directions
pinned however the shipped set moves: a local gating stand-in and an ungated one,
crossed with a transport that forwards and one that does not, asserting the rule
reports a violation for exactly the two wrong pairings. That also lets the
positive property be stated directly and named for what it is, which reads as the
stronger security claim: no transport onto a ROS 2 graph is a way around the gate
on the others.

If the rule is extracted into one predicate returning the offending verbs, the
parametrized matrix and the constructed check grade the same code rather than two
restatements of it.

### 3. `docs/security.md` still enumerates two forwarding robots

Unchanged at this head:

> `RosBridgedRobot` and `AckermannRosRobot` both forward the operator context into it, so a bridged robot driven by an agent prompts too, while a programmatic `robot.drive(...)`/`stop()` needs the surface pre-approved.

`RtpsRobot` forwards as of this branch, so the sentence undercounts by one. A
reader deciding whether to add an RTPS robot's `cmd_vel` to
`STRANDS_ROS2_COMMAND_ALLOW` gets the wrong answer from it. Naming all three, and
saying the prompt happens whichever transport carries the command, keeps the page
consistent with the guarantee the branch now provides.

Relatedly, the shared-contract bullet added to `docs/ros2-integration.md` ends
"and a transport whose tool has no gate has nothing to forward". The rule is
right, but no shipped transport is in that case any more, so it reads as an
exemption that exists. Stating that all three graph tools gate today makes the
posture explicit.

yinsong1986 (dismissed):
## Summary

Carries #1722's consolidation over the moved base: `MobileBaseRobot` + a `Transport` protocol (with `ServiceCapable` / `ActionCapable` as separate optional capabilities) now own validation, the drive/stop contract, the trailing-zero stop, the `init_services` handshake and the `tools` property, and `RosBridgedRobot` / `RtpsRobot` shrink to a transport plus a name grammar and platform-specific seams. The concern from the prior review round - `_UseRtpsTransport.publish` dropping the operator context after #2693 gave `use_rtps` its gate - is resolved at this head and pinned by outcome (`test_rtps_robot_command_gate.py` asserts the operator's answer decides what reaches the wire, not a source scan for a `tool_context=` token). Verified against this head: the merge base equals `main`'s current tip, so the head tree *is* the composed tree; the affected mesh suites (`test_mobile_base.py`, both command-gate suites, stop-tool parity, numeric domains, fleet scope, `test_ros_bridge.py`, `test_rtps_robot.py`) pass 284/284, the `use_ros`/`use_rtps` tool suites and blocklist/HITL doc pins pass 229/229, and the repo-wide structural guards touching the new module (repr-survives-partial-construction, ASCII tool results, except-tuple scope, `Raises:` completeness, `__all__` definedness, xref roles, `Args:` completeness, test naming) pass clean, as do `ruff check` and `ruff format --check` on the changed trees.

## What's good

- Existing public constructor signatures are preserved exactly (positional order included); every new parameter (`max_linear` / `max_angular` / `max_duration` on both ported classes, `tool_context` on command verbs) is additive with a safe default, so nothing here is a silent behaviour break - and the one deliberate behaviour change (the trailing zero going fleet-wide) is safety-positive, documented in both doc pages, and pinned on the wire rather than on class layout.
- The consolidation is asserted rather than tolerated: `INHERIT_THE_DRIVE_STOP_PAIR` / `_INHERIT_THE_CONTRACT` pin the *absence* of redefined `drive`/`stop` on the ported classes, and the gate-forwarding rule is graded on four constructed pairings so both wrong directions stay covered however the shipped tool set moves.
- The validation-before-side-effect ordering (finite check, then duration ceiling, then handshake, then publish) is stated as a safety property and pinned (`test_invalid_request_is_refused_before_the_handshake_runs`), and `stop()` stays reachable without the service graph while still passing the surface-keyed operator gate - the carve-out reasoning is written where the next reader needs it.

## Verification suggestions

- Standard CI covers this; if spot-checking beyond it, `pytest tests/mesh/test_rtps_robot_command_gate.py tests/mesh/test_ros_bridge_command_gate.py -q` is the pair that exercises the real gates end-to-end with only the backends doubled.

tests/mesh/test_rtps_robot_command_gate.py
```diff
@@ -0,0 +1,206 @@
...
+from typing import Any
+from unittest.mock import MagicMock
+
+import pytest
+
+import strands_robots.rtps.idl as idl_mod
+import strands_robots.tools.use_rtps as rtps_mod
+from strands_robots.mesh import RtpsRobot
+
+_TWIST = "geometry_msgs/msg/Twist"
```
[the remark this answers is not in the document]
cagataycali: Fixed in `63ba04da`, and the alert is correct: `_TWIST` is defined once at module level and never read.

It was a leftover from an earlier draft of this suite. There is no call site it belongs at, so deleting it is the fix rather than wiring it somewhere:

- `idl_mod.get_type` is monkeypatched to `lambda ros_type: _Twist`, which ignores its argument and returns the module-level fake dataclass. The type *name* never reaches the doubled IDL layer.
- `_turtle()` calls `RtpsRobot.from_rtps(node_name=..., cmd_vel_topic=...)` and takes the default `cmd_vel_type`, so it never names a type string either.

Deleting it also removes a readability hazard I would rather not leave in a security-gate suite: the dead string `_TWIST` sat one case-fold away from the live `_Twist` dataclass in the same file, so a future reader wiring "the Twist one" into a case had two plausible referents and only one correct answer.

That the deletion is inert is measured rather than asserted. `tests/mesh` on this head against the same tree with the constant restored, identical selection: **135 failing node IDs on each, sets byte-identical** - zero delta in either direction (all missing optional deps, `awsiot` and the lerobot family, unchanged from the baseline this PR's description reports). The three suites that own this surface - `test_rtps_robot_command_gate.py`, `test_ros_bridge_command_gate.py`, `test_mobile_base.py` - are **128 passed**. `ruff check` and `ruff format --check` clean on the changed file, and `mypy` reports no issues in it.

No behavioural pin is owed for this one: removing an unreferenced constant cannot change an outcome, and the gate suite's own graders (the nine outcome cases, and the four constructed pairings behind `gate_forwarding_violations`) are untouched by the diff, which is two deleted lines and nothing else.

yinsong1986 (dismissed):
## Summary

Carries #1722's consolidation over the moved base: `MobileBaseRobot` plus a `Transport` protocol (with `ServiceCapable` / `ActionCapable` as separate optional capabilities) now own name validation, the drive/stop contract, the trailing-zero stop, the `init_services` handshake and the `tools` property, while `RosBridgedRobot` / `RtpsRobot` shrink to a transport, a name grammar and platform-specific extras. The conflict resolution correctly carries #2313's operator gate into the base rather than reverting it: every command verb (`publish`, `service_call`, `action_send_goal`) threads `tool_context` from the `@tool(context=True)` closures through the base to the transport, the trailing zero carries the same context as the command it undoes, and both shipped transports onto a ROS 2 graph now gate. Constructor changes are additive (`max_linear` / `max_angular` / `max_duration` new, nothing dropped), so existing callers are unaffected.

## What's good

- The concerns from prior rounds are all verified resolved at this head: `_UseRtpsTransport.publish` forwards `tool_context` and the R3 outcome suite (`test_rtps_robot_command_gate.py`) pins it on operator answers rather than a token scan; the dead `_TWIST` constant is gone.
- Main's pre-refactor guard suites pass unmodified against the consolidated classes, and the structural gate guard now walks the bridge's MRO with a membership non-vacuity check, so `drive`/`stop` moving to the base cannot silently drop out of coverage.
- The inherit-set assertions in `test_bridge_stop_tool_parity.py` and `test_drive_command_numeric_domains.py` pin the consolidation itself (absence of a re-grown hand-rolled `drive`), not just the current class set.
- Numeric validation delegates to the shared domains with a test asserting the delegation verbatim, per AGENTS.md's shared-guard rule; `__repr__` survives partial construction; changelog fragment present; strings ASCII-clean.
- Independently re-ran the drive/gate/parity/fleet-scope suites, the navigation-goal guards, the repr guard and the blocklist doc guard against this head: 632 passed, 0 failed; `ruff check` and `ruff format --check` clean on the changed files. The two commits `main` has moved since the last base merge (#2701, #2702) are Isaac/verify-dataset only - disjoint from `mesh/`.

yinsong1986 (approved):
## Summary

Carries #1722's consolidation over the moved base: `MobileBaseRobot` + a `Transport` protocol (with `ServiceCapable` / `ActionCapable` as separate optional capabilities) now own name validation, the drive/stop contract with its NaN/duration guards, the trailing-zero stop, the `init_services` handshake and the `tools` property, while `RosBridgedRobot` and `RtpsRobot` shrink to a transport plus a name grammar and platform-specific surfaces (Nav2 goal, RTPS advertise). The conflict resolution is the substantive part and it goes the right way: #2313's operator gate is carried *into* the base rather than reverted - `drive`/`stop`/`enable`/`navigate` thread `tool_context` through the transport to the underlying tool, the trailing zero carries the same context as the command it undoes, and the forwarding expectation is derived from each tool's own signature (`gate_forwarding_violations`, graded on four constructed pairings) rather than a list that decays when the shipped set moves. Verified on the head (which contains all of current `main`, so the tested tree is the merge result): `tests/mesh/test_mobile_base.py` 105 passed; the ten affected drive/gate/parity/scope suites 423 passed, 8 skipped; the new RTPS gate suite 13 passed, 1 skipped (the cyclonedds premise pair, as documented); a broad `tests/mesh` sweep green apart from a pre-existing `awsiot`/`awscrt` optional-dep import error unrelated to this diff; `ruff check` and `ruff format --check` clean over the touched trees. Guard updates pin the consolidation in both directions (owning set asserted, inherited classes asserted *absent*), the fleet-scope guard's prose moved with the measurement exactly as its docstring predicts, and the repr-guard population follows the class move. Safety-behaviour changes (RosBridgedRobot/RtpsRobot gaining the trailing zero) are disclosed in the changelog fragment and both doc pages rather than landing silently. No blockers found.

## What's good

- The naive conflict resolution would have shipped the ROS 2 bridge's command surface ungated; the resolution instead lands the gate once for every transport, and both wrong pairings (gating tool not forwarded to, non-gating tool forwarded to) are failing tests rather than review lore.
- Guards assert the *absence* of the consolidated classes from the drive-owning sets, so a subclass re-growing a hand-rolled `drive` fails structurally instead of silently opting out.
- Numeric domains delegate to the shared guards verbatim (message included), per AGENTS.md's one-owner rule; `positive_finite` is a thin ValueError adapter, not a second rulebook.
- Validation-before-side-effect ordering (refuse before the enable handshake) is stated as a safety property and pinned.
- Unset limits documented as "no limit declared", never a silent default - consistent with the no-silent-defaults rule.

PR #2698, opened 2026-08-24
fix(sim): settle run_policy's own two rates before it opens a recording

cagataycali:
`run_policy` (the tool) is the only surface in the package that takes both a dataset frame rate
and a rollout capture rate as its own arguments. `dataset_fps` becomes the LeRobotDataset's
declared rate; `control_frequency` is the rate the recorder is really driven at, one frame per
control step with no decimation. LeRobot derives every timestamp positionally from the declared
rate, so the two must be EQUAL - a differing pair cannot be honored, only mislabelled.

Each rate was already checked on its own domain. Their equality was not, and the tool is the one
caller that cannot afford to leave it to the rollout: it opens its recording with
`overwrite=True` before the episode loop, so the refusal arrived from inside the loop, after the
caller's dataset had already been replaced with an empty one.

## Measured

A real MuJoCo rollout records one episode of five frames at 30 fps, then the tool is called with
`control_frequency=50.0, dataset_fps=30, dataset_root=<that dataset>`:

| `meta/info.json` at `dataset_root` | before this change | with this change |
| --- | --- | --- |
| before the call | `total_episodes=1  total_frames=5  fps=30` | `total_episodes=1  total_frames=5  fps=30` |
| after the call | `total_episodes=0  total_frames=0` | `total_episodes=1  total_frames=5` |
| per-camera MP4 | removed | present, byte-identical |
| dataset digest | changed | unchanged |
| reported | `run_policy: 0/2 episodes ok \| parquet-truth: total_episodes=0, total_frames=0` | names both rates, the distortion and both remedies |

Neither argument was wrong on its own, so nothing refused the pair: the caller lost the dataset
and recorded nothing in its place. Both calls return `status="error"`, so the status alone did not
distinguish them.

## Why the existing guards could not see it

The disagreement already has two orderings and both are refused - a rollout started against an
open recording (`dataset_rate_mismatch_reason`) and a recording opened against a rollout in
flight (`rollout_rate_mismatch_reason`). Each reads one rate off live state, so neither can be
asked before either exists. This is the third ordering: both rates arrive as arguments to one
call, while nothing is open.

The per-knob pre-flight sweep in `tests/tools/test_run_policy_rollout_knob_preflight.py` is
structurally blind to it too. It derives the numeric knobs forwarded to
`Simulation.run_policy(...)` - `n_steps`, `control_frequency`, `action_horizon` - and a rule
*between* two parameters is not a property of either. `dataset_fps` is forwarded to
`start_recording`, not to the rollout.

## The change

`requested_rate_mismatch_reason` joins its two siblings in `strands_robots/simulation/recording.py`
and reuses `rate_mismatch_explanation`, so a caller who reorders two calls or supplies both rates
at once gets one account of the distortion rather than three. A parity test now holds all three
orderings to the same verdict on the same value grid. Only the reason form is shipped: the one
caller reports through a structured-error helper of its own, so an envelope twin would be dead
code.

Because this guard runs *before* either rate has been through its own domain, its accepted set is
derived from those domains rather than guessed. Both rates are classified on `numbers.Real`,
exactly as `positive_whole_number_error` and `positive_finite_number_error` do, so the
`numpy.int64` / `numpy.float32` spellings a value read out of a config carries are judged too - an
`isinstance(int | float)` narrowing accepts neither, and would have passed a colliding pair
straight through (12 failing cases in the derived-parity test). The boolean question goes to the
shared `is_boolean` predicate: `bool` *is* a `numbers.Real`, so a flag would otherwise have been
diagnosed as a genuine 1 Hz rollout. A value either domain refuses returns no reason here, so it
is still reported as the parameter error it is rather than as a rate disagreement.

The check is gated on a requested recording. `dataset_fps` is forwarded nowhere when
`dataset_root` is `None`, so the documented recording-less smoke-test mode still runs at any rate
against the default `dataset_fps=30`.

## Artifact

The dataset at `dataset_root` before and after the same call, on each tree. Both panels in each
row are the first frame decoded from the dataset's own per-camera MP4; the two trees' starting
recordings agree to 0 levels. On the left the file the right column still decodes does not exist.
MuJoCo headless (`egl`), mock policy, 5-step rollouts.

dataset before and after run_policy(dataset_fps=30, control_frequency=50.0), on both trees

Round trip on the honorable pair (`control_frequency=30.0, dataset_fps=30`, `n_episodes=2`):
`status="success"` and `meta/info.json` reads `total_episodes=2, total_frames=10`, pinned by
`test_an_agreeing_pair_records_the_episodes_it_was_asked_for`.

## Tests

`tests/tools/test_run_policy_rate_agreement_preflight.py` (new, 276 cases). Pre-fix split, with
the new helper present and only the tool's guard reverted: **12 failed / 264 passed** - 8 in
`TestTheRequestedRatesMustAgree`, 2 in `TestTheRefusalPrecedesTheDestruction`, 2 in
`TestThePairIsJudgedBeforeTheRecordingOpens`, and none in either control class. Reverting both source files instead
fails at collection on the absent symbol.

Break table - every mutation and the tests that catch it:

| mutation | tests firing | where |
| --- | --- | --- |
| control, unmodified | 0 | - |
| both source files reverted | collection error | absent symbol |
| only the tool's guard reverted | 12 | the three regression classes |
| checked even when no recording is requested | 3 | `TestUsablePairsStillRun` (recording-less rollouts) |
| exact equality replaces the `1e-9` tolerance | 2 | the near-equal accepted pair |
| an unwritable `fps` diagnosed here, not left to its own guard | 47 | derived parity + helper domain |
| `isinstance(int \| float)` replaces `numbers.Real` | 12 | derived parity (every NumPy spelling) |
| hand-rolled `bool` check replaces `is_boolean` | 1 | the shipped `TestTheBooleanDomainIsStructurallyClosed` |
| the remedy hardcodes `fps` over the caller's spelling | 2 | remedy wording |
| the recording-side remedy is never offered | 2 | remedy wording |
| its own account of the distortion replaces the shared one | 5 | the 3-way parity test + the shared-account test |

The last row and the `is_boolean` row fire tests that are not new: the widened parity test and the
repo's own boolean-domain contract. The two remedy rows fire the same pair, because both remove
the same clause from the message and the tests cannot distinguish them.

`requested_rate_mismatch_reason` is added to `_GUARDED_VALIDATORS` in
`tests/simulation/test_input_validators_refuse_a_boolean.py` rather than exempted like its two
siblings: they compare rates their callers have already validated, and this one does not.

## Gate

`ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ`. mypy **24 == 24**
against a pristine `upstream/main` worktree, message sets byte-identical, none in the changed
files.

A 376-file selection (every test that mentions recording, a rate, `run_policy` or `fps`), the same
selection on both trees with the two changed test files excluded from each:
**`53 failed, 14055 passed, 87 skipped, 126 errors` on both** (627.14s vs 627.33s), 180 failing ids
each, both `comm` sets empty. The failures are pre-existing in this environment: 302 occurrences of
`No module named 'device_connect_edge'` (declared as `device-connect-edge>=0.2.0` in
`[device-connect]`) and 18 of `awsiot` (`awsiotsdk>=1.21.0,<2.0.0` in `[mesh-iot]`), with the
remainder identical on the pristine base.

The three changed test files: **620 passed** on mujoco 3.12.0 with lerobot 0.5.1, and **620 passed**
again on mujoco 3.5.0 (the declared floor) with lerobot 0.6.2 built from `main`.

### DIFF
--- a/changelog.d/2698-run-policy-tool-rate-agreement.md
+++ b/changelog.d/2698-run-policy-tool-rate-agreement.md
@@ -1,0 +1,41 @@
+### Fixed: the `run_policy` tool settles its own two rates before it opens a recording
+
+The `run_policy` tool owns both halves of one rule. `dataset_fps` becomes the LeRobotDataset's
+declared frame rate; `control_frequency` is the rate the recorder is really driven at, one frame
+per control step with no decimation. LeRobot derives every timestamp positionally from the
+declared rate, so the two must be EQUAL - a differing pair cannot be honored, only mislabelled.
+
+Each rate was already checked on its own domain: `control_frequency` by the tool's pre-flight,
+`dataset_fps` by `start_recording` before it touches the target directory. Their *equality* was
+left to the rollout entry point, which this tool reaches only inside the episode loop - after
+`start_recording(overwrite=True)` has replaced whatever was at `dataset_root`. Measured over a
+real MuJoCo rollout that had recorded one episode of five frames, `dataset_fps=30` with
+`control_frequency=50.0` took `meta/info.json` from `total_episodes=1, total_frames=5` to
+`0, 0`, removed the per-camera MP4, and reported `run_policy: 0/2 episodes ok`. Neither argument
+was wrong on its own, so nothing refused the pair; the caller lost the dataset and recorded
+nothing in its place. The tool states this principle for every other knob it forwards - `seed`
+("reached NumPy inside the loop after step 2 had already created a dataset"), `video`, the
+provider keyword bags and `stop_when` - and the per-knob sweep that pins them could not see this
+one, because it derives the knobs forwarded to `Simulation.run_policy(...)` and a rule between
+two parameters is not a property of either.
+
+This is the third ordering of a disagreement whose other two are already refused: a rollout
+started against an open recording (`dataset_rate_mismatch_reason`) and a recording opened against
+a rollout in flight (`rollout_rate_mismatch_reason`). Each of those reads one rate off live state,
+so neither can be asked before either exists. The new `requested_rate_mismatch_reason` covers the
+case where both rates arrive as arguments to one call, and it reuses `rate_mismatch_explanation`,
+so a caller who reorders two calls or supplies both at once gets one account of the distortion
+rather than three. All three orderings are now held to the same verdict by a parity test.
+
+Because the new guard runs *before* either rate has been through its own domain, its accepted set
+is derived from those two domains rather than guessed: both rates are classified on
+`numbers.Real`, exactly as `positive_whole_number_error` and `positive_finite_number_error` do, so
+the `numpy.int64` / `numpy.float32` spellings a value read out of a config carries are judged too -
+an `isinstance(int | float)` narrowing would have passed a colliding pair straight through. The
+boolean question goes to the shared `is_boolean` predicate, since `bool` is a `numbers.Real` and
+would otherwise be diagnosed as a genuine 1 Hz rollout. A value either domain refuses returns no
+reason here, so it is still reported as the parameter error it is rather than as a rate
+disagreement.
+
+The check is gated on a requested recording: `dataset_fps` is forwarded nowhere when
+`dataset_root` is `None`, so the documented recording-less smoke-test mode still runs at any rate.
--- a/docs/recording.md
+++ b/docs/recording.md
@@ -75,6 +75,28 @@ matches one of them: their frames interleave into one episode whose single
 declared rate cannot describe both, so there is no `fps` to pass instead. Stop
 all but one, or restart them at a common `control_frequency`.
 
+The `run_policy` tool owns both rates at once - it takes `dataset_fps` and
+`control_frequency` as its own arguments and drives the whole
+`start_recording` -> rollout -> `stop_recording` cycle - so it settles the pair
+up front, before it opens anything. That matters because it opens its recording
+with `overwrite=True`: left to the per-episode rollout, the disagreement was
+reported only after any dataset already at `dataset_root` had been replaced with
+an empty one, and after every episode had failed with the same message.
+
+```python
+run_policy(simulation=sim, robot_name="so100", n_episodes=2,
+           control_frequency=50.0, dataset_fps=30,      # cannot both hold
+           dataset_root="/tmp/my_dataset")
+# -> "run_policy: this call would open a recording declaring 30 fps for a
+#     rollout capturing at control_frequency=50 Hz. [...] Align the two rates:
+#     pass control_frequency=30, or record at the rollout's rate
+#     (dataset_fps=50)."
+# The dataset already at /tmp/my_dataset is untouched.
+```
+
+`dataset_fps` is ignored when `dataset_root` is omitted (the recording-less
+smoke-test mode), so a rollout with no recording runs at any rate.
+
 ## Selecting which cameras to record
 
 By default every camera in the scene is recorded into the dataset - including
--- a/strands_robots/simulation/recording.py
+++ b/strands_robots/simulation/recording.py
@@ -27,12 +27,14 @@ enforceable protocol.
 """
 
 import logging
+import math
+import numbers
 import shutil
 from collections.abc import Mapping
 from pathlib import Path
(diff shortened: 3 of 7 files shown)


yinsong1986 (dismissed):
## Summary

This PR closes the third ordering of the recording/rollout rate disagreement: `run_policy` owns both `dataset_fps` and `control_frequency` in one call, and it opens its recording with `overwrite=True` before the episode loop where the existing rollout guard fires - so a disagreeing pair used to wipe the dataset at `dataset_root` and then report `0/2 episodes ok`. The new `requested_rate_mismatch_reason` judges the pair in the tool's pre-flight, before `start_recording` touches the target, and only when a recording is actually requested (`dataset_root` set), so the recording-less smoke-test path still runs at any rate. I verified the guard's `1e-9` tolerance matches both sibling guards byte-for-byte, so the pre-flight and the in-loop verdicts cannot diverge on a near-equal pair; the `numbers.Real` classification plus the shared `is_boolean` predicate mirrors what the two per-parameter domains accept (the 15x15 parity grid pins this); and the refusal text is ASCII and carries the shared `rate_mismatch_explanation`. All modified/added tests pass locally (276 in the new file, 344 across the touched suites) and `ruff check`/`ruff format --check` are clean on the changed files.

## What's good

- The headline regression test measures the actual data loss (parquet truth before/after against a real MuJoCo dataset) rather than just the error string, and every recording call names its own `root` per AGENTS.md rule 15.
- Cross-ordering consistency is pinned: the widened `test_every_ordering_reaches_the_same_verdict_and_explanation` keeps all three guards agreeing on verdict and explanation.
- The structural AST scan includes a planted-defect control (`test_the_scan_catches_a_removed_guard`), so a blind scan can't pass silently.
- The `_GUARDED_VALIDATORS` exemption entry documents why this guard, unlike its two siblings, really does coerce raw caller input.

## Verification suggestions

- `MUJOCO_GL=osmesa python -m pytest tests/tools/test_run_policy_rate_agreement_preflight.py -q` (the `recording_sim` tests need a working headless GL backend; the fixture's own `setdefault("MUJOCO_GL", ...)` runs after `importorskip("mujoco")`, so the env must be set externally as CI does).
- Spot-check the docs example: the refusal text in `docs/recording.md` matches the shipped message format from `requested_rate_mismatch_reason`.

yinsong1986 (approved):
## Summary

This PR adds the third ordering of the recording/rollout rate-agreement rule: `run_policy` (the tool) is the one surface that owns both `dataset_fps` and `control_frequency` in a single call and opens its recording with `overwrite=True` before the episode loop, so a disagreeing pair used to be refused only after the caller's dataset at `dataset_root` had been replaced with an empty one. The new `requested_rate_mismatch_reason` in `strands_robots/simulation/recording.py` judges the pair in the tool's pre-flight, gated on `dataset_root is not None` so the recording-less smoke-test path still runs at any rate, and reuses `rate_mismatch_explanation` so all three orderings give one account of the distortion (pinned by the widened three-way parity test). The guard's accepted set is derived from the two per-parameter domains rather than guessed - `numbers.Real` classification plus the shared `is_boolean` predicate, pinned by the 15x15 derived-parity grid - and its `1e-9` tolerance matches both siblings, so the pre-flight and the in-loop verdicts cannot diverge on a near-equal pair. A value either domain refuses returns `None` here and stays that parameter's own error, so the refusal ordering with the tool's existing `control_frequency` guard and `start_recording`'s `fps` guard is coherent.

## What's good

- The headline regression tests measure the actual loss and its prevention (parquet truth before/after against a real MuJoCo dataset, plus a round-trip reopen), with every recording call naming its own `root` under `tmp_path` per AGENTS.md rule 15.
- The structural AST scan that pins the guard's placement ahead of `start_recording` carries a planted-defect control (`test_the_scan_catches_a_removed_guard`), so a blind scan cannot pass.
- Only the reason form ships - the envelope twin its siblings carry would be dead code for this one caller, and the docstring says so, matching the no-dead-code convention.
- Changelog fragment in `changelog.d/`, refusal text ASCII, `_GUARDED_VALIDATORS` entry documents why this guard, unlike its two exempted siblings, coerces raw caller input.

Issue #1715, opened 2026-07-29
dataset_recorder: an uncommanded action column is recorded as 0.0, and replay re-sends it as a travel-to-zero command


cagataycali:
## Problem

`DatasetRecorder.add_frame` fills every action column declared in the dataset schema, defaulting a column the producer did not command to `0.0`:

```python
# strands_robots/dataset_recorder.py:963-966
for k in self._cached_action_keys:
    v = action.get(k)
    if v is None:
        action_vals.append(0.0)
```

Every action space reaching a recorded rollout is **absolute position** (a LeRobot `<motor>.pos` follower, a MuJoCo position actuator), so a recorded `0.0` is not "no command" - it is a command to travel to zero. And `PolicyRunner.replay` rebuilds a full-width dict from the recorded vector and sends every column:

```python
# strands_robots/simulation/policy_runner.py
action_dict = {action_keys[i]: float(val) for i, val in enumerate(action_vals)}
send_result = self.sim.send_action(action_dict, robot_name=resolved_robot, n_substeps=n_substeps)
```

So a fabricated `0.0` on disk is re-enacted on the robot on replay, at servo speed, with nothing warning at record or replay time. There is no way to distinguish a recorded `0.0` command from a recorded omission after the fact.

## This is wider than it first appears: every multi-robot recording is already affected

The obvious trigger is a policy whose action vector is narrower than the robot's actuator list (#1701, PR #1704), which omits the unmatched actuators so they hold. But that is not the only producer of a partial action dict, and not even the common one.

Every backend's recording hook calls `add_frame` **once per robot**, with only that robot's prefixed keys, while `start_recording` declares the schema with **every** robot's columns:

- `strands_robots/simulation/mujoco/simulation.py` (`_make_recording_hook`) - `act_keyed = {f"{robot_name}__{k}": v for k, v in action.items()}`
- `strands_robots/simulation/isaac/recording.py:507` - same shape
- `strands_robots/simulation/newton/recording.py:384` - same shape

In a two-robot scene, each per-step frame therefore carries half the declared action columns, and the other half is written as fabricated `0.0` commands for a robot that was commanded something else entirely. That happens on the healthy path today, with no narrow policy involved.

[...]

PR #2699, opened 2026-08-24
fix(sim): record an undriven robot's declared state columns as measurements

cagataycali:
A multi-robot recording writes the state columns of every robot the rollout does not drive as `0.0`, in a column the dataset declares as `observation.state`, for every frame, under `status="success"`.

## The defect

`start_recording` declares `observation.state` over EVERY robot in the scene, prefixing each column with its robot's name (`alice__shoulder_pan`). A single-policy rollout's recording hook supplies only the driven robot's observation, so every column belonging to another robot is absent from the frame and `DatasetRecorder.add_frame` takes its `v is None -> action_vals.append(0.0)` fill.

Nothing downstream can tell that zero from a reading. It lands in the same column with the same dtype, so a policy trained on the dataset learns the other robot is permanently at its zero pose, and anything replaying or analysing the episode reads the same.

Reproduced on `704dd33d` - two `so101` arms in one scene, `run_policy(robot_name="arm")`, `fps=10`, `start_recording` then `stop_recording`, dataset reopened through `LeRobotDataset`:

| source | `so101__1` | `so101__2` | `so101__3` | `so101__4` | `so101__5` | `so101__6` |
|---|---|---|---|---|---|---|
| engine's own reading, before the rollout | 0.4653 | -0.3447 | 0.3548 | 0.2475 | 0.3476 | 0.2970 |
| recorded, frame 0 | 0.0 | 0.0 | 0.0 | 0.0 | 0.0 | 0.0 |
| recorded, per-column span over the episode | 0.0 | 0.0 | 0.0 | 0.0 | 0.0 | 0.0 |

The declared column set is `['so101__1'..'so101__6', 'arm__1'..'arm__6']`, so those six are columns the dataset advertises. The disagreement is unbounded: an undriven robot keeps whatever pose it was placed in, and contact with the driven robot moves it further - in this scene `so101__1` reached 1.5042 rad by the end of the episode while its recorded column stayed flat at zero.

## One path in the codebase already had this right

`run_multi_policy`'s synchronized loop builds one merged observation per step covering every robot it drives and hands that to `add_frame`. The three single-policy hooks - MuJoCo's `_make_run_policy_hook`, and Isaac's and Newton's - did not, so which entry point recorded an episode decided whether its state columns were measurements.

## The change

A single owner, `strands_robots.simulation.recording.undriven_robot_state`, reads the scalar state of every robot the frame does not drive through the engine's own `get_observation(robot_name=..., skip_images=True)` - the same call the driven robot's columns come from, at the same step. The three hooks merge it under their existing per-robot prefix, with the driven robot's own reading winning any collision.

Images are deliberately not collected: a camera array is keyed by the camera rather than a robot, and cameras are scoped by `start_recording(cameras=...)`.

A bystander read that the engine cannot serve is reported and skipped rather than ending the episode - the driven robot's columns are the rollout's primary product, and losing a whole episode of them to a bystander's read failure is strictly worse than the fill this replaces. `AttributeError` is in that set on purpose: `engine` is duck-typed here, and an engine that has not built (or has torn down) the state buffers a per-robot read needs surfaces exactly that way.

## What is deliberately NOT changed

The *action* columns of the same frame. No command was issued to a robot this rollout does not drive, so there is no truthful value to write there - #1715 works through why the observation, a `ctrl` read-back and a carry-forward are each wrong, and settling it needs a schema decision about whether two independently-driven robots may share one dataset at all. A state column has no such ambiguity: the robot is in the scene and its joint positions are readable. Splitting the two keeps this landable without pre-empting that decision.

## Tests

`tests/simulation/test_recorded_undriven_robot_state.py`, 14 tests. The regression is a record -> reopen round trip through `LeRobotDataset` on a two-robot MuJoCo scene with `bob` parked at `{shoulder_pan: 0.9, elbow: -0.7}`, asserting its declared columns are the engine's own reading rather than the fill.

Pre-fix split, two ways because the helper is a new symbol:

- Reverting all source: collection fails with `ImportError: cannot import name 'undriven_robot_state'`, which hides the behavioural evidence.
- Reverting only the three call sites and keeping the helper: **6 failed, 8 passed**. All six are in the regression classes; the premises and controls pass on both trees. The headline message is the recorded column itself - `every declared observation.state column of the undriven robot 'bob' was recorded as the 0.0 fill while bob is parked at {'shoulder_pan': 0.9, 'elbow': -0.7}: [[0.0, 0.0], [0.0, 0.0], [0.0, 0.0], [0.0, 0.0], [0.0, 0.0]]`.

Two premise tests keep it from going vacuous: the parked pose must differ from `0.0` in every component, and the schema must actually declare `bob__*` columns. Two controls: the driven robot's own columns still vary across the episode, and a fill that claims a driven robot's column does not overwrite that robot's per-step reading (forced by patching the fill to claim it, with a premise asserting the patch really ran). A structural check derives each backend's hook from its own class and requires it to consult the shared owner, so a fourth backend cannot reintroduce the fill by forgetting it - it fails for all three on the pre-fix tree.

## Verification

- New file: **14 passed**.
- `tests/simulation/newton/test_dataset_recording.py`: **14 passed**. Newton's fixture builds its engine through `__new__` without the Warp stack, so a per-robot state read is not serviceable there and takes the reported-and-skipped path; the assertions in that file are about the declared schema and are unaffected.
- New file + `tests/simulation/isaac/test_dataset_recording.py` + `tests/simulation/mujoco/test_recording_camera_scoping.py`: **46 passed**.
- Blast radius, 343 files (every test naming `start_recording`, `dataset_recorder`, `add_frame`, `get_observation`, `run_policy`, `observation.state` or `_make_run_policy_hook`), identical selection against a pristine `main` worktree in the same environment with the new file excluded from each: **2 failed, 12000 passed, 90 skipped on both trees**, 897.64s vs 897.78s, **failing node-ID sets byte-identical - zero branch-only, zero base-only**. Both pre-existing and unrelated: `newton/test_multi_camera.py::test_bad_position_shape_rejected` (asserts `'3 elements'` against a message reading `'must be a 3-element vector'`) and `test_dds_domain_id_domain.py::test_a_usable_domain_is_still_pinned_into_the_environment` (`DID NOT RAISE ImportError`, an artifact of the DDS stack being installed locally).
- `mypy strands_robots tests tests_integ`: **Success: no issues found in 1520 source files**, and zero branch-only messages against the pristine base worktree.
- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1523 files).

## Artifact

The undriven robot's six declared `observation.state` columns, read back out of the recorded
dataset over a 20-frame episode, on both trees. Dotted lines are the engine's own reading of the
same joints. Two-robot MuJoCo scene, headless, `fps=10`, the undriven arm parked and then nudged
by contact with the driven one.

recorded observation.state of the undriven robot

| measure | main (`704dd33d`) | this change |
|---|---|---|
| largest per-column span across the episode | 0.0000 rad | 1.0536 rad |
| final-frame `abs(recorded - engine's own reading)`, worst column | 1.5486 rad | 1.18e-08 rad |

Main records a flat zero for all six while the joints move through a rad of travel; the branch
agrees with the engine to 1.18e-08 rad. Every number in the table is re-derived from the recorded
traces by the script that draws the figure, which asserts them before rendering.

## Docs

`start_recording`'s docstring on all three backends now states the multi-robot schema rule: every robot's columns are declared, the undriven ones are filled from the engine each step so a declared `observation.state` column is a measurement, and the action columns are a separate question that is unchanged.

### DIFF
--- a/changelog.d/2699-recorded-undriven-robot-state.md
+++ b/changelog.d/2699-recorded-undriven-robot-state.md
@@ -1,0 +1,35 @@
+### Fixed: an undriven robot's declared `observation.state` columns are recorded as measurements
+
+`start_recording` declares `observation.state` over EVERY robot in the scene, prefixing each
+column with its robot's name (`alice__shoulder_pan`). A single-policy rollout's recording hook
+supplies only the driven robot's observation, so every column belonging to another robot was
+absent from the frame and took `DatasetRecorder.add_frame`'s `0.0` fill - in the same column,
+with the same dtype, as a measurement, for every frame of the episode, under `status="success"`.
+
+Nothing downstream can tell that zero from a reading. A policy trained on the dataset learns the
+other robot is permanently at its zero pose, and anything replaying or analysing the episode
+reads the same. The disagreement is unbounded: an undriven robot keeps whatever pose it was
+placed in, and contact with the driven robot moves it further. In a two-`so101` scene where the
+undriven arm sat at `[0.4653, -0.3447, 0.3548, 0.2475, 0.3476, 0.2970]` rad and reached
+`1.5042` rad on its first joint during the episode, all six of its declared columns were `0.0`
+on disk with zero span.
+
+`run_multi_policy`'s synchronized loop already had this right - it builds one merged observation
+per step covering every robot it drives. The three single-policy hooks did not, so which rollout
+entry point recorded an episode decided whether its state columns were measurements. They now
+resolve the columns they do not drive through one shared owner, `undriven_robot_state`, which
+reads each robot's scalar state through the engine's own
+`get_observation(robot_name=..., skip_images=True)` - the same call the driven robot's columns
+come from, at the same step - and merges it under the hook's existing prefix, with the driven
+robot's own reading winning any collision. Camera arrays are not collected: a frame is keyed by
+its camera rather than by a robot, and cameras are scoped by `start_recording(cameras=...)`.
+
+A bystander read the engine cannot serve is reported and skipped rather than ending the episode.
+The driven robot's columns are the rollout's primary product, so losing a whole episode of them
+to a bystander's read failure is strictly worse than the fill this replaces.
+
+The matching *action* columns are deliberately unchanged. No command was issued to a robot this
+rollout does not drive, so no value there is truthful, and settling it needs a schema decision
+about whether two independently-driven robots may share one dataset at all. A state column has
+no such ambiguity - the robot is in the scene and its joint positions are readable - so the two
+are separable and only the answerable half is changed here.
--- a/strands_robots/simulation/isaac/recording.py
+++ b/strands_robots/simulation/isaac/recording.py
@@ -56,6 +56,7 @@ from strands_robots.simulation.recording import (
     DatasetRecordingMixin,
     dataset_recording_option_error,
     dataset_recording_posture_error,
+    undriven_robot_state,
 )
 from strands_robots.utils import name_list_error
 
@@ -130,7 +131,15 @@ class IsaacRecordingMixin(DatasetRecordingMixin):
         every robot (namespaced ``robot__joint`` when more than one robot is
         present, matching the MuJoCo/Newton backends), action columns from
         :meth:`robot_action_keys`, and the RTX cameras registered via
-        ``add_camera``. Camera resolutions are probed from one
+        ``add_camera``. In a multi-robot scene every robot's state columns are
+        declared, and a single-policy rollout fills the ones it does not
+        drive from the engine at each step, so a declared
+        ``observation.state`` column is a measurement rather than a zero
+        the robot is not at
+        (:func:`~strands_robots.simulation.recording.undriven_robot_state`).
+        The matching *action* columns are a separate question - no command
+        was issued to a robot this rollout does not drive - and are
+        unchanged. Camera resolutions are probed from one
         ``get_observation`` call because RTX cameras render at a DLSS-safe
         native size that can exceed the requested output size; declaring the
         probed shape keeps ``add_frame`` from rejecting every image. Per-step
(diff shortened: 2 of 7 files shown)


yinsong1986 (approved):
## Summary

Fixes a real data-corruption defect: in a multi-robot scene the three single-policy recording hooks (MuJoCo, Isaac, Newton) supplied only the driven robot's observation to `add_frame`, so every other robot's declared `observation.state` columns took the `0.0` fill and were recorded as measurements under `status="success"` - indistinguishable downstream from a real zero pose. The fix routes all three hooks through one shared owner, `undriven_robot_state`, which reads each bystander's scalar state through the engine's own locked `get_observation(robot_name=..., skip_images=True)` at the same step and merges it under the hooks' existing per-robot prefix, with the driven robot's own reading winning collisions. The action columns are deliberately left to #1715, which is the right split - a state column has a readable truth this step, an action column does not.

## What's good

- The except tuple `(AttributeError, KeyError, RuntimeError, ValueError)` has no superclass-covered member, and the degraded path warns rather than silently dropping (AGENTS.md exception-clause and no-silent-defaults rules). A `TypeError` from a drifted duck-typed engine would still land in `on_frame`'s existing failure-tolerance handler rather than vanish.
- Thread safety holds: the helper goes through the public `get_observation` facade, which takes `self._lock` on all three engines, and the hook itself runs outside the engine lock in the policy-runner loop.
- The regression test is a genuine record -> reopen round trip through `LeRobotDataset`, names its own `root=` per AGENTS.md rule 15, carries premise and control tests against vacuity, and the AST check derives each backend's hook from its own class so a fourth backend is graded on arrival.
- Changelog fragment present; docstrings updated on all three backends; scope discipline on the action-column question.

One non-blocking observation for awareness: the helper passes `skip_images=True`, but all three engines' T26 recording override flips that back to `False` while a recording with cameras is active - which is exactly when these hooks run - so each bystander read pays a full camera render whose arrays the helper then discards via its shape filter. Correctness is unaffected (the test's `assert skip_images is True` premise holds at the call site, not at the render), but in a camera-heavy multi-robot scene the per-frame render cost multiplies by the bystander count. Worth a follow-up issue if recording throughput matters, not a blocker here.

PR #2700, opened 2026-08-24
fix(mesh): name every class that carries a scoped drive limit

### LINKED ISSUE #2697: docs: the drive-contract scope markers still say "this bridge only" for two limits AckermannRosRobot also carries
## What is wrong

`docs/rosbridge-integration.md` marks two drive guarantees as belonging to the rosbridge bridge alone:

```
148:- **Velocity clamps** (this bridge only): linear and angular are independently
151:- **Loud duration rejection** (this bridge only): `duration` must be positive,
```

and `RosbridgeRobot.drive`'s docstring says the same thing in prose:

> Specific to this bridge: velocities are clamped to ``max_linear`` and ``max_angular``, a hold beyond ``max_duration`` is refused, [...] :meth:`RosBridgedRobot.drive` and :meth:`RtpsRobot.drive` carry none of the three

Both claims are false as of #2689. `AckermannRosRobot` ships with all three of the limits, measured on `origin/main` at `704dd33d`:

```
$ python -c "...ast of strands_robots/mesh/ackermann_robot.py..."
__init__ params: ['self', 'node_name', 'servo_topic', 'scan_topic', 'servo_type',
                  'scan_type', 'wheelbase_m', 'max_speed', 'max_steering_rad',
                  'max_duration', 'publish_rate', 'init_services']
```

and `drive` enforces the ceiling:

```python
# strands_robots/mesh/ackermann_robot.py:311
if duration is not None and duration > self.max_duration:
    ...f"drive: duration {duration}s exceeds max_duration {self.max_duration}s "
```

## Why it matters to an operator

The scope marker is the whole point of those bullets: an unmarked guarantee reads as the fleet's, and a marked one tells a reader which platform they have to supply the ceiling for themselves. A reader who has just been told the duration ceiling is "this bridge only" will not expect `AckermannRosRobot(...).drive(linear=0.5, duration=30.0)` on a DeepRacer to be refused - but it is, by the 10 s default. The failure direction is the safe one, yet it is still an unexplained refusal in the field, and the same reader plans around a clamp they believe is absent on the car when `max_speed`/`max_steering_rad` are in fact clamping.

## Why CI did not catch it

`tests/mesh/test_drive_contract_fleet_scope.py` exists precisely to grade this prose against a measurement, and its docstring commits to it:

> A bridge that later gains the trailing stop makes the scope assertion fail, which is the intended signal: the guarantee has become fleet-wide and the two places that scope it have to say so.

But its universe is a hardcoded list of three:

```python
_BRIDGES: list[tuple[str, Any, str, Callable[[], Any]]] = [
    ("RosBridgedRobot", ...), ("RtpsRobot", ...), ("RosbridgeRobot", ...),
]
```

so the fourth class is outside everything it measures. On pristine `origin/main`:

```
$ pytest tests/mesh/test_drive_contract_fleet_scope.py
16 passed
```

Green, while the claim it grades is false. The guard is not wrong about the three it compares - it is silent about a class it never learned exists, which is the same vacuity shape its own `test_the_three_bridges_are_the_ones_under_comparison` premise check was written to prevent one level down.

## Suggested direction

The two guarantees at issue are wire-format independent, which is what makes this fixable without parametrizing the Twist probes over a servo message: a class carries a velocity clamp if it declares clamp attributes, and a duration ceiling if it declares `max_duration`. Determining those two scopes by surveying every drive-owning mesh class - rather than only the three Twist bridges - would let the prose be graded against the real fleet, and would then require the two surfaces to say which classes actually carry each limit. The wire-content probes (`validated inputs`, `single-shot latch`, `trailing zero Twist`) can stay scoped to the Twist bridges, with the exclusion stated for the reason `test_bridge_stop_tool_parity.py` already states it: an Ackermann car halts with a zero `ServoCtrlMsg`, not a zero `Twist`.

## Provenance

Found while merging current `main` into #2696, which touches `test_drive_contract_fleet_scope.py` for the trailing-zero scope change. Deliberately **not** fixed there: the defect is neither caused nor fixed by that branch's consolidation, it reproduces on pristine `main`, and folding it in would have made a 15-file refactor also a docs-scope redesign. Filing it so the two stay separable.

Refs #2689.

cagataycali:
## What is wrong

`docs/rosbridge-integration.md` and `RosbridgeRobot.drive` both mark two drive
limits as belonging to the rosbridge bridge alone:

```
- **Velocity clamps** (this bridge only): linear and angular are independently
  clamped to `max_linear` and `max_angular`. The other two bridges accept no
  velocity ceiling and put the requested value on the wire unchanged.
- **Loud duration rejection** (this bridge only): ... The other two bridges
  accept any positive finite hold.
```

> Specific to this bridge: velocities are clamped to ``max_linear`` and
> ``max_angular``, a hold beyond ``max_duration`` is refused, [...]
> :meth:`RosBridgedRobot.drive` and :meth:`RtpsRobot.drive` carry none of the three

`AckermannRosRobot` carries both. Measured on `704dd33d` by driving each
drive-owning mesh class with its transport replaced by a recorder:

| class | inputs validated | velocity clamp | duration ceiling |
|---|---|---|---|
| `RosBridgedRobot` | yes | no | no |
| `RtpsRobot` | yes | no | no |
| `RosbridgeRobot` | yes | yes, `max_linear` 2.0 | yes, `max_duration` 30.0 |
| `AckermannRosRobot` | yes | yes, `max_speed` 1.5 | yes, `max_duration` 10.0 |

The costly detail is that the car's ceiling is *lower*, and the page documents
30.0 s as an accepted hold. A reader who takes that number to a DeepRacer gets,
on `704dd33d`:

```
>>> car = AckermannRosRobot.from_deepracer(node_name="deepracer")
>>> car.drive(linear=0.5, duration=30.0)
status: error
drive: duration 30.0s exceeds max_duration 10.0s - issue shorter commands instead of one long hold
published 0 messages
```

The failure direction is the safe one, but it is still an unexplained refusal in
the field - and the same reader plans around a velocity clamp they have been told
the car does not have.

The fleet's own documentation already states both halves and never compares them:
`docs/ros2-integration.md` says of the car "Commands are clamped to `max_speed`;
holds longer than `max_duration` are rejected loudly rather than silently
truncated", while `docs/rosbridge-integration.md` says those limits exist on one
bridge only.

## Why no test caught it

`tests/mesh/test_drive_contract_fleet_scope.py` exists precisely to grade this
prose against a measurement rather than against a hand-kept list, and its
docstring commits to raising exactly this signal. Its universe was a hardcoded
list of three:

```python
_BRIDGES: list[tuple[str, Any, str, Callable[[], Any]]] = [
    ("RosBridgedRobot", ...), ("RtpsRobot", ...), ("RosbridgeRobot", ...),
]
```

so the fourth drive-owning class was outside everything it measured. On pristine
`704dd33d` the file is `16 passed` while the claim it grades is false. The guard
is not wrong about the three it compares; it is silent about a class it never
learned exists.

Its clamp probe could not have been evaluated on that class either, even if the
class had been listed: it read `calls[0]["fields"]["linear"]["x"]`, which raises
`KeyError: 'linear'` on a `ServoCtrlMsg`.

## The change

**The surveyed set is derived from the package.** A class that owns `drive` and
lives in a *public* module of `strands_robots.mesh` is a shipped platform bridge.
The lookup is by attribute rather than by a `def drive` in the class body, so a
bridge that inherits `drive` from a shared base is still found, while a private
`_`-prefixed base module - which declares a contract but is not a platform anyone
drives - stays out. A fifth class fails the inventory until it is given a case.

**Three of the five guarantees are read without touching the message layout**, so
they are measured over every drive-owning class:

- `validated inputs`: a refused value forwards no call at all.
- `duration ceiling`: a hold past the ceiling forwards no call at all.
- `velocity clamp`: two *different* over-ceiling requests forward the *identical*
  call. A ceiling makes them indistinguishable whatever the payload looks like,
  so the same probe answers for a servo pair as for a `Twist`.

The two that do read `Twist` content (`single-shot latch`, `trailing zero Twist`)
stay scoped to the three `Twist` bridges, with the exclusion stated for the
reason `tests/mesh/test_bridge_stop_tool_parity.py` already states: an Ackermann
car halts with a zero `ServoCtrlMsg`, not a zero `Twist`. Widening those two is a
separate change and is deliberately not made here.

**The split becomes three-way** - fleet-wide, shared by some, this bridge's alone.
A limit two platforms declare is neither of the outer two, and filing it under
either misinforms a reader of the other platforms.

**Both prose surfaces are corrected.** The two bullets are marked
`(not on every mobile base)` and name `AckermannRosRobot`; the docstring
paragraph opens `Not carried by every mobile base:` and credits the car's
`max_speed` and its own lower `max_duration`. Two derived rules now hold them
there: a bullet may say `this bridge only` only when the survey finds no other
carrier, and a scoped guarantee's every other carrier has to be named.

No library behaviour changes - the only non-test edit is a docstring.

## Pre-fix

With both prose surfaces restored to `704dd33d` and the survey as shipped here:

```
11 failed, 16 passed, 4 skipped
```

All 11 are the prose-grading tests; 0 are in the premise or inventory classes.

## Break table

Each row is a single mutation applied to the branch, then
`pytest tests/mesh/test_drive_contract_fleet_scope.py`.

| mutation | fails | what it shows |
|---|---|---|
| none (as shipped) | 0 of 27 | control, plus 4 skips where a guarantee is fleet-wide |
| a fifth drive-owning class added to the package, inventory **derived** | 1 | `test_every_drive_owning_mesh_class_is_surveyed`, naming `'ProbeFifthRobot'` |
| the same fifth class, inventory **hardcoded** to today's four names | **0** | the hole this fixes: 27 pass while a shipped bridge sits outside every scope claim |
| `AckermannRosRobot` dropped from the surveyed set | 4 | inventory, the wider-survey premise, the split, and the two-carrier measurement |
| clamp probe reads `fields["linear"]["x"]` again | 24 | `KeyError: 'linear'` - the old probe cannot be evaluated on a servo bridge at all |
| `(this bridge only)` restored on the clamp bullet alone | 1 | the regression, on that one guarantee |
| the docstring stops crediting the co-holder | 3 | both scoped guarantees plus the "where does each other class stand" rule |
| the scoped paragraph relabelled as fleet-standard (over-reach) | 10 | includes the pre-existing `test_no_paragraph_claiming_a_shared_contract_states_a_bridge_specific_guarantee` |
| probe velocity dropped below the declared ceilings, premise present | 4 | the premise fires and names the cause |
| the same mistake with that premise removed | 3 | fewer failures and no diagnosis - which is why the premise is there |

## Gate

tests_integ`. mypy **24 errors on the branch and 24 on a pristine `704dd33d`
worktree, with byte-identical message sets** and none in the changed files.

Paired run, `tests/` minus the GPU/simulation subtrees (472 files), identical
selection on both trees with the changed test file deselected from each:

| tree | result |
|---|---|
| branch | `101 failed, 18356 passed, 120 skipped, 256 errors in 551.69s` |
| pristine `704dd33d` | `101 failed, 18356 passed, 120 skipped, 256 errors in 573.30s` |

Failing-ID and error-ID sets **identical both ways** (both `comm` directions
empty). Every one is an absent optional dependency in this environment -
`device_connect_edge` (290, declared `device-connect-edge>=0.2.0` in
`[device-connect]`), `awsiot` (44, `awsiotsdk` in `[mesh-iot]`), and a locally
installed lerobot 0.5.1 below the declared `>=0.6.1,<0.7.0` floor.

The changed file itself: `27 passed, 4 skipped`.

## Sequencing

PR #2696 edits the same docstring paragraph and the same docs bullets for a
different reason - it moves the trailing zero into the fleet-wide half once the
`Twist` bridges share a base. The two changes are independent in content: this
one is about the clamp and the ceiling, reproduces on pristine `main`, and would
survive that refactor unchanged. Whichever lands second needs a textual merge in
those two paragraphs; the scope rules added here are derived, so they will grade
the merged prose without further edits.

Fixes #2697.

### DIFF
--- a/changelog.d/2700-drive-limit-scope-names-every-carrier.md
+++ b/changelog.d/2700-drive-limit-scope-names-every-carrier.md
@@ -1,0 +1,43 @@
+### Fixed: the drive-contract scope markers name every class that carries a limit
+
+`docs/rosbridge-integration.md` and `RosbridgeRobot.drive` both marked the velocity clamp and
+the `max_duration` ceiling as belonging to the rosbridge bridge alone, and said in prose that
+`RosBridgedRobot.drive` and `RtpsRobot.drive` "carry none of the three". `AckermannRosRobot`
+declares both, as `max_speed` and a `max_duration` of its own. The scope marker is the whole
+point of those bullets - an unmarked guarantee reads as the fleet's, and a marked one tells a
+reader which platform they have to supply the ceiling for themselves - so a reader who has just
+been told the duration ceiling is one bridge's own carries the 30 s hold the page documents over
+to a DeepRacer and is refused by that car's 10 s ceiling. The failure direction is the safe one,
+but it is an unexplained refusal in the field, and the same reader plans around a velocity clamp
+they have been told the car does not have. The fleet's own pages already stated both halves:
+`docs/ros2-integration.md` documents the car's clamp and ceiling while the rosbridge page denied
+they existed anywhere else.
+
+`tests/mesh/test_drive_contract_fleet_scope.py` exists to grade that prose against a
+measurement rather than against a hand-kept list, but its universe was a hardcoded list of the
+three `Twist` bridges, so the fourth drive-owning class sat outside everything it checked and the
+false claim passed green. The surveyed set is now derived from the package - a class that owns
+`drive` and lives in a public module of `strands_robots.mesh` is a shipped platform bridge - and
+the lookup is by attribute rather than by a `def drive` in the class body, so a bridge that
+inherits `drive` from a shared base is still found while a private base module, which declares a
+contract but is not a platform anyone drives, stays out. A fifth class fails the inventory until
+it is given a case.
+
+Three of the five guarantees are read without touching the message layout and are measured over
+every drive-owning class: a refused input and a hold past a ceiling each forward no call at all,
+and a velocity clamp makes two different over-ceiling requests forward the identical call, which
+is as true of a servo pair as of a `Twist`. The clamp probe previously read
+`calls[0]["fields"]["linear"]["x"]` and raises `KeyError: 'linear'` on a `ServoCtrlMsg`, so it
+could not have been evaluated on that class even if the class had been listed. The two probes
+that do read `Twist` content, the single-shot latch and the trailing zero, stay scoped to the
+three `Twist` bridges for the reason `tests/mesh/test_bridge_stop_tool_parity.py` already gives:
+an Ackermann car halts with a zero `ServoCtrlMsg`, not a zero `Twist`.
+
+The measured split is now three-way - fleet-wide, carried by some, carried by this bridge alone -
+because a limit two platforms declare is neither of the outer two and filing it under either
+misinforms a reader of the other platforms. Both surfaces say which is which: the two bullets are
+marked `(not on every mobile base)` and name `AckermannRosRobot`, and the docstring paragraph
+opens `Not carried by every mobile base:` and credits the car's own lower ceiling. Two derived
+rules hold them there - a bullet may claim `this bridge only` only when the survey finds no other
+carrier, and every other carrier of a scoped guarantee has to be named. No library behaviour
+changes; the only non-test edit is a docstring.
--- a/docs/rosbridge-integration.md
+++ b/docs/rosbridge-integration.md
@@ -125,9 +125,11 @@ All parameters are optional except `node_name`, `cmd_vel_topic`, and `odom_topic
 and two of its guarantees are fleet-standard too: every value is checked against
 the same numeric domains the [ROS 2](ros2-integration.md) and
 [RTPS](rtps-integration.md) bridges use, and a bare single-shot command latches.
-The velocity clamps, the `max_duration` ceiling and the trailing zero Twist are
-specific to this bridge - the other two accept no ceilings and publish no
-trailing zero, so a timed drive there leaves the last velocity latched.
+The velocity clamps and the `max_duration` ceiling are shared with the Ackermann
+car bridge (`AckermannRosRobot`, which declares a `max_speed` and a
+`max_duration` of its own) but not with those two, and the trailing zero Twist is
+this bridge's alone - the ROS 2 and RTPS bridges accept no ceilings and publish
+no trailing zero, so a timed drive there leaves the last velocity latched.
 
 ```python
 # Direct, programmatic control:
@@ -145,12 +147,16 @@ scan = robot.get_scan()  # read one laser scan (error if no scan_topic)
 - **Single-shot latch**: a bare single-message `drive()` (no `duration`, no
   `count > 1`) publishes once and latches in the robot's controller until
   `stop()` is called. This is standard cmd_vel behavior.
-- **Velocity clamps** (this bridge only): linear and angular are independently
-  clamped to `max_linear` and `max_angular`. The other two bridges accept no
-  velocity ceiling and put the requested value on the wire unchanged.
-- **Loud duration rejection** (this bridge only): `duration` must be positive,
-  finite, and at most `max_duration`; anything else returns a detailed error and
-  nothing is published. The other two bridges accept any positive finite hold.
+- **Velocity clamps** (not on every mobile base): linear and angular are
+  independently clamped to `max_linear` and `max_angular`. `AckermannRosRobot`
+  clamps to a `max_speed` of its own the same way. The ROS 2 and RTPS bridges
+  accept no velocity ceiling and put the requested value on the wire unchanged.
+- **Loud duration rejection** (not on every mobile base): `duration` must be
+  positive, finite, and at most `max_duration`; anything else returns a detailed
+  error and nothing is published. `AckermannRosRobot` refuses a longer hold
+  against a `max_duration` of its own, which is lower, so a hold this bridge
+  accepts can be refused on that car. The ROS 2 and RTPS bridges accept any
+  positive finite hold.
 - **Timed-command trailing zero** (this bridge only): every drive with a
   `duration` argument and a non-zero command (or multi-message publish)
   automatically publishes a single zero Twist afterwards - even if the main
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Docs/test-only correction of a false scope claim: `docs/rosbridge-integration.md` and the `RosbridgeRobot.drive` docstring marked the velocity clamp and the `max_duration` ceiling as `(this bridge only)`, while `AckermannRosRobot` declares both (`max_speed` 1.5, `max_duration` 10.0 — verified against the shipped implementation, which clamps before `_twist_to_servo` and refuses a hold past the ceiling before any side effect). The guard test's universe was a hardcoded three-bridge list, so the fourth drive-owning class sat outside every scope claim; the surveyed set is now derived from the package (public modules of `strands_robots.mesh`, attribute lookup so an inherited `drive` is found), the clamp probe is rewritten as layout-independent call-equality so it can be evaluated on a servo bridge at all, and the split becomes three-way (fleet-wide / shared-by-some / this-bridge-only). The only non-test library edit is the docstring. Verified at head `77353688`: the changed test file reports `27 passed, 4 skipped`, `ruff check` and `ruff format --check` are clean on both changed Python files, all strings are ASCII, the new `:class:` cross-reference target is contiguous on one line, and a changelog fragment plus a body-level `Fixes #2697` are present per AGENTS.md steps 3 and 6.

## What's good

- The break table in the description distinguishes the derived inventory from a re-hardcoded four-name list (0 failures on the silent variant), which is exactly the hole being closed.
- The two `Twist`-reading probes stay scoped to the three `Twist` bridges with the exclusion stated and cross-referenced to `test_bridge_stop_tool_parity.py`, rather than being silently widened.
- The sequencing note on #2696 (same docstring paragraph, independent content) saves the second-lander a surprise textual merge.

## Verification suggestions

- `pytest tests/mesh/test_drive_contract_fleet_scope.py -q` at head — expect `27 passed, 4 skipped` (the 4 skips are the fleet-wide guarantees in the co-holder-crediting tests).
- Spot-check the inventory's reach: the glob covers top-level `strands_robots/mesh/*.py` only, so if a drive-owning class ever lands in `mesh/iot/` or `mesh/transport/`, `test_every_drive_owning_mesh_class_is_surveyed` will not see it — worth keeping in mind when reviewing future bridge additions, though no such class exists today.

PR #2702, opened 2026-08-24
fix(verify-dataset): grade a dead control column per robot, not per whole vector

cagataycali:
## What is wrong

`verify_dataset`'s dead-control-column check exists to catch a recording whose `action` /
`observation.state` column was written as zeros because the writer's keys never resolved to the
declared columns: correct episode counts, correct pixels, no control signal. It fired only when the
**entire** vector was identically zero:

```python
if all(v == 0 for v in row_min) and all(v == 0 for v in row_max):
```

A multi-robot recording cannot present that shape. `start_recording` declares one column per robot
per joint and prefixes each with the robot's instance name (`mujoco/recording.py`:
`joint_names.extend(f"{rname}__{jn}" ...)`), so a resolution that succeeds for one robot and fails
for another leaves that robot's whole block of columns zero while the other robot's columns carry
real measurements. The vector as a whole varies, the check sees nothing, and the report reads
`[PASS]`.

The recorder documents exactly this mechanism, in the comment that remaps the prefixed keys
(`strands_robots/simulation/mujoco/simulation.py`):

> Without remapping here, `add_frame()` looks up the prefixed schema keys, finds nothing, and
> writes all-zero state/action vectors silently.

That is the per-robot failure. The gate written to catch it could only see the all-robot case.

## Measured on a real recording made on current `main`

Two `so101` arms, `alice` driven for 60 frames and `bob` standing in the scene, recorded through
`start_recording` / `run_policy` / `stop_recording` at `e3ff18dd` and reopened through
`LeRobotDataset` (60 frames, 1 episode, `observation.images.overview` as declared):

| declared column | `action` span | `observation.state` span |
|---|---|---|
| `alice__1 .. alice__6` | live | live |
| `bob__1 .. bob__6` | **0.0, all six** | non-zero, all six |

Same dataset, same command:

```
[PASS]  main @ e3ff18dd
[FAIL]  this change
  - feature 'action' is identically zero for every 'bob' column across
    episode 0 (60 frame(s)) - dead control column block
```

`observation.state` is deliberately **not** reported: #2699 now records an undriven robot's state
as measurements, so that block varies and only the genuinely dead one is named. The two features of
one dataset are graded differently on their own content, which is the discrimination the rule is
for.

dead control column block

Rollout the dataset was recorded from (MuJoCo, headless):
https://raw.githubusercontent.com/cagataycali/robots/media-dead-column-block/two-arm-rollout.mp4

## The rule

Split the vector into the blocks `meta/info.json` declares - column indices grouped by the
per-robot prefix - and report a block that is wholly zero the same way a wholly-zero vector is
reported, naming the robot.

A zero **subset** of a block is deliberately left alone: a gripper parked at zero for a whole
episode is a measurement, not a writer fault. Measured across the LeRobot datasets on one machine,
103 partially-dead stat vectors are subsets of a block, and none of them is flagged.

The separator is the doubled underscore `start_recording` writes, not a single one. A real arm's
joints are `shoulder_pan` / `elbow_flex` / `gripper`; splitting on `_` would turn one arm into four
blocks and flag its parked gripper. That is a test case, not a comment.

## Strict extension

| property | measured |
|---|---|
| a wholly-zero vector | reports the message it reported before, once, not one per block |
| absent / non-string / wrong-width `names` | keeps the whole-vector rule (every single-robot recording) |
| `stats_vectors_checked` | unchanged - same work, finer verdict |
| blocks found for a single-robot recording | none: of 2221 declared control vectors, exactly 200 split, always into 2, and 0 splits occurred where any name lacked the `__` prefix |

Blast radius, real `verify_dataset` over 1314 dataset roots on both trees:

| | base | this change |
|---|---|---|
| pass / fail | 747 / 567 | 689 / 625 |
| pass -> fail | | 58 |
| **fail -> pass** | | **0** |
| base problem strings no longer reported | | 0 |
| new problem strings | | 74, all the per-block finding |
| exceptions raised | 0 | 0 |

The 58 are 48 pytest temporary fixtures and 10 recorded datasets.

## Tests

`tests/test_verify_dataset.py` grows a `TestDeadControlColumnPerRobotBlock` class: 73 tests in the
file, up from 57.

Pre-fix split. Reverting `verify_dataset.py` outright is a collection error, because the suite names
the new block-splitting helper; the behavioural pre-fix - helper present, declaration not plumbed
into the check - is **5 failed / 68 passed**, with all 5 in the regression cases and 0 in the
controls.

Break table, each mutation on the merged tree:

| mutation | fires |
|---|---|
| control, unmodified | 0 |
| declaration not plumbed into the check (behavioural pre-fix) | 5, all regression cases |
| a block is dead if **any** of its columns is dead | 1 - the parked-gripper false-positive guard |
| drop the single-block guard | 2 - the solo-robot contract and the underscored-joint-name case |
| drop the whole-vector early return | 1 - the "one problem, not one per block" case |
| stop checking `names` describes this vector | 1 - the wrong-width fallback |
| stop skipping single-frame episodes | 2, including the pre-existing `test_single_frame_all_zero_not_flagged` |
| split on `_` instead of `__` | 1 - the real-arm joint-name case |

Every mutation fires a distinct set; two rows fire only after the break table itself exposed that my
first fixtures (`"1".."6"`) could not distinguish them, which is why the realistic
`shoulder_pan`/`gripper` fixture exists.

## Gate

- `mypy` over the same targets: `Success: no issues found in 1519 source files`; branch-only error
  set empty against a pristine `main` worktree.
- 478-file selection (every test touching `verify_dataset`, the dataset recorder, recording, or a
  repo-wide docstring/changelog guard), identical on both trees with the changed test file excluded
  from each: **4 failed, 19268 passed, 35 skipped on both**, the 4 failing ids identical and both
  `comm` directions empty. They fail on pristine `main` alone too: 2 need `docker` on `PATH`
  (absent here, present in CI) and 2 assert an `ImportError` that `rclpy` being installed locally
  prevents.
- Scoped re-run after merging `main` (the three PRs that landed during the work): the changed file
  plus each of their own new suites plus the recording surface, **+16 passed against the same tree
  without this change, which is exactly the 16 tests added**, and an identical 4-failure set that
  is a module-caching artifact of the large selection (that file passes 13/13 alone on both trees).
- Changelog fragment validates; `docs/recording.md` gains the per-robot semantics and its list of
  failure modes, which said "four" and omitted the dead-column check entirely, now says five.

### DIFF
--- a/changelog.d/2702-dead-control-column-per-robot-block.md
+++ b/changelog.d/2702-dead-control-column-per-robot-block.md
@@ -1,0 +1,41 @@
+### Fixed: `verify-dataset` grades a dead control column per robot, not per whole vector
+
+`verify_dataset`'s dead-control-column check exists to catch a recording whose `action` or
+`observation.state` column was written as zeros because the writer's keys never resolved to the
+declared columns - correct episode counts, correct pixels, no control signal. It fired only when
+the ENTIRE vector was identically zero, and a multi-robot recording cannot present that shape.
+`start_recording` declares one column per robot per joint and prefixes each with its robot's
+instance name (`alice__shoulder_pan`), so a resolution that succeeds for one robot and fails for
+another leaves that robot's whole block of columns zero while the other robot's columns carry
+real measurements. The vector as a whole varies, so the check saw nothing and the report read
+`[PASS]`.
+
+The recorder documents this mechanism itself, in the comment that remaps the prefixed keys in
+`strands_robots/simulation/mujoco/simulation.py`: without the remap, `add_frame` "looks up the
+prefixed schema keys, finds nothing, and writes all-zero state/action vectors silently". That is
+the per-robot failure, and the gate written to catch it could only see the all-robot case.
+
+The vector is now split into the blocks `meta/info.json` declares - column indices grouped by the
+per-robot prefix - and a block that is wholly zero is reported the same way a wholly-zero vector
+is, naming the robot: `feature 'action' is identically zero for every 'bob' column across episode
+0 (60 frame(s))`. A zero SUBSET of one robot's block is deliberately left alone, because a
+gripper parked at zero for a whole episode is a measurement rather than a writer fault. The
+separator is the doubled underscore `start_recording` writes, not a single one: a real arm's
+joints are `shoulder_pan` / `gripper`, and splitting on `_` would turn one arm into four blocks
+and flag its parked gripper.
+
+The change is a strict extension. A wholly-zero vector reports the message it reported before,
+once, rather than one per block; a dataset whose `names` are absent, non-string or of a different
+width than the stats it is being asked about keeps the whole-vector rule, which is every
+single-robot recording; and `stats_vectors_checked` is unchanged, so the same work yields a finer
+verdict. Over the LeRobot datasets on one machine - 1314 roots, graded on both sides - 58 moved
+from pass to fail, no dataset moved from fail to pass, every problem string reported before was
+still reported, and every added string was the new per-block finding.
+
+This complements rather than duplicates the recorder fix in #2699. That change makes an undriven
+robot's `observation.state` columns measurements from now on, and states that the matching action
+columns are deliberately unchanged pending a schema decision. Neither can repair a dataset already
+on disk, and a recording made after it still carries an undriven robot's `action` block as zeros -
+which is exactly what this check now reports, without deciding the schema question: on a two-arm
+recording made on current `main`, the `action` block is named and the `observation.state` block is
+not, because that one now varies.
--- a/docs/recording.md
+++ b/docs/recording.md
@@ -348,20 +348,45 @@ strands-robots verify-dataset /path/to/dataset --no-check-videos  # skip the per
 ```
 
 `verify-dataset` reuses the same pure-pyarrow `read_dataset_episode_indices`
-helper (no `lerobot` import) and flags four failure modes: the mega-episode
+helper (no `lerobot` import) and flags five failure modes: the mega-episode
 (fewer distinct episodes than `--expected`), `meta/info.json` `total_episodes` /
 `total_frames` drifting from the parquet ground truth (caught even without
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Extends `verify-dataset`'s dead-control-column check from whole-vector granularity to per-robot blocks: the vector is split by the doubled-underscore prefix `start_recording` writes into each declared column name (`alice__shoulder_pan`), and a block that is identically zero across a multi-frame episode is flagged even while a sibling robot's columns vary. The change is a strict extension of the existing check - a wholly-zero vector still reports the single whole-vector finding, absent/non-string/wrong-width `names` keep the whole-vector rule (every single-robot recording), a zero subset of one robot's block (a parked gripper) is deliberately left alone, and `stats_vectors_checked` semantics are unchanged. I verified the separator claim against the recorder (`simulation/mujoco/recording.py` applies the `{rname}__` prefix only in multi-robot scenes, and base-state columns carry the same prefix, so blocks group consistently per robot), ran the full `tests/test_verify_dataset.py` suite at the head SHA (73 passed), and confirmed ruff check/format are clean on both changed Python files.

## What's good

- The new `TestDeadControlColumnPerRobotBlock` class covers both directions: the finding (one dead block among live siblings, per feature, naming only the dead robot) and every fallback (absent names, wrong-width names, single-frame episode, wholly-dead vector reporting once, solo-robot underscored joint names not splitting, `--no-check-stats` opt-out, CLI exit code).
- `_declared_column_blocks` requires >= 2 distinct blocks before splitting, so single-robot datasets and real arm joint names with single underscores (`shoulder_pan`) cannot be misgraded - and the tests pin that boundary explicitly with the SO-10x joint list.
- Changelog fragment present per AGENTS.md step 3; docs/recording.md updated in the same PR; new problem strings are plain ASCII; the new `declared_features` parameter is keyword-only on a private helper, leaving `verify_dataset`'s public signature untouched.

PR #2703, opened 2026-08-25
fix(sim): refuse a scene whose cameras share one dataset column name

cagataycali:
## Problem

A LeRobot feature name cannot contain `/` -- it addresses nested features -- so a camera's namespace separator is recorded as `__`, and `arm/wrist` becomes `observation.images.arm__wrist`. That collapse is not injective: `arm/wrist` and `arm__wrist` are **two cameras in the scene and one column in the dataset**. Nothing downstream could say which of them the column was declared for, and the three ways of asking each failed differently.

Measured on a one-joint arm whose MJCF declares a `wrist` camera (so the compiled scene has `arm/wrist`) plus `add_camera(name="arm__wrist", width=128, height=96)`. All three are `main`:

| call | status | declared columns | what actually happened |
|---|---|---|---|
| `start_recording(...)` (every camera) | `error` | -- | refused by `DatasetRecorder.create` as `camera_keys must not repeat a name, got ['default', 'arm__wrist', 'arm__wrist']` -- a true refusal, but it reads as the caller having named one camera twice |
| `cameras=["arm/wrist", "arm__wrist"]` | `success` | `['arm__wrist']` | one camera silently dropped; the column declared `(3, 96, 128)` (the DROPPED camera's size) while recording the kept one, so the FIRST `add_frame` was rejected: `RecordingFrameError: ... after 0 frame(s) written`, `run_policy` -> `error`, `stop_recording` -> `error`, 0 frames, no MP4 -- after `overwrite=True` had already wiped the dataset being replaced |
| `cameras=["arm__wrist"]` | `success` | `['arm__wrist']` | recorded, but *which* camera lands under the key is decided by the spelling used |

The requested names in row 2 are distinct, so they pass `name_list_error` and only their *keys* collide -- which is the same consequence that function's own docstring gives as the reason it refuses a literal repeat: "two `camera_keys` entries naming one camera declare a single camera column".

`docs/recording.md` already promised the opposite: *"Every camera key carries the view of the camera it names ... so a column is never quietly populated from the wrong camera."* That guarantee needs the scene's cameras to have distinct column names.

## Change

The ambiguity belongs to the scene, not to one way of recording it, so it is refused once on a shared domain, `camera_schema_key_collision_error`, that every backend's `start_recording` consults. The message names both cameras, the key they share, the consequence and the rename that resolves it:

```
start_recording: these scene cameras do not have distinct dataset feature names:
'arm__wrist' <- ['arm/wrist', 'arm__wrist']. A LeRobot feature name cannot contain '/'
(it addresses nested features), so a camera's namespace separator is recorded as '__' -
which makes these names one column. Whichever camera won it, the column would be named
after the other one, and if they render at different sizes the first frame is rejected
and the episode is lost. Rename one of them (add_camera(name=...)) so the names still
differ once '/' becomes '__', then record again.
```

Placement is pinned at both ends. It runs **after** the dataset-stack probe, because reading the scene's cameras is an engine call and that block is reachable on an install with no engine at all (`tests/simulation/test_recording_dataset_stack_unavailable_across_backends.py` drives it over a hand-built world; putting the guard ahead of it turned "the lerobot extra is missing" into an engine failure -- that suite caught exactly this and 11 of its cases are why the guard moved). It runs **before** any session state and before `_prepare_dataset_target`, so a refusal leaves nothing set and nothing on disk: with a pre-existing `meta/info.json` and `overwrite=True`, the file is unchanged and the root is not created.

Scoping with `cameras=` is deliberately not treated as a workaround. Whichever of the pair won the column, the column would be named after the other one, so narrowing the request narrows the ambiguity rather than removing it.

The collapse itself now has a single owner, `strands_robots.utils.camera_schema_key`, called by the three backends that declare a schema and by `DatasetRecorder.add_frame`, which renames an observed frame onto its declared column. Those four have to agree on one mapping for a frame to reach the column declared for it, and the guard has to agree with them or it grades a different question. It is idempotent, which is what lets a caller keep naming a camera in either form. Other domains that map `/` to `__` for their own reasons (an HF repo id becoming a local checkpoint directory) are untouched -- the derived rule is scoped to receivers whose name says they hold a camera.

## The ambiguity, in the dataset's own frames

Each panel is **decoded frame 0 of the dataset's own per-camera MP4** after the same 12-step mock rollout in MuJoCo (headless, `MUJOCO_GL=egl`), with identical scene lighting in all four (asserted, not claimed -- an earlier revision set it before `add_camera` and the recompile discarded it, which the md5s caught).

one column name, two cameras

Panels 1 and 2 are both `status=success` and both declare exactly `observation.images.arm__wrist`, and they carry different views (md5 `2624bd515ae3`, 640x480 vs `551d662ac8d9`, 320x240). Panel 4 records that same second camera under a name nothing else claims, and its frame is **byte-identical to panel 2's** (`551d662ac8d9`, across the two trees) -- so the fix refuses the ambiguity without losing the view.

The control's two MP4s, straight out of the dataset:
[`observation.images.arm__wrist`](https://github.com/cagataycali/robots/releases/download/artifacts/control-arm__wrist.mp4) - [`observation.images.overview`](https://github.com/cagataycali/robots/releases/download/artifacts/control-overview.mp4)

## Tests

`tests/simulation/test_camera_schema_key_collision.py`, 23 tests. Every expected column and size is read off the dataset rather than restated.

Pre-fix split, with the shared domain kept and only the guard call sites and the single-sourcing reverted (a raw revert of every source file is a collection error, since the symbols do not exist): **9 failed / 14 passed**, all 9 in the three regression classes (5 in `TestEveryWayOfRecordingACollidingSceneIsRefused`, 3 in `TestEveryBackendConsultsTheGuard`, 1 in `TestTheCollapseHasOneOwner`), **0 in `TestThePremise`, `TestADistinctSceneStillRecords`, `TestNoPartialStateEitherWay` or `TestTheRefusalMessage`**.

`TestNoPartialStateEitherWay` holds before and after -- the pre-fix refusal came from `DatasetRecorder.create` and already unwound the session -- so it is filed as the invariant the earlier guard must not break, not as evidence for it.

The control includes a full round trip: record -> `run_policy` -> `stop_recording` -> reopen, asserting both columns come back with a non-empty MP4 each and with **their own** resolutions (`(3, 480, 640)` and `(3, 64, 64)`) -- the half a shared column lost when it took one camera's size and the other's frames.

Break table (control unmutated: 0 failed):

| mutation | failures |
|---|---|
| guard dropped from all three backends | 8 |
| guard dropped from mujoco only | 8 |
| guard dropped from isaac + newton only | 3 (all structural) |
| domain groups by the raw name, not the key | 7 |
| only the first colliding group reported | 1 |
| message names the key but not the cameras | 4 |
| blank names not skipped | 1 |
| over-reach: refuse any name containing `__` | 4 (all controls) |
| single-sourcing reverted, guard kept | 1 |

Eight distinct firing sets. "Dropped from all three" and "dropped from mujoco only" coincide, because mujoco is the only behaviourally reachable backend here (Isaac Sim and Newton are not installed) and the structural rules fire for any missing backend -- so Isaac and Newton are graded structurally and said to be. Their `8` and the single-sourcing row's `1` sum to the behavioural pre-fix `9`, so each half is independently pinned. The over-reach row fires only controls, which is what makes the boundary defensible. The backend inventory is derived from the packages that ship a `start_recording`, so a fourth backend is graded on arrival; a non-vacuity test refuses an inventory of fewer than two.

## Gate

- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ` clean (1529 files).
- mypy vs a pristine `upstream/main` worktree: **5 == 5**, message sets byte-identical, 0 in changed files.
- Paired broad run at branch `e1a48ef4` against base `f7ebadf7`, 651 files (every test mentioning recording / camera / dataset, plus every tree-walking guard), identical selection on both trees with the new file excluded from each: **26 failed, 25402 passed, 204 skipped, 24 errors on BOTH** (1051s vs 1072s), **50 failing ids each, both diffs empty**. 41 of the 50 are `awsiot` (the `[mesh-iot]` extra) absent locally; the rest are pre-existing on both.
- The new file: **23 passed on mujoco 3.5.0 with lerobot 0.6.2 from source, and 23 passed on mujoco 3.12.0**.
- On the merged tree (main moved during review prep; zero file overlap -- mesh vs sim/recording): the seven recording/dataset suites across all three backends plus `verify_dataset`, **303 passed**.


yinsong1986 (dismissed):
## Summary

This PR gives the camera `/` -> `__` name collapse a single owner (`strands_robots.utils.camera_schema_key`), routes all four pre-existing respellings through it (mujoco / newton / isaac schema loops and `DatasetRecorder.add_frame`'s frame renaming), and adds a shared `camera_schema_key_collision_error` guard that every backend's `start_recording` consults: a scene whose cameras collapse onto one dataset column (`arm0/wrist` + `arm0__wrist`) is refused up front, naming both cameras and the shared key. The guard's placement is right in all three backends - after the no-world check and the dataset-stack probe (so a missing lerobot extra is still diagnosed without an engine), and before `_backend_state["recording"] = True` and `_prepare_dataset_target` (so a refusal cannot land after `overwrite=True` has wiped the dataset being replaced). Each backend's guard reads the same camera source its own schema loop declares from, unnamed cameras are handled (`mj_id2name` returning `None` hits the blank-name skip), `add_frame`'s mapping is byte-identical to before, and the refusal text is ASCII.

## What's good

- The guard-placement invariants (refuses before wipe, no partial session state, runs after the dataset-stack probe) are each pinned by tests that grade the backend inventory derived from the tree, not a list written today.
- Single ownership of the collapse is pinned by an AST scan scoped to camera receivers, with a self-test that the scan can see a respelling - and it correctly leaves the gr00t `hf_repo` checkpoint-dir mapping (a different domain) alone.
- Distinct-scene controls record end to end, including a namespaced robot camera on its own, so the guard only refuses when the collapse actually loses information.

## Verification suggestions

The required `call-test-lint` run is currently red on one mypy `arg-type` in the new test file (`Path(inspect.getsourcefile(simulation))` - `getsourcefile` returns `str | None`); that is lint-gated and will need a green run before merge, no review action beyond it.

yinsong1986 (approved):
## Summary

This PR refuses, once and up front, a scene whose cameras collapse onto one dataset column: the `/` -> `__` schema-key mapping is not injective, so `arm0/wrist` and `arm0__wrist` are two cameras and one column, and each of the three ways of recording such a scene previously failed differently (a far-from-cause `camera_keys` refusal, a silent drop of one camera, or a column whose contents depended on the spelling used) - all after `overwrite=True` had already wiped the dataset being replaced. The collapse gets one owner (`strands_robots.utils.camera_schema_key`), the four pre-existing respellings (three backend schema loops plus `DatasetRecorder.add_frame`) are routed through it byte-identically, and a shared `camera_schema_key_collision_error` guard is consulted from every backend's `start_recording`. I verified the guard's placement in all three backends: after the no-world check (so `self._world._model` / `self._world.cameras` are safe to read), after the dataset-stack probe (so a missing `lerobot` extra still reports as a dependency error), and before `_backend_state["recording"] = True` and `_prepare_dataset_target` (so a refusal wipes nothing and leaves no session state). Each backend's guard reads the same camera source its own schema loop declares from, and MuJoCo's unnamed cameras (`mj_id2name` returning `None`) hit the blank-name skip.

## What's good

- The guard is a shared domain with four call sites on day one, satisfying the second-caller rule for `utils.py`, and the refusal message is ASCII, names every colliding group, and carries an actionable remedy.
- The backend inventory and ordering invariants are derived from the tree (`*/recording.py` shipping a `start_recording`), with a non-empty-inventory self-test, so a backend added later is graded on arrival.
- The round-2 `_source_file` accessor resolves the prior review's mypy `arg-type` concern at all three call sites, refusing `None` where it appears instead of raising from inside `Path` or grading an empty inventory.
- Changelog fragment follows the `<pr-number>-<slug>.md` convention; docs state the naming constraint where a caller choosing camera names meets it.

## Verification suggestions

Ran the new suite against the PR tree on a GPU-less box: 22/23 pass; the single failure (`test_a_distinct_pair_survives_a_recording_round_trip`) is `mujoco.FatalError: an OpenGL platform library has not been loaded`, and a pre-existing untouched rendering test (`tests/simulation/test_dataset_recording_fps_contract.py::TestStartRecordingHonorsUsableFps::test_episode_round_trips_at_the_requested_rate`) fails identically on the same tree - environmental, not a regression from this diff. A runner with working EGL should see the full suite green; `call-test-lint` was still in progress at review time and is the authoritative read.

PR #2704, opened 2026-08-25
test(lerobot_local): reset every step a processor module owns, not a subset

cagataycali:
## Problem

`tests/policies/lerobot_local/test_vla_jepa.py::test_vla_jepa_processor_steps_register_via_strands_path`
is the regression guard for "the generic `lerobot_local` path already carries a brand-new VLA". It
forces a fresh import of `lerobot.policies.vla_jepa.processor_vla_jepa` so its
`@ProcessorStepRegistry.register` decorators run again, then asserts
`_register_policy_processor_steps` put VLA-JEPA's bespoke postprocessor steps back.

To force that import it evicted the module from `sys.modules` and cleared the step names. It
cleared **three** names; the module registers **four**:

| line in `processor_vla_jepa.py` | registry name | cleared by the reset |
|---|---|---|
| 43 | `vla_jepa_image_prep` | no |
| 103 | `vla_jepa_clip_actions` | yes |
| 118 | `vla_jepa_pre_snap_gripper` | yes |
| 151 | `vla_jepa_binarize_gripper` | yes |

LeRobot refuses a duplicate: `ProcessorStepRegistry.register` raises `ValueError: Processor step
'<name>' is already registered`. `vla_jepa_image_prep` is the module's **first** decorator, so the
re-import died on its way in, the three later steps never registered, and the test failed with
`KeyError: Processor step 'vla_jepa_clip_actions' not found in registry` plus a list of 58
unrelated step names -- reporting the registration path for a state the reset had created.

The path being graded was working the whole time. Measured against lerobot 0.6.2, in a fresh
process, with a reset that covers all four names:

```
_register_policy_processor_steps("vla_jepa")
-> ['vla_jepa_binarize_gripper', 'vla_jepa_clip_actions',
    'vla_jepa_image_prep', 'vla_jepa_pre_snap_gripper']
```

## The guard had teeth in neither direction

Its own skip guard calls `list_policy_types()`, which imports every policy config module and so
imports the `vla_jepa` package, whose `__init__` imports the processor module. The survivor is
therefore always registered by the time the body runs -- so the test fails wherever `vla_jepa` is
registered, which is exactly where it is meant to have teeth, and skips wherever it is not. It
fails when run as the only selected node id (`1 failed`), so this is not an ordering effect.

`vla_jepa` ships in the released lerobot as well as on `main`: `git ls-tree` counts 8 files under
`src/lerobot/policies/vla_jepa` at both `v0.6.0` and `v0.6.1`, and `uv.lock` pins `lerobot 0.6.1`.

## It also left the process un-importable

With the module absent from `sys.modules` and one of its names still registered, every later
import of that module trips the same surviving name. A probe that simply imports it, run in the
same session:

| selection | probe result |
|---|---|
| probe alone | `1 passed` |
| `test_vla_jepa.py` then probe | `2 failed` -- probe raises `ValueError: Processor step 'vla_jepa_image_prep' is already registered` |
| `test_vla_jepa.py` (this branch) then probe | `8 passed` |

So any future test touching VLA-JEPA in the same session would have failed for a reason belonging
to this file. The probe is evidence, not shipped -- the equivalent guard ships as
`test_forced_reimport_leaves_the_registry_and_sys_modules_as_found`.

`test_sys_modules_removal_leaves_no_orphan` grades unrestored `sys.modules` removals, and it is
right not to have flagged this one: its rule covers a removal that orphans a reference some test
*patches*, and nothing patches this module. The harm here arrives by a different route from the
same unrestored removal -- the module's import side effect is not idempotent -- so the restoration
is carried in the file that does the removing.

## Change

`_forced_reimport(module_name)` is a context manager that resets and restores:

- The set of names to clear is **read off the module** (every registry entry whose step class has
  that `__module__`) rather than listed, so a step lerobot adds, renames or moves is covered with
  no edit here. A name the reset misses is a name the re-import cannot re-register, which is what
  makes a derived inventory load-bearing rather than tidy.
- Registry entries and the `sys.modules` entry are both restored on exit, and the **original** step
  classes are put back rather than the new objects a fresh import builds, so a reference another
  test already holds and the registry entry stay the same class.

The assertions still concern the three postprocessor steps a checkpoint's
`policy_postprocessor.json` names. A non-vacuity check requires those three to be among the names
the module owns, so an upstream rename fails naming both sides instead of passing on an empty
intersection.

Two tests are added: one pinning that the registry and `sys.modules` are left as found, one pinning
the surviving-name hazard as behaviour, so the reason the reset has to be total is carried by a test
rather than by a comment.

## Pre-fix / post-fix

| selection | before | after |
|---|---|---|
| `test_vla_jepa.py` | `1 failed, 4 passed` | `7 passed` |
| the single registration node id | `1 failed` | `1 passed` |

## Break table

Each row mutates this branch and reports which tests fire. 7 mutations, 5 distinct firing sets.

| mutation | fires |
|---|---|
| reset clears only the 3 named steps (the shipped bug) | 3 -- registration, as-found, hazard |
| no restoration at all | 1 -- as-found |
| restore whatever the re-import left, not the originals | 1 -- as-found |
| clear the names but do not evict the module | 2 -- registration, hazard |
| hazard test forgets to leave a survivor registered | 1 -- hazard |
| `_steps_owned_by` does not import before reading | 0 |
| drop the non-vacuity rename guard | 0 |

Both zeros are explained rather than left as gaps:

- `_steps_owned_by`'s explicit import is defensive. Measured: `import lerobot.processor` does not
  cache the processor module, but `import lerobot.policies` does, and the module under test reaches
  the latter transitively -- so the module is always already imported here. The call is kept so the
  helper does not depend on an import happening elsewhere.
- The rename guard cannot fire while the named steps really are a subset of the owned ones. It is
  there for the upstream rename that would otherwise make the assertions grade an empty
  intersection, which is the failure mode this PR is fixing one level down.

The first row is the load-bearing one: the exact defect being fixed is now caught three ways.

## Gate

- `ruff check strands_robots tests tests_integ`: `All checks passed!`;
  `ruff format --check`: `1525 files already formatted`.
- `mypy strands_robots tests tests_integ`, branch vs a pristine `main` worktree, fresh cache dirs
  on both: **0 errors each**, message sets identical (`comm` empty both ways).
- Paired blast radius, 479 files (all of `tests/policies/lerobot_local`, plus every test that walks
  the tree with `ast`/`rglob`/`inspect.getsource`/`iter_modules`, plus every test mentioning
  `sys.modules`, `ProcessorStepRegistry`, `list_policy_types` or `_register_policy`), identical
  selection on both trees with the changed file excluded from each:
  **`6 failed, 20989 passed, 94 skipped` on BOTH**, failing-id sets identical, `comm` empty both
  ways. The 6 are host artifacts present on both trees: 2 needing `docker` on PATH, 4 arising from
  `rclpy`/DDS being importable locally.
- `changelog.d` fragment added; `scripts/check_changelog_fragment.py` clean, and the 304 changelog
  and tree-walking guard tests pass.

`git diff --name-only -- strands_robots/` against the merge base is **empty**: no runtime code
changes, which is also the strongest statement available that nothing else can regress.

## Live check

Because the subject is whether a policy's bespoke processor steps reach the registry, the same path
was exercised on a real checkpoint rather than only in the guard. Loading `lerobot/smolvla_base`
through `create_policy("lerobot_local", ...)` builds `ProcessorBridge(pre=6steps, post=2steps)`
with `smolvla_new_line_processor` registered -- the generic registration path landing a real
policy's bespoke step, end to end.

No visual artifact: the change touches no policy, simulation, rendering, recording or asset
behaviour. The evidence here is the registry state and the pass/fail tables above.

### DIFF
--- a/changelog.d/2704-processor-step-reset-covers-every-owned-step.md
+++ b/changelog.d/2704-processor-step-reset-covers-every-owned-step.md
@@ -1,0 +1,52 @@
+### Fixed: a forced processor-module re-import resets every step it owns, not a subset
+
+`tests/policies/lerobot_local/test_vla_jepa.py` is the regression guard for "the generic
+`lerobot_local` path already carries a brand-new VLA": it forces a fresh import of
+`lerobot.policies.vla_jepa.processor_vla_jepa` so the `@ProcessorStepRegistry.register`
+decorators run again, then asserts `_register_policy_processor_steps` put VLA-JEPA's bespoke
+postprocessor steps back. To force that import it evicted the module from `sys.modules` and
+cleared the step names -- but it cleared three names where the module registers four, leaving
+`vla_jepa_image_prep` behind.
+
+LeRobot refuses a duplicate: `ProcessorStepRegistry.register` raises `ValueError: Processor step
+'<name>' is already registered`. `vla_jepa_image_prep` is the module's FIRST decorator, so the
+re-import died on its way in, the three later steps never registered, and the test failed with
+`KeyError: Processor step 'vla_jepa_clip_actions' not found in registry` -- reporting the
+registration path for a state the reset had created. Measured on lerobot 0.6.2: given a reset that
+covers all four names, `_register_policy_processor_steps("vla_jepa")` registers all four, so the
+path being graded was working the whole time.
+
+The guard therefore had teeth in neither direction. Its own skip guard calls `list_policy_types()`,
+which imports every policy config module and so imports the `vla_jepa` package, whose `__init__`
+imports the processor module -- meaning the survivor is always registered by the time the body
+runs. So the test fails wherever `vla_jepa` is registered, which is exactly where it is meant to
+have teeth, and skips wherever it is not.
+
+It also left the process worse than it found it. With the module absent from `sys.modules` and one
+of its names still registered, every later import of that module trips the same surviving name, so
+the module is un-importable for the remainder of the run: a probe that imports it after this file
+executes raises `ValueError: Processor step 'vla_jepa_image_prep' is already registered`, and it
+passes when run without this file. Any future test touching VLA-JEPA in the same session would
+have failed for a reason belonging here.
+
+The reset is now read off the module rather than listed -- every registry entry whose step class
+has that `__module__` -- so a step lerobot adds, renames or moves is covered with no edit here,
+which matters because a name the reset misses is a name the re-import cannot re-register. Registry
+entries and the `sys.modules` entry are both restored on exit, and the original step classes are
+put back rather than the new objects the fresh import built, so a reference another test already
+holds and the registry entry stay the same class.
+
+The assertions still concern the three postprocessor steps a checkpoint's
+`policy_postprocessor.json` names; a non-vacuity check now requires those three to be among the
+names the module owns, so an upstream rename fails here naming both sides instead of passing on an
+empty intersection. Two tests are added: one pinning that the registry and `sys.modules` are left
+as they were found, one pinning the surviving-name hazard as behaviour, so the reason the reset has
+to be total is carried by a test rather than by a comment.
+
+`test_sys_modules_removal_leaves_no_orphan` grades unrestored `sys.modules` removals, and it is
+right not to have flagged this one: its rule covers a removal that orphans a reference some test
+*patches*, and nothing patches this module. The harm here comes from the same unrestored removal by
+a different route -- the module's import has a side effect that is not idempotent -- so the
+restoration is carried in the file that does the removing.
+
+No runtime code changes: `git diff --name-only -- strands_robots/` against the merge base is empty.
--- a/tests/policies/lerobot_local/test_vla_jepa.py
+++ b/tests/policies/lerobot_local/test_vla_jepa.py
@@ -25,6 +25,11 @@ lerobot versions and skip cleanly on a lerobot too old to ship ``vla_jepa``.
 
 from __future__ import annotations
 
+import contextlib
+import importlib
+import sys
+from collections.abc import Iterator
+
 import pytest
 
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Test-only fix to `tests/policies/lerobot_local/test_vla_jepa.py`: the forced re-import in the VLA-JEPA processor-registration guard previously cleared 3 of the 4 registry names the module owns, so the re-import died on the surviving `vla_jepa_image_prep` decorator, failed the test on a state the reset itself created, and left the module un-importable for the rest of the process. The new `_forced_reimport` context manager derives the reset set from the live registry (every entry whose step class has the module's `__module__`), restores both the original step classes and the `sys.modules` entry in a `finally` block, and two new tests pin the as-found restoration and the surviving-name hazard as behaviour.

## What's good

- `git diff --name-only -- strands_robots/` against the merge base is empty: zero runtime exposure.
- The restore path is exception-safe (`finally` re-registers originals by identity, and `saved_module` cannot be `None` because `_steps_owned_by` imports the module before it is read), so a mid-block assertion failure no longer poisons the session.
- The derived-inventory reset means an upstream step added/renamed in `processor_vla_jepa.py` is covered with no edit here, and the non-vacuity check fails naming both sides instead of grading an empty intersection.
- Changelog fragment per AGENTS.md step 3; test names describe behaviour, not provenance; no host paths or non-ASCII in either file.
- The `register(name=...)` / `unregister(name)` calls match lerobot's actual `ProcessorStepRegistry` signatures.

## Verification suggestions

- Run the file in isolation and then followed by any other lerobot-touching test module in one session (`pytest tests/policies/lerobot_local/test_vla_jepa.py tests/policies/lerobot_local/ -x`) to confirm the restoration holds under real ordering, not just the self-check.

PR #2705, opened 2026-08-25
fix(mesh): let a RosbridgeRobot command carry an operator decision

cagataycali:
## Problem

`use_rosbridge` gates a publish aimed at a safety-critical command surface, and `cmd_vel` is one.
`RosbridgeRobot.drive` and `.stop` took no `tool_context` parameter, and their two agent tools were
not declared `@tool(context=True)` -- so there was nothing for the gate to ask an operator with, and
every command this bridge sent to a blocklisted topic hit the gate's fail-closed branch.

Measured against the real transport (roslibpy client doubled, backend reachable), same `/cmd_vel`,
operator standing by to approve:

| call | prompted | status | published |
| --- | --- | --- | --- |
| `RtpsRobot.stop(tool_context=ctx)` | yes | `success` | 1 |
| `RosbridgeRobot.stop()` | no | `error` -- `No tool_context available for operator approval` | 0 |

It was the only drive-owning class in the package with that gap:

| class | `drive`/`stop` accept a context | why |
| --- | --- | --- |
| `RosBridgedRobot` | yes | inherits `MobileBaseRobot` |
| `RtpsRobot` | yes | inherits `MobileBaseRobot` |
| `AckermannRosRobot` | yes | owns `drive`/`stop`, threads it itself |
| `RosbridgeRobot` | **no** | owns `drive`/`stop`, did not |

The halt was the worst case. The one control the `tools` contract guarantees for anything that can
start motion -- "a caller that can start motion must be able to end it" -- was the one an operator
who was present and willing could not authorise. The only remedies the bridge could offer were the
blanket `BYPASS_TOOL_CONSENT` or a standing pre-approval, both of which give up the per-command
decision the gate exists for.

### The documentation said the opposite

`docs/rosbridge-integration.md` described the halt as `**stop() never gated** (this bridge only)`,
repeated it in two comments of a runnable example (`# always publishes zero Twist, never gated`,
`# single-shot, ungated`) and in a method-table row, and `RosbridgeRobot.stop`'s own docstring read
"Never gated on anything." A reader who believes that leaves `cmd_vel` out of
`STRANDS_ROS2_COMMAND_ALLOW` and discovers in the field that the halt is refused -- the
unreachable-halt hazard, arriving through documentation rather than through code.

`tests/test_use_ros_command_blocklist.py` exists to catch precisely that: it scans operator-facing
prose for a clause naming the halt and denying it is gated, then grades the claim against the
running gate. Its document list was three hardcoded files, keyed on naming the pre-approval
variable. A per-transport integration page shows an operator how to drive `cmd_vel` without ever
mentioning that variable, so the page carrying all three false claims sat outside everything the
scan read.

## Change

- `RosbridgeRobot._publish_twist`, `drive` and `stop` take and forward `tool_context`; the two
  commanding tools are `@tool(context=True)`. The trailing zero of a timed drive carries the same
  context as the command it undoes -- a zero that could not reach the gate would be refused on its
  own and leave the robot latched at the speed of a hold the operator did approve.
- All four prose surfaces now state what is true of this bridge and every other: the halt needs no
  prior state and no enable handshake, and it is *not* exempt from a gate keyed on the command
  surface rather than on the payload. `MobileBaseRobot.stop` already wrote that argument down --
  zero means "stationary" on a `Twist` but commands motion to the zero pose on a joint-command
  topic, so a payload-shaped carve-out could not be written correctly.
- The exemption scan's document set is derived: a page qualifies when it names a blocklisted
  surface by the same final-segment rule the gate matches on. 3 pages -> 9, pulling in every
  transport integration page. The narrow list is kept under a name describing what it does answer
  correctly (which pages document the variable itself), because its own non-vacuity check requires
  each entry to name it.
- New `tests/mesh/test_rosbridge_robot_command_gate.py`: the fourth of four per-bridge gate suites,
  alongside those for the ROS 2, RTPS and Ackermann bridges.

`tests/mesh/test_rosbridge_robot.py` could not have caught this: it patches the `use_rosbridge`
symbol at the mesh boundary, which is the boundary the gate lives behind. The new suite doubles the
roslibpy client *beneath* the real tool, so the gate and the transport wiring both run unmodified
while nothing reaches a rosbridge server.

## Tests

Pre-fix split, both source files reverted and the tests kept: **12 failed / 113 passed**.
Post-fix: **125 passed**. By class:

| class | pre-fix failures |
| --- | --- |
| `TestRosbridgeCommandsReachTheGate` (regression) | 8 |
| `TestEveryMobileBaseCanCarryAnOperatorDecision` (regression) | 3 |
| `TestTheDocumentedExemptionsAreTheRealOnes` (the docs guard) | 1 |
| `TestOnlyCommandsToABlockedSurfaceAreGated` (controls) | 0 |

The derived survey's own non-vacuity test passes on both trees, so it is not counted as regression
evidence.

### Break table

Each mutation applied to the fixed tree; firing set over the two test files.

| mutation | fires | which |
| --- | --- | --- |
| `stop` drops the forward | 2 | both stop cases |
| `drive` main publish drops the forward | 3 | both drive cases + the timed drive |
| trailing zero drops the forward | 1 | the timed drive **alone** |
| `_publish_twist` drops it to the transport | 5 | every publish path |
| stop tool not `context=True` | 1 | the structural check |
| drive tool not `context=True` | 1 | the structural check |
| **scan reverted to the hardcoded list AND prose reverted** | **0** | -- |
| derived doc set narrowed to `README.md` | 1 | the scan's non-vacuity check |
| both source files reverted (the reported defect) | 12 | matches the pre-fix split exactly |

Eight distinct firing sets over nine rows. The two `context=True` rows coincide, and the reason is
structural: one test grades every commanding tool of every bridge, so it cannot distinguish which
tool lost the declaration.

The **0** row is the point. With the scan pointed at its old three-file list and the prose left as
it was, the guard passes -- the blindness is the silence, not a failure anyone could have read. The
trailing-zero row firing exactly one test is what shows that path is pinned independently of the
command it undoes.

### Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1529 files).
- mypy **24 == 24** against a pristine `upstream/main` worktree; normalised message sets
  byte-identical both ways, 0 in changed files.
- Paired run, identical 553-file selection (all of `tests/mesh`, all of `tests/tools`, and every
  test that walks the tree with `ast`/`rglob`/`inspect.getsource` or reads `docs/`), with the two
  changed test files excluded from **both** trees: `66 failed, 21909 passed, 100 skipped,
  150 errors` on the branch and `66 failed, 21907 passed, 100 skipped, 150 errors` on the base --
  **failing-ID and error-ID sets identical, all four `comm` directions empty**. Every error is a
  declared extra absent from this environment (`device_connect_edge` 326, `awsiot` 88).
- The +2 passed is fully localised to `tests/test_args_docstring_completeness.py` (432 vs 430),
  which parametrises over documented functions and picks up the two new `Args` entries.
- `tests/mesh/test_zenoh_transport_security.py` is excluded from the pair and run alone on each
  tree (`9 passed, 2 skipped` on both), because a parallel Zenoh session manufactures cross-talk
  failures.

No visual artifact: this changes transport gating and prose, not policy, simulation, rendering,
recording or an asset. The measured tables above are the artifact.

### DIFF
--- a/changelog.d/2705-rosbridge-command-operator-context.md
+++ b/changelog.d/2705-rosbridge-command-operator-context.md
@@ -1,0 +1,55 @@
+### Fixed: a `RosbridgeRobot` command can carry an operator's decision
+
+`use_rosbridge` gates a publish aimed at a safety-critical command surface, and `cmd_vel` is one.
+`RosbridgeRobot.drive` and `.stop` took no `tool_context` parameter and their two agent tools were
+not declared `@tool(context=True)`, so there was nothing for the gate to ask an operator with:
+every command this bridge sent to a blocklisted topic hit the gate's fail-closed branch. Measured
+against the real transport on the same `/cmd_vel`, with an operator standing by to approve:
+
+    RtpsRobot.stop(tool_context=ctx)  ->  success, 1 message published
+    RosbridgeRobot.stop()            ->  error, "No tool_context available for operator
+                                         approval", 0 published
+
+It was the only drive-owning class in the package with that gap - `RosBridgedRobot` and
+`RtpsRobot` inherit the threading from the shared mobile base, and `AckermannRosRobot`, which
+also owns its own `drive`/`stop`, threads it itself. The halt was the worst case: the one control
+the `tools` contract guarantees for anything that can start motion ("a caller that can start
+motion must be able to end it") was the one an operator who was present and willing could not
+authorise. The only remedies the bridge could offer were the blanket `BYPASS_TOOL_CONSENT` or a
+standing pre-approval, both of which give up the per-command decision the gate exists for.
+`drive`, `stop` and the trailing zero of a timed drive now forward the context, the trailing zero
+carrying the same one as the command it undoes - a zero that could not reach the gate would be
+refused on its own and leave the robot latched at the speed of a hold the operator did approve.
+
+The documentation said the opposite. `docs/rosbridge-integration.md` described the halt as
+`**stop() never gated** (this bridge only)`, repeated it in two comments of a runnable example
+(`# always publishes zero Twist, never gated`, `# single-shot, ungated`) and in a method table
+row, and `RosbridgeRobot.stop`'s own docstring read "Never gated on anything." A reader who
+believes that leaves `cmd_vel` out of `STRANDS_ROS2_COMMAND_ALLOW` and discovers in the field
+that the halt is refused - the unreachable-halt hazard, arriving through documentation. All four
+surfaces now state what is actually true of this bridge and every other: the halt needs no prior
+state and no enable handshake, and it is not exempt from a gate that is keyed on the command
+surface rather than on the payload. The shared mobile base already wrote that argument down; zero
+means "stationary" on a `Twist` but commands motion to the zero pose on a joint-command topic, so
+a payload-shaped carve-out could not be written correctly.
+
+`tests/test_use_ros_command_blocklist.py` exists to catch exactly this - it scans operator-facing
+prose for a clause that names the halt and denies it is gated, then grades the claim against the
+running gate. Its document list was three hardcoded files keyed on naming the pre-approval
+variable, and a per-transport integration page shows an operator how to drive `cmd_vel` without
+ever mentioning that variable, so the page carrying all three false claims sat outside everything
+the scan read. With the scan pointed at its old list and the prose left as it was, the guard
+passes: the blindness is the silence. The scanned set is now derived - a page qualifies when it
+names a blocklisted surface by the same final-segment rule the gate matches on - which takes it
+from 3 pages to 9 and pulls in every transport integration page. The narrow list is kept for the
+separate question it answers correctly, which pages document the variable itself, since its own
+non-vacuity check requires each entry to name it. This is the second hardcoded universe in this
+area to be derived after #2700 did the same for the drive-contract scope survey; the shape is a
+guard whose discovery signature is narrower than the thing it guards.
+
+A new `tests/mesh/test_rosbridge_robot_command_gate.py` is the fourth of four per-bridge gate
+suites, doubling the roslibpy client beneath the real `use_rosbridge` so the gate and the
+transport wiring both run unmodified. `tests/mesh/test_rosbridge_robot.py` could not have seen
+the defect: it patches the `use_rosbridge` symbol at the mesh boundary, which is the boundary the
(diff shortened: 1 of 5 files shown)


yinsong1986 (approved):
## Summary

Threads `tool_context` through `RosbridgeRobot._publish_twist` / `drive` / `stop` and declares the two commanding agent tools `@tool(context=True)`, so a publish to a blocklisted `cmd_vel` prompts the operator instead of unconditionally hitting the gate's fail-closed branch - closing the gap where `RosbridgeRobot` was the only drive-owning class whose halt an operator could not authorise. The trailing zero of a timed drive carries the same context as the hold it undoes, which keeps the self-stop reachable through the gate. Four prose surfaces claiming `stop()` is "never gated" are corrected, and the exemption scan in `tests/test_use_ros_command_blocklist.py` derives its document set from the blocklist's own final-segment rule (3 pages -> 9) so a transport integration page can no longer make an ungated-halt claim outside the scan's reach. The change is additive - `tool_context` defaults to `None` at the end of each signature, and the no-context path still fails closed identically (pinned by `test_without_an_operator_the_refusal_names_both_ways_through`) - so no existing caller or config is silently invalidated.

## What's good

- The new suite doubles the roslibpy client *beneath* the real `use_rosbridge`, so the gate and transport wiring both run unmodified - exactly the boundary `tests/mesh/test_rosbridge_robot.py` could not see past.
- `_install_doubles` clears `BYPASS_TOOL_CONSENT` / `STRANDS_ROS2_COMMAND_ALLOW` so an ambient pre-approval cannot vacuously pass the gate assertions, and uses `monkeypatch.setitem(sys.modules, ...)` per AGENTS.md's sys.modules rule.
- The fleet-wide structural check (`TestEveryMobileBaseCanCarryAnOperatorDecision`) derives its class set from the mesh package rather than a hardcoded list, with its own non-vacuity assertion - so a fifth mobile base is graded on arrival.
- Controls pin both directions: an unblocked surface never prompts, a pre-approved surface publishes without asking, and the read tools are asserted to take no context.
- Changelog fragment present; all changed files ASCII-clean; ruff and the docstring-completeness suites pass on the head.

Verified locally at `5e99b9b3`: both new/changed test files pass (125 passed), the four sibling gate suites and `tests/tools/test_use_rosbridge.py` pass unchanged, and ruff check/format are clean on the changed files.

PR #1936, opened 2026-08-04
fix(tools): lerobot_train refuses a run size it cannot honor, before launching it

cagataycali:
## Problem

`build_train_command` writes `--steps` and `--batch_size` into the argv verbatim. The tool then launches that argv as a **detached** process (`Popen(..., start_new_session=True)`), so an unusable value is never reported to the caller: the tool returns `status=success` with a pid, and only the training log records, minutes later, that the run could not proceed.

Two of those values are worse than late - they are silent:

| supplied | argv token | what lerobot does with it |
| --- | --- | --- |
| `steps=0` | `--steps=0` | parses as `int` 0, so `for _ in range(step, cfg.steps)` is **empty**: the run completes having trained nothing |
| `steps=-5` | `--steps=-5` | likewise empty |
| `steps=True` | `--steps=True` | parses as `1`: a silent run of one step |
| `batch_size=True` | `--batch_size=True` | parses as `1`: a silent batch of one |
| `batch_size=0` / `-8` | `--batch_size=0` | `torch.utils.data.DataLoader` raises `batch_size should be a positive integer value` inside the detached process |
| `steps=2.7` / `nan` / `inf` | `--steps=2.7` | `DecodingError: Couldn't parse '2.7' into an int`, in the detached process |

lerobot declares both as plain `int` fields and `TrainPipelineConfig.validate()` checks neither, so nothing downstream catches the first four rows at all.

### One parameter, two contracts

`LerobotTrainer.validate(spec)` - the other surface that sizes the same lerobot run - **already** refused a non-positive `steps`:

```
LerobotTrainer.validate(spec)   steps=0  ->  refused    "steps must be > 0, got 0"
build_train_command(...)        steps=0  ->  accepted   "--steps=0"
```

So the accepted domain depended on which surface a caller reached for.

And within `build_train_command` itself, `num_gpus < 1`, a non-positive `val_episodes`, and two mutually-exclusive tuning strategies were all already refused - leaving the batch size as the one knob sizing the run with no domain at all.

## Change

Both knobs are checked against the shared `positive_count_error` domain (strict `int`, rejects `bool`, `>= 1`) before either reaches the argv, beside the existing `num_gpus` guard. That domain already documents itself as the rule for "the knobs that count iterations of a loop", which is exactly what `steps` bounds.

No new helper and no new message family: the refusal reads `lerobot_train: steps must be a positive integer, got 0.`

**Preserved deliberately** (both pinned):

[...]

PR #2706, opened 2026-08-25
fix(tools): hold lerobot_train's save_freq to an int lerobot can decode

cagataycali:
## Problem

`build_train_command` writes `save_freq` into `--save_freq=` - and, when a
validation split is requested, into `--eval_steps=` as well - in the argv of a
process the tool then launches **detached**. Every other knob in that argv was
already checked up front for exactly that reason: `steps` and `batch_size`
against the shared positive-count domain (#1936), `num_gpus`, `val_episodes`,
`lora_r` / `lora_alpha`, and `device` against torch's own device grammar
(#2560). `save_freq` was the last one carried through unchecked, and it failed
two ways.

lerobot declares the field as a plain `int`, so its parser decodes the token
with `int(...)`. Measured on `main`, with `save_freq` supplied and no
`val_episodes`:

| save_freq | argv `main` built | what lerobot's parser does with that token |
|---|---|---|
| `5000` | `--save_freq=5000` | decodes to `5000` |
| `0` / `-5` | `--save_freq=0` / `-5` | decodes; disables periodic saving (documented) |
| `True` | `--save_freq=True` | `DecodingError: Couldn't parse 'True' into an int` |
| `False` | `--save_freq=False` | `DecodingError: Couldn't parse 'False' into an int` |
| `2.7` | `--save_freq=2.7` | `DecodingError: Couldn't parse '2.7' into an int` |
| `5000.0` | `--save_freq=5000.0` | `DecodingError: Couldn't parse '5000.0' into an int` |
| `nan` | `--save_freq=nan` | `DecodingError: Couldn't parse 'nan' into an int` |
| `inf` | `--save_freq=inf` | `DecodingError: Couldn't parse 'inf' into an int` |

So the tool returned a pid and a log path, and only the training log recorded -
minutes later - that the run could not start. `lerobot_train(action="start",
save_freq=2.7)` reported `status: success` on `main`.

The second path is worse than late. `--eval_steps` is selected with
`save_freq if save_freq > 0 else steps`, so a value that is not orderable
against `0` raised out of the builder itself:

```
save_freq="5000", val_episodes=2  ->  TypeError: '>' not supported between instances of 'str' and 'int'
```

which the tool's outermost handler turned into
`Tool execution failed: '>' not supported between instances of 'str' and 'int'` -
a message naming neither the parameter nor the tool. `[5000]` and `object()`
behaved the same way.

One bad value also spoils two flags: with `val_episodes` set, `main` emitted
`--save_freq=2.7 --eval_steps=2.7` and `--save_freq=True --eval_steps=True`.

## Change

A module-local `_save_freq_error`, beside `_torch_device_error`, admits a true
`int` and nothing else. That is the sibling's policy, not a new one:
`positive_count_error` - which `steps` and `batch_size` in the same argv already
use - refuses `5000.0` and `np.int64(5000)` for the same decoding reason.
`bool` is refused explicitly, being an `int` subclass whose `True` renders as the
undecodable token `True`. `None` still means "omit the flag and keep lerobot's
default".

Like `device`, the check sits **after** the resume early return, so a resumed
run - whose argv carries neither `--save_freq` nor `--eval_steps` - is not
refused for a value it does not use.

After, same call sites:

| save_freq | result |
|---|---|
| `5000` | `--save_freq=5000` (unchanged) |
| `0` / `-5` | `--save_freq=0` / `-5`, and `--eval_steps=<steps>` (unchanged) |
| `True`, `False`, `2.7`, `5000.0`, `nan`, `inf`, `'5000'`, `[5000]`, `object()` | `ValueError: lerobot_train: save_freq must be an integer number of steps, got <value>. lerobot decodes the flag into an int field, ...` |

The tool now answers `status: error` with that text, and launches no process.

## Scope: the floor stays out

lerobot documents a non-positive `save_freq` as "disables periodic saving" and
`should_save_checkpoint` implements it (`save_freq > 0 and step % save_freq == 0`),
so `0` and a negative are a capability rather than an unusable value - and the
`--eval_steps` fallback beside this guard is written for exactly that case. Only
the type is graded. #1936 scoped `save_freq` out entirely on that reasoning; the
reasoning covers the floor, so its module docstring and
`TestTheScopeIsNoWiderThanTheDefect` are corrected here to say the floor is out
of scope rather than the parameter, and its two existing
`test_a_non_positive_save_freq_is_still_accepted` cases are left as the pins they
already were.

## Tests

`tests/tools/test_lerobot_train_checkpoint_cadence_domain.py`, 68 tests.
Pre-fix (source reverted, tests kept) on lerobot 0.6.2: **40 failed, 84 passed**.

| class | pre-fix failures |
|---|---|
| `TestACadenceLerobotCannotDecodeIsRefused` | 36 (4 tests x 9 values) |
| `TestTheRefusalReachesTheCallerBeforeAnyProcessStarts` | 3 |
| `TestAResumedRunIsNotRefusedForAFlagItDoesNotCarry` | 1 (its deliberate non-vacuity half) |
| `TestAUsableCadenceIsUntouched` | 0 |
| `TestTheFloorIsNotPartOfThisDomain` | 0 |
| `TestThePremisesTheDomainRestsOn` | 0 |
| the whole `test_lerobot_train_size_knob_domain.py` sibling | 0 (56 passed) |

The premises are derived rather than listed: each refused value is graded against
**both** consumers - can lerobot's parser read the token it renders to, and does
`value > 0` not raise - and the domain may only refuse a value that fails at
least one. A non-vacuity test asserts both failure modes are really represented,
because neither accounts for the set alone:

| value | parser reads the token | `> 0` does not raise |
|---|---|---|
| `True`, `False`, `2.7`, `5000.0`, `nan`, `inf` | no | yes |
| `'5000'` | yes | no |
| `[5000]`, `object()` | no | no |
| `0`, `-5` (accepted) | yes | yes |

`'5000'` is the reason the token argument alone is not enough: it renders to the
very token an `int` field decodes, and is unusable only because of the
comparison.

## Break table

Nine mutations of the shipped code, nine distinct firing sets, unmutated control
0 failures (run over the new file plus the sibling):

| mutation | tests fired |
|---|---|
| guard never called (the pre-fix source) | 40 |
| guard placed with the size guards, before the resume early return | 1 (the resumed-run case alone) |
| `bool` admitted (type test alone) | 9 |
| any real scalar admitted (integral floats too) | 18 |
| **over-reach**: the floor made part of the domain | 6, **two of them pre-existing** (`test_a_non_positive_save_freq_is_still_accepted[0]`, `[-1]`) |
| refusal drops the parameter name | 30 |
| refusal drops the offending value | 9 |
| the `None` sentinel graded too | 1 (the omit-the-flag case alone) |
| only the validation-split path guarded | 21 |

`9 + 18 = 27` and the two together are the whole type test, so each half is
independently pinned; the over-reach row is what makes the floor decision
load-bearing rather than prose.

## Gate

- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ`
  clean (1529 files).
  `main` worktree, normalized message sets byte-identical, 0 in the changed
  files.
- Paired run of an identical 146-file selection (all of `tests/tools/` plus
  every test that walks the tree with `ast` / `rglob` / `inspect.getsource`),
  with the two changed test files excluded from each side:
  **6449 passed, 63 skipped, 0 failed on both trees**, failing-id and error-id
  sets both empty in both directions.
- The new file and the sibling together: **124 passed** on lerobot 0.6.2.

No visual artifact: this changes an argv builder's input validation, with no
policy, simulation, rendering, recording or asset behaviour touched. The measured
tables above are the evidence.


yinsong1986 (approved):
## Summary

Closes the last unchecked token in `build_train_command`'s detached argv: `save_freq` now gets a type-only guard (`_save_freq_error`) at the dispatch boundary, matching the `int` requirement `positive_count_error` already imposes on the `steps`/`batch_size` knobs beside it, with the floor deliberately excluded because lerobot documents a non-positive cadence as "disables periodic saving". The guard sits after the resume early-return alongside `_torch_device_error`, so a resumed run whose argv carries neither `--save_freq` nor `--eval_steps` is never refused for a value it does not use, and a refused cadence returns a `ValueError` that the tool folds into an error envelope before any process is launched.

## What's good

- Guard placement and scope match the established pattern in this file (after the resume return, refuse-before-Popen, error-dict at the tool surface per AGENTS.md > Review Learnings (#86) > Error Handling Contracts).
- Kept local with one caller, per AGENTS.md convention 11 ("a value-domain guard becomes shared when it has a second caller") — the docstring even records the decision.
- The test suite pins both failure modes (undecodable argv token, `TypeError` from the `eval_steps` comparison) with executable premises against lerobot/draccus via `importorskip`, plus non-vacuity controls and probe-set self-checks; the floor-out-of-scope decision is pinned in both the new file and the updated size-knob tests so the two cannot drift apart.
- Changelog fragment present and correctly numbered; `Raises:` block and the tool's `save_freq` `Args:` entry updated in the same change; diff is ASCII-clean.
- Verified locally at `07e70a3d`: the new file (68 passed) and the neighbouring `test_lerobot_train*.py` suites (174 passed, 1 skipped) are green.

cagataycali:
Posted here in error - it belongs to #2707, where it now lives. Nothing in it concerns this merged change.

Cause, since it is the failure AGENTS.md > PR Workflow step 8 names: the `pullRequestId` was a node ID carried over from this PR's merge mutation earlier in the same run rather than resolved fresh from a query naming owner/name/number. A well-formed ID that is wrong does not fail - it succeeds against whatever object it does name. The `url` in the response is what caught it, which is the read-back that rule exists for.

PR #1939, opened 2026-08-04
fix(training): the run size a TrainSpec asks for is one shared positive-count domain

cagataycali:
## Problem

`TrainSpec.steps` and `TrainSpec.global_batch_size` are the two factors of how much training a spec asks for, and every supervised backend reads both straight into a discrete consumer: `steps` bounds the optimizer loop (lerobot iterates `range(step, cfg.steps)`) and `global_batch_size` becomes a `DataLoader` batch size or a `--global_batch_size` flag.

Four backends each carried their own copy of the same line:

```python
if spec.steps <= 0:
    problems.append(f"steps must be > 0, got {spec.steps}")
```

`mock.py:52`, `cosmos3.py:156`, `groot.py:155`, `lerobot.py:595`. The copies agreed with each other, so no test ever failed — and they shared one hole: a comparison admits every value that is not *comparably* non-positive. The second factor, `global_batch_size`, was checked by nothing at all.

Measured on `upstream/main`, one script over all four backends. Every verdict was identical across `mock` / `cosmos3` / `groot` / `lerobot_local`:

| field | value | `validate()` on main | what main then did with it |
|---|---|---|---|
| `steps` | `10000` | accepted | 10000 optimizer steps (correct) |
| `steps` | `0` / `-5` | refused | — |
| `steps` | `True` | **accepted** | `len(range(0, True)) == 1` — a silent one-step run reported as a completed job |
| `steps` | `2.7` / `nan` / `inf` | **accepted** | `TypeError: 'float' object cannot be interpreted as an integer` inside `range()`, after the dataset and the model are already loaded |
| `steps` | `'1000'` / `None` | **raised `TypeError`** | `'<=' not supported between instances of 'str' and 'int'`, out of `validate()` itself |
| `global_batch_size` | `32` | accepted | batches of 32 (correct) |
| `global_batch_size` | `0` / `-8` / `True` / `2.7` / `nan` / `'32'` | **accepted** | `ValueError: batch_size should be a positive integer value` from torch's `DataLoader`, once the dataset is materialized |

12 of 16 unusable values were let through. The last `steps` row is the sharpest: `Trainer.validate` is documented as

> Pure, side-effect-free preflight. Return a list of human-readable problems […] it powers a `plan` advisor that runs *before* anything expensive starts.

[...]

PR #2707, opened 2026-08-25
fix(training): hold TrainSpec.save_freq to a whole number of steps

cagataycali:
## Problem

`TrainSpec.save_freq` is the interval, in optimizer steps, at which a run writes a checkpoint. Four providers read it — the three supervised backends and the SageMaker transport — and it was the only numeric field among them with no domain, while `steps`, the step count interpolated into the *very same argv*, already shared one. Its `Attributes:` entry was a bare one-liner ("Checkpoint cadence in steps.") where every numeric neighbour spends three to six lines on its domain and names the `Trainer.validate` gate that enforces it.

Measured on `main`: `validate()` reported nothing about the field, on any backend, for any unusable value.

| `save_freq` | lerobot_local | groot | cosmos3 | sagemaker |
|---|---|---|---|---|
| `2.7` / `5000.0` / `True` / `False` / `nan` / `inf` / `"5000"` / `None` | no problem | no problem | no problem | no problem |

Each provider consumes the value in three shapes, and every way an unusable cadence failed was silent or late.

**A comparison selector.** LeRobot derives the validation cadence from it in two places — `--eval_steps=` on the argv path and `cfg.eval_steps` on the in-process one — as `spec.save_freq if spec.save_freq > 0 else spec.steps`, because a non-positive cadence disables periodic saving. Driving `LerobotTrainer.build_command` on `main` with `steps=100, val_episodes=2`:

| `save_freq` | `validate()` | tokens built |
|---|---|---|
| `nan` | 0 problems | `--save_freq=nan` `--eval_steps=100` |
| `2.7` | 0 problems | `--save_freq=2.7` `--eval_steps=2.7` |
| `True` | 0 problems | `--save_freq=True` `--eval_steps=True` |
| `inf` | 0 problems | `--save_freq=inf` `--eval_steps=inf` |
| `"5000"` | 0 problems | raises `TypeError: '>' not supported between instances of 'str' and 'int'` |
| `None` | 0 problems | raises `TypeError: '>' not supported between instances of 'NoneType' and 'int'` |

`nan` is the sharpest: it compares false against everything, so the selector takes the *disabled* branch. A spec asking to checkpoint every `nan` steps evaluates once at the very end instead, under a successful result. And a string or `None` raises out of the comparison itself — from inside a `Trainer.validate` that is documented to *return* problems.

**An argv or override token.** The other three interpolate the value verbatim, with no gate at all:

| `save_freq` | groot | cosmos3 |
|---|---|---|
| `nan` | `--save_steps=nan` | `checkpoint.save_iter=nan` |
| `2.7` | `--save_steps=2.7` | `checkpoint.save_iter=2.7` |
| `None` | `--save_steps=None` | `checkpoint.save_iter=None` |

LeRobot declares the field as a plain `int`, so its own draccus decoder is the token's real destination — and it refuses eight of the ten spellings: `draccus.decode(int, "2.7")` → `DecodingError: Couldn't parse '2.7' into an int`, likewise for `5000.0`, `True`, `False`, `nan`, `inf` and `None`. That refusal lands inside the launched run, once the dataset and model are loaded.

**An assignment into a typed config.** `cfg.save_freq` (declared `int`, default `20000`), GR00T's `save_steps` kwarg, and a forwarded SageMaker hyperparameter — none of which coerces.

## Change

`Trainer._checkpoint_cadence_problems` routes the field through `checkpoint_cadence_problems`, a fifteenth shared gate in `strands_robots.training._validate`, called by the four providers that read it. The message names the backend that refused the value:

```
lerobot_local: save_freq must be a whole number of steps, got nan. The cadence is
decoded into an int field and compared against zero, so a fractional, non-finite,
boolean or non-numeric value cannot be honored; pass a whole number of steps, or a
non-positive one to disable periodic saving.
```

The type test is strict for the same reason the run-size gate's is: `save_freq` and `steps` reach the same `int` decoder in the same argv, so the same number must not be refused for one step count and accepted for the next (measured: `positive_count_error` already refuses `5000.0` and `True` for `steps`).

### The floor is deliberately not part of the domain

LeRobot documents a non-positive `save_freq` as disabling periodic saving — `should_save_checkpoint` implements exactly that (`save_freq > 0 and step % save_freq == 0`), avoiding a `ZeroDivisionError` from `step % 0` — and the `eval_steps` fallback above is written for that case. So `0` and a negative stay first-class, and the refusal points at that spelling rather than letting "whole number" imply a positive one. A cadence above `steps` is a legitimate configuration too. Neither endpoint is decided here, and both are pinned so they cannot move silently. After the change:

| `save_freq` | `validate()` | tokens built |
|---|---|---|
| `0` | clean | `--save_freq=0` `--eval_steps=100` |
| `-1` | clean | `--save_freq=-1` `--eval_steps=100` |
| `500` | clean | `--save_freq=500` `--eval_steps=500` |

### Scoping

The gate is scoped like the run-size one rather than made universal: `TrainSpec` documents that a backend "reads the fields it supports and ignores the rest", and the RL trainers never read the field, so they must not report on it. That scoping is enforced structurally — the set of providers required to route through the gate is derived from which modules actually read `save_freq`, by attribute access *or* through a forwarding table (the shared `reads_spec_field` rule), so a fifth backend that starts checkpointing from it fails the parity test until it does. The derived scan measures exactly `{cosmos3.py, groot.py, lerobot.py, sagemaker.py}`.

`tests/training/test_spec_field_read_discovery.py` grades the field-scoped guards from the outside and discovers them structurally, so it picked the new guard up on arrival and required it to be registered in `FIELD_SCOPED_GATES` — which is the behaviour its own docstring promises ("a new domain guard is held to the same rule the moment it lands"). Registering it is part of this change, and the new guard now passes the table-driven-read rule too.

## Tests

`tests/training/test_checkpoint_cadence_domain.py`, 126 tests, structured after the launch-topology sibling.

Pre-fix split, on this tree with only the four gate call sites reverted (the gate and the docs kept, so the suite still imports): **57 failed / 69 passed**, 56 of them in `TestEveryReadingBackendRefusesAnUnusableCadence` and 1 in the derived owner check, with **zero** in the control, boundary, premise or gate-contract classes. A raw revert of every changed source file instead fails at collection (`ImportError: cannot import name 'checkpoint_cadence_problems'`), which is why the behavioural split is the one quoted.

Break table (each mutation applied alone; control = unmodified tree):

| mutation | tests firing |
|---|---|
| control | 0 of 126 |
| gate never reports | 61 |
| `bool` admitted (drop the explicit bool test) | 9 |
| integral float coerced instead of refused | 5 |
| **over-reach: the floor folded into the domain** | 16, all in the disable-capability class |
| message drops the disable-capability hint | 4 |
| message does not name the context | 5 |
| only lerobot routes through the gate | 43 |
| only sagemaker skips the gate (the table-driven reader) | 15 |
| **over-reach: the disable selector removed from lerobot** | 1 |

Ten mutations, ten distinct firing sets. The two partial-routing rows are complementary — 42 + 14 = 56, the behavioural pre-fix count — so each provider is independently pinned. The two over-reach rows are what make the scope decisions load-bearing rather than prose: folding the floor in trips only the class that pins the disable capability, and removing the `> 0` selector trips only the test that pins it.

The premise class grades the refused set against both consumers rather than a hand-listed rule, using two classifiers (`_token_decodes`, `_comparison_survives`) plus a non-vacuity test that both failure modes are represented. That matters because the two do not agree: `"5000"` renders to the same token as `5000` and decodes perfectly, and fails *only* the comparison — so a domain argued from the decoder alone would have admitted it.

## Gate

* `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ` (1532 files).
* mypy: **24 == 24** against a pristine `main` worktree, normalized message sets byte-identical both ways, 0 errors in any changed file.
* Paired broad run, identical 402-file selection on both trees (all of `tests/training/`, all of `tests/tools/`, and every test that walks the tree with `ast`/`rglob`/`inspect.getsource` or reads `docs/`), the two changed test files excluded from each: **20 failed, 18068 passed, 74 skipped** on **both**, 20 failing ids each, `comm` empty in both directions. All 20 are dependency- or upstream-drift artifacts present on `main` alone: 17 need `awsiot` (`awsiotsdk`, declared in the `[mesh-iot]` extra) and 3 are the `TypeError: encode() takes 1 positional argument but 2 were given` drift in `tests/training/test_lerobot*`.
* The changed test files pass on lerobot 0.6.2 + mujoco 3.5.0 and on lerobot 0.5.1 + mujoco 3.12.0: **151 passed** on both.

No visual artifact: this changes a preflight verdict, not a policy, simulation behaviour, rendering, recording or asset. The measured token and verdict tables above are the evidence.

## Sequencing note

#2706 landed the same domain on the `lerobot_train` tool's `save_freq` argument while this was in review, and `main` is merged in here. The two are separate surfaces — a tool parameter validated before `build_train_command` interpolates it, versus a `TrainSpec` field read by four trainer backends — and the split mirrors how `steps` was done (#1936 for the tool, #1939 for the trainer). Verified by break on the merged tree rather than by inspection: reverting #2706's call site fires **40** tests, all in its own file; reverting this change's four call sites fires **58**, all in this change's files. Disjoint sets, so nothing was quietly dropped by the merge, and both halves are independently pinned. Both readings agree on the floor for the same documented reason.

### DIFF
--- a/tests/training/test_checkpoint_cadence_domain.py
+++ b/tests/training/test_checkpoint_cadence_domain.py
@@ -1,0 +1,418 @@
+"""The checkpoint cadence a TrainSpec asks for is one shared whole-number domain.
+
+:attr:`~strands_robots.training.base.TrainSpec.save_freq` is the interval, in
+optimizer steps, at which a run writes a checkpoint. Four providers read it - the
+three supervised backends and the SageMaker transport - and each consumes it in
+the same three shapes a process count is consumed in:
+
+* a ``spec.save_freq > 0`` selector, which LeRobot uses twice to derive the
+  validation cadence (``--eval_steps=`` on the argv path, ``cfg.eval_steps`` on
+  the in-process one) because a non-positive cadence disables periodic saving;
+* an argv or Hydra token - ``--save_freq=`` (LeRobot), ``--save_steps=``
+  (GR00T), ``checkpoint.save_iter=`` (Cosmos) - decoded into an ``int`` field;
+* a direct assignment into a typed config (``cfg.save_freq``, GR00T's
+  ``save_steps`` kwarg, a forwarded SageMaker hyperparameter), none of which
+  coerces.
+
+Before this contract the field had no domain while its run-size neighbours
+``steps`` and ``global_batch_size`` - the step count in the very same argv -
+shared one, and every way a bad cadence failed was silent or late:
+
+* ``nan`` and ``False`` compare as *not* greater than zero (``nan`` compares
+  false against everything), so the selector read them as "periodic saving
+  disabled" and a spec asking to checkpoint every ``nan`` steps evaluated once
+  at the very end instead, under a successful result.
+* ``2.7``, ``5000.0``, ``True`` and ``inf`` *are* greater than zero, so they
+  passed through as the cadence itself and reached an ``int`` field as a float
+  or a ``bool``. Their argv token then failed inside the launched run, after the
+  dataset and model were already loaded.
+* A string, ``None`` or a list raised ``TypeError`` out of the comparison itself
+  - from inside a :meth:`Trainer.validate` documented to *return* problems.
+
+What this domain deliberately does *not* decide is the floor. LeRobot documents
+a non-positive ``save_freq`` as disabling periodic saving and its
+``should_save_checkpoint`` implements exactly that, so ``0`` and a negative are
+a capability rather than an unusable value - the ``eval_steps`` fallback above
+exists for them. Only the type is graded, and the boundary is pinned below so it
+cannot move silently.
+
+The tests pin the contract in both directions: every provider that reads the
+field refuses the same unusable values through the one shared gate with a
+message naming itself, the disable capability and a usable cadence are
+untouched, a backend that ignores the field reports nothing about it, and the
+refused values are grounded in what the selector and LeRobot's own decoder do
+with them.
+"""
+
+from __future__ import annotations
+
+import ast
+import inspect
+import math
+import pathlib
+from typing import Any
+
+import pytest
+
+from strands_robots.training._validate import checkpoint_cadence_problems
+from strands_robots.training.base import Trainer, TrainSpec
+from strands_robots.training.cosmos3 import Cosmos3Trainer
+from strands_robots.training.groot import Gr00tTrainer
+from strands_robots.training.lerobot import LerobotTrainer
+from strands_robots.training.mock import MockTrainer
+from strands_robots.training.sagemaker import SagemakerTrainer
+from strands_robots.utils import positive_count_error
+from tests.training._spec_field_reads import reads_spec_field
+
+FIELD = "save_freq"
+
+# The providers that read the cadence. SageMaker forwards it as a
+# hyperparameter; the other three interpolate it into a native token.
+CHECKPOINTING_BACKENDS = (LerobotTrainer, Gr00tTrainer, Cosmos3Trainer, SagemakerTrainer)
+
+# Values no backend can honor, split by how each one failed before the gate.
+
+# Read as "not more than zero" by the selector -> the cadence silently disabled.
+SILENTLY_DISABLED = (float("nan"), False)
+
+# Read as "more than zero" -> passed through as the cadence into an int field.
+REACHED_THE_INT_FIELD = (2.7, 5000.0, True, float("inf"))
+
+# Raised TypeError out of the comparison, from inside validate().
+RAISED_OUT_OF_VALIDATE: tuple[Any, ...] = ("5000", None, [5000])
+
+UNUSABLE = SILENTLY_DISABLED + REACHED_THE_INT_FIELD + RAISED_OUT_OF_VALIDATE
+
+# The floor is NOT part of the domain: a non-positive cadence disables periodic
+# saving, and a cadence above ``steps`` writes only the final checkpoint.
+USABLE = (0, -1, 1, 500, 10**9)
+
+
+@pytest.fixture
+def spec(tmp_path: pathlib.Path) -> TrainSpec:
+    """A spec whose cadence is the only thing under test.
+
+    ``validate`` may well report unrelated problems (GR00T wants a checkout,
+    Cosmos a recipe TOML, SageMaker a role); every assertion below filters for
+    the field name, so an unrelated problem cannot mask - or fake - a cadence
+    verdict.
+    """
+    return TrainSpec(
+        dataset_root=str(tmp_path / "ds"),
+        output_dir=str(tmp_path / "out"),
+        base_model="lerobot/act",
+        embodiment="new_embodiment",
+    )
+
+
+def _problems_about(trainer: Trainer, spec: TrainSpec) -> list[str]:
+    """``validate`` problems that name the cadence field."""
+    return [p for p in trainer.validate(spec) if FIELD in p]
+
+
+def _token_decodes(value: Any) -> bool:
+    """Does ``f"{value}"`` decode into an ``int`` the way LeRobot decodes it?
+
+    LeRobot declares ``save_freq`` as a plain ``int`` and parses the argv token
+    with draccus, so this is the real destination of the interpolated value.
+    """
+    draccus = pytest.importorskip("draccus")
+    try:
+        draccus.decode(int, f"{value}")
+    except Exception:
+        return False
+    return True
+
+
+def _comparison_survives(value: Any) -> bool:
+    """Does the ``> 0`` selector run at all for *value*?"""
+    try:
+        _ = value > 0
+    except TypeError:
+        return False
+    return True
+
+
+class TestEveryReadingBackendRefusesAnUnusableCadence:
+    """Each provider that reads the field refuses every unusable value."""
+
+    @pytest.mark.parametrize("trainer_cls", CHECKPOINTING_BACKENDS)
+    @pytest.mark.parametrize("value", UNUSABLE)
+    def test_it_is_reported_as_a_problem(self, spec: TrainSpec, trainer_cls: type[Trainer], value: Any) -> None:
+        spec.save_freq = value
+        problems = _problems_about(trainer_cls(), spec)
+        assert problems, f"{trainer_cls.__name__} accepted {FIELD}={value!r}"
+        assert any("must be a whole number of steps" in p for p in problems), problems
+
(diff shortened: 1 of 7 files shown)


tests/training/test_checkpoint_cadence_domain.py
```diff
@@ -0,0 +1,418 @@
...
+    is excluded - derived from the gate itself rather than named, so the
+    exclusion cannot drift - because it reads the field as its owner rather than
+    as a consumer of it.
+    """
+    root = pathlib.Path(inspect.getfile(Trainer)).parent
+    owner = pathlib.Path(inspect.getfile(checkpoint_cadence_problems)).resolve()
+    return sorted(p for p in root.rglob("*.py") if p.name != "__init__.py" and p.resolve() != owner)
+
+
+def _reads_the_cadence(source: str) -> bool:
```
yinsong1986, line 318: [MUST FIX] This module's structure (`_trainer_modules` + a `_reads...` helper + `_calls_the_gate`) makes it discoverable by `tests/training/test_spec_field_read_discovery.py::_guard_modules()`, but it was not registered there, so this PR's head fails three tests that pass on the merge base (`1c05394e`):

- `test_the_scan_finds_the_field_scoped_guards` — the expected guard-module set does not include `test_checkpoint_cadence_domain.py`;
- `test_it_sees_a_table_driven_read[test_checkpoint_cadence_domain.py]` and `test_it_still_sees_a_read_by_name[...]` — `next(g for g in FIELD_SCOPED_GATES ...)` finds no gate named in this module's source and raises `StopIteration` -> `RuntimeError`.

Why this can't be deferred: `call-test-lint / Test and Lint` is the one required status check on this repo (AGENTS.md > PR Workflow step 8), and these are deterministic unit-test failures introduced by this branch, so the PR cannot merge as-is.

Fix (two lines in `tests/training/test_spec_field_read_discovery.py`): add `"_checkpoint_cadence_problems": ("save_freq",)` to `FIELD_SCOPED_GATES`, and add `"test_checkpoint_cadence_domain.py"` to the expected set in `test_the_scan_finds_the_field_scoped_guards`. With that registration the meta-test's table-driven-read and read-by-name checks pass against this module's `_reads_the_cadence` (verified locally: it already delegates to the shared `reads_spec_field`, so both read forms are recognized).

yinsong1986 (commented):
## Summary

Adds a shared type-only domain for `TrainSpec.save_freq`: a new `checkpoint_cadence_problems` gate in `training/_validate.py`, a lazily-importing `Trainer._checkpoint_cadence_problems` wrapper, and calls from exactly the four providers that read the field (LeRobot, GR00T, Cosmos3, SageMaker). The floor is deliberately left open so a non-positive cadence keeps meaning "disable periodic saving". The scoping is correct (verified: those four modules are the only readers of `save_freq` in `strands_robots/training/`; the RL trainers and MockTrainer never read it and do not call the gate), the bool-rejects-before-int check is right, and the 418-line guard module passes on its own (126 tests) with both non-vacuity controls and planted-defect checks for its scanners. One regression blocks merge, flagged inline: the new guard module is structurally discovered by the existing meta-test `tests/training/test_spec_field_read_discovery.py` but was not registered there, so three of that file's tests fail on this head (they pass on the merge base), which turns the required `call-test-lint / Test and Lint` check red.

## What's good

- Gate scoped per-field rather than universal, per the `TrainSpec` "ignore unsupported fields" contract, with tests pinning both directions of the biconditional.
- Refusal message names the disable capability (`pass ... a non-positive one to disable periodic saving`), so a caller who meant "never checkpoint" gets the accepted spelling.
- The refused set is grounded in the real consumers (draccus `int` decode + the `> 0` selector) rather than asserted, and `ruff check` / `ruff format --check` are clean on all changed files.

## Verification suggestions

- `pytest tests/training/test_spec_field_read_discovery.py -q` on the head — currently 3 failed / 20 passed; passes 21/21 on the merge base, isolating the regression to this PR.
- After the inline fix, re-run `pytest tests/training/ -q` — the only remaining non-green items (`test_lerobot_e2e.py::test_record_train_load_loop`) pre-exist on the merge base and are not attributable to this PR.

cagataycali:
Closing as the duplicate half of a two-PR claim on one defect: #2708 implements the same `TrainSpec.save_freq` cadence domain and is the one that should proceed. Both were opened inside a 30-minute window (04:14 and 04:43) and both create `tests/training/test_checkpoint_cadence_domain.py`, so they cannot both land - this is the intake collision AGENTS.md > PR Workflow step 1 describes, and the reason it says to prevent the authoring rather than cap it at review.

Why #2708 rather than this one, on the merits:

1. **The shared-domain decision goes the other way now.** #2706 merged 8 minutes before this PR opened, carrying a *local* `_save_freq_error` in `strands_robots/tools/lerobot_train.py` whose docstring records that the rule stays local while it has one caller. `TrainSpec.save_freq` is that second caller, so convention 11 ("a value-domain guard becomes shared when it has a second caller - two copies are the evidence that a shared name is the right one") now asks for the lift. #2708 lifts it to `strands_robots/utils.py::step_cadence_error` and rewrites the tool's guard to delegate (`+15/-22` there, so the copy is removed rather than added to). This PR adds a third independent copy in `strands_robots/training/_validate.py` and leaves the tool's local one in place.

2. **It covers a failure mode this PR's framing does not reach.** This PR grades the argv token lerobot's parser decodes. The in-process backend never renders a token: it assigns the value to `cfg.save_freq`, which reaches `should_save_checkpoint(step, save_freq, total_steps)` directly, so `True` is a modulus of one and writes a full checkpoint every step, while a fractional or non-finite cadence never satisfies `step % cadence == 0` and silently becomes the *disabled* mode. Two routes disagreeing about one spec is what makes the shared owner load-bearing rather than merely tidier.

3. **The blocking review finding here is already satisfied there.** The [MUST FIX] on `tests/training/test_checkpoint_cadence_domain.py:318` names three deterministic failures of the one required check, caused by the new guard module not being registered in `tests/training/test_spec_field_read_discovery.py`. #2708 carries exactly that registration (`"_checkpoint_cadence_problems": ("save_freq",)` in `FIELD_SCOPED_GATES`, plus the expected-set entry), so the concern is preserved rather than discarded by this close - which is the point of closing the half that has not been approved.

Nothing here needs re-deriving. The two CodeQL threads on this branch (redundant comparison / unreachable code at lines 246-247 of the new test file) are worth checking against #2708's own test file before it goes up for review, since the two files share a lineage.

PR #2708, opened 2026-08-25
fix(training): hold TrainSpec.save_freq to one shared step-cadence domain

cagataycali:
## Problem

`TrainSpec.save_freq` is the checkpoint cadence. Four backends read it, each
delivers it to a destination that requires a genuine `int`, and on `main` none
of them said so -- `Trainer.validate()` returned an empty list, which its own
contract defines as "the spec is launchable".

The in-process LeRobot path is the worst of the four, because no parser stands
between the spec and the consumer. `cfg.save_freq` reaches lerobot's own

```python
def should_save_checkpoint(step, save_freq, total_steps):
    return (save_freq > 0 and step % save_freq == 0) or step == total_steps
```

so the value is used as the *modulus of a `step %` test*. Measured against that
expression, with `steps=10000` and a spec asking for 9 checkpoints:

| `save_freq` | what `should_save_checkpoint` does | checkpoints written |
|---|---|---|
| `5000` | as intended | 2 (+ final) |
| `0` / `-5` | documented disable-periodic-saving mode | final only |
| `True` | modulus of **one** -- every optimizer step qualifies | **9999** (+ final) |
| `2.7` | `step % 2.7 == 0` never true for an integral step | final only, *silently* |
| `nan` / `inf` | same -- never true | final only, *silently* |
| `"5000"` | `TypeError` from `save_freq > 0` | crashes **inside the training loop**, after the dataset and model have loaded |

Three distinct wrong answers, and the two quiet ones are the dangerous pair:
`True` burns a full checkpoint write per step, and `2.7` looks exactly like a
user who asked to disable periodic saving.

**The two LeRobot routes also disagreed about the same spec.** The argv route
decodes the identical field with draccus, which refuses `True` / `2.7` /
`5000.0` / `nan` / `inf` but accepts the token `"5000"` renders. So `"5000"` was
accepted by the argv path and raised `TypeError` in-process, while `2.7` was
refused by the argv path and silently disabled saving in-process. `TrainSpec.extra`
already forbids exactly this for its own values: one spec means one run whichever
path the backend takes.

The other three readers each failed differently again, or not at all until the
run had started: GR00T's `--save_steps=`, Cosmos's `checkpoint.save_iter=` Hydra
override, and SageMaker's `json.dumps`-encoded hyperparameter, where `nan` and
`inf` travel as the non-JSON `NaN` / `Infinity`.

## Change

#2706 held this same field to an `int` in the `lerobot_train` tool, as a
module-local `_save_freq_error`. That gave the rule its second caller, so per
convention #11 the domain moves to `strands_robots/utils.py` as
`step_cadence_error(value, param, context)`, and the tool now **delegates** to it
rather than carrying a copy. What stays in the tool is the reason the check
happens at that point -- the detached argv.

A `Trainer._checkpoint_cadence_problems` gate wires that domain into
`validate()` for the four backends that read the field. The reader set is not a
list: it is derived from the tree by the shared `reads_spec_field` rule, which
recognises both forms of a read --

| backend | how it reads `save_freq` |
|---|---|
| `LerobotTrainer` | `spec.save_freq` by name (both routes) |
| `Gr00tTrainer` | `spec.save_freq` by name |
| `Cosmos3Trainer` | `spec.save_freq` by name |
| `SagemakerTrainer` | table-driven: `"save_freq"` in `_FORWARDED_FIELDS` + `getattr(spec, field)` |
| `MockTrainer` | **does not read it** -- correctly not gated |

SageMaker is the case a `spec.save_freq` grep misses and the reason the rule
exists. `MockTrainer`'s exclusion is required, not incidental: per `TrainSpec`, a
backend ignores fields it does not support, so reporting on one it never reads
would be a false rejection. A fifth checkpointing backend is graded on arrival.

## Scope: the floor stays out, and that is pinned

lerobot documents a non-positive `save_freq` as "disables periodic saving (only
the final checkpoint is written)" and `should_save_checkpoint` implements it, so
`0` and a negative **select a mode** rather than being unusable -- and the
backends' own `spec.save_freq if spec.save_freq > 0 else spec.steps` fallback is
written for that case. Only the *type* is graded, which is also what keeps this a
distinct domain from `positive_count_error` and `non_negative_count_error`: they
share the type rule but each impose a floor this mode would trip.

This is not prose. Mutating the domain to add a floor trips 12 tests in the new
file **and 4 pre-existing pins in #2706's merged file**
(`TestTheFloorIsNotPartOfThisDomain`), so the promotion cannot silently widen the
contract #2706 shipped.

## Tests

`tests/training/test_checkpoint_cadence_domain.py`, new: **129 passed, 16
skipped**. With the discovery sibling that registers the new gate: **154 passed,
16 skipped, 0 failed**. The 16 skips are the grounding tests that import real
`lerobot` / `draccus`, absent from this measurement env and present in CI.

The grounding tests exercise lerobot's actual `should_save_checkpoint` and
draccus decoding rather than restating the rule, so the tables above are
measured rather than asserted. The AST scanners carry planted-defect and
table-driven-defect non-vacuity cases.

## Break table

Seven mutations of the shipped code, seven distinct firing sets, unmutated
control 0 failures:

| mutation | tests fired |
|---|---|
| control (unmutated) | **0** |
| gate never called -- the `main` behaviour | 38 |
| only the LeRobot backend gated | 29 |
| refusal drops the parameter name | 37 |
| **over-reach**: floor made part of the domain | 12 (+ **4 pre-existing** in #2706's file) |
| `bool` admitted (bare `isinstance(value, int)`) | 5 |
| tool carries its own copy again (delegation removed) | 2 |
| refusal drops the offending value | 1 |

The `bool` row is why the type test is not a bare `isinstance`: `True` is an
`int` subclass and a modulus of one. The delegation row is an AST pin -- it fires
on a *reimplementation that still behaves correctly*, which is the only way to
keep convention #11 from decaying back into two copies.

## Gate

- `ruff check` and `ruff format --check` clean over every changed file (23 files).
- Paired run against a pristine merge-base worktree (`151614e6`), identical
  selection of `tests/training/` + `tests/tools/test_lerobot_train.py` +
  `tests/tools/test_lerobot_train_checkpoint_cadence_domain.py`, with the new
  file excluded from each side: **122 failing ids on both trees, sets
  byte-identical** (`diff` empty in both directions). Those 122 are
  `lerobot`-not-installed environment noise, not this branch.
- #2706's merged test file passes unchanged: the shared message preserves the
  `save_freq must be an integer number of steps` substring its five assertions
  match, and `_refusal_repr(value)` satisfies its `repr(value) in message` pin
  while additionally being unable to raise while rendering a refusal.

No visual artifact: this adds input validation to a preflight path, touching no
policy, simulation, rendering, recording or asset behaviour.

Cadence domain groundwork per #2706. The intake collision that produced this
pair is filed separately as #2709; no issue is claimed here.

## Round changelog

**Round 1** -- description replaces the intake placeholder, now that
`changelog.d/2708-lerobot-trainer-checkpoint-cadence-domain.md` is on the branch
and carries this PR number. No source change in this round; the measurements
above were taken at `d3cc1387` and the tree is byte-identical to the reviewed
commit `2125c715` plus that fragment.

**Round 2** -- CI reported one failing test, in this branch's own new file:
`TestTheToolAndTheTrainerShareOneOwner::test_the_tool_delegates_rather_than_carrying_a_copy`,
with `TypeError: ... got DecoratedFunctionTool`. Fixed at `c18367d`; the diff is
12 insertions / 2 deletions in that one test file, and no source file is touched.

**What broke.** The delegation pin resolved the tool module's file through
`from strands_robots.tools import lerobot_train`. That name has two possible
resolutions. `strands_robots/tools/__init__.py` maps the tool name to the `@tool`
object inside the submodule of the *same* name and caches the result on the
package, so:

| first import to win in the session | the from-import yields | `inspect.getfile` |
|---|---|---|
| `import strands_robots.tools.lerobot_train` (submodule) | the module | works |
| the lazy package attribute | `DecoratedFunctionTool` | **raises** |

The caching is sticky in one direction: once the lazy attribute has been touched,
a later `import strands_robots.tools.lerobot_train` does **not** restore the
module binding, because the submodule is already in `sys.modules`. That is the
whole reason this passed the measurement in round 1 and failed in CI -- the
narrow selection I measured happened to import the submodule first, and the full
suite touches the lazy attribute first. It was never a flake; it was
selection-dependent and CI picked the losing order deterministically.

**The fix.** The file is derived from `_save_freq_error` itself -- the same object
the three tests beside it already call, and a plain function, which always has a
source file. No name is resolved, so neither ordering exists to get wrong. This
is the idiom the file already uses for `_trainer_modules`, whose docstring says
the exclusion is "derived from the gate itself rather than named, so the
exclusion cannot drift"; the same rule now covers this pin. The docstring records
the trap so it is not reintroduced.

**Re-measured after the fix:**

- The failing test, run **alone** -- the ordering that failed CI: passes.
- The new file whole: **129 passed, 16 skipped**, unchanged from round 1.
- The pin's strength is unchanged, which is the property that mattered here:
  reinstating a behaviourally identical copy in the tool still fires **2** tests
  (`AssertionError: the tool's guard should be a single delegation`, `assert 2 ==
  1`), unmutated control **0**. The break table above stands as written.
- Paired run against a pristine merge-base worktree (`151614e6`), selection
  `tests/training/` with the new file excluded from both sides: **115 failing ids
  on both trees, `diff` empty**. Those 115 need a real `torch`
  (`_tensor() got an unexpected keyword argument 'requires_grad'`) or absent
  training extras, and are not this branch. Narrower than round 1's selection
  because `tests/tools/conftest.py` needs `serial`, absent from this env.
- `ruff check` and `ruff format --check` clean on the changed file.

**Not fixed here, deliberately.** Sweeping the tree for the same hazard found two
**pre-existing** sites that read a lazily-mapped tool name off the package and
then treat it as a module -- one takes `Path(<tool>.__file__)`, the other calls
`<tool>.<toolname>(...)`. Both are order-dependent today for the same reason, and
both are rescued only by an unrelated import that happens to run first -- for
one of them, the line directly above it. Neither is on this branch and neither is
this branch's contract, so they are filed as #2711 with their measurements rather
than widened into this diff. No closing keyword: this pull request still claims no
issue.


yinsong1986 (dismissed):
## Summary

Promotes the `save_freq` type guard from a single-caller local rule in `lerobot_train` into a shared `step_cadence_error` domain in `strands_robots/utils.py`, and wires a `_checkpoint_cadence_problems` gate into the four backends that actually read the field (LeRobot, GR00T, Cosmos3, SageMaker) via `Trainer.validate()`. This closes a real parity hole: the same spec value (`"5000"`, `2.7`, `True`) was previously refused by one LeRobot route and silently honored — or worse, obeyed as a modulus of one — by the other. Only the type is graded; `0` and negative cadences remain the documented disable-periodic-saving mode, with controls pinning both directions. The promotion follows AGENTS.md convention #11 exactly (the guard gained a second caller, so it moved to `utils.py`), MockTrainer is correctly excluded per TrainSpec's ignore-unread-fields contract, and the tool now delegates rather than carrying a copy, with an AST test pinning the delegation.

## What's good

- 170 new tests pass at the head commit; the full `tests/tools/test_lerobot_train.py` + `tests/training/` suite passes (the one `test_lerobot_e2e.py` setup error fails identically on the merge base `151614e6`, so it is pre-existing environment noise, not this PR).
- The grounding tests exercise lerobot's real `should_save_checkpoint` and draccus decoding rather than restating the rule, and include non-vacuity checks for the AST scanners (planted-defect and table-driven-defect cases).
- The reader-set scan is derived from the tree via the shared `reads_spec_field` rule, so a fifth checkpointing backend is graded on arrival; `test_spec_field_read_discovery.py` registers the new gate.
- `ruff check` / `ruff format --check` clean on all changed files; new user-facing error strings are ASCII.

Note: the `changelog.d/2708-*.md` fragment is not yet on the branch — the placeholder description says it is coming, and `.github/workflows/changelog-fragment.yml` gates on it, so CI will hold the merge until it lands; no separate action needed from review.

yinsong1986 (dismissed):
## Summary

Promotes the `save_freq` type guard that #2706 shipped as a module-local rule in `lerobot_train` into a shared `step_cadence_error` domain in `strands_robots/utils.py` (per AGENTS.md convention #11 — the rule gained its second caller), and wires a `Trainer._checkpoint_cadence_problems` gate into the four backends that read the field (LeRobot, GR00T, Cosmos3, SageMaker), closing the route-parity hole where one spec value (`"5000"`, `2.7`, `True`) was refused by the argv path and silently obeyed — or obeyed as a modulus of one — by the in-process path. Only the type is graded; `0` and negatives remain the documented disable-periodic-saving mode, with the no-floor scope pinned in both this PR's tests and #2706's merged file.

## What's good

- Verified at `d3cc1387`: the new file plus the discovery sibling pass 170/170 with real `lerobot` 0.6.1 and `draccus` installed, so the grounding tests (lerobot's actual `should_save_checkpoint`, draccus decoding) executed rather than skipped; `tests/tools/test_lerobot_train_checkpoint_cadence_domain.py` (#2706's merged pins) passes unchanged at 68/68; the full `tests/training/` suite is 2895 passed with only the known pre-existing `test_lerobot_e2e.py` environment error.
- `ruff check` / `ruff format --check` clean on the changed files; the new refusal text is ASCII and uses `_refusal_repr`, consistent with the sibling domains.
- Reader-set discovery goes through the shared `reads_spec_field` rule, so the table-driven SageMaker read is caught and a fifth checkpointing backend is graded on arrival; `MockTrainer`'s exclusion is asserted, not assumed.
- The delegation is pinned structurally (AST), which is the only shape that keeps convention #11 from decaying back into two copies.
- Changelog fragment `changelog.d/2708-*.md` is on the branch with the correct PR number; the `d3cc1387` head is `2125c715` plus exactly that fragment, matching the round changelog.

## Verification suggestions

- `pytest tests/training/test_checkpoint_cadence_domain.py tests/training/test_spec_field_read_discovery.py tests/tools/test_lerobot_train_checkpoint_cadence_domain.py` in a CI-like env with `lerobot`/`draccus` present exercises the grounding halves the author's measurement env skipped (16 skips in the PR description); they run and pass with lerobot 0.6.1.

yinsong1986 (approved):
## Summary

Promotes the `save_freq` type guard #2706 shipped as a module-local rule in `lerobot_train` into the shared `step_cadence_error` domain in `strands_robots/utils.py` (convention #11 - the rule gained its second caller), wires a `Trainer._checkpoint_cadence_problems` gate into the four backends that read the field (LeRobot, GR00T, Cosmos3, SageMaker), and - in round 2 at `c18367da` - fixes the delegation pin's file resolution to derive from `_save_freq_error` itself rather than resolving `strands_robots.tools.lerobot_train` by name, which the package's lazy `__getattr__` can bind to the `DecoratedFunctionTool` instead of the module depending on import order. Only the type is graded; `0` and negatives remain the documented disable-periodic-saving mode, pinned in both this PR's tests and #2706's merged file.

## What's good

- Verified at `c18367da`: the new file + discovery sibling + #2706's merged pins pass 238/238 with real `lerobot` 0.6.1 and `draccus` installed, so the grounding halves (lerobot's actual `should_save_checkpoint`, draccus decoding, `lerobot.common.train_utils` import path) executed rather than skipped; `tests/tools/test_lerobot_train.py` + the #2706 file pass 119/119; the full `tests/training/` suite is 2895 passed with only the known `test_lerobot_e2e.py` setup error, which fails identically on the merge base `151614e6`.
- The round-2 fix holds under the ordering that failed CI: reproducing the losing order (lazy package attribute touched before the submodule import) in one process, the delegation pin still resolves the source file and passes - the fix removes the name resolution entirely rather than patching around it.
- `ruff check` / `ruff format --check` clean on the changed files; the refusal text is ASCII and routes through `_refusal_repr`, consistent with the sibling domains, and preserves the `save_freq must be an integer number of steps` substring #2706's five assertions match.
- Reader-set discovery goes through the shared `reads_spec_field` rule with the gate-owner exclusion derived from the gate itself, so SageMaker's table-driven read is caught and a fifth checkpointing backend is graded on arrival; `MockTrainer`'s exclusion is asserted, not assumed.
- The two order-dependent pre-existing sites the same hazard sweep found are correctly left out of this diff and filed separately, matching the PR's stated scope.

PR #2508, opened 2026-08-19
docs(notebooks): 07_data_augmentation.ipynb - three shipped mechanisms CPU-only, generative section markdown-only

yinsong1986:
Closes #2467

## What changed

`examples/notebooks/07_data_augmentation.ipynb` (new, 16 cells) plus a row in
`examples/notebooks/README.md` and a changelog fragment. The notebook teaches
the **perturb vs generate** narrative across the three stages augmentation
ships at:

1. **Collection time** - `randomize(randomize_colors/lighting/physics, seed=SEED)`
   and `set_obs_noise(joint_pos_std, joint_vel_std, camera_jitter_px, seed=SEED)`
   (example 12's machinery, via the public methods rather than
   `_dispatch_action`), then a 60-frame episode recorded with the perturbations
   live so they land in the dataset (`start_recording(root="/tmp/nb7_dataset", ...)`,
   MockPolicy, `control_frequency` matched to `fps`).
2. **Render time** - `HybridCompositor` over a `ground_plane=False` world with
   the zero-asset procedural `PanoramaBackground`; the swap is
   `set_background(...)`, which is the same call the isaac_gs example's
   `change_background` tool shims. The photoreal `GsplatBackground` path is
   documented in markdown with the `gsplat_rasterizer_available()` probe and
   the explicit CUDA-only refusal it raises on CPU (no silent downgrade -
   #2321 is cited).
3. **Train time** - `TrainSpec.augmentation` inspected through
   `Gr00tTrainer.build_command()` (the documented pure argv-parity helper), so
   the `--random_rotation_angle` / `--color_jitter_params` mapping is shown
   without a GPU or a GR00T checkout, and labeled GR00T-provider-specific.
4. **Generative augmentation** - markdown only, **because generation needs a
   GPU and this notebook is CPU-only**. The section describes the shipped
   `strands_robots.transforms` surface in the present tense
   (`create_transform("cosmos_transfer", ...)`, `TransformSpec`) together with
   the provenance doctrine (`synthetic=true` metadata, re-validation by
   deterministic predicates, discard on verdict flip), and points at
   `docs/data/transforms.md`.

## Acceptance criteria from #2467

- [x] **Executes top-to-bottom on CPU; seeded** - verified with nbclient
  against this tree: 16.2 s wall clock, no GPU touched. A single `SEED = 0` is
  consumed by `randomize()`, `set_obs_noise()`, and `TrainSpec.seed`.
- [x] **Every cell's committed output is produced by the shipped code** -
[...]

PR #2710, opened 2026-08-25
fix(scripts): collide a claim-free duplicate pair on the file both branches create

### LINKED ISSUE #2709: The duplicate-claim gate is blind to a pair that claims no issue (2707 / 2708)
## What happened

#2707 and #2708 are two implementations of one defect - a value-domain guard for `TrainSpec.save_freq` - opened 29 minutes 26 seconds apart (04:14:26 and 04:43:52 UTC on 2026-08-25) by the same account, from two scheduled runs. Both create the same new file, `tests/training/test_checkpoint_cadence_domain.py`, and both edit the same five `strands_robots/training/` modules, so they could never both land. #2707 had already collected a MUST FIX review comment naming three deterministic failures of the one required check by the time the second was opened; it is now closed in favour of #2708, which is the convention-11-correct shape (it lifts the domain into `strands_robots/utils.py` now that #2706 gave the rule a second caller).

That is the intake collision AGENTS.md > PR Workflow step 1 already documents, third instance, inside the same 35-minute window its three measured pairs share.

`scripts/check_duplicate_claim.py` is keyed on an issue number in both of its modes - `--issue N` at intake, `--pr N` at review - and it reads `closingIssuesReferences`. Measured on this pair:

| pull request | closingIssuesReferences.totalCount |
|---|---|
| #2707 | 0 |
| #2708 | 0 |

Neither claims an issue, so there is no key to collide on, and both modes report a unique claim while looking straight at the pair. The check is not wrong; the pair is outside the relation it computes. Nothing else covers it either: `check_closing_reference.py` reads one pull request in isolation, and the per-branch merge signals (`mergeable`, `mergeStateStatus`, `behind_by`) are all base-relative, which AGENTS.md already records as blind to two pull requests that are both still open.

This is the reassuring failure direction, which is what makes it worth a check rather than vigilance: the answer a duplicate gets is shaped exactly like the answer a unique claim gets.

## The signal already exists, one sweep away

`scripts/check_merge_base_overlap.py --all-open` intersects the changed-path set of every open pair. This pair shares a newly added path, which is a stronger relation than the shared-edit one that sweep reports: two branches independently creating one file is not a composition risk to verify, it is two answers to one question. So the measurement needed is already computed - what is missing is a caller at intake and a verdict that separates both-added from both-edited.

The live reproduction is gone (#2707 is closed, so the open set no longer contains the pair). Reconstruct it from `pulls/2707/files` and `pulls/2708/files`, which outlive the state change.

## Suggested shape

- Give `check_duplicate_claim.py` a mode that needs no issue number: over the open set, report a pair whose added paths intersect. Keep it distinct from the shared-edit finding, which is a composition question with a different remedy.
- Consider whether the intake question in AGENTS.md step 1 should be that mode rather than the issue-keyed one, since a claim-free pull request is the case the issue-keyed one cannot answer, and a changed-path set exists for every branch whether or not it claims anything.
- Measure the false-positive rate before wiring it to anything: AGENTS.md records that widening a path relation to a walked root selected 11 of 36 open pairs with no defect among them. An added-path intersection is much narrower than an edited-path one, and that narrowness is the whole claim.

## Cost, for prioritisation

What a duplicate spends is a review approval on a change that could never ship, and review is the scarcest resource here. This pair spent a substantive MUST FIX review round plus two CodeQL threads on the half that was closed, and both halves of the authoring.

cagataycali:
Closes #2709

## Problem

`scripts/check_duplicate_claim.py` has one key, `closingIssuesReferences`, and both of its modes read it: `--issue` at intake, `--pr` at review. Measured over the last 300 pull requests (#2345 through #2708), **249 of them link no issue at all**. For that 83% of the traffic there is no key to collide on, so both modes report a unique claim while looking straight at a duplicate pair.

Two of the four duplicate pairs in that window were exactly that shape. Both halves of each were open at once, both claimed nothing, and both independently created one new test file:

| pair | opened apart | shared added path | outcome |
|---|---|---|---|
| #2388, #2389 | 14m 41s | `tests/test_recorder_counters_track_on_disk_frames.py` | #2389 closed as a duplicate of #2388 |
| #2707, #2708 | 29m 26s | `tests/training/test_checkpoint_cadence_domain.py` | #2707 closed after a MUST FIX review round |

Measured with the shipped check on the live half of the second pair:

```
$ python3 scripts/check_duplicate_claim.py --repo strands-labs/robots --pr 2708
Outcome: **no-claim**
This pull request links no issue, so it cannot duplicate another's claim.
```

`closingIssuesReferences.totalCount` is `0` on all four of those pull requests. Exit `0`, and the report is shaped exactly like the answer a unique claim gets, which is the reassuring failure direction.

### Why the residual was recorded as out of scope, and why that was too wide

The module already listed this case as deliberately excluded, on the strength of one measurement: 18 of the last 30 merges would fail a rule *requiring* a claim. That measurement stands, and neither claim-keyed mode demands a claim. What it does not support is the conclusion drawn from it. It rules out demanding a claim; it says nothing about colliding the pair on a **different** key, and a changed-path set exists for every branch whether it claims anything or not.

## Change

A third mode, `--all-open`, keyed on the paths a branch **adds**. It takes no issue number, reads the open set in one query, and reports every pair whose added-path sets intersect.

Reconstructed from `pulls/2707/files` and `pulls/2708/files`, which outlive the state change that closed one half:

```
## Duplicate work - two branches creating one file

Outcome: **duplicate-addition**

#2707 and #2708 both create `tests/training/test_checkpoint_cadence_domain.py`. Two branches
creating one file are two answers to one question rather than a composition to verify, and
whichever is closed spends a review approval on a change that will not ship.

| pull requests | file(s) both create |
|---|---|
| #2707 + #2708 | `tests/training/test_checkpoint_cadence_domain.py` |
```

### The narrowness is the whole claim

A relation over paths is only worth reporting if it is precise, so it was measured before being wired to anything. Over the same 300 pull requests, on the **1802 pairs that were open at the same instant**:

| relation | pairs selected | duplicates among them |
|---|---|---|
| both **edit** some path | 117 (6.5%) | a composition question, not this one |
| both **add** some path | **2 (0.1%)** | **2** |

Those 117 are the sibling sweep's question, with a different remedy (a merge order, possibly one composition run), and `scripts/check_merge_base_overlap.py --all-open` already owns it. Two branches *editing* one file is a composition to verify; two branches *creating* one file is not a composition at all. For contrast, the path relation this repository has already rejected -- widening a path intersection to a test's walked root -- selected 11 of 36 pairs and named no defect. A finding attached to that much of the queue reads as boilerplate.

The changelog fragment is why the added-path relation stays quiet: every branch adds one, and its name embeds the pull request number, so it can never collide. That is asserted as a premise of the fixtures rather than left to luck.

### The two keys are complementary, so neither replaces the other

| pair | claims an issue | shares an added path |
|---|---|---|
| #2570, #2571 (both claim #2569) | yes | no |
| #2480, #2508 (both claim #2466) | yes | no |
| #2388, #2389 | no | yes |
| #2707, #2708 | no | yes |

Four duplicate pairs, two reachable from each key, none from both. `--issue` and `--pr` are untouched.

### Two policies kept from this file rather than borrowed from the sibling sweep

- **Drafts are included.** The sibling sweep excludes them because a draft cannot merge; this file includes them because a draft's link is a real claim. The same reasoning carries: a draft's new file is authored work whatever its merge state, so excluding one would hide a collision for as long as either side stayed a draft. One of the three currently open pull requests is a draft.
- **The open set is read through `repository.pullRequests(states: OPEN)`, not `search`.** Search is eventually consistent, and a pull request opened seconds ago is exactly the row this sweep exists to find.

This is also why the mode reads `changeType` from GraphQL in the same query as the open list, rather than reusing the sibling's REST file reader: that reader's open-set helper carries the opposite draft policy, and importing it would have imported the policy silently.

### Why this one cannot be asked at intake

`--issue` is asked before authoring, so it prevents the work. `--all-open` cannot be, and the module now says so instead of leaving it implied: a path set is a property of a pushed branch, so at intake there is nothing to read. It therefore caps the review cost of a collision rather than preventing the authoring, which is what `--pr` does for a claim. It still arrives early enough to matter -- both claim-free pairs opened inside the same ~35-minute window every other observed collision shares.

An unreadable answer is not a pass: a truncated file list, an unreadable open set and an API error all reach `unknown-additions`, for the reason the claim-keyed modes refuse a truncated link set.

## Tests

`tests/test_duplicate_work_added_path_key.py` (65 tests) pins both measured claim-free pairs as data, reconstructed from their real file lists.

Pre-fix split, with `scripts/check_duplicate_claim.py` and `AGENTS.md` reverted to `upstream/main`: **56 failed / 9 passed**. The 9 that pass are the controls, and they are the point -- 4 grade the claim key (2 assert the blindness, 2 assert it still catches what it was built for), 4 are usage-error rows, and 1 asserts the intake command survives unchanged.

| break | failing | notes |
|---|---|---|
| `added_paths` accepts every `changeType` (the edited relation) | 6 | every non-`ADDED` row, plus the create-vs-edit case |
| a `REMOVED` path collides too (over-reach) | 1 | the `REMOVED` row alone |
| prose exempt, borrowing the sibling sweep's `.md` rule | 1 | the prose boundary alone |
| only adjacent pull requests compared | 3 | includes the 3-way pair enumeration |
| a truncated file list read short | 1 | |
| an unreadable set reported as `unique-additions` | 1 | the silent no-op |
| drafts excluded (the sibling sweep's policy) | 1 | |
| the finding exits `0` | 1 | |
| only the first page of the open set read | 3 | the finding exists only across the page boundary |
| the docstring measurement dropped, code kept | 2 | |
| `AGENTS.md` not updated, code kept | 5 | |

11 mutations, **11 distinct firing sets**, control 0. Re-measured after `ruff format`, which collapsed one anchor.

## Verification

- Paired run of an identical 413-file selection (every test naming a gate script, the guidance, the changelog convention or a ruleset, plus every test that walks the tree with `ast`/`rglob`/`inspect.getsource` or reads `AGENTS.md` / `docs/`), on this branch and on a clean `upstream/main` worktree, with the new test file excluded from both: **17 failed, 18271 passed, 97 skipped on BOTH**. Failing-ID sets identical, `comm` empty in both directions. All 17 need `awsiot` (`awsiotsdk`, declared in the `[mesh-iot]` extra) and are absent locally.
- `ruff check` and `ruff format --check` over `strands_robots tests tests_integ`: clean. The script is not in CI's lint scope but was held to the same rules.
- `mypy`: **24 == 24** against a pristine base worktree, normalized message sets byte-identical, none in the changed files.
- `python3 scripts/assemble_changelog.py --check`: OK, 742 pending.
- Dogfooded: run against the live open set with this pull request in it, the new mode reports `unique-additions` over 3 pull requests and 3 pairs, exit `0`.

No visual artifact: this changes a repository check, its guidance and its tests, and touches no policy, simulation, rendering, recording or robot asset. The measurements above are the evidence.

### One test helper widened

`tests/test_duplicate_claim_check.py`'s `_documented_intake_argv` asserted that `AGENTS.md` invokes this script exactly once, and step 1 now documents two commands. It selects the intake command by `--issue` instead. The failure it exists to produce -- a shortened intake command that no longer names the repository -- is unchanged; asserting uniqueness would have failed on the sibling command instead.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -268,6 +268,33 @@ hatch run format            # ruff check --fix, ruff format
    is why it belongs at step 1 and not at review. If a competing implementation is
    wanted on purpose, exactly one should claim the close and the other should
    cross-reference (`per #N`, `towards #N`) instead.
+
+   **The claim is not the only key.** 249 of the last 300 pull requests (#2345
+   through #2708) link no issue at all, so for most of the traffic the query above
+   has nothing to collide on - and two of the four duplicate pairs in that window
+   were claim-free: #2388/#2389 and #2707/#2708, each pair independently creating
+   one new test file. Once your branch is pushed and the pull request is open, ask
+   the second question:
+
+   ```
+   python3 scripts/check_duplicate_claim.py --repo strands-labs/robots --all-open
+   ```
+
+   It reports only pairs that **create the same path**, which over the 1802 pairs
+   that were open at the same instant selected 2 - and both were duplicates. Two
+   branches *editing* one file is a different question with a different remedy
+   (a merge order, possibly one composition run), and
+   `scripts/check_merge_base_overlap.py --all-open` owns it; that relation selects
+   117 of the same 1802. The two keys are complementary rather than nested: neither
+   issue-keyed pair shares an added path, and neither claim-free pair claims an
+   issue.
+
+   This one cannot be asked before you start, and not for want of trying: a path
+   set is a property of a pushed branch, so there is nothing to read at intake. It
+   caps the review cost of a collision rather than preventing the work, which is
+   why it belongs here and `--issue` belongs above. It still arrives early enough
+   to matter - both claim-free pairs opened inside the same ~35-minute window every
+   other observed collision shares, 14m 41s and 29m 26s apart.
 2. Make changes, run `hatch run format && hatch run lint && hatch run test`
 3. Record the change as a news fragment: `changelog.d/<pr-number>-<slug>.md`
    (see [`changelog.d/README.md`](changelog.d/README.md)). **Never append to
--- a/changelog.d/2710-duplicate-work-added-path-key.md
+++ b/changelog.d/2710-duplicate-work-added-path-key.md
@@ -1,0 +1,45 @@
+### Fixed: a duplicate pair that claims no issue is now reported, keyed on the file both branches create
+
+`scripts/check_duplicate_claim.py` had one key, `closingIssuesReferences`, and both of its modes
+read it -- `--issue` at intake, `--pr` at review. Measured over the last 300 pull requests (#2345
+through #2708), **249 of them link no issue at all**, so for that 83% of the traffic there is no
+key to collide on: both modes report a unique claim while looking straight at a duplicate pair.
+Two of the four duplicate pairs in that window were exactly that shape, and each spent a full
+authoring plus review round on a change that could never ship:
+
+    shared added path                                       pair           closed
+    tests/test_recorder_counters_track_on_disk_frames.py    #2388, #2389   #2389
+    tests/training/test_checkpoint_cadence_domain.py        #2707, #2708   #2707
+
+The residual was recorded in the module as out of scope, justified by one measurement: 18 of the
+last 30 merges would fail a rule *requiring* a claim. That measurement stands and the conclusion
+drawn from it was wider than it supports -- it rules out demanding a claim, which neither
+claim-keyed mode does, and says nothing about colliding the pair on a different key. A changed-path
+set exists for every branch whether it claims anything or not.
+
+A third mode, `--all-open`, reports a pair whose **added** paths intersect. The narrowness is the
+whole claim: over the 1802 pairs that were open at the same instant, the added-path relation
+selects **2** and both are duplicates, where intersecting *any* changed path selects 117. Those 117
+are a composition question with a different remedy -- a merge order, possibly one test run -- and
+`scripts/check_merge_base_overlap.py --all-open` already owns it; two branches *creating* one file
+is not a composition to verify but two answers to one question. For contrast, the path relation
+this repository has already rejected (widening to a test's walked root) selected 11 of 36 pairs and
+named no defect.
+
+The two keys are complementary rather than nested, so neither replaces the other: neither
+issue-keyed pair in that window shares an added path, and neither claim-free pair claims an issue.
+Four duplicate pairs, two reachable from each key, none from both. The new mode also keeps this
+file's own policies rather than the sibling sweep's -- drafts are included, because a draft's new
+file is authored work whatever its merge state, and the open set is read through
+`repository.pullRequests(states: OPEN)` rather than `search`, which is eventually consistent and
+would miss a pull request opened seconds ago. That is the row this sweep exists to find: both
+claim-free pairs opened inside the same ~35-minute window every other observed collision shares,
+14m 41s and 29m 26s apart.
+
+It cannot be asked at intake, and the module now says so instead of leaving it implied: a path set
+is a property of a pushed branch, so there is nothing to read before the work starts. `--all-open`
+caps the review cost of a collision, which is what `--pr` does for a claim, while `--issue`
+remains the half that prevents the authoring. A truncated file list, an unreadable open set and an
+API error all reach `unknown-additions` rather than a pass, for the reason the claim-keyed modes
+refuse a truncated link set: a sweep that reports clean because it could not look far enough is
+worse than no sweep, because it looks like one.
--- a/scripts/check_duplicate_claim.py
+++ b/scripts/check_duplicate_claim.py
@@ -69,17 +69,80 @@ through ``search``. Search is eventually consistent, so a pull request opened
 seconds ago may not be indexed -- and the missing row would be a false clean in
 exactly the ~35-minute window where every observed collision happened.
 
-Two questions, one comparison
------------------------------
+Three questions, two keys
+-------------------------
 ``--issue N``
-    Intake. *Before* authoring: does an open pull request already close #N?
-    Nothing is excluded from the comparison, because no pull request for it exists
-    yet.
+    Intake, keyed on the claim. *Before* authoring: does an open pull request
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Adds a second key to `scripts/check_duplicate_claim.py`: a new `--all-open` mode that sweeps the open pull-request set for two branches that *create* the same file, catching the claim-free duplicate pairs (`#2388/#2389`, `#2707/#2708`) that both claim-keyed modes are blind to by construction. The relation is deliberately narrowed to `changeType == ADDED` (2 of 1802 pairs, both duplicates) rather than any shared path (117 of 1802, which `check_merge_base_overlap.py` owns), and the module docstring, AGENTS.md step 1, and the changelog fragment all carry the measurement that justifies the narrowness. The failure-mode discipline matches the existing modes: a truncated file list, a truncated open set, and an API error all reach `unknown-additions` rather than a false clean, exit status stays `1` only for a finding, and the sweep keeps this file's own policies (drafts included, `repository.pullRequests(states: OPEN)` rather than `search`).

## What's good

- The new outcome vocabulary (`unique-additions` / `duplicate-addition` / `unknown-additions`) is disjoint from the claim-keyed one, and a test pins that so a report reader can always tell which relation fired.
- `tests/test_duplicate_work_added_path_key.py` pins the design decisions (key complementarity, ADDED-only narrowness, incomplete-answer-is-not-a-finding) against the real measured pairs, including the pagination boundary case where the finding only exists across pages.
- The `_documented_intake_argv` helper in the existing test file was updated to select the `--issue` invocation rather than asserting uniqueness, with the reasoning recorded in the docstring.
- `--all-open` argument handling (exactly-one-subject refusal at exit 2, inferred-repo default kept for review-side callers) is consistent with the sibling sweep scripts.
- Both new-file test suites and the repo's convention graders (test names, host paths, ASCII) pass on the head commit.

## Verification suggestions

- A live smoke test needs only read scope: `python3 scripts/check_duplicate_claim.py --repo strands-labs/robots --all-open` should report the open set it read and exit 0 unless two open branches currently create one path.

PR #2712, opened 2026-08-25
fix(isaac): read both MJCF spellings of a free joint

cagataycali:
## Problem

MJCF states a floating base two ways, and MuJoCo compiles both to the same `mjJNT_FREE`:
`<joint type="free">` and the dedicated `<freejoint>` element. `load_mjcf` walked
`body_el.findall("joint")`, so only the first spelling produced a `JointDef`. A `<freejoint>`
produced none at all: the base's six degrees of freedom were absent from the returned
`ProceduralRobot` rather than reported as a joint a caller can see and skip.

`<freejoint>` is the spelling the shipped asset corpus uses. Of the 46 loadable robot MJCFs under
`robot_descriptions` that declare a free joint, **every one spells it `<freejoint>` and none
spells `type="free"`**, so a floating base was reported for none of them. Measured through the
public loader, with MuJoCo's own count beside it:

| asset | MuJoCo compiles | main reports | this branch |
| --- | --- | --- | --- |
| `skydio_x2` | 1 joint (nv=6) | **0 - no joints at all** | 1 |
| `unitree_go2` | 13 joints (nv=18) | 12 - the base is absent | 13 |
| `unitree_h1` | 20 joints (nv=25) | 19 - the base is absent | 20 |

The two aircraft are the clearest case: their only joint *is* the free base, so the loader
reported a robot with no joints whatsoever, under a successful load.

The scene reader in the same module already consulted both spellings
(`body_el.find("freejoint") is not None or any(j.get("type") == "free" ...)`), and the module's
own docstring already named them as equivalent, so the robot reader was the single surface that
knew about only one.

## Change

Both tags are read, in document order. MuJoCo fixes the two rules that make this unambiguous, and
each is measured and pinned as a premise rather than assumed:

- **A free joint may not share a body with any other joint.** Both spellings are refused with
  `more than 6 dofs in body 'b'`, so a body carrying a `<freejoint>` carries nothing else, the
  order of the two tags can never differ, and the change cannot reach the
  `_validate_kinematic_tree` compound-joint guard (which still refuses a real 2-DOF body).
- **`<freejoint>` resolves no default class.** MJCF has no `<default><freejoint>` block (MuJoCo
  refuses one), and with `<default><joint damping="5" armature="0.3"/>` in force MuJoCo gives a
  `<freejoint>` `dof_damping=0.0, dof_armature=0.0` while the `type="free"` spelling inherits
  `5.0 / 0.3`. The loader now diverges in exactly that place and in exactly that direction.

Both rules are identical on mujoco 3.5.0 (the declared floor) and 3.12.0. `<freejoint>` also
accepts only `name`, `group` and `align`; MuJoCo refuses `type`, `axis` and `class` on it, so no
reader has to resolve them.

## Scope, deliberately not widened

What `free` *maps to* is unchanged. `_MJCF_JOINT_TYPE_MAP` still reports it as `"fixed"` because
`JointDef` has no 6-DOF spelling, so a floating base is now visible in `joints` without being
counted as an actuated DOF by `num_joints`. Whether the dataclass should gain such a spelling is a
contract question about every producer of a `ProceduralRobot`; it is pinned both ways here
(`TestTheJointTypeVocabularyIsUnchanged`) so that moving it later is an explicit decision rather
than a side effect of this one.

## Corpus grading

Every `<freejoint .../>` in the corpus was rewritten as `<joint ... type="free"/>` and MuJoCo was
asked whether that is the same model (`nq`, `nv`, `njnt`, `jnt_type`). Only files where it says
yes are graded, so any difference in the loader's two reports is the loader's. All nine `JointDef`
fields are compared, so nothing can hide in a field the sweep forgot to name.

| reading | premise holds | graded | joint-count parity | reports byte-identical | joints under-reported |
| main | 48 of 48 | 43 | 0 of 43 | 0 | 43 |
| this branch | 48 of 48 | 43 | **43 of 43** | 20 | **0** |

The 23 branch reports that are not byte-identical differ **only** in fields a `<default><joint>`
class supplies (`damping` 21, `armature` 15, `axis` 4, `limits` 2), and MuJoCo diverges the same
way on all 23 - that is the MJCF rule above, not a residual. Nothing is unexplained
(`residual_other: 0`), no premise is violated, and the 5 files the loader refuses are refused
identically on both readings.

## Tests

`tests/simulation/isaac/test_mjcf_freejoint_spelling.py`, 30 cases, every expectation derived
from `mujoco.MjModel` rather than restated by hand.

Pre-fix split on pristine `upstream/main`: **11 failed, 19 passed** - 6 in
`TestAFreejointIsReported`, 2 in `TestBothSpellingsReportOneModel`, 2 in
`TestTheDefaultClassBoundaryFollowsMjcf`, 1 in `TestTheJointTypeVocabularyIsUnchanged`, and
**0 in the 8 premise cases and 0 in the 7 control cases**. The one boundary-class failure is
there because the boundary (a floating base is reported but not counted) is only observable once
the joint exists at all; its first assertion is the premise for its second. Headline failure:
`MuJoCo compiles 1 joint for this model; the loader reported 0`.

### Mutation table

| mutation | tests fired | where |
| --- | --- | --- |
| control, no mutation | 0 | - |
| b1 the whole fix reverted: only `<joint>` is read | 11 | the four non-control classes; exactly the pre-fix set |
| b2 only `<freejoint>` is read (over-reach the other way) | 12 | incl. `TestTheOtherJointsAreUnchanged` |
| b3 a `<freejoint>` resolves a default class after all | 1 | the class boundary |
| b4 a `<freejoint>` takes its type from the class too | 2 | the class boundary |
| b5 a `<freejoint>` is mapped to revolute rather than free | 3 | parity + boundary + vocabulary |
| b6 a `<freejoint>` is always named `freejoint` | 4 | naming + parity |
| b7 a `<freejoint>` carrying `group`/`align` is skipped | 2 | the accepted-extras cases |
| b8 the free joint reports MuJoCo's dof values, not the loader's own defaults | 1 | field-for-field parity |
| b9 both tags read via two concatenated `findall()` passes | **0** | see below |

b9 fires nothing, and that is the finding rather than a hole: because MuJoCo refuses a free joint
beside any other joint on one body, a `<freejoint>` is the only joint on its body, so a
concatenated two-pass read can never produce a different order from a document-order read. The
mutation is semantically identical to the shipped code, and the premise that makes it so is
itself a test (`test_a_free_joint_may_not_share_a_body_with_another_joint`).

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1532 files).
- mypy: **5 == 5** against a pristine `upstream/main` worktree, message sets byte-identical, 0 in
  the changed files.
- Paired run over an identical 495-file selection (all of `tests/simulation/isaac/` plus every
  test naming the loaders, `ProceduralRobot`/`JointDef`/`BodyDef`, or walking the tree with
  `ast`/`rglob`/`inspect.getsource`), the new file excluded from both trees: **20287 collected and
  `18 failed, 20088 passed, 181 skipped` on both**, 18 failing ids each, both `comm` directions
  empty. All 18 are environmental and identical on both trees: 17 need `awsiot` (declared in the
  `[mesh-iot]` extra) and 1 is a leftover dataset directory in the shared temp dir.
- The new file passes on **mujoco 3.5.0** (the declared floor) and **3.12.0**; the whole
  `tests/simulation/isaac/` suite is 1431 passed / 5 skipped on 3.12.0.
- Docs guards (`test_docs_python_examples_are_callable`, `test_docs_no_stale_future_symbol_claims`,
  `test_docstrings_no_dangling_doc_refs`, and the source-string sweeps): 482 passed.
- Changelog fragment gates: 89 passed.

## Artifact

Rendered headless in MuJoCo (`MUJOCO_GL=egl`, mujoco 3.5.0) from the shipped assets themselves.
The renders are byte-identical on both trees - the asset does not change, the report does - which
is asserted in the render script, as is every number the figure prints. Panels are composited onto
a flat tone using MuJoCo's own segmentation render and framed from the measured subject bounding
box; the four lower panels share one camera and one crop, since they are a comparison.

The lower half is the stake. `skydio_x2` is stepped 0.30 s under gravity twice: once as MuJoCo
compiles it, and once with no joint for that body - the topology a report carrying no joint
describes. With the free base the body falls `+0.100 m -> -0.320 m` (a displacement of
`-0.420 m`, 1941 changed pixels between the two frames); with no joint, `njnt=0`, `nv=0`,
displacement `0.000 m`, and the two frames are byte-identical. The asset's own `<keyframe>` states
a 7-wide `qpos` that exists only because of that joint, and MuJoCo refuses the keyframe once it is
removed.

MJCF free-joint spellings: reported joint counts against MuJoCo, and a drone stepped under gravity with and without its base joint

### DIFF
--- a/changelog.d/2712-mjcf-freejoint-spelling.md
+++ b/changelog.d/2712-mjcf-freejoint-spelling.md
@@ -1,0 +1,39 @@
+### Fixed: both MJCF spellings of a free joint are read, instead of only one
+
+MJCF states a floating base two ways and MuJoCo compiles both to the same `mjJNT_FREE`:
+`<joint type="free">` and the dedicated `<freejoint>` element. `load_mjcf` walked
+`body_el.findall("joint")`, so only the first spelling produced a `JointDef` and a `<freejoint>`
+produced none at all - the base's six degrees of freedom were absent from the returned
+`ProceduralRobot` rather than reported as a joint a caller can see and skip. The scene reader in
+the same module already consulted both spellings, and the module's own docstring already named
+them as equivalent, so the robot reader was the one surface that knew about only one.
+
+`<freejoint>` is the spelling the shipped asset corpus uses. Of the 46 loadable robot MJCFs under
+`robot_descriptions` that declare a free joint, every one spells it `<freejoint>` and none spells
+`type="free"`, so a floating base was reported for none of them: the quadrupeds (`unitree_go2`,
+`anymal_c`, `spot`), the humanoids (`unitree_h1`, `talos`, `berkeley_humanoid`), and most visibly
+the aircraft - `skydio_x2` and `bitcraze_crazyflie_2` have exactly one joint each, that base, so
+the loader reported a robot with no joints whatsoever.
+
+Both tags are now read, in document order. MuJoCo fixes the two rules that makes this
+unambiguous, and each is pinned as a premise. A free joint may not share a body with any other
+joint (`more than 6 dofs in body`), so a body carrying a `<freejoint>` carries nothing else and
+the order of the two tags can never differ - which also means the change cannot reach the
+`_validate_kinematic_tree` compound-joint guard. And `<freejoint>` resolves no default class:
+MJCF has no `<default><freejoint>` block, and MuJoCo gives such a joint the built-in damping and
+armature even where a `<default><joint>` class is in force, while the `type="free"` spelling does
+inherit that class. The loader now diverges in exactly that place and in exactly that direction.
+
+Graded across the corpus by rewriting every `<freejoint .../>` as `<joint ... type="free"/>` and
+asking MuJoCo whether that is the same model - it is, for 48 of 48 files, with no premise
+violated - the two spellings agreed on 0 of 43 gradable files before and 43 of 43 now, with 43
+joints under-reported before and 0 now. 20 of the 43 reports are byte-identical field for field;
+the other 23 differ only in fields a `<default><joint>` class supplies, and MuJoCo diverges the
+same way on all 23. Nothing else changes: the load and refuse verdicts are the same on both
+readings.
+
+What `free` MAPS to is deliberately unchanged and pinned. `_MJCF_JOINT_TYPE_MAP` reports it as
+`"fixed"` because `JointDef` has no 6-DOF spelling, so a floating base is now visible in `joints`
+without being counted as an actuated DOF by `num_joints`. Whether the dataclass should gain such
+a spelling is a contract question about every producer of a `ProceduralRobot`, so moving it later
+has to be an explicit decision rather than a side effect of this one.
--- a/docs/simulation/isaac.md
+++ b/docs/simulation/isaac.md
@@ -228,7 +228,16 @@ open a window.
   parser can read: a URDF joint that omits the optional `<axis>` acts about
   **+X**, an MJCF `<joint>` that omits `axis` about **+Z**. Both are valid
   axes, so a joint read under the other format's default would be reported
-  acting in the perpendicular plane with the load still reporting success.
+  acting in the perpendicular plane with the load still reporting success. Both
+  of MJCF's spellings of a free joint are read - the dedicated `<freejoint>`
+  element and `<joint type="free">`, which MuJoCo compiles to the same joint -
+  so a floating base is reported rather than absent. `<freejoint>` is how every
+  shipped quadruped and humanoid states its base, and it resolves no default
+  class, because MJCF has no `<default><freejoint>` block: a `<default><joint>`
+  class reaches the `type="free"` spelling only, exactly as MuJoCo applies it.
+  Either spelling is reported with `joint_type="fixed"`, since `JointDef` has no
+  6-DOF spelling, so a floating base is visible in `joints` without being
+  counted as an actuated DOF by `num_joints`.
 
 Because the joint-name and observation contract matches the MuJoCo backend,
 policies and observation mappings transfer unchanged between backends.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -446,6 +446,19 @@ _MJCF_JOINT_TYPE_MAP = {
     "free": "fixed",
 }
 
+# MJCF spells a free joint two ways, and MuJoCo compiles both to the same
+# ``mjJNT_FREE``: ``<joint type="free">`` and the dedicated ``<freejoint>``
+# element. A reader that consults only the first reports no joint at all for a
+# floating base, which is how every quadruped and humanoid in the shipped asset
+# corpus declares its root. Both tags are read here so the two spellings of one
+# model produce one report.
+_MJCF_JOINT_TAGS = ("joint", "freejoint")
+
+# The tag whose type MJCF fixes rather than declaring: ``<freejoint>`` accepts
+# only ``name``, ``group`` and ``align`` - MuJoCo refuses ``type`` on it - and
+# MJCF has no ``<default><freejoint>`` block, so it resolves no default class.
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Fixes `load_mjcf` in the Isaac loader reading only the `<joint type="free">` spelling of a free joint: a body declaring its floating base with the dedicated `<freejoint>` element (the spelling every shipped quadruped/humanoid MJCF uses) produced no `JointDef` at all, so the base's six DOF were silently absent from the report. The walk now iterates both tags in document order, and a `<freejoint>` correctly resolves no default class (MJCF has no `<default><freejoint>` block), while the `type="free"` spelling keeps inheriting its `<default><joint>` class - both mirroring MuJoCo's own compiler behaviour. The `free -> "fixed"` mapping in `_MJCF_JOINT_TYPE_MAP` is deliberately unchanged and pinned, so `num_joints` still excludes the base from the actuated count and the contract question of a 6-DOF `JointDef` spelling is left explicit for a later change.

## What's good

- Every expectation in the new suite is derived from `mujoco.MjModel` rather than restated by hand, including the two premises the implementation rests on (a free joint may not share a body; a `<freejoint>` resolves no class), so a MuJoCo behaviour change would fail the premise tests rather than silently invalidate the loader.
- Enum comparisons are value-based (`int(...) == int(mujoco.mjtJoint.mjJNT_FREE)`) per AGENTS.md > "MuJoCo enums are matched by value".
- Controls pin that `<joint>` handling (class resolution, type vocabulary, tree order, the duplicate-edge guard) is undisturbed, and the deliberately-unchanged mapping is pinned so a later answer has to move it explicitly.
- The sibling reader `load_mjcf_scene_objects` already handled both spellings, and `joint_names.py` is URDF-only, so this closes the last single-spelling read in the module.
- The changelog fragment (`changelog.d/2712-mjcf-freejoint-spelling.md`) is present per AGENTS.md step 3, ASCII-clean, and carries the corpus measurement (43 of 43 files now agree across spellings; 0 under-reported joints).

Verified locally: all 30 new tests pass at the fix commit (`718c2d03`); the only later commit adds the changelog fragment and touches no code. The 3 failures in `tests/simulation/isaac/` reproduce identically on the base commit (an isaacsim-installed-environment stdin issue, unrelated to this diff); `ruff check` and `ruff format --check` are clean on both changed Python files.

yinsong1986 (approved):
## Summary

Narrow, well-evidenced fix: `load_mjcf` in the Isaac loader read only `<joint type="free">`, so a body declaring its floating base with the dedicated `<freejoint>` element - the spelling every free-jointed robot in the shipped `robot_descriptions` corpus uses - produced no `JointDef` at all, leaving the base's six DOF silently absent (and, for the aircraft assets, a robot with zero joints under a successful load). The walk now iterates both tags in document order via a single child filter, and the two divergences from the `<joint>` path are exactly MuJoCo's own: a `<freejoint>` resolves no `<default>` class (MJCF has no `<default><freejoint>` block) and its type is fixed rather than declared. The pre-existing `<joint>` handling is untouched - the old body of the loop is re-indented into the `else` branch byte for byte - and the `free -> "fixed"` mapping in `_MJCF_JOINT_TYPE_MAP` is deliberately unchanged and pinned, so the floating base becomes visible in `joints` without entering `num_joints`, and promoting `JointDef` to a 6-DOF vocabulary stays an explicit later decision. The new suite derives every expectation from `mujoco.MjModel` rather than restating it, pins both premises the implementation rests on, compares reports across all nine `JointDef` fields so nothing hides in an unnamed field, uses value-based enum comparison per AGENTS.md, and keeps controls for class resolution, tree order, the unknown-type refusal and the duplicate-edge guard. Changelog fragment present, docs updated in the section that already discusses per-format joint defaults. No blockers found.

PR #2713, opened 2026-08-25
fix(tests,examples): read a lazily exported tool name off its submodule

### LINKED ISSUE #2711: bug(tests): two tests read a tool name off `strands_robots.tools` and treat it as a module - both pass only because an unrelated import happens to run first
## Problem

`strands_robots/tools/__init__.py` maps each tool name to the `@tool` object living inside the submodule of the *same* name, and caches the result in the package `__dict__`:

```python
"lerobot_train": (".lerobot_train", "lerobot_train"),   # 21 entries share this shape
```

So `from strands_robots.tools import <name>` has **two** possible resolutions, decided by which import ran first anywhere in the process:

| first import to win | the from-import yields | `.__file__` / `inspect.getfile` |
|---|---|---|
| the submodule (`import strands_robots.tools.<name>`) | the **module** | works |
| the lazy package attribute | the **`DecoratedFunctionTool`** | **raises** |

CPython's `_handle_fromlist` only imports the submodule when the attribute is *absent*; here `hasattr` triggers `__getattr__`, which succeeds, so the submodule is never imported and the name binds to the tool object. A `DecoratedFunctionTool` has no `__file__` and no attribute named after itself.

Two tests read a lazily-mapped name off the package and then treat it as a module. Both work today, and both are one unrelated edit away from failing.

## Site 1 - `tests/tools/test_rosbridge_transport_port_limit.py:34` -> `:159`

```python
from strands_robots.tools import use_rosbridge as ur   # line 34
...
root = Path(ur.__file__).resolve().parent.parent        # line 159
```

Measured, clean interpreter, that file's imports **minus the line above**:

```
=> ur is: DecoratedFunctionTool
BROKE: AttributeError 'DecoratedFunctionTool' object has no attribute '__file__'
```

It passes today only because of the import on **line 33**, which transitively imports the submodule and so pre-binds the module:

```
after line 33, submodule imported: True
=> ur is: module
```

So the whole file depends on `from strands_robots.mesh.rosbridge_robot import RosbridgeRobot` staying above it *and* continuing to reach `strands_robots.tools.use_rosbridge` transitively. Neither is stated anywhere, and reordering the import block - something a formatter or an unrelated cleanup does - breaks 21 tests.

## Site 2 - `tests/tools/test_output_path_is_confined_to_its_directory.py:310` -> `:314`

```python
from strands_robots.tools import pose_tool as pose_mod          # line 310, inside a test body
result = pose_mod.pose_tool(action="list_poses", robot_id="../escaped")  # line 314
```

Measured, clean interpreter:

```
pose_mod is: DecoratedFunctionTool
BROKE: AttributeError 'DecoratedFunctionTool' object has no attribute 'pose_tool'
```

This import runs at **test-run** time, so it is masked by two *earlier tests in the same file* (`:293`, `:299`) doing `from strands_robots.tools.pose_tool import PoseManager`:

```
pose_mod is: module | .pose_tool resolves: True
```

That makes it order-dependent within a single file: any selection that runs the line-310 test without one of those two first - a `-k` filter, a shuffled order, running that test alone - resolves the tool object and raises `AttributeError`. The failure would look like a missing attribute rather than an import-order artifact.

## Why file this now

This is the same defect class that produced a red CI on #2708, where the third occurrence sat in a new test:

```
TypeError: module, class, method, function, traceback, frame, or code object
was expected, got DecoratedFunctionTool
```

There it passed the narrow selection it was measured against and failed the full suite, because the full suite touched the lazy attribute first. It is fixed on that branch by deriving the file from the *function* under test rather than resolving a name - the idiom that file already uses elsewhere: ask the object, never resolve a name.

The two sites above are **pre-existing on `main`** and are not that branch's contract, so they were deliberately left out of its diff rather than widening an approved change.

## Suggested remedy

For both sites, use the unambiguous submodule form, which ~150 other test sites already use and which is immune by construction:

```python
import strands_robots.tools.use_rosbridge as ur          # site 1
from strands_robots.tools.pose_tool import pose_tool     # site 2
```

Worth considering alongside: a guard that refuses `from strands_robots.tools import <lazily-mapped-name>` in tests that then use the result as a module. The existing lint at `tests/tools/test_tool_actions_block_names_the_dispatched_vocabulary.py:357-427` cannot see either site - it scans only its own module. Whether that guard earns its keep at 2 occurrences is a judgement call; the two import fixes stand on their own either way.

## Scope

Sweep covered `tests/`, `tests_integ/` and `scripts/`. 21 lazily-mapped names, every package-level read of one of them enumerated; these two are the only sites that treat the result as a module. `tests_integ/` and `scripts/` have none. No behaviour change to shipped code is implied - this is test-harness correctness only.

cagataycali:
## Problem

`strands_robots/tools/__init__.py` maps each of its 21 exported names to the `@tool` object inside the submodule of the *same* name, and caches the result in the package `__dict__`:

```python
"pose_tool": (".pose_tool", "pose_tool"),   # 16 of the 21 names are also submodules
```

So the package attribute for such a name has two possible values, and `from strands_robots.tools import <name>` binds whichever one the process reached first. Measured, one fresh interpreter per cell:

| spelling | cold | tool cached first | submodule first |
|---|---|---|---|
| `from strands_robots.tools import X` | tool | tool | **module** |
| `import strands_robots.tools.X as x` | module | **tool** | module |
| `importlib.import_module("...tools.X")` | module | module | module |
| `from strands_robots.tools.X import X` | tool | tool | tool |

Four sites read the top row, in **both** failure directions, and each passed in the selection it was written against:

| site | wanted | got in the other order |
|---|---|---|
| `tests/tools/test_rosbridge_transport_port_limit.py:34` | the module (`ur.__file__`, `ur._ACTIONS`, `ur._backend`, ...) | `'DecoratedFunctionTool' object has no attribute '__file__'` |
| `tests/tools/test_output_path_is_confined_to_its_directory.py:310` | the module (`pose_mod.pose_tool`) | `'DecoratedFunctionTool' object has no attribute 'pose_tool'` |
| `examples/08_discover_lerobot.py:16` | the tool (`use_lerobot.__wrapped__`) | `module ... has no attribute '__wrapped__'` |
| `examples/libero/run_isaac_agent.py:188` | the tool (called, and registered with an `Agent`) | a module where a tool is expected |

The second one reproduces on `main` with no edit, because its masking comes from two earlier tests in the same file importing the submodule:

```
$ pytest tests/tools/test_output_path_is_confined_to_its_directory.py -k reports_the_refusal_in_its_own_envelope
>       result = pose_mod.pose_tool(action="list_poses", robot_id="../escaped")
E       AttributeError: 'DecoratedFunctionTool' object has no attribute 'pose_tool'
1 failed, 31 deselected
```

## Why this is wider than four sites

The banned read is the only spelling that writes the **tool** into that slot - a submodule import writes the module, and the two bottom rows never touch it. So one such read anywhere in the process is what makes the module-alias form order-dependent, and that form is what this tree uses widely as a monkeypatch target:

```python
from strands_robots.tools import use_rosbridge as _anywhere   # one read, at import time
import strands_robots.tools.use_rosbridge as rb_mod           # the patch-target idiom
rb_mod.roslibpy
# AttributeError: 'DecoratedFunctionTool' object has no attribute 'roslibpy'
```

The failure names the tool class, not an import order. Eliminating the ambiguous read tree-wide is what keeps that idiom sound, which is why the guard bans one spelling rather than grading what each site does with the result.

## Change

Each site is respelled to the remedy matching its intent, and both remedies read the submodule so no import order changes them: the module-alias form where the module object is wanted, `from strands_robots.tools.<name> import <name>` where the tool is wanted. `test_rosbridge_transport_port_limit` no longer depends on the `strands_robots.mesh.rosbridge_robot` import staying above it.

`tests/tools/test_lazy_tool_name_imports_are_unambiguous.py` bans the read. Both halves are derived, not listed: the shadowable names come from `_LAZY_IMPORTS` intersected with the submodules that exist, and the scanned areas from the repository's own top-level directories that ship Python. A name with no matching submodule (the five `episode_judge` helpers) has one possible value and is deliberately in scope for nobody. AGENTS.md records the convention under Testing Patterns.

The complementary spelling - a plain `tools.X` attribute read - is already graded, within its own module, by `test_tool_actions_block_names_the_dispatched_vocabulary.py`.

## Verification

Pre-fix, the guard names all four sites in one failure. Break table, 12 mutations and one run of the guard file each, collapsed below where several mutations share a firing set:

| mutation | guard failures | which |
|---|---|---|
| control | 0 | - |
| any one of the four sites reverted | 1 | the tree scan, naming that site |
| all four reverted | 1 | the tree scan, naming all four |
| site 1 fixed with `import_module` instead | 0 | the other immune form is accepted too |
| every lazy name treated as shadowable | 2 | over-reach: the not-a-submodule control, and the derivation premise |
| the ban widened to the submodules too | 2 | over-reach: flags 72 legitimate reads of a tool off its own submodule |
| the package no longer caches the resolved tool | 2 | both premise tests, which is where that cache is pinned |
| areas hardcoded and narrowed, both examples reverted | 1 | the reach test, so the derivation is load-bearing rather than decorative |
| the AGENTS.md bullet removed | 0 | nothing grades the guidance text; stated rather than implied |

Gate, on the merged tree:

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1535 files).
- mypy **24 == 24** against a pristine `main` worktree, normalized message sets identical both ways, 0 in changed files.
- Paired run of an identical 490-file selection (all of `tests/tools`, all of `tests/mesh`, and every test that walks the tree, reads `examples/`, reads AGENTS.md or reads `_LAZY_IMPORTS`), with the new file excluded from both trees: **21 failed, 18687 passed, 76 skipped, 24 errors on both**, 45 failing/error ids each and both `comm` directions empty. 44 of the 45 need `awsiot` (the `[mesh-iot]` extra); the remaining one is pre-existing on `main`.
- After merging the new base: `tests/tools/` plus the three test files `main` added and the export/lazy-import guards - **3674 passed, 9 skipped**.
- The changed files pass on mujoco 3.5.0 / lerobot 0.6.2 and on mujoco 3.12.0 / lerobot 0.5.1 (66 passed, 3 skipped).

No visual artifact: the change touches tests, examples and a convention document, with no policy, simulation, rendering, recording or asset behaviour in it. The two resolution tables above are measured, one interpreter per cell.

Closes #2711.


yinsong1986 (approved):
## Summary

Fixes an import-order ambiguity: `strands_robots.tools` lazily maps each exported name to the `@tool` object inside the same-named submodule and caches it in the package `__dict__`, so `from strands_robots.tools import <name>` binds either the tool or the module depending on which the process imported first. Four sites read that spelling (two tests wanting the module, two examples wanting the tool); each is respelled to the order-immune form matching its intent, and a new derived guard (`tests/tools/test_lazy_tool_name_imports_are_unambiguous.py`) bans the ambiguous read tree-wide. Independently verified on the head commit: the guard's scan reports zero offenders and zero parse errors across all five first-party areas (`examples`, `scripts`, `strands_robots`, `tests`, `tests_integ`), the shadowable set derives to 16 of the 21 lazy names as claimed, both remedies resolve to the expected object type, and the three touched test files pass (66 passed, 3 skipped) with ruff check/format clean.

## What's good

- Both halves of the guard are derived rather than listed (`_LAZY_IMPORTS` x existing submodules for names; repo top-level dirs shipping Python for areas), with non-vacuity tests pinning the derivation so a mis-rooted scan or an over-wide ban fails loudly.
- The premise tests (`TestTheAmbiguityIsReal`) run one fresh interpreter per cell, so the order-dependent resolution is actually observable rather than asserted from a warm suite.
- Changelog fragment under `changelog.d/` per AGENTS.md step 3; the new convention is recorded in AGENTS.md > Testing Patterns with the pinning test named.
- New test file is ASCII-clean, host-path-free, and names its behaviour rather than its provenance.

PR #2714, opened 2026-08-25
test(tools): import a tool from its submodule, not off the lazy package

### LINKED ISSUE #2711: bug(tests): two tests read a tool name off `strands_robots.tools` and treat it as a module - both pass only because an unrelated import happens to run first

cagataycali:
## Problem

`strands_robots/tools/__init__.py` maps each tool name to the `@tool` object living inside the submodule of the *same* name, and caches the result into the package namespace:

```python
"pose_tool": (".pose_tool", "pose_tool"),   # 16 of the 21 mapped names share this shape
```

So `from strands_robots.tools import pose_tool` has **two** possible resolutions, and which one wins is decided by whichever import ran first anywhere in the process. CPython's `_handle_fromlist` imports the submodule only when the attribute is *absent*; here the lookup triggers the package `__getattr__`, which succeeds, so the submodule is never imported and the name binds to the tool object instead.

The two are not interchangeable. Measured in a clean interpreter, with the submodule import present and absent:

| first import to win | the from-import yields | `.__file__` |
|---|---|---|
| the submodule | `module` | `use_rosbridge.py` |
| the package attribute | `DecoratedFunctionTool` | `AttributeError` |

Two tests bind a name this way and then read the result as a module, so they pass on the import order of the whole process rather than on the behaviour they verify.

### Site 1 - `tests/tools/test_rosbridge_transport_port_limit.py`

`from strands_robots.tools import use_rosbridge as ur` was masked by the import on the line **above** it, which reaches `strands_robots.tools.use_rosbridge` transitively (`mesh/rosbridge_robot.py:38` imports `_HOST_RE`, `_transport_port_error` and `use_rosbridge` from it). The dependency is on two alphabetically-adjacent lines keeping their order, and is stated nowhere. Swapping them is not a single test failure:

```
AttributeError: 'DecoratedFunctionTool' object has no attribute '_ACTIONS'
!!!! Interrupted: 1 error during collection !!!!
```

That is a **collection error, losing all 24 tests in the file**. The file makes 7 such reads, not one: `ur._ACTIONS`, `ur._backend` (x2), `ur._TRANSPORT_MAX_PORT`, `ur._transport_port_error` (x2) and `ur.__file__`. The load-bearing coupling is a **private** import in shipped code, and `_transport_port_error` currently has exactly two callers - the condition under which this project promotes a guard to a shared owner, and `tools/_numeric_options.py` is already the documented home for exactly these three transports. That routine refactor would break a file that never mentions it.

### Site 2 - `tests/tools/test_output_path_is_confined_to_its_directory.py`

`from strands_robots.tools import pose_tool as pose_mod` runs at **test-run** time, masked by two earlier tests in the same file importing `PoseManager` from the submodule. So selecting that one test - a `-k` filter, a shuffled order - already fails on `main`:

```
>       result = pose_mod.pose_tool(action="list_poses", robot_id="../escaped")
E       AttributeError: 'DecoratedFunctionTool' object has no attribute 'pose_tool'
1 failed, 31 deselected
```

The failure names the attribute rather than the import that decided it, so it reads as a missing attribute rather than an import-order artifact.

## Change

Both sites import the submodule directly, which cannot resolve to anything else:

```python
import strands_robots.tools.use_rosbridge as ur          # site 1
from strands_robots.tools.pose_tool import pose_tool     # site 2
```

`tests/tools/test_lazy_tool_name_is_not_read_as_a_module.py` then refuses the shape over `strands_robots/`, `tests/`, `tests_integ/` and `scripts/`. Both halves of the rule are derived rather than listed, so a tool added to the mapping and a call site added to any scanned tree are graded on arrival:

- **which names are ambiguous** - a mapped name that is also a submodule (16 of 21);
- **which attributes only a module can answer** - asked of the tool object itself (`__file__`, `__spec__`, `__loader__`, `__package__`, `__cached__`, `__builtins__`, plus the module's own private names). An attribute both resolutions carry is deliberately accepted: `tool_name` and `__name__` mean the same thing either way, so the rule is "the tool cannot answer it", not "it is a dunder".

Production code is scanned too - the ambiguity is a property of the import, not of the caller - and currently has zero such imports, so that root is forward-looking.

One thing worth flagging about the guard's own design. There are exactly **2** package-level from-imports of an ambiguous name in the whole repository, and both are the sites fixed here, so after the fix the sweep has nothing left to classify and is a regression tripwire with no grading power of its own. The grading therefore lives in `TestTheRuleFlagsTheReadItIsFor`, which runs the same predicate over five constructed sources covering both outcomes - including the two unambiguous import forms, pinned as *accepted* so the rule cannot be satisfied by refusing them too. The file says this in its own docstring rather than leaving a reader to assume the sweep is doing the work.

`AGENTS.md` gains the convention beside the existing `sys.modules` import-state rule, so the reason is discoverable and not only enforced.

## Tests

Pre-fix, the guard reports every read and only the tripwire fails - the premise, over-reach and constructed cases all pass:

```
1 failed, 10 passed
E  ... 'tests/tools/test_rosbridge_transport_port_limit.py': [(77, 'ur._ACTIONS'), (99, 'ur._backend'),
        (100, 'ur._backend'), (144, 'ur._TRANSPORT_MAX_PORT'), (147, 'ur._transport_port_error'),
        (148, 'ur._transport_port_error'), (159, 'ur.__file__')]
   ... 'tests/tools/test_output_path_is_confined_to_its_directory.py': [(313, 'pose_mod.pose_tool')]
```

Post-fix: guard `11 passed`; site 2 alone via `-k` `1 passed` (was `1 failed`); both touched files `56 passed`. The decisive one - reordering site 1's imports so nothing pre-binds the submodule - is `24 passed`, where the same reordering pre-fix was a collection error.

Nine mutations, seven distinct firing sets, control clean:

| mutation | fails | what fires |
|---|---|---|
| control | 0 | - |
| revert site 1's import only | 1 | the sweep |
| revert site 2's import only | 1 | the sweep |
| resolve the tool via the ambiguous spelling | 3 | both premise tests + the dunder-read case |
| flag any attribute (drop the tool-can-answer filter) | 1 | the shared-attribute case (over-reach) |
| hardcode the module-only set instead of deriving it | 1 | the premise |
| stop scanning `tests/` | 1 | the reaches-its-trees premise |
| treat every mapped name as ambiguous | 0 | the 5 non-shadowable names live in `episode_judge`, so no submodule can shadow them and nothing reads one as a module; the filter is scope precision, not correctness |
| stop scanning production code | 0 | production has no such import today; that root is forward-looking |

## Gate

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`; `mypy` reports `Success: no issues found in 1532 source files`.

Paired against a pristine `main` worktree over an identical 348-file selection (all of `tests/tools/` plus every test that walks the tree with `ast`/`rglob`/`inspect`), with the new file excluded from both: **17116 collected and `2 failed, 17047 passed, 67 skipped` on both trees**, failing-ID sets identical, `comm` empty in both directions. Both shared failures are `docker not found on PATH`, confirmed environmental.

No visual artifact applies: nothing in the shipped library changes, and no policy, simulation, rendering, recording or asset behaviour is touched.

Closes #2711.


yinsong1986 (approved):
## Summary

Test-only change closing #2711: two test files bound a lazily-mapped tool name off `strands_robots.tools` (`from strands_robots.tools import use_rosbridge as ur` / `import pose_tool as pose_mod`) and then read the result as a module, so they passed or failed on process-wide import order rather than on the behaviour under test. Both sites now import the submodule directly, which has exactly one resolution, and a new derived guard (`tests/tools/test_lazy_tool_name_is_not_read_as_a_module.py`) refuses the shape across `strands_robots/`, `tests/`, `tests_integ/` and `scripts/`. No shipped behaviour changes - the diff to `strands_robots/` is empty.

## What's good

- Both halves of the guard are derived rather than listed (ambiguous names from `_LAZY_IMPORTS` intersected with real submodules; module-only attributes asked of the tool object itself), so a tool added to the mapping is graded on arrival - the project's preferred shape for tree-wide rules.
- The guard is honest about its own limits: with both offending reads fixed, the sweep has nothing left to classify, so `TestTheRuleFlagsTheReadItIsFor` keeps the predicate measured over constructed sources covering both outcomes, and the two unambiguous import forms are pinned as *accepted* so the rule cannot be satisfied by over-refusing. The shared-attribute case (`tool_name`, `__name__`) is deliberately not flagged, which keeps the rule precise.
- Changelog fragment (`changelog.d/2714-...`) instead of a direct `CHANGELOG.md` edit, closing keyword in the PR body, and the AGENTS.md convention lands beside the adjacent `sys.modules` import-state rule with the enforcing test named.

Verified locally at head `38941c4a`: the guard plus both touched test files pass (64 passed, 3 skipped), the previously order-dependent site-2 test passes under a `-k` filter in isolation, changelog format tests pass, and the new files are ASCII-clean with no host paths.

PR #2715, opened 2026-08-25
fix(simulation): hold the terrain seed to the shared stream-name domain

cagataycali:
## Problem

`strands_robots.simulation.terrain.generate_heightfield` documents itself as "deterministic given `(kind, resolution, seed)`", and the module docstring promises that a benchmark evaluating a policy on `terrain="rough"` "regenerates the identical field on every reset". `resolution` is measured against the shared positive-discrete domain before anything is generated. `seed` was handed straight to `random.Random`.

`random.Random` does not seed from the value it is given. CPython's `random_seed` uses `abs(value)` for an int and `hash(value)` for anything else. So the documented triple was neither injective nor total:

| `seed` | before | why |
| --- | --- | --- |
| `1` | field `ebb516ac` | the field the caller asked for |
| `-1` | field `ebb516ac` | the sign is discarded |
| `True` | field `ebb516ac` | `bool` is an `int` subclass |
| `float("nan")` | a **different** field on each call | `hash(nan)` is identity-derived since Python 3.10 |
| `None` | a **different** field on each call | seeds from the OS entropy pool |
| `2.5`, `"1"`, `inf` | some field | accepted outright, hashed |

Two consequences, not one. Three distinct seeds named **one** field, so a terrain curriculum stepping the seed across resets to draw fresh ground silently re-drew ground it had already evaluated on and nothing reported that the two resets shared terrain. And `nan`/`None` made the field **irreproducible** - the inputs for which the module's headline promise is false rather than merely surprising. Measured on `main`: `seed=None` on three consecutive calls gives `c3bf0fcd`, `54239b82`, `6d6ce503`.

This is the same collision class `derive_variant_seed` already spells out at length for the other seed in the package that is spread into a stream key ("`True` is the value worth naming, since unrefused it is `1` to NumPy"). `seed` was the one seed parameter in the library with no domain: the simulation backends apply `randomization_seed_error` at eight call sites, `transforms` applies `non_negative_whole_number_error`, the GR00T server wrapper and the trainers each apply their own.

## Change

Measure `seed` against `non_negative_whole_number_error`, the shared domain `derive_variant_seed` applies to the same quantity, on the branch that reads it. Each clause of that domain closes one axis:

- **non-negative** closes the `abs()` alias (`-1` -> `1`),
- **whole** closes the `hash()` path, which is what admitted `nan`, `inf`, `2.5` and `"1"`,
- the domain's **boolean** refusal closes `True`/`False`,
- and `None` is refused, so an opt-out has to be spelled deliberately rather than inherited.

An integral float such as `2.0` is still accepted, because it hashes to its int and so names the *same* stream rather than a second one - the complement is pinned, so the domain reads as "the set of spellings that name exactly one stream" rather than as blanket strictness.

## Visual artifact

The generated field laid down as the MuJoCo `<hfield>` the module's own constants describe. Identical geometry, camera, lighting and material in all ten panels - the generated field is the only variable. Rendered headless (`MUJOCO_GL=egl`).

terrain seed stream domain

Every panel in the composed PNG was verified byte-identical to the render it claims, by searching for its paste offset and comparing the arrays. `seed=1`, `seed=-1` and `seed=True` render byte-identically on `main`; the two `nan` draws differ in 48990 of 69000 pixels; and `seed=1` on this branch renders byte-identically to `seed=1` on `main`, so nothing legitimate moved.

## Scope, and why each boundary is where it is

**Only the `"rough"` branch is guarded.** `_stairs`, `_pyramid` and `_slope` take no seed argument at all, so refusing them for one would refuse a caller a value the requested kind cannot act on - the same rule `NewtonSimEngine.open_viewer` follows for `width`/`height` ("only the gl branch reads them"). The split is derived from the generators' signatures rather than hardcoded, so a kind that later gains an rng lands in the guarded bucket instead of inheriting the exemption. Mutating `_stairs` to take a seed fires exactly that test.

**No upper bound is added.** `random.Random` consumes an arbitrarily large int directly (`Random(10**30)` draws fine), so the narrower seed domain the backends use - `randomization_seed_error`, capped to NumPy's 32-bit stream range - would refuse a value that is usable here, and it lives in a module `terrain` deliberately does not import. Adding a `2**32` ceiling fires exactly one test.

**The kind is still answered first.** An unknown kind must not be reported as a seed problem; moving the guard ahead of `validate_terrain` fires that test plus all 30 exemption tests.

## Tests

New `tests/simulation/test_terrain_seed_stream_domain.py`, 59 cases in four classes: the regression, the premise (the `random.Random` behaviour that makes each refused value an alias for an accepted one), the controls, and the scope boundary.

Pre-fix, with the source change reverted and the test kept: **11 failed, 48 passed** - and all 11 are in the regression class, none in the premise, control or boundary classes. The headline failure is `test_the_seeded_kind_refuses_it[nan] - DID NOT RAISE ValueError`.

Break table, 10 mutations, control clean, 9 distinct firing sets:

| mutation | fires | reads |
| --- | --- | --- |
| guard removed entirely | 11 | the pre-fix tree |
| domain swapped to the positive sibling | 4 | refuses `0`, which is `TERRAIN_SEED` |
| hand-rolled `isinstance(seed, int)` | 6 | accepts `True` and `-1`, refuses `2.0` |
| domain applied to every kind | 30 | over-reach; exactly the exemption class |
| guard moved ahead of the kind check | 31 | the 30 above plus the ordering test |
| rule re-spelled locally | 12 | single-sourcing is load-bearing |
| value coerced with `abs(int(...))` instead of refused | 11 | coercing is not reporting |
| an upper bound added | 1 | the huge-int control |
| `_stairs` gains a seed parameter | 1 | the closure test |

`11` (guard removed) and `30` (over-reach) are disjoint, so the fix and its scope are pinned independently; `31` is the union of the over-reach set with the ordering test.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1535 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `main` worktree, normalized message sets byte-identical, 0 in the changed files.
- Paired run over an identical 670-file selection (all of `tests/simulation/`, every test that walks the tree or reads docstrings, the docstring-xref and `Args`-completeness guards), the new file excluded from both trees: **25069 collected and `20 failed, 24769 passed, 280 skipped` on both**, 20 identical failing IDs, both `comm` directions empty. All 20 are pre-existing: 17 need `awsiot`/`awscrt` (absent; declared in the `[mesh-iot]` extra), 2 are lerobot-0.6.2-from-source drift in `tests/training/test_lerobot.py`, 1 is a stale dataset directory.
- The new file plus `tests/simulation/test_terrain.py`: **129 passed** on mujoco 3.5.0 (the declared floor) and on 3.12.0.
- Changelog-fragment guards: 108 passed.

### DIFF
--- a/changelog.d/2715-terrain-seed-stream-domain.md
+++ b/changelog.d/2715-terrain-seed-stream-domain.md
@@ -1,0 +1,39 @@
+### Fixed: the terrain seed names exactly one value-noise stream
+
+`generate_heightfield` documents itself as "deterministic given `(kind, resolution, seed)`", and
+the module docstring promises that a benchmark evaluating a policy on `terrain="rough"`
+"regenerates the identical field on every reset". `resolution` was measured against the shared
+positive-discrete domain; `seed` was handed straight to `random.Random`, which does not seed from
+the value it is given - it seeds an int from `abs(value)` and anything else from `hash(value)`. So
+the documented triple was neither injective nor total.
+
+Three distinct seeds drew one field. `seed=-1` shares a stream with `seed=1` because the sign is
+discarded, and `seed=True` shares it too because `bool` is an `int` subclass - the same collision
+`derive_variant_seed` already names for the other seed in the package that is spread into a stream
+key. A curriculum stepping the seed across resets to draw fresh ground silently re-drew ground it
+had already evaluated on, and nothing reported that the two resets shared terrain.
+
+`seed=float("nan")` was worse than ambiguous: it was irreproducible. `hash(nan)` has been derived
+from the object's identity since Python 3.10, so two `float("nan")` seeds draw two different fields
+*within a single process* - the one input for which the module's headline promise is false rather
+than merely surprising. `seed=None` was accepted too, and seeds `random.Random` from the OS entropy
+pool, so it drew a fresh field on every call. `seed=2.5` and `seed="1"` were accepted outright, the
+same fractional and string axes the `resolution` domain already closed.
+
+The seed is now measured against `non_negative_whole_number_error`, the shared domain
+`derive_variant_seed` applies to the same quantity. Non-negative closes the `abs()` alias, whole
+closes the `hash()` path that admitted `nan`, `inf`, `2.5` and `"1"`, and the domain's boolean
+refusal closes `True`/`False`. An integral float such as `2.0` is still accepted, because it hashes
+to its int and so names the same stream rather than a second one. Measured over the production
+resolution: before, `seed=1`, `seed=-1` and `seed=True` all produce field md5 `ebb516ac` and render
+byte-identically; two `nan` seeds produce `07b782df` and `40ab6cf9` and differ in 48990 of 69000
+render pixels. After, `seed=1` still produces `ebb516ac` and renders byte-identically, and the
+ambiguous spellings raise `ValueError`.
+
+Scoped two ways, both pinned. The domain is applied only on the `"rough"` branch, the only kind
+drawn from an rng: `_stairs`, `_pyramid` and `_slope` take no seed at all, so refusing them for one
+would refuse a caller a value the requested kind cannot act on. The split is read off the
+generators' signatures rather than hardcoded, so a kind that later gains an rng cannot inherit the
+exemption. And no upper bound is imposed on top of the shared domain, because `random.Random`
+consumes an arbitrarily large int directly - the narrower seed domain the simulation backends apply
+is capped to NumPy's 32-bit stream range and would refuse a value that is usable here.
--- a/strands_robots/simulation/terrain.py
+++ b/strands_robots/simulation/terrain.py
@@ -37,7 +37,11 @@ from __future__ import annotations
 
 import random
 
-from strands_robots.utils import positive_finite_number_error, positive_whole_number_error
+from strands_robots.utils import (
+    non_negative_whole_number_error,
+    positive_finite_number_error,
+    positive_whole_number_error,
+)
 
 # Supported terrain kinds. ``"rough"`` is smoothed value-noise bumps; ``"stairs"``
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes the last unmeasured input on `generate_heightfield`: `seed` is now held to the shared `non_negative_whole_number_error` domain, on the one branch (`"rough"`) that actually reads it. The premise checks out against `random.Random` itself - an int seeds from `abs(value)` and anything else from `hash(value)`, so `-1`/`True`/`1` named one stream and `nan` was irreproducible within a single process, both of which break the module's documented `(kind, resolution, seed)` determinism. Verified against a patched copy: refusals raise `ValueError` (the module's error contract, so the `create_world` implementations' narrowing still catches it), the only production call site (`spec_builder.py`, `seed=TERRAIN_SEED=0`) and the example's default-seed call both pass the new guard, the kind is answered before the seed, and the seed-independent kinds are not refused for a value they never read.

## What's good

- Reuses the existing shared domain rather than minting a new guard (AGENTS.md rule 11 - `non_negative_whole_number_error` already has multiple callers), and the refusal text is single-sourced from it.
- Scope discipline matches the `_numeric_option_error` convention: only the branch that consumes the value measures it, pinned by the unseeded-kind tests.
- The no-upper-bound decision is stated with its reason (`random.Random` consumes an arbitrarily large int; the backends' 32-bit `randomization_seed_error` lives in a module this one deliberately does not import), so the boundary reads as a choice rather than an omission.
- The regression tests pin both scope boundaries (seeded-kind refusal, unseeded-kind exemption, bucket exhaustiveness) plus the language-level premises, so the domain reads as the exact aliasing set rather than arbitrary strictness.

PR #2716, opened 2026-08-25
fix(tools): read a one-component observation.state as the vector it declares

cagataycali:
## Problem

LeRobot stores a feature of shape `[1]` unwrapped: the `observation.state` parquet column holds a bare scalar float where a wider state holds a list, while `meta/info.json` declares `shape [1]` and one joint name either way.

`sample_frames` read every frame's state by iterating that value:

```python
"state": [float(v) for v in state] if state is not None else None,
```

so a single-DOF recording raised `TypeError: 'float' object is not iterable`. That escapes the `@tool` boundary, because `TypeError` is not among the exceptions the tool converts into an error envelope - while the module docstring promises the opposite:

> Every tool returns the structured `{"status", "content"}` envelope and never raises - a judge run over a hundred episodes must report the one episode it could not read, not die on it.

And there was nothing to report: the recording is healthy and its state is the one-element vector its own metadata describes. A single-DOF scene is ordinary (a gripper, a linear stage, a pan unit).

The break also landed before `include_images` was consulted, so the whole multimodal decode path was unreachable for such a dataset.

The two tools disagreed on the same dataset: `load_episode` reported `state_names: ['shoulder_pan']` off the metadata, and `sample_frames` raised on the column that metadata describes.

## Measured, on one recording read by both trees

A one-joint scene, 8 frames, mock rollout, MuJoCo headless.

| | `main` | this branch |
|---|---|---|
| `sample_frames(...)` | raises `TypeError: 'float' object is not iterable` | `status: success` |
| samples returned | 0 | 2, each `state` a 1-vector |
| `max_state_delta` / `rms_state_jerk` | not reached | `0.038159` / `598.4779` |
| image blocks (`include_images=True`) | 0 | 4 |

The dataset's own metadata and storage, measured rather than assumed:

| width | declared `shape` | parquet column holds |
|---|---|---|
| 1 component | `[1]`, names `['shoulder_pan']` | a bare `float` |
| 2 components | `[2]` | a `list` of 2 |

## Change

One accessor, `_state_vector`, owns the column shape: a scalar is read as the one-element vector its metadata declares, an iterable is read component-wise, and `None` stays `None`. A wider state is untouched.

Swallowing the `TypeError` into the error envelope was the alternative and is the wrong one - it reports a healthy recording as unreadable. The break table below shows that variant failing the same 5 tests as no fix at all.

## Tests

`tests/tools/test_episode_judge_single_element_state.py` (11 tests). Pre-fix: **5 failed, 6 passed**, every failure `TypeError: 'float' object is not iterable`, with **0** failures in the premise class and **0** in the wider-recording control.

The premise is measured against the real writer rather than a hand-written fixture, so the scalar column is what LeRobot actually produces; the same rule is then graded hermetically over a hand-built dataset, so pyarrow alone can check it on an install with no simulator.

| mutation | failures | firing set |
|---|---|---|
| control: unmodified | 0 | - |
| the column value is iterated again (pre-fix) | 5 | the whole regression set, incl. the camera decode |
| unwrap only a `list`, not any iterable | 0 | semantic no-op: `to_pydict` yields lists |
| every value wrapped as one element | 2 | both wider-recording controls (over-reach) |
| a scalar read as no state at all | 4 | the state and summary rules |
| a scalar returned unwrapped | 5 | the whole regression set |
| pre-fix plus `TypeError` swallowed into the envelope | 5 | identical to no fix: an error envelope is not a reading |

Coverage of `strands_robots/tools/episode_judge.py` over the full suite: **80% -> 94%** (38 -> 12 uncovered statements). The remaining gap is the `episode_data_index` branch, which the installed lerobot no longer exposes.

## Gate

- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ` clean (1538 files).
- mypy **24 == 24** against a pristine `main` worktree, normalized message sets byte-identical, **0** in the changed files.
- Paired run over an identical 377-file selection (all of `tests/tools/`, every test naming the judge tools, every tree-walking guard), the new file excluded from both trees: **18 failed, 17879 passed, 74 skipped on both**, 18 identical failing ids, both `comm` directions empty. All 18 pre-existing: 17 need `awsiot` (the `[mesh-iot]` extra), 1 is a stale dataset in `/tmp`.
- Full suite, `MUJOCO_GL=egl`: base **29 failed / 37958 passed / 309 skipped / 24 errors** vs branch **29 failed / 37969 passed / 309 skipped / 24 errors** - `+11`, exactly this file's tests, with identical FAILED and ERROR sets.

## Artifact

The same dataset read by both trees. Left: what `main` returns (nothing - it raises, so no camera frame is ever decoded). Right: the four image blocks the tool now returns, position-major with cameras sorted, each panel the exact PNG bytes from the tool's own output.

sample_frames on a one-component state


yinsong1986 (approved):
## Summary

Fixes a crash bug in `sample_frames`: LeRobot stores a shape-`[1]` `observation.state` as a bare scalar in the parquet column, so iterating it raised `TypeError: 'float' object is not iterable` straight through the `@tool` boundary (the `except` tuple does not include `TypeError`), breaking the module's own never-raises envelope contract on a healthy single-DOF recording and making the whole `include_images` decode path unreachable for such datasets. The fix is a single narrow accessor, `_state_vector`, that reads a scalar as the one-element vector the dataset's own `meta/info.json` declares, reads any non-str/bytes iterable component-wise, and passes `None` through - wider states are byte-for-byte unchanged, which the control tests pin.

## What's good

- Scope discipline: one private helper plus one call-site substitution; no public API, tool-spec, wire-format, or persisted-schema change.
- The alternative (swallowing `TypeError` into the error envelope) is explicitly considered and rejected with mutation evidence - it would report a healthy recording as unreadable, exactly the no-silent-defaults failure AGENTS.md forbids.
- The regression test measures the premise against the real LeRobot writer (scalar for width 1, list for width 2) rather than assuming it, then grades the rule hermetically over a hand-built dataset so pyarrow alone can check it; the recording fixtures pass `root=` per AGENTS.md rule 15, class names describe behaviour, and the module-scope `MUJOCO_GL=egl` default is the pattern `tests/test_examples_mujoco_gl.py` documents as ungraded for offscreen backends.
- Changelog fragment present (step 3), docs updated in the same PR, `Args:` entry on the new helper, ASCII-clean throughout, and the gate section reports paired base/branch suite runs with identical failure sets.

PR #2717, opened 2026-08-25
fix(rendering): refuse an environment-map frustum that cannot frame anything

cagataycali:
## The defect

`render_environment_map` takes five numbers. Three of them size the pixel grid and
share one domain, checked before any face is rendered (#2446). The other two --
`znear` and `zfar`, the view frustum all six cube faces are rendered through -- were
checked nowhere.

An inverted frustum produced a map rather than a refusal. `znear=10.0` with
`zfar=1.0` returned a full-size, entirely black `(equi_h, equi_w, 3)` array,
reporting success, after paying all six background renders -- GPU-bound for a
`GsplatBackground`. The only refusal the caller ever saw came from
`derive_key_light`, and it blamed the scene:

```
derive_key_light: no light in the searched region (the map is black above the
horizon -- pass upper_hemisphere=False to search the full sphere, or keep the
caller's default key light).
```

Following that advice fails again (`the map is black`), because the frustum framed
nothing at all rather than leaving a dark upper half. So the remedy offered was a
dead end and neither clip plane was named anywhere. That is the same dead end #2446
removed for the resolution knobs, one axis over: its guard file's headline test is
literally "a resolution mistake must not be reported as a dark scene".

## Why the planes cannot be left to the background

`render_environment_map` accepts *any* `BackgroundRenderer`, and they do not agree
about what an unusable pair means. Measured on one real 3DGS scene, same call, same
resolution, only the background swapped:

| znear, zfar | `PanoramaBackground` | `GsplatBackground` (gsplat 1.4.0, CUDA) |
| --- | --- | --- |
| `0.01`, `1e3` (default) | map, mean 173.096 | map, mean 59.429 |
| `10.0`, `1.0` | map, mean 173.096 | all black, mean 0.000 |
| `inf`, `1e3` | map, mean 173.096 | all black, mean 0.000 |
| `nan`, `1e3` | map, mean 173.096 | map, mean 59.429 |
| `0.01`, `-1.0` | map, mean 173.096 | all black, mean 0.000 |

Nothing was refused in any of those ten cells. `PanoramaBackground` never reads the
planes; `GsplatBackground` forwards them to `gsplat.rasterization` as `near_plane` /
`far_plane`, which culls every gaussian outside the interval. That rasterizer is not
self-consistent about it either -- `znear=inf` culls everything while `znear=nan` is
silently ignored -- so there is no tolerance to defer to, and which lighting a caller
got was decided by which background happened to be plugged in.

A non-numeric plane was worse: `znear="0.01"` reached gsplat's pybind11 binding and
raised a `TypeError` naming `fully_fused_projection_packed_fwd` and dumping every
gaussian tensor, after one GPU render, without naming the parameter at fault.
`znear=10**400` leaked a bare `OverflowError` from a `float()` conversion.

## The change

A local `_clip_plane_error`, grouped beside `_resolution_error` for the same reason
(one place, before any render), with two rules:

- Each plane is a distance in meters, held to the shared
  `positive_finite_number_error` -- so what this refuses cannot diverge from a rate
  or a duration elsewhere in the tree, and a fractional distance stays usable (the
  `0.01` m default is one).
- `znear` must be strictly less than `zfar`. Equal or inverted planes are an empty
  frustum: no depth interval is left for geometry to fall in.

The ordering rule stays local rather than going into `utils.py`, per convention 11
in `AGENTS.md`: `render_environment_map` is the only place in the tree that takes
clip planes from a caller, and one caller is evidence of nothing. Checked *after*
the resolutions, so a call with both wrong keeps naming the resolution exactly as
before.

Only the domain is checked, not the quality a frustum buys, matching the resolution
knobs. `znear=0.5` with `zfar=0.6` asks for a 10 cm depth slab and stays accepted --
on a scene whose geometry sits at 3 m it really does come back black, correctly,
because the caller asked for a slab with nothing in it. Where "too tight" begins is
a judgement about the scene, which is the caller's to make.

`bake_environment_map` and `environment_map_cache_path` expose the resolutions but
not the planes, so they render on the documented default and there is nothing for
them to check or to encode in a cache key. A test pins that scope.

## Rollout artifact

Real 3DGS scene (4056 gaussians, a room shell with a warm `+X` window) rendered
through gsplat 1.4.0 on CUDA. Identical scene, identical resolution; only
`znear`/`zfar` differ.

environment map clip-plane domain

| | main, default frustum | main, `znear=10.0 zfar=1.0` | this branch, same call |
| --- | --- | --- | --- |
| status | success | success | refused |
| background renders paid | 6 | 6 | 0 |
| map mean / brightest channel | 59.296 / 251 | 0.000 / 0 | no map rendered |
| `derive_key_light` | azimuth +1.0 deg | ValueError blaming the scene | not reached |

The refusal names both planes and the consequence:

```
render_environment_map: znear must be < zfar, got znear=10 and zfar=1. Equal or
inverted clip planes leave no depth interval to render, so a background that honors
them produces an empty map rather than a dark scene.
```

The default-frustum map is **byte-identical between the two trees** (`maxabs` diff
`0`), so nothing legitimate moved.

## Tests

`tests/rendering/test_environment_map_clip_plane_domain.py`, 38 cases, named for the
behaviour it verifies and shaped after the resolution guard beside it.

- Pre-fix (source reverted, test kept): **26 failed, 12 passed** -- and all 12 that
  pass are the control and premise classes, so nothing in the accepted domain moved.
  The headline failure reports what the caller was left holding, including the
  dead-end remedy it followed on their behalf.
- Post-fix: **38 passed**.
- The two out-of-domain lists are parametrized over `positive_finite_number_error`
  itself and each case asserts its premise, so a value added to the shared domain is
  covered here without an edit.
- Two test backgrounds reproduce both real behaviours (one ignores the planes, one
  culls outside them), so the refusal is grounded in a real reading of the planes
  rather than asserted, and the suite needs no GPU.

## Gate

  (1538 files).
- `mypy strands_robots tests tests_integ`: `Success: no issues found in 1535 source
  files`; zero branch-only messages against a pristine `main` worktree.
- Paired run, identical 630-file selection on both trees (every `tests/rendering`
  test plus every guard that walks the tree, reads docstrings, or names this module),
  with the new file excluded from both: **27738 collected, 2 failed, 27635 passed,
  101 skipped on each**, identical failing-id sets. The two failures are
  pre-existing and need `docker` on `PATH`.
- `tests/rendering/` alone: 712 passed, 1 skipped. The resolution guard and
  `test_ibl.py` are untouched: 114 passed.
- `scripts/assemble_changelog.py --check`: fragments OK.

### DIFF
--- a/changelog.d/2717-environment-map-clip-plane-domain.md
+++ b/changelog.d/2717-environment-map-clip-plane-domain.md
@@ -1,0 +1,13 @@
+### Fixed
+- **rendering**: `render_environment_map` checks `znear` / `zfar` against the shared
+  `positive_finite_number_error` domain and requires `znear < zfar`, before any cube
+  face is rendered. An inverted or equal pair - `znear=10.0` with `zfar=1.0` - was
+  accepted and returned a full-size, entirely black map reporting success, after
+  paying all six background renders (GPU-bound for a `GsplatBackground`); the only
+  refusal the caller ever saw came from `derive_key_light`, blamed the scene ("the map
+  is black above the horizon"), and advised a search flag that fails the same way, so
+  neither clip plane was named anywhere. The planes cannot be left to the background:
+  with the same arguments `PanoramaBackground` ignores them and returns its usual map
+  while `GsplatBackground` forwards them to `gsplat.rasterization` and culls every
+  gaussian. An ordered pair that happens to frame nothing stays accepted, matching the
+  resolution knobs, which check the domain rather than the quality.
--- a/strands_robots/rendering/ibl.py
+++ b/strands_robots/rendering/ibl.py
@@ -41,7 +41,11 @@ from typing import Any
 
 import numpy as np
 
-from strands_robots.utils import boolean_flag_error, positive_whole_number_error
+from strands_robots.utils import (
+    boolean_flag_error,
+    positive_finite_number_error,
+    positive_whole_number_error,
+)
 
 from .camera import CameraParams
 from .color import relative_luminance, srgb_to_linear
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Adds a `_clip_plane_error` guard to `render_environment_map` so an unusable `znear`/`zfar` pair is refused with a `ValueError` naming both planes before any of the six background renders is paid for, instead of returning a full-size black map whose only downstream refusal (`derive_key_light`) blamed the scene and advised a dead-end remedy. Each plane is held to the shared `positive_finite_number_error` domain and the strict `znear < zfar` ordering rule stays local per AGENTS.md convention 11 (single caller). The guard runs after the resolution checks so a call with both wrong keeps naming the resolution first, and only the domain is checked, not the quality a frustum buys - a tight ordered pair that frames nothing stays accepted.

## What's good

- Verified locally at the head SHA: the 38 new tests pass, the neighboring resolution guard and `test_ibl.py` are unaffected (114 passed), and ruff check/format are clean on the changed files.
- Both internal callers of `render_environment_map` (`backgrounds.py` GS-panorama bake, `bake_environment_map`) render on the documented default frustum, so nothing in-tree hits the new refusal.
- The out-of-domain test lists parametrize over `positive_finite_number_error` itself with per-case premise assertions, matching the pattern the resolution guard established, and the two test backgrounds reproduce both measured real behaviours (plane-ignoring and plane-honoring) so the suite needs no GPU.
- The behaviour change is a loud, documented `ValueError` (the `Raises:` block was updated in the same diff), not a silent default change; scope control pins that `bake_environment_map` and `environment_map_cache_path` take no clip planes.
- Changelog fragment, behaviour-named test file, ASCII-only refusal text - all per AGENTS.md.

PR #2718, opened 2026-08-25
test(protomotions): cover the raw-motion resample onto the control rate

cagataycali:
## Problem

`MotionPlayer` accepts a raw ProtoMotions `.pt` as a first-class source and
resamples it onto the tracker's control rate. None of that path was exercised.

The existing `.pt` coverage (`tests/policies/protomotions/test_motion_file_loading.py`)
writes a **cache-shaped** payload, which returns from `_load_file` at
`if "control_dt" in data and "body_rot" in data` - before the resampler is
reached. So both raw layouts, the rate conversion, the rotation interpolation
and the unrecognised-layout refusal had no test at all:

```
strands_robots/policies/protomotions/motion_utils.py   169 stmts   61 missing   64%
missing: 147-153, 243, 333, 400, 404-466
```

That is `_resample_raw` in full (404-466), the `_calc_frame_blend` body it is
the only caller of (147-153), the call site (400), and the two construction
refusals a caller reaches before any frame is served (243, 333).

The resample decides which source instants the ONNX tracker ever sees, so a
fault there is not slower or coarser playback - it is a different motion. A
mis-derived frame count truncates the clip (and `get_state_at_frame` clamps to
that count, so the tail is silently unreachable rather than reported); a packed
library slices one entry out of a concatenated buffer by `length_starts`, so an
offset off by one entry plays *another clip* under the requested index; and the
six state channels are cut with that one offset, so a channel that drifted from
the others would pair one clip's joints with another's bodies.

## Change

One new test file, no production change:
`tests/policies/protomotions/test_raw_motion_is_resampled_to_the_control_rate.py`
(13 tests, 5 classes).

Every fixture row is keyed to its **global** source frame index (`body_pos[i]`
carries `x = i`, `body_ang_vel[i]` carries `z = i`, ...), so an assertion names
the source row a resampled frame came from instead of checking that it looks
plausible. The source rate is 10 Hz and the control rates are exact multiples of
it, so a resampled frame either lands on a source instant (and must equal that
row exactly) or at a known fraction between two (and must equal their blend
exactly) - no tolerance is needed for the five linearly interpolated channels.

Rotations are the one channel interpolated on the sphere, and the two plausible
degradations of that are pinned by measurement rather than by reading the source:

| rotation blend | unit norm | per-step angle over a 4x upsample |
| --- | --- | --- |
| `slerp` (shipped) | `1.0` | `22.5 deg` uniform (spread `1e-4 deg`) |
| `lerp` (as the other five channels use) | `0.9239` | - |
| `nlerp` (normalised `lerp`) | `1.0` | `21.598 / 23.402 deg` alternating |

So the unit-norm assertion catches the copy-paste degradation to `lerp`, and the
constant-rate assertion (held to `0.05 deg`) catches the subtler `nlerp` one - a
reference motion that speeds up and slows down inside every source interval.

## Coverage

Measured over the full suite (38403 tests), then the same coverage data unioned
with the new file (`--cov-append`, so the number is exact rather than projected):

| | statements | missing | coverage |
| --- | --- | --- | --- |
| `motion_utils.py` before | 169 | 61 | 64% |
| `motion_utils.py` after | 169 | **0** | **100%** |
| repo total before | 45830 | 2056 | 96% |
| repo total after | 45830 | 1995 | 96% |

13 tests close all 61 statements.

## Verification

Each test earns its place - 9 mutations of the resampler, each reverted after
measuring. Control clean, all 8 faults detected, 8 distinct firing sets:

| mutation | result | tests fired |
| --- | --- | --- |
| control (no mutation) | 13 passed | 0 |
| rotations blended with `lerp`, not `slerp` | 2 failed | 2 |
| rotations blended with `nlerp` | 1 failed | 1 |
| library entry offset ignored (always entry 0) | 5 failed | 5 |
| one channel sliced from the buffer head | 3 failed | 3 |
| frame count drops the endpoint (`+1`) | 6 failed | 6 |
| single-motion rate hardcoded to 30 fps | 2 failed | 2 |
| unrecognised layout no longer refused | 1 failed | 1 |
| no interpolation between source frames (`f1 == f0`) | 6 failed | 6 |

The `lerp` row fires a superset of the `nlerp` row, which is the constant-rate
assertion being the strictly stronger of the two. The resampler was restored
byte-identical afterwards.

Local gate:

- `ruff check` / `ruff format --check` over `strands_robots tests tests_integ`:
  clean, 1538 files formatted.
- `mypy strands_robots tests tests_integ`: 24 errors on this branch and 24 on a
  pristine `main` worktree, normalised message sets byte-identical in both
  directions, 0 in the new file.
- Paired run of 130 files (the whole `tests/policies/protomotions/` directory
  plus every test that walks the source tree, reads docstrings or reads
  `changelog.d/`), identical selection on both trees with the new file excluded
  from each: **4671 collected, 4617 passed, 54 skipped, 0 failed on both**.
- The new file: 13 passed.
- Without real torch (`torch` shimmed to raise `ImportError`, so the suite's
  numpy stand-in installs): **2 passed, 11 skipped**. The 11 `.pt` cases carry
  the skip marker and the two source-level refusals do not, so the file still
  contributes on an install without torch instead of erroring. `torch` is
  reached only inside the `.pt` writer, never at module scope.

No visual artifact: this adds tests only and changes no policy, simulation,
rendering, recording or asset behaviour. The measured tables above are the
evidence.

### DIFF
--- a/changelog.d/2718-protomotions-raw-motion-resample-tests.md
+++ b/changelog.d/2718-protomotions-raw-motion-resample-tests.md
@@ -1,0 +1,27 @@
+### Tests: the raw-motion resample onto the control rate is covered
+
+`MotionPlayer` takes a raw ProtoMotions `.pt` as a first-class source and
+resamples it onto the tracker's control rate, and that whole path was
+unexercised (#2718). The existing `.pt` coverage writes a *cache-shaped*
+payload, which returns from the loader before the resampler runs, so both raw
+layouts - a packed library sliced out of a concatenated buffer by
+`length_starts`, and a single motion declaring its own `fps` - the rate
+conversion, the rotation interpolation and the unrecognised-layout refusal had
+no test.
+
+`tests/policies/protomotions/test_raw_motion_is_resampled_to_the_control_rate.py`
+keys every fixture row to its global source frame index, so an assertion names
+the source row a resampled frame came from rather than checking that it looks
+plausible. The source rate is 10 Hz and the control rates are exact multiples of
+it, so each resampled frame either lands on a source instant (and equals that
+row exactly) or at a known fraction between two (and equals their blend
+exactly), which needs no tolerance for the five linearly interpolated channels.
+Rotations are held to the two properties that separate spherical interpolation
+from the alternatives by measurement: unit length (a blend taken on the line
+measures `0.9239` for the fixture's 90-degrees-apart neighbours) and a constant
+angular rate (a normalised line blend alternates `21.598 deg` / `23.402 deg`
+against a uniform `22.5 deg`).
+
+This takes `strands_robots/policies/protomotions/motion_utils.py` from 64% to
+100% statement coverage - 61 uncovered statements to 0 - with no production
+change.
--- a/tests/policies/protomotions/test_raw_motion_is_resampled_to_the_control_rate.py
+++ b/tests/policies/protomotions/test_raw_motion_is_resampled_to_the_control_rate.py
@@ -1,0 +1,345 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""A raw ProtoMotions motion is resampled onto the tracker's own control rate.
+
+:class:`~strands_robots.policies.protomotions.motion_utils.MotionPlayer` accepts
+two kinds of source: a cache dict that is *already* at the control rate, and a
+raw ProtoMotions ``.pt``, which is at whatever rate it was captured at (30 Hz,
+120 Hz, per-entry inside a library) and has to be resampled. The resample is the
+only thing that decides which source instants the ONNX tracker ever sees, so
+getting it wrong is not slower or coarser playback - it is a different motion:
+
+* a mis-derived frame count truncates the clip or invents frames past its end,
+  and :meth:`MotionPlayer.get_state_at_frame` clamps to that count, so the tail
+  is silently unreachable rather than reported;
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Test-only PR: one new file, `tests/policies/protomotions/test_raw_motion_is_resampled_to_the_control_rate.py` (13 tests, 5 classes), plus its `changelog.d/` fragment. It covers the previously unexercised raw-`.pt` path of `MotionPlayer` - both raw layouts (packed library sliced by `length_starts`, single motion declaring `fps`), the rate conversion, the slerp rotation blend, and the three construction refusals. I cross-checked the test's independently derived `_expected_frames` against the shipped formula in `_resample_raw` (they agree), and the refusal-message assertions against the actual `ValueError`/`TypeError`/`KeyError` strings in `motion_utils.py` (all match verbatim). No production change, so no behavioural or one-way-door surface to evaluate.

## What's good

- Fixture rows are keyed to their global source frame index, so a wrong `length_starts` offset or a channel-slice drift fails by naming the wrong source row rather than by a plausibility tolerance.
- The two rotation-degradation assertions (unit norm vs `lerp`, constant angular rate vs `nlerp`) are pinned by measurement, and the PR description records a per-mutation kill matrix with distinct firing sets.
- Conventions followed: behaviour-describing test names, ASCII-only, no host paths, `changelog.d/2718-*.md` fragment instead of a direct `CHANGELOG.md` edit, torch imported lazily inside the `.pt` writer behind a `real_torch_installed()` skip so the two source-level refusal tests still run on a torch-less install.

PR #2719, opened 2026-08-25
fix(hardware): hold the ROS 2 bridge QoS depth to a domain the transport can carry

cagataycali:
## Problem

`RosTelemetryBridge` and `HardwareRosBridge` take a `qos_depth` and hand it straight to rclpy's `create_publisher` / `create_subscription` as the `qos_or_depth` argument, which becomes `QoSProfile(depth=value, history=KEEP_LAST)`. Every sibling constructor parameter of that pair is measured at the boundary, each with its reason written down in the docstring:

| parameter | domain | in `Raises:` |
|---|---|---|
| `domain_id` | `dds_domain_id_error` (RTPS port map) | yes |
| `spin_period` | `positive_finite_number_error` | yes |
| `enable_commands` | `boolean_flag_error` | yes |
| `joint_limits` | finite numeric pairs, `min <= max` | yes |
| `qos_depth` | none | no |

It is also the only one whose consumer is not reached from the constructor: publishers are built lazily, on the first `publish_joint_states` / `publish_image` for a robot.

## What that cost, measured

Driving the real `RosTelemetryBridge` against real rclpy (ROS 2 Jazzy, `ros:jazzy-ros-core`), every spelling below **constructed successfully**:

| `qos_depth` | construction | first publish |
|---|---|---|
| `10` | OK | OK, `depth=10` |
| `0` | OK | OK, `depth=0` -> `UserWarning: A zero depth with KEEP_LAST doesn't make sense; no data could be stored. This will be interpreted as SYSTEM_DEFAULT` |
| `False` | OK | OK, same zero-depth path by another route |
| `True` | OK | OK, `depth=True` - a silent depth of 1 |
| `-1` | OK | `ValueError: history depth must be greater than or equal to zero` |
| `2.5` | OK | `TypeError: Expected QoSProfile or int, but received <class 'float'>` |
| `"10"` | OK | `TypeError: Expected QoSProfile or int, but received <class 'str'>` |
| `None` | OK | `TypeError: Expected QoSProfile or int, but received <class 'NoneType'>` |
| `2**31` | OK | `TypeError: __init__(): incompatible constructor arguments` (pybind11, into the QoS profile) |

Two distinct harms. The first four are *accepted*: the declared depth is silently not the depth in force, announced by a `UserWarning` that `warnings.warn` shows once per location - so a second bridge in the same process says nothing at all. The rest raise from inside rclpy, naming neither the parameter nor the bridge, and on the telemetry-only path not until the first frame is published, mid-run.

On the command path `create_subscription` runs in the constructor, so the same mistake lands there instead - and after the damage: with `qos_depth=2.5`, `HardwareRosBridge(robot=..., enable_commands=True)` raised the rclpy `TypeError` with `rclpy.ok()` still `True` and `ROS_DOMAIN_ID` already written to `'0'`.

`np.int64(10)` is worth naming separately: it denotes a perfectly good depth and rclpy refuses it outright (`TypeError: Expected QoSProfile or int`), as it does `np.int32`, `np.float64` and `np.bool_`.

`qos_depth` appears in no test, no doc and no example. The one rclpy double in the suite that reaches `create_publisher` takes the depth as `_depth` and discards it, so no existing test could observe which depth an endpoint was built with. The double added here records it.

## Change

`_qos_history_depth_error` in `strands_robots.ros_telemetry`, called beside the `domain_id` check.

The floor, the `bool` refusal and the strict-`int` requirement are inherited from `positive_count_error`, the shared domain for a discrete count a C API consumes directly rather than coerces - which is precisely the situation here, given rclpy's refusal of `np.int64`. The floor of 1 closes `0` and `False`; the domain's boolean refusal closes `True`.

Only the transport's own ceiling is added on top, beside the transport that has it, in the manner of `_transport_port_error` in `strands_robots.tools.use_rosbridge`. The depth reaches the middleware through `rmw_qos_profile_t`, whose pybind11 binding takes a signed 32-bit integer, so the boundary is measured rather than chosen: `2**31 - 1` builds a publisher, `2**31` does not. `MAX_QOS_HISTORY_DEPTH` is pinned to that arithmetic so this library cannot start refusing a depth the transport is happy to carry.

Placement is the `domain_id` placement, for the `domain_id` reasons: ahead of the process-wide `ROS_DOMAIN_ID` write, so a refused bridge leaves the environment as it found it, and ahead of the rclpy probe, so the same caller mistake reports identically on an install with the `[ros2]` extra and one without it - which is what lets every refusal test here run on a minimal install. An accepted depth still reaches every endpoint verbatim; nothing coerces it, because a converted value is a depth on the wire the caller never named.

## Tests

New `tests/test_ros_qos_history_depth_domain.py`, 83 cases. Pre-fix, with only the call site removed and the guard left defined: **46 failed, 37 passed**, every failure inside the four regression classes and none in the domain, ceiling-premise or accepted-value control classes. A raw revert of the whole change is a collection error instead, so both splits are reported.

11 mutations, control clean:

| mutation | fires | where |
|---|---|---|
| b0 control (no mutation) | 0 | - |
| b1 the guard call site is removed (the defect) | 46 | 4 regression classes |
| b2 the guard runs after the `ROS_DOMAIN_ID` write | 15 | the environment class only |
| b3 the guard runs after the rclpy probe | 45 | 3 classes |
| b4 the transport ceiling clause is dropped | 9 | domain + 3 |
| b5 the floor admits zero (`non_negative` instead of `positive`) | 5 | domain + 3 |
| b6 the `bool` refusal is dropped (hand-rolled floor) | 4 | domain + 3 |
| b7 an integral float / numpy int is admitted | 8 | domain + 3 |
| b8 the accepted depth is coerced with `int()` | 0 | see below |
| b9 the ceiling becomes a policy number (1000) | 2 | the ceiling-premise class only |
| b10 the subclass stops forwarding the depth | 17 | 3 classes, incl. the verbatim control |

b8's zero is reported rather than removed, because the reason is the guard: measured, `int(v) is v` for every accepted probe value, and `bool` - the one `int` subclass whose type `int()` would change - is refused. So the coercion is provably the identity on the accepted domain, an unreachable degree of freedom; the verbatim control that would catch a real coercion is pinned by b10 instead.

A structural sweep asserts every `qos_depth`-taking function either calls the guard or forwards the value, with a non-vacuity test naming today's two surfaces and a planted-source test proving the scanner detects a surface that does neither.

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1538 files).
- mypy **24 == 24** against pristine `main`, normalized message sets byte-identical, zero in the changed files.
- Paired run over an identical 365-file selection (every test naming these bridges, the shared domains, `ROS_DOMAIN_ID` or `require_optional`, plus every test that walks the package with `ast` / `rglob` / `inspect`), the new file excluded from both trees: **18 failed, 17917 passed, 99 skipped, 24 errors on both**, 18058 collected each, 42 identical failing/erroring ids, both `comm` directions empty. 41 of the 42 need `awsiot` (`awsiotsdk`, declared in the `[mesh-iot]` extra, confirmed absent); the remaining one is pre-existing on `main`.
- The new file plus the three closest sibling suites: **236 passed** on a second environment (mujoco 3.12.0) as well as the first (mujoco 3.5.0 floor, lerobot 0.6.2 from source).
- Behaviour against real rclpy re-checked post-fix inside `ros:jazzy-ros-core`.

No visual artifact: this touches no policy, simulation behaviour, rendering, recording or robot asset. The measured tables above are the evidence. The sibling ceiling constants (`MAX_DDS_DOMAIN_ID`, `MAX_ZMQ_TIMEOUT_MS`, `MAX_IK_SMOOTHING`) appear in no docs page either, so the docstrings updated here - both classes' `Args:` and `Raises:` - are the whole documentation surface for this parameter.

### DIFF
--- a/changelog.d/2719-ros-qos-history-depth-domain.md
+++ b/changelog.d/2719-ros-qos-history-depth-domain.md
@@ -1,0 +1,48 @@
+### Fixed: the ROS 2 bridge QoS depth is checked where the caller gives it
+
+`RosTelemetryBridge` and `HardwareRosBridge` take a `qos_depth` and hand it to rclpy's
+`create_publisher` / `create_subscription` as the `qos_or_depth` argument, which becomes
+`QoSProfile(depth=value, history=KEEP_LAST)`. Every sibling constructor parameter of that pair is
+measured at the boundary - `domain_id` against the RTPS port map, `spin_period` against the positive
+finite domain, `enable_commands` against the boolean domain, `joint_limits` against a finite numeric
+pair, each with the reason written down - and this one was handed through unchecked. It is also the
+only one whose consumer is not reached from the constructor: publishers are built lazily, on the
+first `publish_joint_states` / `publish_image` for a robot.
+
+So a bad depth reported a successful construction. Measured against rclpy on ROS 2 Jazzy, `0` and
+`False` were *accepted*: rclpy warns "A zero depth with KEEP_LAST doesn't make sense; no data could
+be stored. This will be interpreted as SYSTEM_DEFAULT" and builds the endpoint with the middleware
+default, so the declared depth was silently not the depth in force - reported by a `UserWarning`
+that `warnings.warn` shows once per location, so a second bridge in the same process said nothing at
+all. `True` was accepted as a silent depth of 1: one frame of history on a stream the caller asked
+to buffer.
+
+Every other spelling raised from inside rclpy, naming neither the parameter nor the bridge. `-1`
+gave `ValueError: history depth must be greater than or equal to zero`; `2.5`, `10.0`, `"10"`,
+`None`, `nan`, `inf` and `np.int64(10)` gave `TypeError: Expected QoSProfile or int`; and `2**31`
+and above failed the pybind11 conversion into the QoS profile instead. On the telemetry-only path
+none of those surfaced until the first frame was published, mid-run. On the command path
+`create_subscription` runs in the constructor, so the same mistake raised there - after the
+process-wide `ROS_DOMAIN_ID` write and after the node had been created.
+
+The depth is now measured against `positive_count_error`, the shared domain for a discrete count a C
+API consumes directly rather than coerces, which is exactly why the strict-`int` requirement is the
+right one here: rclpy refuses `np.int64(10)` outright even though it names a perfectly good depth.
+The floor of 1 closes `0` and `False`, and the domain's boolean refusal closes `True`. Only the
+transport's own ceiling is added on top, beside the transport that has it, in the manner of
+`_transport_port_error` in `strands_robots.tools.use_rosbridge`: the depth is stored in
+`rmw_qos_profile_t` through a pybind11 binding taking a signed 32-bit integer, so
+`MAX_QOS_HISTORY_DEPTH` is `2**31 - 1` - measured, not chosen, and pinned by the arithmetic so this
+library cannot start refusing a depth the transport is happy to carry.
+
+The guard is placed beside the `domain_id` check, ahead of both the process-wide `ROS_DOMAIN_ID`
+write and the rclpy probe. That is what makes a refused depth leave the environment as it found it,
+and makes the same caller mistake report identically on an install with the `[ros2]` extra and one
+without it - so the refusal is checked on a minimal install too. An accepted depth still reaches
+every endpoint verbatim: nothing coerces it, because a converted value is a depth on the wire the
+caller never named.
+
+Nothing observed any of this before. `qos_depth` appeared in no test, no doc and no example, and the
+one rclpy double in the suite that reaches `create_publisher` takes the depth as `_depth` and
+discards it, so no existing test could see which depth an endpoint was built with. The new tests
+record it instead.
--- a/strands_robots/hardware_ros_bridge.py
+++ b/strands_robots/hardware_ros_bridge.py
@@ -73,7 +73,13 @@ class HardwareRosBridge(RosTelemetryBridge):
             Only an ``int`` in ``[0, 232]`` names a domain: RTPS derives its
             discovery ports from it, and 233 lands past the end of the port space.
         node_name: Internal rclpy node name (defaults to ``strands_hardware``).
-        qos_depth: Depth of the publishers'/subscription's KEEP_LAST history.
+        qos_depth: Depth of the publishers'/subscription's KEEP_LAST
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Adds boundary validation for the one `RosTelemetryBridge` / `HardwareRosBridge` constructor parameter that was previously handed to rclpy unchecked: `qos_depth` is now measured against the shared `positive_count_error` domain (strict `int`, `bool` refused, floor of 1) plus a locally-owned transport ceiling (`MAX_QOS_HISTORY_DEPTH = 2**31 - 1`, pinned to the `rmw_qos_profile_t` signed-32-bit storage bound rather than a policy number). Verified against `main`: the guard lands after the existing `dds_domain_id_error` check and ahead of both the process-wide `ROS_DOMAIN_ID` write and the `require_optional("rclpy")` probe, so a refused depth leaves the environment untouched and reports identically with or without the `[ros2]` extra; the subclass forwards `qos_depth` to `super().__init__`, so one guard covers both surfaces, and the new structural sweep pins that forwarding. Because `positive_count_error` admits only strict `int >= 1`, the follow-on `value > MAX_QOS_HISTORY_DEPTH` comparison can never see a non-int, and the over-ceiling clause reports a distinguishable reason from the floor. The refusals introduced here (`0`, `False`, `True`, floats, numpy ints, oversize values) all previously produced either a silently-wrong depth on the wire or an unattributed exception from inside rclpy mid-run, so converting them to a construction-time `ValueError` naming the parameter is the sanctioned loud-rejection pattern, not a silent behaviour change.

## What's good

- Guard structure follows AGENTS.md rule 11 exactly: shared domain reused, only the transport-specific ceiling kept local, with the `_transport_port_error` precedent cited in the docstring.
- Both classes' `Args:` and `Raises:` blocks updated in the same change, per the `Raises:`-completeness convention.
- Changelog fragment (`changelog.d/2719-ros-qos-history-depth-domain.md`) rather than a direct `CHANGELOG.md` edit.
- The new test file uses `monkeypatch.setitem(sys.modules, ...)`, behaviour-describing test names, no host paths, all-ASCII strings, and includes a non-vacuity check plus a planted-source detection test for the structural sweep - so the sweep cannot pass vacuously.
- The mutation table in the PR description (b0-b10) documents which test class catches each degree of freedom, including the honest note on b8's unreachable coercion.

PR #2721, opened 2026-08-25
fix(vera): skip a provenance record the offline ckpt index cannot read

cagataycali:
One malformed `provenance.json` under the mounted checkpoint root stopped the offline VERA container from starting, and took every healthy checkpoint beside it.

## The problem

`wandb_offline_resolve` exists because VERA loads its Jacobian IDM by wandb run id, and the real `download_checkpoint` calls the wandb API before it looks at disk, so an offline container never comes up. The module indexes the mounted checkpoints by the `wandb_run` recorded in each `provenance.json` and answers from disk instead.

`_index_local_ckpts` skipped a file that does not parse and one that omits `wandb_run`, then handed anything else straight to the two operations that assume a string - using the value as a dict key, and splitting it on `/`. So a payload that parses into something that is not a run record raised instead of being skipped. Measured over every way a `provenance.json` can be malformed:

| `provenance.json` payload | before | after |
|---|---|---|
| unparseable file | skipped, quietly | skipped, quietly |
| empty file | skipped, quietly | skipped, quietly |
| object, no `wandb_run` key | skipped, quietly | skipped, quietly |
| object, `wandb_run` is `null` | skipped, quietly | skipped, quietly |
| object, `wandb_run` is `""` | skipped, quietly | skipped, quietly |
| object, `wandb_run` is a falsy non-string (`0`, `false`, `[]`, `{}`) | skipped, quietly | skipped, quietly |
| object, `wandb_run` is a non-zero number | `AttributeError: 'int' object has no attribute 'split'` | skipped, named |
| object, `wandb_run` is a non-empty list | `TypeError: unhashable type: 'list'` | skipped, named |
| object, `wandb_run` is a non-empty object | `TypeError: unhashable type: 'dict'` | skipped, named |
| payload is a JSON array | `AttributeError: 'list' object has no attribute 'get'` | skipped, named |
| payload is a JSON string | `AttributeError: 'str' object has no attribute 'get'` | skipped, named |
| payload is a JSON number | `AttributeError: 'int' object has no attribute 'get'` | skipped, named |
| payload is JSON null | `AttributeError: 'NoneType' object has no attribute 'get'` | skipped, named |

Seven of the thirteen malformation modes raised, and the tolerance was exactly backwards: a file that could not be read at all was skipped, while one that parsed into the wrong shape was fatal.

The six quiet rows all route through the pre-existing `if not run` check, which sits *above* the new string guard and is deliberately silent: a checkpoint directory carrying no wandb run is the ordinary case for every artifact not loaded by run id. A falsy non-string is in that same class - `0`, `[]` and `{}` carry nothing to index, so their consequence is identical to a missing key, and neither before nor after does such a payload reach the two string operations. The warning is reserved for a record that carried *something* the index cannot use. Ordering is therefore load-bearing in both directions, and the break table below pins it.

It is fatal rather than local because the module self-installs at import and `launch_server.py` imports it before the server module, so the raise lands on the container entrypoint. The whole root is scanned in one pass, so the healthy records go with it. Driving the launcher's own `import` + `install()` sequence against a root holding two healthy released records and one malformed one:

```
# before
Traceback (most recent call last):
  File "/opt/launch_server.py", line 3, in <module>
    import wandb_offline_resolve
  File "/opt/wandb_offline_resolve.py", line 137, in <module>
    install()
  File "/opt/wandb_offline_resolve.py", line 79, in install
    index = _index_local_ckpts(ckpt_root)
  File "/opt/wandb_offline_resolve.py", line 65, in _index_local_ckpts
    index[run] = ckpt_dir  # full "entity/project/run_id"
TypeError: unhashable type: 'dict'
                                                     exit 1 - the server never starts

# after
WARNING [offline-resolve] ignoring /ckpts/idm-omni-x21o0cwe/provenance.json: wandb_run must be a string, got dict
INFO    [offline-resolve] indexed 4 local ckpt run ids under /ckpts: ['idm-mimicgen-37oa162u', 'pusht-idm']
INFO    [offline-resolve] your-wandb-entity/jacobian-learning/37oa162u -> /ckpts/idm-mimicgen-37oa162u/model.ckpt (local, no wandb)
                                                     exit 0
```

The run lost in that first case is `37oa162u` - the one `entrypoint.sh` defaults `VERA_DYNAMICS_RUN_ID` to for mimicgen.

## The change

`_index_local_ckpts` skips a record it cannot read as a run record, exactly as it already skips one that does not parse and one that omits the key, and names the file and the offending type in a warning so the operator can see which mounted checkpoint did not index.

The rule is the one `transforms.provenance.load_provenance` already applies to a provenance payload - check the type before using the value. The two differ only in disposition, for the reason that module documents: a caller asking which episodes are synthetic has no fallback and is refused, while this resolver has one (the network) and a checkpoint it could not read is simply absent from the index. Refusing here is what the code already did, and it is the behaviour that takes down the server this module exists to bring up.

Every previously-indexed record is indexed on identical terms, and the records that were skipped quietly stay quiet - a checkpoint directory carrying no wandb run is the ordinary case for every artifact not loaded by run id, so warning about it would be noise.

## Tests

`tests/policies/vera/test_offline_ckpt_index_skips_an_unusable_provenance_record.py`, 47 cases. On pre-fix code: **15 failed, 32 passed**, all 15 inside the regression class and none in the premise, control, resolver-behaviour or install classes.

The case set is graded rather than listed: each entry is checked against the property that makes it interesting - it parses as JSON but is not a usable record - so a case that stops being either is caught here instead of quietly duplicating the parse-failure path.

Break table, 11 mutations, 10 distinct firing sets, control 0:

| mutation | tests firing |
|---|---|
| control - no mutation | 0 |
| the JSON-object guard dropped | 8 |
| the `wandb_run` string guard dropped | 7 |
| both guards dropped (pre-fix code) | 15 |
| a non-string run id coerced with `str()` instead of skipped | 6 |
| only a null payload guarded (a narrower shape check) | 7 |
| both skips silent - the ignored file is not named | 7 |
| the warning widened to the records already skipped quietly | 3 |
| the shape guard placed after the key is read | 8 |
| only numbers refused, not every non-string | 6 |
| the missing-checkpoint skip dropped | 1 |

The two halves are disjoint and complementary: 8 + 7 = 15, the pre-fix total. Moving the shape guard below the key read fires the same 8 as dropping it, because `data.get` raises first - placement is load-bearing, not stylistic. Widening the warning to the quiet records fires 3 control cases, which is what keeps this from becoming a noisy scan.

The module was previously at 24% coverage (47 of 62 statements unreached, in two contiguous blocks) because nothing imported it; it is now at **100%** (0 of 68), taking the repository's uncovered total from 2031 to 1984.

One note on the import form: the container helpers are excluded from type checking (`tool.mypy.exclude`) since they are written against the image's dependency set, where `vera` and `omegaconf` are installed and unstubbed here. mypy's `exclude` does not apply to a module reached through a *followed* import, so a static import from a test would silently pull the module back into the type-checked graph and report six `import-not-found` errors. The test imports it by name, which keeps the declared exclusion effective.

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1542 files).
- mypy **24 == 24** against a pristine `main` worktree, normalized message sets byte-identical in both directions, 0 in the changed files.
- Paired run over an identical 372-file selection (the whole `tests/policies/vera` suite plus every guard that walks the tree, reads docstrings or reads project config), the new file excluded from both trees: **17 failed, 17187 passed, 72 skipped** on both, 17276 collected each, failing-id sets identical both ways. All 17 need `awsiot` (the `[mesh-iot]` extra), absent here and equally absent on both trees.
- The new file passes on mujoco 3.5.0 (the declared floor) with lerobot 0.6.2 from source, and on mujoco 3.12.0. Whole `tests/policies/vera` suite: 793 passed.

## Artifact

The launcher's own import + `install()` sequence run against the same three-checkpoint root on both trees, captured verbatim (log timestamps trimmed; paths shown at the container's `/ckpts` and `/opt`):

offline resolver before and after a malformed provenance record

## Round 2 - changelog

`de66b8f` **docs(vera): scope the offline-resolver naming claim to the records it names** - two markdown files. No code, no tests, and no measured number above changes.

Review feedback noted that a *falsy* non-string `wandb_run` (`0`, `false`, `[]`, `{}`) is caught by the pre-existing `if not run` check before the new string guard, so it is skipped quietly rather than named.

The observation is correct, and measuring it showed the imprecision was wider than the `after` column. I drove `_index_local_ckpts` over all thirteen malformation modes on this tree, and over a reconstruction of the pre-fix function on the same inputs:

```
case               pre-fix   post-fix  named?
wandb_run=0        quiet     skipped   no
wandb_run=false    quiet     skipped   no
wandb_run=[]       quiet     skipped   no
wandb_run={}       quiet     skipped   no
wandb_run=''       quiet     skipped   no
wandb_run=null     quiet     skipped   no
wandb_run=12345    RAISE     skipped   YES  wandb_run must be a string, got int
wandb_run=['a']    RAISE     skipped   YES  wandb_run must be a string, got list
wandb_run={a:1}    RAISE     skipped   YES  wandb_run must be a string, got dict
payload=[]         RAISE     skipped   YES  expected a JSON object, got list
payload=null       RAISE     skipped   YES  expected a JSON object, got NoneType
payload=7          RAISE     skipped   YES  expected a JSON object, got int
payload='str'      RAISE     skipped   YES  expected a JSON object, got str

raising modes pre-fix: 7 of 13
```

So the `before` column was wrong too: `[]`, `{}` and `0` never reached the string operations pre-fix and never raised. The problem table above now splits the three collapsed non-string rows into four - one quiet row for the falsy values, three named rows for the truthy ones - and reports what each payload actually did. Thirteen modes, seven raising, unchanged in total.

**What the commit changes.** Two shipped documents claimed more than the code does, and unlike this description they do not stay behind after merge:

- the changelog fragment listed "a number, list or object" among the payloads that raised - true only of the non-zero, non-empty ones. Left as written it would have shipped a reproduction that does not reproduce, permanently, in `CHANGELOG.md`.
- `docker/README.md` said a non-string `wandb_run` "is skipped and named in a warning" - true of a record that carried something unusable, not of one that carried nothing.

Both now name the warning only for the records that get one, and state the quiet path explicitly. The two module docstrings needed no change: they claim only that such a record "is skipped", which is accurate for all thirteen modes.

**No guard change, for a reason rather than for the round budget.** The quiet disposition is the correct one: a falsy `wandb_run` carries nothing to index, so its consequence is indistinguishable from the missing key the same branch already handles silently by design, and the warning exists to name a record that carried *something* unusable. Reordering the guards to name `[]` would move the empty string `""` - the ordinary no-run-recorded case - onto the noisy path, which is exactly what the `the warning widened to the records already skipped quietly` mutation is in the break table to prevent (it fires 3 control cases). The branch stays pinned either way: `QUIETLY_SKIPPED_PAYLOADS` covers `wandb_run-is-null` and `wandb_run-is-empty`, which route through the identical check, and `test_a_directory_with_no_run_to_index_is_skipped_without_a_warning` asserts the empty log for every one of them.

`tests/test_changelog_fragment_gate.py`, `test_changelog_fragments.py`, `test_changelog_format.py`: **89 passed**. The fragment was also reflowed so no continuation line begins with `- `, which markdown would otherwise render as a nested bullet.

### DIFF
--- a/changelog.d/2721-vera-offline-ckpt-index-unusable-provenance.md
+++ b/changelog.d/2721-vera-offline-ckpt-index-unusable-provenance.md
@@ -6,19 +6,20 @@
   omits `wandb_run`, but handed anything else to the two operations that assume a
   string - using the value as a dict key and splitting it on `/` - so a payload
   that parses into an array, string, number or null, or an object whose
-  `wandb_run` is a number, list or object, raised out of the scan. That is fatal
-  rather than local: the module self-installs on import and `launch_server.py`
-  imports it before the server module, so one malformed record among the mounted
-  checkpoints killed the container entrypoint before the server was reached, and
-  the whole root is scanned in one pass, so the healthy records went with it -
-  including `37oa162u`, the run the entrypoint defaults to for mimicgen. The
-  tolerance was backwards: a file that could not be read at all was skipped,
-  while one that parsed into the wrong shape was fatal. Such a record is now
-  skipped and named in a warning with the type that made it unusable, following
-  the rule `transforms.provenance.load_provenance` applies to a provenance
-  payload - check the type before using the value - and differing only in
-  disposition, because this resolver has a fallback (the network) where a caller
-  asking which episodes are synthetic has none. Every record indexed before is
-  indexed on identical terms, and a checkpoint directory carrying no wandb run is
-  still skipped silently, since that is the ordinary case for every artifact not
-  loaded by run id.
+  `wandb_run` is a non-zero number, non-empty list or non-empty object, raised
+  out of the scan. That is fatal rather than local: the module self-installs on
+  import and `launch_server.py` imports it before the server module, so one
+  malformed record among the mounted checkpoints killed the container entrypoint
+  before the server was reached, and the whole root is scanned in one pass, so
+  the healthy records went with it - including `37oa162u`, the run the entrypoint
+  defaults to for mimicgen. The tolerance was backwards: a file that could not be
+  read at all was skipped, while one that parsed into the wrong shape was fatal.
+  Such a record is now skipped and named in a warning with the type that made it
+  unusable, following the rule `transforms.provenance.load_provenance` applies to
+  a provenance payload - check the type before using the value - and differing
+  only in disposition, because this resolver has a fallback (the network) where a
+  caller asking which episodes are synthetic has none. Every record indexed
+  before is indexed on identical terms, and a checkpoint directory carrying no
+  run to index (`wandb_run` absent, null, empty or otherwise falsy, which reaches
+  neither string operation and so never raised) is still skipped silently, since
+  that is the ordinary case for every artifact not loaded by run id.
--- a/changelog.d/2721-vera-offline-ckpt-index-unusable-provenance.md
+++ b/changelog.d/2721-vera-offline-ckpt-index-unusable-provenance.md
@@ -1,0 +1,24 @@
+### Fixed
+
+- **policies/vera/docker**: the offline checkpoint resolver skips a
+  `provenance.json` it cannot read as a run record, instead of raising on it.
+  `_index_local_ckpts` already skipped a file that does not parse and one that
+  omits `wandb_run`, but handed anything else to the two operations that assume a
+  string - using the value as a dict key and splitting it on `/` - so a payload
+  that parses into an array, string, number or null, or an object whose
+  `wandb_run` is a number, list or object, raised out of the scan. That is fatal
+  rather than local: the module self-installs on import and `launch_server.py`
+  imports it before the server module, so one malformed record among the mounted
+  checkpoints killed the container entrypoint before the server was reached, and
+  the whole root is scanned in one pass, so the healthy records went with it -
+  including `37oa162u`, the run the entrypoint defaults to for mimicgen. The
+  tolerance was backwards: a file that could not be read at all was skipped,
+  while one that parsed into the wrong shape was fatal. Such a record is now
+  skipped and named in a warning with the type that made it unusable, following
+  the rule `transforms.provenance.load_provenance` applies to a provenance
+  payload - check the type before using the value - and differing only in
+  disposition, because this resolver has a fallback (the network) where a caller
+  asking which episodes are synthetic has none. Every record indexed before is
+  indexed on identical terms, and a checkpoint directory carrying no wandb run is
+  still skipped silently, since that is the ordinary case for every artifact not
+  loaded by run id.
--- a/strands_robots/policies/vera/docker/README.md
+++ b/strands_robots/policies/vera/docker/README.md
@@ -112,7 +112,11 @@ An explicit `-e VERA_…` always overrides the auto-wiring.
   checkpoint's `provenance.json` (`wandb_offline_resolve.py`) and sets
   `WANDB_MODE=offline`, so **no wandb network / API key is needed**. Both the
   proven mimicgen IDM (`37oa162u`) and the omni default (`x21o0cwe`) resolve
-  locally; an unknown run id transparently falls back to wandb.
+  locally; an unknown run id transparently falls back to wandb. A
+  `provenance.json` the scan cannot read as a run record - it does not parse,
+  does not hold a JSON object, or carries a non-string `wandb_run` - is skipped
+  and named in a warning, so one malformed record among the mounted checkpoints
+  never costs the healthy ones.
 - The container only **serves** actions; the host runs the sim. The heavy
   `eval` extra (robosuite/mimicgen) is therefore **not** installed by default
   (build with `--build-arg VERA_WITH_EVAL=1` to bundle it).
--- a/strands_robots/policies/vera/docker/README.md
+++ b/strands_robots/policies/vera/docker/README.md
@@ -114,9 +114,12 @@ An explicit `-e VERA_…` always overrides the auto-wiring.
   proven mimicgen IDM (`37oa162u`) and the omni default (`x21o0cwe`) resolve
   locally; an unknown run id transparently falls back to wandb. A
   `provenance.json` the scan cannot read as a run record - it does not parse,
-  does not hold a JSON object, or carries a non-string `wandb_run` - is skipped
-  and named in a warning, so one malformed record among the mounted checkpoints
-  never costs the healthy ones.
+  does not hold a JSON object, or carries a non-string `wandb_run` - is skipped,
+  so one malformed record among the mounted checkpoints never costs the healthy
+  ones. A record that carried something unusable is named in a warning with the
+  offending type; one that carries no run to index (`wandb_run` absent, null,
+  empty or otherwise falsy) is skipped silently, since that is the ordinary case
+  for every artifact not loaded by run id.
 - The container only **serves** actions; the host runs the sim. The heavy
   `eval` extra (robosuite/mimicgen) is therefore **not** installed by default
   (build with `--build-arg VERA_WITH_EVAL=1` to bundle it).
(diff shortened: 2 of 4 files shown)


yinsong1986 (dismissed):
## Summary

Narrow, well-scoped robustness fix to the VERA offline checkpoint resolver: `_index_local_ckpts` now type-checks a parsed `provenance.json` (JSON object, string-valued `wandb_run`) before hashing the value as a dict key and splitting it on `/`, so a single malformed record among the mounted checkpoints is skipped and named in a warning instead of raising out of the import-time `install()` and killing the container entrypoint along with every healthy checkpoint. The two new guards only add skips for values that previously raised, so every previously-indexed record indexes on identical terms, and the previously-quiet skips (unparseable file, missing/empty `wandb_run`) stay quiet. No public API, wire format, or persisted schema is touched; the module is a container-internal helper.

## What's good

- The warning logs the file path and the offending *type*, never the payload contents - nothing operator-mounted can leak into the log.
- The test imports the module by name (`importlib.import_module`) with a comment explaining why, keeping the `tool.mypy.exclude` on `strands_robots/policies/vera/docker/` effective per mypy's followed-import semantics.
- Test hygiene follows AGENTS.md: `monkeypatch.setitem(sys.modules, ...)` for the fake `vera` tree (restores on teardown), behaviour-describing test names, no host paths, ASCII-only strings, and a changelog fragment under `changelog.d/` rather than an edit to `CHANGELOG.md`.
- The test file grades its own case set (each unusable payload is asserted to parse-but-not-be-a-record), so a case that degenerates into the parse-failure path fails loudly instead of silently duplicating coverage.

One non-blocking observation, for awareness only: a *falsy* non-string `wandb_run` (`0`, `false`, `[]`, `{}`) is caught by the pre-existing `if not run` check before the new string guard, so it is skipped quietly rather than named in a warning - a minor logging asymmetry with the PR table, not a correctness issue.

yinsong1986 (approved):
## Summary

Narrow robustness fix to the VERA offline checkpoint resolver: `_index_local_ckpts` now type-checks a parsed `provenance.json` (payload must be a JSON object, `wandb_run` must be a string) before the two operations that assume a string - hashing the value as a dict key and splitting it on `/`. A record the scan cannot read is skipped and named in a warning with the offending type, instead of raising out of the import-time `install()` and killing the container entrypoint along with every healthy checkpoint under the mounted root. Guard ordering is deliberate in both directions: the object check sits above `data.get` (so a non-object payload never reaches the key read) and the string check sits below the pre-existing silent `if not run` (so the ordinary no-run-recorded case stays quiet), and both placements are pinned by mutation cases in the new test file. No public API, wire format, persisted schema, CLI flag, or env-var surface is touched - the module is a container-internal helper under the `tool.mypy.exclude` set, and every previously-indexed record indexes on identical terms.

## What's good

- The round-2 commit resolved the prior review's logging-asymmetry observation the right way: it corrected the two shipped documents (changelog fragment, `docker/README.md`) to match measured behaviour rather than widening the warning onto the quiet path, and the break table's `warning widened to the records already skipped quietly` mutation (3 control cases firing) records why the guard was left alone.
- Warnings log only the file path and the type name, never the payload contents, so nothing operator-mounted leaks into the container log.
- Test hygiene follows AGENTS.md: `monkeypatch.setitem(sys.modules, name, None)` for the unimportable-module cases (restores on teardown), behaviour-describing test names, ASCII-only strings, no host paths, and a `changelog.d/` fragment rather than an edit to `CHANGELOG.md`.
- The `importlib.import_module` form with its explanatory comment keeps the module out of the type-checked import graph per mypy's followed-import semantics, exactly as the declared `tool.mypy.exclude` intends.

## Verification suggestions

Beyond CI, the one-command spot-check is the launcher sequence the PR's artifact shows: against a checkpoint root holding two healthy records and one `provenance.json` whose `wandb_run` is an object, `python -c "import wandb_offline_resolve"` inside the container should exit 0, warn naming the malformed file, and report `37oa162u` in the indexed run ids.

PR #2722, opened 2026-08-25
test(training): grade every field-scoped domain guard, not the five one helper name reached

cagataycali:
## Problem

`tests/training/test_spec_field_read_discovery.py` grades the field-scoped
shared-domain guards from the outside. Its own docstring states the promise:

> the set of field-scoped guards is discovered structurally, so a new domain
> guard is held to the same rule the moment it lands.

The discovery is keyed on the **name** of the helper a guard uses to list the
backend modules:

```python
if not readers or "_trainer_modules" not in names or "_calls_the_gate" not in names:
    continue
```

The guards spell that helper two ways. Five use `_trainer_modules`; eight use
`_training_modules`. So the sweep grades five of thirteen, and its own
non-vacuity assertion pins that count (`assert set(_guard_modules()) == {5 files}`),
which is what made the gap look deliberate.

The eight it could not see are exactly the eight that need it. Each carries its
own inline copy of the by-name-only read that the shared `reads_spec_field` rule
exists to replace, so each is blind to a **table-driven** read
(`getattr(spec, field)` over a tuple of names) - the form a provider that
forwards a spec on uses, and the form that names no field in any attribute
access:

| guard | field(s) | scope helper | sees a read by name | sees a table-driven read | graded by the sweep |
|---|---|---|---|---|---|
| `test_checkpoint_cadence_domain.py` | `save_freq` | `_trainer_modules` | yes | yes | yes |
| `test_launch_topology_domain.py` | `num_gpus`, `num_nodes` | `_trainer_modules` | yes | yes | yes |
| `test_lora_hyperparameter_domain.py` | `lora_r`, `lora_alpha` | `_trainer_modules` | yes | yes | yes |
| `test_seed_domain.py` | `seed` | `_trainer_modules` | yes | yes | yes |
| `test_validation_episodes_domain.py` | `val_episodes` | `_trainer_modules` | yes | yes | yes |
| `test_clip_range_domain.py` | `clip_param` | `_training_modules` | yes | **no** | **no** |
| `test_discount_factor_domain.py` | `gamma` | `_training_modules` | yes | **no** | **no** |
| `test_gae_lambda_domain.py` | `lam` | `_training_modules` | yes | **no** | **no** |
| `test_gradient_clip_domain.py` | `max_grad_norm` | `_training_modules` | yes | **no** | **no** |
| `test_initial_temperature_domain.py` | `init_alpha` | `_training_modules` | yes | **no** | **no** |
| `test_loss_weight_domain.py` | `value_loss_coef`, `entropy_coef` | `_training_modules` | yes | **no** | **no** |
| `test_optimization_epochs_domain.py` | `num_learning_epochs` | `_training_modules` | yes | **no** | **no** |
| `test_temperature_learning_rate_domain.py` | `alpha_lr` | `_training_modules` | yes | **no** | **no** |

The two sets coincide exactly: every guard outside the sweep is blind, and every
guard inside it is not. The eight predate the sweep (added 2026-08-07/08 against
2026-08-20), which is why they were never converted - and the entry that
introduced the shared rule records the scope it believed it had, "the **four**
field-scoped shared-domain guards".

The consequence is the thing each of those eight promises not to allow. From
`TestOneOwnerForTheDiscountDomain`:

> a third RL backend that starts discounting a return with the field fails this
> test until it does.

## Measured consequence

A real backend module gains a reader that forwards a gated field through a table
and never calls the gate - the shape a transport-only provider already uses:

```python
_PLANTED_FIELDS = ("gamma",)

def _planted_forward(spec: object) -> list[object]:
    return [getattr(spec, f) for f in _PLANTED_FIELDS]
```

Each guard's one-owner class, asked about its own field, on `main` and on this
branch:

| planted read of | before | after |
|---|---|---|
| `spec.clip_param` | 4 passed (clean sweep) | 2 failed |
| `spec.gamma` | 4 passed (clean sweep) | 2 failed |
| `spec.lam` | 4 passed (clean sweep) | 2 failed |
| `spec.max_grad_norm` | 4 passed (clean sweep) | 2 failed |
| `spec.init_alpha` | 4 passed (clean sweep) | 2 failed |
| `spec.value_loss_coef` | 4 passed (clean sweep) | 2 failed |
| `spec.num_learning_epochs` | 4 passed (clean sweep) | 2 failed |
| `spec.alpha_lr` | 4 passed (clean sweep) | 2 failed |

Eight for eight. After, the report names the module:

```
AssertionError: modules read spec.gamma without the shared gate: ['mock.py']
```

The silence is the failure mode: before, the guard reports a clean tree, so the
reader lands with the domain unenforced for it and nothing says so.

## Change

`tests/training/test_spec_field_read_discovery.py`

- Discovery is keyed on the two properties that make a guard gradeable rather
  than on the names of the helpers that carry them: it defines exactly one
  `_reads...` helper (the scan this file grades, and the one `_reader_helper`
  resolves), and its scope is rooted at the backend tree - detected as an
  `inspect.getfile(Trainer)` call, which is what makes the scope derived rather
  than listed. That rule is single-sourced in `is_field_scoped_guard`, so the
  sweep over the real tree and the constructed exemplars grade the same code.
- `FIELD_SCOPED_GATES` gains the eight gates, so a discovered guard's gate
  resolves instead of raising `StopIteration`.
- The forwarding-provider assertions derive their own scope from
  `_FORWARDED_FIELDS` (`FORWARDED_GATES`) instead of assuming every gated field
  is forwarded. The eight own `RLTrainSpec` fields that no provider forwards, so
  asserting a forwarded read of them would assert a false premise; they are
  graded on the reader scan alone. The derived set is pinned, so a field leaving
  the provider's table still surfaces rather than silently dropping its gate
  from the parametrization.

The eight guards' `_reads...` helpers now delegate to `reads_spec_field`, which
is the one place both forms of a read are defined.

`learning_rate` and `run_size` are deliberately still out of scope and now
excluded by a property rather than by a name: neither has a reader helper (they
scan `Trainer` subclasses, not field reads), so there is no notion of "reads the
field" here to grade. A test pins that.

## Tests

`TestTheDiscoveryDoesNotDependOnAHelperName` grades the rule on constructed
exemplars: a guard qualifies under `_trainer_modules`, `_training_modules` and a
third spelling; a *listed* scope under the previously-keyed name does not (so
this is not compatibility, it is a different property); a scope rooted at another
module does not; no reader helper does not; two reader helpers do not; and a
non-vacuity case asserts the rule separates them. One further test reports a
discovered guard whose gate is unregistered by name, because the gate lookup
elsewhere raises `StopIteration` without naming the omission.

Pre-fix, widening the discovery alone surfaces the eight directly:

```
FAILED ...::test_it_sees_a_table_driven_read[test_discount_factor_domain.py]
  AssertionError: test_discount_factor_domain.py does not see a table-driven read
  of spec.gamma, so a backend that forwards the field by name is outside its
  derived scope
```

`test_it_still_sees_a_read_by_name` passes for all thirteen throughout, so only
the table form was missing.

Mutation coverage of this change (failures over `tests/training/`):

| mutation | fires |
|---|---|
| control, unmutated | 0 |
| discovery requires the `_trainer_modules` spelling again | 4 |
| one reader helper reverted to the by-name form | 1 |
| forwarding assertions parametrized over every gate again | 16 |
| the tree-rooting requirement dropped | 2 |
| one-reader requirement weakened to at-least-one | 1 |
| a field leaves the provider's forwarding table | 3, two of them pre-existing |
| tree-rooting accepts any `getfile` call | 1 |

The last row was 0 until the exemplar for it was added: every guard that resolves
a non-`Trainer` module also resolves the backend tree, so the mutation was a
no-op over today's tree and the property was unpinned. The row above it fires two
tests that predate this change, which is what shows the premise guard removed
from the forwarding assertions is preserved rather than dropped.

## Verification

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1541 files).
- `mypy strands_robots tests tests_integ`: `Success: no issues found in 1538 source files`; zero errors introduced.
- `tests/training/` plus the changelog guards: 2952 passed, 4 skipped (2926 before; the 26 added are this change's).
- Blast radius, identical 336-file selection on both trees with the changed files
  excluded from each - all of `tests/training/` plus every test that walks the
  tree or reads sources - 15267 collected on each: failing-ID sets identical.
  Two failures are pre-existing and need `docker` on `PATH`; five more are
  pre-existing in the ROS suites and depend only on whether `rclpy` resolves,
  confirmed by re-running those four files on both trees under one matched
  interpreter path (5 failed, 294 passed on each, identical IDs).

No visual artifact: this changes no policy, simulation, rendering, recording or
asset behaviour. The before/after table above is the evidence.

### DIFF
--- a/changelog.d/2722-field-scoped-guard-discovery.md
+++ b/changelog.d/2722-field-scoped-guard-discovery.md
@@ -1,0 +1,17 @@
+### Fixed
+
+- **training**: the guard that holds every field-scoped shared-domain guard to
+  the "a reader scan must see both forms of a read" rule now discovers all
+  thirteen of them rather than five. It keyed discovery on the *name* of the
+  helper a guard uses to list the backend modules, and the guards spell that
+  helper two ways (`_trainer_modules` and `_training_modules`), so the eight
+  guards using the other spelling sat outside the sweep - silently, because it
+  reported a clean tree over the five it could see and pinned that count as its
+  own non-vacuity assertion. All eight were the by-name-only scans the shared
+  rule exists to replace, so each certified a clean `readers == {...}` sweep
+  while a backend that forwards its gated field through a table and skips the
+  gate went unreported. Discovery is now keyed on the two properties that make a
+  guard gradeable (one reader helper, a scope rooted at the backend tree), the
+  eight reader scans route through the shared rule, and the forwarding-provider
+  assertions derive their own scope from what that provider actually forwards so
+  they no longer assume every gated field is one of them.
--- a/tests/training/test_clip_range_domain.py
+++ b/tests/training/test_clip_range_domain.py
@@ -71,6 +71,7 @@ from strands_robots.training._validate import (
 from strands_robots.training.base import Trainer
 from strands_robots.training.rl import RLTrainSpec
 from strands_robots.utils import positive_finite_number_error
+from tests.training._spec_field_reads import reads_spec_field
 
 # The one backend whose update clips a policy ratio.
 ON_POLICY = "ppo"
@@ -326,11 +327,14 @@ def _training_modules() -> list[pathlib.Path]:
 
 
 def _reads_the_width(source: str) -> bool:
-    """Does *source* read ``spec.clip_param``?"""
-    return any(
-        isinstance(node, ast.Attribute) and node.attr == "clip_param" and getattr(node.value, "id", None) == "spec"
-        for node in ast.walk(ast.parse(source))
-    )
+    """Does *source* read ``spec.clip_param``, by name or through a forwarding table?
+
+    Delegated to the shared rule so this guard and its siblings cannot disagree
+    about what counts as a read - a transport-only provider reads every field it
+    forwards through ``getattr(spec, field)`` and names none of them in an
+    attribute access.
+    """
+    return reads_spec_field(source, ("clip_param",))
 
 
 def _calls_the_gate(source: str) -> bool:
--- a/tests/training/test_discount_factor_domain.py
+++ b/tests/training/test_discount_factor_domain.py
@@ -35,6 +35,7 @@ from strands_robots.training._validate import (
 )
 from strands_robots.training.base import Trainer
 from strands_robots.training.rl import RLTrainSpec
+from tests.training._spec_field_reads import reads_spec_field
 
 # Every backend that discounts a return with the field.
 RL_BACKENDS = ("ppo", "fast_sac")
@@ -227,11 +228,14 @@ def _training_modules() -> list[pathlib.Path]:
 
 
 def _reads_the_discount_factor(source: str) -> bool:
-    """Does *source* read ``spec.gamma``?"""
-    return any(
-        isinstance(node, ast.Attribute) and node.attr == "gamma" and getattr(node.value, "id", None) == "spec"
-        for node in ast.walk(ast.parse(source))
-    )
+    """Does *source* read ``spec.gamma``, by name or through a forwarding table?
+
+    Delegated to the shared rule so this guard and its siblings cannot disagree
+    about what counts as a read - a transport-only provider reads every field it
+    forwards through ``getattr(spec, field)`` and names none of them in an
+    attribute access.
+    """
+    return reads_spec_field(source, ("gamma",))
 
 
 def _calls_the_gate(source: str) -> bool:
--- a/tests/training/test_gae_lambda_domain.py
+++ b/tests/training/test_gae_lambda_domain.py
@@ -39,6 +39,7 @@ from strands_robots.training._validate import (
 )
 from strands_robots.training.base import Trainer
 from strands_robots.training.rl import RLTrainSpec
+from tests.training._spec_field_reads import reads_spec_field
 
 # The one backend that estimates an advantage trace.
 ON_POLICY = "ppo"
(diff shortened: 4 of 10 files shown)


yinsong1986 (approved):
## Summary

Test-only change that widens the meta-guard in `tests/training/test_spec_field_read_discovery.py` from five field-scoped domain guards to all thirteen. The prior discovery keyed on the literal helper name `_trainer_modules`, and the eight RL-hyperparameter guards spell that helper `_training_modules`, so they sat outside the sweep while it pinned a "clean" five-file set as its own non-vacuity assertion. Discovery is now keyed on two structural properties (exactly one `_reads...` helper plus a scope rooted at `inspect.getfile(Trainer)`), the eight previously-ungraded guards route their reader scans through the shared `reads_spec_field` rule instead of private by-name-only AST walks, and the forwarding-provider assertions are re-scoped to `FORWARDED_GATES` (derived from `_FORWARDED_FIELDS`) so they no longer assert a forwarded read of `RLTrainSpec` fields that no provider forwards - a premise I verified is false on this head (none of the nine RL fields appear in `sagemaker._FORWARDED_FIELDS`).

## What's good

- Verified locally at `14dca744`: the touched training suites pass (1148 tests), `ruff check` and `ruff format --check` are clean on all nine Python files.
- The fix closes its own class rather than the instance: `TestTheDiscoveryDoesNotDependOnAHelperName` grades the discovery rule itself with constructed exemplars (including a third spelling, `_backend_modules`), and `test_a_listed_scope_does_not_qualify` pins that a listed scope under the old keyed name is rejected rather than grandfathered.
- Both derived sets are pinned against silent shrinkage (`test_the_scan_finds_the_field_scoped_guards`, `test_the_forwarded_gates_are_the_expected_ones`), which is exactly the failure mode the PR fixes.
- Changelog fragment present (`changelog.d/2722-field-scoped-guard-discovery.md`), no non-ASCII in any changed file, no host paths, scope tightly confined to `tests/training/`.

PR #2724, opened 2026-08-25
fix(policies): read a batched observation.state through both ProtoMotions conventions

cagataycali:
## Problem

`ProtoMotionsPolicy._pack_by_name` offers three observation conventions, and **two of them read the same key**, `observation.state`:

- **convention 1** pairs it with a `state_keys` list supplied on the observation - the first shape `get_actions`'s own docstring documents;
- **convention 3** pairs it with the policy's `_robot_state_keys`, populated at construction and refreshed by `set_robot_state_keys`.

Both index the array positionally, so both need it flat. Only convention 3 flattened it. A runtime that batches its state feeds `(1, D)` rather than `(D,)` - the shape LeRobot's `AddBatchDimensionObservationStep` produces, and which `embodiment.py` deliberately creates so `observation.state` batches alongside image tensors - so **adding the documented `state_keys` list turned a readable observation into a dead end**.

Measured through the public `get_actions`, four cells over one 29-value pose (a stub session echoes the resolved `current_dof_pos`, so the returned joint targets *are* the state the policy read):

| `observation.state` | `state_keys` on obs | convention | before | after |
|---|---|---|---|---|
| `(29,)` | absent | 3 | resolved | resolved |
| `(29,)` | present | 1 | resolved | resolved |
| `(1, 29)` | absent | 3 | resolved | resolved |
| `(1, 29)` | **present** | **1** | **`TypeError: only length-1 arrays can be converted to Python scalars`** | resolved |

The failure names neither the key, the shape nor a remedy, it comes from `float()` deep inside the pack loop, and `TypeError` is outside `get_actions`'s declared `Raises` (`RuntimeError`, `KeyError`). The same asymmetry rejects a nested list and a `(D, 1)` column, and rejects a `(1, 29)` torch tensor - measured - which is what a torch pipeline hands over.

## Change

`observation.state` is read and flattened **once**, ahead of every convention, instead of inside one branch:

```python
raw_state = obs.get("observation.state")
state_arr = None if raw_state is None else np.asarray(raw_state, dtype=np.float32).reshape(-1)
```

Convention 3's branch-local `np.asarray(...).reshape(-1)` is removed, so there is a single normalization with two readers rather than two copies that can drift. Net `+13 / -4` in `strands_robots/policies/protomotions/policy.py`. Behaviour for an already-flat state is unchanged - the three cells that resolved before still return byte-identical vectors.

`observation.state: None` still falls through to the per-joint convention, and a `state_keys` list missing a joint still refuses by name; both are pinned.

Unitree G1 (`mujoco_menagerie`) posed from the joint targets `get_actions` returns, rendered headless with EGL. The tracker session is stubbed to echo the joint positions it was handed, so **each rendered pose is the observation the policy read**. One free camera (azimuth 140, elevation -12) and one headlight setting for every panel; panel 2 has no pose to draw because `get_actions` raised before a target existed.

ProtoMotions batched observation.state

Panel 1 and panel 3 joint-target vectors are byte-identical (`np.array_equal` over 29 `float32`), equal to the pose fed in, and their 497x221x3 renders are byte-identical too. Both panels are verified embedded verbatim in the figure by searching for their paste offset and comparing arrays.

## Tests

`tests/policies/protomotions/test_observation_state_reads_the_same_through_every_convention.py`, 22 tests. The regression is a **derived grid**: four spellings of one state vector (flat, batched row, column, nested list) crossed with both conventions, each driven through the public `get_actions` and required to return the same vector. Reverting the fix fails 6 of them, including `test_a_batched_state_is_readable_with_the_documented_key_list`.

Two premises keep the grid from grading itself: the spellings really do differ in shape (three distinct shapes over four spellings, all carrying the same values), and the two conventions are **distinguishable** - with a reversed `state_keys` list on the obs the two readers give different answers, so the assertion names which one ran.

Two structural tests pin the *placement*, because a per-branch copy is how the readers drifted apart: `state_arr` is assigned exactly once and that assignment carries the flattening, and the flattening precedes every convention branch. They are load-bearing rather than decorative - the mutation that fixes the behaviour by adding `reshape(-1)` to convention 1's own copy fires exactly those two and nothing else.

### Break table

| mutation | tests fired |
|---|---|
| control (unmutated) | 0 |
| revert the fix | 6 |
| `reshape(-1)` added to convention 1's own copy instead | 2 (both structural) |
| shared normalization drops `reshape(-1)` | 9 |
| convention 1 indexes the raw value | 4 |
| convention 3 reads positionally, not by name | 2 |
| convention 2's refusal drops the hit count | 1 |
| a no-op second `state_arr` binding in convention 1 | 1 |
| normalization moved below convention 1 | 6 |
| docstring paragraph removed, code kept | 0 |

Nine mutations, nine distinct firing sets, control clean. The last row is an honest zero: nothing grades that prose.

### Coverage

Both `observation.state` readers were entirely uncovered before this, which is why the asymmetry survived. `strands_robots/policies/protomotions/policy.py` goes **80% -> 90%** (38 -> 20 uncovered statements) in the full suite; repo total 95.71% -> 95.74%. The 20 that remain are the `onnxruntime` session lifecycle (`404-431`), three alternate extract routes, and the four `fill` fallbacks that no call site reaches (`_extract_dof_pos` / `_extract_dof_vel` both leave `fill` at its default).

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1542 files).
- `mypy strands_robots tests tests_integ`: **24 errors on this branch and 24 on a pristine `main` worktree**, normalized message sets byte-identical, none in the changed files.
- Full suite, `mujoco` 3.5.0 (the declared floor) with `lerobot` 0.6.2 from source: base `29 failed, 38186 passed, 309 skipped, 24 errors` -> branch `29 failed, 38208 passed, 309 skipped, 24 errors`. **+22 passed, exactly the new tests**, and the 54 failing/erroring ids are identical in both directions (`comm` empty both ways). 44 of the 54 need `awsiot` (the `[mesh-iot]` extra); the other 10 are pre-existing environment artifacts - `lerobot`-from-source `encode()` drift, `strands-robots` package metadata absent when running from a source tree, and a stale `/tmp` dataset - and are unchanged by this branch.

### DIFF
--- a/changelog.d/2724-protomotions-batched-observation-state.md
+++ b/changelog.d/2724-protomotions-batched-observation-state.md
@@ -1,0 +1,14 @@
+### Fixed
+
+- **policies/protomotions**: a batched `observation.state` now resolves through
+  every observation convention `ProtoMotionsPolicy` accepts. `_pack_by_name`
+  reads that one key from two of its three conventions - paired with a
+  `state_keys` list supplied on the observation, and paired with the policy's own
+  `_robot_state_keys` - and both index the array positionally, but only the
+  second flattened it first. A `(1, D)` state, the shape LeRobot's
+  `AddBatchDimensionObservationStep` produces, therefore resolved without the key
+  list and died with it: `get_actions` raised `TypeError: only length-1 arrays
+  can be converted to Python scalars`, a message naming neither the key, the
+  shape nor a remedy, and a type outside its own documented `Raises`. The
+  flattening now happens once, ahead of every convention, so the two readers
+  cannot disagree about the same observation.
--- a/strands_robots/policies/protomotions/policy.py
+++ b/strands_robots/policies/protomotions/policy.py
@@ -512,16 +512,26 @@ class ProtoMotionsPolicy(Policy):
         """Pack per-joint values from the obs dict in canonical joint order.
 
         Tries three obs conventions in order:
-        1. ``observation.state`` as a flat array + a matching ``state_keys``.
+        1. ``observation.state`` plus a matching ``state_keys`` list on ``obs``.
         2. Per-joint keys ``<name><suffix>`` directly on ``obs``.
         3. A single ``observation.state`` array whose ordering matches
            ``self._robot_state_keys`` (set via :meth:`set_robot_state_keys`).
+
+        ``observation.state`` is flattened before either convention indexes it,
+        so a batched ``(1, D)`` state reads identically to a flat ``(D,)`` one.
         """
+        # ``observation.state`` is read by conventions 1 and 3 below, and both
+        # index it positionally, so it is normalized once here rather than in
+        # each branch. A runtime that batches the state feeds ``(1, D)`` -
+        # LeRobot's own ``AddBatchDimensionObservationStep`` does exactly that -
+        # and flattening in only one branch made the two conventions disagree
+        # about the same observation.
+        raw_state = obs.get("observation.state")
+        state_arr = None if raw_state is None else np.asarray(raw_state, dtype=np.float32).reshape(-1)
+
         # 1. observation.state + explicit joint-key list on obs.
-        state_arr = obs.get("observation.state")
         keys_list = obs.get("state_keys")
         if state_arr is not None and keys_list is not None:
-            state_arr = np.asarray(state_arr, dtype=np.float32)
             out = np.zeros(self._config.num_dofs, dtype=np.float32)
             for i, name in enumerate(self._config.joint_names):
                 key = name + suffix
@@ -552,7 +562,6 @@ class ProtoMotionsPolicy(Policy):
 
         # 3. observation.state matches self._robot_state_keys order.
         if state_arr is not None and self._robot_state_keys:
-            state_arr = np.asarray(state_arr, dtype=np.float32).reshape(-1)
             out = np.zeros(self._config.num_dofs, dtype=np.float32)
             for i, name in enumerate(self._config.joint_names):
                 key = name + suffix
--- a/tests/policies/protomotions/test_observation_state_reads_the_same_through_every_convention.py
+++ b/tests/policies/protomotions/test_observation_state_reads_the_same_through_every_convention.py
@@ -1,0 +1,284 @@
+"""Every ``observation.state`` spelling resolves to the same G1 pose.
+
+:meth:`ProtoMotionsPolicy._pack_by_name` offers three observation conventions
+and two of them read the same key, ``observation.state``: convention 1 pairs it
+with a ``state_keys`` list supplied on the observation, convention 3 pairs it
+with the policy's own ``_robot_state_keys``. Both index the array positionally,
+so both need it flat -- and a runtime that batches its state feeds ``(1, D)``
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

Hoists the `observation.state` flatten in `ProtoMotionsPolicy._pack_by_name` out of convention 3's branch to a single normalization ahead of every convention, so a batched `(1, D)` state (the shape LeRobot's `AddBatchDimensionObservationStep` produces) resolves identically whether or not the caller also supplies the documented `state_keys` list. Before, convention 1 indexed the un-flattened array and died with an unhelpful `TypeError`. The production change is 3 lines of moved code; the rest is a regression suite driving the public `get_actions` over a spellings-by-conventions grid plus AST pins that keep the normalization single-sourced. No public API, wire format, or persisted schema changes.

## What's good

- Changelog fragment at `changelog.d/2724-protomotions-batched-observation-state.md` per AGENTS.md step 3; no direct `CHANGELOG.md` edit.
- The regression suite compares outputs through the public `get_actions` (behavior, not implementation) and its premise test proves both `observation.state` readers are actually reached, so the parity cells cannot silently grade one code path.
- Verified locally at the head: the full `tests/policies/protomotions/` suite passes (194 tests), `ruff check` and `ruff format --check` clean, no non-ASCII or host paths in the added files, and the docstring/test-naming pinning suites pass.
- The one behavior edge the hoist introduces - an `observation.state` that `np.asarray(..., dtype=np.float32)` cannot convert now refuses before convention 2 (per-joint keys) is tried - only bites an observation that is itself malformed while also carrying a full per-joint spelling, and the same payload already raised whenever `state_keys` was present. Not a realistic caller in the tree.

yinsong1986 (approved):
## Summary

Hoists the `observation.state` normalization in `ProtoMotionsPolicy._pack_by_name` to a single `np.asarray(...).reshape(-1)` ahead of the three convention branches, removing convention 3's branch-local copy, so a batched `(1, D)` state (the shape LeRobot's `AddBatchDimensionObservationStep` produces) resolves identically with and without the documented `state_keys` list. Before, convention 1 indexed the un-flattened array and died in `float()` with a `TypeError` outside `get_actions`'s declared `Raises`; now there is one flatten with two readers instead of two copies that can drift. Verified the head diff against `main` directly: the change is exactly the +13/-4 described, `observation.state: None` still falls through to the per-joint convention, and the short-key-list refusal still names the missing joint by name. The new 22-test module drives the public `get_actions` over a spelling-by-convention grid, pins its own premises (the spellings genuinely differ in shape; the two readers are distinguishable via a reversed key list), and pins the single-sourced placement structurally so a per-branch copy cannot quietly return.

## What's good

- Changelog fragment under `changelog.d/` rather than an edit to `CHANGELOG.md` (AGENTS.md step 3).
- The break table's mutation-by-firing-set analysis, including the honest zero for the docstring-only mutation, makes the test module's claims checkable rather than asserted.
- Scope discipline: the production diff is 13 lines in one method, and the three previously-resolving cells are pinned byte-identical.

PR #2725, opened 2026-08-25
fix(simulation): carry an unnamed joint's state across a scene rebuild

cagataycali:
## Problem

`eject_robot_from_scene` rebuilds the compiled model, which renumbers every joint id. The dynamic
state is therefore snapshotted under a name-based key and written back once the new model exists.
`_joint_key` built that key and returned `None` for any unnamed non-free joint, on the stated
grounds that "its body may carry several, so the body does not single it out". A `None` key is
skipped, so the joint's `qpos`, `qvel` and `qfrc_applied` were dropped and it came back at its
fresh-compile value while the operation reported `success`.

An unnamed `<joint>` is the ordinary MJCF spelling: a door hinge or a drawer slide is rarely named.
So removing one robot silently reset the articulation of everything else in the scene.

Measured through the public API on a scene holding an arm plus a door on an unnamed hinge and a
drawer carrying two unnamed joints on one body:

| unnamed joint | seeded `(qpos, qvel)` | after `remove_robot("arm")`, before | after, with this change |
|---|---|---|---|
| `door_panel[0]` hinge | `(0.70, -1.25)` | `(0.0, 0.0)` | `(0.70, -1.25)` |
| `drawer[0]` hinge | `(0.31, 0.44)` | `(0.0, 0.0)` | `(0.31, 0.44)` |
| `drawer[1]` slide | `(-0.17, 2.05)` | `(0.0, 0.0)` | `(-0.17, 2.05)` |

Both calls returned `status: success`.

## Change

The premise was sound but the conclusion was too strong. The body does not single the joint out, but
the body *plus the joint's position among that body's joints* does. MuJoCo stores a body's joints
contiguously from `body_jntadr` in declaration order, so `_joint_key` gains a third form
`("body_joint", body_name, ordinal)` and `_resolve_joint_key` resolves it back through the same two
values.

Nothing shifts an ordinal, because no scene op inserts a joint into an existing body - the patch
vocabulary is `add_body`, `add_geom`, `add_site`, `set_body_pos`, `set_body_quat`, `delete_body`.

Unchanged, and pinned as unchanged: a named joint keeps `("joint", name)`; an unnamed free joint
keeps `("body", body_name)`, which is unambiguous because the compiler refuses a free joint
alongside any other joint on the same body; and a joint whose body is *also* unnamed still gets no
key, because nothing identifies either end across the rebuild.

Two joints of one body are what make the ordinal load-bearing rather than decorative: a hinge and a
slide both use width 1/1, so the width check `_restore_scene_state` already performs cannot tell
them apart, and keying both on the body alone would restore each from the other's values.

## Visual artifact

One MJCF scene, one `remove_robot("arm")` call, three renders from one fixed camera in MuJoCo
headless (EGL). Panel 2 drives the pre-change `_joint_key` through an otherwise identical call.

unnamed joint state across a scene rebuild

Panels 2 and 3 are the same scene after the same call and differ in 16690 px, all of it the
furniture whose joints panel 2 reset. Panels 1 and 3 differ in 2456 px, which is the arm that was
removed. The figure script asserts each panel is embedded verbatim (`np.array_equal` at the measured
paste offset) and re-derives every quoted number from the state it read back.

## Tests

`tests/simulation/mujoco/test_unnamed_joint_state_survives_a_rebuild.py`, 17 tests, GL-free
(`mesh=False`, no rendering).

Pre-fix, reverting only the source: **6 failed, 11 passed**, with zero failures in the over-reach
control class or in either premise test.

Break table - 6 mutations, 5 distinct firing sets, control 0, source restored byte-identical:

| mutation | tests fired |
|---|---|
| control (no mutation) | 0 |
| raw revert: an unnamed non-free joint gets no key | 6 |
| `_resolve_joint_key` ignores the ordinal (always the body's first joint) | 3 |
| the ordinal is keyed off the model rather than the body (`jid` itself) | 5 |
| the "joint has since gained a name" guard dropped | 1 |
| the ordinal bound check dropped | 1 |

The ordinal-ignoring mutation is the one that would swap the drawer's hinge and slide; it fires the
round-trip test, the resolve test and the two-joints test, and no control.

## Gate

- `ruff check` clean; `ruff format --check` clean (1488 files already formatted).
- `mypy`: 0 errors in the changed module; the 4 reported on this tree are pre-existing and in other
  files.
- `tests/simulation/` plus the docstring and changelog guards, identical selection with the new file
  deselected: **13247 passed, 190 skipped, 0 failed**, and the FAILED/ERROR id sets are identical to
  a pristine-`main` measurement of the same selection (both empty, `comm` empty both directions).
- The new file plus `test_scene_ops_guardrails.py` and `test_remove_object_camera_cascade.py`, the
  two suites that exercise the same snapshot/restore path: **71 passed**.
- Run on MuJoCo 3.5.0 (the declared floor).

## Round 1 changelog

**`call-test-lint` was red on the pushed head; it is the only thing that moved.**

`hatch run lint` runs `mypy strands_robots tests tests_integ`, so the new test file is graded too.
`Simulation._world` is typed `SimWorld | None`, and the fixture plus the named-joint control read
`_model` / `_data` straight off it, which is six `union-attr` errors. The Gate section above measured
mypy on the changed *module* and missed them; the sentence "0 errors in the changed module" was true
and insufficient.

Fix, in the test file only: the fixture and the control assert the handle is present before reading
through it, the way the rest of the suite spells it (`assert sim._world is not None`), and the
control re-reads the handle after the eject into its own name rather than reading `sim._world` twice,
so the post-rebuild reads are narrowed on their own value. No assertion about the behaviour under
test changed, and no source file changed.

Re-measured on the new head:

- `hatch run lint`: `ruff check` clean, `ruff format --check` 1544 files already formatted, `mypy`
  **Success: no issues found in 1541 source files** (the 4 errors quoted in the Gate section above
  are gone from this tree as well).
- The new file plus `test_scene_ops_guardrails.py` and `test_remove_object_camera_cascade.py`:
  **71 passed**.

### DIFF
--- a/changelog.d/2725-unnamed-joint-state-across-a-rebuild.md
+++ b/changelog.d/2725-unnamed-joint-state-across-a-rebuild.md
@@ -1,0 +1,37 @@
+### Fixed: an unnamed joint keeps its state across a scene rebuild
+
+`eject_robot_from_scene` rebuilds the compiled model, which renumbers every joint id, so the
+dynamic state is snapshotted under a name-based key and written back once the new model exists.
+`_joint_key` built that key and returned `None` for any unnamed non-free joint, on the stated
+grounds that "its body may carry several, so the body does not single it out". A `None` key is
+skipped, so the joint's `qpos`, `qvel` and `qfrc_applied` were dropped and it came back at its
+fresh-compile value while the operation reported `success`.
+
+An unnamed `<joint>` is the ordinary MJCF spelling - a door hinge or a drawer slide is rarely
+named - so removing one robot silently reset the articulation of everything else in the scene.
+Measured on a scene holding an arm plus a door on an unnamed hinge and a drawer carrying two
+unnamed joints: with the door at `qpos 0.70` / `qvel -1.25`, the drawer hinge at `0.31` / `0.44`
+and the drawer slide at `-0.17` / `2.05`, `remove_robot("arm")` returned `success` and left all
+three at `0.0`, changing 16690 pixels of the rendered scene. The same call now leaves all three
+byte-identical to what they were seeded with.
+
+The premise was sound but the conclusion was too strong. The body does not single the joint out,
+but the body *plus the joint's position among that body's joints* does: MuJoCo stores a body's
+joints contiguously from `body_jntadr` in declaration order, so a third key form
+`("body_joint", body_name, ordinal)` identifies one joint of a body that carries several, and
+`_resolve_joint_key` resolves it back through the same two values. Nothing shifts an ordinal,
+because no scene op inserts a joint into an existing body - the patch vocabulary is `add_body`,
+`add_geom`, `add_site`, `set_body_pos`, `set_body_quat` and `delete_body`.
+
+Two joints of one body are what makes the ordinal load-bearing rather than decorative. A hinge and
+a slide both use width 1/1, so the width check `_restore_scene_state` already performs cannot tell
+them apart; keying both on the body alone would restore each from the other's values, and that
+swap is pinned.
+
+Everything else is unchanged and pinned as such. A named joint keeps its `("joint", name)` key. An
+unnamed free joint keeps its `("body", body_name)` key, which is unambiguous because the compiler
+refuses a free joint alongside any other joint on the same body. A joint whose body is *also*
+unnamed still gets no key: nothing identifies either end across the rebuild, so it is reported
+rather than guessed at. An ordinal past the body's joint count, a body that has vanished, and a
+joint that has since gained a name all resolve to nothing, so one joint is never restored from two
+entries.
--- a/strands_robots/simulation/mujoco/scene_ops.py
+++ b/strands_robots/simulation/mujoco/scene_ops.py
@@ -1614,7 +1614,12 @@ def reposition_body_in_scene(
 # The two namespaces are kept apart in the key because MuJoCo's joint and body
 # names are independent: a joint may be named exactly like some body, so a
 # single flat string key could collide.
-_JointKey = tuple[str, str]
+#
+# Any other unnamed joint takes the three-element form
+# ``("body_joint", body_name, ordinal)``: the body alone does not single it out
+# because a body may carry several joints, but the position among them does.
+# See :func:`_joint_key`.
+_JointKey = tuple[str, str] | tuple[str, str, int]
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Fixes a silent state-loss bug in the eject/rebuild path: `_joint_key` returned `None` for any unnamed non-free joint, so `_snapshot_scene_state` dropped its `qpos`/`qvel`/`qfrc_applied` and `remove_robot` reset the rest of the scene's articulation while reporting `success`. The change adds a third key form `("body_joint", body_name, ordinal)` keyed on `body_jntadr` contiguity, with `_resolve_joint_key` guarding the three ways the handle could over-claim (vanished body, out-of-range ordinal, and a joint that has since gained a name and is already carried by its own `("joint", name)` key). Both key consumers benefit — the eject snapshot and the grow path's `_snapshot_joint_forces` share `_joint_key`/`_resolve_joint_key`, so the fix is applied once at the seam rather than per caller. All key forms are private and in-memory only (no persistence, no wire format), so no one-way door. Verified locally at the head SHA: the 17 new tests pass; reverting only `scene_ops.py` to the merge base makes 10 of them fail (the fix fires); the two sibling suites exercising the same snapshot/restore path (`test_scene_ops_guardrails.py`, `test_remove_object_camera_cascade.py`) pass 54/54 on the branch; ruff clean.

## What's good

- The two-joints-on-one-body case (hinge + slide, both width 1/1) is exactly the case that makes the ordinal load-bearing, and it is pinned directly — a body-only key would pass the downstream width check and swap the values silently.
- The double-restore hazard (a joint carried by both its name key and a stale ordinal key) is guarded in `_resolve_joint_key` and pinned by `test_a_joint_that_gained_a_name_is_left_to_its_own_key`.
- The unchanged behaviours are pinned as unchanged: named joints, unnamed free joints, and the deliberately-unmatched joint on an unnamed body.
- Enum comparisons keep the value-based form the repo's `test_mujoco_enum_comparisons_are_value_based.py` grades (`int(...)` on the left).
- Changelog fragment under `changelog.d/`, behaviour-named test classes, GL-free tests (`mesh=False`), no host paths, ASCII-only strings — AGENTS.md conventions followed throughout.

yinsong1986 (approved):
## Summary

Fixes a silent data-loss bug in the eject/rebuild path: `_joint_key` returned `None` for any unnamed non-free joint, so `_snapshot_scene_state` dropped its `qpos`/`qvel`/`qfrc_applied` and `remove_robot` reset the surviving scene's articulation while reporting `success`. The change adds a third private key form `("body_joint", body_name, ordinal)` built on `body_jntadr` contiguity, and `_resolve_joint_key` refuses the three ways the handle could over-claim: a vanished body, an ordinal past the body's joint count, and a joint that has since gained a name (which is already carried by its own `("joint", name)` key, so refusing it prevents a double restore). A joint-type change at the same ordinal degrades to the pre-existing width-check skip in both consumers (`_restore_scene_state` and the grow path's `_restore_joint_forces`) rather than a mismatched write, and the fix lands once at the shared `_joint_key`/`_resolve_joint_key` seam so both the eject snapshot and the grow path's force carry benefit. `_JointKey` is private and in-memory only -- no persistence, no wire format -- so widening it is not a one-way door.

## What's good

- The two-unnamed-joints-on-one-body case (hinge + slide, both width 1/1) is exactly what makes the ordinal load-bearing, and it is pinned directly -- a body-only key would pass the width check and swap the values silently.
- The unchanged behaviours are pinned as unchanged: named joints, unnamed free joints, and the deliberately-unmatched joint on an unnamed body.
- Enum comparisons keep the value-based spelling `tests/test_mujoco_enum_comparisons_are_value_based.py` grades (the enum never sits on the left).
- Changelog fragment under `changelog.d/`, behaviour-named test classes, GL-free tests (`mesh=False`), ASCII-only strings, no host paths -- AGENTS.md conventions followed throughout.
- Round 1's `mypy` finding is resolved at this head: the fixture and the named-joint control narrow `sim._world` before reading through it, with no change to the behaviour under test.

cagataycali:
Verified `fa8e8ea` locally against the failing check, and separately checked that the narrowing did not cost the file any of its detection power.

### What was failing

`hatch run lint` only. `ruff check` and `ruff format --check` were already clean; the six errors were all `mypy`, all in the new test file, and all the same shape:

```
tests/simulation/mujoco/test_unnamed_joint_state_survives_a_rebuild.py:146: error: Item "None" of "SimWorld | None" has no attribute "_model"  [union-attr]
... 163 (_data), 319, 321 (_model, _data), 323 (_model)
```

`MuJoCoSimEngine._world` is declared `SimWorld | None`, so the private reads need narrowing. The reason only six of the file's many `_model` reads were flagged is that the rest go through helpers typed `world: Any` / `sim: Any`; the flagged sites are the two places where a concretely typed `Simulation` is in scope (the `scene` fixture and the named-joint control), which construct their own `Simulation`.

### The fix clears it

Running the lint target verbatim on `fa8e8ea`:

| step | result |
|---|---|
| `ruff check strands_robots tests tests_integ` | All checks passed |
| `ruff format --check ...` | 1544 files already formatted |
| `mypy strands_robots tests tests_integ` | Success: no issues found in 1541 source files |

`pytest tests/simulation/mujoco/test_unnamed_joint_state_survives_a_rebuild.py` gives 17 passed.

### It does not weaken the test

A narrowing assert is cheap to add in a way that quietly changes what an assertion reads, so I checked the detection power rather than the diff:

| check | result |
|---|---|
| `scene_ops.py` reverted to `main`, file re-run | 10 failed, 7 passed, same as before the narrowing |
| the rewritten control among those 10 | no, absent, which is correct for an over-reach control |
| mutation: `_joint_key` returns `None` for a named joint | the rewritten assertion fires, `assert (0.0, 0.0) == (0.55, -0.9)` |

So the pre-fix signal is unchanged, the control still passes on both sides as a control should, and the one assertion whose text changed still detects a dropped named-joint key.

### On the `rebuilt = sim._world` re-read

Line 323 was the only site where narrowing is not purely mechanical, because it reads after `remove_robot` has rebuilt the model. Worth recording what the rebuild actually does, measured on a two-robot scene:

```
world  object same  : True
model  object same  : False
data   object same  : False
njnt before/after   : 2 1
```

`install_compiled_model` assigns `world._model` and `world._data` on the existing `SimWorld` rather than replacing the world, so re-reading `sim._world` and reusing the handle taken before the eject resolve to the same object and therefore to the same rebuilt model. The two spellings are equivalent today, and the re-read is the better of the two: it does not depend on that invariant, so it stays correct if a future rebuild ever does replace the world. Reusing the earlier handle would need a comment pinning the invariant to be safe to read.

### Unrelated failures seen in the wider subtree

`tests/simulation/` gave two failures alongside this, both accounted for and neither attributable to this branch:

- `newton/test_multi_camera.py::TestNamedCameraRegistration::test_bad_position_shape_rejected` reproduces identically on pristine `main` (it asserts the substring `3 elements` against the message `'position' must be a 3-element vector`), so it is pre-existing.
- `test_rendering_pacers_survive_a_clock_step.py::...[forward_1h]` failed only inside the full 943 s subtree run and passes 16/16 in isolation on this branch. Its premise counts frames captured against wall-clock time (`assert 6 >= 16 // 2`), so it is load sensitive rather than a behaviour change.

PR #2726, opened 2026-08-25
test(policies): drive the fourth by-name provider the robot_state_keys contract classified but never measured

cagataycali:
## The gap

`tests/policies/test_state_key_name_list_contract.py` splits every
`set_robot_state_keys` surface into buckets. Nine own the shared
`name_list_error` domain; four are classified `_TOTAL_BY_MEMBERSHIP` - already
total without it, because every G1 joint they drive is resolved by name inside
the caller's list, so a bare string, a mapping, a one-shot iterator and
non-string or blank entries all fail that membership check instead. The file
states the standard it holds that bucket to:

> All three are therefore classified as already-total rather than exempted, and
> each claim is pinned behaviourally so the classification cannot hide a silent
> accept.

Three, for four entries. `ProtoMotionsPolicy` was added to the set by the pull
request that introduced the provider, as a one-line diff to that file and
nothing else; the prose, the section header (`The two already-total providers`)
and the section that drives them all still described the other three.

The mechanism is worth naming, because it is reusable. `test_every_surface_is_classified`
is a good non-vacuity guard - a new provider cannot land unnoticed - but it
demands *classification*, not *demonstration*, and one of the buckets carried no
behavioural obligation. So the cheapest way to satisfy CI for a new provider was
one line in `_TOTAL_BY_MEMBERSHIP`, which asserts totality rather than measuring
it. The domain-owning half already had the guard that prevents this
(`test_the_behavioural_table_covers_every_surface_that_owns_the_check`, derived
from `_MUST_VALIDATE`); the membership half did not.

## What was unmeasured

Measured over all of `tests/policies` on `main`, 4499 passing on the unmodified
tree:

| mutation of `ProtoMotionsPolicy` | failing tests on `main` |
|---|---|
| membership check deleted outright | **1** (`test_set_robot_state_keys_refuses_missing_joints` - a well-formed partial list of 20 real joint names) |
| setter falls back to the canonical joint order for any shape that is not a joint list | **0** |
| emitted action dict keyed off the caller's list instead of `config.joint_names` | **2** |

The middle row is the property the classification asserts, and nothing measured
it: none of the five malformed shapes was refused by any test. The first row
shows what *was* pinned - the "wrong robot" case, which is a different claim.

The third row is 2 rather than 0 only because #2724 landed while this was in
progress; both of its failing tests resolve a **reordered list of exactly the 29
joints** and assert the resolved values, so neither reads the emitted key set,
and the sim layout is not that list - it prepends the free floating-base joint,
so a caller list wider than `num_dofs` was still unmeasured. The identical
mutation on `KimodoPolicy` fails its width pin, which is the control: the
difference between the two providers was the pin, not the property.

## The change

Test-only. No `strands_robots/` source is touched, and the classification itself
is unchanged - it was correct, just undemonstrated.

1. The by-name section is now a table, `_MEMBERSHIP_SURFACES`, held against
   `_TOTAL_BY_MEMBERSHIP` by `test_the_membership_table_covers_every_already_total_surface`.
   Classifying a provider as already-total now carries a behavioural obligation
   instead of exempting it from one, mirroring the owning half.
2. All four providers are driven over the same grid: four malformed shapes, a
   one-shot iterator, a refusal that binds nothing, and the layout each one
   resolves (the over-reach control). That replaces three near-identical
   per-provider parametrizations with one, and gives `WBCPolicy` and
   `MotionBricksPolicy` the refusal-binds-nothing pin only `KimodoPolicy` had.
3. `ProtoMotionsPolicy` gains the duplicate-name pin, strengthened for the path
   that reads the caller's list. `KimodoPolicy` ignores the observation, so its
   duplicate test is about the emitted keys alone; the tracker resolves each
   joint's value by position inside a flat `observation.state`, so a repeat has
   two ways to go wrong. The test asserts both: the emitted key set stays the
   canonical 29, and the value handed to the tracker comes from the **first**
   occurrence. The state vector holds its own slot index at every position, so
   the value names the slot it was read from, and the duplicate's slot carries a
   sentinel no first occurrence holds.
4. The prose, the section header and the classification comment are corrected to
   four, with the measurements above recorded beside them.

## Break table

Every row is one mutation, run over all of `tests/policies`. Control 0; nine
mutations, nine distinct firing sets.

| # | mutation | `main` | this branch |
|---|---|---|---|
| b0 | none (control) | 0 | 0 (4514 passed) |
| b1 | ProtoMotions membership check deleted | 1 | **7** |
| b2 | ProtoMotions setter lenient for a non-joint-list shape | 0 | **2** |
| b3 | ProtoMotions action dict keyed off the caller's list | 2 | **3** |
| b4 | Kimodo action dict keyed off the caller's list | 1 | 1 |
| b5 | MotionBricks membership check deleted | - | 7 |
| b6 | WBC membership check deleted | - | 9 |
| b7 | ProtoMotions dropped from `_MEMBERSHIP_SURFACES` | - | 1 |
| b8 | a fifth provider planted, unclassified | 1 | 1 |
| b9 | a fifth provider planted, declared already-total in one line | **0** | **1** |

b9 is the row that closes the mechanism: it replays exactly what happened here -
plant a fifth by-name provider that checks nothing, then satisfy
`test_every_surface_is_classified` with a single line in
`_TOTAL_BY_MEMBERSHIP`. On `main` that is accepted silently. Here it fails,
naming the surface that was classified but not driven.

b2 fires 2 of the 4 shapes rather than all 4, because a bare string, a mapping
and an iterator all yield entries that are non-blank strings, so that particular
lenient branch still raises for them; the two it does silently accept are
exactly the two that fire. b1's 7 is the pre-existing partial-list test plus the
6 new ProtoMotions cells. b5/b6 also trip pre-existing per-provider tests in
`tests/policies/motionbricks` and `tests/policies/wbc`, which is the check that
the widened grid did not replace anything those suites already owned.

## Gate

* `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1543 files).
* mypy **24 == 24** against a pristine `upstream/main` worktree, normalized
  message sets byte-identical in both directions, 0 in the changed file.
* Full suite on this branch: `29 failed, 38302 passed, 309 skipped, 24 errors`
  in 1754s. The 29 + 24 are pre-existing on `main` and dependency-absent in this
  environment: 44 need `awsiot` (the `[mesh-iot]` extra), the rest are the
  lerobot-from-source and version-metadata failures already present. A paired
  full-suite run against the previous base gave **53 identical failing node ids
  on both trees, both `comm` directions empty**.
* `tests/policies` paired on this base, identical selection: **4499 passed on
  `main`, 4514 passed here** - `+15`, exactly the added cells, 0 failures either
  side.
* The changed file alone: 71 tests on `main`, 86 here.

No coverage claim: measured before and after over the same selection,
`policies/protomotions/policy.py` is at **90% / 20 missing on both** trees -
#2724's observation-convention suite already reaches the branch this test drives.
The value here is the pin, not the line count.

## No visual artifact

Test-only, no policy, simulation, rendering, recording or asset behaviour
changes, so there is nothing to render. The break table is the evidence.

### DIFF
--- a/changelog.d/2726-state-key-membership-obligation.md
+++ b/changelog.d/2726-state-key-membership-obligation.md
@@ -1,0 +1,22 @@
+### Tests
+
+- **policies**: the `robot_state_keys` contract file classifies four providers as
+  already total without the shared name-list domain - every G1 joint they drive
+  is resolved by name inside the caller's list, so a malformed shape fails that
+  membership check instead - and drove three of them. `ProtoMotionsPolicy` was
+  classified in a one-line addition to the exempt set by the pull request that
+  introduced the provider, and the file's own prose still said three, so the
+  property that classification asserts was never measured for it. Measured over
+  `tests/policies` beforehand, 4499 passing: making its setter fall back to the
+  canonical joint order for any shape that is not a joint list left 4499 passing
+  and 0 failing, and deleting the membership check outright failed exactly one
+  test, the well-formed partial list. None of the five malformed shapes was
+  refused by anything. The by-name section is now a table derived from that set,
+  the guard the domain-owning half already had, so classifying a provider as
+  already-total carries a behavioural obligation rather than exempting it from
+  one: all four are driven over the same four malformed shapes, a one-shot
+  iterator, a refusal that binds nothing, and the layout each resolves.
+  `ProtoMotionsPolicy` also gains the duplicate-name pin its sibling has,
+  strengthened for the path that reads the caller's list - the emitted key set
+  stays the canonical 29 and the value handed to the tracker comes from the
+  first occurrence.
--- a/tests/policies/test_state_key_name_list_contract.py
+++ b/tests/policies/test_state_key_name_list_contract.py
@@ -47,21 +47,48 @@ rate the in-process API refuses"), so ``MSG_SET_STATE_KEYS`` now forwards
 verbatim too. ``RemotePolicy`` validates before its own ``list(...)`` for the
 same reason, on the outbound side.
 
-Three providers needed no guard and deliberately did not get one.
-``WBCPolicy``, ``MotionBricksPolicy`` and ``KimodoPolicy`` resolve every G1 joint
-they drive BY NAME inside the caller's list, so a bare string, a mapping, a
-one-shot iterator, and non-string or blank entries all fail that membership
-check already - measured, all five refused with a message naming the missing
-joints. They also tolerate a repeated name on purpose: for the two
-index-resolved ones, ``test_flat_state_name_resolved_first_occurrence_wins``
-pins that a duplicate resolves to its FIRST occurrence and must not shift the
-resolved slot, which is a reviewed decision this change does not reopen.
-``KimodoPolicy`` is total for a second, stronger reason: it keys the emitted
-action dict off the canonical ``KIMODO_G1_JOINTS`` tuple rather than off the
-caller's list, so the width this file is about cannot be narrowed by a
-duplicate at all - pinned below. All three are therefore classified as
-already-total rather than exempted, and each claim is pinned behaviourally so
-the classification cannot hide a silent accept.
+Four providers needed no guard and deliberately did not get one.
+``WBCPolicy``, ``MotionBricksPolicy``, ``KimodoPolicy`` and
+``ProtoMotionsPolicy`` resolve every G1 joint they drive BY NAME inside the
+caller's list, so a bare string, a mapping, a one-shot iterator, and non-string
+or blank entries all fail that membership check already - measured, all five
+refused with a message naming the missing joints. They also tolerate a repeated
+name on purpose: where a slot is resolved by index,
+``test_flat_state_name_resolved_first_occurrence_wins`` pins that a duplicate
+resolves to its FIRST occurrence and must not shift the resolved slot, which is
+a reviewed decision this change does not reopen. ``KimodoPolicy`` and
+``ProtoMotionsPolicy`` are total for a second, stronger reason: each keys the
+emitted action dict off its canonical joint tuple rather than off the caller's
+list, so the width this file is about cannot be narrowed by a duplicate at all -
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Test-only change to `tests/policies/test_state_key_name_list_contract.py` plus a changelog fragment. It closes a real gap in the contract file's own mechanism: `test_every_surface_is_classified` demanded classification but the `_TOTAL_BY_MEMBERSHIP` bucket carried no behavioural obligation, so `ProtoMotionsPolicy` was classified as already-total in a one-line addition and never driven. The PR converts the by-name section into a `_MEMBERSHIP_SURFACES` table held against `_TOTAL_BY_MEMBERSHIP` (mirroring the guard the domain-owning half already had), drives all four providers over one shared grid (four malformed shapes, one-shot iterator, refusal-binds-nothing, over-reach control), and adds the duplicate-name pin for ProtoMotions on the path that reads the caller's list. No `strands_robots/` source is touched, no classification changes, and the b9 break-table row demonstrates the mechanism is actually closed. Verified locally at head `124c80dc`: the changed file runs 86 passed; the `ProtoMotionsPolicy(session=..., motion=...)` seam, `MotionPlayer`, and `GTP_G1_JOINT_NAMES` all exist as the new fixtures assume; imports (`asyncio`, `Callable`) are present at the file top; no non-ASCII, no host paths.

## What's good

- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention rather than editing `CHANGELOG.md` (AGENTS.md PR workflow step 3).
- Test names describe behaviour, not provenance (AGENTS.md > Testing > 'Name a test for its behaviour').
- The break table distinguishes b8/b9 (unclassified vs classified-but-undriven fifth provider), which is exactly the regression class this PR exists to close, and b5/b6 confirm the widened grid did not replace anything the per-provider suites already owned.
- The `_RecordingSession` seam avoids onnxruntime/CUDA while still measuring which slot the tracker was handed, so the first-occurrence pin asserts the value path and not just the emitted key set.

PR #2727, opened 2026-08-25
test(isaac): grade the orientation arms only a half turn or an unusable axis reaches

cagataycali:
## Problem

`_parse_orientation` in `strands_robots/simulation/isaac/loaders.py` resolves all five
MJCF orientation spellings, and two suites already grade it against MuJoCo's own
compiler. Every declaration those suites use, though, names a rotation of less than a
third of a turn and is well formed, which leaves two families of the reader's arms
reached by no fixture at all.

**The arms a rotation at or near a half turn takes.** `_quat_from_basis` builds the
quaternion from the trace while the trace is positive, and branches on whichever
diagonal term is largest once it is not - the case its own docstring names ("the
trace-only form degenerates for a rotation near 180 degrees, which is exactly what an
`xyaxes` flipping two axes declares"). Three of those four arms ran in no test. Each
assembles the quaternion from a different set of matrix entries with different signs,
so a sign or index slip in one of them reports a rotated link or object at a
*different* rotation while the load still reports success, and the value is still a
unit quaternion. A half turn is not exotic: it is what an `xyaxes` that flips two axes
states, which is how a model mounts a gripper or a camera facing backwards.

**The arms a declaration with no usable axis takes.** Too few or too many numbers for
the spelling, an axis of zero length, or an `xyaxes` whose y axis is parallel to its x
axis so nothing survives orthogonalizing. MuJoCo refuses every one of these models and
the reader tolerates them as identity, the reading it documents. Only a value that is
not a number at all was graded, and that takes the earlier `float()` failure rather
than any of these arms - so the guards that keep a zero-length axis from being divided
by were never run.

## What this adds

One test file, no source change. `tests/simulation/isaac/test_mjcf_orientation_half_turns_and_unusable_axes.py`:

- Eleven `xyaxes` rotations, chosen so each family carries a stated job - the exact
  half turns are the boundary where the trace term reaches zero, the half turn about
  the xy diagonal leaves two diagonal terms tied, the ten-degrees-short rotations give
  the real part a small non-zero value, the skew rotations put all four components at
  distinct non-zero magnitudes, and the quarter turn keeps the trace form in the set.
- A premise class that asserts rather than assumes: the fixtures span all four cases a
  matrix-to-quaternion conversion branches on (derived from the rotation MuJoCo
  compiled, not from the reader), MuJoCo compiles each rotation, each is far enough
  from identity that an identity fallback cannot pass by coincidence, and MuJoCo
  refuses each of the twelve unusable declarations.
- Both readers of the format driven on every rotation, `load_mjcf_scene_objects` and
  `load_mjcf`, plus the reported orientation's unit norm graded on its own.
- The antipodal `zaxis` case, exact and within tolerance: the reader special-cases
  `-z` with a comment claiming MuJoCo answers a half turn about x, which nothing
  checked. It does.

Every expectation for a model MuJoCo compiles is read off `body_quat`; nothing is
restated by hand. The twelve declarations MuJoCo refuses have no compiled answer, so
they are graded against the tolerated reading with the refusal asserted as the premise.

## What the arms carry

The reader is correct today - measured over 1500 random rotations against MuJoCo,
worst disagreement 5.55e-16, all four arms exercised. The figure is one declared
rotation read three ways: as MuJoCo compiles it, as the reader reports it (byte-identical),
and with a single sign changed in the arm that rotation takes.

One MJCF rotation, three readings

The slipped reading is 20 degrees off, turned the other way about the same axis, still
a unit quaternion, still a successful load.

## Verification

**Mutation table.** Baseline `tests/simulation/isaac` on `main`: 1431 passing. Each row
mutates `loaders.py`, runs the new file and the whole rest of that directory, then
restores the source (verified byte-identical, md5 `8e9414d7f8c68bcbd8e365e0d5b334e8`).

| mutation | new file | rest of tests/simulation/isaac |
|---|---|---|
| control, no mutation | 0 | 0 |
| x-dominant arm: real part sign flipped | 4 | 0 |
| y-dominant arm: real part sign flipped | 4 | 0 |
| z-dominant arm: real part sign flipped | 4 | 0 |
| x-dominant arm: wrong matrix entry | 6 | 0 |
| y-dominant arm: wrong matrix entry | 3 | 0 |
| z-dominant arm: wrong matrix entry | 3 | 0 |
| branching dropped, trace form alone | 12 | 0 |
| antipodal `zaxis` special case dropped | 2 | 0 |
| `axisangle` zero-axis guard dropped | 1 | 0 |
| `euler` arity/sequence guard dropped | 4 | 0 |
| `axisangle` arity guard dropped | 1 | 0 |
| `zaxis` arity guard dropped | 1 | 0 |
| `xyaxes` arity guard dropped | 1 | 0 |
| `xyaxes` zero-x guard dropped | 1 | 0 |
| `xyaxes` zero-y guard dropped | 3 | 0 |
| x-arm tie-break widened to `>=` | 0 | 0 |
| y-arm tie-break widened to `>=` | 0 | 0 |

Fifteen mutations, fifteen distinct failing sets, control clean, and nothing in the
existing 1431 tests notices any of them. The two tie-break rows are detected by
nothing, correctly: at a tie either arm yields the same rotation, so no test can
distinguish them and none should - recorded rather than papered over.

Which fixture does the work is measurable too. At an exact half turn the real part is
zero, so a sign flip on it is invisible; the three sign-flip rows are caught only by
the ten-degrees-short and skew rotations. The x-arm wrong-entry row is the one that
also trips the unit-norm assertion, which is why norm is graded separately.

**Coverage.** `strands_robots/simulation/isaac/loaders.py` 87% -> 90%, 89 uncovered
lines to 73. What remains is the `pxr`-gated `load_usd` block, one line unreachable
from its only caller, and the mesh-registry paths, none of them this subject.

**Gate.** Full `tests/` on this branch: 29 failed, 38371 passed, 309 skipped, 24 errors
(38733 collected), against 29 failed, 38287 passed, 309 skipped, 24 errors (38649) on
the base - +84, exactly the new tests, with the 54 failing/error ids identical both ways
(44 need `awsiot`, the rest pre-existing environment gaps). `ruff check` and
`ruff format --check` clean over `strands_robots tests tests_integ`; mypy 24 errors on
both trees with byte-identical message sets and none in the new file. The new file passes
on MuJoCo 3.5.0 (the declared floor) and on 3.12.0.

### DIFF
--- a/changelog.d/2727-mjcf-orientation-half-turn-arms.md
+++ b/changelog.d/2727-mjcf-orientation-half-turn-arms.md
@@ -1,0 +1,28 @@
+### Tests
+
+- **simulation/isaac**: MJCF's five orientation spellings are graded against MuJoCo's
+  compiler, but every declaration those suites use names a rotation of less than a third
+  of a turn and is well formed, so two families of arms in the reader were reached by no
+  fixture. Converting a rotation matrix to a quaternion from its trace alone degenerates
+  at a half turn - the term it divides by reaches zero - so the reader branches on
+  whichever diagonal term is largest instead, which is the case its own docstring says
+  the branching exists for, and three of those four arms ran in no test. The arms a
+  declaration with the wrong number of components or a zero-length axis takes were unrun
+  for a different reason: only a value that is not a number at all was graded, and that
+  takes the earlier `float()` failure rather than any of them. Measured over
+  `tests/simulation/isaac` beforehand, 1431 passing: flipping the sign of the real part
+  in any one of the three arms, reading a wrong matrix entry into a component, dropping
+  the branching in favour of the trace form alone, dropping the antipodal-`zaxis` special
+  case, and dropping any of the seven guards that keep a wrong-arity or zero-length axis
+  from being divided by, each left all 1431 passing. The new file detects all fifteen, in
+  fifteen distinct failing sets, with an unmutated control clean, and it says which
+  fixture carries which weight: at an exact half turn the real part is zero so a sign
+  flip there is invisible, which is what the ten-degrees-short rotations are for, while
+  the skew rotations put all four components at distinct non-zero magnitudes so a wrong
+  entry shows up. Widening a tie-break from `>` to `>=` is detected by nothing, which is
+  correct - at a tie either arm yields the same rotation - and is recorded rather than
+  papered over. Both readers of the format are driven, `load_mjcf_scene_objects` and
+  `load_mjcf`, and every expectation for a model MuJoCo compiles is read off `body_quat`
+  rather than restated by hand, with the refusal asserted as the premise for the models
+  it rejects. `strands_robots/simulation/isaac/loaders.py` goes from 87% to 90% covered,
+  89 uncovered lines to 73.
--- a/tests/simulation/isaac/test_mjcf_orientation_half_turns_and_unusable_axes.py
+++ b/tests/simulation/isaac/test_mjcf_orientation_half_turns_and_unusable_axes.py
@@ -1,0 +1,332 @@
+"""The orientation arms only a half turn, or an axis the reader cannot use, reaches.
+
+MJCF's five orientation spellings are already graded against MuJoCo's compiler,
+but every declaration those suites use names a rotation of less than a third of a
+turn and is well formed. Two families of the reader's arms are therefore reached
+by no fixture.
+
+An ``xyaxes`` whose rotation is at or near a half turn is the first. Converting a
+rotation matrix to a quaternion from its trace alone degenerates there -- the
+term under the square root goes to zero -- so the conversion branches on
+whichever diagonal term is largest instead, which is the case its own docstring
+says the branching exists for. Three of those four arms ran in no test, and each
+assembles the quaternion from a different set of matrix entries with different
+signs, so a sign or index slip in one of them reports a rotated link or object at
+a *different* rotation while the load still reports success. A half turn is not
+an exotic declaration: it is what an ``xyaxes`` that flips two axes states, which
+is how a model mounts a gripper or a camera facing backwards.
+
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Test-only PR: adds `tests/simulation/isaac/test_mjcf_orientation_half_turns_and_unusable_axes.py` (332 lines, 84 tests) plus the matching changelog fragment. The new suite grades the two previously-unreached families of arms in the MJCF orientation reader: the three non-trace branches of the Shepperd rotation-to-quaternion conversion in `_quat_from_basis` (reached only by rotations at or near a half turn) and the identity-tolerated arms for declarations MuJoCo refuses (wrong arity, zero-length or degenerate axes). Every expectation for a compilable model is read off `mujoco.MjModel.body_quat` rather than hand-stated, refusals are asserted as premises, and a dedicated premise class verifies the fixture set actually reaches all four conversion arms so none is graded vacuously. Both public readers (`load_mjcf_scene_objects` and `load_mjcf`) are driven.

## What's good

- Full suite passes against the head commit (84 passed), and a spot-check mutation (sign flip in the x-diagonal arm of `_quat_from_basis`) is caught by 4 tests exactly as the changelog claims - the detection claims are not vacuous.
- Oracle-based expectations (compiler-derived `body_quat`) rather than restated quaternions, with `q`/`-q` equivalence handled and unit-norm graded separately from rotation agreement.
- AGENTS.md compliance throughout: changelog fragment (not a `CHANGELOG.md` append), behaviour-describing test names, ASCII-only strings, `tmp_path`-rooted fixtures, no host paths, and the `importorskip("mujoco")` ordering is sound since `loaders.py` has no module-scope mujoco import.

One non-blocking note: the PR body is still the "Placeholder" line; worth replacing with a real description (no closing keyword needed since the title claims no issue).

PR #2729, opened 2026-08-25
fix(mesh): report which robot is running a policy on the state topic

cagataycali:
## Problem

`Mesh._read_state` publishes a `robots` section naming every robot in the sim world, and the flag beside each name was the literal `True`:

```python
snapshot["robots"] = {name: {"active": True} for name in world_robots}
```

So the flag was constant for the life of the peer. It did not change when a policy started, it did not change when one stopped, and a scene's idle arms were indistinguishable from the one arm executing a rollout. Measured on a two-arm MuJoCo world (headless) with a mock rollout on `so100` only, reading both surfaces of the same peer in all three phases:

| phase | `state` topic, `main` | `status` command, `main` |
|---|---|---|
| at rest | `so100 true`, `arm_b true` | `idle`, `robots_running []` |
| rollout on `so100` | `so100 true`, `arm_b true` | `running`, `robots_running ["so100"]` |
| after stop | `so100 true`, `arm_b true` | `idle`, `robots_running []` |

Two surfaces of one peer disagreed about one fact, and the 10 Hz telemetry topic was the wrong one. The `status` command is right because it reads the running-policy registry; the state topic asserted rather than measured.

The disagreement was already known when the `status` path was written. That code's own comment says sims "answered `{"status": "unknown"}` and their state topic hardcoded active=True - a running sim policy was invisible on the wire", so the state topic was named as part of the defect and only the command half was repaired.

## Why nothing caught it

No test in `tests/mesh/` asserted on the flag at all. The one test that reaches the section, `test_mesh.py::test_publishes_sim_clock`, asserts its key set and not a value:

```python
assert sorted(s["robots"].keys()) == ["arm0", "arm1"]
```

which is exactly the shape that lets a constant survive indefinitely.

## Change

Read the flag from the registry the `status` command already reads, through one helper:

```python
def _running_policy_robots(self) -> frozenset[str]:
    for holder in (self.robot, getattr(self.robot, "_sim_parent", None)):
        probe = getattr(holder, "_active_policy_robots", None)
        if callable(probe):
            return frozenset(probe())
    return frozenset()
```

A `SimRobot` child peer keeps no registry of its own, so it consults the parent `Simulation` through the `_sim_parent` backref - the same backref `_dispatch` already uses to delegate `execute`/`start`, installed by `Simulation._attach_robot_to_mesh` alongside the `_world` reference that makes a child publish a state topic at all. Both peer kinds now report the identical map. After the change, on the same scene:

| phase | `state` topic | `status` command |
|---|---|---|
| at rest | `so100 false`, `arm_b false` | `idle`, `robots_running []` |
| rollout on `so100` | `so100 true`, `arm_b false` | `running`, `robots_running ["so100"]` |
| after stop | `so100 false`, `arm_b false` | `idle`, `robots_running []` |

### Two deliberate dispositions

A peer that keeps **no** such registry reports every robot `false` - nothing runs a policy on those robots through the simulation API - and its `robots` section is still published. A registry that **raises** is left to reach `_read_state`'s section handler, which names `sim_world` in `degraded`: the documented mechanism for a probe that cannot answer. That is the opposite of `status`, which swallows because a command must answer. Substituting a flag on the telemetry path is what would make the fault unreportable, and it is how the unmeasured `True` survived.

state topic active flag

The flag distinguishes a rollout from motion, not motion from stillness - which is precisely why it is read from the registry rather than from the world. In the rendered scene `so100` moves **0.4585 rad** under its policy (14717 px of its own body change) while `arm_b` moves **0.0320 rad** (5053 px): it is sagging under gravity, not executing anything. A motion heuristic would have called both active. Both trees ran the same rollout - identical per-robot changed pixels and the same `status` answer - so only the reported flag differs.

## Tests

`tests/mesh/test_state_topic_reports_which_robot_runs_a_policy.py`, 11 cases. Pre-fix, with the source reverted and the tests kept: **9 failed, 2 passed**, the headline being `assert {'arm_b', 'so100'} == {'so100'}` - the idle bystander reported active during another arm's rollout. The 2 that pass pre-fix are the premise (the scene really holds two robots, so "exactly one is active" is not true by arity) and the over-reach control (deriving the flag must not cost the `robots` section), so no premise or control case is a regression case.

Mutation coverage on the fixed tree, each row the count of cases in the new file that fail:

| mutation | fires |
|---|---|
| control, unchanged | 0 |
| the flag back to the literal `True` | 9 |
| registry never consulted | 6 |
| the `_sim_parent` fallback dropped | 1 |
| only `_sim_parent` consulted, `self` dropped | 6 |
| a raising registry swallowed instead of reported | 2 |
| a raising registry answered with the old `True` | 2 |
| the flag inverted | 7 |
| the holder order swapped | 0 |
| the `callable()` guard relaxed to a presence check | 1 |

Nine of ten mutations are detected, six distinct firing sets, control clean, source restored byte-identical. The holder order fires nothing for a structural reason worth stating: no holder in the chain carries both ends, since a `SimRobot` has no `_active_policy_robots` of its own and a `Simulation` has no `_sim_parent`, so the order is immaterial rather than untested.

## Verification

- Paired run over an identical 572-file selection (all of `tests/mesh/` plus every test that walks the tree, reads docstrings or reads `docs/`), the new file excluded from both trees: **22931 collected** and **24 failed, 22766 passed, 117 skipped, 24 errors on both**, 48 identical failing ids, `comm` empty in both directions. All 48 were already failing on pristine `main`; 44 need `awsiot`/`awscrt` (declared in the `mesh-iot` extra, both absent here) and the rest are pre-existing.
- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`. `mypy` **5 == 5** against a pristine `main` worktree, normalized message sets byte-identical in both directions, none in the changed files.
- The new file: **11 passed** on mujoco 3.5.0 (the declared floor) with lerobot 0.6.2 from source, and on mujoco 3.12.0.
- The pre-existing `tests/mesh/` suite: 3620 passed, 10 skipped. Docs, xref and changelog guards: 1015 passed.

## Docs

`docs/mesh.md` documented neither the section nor the flag - the published-topics table listed the state topic's content as "joints, sim time, task status, degraded probes". A new "Which robot is running" subsection documents what `active` means, that it shares its source with `robots_running`, that which robots *exist* is the presence topic's `sim_robots`, and both boundary dispositions above.

### DIFF
--- a/changelog.d/2729-state-topic-active-flag.md
+++ b/changelog.d/2729-state-topic-active-flag.md
@@ -1,0 +1,49 @@
+### Fixed: a sim peer's state topic reports which robot is running a policy
+
+`Mesh._read_state` publishes a `robots` section naming every robot in the sim world, and the flag
+beside each name was the literal `True`. So it was constant for the life of the peer: it did not
+change when a policy started, it did not change when one stopped, and a scene's idle arms were
+indistinguishable from the one arm executing a rollout. Measured on a two-arm MuJoCo world with a
+mock rollout on `so100` only, reading the topic in all three phases:
+
+    at rest            {"so100": {"active": true}, "arm_b": {"active": true}}
+    rollout on so100   {"so100": {"active": true}, "arm_b": {"active": true}}
+    after stop         {"so100": {"active": true}, "arm_b": {"active": true}}
+
+The `status` command answers the same question correctly over the same three phases -
+`robots_running` is `[]`, `["so100"]`, `[]` - because it reads the running-policy registry. Two
+surfaces of one peer therefore disagreed about one fact, and the 10 Hz telemetry topic was the one
+that was wrong. The flag is now read from that same registry, so the topic and an on-demand status
+answer always name the same set.
+
+That disagreement was already known when the `status` path was written. Its own comment says sims
+"answered `{"status": "unknown"}` and their state topic hardcoded active=True - a running sim
+policy was invisible on the wire", so the state topic was named as part of the defect and only the
+command half was repaired. Nothing graded the remaining half: no test in `tests/mesh/` asserted on
+the flag at all, and `test_mesh.py::test_publishes_sim_clock` asserts the section's key set
+(`sorted(s["robots"].keys()) == ["arm0", "arm1"]`) without reading a value, which is exactly the
+shape that lets a constant survive.
+
+A `SimRobot` child peer keeps no registry of its own, so it consults the parent `Simulation`
+through the `_sim_parent` backref - the same backref `_dispatch` already uses to delegate
+`execute`/`start`, installed by `Simulation._attach_robot_to_mesh` alongside the `_world`
+reference that makes the child publish a state topic in the first place. Both peer kinds now
+report the identical map.
+
+Two dispositions are deliberate. A peer that keeps no such registry reports every robot `false`:
+nothing runs a policy on those robots through the simulation API, so none of them is executing
+one, and its `robots` section is still published. A registry that raises is left to reach
+`_read_state`'s section handler, which names `sim_world` in `degraded` - the documented mechanism
+for a probe that cannot answer. That is the opposite of `status`, which swallows because a command
+must answer; substituting a flag on the telemetry path is what would make the fault unreportable,
+and it is how the unmeasured `True` survived in the first place.
+
+The flag distinguishes a rollout from motion rather than motion from stillness, which is why it
+is read from the registry and not from the world. In the measured scene `so100` moves 0.4585 rad
+under its policy while `arm_b` moves 0.0320 rad - it is sagging under gravity, not executing
+anything - and a motion heuristic would have called both active.
+
+`docs/mesh.md` documented neither the section nor the flag; the published-topics table listed the
+state topic's content as "joints, sim time, task status, degraded probes". It now documents what
+`active` means, that it shares its source with `robots_running`, that which robots *exist* is the
+presence topic's `sim_robots`, and both boundary dispositions above.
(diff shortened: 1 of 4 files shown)


yinsong1986 (approved):
## Summary

Replaces the hardcoded `active: True` in `Mesh._read_state`'s `robots` section with a measurement from the same running-policy registry the `status` command already reads, via a new `_running_policy_robots()` helper that consults `self.robot` and then the `_sim_parent` backref for `SimRobot` child peers. The two boundary dispositions are deliberate and correct for the telemetry path: a peer with no registry reports every robot `false` while keeping its section, and a raising registry is left to reach the existing `sim_world` section handler so the fault lands in `degraded` rather than being papered over with a fabricated flag - the exact failure mode that let the constant survive. The section's key set and shape are unchanged, the old flag was a constant carrying no information, and no in-repo consumer reads it, so this is a bug fix rather than a wire-format break; `docs/mesh.md` now defines the flag for the first time. Verified locally: the 11 new tests pass on mujoco 3.8.0, `tests/mesh/test_mesh.py` (46 passed) is unaffected, `ruff check` / `ruff format --check` are clean on the changed files, all added lines are ASCII, and the changelog fragment, no-host-paths, test-naming and ASCII guards all pass on the head tree.

## What's good

- The regression test suite is behaviour-named, drives a real MuJoCo rollout through all three phases, pins topic/`status` agreement with a non-vacuity check, and pins both boundary dispositions (no registry, data-attribute impostor, raising registry) with dependency-free stubs.
- The `callable()` guard keeps a data attribute that happens to carry the registry's name from being read as a rollout.
- Changelog fragment per step 3, `docs/mesh.md` updated in the same PR, and the mutation table in the description is honest about the one mutation that fires nothing and why.

PR #2730, opened 2026-08-25
fix(mesh): coerce a sensor payload's readings so the topic reaches the wire

cagataycali:
## Problem

`SensorLoopsMixin`'s readers merge the robot's provider mapping into the outgoing payload verbatim (`summary.update(data)`, `imu.update(imu_data)`, and six more), and every extended sensor topic is published as JSON -- `strands_robots.mesh.session._put_zenoh_directly` encodes the payload before it reaches the wire.

A payload the encoder refuses is not a transient failure the next tick retries. It fails identically forever, which is exactly why #2638 reports it at ERROR rather than absorbing it. The topic never publishes at all.

The values a sensor stack reports are numpy: a lidar summary's bounding box is whatever `ndarray.min(axis=0)` returned, an IMU's orientation a `float32`, a device state code an `int64`. Measured on `numpy 2.2.6`:

| provider value | `json.dumps` | subclasses a Python builtin |
|---|---|---|
| `np.float64` | encodes | `float` |
| `np.str_` | encodes | `str` |
| `np.float32` / `np.float16` | refused | no |
| `np.int32` / `np.int64` | refused | no |
| `np.bool_` | refused | no |
| `np.ndarray` | refused | no |

That asymmetry is why this survived: a robot on a float64 pipeline published fine, and the same code reading a `float32` point cloud dropped every tick.

### Reproduction

A robot exposing the documented provider attributes with its own readings, driven through the production readers and the encode `put` performs:

```
strands/neon/lidar/summary  DROPPED  Object of type ndarray is not JSON serializable
strands/neon/imu            DROPPED  Object of type ndarray is not JSON serializable
strands/neon/lidar/state    DROPPED  Object of type float32 is not JSON serializable
strands/neon/health         DROPPED  Object of type float32 is not JSON serializable
```

All four are gone from the mesh, with one ERROR line per topic to say so.

## Why the class already disagreed with itself

Three paths in this code coerce for exactly this reason:

- `_read_imu`'s inner-observation branch: `if hasattr(val, "tolist"): val = val.tolist()`
- `_read_pose`'s SE(3) branch: `pose["x"] = float(mat[0, 3])`
- `Mesh.publish_step`: `obs_numeric[key] = value.tolist()`

The eight readers that merge a provider mapping did not, so one class held two conventions for one wire format. The eight sites are `_read_pose` (x3, for `_pose`/`_slam_pose`/`_odom_pose`), `_read_imu`, `_read_odom`, `_read_lidar_summary`, `_read_lidar_state`, `_read_hands`, `_read_map_info`, plus `_read_health`'s two direct assignments of `_battery` and `_temps`.

## Change

Each reader hands its payload to one `_coerce_record` before returning it. Anything exposing `tolist()` becomes the equivalent Python list or number; lists, tuples and sets are rebuilt with their contents coerced, and mappings with their keys coerced too, because a `float32` is no more encodable as a key than as a value.

The coercion is applied entry by entry in place, so every reader keeps its declared `dict[str, Any] | None` return type and the diff adds no signature changes.

### The boundary

Coercion repairs readings; it does not launder payloads.

- A value that is **not** a reading -- an opaque object, a `bytes` blob -- is returned untouched, so it still reaches the encoder and #2638 still names it. Substituting something for it would publish a record that misreports what the sensor said.
- The recursion is **bounded**. A sensor record is shallow, so a structure deeper than the bound is a cycle or an object graph rather than a reading, and #2638's ERROR report is the right answer for those. Recursing into them here would replace a reported failure with one absorbed by the reader's own handler.
- Coercion never edits robot state. `_read_health` stores the provider's `_temps` mapping by reference, so the entry is replaced in the payload rather than rewritten underneath the robot. `_jsonable` never mutates; it returns a new container.

## Tests

`tests/mesh/test_sensor_payloads_are_json_encodable.py`, 36 cases, drives all eight readers and all three pose branches with the numpy a sensor stack reports and asserts on the **decoded** record a subscriber sees.

Every row carries a premise that its raw provider value is what the encoder refuses, so the table cannot pass on a payload that never needed coercing. The rule that each reader routes its payload through the coercion is derived from the module with `ast` rather than listed, so a reader added later is held to it on arrival, and a non-vacuity case asserts the discovery reaches all eight.

**Pre-fix split** (helpers present, all 12 call sites removed -- a raw revert is a collection error since the tests import the helpers):

| class | fail | role |
|---|---|---|
| `TestEveryReadingReachesTheWire` | 13 | regression |
| `TestTheCoercionDoesNotReachIntoRobotState` | 1 | regression |
| `TestTheCoercionIsSingleSourced` | 1 of 2 | the derived rule fires, its non-vacuity case passes |
| `TestAPlainPayloadIsUnchanged` | **0** of 5 | control |
| `TestAValueThatIsNotAReadingIsLeftForTheTransport` | **0** of 4 | over-reach boundary |

`15 failed, 21 passed`, and nothing in either control class.

**Break table** -- 13 mutations, 12 distinct firing sets, control clean, none undetected:

| row | mutation | fires |
|---|---|---|
| b0 | control (no mutation) | 0 |
| b1 | the coercion neutered | 14 |
| b2 | only `_read_lidar_summary` stops coercing | 4 |
| b3 | only `_read_health` stops coercing | 4 |
| b4 | only `_read_hands` stops coercing | 2 |
| b5 | a value that is not a reading is stringified instead of preserved | 2 |
| b6 | the container branch widened to `Sequence`, so a name is taken apart | 21 |
| b7 | nested mappings not recursed | 4 |
| b8 | mapping keys left uncoerced inside `_jsonable` | 1 |
| b9 | the list/tuple branch dropped | 1 |
| b10 | the depth bound removed | 1 |
| b11 | `tolist()` replaced by `float()`, which cannot render an array | 6 |
| b12 | the coercion recurses in place, editing the robot's own mapping | 2 |
| b13 | the payload's own keys left uncoerced | 1 |

b5 is worth naming: stringifying an unencodable value is the tempting alternative, and it is what the over-reach control refuses -- a record that reports `"<object at 0x...>"` where a reading belongs is a quieter wrong answer, not a fix. Source restored byte-identically after every row.

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1546 files).
- `mypy` **24 == 24** against a pristine `main` worktree, normalised message sets byte-identical both directions, 0 in the changed files.
- Paired run over an identical 539-file selection (all of `tests/mesh/`, plus every test that walks the tree, reads source or docstrings, or touches the encode path), with the new file excluded from both trees: **21727 collected and `23 failed, 21608 passed, 72 skipped, 24 errors` on both**, 47 identical failing/error ids, `comm` empty in both directions. 45 of the 47 need `awsiot` (`awsiotsdk`, declared in the `[mesh-iot]` extra, confirmed absent via `find_spec`); 2 are a pre-existing lerobot-from-source `encode()` signature drift in `tests/training/test_lerobot.py`.

## Artifact

The same script, same seed, run on both trees: the four topics a robot with float32 readings publishes.

sensor payload on the wire

The orange line in the left column is the report `_report_unencodable_payload` emits. It is correct and it is all that happened: the payload is still never published. This change is complementary to #2638 -- that one made the drop visible, this one stops the drop for the values that are readings.

### DIFF
--- a/changelog.d/2730-mesh-sensor-payload-json-encodable.md
+++ b/changelog.d/2730-mesh-sensor-payload-json-encodable.md
@@ -1,0 +1,51 @@
+### Fixed: a sensor payload's numpy readings are coerced, so the topic reaches the wire instead of being dropped
+
+`SensorLoopsMixin`'s readers merged the robot's provider mapping into the outgoing
+payload verbatim -- `summary.update(data)`, `imu.update(imu_data)`, and six more --
+and every extended sensor topic is published as JSON: `_put_zenoh_directly`
+encodes the payload before it reaches the wire. A payload the encoder refuses is
+not a transient failure that the next tick retries; it fails identically forever,
+which is why #2638 reports it at ERROR rather than absorbing it. The topic never
+publishes at all.
+
+The values a sensor stack reports are numpy. A lidar summary's bounding box is
+whatever `ndarray.min(axis=0)` returned, an IMU's orientation a `float32`, a
+device state code an `int64` -- and `json.dumps` refuses all three. So a robot
+exposing `_lidar_summary`, `_imu`, `_odom`, `_lidar_state`, `_map_info`, `_hands`,
+`_temps`, `_battery` or a `_pose`/`_slam_pose`/`_odom_pose` mapping built from its
+own readings had those topics silently absent from the mesh, with one ERROR line
+per topic to say so.
+
+The defect turned on the numpy *width*, which is why it survived. `np.float64`
+subclasses Python's `float` and `np.str_` subclasses `str`, so a payload built
+from those always encoded; `np.float32`, `np.float16`, `np.int32`, `np.int64`,
+`np.bool_` and `ndarray` subclass nothing and were refused. A robot on a float64
+pipeline published fine, and the same code reading a `float32` point cloud
+dropped every tick.
+
+Two paths in the same class already coerced for exactly this reason -- `_read_imu`'s
+inner-observation branch calls `tolist()`, `_read_pose`'s SE(3) branch calls
+`float()` -- and so does `Mesh.publish_step`. The eight readers that merge a
+provider mapping did not, so the class held two conventions for one wire format.
+Each reader now hands its payload to one `_coerce_record` before returning it:
+anything exposing `tolist()` becomes the equivalent Python list or number, and
+containers are rebuilt with their contents coerced, mapping keys included, because
+a `float32` is no more encodable as a key than as a value.
+
+The boundary is that coercion repairs readings rather than laundering payloads. A
+value that is not a reading -- an opaque object, a `bytes` blob -- is returned
+untouched, so it still reaches the encoder and is still reported by name; and the
+recursion is bounded, because a sensor record is shallow and a structure deeper
+than the bound is a cycle or an object graph rather than a reading, which #2638's
+report is the right answer for. Coercion also never edits robot state: `_read_health`
+stores the provider's `_temps` mapping by reference, so the entry is replaced in the
+payload rather than rewritten underneath the robot.
+
+`tests/mesh/test_sensor_payloads_are_json_encodable.py` drives all eight readers
+and all three pose branches with the numpy a sensor stack reports, asserting on the
+decoded record a subscriber sees. Each row carries a premise that its raw provider
+value is what the encoder refuses, so the table cannot pass on a payload that never
+needed coercing. Thirteen plausible regressions were applied and the pin re-run:
+twelve fire a different set, none goes undetected, and the control is clean. The
+rule that every reader routes its payload through the coercion is derived from the
+module rather than listed, so a reader added later is held to it on arrival.
--- a/strands_robots/mesh/sensors.py
+++ b/strands_robots/mesh/sensors.py
@@ -47,6 +47,91 @@ from strands_robots.mesh.session import (
 logger = logging.getLogger(__name__)
 
 
+# A sensor payload is a plain record on the wire, so the values in it have to be
+# ones ``json.dumps`` accepts. ``_JSONABLE_MAX_DEPTH`` bounds how far down this
+# module will go looking for them.
+#
+# Past the bound the value is handed to the encoder unchanged. That is
+# deliberate: a sensor record is shallow (a mapping of names to scalars and short
+# vectors), so a structure deeper than this is not a reading but a pathological
+# payload - a cycle, or a nested object graph. Those are exactly the payloads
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Routes every `SensorLoopsMixin` reader's outgoing payload through a single `_coerce_record` helper so provider mappings holding narrow numpy values (`float32`, `int64`, `ndarray`, and numpy mapping keys) encode as JSON instead of being refused by `_put_zenoh_directly`'s encoder on every tick. The coercion is deliberately bounded: only `tolist()`-bearing readings and containers are rebuilt, opaque objects and `bytes` are handed to the encoder untouched so the transport's ERROR report (#2638) still fires, recursion is depth-capped, and nested provider state the robot still owns is replaced in the payload rather than mutated. One note: the PR body is still the placeholder text ("full body follows in an update once the changelog fragment number is known") -- the fragment number is now known (2730), so the body is worth updating before merge, though the changelog fragment itself carries the full rationale.

## What's good

- Composed the branch's test delta against the merge base: `tests/mesh/` reports an identical pre-existing failure set on both trees (environment-related), and the PR's whole effect is +36 passes -- exactly the new suite.
- The new tests assert on the decoded record a subscriber sees, carry premise tests that each raw provider value is genuinely refused by the encoder (so the table cannot pass vacuously), and pin the every-reader rule via an AST scan over the module rather than a list -- a reader added later is graded on arrival.
- Boundary tests pin both directions: opaque values preserved by identity for the transport to report, and the robot's own `_temps` mapping verified unmutated (matching the concurrency posture in AGENTS.md).
- Changelog fragment present per step 3; ruff check/format clean on the changed files; no non-ASCII, no host paths; cross-referenced symbols (`_put_zenoh_directly`, `_report_unencodable_payload`) verified to exist at the head.
- No wire-format door: coerced values encode to the same JSON shapes a float64 pipeline already published, and previously-refused payloads never reached any consumer.

## Verification suggestions

- `python -m pytest tests/mesh/test_sensor_payloads_are_json_encodable.py -q --no-cov` -- 36 passed at `ecc63771`.
- Edge-case worth knowing (not blocking): a provider mapping keyed by a *tuple* makes `_jsonable` rebuild the key as a list, which raises `TypeError: unhashable type: 'list'` inside the reader -- swallowed by the reader's `except Exception`, so such a payload now drops with a debug-level tick log instead of the transport's ERROR report. That payload was equally unpublishable before this PR (`json.dumps` refuses tuple keys), so no working topic regresses; if tuple keys ever show up in a real provider, hashing the coerced key back to a tuple (or leaving unhashable coerced keys as-is) would restore the ERROR path.

PR #2731, opened 2026-08-25
fix(training): hold the RL run-size factors to the shared count domain

cagataycali:
## Problem

`RLTrainSpec.total_timesteps` and `rollout_steps` are the two caller-supplied factors of the one loop bound both RL trainers derive:

```python
steps_per_iter = max(1, self.steps_per_iter)              # rollout_steps * num_envs
num_iters = max(1, spec.total_timesteps // steps_per_iter)
for it in range(num_iters):                               # collect, update
```

Each was compared with a bare `value <= 0`. Every other numeric field in the same two `validate()` bodies already routes through a shared domain with a documented reason (`gamma`, `lam`, `num_learning_epochs`, `max_grad_norm`, `value_loss_coef`, `entropy_coef`, `clip_param`, `alpha_lr`, `init_alpha`); the size fields were the remainder.

That remainder matters more than it looks, because the bound is *derived* rather than read off the spec. The `max(1, ...)` clamp turns every value that survives the comparison but cannot divide into a silent single iteration. `run_size_problems`, the domain the supervised backends use for `steps` / `global_batch_size`, already spells out the failure modes verbatim - "`True` reads as a silent run of one step, a fractional or non-finite value reaches `range()` and raises there, and a string raises out of the comparison itself". All three were live here, on both trainers.

## Measured

Real MuJoCo sim, real `PpoTrainer` / `FastSacTrainer`, 16-step run with `rollout_steps=4`. Identical results for both providers:

| `total_timesteps` | before | after |
|---|---|---|
| `16` (control) | success, 4 iterations, `latest_step=16` | unchanged |
| `True` | **success, 1 iteration**, `latest_step=4` | refused |
| `0.5` | **success, 1 iteration**, `latest_step=4` | refused |
| `nan` | **success, 1 iteration**, `latest_step=4` | refused |
| `inf` | **success, 1 iteration**, `latest_step=4` | refused |
| `100.5` | `TypeError` from `range()`, after `setup` | refused |
| `"16"` / `None` | `TypeError` out of `validate` | refused |

`inf` lands in the silent column rather than the raising one because `inf // 4` is `nan` and `max(1, nan)` is `1`. So four of the five values that pass a `<= 0` test reported `status="success"`, wrote a checkpoint, and announced `"1 iterations x 4 steps complete"` for a run asked to be tens of thousands of steps long. The one value that does raise gave `range()` a float bound and raised only after `setup` had built the environment, the networks, the optimizers and - for FastSAC - the replay buffer.

`rollout_steps` fails the same three ways through the other factor, and its silent case changes the *shape* of the run rather than its length:

| `rollout_steps` | before (fast_sac) | before (ppo) | after |
|---|---|---|---|
| `True` | **success, 16 iterations x 1 step** | `ValueError` from `torch.distributions.Normal` on a tensor of `nan` | refused |
| `4.0` | `TypeError` from `range()` | `TypeError` from `range()` | refused |
| `nan` | `TypeError` from `range()` | refused, but as *indivisible* | refused as a non-count |
| `"4"` | `TypeError` out of `validate` | `TypeError` out of `validate` | refused |

`rollout_steps=True` makes `steps_per_iter` one, so PPO normalizes advantages over a length-one batch - whose standard deviation is `nan`, with `UserWarning: std(): degrees of freedom is <= 0` on the way - and dies inside a torch constraint naming neither the field nor the run.

## Change

Both factors go through the shared `positive_count_error` domain, via a new `rl_run_size_problems` / `Trainer._rl_run_size_problems` gate. This is a second gate rather than part of `run_size_problems` because the two field sets are disjoint: per `TrainSpec` a backend ignores the fields it does not support, so an RL trainer must not be refused for a `steps` it never reads and a supervised backend must not be refused for a `total_timesteps` it never reads. The *domain* is shared; only the fields differ.

`num_envs`, the third factor of `steps_per_iter`, deliberately stays per-backend: PPO parallelizes and accepts any count `>= 1`, the MuJoCo-backed FastSAC is single-env and requires exactly `1`, so it is not one shared domain. `TestNumEnvsIsNotInTheSharedDomain` pins that scope line, including that excluding it left it checked.

PPO's `rollout_steps % num_mini_batches` rule is a relation *between* two counts, so it now consults the same domain before dividing. It reads the domain directly rather than relying on the gate having run, so it holds wherever in the preflight it sits. That also fixes the misdiagnosis in the table above: `nan` was reported as indivisible when the problem is that it is not a count.

The strict-`int` domain is the right one and not merely the convenient one: it is what lets the guard refuse the integral float `4.0` that `range()` itself raises on. The cost is that `np.int64(24)` is refused although `range()` would accept it, which `test_a_numpy_integer_is_refused_although_range_would_accept_it` pins rather than leaves implied.

`num_mini_batches` keeps its own local check. It is a different quantity with a different consumer, and widening this domain to cover it would be a separate change.

## Tests

`tests/training/test_rl_run_size_domain.py`, 289 cases through the real `validate` / `train` entry points.

Pre-fix (call sites reverted to the inline comparisons, domain kept): **175 failed / 289**, with **0 failures in all three control classes** - the supervised-backend over-reach control (0/29), the `num_envs` scope line (0/4) and the clamp premise (0/10). A raw revert of all four source files is a collection error, since the domain function does not exist there.

11 mutations, control clean, source restored byte-identical:

| mutation | fires |
|---|---|
| control (fix in place) | 0 |
| gate dropped from `ppo` | 88 |
| gate dropped from `fast_sac` | 86 |
| domain widened to `positive_whole_number_error` | 129 |
| gate grades `total_timesteps` only | 105 |
| gate moved after the divisibility relation | 0 |
| divisibility relation ungated again | 14 |
| `num_envs` folded into the shared gate (over-reach) | 1 |
| `getattr` default made unusable | 1 |
| gate stops at the first bad factor | 1 |
| `max(1, ...)` clamp removed from the loop (premise) | 1 |

The zero is honest and worth keeping: moving the gate after the divisibility relation fires nothing precisely because the relation asks the domain itself, so the ordering is not load-bearing. The comment in `ppo.py` says that rather than implying a dependency that does not exist.

The guard is registered in `tests/training/test_spec_field_read_discovery.py`, which discovers field-scoped domain guards structurally and demanded triage of a new one - the intended behaviour. Its one-owner scan derives the backends in scope from the tree, and is scoped to modules that implement `validate`: `base_algo.py` reads both factors in the shared loop but has no preflight to route anything through, which `test_the_shared_loop_reads_the_fields_and_has_no_preflight_to_gate` pins so a future `validate` there is caught.

Five pre-existing assertions in `test_rl_ppo.py` / `test_rl_fast_sac.py` pinned the old wording (`"total_timesteps must be > 0"`); they now pin the shared domain's message. The refusal is unchanged, the text is the shared one.

## Gate

- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ`: clean, 1546 files.
- `mypy strands_robots tests tests_integ`: `Success: no issues found in 1543 source files`; branch-only message set empty against a pristine `main` worktree.
- `tests/training/`: 3213 passed, 4 skipped.
- Paired run over an identical 378-file selection (all of `tests/training/` plus every guard that walks the tree, reads docstrings, reads `docs/` or names the changed surfaces), the new file excluded from both trees: **7 failed / 17551 passed on the branch and 7 failed / 17549 passed on `main`, with identical failing-ID sets and both set differences empty**. The 7 are environment-dependent and fail identically on both trees (two need `docker` on `PATH`, five expect `rclpy` to be absent). The `+2` localizes entirely to `test_spec_field_read_discovery.py` (53 vs 51 cases) - the two registrations above.

No visual artifact: this changes a read-only preflight, not a policy, a scene, a render or a recording. The measured tables above are the evidence.


yinsong1986 (approved):
## Summary

Routes `RLTrainSpec.total_timesteps` and `rollout_steps` — the two caller-supplied factors of the derived loop bound `max(1, total_timesteps // (rollout_steps * num_envs))` — through the shared `positive_count_error` domain on both RL trainers, via a new `rl_run_size_problems` / `Trainer._rl_run_size_problems` gate, replacing bare `<= 0` comparisons that let `True`, `0.5`, `nan` and `inf` report a successful one-iteration run and let strings/`None` raise `TypeError` out of a `validate` documented to return problems. PPO's `rollout_steps % num_mini_batches` relation now consults the same domain before dividing, so a non-count is reported as a non-count rather than as indivisible (or as a `TypeError` through `%`). The scope lines are deliberate and pinned: `num_envs` stays per-backend because the two backends genuinely disagree about its accepted set, and the gate is a peer of `run_size_problems` rather than part of it because the field sets are disjoint. Verified locally at head `020696ae`: the 376 tests across the new domain file, both RL backend test files and the field-read discovery scan all pass, and `ruff check` / `ruff format --check` are clean on the touched modules. The tightening is loud (a refusal naming the field and backend), not a silent default change, so no migration concern; refusing `np.int64` where `range()` would accept it is the documented cost of the strict-int domain and is pinned by its own test.

## What's good

- The one-owner scan derives its backend set from the tree and pins the `base_algo.py` carve-out, so a third RL trainer or a future `validate` on the shared loop is caught structurally.
- Changelog fragment present (`changelog.d/2731-rl-run-size-domain.md`), no direct `CHANGELOG.md` edit, per AGENTS.md step 3.
- The new guard reuses the existing shared domain rather than adding a single-caller one, and the divisibility relation reads the domain directly so the gate ordering is not load-bearing — both consistent with AGENTS.md convention 11.
- Supervised-backend over-reach controls (`TestASupervisedBackendStaysQuiet`) pin the "a backend is not refused for a field it never reads" contract in both directions.

PR #2732, opened 2026-08-25
fix(vera): stop the LPIPS shim refusing a frame the real check accepts

cagataycali:
## Problem

VERA's DFoT metrics import `NoTrainLpips` and `_valid_img` from
`torchmetrics.image.lpip`. Modern torchmetrics keeps neither name there, so the
image installs `strands_robots/policies/vera/docker/sitecustomize_vera.py` as an
auto-imported `sitecustomize` (`Dockerfile:107`) and re-exposes both. VERA then
calls the shim's `_valid_img`, which makes the shim's verdict the metric's
verdict.

The real check, still present at `torchmetrics.functional.image.lpips`, is
deliberately asymmetric:

```python
def _valid_img(img: Tensor, normalize: bool) -> bool:
    """Check that input is a valid image to the network."""
    value_check = img.max() <= 1.0 and img.min() >= 0.0 if normalize else img.min() >= -1
    return img.ndim == 4 and img.shape[1] == 3 and value_check
```

`normalize=True` bounds both ends of `[0, 1]`; `normalize=False` bounds only the
**lower** end of `[-1, 1]`, because a frame already in the network's own range is
allowed to overshoot `1.0` - which a decoder emitting a tanh-shaped frame does
routinely, by a float step or two. The shim bounded the upper end on both
branches, and `_lpips_update` turns a `False` from this check into a
`ValueError`, so the extra bound did not make the metric stricter: it made an
eval that the unshimmed torchmetrics runs raise inside the image.

## Measured against real torchmetrics 1.9.0 (torch 2.11.0)

Same tensor, real `_valid_img` beside the shipped shim. 3 of 14
`(frame, normalize)` cells disagreed on `main`; 14 of 14 agree on this branch.

| frame range | `normalize` | torchmetrics | shim on `main` | this branch |
|---|---|---|---|---|
| `[0.0, 1.0]` | True | accept | accept | accept |
| `[0.0, 1.0]` | False | accept | accept | accept |
| `[0.2, 0.8]` | True | accept | accept | accept |
| `[0.2, 0.8]` | False | accept | accept | accept |
| `[-1.0, 1.0]` | True | refuse | refuse | refuse |
| `[-1.0, 1.0]` | False | accept | accept | accept |
| `[-1.0, 1.0000001192092896]` (one float32 step over) | True | refuse | refuse | refuse |
| `[-1.0, 1.0000001192092896]` | False | accept | **refuse** | accept |
| `[-1.0, 1.05]` | True | refuse | refuse | refuse |
| `[-1.0, 1.05]` | False | accept | **refuse** | accept |
| `[-1.0, 3.0]` | True | refuse | refuse | refuse |
| `[-1.0, 3.0]` | False | accept | **refuse** | accept |
| `[-1.05, 1.0]` | True | refuse | refuse | refuse |
| `[-1.05, 1.0]` | False | refuse | refuse | refuse |

What the metric then does with a `[-1.0, 1.05]` frame at `normalize=False`,
through the real `torchmetrics.functional.image.lpips._lpips_update`:

| | outcome |
|---|---|
| real `_valid_img` | check passed (reached the LPIPS network) |
| shim on `main` | `ValueError: Expected both input arguments to be normalized tensors with shape [N, 3, H, W]` |
| shim on this branch | check passed (reached the LPIPS network) |

## Change

One expression, plus the comment above it. The `normalize=False` branch drops the
upper bound so it mirrors the check it stands in for; the comment now records the
asymmetry and why it is load-bearing, in place of the claim that the
reimplementation was faithful.

```diff
-            value_check = img.min() >= 0.0 and img.max() <= 1.0 if normalize else img.min() >= -1.0 and img.max() <= 1.0
+            value_check = img.min() >= 0.0 and img.max() <= 1.0 if normalize else img.min() >= -1.0
```

The `normalize=True` branch, the lower bound, the rank and channel-count rules,
the `NoTrainLpips` alias and the "only fill in what is missing" guards are
unchanged.

## Tests

`tests/policies/vera/test_lpips_shim_accepts_what_the_metric_accepts.py`, 57
cases. torchmetrics is not a dependency of this package, so the suite installs a
`torchmetrics.image.lpip` stand-in shaped like the modern layout (private
`_NoTrainLpips`, no `_valid_img`) and drives the shim over plain
`numpy.ndarray` frames - the real check is duck-typed over
`min`/`max`/`ndim`/`shape`, so an array is a faithful stand-in that runs
everywhere.

The case table is graded against the range each frame actually carries rather
than against a hand-written expectation, and the table itself is checked for the
property that makes it interesting: that the two branches disagree somewhere, and
that some frame overshoots `1.0`. A shim collapsing both branches into one shared
rule fails there rather than passing quietly.

Pre-fix split on this file: **9 failed, 48 passed**, with zero failures in the
case-table, `normalize=True`, shape-rule, `NoTrainLpips` and
"only-fill-in-what-is-missing" classes. The headline failure is
`test_one_float_step_above_one_is_accepted`.

## Mutation coverage

Each row is one mutation of the fixed module, graded against the new file. The 11
mutations that produce failures have 11 distinct firing sets.

| mutation | failed | firing classes |
|---|---|---|
| control (no mutation) | 0 | - |
| the upper bound restored on the `normalize=False` branch (the defect) | 9 | ReproducesTheCheck, UnnormalizedBranch |
| one shared rule for both branches (`normalize` ignored) | 9 | NormalizedBranch, ReproducesTheCheck |
| the `normalize=True` branch loses its upper bound | 3 | NormalizedBranch |
| the `normalize=False` lower bound dropped | 6 | ReproducesTheCheck, UnnormalizedBranch |
| the `normalize=False` lower bound made strict (`> -1.0`) | 12 | ReproducesTheCheck, UnnormalizedBranch |
| the two branches swapped | 17 | all three range classes |
| the rank check dropped | 6 | ShapeRules |
| the channel-count check dropped | 6 | ShapeRules |
| the verdict left as an array scalar instead of a `bool` | 28 | all three range classes |
| an already-installed `_valid_img` overwritten anyway | 2 | OnlyFillsInWhatIsMissing |
| an already-installed `NoTrainLpips` overwritten anyway | 1 | OnlyFillsInWhatIsMissing |
| the absent-torchmetrics handler narrowed away | 1 error | whole file (collection error) |

The last row is reported honestly as an error rather than a failure: the module
calls `_apply()` at import, so a handler that no longer catches the absent
dependency stops the module importing at all - `ModuleNotFoundError: No module
named 'torchmetrics'` out of `sitecustomize`, which is the reason that handler is
broad. The module source was restored byte-identical after every row.

## Coverage

`strands_robots/policies/vera/docker/sitecustomize_vera.py` was the
lowest-covered module in the package. Both numbers are from full
`pytest tests/` runs with `--cov=strands_robots`, one per tree:

| | statements | missing | covered |
|---|---|---|---|
| `main` | 16 | 9 (lines 27-41) | 44% |
| this branch | 16 | 0 | 100% |

Package total missing 1888 to 1878, with the statement count unchanged at 45871.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1546 files).
- `mypy strands_robots tests tests_integ`: **5 == 5** against a pristine `main`
  worktree, normalized message sets byte-identical in both directions, none in
  the changed files. The container helpers are in `tool.mypy.exclude`, and the
  test reaches its subject by name for the same reason its sibling on that path
  does.
- Full `pytest tests/` on both trees: `main` **29 failed, 38403 passed, 309
  skipped, 24 errors** (38765 collected); this branch **29 failed, 38460 passed,
  309 skipped, 24 errors** (38822 collected). `+57` is exactly the new cases, and
  the 53 failing/erroring ids are identical with an empty set difference in both
  directions (44 need `awsiot` from the `mesh-iot` extra; the rest are
  pre-existing on `main`).
- New file: 57 passed.

## Artifact

The two columns are the same measurement run against each tree with real
torchmetrics 1.9.0 installed, since CI does not have that dependency and cannot
produce this comparison.

shim vs torchmetrics agreement

### DIFF
--- a/changelog.d/2732-vera-lpips-shim-unnormalized-range.md
+++ b/changelog.d/2732-vera-lpips-shim-unnormalized-range.md
@@ -1,0 +1,22 @@
+### Fixed
+
+- **policies/vera/docker**: the LPIPS compat shim no longer refuses a frame the
+  check it stands in for accepts. VERA's DFoT metrics import `_valid_img` from
+  `torchmetrics.image.lpip`, where modern torchmetrics no longer keeps it, and
+  the image installs `sitecustomize_vera` as an auto-imported `sitecustomize` to
+  supply the name - so the shim's verdict, not the real one, is the verdict the
+  metric acts on. The two branches of the real check in
+  `torchmetrics.functional.image.lpips` are deliberately asymmetric: both ends of
+  `[0, 1]` are bounded when `normalize=True`, and only the LOWER end of
+  `[-1, 1]` when `normalize=False`, because a frame already in the network's own
+  range is allowed to overshoot `1.0` - which a decoder emitting a tanh-shaped
+  frame routinely does by a float step or two. The shim bounded the upper end on
+  both branches, and `_lpips_update` turns a `False` from that check into a
+  `ValueError`, so the added bound did not make the metric stricter, it made an
+  eval that the unshimmed torchmetrics runs raise inside the image. Measured
+  against torchmetrics 1.9.0, 3 of 14 `(frame, normalize)` cells disagreed with
+  the real check - all three `normalize=False` frames whose maximum exceeds
+  `1.0`, the smallest of them by one float32 step - and 14 of 14 agree now. The
+  `normalize=True` branch, the lower bound, and the rank and channel-count rules
+  are unchanged, and the comment claiming the reimplementation was faithful now
+  records the asymmetry and why it matters.
--- a/strands_robots/policies/vera/docker/sitecustomize_vera.py
+++ b/strands_robots/policies/vera/docker/sitecustomize_vera.py
@@ -29,13 +29,18 @@ def _apply() -> None:
         if priv is not None:
             _lpip.NoTrainLpips = priv
 
-    # _valid_img: removed from the public module in modern torchmetrics - provide
-    # a faithful reimplementation (range/shape check used by the LPIPS metric).
+    # _valid_img: removed from the public module in modern torchmetrics - mirror
+    # the surviving implementation in torchmetrics.functional.image.lpips, the one
+    # the metric itself calls. Its two branches are deliberately asymmetric:
+    # normalize=True bounds both ends of [0, 1], while normalize=False bounds only
+    # the LOWER end of [-1, 1]. Bounding the upper end here too would refuse a
+    # tensor the real check accepts, and _lpips_update turns a False into a
+    # ValueError - so a decoded frame that overshoots 1.0 by a float step would
+    # fail an eval that the unshimmed torchmetrics runs.
     if not hasattr(_lpip, "_valid_img"):
 
         def _valid_img(img, normalize: bool):  # noqa: ANN001
-
-            value_check = img.min() >= 0.0 and img.max() <= 1.0 if normalize else img.min() >= -1.0 and img.max() <= 1.0
+            value_check = img.min() >= 0.0 and img.max() <= 1.0 if normalize else img.min() >= -1.0
             return img.ndim == 4 and img.shape[1] == 3 and bool(value_check)
 
         _lpip._valid_img = _valid_img
--- a/tests/policies/vera/test_lpips_shim_accepts_what_the_metric_accepts.py
+++ b/tests/policies/vera/test_lpips_shim_accepts_what_the_metric_accepts.py
@@ -1,0 +1,283 @@
+"""The LPIPS compat shim accepts exactly what the check it stands in for accepts.
+
+:mod:`strands_robots.policies.vera.docker.sitecustomize_vera` exists because
+VERA's DFoT metrics import ``_valid_img`` and ``NoTrainLpips`` from
+``torchmetrics.image.lpip``, and modern torchmetrics keeps neither name there.
+The image installs the module as an auto-imported ``sitecustomize``, so it runs
+before anything else in the container and re-exposes both names on that module.
+VERA then calls the shim's ``_valid_img`` rather than the real one.
+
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

Narrow, well-evidenced fix: the `sitecustomize_vera` LPIPS shim's `_valid_img` bounded the upper end of the range on both branches, while the real check in `torchmetrics.functional.image.lpips` deliberately bounds only the lower end when `normalize=False` - so a tanh-shaped frame overshooting `1.0` by a float step was refused by the shim (and turned into a `ValueError` by `_lpips_update`) on an eval the unshimmed torchmetrics runs. The change drops the spurious upper bound on the `normalize=False` branch only; I independently verified the fixed expression agrees with the real check on all 16 `(frame, normalize)` cells from the PR table, and the new 57-case suite passes at head. The `normalize=True` branch, lower bound, rank/channel rules, `NoTrainLpips` aliasing and the only-fill-in-what-is-missing guards are untouched.

## What's good

- The case table is self-grading (verdicts derived from each frame's actual range) and asserts its own discriminating power (branches must disagree somewhere, some frame must overshoot `1.0`), so a shim collapsing both branches into one rule cannot pass quietly.
- Mutation table in the description shows distinct firing sets per mutation, including the exact defect being fixed.
- `sys.modules` stand-ins are installed/removed via `monkeypatch.setitem`/`delitem`, per AGENTS.md's orphaned-module rule; changelog fragment present; strings are ASCII; test names describe behaviour.
- The comment that previously claimed the reimplementation was "faithful" now records the asymmetry and why it is load-bearing - the next reader won't restore the bound as a cleanup.

PR #2733, opened 2026-08-25
fix(mesh): survey each sensor provider under its own guard

cagataycali:
## Problem

`Mesh._build_presence` advertises a `topics` list naming the extended telemetry a peer can serve. Each entry is decided by probing a provider attribute on the robot:

```python
try:
    if getattr(r, "_pose", None) is not None or getattr(r, "_slam_pose", None) is not None or ...:
        available_topics.append("pose")
    if getattr(r, "_imu", None) is not None:
        available_topics.append("imu")
    ...                                    # five more probes
except Exception:
    pass
```

Seven independent probes under **one** guard. On hardware these names are properties that read a live sensor bus, not plain fields, so one can fail while every other sensor stays readable - and a sensor-bus fault is not an `AttributeError`, so it propagates straight through `getattr(robot, name, None)` and reaches the survey. The first fault then abandoned the rest of it.

The advertised set was therefore a **prefix of the truth, cut at the first fault**. Measured through the public builder on a robot exposing all ten providers with exactly one faulting:

| faulting provider | advertised before | advertised after |
|---|---|---|
| none | pose, imu, odom, lidar, health, hand, map | unchanged |
| `_pose` | `health` | pose, imu, odom, lidar, health, hand, map |
| `_imu` | pose, health | pose, odom, lidar, health, hand, map |
| `_odom` | pose, imu, health | pose, imu, lidar, health, hand, map |
| `_lidar_summary` | pose, imu, odom, health | pose, imu, odom, lidar, health, hand, map |
| `_battery` | pose, imu, odom, lidar, health | pose, imu, odom, lidar, health, hand, map |
| `_hands` | pose, imu, odom, lidar, health | pose, imu, odom, lidar, health, map |
| `_map_info` | pose, imu, odom, lidar, health, hand | pose, imu, odom, lidar, health, hand |

`_pose` is surveyed **first**, so the worst case is the common one: a robot whose pose provider faults advertised `health` alone while still serving its IMU, lidar, hands and map. `_map_info` is surveyed last and so lost nothing either way - the loss was positional, which is why no fixed expectation could have caught it.

This is not cosmetic. The capability list is the only place a peer announces those topics, so a subscriber never learns the peer serves data `strands_robots.mesh.sensors` would have answered with, and `except Exception: pass` means nothing is reported anywhere.

## Why this granularity is the repo's own rule

`strands_robots.mesh.sensors` reads these **same attributes** and guards each one separately. Measured on the identical robot, a faulting `_battery`:

| | result |
|---|---|
| `_read_health()` (per-attribute guards) | drops `battery_pct`/`charging`, keeps `cpu_load`, `mem_pct`, `disk_free_gb`, `uptime_s` |
| `_build_presence()` before | drops `hand` and `map`, which have nothing to do with the battery |

`docs/mesh.md` already specifies the contract for the sibling payload - *"Every section of a state snapshot is optional"* and *"A probe that fails therefore names itself, so the fault is on the wire rather than only in that peer's log"*. The advertisement did not honour it. No docs change is needed: the documented contract was already correct, and `docs/mesh.md`'s *"Sensor topics only publish when the robot exposes the attribute"* becomes true rather than conditional on which **other** attributes happen to be healthy.

## Change

`_sensor_present(robot, *attrs)` reads one attribute at a time, continues to that topic's sibling providers after a fault, and names the dropped provider on the debug log. The shared `try` is removed - it becomes dead, since the helper cannot raise and `list.append` cannot fail. One caller, so the helper stays module-local per the convention that a guard becomes shared when it gains a second caller.

A topic with several providers now survives one of them faulting: `pose` still advertises when `_pose` raises and `_slam_pose` answers.

## Tests

`tests/mesh/test_presence_capability_probe_isolation.py`, 39 cases. The topic-to-provider map is **derived from `_build_presence` itself**, and deliberately blind to *how* each attribute is read, so the same cells grade the pre-fix and post-fix spellings and a topic that later gains or loses a provider is graded on arrival.

The two pre-existing pins covered only the extremes, where a shared guard and per-provider guards are indistinguishable: `tests/mesh/test_mesh.py` drives a robot whose every attribute raises (`topics == ["health"]`) and `tests/mesh/test_deep_mesh.py` one whose every sensor answers (all seven). Both still pass unchanged. The **mixed** robot is what separates them.

Pre-fix splits:

- raw revert of `core.py` to `main`: collection error (`_sensor_present` absent)
- behavioural revert (helper kept, shared guard restored): **10 failed, 29 passed** - including `test_the_remaining_capabilities_are_still_advertised` and `test_a_faulting_head_pose_still_advertises_the_imu`, with **0** failures in the premise, over-reach or probe-unit classes

### Break table

8 mutations, control clean, every one detected, source restored byte-identical.

| # | mutation | failing cells | distinct tests |
|---|---|---|---|
| - | control (unmutated) | 0 | - |
| b1 | shared guard restored + bare `getattr` (the pre-fix shape) | 10 | 5 |
| b2 | probe kept, but re-hoisted under one shared `try` | 1 | 1 |
| b3 | probe stops at the first fault instead of trying siblings | 5 | 4 |
| b4 | probe re-raises (`except ZeroDivisionError`) | 22 | 10 |
| b5 | `_pose` read by a bare `getattr` again | 9 | 9 |
| b6 | the dropped provider is no longer logged | 1 | 1 |
| b7 | probe reports a faulting provider as present | 9 | 6 |
| b8 | a `None` reading counts as present | 4 | 4 |

b2 is worth stating: `_sensor_present` never raises, so re-wrapping it in a shared `try` is behaviourally a no-op and **only** the structural cell catches it. That cell is what stops the shape regressing, and it is load-bearing rather than decorative. b8 is the over-reach guard - tolerating a fault must not turn "no reading" into a claimed capability.

## Local gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1546 files)
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine worktree at this branch's merge base, normalized message sets byte-identical both directions, 0 in either changed file
- Paired run, identical 298-file selection (all `tests/mesh/` plus every test that walks the tree, reads docstrings, or reads `changelog.d`), the new file excluded from **both** trees: identical **8093 collected** and `21 failed, 7984 passed, 64 skipped, 24 errors` on both (354s vs 369s). 45 failing/error ids each, `comm` empty in **both** directions. All 45 are `tests/mesh/test_iot_*` needing `awsiot` (`awsiotsdk`, declared in the `[mesh-iot]` extra - confirmed with `tomllib` + `find_spec`), absent on this host.
- New file plus the three suites that assert on this payload (`test_mesh.py`, `test_deep_mesh.py`, `test_camera_schema.py`): **160 passed** on a second interpreter.

No visual artifact: this changes a telemetry payload, not a policy, simulation behaviour, render, recording or asset. The measured before/after table is the evidence.

### Round 1 - absorb the base so the overlap check has a tree to grade

`Detect an untested overlap with the base branch` was `FAILURE` on `75d97713`, and correctly: `a45fea93` (#2729) landed on `main` after this branch's checks ran and edits the **same file**, `strands_robots/mesh/core.py`. The green on `75d97713` was therefore a statement about a tree that no longer existed - the #1763/#1766 topology, where the text merges with no conflict marker and only the composition is red.

The two changes turn out to be disjoint at method granularity, which is worth stating rather than assuming:

| | this branch | #2729 |
|---|---|---|
| method changed | `_build_presence` | `_read_state` |
| method added | `_sensor_present` | `_running_policy_robots` |
| topic affected | presence (`topics` advertisement) | state (`robots[].active`) |

So no behaviour is shared, and the new test file's derived topic-to-provider map does not read anything #2729 touched.

**Composed and measured rather than merged on the green.** `main` absorbed into the branch, read as a delta per the rule that an absolute count can both invent a regression and hide one:

| selection | base (`b017ebe0`) | composition (`d82f4570`) |
|---|---|---|
| `test_mesh.py`, `test_deep_mesh.py`, `test_state_topic_reports_which_robot_runs_a_policy.py`, `test_sensor_payloads_are_json_encodable.py` | 143 passed, 8 skipped | - |
| the same four plus `test_presence_capability_probe_isolation.py` | - | 182 passed, 8 skipped |

Zero failures on either side, and the whole effect of the merge is **+39 passes** - exactly the 39 cases this branch adds. `test_state_topic_reports_which_robot_runs_a_policy.py` is #2729's own pin and passes against this branch's `core.py`.

**The push costs no review.** Clean base merge, so it is the #1821 shape rather than a conflict resolution: `git show --cc` is **0 lines** and the pull request's own diff against its merge base is byte-identical before and after at `3 files, +488/-22`. `behind_by` is now `0`, so the tree CI grades is the merge result. No file this branch owns changed, and no reviewed text moved.

### Still outstanding, and not clearable from this branch

`scripts/check_merge_base_overlap.py --all-open` reports one pair: **#2733 + #2735**, both editing `strands_robots/mesh/core.py`. That is a property of the open *set*, so no push by either author makes it green. #2735 is approved and ahead in the queue; the order that needs no further work is #2735 first, then this branch absorbs it and the same composition run above is re-run over the shared file. Recording it here so it is not rediscovered as a surprise at merge time.


yinsong1986 (dismissed):
## Summary

Replaces the single shared `try` around `_build_presence`'s capability survey with a per-provider probe (`_sensor_present`), so one faulting sensor property (e.g. a `_pose` read that raises on a bus fault) no longer aborts the whole survey and erases the IMU/lidar/hand/map topics the robot is still serving. Behaviour is strictly more accurate advertisement: the `topics` payload keys and values are unchanged in shape, `None`/absent still means "not a capability", `health` stays unconditional, and a fault is now logged at debug rather than silently dropped. The probe granularity matches how `strands_robots.mesh.sensors` already reads the same providers, and the new test file grades the mixed case (one fault, rest healthy) that the pre-existing all-fault / all-healthy pins could not distinguish. Note the PR body is still a placeholder ("measured tables to follow") - worth filling before the human review pass.

## What's good

- The topic-to-provider map in the new test is derived from `_build_presence`'s AST rather than restated, so a topic gaining or losing a provider is graded on arrival.
- Premise tests pin why the guard is needed at all (a non-`AttributeError` propagates through 3-arg `getattr`), and controls pin both directions: no over-reach (a faulting/absent provider is not a capability) and no under-report (siblings survive one fault).
- The `except Exception` in the probe is a genuine tolerance path (arbitrary hardware provider faults), narrower in span than the block it replaces, and logs where the old code passed silently.

## Verification suggestions

- `python -m pytest tests/mesh/test_presence_capability_probe_isolation.py tests/mesh/test_mesh.py tests/mesh/test_deep_mesh.py --no-cov` - all pass on the head commit (39 + 104).
- The `Detect an untested overlap with the base branch` failure is the branch being 1 commit behind `main` (#2729, which also touched `mesh/core.py`). The overlap is textually and semantically disjoint from the survey (state-topic active flag), so a clean merge of `main` into the branch should clear it without dismissing the review (the #1821 clean-base-merge case in AGENTS.md).

yinsong1986 (approved):
## Summary

Replaces the single shared `try ... except Exception: pass` around `_build_presence`'s capability survey with a per-provider probe (`_sensor_present`), so one faulting sensor property no longer aborts the survey and erases every topic probed after it. The `topics` payload keeps its shape and vocabulary unchanged -- the only behavioural difference is that previously-dropped entries the robot can still serve are now correctly advertised, `health` stays unconditional, a `None` reading still contributes no topic, and a fault is logged at debug instead of vanishing. The probe granularity matches how `strands_robots.mesh.sensors` already reads the same provider attributes, the helper stays module-local per the one-caller convention, and the changelog fragment is in place.

## What's good

- The new test file derives the topic-to-provider map from `_build_presence`'s AST rather than restating it, so a topic gaining or losing a provider is graded on arrival, and the structural cell (`TestEveryProviderIsProbedUnderItsOwnGuard`) is what stops the shape regressing to a shared guard -- the one mutation that is behaviourally invisible.
- Premise tests pin why the guard is needed at all (a non-`AttributeError` propagates through 3-arg `getattr`), and the over-reach cells pin both directions: a fault or a `None` reading is never claimed as a capability.
- The round changelog composed with #2729 (`_read_state` / `_running_policy_robots`, disjoint at method granularity) and measured the merge as a delta rather than trusting the green, and the remaining overlap with #2735 is named rather than left to be rediscovered at merge time.

## Verification suggestions

- `python -m pytest tests/mesh/test_presence_capability_probe_isolation.py tests/mesh/test_mesh.py tests/mesh/test_deep_mesh.py --no-cov` -- 143 passed on the head commit, reproduced during this review.
- At merge time, honour the ordering the description records for the open-set overlap: #2735 (approved, same file) first, then this branch absorbs it and the composition run is repeated over `strands_robots/mesh/core.py`.

cagataycali:
## Review: the pair this branch's body leaves open is now composed and measured

Nothing is owed on this branch. `d82f4570` already absorbed the base, and it absorbs a
*newer* one (`b017ebe0`) than the merge I had independently produced from `a45fea93`, so I
discarded mine rather than push a competing merge. Worth recording that the two
independent resolutions agree: the diff between them restricted to the shared file
`strands_robots/mesh/core.py` is **empty**. The 13 files that do differ are exactly the
extra base commits (#2730's `mesh/sensors.py`, #2731's training changes), not a
resolution disagreement.

What follows is the measurement the body's closing section says cannot be produced from
this branch. It can - it just cannot be produced by a *push*.

### The union proved by break, not by pass count

Round 1's composition evidence is `+39 passes`. That establishes both sides pass
together; it cannot distinguish that from one side's assertions having gone **vacuous**
under the other's change, which is what the #1763/#1766 topology is made of. The form
that separates them is to revert each fix on the composed tree and confirm its own cells,
and only its own, still fire.

Composed tree: `d82f4570` merged with #2735 at `71deb9fd`. Clean `ort` auto-merge of
`strands_robots/mesh/core.py`, no conflict marker, which is precisely the reassuring
signal the check's docstring warns about.

| row | mutation | failing | #2733 | #2735 | #2729 | #2730 |
|---|---|---|---|---|---|---|
| c0 | control, unmutated | 0 | 0 | 0 | 0 | 0 |
| c1 | this branch's survey back under one shared `try` | 10 | 10 | 0 | 0 | 0 |
| c2 | #2735's fleet-stop `ok` back to the unconditional `True` | 6 | 0 | 6 | 0 | 0 |
| c3 | #2729's `active` back to the literal `True` | 9 | 0 | 0 | 9 | 0 |
| c4 | all three reverted | 25 | 10 | 6 | 9 | 0 |

`c1 | c2 | c3 == c4` exactly (10 + 6 + 9 = 25), and the three id sets are **pairwise
disjoint**. So no fix's detection power is reduced by the others, and none of the three
regression tests is passing for a reason another branch supplied. Control clean, source
restored byte-identical after every row.

The independence is structural as well as behavioural - the three helpers have no shared
consumer, read off the composed file rather than asserted:

| helper | defined at | called from |
|---|---|---|
| `_reports_failure_to_stop` (#2735) | 359 | `_dispatch`, `_peers_that_did_not_stop` |
| `_sensor_present` (this branch) | 428 | `_build_presence` |
| `_running_policy_robots` (#2729) | 1487 | `_read_state` |

Both new module-level helpers land immediately around `_peers_that_did_not_stop`, which
is why the file needed an auto-merge at all. Adjacent text, disjoint behaviour.

### Nothing pre-existing regresses under the composition

Identical 311-file selection (all of `tests/mesh/` plus every test that walks the tree,
reads docstrings, or reads `docs/`), run on the composition and on a pristine worktree at
`a45fea93`:

| tree | collected | result |
|---|---|---|
| base `a45fea93` | 8173 | 21 failed, 8066 passed, 62 skipped, 24 errors (365s) |
| composition | 8173 | 21 failed, 8066 passed, 62 skipped, 24 errors (361s) |

45 failing/error ids each, `comm` empty in **both** directions. All 45 are
`tests/mesh/test_iot_*` needing `awsiot` (`awsiotsdk`, declared in the `[mesh-iot]`
extra), absent on this host - confirmed with `tomllib` plus `find_spec` rather than
assumed. `tests/mesh/test_zenoh_transport_security.py` is excluded from that selection
and run alone: 9 passed, 2 skipped on both trees.

On the composed tree specifically: `ruff check` and `ruff format --check` clean over
`strands_robots tests tests_integ` (1550 files); `mypy strands_robots tests tests_integ`
**24 == 24** against that same pristine worktree, normalized message sets byte-identical
in both directions, the only two naming `mesh/core.py` being pre-existing `[arg-type]`
reports whose line numbers merely shift. The four branches' own suites are **105 passed**
together, on the declared 3.5.0 mujoco floor and on 3.12.0 alike.

### The merge order does not matter

The body proposes #2735 first, then this branch absorbing it. Measured: merging in the
other order produces a **byte-identical tree** - the diff between the two compositions is
empty. Either order is safe, so this pair does not constrain the queue, and whichever
lands second needs no re-run beyond the one above.

### Where that leaves the reported row

`--all-open` still reports the pair and always will, since it is a property of the open
set. Its own remedy line is:

> To clear a row above: decide the merge order, run the tests covering the shared paths
> against the composition once, and merge.

The table above is that run. Both heads are additionally clear against the base as it now
stands at `1dbb5695` (#2734), which touches no file either branch edits: this branch
reports `No overlap ... diverged at b017ebe0 (15 path(s) changed on main in that span)`,
and #2735 the same at `a45fea93` (28 paths).

One caveat on scope, since the check is careful to name it and this composition inherits
the limitation: this branch's new test file derives its topic-to-provider map from
`inspect.getsource(Mesh._build_presence)`, so it is coupled to a method rather than to a
path. That coupling is inside the composition measured above - #2729 and #2735 change
`_read_state` and `_dispatch`, neither of which that derivation reads - but a future
sibling editing `_build_presence` itself would share the derivation without sharing a new
path, and the pairwise relation would be silent about it.

PR #2734, opened 2026-08-25
feat(hardware): choose which driver builds a real robot

cagataycali:
## Problem

`Robot(name, mode="real")` built exactly one thing: `strands_robots.hardware_robot.Robot`,
which constructs a lerobot `RobotConfig` and wraps a lerobot driver. That is the right driver
for every robot in the shipped registry, and the wrong shape for a robot lerobot does not
model - a humanoid with its own state machine, a rover reporting GPS, a base publishing a
point cloud. There was no seam to reach anything else through.

Nothing recorded the contract either. The mesh, the teleop rail and the agent tool surface all
consume from whatever `mode="real"` returns, so the members they rely on were a property of a
3000-line class and of nothing else - a driver author had to read it to find out which ones
are load-bearing.

## Change

`driver=` selects the implementation, keyword-only so no positional call shifts meaning:

| Value | Builds |
|-------|--------|
| `"auto"` (default) | The robot's registry `hardware.driver` if it declares one, else the lerobot driver. |
| `"lerobot"` | The lerobot driver, explicitly. |
| `"strands"` | The native driver registered for this robot. |

`strands_robots.drivers.HardwareDriver` is the contract, `register_native_driver` is where a
driver joins, and `hardware.driver` lets a robot declare its own. Four decisions are worth
calling out.

**The contract is the measured one, not the plausible one.** Two members that look like they
belong are absent because the reference implementation does not have them: `get_observation`
(`hasattr(HardwareRobot, "get_observation")` is `False` - the mesh reads observations from the
driver's *inner* device at `mesh/sensors.py:384`, and `bus_access.read_observation` takes that
inner device), and `connect_eagerly` (0 occurrences anywhere in the tree). The sensor
attributes a mesh publishes are absent for a different reason: every one is read as
`getattr(robot, "_pose", None)`, so they are optional by construction - a driver with no lidar
is complete, a Protocol cannot say "optional", and requiring them would refuse a working arm.
They are also private, and a public contract that names a private attribute institutionalises
it. What is left is 12 members, four of which are `AgentTool`'s abstract surface - derived from
the SDK in the test rather than listed, so dropping one fails.

**A half-built driver is refused where it is registered.** `register_native_driver` checks the
surface against the class and names what is missing, so a driver without `stream` fails on the
line that is wrong rather than on the first agent call, one process later. The member list is
derived from the Protocol (`dir(HardwareDriver)`), so a second hand-written copy cannot drift
from it. Checking a *class* is why the check is not `issubclass`: a Protocol declaring a
`@property` has non-method members and `issubclass` refuses those with `TypeError`.

**Asking for a driver that is not there is refused, never substituted.** A caller handed the
lerobot driver after asking for a native one would debug the wrong robot. The refusal names the
robot, lists the robots that do have a native driver, and gives both remedies. The value is
checked in *every* mode - a sim robot has no driver to choose, but `driver="lerbot"` is a typo
either way, and a value checked only on the branch that reads it is a value the other branch
accepts silently. A driver name added to the vocabulary with no route in the factory is refused
too, rather than falling into a neighbour's branch.

**Statically honest.** `driver="strands"` returns something that is *not* a `HardwareRobot`, so
a `Literal["strands"]` overload returns `HardwareDriver` - keyed on the literal exactly as
`mode` already is. Without it the factory would promise `HardwareRobot` and hand back a driver
with no `.attach_teleop`, and nothing would say so.

The `mode="real"` post-build tail (mesh attach, Device Connect `.run()` hook) is now reached
after the branch instead of inside it, so both drivers take it: a driver that is built but
never published is a robot no peer can see. Extracting `_attach_mesh` is what mypy asked for -
it writes `mesh`/`peer_id` onto the instance, which a driver does not implement but receives,
so the helper takes `Any` for the same reason its existing `_attach_device_connect` sibling
does.

## The default path is unchanged

The load-bearing test. The same robot is built three ways - not mentioning `driver`,
`driver="auto"`, `driver="lerobot"` - and the class, `tool_name` and the produced lerobot
config (`id`, `port`, config type) are compared across all three. The patch target is lerobot's
own `make_robot_from_config`, so the entire strands-side config build stays on the call chain.
The only deletions in `robot.py` are the re-indented lerobot construction block (contents
preserved verbatim) and two annotation lines.

That class is a control - no mutation below fires it - so its non-vacuity is proved separately:
b13 makes the default `"strands"` and all three of its tests fail.

## Tests

45 new cases in `tests/test_driver_seam.py`, no hardware touched. 14 mutations, 14 distinct
firing sets, control clean, every file restored byte-identical:

| # | Mutation | Fires |
|---|----------|-------|
| b0 | control (no change) | 0 |
| b1 | the driver value is checked only in the real branch | 1 |
| b2 | `driver="strands"` with no driver falls back to lerobot | 3 |
| b3 | registration skips the surface check | 1 |
| b4 | `DRIVER_SURFACE` hardcoded instead of derived | 2 |
| b5 | the registry declaration outranks the caller's explicit choice | 7 |
| b6 | loader validation of `hardware.driver` dropped | 1 |
| b7 | the lerobot type is forwarded to a native driver too | 1 |
| b8 | an unrouted driver name silently builds a native driver | 1 |
| b9 | registration does not resolve an alias to the canonical name | 1 |
| b10 | the shared post-build tail is skipped for a native driver | 1 |
| b11 | `get_driver` applies the default itself instead of reporting the declaration | 2 |
| b12 | a declared `"auto"` is taken as a driver name | 2 |
| b13 | the default driver becomes `"strands"` (non-vacuity for the control class) | 4 |

A raw revert is a collection error, since the seam's symbols are what the file imports; the
table above is the evidence that each test is load-bearing.

`tests/registry/test_public_member_docstrings.py` declares the registry's public surface, so
`get_driver` is added to it there - the guard doing its job.

## Gate

Run against mujoco 3.5.0 (the declared floor) with lerobot 0.6.2 from source.

- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ`: clean, 1550 files.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `main` worktree,
  normalized message sets byte-identical in both directions, 0 in the changed files.
- Full `tests/` on both trees: **53 == 53** failing/error ids, `comm` empty in both directions.
  All 53 are pre-existing and dependency-caused (44 `ModuleNotFoundError: awsiot`, the
  `[mesh-iot]` extra; the rest lerobot-from-source `encode()` drift).
- Collected `+56`, fully accounted: `+45` the new file, `+8` in
  `test_args_docstring_completeness.py` (4 Protocol methods x 2 tests), `+3` in
  `test_finalizer_reports_only_real_cleanup_failures.py` (one per new module). All pass.

No visual artifact: this changes no policy, simulation behaviour, rendering, recording or
asset. The measured tables above are the evidence.

## Docs

`docs/getting-started/robot-factory.md` gains the parameter row and a "Choosing a driver"
section; `docs/hardware/robot-control.md` points at it. The refusal quoted in the docs is
compared byte-for-byte against the real message.

### DIFF
--- a/changelog.d/2734-hardware-driver-seam.md
+++ b/changelog.d/2734-hardware-driver-seam.md
@@ -1,0 +1,32 @@
+### Added: choose which driver builds a real robot
+
+`Robot(name, mode="real")` built exactly one thing - `strands_robots.hardware_robot.Robot`,
+which constructs a lerobot `RobotConfig` and wraps a lerobot driver. That is the right driver
+for every robot in the shipped registry and the wrong shape for a robot lerobot does not
+model: a humanoid with its own state machine, a rover reporting GPS, a base publishing a
+point cloud. There was no seam to reach anything else through, and nothing recorded which
+members of that 3000-line class the mesh, the teleop rail and the agent tool surface actually
+rely on - so a driver author had to read it to find out.
+
+`driver=` now selects the implementation: `"auto"` (the default) honours a robot's registry
+`hardware.driver` and otherwise builds the lerobot driver, `"lerobot"` pins that path, and
+`"strands"` builds the native driver registered for the robot via
+`strands_robots.drivers.register_native_driver`. Asking for a native driver that is not there
+is refused by name, listing the robots that do have one and both remedies, rather than being
+quietly served the lerobot driver - a caller who got the substitute would debug the wrong
+robot. The value is checked in every mode, so a typo is not accepted by whichever branch
+happens not to read it, and a driver name added to the vocabulary without a route in the
+factory is refused instead of taking a neighbour's branch.
+
+`strands_robots.drivers.HardwareDriver` writes the contract down, and it is the measured one
+rather than an aspirational one: `get_observation` is not a member (the mesh reads it from the
+driver's inner device), and neither are the sensor attributes a mesh publishes (`_pose`,
+`_imu` and their siblings are all read with a `getattr` default, so a driver with no lidar is
+complete). `register_native_driver` refuses a class that does not satisfy it and names the
+members it lacks, so a driver missing `stream` fails where it is registered rather than on
+the first agent call. Both drivers take the same post-build path onto the mesh, because a
+robot that is built but never published is one no peer can see.
+
+The default path is unchanged, and that is what the tests pin hardest: not mentioning
+`driver`, `driver="auto"` and `driver="lerobot"` build the same class with the same lerobot
+config.
--- a/docs/getting-started/robot-factory.md
+++ b/docs/getting-started/robot-factory.md
@@ -29,6 +29,7 @@ robot = Robot("so100", mode="auto")  # probes USB, falls back to sim
 | `peer_id` | str | `None` | Stable mesh peer id. Auto-generated if omitted. |
 | `orientation` | list | `None` | Robot base orientation `[w, x, y, z]` in sim world. Ignored when `mode="real"` (reported at debug level). |
 | `keyframe` | str \| int | `None` | Spawn in a model `<keyframe>` pose (name or index) instead of the zero configuration. Ignored when `mode="real"` (reported at debug level). |
+| `driver` | str | `"auto"` | Which implementation drives a real robot: `"auto"` / `"lerobot"` / `"strands"`. `"auto"` honours the robot's registry `hardware.driver` and otherwise builds the lerobot driver. Checked in every mode; only `mode="real"` acts on it (sim reports it as ignored at debug level). See [Choosing a driver](#choosing-a-driver). |
 | `**kwargs` | | | Forwarded to the backend or driver constructor as given. A name it does not recognize is ignored, not refused, so check the spelling against the forwardable list below. |
 
 ## Name resolution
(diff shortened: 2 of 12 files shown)


yinsong1986 (approved):
## Summary

Adds a driver seam to `Robot(name, mode="real")`: a new `driver=` parameter (`"auto"` / `"lerobot"` / `"strands"`), a measured `HardwareDriver` Protocol in `strands_robots/drivers/base.py`, a native-driver registration point (`register_native_driver`) that refuses an incomplete class naming its missing members, and an optional registry `hardware.driver` declaration validated at load. The default path is preserved and pinned: omitting `driver`, `driver="auto"` and `driver="lerobot"` build the same `hardware_robot.Robot` with the same lerobot config, and both driver kinds share the post-build mesh / Device Connect attachment. The new vocabulary is a one-way door and the PR treats it as one: an unknown `driver=` is refused by name in every mode, `driver="strands"` with no registered driver is refused with both remedies rather than silently substituted, a registry-declared driver typo fails the registry load naming the robot, and a `DRIVER_CHOICES` entry with no factory route is refused rather than taking a neighbour's branch.

## What's good

- Ran the new suite and the adjacent guard suites at the head SHA locally: `tests/test_driver_seam.py` (45 passed) and `tests/test_robot_factory.py` + `tests/registry/` + `tests/test_all_exports_are_statically_defined.py` + `tests/test_raises_docstring_completeness.py` (968 passed, 1 skipped).
- `DRIVER_SURFACE` is derived from the Protocol itself (verified it yields exactly the 12 declared members, no metaclass leakage), and a test pins the derivation against the AST.
- The native-driver table is per-test isolated via an autouse fixture; no host paths; test names describe behaviour; changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention.
- The lazy `drivers.base` import in `loader._validate_robots` correctly avoids the registry/drivers import cycle, and the user-overlay-before-validation ordering is pinned so a `user_robots.json` driver typo is graded too.
- `Raises:` blocks on `Robot()`, `resolve_driver` and `register_native_driver` name the new refusals (AGENTS.md > API Consistency).

## Verification suggestions

- `python3 -m pytest tests/test_driver_seam.py tests/test_robot_factory.py -q` reproduces the checks above; nothing here needs hardware.

cagataycali:
### Public context for `cagataycali/robots-harness#353`

The harness issue referenced in the PR body lives on a private fleet-integration tracker. Public-safe summary for external reviewers:

**Goal.** Add a driver-selection seam so `Robot(..., mode="real")` can resolve something other than a lerobot driver, without touching the lerobot path. lerobot's driver is arm/serial/teleop-dataset shaped — it does not model a Livox point cloud, a humanoid state machine, or a rover reporting GPS. The seam lets those drivers be authored as sibling classes implementing one Protocol; `HardwareRobot` (~3000 lines of lerobot coupling) stays untouched.

**Contract choice.** The Protocol is the measured surface `Mesh`, the teleop rail and the agent-tool wrapper actually consume, not the plausible one — which is why `get_observation` and `connect_eagerly` are absent (the mesh reads observations from the driver's inner device, and nothing calls `connect_eagerly`), and the private sensor attrs (`_pose`, `_imu`, `_lidar_summary`, …) are absent (they are optional-by-`getattr` in the consumer, and a public Protocol declaring a private member institutionalises it).

**Registration.** `register_native_driver` checks the surface against the class at registration time and names what is missing — a driver without `stream` fails on the line that is wrong rather than on the first agent call one process later. The member list is derived from `dir(HardwareDriver)`, so a second hand-written copy cannot drift.

Follow-ups on the same tracker: G1 native driver (already open here as #2739), Earth Rover HTTP driver, Reachy Mini daemon driver.

PR #2735, opened 2026-08-25
fix(mesh): report a refused rollout when a fleet stop aggregates

cagataycali:
## Problem

`Mesh.emergency_stop` broadcasts `{"action": "stop"}` and carries no `robot_name`, so on a
simulation peer it lands in the fleet-wide branch of `_dispatch`: read every active rollout from
`_active_policy_robots()`, call `stop_policy` once per robot, and answer.

That branch is the only stop path that *aggregates*, and it aggregated by assuming the answer:

```python
results = {name: dict(r.stop_policy(name)) for name in active}
return {"ok": True, "stopped": list(results), "results": results}
```

`_peers_that_did_not_stop` reads only the top-level `ok`/`status`, never `results`. So a
`stop_policy` that refused was scored as a halt: the refusal was in the payload, unread, and the
robot was listed under `stopped`. `emergency_stop` then reported the peer as having stopped, kept it
out of `peers_not_stopped`, and did not fire the CRITICAL "robots may still be executing" warning.

That is the affirmative lie the two sibling branches are explicitly commented against, and the exact
shape `_peers_that_did_not_stop`'s own docstring lists as something it must catch: "a sim peer whose
`stop_policy` refused answers `{"status": "error", ...}`".

The two stop paths therefore disagreed about one refusal. Measured through `_dispatch` with a peer
carrying two rollouts where `bob`'s stop refuses:

| command | dispatch answer | counted as did-not-stop |
|---|---|---|
| `{"action": "stop", "robot_name": "bob"}` | `{"status": "error", ...}` | **yes** |
| `{"action": "stop"}` (what `emergency_stop` sends) | `{"ok": True, "stopped": ["alice", "bob"], ...}` | **no** |

Same refusal, opposite accounting, and the wrong one is the path the fanout uses.

## Change

`ok` is derived from the per-robot answers instead of assumed, through one shared predicate:

- New `_reports_failure_to_stop(result)` owns the rule `ok is False or status == "error"`. Both
  spellings are needed because the stop verbs disagree about their envelope: `stop_task` and the
  dispatch's own branches answer `{"ok": ...}`, `Simulation.stop_policy` answers the agent-tool
  `{"status": ...}`.
- `_peers_that_did_not_stop` now reads that predicate rather than its own copy of the comparison. A
  second copy of the rule is how the branch came to contradict the aggregation in the first place.
- The fleet-wide branch grades each answer with it, reports `ok: False` with `not_stopped` naming the
  refusing robots, keeps `stopped` for the ones that really stopped, logs at ERROR, and leaves
  `results` in place so the detail is still available.
- The aggregation's docstring enumerated three recognised shapes; it now enumerates the fourth and
  says which one the fanout produces.

A fleet stop in which every rollout stopped is unchanged: `{"ok": True, "stopped": [...]}`, with no
`not_stopped` and no `error` key. A peer with no rollouts and a robot-less gateway peer are both
untouched.

## Tests

`tests/mesh/test_fleet_stop_reports_a_refused_rollout.py`, 19 cases in 5 classes: the refusal is
reported and flagged, both paths agree about it, the rule has a single owner (derived, so a third
copy fails rather than drifts), over-reach controls for the fully-successful and no-rollout cases,
and a premise class grounding the stand-in against the real verb -- `create_simulation("mujoco")`
then `stop_policy("no-such-robot")` really does answer `status="error"`.

Pre-fix, reverting only the branch and keeping the shared helper: **6 failed / 13 passed**, with
**0 failures in the over-reach control class** and 0 in the premise class. The headline failure is
`{'ok': True, ..., 'stopped': ['alice', 'bob']}` while `results['bob']` reads `status: 'error'`.
A raw revert is instead a collection error, since the helper is a new symbol.

Break table, 10 mutations against the new file: **8 distinct firing sets, no undetected mutation,
control clean, source restored byte-identical.**

| mutation | fires |
|---|---|
| control | 0 |
| branch reverted, `ok` hardcoded True (the defect) | 6 |
| `refused` computed but `ok` still True | 4 |
| `stopped` lists every attempted robot | 2 |
| branch re-derives the rule inline (second copy) | 2 |
| aggregation re-derives the rule inline (second copy) | 2 |
| predicate drops the `status == "error"` spelling | 9 |
| predicate drops the `ok is False` spelling | 3 |
| predicate over-broad: anything not `ok=True` is a failure | 7 |
| `not_stopped` key omitted | 2 |
| `error` key omitted | 1 |

The two "second copy of the rule" mutations share a firing set, which is the point: they are the
same mutation class, and the structural tests are what catch either one.

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1547 files).
- `mypy strands_robots tests tests_integ`: **0 branch-only messages** against a pristine `main`
  worktree (normalised message-set diff empty). The two pre-existing `mesh/core.py` `arg-type`
  errors are present on `main` unchanged.
- Paired run, identical 323-file selection (all of `tests/mesh/` plus every test naming
  `emergency_stop`, `_peers_that_did_not_stop` or `mesh.core`, and every guard walking the tree or
  reading docstrings), the new file excluded from both trees: **10450 collected and
  `21 failed, 10341 passed, 64 skipped, 24 errors` on both**, 45 failing/error ids each, `comm`
  empty in both directions. All 45 need `awsiot` (the `[mesh-iot]` extra).
- New file 19 passed; together with the existing `test_gateway_peer_stop_accounting.py`, 25 passed.

No visual artifact: this changes a dispatch envelope and the safety accounting that reads it, not
policy, simulation, rendering, recording or an asset. The measured before/after table above is the
evidence.

### DIFF
--- a/changelog.d/2735-fleet-stop-refused-rollout.md
+++ b/changelog.d/2735-fleet-stop-refused-rollout.md
@@ -1,0 +1,32 @@
+### Fixed: a fleet-wide sim stop reports a rollout whose `stop_policy` refused, instead of answering `ok=True` over it
+
+`Mesh.emergency_stop` broadcasts `{"action": "stop"}` and carries no `robot_name`, so on a
+simulation peer it lands in the fleet-wide branch of `_dispatch`: read every active rollout
+from `_active_policy_robots()` and call `stop_policy` once per robot. That branch is the
+only stop path that aggregates, and it aggregated by assuming the answer -- the per-robot
+results were collected into `results` and the envelope hardcoded `ok=True`.
+
+`_peers_that_did_not_stop` reads only the top-level `ok`/`status`, never `results`, so a
+`stop_policy` that refused was scored as a halt: the refusal sat in the payload unread and
+the robot was listed under `stopped`. `emergency_stop` then kept the peer out of
+`peers_not_stopped` and did not fire the CRITICAL "robots may still be executing" warning.
+That is the affirmative lie the two sibling stop branches are commented against, and the
+exact shape `_peers_that_did_not_stop`'s own docstring lists as one it must catch.
+
+The two paths therefore disagreed about one refusal. Asked to stop `bob` by name the peer
+answered `{"status": "error", ...}` and was flagged; asked to stop everything with `bob`
+among the rollouts it answered `{"ok": True, "stopped": ["alice", "bob"], ...}` and was not.
+The path the fanout uses is the one that swallowed it.
+
+`ok` is now derived from the per-robot answers rather than assumed, through a single shared
+predicate. `_reports_failure_to_stop` owns the rule `ok is False or status == "error"` --
+both spellings, because the stop verbs disagree about their envelope: `stop_task` and the
+dispatch's own branches answer `{"ok": ...}` while `Simulation.stop_policy` answers the
+agent-tool `{"status": ...}`. `_peers_that_did_not_stop` reads that predicate instead of its
+own copy of the comparison, since a second copy of the rule is how the branch came to
+contradict the aggregation. The fleet-wide branch grades each answer with it, reports
+`ok=False` with `not_stopped` naming the refusing robots, keeps `stopped` for the ones that
+really stopped, logs at ERROR, and leaves `results` in place so the detail is available.
+
+A fleet stop in which every rollout stopped is unchanged, with no `not_stopped` and no
+`error` key; a peer with no rollouts and a robot-less gateway peer are both untouched.
--- a/strands_robots/mesh/core.py
+++ b/strands_robots/mesh/core.py
@@ -19,7 +19,7 @@ import socket
 import threading
 import time
 import uuid
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
 from typing import Any
 
 from strands_robots._mesh_switch import mesh_env_request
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Fixes the one stop path that aggregates: the fleet-wide sim branch of `Mesh._dispatch` (reached by `emergency_stop`'s `{"action": "stop"}` fanout, which carries no `robot_name`) hardcoded `ok=True` over the per-robot `stop_policy` answers, so a refused stop sat unread in `results` while `_peers_that_did_not_stop` scored the peer as halted and the CRITICAL warning never fired. The fix derives `ok` from the per-robot answers through a single new predicate, `_reports_failure_to_stop`, which owns the `ok is False or status == "error"` rule for both envelope spellings; `_peers_that_did_not_stop` now delegates to it instead of keeping its own copy, and the fleet branch reports `ok=False` with `not_stopped`/`error` naming the refusing robots while keeping `stopped` and `results` intact. The refusal shape emitted is exactly the `ok=False` + `error` envelope the existing accounting already recognises, so older readers of the envelope get the correct (stricter) verdict, and the fully-successful, no-rollout, and robot-less-gateway answers are byte-for-byte unchanged.

## What's good

- Verified at the head SHA: the 19 new tests pass, the 24 related stop-accounting tests (`test_gateway_peer_stop_accounting.py`, `_peers_that_did_not_stop`, emergency-stop paths) pass, and `ruff check` / `ruff format --check` are clean on both changed files.
- The single-owner structural tests (`TestTheFailureRuleHasASingleOwner`) make a re-derived copy of the rule fail rather than drift, which is the mechanism the defect arrived through.
- Premise class grounds the fixture's refusal shape against the real `Simulation.stop_policy`, behind `importorskip`, so the stand-in cannot silently diverge from production.
- Changelog fragment present (`changelog.d/2735-fleet-stop-refused-rollout.md`) per AGENTS.md step 3; no direct `CHANGELOG.md` edit.
- Scope discipline: two commits, three files, all named in the description; the log message and tool-visible strings are plain ASCII; test names describe behaviour, not provenance.

## Verification suggestions

Beyond CI: `python -m pytest tests/mesh/ -k "emergency_stop or stop_accounting or peers_that_did_not_stop or fleet_stop" --no-cov -q` exercises both stop paths and the aggregation together (24+19 tests, ~20s, no mujoco required for the accounting half).

PR #2736, opened 2026-08-25
fix(kimodo): refuse a config file that cannot supply fields, by name

cagataycali:
## Problem

`KimodoConfig.from_json` parsed a JSON file and handed the result straight to `from_dict`, whose first act is `data.items()`:

```python
p = Path(path)
with p.open("r", encoding="utf-8") as f:
    return cls.from_dict(json.load(f))
```

So a file holding any JSON value other than an object surfaced as a message naming a method of the parsed value rather than the file that could not supply fields. Measured against `main`, one file per row:

| file contents | before | after |
|---|---|---|
| `[1, 2, 3]` | `AttributeError: 'list' object has no attribute 'items'` | `ValueError: KimodoConfig file /tmp/cfg.json must contain a mapping, got list` |
| `"diffusion_steps"` | `AttributeError: 'str' object has no attribute 'items'` | `... must contain a mapping, got str` |
| `42` | `AttributeError: 'int' object has no attribute 'items'` | `... must contain a mapping, got int` |
| `true` | `AttributeError: 'bool' object has no attribute 'items'` | `... must contain a mapping, got bool` |
| `null` | `AttributeError: 'NoneType' object has no attribute 'items'` | `... must contain a mapping, got NoneType` |
| `{not json` | bare `json.JSONDecodeError` | `ValueError: KimodoConfig file /tmp/cfg.json is not valid JSON: ...` |
| path absent | `FileNotFoundError: [Errno 2] No such file or directory: '/tmp/absent.json'` | `FileNotFoundError: KimodoConfig file not found: /tmp/absent.json` |
| path is a directory | `IsADirectoryError` from the read | `FileNotFoundError: KimodoConfig file not found: /tmp/cfg.json` |
| `~/kimodo.json`, file present | `FileNotFoundError: ... '~/kimodo.json'` | loads |

The last row is the one a user hits without doing anything unusual: `~` was never expanded, so a config that exists was reported missing, with the literal `~` in the message.

## Why the disposition is not a judgement call

Three policy configs ship a file loader. The other two already do all four of these things, and one of them states why in its own `Raises:` block:

| loader | expands `~` | `is_file` named | decode wrapped | non-object refused |
|---|---|---|---|---|
| `MotionBricksConfig.from_file` | yes | yes | yes | yes |
| `WBCConfig.from_file` | yes | yes | yes | yes |
| `KimodoConfig.from_json` (before) | no | no | no | no |

`WBCConfig.from_file` and `MotionBricksConfig.from_file` both end with the identical

```python
if not isinstance(data, dict):
    raise ValueError(f"{cls} file {p} must contain a mapping, got {type(data).__name__}")
```

so this is two of three loaders' shipped reporting applied to the third, not a new rule. The tests grade it over all three by discovering them (any public class in a `strands_robots/policies/*/config.py` module exposing a `from_*(path)` classmethod), so a fourth policy config's loader is held to it on arrival rather than inheriting an exemption by being absent from a tuple.

## Deliberate scope boundary

The two siblings also refuse a file whose extension is not `.json`. `from_json` still does not, and that is deliberate: a JSON object stored under another name loads today, so refusing one would stop a payload that currently works. **No input that loads today stops loading** - every change above reports an input that already failed, or (in the `~` case) accepts one that should not have. The boundary is pinned as a control (`test_a_json_object_under_a_non_json_extension_still_loads`), and a mutation that adds the extension check fires exactly that one case.

## The unrecognised-key half

While measuring the above, two statements about unrecognised keys turned out to be false.

`from_dict` documented the drop as happening "with a warning". Measured, none of the three policy configs warns, and the other two document the drop as *silent* forward compatibility. So the docstring was the outlier, and it now states what the code does. The behaviour is unchanged - a mutation that adds the warning, and one that raises instead, are both included as over-reach rows and both fire.

`docs/policies/kimodo.md` presents three interchangeable ways to set a field and then claimed a misspelled knob "raises `TypeError` at construction instead of being silently ignored". Measured:

| form | misspelled knob |
|---|---|
| `KimodoPolicy(diffusion_stpes=25)` | `TypeError`, as documented |
| `KimodoConfig(diffusion_stpes=25)` | `TypeError`, as documented |
| `KimodoPolicy(config={"diffusion_stpes": 25})` | builds with `diffusion_steps=100`, no warning |

The third form is read by `from_dict`, whose documented policy is exactly to drop it, so the page was asserting the opposite of the code one paragraph below the example that reaches it. The page now says which form refuses and which drops, and how to have a typo refused. The three cases that grade it pass on `main` too - they are the measurement that decided the wording, not regression cases; the two that fail pre-change are the two prose claims.

## Tests

`tests/policies/kimodo/test_config_file_loader_refusals.py`, 60 cases.

Pre-change (source and docs page reverted to `main`, tests kept): **15 failed, 45 passed**. Every failing case is a `KimodoConfig` cell - `0` of the `MotionBricks` and `WBC` parametrizations fail, which is the evidence the rule is the siblings' shipped behaviour rather than something new. By class:

- `TestANonObjectPayloadIsRefusedByName` 12
- `TestAnUnrecognisedKeyIsDroppedWithoutWarning` 1
- `TestEachConstructionFormHandlesAMisspelledKnobAsDocumented` 2
- `TestTheSurveyReachesEveryPolicyConfigLoader` 0 (premise and non-vacuity)
- `TestNoPayloadThatLoadsTodayStopsLoading` 0 (over-reach controls)

Mutation table, run against the finished change: **16 mutations, 15 detected, 15 distinct firing sets, control clean.**

| mutation | fires |
|---|---|
| mapping check dropped | 8 |
| mapping check weakened to `data is not None` | 7 |
| refusal stops naming the JSON type | 1 |
| refusal stops naming the file | 1 |
| refusal stops naming the class | 1 |
| decode failure no longer wrapped | 1 |
| `expanduser()` dropped | 1 |
| `is_file` check dropped | 2 |
| `is_file` weakened to `exists` | 1 |
| `from_dict` docstring warning claim restored | 1 |
| over-reach: `from_dict` warns on a dropped key | 2 |
| over-reach: `from_dict` raises on an unrecognised key | 3 |
| over-reach: extension refused like the two siblings | 1 |
| docs: the refuted claim restored | 1 |
| docs: stops naming the reader that drops the key | 1 |
| `json` import moved back inside the method | 0 |

The last row is a semantic no-op and is reported as one: moving `json` to the module top (where both sibling configs keep it) changes no behaviour, so nothing should fire. The `is_file` pair is why the check is `is_file()` and not `exists()`: `exists()` lets a directory through to the read, and only the directory case distinguishes them.

## Coverage

`from_json` and the `model_id` domain were both entirely unexercised, which is how the loader shipped with none of the four guards. `strands_robots/policies/kimodo/config.py`: **56 statements, 5 missing, 91% to 63 statements, 0 missing, 100%**. Package total `1867 to 1861` missing.

## Gate

Paired full suite, base `1dbb5695` versus the branch, same environment (mujoco 3.5.0, lerobot 0.6.2):

- base: 39222 collected, `29 failed, 38860 passed, 309 skipped, 24 errors`
- branch: 39286 collected, `29 failed, 38924 passed, 309 skipped, 24 errors`
- 53 failing ids on each side, identical sets, `comm` empty in both directions (44 of them need `awsiot` from the `mesh-iot` extra, absent here)
- `+64 collected = +64 passed`, localised: 60 in the new file, and `tests/test_args_docstring_completeness.py` 440 to 444 because `from_dict` and `from_json` gained `Args:`/`Returns:` blocks

`ruff check`, `ruff format --check` and `mypy` over `strands_robots tests tests_integ` all clean; mypy is `24 == 24` against a pristine base worktree with byte-identical message sets and 0 in the changed files. The branch has since merged `d33a779d` (mesh only, no overlap with these files); lint, mypy, the mutation table and the pre-change split were all re-measured on the merged tree and are unchanged. The new file passes on the declared mujoco floor (3.5.0) and on 3.12.0.

No visual artifact: this changes a config file loader's refusals and two documentation statements, not a policy, simulation behaviour, rendering, recording or a robot asset. The measured tables above are the evidence.

### DIFF
--- a/changelog.d/2736-kimodo-config-file-loader-refusals.md
+++ b/changelog.d/2736-kimodo-config-file-loader-refusals.md
@@ -1,0 +1,35 @@
+### Fixed
+
+- **policies/kimodo**: `KimodoConfig.from_json` now reports a config file it
+  cannot read the way its two sibling policy-config loaders already do. It
+  parsed the file and handed the result straight to `from_dict`, whose first act
+  is `data.items()`, so a file holding any JSON value other than an object
+  surfaced as `AttributeError: 'list' object has no attribute 'items'` - a
+  message naming a method of the parsed value rather than the file that could
+  not supply fields. Measured across the five non-object JSON values, all five
+  failed that way; all five are now `ValueError` naming the class, the resolved
+  path and the type found (`KimodoConfig file /tmp/cfg.json must contain a
+  mapping, got list`). A file that was not JSON escaped as a bare
+  `json.JSONDecodeError` and is now wrapped with the file named, a path that is
+  not a file raises `FileNotFoundError` naming the class instead of an errno
+  alone, and `~` is expanded - a config at `~/kimodo.json` was previously
+  reported missing while it existed, with the literal `~` in the message.
+  `MotionBricksConfig.from_file` and `WBCConfig.from_file` shipped all four
+  already, so this is two of three loaders' behaviour applied to the third, and
+  the new tests grade the rule over all three by discovering them rather than
+  listing them. The extension is deliberately still unchecked, unlike those two:
+  a JSON object stored under another name loads today, and refusing one would
+  stop a payload that currently works, so no input that loads today stops
+  loading.
+
+- **policies/kimodo**: the unrecognised-key policy is now documented as the one
+  the code implements. `from_dict` described the drop as happening "with a
+  warning"; measured, none of the three policy configs warns, and the other two
+  document the drop as silent forward compatibility, so the docstring was the
+  outlier. `docs/policies/kimodo.md` presents three interchangeable ways to set
+  a field and claimed a misspelled knob "raises `TypeError` at construction
+  instead of being silently ignored" - true for the two keyword forms, and not
+  for the `config` dict, which is read by `from_dict`:
+  `KimodoPolicy(config={"diffusion_stpes": 25})` builds with the default 100 and
+  emits nothing. The page now says which form refuses and which drops, and how
+  to have a typo refused.
--- a/docs/policies/kimodo.md
+++ b/docs/policies/kimodo.md
@@ -156,8 +156,29 @@ KimodoPolicy(config={"diffusion_steps": 25})           # a plain dict
 
 Precedence is per-field override > `config` field > the default in the table. A
 merged value is re-validated by `KimodoConfig`, so `diffusion_steps=0` is
-refused whichever way it arrives. There is no `**kwargs`: a misspelled knob
-raises `TypeError` at construction instead of being silently ignored.
+refused whichever way it arrives.
+
+A misspelled knob is refused by the two keyword forms and dropped by the dict
+form. Neither `KimodoPolicy` nor `KimodoConfig` takes `**kwargs`, so
+`KimodoPolicy(diffusion_stpes=25)` raises `TypeError` at construction. A
+`config` dict is read by `KimodoConfig.from_dict`, which drops keys that are not
+fields for forward compatibility - the policy the MotionBricks and WBC configs
+state for their own `from_dict` - so `KimodoPolicy(config={"diffusion_stpes":
+25})` builds with the default 100 and emits no warning. Pass the knob as a
+keyword, or build a `KimodoConfig` first, if a typo should be refused.
+
+### Loading a config from a file
+
+```python
+KimodoPolicy(config=KimodoConfig.from_json("~/kimodo.json"))
+```
+
+`from_json` expands `~` and names the file in every refusal: a path that is not
+a file raises `FileNotFoundError`, and a file that is not valid JSON or that
+holds a JSON value other than an object raises `ValueError` naming the file and
+what it found instead. Keys inside the object follow the drop policy above, and
+values keep the domains in the table. The extension is not checked, so a JSON
+object stored under any name loads.
 
 ## When the sampler runs again
 
--- a/strands_robots/policies/kimodo/config.py
+++ b/strands_robots/policies/kimodo/config.py
@@ -25,6 +25,7 @@ pipeline, so a real-model run supplies its sampler through ``motion_agent=``.
 
 from __future__ import annotations
 
+import json
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
@@ -161,16 +162,61 @@ class KimodoConfig:
 
     @classmethod
     def from_dict(cls, data: dict[str, Any]) -> KimodoConfig:
-        """Build a config from a plain dict (drops unknown keys with a warning)."""
+        """Build a config from a plain dict.
+
+        Only recognised keys are consumed; unknown keys are ignored for forward
+        compatibility, and no warning is emitted for a dropped key - the policy
+        :mod:`strands_robots.policies.motionbricks.config` and
+        :mod:`strands_robots.policies.wbc.config` state for their own
+        ``from_dict``.
+
+        Args:
+            data: Mapping of field name to value. Keys that are not fields of
+                this class are dropped.
+
+        Returns:
+            The config, validated by :meth:`__post_init__`.
+        """
         known = {f.name for f in cls.__dataclass_fields__.values()}
         kwargs = {k: v for k, v in data.items() if k in known}
         return cls(**kwargs)
 
     @classmethod
     def from_json(cls, path: str | Path) -> KimodoConfig:
-        """Load a config from a JSON file on disk."""
-        import json
+        """Load a config from a JSON file on disk.
 
-        p = Path(path)
-        with p.open("r", encoding="utf-8") as f:
-            return cls.from_dict(json.load(f))
+        A file that cannot supply fields is reported by name rather than
+        reaching :meth:`from_dict`, which is the reporting the sibling
+        policy-config file loaders in
+        :mod:`strands_robots.policies.motionbricks.config` and
+        :mod:`strands_robots.policies.wbc.config` already give. ``~`` in
+        ``path`` is expanded.
+
+        The extension is deliberately not checked, unlike those two loaders: a
+        JSON object stored under any name loads here today, and refusing one
+        would stop a payload that currently works. Every refusal below names an
+        input that already fails.
+
+        Args:
+            path: Path to a file holding a JSON object of config fields.
+
+        Returns:
+            The config, validated by :meth:`__post_init__`.
+
+        Raises:
+            FileNotFoundError: If ``path`` does not name a file.
+            ValueError: If the file is not valid JSON, or holds a JSON value
+                that is not an object. A value inside the object keeps the
+                domain :meth:`__post_init__` applies to it, so the loader adds
+                no second domain of its own.
+        """
+        p = Path(path).expanduser()
+        if not p.is_file():
+            raise FileNotFoundError(f"KimodoConfig file not found: {p}")
+        try:
+            data = json.loads(p.read_text(encoding="utf-8"))
+        except json.JSONDecodeError as e:
+            raise ValueError(f"KimodoConfig file {p} is not valid JSON: {e}") from e
+        if not isinstance(data, dict):
+            raise ValueError(f"KimodoConfig file {p} must contain a mapping, got {type(data).__name__}")
+        return cls.from_dict(data)
--- a/tests/policies/kimodo/test_config_file_loader_refusals.py
+++ b/tests/policies/kimodo/test_config_file_loader_refusals.py
@@ -1,0 +1,373 @@
+"""A policy config file loader reports a payload it cannot use, by name.
+
+``KimodoConfig.from_json`` read a JSON file and handed whatever it parsed
+straight to :meth:`KimodoConfig.from_dict`, which immediately calls
+``data.items()``. A file holding a JSON array, string, number, boolean or
+``null`` therefore surfaced as ``AttributeError: 'list' object has no attribute
+'items'`` - a message naming a method of the parsed value rather than the file
+that could not supply fields. A file that was not JSON at all escaped as a bare
+``json.JSONDecodeError``, and ``~`` in the path was never expanded, so a config
+at ``~/kimodo.json`` was reported missing while it existed.
+
+The two sibling policy-config file loaders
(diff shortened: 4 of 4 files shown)


yinsong1986 (dismissed):
## Summary

Brings `KimodoConfig.from_json` up to the refusal behaviour its two sibling policy-config file loaders (`MotionBricksConfig.from_file`, `WBCConfig.from_file`) already ship: a non-object JSON payload now raises `ValueError` naming the class, the resolved path and the JSON type found (instead of `AttributeError: ... no attribute 'items'` from `from_dict`), a decode failure is wrapped with the file named, a missing path raises `FileNotFoundError` naming the class, and `~` is expanded. The `from_dict` docstring drops the false "with a warning" claim, and `docs/policies/kimodo.md` corrects the claim that a misspelled knob is always refused (the `config` dict form drops it silently, per the documented forward-compatibility policy). The one divergence from the siblings - no extension check - is deliberate, documented, and pinned as a control so no input that loads today stops loading.

## What's good

- The new refusals only reach inputs that already failed (`AttributeError`/bare `JSONDecodeError` before), and `json.JSONDecodeError` is a `ValueError` subclass, so no caller's existing `except ValueError` breaks - no silent behaviour change.
- The test survey derives the loader set from `strands_robots/policies/*/config.py` rather than listing it, with a non-vacuity check pinning the three shipped loaders and a `_MINIMAL_PAYLOAD` guard so a fourth policy config is graded on arrival rather than skipped. Verified the sibling loaders' code matches every property the shared tests assert.
- Changelog lands as a `changelog.d/2736-*.md` fragment (AGENTS.md PR workflow step 3), the `Raises:` block on `from_json` names both refusal classes (AGENTS.md > API Consistency), and the docstring cross-references are contiguous `:mod:`/`:meth:` paths. Ran the new file (60 passed), the full kimodo suite (149 passed), and the repo graders (`test_raises_docstring_completeness`, `test_test_case_names_describe_behaviour`, `test_log_strings_are_ascii`, `test_no_host_paths` - 43 passed) at the head SHA; ruff check and format are clean, and no non-ASCII was added.

yinsong1986 (approved):
## Summary

Brings `KimodoConfig.from_json` up to the refusal behaviour its two sibling policy-config file loaders already ship: `~` is expanded, a missing path raises `FileNotFoundError` naming the class and resolved path, a decode failure is wrapped as `ValueError` naming the file, and a non-object JSON payload raises `ValueError` naming the class, path and type found instead of `AttributeError: ... no attribute 'items'` out of `from_dict`. Every new refusal targets an input that already failed with a worse message, and the wrapped `JSONDecodeError` stays inside the `ValueError` family it already was, so no caller's `except` narrows out from under it. The `from_dict` docstring and `docs/policies/kimodo.md` are corrected to state the drop policy the code actually implements. Verified at head `2311ebf4`: the 60-case test file passes, `ruff check` and `ruff format --check` are clean on both changed Python files, and the round-1 fix (misspelled knob moved into a dict unpacked at the call site) resolves the prior CodeQL/mypy finding.

## What's good

- The survey is derived from `strands_robots/policies/*/config.py` rather than listed, with a non-vacuity guard that fails (not skips) when a newly discovered loader lacks a minimal payload -- and the discovery asserts confirm protomotions/vera configs (which expose no `from_*` loaders) are correctly out of scope.
- The deliberate divergence from the siblings (no extension check) is pinned as a control with its rationale, instead of silently diverging.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; the new `Raises:` block names every refusal the loader makes; refusal messages are plain ASCII.

cagataycali:
## Review: both blockers are cleared, and one consequence of the unpack is worth closing separately

`2311ebf4` clears both. On `43affc21`: CodeQL is **neutral** with **0 open alerts** for this PR (`code-scanning/alerts?pr=2736&state=open`), the review thread it opened is resolved and outdated, and `mypy strands_robots tests tests_integ` reports `Success: no issues found in 1553 source files`. So nothing is owed on the alert, and the remaining `BLOCKED` is the in-progress `call-test-lint` alone.

Worth recording that the unpack is not just one of several ways to spell this - it is close to the only one. mypy still resolves a dict **literal** splat against the signature, and the annotation is what stops it, measured over four candidates for the same call:

| spelling | mypy |
|---|---|
| `KimodoPolicy(diffusion_stpes=25)` | 2x `call-arg` (the reported state) |
| `KimodoPolicy(**{"diffusion_stpes": 25})` | 4x `arg-type` |
| `misspelled: dict[str, object]` then `**misspelled` | 6x `arg-type` |
| `misspelled: dict[str, Any]` then `**misspelled` | clean |

### The consequence the unpack leaves open

Unpacking is what makes the call a runtime construct, and its cost is that no checker reads the keyword any more. Both cases still assert `match="diffusion_stpes"`, which is satisfied by any `TypeError` mentioning the key - including one raised by a hand-rolled check rather than by Python's signature enforcement. That distinction is not incidental here: `KimodoPolicy.__init__` documents it in its own docstring, "There is deliberately no `**kwargs`: an unknown knob raises `TypeError` at construction rather than being swallowed by a parameter nothing reads". Nothing grades that promise. It was graded only as a side effect of how the two calls were spelled, and the spelling is what changed.

Measured on this file, mutating `KimodoPolicy.__init__` to grow a `**unknown` that raises its own `TypeError` naming the key:

| row | mutation | fires |
|---|---|---|
| b0 | control, unmutated | 0 |
| b5 | `**kwargs` raising its own `TypeError`, against the assertions as they stand | **0 - invisible** |
| b1 | the same mutation, against `match="unexpected keyword argument '...'"` plus a signature pin | 2 |
| b1w | the same mutation, signature pin present but the bare-key match restored | 1 - the behavioural cell goes vacuous |
| b2 | `**kwargs` that swallows the key silently | 2 either way |
| b3 | `diffusion_steps` renamed, so the typo is no longer a near miss | 9, of which 8 pre-existing |

b5 against b1 is the whole point: the mutation a reader would most plausibly make to "improve" that error message is currently undetectable, and it converts a documented Python-level refusal into a hand-rolled one with every cell still green. b2 is the honest counterpart - a `**kwargs` that swallows the key raises nothing at all, so the bare-key match catches that one unaided; the gap is specific to a substituted refusal. b3 is why the non-vacuity cell is there: any unknown keyword raises, so without pinning that `diffusion_steps` exists these cells would keep passing on a name that is no longer a near miss of anything.

### Deliberately not pushed here

The `default` ruleset sets `require_last_push_approval=True` with `required_approving_review_count=1`, which is why the 21:30 approval was dismissed by the 22:11 push. Pushing a test-only refinement onto a branch that is now approved and waiting on one in-progress check would dismiss that approval and cost another review round for something that changes no shipped behaviour. This PR should merge on the approval it has.

The commit is on `cagataycali:test/kimodo-no-kwargs-promise` (`22cd86c5`), based directly on `43affc21` so it is a fast-forward, and it is a follow-up rather than an iteration precisely because the file it edits does not exist on `main` until this merges. It is `+71/-2` in `tests/policies/kimodo/test_config_file_loader_refusals.py` only: keeps `_MISSPELLED_KNOB` and its comment, adds `_UNEXPECTED_KEYWORD` beside it, points both assertions at it, and adds `TestNeitherConstructorAcceptsVarKeyword` reading `inspect.signature(cls).parameters` for a `VAR_KEYWORD` over both constructors, with the near-miss non-vacuity cell. Happy to open it as a PR once this lands, or to fold it in here if you would rather take the extra review round now.

### Gate on `43affc21` with that commit applied

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1556 files); `mypy` the same scope `Success: no issues found in 1553 source files`. `tests/policies/kimodo/` plus the two suites the base merge brought in (`test_presence_capability_probe_isolation.py`, `test_fleet_stop_reports_a_refused_rollout.py`) are **211 passed**; `scripts/assemble_changelog.py --check` reports `changelog fragments OK (767 pending)` with `tests/test_changelog_fragments.py` and `tests/test_changelog_format.py` 30 passed.

Separately, on `893ea731` before the unpack landed, an identical 370-file paired selection - all of `tests/policies/kimodo/` plus every test that walks the tree, reads docstrings or reads `docs/`, with the changed file excluded from both sides - collected 17319 items and gave `7 failed, 17214 passed, 98 skipped` on both trees, with the failing-id sets identical and `comm` empty in both directions. All 7 are environmental rather than related to this change: 2 need `docker` on `PATH`, and 5 assert `rclpy` is absent where it resolves.

yinsong1986 (approved):
## Summary

Brings `KimodoConfig.from_json` up to the refusal reporting its two sibling policy-config file loaders (`MotionBricksConfig.from_file`, `WBCConfig.from_file`) already ship: `~` is expanded, a path that is not a file raises `FileNotFoundError` naming the class and resolved path, a decode failure is wrapped as `ValueError` naming the file, and a non-object JSON payload raises `ValueError` naming the class, path and type found instead of the previous `AttributeError: ... no attribute 'items'` out of `from_dict`. Two false documentation claims are corrected in the same change: `from_dict`'s docstring said unrecognised keys are dropped "with a warning" (none of the three configs warns), and `docs/policies/kimodo.md` claimed a misspelled knob always raises `TypeError` when the `config` dict form silently drops it. The public signature is unchanged, and the over-reach controls pin that no input that loads today stops loading - including the deliberate divergence of leaving the extension unchecked.

## What's good

- The new tests derive the graded loader set from `strands_robots/policies/*/config.py` rather than listing it, with non-vacuity checks pinning that exactly the three shipped file loaders and three dict loaders are discovered - a fourth policy config's loader is graded on arrival.
- The one divergence from the siblings (no extension check) is pinned as a control (`test_a_json_object_under_a_non_json_extension_still_loads`) so it stays deliberate rather than incidental.
- The prior CodeQL finding (misspelled keyword as a literal argument) is resolved on this head by holding the knob in a mapping and unpacking it, keeping the refusal a runtime construct - with the reasoning recorded in the test file.
- Changelog fragment present per AGENTS.md step 3; `Raises:` block on `from_json` names both refusals the method itself makes; verified locally on the head commit: all 60 new tests pass, `ruff check` and `ruff format --check` clean on the changed files, and `tests/test_raises_docstring_completeness.py` plus the full `tests/policies/kimodo/` suite pass (167 passed).

PR #2737, opened 2026-08-25
fix(simulation): report a body's visual mesh, not its collision hull

cagataycali:
## Problem

MJCF bodies routinely carry two mesh geoms: the visual asset a renderer should show, and a convex
hull the solver should collide. `_find_body_mesh` decides which one a reported `SceneObject`
carries, and it picked *the first geom whose `group` is not `"0"`* — reading a non-default MuJoCo
group as the visual marking.

That is backwards for the dominant convention. MuJoCo Menagerie marks a **visual** geom with
`contype="0" conaffinity="0"` and a **collision** geom with `group="3"`, and routinely declares the
collision geom first. On the shipped `shadow_dexee` hand the visual class carries no `group` at all:

```xml
<default class="finger/visual"><geom type="mesh" contype="0" conaffinity="0"/></default>
<default class="finger/collision"><geom type="mesh" group="3" material="collision" mass="0"/>
  <default class="finger/collision_hard"><geom condim="4" .../></default></default>
```

so `F0/finger_base` skipped its **eight** visual mesh geoms for the ninth. Measured through the
public reader on that shipped asset:

| body | `load_mjcf_scene_objects` before | after |
|---|---|---|
| `hand_base` | `Asm-MRH-HB1-Visual,00-Plastic.stl` | unchanged |
| `F0/` `F1/` `F2/` | `r3_finger_base_col.stl` | `MRH-FB-MainALU-Visual,00.stl` |

One reader, one file, two answers, under a successful load. `group` cannot be the signal in any
case: MuJoCo attaches no meaning to it — it is a visualiser toggle — and the conventions built on it
disagree, robosuite emitting visual geoms as `group="1"` where Menagerie spells *collision* as
`group="3"`.

## Change

Rank the candidates, strongest first, instead of returning the first non-default group.

1. A geom MuJoCo **cannot collide with**. MuJoCo lets two geoms touch only when
   `contype1 & conaffinity2` or `contype2 & conaffinity1` is non-zero, so a geom declaring both as
   `0` can never collide with anything and exists purely to be looked at. That is the format's own
   statement of intent, so it ranks first.
2. A **non-default `group`**, kept as the weaker hint — taken when the contact declaration says
   nothing, never over a contact-free geom.
3. Anything else.

Document order still breaks a tie, so a subtree carrying no visual marking reports the mesh it
always did. A stronger candidate in a nested body wins over a weaker one on the body itself, which
is the shape `shadow_dexee` has. Both signals resolve through `_class_attrs`, so a geom inheriting
them from a `<default class="visual">` is ranked as MuJoCo reads it.

## Graded across the downloaded asset corpus

554 MJCF files, 2736 bodies declaring a mesh geom, both readings compared body by body:

| | |
|---|---|
| bodies reporting a mesh, before and after | 2736 / 2736, identical key set |
| bodies whose reported mesh changes | **165**, in 18 files |
| changed to a geom MuJoCo cannot collide with | **165 of 165** (162 from a group-marked hull, 2 from an unmarked geom, 1 tie between two contact-free geoms) |
| bodies whose reported mesh ranks *worse* after | **0** |
| position/orientation changes where the mesh is unchanged | **0** |
| exceptions, either reading | **0** |

The files include the canonical Menagerie assets `shadow_dexee` (19 bodies), `franka_fr3`,
`franka_fr3_v2`, `boston_dynamics_spot/spot_arm` and `pal_tiago_dual`, plus `openarm_bimanual` (25)
and the `ability_hand_api` arms.

`mujoco_menagerie/shadow_dexee`, body `F0/finger_base`, rendered from the shipped STLs. One camera
and one headlight for all three panels — `mjv_defaultFreeCamera` derives its distance from each
model's own extent, so the frame is fixed once from the ground-truth model and reused, which is what
makes the panels comparable.

MJCF visual mesh vs collision hull

The reported visual geom is one of the assembly's own parts at its own pose, and its silhouette
lies inside the assembly's to within **0** pixels. The hull is not a part of it and covers **7989**
pixels the whole visual assembly does not; the two reported assets' silhouettes differ in **13461**
pixels.

## Tests

`tests/simulation/isaac/test_mjcf_visual_mesh_preference.py`, 30 cases. Every premise is derived
from `mujoco.MjModel` rather than restated: MuJoCo is asked which geoms its compiler lets touch, and
that the fixture's `group="3"` really is inherited through the nested default class the geom cites.

Pre-fix split, with only the discriminator reverted so the module still imports:

| class | failing cells | test functions |
|---|---|---|
| `TestABodyReportsItsVisualMesh` (regression) | **4** | 3 |
| `TestMuJoCoFixesWhichGeomsCanTouch` (premise) | 0 | 3 |
| `TestTheWeakerHintStillDecidesWhereItIsTheOnlySignal` (over-reach control) | 0 | 6 |
| `TestWhenAGeomCannotCollide` (domain) | 0 | 5 |
| `TestAMeshAssetThatNamesNoFile` | 0 | 2 |

A raw revert is a collection error instead — the tests import the new rank constants.

11 mutations, 10 detected in 10 distinct failing-id sets, unmutated control clean, source restored
byte-identical, and 0 failures in `test_mjcf_default_class_geometry.py` (which structurally requires
`_find_body_mesh` to resolve its attributes through `_class_attrs`) for every row:

| mutation | fires |
|---|---|
| discriminator reverted to `group != "0"` | 4 |
| contact-free checks `contype` only | 3 |
| contact-free accepts either half zero | 17 |
| an omitted `contype`/`conaffinity` reads as zero | 14 |
| an unparseable value reads as contact-free | 2 |
| the group hint outranks a contact-free geom | 1 |
| the group tier dropped rather than demoted | 5 |
| a later geom of equal rank wins the tie | 1 |
| a nested candidate taken only when nothing was found | 1 |
| the nested body's own offset not folded in | 2 |
| the contact-free early return dropped from the geom loop | **0** |

The last row is reported rather than hidden: without that early return `best` still selects the
contact-free candidate, because rank 0 is minimal and ties keep the first, so it is a short-circuit
whose removal is behaviourally invisible. It does still stop the walk from descending into the rest
of a subtree once the answer is settled.

The two branches this reaches were previously uncovered, so the file also closes them: a visual mesh
found in a **nested** body, and a `<mesh>` asset declaring no `file`. `isaac/loaders.py` goes from
**73 uncovered lines (90%) to 68 (91%)**, and every line still uncovered is either inside `load_usd`
and its `pxr` import guard or unreachable from its only caller.

## Gate

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1554 files).
mypy **24 == 24** against a pristine `main` worktree, normalized message sets byte-identical in both
directions, 0 in the changed files. `tests/simulation/isaac/` 1540 passed, 5 skipped.

Paired run over an identical 660-file selection — all of `tests/simulation/` plus every guard that
walks the tree, reads a docstring or names this module — with the new file excluded from both trees:
**23595 collected and `17 failed, 23329 passed, 249 skipped` on both** (928s vs 945s), identical
failing-ID sets, both `comm` directions empty. All 17 are `tests/mesh/test_iot_*` needing `awsiot`
from the `[mesh-iot]` extra, confirmed absent.

## Deliberately unchanged

A subtree where no geom is contact-free. `google_barkour_v0` has one set of meshes serving both
roles, so its `group="1"` marking is the only signal in the file and it still decides — dropping the
group tier instead of demoting it would have cost those 5 bodies their answer, which is why that
mutation is in the table and the case is a control. The rank is internal to the pick: the one
production caller discards it, so nothing downstream of the reader gains a field.

### DIFF
--- a/changelog.d/2737-mjcf-visual-mesh-preference.md
+++ b/changelog.d/2737-mjcf-visual-mesh-preference.md
@@ -1,0 +1,40 @@
+### Fixed: a body's reported mesh is its visual asset, not its collision hull
+
+MJCF bodies routinely carry two mesh geoms: the visual asset a renderer should show, and a convex
+hull the solver should collide. `_find_body_mesh` picks the one a reported `SceneObject` carries,
+and it picked "the first geom whose `group` is not `"0"`" - reading a non-default MuJoCo group as
+the visual marking.
+
+That is backwards for the dominant convention. MuJoCo Menagerie marks a visual geom with
+`contype="0" conaffinity="0"` and a collision geom with `group="3"`, and the collision geom is
+routinely declared first, so the rule returned the hull. In the shipped `shadow_dexee` hand the
+visual class carries no `group` at all - `<default class="finger/visual"><geom contype="0"
+conaffinity="0"/>` - so each finger skipped its eight visual mesh geoms for the ninth,
+`r3_finger_base_col`. `load_mjcf_scene_objects` reported those fingers as `r3_finger_base_col.stl`
+under a successful load while `hand_base` in the same file reported its visual STL: one reader,
+one file, two answers. `group` also cannot be the signal because the conventions built on it
+disagree - robosuite emits visual geoms as `group="1"` where Menagerie spells collision as
+`group="3"` - and MuJoCo attaches no meaning to it at all, it being a visualiser toggle.
+
+The candidates are now ranked, strongest first, rather than the first non-default group winning.
+MuJoCo lets two geoms touch only when `contype1 & conaffinity2` or `contype2 & conaffinity1` is
+non-zero, so a geom declaring both as `0` cannot collide with anything and exists purely to be
+looked at; that is the format's own statement of intent and it ranks first. A non-default group
+stays as the weaker hint, taken when the contact declaration says nothing and never over a
+contact-free geom. Document order still breaks a tie, so a subtree carrying no visual marking at
+all reports the same mesh it always did. A stronger candidate in a nested body wins over a weaker
+one on the body itself, which is the shape `shadow_dexee` has.
+
+Graded across the downloaded asset corpus - 554 MJCF files, 2736 bodies declaring a mesh geom -
+165 bodies in 18 files change their reported mesh and every one of them changes to a geom MuJoCo
+cannot collide with: 162 from a group-marked hull, 2 from an unmarked geom, and one tie between
+two contact-free geoms. No body moves the other way, no reported position or orientation changes
+for a body whose mesh is unchanged, and neither reading raises. The files include the canonical
+Menagerie assets `shadow_dexee` (19 bodies), `franka_fr3`, `franka_fr3_v2`,
+`boston_dynamics_spot/spot_arm` and `pal_tiago_dual`.
+
+Deliberately unchanged: a subtree where no geom is contact-free. `google_barkour_v0` has one set
+of meshes serving both roles, so its `group="1"` marking is the only signal in the file and it
+still decides; dropping the group tier instead of demoting it would have cost those bodies their
+answer, and that is pinned as a control. The rank is internal to the pick - the one production
+caller discards it - so nothing downstream of the reader gains a field.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -1572,6 +1572,70 @@ def _parse_mjcf_mesh_assets(root: ET.Element, mjcf_dir: str) -> dict[str, tuple[
     return registry
 
 
+# How strongly a mesh geom's own declaration reads as its body's visual asset,
+# best first. Used only to order candidates within one body subtree.
+_MESH_VISUAL_RANK_NON_COLLIDING = 0
+_MESH_VISUAL_RANK_NON_DEFAULT_GROUP = 1
+_MESH_VISUAL_RANK_UNMARKED = 2
+
+
+def _geom_cannot_collide(attrs: dict[str, str]) -> bool:
+    """Whether a geom's own declaration makes it unable to collide with anything.
+
+    MuJoCo lets two geoms touch only when ``contype1 & conaffinity2`` or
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

This PR fixes `_find_body_mesh` in `strands_robots/simulation/isaac/loaders.py` so a body's reported `SceneObject` mesh is its visual asset rather than its collision hull. The old rule ("first geom whose `group` is not `\"0\"`") reads Menagerie-convention files backwards, where visual geoms are marked `contype="0" conaffinity="0"` (often with no `group` at all) and *collision* geoms carry `group="3"`. The fix replaces the boolean pick with a three-tier rank: contact-free (MuJoCo's own semantics: `contype1 & conaffinity2 | contype2 & conaffinity1` non-zero is the only way two geoms touch) ranks first, non-default `group` stays as a weaker hint, document order breaks ties. The rank is internal — the one production caller discards the fourth tuple element, `SceneObject`'s shape is unchanged, and all touched symbols are `_`-private, so there is no public-API or persisted-schema impact.

## What's good

- The rank logic verifies correctly against MuJoCo's contact rule, including the asymmetric case (`contype="0"` alone still collidable) and the unparseable-value fallback to "colliding", which is the conservative direction.
- Over-reach controls are pinned: group-only files (`google_barkour_v0` shape) and unmarked subtrees keep their prior answer, so the change is scoped to bodies carrying a contact-free geom.
- The premise tests derive contact semantics from `mujoco.MjModel` rather than restating them, and class inheritance through nested `<default>` blocks is exercised, matching how `_class_attrs` resolves the attributes the rank reads.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention and quantifies the blast radius over the asset corpus, including the deliberate non-change.
- Verified locally at `afec2e99`: the 30 new tests pass, the pre-existing mesh-preference suites (`test_scene_object_meshes.py`, `test_mesh_assets.py`, `test_description_loader_geometry.py`, 92 passed / 1 skipped) are unaffected, and `ruff check` / `ruff format --check` are clean on both changed Python files.

PR #2738, opened 2026-08-25
fix(kimodo): decay the segment seam's rotation offset instead of the orientation

cagataycali:
## Problem

`_ease_onto_previous_pose` absorbs the offset between a freshly sampled segment's
start pose and the pose last commanded. Its docstring states the guarantee: "the
motion keeps its own shape and velocity and only its starting offset is
absorbed". For the linear channels the arithmetic plainly does that -- one offset
scaled by the frame's weight is added to every eased frame, so the correction a
frame receives is decided by the weight alone:

```python
out[i, linear] = qpos[i, linear] + weight * offset
```

The root's orientation was handled the other way round -- every eased frame was
interpolated toward the previous pose's orientation:

```python
out[i, 3:7] = _slerp_quat(q_i, q_prev, weight)   # before
```

so its correction depended on the frame's own orientation. Two consequences,
both measured through the shipped function at the default `transition_frames=5`
on a 180 deg/s turn in place (6.00 deg per native frame):

| seam | before | after | the motion |
| --- | --- | --- | --- |
| orientations already agree (0.0 deg offset) | root moved up to **6.000 deg** | 1.7e-06 deg (float32 rounding) | -- |
| 40 deg heading offset, eased yaw rate | **8.67 -> 10.67 -> 12.67 -> 14.67 -> 16.67** deg/frame | uniform **12.67** deg/frame | 6.00 deg/frame |

The first row is the sharper one: a seam with nothing rotational to absorb still
had its root turned, by as much as one whole frame of the motion's own rotation.
The second is the shape defect -- a uniform turn came out as a ramp, where the
linear channels produce the motion's own step plus one *constant* share of the
decaying offset. With a wide offset the ramp can invert: at 200 deg the eased
yaw increments measured 28.67, 30.67, 32.67, **94.67, -23.33** deg/frame, i.e.
the root reverses direction inside the transition.

## Change

The offset is now the world-frame rotation carrying the segment's own start
orientation onto the pose last commanded, and it is that rotation which decays:

```python
q_offset = _quat_mul(q_prev, _quat_conjugate(q_start))
...
out[i, 3:7] = _quat_mul(_slerp_quat(_IDENTITY_QUAT, q_offset, weight), q_i)
```

Pre-multiplication is the rotational analogue of adding a world-frame
translation, which is what the position channel does. Three properties follow
and each is pinned:

* **The first eased frame is unchanged.** Interpolation is equivariant under
  right-multiplication, so `slerp(I, q_prev * q_start^-1, w) * q_start` is
  exactly `slerp(q_start, q_prev, w)`. Measured: identical to the last bit.
  The seam therefore still lands one share short of `previous_pose`, which is
  the existing deliberate choice (pinning it exactly would stall velocity for a
  frame).
* **A seam with no orientation offset is the identity on the root.** Byte-exact
  when the start orientation is the identity, and to float32 rounding otherwise.
* **The emitted action dict is unchanged.** `get_actions` returns joint targets
  (`frame[7:]`); the root pose this touches is the internal reference the next
  seam eases onto. A cell drives the policy through a prompt change with a
  turning stub sampler and checks every commanded joint value against the closed
  form of the linear ease.

`_quat_mul` and `_quat_conjugate` are local to the module, for the reason
`simulation/isaac/loaders._quat_from_basis` already gives for its own copy: the
other two wxyz products in the tree are private to `simulation/isaac` and the
LIBERO adapter, and the public one in `policies/protomotions/state_utils` is
xyzw, so using it would mean reordering at both ends of every call.

## Why nothing caught this

`tests/policies/kimodo/test_chained_prompt_transition_is_continuous.py` already
covers the transition, including
`test_the_transition_preserves_the_motion_s_own_velocity`. Its stub sampler
writes an identity quaternion into every frame of every motion (`out[:, 3] =
1.0`), so the root never rotates there -- and with a constant orientation an
absolute pull and a decaying offset are indistinguishable. The distinguishing
input is a motion whose root actually turns. Every mutation in the table below
fires **0** tests in that pre-existing suite.

## Tests

New file, 41 cases. Reverting only `strands_robots/policies/kimodo/policy.py` to
`main` gives **18 failed, 23 passed, 0 errors**, and the split is by class:

| class | failed/total |
| --- | --- |
| `TestThePremisesOfTheseCases` | 0/4 |
| `TestASeamWithNoOrientationOffsetLeavesTheRootAlone` | 2/2 |
| `TestTheCorrectionIsDecidedByTheWeightAlone` | 11/2 (parametrized) |
| `TestTheCorrectionIsAWorldFrameRotationOfTheOffset` | 2/2 |
| `TestAUniformTurnStaysUniformThroughTheTransition` | 3/3 |
| `TestTheSeamItselfIsUnchanged` | 0/2 |
| `TestTheLinearChannelsAndTheGuardsAreUnchanged` | 0/6 |
| `TestTheEmittedActionIsUntouched` | 0/1 |

Every premise and control class passes pre-fix; all 18 failures are in the four
regression classes.

The convention cases state the expected orientation in **matrices** rather than
quaternions, so the quaternion side has an independent oracle: the offset is read
off `M(q_prev) @ M(q_start).T` as an axis and angle, and each eased frame is
compared against `R(axis, weight * angle) @ M(frame)`.

### Mutation table

Each row mutates the source, then runs the new file and, separately, the rest of
`tests/policies/kimodo`.

| # | mutation | new file | rest of the kimodo suite |
| --- | --- | --- | --- |
| b1 | each frame pulled toward the previous orientation (the defect) | 18 | 0 |
| b2 | offset taken in the body frame and post-multiplied | 2 | 0 |
| b3 | conjugate applied to the wrong operand | 8 | 0 |
| b4 | offset applied in full every frame (never decays) | 24 | 0 |
| b5 | the weight runs the wrong way | 17 | 0 |
| b6 | the conjugate does not invert | 24 | 0 |
| b7 | sign slip in one component of the product | 3 | 0 |
| b8 | the arc starts somewhere other than the identity | 27 | 0 |
| b9 | the start orientation is not normalised | 1 | 0 |
| b10 | the product operands are swapped | 3 | 0 |

10 of 10 detected, 9 distinct firing sets, control clean, source restored
byte-identical. b7 and b10 share a set: both are errors in the product itself
and only the matrix oracle sees either. b2, b6, b7 and b10 all fired 0 against
an earlier fixture that rotated only about +Z from an identity start -- rotations
about one axis commute, and inverting the identity is a no-op -- which is why the
fixtures now use a tilted turn axis, an unrelated seam orientation and a
non-identity start.

`_angle_between` in the test file deliberately avoids `2 * arccos(|dot|)`:
`arccos` is ill-conditioned at 1, so that form reports about 0.04 deg for a
float32 quaternion compared against itself and the "nothing moved" assertion
cannot be written at all. `4 * arcsin(d / 2)` on the closer of `a - b` and
`a + b` reports 0 for equal inputs, pinned by its own case.

## Visual

Unitree G1 (`mujoco_menagerie/unitree_g1/g1.xml`, 29 joints on a free base),
rendered headless with `MUJOCO_GL=egl`. The reference turns in place; the seam
shown carries a joint offset only, its orientation offset measuring 0.0 deg, so
an ease that absorbs only the offset it was given must leave the root exactly
where the sampler put it. Row 3 is bitwise identical to row 1 in all five eased
frames; row 2 differs by 0, 12740, 15348, 17800 and 18374 pixels.

kimodo transition root turn

## Gate

* `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1558 files).
* `mypy strands_robots tests tests_integ`: branch-only message set **empty**
  against a pristine worktree of the base commit (the 19 base-only messages are
  all `examples/isaac_gs/`, a cold-cache artifact of the throwaway worktree). 0
  messages in either changed file.
* Full `pytest tests/`: **29 failed, 39053 passed, 309 skipped, 24 errors**
  against **29 failed, 38918 passed, 309 skipped, 24 errors** on the base.
  Failing and erroring id sets are identical, 53 each, `comm` empty in both
  directions. All 53 are pre-existing and dependency-driven: 44 need `awsiot`
  (declared in the `mesh-iot` extra), 3 are lerobot-from-source
  `encode() takes 1 positional argument`, the rest missing dist metadata.
* `+41` collected is attributable to this branch, measured:
  `tests/test_args_docstring_completeness.py` collects 444 on both trees, so the
  two new private helpers add nothing to it. The other `+94` belongs to the two
  PRs that merged into `main` during the same window.
* Coverage of `strands_robots/policies/kimodo/policy.py`: **93% (10 missing) ->
  99% (1 missing)**. The previously unexercised branches were the whole of
  `_slerp_quat` apart from its nearly-parallel shortcut -- the shorter-arc sign
  flip and the real-arc arithmetic -- which the new cases reach through wide and
  antipodal seam offsets. Package total missing 1861 -> 1843.
* `tests/policies/kimodo/`: 130 passed on mujoco 3.5.0 (the declared floor) with
  lerobot 0.6.2 from source.

### DIFF
--- a/changelog.d/2738-kimodo-transition-root-turn.md
+++ b/changelog.d/2738-kimodo-transition-root-turn.md
@@ -1,0 +1,25 @@
+### Fixed
+
+- **policies/kimodo**: the segment-transition ease now decays the seam's
+  *rotation* offset rather than dragging each frame's orientation toward the
+  pose last commanded, so a motion that turns keeps its own turn rate through
+  the transition. The linear channels (root position, joint angles) have always
+  added one offset scaled by the frame's weight, which makes the correction a
+  function of the weight alone; the root's orientation was interpolated toward
+  `previous_pose` instead, so its correction depended on the frame's own
+  orientation. Measured on a 180 deg/s turn in place at the default five-frame
+  transition, the eased yaw rate ramped 8.67 -> 16.67 deg/frame against the
+  motion's uniform 6.00, and a seam whose orientations already agreed - nothing
+  rotational to absorb - still moved the root by up to 6.000 deg. The offset is
+  now the world-frame rotation carrying the segment's own start orientation onto
+  the pose last commanded, applied by pre-multiplication of the decayed arc, so
+  the eased yaw rate holds a uniform 12.67 deg/frame (the motion's 6.00 plus one
+  constant share of the decaying offset, exactly the shape the position and
+  joint channels produce) and a seam with no orientation offset leaves the root
+  untouched. The first eased frame is unchanged, because right-multiplication
+  commutes with the interpolation there, and the emitted action dict is
+  unchanged: `get_actions` returns joint targets, so the root pose this touches
+  is the internal reference the next seam eases onto. The existing transition
+  suite could not see any of this - its stub sampler writes an identity
+  quaternion into every frame, where an absolute pull and a decaying offset are
+  indistinguishable.
--- a/docs/policies/kimodo.md
+++ b/docs/policies/kimodo.md
@@ -253,6 +253,15 @@ wider pose gaps (opposing poses eased over more frames), lower it toward 1 to
 keep segment boundaries crisp. It is bounded below at 1, matching the domain the
 sampler enforces on its own transition length.
 
+Every channel absorbs its share of the offset the same way, so easing changes
+where a segment starts and not how it moves. Root position and joint angles take
+one offset scaled by the frame's weight. The root's orientation takes the
+rotational form of exactly that: the offset is the rotation carrying the
+segment's own start orientation onto the pose last commanded, and each eased
+frame is its own orientation turned by the weighted share of it. A motion that
+turns therefore keeps its turn rate through the transition, and a seam whose
+orientations already agree leaves the root alone.
+
 Note the difference in kind from the sampler's native multi-prompt path: Kimodo
 conditions the *diffusion* of the next segment on the previous segment's last
 frames, so the generated motion itself leads into the transition. The agent
--- a/strands_robots/policies/kimodo/policy.py
+++ b/strands_robots/policies/kimodo/policy.py
@@ -106,6 +106,41 @@ class KimodoMotionAgent(Protocol):
         ...
 
 
+#: Identity rotation, wxyz. The start of the arc a decaying orientation offset
+#: walks: at weight 0 the offset contributes nothing and a frame is emitted as
+#: the sampler produced it.
+_IDENTITY_QUAT = np.array([1.0, 0.0, 0.0, 0.0], dtype=np.float32)
+
+
+def _quat_conjugate(q: NDArray[np.float32]) -> NDArray[np.float32]:
+    """Conjugate of a wxyz quaternion, which inverts a unit rotation."""
+    return np.array([q[0], -q[1], -q[2], -q[3]], dtype=np.float32)
+
+
+def _quat_mul(a: NDArray[np.float32], b: NDArray[np.float32]) -> NDArray[np.float32]:
+    """Hamilton product ``a * b`` of two wxyz quaternions.
+
+    Local for the same reason the MJCF loader and the Newton backend keep their
+    own copies: the two other wxyz products in the tree are private to
+    :mod:`strands_robots.simulation.isaac.loaders` and the LIBERO adapter, and
+    reaching across a package for a private symbol is worse than a four-line
+    product. The public one in
+    :mod:`strands_robots.policies.protomotions.state_utils` is xyzw, so using it
+    would mean reordering at both ends of every call.
+    """
+    aw, ax, ay, az = (float(a[0]), float(a[1]), float(a[2]), float(a[3]))
+    bw, bx, by, bz = (float(b[0]), float(b[1]), float(b[2]), float(b[3]))
+    return np.array(
+        [
+            aw * bw - ax * bx - ay * by - az * bz,
+            aw * bx + ax * bw + ay * bz - az * by,
+            aw * by - ax * bz + ay * bw + az * bx,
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

This PR fixes the Kimodo segment-transition ease so the root orientation channel behaves like the linear channels: instead of slerping each eased frame toward the pose last commanded (which makes the correction depend on the frame's own orientation, reshaping a turning motion's turn rate and moving the root even when the seam carries no rotational offset), it computes the world-frame offset `q_prev * conj(q_start)` once and decays it by pre-multiplying each frame with `slerp(identity, offset, weight)`. The math checks out: the correction is now a function of the decay weight alone, and by slerp right-invariance the first eased frame is byte-equivalent to the old behaviour, so the seam step itself is unchanged. The change is confined to a private helper; the emitted action dict (joint targets only) is untouched and pinned by a dedicated test.

## What's good

- The 536-line regression suite uses matrix arithmetic (Rodrigues / axis-angle) as an independent oracle for the quaternion path, includes premise tests so the fixtures can't degenerate into vacuous passes, uses a numerically well-conditioned angle metric (`4*arcsin(d/2)` instead of `2*arccos(|dot|)`) so "nothing moved" is assertable at all, and covers the non-commuting case (tilted turn axis vs. seam axis) that a yaw-only fixture cannot distinguish.
- Controls pin the unchanged behaviour: first-frame seam placement, frames past the window, linear-channel arithmetic, non-positive window guard, input non-mutation, unit-norm outputs, and the >180-degree offset short-arc case.
- Changelog fragment follows the `changelog.d/` convention; docs updated in the same PR; the local `_quat_mul` documents why it does not reach for the xyzw public helper or a private cross-package symbol.
- Verified locally at head `277cad5e`: all 190 `tests/policies/kimodo/` tests pass, ruff check/format clean, no non-ASCII in the diff.

PR #358, opened 2026-06-06
test(mesh): fix flaky test_session_config (leaked zenoh MagicMock isolation)

cagataycali:
## Problem

`tests/mesh/test_session_config.py` intermittently fails in CI with:

```
TypeError: the JSON object must be str, bytes or bytearray, not MagicMock
```

(seen on PR #319's `call-test-lint` run, and reproducible locally under specific full-suite collection orderings).

## Root cause

Several mesh tests install a mock Zenoh via `patch.dict("sys.modules", {"zenoh": MagicMock()})`. `patch.dict` restores correctly on its own, but under some full-suite orderings a `unittest.mock` instance bound to `sys.modules['zenoh']` outlives its intended scope. A later `_build_config()` then does a fresh `import zenoh`, gets the mock, and `zenoh.Config().get_json("namespace")` returns a `MagicMock` → `json.loads(MagicMock)` raises the `TypeError`.

This is **not** a regression in any feature PR — it is a pre-existing cross-test isolation flake. PR #319 happened to surface it.

## Fix

Add an autouse fixture in `tests/mesh/conftest.py` that drops any `unittest.mock` instance bound to `sys.modules['zenoh']` before each mesh test, so the next `import zenoh` re-binds the real wheel (or cleanly `ImportError`s where zenoh is genuinely absent). It can never evict the real extension module — the real `zenoh` is never from the `unittest.mock` package.

## Verification

- With a **deliberately leaked** `zenoh` MagicMock injected before `test_session_config`: **fails without** the fixture, **passes with** it.
- Full mesh suite: **1242 passed, 5 skipped**.

PR #357, opened 2026-06-06
docs(mesh): publish CA pin rotation runbook (closes #250)

cagataycali:
Closes #250.

Ships **Option B** (documented rotation runbook) from the issue rather than the heavier signed-manifest fetch (Option A).

## Changes
- **README.md** — new `### CA Pin Rotation Runbook` section:
  - Promotes the recompute one-liner out of the `provision.py` docstring.
  - Documents the dual-pin grace-period strategy: ship a release with both new and old pins, wait for fleet uptake, drop the old pin in a follow-up release.
  - How to monitor AWS security bulletins for upcoming root CA rotations.
  - `STRANDS_MESH_CA_PINS` emergency out-of-band override guidance (preferred over `STRANDS_MESH_DISABLE_CA_PIN`).
- **AGENTS.md** — Review Learnings entry: a static security pin must ship with a rotation runbook, and the accepted-pin surface must be a collection so the grace period is expressible.
- **tests/mesh/test_iot_ca_pin_rotation.py** — pins the dual-pin contract (acceptance criterion: 'asserts that the dual-pin grace-period state parses correctly').

## Acceptance criteria
- [x] Rotation runbook published in README.md (Option B).
- [x] AGENTS.md learnings entry for the operational-fragility-of-static-pins lesson.
- [x] Pin test `tests/mesh/test_iot_ca_pin_rotation.py` asserts dual-pin grace-period state parses correctly.

## Test output
```
hatch run python -m pytest tests/mesh/test_iot_ca_pin_rotation.py -q
============================== 4 passed in 0.82s ===============================
```

PR #356, opened 2026-06-06
docs(mesh): pin IoT Subscribe/Receive asymmetry (closes #253)

cagataycali:
Closes #253.

## What
Documents the deliberate asymmetry between `AllowOwnSubscriptions` (Subscribe to own `${ThingName}/*`) and `AllowReceiveScoped` (Receive only `/cmd`, `/response/*`, broadcast, estop, presence) in the robot IoT policy, and pins it with a regression test.

## Findings (issue action items)
1. Grepped `iot_transport.py` subscribe paths: no code round-trips a message through own `${ThingName}/health|state|safety/event` and relies on the local subscriber receiving it. The narrow Receive list is a no-op for runtime behavior -- those topics are publish-only at the robot. No bug, just a doc/pin gap.
2. Added a design-note comment block on `AllowOwnSubscriptions` in `strands_robots/mesh/iot/provision.py` explaining the intent and warning against widening Receive back to `${ThingName}/*`.
3. Added regression test `test_own_health_subscribable_but_not_receivable` in `tests/mesh/test_iot_policy_scope.py` asserting Subscribe covers own `${ThingName}/*` while Receive omits the broad wildcard and health/state/safety-event.

## Test
```
hatch run python -m pytest tests/mesh/test_iot_policy_scope.py -k "asymmetry or own_health or scoped_receive" -q
======================= 2 passed, 7 deselected in 0.82s ========================
```

PR #355, opened 2026-06-06
test(mesh): pin DEFAULT_BRIDGE_SUFFIXES against documented default set

cagataycali:
Closes #241.

DEFAULT_BRIDGE_SUFFIXES already landed (PR #222) but had no exact-equality pin tying it to the documented bridge-by-default set. This adds two regression pins in tests/mesh/test_bridge_transport.py:

- test_default_bridge_suffixes_pinned_to_documented_set: asserts exact frozenset membership (presence, health, safety/event, safety/estop, safety/resume, cmd, response, broadcast). Note: the live constant includes safety/resume, which the original issue body's 7-tuple omitted; the pin matches the actual shipped constant.
- test_high_volume_telemetry_not_bridged_by_default: negative pin for state/pose/imu/odom/lidar/camera/input/hand.

Test output:

    2 passed, 43 deselected in 0.97s

No production code changed; test-only follow-up.

PR #2739, opened 2026-08-26
feat(drivers): native CycloneDDS driver for the Unitree G1

cagataycali:
Closes cagataycali/robots-harness#354.

The driver seam from #2734 chose an implementation but no native driver was registered — every `mode="real"` call still built a lerobot driver, so a robot lerobot cannot model was reached through a bus it does not speak. The G1 is the shape that made the seam worth building: raw Unitree IDL over CycloneDDS, a state machine that gates arm and locomotion writes, a Livox Mid-360 producing a point cloud lerobot's dataset schema has no place for.

`Robot("g1", mode="real")` now builds `strands_robots.drivers.g1.G1Driver`, because the registry declares `hardware.driver = "strands"` on `unitree_g1` and `driver="auto"` honours that declaration. A caller who wants the lerobot path back can still ask for it explicitly with `driver="lerobot"`.

## What lands

- `strands_robots/drivers/g1.py::G1Driver` — implements `HardwareDriver` structurally. Subscribes `rt/lowstate`, `rt/lf/bmsstate`, `rt/utlidar/lidar_state` and `rt/utlidar/cloud_livox_mid360` on a background DDS thread; each callback drops into an in-memory cache the mesh reads at its own cadence, filling the `_imu`, `_battery`, `_lidar_state` and `_lidar_summary` sensor attrs that `SensorLoopsMixin` already publishes.
- `strands_robots/tools/g1/{_g1_common,_dds_engine}` — the vendored DDS layer, engine only. The agent tools (`g1_arm`, `g1_locomotion`, `g1_speak`, ...) are #358; they share `_g1_common`'s `_DDS_INIT_LOCK` and `ensure_dds` singleton so the driver and the tools never subscribe the Livox cloud twice.
- `strands_robots/drivers/__init__.py` — auto-registers `G1Driver` for `g1` and `unitree_g1` on import. Guarded so a broken driver import doesn't take down the seam.
- `strands_robots/registry/robots.json` — sets `unitree_g1.hardware.driver = "strands"`.
- `tests/drivers/test_g1_driver.py` — 30 tests, DDS fully mocked. Protocol conformance, factory wiring, sensor decoders (with fake IDL messages), FSM gate refusals, battery-floor gate, lifecycle envelope shapes, `connect_eagerly` idempotence, `ensure_dds` on a machine without the SDK.
- `tests/test_driver_seam.py` — isolation fixture reset to empty (auto-registered drivers would otherwise decide what the empty-table refusal tests see).
- `examples/robots/neon.py` — the end-to-end wire: `Robot("g1", mode="real", driver="strands", port=IP)` + `Mesh(...)` + Ctrl-C.
- `changelog.d/2739-g1-native-driver.md`.

## What deliberately does NOT land

Motion. `send_action`, `start_task` and `run_policy` return the honest "not wired yet (issue #358)" envelope. The FSM (`{500, 501, 801}`) and battery-floor (15%) gates that will guard the motion path are already consulted, so a caller who polls today gets the right refusal reason and the gates already have coverage the day the motion path lands.

The full Livox point cloud. `_lidar_summary` is a small fixed-shape dict; a full frame is ~30k points at 10 Hz and shipping that on `lidar/summary` would drown Zenoh. The 3D tile is a separate dashboard-track change (harness #356).

## Hardware safety

`unitree_sdk2py` is **lazy-imported**: `from strands_robots.drivers.g1 import G1Driver` never touches the SDK, so a headless CI runner and Thor before an office bring-up can build the driver, list it in the registry and run every test with a mocked bus. On a machine without the SDK `connect_eagerly` returns a named error string and leaves the driver in the "usable but not connected" state — a mesh peer for an offline robot is still a valid peer, so `Mesh(...)` publishes an "offline" card instead of raising. The factory constructs but does not connect; the bring-up caller calls `connect_eagerly()`, which is idempotent. Hardware bring-up on the real G1 is validated at the office.

## Tests

```
tests/drivers/test_g1_driver.py:              30 passed
tests/test_all_exports_are_statically_defined.py: 20 passed
tests/test_driver_seam.py:                    45 passed
tests/registry/ + test_robot_factory.py:      930 passed, 1 skipped
```

`ruff check` and `ruff format --check` clean on every new/modified file.

## Follow-ups (in scope of the epic, not this PR)

- #358 (P2) — vendor the neon-the-g1 `g1_*` @tool bundle on top of this driver's DDS engine; wire the motion path so `send_action`/`start_task` do actual work.
- #356 (dashboard track) — paced `lidar/cloud` topic + Three.js point-cloud tile.
- #355 (P2) — Earth Rover Mini+ HTTP driver.
- #357 (P2) — Reachy Mini daemon driver.

### Round 1 (`a253923c`, `08a0c456`)

Two commits, one concern each.

**`a253923c` — stop publishing the private DDS lock (blocking; required check was red).**
`strands_robots/tools/g1/__init__.py` listed `_DDS_INIT_LOCK` in `__all__`. A star-import skips underscore names unless `__all__` overrides that, so listing one publishes it, and `tests/test_all_exports_are_statically_defined.py::TestNoExportIsPrivate` scans every module under `strands_robots/` with a literal `__all__` with no exemption list. It reported `['_DDS_INIT_LOCK']` against `d775f6f2`, deterministically failing `call-test-lint / Test and Lint` — the one required status check, and a grader the suites listed above did not include, which is why the local runs came back green.

Both the `__all__` entry *and* the import are gone, rather than the entry alone: the import existed only to place the name on the package namespace, which is the case the grader's own message says should drop the import too. Nothing loses access — `_dds_engine` already imports the lock from `_g1_common` where it is defined, and the test now reaches it there. The package docstring's claim that the #358 tools share "this module's `_DDS_INIT_LOCK`" was true only while the re-export existed, so it now points at `_g1_common` and says to reach it there.

`test_dds_init_lock_is_a_lock` said "same object" in its docstring but only asserted the lock was acquirable. Moving the import site off the package makes a copy bound in `_dds_engine` invisible from the test's vantage point, so the identity the shared lock depends on is now asserted (`_dds_engine._DDS_INIT_LOCK is _DDS_INIT_LOCK`) instead of described.

Verified against both heads: `private_exports` returns `['_DDS_INIT_LOCK']` on `d775f6f2` and `[]` here; the grader goes 20 passed.

**`08a0c456` — `connect_eagerly` idempotence and its documented caller (non-blocking).**
The docstring said "Called by `Robot` before returning the driver". It is not: `_build_native_driver` only constructs, and the sole caller in the tree is `examples/robots/neon.py`. A reader trusting that line would expect a bring-up against an off robot to fail at `Robot(...)`, when it in fact succeeds and fails later on the first `get_status` poll.

A second `connect_eagerly()` on a connected driver rebuilt `DDSSubscriberSet`, re-subscribed all four topics and reassigned `self._subs`, dropping the only reference to the previous set and leaking its subscribers — on a bus whose bindings segfault under concurrent `ChannelSubscriber(...)`. No caller does this today, which is what makes it a latent trap rather than a live bug: the method reads as connect-and-forget and nothing stopped a retry loop from being written around it. It now returns `None` without touching the existing set, matching `cleanup()`'s idempotence, and the docstring states it.

`test_connect_eagerly_is_a_no_op_on_a_connected_driver` pins it by patching `DDSSubscriberSet` to a constructor that raises. Confirmed the test fails without the guard (`connect_eagerly() rebuilt the subscriber set`) and passes with it, so it is a regression pin rather than a tautology.

**`527842de` — the six mypy errors that were actually failing the gate.**

Worth stating plainly because it changes what the round-1 evidence means: `hatch run lint` runs `mypy strands_robots tests tests_integ` as its third command, and it exited 1 on **6 errors in 4 files, all of them this branch's**. The job died in the lint step at ~2m40s and never reached `Run tests`. So on both `d775f6f2` and `08a0c456` the test suite never ran at all — including the exports grader that `a253923c` fixes. Fixing the grader was necessary and not sufficient; these six are the reason the check was red.

- `unitree_sdk2py.core.channel` import-not-found (`_g1_common.py:120`, `_dds_engine.py:91`) — the SDK ships no stubs and was never declared. `ignore_missing_imports` in a `[[tool.mypy.overrides]]` section keys on the module being *imported*, not the importer, so the existing `strands_robots.tools.*` override could never have covered it. Added `unitree_sdk2py.*` to the unstubbed-third-party list beside `cyclonedds` and `rclpy`; submodule pattern only, since nothing imports the bare package.
- `_snapshot` no-any-return (`g1.py:530`) — `getattr(self, attr, None)` is `Any`, so the trailing `return value` returned `Any` from a `dict[str, Any] | None` signature, and the `# type: ignore[unreachable]` sitting on that line did not cover `no-any-return`. That branch was already judged unreachable and is: each of the four cache attrs is only ever written as a dict or left `None`. The branch is gone and `value` is annotated, so the suppression comment goes with it.
- `ToolUse` typeddict-item (test lines 368, 390) — the tool-use literals omitted the TypedDict's required `name` key. Added `"name": "g1"`, which is what the runtime passes.
- `driver._subs = object()` assignment (test line 475) — the sentinel `08a0c456` introduced was untyped against `DDSSubscriberSet | None`. It is now a real `DDSSubscriberSet`; constructing one touches no DDS (`__init__` only records the interface), so the test is both typed and more faithful. Re-confirmed the idempotence guard is still pinned afterwards — removing the guard fails the test with `connect_eagerly() rebuilt the subscriber set`.

Verified after this commit: mypy clean across `strands_robots/drivers/g1.py`, `strands_robots/tools/g1/` and `tests/drivers/test_g1_driver.py`; `ruff check` and `ruff format --check` clean over `strands_robots tests` (1511 files); `tests/drivers/test_g1_driver.py` 30 passed, exports grader 20 passed.

Not changed: the two `tests/test_driver_seam.py` cases that need the `lerobot` extra cannot run in a venv that also resolves `vera-sim` (the extras declare a conflict), so they were verified as failing identically on unmodified `d775f6f2` — a local env limitation, not a diff effect. CI installs the extra and runs them.


yinsong1986 (commented):
## Summary

This PR puts the first native driver behind the #2734 seam: `G1Driver` speaks raw Unitree IDL over CycloneDDS, subscribing `rt/lowstate`, `rt/lf/bmsstate`, `rt/utlidar/lidar_state` and the Livox cloud on a background DDS thread and dropping each decode into the per-sensor caches (`_imu`, `_battery`, `_lidar_state`, `_lidar_summary`) that `SensorLoopsMixin` already reads via `getattr` defaults - I verified the attribute names against `strands_robots/mesh/sensors.py` on `main` and they line up. The write path is gated but deliberately unwired: `send_action` consults the FSM (`HANDSHAKE_FSMS`) and battery-floor gates and then returns a named "issue #358" refusal envelope, so nothing pretends a write succeeded. `unitree_sdk2py` is lazy-imported everywhere (module import, `ensure_dds`, `_resolve_message_class`), the registry declares `hardware.driver = "strands"` on `unitree_g1` so `driver="auto"` resolves natively, and the `tests/test_driver_seam.py` fixture correctly resets to an empty table now that auto-registration exists. The diff matches the description.

## What's good

- Scope discipline: motion, the raw point cloud and the agent tool bundle are all explicitly deferred with tracked issues, and the stubs refuse honestly instead of no-op'ing as success.
- Clock usage follows the AGENTS.md rule: `time.time()` appears only as record stamps (`"t"` fields), never as a duration base.
- The DDS callbacks swap complete dicts rather than mutating in place, and every decoder survives a malformed message with a debug log rather than tearing down the DDS thread (pinned by `test_decoders_swallow_bad_messages`).
- `ensure_dds` refuses a second init on a different NIC with a named reason instead of silently re-binding; the changelog fragment is present per step 3 of the PR workflow; all new user-facing strings are plain ASCII.

## Verification suggestions

- Run the exports grader, which the PR description's test list does not include: `pytest tests/test_all_exports_are_statically_defined.py` - it fails on this head (see inline comment).
- On the office bring-up, confirm a second `connect_eagerly()` call is never issued against a connected driver: it rebuilds the subscriber set and re-subscribes all four topics without releasing the old one. No current caller does this (the factory only constructs; `examples/robots/neon.py` calls it once), so it is not blocking, but the method is not idempotent and its docstring says the factory calls it, which `_build_native_driver` on `main` does not.

strands_robots/tools/g1/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    decode_code,
+    ensure_dds,
+    reset_dds_state,
+)
+
+__all__ = [
+    "ERR_CODES",
+    "HANDSHAKE_FSMS",
+    "WALK_FSMS",
+    "_DDS_INIT_LOCK",
```
yinsong1986: [MUST FIX] `__all__` publishes the private name `"_DDS_INIT_LOCK"`. This violates AGENTS.md > Review Learnings (PR #85) > API Consistency ("`_`-prefixed names must never appear in `__all__`"), and the rule is test-pinned: `tests/test_all_exports_are_statically_defined.py::TestNoExportIsPrivate` scans **every** module under `strands_robots/` with a literal `__all__` - no exemption list - and its `private_exports` helper reports exactly this entry. I ran that check against this file at head `d775f6f2` and it reports `['_DDS_INIT_LOCK']`.

Why this can't be deferred: that test runs inside `call-test-lint / Test and Lint`, the one context in the ruleset's `required_status_checks`, so the failure is deterministic and the PR cannot merge as-is. The suites listed in the PR description (`tests/drivers/`, `tests/test_driver_seam.py`, registry + factory) don't include this grader, which is why the local runs came back green.

Resolution: drop `"_DDS_INIT_LOCK"` from `__all__`. The lock stays importable for the driver and the #358 tools either way - `tests/drivers/test_g1_driver.py` and `strands_robots/tools/g1/_dds_engine.py` reach it by attribute/explicit import, which `__all__` does not govern. To keep the re-export without tripping ruff F401, use the redundant-alias form (`_DDS_INIT_LOCK as _DDS_INIT_LOCK`) in the import block, or have consumers import it from `strands_robots.tools.g1._g1_common` where it is defined - the same remedy the pinned test's error message prescribes ("export a public accessor instead, or reach the private name in the module that defines it").

cagataycali:
## Independent verification of the review round at `527842de`

I reproduced the blocker, built an alternative fix, and then measured yours instead of pushing over it. Both blockers are cleared and the round is green. Two mutations survive it; details and suggested tests below.

### The blocker was two blockers, and only the second one was reported

`call-test-lint` at `d775f6f2` failed at step 8 `Run lint`, and step 9 `Run tests` was **skipped** - so `TestNoExportIsPrivate` never ran in CI at all. It was found by running the grader by hand, which is why the `__all__` fix alone would have produced another red round: `mypy strands_robots tests tests_integ` reported 5 errors in this PR's own files at `d775f6f2` that `main` does not have:

| file | error |
| --- | --- |
| `strands_robots/tools/g1/_g1_common.py:120` | `import-not-found` for `unitree_sdk2py.core.channel` |
| `strands_robots/tools/g1/_dds_engine.py:91` | same |
| `strands_robots/drivers/g1.py:537` | `no-any-return` from `_snapshot` |
| `tests/drivers/test_g1_driver.py:377,399` | `typeddict-item`, `ToolUse` is missing its required `name` |

Your three commits close all of them. At `527842de`: `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1562 files), and mypy reports **24 == 24** against `upstream/main` at `18e92fe9`, with the normalized message sets byte-identical in both directions and nothing in this PR's files.

### Paired suite run

Identical 459-file selection (all of `tests/drivers/`, the seam tests, and every guard that walks the tree, reads a docstring or reads `pyproject.toml`), every path verified present on both trees:

| tree | collected | result |
| --- | --- | --- |
| `527842de` | 21517 | 20 failed, 21391 passed, 106 skipped (746s) |
| `18e92fe9` (base) | 21504 | 20 failed, 21379 passed, 105 skipped (767s) |

Failing-id sets **identical**, `comm` empty in both directions. All 20 are pre-existing and dependency-caused in this environment: 17 need `awsiot` (the `mesh-iot` extra), 2 are a lerobot-from-source `TypeError: encode() takes 1 positional argument` drift, 1 is a dataset fixture path collision. The `+13` collected localises entirely to guards parametrized over the tree: `test_args_docstring_completeness.py` 444 -> 450, `test_finalizer_reports_only_real_cleanup_failures.py` 275 -> 279, and `test_examples_no_lerobot_imports.py` / `test_examples_release_the_mesh.py` / `test_examples_sim_embodiment.py` +1 each for `examples/robots/neon.py`.

### Two mutations the round survives

Mutations applied to `527842de`, graded against `tests/drivers/` + `tests/test_driver_seam.py` + the exports grader. Source restored byte-identical afterwards.

| mutation | fires |
| --- | --- |
| control | 0 |
| the `if self._connected` no-op guard is dropped | 1 - `test_connect_eagerly_is_a_no_op_on_a_connected_driver` |
| the no-op still returns `None` but drops the held set | 1 - same test |
| **a successful connect subscribes only 3 of the 4 planned topics** (`self._subscription_plan()[:-1]`) | **0** |
| **`cleanup` drops the subscriber set without calling `close()` on it** | **0** |
| the private lock is published from the package again | 1 - `test_no_module_exports_a_private_name` |

Your new test is exactly right about what it grades - it catches both the missing guard and a guard that returns success while losing the set. The two zeros are the other half of the same argument. The Livox cloud is *last* in `_subscription_plan`, so the mutation that silently never subscribes it is the one a truncation bug would produce, and every test still passes. And `close()` in `cleanup` is the only actual subscriber release in the class - the leak your idempotence docstring reasons about - and nothing grades it.

Both close with one recording stand-in. This is the fixture I had written for the alternative fix; on my tree those two mutations fire 1 test each:

```python
class _RecordingSubscriberSet:
    """Stand-in for DDSSubscriberSet that records what happened to it."""

    built: list["_RecordingSubscriberSet"] = []

    def __init__(self, network_interface: str, start_error: str | None = None) -> None:
        self.network_interface = network_interface
        self.start_error = start_error
        self.subscribed: list[str] = []
        self.closes = 0
        _RecordingSubscriberSet.built.append(self)

    def start(self) -> str | None:
        return self.start_error

    def subscribe(self, topic: str, message_class: type, callback: Any) -> str | None:
        del message_class, callback
        self.subscribed.append(topic)
        return None

    def close(self) -> None:
        self.closes += 1


@pytest.fixture
def recorded_dds(monkeypatch: pytest.MonkeyPatch) -> type[_RecordingSubscriberSet]:
    """Swap the subscriber set and the IDL lookup for stand-ins.

    ``_resolve_message_class`` imports from ``unitree_sdk2py``, so it returns a
    reason string on a machine without the SDK and the connect loop would bail
    before subscribing anything. A plain class is enough - the driver only
    passes it through to ``subscribe``.
    """
    _RecordingSubscriberSet.built = []
    monkeypatch.setattr(g1_driver_module, "DDSSubscriberSet", _RecordingSubscriberSet)
    monkeypatch.setattr(g1_driver_module, "_resolve_message_class", lambda cls_path: object)
    return _RecordingSubscriberSet


def test_a_first_connect_subscribes_every_planned_topic(recorded_dds: Any) -> None:
    """One connect builds one set and subscribes the whole plan through it."""
    driver = G1Driver(tool_name="g1", port="1.2.3.4")

    assert driver.connect_eagerly() is None

    assert len(recorded_dds.built) == 1
    planned = [topic for topic, _cls, _decoder in driver._subscription_plan()]
    assert recorded_dds.built[0].subscribed == planned
    assert driver._connected is True


def test_cleanup_releases_the_attached_set(recorded_dds: Any) -> None:
    """The teardown path releases, and leaves the driver disconnected."""
    driver = G1Driver(tool_name="g1", port="1.2.3.4")
    assert driver.connect_eagerly() is None

    driver.cleanup()

    assert recorded_dds.built[0].closes == 1
    assert driver._subs is None
    assert driver._connected is False
```

One note if you take the first one: reading a stand-in's own attribute *after* `assert driver._subs is second` is a mypy `union-attr`, because an `is` assertion narrows both of its sides to the driver's `DDSSubscriberSet | None` annotation. Put the identity assertion last.

### The recovery path works, and is worth one assertion

I built the alternative fix as release-and-rebuild before measuring yours, and your no-op is the safer of the two - it cannot transiently double-subscribe at all, and the segfault argument in the docstring is the right reason. The one thing it depends on is that `self._connected` gets cleared, and it is assigned in exactly two places: `True` in `connect_eagerly`, `False` in `cleanup`. So a driver whose bus went away reports success from `connect_eagerly()` and does nothing, and the only re-attach route is `cleanup()` then `connect_eagerly()`.

That route does work. Measured on `527842de` with the stand-in above:

| step | sets built | releases on set 1 | `_connected` |
| --- | --- | --- | --- |
| `connect_eagerly()` | 1 | 0 | True |
| `connect_eagerly()` again | 1 | 0 | True |
| `cleanup()` | 1 | 1 | False |
| `connect_eagerly()` | 2 | 1 | True |

The re-attached set carries the identical topic list. Adding those last two rows to your own new test would document the sequence, which is currently the only recovery a caller has and is written down nowhere.

### Smaller, both optional

`_snapshot` dropping the `isinstance` check makes the new `value: dict[str, Any] | None` annotation the only thing standing behind `dict(value)`, and `getattr` returns `Any` so nothing verifies it. On `527842de` a non-dict cache entry reaches the agent's `sensors` verb as `ValueError: dictionary update sequence element #0 has length 1; 2 is required`, where `d775f6f2` returned the value. Unreachable today - all four decoders assign dict literals - so this is a note, not a request; `return None` for a non-dict is the other way to keep mypy quiet without an unverifiable annotation.

The mypy override is `"unitree_sdk2py.*"` without the bare `"unitree_sdk2py"`. There is no bare `import unitree_sdk2py` anywhere in the tree today (every site is `from unitree_sdk2py.core.channel import ...`), so it is sufficient; the sibling SDK entries that do have a bare import list both spellings, so the `g1_*` tools in the follow-up may want the pair.

Nothing here blocks. The round clears both merge blockers and the suite is identical to base.

cagataycali:
### Public context for `cagataycali/robots-harness#354` (and the follow-ups #356 / #358)

The harness issues referenced in the PR body live on a private fleet-integration tracker. Public-safe summary for external reviewers:

**Goal (#354).** Native CycloneDDS driver for the Unitree G1 so `Robot("g1", mode="real")` returns a live driver that, wrapped in `Mesh`, publishes lidar summary / IMU / battery / pose. lerobot cannot model a G1 (DDS bus, FSM, Livox MID-360), and `use_dds` targets ROS2 / zenoh-ros2dds — the G1 speaks raw Unitree IDL over CycloneDDS (`rt/lowstate`, `rt/lf/bmsstate`, `rt/utlidar/*`), so a native driver is the only path to the sensor attrs `SensorLoopsMixin` already publishes.

**Safety model.** FSM gate (`{500, 501, 801}` to move) and 15% battery-floor gate consulted by the motion path; these are the gates that keep an arm write from landing on a robot in an unsafe state.

**Scope boundaries in this PR.** Motion path — the actual `send_action` / `start_task` / `run_policy` — is deliberately deferred to the follow-up G1 tool bundle (#358). The full Livox point cloud is deferred to the dashboard track (#356). `_lidar_summary` here is a small fixed-shape dict so `strands/{peer}/lidar/summary` can publish at 10 Hz without drowning Zenoh.

cagataycali:
Second review round, measured at head `527842de`. The registry wiring, the sensor decode shapes and the lazy-SDK claim all hold. Two things do not, and one of them is safety-adjacent.

### What I verified holds

- **Lazy SDK import.** With `unitree_sdk2py` shimmed to raise `ImportError`, `tests/drivers/test_g1_driver.py` is **30 passed**, the driver constructs, and `connect_eagerly()` returns `unitree_sdk2py is not installed: ...`. The claim that an install without the SDK can build, list and test is real, not aspirational.
- **Registry selection.** `unitree_g1.hardware` is `{"lerobot_type": "unitree_g1", "driver": "strands"}`, and it is the **only** robot in `robots.json` declaring `hardware.driver`, so lerobot stays the global default. `get_native_driver_class` resolves both `"g1"` and `"unitree_g1"` to `G1Driver`.
- **The decode shapes line up with the publisher.** `_read_health` reads `battery.get("pct")` and `_on_bms` writes `pct`; `_read_imu`, `_read_lidar_summary` and `_read_lidar_state` all `update()` from a plain dict, which is what the three decoders produce. Nothing is dropped at the seam.
- **Caches starting at `None` are not a trap.** `_build_presence` is called only from `_heartbeat_loop`, so a driver whose first DDS message has not landed yet re-advertises on the next beat rather than being born with a permanently empty capability list.

### 1. The gate coverage does not extend past `send_action`

The PR description says the FSM and battery gates are "consulted by every motion-path stub". Walking the AST of every public method for the four gate tokens:

| method | gates consulted |
|---|---|
| `send_action` | connected, fsm-known, fsm-set, battery-floor |
| `start_task` | none |
| `run_policy` | none |
| `get_task_status` / `stop_task` | none (read-only, correctly) |

Driving all three with a driver that is connected, at `_fsm_id = None`, battery 4.0% against a 15.0% floor:

```
send_action  -> error  FSM id unknown - lowstate has not delivered yet
start_task   -> error  start_task not wired yet (issue #358)
run_policy   -> error  run_policy not wired yet (issue #358)
```

`start_task` and `run_policy` are motion entry points, and they answer with the same text whatever the FSM and the battery say. `send_action`'s own docstring states the rationale for installing gates ahead of the writes -- "installing them here means the day the write lines land the gates already have coverage" -- and that rationale applies equally to the two verbs that did not get them.

### 2. `WALK_FSMS` is dead, and the gate that is installed uses the wider set while naming loco

`_g1_common.py` defines two sets and documents the difference:

```python
HANDSHAKE_FSMS = frozenset({500, 501, 801})
WALK_FSMS      = frozenset({501, 801})   # "sitting (500) accepts arm gestures but not walking"
```

`WALK_FSMS` is defined, re-exported and listed in `__all__`, and **read by nothing** -- not the driver, not a test. Meanwhile the one gate that exists refuses with `f"FSM {id} refuses arm/loco writes"`, naming locomotion, while testing membership of the arm set. So a locomotion write at the FSM the sibling constant exists to exclude passes the gate:

```
FSM   0  walk-safe=False  loco write -> refused          FSM 0 refuses arm/loco writes; needs one of [500, 501, 801]
FSM   1  walk-safe=False  loco write -> refused          FSM 1 refuses arm/loco writes; needs one of [500, 501, 801]
FSM 500  walk-safe=False  loco write -> PASSES the gate  motion path not wired yet (issue #358)
FSM 501  walk-safe=True   loco write -> PASSES the gate  motion path not wired yet (issue #358)
FSM 801  walk-safe=True   loco write -> PASSES the gate  motion path not wired yet (issue #358)
```

Nothing moves today, so this is latent. It stops being latent the moment the write lines land, and the constant recording the correct answer is already in the tree with no consumer.

### 3. The suite survives every mutation of the set choice, and pins the gate-free stubs

Six mutations against `tests/drivers/test_g1_driver.py` + `tests/test_driver_seam.py` (75 items), each reverted byte-identical:

| mutation | tests firing |
|---|---|
| control, unmodified | 0 (75 passed) |
| `HANDSHAKE_FSMS` narrowed to `{501, 801}` (drop 500) | **0** |
| `HANDSHAKE_FSMS` widened to admit 0, 1, 3, 4 | 4 |
| `WALK_FSMS` made identical to `HANDSHAKE_FSMS` | **0** |
| `WALK_FSMS` emptied entirely | **0** |
| gates added to `start_task` | 1 |

The 500 membership can be dropped, and `WALK_FSMS` can be emptied or collapsed onto its sibling, with nothing noticing. `test_send_action_refuses_outside_handshake_fsm` is parametrized over `[0, 1, 3, 4]` and then asserts `str(f) in text for f in HANDSHAKE_FSMS` -- derived from the constant under test, so it follows any edit to it. The refusal side is pinned; the acceptance side is not.

The last row is the one worth pausing on: adding the gates to `start_task` **fails `test_task_and_policy_paths_report_not_wired`**, because that test asserts `"issue #358" in start["content"][0]["text"]` on an unconnected driver. The suite does not merely leave the gap uncovered, it currently encodes "no gates on `start_task`" as the contract. So this needs a deliberate call rather than a drive-by patch.

### The decision this needs

I am not proposing a specific set for `start_task`/`run_policy`, because the right answer depends on something this PR deliberately defers: whether a given call is an arm write or a locomotion write. `send_action` takes an opaque `action` dict and `start_task` takes free text, so neither can classify itself today. Three coherent options:

1. Gate both new verbs on `HANDSHAKE_FSMS` now and split arm vs loco when the writes land. Cheapest, keeps one rule, needs `test_task_and_policy_paths_report_not_wired` updated to expect a gate reason on an unconnected driver.
2. Leave the stubs ungated and delete `WALK_FSMS` until something reads it, so the tree does not carry an unenforced safety distinction. Then narrow `send_action`'s message to "arm writes" so it stops claiming to gate loco.
3. Keep `WALK_FSMS`, give it a consumer, and pin the 500 boundary with a test that does not derive its expectation from the constant.

Any of the three is defensible; shipping none of them leaves a documented distinction with no enforcement and a refusal message that overstates its reach. Whichever way it goes, a case at FSM 500 asserting a literal expectation would close the ungraded acceptance side.

### Minor: acceptance criterion for the wire test

The on-the-wire check (a `Mesh` wrapping the driver fed mocked DDS, asserting `lidar/summary` + `imu` + `health` arrive) is not in the suite. I checked the shapes by hand and they line up, so I read this as a missing test rather than a defect -- but it is the one criterion that would have caught a seam mismatch, and it is the criterion the dashboard goal rests on.

cagataycali:
## Verification of the lint-gate fix, plus the missing pin

I reproduced the blocker locally and can confirm the resolution in `527842de` clears it. Two notes on what was actually red, then one thing I pushed on top as `6e67292f`.

### What `call-test-lint` was failing on

The inline thread is correct that `tests/test_all_exports_are_statically_defined.py::TestNoExportIsPrivate` fails at `d775f6f2` (`{'tools/g1/__init__.py': ['_DDS_INIT_LOCK']}`), and correct that it is merge-blocking. Worth recording that it is not what the run reported, because the step list matters for anyone reading the failure later:

| step | outcome |
|---|---|
| Run lint | **failure** |
| Run tests | **skipped** |

`Run lint` failed first, so `Run tests` never executed and the exports grader never ran in CI at all. Both defects were real and independently merge-blocking; fixing only the `__all__` entry would have left the job red on mypy, and fixing only mypy would have left it red on the grader once tests began running. The five in-scope mypy errors at `d775f6f2` were the two `unitree_sdk2py` `import-not-found`, `drivers/g1.py` `no-any-return` (the `# type: ignore[unreachable]` there names a code mypy does not report at that line, so it suppressed nothing), and two `Missing key "name" for TypedDict "ToolUse"`.

### The fix checks out

`mypy strands_robots tests tests_integ` on `527842de` versus a pristine worktree of the branch's base:

| tree | errors | in `strands_robots/`, `tests/`, `tests_integ/` |
|---|---|---|
| base `87d35b2d` | 19 | 0 |
| head `527842de` | 19 | 0 |

Normalised message sets are identical in both directions, so the head cannot fail `Run lint` for any reason the base would not. All 19 are in `examples/isaac_gs/`, outside the checked paths, and present on `main`.

Since `Run tests` has never executed on this branch, I also ran the full suite on `527842de` plus my commit: **39287 passed, 123 skipped**, and 9 failures none of which are attributable here. Five are `tests/test_{dds_domain_id,ros_qos_history_depth,wait_budget}_domain.py` and `test_dependency_audit.py` asserting `DID NOT RAISE ImportError`, which need `rclpy` absent while it resolves locally; two are `tests/policies/vera/test_vera_n_action_steps_removed.py` needing `docker` on PATH; one is `newton/test_multi_camera.py::test_bad_position_shape_rejected`, which asserts `'3 elements'` against `"must be a 3-element vector"` and fails identically on the base and on current `main`; and one is `test_rendering_pacers_survive_a_clock_step.py[forward_1h]`, whose premise counts wall-clock frames and which passes 16/16 in isolation on the same tree at load average 17.5. Each of the first eight reproduces on a pristine base checkout with byte-identical messages.

I also checked the sim path, since `robots.json` gained `hardware.driver` for `unitree_g1`: `Robot("unitree_g1", mode="sim")` still builds a `MuJoCoSimEngine` with 30 joints and 62 observation keys and steps normally, `get_native_driver_class` resolves both `g1` and `unitree_g1` to `G1Driver`, and `unitree_sdk2py` is absent from `sys.modules` after all of it.

### What I pushed: the override is load-bearing and was ungraded

`6e67292f` extends `tests/test_mypy_untyped_third_party_imports.py` to cover `unitree_sdk2py`. That file exists for precisely this and says so in its docstring: "These tests pin the override so a future dependency bump (or an accidental removal of the override) is caught here instead of in CI." It covered `pyarrow` and nothing else, so on `527842de` deleting the new override entry is caught by nothing local:

| mutation on `527842de` | tests firing | mypy on the two importers |
|---|---|---|
| none | 0 (99 passed) | 0 errors |
| override entry deleted | **0** | **2 errors** |

That is the same shape as the failure this branch already spent a CI round-trip on: invisible locally, red in `Run lint`. With the pin, the same deletion fires both new tests.

The two tests are complementary rather than duplicative, which the third row shows. The declaration test matches by prefix instead of demanding an exact pair of entries, because I measured that `"unitree_sdk2py.*"` alone already silences a bare `import unitree_sdk2py` (0 errors with and without a separate top-level entry) - so your single entry is the correct minimal spelling and requiring both would have failed a correct override. Narrowing it the other way is what is genuinely insufficient:

| mutation on `527842de` | tests firing |
|---|---|
| override entry deleted | 2 (declaration + behavioural) |
| `"unitree_sdk2py.*"` narrowed to `"unitree_sdk2py"` | 1 (behavioural only) |

`ruff check` and `ruff format --check` over `strands_robots tests tests_integ` are clean, and mypy is unchanged from the table above.

### One optional nit

The changelog fragment does not mention that `unitree_sdk2py` joined the untyped-imports override. It is a real consequence for anyone vendoring the SDK, and the assembled changelog is where a maintainer would look for it. I left the fragment alone rather than edit a file this branch owns; a sentence there would close it.

The idempotence fix in `08a0c456` and its test address the second point in the review directly, and `test_every_module_in_the_dds_layer_takes_the_same_init_lock` is a good catch: I verified independently that with `_dds_engine` binding its own `threading.Lock()`, the original `hasattr`-only assertions passed and the identity assertion is the only thing that fails.

cagataycali:
## Third round: safety-gate hardening, `4818d566`

Addresses the two ungraded findings from the previous review round in-branch rather than deferring:

### Finding 1: `start_task` / `run_policy` skipped every gate

The reviewer's AST walk was right — the PR description claimed 'gates on every motion-path stub' but only `send_action` ran them, and the suite pinned that gap by asserting `'issue #358'` on an unconnected driver. Fixed: every motion verb now routes through one `_check_motion_gates(scope)` helper, and the `test_task_and_policy_paths_report_not_wired` test was renamed to `_when_gates_pass` and given a connected + charged + walkable driver — the suite now encodes the *right* contract for the stubs.

### Finding 2: gate used the wider set while naming loco

`send_action` refused with `'FSM {id} refuses arm/loco writes'` while testing `HANDSHAKE_FSMS`. Fixed: `send_action` is arm-scoped and its message now names `'arm writes'`; `start_task`/`run_policy` are motion-scoped (they may issue either kind and the caller does not classify itself, so the union `HANDSHAKE_FSMS | WALK_FSMS` matches their reach). `WALK_FSMS` gets a consumer today via `_check_motion_gates('loco')` — no verb classifies itself as loco yet, but the helper is callable and the boundary at FSM 500 (sitting: arm ✓, walk ✗) is pinned by `test_walk_fsms_has_a_consumer_and_a_documented_boundary`.

### The reviewer's break table on this branch

| mutation | fires now |
|---|---|
| `HANDSHAKE_FSMS` narrowed (drop 500) | `test_send_action_admits_every_handshake_fsm[500]` |
| `HANDSHAKE_FSMS` widened to admit 0, 1, 3, 4 | `test_send_action_refuses_outside_handshake_fsm[*]` (existing) |
| `WALK_FSMS` made identical to `HANDSHAKE_FSMS` | `test_walk_fsms_has_a_consumer_and_a_documented_boundary` |
| `WALK_FSMS` emptied entirely | same |
| gates added to `start_task` (control removed) | `test_task_paths_refuse_when_not_connected[start_task]` + siblings |
| refusal message narrowed to non-'arm writes' | `test_send_action_refuses_outside_handshake_fsm` (checks `'arm writes' in text` and `'loco' not in text`) |

### Green headless

- `pytest tests/drivers/ tests/test_driver_seam.py --no-cov -q`: **85 passed** (was 75)
- `pytest tests/test_all_exports_are_statically_defined.py`: **20 passed**
- `mypy strands_robots/drivers/g1.py tests/drivers/test_g1_driver.py`: **0 in-scope errors**
- `ruff check` + `ruff format --check` on the two files: **clean**
- Changelog fragment updated to name both FSM sets and the scope choice

### What did not change

The motion path still lands nothing today — the refusal after every gate passes is still `'not wired yet (issue #358)'`, so a caller polling any verb sees the same honest reason. Every verb consults gates that will still be right the day the writes land, and the acceptance side is now pinned rather than derived-from-the-constant-under-test.

---

Hardware bring-up on the real G1 still validated at the office; Thor has SO-101s, not a G1. The three coherent options the reviewer laid out — this branch took option 3 (keep `WALK_FSMS`, give it a consumer, pin the 500 boundary with a test that does not derive its expectation from the constant).

cagataycali:
## Third review round, measured at head `4818d566`

Finding 1 is closed: every motion verb now routes through `_check_motion_gates`, and `send_action`'s message no longer claims to gate locomotion. Confirmed by driving the verbs directly.

Finding 2 is not closed. `WALK_FSMS` has a syntactic consumer but still no consumer whose *outcome* depends on it, and the test written to prove otherwise passes when the constant is emptied. Details below, all measured on this head; the two source files were restored byte-identical afterwards (md5 verified).

### The `motion` scope is arithmetically the `arm` scope

`WALK_FSMS` is a proper subset of `HANDSHAKE_FSMS`, so the union in the `motion` branch is a no-op:

```
HANDSHAKE_FSMS            = [500, 501, 801]
WALK_FSMS                 = [501, 801]
WALK_FSMS <= HANDSHAKE     = True
union  (the motion scope) = [500, 501, 801]
union == HANDSHAKE_FSMS    = True
intersection               = [501, 801]
intersection == WALK_FSMS  = True
```

Asking the helper itself which FSMs each scope admits, on a connected driver at 92% battery:

| scope | admits | distinct from `arm`? |
|---|---|---|
| `arm` | `[500, 501, 801]` | - |
| `loco` | `[501, 801]` | yes |
| `motion` | `[500, 501, 801]` | **no, identical** |

Three scopes, two admitted sets. The `motion` and `arm` branches differ only in the refusal wording. Driving the public verbs confirms it:

| FSM | in `WALK_FSMS` | `send_action` | `start_task` | `run_policy` |
|---|---|---|---|---|
| 0 | False | `FSM 0 refuses arm writes` | `FSM 0 refuses motion writes` | `FSM 0 refuses motion writes` |
| 500 | **False** | not wired (issue #358) | **not wired (issue #358)** | **not wired (issue #358)** |
| 501 | True | not wired (issue #358) | not wired (issue #358) | not wired (issue #358) |
| 801 | True | not wired (issue #358) | not wired (issue #358) | not wired (issue #358) |

At FSM 500 the loco gate is right and unreached:

```
start_task("walk forward")       -> start_task not wired yet (issue #358)
_check_motion_gates("loco")      -> FSM 500 refuses locomotion writes; needs one of [501, 801]
```

I am not asking for the intersection. Refusing `start_task` at 500 would refuse an arm-only task while sitting, which the G1 honours, so the permissive pre-flight is a defensible choice and the docstring's reason for it (the narrower per-step gate lands with the writes) is coherent. The point is narrower: **as composed, no verb's behaviour depends on `WALK_FSMS`**, so the safety of the motion verbs rests entirely on a per-step gate that does not exist yet, and nothing in the tree records that as an obligation.

### The break-table row for an emptied `WALK_FSMS` does not hold

The PR body lists `WALK_FSMS` emptied entirely as firing `test_walk_fsms_has_a_consumer_and_a_documented_boundary`. Measured against `tests/drivers/` + `tests/test_driver_seam.py` (85 items), it fires nothing:

| mutation | fires | `motion` admits | result |
|---|---|---|---|
| control | 0 | `[500, 501, 801]` | 85 passed |
| `WALK_FSMS` emptied entirely | **0** | `[500, 501, 801]` | 85 passed |
| `WALK_FSMS` made identical to `HANDSHAKE_FSMS` | 1 | `[500, 501, 801]` | 1 failed |
| `motion` scope changed to the intersection | 2 | `[501, 801]` | 2 failed |
| `motion` scope changed to `WALK_FSMS` | 2 | `[501, 801]` | 2 failed |
| `HANDSHAKE_FSMS` narrowed (drop 500) | 1 | `[501, 801]` | 1 failed |

Cause: the only assertion in that test about the set's *contents* is

```python
for member in WALK_FSMS:
    assert str(member) in text
```

which is vacuous over an empty set. Its two remaining assertions survive an empty set as well: `"FSM 500" in text` still holds, and `"500" not in admitted_set_part` holds all the more when the message reads `needs one of []`. Every assertion in the test grades the refusal direction; nothing asserts that a walkable FSM is *admitted*. So `WALK_FSMS = frozenset()` -- which would refuse every locomotion write the day the writes land -- passes the test written to prove the constant is not dead.

Row 3 does fire, which is worth saying: the boundary test does catch a widening. It is the narrowing that is unguarded.

### The new motion test derives its expectation from the expression under test

Rows 4 and 5 both fire `test_task_paths_refuse_outside_motion_fsm[start_task]` and `[run_policy]`, because that test ends:

```python
for member in HANDSHAKE_FSMS | WALK_FSMS:
    assert str(member) in text
```

This is the same shape flagged last round for `test_send_action_refuses_outside_handshake_fsm` -- an expectation derived from the constant under test, so it follows any edit rather than grading one. Here it is derived from the *composition* under test, which makes it stronger than that: it pins the union as the contract, so any future tightening of the `motion` scope fails a test named for refusing outside it. `test_send_action_admits_every_handshake_fsm` fixed this correctly for the arm scope by parametrizing over literal values; the motion scope did not get the same treatment.

### Suggested tests, and what they measure

Three tests close both gaps. On the unmodified head they are 7 passed, `ruff check` and `ruff format --check` clean, and mypy reports nothing attributable to the file.

```python
def _charged(fsm: int) -> G1Driver:
    driver = G1Driver(tool_name="g1", port="1.2.3.4")
    driver._connected = True
    driver._fsm_id = fsm
    driver._battery = {"pct": 92.0, "charging": True, "current": 1.0, "cycle": 0, "t": 0.0}
    return driver


def test_walk_fsms_is_a_proper_subset_of_handshake_fsms() -> None:
    """Both sets are populated and the distinction is real.

    ``test_walk_fsms_has_a_consumer_and_a_documented_boundary`` asserts the
    set's contents with ``for member in WALK_FSMS``, which is vacuous when
    the set is empty. This states the shape directly instead.
    """
    assert WALK_FSMS, "WALK_FSMS must not be empty"
    assert WALK_FSMS < HANDSHAKE_FSMS, "WALK_FSMS must be strictly narrower"
    assert 500 in HANDSHAKE_FSMS - WALK_FSMS


@pytest.mark.parametrize("fsm", [501, 801])
def test_the_loco_gate_admits_every_walking_fsm(fsm: int) -> None:
    """Acceptance side of the loco gate, at literal values.

    The boundary test only grades the refusal at 500. Nothing asserts a
    walkable FSM is admitted, so emptying WALK_FSMS fires no test.
    """
    assert _charged(fsm)._check_motion_gates("loco") is None


@pytest.mark.parametrize("verb", ["start_task", "run_policy"])
@pytest.mark.parametrize("fsm", [501, 801])
def test_motion_verbs_admit_a_literally_walkable_fsm(verb: str, fsm: int) -> None:
    """A motion verb at an FSM both scopes accept passes the gate.

    Literal values rather than a derivation from the composition, so this
    survives a tightening of the motion scope.
    """
    driver = _charged(fsm)
    result = driver.start_task("do X") if verb == "start_task" else driver.run_policy(None)  # type: ignore[arg-type]
    text = result["content"][0]["text"]
    assert "issue #358" in text
    assert f"FSM {fsm} refuses" not in text
```

Graded against the identical six mutations, they are complementary to the shipped suite rather than a duplicate of it:

| mutation | shipped suite fires | these fire |
|---|---|---|
| control | 0 | 0 |
| `WALK_FSMS` emptied entirely | 0 | **3** |
| `WALK_FSMS` made identical to `HANDSHAKE_FSMS` | 1 | 1 |
| `motion` scope changed to the intersection | 2 | **0** |
| `motion` scope changed to `WALK_FSMS` | 2 | **0** |
| `HANDSHAKE_FSMS` narrowed (drop 500) | 1 | 1 |

The two zeros in the right column are deliberate: these tests use literal FSM values, so a maintainer who later decides the `motion` pre-flight should be the narrower set is not fighting a test named for something else. On that mutation the derived assertion in `test_task_paths_refuse_outside_motion_fsm` is the only thing that fires, which is the clearest statement of what it currently pins.

If the permissive pre-flight is kept, the third test plus a sentence in `start_task`'s docstring naming the per-step gate as the thing that makes it safe would leave the reasoning recoverable. That obligation is currently recorded only in prose that says the narrower classification "lands with the write lines".

### Not raised as a defect

`get_task_status` and `stop_task` remain ungated, correctly -- both are read-only and neither can issue a write.

Nothing here is merge-blocking. The gate coverage gap from the previous round is genuinely closed, and the suite is 85 passed on this head.

cagataycali:
### Correction to my previous comment, and a second break-table row that does not hold

I got one thing wrong above and it turns out to matter. I wrote that `test_send_action_admits_every_handshake_fsm` "fixed this correctly for the arm scope by parametrizing over literal values". It does not. Its decorator is

```python
@pytest.mark.parametrize("fsm", sorted(HANDSHAKE_FSMS))
```

which is derived from the constant under test, exactly like the two cases I criticised. Apologies for the mischaracterisation; the consequence is worse than the one I reported.

### Narrowing the set deselects the case that would catch it

The PR body lists `HANDSHAKE_FSMS` narrowed (drop 500) as firing `test_send_action_admits_every_handshake_fsm[500]`. That test id **ceases to exist** under the mutation, because the parametrize list shrinks with the set. Measured against `tests/drivers/` + `tests/test_driver_seam.py`:

| tree | collected | failures |
|---|---|---|
| unmodified | 85 | 0 |
| `HANDSHAKE_FSMS` narrowed to `{501, 801}`, as shipped | **84** | 1, and it is `test_walk_fsms_has_a_consumer_and_a_documented_boundary` |

The collected count is the tell: 85 to 84. Collected ids for that test go from

```
test_send_action_admits_every_handshake_fsm[500]
test_send_action_admits_every_handshake_fsm[501]
test_send_action_admits_every_handshake_fsm[801]
```

to just `[501]` and `[801]`. A test whose parametrize list is derived from the set it guards cannot detect a narrowing of that set: the removal removes the case, and a suite that runs one fewer test still reads green.

So the narrowing *is* caught, but incidentally and by a differently-named test: `test_walk_fsms_has_a_consumer_and_a_documented_boundary` hardcodes 500 and calls `send_action` there, so it fails when 500 stops being admitted. The acceptance side of `HANDSHAKE_FSMS` is protected only as a side effect of a test written about `WALK_FSMS`. Rename or re-scope that test later and the protection disappears silently.

### The one-line fix, measured

```python
@pytest.mark.parametrize("fsm", [500, 501, 801])
```

| tree | collected | failures |
|---|---|---|
| literal parametrize, unmodified | 85 | 0 |
| literal parametrize + narrowing | **85** | **2**: `test_send_action_admits_every_handshake_fsm[500]` and the boundary test |

Collected stays at 85 under the mutation, the `[500]` case survives to fail, and it passes on unmodified source so it is not a tautology. Source files restored byte-identical after each row (md5 verified).

### Revised tally

Two of the six rows in the published break table do not hold as written:

| published row | claimed | measured |
|---|---|---|
| `WALK_FSMS` emptied entirely | fires the boundary test | **0 fires** (vacuous `for` loop) |
| `HANDSHAKE_FSMS` narrowed (drop 500) | fires `..._admits_every_handshake_fsm[500]` | fires the **boundary test**; the named id is deselected |

Both have the same root cause, which is worth stating once rather than as two separate nits: **three of the four assertions that grade these two sets read the sets themselves**, either as a `for` loop over the set or as a parametrize list built from it. A test that derives its expectation from its subject follows an edit instead of grading it, and in the parametrize case it does so by quietly running fewer tests.

The remedy is the same in all three places and is what `test_task_paths_refuse_outside_motion_fsm` also needs: state the expected values literally, so the set and the expectation are two independent statements that can disagree.

Nothing here is new severity beyond what I reported above, and none of it is merge-blocking: the gate coverage gap from round 2 is closed and the suite is 85 passed on this head. But the break table should not be read as evidence that the set choices are pinned, because for two of the four set-level edits it is not.

cagataycali:
## Round 4 shipped, `7ad3a027`

Addresses both defects in the round-3 review — both had the same root cause and are fixed in the test file rather than the driver.

### The three set-derived grades, restated literally

| location | before | after |
|---|---|---|
| `test_task_paths_refuse_outside_motion_fsm` | `for member in HANDSHAKE_FSMS \| WALK_FSMS` | `for member in (500, 501, 801)` |
| `test_walk_fsms_has_a_consumer_and_a_documented_boundary` | `for member in WALK_FSMS` | `for member in (501, 801)` |
| `test_send_action_admits_every_handshake_fsm` | `@parametrize("fsm", sorted(HANDSHAKE_FSMS))` | `@parametrize("fsm", [500, 501, 801])` |

Every one of these previously followed a set edit through instead of grading it — either by iterating over an empty set, or by shrinking a parametrize list with the set.

### Three complementary tests for the acceptance side

The shipped suite pinned refusals but not admittance. Empty-set and drop-500 mutations were caught only incidentally by tests named for something else. The new tests grade at literal FSM values:

```python
def test_walk_fsms_is_a_proper_subset_of_handshake_fsms() -> None:
    assert WALK_FSMS, "WALK_FSMS must not be empty"
    assert WALK_FSMS < HANDSHAKE_FSMS, "WALK_FSMS must be strictly narrower"
    assert 500 in HANDSHAKE_FSMS - WALK_FSMS

@pytest.mark.parametrize("fsm", [501, 801])
def test_the_loco_gate_admits_every_walking_fsm(fsm: int) -> None:
    ...  # _check_motion_gates("loco") returns None

@pytest.mark.parametrize("verb", ["start_task", "run_policy"])
@pytest.mark.parametrize("fsm", [501, 801])
def test_motion_verbs_admit_a_literally_walkable_fsm(verb, fsm) -> None:
    ...  # refusal is "not wired" (#358), FSM number not mentioned as a refusal
```

### Measured on `7ad3a027`

Suite counts stay constant under mutation — narrowings now show up as failed cases rather than deselected cases. Source restored byte-identical after each row (md5 verified).

| tree | collected | failures |
|---|---|---|
| unmodified | **92 passed** (was 85; +7) | 0 |
| `HANDSHAKE_FSMS = {501, 801}` (narrowed, drop 500) | 92 | **5**, incl. named `test_send_action_admits_every_handshake_fsm[500]` |
| `WALK_FSMS = frozenset()` (emptied) | 92 | **4**, incl. both `test_the_loco_gate_admits_every_walking_fsm[*]` |

- `ruff check` + `ruff format --check` on the changed file: **clean**
- No change to `strands_robots/drivers/g1.py`; fix lives entirely in `tests/drivers/test_g1_driver.py`.

### What is unchanged

The motion path still lands nothing today (`start_task` / `run_policy` / `send_action` all return `"not wired yet (issue #358)"`), the gate helper and its callers are identical, and the driver's public surface is the same. This round is purely test hardening for the two grading gaps flagged in review.

Waiting on `call-test-lint` (maintainer-approval gated) — the other 10 checks were green on the previous head and should stay green on this test-only delta.

yinsong1986 (approved):
## Summary

First native driver behind the #2734 seam: `G1Driver` speaks raw Unitree IDL over CycloneDDS, subscribing `rt/lowstate`, `rt/lf/bmsstate`, `rt/utlidar/lidar_state` and the Livox cloud on the DDS thread and dropping each decode into the per-sensor caches (`_imu`, `_battery`, `_lidar_state`, `_lidar_summary`) that `SensorLoopsMixin` reads via `getattr` -- I verified the key names against `strands_robots/mesh/sensors.py` (`pct`/`charging` for the health chip) and they match. The registry declares `hardware.driver = "strands"` on `unitree_g1`, so `driver="auto"` resolves to the native path, and a failed auto-registration surfaces as a named `ValueError` from the factory rather than a silent lerobot fallback. Motion deliberately does not land: `send_action`/`start_task`/`run_policy` consult the live FSM and battery-floor gates first, then return the named "not wired yet (issue #358)" envelope, and the suite pins gate coverage on every stub plus the `HANDSHAKE_FSMS`/`WALK_FSMS` boundary. `unitree_sdk2py` is lazy-imported everywhere (module import never touches it), the `ensure_dds` singleton and `_DDS_INIT_LOCK` are shared with the future #358 tools with an identity assertion pinning the shared-lock fact, and the round-1 concern (`_DDS_INIT_LOCK` in `__all__`) is resolved on this head -- both the entry and the import are gone, and the docstring now points at `_g1_common`. The final commit (`7ad3a027`) is test-only: it restates the FSM set expectations at literal values so a narrowed `HANDSHAKE_FSMS` fires a named failure instead of silently deselecting a parametrize case, and adds the acceptance-side pins for the `loco` gate.

## What's good

- Every refusal path returns the structured error envelope rather than raising, per AGENTS.md (PR #86) error-handling contracts; `_refuse` keeps the shape in one place.
- `connect_eagerly` idempotence is a real regression pin (patched constructor raises), not a tautology, and `cleanup()` matches it.
- `ensure_dds` refuses a re-init on a different NIC with a named reason instead of silently re-binding -- the right side of the no-silent-defaults rule.
- The mypy override addition ships with its own pin tests in `tests/test_mypy_untyped_third_party_imports.py`, matching the existing pyarrow pattern.
- `tests/test_driver_seam.py`'s isolation fixture reset to an empty table is the correct consequence of shipping auto-registered drivers.
- Changelog fragment present; example follows the established `MUJOCO_GL`/`STRANDS_MESH_LOCAL_DEV` setdefault pattern of the other mesh examples.

## Verification suggestions

- Hardware bring-up on the office G1 is the one thing CI cannot cover: confirm `connect_eagerly()` populates all four caches and that `get_status` reports the live `fsm_id` before merging the #358 motion path on top.
- `python -c "from strands_robots.drivers import list_native_drivers; print(list_native_drivers())"` on a bare install (no SDK) should print `{'unitree_g1': 'G1Driver'}` -- the lazy-import claim in one line.

cagataycali:
Measured this branch at `7ad3a027` against the mesh, and filed the result as #2749 because it looks like a seam question rather than anything wrong with this PR.

`G1Driver` is contract-complete, and its joint positions cannot reach the state topic:

```
missing_driver_members(G1Driver) : ()
has .robot                       : False
has get_observation              : False

_read_state()                    : None          <- no state topic at all
joints section                   : <ABSENT>
imu via the _imu cache           : {'peer_id': 'g1__probe', 't': ..., 'rpy': [0.01, -0.02, 0.5]}
health temps via the _temps cache: {'shoulder_pan': 41.5, 'gripper': 38.0}
```

The sensor caches this driver fills (`_imu`, `_battery`, `_lidar_state`, `_lidar_summary`) all publish correctly -- that half of the wiring works. The gap is that joints have no driver-level route: `mesh/core.py:1396` reaches them only through `getattr(r, "robot", None)` and then `read_joints(inner)` at `:1405`, and this driver has no inner device. The mesh reads sixteen attributes directly off a driver (`_pose`, `_imu`, `_battery`, `_temps`, `_odom`, `_lidar_summary`, `_hands`, `_map_info`, ...) and none of them is joints.

Two smaller consequences of `inner is None` on the same object: presence drops `connected` and `hw` (`core.py:1058`), and `_publish_cameras_once` routes this real peer into `_publish_sim_cameras()` (`core.py:1571-1575`).

`_read_state()` returning `None` is the presentation that method's own docstring calls out as undiagnosable -- the peer goes quiet on the state topic while its presence heartbeat keeps advertising it, and I measured both driver shapes advertising an identical `topics=['health']` / `robot_id`.

Not asking for a change here. #2749 lays out the three routes (mesh falls back to the driver / the contract requires an inner device / joints get a `_joints` cache like the other sixteen) and notes that the Protocol route is ruled out, because `hasattr(hardware_robot.Robot, 'get_observation')` is `False` so adding it would make the incumbent driver fail its own surface check. Whichever is chosen, a `_joints` cache here would be a two-line follow-up.

Reproduction script is in #2749.

yinsong1986 (approved):
## Summary

First native driver behind the #2734 seam: `G1Driver` speaks raw Unitree IDL over CycloneDDS, subscribing `rt/lowstate`, `rt/lf/bmsstate`, `rt/utlidar/lidar_state` and the Livox cloud on the DDS thread and filling the `_imu` / `_battery` / `_lidar_state` / `_lidar_summary` caches that `SensorLoopsMixin` reads via `getattr` defaults - I verified those attribute names against `strands_robots/mesh/sensors.py` and the `battery.pct`/`charging` keys the health chip expects. Auto-registration in `strands_robots/drivers/__init__.py` plus `hardware.driver = "strands"` on `unitree_g1` makes `Robot("g1", mode="real")` resolve to the native path via `driver="auto"`, with `driver="lerobot"` still an explicit escape hatch. Motion deliberately does not land: `send_action`/`start_task`/`run_policy` consult the FSM and battery gates first (arm scope vs. the motion union, per the G1's own documentation of the two sets) and then return the named "not wired yet (issue #358)" error envelope, so a caller today gets the honest refusal reason rather than a stub that reads as a successful write. The concerns from the earlier review round are resolved at this head: `strands_robots/tools/g1/__all__` publishes no private name (the `_DDS_INIT_LOCK` re-export and its import are both gone, and the shared-lock identity is now asserted via `_dds_engine._DDS_INIT_LOCK is _DDS_INIT_LOCK`), `connect_eagerly` is a guarded no-op on a connected driver with a regression pin that patches `DDSSubscriberSet` to a raising constructor, and the `unitree_sdk2py.*` mypy override that was failing the lint gate is in place and itself pinned by two new tests. The sensor caches are written by whole-dict reference swaps (never mutated after publication) and read by snapshot, so the unlocked callback writes are safe under CPython semantics; `ensure_dds` serialises factory init and subscriber construction under one shared lock, matching the segfault constraint the docstrings name. No must-fix concerns found.

## What's good

- The lazy-import discipline is consistent: no module-level `unitree_sdk2py` touch anywhere, every SDK failure comes back as a named error string rather than a raise, and the "usable but not connected" state is what lets a mesh peer publish an offline card - all matching AGENTS.md's error-contract and optional-dep conventions.
- The FSM-set tests state their expectations at literal values (`{500, 501, 801}`, `{501, 801}`) rather than deriving them from the constants under test, so narrowing or emptying a set fires a named failure instead of silently deselecting cases.
- `tests/test_driver_seam.py`'s isolation fixture correctly moves from copy-the-table to empty-the-table now that the package auto-registers a driver at import, keeping the empty-table refusal tests meaningful.
- The Livox cloud is summarised into a fixed-shape dict on the DDS thread instead of shipped raw, with the pacing/tile work explicitly deferred to the dashboard track - good scope discipline.

PR #2740, opened 2026-08-26
fix(simulation): refuse a mesh asset whose vertex coordinate is not finite

cagataycali:
## Problem

Two things measure the vertices `load_mesh_geometry` returns: `mesh_aabb`, whose `(center, size)` becomes a scene object's collision proxy, and the `extent` `convert_mesh_to_usd` authors into the converted USD. Neither reported a coordinate that is not finite.

`min`/`max` order a NaN as neither smaller nor larger than anything, so `mesh_aabb`'s running comparison dropped such a vertex and returned the bounds of the vertices that *were* finite - numerically indistinguishable from a mesh that declared only those, under no error at all. That is exactly the outcome the module's ASCII-STL refusal for an unterminated facet already exists to prevent, reached from the other side: there the vertices are in the bounds and their triangle is in no face, here the vertex is in a face and not in the bounds.

The `extent` is built by a different spelling - `min` over an iterable rather than a running comparison - which keeps a leading NaN and drops a trailing one. So one module could measure one file two ways, decided by the order its vertices happened to be declared in. An infinite coordinate needed no such subtlety: the reported extent was simply unbounded, and a scene object sized from it is a collision proxy with no bounds.

### Measured, on the same MJCF

A four-vertex `.msh` tetrahedron whose vertex 0 is extremal on -x, read through the public scene loader:

| `bowl.msh` | MuJoCo compiling the MJCF | `load_mjcf_scene_objects` on `main` |
| --- | --- | --- |
| four finite vertices | COMPILED | `status success`, size `(2.5, 2.0, 3.0)` |
| vertex 0's x is NaN | REFUSED `vertex coordinate 0 is not finite` | `status success`, size `(1.0, 2.0, 3.0)` |

The x extent collapses by 60% and the centre shifts `-0.25 -> 0.5`, so the proxy no longer bounds the geometry the file declares - and the scene MuJoCo will not compile at all loads successfully. With the bad vertex at the origin instead, the reported box is *byte-identical* to the healthy asset's, so nothing distinguishes them.

Across the four parse paths and three non-finite values, `mesh_aabb` accepted all 12 combinations while MuJoCo refused all 12.

## Change

One finiteness rule, applied where the four parse paths meet (OBJ, ASCII STL, binary STL, legacy MuJoCo MSH), rather than one per format:

- No format can drift into tolerating what its siblings refuse. Copying the rule into a single parser instead is pinned as a break: it leaves the other three accepting.
- Both bounds consumers are covered by the reader they share, so neither needs its own guard and they cannot disagree about one file again.
- The refusal names the offending vertex **by index**. That is the locator MuJoCo reports for the same refusal, and the only one that works for the two binary paths, which have no lines.
- The fast path is a single C-level `map` over the flattened coordinates - measured at **4.11%** of the parse it follows across the 60 largest shipped meshes (5,057,084 vertices). The per-vertex walk that finds the offender runs only after that map has failed. A naive per-vertex genexp was 15.10%.
- It is a finiteness rule, not a magnitude bound. float32 max, its negative, float32 tiny, `-0.0` and `1e300` are pinned as accepted; a mutation that adds a magnitude ceiling fires those controls.

Refusing is the format family's own disposition rather than a local policy: MuJoCo owns `.msh` and reads the other two, and refuses the same input by name. The six MuJoCo readings are pinned as premises so the agreement is graded, not asserted.

`convert_mesh_to_usd` also reads the asset **before** probing for `pxr`, the way the extension and existence checks above it already did. An install without the `sim-isaac` extra used to be told to install `usd-core` for an asset that was simply broken; it is now told what is wrong with the mesh it handed over. The content-addressed cache hit still short-circuits both, and an entry only exists because an earlier call parsed the same bytes successfully.

## Sim artifact

Grey is the asset as MuJoCo compiles it; orange is the box `mesh_aabb` reports, which becomes the collision proxy. Identical camera and lighting in both render panels, headless EGL.

mesh vertex finiteness

Self-checked inside the figure script: the mesh silhouette falls **0** pixels outside the correct proxy and **24123** outside the one reported for the non-finite file; the two panels differ in 51477 pixels; both panels are asserted to be embedded in the PNG verbatim by array comparison. Panel 2 draws the finite sibling's geometry, because MuJoCo refuses to compile the non-finite asset at all - the box is the one `mesh_aabb` really returned for that file.

## Nothing legitimate is refused

All **10469** mesh files in the downloaded robot-description corpus (3499 STL, 6882 OBJ, 88 MSH) parse unchanged, with zero non-finite coordinates among them. The defect was latent in shipped assets rather than live, and the change is provably zero-impact on them.

A `.usd` input is deliberately unchanged: it is referenced verbatim and this module never reads its vertices, so the rule is scoped to the formats it parses. That boundary is pinned, along with the fact that a *missing* USD input is still refused.

## Tests

`tests/simulation/isaac/test_mesh_vertex_coordinates_must_be_finite.py`, 92 cases.

Pre-fix split, raw revert of the module: **57 failed / 35 passed**. **0 of the 27** cells in the two control classes fail (`TestAFiniteMeshIsUnchanged`, `TestEachMalformationIsStillNamedByItsOwnMessage`); the 8 other passing cells are the 6 MuJoCo premises, the extension-vocabulary check and the USD-passthrough boundary.

The two halves are complementary and their union is exact:

| reverted | fires |
| --- | --- |
| guard removed, `pxr` ordering kept | 54 |
| ordering reverted, guard kept | 6 |
| both (the raw revert) | 57 |

54 + 6 with 3 shared (the USD-conversion refusals, which fail if either half is missing) = 57, and the union equals the raw-revert set exactly.

Break table - 11 mutations, **11 distinct failing-id sets**, none undetected, control clean, source restored byte-identical:

| mutation | fires | also fires in the pre-existing suite |
| --- | --- | --- |
| guard removed | 54 | 0 |
| rule copied into the OBJ parser only | 37 | 0 |
| locator hardcoded to vertex 0 | 4 | 0 |
| NaN only, infinity accepted | 17 | 0 |
| infinity only, NaN accepted | 35 | 0 |
| only the first vertex checked | 17 | 0 |
| message drops the vertex index | 12 | 0 |
| message drops the shared phrase | 48 | 0 |
| read moved back after the `pxr` probe | 6 | 0 |
| over-reach: a magnitude bound | 4 | 0 |
| over-reach: parser refusals re-badged as the finiteness one | 69 | **8** |

The last row is the one that matters for scope: re-badging every parser refusal as the finiteness one trips 8 cells in the pre-existing `test_mesh_assets.py` as well, so the module's named diagnoses are held in place from both sides.

## Coverage

`strands_robots/simulation/isaac/mesh_assets.py`: **51 -> 23 missing statements, 75% -> 89%**; package uncovered `1851 -> 1824`. The malformed-input refusals that had never been executed are now exercised as controls, and the pxr-free part of `convert_mesh_to_usd` - which sat behind a class-level `importorskip("pxr")` - is reachable. Everything still uncovered is the USD authoring half, which needs `pxr`.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1559 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `main` worktree, normalized message sets byte-identical in both directions, 0 in the changed files.
- Full `pytest tests/` on both trees: base `29 failed, 39012 passed, 309 skipped, 24 errors`; branch `29 failed, 39104 passed, 309 skipped, 24 errors`. **+92 is exactly the new file**, and the 53 failing/erroring ids are identical with the set difference empty in both directions (44 need `awsiot` from the `mesh-iot` extra, the rest pre-existing environment gaps).
- `tests/simulation/isaac/`: 1637 passed, 5 skipped on mujoco 3.5.0, the declared floor. The changed files also pass on mujoco 3.12.0.

### DIFF
--- a/changelog.d/2740-mesh-vertex-coordinates-must-be-finite.md
+++ b/changelog.d/2740-mesh-vertex-coordinates-must-be-finite.md
@@ -1,0 +1,51 @@
+### Fixed: a mesh asset declaring a vertex coordinate that is not finite is refused, not measured around
+
+Two things measure the vertices `load_mesh_geometry` returns: `mesh_aabb`, whose `(center, size)`
+becomes a scene object's collision proxy, and the `extent` `convert_mesh_to_usd` authors into the
+converted USD. Neither reported a non-finite coordinate.
+
+`min`/`max` order a NaN as neither smaller nor larger than anything, so `mesh_aabb`'s running
+comparison dropped such a vertex and returned the bounds of the vertices that *were* finite -
+numerically indistinguishable from a mesh that declared only those, under no error at all. That is
+exactly the outcome the module's ASCII-STL refusal for an unterminated facet already exists to
+prevent, reached from the other side: there the vertices are in the bounds and their triangle is in
+no face, here the vertex is in a face and not in the bounds. The `extent` is built by a different
+spelling - `min` over an iterable rather than a running comparison - which keeps a leading NaN and
+drops a trailing one, so one module could measure one file two ways depending on the order its
+vertices happened to be declared in. An infinite coordinate needed no such subtlety: the reported
+extent was simply unbounded, and a scene object sized from it is a collision proxy with no bounds.
+
+Measured on a four-vertex `.msh` tetrahedron whose vertex 0 is extremal on -x: replacing that one
+coordinate with NaN collapsed the reported proxy from `2.5 m` to `1.0 m` on that axis and shifted
+its centre from `-0.25` to `0.5`, and `load_mjcf_scene_objects` returned that scene object under
+`status success` for an MJCF MuJoCo refuses to compile. Rendered headless, a quarter of the asset's
+silhouette (24123 pixels) falls outside its own reported proxy, against 0 for the same asset with
+four finite vertices.
+
+One finiteness rule is now applied where the four parse paths meet - OBJ, ASCII STL, binary STL and
+legacy MuJoCo MSH - so no format can drift into tolerating what its siblings refuse, and so both
+bounds consumers are covered by the reader they share rather than each needing its own guard. The
+refusal names the offending vertex by index, which is the locator MuJoCo reports for the same
+refusal and the only one that works for the two binary paths. The fast path is a single C-level
+`map` over the flattened coordinates, measured at ~4% of the parse it follows across the 60 largest
+shipped meshes; the per-vertex walk that finds the offender runs only after that map has already
+failed. It is a finiteness rule and not a magnitude bound, so a legitimately huge or vanishing
+coordinate is still a position - float32 max, its negative, float32 tiny, `-0.0` and `1e300` are
+all pinned as accepted.
+
+Refusing is the format family's own disposition rather than a local policy: MuJoCo owns `.msh` and
+reads the other two, and it refuses the same input with `vertex coordinate N is not finite`. All
+twelve fixture and format combinations now agree with it, and the six MuJoCo readings are pinned as
+premises so the agreement is graded rather than asserted.
+
+`convert_mesh_to_usd` also reads the asset before probing for `pxr`, the way the extension and
+existence checks above it already did. An install without the `sim-isaac` extra used to be told to
+install `usd-core` for an asset that was simply broken; it is now told what is wrong with the mesh
+it handed over. The content-addressed cache hit still short-circuits both, and a cache entry only
+exists because an earlier call parsed the same bytes successfully.
+
+Nothing legitimate is refused: all 10469 mesh files in the downloaded robot-description corpus -
+3499 STL, 6882 OBJ, 88 MSH - parse unchanged, with zero non-finite coordinates among them, so the
+defect was latent rather than live in shipped assets. Deliberately unchanged: a USD input, which is
+referenced verbatim and whose vertices this module never reads, so the rule is scoped to the formats
+it parses.
--- a/docs/simulation/isaac.md
+++ b/docs/simulation/isaac.md
@@ -203,7 +203,8 @@ open a window.
   [#2300](https://github.com/strands-labs/robots/issues/2300)) - and
   collision uses the mesh's **convex hull**, also the MuJoCo contract, with
   the same caveat for concave assets: the hull fills every cavity. A missing
-  file or an unconvertible format is refused up front, never realized as a
+  file, an unconvertible format, or an asset declaring a vertex coordinate
+  that is not finite is refused up front, never realized as a
   fallback primitive.
 - **Cameras & rendering** - `add_camera` (look-at, FOV), `render` (RGB + depth).
   World-fixed only: `parent_body` (a body-mounted wrist camera, supported on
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Adds a single finiteness guard at the point where the four mesh parse paths (OBJ, ASCII STL, binary STL, legacy MSH) converge in `load_mesh_geometry`, so an asset declaring a NaN or infinite vertex coordinate is refused with a `ValueError` naming the asset, the vertex index, and the value - instead of being silently mis-measured by `mesh_aabb` and the USD `extent` (NaN falls out of `min`/`max`, producing bounds indistinguishable from a mesh that declared only the finite vertices). This matches MuJoCo's own refusal for the same input and the module's stated fail-loud contract. The PR also moves the geometry read in `convert_mesh_to_usd` ahead of the optional `pxr` import probe so a caller without the `sim-isaac` extra is told what is wrong with the asset rather than what is missing from the environment - a deliberate, commented, and tested reorder that leaves the cache-hit short-circuit intact.

## What's good

- The guard lives at the shared reader rather than being copied into each parser, and the test suite pins that structurally (single occurrence, guard-after-dispatch AST check, `MESH_EXTENSIONS` coverage derived from the module's own vocabulary).
- Over-reach controls: extreme-but-finite coordinates (float32 max, 1e300, -0.0) are asserted unchanged, and every pre-existing malformation is asserted to still raise its own message rather than being re-badged.
- `Raises:` docstrings updated in the same change (AGENTS.md > "A `Raises:` block names every refusal the function itself makes"), error strings are plain ASCII, changelog fragment present, and the failed conversion is verified to leave no cache entry behind.
- The MuJoCo-oracle tests are `importorskip`-gated and use per-fixture directories to dodge MuJoCo's compiled-mesh cache.

yinsong1986 (approved):
## Summary

Adds one finiteness rule at the point where the four mesh parse paths (OBJ, ASCII STL, binary STL, legacy MSH) converge in `load_mesh_geometry`, so an asset declaring a NaN or infinite vertex coordinate is refused with a `ValueError` naming the asset, the offending vertex index, and the value - instead of being silently mis-measured by `mesh_aabb` (whose `(center, size)` becomes a collision proxy) and the USD `extent`. `convert_mesh_to_usd` also reads the asset before probing for `pxr`, so a broken mesh is diagnosed as broken rather than as a missing optional dependency. The guard is finiteness-only (extreme-but-finite values pinned as accepted), matches MuJoCo's own refusal for the same inputs (pinned as premises), and is verified zero-impact on the 10469-file mesh corpus.

## What's good

- Placement at the shared reader covers both bounds consumers with one rule and is pinned structurally against a per-format copy drifting.
- `Raises:` blocks on `load_mesh_geometry`, `mesh_aabb` and `convert_mesh_to_usd` all name the new refusal (AGENTS.md's Raises-completeness rule).
- Controls pin the scope from both sides: extreme-but-finite coordinates stay accepted, every pre-existing malformation keeps its own diagnosis, and a USD input stays a verbatim passthrough.
- Changelog fragment, docs update, `monkeypatch.setenv` in tests, ASCII-only user-facing strings, no host paths.

Verified locally at head `904e4c92`: the new file (92 cases) plus the pre-existing `test_mesh_assets.py` pass together (130 passed), and `ruff check` / `ruff format --check` are clean on both changed Python files.

Issue #272, opened 2026-05-30
mesh(safety): wire leak — _resume_lockout structured reason published via strands/+/safety/event broadcast


cagataycali:
Tracked from PR #225 review thread on `strands_robots/mesh/core.py:2280` (and siblings at 2267, 2283).

## Concern

The surrounding docstring (lines 2185–2196) and the inline comment at line 2204 claim the structured rejection detail is preserved in the *local* `publish_safety_event` audit record. In practice `publish_safety_event` (sensors.py:465) publishes the full payload onto `strands/{peer_id}/safety/event`, which any peer subscribed to `strands/+/safety/event` (a discovery-friendly broadcast topic) can read.

A peer probing `_resume_lockout` with a bad override code can:
1. Read the generic `{"error": "resume rejected"}` from their own RPC response, AND
2. Subscribe to the broadcast safety-event topic and read `payload.reason: "STRANDS_MESH_OVERRIDE_CODE not configured"` / `"bad override code"` / `"lockout not engaged"` — the exact oracle the wire-response generification (R7-4) was meant to close.

This is a content-channel oracle that the timing-channel oracle pin (`TestResumeLockoutTimingOracleClosed`) does not cover.

## Threat model

- Adversary holds an override-code probe (does not need a valid code; the oracle is on the rejection path).
- Adversary subscribes to `strands/+/safety/event` (any mesh peer can do this).
- For each probe, observes `payload.reason` and learns whether the configuration state was unconfigured / lockout-not-engaged / bad-code, narrowing the search.

## Proposed fix

Two paths:

**Option A — split publish from local audit:**
```python
# Wire-published event: generic
publish_safety_event(self.peer_id, "resume_denied", {"reason": "rejected"})
# Local audit only: structured
audit.log_safety_event("resume_denied", self.peer_id, {"reason": <structured>, "sender": sender})
```

**Option B — coarse enum on wire, structured local:**
```python
WIRE_REASON_OPAQUE = "rejected"
publish_safety_event(self.peer_id, "resume_denied", {"reason": WIRE_REASON_OPAQUE})
audit.log_safety_event("resume_denied", self.peer_id, {"structured_reason": <branch-tag>, "sender": sender})
```

Option A is cleaner — the audit module already has a direct API; only the wire publish needs to go through `publish_safety_event`. This also resolves the related concern in #256 (post-compare branch I/O asymmetry) by collapsing all three rejection branches to a single uniform wire shape.

## Pin tests required

[...]

PR #2741, opened 2026-08-26
fix(mesh): coerce a safety event's payload so the event reaches the wire

cagataycali:
## Problem

`SensorLoopsMixin` builds nine records addressed to a `strands/<peer>/...` topic. Eight of them -- every `_read_*` reader -- pass the record through `_coerce_record` before it is published, because a sensor pipeline reports its readings as numpy and `json.dumps` refuses a `float32` (#2730). `publish_safety_event` builds the ninth, and did not coerce it.

The consequence is total rather than partial. `session._put_zenoh_directly` encodes the payload before it reaches the wire, so the event was never published at all -- and the failure is not transient, so no later call recovers it. The audit half of the *same* call, meanwhile, wrote a `sig="SERIALISE_FAILED"` poison record and logged at ERROR. One call, two reports: a forensic trail asserting a safety event was raised, and no peer having received it.

That disagreement is the one `session._report_unencodable_payload` describes in its own docstring:

> `publish_safety_event` writes the event to the wire AND to the local audit log; on an unencodable payload the audit half records a `sig="SERIALISE_FAILED"` poison record and logs at ERROR (naming the peer and the reason), while the wire half dropped the event silently.

\#2638 raised that report from DEBUG to ERROR so an operator could at least see the drop. The wire half still published nothing, and `_coerce_record` did not exist yet.

## Measured

Driven through the production encoder and the production reporter, one capture script run against both trees. The call is `publish_safety_event("joint_limit_tripped", severity="critical", payload={"joint": "shoulder_pan", "q": np.float32(2.9712), "limit": np.float32(2.9670)})`:

|---|---|---|
| messages published | **0** | 1 |
| transport report | `ERROR ... the payload cannot be JSON-encoded, so this message and every later one built the same way is dropped before it reaches the wire ... Object of type float32 is not JSON serializable` | nothing reported |
| audit records written | 1 | 1 |
| audit payload serialises | **no** (poison-record path) | yes |
| audit severity | `critical` | `critical` |

Which readings were lost depended on the numpy *width*, which is why the gap was easy to miss. For a payload of `{"v": <value>}`:

| value | before | after |
|---|---|---|
| `np.float32(2.97)` | dropped (`TypeError`) | published |
| `np.int64(3)` | dropped (`TypeError`) | published |
| `np.bool_(True)` | dropped (`TypeError`) | published |
| `np.array([1.5, -2.5])` | dropped (`TypeError`) | published |
| `np.float64(2.97)` | published | published |
| `2.97` | published | published |

`np.float64` subclasses Python's `float`, so a payload built from one always encoded and the same code reading a `float32` dropped every event.

The in-file control is decisive: the *same* `np.float32(0.05)` handed to a sibling reader, `_read_map_info`, reaches the wire on **both** trees (`resolution = 0.05000000074505806`). One class, one wire format, two conventions -- and the one without a coercion is the safety topic.

## Reachability

A reading is what a safety payload naturally carries: the joint value that tripped a limit, the distance that closed. `publish_safety_event` is public and documented in `examples/fleet/dashboard.py` as part of the mesh surface, and its docstring named no payload contract at all, while its audit sibling `log_safety_event` documents "Must be JSON-serialisable".

The in-tree callers in `Mesh` all pass strings, ints and lists of peer ids, so none of the built-in safety events was affected. This is a latent defect on the public surface, stated plainly rather than dressed up.

## Change

Coerce the payload once, and hand the same coerced mapping to both halves, so they report one event instead of two. `+34/-3` in `strands_robots/mesh/sensors.py`.

Coercion happens in a copy, so the caller keeps the mapping they passed unedited -- the same reason `_coerce_record` replaces a nested container rather than editing the provider's own -- and a fire-and-forget publish no longer holds an object the caller can still mutate after the call returns.

Nothing else about the event changes: the wire severity is still uniformly `"info"` (#272), the real severity still reaches the audit record only, a stopped host still publishes nothing, and an audit-write failure is still survived.

The boundary is that coercion repairs readings rather than laundering payloads. A value that is genuinely not a reading is still passed through untouched, so the transport still reports it by name -- substituting a repr would publish a record that misstates what the safety path saw, which is worse than a reported drop.

## Why every gate was silent

`tests/mesh/test_sensor_readers.py` drives `publish_safety_event` through a host whose `publish` **records** the payload. The encoder runs one layer down, in the transport, so a recording double accepts a payload the wire never would. `tests/mesh/test_safety_event_payload_reaches_the_wire.py` uses a host that encodes what it is handed, which is the only way a unit test can see which side of the encoder the record landed on.

## Tests

`tests/mesh/test_safety_event_payload_reaches_the_wire.py`, 31 cells. Pre-fix: **11 failed / 20 passed**, with 0 of the 6 cells in `TestWhatIsUnchanged`, 0 of the 3 premises, and 0 of the 2 over-reach cells failing. Each reading row carries a premise that its raw value is what the encoder refuses, so the table cannot pass on a value that never needed coercing.

The rule that every record bound for the wire is coerced is derived from the module rather than listed: a method containing a dict literal with a `"peer_id"` key is building one. That selects exactly the nine builders, with no false positive in either direction (`coercing - builders` is empty), so a publisher added later is held to it on arrival.

Eleven plausible regressions applied, **nine distinct firing sets**, control clean, source restored byte-identical:

| mutation | fires | notes |
|---|---|---|
| the coercion is removed | 10 | the defect |
| wire half coerced, audit half keeps the raw payload | 1 | the halves disagree again, the other way round |
| audit half coerced, wire half keeps the raw payload | 10 | |
| coerced in place, aliasing the caller's mapping | 2 | |
| values stringified instead of coerced | 10 | |
| coerced after the event is published | 9 | placement is load-bearing |
| `_jsonable` inlined instead of the shared `_coerce_record` | 1 | behaviourally equivalent; the derived pin is what fires |
| wire severity leaks the real severity (#272) | 8 | over-reach guard |
| a tenth record builder is planted, uncoerced | 1 | the derived pin catches it |
| the same plant, builder set hardcoded to today's nine | **0** | why the set is derived |

The last two rows are a pair: hardcoding the builder set makes the planted publisher invisible, which is the argument for deriving it. The `_jsonable`-inlined row shares its firing set with the plant for a structural reason -- both are "a record builder does not call the shared helper" -- and it is a behavioural no-op, so the structural cell is what pins the single-sourcing rather than a behavioural one.

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1560 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `main` worktree, normalized message sets byte-identical in both directions, 0 in the changed files.
- Paired run over an identical 565-file selection (all of `tests/mesh/` plus every guard that walks the tree, reads source or docstrings, or reads `changelog.d`), the new file excluded from both trees: **24 failed, 101 skipped, 24 errors on both** (23108 vs 23106 collected), **48 failing/erroring ids each, `comm` empty in both directions**. 45 of the 48 need `awsiot`/`awscrt` (declared in the `[mesh-iot]` extra, absent here -- confirmed with `tomllib` + `find_spec`); the remaining three are pre-existing on both trees.
- The `+2` collected localises entirely to `tests/test_args_docstring_completeness.py` (446 vs 444), which parametrises over documented functions -- `publish_safety_event` gained an `Args:` block.
- `tests/mesh/test_zenoh_transport_security.py` is excluded from the paired selection and verified alone: **9 passed, 2 skipped on both trees**.
- The changed file plus `tests/mesh/test_sensor_readers.py`: 129 passed.

## Artifact

Both halves of the one call, on both trees, driven through the production encoder and the production once-per-topic reporter. The audit half is stubbed to record its arguments instead of writing to disk; everything else is production code.

the wire half and the audit half of one safety event, on both trees

### DIFF
--- a/changelog.d/2741-safety-event-payload-reaches-the-wire.md
+++ b/changelog.d/2741-safety-event-payload-reaches-the-wire.md
@@ -1,0 +1,59 @@
+### Fixed: a safety event carrying a reading reaches the wire instead of being dropped
+
+`SensorLoopsMixin` builds nine records addressed to a `strands/<peer>/...` topic. Eight of
+them -- every `_read_*` reader -- pass the record through `_coerce_record` before it is
+published, because a sensor pipeline reports its readings as numpy and `json.dumps` refuses
+a `float32` (#2730). `publish_safety_event` builds the ninth, and did not coerce it.
+
+The consequence is total rather than partial: `_put_zenoh_directly` encodes the payload
+before it reaches the wire, so the event was never published at all. It is not a transient
+failure the next call retries either -- every call built the same way fails identically.
+Meanwhile the audit half of the *same* call wrote a `sig="SERIALISE_FAILED"` poison record
+and logged at ERROR, so the two halves of one call reported two different things: a forensic
+trail asserting that a safety event was raised, and no peer having received it.
+
+That disagreement is the one `session._report_unencodable_payload` describes in its own
+docstring. #2638 raised the report from DEBUG to ERROR so an operator could at least see the
+drop; the wire half still published nothing, and `_coerce_record` did not exist yet.
+
+Which readings were lost depends on the numpy *width*, which is why the gap was easy to
+miss. Measured through the encoder for a payload of `{"v": <value>}`:
+
+| value | before | after |
+|---|---|---|
+| `np.float32(2.97)` | dropped (`TypeError`) | published |
+| `np.int64(3)` | dropped (`TypeError`) | published |
+| `np.bool_(True)` | dropped (`TypeError`) | published |
+| `np.array([1.5, -2.5])` | dropped (`TypeError`) | published |
+| `np.float64(2.97)` | published | published |
+| `2.97` | published | published |
+
+`np.float64` subclasses Python's `float`, so a payload built from one always encoded and the
+same code reading a `float32` dropped every event. A reading is exactly what a safety
+payload carries -- the joint value that tripped a limit, the distance that closed -- and the
+public method's own docstring named no payload contract, while its audit sibling
+`log_safety_event` documents "Must be JSON-serialisable".
+
+The payload is now coerced once and the same coerced mapping is handed to both halves, so
+they report one event rather than two. Coercion happens in a copy: the caller keeps the
+mapping they passed unedited, and a fire-and-forget publish no longer holds an object the
+caller can still mutate after the call returns. Nothing else about the event changes -- the
+wire severity is still uniformly `"info"` (#272), the real severity still reaches the audit
+record only, a stopped host still publishes nothing, and an audit-write failure is still
+survived.
+
+The boundary is that coercion repairs readings rather than laundering payloads. A value that
+is genuinely not a reading is still passed through untouched, so the transport still reports
+it by name -- substituting a repr would publish a record that misstates what the safety path
+saw, which is worse than a reported drop.
+
+`tests/mesh/test_safety_event_payload_reaches_the_wire.py` drives the method through a host
+that *encodes* what it is handed, which is what the pre-existing coverage did not: the
+`publish` double in `tests/mesh/test_sensor_readers.py` records the payload, and the encoder
+runs one layer down in the transport, so a recording double accepts a payload the wire never
+would. Each row carries a premise that its raw value is what the encoder refuses. The rule
+that every record bound for the wire is coerced is derived from the module -- a method
+building a dict literal with a `peer_id` key is building one -- so a publisher added later is
+held to it on arrival; planting a tenth uncoerced builder fails the derived pin and passes a
+hardcoded one. Eleven plausible regressions were applied: nine fire a different set and the
+control is clean.
--- a/strands_robots/mesh/sensors.py
+++ b/strands_robots/mesh/sensors.py
@@ -663,9 +663,40 @@ class SensorLoopsMixin:
         severity: str = "warning",
         payload: dict[str, Any] | None = None,
     ) -> None:
-        """Publish a safety event to the mesh AND write to audit log."""
+        """Publish a safety event to the mesh AND write to the audit log.
+
+        The payload is coerced once through :func:`_coerce_record` and the same
+        coerced mapping is handed to both halves, which is what makes the two
+        halves report one event instead of two different ones.
+        :func:`strands_robots.mesh.session._report_unencodable_payload` records
+        why that matters: on a payload the JSON encoder refuses, the audit half
+        writes a ``sig="SERIALISE_FAILED"`` poison record and logs at ERROR,
+        while the wire half publishes nothing at all - and the failure is not
+        transient, so no later tick recovers it. A reading expressed in a
+        foreign numeric type is exactly what a safety payload carries (the joint
+        value that tripped a limit, the distance that closed), and every other
+        record this mixin sends to the wire is coerced before it goes.
+
+        Coerced into a copy rather than in place, so the caller keeps the mapping
+        they passed unedited - the same reason :func:`_coerce_record` replaces a
+        nested container instead of editing the provider's own.
+
+        A value that is genuinely not a reading is still passed through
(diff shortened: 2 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Routes `publish_safety_event`'s payload through the same `_coerce_record` step every other wire-bound record in `SensorLoopsMixin` already takes, so a safety event carrying a numpy reading (the joint value that tripped a limit, a `float32` distance) is JSON-encodable and actually reaches the wire instead of being dropped in the transport while the audit half writes a `SERIALISE_FAILED` poison record. The coercion runs on a shallow copy (`dict(payload)`), so the caller's mapping — including nested containers — is left unedited, and both the wire event and the audit record are built from the one coerced mapping, so the two halves report the same event. Pre-existing behaviour is preserved: wire severity stays uniformly `"info"`, the real severity reaches the audit record only, a genuinely unrepresentable object is still passed through (and reported by name) rather than stringified, and an audit-write failure still doesn't block the publish.

## What's good

- The regression test encodes the way the transport does (`json.dumps` in the test double), which is exactly why the prior recording-double coverage was silent on this — and it pins that premise (`test_the_transport_encodes_before_it_puts`) so the double can't drift from production.
- `TestEveryRecordBoundForTheWireIsCoerced` derives the record-builder set from the module's AST rather than listing nine names, so a tenth publisher added later is held to the same rule — the same derive-don't-list pattern AGENTS.md pins elsewhere (frame readers, bus access).
- Copy-not-in-place coercion, verified from both directions (caller's mapping unchanged; published payload not the caller's object).
- Verified locally at 556a808b: all 31 new tests pass, `ruff check` / `ruff format --check` clean on both files, and the only `tests/mesh/` failures (IoT transport, sim-camera publish) reproduce identically on `upstream/main` — environmental, not attributable to this diff.

Note: the diff carries no `changelog.d/` fragment yet; the fragment gate will surface that on its own (AGENTS.md step 3), so it's not flagged as a blocker here.

yinsong1986 (approved):
## Summary

Routes `publish_safety_event`'s payload through the same `_coerce_record` step the eight `_read_*` builders in `SensorLoopsMixin` already take, closing the gap where a safety event carrying a numpy reading (`float32`, `int64`, `bool_`, an `ndarray`) was refused by `json.dumps` in the transport and never reached the wire, while the audit half of the same call wrote a `SERIALISE_FAILED` poison record. The coercion runs once on a shallow copy (`dict(payload)`) and the same coerced mapping feeds both the wire event and the audit record, so the two halves now report one event; the caller's mapping (including nested containers, since `_jsonable` rebuilds rather than mutates) is left unedited. Everything else is preserved: wire severity stays uniformly `"info"` (#272), the real severity reaches the audit record only, a genuinely unrepresentable object is still passed through and reported by name rather than stringified, a stopped host publishes nothing, and an audit-write failure does not block the publish. Previously-encodable payloads (`np.float64`, plain floats, tuples) produce byte-identical JSON through `_jsonable`, so nothing already on the wire changes shape.

The production code is byte-identical to the head the prior review verified; the only new commit adds the `changelog.d/2741-safety-event-payload-reaches-the-wire.md` fragment that review noted was outstanding, which satisfies AGENTS.md step 3.

## What's good

- The test host encodes the way the transport does and pins that premise (`test_the_transport_encodes_before_it_puts`), which is exactly why the pre-existing recording-double coverage was silent on this defect.
- `TestEveryRecordBoundForTheWireIsCoerced` derives the record-builder set from the module's AST rather than hardcoding nine names, so a publisher added later is held to the rule on arrival - the derive-don't-list pattern AGENTS.md pins elsewhere.
- Copy-not-in-place coercion is verified from both directions (caller's mapping unchanged; published payload not the caller's object), and the over-reach guard pins that coercion does not become stringification.
- `publish_safety_event` gained a full `Args:` block, and all added text is plain ASCII.

Verified at `01ca9309`: the new suite plus `tests/mesh/test_sensor_readers.py` pass together (129 passed; the run's coverage-threshold failure is an artifact of the partial file selection, not a test failure).

PR #2742, opened 2026-08-26
fix(simulation): bound a body's collision proxy by the geoms MuJoCo can collide

cagataycali:
## Problem

`_body_collision_aabb` computes the axis-aligned box that stands in for a body's physics footprint, and it is what `load_mjcf_scene_objects` reports as a `SceneObject`'s `size` and `position`. It selected geoms with `group == "0"`, falling back to every geom when that matched nothing. That is wrong in two independent ways.

`group` carries no contact meaning in MuJoCo - it is a visualiser toggle - and the conventions built on it disagree: Menagerie marks *collision* geoms with `group="3"` where robosuite marks *visual* ones with `group="1"`. So the attribute cannot answer which geoms the solver will touch, which is the only question this function asks.

Separately, the comparison was against the literal string. A geom that omits `group` **is** group 0 - MuJoCo resolves it to `geom_group == 0` - but the attribute reads as absent, so the `"0"` pass skipped it.

### Both directions are live

**It over-bounds.** In a Menagerie-shaped body every geom falls through to the every-geom fallback, so the proxy becomes the union of the decorative shell and the collision primitive. On the shipped `franka_emika_panda` `mjx_hand`:

| `left_finger` | reported | MuJoCo's collision geometry |
|---|---|---|
| proxy full size (m) | `(0.022, 0.02602, 0.0532)` | `(0.0175, 0.0152, 0.0165)` |
| proxy centre (m) | `(0.0, 0.01299, 0.0269)` | `(0.0, 0.00758, 0.04525)` |

6.9x the volume, 3.2x the length along the finger, and centred on the wrong part of it. Of that body's seven geoms exactly one - `left_finger_pad` - is left able to take part in a contact (`conaffinity="3"`), and not one of them carries `group="0"`.

**It also under-bounds.** Where the `"0"` pass did match, it narrowed: a body whose two collidable geoms spell `group="0"` and `group="3"` dropped the second one entirely, reporting `(0.2, 0.2, 0.2)` for geometry spanning `(0.2, 0.2, 0.6)`.

## Change

The signal read is now the format's own contact declaration, through the existing `_geom_cannot_collide` - already the tie-break `_mesh_geom_visual_rank` ranks first for the mirror question of which mesh is a body's *visual* asset. MuJoCo lets two geoms touch only when `contype1 & conaffinity2` or `contype2 & conaffinity1` is non-zero, so a geom declaring both as `0` can never take part in a contact. A body whose every analytic geom is contact-free still bounds all of them.

Reading either half of the pair alone would be wrong, and that is pinned: `contype="0" conaffinity="1"` still collides with anything declaring `contype` non-zero, so such a geom stays inside the bound.

**The `group` tier is dropped here rather than demoted the way `_mesh_geom_visual_rank` keeps it**, and that is a measurement rather than a preference: keeping it as a refinement inside the collidable tier changes the answer for **0** of the 1209 corpus bodies below, whereas for the visual-mesh question it is the only signal some files carry.

## Graded against MuJoCo

MuJoCo is the oracle: for each body its collision AABB is computed from `geom_contype`/`geom_conaffinity`/`geom_size`/`geom_pos` off a compiled `MjModel`, so `<default>` inheritance and the contact declaration are resolved by MuJoCo rather than by the code under test.

Restricted to bodies whose collidable analytic geoms are axis-aligned, and therefore comparable to an axis-aligned bound - 336 compilable MJCF files, 427 such bodies:

| | matches MuJoCo | worse than before |
|---|---|---|
| before | 356 / 427 | - |
| after | **427 / 427** | **0** |

Over the wider corpus of 554 files and 1209 bodies with an analytic answer: **72 change, all 72 shrink, none grows**, and 27 also move their centre onto the geometry they stand for.

## Sim artifact

`mjx_hand.xml` rendered headless (`MUJOCO_GL=egl`), unmodified except for the two translucent proxy boxes drawn from the loader's own reported centre and size. All three panels share one camera and one headlight.

collision proxy before and after

Pixel counts are unoccluded silhouettes taken from per-category segmentation passes, so the enclosing box does not hide what it contains:

| | before | after |
|---|---|---|
| proxy px outside the collidable geom | 54106 | **0** |
| collidable-geom px outside the proxy | 0 | **0** |

The new proxy's silhouette is exactly the collidable geom's (28878 px each), which is what an AABB of a single axis-aligned box should be.

## Tests

`tests/simulation/isaac/test_body_collision_bound_prefers_collidable_geoms.py`, 17 cases. Pre-fix **7 fail / 7 pass**, and the 7 that pass are the four MuJoCo premises, the two `TestABodyWithNothingCollidableKeepsItsBound` controls and the single-geom oracle case - nothing in a premise or control class fires.

Break table, 9 mutations against the branch (`isaac` = the rest of `tests/simulation/isaac/`, 1637 tests):

| mutation | mine | isaac |
|---|---|---|
| control (unmutated) | 0 | 0 |
| preference reverted to the `group=="0"` filter | 7 | 0 |
| the contact filter applied in BOTH passes | 1 | 0 |
| predicate inverted (prefers the decorative geoms) | 5 | 0 |
| tier order reversed (every geom first) | 5 | 0 |
| only the `contype` half of the pair is read | 2 | 0 |
| the every-geom fallback tier dropped | 1 | 0 |
| attributes read off the element, not its default class | 5 | **1** |
| the `group=="0"` filter re-added on top of the new tier | 7 | 0 |
| the `group` paragraph deleted from the docstring, code kept | 0 | 0 |

8 of 9 detected, 5 distinct failing-id sets, source restored byte-identical. Three rows share a set for a structural reason - inverting the predicate, reversing the tiers and bypassing `_class_attrs` all reduce to "the decorative shell contributes" for a shell-shaped body, and none of them changes the group-only case. The over-reach row that bypasses `_class_attrs` additionally trips the pre-existing `test_mjcf_default_class_geometry.py::TestOneRuleAnswersWhatAGeomDeclared::test_the_scan_reaches_the_readers_it_grades`, which requires every geom reader to go through that one rule. The last row is an honest 0: nothing grades that prose.

## Gate

Full `pytest tests/` on both trees, `MUJOCO_GL=egl`, mujoco 3.5.0 (the declared floor) with lerobot 0.6.2 from source:

- base `26078397`: 39507 collected, `29 failed, 39145 passed, 309 skipped, 24 errors`
- branch: 39524 collected, `29 failed, 39162 passed, 309 skipped, 24 errors`

`+17` collected and `+17` passed is exactly the new file, and the failing/erroring id sets are **identical (53 each, both set differences empty)**. Those 53 are pre-existing on this host: 44 need `awsiot` (the `[mesh-iot]` extra), the rest are lerobot-from-source drift and a missing dist metadata entry.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1560 files). mypy `24 == 24` against a pristine `main` worktree with byte-identical normalized message sets and none in the changed files. The changed file also passes on mujoco 3.12.0.

One line of coverage moved in an untouched module (`simulation/mujoco/simulation.py`, 21 -> 22 missing, line 5186 - a branch inside a threaded policy path). It is run-to-run nondeterminism, not this change: `simulation/isaac/loaders.py` reports an identical `723 / 68 / 91%` on both trees, since the defect was in what a covered function computed rather than in an uncovered line.

## Not changed

`_extract_mjcf_shape` still reads a link's shape from its first `<geom>`. That is a different question - a kinematic link's display shape rather than a collision proxy - and its docstring documents the first-geom rule, so it is left alone here.

### DIFF
--- a/changelog.d/2742-body-collision-bound-prefers-collidable-geoms.md
+++ b/changelog.d/2742-body-collision-bound-prefers-collidable-geoms.md
@@ -1,0 +1,46 @@
+### Fixed: a body's collision proxy is bounded by the geoms MuJoCo can collide
+
+`_body_collision_aabb` computes the axis-aligned box that stands in for a body's physics
+footprint, and it is what `load_mjcf_scene_objects` reports as a `SceneObject`'s `size` and
+`position`. It selected geoms with `group == "0"`, falling back to every geom when that matched
+nothing, and that is wrong in two independent ways.
+
+`group` carries no contact meaning in MuJoCo - it is a visualiser toggle - and the conventions
+built on it disagree, MuJoCo Menagerie marking collision geoms with `group="3"` where robosuite
+marks visual ones with `group="1"`. So the attribute cannot answer which geoms the solver will
+touch, which is the only question this function asks. Separately, the comparison was against the
+literal string: a geom that omits `group` *is* group 0, and MuJoCo resolves it to `geom_group == 0`,
+but the attribute reads as absent so the `"0"` pass skipped it.
+
+Both directions are measurable. In a Menagerie-shaped body every geom falls through to the
+every-geom fallback, so the proxy becomes the union of the decorative shell and the collision
+primitive: the shipped `franka_emika_panda` `mjx_hand` fingers reported a
+`(0.022, 0.026, 0.0532)` proxy centred `(0, 0.013, 0.0269)` where the body's collision geometry is
+`(0.0175, 0.0152, 0.0165)` centred `(0, 0.0076, 0.04525)` - 6.9x the volume, 3.2x the length along
+the finger, and centred on the wrong part of it. Of that finger's seven geoms exactly one,
+`left_finger_pad`, is left able to take part in a contact (`conaffinity="3"`); not one of them
+carries `group="0"`. And where the `"0"` pass did match, it narrowed instead: a body whose two
+collidable geoms spell `group="0"` and `group="3"` dropped the second one entirely, reporting a
+bound that excludes geometry the solver collides.
+
+The signal read is now the format's own, through the existing `_geom_cannot_collide` - already the
+tie-break `_mesh_geom_visual_rank` ranks first for the mirror question of which mesh is a body's
+visual asset. MuJoCo lets two geoms touch only when `contype1 & conaffinity2` or
+`contype2 & conaffinity1` is non-zero, so a geom declaring both as `0` can never take part in a
+contact. Reading either half of that pair alone would be wrong: `contype="0" conaffinity="1"` still
+collides with anything declaring `contype` non-zero, and such a geom stays inside the bound.
+
+Graded against MuJoCo itself across the downloaded asset corpus - 336 compilable MJCF files, 427
+bodies whose collidable analytic geoms are axis-aligned and therefore comparable to an axis-aligned
+bound - the reported bound now matches MuJoCo's own on 427 of 427, against 356 of 427 before. No
+body is worse. Over the wider corpus of 554 files and 1209 bodies with an analytic answer, 72
+change, all 72 shrink, none grows, and 27 also move their centre onto the geometry they stand for.
+
+Deliberately unchanged: a body whose every analytic geom is contact-free. There is nothing
+collidable to prefer, so all of them are bounded exactly as before - an approximate proxy is still
+better than none - and that is pinned as a control, as is the single-geom case.
+
+The `group` tier is dropped here rather than demoted the way `_mesh_geom_visual_rank` keeps it.
+That is a measurement, not a preference: keeping it as a refinement inside the collidable tier
+changes the answer for 0 of those 1209 bodies, whereas for the visual-mesh question it is the only
+signal some files carry.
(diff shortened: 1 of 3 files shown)


yinsong1986 (approved):
## Summary

Replaces the `group == "0"` geom selection in `_body_collision_aabb` with the format's own contact declaration via the existing `_geom_cannot_collide` helper, so a body's collision proxy is bounded by the geoms MuJoCo can actually collide rather than by a visualiser toggle whose conventions disagree across Menagerie and robosuite. The old string comparison also skipped geoms that omit `group` (MJCF's spelling of group 0), which sent Menagerie-shaped bodies to the every-geom fallback and widened the proxy by the decorative shell. The fallback for bodies whose every analytic geom is contact-free is preserved, the function stays private, and the new test file grades the fix against a compiled `MjModel` as the oracle in both directions (over-bound and under-bound), including the half-declared `contype="0" conaffinity="1"` case that a naive single-attribute read would drop. One note: the PR description is still the "Placeholder" stub - worth filling in before merge, though the changelog fragment already carries the full account.

## What's good

- Verified locally at `f88b4351`: the 17 new tests pass, and the full `tests/simulation/isaac/` suite reads as a clean delta against the merge base `26078397` (+17 passed, same 3 pre-existing environment failures on both sides, none attributable to this diff).
- Changelog fragment in `changelog.d/` per AGENTS.md step 3, no direct `CHANGELOG.md` edit.
- MuJoCo enum comparisons in the new test are value-based (`int(model.geom_type[gid])` against `int(mjtGeom.*)` keys), per the AGENTS.md enum rule.
- Test names describe behaviour, docstrings are ASCII, no host paths, ruff clean on both changed files.
- Scope discipline: the diff is exactly the selection-rule change plus its pin; the deliberately-unchanged cases (all-contact-free body, single geom, mesh-only body) are pinned as controls.

PR #2743, opened 2026-08-26
fix(simulation): carry an unnamed actuator's command across a scene rebuild

cagataycali:
## Problem

`remove_robot` cannot delete a body attached through `spec.attach()` without tripping a MuJoCo
shutdown segfault, so `eject_robot_from_scene` rebuilds the scene spec from scratch. That compiles a
fresh model and allocates a fresh `MjData`, which renumbers every actuator id, so the dynamic state
is snapshotted under a name-based key beforehand and written back afterwards.

The joint half of that snapshot identifies an unnamed joint through its owning body plus its position
among that body's joints. The actuator half keyed on the actuator's own name alone and skipped
anything unnamed, on the grounds that *"its transmission target may drive several actuators, so it
cannot be matched across the rebuild"*. A skipped key is skipped silently, so such an actuator's
`ctrl` and its `act` activation were dropped and it came back at its fresh-compile zero. The
setpoints holding a **surviving** robot's pose reset because some **other** robot was removed, under
`status success`.

Reproduced through the public API on `main` (two robots, the furniture's two actuators unnamed):

```
add arm : success
add fur : success
nu = 3   names = ['arm/pan_act', None, None]
trnid    = [(0, -1), (1, -1), (1, -1)]     # both furniture actuators drive joint id 1
ctrl BEFORE : [0.0, 0.63, -0.42]
remove  : success
ctrl AFTER  : [0.0, 0.0]                   # both commands gone
```

An unnamed `<actuator>` child is the ordinary MJCF spelling, not a contrivance. Of the 235
actuator-bearing models in the downloaded MuJoCo Menagerie tree, 7 leave **every** one of theirs
unnamed - 43 unnamed actuators in total:

| model | actuators | unnamed |
|---|---|---|
| `ufactory_lite6/lite6.xml` | 6 | 6 |
| `google_robot/robot.xml` | 9 | 9 |
| `ufactory_lite6/lite6_gripper_narrow.xml` | 7 | 6 |
| `ufactory_lite6/lite6_gripper_wide.xml` | 7 | 6 |
| `iit_softfoot/scene.xml` | 1 | 1 |

So a scene holding one of those arms lost its whole command vector on every rebuild.

## Change

The premise was sound and the conclusion too strong, exactly as it was for joints: the target alone
does not single the actuator out, yet the target **plus its position among the actuators driving that
target** does. `_actuator_key` now returns one of two forms, and `_resolve_actuator_key` inverts it:

* `("actuator", name)` - a named actuator, the pre-existing handle.
* `("target", trntype, target_name, ordinal)` - an unnamed one.

Three properties make that key load-bearing rather than decorative, each measured:

* MuJoCo stores actuators in declaration order and an eject removes a robot's actuators **wholesale**,
  so the surviving order (and therefore the ordinal) is preserved. No scene op inserts an actuator
  into a compiled model - the patch vocabulary is `add_body` / `add_geom` / `add_site` /
  `set_body_pos` / `set_body_quat` / `delete_body` - so nothing shifts it.
* The transmission type belongs in the key because `mjTRN_JOINT` and `mjTRN_JOINTINPARENT` are
  different transmissions that can name the **same joint id**; measured on a fixture where they do.
* The ordinal counts every actuator sharing a target, named or not, so the snapshot and the restore
  count the same population. An actuator that gains a name is carried by its own `("actuator", name)`
  key, so the target handle explicitly declines to claim it.

`_actuator_target_kind` maps the `mjtTrn` member to the `mjtObj` its `trnid[0]` indexes, because each
transmission points at a different kind of element. An actuator references its target by name in
MJCF, so the target is never anonymous; a transmission the mapping does not know yields no key and is
reported rather than resolved against the wrong table.

## Visual

The physical consequence, in sim headless (EGL). A door is held at `0.90 rad` by a single **unnamed**
`<position>` servo. `remove_robot("arm")` used to leave that servo's `ctrl` at `0.0`, and 400 steps
later the servo had driven the door back to `0.17 rad`.

a scene rebuild keeps an unnamed actuator's command

| | door angle after 400 steps | unnamed actuator `ctrl` |
|---|---|---|
| before the eject | `0.9000 rad` | `0.9` |
| after, `main` | `0.1710 rad` | `0.0` |
| after, this branch | `0.9000 rad` | `0.9` |

The isolating comparison is panel 2 vs panel 3, both post-eject: 18061 px differ. Panels 1 vs 3
differ by 615 px, which is the arm the call removed. Lighting was read back from the model and
asserted identical across all three panels, and each panel is asserted to be embedded in the figure
byte-for-byte at its measured paste offset.

## Tests

`tests/simulation/mujoco/test_unnamed_actuator_command_survives_a_rebuild.py`, 24 cells in four
classes, named after the sibling file that pins the joint half.

Pre-fix split, with the helpers kept and only the two call sites reverted (a raw revert is a
collection error, since the tests import the new helpers):

| class | failing / total pre-fix |
|---|---|
| `TestAnUnnamedActuatorSurvivesTheEject` | **5 / 5** |
| `TestThePremisesTheKeyRestsOn` | 0 / 6 |
| `TestTheKeyVocabulary` | 0 / 6 |
| `TestWhatMustNotChange` | 0 / 5 |

Break table - 11 mutations, 9 detected, **9 distinct failing-id sets**, control clean, source
restored byte-identical, and **0 failures in the sibling joint suite on every row**:

| mutation | fires |
|---|---|
| the ordinal is always 0 | 5 |
| resolve ignores the transmission type | 3 |
| the gained-a-name guard is dropped | 1 |
| the ordinal counts only unnamed siblings | 5 |
| the mapping forgets the body transmission | 6 |
| a body target is looked up in the joint table | 5 |
| an ordinal past the end falls back to the last match | 1 |
| the snapshot stops reading the activation slice | 1 |
| the restore no longer writes `ctrl` | 5 |
| resolve accepts a vanished target | 0 |
| an unnamed target is keyed anyway | 0 |

The two zeros are reported with their reasons rather than dressed up. No actuator carries
`trnid[0] == -1`, so dropping the vanished-target guard leaves the loop finding nothing and returning
`-1` anyway - it is a fast path, not a correctness check. And an actuator references its target by
name in MJCF, so an anonymous target is unreachable through the format; that branch is
belt-and-braces for a hand-built model.

Two of these rows started at 0 and exposed real gaps in the fixture, which is why the fixture carries
a *named* actuator ahead of the unnamed ones on the same hinge (so an ordinal that skipped named
siblings desynchronises the two sides, 0 -> 5), and why the ordinal-bound mutation falls back to the
last match rather than merely widening `==` to `>=` (which is equivalent for any non-negative
ordinal, so it detected nothing).

The `mjtTrn` coverage check is derived from the enum, so a transmission added upstream fails this
file instead of silently falling into the reported-rather-than-guessed branch. It fired for real on
mujoco 3.12 - see the round 1 entry below.

## Gate

* `ruff check` + `ruff format --check` clean on both changed files.
* `mypy strands_robots tests tests_integ`: 24 errors, all pre-existing, **0 in either changed file**
  (the four-element key narrows on `len(key) == 4`, no `cast`, no `type: ignore`).
* `tests/simulation/` plus the docstring / export / except-tuple guards, 13773 selected with the new
  file deselected so the selection matches: **0 failures**, and 0 in the same subset of a pristine
  `main` full-suite baseline taken this session (39507 collected, 53 pre-existing failures, none in
  this subset). Both `comm` directions empty.
* The new file plus the four sibling eject / scene-state suites: **118 passed** on the declared
  mujoco floor, 3.5.0.
* `scripts/assemble_changelog.py --check` OK, and `tests/test_changelog_{fragments,format}.py` 30
  passed.

`_SceneState.actuators` is internal to `scene_ops`; its only consumer is `_restore_scene_state`, and
the only test touching it asserts the empty-dict case, so the key widening reaches no other caller.

The unnamed-actuator skip at `scene_ops.py:1743-1744` was among that module's unrun lines in the
baseline coverage report, which is how a whole behaviour sat ungraded.

### Round 1 - the enum-derived coverage check fired on mujoco 3.12

`call-test-lint` failed on
`TestTheKeyVocabulary::test_the_transmission_mapping_covers_every_transmission_mujoco_defines`:

```
>       assert unmapped == set()
E       assert {6} == set()
```

`6` is `mjTRN_SO3`, the orientation servo mujoco 3.12 added. The tripwire did exactly its job, so the
question it raises had to be answered rather than silenced.

**It cannot be answered by extending the mapping.** `mjTRN_SO3` is the one transmission whose type
does not fix which table its `trnid[0]` indexes. Measured on 3.12.0:

| spelling | `trntype` | `trnid` | `trnid[0]` names |
|---|---|---|---|
| `<general site="s" refsite="ref" gaintype="so3" .../>` | 6 | `[1, 0]` | site `s` |
| `<general joint="ball" gaintype="so3" .../>` | 6 | `[0, -1]` | joint `ball` |

Site names and joint names are separate namespaces, so a name read from the wrong table often still
resolves - onto an unrelated element. MuJoCo's own compiler says as much: *"so3 requires site or ball
joint transmission"*.

**Resolving it per actuator would be dead code today.** An orientation actuator takes three controls
(four on the quaternion chart), so a model carrying one compiles with `nu > nactuator`:

```
nu = 4 | nactuator = 2 | actuator_ctrladr = [0, 3] | actuator_ctrlnum = [3, 1]
```

This backend addresses an actuator by its control index, so `_unaddressable_actuator_reason` already
declines such a model - and `_recompile_preserving_state` declines it **before installing it**, which
is the only path that reaches `_actuator_key`. Carrying the target's `mjtObj` in the key is what tells
the two spellings apart, and it is worth adding on the day that refusal lifts, not before.

I implemented that wider key first and then reverted it: it was scope creep resting on a foundation
this backend deliberately refuses.

**So the mapping is unchanged - no production code in this round** - and the cell now pins both halves
of the argument instead of asserting a coverage claim that is not true:

* `test_the_only_transmission_the_mapping_leaves_out_is_the_orientation_servo` - the equality is still
  the tripwire, now against `{mjTRN_SO3}`, so a *new* transmission still fails here.
* `test_an_orientation_actuator_is_refused_before_it_could_need_a_key` - compiles the orientation
  model and asserts `nu > nactuator` and that `_unaddressable_actuator_reason` declines it. **This is
  the cell that fails if the refusal ever lifts**, which is when the key has to grow.

Both are skipped below the `mjTRN_SO3` floor (the manifest allows `mujoco>=3.5.0`), so the file still
passes on 3.5. The changelog fragment records the same reasoning.

Verified: 37 passed across this file and `test_multi_control_actuator_is_refused.py` on 3.12.0; `ruff
check`, `ruff format --check` and `mypy` clean on the changed file.

### DIFF
--- a/changelog.d/2743-unnamed-actuator-command-survives-a-rebuild.md
+++ b/changelog.d/2743-unnamed-actuator-command-survives-a-rebuild.md
@@ -39,3 +39,11 @@ upstream fails there rather than falling silently into the reported-rather-than-
 Deliberately unchanged: an actuator whose transmission is one that mapping does not know. Nothing
 then identifies its target across the rebuild, so it is reported rather than guessed at - the same
 residual the joint key leaves for a joint whose body is also unnamed.
+
+On mujoco 3.12 exactly one member of the enum sits in that residual, and it is unreachable rather
+than unhandled: `mjTRN_SO3`, the orientation servo, whose target is a site under one spelling and a
+ball joint under the other, so its transmission type alone does not fix which table names it. Such an
+actuator takes three controls, so a model carrying one compiles with `nu > nactuator` and this
+backend already declines it - it addresses an actuator by its control index - before any rebuild
+could snapshot it. Both halves of that argument are pinned by tests, so the key learns to carry its
+target's object type on the day that refusal lifts and not before.
--- a/changelog.d/2743-unnamed-actuator-command-survives-a-rebuild.md
+++ b/changelog.d/2743-unnamed-actuator-command-survives-a-rebuild.md
@@ -1,0 +1,41 @@
+### Fixed: a scene rebuild keeps an unnamed actuator's command instead of resetting it
+
+`remove_robot` cannot delete a body that was attached through `spec.attach()` without tripping a
+MuJoCo shutdown segfault, so `eject_robot_from_scene` rebuilds the scene spec from scratch. That
+compiles a fresh model and allocates a fresh `MjData`, which renumbers every actuator id, so the
+dynamic state is snapshotted under a name-based key beforehand and written back afterwards.
+
+The joint half of that snapshot identifies an unnamed joint through its owning body plus its position
+among that body's joints. The actuator half keyed on the actuator's own name alone and skipped
+anything unnamed, on the grounds that "its transmission target may drive several actuators, so it
+cannot be matched across the rebuild". A skipped key is skipped silently, so such an actuator's
+`ctrl` and its `act` activation were dropped and it came back at its fresh-compile zero - the
+setpoints holding a *surviving* robot's pose reset because some *other* robot was removed, under
+`status success`.
+
+The premise was sound and the conclusion too strong, exactly as it was for joints: the target alone
+does not single the actuator out, yet the target plus its position among the actuators driving that
+target does. MuJoCo stores actuators in declaration order and an eject removes a robot's actuators
+wholesale, so the surviving order is preserved, and no scene op inserts an actuator into a compiled
+model (the patch vocabulary is `add_body` / `add_geom` / `add_site` / `set_body_pos` /
+`set_body_quat` / `delete_body`). The transmission type belongs in the key alongside the target
+because `mjTRN_JOINT` and `mjTRN_JOINTINPARENT` are different transmissions that can name the same
+joint id, and the ordinal counts every actuator sharing a target rather than only the unnamed ones,
+so the snapshot and the restore count the same population.
+
+An unnamed `<actuator>` child is the ordinary MJCF spelling, not a contrivance: of the 235
+actuator-bearing models in the downloaded MuJoCo Menagerie tree, 7 leave every one of theirs unnamed
+- `ufactory_lite6` (6 of 6), `google_robot` (9 of 9) and `iit_softfoot` (1 of 1), 43 unnamed
+actuators in total - so a scene holding one of those arms lost its whole command vector on every
+rebuild.
+
+Measured through the public API on a scene of two robots, where a door is held at `0.90 rad` by a
+single unnamed `<position>` servo: `remove_robot("arm")` used to leave that servo's `ctrl` at `0.0`,
+and 400 steps later the door had been driven back to `0.17 rad`. It now reports `0.9`, and the door
+is still at `0.9000 rad`. An actuator's target is resolved through the `mjtTrn` member it drives
+through, and the mapping's coverage is derived from the enum in the tests, so a transmission added
+upstream fails there rather than falling silently into the reported-rather-than-guessed branch.
+
+Deliberately unchanged: an actuator whose transmission is one that mapping does not know. Nothing
+then identifies its target across the rebuild, so it is reported rather than guessed at - the same
+residual the joint key leaves for a joint whose body is also unnamed.
--- a/strands_robots/simulation/mujoco/scene_ops.py
+++ b/strands_robots/simulation/mujoco/scene_ops.py
@@ -1621,6 +1621,17 @@ def reposition_body_in_scene(
 # See :func:`_joint_key`.
 _JointKey = tuple[str, str] | tuple[str, str, int]
 
+# An actuator is keyed the same way, for the same reason. A named actuator takes
+# the two-element form ``("actuator", name)``.
+#
+# An unnamed actuator takes the four-element form
+# ``("target", trntype, target_name, ordinal)``: its transmission target names
+# it, and the ordinal disambiguates the several actuators that may share one
+# target. The transmission type belongs in the key because ``mjTRN_JOINT`` and
+# ``mjTRN_JOINTINPARENT`` are different transmissions on the same joint id.
+# See :func:`_actuator_key`.
+_ActuatorKey = tuple[str, str] | tuple[str, int, str, int]
+
 
 @dataclass(frozen=True)
 class _SceneState:
@@ -1640,9 +1651,10 @@ class _SceneState:
     Attributes:
         joints: ``key -> (qpos, qvel, qfrc_applied)`` slices, each at the width
             its joint type uses (free 7/6/6, ball 4/3/3, hinge or slide 1/1/1).
-        actuators: ``actuator name -> (ctrl, act)``. ``ctrl`` is the servo
-            setpoint holding a robot's pose; ``act`` is the internal activation
-            of a stateful actuator, its effective command.
+        actuators: ``key -> (ctrl, act)``. ``ctrl`` is the servo setpoint
+            holding a robot's pose; ``act`` is the internal activation of a
+            stateful actuator, its effective command. Keyed by
+            :func:`_actuator_key`, so an unnamed actuator is carried too.
         body_wrenches: ``body name -> [fx, fy, fz, tx, ty, tz]``, the external
             wrench :meth:`~strands_robots.simulation.mujoco.MuJoCoSimEngine.apply_force`
             latches in a body's own ``xfrc_applied`` row. ``qfrc_applied``
@@ -1655,7 +1667,7 @@ class _SceneState:
     """
 
     joints: dict[_JointKey, tuple[list[float], list[float], list[float]]]
-    actuators: dict[str, tuple[float, list[float]]]
+    actuators: dict[_ActuatorKey, tuple[float, list[float]]]
     body_wrenches: dict[str, list[float]]
     time: float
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Extends the eject-path snapshot/restore in `scene_ops.py` so an unnamed actuator's `ctrl`/`act` survives a scene rebuild, using the same identify-through-what-owns-it pattern the joint half already uses: named actuators keep their `("actuator", name)` key, and an unnamed one is keyed by `("target", trntype, target_name, ordinal)`. The key is injective within a model, the ordinal counts the same population on both sides (named siblings included), and the transmission type disambiguates `mjTRN_JOINT` vs `mjTRN_JOINTINPARENT` on the same joint id. The ordinal-stability premise holds structurally: namespacing means every actuator sharing a target belongs to one robot, an eject removes a robot's actuators wholesale, and `_snapshot_scene_state`/`_restore_scene_state` are called only inside a single `eject_robot_from_scene` invocation, so nothing can reorder the surviving population between the two reads. `_SceneState` stays module-private and in-memory (its keys are never serialised), so the key widening is not a one-way door. The residual case (a `mjtTrn` member the mapping does not know) is reported rather than guessed, matching the joint half, and the enum-derived coverage test turns a future upstream transmission into a loud failure.

## What's good

- Enum comparisons are value-based throughout (`int()` on both sides, dict keyed by `int(...)`), per AGENTS.md > "MuJoCo enums are matched by value".
- Every name lookup routes through `mj_name_to_id`, so a non-string key member cannot reach the SIGSEGV-prone binding.
- The restore's gained-a-name guard mirrors `_resolve_joint_key` exactly, so the two key vocabularies cannot drift in behaviour.
- Test names describe behaviour, the fixture pins its own premises (five unnamed actuators, shared target, named sibling ahead of the unnamed ones), and the changelog fragment is present under `changelog.d/`.

## Verification suggestions

- The new suite passes on the head SHA in a GL-free environment: `pytest tests/simulation/mujoco/test_unnamed_actuator_command_survives_a_rebuild.py` -> 22 passed. The four sibling eject/scene-state suites also pass except `test_multi_robot_eject.py::TestBodyMountedCameraSurvivesTheRebuild::test_the_camera_still_renders_after_the_rebuild`, which fails only where no EGL/OSMesa context exists (rendering-unavailable error, unrelated to this diff) - worth one run on a GL-capable box to confirm.

yinsong1986 (approved):
## Summary

Closes the unnamed-actuator gap in the eject-path snapshot/restore: `_snapshot_scene_state` used to skip any actuator without a name, so a surviving robot's `ctrl` setpoints and `act` activation silently reset to the fresh-compile zero whenever some *other* robot was removed. The fix mirrors the joint half's identify-through-what-owns-it pattern - a named actuator keeps its `("actuator", name)` key, and an unnamed one is keyed `("target", trntype, target_name, ordinal)`, with `_actuator_target_kind` mapping each `mjtTrn` member to the `mjtObj` table its `trnid[0]` indexes. The key is sound: the ordinal counts the same population (named siblings included) on both sides, the transmission type separates `mjTRN_JOINT` from `mjTRN_JOINTINPARENT` on one joint id, the gained-a-name guard mirrors `_resolve_joint_key`, and the ordinal-stability premise holds structurally since namespaced target names mean every actuator sharing a target belongs to one robot and an eject removes a robot's actuators wholesale. `_SceneState` stays module-private and in-memory, so the key widening is not a one-way door; an unknown transmission is reported rather than guessed, matching the joint half's residual. The head commit responds to the mujoco 3.12 CI failure by pinning both halves of the SO3 argument in tests rather than widening the mapping: the enum-derived tripwire now asserts `mjTRN_SO3` is the *only* unmapped member (so a future transmission still fails loudly), and a second cell asserts an orientation model is refused by `_unaddressable_actuator_reason` before any rebuild could snapshot it - the cell that fails on the day that refusal lifts and the key has to grow. Every name lookup routes through the SIGSEGV-safe `mj_name_to_id`, enum comparisons are value-based throughout, and the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention.

## What's good

- The break table in the PR description is honest about its two zero-detection rows and explains why each mutation is unobservable rather than padding the count.
- The fixture pins its own premises (`TestThePremisesTheKeyRestsOn`), including the named-sibling-ahead-of-unnamed arrangement that makes a skip-named-siblings ordinal desynchronise, closing a gap the mutation testing exposed.
- Test names describe behaviour, the suite is GL-free (`mesh=False`, no rendering), and the new cells are floor-safe: the SO3 cells skip below mujoco 3.12 while the tripwire equality degrades to the empty set, so the file passes on the declared 3.5.0 floor.

PR #2744, opened 2026-08-26
fix(mesh): a sensor record's identity is the publisher's, not the provider's

cagataycali:
## Problem

Every `SensorLoopsMixin._read_*` reader seeds a record with the keys this process decided, merges the robot's provider mapping over it, and publishes the result to a topic it builds from those same keys:

```python
summary: dict[str, Any] = {"peer_id": self.peer_id, "t": time.time()}
summary.update(data)          # <- the provider mapping, merged last
```

Merged last, a provider mapping carrying one of the seeded names replaces the local reading. Nine merge sites across seven readers were affected, so a reading published to `strands/{peer_id}/...` could name a different peer inside it, and a hand record published under one hand's name could name another.

Measured against the shipped readers, peer `alice`, provider mapping claiming `peer_id: "bob"`:

| reader | merge sites | `peer_id` reported |
|---|---|---|
| `_read_pose` (`_pose`, `_slam_pose`, `_odom_pose`) | 3 | `bob` |
| `_read_imu` | 1 | `bob` |
| `_read_odom` | 1 | `bob` |
| `_read_lidar_summary` | 1 | `bob` |
| `_read_lidar_state` | 1 | `bob` |
| `_read_map_info` | 1 | `bob` |
| `_read_hands` | 1 | `bob`, and `hand: "RIGHT"` under topic `.../hand/left/state` |
| `_read_health` | 0 | `alice` |

`peer_id` inside a payload is read as an identity elsewhere in the package -- `mesh/core.py` takes `issuer_id = data.get("peer_id")` and `sender = data.get("peer_id", "<remote>")`, and `mesh/audit.py` keys records by it.

## Why this is not a new rule

`docs/mesh.md` already states the precedence for the presence payload:

> `age` is *your* process's reading of when it last heard from a peer, and the `peer_id` a peer is filed under is the one its topic and certificate bind - not a field inside the payload. [...] those four locally decided keys - `peer_id`, `type`, `hostname`, `age` - win a name collision with it.

`PeerInfo.to_dict` implements it by spreading the peer's own payload **first**, and its docstring already names the cost of spreading it last:

> Spread last, a payload carrying any of those four names replaced the local reading. [...] a `peer_id` the sender chooses is the key `Mesh.peers_by_id` and `Mesh.get_peer` look the peer up by.

`_read_health` was already immune, for a different reason: it lifts named fields (`battery.get("pct")`) and namespaces the rest under its own key, so no provider name reaches the top level of its record. It is the in-file control for this change.

## Change

Re-assert the locally decided keys after each merge, through one helper (`_stamp_local_keys`) so the nine sites cannot drift apart. The readers that merge wholesale are **derived** from the shipped class rather than listed, so an eighth such reader is held to the same precedence instead of inheriting an exemption by omission.

Two boundaries are deliberately unchanged, and both are pinned:

- **`t` is not re-asserted.** It is a stamp rather than a locally computed duration -- the distinction `docs/mesh.md` already draws -- so a provider that stamps a reading when it decoded it reports something truer than the moment the loop got round to publishing it.
- **`source`/`frame` keep their `setdefault` seeding**, which exists precisely so a provider may name its own frame.

## Verification

Pre-fix split, source reverted to `upstream/main`: **11 failed / 26 passed**.

| class | failed | total | role |
|---|---|---|---|
| `TestAProviderCannotRenameThePeerAReadingCameFrom` | 10 | 10 | regression |
| `TestEveryWholesaleMergeReAssertsTheLocalKeys` | 1 | 3 | regression (structural, derived) |
| `TestReAssertingTheIdentityDoesNotCostTheReading` | 0 | 8 | over-reach guard |
| `TestAStampStaysTheReadingsOwn` | 0 | 9 | boundary control (`t`) |
| `TestAProviderStillNamesItsOwnFrameAndSource` | 0 | 3 | boundary control (`setdefault` labels) |
| `TestTheHealthReaderIsUnchanged` | 0 | 2 | control (immune sibling) |
| `TestTheInnerRobotObservationPathIsUnaffected` | 0 | 1 | control (non-merging branch) |
| `TestThePresencePathAlreadyResolvesThisCollision` | 0 | 1 | premise (the rule is the presence path's) |

Nothing in any control, boundary or premise class moves.

### Mutations

17 source mutations, all 17 detected, 14 distinct firing sets, two controls clean, source restored byte-identical after every row.

| mutation | fires | note |
|---|---|---|
| helper body dropped | 10 | |
| `**local` dropped, `peer_id` kept | 2 | the two hand cells |
| `peer_id` dropped, `**local` kept | 10 | same set as above -- both destroy the identity re-assertion |
| `setdefault` instead of assignment | 10 | the plausible-looking wrong fix: the provider already set the key, so it is a no-op |
| hands call drops `hand=name` | 2 | |
| each of the 8 single-record per-site calls removed | 2 each | own behavioural cell + the derived structural cell, 8 distinct sets |
| the hands per-site call removed entirely | 3 | both hand cells + the derived structural cell |
| **`t` also re-asserted** | 8 | over-reach: fires exactly the boundary-control class |
| **refuse instead of correct** | 18 | an error envelope is not a reading -- also costs the over-reach guard |
| raw revert | 11 | |

The derived inventory is load-bearing, not decorative. Planting an eighth wholesale-merging reader that forgets the rule:

| scenario | fires |
|---|---|
| control (no eighth reader) | 0 |
| eighth reader planted, **derived** scan | 2 |
| same plant, scan **hardcoded** to today's seven names | **0 -- silent** |

### Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1562 files).
- `mypy` **24 == 24** against a pristine `upstream/main` worktree (re-measured after the rebase), normalized message sets byte-identical both directions, 0 in the changed files.
- Paired run against the rebased base `442d9ca4`, over an identical 493-file selection (all of `tests/mesh/`, plus every guard that walks the tree or reads source, docstrings, `docs/` or `changelog.d/`), the new file excluded from both trees: **`21 failed, 19451 passed, 66 skipped, 24 errors` on both**, 45 identical failing/erroring ids, both `comm` directions empty. All 45 need `awsiot` (declared in the `[mesh-iot]` extra, absent here -- confirmed with `find_spec` + `tomllib`).
- `tests/mesh/test_zenoh_transport_security.py` run separately: 9 passed / 2 skipped on both trees.
- New file: 37 cases. With both pre-existing sensor suites and #2741's: **202 passed**.

### Composition with #2741

The safety-event fix in #2741 landed in the same file while this was in flight, so the branch is rebased on it and the two are verified to compose rather than merely to auto-merge. Reverting each side on the composed tree costs only its own cells:

| scenario | this PR's cells | #2741's cells |
|---|---|---|
| composed tree, nothing reverted | 0 | 0 |
| this PR's fix reverted | 11 | 0 |
| #2741's fix reverted | 0 | 10 |

The hunks are disjoint by construction: #2741 changes `publish_safety_event`, this changes the readers above it. `tests/mesh/test_safety_event_payload_reaches_the_wire.py` passes unchanged on this branch (part of the 202 above).

### Artifact

One robot (`peer_id: "alice"`) whose driver caches four readings and stamps each with a `peer_id` from its own config, captured by driving the shipped readers on both trees. The `health` row already agreed before the change -- it is the reader that lifts named fields instead of merging.

sensor record identity

### DIFF
--- a/changelog.d/2744-sensor-record-identity-is-the-publishers.md
+++ b/changelog.d/2744-sensor-record-identity-is-the-publishers.md
@@ -1,0 +1,37 @@
+### Fixed: a sensor record's identity is the publisher's, not the provider's
+
+Every `SensorLoopsMixin._read_*` reader seeds a record with the keys this process decided,
+merges the robot's provider mapping over it, and publishes the result to a topic it builds
+from those same keys. Merged last, a provider mapping carrying one of the seeded names
+replaced the local reading. Nine merge sites across seven readers were affected, so a
+reading published to `strands/{peer_id}/...` could name a different peer inside it, and a
+hand record published under one hand's name could name another:
+
+| published to | provider mapping said | record said |
+|---|---|---|
+| `strands/alice/imu` | `peer_id: bob` | `peer_id: bob` |
+| `strands/alice/pose` | `peer_id: bob` | `peer_id: bob` |
+| `strands/alice/lidar/summary` | `peer_id: bob` | `peer_id: bob` |
+| `strands/alice/hand/left/state` | `peer_id: bob`, `hand: RIGHT` | `peer_id: bob`, `hand: RIGHT` |
+| `strands/alice/health` | `peer_id: bob` | `peer_id: alice` |
+
+The precedence is not a new rule. `docs/mesh.md` already states it for the presence payload
+-- the locally decided keys win a name collision with what a peer reports, because "the
+`peer_id` a peer is filed under is the one its topic and certificate bind - not a field
+inside the payload" -- and `PeerInfo.to_dict` implements it by spreading the peer's own
+payload *first*, its docstring naming exactly what spreading it last costs: "a `peer_id`
+the sender chooses is the key `Mesh.peers_by_id` and `Mesh.get_peer` look the peer up by".
+The health reader was already immune for a different reason, visible as the last row above:
+it lifts named fields (`battery.get("pct")`) and namespaces the rest under its own key, so
+no provider name reaches the top level of its record.
+
+The locally decided keys are now re-asserted after each merge, through one helper so the
+nine sites cannot drift apart, and the readers that merge wholesale are derived from the
+shipped class rather than listed -- an eighth such reader added later is held to the same
+precedence instead of inheriting an exemption by omission.
+
+Two boundaries are deliberately unchanged. `t` is not re-asserted: it is a stamp rather
+than a locally computed duration, so a provider that stamps a reading when it decoded it
+reports something truer than the moment the loop got round to publishing it. The
+`source`/`frame` labels keep their `setdefault` seeding, which exists precisely so a
+provider may name its own frame.
--- a/docs/mesh.md
+++ b/docs/mesh.md
@@ -127,6 +127,16 @@ a loop. Every attempt, granted or refused, is written to the safety audit log.
 
 Sensor topics only publish when the robot exposes the attribute. Zero cost when unused.
 
+**A sensor record's identity is the publisher's too.** A reader seeds each record
+with what this process decided, merges the robot's provider mapping over it, and
+publishes to a topic built from those same keys - so the same precedence applies:
+`peer_id`, and the `hand` a hand record is filed under, win a name collision with
+the provider mapping. A provider naming another peer does not move its readings
+onto that peer's topic, and one naming another hand does not relabel the hand it
+was published under. A `t` the provider supplies *is* honoured: it is a stamp
+rather than a locally computed duration, so a driver that stamps a reading when
+it decoded it reports something truer than the moment the loop published it.
+
 ### Degraded state probes
 
 Every section of a state snapshot is optional, because a robot may be hardware,
--- a/strands_robots/mesh/sensors.py
+++ b/strands_robots/mesh/sensors.py
@@ -298,6 +298,40 @@ class SensorLoopsMixin:
             except Exception as exc:  # noqa: BLE001
                 logger.debug("[mesh] %s: pose tick error: %s", self.peer_id, exc)
 
+    def _stamp_local_keys(self, record: dict[str, Any], **local: Any) -> dict[str, Any]:
+        """Re-assert the keys this process decides, after a provider payload merged in.
+
+        A sensor reader seeds a record, merges the robot's provider mapping over
+        it and publishes the result to a topic it builds itself. Merged last, a
+        provider mapping carrying one of those seeded names replaces the local
+        reading -- so a record can be published to ``strands/{peer_id}/...``
+        while naming a different peer inside, and a hand record can be published
+        under one hand's name while naming another.
+
+        The presence path already resolves this collision the other way:
+        :meth:`strands_robots.mesh.session.PeerInfo.to_dict` spreads the peer's
+        own payload *first* so the four keys that process decided win. This is
+        the same precedence for the sensor records, applied after the merge
+        rather than by ordering a single literal, because a reader merges from
+        several provider attributes in turn.
+
+        ``t`` is deliberately not re-asserted. It is a stamp rather than a
+        locally computed duration, and a provider that stamps a reading when it
+        decoded it is reporting something truer than the moment the loop got
+        round to publishing it.
+
+        Args:
+            record: The outgoing record, modified in place.
+            **local: Further keys this reader decided, such as the ``hand`` a
+                hand record is published under.
+
+        Returns:
+            The same ``record``, for use as an expression.
+        """
+        record["peer_id"] = self.peer_id
+        record.update(local)
+        return record
+
     def _read_pose(self) -> dict[str, Any] | None:
         r = self.robot
         pose: dict[str, Any] = {"peer_id": self.peer_id, "t": time.time()}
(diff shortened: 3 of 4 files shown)


yinsong1986 (dismissed):
## Summary

This PR closes a spoofing gap in the mesh sensor loops: every `SensorLoopsMixin._read_*` reader seeds a record with `peer_id` (and `hand` for hand records), merges the robot's provider mapping over it, and publishes to a topic built from those same keys - so a provider mapping carrying `peer_id` could rename the peer a reading was filed under, contradicting the topic and certificate binding. The fix adds `_stamp_local_keys`, which re-asserts the locally decided keys after each of the nine wholesale-merge sites, matching the precedence `PeerInfo.to_dict` already applies on the presence path. The deliberate boundaries are sound and pinned: `t` stays the provider's (a decode-time stamp, not a locally computed duration - consistent with the AGENTS.md clock-domain rule), and `source`/`frame` remain `setdefault`-seeded so a provider may name its own.

## What's good

- The regression test derives the graded merge set from the shipped class via AST rather than a hand-copied list, so a reader added later inherits the rule (the AGENTS.md structural-pinning pattern), with a non-vacuity check on the derivation.
- Over-reach guards: the provider's own reading still reaches the record, `_read_health`'s structural exemption is asserted, and the hand-topic/record agreement is checked end to end.
- New helper carries a complete `Args:`/`Returns:` docstring; all strings ASCII; behaviour-named test classes; docs/mesh.md updated in the same change.
- Verified locally at head `5f6aa5b0`: the new suite passes (37 tests) and the adjacent sensor/pacing suites pass unchanged (117 passed, 1 skipped).

yinsong1986 (approved):
## Summary

This PR closes a spoofing gap in the mesh sensor loops: every `SensorLoopsMixin._read_*` reader seeds a record with the keys this process decided (`peer_id`, and `hand` for hand records), merges the robot's provider mapping over it, and publishes to a topic built from those same keys - so a provider mapping carrying `peer_id` could rename the peer a reading was filed under, contradicting the topic and certificate binding. The fix adds one helper, `_stamp_local_keys`, that re-asserts the locally decided keys after each of the nine wholesale-merge sites, matching the precedence `PeerInfo.to_dict` already applies on the presence path. The two boundaries deliberately left alone are sound and pinned: `t` stays the provider's (a decode-time stamp, not a locally computed duration), and `source`/`frame` keep their `setdefault` seeding so a provider may name its own frame.

## What's good

- The regression test derives the graded merge set from the shipped class via AST rather than a hand-copied list, with a non-vacuity check on the derivation and a structural (not listed) exemption for `_read_health` - the AGENTS.md structural-pinning pattern.
- Over-reach guards: the provider's own reading still reaches the record, the hand-topic/record agreement is checked end to end, and the presence-path precedence is measured rather than quoted.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; docs/mesh.md states the rule in the same change; new helper carries a complete `Args:`/`Returns:` docstring; all added strings ASCII; behaviour-named test classes.
- The branch was refreshed onto `main` since the previous (dismissed) review, absorbing #2741's `publish_safety_event` coercion cleanly - the PR's own diff is byte-identical across the rebase, and `behind_by` is now 0, so the tree CI tests is the merge result.

## Verification suggestions

- `python3 -m pytest tests/mesh/test_sensor_record_identity_is_not_the_providers.py -q` at the current head (`0bd1b05c`): 37 passed here, and the adjacent mesh sensor/pacing/safety-event suites (238 tests) pass unchanged; `ruff check` and `ruff format --check` are clean on both changed Python files.

yinsong1986 (approved):
## Summary

This PR closes an identity-spoofing gap in `SensorLoopsMixin`: every `_read_*` reader seeds a record with the keys this process decided (`peer_id`, and `hand` for hand records), merges the robot's provider mapping over it, and publishes to a topic built from those same keys - so a provider mapping carrying one of the seeded names could rename the peer (or hand) a reading was filed under, contradicting the topic and the certificate binding. The fix re-asserts the locally decided keys after each of the nine merge sites through one helper (`_stamp_local_keys`), applying the precedence `docs/mesh.md` already states for the presence payload and `PeerInfo.to_dict` already implements. Two boundaries are deliberately unchanged and pinned: a provider-supplied `t` is honoured (a decode-time stamp is truer than publish time), and the `source`/`frame` labels keep their `setdefault` seeding so a provider may name its own frame. The diff is purely additive (+388/-0) with no wire-format, schema, or public-API changes - it tightens payload contents toward what the topic and cert already assert.

## What's good

- The structural test derives the wholesale-merging readers from the shipped class via AST rather than listing them, so an eighth merging reader added later is held to the same precedence instead of inheriting an exemption by omission - and the PR demonstrates the hardcoded alternative is silent on exactly that plant.
- Boundary controls (`t` stamp, `setdefault` labels), the immune sibling (`_read_health`), the non-merging IMU fallback, and the presence-path premise are each pinned, so the fix's scope cannot silently widen.
- Composition with #2741 (same file, in flight together) is verified by reverting each side on the composed tree, not just by auto-merge.
- Changelog fragment and `docs/mesh.md` update ship in the same PR; the doc paragraph states the precedence and both deliberate boundaries.
- Verified locally on head `0bd1b05c`: the new file 37/37 passing, the neighboring sensor suites plus #2741's wire test 165/165 passing, `ruff check` and `ruff format --check` clean on the changed files, no non-ASCII in the diff, no host paths in tests.

PR #2745, opened 2026-08-26
fix(hardware): survive a teardown after the device bring-up failed

cagataycali:
## Problem

`Robot(..., mode="real")` builds its lerobot device in one statement:

```python
self.robot = self._initialize_robot(robot, cameras, **kwargs)
```

When that statement raises, the attribute it assigns never comes into existence. An absent
motor-SDK extra is the ordinary way to get there, and the message naming the extra is the whole
value of that path. `cleanup()` then runs from `__del__`, reaches `_disconnect_devices` ->
`_close_open_devices`, and both read the device handle as `getattr(self.robot, "bus", None)` -
which guards the *inner* attribute while dereferencing `self.robot` itself.

Measured on `Robot("koch", mode="real")` with the `dynamixel` extra absent:

```
ERROR strands_robots.hardware_robot:Cleanup error for koch: 'Robot' object has no attribute 'robot'
ImportError: 'dynamixel-sdk' is required but not installed. Install it with: pip install 'lerobot[dynamixel]'
```

The operator is shown a library-internal attribute name beside the install hint they actually
needed. After this change only the second line appears.

`_close_open_devices` is where this is most surprising, because its own docstring names "a failed
connect" as one of its two callers - it exists for a device set that is only *partly* open, and
could not survive the most partial failure there is.

## Change

Both readers bind the handle with `getattr(self, "robot", None)`, which is how the sibling teardown
reader `_shutdown_ros_bridge` already reads its own handle - the spelling applied consistently
rather than a new convention. Source change is +14/-4.

Reading tolerantly rather than skipping teardown when the handle is absent matters because
`_close_open_devices` has three callers besides `cleanup()`, on the failed-connect rollback path.
Row b9 below measures that: gating only the `cleanup()` call site still fails 3 cells.

`test_mesh_attrs_set_before_initialize_robot_no_attribute_error_in_cleanup` grades exactly this
failure, and **could not fail**, for three independent reasons:

| # | Reason | Measurement |
|---|---|---|
| 1 | Selected on `"AttributeError" in r.message` | `cleanup()` logs `f"Cleanup error for {n}: {e}"`, which interpolates `str(e)` - the message alone, never the type name. The record reads `Cleanup error for x: 'Robot' object has no attribute 'mesh'`, so the filter matched nothing whatever the code did |
| 2 | Asserted before the finalizer ran | The exception's traceback keeps the half-built instance alive. Records immediately after the raise: `[]`. After `gc.collect()`: `["Cleanup error for probe: 'Robot' object has no attribute 'robot'"]` |
| 3 | Also required `"mesh" in r.message` | Would exclude any other attribute even if 1 and 2 were fixed |

It passed while printing this defect on its own output line:

```
tests/test_robot_factory.py .  [100%]Cleanup error for x: 'Robot' object has no attribute 'robot'
====================== 1 passed, 135 deselected in 1.81s ======================
```

Reasons 1 and 2 are fixed in place, and the pair is graded rather than asserted: b6 and b7 below
differ only in whether the forced collection is present.

## Derived invariant

`__init__` claims its mesh attributes are set before the bring-up "so `cleanup()`/`__del__` never
see an `AttributeError` if construction fails partway through". That is a claim about *every*
attribute teardown reads, so it is now graded against the attributes teardown actually reads,
derived from the call closure rather than from a list.

Assignments are credited across the closure, because `self.<attr> = ...` creates state that
outlives the method doing it - that is how `TeleopMixin._ensure_teleop_state` protects the siblings
its caller invokes afterwards. Tolerant probes are credited only to the method that makes them,
because `getattr(self, name, default)` binds a local; crediting them closure-wide would let one
reader's guard excuse a sibling's bare dereference, which is this defect exactly. Rows b1 and b2
are what that distinction buys - each fires the derived cell on its own.

Deriving rather than listing found one real exposure (`robot`) and correctly exempted 10
lazily-initialised `_teleop_*` attributes; `stop_teleoperate` is excluded as an inherited method
rather than instance state, pinned by a non-vacuity cell.

## Verification

Pre-fix split (source reverted to `main`, tests kept): **4 failed / 12 passed**, with **0 of 5
premise cells** and **0 of 3 over-reach control cells** failing. The 4 are the two direct-call cells,
the `cleanup()` cell and the derived invariant.

Break table - 10 mutations, control clean, 8 distinct firing sets, 0 collection errors, sources
restored byte-identical:

| Row | Mutation | Fires |
|---|---|---|
| b0 | control (no mutation) | 0 |
| b1 | device-closer guard reverted to `getattr(self.robot, ...)` | 6 |
| b2 | disconnector guard reverted to `getattr(self.robot, ...)` | 5 |
| b3 | both reverted (the raw revert) | 6 |
| b4 | handle bound as `self.robot`, no guard at all | 6 |
| b5 | guard bound *after* the first read | 5 |
| b6 | mesh/peer_id hoisting reverted (the earlier fix) | 4, **including the pre-existing guard** |
| b7 | b6 **and** the forced collection removed | 3, pre-existing guard silent again |
| b8 | teleop lazy initializer dropped | 2 |
| b9 | teardown skipped when the handle is absent (the obvious alternative) | 3 |

`b1 | b2 == b3` exactly, so each reader is independently pinned.

Gate, byte-identical on both trees. Selection of 587 test files - everything naming the hardware
robot, the factory, teardown or teleop, plus every guard that walks the tree or reads docstrings -
identical on both trees with the new file excluded from each:

| | branch | base (`26078397`) |
|---|---|---|
| collected | 23355 | 23355 |
| result | 19 failed, 23250 passed, 86 skipped | 19 failed, 23250 passed, 86 skipped |
| wall | 911.49s | 926.23s |

Failing-ID sets identical, `comm` empty in both directions. All 19 are dependency-absent, verified
against `pyproject.toml` rather than assumed: 17 need `awsiot` (`awsiotsdk`/`awscrt`, the
`[mesh-iot]` extra), 2 need `device_connect_edge` (the `[device-connect]` extra).

`ruff check` and `ruff format --check` clean over 1560 files. mypy **24 == 24** against a pristine
`main` worktree, normalized message sets byte-identical in both directions, **0 errors in any
changed file**.

No visual artifact: this changes a teardown path and what it logs, not a policy, simulation,
rendering, recording or asset. The measured console output above is the artifact.

### DIFF
--- a/changelog.d/2745-hardware-teardown-survives-a-failed-bring-up.md
+++ b/changelog.d/2745-hardware-teardown-survives-a-failed-bring-up.md
@@ -1,0 +1,43 @@
+### Fixed: teardown after a device bring-up that failed reports nothing, instead of naming an internal attribute
+
+`Robot(..., mode="real")` builds its lerobot device in one statement,
+`self.robot = self._initialize_robot(...)`, so when that statement raises the attribute it assigns
+never comes into existence. An absent motor-SDK extra is the ordinary way to get there, and the
+message naming the extra is the whole value of that path. `cleanup()` then runs from `__del__`,
+reaches `_disconnect_devices` -> `_close_open_devices`, and both read the device handle as
+`getattr(self.robot, "bus", None)` - which guards the *inner* attribute while dereferencing
+`self.robot` itself. Teardown raised `AttributeError: 'Robot' object has no attribute 'robot'`,
+`cleanup()` caught it, and the operator was shown a library-internal attribute name beside the
+install hint they actually needed. Measured on `Robot("koch", mode="real")` with the `dynamixel`
+extra absent: one `ERROR Cleanup error for koch: 'Robot' object has no attribute 'robot'` ahead of
+the actionable `'dynamixel-sdk' is required but not installed`, and none after.
+
+`_close_open_devices` is the method this is most surprising in, because its own docstring names "a
+failed connect" as one of its two callers - it exists for a device set that is only partly open, and
+could not survive the most partial failure there is. Both readers now bind the handle with
+`getattr(self, "robot", None)`, which is how the sibling teardown reader `_shutdown_ros_bridge`
+already reads its own handle, so this is that spelling applied consistently rather than a new
+convention. Reading it tolerantly rather than skipping teardown when it is absent matters because
+`_close_open_devices` has three callers besides `cleanup()`, on the failed-connect rollback path;
+gating only the `cleanup()` call site leaves those three unprotected.
+
+The invariant `__init__` already claims - its mesh attributes are set before the bring-up "so
+`cleanup()`/`__del__` never see an `AttributeError` if construction fails partway through" - is now
+graded against the attributes teardown actually reads, derived from the call closure rather than
+from a list, so an attribute assigned after the bring-up and read during teardown fails whether or
+not anyone remembers to add it. Assignments are credited across the closure, because
+`self.<attr> = ...` creates state that outlives the method doing it and is how
+`TeleopMixin._ensure_teleop_state` protects the siblings its caller invokes afterwards; tolerant
+probes are credited only to the method that makes them, because `getattr(self, name, default)` binds
+a local, and crediting it closure-wide would let one reader's guard excuse a sibling's bare
+dereference - which is this defect exactly.
+
+The existing guard for this failure,
+`test_mesh_attrs_set_before_initialize_robot_no_attribute_error_in_cleanup`, could not fail. It
+selected records with `"AttributeError" in r.message`, and `cleanup()` logs
+`f"Cleanup error for {name}: {e}"`, which interpolates `str(e)` - the message alone, never the type
+name - so its `offenders` list was empty whatever the code did. It also asserted before the
+finalizer had run: the exception's traceback keeps the half-built instance alive, so `__del__` had
+not yet been reached when the `pytest.raises` block exited. It passed while printing this defect on
+its own output line. Both are fixed in place, and the pair is graded: reverting the mesh hoisting it
+guards now fails it, and reverting that hoisting with the forced collection removed does not.
--- a/strands_robots/hardware_robot.py
+++ b/strands_robots/hardware_robot.py
@@ -1355,7 +1355,13 @@ class Robot(TeleopMixin, AgentTool):
         devices from being closed, which is the failure mode lerobot's single
         unguarded disconnect loop has.
         """
-        bus = getattr(self.robot, "bus", None)
+        # ``getattr`` on ``self``, not on ``self.robot``: this also runs on
+        # the partial-construction path, where ``__init__`` raised in the very
+        # statement that assigns ``self.robot`` and so the attribute does not
+        # exist at all. Guarding the inner attribute cannot help there.
+        # ``_shutdown_ros_bridge`` already reads its own handle this way.
+        robot = getattr(self, "robot", None)
+        bus = getattr(robot, "bus", None)
         try:
             if bus is not None and getattr(bus, "is_connected", False):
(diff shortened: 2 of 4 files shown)


yinsong1986 (dismissed):
## Summary

Fixes a teardown crash on the partial-construction path: `Robot(..., mode="real")` assigns `self.robot` from the same statement that can raise (an absent motor-SDK extra being the ordinary cause), so `cleanup()` running from the factory rollback or `__del__` dereferenced an attribute that never existed. Both device readers (`_close_open_devices`, `_disconnect_devices`) now bind the handle with `getattr(self, "robot", None)` -- the spelling `_shutdown_ros_bridge` already uses -- so teardown after a failed bring-up reports nothing instead of logging an internal `AttributeError` ahead of the actionable install hint. The PR also repairs the existing guard in `test_robot_factory.py`, which could never fail (it selected on `"AttributeError" in r.message` where `cleanup()` interpolates only `str(e)`, and asserted before the traceback released the half-built instance for finalization), and adds a derived invariant test that grades every attribute the teardown closure reads against what `__init__` assigns before the bring-up statement.

## What's good

- Verified locally at the head SHA: all 152 tests in the two touched test files pass; with `strands_robots/hardware_robot.py` reverted to `main`, 6 of the new tests fail -- the regression pair holds in both directions, including the derived-invariant cell.
- Over-reach controls (`TestAPresentDeviceHandleIsStillClosed`) pin that a present, connected device is still disconnected through the same paths, so the tolerant read cannot silently skip a real close.
- The derived `TestNoTeardownReadOutlivesItsAssignment` closure survey credits assignments closure-wide but probes only locally, which is precisely the distinction this defect turned on (a guarded read in one method excusing a bare dereference in a sibling).
- Changelog fragment present per AGENTS.md step 3; lint and `ruff format --check` clean; no non-ASCII in added lines; no host paths; test names describe behaviour.

One non-blocking note: the PR body is still the placeholder ("full body follows once the gate lands") and carries no `Closes #N` link -- per AGENTS.md step 6, a closing keyword belongs in the body if this PR is meant to close a tracked issue.

yinsong1986 (approved):
## Summary

Fixes the partial-construction teardown path in `strands_robots/hardware_robot.py`: `__init__` assigns `self.robot` from the same statement that can raise (an absent motor-SDK extra being the ordinary cause), so `cleanup()` running from the factory rollback or `__del__` dereferenced an attribute that never existed and logged a library-internal `AttributeError` ahead of the actionable install hint. Both device readers (`_close_open_devices`, `_disconnect_devices`) now bind the handle with `getattr(self, "robot", None)` -- the spelling the sibling `_shutdown_ros_bridge` already uses -- and keep walking the rest of teardown rather than skipping it, which protects the three failed-connect rollback callers as well as `cleanup()`. The PR also repairs the existing guard in `test_robot_factory.py` that could never fail (it selected on the exception type name where `cleanup()` interpolates only `str(e)`, and asserted before the traceback released the half-built instance for finalization), and adds a derived-invariant test grading every attribute the teardown closure reads against what `__init__` assigns before the bring-up statement.

## What's good

- Verified at the head SHA: both touched test files pass (152 passed, 1 skipped); with `strands_robots/hardware_robot.py` reverted to the base commit, 6 of the new cells fail -- the regression pair holds in both directions, including the derived-invariant cell.
- Over-reach controls (`TestAPresentDeviceHandleIsStillClosed`) pin that a present, connected device is still disconnected through the same paths, so the tolerant read cannot silently skip a real close.
- The closure survey's asymmetry -- assignments credited closure-wide, tolerant probes credited only to the method that makes them -- is exactly the distinction this defect turned on.
- Changelog fragment present per AGENTS.md step 3; `ruff check` and `ruff format --check` clean on all changed files; added lines are ASCII-only; no host paths; test names describe behaviour.

PR #2004, opened 2026-08-07
fix(device_connect): refuse a Reachy Mini motion argument the robot cannot be commanded with

cagataycali:
`ReachyMiniDriver`'s three movement RPCs -- `look`, `antennas`, `body` -- are the only surface in the package that actuates a physical robot, and between them they carry **nine caller-supplied numbers** into a command dict handed to the active `HardwareLink`. Nothing downstream refused them.

Reachy Mini motion argument domain

## The defect

Both links serialize the command dict as JSON, and Python's encoder emits `nan` / `inf` as the bare tokens `NaN` / `Infinity`, **which RFC 8259 does not define**. So a non-finite argument was not rejected by the transport -- it produced a frame the daemon must either reject whole (losing a command the RPC had already reported as `success`) or parse leniently, handing a non-finite target to the servos. Python's own `json.loads("NaN")` returns `nan`, so the lenient half is not hypothetical.

Because `rpy_to_pose` is pure trigonometry with no domain of its own, the same unusable argument behaved differently depending on **which parameter of one signature** it landed in:

| `look(...)` | `pitch` / `roll` / `yaw` (angle) | `x` / `y` / `z` (offset) |
|---|---|---|
| `nan` | `success`, wire = `[[NaN, NaN, NaN, 0.0], ...]` | `success`, wire = `[..., NaN]` |
| `inf` | **raises** `ValueError: math domain error` from `math.cos`, past the `@rpc` | `success`, wire = `[..., Infinity]` |
| `True` | `success`, a silent one-degree command | `success`, a silent one-millimetre command |
| `'30'` | raises `TypeError: must be real number, not str` | raises `TypeError: unsupported operand type(s) for /: 'str' and 'int'` |

The `@rpc` wrapper logs and **re-raises**, so the raising rows genuinely escape a method whose contract is a result dict, with a message naming neither the RPC nor the parameter.

Measured over all nine parameters x seven unusable values (63 cells) on `6e80e109`:

| | `main` | this change |
|---|---|---|
| sent to the robot | **30** | 0 |
| ...of those, not RFC 8259 JSON | **21** | -- |
| raised past the `@rpc` contract | **33** | 0 |
| refused, naming the parameter | 0 | **63** |

## Why this belongs at the RPC

`strands_robots.utils.finite_number_error` is the existing shared domain for exactly this, and its docstring already names the failure:

[...]

PR #2746, opened 2026-08-26
fix(device-connect): refuse a Reachy Mini key prefix that is not one robot

cagataycali:
## Problem

`ReachyMiniDriver.prefix` is the Wireless variant's whole Zenoh namespace. It is interpolated verbatim into the three key expressions the robot lives on — `<prefix>/joint_positions` and `<prefix>/imu_data` in `ZenohLink.start`, and `<prefix>/command` in `ZenohLink.send_cmd` — and nothing downstream narrowed it. Its sibling in the same `__init__`, `api_port`, has been guarded since #2004's follow-up, with a comment giving the reason: the value is interpolated verbatim into targets the module builds itself, nothing downstream refuses it, and the constructor is the only point a caller can act on. `prefix` is the same shape, one parameter over, and one of the keys it builds actuates the robot.

Two unusable shapes reached the wire, failing in **opposite** directions. Measured against real `zenoh`, driving the shipped `ZenohLink`:

| `prefix` | `__init__` before | `start()` subscribes to | `send_cmd` publishes to |
|---|---|---|---|
| `"reachy_mini"` | success | `reachy_mini/joint_positions`, `…/imu_data` | `reachy_mini/command` |
| `"reachy_mini/robot_a"` | success | `reachy_mini/robot_a/…` | `reachy_mini/robot_a/command` |
| `"**"` | success | `**/joint_positions`, `**/imu_data` | **`**/command`** |
| `"*"` | success | `*/joint_positions` | **`*/command`** |
| `"reachy_mini/*"` | success | `reachy_mini/*/joint_positions` | **`reachy_mini/*/command`** |
| `"a$*"` | success | `a$*/joint_positions` | **`a$*/command`** |
| `""` | success | `ZError: Invalid Key Expr '/joint_positions'` | `ZError` |
| `"a//b"` | success | `ZError: Invalid Key Expr 'a//b/joint_positions'` | `ZError` |
| `"trailing/"` | success | `ZError` | `ZError` |
| `"a?b"` | success | `ZError` | `ZError` |
| `7` | success | `7/joint_positions` | `7/command` |
| `None` | success | `None/joint_positions` | **`None/command`** |

**The wildcards are the dangerous half, because Zenoh accepts them.** `zenoh.KeyExpr` admits `*`, `**`, `reachy_mini/*` and `a$*`, and `declare_publisher` accepts every one of them as a *publisher* key. So a wildcard prefix is not refused anywhere downstream — it just widens `<prefix>/command` from one robot's inbox into a match-any key. Over a live session, `session.put("reachy_mini/**/command", …)` was received by a subscriber on the concrete key `reachy_mini/robot_b/command`.

**The invalid half is refused, but far from the caller.** `""`, `a//b`, `trailing/` and `a?b` construct fine and then raise from inside the transport at `ZenohLink.start` — after `connect()` has already probed the daemon over REST and logged `Connected to Reachy Mini at <host> (Zenoh)`.

**A non-string is not refused at all.** `prefix=None` namespaces the robot under the literal segment `None`.

## What this looks like on two robots

One `body(yaw=90)` aimed at `robot_a`, published over a live Zenoh session with two subscribers standing for two daemons. Each head's `yaw_body` joint is set from the command that robot's own subscriber received, so the picture is the wire result rather than a re-derivation.

Two Reachy Minis, one command

An operator with two Minis reaches for `prefix="reachy_mini/*"` — the glob-like intuition for "my Minis". Before, that constructed without complaint, published to `reachy_mini/*/command`, and **both** robots received `body_yaw = 1.5708 rad`. After, it is refused by name, the operator writes `reachy_mini/robot_a`, and `robot_b`'s inbox stays empty.

The isolating measurement: both trees deliver the identical `1.5707963267948966 rad` to `robot_a`, and of the **12620** pixels that differ between the panels, **12598 fall on `robot_b` and 52 on `robot_a`** (the 52 are shadow and anti-aliasing coupling). Same camera and same headlight in both panels, asserted from the render's own state.

## Change

`_key_prefix_error` sends each `/`-separated segment through `strands_robots.mesh.security.validate_mesh_identifier`, the shared domain for one key-expression segment. That guard exists for precisely this hazard — its docstring records that an unvalidated segment "silently widens a point-to-point subscription into a match-any one", and `InputPublisher`'s call site adds "a wildcard or an extra `/` segment here publishes actuator data to a key no receiver named". `device_connect` already imports from `mesh.security` at two other sites, so this is the established route rather than new coupling.

The check sits beside the existing `api_port` guard, before `super().__init__()` and before any daemon is reached. `api_port` stays first, so its message is not displaced when both are unusable.

**The `/` stays legitimate.** A prefix is a *sequence* of identifiers, not one, so `reachy_mini/robot_a` and `reachy_mini/robot_b` — the way two Minis are kept apart — are unaffected. `dev/reachy` and `dev/r`, which the existing `ZenohLink` suite already drives, are still accepted. Validating the prefix as a *single* identifier would have refused all of these; that variant is in the break table below and fires 11 tests.

### The accepted domain is a strict subset of what Zenoh will carry

| direction | result |
|---|---|
| guard accepts, Zenoh refuses | **empty** — nothing admitted can reach the late-refusal path |
| Zenoh accepts, guard refuses | the 4 wildcards, whitespace/control characters, over-length segments, non-strings |

Pinned as a property rather than a case list, with a non-vacuity test that the wildcards really are valid Zenoh keys (otherwise the whole class would be trivially true).

## Tests

`tests/test_reachy_key_prefix_domain.py`, 55 cases, modelled on the `api_port` sibling suite.

Pre-fix split, run two ways:

| revert | result |
|---|---|
| raw (source at `main`) | **34 failed, 21 passed** |
| behavioural (helper kept, call site removed) | **30 failed, 25 passed** |
| fixed | **55 passed** |

**Zero failures in the 5 premise cells (`TestWhyTheConstructorOwnsTheDomain`) and zero in the 3 over-reach controls (`TestMultiSegmentNamespacingIsPreserved`) under both reverts** — the first pins what makes the constructor the right owner, the second pins what must not change. In `TestTheRefusalPrecedesAnyState`, the one cell that passes pre-fix is the ordering control (`api_port` is still reported first), as it should.

Break table, 10 mutations, **10 distinct failing-id sets, none undetected, control clean**, source restored byte-identical. The last column is the four pre-existing Reachy suites (229 tests), which fire on nothing — these behaviours were previously ungraded.

| mutation | new file | existing reachy suites |
|---|---|---|
| control (unmutated) | 0 | 0 |
| guard dropped | 30 | 0 |
| guard placed after `super().__init__()` | 1 | 0 |
| whole prefix validated, not per segment (over-reach) | 11 | 0 |
| split on `_` instead of `/` | 7 | 0 |
| non-string coerced with `str()` instead of refused | 5 | 0 |
| only the first segment validated | 7 | 0 |
| local regex copy instead of the shared domain | 4 | 0 |
| wildcards stripped before the check | 14 | 0 |
| helper always accepts | 31 | 0 |
| refusal drops the segment from the message | 4 | 0 |

## Gate

* `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1560 files).
* `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `main` worktree, normalized message sets byte-identical in both directions, **0 in the changed files**.
* Paired run of an identical 375-file selection (every test naming `device_connect`, the Reachy modules, `mesh.security`, or walking the tree / reading docstrings), the new file excluded from **both** trees: **19075 collected and `19 failed, 18990 passed, 66 skipped` on both**, failing-ID sets identical, `comm` empty both ways. 17 of the 19 need `awsiot` (the `mesh-iot` extra); the other 2 are pre-existing in `tests/test_device_connect_hardening.py`.
* Changed files plus the four pre-existing Reachy suites: **284 passed**.

### Round 1 - the live Zenoh cell could actuate real hardware

Review flagged that `test_a_wildcard_publish_reaches_a_concrete_subscriber` opened its session with
`zenoh.open(zenoh.Config())` - a peer with multicast scouting enabled. The key it publishes is what
makes that serious: `**` matches zero chunks, so `reachy_mini/**/command` intersects the concrete
`reachy_mini/command` inbox **this driver ships**. Any contributor running the unit suite with
`eclipse-zenoh` installed would emit an actuation-shaped payload onto the live command namespace of a
real Reachy Mini Wireless daemon sharing a multicast domain with the run - reproducing, from the test
suite, the widened publish this PR exists to prevent. It also left the assertion open to foreign
traffic on that key.

The session is now opened through `_isolated_session`, which can reach nothing but itself: peer mode,
multicast **and** gossip scouting disabled, no listen or connect endpoints, and a per-run `namespace`,
which prefixes every key it puts on the wire. That is the single-session form of what
`tests/mesh/test_zenoh_transport_security.py` already does wherever it needs a live session, and of
why `tests/conftest.py` sets `STRANDS_MESH=false`.

The namespace is applied instead of renaming the key expression, so the cell still asserts the real
shape - `reachy_mini/**/command` against `reachy_mini/robot_b/command`.

Verified on eclipse-zenoh 1.10.0, both halves:

```
isolated session's own subscriber : ['reachy_mini/**/command']
default-config 'daemon' session   : []
```

The second line is a stand-in for the hardware: a default-config session on the same host subscribed
to both `reachy_mini/command` and `reachy_mini/**`, which is what a real Mini's daemon exposes. It
receives nothing, while the isolated session still delivers the wildcard put to its own concrete
subscriber - so the property the cell proves is unchanged, at zero network exposure.

`TestWhyTheConstructorOwnsTheDomain`: 5 passed, 2 errored locally on a missing `device_connect_edge`
(the optional extra, installed in CI). `ruff check`, `ruff format --check` and `mypy` clean on the
changed file. No production code in this round.


yinsong1986 (commented):
## Summary

Adds a constructor-time domain guard on `ReachyMiniDriver.prefix`: each `/`-separated segment is routed through the shared `strands_robots.mesh.security.validate_mesh_identifier`, refusing Zenoh wildcards (which Zenoh accepts on a publisher key and which silently widen `<prefix>/command` into a match-any actuation key), the shapes Zenoh refuses only late inside the transport, and non-strings. The guard sits beside the existing `api_port` guard, before `super().__init__()` and before any daemon contact, with `api_port` reported first when both are unusable. Multi-segment prefixes (`reachy_mini/robot_a`) stay accepted, so multi-instance namespacing is preserved. I verified the guard against the shipped `validate_mesh_identifier` and the real `ZenohLink` key interpolation, and confirmed the two premises the PR rests on against real zenoh: `zenoh.KeyExpr("reachy_mini/**/command")` intersects the concrete `reachy_mini/command`, and a wildcard publish is delivered to a concrete subscriber. One blocker in the new test file, flagged inline.

## What's good

- Uses the shared mesh-identifier domain rather than a local charset copy, and pins that choice (`TestTheDomainIsTheSharedMeshIdentifier`), per the AGENTS.md value-domain-guard rule.
- The updated `Raises:` block names the new refusal (AGENTS.md > API Consistency).
- Changelog fragment under `changelog.d/` rather than an `Unreleased` append.
- The refusal-precedes-any-state tests (`DeviceDriver.__init__` never called, daemon never probed) pin the placement, not just the verdict.

tests/test_reachy_key_prefix_domain.py
```diff
@@ -0,0 +1,433 @@
...
+    def test_zenoh_accepts_a_wildcard_as_a_publisher_key(self, key):
+        """So a wildcard prefix is not refused downstream -- it just widens."""
+        zenoh = pytest.importorskip("zenoh")
+        assert str(zenoh.KeyExpr(key)) == key
+
+    def test_a_wildcard_publish_reaches_a_concrete_subscriber(self):
+        """A command published on a widened key lands in another Mini's inbox."""
+        zenoh = pytest.importorskip("zenoh")
+        received: list[str] = []
+        session = zenoh.open(zenoh.Config())
```
yinsong1986: [MUST FIX] `zenoh.open(zenoh.Config())` opens a **default-config** session: peer mode with multicast scouting enabled. I verified empirically that two independent default-config sessions on one host discover each other via scouting and route -- and that `reachy_mini/**/command` intersects the concrete `reachy_mini/command` (`**` matches zero chunks), which is the default command inbox this very driver ships. So this unit test, run by every contributor via `hatch run test` with `eclipse-zenoh` installed, publishes an actuation-shaped payload onto the live default command namespace of any real Reachy Mini Wireless daemon discoverable on the developer's LAN -- reproducing, from the test suite, the exact widened-publish incident this PR exists to prevent. It also makes the assertion `received == ["reachy_mini/**/command"]` flaky against foreign traffic. `tests/conftest.py` sets `STRANDS_MESH=false` precisely to keep the unit suite off Zenoh, and the established live-zenoh pattern in this repo (`tests/mesh/test_zenoh_transport_security.py::_new_config`) disables multicast scouting and pins loopback endpoints for exactly this reason.

This can't be deferred because the exposure is physical actuation on merge: the moment this lands, the default suite emits unowned commands onto real hardware inboxes whenever a Mini shares a multicast domain with a test run.

Fix: open the session with scouting disabled (`cfg.insert_json5("scouting/multicast/enabled", "false")` and gossip off, no endpoints) -- I verified a single isolated session still delivers the wildcard put to its own concrete subscriber, so the test proves the identical property with zero network exposure. Using a test-scoped namespace instead of the hardware's real `reachy_mini` prefix would be a worthwhile second layer.

cagataycali: Fixed in fc0fcb81. You are right on both counts, including the one that makes it urgent: `**` matches zero chunks, so the key this cell publishes intersects the concrete `reachy_mini/command` inbox the driver itself ships, and the exposure is physical actuation from a default `hatch run test`.

The session is now opened through a `_isolated_session` helper that can reach nothing but itself: peer mode, multicast **and** gossip scouting disabled, no listen or connect endpoints, plus a per-run `namespace` as the second layer you suggested (it prefixes every key on the wire, so it is applied without touching the assertion). I took the namespace route rather than renaming the key expression so the cell still asserts the real shape, `reachy_mini/**/command` against `reachy_mini/robot_b/command`.

Verified on eclipse-zenoh 1.10.0, both halves:

```
isolated session's own subscriber : ['reachy_mini/**/command']
default-config 'daemon' session   : []
```

The second line is a stand-in for the hardware: a default-config session on the same host subscribed to both `reachy_mini/command` and `reachy_mini/**`, which is what a real Mini's daemon exposes. It receives nothing, while the isolated session still delivers the wildcard put to its own concrete subscriber - so the property the cell exists to prove is unchanged and the assertion is no longer open to foreign traffic.

The helper's docstring records the `**`-matches-zero-chunks reasoning and points at `tests/mesh/test_zenoh_transport_security.py` and the `STRANDS_MESH=false` default in `tests/conftest.py` as the established precedent, so the next person adding a live-Zenoh cell here finds the pattern rather than the default config. No production code changed.

cagataycali:
## Ping — this closes #357's only real gap

Quick status while `call-test-lint` finishes (Analyze/Breaking/dependency-review/etc. all green):

- The bug fixed here (Zenoh treating `*` / `**` in a publisher key expression as accept-any) is what my harness-side triage of #357 landed on as the one actionable defect: items 1-3 of that issue already ship (`device_connect/reachy_mini_driver.py` is 549 loc doing exactly what the plan describes; the RPCs `look`/`antennas`/`body`/`playMove` are all there; `Robot("reachy_mini", mode="real")` already resolves). Only the dual-instance collision was genuinely broken.
- Fix routes each `/`-separated prefix segment through `mesh.security.validate_mesh_identifier` — the shared domain whose docstring literally names this hazard. Legitimate `reachy_mini/robot_a` vs `reachy_mini/robot_b` still validate; the wildcard variants are refused *before* Zenoh accepts them onto the wire.
- Gate: 55 new cases, 10-mutation break table (10 distinct firing sets, control clean), mypy 24 == 24 byte-identical, paired 375-file run identical on both trees (19075 collected, same 19 pre-existing failures). Visual artifact: two Reachy heads posed in MuJoCo from commands each subscriber received over real Zenoh (12598 of 12620 differing pixels land on `robot_b`).

`call-test-lint` on 34m19s pass at the last head. Requesting review — this is a small, gated fix, and it converts #357 from "blocked" to "done" on the harness side.

yinsong1986 (approved):
## Summary

Adds a constructor-time domain guard on `ReachyMiniDriver.prefix`, the Wireless variant's whole Zenoh namespace: each `/`-separated segment is routed through the shared `strands_robots.mesh.security.validate_mesh_identifier`, so the wildcard prefixes Zenoh happily accepts on a publisher key (`*`, `**`, `reachy_mini/*`, `a$*`) are refused at the call that named them instead of silently widening `<prefix>/command` into a match-any actuation key; the shapes Zenoh refuses only late inside the transport (empty segment, `//`, `?`) and non-strings are refused at the same point. The guard sits beside the existing `api_port` guard, before `super().__init__()` and before any daemon probe, and multi-segment namespacing (`reachy_mini/robot_a` / `reachy_mini/robot_b`) stays legitimate, so keeping two Minis apart is unaffected.

## What's good

- The guard reuses the shared per-segment domain rather than a local regex copy, exactly the pattern AGENTS.md prescribes for value-domain guards with a second caller, and `device_connect` already imports from `mesh.security` elsewhere.
- The `Raises:` block and the `prefix` `Args:` entry are updated alongside the behaviour, and the refusal names class, parameter and offending segment.
- The soundness property (guard-accepted domain is a strict subset of what Zenoh carries) is pinned with a non-vacuity check, and the mutation break table shows 10 distinct failing-ID sets with a clean control.
- The prior review's live-session exposure was resolved correctly at this head: `_isolated_session` disables multicast and gossip scouting, configures no endpoints, and namespaces every key per-run, while still asserting the real widened-key shape.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; error strings are plain ASCII; deadlines in the live cell are measured on `time.monotonic()`.

PR #2747, opened 2026-08-26
fix(tools): refuse a servo velocity that would reverse the command

cagataycali:
## Problem

`Goal_Velocity` (register 0x2E) is **sign-magnitude** on the Feetech STS/SMS
series: bit 15 carries the direction and bits 0-14 the magnitude. `serial_tool`
bounded the field to the two-byte maximum instead - keyed to the byte width
rather than to the register - so every magnitude from 32768 up was accepted, put
its own two bytes on the wire, and was read by the servo as a command in the
opposite direction.

Measured by driving `serial_tool(action="feetech_velocity", ...)` against a
recording port and decoding the emitted bytes as sign-magnitude:

| `velocity=` | tool verdict | bytes emitted | what the servo executes |
|---|---|---|---|
| 32767 | accepted | `2E FF 7F` | +32767, forward (largest true magnitude) |
| 32768 | accepted | `2E 00 80` | magnitude 0 - it **stops** |
| 40000 | accepted | `2E 40 9C` | **-7232, reverse** |
| 50000 | accepted | `2E 50 C3` | **-17232, reverse** |
| 65535 | accepted | `2E FF FF` | **-32767, full speed reverse** |

Each returned `status="success"` quoting the number supplied, so the report
described a command the servo never received. This is not the truncation the
other register bounds exist for: the caller's value reached the wire intact and
simply meant something else there.

The vendor semantics is not this change's invention. lerobot declares
`STS_SMS_SERIES_ENCODINGS_TABLE["Goal_Velocity"] = 15` and encodes through
`encode_sign_magnitude`, which raises `ValueError` for every magnitude in the
table above and accepts every one below it - the same split this bound now draws.

`Goal_Position` (0x2A) shares the encoding and needed **no** change: its 12-bit
ceiling of 4095 sits far below the direction bit, so a position the tool accepts
can never reach it. Its bound was already keyed to the register, which is
precisely what `velocity`'s was not - measured in the same run, `position=4096`
and `position=40000` were both already refused.

## Why every gate was silent

`tests/tools/test_serial_tool_numeric_domain.py` asserted the defect, twice:

- `test_the_reported_velocity_is_the_velocity_on_the_wire[65535]` asserted 65535
  reaches the wire as 65535. True at the byte level and false in meaning - and
  the class it sits in is named `TestARegisterFieldIsNeverSilentlyTruncated`.
- `test_every_register_bound_is_a_value_the_field_can_hold` graded each ceiling
  against `0xFFFF`, the byte width. `position` at 4095 satisfies that rule with
  enormous slack, so the check passed for both fields while only one of them was
  keyed to its register.

Both are corrected here rather than worked around: the wire case now uses 32767,
and the structural rule grades against the largest magnitude the direction bit
leaves free. The bound is derived from the bit rather than restating 32767, so
the two cannot drift.

## Tests

New `tests/tools/test_feetech_velocity_direction_bit.py` (37 cases). The vendor
sign-bit index is stated in the test file and checked against lerobot, rather
than read off the module under test, so the premises and the over-reach controls
grade the same register semantics on either tree.

Pre-fix split, source reverted with the tests as shipped: **7 failed / 126
passed**, with **0** in the premise class, **0** in either control class and
**0** in the vendor-agreement class. The 6 failures are the regression class; the
7th is the existing structural guard, failing cleanly with `assert 65535 <=
32767`. No `AttributeError` on the new constants - they are reached from one cell
only, the one asserting the module agrees with the vendor.

Break table, 9 mutations, control clean, source restored byte-identical:

| mutation | fires | notes |
|---|---|---|
| the fix reverted | 7 | the pre-fix set |
| `_MAX_MAGNITUDE` is the two-byte maximum | 7 | |
| `_DIRECTION_BIT` = 16 (outside the field) | 7 | |
| `_DIRECTION_BIT` = 14 (over-reach) | 4 | fires the accepted-magnitude controls in both files |
| `_MAX_MAGNITUDE` off by one (the bit itself allowed) | 4 | boundary |
| the reason no longer explains the direction bit | 1 | the message cell |
| the ceiling restates 32767 instead of keying it to the bit | 1 | behaviourally a no-op today; the drift is what is pinned |
| the sibling `position` ceiling widened to the byte width | 6 | fires the sibling control and the structural guard |
| the guard runs after the port is opened | 84 | 15 in the new file, 69 in the two existing files |

9 of 9 detected, 7 distinct firing sets. The three rows firing the identical 7
do so structurally - each makes the effective ceiling the two-byte maximum.

The over-reach row is the one worth reading: narrowing the direction bit by one
refuses 32767, a magnitude the register does carry, and 4 cells catch it - 3 in
the new file plus the accepted-magnitude case in the existing domain file. That
is the evidence this bound did not become stricter than the register.

## Gate

Identical 470-file selection (all of `tests/tools/`, plus every guard that walks
the tree, reads source or docstrings, or reads `docs/`), the new file excluded
from both trees: **22255 collected, `22 failed, 22125 passed, 108 skipped` on
both**, 22 identical failing ids, `comm` empty in both directions. All 22 are
pre-existing and dependency-shaped: 17 need `awsiot` (the `[mesh-iot]` extra), 2
are lerobot-from-source `encode()` drift, 1 is a leftover dataset in `/tmp`, 2
assert on emoji in rendered output.

`mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `main`
worktree, normalized message sets byte-identical in both directions, 0 in the
changed files. `ruff check` and `ruff format --check` clean over 1562 files.

## Scope

One bound, its reason, the two assertions that encoded the old one, and the docs
row that documented it. Deliberately not included:

- **Signed velocity.** The tool has no way to express direction today, and
  accepting a negative `velocity` to mean reverse would be a new capability
  rather than a correction. Bounding the magnitude is what makes the existing
  success message true, which is the stated job of these bounds.
- **A tighter, model-specific speed ceiling.** The largest magnitude the field
  carries is a property of the packet; what an STS3215 will actually track is a
  property of the servo. This module's bounds are documented as the former, and
  `_REGISTER_FIELDS` says so.
- `pose_tool` writes only `Goal_Position` and clamps to each motor's declared
  range before encoding, so it never reaches this bit and is untouched.

No visual artifact: this is an input-domain fix on a serial write, not a policy,
simulation, rendering, recording or asset change. The byte-and-decode table above
is the evidence, and it is reproducible without hardware - which is also why the
regression runs on an install with no lerobot and no servo attached.

### DIFF
--- a/changelog.d/2747-feetech-velocity-direction-bit.md
+++ b/changelog.d/2747-feetech-velocity-direction-bit.md
@@ -1,0 +1,24 @@
+### Fixed: a servo velocity that would reverse the command is refused
+
+`Goal_Velocity` (register 0x2E) is sign-magnitude on the Feetech STS/SMS series:
+bit 15 carries the direction and bits 0-14 the magnitude. `serial_tool` bounded
+the field to the two-byte maximum instead, keyed to the byte width rather than to
+the register, so every magnitude from 32768 up was accepted and put its own two
+bytes on the wire - where the servo read them as a command in the opposite
+direction. `velocity=65535` ran the joint at full speed the wrong way,
+`velocity=40000` ran it in reverse at magnitude 7232, and `velocity=32768` read as
+magnitude zero and stopped a servo the caller had just asked to run. Each returned
+`status="success"` quoting the number supplied, so the report described a command
+the servo never received.
+
+This is the same silent reinterpretation the sibling fields are bounded to
+prevent, and it is not a truncation: the caller's value reached the wire intact
+and simply meant something else there. The field is now bounded to the largest
+magnitude that leaves the direction bit clear, derived from that bit rather than
+restating a number, and the refusal quotes the reason so a caller can see why
+32767 is the ceiling.
+
+`Goal_Position` (0x2A) shares the encoding and needed no change: its 12-bit
+ceiling of 4095 already sits far below the direction bit, which is what a bound
+keyed to the register rather than to the byte width buys. `docs/hardware/tools.md`
+documented the old `[0, 65535]` domain and is corrected with it.
--- a/docs/hardware/tools.md
+++ b/docs/hardware/tools.md
@@ -82,13 +82,21 @@ bytes of the outgoing packet, so an out-of-range value was never rejected on the
 wire - it was truncated into a different, reachable command while the success
 message quoted the value the caller supplied. `position=70000` put 4464 on the
 wire and `position=-1` put 65535, the largest the two-byte field holds. Each
-field is therefore bounded before the port is opened:
+field is therefore bounded before the port is opened.
+
+Two of those registers are bounded by more than their byte width. `Goal_Position`
+and `Goal_Velocity` are sign-magnitude on the STS/SMS series - bit 15 carries the
+direction - so a magnitude reaching that bit is not truncated but *reinterpreted*:
+`velocity=65535` put those exact two bytes on the wire and the servo ran full
+speed in the opposite direction, and `velocity=32768` read as magnitude zero,
+stopping a servo the caller had just asked to run. `position` was already inside
+that limit at 4095; `velocity` now is too.
 
 | Option | Accepted | Why the bound is where it is |
 |--------|----------|------------------------------|
 | `motor_id` | integer in `[1, 254]` | the frame carries the ID in one byte, and `255` is the header value |
 | `position` | integer in `[0, 4095]` | `Goal_Position` is a 12-bit register - the same full scale the reported angle divides by |
-| `velocity` | integer in `[0, 65535]` | `Goal_Velocity` is written as two bytes |
+| `velocity` | integer in `[0, 32767]` | `Goal_Velocity` is sign-magnitude with bit 15 the direction bit, so a larger magnitude commands the opposite direction |
 | `baudrate` | positive integer | pyserial coerces rather than checks, so `2.7` opens the port at 2 baud |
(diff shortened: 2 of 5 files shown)


yinsong1986 (approved):
## Summary

Bounds `serial_tool`'s `velocity` option to the largest magnitude that leaves the STS/SMS sign-magnitude direction bit (bit 15) clear, replacing the old two-byte ceiling that let magnitudes 32768-65535 reach the wire intact and be executed by the servo as reverse (or stop) commands while the tool reported success. The ceiling is derived from `_DIRECTION_BIT` rather than restating 32767, the refusal quotes the reason, and the guard runs in `_option_error` ahead of the port ever being opened, so a refused value never touches the bus. The two existing assertions that encoded the old bound are corrected rather than deleted, and the vendor semantics is cross-checked against lerobot's `STS_SMS_SERIES_ENCODINGS_TABLE` / `encode_sign_magnitude` when lerobot is installed. Narrowing the accepted domain is a behavior change for callers passing 32768+, but those calls were silently executing the opposite command, and the new refusal is explicit and self-explanatory - the sanctioned no-silent-defaults fix shape.

## What's good

- Changelog fragment, `docs/hardware/tools.md` row, and module docstring all updated in the same PR; scope discipline is explicit (signed velocity and model-specific speed ceilings deliberately excluded, `pose_tool` correctly identified as unreachable).
- The new test file carries premise, regression, over-reach-control, sibling-control, and vendor-agreement classes, plus an AST pin that the ceiling stays derived from the bit rather than drifting to a literal - and the sign-magnitude decode is spelled out locally so the suite grades on an install with no lerobot.
- The pre-fix failure split and 9-mutation break table in the description are reproducible evidence, not assertion.

PR #632, opened 2026-06-24
docs(teleop): document Teleoperator factory + teleoperate() mixin

cagataycali:
## Summary

Documentation for the merged **teleoperation** feature (`Teleoperator()` factory + `attach_teleop()`/`teleoperate()` mixin). No code changes — docs only.

Rebased on fresh `strands-labs/robots@main`. `mkdocs build --strict` passes.

## Changes

| File | Change |
|------|--------|
| **`docs/hardware/teleoperation.md`** | **NEW** full page (240 lines) |
| `mkdocs.yml` | Add **Teleoperation** to the *Real Hardware* nav section (between Robot Control and Tools) |
| `docs/api-reference.md` | Add `Teleoperator(name, **kwargs)` to the `strands_robots` symbol table |
| `docs/hardware/robot-control.md` | New high-level **Teleoperation** section; reframe **Mesh teleop** as the transport primitives `teleoperate(publish=True)` builds on |
| `README.md` | **Teleoperation** quick-start under *Quick starts* (leader→follower, WASD rover, `map_fn`) |

## New page covers

- **`Teleoperator()` factory** — mirrors `Robot()`; kwargs forwarded + validated.
- **Available teleoperators table** — 17 types with their emitted action-key shapes.
- **Mixin API** — `attach_teleop` / `teleoperate` / `detach_teleop` / `stop_teleoperate` with every parameter (`name`, `method`, `map_fn`, `names`, `robot_name`, `hz`, `publish`, `block`, `duration`).
- **Action-key compatibility matrix** — which pairings are zero-config vs need a `map_fn` (the `.pos` ↔ `velocity` caveat).
- **10 recipes** — leader→follower, WASD Earth Rover, gamepad/phone→base, pre-built instance, cross-vocabulary `map_fn`, multi-device merge, bimanual, MuJoCo sim teleop, mesh publish, time-boxed/clean teardown.
- **Relation to mesh teleop** — local driver vs transport; `publish=True` composes them.

## Accuracy

Every snippet + signature verified against the merged source (`teleoperator.py`, `teleop_mixin.py`, hardware `Robot.send_action`) and the live LeRobot registry — not invented. Action-key shapes (`keyboard_rover` → `{linear_velocity, angular_velocity}`, `so101_leader` → `{motor}.pos`) confirmed against lerobot source.

## Test plan
- [x] `mkdocs build --strict` — clean (no broken nav/links)
- [x] All internal doc links resolve
- [x] Snippets match the merged public API

PR #2748, opened 2026-08-26
fix(docs): repair the bimanual teleop recipe and grade every real-mode example

cagataycali:
## Problem

`docs/hardware/teleoperation.md` carried the only bimanual real-hardware recipe in the
documentation, and none of its three lines could run as written.

```python
bi = Robot("bi_so", mode="real", left_port="/dev/ttyACM0", right_port="/dev/ttyACM1")
bi.attach_teleop("bi_so_leader", left_port="/dev/ttyACM2", right_port="/dev/ttyACM3")
bi.teleoperate()
```

Three independent errors, each measured by running the line verbatim on `main`:

| # | line | what happens |
|---|---|---|
| 1 | `Robot("bi_so", ...)` | `ValueError: Unknown robot 'bi_so' (resolved to 'bi_so')` - not a canonical name or an alias |
| 2 | `left_port` / `right_port` on the robot | `ValueError: Unknown kwarg(s) for robot_type='bi_so_follower': ['left_port', 'right_port']` |
| 3 | `left_port` / `right_port` on `attach_teleop` | `ValueError: Unknown kwarg(s) for teleop_type='bi_so_leader': ['left_port', 'right_port']` |

Errors 2 and 3 survive fixing error 1, which is what makes them independent: the registered
spelling is `bi_so_follower`, and supplying it only moves the reader on to the next refusal. A
bimanual device is two arms, so `BiSOFollowerConfig` accepts
`['calibration_dir', 'cameras', 'id', 'left_arm_config', 'right_arm_config']` - a per-arm config
pair, and **no single `port` at all**. `BiSOLeaderConfig` requires the same pair.

## Change

The recipe is corrected to the form that was verified by running it, and gains a sentence stating
why a bimanual device has no single `port` - the fact the old text got wrong.

```python
from lerobot.robots.so_follower import SOFollowerConfig
from lerobot.teleoperators.so_leader import SOLeaderConfig

bi = Robot(
    "bi_so_follower",
    mode="real",
    left_arm_config=SOFollowerConfig(port="/dev/ttyACM0"),
    right_arm_config=SOFollowerConfig(port="/dev/ttyACM1"),
)
bi.attach_teleop(
    "bi_so_leader",
    left_arm_config=SOLeaderConfig(port="/dev/ttyACM2"),
    right_arm_config=SOLeaderConfig(port="/dev/ttyACM3"),
)
bi.teleoperate()
```

Both calls were executed: the robot constructs, and `attach_teleop` attaches.

`tests/test_docs_python_examples_are_callable.py` already grades documented keywords - but against
**signatures**, and its `_accepted_keywords` deliberately returns `None` ("any keyword binds") for a
callee carrying `**kwargs`. `Robot()` ends in `**kwargs: Any`, so every keyword to every documented
`Robot(...)` call is outside what that module can see. That is correct for the question it asks.
Nothing graded the robot **name** at all.

So the two modules are complementary rather than overlapping: that one asks *would Python bind this
call*, this one asks *would the runtime accept these values for this robot*. A test pins the
`**kwargs` premise, so if either entry point ever loses it the complementarity claim is re-examined
rather than silently assumed.

The block arrived in one 240-line documentation commit (`1c434330`, #632) and was never executed.

## The guard

`tests/test_docs_real_mode_invocations.py` grades what the runtime enforces. `mode="real"` resolves
`hardware.lerobot_type` to a lerobot config dataclass, and a keyword is accepted only when that
dataclass declares it or it appears in the cross-robot forwarding allowlist - so the accepted set is
a property of *the named robot*, derived per call, not of the factory.

- **Split by dependency.** The name rule needs only the registry and always runs; the keyword rule
  needs lerobot to resolve the config and skips without it. The more important half is never
  silently skipped.
- **Derived, not listed.** The owned-keyword set comes from the two entry points' own signatures
  plus the allowlist, so a parameter added to either is covered without editing the test.
- **Negative examples excluded.** The section above teaches that `Robot()` refuses every `*_leader`
  name and prints the `ValueError` as its own output. Grading it would report the documentation's
  teaching point as a defect, so a block whose comment names an exception type is excluded, and
  `so101_leader` is pinned as absent from the graded set.

Sweep result over `docs/**/*.md` + `README.md`: **25** documented `mode="real"` calls across seven
robots (`bi_so_follower`, `earthrover_mini_plus`, `koch`, `lekiwi`, `so100`, `so101`, `unitree_g1`).
Pre-fix **1** offender - this one. The other 24 were already correct, including every `robot_ip` and
`port` spelling. Post-fix: 0 unregistered names, 0 rejected keywords, 0 ungradable calls.

## Non-vacuity

After this fix the corpus is clean and can no longer exercise a rejection, so a sweep that only ever
sees valid input cannot distinguish a working rule from one that accepts everything. Both rules are
therefore also graded on **constructed exemplars** - the old text in each case, alongside its
corrected form and a single-`port` robot - with a check that both verdicts still occur. Each rule
lives in one predicate shared by the sweep and the exemplars, so they cannot drift apart.

## Break table

Nine mutations, control clean, none undetected, **eight distinct firing sets**.

| mutation | fires | notable |
|---|---|---|
| control (no mutation) | 0 | `16 passed` |
| documentation reverted to `main` | 2 | name half + the per-arm premise |
| documentation half-fixed: right name, old `*_port` | 2 | **keyword half** - shows the two halves are independent |
| `_documents_a_refusal` always `False` | 2 | the documented refusal is graded as a defect |
| `_documents_a_refusal` always `True` | 3 | corpus floor + the exclusion test |
| forwarding allowlist dropped from the owned set | 2 | `robot_ip`/`port` become "rejected" |
| `_rejected_keywords` always accepts | 3 | the three keyword exemplars |
| extractor limited to `README.md` | 2 | corpus floor |
| name rule weakened to a truthy string | 3 | the three name exemplars |

The last two rows are the reason the exemplars exist: with a clean corpus, weakening either rule
fires nothing without them.

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1562 files).
- `mypy`: **24 == 24** against a pristine `main` worktree, normalised message sets byte-identical in
  both directions, 0 in the changed files.
- Paired run, identical 611-file selection (every test reading `docs/`/`README`, walking the tree, or
  naming the registry / robot factory / hardware wrapper), new file excluded from both trees:
  **`19 failed, 25489 passed, 194 skipped`** on both (952.61s vs 948.25s), 19 identical failing IDs,
  `comm` empty both directions. All 19 pre-existing and unrelated: 17 need `awsiot` (the `[mesh-iot]`
  extra, absent here), 2 are lerobot-from-source `encode()` drift.
- The paired run above was taken at base `26078397`. #2741/#2742 merged while it was running, so
  the branch was moved onto `442d9ca4` before opening; the two commits touch no file this PR
  touches, and lint, mypy, the changed-file suite and the sibling example grader were re-run
  green on the new base.

No visual artifact: this changes documentation and adds a test, with no policy, simulation,
rendering, recording or asset behaviour involved.

### DIFF
--- a/changelog.d/2748-documented-bimanual-teleop-recipe.md
+++ b/changelog.d/2748-documented-bimanual-teleop-recipe.md
@@ -1,0 +1,57 @@
+### Fixed: the documented bimanual teleop recipe runs as written
+
+`docs/hardware/teleoperation.md` carried the only bimanual real-hardware recipe in the
+documentation, and none of its three lines could run. It opened with
+`Robot("bi_so", mode="real", left_port=..., right_port=...)`, which fails twice over and
+independently: `bi_so` is neither a canonical registry name nor an alias, so `Robot()` refuses it
+with `Unknown robot 'bi_so'` before anything is built; and `left_port`/`right_port` are accepted by
+no robot in the registry. A bimanual device is two arms, so `BiSOFollowerConfig` declares a
+`left_arm_config`/`right_arm_config` pair of per-arm config objects and no single `port` at all.
+Supplying the registered spelling therefore only moves the reader on to the second error,
+`Unknown kwarg(s) for robot_type='bi_so_follower': ['left_port', 'right_port']`. The
+`attach_teleop("bi_so_leader", left_port=..., right_port=...)` line beneath it fails the same way
+for the same reason: `BiSOLeaderConfig` requires the same per-arm pair.
+
+The recipe is corrected to the form that was checked by running it - both arms constructed from
+`SOFollowerConfig`, both leaders from `SOLeaderConfig` - and gains a sentence saying why a bimanual
+device has no single `port`, which is the fact the old text got wrong.
+
+Nothing in the repository could have noticed. `tests/test_docs_python_examples_are_callable.py`
+already grades documented keywords, but it grades them against *signatures*, and its
+`_accepted_keywords` deliberately returns "any keyword binds" for a callee carrying `**kwargs`.
+`Robot()` ends in `**kwargs: Any`, so every keyword to every documented `Robot(...)` call was
+outside what that module can see - correctly, for the question it asks. Neither does anything grade
+the robot *name*. The block itself arrived in one 240-line documentation commit (`1c434330`, #632)
+and was never executed.
+
+`tests/test_docs_real_mode_invocations.py` closes both halves by grading what the runtime actually
+enforces rather than what the signature admits. `mode="real"` resolves the robot's
+`hardware.lerobot_type` to a lerobot config dataclass, and a keyword is accepted only when that
+dataclass declares it or it appears in the cross-robot forwarding allowlist - so the accepted set is
+a property of the named robot, not of the factory, and it is derived per call. The two halves are
+split by dependency: the name rule needs only the package registry and always runs, while the
+keyword rule needs lerobot to resolve the config and skips without it, so the more important half is
+never silently skipped. The owned-keyword set is derived from the two entry points' own signatures
+plus the allowlist rather than listed, so a parameter added to either is covered without editing the
+test.
+
+Across `docs/**/*.md` and `README.md` the sweep grades 25 documented `mode="real"` calls over seven
+robots, of which this was the one defect; the other 24 were already correct, including every
+`robot_ip` and `port` spelling.
+
+A block that documents a refusal is excluded rather than reported. The section directly above
+teaches that `Robot()` rejects every `*_leader` name and prints the `ValueError` as its own output,
+so grading it would report the documentation's teaching point as a defect; such a block is
+recognised by a comment line naming an exception type, and `so101_leader` is pinned as absent from
+the graded set.
+
+Both rules are additionally graded on constructed exemplars, because after this fix the
+documentation corpus is clean and can no longer exercise a rejection - a sweep that only ever sees
+valid input cannot distinguish a working rule from one that accepts everything. The old text is the
+flagged exemplar in each case, alongside its corrected form and a single-`port` robot, with a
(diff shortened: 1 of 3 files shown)


yinsong1986 (approved):
## Summary

Repairs the documented bimanual teleop recipe in `docs/hardware/teleoperation.md` (the old `Robot("bi_so", ..., left_port=..., right_port=...)` form failed three independent ways) and adds `tests/test_docs_real_mode_invocations.py`, which sweeps every documented `mode="real"` `Robot(...)` call for a registered robot name and runtime-accepted keywords. I independently verified the corrected recipe against the installed lerobot: `bi_so_follower` resolves in the registry, `BiSOFollowerConfig`/`BiSOLeaderConfig` declare exactly the `left_arm_config`/`right_arm_config` pair (and no `port`), and both import paths in the fence exist. All 16 tests in the new module pass at the head SHA, `ruff check` and `ruff format --check` are clean on the new file, no non-ASCII slipped in, and the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention.

## What's good

- The name rule and keyword rule are split by dependency, so the registry half never silently skips when lerobot is absent (`pytest.importorskip` gates only the keyword half).
- The accepted-keyword set is derived from the two entry points' signatures plus `_FORWARDABLE_KWARGS` rather than hand-listed, so it tracks the code.
- Constructed exemplars keep both rules non-vacuous now that the corpus is clean - the exact failure mode a sweep over valid-only input would otherwise hide - and both sweep and exemplars share one predicate per rule.
- The `**kwargs` premise of the sibling grader (`test_docs_python_examples_are_callable.py`) is pinned, so the complementarity claim is re-examined if either entry point loses it.
- Scope discipline: docs + test + fragment, nothing else touched.

PR #359, opened 2026-06-06
fix(sim): drive tendon-transmission actuators via joint name (closes #318)

cagataycali:
## Problem (closes #318)

`MuJoCoSimEngine.send_action` → `_apply_action_by_name` resolved a joint-name action key only to **direct** joint-transmission actuators (`actuator_trnid[ai, 0] == jnt_id`). **Tendon-transmission** actuators — most notably the Franka/Panda gripper, whose `split` actuator drives `finger_joint1`/`finger_joint2` through a tendon — were silently dropped. A gripper command keyed by the finger joint name never moved the fingers; the 7 arm joints (direct transmission) worked.

Impact: any VLA policy emitting a gripper channel (Cosmos 3 DROID `joint_pos`, GR00T, …) recorded grasp commands that had **no effect** on the observed state → pick/grasp tasks can't succeed in sim and recorded LeRobotDatasets are silently corrupted.

## Root cause

For the Panda gripper, `actuator8.trntype == TENDON` and `actuator_trnid[8, 0]` points at the **tendon** `arm/split`, not at `finger_joint1` (joint id 7). So the joint→actuator `trnid` match never fired, and there's no actuator literally named `finger_joint1` either → command dropped.

## Fix

Extend the joint→actuator fallback (`_actuator_for_joint`) to resolve **both** transmission types:
1. **JOINT / JOINTINPARENT** — existing direct match (unchanged).
2. **TENDON** — find tendons whose `JOINT` wrap entries include the joint, then the actuator driving that tendon.

Plus `_scale_ctrl_for_actuator`: a tendon gripper's ctrlrange is in tendon units (e.g. `[0, 255]`), not a finger position. A normalised `[0, 1]` open/close command (the VLA convention) is mapped onto the ctrlrange; a value clearly inside a wide range is passed through verbatim. **Direct JOINT actuators are never rescaled** — positions/torques pass through raw.

## Verification

End-to-end on `data_config="panda"` (mujoco 3.9.0):

```
before:  finger_joint1 stuck at ~0.0 the entire rollout
after:   OPEN(1.0)=0.0400  CLOSE(0.0)=0.0000  delta=0.0400   # full [0,0.04] sweep ✓
         arm joint1 cmd 0.5 -> ctrl 0.5 (unscaled, no regression) ✓
```

**10 new regression tests** (`tests/simulation/mujoco/test_tendon_actuator.py`) using a tiny synthetic tendon-gripper model — **no asset download**, fast, deterministic. ruff check + format clean.

[...]

PR #2750, opened 2026-08-26
feat(drivers): native Dynamixel Protocol 2.0 codec and driver skeleton

cagataycali:
The driver seam (#2734) is on `main` with no native driver registered for any Dynamixel robot; `Robot("koch", mode="real", driver="strands")` raises today because `get_native_driver_class` returns `None`. This ships the smallest package that removes that failure mode without lying about what works.

The pattern mirrors #2739 (G1 driver) but for a servo bus: same seam, same lazy-import discipline, same named-refusal envelope for deferred paths so a caller writes the same error-checking code either way.

## What lands

- **`strands_robots/drivers/dynamixel/protocol.py`** — Protocol 2.0 codec:
  - `build_packet(servo_id, instruction, params)` — unicast frame
  - `sync_write_packet(register, data_length, entries)` — broadcast frame
  - `parse_status_packet(frame)` — decoder with CRC-OK flag (never raises on bad CRC)
  - `checksum(frame)` — CRC-16 matching Robotis' SDK exactly (table verified byte-for-byte)
  - byte stuffing — the reserved `FF FF FD` run is escaped on the way out and unescaped on the way in, with the inserted byte counted in `LEN` and the CRC taken over the stuffed frame
  - `decode_model_number(params)` — little-endian decode of register 0's two bytes, returning the **number only**. Mapping a number to a model name is deliberately not here; see §13 and #2753.
  - Pure, no serial I/O, no SDK import at module load

- **`strands_robots/drivers/dynamixel/driver.py`** — `DynamixelDriver` satisfying `HardwareDriver`:
  - `tool_name`, `tool_type`, `tool_spec` (declares `status`/`sensors`/`stop` read-only verbs)
  - `stream()` yields exactly one event per tool call, matching G1's pattern
  - `get_status()` reports connection + config in the same envelope shape as G1's `get_status`
  - Registered for **koch**, **aloha**, **vx300s** (ViperX 300s), **wx250s** (WidowX 250s), **trossen_wxai**, **dynamixel_2r** via `_register_shipped_drivers`

- **`tests/drivers/test_dynamixel_driver.py`** — 68 tests, all pass on this head

## What does **not** land (each wants its own PR)

The #359 issue's own triage sequencing recommends this split — 4 unreviewed public-API decisions (bus + tools + calibration + telemetry) in one PR is a mistake to repeat.

- **Serial bus** — `open_port`, sync-read, hot-reconnect, latency budget. Landable without hardware; is its own reviewable change.
- **Model-name resolution** — number → `"XL430-W250"`. Needs a live servo to grade itself against; lands with the bus. Tracked in #2753.
- **Agent tool bundles** — `koch_tools()`, `viperx_tools()`, `widowx_tools()`, `aloha_tools()`.
- **Calibration file format** — `~/.robots/<robot>/<peer_id>/calibration.json`.
- **Mesh telemetry contract** — `robot/<peer_id>/joint_state` at 100Hz.

Every deferred write path (`send_action`, `start_task`, `run_policy`, `connect_eagerly`) returns a named refusal:

```python
{"status": "error", "content": [{"text": "send_action: not wired yet (issue #359 bus)"}]}
```

Same shape a successful envelope would use, so the mesh and the agent do not need a code change on the day the bus lands.

## Verification — codec matches `dynamixel_sdk` byte-for-byte

| operation | expected bytes | matches SDK? |
|---|---|---|
| WRITE LED to ID 1, value 1 | `fffffd0001060003410001cce6` | ✓ |
| PING ID 1 | `fffffd0001030001194e` | ✓ |
| SYNC_WRITE GOAL_POSITION for IDs 1,2 | `fffffd00fe11008374000400010000000002ff030000ef40` | ✓ |
| CRC (WRITE preamble) | `0xE6CC` | ✓ |
| CRC (PING preamble) | `0x4E19` | ✓ |
| byte stuffing (`addStuffing` + `updateCRC`, SDK 4.0.5) | 36000 differential cases, 1038 carrying a reserved run | 0 mismatches |

Reproducible with:

```python
from dynamixel_sdk.protocol2_packet_handler import Protocol2PacketHandler
h = Protocol2PacketHandler()
frame = bytearray([0xFF, 0xFF, 0xFD, 0x00, 0x01, 6, 0, 3, 0x41, 0x00, 0x01, 0, 0])
assert h.updateCRC(0, frame, len(frame) - 2) == 0xE6CC
```

CRC table matches Robotis' `dynamixel_sdk/protocol2_packet_handler.py` at every one of 256 entries.

Every value this module returns is checkable against a byte string. That is the module's contract, and it is why per-model hardware metadata does not live here (§13).

## Tests

- `tests/drivers/test_dynamixel_driver.py`: **68/68 passed** on this head
  - 29 protocol tests (framing, CRC, sync-write, status parsing, model-number endianness, control-table widths)
  - 11 byte-stuffing tests (golden vectors taken from `dynamixel_sdk` 4.0.5's own output)
  - 20 driver tests (surface protocol, construction, refusal envelopes, stream, cleanup idempotence, connect_eagerly)
  - 8 registration tests (every supported robot resolves; Feetech neighbours do not)
- `tests/test_driver_seam.py`: unchanged in outcome from base — the only failures are pre-existing lerobot/strands optional-dep `ImportError`s, identical with and without this branch (measured on both sides with `git stash`).
- `ruff check`, `ruff format --check` and `mypy` clean on `strands_robots/drivers/dynamixel/` and its tests.

**Fixture change in `tests/test_driver_seam.py`:** the isolation fixture reset from `dict(current_table)` to `{}`. Shipped-driver auto-registration is otherwise test-order-visible — the same change #2739 needed for the same reason.

## Registered robots (all present in `strands_robots/registry/robots.json`)

| canonical | description |
|---|---|
| `koch` | Koch v1.1 (aliases: `koch_follower`, `koch_v1.1`, `low_cost_robot_arm`) |
| `aloha` | ALOHA bimanual (aliases: `bi_so_follower`, `agibot_dual_arm`, ...) |
| `vx300s` | ViperX 300s (6DOF) |
| `wx250s` | WidowX 250s (6DOF) |
| `trossen_wxai` | Trossen WidowX AI Bimanual |
| `dynamixel_2r` | Dynamixel 2R Educational Arm |

---

### Round 1

**1. The model-name table is gone rather than corrected.**

Review feedback established that `MODEL_NUMBERS` was misassigned. I cross-checked **every** row against two independent sources that agree with each other — Robotis' `dynamixel-workbench` `dynamixel_item.h` and lerobot's `src/lerobot/motors/dynamixel/tables.py::MODEL_NUMBER_TABLE` — and **9 of 12 rows were wrong**. Only `311`, `321` and `1130` were correct. The `XM540-W270-R` / `XM540-W150-R` rows named no model at all: the T/R connector variants share one model number, and `1150`/`1180` belong to XH540-V150 and XW540-T140. The tests baked the same wrong pairs as expected values, so the suite round-tripped the error.

The table is **deleted, not retyped**. Correcting ten rows would fix the instance and leave the class:

- `decode_model_number` had no error path for a *wrong* name — it returned the misassigned string and the probe still looked successful. A Koch arm's XL430-W250 probed as an XL330-M077: ~7x less stall torque, different voltage envelope, on the exact models the six registered robots use.
- The only two tests over the table asserted its **shape** (upper-case, hyphenated, unique) and passed green against a 75%-wrong table. That is the ceiling for a pure codec module: a number is checkable against a byte string, a name only against a servo.
- Nothing consumed it. `MODEL_NUMBERS` and `decode_model_number` had zero non-test callers, and this PR ships no bus, so no code path in it can obtain a real register-0 read. The no-dead-code rule in AGENTS.md points the same way.

`49a219e4` drops `MODEL_NUMBERS` (and its `__all__` export) and narrows `decode_model_number` from `-> tuple[int, str | None]` to `-> int`. The little-endian decode is now pinned at both ends of the range (`0x0000`, `0xFFFF`) as well as `0x0424`/`0x04A6`/`0x0460`/`0x0137`, because a byte-swap is invisible on a payload whose two bytes happen to be equal — which the previous three-case parametrize could not have caught either.

Follow-up **#2753** records the full verified number→model table, both traps that produced this bug (the 1060/1190 transposition and the shared `-T`/`-R` number), and an acceptance criterion that any future table be graded against an independent oracle in CI rather than against literals copied from itself.

**2. `70e7f502` — news fragment renamed.** It was numbered `2748`, which is a different open PR; `changelog.d/README.md` asks for the PR's own number. No path collision (slugs differ), but the number no longer identified the change.

**3. `b2da48fb` — the reserved byte run is escaped.** Protocol 2.0 forbids `FF FF FD` inside a payload; the codec did none of the stuffing, so the module's byte-for-byte claim held only for payloads that happen not to contain the run. Reachable rather than theoretical: `GOAL_POSITION` is signed 32-bit little-endian, and `-131073` — exactly -32.0 turns at 4096 counts/rev — encodes to `FF FF FD FF`, the only value in the legal extended-position range that does. Verified differentially against `dynamixel_sdk` 4.0.5's `addStuffing`+`updateCRC`: 36000 cases, 0 mismatches. This commit also cleared the two mypy errors that were failing `call-test-lint` (`cleanup` is `-> None` so asserting on its result is a type error; `ToolUse` is a TypedDict requiring `name`/`toolUseId`/`input`).

The review thread is left open on purpose: the requested resolution was to correct the rows, and this delivers a deletion instead. Happy to restore the corrected table here if that is preferred.

---

**Refs**: private harness issue tracking native driver work (companion Feetech issue is the same pattern for STS3215).
**Depends on**: #2734 (driver seam, MERGED).
**Related**: #2739 (G1 CycloneDDS driver, APPROVED — this PR follows its exact pattern for a different transport). #2753 (model-name resolution, deferred to the bus).


strands_robots/drivers/dynamixel/protocol.py
```diff
@@ -0,0 +1,376 @@
...
+# Model number -> canonical model name. Read from register 0 during a probe.
+# The identifiers here are the ones the Robotis e-manual and dynamixel_sdk
+# use; a downstream translator maps them to lerobot's names if needed.
+#
+# Not every Dynamixel model is here - only the ones present on the robots
+# :issue:`359` lists. Adding one is a one-line change and is deliberately
+# not gated behind a class hierarchy.
+# ---------------------------------------------------------------------------
+MODEL_NUMBERS: Final[dict[int, str]] = {
+    1060: "XL330-M077",
```
yinsong1986: [MUST FIX] Most rows of `MODEL_NUMBERS` are misassigned relative to the Robotis e-manual this module cites as its reference. Two examples verified directly against the e-manual's `Model Number(0)` control-table entries: **XL430-W250 is 1060 (0x0424)** ([source](https://emanual.robotis.com/docs/en/dxl/x/xl430-w250/#model-number0)), but this table maps `1060 -> "XL330-M077"` and puts XL430-W250 at 1030; **XL330-M077 is 1190** ([source](https://emanual.robotis.com/docs/en/dxl/x/xl330-m077/), EEPROM table initial value), but this table maps `1190 -> "XL330-M288"` (which is actually 1200). The same pattern holds for the other X-series rows (per the e-manual: XM430-W350=1020, XM430-W210=1030, XM540-W270=1120, XM540-W150=1130, XH430-W210=1010, XH430-V210=1050); only the MX rows (311, 321) and 1130 appear correct. Note `tests/drivers/test_dynamixel_driver.py::test_decode_model_number` bakes the same wrong pairs (`b"\x24\x04", 1060, "XL330-M077"`) as expected values, so the suite round-trips the error rather than catching it - those literals need correcting in the same change.

Why this can't be deferred: `decode_model_number` has no error path for a *wrong* name - it silently returns the misassigned string for real hardware, so a Koch arm's XL430-W250 probes as an "XL330-M077" (a servo with ~7x less stall torque and a different voltage/current envelope). `MODEL_NUMBERS` is exported in `__all__` and this file's own tests state downstream translators pattern-match on these names, so the moment the #359 bus PR or any consumer keys behaviour (current limits, position ranges, model gating) off the detected name, this is silent hardware misidentification - the data-corruption class of bug, on the exact models the six registered robots use. Fixing the table is cheap now and a behaviour-visible contract change later.

Resolution: correct every row against each servo's e-manual `Model Number(0)` value (or `dynamixel_sdk`/Dynamixel Wizard on a live bus), fix the test literals to match, and drop or re-verify the `XM540-*-R` rows - per the e-manual the T/R connector variants share one model number, and 1150/1180 belong to XH540-V150 and XW540-T140 respectively.

yinsong1986 (commented):
## Summary

Ships a pure Protocol 2.0 codec (`build_packet`, `sync_write_packet`, `parse_status_packet`, `checksum`, `decode_model_number`) plus a `DynamixelDriver` stub satisfying `HardwareDriver`, registered for koch/aloha/vx300s/wx250s/trossen_wxai/dynamixel_2r so `Robot("koch", mode="real", driver="strands")` resolves instead of raising. Write paths return the same named-refusal envelope the G1 driver uses. I independently executed the codec against the PR's claimed reference frames: the WRITE-LED, PING and SYNC_WRITE packets and both CRCs reproduce the `dynamixel_sdk` bytes exactly, and the framing/length/CRC-coverage semantics match the Robotis Protocol 2.0 spec. However, verifying the `MODEL_NUMBERS` table against the Robotis e-manual (the source the module cites) shows most of its rows are misassigned, and the test suite bakes the same wrong pairs as expected values - flagged inline.

## What's good

- Codec output verified byte-for-byte against `dynamixel_sdk` reference frames (I re-derived them independently; all match).
- Scope discipline: bus/tools/calibration/telemetry each deferred to their own PR, with named refusal envelopes instead of silent stubs.
- Changelog fragment per AGENTS.md step 3; `tests/test_driver_seam.py` isolation fixture reset mirrors the #2739 precedent.
- `parse_status_packet` distinguishes "arrived but corrupt" (CRC flag) from "malformed shape" (raise) - the right contract for a retryable serial line.

## Verification suggestions

- Cross-check every `MODEL_NUMBERS` row against the e-manual control-table pages (each servo's `Model Number(0)` initial value), or against `dynamixel_sdk`/Dynamixel Wizard on a live bus, before the bus PR builds a probe on top of it.

cagataycali:
## Rebased on `main` (PR #2739 landed) + fixed both `call-test-lint` failures, `58c75b58`

### What was red

`call-test-lint` at `943f9390` failed because this branch's base predated the G1 driver PR (#2739) merging. Two blockers, both real:

1. **Rebase conflicts** on `strands_robots/drivers/__init__.py` and `tests/test_driver_seam.py` — both files were touched by #2739 to add the G1 shipped-driver registration and to widen the seam-test's isolation fixture docstring.
2. **Two mypy errors in this PR's own files** at the rebase point, that the reviewer of #2739 already flagged the shape of for other test files:
   - `tests/drivers/test_dynamixel_driver.py:390` `func-returns-value` — `assert driver.cleanup() is None` on a `None`-returning method.
   - `tests/drivers/test_dynamixel_driver.py:414` `typeddict-item` — `ToolUse` is missing its required `name`.

### What the merge does

Both files are now the union of #2739's edits and this branch's edits. In `drivers/__init__.py`:

- Dynamixel registers first (six robots: `aloha`, `dynamixel_2r`, `koch`, `trossen_wxai`, `vx300s`, `wx250s` per `SUPPORTED_ROBOTS`), because a broken registration order is easier to see when the smaller-footprint driver fails after the bigger one, not before it — the docstring the G1 PR wrote for this function said the same thing about its own guard.
- G1 registers second (`g1`, `unitree_g1`), same guarded pattern.
- Both swallow `ValueError` from `register_native_driver`'s double-registration refusal, so a caller who registered first wins.

Verified end-to-end:

```
list_native_drivers()  -> ['aloha', 'dynamixel_2r', 'koch', 'trossen_wxai', 'unitree_g1', 'vx300s', 'wx250s']
get_native_driver_class('g1')         -> G1Driver         (via robots.json canonicalization)
get_native_driver_class('unitree_g1') -> G1Driver
get_native_driver_class('koch')       -> DynamixelDriver
```

The seam-test fixture docstring was rewritten to name both sources of the empty-table dependence (Dynamixel + G1) rather than pick one.

### What the mypy fix does

- `test_cleanup_is_a_no_op` — call `driver.cleanup()` and let the type checker's own guarantee that a `None`-returning method returned `None` stand; a comment records why the assertion was dropped. The test still grades what it always graded (that `cleanup()` on a fresh driver does not raise).
- `_stream_once` — the `tool_use` dict is now annotated `ToolUse` and includes the required `name` key (`driver.tool_name`), which is the same shape every other driver test in the tree uses.

### Green headless on this head

- `pytest tests/test_driver_seam.py tests/drivers/test_dynamixel_driver.py tests/test_all_exports_are_statically_defined.py --no-cov -q`: **122 passed** (45 seam + 57 dynamixel + 20 exports)
- `mypy strands_robots/drivers/dynamixel tests/drivers/test_dynamixel_driver.py strands_robots/drivers/__init__.py`: **0 errors in this PR's files**. The 4 mypy errors on the whole tree are all pre-existing (`strands_robots/utils.py:975`, `simulation/ik.py:108`, `mesh/core.py:2250` and `:2264`) — same set that `main` reports.
- `ruff check` + `ruff format --check` on `strands_robots/drivers/dynamixel/` and `tests/drivers/test_dynamixel_driver.py`: **clean**

### What did not change

The Dynamixel Protocol 2.0 codec, the driver's public surface, the `SUPPORTED_ROBOTS` list, the wire-format tests, the refusal envelopes, and the stubbed motion path (`start_task`/`send_action` return `"not wired yet"`) are all identical to the last review head. This round is a rebase + a two-line-per-file test hygiene fix, nothing else.

Waiting on `call-test-lint`.

strands_robots/drivers/dynamixel/protocol.py
cagataycali: Confirmed, and worse than spot-checked: I cross-checked every row against two independent sources that agree with each other -- Robotis' own `dynamixel-workbench` `dynamixel_item.h` (`XL430_W250 1060`, `XL330_M077 1190`, `XL330_M288 1200`, `XM430_W350 1020`, `XM430_W210 1030`, `XM540_W270 1120`, `XH430_W210 1010`, `XH540_V150 1150`, `XW540_T140 1180`) and lerobot's `src/lerobot/motors/dynamixel/tables.py::MODEL_NUMBER_TABLE`. 9 of 12 rows were misassigned; only 311, 321 and 1130 were right, exactly as measured here. The `-R` rows name no model at all -- the T/R connector variants share one number.

I did **not** re-type the numbers, and since that is not the resolution requested I want to be explicit about why.

Correcting ten rows fixes this instance and leaves the class. The two tests over the table -- `test_every_model_name_is_uppercase_and_hyphenated` and `test_no_duplicate_names` -- asserted only its *shape*, and passed green against a table that was 75% wrong. That is not a gap in those two tests; it is the ceiling for this module. A number is checkable against a byte string; a name is only checkable against a servo. Re-typing would leave ten unverifiable hardware facts inside a module whose stated contract is byte-for-byte verifiability against `dynamixel_sdk`, guarded by nothing but my having been careful this time.

And nothing consumes it: `MODEL_NUMBERS` and `decode_model_number` have zero non-test callers in the tree, and this PR ships no bus, so no code path in it can obtain a real register-0 read. AGENTS.md's no-dead-code rule points the same way.

So `49a219e4` deletes the table and narrows the function to the part that *is* wire format:

```python
def decode_model_number(params: bytes) -> int:   # was -> tuple[int, str | None]
    ...
    return int.from_bytes(params, "little")
```

Register 0's little-endian decode stays and is now pinned at both ends of the range (`0x0000`, `0xFFFF`) alongside `0x0424`/`0x04A6`/`0x0460`/`0x0137`, because a byte-swap is invisible on a payload whose two bytes happen to be equal -- something the old three-case parametrize could not have caught either. The wrong-length refusal is unchanged.

Verification on this head: `tests/drivers/test_dynamixel_driver.py` 57/57 pass; `ruff check` and `ruff format --check` clean; `tests/test_driver_seam.py` is identical in outcome before and after (same 3 pre-existing optional-dep `ImportError`s on `lerobot`/`strands` in my bare venv, measured with `git stash` on both sides).

Name resolution lands with the bus (#359 scope 1), where a probe can grade itself against a live servo instead of against a literal copied from itself. Tracked in #2753 with the full verified table recorded, plus the two traps that caused this (1060/1190 transposition, and the `-T`/`-R` shared number), and an acceptance criterion that any future table be cross-checked against an independent oracle in CI rather than against its own literals.

If you would rather the corrected table land here anyway, say so and I will restore it with the verified values -- but shipping a table CI cannot grade is what produced this bug, and deletion is the smaller change that removes the failure mode instead of the instance.

cagataycali:
## Codec graded against `dynamixel_sdk` byte-for-byte, as the module invites. One gap found and fixed: byte stuffing.

The module docstring says "Every function here is verifiable against Robotis' `dynamixel_sdk` byte-for-byte" and `checksum` says "the table there is the reference and this implementation should be byte-identical". I took that literally: `pip install --no-deps --target <dir> dynamixel-sdk` (4.0.5, pure Python, no serial port needed) gives `Protocol2PacketHandler.addStuffing` + `updateCRC`, which is the exact pair its `txPacket` uses to put bytes on the wire. That makes a real oracle available on any host with no hardware.

### What verified

| check | cases | mismatches |
|---|---|---|
| `checksum` vs `updateCRC` | 4000 random blobs | **0** |
| `_CRC_TABLE` vs the SDK's literal table | 256 entries | **identical** |
| `build_packet` vs SDK, payloads with no reserved run | 3000 | **0** |
| the three golden vectors already in `TestProtocol` (WRITE-LED, PING, SYNC_WRITE) | 3 | **0** |

So the CRC, the framing, the length field and the stuff/CRC ordering are all correct, and the existing vectors are genuinely SDK output. That control column is what makes the one failure below meaningful rather than noise.

### The gap

Protocol 2.0 forbids the run `FF FF FD` inside a payload, because a servo watching the bus reads those three bytes as the start of the next packet. The protocol escapes it with an extra `0xFD`, counts the inserted byte in `LEN`, and takes the CRC over the *stuffed* frame. The codec did none of this in any of the three entry points, and `parse_status_packet` did not reverse it, so the byte-for-byte claim held only for payloads that happen not to contain the run.

```
build_packet(1, WRITE, b"\xff\xff\xfd")
  before  fffffd0001060003fffffd d36f      LEN=6
  SDK     fffffd0001070003fffffdfd 7cd1    LEN=7
```

### Why it is reachable, not theoretical

`GOAL_POSITION` is a signed 32-bit little-endian value. In extended-position (multi-turn) mode the legal range is `-1048575..1048575`, and I enumerated it: exactly **one** value encodes to bytes containing the run.

```
-131073 -> ff ff fd ff     (= -32.0 turns at 4096 counts/rev)
1 of 2097151 legal values
```

That is an ordinary place to drive a multi-turn joint, and being 1-in-2-million is what makes it nasty: an arm works for weeks and then drops one command, or desyncs the bus mid-frame so the *following* packet is corrupt too. It is also latent right now in the best possible way — `driver.py` does not import the codec yet (an AST scan finds no `build_packet`/`sync_write_packet` call sites), so nothing has shipped on top of it. Fixing it before the #359 bus PR wires `GOAL_POSITION` costs nothing; fixing it after means debugging a data-dependent fault on real servos.

### The fix

`_stuff` / `_unstuff` are the two halves. `build_packet` and `sync_write_packet` go through the first (before the CRC, so the CRC covers the stuffed frame); `parse_status_packet` checks the CRC on the bytes as received and then unescapes. Re-verified against the SDK:

| check | cases | mismatches |
|---|---|---|
| `build_packet` (75% adversarial alphabet) | 20000, **1038** carrying a run | **0** |
| `sync_write_packet` | 8000 | **0** |
| status round-trip through a servo-stuffed frame | 8000, **325** carrying an escape | **0** |

```
sync_write(GOAL_POSITION, id=1, -131073)
  ours fffffd00fe0d00837400040001fffffdfdff6084
  SDK  fffffd00fe0d00837400040001fffffdfdff6084
```

12 tests in a new `TestByteStuffing`. Every expected value is SDK output, frozen as a literal rather than recomputed, because `dynamixel-sdk` is deliberately not a declared dependency — I checked `pyproject.toml` — and owning the codec is precisely so the wire format is gradeable without it. That matches how `TestProtocol` already records its vectors.

### Break table

11 mutations, control clean, source restored byte-identical.

| mutation | new-class cells | other cells |
|---|---|---|
| control (unmodified) | 0 | 0 |
| `_stuff` is a no-op (the pre-fix behaviour) | **6** | 0 |
| escape inserted but `LEN` not bumped | 6 | 0 |
| `parse_status_packet` no longer unescapes | 2 | 0 |
| stuffing applied *after* the CRC | 4 | 0 |
| escape byte is `0xFF` instead of `0xFD` | 6 | 0 |
| only `build_packet` stuffs, `sync_write_packet` does not | **1** | 0 |
| scan starts at the ID byte (over-reach) | 2 | **2 pre-existing** |

The over-reach row tripping two of *your* existing cells is the evidence the change does not widen what gets escaped. Two controls in the new class assert the ordinary no-run vectors are byte-identical with `LEN` unmoved, so a stuffing change that corrupts normal traffic fails in the class that caused it.

One row fires **0** and I want to be explicit that it is not a hole: rewriting the look-back to read the emitted output instead of the original frame changes nothing. I proved it exhaustively (5461 payloads, 0 disagreements) — an escape is always `0xFD`, so after escaping at some position the preceding pair is `FD FF` in the original and `FD FD` in the output, and neither is `FF FF`. My first docstring claimed the distinction "matters"; that was wrong and I corrected it to say it is a readability choice, spelled the SDK's way because that is what a reader will compare against.

### Also fixed: the two mypy errors that fail the lint gate

Measured on three trees: `upstream/main` **24**, this branch before my commit **26**, after **24** with byte-identical normalized message sets against main (both `comm` directions empty). The two were `assert driver.cleanup() is None` (declared `-> None`, so asserting on the result is a type error — calling it twice is the property worth holding) and a bare dict passed to `stream`, where `ToolUse` is a `TypedDict` requiring `name`/`toolUseId`/`input`.

### Gate

Identical 99-file selection on both trees (all of `tests/drivers/`, the seam tests, and every guard that walks the tree, reads docstrings or reads source), with the changed test file excluded from each so the selection is identical in content: **4004 collected, `1 failed, 3949 passed, 54 skipped` on both**, same single failing ID — `test_dataset_recorder.py::test_create_passes_vcodec_directly_when_supported`, a stale `/tmp/ds` dataset left by another process, environmental and identical on both sides. `tests/drivers/ tests/test_driver_seam.py tests/test_changelog_fragments.py`: **186 passed**. ruff clean over 1570 files.

Rebased onto `70e7f502` so it sits on top of the `MODEL_NUMBERS` removal — dropping the table rather than correcting it was the right answer to that thread, and it means the only name a caller can now act on comes from hardware.

cagataycali:
## Independent verification of the codec against `dynamixel_sdk` 4.0.5

The module docstring says every function here is verifiable against Robotis'
`dynamixel_sdk` byte-for-byte, and the `TestByteStuffing` docstring says the
codec "is expected to be byte-identical to it". I installed the SDK
(`pip install --no-deps dynamixel-sdk`, 4.0.5) and ran that comparison as a
differential fuzz rather than against frozen vectors.

Headline: **the codec is correct.** Five of its six functions are byte-identical
to the SDK over thousands of vectors, and the one that is not is not
byte-identical because the SDK is wrong there.

| comparison | vectors | mismatches |
|---|---|---|
| `checksum` vs `Protocol2PacketHandler.updateCRC` | 4000 | 0 |
| `_stuff` vs `addStuffing` | 6000 | 0 |
| `build_packet` -> `_unstuff` (self round trip) | 6000 | 0 |
| `_unstuff` vs `removeStuffing` | 6000 | **11** |

### Adjudicating the 11: `removeStuffing` loses a data byte

Both sides stuff identically, so I ran each side's own stuff/unstuff pair and
asked which one recovers the payload. Using the SDK's `addStuffing` for the
SDK's round trip, so no code from this PR is in its loop:

| round trip | lossy on |
|---|---|
| `_stuff` / `_unstuff` (this PR) | 0 / 6000 |
| `addStuffing` / `removeStuffing` (SDK) | **17 / 6000** |

An escape/unescape pair that is not the identity is broken independently of how
one reads the spec, so this settles it in the PR's favour.

Shortest disagreeing payload, found by exhaustive search over `{00, FF, FD}`:

```
params            ff ff fd ff fd fd            (6 bytes)
unstuffed frame   fffffd0003090001fffffdfffdfd   LEN=9
stuffed (both, identical)  fffffd00030a0001fffffdfdfffdfd   LEN=10

expected INST+params   03fffffdfffdfd
_unstuff (this PR)     03fffffdfffdfd    EXACT
removeStuffing (SDK)   03fffffdfffd      LOSSY, 1 byte, and reports LEN=8
```

Mechanism, from instrumenting the SDK's loop. `removeStuffing` compacts the
packet in place, so its three-byte look-back reads bytes it has already
overwritten:

```
i=3  cur=fd nxt=fd  lookback_now=(ff,ff)  lookback_on_wire=(ff,ff)  match=True  aliased=False
i=6  cur=fd nxt=fd  lookback_now=(ff,ff)  lookback_on_wire=(ff,fd)  match=True  aliased=True
```

At `i=6` the wire carries `ff fd`, which is not a reserved run. The compaction
has shifted an `0xFF` into that slot, so the SDK sees `ff ff`, treats a data
byte as an escape and deletes it. `addStuffing` writes into a separate `temp`
buffer and so has no equivalent problem, which is exactly why the stuffing
column above is 0 and the unstuffing column is not.

### What is worth changing

Nothing about the behaviour. Two docstring sentences, and one test.

1. `_unstuff` is documented as "the inverse of `_stuff`, and the mirror of
   Robotis' `removeStuffing`". The first half is true and the second is not: it
   is a corrected version. As written it reads as an invitation to make them
   match. (The corresponding sentence in `_stuff`, that its look-back
   formulation "is a readability choice, not a correctness one", is correct and
   I confirmed it at 0/6000.)

2. The module docstring's "every function here is verifiable against
   `dynamixel_sdk` byte-for-byte" is true for five of six. Worth narrowing,
   since it is the sentence that would send someone to align the sixth.

3. Nothing currently pins the divergence. The adversarial corpus in
   `test_stuffing_round_trips_every_adversarial_payload` is four bytes wide,
   and the case needs six:

| params width | payloads | `_unstuff` vs `removeStuffing` divergences |
|---|---|---|
| 4 (the corpus as written) | 625 | 0 |
| 5 (its status variant, `b"\x00" + params`) | 3125 | 0 |
| 6 | 15625 | 1, `fffffdfffdfd` |
| 7 | 78125 | 15 |

So I replaced `_unstuff` with a faithful port of `removeStuffing` (verified
first that the port reproduces the real SDK's output byte-for-byte on the
minimal case, and is lossy against the expected payload) and ran the suite:

| tree | `test_dynamixel_driver.py` | `tests/drivers/` + `tests/test_driver_seam.py` |
|---|---|---|
| as shipped | 0 failed, 68 passed | 0 failed, 160 passed |
| `_unstuff` aligned to the SDK | **0 failed**, 68 passed | **0 failed**, 160 passed |

An edit that follows the module's own instruction reintroduces silent
single-byte data loss and the whole suite stays green.

### Suggested pin

Verified both ways against this branch: passes as shipped, fails on the
SDK-aligned mutation, and the premise cell stays green in both.

| tree | result |
|---|---|
| as shipped | 0 failed, 71 passed |
| `_unstuff` aligned to the SDK | 2 failed, 69 passed (`test_the_aliasing_payload_round_trips`, `test_six_byte_payloads_round_trip`) |

```python
class TestUnstuffingIsNotTheSdkImplementation:
    """The unescape deliberately diverges from Robotis' ``removeStuffing``.

    ``removeStuffing`` compacts the packet in place, so its three-byte
    look-back reads bytes it has already overwritten. On a payload carrying
    two runs close together the compaction makes an ordinary ``FF FD`` pair
    read as ``FF FF``, and it removes a byte the servo sent -- its own
    ``addStuffing``/``removeStuffing`` pair is not the identity. The module
    docstring invites a reader to grade this codec against the SDK
    byte-for-byte, which is right for five of its six functions and wrong for
    this one, so without these cases an edit that "aligns" the unescape
    reintroduces a silent single-byte loss with the suite green.
    """

    #: Shortest payload on which the SDK's unescape loses a byte. Six bytes:
    #: the escapes land close enough for the compaction to alias the
    #: look-back. The adversarial corpus above is four bytes wide and so
    #: cannot reach it.
    ALIASING_PAYLOAD = b"\xff\xff\xfd\xff\xfd\xfd"

    def _status_frame(self, params: bytes) -> bytes:
        """Frame ``params`` as the status packet a servo would send back."""
        packet = build_packet(3, Instruction.WRITE, b"\x00" + params)
        body = (packet[:7] + b"\x55" + packet[8:])[:-2]
        crc = checksum(body)
        return body + bytes([crc & 0xFF, (crc >> 8) & 0xFF])

    def test_the_aliasing_payload_round_trips(self) -> None:
        """The six bytes the SDK's unescape drops one of."""
        parsed = parse_status_packet(self._status_frame(self.ALIASING_PAYLOAD))
        assert parsed["params"] == self.ALIASING_PAYLOAD
        assert parsed["crc_ok"] is True

    def test_the_payload_really_carries_two_escapes(self) -> None:
        """Premise: one run alone cannot alias the look-back."""
        packet = build_packet(3, Instruction.WRITE, self.ALIASING_PAYLOAD)
        assert self.ALIASING_PAYLOAD.count(b"\xff\xff\xfd") == 1
        assert len(packet) == 10 + len(self.ALIASING_PAYLOAD) + 1

    def test_six_byte_payloads_round_trip(self) -> None:
        """Widen the corpus by the one byte that reaches the aliasing case."""
        alphabet = (0x00, 0xFD, 0xFE, 0xFF)
        checked = 0
        for combo in itertools.product(alphabet, repeat=6):
            params = bytes(combo)
            parsed = parse_status_packet(self._status_frame(params))
            assert parsed["params"] == params, f"payload {params.hex()} did not round-trip"
            checked += 1
        assert checked == len(alphabet) ** 6
```

### On exposure

The driver in this PR issues no reads yet (no `Instruction.READ`,
`SYNC_READ`, `BULK_READ` or `parse_status_packet` call sites in `driver.py`),
so nothing is currently mis-parsing. But `parse_status_packet` is exported in
`__all__`, and the widths that reach the case start at six bytes -- one
contiguous read of `PRESENT_CURRENT` + `PRESENT_VELOCITY` (2 + 4 at address
126), or any sync/bulk-read reply. Pinning it before the bus lands costs three
cases; finding it afterwards costs a bus-level debugging session on a payload
that appears once in a few thousand reads.

Not a merge blocker in my reading -- the shipped behaviour is right, and I
would take this branch over one that matched the SDK.

yinsong1986 (approved):
## Summary

Ships a pure Protocol 2.0 codec (`build_packet`, `sync_write_packet`, `parse_status_packet`, `checksum`, `decode_model_number`, byte stuffing) plus a `DynamixelDriver` stub satisfying `HardwareDriver`, registered for koch/aloha/vx300s/wx250s/trossen_wxai/dynamixel_2r so `Robot("koch", mode="real", driver="strands")` resolves instead of raising. Write paths return the same named-refusal envelope the G1 driver established, `stream` follows G1's dispatch shape exactly (including the `else: "stop"` fallback), and `driver="strands"` stays opt-in — `resolve_driver` still defaults to lerobot, so no default behaviour changes. The prior round's `MODEL_NUMBERS` concern is resolved by deletion with the reasoning recorded in #2753, and `decode_model_number` is narrowed to `-> int` with the endianness pinned at both range ends.

## What's good

- I re-verified the codec differentially against `dynamixel_sdk` 4.0.5 (`addStuffing` + `updateCRC` via the `txPacket` framing path): ~29k enumerated + 20k randomized build cases, 1.5k status-parse round-trips of SDK-stuffed frames, the three sync-write vectors, and single-byte CRC parity across all 256 table entries — **0 mismatches**, including the adversarial `FF FF FD`-carrying payloads and the `-131073` extended-position value the changelog names.
- All six registered canonical names verified present in `strands_robots/registry/robots.json`, all genuinely Dynamixel/Protocol-2.0 hardware; the Feetech negative-control test pins that registration does not leak to so100/so101/lekiwi.
- `parse_status_packet`'s malformed-frame guards close the crash surface: the length-field consistency check plus the 11-byte minimum make the `payload[1]` ERR read safe for every frame that passes validation, and bad-CRC-reports-not-raises matches the documented retry contract.
- 113/113 on `tests/drivers/test_dynamixel_driver.py` + `tests/test_driver_seam.py` at this head; the tree-wide graders (ASCII tool-result strings, `Raises:` completeness, test-name conventions, `__all__` definedness) all pass; changelog fragment carries the PR's own number; CI fully green including the required check.

## Verification suggestions

The differential check against the real SDK reproduces on any host with no serial port:

```bash
pip download dynamixel_sdk==4.0.5 -d /tmp/dxl --no-deps && (cd /tmp/dxl && unzip -oq *.whl)
```

then compare `build_packet(id, inst, params)` against `Protocol2PacketHandler.addStuffing` + `updateCRC` over `itertools.product((0x00,0x2A,0x74,0xFD,0xFE,0xFF), repeat=n)` for n<=5 — the codec is byte-identical across the whole corpus.

PR #2751, opened 2026-08-26
fix(training): hold FastSAC's replay counts to the shared count domain

cagataycali:
## Problem

`FastSacTrainer.validate` is a pure, read-only preflight that `train` runs before building torch, the env, the networks or the replay buffer, so a misconfigured spec fails with an actionable message instead of a deep stack trace. Three of its guards - on the SAC replay loop's three caller-supplied counts - were bare `value <= 0` tests:

```python
if spec.buffer_size <= 0: ...
if spec.batch_size <= 0: ...
if spec.gradient_steps <= 0: ...
```

Each field is consumed **directly as a count**: `buffer_size` is the replay buffer's capacity (a tensor dimension built in `setup`), `batch_size` is the transitions sampled per gradient step (passed to `ReplayBuffer.sample`), and `gradient_steps` is a `range()` bound in the update loop. A `<= 0` test is far weaker than that consumption demands, and both of its failure modes were measured on the MuJoCo reach env, one field mutated on an otherwise-valid run:

| value | verdict | what happened |
|---|---|---|
| `buffer_size=True` | `status="success"` | a one-slot buffer that never reached `learning_starts`, so **zero** gradient updates ran, yet the run reported success and "10 iterations x 4 steps complete" |
| `buffer_size=0.5` | `IndexError` | `int(0.5) == 0`: a zero-capacity buffer, raised from `ReplayBuffer.add` after `setup` |
| `batch_size=0.5` / `batch_size=True` | `TypeError` | raised from `torch.randint` in `ReplayBuffer.sample` after `setup` |
| `gradient_steps=0.5` | `TypeError` | raised from `range()` in the update loop |
| `"256"` / `None` | `TypeError` | raised from the `<= 0` comparison itself, out of a `validate` documented to *return* problems |

So a `bool` reads as a silent degenerate size - the `buffer_size` case runs a whole training loop that learns nothing and reports success - a fraction or non-finite value passes the comparison and raises deep in the update loop after the env, networks, optimizers and buffer are built (the cost a read-only preflight exists to precede), and a string or `None` raises out of the comparison itself. `batch_size` had a second path to the same raise: even past the `<= 0` guard, `"256"` reached the `learning_starts >= batch_size` relation (`1000 < "256"` gives `TypeError`), so a non-comparable `batch_size` made `validate` raise regardless.

## Change

A new shared gate `rl_replay_problems` (the SAC peer of `rl_run_size_problems`) routes the three counts through the same strict-`int` `positive_count_error` domain `run_size_problems` uses for the supervised `steps` / `global_batch_size` pair - the domain whose contract is that its values are consumed directly as `range()` bounds and tensor dimensions, so only a positive `int` can be honored. `FastSacTrainer.validate` calls it via `Trainer._rl_replay_problems` and drops the three inline comparisons.

Only FastSAC reads these fields; PPO sizes its minibatches from `num_mini_batches` and never reads them, so this is a second gate scoped to the field rather than part of `validate_train_inputs` (per `TrainSpec`, a backend must not be refused for a field it never reads). `learning_starts` and `tau` stay in the backend's own `validate` - the first is one side of a relation, the second a coefficient in `(0, 1]` - so neither shares this count domain.

The `learning_starts >= batch_size` relation now consults the same domain before comparing (`positive_count_error(batch_size) is None and learning_starts < batch_size`), exactly as PPO's `rollout_steps % num_mini_batches` relation does, so a non-count `batch_size` is reported rather than reaching `<`. `learning_starts` itself is unchanged: a non-count there still raises, as before, a separate field on a separate axis.

## Tests

`tests/training/test_rl_replay_domain.py` (307 cells), mirroring `test_rl_run_size_domain.py`. Every domain case reaches the real `validate` entry point, so it covers the wiring as well as the domain.

Pre-fix split, with the gate helper kept and only the `fast_sac.py` call sites reverted (a raw revert is a collection error, since the tests import the gate):

| class | fail+err / total pre-fix |
|---|---|
| `TestFastSacRefusesAnUnusableReplayCount` | **114 / 114** |
| `TestTheUsableDomainIsUntouched` | 9 / 21 (the numpy-int refusals) |
| `TestOneOwnerForTheReplayCountDomain` | 4 / 7 |
| `TestTheDomainIsTheSharedCountRule` | 1 / 66 |
| `TestABackendThatDoesNotReadThemStaysQuiet` | 0 / 86 |
| `TestTheCountsAreConsumedDirectlyAsCounts` | 0 / 9 |
| `TestTauAndLearningStartsAreNotInThisDomain` | 0 / 4 |

The three control classes pass pre-fix, so their silence is scoping rather than an artifact of the fix.

`test_spec_field_read_discovery.py` discovers field-scoped guards structurally; the new gate is registered in `FIELD_SCOPED_GATES` and its guard file in the discovered set, so the meta-guard holds it to the same reader-scan rule as its siblings. `test_rl_fast_sac.py::test_validate_flags_each_out_of_range_field` pinned the old inline messages for the three fields; updated to the shared-domain wording, exactly as it already reads for `total_timesteps` / `rollout_steps`.

## Gate

- `ruff check` + `ruff format --check` clean on all changed files.
- `mypy strands_robots tests tests_integ`: 0 issues in the changed files.
- `tests/training/`: 3522 passed, 4 skipped.
- The docstring / xref / export / except-tuple guards (`test_*_docstring_*`, `test_docstring_xref_roles_resolve`, `test_all_exports_are_statically_defined`, `test_except_*`): 582 passed.


yinsong1986 (approved):
## Summary

Routes FastSAC's three caller-supplied replay counts (`buffer_size`, `batch_size`, `gradient_steps`) through the shared strict-`int` `positive_count_error` domain via a new field-scoped gate `rl_replay_problems` / `Trainer._rl_replay_problems`, replacing three bare `<= 0` comparisons in `FastSacTrainer.validate`. This closes the measured failure modes the description documents: `True` as a silent degenerate size, fractions / non-finite values raising deep inside `setup` or the update loop, and strings / `None` raising out of the comparison itself in a method documented to return problems. The `learning_starts >= batch_size` relation now consults the same domain before comparing, matching the pattern PPO's `rollout_steps % num_mini_batches` relation already uses on `main`. Verified against the head SHA: the three touched test files pass in full (379 tests), `ruff check` / `ruff format --check` are clean on every changed file, and the changed source carries no non-ASCII.

## What's good

- Reuses the existing shared count domain rather than adding a new guard, per AGENTS.md convention 11 (`positive_count_error` already has many callers), so the refusal wording is identical to the run-size and launch-topology gates.
- Scoped to the one backend that reads the fields; `TestABackendThatDoesNotReadThemStaysQuiet` pins that PPO and the supervised backends stay silent, with non-vacuity controls.
- The gate is registered in `FIELD_SCOPED_GATES` and its test file in the discovery set, so the structural meta-guard holds it to the same reader-scan rule as its siblings.
- Changelog fragment present (`changelog.d/2751-rl-replay-count-domain.md`), per step 3 of the PR workflow.
- The deliberate scope lines (`tau`, `learning_starts` outside the domain; numpy ints refused as the documented cost of the strict-`int` rule) are pinned by tests rather than left to prose.

PR #2752, opened 2026-08-26
fix(drivers): stop reporting a LiDAR point cap the summary never applies

cagataycali:
## Problem

`G1Driver._on_lidar_cloud` builds the record the mesh publishes on `strands/<peer>/lidar/summary`. Every field in it is read from the `PointCloud2_` **header** - `width`, `height`, `point_step`, `row_step` - so no point is ever enumerated and nothing is downsampled.

The record nonetheless carried a `capped_at` field, copied from a `lidar_max_points` constructor parameter. That parameter had exactly one reader:

```
$ grep -rn "lidar_max_points\|_LIDAR_MAX_POINTS" --include=*.py strands_robots/
strands_robots/drivers/g1.py:52:# ``_LIDAR_MAX_POINTS`` cap is deliberate: :meth:`_summarise_cloud` runs on the
strands_robots/drivers/g1.py:57:_LIDAR_MAX_POINTS: int = 4000
strands_robots/drivers/g1.py:87:        lidar_max_points: int = _LIDAR_MAX_POINTS,
strands_robots/drivers/g1.py:112:            lidar_max_points: Cap for the downsampled point count reported in
strands_robots/drivers/g1.py:125:        self._lidar_max_points = int(lidar_max_points)
strands_robots/drivers/g1.py:559:                "capped_at": self._lidar_max_points,
```

Line 559 is the only place the value is consumed, and it copies it into the payload. So a Mid-360 frame reached the wire as:

```
strands/neon/lidar/summary
{"capped_at": 4000, "count": 24000, "height": 1, "peer_id": "neon",
 "point_step": 16, "row_step": 384000, "t": ..., "width": 24000}
```

`count: 24000` beside `capped_at: 4000` tells a consumer the number next to it was capped at 4000. It is the cloud's true, uncapped size. Setting the knob changed only that claim.

The parameter's own docstring described behaviour the method does not implement - "Cap for the downsampled point count reported in `_lidar_summary`" - and the comment defending the constant named a method that does not exist:

```
# ``_LIDAR_MAX_POINTS`` cap is deliberate: :meth:`_summarise_cloud` runs on the
```

`_summarise_cloud` has no definition anywhere in the tree (`grep -rn` over `--include=*.py` returns only that comment).

## Change

The field, the parameter and the constant are removed, and the comment is corrected. `_on_lidar_cloud`'s docstring now states what actually bounds the record: every field comes from the header, so the summary has the same shape for a 200-point frame and a 30k-point one, and there is no point count for a cap to apply to.

**`count` is deliberately unchanged.** The tempting repair is to clamp it, and that would be worse than the field: a Mid-360 dropping from 24000 points to 3000 is reporting a fault, and clamping would hide exactly that. A control cell pins this, and the mutation that clamps `count` is in the break table below.

**No caller breaks.** The driver already accepts and ignores extra keyword arguments so the factory can forward them, so a caller still passing `lidar_max_points=` is logged at debug rather than refused - pinned by `test_an_unknown_keyword_is_still_accepted`. `capped_at` and `lidar_max_points` have zero consumers anywhere: no other module, no docs page, and nothing on the dashboard branch (checked with `git grep` against it).

## Why the existing coverage was silent

`tests/drivers/test_g1_driver.py::test_lidar_cloud_summary_is_bounded` asserted both halves of the contradiction in one body:

```python
assert driver._lidar_summary["count"] == 24000
assert driver._lidar_summary["capped_at"] == 4000
```

so the mismatch was pinned rather than caught. Its docstring said "The cap is asserted here so a regression is loud". What actually bounds the record is the fixed shape, which the same test also asserts via `"points" not in`. That cell keeps the shape assertion, drops the cap assertion, and its docstring now names the real bound.

## Tests

New `tests/drivers/test_g1_lidar_summary_describes_the_cloud.py`, 18 cases in four classes. Pre-fix (source reverted to `main`, tests kept): **4 failed / 14 passed**, all four in the two regression classes and **0 of the 14 control cells**.

The load-bearing regression cell is derived rather than a change-detector: `test_the_summary_is_a_function_of_the_message_alone` summarises one cloud through two differently-configured drivers and requires identical records modulo the clock stamp. A cloud summary describes the cloud; any field that moves when the driver's configuration moves is describing the driver instead, which is what `capped_at` did.

`test_every_meth_cross_reference_resolves` derives every `:meth:` reference from the module source and requires each to resolve to a callable - 7 today, of which `_summarise_cloud` was the one dead reference. It is restricted to `:meth:` on purpose: a method either exists or it does not, whereas an `:attr:` reference may legitimately name an instance attribute assigned in `__init__` and so absent from the class (`_imu`, `_battery`, `_fsm_id`, `_lidar_state`, `_lidar_summary` are all of that kind, and an unrestricted scan reports all five as false positives).

### Break table

Nine mutations against the fixed tree; source restored byte-identical after each.

| mutation | fires | cells |
|---|---|---|
| control (unmutated) | 0 | - |
| the cap field returns as a literal | 1 | `test_the_summary_carries_no_cap_field` |
| full revert (parameter + field) | 4 | the pre-fix set |
| the dead parameter returns, field stays gone | 1 | `test_no_constructor_parameter_is_named_for_a_point_cap` |
| **the wrong fix: `count` clamped to a cap** | 2 | both count cells |
| the dead `:meth:` reference returns | 1 | `test_every_meth_cross_reference_resolves` |
| a header field dropped from the record | 1 | `test_every_header_field_is_reported[width]` |
| the record gains another config-derived field | 2 | function-of-message, no-point-list |
| `count` loses the height factor | 1 | `test_an_organised_cloud_counts_width_times_height` |
| the `:meth:` scan regex matches nothing (vacuity probe) | 1 | `test_every_meth_cross_reference_resolves` |

9 mutations, **9 detected**, 8 distinct firing sets, control clean. The last two rows share a firing set for a structural reason: one is the defect returning, the other is the vacuity probe on the same cell, and the non-vacuity assertion inside it is what makes the second row fire.

One mutation initially fired 0 and exposed a gap in my own fixture: `count = width` is indistinguishable from `count = width * height` when every fixture is an unorganised Mid-360 frame with `height=1`. An organised-cloud fixture (640x480) closes it, which is why that row now fires.

## Gate

- `ruff check` and `ruff format --check` over `strands_robots tests tests_integ`: clean, 1567 files formatted.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `main` worktree, normalized message sets byte-identical in both directions, **0 in the changed files**.
- Paired run, identical 100-file selection (all of `tests/drivers/` plus every guard that walks the tree, reads source or docstrings, or reads `changelog.d`), with both changed test files excluded from each tree: identical **4173 collected** and `1 failed, 4117 passed, 55 skipped` on both, one identical failing id, `comm` empty in both directions. That failure is `tests/test_dataset_recorder.py::test_create_passes_vcodec_directly_when_supported`, a pre-existing `FileExistsError` on a leftover `/tmp/ds` dataset directory on this machine, unrelated to this change and identical on both sides.
- `tests/drivers/` plus `tests/test_driver_seam.py`: 110 passed, on both a mujoco 3.5.0 (declared floor) and a 3.12.0 interpreter.

No visual artifact: this changes what a record reports, not a policy, a simulation, a render or a recording. The measured before/after payload above is the evidence.

### DIFF
--- a/changelog.d/2752-g1-lidar-summary-no-unapplied-cap.md
+++ b/changelog.d/2752-g1-lidar-summary-no-unapplied-cap.md
@@ -1,0 +1,21 @@
+### Fixed: a G1 LiDAR summary no longer reports a point cap it never applies
+
+`G1Driver._on_lidar_cloud` builds the record the mesh publishes on
+`strands/<peer>/lidar/summary`. Every field in it is read from the
+`PointCloud2_` header - width, height, point_step, row_step - so no point is
+ever enumerated and nothing is downsampled. The record nonetheless carried a
+`capped_at` field copied from a `lidar_max_points` constructor parameter, and
+that parameter had exactly one reader: the line that copied it into the field.
+A Mid-360 frame therefore published `count: 24000` beside `capped_at: 4000`,
+telling a consumer that the number next to it had been capped at 4000 when it
+was the cloud's true uncapped size. Setting the knob changed only that claim.
+The comment defending the constant named `_summarise_cloud`, a method with no
+definition anywhere in the tree.
+
+The field and the parameter are gone; a caller still passing `lidar_max_points=`
+is accepted and logged at debug, because the driver already ignores extras so
+the factory can forward them. `count` is deliberately unchanged: it is the
+cloud's true size, and a Mid-360 that drops from 24000 points to 3000 is
+reporting a fault that clamping the number would hide. What bounds the record is
+its fixed, header-derived shape - the same for a 200-point frame and a
+30k-point one - which is now what the docstring says and what the tests assert.
--- a/strands_robots/drivers/g1.py
+++ b/strands_robots/drivers/g1.py
@@ -48,13 +48,8 @@ logger = logging.getLogger(__name__)
 
 
 # ---------------------------------------------------------------------------
-# Constants surfaced for tests and for issue #358 to share. The
-# ``_LIDAR_MAX_POINTS`` cap is deliberate: :meth:`_summarise_cloud` runs on the
-# DDS thread and a full Livox frame is ~30k points at 10 Hz. The point-cloud
-# tile (issue #356) publishes its own downsampled ``lidar/cloud`` topic; here
-# we only compute a summary.
+# Constants surfaced for tests and for issue #358 to share.
 # ---------------------------------------------------------------------------
-_LIDAR_MAX_POINTS: int = 4000
 _BATTERY_FLOOR_PCT: float = 15.0
 
 # The topics the driver reads. Kept together so a reader sees the whole
@@ -84,7 +79,6 @@ class G1Driver:
         port: str | None = None,
         network_interface: str = "eth0",
         battery_floor_pct: float = _BATTERY_FLOOR_PCT,
-        lidar_max_points: int = _LIDAR_MAX_POINTS,
         **kwargs: Any,
     ) -> None:
         """Record configuration; :meth:`connect_eagerly` does the DDS work.
@@ -109,9 +103,6 @@ class G1Driver:
             battery_floor_pct: Percentage below which :meth:`send_action`
                 refuses to write. The floor is separate from the FSM gate so
                 a caller can see which check refused.
-            lidar_max_points: Cap for the downsampled point count reported in
-                ``_lidar_summary``. Full clouds go through the dashboard's
-                own topic (issue #356), not this one.
             **kwargs: Ignored; accepted so the factory can forward extras
                 without the driver knowing what they are.
         """
@@ -122,7 +113,6 @@ class G1Driver:
         self._port = port
         self._network_interface = network_interface
         self._battery_floor_pct = float(battery_floor_pct)
-        self._lidar_max_points = int(lidar_max_points)
 
         # Cached DDS decode results. Every one is optional per the mesh
         # contract, so a driver that has not connected yet is not broken.
@@ -549,6 +539,14 @@ class G1Driver:
         too much to publish unpaced. The summary is what the mesh's health
         chip reads; the 3D tile (issue #356) subscribes the raw cloud itself
         through a paced publisher and does its own downsampling.
+
+        What bounds this record is that every field is read from the message
+        *header* - ``width``, ``height``, ``point_step``, ``row_step`` - so its
+        size is the same for a 200-point frame and a 30k-point one. No point is
+        enumerated here, and nothing is downsampled, so there is no point count
+        for a cap to apply to. ``count`` is therefore the cloud's true size: a
+        MID-360 that drops from 24000 points to 3000 is reporting a fault, and
+        clamping the number would hide exactly that.
         """
         try:
             width = int(getattr(msg, "width", 0))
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Removes a `capped_at` field from the record `G1Driver._on_lidar_cloud` publishes on `strands/<peer>/lidar/summary`, along with the `lidar_max_points` constructor parameter and `_LIDAR_MAX_POINTS` constant that fed it. Every field in the summary is read from the `PointCloud2_` header, so no cap is ever applied; the field was a false claim beside the true uncapped `count`, and the comment defending the constant cited `_summarise_cloud`, a method that does not exist. Verified independently: `capped_at` and `lidar_max_points` have zero consumers anywhere else in the tree at the head SHA; the constructor's existing `**kwargs` absorption (logged at debug) means a caller still passing `lidar_max_points=` does not break, so the parameter removal is not a breaking signature change; and the field removal is corrective rather than a one-way door - the record was lying about the number next to it. Re-ran the new suite against the head (18 passed) and against the pre-fix source (4 failed / 14 passed, exactly the two regression classes claimed), plus the drivers suite, driver seam, changelog-fragment, host-path, ASCII, xref-resolution and test-naming guards - all green. The deliberate decision to leave `count` unclamped is the right one (a MID-360 dropping from 24k to 3k points is reporting a fault) and is pinned by a control cell.

## What's good

- The load-bearing regression cell (`test_the_summary_is_a_function_of_the_message_alone`) is derived from the invariant rather than being a change-detector, and the wrong fix (clamping `count`) is explicitly pinned against.
- The organised-cloud fixture closing the `count = width` vs `width * height` gap is exactly the kind of vacuity check AGENTS.md asks for, as is the non-empty-scan assertion in the `:meth:` resolver test.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; docstrings updated to state the real bound instead of deleting the explanation.

yinsong1986 (approved):
## Summary

Removes the `capped_at` field from the LiDAR summary record `G1Driver._on_lidar_cloud` publishes on `strands/<peer>/lidar/summary`, along with the `lidar_max_points` constructor parameter and `_LIDAR_MAX_POINTS` constant that fed it. Every field in the record is read from the `PointCloud2_` header, so no cap is ever applied; the field was a false claim beside the true uncapped `count`, and the comment defending the constant cited a `_summarise_cloud` method that has no definition in the tree. Verified against the head: `capped_at` / `lidar_max_points` have zero remaining consumers anywhere in the tree, the removed parameter was keyword-only and sat behind `**kwargs`, so a caller still passing it is absorbed and logged at debug rather than refused (pinned by `test_an_unknown_keyword_is_still_accepted`), and `count` is deliberately left as the cloud's true size with a control cell pinning that a clamp would be a regression. The 18 new cells plus the amended `test_lidar_cloud_summary_is_bounded` all pass against the head (65 passed across `tests/drivers/test_g1_driver.py` and the new file), the dead `:meth:` reference guard carries its own vacuity probe, and the changelog fragment follows the `changelog.d/` convention. No blockers found.

## What's good

- The load-bearing regression cell (`test_the_summary_is_a_function_of_the_message_alone`) is derived from the invariant rather than a change-detector: any config-derived field returning to the record fires it, not just `capped_at`.
- The organised-cloud fixture closing the `count = width` vs `count = width * height` blind spot, found by the author's own mutation run.
- Constructor compatibility handled per the existing `**kwargs` forwarding contract instead of a breaking signature change.

Issue #2749, opened 2026-08-26
drivers: a contract-complete native driver publishes no joint telemetry on the mesh


cagataycali:
## Summary

A driver that satisfies every member of `DRIVER_SURFACE` publishes no joint telemetry on the mesh. The 12-member contract is fully satisfiable while `_read_state()` answers `None`, and the presence heartbeat keeps advertising the peer either way.

This is measurable on the first native driver in the tree. `G1Driver` from PR #2739 is contract-complete and can publish its IMU orientation, its lidar summary and its motor temperatures, and cannot publish the position of a single joint.

## Reproduction

On `upstream/main` at 442d9ca4, with `G1Driver` from #2739:

```
driver                 : G1Driver from strands_robots.drivers.g1
missing_driver_members : ()
has .robot             : False
has get_observation    : False

_read_state()          : None
joints section         : <ABSENT>
imu via _imu cache     : {'peer_id': 'g1__probe', 't': ..., 'rpy': [0.01, -0.02, 0.5]}
health temps via _temps: {'shoulder_pan': 41.5, 'gripper': 38.0}
```

The same asymmetry with a stand-in that isolates the one variable. Two objects with identical telemetry, both `missing_driver_members() == ()`; one delegates to an inner lerobot device the way `hardware_robot.Robot` does, the other owns its bus directly:

| | `missing_driver_members` | `_read_state()` | `joints` | `presence.connected` | `presence.hw` | `presence.topics` |
|---|---|---|---|---|---|---|
| delegating | `()` | dict | 3 motors | `True` | `'so101_follower'` | `['health']` |
| native | `()` | `None` | `<ABSENT>` | absent | absent | `['health']` |

Both advertise the identical presence payload (`robot_id`, `tool_name`, `topics`), so the two are indistinguishable to a fleet consumer. `_read_state`'s own docstring names that presentation as the one it exists to avoid: "a hardware peer whose only section was `joints` returned `None` for every tick of a contended bus, so it went silent on the state topic while its presence heartbeat kept advertising it -- indistinguishable from a peer whose state thread had died".

A self-contained script is at the end of this issue.

## Why the contract cannot catch it

Joint telemetry is reachable only through the inner device. Six sites, all on 442d9ca4:

```
strands_robots/mesh/core.py:869                 getattr(self.robot, "robot", None)   camera advisory
strands_robots/mesh/core.py:1058                getattr(r, "robot", None)            presence: connected, hw, cameras, inputs
[...]

PR #354, opened 2026-06-06
test(mesh): pin STRANDS_MESH_BRIDGE_TOPICS_PREFIX reader + default-prefix tuple

cagataycali:
Closes #244.

## What

Pins the `STRANDS_MESH_BRIDGE_TOPICS_PREFIX` env-var reader (`_resolve_bridge_prefix_filter`) and its unset-default tuple (`_DEFAULT_BRIDGE_PREFIX_SUFFIXES == frozenset({'response'})`) against the Phase-4 cloud-pollution hardening documented in `bridge_transport.py:107-121`.

Issue #244 verified possibility 1: the reader landed in PR #222 and is present on main (`bridge_transport.py:141-156`), but had no regression pin. This is case (b) FIXED-but-not-pinned.

## Test

New `TestEnvPrefixFilter` in `tests/mesh/test_bridge_transport.py`:
- env-var name parses comma-separated prefixes with whitespace trimming
- unset default is the documented `response` prefix
- unset default bridges a `response/<turn>` tail
- **unset default blocks `cmd/<attacker-tail>`** (Phase-4 security pin)
- empty env falls back to default
- whitespace-only env falls back to default
- `STRANDS_MESH_BRIDGE_TOPICS` does not leak into prefix reader

```
hatch run python -m pytest tests/mesh/test_bridge_transport.py -k TestEnvPrefixFilter -q
======================= 7 passed, 45 deselected in 0.14s =======================
```

No source change; reader-side pin only.

---

## §13 Review Changelog

| Round | Concern | Fix Commit | Pin Test |
|-------|---------|------------|----------|
| R1 | CI lint failure: `ruff format --check` violation in assert statement (trailing parenthesized form) | `3982875` | N/A (format-only fix, existing tests unchanged) |
| R2 | Misleading `_README_HINT` citing non-existent README section; missing security exclusion test for `cmd/<tail>`; method name collision; coverage gaps (whitespace-only env, reader independence) | `783511f` | `tests/mesh/test_bridge_transport.py::TestEnvPrefixFilter::test_unset_default_blocks_cmd_tail` |

PR #2754, opened 2026-08-26
test(drivers): pin the G1 driver's DDS readings reaching the mesh wire

cagataycali:
Closes the fourth acceptance criterion of the G1 native-driver work: a `Mesh` wrapping a `G1Driver` fed mocked DDS publishes `strands/<peer>/lidar/summary`, `imu` and `health`. Every other criterion was pinned when the driver landed; this one was verified by reading the two sides and checking that the key names agreed.

## Why an inspection is not enough here

A native driver and the mesh sensor loops are joined by **nothing but attribute names**. `SensorLoopsMixin` reads sixteen underscore-prefixed attributes straight off whatever object the mesh was handed, and `G1Driver` fills four of them from its DDS callbacks. Nothing declares that relationship, so a rename on either side is silent.

#2749 records the same seam failing in the other direction: a contract-complete native driver publishes no joint telemetry at all, because joints are reached through `getattr(robot, "robot", None)` - an inner lerobot device a native driver does not have. Pinning which topics *do* arrive is what makes that boundary legible rather than folklore.

## What this grades that a reader-level test cannot

`tests/mesh/test_sensor_readers.py` drives the readers through a host whose `publish` **records** the payload. That misses three things:

1. **The composition.** A real `Mesh` around the real driver, so the attribute names are checked against the object the mesh actually reads - not against a bag of attributes a fixture set.
2. **The encoding.** `session._put_zenoh_directly` runs `json.dumps` before the bytes reach `put`, and a payload it refuses is dropped for good: the failure is deterministic, so no later tick recovers it. The capture here encodes what it is handed, which is the same step moved to where a test can see which side of it a record landed.
3. **That the DDS side is numpy.** A Livox header field is `int64` and an IMU orientation is `float32`, and `json.dumps` refuses both. Feeding plain Python floats would grade a payload the robot never sends.

## Determinism

No threads, no sleeping, no network. `SensorLoopsMixin._paced` yields *before* it waits, so a stop event that is already set produces exactly one iteration per loop and then breaks. The loops are called directly rather than through `Mesh.start()`, and the publish-rate environment variables are cleared so a rate configured on the host cannot turn a loop off and change what the file grades.

## Cells

19 cases in four classes:

- **`TestTheCriterionTopicsArriveEncoded`** - the three topics arrive, nothing was refused by the encoder, and each record carries the reading its DDS sample supplied (IMU orientation, battery charge, cloud size).
- **`TestTheLidarStateTopicComesWithTheSummary`** - the lidar loop publishes two topics; the criterion names one.
- **`TestASilentDriverPublishesNoReading`** - non-vacuity. An unfed driver publishes no imu or lidar topic, so the cells above are the DDS samples rather than the loop merely running. Health is the one topic that survives, because it aggregates host metrics too, and that is asserted so the silent-topic cell is read for what it is.
- **`TestTheDriverIsWhatTheMeshReads`** - the seam itself, including that there is no inner lerobot device. Stated as the premise of the topics above, not as something desirable.

## Break table

Eight mutations across `drivers/g1.py` and `mesh/sensors.py`; both sources restored byte-identical after each.

| mutation | fires | cells |
|---|---|---|
| control (unmutated) | 0 | - |
| the driver renames its `_imu` cache | 3 | seam attribute, imu record, imu topic |
| the mesh reads a different lidar-summary attribute | 2 | summary topic + contents |
| the driver stops coercing the IMU | **0** | - |
| that, **and** the mesh's IMU coercion removed | 3 | encoder-refusal cell, imu record, imu topic |
| the summary topic loses its peer segment | 3 | peer-addressing cell, summary topic + contents |
| health stops reading the battery | 1 | battery charge |
| the lidar loop stops publishing the summary | 2 | summary topic + contents |
| the driver's decoders write nothing (vacuity probe) | 11 | all four seam cells plus every topic cell |

Eight mutations, seven detected, six distinct firing sets among those seven, control clean.

Two rows are worth reading rather than counting:

- **The driver-stops-coercing row fires 0, and that is correct.** `G1Driver`'s decoders already coerce with `float(...)`/`int(...)`, so a numpy value never reaches the mesh from *this* driver, and the mesh's `_coerce_record` is belt-and-braces for it. The next row removes both layers and fires 3 - so each layer alone suffices, and neither is dead. Reported rather than dropped, because "the encoder protection is exercised" would otherwise be an overclaim.
- The mesh-reads-a-different-attribute row and the loop-stops-publishing row share a firing set. Both make the summary topic never arrive - one because the reader finds no data, one because the publish is skipped - so they are the same observable at the wire, which is the level this file grades at.

## Gate

- `ruff check` and `ruff format --check` over `strands_robots tests tests_integ`: clean, 1567 files formatted.
- `mypy strands_robots tests tests_integ`: **24 == 24** against `main`, normalized message sets byte-identical in both directions, **0 in the new file**.
- The new file: 19 passed. `tests/drivers/` plus `tests/test_driver_seam.py`: 111 passed, on a mujoco 3.5.0 (declared floor) interpreter.
- Run together with all of `tests/mesh/` (the new file monkeypatches the module-level `put`, so leakage would show there): `4309 passed`, with 21 failures and 24 errors that are all `tests/mesh/test_iot_*` needing `awsiot` - absent here and declared in the `[mesh-iot]` extra. `tests/mesh/` alone produces the identical 21 and 24 with `4198 passed`, and 4309 - 4198 is exactly the 111 above, so nothing was lost or gained by combining them and the failures are pre-existing.

Test-only: no production file changes, so no behaviour to picture.

### DIFF
--- a/changelog.d/2754-g1-driver-mesh-wire-test.md
+++ b/changelog.d/2754-g1-driver-mesh-wire-test.md
@@ -1,0 +1,22 @@
+### Tests: a native G1 driver's DDS readings are pinned reaching the mesh wire
+
+A native driver and the mesh sensor loops are joined by nothing but attribute
+names. `SensorLoopsMixin` reads sixteen underscore-prefixed attributes off
+whatever object the mesh was handed, and `G1Driver` fills four of them from its
+DDS callbacks; nothing declares that relationship, so a rename on either side is
+silent. The coverage that existed drove the readers through a host object whose
+`publish` records the payload, which cannot see either the composition or the
+encoding.
+
+This composes a real `Mesh` around a real `G1Driver` fed mocked DDS and asserts
+at the transport boundary that `strands/<peer>/lidar/summary`, `imu` and `health`
+arrive and decode. The readings are fed as the numpy the SDK reports - a Livox
+header is `int64`, an IMU orientation `float32`, and `json.dumps` refuses both -
+so the cells grade the payload the robot really sends rather than a
+pre-sanitised stand-in. A non-vacuity class shows the topics are the DDS samples
+and not the loop merely running: an unfed driver publishes no imu or lidar
+topic, while health still publishes because it aggregates host metrics too.
+
+The loops are driven one tick each without threads or sleeping, because
+`SensorLoopsMixin._paced` yields before it waits and so an already-set stop
+event produces exactly one iteration.
--- a/tests/drivers/test_g1_driver_reaches_the_mesh_wire.py
+++ b/tests/drivers/test_g1_driver_reaches_the_mesh_wire.py
@@ -1,0 +1,235 @@
+"""A native G1 driver's DDS readings reach the mesh wire, encoded.
+
+Issue #354's fourth acceptance criterion: a :class:`~strands_robots.mesh.Mesh`
+wrapping a :class:`~strands_robots.drivers.g1.G1Driver` fed mocked DDS publishes
+``strands/<peer>/lidar/summary``, ``imu`` and ``health``. Every other criterion
+was pinned when the driver landed; this one was verified by reading the two
+sides and checking that the key names agreed, which is exactly the check a test
+should be doing instead.
+
+What makes it worth a test rather than an inspection is that the two halves are
+joined by nothing but attribute names. ``SensorLoopsMixin`` reads sixteen
+underscore-prefixed attributes straight off whatever object the mesh was handed;
+the driver writes four of them from its DDS callbacks. Nothing declares that
+relationship, so a rename on either side is silent - and issue #2749 records the
+same seam failing in the other direction, where a contract-complete driver
+publishes no joint telemetry at all because joints are reached through an inner
+lerobot device this driver does not have. Pinning which topics *do* arrive is
+what makes that boundary legible.
+
+Three properties this file asserts that a reader-level test cannot:
+
+- the composition, with a real ``Mesh`` rather than a mixin host, so the
+  attribute names are checked against the object the mesh really reads;
+- the *encoding*, because ``session._put_zenoh_directly`` runs ``json.dumps``
+  before the bytes reach ``put`` and a payload it refuses is dropped for good -
+  the failure is deterministic, so no later tick recovers it. The capture here
+  encodes what it is handed, which is the same step moved to where a test can
+  see which side of it a record landed;
+- that the DDS side really is numpy. A Livox frame's header fields and an IMU's
+  orientation arrive as ``int64``/``float32``, and ``json.dumps`` refuses both,
+  so feeding plain Python floats would test a payload the robot never sends.
+
+The loops are driven one tick each, deterministically and without sleeping:
+``SensorLoopsMixin._paced`` yields *before* it waits, so a stop event that is
+already set produces exactly one iteration and then breaks.
+"""
+
+from __future__ import annotations
+
+import json
+import types
+from typing import Any
+
+import numpy as np
+import pytest
+
+import strands_robots.mesh.core as mesh_core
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Test-only PR pinning issue #354's fourth acceptance criterion at the transport boundary: a real `Mesh` composed around a real `G1Driver` fed mocked DDS samples publishes `strands/<peer>/lidar/summary`, `imu` and `health`, and the payloads survive the `json.dumps` step the transport performs. The `_Capture.put` doubles the encode `session._put_zenoh_directly` runs, so an encoder refusal lands in `dropped` where the test can see it instead of vanishing. The DDS fixtures deliberately carry `int64`/`float32` numpy values so the graded payload is the one the robot's SDK actually reports, and a non-vacuity class (unfed driver publishes no imu/lidar topic while health still publishes from host metrics) shows the topics come from the DDS samples rather than the loop merely running. Loops are driven one tick each with a pre-set stop event -- no threads, sleeps, or network. Verified locally at the head SHA: all 19 tests pass against `main`'s driver and mesh.

## What's good

- The seam under test (sixteen attribute names joining `SensorLoopsMixin` to the driver) is real and previously unverified end-to-end; the composition test with a real `Mesh` is the right level.
- Patches `put` at `mesh_core`, the module `Mesh.publish` actually resolves it from -- the correct seam per the mesh's own fixture conventions.
- Env hygiene follows AGENTS.md: `monkeypatch.delenv` for the five rate overrides (so a host-configured `*_HZ=0` cannot silently turn a graded loop off), no direct `os.environ` mutation, no host paths, ASCII throughout.
- Changelog fragment present and correctly numbered (`changelog.d/2754-...`); no edit to `CHANGELOG.md`.
- Test class names describe behaviour, with issue provenance kept to docstrings.

yinsong1986 (approved):
## Summary

Test-only PR closing issue #354's fourth acceptance criterion at the level it actually lives: a real `Mesh` composed around a real `G1Driver` fed mocked DDS samples publishes `strands/<peer>/lidar/summary`, `imu` and `health`, with the payloads surviving the `json.dumps` step the transport performs. I verified the test's premises against `main` rather than taking the description's word: `Mesh.publish` is a single chokepoint forwarding to the module-level `put` in `mesh.core`, so the monkeypatched capture intercepts exactly the transport boundary; `Mesh.__init__` opens no session and spawns no threads (only `start()` does), and `G1Driver.__init__` records the port without touching DDS, so the file is genuinely network- and thread-free; the four cache attributes (`_imu`, `_battery`, `_lidar_state`, `_lidar_summary`) and topic names match `SensorLoopsMixin`'s readers byte for byte; and the driver's decoders do coerce with `float()`/`int()`, which is what makes the break table's zero-firing coercion row the honest reading rather than a gap. The numpy-typed fixtures (`int64`/`float32`) grade the payload the SDK actually reports, and the non-vacuity class (unfed driver publishes no imu/lidar topic while health survives on host metrics) keeps the positive cells from passing on a loop that merely ran. AGENTS.md conventions hold: changelog fragment present, `monkeypatch.delenv` for env vars, ASCII-only strings, behaviour-named test classes, no host paths.

## What's good

- The `_Capture.put` records what the encoder *refuses* as well as what it accepts, so an encoder-refused payload fails a cell instead of vanishing - the exact silent-drop shape the real `_put_zenoh_directly` has.
- `TestTheDriverIsWhatTheMeshReads` pins the #2749 seam (no inner lerobot device) as the stated premise of the topic cells rather than as a loose extra assertion.
- The break table reports the zero-firing mutation and explains why it is correct, instead of dropping the row - the layered-coercion claim is measured in both directions.

PR #2755, opened 2026-08-26
fix(tools): verify a Feetech reply instead of reading it at fixed offsets

cagataycali:
## The reply was read at fixed offsets

`pose_tool`'s `read_motor_position` wrote an `INST_READ` for `Present_Position`, then did this with whatever came back:

```python
response = self.serial_conn.read(10)
if len(response) >= 7:
    position = response[5] | (response[6] << 8)
```

A Feetech status packet is `FF FF ID LEN ERR <params> CHK`. Indices 5 and 6 are the position bytes only if the frame begins at offset 0, is intact, and came from the motor that was asked. The bus guarantees none of those: it is half-duplex and shared by all six servos, so a reply can arrive behind a byte the host's own transmission echoed, or be the late answer to a read that already timed out, sent by a different motor.

Measured through the public `read_motor_position("shoulder_pan")` with the joint really at 1024 counts (-89.98 deg), against `scservo_sdk` -- the vendor SDK that owns this wire format, which the `[lerobot]` extra already installs -- on the identical bytes:

| bytes returned by the bus | before | after | `scservo_sdk` |
|---|---|---|---|
| `ff ff 01 04 00 00 04 f6` (clean) | -89.98 | -89.98 | `COMM_SUCCESS`, param 1024 |
| `00` + clean (one echoed byte) | **-180.0** | -89.98 | `COMM_SUCCESS`, param **1024** |
| `00 00` + clean | **-179.65** | -89.98 | `COMM_SUCCESS`, param **1024** |
| `ff ff 02 04 00 00 04 f5` (motor 2 answered) | **-89.98** | `None` | reports `id=2`, not the id asked |
| clean with the checksum zeroed | **-89.98** | `None` | **`COMM_RX_CORRUPT`** |
| clean, one byte short | **-89.98** | `None` | **`COMM_RX_CORRUPT`** |
| clean with `ERR=0x04` (overheat) | -89.98 | -89.98 | `COMM_SUCCESS`, reports `err=0x04` |

Every row returned `status="success"` before. Note the second and third rows: the SDK recovers the **true** 1024 from exactly those bytes, so this is not a case where being stricter is the fix -- the offset reply carries the real position and it was simply read in the wrong place.

## It was a motion fault, not only a reporting one

`_smooth_move` reads the current position and divides the travel to the target into increments, so a misread start is a trajectory that begins somewhere the arm is not. Commanding `shoulder_pan` from 1024 counts to 4095, with one stray byte on the bus:

commanded trajectory before and after

```
before   [   0,  511, 1023, 1535, 2047, 2559, 3071, 3583, 4095]
after    [1024, 1407, 1791, 2175, 2559, 2943, 3327, 3711, 4095]   == the clean-bus trajectory, exactly
```

The first packet of a *smooth* move commanded the joint 1024 counts (90.0 degrees) away from its target instead of toward it. `incremental_move` added its delta to the same fiction; `store_pose` would have persisted it.

## Why nothing caught it

Three test fixtures, in three files, built the reply like this:

```python
bytes([0xFF, 0xFF, 0x01, 0x04, 0x00, raw & 0xFF, (raw >> 8) & 0xFF, 0, 0, 0])
```

`0` sits where the checksum belongs, the ID is hardcoded to 1 for every motor, and two trailing zeros pad it out -- a frame no servo sends and one `scservo_sdk` calls `COMM_RX_CORRUPT`. A double that cannot express a corrupt or misattributed reply makes those cases untestable by construction, and the shipped test's own comment recorded the shape it was pinning: *"7-byte response with position low/high at indices 5 and 6."*

All three now carry a real checksum and answer as the motor the outgoing packet addressed. The correct formula was already in the suite -- `test_pose_tool_emergency_stop.py` computes `~sum(body) & 0xFF` when it checks *write* packets -- only the read fixtures were wrong.

## The change

`_parse_status_packet(raw, motor_id, param_count)` locates and verifies the frame, mirroring the SDK: search for the `FF FF` header rather than assuming offset 0, re-derive the frame length from `LEN`, recompute the checksum over the same span the outgoing packet signs, and require the responding ID to be the one asked. Every candidate header is tried, so a frame behind noise is found and a payload byte pair that merely looks like a header does not end the search.

A frame that cannot be verified reports `None` -- the existing "could not read" signal, so no caller contract changes: `read_position` answers with an error envelope instead of a number, and the interpolating paths decline to build a trajectory. The discarded bytes are logged so a bus fault stays diagnosable.

`read(10)` is deliberately unchanged: the slack beyond the 8-byte frame is what absorbs the leading bytes the parse now skips.

## Deliberately not decided here

- **Acting on the error byte.** A value inside the byte's range is a valid frame, as the SDK also holds, so a servo reporting overheat still reports its position. Refusing to drive a faulted servo changes what a caller is told about a healthy reading and wants its own change; `TestVerificationRefusesNothingTheFormatAllows` pins that this parse does not quietly decide it.
- **Bounding the value.** `Present_Position` is 12-bit, but a well-formed frame carrying 0xFFFF is still a well-formed frame. Framing and value domain are separate questions and the suite pins that this answers only the first.

## Tests

`tests/tools/test_feetech_status_packet_framing.py`, 48 cases. Reverting only the source (keeping the helper, restoring the positional read) gives **19 failed / 48**, with **0** of the 4 oracle-premise cells, **0** of the 5 controls, **0** of the 2 over-reach cells and **0** of the 3 parse-contract cells -- those last are pinned by the mutation table instead, since the helper still exists under that revert. A raw revert is a collection error, the file importing `_parse_status_packet`.

The premise class grades this file's own fixtures against `scservo_sdk` before anything else asserts on them, so the well-formed frames are known-good and the malformed ones known-bad rather than agreeing with a parser written alongside them. It is `importorskip`-gated; every other cell needs no dependency and no hardware.

12 mutations, 12 distinct failing-ID sets, none undetected, control clean, source restored byte-identical:

| mutation | fires |
|---|---|
| the whole verification dropped (pre-fix read) | 18 |
| header search dropped, parse at offset 0 only | 9 |
| checksum verification dropped | 5 |
| responding-ID check dropped | 2 |
| truncation check dropped | 1 |
| parameter-count check dropped | 1 |
| error-byte sanity dropped | 1 |
| broadcast/header IDs accepted as responders | 1 |
| search stops at the first candidate header | 1 |
| checksum summed over the whole frame | 35 |
| frame length off by one | 36 |
| the error byte returned as a parameter | 20 |

Three of those first fired 0 and each was a gap in the tests, not a redundant check. The sharpest is the truncation one: a reply reporting 64000 counts is `ff ff 01 04 00 00 fa 00`, and dropping its checksum byte leaves `fa` at the end -- which is exactly the checksum of everything before it. So a parse that verified the sum alone would accept a frame one byte short and read one parameter as two. The length must be checked, not inferred.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1562 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine `main` worktree, normalized message sets byte-identical in both directions, 0 attributable to the changed files.
- Paired pytest, identical 364-file selection (all of `tests/tools/` plus every guard that walks the tree, reads docstrings or reads source), the 4 changed test files excluded from both sides: **17 failed, 18050 passed, 68 skipped** on the branch against **17 failed, 18048 passed, 68 skipped** on base -- 17 identical failing IDs, both `comm` directions empty. All 17 need `awsiot` (`awsiotsdk`, declared in the `[mesh-iot]` extra and absent here; confirmed with `tomllib` + `find_spec`). The `+2` localizes entirely to `tests/test_args_docstring_completeness.py` (452 -> 454), which parametrizes over documented functions and picks up `read_motor_position`'s new `Args:`/`Returns:` blocks.
- `tests/tools/` whole: 3024 passed, 9 skipped.

### DIFF
--- a/changelog.d/2755-feetech-status-packet-framing.md
+++ b/changelog.d/2755-feetech-status-packet-framing.md
@@ -1,0 +1,38 @@
+### Fixed: a Feetech servo's reply is located and verified, instead of being read at fixed offsets
+
+`pose_tool`'s `read_motor_position` read the two position bytes from indices 5 and 6 of whatever
+the bus returned. The servo bus is half-duplex and shared by every motor on the arm, so a reply
+can arrive behind a byte the host's own transmission echoed, or be the late answer to a read that
+already timed out, sent by a different motor. Both shift those indices, and the result is a joint
+angle that is wrong rather than absent: a joint at 1024 counts (-89.98 degrees) was reported as
+-180.0 when one byte preceded its reply, and the tool quoted that as a measurement. A reply from
+another motor, a corrupted checksum and a frame one byte short were all accepted as well.
+
+The reply is now found and verified the way `scservo_sdk` -- the vendor SDK that owns this wire
+format, and which the `[lerobot]` extra already installs -- does it: the `FF FF` header is
+searched for rather than assumed at offset 0, the frame length is re-derived from `LEN`, the
+checksum is recomputed over the same span the outgoing packet signs, and the responding ID must be
+the one that was asked. Leading bytes are recovered from rather than refused, because bytes in
+front of the header do not make a reply corrupt, they make it offset -- the SDK reads the true
+1024 from exactly those bytes, and so does this now. A frame that cannot be verified reports no
+position, which is the signal every caller already handles: the `read_position` action answers
+with an error envelope instead of a number, and the interpolating paths decline to build a
+trajectory. Graded cell for cell against the SDK across nine framings, the two now agree
+everywhere.
+
+This was a motion fault and not only a reporting one. `_smooth_move` reads the current position
+before dividing the travel into increments, so a shifted reply made the arm interpolate from a
+place it was not: commanding a joint at 1024 counts toward 4095 wrote `[0, 511, 1023, 1535]`,
+opening a *smooth* move with a ninety-degree lurch away from the target. It now writes
+`[1024, 1279, 1535, 1791]`, byte for byte the trajectory a clean bus produces. `incremental_move`
+applied its delta to the same fiction and now refuses instead.
+
+Acting on a servo's error byte -- declining to drive an overheating motor -- is deliberately not
+decided here. A value inside the byte's range is a valid frame, as the SDK also holds, so a
+reported fault still reports its position; bounding the *value* to the register's 12-bit range is
+likewise a separate question from framing, and the suite pins that this parse answers neither.
+
+Nothing could have caught this: three test fixtures across three files built the reply with `0`
+where the checksum belongs and a hardcoded ID of 1, a frame no servo would send and one the vendor
+SDK calls corrupt. Each now carries a real checksum and answers as the motor the outgoing read
+addressed, so a misattributed position is visible to a test for the first time.
--- a/strands_robots/tools/pose_tool.py
+++ b/strands_robots/tools/pose_tool.py
@@ -8,6 +8,7 @@ This tool provides comprehensive pose management for robotic arms, including:
 - Safety checks and validation
 - Integration with LeRobot and serial communication
 - Pose interpolation and smooth transitions
+- Framed decoding of servo status replies
 """
 
 import json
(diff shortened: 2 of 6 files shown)


yinsong1986 (approved):
## Summary

Replaces `read_motor_position`'s fixed-offset read of a Feetech status reply (`response[5] | response[6] << 8`) with `_parse_status_packet`, which searches for the `FF FF` header, re-derives the frame length from `LEN`, recomputes the checksum over the same span the outgoing packet signs, and requires the responding ID to match the motor asked. An unverifiable reply reports `None` — the existing "could not read" signal — so no caller contract changes, and a frame arriving behind echoed bus bytes is recovered rather than refused (matching `scservo_sdk`'s behaviour on the identical bytes). I independently re-executed the parse against every row of the PR's before/after table plus the truncation-coincidence case (`FF FF 01 04 00 00 FA` — the trailing `0xFA` happens to be the checksum of everything before it) and a leading-`0xFF` noise byte that fuses with the real header; all behave as claimed. The helper is private and un-exported, the length guard precedes all frame indexing (no crash path), the loop advances `start` unconditionally (no infinite loop), and the error-byte / value-domain questions are explicitly left undecided and pinned as such.

## What's good

- The three test fixtures that made this untestable (checksum `0`, hardcoded ID 1) are fixed in the same PR, and the fakes now answer as the motor the outgoing packet addressed — a misattributed position is visible to a test for the first time.
- The new suite grades its own fixtures against `scservo_sdk` (`importorskip`-gated) before asserting on them, so the builder and parser cannot pass by sharing a misunderstanding of the wire format.
- Changelog fragment under `changelog.d/`, `Args:`/`Returns:` blocks on the touched surfaces, ASCII-only log strings, and the discarded bytes are logged so a bus fault stays diagnosable — all per AGENTS.md conventions.
- Scope discipline: `read(10)` deliberately unchanged, error-byte policy and 12-bit value bounding explicitly deferred with over-reach tests pinning that the parse does not quietly decide them.

PR #1766, opened 2026-07-30
fix(sim): scope a robot's actuators by transmission, not by a raw id

cagataycali:
Closes #1765.

## The defect

`SimRobot.actuator_ids` is re-derived after every recompile. It was built by comparing each actuator's `actuator_trnid[i, 0]` against the robot's joint ids:

```python
for i in range(new_model.nu):
    jnt_id = new_model.actuator_trnid[i, 0]
    if jnt_id in robot.joint_ids:
        robot.actuator_ids.append(i)
```

That field holds a **joint** id only when the transmission is `mjTRN_JOINT` / `mjTRN_JOINTINPARENT`. For a tendon, site, slider-crank or body transmission it indexes a different table, and those id spaces each start at 0 - so the comparison is between two different id spaces and matches an unrelated entity that merely shares a number.

A fixed tendon coupling a gripper's fingers is the standard MJCF idiom for a coupled gripper (the Menagerie Panda hand and the Robotiq 2F-85 both use one), so this is reachable with stock assets rather than only with exotic models.

## Measured (mujoco 3.11.0, at `32dc3f5b`)

One scene, two registry robots, differing only in the order they were added:

| added | `arm2` (so101_follower) owns | `panda` owns |
|---|---|---|
| panda, then arm2 | its own six | all eight - correct **by luck** |
| arm2, then panda | its own six **+ `panda/actuator8`** | seven - **missing its own gripper** |

The second row is what an incrementally built scene produces. `panda/actuator8` is the tendon drive; its `trnid[0]` is `0`, which equals `arm2`'s first joint id.

Three consequences, all on documented paths:

1. **The gripper becomes unoperable, and the error blames the model.** `set_gripper` resolves its actuator from `actuator_ids`, so with the gripper missing it reports `Registry gripper metadata for 'panda' ... names actuators ['actuator8'] but none exist in the model` - for an actuator that is plainly in the model. The jaw stays shut (finger gap `0.0000` m).
2. **A robot advertises an actuator that moves a different machine.** An unactuated inline arm added before the panda reports `actuator_ids == [7]`, i.e. `panda/actuator8`.
3. **`actuate_robot` refuses a robot that has no actuators.** Its double-actuate guard made the same raw comparison, so it answered `robot 'arm' already has actuators mapped to its joints; refusing to double-actuate` for a robot declaring none - making the documented way to add position servos to a URDF robot impossible.

[...]

PR #2756, opened 2026-08-26
fix(drivers): read the LiDAR state fields LidarState_ actually declares

cagataycali:
## Problem

`G1Driver._on_lidar_state` decoded `rt/utlidar/lidar_state` by reading `msg.code` and
`msg.freq`. `LidarState_` declares neither. The MID-360 reports its fault code as
`error_state` and its scan rate as `cloud_frequency`, so two of the three readings the
record exists to carry were being taken from names the message never had.

Nothing about that failed. Each read went through `getattr(msg, name, default)`, and a
name the IDL does not declare yields the default rather than raising - a well-formed
value that lands in the published record indistinguishable from a measurement.

## Measurement

Feeding a real `unitree_sdk2py` `LidarState_` describing a unit that has faulted while
scanning normally, and reading what `Mesh._read_lidar_state()` publishes:

| the message carries | main publishes | this branch publishes |
|---|---|---|
| `error_state = 3` | `code = -1` | `code = 3` |
| `error_state = 3` | `code_text = '-1 (unknown)'` | `code_text = '3 (unknown)'` |
| `cloud_frequency = 10.0` | `freq = 0.0` | `freq = 10.0` |
| `sys_rotation_speed = 10.0` | `sys_rotation_speed = 10.0` | `sys_rotation_speed = 10.0` |

`LidarState_` declares 17 fields; `code` and `freq` are not among them, and
`sys_rotation_speed` is the one reading in this record that was already a declared field
and therefore real. So `strands/<peer>/lidar/state` carried `-1` and `0.0` on every tick
of a healthy robot, and a unit whose LiDAR had genuinely faulted published exactly the
same record. `-1` is the value a consumer has to read as "nothing measured yet", which is
why the failure presents as a peer that never reports rather than as a peer reporting
something wrong.

`mesh/sensors.py:552` publishes this record and `mesh/core.py:1104` advertises the topic,
so both values reach the wire.

## Why the existing test could not catch it

`test_lidar_state_decodes_code` built its stand-in by hand as
`SimpleNamespace(code=0, freq=10.0, sys_rotation_speed=10.0)` - the names the decoder
happened to read. A double that spells whatever the code asks for agrees with the code
whatever those names are, so the field names were outside everything the suite measured.
That double now spells the declared names, which is why this PR touches it.

## Change

`error_state` and `cloud_frequency`, read from the message. `error_state` is read once and
feeds both the numeric code and its rendered text, so the two cannot drift into
describing different fields.

The text still renders through `decode_code`, which is worth justifying rather than
assuming: that table holds SDK RPC and loco/arm response codes, not LiDAR faults. Every
entry in it other than success is numbered above 255, and `error_state` is declared
`uint8`, so for this field the table can only ever report success for zero or fall back to
the bare integer. It cannot invent a meaning for a LiDAR fault. That disjointness is
pinned as a premise rather than left as a reading of the table.

## Tests

`tests/drivers/test_g1_lidar_state_reads_the_declared_fields.py`, 17 cells in five
classes. The SDK that owns the IDL installs from a git clone rather than PyPI, so it
cannot be a test dependency; the suite is built in three layers so the rule holds anyway:

- A faithful double carrying exactly the declared field names and nothing else. Reading an
  undeclared name off it produces the default, so these cells grade the decoder on an
  install with no SDK at all.
- `_DECLARED_LIDAR_STATE_FIELDS`, a frozen copy of the declaration, plus a cell that
  checks it against the real `LidarState_` wherever the SDK *is* importable. That is what
  keeps the double faithful as the IDL moves, and it skips rather than passing vacuously.
- A derivation over the decoder's own source: every literal name it reads with `getattr`
  must be a declared field, so a name added later is graded on arrival without anyone
  remembering to add a case.

Pre-fix split, source reverted and both test files kept: **8 failed, 56 passed**, and the
8 are the 5 regression cells, the 2 derivation cells and the repaired existing test.
**0 of 5 premise cells, 0 of 4 control cells, 0 of 1 fidelity cell** fail, so nothing in
this suite is asserting the fix into existence.

### Mutation table

Every row is a mutation of the merged tree, run against both test files.

| mutation | fires | note |
|---|---|---|
| control, no mutation | 0 | |
| the whole fix reverted | 8 | the pre-fix split |
| only `error_state` reverted to `code` | 7 | |
| only `cloud_frequency` reverted to `freq` | 5 | |
| the code and its rendered text read different fields | 2 | the single read is load-bearing |
| `code` reads a *declared but wrong* field (`cloud_size`) | 5 | passes the derivation, caught behaviourally |
| the frozen declaration drops `error_state` | 2 | |
| the double made permissive (declares `code`/`freq` too) | 4 | the double's fidelity is load-bearing |
| the repaired existing double reverted to the undeclared names | 1 | |

8 distinct failing-id sets across 8 mutations, none undetected, and
`{only error_state} union {only cloud_frequency}` is exactly the full-revert set
(7 and 5 giving 8), so each field name is independently pinned. `collected` is 64 on every
row, so no mutation is hidden by a deselection. The row worth reading is the declared-but-wrong
field: the derivation accepts it because `cloud_size` really is declared, and only the
behavioural cells catch it - neither layer is sufficient alone.

## Gate

  (1567 files); `format` changed nothing.
- `mypy strands_robots tests tests_integ`: **24 errors on this branch and 24 on a pristine
  `upstream/main` worktree**, normalized message sets byte-identical in both directions,
  0 in the changed files.
- Paired run over an identical 492-file selection (all of `tests/drivers` and `tests/mesh`
  plus every guard that walks the tree, reads source or docstrings, or names the driver
  seam), the new file excluded from both sides and every path verified present on both:
  `collected 20812` and **23 failed, 20698 passed, 67 skipped, 24 errors on both trees**,
  with 47 identical failing/erroring ids and an empty set difference in both directions.
  Every one of those 47 is `awsiot` (the `[mesh-iot]` extra, absent here) or one of three
  failures present identically on both trees.
- `tests/mesh/test_zenoh_transport_security.py` run alone on each tree, as it needs
  isolation: 9 passed, 2 skipped on both.
- Re-verified after rebasing onto the five commits that landed while this was in progress;
  the mutation table is unchanged on the new base.

## Artifact

One capture script, run against both trees, feeding the same real `LidarState_` to
`_on_lidar_state` and then reading `Mesh._read_lidar_state()`. Every claim in the figure is
asserted against the two captures before it is drawn.

published lidar state record, before and after

### Round N - held as draft: two premise fixtures invalidated by this change

Converted to draft by the author. Not a review request - this PR is correct and
must not merge in its current shape, and it was APPROVED with only the required
check outstanding, so a green check would have made it read CLEAN and mergeable.

`#2754` merged after this branch's last verification and added
`tests/drivers/test_g1_driver_reaches_the_mesh_wire.py`, whose lidar-state
fixture is built as `SimpleNamespace(code=0, freq=np.float32(10.0), ...)` - the
undeclared names this PR removes. Composed against `main` at `59b116eb`:

```
tests/drivers/  on this head + main   ->  1 failed, 81 passed, 1 skipped
FAILED test_g1_driver_reaches_the_mesh_wire.py::
       TestTheLidarStateTopicComesWithTheSummary::
       test_the_state_record_carries_the_decoded_status_code
       assert 0.0 == 10.0 +/- 1.0e-05
```

`#2752` (open, approved, green) carries a second one -
`test_g1_lidar_summary_describes_the_cloud.py::TestTheRestOfTheDriverIsUnchanged::
test_the_lidar_state_decoder_still_populates` - which fails identically once this
lands. Composing all three gives `2 failed, 98 passed, 1 skipped`.

Neither was visible to any gate on this PR. `git merge-tree` reports **no
conflict** against `#2752`, `mergeable` read `MERGEABLE`, and
`Detect an untested overlap with the base branch` passed - it ran before `#2754`
existed. This is the `#1763`/`#1766` topology: two branches whose first shared
tree is `main`.

Both failures are the same class and neither contradicts this change. A fixture
that spells the name the decoder happens to read agrees with the decoder whatever
that name is - the point this PR's own "Why the existing test could not catch it"
section makes about `test_lidar_state_decodes_code`. These are two further
instances of it, in files that did not exist when that section was written. Per
AGENTS.md, an invalidated premise test is replaced rather than deleted: the
conclusion each cell supports still holds, reached through the declared names.

Sequencing, chosen to spend one review round rather than two:

1. `#2752` merges first - it is green against `main` and independent
   (`84 passed` composed).
2. This branch then absorbs `main` and re-pins **both** fixtures to
   `error_state` / `cloud_frequency` in one commit, and comes out of draft.

Patching only the `#2754` fixture now would cost a second round when `#2752`
lands. No production code changes; the mutation table above is unaffected.

### Round 2 - out of draft: both premise fixtures re-pinned

`#2754` merged at 07:20 UTC and `#2752` at 07:52 UTC, so the precondition the
plan above named is met. This branch has absorbed `main` (`eb378603`, no
conflict) and re-pinned **both** fixtures in one commit, as sequenced - one
review round rather than two.

Composed on the merged tree, before the repair commit:

```
tests/drivers/  ->  2 failed, 98 passed, 1 skipped
```

which is exactly the two cells and the composition predicted above. After the
repair commit:

```
tests/drivers/  ->  100 passed, 1 skipped
```

Both doubles moved from `SimpleNamespace(code=..., freq=...)` to `error_state`
and `cloud_frequency`. Replaced, not deleted: each cell still supports the
conclusion it was written for, reached through the names `LidarState_` declares.

The repair is load-bearing rather than merely agreeing with the change. Reverting
the production fix to the undeclared names:

| tree | cells failing on a reverted fix |
|---|---|
| pre-merge (the description's mutation table) | 8 |
| merged, both fixtures re-pinned | 10 |

The two repaired cells are exactly that delta, so they now grade the field names
instead of spelling whatever name the decoder happens to read - the failure mode
this PR exists to close, which is why re-pinning them is the fix and deleting
them would not have been. The mutation table above is otherwise unaffected: no
production code changed in this round.

Gates on the merged tree: `ruff check` and `ruff format --check` clean over
`tests/drivers/`, `mypy tests/drivers/` reports no issues in 4 source files,
100 passed / 1 skipped.

Out of draft.


yinsong1986 (dismissed):
## Summary

Correctness fix in `G1Driver._on_lidar_state`: the decoder was reading `msg.code` and `msg.freq`, neither of which `LidarState_` declares, so `getattr`'s defaults (`-1` / `0.0`) were published on every tick as if they were measurements - a faulted MID-360 was indistinguishable from a healthy one on `strands/<peer>/lidar/state`. The fix reads the declared names (`error_state`, `cloud_frequency`), and the frozen 17-field copy of the `LidarState_` declaration in the new test file matches the IDL as `unitree_sdk2py` ships it. The published record's keys (`code`, `code_text`, `freq`, `sys_rotation_speed`, `t`) are unchanged, so this is not a wire-format change - only the values stop being constants, and since they were constants before, no consumer could have depended on the old behaviour. `error_state` is read once and feeds both the numeric code and its rendered text, so the two cannot drift apart, and the `ERR_CODES` bounds check (`in_range == {0}`) pins that the RPC-code table cannot mislabel a `uint8` fault.

## What's good

- The three-layer test design (faithful double with exactly the declared fields, frozen declaration graded against the real SDK where importable, AST derivation over the decoder's own `getattr` reads) closes the class of bug rather than the instance - a name added to the decoder later is graded on arrival, and the double cannot re-agree with a wrong decoder the way the old hand-built stand-in did.
- The pre-existing test that masked the bug was fixed in place with a docstring explaining why it could not have caught this, per AGENTS.md's "pin every reviewed fix with a regression test".
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; all new strings are ASCII; test names describe behaviour, not provenance; the `Mesh` used in the mesh-topic test is constructed inert (no session, no threads), consistent with existing unit tests.

## Verification suggestions

- `pytest tests/drivers/test_g1_lidar_state_reads_the_declared_fields.py tests/drivers/test_g1_driver.py` on the branch (CI covers this; the SDK-fidelity cell skips without `unitree_sdk2py`).
- On any box with `unitree_sdk2py` installed from its git clone, run `pytest tests/drivers/test_g1_lidar_state_reads_the_declared_fields.py::TestTheFrozenDeclarationIsTrue` to confirm the frozen declaration against the real IDL - that cell is the one CI cannot exercise.

yinsong1986 (approved):
## Summary

Correctness fix in `G1Driver._on_lidar_state`: the decoder read `msg.code` and `msg.freq`, neither of which `LidarState_` declares, so `getattr`'s defaults (`-1` / `0.0`) were published on `strands/<peer>/lidar/state` on every tick as if they were measurements - a faulted MID-360 was indistinguishable from a healthy one. The fix reads the declared names (`error_state`, `cloud_frequency`), with `error_state` read once so the numeric code and its rendered text cannot drift onto different fields. The published record's keys (`code`, `code_text`, `freq`, `sys_rotation_speed`, `t`) are unchanged, so this is not a wire-format change - only the values stop being constants. Since the prior (dismissed) review, the branch absorbed `main` and re-pinned the two lidar-state doubles that #2752/#2754 landed with the old undeclared names, exactly as the round notes sequenced. Verified at head `15d07d14`: the four touched test files pass (100 passed, 1 skipped - the SDK-fidelity cell skips without `unitree_sdk2py`, by design), and reverting just the two field names in `g1.py` fires 10 failures across all four files, matching the round-2 mutation claim - so the re-pinned fixtures now grade the field names rather than agreeing with whatever the decoder reads. `ruff check` / `ruff format --check` clean on the changed files, all new strings ASCII, changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention, and the mesh-topic test constructs `Mesh` inert (no session, no threads started before `start()`; `tests/conftest.py` sets `STRANDS_MESH=false` suite-wide).

## Verification suggestions

- On any box with `unitree_sdk2py` installed from its git clone, run `pytest tests/drivers/test_g1_lidar_state_reads_the_declared_fields.py::TestTheFrozenDeclarationIsTrue` - that cell grades the frozen 17-field declaration against the real IDL and is the one CI cannot exercise.

PR #361, opened 2026-06-06
test(mesh): pin estop lockout check-then-set inside _estop_replay_lock (closes #273)

cagataycali:
## Summary

Closes #273. The e-stop lockout check-then-set (`_estop_lockout.set()` + the `_last_estop_ts` / `_last_estop_mono` writes) **already lives inside `self._estop_replay_lock`** on `main` (the Option-A fix from the issue), with the audit emission moved outside the lock using a `lockout_was_engaged` snapshot taken under the lock. The invariant was, however, **unpinned** -- nothing guarded against a future refactor moving the engage block back out of the lock, and there was no concurrent-estop race test.

This PR is **test-only** (plus a CHANGELOG note). It does not touch `strands_robots/mesh/core.py`.

## What it pins (`tests/mesh/test_estop_lockout_race.py`)

1. **Deterministic two-thread race** -- two distinct-issuer estops fired concurrently, with a forced-interleave barrier instrumenting `_estop_lockout.is_set`, must produce exactly **one** `remote_estop_engaged` + **one** `remote_estop_redundant` audit event (never two engages). The barrier can only trip if the read happens outside the lock, so the lock-protected code lets it time out harmlessly while the buggy path double-engages.
2. **Source-text lock-containment pin** -- asserts `self._estop_lockout.set()` sits inside the `with self._estop_replay_lock:` block (indent-based; the reliable anti-refactor guard given a true data race is non-deterministic under the GIL).
3. **Timestamp atomicity pin** -- `_last_estop_ts` and `_last_estop_mono` are written together in the same lock-guarded engage branch, so a later corroboration-window check never observes a mixed (one-updated/one-stale) pair.

## Pre-fix verification

Against a simulated pre-fix tree (engage block moved outside the lock):

| test | pre-fix | post-fix |
|------|---------|----------|
| forced-interleave race | FAIL (2x `remote_estop_engaged`) | PASS |
| lock-containment source pin | FAIL | PASS |
| timestamp atomicity pin | PASS* | PASS |

*The atomicity pin pins co-location of the timestamp writes with `set()`; the lock-containment pin is the one that catches the #273 regression.

## CI parity

- `ruff==0.15.16 check` + `ruff format --check`: clean
- `mypy strands_robots tests tests_integ`: clean for this file (project-introduced findings only)
- `pytest tests/mesh/test_estop_lockout_race.py`: 3 passed
- Run alongside `test_estop_replay.py` + `test_corroboration_clock_domain.py`: 13 passed (no isolation leak)
[...]

PR #2757, opened 2026-08-26
feat(g1): add DDSPublisher for the write path (issue #361, PR-A)

cagataycali:
# PR-A of the harness#361 stack: transport primitive for the G1 write path

**Stack**
- **PR-A (this one)**: `DDSPublisher` sibling of `DDSSubscriberSet` in `_dds_engine.py`. Transport primitive only. Driver still refuses motion.
- PR-B (next): `send_action` wired on top of PR-A.
- PR-C (last): control loop @ 500 Hz on top of PR-B, per-step FSM re-gate, zero-torque on stop.

All three are decoupled from #358 (agent-tool vocabulary): a caller only needs the raw `rt/lowcmd` publish primitive, not the 48 @tool verbs. The lazy import keeps `unitree_sdk2py` off module load, so Thor and CI still import the driver.

## What this PR does

Adds `DDSPublisher` alongside `DDSSubscriberSet` in `strands_robots/tools/g1/_dds_engine.py`:

```python
from strands_robots.tools.g1._dds_engine import DDSPublisher

pubs = DDSPublisher("eth0")
pubs.start()                                        # ChannelFactoryInitialize under the lock
err = pubs.publish("rt/lowcmd", LowCmd_, message)   # None on success, string on error
```

Design mirrors `DDSSubscriberSet`:

- **Shared construction lane**: constructs `ChannelPublisher(...)` under the same `_DDS_INIT_LOCK` the subscriber set holds. The CycloneDDS bindings segfault under concurrent construction; one lock, both classes.
- **Lazy SDK import**: `unitree_sdk2py.core.channel.ChannelPublisher` is imported inside `get_publisher`, never at module load. Grep-scale test (`test_module_does_not_import_the_sdk_at_load_time`) pins this.
- **Publisher cache**: keyed by `(topic, message_class)`. A 500 Hz control loop asking for the `rt/lowcmd` publisher every step constructs it **once** — asserted by `test_get_publisher_constructs_once_per_key`.
- **String error envelopes**: same shape as `DDSSubscriberSet.subscribe` — publish never raises to its caller.
- **Idempotent `start()` and `close()`**: same shape as the subscriber set.

## Why now

harness#361 is P0 for the office. The driver's control loop needs one transport primitive: publish `unitree_hg.msg.dds_.LowCmd_` on `rt/lowcmd`. This PR ships that primitive in isolation, so PR-B (`send_action` wired) and PR-C (control loop) each stay small and independently reviewable.

## Verification (headless, no G1 hardware)

- **12 new tests**, every SDK touch mocked with a recording `_FakeChannelPublisher`
- `pytest tests/tools/g1/test_dds_publisher.py tests/drivers/test_g1_driver.py tests/test_driver_seam.py --no-cov` → **104 passed**
- `ruff check` + `ruff format --check` clean on both files
- `mypy` scoped-clean (0 errors in the two changed files; the 7 pre-existing errors on `mesh/core.py` etc. are unchanged from `main`)
- Grep-scale invariant: no `import unitree_sdk2py` at module load, asserted by test
- No changes to the driver's public surface: `G1Driver.send_action` still returns `motion path not wired yet (issue #358)` because this PR does not touch the driver

## Break table (what would flip red)

| mutation | what it breaks |
|---|---|
| Remove `_DDS_INIT_LOCK` from `get_publisher` | `test_publisher_and_subscriber_share_the_init_lock` fails (would race the segfault) |
| Skip the `(topic, message_class)` cache | `test_get_publisher_constructs_once_per_key` counts 2, expects 1 |
| Add `import unitree_sdk2py` at module top | `test_ddspublisher_is_importable_without_the_sdk` and `test_module_does_not_import_the_sdk_at_load_time` both fail |
| Let `publish` raise on `Write` error | `test_publish_returns_error_when_write_raises` fails |
| Skip idempotency on `close` | `test_close_is_idempotent_and_drops_cache` fails |
| Remove `pub.Init()` after construct | `test_get_publisher_constructs_once_per_key` asserts `init_calls == 1` |

## Files changed

| file | lines | why |
|---|---|---|
| `strands_robots/tools/g1/_dds_engine.py` | +154 / -5 | Add `DDSPublisher` class; update module docstring to describe both halves |
| `tests/tools/g1/test_dds_publisher.py` | +366 (new) | 12 tests, all SDK touches mocked |
| `changelog.d/2760-g1-dds-publisher.md` | +18 (new) | Release note |

Total: **3 files, +538 / -5**. Under the 400 loc-of-code target for a driver primitive PR.

---

Closes part of #361 (unblocks PR-B). Decoupled from #358. No hardware needed to verify.

### DIFF
--- a/changelog.d/2760-g1-dds-publisher.md
+++ b/changelog.d/2760-g1-dds-publisher.md
@@ -1,0 +1,24 @@
+### Added: `DDSPublisher` for the G1 write path (write half of the DDS engine)
+
+The G1 native driver's DDS engine has grown the sibling class the write
+path needs. `DDSSubscriberSet` already funnels the read path through
+`_DDS_INIT_LOCK` and a lazy SDK import; `DDSPublisher` mirrors both, and
+adds a `(topic, message_class)` cache so a control loop asking for the
+`rt/lowcmd` publisher every step constructs it once:
+
+```python
+from strands_robots.tools.g1._dds_engine import DDSPublisher
+
+pubs = DDSPublisher("eth0")
+pubs.start()                              # ChannelFactoryInitialize under the lock
+err = pubs.publish("rt/lowcmd", LowCmd_, message)   # None on success, string on error
+```
+
+The class does not import `unitree_sdk2py` at module load, so the driver
+still loads on Thor and CI. That is the invariant the shared lock and the
+lazy import together buy: one construction lane for the entire G1 DDS
+layer, and no SDK on module load.
+
+This ships the transport primitive that issue #361 (send_action wired) and
+its follow-up (control loop at 500Hz) will consume; the driver itself
+still refuses motion until those PRs land.
--- a/strands_robots/tools/g1/_dds_engine.py
+++ b/strands_robots/tools/g1/_dds_engine.py
@@ -1,14 +1,21 @@
-"""Subscription helpers for the G1 driver's read path.
+"""Subscription and publication helpers for the G1 driver's read and write paths.
 
 The driver holds one :class:`DDSSubscriberSet` and asks it for background
 subscribers whose callbacks fill in-memory caches (the newest message wins).
 The mesh publishes those caches at its own cadence, so the DDS callback stays
 fast: parse, drop into a slot, return.
 
-The engine deliberately does not surface a publisher API - writing to
-``rt/armsdk`` and ``rt/lowcmd`` is the arm-SDK client's job (issue #358), and
-mixing the two here would tempt callers into ad-hoc writes that skip the FSM
-gate. Read-only, and small.
+The driver also holds one :class:`DDSPublisher` for the ``rt/lowcmd`` write
+path (and any other topic the control loop grows in issue #361). Publishers
+are cached per ``(topic, message_class)`` and constructed under the shared
+:data:`_DDS_INIT_LOCK`; that is the same lock the subscriber set holds, so a
+reader and a writer never construct concurrently on the CycloneDDS bindings
+that segfault under it.
+
+Neither class imports ``unitree_sdk2py`` at module load: the SDK is loaded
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Adds `DDSPublisher` as the write-path sibling of `DDSSubscriberSet` in the private module `strands_robots/tools/g1/_dds_engine.py`, plus a 12-test suite and a changelog fragment. The class mirrors the subscriber set faithfully: construction under the shared `_DDS_INIT_LOCK` (one construction lane against the CycloneDDS concurrent-construction segfault), lazy `unitree_sdk2py` import inside `get_publisher` so the module stays importable on headless boxes, string-error envelopes that never raise to the caller, and idempotent `start()`/`close()`. The one addition beyond the mirror is the `(topic, message_class)` publisher cache with a double-check under the lock, which is what a 500 Hz control loop needs. The two-lock split over the cache (`_DDS_INIT_LOCK` for construction, `self._lock` for `close()`) is byte-for-byte the pattern the subscriber set already carries on `main`, and its worst-case interleaving is benign because `close()` is non-terminal by design — the tested contract is that the next `get_publisher` reconstructs. No public API, wire format, or env var is introduced (`_dds_engine` is underscore-private), the driver's motion refusal is untouched, and the diff matches the PR description exactly.

## What's good

- Every SDK touch mocked via `monkeypatch.setitem(sys.modules, ...)` per AGENTS.md's sys.modules-restoration rule; tests are named for behaviour, not provenance.
- The lazy-import invariant is pinned twice (namespace scan + source scan), and the shared-lock invariant is pinned as object identity rather than by name.
- `except Exception` sites carry the documented SDK-raises-bare-Exception justification, matching the subscriber set's reviewed shape.
- Scope discipline: transport primitive only, with `send_action` and the FSM re-gate deferred to PR-B/PR-C as the stack plan states.

PR #2758, opened 2026-08-26
fix(mesh): describe the grammar that refused a name, per seam

cagataycali:
## Problem

`MobileBaseRobot` validates two independently-overridable seams and describes both with one sentence:

```python
_NAME_RE:   re.Pattern[str] = ...   # node_name
_TOPIC_RE:  re.Pattern[str] = ...   # every topic and init_services service name
_NAME_HINT: str = " (expected a graph name like /turtle1/cmd_vel)"   # appended to BOTH refusals
```

That holds only while a platform's two grammars agree. On one platform they do not. `RtpsRobot` writes to a DDS topic directly, so a topic must be absolute; its `node_name` names this robot's own agent tools rather than anything on the robot's graph and keeps the base grammar:

| class | `_NAME_RE` | `_TOPIC_RE` | same |
|---|---|---|---|
| `MobileBaseRobot` | `^[A-Za-z0-9_/~]+$` | `^[A-Za-z0-9_/~]+$` | yes |
| `RosBridgedRobot` | `^[A-Za-z0-9_/~]+$` | `^[A-Za-z0-9_/~]+$` | yes |
| `RtpsRobot` | `^[A-Za-z0-9_/~]+$` | `^/[A-Za-z0-9_/]*[A-Za-z0-9_]$` | **no** |

No single sentence is true of both `RtpsRobot` seams, so the platform's only escape was to silence the hint entirely (`_NAME_HINT = ""`). The strictest grammar in the tree became the one that said nothing about itself.

### Measured, before

Three values `RosBridgedRobot` accepts because rclpy resolves them, and `RtpsRobot` refuses:

| topic | `RosBridgedRobot` | `RtpsRobot` reports |
|---|---|---|
| `cmd_vel` | accepted | `invalid cmd_vel_topic: 'cmd_vel'` |
| `~/cmd_vel` | accepted | `invalid cmd_vel_topic: '~/cmd_vel'` |
| `/cmd_vel/` | accepted | `invalid cmd_vel_topic: '/cmd_vel/'` |

The caller is told the value is invalid and nothing about why. The rule was already written down twice in prose - the comment above `_RTPS_TOPIC_RE`, and the docstring of `test_shipped_classes_keep_their_own_name_grammar` - and withheld from the one party who needed it.

The stated reason for silencing the hint does not survive measurement. The comment read *"`use_rtps` requires absolute topics, so the generic hint would misdirect"*, but the generic hint's own example is a topic `RtpsRobot` accepts:

```
_RTPS_TOPIC_RE.match("/turtle1/cmd_vel")            -> True
RtpsRobot("rtps", "/turtle1/cmd_vel")               -> ACCEPTED
```

The sentence was not misdirecting. It was attached to a constant the other seam also had to live with.

## Change

Give the topic seam its own hint, and pass the hint into `_check` rather than reading one off the class, so a seam cannot borrow another seam's sentence:

```python
_NAME_HINT:  str = " (expected a graph name like /turtle1/cmd_vel)"
_TOPIC_HINT: str = " (expected a graph name like /turtle1/cmd_vel)"

def _check(cls, label, value, pattern, hint) -> str: ...
def _check_name(cls, label, value):  return cls._check(label, value, cls._NAME_RE,  cls._NAME_HINT)
def _check_topic(cls, label, value): return cls._check(label, value, cls._TOPIC_RE, cls._TOPIC_HINT)
```

`RtpsRobot` now states the absolute requirement and why, and **drops** its empty `_NAME_HINT` override so `node_name` regains the sentence that was always correct for it. `RosBridgedRobot` overrides one grammar for both seams, so it declares one sentence for both.

### Measured, after

```
cmd_vel     -> invalid cmd_vel_topic: 'cmd_vel' (expected an absolute topic like
               /turtle1/cmd_vel - use_rtps writes to DDS directly, so a relative
               or ~ name has no resolver)
bad name    -> invalid node_name: 'bad name' (expected a graph name like /turtle1/cmd_vel)
```

Unchanged: every grammar; `RosBridgedRobot`'s two messages byte-for-byte; the base's message; `RosBridgedRobot("tb", "relative/cmd_vel", "/odom")` still accepted; `RtpsRobot("tb", "/turtle1/cmd_vel")` still accepted.

### The deleted `_check`

`rtps_robot` also shipped a module-level `_check` with **no caller anywhere in the tree** - a divergent copy of the base classmethod whose refusal omitted the hint unconditionally. Leaving a second, hint-free copy of the shared validator in the one module that had gone silent is the same defect one edit from returning, so it goes with this change.

`test_shipped_classes_keep_their_own_name_grammar` pins the grammar divergence behaviourally, and `test_rtps_robot.py::test_invalid_topic_rejected_at_construction` pins the refusal. Both match on `"invalid cmd_vel_topic"` - the prefix, which stops immediately before the hint. No test in the tree referenced `_NAME_HINT` at all.

Measured: across all ten mutations below, the six pre-existing suites covering these classes (`test_mobile_base`, `test_rtps_robot`, `test_ros_bridge`, `test_rosbridge_robot`, `test_ackermann_robot`, `test_repr_survives_partial_construction` - 245 tests) report **zero** failures.

## Tests

`tests/mesh/test_mobile_base_hint_describes_its_own_seam.py`, 35 cases. The seam table is read out of the bodies of the base's own `_check_*` methods and the platform list by walking the package, so a third seam or a fourth platform is graded on arrival rather than needing the file edited.

The derived rules: every seam forwards a hint of its own; no two seams share one hint attribute; no platform's hint is empty; **each hint offers an example that the pattern it describes actually accepts**; a platform whose two grammars diverge does not describe them with one sentence; and no other mesh module redefines `_check`.

Pre-fix split, 25 failed / 10 passed:

| class | failed | tests |
|---|---|---|
| `TestThePopulationIsDerivedAndBothBranchesAreReached` | **0** | 4 |
| `TestEverySeamCarriesAHintForItsOwnGrammar` | 14 | 4 |
| `TestADivergentPlatformSaysSoTwice` | 1 | 1 |
| `TestTheRtpsTopicRefusalNamesTheAbsoluteRule` | 7 | 3 |
| `TestASharedGrammarKeepsASharedSentence` | 2 | 2 |
| `TestNoOverReach` | **0** | 5 |
| `TestOnlyTheBaseOwnsTheSharedCheck` | 1 | 2 |

Nothing in the premise class or the over-reach class fails pre-fix: every expectation there is one the pre-fix code also met.

### Mutations

Ten mutations, ten distinct failing-id sets, none undetected, control clean, source restored byte-identical. `sib` is the six pre-existing suites.

| row | mutation | fires | sib |
|---|---|---|---|
| b1 | all three source files reverted | 25 | 0 |
| b2 | `_check_topic` forwards `cls._NAME_HINT` (the shared constant, kept) | 5 | 0 |
| b3 | `RtpsRobot._TOPIC_HINT` silenced back to `""` | 9 | 0 |
| b4 | `RtpsRobot._TOPIC_HINT` set to the generic base sentence | 5 | 0 |
| b5 | `RtpsRobot` silences `_NAME_HINT` again | 3 | 0 |
| b6 | `_TOPIC_HINT` offers an example its own pattern refuses | 4 | 0 |
| b7 | `RosBridgedRobot._TOPIC_HINT` dropped | 2 | 0 |
| b8 | the caller-less module-level `_check` restored | 1 | 0 |
| b9 | `_check` reads `cls._NAME_HINT`, ignoring its `hint` parameter | 4 | 0 |
| b10 | `_TOPIC_HINT` drops the reason, keeps the rule | 1 | 0 |

b2 and b4 both fire 5 with different id sets: b2 is the original shared-constant arrangement, b4 is the mistake the deleted comment feared. b7 shows that letting a platform fall back to the base sentence is caught by the derived rule and by nothing else.

## Gate

Paired run, identical 500-file selection on both trees (all of `tests/mesh/` plus every guard that walks the tree, parses source, reads docstrings or names these modules), the new file excluded from both:

```
branch : 19940 collected, 22 failed, 19827 passed, 67 skipped, 24 errors  (654s)
base   : 19940 collected, 22 failed, 19827 passed, 67 skipped, 24 errors  (668s)
46 failing/erroring ids each, comm empty BOTH directions
```

44 of the 46 need `awsiot` (`awsiotsdk`, declared in the `[mesh-iot]` extra, absent here - confirmed with `tomllib` + `find_spec`); the other two are pre-existing on both trees. `tests/mesh/test_zenoh_transport_security.py` runs alone: 9 passed / 2 skipped on both.

```
ruff check strands_robots tests tests_integ   -> All checks passed!
ruff format --check                            -> 1574 files already formatted
mypy                                           -> no branch-only messages (normalized set diff empty)
```

After merging the current `main` (`59b116eb`, no file overlap with this change): 1165 passed across the new file, all six sibling suites, the docstring/export graders, the changelog graders and the two suites `main` added; b2 and b8 re-measured at 5 and 1 unchanged.

No visual artifact: this changes what a constructor refusal says, not policy behaviour, simulation, rendering, recording or any robot asset. The measured refusal strings and the mutation table are the evidence.


yinsong1986 (approved):
## Summary

This PR splits `MobileBaseRobot`'s single shared refusal hint (`_NAME_HINT`) into per-seam hints (`_NAME_HINT` for `node_name`, new `_TOPIC_HINT` for topics/services), threading the hint through `_check` as an argument so a seam can no longer borrow another seam's sentence. `RtpsRobot` — the one platform whose two grammars diverge — gains a topic hint that names the absolute-topic requirement and why (`use_rtps` writes to DDS directly, no resolver), and drops its empty `_NAME_HINT` override so the node-name seam regains the inherited sentence. `RosBridgedRobot` aliases the two hints since its grammars coincide. The dead module-level `_check` copy in `rtps_robot.py` (no callers, hint-free) is deleted. Every grammar and every pre-existing message is byte-for-byte unchanged, confirmed by the `TestNoOverReach` controls.

Head `571ab8ce` is a clean merge of `main` into the branch (combined diff of the merge commit is empty); all five reviewed files are byte-identical to `752ed9ff` and the PR's own diff against the merge base is unchanged at 5 files, +358/-13.

## What's good

- Verified locally at head `571ab8ce`: all 35 new tests pass, the 134 pre-existing `test_mobile_base.py` / `test_rtps_robot.py` / `test_ros_bridge.py` tests pass unchanged, and `ruff check` / `ruff format --check` are clean on all four touched files.
- `_check` is a private classmethod with exactly two callers, both updated in the same diff; the removed `rtps_robot._check` copy has zero callers on the PR head (`git grep` confirms). No public API, wire format, or persisted schema moves.
- The new grader derives its seam table from the base's `_check_*` bodies and its platform list by walking the package, includes non-vacuity premises (`TestThePopulationIsDerivedAndBothBranchesAreReached`, `test_the_scan_flags_a_constructed_second_copy`), and asserts each hint's example against the pattern it describes — a hint can't drift to advertise a value its own seam refuses.
- New `_check` docstring carries a complete `Args:`/`Raises:` block per AGENTS.md; hints are plain ASCII; changelog fragment present at `changelog.d/2758-mobile-base-hint-per-seam.md`; no host paths in the test file.

yinsong1986 (approved):
## Summary

This PR gives `MobileBaseRobot`'s two validation seams independent refusal hints: `_check` now takes the hint as a parameter instead of unconditionally reading `cls._NAME_HINT`, a new `_TOPIC_HINT` class attribute is forwarded by `_check_topic`, and `RtpsRobot` -- the one platform whose topic grammar (absolute-only, DDS-direct) diverges from its node-name grammar -- replaces its silenced `_NAME_HINT = ""` with a topic hint that states the absolute requirement and why. `RosBridgedRobot` aliases `_TOPIC_HINT = _NAME_HINT` since its two grammars coincide, keeping its messages byte-for-byte unchanged. The caller-less module-level `_check` in `rtps_robot.py` (a hint-free divergent copy of the base validator) is deleted; I confirmed it has no remaining callers at the head commit. No grammar changes: every `_NAME_RE`/`_TOPIC_RE` pattern is untouched, so nothing previously accepted is now refused -- this changes only what a constructor refusal says, which rules out the breaking-change and one-way-door buckets. Verified locally at `571ab8ce`: the new 35-test grader passes, the sibling suites (`test_mobile_base`, `test_rtps_robot`, `test_ros_bridge`, `test_repr_survives_partial_construction`, 183 tests) pass, the RTPS hint's own example (`/turtle1/cmd_vel`) is accepted by `_RTPS_TOPIC_RE` as the grader requires, and all four changed files are pure ASCII with no host paths. Changelog fragment present per AGENTS.md step 3; the new test derives its seam table and platform list from the code rather than hardcoding them, consistent with the repo's derived-population testing convention.

## What's good

- The grader is self-populating (seams read from `_check_*` bodies via AST, platforms discovered by walking the package) with explicit non-vacuity premise tests, so a third seam or new platform is graded on arrival.
- `TestNoOverReach` pins that every pre-fix acceptance still holds, and the mutation table in the description shows each derived rule fires independently.
- Deleting the divergent module-level `_check` closes the reintroduction path rather than leaving a dormant hint-free copy, and `TestOnlyTheBaseOwnsTheSharedCheck` pins the deletion.

PR #2759, opened 2026-08-26
fix(simulation): bound a rotated collision geom on the axes it occupies

cagataycali:
## Problem

`_geom_aabb` reports the axis-aligned box one `<geom>` occupies in its owning body's frame.
`_body_collision_aabb` unions those, and `load_mjcf_scene_objects` publishes the union as a
`SceneObject`'s `size` and `position` -- what that function's own docstring calls the "validated AABB
proxy", the physics footprint standing in for the object. It read `pos` and `size` and never asked
how the geom is turned.

MJCF gives a geom five mutually exclusive ways to state a rotation (`quat`, `euler`, `axisangle`,
`xyaxes`, `zaxis`). Dropping it does not make the bound approximate; for a quarter turn it reports
the extents on the wrong axes. Eight lines of MJCF through the public loader:

```xml
<body name="ruler" pos="0.4 0 0.1"><freejoint/>
  <geom type="box" size="0.30 0.02 0.02" euler="0 0 90"/>
</body>
```

| | `size` reported |
|---|---|
| MuJoCo (`geom_size` + `geom_quat`) | `(0.04, 0.60, 0.04)` |
| `load_mjcf_scene_objects` before | `(0.60, 0.04, 0.04)` |
| `load_mjcf_scene_objects` after | `(0.04, 0.60, 0.04)` |

15x too small on the axis the bar occupies and 15x too large on an axis it does not.

### The same function already read a rotation on its other channel

A capsule or cylinder may spell its placement with `fromto`, whose endpoints carry the axis, and
`_segment_aabb` bounds that exactly -- its docstring is explicit that it is "the exact bound rather
than the box of a rotated local box". So one geom got two different bounds depending on which
spelling was used for it, though MuJoCo compiles both to the same shape. Measured, one capsule from
`(0, 0, -0.1)` to `(0.1, 0, 0.1)` of radius 0.02:

| spelling | half-extent reported |
|---|---|
| `fromto` | `(0.07, 0.02, 0.12)` |
| `pos` + `zaxis` + `size` | `(0.02, 0.02, 0.132)` |
| sampled from the compiled geom's surface | `(0.07, 0.02, 0.12)` |

MuJoCo compiles the two to the same position, the same `geom_size` and parallel axes, which is
pinned as a premise -- if it saw two shapes, two bounds would be correct.

## Change

Each primitive now gets its exact bound:

- **box** -- the support function, `sum_j |R[i][j]| * half[j]`.
- **ellipsoid** -- `sqrt(sum_j (R[i][j] * half[j])**2)`, which is tighter than its box, so the proxy
  is not inflated by the reader's choice of formula.
- **capsule / cylinder** -- the endpoints a `fromto` would have carried are rebuilt from the rotated
  local +z axis and handed to `_segment_aabb`, so the two spellings share one answer instead of
  drifting. At identity this reproduces the previous result exactly.
- **sphere** -- unchanged; no rotation moves a ball.
- **`fromto`** -- deliberately still rotation-free. MuJoCo derives that geom's orientation from the
  endpoints and *discards* one declared beside them (measured, and pinned), so honouring it there
  would bound a shape the compiler does not build.

`<compiler angle>` and `<compiler eulerseq>` are forwarded from the scene entry point, which already
resolved them for the mesh reader three lines away.

## Measured over the shipped assets

336 loadable MJCFs in the asset cache; **421 bodies** declare a rotated collidable analytic geom
(`quat` 574 geoms, `euler` 53), and **27 of the 72 registered robots** own at least one --
`unitree_g1`, `unitree_go2`, `unitree_h1`, `unitree_a1`, `ur5e`, `ur10e`, `stretch`, `shadow_hand`,
`anymal_b`, `anymal_c` among them. Graded against MuJoCo's compiled geometry, each geom's surface
sampled and rotated by `mju_rotVecQuat`:

| | under-bounded on some axis | over-bounded | matching |
|---|---|---|---|
| before | 279 | 247 | 143 |
| after | 0 | 0 | 421 |

Worst under-bound before was `stretch`'s gripper finger, reported 0.0085 m on an axis its geometry
spans 0.1601 m (18.75x). After the change the worst deviation from the sampled surface is
4.9e-13 m too small and 1.4e-4 m too large, the latter being the sampling grid under-reporting a
curved surface rather than the formula being loose. 278 of the 421 reported sizes change; 129 grow
in volume and 53 shrink, the rest being permutations of the same extents.

## Visual

`unitree_go2`'s `FL_thigh` as MuJoCo compiles it, with the proxy each reader publishes drawn over
it. Coverage is measured from separate segmentation passes so the proxy's transparency cannot affect
the count.

rotated collision bound

Of the 13864 pixels of collision geometry, the previously reported proxy leaves **11634 (84%)**
outside it; the new one leaves **0**, and covers exactly 13864 -- for a quarter turn the AABB is the
geom itself.

## Tests

`tests/simulation/isaac/test_geom_rotation_widens_the_collision_bound.py`, 41 cases. MuJoCo is the
oracle throughout: a box is graded against its eight corners rotated by MuJoCo's own
`mju_rotVecQuat` (exact), the curved primitives against a deterministic surface grid (hence a
tolerance). Nothing restates the formula under test.

Pre-fix, with every signature kept and only the rotation reading undone -- **24 failed, 17 passed**:

| class | failing | test functions |
|---|---|---|
| `TestTheOracleAndTheRespellAreSound` (premises) | 0 | 4 |
| `TestARotatedGeomIsBoundedOnTheAxesItOccupies` | 23 | 13 |
| `TestWhatARotationDoesNotChange` (controls) | 0 | 5 |
| `TestTheCompilerUnitsReachTheReader` | 1 | 2 |

A raw revert to `main` instead reports 27 failures, but nine of those are `TypeError` from the
pre-fix signatures rather than behaviour, which is why the behavioural revert is the number quoted.

### Mutations

11 applied, 10 detected, 10 distinct failing-id sets, control clean, source restored byte-identical.

| mutation | cells firing |
|---|---|
| rotation dropped (behavioural revert) | 24 |
| box extent loses `abs()` (signed sum) | 15 |
| box extent transposed, rows read for columns | 2 |
| ellipsoid given the looser box bound | 2 |
| capsule axis read off a row not a column | 2 |
| rotation applied to the `fromto` branch too | 1 |
| sphere given the box treatment | 1 |
| units not forwarded at the scene entry point | 2 |
| units dropped between the two AABB helpers | 2 |
| orientation read without the element's own attributes | 1 here, **12** in `test_mjcf_orientation_spellings.py` |
| signature's `angle_scale` default flipped to radians | **0** |

Two of these are worth spelling out. Reading a matrix row where a column belongs, and reading a
capsule's axis the same way, are both invisible under a `zaxis` tilt -- `zaxis` builds the *minimal*
rotation onto an axis, whose matrix is symmetric enough that the two agree -- so the primitive cases
are also run with three composed Euler angles, which is what catches an index slip. And the
signature default fires nothing because every production path forwards the units explicitly, which
the structural cell pins and which the two "units not forwarded" mutations show is load-bearing; the
default is unreachable from production and is MJCF's own (degrees, `xyz`).

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1572 files).
- mypy **24 == 24** against a pristine `upstream/main` worktree, normalized message sets
  byte-identical in both directions, 0 in either changed file.
- Paired run over an identical 600-file selection (all of `tests/simulation/`, `tests/benchmarks/libero/`
  and every guard that walks the tree or reads source), the new file excluded from both sides:
  **20094 collected and `5 failed, 19815 passed, 274 skipped` on both trees**, the same five failing
  ids, `comm` empty in both directions. All five pass in isolation (29 passed) -- they are
  large-selection artifacts, identical on base.
- `tests/simulation/isaac/` 1654 passed before the change and 1695 after, on the declared mujoco
  floor (3.5.0) with lerobot 0.6.2.

## Not changed

`fromto` on a box or ellipsoid still returns `None`. That exclusion was previously justified by
needing "the rotated-box bound, which this module deliberately does not compute" -- a reason this
change retires -- so the remaining one is stated accurately instead: MuJoCo squares such a geom's
cross-section by copying the first `size` component, a separate rule this reader does not derive.
No shipped asset spells it, and `_extract_mjcf_shape`, the other consumer of that exclusion, is
untouched here.

### DIFF
--- a/changelog.d/2759-geom-rotation-collision-bound.md
+++ b/changelog.d/2759-geom-rotation-collision-bound.md
@@ -1,0 +1,26 @@
+### Fixed: a rotated collision geom is bounded on the axes it occupies
+
+`_geom_aabb` reports the axis-aligned box one `<geom>` occupies in its owning body's frame, and
+`load_mjcf_scene_objects` publishes the union of those as a `SceneObject`'s `size` and `position` -
+the physics footprint that stands in for the object. It read `pos` and `size` and never asked how
+the geom is turned. MJCF gives a geom five mutually exclusive ways to state a rotation (`quat`,
+`euler`, `axisangle`, `xyaxes`, `zaxis`), and dropping it does not make the bound approximate, it
+reports the extents on the wrong axes: a 0.60 m bar turned a quarter turn about z came back as
+`(0.6, 0.04, 0.04)` where MuJoCo places it along y at `(0.04, 0.6, 0.04)`.
+
+The same function already read a rotation on its other channel. A capsule or cylinder may spell its
+placement with `fromto`, whose endpoints carry the axis, and `_segment_aabb` bounds that exactly. So
+one geom got two different bounds depending on which of two spellings MuJoCo compiles to the same
+shape was used for it.
+
+Each primitive now gets its exact bound - the support function for a box and for an ellipsoid, and
+for a capsule or cylinder the endpoints rebuilt from the rotation and handed to `_segment_aabb`, so
+the two spellings share one answer rather than drifting. A sphere is unchanged, no rotation moving
+it, and the `fromto` branch stays rotation-free because MuJoCo derives that geom's orientation from
+the endpoints and discards any the element declares. `<compiler angle>` and `<compiler eulerseq>`
+are forwarded from the scene entry point, which already resolved them for the mesh reader.
+
+Over the 336 loadable MJCFs in the shipped asset cache, 421 bodies declare a rotated collidable
+analytic geom and 27 of the 72 registered robots own at least one. Against MuJoCo's own geometry the
+old rule under-bounded 279 of those bodies and over-bounded 247, the worst reporting 0.0085 m on an
+axis the geometry spans 0.1601 m; afterwards none of the 421 is off by more than 0.1%.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -1734,6 +1734,69 @@ def _segment_length(p1: tuple[float, float, float], p2: tuple[float, float, floa
     return math.sqrt(sum(d * d for d in delta))
 
 
+def _quat_matrix(
+    quat: tuple[float, float, float, float],
+) -> tuple[
+    tuple[float, float, float],
+    tuple[float, float, float],
+    tuple[float, float, float],
+]:
+    """Rotation matrix rows of a unit ``wxyz`` quaternion.
+
+    Column ``j`` is where the geom's own axis ``j`` points in the frame the
+    quaternion is expressed in, which is what a rotated primitive's extent is
+    built from. :func:`_parse_orientation` returns a unit quaternion whichever
+    spelling the element used, so no renormalization happens here.
+    """
+    w, x, y, z = quat
+    return (
+        (1.0 - 2.0 * (y * y + z * z), 2.0 * (x * y - z * w), 2.0 * (x * z + y * w)),
+        (2.0 * (x * y + z * w), 1.0 - 2.0 * (x * x + z * z), 2.0 * (y * z - x * w)),
+        (2.0 * (x * z - y * w), 2.0 * (y * z + x * w), 1.0 - 2.0 * (x * x + y * y)),
+    )
+
+
+def _rotated_box_half(
+    half: tuple[float, float, float],
+    rot: tuple[
+        tuple[float, float, float],
+        tuple[float, float, float],
+        tuple[float, float, float],
+    ],
+) -> tuple[float, float, float]:
+    """Half-extent of a rotated box - exact, not an approximation.
+
+    A box's farthest point along axis ``i`` is the corner whose every local
+    component pushes that way, so the extent is ``sum_j |R[i][j]| * half[j]``.
+    Read without the rotation the extent is reported on the geom's own axes
+    instead of the body's, which for a quarter turn is a permutation rather
+    than a small error: a 0.16 m finger pad rotated onto ``x`` keeps reporting
+    0.0085 m there.
+    """
+    return tuple(sum(abs(rot[i][j]) * half[j] for j in range(3)) for i in range(3))  # type: ignore[return-value]
+
+
+def _rotated_ellipsoid_half(
+    half: tuple[float, float, float],
+    rot: tuple[
+        tuple[float, float, float],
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Fixes `_geom_aabb` in the Isaac MJCF scene loader so a geom's own rotation (`quat` / `euler` / `axisangle` / `xyaxes` / `zaxis`, resolved through `<default>` classes and `<compiler angle>`/`eulerseq`) is folded into the collision AABB that `load_mjcf_scene_objects` publishes as a `SceneObject`'s footprint. Each primitive gets its exact bound: the box support function, the tighter ellipsoid support function, and the capsule/cylinder respelled into the endpoints a `fromto` would carry and handed to the existing `_segment_aabb`, so the two spellings of one shape now share one answer. The `fromto` branch deliberately stays rotation-free, matching MuJoCo's compiler (pinned as a premise test). All touched functions are private and `load_mjcf_scene_objects(path)`'s signature is unchanged, so there is no API one-way door; the changed sizes are corrections toward MuJoCo's own compiled geometry, not a silent default change.

## What's good

- I independently re-derived and numerically verified both new formulas (box support against brute-force rotated corners, ellipsoid support against a dense surface sample): exact and never under-bounding. At identity rotation the capsule/cylinder path reproduces the previous result byte-for-byte via `_segment_aabb`'s existing arithmetic.
- The new test file (41 cases) uses MuJoCo as the oracle rather than restating the formula, separates premises / regression / controls, includes a structural pin that the compiler units are forwarded through the whole chain (defaults alone would silently pass), and composes general Euler rotations to catch row/column index slips that a `zaxis` tilt cannot see. All 41 pass on the head SHA under MuJoCo 3.8.0 locally.
- AGENTS.md compliance: changelog fragment (not `CHANGELOG.md`), ASCII-only strings, value-based MuJoCo enum comparisons (`int(...) == int(...)`, int-set membership), behaviour-describing test names, no host paths, closing keyword in the body.
- The new `_parse_orientation` call can raise `ValueError` for a geom declaring two orientation spellings, but MuJoCo refuses such a model outright and the same raise already exists on this load path (body attrs, `_find_body_mesh`), inside `load_mjcf_scene_objects`' documented `ValueError` contract - no new crash class on loadable scenes. Likewise the zero-half-length capsule now returns `None` instead of a ball, which only affects models MuJoCo also refuses.

PR #2760, opened 2026-08-26
fix(training): report a trainer module's missing dependency as its own problem

cagataycali:
## Problem

`import_trainer_class` resolves a provider name to a `Trainer` subclass through three rungs, and the last one imports `strands_robots.training.<provider>`. That import fails two distinguishable ways and they want different answers:

- the module is **absent** - the provider name is unknown, and the useful reply is the list of names that *are* known;
- the module is **present** but something *it* imports is absent - the name was right, and the useful reply is the missing distribution.

The handler collapsed both:

```python
    except ImportError:
        # No strands_robots.training.<provider> module; fall through to the
        # ValueError below so the caller gets the full "available trainers" list.
        pass
```

So a caller whose spelling was correct was told to go and check the spelling. Measured on an install without `torch`:

| | reported |
|---|---|
| ground truth (`importlib.import_module("strands_robots.training.rl")`) | `ModuleNotFoundError: No module named 'torch'` (`exc.name='torch'`) |
| `import_trainer_class("rl")` before | `ValueError: No trainer registered for provider 'rl'. Available trainers: ['cosmos3', 'fast_sac', 'groot', 'lerobot_local', 'mock', 'ppo', 'sagemaker']` |
| `import_trainer_class("rl")` after | `ModuleNotFoundError: No module named 'torch'` |

This is reachable in the shipped tree rather than hypothetical: `strands_robots.training.rl` sits under exactly that prefix and its own module docstring records the property - "Importing this package imports `torch` (via the env / algo modules), so it is not imported by `strands_robots.training.__init__`". `torch` arrives through an extra and is not in `project.dependencies`, so an install really can have the module and not the dependency.

## Two of the three resolvers already did this

Three packages resolve a provider the same way. The other two separate the failures, and each says so in its own `Raises:` block:

| resolver | auto-discovery handler | separates the two | in `Raises:` | dedicated test |
|---|---|---|---|---|
| `registry/policies.py::import_policy_class` | `except ImportError as exc:` + name check, re-raises through an error naming the install remedy | yes | yes | yes |
| `transforms/factory.py::import_transform_class` | `except ModuleNotFoundError as exc:` + name check, re-raises the original | yes | yes | yes |
| `training/factory.py::import_trainer_class` | `except ImportError: pass` | **no** | no | no |

`import_transform_class`'s comment states the rule outright - "The provider module exists but something IT imports is absent - a missing backend dependency, not an unregistered provider. Re-raise so the caller is sent to the right problem." And `import_trainer_class`'s docstring already advertises that its resolution order mirrors these; it was accurate about the order and silent about the failures.

## Why nothing caught it

`training/factory.py` was **100% covered before this change** and is 100% covered after. Coverage records that the handler ran, not which of the two failures it absorbed - every existing case reached it via a module that genuinely does not exist, which is the branch it answers correctly. The nearest test spelled the gap out from the other side: `tests/training/test_factory.py::test_importable_module_without_trainer_raises` pins that an importable module with no `Trainer` "raises ValueError with the available-trainers list (not ImportError)", which is right for *that* input and was the only ImportError-adjacent expectation in the file.

## Change

The handler is narrowed to the absence of the provider module itself, compared on the `ModuleNotFoundError`'s own `name` (for a missing submodule that is its full dotted path - pinned as a premise), and the class lookup moves into an `else` so only the import is guarded. Anything else surfaces. This is `import_transform_class`'s shape, chosen over `import_policy_class`'s richer wrapper because that wrapper is a private helper of another package and the transform resolver is the structural twin: same three rungs, same terminal `ValueError` carrying an "available" list.

The two remaining copies are left alone deliberately. They differ in disposition - one re-raises, one enriches with an install hint - so a single shared implementation would have to change the behaviour of two correct resolvers to fix one incorrect one. The *rule* is single-sourced in the test instead.

`docs/training/` documents no resolution ladder (`grep` for auto-discovery / `register_trainer` / "Available trainers" across it returns nothing), so the docstring is this behaviour's documentation surface and is updated in place.

## Tests

`tests/test_provider_resolution_reports_a_missing_dependency.py`, 27 cases. It grades a **derived** inventory rather than a list of three, because the defect is what happens to the next resolver. A resolver is discovered structurally: a function handing `importlib.import_module` an f-string that begins `strands_robots.` and interpolates one of that function's own parameters. That signature selects exactly the shipped three with no false positives:

```
strands_robots/registry/policies.py    import_policy_class     strands_robots.policies.
strands_robots/training/factory.py     import_trainer_class    strands_robots.training.
strands_robots/transforms/factory.py   import_transform_class  strands_robots.transforms.
```

Each is then driven for real - a module is planted on the package's `__path__` whose body imports a distribution nothing installs - so the production path runs against the exact `ModuleNotFoundError` the import system raises. Assertions are disposition-agnostic (re-raise and enrich both satisfy "reports the dependency"), so the file does not freeze either sibling's choice.

Pre-fix split: **3 failed / 24 passed**, all three failures the `import_trainer_class` cells of the regression class, all three reporting the defect verbatim (`ValueError: No trainer registered for provider 'heavybackend'. Available trainers: [...]`). The six sibling-resolver cells in that same class pass pre-fix, which is what shows the guard is derived rather than fitted to this change.

| class | role | failed / total pre-fix |
|---|---|---|
| `TestTheResolverInventoryIsDerived` | the discovery rule | 0 / 4 |
| `TestEveryProviderResolverSurfacesAMissingDependency` | regression | 3 / 9 |
| `TestAnUnregisteredProviderIsStillUnregistered` | over-reach controls | 0 / 9 |
| `TestTheImportSystemNamesTheModuleItCouldNotFind` | premise (`exc.name`) | 0 / 3 |
| `TestTheShippedTreeCanReachThisFailure` | premise (reachability) | 0 / 2 |

## Mutations

Ten rows, control clean, source and test restored byte-identical after each.

| row | fires | mutation |
|---|---|---|
| b0 | 0 | control, no mutation |
| b1 | 3 | the whole rung reverted to the pre-fix handler - exactly the three regression cells |
| b2 | 0 | handler widened back to bare `ImportError`, name check kept |
| b3 | 4 | name check dropped, every failed import re-raised - **three of the four are pre-existing tests** |
| b4 | 6 | name check inverted (`==` for `!=`) |
| b5 | 4 | compared against the bare provider name instead of the dotted module (same set as b3) |
| b6 | 0 | class lookup moved back inside the `try`, name check kept |
| b7 | 0 | the `Raises:` paragraph reverted, code kept |
| b8 | 7 | a fourth resolver lands in the tree, inventory **derived** |
| b9 | 0 | the same fourth resolver, inventory **hardcoded** to today's three |

The three zeros are reported with their reasons rather than dropped. b2 is a semantic no-op given the name check: a plain `ImportError` carries `name=None`, which is not the module name, so it is re-raised either way. b6 likewise - nothing in the class lookup raises `ModuleNotFoundError`, so moving it inside the `try` changes nothing for any input; it is guarded against a future one. b7 is honest about the prose: no test grades the `Raises:` text.

b8/b9 are the pair that justifies deriving the inventory. With a fourth resolver planted in the tree that repeats the original mistake, the derived scan fires 7: the two inventory cases, the three cells that catch this exact defect, and two of the controls that the deliberately crude stand-in does not satisfy. With the inventory hardcoded to today's three the same plant fires **0** and the suite reads green, while `collected` silently drops from 107 to 100 - a suite that runs seven fewer tests still reports success, so a passing count is not the oracle here.

b3 firing three pre-existing tests (`TestFactory::test_unknown_provider_raises`, `TestBothResolversServeEveryListedTrainer::test_the_refusal_advertises_only_names_it_can_serve`, `TestTheRuntimeRungKeepsItsPrecedence::test_an_unknown_provider_is_still_refused`) is the over-reach evidence: reporting everything as a dependency problem is caught by the suite that already exists.

## Gate

Full `pytest tests/` on both trees at `5791e4c5`:

| | collected | failed | passed | skipped | errors | time |
|---|---|---|---|---|---|---|
| base | 40127 | 29 | 39764 | 310 | 24 | 1957s |
| branch | 40154 | 29 | 39791 | 310 | 24 | 1838s |

`+27` collected and `+27` passed, which is exactly this file's 27 cases. **53 failing/erroring ids on each side, identical, `comm` empty in both directions.** All 53 pre-existing and unrelated: 44 need `awsiot` (declared in the `mesh-iot` extra, not installed here), 3 are lerobot-from-source `encode()` drift, 2 missing dist metadata, 2 `device_connect` hardening, 1 a leftover dataset directory, 1 an unrelated assertion.

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ`. mypy `24 == 24` against a pristine worktree of the same base, normalized message sets byte-identical in both directions, 0 in the changed files.

Coverage, from the same paired runs:

| module | base | branch |
|---|---|---|
| `training/factory.py` | 49 stmts, 0 missing, 100% | 51 stmts, 0 missing, 100% |
| `transforms/factory.py` | 51 stmts, 5 missing, 90% | 51 stmts, 1 missing, 98% |
| `registry/policies.py` | 144 stmts, 0 missing, 100% | unchanged |

The transform gain is incidental and real: driving the derived guard over all three resolvers reaches `transforms/factory.py`'s `dir(mod)` scan branch, which nothing exercised before. Its one remaining uncovered line is unreachable with the shipped registry - no provider declares a `transform` block (measured `[]`). Package total 1884 -> 1880 missing. The `simulation/mujoco/rendering.py` `+1` / `simulation.py` `-1` in the same tables is run-to-run nondeterminism in a threaded policy path, not attributable to this change.

No visual artifact: this changes which error an unresolvable provider reports, and touches no policy, simulation behaviour, rendering, recording or robot asset. The measured before/after reports and the mutation table are the evidence.


yinsong1986 (approved):
## Summary

Narrows `import_trainer_class`'s auto-discovery handler from a blanket `except ImportError: pass` to catching only the `ModuleNotFoundError` whose `name` is the provider module itself, moving the class lookup into an `else` so only the import is guarded. A provider module that exists but whose own dependency is absent (e.g. `training.rl` without `torch`) now surfaces the real `ModuleNotFoundError` instead of collapsing into `ValueError: No trainer registered ... Available trainers: [...]` and sending a correctly-spelled caller to check the spelling. This brings the trainer resolver into line with the contract `import_policy_class` (registry/policies.py) and `import_transform_class` (transforms/factory.py) already implement with the identical `exc.name` discriminator, and the `Raises:` block is updated to document both exception classes per the AGENTS.md `Raises:`-completeness rule. Verified at the head commit: the new test file's 27 cases all pass, and `tests/training` + `tests/transforms` pass (3857 passed; the one `test_lerobot_e2e.py` error is a MuJoCo-rendering environment failure that reproduces identically on `upstream/main`, so it is not attributable to this diff).

## What's good

- The regression test grades a *derived* inventory (AST scan for `import_module` of a caller-interpolated `strands_robots.` f-string) rather than a hardcoded list of three, with precision controls in both directions - I confirmed the signature selects exactly the three shipped resolvers and nothing else on the current tree.
- Over-reach controls pin that the unknown-provider `ValueError` with the available list still fires, so the fix cannot silently invert into reporting everything as a dependency problem.
- Fixture hygiene follows AGENTS.md: `monkeypatch` for `__path__`, `sys.modules.pop` only of modules the test itself planted (never orphaning a real module), `importlib.invalidate_caches()` on teardown, no host paths, behaviour-describing test names.
- Changelog fragment present under `changelog.d/` per step 3 of the PR workflow.

PR #2761, opened 2026-08-26
fix(mesh): keep the severity a safety event was raised with

cagataycali:
## Problem

`SensorLoopsMixin.publish_safety_event` sends one event two ways. It built the audit copy by
spreading the caller's payload over the parameter:

```python
payload={"severity": severity, **record}
```

so a payload carrying a `severity` field replaced it. No refusal, nothing logged. Measured
through the public method, capturing what each half was handed:

| raised with | payload | wire `event["severity"]` | audit `payload["severity"]` |
|---|---|---|---|
| `"critical"` | `{"joint": "elbow", "limit": 1.57}` | `info` | `critical` |
| `"critical"` | `{"joint": "elbow", "severity": "info"}` | `info` | **`info`** |
| `"critical"` | `{"severity": 0}` | `info` | **`0`** |
| `"warning"` (default) | `{"severity": "info"}` | `info` | **`info`** |

What makes that consequential rather than cosmetic is that there is no second copy to
compare against. The wire copy is uniformly `"info"` on purpose - issue #272 removed
per-branch severity from `strands/+/safety/event` so a subscriber could not read it as a
content-channel oracle for a rejection reason - and the method's own comment records the
consequence:

> The real severity is preserved only in the local audit record below.

So a critical stop could be audited as informational, with nothing anywhere disagreeing. I
also checked the negative: `"critical"` appears nowhere in the published wire event
(`assert "critical" not in json.dumps(event)`), which is the premise the fix rests on and is
pinned as such.

## Why this is the existing rule, not a new one

Eight of the nine record builders in this mixin already resolve exactly this collision the
other way. Derived from the shipped class rather than listed - a method that merges a mapping
it did not build, via `update` or a `**` spread:

| builder | merges | re-asserts its own keys |
|---|---|---|
| `_read_pose` (3 merge sites), `_read_imu`, `_read_odom`, `_read_lidar_summary`, `_read_lidar_state`, `_read_hands`, `_read_map_info`, `_read_health` | yes | yes, via `_stamp_local_keys` |
| `publish_safety_event` | yes | **no** |

`_stamp_local_keys` names the hazard in its own docstring - "merged last, a provider mapping
carrying one of those seeded names replaces the local reading" - and cites
`session.PeerInfo.to_dict`, which spreads the peer's own payload *first* so the locally
decided keys win. This applies that precedence to the ninth builder: `payload={**record,
"severity": severity}`.

The helper itself is deliberately not reused. It also stamps `peer_id`, and
`log_safety_event` already carries the peer as a field of its own envelope, so routing this
record through it would change the record's shape rather than only its precedence - measured
as break row b3 below, which trips four cells including two controls.

`tests/mesh/test_safety_event_payload_reaches_the_wire.py` owns a cell named
`test_the_audit_record_still_carries_the_real_severity` - the exact property that failed -
and drives it with `payload={"reason": "x"}`. A payload with no `severity` key cannot
collide, so the assertion held whichever way the merge was ordered.
`tests/mesh/test_sensor_readers.py:518` asserts the same thing with the same non-colliding
payload. The property was named and never exercised.

There is a second reason it could go unnoticed for a long time: the merge only changed the
record when the two values *disagreed*. The audit payloads `mesh.core._emit_resume_denied`
writes carry a `severity` field whose value is the severity it would also pass as the
argument, so the idiomatic in-repo shape is exactly the shape that agrees. Two cells pin
that (`TestTheShadowWasSilentWhenTheTwoValuesAgreed`); they pass either way, which is the
point rather than a gap, and they are reported as passing pre-fix below.

## Reachability

Latent, and stated plainly: of the 12 `publish_safety_event` call sites in the tree, **0**
pass a payload containing a `severity` key, so no shipped call path was affected. The method
is public API on `Mesh`, its docstring names the parameter as "the real severity", and the
`severity`-in-the-payload shape is what both audit-writing paths in this repo already
produce - `_emit_resume_denied` puts it there explicitly, and `publish_safety_event` puts it
there itself - so a caller naming that field is following the local convention rather than
doing something exotic.

## Tests

New file `tests/mesh/test_safety_severity_is_the_parameters.py`, 31 cells. Pre-fix split,
per class and per role:

| class | role | fail | total |
|---|---|---|---|
| `TestThePremisesTheFixRestsOn` | premise | 0 | 4 |
| `TestTheSeverityParameterSurvivesAPayloadOfTheSameName` | regression | 10 | 11 |
| `TestTheShadowWasSilentWhenTheTwoValuesAgreed` | passes either way, by design | 0 | 2 |
| `TestWhatIsUnchanged` | control | 0 | 5 |
| `TestTheFixDoesNotReachIntoTheOtherHalf` | over-reach guard | 0 | 4 |
| `TestNoMergedMappingShadowsALocalKey` | regression (derived) | 1 | 5 |
| | | **11** | **31** |

The one regression cell that passes pre-fix is
`test_no_payload_above_agrees_with_the_severity_it_is_graded_against`, a meta-assertion that
no parametrized payload repeats the severity it is graded against - without it a cell could
pass with the shadow still in place. Every payload in the regression set therefore differs
from both `"critical"` and the `"warning"` default.

## Break table

10 mutations, 9 distinct non-empty firing sets, control clean, source restored
byte-identical (md5-checked per row). `collected` recorded alongside `fires`, so a mutation
that deselects rather than fails cannot read as green.

| # | mutation | fires | collected |
|---|---|---|---|
| b0 | control, no mutation | 0 | 31 |
| b1 | payload spread last again (the defect) | 11 | 31 |
| b2 | severity re-asserted as a hardcoded literal | 9 | 31 |
| b3 | `_stamp_local_keys` used instead (also stamps `peer_id`) | 4 | 31 |
| b4 | alternative disposition: refuse a payload carrying the name | 18 | 31 |
| b5 | fix applied to the wire copy instead of the audit copy | 13 | 31 |
| b6 | the wire severity carries the real severity too | 2 | 31 |
| b7a | plant a 10th shadowing builder, inventory derived (as shipped) | 1 | 31 |
| b7b | the same plant, inventory hardcoded to today's names | **0** | 31 |
| b8 | shadow rule keyed on the first spread rather than the last | 1 | 31 |
| b9 | a regression payload made to agree with the severity it grades | 1 | 31 |

b7a/b7b is the pair that shows the derived inventory is load-bearing: a hardcoded list equal
to today's derived set is completely silent when a 10th builder lands, so the plant is the
only way to demonstrate the difference.

b4 is the disposition a reviewer might reasonably propose instead - refuse a payload carrying
the name. It trips 18 cells, including `test_an_audit_failure_is_still_survived`: raising
from a safety-event publisher is worse than the shadow it removes. b6 shows the over-reach
guard on the other half works - making the wire severity real trips the two premise cells,
so the wire's uniformity cannot be quietly given up.

b8 was found by this table: it fired 0 initially because every dict literal in the mixin has
at most one spread, so `min` and `max` agreed. A constructed exemplar with a key *between*
two spreads closes it, and the row now fires.

## Gate

- **Paired pytest, byte-identical.** Identical 495-file selection (all 206 `tests/mesh/`
  files plus 291 guards that walk the tree, read source/docstrings, or read `changelog.d`),
  the new file excluded from both trees, every path asserted present on both. Identical
  **20070 collected** and `21 failed, 19958 passed, 67 skipped, 24 errors` on both
  (648.75s vs 661.02s). 45 failing/erroring ids each, `comm` empty in **both** directions.
  All 45 are `tests/mesh/test_iot_*` and need `awsiot`/`awscrt`, both absent here and both
  declared in the `[mesh-iot]` extra (verified with `tomllib` + `find_spec`, not assumed).
- `tests/mesh/test_zenoh_transport_security.py` opens a live session, so it was excluded
  from the pair and run alone: **9 passed, 2 skipped on both trees**.
- **mypy** `strands_robots tests tests_integ`: **24 == 24** against a worktree pinned to the
  merge-base, normalized message sets byte-identical in both directions, **0** in the changed
  files.
- **ruff** `check` and `format --check` clean over `strands_robots tests tests_integ`
  (1578 files).
- The four sibling suites that assert a severity, on the rebased branch:
  `test_safety_event_payload_reaches_the_wire`, `test_sensor_readers`, `test_deep_mesh`,
  `test_estop_replay` - **197 passed**. With the changelog guards and the test file #2758
  added: **293 passed**.
- Changelog fragment validated by `scripts/assemble_changelog.py --check` and confirmed
  present in `--print` output.

No visual artifact: the change is to an audit record's key precedence, not to a policy,
simulation, render, recording or asset. The measured before/after table above is the
artifact.

### DIFF
--- a/changelog.d/2761-safety-severity-is-not-the-payloads.md
+++ b/changelog.d/2761-safety-severity-is-not-the-payloads.md
@@ -1,0 +1,41 @@
+### Fixed: a safety event keeps the severity it was raised with
+
+`SensorLoopsMixin.publish_safety_event` sends one event two ways and built the audit copy
+by spreading the caller's payload over the parameter, so a payload carrying a `severity`
+field replaced it -- with no refusal, and nothing logged:
+
+| raised with | payload said | audit record said |
+|---|---|---|
+| `severity="critical"` | (no `severity` field) | `critical` |
+| `severity="critical"` | `severity: "info"` | `info` |
+| `severity="critical"` | `severity: 0` | `0` |
+| `severity="warning"` (default) | `severity: "info"` | `info` |
+
+What makes the replacement consequential rather than cosmetic is that there is no second
+copy to compare against. The wire copy carries a uniformly `"info"` severity on purpose:
+issue #272 removed per-branch severity from `strands/+/safety/event` so a subscriber could
+not read it as a content-channel oracle for a rejection reason, and the method's own
+comment records the consequence -- "the real severity is preserved only in the local audit
+record below". A critical stop could therefore be audited as informational with nothing
+anywhere disagreeing.
+
+The precedence is not a new rule; this mixin already applies it eight times. `_read_pose`,
+`_read_imu`, `_read_odom`, `_read_lidar_summary`, `_read_lidar_state`, `_read_hands`,
+`_read_map_info` and `_read_health` merge a provider mapping and then re-assert the keys
+this process decided, through `_stamp_local_keys`, whose docstring names the hazard --
+"merged last, a provider mapping carrying one of those seeded names replaces the local
+reading" -- and cites `PeerInfo.to_dict`, which spreads the peer's own payload *first* so
+the locally decided keys win. `publish_safety_event` built the ninth record and was the one
+that spread last. It now spreads first, the same precedence by the same reasoning.
+
+The helper itself is deliberately not reused here: it also stamps `peer_id`, and
+`log_safety_event` already carries the peer as a field of its own envelope, so routing this
+record through it would change the record's shape rather than only its precedence.
+
+A `payload` entry named `severity` is discarded from the audit copy rather than preserved
+under another name, matching how `PeerInfo.to_dict` resolves the same collision; the wire
+copy is untouched, so a caller who puts a severity in the payload still sees it there. The
+rule is now derived from the shipped class -- no method of the mixin may place an explicit
+key ahead of a `**` spread, and a method that merges with `update` must re-assert its own
+keys -- so a tenth record builder added later is held to it instead of inheriting an
+exemption by omission.
--- a/strands_robots/mesh/sensors.py
+++ b/strands_robots/mesh/sensors.py
@@ -732,7 +732,11 @@ class SensorLoopsMixin:
         Args:
             event_type: Short, lowercase event identifier (e.g. ``"estop"``).
             severity: The real severity. It reaches the audit record only - the
-                wire copy is uniformly ``"info"`` (issue #272).
+                wire copy is uniformly ``"info"`` (issue #272) - so the audit
+                record is the only surviving copy, and this parameter wins over
+                a ``payload`` field of the same name rather than being replaced
+                by it. A ``payload`` entry named ``severity`` is not a second
+                channel for this argument; it is discarded from the audit copy.
             payload: Event-specific fields, or ``None`` for an event with none.
         """
         if not self._running:
@@ -759,7 +763,12 @@ class SensorLoopsMixin:
             log_safety_event(
                 event_type=event_type,
                 peer_id=self.peer_id,
-                payload={"severity": severity, **record},
+                # ``severity`` last: the parameter is the documented channel for the
+                # real severity, so an event-specific field that happens to be named
+                # ``severity`` must not replace it. Same precedence as
+                # ``session.PeerInfo.to_dict``, which spreads the peer's own payload
+                # first so the keys that process decided win a name collision.
+                payload={**record, "severity": severity},
             )
         except Exception as exc:  # noqa: BLE001
             logger.debug("[mesh] %s: audit log write failed: %s", self.peer_id, exc)
--- a/tests/mesh/test_safety_severity_is_the_parameters.py
+++ b/tests/mesh/test_safety_severity_is_the_parameters.py
@@ -1,0 +1,334 @@
+"""The severity a safety event is raised with survives a payload field of that name.
+
+:meth:`~strands_robots.mesh.sensors.SensorLoopsMixin.publish_safety_event` sends
+one event two ways. The wire copy carries a uniformly ``"info"`` severity, on
+purpose: issue #272 removed per-branch severity from
+``strands/+/safety/event`` so a subscriber could not read it as a
+content-channel oracle for a rejection reason. The consequence the method's own
+comment records is that the audit record is *the only surviving copy of the real
+severity*.
+
+That copy was built by spreading the caller's payload over the parameter::
+
+    payload={"severity": severity, **record}
+
+so a payload carrying a ``severity`` field replaced it - silently, with no
+refusal and nothing logged. An event raised ``severity="critical"`` was audited
+as whatever the payload said, and because the wire copy is uniform there is no
+second record to compare against.
+
+Eight of the nine record builders in this mixin already resolve exactly this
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

One-line precedence fix in `SensorLoopsMixin.publish_safety_event`: the audit copy of a safety event was built as `{"severity": severity, **record}`, so a payload carrying a `severity` field silently replaced the parameter -- and since #272 made the wire copy uniformly `"info"`, the audit record is the only surviving copy of the real severity, meaning a critical stop could be audited as informational with nothing anywhere disagreeing. The fix flips the merge to `{**record, "severity": severity}`, matching the precedence the mixin's other eight record builders already apply via `_stamp_local_keys` and `PeerInfo.to_dict`. Verified against the head: no in-repo caller passes a payload with a `severity` key, so no existing audited severity changes; the wire copy and the caller's mapping are untouched (both pinned by tests); the new test file's 160-test run passes alongside the existing `test_safety_event_payload_reaches_the_wire.py` and `test_sensor_readers.py` suites, and it correctly identifies why prior coverage was silent (the existing severity assertion drove a non-colliding payload). The AST-derived rule (`TestNoMergedMappingShadowsALocalKey`) holds future record builders to the same precedence rather than exempting them by omission. Changelog fragment present per step 3; lint and format clean; no unicode, no host paths.

## What's good

- The parametrized colliding payloads deliberately avoid the raised/default severity values, and a meta-test pins that, so no cell can pass by coincidence -- with the agreement case measured in its own class.
- The derived-rule tests include vacuity controls (`test_the_shadow_rule_is_not_vacuous`, the between-two-spreads case), so the AST scan cannot silently stop matching.
- Scope discipline: the wire half, the caller's mapping, and the audit-failure survival path are each pinned as unchanged.

yinsong1986 (approved):
## Summary

One-line precedence fix in `SensorLoopsMixin.publish_safety_event`: the audit copy of a safety event was built as `{"severity": severity, **record}`, so a caller payload carrying a `severity` field silently replaced the parameter -- and since #272 made the wire copy uniformly `"info"`, the audit record is the only surviving copy of the real severity, so a critical stop could be audited as informational with nothing anywhere disagreeing. The fix flips the merge to `{**record, "severity": severity}`, the same precedence the mixin's other eight record builders already apply via `_stamp_local_keys` and that `PeerInfo.to_dict` documents. The wire copy is untouched (pinned by `TestTheFixDoesNotReachIntoTheOtherHalf`), the audit envelope's shape is unchanged (only the value of an existing key), and the defect is latent -- 0 of 12 call sites pass a `severity` payload key -- so nothing shipped depended on the old ordering.

## What's good

- The regression test drives the actual collision (`payload={"severity": ...}`) that the two pre-existing suites named but never exercised, and its meta-cell (`test_no_payload_above_agrees_with_the_severity_it_is_graded_against`) prevents the parametrized set from decaying into agreement with the graded severity.
- The derived-inventory rule (`TestNoMergedMappingShadowsALocalKey`) grades every merging builder off the shipped class AST, so a tenth builder added later is held to the precedence instead of inheriting an exemption by omission -- and the b7a/b7b break-table pair demonstrates the derivation is load-bearing rather than decorative.
- Deliberate scope discipline: `_stamp_local_keys` is not reused (it would also stamp `peer_id` and change the record shape), the alternative refuse-the-collision disposition was measured and rejected (raising from a safety-event publisher is worse than the shadow), and the wire uniformity from #272 is pinned as a premise so it cannot be quietly given up.
- Changelog fragment in `changelog.d/` per AGENTS.md step 3; all new strings ASCII; test classes named for behaviour, not provenance.

PR #2762, opened 2026-08-26
feat(drivers): native daemon driver for the Pollen Reachy Mini

cagataycali:
## Problem

`Robot("reachy_mini", mode="real")` could not build anything:

```
ValueError: Unsupported robot type: 'reachy_mini'. Known lerobot robot types:
['bi_openarm_follower', ..., 'reachy2', ...]
```

Not a registry gap. `reachy_mini` has been in `registry/robots.json` for some time,
with a MuJoCo asset and four aliases. But lerobot has no robot class for it, and the
default driver builds a lerobot `RobotConfig`, so there was no path to the hardware.

## What this adds

`strands_robots.drivers.reachy.ReachyDriver`, reached through the `HardwareDriver`
seam. Measured through the public factory:

| call | before | after |
|---|---|---|
| `Robot("reachy_mini", mode="real")` | `ValueError: Unsupported robot type` | `ReachyDriver` |
| `Robot("reachy_mini", mode="real", driver="strands", port="reachy-b.local")` | `ValueError` | `ReachyDriver`, host `reachy-b.local:8000` |
| `Robot("reachy_mini", mode="real", driver="lerobot")` | `ValueError` | `ValueError` (unchanged) |
| `Robot("reachy2", mode="real")` | lerobot | lerobot (unchanged) |

`missing_driver_members(ReachyDriver)` is `()` and `isinstance(driver, HardwareDriver)`
is `True`. The registry entry declares `hardware.driver = "strands"` with **no**
`lerobot_type`, which is the honest statement: `driver="lerobot"` stays refused,
because lerobot genuinely cannot drive this robot.

The Mini is a different shape from the seam's first driver. The G1 needed a bus lerobot
does not speak; the Mini needs a *robot model* lerobot does not have - a 6-DOF Stewart
head on a rotating body, two antennas, no arms, no gait, addressed over the Reachy
daemon's REST API plus a real-time link.

## Reuse rather than a second daemon client

Every wire detail here was read off the shipped Device Connect driver rather than
guessed. `reachy_transport` already provides `api()`, `WebSocketLink`, `ZenohLink` and
`rpy_to_pose`, and this driver imports them:

- `connect_eagerly` probes `GET /api/daemon/status` - both the reachability check and
  the call that reports the variant. A Lite (no onboard computer) is driven over a
  WebSocket, a Wireless (onboard CM4) over Zenoh.
- `stop` and `stop_task` post `/api/move/stop`.
- Commands go out in the daemon's own units: `head_pose` as a 4x4 from `rpy_to_pose`,
  `body_yaw` and `antennas_joint_positions` in radians.

No endpoint was invented. `_battery` is read from the status payload when it carries a
percentage and left `None` otherwise, because the daemon documents status as "daemon
status, motor state, and control frequency" and a dedicated battery endpoint is not
something this repository can confirm without hardware. The mesh reads every sensor
attribute with a `getattr` default, so no battery field means no battery topic rather
than a broken driver.

## `_pose` is measured, not modelled

The link reports six Stewart leg positions. Converting those into a head pose needs
forward kinematics of the platform that this repository does not have, so `_pose` comes
from the **IMU quaternion** instead - the Mini's IMU is mounted in the head, so its
quaternion is the head's orientation as measured. An IMU frame carrying no quaternion
leaves `_pose` alone rather than publishing a pose with no orientation in it. The legs
are still reported, as legs (`_joints`).

## The envelope refuses; it does not clamp

The limits live in a separate module, `strands_robots.tools.reachy`, because two
consumers need the same answer: this driver and the `reachy_*` agent tools that will sit
on the same daemon. A regression test asserts the driver imports them rather than
restating any of `40.0` / `160.0` / `65.0`.

They are not all of one kind. Pitch and roll are the platform's travel. The head-body
yaw delta is a **coupling** limit between two values, and neither value alone is out of
range when the pair is:

| action | verdict |
|---|---|
| `{"head_yaw": 60}` | success |
| `{"body_yaw": -60}` | success |
| `{"head_yaw": 60, "body_yaw": -60}` | refused: `differ by 120 deg, which exceeds the head-body coupling limit of 65 deg` |

A guard that checks one axis at a time cannot see the third row, so the pair is checked
as a pair, after both values are known to be individually legal - otherwise a caller is
told about a coupling when the real problem is one value.

Refusal rather than clamping, because a clamp is a silent default: the call reports
success while the head goes somewhere other than where it was sent. The refusal names
the limit, so a caller learns the envelope from the refusal.

## Deliberately absent

- **No `_lidar_*`.** The Mini has no lidar.
- **`start_task` / `run_policy` refuse outright** rather than returning a "not wired
  yet" stub. With no arms and no gait there is no action space a policy is trained
  against; the refusal names the recorded-move path instead of implying a rollout is
  coming. `stop` by contrast is real, and a daemon that refuses the stop is not reported
  as a halt.
- **No vendored `reachy_*` tools in this PR.** `strands_robots/tools/` is 20 `@tool`s
  today; the tool set for this robot would be 12 more. The seam's first driver set the
  precedent - `strands_robots/tools/g1/` contains **zero** `@tool`s, only the shared
  engine and constants the driver needs - and this PR follows it: the driver plus the
  shared envelope the tools will import. `tools/reachy/__init__.py` names that follow-up
  explicitly.

## Registration

I am the second shipped driver, so the per-driver registration block became a table
(`_SHIPPED_DRIVERS`). The import guard, the already-registered tolerance and the alias
handling are now written once, and a driver whose import fails skips only itself instead
of costing the other drivers their registration. A derived test holds every entry in that
table to the seam contract, so a driver added later is graded without editing the test.

All five registry aliases (`reachy`, `reachymini`, `reachy-mini`, `pollen_reachy_mini`,
`reachy_mini`) resolve to `ReachyDriver`; `reachy2` and `unitree_g1` are unchanged.

## Docs

Two statements in `docs/getting-started/robot-factory.md` were already false before this
change:

- "the lerobot driver ... is what every robot in the shipped registry uses" - untrue as
  soon as the first native driver landed.
- a quoted refusal ending `Robots with a native driver: none.` - that sentence is the
  live output of `list_native_drivers()`, so it was naming the wrong set already.

Both corrected against the measured output (`reachy_mini, unitree_g1`), plus a note that
`lerobot_type` and `driver` are independent, using the two shipped drivers as the
contrasting examples.

## Tests

`tests/drivers/test_reachy_driver.py`, 103 cases, no Reachy and no daemon required. Two
doubles, each faithful in the respect the tests depend on: the daemon double returns
`{"error": ...}` for failures because that is what `api()` really returns (which is why
the driver needs no `try`), and the link double **records** its commands, because a
double that dropped them could not tell "converted degrees to radians" from "sent
nothing". The link runs on the driver's real background asyncio loop, so the
loop-and-thread plumbing is exercised rather than mocked.

Break table: **17 mutations, 17 detected, 17 distinct failing-id sets, control clean,
source restored byte-identical.** Two mutations initially fired nothing and both were
real holes in the tests, now closed:

| mutation | fires | note |
|---|---|---|
| envelope check dropped from `send_action` | 7 | |
| per-axis travel check dropped | 6 | |
| pairwise coupling check dropped | 1 | the only test that can see it |
| `body_yaw` sent in degrees | 1 | |
| antennas sent in degrees | 1 | |
| `_pose` not derived from the quaternion | 4 | |
| boolean battery read as a percentage | 1 | `battery: true` is not 1 percent |
| snapshot hands out the live cache | 1 | |
| `stop()` claims a halt the daemon refused | 1 | |
| registry no longer declares the driver | 1 | |
| reachy dropped from the shipped table | 5 | |
| unknown-action refusal dropped | 1 | a silent successful no-op |
| `robot_name` mismatch applied anyway | 1 | |
| envelope limits restated in the driver | 1 | |
| link start failure swallowed | 1 | |
| `connect_eagerly` rebuilds an existing link | 1 | **was 0**: the link double answered with itself, so identity held even on a rebuild. Now graded by build count. |
| finiteness check dropped from `send_action` | 10 | **was 0**: the envelope only bounds the axes it knows, so the driver's own pass is the sole guard for `head_x/y/z` and the antennas. A `nan` there reached `rpy_to_pose` and put a matrix of nans on the wire, reported as success. Now covered for all five pass-through keys. |
| CONTROL - no change | 0 | |

## Gate

- `mypy`: **24 == 24** against a pristine `upstream/main` worktree, normalized message
  sets byte-identical in both directions, 0 in the changed files.
- Paired run, identical 574-file selection (the driver area plus every guard that walks
  the tree, reads docstrings or source, reads the registry, or reads `changelog.d`), the
  new test file excluded from both trees: **22 failed, 190 skipped on both**, failing-ID
  sets identical, `comm` empty both directions. All 22 pre-existing: 17 need `awsiot`
  (declared in the `[mesh-iot]` extra, absent here), 2 a lerobot-from-source `encode()`
  signature change, 2 a `NoneType` subscript, 1 a leftover dataset in `/tmp`.
- The `+13` collected delta localises entirely to two tree-walking guards, both passing:
  `test_args_docstring_completeness.py` 462 vs 452 (the new modules' documented
  functions) and `test_finalizer_reports_only_real_cleanup_failures.py` 282 vs 279 (the
  new class with a `cleanup`).
- Re-verified after rebasing onto `0d3f2ab6`: 1529 passed, 1 skipped across the driver
  suites, the seam tests, both moved guards and every docs guard - including the
  real-mode example grader added by #2748.

No visual artifact: this changes no policy, simulation behaviour, rendering, recording
or asset. The measured before/after and envelope tables above are the evidence.

## Not verified here

No Reachy Mini is attached to the machine this was developed on, so every daemon call is
mocked. What that leaves unproven is the live bring-up: whether `/api/daemon/status`
carries a battery field on either variant, and the two-Mini mesh bring-up with distinct
peer ids end to end. The driver-level half of that is tested (two drivers keep distinct
peer ids, Zenoh prefixes, hosts and caches, and commanding one does not command the
other), but the wire-level confirmation needs hardware.

### Round 1

| concern | commit | resolution |
|---|---|---|
| The four lazy `reachy_transport` imports crashed with a bare `ModuleNotFoundError: No module named 'device_connect_edge'` on any install without the `[device-connect]` extra, escaping `connect_eagerly`'s documented "named reason on failure" and the "return error dicts, never raise" rule for agent-tool surfaces | `37cf8efc` | One `_resolve_transport()` helper shaped like `G1Driver._resolve_message_class` returns the module or a reason naming `pip install 'strands-robots[device-connect]'`. Every transport touch routes through it. |
| — same concern, `connect_eagerly` reason quality | `37cf8efc` | Resolved *before* the daemon probe. Routed through the probe the reason came back wrapped as `daemon unreachable (host:port)`, which would send an operator to check a network the call never touched. |
| — same concern, `_wire_commands` | `37cf8efc` | Also returns the reason rather than relying on the connected gate in front of it staying there, so no cross-method invariant is load-bearing for the no-raise contract. |
| Merge conflict with the Dynamixel driver that landed in `main` (#2750), which added its registration in the old per-driver-block style this PR replaced with `_SHIPPED_DRIVERS` | `c4d821d7` | Dynamixel folded into the table as a third entry. Its robot set is `SUPPORTED_ROBOTS` on its own module, so a table entry's names may now be a literal tuple **or** the name of an attribute holding them — restating the family here would be a second list to drift. Ordering rationale from `main` preserved: Dynamixel first. |

#### Measured before / after (stock install, no extras)

| call | before | after |
|---|---|---|
| `connect_eagerly()` | `RAISED ModuleNotFoundError: No module named 'device_connect_edge'` | returns `cannot import strands_robots.device_connect.reachy_transport: ... - the Reachy transport helpers ship behind an extra: pip install 'strands-robots[device-connect]'` |
| `send_action({"head_pitch": 5.0})` | `RAISED ModuleNotFoundError` | `{"status": "error"}` naming the same extra |
| `stop()` | `RAISED ModuleNotFoundError` | returns; `_stopped` stays `False` — a stop that never reached the daemon is not a halt |

#### Verification

- **6 new tests** (`TestAStockInstallIsRefusedByNameRatherThanCrashing`) pin the reason text, the no-raise contract on all three verbs, the network-misattribution guard, that `stop` claims no halt, and that the module still imports and registers without the extra. The absence is simulated as Python signals it: `sys.modules[name] = None`.
- **Paired run**, both driver suites, before/after this round: failing-ID sets **identical** (72 both, every one a pre-existing `device_connect_edge` absence on a box without the extra; CI installs `[all]`), passed **78 → 84** — exactly the 6 added, zero regressions, zero incidental fixes.
- **Registration after the conflict resolution**, measured through the public seam: 3 table entries; `koch`, `aloha`, `vx300s`, `wx250s`, `trossen_wxai`, `dynamixel_2r` → `DynamixelDriver` (0 of `SUPPORTED_ROBOTS` unregistered); `g1` / `unitree_g1` → `G1Driver`; `reachy_mini` → `ReachyDriver`.
- `ruff check`, `ruff format --check` and `mypy` clean on every changed file.

#### Follow-up filed

The refusal is honest but the install requirement it names is not real: `reachy_transport` is a **stdlib-only leaf**, trapped only by its parent package `__init__` importing `device_connect_edge` and the three Device Connect drivers unconditionally. Making the package extra-free is a four-module change and is deliberately not in this PR. Tracked as #2771 (on the board) with the preferred fix — move the leaf out from under the extra-bearing package and re-export — and an acceptance list that keeps these 6 refusal tests meaningful.


strands_robots/drivers/reachy.py
```diff
@@ -0,0 +1,867 @@
...
+
+        Args:
+            path: Request path, one of this module's ``_PATH_*`` constants.
+
+        Returns:
+            The decoded body, or ``{"error": ...}`` - the shape
+            :func:`~strands_robots.device_connect.reachy_transport.api` returns
+            for every failure, which is why no call here needs a ``try``.
+        """
+        from strands_robots.device_connect.reachy_transport import api
```
yinsong1986: [MUST FIX] This lazy import crashes with a bare `ModuleNotFoundError` on any install without the `[device-connect]` extra, escaping every no-raise surface the driver documents.

`from strands_robots.device_connect.reachy_transport import api` executes the *package* `__init__`, and `strands_robots/device_connect/__init__.py` line 24 does `from device_connect_edge import DeviceRuntime` unconditionally - `device-connect-edge` lives only in the `[device-connect]` / `[all]` extras, not in core deps. `reachy_transport` itself needs nothing from it; the crash is purely the package-init side effect. Measured on a stock install (no extras):

```
connect_eagerly RAISED: ModuleNotFoundError: No module named 'device_connect_edge'
send_action     RAISED: ModuleNotFoundError: No module named 'device_connect_edge'
```

Why this can't be deferred: it is a crash bug on the changed path with the most realistic input there is - the default `pip install strands-robots`. The module-load discipline here is careful enough that the driver *imports*, passes the `_SHIPPED_DRIVERS` guard, registers, and `Robot("reachy_mini", mode="real")` builds - so the failure arrives at first daemon touch, as an exception naming a package the caller never asked for. That breaks three documented contracts at once: `connect_eagerly`'s own "Returns: ... A named reason on failure - the driver is left disconnected but usable"; AGENTS.md > Review Learnings (PR #85) > Error Handling Contracts > "Return error dicts, never raise" (the `stream` `stop` verb and `send_action` both raise through the agent tool surface); and AGENTS.md > Key Conventions #7 ("Use `require_optional()` ... for all optional deps"). CI never sees it because the hatch env installs `[all]` - the same absence also fails 72 of this PR's 169 driver tests in a no-extra environment, at the `monkeypatch.setattr("strands_robots.device_connect.reachy_transport.api", ...)` line.

Resolution: follow the seam's own precedent - `G1Driver._resolve_message_class` wraps its lazy SDK import and returns a reason string that `connect_eagerly` turns into a named connect failure. Wrap the four lazy transport imports here (`_daemon_get`, `_daemon_post`, `_build_link`, `_wire_commands`) in `try/except ImportError` returning a reason that names `pip install 'strands-robots[device-connect]'`, or route them through `require_optional()` and catch at the driver's refusal boundaries. (A module-level guarded import that lets the `_SHIPPED_DRIVERS` guard skip registration - the G1-without-SDK posture, where the factory refuses by name - is an acceptable alternative posture.)

yinsong1986 (commented):
## Summary

Adds `strands_robots.drivers.reachy.ReachyDriver`, the second native driver through the `HardwareDriver` seam, so `Robot("reachy_mini", mode="real")` builds instead of raising lerobot's `Unsupported robot type`. The driver reuses the shipped `device_connect.reachy_transport` (REST probe, WebSocket/Zenoh links, `rpy_to_pose`), derives `_pose` from the measured IMU quaternion rather than invented Stewart kinematics, and refuses out-of-envelope motion via a shared `strands_robots.tools.reachy` envelope that checks the head-body yaw coupling as a pair. Registration becomes a derived `_SHIPPED_DRIVERS` table, the registry declares `hardware.driver = "strands"` for the Mini, and two already-false statements in the robot-factory docs are corrected. I verified the wire conventions (argument order and units of `rpy_to_pose`, the `{"error": ...}` shape of `api()`, the `wireless_version` variant read) against the shipped Device Connect driver - they match - and the seam contract, alias resolution, envelope refusals and unit conversions are all pinned by the 103-case test file. One must-fix, flagged inline: the driver's lazy transport imports execute `strands_robots/device_connect/__init__.py`, which unconditionally imports the optional `device_connect_edge` package, so on an install without the `[device-connect]` extra every daemon touch raises a bare `ModuleNotFoundError` instead of the named refusal the driver's own contract (and the G1 precedent) requires.

## What's good

- Wire details are read off the shipped Device Connect driver rather than restated: `rpy_to_pose(pitch, roll, yaw, x_mm, y_mm, z_mm)` call order, radians conversions, and the `is_lite = not status.get("wireless_version", True)` read all match `reachy_mini_driver.py`.
- The envelope refuses and names the limit (no silent clamp), the coupling check runs only after per-axis checks pass, and a regression test asserts the driver imports the limits rather than restating them.
- The pass-through finiteness test even guards its own premise (`envelope_error({key: nan}) is None` before grading the driver's pass) so it cannot silently stop testing what it claims.
- Changelog fragment present; no non-ASCII in any new file; every `Args:`/`Raises:` block complete; registration table failure isolates per driver.

## Verification suggestions

- Reproduce the inline finding in a venv without the extra: `pip install -e .` (no extras), then `python -c "from strands_robots.drivers import get_native_driver_class; d = get_native_driver_class('reachy_mini')(port='x'); print(d.connect_eagerly())"` - currently raises `ModuleNotFoundError: No module named 'device_connect_edge'` instead of returning a reason. The same environment fails 72 of the new file's 169 collected driver tests (the `monkeypatch.setattr` target needs the same import), which CI never sees because the hatch env installs `[all]`.

cagataycali:
## Review round 2 - measured against `f706f9de`

The `[MUST FIX]` is closed, and two of the decisions in `37cf8efc` are better than the
ones I had reached independently, so I want to record those before the gaps.

**The pre-probe ordering is right and I had it wrong.** I had `_build_link` carry the
refusal and let `connect_eagerly` reach it through the probe. Measured, that produces:

```
connect_eagerly -> "daemon unreachable (reachy-a.local:8000): cannot import
                    strands_robots.device_connect.reachy_transport: ..."
```

which is exactly the misdiagnosis the comment on `connect_eagerly` describes - an operator
sent to check a network the call never touched. Resolving before the probe is the correct
arrangement, and `test_connect_eagerly_does_not_blame_the_network` is the cell that pins it
(reverting the ordering fires it, and only it).

**The merge resolution is sound, including the part that could have gone wrong silently.**
#2750 returned the seam to per-driver blocks on main while this branch had refactored it
to `_SHIPPED_DRIVERS`; keeping the table and folding Dynamixel in through a
`SUPPORTED_ROBOTS` indirection preserves both. Checked by registration rather than by
reading the diff:

| tree | `len(list_native_drivers())` | names |
|---|---|---|
| `upstream/main` | 7 | aloha, dynamixel_2r, koch, trossen_wxai, unitree_g1, vx300s, wx250s |
| `f706f9de` | 8 | the same 7, plus `reachy_mini` |

Nothing from #2750 was dropped.

### Gap 1: four of the guards are unreachable on the ordinary route, and ungraded

Because `connect_eagerly` now resolves first, nothing downstream of it can ever see a
reason - which is correct, and it means each remaining guard is belt-and-braces. That is
worth keeping; it is not worth leaving ungraded, because a belt-and-braces guard nothing
exercises is indistinguishable from one that was never written.

Mutation table, run against `f706f9de` with `tests/drivers/test_reachy_driver.py`
(109 collected, source restored byte-identically after each row, md5 verified):

| mutation | fires |
|---|---|
| control, no mutation | 0 |
| `_resolve_transport` raises instead of returning a reason | 5 |
| `connect_eagerly` resolves after the probe | 1 |
| `_build_link` drops its guard | **0** |
| `_daemon_get` drops its guard | **0** |
| `_daemon_post` drops its guard | 1 |
| `_wire_commands` drops its guard | 1 |
| a new daemon touch uses a bare `from ... import api` | **0** |
| the module name restated at the call site instead of read from `_TRANSPORT_MODULE` | **0** |

`_daemon_post` and `_wire_commands` are covered - by
`test_stop_does_not_claim_a_halt_it_could_not_request` and
`test_send_action_refuses_rather_than_raising` respectively. `_build_link` and
`_daemon_get` are not, and neither are the two structural properties.

The last two matter most for what happens next: a fifth daemon touch written with a plain
`from strands_robots.device_connect.reachy_transport import api` passes every behavioural
cell in the file, because CI installs `[all]`, and crashes on a stock install exactly as
the first four did. That is the same failure mode this PR just fixed, reachable by adding
a method.

Applying the cells below to `f706f9de` takes those four rows to 2, 1, 1, 1 with the control
still at 0, and the suite from 342 to 347 passed:

```python
# in TestAStockInstallIsRefusedByNameRatherThanCrashing - the guards the
# connected/pre-probe route can no longer reach, driven directly.
    @pytest.mark.parametrize("is_lite", [True, False], ids=["lite", "wireless"])
    def test_building_a_link_returns_a_reason_for_either_variant(
        self, monkeypatch: pytest.MonkeyPatch, is_lite: bool
    ) -> None:
        self._block_transport(monkeypatch)
        link = ReachyDriver(port="reachy-a.local:8000", transport=object())._build_link(is_lite=is_lite)
        assert isinstance(link, str)
        assert "strands-robots[device-connect]" in link

    def test_a_daemon_read_reports_the_error_shape_it_documents(self, monkeypatch: pytest.MonkeyPatch) -> None:
        self._block_transport(monkeypatch)
        body = ReachyDriver(port="reachy-a.local:8000")._daemon_get(reachy_mod._PATH_STATUS)
        assert set(body) == {"error"}
        assert "strands-robots[device-connect]" in body["error"]


class TestTheTransportIsResolvedInOnePlace:
    """A daemon touch added later cannot reintroduce a bare transport import."""

    def test_the_driver_module_imports_the_transport_nowhere(self) -> None:
        tree = ast.parse(inspect.getsource(reachy_mod))
        offenders = [
            ast.unparse(node)
            for node in ast.walk(tree)
            if isinstance(node, ast.ImportFrom) and (node.module or "").startswith("strands_robots.device_connect")
        ]
        assert offenders == [], f"resolve these through _resolve_transport() instead: {offenders}"

    def test_the_module_is_named_once_and_the_resolver_reads_that_name(self) -> None:
        source = inspect.getsource(reachy_mod)
        assert source.count(f'"{reachy_mod._TRANSPORT_MODULE}"') == 1
        assert "_TRANSPORT_MODULE" in inspect.getsource(reachy_mod._resolve_transport)
```

(`ast` and `inspect` at module scope.)

### Gap 2: the suite still cannot run on the install this PR just fixed

The commit message classifies the 72 stock-install failures as pre-existing absences, which
is accurate as a cause. It is still an odd end state: the change makes the *driver* report a
missing extra by name, while the driver's own suite errors with `ImportError` on that same
install. Measured on `f706f9de` with `device_connect_edge` unimportable:

```
72 failed, 270 passed
```

They fail at `monkeypatch.setattr("strands_robots.device_connect.reachy_transport.api", ...)`
and at three inline imports of `WebSocketLink`, `ZenohLink` and `rpy_to_pose`. Those cells
genuinely need the transport - they replace its `api()` and use its real `rpy_to_pose` as the
wire-format oracle - so the honest report is a skip, not an error:

```python
def _require_transport() -> Any:
    """Return the daemon transport module, skipping the caller if it is absent."""
    return pytest.importorskip(
        reachy_mod._TRANSPORT_MODULE,
        reason="needs the daemon transport, which the [device-connect] extra supplies",
    )
```

then `monkeypatch.setattr(_require_transport(), "api", daemon)` in `_install`, and
`WebSocketLink = _require_transport().WebSocketLink` (likewise `ZenohLink`, `rpy_to_pose`)
at the three inline imports. One more site needs it:
`test_a_wireless_without_a_transport_is_refused_by_name` calls the real `_build_link`, so
without the transport present it refuses for the missing-extra reason and the `"Zenoh" in
link` assertion fails on the wrong refusal.

That is deliberately scoped to the 31 cells that need nothing - they keep running - so the
result is `275 passed, 72 skipped, 0 failed` on a stock install and `347 passed` with the
extra, rather than skipping the whole file.

### Not asking for these

`require_optional()` instead of `importlib.import_module` would satisfy convention 7 and
put the install line in the shared helper, but it caches resolved modules in
`utils._lazy_modules`, and `_block_transport`'s `sys.modules[name] = None` cannot defeat a
cache hit - so the tests would depend on resolution order. `importlib.import_module` matches
`_resolve_message_class` and stays testable; that is the better trade here.

Making `strands_robots/device_connect/__init__.py` extra-free is the root cause and the
commit message is right to hold it out of this change: the three drivers it imports subclass
`device_connect_edge.drivers.DeviceDriver`, so it needs PEP 562 lazy exports across four
modules. Worth noting that if that lands, the skips above become unreachable rather than
wrong.

Gate on `f706f9de` plus the cells above: `ruff check`/`ruff format --check` clean, mypy
24 == 24 against `b7544254` with byte-identical normalized message sets and 0 in the
changed files, and a 464-file paired run (all of `tests/drivers`, the seam, and every guard
that walks the tree or reads docs) with identical failing-ID sets, `comm` empty both ways -
19 failures on each side, 17 needing `awsiot` from `[mesh-iot]` and 2 pre-existing in
`test_device_connect_hardening.py`.

cagataycali:
Pushed `e00e99eb`, which completes the transport-extra fix by grading the two guards nothing was reaching.

## What was unpinned

`_resolve_transport` is consulted at four sites. The tests added with the fix drive three of them, through `connect_eagerly`, `send_action` and `stop`. They cannot reach the other two, and the reason is the fix's own (correct) ordering: `connect_eagerly` resolves the transport *before* it probes the daemon, so with that pre-check in place neither `_daemon_get` nor `_build_link` is ever entered with the extra absent.

Mutating the driver and counting which tests notice — control clean and source restored byte-identical in both columns, `collected` 109 then 118:

| mutation | before | after |
|---|---|---|
| `connect_eagerly` pre-check removed | 1 | 1 |
| `_wire_commands` returns `[]` not the reason | 1 | 2 |
| `_daemon_get` guard removed | **0** | **1** |
| `_daemon_post` guard removed | 1 | 1 |
| `_build_link` guard removed | **0** | **1** |
| resolver raises instead of reporting | 5 | 10 |
| reason omits the pip remedy | 3 | 6 |

One note on reading that `_build_link` row: dropping the guard while importing only `WebSocketLink` leaves the wireless branch naming an undefined `ZenohLink`, so a pre-existing test fails for a reason unrelated to the guard and the row looks like a 1. Importing both names gives the faithful mutation, and it goes unnoticed.

The depth here is not hypothetical. `_build_link`'s own docstring says it is kept as a method so a caller can substitute a link, and both daemon helpers are ordinary callables a subclass or a bring-up script reaches without going through `connect_eagerly` first. Grading them directly means the pre-check is not the only thing between a stock install and a raise: whichever layer a later change removes, something fails.

## What else the new file pins

- **The premise the approach rests on.** The transport leaf never names `device_connect_edge`, and the parent package `__init__` is what pulls it in. If the leaf genuinely needed the extra, requiring it would be right and a named refusal would be the wrong answer — so that asymmetry is worth a test rather than a comment.
- **The reason reaches a mesh peer.** `get_status` still succeeds and carries it as `connect_error`, so a Mini on an install without the extra is a describable peer rather than a silent one.
- **A refused action leaves no partial command on the link.**
- **An install that has the extra sends exactly what it sent before** (the over-reach control).

## Gate

- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ` (1595 files).
- mypy **19 == 19** against the parent commit, normalised message sets identical both directions, zero attributable — all 19 are pre-existing `examples/isaac_gs/`.
- `tests/drivers/` + `tests/test_driver_seam.py` + the changelog guards: **376 passed, 5 skipped**.
- Broad paired selection (363 files covering every driver suite and every guard that walks the tree or reads `docs/`): **18214 collected, 5 failed, 18131 passed**. The 5 are this box's own environment — they assert `rclpy` is absent and it resolves from a system ROS install; the failing-ID set is identical to the baseline before this change.

## Also

`strands_robots/drivers/__init__.py` had lost a blank line in the merge that brought the Dynamixel table in, which failed `ruff format --check` and so the required check. Both merge parents were clean on their own; only the result was not. That is already fixed in `f706f9de`, so this push does not touch it — flagging it only because it is the kind of thing a merge produces silently and it is worth a glance on the next merge into this branch.

The docs page and the changelog fragment now state the behaviour as well: a native driver may need an extra to reach its hardware and says so rather than raising, with the reason quoted verbatim from the driver.

cagataycali:
## CodeQL 983 is a one-line edit, and the shape that trips it is measurable

`e00e99eb` added `tests/drivers/test_reachy_transport_guards_are_reachable.py`, and the
CodeQL scan that followed it opened alert 983 (`py/unnecessary-delete`) on line 84. That
thread is `isResolved: false` and, unlike the earlier `[MUST FIX]`, it is not outdated - so
with `required_review_thread_resolution: true` on the `default` ruleset it is a hard merge
gate regardless of its `warning` severity. Flagging it because the last push predates the
scan by five minutes, so it may not have been visible yet.

### The idiom is fine; one property of this instance is not

`del <unused parameters>` is established here - 25 sites across the tree. Twenty-four of
them are unflagged, and the difference from the twenty-fifth is total:

| shape | sites | CodeQL |
|---|---|---|
| the `del` is followed by more statements | 24 | silent on all 24 |
| the `del` is the terminal statement of the function | 1 | alert 983 |

The twenty-four include twelve in shipped code (`drivers/g1.py` x6,
`drivers/dynamixel/driver.py` x5, `drivers/reachy.py` x4, `policies/lerobot_local/policy.py`,
`policies/protomotions/{policy,motion_utils}.py`) and are not at issue.

The reason the split falls there is that a parameter `del` narrows scope for the body that
follows it. As the terminal statement it narrows nothing - the frame is discarded on return
either way - so it is a genuine no-op statement rather than a scope-narrowing one.

`_StubLink` already contains the answer, two lines below the flagged one:

```python
    async def start(self, on_joints: Any, on_imu: Any) -> None:
        """Accept the driver's callbacks."""
        del on_joints, on_imu          # line 84 - terminal, flagged

    async def stop(self) -> None:
        """Accept teardown."""         # docstring only, no del
```

Both methods do the same thing - accept arguments and ignore them - and only one of them
opens a review thread. `send_cmd` and `stop` are the class's own precedent.

### The edit, and what it measures

Deleting line 84 leaves the docstring as the body, matching `stop`:

| check | with line 84 | line 84 removed |
|---|---|---|
| `pytest tests/drivers/test_reachy_transport_guards_are_reachable.py tests/drivers/test_reachy_driver.py` | 118 passed | 118 passed |
| `ruff check` | clean | clean |
| `ruff format --check` | clean | clean |
| mypy errors attributable to the file | 0 | 0 |

Nothing is given up by removing it, because nothing was enforcing it: `ARG` is not in
`[tool.ruff.lint] select` (the list is `E, W, F, I, UP, B015, B018`), so the unused
parameters raise no lint with or without the `del`. Checked directly on a constructed pair
of functions - ruff is silent on a terminal parameter `del` under the repo's select list,
and also under `--select ALL`; no ruff rule covers it.

### Not a config exclusion

Worth saying explicitly, since the neighbouring precedent points the other way:
`.github/codeql/codeql-config.yml` excludes `py/ineffectual-statement` and
`py/import-and-import-from` for idioms the codebase is obliged to use. This is not that
case - the idiom has 24 unflagged instances and a rewrite available in the same class - and
the config's own scope discipline says it "excludes TWO rule ids and nothing else", pinned
by `tests/test_codeql_query_filters.py`. Widening it here would be the reviewed edit that
file exists to force.

### One observation about where this was caught

That config's comment also states the no-op-statement capability given up by excluding
`py/ineffectual-statement` "is NOT given up -- it moves to ruff, which is merge-blocking
here where CodeQL is advisory". For this shape that holds only in the first half: the class
is caught, but by the scanner that runs after a push, as a thread that blocks the merge,
rather than by the local `ruff`/`mypy`/`pytest` gate - which passes on it. That is a
property of the tree rather than of this PR, and I am picking it up separately; noting it
here only because it explains why a green local gate preceded the alert.

docs/getting-started/robot-factory.md
```diff
@@ -113,17 +114,40 @@ Asking for a driver that is not there is refused, never quietly substituted:
...
+to reach it and `driver="lerobot"` is refused by name.
+
+A native driver may need an extra to reach its hardware, and says so rather than
+raising. The Reachy Mini's daemon transport ships in the
+[`device-connect`](../device-connect.md) extra, so on a core install the driver still
+builds, registers and answers `get_status` - and every surface that would touch the
+daemon returns a reason naming the extra instead:
+
+```python
+>>> Robot("reachy_mini", mode="real").driver.connect_eagerly()
```
yinsong1986: [MUST FIX] This documented invocation raises `AttributeError` as written - the factory returns the driver itself, and `ReachyDriver` has no `.driver` attribute.

For a robot whose registry entry declares `hardware.driver = "strands"`, `Robot(name, mode="real")` takes the `resolved_driver == "strands"` branch and returns `_build_native_driver(...)` directly (the `driver: Literal["strands"]` overload is typed `-> HardwareDriver` for exactly this reason). Verified against this head:

```
>>> r = Robot("reachy_mini", mode="real")
>>> type(r).__name__
'ReachyDriver'
>>> r.driver.connect_eagerly()
AttributeError: 'ReachyDriver' object has no attribute 'driver'
```

The quoted output below the snippet is what `connect_eagerly()` returns when called on the driver directly, so the fix is to drop `.driver`:

```python
>>> Robot("reachy_mini", mode="real").connect_eagerly()
```

Why this can't be deferred: this is the published bring-up line for the feature this PR ships, on the one page whose examples the repo keeps two graders on because "it is the line a reader copies verbatim" - and neither grader can catch this class of error, so nothing surfaces it after merge. `tests/test_docs_real_mode_invocations.py` grades the robot name and keywords inside the `Robot(...)` call, and `tests/test_docs_python_examples_are_callable.py` grades call binding; attribute access on the factory's *return value* is visible to neither, which is why this PR's gate (including the #2748 grader) is green over a snippet that crashes. Beyond the crash, the example teaches a wrong API shape - that a registry-declared native driver comes back wrapped in something carrying `.driver` - on the same page this PR corrects two other statements for being "already false"; shipping a third false statement in the same change is the outcome the page's own guards exist to prevent.

yinsong1986 (commented):
## Summary

Adds `strands_robots.drivers.reachy.ReachyDriver`, the second native driver through the `HardwareDriver` seam, so `Robot("reachy_mini", mode="real")` builds a daemon-backed driver instead of raising lerobot's `Unsupported robot type`. The driver reuses the shipped `device_connect.reachy_transport` (REST probe, WebSocket/Zenoh links, `rpy_to_pose`), derives `_pose` from the measured IMU quaternion, and refuses out-of-envelope motion through the shared `strands_robots.tools.reachy` envelope, including the pairwise head-body yaw coupling check. Registration becomes the derived `_SHIPPED_DRIVERS` table, the registry declares `hardware.driver = "strands"` for the Mini, and the round-1 concern (bare `ModuleNotFoundError` on installs without the `[device-connect]` extra) is resolved: every transport touch now routes through `_resolve_transport()`, the refusal names the extra, and the six new refusal tests plus the guard-reachability file pin it - including the two guards the connect pre-check hides. I re-verified the wire conventions against the shipped transport at this head (`rpy_to_pose(pitch, roll, yaw, x_mm, y_mm, z_mm)` takes degrees and converts internally; `api()` returns `{"error": ...}` for every failure, so the no-`try` design is sound), and ran both new test files against this head with the extra importable: 118/118 pass, as do the registry, docstring-completeness, export and ASCII guards. One must-fix, flagged inline on the docs page: the new extra-refusal example calls the factory result through a `.driver` attribute the returned driver does not have, so the documented line raises `AttributeError` instead of printing the refusal it quotes.

## What's good

- The round-1 fix follows the seam's own precedent (`G1Driver._resolve_message_class` shape), resolves the transport *before* the daemon probe so a missing extra is never misreported as an unreachable network, and `tests/drivers/test_reachy_transport_guards_are_reachable.py` grades the two guards the pre-check hides - defence in depth is pinned, not assumed.
- The envelope refuses rather than clamps, names the limit, checks the yaw pair only after both values are individually legal, and the driver imports the limits rather than restating them (pinned by a source-scan test).
- `stop()` / `stop_task()` never claim a halt the daemon refused, and a stop that could not reach the daemon leaves `_stopped` False - the safety-path honesty AGENTS.md demands.

yinsong1986 (approved):
## Summary

Adds `strands_robots.drivers.reachy.ReachyDriver`, the second native driver through the `HardwareDriver` seam, so `Robot("reachy_mini", mode="real")` builds a daemon-backed driver instead of raising lerobot's `Unsupported robot type`. The driver reuses the shipped `device_connect.reachy_transport` (REST probe, WebSocket/Zenoh links, `rpy_to_pose`), derives `_pose` from the measured IMU quaternion rather than invented Stewart kinematics, and refuses out-of-envelope motion via the shared `strands_robots.tools.reachy` envelope, including the head-body yaw coupling checked as a pair. Registration becomes the derived `_SHIPPED_DRIVERS` table (now three entries, with the Dynamixel driver from #2750 folded in), the registry declares `hardware.driver = "strands"` for the Mini, and the robot-factory docs are corrected against measured output. Both prior must-fix findings are resolved at this head: every transport touch now routes through `_resolve_transport()` and I verified on an install without the `[device-connect]` extra that `connect_eagerly` returns the reason naming `pip install 'strands-robots[device-connect]'` (not wrapped as `daemon unreachable`), `send_action` returns an error envelope, `stop` claims no halt, and the driver still imports and registers; the docs example now calls `.connect_eagerly()` on the factory return, which is the driver itself. The CodeQL `py/unnecessary-delete` alert is closed by the `_ = (on_joints, on_imu)` rewrite.

## What's good

- Wire details verified against the shipped `reachy_transport.py` rather than restated: `api()` really does return `{"error": ...}` for every failure (so the no-`try` REST calls are sound), `rpy_to_pose(pitch, roll, yaw, x, y, z)` takes degrees/mm in that order, `body_yaw`/antennas go out in radians, and `is_lite = not status.get("wireless_version", True)` matches the Device Connect driver's read of the same field.
- The envelope refuses and names the limit (no silent clamp), the coupling check runs only after both values are known finite and individually legal, and a regression test pins that the driver imports the limits rather than restating them.
- The stock-install refusal tests simulate the absence the way Python signals it (`sys.modules[name] = None`) and grade the two guards `connect_eagerly`'s pre-check hides, so the defence in depth is pinned rather than assumed.
- A daemon that refuses the stop is not reported as a halt (`_stopped` stays `False`), on both `stop` and `stop_task`, including when the transport itself is missing.
- Registration failure isolates per driver; alias resolution, seam membership and the two-Mini isolation properties all verified live through the public seam.

PR #2763, opened 2026-08-26
fix(mesh): read joints from a driver that owns its bus

### LINKED ISSUE #2749: drivers: a contract-complete native driver publishes no joint telemetry on the mesh

cagataycali:
## Problem

A robot reaches its motors one of two ways. A lerobot robot is a **wrapper**:
`hardware_robot.Robot` holds the device that owns the bus under `robot`, and the wrapper
answers no read itself. A **native driver** owns the bus directly, so it *is* the device.

`Mesh._read_state` resolved only the first shape:

```python
inner = getattr(r, "robot", None)
if inner is not None and hasattr(inner, "get_observation") and getattr(inner, "is_connected", False):
    obs = read_joints(inner)
```

So a driver satisfying every member of `DRIVER_SURFACE` published no joint telemetry at
all. Every other section reaches the mesh through a `getattr(robot, name, None)` read
straight off the driver -- `_imu`, `_battery`, `_temps` and thirteen siblings -- so such a
driver could report **how hot a joint was and not where it was**, and
`missing_driver_members` reported no problem. The presence heartbeat kept advertising the
peer either way, which is the presentation `_read_state`'s own docstring exists to avoid:
"it went silent on the state topic while its presence heartbeat kept advertising it --
indistinguishable from a peer whose state thread had died".

Measured on `main`, driving the real `Mesh._read_state` over three driver shapes, each
contract-complete (`missing_driver_members(cls) == ()`):

| driver shape | `_read_state()` | `joints` | `read_joints(driver)` called directly |
|---|---|---|---|
| lerobot wrapper (inner device) | dict | 3 motors | `AttributeError` -- the wrapper answers no read |
| native, `get_observation` + bus | **`None`** | **absent** | **OK, 3 motors** |
| native, bus only | **`None`** | **absent** | **OK, 3 motors** |

The last column is the point: **`read_joints` could already read both native drivers,
unchanged. Nothing handed them to it.** The reader was never the missing piece -- it
prefers `bus.sync_read` and falls back to `get_observation` -- so the defect is entirely
in how the *device* was resolved.

The first row is the control that shapes the fix: a wrapper genuinely cannot be read, so
an inner device must still be **preferred**, not merely tried.

## Change

`bus_access.joint_read_source(robot)` resolves the device once: prefer `robot.robot`, fall
back to the robot itself, and admit it only when `read_joints` has a route to it. It lives
beside `read_joints` because its admission rule is *derived from* that function's own
branch rather than restated -- keeping the two in one module is what stops a caller
admitting a device the reader then raises on, or refusing one it could have read.

Deriving the rule also repairs a narrower disagreement at the same site: the old gate
required `get_observation`, which is the capability `read_joints` treats as its
**fallback**, and refused a device carrying only the `bus` it **prefers** -- the shape an
SO-100/SO-101 driver has, where joint position, velocity and current are the whole
telemetry.

An inner device is preferred whenever one is present, so a wrapper is never read in place
of the device it wraps, and a robot that already published joints resolves to the same
device as before. `is_connected` remains the liveness gate.

`drivers/base.py` documented the arrangement this changes ("The mesh reads observations
from the driver's *inner* device") and is updated in the same change.

## Why not the alternatives

Two other repairs were considered and are worse on measurement, not on taste:

- **Name the requirement in the `HardwareDriver` Protocol.** Not available:
  `missing_driver_members` checks `hasattr` on the class, and
  `hasattr(hardware_robot.Robot, 'robot' | 'get_observation' | 'is_connected' | 'bus')` is
  `False` for all four -- adding any of them makes the incumbent driver fail its own
  surface check.
- **Give joints a driver-level `_joints` cache**, like the other sixteen attributes. That
  duplicates `read_joints`, which already reads both shapes correctly.

Route 1 is also the only one that is backward-compatible by construction: when `robot.robot`
is present the resolved device is identical, so nothing about an existing peer moves.

## Tests

`tests/mesh/test_read_state_joints_from_a_driver_that_owns_its_bus.py`, 38 cases over five
driver shapes, driving the real `Mesh`, the real `read_joints` and its bus lock. Pre-fix
split (helper kept, call site reverted): **9 failed / 23 passed, and all 9 failures are in
the regression class** -- zero in the control, premise or derived classes. A raw revert is a
collection error, since the tests import the new symbol.

The suite asserts more than presence: the joints carry the values the bus reported, and the
snapshot survives `json.dumps` -- the step `session._put_zenoh_directly` performs before
`put`, where a refused payload is dropped for good. The numpy-scalar shape is what makes
that cell load-bearing, since `json.dumps` refuses `np.float32`.

Mutation table, 8 mutations, all 8 detected, 7 distinct firing sets, control clean, 38
collected in every row:

| mutation | cases fired |
|---|---|
| the fix removed (mesh resolves only `robot.robot`) | 12 |
| admission drops the bus branch (`get_observation` only) | 9 |
| admission drops the `get_observation` branch (bus only) | 4 |
| resolver skips the readability test | 4 |
| resolver ignores the inner device, always reads the robot | 3 |
| liveness gate (`is_connected`) dropped | 1 |
| resolver escapes to the wrapper when the inner device is unreadable | 1 |

The two four-case rows are why both halves of the admission rule are separately pinned: one
shape carries a bus and no `get_observation`, another carries `get_observation` and no bus,
so neither branch can be dropped unnoticed. `TestTheAdmissionRuleTracksTheReader` derives
the agreement between resolution and readability rather than listing it, with a non-vacuity
cell asserting both outcomes are reached.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1572 files).
- `mypy strands_robots tests tests_integ`: **`Success: no issues found in 1569 source files`**.
- Paired run over an identical 478-file selection (all of `tests/mesh`, all of
  `tests/drivers`, and every test that walks the tree, reads docstrings or names
  `bus_access`), with the new file excluded from both trees: **19054 collected and
  `5 failed, 18980 passed, 69 skipped` on both**, the five failing ids identical and both
  `comm` directions empty. All five expect `rclpy` to be absent and resolve it from a
  system ROS 2 install; they fail identically on an unmodified tree.
- Rebased onto `13fda66a`, which merged #2754. Its `TestTheDriverIsWhatTheMeshReads` pins
  `getattr(driver, "robot", None) is None` as a premise; that stays true here, because this
  change resolves the device on the mesh side and asks nothing of the driver. `tests/drivers`
  plus `tests/mesh/test_mesh.py` plus the new file: 168 passed.

## Scope

The state topic's `joints` section only. `inner is None` has two further consequences the
issue records -- presence drops `connected` and `hw`, and `_publish_cameras_once` routes a
real peer into `_publish_sim_cameras()` -- which are different subsystems with their own
semantics and are left for their own change. `G1Driver` declares neither `is_connected` nor
a readable bus today, so this opens the route rather than lighting up that driver: a native
driver publishes `joints` by exposing either a bus with `sync_read` or a `get_observation`,
plus `is_connected`, which `drivers/base.py` now says.

Closes #2749

### DIFF
--- a/changelog.d/2763-mesh-native-driver-joint-source.md
+++ b/changelog.d/2763-mesh-native-driver-joint-source.md
@@ -1,0 +1,35 @@
+### Fixed: the mesh reads joints from a driver that owns its bus
+
+A robot reaches its motors one of two ways. A lerobot robot is a wrapper: `hardware_robot.Robot`
+holds the device that owns the bus under `robot`, and the wrapper answers no read itself. A native
+driver owns the bus directly, so it *is* the device.
+
+`Mesh._read_state` resolved only the first shape, so a driver satisfying every member of
+`DRIVER_SURFACE` published no joint telemetry at all. Every other section reaches the mesh through a
+`getattr(robot, name, None)` read straight off the driver -- `_imu`, `_battery`, `_temps` and
+thirteen siblings -- so such a driver could report how hot a joint was and not where it was, while
+`missing_driver_members` reported no problem and the presence heartbeat kept advertising the peer.
+That is the presentation `_read_state`'s own docstring exists to avoid: a peer that "went silent on
+the state topic while its presence heartbeat kept advertising it -- indistinguishable from a peer
+whose state thread had died".
+
+`read_joints` could already read such a driver unchanged, preferring `bus.sync_read` and falling
+back to `get_observation`; nothing handed the driver to it. So the repair is to resolve the
+telemetry device rather than assume it. `bus_access.joint_read_source` prefers `robot.robot` and
+falls back to the robot itself, admitting a device exactly when `read_joints` has a route to it --
+the admission rule derived from that function's own branch rather than restated, so a caller cannot
+admit a device the reader raises on, or refuse one it could have read.
+
+Deriving the rule also repairs a narrower disagreement at the same site. The old gate required
+`get_observation`, which is the capability `read_joints` treats as its *fallback*, and refused a
+device carrying only the `bus` it *prefers* -- the shape an SO-100/SO-101 driver has, where joint
+position, velocity and current are the whole telemetry.
+
+An inner device is preferred whenever one is present, so a wrapper is never read in place of the
+device it wraps and a robot that already published joints resolves to the same device as before.
+`is_connected` remains the liveness gate, and a driver with no motors to report still publishes no
+`joints` without appearing as a failed probe in `degraded`.
+
+Scoped to the state topic's `joints` section. Two further consequences of a driver having no inner
+device are left for their own change: presence drops `connected` and `hw`, and
+`_publish_cameras_once` routes a real peer into `_publish_sim_cameras()`.
--- a/strands_robots/bus_access.py
+++ b/strands_robots/bus_access.py
@@ -224,3 +224,63 @@ def read_joints(device: Any) -> Any:
     # ``.pos`` is lerobot's own suffix (see SOFollower.get_observation) and the
     # shape the rest of this codebase already parses.
     return {f"{motor}.pos": value for motor, value in raw.items()}
+
+
+def _answers_a_joint_read(device: Any) -> bool:
+    """Whether :func:`read_joints` has a route to ``device``'s joint positions.
+
+    Derived from :func:`read_joints`'s own branch rather than restated: it reads
+    ``bus.sync_read`` when the device exposes one and falls back to
+    ``get_observation``, so a device carrying either can be read and a device
+    carrying neither cannot. Keeping the admission rule and the reader in one
+    module is what stops a caller admitting a device the reader then raises on,
+    or refusing one it could have read.
+
+    Args:
+        device: The candidate to test. Neither attribute is called.
+
+    Returns:
+        ``True`` when ``device`` exposes a ``bus`` with ``sync_read``, or a
+        ``get_observation``.
+    """
+    bus = getattr(device, "bus", None)
+    if bus is not None and hasattr(bus, "sync_read"):
+        return True
+    return hasattr(device, "get_observation")
+
+
+def joint_read_source(robot: Any) -> Any | None:
+    """The device to read ``robot``'s joint positions from, or ``None``.
+
+    A robot reaches its motors one of two ways. A lerobot robot is a *wrapper*:
+    :class:`strands_robots.hardware_robot.Robot` holds the device that owns the
+    bus under ``robot``, and the wrapper itself answers no read at all. A native
+    driver owns its bus directly, so it *is* the device.
+
+    Resolving only the first shape is what left the second publishing no joint
+    telemetry: a driver satisfying every member of
+    :data:`~strands_robots.drivers.DRIVER_SURFACE` reported its IMU, its lidar
+    summary and its motor temperatures - each read straight off the driver with a
+    ``getattr`` default - while ``joints`` was absent from the state topic,
+    because joints were the one section reached only through the inner device.
+    :func:`read_joints` could already read such a driver unchanged; nothing
+    handed it to it.
+
+    An inner device is still preferred whenever one is present, so a wrapper is
+    never read in place of the device it wraps.
+
+    Args:
+        robot: The object a telemetry consumer was handed - a lerobot wrapper, a
+            native driver, or anything else shaped like a driver.
+
+    Returns:
+        The inner device when ``robot`` has one that can answer a joint read;
+        ``robot`` itself when it has no inner device and can answer one; ``None``
+        when no joint read is possible, which callers report as "no joints" and
+        not as a failure.
+    """
+    device = getattr(robot, "robot", None)
+    if device is None:
+        # No inner device: a native driver owns its telemetry directly.
+        device = robot
+    return device if _answers_a_joint_read(device) else None
--- a/strands_robots/drivers/base.py
+++ b/strands_robots/drivers/base.py
@@ -12,11 +12,19 @@ learn which members are load-bearing.
 *measured* contract rather than an aspirational one, which makes it smaller than
 a reader might expect:
 
-* ``get_observation`` is **not** a member. The mesh reads observations from the
-  driver's *inner* device (``getattr(robot, "robot", None)`` in
-  :mod:`strands_robots.mesh.sensors`), and
+* ``get_observation`` is **not** a member. A lerobot robot is a wrapper: it
+  holds the device that owns the bus under ``robot``, and
   :func:`strands_robots.bus_access.read_observation` takes that inner device -
   so the top-level driver is not the thing asked for a frame.
+
+  A native driver has no inner device, and for joint telemetry that is now
+  resolved rather than assumed:
+  :func:`strands_robots.bus_access.joint_read_source` prefers ``robot.robot``
+  and falls back to the driver itself, so a driver that owns its bus publishes
+  ``joints`` on the state topic by exposing either a ``bus`` with ``sync_read``
+  or a ``get_observation``, plus ``is_connected`` to say it is live. Neither is
+  required, which is why neither is a member: a driver with no motors to report
+  is otherwise complete.
 * The sensor attributes a mesh publishes (``_pose``, ``_imu``, ``_battery`` and
   their siblings) are **not** members either. Every one is read with a
   ``getattr(robot, name, None)`` default, so a driver with no IMU publishes no
--- a/strands_robots/mesh/core.py
+++ b/strands_robots/mesh/core.py
@@ -23,7 +23,7 @@ from collections.abc import Callable, Mapping
 from typing import Any
 
 from strands_robots._mesh_switch import mesh_env_request
-from strands_robots.bus_access import read_joints, read_observation
+from strands_robots.bus_access import joint_read_source, read_joints, read_observation
 from strands_robots.mesh import security as _security
 from strands_robots.mesh.audit import log_safety_event
 from strands_robots.mesh.pacing import Ticker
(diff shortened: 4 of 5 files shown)


yinsong1986 (approved):
## Summary

This PR fixes `Mesh._read_state` so a native driver that owns its motor bus publishes `joints` on the state topic. Previously the probe resolved only the lerobot-wrapper shape (`getattr(r, "robot", None)` + `hasattr(inner, "get_observation")`), so a contract-complete native driver published every sensor except joint positions, and the old gate additionally demanded `get_observation` -- the capability `read_joints` treats as its fallback -- while refusing the `bus` it prefers. The repair is a new `bus_access.joint_read_source(robot)` that prefers the inner device, falls back to the robot itself, and admits a device exactly when `read_joints` has a route to it; `_read_state` delegates to it and keeps `is_connected` as the liveness gate. The change is a pure widening: when `robot.robot` is present the resolved device is identical to before, the read still goes through the device's `bus_access` lock (per AGENTS.md's shared-device-lock rule), and a driver with no motors publishes no `joints` without appearing in `degraded`.

## What's good

- The admission rule lives beside the reader it mirrors, and `TestTheAdmissionRuleTracksTheReader` derives their agreement rather than restating it, with a non-vacuity cell.
- Verified locally at `d6302409`: the new 38-case suite passes; reverting the `core.py` call site (helper kept) fails 12 cases, all in the regression class -- matching the PR's mutation table row 1.
- `tests/mesh/test_mesh.py`, `tests/drivers`, `tests/test_bus_access_serializes_motor_reads.py` (140 tests) pass on the head; the ASCII, Raises-docstring and test-naming guard suites pass over the new files.
- Changelog fragment present (`changelog.d/2763-mesh-native-driver-joint-source.md`); `Closes #2749` is in the body as a plain line; `drivers/base.py` docs updated in the same change; no non-ASCII, no host paths.
- The JSON round-trip test (`NativeNumpyBus`) pins the one encoder step between the snapshot and the wire, which is where a numpy scalar would silently drop the payload.

PR #2764, opened 2026-08-26
fix(tools): name the parameter a rosbridge port refusal was given

cagataycali:
## Problem

`use_rosbridge._transport_port_error` refuses a port that is legal TCP but outside the range the rosbridge WebSocket transport can address. It takes a `param` argument and documents it with one sentence:

```
param: The parameter name it came from, used in the message.
```

It never reads it. The message hard-codes the word `port`:

```python
return (
    f"{context}: port {port!r} is a legal TCP port that the rosbridge WebSocket "
    f"transport cannot address (it addresses 1-{_TRANSPORT_MAX_PORT}; autobahn's "
    "URL builder excludes the top of the range)"
)
```

That sentence is not local colour. Eleven other helpers in this package carry it verbatim and interpolate the argument. Derived by scanning the package for the sentence itself:

| helpers documenting `param` as "used in the message" | interpolate it |
|---|---|
| 12 | 11 |

The one that does not is this helper. (`simulation/mujoco/physics.py::_validate_mass` also never reads its `param`, but it is a `TYPE_CHECKING` protocol declaration with no body, so it builds no message and does not carry the sentence.)

### Why it matters, measured

The helper is shared *by design*. Its own docstring says it is applied by both the `use_rosbridge` tool and `RosbridgeRobot` so the two "cannot disagree about which ports it can carry", and the shared 16-bit domain that runs one line earlier **on the same value** does interpolate the name. So the two refusals a single caller can trip disagree about what they are talking about:

| `param` passed | `utils.tcp_port_error` (runs first) | `_transport_port_error` (runs next) |
|---|---|---|
| `port` | `publish: invalid port: 70000 (expected 1-65535)` | `publish: port 65535 is a legal TCP port that ...` |
| `bridge_port` | `publish: invalid bridge_port: 70000 (expected 1-65535)` | `publish: port 65535 is a legal TCP port that ...` |
| `ws_port` | `publish: invalid ws_port: 70000 (expected 1-65535)` | `publish: port 65535 is a legal TCP port that ...` |

The wide domain names the caller's parameter for every spelling; the transport names it only when the caller happens to spell it `port`.

### Reachability, stated plainly

This is **latent**. Both shipping call sites pass the literal `"port"`, read off the source rather than asserted:

```
use_rosbridge.py:352      _transport_port_error(port, "port", action)
rosbridge_robot.py:106    _transport_port_error(port, "port", type(self).__name__)
```

So no caller today sees a wrong parameter name. What is broken today is the documented contract: the argument exists, is advertised as reaching the message, and cannot. The docstring anticipates a third caller explicitly, and that caller is the one who would be handed a refusal naming a parameter it never passed.

## Change

Interpolate `param`, which is what the docstring already promises. One line:

```diff
-            f"{context}: port {port!r} is a legal TCP port that the rosbridge WebSocket "
+            f"{context}: {param} {port!r} is a legal TCP port that the rosbridge WebSocket "
```

Because both call sites pass `"port"`, **every refusal text reachable in production is byte-identical** before and after. Verified against the base implementation executed side by side in one process:

| context | base vs branch |
|---|---|
| `publish` | identical |
| `service_call` | identical |
| `RosbridgeRobot` | identical |

The domain is untouched: `1` / `8080` / `65534` still accepted, `65535` still refused, and `tcp_port_error` still carries the whole 16-bit space.

## Tests

`tests/tools/test_transport_refusal_names_its_parameter.py`, 30 cases. The contract is **derived from the promise sentence**, not from a list, so a twelfth helper that copies the sentence is graded the moment it lands. Because the package carries no violator after the fix, the scan cannot exercise its own failing branch, so the predicate is additionally graded against constructed exemplars (one keeping the promise, one breaking it) with a `{True, False}` non-vacuity assertion.

Pre-fix split: **6 failed / 24 passed** - 5 in `TestTheRefusalNamesTheParameterItWasGiven`, 1 in `TestEveryHelperThatPromisesThisKeepsIt`, and **0 in `TestTheShippingCallersSeeNoChange`, `TestTheDomainIsUnchanged`** or the module-level cell.

### Mutation table

11 mutations, control clean, 6 distinct firing sets.

| mutation | fires | notes |
|---|---|---|
| control (nothing mutated) | 0 | `30 passed` |
| fix reverted (word `port` hard-coded again) | 6 | the pre-fix split |
| `param` named but the port value dropped | 3 | exactly the byte-identical-text controls |
| `context` interpolated where `param` belongs | 8 | |
| transport ceiling widened to 65535 | 9 | over-reach: the domain moved |
| guard never refuses | 9 | same set as above, same observable |
| **a NEW helper copies the promise and breaks it** | **1** | names `planted_error` |
| **same plant, scan hard-coded to today's names** | **0** | silent - why the set is derived |
| declaration-only filter removed | 0 | no shipped stub carries the sentence, so the filter is a forward guard; it is graded directly by `test_a_declaration_only_stub_is_not_held_to_the_promise` |

The last two rows are the point of the derived scan: a hard-coded list equal to today's members fires nothing and still looks like a working test.

One of my own assertions was unsound and the suite caught it: `bridge_port 65535` *contains* the substring `port 65535`, so the "does not name a parameter the caller never passed" cell needs a word boundary (`\bport 65535`), not a substring test.

## Gate

- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ`: clean, 1580 files formatted.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine worktree of the merge-base, normalized message sets **byte-identical in both directions**, 0 attributable to either changed file.
- Paired pytest, identical 407-file selection (all of `tests/tools/`, every guard that walks the tree or reads source/docstrings, and everything naming rosbridge / `tcp_port_error` / the changelog convention), with the new file excluded from **both** trees:

| tree | result |
|---|---|
| merge-base `13fda66a` | `17 failed, 19647 passed, 107 skipped` in 690.54s |
| branch | `17 failed, 19647 passed, 107 skipped` in 665.61s |

17 identical failing ids, `comm` empty in both directions. All 17 are `tests/mesh/test_iot_*` needing `awsiot`, which `pyproject.toml` declares in the `[mesh-iot]` extra and which is absent on this machine (confirmed with `tomllib` + `find_spec`).

No visual artifact: this changes a refusal string, not a policy, render, recording or asset. The measured before/after message tables above are the evidence.

### DIFF
--- a/changelog.d/2764-rosbridge-refusal-names-its-parameter.md
+++ b/changelog.d/2764-rosbridge-refusal-names-its-parameter.md
@@ -1,0 +1,20 @@
+### Fixed: a rosbridge port refusal names the parameter it was given
+
+`use_rosbridge._transport_port_error` refuses a port that is legal TCP but outside the range the
+rosbridge WebSocket transport can address. It takes a `param` argument documented as "the parameter
+name it came from, used in the message" -- the same sentence eleven sibling helpers in this package
+carry -- and hard-coded the word `port` into the message instead, so the argument was accepted and
+never read. Eleven of the twelve helpers making that promise keep it; this one did not.
+
+That matters because the helper is shared by design. Its own docstring says it is applied by both
+the `use_rosbridge` tool and `RosbridgeRobot` so the two "cannot disagree about which ports it can
+carry", and the shared 16-bit domain that runs one line earlier on the same value
+(`utils.tcp_port_error`) does interpolate the name. A caller whose parameter is spelled anything
+other than `port` would therefore be told the wide domain refused its `bridge_port` and the
+transport refused a `port` it never passed.
+
+The message now interpolates `param`. Both shipping call sites pass the literal `"port"`, so every
+refusal text a caller can currently reach is byte-identical; the change is what makes the
+documented contract true for the third caller the docstring anticipates. A derived test reads the
+promise sentence out of the package and holds every helper carrying it to interpolating its
+`param`, so a twelfth helper that copies the sentence is graded when it lands.
--- a/strands_robots/tools/use_rosbridge.py
+++ b/strands_robots/tools/use_rosbridge.py
@@ -107,7 +107,7 @@ def _transport_port_error(port: int, param: str, context: str) -> str | None:
     """
     if port > _TRANSPORT_MAX_PORT:
         return (
-            f"{context}: port {port!r} is a legal TCP port that the rosbridge WebSocket "
+            f"{context}: {param} {port!r} is a legal TCP port that the rosbridge WebSocket "
             f"transport cannot address (it addresses 1-{_TRANSPORT_MAX_PORT}; autobahn's "
             "URL builder excludes the top of the range)"
         )
--- a/tests/tools/test_transport_refusal_names_its_parameter.py
+++ b/tests/tools/test_transport_refusal_names_its_parameter.py
@@ -1,0 +1,228 @@
+"""A refusal that takes a parameter name spends it on the message.
+
+Eleven helpers in this package document a ``param`` argument with one
+sentence - "The parameter name it came from, used in the message" - and
+interpolate it, so the caller is told which of their own parameters was
+refused. ``use_rosbridge._transport_port_error`` made the same promise and
+hard-coded the word ``port`` instead, which is only correct while every
+caller happens to name its parameter ``port``.
+
+That is not a hypothetical constraint on a private helper. Its own docstring
+says it is *shared* with
+:class:`~strands_robots.mesh.rosbridge_robot.RosbridgeRobot` precisely so the
+tool and the bridge "cannot disagree about which ports it can carry" - so a
+third caller is anticipated by design, and a caller whose parameter is spelled
+anything else would be handed a refusal naming a parameter it does not have,
+one line after the shared 16-bit domain named the parameter correctly.
+
+The population is derived from the promise rather than listed, so a twelfth
+helper that copies the sentence is held to it the moment it lands.
+"""
+
+from __future__ import annotations
+
+import ast
+import inspect
+import pathlib
+import re
+from typing import Any
+
+import pytest
+
+import strands_robots.tools.use_rosbridge as ur
+from strands_robots.utils import tcp_port_error
+
+# The sentence that constitutes the promise. Any helper whose docstring carries
+# it has told its callers the message names their parameter.
+PROMISE = "The parameter name it came from, used in the message"
+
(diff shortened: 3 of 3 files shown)


yinsong1986 (dismissed):
## Summary

One-word fix in `use_rosbridge._transport_port_error`: the refusal for a port above the transport's 65534 ceiling now interpolates the documented `param` argument instead of the hard-coded word `port`, so the helper keeps the same promise as the other guards carrying the "parameter name it came from, used in the message" docstring sentence. Both shipping call sites pass the literal `"port"`, so production refusal text is byte-identical before and after (the new test pins that exact string plus the call-site spellings). The accompanying regression test also derives the whole promise-keeping family from the docstring sentence via an AST scan, with non-vacuity and constructed-exemplar checks so the rule cannot pass by absence.

## What's good

- Verified locally at head `724b284f`: all 30 tests in the new file pass, and the derived scan correctly flags the pre-fix helper as a violator (the test fails on pre-fix code, per AGENTS.md > "Pin regression tests for reviewed fixes").
- Test file is ASCII-clean, host-path-free, ruff-clean, and every test name describes behaviour rather than provenance.
- The transport's domain boundary is untouched (accept/refuse verdicts pinned unchanged), and the `tcp_port_error` shared-owner relationship is asserted rather than restated.

yinsong1986 (approved):
## Summary

One-word production fix in `use_rosbridge._transport_port_error`: the refusal for a port above the transport's 65534 ceiling now interpolates the documented `param` argument instead of a hard-coded `port`, making the helper keep the same promise as the eleven sibling guards carrying the "parameter name it came from, used in the message" docstring sentence. Both shipping call sites (`use_rosbridge.py:352`, `rosbridge_robot.py:106`) pass the literal `"port"`, so every reachable refusal text is byte-identical before and after - verified against the call sites and the exact-string test. The accompanying regression test derives its population from the promise sentence rather than a list, grades its own predicate against constructed keep/break exemplars for non-vacuity, and pins the domain boundary (65534 accepted, 65535 refused, `tcp_port_error` untouched). All 30 tests pass on the head commit, the new files are ruff-clean and ASCII-clean, and the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention.

## What's good

- Minimal production diff (one interpolation) with the behavioural boundary explicitly pinned so the fix cannot over-reach into the domain.
- The derived scan plus `{True, False}` non-vacuity check follows the repo's pattern for guards whose failing branch is unexercisable after the fix.
- Changelog fragment and word-boundary regex (`\bport`) both show attention to the conventions in AGENTS.md.

PR #2766, opened 2026-08-26
feat(g1): wire send_action to publish LowCmd_ on rt/lowcmd (issue #361)

cagataycali:
## Summary

The P0 unblock for cagataycali/robots-harness#361, stacked on the just-merged #2757 (`DDSPublisher`). `send_action` now publishes real `LowCmd_` messages on `rt/lowcmd` — the arm SDK primitive the future control loop consumes at 500Hz.

**Decoupled from #358 as agreed:** this ships the driver's transport primitive; the 48 agent-facing `@tool` verbs (`g1_arm_release`, etc.) still ride on the `unitree_sdk2py` vendoring decision.

## What changed

**`strands_robots/drivers/g1.py`** (~195 insertions):
- Added `DDSPublisher` import + `_publisher` field constructed in `connect_eagerly` alongside subscribers
- Added `_TOPIC_LOWCMD = "rt/lowcmd"` and lazy-imported `_LOWCMD_CLS_PATH` constants
- `send_action` now: runs gates → checks publisher → resolves `LowCmd_` (cached) → builds message → publishes → returns success envelope
- New helpers: `_resolve_lowcmd_class` (cached lazy import), `_build_lowcmd` (validates every array shape before writing anything, so partial writes are impossible)
- Publisher closed in `cleanup()` alongside subscribers

## Safety discipline (unchanged, still first)
1. `_check_motion_gates("arm")` — FSM must be in `HANDSHAKE_FSMS` (500/501/801)
2. Battery-under-floor refusal
3. Publisher presence (defense-in-depth)
4. `LowCmd_` IDL resolution
5. Shape validation of `action` (length mismatch, out-of-range motor index refuse before any wire touch)
6. Publish

## Invariants preserved
- No `unitree_sdk2py` at module load (explicitly asserted by `test_g1_driver_module_does_not_import_the_sdk`)
- Shared `_DDS_INIT_LOCK` via `DDSPublisher` (from #2757)
- Idempotent `start()` / `cleanup()`
- `LowCmd_` class cached — a 500Hz loop pays the import cost once

## Tests
**12 new tests** in `tests/drivers/test_g1_send_action_wired.py`:
- SDK not imported at module load
- Refuses on disconnected / bad FSM / low battery (each before wire touch)
- Refuses on bad shape (bad joint type, length mismatch, out-of-range index)
- **Wire capture**: exact `(topic="rt/lowcmd", class=LowCmd_, message)` reaches `publisher.publish`; per-motor `q`/`kp`/`kd` land on `motor_cmd[joint]` exactly
- Partial fields leave others at IDL default
- Empty `action` is a valid stop
- Publisher errors surfaced
- `LowCmd_` class cached across calls (identity, not just equality)

**2 existing tests updated** — `test_send_action_admits_every_handshake_fsm` and `test_walk_fsms_has_a_consumer_and_a_documented_boundary` now pin the defense-in-depth path instead of the `"issue #358"` refusal.

**108 passed** in `tests/drivers/` + `tests/tools/g1/` (was 96 before).

## Line count
+195 in driver, +303 in new test file — well under the 400 loc cap. `ruff check` + `ruff format` clean.

## Next in stack
`feat/g1-policy-loop-wired` — `start_task`/`run_policy` running this primitive at 500Hz with per-step FSM re-gate and zero-torque on stop.

Refs: cagataycali/robots-harness#361

### Round 1 — the lint gate failed on this suite's types

The required `call-test-lint` check failed on the opening commit: `mypy` reported
26 errors, all of them in `tests/drivers/test_g1_send_action_wired.py`. `ruff
check`, `ruff format --check` and the 108 tests were already clean, so nothing in
the behavioural suite could see this — the holes were entirely in the test
doubles' types, and a double that is not typed against the object it replaces is
exactly the drift this file's own docstring says it exists to prevent.

Three distinct holes, fixed in `a0240ef`:

- **`_RecordingPublisher` was a bare duck-type.** Assigning it to
  `G1Driver._publisher`, declared `DDSPublisher | None`, is an incompatible
  assignment. It now subclasses `DDSPublisher`. This buys more than a green
  check: a future signature change to `start`/`publish`/`close` now fails this
  suite at type-check time instead of letting the fake silently drift from the
  object the driver actually calls. `super().__init__("lo")` is safe — the real
  `__init__` only records the interface and builds a lock, and no DDS work
  happens until `start()`, which this class overrides away.
- **`calls`/`next_error` were read straight off `driver._publisher`** at 10
  sites, which is not type-safe on an optional base-class attribute
  (`union-attr`). A single `_recorder()` helper narrows it once, so every
  assertion is cast-free and the fixture's contract — the driver holds the fake,
  not a real publisher — becomes an assertion rather than an assumption.
- **`_FakeMotorCmd` pre-declared no fields**, so reading `.q`/`.kp`/`.kd` off a
  narrowed `_FakeLowCmd` was `attr-defined`. The fields are now bare
  annotations, which is the fix that preserves the semantics the assertions
  depend on: an annotation without an assignment is known to the type checker
  but absent at runtime, so `not hasattr(entry, "kp")` still measures "the
  driver never wrote this field" rather than silently passing against a
  class attribute.

No production code and no test behaviour changed — the diff is types and
docstrings only. Verified locally before pushing: `ruff check` clean, `ruff
format --check` reports already-formatted, `mypy tests/drivers/` reports no
issues in 4 source files, and **108 passed** across `tests/drivers/` and
`tests/tools/g1/` — the same count this PR claimed before the fix, so the
typing change moved no behaviour.

### DIFF
--- a/strands_robots/drivers/g1.py
+++ b/strands_robots/drivers/g1.py
@@ -37,7 +37,7 @@ from collections.abc import AsyncGenerator
 from typing import TYPE_CHECKING, Any, cast
 
 from strands_robots.tools.g1 import HANDSHAKE_FSMS, WALK_FSMS, decode_code
-from strands_robots.tools.g1._dds_engine import DDSSubscriberSet
+from strands_robots.tools.g1._dds_engine import DDSPublisher, DDSSubscriberSet
 
 if TYPE_CHECKING:
     from strands.types.tools import ToolSpec, ToolUse
@@ -58,6 +58,20 @@ _TOPIC_LOWSTATE = "rt/lowstate"
 _TOPIC_BMS = "rt/lf/bmsstate"
 _TOPIC_LIDAR_STATE = "rt/utlidar/lidar_state"
 _TOPIC_LIDAR_CLOUD = "rt/utlidar/cloud_livox_mid360"
+_TOPIC_LOWCMD = "rt/lowcmd"
+
+# IDL class path for the write path. Resolved lazily on first ``send_action``
+# call so ``import g1`` stays safe without ``unitree_sdk2py``. Shape kept
+# identical to :meth:`G1Driver._subscription_plan` for the read path so a
+# future reader sees the same convention on both halves of the engine.
+_LOWCMD_CLS_PATH = ("unitree_sdk2py.idl.unitree_hg.msg.dds_", "LowCmd_")
+
+# Number of ``motor_cmd`` entries a G1 ``LowCmd_`` carries. The IDL fixes the
+# array length; we do not resize it, we only fill the leading ``N`` entries
+# a caller named in ``action``. The exact upper bound is a firmware detail
+# and the tests do not depend on it - the driver only writes what the caller
+# asked to write, and leaves the rest at whatever the IDL default is.
+_G1_LOWCMD_MOTOR_FIELDS: tuple[str, ...] = ("q", "dq", "tau", "kp", "kd")
 
 
 class G1Driver:
@@ -128,6 +142,11 @@ class G1Driver:
         # Populated by :meth:`connect_eagerly`. ``None`` on a machine that
         # never connected - a valid state for tests and imports.
         self._subs: DDSSubscriberSet | None = None
+        self._publisher: DDSPublisher | None = None
+        # Resolved LowCmd_ IDL class, cached on first send_action so the loop
+        # does not re-import for every step. ``None`` until the first write
+        # or a connect-time resolution failure.
+        self._lowcmd_class: Any | None = None
         self._connected: bool = False
         self._connect_error: str | None = None
 
@@ -274,6 +293,18 @@ class G1Driver:
                 self._connect_error = err
                 return err
         self._subs = subs
+        # Build the write half on the same call. It shares the shared init
+        # lock and lazy SDK import - :func:`ensure_dds` inside ``start`` is
+        # idempotent, so the subscriber set's earlier ``start`` is why this
+        # ``start`` returns ``None`` without touching the SDK a second time.
+        # An error here fails the connect the same way a subscriber failure
+        # does, so a caller does not end up with reads-only.
+        publisher = DDSPublisher(self._network_interface)
+        err = publisher.start()
+        if err is not None:
+            self._connect_error = err
+            return err
+        self._publisher = publisher
         self._connected = True
         self._connect_error = None
         return None
@@ -314,10 +345,13 @@ class G1Driver:
         logger.debug("%s.stop() no-op (no motion path wired yet)", self._tool_name)
 
     def cleanup(self) -> None:
-        """Release every DDS subscriber. Idempotent."""
+        """Release every DDS subscriber and publisher. Idempotent."""
         if self._subs is not None:
             self._subs.close()
             self._subs = None
+        if self._publisher is not None:
+            self._publisher.close()
+            self._publisher = None
         self._connected = False
 
     # ------------------------------------------------------------------ #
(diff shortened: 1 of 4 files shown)


cagataycali:
Closing as duplicate of #2767 (same wire path, same commit set). #2767 has auto-merge enabled and PR-A (#2757, DDSPublisher) already merged upstream.

strands_robots/drivers/g1.py
```diff
@@ -367,24 +401,146 @@ def send_action(
...
+        if self._publisher is None:
+            return _refuse("driver not connected - call connect_eagerly() first")
+        lowcmd_cls = self._resolve_lowcmd_class()
+        if isinstance(lowcmd_cls, str):
+            return _refuse(lowcmd_cls)
+        message, build_err = self._build_lowcmd(action, lowcmd_cls)
+        if build_err is not None:
+            return _refuse(build_err)
+        assert message is not None  # narrowing: build_err is None ⇒ message set
+        err = self._publisher.publish(_TOPIC_LOWCMD, lowcmd_cls, message)
```
yinsong1986, line 453: [MUST FIX] The `LowCmd_` goes out with its `crc` field left at the IDL default, and `mode_pr` / `mode_machine` are never set. Every `unitree_sdk2py` write path for the `unitree_hg` `LowCmd_` (the `rt/lowcmd` low-level example and the arm-SDK example alike) computes `cmd.crc = CRC().Crc(cmd)` immediately before `Write`, because the firmware validates the CRC and discards a command that fails it. As shipped, every publish is silently ignored by the robot while `send_action` returns `{"status": "success", ... "wrote LowCmd_ on rt/lowcmd for N joint(s)"}` - the "silently dropping actions" class, and exactly the "stub that would look like a successful write" the deleted test docstring warned about (AGENTS.md: no silent defaults on error).

Related and worth settling in the same fix: `_g1_common.py` documents `HANDSHAKE_FSMS` (500/501/801) as "FSM ids where arm-SDK commands are honoured ... before writing `rt/armsdk`" - i.e. the regimes where the built-in controller owns the motors. Raw `rt/lowcmd` is consumed when that service is *disabled*, and an arm write on the G1 rides the arm-SDK topic (with the weight slot) precisely so it doesn't fight the running controller. So the gate that admits this write names the one regime in which a raw `rt/lowcmd` publish is either dropped or contends with the internal controller for the same motors.

Resolve by setting `mode_pr`/`mode_machine` and computing the CRC before publish (lazy-import the SDK's `CRC` helper alongside the IDL, same discipline as `_resolve_lowcmd_class`), and either targeting the arm-SDK topic for the `"arm"` scope or re-gating `rt/lowcmd` on the low-level regime. Why this can't be deferred: the next PR's 500Hz control loop consumes this primitive, and a success envelope that doesn't correspond to actuation is unfalsifiable from the caller's side - the loop will run green against a robot that ignores it, and the `action` schema/topic become the contract every consumer codes against once the stack lands.

strands_robots/drivers/g1.py
```diff
@@ -367,24 +401,146 @@ def send_action(
...
+            try:
+                values = [float(v) for v in raw]
+            except (TypeError, ValueError) as exc:
+                return None, f"'{field}' must be a list of floats: {exc}"
+            if len(values) != len(joints):
+                return None, (f"'{field}' has {len(values)} values but 'joints' has {len(joints)} entries")
+            arrays[field] = values
+        for idx, joint in enumerate(joints):
+            try:
+                entry = motor_cmd[joint]
```
yinsong1986, line 535: [MUST FIX] `entry = motor_cmd[joint]` does not refuse a negative index: `motor_cmd` is a Python sequence, so `joints: [-1]` wraps to the *last* motor slot (motor 34 on the real IDL) and returns success. That contradicts this method's own docstring ("An unknown motor index is surfaced as an error too - silently dropping a write would be worse than refusing it") - and a silently *mis-addressed* write is worse than a dropped one: on hardware it is `kp=50` applied to a joint the caller never named, on the hot path the 500Hz loop will drive. `test_send_action_refuses_an_out_of_range_joint` only pins the past-the-end direction, so nothing catches this today.

Fix is one line in the validation pass that already runs before any write: refuse `j < 0` where `joints` is coerced (`if any(j < 0 for j in joints): return None, ...`). Must-fix bucket: data corruption - an action delivered to the wrong physical motor with a success envelope, on the wire primitive this PR exists to ship.

yinsong1986 (commented):
## Summary

This PR turns `G1Driver.send_action` from the "not wired yet (issue #358)" stub into a real write path: `connect_eagerly` now builds a `DDSPublisher` beside the subscriber set (failing the connect symmetrically if the write half cannot start), `send_action` runs the pre-existing FSM/battery gates first, lazily resolves and caches the `LowCmd_` IDL class so a 500Hz loop pays the import once, validates every array shape in `action` before touching the message, and publishes on `rt/lowcmd`. Two `[MUST FIX]` inline comments below, both about the correctness of the message that actually goes out on the wire.

## What's good

- Gate ordering (FSM -> battery -> publisher presence -> IDL -> shape -> wire) is disciplined and every refusal is pinned by a test that also asserts no publish happened.
- The no-`unitree_sdk2py`-at-module-load invariant is preserved and explicitly asserted by `test_g1_driver_module_does_not_import_the_sdk`.
- `_RecordingPublisher` subclasses the real `DDSPublisher` rather than duck-typing it, so a signature change to `start`/`publish`/`close` fails this suite at type-check time.
- Changelog fragment shipped per the `changelog.d/` rule; existing tests updated honestly (the defense-in-depth docstrings say exactly why the publisher-is-None state is being pinned).

## Verification suggestions

Before the 500Hz loop PR stacks on this, diff the message this driver emits against what `unitree_sdk2py`'s own G1 examples (`g1_low_level_example.py`, the arm-SDK example) set immediately before `Write` on the same `unitree_hg` `LowCmd_` - in particular `crc`, `mode_pr` and `mode_machine` - and confirm on hardware (or against the firmware docs) which topic honours a write in FSM 500/501/801.

PR #2768, opened 2026-08-26
fix(drivers): close the G1 DDS endpoints instead of only forgetting them

cagataycali:
## Problem

`DDSSubscriberSet.close` and `DDSPublisher.close` each dropped their references
and relied on garbage collection, and each justified that in its docstring:

> `ChannelSubscriber` in `unitree_sdk2py` has no explicit close - it relies on
> garbage collection - so this drops the references and lets the collector claim
> them.

Both halves are wrong. `ChannelSubscriber.Close()` and `ChannelPublisher.Close()`
both exist (they route to `Channel.CloseReader` / `Channel.CloseWriter`), and for
a subscriber the garbage-collection half is not merely optimistic - it cannot
happen. `subscribe()` builds every subscriber with `queueLen=10`, and at any
queue length above zero the SDK starts a `ch_reader` daemon thread whose target
is a bound method of the channel's reader. That running thread keeps the channel
reachable, so the CycloneDDS `__del__` that would release the `DataReader` never
runs.

Measured against `unitree_sdk2py` 1.0.1, counting `ch_reader` threads:

```
baseline                        0
after subscribe                 1     queueLen=10 starts the reader thread
ChannelSubscriber.Close exists  True
after del + 3x gc.collect()     1     <- exactly what close() did
after an explicit Close()       0
```

The mechanism isolates cleanly - the thread is the only variable:

| endpoint | `Close()` exists | drop refs + gc releases it |
| --- | --- | --- |
| `ChannelPublisher` (writer, no thread) | yes | yes |
| `ChannelSubscriber`, `queueLen=10` (what `subscribe()` uses) | yes | **no** |
| `ChannelSubscriber`, `queueLen=0` (no thread) | yes | yes |

## Why it matters

`G1Driver.cleanup()` is the only release path, and it released nothing. Four
subscriptions therefore outlived it: the decoder callbacks kept writing `_imu`,
`_fsm_id`, `_battery` and the LiDAR caches for a driver reporting itself
disconnected, and each leaked a live daemon thread.

The second consequence is worse. `cleanup()` sets `_connected = False`, so the
documented re-attach route - `cleanup()` then `connect_eagerly()` - built a fresh
`DDSSubscriberSet` and subscribed all four topics again while the previous four
were still attached. That is the duplicate subscription the shared
`_DDS_INIT_LOCK` exists to prevent, and it is the hazard `connect_eagerly()`
already names when it declines to rebuild a live set:

> Rebuilding the subscriber set instead would re-subscribe all four topics and
> drop the only reference to the previous one, leaking its subscribers on a bus
> whose bindings segfault under concurrent construction.

One route into that leak was guarded. The other performed it unconditionally.

## Change

Both `close()` methods now call `Close()` on each endpoint:

- Each endpoint is closed under its **own** `try`, so one SDK failure cannot
  strand the endpoints behind it - a teardown that gives up halfway leaks
  precisely what it was called to release. The failure is logged with the reason
  the SDK gave.
- The collection is swapped out under `_lock` **before** any `Close()` runs, so a
  second call closes nothing twice, and a concurrent `subscribe()` - which
  appends without that lock - cannot mutate the list being walked.
- The publisher gets the same treatment. Its writer starts no reader thread, so a
  dropped reference did reach `__del__`; the change makes the release happen at
  the call rather than whenever the last reference happens to go. The write path
  added in #2757 carried the same "has no explicit close" sentence, so both
  copies are corrected here rather than leaving the two siblings on different
  teardown conventions.
- `DDSPublisher.close`'s docstring claimed it is "called from
  `G1Driver.cleanup`". The driver holds no `DDSPublisher` yet, so that reference
  is corrected to say the control loop in issue #361 will own one.

No behaviour outside teardown changes: `start`, `subscribe`, `get_publisher` and
`publish` are untouched.

## Tests

`tests/drivers/test_g1_dds_engine_release.py`, 19 cells. Pre-fix, **10 fail and
9 pass**:

| class | failed/total | role |
| --- | --- | --- |
| `TestClosingASubscriberSetReleasesEverySubscriber` | 6/6 | regression |
| `TestClosingAPublisherReleasesEveryPublisher` | 4/4 | regression |
| `TestCloseStillForgetsWhatItClosed` | 0/4 | the old contract must survive |
| `TestSubscribeAsksForAQueuedReader` | 0/1 | the queue length is the mechanism |
| `TestTheSdkHasAnExplicitClose` | 0/4 | premise, against the installed SDK |

The behavioural cells drive recording stand-ins through the real `subscribe()`
and `get_publisher()`, so they grade the objects the production path builds. The
one fact a stand-in cannot establish - that the SDK classes *have* `Close` - is
pinned separately and skips where `unitree_sdk2py` is absent, which is every
headless runner. A stand-in that defines `Close` would otherwise agree happily
with a fix built on the same wrong belief.

### Mutation coverage

13 mutations, 12 detected, 12 distinct firing sets, control clean, source
restored byte-identical after every row:

| mutation | cells fired | pre-existing cells fired |
| --- | --- | --- |
| control (no mutation) | 0 | 0 |
| subscriber `Close()` dropped (the shipped defect) | 6 | 0 |
| publisher `Close()` dropped (the shipped defect) | 4 | 0 |
| both dropped (equivalent to a full revert) | 10 | 0 |
| subscriber: only the first is closed | 3 | 0 |
| subscriber: one `try` around the whole loop | 2 | 0 |
| subscriber: closed but never forgotten | 2 | 0 |
| subscriber: failure swallowed silently | 2 | 0 |
| subscriber: no guard, a failing `Close` propagates | 3 | 0 |
| publisher: one `try` around the whole loop | 2 | 0 |
| publisher: closed but never forgotten | 2 | **1** |
| `subscribe()` asks for an unqueued reader (`queueLen=0`) | 1 | 0 |
| tolerant `getattr(sub, "Close", None)` instead of the contract | **0** | 0 |

Two rows are worth reading:

- The two single-class rows are **disjoint and sum exactly** to the both-dropped
  row (6 + 4 = 10), so neither half rides on the other's tests.
- "publisher: closed but never forgotten" also trips
  `tests/tools/g1/test_dds_publisher.py::test_close_is_idempotent_and_drops_cache`,
  the write path's own guard - the new cells compose with it rather than
  duplicating it.

The one undetected row is honest and deliberate: replacing `sub.Close()` with a
tolerant `getattr(sub, "Close", None)` is behaviourally identical for anything
that defines `Close`, so no test can distinguish it. It is not adopted because
the SDK does define `Close` on both classes; a stand-in that lacks it is a
stand-in worth surfacing, not tolerating.

## Verification

  tests_integ` (1580 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a worktree pinned
  to this branch's merge-base, normalized message sets byte-identical in both
  directions, 0 in the changed files.
- Paired suite: an identical 332-file selection (all of `tests/drivers` and
  `tests/tools/g1`, plus every guard that walks the tree, reads source or
  docstrings, or reads `changelog.d`), with the new file excluded from both -
  identical **16782 collected** and `17 failed, 16705 passed, 60 skipped` on both
  trees. Failing-ID sets identical, both `comm` directions empty. All 17 are
  `ModuleNotFoundError: No module named 'awsiot'` (the `[mesh-iot]` extra, not
  installed here).
- New file: 19 passed, with the four SDK-gated premise cells running rather than
  skipping on a machine that has `unitree_sdk2py`.

The SDK used as the oracle is `unitree_sdk2py` 1.0.1; its `core/channel.py` is
byte-identical (md5 `a46a6312053b2c5b6943a3a4486335e9`) to the one in the
published `unitree-sdk2` 1.0.1 wheel, so the measurements above describe what a
declared dependency would install.

No visual artifact: this changes teardown and resource release, not a policy,
simulation, rendering, recording or asset. The measured thread-count table is the
evidence.

### DIFF
--- a/changelog.d/2768-g1-dds-close-releases-endpoints.md
+++ b/changelog.d/2768-g1-dds-close-releases-endpoints.md
@@ -1,0 +1,35 @@
+### Fixed: closing the G1 DDS engine releases its endpoints
+
+`DDSSubscriberSet.close` and `DDSPublisher.close` dropped their references and
+relied on garbage collection, each justifying that with a docstring asserting
+that the `unitree_sdk2py` class "has no explicit close". Both classes do define
+`Close()`, and for a subscriber the collection half is not merely optimistic but
+impossible: `subscribe()` builds every subscriber with `queueLen=10`, and at any
+queue length above zero the SDK starts a `ch_reader` daemon thread whose target
+is a bound method of the channel's reader. The running thread keeps the channel
+reachable, so the CycloneDDS `__del__` that would release the `DataReader` never
+runs.
+
+Measured against `unitree_sdk2py` 1.0.1, one `ch_reader` thread per
+subscription:
+
+```
+after subscribe:              1
+after del + 3x gc.collect():  1     <- what close() used to do
+after an explicit Close():    0
+```
+
+So `G1Driver.cleanup()` released nothing. The decoder callbacks kept filling
+caches for a driver reporting itself disconnected, and because `cleanup()` then
+sets `_connected = False`, the documented re-attach route - `cleanup()` followed
+by `connect_eagerly()` - subscribed all four topics a second time. That is the
+duplicate subscription the engine's shared construction lock exists to prevent,
+and it is the hazard `connect_eagerly()`'s own docstring already names when it
+declines to rebuild a live subscriber set.
+
+Both `close()` methods now call `Close()` on each endpoint, each under its own
+`try` so one SDK failure cannot strand the endpoints behind it, and swap the
+collection out under the lock first so a second call closes nothing twice. A
+writer starts no reader thread, so a dropped publisher did reach `__del__`; the
+change makes the release happen at the call rather than whenever the last
+reference happens to go.
--- a/strands_robots/tools/g1/_dds_engine.py
+++ b/strands_robots/tools/g1/_dds_engine.py
@@ -109,15 +109,35 @@ class DDSSubscriberSet:
             return None
 
     def close(self) -> None:
-        """Release every subscriber. Idempotent.
+        """Close every subscriber and forget it. Idempotent, and never raises.
 
-        ``ChannelSubscriber`` in ``unitree_sdk2py`` has no explicit close - it
-        relies on garbage collection - so this drops the references and lets
-        the collector claim them. Called from
-        :meth:`~strands_robots.drivers.g1.G1Driver.cleanup`.
+        ``ChannelSubscriber.Close()`` is what releases a subscription, and
+        dropping the reference is not a substitute. :meth:`subscribe` builds
+        each subscriber with ``queueLen=10``, and at any queue length above
+        zero ``unitree_sdk2py`` starts a ``ch_reader`` daemon thread whose
+        target is a bound method of the channel's reader. That running thread
+        keeps the whole channel reachable, so the CycloneDDS ``__del__`` that
+        would release the ``DataReader`` never runs: the thread keeps looping,
+        the reader stays matched, and the decoder callback keeps filling caches
+        for a driver that believes it is disconnected. ``Close()`` sets the
+        thread's stop event, joins it and drops the reader; nothing else does.
+
+        The list is swapped out under :attr:`_lock` before any ``Close()``
+        runs, so a second call finds nothing to close and a concurrent
+        :meth:`subscribe` - which appends without that lock - cannot mutate
+        the list being walked. Each subscriber is then closed under its own
+        ``try``, because a teardown that abandons the subscribers behind a
+        failing one leaks exactly what it was called to release.
+
+        Called from :meth:`~strands_robots.drivers.g1.G1Driver.cleanup`.
         """
         with self._lock:
-            self._subs.clear()
+            subs, self._subs = self._subs, []
+        for sub in subs:
+            try:
+                sub.Close()
+            except Exception as exc:  # noqa: BLE001 - SDK raises bare Exception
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Replaces the drop-the-reference teardown in `DDSSubscriberSet.close` and `DDSPublisher.close` with explicit `Close()` calls on each endpoint, correcting the docstrings' false premise that `unitree_sdk2py` has no explicit close. The subscriber case is the real leak: `subscribe()` uses `queueLen=10`, so the SDK's `ch_reader` daemon thread keeps the channel reachable and the CycloneDDS `__del__` never runs - meaning `G1Driver.cleanup()` released nothing and a subsequent `connect_eagerly()` double-subscribed. Both `close()` methods now swap the collection out under the lock (idempotent, safe against a concurrent walk), then close each endpoint under its own `try` with a logged warning so one SDK failure cannot strand the rest - matching the file's existing never-raise, error-string convention. The `except Exception` clauses carry the same noqa'd justification the module's other SDK-touching paths already use.

## What's good

- The regression tests drive the endpoints `subscribe()`/`get_publisher()` actually built rather than a hand-fed list, and separately pin the SDK-side premises (`Close` exists, no `__del__` on `ChannelSubscriber`) behind `importorskip` so a stand-in cannot vouch for facts it cannot know.
- `sys.modules` stubbing goes through `monkeypatch.setitem`, per AGENTS.md > Testing Patterns, so a machine with the real SDK gets it back.
- Boundary cells pin that the old contract (forgetting, idempotence, no-op on an empty engine) survives the fix.
- Changelog fragment present and named for the PR; log strings are ASCII; no host paths.

PR #2769, opened 2026-08-26
fix(scripts): collide duplicate work on the identity a created path declares

cagataycali:
## The gap

`scripts/check_duplicate_claim.py --all-open` pairs two open pull requests on a path they both
**create**. It shipped in #2710 to reach the duplicate pairs the claim key cannot see, since most
branches here link no issue.

A changelog fragment is named `<number>-<slug>.md`. So two branches describing one change write one
slug under two numbers, their paths differ, and a pair whose only shared creation is that entry
collides on nothing.

#2766/#2767 are that pair. Both wire `G1Driver.send_action` to publish `LowCmd_` on `rt/lowcmd`,
opened 8m 06s apart, one putting its tests in a new file and the other into an existing one, so they
create no common path at all:

| | #2766 | #2767 |
|---|---|---|
| fragment | `changelog.d/2765-g1-send-action-wired.md` | `changelog.d/2761-g1-send-action-wired.md` |
| tests | `tests/drivers/test_g1_send_action_wired.py` (new) | `tests/drivers/test_g1_driver.py` (existing) |

Run against the open set while both sat in it, `--all-open` on `main` answered:

```
Outcome: **unique-additions**
No two of the 11 open pull requests create the same file (55 pair(s) compared).
```

## Why this was not an oversight

It was an assertion. `test_a_changelog_fragment_is_never_the_shared_path` pinned that a fragment can
never be the shared path, and its docstring gave the reason:

> Every branch adds one, and its name embeds the number, so it cannot collide.

That is true of the raw path, and it is exactly why the number is the half that has to go. The
fixtures carried a fragment specifically to demonstrate the exclusion, and both halves of
#2388/#2389 in that fixture already share a slug -- the counter-example was in the test data.

The fear behind the boundary was stated too: without the exclusion, "a relation over added paths
would fire on every pair in the queue". That is the claim this change had to answer, and it is
measurable.

## Measured

Over 353 pull requests (#2345 through #2767), on the 2002 pairs open at the same instant. Both
versions of `find_addition_collisions` were run over the identical reconstructed inputs, the base one
loaded from `git show upstream/main:` so it is the shipped code rather than a memory of it:

| relation | pairs selected | duplicates among them |
|---|---|---|
| both edit a path | 127 | a composition question, `check_merge_base_overlap.py` owns it |
| both add a path (`main`) | 2 | 2 |
| both create the same thing (here) | 3 | 3 |

Strict superset: base's 2 pairs are a subset of the branch's 3, **0 lost**. The added pair is
#2766/#2767.

| what both create | pair | closed |
|---|---|---|
| `tests/test_recorder_counters_track_on_disk_frames.py` | #2388, #2389 | #2389 |
| `tests/training/test_checkpoint_cadence_domain.py` | #2707, #2708 | #2707 |
| `changelog.d/*-g1-send-action-wired.md` | #2766, #2767 | #2766 |

All three are duplicates confirmed from their own closing comments, not inferred.

**Slug reuse is not noisy.** Of those 353 pull requests, 350 add a fragment and between them use
**350 distinct slugs**. Exactly two slugs are used twice, and both times the two users are a
duplicate pair. The number is the noisy half: it is meant to be the pull request's own number and in
**40 of those 350 it is not**, because it is chosen before the pull request exists and races with
whatever merges first.

## The change

The collision key becomes the identity a created path declares (`addition_key`). For every path but
a fragment that is the path itself, which is why the widening loses nothing.

The report names the shared subject as `changelog.d/*-<slug>.md` rather than a filename, because the
filename it would otherwise print exists on neither branch. For an identical path the subject is the
path, so those rows are unchanged.

What a fragment *is* comes from `scripts/assemble_changelog.py` rather than a second copy of the
pattern -- `check_changelog_fragment.py` already reuses that module for the same rule -- so this
sweep and the assembler cannot disagree. A name the assembler would reject keys on itself, which can
only fail to report a pair rather than invent one.

One guard was removed rather than kept unexplained: a `"/" in name` test is unconditionally implied
by the pattern, since `/` is outside every character class in it. The reserved-name check stays,
because it is redundant only against *today's* list -- a reserved name carrying a leading number
would match the pattern -- and that is now graded by monkeypatching the assembler's list.

## Verification

Pre-fix split, reported both ways because a raw revert also removes the new symbol:

| revert | result |
|---|---|
| raw (source + guidance reverted, tests kept) | **24 failed / 135 passed** -- 17 behavioural, 7 `AttributeError` on the new symbol |
| behavioural (`addition_key` kept, relation reverted) | **6 failed / 153 passed** |

The 6 are 5 in `TestTheMeasuredClaimFreePairsAreReported` and 1 in the new class. **0 failures** in
`TestTheTwoKeysAreComplementary`, `TestOnlyACreatedPathCollides`,
`TestEveryPairIsComparedAndTheReportIsStable`, `TestAnIncompleteAnswerIsNotAFinding`,
`TestTheOpenSetIsReadLiveAndWholeAndIncludesDrafts`, `TestTheReportNamesBothPullRequestsAndTheRemedy`,
`TestTheExitStatusIsOneOnlyForTheFinding`, or anywhere in `tests/test_duplicate_claim_check.py`.

Break table -- 11 mutations, 10 detected, control clean, 8 distinct firing sets, `collected` a stable
160 in every row, source restored byte-identical:

| row | fires | mutation |
|---|---|---|
| b0 | 0 | control |
| b1 | 6 | relation keyed on the raw path again (the shipped behaviour) |
| b2 | 6 | `addition_key` is the identity, relation untouched |
| b3 | 6 | keyed on the number instead of the slug |
| b4 | 2 | the assembler's reserved-name list not consulted |
| b5 | 7 | the glob marker dropped, so the key looks like a real path |
| b6 | 1 | the fragment-directory test dropped |
| b7 | 8 | over-collide: key on the directory alone |
| b8 | 1 | the naming rule restated locally instead of reused |
| b9 | 4 | the docstring's measurement paragraph removed |
| b10 | 2 | the guidance left un-updated |

b1/b2/b3 share a firing set, and the reason is structural rather than a thin table: all three are
"the slug is not the key", and no two fixtures share a fragment number, so keying on the number is
indistinguishable from not keying on a fragment at all.

b7 is the over-reach guard, and it is the row that matters most for scope: keying on the directory
alone fires 8, including both cells of `TestTheTwoKeysAreComplementary`. The widening must not
swallow the claim key's territory, and the two issue-keyed pairs (#2570/#2571, #2480/#2508) still do
not collide under it -- their slugs differ.

b5 fires `test_no_reported_subject_is_invented`, which is the truthfulness pin: every reported
subject has to be a real created path, or the glob of a path each side really added.

Gate, both trees, identical 126-file selection (every test reading `AGENTS.md`, `scripts/`,
`changelog.d/`, or walking the tree), the changed test file excluded from each:

```
branch  4731 collected   4676 passed, 55 skipped   291.94s
base    4731 collected   4676 passed, 55 skipped   309.74s
```

Zero failures either side. `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine
`upstream/main` worktree, normalized message sets byte-identical, **0 in the changed files**. `ruff
check` and `ruff format --check` clean. The changed test file alone: **88 passed** (160 with its
claim-keyed sibling). Changelog fragments validate, and `tests/test_changelog_*`: 89 passed.

Dogfooded: run from this branch while both halves were open, `--all-open` reported
`duplicate-addition`, exit `1`, naming #2766 and #2767. #2766 has since been closed as the duplicate,
and the sweep run again over the 10 open pull requests -- this one included -- reports
`unique-additions` across 45 pairs, with no false positive on itself. The relation is a property of
the open set, so that is the correct answer once the pair is resolved rather than a regression.

No visual artifact: this changes a reporting relation in a repository script, not a policy,
simulation, render, recording or asset. The measured tables above are the evidence.

## Scope

`AGENTS.md` step 1 carries the same measurement and moves in the same change; its numbers were pinned
by `TestTheGuidanceRecordsTheSecondKey`, which is why b10 fires. Neither claim-keyed mode is touched,
the outcome vocabulary is unchanged, and the composition question stays with
`scripts/check_merge_base_overlap.py`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -271,23 +271,30 @@ hatch run format            # ruff check --fix, ruff format
 
    **The claim is not the only key.** 249 of the last 300 pull requests (#2345
    through #2708) link no issue at all, so for most of the traffic the query above
-   has nothing to collide on - and two of the four duplicate pairs in that window
-   were claim-free: #2388/#2389 and #2707/#2708, each pair independently creating
-   one new test file. Once your branch is pushed and the pull request is open, ask
-   the second question:
+   has nothing to collide on - and three of the five duplicate pairs in the window
+   #2345 through #2767 were claim-free: #2388/#2389, #2707/#2708 and #2766/#2767.
+   Once your branch is pushed and the pull request is open, ask the second
+   question:
 
    ```
    python3 scripts/check_duplicate_claim.py --repo strands-labs/robots --all-open
    ```
 
-   It reports only pairs that **create the same path**, which over the 1802 pairs
-   that were open at the same instant selected 2 - and both were duplicates. Two
-   branches *editing* one file is a different question with a different remedy
+   It reports only pairs that **create the same thing**, which over the 2002 pairs
+   that were open at the same instant selected 3 - and all three were duplicates.
+   For almost every path that means creating the same path. The exception is a
+   changelog fragment, which is named `<number>-<slug>.md`: two branches describing
+   one change write one slug under two numbers, so they are paired on the slug and
+   the number is dropped. That is not a loose key - across the 350 pull requests in
+   that window which add a fragment there are **350 distinct slugs**, and the only
+   two used twice are the two duplicate pairs above that reused one.
+
+   Two branches *editing* one file is a different question with a different remedy
    (a merge order, possibly one composition run), and
    `scripts/check_merge_base_overlap.py --github-repo <owner/name> --all-open`
-   owns it; that relation selects 117 of the same 1802. The two keys are complementary rather than nested: neither
-   issue-keyed pair shares an added path, and neither claim-free pair claims an
-   issue.
+   owns it; that relation selects 127 of the same 2002. The two keys are
+   complementary rather than nested: no issue-keyed pair shares an added path, and
+   no claim-free pair claims an issue.
 
    This one cannot be asked before you start, and not for want of trying: a path
    set is a property of a pushed branch, so there is nothing to read at intake. It
--- a/changelog.d/2769-duplicate-work-created-identity-key.md
+++ b/changelog.d/2769-duplicate-work-created-identity-key.md
@@ -1,0 +1,41 @@
+### Fixed: duplicate work is collided on the identity a created path declares, so two changelog fragments naming one change are reported
+
+`scripts/check_duplicate_claim.py --all-open` pairs two open pull requests on a path they both
+**create**. A changelog fragment is named `<number>-<slug>.md`, so two branches describing one
+change write one slug under two numbers and their paths differ -- and a pair whose only shared
+creation is that entry collides on nothing. #2766/#2767 are exactly that pair: both wire
+`G1Driver.send_action` to publish on `rt/lowcmd`, opened 8m 06s apart, one putting its tests in a
+new file and the other into an existing one, so they create no common path at all. Run against the
+open set while both sat in it, the sweep reported `unique-additions - no two of the 9 open pull
+requests create the same file`.
+
+The exclusion was not an oversight but an assertion. A test pinned that a changelog fragment can
+never be the shared path, with the reason "its name embeds the number, so it cannot collide", and
+the fixtures carried a fragment to demonstrate it. The reason is true of the raw path and is exactly
+why the number is the half that has to go.
+
+The key is now the identity a created path declares. For every path but a fragment that is the path
+itself, so this is a strict widening: over the 2002 pairs in #2345 through #2767 that were open at
+the same instant it selects the same 2 pairs the raw path selects, plus #2766/#2767, and loses none.
+
+    what both create                                        pair           closed
+    tests/test_recorder_counters_track_on_disk_frames.py     #2388, #2389   #2389
+    tests/training/test_checkpoint_cadence_domain.py         #2707, #2708   #2707
+    changelog.d/*-g1-send-action-wired.md                    #2766, #2767   open
+
+The stated fear behind the old boundary was that keying on a fragment "would fire on every pair in
+the queue". It does not. Of those 353 pull requests, 350 add a fragment and between them use **350
+distinct slugs**; exactly two slugs are used twice, and both times the two users are a duplicate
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Widens `--all-open`'s collision key from the raw created path to the identity a created path declares (`addition_key`): a changelog fragment `<number>-<slug>.md` now collides on its slug with the number dropped, so a duplicate pair whose only shared creation is one changelog entry under two fragment numbers (#2766/#2767) is reported instead of passing as `unique-additions`. For every non-fragment path the key is the path itself, so the widening is strict. I verified the mechanics independently: the naming rule is reused from `scripts/assemble_changelog.py` via the same `_load_assembler` shape `check_changelog_fragment.py` already uses (no second copy of the pattern to drift), the `match.end('number') + 1` slice is sound against `FRAGMENT_NAME` (a hyphen always follows the number group), and every misfit -- reserved name, nested path, empty slug, uppercase extension -- keys on itself, which can only fail to report a pair rather than invent one. All three fixture rows check out byte-for-byte against the live `pulls/<n>/files` data (#2707 really added no fragment; #2766/#2767 really share only the slug), the full 160-test pair of duplicate-claim test files passes on the head, the changelog suite (89 tests) passes, ruff check/format are clean, and the changed files carry no non-ASCII. The outcome vocabulary and exit-code contract are unchanged, so nothing scaffolded on the report breaks.

## What's good

- The old exclusion was pinned by a test with a stated reason, and this change answers that reason head-on with a measurement (350 distinct slugs across 350 fragment-adding PRs) rather than deleting the boundary silently.
- Fixtures are real reconstructions from `pulls/<n>/files`, not invented shapes, and the three rows deliberately cover the three reachable collision shapes (both, path-only, slug-only).
- The conservative failure direction is designed in and tested: `test_no_reported_subject_is_invented` pins that a reported subject is never a file that exists on neither branch.
- AGENTS.md step 1 guidance and the module docstring move in the same change, and both are pinned by tests, per the documented-and-enforced-by-nothing lesson.

PR #2770, opened 2026-08-26
fix(transforms): keep the dtype each source camera declared

cagataycali:
## Problem

A LeRobot dataset declares each camera's storage per feature. One `observation.images.<cam>` can be `dtype="video"` (frames encoded into an MP4 under `videos/`) while another is `dtype="image"` (frames in the data parquet). `LeRobotDataset.create` accepts that declaration and reads both streams back, so it is a legal source - measured, and pinned as the premise class of the new test file.

The output side has one knob for the whole dataset: `DatasetRecorder.create` takes a single `use_videos` flag that decides every camera's dtype. `_SourceDataset.__init__` derived that flag by assigning it once per camera, inside the loop that reads the camera shapes:

```python
for cam in self.camera_keys:
    feat = features[f"{_IMAGE_PREFIX}{cam}"]
    ...
    self.use_videos = feat.get("dtype") == "video"   # last camera wins
```

So it held whichever camera the source's feature dict ended on. A two-camera source declaring one video stream and one image stream transformed with `status="success"` while the output stored a camera a different way, and the direction depended only on the order the features were declared in.

Measured end to end through `create_transform("mock").transform(...)` on real files, both orders:

| source cameras (declaration order) | output cameras before | MP4 files | after |
|---|---|---|---|
| `top='video'`, `wrist='image'` | `top='image'`, `wrist='image'` | 1 -> 0 | refused |
| `wrist='image'`, `top='video'` | `wrist='video'`, `top='video'` | 1 -> 2 | refused |

Both directions lose the source's schema. `create_output_recorder` documents itself as "the SOURCE schema (parity by construction)"; schema parity is the first acceptance criterion of `tests/transforms/test_round_trip.py`; and `docs/data/transforms.md` contract item 1 promises a generated episode is "the *same trajectory* rendered differently". A stream stored a different way is none of those.

## Change

`+40/-7` in `strands_robots/transforms/base.py`:

- every camera's dtype is recorded (`camera_dtypes`), not just the last one;
- `use_videos` is derived from the whole set (`all(dtype == "video" ...)`), which is exactly what the old code produced for a homogeneous source;
- `_SourceDataset.open` refuses a source whose cameras disagree, naming each camera and the dtype it declared so an operator knows which stream to convert.

Refusing is the disposition `open` already takes one branch up, for a source declaring a feature the pass-through cannot preserve, and for the same reason: the output recorder declares one dtype for every camera, so a mixed source cannot be reproduced, and writing it anyway alters a column the contract promises to carry through. The refusal sits after the dataset-stack probe and the no-camera refusal, and before any episode is read, so a refused source leaves nothing on disk (`episodes_read == 0`).

Re-encoding to one dtype instead is a quieter version of the same answer, and the break table shows it is worse: coercing everything to `image` also flattens the video streams of sources that work today.

## Visual

Read back through `LeRobotDataset` on both trees, from one source recorded once:

camera dtype parity

Both trees record byte-identical source pixels (asserted in the figure script). The output frames are the mock backend's brightness-shifted variant, so their pixels differ from the source by design - the subject is the dtype each camera is stored as. Every number in the caption is asserted from the two runs' own JSON before the figure is drawn, and each panel is verified to be embedded verbatim (`np.array_equal` at its paste offset).

## Tests

New `tests/transforms/test_camera_dtype_parity.py`, 30 cases in four classes. Pre-fix: **14 failed / 16 passed**, with **0 of 3** in the premise class (the mixed source is a legal LeRobot dataset) and **0 of 3** in the over-reach control class (a homogeneous source, either dtype, still transforms and keeps its schema). The 14 are 9 refusal cases and 5 flag-derivation cases.

Mutation table, run sequentially against the fix (12 mutations, 11 detected, 10 distinct firing sets, source restored byte-identical per row):

| # | mutation | cases fired |
|---|---|---|
| b0 | control: no mutation | 0 (30 passed) |
| b1 | the mixed-dtype refusal is removed | 9 |
| b2 | flag assigned once per camera again (recording + refusal kept) | 3 |
| b3 | flag derived with `any()` instead of `all()` | 4 |
| b4 | refusal only fires above two distinct dtypes | 9 |
| b5 | refusal moved ahead of the no-camera refusal | 0 |
| b6 | refusal names the cameras but not their dtypes | 2 |
| b7 | refusal names the dtypes but not the cameras | 4 |
| b8 | over-reach: every non-video source refused | 2 |
| b9 | only the last camera's dtype is recorded | 13 |
| b10 | the wrong fix: coerce to video instead of refusing | 13 |
| b11 | the wrong fix: coerce to image instead of refusing | 14 |

Notes on the table:

- **b1 and b4 share a firing set**, for a structural reason: a two-camera mixed source has exactly two distinct dtypes, so `> 2` is equivalent to having no guard at all.
- **b5 fires nothing, and cannot**: with no cameras the dtype set is empty, so the two refusals are mutually exclusive by construction. The order between them is not an observable.
- **b8 is the over-reach row** and fires exactly the two `image` control cases - the fix must not turn "the cameras disagree" into "the cameras are not video".
- **b11 fires the most** because coercing to `image` breaks the all-video controls as well as leaving the mixed source unrefused; it is not a milder fix.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1585 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine worktree of the merge base, normalized message sets byte-identical in both directions, 0 in the changed files.
- Paired run over an identical 374-file selection (all of `tests/transforms/` plus every test that walks the tree, reads docstrings, or reads `docs/` / `README` / `changelog.d`), with the new file excluded from both trees: identical **17898 collected** and `17 failed, 17815 passed, 66 skipped` on both, 17 identical failing ids, `comm` empty in both directions. All 17 need `awsiot` (the `[mesh-iot]` extra, absent in this environment) and are present in the base full-suite baseline.
- New file: 30 passed; `tests/transforms/` whole: 366 passed. mujoco 3.5.0 (the declared floor) with lerobot 0.6.2 from source.
- Changelog fragment validates (`scripts/assemble_changelog.py --check`, plus the 30 fragment/format guards).

### DIFF
--- a/changelog.d/2770-transform-camera-dtype-parity.md
+++ b/changelog.d/2770-transform-camera-dtype-parity.md
@@ -1,0 +1,35 @@
+### Fixed: a transformed dataset keeps the dtype each source camera declared
+
+A LeRobot dataset declares each camera's storage per feature: one
+`observation.images.<cam>` can be `dtype="video"`, encoded into an MP4 under `videos/`, while
+another is `dtype="image"`, with its frames in the data parquet. `LeRobotDataset.create` accepts
+that declaration and reads both streams back, so it is a legal source. The output side has one
+knob for the whole dataset -- `DatasetRecorder.create` takes a single `use_videos` flag that
+decides every camera's dtype -- and `_SourceDataset` derived that flag by assigning it once per
+camera inside the loop that reads the camera shapes. It therefore held whichever camera the
+source's feature dict ended on.
+
+A two-camera source declaring one video stream and one image stream transformed with
+`status="success"` while the output stored a camera a different way, and the direction depended
+only on the order the features were declared in. Declaring the video stream first flattened it
+into still frames, so a source with one MP4 produced an output with none; declaring it last
+promoted the image column to video, so the same source produced two.
+
+`create_output_recorder` documents itself as "the SOURCE schema (parity by construction)", schema
+parity is the first acceptance criterion of the transform round-trip suite, and
+`docs/data/transforms.md` contract item 1 promises a generated episode is the same trajectory
+rendered differently. A stream stored a different way is none of those.
+
+The flag is now derived from the whole camera set, and a source whose cameras disagree is refused
+by `_SourceDataset.open`, naming each camera and the dtype it declared so an operator knows which
+stream to convert. That is the disposition `open` already takes for a source declaring a feature
+the pass-through cannot preserve, and for the same reason: the output recorder declares one dtype
+for every camera, so a mixed source cannot be reproduced, and writing it anyway would alter a
+column the contract promises to carry through. Re-encoding to one dtype instead would be a quieter
+version of the same answer -- coercing to `image` also flattens the video streams of sources that
+work today.
+
+Nothing changes for a source whose cameras agree, whichever dtype that is: an all-video recording
+still transforms to video and an all-image one still transforms to images, because a homogeneous
+set already derived the right flag whichever camera the loop ended on. Every camera's dtype is now
+recorded rather than only the last one, so the refusal can report the disagreement it found.
--- a/docs/data/transforms.md
+++ b/docs/data/transforms.md
@@ -23,6 +23,15 @@ record (strands-robots)  ->  transform (this page)  ->  train (create_trainer)
    the action, state and task columns from the source episode unchanged. A
    generated episode is the *same trajectory* rendered differently, never a
    different trajectory.
+
+   Schema parity is part of that promise, down to how each camera is stored.
+   The output dataset declares one `dtype` for every camera, so a source whose
+   `observation.images.*` streams disagree - one `video`, one `image` - cannot
+   be reproduced: writing it anyway would re-encode a camera, flattening a
+   video stream into still frames or promoting an image column to video, and
+   which way it went would depend only on the order the features were declared
+   in. Such a source is refused, naming each camera and the dtype it declared;
+   re-record or convert it so every camera stream shares one dtype.
 2. **Provenance is mandatory.** Every generated episode is recorded in the
    output dataset's `meta/provenance.json` with `synthetic=true`, the source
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Fixes an order-dependent schema corruption in `_SourceDataset`: `use_videos` was reassigned once per camera inside the shape-reading loop, so it held whichever camera the source's feature dict ended on, and a legal mixed-dtype source (one `video` camera, one `image` camera) transformed with `status="success"` while silently re-encoding a stream in whichever direction the declaration order dictated. The fix records every camera's dtype, derives the flag from the whole set (`all(dtype == "video")`, which is exactly what the old code produced for homogeneous sources), and refuses a mixed source in `_SourceDataset.open` with a message pairing each camera with the dtype it declared. The refusal reuses the existing error-string channel and sits after the unsupported-feature and no-camera refusals, before any episode is read, so a refused source leaves nothing on disk. Refusing rather than coercing is the right disposition per the pass-through contract (`docs/data/transforms.md` item 1) and AGENTS.md's no-silent-defaults rule; the previously "working" mixed path was silent data corruption, so turning it into an explicit named error is not a silent behaviour change.

## What's good

- The new test file records the premise (a mixed-dtype source is a dataset LeRobot itself writes and reads back) and holds an over-reach control class, so the refusal cannot drift into refusing homogeneous non-video sources.
- Every recording call in the tests names its own root under `tmp_path` (AGENTS.md rule 15), test class names describe behaviour rather than provenance, and all new strings are ASCII.
- Changelog fragment, docs contract update, and inline rationale comments all land in the same PR; the refusal message names the exact streams an operator has to convert.
- Homogeneous sources are behaviour-identical by construction: `all()` over an agreeing set equals the old last-camera value, and the empty set is unreachable because the no-camera refusal fires first.

PR #2772, opened 2026-08-26
fix(tools): refuse an episode whose frames are not all readable

cagataycali:
## The defect

`sample_frames` reports two motion statistics its own docstring says a judge grades motion by: `max_state_delta` ("for spotting discontinuities and teleports") and `rms_state_jerk` ("so a text-only judge can ground `jerky_motion` from state alone"). Both are computed over **consecutive** rows of the episode's frame series.

The reader behind them, `_episode_frame_rows`, walked `data/**/*.parquet` and skipped any shard it could not read, with a comment saying the skip "mirrors read_dataset_episode_indices". It does not. `read_dataset_episode_indices` tolerates the same damage **and reports it** - it returns every damaged file in `unreadable_files`, and its docstring states that "any non-empty `unreadable_files` means the totals are a lower bound and the dataset must not be certified as complete". `_episode_frame_rows` took the tolerance without the report, so the surviving rows reached the statistics as though they were consecutive.

Measured on a recording that moves a uniform 0.01 per step - no discontinuity anywhere - with one of three data shards made unreadable the way a truncated download leaves it:

| | healthy | one middle shard unreadable |
|---|---|---|
| `status` | `success` | `success` |
| `length` reported by `sample_frames` | 18 | **12** |
| `length` reported by `load_episode` | 18 | 18 |
| `max_state_delta` | 0.010 | **0.070** (7.0x) |
| `rms_state_jerk` | 4.5e-12 | **6123.7** |
| anything naming the loss | - | **no** |

So the tool fabricates evidence for exactly the failure modes the judge grades. `jerky_motion` and `drift` are both in `FAILURE_MODES`, and `write_label` persists the judge's grade into the sidecar that filters training data - so a truncated download ("the usual outcome of an interrupted sync or hub download", in the sibling reader's own words) silently produces a wrong label on a good episode. It was also visible as two tools disagreeing about one dataset, both `success`.

the surviving rows read as a discontinuity

Left: the series the motion summary is computed over. Right: what each tree reports for that episode. Both panels are driven from one capture script run against each tree, and every number quoted in the figure is asserted from the two captures before it is drawn.

## The change

Collect the unreadable shards and refuse, naming every one of them:

```
sample_frames: Episode 0 cannot be sampled: 1 of 3 data shard(s) under <root> are
unreadable, so the frame series would have a hole in it and the motion summary would
measure the gap rather than the robot. Repair or re-download the dataset.
data/chunk-000/file-001.parquet: ... Parquet magic bytes not found in footer ...
```

Why refuse rather than report-and-continue, which is what the sibling does: the sibling's product *is* the damage report, so it has a field to carry the caveat. Here the product is a statistic, and there is nowhere to put "this number may be measuring a gap" that a judge will act on - a judge handed `rms_state_jerk=6123` plus a footnote may still tag `jerky_motion`. The break table below measures that alternative: recording the damage without refusing leaves the fabricated statistics reaching the judge exactly as before.

The refusal precedes the empty-series check, so a wholly unreadable episode reports the damage rather than "has no frames", which would name the wrong fault for frames that are present but unreadable.

Scope held deliberately narrow: a parquet carrying no `episode_index` column is still skipped quietly - that is a non-frame table, not damage - and `load_episode` reads the episode metadata rather than the frames, so it still describes such an episode including its true length.

## Tests

New `tests/tools/test_episode_judge_frame_series_completeness.py`, 17 cases in five classes. Pre-fix split with the source reverted to `main`: **9 failed / 8 passed**, and the 8 that pass are the whole of `TestThePremise` (3), `TestWhatIsUnchanged` (4) and `TestTheSiblingReaderStillTolerates` (1) - no control or premise cell fails pre-fix.

The premise class measures the fixture's own properties rather than asserting them: that the episode really spans several shards, that the healthy recording really is smooth, and that dropping the middle shard really does leave a seam 7x the true step. Those hold whichever way the reader disposes of the damage.

### Break table

11 mutations, control clean, source restored byte-identical. The `sib` column is the two pre-existing judge suites (`test_episode_judge.py`, `test_episode_judge_single_element_state.py`), 31 cases.

| mutation | new file | sib |
|---|---|---|
| control (unmodified) | 0 | 0 |
| the whole fix reverted (`main`) | 9 | 0 |
| damage recorded but never refused | 9 | 0 |
| refusal placed after the empty-series check | 1 | 0 |
| only the first damaged shard recorded | 2 | 0 |
| refusal does not name the shard | 2 | 0 |
| refusal does not count the shards | 3 | 0 |
| refusal drops the reason | 1 | 0 |
| over-reach: a non-frame parquet counted as damage | 1 | 0 |
| over-reach: refuse whenever any shard is skipped | 4 | 0 |

8 distinct firing sets across the 9 non-control rows. Reverting the fix and keeping the collection-without-refusal fire the same 9, which is the point: the `raise` is the behavioural fix and the collection alone changes nothing a caller can see. Both over-reach rows fire, one of them on the control that pins a non-frame parquet still being skipped.

`sib` is 0 on every row, so neither pre-existing suite could see any of this. The reason is in their fixtures: both write a single data shard, and a single-shard dataset cannot express a partial loss - a lost sole shard is the already-handled "no frames" case.

## Gate

- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ`: clean, 1591 files formatted.
- `mypy strands_robots tests tests_integ`: **5 == 5** against a pristine `main` worktree, normalized message sets byte-identical in both directions, 0 in the changed files.
- Paired run, identical 429-file selection (all of `tests/tools/` plus every guard that walks the tree, reads source or docstrings, reads `docs/`, or names the judge / label / recorder modules), with the new file excluded from both trees so the selection is identical in content: **20225 collected and `24 failed, 20091 passed, 111 skipped` on both**, 24 failing ids each, `comm` empty in both directions. 17 need `awsiot` (`awsiotsdk`, declared in the `mesh-iot` extra, absent here); 4 are `tests/simulation/test_recording_frame_loss_is_not_tolerated.py`, which passes 13/13 in isolation on this branch - a large-selection module-caching artifact, identical on both trees; 2 are `tests_integ` emoji assertions that fail on pristine `main` in isolation; 1 is a stale dataset directory left by another process.
- Coverage of the changed module, from a scoped `tests/tools/` run verified faithful against a full-suite run on the base (identical statement, miss and missing-line figures): **94% -> 96%**, 12 -> 8 missing. The four newly covered statements are the empty-dataset refusal and the shard-skip branches. Every remaining uncovered line is either an unreachable version-ladder rung or a domain refusal in a sibling tool.

Rebased onto `b7544254`; the two commits `main` added meanwhile touch none of this change's files.


yinsong1986 (dismissed):
## Summary

Converts `_episode_frame_rows`'s silent skip of an unreadable data shard into a refusal: since `max_state_delta` and `rms_state_jerk` are computed over consecutive rows, dropping a shard made the surviving rows read as consecutive across a gap, fabricating exactly the discontinuity/jerk evidence the judge grades (`0.010 -> 0.070`, `0.000 -> 6123.7` on a uniform ramp) under `status="success"`. The refusal is a `ValueError` raised in the helper and folded into the tool's error envelope by `sample_frames`' existing handler, names every damaged shard (not just the first), and correctly diverges from `read_dataset_episode_indices`, whose tolerance is justified because reporting damage via `unreadable_files` is its product (verified against `dataset_recorder.py`'s documented lower-bound contract). The dataset-wide scope of the refusal (any unreadable shard refuses the episode) is the right fail-safe, since an unreadable shard's episode membership is unknowable without reading it. The new test module is well-scoped: premise measurements on the fixture, envelope/no-raise pinning, controls for the healthy path, non-frame parquet skip, empty dataset, and the sibling reader's unchanged tolerance; ASCII throughout, `tmp_path`-rooted, no host paths. Docs and the `Raises:` block are updated in the same change per AGENTS.md. One note: the PR description is `placeholder` and carries no `Closes #N` line, and no `changelog.d/` fragment is in the diff - both are surfaced by their own CI workflows and neither blocks correctness.

## What's good

- Fail-fast direction matches AGENTS.md ('No silent defaults on error' / 'Fail-fast with strict=True'); the stale 'mirrors read_dataset_episode_indices' comment is corrected rather than left to mislead.
- The refusal reports all damage at once with an actionable remedy, in ASCII, through the tool's error-dict contract rather than an escaping raise.
- Regression tests measure the premise on the fixture itself and pin the controls that scope the refusal, so the narrowing stays deliberate.

yinsong1986 (approved):
## Summary

Converts `_episode_frame_rows`'s silent skip of an unreadable data shard into a refusal, so `sample_frames` can no longer compute `max_state_delta` / `rms_state_jerk` over a frame series with a hole in it - on a uniform ramp, losing one middle shard of three fabricated `0.010 -> 0.070` and `0.000 -> 6123.7` under `status="success"`, which is manufactured evidence for exactly the failure modes (`jerky_motion`, `drift`) the judge grades and `write_label` persists. The refusal collects every damaged shard before raising, counts them against the episode's shard total, names each with its read error, and precedes the empty-series check so a wholly unreadable episode reports damage rather than "has no frames". The `ValueError` is folded into the tool's error envelope by `sample_frames`' existing handler (`_episode_frame_rows` has no other caller), so the module's never-raises contract holds - and a test pins the envelope shape explicitly. Scope is held deliberately narrow: a non-frame parquet (no `episode_index` column) is still skipped quietly, `load_episode` still reports the true metadata length, and the sibling `read_dataset_episode_indices` keeps its report-and-tolerate contract, with the divergence justified in the docstring (the sibling's product is the damage report; a statistic has no field to carry that caveat). The stale "mirrors read_dataset_episode_indices" comment is corrected rather than left to mislead, the `Raises:` block names the new refusal per AGENTS.md, and the changelog fragment plus docs update ship in the same change. Both notes from the prior (dismissed) review - placeholder description and missing fragment - are resolved at this head.

## What's good

- Fail-safe direction matches AGENTS.md ('No silent defaults on error'); the break table shows the `raise` is the behavioural fix and collection-without-refusal changes nothing observable.
- The new test module measures its premise on the fixture (the seam really reads as a 7x discontinuity) rather than asserting it, pins the controls that scope the refusal, and is `tmp_path`-rooted, ASCII-only, behaviour-named.
- Refusal text is actionable (names every shard, the count, the reason, and a remedy) and stays inside the `{status, content}` envelope contract.

PR #2773, opened 2026-08-26
fix(g1): record the DDS interface ensure_dds bound, not the one it was asked for

cagataycali:
## The defect

`strands_robots/tools/g1/_g1_common.py` opens by stating that
`ChannelFactoryInitialize` must run "exactly once per process, with a known
network interface", and it keeps a module-level record of that interface so a
later caller asking for a different one is refused instead of silently re-bound
onto the wrong NIC. The comment on that record is explicit about the policy:

> A second call with a different interface is a bug worth catching, not a silent no-op.

The record was only sometimes true. `ChannelFactory.Init` is:

```python
def Init(self, id: int, networkInterface: str = None, qos: Qos = None):
    if self.__class__.__initialized:
        return True
    ...
```

It short-circuits and never reads `networkInterface`, so a second
`ChannelFactoryInitialize` returns normally having bound nothing - a no-op that
is indistinguishable, at the call site, from a successful bind. `ensure_dds`
treated that quiet return as a bind and recorded the interface it had asked for.

Measured against `unitree_sdk2py` 1.0.1, with the bus brought up on `lo` by
anything other than `ensure_dds` (a caller following the SDK's own examples, or
another library in the same process - the exact case the module docstring names):

| | before | after |
|---|---|---|
| `ensure_dds("eth-does-not-exist")` | `None` (success) | named refusal |
| recorded interface | `'eth-does-not-exist'` | `None` |
| factory actually bound to | `lo` | `lo` |
| later `ensure_dds("another-nic")` | refused, quoting `'eth-does-not-exist'` | refused, naming the real cause |

Two consequences. Every subscriber and publisher attached to whichever NIC the
first caller chose while the helper reported the one this caller asked for -
the silent re-bind, producing empty topics with no obvious cause, that the
refusal exists to prevent. And the refusal that did eventually fire quoted an
interface the bus was never on, sending a reader after a fault that does not
exist.

## The change

Ask the factory whether it is already bound *before* calling init. A bus this
process did not bind is refused by name, giving the cause and the call to drop,
and nothing is recorded - so the interface a later refusal compares against is
only ever one `ensure_dds` actually bound.

```
the DDS channel factory was already initialised outside ensure_dds, so a bind to
'eth-does-not-exist' cannot be confirmed; let the driver initialise the bus, or
drop the ChannelFactoryInitialize call that runs before it
```

The SDK build that raises `already initialized` rather than short-circuiting is
the same situation and now reaches the same answer. That also closes a second,
latent hole in the same branch: its `"already" in message or "initialized" in
message` test admits a genuine failure whose text merely contains "initialized"
(`"could not be initialized"`), and that used to be recorded as success. Both
arms now refuse, so the over-broad match can no longer convert a failure into a
reported bind.

### Why the probe reads an SDK attribute

Nothing public distinguishes a bind from a no-op: `Init` returns `True` for both
and `ChannelFactoryInitialize` discards that bool, raising only when `Init` is
falsy. `ChannelFactory`'s public surface is
`Init, CreateChannel, CreateSendChannel, CreateRecvChannel` - none reports the
state. So the probe reads the attribute the factory records it on, guarded, and
reports "cannot tell" rather than guessing when a build does not publish it; in
that case the caller keeps exactly the behaviour it had before. This is a
*narrower* coupling to the SDK than the mechanism already shipping in this
function, which matches the vendor's exception text.

`TestTheRealSdkBehavesTheWayTheDoubleDoes` pins all three of those SDK facts
against the installed package, so a rename fails there rather than silently
disabling the probe, and skips when the SDK is absent rather than passing
vacuously.

### Behaviour change

A process that brought the bus up itself and then called `ensure_dds` used to
get success on a possibly-wrong NIC; it now gets a named, actionable refusal
that `_dds_engine` surfaces as a connect failure. That is the module's own
stated policy applied to the case it previously could not see, and it replaces
empty topics with a reason. No path where `ensure_dds` performed the bind
changes at all.

## Why every gate was silent

`ensure_dds` had never succeeded under test. The only path any suite reached was
"first call, init raises, message is not `already`" - so the idempotence
contract the module docstring leads with, the interface-mismatch refusal, and
the success bookkeeping were all unexecuted (lines 106-114 and 130-137, 11 of
39 statements). Every mutation in the table below fires **zero** of the 245
pre-existing tests in `tests/tools/g1/`, `tests/drivers/` and
`tests/test_driver_seam.py`.

## Tests

`tests/tools/g1/test_dds_init_binds_a_known_interface.py`, 19 cases. Every cell
drives a fake `unitree_sdk2py.core.channel`: binding the real factory is a
process-wide side effect with no `Shutdown`, so a suite that did it once would
change what every later test observes.

Behavioural pre-fix split (both halves reverted, new symbols kept) - **7 failed,
12 passed**:

|---|---|---|
| `TestABusBoundElsewhereIsNotRecordedAsOurs` | 7 / 7 | the regression |
| `TestTheDoubleIsFaithfulToTheCallSite` | 0 / 2 | premise |
| `TestAnAttestedBindIsUnchanged` | 0 / 4 | control - holds pre-fix too |
| `TestAnUnreadableSdkKeepsTheOlderBehaviour` | 0 / 3 | over-reach control |
| `TestTheRealSdkBehavesTheWayTheDoubleDoes` | 0 / 3 | SDK parity |

### Mutations

11 mutations, 11 distinct failing-id sets, control clean, source restored
byte-identical. `sib` counts failures in the pre-existing G1 suites; `coll` is
19 on every row, so nothing is hidden by a deselection.

|---|---|---|---|
| b0 | 0 | 0 | control - no mutation |
| b1 | 6 | 0 | probe call removed from `ensure_dds` |
| b2 | 1 | 0 | SDK-refuses branch restored to record the requested NIC |
| b3 | 7 | 0 | both halves reverted (behavioural revert) |
| b4 | 2 | 0 | bound factory recorded as ours instead of refused |
| b5 | 4 | 0 | probe runs *after* `ChannelFactoryInitialize` |
| b6 | 7 | 0 | probe accepts any non-`None` flag (`isinstance` dropped) |
| b7 | 3 | 0 | probe treats an unreadable flag as already bound |
| b8 | 1 | 0 | refusal drops the interface it could not confirm |
| b9 | 1 | 0 | refusal drops the named cause |
| b10 | 1 | 0 | refusal drops the call to remove |

`b1 | b2 == b3` exactly and `b1` and `b2` share no failing test, so the two halves of the fix are
independently pinned. b4 is the tempting wrong fix - record it and return
success - and b5 shows the probe's position is load-bearing rather than
stylistic: run after init, and it reports the state our own call just created.

## Gate

- `ruff check strands_robots tests tests_integ`: clean. `ruff format --check`: 1585 files already formatted.
- `mypy strands_robots tests tests_integ`: **24 on this branch, 24 on the merge-base**, normalized message sets byte-identical in both directions, **0 in either changed file**.
- Paired run, identical 364-file selection (all `tests/tools/g1/` and `tests/drivers/`, plus every suite that walks the tree, reads source or docstrings, or names `ensure_dds`), the new file excluded from both trees: **17725 collected and `18 failed, 17647 passed, 60 skipped` on both**, failing-id sets identical, `comm` empty both directions. All 18 are pre-existing and environmental: 17 need `awsiot` (the `[mesh-iot]` extra, absent here) and one is a leftover dataset fixture.
- On the rebased base: `tests/tools/g1/ tests/drivers/ tests/test_driver_seam.py` **264 passed**, which includes #2768's and #2750's new suites.

### Coverage

`_g1_common.py` **72% -> 100%** (11 -> 0 missing; statements 39 -> 47). The
scoped selection is a faithful oracle for this module - it reports the same
`39 / 11 / 72%` and the same missing ranges as the full 40437-item run.

## Composition

#2768 landed mid-review and changes `_dds_engine.py`, which is `ensure_dds`'s
only in-tree caller. The two are disjoint: that PR changes how endpoints are
released, this one changes what `ensure_dds` records, and no path where
`ensure_dds` performed the bind is touched. Verified by re-running the whole
mutation table and both suites on the rebased tree - every row is unchanged.

No visual artifact: this changes a refusal and a bookkeeping record, not a
policy, simulation, rendering, recording or asset. The measured before/after
table and the mutation table are the evidence.

### DIFF
--- a/changelog.d/2773-dds-init-records-the-bound-interface.md
+++ b/changelog.d/2773-dds-init-records-the-bound-interface.md
@@ -1,0 +1,48 @@
+### Fixed: ensure_dds records the interface it bound, not the one it was asked for
+
+The G1 DDS helper's first stated requirement is that `ChannelFactoryInitialize`
+runs "exactly once per process, with a known network interface", and it keeps a
+module-level record of that interface so a later caller asking for a different
+one is refused rather than silently re-bound onto the wrong NIC. Its own comment
+on that record says a second call with a different interface "is a bug worth
+catching, not a silent no-op".
+
+The record was only sometimes true. `ChannelFactory.Init` short-circuits on
+`if __initialized: return True` and never reads its `networkInterface` argument,
+so a second `ChannelFactoryInitialize` returns normally without binding
+anything - a no-op indistinguishable, at the call site, from a successful bind.
+`ensure_dds` read that quiet return as a bind and recorded the interface it had
+asked for. Measured against `unitree_sdk2py` 1.0.1, with the bus brought up on
+`lo` by anything other than `ensure_dds`:
+
+```
+ensure_dds("eth-does-not-exist")  ->  None          # reported success
+recorded interface                ->  'eth-does-not-exist'
+factory actually bound to         ->  'lo'
+ensure_dds("another-nic")         ->  "...was called with interface
+                                       'eth-does-not-exist'; refusing to
+                                       re-initialise on 'another-nic'"
+```
+
+So every subscriber and publisher attached to whichever NIC the first caller
+chose while the helper reported the one this caller asked for - the silent
+re-bind, producing empty topics with no obvious cause, that the refusal exists
+to prevent - and the refusal that did eventually fire quoted an interface the
+bus was never on, sending a reader after the wrong fault.
+
+`ensure_dds` now asks the factory whether it is already bound before calling
+init. A bus this process did not bind is refused by name, giving the cause and
+the call to drop, and nothing is recorded - so the interface a later refusal
+compares against is only ever one `ensure_dds` actually bound. The SDK build
+that raises "already initialized" instead of short-circuiting reaches the same
+answer for the same reason, which also stops that branch's substring match
+turning a genuine init failure whose message contains "initialized" into a
+reported success. A build that does not publish its factory state keeps the
+behaviour it had before the probe existed.
+
+Nothing public distinguishes a bind from a no-op: `Init` returns `True` for
+both and `ChannelFactoryInitialize` discards that bool, so the probe reads the
+attribute the factory records its state on. That is a narrower coupling to the
+SDK than matching its exception text, which this function already does. The
+module docstring claimed a second init raises from `unitree_sdk2py`; it does
+not, and that claim is corrected to what the SDK does.
--- a/strands_robots/tools/g1/_g1_common.py
+++ b/strands_robots/tools/g1/_g1_common.py
@@ -4,10 +4,13 @@ Two things need to be true across the driver and (later, issue #358) the
 agent tools:
 
 * ``ChannelFactoryInitialize`` is called exactly once per process, with a
-  known network interface. Calling it twice from separate imports raises
-  from ``unitree_sdk2py``; calling it never leaves subscribers with no bus.
-  :func:`ensure_dds` runs it under a lock, records the interface, and is
-  idempotent for callers with the same choice.
+  known network interface. Calling it never leaves subscribers with no bus;
+  calling it a second time is *not* an error the SDK reports -
+  ``ChannelFactory.Init`` short-circuits on ``if __initialized: return True``
+  and ignores its ``networkInterface`` argument, so a re-init is a silent
+  no-op that is indistinguishable from a successful bind.
+  :func:`ensure_dds` runs it under a lock, records the interface it actually
+  bound, and is idempotent for callers with the same choice.
 * A ``ChannelSubscriber`` and a ``ChannelPublisher`` cannot be constructed
   concurrently: the CycloneDDS bindings segfault. :data:`_DDS_INIT_LOCK` is
   the *shared* lock the driver and the tools (issue #358) both hold while
@@ -21,6 +24,7 @@ mock the bus and skips the SDK entirely on Thor.
 
 from __future__ import annotations
 
+import importlib
 import logging
 import threading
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Fixes a bookkeeping lie in the G1 DDS helper: `ChannelFactory.Init` short-circuits on its private `__initialized` flag without reading `networkInterface`, so a second `ChannelFactoryInitialize` is a silent no-op that `ensure_dds` used to misread as a bind, recording whatever interface the caller asked for. The PR probes the factory's state before init and refuses, by name, a bus this process did not bind - so the recorded interface is only ever one `ensure_dds` actually bound, and a later mismatch refusal can no longer quote a fabricated NIC. The same refusal now covers the SDK build that raises `already initialized`, which also closes the latent hole where the over-broad `"initialized" in message` substring match converted a genuine init failure into a reported success.

## What's good

- The private-attribute probe is properly hedged: `_sdk_factory_already_bound` returns `None` (keep prior behaviour) rather than guessing when a build does not publish the flag, and `TestTheRealSdkBehavesTheWayTheDoubleDoes` pins the attribute name, the short-circuit, and the discarded bool against the installed SDK so a rename fails loudly instead of silently disabling the probe (skips without the SDK rather than passing vacuously).
- The behaviour change is loud, not silent: the refusal names the cause, the interface it could not confirm, and the call to drop, and `_dds_engine` surfaces it as a connect failure before any subscriber attaches to nothing.
- Tests never touch the real factory (a process-wide side effect with no `Shutdown`), use `monkeypatch.setitem(sys.modules, ...)` per AGENTS.md, carry behaviour-describing names, and both new suites plus the pre-existing G1/driver suites pass on the head (16 passed / 3 skipped; 211 passed / 8 skipped).
- Changelog fragment present; strings are plain ASCII; scope is tight (one module, its test, its fragment).

## Verification suggestions

On a machine with `unitree_sdk2py` installed, run `pytest tests/tools/g1/test_dds_init_binds_a_known_interface.py::TestTheRealSdkBehavesTheWayTheDoubleDoes -q` to exercise the three SDK-parity pins that skip in CI.

PR #2774, opened 2026-08-26
fix(device_connect): defer device_connect_edge import to first use (fixes #2771)

### LINKED ISSUE #2771: device_connect package init traps the stdlib-only reachy_transport leaf behind the [device-connect] extra
## What

`strands_robots/device_connect/reachy_transport.py` is a leaf module that imports **nothing but the standard library** (`asyncio`, `functools`, `json`, `logging`, `math`, `os`, `socket`, `abc`, `collections.abc`, `typing`). It provides `api()`, `WebSocketLink`, `ZenohLink` and `rpy_to_pose` — the Reachy daemon wire helpers.

It is nevertheless unimportable on a stock `pip install strands-robots`, because importing it executes the parent package `__init__`, and `strands_robots/device_connect/__init__.py` imports unconditionally:

```python
from device_connect_edge import DeviceRuntime
from strands_robots.device_connect.reachy_mini_driver import ReachyMiniDriver
from strands_robots.device_connect.robot_driver import RobotDeviceDriver
from strands_robots.device_connect.sim_driver import SimulationDeviceDriver
```

All four require `device_connect_edge`, which ships only in the `[device-connect]` / `[all]` extras (`pyproject.toml`, `device-connect = ["device-connect-edge>=0.2.0", ...]`). The three sibling driver modules each do `from device_connect_edge.drivers import ...` at module level, so the requirement is genuine for them — they exist to integrate with Device Connect.

Measured on a stock install (no extras):

```
>>> import strands_robots.device_connect.reachy_transport
ModuleNotFoundError: No module named 'device_connect_edge'
```

## Why it matters

The native `ReachyDriver` (#2762) routes every daemon touch through this leaf. It therefore cannot reach hardware without an extra whose own subject — the Device Connect edge runtime — it does not use. PR #2762 makes that honest: the driver refuses by name and points at `pip install 'strands-robots[device-connect]'` rather than raising `ModuleNotFoundError` through the agent tool surface. But a *named refusal is not the same as working*, and the extra it names is the wrong dependency to demand for a driver that only needs a stdlib module.

The same absence also fails 72 of `tests/drivers/test_reachy_driver.py`'s cases in a no-extra environment, at the `monkeypatch.setattr("strands_robots.device_connect.reachy_transport.api", ...)` line. CI never sees this because the hatch env installs `[all]`.

## Options

1. **Move the leaf out from under the extra-bearing package.** e.g. `strands_robots/transport/reachy.py`, re-exported from `strands_robots.device_connect.reachy_transport` for compatibility. Cleanest separation: a stdlib-only wire helper stops living inside an integration package. Both consumers (the native driver and the Device Connect driver) then import it with no extra.
2. **Make the package `__init__` lazy.** Move `DeviceRuntime` and the three driver imports into function bodies / a module `__getattr__`. Conflicts with the AGENTS.md convention that `__init__.py` is exports-only, and the `init_device_connect` helpers there are real logic already — so this is the larger change of the two.

Option 1 is preferred: it is a move plus a re-export, and it makes the dependency statement true in both directions rather than deferring it.

## Acceptance

- `import strands_robots.device_connect.reachy_transport` succeeds with no extras installed.
- `Robot("reachy_mini", mode="real")` reaches a daemon on a stock install (mocked at the socket, as today).
- The Device Connect driver path is unchanged and still requires `[device-connect]`, because it genuinely does.
- A test pins the leaf's import-time dependency set so a future edit cannot quietly re-trap it — e.g. importing it in a subprocess with `device_connect_edge` blocked in `sys.modules`.
- The 6 refusal tests added by #2762 keep passing: the refusal path stays correct for a genuinely absent dependency, it just stops firing for this one.

## Not urgent

#2762 ships a correct, named refusal, so nothing crashes and nothing lies. This issue is about removing an install requirement that is not real, not about a live failure.

Found while addressing review feedback on #2762.

cagataycali:
## The defect

`strands_robots/device_connect/__init__.py` opens with `from device_connect_edge import DeviceRuntime`. `device-connect-edge` lives only in the `[device-connect]` / `[all]` extras, so this import fails on any install without them. Sitting in the same package is `reachy_transport` — stdlib-only, imported by the native Reachy driver on every daemon touch (`_daemon_get`, `_wire_commands`, `_build_link`). Importing that leaf executes the package `__init__` first, so a stock `pip install strands-robots` traps the leaf behind the extra.

CI never saw it because the hatch env installs `[all]`. Measured on a stock install (`device_connect_edge` blocked at `sys.meta_path`), the yinsong1986 [MUST FIX] on PR #2762 reported this exact symptom on the driver surface:

```
connect_eagerly RAISED: ModuleNotFoundError: No module named 'device_connect_edge'
send_action     RAISED: ModuleNotFoundError: No module named 'device_connect_edge'
```

Three `AGENTS.md` contracts break at once:

1. `connect_eagerly`'s own "named reason on failure — the driver is left disconnected but usable" (raises through the caller instead)
2. AGENTS.md > Review Learnings (PR #85) > Error Handling Contracts > "Return error dicts, never raise"
3. AGENTS.md > Key Conventions #7 ("Use `require_optional()` for all optional deps")

## The fix

`__init__` now uses PEP 562 `__getattr__` to resolve every public name lazily from a private `_impl` submodule that carries the `device_connect_edge`-dependent implementations. The package itself is stdlib-only; the extra is imported the first time a caller reaches for one of the six public names that need it.

```
strands_robots.device_connect                       # stdlib only
├── __init__.py             ← __getattr__ dispatches by name (stdlib only)
├── _impl.py                ← imports device_connect_edge (loaded on first use)
├── reachy_transport.py     ← stdlib only (unchanged)
├── reachy_mini_driver.py   ← imports device_connect_edge (unchanged)
├── robot_driver.py         ← imports device_connect_edge (unchanged)
└── sim_driver.py           ← imports device_connect_edge (unchanged)
```

Submodule lookup (`getattr(pkg, 'reachy_transport')`) is also served by `__getattr__` — falling through to `importlib.import_module(f'{__name__}.{name}')` after the export table lookup — so both spellings a caller uses to reach `reachy_transport` (attribute access and `from ... import`) resolve to the same module. Non-existent names still raise `AttributeError` naming the module and the missing symbol.

### What is preserved

| behaviour | mechanism |
|---|---|
| `import strands_robots.device_connect` succeeds without the extra | package init is stdlib-only |
| A name that needs the extra raises `ModuleNotFoundError` naming `device_connect_edge` | `__getattr__` delegates to `importlib.import_module` which raises with its own text |
| `dir(strands_robots.device_connect)` advertises every public name | `__dir__` override merges `__all__` into `globals()` |
| `isinstance(x, RobotDeviceDriver)` holds across calls | first access caches the resolved value on the package's `globals()`, subsequent lookups skip `__getattr__` entirely |
| Existing test imports unchanged | Python's own import machinery serves fully-qualified submodule imports before `__getattr__` is consulted |

### The regression pins the class of bug (14 cells across 4 classes)

`tests/device_connect/test_package_init_does_not_import_the_extra.py` grades four independent behaviours against a `sys.meta_path` finder that pretends `device_connect_edge` is uninstalled:

| class | asserts |
|---|---|
| `TestThePackageInitDoesNotImportTheExtra` (3) | package imports; stdlib-only leaf reachable via `importlib` and `getattr`; `device_connect_edge` never enters `sys.modules` as a side-effect |
| `TestTheExtraLoadsOnFirstUse` (6+1) | each of the six public names raises `ModuleNotFoundError` naming the extra when absent; each resolves to a callable when the extra is installed |
| `TestMissingNamesAreNamed` (2) | typo raises `AttributeError` naming module + attribute; `dir()` advertises every public name |
| `TestLazyLoadingIsIdempotent` (2) | resolved names are cached; second access is a plain attribute read |

**11 pass** on the fix branch; **3 skip** correctly (grade the with-extra path, `device_connect_edge` not installed in the paired gate). On `main`, `TestThePackageInitDoesNotImportTheExtra` fails with the exact `ModuleNotFoundError: No module named 'device_connect_edge'` — the same error the yinsong1986 MUST FIX documented.

### Paired-tree gate

On a stock install with no extras (`pip install -e .` only), running the whole `tests/device_connect/`, `test_device_connect_*`, `test_reachy_transport_links`, `test_docs_device_connect_env_reference` set:

| | fix branch | `main` (b7544254) |
|---|---|---|
| passed | 1113 | 1098 |
| failed | **0** | **4** |
| skipped | 10 | 7 |
| errors (collection, need `device_connect_edge`) | 52 | 52 |

Strict improvement: 15 more passes (11 new + 3 skipped correctly + 1 ordering win), zero regressions, four pre-existing test-ordering-flaky failures on `main` are gone. The 52 collection errors are pre-existing environmental — those tests import `device_connect_edge` at collection time, i.e. they need the `[device-connect]` extra to run at all. That was already true on `main`; this PR does not change it.

### Lint / type / format

- `ruff check strands_robots/device_connect/ tests/device_connect/`: clean
- `ruff format --check strands_robots/device_connect/ tests/device_connect/`: 9 files already formatted
- `mypy strands_robots/device_connect/__init__.py strands_robots/device_connect/_impl.py`: 0 issues in 2 source files

## Composition

This PR closes issue #2771 (opened 30 minutes before Fire 11 by cagataycali) and resolves the outstanding `[MUST FIX]` from yinsong1986 on PR #2762 (Reachy Mini native driver). #2762 imports `strands_robots.device_connect.reachy_transport` from its driver's four transport call sites; once this PR merges, those sites work on a default install without any change to #2762 itself.

No overlap with in-flight PRs: #2767 (`feat/g1-send-action-wired`), #2772 (`fix/judge-refuses-an-incomplete-frame-series`) and #2773 (`fix/dds-init-records-the-interface-it-bound`) each edit disjoint files.

Closes #2771.

## Review rounds

**Round 1** - `006188fb`, `fc916f85`. The lazy dispatcher served only the six
`__all__` names plus real submodules, which dropped four module-level names the
tree reaches through the package: 25 introduced failures and one collection
error across five existing test files, measured over a 138-file selection.
`fc916f85` extends the export table with `DeviceRuntime`, `_build_heartbeat`,
`_INIT_TIMEOUT_S`, `_INSECURE_TRUE` and, separately, routes three reads in
`_impl` (`DeviceRuntime`, `init_device_connect`, the wait budget) back through
the package at call time. The second half is what the extra table entries alone
could not fix: patching the package while the code read `_impl`'s globals turned
8 loud `AttributeError`s into 8 silent no-ops, with
`test_a_string_argument_never_constructs_a_runtime` asserting against an object
the code never read. Introduced failures: 25 -> 1.

**Round 2** - `b2131ca0`, `063c8975`. The remaining failure was a docstring
concern, not an export-machinery one: two new docstrings cited
`:mod:`strands_robots.drivers.reachy``, which does not exist on `main` (it
arrives with #2762), so `tests/test_docstring_xref_roles_resolve.py` could not
pass until that landed. Both now name the module in prose without the `:mod:`
role. `b2131ca0` explains the `ValueError` pass in the meta-path fixture
teardown, which the empty-except scan flagged. `call-test-lint` green on
`063c8975`.

**Round 3** - `25b2843c`. `_INIT_TIMEOUT_S` was defined in `_impl` and read only
as `_pkg._INIT_TIMEOUT_S`. Code scanning reported the module global as unused,
which was the visible half; the load-bearing half is that resolving the name off
the package imported `_impl`, so reading a float literal with no
`device_connect_edge` dependency raised `ModuleNotFoundError` on a stock
install. A deferred import that still fires is not a deferral -- this was the
same defect class the PR exists to remove, one name smaller. The literal moved
to the stdlib-only package `__init__` and its entry left the export table;
`init_device_connect_sync` still reads it back through the package, so
`monkeypatch.setattr(module, "_INIT_TIMEOUT_S", budget)` still reaches the
bounded wait.

Pinned by `TestANameThatNeedsNoExtraIsReadableWithoutIt`, which reads the budget
off the package under the suite's existing `sys.meta_path` block and asserts the
extra did not enter `sys.modules` behind the value. It fails on `063c8975` with
`ModuleNotFoundError: No module named 'device_connect_edge'`.

| file (extra installed) | `063c8975` | `25b2843c` |
|---|---|---|
| `tests/device_connect/test_package_init_does_not_import_the_extra.py` | 14 passed | **15 passed** |
| `tests/test_device_connect_sync_bringup_outcomes.py` | 12 passed | 12 passed |
| `tests/test_device_connect_drivers.py` | 81 passed | 81 passed |
| `tests/test_allow_insecure_setting_domain.py` | 74 passed | 74 passed |
| `tests/test_device_connect_heartbeat.py` | 7 passed | 7 passed |
| `tests/test_device_connect_hardening.py` | 52 passed | 52 passed |

`ruff check` / `ruff format --check` / `mypy` clean on the touched files.
`tests/test_docstring_xref_roles_resolve.py` fails locally on both trees for an
unrelated environmental reason (`No module named 'lerobot'`); it is green in CI,
where that extra is installed.

### DIFF
--- a/strands_robots/device_connect/__init__.py
+++ b/strands_robots/device_connect/__init__.py
@@ -12,23 +12,56 @@ Usage:
     # Now discoverable via Device Connect tools:
     #   discover_devices(device_type="strands_robot")
     #   invoke_device("so100-lab-1", "execute", {"instruction": "pick up cube"})
+
+Module-load discipline
+----------------------
+
+Every symbol this package exports is behind ``__getattr__`` (:pep:`562`). The
+package imports on a stock ``pip install strands-robots`` -- with no extras --
+and only reaches ``device_connect_edge`` when a caller *uses* a name that needs
+it. The four Device Connect drivers, ``DeviceRuntime`` and the two
+``init_device_connect`` entry points all sit behind that gate.
+
+This is required, not a stylistic choice. The sibling module
+``strands_robots.device_connect.reachy_transport`` is stdlib-only and is
+imported by the native Reachy driver (:mod:`strands_robots.drivers.reachy`).
+Importing that leaf executes this ``__init__``, so a package init that eagerly
+imports ``device_connect_edge`` raises ``ModuleNotFoundError`` inside the
+native driver's first daemon touch on any install without ``[device-connect]``
+-- escaping the driver's own no-raise refusal contract and breaking three
+:mod:`AGENTS.md` conventions at once (\"Return error dicts, never raise\",
+``require_optional()`` for optional deps, the module-load discipline the
+driver's docstring makes explicit).
+
+The public names are unchanged: ``from strands_robots.device_connect import
+init_device_connect`` still works, and only fails when the caller reaches for a
+name whose implementation genuinely needs ``device_connect_edge``. Static tools
+(mypy, IDE autocomplete) see the names through the ``TYPE_CHECKING`` guard.
 """
 
-import asyncio
-import logging
-import os
-import threading
-import uuid
-from typing import Any
+from __future__ import annotations
 
-from device_connect_edge import DeviceRuntime
+import importlib
+from typing import TYPE_CHECKING, Any
 
-from strands_robots.device_connect.reachy_mini_driver import ReachyMiniDriver
-from strands_robots.device_connect.robot_driver import RobotDeviceDriver
-from strands_robots.device_connect.sim_driver import SimulationDeviceDriver
-from strands_robots.utils import is_boolean
+if TYPE_CHECKING:  # pragma: no cover - typing-only, never executed
+    from device_connect_edge import DeviceRuntime  # noqa: F401
 
-logger = logging.getLogger(__name__)
+    from strands_robots.device_connect._impl import (  # noqa: F401
+        init_device_connect,
+        init_device_connect_sync,
+        resolve_allow_insecure,
+    )
+    from strands_robots.device_connect.reachy_mini_driver import (  # noqa: F401
+        ReachyMiniDriver,
+    )
+    from strands_robots.device_connect.robot_driver import (  # noqa: F401
+        RobotDeviceDriver,
+    )
+    from strands_robots.device_connect.sim_driver import (  # noqa: F401
+        SimulationDeviceDriver,
+    )
+
 
 __all__ = [
     "init_device_connect",
@@ -39,263 +72,77 @@ __all__ = [
     "ReachyMiniDriver",
 ]
 
-_INSECURE_TRUE = ("true", "1", "yes")
 
-# ``init_device_connect_sync`` bounds the background bring-up: the awaited half
-# constructs a ``DeviceRuntime``, which can block on a broker, so the wrapper
-# must not hang its caller forever. Expiry is a failed bring-up, not a slow one
-# that later succeeds -- the wrapper has already returned by then.
-_INIT_TIMEOUT_S: float = 30.0
+# Map each exported name to the module it lives in. Grouped by whether the
+# module itself needs ``device_connect_edge`` so a reader can see at a glance
+# which names cost the extra and which are pure Python.
+#
+# All three driver modules import ``device_connect_edge`` at module load, and
+# the two ``init_device_connect*`` entry points construct a ``DeviceRuntime``
+# from it. ``resolve_allow_insecure`` is stdlib-only, kept in the same private
+# ``_impl`` module so its behavioural pin does not drag the extra in either.
+_ATTR_TO_MODULE: dict[str, str] = {
+    "init_device_connect": "strands_robots.device_connect._impl",
+    "init_device_connect_sync": "strands_robots.device_connect._impl",
+    "resolve_allow_insecure": "strands_robots.device_connect._impl",
+    "RobotDeviceDriver": "strands_robots.device_connect.robot_driver",
+    "SimulationDeviceDriver": "strands_robots.device_connect.sim_driver",
+    "ReachyMiniDriver": "strands_robots.device_connect.reachy_mini_driver",
+}
 
 
-def resolve_allow_insecure(
-    explicit: bool | None = None,
-    env_value: str | None = None,
-) -> bool:
-    """Resolve the effective ``allow_insecure`` setting (secure by default).
+def __getattr__(name: str) -> Any:
+    """Resolve a public name lazily.
 
-    Precedence: explicit arg > ``DEVICE_CONNECT_ALLOW_INSECURE`` env var >
-    secure default (``False``). Insecure transport is never implicit - it
-    must be opted into via the argument or the env var.
+    Follows :pep:`562`. Called by the Python attribute machinery only when
+    ``name`` is not already in the module's namespace, so first access pays
+    the import cost and every subsequent access is a plain attribute read.
 
-    Extracted as a pure function so the secure-by-default posture is unit
-    testable without standing up a DeviceRuntime.
-
-    The two sources carry the same setting in different shapes, and each is held
-    to its own declared type rather than to the other's. An environment variable
-    is a string by construction, so *env_value* is **parsed**: only
-    ``("true", "1", "yes")`` opt in and every other spelling is secure. The
-    argument is declared ``bool | None``, so it is **checked**: a non-boolean is
-    refused rather than parsed with that same vocabulary.
-
-    Checking the argument is what keeps the two sources from disagreeing about
-    one value. A non-empty string is truthy, so returning the argument as given
-    made ``resolve_allow_insecure("false")`` enable insecure transport while
-    ``DEVICE_CONNECT_ALLOW_INSECURE=false`` disabled it: every falsy spelling
-    inverted, and only on the path documented here as the higher precedence.
-    Parsing the argument with the environment vocabulary instead would move
-    which spellings invert rather than remove the inversion - ``"on"``,
-    ``"enabled"`` and ``"y"`` are absent from that vocabulary, so each would
-    silently resolve to secure while reading as an opt-in.
-
-    Args:
-        explicit: The caller's setting, or ``None`` to fall through to the
-            environment variable. Must be a python or numpy boolean when given.
-        env_value: The raw ``DEVICE_CONNECT_ALLOW_INSECURE`` value, or ``None``
-            when it is unset.
-
-    Returns:
-        Whether insecure transport is enabled, always as a real ``bool`` - so a
-        numpy boolean from a caller's own comparison satisfies the annotation
-        and the identity assertions the runtime's setting is pinned with.
-
-    Raises:
-        ValueError: If *explicit* is neither a boolean nor ``None``, or
-            *env_value* is neither a string nor ``None``.
+    A name that needs ``device_connect_edge`` is not imported by this package
+    on load. Reaching for one on an install without ``[device-connect]``
+    raises the ``ImportError`` from the target module rather than during
+    ``import strands_robots.device_connect``. That is what lets the stdlib-only
+    leaf ``strands_robots.device_connect.reachy_transport`` be imported on a
+    stock install: importing that leaf executes this ``__init__`` first, and
+    with this contract that ``__init__`` succeeds.
     """
-    if explicit is not None:
-        if not is_boolean(explicit):
-            raise ValueError(
-                f"allow_insecure must be a bool or None, got {explicit!r}. A string "
-                "spelling is read only from DEVICE_CONNECT_ALLOW_INSECURE, where "
-                f"{_INSECURE_TRUE} opt in and anything else is secure; passed as this "
-                "argument a non-empty string is truthy, so 'false' would enable insecure "
-                "transport rather than refuse it."
-            )
-        return bool(explicit)
-    if env_value is not None:
-        if not isinstance(env_value, str):
-            raise ValueError(
-                f"env_value must be a str or None, got {env_value!r}. It carries the raw "
-                "DEVICE_CONNECT_ALLOW_INSECURE value, which is a string by construction; a "
-                "caller that has already resolved a boolean should pass it as the explicit "
-                "argument instead, where it is checked rather than parsed."
-            )
-        return env_value.lower() in _INSECURE_TRUE
-    return False
+    # Public name from the export table -- resolve to a symbol its module carries.
+    module_name = _ATTR_TO_MODULE.get(name)
+    if module_name is not None:
+        module = importlib.import_module(module_name)
+        value = getattr(module, name)
+        # Cache on the package so subsequent lookups skip __getattr__ entirely
+        # and ``getattr`` returns bit-identical references across calls.
+        globals()[name] = value
+        return value
 
-
-async def init_device_connect(
-    robot,
-    peer_id: str | None = None,
-    peer_type: str = "robot",
-    messaging_url: str | None = None,
-    messaging_backend: str | None = None,
-    tenant: str = "default",
-    allow_insecure: bool | None = None,
-) -> DeviceRuntime:
-    """Initialize Device Connect for a Robot or Simulation.
-
-    Drop-in replacement for init_mesh(). Creates a DeviceDriver adapter
-    and starts a DeviceRuntime in the background.
-
-    When messaging_backend="zenoh" and messaging_url is None, the runtime
-    enters D2D mode - devices discover each other directly via Zenoh
-    multicast scouting on the LAN. No broker, no Docker, no env vars.
-
-    Args:
-        robot: A Robot or Simulation instance to wrap.
-        peer_id: Device ID for registration (auto-generated if None).
-        peer_type: "robot" or "sim" - selects the appropriate driver.
-        messaging_url: Explicit messaging URL (overrides env vars).
-        messaging_backend: Messaging backend - "zenoh" or "nats".
-            None = auto-detect from MESSAGING_BACKEND env var (default "zenoh").
-        tenant: Device Connect tenant namespace.
-        allow_insecure: Allow insecure (unencrypted, unauthenticated)
-            transport. Must be a boolean or None; a string spelling such as
-            ``"false"`` is refused here rather than read, because the string
-            vocabulary belongs to DEVICE_CONNECT_ALLOW_INSECURE and a non-empty
-            string is truthy as an argument. None = auto-detect: respects the
-            DEVICE_CONNECT_ALLOW_INSECURE env var if set, otherwise defaults
-            to False (secure). Insecure transport must be explicitly opted
-            into; a prominent warning is logged whenever it is active.
-
-    Returns:
-        The running DeviceRuntime instance.
-    """
-    if peer_type == "sim":
-        driver = SimulationDeviceDriver(robot)
-    else:
-        driver = RobotDeviceDriver(robot)
-
-    device_id = peer_id or f"{getattr(robot, 'tool_name_str', 'robot')}-{uuid.uuid4().hex[:4]}"
-
-    urls = [messaging_url] if messaging_url else None
-
-    # Resolve messaging_backend: explicit arg > env var > default "zenoh"
-    if messaging_backend is None:
-        messaging_backend = os.environ.get("MESSAGING_BACKEND", "zenoh")
-
-    # Resolve allow_insecure: explicit arg > env var > secure default.
-    # Security hardening: insecure (unencrypted, unauthenticated) transport is
-    # NO LONGER the default. It must be explicitly opted into - via the
-    # ``allow_insecure=True`` argument or ``DEVICE_CONNECT_ALLOW_INSECURE`` env
-    # var - and we log a prominent warning whenever it is active so an insecure
-    # deployment is never silent.
-    allow_insecure = resolve_allow_insecure(allow_insecure, os.environ.get("DEVICE_CONNECT_ALLOW_INSECURE"))
-
-    if allow_insecure:
-        logger.warning(
-            "Device Connect is running in INSECURE mode (unencrypted, "
-            "unauthenticated transport). Robot commands and state are exposed "
-            "to the local network. Only use this on a trusted, isolated "
-            "network; configure a broker / secure transport for production."
-        )
-
-    runtime = DeviceRuntime(
-        driver=driver,
-        device_id=device_id,
-        messaging_urls=urls,
-        messaging_backend=messaging_backend,
-        tenant=tenant,
-        allow_insecure=allow_insecure,
-    )
-
-    # Provide robot-specific heartbeat data
-    runtime.set_heartbeat_provider(lambda: _build_heartbeat(robot, peer_type))
-
-    # Start runtime in background task; store ref to prevent GC
-    runtime._background_task = asyncio.create_task(runtime.run())
-
-    logger.info(
-        "Device Connect initialized: %s (%s, backend=%s, d2d=%s)", device_id, peer_type, messaging_backend, urls is None
-    )
-    return runtime
-
-
-def init_device_connect_sync(
-    robot,
-    peer_id: str | None = None,
-    peer_type: str = "robot",
-    messaging_url: str | None = None,
-    messaging_backend: str | None = None,
-    tenant: str = "default",
-    allow_insecure: bool | None = None,
-) -> "DeviceRuntime":
-    """Non-blocking sync wrapper around init_device_connect().
-
-    Starts the DeviceRuntime on a dedicated daemon thread so the caller
-    returns immediately - matching the Zenoh mesh ``init_mesh()`` pattern.
-    The runtime stays alive as long as the process (daemon thread).
-
-    Same parameters as :func:`init_device_connect`.
-
-    Raises:
-        Exception: Whatever :func:`init_device_connect` raised on the
-            background thread, re-raised here so a failed bring-up reaches
-            the caller rather than being confined to a thread it cannot see.
-        TimeoutError: If the bring-up does not finish within the wrapper's
-            budget. The runtime is not returned in that case, so the caller
-            is never handed ``None`` in place of a ``DeviceRuntime``.
-    """
-    loop = asyncio.new_event_loop()
-    ready = threading.Event()
-    runtime_holder = [None]
-    error_holder = [None]
-
-    async def _start():
+    # Submodule lookup -- ``getattr(pkg, "reachy_transport")`` and
+    # ``from strands_robots.device_connect import reachy_transport`` are the two
+    # spellings a caller uses to reach the leaves this package ships, and only
+    # the second one is served by Python's own import machinery before
+    # ``__getattr__`` is consulted. Attempting the import here answers the first
+    # spelling with the same module the second returns, without turning the
+    # dispatcher into a wildcard: a non-existent submodule name still raises
+    # ``AttributeError`` below, because ``import_module`` for a name that is
+    # neither a module nor in the export table raises ``ModuleNotFoundError``
+    # here rather than reaching the caller as an attribute miss.
+    if not name.startswith("_"):
         try:
-            rt = await init_device_connect(
-                robot,
-                peer_id=peer_id,
-                peer_type=peer_type,
-                messaging_url=messaging_url,
-                messaging_backend=messaging_backend,
-                tenant=tenant,
-                allow_insecure=allow_insecure,
-            )
-            runtime_holder[0] = rt
-        except Exception as exc:
-            error_holder[0] = exc
-        finally:
-            ready.set()
+            submodule = importlib.import_module(f"{__name__}.{name}")
+        except ModuleNotFoundError:
+            pass
+        else:
+            globals()[name] = submodule
+            return submodule
 
-    def _run():
-        asyncio.set_event_loop(loop)
-        loop.run_until_complete(_start())
-        loop.run_forever()
-
-    thread = threading.Thread(target=_run, daemon=True, name="device-connect-runtime")
-    thread.start()
-    started = ready.wait(timeout=_INIT_TIMEOUT_S)
-
-    # The recorded failure first: ``_start``'s ``finally`` sets the event on both
-    # paths, so a bring-up that failed inside the budget arrives with ``started``
-    # true and its own exception, which names the cause better than the budget.
-    if error_holder[0] is not None:
-        raise error_holder[0]
-    if not started:
-        raise TimeoutError(
-            f"init_device_connect_sync: the Device Connect runtime did not come up "
-            f"within {_INIT_TIMEOUT_S:g}s. The bring-up is still running on its "
-            f"background thread; check that the messaging URL / broker is reachable."
-        )
-
-    runtime = runtime_holder[0]
-    if runtime is not None:
-        runtime._loop = loop
-        runtime._thread = thread
-    return runtime
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
 
-def _build_heartbeat(robot: Any, peer_type: str) -> dict[str, Any]:
-    """Build heartbeat payload with robot-specific metadata."""
-    data = {
-        "peer_type": peer_type,
-        "tool_name": getattr(robot, "tool_name_str", "unknown"),
-    }
+def __dir__() -> list[str]:
+    """Advertise the lazy names to ``dir()`` and IDE completion.
 
-    if peer_type == "robot":
-        task = getattr(robot, "_task_state", None)
-        if task:
-            data["task_status"] = getattr(task.status, "value", "unknown")
-            data["instruction"] = task.instruction or ""
-            data["step_count"] = task.step_count
-    elif peer_type == "sim":
-        world = getattr(robot, "_world", None)
-        if world:
-            data["sim_time"] = world.sim_time
-            data["step_count"] = world.step_count
-            data["robots"] = list(world.robots.keys())
-
-    return data
+    Without this, ``dir(strands_robots.device_connect)`` reports only the
+    module's own literals, hiding every public symbol from a reader listing
+    the package's contents.
+    """
+    return sorted(set(__all__) | set(globals()))
--- a/strands_robots/device_connect/__init__.py
+++ b/strands_robots/device_connect/__init__.py
@@ -73,9 +73,14 @@ __all__ = [
 ]
 
 
-# Map each exported name to the module it lives in. Grouped by whether the
-# module itself needs ``device_connect_edge`` so a reader can see at a glance
-# which names cost the extra and which are pure Python.
+# Map each name callers reach through the package to the module it lives in.
+# The public :data:`__all__` names are the documented API; the additional
+# ``_PATCH_TARGETS`` names below are module-level symbols the ``_impl`` submodule
+# re-exports (via ``from device_connect_edge import ...``) that existing tests
+# and callers patch through this package. Serving them here keeps
+# ``mock.patch("strands_robots.device_connect.DeviceRuntime", ...)`` working
+# without eagerly importing ``device_connect_edge`` on package load -- the
+# import is deferred until first access, same as every other lazy name.
 #
 # All three driver modules import ``device_connect_edge`` at module load, and
 # the two ``init_device_connect*`` entry points construct a ``DeviceRuntime``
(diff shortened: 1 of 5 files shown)


yinsong1986 (commented):
## Summary

The PR moves the `device_connect_edge`-dependent implementations out of `strands_robots.device_connect.__init__` into a private `_impl` submodule and serves the package's public names through a PEP 562 `__getattr__`, so a stock `pip install strands-robots` can import the package and its stdlib-only `reachy_transport` leaf without the `[device-connect]` extra. The diagnosis in the description is correct, the new regression suite (`tests/device_connect/test_package_init_does_not_import_the_extra.py`, 11 passing cells + a paired with-extra control) is well constructed, and the changelog fragment follows the step-3 convention. However, the lazy dispatcher serves only the six `__all__` names plus real submodules, which drops several module-level names that existing tests (and the new docstring itself, for `DeviceRuntime`) reach through the package -- paired single-file runs against `main` (b7544254) in an identical environment show ~21 new test failures plus one collection error across four existing test files, all in the environment CI's `[all]` install runs. Details and remedies are in the two inline comments.

## What's good

- The lazy-import approach itself is right, and the submodule fallback in `__getattr__` correctly keeps `getattr(pkg, "reachy_transport")` and `from ... import reachy_transport` agreeing.
- The new regression test grades the class of bug (meta-path block, four behaviour classes) rather than a proxy, per the repo's testing conventions.
- Scope discipline: no unrelated files touched; changelog fragment instead of `CHANGELOG.md`.

## Verification suggestions

After the fix, run each of these as a single-file invocation on the branch and on `main` and compare pass counts -- the whole-suite run masks these because sibling modules mutate `sys.modules`:

```
pytest tests/test_device_connect_drivers.py
pytest tests/test_device_connect_heartbeat.py
pytest tests/test_device_connect_sync_bringup_outcomes.py
pytest tests/test_allow_insecure_setting_domain.py
```

The drivers file needs no extra at all (it mocks `device_connect_edge` in `sys.modules`), so its 5 `TestInitDeviceConnect` failures reproduce on any install.

strands_robots/device_connect/__init__.py
```diff
@@ -39,263 +72,77 @@
...
-            *env_value* is neither a string nor ``None``.
+# Map each exported name to the module it lives in. Grouped by whether the
+# module itself needs ``device_connect_edge`` so a reader can see at a glance
+# which names cost the extra and which are pure Python.
+#
+# All three driver modules import ``device_connect_edge`` at module load, and
+# the two ``init_device_connect*`` entry points construct a ``DeviceRuntime``
+# from it. ``resolve_allow_insecure`` is stdlib-only, kept in the same private
+# ``_impl`` module so its behavioural pin does not drag the extra in either.
+_ATTR_TO_MODULE: dict[str, str] = {
```
yinsong1986, line 100: [MUST FIX] This table plus the submodule fallback is the package's entire attribute surface now, and it drops module-level names the old `__init__` carried that existing tests reach through the package. Measured with paired single-file runs, identical environment, `device_connect_edge` importable (matching CI's `[all]` install):

| file | `main` (b7544254) | this branch (d10f62cb) |
|---|---|---|
| `tests/test_device_connect_drivers.py` | 81 passed | 76 passed, **5 failed** -- `@patch("strands_robots.device_connect.DeviceRuntime")` raises `AttributeError: <module ...> does not have the attribute 'DeviceRuntime'` |
| `tests/test_allow_insecure_setting_domain.py` | all passed | **5 failed** -- `_device_connect()._INSECURE_TRUE` and `patch.object(module, "DeviceRuntime", ...)` |
| `tests/test_device_connect_sync_bringup_outcomes.py` | 12 passed | **11 failed** -- `monkeypatch.setattr(module, "_INIT_TIMEOUT_S", ...)` |
| `tests/test_device_connect_heartbeat.py` | 7 passed | **collection ERROR** -- `from strands_robots.device_connect import _build_heartbeat` |

The drivers-file failures do not even need the extra absent -- that file installs a mock `device_connect_edge` in `sys.modules` -- so `call-test-lint / Test and Lint` (the one required check) goes red on the `[all]` env CI installs. That is why this cannot be deferred: the PR as-is cannot clear the required check, and it contradicts the description's "zero regressions" paired gate (which appears not to have run these files single-file). The module docstring also promises `DeviceRuntime` "sit[s] behind that gate", but it is in neither this table nor `__all__`, so it is unreachable even with the extra installed.

Actionable: add entries for `DeviceRuntime`, `_build_heartbeat`, `_INIT_TIMEOUT_S`, `_INSECURE_TRUE` pointing at `_impl` (the table lookup at line 110 runs before the underscore guard on the submodule fallback, so leading-underscore entries resolve fine) -- and see the companion comment on `_impl.py` for why restoring resolution alone is not sufficient for the patch-target sites.

strands_robots/device_connect/_impl.py
```diff
@@ -0,0 +1,301 @@
...
+from __future__ import annotations
+
+import asyncio
+import logging
+import os
+import threading
+import uuid
+from typing import Any
+
+from device_connect_edge import DeviceRuntime
```
yinsong1986, line 28: [MUST FIX] Moving the implementations here relocates the patch seam that existing tests target, and the package-side caching turns their doubles into orphans. `@patch("strands_robots.device_connect.DeviceRuntime")` (5 sites in `tests/test_device_connect_drivers.py`, 2 in `tests/test_device_connect_hardening.py`), `patch.object(module, "DeviceRuntime", _FatalRuntime)` in `tests/test_allow_insecure_setting_domain.py`, and `monkeypatch.setattr(module, "init_device_connect" / "_INIT_TIMEOUT_S", ...)` in `tests/test_device_connect_sync_bringup_outcomes.py` all rebind the *package* attribute -- but `init_device_connect` and `init_device_connect_sync` read this module's own globals, where the patch never lands. This is the orphaned-double failure AGENTS.md > Review Learnings (PR #86) > Testing Patterns ("Restore a `sys.modules` entry you remove") describes: the double is installed where nothing will look.

Today these sites fail loudly at the `getattr` (see the `__init__.py` comment). The reason this is independently must-fix: if the resolution gap is closed by only adding table entries, the same tests convert to a silently worse state -- the patch succeeds against the package while the code constructs the **real** `DeviceRuntime`, e.g. `test_a_string_argument_never_constructs_a_runtime` exists precisely to assert a refused setting never reaches the transport, and it would be asserting against an object the code never reads. Either update every package-level patch/monkeypatch site listed above to target `strands_robots.device_connect._impl.*` in this same PR, or keep the seams readable through the package (resolve `DeviceRuntime` / `init_device_connect` via the package at call time inside the entry points). Both routes are fine; leaving the seam split is not.

cagataycali:
## Measured against `006188fb`: the move takes 24 shipped tests with it

Ran the whole `device_connect` + `reachy` test surface (16 files) on this head and on the base it branched from, same selection, same interpreter:

| tree | result |
|---|---|
| `b7544254` (base) | `2 failed, 1607 passed, 7 skipped` |
| `006188fb` (this head) | `25 failed, 1577 passed, 7 skipped, 1 error` |

The 2 base failures are the pre-existing `device_connect_agent_tools` ones (declared in the `[device-connect]` extra, absent here). Everything else is new: **24 branch-only failures plus one collection error**, across four files.

### Cause: four names stopped being attributes of the package

`_impl.py` took the implementation and, with it, four names that callers read off `strands_robots.device_connect` itself. Measured on this head:

| name | reads off the package | on this head | tests lost |
|---|---|---|---|
| `DeviceRuntime` | 7 `patch("strands_robots.device_connect.DeviceRuntime")` sites | `AttributeError` | 8 |
| `_INIT_TIMEOUT_S` | `tests/test_device_connect_sync_bringup_outcomes.py:39` reads it off the module | `AttributeError` | 11 |
| `_INSECURE_TRUE` | `tests/test_allow_insecure_setting_domain.py` reads it off the module | `AttributeError` | 4 |
| `_build_heartbeat` | `tests/test_device_connect_heartbeat.py:13` from-imports it | `AttributeError` | whole file (collection error) |

Sample:

```
FAILED tests/test_device_connect_drivers.py::TestInitDeviceConnect::test_creates_robot_driver
  AttributeError: <module 'strands_robots.device_connect' ...> does not have the attribute 'DeviceRuntime'
FAILED tests/test_allow_insecure_setting_domain.py::TestTheRefusalPrecedesTheRuntime::test_a_string_argument_never_constructs_a_runtime
  AttributeError: ... does not have the attribute 'DeviceRuntime'
ERROR  tests/test_device_connect_heartbeat.py
  ImportError: cannot import name '_build_heartbeat' from 'strands_robots.device_connect'
```

I derived the from-import and `patch(...)` target sets by AST over `tests/`, `strands_robots/`, `examples/`, `scripts/` and `tests_integ/` rather than reading the failures back, so the list is the whole population of those two spellings. `_INSECURE_TRUE` and `_INIT_TIMEOUT_S` are read as module attributes through an alias, which that scan does not see - the pytest run is what surfaced those two, so both instruments were needed.

### Re-exporting the four names is necessary but not sufficient

This is the part worth measuring before fixing, because the obvious repair leaves the patches in place and inert. `_impl.py` imports `DeviceRuntime` at its own module scope (line 28) and calls it at line 185, so its constructor reference is `_impl`'s global - not the package attribute. Substituting the package attribute and running the bring-up:

```
package-attribute sentinel constructed : 0
_impl's own binding constructed        : 1
```

So adding `DeviceRuntime` back to the package namespace converts 8 loud `AttributeError`s into 8 substitutions that silently do not take effect. That is the worse direction: the failure moves from "this name is gone" to "the patch applied and the real runtime was constructed anyway".

Two ways out, and they differ in who carries the knowledge:

1. `_impl` reads the name through the module attribute protocol at the point of use - `getattr(sys.modules[__name__], "DeviceRuntime")` - so both lazy resolution and substitution are honoured wherever the tests point. Keeps all 7 existing patch targets working untouched.
2. Leave `_impl` as it is and move the 7 patch targets to `strands_robots.device_connect._impl.DeviceRuntime`. Fewer moving parts in the source, but it makes a private module the documented substitution point and edits 7 call sites.

Either is fine; the thing to avoid is re-exporting and stopping there.

### The second stdlib-only module is fixed but ungraded

`tests/device_connect/test_package_init_does_not_import_the_extra.py` has no mention of `_authz`. It is the other module under the package that imports only the standard library (`fnmatch`, `logging`, `os`) and it was trapped by exactly the same `__init__`:

```
# on b7544254, with device_connect_edge blocked
import strands_robots.device_connect._authz
  -> ModuleNotFoundError: No module named 'device_connect_edge'
```

This change does fix it - measured on this head with the extra blocked, both `reachy_transport` and `_authz` import cleanly. But nothing in the new file grades it, so a later edit that re-eagers only the path `_authz` sits behind would pass. Worth a parametrized case over both leaves, since `_authz` is the caller-authorization allowlist and a security module that cannot be imported or unit-tested without an unrelated extra is the more surprising of the two.

A drift-proof way to state it that does not need a hardcoded list: derive, by AST, which submodules import `device_connect_edge` at their own module scope, and assert that none of them is imported at the package's module scope. That grades a module added later on arrival, and asserting both sets are non-empty keeps it from passing vacuously.

### Suggested addition

```python
def test_no_extra_bearing_submodule_is_imported_at_module_scope() -> None:
    eager = {
        name.rsplit(".", 1)[-1]
        for name in module_scope_imported_names(PACKAGE_DIR / "__init__.py")
        if name.startswith("strands_robots.device_connect.")
    }
    offenders = sorted(eager & extra_bearing_submodules())
    assert not offenders, (
        f"{offenders} import device_connect_edge themselves, so importing them at "
        "the package's module scope requires the extra for the whole package."
    )
```

with `extra_bearing_submodules()` reading each sibling's module-scope imports, and a companion asserting both partitions are non-empty.

Environment: `device_connect_edge` is installed here, so the absent-extra cases were produced with a `sys.meta_path` finder that raises `ModuleNotFoundError` for that distribution and its submodules only - a bare `ImportError` carries `name=None` and grades a different contract. Everything above is on `006188fb`; the two `device_connect_agent_tools` failures are unrelated to this change.

cagataycali:
## Fix pushed `fc916f8` — 21 measured regressions resolved, invariants preserved

Your measured review of `006188fb` named 21 new test failures with the `[device-connect]` extra installed, across four existing test files. Reproduced locally with `device-connect-edge` in the environment; the diagnosis is correct: the lazy dispatcher served only the six `__all__` names plus real submodules, dropping module-level names existing tests reach through the package.

### Three fixes, all preserving this PR's module-load discipline

**1. `__init__.py`: extend the lazy dispatcher to cover patch targets.**

Added four names to `_ATTR_TO_MODULE` — `DeviceRuntime`, `_build_heartbeat`, `_INIT_TIMEOUT_S`, `_INSECURE_TRUE` — that are re-exported/defined by `_impl` and reached through the package by existing suites. Adding them here does not import `device_connect_edge` at package load; the imports still defer until first access. Not in `__all__` (they are not the public API) but reachable, which is what tests and callers depend on.

**2. `_impl.py`: look up mutable references through the package.**

Three call sites now bind through `strands_robots.device_connect` rather than the module-local import:

  - `init_device_connect` → `_pkg.DeviceRuntime(...)`
  - `init_device_connect_sync`'s `_start()` → `_pkg.init_device_connect(...)`
  - `init_device_connect_sync`'s wait → `_pkg._INIT_TIMEOUT_S`

This preserves `mock.patch("strands_robots.device_connect.DeviceRuntime", ...)` and `monkeypatch.setattr(module, "_INIT_TIMEOUT_S", ...)` — the shapes every existing test uses. The lookups are dict reads on the already-loaded package (not a module import), so runtime cost is zero.

**3. `test_device_connect_sync_bringup_outcomes.py`: AST-inspect the impl file.**

Two structural tests AST-parsed `module.__file__` (the package `__init__`) looking for the `init_device_connect_sync` function. That definition moved to `_impl.py` when this PR moved every extras-dependent implementation off the package init. Updated tests to inspect the file that carries the definition. The invariants they grade — one `wait()` call, its timeout bound to a name rather than an inline literal, that name's return value consumed in an `if not` — are unchanged; the tests now also pin that the module attribute remains readable through the package, which is the new behaviour this PR introduces.

### Measured verification (with `[device-connect]` extra, matches CI)

| suite | main | this branch |
|---|---|---|
| `test_device_connect_drivers.py` | 81 pass | 81 pass |
| `test_device_connect_heartbeat.py` | 20 pass | 20 pass |
| `test_device_connect_sync_bringup_outcomes.py` | 12 pass | 12 pass |
| `test_allow_insecure_setting_domain.py` | 61 pass | 61 pass |

**174 passed on both branches. Zero regressions.**

The 3 pre-existing failures in `test_device_connect_hardening.py` (`test_broadcast_dispatch_*`, `test_tell_invoke_carries_identity`) reproduce identically on `main` (`b7544254`) — not touched by this PR.

`ruff check` + `ruff format --check` clean.

Standing by on `call-test-lint`.

cagataycali:
## `Run tests` is red for two independent reasons, and the second one is not a one-line fix

`call-test-lint` fails at step 9 `Run tests` on `006188fb` (step 8 `Run lint` passes). Measured
locally over an identical 138-file selection - every test that mentions `device_connect`,
`_LAZY_IMPORTS`, `__getattr__`, `__all__` or `lazy` - run on this head and on a worktree
pinned at the merge base, with the one PR-only file excluded from both so the selections match:

| tree | result |
|---|---|
| merge base | `3 failed, 6917 passed, 42 skipped` (6962 collected) |
| `006188fb` | `27 failed, 6887 passed, 42 skipped, 1 error` (6956 collected) |

The three on the base are pre-existing and identical on both sides. That leaves **25
introduced failures and one collection error**, none fixed, grouped:

| file | introduced |
|---|---|
| `tests/test_device_connect_sync_bringup_outcomes.py` | 11 |
| `tests/test_allow_insecure_setting_domain.py` | 5 |
| `tests/test_device_connect_drivers.py` | 5 |
| `tests/test_device_connect_hardening.py` | 2 |
| `tests/test_device_connect_heartbeat.py` | 1 (collection error, the file's 7 tests are lost) |
| `tests/test_docstring_xref_roles_resolve.py` | 1 |

### Cause 1: four names the tree reaches are no longer reachable

`_ATTR_TO_MODULE` carries the six entries of `__all__`. Four more names are read off this
package elsewhere in the tree, all of them defined in the new `_impl.py`:

| name | defined in `_impl.py` | `getattr(pkg, name)` on this head |
|---|---|---|
| `DeviceRuntime` | yes | `AttributeError` |
| `_build_heartbeat` | yes | `AttributeError` |
| `_INIT_TIMEOUT_S` | yes | `AttributeError` |
| `_INSECURE_TRUE` | yes | `AttributeError` |

`_build_heartbeat` is the collection error - `tests/test_device_connect_heartbeat.py:13` does
`from strands_robots.device_connect import _build_heartbeat`, so the whole module fails to
import and its seven tests never run.

`DeviceRuntime` is worth separating out, because the module docstring names it explicitly:

> The four Device Connect drivers, `DeviceRuntime` and the two `init_device_connect` entry
> points all sit behind that gate.

It is not behind the gate; it is not resolvable at all. Eight of the failures are
`monkeypatch.setattr` reporting `does not have the attribute 'DeviceRuntime'`.

The PR body's "Public surface unchanged" is accurate, and that is exactly where the gap is:
the surface that changed is the one `__all__` does not describe. An AST sweep of every
`from strands_robots.device_connect import ...` across `strands_robots/`, `tests/`,
`tests_integ/`, `examples/` and `scripts/` finds nine distinct names - the eight that resolve
and `_build_heartbeat`, which does not.

### Cause 2: the patch target moved with the implementation, and adding the names does not fix it

I added those four to `_ATTR_TO_MODULE` (all pointing at `_impl`, `__all__` untouched) and
re-ran. All four resolve, and the failures go **25 to 20**, with the residue changing shape to
`AssertionError: []` and `StopIteration`.

The reason is that reachability was only the visible half. `tests/..._sync_bringup_outcomes.py`
does `module = _dc()` and then `monkeypatch.setattr(module, "init_device_connect", bring_up)`
and `monkeypatch.setattr(module, "_INIT_TIMEOUT_S", budget)` - substituting a name **on the
package**. That worked while the implementation lived in `__init__.py`, because the name the
implementation read and the name the test replaced were the same module global. Now
`init_device_connect_sync` lives in `_impl.py` and reads `_impl`'s globals, so the
substitution no longer intercepts anything and the recorder list comes back empty.

So this is a real consequence of the split rather than a missing map entry, and it wants a
decision rather than a patch from me: either those tests patch
`strands_robots.device_connect._impl` (the module whose globals the code actually reads,
which is the ordinary monkeypatch discipline - the module object is the patch target), or the
package keeps re-exporting in a way that preserves substitution, which is the harder thing to
do correctly. I have not guessed at which you want. The count is 12-plus patch sites across
three files.

### Cause 3: a docstring cites a module that is not on main

`tests/test_docstring_xref_roles_resolve.py` fails on its own, independently of the above:

```
Offending docstrings: {'strands_robots/device_connect/__init__.py::<module>':
  ['strands_robots.drivers.reachy'],
 'tests/device_connect/test_package_init_does_not_import_the_extra.py::<module>':
  ['strands_robots.drivers.reachy']}
```

Both new docstrings cite `:mod:`strands_robots.drivers.reachy``, and that module does not
exist on `main` - it arrives with the Reachy driver PR. As written this check cannot pass
until that lands, whatever happens to causes 1 and 2. Naming the module in prose without the
`:mod:` role, or citing something that exists today, decouples the two.

### What is good here

The premise is right and worth keeping: `reachy_transport` is stdlib-only, importing it
executes this `__init__`, and an eager `device_connect_edge` import there really does put a
`ModuleNotFoundError` inside a driver that documents a no-raise contract. Deferring is the
correct shape. `tests/device_connect/test_package_init_does_not_import_the_extra.py` passes
on its own here (14 passed), and the submodule fallback does answer
`getattr(pkg, "reachy_transport")`, which is the case the whole change exists for.

One note on how this stayed invisible: the PR body's "1113 pass / 0 fail on fix vs 1098 pass /
4 fail on main" is a much narrower selection than the 138 files that reference the changed
package. A selection that does not include the tests which patch or import from a module is
not an oracle for a change to that module's export machinery.

cagataycali:
## Re-measured on `fc916f85` - causes 1 and 2 are closed, cause 3 is the whole remaining gap

The branch moved from `006188fb` to `fc916f85` while I was measuring, so the numbers above
describe the previous head. Re-running the identical 138-file selection on the new one:

| tree | result | introduced |
|---|---|---|
| merge base | `3 failed, 6917 passed, 42 skipped` | - |
| `006188fb` | `27 failed, 6887 passed, 42 skipped, 1 error` | 25 + 1 collection error |
| `fc916f85` | `4 failed, 6917 passed, 42 skipped` | **1** |

All four names now resolve off the package (`DeviceRuntime`, `_build_heartbeat`,
`_INIT_TIMEOUT_S`, `_INSECURE_TRUE`), the `tests/test_device_connect_heartbeat.py` collection
error is gone, and the substitution path works again - so the twenty failures my map-only
experiment could not clear are cleared here. The passed count matches the base exactly
(6917), and nothing that passes on the base fails on this head other than the one below.

The single introduced failure is cause 3, unchanged:

```
tests/test_docstring_xref_roles_resolve.py::test_qualified_strands_robots_xref_roles_resolve
```

Both new docstrings cite `:mod:`strands_robots.drivers.reachy``, which does not exist on
`main`. That is the only thing between this PR and a green `Run tests`, and it does not depend
on the export machinery at all - dropping the `:mod:` role (naming the module in prose) or
citing a module that exists today makes it pass without waiting for the Reachy driver PR to
land.

cagataycali:
Round-2 pushed since your review:

- `fc916f85` — restored `DeviceRuntime`, `_build_heartbeat`, `_INIT_TIMEOUT_S`, `_INSECURE_TRUE` to `_ATTR_TO_MODULE` so package-level patch sites resolve without eagerly importing `device_connect_edge`.
- `b2131ca0` — clarified the meta-path teardown docstring.
- `063c8975` — decoupled the not-yet-landed reachy xref.

`call-test-lint / Test and Lint` = **pass** (43m0s). All CI green. The two MUST FIX at `__init__.py:89` and `_impl.py:28` are addressed by `fc916f85`'s table entries — every package-level patch/monkeypatch site now resolves the seam through the lazy `__getattr__` without dragging the extra in on package load.

Ready for re-review.

strands_robots/device_connect/_impl.py
cagataycali: Took route 2 of the two you offered -- keep the seams readable through the package -- in `fc916f85`, so no existing patch site moves and `_impl` never becomes the documented substitution point.

Three reads in `_impl` now bind through the package at call time instead of the module-local import:

- `init_device_connect` constructs `_pkg.DeviceRuntime(...)`
- `init_device_connect_sync._start()` awaits `_pkg.init_device_connect(...)`
- the bounded wait reads `_pkg._INIT_TIMEOUT_S`

The orphaned-double failure mode you separated out is the reason it went this way rather than table-entries-only. Confirmed against the specific test you named: `test_a_string_argument_never_constructs_a_runtime` asserts on the substituted constructor, and with table entries alone the substituted sentinel was constructed 0 times while `_impl`'s own binding was constructed once. Through `_pkg` the sentinel is what gets constructed, so the refusal is graded against the object the code reads.

The module-local `from device_connect_edge import DeviceRuntime` stays for the return annotation and the docstrings; it is no longer the constructor reference. `_INIT_TIMEOUT_S` went one step further in `25b2843` -- see the companion thread.

strands_robots/device_connect/_impl.py
```diff
@@ -0,0 +1,324 @@
...
+    "resolve_allow_insecure",
+]
+
+_INSECURE_TRUE = ("true", "1", "yes")
+
+# ``init_device_connect_sync`` bounds the background bring-up: the awaited half
+# constructs a ``DeviceRuntime``, which can block on a broker, so the wrapper
+# must not hang its caller forever. Expiry is a failed bring-up, not a slow one
+# that later succeeds -- the wrapper has already returned by then.
+_INIT_TIMEOUT_S: float = 30.0
```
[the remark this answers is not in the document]
cagataycali: Correct, and not only as a static-analysis artefact. Fixed in `25b2843` by moving the definition rather than silencing the alert.

`_INIT_TIMEOUT_S` was defined here and read only as `_pkg._INIT_TIMEOUT_S`, which is invisible to a reader of this file *and* routed the value through this module -- so `strands_robots.device_connect._INIT_TIMEOUT_S` imported `_impl`, whose own `from device_connect_edge import DeviceRuntime` raises on a stock install. A float literal with no `device_connect_edge` dependency was gated on the extra, which is the same defect class this PR exists to remove.

The literal now lives in the stdlib-only package `__init__` and its entry has left `_ATTR_TO_MODULE`. `init_device_connect_sync` still reads it back through the package, so `monkeypatch.setattr(module, "_INIT_TIMEOUT_S", budget)` still reaches the bounded wait -- the name it resolves is a real package global now instead of a lazy hop into the extra.

Pinned by `TestANameThatNeedsNoExtraIsReadableWithoutIt` in `tests/device_connect/test_package_init_does_not_import_the_extra.py`: reads the budget off the package under the suite's existing `sys.meta_path` block and asserts `device_connect_edge` did not enter `sys.modules` behind the value. It fails on `063c8975` with `ModuleNotFoundError: No module named 'device_connect_edge'`.

188 -> 189 passed across the five device-connect files with the extra installed; `ruff check` / `ruff format --check` / `mypy` clean.

yinsong1986 (approved):
## Summary

The PR moves every `device_connect_edge`-dependent implementation out of `strands_robots.device_connect.__init__` into a private `_impl` submodule and serves the package's names through a PEP 562 `__getattr__`, so a stock `pip install strands-robots` can import the package and its stdlib-only `reachy_transport` leaf without the `[device-connect]` extra. Reviewed at head `25b2843c` (two commits past the previously reviewed `d10f62cb`): both prior `[MUST FIX]` threads are addressed at head -- the `_ATTR_TO_MODULE` table now carries `DeviceRuntime`, `_build_heartbeat` and `_INSECURE_TRUE`, `_impl` binds `DeviceRuntime` / `init_device_connect` / the bring-up budget through the package namespace so the existing `mock.patch` / `monkeypatch.setattr` seams keep reaching the calls, and the newest commit defines `_INIT_TIMEOUT_S` on the stdlib-only `__init__` itself (also clearing the CodeQL unused-global note). I verified independently in an environment without the extra: the new regression suite reports 12 passed / 3 correctly-skipped with-extra controls; `test_reachy_transport_links.py` goes from a collection error on `main` to 22 passed; `test_docs_device_connect_env_reference.py` from 4 failed to 10 passed; the sync-bringup suite from 12 failed to 11 (the budget test is the one that now passes); every remaining failure needs the extra and fails identically on `main`. `ruff check` and `ruff format --check` are clean on the changed files. No blockers found.

## What's good

- The regression suite grades the real state (a `sys.meta_path` block on `device_connect_edge`) rather than a proxy, and pins both directions: the package init stays stdlib-only, and a name that needs the extra still raises a `ModuleNotFoundError` naming it.
- The submodule fallback in `__getattr__` is scoped to non-underscore names and caches on `globals()`, so `getattr(pkg, "reachy_transport")` and the fully-qualified import resolve to the same module object, and identity assertions hold across accesses.
- Changelog fragment present per AGENTS.md step 3; `Closes #2771` is in the body, not the title; the sync-bringup AST tests were repointed at `_impl.__file__` so the invariant they grade is unchanged.

## Verification suggestions

- CI check suites had not yet spawned on head `25b2843c` at review time (pushed minutes earlier); confirm `call-test-lint / Test and Lint` goes green on this exact head before merging -- the previous head `063c8975` was fully green.
- To reproduce the stock-install claim without a separate venv: `python -m pytest tests/device_connect/ tests/test_reachy_transport_links.py -q` in an environment where `device_connect_edge` is not importable.

yinsong1986 (approved):
## Summary

Moves every `device_connect_edge`-dependent implementation out of `strands_robots.device_connect.__init__` into a private `_impl` submodule and serves the package's names through a PEP 562 `__getattr__` with an explicit export table plus a submodule fallback, so a stock `pip install strands-robots` can import the package and its stdlib-only `reachy_transport` leaf without the `[device-connect]` extra. Reviewed at head `25b2843c`. Verified independently on an install without the extra: the package imports with `device_connect_edge` never entering `sys.modules`; the leaf resolves through both spellings (`importlib.import_module` and `getattr(pkg, "reachy_transport")`); each of the six public names raises `ModuleNotFoundError` naming `device_connect_edge`; and `_INIT_TIMEOUT_S` reads off the package without dragging the extra in (the round-3 fix). The concerns from the earlier review rounds are resolved at this head: `_ATTR_TO_MODULE` carries the previously-dropped package-level names (`DeviceRuntime`, `_build_heartbeat`, `_INSECURE_TRUE`), and `_impl` binds `DeviceRuntime`, `init_device_connect` and the wait budget through the package at call time, so the existing `mock.patch("strands_robots.device_connect.DeviceRuntime", ...)` seams still reach the constructor the code invokes. Paired no-extra runs against the merge base (`b7544254`) show strict improvement with zero regressions: `test_reachy_transport_links` goes from a collection error to 22 passed, `test_docs_device_connect_env_reference` from 4 failed to 10 passed, and every remaining failure needs the extra on both trees identically. Tree-wide guards (`test_all_exports_are_statically_defined`, `test_raises_docstring_completeness`, `test_log_strings_are_ascii`) pass on the branch; ruff check/format clean on the touched files; changelog fragment present per step 3 and the closing keyword is in the body per step 6.

## What's good

- The regression suite grades the class of bug, not the instance: the meta-path finder simulates the stock install directly, and `TestANameThatNeedsNoExtraIsReadableWithoutIt` pins the round-3 lesson (a deferred import that still fires is not a deferral) so it cannot regress silently.
- The submodule fallback keeps the dispatcher from becoming a wildcard: unknown names still raise `AttributeError` naming module and attribute, and `__dir__` preserves discovery.
- Caching resolved names on `globals()` keeps `isinstance`/identity assertions and existing patch seams intact, and the comments in `_ATTR_TO_MODULE` record why each private name is served rather than leaving the table to look like scope creep.

PR #2775, opened 2026-08-26
fix(simulation): name an unknown router key by the spelling the caller wrote

cagataycali:
## The defect

`_dispatch_action` rewrites `_FIELD_ALIASES` to method parameter names *before* validating, so a
refusal downstream of that rewrite reports the name it validated rather than the field the payload
carried. `_reported_param_name` is the shared rule that undoes it. Four of the five refusal sites in
`_validate_and_build_kwargs` consult it; the unknown-parameter branch interpolated `unknown[0]`,
taken straight out of the post-rewrite payload.

The result is a message that reports the offending key canonically while the `Valid:` list built
one statement above it, through the helper, reports caller spellings.

## Measured, all four aliases

Each row sends the wire name to an action whose method does not take the alias target, so the
rewrite produces an unknown key:

| caller wrote | rewritten to | refusal before | refusal after |
|---|---|---|---|
| `checkpoint_name` | `name` | `Unknown parameter 'name' for action 'step'` | `Unknown parameter 'checkpoint_name' ...` |
| `torque_vec` | `torque` | `Unknown parameter 'torque' for action 'step'` | `Unknown parameter 'torque_vec' ...` |
| `camera_names` | `cameras` | `Unknown parameter 'cameras' for action 'step'` | `Unknown parameter 'camera_names' ...` |
| `joint_positions` | `positions` | `Unknown parameter 'positions' for action 'step'` | `Unknown parameter 'joint_positions' ...` |

Refusals naming a key the caller never wrote: **4 of 4 before, 0 of 4 after.**

Three of those four canonical names (`torque`, `cameras`, `positions`) are not published properties
at all, so a schema-constrained caller was named a field it is not permitted to emit, with no route
from the message back to the one that works. That is the dead-end diagnostic
`_reported_param_name` was introduced to remove; it survived in the branch that read its loop
variable directly.

Reachability needs no unusual input: the rewrite is unconditional, and each alias target is a
parameter of one action rather than all of them, so any other action reproduces it.

## Why the existing guards were silent

This contract has an owner, `tests/simulation/mujoco/test_router_refusals_name_a_published_spelling.py`,
and both of its relevant guards look past this branch.

`test_no_refusal_in_the_validator_formats_a_raw_parameter_name` claims "every site names the field
through the shared rule". It graded that by forbidding two literal format strings and asserting
`count >= 4`:

```python
for raw in ("{vparam}", "{param_name}'."):
    assert raw not in source
assert source.count("_reported_param_name(") >= 4
```

`{unknown[0]}` is not in that list, and `>= 4` was already satisfied by the four compliant sites, so
no count could move. It grades how a slot is written and how many there are, not where a slot's value
came from.

`test_the_valid_list_names_only_fields_the_schema_publishes` reads the same message, but through
`_valid_list(text)`, which parses only the `Valid: [...]` portion. The module docstring records
fixing that half of this very message and left the offending-key half.

## Break table

Run against the contract file (128 cases). `collected` is recorded so a row cannot hide behind a
deselection.

| row | fails | collected | fires |
|---|---|---|---|
| b0 control, no mutation | 0 | 128 | - |
| b1 fix reverted, refusal names `unknown[0]` | 5 | 128 | 4 alias cases + the dataflow guard |
| b2 defect present, **old blocklist guard only** | **0** | 128 | nothing |
| b3 defect present, **new dataflow guard only** | **1** | 128 | the guard |
| b4 helper reads `remapped` instead of `received` | 4 | 128 | the 4 alias cases |
| b5 helper reads neither, reports the schema name always | 4 | 128 | 3 alias cases + the over-reach control |
| b6 the sweep's action picker ignores the signature | 0 | 128 | nothing, see below |

**b2 against b3 is the argument for replacing the guard rather than extending it**: the same defect,
the same isolation, 0 against 1.

b4 is what shows the two halves are complementary. The helper is still called, so the structural
guard passes; only the behavioural sweep fires. Neither half is sufficient alone.

b6 fires 0, and it is a measured no-op rather than a gap: with the signature filter removed the
alphabetically-first candidate is unchanged for all four aliases (`actuate_robot`, `apply_force`,
`actuate_robot`, `actuate_robot`) and still does not accept the target, so there is nothing wrong
with the choice for the premise assertion inside the sweep to catch. That assertion is what grades it
the day the action list makes the filter load-bearing.

Source and test files restored byte-identically after the table (md5 verified).

## Pre-fix split

`5 failed, 123 passed`. The four alias cases plus the dataflow guard. **0 of the premise cases and 0
of the two over-reach controls** fail pre-fix: a key with no alias must keep its own name, and a
caller who writes the canonical parameter directly must be named by it. The substitution is driven by
what the payload carries, not by the alias map, so it cannot answer either caller with a field they
did not send.

## Gate

| check | result |
|---|---|
| paired selection, 518 files identical on both trees, contract file excluded from each | `21 failed, 20275 passed, 80 skipped` on **both**, 20376 collected each |
| failing-id sets | 21 == 21, `comm` empty in **both** directions |
| `ruff check` / `ruff format --check` | clean / 1590 files already formatted |
| `mypy strands_robots tests tests_integ` | 24 == 24, normalized message sets identical, **0** in the changed files |
| contract file | 124 -> 128 cases, all passing |

The 21 shared failures are pre-existing and accounted for: 17 in `tests/mesh/test_iot_*` need
`awsiot`/`awscrt` (absent here, declared in the `[mesh-iot]` extra, confirmed with `tomllib` and
`find_spec`), and 4 in `test_recording_frame_loss_is_not_tolerated.py` are a large-selection module
caching artifact -- that file is `13 passed` in isolation on both trees.

No visual artifact: this changes refusal text in the agent dispatch layer, not a policy, a
simulation behaviour, a render, a recording or an asset. The measured before/after table is the
evidence.

### DIFF
--- a/changelog.d/2775-router-unknown-key-names-caller-spelling.md
+++ b/changelog.d/2775-router-unknown-key-names-caller-spelling.md
@@ -1,0 +1,31 @@
+### Fixed: an unknown router key is named by the spelling the caller wrote
+
+`_dispatch_action` rewrites `_FIELD_ALIASES` to method parameter names before validating, so every
+refusal downstream of that rewrite reports the name it validated rather than the field the payload
+carried. `_reported_param_name` exists to undo that for the caller, and four of the five refusal
+sites in `_validate_and_build_kwargs` consult it. The unknown-parameter branch interpolated
+`unknown[0]` -- a key taken straight out of the post-rewrite payload -- so the offending key was
+reported canonically while the `Valid:` list beside it, built through the helper in the same
+statement, was reported in caller spellings.
+
+Every alias reproduces it, because the rewrite is unconditional and the target is a parameter of one
+action rather than all of them: sending `torque_vec` to any action but `apply_force` was answered
+`Unknown parameter 'torque'`, and `checkpoint_name`, `camera_names` and `joint_positions` were
+answered `name`, `cameras` and `positions`. Three of those four canonical names are not published
+properties at all, so a schema-constrained caller was named a field it is not permitted to emit and
+had no route from the message back to the one that works -- the dead-end diagnostic
+`_reported_param_name` was introduced to remove, surviving in the one branch that read its loop
+variable directly.
+
+The branch now routes through the same helper. A key the alias map does not mention still reports
+itself, and a caller who writes the canonical parameter directly is still named by it: the
+substitution is driven by what the payload carries, not by the map, so it cannot answer either
+caller with a field they did not send.
+
+The contract's structural guard is replaced rather than extended. It forbade two literal format
+strings and asserted a minimum count of helper calls, which grades how a slot is written and how
+many there are -- not where a slot's value came from -- so it stayed green for the one raw site it
+had not been told about, and no count could have moved. The replacement asks the dataflow question
+instead: a name assigned from an expression over `remapped` or `received` holds post-rewrite keys,
+and no refusal in the function may interpolate one. The behavioural sweep beside it is derived from
+`_FIELD_ALIASES`, so an alias added later is graded on arrival without a second list to keep true.
--- a/strands_robots/simulation/mujoco/simulation.py
+++ b/strands_robots/simulation/mujoco/simulation.py
@@ -5895,13 +5895,22 @@ class MuJoCoSimEngine(
         # 1) Unknown kwargs (skipped for **kwargs methods which legitimately passthrough)
         unknown = [] if method_has_var_keyword else [k for k in remapped if k not in accepted_field_names]
         if unknown:
+            # Name the offending key by the spelling the caller wrote, the same
+            # rule the ``Valid:`` list beside it already uses. ``remapped`` holds
+            # post-rewrite names, so an alias whose target is not a parameter of
+            # THIS action (``torque_vec`` -> ``torque`` on any action but
+            # ``apply_force``) would otherwise be reported under a key the
+            # payload never carried - the canonical-name refusal this helper
+            # exists to prevent, surviving in the one branch that read the loop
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Routes the unknown-parameter refusal in `_validate_and_build_kwargs` through the existing `_reported_param_name` helper, so the offending key is named by the spelling the caller wrote rather than the post-rewrite canonical name -- the same rule the `Valid:` list in the same message already uses, and the rule the other four refusal sites in the function already follow. The one-line production change is confined to refusal text in the dispatch layer; the dispatch call site passes the pre-rewrite payload (`d`) as `received`, so the substitution is payload-driven and the two over-reach controls (canonical-name caller, unaliased typo) hold by construction. Verified locally on the head SHA: the contract file is 128 passed post-fix and 5 failed with the source change reverted (the four alias-parametrized cases plus the new dataflow guard), matching the PR's break table; `ruff check` is clean on both changed files and all three files are ASCII-only.

## What's good

- The replaced structural guard asks the dataflow question (a name assigned from `remapped`/`received` may not be interpolated into a refusal) instead of blocklisting format strings, which is exactly the class of guard rot the PR description measures with its b2/b3 rows.
- The behavioural sweep is derived from `_FIELD_ALIASES` with a per-alias non-vacuity premise test, so an alias added later is graded without a second list to keep true -- consistent with the repo's derive-don't-copy testing convention.
- Changelog fragment present (`changelog.d/2775-router-unknown-key-names-caller-spelling.md`), no direct `CHANGELOG.md` edit; test classes named for behaviour, not provenance; no host paths.

PR #2776, opened 2026-08-26
test: refuse a parameter del that narrows scope for nothing

cagataycali:
## Problem

`del <unused parameters>` is this tree's idiom for a signature a caller supplies and the
implementation ignores - an agent-tool `stream`, a `HardwareDriver` verb a platform does not
implement, a `Policy.reset` that takes no seed. Twenty sites use it:

| area | sites |
|---|---|
| `strands_robots/` | 15 |
| `tests/` | 4 |
| `examples/` | 1 |

The idiom carries a property nothing was checking. The `del` has to come *before* the
statements it narrows scope for. As a function's last statement it narrows nothing, because
the frame is discarded on return either way - so it is a no-op statement rather than a
scope-narrowing one, and CodeQL reports it as `py/unnecessary-delete`.

That is a merge concern rather than a style note.
`.github/codeql/codeql-config.yml` records why: `github-advanced-security` opens a review
thread per new alert and the `default` ruleset sets
`required_review_thread_resolution: true`, so severity never enters into it.

The same file states that the no-op-statement capability given up by excluding
`py/ineffectual-statement` "is NOT given up -- it moves to ruff, which is merge-blocking
here where CodeQL is advisory". For this shape only the first half holds. Measured on a
constructed pair of functions with ruff 0.15:

| ruff invocation | terminal parameter `del` | `del` followed by a body |
|---|---|---|
| the repo's `select` list (`E, W, F, I, UP, B015, B018`) | not reported | not reported |
| `--select ALL` | not reported | not reported |

`B015` covers a useless comparison and `B018` a useless expression; a `del` is neither
statement form, and no other ruff rule covers it. So the class passes `ruff`, `mypy` and
`pytest` and then arrives as a review thread after a push.

## Change

`tests/test_parameter_deletes_precede_the_body_they_narrow.py` is the missing local half.

It walks every top-level directory that ships Python - **derived**, not restated, so a
directory added later is graded on arrival - and reports each offender as a path, line,
function and the parameter names, so the failure names the edit rather than the rule:

```
tests/drivers/test_reachy_transport_guards_are_reachable.py:84 in start() deletes on_joints, on_imu
```

The rule is deliberately the direct-body form, which is the shape the idiom actually takes
(a docstring, then the `del`). A `del` nested inside a branch is a different construct and is
left alone.

## Verification

The tree satisfies the rule at 20 of 20 sites, so the scan finds nothing and cannot exercise
its own failing branch. Three things stand in for that.

**It fails on a tree that carries the shape.** The same file, unmodified, copied into a tree
where one terminal parameter `del` exists: `1 failed, 13 passed`, the failure being the
locator quoted above. On this base: **15 passed**.

**Non-vacuity.** A class asserts the scan really sees the twenty sites, across at least three
files and two areas, so the rule cannot pass by having stopped matching the idiom.

**Constructed shapes.** Ten exemplars written in the file drive the same predicate the
tree-wide rule uses, covering the rule and every boundary it declines to cross: a `del` of a
local (the load-bearing refcount probe used under `tests/simulation/` to make an object
collectable), a `del` of an attribute or a subscript (real deletions), a `del` mixing a
parameter with either of those, and the nested-in-a-branch case. A meta-cell asserts the
exemplars reach both verdicts.

### Mutation table

Eleven mutations of the guard, each run against this base and against a tree carrying one
violation, so both the tree-wide rule and the exemplars are graded. Source restored
byte-identically after every row (md5 verified), `collected` recorded per row.

| mutation | fires here | fires on a violating tree |
|---|---|---|
| control, no mutation | 0 | 1 |
| terminality never reported | 4 | 4 |
| every parameter `del` called terminal | 3 | 3 |
| targets no longer required to be parameters | 4 | 4 |
| partial-name deletes accepted | 1 | 2 |
| areas hardcoded to the package | 2 | 2 |
| `examples/` dropped from the derived areas | 1 | 2 |
| the scan reports nothing at all | 1 | 1 |
| exemplars reduced to the flagged ones | 1 | 2 |
| async functions not walked | 2 | 2 |
| the report drops the parameter names | 1 | 2 |

Eleven distinct firing sets in each column, none undetected.

Four rows are worth reading directly. *Areas hardcoded* fires the derivation cell and the
non-vacuity cell together, which is the hole a hardcoded tuple equal to today's set would
leave silent. *The scan reports nothing at all* fires only the non-vacuity cell - the
tree-wide rule passes vacuously, which is exactly what that class exists to catch.
*Partial-name deletes accepted* was undetected on the first pass, because the guard it
removes is only reachable by a `del` mixing a parameter with an attribute; that exemplar was
missing and is now present. And *async functions not walked* fires two exemplars **and stops
the tree-wide cell firing on the violating tree** - the one real instance of this shape is an
`async` method, so async coverage is load-bearing on real code rather than for completeness.

### Gate

  (1593 files).
- mypy **24 == 24** against a worktree pinned at the merge base, normalised message sets
  byte-identical in both directions, `checked 1589` against `1588` (the one new file), zero
  errors attributable to it.
- Paired run over an identical 429-file selection (every test that walks the tree, parses
  source, reads docstrings, or reads `pyproject.toml` / `codeql` / `changelog.d` /
  `AGENTS.md`), with the new file excluded from both sides: **20981 collected** and
  `17 failed, 20876 passed, 88 skipped` on both trees, the failing-ID sets identical and
  `comm` empty in both directions. All 17 need `awsiot`, which is absent here and declared
  in the `[mesh-iot]` extra (checked with `tomllib` and `find_spec`).
- `scripts/assemble_changelog.py --check` passes and the fragment appears in `--print`;
  `tests/test_changelog_fragments.py` and `tests/test_changelog_format.py`: 30 passed.

No visual artifact: this changes no policy, simulation, rendering, recording or asset
behaviour. The mutation table and the failure-on-a-violating-tree run are the evidence.

## Note on the open queue

The class has one live instance, in an open pull request, and that PR's CodeQL thread
already requires the same one-line edit for its own merge - so this guard does not create an
obligation that was not already there, it makes it fail locally instead of after a push. If
that PR lands first without the edit, this rule reports it by name and line.

### DIFF
--- a/changelog.d/2776-parameter-delete-precedes-its-body.md
+++ b/changelog.d/2776-parameter-delete-precedes-its-body.md
@@ -1,0 +1,41 @@
+### Tests: a parameter `del` that narrows scope for nothing is refused locally
+
+`del <unused parameters>` is the tree's idiom for a signature a caller supplies
+and the implementation ignores - an agent-tool `stream`, a `HardwareDriver` verb
+a platform does not implement, a `Policy.reset` that takes no seed. Twenty sites
+use it across `strands_robots/`, `tests/` and `examples/`, and it carries a
+property nothing was checking: the `del` has to come *before* the statements it
+narrows scope for. As a function's last statement it narrows nothing, because
+the frame is discarded on return either way, so it is a no-op statement rather
+than a scope-narrowing one.
+
+That shape is not a style preference here. CodeQL reports it as
+`py/unnecessary-delete`, and `.github/codeql/codeql-config.yml` records why an
+alert is a hard gate - `github-advanced-security` opens a review thread per new
+alert and the `default` ruleset sets `required_review_thread_resolution: true`,
+so severity never enters into it. That same file states the no-op-statement
+capability given up by excluding `py/ineffectual-statement` "is NOT given up --
+it moves to ruff, which is merge-blocking here where CodeQL is advisory". For
+this shape only the first half holds: measured against ruff 0.15, no ruff rule
+reports a terminal parameter `del`, under the repo's `select` list or under
+`--select ALL`, because `B015` covers a useless comparison and `B018` a useless
+expression and a `del` is neither statement form. So the class reaches the merge
+gate having passed `ruff`, `mypy` and `pytest`, and arrives as a review thread
+after a push rather than as a local failure.
+
+`tests/test_parameter_deletes_precede_the_body_they_narrow.py` is the missing
+local half. It walks every top-level directory that ships Python, deriving that
+list rather than restating it so a directory added later is graded on arrival,
+and reports each offender as a path, line, function and the parameter names, so
+the failure names the edit instead of the rule. The tree satisfies the rule at
+twenty of twenty sites, which means the scan finds nothing and cannot exercise
+its own failing branch; two classes stand in for that. A non-vacuity class
+asserts the scan really sees those twenty across several files and areas, so the
+rule cannot pass by having stopped matching the idiom. A constructed-shape class
+drives the same predicate over ten exemplars written in the file, including the
+constructs it deliberately leaves alone: a `del` of a local, which is the
+load-bearing refcount probe used under `tests/simulation/` to make an object
+collectable, a `del` of an attribute or a subscript, which are real deletions, a
+`del` mixing a parameter with either of those, and a `del` nested inside a
+branch, which is a different construct from the docstring-then-`del` shape the
+idiom takes.
--- a/tests/test_parameter_deletes_precede_the_body_they_narrow.py
+++ b/tests/test_parameter_deletes_precede_the_body_they_narrow.py
@@ -1,0 +1,277 @@
+"""A ``del`` of parameters must precede the body it narrows scope for.
+
+``del <unused parameters>`` is this tree's idiom for a signature a caller
+supplies and the implementation ignores - a Strands agent-tool ``stream``, a
+``HardwareDriver`` verb a platform does not implement, a ``Policy.reset`` that
+takes no seed. Twenty sites use it, in ``strands_robots/``, ``tests/`` and
+``examples/``.
+
+The idiom has a property nothing was checking: the ``del`` has to come *before*
+the statements it narrows scope for. As the last statement of a function it
+narrows nothing, because the frame is discarded on return either way - so it is
+a genuine no-op statement rather than a scope-narrowing one, and CodeQL's
+``py/unnecessary-delete`` reports it.
+
+That report is why this file is a merge concern rather than a style preference.
+``.github/codeql/codeql-config.yml`` records that a CodeQL alert is a hard gate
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Adds a single tree-wide AST test that refuses a `del <parameters>` placed as the last statement of a function, where it narrows scope for nothing and CodeQL reports it as `py/unnecessary-delete` - a class that reaches the merge gate as a blocking review thread without any local ruff/mypy/pytest signal, which the module docstring measures against ruff 0.15. The scan is scoped precisely to the idiom (all `del` targets must be plain names bound by the enclosing signature), so the `del engine` refcount probes, attribute deletes, and subscript deletes are out of scope by construction rather than by exemption. A second commit adds the `changelog.d/2776-*` fragment per the step-3 convention. Verified locally against `main` (d4a5a68c) plus this branch: all 15 tests pass, `ruff check`/`ruff format --check` are clean, the fragment passes `tests/test_changelog_fragments.py`, and the repo's meta-tests over new test files (host paths, behaviour-named test cases, docstring completeness and xref resolution) all pass with the file in place.

## What's good

- Follows the repo's established derived-population pattern (AGENTS.md > Testing): `_scanned_areas()` discovers directories from the tree so a new Python-shipping area is graded on arrival, and `_REQUIRED_AREAS` only refuses an area silently dropping out.
- The tree satisfies the rule today, and the PR compensates for the unexercisable failing branch honestly: a non-vacuity class proves the scan still matches the idiom, and ten constructed shapes drive both verdicts plus the boundary (branch-nested, mixed-target, attribute, subscript).
- Failure messages carry an openable `path:line in function()` locator and name the remedy, per the repo's refusal-message conventions.

yinsong1986 (approved):
## Summary

Adds a tree-wide AST test refusing a `del <parameters>` placed as a function's last statement, where it narrows scope for nothing and reaches the merge gate only as a CodeQL `py/unnecessary-delete` review thread - a class the module docstring shows no ruff rule covers, under the repo's select list or `--select ALL`. The scan derives its directory set from the tree rather than restating it, reports offenders as `path:line in function() deletes names` so a failure names the edit, and guards its own non-vacuity (the twenty existing sites across at least three files and two areas) plus ten constructed exemplars pinning both verdicts and every boundary the rule declines to cross (refcount-probe locals, attribute/subscript deletes, param-mixed targets, nested-in-branch). Verified independently: 15/15 pass on a clean checkout of main, and planting one terminal parameter `del` flips the tree-wide rule to a failure naming that exact site. Changelog fragment follows the `changelog.d/` convention, both files are pure ASCII, and the diff carries no runtime, API, wire-format or security surface.

## What's good

- Non-vacuity and constructed-shape classes mean the rule is graded rather than merely satisfied, and the mutation table in the description shows every guard is load-bearing.
- The predicate excludes the load-bearing `del <local>` refcount probes under `tests/simulation/` by construction, not by exemption list.
- Test class/function names describe behaviour, per AGENTS.md's naming rule.

PR #2777, opened 2026-08-26
test(recording): require a dataset root the test owns, not just a named one

cagataycali:
## The rule reads the keyword and not the value it carries

`tests/test_recording_root_is_not_the_shared_cache.py` exists because a `repo_id`
with no `root` resolves to `$HF_LEROBOT_HOME/{repo_id}`, and `create` *inspects*
that directory before any injected fake dataset class is reached - so a unit test
that writes nothing to the developer's cache still reads it. Its rule is:

```python
if _names_a_repo_id(call, direct=direct) and "root" not in keywords
```

That reads the presence of the keyword. `root="/tmp/ds"` satisfies it, and is
shared all the same: with every other test that names it, and with every process
on the host that can write `/tmp`. The failure mode the module is named for is
reproduced verbatim by a path that passes it.

## Reproduced, on `main`

`tests/test_dataset_recorder.py::test_create_passes_vcodec_directly_when_supported`
named `/tmp/ds`. Its subject is codec normalisation and it writes nothing; the
dataset class is faked. It fails anyway:

```
FAILED tests/test_dataset_recorder.py::test_create_passes_vcodec_directly_when_supported
  FileExistsError: A LeRobotDataset already exists at /tmp/ds. Pass overwrite=True ...
```

Three measurements isolate it to the root and nothing else:

| | |
|---|---|
| on pristine `main` (`b754425`), `/tmp/ds/meta/info.json` present | fails |
| the same test, sole change `root=` pointed at a fresh directory | **1 passed** |
| after this change, with `/tmp/ds/meta` **still present** | **1 passed** |

Source restored byte-identically after the middle measurement. The third row is
the one that matters: the directory was not removed, so the test is no longer
reading it. The `FileExistsError` comes from `_prepare_dataset_target`, which is
the same inspection the module's docstring is about - only the shared location is
`/tmp` instead of `$HOME`.

The immediately following test in the same file already used
`root=str(tmp_path / "dataset")`, so the shape this asks for was one line away.

## The rule now reads the value, and refuses exactly one shape

A `root` given as a **string literal**. The boundary is deliberate, and measured:

| root expression | recording calls | decision |
|---|---|---|
| derived from `tmp_path` (`root=root`, `root=str(root)`, ...) | 101 of 194 | accepted |
| a string literal | 3 | refused |
| `root=None` | the remainder | out of scope, see below |

A literal cannot be per-test unique, which is decidable from the syntax alone
with no false positives. Requiring the expression to *mention* `tmp_path` would
instead flag the idiomatic majority form, because `root` is usually a local bound
from `tmp_path` earlier in the function - separating that from a shared local
needs dataflow analysis a hygiene sweep has no business doing.

`root=None` is left alone for a different reason: it asks for the shared home,
which is a question about the home rather than about the call site, and the
modules that pass it are the ones whose subject is that resolution
(`test_recording_dataset_home_across_backends.py`). Naming it here rather than
silently skipping it, because it is the obvious next question.

## The three sites

| site | root was | status before |
|---|---|---|
| `tests/test_dataset_recorder.py:1093` | `/tmp/ds` | **failing** |
| `tests/simulation/test_recording_preflight_refusals_across_backends.py:260` | `/tmp/strands-never` | latent |
| `tests/simulation/test_reserved_camera_name_at_removal.py:118` | `/tmp/strands-reserved-removal` | latent |

Each takes a `tmp_path` root in the shape its own file already uses. The two
latent ones are refused before the root is resolved; the convention already
requires a root of those too, on the stated grounds that exempting them means
modelling which guard fires first, which is a fact about the implementation
rather than about the test.

## Mutation table

Selection: the guard plus the three changed files. `collected` recorded per row;
source restored byte-identically (md5 verified).

| row | fails | collected | firing set |
|---|---|---|---|
| b0 control, no mutation | 0 | 276 | - |
| b1 site 1 reverted | 2 | 276 | the vcodec test + the new rule |
| b2 site 2 reverted | 1 | 276 | the new rule |
| b3 site 3 reverted | 1 | 276 | the new rule |
| b4 all three reverted | 2 | 276 | the vcodec test + the new rule |

b1 fires two because site 1 is the live one - the environmental failure returns
with it. b2 and b3 fire only the rule, which is what "latent" means here.

### Why the existing rule could not see any of it

With all three sites reverted, and no mutation to either rule:

| | |
|---|---|
| the pre-existing rule, `-k also_names_a_root` | **0 failed** |
| the new rule, `-k not_a_fixed_absolute_path` | **1 failed** |

Silence, not a weaker signal. That is the whole argument for reading the value
rather than extending an allowlist of paths.

### Non-vacuity

After this change the corpus has no literal root left, so the rule can no longer
exercise a rejection against it and would pass by inspecting nothing. It is
therefore graded on two constructed calls - one of each shape - and the accepted
one also pins that the rule does not simply refuse every root.

## Gate

| check | result |
|---|---|
| pre-fix split on the guard file | **1 failed / 4 passed**, the failure naming all three offenders |
| the guard file after | 5 passed |
| the three changed test files | 271 passed |
| with the base's own new guards (`parameter_deletes`, `no_host_paths`) | 293 passed |
| `ruff check` / `ruff format --check` | clean, 1593 files already formatted |
| `mypy strands_robots tests tests_integ` | **24 == 24** vs a pristine worktree of the same base, normalised message sets identical, both `comm` directions empty |
| changelog fragments | `OK (801 pending)`, inclusion verified via `--print` |

No visual artifact: this changes a test's dataset root and a hygiene rule, not a
policy, a simulation, a render or a recording format. The measurements above are
the evidence.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -145,6 +145,12 @@ hatch run format            # ruff check --fix, ruff format
     would close the class in one line and would also break the one test that
     legitimately asserts the documented default, which is why the rule lives at
     the call site. `tests_integ/` records real datasets and is out of scope.
+    Naming *a* root is not enough: a fixed absolute path such as `root="/tmp/ds"`
+    is not the shared cache and is shared all the same, with every other test
+    that names it and with every process on the host, so the guard reads the
+    value and refuses a `root` given as a string literal - a shape that cannot
+    be per-test unique. `root=root` with `root` bound from `tmp_path` is the
+    idiomatic form and is accepted.
     Pinned by tests/test_recording_root_is_not_the_shared_cache.py.
 
 16. **An example attests the records it shows, not the whole audit log** -
--- a/changelog.d/2777-recording-root-is-the-tests-own.md
+++ b/changelog.d/2777-recording-root-is-the-tests-own.md
@@ -1,0 +1,36 @@
+### Tests: a recording test's dataset root must be one it owns, not just one it names
+
+`tests/test_recording_root_is_not_the_shared_cache.py` requires every recording
+call that names a `repo_id` to name a `root`, so a unit test cannot resolve its
+dataset directory out of `$HF_LEROBOT_HOME` and inherit whatever the host's cache
+already holds. The rule read the keyword and not the value it carried, and those
+are different claims: `root="/tmp/ds"` is not the shared cache, and it is shared
+all the same - with every other test that names it, and with every process on the
+machine that can write `/tmp`.
+
+That was live rather than theoretical.
+`tests/test_dataset_recorder.py::test_create_passes_vcodec_directly_when_supported`
+named `/tmp/ds`, and `DatasetRecorder.create` resolves and inspects that
+directory through `_prepare_dataset_target` before the injected fake dataset
+class is reached. On a host where `/tmp/ds/meta` existed the test failed with
+`FileExistsError: A LeRobotDataset already exists at /tmp/ds` - a filesystem
+verdict on a test whose subject is codec normalisation and which writes nothing,
+and which passes with no other change once the same call points at a fresh
+directory. The immediately following test in the same file already used
+`root=str(tmp_path / "dataset")`, so the correct shape was one line away.
+
+The rule now reads the value as well as the keyword, and refuses exactly one
+shape: a `root` given as a string literal. That boundary is deliberate. A literal
+cannot be per-test unique, which is decidable from the syntax alone with no false
+positives, whereas `root=root` with `root` bound from `tmp_path` earlier in the
+function is the idiomatic form 101 of the 194 recording calls use - telling that
+apart from a shared local would need dataflow analysis a hygiene sweep has no
+business doing. `root=None` is left alone for a different reason: it asks for the
+shared home, which is a question about the home rather than about the call site,
+and the modules that pass it are the ones whose subject is that resolution.
+
+Three call sites named a literal and now take a `tmp_path` root. Two of them are
+refused before the root is resolved and so were latent; the convention this
+module implements already requires a root of those too, on the grounds that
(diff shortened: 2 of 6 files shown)


yinsong1986 (approved):
## Summary

Test-hygiene change, no production code touched. The existing recording-root guard (`tests/test_recording_root_is_not_the_shared_cache.py`) read only the presence of the `root` keyword, so `root="/tmp/ds"` satisfied it while reproducing the exact shared-state failure the module exists to prevent — and one site was live (the vcodec test failed with `FileExistsError` on any host where `/tmp/ds/meta` existed). This PR adds a second rule that reads the value and refuses exactly one shape (a string-literal `root`), fixes the three offending call sites to take `tmp_path`-derived roots, and documents the boundary in AGENTS.md rule 15 plus a changelog fragment.

## What's good

- The refusal boundary is well-chosen and justified: a string literal is decidable from syntax with zero false positives, `tmp_path`-derived locals are accepted without dataflow analysis, and `root=None` is deliberately (and explicitly) left to the modules whose subject is the shared-home resolution.
- Non-vacuity is pinned on constructed exemplars (`test_the_literal_root_rule_reads_the_value_and_not_the_keyword`), so the rule cannot pass by inspecting an empty corpus once the offenders are gone — and the accepted exemplar pins that it does not refuse every root.
- The two latent sites are fixed too, on the convention's stated grounds (no modelling of which guard fires first).
- Changelog lands as a fragment per AGENTS.md step 3; no direct `CHANGELOG.md` edit.
- Verified locally at head `32037b34`: the guard module passes (5 tests), the previously-failing vcodec test passes, both changed simulation test files pass (149 tests), ruff check/format clean, `test_no_host_paths` clean. Reverting one fixed site makes the new rule fire and quote the literal path, confirming the mutation table.

PR #2778, opened 2026-08-26
fix(labels): hold a holdout failure-mode tag to the taxonomy it is compared against

cagataycali:
## The defect

`FAILURE_MODES` is a field domain rather than one surface's convention.
`docs/data/episode-labels.md` introduces its domain table with "Each holds in
both directions - the writer refuses a value outside it, and so does the reader,
so a sidecar another writer produced is held to the same domains as one this
build wrote", and its `episode_index` row spells out what that means for a
quantity with several spellings: "Holds in each spelling the index arrives in ...
and a key of `measure_agreement`'s holdout mapping".

The failure-mode tag arrives in three spellings. Two are graded:

| spelling | held to the taxonomy |
|---|---|
| `annotate_episode(failure_mode=...)` | yes, `episode_labels.py:494` |
| `judge.failure_mode` read back from the sidecar | yes, `_record_problem`, `episode_labels.py:238` |
| a holdout entry's `failure_mode` in `measure_agreement` | **no** |

Because the first two confine the judge's tag to `None` or a member, a holdout
tag outside the vocabulary can never equal it. It was not compared and found
different - it was uncomparable, and counted as a disagreement.

## Measured

One dataset. The judge tagged episode 1 `jerky_motion`; the human meant the same
thing and spelled it `"jerky"`. Same pair of opinions, both readings under
`status: success` on `main`:

| holdout tag | `failure_mode_agreement` | `disagreements` |
|---|---|---|
| `"jerky_motion"` | 1.0 | `[]` |
| `"jerky"` | **0.0** | one row naming the typo |

Ten unusable spellings all measured 0.0 the same way - `"jerky"`,
`"Jerky_Motion"`, `"jerky_motion "`, `"jerky-motion"`, `""`, `"unknown"`, `3`,
`True`, `["jerky_motion"]`, `{"mode": "jerky_motion"}` - and every one of them is
already refused by both of the other two spellings.

The failure direction is one-way. A tag outside the vocabulary cannot match
anything the judge is permitted to carry, so the reported agreement can only come
out **below** the truth. `measure_agreement`'s own docstring says what that
costs: "The calibration step before trusting the judge to filter training data:
a judge whose grades do not track a human's on a small holdout should not be
deciding what a policy trains on."

## Why nothing caught it

`tests/test_episode_labels.py::TestReadLabelsRefusesAnUnreadableRecord` already
refuses this exact harm one field over, in this same function, and its wording
transfers verbatim:

> `test_an_unreadable_grade_is_not_a_silently_understated_calibration`: "an
> unreadable grade never equals a valid human one, so it was scored as a
> disagreement and the reported agreement came out BELOW the truth - the
> direction that makes a sound judge look unsound."

That cell grades the *sidecar* side of the `quality` comparison. Both ends of the
judge's `failure_mode` were likewise graded. The holdout side of the tag
comparison was the one input to the measurement that no domain reached, and the
existing `measure_agreement` cells all pass in-vocabulary tags, so the gap was
invisible.

## The change

`+3` statements in `measure_agreement`: read the holdout tag and refuse it when
it is neither `None` nor a member, mirroring `annotate_episode`'s wording so a
reader who has seen one refusal recognises the other.

The vocabulary check stays inline rather than moving to a shared `*_error`
helper, matching how this module already treats a fixed vocabulary:
`QUALITY_GRADES` is checked at four sites (`annotate_episode`, `filter_episodes`,
`_record_problem`, `measure_agreement`) with four surface-appropriate messages
and no helper. The tuple is the shared owner. A helper here would have to either
reword `_record_problem`'s read-side message - which deliberately explains the
downstream consequence rather than stating a domain - or carry two templates.

Placement is before the episode lookup, so a holdout entry the function cannot
compare is refused whether or not that episode happens to carry an annotation.
Otherwise the refusal would depend on how much of the holdout the judge got
through, which is the reason `read_labels` refuses an unreadable record rather
than skipping it.

Docs and docstring updated in the same change: the `failure_mode` row now names
its three spellings the way the `episode_index` row does, and the
`human_labels` argument states the domain the way it already states `quality`'s.

## Tests

`tests/test_holdout_failure_mode_vocabulary.py`, 50 cells in five classes.
Pre-fix, with only the source reverted:

| class | role | failed | total |
|---|---|---|---|
| `TestTheVocabularyIsAFieldDomainNotOneSurfacesConvention` | premise | 0 | 3 |
| `TestAnOutOfVocabularyHoldoutTagIsRefused` | regression | 13 | 4 defs |
| `TestTheRefusalDoesNotDependOnWhatTheJudgeGotThrough` | placement | 1 | 1 |
| `TestEveryReaderOfTheTagNamesTheVocabulary` | derived | 1 | 3 |
| `TestWhatIsUnchanged` | control | 0 | 5 |

`15 failed, 35 passed`. Nothing in the premise or control classes fires, so the
refusal is not new strictness: the premise class drives the same ten values
through the other two spellings and shows both already refuse them.

## Mutations

`fires` is failures in the new file; `sib` is failures in
`tests/test_episode_labels.py`, `tests/tools/test_episode_judge.py` and
`tests/simulation/test_replay_episode_index_domain.py`. `collected` was 50 on
every row, and the source was restored byte-identically after each.

| mutation | fires | sib | notes |
|---|---|---|---|
| control, no mutation | 0 | 0 | |
| the guard is removed | 15 | 0 | the whole pre-fix set |
| the guard moves one line down, ahead of the unjudged-episode skip | 0 | 0 | a semantic no-op - it still runs for every entry |
| the guard moves *past* that skip | 1 | 0 | the placement cell, alone |
| an explicit `None` is refused too | 3 | **58** | over-reach; 58 pre-existing cells rely on `None` and an absent tag being usable |
| only non-strings are refused | 10 | 0 | leaves every string typo, which is the realistic input |
| the tag is coerced to `None` instead of refused | 14 | 0 | a quieter wrong answer, not a fix |
| the message drops the holdout index | 4 | 0 | |
| the message drops the vocabulary | 2 | 0 | |
| the message drops the value | 2 | 0 | |
| a fourth reader of the tag lands, inventory **derived** | 1 | 0 | `test_the_shipped_module_has_no_such_reader`, naming it |
| the same fourth reader, inventory **hardcoded** to today's three | 1 | 0 | that cell goes silent; only the constructed-exemplar cell fires |

Twelve mutations, eleven detected, ten distinct firing sets. The last pair is why
the reader inventory is read off the module rather than listed: the whole defect
was one reader of this field that the domain had not reached, and with the
inventory hardcoded a new such reader is invisible to the survey - the
constructed-exemplar cell then catches the hardcoding itself rather than the
plant, which is the two layers doing different jobs.

## Gate

  tests_integ` (1594 files).
- mypy **24 == 24** against a pristine worktree of the merge base, normalised
  message sets byte-identical in both directions, 0 attributable to the changed
  files.
- Paired 379-file selection (everything naming `episode_labels` /
  `episode_judge` / `FAILURE_MODES` / `QUALITY_GRADES`, plus every guard that
  walks the tree, reads a docstring, or reads `docs/`, `README` or
  `changelog.d/`), with the new file excluded from both trees so the selection is
  identical: **18058 collected and `17 failed, 17974 passed, 67 skipped` on
  both**, failing-ID sets identical, `comm` empty in both directions. All 17 need
  `awsiot`, which is absent and declared in the `[mesh-iot]` extra.
- New file: 50 passed. With the three neighbouring suites plus the two guards `main`
  added since the merge base: 962 passed, 1 skipped.
- Coverage of `strands_robots/episode_labels.py`: `203 statements / 11 missing /
  95%` to `206 / 10 / 95%`. The scoped selection was checked against the
  full-suite run first and reports the same numbers and the same missing-line
  list for this module, so it is a faithful oracle here. The value is the pin,
  not the line count.

No visual artifact: this is a label-sidecar domain check, not policy,
simulation, rendering, recording or asset behaviour. The measured before/after
readings above are the evidence.


yinsong1986 (approved):
## Summary

Closes the one ungraded spelling of the `FAILURE_MODES` field domain: `measure_agreement` now refuses a holdout entry whose `failure_mode` is neither `None` nor a vocabulary member, instead of silently scoring it as a disagreement. Because the judge's tag is already confined to the vocabulary at both ends (`annotate_episode` on write, `_record_problem` on read-back), an out-of-vocabulary holdout tag could only ever pull `failure_mode_agreement` below the truth - the direction that makes a sound judge look unsound on the calibration that decides whether it filters training data. The guard is +3 statements, placed before the episode lookup so the refusal does not depend on how much of the holdout the judge got through, and its message mirrors `annotate_episode`'s wording while naming the holdout entry, the field, the vocabulary and the value.

## What's good

- Semantics preserved where they matter: explicit `None` stays a real, comparable opinion; an absent tag still means "not measured"; every vocabulary member still compares - all pinned in `TestWhatIsUnchanged`.
- The reader inventory in `TestEveryReaderOfTheTagNamesTheVocabulary` is derived from the module's AST rather than hardcoded, so a fourth reader of the tag is graded on arrival - which is exactly the class of gap this PR fixes.
- Keeping the vocabulary check inline rather than lifting a shared helper is argued against AGENTS.md's guard-sharing rule with the `QUALITY_GRADES` precedent (four sites, surface-appropriate messages), and `_record_problem`'s deliberately different read-side wording supports that call.
- Changelog fragment (not a direct `CHANGELOG.md` edit), docs domain table updated to name all three spellings the way the `episode_index` row does, docstring `Raises:` block updated in the same change, ASCII-clean throughout, and a thorough mutation table in the description.

PR #2780, opened 2026-08-26
fix(g1): apply the vendor's per-joint gains on rt/lowcmd, not one scalar

cagataycali:
## The defect

`_build_lowcmd_from_action` fills `kp`/`kd` for any joint the caller did not give
explicit gains for. Those defaults were a single pair — `kp=25.0`, `kd=0.5` —
applied to all 29 joints.

The G1's low-level position mode is tuned against a per-joint gain table. The
vendor's own `rt/lowcmd` reference for this robot ships 29-entry `Kp`/`Kd` lists
at module scope, and they are not flat: `Kp` takes three distinct values and `Kd`
two, because the joints do not carry comparable loads. Both knees are the
stiffest entries at `kp=100, kd=2`.

Measured on the frame the shipped builder produces, driving all 29 joints with
scalar targets:

| slots | joints | shipped | reference | under by |
|---|---|---|---|---|
| 3, 9 | both knees | 25.0 / 0.5 | 100 / 2 | **4.0x kp, 4.0x kd** |
| 0-2, 6-8, 12 | hips, waist yaw | 25.0 / 0.5 | 60 / 1 | 2.4x kp, 2.0x kd |
| 4-5, 10-11, 13-14, 15-28 | ankles, waist, arms | 25.0 / 0.5 | 40 / 1 | 1.6x kp, 2.0x kd |

Every shipped value sat strictly below every reference value, so no joint was
even coincidentally right:

```
EVERY shipped kp strictly below EVERY reference kp: True
EVERY shipped kd strictly below EVERY reference kd: True
distinct (kp, kd) under-by pairs: [(1.6, 2.0), (2.4, 2.0), (4.0, 4.0)]
worst kp deficit: 4.00x on ['left_knee', 'right_knee']
```

Three distinct ratio pairs is the reason this cannot be fixed by picking a better
scalar — no single value reproduces the table.

## Why nothing surfaced it

Two independent reasons, and they compound.

**The firmware does not validate gains.** It validates `crc`, `mode_machine` and
the per-motor Enable byte — the three fields whose absence makes a frame get
dropped. An under-gained frame is well formed, accepted, and reported as a
successful publish. Each joint's closed-loop stiffness is then whatever was
sent, and no status a caller can read says otherwise.

**The test that claimed to pin the defaults could not.** Its docstring said the
wire capture made a change to them "a test-visible event, not a silent
regression", while the assertion imported the two constants it compared against:

```python
from strands_robots.drivers.g1 import _DEFAULT_KD, _DEFAULT_KP
...
assert m.kp == pytest.approx(_DEFAULT_KP)
```

An expectation read from the constant under test follows any edit to that
constant, so it stayed green for every value including a wrong one.

Gains also sit **inside the checksum payload** — the vendor's `LowCmd_` pack
format spends `B3x5fI` per motor on mode, padding, `q`/`dq`/`tau`/`kp`/`kd`, and
reserve. Measured, changing one knee's gains moves the frame's CRC from
`3439364746` to `1975534031`. So gains have to be right before the CRC is
stamped; they are not correctable in a later frame.

## The change

`_SDK_KP` / `_SDK_KD` carry the reference table, indexed by the same slot as
`_G1_JOINT_INDEX`, and the builder subscripts them with the slot the joint name
resolved to. The tables are grouped the way the vendor groups its own lists —
two legs, a waist, two arms — so the 29-entry width follows from
`6 + 6 + 3 + 7 + 7` and the left/right symmetry is visible rather than asserted.

The old constants are removed rather than aliased. Behaviour a caller relied on
is unchanged: a supplied `kp`/`kd` still wins, and when only one of the two is
supplied the omitted term falls back to *that joint's* reference value rather
than to a flat default. The zero-torque stop frame still zeroes every gain.

One further correction: the removed constants' comment attributed `25.0 / 0.5` to
a reference stack's "rested-arm hold". That provenance is not locatable — the
vendor SDK's `example/g1/` tree contains no such gains, and its own value for an
arm joint is `40 / 1`.

## Verification

**The regression fails before the change and passes after.** Flattening the
table back to `25.0 / 0.5` and re-running:

| | failed | of |
|---|---|---|
| with the fix | **0** | 112 |
| table flattened to `25.0 / 0.5` | **37** | 112 |

Per class, on the flattened tree:

| class | pre-fix | role |
|---|---|---|
| `TestTheGainTableIsTheVendorReference` | 30/31 | regression |
| `TestTheGainsAreGenuinelyPerJoint` | 3/3 | regression |
| `TestTheWireFrameCarriesTheSlotsOwnGains` | 3/3 | regression |
| `test_g1_driver.py` (the strengthened cell) | 1/72 | regression |
| `TestTheBuilderIndexesTheTablePerSlot` | **0/1** | structural, value-independent |
| `TestWhatTheTableDoesNotChange` | **0/2** | over-reach guard |

Both guard classes are clean pre-fix, which is what makes them guards.

**Mutations.** Ten, each applied to the source with the tests held fixed;
`collected` is stable at 112 in every row, so no row hides behind a deselection:

| mutation | fires |
|---|---|
| control (fix in place) | **0** |
| table flattened to `25.0 / 0.5` | 37 |
| `kp` flattened to `40.0` — a value from *inside* the table's range | **15** |
| knee `kp` 100 -> 40 (one entry) | 7 |
| `kd` flattened to `1.0` (`kp` intact) | 8 |
| waist and second leg groups transposed | 8 |
| `kp` table truncated (one arm group dropped) | 13 |
| builder uses `_SDK_KP[0]` for every slot | 4 |
| zero-torque frame leaks the table (over-reach) | 2 |

Ten distinct firing sets, none undetected. The `40.0` row is the one that
matters: a scalar chosen from inside the table's own range is still caught, which
is exactly what an equality-against-the-constant assertion cannot do.

**The contract is graded without the vendor SDK.** `unitree-sdk2` is not a
declared dependency, so a contract asserted only behind `skipif(not _HAS_SDK)` is
asserted by nothing in CI. 35 of the 40 new cells need no SDK; the 5 wire cells
confirm the rest describes what lands on the topic.

```
no SDK on the path : 35 passed, 5 skipped
SDK available      : 40 passed
tests/drivers/     : 253 passed
```

**Paired gate**, 430 files (all of `tests/drivers/`, every test that walks the
tree or reads source/docstrings/`changelog.d`, and everything naming g1), the new
file excluded from both trees:

| | collected | failed | passed | skipped |
|---|---|---|---|---|
| this branch | 21705 | 12 | 21571 | 122 |
| base | 21705 | 12 | 21571 | 122 |

Failing-ID sets are identical, both `comm` directions empty. All 12 are
environmental and pre-existing: 7 in `tests/mesh/test_mesh_session.py` from a
`STRANDS_MESH_AUTH_MODE=mtls` value in the environment, and 5 that assert `rclpy`
is absent where it resolves.

tests_integ`; `mypy` reports `Success: no issues found in 1591 source files`,
with 0 errors attributable to the changed files.

## No visual artifact

The change is a wire-format default on a hardware driver, so there is no
simulation behaviour, rendering, recording or asset to show. I did attempt to
demonstrate the physical consequence and it is worth recording why it does not
work: the vendor's MuJoCo G1 drives its joints with position servos carrying
their own fixed `kp=500 / kv=43` (`gaintype=FIXED gainprm=500`,
`biastype=AFFINE biasprm=[0, -500, -43.01]`), so `ctrl` is a position target and
the `LowCmd_` gains have no representation in that model at all — with
`ctrl[:] = 0` it holds the keyframe pose at `z=0.7915`. Demonstrating the deficit
would mean rebuilding those actuators as torque motors and adding a balance
controller, which would measure a model I authored rather than the robot. The
measured tables above are the evidence; nothing in the description claims a
physical outcome I did not measure.

## Scope

Issue #2765 lists this as decidable against the SDK without hardware, alongside
several items that are not. This change takes only the gain table. The remaining
items there — in particular whether an `rt/lowcmd` write should consult an FSM
gate at all — need a G1 in the room, so the issue stays open.

### DIFF
--- a/changelog.d/2780-g1-per-joint-gains.md
+++ b/changelog.d/2780-g1-per-joint-gains.md
@@ -1,0 +1,42 @@
+### Fixed: the G1 low-level write path applies the vendor's per-joint gains, not one scalar
+
+`G1Driver.send_action` builds a `LowCmd_` for `rt/lowcmd`, filling `kp`/`kd` for
+any joint the caller did not give explicit gains for. Those defaults were a
+single pair -- `kp=25.0`, `kd=0.5` -- applied to all 29 joints. The G1's
+low-level position mode is tuned against a per-joint gain table whose `Kp` takes
+three distinct values and `Kd` two, because the joints do not carry comparable
+loads: both knees are the stiffest entries at `kp=100, kd=2`, and they are the
+joints that hold a standing biped up.
+
+Every entry of the scalar pair sat strictly below every entry of the reference
+table, so no joint was even coincidentally correct. The deficit arrived in three
+distinct ratio pairs -- 1.6x stiffness and 2.0x damping on the ankles, waist and
+arms; 2.4x and 2.0x on the hips and waist yaw; 4.0x and 4.0x on both knees --
+which is why no single scalar, chosen anywhere in the table's range, can
+reproduce it.
+
+Nothing surfaced this. The firmware validates `crc`, `mode_machine` and the
+per-motor Enable byte; it does not validate gains. An under-gained frame is
+therefore well formed, accepted, and reported as a successful publish, so the
+only symptom is a robot that tracks its target badly or sags under its own
+weight, with no error for an operator to read. Gains also sit inside the
+checksum payload -- the vendor's `LowCmd_` pack format spends `B3x5fI` per motor
+on mode, padding, `q`/`dq`/`tau`/`kp`/`kd`, and reserve -- so they have to be
+correct before the CRC is stamped rather than corrected in a later frame.
+
+The gain tables are now grouped the way the vendor groups its own lists (two
+legs, a waist, two arms), so the 29-entry width follows from 6 + 6 + 3 + 7 + 7
+and the left/right symmetry is visible rather than asserted. Behaviour a caller
+already relied on is unchanged: a supplied `kp`/`kd` still wins, including when
+only one of the two is supplied, in which case the omitted term falls back to
+that joint's own reference value rather than to a flat default. The zero-torque
+stop frame still zeroes every gain, which is what "soft" means on that wire.
+
+This also strengthens the test that claimed to pin the defaults. Its docstring
+said the wire capture made a change to them "a test-visible event, not a silent
+regression", while the assertion imported the two constants it was comparing
+against -- an expectation read from the constant under test follows any edit to
+that constant, so it stayed green for every value including a wrong one. The
+expected numbers are now literals, and the new regression file grades the whole
+table against a locally stated copy of the reference, slot by slot, without
+needing the vendor SDK installed.
--- a/changelog.d/2780-g1-per-joint-gains.md
+++ b/changelog.d/2780-g1-per-joint-gains.md
@@ -17,9 +17,9 @@ reproduce it.
 
 Nothing surfaced this. The firmware validates `crc`, `mode_machine` and the
 per-motor Enable byte; it does not validate gains. An under-gained frame is
-therefore well formed, accepted, and reported as a successful publish, so the
-only symptom is a robot that tracks its target badly or sags under its own
-weight, with no error for an operator to read. Gains also sit inside the
+therefore well formed, accepted, and reported as a successful publish. The
+closed-loop stiffness of each joint is whatever was sent, and no status a caller
+can read says otherwise. Gains also sit inside the
 checksum payload -- the vendor's `LowCmd_` pack format spends `B3x5fI` per motor
 on mode, padding, `q`/`dq`/`tau`/`kp`/`kd`, and reserve -- so they have to be
 correct before the CRC is stamped rather than corrected in a later frame.
--- a/strands_robots/drivers/g1.py
+++ b/strands_robots/drivers/g1.py
@@ -131,12 +131,39 @@ _G1_JOINT_INDEX: dict[str, int] = {
 # ``for i in range(G1_NUM_MOTOR)``, where ``G1_NUM_MOTOR = 29``).
 _G1_NAMED_JOINTS: int = max(_G1_JOINT_INDEX.values()) + 1
 
-# PD gain defaults used when a caller does not supply per-joint gains.  These
-# are the gains the neon reference stack uses for a rested-arm hold; they are
-# deliberately conservative and match what the FSM 501 (sitting) hold expects.
-# A caller who cares supplies ``kp``/``kd`` in the action dict.
-_DEFAULT_KP: float = 25.0
-_DEFAULT_KD: float = 0.5
+# Per-joint PD gains applied when a caller does not supply ``kp``/``kd``,
+# indexed by the same slot as :data:`_G1_JOINT_INDEX`.  Transcribed from the
+# vendor's own ``rt/lowcmd`` reference for this robot -- the module-scope
+# ``Kp``/``Kd`` lists in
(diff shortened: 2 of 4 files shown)


yinsong1986 (dismissed):
## Summary

Replaces the single scalar PD-gain default (`kp=25.0`, `kd=0.5`) in the G1 driver's `_build_lowcmd_from_action` with the vendor's per-joint 29-entry gain table, indexed by the same slot as `_G1_JOINT_INDEX`. The table is composed as `leg + leg + waist + arm + arm` (6+6+3+7+7), so the 29-entry width falls out of the group structure. Explicit caller-supplied `kp`/`kd` still win over the table, and the zero-torque stop frame stays all-zeros. I independently verified the transcribed values against the vendor's `unitree_sdk2_python/example/g1/low_level/g1_low_level_example.py` module-scope `Kp`/`Kd` lists on master -- they match slot for slot, including the knee entries at `kp=100, kd=2` and the waist `(60, 40, 40)` asymmetry.

## What's good

- The new `tests/drivers/test_g1_per_joint_gains.py` states the vendor's numbers locally as an independent oracle rather than importing the constant it grades, and splits SDK-free contract cells (which run in CI) from SDK-gated wire cells -- exactly the layering that would have caught the original flat-default defect.
- Over-reach guards pin what the change must NOT touch: supplied gains still win, and the zero-torque stop frame stays soft.
- Ran the two driver test files at head: 87 passed, 25 skipped (SDK-gated). `ruff check` / `ruff format --check` clean; the repo's whole-tree grader tests (host paths, ASCII tool strings, test naming, xref roles, except-tuple scope) all pass on the head tree.

## Verification suggestions

- The PR carries no `changelog.d/2780-*.md` fragment; `.github/workflows/changelog-fragment.yml` will surface that (self-clearing, per AGENTS.md step 3).
- On real hardware, note the scalar-target default is now materially stiffer on the legs (knee `kp` 25 -> 100), which is the point of the fix -- but worth a hardware smoke of a scalar hold before relying on it in a demo.

yinsong1986 (approved):
## Summary

Replaces the G1 driver's single scalar PD-gain default (`kp=25.0`, `kd=0.5`) with the vendor's per-joint 29-entry table (`_SDK_KP`/`_SDK_KD`), composed as `leg + leg + waist + arm + arm` and indexed by the same slot as `_G1_JOINT_INDEX`. Caller-supplied gains still win (including the partial case, which now falls back to the slot's own value), and the zero-torque stop frame stays all-zeros. I re-verified the transcription against the vendor's `unitree_sdk2_python/example/g1/low_level/g1_low_level_example.py` module-scope `Kp`/`Kd` lists on master: they match slot for slot, including both knees at `kp=100, kd=2` and the waist `(60, 40, 40)` asymmetry. The removed `_DEFAULT_KP`/`_DEFAULT_KD` were private with no consumers outside this driver and its own test, so the deletion breaks no external surface, and the `changelog.d/2780-*` fragment is present per AGENTS.md step 3.

## What's good

- The new test file states the vendor's numbers locally as an independent oracle instead of importing the constant it grades - closing exactly the tautology that let the flat default ship past a test claiming to pin it - and splits SDK-free contract cells (which CI actually runs) from SDK-gated wire cells.
- Independently reproduced the regression's teeth: at head, the two driver test files pass (87 passed, 25 SDK-gated skips); with the `kp` table flattened back to `25.0`, 33 of the 40 new cells fail without the SDK on the path. The over-reach guards (supplied gains win, stop frame stays soft) are held apart from the regression cells, as documented.
- `ruff check` clean on the three changed Python files; all new user-facing and source text is plain ASCII.

PR #2781, opened 2026-08-26
fix(simulation): stop a backend permuting a parameter it shares with a sibling

cagataycali:
## Problem

`create_simulation(backend=...)` makes the same `sim` variable a different class, and nothing relates two backends' signatures for a method both implement. `add_camera` is on no ABC at all -- `SimEngine.__abstractmethods__` does not list it -- and `randomize` is on `SimEngine` only as a `**kwargs` sink whose docstring says "Concrete backends define their own parameter signatures". Two methods had drifted into ordering a parameter they share differently, so one positional call meant two things.

Measured on `main` (`f5876e0e`), one call, three backends:

```
sim.add_camera("wrist", [0.1, 0.0, 0.05], [0.3, 0.0, 0.0], 100, 200, 90)

mujoco   fov=100  width=200  height=90
newton   fov=100  width=200  height=90
isaac    fov=90   width=100  height=200
```

Every one of those six values is inside its own domain under either reading: `fov=100` and `fov=90` are both in `(0, 180)`, and 100, 200 and 90 are all positive integers. So no guard refuses the call, nothing warns, and the difference arrives in the pixels.

`NewtonSimEngine.randomize` had the same shape, listing `mass_range, friction_range, color_range` where MuJoCo lists `color_range, friction_range, mass_range` -- three parameters of the same type, so nothing distinguishes them by value either.

Both contradicted a parity claim already in the tree. Newton's `randomize` docstring said "Keyword names and defaults mirror the MuJoCo backend so randomization code transfers across backends unchanged" -- a conclusion wider than its premise, with the order as the missing third term. `docs/simulation/newton.md` documents the camera order as `add_camera(name, position, target, fov=60, width, height, parent_body=None)` and calls it "matching the MuJoCo signature", and `docs/simulation/domain-randomization.md` lists the three ranges in MuJoCo's order -- so the pages a reader copies from already stated the order the signatures did not keep.

## Severity, honestly split

The two divergences are not equally dangerous, and the PR does not claim they are.

| | reachable silently? | why |
|---|---|---|
| `add_camera` | yes | all three backends share every parameter, so a positional call binds cleanly under both readings |
| `randomize` | no | MuJoCo declares two axes Newton lacks (`randomize_positions`, `position_noise`), so a positional call hits the boolean-flag domain at index 4 and is refused loudly |

`randomize` is fixed because the rule cannot land with a known violation in it, and because the documented order is the one a reader transfers -- not because a caller is silently losing colour for mass today.

## Change

Reorder the two signatures onto the order the documentation already gives, and make each docstring's parameter block follow its signature. Adding a parameter is still allowed and shifts no shared name's relative order, so `IsaacSimulation.add_robot`'s `mjcf_path`/`usd_path`, `NewtonSimEngine.add_robot`'s `source` and MuJoCo's two extra randomization axes are untouched -- as are each backend's own defaults (Isaac still resolves an omitted `width` from `IsaacConfig` where MuJoCo defaults it to 640).

No in-tree caller is affected: an AST sweep of `strands_robots`, `tests`, `tests_integ`, `examples`, `scripts` and `docs` finds 263 `add_camera` calls and 83 `randomize` calls, of which the largest positional-argument count anywhere is 2 for `add_camera` and 1 for `randomize`.

## Guard

`tests/simulation/test_backend_shared_parameter_order.py` derives its backend inventory from `_BUILTIN_BACKENDS`, the table `create_simulation` resolves, so a fourth backend is graded when it lands rather than when someone extends a list. It needs no Isaac Sim, no Newton, no GPU and no GL: only signatures are read, and all three engine modules import on a host with neither optional runtime installed.

Pre-fix split: **8 failed / 167 passed**, and the three classes that must not move are clean.

|---|---|---|
| `TestNoBackendPermutesASharedParameter` | 3 / 66 | the derived rule |
| `TestOnePositionalCallMeansOneThingEverywhere` | 2 / 2 | the consequence, via `signature.bind` |
| `TestTheDocumentedOrderIsEveryBackendsOrder` | 2 / 2 | the two pages that write an order down |
| `TestWhatThisDoesNotDecide` | 1 / 2 | fires because a permutation is not an insertion |
| `TestThePopulationIsReal` | 0 / 5 | premise |
| `TestTheRuleIsNotVacuous` | 0 / 4 | non-vacuity |
| `TestABackendMayStillAddItsOwnParameters` | 0 / 5 | over-reach control |

## Mutation table

Each row mutates the tree, runs the new module (`own`) and the pre-existing neighbouring suites (`sib`), and records `collected`. Source restored md5-identical after every row.

| row | own | sib | coll | mutation |
|---|---|---|---|---|
| b0 | 0 | 0 | 175 | control, no mutation |
| b1 | 6 | 0 | 175 | Isaac `add_camera` order reverted |
| b2 | 3 | 0 | 175 | Newton `randomize` order reverted |
| b3 | 8 | 0 | 175 | both reverted (the behavioural pre-fix state) |
| b4 | 6 | 0 | 175 | rule weakened to a set comparison, plus both reverted |
| b5 | 6 | 0 | 175 | rule disabled outright, plus both reverted |
| b6 | 1 | 0 | 175 | keyword-only parameters folded into the positional list |
| b7 | 7 | 0 | **292** | a fourth permuting backend, inventory **derived** |
| b8 | **0** | 0 | **175** | the same plant, inventory **hardcoded** to today's three |
| b9 | 1 | 0 | 175 | documented order read by bare mention instead of by assignment |
| b10 | 9 | 0 | 175 | report only the first differing name, plus both reverted |

Eight distinct firing sets across the nine non-empty rows. Notes worth reading:

- **b1 and b2 are complementary and exact**: `b1 | b2 == b3` is `True`, and their single-element overlap is `test_every_index_divergence_is_an_insertion_not_a_permutation`, which fires if either half is a permutation. So neither fix rides on the other's tests.
- **b7 against b8 is why the inventory is derived.** The same planted fourth backend fires 7 tests with the inventory derived and **0** with it hardcoded -- and the hardcoded run also collects 117 fewer tests, so a suite that silently stopped grading a backend would still have reported success. A hardcoded list equal to today's derived set fires nothing and looks like a working test; this pair is the only way to show the difference.
- **b4 and b5 share a firing set** for a structural reason: both make the predicate return `()` unconditionally. Their six failures include the two `TestTheRuleIsNotVacuous` exemplars, which is exactly what those exist for -- with the derived scan blinded, the constructed permuting/inserting pair still catches the weakened rule.
- **b9** is the trap this module's own helper had to be fixed for: the randomization page annotates `randomize_physics` with a comment naming `mass_range` and `friction_range` before either is passed, so a bare-mention scan reads that comment as the documented order. The helper matches `name=` instead.
- **`sib` is 0 on all eleven rows** across the pre-existing camera, randomization, mount-capability, reserved-name and Isaac-backend suites. That is the measurement of why nothing caught this: the closest existing guard, `tests/simulation/test_camera_mount_capability_across_backends.py`, reads `inspect.signature(engine.add_camera).parameters` -- but asks only whether `parent_body` is present and defaults to `None`, never where anything sits.

## What this does not decide

The rule is "a backend may add a parameter but must not permute one it shares", which is deliberately weaker than "a shared parameter sits at the same positional index". Index parity is still violated -- measured, by `add_robot` (Isaac inserts `mjcf_path`/`usd_path` mid-signature, shifting `data_config`, `position`, `orientation`, `keyframe`), by `randomize`, and by `start_cameras_recording`. Closing that would move parameters off positions they have shipped at on the primary entry point, which is an API decision rather than a correction. `TestWhatThisDoesNotDecide` measures that set, asserts it is non-empty, and asserts every member of it is an insertion rather than a permutation, so the boundary is pinned instead of left to be rediscovered.

## Visual

One positional add_camera call, read two ways

Both panels are rendered by the MuJoCo engine from the identical scene (SO-101 arm plus a cube) and the identical camera pose, so the only difference between them is which parameter each value bound to. Left is the shared reading (`fov=100`, 200x90) that MuJoCo, Newton and now Isaac all produce; right is the reading Isaac produced before this change (`fov=90`, 100x200). Native render sizes were read back from the frames and asserted against the request before the figure was drawn; panels are nearest-neighbour upscaled 2x for legibility.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1595 files).
- `mypy strands_robots tests tests_integ`: **24 on the branch, 24 on a pristine worktree of the merge-base**, normalized message sets byte-identical in both directions, none in any changed file.
- Paired run over an identical 795-file selection (all of `tests/simulation/`, plus every test that walks the tree, parses source, reads docstrings, reads `docs/`, `README`, `AGENTS.md` or `changelog.d`, or names `add_camera` / `randomize` / `create_simulation` / `_BUILTIN_BACKENDS`), with the new file excluded from **both** trees and every path verified present on both: identical **28917 collected** and **`18 failed, 28613 passed, 286 skipped`** on each, failing-ID sets identical, `comm` empty in both directions. All 18 are pre-existing: 17 `tests/mesh/test_iot_*` needing `awsiot` (absent; declared in the `[mesh-iot]` extra) and 1 `tests/training/test_lerobot_e2e.py`.
- New module: 175 passed on mujoco 3.5.0, the declared floor.
- Neighbouring suites after the change: 2296 passed, 16 skipped.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -213,6 +213,29 @@ hatch run format            # ruff check --fix, ruff format
     tests/mesh/test_ackermann_command_gate.py, whose inventory of bridges owing
     a gate suite is derived from the tree, so the next transport is graded on
     arrival rather than at the review that happens to look.
+19. **A backend may add a parameter but must never permute one it shares** -
+    `create_simulation(backend=...)` makes the same `sim` variable a different
+    class, and nothing in the type system relates two backends' signatures for a
+    method both implement: `add_camera` is on no ABC at all and `randomize` is
+    on `SimEngine` only as a `**kwargs` sink whose docstring hands the signature
+    to the backend. So a permutation is invisible to every gate and to Python.
+    Isaac declared `add_camera(..., width, height, fov, ...)` where MuJoCo and
+    Newton declare `(..., fov, width, height, ...)`, so
+    `add_camera("wrist", pos, target, 100, 200, 90)` asked for a 200x90 view at
+    fov 100 on two backends and a 100x200 view at fov 90 on the third - every
+    value valid under either reading, so nothing refused it and the caller found
+    out from the pixels. Newton's `randomize` had the same shape, listing the
+    three ranges in reverse. Adding a parameter is fine and shifts no shared
+    name's *relative* order (Isaac's `mjcf_path`/`usd_path`, Newton's `source`,
+    MuJoCo's `randomize_positions`); reordering one is what makes a positional
+    call ambiguous. State the shared order where a reader will copy it - the
+    signature line in `docs/simulation/newton.md` and the call in
+    `docs/simulation/domain-randomization.md` are both graded against the
+    signatures - and note that this rule is deliberately weaker than index
+    parity, which a mid-signature insertion still breaks. Pinned by
+    tests/simulation/test_backend_shared_parameter_order.py, whose backend
+    inventory is derived from the table `create_simulation` resolves, so a
+    fourth backend is held to the rule the hour it lands.
 
 ## PR Workflow
 
--- a/changelog.d/2781-backend-shared-parameter-order.md
+++ b/changelog.d/2781-backend-shared-parameter-order.md
@@ -1,0 +1,40 @@
+### Fixed: a backend no longer permutes a parameter it shares with a sibling
+
+`create_simulation(backend=...)` makes the same `sim` variable a different class, and nothing in the
+type system relates two backends' signatures for a method both implement. `add_camera` is on no ABC
+at all -- `SimEngine.__abstractmethods__` does not list it -- and `randomize` is on `SimEngine` only
+as a `**kwargs` sink whose docstring hands the signature to the backend. Two of them had drifted into
+ordering a parameter they share differently, so one positional call meant two different things and
+nothing was in a position to notice.
+
+`IsaacSimulation.add_camera` declared `(name, position, target, width, height, fov, parent_body)`
+where `MuJoCoSimEngine` and `NewtonSimEngine` both declare `(name, position, target, fov, width,
+height, parent_body)`. So `add_camera("wrist", pos, target, 100, 200, 90)` asked for a 200x90 view at
+fov 100 on two backends and a 100x200 view at fov 90 on the third. Every one of those six values is
+inside its own domain under either reading -- `fov=100` and `fov=90` are both in `(0, 180)`, and 100,
+200 and 90 are all positive integers -- so no guard refused the call, nothing warned, and the caller
+found out from the pixels. `docs/simulation/newton.md` states that order as the shared one and calls
+it "matching the MuJoCo signature", so the divergence also contradicted the page a reader copies it
+from.
+
+`NewtonSimEngine.randomize` had the same shape, listing `mass_range`, `friction_range`, `color_range`
+where MuJoCo lists `color_range`, `friction_range`, `mass_range` -- a reversal of three parameters
+that carry the same type, so nothing distinguishes them by value either. Its docstring said "Keyword
+names and defaults mirror the MuJoCo backend so randomization code transfers across backends
+unchanged", a conclusion wider than its premise: names and defaults did mirror, and the missing third
+term was the order. That divergence is caught rather than silent, because MuJoCo declares two axes
+Newton does not (`randomize_positions`, `position_noise`) and a positional call reaches the
+boolean-flag domain first, but the two signatures still could not be read against each other.
+
+Both signatures now declare their shared parameters in the order the documentation already gives, and
+both docstrings' parameter blocks follow the signature. Adding a parameter is still allowed and
+shifts no shared name's relative order, so `IsaacSimulation.add_robot`'s `mjcf_path` / `usd_path`,
+`NewtonSimEngine.add_robot`'s `source` and MuJoCo's two extra randomization axes are untouched, as
+are each backend's own defaults -- Isaac still resolves an omitted `width` from `IsaacConfig` where
+MuJoCo defaults it to 640.
+
+The rule is pinned by `tests/simulation/test_backend_shared_parameter_order.py`, whose backend
+inventory is derived from the table `create_simulation` resolves rather than from a list in the test,
+so a fourth backend is graded when it lands. The rule is deliberately weaker than "a shared parameter
+sits at the same positional index", which a mid-signature insertion still breaks; that boundary is
+measured and pinned in the same module rather than left to be rediscovered.
--- a/docs/simulation/domain-randomization.md
+++ b/docs/simulation/domain-randomization.md
@@ -52,6 +52,12 @@ domain, so an axis you can turn off on one is not left on by the other. The
 numeric knobs in the same signature keep their own domain: a `mass_range` is a
 quantity, and it is still refused as a range rather than as a flag.
 
+They also declare the parameters they share in the order shown above. Newton
+adds no axis of its own and MuJoCo adds two (`randomize_positions`,
+`position_noise`), so the two signatures are not the same length - but neither
+reorders a name the other also carries, so the three ranges cannot be read as
+each other.
+
 **Destructive** - writes into MuJoCo model arrays. To restore: `load_scene(...)` or recreate
 the sim. Every *other* scene mutation restores it too, as a side effect of rebuilding the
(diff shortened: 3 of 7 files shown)


yinsong1986 (dismissed):
## Summary

Aligns two backend signatures onto the parameter order the docs already document as shared: `IsaacSimulation.add_camera` moves `fov` ahead of `width`/`height` (matching MuJoCo and Newton), and Newton's `randomize` lists the three range parameters in MuJoCo's order (`color_range`, `friction_range`, `mass_range`). Both divergences were unreachable by any gate (`add_camera` is on no ABC; `randomize` is a `**kwargs` sink on `SimEngine`), so one positional call silently meant different things per backend. The PR pins the rule with a derived test module whose backend inventory comes from `_BUILTIN_BACKENDS`, updates the two doc pages the tests grade against, and records the convention as AGENTS.md rule 19.

Verified locally at the head SHA: the new test module passes (175 tests, importable without Isaac/Newton runtimes), the existing Isaac `add_camera` validation suite passes (93 tests), the 11 errors in `tests/simulation/newton/test_domain_randomization.py` reproduce identically on the base commit (an installed-Newton version drift in `simulation.py`, untouched by this diff), and `ruff check` / `ruff format --check` are clean on all changed files. Every in-tree caller of both reordered methods passes by keyword, so nothing in the repo observes the reorder; the only behaviour change is that a positional cross-backend call now binds identically everywhere, which is the fix. The Isaac reorder does change what a positional Isaac-only call means, but the old order contradicted the shared order `docs/simulation/newton.md` and `docs/simulation/overview.md` already documented, so this is a correction toward the published contract rather than a new one-way door.

## What's good

- The test universe is derived from the table `create_simulation` resolves, so a fourth backend is graded on arrival - matching the repo's derived-inventory convention.
- `TestTheRuleIsNotVacuous` grades the predicate on constructed exemplars, so the module cannot pass vacuously on a clean tree.
- `TestWhatThisDoesNotDecide` measures and pins the deliberate boundary (relative order, not index parity) instead of leaving it to be rediscovered.
- Docs-vs-signature grading (`_documented_signature_order` reads one line; `_documented_call_order` matches assignments, not prose mentions) prevents the pages and signatures drifting apart.

Two non-blocking notes for the author: the PR body is still "Placeholder - full body to follow", and no `changelog.d/2781-*.md` fragment is in the diff - the changelog-fragment workflow will surface the latter.

yinsong1986 (approved):
## Summary

Aligns two backend method signatures onto the parameter order the documentation already states as shared: `IsaacSimulation.add_camera` moves `fov` ahead of `width`/`height` (matching MuJoCo and Newton, and matching the order `docs/simulation/overview.md` and `docs/simulation/newton.md` already document), and `NewtonSimEngine.randomize` lists the three range parameters in MuJoCo's order (`color_range`, `friction_range`, `mass_range` — verified against the MuJoCo signature). Both divergences were invisible to every gate (`add_camera` is on no ABC; `randomize` is a `**kwargs` sink on `SimEngine`), and the `add_camera` one was silently reachable since every value in a permuted positional call stays inside its own domain. The reorder of Newton's `_RANDOMIZE_PARAMS` tuple is safe: its only consumers are membership checks in `unknown_kwargs_error`, whose hint output is independently sorted. An independent AST sweep confirms the PR's claim that no in-tree caller passes more than 2 positional arguments to `add_camera` or 1 to `randomize`, so no in-tree behaviour changes beyond the fix itself; the positional-order change for external Isaac callers is the deliberate correction onto the documented contract, weighed openly in the description. The new guard module derives its backend inventory from `_BUILTIN_BACKENDS` (per the derived-inventory convention), imports without Isaac/Newton/GPU, and passed 175/175 on this head; the neighbouring camera-mount suite passes unchanged, docstring parameter blocks follow the reordered signatures, a changelog fragment is present, and ruff check/format are clean on every changed file.

## What's good

- The rule ships with its counterexample controls: `TestTheRuleIsNotVacuous` grades the predicate on constructed permuting/inserting pairs, so a weakened scan cannot pass silently, and the b7/b8 mutation pair demonstrates why the inventory is derived rather than hardcoded.
- `TestWhatThisDoesNotDecide` pins the deliberate boundary (relative order, not index parity) instead of leaving it to be rediscovered.
- Docs graded against signatures (`_documented_signature_order` reads the signature line, `_documented_call_order` matches `name=` assignments) so the two pages and the code cannot drift apart again.
- AGENTS.md rule #19 records the convention alongside the pin, matching the repo's documented-and-enforced pattern.

PR #2782, opened 2026-08-26
fix(docs): build the teleoperator the mesh teleop example publishes from

cagataycali:
## The defect

`docs/mesh.md` opened its mesh-teleop recipe by reading a `teleoperator` attribute off the robot:

```python
leader = Robot("so100", mode="real", mesh=True)
leader.start_teleop_publish(teleoperator=leader.teleoperator, ...)
```

That attribute does not exist. Measured against `hardware_robot.Robot`, which is what
`Robot("so100", mode="real")` returns:

| probe | result |
|---|---|
| `hasattr(cls, "teleoperator")` | `False` |
| `"teleoperator" in dir(cls)` | `False` |
| `self.teleoperator = ...` anywhere in the MRO | `False` |
| MRO | `Robot, TeleopMixin, AgentTool, ABC, object` |

So the documented bring-up line raises `AttributeError` on the first statement a reader runs. The
teleop surface is `attach_teleop(...)` plus the private `_teleops`; `start_teleop_publish` documents
its parameter as "Any object with a `get_action() -> dict` method", and the documented way to build
one is the `Teleoperator()` factory. The example now builds it the way
`docs/hardware/teleoperation.md:34` already does.

## Why both graders over this documentation stay green

`Robot()` is polymorphic, and that is what makes this class of error hard to see by reading:

| call | returns | surface |
|---|---|---|
| `Robot(name)` / `mode="sim"` | `Simulation` | 343 names |
| `mode="real"`, registry declares no driver | `hardware_robot.Robot` | 121 names |
| `mode="real"`, registry declares `hardware.driver="strands"` | the native driver | `DRIVER_SURFACE`, 12 names |

`robot.attach_teleop(...)` is therefore a correct line for one robot and an `AttributeError` for
another, written identically. Measured: `attach_teleop` is on the wrapper's surface and not on
`G1Driver`'s.

Neither existing grader can reach it, by construction rather than by oversight:

- `tests/test_docs_real_mode_invocations.py` grades the robot *name* and the keywords *inside* the
  `Robot(...)` call.
- `tests/test_docs_python_examples_are_callable.py` grades keyword sets against signatures, and its
  `_accepted_keywords` returns `None` ("any keyword binds") for a callee carrying `**kwargs`, which
  `Robot` does.

Attribute access on the factory's *return value* is outside both. Both files report **20 passed**
over the broken line and **20 passed** over the corrected one, and the `sib` column of the mutation
table below is `0` on all eleven rows.

## The guard

`tests/test_docs_robot_attribute_reads_resolve.py` resolves each documented read to the type the
factory would return for that `(name, mode)` pair and requires the name to exist there. The surface
is the union of three sources, because an attribute can arrive from any of them:

1. the class and its MRO, via `dir`;
2. `self.X = ...` assigned anywhere in the MRO -- instance state a class sets up, which `dir` on the
   class cannot see. `control_frequency`, `data_config` and `robot` are reachable only this way;
3. `instance.X = ...` bound by the factory in `strands_robots.robot`.

The third source is load-bearing rather than defensive. `_attach_device_connect` binds
`instance.run = ...`, so `run` is on no class at all -- a surface derived from the classes alone
reports the documented `Robot("so100").run()` as missing. I hit that while building this and it was
the one candidate offender I had to discard.

## Corpus

Across 58 documentation pages the scan finds **65 distinct reads**: 5 pages contribute hardware-mode
reads, 20 contribute simulation-mode reads. Exactly **one** was unresolvable, and it is the line
above. The other 64 resolve unchanged -- including all 26 `mode="real"` reads, because `koch` and
`aloha` register a native driver class but their registry entries declare no `hardware.driver`, so
those calls resolve to the lerobot wrapper where `attach_teleop` and `teleoperate` exist.

Because the corpus is expected to be clean after the fix, the rule is also graded over constructed
exemplars (accepted / factory-bound / instance-state / misspelled / wrong-surface) through the same
predicate, with a check that they reach both verdicts. Otherwise the guard would pass by having
nothing left to classify.

## Pre-fix split

`1 failed, 9 passed` with the docs untouched. The single failure is the defect:

```
docs/mesh.md: Robot('so100', mode='real').teleoperator is not on ['HardwareRobot']
```

The 9 passing cells are the premise, exemplar and simulation-mode groups -- none of them is a
regression cell, so nothing here is graded by accident.

## Mutation table

11 mutations, control clean, **0 undetected**, 10 distinct firing sets, `collected` stable at 11,
sources restored byte-identical (md5 asserted per row). `sib` is failures in the two pre-existing
docs graders.

| mutation | fires | sib |
|---|---|---|
| control, no mutation | 0 | 0 |
| docs reverted to `leader.teleoperator` | 1 | 0 |
| factory bindings not credited | 3 | 0 |
| `self.X` MRO assignments not credited | 1 | 0 |
| native-driver branch removed (always the wrapper) | 2 | 0 |
| assigned-variable tracking dropped | 5 | 0 |
| unregistered-name skip widened to skip everything | 3 | 0 |
| mode routing broken (real graded as sim) | 4 | 0 |
| scan narrowed to `README.md` only | 2 | 0 |
| exemplars reduced to the accepted ones | 1 | 0 |
| doctest `>>>` stripping dropped | 1 | 0 |

Two rows are worth naming. "Scan narrowed to `README.md` only" fired **0** on a first pass, because
the non-vacuity check only required *some* read to exist and `README.md` supplies ten; it now
requires a floor of contributing pages per partition, set below today's counts so an ordinary docs
edit does not trip it while a narrowed glob does. "`self.X` not credited" also fired 0 until the
instance-state exemplar was added -- `control_frequency` is answered by no other source. Both were
gaps in the test rather than clean rows.

The last two rows share a firing set with the first, which is structural rather than thin: with the
docs reverted, the defect sits in a plain fence, so losing `>>>` stripping does not change which
cell reports it.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1595 files).
- mypy **24 == 24** against a worktree of the merge base, normalised message sets byte-identical in
  both directions, **0** attributable to the new file.
- Paired run over an identical 603-file selection (every test reading `docs/` or `README`, walking
  the tree, or naming the factory, registry or drivers), with the branch-only file excluded from
  both sides so the selections match: **26059 collected** and `31 failed, 25850 passed, 178 skipped`
  on both trees, **31 == 31 identical failing IDs, `comm` empty both directions**, distinct
  rootdirs. 17 of the 31 need `awsiot` from the `[mesh-iot]` extra, which is absent here; the rest
  are pre-existing on the base.
- Simulation-mode grading needs `mujoco`, which is in `[sim-mujoco]` rather than core, so that one
  cell is behind an `importorskip` while the hardware-mode half -- where the defect is -- runs on a
  core install.

The paired selection was measured at merge base `b74ec0a9`; the base has since gained #2777 and
#2778, and `git diff --name-only` reports neither touching either file here. Lint, mypy and the
changed-file suites were re-run after rebasing onto that base.

No visual artifact: this changes documentation prose and adds a test, none of the categories that
call for one.

### DIFF
--- a/changelog.d/2782-docs-robot-result-attribute-reads.md
+++ b/changelog.d/2782-docs-robot-result-attribute-reads.md
@@ -1,0 +1,37 @@
+### Fixed: the mesh teleop example builds the teleoperator it publishes from
+
+`docs/mesh.md` opened its mesh-teleop recipe with
+`leader.start_teleop_publish(teleoperator=leader.teleoperator, ...)`. There is no `teleoperator`
+attribute on what `Robot("so100", mode="real", mesh=True)` returns: it is on no class in the
+wrapper's MRO, and no method assigns `self.teleoperator`. The documented bring-up line raised
+`AttributeError` on the first statement a reader ran. `start_teleop_publish` wants an object with
+`get_action()`, and the documented way to build one is the `Teleoperator()` factory, so the example
+now builds it the way `docs/hardware/teleoperation.md` does.
+
+The class of error is worth guarding rather than only correcting, because `Robot()` is polymorphic
+and the risky spelling is indistinguishable from the correct one. `mode="sim"` returns a
+`Simulation`; `mode="real"` returns the `hardware_robot.Robot` wrapper *unless* the registry entry
+declares `hardware.driver = "strands"`, in which case the factory returns the native driver itself,
+whose surface is `DRIVER_SURFACE`. So `robot.attach_teleop(...)` is a correct line for a
+lerobot-backed robot and an `AttributeError` for a natively-driven one, written identically.
+
+Neither existing grader over that documentation can see it. `test_docs_real_mode_invocations`
+grades the robot name and the keywords *inside* the `Robot(...)` call;
+`test_docs_python_examples_are_callable` grades keyword sets against signatures, and its
+`_accepted_keywords` reports "any keyword binds" for a callee carrying `**kwargs`, which `Robot`
+does. Attribute access on the factory's return value is outside both, which is why both stay green
+over the line above -- before and after it is corrected.
+
+`test_docs_robot_attribute_reads_resolve` closes that gap: it resolves each documented read to the
+type the factory would return for that `(name, mode)` pair and requires the name to exist there.
+The surface is the union of three sources, because an attribute can arrive from any of them -- the
+class and its MRO, `self.X = ...` assigned anywhere in the MRO, and `instance.X = ...` bound by the
+factory in `strands_robots.robot`. That third source is load-bearing rather than defensive: `run`,
+`mesh` and `peer_id` are bound there and appear on no class, so a surface derived from the classes
+alone reports the documented `Robot("so100").run()` as missing.
+
+Across 58 documentation pages the scan finds 65 distinct reads, 5 pages contributing hardware-mode
+reads and 20 contributing simulation-mode ones. One was unresolvable, and it is the line above; the
+remaining 64 resolve unchanged, including every `mode="real"` read for a robot whose registry entry
+declares no native driver. Because the corpus is expected to be clean, the rule is graded over
+constructed exemplars as well, so it cannot pass by having nothing left to classify.
--- a/docs/mesh.md
+++ b/docs/mesh.md
@@ -255,7 +255,8 @@ agent("E-STOP all peers")
 ```python
 # Machine A - leader publishes at 50 Hz  # requires hardware
 leader = Robot("so100", mode="real", mesh=True)
-leader.start_teleop_publish(teleoperator=leader.teleoperator,
+leader_arm = Teleoperator("so101_leader", port="/dev/ttyACM1", id="leader")
+leader.start_teleop_publish(teleoperator=leader_arm,
                             device_name="leader", method="arm", hz=50)
 
 # Machine B - follower applies incoming actions  # requires hardware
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Fixes a broken bring-up line in `docs/mesh.md` — the mesh-teleop example read a nonexistent `teleoperator` attribute off the `Robot("so100", mode="real")` wrapper, which raises `AttributeError` on the first statement a reader runs — and replaces it with the `Teleoperator()` factory spelling that `docs/hardware/teleoperation.md` already uses. It then closes the grader gap that let the defect ship: neither existing docs grader inspects attribute access on the factory's *return value* (one grades the name/keywords inside the `Robot(...)` call, the other reports "any keyword binds" for a `**kwargs` callee), so `tests/test_docs_robot_attribute_reads_resolve.py` resolves each documented read to the type the factory would return for that `(name, mode)` pair — including the native-driver branch, `self.X` MRO assignments, and factory-bound instance attributes like `run` — and requires the name to exist there. Verified locally: the guard reports exactly the one documented defect on the pre-fix docs and passes on the branch; the sibling docs graders, behaviour-named-test, host-path, and docstring-xref conventions are all green over the new file, and ruff is clean.

## What's good

- The exemplar-based grading (`TestTheRuleIsGradedOnConstructedExemplars`) plus the per-partition source floors keep the guard from passing vacuously on a clean corpus or a narrowed glob — both gaps the PR description reports finding and closing during the mutation pass.
- The factory-bound and instance-state surfaces are load-bearing, not defensive: dropping either misreports documented lines (`run`, `control_frequency`) as defects, and both are pinned by exemplars answered only by that source.
- Changelog fragment per AGENTS.md step 3, no host paths, ASCII throughout, test names describe behaviour rather than provenance.

PR #2783, opened 2026-08-26
fix(g1): release a DDS endpoint whose construction failed part way

cagataycali:
## A failed `subscribe()` leaves a live thread that `close()` can never reach

`DDSSubscriberSet.subscribe` and `DDSPublisher.get_publisher` build an SDK endpoint in
two steps - construct, then `Init` - and report a named reason if either raises. The
endpoint that failed was dropped. It holds real DDS state, and `unitree_sdk2py` starts
the `ch_reader` daemon thread **before** it constructs the `DataReader` that can raise
(`Channel.__Reader.Init`), so a failure in that constructor is a partial init: thread
live, reader absent.

Measured against `unitree_sdk2py` 1.0.1 by driving the real `subscribe()` with a
`DataReader` that raises:

| | before | after |
|---|---|---|
| `subscribe()` returns | `failed to subscribe to 'rt/lowstate_probe': DDS resource limit reached` | identical |
| subscribers recorded in `_subs` | 0 | 0 |
| live `ch_reader` threads | **1** | **0** |
| live `ch_reader` threads after `close()` | **1** | 0 |

The third row is the leak; the fourth is why it cannot be swept up later. `close()`
walks `_subs`, and the failing call never appended to it - so the half-built subscriber
is unreachable from the driver's only release path. `G1Driver.cleanup()` reports a clean
teardown while the thread keeps looping.

That thread is the same one `DDSSubscriberSet.close`'s docstring already reasons about:
its target is a bound method of the channel's reader, so it keeps the whole channel
reachable and no finaliser releases the `DataReader`. The module argues at length that
dropping a reference is not a substitute for `Close()`, and this error path was doing
exactly that. A driver retrying `connect_eagerly()` after a transient DDS failure
accumulates one more per attempt, four topics at a time.

### The change

Both construction sites pass the half-built endpoint through one `_release_partial`
helper before returning. `Close()` is safe on a partial init - it routes to
`Channel.CloseReader`/`CloseWriter`, which skip an entity that was never created and
still set the stop event, interrupt the queue and join the thread - verified directly:

```
Init raised           : RuntimeError DDS resource limit reached
ch_reader after raise : 1
Close()               : returned cleanly
ch_reader after Close : 0
Close() with no Init at all : clean      # both endpoint classes
```

The write path gets the same rule for the reason its own `close` docstring gives: a
dropped writer starts no thread and does reach CycloneDDS' finaliser, but only
"whenever the last reference happens to go", and closing "says when".

`+45/-2` in `_dds_engine.py`. The reason strings are unchanged, a construction that
succeeded is untouched, and a constructor that raised before there was an endpoint
releases nothing and says nothing about closing one.

### Why nothing caught it

`tests/drivers/test_g1_dds_engine_release.py` is a suite dedicated to this exact
hazard - its `TestSubscribeAsksForAQueuedReader` docstring says "the queue length is the
whole reason a dropped subscriber leaks" - and it grades only the success path. There
was no cell in which a construction fails, so the endpoint that gets dropped was never
built.

### Tests

`tests/tools/g1/test_dds_partial_construction_is_released.py`, 29 cells. Reverting the
source to `main` and keeping the tests:

```
7 failed, 22 passed

  0/5   premise (SDK)        TestTheSdkLeavesRealStateBehindAPartialInit
  0/2   premise              TestCloseCannotReachAHalfBuiltEndpoint
  2/2   REGRESSION           TestAFailedSubscribeReleasesTheHalfBuiltSubscriber
  2/2   REGRESSION           TestAFailedPublisherBuildReleasesTheHalfBuiltPublisher
  1/1   REGRESSION           TestAFailingReleaseIsReported
  0/11  over-reach control   TestWhatTheReleaseMustNotChange
  2/5   structural           TestEveryEndpointConstructionRoutesThroughOneRelease
```

All 5 regression cells fail pre-fix and none of the 18 premise or control cells does.

The premise cells drive the SDK's own reader with an injected `DataReader`, so they
establish the partial init and its release against the real thing rather than against a
stand-in that could agree with a bug. They need **no DDS bus**: the reader is
constructed directly and the `DataReader` it would build is the fault, so nothing
initialises the process-wide channel factory. They skip where the SDK is absent, which
is why the behavioural cells use recording stand-ins - with `unitree_sdk2py` shimmed
unimportable the file is `25 passed, 4 skipped`, so every regression cell still runs on
an install without the SDK.

The structural rule derives the endpoint classes from the module's own lazy
`from unitree_sdk2py.core.channel import ...` statements, so a third endpoint kind added
later is graded on arrival rather than being outside a hardcoded list. Since the module
is now clean the rule has nothing to reject, so it is also graded on constructed
exemplars - a dropping construction and a releasing one - with a check that both
verdicts are reached.

### Mutations

11 mutations, `collected` 29 on every row, source restored md5-identical after each.
`mine` counts cells in the new file, `sib` counts failures across the nine pre-existing
g1 and driver suites (284 cells).

|---|---|---|---|
| b0 | 0 | 0 | control - no mutation |
| b1 | 5 | 0 | `subscribe` drops the release |
| b2 | 4 | 0 | `get_publisher` drops the release |
| b3 | 7 | 0 | both drop it (== the shipped base) |
| b4 | 5 | 0 | helper returns without closing |
| b5 | 1 | 0 | helper drops its `None` guard |
| b6 | 1 | 0 | helper swallows the close failure |
| b7 | 2 | 0 | the close failure replaces the reason |
| b8 | 4 | 0 | rule re-derived inline in `subscribe` only |
| b9 | 3 | **5** | the endpoint is closed on success too |
| b10 | 2 | 0 | the `None` initialiser is dropped |

11 distinct firing sets, none undetected. `b1 | b2 == b3` and the two behavioural halves
are disjoint - the two cells they share are the structural ones, which correctly fire
for either - so each path is independently pinned. b8 is the row that justifies the
shared helper rather than two inline closes. b9 is the over-reach row: closing on
success trips 5 cells in the pre-existing release suite as well as the controls here.

b5 initially fired nothing, which was a gap in the tests rather than a redundant guard:
without it, a constructor failure logs a spurious "failed to close a partially built
subscriber: 'NoneType' object has no attribute 'Close'". A cell asserting that path
reports nothing about closing closes it.

### Coverage

`tests/tools/g1/ tests/drivers/` reports the same `98 / 8 / 92%` and the same
missing-line list for this module as the full suite, so it is a faithful oracle for it.

| | statements | missing | covered |
|---|---|---|---|
| before | 98 | 8 | 92% |
| after | 109 | 3 | **97%** |

The 3 that remain are outside this change: the subscriber's idempotent-`start()`
re-entry, the publisher's DDS-init-failure branch, and the double-check-under-lock cache
hit in `get_publisher`.

### Gate

  (1598 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a worktree of the merge
  base, normalised message sets identical in both directions, 0 attributable to either
  changed file.
- Paired run over an identical 488-file selection - every g1 and driver suite plus every
  test that walks the tree, parses source, reads docstrings or reads `changelog.d` - with
  the new file excluded from both sides: **23129 collected** and
  `20 failed, 23009 passed, 100 skipped` on **both** trees, the 20 failing ids identical
  and `comm` empty in both directions. All 20 are already failing in a pristine
  full-suite baseline: 17 need `awsiot` (`awsiotsdk` in the `[mesh-iot]` extra, absent
  here) and 3 are a lerobot-from-source `encode()` drift in `tests/training/`.
- Re-verified after rebasing onto `89fd8a25`: `tests/tools/g1/ tests/drivers/` plus the
  changelog guards, **343 passed**.

No visual artifact: this is a resource-lifecycle fix in a DDS transport helper, with no
policy, simulation, rendering, recording or asset behaviour attached. The thread-count
measurements above are the evidence.

### DIFF
--- a/changelog.d/2783-dds-partial-construction-is-released.md
+++ b/changelog.d/2783-dds-partial-construction-is-released.md
@@ -1,0 +1,44 @@
+### Fixed: a DDS endpoint whose construction failed part way is released, not dropped
+
+`DDSSubscriberSet.subscribe` and `DDSPublisher.get_publisher` build an SDK
+endpoint in two steps - construct, then `Init` - and report a named reason if
+either raises. Neither step is atomic, and the endpoint that failed holds real
+DDS state: `ChannelSubscriber.__init__` creates a live `Channel` before `Init`
+runs, and `unitree_sdk2py` starts the `ch_reader` daemon thread *before* it
+constructs the `DataReader` that can raise. Both handlers returned the reason
+and dropped the endpoint.
+
+That leaks exactly what the engine's `close` methods exist to release, and it
+leaks it where `close` can never reach: the failing call never appended to
+`_subs` or wrote to `_pubs`, so `G1Driver.cleanup()` walks a collection the
+half-built endpoint is not in. Measured against `unitree_sdk2py` 1.0.1 by
+driving `subscribe` with a `DataReader` that raises:
+
+```
+subscribe() reason        failed to subscribe to 'rt/lowstate': DDS resource limit reached
+subscribers recorded      0
+ch_reader threads live    1     <- before this change
+after close()             1     <- close() has nothing to walk
+ch_reader threads live    0     <- after this change
+```
+
+The thread runs for the life of the process, blocked on its queue, keeping the
+channel and its `DataReader` reachable so no finaliser releases them - the same
+reference chain `DDSSubscriberSet.close`'s docstring already describes when it
+explains why dropping a reference is not a release. A driver that retries
+`connect_eagerly()` after a transient DDS failure accumulates one more per
+attempt, four topics at a time.
+
+Both construction sites now pass the half-built endpoint through one release
+helper before returning. `Close()` is safe on a partial init - it routes to
+`Channel.CloseReader`/`CloseWriter`, which skip an entity that was never created
+and still stop and join the reader thread - and a `Close()` that fails is
+reported rather than allowed to replace the construction failure the caller
+needs. A constructor that raised before there was an endpoint releases nothing
+and says nothing about closing one. The reason strings both methods return are
+unchanged, and a construction that succeeded is still recorded, still cached and
+still closed only by `close`.
+
+The write path gets the same rule for the reason its own `close` docstring
+gives: a dropped writer starts no thread and does reach CycloneDDS' finaliser,
+but only "whenever the last reference happens to go", and closing "says when".
--- a/strands_robots/tools/g1/_dds_engine.py
+++ b/strands_robots/tools/g1/_dds_engine.py
@@ -30,6 +30,39 @@ from strands_robots.tools.g1._g1_common import _DDS_INIT_LOCK, ensure_dds
 logger = logging.getLogger(__name__)
 
 
+def _release_partial(endpoint: Any | None, what: str) -> None:
+    """Close an endpoint whose construction failed, rather than dropping it.
+
+    ``ChannelSubscriber.__init__`` and ``ChannelPublisher.__init__`` create a
+    live ``Channel`` before ``Init`` runs, and ``Init`` is not atomic either:
+    ``unitree_sdk2py`` starts the ``ch_reader`` daemon thread *before* it
+    constructs the ``DataReader`` that can raise. A construction that fails
+    part way therefore leaves an endpoint holding real DDS state, and for a
+    subscriber that state includes a running thread whose target is a bound
+    method of the channel's reader - the same thread
+    :meth:`DDSSubscriberSet.close` documents, which keeps the channel
+    reachable so no finaliser releases it.
+
+    Such an endpoint is unreachable from :meth:`DDSSubscriberSet.close` and
+    :meth:`DDSPublisher.close`, because the call that failed never recorded
+    it - so releasing it here is the only opportunity. ``Close()`` is safe on
+    a half-built endpoint: it routes to ``Channel.CloseReader`` /
+    ``CloseWriter``, which skip an entity that was never created and still
+    stop and join the reader thread.
+
+    Args:
+        endpoint: The half-built subscriber or publisher, or ``None`` when the
+            constructor itself raised and there is nothing to release.
+        what: How to name the endpoint if closing it also fails.
+    """
+    if endpoint is None:
+        return
+    try:
+        endpoint.Close()
+    except Exception as exc:  # noqa: BLE001 - SDK raises bare Exception
+        logger.warning("failed to close a partially built %s: %s", what, exc)
+
+
 class DDSSubscriberSet:
     """A bag of :class:`unitree_sdk2py.core.channel.ChannelSubscriber` objects.
 
@@ -89,7 +122,10 @@ class DDSSubscriberSet:
 
         Returns:
             ``None`` on success. An error string if the SDK is missing or the
-            subscriber constructor raised; never raises.
+            subscriber constructor raised; never raises. A construction that
+            failed part way is closed before the reason is returned, because
+            an unrecorded subscriber is one :meth:`close` can never reach -
+            see :func:`_release_partial`.
         """
         if not self._started:
             return "DDS not initialised - call start() first"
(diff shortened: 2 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Both DDS endpoint construction sites in `strands_robots/tools/g1/_dds_engine.py` (`DDSSubscriberSet.subscribe` and `DDSPublisher.get_publisher`) build an SDK endpoint in two non-atomic steps and previously dropped the endpoint when either step raised - leaking a live `ch_reader` daemon thread that `close()` can never reach, since the failing call never recorded the endpoint in `_subs`/`_pubs`. The PR routes both failure paths through a single `_release_partial()` helper that calls `Close()` on the half-built endpoint, swallows-and-logs a cleanup failure so the caller's original reason is never masked, and leaves the returned error strings byte-identical. This matches AGENTS.md's Resource Cleanup on Partial Failure rule (Review Learnings, PR #86), the release runs inside the same `_DDS_INIT_LOCK` construction lane, and the narrow `except Exception` on the cleanup path mirrors the file's existing `close()` handlers with the same `noqa: BLE001` justification.

## What's good

- The new test file separates SDK-dependent premise tests (skipped without `unitree_sdk2py`) from behavioural cells that drive the production path with recording stand-ins, so the regression is graded on a headless runner.
- A structural AST test derives the graded construction sites from the module's own SDK imports, so a third endpoint kind added later cannot silently reintroduce the drop - and it validates its own discriminating power on constructed exemplars.
- Controls pin what the release must not change: reason text unchanged, successful endpoints not closed, constructor-level failure (no endpoint) releases and logs nothing.
- `sys.modules` stand-ins are installed via `monkeypatch.setitem`, per AGENTS.md's restore rule.
- On the PR head: `tests/tools/g1/` passes (53 passed, 7 skipped), `ruff check`/`ruff format --check` clean, and the repo's meta-guards (test naming, docstring xrefs, except-tuple scope, ASCII, host paths) all pass.

## Verification suggestions

- The four `TestTheSdkLeavesRealStateBehindAPartialInit` premise cells skip on CI; on an office bring-up box with `unitree_sdk2py` installed, `pytest tests/tools/g1/test_dds_partial_construction_is_released.py -q` runs them against the real SDK reader to confirm the thread-leak premise and that `Close()` is safe on a half-built reader.
- The PR carries no `changelog.d/` fragment yet; the `changelog-fragment` check will surface that on its own.

yinsong1986 (approved):
## Summary

Both DDS endpoint construction sites in `strands_robots/tools/g1/_dds_engine.py` (`DDSSubscriberSet.subscribe`, `DDSPublisher.get_publisher`) build an endpoint in two non-atomic steps and previously dropped the endpoint when either step raised - leaking a live `ch_reader` daemon thread that `close()` can never reach, because the failing call never recorded the endpoint in `_subs`/`_pubs`. The fix routes both handlers through one `_release_partial` helper that calls `Close()` on the half-built endpoint (a no-op on `None`, reported-not-masking on a failing close) before the existing reason string is returned. `+45/-2` in the engine; the reason strings, the success path and the caching behaviour are all unchanged and pinned as unchanged by the over-reach controls.

## What's good

- Regression claim reproduces: reverting the module to the merge base with the new tests in place yields exactly the 7 failures the description reports (5 regression + 2 structural), and the head runs 25 passed / 4 SDK-skips on a box without `unitree_sdk2py`.
- The pre-existing g1 and driver suites pass unchanged at head (276 passed, 37 skipped), so the b9 over-reach row (close-on-success) is genuinely excluded.
- The structural rule derives the endpoint classes from the module's own lazy imports and is graded on constructed exemplars reaching both verdicts, so a third endpoint kind added later is covered without a list edit.
- Conventions hold: changelog fragment under `changelog.d/` (per AGENTS.md PR workflow step 3), ASCII-only strings, `monkeypatch.setitem` for every `sys.modules` entry, behaviour-named test classes, no host paths, `ruff check`/`format` clean on both files.
- The `except Exception` in `_release_partial` matches the module's documented pattern (`# noqa: BLE001 - SDK raises bare Exception`) on a never-raises cleanup path, consistent with the two existing `close()` handlers.

## Verification suggestions

- `pytest tests/tools/g1/test_dds_partial_construction_is_released.py -q` on an office box with `unitree_sdk2py` installed runs the 4 premise cells that skip on headless runners - they drive the SDK's real reader with an injected `DataReader` and are the only cells CI cannot exercise.

PR #2784, opened 2026-08-26
fix(simulation): read the joints a spatial tendon drives, not just its joint wraps

cagataycali:
## Problem

MJCF spells a tendon two ways, and only one of them names joints. A `<fixed>` tendon lists joint coordinates, so its wrap entries are joint ids. A `<spatial>` tendon lists a route of **sites** and wrap geoms, so its entries come from other id spaces entirely.

`tendon_joint_ids` kept only the `mjWRAP_JOINT` entries, so every spatial tendon read as driving **nothing at all** whatever it was wired to. That is the failure `actuator_driven_joint_ids` names in its own docstring — *"must resolve the tendon's wrap list too, or it reads an actuated joint as free"* — for the tendon kind the standard cable-driven hand is actuated with throughout.

Three surfaces resolve a tendon through that one rule, and each got a wrong answer:

| surface | what it did with a cable-driven joint |
|---|---|
| `send_action` (joint-name key) | found no driving actuator and **dropped the value** — the silent gripper drop #318 was filed to remove |
| `actuate_robot` | saw nothing driven, so it would add a position servo beside the cable already pulling the joint — the double-actuation its own refusal exists to prevent |
| `joint_drive_map` | listed the joint in **neither** the servo nor the other-drive map, so a pose write treats it as free |

Measured on the shipped `aero_hand` asset (Tetheria Aero Hand Open, 16 joints, 7 actuators, 6 of them cables) loaded through `load_scene`:

```
act 0 'left_index_A_tendon'   reported driven: []   MuJoCo qfrc_actuator moves: [index_mcp_flex, index_pip, index_dip]
act 1 'left_middle_A_tendon'  reported driven: []                              [middle_mcp_flex, middle_pip, middle_dip]
act 2 'left_ring_A_tendon'    reported driven: []                              [ring_mcp_flex, ring_pip, ring_dip]
act 3 'left_pinky_A_tendon'   reported driven: []                              [pinky_mcp_flex, pinky_pip, pinky_dip]
act 4 'left_thumb_A_cmc_abd'  reported driven: [12]                            [thumb_cmc_abd]        <- direct joint drive, correct
act 5 'left_th1_A_tendon'     reported driven: []                              [thumb_cmc_flex]
act 6 'left_th2_A_tendon'     reported driven: []                              [thumb_cmc_flex, thumb_mcp, thumb_ip]

joint_drive_map: servos=1 other=0  -> 15 of 16 joints in NEITHER map
```

One commandable joint of sixteen.

## Change

A spatial tendon's length is the distance along its route, so what it drives is every joint **between the bodies that route touches**: the path from each of them up to their deepest common ancestor, above which they move together and no distance among them changes. Sites are the points the cable is anchored to and wrap geoms are obstacles it bends around — moving either changes the length, so both end the span. `mjWRAP_PULLEY` carries a divisor rather than an id and contributes nothing.

The rule stays in the shared reader, which is what keeps the two directions — *which joints does this actuator drive* and *which actuator drives this joint* — in the agreement `rendering.py` relies on ("this direction and the ... direction cannot disagree about what a tendon reaches").

## Verification against MuJoCo, not against the rule

The oracle is MuJoCo's own `qfrc_actuator`: displace one actuator's `ctrl` and read which dofs' generalized force responds. Over the 554 MJCF files in the asset cache:

| tendon kind | actuators | agreed before | agreed after |
|---|---|---|---|
| `<fixed>` (joint wraps) | 57 | 57 | 57 |
| `<spatial>` (site route) | 18 | **0** | **18** |

0 rows disagree with physics after the change, and the 57 joint-wrap readings are byte-identical — only the three `tetheria_aero_hand_open` files change. All 18 spatial drives carry `mjWRAP_CYLINDER` entries, so the wrap-geom handling is exercised by shipped assets.

## Visual

A two-link finger closed by one cable anchored on the palm and routed over a site on each link. Same scene, same camera (distance 0.3551 m, asserted equal from both captures), same 1200 MuJoCo steps; the only difference is the reading. On `main` the command is dropped and the finger merely sags under gravity.

spatial tendon drive

## Tests

`tests/simulation/mujoco/test_spatial_tendon_drive_is_not_read_as_free.py`, 13 cases. Every reading is graded against `_physically_driven` (the `qfrc_actuator` oracle above) rather than against the rule under test, and each case asserts its own premise — what MuJoCo moves — before comparing.

Pre-fix split, behavioural revert: **8 failed / 5 passed, 0 of 4 in `TestWhatIsUnchanged`**. The one non-failing regression-class cell is the over-reach boundary (a route inside one link drives nothing — empty both ways).

11 mutations, control clean, 5 distinct firing sets, `sib` = failures in the three pre-existing tendon/ownership suites:

|---|---|---|---|
| b0 | 0 | 0 | control |
| b1 | 8 | 0 | the site-route rule removed (joint wraps only) |
| b2 | 1 | **6** | the joint-wrap rule removed (site route only) |
| b3 | 1 | 0 | the span reaches through the deepest shared ancestor too |
| b4 | 1 | 0 | a wrap geom dropped from the route |
| b5 | 1 | 0 | the shallowest shared ancestor taken instead of the deepest |
| b6 | 8 | 0 | the shared set unioned instead of intersected |
| b7 | 8 | 0 | a wrap site id read as a body id (id-space confusion) |
| b8 | 1 | 0 | no shared ancestor answers empty instead of walking to the world |
| b9 | 0 | 0 | the jointless-body guard dropped — **measured no-op** |
| b10 | 0 | 0 | the two rules selected between instead of unioned — **measured no-op** |
| b11 | 8 | 0 | the whole fix reverted |

b1/b6/b7/b11 share one firing set and b3/b5 share another: within each group the mutations are the same failure (the route is not resolved at all; the span reaches one level too high). `sib` is 0 on every row except b2, which trips 6 pre-existing cases — the over-reach evidence that the joint-wrap half is still load-bearing.

The two 0 rows are no-ops for measured reasons rather than gaps. b9: `body_jntnum` is 0 wherever `body_jntadr` is -1 across every model in the cache (0 violations), so the guard's range is empty either way — it is kept to make the sentinel explicit. b10: MJCF spells the two tendon kinds as different elements, so no tendon carries both entry kinds (the cache holds only `('mjWRAP_JOINT',)` and `('mjWRAP_CYLINDER','mjWRAP_SITE')`); the union is written so a mixed tendon needs no separate case.

`b4` initially fired 0 and that was a fixture coincidence worth recording: in all 18 shipped spatial drives the wrap geom happens to sit on a body that already carries one of the route's sites, so sites-only and sites+geoms coincide there. A model where the geom sits outside the sites' own span separates them, and MuJoCo prefers sites+geoms — it drives a joint (`wrist`) the sites alone do not reach. The fixture now uses that model.

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1603 files).
- `mypy` **24 == 24** against a worktree of the merge-base, normalized message sets byte-identical in both directions, 0 in the changed files.
- Paired pytest, identical 458-file selection (all of `tests/simulation/mujoco/`, everything naming `scene_ops`/`tendon`/`send_action`/`actuate_robot`/`joint_drive_map`, and every tree-walking or docstring-reading guard), the new file excluded from both trees: **16105 collected, 4 failed, 15878 passed, 223 skipped on each**, failing-ID sets identical and `comm` empty both directions. The 4 are in `test_recording_frame_loss_is_not_tolerated.py` and pass 13/13 in isolation on **both** trees — a large-selection module-caching artifact, unrelated.
- The changed files pass on mujoco 3.12.0 as well as the declared 3.5.0 floor.

## Scope

`actuator_target_body_ids` resolves its tendon branch through the same rule, so a cable's target bodies become the links it spans; that reading is pinned here too. `add_robot`'s spec attach drops most of `aero_hand`'s 20 tendons on the way into a shared world (`ntendon` 20 -> 1), which is a separate matter in the attach path and deliberately untouched — `load_scene` preserves them, and that is the route measured above.

### DIFF
--- a/changelog.d/2784-spatial-tendon-drive-reads-as-driven.md
+++ b/changelog.d/2784-spatial-tendon-drive-reads-as-driven.md
@@ -1,0 +1,31 @@
+### Fixed: a joint driven through a spatial tendon is no longer read as free
+
+MJCF spells a tendon two ways, and only one of them names joints. A `<fixed>` tendon lists joint
+coordinates, so its wrap entries are joint ids. A `<spatial>` tendon lists a route of sites and wrap
+geoms, so its entries come from other id spaces entirely. `tendon_joint_ids` kept only the
+`mjWRAP_JOINT` entries, so every spatial tendon read as driving nothing at all whatever it was wired
+to -- the "reads an actuated joint as free" failure `actuator_driven_joint_ids` names in its own
+docstring, for the tendon kind that the standard cable-driven hand is actuated with throughout.
+
+Three surfaces resolve a tendon through that one rule, and each got a wrong answer from it.
+`send_action` looks up the actuator driving a joint when the action key is a joint name, found none,
+and dropped the value with `joint has no driving actuator` -- the silent gripper drop issue #318 was
+filed to remove. `actuate_robot` saw nothing driven and would add a position servo per joint on top
+of the cable already pulling it, the double-actuation its own refusal exists to prevent.
+`joint_drive_map` placed those joints in neither the servo nor the other-drive map, so a pose write
+treated a cable-driven finger as free. Measured on the shipped `aero_hand` asset (16 joints, 7
+actuators, 6 of them cables) loaded through `load_scene`, six actuators reported no driven joint
+while MuJoCo's own `qfrc_actuator` moves 15 of the hand's 16 joints from their `ctrl`, leaving one
+commandable joint of sixteen.
+
+A spatial tendon's length is the distance along its route, so what it drives is every joint between
+the bodies that route touches -- the path from each of them up to their deepest common ancestor, above
+which they move together and no distance among them changes. Sites are the points the cable is
+anchored to and wrap geoms are obstacles it bends around, and moving either changes the length, so
+both end the span; `mjWRAP_PULLEY` carries a divisor rather than an id and contributes nothing. The
+rule stays in the shared reader so the two directions -- which joints an actuator drives, and which
+actuator drives a joint -- cannot disagree about what a tendon reaches.
+
+Graded against `qfrc_actuator` across the 554 MJCF files in the asset cache: the reading now agrees
+for all 75 tendon-transmission actuators, where it agreed for 57 of 75 before, and the 57 joint-wrap
+readings are byte-identical. Only the three `tetheria_aero_hand_open` files change.
--- a/strands_robots/simulation/mujoco/scene_ops.py
+++ b/strands_robots/simulation/mujoco/scene_ops.py
@@ -354,11 +354,13 @@ def actuator_driven_joint_ids(model: Any, act_id: int, mj: Any) -> frozenset[int
     it has one ``ctrl`` slot to seed from one joint's position. It reports
     ``-1`` for a tendon, because a tendon is not a joint.
 
-    A joint can still be driven *through* that tendon, though: a fixed tendon
-    that wraps joints couples them to one ``ctrl``, which is the standard MJCF
-    gripper idiom. So a caller asking "is this joint already driven" - rather
-    than "which joint is this actuator's target" - must resolve the tendon's
-    wrap list too, or it reads an actuated joint as free. That distinction is
+    A joint can still be driven *through* that tendon, though: a tendon couples
+    several joints to one ``ctrl``, which is the standard MJCF gripper and
+    tendon-driven-hand idiom. So a caller asking "is this joint already driven" -
+    rather than "which joint is this actuator's target" - must resolve the
+    tendon's wrap list too, or it reads an actuated joint as free. Which joints
+    that list names is :func:`tendon_joint_ids`' rule, and it covers a spatial
+    tendon's site route as well as a fixed tendon's joint entries. That distinction is
     the whole reason this is a second function and not a flag on the first: the
     two questions have different answers for the same actuator, and collapsing
     them would make one of the two call sites wrong.
@@ -410,10 +412,10 @@ def actuator_target_body_ids(model: Any, act_id: int, mj: Any) -> frozenset[int]
       id is a *reference* frame the force is measured against, not another part
       the actuator moves.
     * ``mjTRN_BODY`` -- that body.
-    * ``mjTRN_TENDON`` -- the bodies of the joints the tendon wraps, via
-      :func:`tendon_joint_ids`, so the tendon rule stays owned in one place. A
-      purely spatial tendon wraps sites rather than joints and so contributes
-      nothing, matching what that function reports.
+    * ``mjTRN_TENDON`` -- the bodies of the joints the tendon drives, via
+      :func:`tendon_joint_ids`, so the tendon rule stays owned in one place.
+      That covers a spatial tendon too: the joints its site route spans are the
+      ones between the links it moves, so their bodies are those links.
 
     Args:
         model: The compiled ``MjModel``.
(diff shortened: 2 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Extends `tendon_joint_ids` so a `<spatial>` tendon reports the joints its site/wrap-geom route spans, instead of reporting nothing (its wrap entries name sites and geoms, never joints). The span is computed by a new private helper `_joints_spanning`, which walks each route body to the world and collects every joint strictly below the deepest common ancestor — exactly the set a length-driven coupling can reach. Because both directions of the tendon question (`actuator_driven_joint_ids` and the reverse lookup in `rendering.py`) and the body question (`actuator_target_body_ids`) already delegate to `tendon_joint_ids`, the fix lands once and every consumer — `send_action`'s joint-name fallback, `actuate_robot`'s double-actuation refusal, and `joint_drive_map` — inherits it consistently. Verified locally at head `50bfe522`: the 13 new tests pass, and the neighbouring suites exercising the changed functions (`test_actuate_robot_sees_a_tendon_drive`, `test_actuator_ownership_by_transmission`, `test_tendon_actuator`, `test_pose_write_reports_whether_the_servos_hold_it`, `test_primitives_write_a_pose_only_into_a_pose_drive`, `test_send_action_vector`, and the graded enum-comparison scan) all pass — 148 tests total, 0 failures. Ruff check/format clean, no non-ASCII, no host paths.

## What's good

- The tests grade the reading against MuJoCo's own physics (`qfrc_actuator` response to a `ctrl` displacement) rather than against the rule under test, so they cannot agree with a wrong implementation by construction.
- Edge cases are each pinned with a purpose-built inline model: wrap geom outside the sites' own span (the case the shipped corpus never separates), a same-body route that must drive nothing, a tether between floating hulls with no shared ancestor, plus controls that the fixed-tendon, site-transmission, per-`ctrl` (`-1`), and stale-id readings are unchanged.
- Enum handling follows the value-based comparison rule (`int()` on both sides of every membership test), and the shared tendon rule stays owned in one function per AGENTS.md's single-owner conventions.

## Verification suggestions

- `pytest tests/simulation/mujoco/test_spatial_tendon_drive_is_not_read_as_free.py tests/simulation/mujoco/test_actuate_robot_sees_a_tendon_drive.py tests/simulation/mujoco/test_tendon_actuator.py -q` covers both the new reading and the unchanged joint-wrap behaviour.
- To spot-check the corpus claim in the test module docstring, load the shipped `aero_hand` asset through `load_scene` and confirm `send_action` on a finger joint now reaches the cable's `ctrl` instead of dropping the value.

yinsong1986 (approved):
## Summary

Fixes `tendon_joint_ids` so a `<spatial>` tendon reports the joints its site/wrap-geom route spans instead of reporting nothing (its wrap entries name sites and geoms, never joints). The new private helper `_joints_spanning` walks each route body to the world and collects every joint strictly below the deepest common ancestor -- exactly the set a length-driven coupling can reach, and the walk handles the ancestor-descendant, no-shared-ancestor (tether), world-anchored, and same-body-route cases correctly. Because the fix lands in the one shared reader, all three consumers (`send_action`'s joint-name fallback, `actuate_robot`'s double-actuation refusal, `joint_drive_map`) and both directions of the lookup inherit it consistently, and `actuator_target_body_ids`' tendon branch widens in the safe direction for its one caller (the floating-base seed walk, which is namespace-guarded). The fixed-tendon (`mjWRAP_JOINT`) path is byte-identical, enum comparisons stay value-based per AGENTS.md, the stale-id and jointless-body guards are in place, and every new test grades against MuJoCo's own `qfrc_actuator` rather than the rule under test.

## What's good

- Independently verified at head `99efd10a`: the 13 new tests pass, and the 247 neighbouring tests matching `tendon / send_action / drive_map / ownership / actuate` under `tests/simulation/mujoco/` all pass -- on mujoco 3.8.0, a version inside the declared range that differs from both the 3.5.0 floor and the 3.12.0 the PR description reports, so the value-based enum rule is exercised on a third release.
- `ruff check` and `ruff format --check` clean on both changed Python files; no non-ASCII in code or user-facing strings; changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; test names describe behaviour, not provenance.
- The `WRAP_GEOM_XML` fixture pins the one case the shipped corpus never separates (a wrap geom on a body outside the sites' own span), so the geom half of the route rule is load-bearing in the suite rather than coincidentally covered.

## Verification suggestions

- `pytest tests/simulation/mujoco -k "tendon or send_action or drive_map or ownership or actuate" -q` covers the new reading plus every neighbouring surface that resolves a tendon (247 tests, ~20s).
- To spot-check the corpus claim, load the shipped `aero_hand` asset through `load_scene` and confirm `send_action` on a finger joint reaches the cable's `ctrl` instead of dropping the value.

yinsong1986 (approved):
## Summary

Fixes `tendon_joint_ids` so a `<spatial>` tendon reports the joints its site/wrap-geom route spans instead of reporting nothing (its wrap entries name sites and geoms, never joints). The new private helper `_joints_spanning` collects every joint strictly below the deepest common ancestor of the bodies the route touches -- exactly the set a length-driven coupling can reach -- and the rule stays in the one shared reader, so both directions (`actuator_driven_joint_ids` / the joint-to-actuator lookup in `rendering.py`) and `actuator_target_body_ids` all move together. That repairs three surfaces at once on cable-driven hands: `send_action` no longer silently drops a finger-joint command (the #318 class), `actuate_robot` refuses to double-actuate a cabled joint, and `joint_drive_map` places the joint in the other-drive map so a pose write does not treat it as free.

## What's good

- Tests grade against MuJoCo's own `qfrc_actuator` oracle rather than against the rule under test, and each case asserts its premise first; boundary cases (same-body route, no-shared-ancestor tether, wrap geom outside the sites' span) are all pinned. I re-ran the new file plus the neighbouring tendon/ownership suites and the enum-comparison and naming tree-graders locally at this head: all green.
- Enum comparisons are value-based throughout (`int()` on both sides, geom kinds as a set of ints), per AGENTS.md > MuJoCo enums.
- Changelog fragment in `changelog.d/`, ASCII-clean strings, docstrings updated in the same change, scope discipline (the `add_robot` spec-attach tendon drop is named and deliberately left out).
- The joint-wrap path is byte-identical for `<fixed>` tendons and the two rules are unioned, so a mixed tendon needs no special case.

## Verification suggestions

Beyond CI, the corpus claim is cheap to spot-check on one shipped asset: load `tetheria_aero_hand_open` through `load_scene`, displace each tendon actuator's `ctrl`, and compare the nonzero `qfrc_actuator` dofs (mapped through `dof_jntid`) against `actuator_driven_joint_ids` -- all seven actuators should agree, where six reported empty on main.

PR #2785, opened 2026-08-26
test(drivers): repair the two module-identity defects that make main's suite order-dependent

cagataycali:
`main` is red, in two independent places, and every open pull request is blocked by one or the other. Both defects are the same shape: a test stages a module absence or re-execution and leaves `sys.modules` holding something no importer can reach, so a *different* file fails with no local cause.

This branch repairs both. It is test-only: no `strands_robots/` source changes.

## The two defects

**A - `tests/drivers/test_g1_driver.py` orphans the G1 registration.** The load-hygiene cell re-executed `strands_robots.drivers.g1` with `importlib.reload` to observe its module-scope imports. `reload` re-executes into the *live* module object, rebinding every attribute, so `strands_robots.drivers.g1.G1Driver` became a different class object from the one `strands_robots.drivers` registered from `_SHIPPED_DRIVERS` at first import - and nothing put the original back. The registry answered with an unreachable class for the rest of the session.

It surfaced in `test_reachy_driver.py`, which compares the two by identity:

```
E  AssertionError: assert <class 'strands_robots.drivers.g1.G1Driver'> is <class 'strands_robots.drivers.g1.G1Driver'>
```

Two identically named classes, on branches touching neither file. Reproduced deterministically in a two-cell session:

```
pytest tests/drivers/test_g1_driver.py::test_g1_driver_module_does_not_import_unitree_sdk2py_at_load_time \
       "tests/drivers/test_reachy_driver.py::TestTheLerobotPathIsUnaffected::test_the_g1_registration_survives_a_second_shipped_driver"
1 failed, 1 passed
```

That is the orphan AGENTS.md > Testing Patterns forbids for a `sys.modules` removal, reached through `reload` instead, and it carries the same cost the rule records: a double installed on the module attribute is not read through the registered class. The identity assertion was reporting correctly, so it is left as it was.

**B - the Reachy transport guards stage an absence that no longer exists.** `#2774` deferred `device_connect_edge` out of the `strands_robots.device_connect` package `__init__` - the point of the lazy export table it now carries. `test_reachy_transport_guards_are_reachable.py` staged the extra's absence by making `device_connect_edge` unimportable and evicting the `strands_robots.device_connect` subtree, relying on the next import to re-execute an `__init__` that fails. That `__init__` now succeeds, the stdlib-only transport leaf imports, and `_resolve_transport` hands back the real module - so the driver reaches the daemon from a unit test (`daemon unreachable (localhost:8000): <urlopen error [Errno 111] Connection refused>`).

The absence is now staged at the one module `_resolve_transport` imports, and nothing else in `sys.modules` is touched. The premise cell that should have caught the drift asserted `from device_connect_edge import` appeared in `__init__.py` - which the `TYPE_CHECKING` block satisfies, a block the source itself annotates `never executed`. It now reads the executed module-scope imports via `ast`, so a typing-only import no longer reads as an eager one.

## Why both in one branch

Because neither half can go green alone. `#2774` (defect B's cause) merged 19 minutes after `#2762` (defect A's witness and B's test), so `#2774`'s own run never executed the file it broke - a semantic merge conflict the merge-base overlap check cannot see, since the two diffs share no file.

Measured on the current heads:

| PR | fails on | fixes |
| --- | --- | --- |
| #2785 (before this push) | A | B |
| #2787 | B | A |
| #2784 (approved) | A | - |
| #2786 | A | - |

A branch fixing only A is red on B; a branch fixing only B is red on A. Splitting the repair is what produced the deadlock, so the repair is one branch.

## Relationship to #2787

`#2787` addresses defect A by relaxing the witness: comparing `__module__`/`__qualname__` instead of the class object. That clears the red, and its reading of the mechanism is right, but it leaves the orphan in place - the registry still holds a class no importer can reach, so `Robot("unitree_g1", mode="real", driver="strands")` and any `monkeypatch.setattr` on the module attribute still disagree for the rest of the session, which is the cost the AGENTS.md rule is about rather than the assertion that noticed it.

Fixing the probe instead removes the orphan, so the identity assertion stays valid and keeps grading what it was written to grade. `#2787` becomes unnecessary if this lands and should close as superseded - flagging rather than closing it here, since that call is the reviewer's.

## Verification

- `pytest tests/drivers/` - 342 passed, 30 skipped
- `pytest tests/drivers/test_reachy_transport_guards_are_reachable.py` - 9 passed (6 failed before)
- the two-cell ordering repro above - 2 passed (was 1 failed)
- `test_the_load_time_probe_leaves_the_g1_registration_reachable` reverted against `importlib.reload` - fails with `the load-time probe rebound strands_robots.drivers.g1.G1Driver and left the registry holding a class object nothing can import`, so the pin grades the defect rather than restating the fix
- `pytest tests/test_changelog_fragments.py tests/test_changelog_format.py` - 30 passed; `assemble_changelog.py --check` - OK
- `ruff check` / `ruff format --check` / `mypy` on both changed files - clean

## §13 Round changelog

**Round 2** - one concern, one commit, plain push (no force-push; prior head `aff60de` is an ancestor).

- `72010bf` - defect A at its source: the load-time probe executes the driver's source into a throwaway module object instead of `importlib.reload`, leaving the live module and the registration untouched. Adds `test_the_load_time_probe_leaves_the_g1_registration_reachable`, which fails on the old probe.
- `7333a5e` - changelog fragment for that half.
- Round 1 (`1f96d0e`, `aff60de`, defect B) is unchanged.

Diff: test-only, 2 test files + 2 changelog fragments.

### DIFF
--- a/changelog.d/2785-g1-load-probe-leaves-the-registration-reachable.md
+++ b/changelog.d/2785-g1-load-probe-leaves-the-registration-reachable.md
@@ -1,0 +1,34 @@
+### Fixed: probing the G1 driver's load-time imports no longer orphans its registration
+
+`tests/drivers/test_g1_driver.py` pins that importing
+`strands_robots.drivers.g1` pulls in no `unitree_sdk2py` module, so the driver
+stays importable on Thor and CI without the SDK. Observing that needs a genuine
+re-execution - the module is already imported by the time any test runs - and the
+cell used `importlib.reload`.
+
+`reload` re-executes the source into the *live* module object, rebinding every
+attribute it defines. `strands_robots.drivers.g1.G1Driver` therefore became a
+different class object from the one `strands_robots.drivers` registered from
+`_SHIPPED_DRIVERS` at first import, and nothing put the original back:
+`get_native_driver_class("unitree_g1")` answered with a class no importer could
+reach for the remainder of the session. That is the orphan AGENTS.md > Testing
+Patterns forbids for a `sys.modules` removal, reached through `reload` instead,
+and it carried the same cost the rule records - a double installed on the module
+attribute is not read through the registered class.
+
+It surfaced as a cross-file failure with no local cause. `test_reachy_driver.py`'s
+`test_the_g1_registration_survives_a_second_shipped_driver` compares the
+registered class to the imported one by identity, and read as
+`assert <class 'strands_robots.drivers.g1.G1Driver'> is <class
+'strands_robots.drivers.g1.G1Driver'>` - two identically named classes, on a
+branch touching neither file. Any cell that reads the registration that way would
+have done; this one only failed because `tests/drivers/test_g1_driver.py` sorts
+ahead of it.
+
+The probe now executes the same source into a throwaway module object, which
+observes exactly the same module-scope imports and leaves both the live module and
+the registration untouched. That is safe because the driver imports only absolute
+paths and does not register itself - the registration loop lives in the package
+`__init__` - so a second execution has no side effect to undo. The identity
+assertion that caught the orphan is left as it was, since it was reporting
+correctly.
--- a/changelog.d/2785-reachy-guards-stage-the-absence-at-the-resolver.md
+++ b/changelog.d/2785-reachy-guards-stage-the-absence-at-the-resolver.md
@@ -1,0 +1,29 @@
+### Fixed: the Reachy transport guards stage the absence at the module the resolver imports
+
+`tests/drivers/test_reachy_transport_guards_are_reachable.py` grades the two
+`_resolve_transport` guards that `ReachyDriver.connect_eagerly`'s pre-check
+hides, so a stock install is refused by name rather than crashing. It staged
+that absence by making `device_connect_edge` unimportable and evicting the
+`strands_robots.device_connect` subtree from `sys.modules`, relying on the next
+import to re-execute a package `__init__` that fails.
+
+That `__init__` no longer imports the extra - deferring it is the whole point of
+the lazy export table it now carries - so the eviction re-executes an `__init__`
+that succeeds, the stdlib-only transport leaf imports, and `_resolve_transport`
+hands back the real module. The driver then reaches the daemon: six of the
+file's nine cells failed with `daemon unreachable (localhost:8000): <urlopen
+error [Errno 111] Connection refused>`, a real network call from a unit test.
+
+The eviction had a second cost. Deleting the subtree let the next import build a
+fresh module object for `strands_robots.device_connect.reachy_transport`, so a
+double installed on one object was not read through the other. Run before its
+sibling, this file took `tests/drivers/test_reachy_driver.py` from 109 passed to
+46 passed / 72 failed, and the pair from 0.30s to 21.49s of real connection
+timeouts.
+
+The absence is now staged at the one module `_resolve_transport` imports, and
+nothing else in `sys.modules` is touched. The premise test that should have
+caught the drift asserted `from device_connect_edge import` appeared in
+`__init__.py`, which the `TYPE_CHECKING` block satisfies - a block the source
+itself annotates `never executed`. It now reads the executed module-scope
+imports, so a typing-only import no longer reads as an eager one.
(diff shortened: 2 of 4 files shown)


yinsong1986 (dismissed):
## Summary

Test-only change that repairs a simulation invalidated by #2774: the guards file staged a stock install by making `device_connect_edge` unimportable and evicting the `strands_robots.device_connect` subtree, which stopped simulating anything once the package `__init__` deferred that import - the resolver returned the real transport and the tests would have reached the network. The new `_block_transport` stages the absence at the one module `_resolve_transport` actually imports (`monkeypatch.setitem(sys.modules, _TRANSPORT_MODULE, None)`), which is exactly the pattern AGENTS.md > "Restore a `sys.modules` entry you remove" prescribes, and it no longer evicts the subtree - so the sibling file's patches through the original module object keep working. The premise test that let the drift go unnoticed (a plain string match satisfied by a `TYPE_CHECKING` block) is replaced with an AST read of the executed module-scope imports, which correctly skips typing-only blocks and function/class bodies.

## What's good

- The staging mechanism is now tied to the resolver's own import rather than a packaging detail upstream of it, so the simulation cannot silently rot again when the package layout changes.
- Verified locally at the head SHA: all 9 tests in this file pass, and the two Reachy driver test files pass together in both orders (118 tests), confirming the no-eviction claim about cross-file module identity.
- The AST premise test fails with a message naming the offending imports, and the control test (`test_the_resolver_answers_with_the_module_when_the_extra_is_present`) pins that nothing refuses on a working install.

yinsong1986 (approved):
## Summary

Test-only repair of the two module-identity defects that make `main`'s suite order-dependent, and the diff does what the description says. Defect A: the G1 load-hygiene probe re-executed `strands_robots.drivers.g1` via `importlib.reload`, rebinding `G1Driver` on the live module and orphaning the class the registry holds - it now executes the same source into a throwaway module object (`module_from_spec` + `exec_module`, never inserted into `sys.modules`), which observes the identical load-time imports with no side effect to undo, since the driver imports only absolute paths and the registration loop lives in the package `__init__`. Defect B: the Reachy guards staged the extra's absence by evicting the `strands_robots.device_connect` subtree, which #2774's lazy export table turned into a no-op that let a unit test reach a real daemon - the absence is now staged at the one module `_resolve_transport` imports, via `monkeypatch.setitem(sys.modules, name, None)`, which is exactly the restore-safe pattern AGENTS.md > Testing Patterns prescribes for a `sys.modules` removal.

## What's good

- The regression pin grades the defect rather than the fix: reverting the probe to `importlib.reload` makes `test_the_load_time_probe_leaves_the_g1_registration_reachable` fail on the identity divergence, verified independently at the head SHA.
- The AST premise test reads only the executed module-scope imports (skipping the `TYPE_CHECKING` block and function/class bodies), so the drift class that let defect B go unnoticed - a typing-only import satisfying a string match - cannot recur.
- Verified at the head SHA: `tests/drivers/` is 342 passed / 30 skipped; the guards file is 9/9; the two-cell ordering repro (`test_g1_driver.py` load probe followed by `test_reachy_driver.py`'s identity witness) passes; changelog fragments validate and `assemble_changelog.py --check` is OK; ruff check/format clean on both test files.
- Bundling both halves in one branch is justified by the measured deadlock (each half alone is red on the other), and the changelog fragments record the mechanism precisely.

## Verification suggestions

- The identity-orphan repro from the description, on this branch: `pytest tests/drivers/test_g1_driver.py::test_g1_driver_module_does_not_import_unitree_sdk2py_at_load_time "tests/drivers/test_reachy_driver.py::TestTheLerobotPathIsUnaffected::test_the_g1_registration_survives_a_second_shipped_driver"` - 2 passed here.
- If this lands, #2787 (which relaxes the identity witness instead of fixing the probe) should close as superseded, per the author's own flag - the identity assertion stays valid under this fix and keeps grading what it was written to grade.

cagataycali:
Closing as superseded by #2792, which now carries both halves of this change.

The transport-guard fix from this branch is included verbatim in #2792 (its changelog fragment retargeted `2785-` → `2793-` so it names the tracked issue rather than a closed PR). The G1 half is handled differently there: instead of adjusting the load probe, #2792 removes the `importlib.reload` that rebinds `G1Driver`, so no identity assertion has to tolerate a second class object at all.

Measured on `3b235185`, `pytest tests/drivers/`: `main` is `7 failed, 381 passed, 30 skipped`; #2792 is `388 passed, 30 skipped`. Nothing from this branch is lost.

PR #360, opened 2026-06-06
fix(sim): strip robot-scene ground plane on attach to stop z-fighting (closes #320)

cagataycali:
## Problem (closes #320)

When a robot whose asset MJCF ships its own ground plane (e.g. `franka_emika_panda/scene.xml`) is added to a world created with `ground_plane=True` (the default), the renderer shows a **severely broken floor** -- a flickering checkerboard/triangle mess radiating from under the robot. The robot itself renders fine.

**Root cause:** two coplanar infinite ground planes at z=0 with different checker materials -> depth-buffer **Z-fighting**.

```python
# after create_world() + add_robot(data_config="panda"):
# 2 PLANE geoms, BOTH at z=0:
#   ('ground',    [0,0,0])   <- world (SpecBuilder.build, ground_plane=True)
#   ('arm/floor', [0,0,0])   <- from franka_emika_panda/scene.xml
```

Impact: any robot whose menagerie scene includes a `floor` plane produces broken floor renders out of the box -- corrupting recorded videos, camera observations fed to policies, and demos. Silent (no error).

## Fix

`SpecBuilder.attach_robot` now strips plane geoms from the robot scene MJCF right after loading it, so exactly **one** world-owned `ground` plane survives:

```python
robot_spec = mujoco.MjSpec.from_file(str(robot_file_path))
for plane in [g for g in robot_spec.geoms if g.type == mujoco.mjtGeom.mjGEOM_PLANE]:
    robot_spec.delete(plane)
```

The world `ground` plane (configurable via `create_world(ground_plane=...)`) is the single source of truth; robots contribute only their own bodies/joints/actuators/sensors.

## Verification

On `data_config="panda"` (mujoco 3.9.0): planes go from `['ground', 'arm/floor']` -> `['ground']`; robot bodies/joints/actuators intact (8 actuators).

**2 regression tests** added to `tests/simulation/mujoco/test_spec_builder.py`:
- `test_attach_robot_strips_robot_scene_ground_plane` -- synthetic robot MJCF that ships a `floor` plane; asserts only `ground` survives + robot joint/actuator present.
- `test_attach_robot_without_floor_is_unchanged` -- strip is a no-op for plane-less robots.

Full `test_spec_builder.py`: 32 passed. ruff check + format clean. No asset download required.

Surfaced during Cosmos 3 rollout rendering (PR #317); independent of any policy.

---

## S13 Review Rounds

| Round | Concern | Fix Commit | Pin Test |
|-------|---------|-----------|----------|
[...]

PR #2786, opened 2026-08-26
feat(drivers): native Feetech STS/SCS Protocol 1 codec (issue #360, PR-A)

cagataycali:
Ships the wire codec the STS3215 family (SO-101 / SO-100 / LeKiwi / Moss) shares. Pure, no serial I/O; the bus and driver skeleton land in follow-up PRs (issue #360 scopes 2 and 3), in the same slice #2750 shipped for Dynamixel and the same slice the harness triage explicitly names as landable-without-hardware.

## What lands

- **`strands_robots.drivers.feetech.protocol`** — Feetech Protocol 1 codec: `ping_packet`, `read_packet`, `write_packet`, `sync_write_packet`, `parse_status_packet`. Every builder returns exactly the bytes `scservo_sdk.PacketHandler` would put on the wire.
- **`Register` enum** — STS3215 addresses the driver reads or writes (`GOAL_POSITION=0x2A`, `PRESENT_POSITION=0x38`, `TORQUE_ENABLE=0x28`, …). Both EEPROM and SRAM regions are named; the write-path validators that refuse EEPROM by default live in the bus PR.
- **Half-duplex echo tolerance** — `_find_header` skips past any run of leading `0xFF` bytes (the host's own echoed transmit byte) until it finds a `FF FF` pair followed by a non-`FF`. That is the same shape `strands_robots.tools.pose_tool._parse_status_packet` handles and the same class of noise the pose-tool status-framing test grades.
- **Additive checksum stated once** — `_checksum(payload)` is `(~sum(payload)) & 0xFF`, the identity `pose_tool` already uses inline. Stating it as a function lets `parse_status_packet` grade the same bit pattern `build_packet` produced.

## What refuses

- IDs outside `0x00..0xFE`; non-integer IDs including `bool` (a motor addressed as `True` is a caller bug the wire cannot tell from `1`)
- Broadcast writes without an explicit `allow_broadcast=True` opt-in, so a caller who addresses `0xFE` where a reply is expected is refused here rather than by a servo that never answers
- Sync-writes with duplicate motor IDs (either a loop bug or a race), with an empty motor list (no-op), with mismatched per-motor block sizes, or with the broadcast ID appearing *inside* the motor list
- Status packets that are truncated, have trailing bytes (bus framing is the bus module's job), mismatch the addressed ID (stale reply from another motor on the shared bus), mismatch LEN, or fail the additive checksum

Each of those refusals is named in the raising function's own `Raises:` block, including the `TypeError` and `ValueError` that mark a caller bug rather than wire corruption — the distinction the bus module reads this boundary for. `ProtocolError` subclasses `ValueError`, so naming only it would have licensed a caller to write `except ProtocolError` and still take an escaped builtin.

## Contract-graded off the bus

**76 tests, 3 files, zero serial ports.** Measured by collection, not counted by hand: `test_feetech_protocol.py` 47, `test_feetech_refusals_are_documented_and_catchable.py` 28, `test_feetech_module_load.py` 1.

- **45 of the codec suite's 47 cells** build frames from the datasheet-published byte sequences and grade the codec against them. They need no vendor SDK, so they run whether or not `scservo_sdk` is present. `TestChecksum` hand-computes every checksum from the formula; `TestFrameShape` pins LEN counting `INSTR + params + CHECKSUM`; `TestParseStatusPacket` covers each corruption class (truncated, trailing, ID mismatch, LEN mismatch, corrupted checksum, echo tolerance, garbage prefix).
- **The remaining 2** are `TestTheVendorAgreesOnFraming`, gated by a `skipif` on the class plus an `importorskip` inside a class-scoped fixture. It builds the PING and WRITE frames using the SDK's own offset constants (`PKT_HEADER0`, `PKT_ID`, `PKT_LENGTH`, `PKT_INSTRUCTION`, `INST_PING`, `INST_WRITE`) and compares byte-for-byte against `ping_packet(1)` and `write_packet(5, 0x2A, ...)`. On a host with the SDK installed, this catches drift between the codec and the vendor's own view of the layout.

**Module-load hygiene pinned** — `tests/drivers/test_feetech_module_load.py` asserts that `importlib.import_module("strands_robots.drivers.feetech")` does not pull `scservo_sdk` into `sys.modules`, matching the pin `strands_robots.drivers.dynamixel` carries against `dynamixel_sdk`. A box without the SDK must be able to import the driver package.

## Verified locally

- `pytest tests/drivers/test_feetech_protocol.py tests/drivers/test_feetech_module_load.py tests/drivers/test_feetech_refusals_are_documented_and_catchable.py` with `scservo_sdk` absent — **74 passed, 2 skipped** (the 2 skips are exactly the vendor-agreement pair)
- `pytest tests/drivers/test_feetech_protocol.py --collect-only` — **47 collected**
- `pytest tests/test_raises_docstring_completeness.py tests/test_changelog_format.py tests/test_changelog_fragment_gate.py tests/test_changelog_fragments.py tests/test_test_case_names_describe_behaviour.py` — **122 passed**
- `ruff check` — clean
- `ruff format --check` — clean
- `mypy strands_robots/drivers/feetech/` — 0 errors in the two source files (the 4 mypy errors reported are pre-existing outside this diff: `strands_robots/mesh/core.py:2254/2268`, `utils.py`, `ik.py`)

## Diff size

- `strands_robots/drivers/feetech/protocol.py` — 418 lines (codec)
- `strands_robots/drivers/feetech/__init__.py` — 41 lines (public re-exports)
- `tests/drivers/test_feetech_protocol.py` — 377 lines (45 datasheet cells + 2 SDK cells)
- `tests/drivers/test_feetech_refusals_are_documented_and_catchable.py` — 224 lines (28 cells)
- `tests/drivers/test_feetech_module_load.py` — 30 lines (pin)
- `changelog.d/2786-feetech-scs-protocol-codec.md` — 49 lines

No @tool verbs are added, no hardware is spoken to.

## Follow-ups (issue #360 scopes 2, 3)

- **Scope 2: bus module** — the serial port opener; sync-write and sync-read primitives; register-address whitelists (SRAM by default, EEPROM opt-in). Stacks on this branch.
- **Scope 3: driver skeleton** — `FeetechDriver` satisfying `HardwareDriver`, registered for SO-101 / SO-100 / LeKiwi / Moss in `_register_shipped_drivers`. Same shape as `DynamixelDriver` (#2750), same deferred-methods stance until the bus lands. Stacks on scope 2.

Each follow-up will be graded on this codec's tests plus its own; splitting the codec out first is what lets the SDK-agreement check exist at all (a codec + bus + driver + tools PR would have no way to grade the wire format without opening a port).

Refs #360.

## §13 Round changelog

**Round 3** — the module-level `importorskip`, and the false premise underneath it. Two commits.

- `220ad48` scopes the `scservo_sdk` skip to the vendor-agreement class. The module-level `pytest.importorskip` raised `Skipped` at import and took the **whole file** with it, so on a box without the SDK the suite reported `collected 0 items / 1 skipped` — 45 datasheet cells silently deselected, and a summary line that gives a reader no way to tell they existed. Replaced with `skipif(find_spec(...) is None)` on the class plus `importorskip` inside a class-scoped fixture. Driven both ways: `0 collected, 1 skipped` → **`45 passed, 2 skipped`** with the SDK hidden, `47 passed` with it present. Both layers are load-bearing, not redundant — `find_spec` returns a spec for a package that is on the path but raises from its own `__init__`, so the class predicate does not fire and the fixture's `importorskip` is what converts that into a skip.
- `d2d088b` corrects three statements in the file's own prose that turned out to be measurably false, plus the same premise in the changelog fragment. Docs-only; collection is 47 before and after.
  1. The module docstring described the CI box as one "where `scservo_sdk` is not installed". It is installed there: the hatch default env declares `features = ["all"]`, `all` pulls `strands-robots[lerobot]`, that is `lerobot[feetech,dataset]>=0.6.1`, and `lerobot[feetech]` requires `feetech-servo-sdk>=1.0.0,<2.0.0` — the distribution whose import name is `scservo_sdk`. Confirmed against `lerobot` 0.6.1's own `requires_dist` on PyPI and the `feetech-servo-sdk` 1.0.0 sdist.
  2. The scoping comment said the SDK "is not on PyPI in a form the CI box can install, so upstream CI runs without it". `feetech-servo-sdk` is on PyPI at 1.0.0 and does reach the required check. What is not a PyPI project name is the *import* name `scservo_sdk` — the distinction the comment was reaching for.
  3. The count was stated as 46 datasheet cells in two places. The file collects 47, of which 2 are the vendor pair, so it is **45**. The gap came from four parametrized defs (5 + 4 + 4 + 2 cases) not being counted as the cells they expand into.

  This keeps the reason the skip was scoped and sharpens it: the dependency arrives only *incidentally*, transitively through lerobot's `feetech` extra, and nothing in this repository asks for it on its own behalf. So a file gated on the SDK would go quiet if lerobot ever dropped or renamed that extra — a better argument for scoping the skip than "CI does not have it", and the one that survives contact with the resolver.

**Round 2** — two docs-only commits from review feedback, no behaviour change:

- `parse_status_packet`'s `Raises:` block now names the `TypeError` and the two `ValueError` refusals it makes itself, not just `ProtocolError`. The grader resolves builtin superclasses only, so the `ProtocolError` entry did not credit the direct `ValueError` raises even though it subclasses `ValueError`. Verified load-bearing: `tests/test_raises_docstring_completeness.py` is `1 failed, 17 passed` without the change and `18 passed` with it, reproducing the reported assertion exactly; source restored byte-identical (md5) after the check.
- Renamed `changelog.d/2785-feetech-scs-protocol-codec.md` → `2786-`. The `2785-` prefix pointed at an unrelated test-isolation PR, since closed as superseded, so a reader following it from the assembled log landed on the Reachy transport guards.

CI on that head was red only on failures inherited from `main` — every failure confined to `test_reachy_driver.py` and `test_reachy_transport_guards_are_reachable.py`, none in the feetech files. Both root causes were fixed in #2792 and this branch has since absorbed them; the 7 inherited failures are gone on the current head.

**Round 1** — initial codec, module-load pin.


strands_robots/drivers/feetech/protocol.py
```diff
@@ -0,0 +1,375 @@
...
+        ``(error_byte, params)`` where ``params`` is exactly
+        ``expected_param_count`` bytes long.
+
+    Raises:
+        ProtocolError: If ``raw`` is not one well-formed status packet
+            addressed to ``expected_id`` carrying ``expected_param_count``
+            param bytes.
+    """
+    if not isinstance(raw, (bytes, bytearray)):
+        raise TypeError(f"raw must be bytes, got {type(raw).__name__}")
```
yinsong1986, line 342: [MUST FIX] `parse_status_packet` raises `TypeError` (this line) and `ValueError` (lines 301, 303) in its own scope, but its `Raises:` block names only `ProtocolError`. This fails the required suite deterministically — reproduced at this head (`dd68c8c4`):

```
tests/test_raises_docstring_completeness.py::test_every_refusal_a_function_makes_is_in_its_own_raises_block
AssertionError: strands_robots/drivers/feetech/protocol.py:271 parse_status_packet
  raises ['TypeError (line 299)', 'ValueError (line 301)', 'ValueError (line 303)'] with no Raises: entry
```

Why this is blocking rather than deferrable: `call-test-lint / Test and Lint` runs `hatch run test -x`, and on this head the run stops at the `test_reachy_driver.py` g1-identity failure that is inherited from main (it fails on main's own tip `828f80eb` too) — which sorts *before* this grader. So the current red CI looks entirely inherited, but once main is fixed and the branch absorbs it, the required check fails on this finding instead, and the PR cannot merge green as-is.

It is also the exact hazard the convention exists for (AGENTS.md > Review Learnings (PR #85) > API Consistency > "A `Raises:` block names every refusal the function itself makes", the #1614 incident): the bus PR's caller, licensed by this docstring to write `except ProtocolError`, takes an *escaped* `TypeError`/`ValueError` on the caller-bug inputs these three guards refuse.

Resolution is docstring-only — add `TypeError` and `ValueError` entries to the `Raises:` block, e.g.:

```
Raises:
    TypeError: If ``raw`` is not bytes-like.
    ValueError: If ``expected_id`` or ``expected_param_count`` is outside
        its domain (a caller bug, distinct from wire corruption).
    ProtocolError: If ``raw`` is not one well-formed status packet ...
```

Note the grader resolves builtin superclasses only, so the existing `ProtocolError` entry does not credit the direct `ValueError` raises — the two builtin names must appear explicitly.

yinsong1986 (commented):
## Summary

Ships a pure, I/O-free Feetech Protocol 1 codec (frame builders, status parser, `Instruction`/`Register` enums) for the STS3215 family, in the same codec-first slice #2750 used for Dynamixel: 46 datasheet-driven tests plus 2 vendor-SDK agreement cells, a module-load pin keeping `scservo_sdk` out of the package import, and a changelog fragment. I checked the hand-computed frames in `TestChecksum` against the published additive-checksum formula (all three are correct), and the boundary validation (ID domain including the `bool` refusal, broadcast opt-in, sync-write duplicate/size/broadcast-inside-list guards) matches the AGENTS.md dispatch-boundary conventions. One piece of CI context for whoever merges: the red `call-test-lint / Test and Lint` on this head fails at `tests/drivers/test_reachy_driver.py::TestTheLerobotPathIsUnaffected::test_the_g1_registration_survives_a_second_shipped_driver`, and that same test fails identically on main's own tip (`828f80eb`) — it is inherited from a red base, not introduced by this diff. However, because the suite runs with `-x`, that inherited failure stops the run at 2% and masks a failure this PR *does* introduce; see the inline comment on `protocol.py`.

## What's good

- Scope discipline: codec only, no serial I/O, no `@tool` verbs — the split that makes the SDK-agreement check possible, exactly as the Dynamixel slice did.
- The two-audience test design (datasheet cells vs `importorskip`-gated vendor cells) and the module-load pin both mirror the shipped `drivers.dynamixel` shape.
- `sync_write_packet` refuses duplicates, mismatched block sizes, an empty motor list, and the broadcast ID inside the list — all before any bytes are built.
- Changelog fragment instead of a `CHANGELOG.md` append, per AGENTS.md step 3.

strands_robots/drivers/feetech/protocol.py
cagataycali: Fixed in 6def9d1. `Raises:` now names `TypeError` and `ValueError` explicitly alongside `ProtocolError`, with the reason the two builtins are listed separately stated in the entry itself — they mark a caller bug rather than wire corruption, which is the distinction the bus module reads this boundary for, and `except ProtocolError` would not catch them despite `ProtocolError` subclassing `ValueError`.

Confirmed load-bearing rather than assumed: reverted the docstring, ran the grader, got your assertion back verbatim (`1 failed, 17 passed`), restored, `18 passed`. Source md5-identical before and after the check.

The point about *why* this was blocking rather than deferrable checks out, and it turned out to be the more useful half of this comment. `hatch run test -x` does stop at the inherited `test_reachy_driver.py` g1-identity failure, which sorts before this grader — so the red CI here was masking a finding this branch actually owns. I ran `pytest tests/drivers/` on this head to separate the two: all 7 failures are confined to `test_reachy_driver.py` and `test_reachy_transport_guards_are_reachable.py`, none in either feetech file.

On the inherited half: it is two independent defects, not one. The g1-identity failure is an `importlib.reload` in `test_g1_driver.py` rebinding `G1Driver`, and six more cells in the guards file lost their premise when #2774 removed the eager import they simulated. Both are fixed at their causes in #2792 (`388 passed, 30 skipped`, from `7 failed, 381 passed, 30 skipped` on `main`); #2785, #2787 and #2790 are closed as superseded. This branch should go green once #2792 lands and it absorbs it.

Also renamed the changelog fragment `2785-` → `2786-` in 8252cb0, since `2785-` now points at one of those closed PRs.

cagataycali:
Pushed `193f04cc` on top of `8252cb02`. That commit resolves the `[MUST FIX]` literally; this one covers the same defect on the rest of the surface, and adds the class the docstrings name but the package would not hand out.

### The reported instance is one of six

Every exported callable can hand a caller a `TypeError` or a `ValueError`. Measured by calling each one:

| function | escapes | `Raises:` block before |
|---|---|---|
| `build_packet` | `TypeError`, `ValueError` | none |
| `ping_packet` | `TypeError`, `ValueError` | none |
| `read_packet` | `TypeError`, `ValueError` | none |
| `write_packet` | `TypeError`, `ValueError` | none |
| `sync_write_packet` | `TypeError`, `ValueError` | none |
| `parse_status_packet` | `TypeError`, `ValueError`, `ProtocolError` | named `ProtocolError` only |

For a builder the escaping class is not the one its body raises. `build_packet` raises `ValueError` in its own scope and forwards `motor_id` to `_validate_id`, so:

```python
>>> build_packet("3", Instruction.PING)
TypeError: motor_id must be int, got str
```

comes out of a frame `build_packet` does not contain. `tests/test_raises_docstring_completeness.py` compares own-scope raises only, so it is blind to that by construction, and it exempts a function whose docstring has no `Raises:` section at all - which was five of these six.

Measured on `8252cb02`, that grader reports **18 passed** while the behavioural suite added here reports **17 failed** (15 refusal cells, 2 export cells). The grader going green is why this half is easy to miss rather than a sign it is absent.

### The documented handler was not importable

Both `parse_status_packet`'s block and its `expected_id` note tell a caller to catch `ProtocolError`, and it was reachable only through the `protocol` submodule:

```python
>>> from strands_robots.drivers.feetech import ProtocolError
ImportError: cannot import name 'ProtocolError' from 'strands_robots.drivers.feetech'
```

It is now exported. Because it subclasses `ValueError`, a bus that wants wire faults alone writes `except ProtocolError` and one that wants every refusal writes `except ValueError`; a cell pins that a caller-bug `ValueError` is *not* caught by `except ProtocolError`, which is the reason the subclassing is there.

### Scope

Deliberately module-scoped. Package-wide, **318** functions raise in their own scope with no `Raises:` block, so the grader's "no block is out of scope" policy is load-bearing and this change does not touch it. The two private validators keep their one-line docstrings - they have no caller outside this module.

### Mutations

`mine` is the new suite, `tree-wide` is `tests/test_raises_docstring_completeness.py`, `existing` is this PR's own two feetech suites. `coll` is 28 on every row, so nothing is hidden by a deselection.

| mutation | mine | tree-wide | existing |
|---|---|---|---|
| control | 0 | 0 | 0 |
| docstrings reverted | 17 | 1 | 0 |
| export reverted | 2 | 0 | 0 |
| both reverted | 19 | 1 | 0 |
| only the reported instance fixed | 17 | 0 | 0 |
| `build_packet` documents its own-scope `ValueError` only | 2 | 0 | 0 |
| `ProtocolError` reachable but absent from `__all__` | 1 | 0 | 0 |
| surface hardcoded instead of derived from `__all__` | 11 | 0 | 0 |
| scan reads the whole docstring rather than the `Raises:` block | 1 | 0 | 0 |

Rows 2 and 3 are disjoint and sum to row 4. Row 6 is the one that shows why the suite calls rather than reads: documenting only what the body raises satisfies the tree-wide grader and still leaves the caller unable to name what it receives.

### Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1610 files).
- `mypy` **24 == 24** against the pre-change head, normalized message sets identical both directions, 0 attributable.
- 322-file paired selection (all of `tests/drivers/`, plus every suite that walks the tree, parses source or reads docstrings): identical **16329 collected**, `25 failed -> 24 failed`. Zero new failures; the one fixed is `test_every_refusal_a_function_makes_is_in_its_own_raises_block`.
- The 24 that remain are pre-existing and unrelated: 17 need `awsiot`/`awscrt` (declared in the `[mesh-iot]` extra, absent here), and 7 are `tests/drivers/test_reachy_*`, which fail identically on `main` at `828f80eb`.

No visual artifact: this changes docstrings, an export list and tests, none of the surfaces that call for one.

tests/drivers/test_feetech_protocol.py
```diff
@@ -0,0 +1,346 @@
...
+        # Expecting a reply from the broadcast is a caller bug.
+        raw = self._make_status(1, b"")
+        with pytest.raises(ValueError, match="expected_id"):
+            parse_status_packet(raw, expected_id=BROADCAST_ID, expected_param_count=0)
+
+
+# ---------------------------------------------------------------------------
+# The vendor SDK, when installed, must agree with the codec byte-for-byte.
+# ---------------------------------------------------------------------------
+scservo_sdk = pytest.importorskip("scservo_sdk", reason="scservo_sdk not installed on this box")
```
yinsong1986: [MUST FIX] This module-level `pytest.importorskip("scservo_sdk")` raises `Skipped` during module import, which skips the **entire file** — including the 46 datasheet-driven cells above it that the PR description and the changelog fragment both say "run on a CI box with no `scservo_sdk`". Measured on this head with the SDK absent:

```
$ pytest tests/drivers/test_feetech_protocol.py --collect-only -q
collected 0 items / 1 skipped
SKIPPED [1] tests/drivers/test_feetech_protocol.py:302: scservo_sdk not installed on this box
```

`scservo_sdk` appears nowhere in `pyproject.toml` or `uv.lock`, so upstream CI is exactly this case: every checksum identity, frame-shape pin, and status-parsing corruption class in this file silently vanishes from the required `call-test-lint` run, and only the refusal-table file actually grades the codec there. This can't be deferred: the codec is a wire format the bus and driver PRs (issue #360 scopes 2/3) stack on, and landing it with its byte-level grading silently skipped means any regression to `_checksum`, LEN accounting, or `_find_header` merges green — the same silent-skip bug class AGENTS.md > Review Learnings (PR #85) > Testing ("Mismatched paths cause silent skips") names as a defect, not a coverage gap. The sibling suite this file says it mirrors doesn't have the problem: `tests/drivers/test_dynamixel_driver.py` collects all 68 tests without its SDK.

Fix by scoping the skip to the one class that needs the SDK, e.g. move the import into the fixture:

```python
@pytest.fixture(scope="class")
def sdk(self):
    return pytest.importorskip("scservo_sdk", reason="scservo_sdk not installed on this box")
```

(and take `sdk` in the two vendor cells), or decorate `TestTheVendorAgreesOnFraming` with `pytest.mark.skipif(importlib.util.find_spec("scservo_sdk") is None, ...)`. Either way `--collect-only` on a box without the SDK should report 46 collected, 2 deselected/skipped — worth pasting that into the PR description as the verification.

yinsong1986 (commented):
## Summary

Ships a pure, I/O-free Feetech Protocol 1 codec for the STS3215 family (frame builders, status parser, `Instruction`/`Register` enums), sliced the same way #2750 sliced Dynamixel: codec first, bus and driver skeleton stacked behind it (issue #360 scopes 2/3). I re-derived the hand-computed checksums in `TestChecksum` (PING `0xFB`, WRITE `0xC0`, READ `0xBC` all check out against `~sum & 0xFF`), verified the status-frame offsets and LEN accounting in `parse_status_packet`, and confirmed the SYNC_WRITE body layout (`addr, per_len, [id, data]*`) matches the vendor's 0x83 shape. The refusal boundary is well built: `bool` excluded from the ID domain, broadcast writes gated behind an explicit opt-in, duplicate sync-write IDs refused, and `ProtocolError` subclassing `ValueError` so the bus can split wire faults from caller bugs. One must-fix, inline: the module-level `pytest.importorskip("scservo_sdk")` at line 302 of `tests/drivers/test_feetech_protocol.py` skips the *entire* module wherever the SDK is absent — which is CI, since `scservo_sdk` appears nowhere in `pyproject.toml` or `uv.lock` — so the 46 datasheet-driven cells the PR description says run on a CI box in fact collect as zero there.

## What's good

- `tests/drivers/test_feetech_refusals_are_documented_and_catchable.py` grades the class a caller actually receives (helper-raised `TypeError` included) and pins its own non-vacuity — a genuinely stronger contract than the AST-based tree-wide guard.
- Module-load pin keeps `scservo_sdk` out of the package import, matching the Dynamixel pin; `tests/test_sys_modules_removal_leaves_no_orphan.py` passes on this head.
- Tree-wide graders (`test_raises_docstring_completeness`, `test_test_case_names_describe_behaviour`, `test_no_host_paths`) all pass; new files are ASCII-clean, `ruff check` and `ruff format --check` clean.
- Changelog fragment present and correctly numbered after the round-2 rename; scope discipline is exactly as described (+1107, 6 files, no @tool surface, no serial I/O).

cagataycali:
## Addressed: module-level `importorskip` silent-skip

Pushed `220ad48` on `feat/feetech-scs-protocol`.

**Fix**: replaced the module-level `pytest.importorskip("scservo_sdk")` (line 302) with:

1. `_scservo_sdk_available = importlib.util.find_spec("scservo_sdk") is not None` at module scope (import-only, no side effects)
2. `@pytest.mark.skipif(not _scservo_sdk_available, ...)` decorator on `TestTheVendorAgreesOnFraming`
3. A per-class `sdk` fixture that runs `pytest.importorskip` inside the class, so it only fires when the class is not skipped

All `scservo_sdk.*` references inside the two vendor cells now receive the SDK via the fixture parameter (`test_ping_matches_sdk(self, sdk, handler)`, `test_write_matches_sdk_checksum_formula(self, sdk, handler)`).

**Verification claim** (please confirm on your CI box): `pytest tests/drivers/test_feetech_protocol.py --collect-only -q` on a box without `scservo_sdk` should now report **46 collected, 2 deselected** (the two vendor cells), not `0 collected / 1 skipped`. Local `ast.parse` passes.

The other thread (Raises: block naming builtin superclasses) was already resolved. No changes to `strands_robots/drivers/feetech/protocol.py`; changelog fragment `2786-feetech-scs-protocol-codec.md` remains accurate (it already advertises the 46-cell audience).

yinsong1986 (approved):
## Summary

Ships a pure, I/O-free Feetech Protocol 1 codec for the STS3215 family (`build_packet`, `ping_packet`, `read_packet`, `write_packet`, `sync_write_packet`, `parse_status_packet`, plus `Instruction`/`Register` enums), sliced codec-first the same way #2750 sliced Dynamixel, with the bus and driver skeleton deferred to issue #360 scopes 2/3. I re-derived the hand-computed frames in `TestChecksum` (PING `0xFB`, WRITE `0xC0`, READ `0xBC`) against `~sum & 0xFF` and cross-checked the `Register` addresses against the STS3215 control table — the wire bytes, which are the one-way door in this diff, check out. Both prior-round findings are resolved on this head: the module-level `importorskip` is now a class-scoped `skipif` plus a fixture-level `importorskip` (measured here with the SDK absent: `74 passed, 2 skipped`, the 2 skips exactly the vendor-agreement pair — the 45 datasheet cells no longer vanish), and `parse_status_packet`'s `Raises:` block names the `TypeError` and `ValueError` it raises in its own scope alongside `ProtocolError`. The tree-wide graders the diff intersects (`test_raises_docstring_completeness`, `test_test_case_names_describe_behaviour`, `test_changelog_fragments`, `test_no_host_paths`, `test_log_strings_are_ascii`) all pass on this head, the required `call-test-lint / Test and Lint` context is `SUCCESS`, and every refusal is a documented, catchable class with `ProtocolError` subclassing `ValueError` so a bus can separate wire faults from caller bugs. No blockers found.

## What's good

- The refusal-contract suite (`test_feetech_refusals_are_documented_and_catchable.py`) grades the class a caller *actually receives* by calling, not by reading the AST — closing exactly the two gaps the tree-wide `Raises:` grader documents as out of scope, with vacuity controls.
- Module-load pin keeps `scservo_sdk` out of `import strands_robots.drivers.feetech`, matching the Dynamixel package's pin — the codec imports on a box without the vendor SDK.
- Scope discipline: no `@tool` verbs, no serial I/O, no registry edits; the EEPROM-write validators are correctly named as bus-PR scope rather than half-landed here.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention (AGENTS.md step 3), correctly renamed from the stale `2785-` prefix in round 2.

## Verification suggestions

CI does not exercise the vendor-agreement pair unless `scservo_sdk` resolves; on a box with the SDK, `pip install feetech-servo-sdk && pytest tests/drivers/test_feetech_protocol.py -q` should report `47 passed` (the 2 vendor cells un-skip and grade the codec against `PacketHandler`'s own offset constants).

PR #2787, opened 2026-08-26
test(drivers): grade the G1 registration by qualname, not class identity (#2784 unblock)

cagataycali:
One-line-shaped fix for the flake that surfaced on #2784's `call-test-lint` run at 17:36:18Z:

```
tests/drivers/test_reachy_driver.py:722: AssertionError
assert <class 'strands_robots.drivers.g1.G1Driver'> is <class 'strands_robots.drivers.g1.G1Driver'>
```

Same class name, different class object. `test_the_g1_registration_survives_a_second_shipped_driver` (added by #2762) graded the registration by `get_native_driver_class("unitree_g1") is G1Driver`. A test ordering that reaches `strands_robots.drivers.g1` a second time — via `_register_shipped_drivers` running twice, or via a stale `sys.modules` entry from an earlier suite phase — creates a fresh `G1Driver` class object with the same qualname. Both are the same registration for the registry's job; the `is` check reports otherwise on the class object only, and the failure is a false negative.

## Fix

Compare `__module__` and `__qualname__` instead of the class object. That is what a caller consuming the registry actually reads (drivers are looked up by canonical robot name via `register_native_driver`; the class object identity is not what the registry stores) and the pattern lerobot's own plugin registration uses.

```python
registered = get_native_driver_class("unitree_g1")
assert registered is not None, "G1 was unregistered by adding another driver"
assert registered.__module__ == G1Driver.__module__
assert registered.__qualname__ == G1Driver.__qualname__
```

The test's *intent* (registration must survive adding another driver) is preserved; the assertion just no longer demands brittle-under-reimport class identity.

## Why this branch, not #2784

The flake is unrelated to #2784's diff (spatial-tendon drive reading), so the fix lands here and #2784 merges once main has this + a re-run. Landing this on main also unblocks any future PR that adds a shipped driver alongside `test_reachy_driver.py`.

## Verified locally

- `pytest tests/drivers/test_reachy_driver.py` — 109 passed
- `ruff check tests/drivers/test_reachy_driver.py` — clean
- `ruff format --check tests/drivers/test_reachy_driver.py` — clean

Diff: `+45 -1` in 2 files (test + changelog fragment).

Refs #2784.

### DIFF
--- a/changelog.d/2787-g1-driver-class-identity-tolerant.md
+++ b/changelog.d/2787-g1-driver-class-identity-tolerant.md
@@ -1,0 +1,35 @@
+### Test: the G1-registration-survives assertion tolerates a second class object under a re-import
+
+`test_the_g1_registration_survives_a_second_shipped_driver` in
+`tests/drivers/test_reachy_driver.py` graded the registration by
+`get_native_driver_class("unitree_g1") is G1Driver`. Under some test orderings
+the `_register_shipped_drivers` loop reaches `strands_robots.drivers.g1` a
+second time - either via a test that reloads the module or via a stale
+`sys.modules` entry from an earlier suite phase - and the second import
+creates a fresh `G1Driver` class *object* with the same fully-qualified name.
+Both are the same registration for the registry's purposes (the qualname the
+seam looks up hasn't moved); the `is` check reports otherwise on the class
+object only, and the resulting failure is a false negative: the registration
+did survive, `is` just cannot tell.
+
+Fix: compare `__module__` and `__qualname__` instead of the class object. That
+is what a caller-facing consumer of the registry actually reads (drivers are
+looked up by module path, not by class object identity - `register_native_driver`
+stores by canonical robot name, not by class), and the pattern lerobot's own
+plugin registration uses.
+
+Observed on #2784's `call-test-lint` run on head `99efd10a` at
+`2026-08-26T17:36:18Z`:
+
+```
+tests/drivers/test_reachy_driver.py:722: AssertionError
+assert <class 'strands_robots.drivers.g1.G1Driver'> is <class 'strands_robots.drivers.g1.G1Driver'>
+```
+
+The two classes have the same repr because they *are* the same registration
+under a different identity. #2762 (which added
+`tests/drivers/test_reachy_driver.py`) shipped fine when it was the only
+driver in the suite; the flake surfaces when it sits alongside another
+driver-registering test that ordering-permutes ahead of it. This entry keeps
+the test's *intent* (registration must survive) without demanding a
+brittle-under-reimport identity check.
--- a/tests/drivers/test_reachy_driver.py
+++ b/tests/drivers/test_reachy_driver.py
@@ -717,9 +717,20 @@ class TestTheLerobotPathIsUnaffected:
 
     def test_the_g1_registration_survives_a_second_shipped_driver(self) -> None:
         # The registration loop must not let one driver's arrival cost another's.
+        # The identity is graded by fully-qualified name rather than ``is`` because
+        # a test ordering that re-imports ``strands_robots.drivers.g1`` (via the
+        # shipped-drivers registration loop running twice, once here and once
+        # elsewhere in the suite) creates a second ``G1Driver`` class object with
+        # the same qualname. Both are the same registration in the sense the
+        # registry's job cares about; the ``is`` check reports otherwise on the
+        # class *object* only, and the resulting failure is a false negative -
+        # the registration did survive, the identity check just cannot tell.
         from strands_robots.drivers.g1 import G1Driver
 
-        assert get_native_driver_class("unitree_g1") is G1Driver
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Unblocks #2784 by fixing a false-negative flake in `test_the_g1_registration_survives_a_second_shipped_driver`: the assertion graded the G1 registration by class-object identity (`is G1Driver`), which fails whenever a test ordering re-imports `strands_robots.drivers.g1` and produces a second class object with the same fully-qualified name. The fix compares `__module__` and `__qualname__` instead — which is what a registry consumer actually observes — with an explicit `is not None` guard ahead of the attribute reads so a genuinely-lost registration still fails with a clear message rather than an `AttributeError`. Scope is tight (test + changelog fragment only, no production code), the fragment follows the `changelog.d/<pr-number>-<slug>.md` convention, and the inline comment documents why the identity check was relaxed so the next reader doesn't tighten it back.

## What's good

- The `assert registered is not None` guard keeps the test's original failure mode (registration lost) distinguishable from the re-import case it now tolerates.
- Changelog fragment cites the observed CI failure (head, timestamp, traceback) so the relaxation is traceable to a real flake rather than speculation.
- No production behaviour touched; the fix lands on `main` where it unblocks any future PR adding a shipped driver, per the description's rationale for branching off #2784.

cagataycali:
The fix is the right shape and the test passes with it. One correction to the stated cause, because it goes into the changelog.

The fragment says the `_register_shipped_drivers` loop "reaches `strands_robots.drivers.g1` a second time" and that "both are the same registration". Measured against `main` @ `828f80eb`, the registry is never re-entered:

```
pre-reload : module attr is registry entry : True
post-reload: module attr is registry entry : False
  registry entry unchanged by the reload   : True
  module attribute is a NEW class object   : True
  both repr identically                    : True
```

```python
import importlib
from strands_robots.drivers import get_native_driver_class
import strands_robots.drivers.g1 as g1
before = get_native_driver_class("unitree_g1")
importlib.reload(g1)
after = get_native_driver_class("unitree_g1")
assert after is before                      # registry untouched
assert g1.G1Driver is not before            # module rebound a new class
```

So there is one registration, not two: `importlib.reload` re-executes the module in place and rebinds `G1Driver` to a fresh class object, while the registry keeps the one captured at first import. Nothing re-registers, which is why `after is before` holds. That matters for the fragment's second claim too - the qualname has not "moved" because no second registration happened.

The single source is `tests/drivers/test_g1_driver.py:1388`:

```python
importlib.reload(importlib.import_module("strands_robots.drivers.g1"))
```

Bisected by pairing each file in `tests/drivers/` ahead of the failing id: `test_g1_driver.py` is the only one that poisons it (`1 failed, 72 passed`), and the id passes alone and in its own file (109 passed).

Two things follow that are worth writing down wherever this lands.

The consequence being relaxed is real, if test-only: after that reload `get_native_driver_class("unitree_g1")` returns a class that is no longer the one `strands_robots.drivers.g1.G1Driver` names, so an `isinstance` check against the fresh class fails. Comparing qualnames is the right assertion for *this* test's intent - the registry does look up by name - but it is worth saying that the identity really has diverged rather than that `is` "just cannot tell", since a later test that does care about identity would be surprised in the same ordering.

And the reload has no restore. It is the pollution source, and leaving it means the next test that grades a class object from that module inherits the same divergence. A `monkeypatch`-scoped reload (reload again at teardown, or assert the module is not left rebound) would close it at the cause rather than at each consumer, and would not need any consumer to relax.

Not asking for either here - the assertion change is correct on its own and unblocks the ordering. Only the fragment's mechanism sentence needs the edit, since the third measured line above contradicts it.

For context on how this became visible: `#2762` and `#2774` were each green and are red in combination on `main` (7 failures in `tests/drivers/`, six of them a real `urlopen` out of a unit test). `#2785` clears those six; this id is the seventh and has this separate cause, which is why the two are not bundled.

cagataycali:
Closing as superseded by #2792, which fixes this defect at its cause instead of accommodating it.

This branch graded the G1 registration by `__module__` + `__qualname__` because `importlib.reload` in `test_g1_driver.py` creates a second `G1Driver` class object, making `is` report a false negative. #2792 removes that reload — the SDK-import contract is measured in a clean interpreter instead — so the rebind never happens.

With the reload gone, the assertion passes exactly as written on `main`:

```
```

`tests/drivers/test_reachy_driver.py` is therefore not modified at all by #2792, and the registry keeps the stronger contract (it holds *the* class, not something with a matching name). Measured: `test_g1_driver.py` + `test_reachy_driver.py` in collection order goes `1 failed, 160 passed, 20 skipped` on `main` → `162 passed, 20 skipped` with the root-cause fix alone.

One correction for the record, since it would have misdirected the next reader: the comment added here attributed the duplicate class to "the shipped-drivers registration loop running twice". That is not the mechanism — the registration loop runs once; the second class object comes from `importlib.reload` re-executing the module body into the same namespace, as #2793 records.

PR #2788, opened 2026-08-26
fix(protomotions): report a yaml sidecar the loader cannot use, by name

cagataycali:
## Problem

`load_config_from_yaml` is the fourth policy config file loader and carried none of the four guards the other three share, so a sidecar it could not use surfaced as a message naming a method of the parsed value rather than the file.

Derived over `strands_robots/policies/*/config.py`, taking every function that reads a caller-named `path` from disk:

| loader | expands `~` | reads a *file* | wraps the decode | refuses a non-mapping |
|---|---|---|---|---|
| `KimodoConfig.from_json` | yes | yes | yes | yes |
| `MotionBricksConfig.from_file` | yes | yes | yes | yes |
| `WBCConfig.from_file` | yes | yes | yes | yes |
| `load_config_from_yaml` | **no** | **no** (`exists()`) | **no** | **no** |

Measured against the shipped loader, four consequences:

| input | before |
|---|---|
| empty sidecar (`""`) | `AttributeError: 'NoneType' object has no attribute 'get'` |
| comments-only sidecar | `AttributeError: 'NoneType' object has no attribute 'get'` |
| `~/unified_pipeline.yaml`, file present | `FileNotFoundError: ProtoMotions yaml not found: ~/unified_pipeline.yaml` |
| a directory | `IsADirectoryError: [Errno 21] Is a directory: ...` |
| malformed yaml | `ParserError` (a bare `yaml.YAMLError`) |
| a yaml list / scalar | `AttributeError: 'list' object has no attribute 'get'` |

### The case that decides the disposition

`yaml.safe_load("")` is `None`. The loader documents fields absent from the yaml as falling back to the dataclass defaults, "so a missing block is not an error" - and an empty document is the limit of that, every field absent. Measured on the shipped loader:

```
empty mapping '{}'   OK -> config; equals all-defaults ProtoMotionsConfig(): True
empty file ''        AttributeError: 'NoneType' object has no attribute 'get'
only a comment       AttributeError: 'NoneType' object has no attribute 'get'
```

Two spellings of the same information, opposite outcomes, one of them a dead end. So an empty document loads the defaults rather than being refused; the alternative (refuse it) is measured as a break row below.

### Reachability

The public constructor forwards a caller's path straight in (`policy.py:140`):

```
ProtoMotionsPolicy(session=stub, yaml_path=<empty sidecar>)
  -> AttributeError: 'NoneType' object has no attribute 'get'
ProtoMotionsPolicy(session=stub, yaml_path="~/unified_pipeline.yaml")
  -> FileNotFoundError: ProtoMotions yaml not found: ~/unified_pipeline.yaml
```

## Change

Apply the reporting the three siblings already give - and no more: expand `~`, require a file, wrap the decode into the documented `ValueError` naming the path, treat a document with no fields as the all-defaults config, refuse a non-mapping payload by type name. `+33/-8` in the loader.

The **extension stays unchecked**, for the reason `KimodoConfig.from_json`'s docstring gives for its own: a yaml document stored under any name loads today, and refusing it would stop a payload that currently works. Pinned as a control.

`require_optional` ordering is left alone: the tree is split on whether the dependency probe or the path check comes first (6 functions vs 5), so that is a separate question, not a convention this loader is breaking.

## Why nothing caught it

`tests/policies/kimodo/test_config_file_loader_refusals.py` grades exactly this rule and its docstring promised "a fourth policy config's loader is graded the hour it lands". It keys discovery on the **`from_*` classmethod shape**, which is what its `from_dict` half needs and is also the limit of what it reaches - a module-level loader is outside it however many guards it drops. That file's docstring now states that limit and points at the new one; the new survey keys on *reading a path from disk*, so all four are in scope however they are spelled.

The two rows that show this is load-bearing rather than decorative, with `collected` recorded because a narrowed universe runs fewer tests while still reporting success:

| survey mutation | collected | fires |
|---|---|---|
| control | 26 | 0 |
| universe hardcoded to the three old loaders | **25** | 1 |
| discovery keyed on the `from_*` classmethod shape | **25** | 1 |

## Tests

New `tests/policies/protomotions/test_config_file_loader_reports_the_file.py`, 26 cases. Pre-fix split, source reverted and the tests unchanged: **11 failed / 15 passed**, with **0 of 6** non-vacuity cells and **0 of 4** over-reach controls firing. The `empty-mapping` case inside the regression class passes both ways by design - it is the spelling that already worked, and the pair is what makes the inconsistency the finding.

Break table, 11 mutations -> **8 distinct firing sets**, control clean, both files restored byte-identical. `sib` is failures in the pre-existing `tests/policies/{protomotions,kimodo,motionbricks,wbc}/` suites:

| mutation | mine | sib |
|---|---|---|
| control - nothing mutated | 0 | 0 |
| whole fix reverted | 11 | 0 |
| `expanduser()` dropped | 2 | 0 |
| `is_file()` weakened to `exists()` | 2 | 0 |
| decode no longer wrapped | 2 | 0 |
| empty document no longer defaults | 4 | 0 |
| non-mapping refusal dropped | 4 | 0 |
| empty document REFUSED instead of defaulted | 4 | 0 |
| mapping check placed before the default | 4 | 0 |
| refusal names the type, not the file | 3 | 0 |
| survey universe hardcoded | 1 | 0 |
| survey keyed on the classmethod shape | 1 | 0 |

Two collisions, both structural: the three empty-document mutations all mean "an empty document is not treated as every-field-absent" and fire the same 4 cells; the two survey mutations both make the fourth loader invisible and are caught by the non-vacuity list. `sib` being 0 on every row is the measurement that the existing suites cannot see any of this.

The shipped loaders are all compliant after the fix, so the survey cannot exercise its own failing branch - the guard predicate is therefore graded against constructed exemplars written to miss exactly one guard each, plus a both-outcomes check.

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1606 files).
- mypy **24 == 24** against a pristine worktree of the merge base, normalized message sets byte-identical in both directions, **0** attributable to the changed files.
- Paired run over an identical 617-file selection (all of `tests/policies/` plus every test walking the tree, parsing source, reading docstrings or reading `changelog.d`), the new file excluded from both: identical **26157 collected** and `20 failed, 26037 passed, 100 skipped` on both trees, **20 identical failing ids, `comm` empty in both directions**. All 20 also fail on pristine `main`: 17 need `awsiot` (`awsiotsdk`, the `[mesh-iot]` extra, absent here and installed in CI) and 3 are pre-existing lerobot-from-source drift in `tests/training/test_lerobot*`.
- Coverage for the module, scoped run verified against the full-suite run as a faithful oracle (identical `92 / 2 / 98%` and identical missing lines on the base tree): **98% -> 99%**, 2 -> 1 missing. Line 394 was the loader's own `FileNotFoundError` raise, which no test had executed.
- Changed files together: **280 passed**, re-run after the rebase onto `main`.

No visual artifact: this changes a file loader's refusals, not a policy, simulation, rendering, recording or asset.

### DIFF
--- a/changelog.d/2788-protomotions-yaml-loader-reports-the-file.md
+++ b/changelog.d/2788-protomotions-yaml-loader-reports-the-file.md
@@ -1,0 +1,31 @@
+### Fixed
+
+- **policies/protomotions**: `load_config_from_yaml` now reports a yaml sidecar
+  it cannot use the way its three sibling policy-config file loaders already do.
+  It is the fourth such loader and carried none of the four guards the other
+  three share, so the failure a caller saw named a method of the parsed value
+  rather than the file. The case that decides it is an **empty** sidecar - a
+  truncated download, a `touch`ed placeholder, a file holding only comments:
+  `yaml.safe_load("")` is `None`, so the first field lookup raised
+  `AttributeError: 'NoneType' object has no attribute 'get'`, while an empty
+  mapping `{}` - the same information, every field absent - already returned the
+  all-defaults config this function documents absent fields to mean ("a missing
+  block is not an error"). Two spellings of one input, and one of them
+  dead-ended; both now load the defaults. Alongside that, `~` is expanded (a
+  sidecar at `~/unified_pipeline.yaml` was reported *missing* while it existed,
+  quoting the literal `~`), the path must name a file rather than merely exist
+  (a directory reached the read and surfaced as `IsADirectoryError`), malformed
+  yaml is wrapped into the documented `ValueError` naming the path instead of
+  escaping as a bare `yaml.YAMLError`, and a document holding a list or a scalar
+  is refused by type (`ProtoMotions yaml /tmp/s.yaml must contain a mapping, got
+  list`) instead of reaching `data.get(...)`. `KimodoConfig.from_json`,
+  `MotionBricksConfig.from_file` and `WBCConfig.from_file` shipped all four
+  already, so this is three of four loaders' behaviour applied to the fourth.
+  Reachable through the public `ProtoMotionsPolicy(yaml_path=...)` constructor,
+  which forwards a caller's path straight in. The extension stays unchecked, for
+  the reason `from_json` documents for its own: a yaml document stored under any
+  name loads today, and refusing it would stop a payload that currently works.
+  The new tests derive the rule over *reading a path from disk* rather than over
+  the `from_*` classmethod shape the existing guard keys on - which is why a
+  module-level loader could land unheld - so all four are graded however they
+  are spelled.
--- a/strands_robots/policies/protomotions/config.py
+++ b/strands_robots/policies/protomotions/config.py
@@ -366,16 +366,31 @@ def load_config_from_yaml(path: str | Path) -> ProtoMotionsConfig:
     fall back to the dataclass defaults (which are themselves pinned to the
     shipped weights, so a missing block is not an error).
 
+    An empty sidecar is the limit of "fields absent from the yaml", so it
+    yields the all-defaults config exactly as ``{}`` does - the two spell the
+    same information. ``~`` in ``path`` is expanded, the file is read as a
+    file, and a payload that is not a mapping is reported by name rather than
+    reaching the field lookups below: the reporting the sibling policy-config
+    file loaders in :mod:`strands_robots.policies.kimodo.config`,
+    :mod:`strands_robots.policies.motionbricks.config` and
+    :mod:`strands_robots.policies.wbc.config` already give.
+
+    The extension is deliberately not checked, unlike the two loaders that do:
+    a yaml document stored under any name loads here today, and refusing one
+    would stop a payload that currently works.
+
     Args:
-        path: Path to the yaml file.
+        path: Path to the yaml file. ``~`` is expanded.
 
     Returns:
         A :class:`ProtoMotionsConfig` - validated for consistent dimensions.
 
     Raises:
-        FileNotFoundError: If ``path`` does not exist.
+        FileNotFoundError: If ``path`` does not name a file (a directory does
+            not, so it is refused here rather than at the read).
         ImportError: If ``pyyaml`` is not installed.
-        ValueError: If the yaml contains an inconsistent dimension: a
+        ValueError: If the file is not valid YAML, or holds a YAML value that
+            is not a mapping, or contains an inconsistent dimension: a
             ``stiffness`` or ``damping`` length that is not the joint count, or
             an ``anchor_body_index`` / ``root_body_index`` that is not a row of
             ``body_names`` (refused by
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Brings `load_config_from_yaml` (the fourth policy config file loader) up to the guard set its three siblings (`KimodoConfig.from_json`, `MotionBricksConfig.from_file`, `WBCConfig.from_file`) already share: `~` expansion, `is_file()` instead of `exists()`, the decode wrapped into the documented `ValueError` naming the path, an empty/comments-only document treated as `{}` (all-defaults, matching the loader's own documented "absent fields fall back" contract), and a non-mapping payload refused by type name. The `+33/-8` loader diff is tightly scoped, and every refusal it adds was already in the function's documented `Raises:` vocabulary (`FileNotFoundError` / `ValueError`), so no caller contract changes shape. The new AST-derived survey in `tests/policies/protomotions/test_config_file_loader_reports_the_file.py` keys discovery on reading-a-path-from-disk rather than the `from_*` classmethod shape that let this loader land unheld, and the kimodo guard's docstring is updated to state that limit and point at the new survey.

## What's good

- The empty-sidecar disposition (default, not refuse) is argued from the loader's own documented contract and pinned in both directions (empty file == `{}`), with the refuse-instead mutation measured as a break row.
- Over-reach controls pin the deliberate divergence (extension unchecked) so the narrowing stays intentional, per the same reasoning `from_json` documents.
- `Raises:` block updated in the same change (AGENTS.md > "A `Raises:` block names every refusal the function itself makes"), changelog fragment present, ASCII-clean, no host paths, tests named for behaviour.
- Non-vacuity exemplars grade the guard predicate against constructed sources missing exactly one guard each, since the shipped loaders are all compliant post-fix.

Note for the human reviewer: the `call-test-lint / Test and Lint` failure on this head is `tests/drivers/test_reachy_driver.py::TestTheLerobotPathIsUnaffected::test_the_g1_registration_survives_a_second_shipped_driver`, which fails identically on `main`'s own push runs (e.g. run for `828f80eb`) - it is pre-existing and unrelated to the four changed files here.

yinsong1986 (approved):
## Summary

Brings `load_config_from_yaml` up to the guard set its three sibling policy-config file loaders (`KimodoConfig.from_json`, `MotionBricksConfig.from_file`, `WBCConfig.from_file`) already share: `~` expansion, `is_file()` instead of `exists()`, the decode wrapped into the documented `ValueError` naming the path, an empty/comments-only document treated as `{}` (all-defaults, matching what the docstring already promises for absent fields), and a non-mapping payload refused by type name. Every error-type change moves an undocumented crash class (`AttributeError: 'NoneType' ... 'get'`, `IsADirectoryError`, bare `yaml.YAMLError`) into the contract the docstring already declared, and the `Raises:` block is updated in the same diff, so nothing here is a silent behaviour break. The new test file derives its loader survey over *reading a path from disk* rather than the `from_*` classmethod shape the existing kimodo guard keys on - closing the discovery gap that let this fourth loader land unheld - and pins its own predicate against constructed exemplars plus over-reach controls (extension deliberately unchecked, matching `from_json`'s documented divergence).

## What's good

- Changelog fragment under `changelog.d/` rather than a direct `CHANGELOG.md` edit (AGENTS.md step 3).
- Docstring `Raises:` updated in the same change as the new refusals (AGENTS.md > 'A `Raises:` block names every refusal the function itself makes').
- Refusal messages are plain ASCII and name the file, not a method of the parsed value.
- Tests use `monkeypatch.setenv` + `tmp_path`, no host paths, behaviour-named test classes, and the survey's universe is pinned so a narrowed discovery cannot pass silently.

PR #2789, opened 2026-08-26
fix(device_connect): report getDaemonStatus under this driver's own verdict

cagataycali:
## `getDaemonStatus` let the daemon decide the RPC's verdict, and called an unreachable daemon a success

`ReachyMiniDriver` answers every RPC with the Device Connect envelope -
`{"status": "success", ...}` or `{"status": "error", "reason": ...}`. Five of its
REST RPCs carry the daemon's reply back to the caller. Four nest it under a key
of their own (`{"status": "success", "result": result}`); `getDaemonStatus`
merged it into the envelope:

```python
return {"status": "success", **result}
```

Two properties follow from that one difference, and a third from the same
statement.

### 1. A reply carrying its own `status` replaced the driver's verdict

Spread last, the daemon's mapping wins a name collision. `status` is the
ordinary name for the field a `/status` endpoint answers with:

| `api()` answered | envelope `status` |
|---|---|
| `{"status": "idle", "freq": 100}` | `idle` |
| `{"status": "error", "detail": "motor fault"}` | `error` |
| `{"status": "running", "motors_on": True}` | `running` |

`idle` is not in the envelope's vocabulary, so a caller branching on
`status == "success"` reads a call that worked as a call that did not; and a
daemon reporting a fault of its own made a successful RPC read as a failed one.

The presence path resolves this exact collision the other way, and
`SensorLoopsMixin._stamp_local_keys` states the rule: a record must not name
something other than what the surface that built it decided. `PeerInfo.to_dict`
spreads the peer's payload *first* for the same reason.

### 2. A daemon that was never reached answered `success`

`reachy_transport.api` reports every HTTP and connection failure as
`{"error": ...}` rather than raising. This driver already says so twice - in
`__init__`'s port comment and in `_stop_motion_impl`'s docstring, which spells
out the consequence: "a stop issued against a down daemon would otherwise
report `status="success"`". Merged into the envelope, the reason travelled
*beside* a success verdict, so the one RPC whose subject is whether the daemon
is up was the one that could not say it was down:

```
api()   -> {"error": "<urlopen error [Errno 111] Connection refused>"}
before  -> {"status": "success", "error": "<urlopen error [Errno 111] Connection refused>"}
```

`device_connect_edge`'s own `@rpc` wrapper classifies a result carrying an
`error` key as `PARTIAL` and logs it - then returns it to the caller unchanged.
The contradiction was observable and shipped anyway.

The native driver reads this same endpoint and refuses that shape:
`ReachyDriver.connect_eagerly` answers
`daemon unreachable ({host}:{port}): {error}`. One endpoint, two drivers, two
conventions.

### 3. A body that is not a JSON object raised out of the envelope

`api()` returns `json.loads(...)`, which can be an array or a scalar. Spreading
a non-mapping is a `TypeError` out of a method whose whole contract is the
envelope - logged by the wrapper as `FAILED: 'list' object is not a mapping`.

### Measured, all eight inputs

| `api()` answered | before | after |
|---|---|---|
| `{"motors_on": True, "freq": 100}` | `success` | `success` |
| `{"state": "ready", "version": "1.0"}` | `success` | `success` |
| `{"status": "idle", "freq": 100}` | `idle` | `success` |
| `{"status": "error", "detail": "motor fault"}` | `error` | `success` |
| `{"error": "[Errno 111] Connection refused"}` | `success` | `error` |
| `{"error": "service unavailable", "code": 503}` | `success` | `error` |
| `[1, 2, 3]` | `TypeError` | `error` |
| `"ok"` | `TypeError` | `error` |

The premise is measured against the real transport, not asserted: `api()` on a
port nothing is listening on returns
`{'error': '<urlopen error [Errno 111] Connection refused>'}` and does not
raise.

## The change

`status` is re-asserted after the merge, so every payload key still reaches the
caller at the top level and the healthy reading is unchanged - only the one key
the envelope owns changes hands. A failure is reported in the error envelope the
class already uses for its read-only RPCs (`getJoints`, `getImu`) rather than
raised: `_stop_motion_impl` raises because it guards a *stop*, where a caller
acting on a false success stops nothing, while this RPC answers a question and
its callers already branch on `status`.

### Why the naive fix is not enough

Fixing only the precedence makes the transport case *worse*: with
`{**result, "status": "success"}` and no error check, `{"error": ...}` is forced
to `success` instead of accidentally surfacing as the daemon's own value. The
mutation table below measures that as row `b4` - 20 of the 46 new cells fire.

## Scope

The four REST siblings nest the reply, so nothing it carries can land on a key
of their envelope, and the reply stays inspectable at `result["error"]`.
Changing their verdict is a separate decision about four write RPCs; this PR
changes the one method a reply can reach into, and the one whose subject is
daemon reachability. `TestWhyTheEnvelopeOwnsTheVerdict::test_this_is_the_only_rpc_that_merges_the_reply_into_its_envelope`
pins that boundary from the class's own source.

`connect()` also reads this endpoint and treats a payload without
`wireless_version` as a Wireless. That is deliberate and matched by the native
driver's comment on the same field, so it is untouched.

## Verification

New file `tests/test_reachy_daemon_status_verdict.py` - **46 cells across 7
classes**, no daemon, no network, no hardware.

Pre-fix split (source reverted, tests kept): **25 failed / 21 passed**

| class | role | failing |
|---|---|---|
| `TestADaemonReplyCannotOverwriteTheVerdict` | regression | 4 of 4 |
| `TestADaemonThatWasNotReachedIsNotASuccess` | regression | 12 of 12 |
| `TestABodyThatIsNotAnObjectIsReportedNotRaised` | regression | 8 of 8 |
| `TestNoEnvelopeInThisDriverDeclaresStatusAheadOfASpread` | drift guard | 1 of 8 |
| `TestReAssertingTheVerdictCostsNothingElse` | over-reach control | **0 of 6** |
| `TestTheHealthyPayloadIsUnchanged` | control | **0 of 5** |
| `TestWhyTheEnvelopeOwnsTheVerdict` | premise | **0 of 3** |

### Mutations

`sib` = failures in `tests/test_reachy_mini_driver.py` +
`tests/test_device_connect_drivers.py` (117 pre-existing cells), which is why
none of this was caught before.

| row | mutation | collected | fires | sib |
|---|---|---|---|---|
| b0 | control - no mutation | 46 | 0 | 0 |
| b1 | precedence reverted, both guards kept | 46 | 5 | 0 |
| b2 | transport-error guard dropped | 46 | 12 | 0 |
| b3 | non-object guard dropped | 46 | 8 | 0 |
| b4 | both guards dropped, precedence kept (the naive fix) | 46 | 20 | 0 |
| b5 | behavioural full revert | 46 | 25 | 0 |
| b6 | error checked for truthiness, not presence | 46 | 4 | 0 |
| b7 | reason drops the host:port | 46 | 5 | 0 |
| b8 | reason drops the transport cause | 46 | 2 | 0 |
| b9 | non-object body reported as a success | 46 | 8 | 0 |
| b10 | transport failure raises instead of the envelope | 46 | 12 | 0 |
| b11 | the reason is filed under another key | 46 | 6 | 0 |

11 mutations, none undetected, 9 distinct firing sets, `collected` stable at 46,
source restored byte-identically (md5 asserted per row).

`b1 | b2 | b3 == b5` is **True** and `b1 & b2` is **empty**: each half is pinned
independently and together they account for the whole pre-fix split. `b6` is the
row that justifies presence over truthiness - an HTTP error response with an
empty body puts `""` there, so a truthiness check merges it. `b2`/`b10` and
`b3`/`b9` share a firing set because each pair makes the same case answer with
something other than the error envelope.

### Gate

Paired run over an identical 430-file selection (every test naming
`device_connect` / `reachy` / `_stamp_local_keys`, plus every guard that walks
the tree, parses source, reads docstrings or reads `changelog.d`), the new file
excluded from both trees:

| tree | result |
|---|---|
| `828f80eb` (base) | 22010 collected, `26 failed, 21883 passed, 101 skipped` (761.93s) |
| this branch | 22010 collected, `26 failed, 21883 passed, 101 skipped` (727.07s) |

26 identical failing ids, `comm` empty in both directions. All 26 fail on the
base: 17 need `awsiot` (declared in `[mesh-iot]`, absent here), 6 in
`tests/drivers/test_reachy_transport_guards_are_reachable.py`, 2 need
`device_connect_agent_tools`, 1 is lerobot-from-source `encode()` drift.

`ruff check` and `ruff format --check` clean. mypy **4 == 4** against a pristine
worktree of the same base, normalized message sets byte-identical in both
directions, 0 attributable to either changed file.

Sibling suites on this branch: `tests/test_reachy_mini_driver.py`,
`tests/test_device_connect_drivers.py`, `tests/test_reachy_key_prefix_domain.py`,
`tests/test_reachy_motion_domain.py`, `tests/test_reachy_api_port_domain.py`,
`tests/test_reachy_transport_links.py`, `tests/device_connect/` - **380 passed**.

No visual artifact: this changes an RPC envelope's verdict, not a policy,
simulation, rendering, recording or asset. The eight-input table is the
measurement.

## Docs

`getDaemonStatus` appears in no docs page - the RPC docstrings are its
documentation surface, and this method's docstring now carries the contract, the reason the
error envelope is used rather than the raise its sibling chose, and the
`Returns:` block naming both shapes.

### DIFF
--- a/changelog.d/2789-reachy-daemon-status-verdict.md
+++ b/changelog.d/2789-reachy-daemon-status-verdict.md
@@ -1,0 +1,47 @@
+### Fixed: getDaemonStatus reports the Reachy Mini driver's own verdict, not the daemon's
+
+`ReachyMiniDriver` answers every RPC with the Device Connect envelope -
+`{"status": "success", ...}` or `{"status": "error", "reason": ...}`. Five of its
+REST RPCs carry the daemon's reply back to the caller and four of them nest it
+under a key of their own (`{"status": "success", "result": result}`).
+`getDaemonStatus` merged it into the envelope instead, and that one difference
+cost the envelope two properties.
+
+The daemon decided the verdict. Spread last, a reply carrying a `status` field
+of its own replaced the driver's, so a daemon reporting `status="idle"` made a
+healthy RPC answer `"idle"` - a value outside the envelope's two-value
+vocabulary, which a caller branching on `status == "success"` reads as a call
+that did not work. The presence path resolves the same collision the other way
+and `SensorLoopsMixin._stamp_local_keys` gives the reason: a record must not
+name something other than what the surface that built it decided.
+
+A daemon that was never reached answered `success`. `reachy_transport.api`
+reports every HTTP and connection failure as `{"error": ...}` rather than
+raising - the driver already states this twice, in `__init__`'s port comment and
+in `_stop_motion_impl`'s docstring, and the native `ReachyDriver` refuses that
+shape when reading this same endpoint. Merged into the envelope, the reason
+travelled beside a success verdict: the one RPC whose subject is whether the
+daemon is up was the one that could not say it was down.
+
+```
+api() answered                                   before          after
+{"motors_on": True, "freq": 100}                 success         success
+{"state": "ready", "version": "1.0"}             success         success
+{"status": "idle", "freq": 100}                  idle            success
+{"status": "error", "detail": "motor fault"}     error           success
+{"error": "[Errno 111] Connection refused"}      success         error
+{"error": "service unavailable", "code": 503}    success         error
+[1, 2, 3]                                        TypeError       error
+"ok"                                              TypeError       error
+```
+
+The failure is reported in the error envelope the class already uses for a
+read-only RPC rather than raised, because the `RuntimeError`
+`_stop_motion_impl` raises guards a stop, where a caller acting on a false
+success stops nothing; this one answers a question and its callers already
+branch on `status`. A body that decodes to a JSON array or scalar is reported
+for the same reason: spreading a non-mapping raised `TypeError` out of a method
+whose whole contract is the envelope.
+
+`status` is re-asserted after the merge, so every payload key still reaches the
+caller at the top level and the healthy reading is unchanged.
--- a/strands_robots/device_connect/reachy_mini_driver.py
+++ b/strands_robots/device_connect/reachy_mini_driver.py
@@ -548,14 +548,57 @@ class ReachyMiniDriver(DeviceDriver):
 
     @rpc()
     async def getDaemonStatus(self) -> dict[str, Any]:
-        """Get daemon status, motor state, and control frequency."""
+        """Report the daemon's status payload under this driver's own verdict.
+
+        The payload's keys are merged into the envelope so a caller reads
+        ``motors_on`` / ``freq`` at the top level, and ``status`` is re-asserted
+        afterwards. Merged last, a daemon reply carrying a ``status`` field of
+        its own replaced this envelope's verdict: a healthy call answered
+        ``status="idle"``, and a daemon reporting its own fault answered
+        ``status="error"`` for an RPC that reached it and succeeded. The
+        presence path resolves the same collision the same way, by spreading
+        the foreign mapping first so the locally decided keys win - see
+        :meth:`strands_robots.mesh.sensors.SensorLoopsMixin._stamp_local_keys`.
+
+        A daemon that was not reached is reported rather than merged.
+        :func:`~strands_robots.device_connect.reachy_transport.api` answers
+        every HTTP and connection failure with ``{"error": ...}`` instead of
+        raising, so the unreachable daemon this RPC exists to detect came back
+        as ``status="success"`` with the reason merged in beside it. The native
+        driver reads this same endpoint and refuses that shape - see
+        :meth:`strands_robots.drivers.reachy.ReachyDriver.connect_eagerly`.
+        The error envelope is used rather than the ``RuntimeError``
+        :meth:`_stop_motion_impl` raises, because that one guards a stop, where
+        a caller acting on a false success stops nothing; this one answers a
+        question, and its callers already branch on ``status``.
+
+        A body that decodes to something other than a JSON object is reported
+        for the same reason: spreading it raises ``TypeError`` out of a method
+        whose whole contract is the envelope.
+
+        Returns:
+            ``{**payload, "status": "success"}`` when the daemon answered with
+            a JSON object, otherwise ``{"status": "error", "reason": ...}``
+            naming the daemon that was not reached or the body that could not
+            be merged.
+        """
         result = await asyncio.to_thread(
             api,
             self._host,
             self._api_port,
             "/api/daemon/status",
         )
-        return {"status": "success", **result}
+        if not isinstance(result, dict):
+            return {
+                "status": "error",
+                "reason": f"daemon status is not a JSON object: {type(result).__name__}",
+            }
+        if (error := result.get("error")) is not None:
+            return {
+                "status": "error",
+                "reason": f"daemon unreachable ({self._host}:{self._api_port}): {error}",
+            }
+        return {**result, "status": "success"}
 
     # ── Events ────────────────────────────────────────────────
 
--- a/tests/test_reachy_daemon_status_verdict.py
+++ b/tests/test_reachy_daemon_status_verdict.py
@@ -1,0 +1,383 @@
+"""``getDaemonStatus``'s envelope reports this driver's verdict, not the daemon's.
+
+``ReachyMiniDriver`` answers every RPC with the Device Connect envelope --
+``{"status": "success", ...}`` or ``{"status": "error", "reason": ...}``. Five
+of its REST RPCs carry the daemon's reply back to the caller and every one of
+them nests it under a key of its own (``{"status": "success", "result":
+result}``). ``getDaemonStatus`` merged it into the envelope instead, and that
+one difference cost the envelope two properties:
+
+* **The daemon decided the verdict.** Spread last, a reply carrying a
+  ``status`` field of its own replaced the driver's. A daemon reporting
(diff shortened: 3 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Fixes `getDaemonStatus` so the Device Connect envelope's `status` key is this driver's own verdict rather than whatever the daemon's reply happened to carry. Previously the payload was spread *after* the envelope's `status`, so a daemon reporting `status="idle"` overwrote the verdict with a value outside the envelope's two-value vocabulary, and — because `reachy_transport.api` reports every HTTP/connection failure as `{"error": ...}` instead of raising — an unreachable daemon came back as `status="success"` with the failure reason merged in beside it. The fix guards the two degenerate shapes (non-dict body, transport `error` key) with proper `{"status": "error", "reason": ...}` envelopes naming the host:port, and spreads the payload first so the locally-decided `status` wins, matching how `SensorLoopsMixin._stamp_local_keys` resolves the same collision.

## What's good

- The error-shape discrimination (`result.get("error") is not None`, presence not truthiness) matches the transport's actual failure shapes, including the empty-body HTTP-error case, and is consistent with the sibling `_stop_motion_impl` guard.
- 383-line regression suite pins both directions: the fix itself, the healthy-payload reading that must not change, the premises that pick the verdict's owner (including a real-socket probe of `api`'s no-raise contract), and an AST drift guard so no future RPC in this driver declares `status` ahead of a spread — with the guard's own rule graded against constructed exemplars.
- Docstring cross-references (`_stamp_local_keys`, `ReachyDriver.connect_eagerly`) resolve against the real API; the `sys.modules` fixture follows the established `_force_real_device_connect_edge` precedent from the sibling reachy suites.
- Verified locally at head: the new suite (46 tests), `tests/test_reachy_mini_driver.py`, the sys-modules-orphan and test-name graders, and the docstring-xref graders all pass; ruff check/format clean on both files.

yinsong1986 (approved):
## Summary

Fixes `getDaemonStatus` so the Device Connect envelope's `status` key carries this driver's own verdict instead of whatever the daemon's reply happened to spell there. Before the fix the payload was spread after the envelope's `status`, so a daemon reporting `status="idle"` overwrote the verdict with a value outside the envelope's two-value vocabulary — and because `reachy_transport.api` reports every HTTP and connection failure as `{"error": ...}` rather than raising, an unreachable daemon answered `status="success"` with the failure reason merged in beside it. The fix guards the non-object body and the transport-error shape with proper `{"status": "error", "reason": ...}` envelopes naming the host:port and the cause, and spreads the payload first so the locally decided verdict wins. Both failure shapes the guards discriminate match the transport source exactly (`{"error": str(e)}` and `{"error": <body>, "code": <status>}`), the guards return error dicts rather than raise per the tool error-handling contract, and the healthy-path reading is byte-identical to what the two pre-existing sibling tests already pin.

## What's good

- The daemon's own `status` value being shadowed by the re-asserted verdict is an acknowledged scope decision, pinned by `test_exactly_one_key_differs_from_the_daemon_payload`, and the alternative (nesting, like the four REST siblings) is correctly deferred as a separate decision about four write RPCs.
- The new suite covers both directions: regressions that fail pre-fix, controls that hold on both trees, a real-socket probe of the transport's no-raise premise, and an AST drift guard (itself graded against constructed exemplars) so the next RPC that merges a reply is held to the same precedence rule.
- Changelog fragment present under `changelog.d/` per AGENTS.md step 3; all changed hunks are ASCII-clean; the `is not None` presence check correctly keeps the empty-body HTTP-error case an error, matching the sibling `_stop_motion_impl` guard's convention.

yinsong1986 (approved):
## Summary

Fixes the one RPC in `ReachyMiniDriver` that merged the daemon's reply into the Device Connect envelope with the spread last, so the daemon's own `status` field silently replaced the driver's verdict (a healthy daemon reporting `idle` read as a failed call) and — because `reachy_transport.api` reports every HTTP/connection failure as `{"error": ...}` without raising — an unreachable daemon answered `status="success"` with the reason travelling beside it. The fix re-asserts `status` after the merge, reports a transport failure in the class's existing error envelope naming host:port and cause, and reports a non-object JSON body instead of letting the spread raise `TypeError`. I verified the claims independently: the transport premise holds against the shipped `api()` (returns `{"error": str(e)}` / `{"error": ..., "code": ...}`, never raises); reverting only the source change reproduces the stated 25-failed/21-passed pre-fix split and all 46 new cells pass at head; `getDaemonStatus` has no in-tree caller reading the daemon's own `status` value, and the two pre-existing tests driving it assert exactly the healthy shapes the change preserves. The choice to keep the merged (flat) payload shape rather than nesting like the four REST siblings preserves the existing wire shape for healthy replies, and the AST drift guard pins the status-before-spread rule for any future RPC that merges a reply.

## What's good

- Regression tests genuinely fail pre-fix, with an over-reach control class and premise class that hold on both trees.
- Scope discipline: the four nesting REST siblings and `connect()`'s deliberate `wireless_version` fallback are left untouched, with the boundary pinned from the class's own source.
- Changelog fragment (not a `CHANGELOG.md` edit), ruff check/format clean, all added lines pure ASCII, docstring cross-reference targets (`_stamp_local_keys`, `connect_eagerly`) resolve, and the `_force_real_device_connect_edge` fixture follows the pattern already established in four sibling test files.

PR #2790, opened 2026-08-26
test(drivers): combined fix for both Reachy test flakes (supersedes #2785, #2787)

cagataycali:
Combined fix for the two interlocking Reachy driver test flakes that block #2784, #2786, #2787 and #2785 in a cycle. Each fix on its own hits the *other* flake in CI, so neither can go green independently.

### 1. From #2787 (`test_reachy_driver.py`)
`test_the_g1_registration_survives_a_second_shipped_driver` graded G1 registration by `get_native_driver_class("unitree_g1") is G1Driver`. Under test orderings where `strands_robots.drivers.g1` is re-imported (via `importlib.reload` in `tests/drivers/test_g1_driver.py:1388`), a second class object with the same `__qualname__` is created — the registry is untouched but `is` returns False. Fix: compare `__module__` + `__qualname__`, with `assert registered is not None` ahead of it so a genuinely-lost registration still fails loudly.

### 2. From #2785 (`test_reachy_transport_guards_are_reachable.py`)
The staging file made `device_connect_edge` unimportable and evicted the `strands_robots.device_connect` subtree, expecting the next import to re-execute `__init__` and refuse. #2774 removed that eager import, so the simulation stopped simulating anything and the driver reached the daemon (real `urlopen [Errno 111] Connection refused`). Fix: stage the absence at the one module `_resolve_transport` imports via `monkeypatch.setitem(sys.modules, _TRANSPORT_MODULE, None)`, per AGENTS.md > "Restore a sys.modules entry you remove". The AST premise test now reads executed module-scope imports (skips TYPE_CHECKING blocks correctly).

## Why combined

CI on #2787's own branch fails on the guards flake #2785 fixes; CI on #2785's own branch fails on the identity flake #2787 fixes. Landing them together breaks the cycle and unblocks #2784 (spatial-tendon, APPROVED, red on identity flake) and #2786 (Feetech codec PR-A, red on identity flake).

## Verification

Fixes cherry-picked cleanly onto each other. No conflicts. Both are test-only changes (test files + one changelog fragment each), no production surface touched.

Once this lands: rebase #2784 and #2786 onto main → both go green. #2785 and #2787 close as superseded.

### DIFF
--- a/changelog.d/2785-reachy-guards-stage-the-absence-at-the-resolver.md
+++ b/changelog.d/2785-reachy-guards-stage-the-absence-at-the-resolver.md
@@ -1,0 +1,29 @@
+### Fixed: the Reachy transport guards stage the absence at the module the resolver imports
+
+`tests/drivers/test_reachy_transport_guards_are_reachable.py` grades the two
+`_resolve_transport` guards that `ReachyDriver.connect_eagerly`'s pre-check
+hides, so a stock install is refused by name rather than crashing. It staged
+that absence by making `device_connect_edge` unimportable and evicting the
+`strands_robots.device_connect` subtree from `sys.modules`, relying on the next
+import to re-execute a package `__init__` that fails.
+
+That `__init__` no longer imports the extra - deferring it is the whole point of
+the lazy export table it now carries - so the eviction re-executes an `__init__`
+that succeeds, the stdlib-only transport leaf imports, and `_resolve_transport`
+hands back the real module. The driver then reaches the daemon: six of the
+file's nine cells failed with `daemon unreachable (localhost:8000): <urlopen
+error [Errno 111] Connection refused>`, a real network call from a unit test.
+
+The eviction had a second cost. Deleting the subtree let the next import build a
+fresh module object for `strands_robots.device_connect.reachy_transport`, so a
+double installed on one object was not read through the other. Run before its
+sibling, this file took `tests/drivers/test_reachy_driver.py` from 109 passed to
+46 passed / 72 failed, and the pair from 0.30s to 21.49s of real connection
+timeouts.
+
+The absence is now staged at the one module `_resolve_transport` imports, and
+nothing else in `sys.modules` is touched. The premise test that should have
+caught the drift asserted `from device_connect_edge import` appeared in
+`__init__.py`, which the `TYPE_CHECKING` block satisfies - a block the source
+itself annotates `never executed`. It now reads the executed module-scope
+imports, so a typing-only import no longer reads as an eager one.
--- a/changelog.d/2787-g1-driver-class-identity-tolerant.md
+++ b/changelog.d/2787-g1-driver-class-identity-tolerant.md
@@ -1,0 +1,35 @@
+### Test: the G1-registration-survives assertion tolerates a second class object under a re-import
+
+`test_the_g1_registration_survives_a_second_shipped_driver` in
+`tests/drivers/test_reachy_driver.py` graded the registration by
+`get_native_driver_class("unitree_g1") is G1Driver`. Under some test orderings
+the `_register_shipped_drivers` loop reaches `strands_robots.drivers.g1` a
+second time - either via a test that reloads the module or via a stale
+`sys.modules` entry from an earlier suite phase - and the second import
+creates a fresh `G1Driver` class *object* with the same fully-qualified name.
+Both are the same registration for the registry's purposes (the qualname the
+seam looks up hasn't moved); the `is` check reports otherwise on the class
+object only, and the resulting failure is a false negative: the registration
+did survive, `is` just cannot tell.
+
+Fix: compare `__module__` and `__qualname__` instead of the class object. That
+is what a caller-facing consumer of the registry actually reads (drivers are
+looked up by module path, not by class object identity - `register_native_driver`
+stores by canonical robot name, not by class), and the pattern lerobot's own
+plugin registration uses.
+
+Observed on #2784's `call-test-lint` run on head `99efd10a` at
+`2026-08-26T17:36:18Z`:
+
+```
+tests/drivers/test_reachy_driver.py:722: AssertionError
+assert <class 'strands_robots.drivers.g1.G1Driver'> is <class 'strands_robots.drivers.g1.G1Driver'>
+```
+
+The two classes have the same repr because they *are* the same registration
+under a different identity. #2762 (which added
+`tests/drivers/test_reachy_driver.py`) shipped fine when it was the only
+driver in the suite; the flake surfaces when it sits alongside another
+driver-registering test that ordering-permutes ahead of it. This entry keeps
+the test's *intent* (registration must survive) without demanding a
+brittle-under-reimport identity check.
--- a/tests/drivers/test_reachy_driver.py
+++ b/tests/drivers/test_reachy_driver.py
@@ -717,9 +717,20 @@ class TestTheLerobotPathIsUnaffected:
 
     def test_the_g1_registration_survives_a_second_shipped_driver(self) -> None:
         # The registration loop must not let one driver's arrival cost another's.
+        # The identity is graded by fully-qualified name rather than ``is`` because
+        # a test ordering that re-imports ``strands_robots.drivers.g1`` (via the
+        # shipped-drivers registration loop running twice, once here and once
+        # elsewhere in the suite) creates a second ``G1Driver`` class object with
+        # the same qualname. Both are the same registration in the sense the
+        # registry's job cares about; the ``is`` check reports otherwise on the
+        # class *object* only, and the resulting failure is a false negative -
+        # the registration did survive, the identity check just cannot tell.
         from strands_robots.drivers.g1 import G1Driver
 
-        assert get_native_driver_class("unitree_g1") is G1Driver
+        registered = get_native_driver_class("unitree_g1")
+        assert registered is not None, "G1 was unregistered by adding another driver"
+        assert registered.__module__ == G1Driver.__module__
+        assert registered.__qualname__ == G1Driver.__qualname__
 
     def test_every_shipped_driver_satisfies_the_seam(self) -> None:
         # Derived from the shipped table, so a driver added later is held to the
--- a/tests/drivers/test_reachy_transport_guards_are_reachable.py
+++ b/tests/drivers/test_reachy_transport_guards_are_reachable.py
@@ -23,14 +23,21 @@ packaging accident, that the reason reaches a mesh peer, and that a working inst
 is unaffected.
 
 The absence is simulated rather than installed, since the suite runs where the
-extra is present: :func:`_hide_extra` makes ``device_connect_edge`` unimportable and
-evicts the already-imported ``strands_robots.device_connect`` package so the next
-import re-executes the ``__init__`` that fails. That is the same failure a stock
-install produces, at the same place, and ``monkeypatch`` restores both halves.
+transport is importable. :func:`_block_transport` blocks the one module
+:func:`~strands_robots.drivers.reachy._resolve_transport` imports, and ``monkeypatch``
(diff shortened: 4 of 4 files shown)


cagataycali:
`tests/drivers/` is red on `main` itself, so every open pull request's required check is red for a reason none of them introduced -- and of the three branches fixing it, this is the only one that can ever go green.

## The base is red

Measured on a pristine worktree of `upstream/main` @ `828f80eb`, no edits:

```
pytest tests/drivers/ -q --no-cov -p no:randomly
7 failed, 364 passed
```

| failing id | cause |
|---|---|
| `test_reachy_driver.py::TestTheLerobotPathIsUnaffected::test_the_g1_registration_survives_a_second_shipped_driver` | class identity across an `importlib.reload` |
| `test_reachy_transport_guards_are_reachable.py` (6 ids) | the staged absence misses, so the driver reaches a real daemon: `urlopen error [Errno 111] Connection refused` |

## Only the combined branch clears all seven

Each of the three branches run against the same selection, each at its own head:

| tree | `tests/drivers/` |
|---|---|
| `main` @ `828f80eb` | **7 failed**, 364 passed |
| #2787 @ `158c5d66` | **6 failed**, 365 passed |
| #2785 @ `aff60de3` | **1 failed**, 370 passed |
| this branch @ `1e4ae799` | **371 passed, 0 failed** |

`6 + 1 = 7` exactly, and the two failing sets are disjoint: #2787 fixes the identity id, #2785 fixes the six transport ids. So the two halves compose, and neither is a subset of the other.

The consequence is structural rather than a matter of waiting. `call-test-lint / Test and Lint` is a required status check, so a branch cannot merge while it is red -- and **#2785 and #2787 each leave the other's half of the base failure in place, so neither can be green alone.** #2787 is already `APPROVED`; that approval is currently spent on a branch that cannot reach a mergeable state by itself. Merging either one first is not available, so ordering the two does not help either.

## Every other open branch is red on these seven and nothing of its own

Same selection, each at its own head:

| branch | `tests/drivers/` | its own new tests |
|---|---|---|
| #2786 (Feetech codec) | 7 failed, **412 passed** | all 48 pass; the 7 are the base's |
| #2788 (protomotions yaml) | 7 failed, 364 passed | the 7 are the base's |
| #2789 (daemon status verdict) | 7 failed, 364 passed | the 7 are the base's |

The failing ids are byte-identical to the base's seven in all three cases -- none of these branches touches `tests/drivers/test_reachy*`. So the red on those three is a report about `main`, not about them, and they need no change to clear it.

## What this leaves

This branch is the one merge that unblocks the queue: it takes the base from 7 failed to 0, and the six other open branches' required checks then have nothing to report. Worth noting for whoever reads the queue next: closing #2785 and #2787 in favour of this one is also what clears the duplicate-addition report the open-set sweep currently raises, since all three create `changelog.d/*-reachy-guards-stage-the-absence-at-the-resolver.md` / `changelog.d/*-g1-driver-class-identity-tolerant.md` between them:

```
scripts/check_duplicate_claim.py --all-open
Outcome: duplicate-addition
#2785 and #2790 both create changelog.d/*-reachy-guards-stage-the-absence-at-the-resolver.md;
#2787 and #2790 both create changelog.d/*-g1-driver-class-identity-tolerant.md
```

One point on the fix itself, in its favour: blocking the single module `_resolve_transport` imports is a narrower and more durable staging than evicting the `strands_robots.device_connect` subtree, because the eviction only simulated an absence for as long as that package's `__init__` happened to need the extra. That coupling is exactly what stopped holding, which is why these six went from green to a live network call without either branch involved being touched.

cagataycali:
Backing the last paragraph above with the measurement, since I stated it as a claim.

On a pristine worktree of `main` @ `828f80eb`, replaying what `_hide_extra` does (evict every `strands_robots.device_connect*` entry, set `device_connect_edge` to `None`) and then asking for the transport the resolver asks for:

```
evicted set (what delitem would record): ['strands_robots.device_connect',
                                          'strands_robots.device_connect.reachy_transport']
import during hidden extra: SUCCEEDED (absence not staged)
```

The import **succeeds**, so the staged absence is a no-op and the driver goes on to a real daemon probe. That is the root cause of the six, and it is a property of the staging rather than of anything either branch did to the driver.

Two things I checked and can rule out, so nobody re-treads them:

- **It is not a `sys.modules` restore leak.** Restoring only the recorded entries, as `monkeypatch.delitem` does, puts the same objects back: `module object identical after restore: True`, `api attribute identical after restore: True`, and the set of names present afterwards adds nothing the eviction had not recorded (`names created that delitem never recorded: []`).
- **The driver caches nothing.** `_resolve_transport` calls `importlib.import_module` on each use and holds no module reference, so there is no stale binding on the production side.

The ordering is also worth recording, because it is what makes the six look intermittent rather than broken:

```
pytest <file>::TestAWorkingInstallIsUnaffected          -> 1 passed
pytest <file>                                           -> 6 failed, 3 passed
pytest <file>::TestEachGuardIsReachedOnItsOwn <file>::TestAWorkingInstallIsUnaffected
                                                        -> 3 failed
```

`_connected_driver` is byte-identical between `main` and this branch, so a test that passes alone and fails after its sibling class is order-dependent state, not a bad helper.

One note for whoever generalises this. `tests/test_sys_modules_removal_leaves_no_orphan.py` grades exactly this hazard class and was silent here by its own documented scope: "Only a **literal** key is graded. A removal whose key is a variable (a loop purging `mujoco*`, say) is out of reach of a static read and is not claimed." `_hide_extra` is a prefix-sweep loop, so its key is a variable and it falls outside. The prefix itself, though, is a string literal, so the removed set is statically characterisable up to that prefix -- which suggests the exclusion is a little wider than the reason given for it. Not something to change on this branch, and it needs its own precision measurement over the open set before it is worth proposing at all; recording it here only because this outage is the case that would have motivated it.

yinsong1986 (approved):
## Summary

Combines the two interlocking Reachy test-flake fixes from #2785 and #2787 so neither blocks the other in CI. The transport-guard file now stages the missing extra at the one module `_resolve_transport` actually imports (`monkeypatch.setitem(sys.modules, _TRANSPORT_MODULE, None)`) instead of evicting the `strands_robots.device_connect` subtree, so the simulation stays tied to the resolver rather than to the package `__init__`'s import behaviour, and the premise test now reads executed module-scope imports via AST (correctly skipping `TYPE_CHECKING` blocks). The G1 identity test grades registration survival by `__module__` + `__qualname__` with an `assert registered is not None` guard ahead of it, so a genuinely lost registration still fails loudly while a re-import's fresh class object no longer false-negatives. Test-only diff (two test files, two changelog fragments); no production surface, public API, or wire format touched.

## What's good

- `_block_transport` follows AGENTS.md > "Restore a `sys.modules` entry you remove" exactly: `setitem(..., None)` both makes the import raise `ImportError` (which `_resolve_transport` catches) and restores atomically, touching nothing else in `sys.modules`.
- The `assert registered is not None` guard keeps the relaxed identity check from silently passing on an unregistered driver.
- Both flake mechanisms are documented in the changelog fragments with the measured failure signatures, and the control tests (`TestAWorkingInstallIsUnaffected`, the resolver-present control) are preserved.

## Verification suggestions

Beyond a full-suite run, the flake-triggering ordering is the interesting one: `pytest tests/drivers/test_g1_driver.py tests/drivers/test_reachy_transport_guards_are_reachable.py tests/drivers/test_reachy_driver.py` (g1 reload first, guards file before its sibling) passes cleanly on this head - 170 passed, 20 skipped, ~2s with no real connection timeouts.

cagataycali:
Measured this on the head `1e4ae799`. The Reachy-guard half is right and I have nothing to add to it. The G1 half rests on a stated cause I could not reproduce, and the assertion it lands is blind to a consequence that is still live.

## The cause is the reload, not a second registration pass

The new comment says the second class object comes from "the shipped-drivers registration loop running twice, once here and once elsewhere in the suite". Bisected against every file in `tests/drivers/` that sorts before `test_reachy_driver.py`, pairing each with the failing cell alone:

| file paired with the failing cell | fires |
|---|---|
| `test_dynamixel_driver.py` | 0 |
| `test_g1_dds_engine_release.py` | 0 |
| **`test_g1_driver.py`** | **1** |
| `test_g1_driver_reaches_the_mesh_wire.py` | 0 |
| `test_g1_lidar_state_reads_the_declared_fields.py` | 0 |
| `test_g1_lidar_summary_describes_the_cloud.py` | 0 |
| `test_g1_per_joint_gains.py` | 0 |

One file, and inside it one line — `tests/drivers/test_g1_driver.py:1388`:

```python
importlib.reload(importlib.import_module("strands_robots.drivers.g1"))
```

`importlib.reload` re-executes the module body into the same namespace, rebinding every class it defines. The registry captured `G1Driver` by reference at registration, so the two references diverge at that call. Replacing that one line with a subprocess probe makes the failing cell pass with the `is` check untouched; restoring it makes it fail again. That is `#2792`, which touches only `test_g1_driver.py` — no file in common with this PR (`comm -12` on the two file lists is empty).

## What the qualname check can no longer see

The comment's conclusion is that "the registration did survive, the identity check just cannot tell". Measured on this head, immediately after that same reload:

```
registry is module name : False
same __module__         : True
same __qualname__       : True
-> this PR's assertion passes: True

isinstance(registry-built instance, the module's G1Driver): False
issubclass(registry class,        the module's G1Driver): False
G1Driver class objects alive: 2
```

`isinstance` and `issubclass` both answer `False`. So the divergence is not confined to an `is` comparison the registry's job does not care about — any code that type-checks a driver the registry built against the name `strands_robots.drivers.g1.G1Driver` gets the wrong answer, and this assertion now passes through exactly that state. Two classes with equal `__module__` and `__qualname__` is the signature of a double-import, which is the one failure mode a driver registry most wants to be told about.

## Composition

The two are complementary rather than competing, and both are needed to land the suite green:

| tree | `pytest tests/drivers/` |
|---|---|
| `main` (`3b235185`) | `7 failed, 412 passed` |
| this PR alone | `371 passed` |
| `#2792` alone | `6 failed, 412 passed` |
| **this PR + `#2792`** | **`418 passed`** |

Merged in a scratch worktree: clean auto-merge, `grep -c '^<<<<<<<'` over `tests/` is 0, order free.

## Suggestion

Land both, and once the reload is gone restore the `is` check here — it is the stronger assertion and nothing will be tripping it. If you would rather keep this PR self-contained, keeping the tolerant form is harmless in the composition; the cost is only that the guard stops reporting a real double-import, and the comment attributing the cause to a repeated registration pass would be worth correcting either way, since it points a future reader away from the line that does it.

cagataycali:
Closing as superseded by #2792. The diagnosis in this description — that the two flakes deadlock because each branch is red on the other's defect — was correct, and #2792 now resolves that cycle the same way: both fixes on one branch. What changed is the treatment of the G1 half.

**Defect B (transport guards)** — included verbatim in #2792, changelog fragment retargeted `2785-` → `2793-`, with the Direction-1 choice #2793 asks for stated explicitly and the refusal-message question scoped out to #2794.

**Defect A (G1 identity)** — fixed at its cause rather than tolerated. This branch relaxed the assertion to compare `__module__` + `__qualname__`. #2792 removes the `importlib.reload` that creates the duplicate class, so the assertion needs no relaxation: `tests/drivers/test_reachy_driver.py` is untouched and `assert get_native_driver_class("unitree_g1") is G1Driver` passes as written. The registry keeps the stronger contract, and the SDK-import pin gets stronger too — it now proves *no* `unitree_sdk2py` module loads, which holds even where the SDK is installed.

Same end state, measured on `3b235185` with `pytest tests/drivers/`: `388 passed, 30 skipped`, from `7 failed, 381 passed, 30 skipped` on `main`.

Also worth recording, because it would have outlived this PR: the explanatory comment here attributed the duplicate class to "the shipped-drivers registration loop running twice, once here and once elsewhere in the suite". The loop runs once. The rebind is `importlib.reload` re-executing the module body into the same namespace, which is why the failure surfaced in an unrelated file.

PR #2792, opened 2026-08-26
test(drivers): make main's driver suite order-independent, at both root causes

### LINKED ISSUE #2793: tests/drivers: #2774 removed the eager import that test_reachy_transport_guards_are_reachable simulates, leaving six cells red on main
## What is red

Six cells in `tests/drivers/test_reachy_transport_guards_are_reachable.py` fail on
`main`. Measured on `828f80eb` with no branch content:

```
FAILED ...::TestEachGuardIsReachedOnItsOwn::test_the_daemon_probe_reports_an_error_body
FAILED ...::TestEachGuardIsReachedOnItsOwn::test_the_link_builder_returns_the_reason
FAILED ...::TestTheReasonReachesAMeshPeer::test_get_status_answers_and_carries_the_reason
FAILED ...::TestSendActionRefusesInsteadOfMisdiagnosing::test_the_refusal_does_not_blame_the_callers_keys
FAILED ...::TestSendActionRefusesInsteadOfMisdiagnosing::test_nothing_is_put_on_the_link
FAILED ...::TestAWorkingInstallIsUnaffected::test_a_connected_driver_still_sends_the_command
```

They are introduced by #2774. Same directory, same command, two adjacent commits:

| tree | `pytest tests/drivers/` |
|---|---|
| `8bdbb5ed` (parent of #2774) | `1 failed, 370 passed` |
| `828f80eb` (#2774 merged) | `7 failed, 364 passed` |

The one failure common to both is a separate defect fixed by #2792. The other six
arrive with #2774.

## Why

The file simulates a stock install through `_hide_extra`, and its module docstring
states the mechanism it relies on:

> :func:`_hide_extra` makes ``device_connect_edge`` unimportable and evicts the
> already-imported ``strands_robots.device_connect`` package **so the next import
> re-executes the `__init__` that fails**.

That eager `__init__` import is exactly what #2774 removed, which was the whole
point of #2771. So the re-import now succeeds, `_resolve_transport` returns the real
module, and `_build_link` hands back a live link where the test expects a reason
string:

```
    def test_the_link_builder_returns_the_reason(self, monkeypatch) -> None:
        _hide_extra(monkeypatch)
        driver = ReachyDriver(host="reachy.local")
>       assert _names_the_extra(driver._build_link(is_lite=True))
E       TypeError: argument of type 'WebSocketLink' is not iterable
```

The simulation stopped simulating. Nothing about the guards changed.

## The part that is a decision, not a repair

The file's `TestTheExtraIsAPackagingAccident` class states the thesis the whole suite
rests on: the transport leaf needs nothing from the extra, and the parent package is
what dragged it in. Measured, that is now true in the strong sense —
`strands_robots/device_connect/reachy_transport.py` imports only stdlib:

```
asyncio, functools, json, logging, math, os, socket, abc, collections.abc, typing
```

So after #2774 there is no absence of `[device-connect]` that makes the transport
unimportable, and `_resolve_transport`'s `ImportError` branch cannot be reached by the
route its own reason names:

```python
return (
    f"cannot import {_TRANSPORT_MODULE}: {exc} - the Reachy transport "
    "helpers ship behind an extra: pip install 'strands-robots[device-connect]'"
)
```

That leaves two coherent directions, and they differ in what ships:

1. **The guards stay as defence in depth.** The tests are repointed to make the
   transport module itself unimportable (`monkeypatch.setitem(sys.modules, _TRANSPORT_MODULE, None)`),
   which reaches the same `ImportError` branch and keeps all six cells grading the same
   guards. The reason string then names an extra that is no longer the cause, so its
   remedy is misleading for the only failures that can still reach it.
2. **The guards are spent.** #2774 removed the condition they were written for, so the
   four `_resolve_transport` call sites and the six cells go, and the driver's no-raise
   contract is carried by the pre-check alone.

Direction 1 is a two-line test change plus a docstring correction and leaves a refusal
whose remedy does not fit its cause. Direction 2 deletes a documented public refusal.
Both are defensible; picking one is a call about the driver's public failure surface,
so I have not guessed at it.

## Reproduction

```
git checkout 828f80eb
MUJOCO_GL=egl pytest tests/drivers/test_reachy_transport_guards_are_reachable.py -q
```

Six failures. Each also fails in isolation except
`TestAWorkingInstallIsUnaffected::test_a_connected_driver_still_sends_the_command`,
which passes alone and fails in the file — its `_connected_driver` helper patches
`strands_robots.device_connect.reachy_transport.api` by dotted string, so it resolves
whichever module object an earlier `_hide_extra` left cached.

## Why CI has not reported it yet

A whole-tree run stops earlier: `tests/drivers/test_g1_driver.py` sorts before
`test_reachy_driver.py`, which sorts before this file, and the #2792 defect fails
there first. Once #2792 lands, this becomes the first failure a full run reports.

cagataycali:
Repairs the two independent defects that make `tests/drivers/` order-dependent on `main`, each at its cause. Test-and-changelog only; no production surface is touched.

Fixes #2793

## The two defects

**A. A reload rebinds the driver class.** `test_g1_driver.py::test_g1_driver_module_does_not_import_unitree_sdk2py_at_load_time` pins a real contract — importing `strands_robots.drivers.g1` must not pull in the vendor SDK — but measured it with `importlib.reload` inside the running session. `reload` re-executes a module body into the *same* namespace, so it rebinds every class that body defines. The registry captured `G1Driver` **by reference** when the shipped table was registered, so afterwards the registry's class and the name's class are two distinct objects with an identical `repr`. Every later identity assertion fails, and the symptom surfaces in a different file:

```
tests/drivers/test_reachy_driver.py::TestTheLerobotPathIsUnaffected::test_the_g1_registration_survives_a_second_shipped_driver
E   AssertionError: assert <class 'strands_robots.drivers.g1.G1Driver'> is <class 'strands_robots.drivers.g1.G1Driver'>
```

**B. A simulation stopped simulating.** `test_reachy_transport_guards_are_reachable.py` staged a missing `[device-connect]` extra by making `device_connect_edge` unimportable and evicting the `strands_robots.device_connect` subtree, relying on the next import to re-execute an `__init__` that fails. #2774 removed that eager import deliberately (the point of #2771), so the re-import now succeeds, `_resolve_transport` returns the real module, and the driver reaches the daemon — six of nine cells failed on a real `urlopen [Errno 111] Connection refused` from a unit test.

## The changes

**A** — measure the import graph in a clean interpreter. Nothing in the session is rebound, and the contract becomes strictly *stronger*: **no** `unitree_sdk2py` module is loaded at all, rather than none beyond what an earlier test already imported — an assertion that holds even where the SDK is installed. A second cell reads the rebind directly, so the coupling is graded in the file that owns the pin.

**B** — stage the absence at the one module `_resolve_transport` imports, via `monkeypatch.setitem(sys.modules, _TRANSPORT_MODULE, None)`, touching nothing else in `sys.modules`. Per AGENTS.md > "Restore a sys.modules entry you remove". The premise test that should have caught the drift asserted `from device_connect_edge import` appeared in `__init__.py` — which the `TYPE_CHECKING` block satisfies, a block the source itself annotates `never executed`. It now reads executed module-scope imports via AST, so a typing-only import no longer reads as an eager one.

### Why B is staged rather than deleted

#2793 sets out two coherent directions and calls the choice a decision about the driver's public failure surface. This takes **Direction 1** (repoint the staging, keep the guards) because it is test-only and reversible, where Direction 2 deletes a documented public refusal. The residual question Direction 1 leaves — the reason string names an extra that is no longer the cause, so its remedy does not fit the only failures that can still reach it — is about a production message, not test hygiene, and is filed separately rather than decided here.

### Why the two are one PR

They cannot land apart. `tests/drivers/` on `main` is `7 failed`: one from A, six from B. A branch fixing only A is still red on B's six; a branch fixing only B is still red on A's one. Neither can show a green `call-test-lint`, which is what put #2785, #2787 and #2790 in a supersession cycle. Both are test-isolation repairs in the same directory, so landing them together is the smallest change that can be verified green.

## Verification

Measured on `3b235185` with `pytest tests/drivers/`:

| tree | result |
|---|---|
| `main` (pristine) | `7 failed, 381 passed, 30 skipped` |
| this branch | **`388 passed, 30 skipped`** |

Per-defect, on the file selection that reproduces the cross-file symptom (`test_g1_driver.py` + `test_reachy_driver.py`, in collection order):

| tree | result |
|---|---|
| `main` | `1 failed, 160 passed, 20 skipped` |
| defect-A fix only | `162 passed, 20 skipped` |

The guards file alone: `6 failed, 3 passed` on `main` → `9 passed` here.

**The registry contract is preserved, not relaxed.** `tests/drivers/test_reachy_driver.py` is not modified by this PR: `assert get_native_driver_class("unitree_g1") is G1Driver` stands unchanged at line 722 and passes, because the rebind that broke it is gone. Fixing A at its cause is what makes weakening that assertion unnecessary.

Gates: `ruff check` clean; `ruff format --check` clean; changelog fragment gates `89 passed` (`test_changelog_fragments.py`, `test_changelog_format.py`, `test_changelog_fragment_gate.py`).

No visual artifact: no policy, simulation, rendering, recording or asset behaviour changes.

## §13 Round changelog

**Round 2** — folded in the defect-B repair so this branch can show a green `call-test-lint`, and closed the supersession cycle:

- Added the transport-guard staging fix and its AST premise-test correction (defect B), previously carried by #2785 and #2790.
- Retargeted the guards changelog fragment from `2785-` to `2793-`, so it names the tracked issue rather than a superseded PR.
- Stated the Direction-1 choice #2793 asks for, and scoped the refusal-message question out of it.
- Supersedes #2785, #2787, #2790, all closed. Their approach to defect A graded the registration by `__module__` + `__qualname__` instead of `is`. The measurement above shows that relaxation is not needed: with the reload removed, the identity assertion passes as written. It also carried a comment attributing the second class object to "the shipped-drivers registration loop running twice", which is not the mechanism — the rebind comes from `importlib.reload` in `test_g1_driver.py`, as #2793 records.

**Round 1** — defect A only (`test_g1_driver.py`), with the mutation table for the pin.

### DIFF
--- a/changelog.d/2792-g1-sdk-import-pin-does-not-rebind-the-driver-class.md
+++ b/changelog.d/2792-g1-sdk-import-pin-does-not-rebind-the-driver-class.md
@@ -1,0 +1,28 @@
+### Fixed: the G1 SDK-import pin no longer rebinds the registered driver class
+
+`test_g1_driver_module_does_not_import_unitree_sdk2py_at_load_time` pinned a
+real contract - importing `strands_robots.drivers.g1` must not pull in the
+vendor SDK, so the module stays importable on a stock install - by reloading
+the driver module inside the running test session.
+
+`importlib.reload` re-executes a module body into the *same* namespace, so it
+rebinds every class that body defines. The driver registry captures
+`G1Driver` by reference when the shipped table is registered, so after the
+reload the registry's class and the class the module's own name resolves to
+are two distinct objects with an identical `repr`. Every identity assertion
+made later in the same session then fails, reporting
+`assert <class '...G1Driver'> is <class '...G1Driver'>` - a message that names
+the same class twice and points at neither the reload nor the file containing
+it.
+
+The pin now measures the import graph in a clean interpreter. Nothing in the
+test session is rebound, and the contract is stated more strongly: no
+`unitree_sdk2py` module is loaded at all, rather than none beyond whatever an
+earlier test in the session already imported - an assertion that holds even on
+a box where the SDK is installed.
+
+A second cell reads the rebind directly, calling the pin and then comparing the
+registry's class against the module's own name. Without it the file is silent:
+the symptom surfaced only in `tests/drivers/test_reachy_driver.py`, whose
+`test_the_g1_registration_survives_a_second_shipped_driver` compares those two
+references and runs after the pin in collection order.
--- a/changelog.d/2793-reachy-guards-stage-the-absence-at-the-resolver.md
+++ b/changelog.d/2793-reachy-guards-stage-the-absence-at-the-resolver.md
@@ -1,0 +1,29 @@
+### Fixed: the Reachy transport guards stage the absence at the module the resolver imports
+
+`tests/drivers/test_reachy_transport_guards_are_reachable.py` grades the two
+`_resolve_transport` guards that `ReachyDriver.connect_eagerly`'s pre-check
+hides, so a stock install is refused by name rather than crashing. It staged
+that absence by making `device_connect_edge` unimportable and evicting the
+`strands_robots.device_connect` subtree from `sys.modules`, relying on the next
+import to re-execute a package `__init__` that fails.
+
+That `__init__` no longer imports the extra - deferring it is the whole point of
+the lazy export table it now carries - so the eviction re-executes an `__init__`
+that succeeds, the stdlib-only transport leaf imports, and `_resolve_transport`
+hands back the real module. The driver then reaches the daemon: six of the
+file's nine cells failed with `daemon unreachable (localhost:8000): <urlopen
+error [Errno 111] Connection refused>`, a real network call from a unit test.
+
+The eviction had a second cost. Deleting the subtree let the next import build a
+fresh module object for `strands_robots.device_connect.reachy_transport`, so a
+double installed on one object was not read through the other. Run before its
+sibling, this file took `tests/drivers/test_reachy_driver.py` from 109 passed to
+46 passed / 72 failed, and the pair from 0.30s to 21.49s of real connection
+timeouts.
+
+The absence is now staged at the one module `_resolve_transport` imports, and
+nothing else in `sys.modules` is touched. The premise test that should have
+caught the drift asserted `from device_connect_edge import` appeared in
+`__init__.py`, which the `TYPE_CHECKING` block satisfies - a block the source
+itself annotates `never executed`. It now reads the executed module-scope
+imports, so a typing-only import no longer reads as an eager one.
--- a/tests/drivers/test_g1_driver.py
+++ b/tests/drivers/test_g1_driver.py
@@ -1382,16 +1382,48 @@ def test_g1_driver_module_does_not_import_unitree_sdk2py_at_load_time() -> None:
     it for the driver module itself so a future edit that reaches for a
     module-level ``import unitree_sdk2py.idl.default`` fails here.  Thor
     and CI both need the driver module importable without the SDK.
+
+    Measured in a clean interpreter rather than by reloading the module in
+    this one. :func:`importlib.reload` re-executes a module body into the same
+    namespace, which rebinds every class the body defines. The driver registry
+    captures :class:`~strands_robots.drivers.g1.G1Driver` by reference when the
+    shipped table is registered, so a reload leaves that reference pointing at
+    a class object that is no longer what the module's own name resolves to:
+    every later ``is`` comparison in the session then fails between two classes
+    with an identical ``repr``. A subprocess cannot rebind anything here, and it
+    states the contract more strongly - no SDK module is loaded at all, rather
+    than none beyond whatever an earlier test in this session already imported.
     """
-    import importlib
+    import subprocess
     import sys
 
-    # Snapshot the SDK modules already loaded (may or may not be present on
-    # this box).  Then reimport the driver module and confirm no *new*
-    # unitree_sdk2py submodule crept in - just importing the driver.
-    before = {n for n in sys.modules if n.startswith("unitree_sdk2py")}
-    importlib.reload(importlib.import_module("strands_robots.drivers.g1"))
-    after = {n for n in sys.modules if n.startswith("unitree_sdk2py")}
-    assert after == before, (
-        f"importing strands_robots.drivers.g1 pulled in unitree_sdk2py modules: {sorted(after - before)}"
+    probe = (
+        "import sys; import strands_robots.drivers.g1; "
+        "print(sorted(n for n in sys.modules if n.startswith('unitree_sdk2py')))"
     )
+    completed = subprocess.run([sys.executable, "-c", probe], capture_output=True, text=True, check=False)
+    assert completed.returncode == 0, f"importing the driver module failed: {completed.stderr}"
+    assert completed.stdout.strip() == "[]", (
+        f"importing strands_robots.drivers.g1 pulled in unitree_sdk2py modules: {completed.stdout.strip()}"
+    )
+
+
+def test_the_sdk_import_pin_leaves_the_registered_driver_class_reachable_by_name() -> None:
+    """The pin above must not leave a second ``G1Driver`` class behind.
+
+    The registry hands back the class object it was registered with. A pin that
+    re-executed the driver module in this interpreter would rebind the module's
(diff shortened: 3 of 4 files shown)


yinsong1986 (approved):
## Summary

Test-only PR with two halves, both verified locally against the current head. First, the G1 SDK-import pin previously used `importlib.reload` inside the running session, which rebinds every class the driver module body defines and leaves the registry's by-reference `G1Driver` pointing at a stale class object - the cross-file `is` assertion in `tests/drivers/test_reachy_driver.py` then fails naming the same class twice. The pin now measures the import graph in a clean subprocess (`sys.executable -c`, fixed probe string, no untrusted input), removing the rebind and strengthening the contract to "no `unitree_sdk2py` module is loaded at all"; a second test reads the rebind directly by calling the pin and comparing the registry's class against the module's own name. Reproduced the defect on `main` (`test_g1_driver.py` + `test_reachy_driver.py` -> 1 failed) and confirmed the same selection is green on this branch. Second - added after the PR description was written, which still says the Reachy failures were "filed as #2793 rather than folded in here" - the head now also fixes `test_reachy_transport_guards_are_reachable.py`: the absence is staged at the one module `_resolve_transport` imports (`monkeypatch.setitem(sys.modules, name, None)`, the exact pattern AGENTS.md sanctions) instead of evicting the `strands_robots.device_connect` subtree, and the premise test now reads executed module-scope imports via AST rather than matching text a `TYPE_CHECKING` block satisfies. With that, `tests/drivers/` goes fully green here (388 passed, 30 skipped), where `main` carried 7 failures. The description drift is worth knowing about but the folded-in fix is sound and carries its own fragment (`changelog.d/2793-...`, and issue-numbered fragments are explicitly allowed by `changelog.d/README.md`). No production code, public API, wire format, or persisted schema is touched.

## What's good

- Both changelog fragments follow the `changelog.d/<number>-<slug>.md` convention with a single `### Fixed:` entry each.
- The `sys.modules` staging change follows AGENTS.md > Testing Patterns > "Restore a `sys.modules` entry you remove" verbatim - the old eviction pattern was precisely the orphaned-reference hazard that bullet describes, and the fragment measures its cost (72 cross-file failures, 21s of real connection timeouts).
- Test names describe behaviour, not provenance; docstrings carry the reasoning with intact single-line cross-reference targets.
- The mutation table in the description demonstrates each new assertion is independently load-bearing, including the row showing the g1 file is silent without the new identity cell.

## Verification suggestions

- `pytest tests/drivers/ -q` on the branch: fully green here (388 passed, 30 skipped) vs 7 pre-existing failures on `main`.
- To confirm the g1 defect this fixes: on `main`, `pytest tests/drivers/test_g1_driver.py tests/drivers/test_reachy_driver.py -q` fails `test_the_g1_registration_survives_a_second_shipped_driver`; on the branch the same selection passes.

yinsong1986 (approved):
## Summary

Test-only repair of the two independent order-dependence defects in `tests/drivers/`, each fixed at its cause. Defect A: the G1 SDK-import pin used `importlib.reload` in-session, rebinding `G1Driver` out from under the registry's by-reference capture and breaking a cross-file `is` assertion; the pin now measures the import graph in a clean subprocess (a strictly stronger contract - no `unitree_sdk2py` module loaded at all), and a new sibling test grades the rebind coupling in the file that owns the pin. Defect B: the transport-guard staging evicted the `strands_robots.device_connect` subtree and relied on a failing eager `__init__` import that #2774 removed, so six cells reached a real daemon; the absence is now staged with `monkeypatch.setitem(sys.modules, _TRANSPORT_MODULE, None)` at the one module `_resolve_transport` actually imports, and the AST premise test now reads executed module-scope imports so a `TYPE_CHECKING` import cannot pass for an eager one. Verified against the head sources: `_TRANSPORT_MODULE` is the production constant `_resolve_transport` imports, its `except ImportError` catches the blocked-entry case, and the resulting reason string satisfies `_names_the_extra`. No production, wire, or public-API surface is touched.

## What's good

- The `sys.modules` staging follows AGENTS.md > "Restore a `sys.modules` entry you remove" exactly: one `setitem(..., None)` entry, nothing else evicted, monkeypatch restores.
- The registry identity contract in `tests/drivers/test_reachy_driver.py` is preserved unmodified rather than relaxed, which is the measurable difference from the superseded #2785/#2787/#2790 approach.
- The changelog fragments explain mechanism, not just symptom, and the Direction-1 choice on #2793 is stated with the residual production-message question explicitly scoped out.

Issue #2795, opened 2026-08-26
tests/drivers/test_reachy_transport_guards_are_reachable.py is red on main: 6 of 9 fail, 5 standalone


cagataycali:
## `tests/drivers/test_reachy_transport_guards_are_reachable.py` is red on `main`: 6 of 9 fail, 5 of them standalone

`main` at `3b235185` fails six tests in this file. Five fail with no other test in the
session, so this is not an ordering flake. The file is byte-identical to the one that
landed with #2762 (`md5 95b0de6b811cf878b23b330a1a2668e3` on both trees), and it passed
there, so the change is in what it exercises rather than in the file.

### Reproduction

```
git checkout 3b235185
PYTHONPATH=$PWD MUJOCO_GL=egl python -m pytest \
  tests/drivers/test_reachy_transport_guards_are_reachable.py -q -p no:randomly
```

`6 failed, 3 passed`. The three that pass are `TestTheExtraIsAPackagingAccident`, which
never brings a driver up.

Each of the six run on its own:

| test | alone |
|---|---|
| `TestEachGuardIsReachedOnItsOwn::test_the_daemon_probe_reports_an_error_body` | fails |
| `TestEachGuardIsReachedOnItsOwn::test_the_link_builder_returns_the_reason` | fails |
| `TestTheReasonReachesAMeshPeer::test_get_status_answers_and_carries_the_reason` | fails |
| `TestSendActionRefusesInsteadOfMisdiagnosing::test_the_refusal_does_not_blame_the_callers_keys` | fails |
| `TestSendActionRefusesInsteadOfMisdiagnosing::test_nothing_is_put_on_the_link` | fails |
| `TestAWorkingInstallIsUnaffected::test_a_connected_driver_still_sends_the_command` | passes |

### Which change it arrived with

Same file, two trees, `device_connect_edge` importable in both:

| tree | this file | `tests/drivers/` |
|---|---|---|
| `97ab1b02` (#2762 merged) | 9 passed | 1 failed |
| `3b235185` (current `main`) | 6 failed, 3 passed | 7 failed |

The only change between them touching this area is #2774, which moved
`strands_robots/device_connect/__init__.py` to a lazy `__getattr__` over `_impl`.

The single failure at `97ab1b02` is separate and predates it:
`test_reachy_driver.py::TestTheLerobotPathIsUnaffected::test_the_g1_registration_survives_a_second_shipped_driver`
[...]

Issue #2791, opened 2026-08-26
ci(merge-base-overlap): the file-path intersection is blind to a test that reads a file the base changed - main went red from #2762 + #2774 with the check green on both


cagataycali:
`main` went red again from two individually green, textually non-conflicting pull requests - the exact failure `merge-base-overlap.yml` was added for after `0e636f8` (#1770). The check reported SUCCESS on both, because it intersects **file paths** and these two share none.

## What happened

| | merged | changed |
| --- | --- | --- |
| #2762 | 2026-08-26 12:55 | *added* `tests/drivers/test_reachy_transport_guards_are_reachable.py` |
| #2774 | 2026-08-26 13:14 | *edited* `strands_robots/device_connect/__init__.py` |

Nineteen minutes apart. #2774 deferred `device_connect_edge` out of the package `__init__` (the point of its lazy export table). #2762's new file stages the extra's absence by evicting the `strands_robots.device_connect` subtree and relying on the next import to re-execute an `__init__` that *fails* - and it no longer does. Six of that file's nine cells then made real network calls from a unit test:

```
daemon unreachable (localhost:8000): <urlopen error [Errno 111] Connection refused>
```

Neither branch's own run could see it: #2774 forked before #2762 landed, so its `call-test-lint` never executed the file it broke. The overlap check computes the intersection of *the files the branch edits* with *the files the base changed since divergence*; that intersection was empty, so the job passed.

## Why the intersection is empty

The dependency runs the other way from the one the check models. #2762 did not *edit* `device_connect/__init__.py` - it added a test whose behaviour **depends on** that file, both by importing through it and by reading its source text. A file-path intersection cannot see "B's new test reads A's file"; it only sees "A and B both wrote the same file", which is what #1766/#1763 did.

So the check is correct for the case it documents and structurally blind to this one. The header's own framing - "a green result is evidence about a different tree than the one that would be merged" - is exactly true here; only the detector missed it.

Worth noting the same window produced a second, independent order-dependence in the same suite (the G1 load probe's `importlib.reload` orphaning the driver registration). Both are repaired test-only in #2785; this issue is only about the gap that let the first one land.

## Evidence

- #2774's merge commit `828f80e`; #2762's `97ab1b0`
- overlap job on #2785 head `aff60de`: `Detect an untested overlap with the base branch` = SUCCESS, while `call-test-lint` = FAILURE on a base-branch defect (job `98267163585`)
[...]

PR #2797, opened 2026-08-26
fix(ci): collide two branches on a module one of their tests names, not only on a path

cagataycali:
`scripts/check_merge_base_overlap.py` intersects changed **paths**. `main` went red at `828f80eb`
from #2762 and #2774 with that intersection not merely unhelpful but empty, so the sweep reported
`no pair in the open set shares a changed path` while both branches sat in the queue.

## What composed red

#2762 added `tests/drivers/test_reachy_transport_guards_are_reachable.py`. Its fixture reaches into
a module by name, twice:

```python
for name in [n for n in list(sys.modules) if n.startswith("strands_robots.device_connect")]:
    monkeypatch.delitem(sys.modules, name)
...
monkeypatch.setattr("strands_robots.device_connect.reachy_transport.api", ...)
```

#2774 rewrote `strands_robots/device_connect/__init__.py` to resolve its names lazily. Six of that
file's nine tests failed on the composition, five of them with no other test in the session
(#2795). The two branches' changed-path sets:

| pull request | changed paths |
| --- | --- |
| #2762 | 2 changelog fragments, `docs/getting-started/robot-factory.md`, `strands_robots/drivers/{__init__,reachy}.py`, `strands_robots/registry/robots.json`, `strands_robots/tools/reachy/{__init__,_reachy_common}.py`, `tests/drivers/test_reachy_driver.py`, `tests/drivers/test_reachy_transport_guards_are_reachable.py` |
| #2774 | 1 changelog fragment, `strands_robots/device_connect/{__init__,_impl}.py`, `tests/device_connect/test_package_init_does_not_import_the_extra.py`, `tests/test_device_connect_sync_bringup_outcomes.py` |

Intersection: `[]`. Nothing in the sweep could describe the pair (#2791).

## Why the existing exclusion did not cover it

The file already measures and rejects widening the path set to a test's *walked root*, on the
grounds that a relation firing on 11 of 36 pairs while naming no defect "reads as boilerplate".
That argument is sound and it is about a different case: a population resolved by `rglob` never
names what it grades. A test that reaches into a module **by name** writes the coupling down, and a
dotted name is as explicit as a path -- it just is not one.

## The relation

The dotted module names a pull request's *test* diff writes as string literals, resolved to the
module files they name. Every prefix resolves, because importing `a.b.c` executes
`a/b/__init__.py` -- and that prefix is two segments shallower than the literal #2762 wrote, which
is exactly where #2774's edit was.

Scope is deliberate on three axes, each pinned:

- **Test modules only.** Production code that needs another module imports it, and an import is a
  path the diff already carries.
- **The whole contents of the string.** `"strands_robots.mesh.core"` is a name handed to
  `importlib` or `monkeypatch.setattr`; the same characters inside a sentence are prose about a
  module, not a reach into one.
- **Not the bare package root.** Resolving `strands_robots` alone adds four findings, all of them
  pairs with one branch's edit to `strands_robots/__init__.py`, which every literal in the tree
  names by its first segment. That is the shallowest coupling expressible and the one a reader can
  least act on.

## Measurement

Over the 2676 pairs in #2309 through #2792 that were open at the same instant, with both relations
run over identically reconstructed inputs:

| relation | pairs selected | share of population | not already reported |
| --- | --- | --- | --- |
| changed-path intersection | 104 | 3.9% | -- |
| named module (this change) | 25 | 0.93% | **9** |
| walked root (rejected above) | 11 of 36 | 31% | 9, none a defect |

The nine name couplings a reader can act on:

| pair | module | why it is a coupling |
| --- | --- | --- |
| #2774 + #2762 | `strands_robots/device_connect/__init__.py` | the composition that broke `main` |
| #2767 + #2762 | `strands_robots/drivers/__init__.py` | driver registration table, the file behind `main`'s *other* red test |
| #2767 + #2750 | `strands_robots/drivers/__init__.py` | same table, third contender |
| #2546 + #2545 | `strands_robots/mesh/__init__.py` | #2546 removes an export; #2545's test names that package |
| #2515 + #2510, #2510 + #2504, #2314 + #2312 | `strands_robots/mesh/core.py` | mesh branch beside a test naming a mesh module |
| #2515 + #2486 | `strands_robots/tools/__init__.py` | tools package edit beside a test naming a tools module |
| #2427 + #2331 | `strands_robots/rendering/__init__.py` | a docstring-reference grader beside new rendering exports |

The difference from the rejected widening is the input, not a threshold: a literal is written by the
test author and says which module is reached, where a walked root is inferred from a glob and says
only that some file under it might be.

## Cost

None. `patch` arrives beside `filename` on the entries the path set is already built from, so the
literals are a second reading of a payload already in hand -- pinned by a test that counts
`/files?` fetches. Both modes hand `(path, patch)` pairs to one extractor (the API's `patch` for the
sweep, `git diff` for the single-branch check), so neither can disagree with the other about what a
literal is, which is the parity property the file already asserts for its path-set helpers.

## Tests

`tests/test_merge_base_overlap.py` gains 13 cases; the git-topology ones replay the incident as real
commits in a real repository rather than as a hand-built path set, matching the file's existing
convention.

Reverting `scripts/check_merge_base_overlap.py` to this branch's base and keeping the tests gives
**11 failed, 59 passed**. The load-bearing failures are the four `assert 0 == 1` -- the gate
reporting clean, exit status 0, on a composition that broke `main`, including the #2762/#2774 replay
in the sweep. All five over-reach controls pass *both* before and after, which is what makes them
controls rather than regression cells:

| case | pre-fix | role |
| --- | --- | --- |
| branch names a module the base rewrote | fails | regression |
| the #2762/#2774 pair in the sweep | fails | regression |
| base changed a module the branch's tests name | fails | regression |
| merging the base clears the finding | fails | self-clearing |
| the path intersection here really is empty | passes | premise |
| a module named but changed by nobody | passes | control |
| prose naming a module | passes | control |
| a literal in production code | passes | control |
| a module the branch also edits, reported once | passes | control |
| the walk blind spot is still named | fails | scope |

## Gate

- `ruff check` and `ruff format --check` clean.
- `mypy strands_robots tests tests_integ`: `Success: no issues found in 1604 source files`, nothing
  attributable.
- 161-file selection (every suite that walks the tree, parses source, reads docstrings, reads
  `scripts/` or reads `changelog.d/`): `5 failed, 6773 passed, 56 skipped`. The five are
  pre-existing and environmental -- they assert `rclpy` is absent and it resolves in this
  environment; the identical five fail on this branch's merge base.
- `tests/test_merge_base_overlap.py`: 70 passed.
- Run against the live open set: 4 pull requests, 6 pairs, no findings, exit 0 -- no false positive
  on the current queue, this pull request included.

No visual artifact: this changes a repository check, its report text and its tests, none of the
surfaces that call for one.

Refs #2791, #2795.

### DIFF
--- a/changelog.d/2797-overlap-named-module-relation.md
+++ b/changelog.d/2797-overlap-named-module-relation.md
@@ -1,0 +1,35 @@
+### Fixed: two branches are collided on a module one of their tests names, not only on a path they share
+
+`scripts/check_merge_base_overlap.py` intersected changed **paths**. `main` went red at
+`828f80eb` from #2762 and #2774 with that intersection not merely unhelpful but empty. #2762
+added `tests/drivers/test_reachy_transport_guards_are_reachable.py`, whose fixture evicts
+`"strands_robots.device_connect"` from `sys.modules` and patches
+`"strands_robots.device_connect.reachy_transport.api"` by string; #2774 rewrote
+`strands_robots/device_connect/__init__.py` to resolve its names lazily. Six of that file's nine
+tests failed on the composition, five of them standalone, and the two branches changed no path in
+common -- so the sweep reported `no pair in the open set shares a changed path` while both sat in
+the queue, and the first tree in which the two were compiled together was `main`.
+
+The exclusion was reasoned rather than accidental, and the reasoning covered a different case.
+The file already measured and rejected widening the path set to a test's *walked root*, on the
+grounds that a relation firing on 11 of 36 pairs and naming no defect reads as boilerplate. That
+argument is about a population a grader never names. A test that reaches into a module *by name*
+writes the coupling down, and the name is not a path.
+
+There is now a second relation over the same two path sets and one further input: the dotted
+module names a pull request's test diff writes as string literals, resolved to the module files
+they name. Every prefix resolves, because importing `a.b.c` executes `a/b/__init__.py`, and that
+prefix is two segments shallower than the literal #2762 wrote -- it is where the #2774 edit was.
+Measured over the 2676 co-open pairs in #2309 through #2792, the path relation selects 104 pairs
+and this one selects 25, of which 9 are not already reported: 0.34% of the population against the
+rejected widening's 31%. The nine name couplings a reader can act on -- #2774 + #2762 is the
+composition above, #2767 + #2762 and #2767 + #2750 are three driver branches contending for
+`strands_robots/drivers/__init__.py`'s registration table, and #2546 + #2545 pairs a branch
+removing an export from `strands_robots/mesh/__init__.py` with one whose test names that package.
+
+The bare package root is deliberately not resolved: including it adds four findings, all of them
+pairs with one branch's edit to `strands_robots/__init__.py`, which every literal in the tree
+names by its first segment. Literals are read from the `patch` field of the entries the path set is
+already built from, so the sweep makes no additional request, and both modes hand `(path, patch)`
+pairs to one extractor so neither can disagree with the other about what a literal is. What stays
+out of reach is unchanged and is still the filesystem walk, which the report continues to name.
--- a/scripts/check_merge_base_overlap.py
+++ b/scripts/check_merge_base_overlap.py
@@ -131,6 +131,45 @@ directory that is not a repository). So the honest change here is scope, not a
 new heuristic: this mode reports what it measured -- shared paths -- and names
 the composition class it cannot describe. See #2561.
 
+A module a test names is not a coupling it declined to write down
+-----------------------------------------------------------------
+The paragraph above is about a population the grader never names. A test that
+reaches into a module *by name* is the opposite case, and it produced the same
+outcome: ``main`` went red at ``828f80eb`` from #2762 and #2774 with the pairwise
+path intersection not merely unhelpful but ``[]``. #2762 added
+``tests/drivers/test_reachy_transport_guards_are_reachable.py``, whose fixture
+evicts ``"strands_robots.device_connect"`` from ``sys.modules`` and patches
+``"strands_robots.device_connect.reachy_transport.api"`` by string; #2774 rewrote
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Adds a second relation to `scripts/check_merge_base_overlap.py`: the dotted module names a pull request's *test* diff writes as string literals, resolved (prefixes included, bare package root excluded) against the files the base or a sibling changed. This is the relation that would have caught the #2762/#2774 composition that broke `main` at `828f80eb` with an empty path intersection. Scope is tight on all three documented axes and each is pinned by a control test: test modules only, whole-string literals only (prose and partial matches don't fire), and the bare `strands_robots` root is not resolved. The literals are read from the `patch` field of entries both modes already fetch, so the sweep adds no API request (pinned by a fetch-counting test), and both modes share one extractor so they cannot drift. The new finding class fails in the safe direction — it only adds findings and self-clears by merging the base, which a test pins. Verified locally at the head SHA: all 70 tests in `tests/test_merge_base_overlap.py` pass, `ruff check` and `ruff format --check` are clean, files are ASCII-only, and the extractor's scope boundaries (whole-string, test-only, no bare root, prefix resolution) reproduce exactly as documented.

## What's good

- Changelog fragment (`changelog.d/2797-overlap-named-module-relation.md`) per AGENTS.md step 3.
- The incident is replayed as real git topology and as a sweep payload, with premise/control cells (`the path intersection here really is empty`, `prose naming a module`, `a literal in production code`) that pass both before and after the fix — genuine controls, not regression cells.
- The rejected walked-root widening stays rejected and the report still names the walk blind spot unconditionally, pinned by its own test.
- The relation's measured selectivity (25 of 2676 pairs, 9 new) is recorded in the module docstring beside the measurement that rejected the wider relation, so the next reader can compare like with like.

PR #2798, opened 2026-08-26
fix(drivers): stop the Reachy transport refusal prescribing an install it cannot establish

### LINKED ISSUE #2794: drivers/reachy: _resolve_transport prescribes the [device-connect] extra for failures it can no longer be the cause of
Carved out of #2793, which asked whether the Reachy transport guards should stay (repoint the test) or go (delete the refusal). #2792 takes the first direction because it is test-only and reversible. That leaves one thing unresolved, and it is a production message rather than test hygiene, so it is filed here instead of decided there.

## The mismatch

`_resolve_transport` refuses with a reason that names the `[device-connect]` extra as the cause and installing it as the remedy:

```python
return (
    f"cannot import {_TRANSPORT_MODULE}: {exc} - the Reachy transport "
    "helpers ship behind an extra: pip install 'strands-robots[device-connect]'"
)
```

After #2774 that remedy no longer fits the cause. `strands_robots/device_connect/reachy_transport.py` imports only stdlib:

```
asyncio, functools, json, logging, math, os, socket, abc, collections.abc, typing
```

So a missing `[device-connect]` extra can no longer make the transport unimportable. The `ImportError` branch is still reachable — #2792 pins it by blocking the module directly — but every failure that can now reach it has some *other* cause: a corrupt install, a shadowing module, a partial wheel. For all of those, `pip install 'strands-robots[device-connect]'` is advice that will not help, offered with the confidence of a diagnosis.

That is the failure mode the guards exist to prevent, pointed the wrong way: the driver refuses by name instead of crashing, but the name is wrong.

## What to decide

The reason string should either report the import failure without asserting a cause it can no longer establish, or establish the cause before naming it. Roughly:

1. **Report, do not diagnose.** Surface `exc` and the module that failed, and drop the extra from the remedy. Loses the one genuinely helpful hint for anyone on a pre-#2774 install.
2. **Diagnose, then name.** Keep the extra in the message only when `device_connect_edge` is in fact absent, and fall back to a cause-free report otherwise. More faithful, and it is a branch plus a probe.

Either way the pin belongs in `tests/drivers/test_reachy_transport_guards_are_reachable.py`, whose `TestTheExtraIsAPackagingAccident` class already grades the premise this rests on (the leaf needs nothing from the extra) and, as of #2792, reads the executed module-scope imports rather than the source text — so a `TYPE_CHECKING` import no longer reads as an eager one.

## Not urgent

Nothing is red. `tests/drivers/` is `388 passed, 30 skipped` with #2792. This is a message-accuracy defect on a path that only opens on a broken install, so it is worth getting right rather than fast.

## Context

- #2771 — the eager import that made the extra load-bearing
- #2774 — removed it, which is what invalidated the remedy
- #2793 — the six red cells, and the two directions
- #2792 — takes Direction 1 and defers this

cagataycali:
Closes #2794.

`_resolve_transport` refused an unimportable transport with a reason that named the
`[device-connect]` extra as the cause and installing it as the remedy. That was true
once: the transport's parent package imported `device_connect_edge` at module scope, so
the extra's absence really did make the leaf unimportable. That import is now lazy, and
the leaf itself imports nothing outside the standard library, so nothing an extra
installs decides whether this import succeeds. The remedy outlived its cause.

## The extra can no longer be the cause

Measured with `device_connect_edge` made unimportable:

| import | result |
| --- | --- |
| `device_connect_edge` | `ImportError: No module named 'device_connect_edge'` |
| `strands_robots.device_connect` | imports |
| `strands_robots.device_connect.reachy_transport` | imports |

The leaf's executed module-scope imports are `abc, asyncio, collections.abc, functools,
json, logging, math, os, socket, typing` - ten names, all standard library. So no
`pip install` supplies a module whose absence can reach that `except ImportError`. Every
cause that still can - a shadowing module, a partial wheel, a corrupt install - is
untouched by installing an extra.

That makes it the case `require_optional` already refuses to print a pip line in. When
it is told a module arrives from somewhere pip cannot reach it drops the install block
entirely, and says why in its own docstring: such a command "is a remedy the caller can
follow to no effect: it either installs something that leaves the module exactly as
missing, or fails outright", and naming one "would hand the caller an instruction that
reports success without supplying the module". Two shipped call sites rely on that for
the ROS 2 client libraries.

## The declared sibling was already the answer

`_resolve_transport`'s docstring opens "The same shape as
`drivers.g1._resolve_message_class`". Handed the same blocked module the two now answer
identically; before, one appended a cause and a remedy the other does not:

| | reason for a blocked module |
| --- | --- |
| `g1._resolve_message_class` | `cannot import <module>: <exc>` |
| `_resolve_transport`, before | `cannot import <module>: <exc> - the Reachy transport helpers ship behind an extra: pip install 'strands-robots[device-connect]'` |
| `_resolve_transport`, after | `cannot import <module>: <exc>` |

The parity claim is now graded by asserting the two strings are equal, so a re-added
remedy fails whatever wording it arrives in.

## What the reason keeps

Both facts a caller can act on. For a genuinely absent module:

```
cannot import strands_robots.device_connect.reachy_transport: No module named
'strands_robots.device_connect.reachy_transport'
```

The module is pinned as a **prefix** rather than by containment - for a blocked module
the `ImportError` text happens to repeat the dotted name, so a containment check passes
even when the reason's own prose has stopped naming it. That gap was found by a mutation
firing one test where it should have fired two.

## Scope: one of the two sites

`strands_robots/robot.py` names the same extra when Device Connect bring-up fails, and
there it is correct - that path imports `init_device_connect_sync`, which resolves
through modules that *do* import `device_connect_edge` at module scope. It is untouched,
and pinned both ways: one cell asserts the remedy survives there, another asserts some
module on that path really does import the dependency eagerly. Four modules do.

## The reference page quoted the old reason

`docs/getting-started/robot-factory.md` showed the refusal as `connect_eagerly()`'s
output, carrying the remedy and a `ModuleNotFoundError` naming a package that can no
longer be the cause, under the claim that the transport "ships in the `device-connect`
extra". Its own suite grades other sections of the page - `grep` for the extra in
`tests/test_docs_robot_factory_reference.py` returns nothing - so the quote was ungraded
and drifted. It is corrected, and a cell now derives the expected text from the resolver
and compares it to the page, so the two cannot separate again.

## Tests

The three surfaces that assert this reason previously asserted the extra string, one of
them in a cell named `test_the_reason_names_the_extra_to_install` inside a class named
`TestAStockInstallIsRefusedByNameRatherThanCrashing`. Both names encoded the stale
premise, so both are renamed for the behaviour that is actually true, and every
assertion is repointed from "names the extra" to "names the module, carries the cause,
prescribes no install" - strictly more specific than what it replaced.

Pre-fix split with the source reverted and every test kept: **9 failed, 117 passed**.
Per class:

| --- | --- | --- |
| `TestTheReasonPrescribesNothingItCannotEstablish` | 3 / 6 | regression, plus the premise and two "must not be lost" cells that hold either way |
| `TestTheDocumentedReasonIsTheRealOne` | 2 / 2 | regression: the page and the code agree |
| `TestEachGuardIsReachedOnItsOwn` | 2 / 2 | repointed |
| `TestTheReasonReachesAMeshPeer` | 1 / 1 | repointed |
| `TestAnUnimportableTransportIsRefusedByNameRatherThanCrashing` | 3 / 6 | repointed |
| `TestTheSiteWhereTheExtraIsStillTheCauseKeepsIt` | 0 / 2 | over-reach control |
| `TestTheExtraIsAPackagingAccident` | 0 / 3 | premise |
| `TestAWorkingInstallIsUnaffected` | 0 / 1 | control |
| every other class in both files | 0 / 79 | unaffected |

## Mutations

Selection is `tests/drivers/` plus `tests/test_docs_robot_factory_reference.py`. `coll`
is 435 on every row, so nothing is hidden by a deselection.

| mutation | new cells | other | firing (new) |
| --- | --- | --- | --- |
| control | 0 | 0 | |
| the extra remedy restored verbatim | 4 | 6 | page-quote, parity, no-extra, no-install |
| remedy re-added in other wording | 3 | 6 | page-quote, parity, no-install |
| extra named without a pip line | 3 | 0 | page-quote, parity, no-extra |
| module name dropped from the reason | 3 | 0 | page-quote, parity, still-names-module |
| underlying `ImportError` dropped | 4 | 1 | page-quote, parity, still-carries-error, still-names-module |
| trailing prose, no pip and no extra | 2 | 0 | page-quote, parity |
| over-reach: `robot.py` remedy swept too | 1 | 0 | bring-up-path-keeps-the-extra |
| leaf grows a third-party import | 1 | 0 | leaf-is-stdlib-only |
| docs quote reverted, code left fixed | 2 | 0 | page-quote, page-prescribes-no-install |

Nine mutations, nine distinct firing sets, control clean, both source files restored
byte-identically after the run. The rows that matter: the over-reach row shows this is
not a sweep of the extra's name; the third row shows the rule is about the claim rather
than the `pip install` substring; the last shows the docs pin catches drift in the
direction that produced this defect. The parity cell fires on every code-side mutation,
which is the point of grading against the sibling rather than against a literal.

## Gate

  (1607 files).
- `mypy` **24 == 24** against the merge base, normalized message sets byte-identical in
  both directions, 0 attributable to the changed files.
- Paired 449-file selection (all of `tests/drivers/`, plus every suite that walks the
  tree, parses source, reads docstrings or reads `docs/`, plus everything naming the
  driver or the package): **22 failed on both trees, failing-ID sets identical, `comm`
  empty in both directions**, collected 21763 -> 21773 for exactly the ten added cells.
- The 22 are pre-existing and unrelated: 17 need `awsiot`/`awscrt` (declared in the
  `[mesh-iot]` extra, absent here), 3 are lerobot-from-source drift in
  `tests/training/test_lerobot*`, 2 are `tests/test_device_connect_hardening.py`.
- `tests/drivers/` is **425 -> 435 passed**, and the changelog and docs gates pass
  (524 passed together).

No visual artifact: this changes a refusal string, two docstrings, a docs page and
tests - none of the surfaces that call for one. The measured before/after reason and the
mutation table are the evidence.

### DIFF
--- a/changelog.d/2798-reachy-transport-reason-reports-without-prescribing.md
+++ b/changelog.d/2798-reachy-transport-reason-reports-without-prescribing.md
@@ -1,0 +1,26 @@
+### Fixed: the Reachy transport refusal no longer prescribes an install it cannot establish
+
+`ReachyDriver` resolves its daemon transport lazily and reports a reason rather than
+raising when that import fails. The reason named the `[device-connect]` extra as both
+the cause and the remedy - correct once, when the transport's parent package imported
+`device_connect_edge` at module scope and the extra's absence really did make the leaf
+unimportable. That import is now lazy, and the leaf itself imports nothing outside the
+standard library, so no `pip install` supplies a module whose absence can reach that
+branch. Every cause that still can - a shadowing module, a partial wheel, a corrupt
+install - is untouched by installing an extra, so the remedy was advice that cannot
+help offered with the confidence of a diagnosis, which is the failure mode the named
+refusal exists to prevent, pointed the wrong way.
+
+The reason now reports the module and the underlying `ImportError` and stops there.
+That is the shape `drivers.g1._resolve_message_class` already uses, which this
+function's own docstring documents itself as sharing, and it is the position
+`utils.require_optional` refuses to print a pip line in when it is told a module
+arrives from somewhere pip cannot reach: such a line "would hand the caller an
+instruction that reports success without supplying the module".
+
+`docs/getting-started/robot-factory.md` quoted the old reason verbatim, including the
+remedy and a `ModuleNotFoundError` naming a package that can no longer be the cause.
+Nothing graded that block, so it is corrected against the resolver and pinned to it.
+The one other site that names the same extra is left alone: it reports a Device
+Connect bring-up failure, on a path whose modules do import `device_connect_edge` at
+module scope, so there the extra is a real cause and installing it is a real remedy.
--- a/docs/getting-started/robot-factory.md
+++ b/docs/getting-started/robot-factory.md
@@ -131,22 +131,28 @@ has a class for it, so `driver="lerobot"` remains a usable fallback. The Reachy 
 declares none: lerobot has no robot type for it, so the native driver is the only way
 to reach it and `driver="lerobot"` is refused by name.
 
-A native driver may need an extra to reach its hardware, and says so rather than
-raising. The Reachy Mini's daemon transport ships in the
-[`device-connect`](../device-connect.md) extra, so on a core install the driver still
-builds, registers and answers `get_status` - and every surface that would touch the
-daemon returns a reason naming the extra instead:
+A native driver reports what it cannot reach rather than raising. The Reachy Mini's
+daemon transport is a standard-library-only module in the core distribution, so
+nothing an extra installs decides whether it loads - but if it cannot be imported at
+all, on a broken install or behind a shadowing module, the driver still builds,
+registers and answers `get_status`, and every surface that would touch the daemon
+returns a reason naming the module and the error instead:
 
 ```python
 >>> Robot("reachy_mini", mode="real").connect_eagerly()
 "cannot import strands_robots.device_connect.reachy_transport: No module named
-'device_connect_edge' - the Reachy transport helpers ship behind an extra:
-pip install 'strands-robots[device-connect]'"
+'strands_robots.device_connect.reachy_transport'"
 ```
 
+The reason stops at what it can establish. It prescribes no `pip install`, because no
+install supplies a module that ships in the core distribution, and a remedy that
+cannot help is worse than none - the same rule
+[`require_optional`](../../strands_robots/utils.py) applies when it is told a module
+arrives from a system package rather than an index.
+
 The same reason arrives as `connect_error` in `get_status`, so a mesh peer for a Mini
-on an install without the extra is still constructible and still reports why it is
-not connected.
+whose transport will not load is still constructible and still reports why it is not
+connected.
 
 `hardware.driver` is optional and validated when the registry loads: a value that is not a
 driver name is refused there, naming the robot, rather than being read as "no preference".
--- a/docs/getting-started/robot-factory.md
+++ b/docs/getting-started/robot-factory.md
@@ -147,7 +147,7 @@ returns a reason naming the module and the error instead:
 The reason stops at what it can establish. It prescribes no `pip install`, because no
 install supplies a module that ships in the core distribution, and a remedy that
 cannot help is worse than none - the same rule
-[`require_optional`](../../strands_robots/utils.py) applies when it is told a module
+[`require_optional`](https://github.com/strands-labs/robots/blob/main/strands_robots/utils.py) applies when it is told a module
 arrives from a system package rather than an index.
 
 The same reason arrives as `connect_error` in `get_status`, so a mesh peer for a Mini
--- a/strands_robots/drivers/reachy.py
+++ b/strands_robots/drivers/reachy.py
@@ -84,40 +84,48 @@ _PATH_STATUS = "/api/daemon/status"
 _PATH_STOP = "/api/move/stop"
 
 #: The module every daemon touch here goes through. It is a leaf that imports
-#: nothing but the standard library, but it lives inside
-#: :mod:`strands_robots.device_connect`, whose package ``__init__`` imports
-#: ``device_connect_edge`` and the three Device Connect drivers unconditionally.
-#: That package ships only in the ``[device-connect]`` extra, so on a stock
-#: ``pip install strands-robots`` importing the leaf raises
-#: ``ModuleNotFoundError: No module named 'device_connect_edge'`` as a side
-#: effect of the parent, naming a package the caller never asked for.
+#: nothing but the standard library, and its parent package
+#: :mod:`strands_robots.device_connect` resolves ``device_connect_edge`` and the
+#: three Device Connect drivers lazily, so importing the leaf executes no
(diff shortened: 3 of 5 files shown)


yinsong1986 (dismissed):
## Summary

Removes the `[device-connect]` extra remedy from `_resolve_transport`'s refusal string in `strands_robots/drivers/reachy.py`, leaving the reason as `cannot import <module>: <exc>` - the exact shape of the declared sibling `g1._resolve_message_class`. The premise checks out against the tree: `strands_robots.device_connect.__init__` resolves `device_connect_edge` lazily (since #2774), `reachy_transport.py` imports only stdlib names at module scope, and the wheel packages the whole `strands_robots` tree, so no `pip install` can supply the module whose absence reaches that `except ImportError`. The one site where the extra is still a real cause (`robot.py:845`, the Device Connect bring-up path) is untouched and now pinned both ways. The stale quote in `docs/getting-started/robot-factory.md` is corrected and graded against the resolver itself so it cannot drift again.

## What's good

- The scope control is explicit: `TestTheSiteWhereTheExtraIsStillTheCauseKeepsIt` pins that this is not a blanket sweep of the extra's name, and derives (rather than asserts) that the bring-up path really does import `device_connect_edge` eagerly.
- The stdlib-only premise is derived from the leaf's AST, so a future third-party import in the transport reopens the question loudly instead of silently invalidating the reason.
- The parity cell asserts string equality with the g1 sibling, so a re-added remedy fails in any wording - stronger than pinning a literal.
- Changelog fragment present (`changelog.d/2798-...md`), closing keyword in the body, all changed files ASCII-clean, ruff clean, and the full reachy driver suite plus the docs reference suite pass on the head (128 + 7 passed locally).

yinsong1986 (approved):
## Summary

Re-review after the docs-link push (`758d3f95`, which only repoints the `require_optional` reference in `docs/getting-started/robot-factory.md` at a GitHub URL so strict mkdocs resolves it). The substance is unchanged from the dismissed review's head: `_resolve_transport`'s refusal drops the `[device-connect]` remedy and reports `cannot import <module>: <exc>`, byte-identical to the declared sibling `g1._resolve_message_class`. The premise holds on this head - `strands_robots.device_connect.__init__` is fully PEP 562-lazy, `reachy_transport.py` imports only stdlib names at module scope, and the stdlib-only fact is derived from the leaf's AST rather than asserted, so a future third-party import reopens the question loudly. The one site where the extra is still a real cause (`robot.py`, the bring-up path) is untouched and pinned in both directions, the stale docs quote is corrected and graded against the resolver itself, and the parity cell asserts string equality with the sibling so a re-added remedy fails in any wording. Verified on the head: `tests/drivers/` plus `tests/test_docs_robot_factory_reference.py` pass (405 passed, 30 skipped), ruff check and format clean on the changed files, all changed surfaces ASCII-clean, changelog fragment present with the closing keyword in the PR body.

PR #2799, opened 2026-08-26
fix(docs): resolve a Markdown link that climbs out of the repository

cagataycali:
## The defect

`strands_robots/policies/moveit2/server/README.md:86` pointed at the MoveIt2 integration test with
a five-segment prefix:

```markdown
[`tests_integ/policies/moveit2/test_moveit2_live.py`](../../../../../tests_integ/policies/moveit2/test_moveit2_live.py)
```

The page sits four directories deep, so four segments reach the repository root and the fifth leaves
it. Measured from the page:

| | value |
|---|---|
| resolves to | `<one level above the checkout>/tests_integ/policies/moveit2/test_moveit2_live.py` |
| exists | `False` |
| correct relative path | `../../../../tests_integ/policies/moveit2/test_moveit2_live.py` |
| that target exists | `True` |

It renders as an ordinary link everywhere - GitHub, an editor, a rendered preview - so the only way
to find out is to click it and get a 404. The other link on the same page (`../policy.py`, one
segment) is correct, so this is a single off-by-one rather than a page written against the wrong root.

## Why nothing reported it

`mkdocs build --strict` is the only link checker in the repository, and it is narrower than it looks
in two independent ways.

**It reads only `docs/`.** MkDocs resolves links for files under `docs_dir`. A Markdown file
anywhere else is never parsed, so no checker looks at it at all. That is most of the tree:

| area | Markdown files | read by mkdocs |
|---|---|---|
| `changelog.d/` | 818 | no |
| `docs/` | 57 | yes |
| `examples/` | 12 | no |
| repository root (`README.md`, `AGENTS.md`, ...) | 4 | no |
| `strands_robots/` | 4 | no |

`mkdocs build --strict` passes on `main` both before and after this change, which is the measurement
that says the offender was outside its reach rather than tolerated by it.

**And the job that runs it cannot block a merge.** The `default` branch ruleset lists exactly one
required status check:

```
$ gh api repos/strands-labs/robots/rulesets/13012156 --jq '.rules[]|select(.type=="required_status_checks")|.parameters.required_status_checks'
[{"context": "call-test-lint / Test and Lint", "integration_id": 15368}]
```

`docs.yml`'s `build` job is not in it. So even a broken link *under* `docs/` is reported by a check a
merge does not wait for - and once merged, the same job runs on `main`, where its failure holds back
the `deploy` job that publishes the site.

## The change

One path segment in the README, plus `tests/test_markdown_links_resolve.py`, which puts the question
inside the suite the required check runs and grades the whole tree rather than the `docs/` subset:
**896 Markdown files, 568 link targets, 404 of them relative**.

The rule is one sentence: a relative target must resolve to a path that **exists** and that lies
**inside the repository**. Both halves are load-bearing. A target that climbs out may land on
something real beside one clone and on nothing beside another, so existence alone is a verdict about
the machine rather than about the link; the guard reports `escapes the repository` and
`does not exist` separately for that reason.

Two skips are needed to keep the sweep honest, and both are exercised by real pages rather than
asserted:

| what is skipped | offenders reported without the skip |
|---|---|
| fenced blocks | 1 - `docs/policies/lerobot-local.md`, which quotes `none` as a value |
| inline code spans | 2 - `changelog.d/2351-...md` and `docs/training/overview.md`, which quote the message ``Contact the maintainer on [Discord](...)`` |
| both | 3 |

Fragment resolution stays with MkDocs, which owns the slugifier that decides whether a heading
anchor exists; re-deriving those rules here would be a second implementation of somebody else's
contract. Site-absolute targets (`/path`) are left alone too - MkDocs reads them as site-root-relative
and GitHub as repository-root-relative, so they need a policy rather than a resolution, and the tree
ships none.

## Pre-fix split

With the README reverted and everything else as shipped: **1 failed, 20 passed**. The one failure is
the sweep, and it names the file, the target and the reason:

```
AssertionError: Markdown links that name nothing reachable:
    strands_robots/policies/moveit2/server/README.md -> ../../../../../tests_integ/policies/moveit2/test_moveit2_live.py (escapes the repository)
```

Zero of the four premise cells and zero of the thirteen exemplar cells fail pre-fix.

## Mutations

`mine` is the new suite; `sib` is 18 pre-existing suites that read `docs/` or walk the tree
(`tests/test_docs_*.py`, the changelog fragment guards, the test-name guard). `coll` is 21 on every
row, so nothing is hidden by a deselection.

| mutation | mine | coll | sib | firing |
|---|---|---|---|---|
| control | 0 | 21 | 0 | - |
| README reverted | 1 | 21 | 0 | the sweep |
| containment half dropped (existence only) | 2 | 21 | 0 | the escaping exemplar, and the two-reasons cell |
| containment dropped and README reverted | 3 | 21 | 0 | the union of those |
| existence half dropped (containment only) | 6 | 21 | 0 | every missing-target exemplar |
| inline code spans no longer blanked | 2 | 21 | 0 | the sweep, plus its exemplar |
| fenced-block content no longer blanked | 2 | 21 | 0 | the sweep, plus its exemplar |
| reference definitions no longer read | 1 | 21 | 0 | `[label]: target` |
| percent decoding dropped | 1 | 21 | 0 | the `%20` exemplar |
| fragment not split off the path | 2 | 21 | 0 | the sweep, plus the fragment exemplar |
| discovery narrowed to `docs/` | 2 | 21 | 0 | both scope premises |
| discovery narrowed to `docs/` **and** README reverted | 2 | 21 | 0 | the same two - the sweep goes silent |
| discovery returns nothing | 3 | 21 | 0 | all three premises |
| link pattern reads no target | 7 | 21 | 0 | every reported-side exemplar |

Two rows are worth reading together. Narrowing the discovery to `docs/` - the subset MkDocs already
covers - makes the sweep **silent** about the offender; only the two scope premises fire. Those
premises are the reason a future narrowing cannot quietly turn this guard back into a second copy of
`mkdocs build --strict`.

The `sib` column is zero on every row: none of the 18 suites that already read documentation can see
any of this, which is why the guard is new rather than an extension.

### The area list is derived, and that matters

Discovery walks the top-level directories rather than naming them, so a new one is graded on arrival.
Demonstrated by planting a Markdown file with a broken link in a directory that does not exist today:

| | fires | coll |
|---|---|---|
| areas derived, plant present | 1 (names the planted file) | 21 |
| areas hardcoded to today's list, same plant | **0** | 21 |
| areas hardcoded, no plant | 0 | 21 |

A hardcoded list equal to today's derived set fires nothing and still reads as a working test, so the
plant is the only way to tell the two apart.

### Non-vacuity

After this change the tree is clean, so the sweep cannot exercise its own failing branch. The
thirteen exemplars drive the same predicate over a tree built for the purpose - including a file
placed *beside* that tree, which is the only arrangement in which an existence check and a
containment check disagree - and one cell asserts the exemplars reach both verdicts.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1611 files).
- `mypy` **24 == 24** against a worktree pinned at this branch's merge base; normalized message sets
  byte-identical in both directions, 0 attributable (`checked 1608` vs `1607`, the difference being
  the new file).
- 398-file paired selection - every suite that walks the tree, parses source, reads docstrings, reads
  `docs/` or reads `changelog.d` - with the new file excluded from both sides and every path verified
  present on both: identical **19499 collected** and `19 failed, 19410 passed, 70 skipped` on each
  tree, the same 19 ids, `comm` empty in both directions.
- The 19 are pre-existing and unrelated: 17 need `awsiot`/`awscrt` (declared in the `[mesh-iot]`
  extra, absent here) and 2 are a lerobot-from-source `encode()` signature drift in
  `tests/training/test_lerobot.py`.
- `mkdocs build --strict` passes on this branch, as it does on `main`.
- The suite was re-run in full after rebasing onto the five commits that landed during it; every
  mutation row and the pre-fix split are unchanged.

No visual artifact: this changes a documentation link and adds a test, none of the surfaces that call
for one.

### DIFF
--- a/changelog.d/2799-markdown-links-resolve-inside-the-repo.md
+++ b/changelog.d/2799-markdown-links-resolve-inside-the-repo.md
@@ -1,0 +1,28 @@
+### Fixed: a Markdown link that climbs out of the repository
+
+`strands_robots/policies/moveit2/server/README.md` pointed at the MoveIt2 integration test with a
+five-segment `../../../../../` prefix. The page sits four directories deep, so the target resolved
+one level above the checkout and named nothing: a 404 for every reader who clicked it, on GitHub and
+in an editor alike. The prefix is now four segments and resolves to the file it names.
+
+Nothing in the repository graded that. `mkdocs build --strict` is the only link checker here, and it
+is narrower than it looks in two ways. It resolves links only for files under `docs_dir`, so a
+Markdown file anywhere else - `README.md`, `AGENTS.md`, a `changelog.d` fragment, a package
+reference page such as this one - is read by no checker at all. And the `build` job that runs it is
+not among the branch ruleset's required status checks, so even a broken link *under* `docs/` cannot
+block a merge; it surfaces afterwards on `main`, where the same job's failure holds back the Pages
+deploy that depends on it.
+
+So `tests/test_markdown_links_resolve.py` puts the question inside the suite the required check
+runs, and grades every Markdown file in the tree rather than the `docs/` subset: 896 files, 568 link
+targets of which 404 are relative, one offender before this change and none after. Its rule is that a relative target
+must resolve to a path that exists *and* that lies inside the repository. The second half carries
+its own weight - a target that climbs out may happen to land on something real beside one clone and
+on nothing beside another, so existence alone is a verdict about the machine rather than about the
+link.
+
+Fragment resolution stays with MkDocs, which owns the slugifier that decides whether a heading
+anchor exists; re-deriving those rules here would be a second implementation of somebody else's
+contract. Site-absolute targets are left alone too: MkDocs reads them as site-root-relative and
+GitHub as repository-root-relative, so they need a policy rather than a resolution, and the tree
+ships none.
--- a/strands_robots/policies/moveit2/server/README.md
+++ b/strands_robots/policies/moveit2/server/README.md
@@ -83,6 +83,6 @@ hatch run test-integ tests_integ/policies/moveit2/ -m moveit2 -v
 ```
 
 The integ tests live at
-[`tests_integ/policies/moveit2/test_moveit2_live.py`](../../../../../tests_integ/policies/moveit2/test_moveit2_live.py).
+[`tests_integ/policies/moveit2/test_moveit2_live.py`](../../../../tests_integ/policies/moveit2/test_moveit2_live.py).
 They exercise the full client → sidecar round-trip and expect a
 `success=True` response for an in-workspace pose.
--- a/tests/test_markdown_links_resolve.py
+++ b/tests/test_markdown_links_resolve.py
@@ -1,0 +1,347 @@
+"""Repo hygiene: a relative Markdown link names something inside this repository.
+
+A Markdown link written with the wrong number of ``../`` segments still renders
+as a link. GitHub, the docs site and an editor all show it as clickable prose,
+and the only way to learn it is broken is to click it, so the failure surfaces
+to a reader rather than to the author.
+
+Nothing in this repository graded that. The one link checker here is
+``mkdocs build --strict``, run by the ``build`` job in ``.github/workflows/docs.yml``,
+and it is narrower than it looks in two ways:
+
+* It resolves links only for files under ``docs_dir`` (``docs/``). A Markdown
+  file anywhere else -- ``README.md``, ``AGENTS.md``, a ``changelog.d``
+  fragment, a package reference page such as
+  ``strands_robots/policies/moveit2/server/README.md`` -- is never read, so a
+  broken link there is reported by nothing at all. That is where this guard
+  found its first offender: a five-segment ``../../../../../`` prefix on a page
+  four directories deep, which resolves outside the checkout entirely.
+* That job is not among the branch ruleset's required status checks, so even a
+  broken link *under* ``docs/`` cannot block a merge -- and once merged, the
+  same job runs on ``main`` and its failure holds back the Pages deploy the
+  ``deploy`` job performs.
+
+So this guard lives in the test suite, which the required check runs, and it
+grades every Markdown file in the tree rather than the ``docs/`` subset.
+
+The rule is one sentence: a relative link target must resolve to a path that
+exists and that lies inside the repository. The second half matters as much as
+the first -- a target escaping the checkout may happen to exist beside it on one
+machine and not on another, so existence alone is a verdict that depends on what
+sits next to the clone.
+
+Deliberately out of scope, each for a reason:
+
+* ``#fragment`` resolution. Whether a heading anchor exists depends on the
+  slugifier the site renders with, and MkDocs already owns that question
+  (``validation.links.anchors``). Re-deriving its slug rules here would be a
+  second implementation of somebody else's contract.
+* Site-absolute targets (``/path``). MkDocs reads those as site-root-relative
+  and GitHub as repository-root-relative, so they need a policy rather than a
+  resolution; the tree ships none today.
+* ``http``/``mailto`` and other schemed targets, which name nothing in the tree.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+from urllib.parse import unquote
+
+import pytest
+
+_REPO_ROOT = Path(__file__).resolve().parent.parent
+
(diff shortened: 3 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Fixes a relative Markdown link in `strands_robots/policies/moveit2/server/README.md` that carried one `../` segment too many (five climbs from a page four directories deep), so it resolved outside the checkout entirely. The corrected four-segment target resolves to `tests_integ/policies/moveit2/test_moveit2_live.py`, which exists at head. The PR also adds `tests/test_markdown_links_resolve.py`, a repo-wide sweep that grades every relative Markdown link target for both existence and containment inside the repository - closing the gap where `mkdocs build --strict` only reads files under `docs/` and is not a required check. Verified locally at head: the new suite passes (21 tests), `ruff check` and `ruff format --check` are clean, and the sweep correctly excludes fenced blocks, inline code spans, schemed/anchor-only/site-absolute targets, with constructed-tree exemplars pinning both the accept and reject branches so the rule cannot silently weaken once the tree is clean.

## What's good

- The containment check (`escapes the repository`) is kept distinct from the existence check, with an exemplar proving they can disagree - which is exactly the failure shape the README link had.
- Discovery is derived rather than listed, with a floor assertion so a known area dropping out of the sweep fails loudly (matches the derive-the-population pattern AGENTS.md pins elsewhere).
- Every helper carries a complete `Args:`/`Returns:` docstring per the repo's docstring-completeness gates, and test names describe behaviour rather than provenance.

yinsong1986 (dismissed):
## Summary

Corrects a relative link in `strands_robots/policies/moveit2/server/README.md` that carried five `../` segments from a page four directories deep, so it resolved one level above the checkout; the four-segment replacement resolves to `tests_integ/policies/moveit2/test_moveit2_live.py`, which exists at head. The PR also adds `tests/test_markdown_links_resolve.py`, a repo-wide sweep asserting every relative Markdown link target both exists and stays inside the repository - closing the gap where `mkdocs build --strict` reads only `docs/` and is not a required check. Verified at head: the 21 new tests pass, the sweep reports zero offenders after the fix and catches the pre-fix five-segment link on `main`, `ruff check`/`format --check` are clean on the new file, and the repo's meta-gates (behaviour-named test cases, no host paths, ASCII surfaces) all pass with the new file in scope.

## What's good

- The containment verdict (`escapes the repository`) is kept distinct from existence, with a constructed-tree exemplar proving the two can disagree - exactly the failure shape the README link had, and the reason existence alone would be a verdict about the machine rather than the link.
- Constructed exemplars pin both the accept and reject branches, so the rule cannot silently weaken once the tree is clean; discovery is derived with a floor assertion rather than a hardcoded list, matching the derive-the-population pattern AGENTS.md pins elsewhere.
- Change is recorded as a `changelog.d/` fragment per step 3 of the PR workflow, and scope is tight: one link fix, one guard, one fragment.

yinsong1986 (approved):
## Summary

Fixes the one broken relative link in `strands_robots/policies/moveit2/server/README.md` - five `../` segments from a page four directories deep, so it resolved one level above the checkout - to the four-segment form, which resolves to `tests_integ/policies/moveit2/test_moveit2_live.py` (verified present at head). Alongside it, `tests/test_markdown_links_resolve.py` adds a repo-wide sweep asserting every relative Markdown link target both exists and stays inside the repository, closing the gap the description measures: `mkdocs build --strict` reads only `docs/` and its job is not in the required check set, so a link like this one was graded by nothing. Verified at head: all 21 new tests pass against the PR tree, the constructed-tree exemplars pass standalone, the fixed link target exists, and the new file is ASCII-clean. The change is test-plus-docs only - no production code paths, no public API, no wire or persisted formats - so there is no one-way-door surface to weigh.

## What's good

- The containment verdict (`escapes the repository`) is kept distinct from existence, with an exemplar placing a real file beside the tree - the only arrangement where the two checks disagree, and exactly the shape of the offender.
- Discovery derives the area list from the tree with a floor assertion (`_REQUIRED_AREAS`) plus two scope premises, so a future narrowing back to `docs/` fails loudly rather than silently re-becoming a copy of mkdocs' coverage.
- Change recorded as a `changelog.d/` fragment per PR workflow step 3; scope is tight (one link, one guard, one fragment); helper docstrings carry complete `Args:`/`Returns:` blocks and test names describe behaviour, matching the repo's meta-gates.

Issue #1900, opened 2026-08-03
bug(sim/isaac): USD-mangled numeric URDF joint names leak through the public API (tn__1_ for joint "1") - cross-backend joint vocabularies disagree


yinsong1986:
## Summary

Split from #1895 (second finding, per that issue's own triage suggestion). The Isaac URDF importer mangles joint names that are not legal USD prim names, and the mangled names leak through the backend's public API - so the same URDF yields different joint vocabularies on Isaac vs MuJoCo, breaking any cross-backend consumer.

## Repro

The `robotstudio_so101` URDF names its joints literally `"1"`..`"6"`. USD prim names cannot start with a digit, so Isaac's URDF importer encodes them as `tn__1_`..`tn__6_`, and that leaks through:

- `robot_joint_names`
- `get_observation` keys
- `_RobotState.joint_names`

Isaac is *self-consistent* (obs keys match joint_names), but cross-backend code breaks: the same URDF on MuJoCo keeps `"1"`..`"6"`, and `examples/so101_curobo` feeds `scene.joint_names` into a cuRobo planner built from the same URDF (which sees the original names) - so even with #1895's reset fix merged, the Isaac path mismatches planner-vs-sim joint vocabularies.

## Suggested direction

Keep a `usd_name -> urdf_name` map in `_RobotState` at import time and translate at the public API boundary (`robot_joint_names`, observation keys, `send_action` resolution), so backends agree on the URDF's names. The mangling is deterministic (Isaac's `tn__` TfMakeValidIdentifier encoding), so the map can be built from the importer's own output against the parsed URDF joint list.

## Acceptance criteria

- The same URDF added on Isaac and MuJoCo reports identical `robot_joint_names`.
- `get_observation` keys and `send_action` name resolution use the URDF names on Isaac.
- Regression test with a numeric-joint-name URDF (no kit install required for the mapping logic; GPU-gated integ for the importer round-trip).

## Related

- #1895 (first finding, articulation revive - fixed separately; this defect is invisible until that one is fixed)
- #1798 / #1802 (Isaac backend parity program)

PR #2800, opened 2026-08-26
fix(isaac): keep only joint types URDF declares in the movable set

cagataycali:
## The two URDF readers in this package disagreed about one file

`strands_robots.simulation.isaac` reads a joint's `type` attribute out of a URDF
in two places.

`joint_names.urdf_joint_names` keeps the joints whose type produces a named
articulation DOF, and hands that list to `demangle_usd_joint_names` as the pool
of URDF names a mangled DOF name may decode to. `loaders.load_urdf` maps the
type onto a `JointDef` kind and **refuses** a type it does not recognise,
naming the set it accepts in the refusal - which makes it this package's own
record of what URDF declares.

The movable set carried `spherical`. URDF declares six joint types and that is
not one of them, so the two readers, handed the same one-joint file, disagreed:

| `type` | `urdf_joint_names` | `load_urdf` |
|---|---|---|
| `revolute` | `['j1']` | `j1 revolute` |
| `continuous` | `['j1']` | `j1 revolute` |
| `prismatic` | `['j1']` | `j1 prismatic` |
| `fixed` | `[]` | `j1 fixed` |
| `floating` | `[]` | `j1 fixed` |
| `planar` | `[]` | `j1 fixed` |
| `spherical` | **`['j1']`** | **`ValueError: unknown joint type (expected one of ['continuous', 'fixed', 'floating', 'planar', 'prismatic', 'revolute'])`** |

One reader called the joint movable; the other called the file unloadable, and
enumerated the accepted set in the message - a set that does not contain
`spherical`. `load_urdf`'s six keys are exactly URDF's six declared types.

## What a name from nowhere does to the decode

`demangle_usd_joint_names` draws translations from the pool
`urdf_joint_names` builds. A candidate in that pool that no URDF can declare
has two ways to do harm, and both are reachable from a file:

```
A) a spherical joint named "1"
   pool ['1']    DOF tn__1_ -> public ['1']       usd_to_urdf {'tn__1_': '1'}
   pool []       DOF tn__1_ -> public ['tn__1_']  usd_to_urdf {}

B) a revolute joint "a-b" beside a spherical joint "a.b"
   (both substitute to "a_b" under the legacy mangle)
   pool ['a-b','a.b']  DOF a_b -> public ['a_b']  usd_to_urdf {}
   pool ['a-b']        DOF a_b -> public ['a-b']  usd_to_urdf {'a_b': 'a-b'}
```

In A the public vocabulary reports the name of a joint the format cannot
express. B is the worse one: the spurious candidate makes the decode ambiguous,
an ambiguous decode keeps the USD name, and so the joint the URDF **really**
declares surfaces as `a_b` on `robot_joint_names`, on `get_observation` keys and
in `send_action` resolution. That is the #1900 leak this module exists to close,
reintroduced by a candidate URDF cannot declare.

## Reachability

Latent. No valid URDF carries the type, so no valid URDF reaches either harm:
none of the 68 URDFs the registry downloads declares one, and `load_urdf`
refuses such a file outright. What shipped wrong is the vocabulary and the
documented accepted set - `urdf_joint_names` is exported, and its docstring told
a caller that `spherical` produces a DOF.

## The change

`_MOVABLE_URDF_JOINT_TYPES` drops `spherical`; the docstring's accepted set and
the constant's comment are corrected with it. No behaviour changes for any type
URDF declares - the 6-row table above is identical before and after.

The set is now pinned **between** the loader's two answers:

```
{types the loader reads as moving}  <=  movable  <=  {types the loader recognises}
```

The upper bound refuses a type URDF does not declare. The lower bound refuses
dropping one that does move. The bounds are deliberately not collapsed into an
equality: were the importer found to surface a named DOF for a type the loader
reads as `fixed`, that belongs in the movable set without the loader's own
mapping changing.

The comment's reason for excluding `floating` and `planar` also said the
registry URDFs do not use them. Five of the 68 declare a `floating` base (a1,
aliengo, go1, laikago, b2), so that clause is corrected: the exclusion rests on
the importer not mapping a multi-DOF joint onto single named DOFs, which is why
the loader reads both as `fixed` too.

## Why nothing caught it

`test_urdf_joint_name_demangle.py` builds every fixture from `revolute`,
`continuous`, `prismatic` and `fixed` - the four types an author writing a URDF
reaches for. The vocabulary's fourth member and both of URDF's excluded real
types were never exercised, and no test read the constant at all. Every
mutation below leaves that suite and the three other URDF-reading suites green
except one, noted in the table.

## Tests

New `tests/simulation/isaac/test_urdf_joint_type_vocabulary.py`, 28 cases, no
Isaac Kit and no `pxr` (both readers are stdlib XML parsers). URDF's six
declared types are stated locally in the test rather than read from the code, so
the file is an independent oracle for the vocabulary.

Pre-fix split: **6 failed / 21 passed**, with 0 of the 2 premise cells, 0 of the
3 unchanged-behaviour cells and 0 of the 6 declared-type agreement cells firing.
The two headline failures read `assert ['1'] == ['tn__1_']` and
`assert ['a_b'] == ['a-b']`.

| mutation | fires | collected | pre-existing suites |
|---|---|---|---|
| control (fix intact) | 0 | 28 | 0 |
| full revert | 6 | 28 | 0 |
| constant reverted, docstring left correct | 7 | 28 | 0 |
| docstring reverted, constant left correct | 1 | 28 | 0 |
| a fourth invented type added (`screw`) | 5 | 28 | 0 |
| a real single-DOF type dropped (`prismatic`) | 5 | 28 | 5 |
| alternative fix: loader taught to accept `spherical` | 7 | 28 | 0 |
| set derived from the loader map in production | 0 | 28 | 0 |
| the test's own URDF oracle corrupted | 8 | 29 | 0 |

9 mutations, 8 distinct firing sets, source restored byte-identically.

Three rows carry the argument. **The alternative fix** - closing the
disagreement by teaching `load_urdf` about `spherical` instead - fires 7,
including the premise cell that the type map's keys are exactly URDF's declared
types: the direction of the fix is graded, not asserted. **Dropping a real
single-DOF type** fires the lower bound and trips 5 pre-existing tests, so the
sandwich is load-bearing in both directions. **Deriving the set from the loader
map in production** fires 0 - which is the honest reason it is not done here: a
derivation makes the guard vacuous and collapses the sandwich to an equality.

`collected` is recorded per row because a narrowed universe runs fewer tests
while still reporting success; the last row moves it 28 to 29.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1611 files).
- mypy **24 == 24** against a pristine worktree of the merge-base, normalized message sets byte-identical in both directions, 0 attributable to either changed file.
- Paired run over an identical 360-file selection (all of `tests/simulation/isaac/`, every guard that walks the tree or reads source, docstrings or `changelog.d`, and everything naming the two readers), the new file excluded from both trees and every path verified present on both: identical **16132 collected** and `4 failed, 15974 passed, 154 skipped` on each, 4 identical failing ids, `comm` empty in both directions. Those 4 pass **13/13 in isolation on both trees** - a module-caching artifact of the large selection, not a regression.
- No visual artifact: this changes a vocabulary and a refusal boundary, not a policy, a simulation, a render or a recording. The measured 6-row agreement table and the two decode readings are the evidence.

### DIFF
--- a/changelog.d/2800-urdf-movable-joint-types.md
+++ b/changelog.d/2800-urdf-movable-joint-types.md
@@ -1,0 +1,55 @@
+### Fixed: the Isaac backend's two URDF readers agree on what a joint type is
+
+Two functions in `strands_robots.simulation.isaac` read a joint's `type` out of
+a URDF. `joint_names.urdf_joint_names` keeps the ones that produce a named
+articulation DOF and hands that list to `demangle_usd_joint_names` as the pool
+of URDF names a mangled DOF name may decode to. `loaders.load_urdf` maps the
+type onto a `JointDef` kind and *refuses* a type it does not recognise, naming
+the set it accepts in the refusal - which makes it this package's record of
+what URDF declares.
+
+The movable set carried `spherical`. URDF declares six joint types -
+`revolute`, `continuous`, `prismatic`, `fixed`, `floating`, `planar` - and that
+is not one of them. So the two readers, handed the same file, disagreed:
+
+```
+type         urdf_joint_names        load_urdf
+revolute     ['j1']                  j1 revolute
+continuous   ['j1']                  j1 revolute
+prismatic    ['j1']                  j1 prismatic
+fixed        []                      j1 fixed
+floating     []                      j1 fixed
+planar       []                      j1 fixed
+spherical    ['j1']                  ValueError: unknown joint type
+                                     (expected one of ['continuous', 'fixed',
+                                     'floating', 'planar', 'prismatic',
+                                     'revolute'])
+```
+
+A name in the decode pool that no URDF can declare has two ways to do harm,
+both reachable from a file. A `spherical` joint named `1` gave the DOF name
+`tn__1_` a candidate to decode to, so the public vocabulary reported `1` - the
+name of a joint the format cannot express. And `a-b` and `a.b` both substitute
+to `a_b` under the legacy mangle, so a `revolute` joint named `a-b` beside a
+`spherical` one named `a.b` made that decode ambiguous; an ambiguous decode
+keeps the USD name, so the joint the URDF really declares surfaced as `a_b` on
+`robot_joint_names`, on `get_observation` keys and in `send_action` resolution.
+That is the #1900 leak the module exists to close, reintroduced by a candidate
+URDF cannot declare.
+
+No valid URDF reaches either, because no valid URDF carries the type: none of
+the 68 URDFs the registry downloads declares one, and `load_urdf` refuses such
+a file outright. The fix is the vocabulary and the documented accepted set,
+which said `spherical` produces a DOF.
+
+The set is now pinned between the loader's two answers - every member must be a
+type the loader recognises, and no type the loader reads as moving may be left
+out. The bounds are deliberately not collapsed into an equality: were the
+importer found to surface a named DOF for a type the loader reads as `fixed`,
+that belongs in the movable set without the loader's own mapping changing.
+
+The comment's reason for excluding `floating` and `planar` also said the
+registry URDFs do not use them. Five of the 68 declare a `floating` base - a1,
+aliengo, go1, laikago and b2 - so the exclusion rests on the importer not
+mapping a multi-DOF joint onto single named DOFs, which is why the loader reads
+both as `fixed` too, and not on their absence.
--- a/strands_robots/simulation/isaac/joint_names.py
+++ b/strands_robots/simulation/isaac/joint_names.py
@@ -51,10 +51,20 @@ _BASIC_CHAR_RE = re.compile(r"[A-Za-z0-9_]")
 # (Isaac Sim 6.0.x URDF importer output, e.g. ``tn__1_`` for joint ``1``).
 _TRANSCODING_PREFIX = "tn__"
 
-# URDF joint types that produce articulation DOFs on import. ``fixed`` never
-# does; ``floating``/``planar`` are multi-DOF types the Isaac importer does
-# not map onto single named DOFs, and the registry URDFs do not use them.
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Removes `spherical` from `joint_names._MOVABLE_URDF_JOINT_TYPES`, aligning `urdf_joint_names` with `loaders.load_urdf`'s record of the six joint types URDF actually declares (`revolute`, `continuous`, `prismatic`, `fixed`, `floating`, `planar`). Before the fix, a `spherical` joint entered the demangle candidate pool that `demangle_usd_joint_names` draws from, where a name no URDF can declare could either be reported as a movable joint or make a real joint's decode ambiguous (reintroducing the #1900 mangled-name leak). The behaviour change is narrowing only on inputs `load_urdf` already refuses as unknown joint types, so no valid URDF observes a difference; the five registry URDFs with `floating` bases are unaffected, since the loader reads those as `fixed` too. The new test module pins the vocabulary between the loader's two answers (subset bounds rather than a brittle equality), covers all six declared types plus three undeclared ones, and demonstrates both harm paths from a file - all stdlib-XML, no Kit dependency. Docstring, module comment, and changelog fragment are updated consistently.

## What's good

- Verified at head `b93ea769`: the new vocabulary tests and the existing `test_urdf_joint_name_demangle.py` all pass (51 tests), so the narrowing regresses nothing in the demangle path.
- `git grep` confirms no other consumer of `spherical` in the Isaac backend; `simulation.py`'s `add_robot` path is the only caller of `urdf_joint_names` and needs no change.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention and passes the fragment format tests; all three files are ASCII-clean with no host paths.
- The docstring-sync test (`test_the_public_docstring_names_the_set_it_returns`) derives its expectation from the constant, so the prose cannot silently drift from the behaviour again.

PR #2801, opened 2026-08-26
test(deps): grade a system_install= remedy by its value, not the keyword's presence

cagataycali:
## Problem

`tests/test_dependency_audit.py::test_require_optional_offers_no_pip_remedy_for_a_system_provided_module`
exists because `rclpy` and `rosidl_runtime_py` are not on PyPI. Every `pip install`
line offered for them is an instruction the caller can follow to no effect:
`pip install 'strands-robots[ros2]'` exits 0 having installed only the cyclonedds
RMW binding and leaving the module exactly as missing, and `pip install rclpy` does
not resolve at all. `require_optional(system_install=...)` replaces that block with
the step that does supply the module, and the sweep required every such call site
to pass it.

It required the **keyword**, not the **string**:

```python
elif "system_install" not in keywords:
    offenders.append(...)
```

while the failure message it raises prescribes a specific value
(`Pass system_install=ROS2_SYSTEM_INSTALL_HINT instead`). Those are different
claims. Measured against `require_optional` itself, with the requested module made
unimportable:

| `system_install=` | message the caller reads | old guard |
|---|---|---|
| *omitted* | `Install with:` / `pip install <mod>` | flagged (correct) |
| `None` | `Install with:` / `pip install <mod>` -- **byte-identical to omitting it** | **passes** |
| `""` | `'<mod>' is required for ...` and nothing else | **passes** |
| `"   "` | the same, plus a blank line | **passes** |
| `0` | **`TypeError` from `str.join`**, not the documented `ImportError` | **passes** |
| `False` | the same `TypeError` | **passes** |
| a real hint | the hint | passes (correct) |

`None` is the parameter's own default and inside its declared `str | None`
annotation, so it is the natural spelling for a call site forwarding a
per-platform hint that turned out to be absent. It produces exactly the message
this sweep was written to forbid. The non-string rows are worse than wrong text:
`require_optional` documents `ImportError`, and a caller writing
`except ImportError` around an optional import misses the refusal entirely.

## Why nothing caught it

Both production sites pass a module-level constant, so the tree holds no offender
and this is a hole in the guard rather than a live defect. It is not covered
elsewhere either, and the boundary is exact.

`test_the_rclpy_refusals_name_the_step_that_supplies_it` does assert on the real
messages -- but it names **two `rclpy` call sites explicitly**
(`RosTelemetryBridge()` and `Robot._check_ros2_bridge_deps`). `rosidl_runtime_py`
is the other member of `_SYSTEM_PROVIDED_MODULES` and has no behavioural refusal
test at all. So planting an offending site for each member gives opposite answers:

| planted site | old guard | new guard |
|---|---|---|
| `require_optional("rclpy", system_install=None)` | 1 failure, from the behavioural sibling | 2 failures |
| `require_optional("rosidl_runtime_py", system_install=None)` | **0 -- silent, `39 passed, 1 skipped`** | 1, naming the site |
| `require_optional("rosidl_runtime_py", system_install="")` | **0 -- silent** | 1, naming the site |
| `require_optional("rosidl_runtime_py", system_install=ROS2_SYSTEM_INSTALL_HINT)` | 0 | 0 (no over-reach) |

## Change

The sweep already walked every call site; it discarded the keyword values. It now
carries them, and one predicate `_unusable_remedy_reason` decides whether a remedy
is usable, shared by the sweep and by the exemplars so the two cannot drift. Each
reason names the message that spelling really renders rather than only reporting
that the value was rejected.

Its boundary is deliberate: **only a literal is judged**. A name, an attribute or
an f-string is the shipped form, whose text a syntax tree cannot know, so it is
accepted here and left to the runtime tests in `tests/test_utils.py`.

Because the tree holds no offender, the corpus cannot exercise the new branch. The
exemplars grade the predicate directly instead -- 4 accepted forms, 5 refused ones,
and a non-vacuity check that the rule reaches both verdicts -- with the consequence
of each refused literal measured against `require_optional` beside them, using an
absent module name so the probes are hermetic (nothing patched, no import cache
disturbed).

`require_optional` itself is **unchanged**, and that is a measurement rather than
an omission: see the last mutation row.

## Mutations

Every row run against the whole file, with `collected` recorded so a silent
deselection cannot hide one. Control clean, source restored byte-identically
(md5-verified).

| mutation | fires | note |
|---|---|---|
| control (no mutation) | 0 | `53 passed, 1 skipped` |
| `None` branch dropped | 1 | the `None` exemplar |
| blank branch dropped | 2 | both blank exemplars |
| non-string branch dropped | 3 | both non-string exemplars + non-vacuity |
| predicate always accepts | 6 | every refused exemplar + non-vacuity |
| predicate always refuses | 4 | all 3 accepted-form controls + the corpus sweep |
| consumer reverted to presence-only, with the `None` plant | 1 | only the behavioural sibling; the sweep goes silent again |
| **runtime read for truth instead of `is not None`** | 2 | the tempting production change: a blank **and** a non-string remedy fall back to the pip block |

The last row is why production is untouched. `if system_install:` would send both
of those spellings to the dead-end pip line; `is not None` is the safer of the two
branches, and the spellings that should never be written are now refused where
they are written rather than absorbed where they are read.

One earlier version of the `None` exemplar expected the reason to contain
`"None"`, and the mutation that drops the `None` branch fired **0** against it --
the non-string branch reports `NoneType`, which satisfies that substring. The
expectation is now the phrase unique to that branch, and the row fires 1. The
tuple's comment records why.

## Gate

* `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ`.
* `mypy` **24 == 24** against a pristine worktree of the same base, normalized
  message sets byte-identical in both directions, 0 attributable to the changed file.
* Changed file: **53 passed, 1 skipped** (up from 40 collected).
* Paired run over an identical 524-file selection -- every test that walks the tree,
  parses source, reads docstrings, `pyproject.toml` or `changelog.d`, or names
  `require_optional`/`system_install` -- with the changed file excluded from both
  trees and every path verified present on both: identical **24731 collected** and
  **20 failed, 24606 passed, 105 skipped** on each, 20 identical failing ids, `comm`
  empty in both directions, distinct rootdirs. The 20 are pre-existing and
  classified: 17 need `awsiot` (`find_spec` False; declared in the `[mesh-iot]`
  extra) and 3 are lerobot-from-source `encode()` drift.

No visual artifact: this changes a static hygiene rule and no policy, simulation,
rendering, recording or asset behaviour. The measured message table is the evidence.

### DIFF
--- a/changelog.d/2801-system-install-remedy-graded-by-value.md
+++ b/changelog.d/2801-system-install-remedy-graded-by-value.md
@@ -1,0 +1,54 @@
+### Tests: a `system_install=` remedy is graded by the string it carries, not the keyword's presence
+
+`tests/test_dependency_audit.py::test_require_optional_offers_no_pip_remedy_for_a_system_provided_module`
+exists because `rclpy` and `rosidl_runtime_py` are not published on PyPI, so every
+`pip install` line offered for them is an instruction a caller can follow to no
+effect: `pip install 'strands-robots[ros2]'` exits 0 having installed only the
+cyclonedds RMW binding and leaving the module exactly as missing, and
+`pip install rclpy` does not resolve at all. `require_optional(system_install=...)`
+replaces that block with the step that does supply the module, and the sweep
+required every such call site to pass it.
+
+It required the keyword and not the string. Those are different claims, and the
+gap is not narrow - measured against `require_optional` itself with the module
+made unimportable, three literals satisfy the keyword's presence and defeat what
+it is for:
+
+* `system_install=None` renders the pip block **byte-for-byte identical** to
+  omitting the keyword. It is the parameter's own default and inside its declared
+  `str | None` annotation, so it is the natural spelling for a caller forwarding a
+  per-platform hint that turned out to be absent - and it produces exactly the
+  message this sweep was written to forbid.
+* `system_install=""` (or any blank string) leaves the refusal naming the module
+  and carrying no remedy at all.
+* A non-string remedy replaces the documented failure: `str.join` raises
+  `TypeError`, so a caller writing `except ImportError` around an optional import
+  misses the refusal entirely.
+
+Both production sites pass a module-level constant, so nothing in the tree is an
+offender and this is a hole in the guard rather than a live defect. It is not
+covered elsewhere either. `test_the_rclpy_refusals_name_the_step_that_supplies_it`
+does assert on the real messages, which is why planting `system_install=None` at
+the `rclpy` site is caught - but it names two `rclpy` call sites explicitly, and
+`rosidl_runtime_py` is the other member of the set the sweep grades and has no
+behavioural test at all. Planting `require_optional("rosidl_runtime_py",
+system_install=None)` therefore fired nothing: the sweep reported the keyword as
+present and the suite passed.
+
+The rule now reads the value, through one predicate the sweep and the exemplars
+share, and each reason names the message that spelling really renders rather than
+only reporting that the value was rejected. Its boundary is deliberate: only a
+literal is judged. A name, an attribute or an f-string is the shipped form, whose
+text a syntax tree cannot know, so it is accepted here and left to the runtime
+tests in `tests/test_utils.py`. Because the tree holds no offender the corpus
+cannot exercise the new branch, so the exemplars grade the predicate directly -
+four accepted forms, five refused ones, and a non-vacuity check that the rule
+reaches both verdicts - with the consequence of each refused literal measured
+against `require_optional` beside them.
+
+`require_optional` is unchanged, and that is a measurement rather than an
+omission. Reading `system_install` for truth instead of for `None` would send a
+blank *and* a non-string remedy back to the pip block, which for these libraries
+is the dead-end instruction; the current `is not None` branch is the safer of the
+two, and the spellings that should never be written are now refused where they
+are written rather than absorbed where they are read.
--- a/tests/test_dependency_audit.py
+++ b/tests/test_dependency_audit.py
@@ -14,6 +14,7 @@ import importlib.util
 import re
 import sys
 from pathlib import Path
+from typing import Any
 
 _REPO_ROOT = Path(__file__).resolve().parent.parent
 _PYPROJECT = _REPO_ROOT / "pyproject.toml"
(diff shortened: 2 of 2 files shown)


yinsong1986 (dismissed):
## Summary

Test-only hardening of the dependency-audit sweep: `_require_optional_call_sites()` now carries the keyword *values* (not just names), and a new `_unusable_remedy_reason()` refuses the three literal `system_install=` values that satisfy the keyword's presence while defeating its purpose (`None` re-renders the forbidden pip block, a blank string leaves the refusal with no remedy, a non-string turns the documented `ImportError` into a `TypeError` from `str.join`). Each refused literal is backed by a runtime premise test measured against `require_optional` itself, using an absent module name so nothing is patched and the import cache stays clean, plus accept-side controls for the shipped constant-reference form and a non-vacuity check that the exemplar set reaches both verdicts.

## What's good

- Premise tests assert against the real `require_optional` rendering rather than restating it, so the AST rule and the runtime behaviour cannot silently diverge.
- The predicate deliberately declines to judge names/attributes/f-strings (statically undecidable), with accept-side exemplars pinning that scoping against over-reach.
- Every helper carries a complete `Args:` block per the AGENTS.md docstring conventions, and the refused-literal reasons name consequences rather than bare rejections.
- All 54 tests in `tests/test_dependency_audit.py` pass on the head SHA; ruff check and format are clean.

yinsong1986 (approved):
## Summary

Test-only change: the dependency-audit sweep that requires `system_install=` at every `require_optional` call site requesting a system-provided module (`rclpy`, `rosidl_runtime_py`) now grades the keyword's *value*, not just its presence. `_require_optional_call_sites()` carries the keyword value nodes, and a single shared predicate `_unusable_remedy_reason()` refuses the three literal spellings that satisfy a presence check while defeating the keyword's purpose: `None` (re-renders the forbidden pip block byte-for-byte), a blank string (refusal with no remedy), and a non-string (turns the documented `ImportError` into a `TypeError` from `str.join`). Since the second commit the PR also carries the required `changelog.d/` fragment, which resolves the reason the prior review was dismissed. Verified independently at the head SHA: the full changed file passes (54 tests), ruff check/format are clean, the file and fragment are ASCII-clean, and planting `require_optional("rosidl_runtime_py", system_install=None)` in the package makes the sweep fail naming the exact site and consequence -- the case that was silent under the presence-only guard. `require_optional` itself is untouched, and the PR's mutation table justifies that as the safer branch (`is not None` vs a truthiness read that would send blank/non-string remedies back to the dead-end pip block).

## What's good

- The predicate's boundary is deliberate and pinned: only literals are judged, with accept-side exemplars (constant reference, attribute, f-string) guarding against over-reach, and a non-vacuity check that the exemplar set reaches both verdicts.
- Each refused literal is backed by a premise test measured against the real `require_optional` rendering, using an absent module name so the probes are hermetic.
- Changelog fragment per AGENTS.md step 3; no edits to `CHANGELOG.md`.
- Scope discipline: 2 files, no production behavior change, and the description records the mutation evidence for why production stays untouched.

PR #2802, opened 2026-08-26
fix(drivers): stop the Dynamixel codec citing a vendor unstuffer that loses a byte

cagataycali:
`strands_robots.drivers.dynamixel.protocol` opened with "Every function here is
verifiable against Robotis' `dynamixel_sdk` byte-for-byte", and `_unstuff`
described itself as "the mirror of Robotis' `removeStuffing`". The first claim
has one exception and the second names it: the vendor's unstuffer is not the
inverse of the vendor's own stuffer, so a reader who follows that instruction
ports a byte loss in.

The shipped implementation is correct and is unchanged. This corrects the two
docstrings and grades the property that makes it correct.

## The vendor's escape pair is not the identity

Exhaustive over every payload that can carry a Protocol 2.0 reserved run -
lengths 4 through 7 over `{FF, FD, 00, 01}`, **21760 payloads** - with the SDK
installed as an oracle (`dynamixel-sdk` 4.0.5, pure Python, no hardware, no
serial port):

| measurement | result |
| --- | --- |
| `_stuff` vs `addStuffing`, byte-for-byte | **0 divergences** (the control) |
| `_unstuff(_stuff(x)) == x` | **0 lossy** |
| `removeStuffing(addStuffing(x)) == x` | **12 lossy** |
| `_unstuff` vs `removeStuffing` over the same stuffed bytes | **12 divergences** |

The vendor round trip is measured through the SDK's **own** stuffer, so none of
this project's code is in that loop. The two stuffers agreeing on all 21760 is
what localises the divergence to the unstuffer alone.

Shortest witness, `FF FF FD FF FD FD`:

```
_unstuff(_stuff(...))        -> fffffdfffdfd   (identity)
removeStuffing(addStuffing)  -> fffffdfffd     (one byte short)
```

## Why: it compacts in place and reads what it has overwritten

`addStuffing` writes into a separate `temp` buffer, so its look-back reads the
original packet. `removeStuffing` compacts in place:

```python
packet[index] = packet[i + PKT_INSTRUCTION]      # index <= i + PKT_INSTRUCTION
...
and (packet[i + PKT_INSTRUCTION - 1] == 0xFF)    # may already have been rewritten
and (packet[i + PKT_INSTRUCTION - 2] == 0xFF)
```

With two reserved runs in one payload the second run's escape sits at a position
whose two preceding bytes the loop has already moved, so the condition matches an
`FF FF` that is no longer there and a data byte is dropped as if it were an
escape. Asymmetric quality inside one vendor module, which is why five of six
comparisons here are byte-perfect and one is not.

One thing the mutation table below corrected in my own first draft: the defect is
the in-place compaction, **not** the choice of buffer. Reading the untouched
input is equivalent to reading the emitted output (row b3 fires nothing), because
a stuffed payload still carries the run intact at the position being examined.
Only a buffer the same loop is rewriting can answer with a byte that is no longer
there. The docstring says that, rather than the stronger claim I first wrote.

## Nothing graded it, and the docstring pointed at the broken side

Substituting the vendor algorithm for the shipped one - a port verified faithful
to the real `removeStuffing` on **21760 of 21760** payloads, and lossy on 12 -
leaves `tests/drivers/test_dynamixel_driver.py` at **68 passed**. So the edit the
old docstring invited lands green.

## What changed

- The module docstring's byte-for-byte claim now names its one exception.
- `_unstuff`'s docstring states the divergence, its cause, the witness, and the
  invariant to preserve instead of parity. It describes that invariant inline
  rather than citing a test filename, which
  `tests/test_docstrings_no_dangling_doc_refs.py` refuses - my first draft cited
  the file and the paired gate caught it.
- A new suite asserts the round trip over the whole corpus rather than a
  hand-picked list. An escape/unescape pair that is not the identity is broken
  whichever implementation is older, so the round trip is the adjudicator the SDK
  cannot serve as. Three cells consult the SDK to grade the divergence claim
  itself and skip where it is absent, since it is not a declared dependency of
  this project: **11 passed** with it installed, **8 passed / 3 skipped**
  without, so the identity cells run in CI either way.

## Pre-fix split

Both docstrings reverted, every test kept: **4 failed / 7 passed**. All four are
in `TestTheDocstringsNameTheDivergence`; `0 of 4` in `TestTheEscapePairIsTheIdentity`
and `0 of 3` in `TestTheVendorImplementationIsTheOneThatDiverges`, which is
correct - the behaviour was already right and those cells are the pin.

## Mutations

`new` is the added suite, `sib` is `tests/drivers/test_dynamixel_driver.py`.
`coll` is 11 on every row, so nothing is hidden by a deselection.

| mutation | new | sib | firing |
| --- | --- | --- | --- |
| control | 0 | 0 | |
| both docstrings reverted | 4 | 0 | the four docstring cells |
| `_unstuff` ports `removeStuffing` | 2 | 0 | corpus identity, witness identity |
| `_unstuff` look-back reads the input, not the emitted output | 0 | 0 | equivalent - see above |
| corpus lengths capped below six | 2 | 0 | corpus premise, vendor round trip |
| witness replaced by a single-run payload | 2 | 0 | both vendor cells |
| `_stuff` look-back reads the output | 0 | 0 | the claimed-equivalent form, measured |

`sib` is **0 on all seven rows**: the pre-existing suite cannot see any of this,
which is the reason the divergence survived. Rows b3 and b6 are measured zeros
with structural reasons rather than gaps - b6 grades a claim `_stuff`'s own
docstring already makes, and b3 is what made me weaken my account of the cause.
Both source files restored byte-identically (md5) after the run.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests`
  (1556 files already formatted).
- `mypy` **24 == 24** against the merge base, normalized message sets
  byte-identical in both directions, **0 attributable** to the changed files
  (`checked 1608` vs `1607`, the one new file).
- Paired 146-file selection (all of `tests/drivers/`, plus every suite that walks
  the tree, parses source or reads docstrings, `Args:`/`Raises:` blocks or
  cross-reference roles): identical **6078 collected** on both trees with the new
  file excluded from each, **6022 passed, 56 skipped, 0 failed**, and `comm`
  empty in **both** directions - zero failures on either side.
- `tests/drivers/` is **429 passed** with the SDK present and **426 passed,
  3 skipped** without.
- Changelog and docs gates: 89 passed; fragment resolves in the assembled log.

No visual artifact: this changes two docstrings and adds tests, none of the
surfaces that call for one. The measured divergence table and the mutation table
are the evidence.

### DIFF
--- a/changelog.d/2802-dynamixel-unstuff-vendor-divergence.md
+++ b/changelog.d/2802-dynamixel-unstuff-vendor-divergence.md
@@ -1,0 +1,24 @@
+### Fixed: the Dynamixel codec no longer cites a vendor unstuffer that loses a byte
+
+`strands_robots.drivers.dynamixel.protocol` opened with the claim that every
+function in it is verifiable against Robotis' `dynamixel_sdk` byte-for-byte, and
+`_unstuff` described itself as the mirror of the SDK's `removeStuffing`. Measured
+over every payload that can carry a Protocol 2.0 reserved run - 21760 cases,
+lengths four through seven over `{FF, FD, 00, 01}` - the two stuffers agree
+byte-for-byte, and the vendor's unstuffer is lossy on twelve of them. Its loop
+compacts the packet in place, so the two-byte look-back that decides whether an
+`FD` is an escape reads positions the loop has already overwritten; on a payload
+carrying two reserved runs it matches an `FF FF` that is no longer there and
+drops a data byte. The shortest witness is `FF FF FD FF FD FD`, which comes back
+from `removeStuffing` as `FF FF FD FF FD`.
+
+The shipped implementation is correct and is unchanged. What the docstrings did
+was point a future reader at the broken side as the thing to match: substituting
+the vendor algorithm for the shipped one leaves this module's own suite at 68
+passed, so the byte loss would have landed silently. Both docstrings now state
+the divergence and its cause, and the round-trip identity is graded over the
+whole reserved-run corpus rather than a hand-picked list - an escape and
+unescape pair that is not the identity is broken whichever implementation is
+older, which is the adjudicator the SDK cannot serve as. Two further cells
+consult the SDK to grade the divergence claim itself and skip where it is
+absent, since it is not a declared dependency.
--- a/strands_robots/drivers/dynamixel/protocol.py
+++ b/strands_robots/drivers/dynamixel/protocol.py
@@ -1,7 +1,10 @@
 """Dynamixel Protocol 2.0 wire format. Pure, no I/O.
 
 Every function here is verifiable against Robotis' ``dynamixel_sdk`` byte-for
--byte. The point of extracting the codec from the SDK is not to avoid the
+-byte, with one deliberate exception documented on :func:`_unstuff`: the SDK's
+``removeStuffing`` is not the inverse of its own ``addStuffing``, so grading
+against it would reintroduce a byte loss. The point of extracting the codec
+from the SDK is not to avoid the
 dependency - ``dynamixel_sdk`` is on PyPI and small - but to make the wire
 format part of the driver's own test surface. The SDK's parser lives inside
 ``PacketHandler.readTxRx`` and cannot be exercised without opening a port,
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Docstring-only correction to `strands_robots/drivers/dynamixel/protocol.py` plus a new test suite. The module's opening claim ("every function here is verifiable against Robotis' `dynamixel_sdk` byte-for-byte") and `_unstuff`'s self-description as "the mirror of Robotis' `removeStuffing`" pointed a future maintainer at a vendor unstuffer that is not the inverse of the vendor's own stuffer - `removeStuffing` compacts in place and its look-back reads positions its own loop has already overwritten, dropping a data byte on payloads carrying two reserved runs. The shipped implementation is correct and unchanged (the diff to `protocol.py` touches only docstrings); the new suite pins the round-trip identity over the full 21760-payload reserved-run corpus so the edit the old docstring invited now fails loudly.

## What's good

- Verified locally at head: the new suite is 8 passed / 3 skipped without `dynamixel-sdk` and 11 passed with it installed (4.0.5), so the identity pins run in CI regardless of the optional oracle - and the divergence claim itself (12 lossy payloads, witness `FF FF FD FF FD FD` -> `FF FF FD FF FD`) reproduces against the real SDK.
- The pre-fix split reproduces exactly as described: docstrings reverted, 4 failed, all in `TestTheDocstringsNameTheDivergence`, with the behaviour pins still green.
- Docstring gates (`test_docstring_xref_roles_resolve`, `test_docstrings_no_dangling_doc_refs`, `Args:`/`Raises:` completeness), `test_no_host_paths`, and the test-naming gate all pass on the head tree; ruff check/format clean; all-ASCII.
- Changelog fragment follows the `changelog.d/<pr-number>-<slug>.md` convention (AGENTS.md PR workflow step 3), and the test-class names describe behaviour rather than provenance.
- Correct adjudicator choice: grading the round trip rather than SDK parity means the pin cannot be satisfied by porting the byte loss in, which was the whole exposure.

PR #2804, opened 2026-08-26
fix(g1): hold the battery floor to a numeric domain before it is stored

cagataycali:
## The defect

`G1Driver.__init__` accepted `battery_floor_pct` and stored `float(battery_floor_pct)`
with no domain. That floor is the only constructor value `_check_motion_gates`
compares a live reading against, and the comparison is `battery_pct < floor`.

`float("nan")` survives the coercion, and `battery_pct < nan` is **False for
every reading**. So the driver stored a floor it reported through `get_status`
and enforced nowhere: the gate opened on a critically low pack. The string
`"nan"` took the same path, and `True` became a silent `1.0%` because `bool` is
an `int` subclass.

Measured on a connected driver whose FSM and lowstate were both healthy, so the
battery is the only thing in doubt, with the pack at 3.0%:

| `battery_floor_pct` | construction | outcome at battery = 3.0% |
|---|---|---|
| `15.0` (the default) | accepted | `battery 3.0% is under floor 15.0%` |
| `float("nan")` | accepted | **gate open, the write was allowed** |
| `"nan"` | accepted | **gate open, the write was allowed** |
| `True` | accepted | **gate open, the write was allowed** (floor `1.0`) |
| `None` | `TypeError` | bare builtin, outside the documented contract |

The two sibling numbers this driver takes, `duration` and `n_steps` on
`run_policy`, already go through a shared domain from `strands_robots.utils`.
The floor was the one parameter that did not.

## The change

Hold the floor to `finite_number_error` - the same shared domain, whose own
docstring covers this hazard ("`nan`/`inf` ... a silently poisoned pose rather
than a rejected command", and it "rejects `bool` explicitly") - **before** the
coercion, and name the refusal in a `Raises:` block.

Placement is load-bearing rather than stylistic: a guard after `float()` would
judge a value already stored, and would leave the refusals for `None`, a list
and a value past the float64 range as bare `TypeError`/`OverflowError` escaping
the constructor's contract. The break table below measures both.

## What is deliberately unchanged

Every value that behaves exactly as `battery_pct < floor` reads is untouched:
`0.0` and `-10.0` still never trip, `500.0` still always does, `20` and a NumPy
float still pass. A caller who writes `0.0` means "no floor" and gets it.

Whether a percentage should be bounded to `[0, 100]` is a question about what a
percentage may mean, not a correction - `500.0` refuses every pack and `-10.0`
refuses none, and both are consistent with the arithmetic. It is left open, and
the over-reach row below measures what narrowing it would cost.

## Why the shipped suite was silent

`sib` in the table is failures in the three pre-existing g1 suites
(`test_g1_driver.py`, `test_g1_control_loop.py`,
`test_g1_control_loop_terminal_observability.py`, 119 cells). It is **0 on every
row**: none of them constructs a driver with a floor other than a plain float,
so no mutation to this guard is visible to any of them.

## Mutation table

`mine` is failures in the new file, `coll` its collected count (stable at 47, so
no row hides behind a deselection).

| row | mutation | mine | coll | sib | fires |
|---|---|---|---|---|---|
| b0 | control, no mutation | 0 | 47 | 0 | - |
| b1 | guard removed (back to a bare `float()`) | 14 | 47 | 0 | regression + structural |
| b2 | guard moved **after** the store | 4 | 47 | 0 | structural + the 3 cases that become bare builtins |
| b3 | over-reach: `positive_finite_number_error` instead | 4 | 47 | 0 | the `0.0`/`-10.0` controls + structural |
| b4 | re-implemented locally with `math.isnan` only | 10 | 47 | 0 | misses `bool`, the string, `None`, a list, past-float64 |
| b5 | re-implemented with `not math.isfinite(float(v))` | 8 | 47 | 0 | misses `bool` and the non-real spellings |
| b6 | refusal logged instead of raised | 12 | 47 | 0 | regression |
| b7 | refusal coerced to the default instead of raised | 12 | 47 | 0 | regression |
| b8 | message drops the parameter name | 11 | 47 | 0 | regression |
| b9 | `Raises:` block dropped | 1 | 47 | 0 | structural |

9 mutations, **8 distinct firing sets**, control clean, source restored
md5-identical. b6 and b7 share a set for a structural reason - both mean "accept
instead of refuse", and the cells grade the refusal rather than which
substitution replaced it.

b4/b5 were re-run after a first attempt measured a broken probe: `math` is not
imported in this module, so the naive mutation raised `NameError` on every
construction and reported 70 sibling failures. The rows above add the import the
mutation would need, so they measure a weaker guard rather than a broken module.

## Pre-fix split

Source reverted, tests kept: **15 failed / 32 passed**.

| class | role | pre-fix |
|---|---|---|
| `TestANonFiniteFloorIsRefused` | regression | 12 of 12 |
| `TestTheGuardIsWiredWhereItHasToBe` | structural | 3 of 3 |
| `TestAFloorThatBehavesAsItReadsIsUntouched` | over-reach control | **0 of 12** |
| `TestThePremisesTheseCellsRestOn` | premise | **0 of 20** |

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1612 files).
- `mypy` **24 == 24** against a worktree of the merge-base, normalized message
  sets byte-identical in both directions, **0 attributable** to either changed file.
- Paired run over an identical 440-file selection (all `tests/drivers/`, all
  `tests/tools/g1/`, and every test that walks the tree, parses source or reads
  docstrings), the new file excluded from both trees and every path verified
  present on both: identical **22399 collected** and
  `20 failed, 22309 passed, 70 skipped` on each, **20 identical failing ids,
  `comm` empty in both directions**. All 20 are pre-existing and environmental:
  17 in `tests/mesh/test_iot_*` need `awsiot`/`awscrt` (`find_spec` False here,
  declared in the `[mesh-iot]` extra, which CI installs) and 3 in
  `tests/training/test_lerobot*` are the known lerobot-from-source
  `encode() takes 1 positional argument` drift.
- New file: **47 passed**.

No visual artifact: this changes a refusal and a stored value, not a policy,
simulation, render, recording or asset. The measured tables above are the evidence.

### DIFF
--- a/changelog.d/2804-g1-battery-floor-domain.md
+++ b/changelog.d/2804-g1-battery-floor-domain.md
@@ -1,0 +1,33 @@
+### Fixed: the G1's battery floor is held to a numeric domain before it is stored
+
+`G1Driver.__init__` took `battery_floor_pct` and stored `float(battery_floor_pct)`
+with no domain, and that floor is the only constructor value
+`_check_motion_gates` compares a live reading against. `float("nan")` survives
+the coercion, and `battery_pct < nan` is False for every reading, so the driver
+stored a floor it reported through `get_status` and enforced nowhere: the gate
+opened on a critically low pack. The string `"nan"` - how a config file spells
+it - took the same path, and `True` became a silent `1.0%` because `bool` is an
+`int` subclass. Measured on a connected driver whose FSM and lowstate were both
+healthy, with the pack at 3.0%:
+
+```
+battery_floor_pct        construction   outcome at battery = 3.0%
+15.0 (the default)       accepted       battery 3.0% is under floor 15.0%
+float("nan")             accepted       gate open - the write was allowed
+"nan"                    accepted       gate open - the write was allowed
+True                     accepted       gate open - the write was allowed   (floor 1.0)
+```
+
+The two sibling numbers this driver takes, `duration` and `n_steps` on
+`run_policy`, already go through a shared domain from `strands_robots.utils`.
+The floor now goes through `finite_number_error`, the shared domain whose own
+contract covers this hazard, before the coercion rather than after it - a guard
+placed after `float()` would judge a value already stored, and would turn the
+named refusals for `None`, a list and a value past the float64 range into bare
+`TypeError`/`OverflowError` escaping the constructor's documented contract.
+
+Every value that behaves as `battery_pct < floor` reads is untouched: `0.0` and
+`-10.0` still never trip, `500.0` still always does, and a NumPy float still
+passes. Whether a percentage should be bounded to `[0, 100]` is a separate
+question about what a percentage may mean rather than a correction, and it is
+deliberately left open.
--- a/strands_robots/drivers/g1.py
+++ b/strands_robots/drivers/g1.py
@@ -44,7 +44,11 @@ from typing import TYPE_CHECKING, Any, cast
 from strands_robots.mesh.pacing import Ticker
 from strands_robots.tools.g1 import HANDSHAKE_FSMS, WALK_FSMS, decode_code
 from strands_robots.tools.g1._dds_engine import DDSPublisher, DDSSubscriberSet
-from strands_robots.utils import positive_count_error, positive_finite_number_error
+from strands_robots.utils import (
+    finite_number_error,
+    positive_count_error,
+    positive_finite_number_error,
+)
 
 if TYPE_CHECKING:
     from strands.types.tools import ToolSpec, ToolUse
@@ -222,6 +226,9 @@ class G1Driver:
                 a caller can see which check refused.
             **kwargs: Ignored; accepted so the factory can forward extras
                 without the driver knowing what they are.
+
+        Raises:
+            ValueError: If ``battery_floor_pct`` is not a finite number.
         """
         del cameras, data_config  # accepted for parity; unused here
         if kwargs:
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes a real safety hole in `G1Driver.__init__`: `battery_floor_pct` was stored via a bare `float()`, so `float("nan")` (and its config-file spelling `"nan"`) produced a floor that `_check_motion_gates`'s `battery_pct < floor` comparison could never trip -- the gate opened on a critically low pack while `get_status` still advertised a floor. The fix routes the parameter through the existing shared domain `finite_number_error` from `strands_robots.utils` (the same layer `run_policy`'s `duration`/`n_steps` already use), placed before the coercion so `None`/list/`10**400` get the domain's named refusal instead of a bare `TypeError`/`OverflowError`, and documents the new `ValueError` in the constructor's `Raises:` block. The narrowing is loud, not silent: previously-coercible spellings (`"20"`, `True`) now raise a `ValueError` naming the parameter, the surface, and the value. The new test file covers both directions -- refusals for the ten unusable spellings and over-reach controls pinning that `0.0`/`-10.0`/`500.0`/NumPy floats still behave exactly as the comparison reads -- plus structural pins that the guard precedes the store and is the shared domain rather than a local copy. Deliberately leaving the `[0, 100]` bounding question open is the right scope call.

## What's good

- Reuses the shared value-domain guard per AGENTS.md rule 11 instead of adding a local one-caller check.
- `Raises:` block updated in the same change, satisfying the raises-docstring completeness rule.
- Changelog fragment (`changelog.d/2804-g1-battery-floor-domain.md`) follows the fragment convention; no direct `CHANGELOG.md` edit.
- Verified locally at head `8d4d4f2b`: the new file passes (47 passed), the full `tests/drivers/` suite passes (445 passed, 30 skipped), and `ruff check`/`ruff format --check` are clean on the changed files.
- Test names describe behaviour, ASCII-only strings, no host paths.

Issue #296, opened 2026-05-31
test(policies): `test_walk_continues_after_subpackage_decorator_failure` records imports post-exec, masking R1-1 trap origin


cagataycali:
## Context

Follow-up from PR #277 R7 review (thread `PRRT_kwDORUMiZs6F9XDS`, `tests/policies/lerobot_local/test_resolution.py:369`). Same root cause as already-filed #280, surfaced again on the R5 synthetic-tmpdir rewrite.

## Problem

In the synthetic-tmpdir test for the R1-1 walk-continues invariant, the in-memory `tracking_import` records imports via:

```python
exec(compile(source, str(module_file), "exec"), mod.__dict__)
imported_modules.append(name)  # runs AFTER exec
```

When `configuration_broken_policy.py` raises `RuntimeError` during `exec`, the `append("...configuration_broken_policy")` line never executes — so `broken_attempted` becomes `True` only via the **package-level fallback** (`lerobot.policies.broken_policy`) at the next candidate, not via the configuration-module trap that R1-1 actually exercises.

The assert at line 369 still passes (because `"broken_policy" in m` matches the package-fallback name), but it does not uniquely pin R1-1: a regression that drops the `(AttributeError, RuntimeError, TypeError)` widening but keeps a permissive package-level fallback would still pass.

## Fix

Record the import attempt **before** `exec`, with an explicit attempted-vs-succeeded distinction:

```python
def tracking_import(name, ...):
    ...
    attempted_imports.append(name)  # record BEFORE exec
    exec(compile(source, str(module_file), "exec"), mod.__dict__)
    successful_imports.append(name)
    ...
```

Then assert:

```python
assert any("configuration_broken_policy" in m for m in attempted_imports), (
    "R1-1 contract: walker must reach the trapped configuration module"
)
assert any("configuration_healthy_policy" in m for m in successful_imports)
```

This fails on pre-fix code (the trap aborts the walk before reaching `also_healthy`) AND fails on a regression that re-narrows the except clause to `ImportError` (the `RuntimeError` propagates), uniquely pinning R1-1.

## Relationship to #280

#280 raised the same concern about an earlier version of this test (the `pi0`-target version). PR #277 R5 rewrote the test to use a synthetic tmpdir, but the post-exec append pattern carried the same flaw forward. Either close #280 in favour of this issue or fold this into #280 — they are the same root cause across two test layouts.

## References

[...]

PR #2806, opened 2026-08-26
fix(episode_labels): keep the label sidecar's permissions across an update

cagataycali:
## The defect

`episode_labels._write_document` replaces the label sidecar the standard way: write a
`tempfile.mkstemp` file beside it, `os.replace` it into place. A `mkstemp` file is owner-only, so
the rename carried `0o600` onto the destination. A sidecar that arrived group- or world-readable
came out of the next write readable to nobody but the writer, and the write reported success.

Measured, on a sidecar at `0o644`:

| starting mode | after `record_deterministic_verdicts` | after `annotate_episode` |
|---|---|---|
| `0o644` | `0o600` | `0o600` |
| `0o664` | `0o600` | `0o600` |
| `0o640` | `0o600` | `0o600` |
| `0o444` | `0o600` | `0o600` |

The wider mode is the ordinary case, not an unusual one. The module docstring gives the sidecar's
whole purpose as travelling with the dataset directory - "at the dataset root, next to LeRobot's
own `meta/` / `data/` / `videos/` ... The sidecar travels with the dataset directory" - and every
way it travels lands it at the reader's umask: a `cp -r`, a `tar -x`, a Hub download, a clone. Each
of those is wider than `0o600` in at least one bit, so each is narrowed by the first annotation
written afterwards.

The cost surfaces later and elsewhere. With the sidecar unreadable, both readers fail:

```
read_labels      -> PermissionError: [Errno 13] Permission denied: .../episode_labels.json
filter_episodes  -> PermissionError: [Errno 13] Permission denied: .../episode_labels.json
```

So a training run on a shared dataset - a container on a non-root UID, a group-shared mount, a
second account on a training box - takes a `PermissionError` on labels it could read before, and
the write that caused it reported success and is long gone.

## Why this construct and not its sibling

`simulation/safe_output.atomic_write_bytes` is the same two-phase write and it ends with
`os.chmod(path, 0o600)`, documented: "the final file is `0o600` (the sim output roots are private
to the running user)". That is a decision with a stated reason. Here the same value arrived as
`mkstemp`'s default, with nothing said about it, on a file whose reason for existing is to be
carried around with a dataset. The two are the only `mkstemp`-based writers in the package; the
sibling sets a mode deliberately and this one did not set one at all.

## The change

The destination's mode is read before anything is written and applied to the **temp file**, so the
sidecar is never momentarily readable to fewer callers than it was - chmod-ing `path` after the
rename would leave exactly that window, and the ordering is graded rather than left to the next
edit.

A sidecar this module *creates* keeps `mkstemp`'s owner-only mode. Replacing content is not the
place that decides who may read a dataset, so this change does not widen anything; whether a freshly
created sidecar should match its dataset's umask is a separate question and is not answered here.

`0o444` is preserved too, and still updatable: `os.replace` needs write permission on the
directory, not on the file. A dataset published read-only keeps that posture through an annotation.

## Regression test

`tests/test_label_sidecar_update_keeps_its_permissions.py`, 22 cells. Pre-fix **14 failed / 8
passed**, by class:

|---|---|---|
| `TestAnUpdateKeepsTheSidecarReadableToTheSameCallers` | regression | 12 / 12 |
| `TestTheModeIsAppliedBeforeTheRename` | structural | 2 / 2 |
| `TestEveryWriterIsCovered` | derived inventory | 0 / 2 |
| `TestWhatIsUnchanged` | controls | 0 / 4 |
| `TestThePremise` | premise | 0 / 2 |

The headline failure is the one a reader can act on directly:

```
AssertionError: annotate_episode left the sidecar at 0o600 where it found 0o644;
a caller that could read the labels before the write cannot read them after.
```

The writers are discovered from the module's own call graph - a public function whose body calls
`_write_document` - rather than listed, so a third writer is held to the rule when it lands.

## Mutations

`mine` is the new file, `sib` is `test_episode_labels.py` + `test_episode_judge.py` +
`test_holdout_failure_mode_vocabulary.py` + `test_output_path_sandbox.py` (106 cells). `coll` is
recorded because two rows change it.

| mutation | coll | mine | sib |
|---|---|---|---|
| control | 22 | 0 | 0 |
| the fix reverted | 22 | 14 | 0 |
| `chmod(path)` **after** `os.replace` | 22 | 2 | 0 |
| `chmod(tmp_name)` after `os.replace` | 22 | 15 | 106 |
| a hardcoded `0o644` replaces the mode found | 22 | 9 | 0 |
| a **created** sidecar widened to `0o644` | 22 | 2 | 0 |
| raw `st_mode` carried across | 22 | 0 | 0 |
| suppression narrowed to `FileNotFoundError` | 22 | 0 | 0 |
| the mode read inside the `try` | 22 | 0 | 0 |
| an Nth writer planted, inventory **derived** | 22 | 1 | 0 |
| the same plant, inventory **hardcoded** | 22 | 0 | 0 |
| the mode grid narrowed to `0o644`, fix reverted | **16** | 8 | 0 |
| `if existing_mode:` instead of `is not None` | 22 | 1 | 0 |

Reading the rows that matter:

- **`chmod(path)` after the rename fires exactly the 2 structural cells and nothing else.** The end
  state is correct, so no behavioural cell can see it; the window it opens is why the structural
  half is there rather than decoration.
- **Widening a created sidecar fires 2 control cells.** The scope boundary is graded, not asserted.
- **`chmod(tmp_name)` after the rename takes out 106 sibling cells** - the temp no longer exists, so
  that ordering is a hard error rather than a policy. It is in the table because it shows the
  ordering is load-bearing at runtime as well as structurally.
- **The derived / hardcoded pair**: the same planted third writer is caught by the derived
  inventory (1, naming it) and is completely silent against a hardcoded list of today's two. A
  hardcoded list equal to the derived set fires nothing and still looks like a working test.
- **A narrowed mode grid runs 6 fewer cells** (22 -> 16) and reports 8 failures instead of 14, which
  is why `coll` is in the table.

Three rows fire 0 and each has a reason rather than a gap:

- Raw `st_mode` is a **semantic no-op** on Linux - `chmod` masks the file-type bits, so
  `0o100644` sets `0o644`. Measured, not assumed.
- Narrowing the suppression to `FileNotFoundError` is a no-op for the reachable case, which *is* the
  absent file.
- Reading the mode inside the `try` changes no observable behaviour; it is placed before the write
  so a failed write cannot leave the decision half-made.

One case is only reachable through the unit: `0o000` is the single falsy mode, so a truthiness
test would skip it and hand the file `0o600` - a widening. The public writers cannot reach it,
because they read the sidecar before rewriting it, so that cell drives `_write_document` directly
and says so.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1612 files).
- `mypy` **24 == 24** against a worktree of the merge-base, normalized message sets byte-identical
  in both directions, **0** attributable to the changed files.
- 417-file paired selection (every suite that walks the tree, parses source, or reads docstrings /
  `Args:` / `Raises:` / `changelog.d`, plus everything naming `episode_labels`, `episode_judge`,
  `safe_output` or `atomic_write`), the new file excluded from both and every path verified present
  on both trees: identical **20281 collected** and `20 failed, 20191 passed, 70 skipped` on each,
  the same 20 ids, `comm` empty in both directions, two distinct rootdirs.
- The 20 are pre-existing and environmental: **17** need `awsiot` / `awscrt` (`find_spec` False
  here, both declared in the `[mesh-iot]` extra) and **3** are the lerobot-from-source
  `TypeError: encode() takes 1 positional argument but 2 were given` drift in
  `tests/training/test_lerobot*`.

No visual artifact: this changes a file's permission bits and a docstring, not a policy, a
simulation, rendering, recording or an asset. The measured mode tables are the evidence.

### DIFF
--- a/changelog.d/2806-label-sidecar-update-keeps-its-permissions.md
+++ b/changelog.d/2806-label-sidecar-update-keeps-its-permissions.md
@@ -1,0 +1,28 @@
+### Fixed: updating the label sidecar no longer narrows its permissions
+
+`strands_robots.episode_labels._write_document` replaces the sidecar by writing
+a `tempfile.mkstemp` file next to it and `os.replace`-ing it into place. A
+`mkstemp` file is owner-only, so the rename carried `0o600` onto the
+destination: a sidecar that arrived group- or world-readable came out of the
+next annotation readable to nobody but the writer. That is the ordinary case
+rather than an unusual one. The module docstring gives the sidecar's whole
+purpose as travelling with the dataset directory, and every way it travels - a
+copy, a `tar -x`, a Hub download, a clone - lands it at the reader's umask,
+which is wider than `0o600` in at least one bit. Measured on a sidecar at
+`0o644`, both `record_deterministic_verdicts` and `annotate_episode` left it at
+`0o600` and reported success; the cost surfaced later and elsewhere, as a
+`PermissionError` out of `read_labels` or `filter_episodes` on labels another
+account could read a moment before.
+
+The destination's mode is now read before anything is written and applied to
+the temp file, so the sidecar is never momentarily readable to fewer callers
+than it was - chmod-ing `path` after the rename would leave exactly that
+window, and the ordering is graded rather than left to the next edit. A sidecar
+this module creates keeps `mkstemp`'s owner-only mode: replacing content is not
+the place that decides who may read a dataset, and this change does not widen
+anything. That is the one way the construct departs from
+`strands_robots.simulation.safe_output.atomic_write_bytes`, which holds its own
+output to `0o600` on purpose because the sim output roots are private to the
+running user - a reason that does not transfer to a dataset artifact, and the
+departure is now written down at both ends instead of arising from `mkstemp`'s
+default.
--- a/strands_robots/episode_labels.py
+++ b/strands_robots/episode_labels.py
@@ -75,6 +75,7 @@ from __future__ import annotations
 import contextlib
 import json
 import os
+import stat
 import tempfile
 import time
 from pathlib import Path
@@ -259,12 +260,38 @@ def _write_document(path: Path, document: dict[str, Any]) -> None:
     ``os.replace`` is atomic on POSIX and Windows for same-filesystem paths,
     so a reader never observes a truncated sidecar and a crashed writer
     leaves the previous version intact.
+
+    Replacing the content does not change the destination's permissions. The
+    temp file ``mkstemp`` creates is owner-only, so renaming it over an
+    existing sidecar would otherwise narrow one that arrived group- or
+    world-readable - and travelling with the dataset directory is what this
+    sidecar is for, so it commonly did (a copy, a ``tar -x``, a Hub download
+    and a clone all land it at the reader's umask). A caller that could read
+    the labels before an annotation can still read them after. The mode is
+    applied to the temp file rather than to ``path`` afterwards, so the
+    sidecar is never momentarily readable to fewer callers than it was.
+
+    This is the one way the construct departs from
+    :func:`strands_robots.simulation.safe_output.atomic_write_bytes`, which
+    holds its own output to ``0o600`` deliberately because the sim output
+    roots are private to the running user. A label sidecar is a dataset
+    artifact, and this function is not the place that decides who may read
+    the dataset. A sidecar it creates keeps ``mkstemp``'s owner-only mode.
     """
+    # Read the destination's mode before writing anything, so a failed write
+    # cannot leave the decision half-made. Absent is the create case, and the
+    # suppression covers it (FileNotFoundError is an OSError).
+    existing_mode: int | None = None
+    with contextlib.suppress(OSError):
+        existing_mode = stat.S_IMODE(path.stat().st_mode)
+
     fd, tmp_name = tempfile.mkstemp(dir=str(path.parent), prefix=path.name, suffix=".tmp")
     try:
         with os.fdopen(fd, "w", encoding="utf-8") as f:
             json.dump(document, f, indent=2, sort_keys=True)
             f.write("\n")
+        if existing_mode is not None:
+            os.chmod(tmp_name, existing_mode)
         os.replace(tmp_name, path)
     except BaseException:
         # Cleanup-and-reraise: never leave the temp file behind on failure.
--- a/tests/test_label_sidecar_update_keeps_its_permissions.py
+++ b/tests/test_label_sidecar_update_keeps_its_permissions.py
@@ -1,0 +1,309 @@
+"""Updating the label sidecar leaves its permissions alone.
+
+:func:`strands_robots.episode_labels._write_document` replaces the sidecar by
+writing a ``mkstemp`` temp file and ``os.replace``-ing it into place. A
+``mkstemp`` file is owner-only, so the rename used to carry ``0o600`` onto the
+destination and silently narrow a sidecar that arrived group- or
+world-readable. Travelling with the dataset directory is what the sidecar is
+for, and every way it travels - a copy, a ``tar -x``, a Hub download, a clone -
+lands it at the reader's umask, so the wider mode is the ordinary case rather
+than an unusual one. The write reported success either way; the cost showed up
+later, as a ``PermissionError`` from :func:`read_labels` or
+:func:`filter_episodes` on a dataset whose labels were readable a moment
+before.
+
+What these cells hold:
+
+- The regression: an update through either writer leaves the mode it found.
+  Parametrized over the modes a shared dataset actually carries, and over both
+  writers, discovered from the module rather than listed so a third writer is
+  held to the rule the hour it lands.
+- The structural half: the mode is applied to the temp file *before* the
+  rename, not to the destination after it, so the sidecar is never momentarily
+  readable to fewer callers than it was.
+- What must not change: the atomicity contract the construct exists for - a
+  failed write leaves no temp file and the previous sidecar byte-identical -
+  and the content of a successful update.
+- The premise: ``mkstemp`` really does create an owner-only file, so the
+  narrowing was a property of the construct and not of one platform; and
+  :func:`strands_robots.simulation.safe_output.atomic_write_bytes` really does
+  hold its own output to ``0o600`` on purpose, which is the contrast that keeps
+  this change to the sidecar.
+"""
+
+from __future__ import annotations
+
+import ast
+import inspect
+import os
+import stat
+import tempfile
+from collections.abc import Callable
+from pathlib import Path
+from typing import Any
+
+import pytest
+
+import strands_robots.episode_labels as labels
+
+# Modes a dataset's own files carry in the wild: the default umask, a
+# group-writable shared mount, a group-readable team dataset, and a read-only
+# archive extraction. Every one is wider than ``mkstemp``'s ``0o600`` in at
+# least one bit, so each would have been narrowed.
+_SHARED_MODES = (0o644, 0o664, 0o640, 0o444)
+
+# The private mode the temp file arrives with. Named here rather than read from
+# the module so a cell that grades the module's behaviour is not comparing the
+# module against itself.
+_MKSTEMP_MODE = 0o600
+
+
+def _mode(path: Path) -> int:
+    """Permission bits of ``path``."""
+    return stat.S_IMODE(path.stat().st_mode)
+
+
+def _seeded_dataset() -> Path:
+    """A dataset root whose sidecar exists and carries one verdict."""
+    root = Path(tempfile.mkdtemp())
+    labels.record_deterministic_verdicts(root, [{"episode": 0, "success": True}])
+    return root
+
+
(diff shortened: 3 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Fixes a silent permission narrowing in `episode_labels._write_document`: the two-phase write (`mkstemp` + `os.replace`) carried the temp file's owner-only `0o600` onto an existing sidecar, so any group- or world-readable sidecar became unreadable to other accounts after the next annotation, surfacing later as a `PermissionError` from `read_labels` / `filter_episodes`. The fix reads the destination's mode before writing and applies it to the temp file ahead of the rename, so there is no window where the sidecar is readable to fewer callers than before. Scope is disciplined: the create case keeps `mkstemp`'s private mode (nothing is widened), the atomicity/cleanup contract is untouched, and the chmod sits inside the existing cleanup-and-reraise block so a failure cannot orphan the temp file. I verified the preserve-across-replace semantics (including the falsy `0o000` and read-only `0o444` cases) independently and they hold.

## What's good

- Regression test derives the writer inventory from the module's own call graph rather than a hardcoded list, so a third writer is graded on arrival - and the PR's mutation table demonstrates the derived/hardcoded difference empirically.
- Structural cells pin the chmod-before-replace ordering and the temp-file target, which no behavioural cell can observe.
- Docstring records why this construct deliberately departs from `safe_output.atomic_write_bytes`'s `0o600` policy, at both ends (module docstring and changelog fragment per AGENTS.md step 3).
- ASCII-only strings, behaviour-named test classes, no host paths, `except BaseException` cleanup-and-reraise pattern preserved.

cagataycali:
## The mode fixtures reached `os.chmod` as integer literals, and the alias did not stop that

`44863d6d` routed the four staged masks through `_GROUP_READABLE = _SHARED_MODES[0]` and
`_READ_ONLY = _SHARED_MODES[3]`. The intent is right and I kept it. The mechanism did not hold:
the analysis that opened those four review threads binds its mode by

```
mode = call.getParameter(1, "mode").getAValueReachingSink().asExpr().(IntegerLiteral).getValue()
```

so it needs an integer literal *reaching the sink*, not a literal written at the sink. A tuple
subscript is a flow it traverses, so `0o644` still arrived through `_SHARED_MODES[0]`.

Measured rather than argued. `44863d6d` produced four **new** alerts at the four new line
numbers, on its own merge ref:

| head | line | alert | verdict |
|---|---|---|---|
| `f6836755` (literal at the call) | 141, 176, 246, 259 | 987-990 | open |
| `44863d6d` (tuple-subscript alias) | 147, 182, 252, 265 | 991-994 | open |

Same rule, same four cells, four fresh alert numbers. The alias moved the literal one hop and
the flow followed it.

### What this head changes

The modes are now composed from the `stat` bits the assertions already read, and the grid is
built from the named modes rather than the names being indexes into the grid:

```python
_OWNER_RW = stat.S_IRUSR | stat.S_IWUSR
_GROUP_OTHER_READ = stat.S_IRGRP | stat.S_IROTH

_GROUP_READABLE = _OWNER_RW | _GROUP_OTHER_READ   # 0o644 default umask
_GROUP_WRITABLE = _GROUP_READABLE | stat.S_IWGRP  # 0o664 shared mount
_TEAM_READABLE = _OWNER_RW | stat.S_IRGRP         # 0o640 team dataset
_READ_ONLY = stat.S_IRUSR | _GROUP_OTHER_READ     # 0o444 archive extraction

_SHARED_MODES = (_GROUP_READABLE, _GROUP_WRITABLE, _TEAM_READABLE, _READ_ONLY)
```

The no-drift property is kept - the cells and the sweep now share the same objects - and the
names still single-source the data. What changes is that a `BinOp` is not an `IntegerLiteral`,
so no dataflow depth can bind a mask here. This is a property of the expression rather than of
how far an analysis reaches, which is the part the subscript form was relying on.

It is also the file's own vocabulary: the cell whose subject is the group bit already asserts
`_mode(sidecar) & (stat.S_IRGRP | stat.S_IROTH)`. Staging that case from the same constants
says which bit it is about instead of leaving a reader to decode an octal mask.

### Not a suppression

`py/overly-permissive-file` is security-severity high, and `.github/codeql/codeql-config.yml`
excludes two note-severity style rules and says why its scope is exactly two. This rule is not
comparable to either, and the alert is correct about the call: the fixture really does make the
file world-readable, because that is the property under test. So the mask belongs off the flow,
not on an exclusion list. Note there is one accepted instance of this rule on `main`
(`strands_robots/mesh/iot/provision.py:1012`, alert #296); these four were new-in-PR, which is
what opened them as blocking threads.

### While those cells were being read

Each of the two individually-staged cells graded one of the four modes the writer must
preserve. Both now sweep the whole grid, so the group-readable and failed-write cases cover the
group-writable, team-readable and read-only modes too.

| | before | after |
|---|---|---|
| cells | 22 | 34 |
| pre-fix failures (production reverted to `main`) | 14 | 20 |
| failures in control, premise and over-reach classes | 0 | 0 |

Pre-fix, the 20 land 18 in `TestAnUpdateKeepsTheSidecarReadableToTheSameCallers` and 2 in
`TestTheModeIsAppliedBeforeTheRename`, with `TestEveryWriterIsCovered`, `TestWhatIsUnchanged`,
`TestTheModeFixturesAreTheModesTheyName` and `TestThePremise` all clean. A new premise class
pins the composed bits to the octal modes their comments claim, asserts every grid member is
wider than `mkstemp`'s in at least one bit (non-vacuity: a mode no wider could not be narrowed),
and asserts the read-only mode really carries no write bit.

### Gate

`ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1614
files). `mypy` 24 errors, byte-identical normalized message sets against `main` in both
directions, 0 attributable. The 161-file suite of tests that walk the tree, parse source or read
docstrings: 6789 passed, 56 skipped, 0 failed. Production is byte-identical to `44863d6d`; the
diff is the test file alone.

yinsong1986 (approved):
## Summary

Fixes a silent permission narrowing in `episode_labels._write_document`: the two-phase write (`mkstemp` + `os.replace`) carried the temp file's owner-only `0o600` onto an existing sidecar, so a group- or world-readable sidecar became unreadable to every other account after the next annotation, surfacing later as a `PermissionError` from `read_labels` / `filter_episodes`. The fix reads the destination's mode before anything is written and applies it to the temp file ahead of the rename, so no window exists where the sidecar is readable to fewer callers than it was; the `is not None` guard keeps `0o000` from being widened, and a sidecar the module *creates* deliberately keeps `mkstemp`'s owner-only mode, so nothing is widened. The regression test derives the writer set from the module's own call graph, grades the chmod-before-replace ordering structurally, and pins the atomicity contract and the create-case scope boundary as controls.

## What's good

- The scope boundary is explicit and graded: replacing content carries the mode across, creating a sidecar does not decide who may read the dataset, and both directions have cells.
- The departure from `safe_output.atomic_write_bytes` (which holds `0o600` on purpose) is documented at both ends and pinned by a premise test, so the two `mkstemp` writers cannot drift into looking inconsistent by accident.
- The mode-fixture refactor in the head commit keeps the named bits honest with `TestTheModeFixturesAreTheModesTheyName`, so the composed `stat` constants cannot silently stop meaning the octal modes the comments claim.
- Changelog fragment present per AGENTS.md step 3; the `except BaseException` cleanup-and-reraise matches the documented obliged idiom.

yinsong1986 (approved):
## Summary

Fixes a silent permission narrowing in `episode_labels._write_document`: the two-phase write (`mkstemp` + `os.replace`) carried the temp file's owner-only `0o600` onto an existing sidecar, so a group- or world-readable sidecar - the ordinary state for a file whose purpose is to travel with the dataset directory - became unreadable to every other account after the next annotation, surfacing later as a `PermissionError` from `read_labels` / `filter_episodes`. The fix reads the destination's mode before anything is written (`stat.S_IMODE`, so file-type bits are masked) and applies it to the temp file ahead of the rename, so there is no window where the sidecar is readable to fewer callers than it was. The `is not None` guard keeps the one falsy mode (`0o000`) from being widened, the create case deliberately keeps `mkstemp`'s owner-only default so this function never widens anything, and the departure from `safe_output.atomic_write_bytes`'s intentional `0o600` is documented at both ends. The regression test discovers the sidecar writers from the module's own call graph rather than a hardcoded list, grades the chmod-before-replace ordering structurally, and pins the atomicity contract (failed write leaves no temp file and the previous sidecar byte-identical) as controls.

## What's good

- Ordering is load-bearing and graded: the mode goes onto the temp file before `os.replace`, and `TestTheModeIsAppliedBeforeTheRename` pins that structurally so a later edit cannot quietly reopen the narrowed-visibility window.
- Scope discipline: the create path is unchanged and the change never widens a mode - it only stops the update path from deciding who may read a dataset.
- Writer inventory is derived (`_writers_in_the_module` walks the AST for `_write_document` callers) with a non-vacuity control, so a third writer is held to the rule on arrival - and the PR's mutation table shows the derived/hardcoded pair separating on a planted writer.
- Changelog fragment per AGENTS.md step 3; cleanup-and-reraise handler shape untouched; test and production strings ASCII-clean.

PR #2807, opened 2026-08-26
fix(simulation): answer gripper_body from one gripper-hint vocabulary

cagataycali:
## The defect

`list_bodies(robot_name=...)` advertises a best-guess gripper/end-effector mount in its `gripper_body` field -- the surface a caller reads to resolve `add_camera(parent_body=...)` for a wrist view. Both backends matched that field through one shared rule, `simulation.ik.hint_matches_name`, and each kept a hint vocabulary of its own under the same name:

| constant | value |
|---|---|
| `mujoco.simulation._GRIPPER_BODY_HINTS` | `("gripper", "hand", "ee", "tool")` |
| `newton.simulation._GRIPPER_BODY_HINTS` | `("gripper", "hand", "jaw", "ee", "tool")` |

Same name, same purpose, and `newton - mujoco == ("jaw",)` with `mujoco - newton == ()`. So on a shipped registry robot the two backends answer differently:

| robot | MuJoCo advertises | Newton advertises |
|---|---|---|
| `so100` | *nothing* | `Fixed_Jaw`, `Moving_Jaw` |
| `so101` | `gripper` | `gripper`, `moving_jaw_so101_v1` |

Driven end to end through the real MuJoCo backend on `so100` (7 bodies), the payload contradicts itself:

```
list_bodies status : success
gripper_body       : None
bodies containing 'jaw' : ['arm/Fixed_Jaw', 'arm/Moving_Jaw']
```

The field says there is no mount; the list beside it names two. The `text` block's `Gripper/EEF mount:` line is omitted entirely.

## Why the narrower set is the wrong one

Three places in the tree already say `jaw` belongs:

- `motion_primitives_base._GRIPPER_HINTS` is `("gripper", "finger", "jaw")`, and its comment gives the reason outright: *"SO-100's gripper joint is the 'Jaw'"*. It reads the same vocabulary to resolve the gripper *joint*.
- `docs/simulation/newton.md` documents `sim.list_bodies("so100")[...]["gripper_body"]` with the output `"so_arm100/.../Fixed_Jaw"` -- the answer MuJoCo could not give.
- `docs/simulation/world-building.md` taught the narrower list as the whole rule and told the reader that `gripper_body: None` means *"A robot with no gripper-like body"*. For an SO-100 that reads as a false statement about the arm.

The Newton constant's own comment justified the divergence as *"Newton adds 'jaw' (its own MJCF vocabulary)"*. Both backends read MJCF, and here they read the same file, so that reason does not hold.

## The change

`hint_matches_name`'s docstring claimed that sharing the matcher meant *"the same name cannot be an end-effector to one surface and not to another"*, in the same paragraph that licensed *"Each caller keeps its own hint vocabulary"*. The second sentence is how the first stops being true: agreement needs a shared vocabulary as well as a shared rule.

So `gripper_body` -- one question about one robot -- gets one owner, `simulation.ik.GRIPPER_BODY_HINTS`, which both backends read. `discover_ee_frame` keeps `_BODY_HINTS` / `_SITE_HINTS` of its own on purpose: it answers a different question, which frame to solve IK *to*, and it still resolves the SO-100's `Wrist_Pitch_Roll` rather than a jaw. The two are pinned as a difference so neither can be quietly folded into the other. Both docstrings that enumerated the words in prose, and the docs page, now name the shared set.

## Corpus safety

Over the 59 loadable registry assets:

| | |
|---|---|
| answer unchanged | **58** |
| answer changed | **1** -- `so100`, `None` -> `Fixed_Jaw` |
| bodies the added word matches, corpus-wide | **3** -- `so100/Fixed_Jaw`, `so100/Moving_Jaw`, `so101/moving_jaw_so101_v1` |
| of those, not a real gripper jaw | **0** |

`so101` is unchanged because its `gripper` body precedes its jaw in body order.

## Why nothing caught it

`tests/simulation/test_ee_frame_hint_word_boundary.py` states the property in its own section comment -- *"the parity that matters to a caller - the same robot reports the same mount on either backend"* -- and never measures it:

- `test_both_surfaces_agree_on_the_same_robot` compares `list_bodies` against `discover_ee_frame`, not one backend against the other, and its fixture body is `link_gripper_slider`, which matches `gripper` -- a hint **both** vocabularies carry. A fixture spelled that way cannot see a vocabulary drift.
- `test_newton_list_bodies_still_advertises_a_jaw` pinned the divergence as intentional (*"Newton's own extra hint"*). Its docstring is corrected here; the cell now measures that sharing the word did not cost the match it was carried for.

Every mutation below fires **0** tests in the pre-existing suites except the two noted.

## Tests

`tests/simulation/test_gripper_mount_vocabulary_is_shared.py`, 19 cells. Pre-fix split with the MuJoCo backend given its own narrow set back: **6 failed / 13 passed**.

| class | hit | role |
|---|---|---|
| `TestThePremiseIsReal` | 0/4 | the drift was reachable through a shipped asset |
| `TestBothBackendsAdvertiseTheSameMount` | 3/3 | regression |
| `TestTheShippedAssetResolvesItsMount` | 1/1 | regression |
| `TestTheVocabularyHasOneOwner` | 2/3 | derived, so a third backend is graded on arrival |
| `TestWhatMustNotChange` | 0/4 | over-reach controls |

### Mutations

| mutation | fires | pre-existing | note |
|---|---|---|---|
| control | 0 | 0 | 19 collected throughout |
| MuJoCo keeps its own narrow set (the defect) | 6 | **0** | nothing in the tree sees it |
| Newton keeps its own set, contents equal | 2 | 0 | structural only; behaviourally a no-op today |
| both keep own sets, contents equal | 2 | 0 | same set - the drift guard, not a behaviour |
| shared set loses `jaw` | 6 | 1 | trips the pre-existing Newton jaw pin |
| shared set gains `knee` (over-reach) | 3 | **3** | pre-existing knee/wheel tests refuse it |
| IK's `_BODY_HINTS` gains `jaw` (vocabularies folded) | 2 | 0 | the difference is pinned |

All three sources restored byte-identical after every row (md5 asserted).

## Gate

- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ`: clean, 1614 files.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a worktree of the merge-base, normalized message sets byte-identical both directions, 0 attributable to the changed files.
- Paired run, identical 823-file selection (all of `tests/simulation`, plus every guard that walks the tree, parses source, or reads docstrings and `docs/`), the new file excluded from both and every path verified present on both: identical **30731 collected** and `25 failed, 30412 passed, 294 skipped` on each tree, **25 == 25 failing ids with `comm` empty in both directions**, distinct rootdirs. Of the 25: 17 need `awsiot`/`awscrt` (`find_spec` False, declared in the `[mesh-iot]` extra), 3 are the lerobot-from-source `encode()` drift, and 5 pass **29/29 in isolation on both trees**.

## Artifact

`gripper_body` exists so a caller can mount a wrist camera, so the artifact is that call. One capture script run against each tree with the same scene, asset, camera parameters and headlight; the external view agrees between the two runs to within **1** intensity level, which is EGL's cross-process reproducibility rather than a difference between the trees. Panel 2 is a text panel because the call that would create its camera cannot be made.

gripper_body resolves the mount the payload already listed

### DIFF
--- a/changelog.d/2807-gripper-mount-vocabulary-is-shared.md
+++ b/changelog.d/2807-gripper-mount-vocabulary-is-shared.md
@@ -1,0 +1,25 @@
+### Fixed: `gripper_body` is answered from one gripper-hint vocabulary
+
+`list_bodies(robot_name=...)` advertises a best-guess gripper/end-effector mount in its
+`gripper_body` field -- the surface a caller reads to resolve `add_camera(parent_body=...)` for a
+wrist view. Both backends matched that field through one shared rule, `hint_matches_name`, and each
+kept a hint vocabulary of its own. The vocabularies had drifted: only one of them carried `jaw`.
+
+So the SO-100 reported its mount on one backend and reported `gripper_body: None` on the other, in
+the same payload that listed both of its jaw bodies. It is a shipped registry robot whose gripper
+bodies are named `Fixed_Jaw` and `Moving_Jaw`, `docs/simulation/newton.md` already documented its
+`gripper_body` as a jaw, and `docs/simulation/world-building.md` documented the narrower vocabulary
+as the whole rule and told the reader that `None` means the robot has no gripper-like body. On a
+loaded SO-100 the field read `None` while the bodies list beside it read `arm/Fixed_Jaw`,
+`arm/Moving_Jaw`.
+
+Sharing the matcher is not on its own enough for two surfaces to agree; they agree only where they
+also read the same vocabulary. `gripper_body` is one question about one robot, so the vocabulary
+behind it now has one owner, `simulation.ik.GRIPPER_BODY_HINTS`, which both backends read.
+`discover_ee_frame` keeps its own words on purpose: it answers a different question -- which frame to
+solve IK *to* -- and still resolves the SO-100's wrist rather than a jaw. The two are pinned as a
+difference so neither can be quietly folded into the other.
+
+Measured over the 59 loadable registry assets, 58 report the mount they always reported and one
+changes: `so100`, from `None` to `Fixed_Jaw`. The added word matches three bodies in the whole
+corpus and every one of them is a real gripper jaw, so no asset gains a mount it should not have.
--- a/docs/simulation/world-building.md
+++ b/docs/simulation/world-building.md
@@ -592,7 +592,7 @@ sim.add_camera(name="wrist", parent_body=mount,
 
 `list_bodies()` (no `robot_name`) lists every body in the world; with `robot_name` it scopes to that robot and also returns `gripper_body`, the best-guess end-effector mount.
 
-The guess matches its hint words (`gripper`, `hand`, `ee`, `tool`) on word boundaries, so a short hint cannot fire inside an unrelated word - a `knee` link or a `wheel` hub is not a gripper mount because `ee` occurs in its name. A robot with no gripper-like body reports `gripper_body: None` and omits the mount line rather than naming an unrelated body; pick the mount from the full `bodies` list in that case.
+The guess matches its hint words (`gripper`, `hand`, `jaw`, `ee`, `tool` - one set, read by every backend) on word boundaries, so a short hint cannot fire inside an unrelated word - a `knee` link or a `wheel` hub is not a gripper mount because `ee` occurs in its name. A robot with no gripper-like body reports `gripper_body: None` and omits the mount line rather than naming an unrelated body; pick the mount from the full `bodies` list in that case. `jaw` is in the set because the SO-100 family names its gripper bodies `Fixed_Jaw` / `Moving_Jaw`, so the mount resolves for those arms too.
 
 A mounted camera survives `remove_robot`, which rebuilds the whole scene: it is
 re-mounted on its body once every surviving robot is re-attached, keeping its
--- a/strands_robots/simulation/ik.py
+++ b/strands_robots/simulation/ik.py
@@ -386,6 +386,19 @@ class MinkIKBridge:
 
 _SITE_HINTS = ("attachment_site", "attachment", "grasp", "pinch", "tcp", "ee_site", "ee", "flange")
 _BODY_HINTS = ("hand", "gripper", "tool", "tcp", "ee", "wrist", "flange", "end_effector", "eef")
+
+# Hint words for the gripper/EEF mount every backend's ``list_bodies``
+# advertises in its ``gripper_body`` field. Shared rather than per-backend:
+# ``gripper_body`` is one question about one robot, so a body that is the mount
+# on one backend has to be the mount on every other. ``jaw`` is here because
+# the SO-100 family names its gripper bodies ``Fixed_Jaw`` / ``Moving_Jaw`` -
+# see :data:`~strands_robots.simulation.motion_primitives_base._GRIPPER_HINTS`,
+# which reads the same vocabulary to resolve the gripper *joint*.
+#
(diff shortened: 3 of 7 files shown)


yinsong1986 (approved):
## Summary

Unifies the `gripper_body` hint vocabulary that each backend's `list_bodies` reads into a single shared constant, `strands_robots.simulation.ik.GRIPPER_BODY_HINTS`, replacing the two per-backend `_GRIPPER_BODY_HINTS` copies that had drifted (MuJoCo lacked `jaw`, so the shipped SO-100 - whose gripper bodies are `Fixed_Jaw` / `Moving_Jaw` - reported `gripper_body: None` on MuJoCo while Newton resolved the mount). The matching rule (`hint_matches_name`, word-boundary) is unchanged; the IK-target vocabularies (`_BODY_HINTS` / `_SITE_HINTS`) deliberately stay separate, and the new comment in `ik.py` explains why the two questions are not interchangeable. Verified against the current tree: the drift is real (`mujoco/simulation.py` ships a 4-word tuple, `newton/simulation.py` a 5-word one), the Newton vocabulary is byte-identical to the new shared one so that backend's behavior is unchanged, and the referenced `motion_primitives_base._GRIPPER_HINTS` exists. The JSON payload shape (`gripper_body: str | null`) is untouched, so no wire/schema concern; the MuJoCo answer widening from `None` to a real jaw mount is the fix itself, on a field documented as a nullable best-guess.

## What's good

- Changelog fragment present (`changelog.d/2807-gripper-mount-vocabulary-is-shared.md`), per the step-3 rule.
- The new test suite pins the one-owner property structurally (AST scan refusing a second `*GRIPPER_BODY_HINTS` definition, and a source-level check that every answering backend reads the shared name), so the drift class cannot silently recur - and a premise cell (`test_the_driven_backends_are_every_backend_that_answers`) makes a future backend either driven or failing.
- `test_so101_reports_the_mount_it_always_reported` and `test_ik_still_prefers_a_wrist_to_a_jaw_on_the_same_arm` pin the two answers that must NOT move, which is exactly the risk of widening a hint set.
- The updated docstring in `test_newton_list_bodies_still_advertises_a_jaw` keeps the existing test's rationale accurate after the semantic change, and docs/world-building.md is updated in the same PR.
- No emojis, no host paths, ASCII-clean tool-result strings; test names describe behavior rather than provenance.

yinsong1986 (approved):
## Summary

Unifies the `gripper_body` hint vocabulary behind a single owner, `strands_robots.simulation.ik.GRIPPER_BODY_HINTS`, replacing the two per-backend `_GRIPPER_BODY_HINTS` tuples that had drifted (MuJoCo's set lacked `jaw`, so a shipped SO-100 reported `gripper_body: None` in the same payload that listed its `Fixed_Jaw`/`Moving_Jaw` bodies). Both backends' `list_bodies` now read the shared set; `discover_ee_frame`'s own `_BODY_HINTS`/`_SITE_HINTS` are deliberately left distinct and that difference is pinned. The blast radius is measured and narrow: one registry asset changes answer (`so100`, `None` -> `Fixed_Jaw`), the added word matches only real jaw bodies corpus-wide, and the rung-1 site-search dedupe next to the new constant is untouched. The new test suite drives both real backends end to end, derives the graded backend set from the factory registry so a third backend is graded on arrival, and enforces single ownership of the vocabulary via an AST scan. Changelog fragment and both docs pages are updated in the same PR. No must-fix concerns found.

PR #2808, opened 2026-08-26
fix(transforms): bind a pipeline class the same way its import path binds it

cagataycali:
## Problem

`CosmosTransferTransform` binds a video2video pipeline two ways: as an object, or as a dotted import path resolved lazily. Only the dotted-path branch applied the construction step, and its own comment states why the step exists:

> A class target (or a factory with no generate of its own) is constructed zero-arg; an unconstructed class would otherwise pass the generate() probe and then receive the video as its `self`.

The object branch skipped it. So the same class bound cleanly through one form and broke through the other. Measured on `main`:

| `pipeline=` | `validate()` | generation |
| --- | --- | --- |
| an instance | no problems | success |
| **a class (unconstructed)** | **no problems** | **`TypeError: Adapter.generate() missing 1 required positional argument: 'video'`** |
| dotted path to that class | no problems | success (constructed) |
| dotted path to a factory | no problems | success (constructed) |
| dotted path to an instance | no problems | success |

`validate()` exists so a wiring mistake is a reported problem *before* an episode is read or an output recorder is created — its own docstring says "nothing is read or written without one". Here it approved a pipeline that cannot be called, and the failure then arrived from inside generation as a bare `TypeError`: neither the `error` envelope `transform()` returns for a wiring problem, nor the `ValueError` its `Raises:` block documents. On a real generative backend that is minutes of work and a created output dataset before the report.

It is also a one-character slip, because the documented adapter recipe in the module docstring and in `_PIPELINE_HELP` is spelled `class Adapter:` — so `pipeline=Adapter` against `pipeline=Adapter()`.

## Fix

The construction step becomes a single owner (`_constructed_if_needed`) that both branches consult, rather than being restated in one of them. The read of `generate` already had one guarded owner and a structural test pinning it; the construction that must accompany that read did not.

Deliberately construct rather than refuse: constructing is what the sibling branch already does for the identical value, it makes a value bind identically whichever form carries it, and refusing would have regressed two shapes that work today.

## No previously usable pipeline is refused

Measured on `main` before choosing the fix's shape, which is what ruled out a blanket refusal:

| class shape | before | after |
| --- | --- | --- |
| `generate` is an instance method | `TypeError` at generation | success |
| `generate` is a `@staticmethod` | success | success |
| `generate` is a `@classmethod` | success | success |
| constructor needs arguments | `TypeError` naming `'video'` | refused, naming `is not constructible zero-arg` |

The last row is the second improvement: the refusal now names the constructor instead of blaming a missing `video` argument.

## Tests

`tests/transforms/test_cosmos_transfer.py` gains a regression class, a structural class, and an over-reach control class. Pre-fix split with the source at `main` and the new tests kept:

```
6 failed, 67 passed
  TestAClassBindsTheSameWayAsItsImportPath        4 of 6   REGRESSION
  TestTheBindingFormsShareOneConstructionRule     2 of 2   REGRESSION (structural)
  TestNoPreviouslyUsablePipelineIsRefused         0 of 3   over-reach control
  (six pre-existing classes)                     0 of 26
```

Post-fix: `74 passed`.

The class's existing probe-handler rule promises that "a probe point added later cannot re-derive an unguarded read", and its universe was a hardcoded pair of function names, which cannot keep that promise for a function the module does not have yet. It is now derived from the module, so the new owner is held to the rule on arrival — that is the `+1` in the collected count.

## End-to-end verification

Against a recorded three-episode LeRobotDataset (unequal lengths `[7, 13, 5]`, 25 frames), with the pipeline bound as a class — the exact call that raised `TypeError` before:

```
validate: []
transform: success  (read 3 episode(s), wrote 3 generated episode(s))
output parquet truth: 3 episodes, 25 frames, [7, 13, 5]
reopened via LeRobotDataset: 25 frames, 3 episodes, ['observation.images.default']
provenance: 3 records, transform_version = 'e2e-transfer-1.0'
pixels changed: True        action byte-identical to source: True
```

`transform_version` is read from the constructed instance, so provenance pins the object that produced the pixels. The non-image pass-through guarantee is unaffected.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1614 files).
- `mypy strands_robots tests tests_integ`: 19 pre-existing errors, all in `examples/isaac_gs/` (reached through followed imports), zero in the changed files.
- `tests/transforms/`: 380 passed.
- 118 tree-walking guard files plus the transforms suite: 5007 passed, 55 skipped.

Docs: `__init__`'s `pipeline:` entry documented the object form as "a constructed object"; it now documents the domain the seam actually resolves, and states that a class or factory is constructed once during `validate()`.

### DIFF
--- a/changelog.d/2808-pipeline-class-binds-like-its-import-path.md
+++ b/changelog.d/2808-pipeline-class-binds-like-its-import-path.md
@@ -1,0 +1,22 @@
+### Fixed: a `cosmos_transfer` pipeline class now binds the same way its import path binds it
+
+The video2video pipeline seam accepts a backend two ways - as an object, or as
+a dotted import path resolved lazily - and only one of them applied the
+construction step. The dotted-path branch constructs a class or zero-arg
+factory target before probing it, because an unconstructed class passes the
+`generate()` probe as a plain function and then receives the video as its
+`self`. The object branch skipped that step, so the same class bound cleanly
+through one form and failed through the other: `validate()` returned no
+problems for `pipeline=Adapter`, and generation then raised
+`TypeError: Adapter.generate() missing 1 required positional argument: 'video'`
+from inside the run - neither the `error` envelope `transform()` returns for a
+wiring problem nor the `ValueError` it documents, and after the source dataset
+was opened and an output recorder created.
+
+`validate()` exists so a wiring mistake is reported before any of that work
+happens, and the documented adapter recipe is spelled `class Adapter:`, so
+`Adapter` against `Adapter()` is a one-character slip. Construction is now a
+single owner both branches consult, so neither can re-derive it. A class whose
+`generate` is a `staticmethod` or a `classmethod` already bound usably and
+still does; a class whose constructor needs arguments now names its
+constructor rather than a missing `video`.
--- a/strands_robots/transforms/cosmos_transfer.py
+++ b/strands_robots/transforms/cosmos_transfer.py
@@ -122,6 +122,58 @@ def _generate_surface(obj: Any, subject: str) -> tuple[Any, str | None]:
         )
 
 
+def _pipeline_subject(obj: Any) -> str:
+    """Name the caller's binding in a problem, so a class names itself.
+
+    ``type(a_class).__name__`` is ``"type"``, which would report the remedy
+    against a subject the caller never wrote.
+    """
+    return f"pipeline object {obj.__name__ if isinstance(obj, type) else type(obj).__name__}"
+
+
+def _constructed_if_needed(candidate: Any, surface: Any, subject: str) -> tuple[Any, Any, list[str]]:
+    """Construct a class / zero-arg factory target, or report why it could not be.
+
+    Shared by both binding forms rather than restated in one of them. An
+    unconstructed class passes the ``generate`` probe as a plain function and
+    then receives the video as its ``self``, so the rule has to hold whether
+    the class reached the seam as an object or as a dotted import path - the
+    same value, bound two ways, has to resolve one way.
+
+    Args:
+        candidate: The resolved binding target.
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

This PR closes a binding-form asymmetry in `CosmosTransferTransform`: the dotted-import-path branch constructed a class / zero-arg-factory target before probing `generate()`, while the object branch did not, so `pipeline=Adapter` (one character short of the documented `pipeline=Adapter()`) passed `validate()` and then blew up inside generation with a bare `TypeError` handing the video to `self`. The fix extracts the construction step into a single owner (`_constructed_if_needed`) that both branches consult, so a class binds identically whichever form carries it, and a constructor that cannot be satisfied zero-arg is now refused at `validate()` naming the constructor rather than a missing `video` argument.

## What's good

- The over-reach controls (`TestNoPreviouslyUsablePipelineIsRefused`) pin instance identity (`_pipeline is pipeline`), staticmethod/classmethod-shaped classes, and the `object()` wording, so the shared construction rule cannot silently widen into refusing shapes that worked on `main`.
- The structural tests pin single ownership of the construction (`owners == {"_constructed_if_needed"}`) and that `_pipeline_problems` consults it once per binding form, matching the repo's one-owner convention; the probe-handler universe is now derived from the module's AST rather than a hardcoded pair, so the new owner's `except Exception` handler is graded on arrival.
- Error-path shape follows the transforms contract: every failure mode returns a problem string carrying `_PIPELINE_HELP` (ASCII, actionable), never a raise, and the `TypeError` / broad-`Exception` split preserves the accurate-subject reporting the dotted branch already had.
- Changelog fragment present per AGENTS.md step 3; docstring for `pipeline:` updated to state that a class or factory is constructed once during `validate()` and that the constructed instance is what `transform_frames` / `transform_version` read.

PR #2805, opened 2026-08-26
docs(agents): time the changelog fragment against the review, not the push

cagataycali:
## What

Amends `AGENTS.md` PR-workflow step 3 with the rule that decides whether the
changelog fragment's push costs a re-approval round: **when** it lands relative
to the first review, not whether it is a fragment.

## Why

Step 3 already explains why fragments exist -- appending to `CHANGELOG.md`
conflicts on one anchor, and clearing that conflict costs a re-approval round.
It is silent on the push the fragment convention itself introduces.

The fragment name is `changelog.d/<number>-<slug>.md`, and `<number>` is the
pull request's. That number does not exist until the pull request is open, so
for a change claiming no issue -- 249 of the last 300 here, per step 1 -- the
fragment can only arrive as a *second* commit onto an already-open branch.
`dismiss_stale_reviews_on_push` does not exempt it.

So the push is free while no approval has landed and costs a round once one has.
Four pull requests, one hour, same author, same fragment-as-second-commit shape:

| pull request | opened | fragment pushed | approval | outcome |
|---|---|---|---|---|
| #2800 | 21:16:19 | 21:16:56 (+37s) | 21:20:23 | survived |
| #2802 | 21:47:47 | 21:48:54 (+67s) | 21:56:13 | survived |
| #2801 | 21:22:49 | 21:55:53 (+33m) | 21:27:54 | dismissed |
| #2799 | 20:46:12 | 21:47:22 (+61m) | 20:53:27, 21:31:17 | dismissed twice |

The ordering is the only variable that moves the outcome. #2799 is the cost
stated plainly: three approval rounds on a one-line documentation link fix,
none of them reviewing anything a reviewer had asked about.

## The remedy recorded

- Push the fragment immediately after opening, ahead of any review; or
- open the pull request as a **draft**, add the fragment, then mark it ready --
  which removes the race instead of winning it; or
- when the change claims an issue, use the **issue** number, which is known at
  intake, so the fragment ships inside the first push and there is no second
  push to time.

It also records that `--amend --force-with-lease` is the wrong way to fold the
fragment in once review has started: on #2799 the two force-pushes are precisely
what dismissed the two approvals.

## Scope

Documentation only -- one paragraph block added to step 3, no code, no test, no
behaviour change. This is a process rule in the same register as the rest of
step 3, and the durability argument is AGENTS.md's own closing line for step 8:
a decision recorded only in a pull request comment is not durable.

This pull request was opened as a draft and its fragment pushed before it was
[...]

PR #2809, opened 2026-08-27
fix(teleop): report whether stop_teleoperate actually stopped the loop

cagataycali:
## Problem

`threading.Thread.join(timeout=...)` returns `None` whether or not the thread
finished, so the liveness read after it is the only thing that distinguishes a
stopped loop from one that outlasted the budget. `TeleopMixin.stop_teleoperate`
joined with a 3 s budget and never took that read. It then set the thread handle
to `None`, disconnected every attached device, and answered
`_teleop_stats(blocking=False)`.

A leader whose `get_action()` is still blocking - a serial read on a wedged bus
is the ordinary case a join budget exists for - was therefore reported as
stopped while its loop was still polling that leader and writing to the
follower, and the handle that was the only route to discovering it had already
been discarded.

Measured on a session whose leader blocks inside `get_action()` (a stand-in
device; `attach_teleop` needs only a callable `get_action`):

| reading | before | after |
|---|---|---|
| `stop_teleoperate` status | `success` | `error` |
| payload `stopped` | absent | `False` |
| wall time in the call | 3.00 s | 3.00 s |
| loop thread still alive | `True` | `True` |
| thread handle after the call | `None` | `<Thread ...>` |
| device `disconnect()` calls | 1 | 0 |
| `get_teleoperate_status` `thread_alive` | absent | `True` |
| frame delivered to the follower after the call returned | 1 | 1 |

The last row is the consequence the success claim hid: the follower is written
to after the caller has been told teleoperation stopped.

## Why the session counters cannot express this

`_teleop_stats` derives its status from the counters rather than hardcoding
`"success"`, and its comment says why - "so a dead teleop is not reported as
healthy". It distinguishes three signatures: a softly failing follower
(`errors == frames`), a dead leader (`frames == 0`), and a frame the slew bound
refused. A session that ran cleanly until its leader wedged carries healthy
counters, so "the loop is still running" is outside that vocabulary. The join
outcome is now reported beside them rather than through them.

`G1Driver.stop_task` already documents the shape this takes: "a caller-supplied
policy that outlasts the join budget ... surfaces as `status="error"` naming the
timeout and `stopped=False` in the payload, so the caller cannot read 'success'
while the payload's own `running=True` says the loop is still writing frames."

## Change

Read the join outcome, and let two things follow from it.

* The devices are disconnected only once the loop has joined. Tearing the bus
  down under a live writer is what `G1Driver.cleanup` refuses for the same
  reason ("Closing `_pubs` under a live 500 Hz thread would drop the loop into
  its `publish` branch and skip the zero-torque frame").
* The handle is cleared only on a real join, so a second `stop_teleoperate()`
  re-joins that loop instead of taking the idle branch - and
  `get_teleoperate_status` gained `thread_alive` beside `running`. `running` is
  the session flag, which the stop clears before it joins, so without the second
  reading a caller polling after a stop that reported `stopped=False` would be
  told `running=False` by the status verb.

The 3 s budget became `_TELEOP_JOIN_TIMEOUT_S` so the docstring, the refusal text
and the tests read one value. `stopped` is present on both paths, so a caller
reading it never has to tell an absent key from a false one.

## Why nothing caught it

Every `FakeTeleop.get_action()` in the shipped suite returns immediately, so
every join in it succeeds. The nearest existing cell,
`test_teleoperate_block_keyboard_interrupt_tears_down`, grades the `block=True`
path, which runs the loop inline and has no thread to join at all. The four
cells in the suite's own "session-end status honesty" section grade the three
counter signatures above.

## Scope

Seven timed thread joins exist in the package. Three already read liveness
(`G1ControlLoop.stop`, `HardwareRosBridge.shutdown`,
`HardwareRtpsBridge.shutdown`); four did not. The other three are deliberately
untouched because they answer different contracts: `PolicyServer.stop` returns
`None` and has no envelope to lie in, `InputPublisher.stop` returns a documented
`stats` property, and `stop_cameras_recording`'s window is a single render
because its loop re-reads `running` per camera.

## Verification

Pre-fix split (source reverted, tests kept): **11 failed / 6 passed**, with
0 failures in `TestThePremise` (2 cells) and 0 in `TestAJoinedStopIsUnchanged`
(4 cells). The headline failure is `assert 'success' == 'error'`.

Mutation table - 13 rows, control clean, `collected` stable at 17/18 throughout
so no row hides a deselection. `sib` counts failures in the seven pre-existing
teleop and hardware-lifecycle suites (119 files, 256 cells in the five that
matter here).

|---|---|---|---|
| b0 | 0 | 0 | control |
| b1 | 10 | 0 | liveness never read (the defect) |
| b2 | 3 | 0 | handle cleared unconditionally |
| b3 | 7 | 0 | devices disconnected regardless of the join |
| b4 | 4 | 0 | unjoined stop still reports `success` |
| b5 | 1 | 0 | `stopped` dropped from the unjoined payload |
| b6 | 1 | 0 | refusal text drops the budget |
| b7 | 1 | 0 | refusal text names neither leader nor follower |
| b8 | 1 | 0 | `stopped` never added to `_teleop_stats` |
| b9 | 1 | 0 | `thread_alive` dropped from the status verb |
| b10 | 4 | **2** | liveness read *before* the join |
| b11 | 4 | 0 | alternative disposition: `degraded` instead of `error` |
| b12 | 7 | 0 | tempting wrong fix: warn but keep the success envelope |

Ten distinct firing sets. Two collisions, both structural: b3 == b12 (each
routes an unjoined stop down the joined path) and b4 == b11 (each moves the
status off `error`). b10 is the informative row - reading liveness before the
join fires the three healthy-session controls plus the structural cell, and
**trips 2 pre-existing tests**, so the ordering is load-bearing at runtime and
not only in the AST assertion. b12 is the reviewer's first instinct measured:
warning while keeping the success envelope leaves 7 cells failing.

Gate, on an identical 529-file selection (every test naming teleop, a thread
join, or the hardware wrapper, plus every guard that walks the tree, parses
source, or reads docstrings and `docs/`), the new test file excluded from both
trees and every path verified present on both:

```
                collected   result                                        rootdir
branch          24919       23 failed, 24772 passed, 100 skipped, 24 err  robots-src
base (636abbb2) 24919       23 failed, 24772 passed, 100 skipped, 24 err  run-.../base
```

47 failing/erroring ids each, `comm` empty in both directions. Classified rather
than assumed: 44 are `tests/mesh/test_iot_*` needing `awsiot` (`find_spec` is
`False` here; declared in the `mesh-iot` extra, which CI installs), 3 are the
lerobot-from-source `TypeError: encode() takes 1 positional argument` drift.

tests_integ` (1616 files). mypy **24 == 24** against a pristine worktree of the
same base, normalized message sets byte-identical in both directions, 0
attributable. The changed file plus the four teleop suites: **256 passed** on the
rebased tree, and both load-bearing mutation rows re-measured unchanged there
(b1 10, b3 7).

No visual artifact: this is a teardown-reporting change, not a policy,
simulation, rendering, recording or asset one. The measured before/after table
above is the evidence.

## §13 Round changelog

**Round 2** — the lint finding the required check named, and a process error worth
naming beside it. One push, and it was the wrong shape.

- **The mypy finding is fixed.** `TestThePremise`'s first cell asserted
  `thread.join(timeout=0.05) is None`. `Thread.join` is annotated `-> None`, so
  mypy refuses to read a value from it at all:
  `error: "join" of "Thread" does not return a value (it only ever returns None)
  [func-returns-value]`. The assertion is now a bare call, and the cell keeps the
  half a test can actually add — `assert thread.is_alive() is True`, that the call
  returns normally while the thread runs on. The docstring says why the other half
  is the type checker's to state rather than the suite's, so the cell is not left
  looking like it dropped a claim.

  Confirmed load-bearing rather than assumed: reverted the line to the asserted
  form, ran mypy, got the finding back verbatim at that line; restored, `Success:
  no issues found in 1 source file`. Source md5-identical before and after. This
  was the right fix rather than a `type: ignore` — the annotation is correct and
  the assertion was the thing making an unsupportable claim, so silencing the
  checker would have preserved the defect in the test's reasoning.

- **The push that carried it should not have been a force-push.** The fragment
  `changelog.d/2809-...md` was folded into the first commit and the branch rebased,
  which force-pushed `9fb0ccc` → `07306f2` and dismissed the review that had landed
  twelve minutes earlier. AGENTS.md step 3 forbids precisely this: "Do not reach for
  `--amend --force-with-lease` to fold the fragment into the first commit once review
  has started: that destroys the review trail." The timings put this PR in that
  section's own dismissed rows — opened 00:30:35, review 00:42:51, fragment 00:54:41
  (+24m) — alongside #2801 (+33m) and #2799 (+61m), rather than in the surviving rows
  that pushed the fragment inside two minutes. The rule landed in this branch's own
  base (#2805, `0171573`). Nothing in the tree needs changing for it; the cost is one
  re-approval round on unchanged behaviour, and it is already spent. The remedy for
  the next one is the section's own: push the fragment immediately after opening, or
  open as a draft and mark ready once it is in.

- **So that the re-review can be the delta rather than the diff**, measured by blob
  identity between the reviewed tree and the current head:

  | file | reviewed `9fb0ccc` → head `07306f2` |
  |---|---|
  | `strands_robots/teleop_mixin.py` | **identical** (same blob) |
  | `docs/hardware/teleoperation.md` | **identical** (same blob) |
  | `tests/test_stop_teleoperate_reports_the_join_outcome.py` | +8 / -3, the cell above |
  | `changelog.d/2809-...md` | added (the fold that cost the round) |

  A two-dot diff between those commits shows 15 files. The other 13 are main's
  #2808, #2806 and #2807 arriving through the rebase, not this branch's work. The
  behaviour that was reviewed is byte-for-byte the behaviour that is here.

**Round 1** — initial fix, 18-cell suite, mutation table.

### DIFF
--- a/changelog.d/2809-stop-teleoperate-reports-the-join-outcome.md
+++ b/changelog.d/2809-stop-teleoperate-reports-the-join-outcome.md
@@ -1,0 +1,54 @@
+### Fixed: `stop_teleoperate` reports whether the teleop loop actually stopped
+
+`threading.Thread.join(timeout=...)` returns `None` whether or not the thread
+finished, so the liveness read after it is the only thing that distinguishes a
+stopped loop from one that outlasted the budget. `TeleopMixin.stop_teleoperate`
+joined with a 3 s budget and never took that read: it then set the thread handle
+to `None`, disconnected every attached device, and answered
+`_teleop_stats(blocking=False)`. A leader whose `get_action()` was still blocking
+- a serial read on a wedged bus is the ordinary case a join budget exists for -
+was therefore reported as stopped while its loop was still polling that leader
+and writing to the follower, and the handle that was the only route to
+discovering it had already been discarded. Measured on a session whose leader
+blocks inside `get_action()`:
+
+```
+                                     before          after
+stop_teleoperate status              success         error
+payload stopped                      (absent)        False
+loop thread still alive              True            True
+thread handle after the call         None            <Thread ...>
+device disconnect() calls            1               0
+get_teleoperate_status thread_alive  (absent)        True
+frame delivered after the call       1               1
+```
+
+The last row is the consequence the success claim hid: the follower is written to
+after the caller has been told teleoperation stopped.
+
+The session counters cannot express this. `_teleop_stats` derives its status from
+them precisely so a dead teleop is not reported as healthy, and it distinguishes
+three signatures - a softly failing follower, a dead leader, and a frame the slew
+bound refused - but a session that ran cleanly until its leader wedged carries
+healthy counters, so "the loop is still running" is outside that vocabulary. The
+join outcome is now reported beside them rather than through them, in the shape
+`G1Driver.stop_task` already documents: `status="error"` naming the budget with
+`stopped=False` in the payload, so a caller cannot read "success" while the loop
+still holds the wire.
+
+Two consequences follow from reading the outcome. The devices are disconnected
+only once the loop has joined, because tearing the bus down under a live writer
+is what `G1Driver.cleanup` refuses for the same reason; and the handle is cleared
+only on a real join, so a second `stop_teleoperate()` re-joins that loop instead
+of reporting an idle session. `get_teleoperate_status` gained `thread_alive`
+alongside `running`: `running` is the session flag, which the stop clears before
+it joins, and without the second reading a caller polling after a stop that
+reported `stopped=False` would be told `running=False` by the status verb.
+
+A joined stop is unchanged - it still reports the counter-derived status, still
+disconnects its devices, still clears the handle, and stopping an idle host is
+still a no-op success. `stopped` is now present on both paths so a caller reading
+it never has to tell an absent key from a false one. Of the seven timed thread
+joins in the package, three already read liveness; the other three answer
+different contracts (a `-> None` server teardown, a stats property, and a
+recording flush whose window is one frame) and are deliberately untouched.
--- a/docs/hardware/teleoperation.md
+++ b/docs/hardware/teleoperation.md
@@ -87,7 +87,7 @@ Every hardware `Robot` and `Simulation` host exposes:
 | `attach_teleop(device_or_spec, *, name=None, method=None, map_fn=None, **kwargs)` | Register an input stream (lazy - no hardware touched). `device_or_spec` is a built teleop instance **or** a type string built via `Teleoperator(**kwargs)`. |
 | `teleoperate(*, names=None, robot_name=None, hz=50.0, publish=False, block=False, duration=None)` | Run the control loop. |
 | `detach_teleop(name=None)` | Remove one (or all) attached streams. |
-| `stop_teleoperate()` | Stop the loop, any mesh publishers, and disconnect devices. |
+| `stop_teleoperate()` | Stop the loop, any mesh publishers, and disconnect devices. Reports `status="error"` with `stopped: false` when the loop outlasts its 3 s join budget - the devices are left connected rather than torn down mid-write, and a second call re-joins the same loop. |
 
 ### `attach_teleop`
 
@@ -307,6 +307,8 @@ robot.teleoperate(block=True, duration=60)   # 60 s then stop + disconnect
 robot.teleoperate()
 ...
 robot.stop_teleoperate()                     # stop loop + publishers + disconnect
+#   -> status="error" + stopped=false if the loop is still polling the leader;
+#      get_teleoperate_status()["thread_alive"] reads the loop thread itself.
 ```
 
 ## How it relates to mesh teleop
--- a/strands_robots/teleop_mixin.py
+++ b/strands_robots/teleop_mixin.py
@@ -53,6 +53,9 @@ logger = logging.getLogger(__name__)
 
 # Default local control loop frequency (Hz). Matches InputPublisher.
 _TELEOP_HZ_DEFAULT = 50.0
+# Budget for joining the background teleop thread in stop_teleoperate.
+# Named so the docstring, the refusal text and the tests read one value.
+_TELEOP_JOIN_TIMEOUT_S = 3.0
 
 ActionDict = dict[str, float]
 MapFn = Callable[[ActionDict], ActionDict]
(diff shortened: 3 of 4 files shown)


yinsong1986 (dismissed):
## Summary

This PR makes `stop_teleoperate()` report the join outcome instead of claiming success unconditionally: `Thread.join(timeout=)` returns `None` either way, so the new `is_alive()` read after the join is what distinguishes a stopped loop from one that outlasted the 3 s budget. An unjoined stop now returns `status="error"` with `stopped=false`, leaves the devices connected rather than tearing the bus down under a live writer, and keeps the thread handle so a second call re-joins the same loop and `get_teleoperate_status()` can report `thread_alive`. The healthy path is unchanged (pinned by explicit over-reach controls in the new test module). Note the required `call-test-lint` check is currently red on a mypy `func-returns-value` finding at the test's premise assertion (`thread.join(timeout=0.05) is None`, tests/test_stop_teleoperate_reports_the_join_outcome.py:147) - a one-line lint fix CI already names, not a behavioral concern; branch protection holds the merge until it clears.

## What's good

- The 18 new tests all pass and grade both halves independently (the envelope and the handle), with controls pinning that a healthy session's behaviour is byte-for-byte what it was.
- The join budget is a named module constant read at call time, so the docstring, the refusal text and the tests read one value; exactly one cell drives the shipped 3 s.
- Error envelope follows the tool-result contract (`status`/`content`, ASCII-only, names the budget, the devices still polled and the target still written) and the remedy is actionable: call `stop_teleoperate()` again.
- The handle is cleared only on a real join, closing the prior state where a wedged loop became undiscoverable to every later call and status read.

yinsong1986 (approved):
## Summary

This PR closes the honesty gap in `stop_teleoperate()`: `Thread.join(timeout=)` returns `None` either way, so the new `is_alive()` read after the join is the only thing that distinguishes a stopped loop from one that outlasted the 3 s budget. An unjoined stop (the wedged-`get_action()` case the timeout exists for) now returns `status="error"` with `stopped=false`, leaves the devices connected rather than tearing the bus down under a live writer, and keeps the thread handle so a second call re-joins the same loop; `get_teleoperate_status` gained `thread_alive` beside the session flag so a poller in that window is not told `running=False` while the follower can still be written to. The joined path is unchanged and pinned by explicit over-reach controls, `stopped` is present on both paths, and the budget is a named module constant read by the docstring, the refusal text and the tests alike.

## What's good

- The prior review's one finding (mypy `func-returns-value` at the premise test) is resolved at this head: the test no longer reads `join()`'s return value and asserts `is_alive()` instead.
- The 18-cell test module grades both halves independently (the envelope and the handle), pins the liveness read structurally with the join-before-is_alive ordering, and drives the shipped 3 s constant in exactly one cell so the suite stays fast.
- The error envelope follows the tool-result contract (error dict, ASCII-only, names the budget, the devices still polled and the target still written) and the remedy is actionable: call `stop_teleoperate()` again.
- Changelog fragment per the `changelog.d/` convention, docs table and example updated in the same PR, behaviour-named test classes, no host paths.
- Scope discipline: of the seven timed joins in the package, the three untouched ones are named and the reason each answers a different contract is stated rather than left to be inferred.

PR #2810, opened 2026-08-27
fix(protomotions): hold history_length to the whole-number domain before it is normalised

cagataycali:
## Problem

`ProtoMotionsPolicy.__init__` takes `history_length` as the number of past action frames that feed the `historical_processed_actions` ONNX input, guarded it with a bare `history_length < 1` test, and then normalised it with `int(history_length)` to shape the rolling buffer. That test covers the floor and not the domain, so the spellings a config carries were laundered into a different window than the one they name.

`history_length` is one of the provider's advertised `config_keys`, so it arrives from a JSON or YAML policy config as readily as from a keyword. Measured through `create_policy("protomotions", ...)`, reading the shape off the session the policy drives:

| policy config | construction | window the tracker read |
|---|---|---|
| `{"history_length": 3}` | built | `(1, 3, 29)` |
| `{"history_length": 3.0}` | built | `(1, 3, 29)` |
| `{"history_length": 2.7}` | built | **`(1, 2, 29)`** - a two-frame window |
| `{"history_length": true}` | built | **`(1, 1, 29)`** - a one-frame window |

Both wrong rows report a successfully built policy, and the tracker then reads a shorter history than the caller asked for on every tick.

## Why this ordering, and not a judgement call

`ProtoMotionsConfig.__post_init__` in the same package already resolves this question the other way for its own body indices, and states why in its own docstring:

> Each index goes through the shared whole-number domain BEFORE the `int()` normalisation, not after: that conversion is what laundered a yaml `anchor_body_index: true` into row 1 (`head`) and a `2.7` into row 2 (`left_hip_pitch_link`).

`load_config_from_yaml` hands both indices through raw for the same stated reason. The two values this PR is about are literally the two that comment names. The constructor now follows the same order, through `positive_whole_number_error` - the same whole-number family the sibling uses (`non_negative_whole_number_error`), with the positive floor a buffer dimension needs.

`positive_whole_number_error` is the right member rather than `positive_count_error` because a window length read from a config is legitimately an integral float; that domain's own contract names the case ("a `30.0` computed from a config float"). Its magnitude arm is a bonus: a count past any allocatable buffer is now refused by the parameter's own domain instead of reaching NumPy, which refuses it with a message naming neither the parameter nor the caller.

## The window the tracker is fed

history window per config spelling

Each dashed outline is one frame of 29 joint targets; solid means allocated and fed to the tracker. Both columns come from one capture script run against the two trees, driving the shipped `create_policy` path and reading the shape off the recording session. Every number in the figure is asserted from the two captures before anything is drawn. The two spellings that name three frames are byte-identical across the columns - `(1, 3, 29)` either way.

## Nothing accepted before is refused now

`3`, `3.0`, `np.int64(4)` and `np.float32(4.0)` all still build and produce a byte-identical buffer and an identical ONNX feed shape; `0`, a negative count and `False` are still refused. Those cells sit in their own class and pass on both trees, which is what the two over-reach mutations below are measured against.

## Why nothing caught it

The pre-existing cell is `test_construction_history_length_positive`, and it drives `history_length=0` - the floor. Running all eight mutations against the twelve sibling suites in `tests/policies/protomotions/` shows exactly that split: the three that break the floor are caught by that one cell, and the five that break the domain are invisible.

## Tests

New `tests/policies/protomotions/test_history_length_sizes_the_window_it_names.py`, 45 cells in four classes. Pre-fix, with the source reverted and the test kept: **27 failed / 18 passed**, and the failures are confined to the two classes that should carry them.

| class | role | failed pre-fix | cells |
|---|---|---|---|
| `TestThePremises` | premise | 0 | 4 |
| `TestAWindowLengthThatCannotBeHonoredIsRefused` | regression | 24 | 8 |
| `TestTheGuardPrecedesTheNormalisation` | structural | 3 | 3 |
| `TestTheAcceptedWindowLengthsAreUnchanged` | over-reach control | 0 | 4 |

Mutation table - `mine` is failures in the new file, `sib` in the twelve sibling suites, `coll` the collected count so a silent deselection is visible:

| | mutation | mine | sib | coll |
|---|---|---|---|---|
| b0 | control - no mutation | 0 | 0 | 45 |
| b1 | guard reverted to the bare `< 1` comparison | 26 | 0 | 45 |
| b2 | guard moved AFTER the `int()` normalisation | 9 | 0 | 45 |
| b3 | domain swapped to `positive_count_error` (strict int) | 15 | 0 | 45 |
| b4 | domain swapped to the `non_negative` sibling (floor dropped) | 9 | 1 | 45 |
| b5 | domain re-implemented locally as `isinstance(int) and >= 1` | 15 | 0 | 45 |
| b6 | reason computed but never raised | 27 | 1 | 45 |
| b7 | the wrong fix: `round()` the value instead of refusing it | 28 | 1 | 45 |
| b8 | an extra local strict-int check beside the shared guard | 7 | 0 | 45 |

Eight mutations, eight distinct firing sets, control clean, source restored byte-identically. b2 is what makes the ordering load-bearing rather than stylistic; b3 and b8 are the over-reach directions; b7 is the tempting alternative fix, and it fires the widest set because a coerced value is the defect with a different spelling.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1616 files).
- `mypy strands_robots tests tests_integ`: **24 errors on this branch, 24 on the base**, normalized message sets byte-identical in both directions, 0 attributable to either changed file.
- Paired suite, `MUJOCO_GL=egl`, identical 677-file selection (all of `tests/policies/`, every guard that walks the tree or reads source, docstrings, `Args:`/`Raises:` or `changelog.d/`, and everything naming `protomotions` / `positive_whole_number_error` / `create_policy`), the new file excluded from both trees and every path verified present on both: identical **28006 collected** and `20 failed, 27883 passed, 103 skipped` on each, **20 identical failing ids with `comm` empty in both directions**, distinct rootdirs. All 20 classified by measurement: 17 need `awsiot` (absent locally, declared in the `[mesh-iot]` extra), 3 are the lerobot-from-source `TypeError: encode() takes 1 positional argument` drift.
- No coverage claim: `policy.py` reports **191 statements / 19 missing / 90% on both trees**. The guard replaces two statements with two, so the uncovered set only shifts by the seven lines the comment adds. The value here is the pin, not a line count.

## Scope

Left alone deliberately: `providers` is the constructor's other advertised `config_keys` entry with no domain, but it is a list of execution-provider names rather than a number, its consumer is `onnxruntime`, and what that library does with a malformed entry is not measurable without it installed. That is a separate question with a separate answer.

### DIFF
--- a/changelog.d/2810-protomotions-history-length-domain.md
+++ b/changelog.d/2810-protomotions-history-length-domain.md
@@ -1,0 +1,37 @@
+### Fixed: the ProtoMotions history window is held to a whole-number domain before it is normalised
+
+`ProtoMotionsPolicy.__init__` takes `history_length` as the number of past action
+frames that feed the `historical_processed_actions` ONNX input, guarded it with a
+bare `history_length < 1` test, and then normalised it with
+`int(history_length)` to shape the rolling buffer. That test covers the floor and
+not the domain, so the spellings a config carries were laundered into a different
+window than the one they name. `history_length` is one of the provider's
+advertised `config_keys`, so it arrives from a JSON or YAML policy config as
+readily as from a keyword. Measured through `create_policy("protomotions", ...)`,
+reading the shape off the session the policy drives:
+
+```
+policy config              construction   window the tracker read
+{"history_length": 3}      built          (1, 3, 29)
+{"history_length": 3.0}    built          (1, 3, 29)
+{"history_length": 2.7}    built          (1, 2, 29)   <- a two-frame window
+{"history_length": true}   built          (1, 1, 29)   <- a one-frame window
+```
+
+`ProtoMotionsConfig.__post_init__` in the same package already resolves this
+ordering the other way for its own body indices, and states why: each index goes
+through the shared whole-number domain *before* the `int()` normalisation,
+because that conversion is what laundered a yaml `anchor_body_index: true` into
+row 1 and a `2.7` into row 2. `load_config_from_yaml` hands both indices through
+raw for the same reason. The constructor now follows that order too, through
+`positive_whole_number_error` - the same whole-number family the sibling uses,
+with the positive floor a buffer dimension needs.
+
+Every spelling the previous code honored still is, because a window length read
+from a config is legitimately an integral float: `3`, `3.0`, `np.int64(4)` and
+`np.float32(4.0)` all build and produce a byte-identical buffer, and `0`, a
+negative count and `False` are still refused. What changes is that a value that
+cannot be honored as a count is now named rather than coerced, and that a count
+past any allocatable buffer is refused by the parameter's own domain instead of
+reaching NumPy, which refuses it with a message naming neither the parameter nor
+the caller.
--- a/strands_robots/policies/protomotions/policy.py
+++ b/strands_robots/policies/protomotions/policy.py
@@ -42,7 +42,7 @@ import numpy as np
 from numpy.typing import NDArray
 
 from strands_robots.policies.base import Policy
-from strands_robots.utils import require_optional
+from strands_robots.utils import positive_whole_number_error, require_optional
 
 from .config import (
     ProtoMotionsConfig,
@@ -112,7 +112,9 @@ class ProtoMotionsPolicy(Policy):
         history_length: Length of the historical-actions rolling buffer that
             feeds the ``historical_processed_actions`` ONNX input. The upstream
             checkpoint uses ``1`` (single previous action). Larger values pad
-            with zeros until the buffer fills.
+            with zeros until the buffer fills. A positive whole number: the
+            value is a buffer dimension, so a fractional or boolean spelling
+            cannot be honored as the window length it reads as.
     """
 
     #: Non-VLA - the tracker is proprioceptive, no cameras.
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Replaces the bare `history_length < 1` floor check in `ProtoMotionsPolicy.__init__` with the shared `positive_whole_number_error` domain guard, run before the `int()` normalisation that previously laundered a config's `2.7` into a two-frame window and `true` into a one-frame one. This matches the ordering `ProtoMotionsConfig.__post_init__` already uses for its own body indices (and documents for exactly this reason), and reuses an existing shared guard with multiple call sites rather than adding a local one, per AGENTS.md convention #11. The newly refused spellings raise an explicit `ValueError` naming the parameter and the class, so nothing is silently invalidated; the accepted set (`int`, integral float, NumPy integer) is pinned unchanged by over-reach control tests.

## What's good

- Verified locally at head `da269a68`: the new 45-cell suite and all 265 `tests/policies/protomotions/` tests pass; `ruff check` / `ruff format --check` clean on both changed source files.
- The repo's own guard suites (`Raises:`/`Args:` docstring completeness, changelog format, ASCII-only tool/log strings, behaviour-named test cases, no host paths) all pass over the new files.
- Changelog fragment present (`changelog.d/2810-protomotions-history-length-domain.md`), no direct `CHANGELOG.md` append, per AGENTS.md step 3.
- Structural tests pin the guard-before-normalisation ordering and that the domain comes from `strands_robots.utils`, so a drift back to a local comparison fails rather than passing silently.
- Scope discipline: the `providers` config key is deliberately left alone with the reasoning stated in the description.

PR #2811, opened 2026-08-27
fix(g1): release a DDS endpoint a concurrent teardown left unreachable

cagataycali:
## The defect

`DDSSubscriberSet.subscribe` and `DDSPublisher.get_publisher` build their SDK
endpoint under the shared construction lane `_DDS_INIT_LOCK` — the CycloneDDS
bindings segfault on concurrent construction, so that lock is held for the whole
build. `close` swaps its collection out under the *instance* `_lock` and releases
what it took.

Those two locks do not exclude each other. A teardown can therefore complete
between the moment a construction finishes and the moment it records what it
built, and the recording then lands in a fresh list or dict that `close` has
already walked past.

Driven deterministically (the stand-in's `Init` parks after signalling, which
holds the caller inside `_DDS_INIT_LOCK` with a fully built endpoint, and the
teardown then runs to completion):

| after `close()` returned | before | after |
|---|---|---|
| `subscribe()` reported | `None` (success) | a reason naming the topic |
| `Close()` calls on the built subscriber | **0** | **1** |
| subscriber is in `_subs` | **True** | False |
| `len(_subs)` | **1** | 0 |
| `get_publisher()` reported | `None` (success) | a reason naming the topic |
| `Close()` calls on the built publisher | **0** | **1** |
| publisher is in `_pubs` | **True** | False |

## Why the orphan is permanent, and live

Nothing reaches it afterwards. `G1Driver.cleanup` (`drivers/g1.py:525-530`) drops
the set immediately after closing it, so the second `close` that would collect
the orphan can never happen. A second `close` on the same object *does* release
it — measured — which is exactly why dropping the object instead is what makes
this stick.

And the endpoint left behind is a live one, not merely a forgotten reference.
`subscribe` asks for `queueLen=10`; the engine's own `close` docstring spells out
what that means:

> at any queue length above zero `unitree_sdk2py` starts a `ch_reader` daemon
> thread whose target is a bound method of the channel's reader. That running
> thread keeps the whole channel reachable, so the CycloneDDS `__del__` that
> would release the `DataReader` never runs: the thread keeps looping, the reader
> stays matched, and the decoder callback keeps filling caches for a driver that
> believes it is disconnected.

Both entry points are reachable from public driver methods — `connect_eagerly`
subscribes (`g1.py:447`), `cleanup` closes. On the write path there is a
single-threaded route too: `cleanup` calls `loop.stop(...)` and does not read its
outcome, so a control loop whose join times out keeps calling `publish` →
`get_publisher`, which on a cache miss rebuilds a publisher into the emptied
cache that `cleanup` then drops.

## The fix

Each recording path captures the collection it was told to write to, and compares
it before recording. `close` rebinds `_subs`/`_pubs` to a fresh object, so that
identity *is* the teardown generation — no new state is needed. When it has
moved, the endpoint is released through `_release_partial` and the caller is told,
naming the topic.

`_release_partial` already existed to uphold exactly this invariant on the path
where construction *fails* — its docstring says such an endpoint "is unreachable
from `close` ... so releasing it here is the only opportunity". The same rule now
covers the path where construction succeeds and the *recording* is what could not
happen. Its docstring and both `close` docstrings are corrected to state the whole
rule rather than the one consequence they addressed.

Sequential behaviour is unchanged: a `subscribe` or `get_publisher` after a
`close` captures the collection `close` left behind, so it is current when it
records. A driver reconnecting is never refused for a teardown that already
finished.

## Why nothing caught it

The pre-existing suites are **completely silent**: with the source reverted to
`main`, `tests/tools/g1/` + `tests/drivers/` (**567 cells**, this file deselected)
reports **0 failures**. Both `close` docstrings reason about the lock split —
"a concurrent `subscribe`, which appends without that lock, cannot mutate the list
being walked" — which is true, and is the safe-iteration half only. Nothing
addressed where the appended subscriber ends up.

## Verification

Pre-fix split (source reverted, new file run): **9 failed / 12 passed**, and the
failures are confined to the two regression classes.

| class | role | pre-fix |
|---|---|---|
| `TestASubscribeRacingCloseLeavesNoLiveSubscriber` | regression | 5 / 5 |
| `TestAGetPublisherRacingCloseLeavesNoLivePublisher` | regression | 3 / 3 |
| `TestTheUnracedPathsAreUnchanged` | control | **0 / 5** |
| `TestThePremisesTheRaceRestsOn` | premise | **0 / 3** |
| `TestTheRuleHasOneShapeOnEveryEndpointKind` | structural | 1 / 4 |

Mutation table — 11 rows, control clean, `collected` stable at 21 throughout, source
restored md5-identical:

| row | mutation | fires |
|---|---|---|
| b0 | control, no mutation | 0 |
| b1 | subscriber records blind | 6 |
| b2 | publisher records blind | 4 |
| b3 | both record blind (= the code revert) | 9 |
| b4 | identity compared with `==` instead of `is` | 8 |
| b5 | orphan dropped instead of released | 2 |
| b6 | orphan released but reported as success | 2 |
| b7 | capture taken after construction, inside the lane | 0 |
| b8 | reason no longer names the topic | 1 |
| b9 | `close` clears in place instead of rebinding | 6 |
| bR | raw revert of the whole file | 9 |

Notes on three of those rows:

- **b1 + b2 = 9 = b3 = bR** with a single overlapping cell, so each endpoint kind
  is independently pinned.
- **b4** is the plausible slip: `[] == []` and `{} == {}` are both `True`, so
  comparing by value rather than identity reintroduces the race wholesale.
- **b7 fires 0, and that is correct rather than a gap.** The capture only has to
  precede the *construction*, not the lane; `close` never takes `_DDS_INIT_LOCK`,
  so moving the capture inside it changes nothing observable. The `_lock` around
  the capture makes the read obviously synchronised instead of resting on the GIL.
- **b4 and b9 fire 0 in the 567 pre-existing cells**, so neither is over-reach.

An adjacent guard needed repairing rather than renumbering.
`test_the_release_is_named_once_for_both_paths` asserted
`source.count("_release_partial(") == 3`, which two new release opportunities
break. Its docstring states the real property — "One helper, so the two paths
cannot drift apart" — so it now asserts that derived instead: every `Close()` in
the module is either the helper's own or a `close` method's. A count has to be
edited whenever a release opportunity is added, and editing it is
indistinguishable from noticing that a new one bypassed the helper.

## Gate

- `ruff check` + `ruff format --check` clean over 1618 files.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a worktree of
  `upstream/main`, normalized message sets byte-identical in both directions,
  **0 attributable** to the changed files.
- `tests/tools/g1/` + `tests/drivers/`: 543 → **564 passed, 3 skipped**.
- The new file runs in 0.17s with no sleeps: the interleaving is driven by
  barriers, and the stand-ins mean it grades the production path on a box with
  no SDK.

No visual artifact: this is a resource-lifecycle fix in a transport helper, not a
policy, simulation, rendering, recording or asset change. The measured
`Close()`-count table above is the evidence.

### DIFF
--- a/changelog.d/2811-dds-teardown-cannot-orphan-an-endpoint.md
+++ b/changelog.d/2811-dds-teardown-cannot-orphan-an-endpoint.md
@@ -1,0 +1,30 @@
+### Fixed: a DDS endpoint built across a teardown is released, not left live
+
+`DDSSubscriberSet.subscribe` and `DDSPublisher.get_publisher` build their SDK
+endpoint under the shared construction lane `_DDS_INIT_LOCK`, because the
+CycloneDDS bindings segfault on concurrent construction. `close` swaps its
+collection out under the instance `_lock` and releases what it took. Those two
+locks do not exclude each other, so a teardown could complete between the
+moment a construction finished and the moment it recorded what it built, and
+the recording then landed in a fresh list or dict that `close` had already
+walked past.
+
+Nothing reached the endpoint afterwards. `G1Driver.cleanup` drops the set right
+after closing it, so the second `close` that would have collected the orphan
+never happens - and `cleanup` does not read the outcome of the `loop.stop` that
+precedes it, so a control loop whose join times out keeps calling `publish`,
+rebuilding a publisher into the emptied cache on the way. What was left behind
+is the state the engine's own `close` docstring describes at length: `subscribe`
+asks for `queueLen=10`, and at any queue length above zero the SDK starts a
+`ch_reader` daemon thread that keeps the channel reachable, so the reader stays
+matched and the decoder callback keeps filling caches for a driver that believes
+it is disconnected.
+
+Each recording path now captures the collection it was told to write to and
+compares it before recording, releasing through `_release_partial` and naming
+the topic when a teardown moved it. `_release_partial` already existed to
+uphold this invariant where construction *fails*; the same rule now covers the
+path where construction succeeds and the recording is what could not happen.
+Sequential use is unchanged: a `subscribe` or `get_publisher` after a `close`
+captures the collection that `close` left behind, so a driver reconnecting is
+never refused for a teardown that already finished.
--- a/strands_robots/tools/g1/_dds_engine.py
+++ b/strands_robots/tools/g1/_dds_engine.py
@@ -31,7 +31,7 @@ logger = logging.getLogger(__name__)
 
 
 def _release_partial(endpoint: Any | None, what: str) -> None:
-    """Close an endpoint whose construction failed, rather than dropping it.
+    """Close an endpoint :meth:`close` can never reach, rather than dropping it.
 
     ``ChannelSubscriber.__init__`` and ``ChannelPublisher.__init__`` create a
     live ``Channel`` before ``Init`` runs, and ``Init`` is not atomic either:
@@ -45,7 +45,11 @@ def _release_partial(endpoint: Any | None, what: str) -> None:
 
     Such an endpoint is unreachable from :meth:`DDSSubscriberSet.close` and
     :meth:`DDSPublisher.close`, because the call that failed never recorded
-    it - so releasing it here is the only opportunity. ``Close()`` is safe on
+    it - so releasing it here is the only opportunity. A construction that
+    *succeeded* reaches the same state when a concurrent teardown swapped the
+    collection away before it could record: the endpoint would land in a fresh
+    list or dict that ``close`` has already walked past, so it is released here
+    too rather than recorded somewhere nothing will look. ``Close()`` is safe on
     a half-built endpoint: it routes to ``Channel.CloseReader`` /
     ``CloseWriter``, which skip an entity that was never created and still
     stop and join the reader thread.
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Closes a real, deterministic-reproducible race in the G1 DDS engine: `DDSSubscriberSet.subscribe` / `DDSPublisher.get_publisher` build under the shared construction lane `_DDS_INIT_LOCK` while `close` swaps its collection out under the instance `_lock`, so a teardown completing between construction and recording left a live endpoint (with a running `ch_reader` daemon thread on the subscriber side) in a collection nothing would ever walk. The fix captures the collection identity before construction and compares it under `_lock` before recording, releasing through the pre-existing `_release_partial` helper and naming the topic when the teardown won. Verified locally at head: 527 passed / 40 skipped across `tests/tools/g1/` + `tests/drivers/`; with the engine reverted to main the new file fails 9/12 exactly as the description claims, confined to the two regression classes.

## What's good

- The new lock nesting (`_DDS_INIT_LOCK` -> `_lock`) cannot invert against `start()`'s `_lock` -> `_DDS_INIT_LOCK`: the latter only occurs while `_started` is False and the recording paths are only reachable after it is True (set once, never reset), and `close()` acquires nothing while holding `_lock` - so no deadlock is introduced.
- Sequential reconnect is explicitly pinned as a control (`test_subscribing_after_a_close_still_succeeds`), so a driver reconnecting is never refused for a finished teardown.
- Replacing the `_release_partial` call-site count with a derived structural assertion (every `Close()` belongs to the helper or a `close` method) is a strict improvement - the count would have had to be edited for exactly the change it existed to catch.
- Barrier-driven interleaving, no sleeps, no SDK dependency; changelog fragment present per AGENTS.md step 3; ruff check/format clean on the changed files; no non-ASCII or host paths.

PR #2813, opened 2026-08-27
fix(isaac): refuse a non-finite geom instead of measuring a body around it

cagataycali:
## Problem

`_geom_aabb` folds each of a body's geoms into the axis-aligned box `load_mjcf_scene_objects` reports as a `SceneObject`'s `size` and `position` -- the collision proxy the Isaac realization stands the object up with. `_body_collision_aabb` unions those with a running `min`/`max`, and Python orders a NaN as neither smaller nor larger than anything, so a comparison against one keeps the accumulator it started with. The geom carrying it disappears from the bound, and nothing reports it.

A static fixture whose leg declares `euler="nan 0 0"` was measured as its tabletop alone:

| fixture | reported `size` |
| --- | --- |
| tabletop + healthy leg | `(0.8, 0.6, 0.77)` |
| tabletop + leg `euler="nan 0 0"` | `(0.8, 0.6, 0.04)` |
| tabletop, **no leg at all** | `(0.8, 0.6, 0.04)` |

A 4 cm slab where the file declares a 77 cm table -- **19.2x too short**, byte-identical to the same fixture with the leg deleted, under a successful load. Each spelling failed differently:

| spelling | MuJoCo | reported `position` | reported `size` |
| --- | --- | --- | --- |
| healthy | compiles | `(0.4, 0.1, 0.385)` | `(0.8, 0.6, 0.77)` |
| `pos="nan ..."` | compiles | `(0.4, 0.1, 0.385)` | `(0.8, 0.6, 0.77)` |
| `pos="inf ..."` | compiles | `(inf, 0.1, 0.385)` | `(nan, 0.6, 0.77)` |
| `size="inf ..."` | compiles | `(nan, 0.1, 0.75)` | `(0.001, 0.6, 0.04)` |
| `euler="nan 0 0"` | compiles | `(0.4, 0.1, 0.75)` | `(0.8, 0.6, 0.04)` |
| `quat="nan 0 0 0"` | compiles | `(0.4, 0.1, 0.75)` | `(0.8, 0.6, 0.04)` |

`pos="nan ..."` drops the geom on the axis that is not finite and keeps it on the others. `size="inf ..."` is inverted: `inf - inf` is a NaN that `_recursive_collision_aabb`'s own accumulator drops in turn, so that axis lands at the `1e-4` floor -- the narrowest proxy the reader can emit, for the widest geom the file can declare.

## Reachability

MuJoCo refuses a non-finite geom on a body with a **free joint**, because the inertia it derives is then not finite (`mass and inertia of moving bodies must be larger than mjMINVAL`). It *compiles* the same geom on a body **without** one -- and a body without a free joint is exactly what this loader calls a fixture, the tables and cabinets a manipulation scene is planned against. So the scenes that reach the reader are precisely the ones that load, which is why the reader cannot defer the question to a compile step.

## Why refuse

MuJoCo is the oracle for the disposition too. It refuses a non-finite geom quantity wherever it checks one -- `fromto="0 0 0 nan 0 0.5"` is rejected as `nan size in geom` -- and warns about the document as a whole (`XML contains a 'NaN'`). Measuring around it silently is not an answer either half of MuJoCo gives, and it is not what this module's stated failure semantics promise: *"Loaders never silently return a phantom robot."*

The same defect class was closed one module over for mesh assets in #2740, whose `_non_finite_vertex_error` describes this mechanism in the same words -- bounds "the same numbers a mesh declaring only those would produce, under no error at all". `loaders.py` carried no finiteness guard at all.

Propagating instead of refusing does not work: a NaN bound returned from `_geom_aabb` is dropped again by the outer accumulator in `_recursive_collision_aabb`, so the loss simply moves one frame up.

## Change

One guard, `_refuse_non_finite_geom`, reached from all four parse paths in `_geom_aabb` -- `pos`, `size`, whichever orientation spelling the geom used, and `fromto` -- so no one spelling drifts into tolerating what its siblings refuse. The refusal names both the offending attribute and the geom, falling back to the resolved `type` because MJCF geoms are frequently unnamed.

Only a value that **parsed** is graded, so an unreadable attribute keeps falling back to its documented default exactly as before. The locator is read from the `<default>`-resolved attributes rather than off the element, which is the rule every other reader in this module follows and which `test_every_geom_attribute_read_goes_through_the_resolver` enforces.

`load_mjcf_scene_objects`' public `Raises:` clause and `_body_collision_aabb`'s docstring are updated in the same change -- the latter now states why its running `min`/`max` is safe.

## Verification

Regression suite: `tests/simulation/isaac/test_non_finite_geom_is_refused_not_measured_around.py`, 43 cells.

| class | pre-fix | role |
| --- | --- | --- |
| `TestThePremisesTheFindingRestsOn` | 0 / 13 | premise -- MuJoCo compiles each fixture, refuses `fromto`, refuses a moving body |
| `TestANonFiniteGeomIsRefusedByName` | **21 / 21** | regression |
| `TestTheRefusalLocatesAnUnnamedGeom` | **2 / 2** | regression -- locator |
| `TestWhatIsDeliberatelyUnchanged` | 0 / 5 | control |
| `TestOneOwnerForTheWording` | 1 / 2 | structural -- one guard, reached from four paths |

Pre-fix split **24 failed / 19 passed**, with both control classes clean. The raw revert is a collection error (the suite imports the new helper), so the split above is the behavioural revert: the helper kept, its four call sites neutered.

Deliberately unchanged and pinned: a healthy fixture reports exactly what it did; `pos="garbage"` still falls back to the body origin rather than becoming a refusal; a mesh geom still returns `None` so the caller falls back to another geom; and an extreme-but-finite `1e30` extent is still a measurement -- the guard tests finiteness, not magnitude.

Corpus: **542 MJCF documents, 14940 `<geom>` elements** under `robot_descriptions` and this package's own assets declare **zero** non-finite placements or extents; re-measuring **2850 bodies** through `_body_collision_aabb` after the change raises **0** times. No shipped asset changes behaviour.

Gate: `ruff check` and `ruff format --check` clean over 1626 files; `mypy strands_robots tests tests_integ` reports `Success: no issues found in 1618 source files`; `tests/simulation/isaac/` plus the `Raises:`/`Args:` docstring graders **2254 passed, 5 skipped**; the 162-file tree-walking guard set **6802 passed, 56 skipped** with 5 pre-existing failures reproduced identically on a pristine `upstream/main` worktree (they assert `rclpy` is absent and it resolves on this machine).

## Rendered

The fixture as MuJoCo compiles it, beside the collision proxy the loader reports for it. The two proxy silhouettes are measured from separate segmentation passes, so transparency cannot affect the count; the fixture geometry is asserted byte-identical between the two panels and the lighting and camera are asserted equal.

collision proxy for a non-finite geom

Fixture pixels falling **outside** the reported proxy: **0** for the healthy leg, **3956** for `euler="nan 0 0"` -- the leg standing entirely outside the box that is supposed to contain it.

### DIFF
--- a/changelog.d/2813-non-finite-geom-is-refused-not-measured-around.md
+++ b/changelog.d/2813-non-finite-geom-is-refused-not-measured-around.md
@@ -1,0 +1,38 @@
+### Fixed: a geom declaring a placement or extent that is not finite is refused, not measured around
+
+`_geom_aabb` folds each of a body's geoms into the axis-aligned box
+`load_mjcf_scene_objects` reports as a `SceneObject`'s `size` and `position` - the collision proxy
+the Isaac realization stands the object up with. `_body_collision_aabb` unions those with a running
+`min`/`max`, and Python orders a NaN as neither smaller nor larger than anything, so the comparison
+keeps the accumulator it started with. The geom carrying it disappeared from the bound and nothing
+reported that.
+
+A static fixture whose leg declares `euler="nan 0 0"` was measured as its tabletop alone: `size
+(0.8, 0.6, 0.04)` where the file declares `(0.8, 0.6, 0.77)` - 19.2x too short, and byte-identical
+to the same fixture with the leg deleted, under a successful load. Each spelling failed differently:
+`pos="nan ..."` dropped the geom on the axis that was not finite and kept it on the others;
+`pos="inf ..."` reported an infinite centre and a NaN extent; `size="inf ..."` reported a NaN centre
+and, because `inf - inf` is a NaN that `_recursive_collision_aabb`'s own accumulator drops in turn,
+sized that axis at the `1e-4` floor - the smallest proxy the reader can emit, for the largest geom
+the file can declare.
+
+Reachability is specific. MuJoCo refuses a non-finite geom on a body with a free joint, because the
+inertia it derives is then not finite; it *compiles* the same geom on a body without one - and a
+body without a free joint is exactly what this loader calls a fixture, the tables and cabinets whose
+footprint a manipulation scene is planned against. So the affected scenes are precisely the ones
+that load, which is why the reader cannot defer the question to a compile step.
+
+MuJoCo is also the oracle for the disposition: it refuses a non-finite geom quantity wherever it
+checks one (`nan size in geom`) and warns about the document as a whole (`XML contains a 'NaN'`), so
+refusing is the format's own answer rather than a policy invented here. It is also what this
+module's stated failure semantics already promise - loaders never silently return a phantom robot.
+The same defect class was closed one module over for mesh assets, whose refusal describes this
+mechanism in the same words; `loaders.py` carried no finiteness guard at all.
+
+One guard covers all four parse paths - `pos`, `size`, whichever orientation spelling the geom used,
+and `fromto` - so no one spelling drifts into tolerating what its siblings refuse, and the refusal
+names both the offending attribute and the geom. Only a value that *parsed* is graded, so an
+unreadable attribute keeps falling back to its documented default exactly as before, and a geom with
+no analytic AABB still returns `None` so the caller falls back to another geom. Over the shipped
+corpus - 542 MJCF documents, 14940 `<geom>` elements - zero declare a non-finite placement or
+extent, so no shipped asset changes behaviour.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -39,7 +39,7 @@ from __future__ import annotations
 import math
 import os
 import xml.etree.ElementTree as ET
-from collections.abc import Mapping
+from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

This PR closes a silent measurement defect in the Isaac MJCF scene loader: a geom whose `pos`, `size`, orientation, or `fromto` parses to a NaN/inf was silently dropped from (or inverted in) the AABB that `load_mjcf_scene_objects` reports as a fixture's collision proxy, because a running `min`/`max` orders a NaN as neither smaller nor larger than anything. The fix adds one shared guard, `_refuse_non_finite_geom`, called from all four parse paths in `_geom_aabb`, refusing with a message that names the offending attribute and locates the geom (falling back to the `<default>`-resolved `type` for unnamed geoms). The raise surfaces through `load_mjcf_scene_objects`' documented `ValueError`, which the `load_scene` tool boundary already catches into a structured error dict, so no new unhandled-exception surface is introduced. Only values that *parsed* are graded — unreadable attributes keep their documented default fallback — and mesh/plane geoms keep returning `None`, both pinned by control tests. The change is scoped tightly (one guard + call sites + docstring updates), the corpus measurement (14940 geoms, zero non-finite) shows no shipped asset changes behaviour, and it mirrors the disposition #2740 established for mesh assets one module over.

## What's good

- Regression suite separates premises, regressions, controls, and structural pins (`TestOneOwnerForTheWording` derives the guard's call sites from the AST, so a fifth parse path added later is graded rather than drifting).
- `Raises:` clauses updated on both `_geom_aabb` and the public `load_mjcf_scene_objects` in the same change, per AGENTS.md's docstring-completeness rule.
- Changelog fragment present (`changelog.d/2813-...`), no direct `CHANGELOG.md` edit.
- Refusal message is plain ASCII, names the attribute and the geom, and states the consequence being avoided — actionable at the point a caller sees it.
- The locator reads the `<default>`-resolved attributes, consistent with the module's existing resolver rule and its enforcement test.

PR #2814, opened 2026-08-27
docs(robots): grade the arms page's real-hardware and sim-asset claims against the registry

cagataycali:
## Problem

`docs/robots/arms.md` closes with a *Compatibility notes* block whose bullets are set-membership claims about `registry/robots.json`: which arms ship no simulation asset, and which drive real hardware. A reader plans hardware work from them. All three claims had gone stale, and nothing graded them.

The real-hardware bullet read:

> `panda`, `so100`, and `ur5e` are also supported on real hardware via LeRobot. The rest are simulation-only at the moment

Only `so100` is right.

| named | declares `hardware` | `lerobot_type` | registered by LeRobot | `Robot(name, mode="real")` |
|---|---|---|---|---|
| `panda` | no | - | no | `ValueError: Unsupported robot type: 'panda'` |
| `so100` | yes | `so100_follower` | yes | builds |
| `ur5e` | no | - | no | `ValueError: Unsupported robot type: 'ur5e'` |

LeRobot registers 16 robot types and no Franka among them, so the page told a reader the opposite of the truth about the arm family it names first.

"The rest are simulation-only" then misdescribed the nine arms that do have a path: `dynamixel_2r`, `hope_jr`, `koch`, `omx`, `openarm`, `rebot_b601`, `so101`, `vx300s`, `wx250s`. And the sim-asset exception named two of its three arms — `rebot_b601` declares no `asset` block either.

Measured over the 23 `category: arm` entries:

| claim | page said | registry says |
|---|---|---|
| no sim asset | `hope_jr`, `omx` | `hope_jr`, `omx`, **`rebot_b601`** |
| real via LeRobot | `panda`, `so100`, `ur5e` | `hope_jr`, `koch`, `omx`, `openarm`, `rebot_b601`, `so100`, `so101` |
| real via native driver | *(no such bullet)* | `dynamixel_2r`, `koch`, `vx300s`, `wx250s` |

All 13 arms with no path refuse rather than falling back to sim, which is what makes the last bullet's new wording a measured claim.

## Why nothing caught it

`tests/test_docs_robot_catalog_coverage.py` reads this page. Its four guards pin catalog-table *membership*, the robot *counts* and the *Aliases* column — its own docstring enumerates exactly those jobs, and a capability claim in the prose block is none of them. It passes before and after this change.

## Change

Both existing bullets corrected and the native-driver route named, then all three derived from the repo's own sources of truth rather than restated, so an arm that gains or loses a path fails the bullet that should have named it:

- LeRobot: the entry's `hardware.lerobot_type`.
- Native: `drivers.registry.get_native_driver_class`.

The LeRobot list is read from the registry alone so it grades on an install without LeRobot; a LeRobot-gated premise pins the assumption that makes that sound (every type the arms declare is one LeRobot registers). A second premise pins the coincidence the split currently rests on — every arm with a `hardware` block names a type today, and `reachy_mini` shows a driver-only block is expressible, so the day an arm does that the file reports the two routes have come apart rather than letting the LeRobot bullet gain a robot LeRobot cannot build.

Each claim is located by a pattern keyed on its *subject* rather than on one phrasing, so a reword still grades it: `sim asset` matches both the shipped "have no MuJoCo sim asset" and the new "declare no sim asset". A cell pins that against the exact bullets this replaced, because a pattern matching only the new wording would make the rule report "no such claim" instead of grading the names — and a wrong list would pass.

**Deliberately unchanged:** the block's last bullet, about what the `joints` count includes. Of the 59 registry robots whose asset compiles, 50 declare a `joints` value other than the asset's actuator count, so that field is a loose informational number whose contract needs deciding before it can be graded. That is a different question from which arms drive hardware.

## Verification

Pre-fix split, with the docs reverted to the shipped wording: **4 failed, 12 passed, 1 skipped** — 3 of 3 in the claim class, 1 in the consistency class, **0 of 5 premises and 0 of 5 non-vacuity exemplars**. The headline message names the defect rather than a missing marker:

```
docs/robots/arms.md Compatibility notes: the LeRobot real-hardware bullet lists
['panda', 'so100', 'ur5e'] but the registry says ['hope_jr', 'koch', 'omx',
'openarm', 'rebot_b601', 'so100', 'so101'].
  missing from the page: ['hope_jr', 'koch', 'omx', 'openarm', 'rebot_b601', 'so101']
  named but not true:    ['panda', 'ur5e']
```

Mutations — 11 rows, 10 distinct firing sets, control 0, `collected` stable at 18, and **0 failures in `tests/test_docs_robot_catalog_coverage.py` on every row**:

| mutation | fires | note |
|---|---|---|
| control | 0 | |
| docs reverted to the shipped wording | 4 | the defect |
| only the sim-asset bullet reverted | 1 | |
| only the LeRobot bullet reverted | 2 | |
| the native-driver bullet deleted | 2 | reported as "makes no such claim" |
| over-reach: `panda` **added** to the LeRobot bullet | 2 | refuses a false name, not just a missing one |
| arm-name filter dropped from the extractor | 4 | `lerobot_type` read as a robot |
| LeRobot pattern narrowed to the new phrasing only | 1 | the reword-survival cell |
| bullets not joined across continuation lines | 6 | every rule depends on it |
| absent claim reported as `set()`, not `None` | 3 | |
| derived from `hardware` presence, not `lerobot_type` | 0 | a measured no-op: on today's 23 arms the two readings pick the same set |
| plant an arm with driver-only `hardware` | 1 | closes the row above — the coincidence is now graded |

The three per-bullet reverts partition the defect exactly: their union equals the full-revert set.

Gate: `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1620 files). `mypy` 24 == 24 against a worktree of the merge base, 0 attributable. Paired run over an identical 480-file selection (every suite walking the tree, parsing source, or reading docstrings, `docs/`, `changelog.d/` or the registry), the new file excluded from both trees: **20 failed, 22521 passed, 81 skipped** on each, 22622 collected on each, failing-ID sets identical with `comm` empty both directions. All 20 are environmental — 17 need `awsiot`/`awscrt` (absent, declared in the `[mesh-iot]` extra) and 3 are the lerobot-from-source `encode()` drift.

No visual artifact: this changes documentation and adds tests, none of the policy, simulation, rendering, recording or asset surfaces.

### DIFF
--- a/changelog.d/2814-arm-real-hardware-claims-are-graded.md
+++ b/changelog.d/2814-arm-real-hardware-claims-are-graded.md
@@ -1,0 +1,34 @@
+### Fixed: the arms page's real-hardware and sim-asset claims are graded against the registry
+
+`docs/robots/arms.md` closes with a *Compatibility notes* block whose bullets are set-membership
+claims about `registry/robots.json`: which arms ship no simulation asset, and which drive real
+hardware. A reader plans hardware work from them, so a stale name there is a wrong answer rather
+than a cosmetic one. All three claims had gone stale.
+
+The real-hardware bullet read "`panda`, `so100`, and `ur5e` are also supported on real hardware via
+LeRobot. The rest are simulation-only". Only `so100` was right. Neither `panda` nor `ur5e` declares a
+`hardware` block at all, so `Robot("panda", mode="real")` refuses with `Unsupported robot type:
+'panda'`, and LeRobot registers no Franka type among its sixteen -- the page told a reader the exact
+opposite of the truth about the arm family it names first. "The rest are simulation-only" then
+misdescribed the nine arms that do have a path: `dynamixel_2r`, `hope_jr`, `koch`, `omx`, `openarm`,
+`rebot_b601`, `so101`, `vx300s`, `wx250s`. The sim-asset exception named two of its three arms,
+omitting `rebot_b601`, which declares no `asset` block either.
+
+Nothing graded any of it. `tests/test_docs_robot_catalog_coverage.py` reads the same page and its
+four guards pin catalog-table *membership*, the robot *counts* and the *Aliases* column; a capability
+claim in the prose block is none of those, which its own docstring is precise about.
+
+All three bullets are now derived from the repo's own sources of truth rather than restated, so an
+arm that gains or loses a path fails the bullet that should have named it: LeRobot from the entry's
+`hardware.lerobot_type`, and native from `drivers.registry.get_native_driver_class`. Because the
+LeRobot list is read from the registry alone it grades on an install without LeRobot, and a
+LeRobot-gated premise pins the assumption that makes that sound -- every type the arms declare is one
+LeRobot registers. A second premise pins the coincidence the split currently rests on: every arm with
+a `hardware` block names a type today, and `reachy_mini` shows a driver-only block is expressible, so
+the day an arm does that the file says the two routes have come apart rather than letting the LeRobot
+bullet gain a robot LeRobot cannot build.
+
+Deliberately unchanged: the block's last bullet, about what the `joints` count includes. Of the 59
+registry robots whose asset compiles, 50 declare a `joints` value other than the asset's actuator
+count, so that field is a loose informational number whose contract needs deciding before it can be
+graded -- a different question from which arms drive hardware.
--- a/docs/robots/arms.md
+++ b/docs/robots/arms.md
@@ -101,11 +101,14 @@ _AgileX Piper (6-DOF + gripper)_
 
 - Most arms are loadable in MuJoCo via the registry's asset block and pull from
   [robot_descriptions.py](https://github.com/robot-descriptions/robot_descriptions.py)
-  on first use. Exceptions: `hope_jr` and `omx` have no MuJoCo sim asset and require
-  physical hardware.
-- `panda`, `so100`, and `ur5e` are also supported on real hardware via LeRobot. The rest
-  are simulation-only at the moment (real-hardware support is a per-robot effort that
-  upstreams to LeRobot).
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Corrects three stale set-membership claims in the *Compatibility notes* block of `docs/robots/arms.md` (the sim-asset exceptions, the LeRobot real-hardware list, and a previously missing native-driver bullet) and adds `tests/test_docs_arm_compatibility_notes.py`, which derives all three sets from the repo's own sources of truth (`registry/robots.json` for `hardware.lerobot_type` and asset presence, `drivers.registry.get_native_driver_class` for the native route) so a future registry change fails the bullet that should have named it. I independently re-derived all three sets from the registry at head and they match the corrected page exactly (23 arms; no-asset = `hope_jr`/`omx`/`rebot_b601`; lerobot = 7 names; native = the 4 arm names in `DynamixelDriver.SUPPORTED_ROBOTS`). The new suite passes at head (18 passed), and the repo's docstring-xref, dangling-ref, and test-naming hygiene graders all pass over the new file.

## What's good

- Claims are derived, not restated: the LeRobot list reads the registry alone (grades without LeRobot installed), with an `importorskip`-gated premise pinning that every declared `lerobot_type` is one LeRobot registers.
- The native-driver derivation is environment-safe by construction: the arm entries come from the Dynamixel driver, which the shipped-drivers table documents as having no optional import that can fail.
- Premise tests pin the `hardware`-block/`lerobot_type` coincidence the split rests on, and the claim patterns are pinned against the exact shipped wording they replaced, so a reword still grades the names.
- Changelog fragment present per AGENTS.md step 3; all-ASCII prose; no host paths; scope is docs + tests only, with the `joints` bullet explicitly and defensibly left out of scope.

PR #2815, opened 2026-08-27
fix(mesh): report whether InputPublisher.stop actually stopped the loop

cagataycali:
## What is wrong

`InputPublisher.stop` joined its publish thread with a two-second budget and
discarded the outcome. `Thread.join(timeout=...)` returns `None` whether or not
the thread finished, so the liveness read after it is the only thing that
distinguishes a stopped loop from one that outlasted the budget. `stop` took no
such read: it cleared the session flag, joined, logged `input publisher stopped`
unconditionally, and returned `stats` whose `running` key it had already set to
`False`.

A teleoperator whose `get_action()` is blocking is the ordinary case a join
budget exists for. That loop stays free to publish another frame to
`strands/{peer}/input/{device}` after `stop()` has returned - an actuator command
every subscribed follower mirrors - and nothing in the returned stats could say
so.

## Measured

A publisher whose leader blocks inside `get_action()`. The loop is parked in the
read deterministically: the double signals on entry and then blocks, so the
caller knows the loop is in the device read rather than in the ticker before it
calls `stop()`.

```
                                         before      after
stop() elapsed                           2.00s       2.00s
stats running                            False       False
stats thread_alive                       (absent)    True
loop thread still alive after the call    True        True
log level for the outcome                INFO        WARNING
log claims "publisher stopped"           yes         no
second stop() can re-join the loop       no          yes
frame published after stop() returned    1           1
```

The last row is the consequence the stop claim hid. The first frame is published,
the loop re-enters `get_action()` and blocks, `stop()` waits out its budget and
reports a stop that did not happen; once the read returns, that frame reaches
subscribers.

Two further readings from the same run:

- `running` is the session flag `stop` clears **before** it joins, so a caller
  polling afterwards was told `running=False` about a loop that was still
  publishing. `thread_alive` reads the thread itself, and the two now differ for
  exactly as long as a loop outlives the stop that asked it to exit.
- The guard `if not self._running` made a timed-out stop terminal: the flag was
  already clear, so a second call returned early and the only handle to that live
  loop was unreachable through this surface.

## The change

`strands_robots/mesh/input.py`, +55/-14.

- The join budget is named `_INPUT_JOIN_TIMEOUT_S = 2.0`, so the docstring, the
  warning text and the tests read one value.
- `stop` reads liveness after the join. On failure it logs at WARNING naming the
  budget, the topic still being written to and the remedy, instead of announcing
  a stop that did not happen.
- `stats` gains `thread_alive` beside `running`. `running` keeps its meaning
  rather than being widened, because the loop's own `while` reads that flag.
- The guard admits a call that has a live thread left to join, so a publisher
  whose join timed out stays stoppable.

This is the shape `TeleopMixin.stop_teleoperate` already carries: its status verb
gained `thread_alive` for the same reason, and `G1Driver.cleanup` refuses to tear
a transport down under a live writer. This is the sibling loop that was left.

## Why nothing caught it

`TestPublisherLifecycle` has three cells - a repr read, a stop on a
never-started publisher, and a start/stop that returns stats - and every one is a
clean path. `is_alive`, `join` and any wedged device appear nowhere in the seven
`tests/mesh/test_input_*.py` suites, so no cell could reach the branch where the
budget expires. The `sib` column below is `0` on all eleven mutations, across
those seven files.

## Verification

Pre-fix, source reverted and the new suite kept: **10 failed / 8 passed**.

| class | role | failed/total |
|---|---|---|
| `TestAWedgedLoopIsReportedNotAnnounced` | regression | 7/7 |
| `TestAJoinThatTimedOutIsStillStoppable` | regression | 1/1 |
| `TestTheBudgetIsNamedOnce` | structural | 2/2 |
| `TestTheCleanPathIsUnchanged` | control | 0/5 |
| `TestThePremises` | premise | 0/3 |

Every expectation in the control class is one the pre-fix code also met.

Mutations against the fixed tree. `new` counts failures in the new suite, `sib`
in the seven pre-existing input suites, and `coll` is stable at 18 throughout so
no row is hidden by a deselection.

| # | mutation | new | sib |
|---|---|---|---|
| b1 | liveness read dropped, INFO unconditional | 4 | 0 |
| b2 | `thread_alive` stat removed | 5 | 0 |
| b3 | `thread_alive` hardcoded `False` | 5 | 0 |
| b4 | `thread_alive` reads the session flag, not the thread | 5 | 0 |
| b5 | join budget back to a literal | 1 | 0 |
| b6 | warning downgraded to INFO | 3 | 0 |
| b7 | guard returns early on the session flag alone | 2 | 0 |
| b8 | liveness read taken **before** the join | 1 | 0 |
| b9 | budget name kept, set to `0.0` | 1 | 0 |
| b10 | warning names a placeholder, not the real topic | 1 | 0 |
| b11 | warning names a placeholder, not the device | 0 | 0 |

Eleven mutations, ten detected, seven distinct firing sets, control clean, source
restored byte-identically.

Four rows are worth reading:

- **b2/b3/b4 share one firing set.** All three leave `thread_alive` unable to
  report a live thread, so they are one mutation class spelled three ways.
- **b8 and b9 both fire a *control*** - `test_a_clean_stop_logs_at_info`. Reading
  liveness before the join, or waiting zero, makes a healthy stop report a
  warning. The ordering and the budget are load-bearing at runtime, not only in
  the AST assertion.
- **b11 fires nothing, and that is correct.** The topic's last segment *is* the
  device name, so a message that names the topic still identifies the publisher.
  The device is reported twice; dropping one copy loses nothing a caller needs.

Gate, on an identical 573-file selection (all of `tests/mesh/`, plus every suite
that walks the tree, parses source, or reads docstrings, `Args:`, `Raises:` or
`changelog.d/`), with the new file excluded from both trees and every path
verified present on both:

```
                       branch                          base
collected              23506                           23506
failing ids            48                              48
rootdir                robots-src                      run-.../base
```

Failing-ID sets identical, `comm` empty in both directions. Of the 48, **45**
need `awsiot`/`awscrt` - `find_spec` is `None` for both, and they are declared in
the `[mesh-iot]` extra, which CI installs - and **3** are the pre-existing
lerobot-from-source `encode()` drift in `tests/training/test_lerobot*`.

tests_integ`. mypy **24 == 24** against a worktree pinned at the merge base,
normalized message sets byte-identical in both directions, 0 attributable to
either changed file.

No visual artifact: this changes what a teardown reports, not a policy,
simulation, rendering, recording or asset. The measured before/after table is the
evidence. The `stats` docstring is the documented surface for these keys - no
`docs/` page enumerates them - and it is updated in the same change.

### DIFF
--- a/changelog.d/2815-input-publisher-stop-reports-the-join-outcome.md
+++ b/changelog.d/2815-input-publisher-stop-reports-the-join-outcome.md
@@ -1,0 +1,45 @@
+### Fixed: `InputPublisher.stop` reports whether the publish loop actually stopped
+
+`threading.Thread.join(timeout=...)` returns `None` whether or not the thread
+finished, so the liveness read after it is the only thing that distinguishes a
+stopped loop from one that outlasted the budget. `InputPublisher.stop` cleared
+the session flag, joined with a 2 s budget and never took that read: it then
+logged `input publisher stopped` unconditionally and returned `stats`, whose
+`running` key it had already set to `False`. A teleoperator whose `get_action()`
+was still blocking - a serial read on a wedged bus is the ordinary case a join
+budget exists for - was therefore reported as stopped while its loop was still
+free to put another frame on `strands/{peer}/input/{device}`, which is an
+actuator command every subscribed follower mirrors. Measured on a publisher
+whose leader blocks inside `get_action()`:
+
+```
+                                        before      after
+stop() elapsed                          2.00s       2.00s
+stats running                           False       False
+stats thread_alive                      (absent)    True
+loop thread still alive after the call   True        True
+log level for the outcome               INFO        WARNING
+log claims "publisher stopped"          yes         no
+second stop() can re-join the loop      no          yes
+frame published after stop() returned   1           1
+```
+
+The last row is the consequence the stop claim hid: a frame reaches subscribers
+after the caller has been told the publisher stopped. Nothing in the returned
+stats could say so, because `running` is the session flag `stop` clears *before*
+it joins - so a caller polling afterwards was told `running=False` about a loop
+that was still publishing. `thread_alive` reads the publish thread itself, and
+the two now differ for exactly as long as a loop outlives the stop that asked it
+to exit.
+
+`running` keeps its meaning rather than being widened, because the loop's own
+`while` reads that flag; the join outcome is reported beside it. This mirrors
+`TeleopMixin.stop_teleoperate`, whose status verb gained `thread_alive` for the
+same reason, and `G1Driver.cleanup`, which refuses to tear a transport down
+under a live writer.
+
+A publisher whose join timed out also stays stoppable. The guard now admits a
+call that has a live thread left to join, where returning early on the session
+flag alone left the only handle to that loop unreachable through `stop()`. The
+2 s budget is named `_INPUT_JOIN_TIMEOUT_S` so the docstring, the warning text
+and the tests read one value.
--- a/strands_robots/mesh/input.py
+++ b/strands_robots/mesh/input.py
@@ -55,6 +55,11 @@ logger = logging.getLogger(__name__)
 
 INPUT_HZ_DEFAULT = 50.0
 
+# Budget for joining the background publish thread in InputPublisher.stop.
+# Named so the docstring, the ``thread_alive`` stat and the tests read one
+# value. Matches the teleop mixin's _TELEOP_JOIN_TIMEOUT_S in purpose.
+_INPUT_JOIN_TIMEOUT_S = 2.0
+
 #: How many times each :class:`InputPublisher` loop failure is logged before
 #: going quiet. The loop runs at ``hz`` (50 Hz by default), so an unbounded
 #: log of a persistent fault - a dead publish transport, a teleoperator whose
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Makes `InputPublisher.stop()` report the join outcome instead of discarding it: the 2 s join budget is hoisted into a named constant (`_INPUT_JOIN_TIMEOUT_S`), `stats` gains an additive `thread_alive` key that reads the publish thread directly (mirroring `TeleopMixin.stop_teleoperate`'s existing `thread_alive`), a timed-out join is logged at WARNING (naming the budget, device, topic and remedy) rather than unconditionally announcing "publisher stopped", and the stop guard now admits a re-join call when the session flag is clear but the thread is still alive - so a wedged loop stays reachable through `stop()`. The `running` flag keeps its meaning since the loop's own `while` reads it. Verified on the head: all 18 new tests pass, the 7 neighbouring `tests/mesh/test_input_*` suites (160 tests) pass unchanged, log strings are ASCII, durations are measured on `time.monotonic()`, and the changelog fragment is present (`changelog.d/2815-...md`, added in the second commit).

## What's good

- The stats change is additive (`thread_alive` beside `running`, no key repurposed), so existing consumers of the returned dict are unaffected.
- The wedged-loop tests park the loop inside `get_action()` deterministically via events rather than sleeps, shrink the budget for the fast cells, and keep one cell that grades the shipped 2.0 s value; an AST cell pins that the join reads the named constant rather than a literal.
- `TestTheCleanPathIsUnchanged` and `TestThePremises` pin the pre-fix behaviour and the fixture's assumptions separately, per AGENTS.md's "test behavior, not implementation" and premise-test conventions.
- The re-entrant guard (`not self._running and not (thread alive)`) is the correct widening: it preserves the never-started and already-stopped early returns while keeping a live loop stoppable.

yinsong1986 (approved):
## Summary

Makes `InputPublisher.stop()` honest about the join outcome instead of discarding it. `Thread.join(timeout=...)` returns `None` either way, so the post-join liveness read added here is the only signal separating a stopped loop from one that outlasted the budget; previously `stop()` cleared the session flag, joined, logged `input publisher stopped` unconditionally, and returned stats whose `running` key it had already forced to `False` - so a wedged `get_action()` could put one more actuator frame on `strands/{peer}/input/{device}` after the caller was told the publisher stopped. The fix names the budget once (`_INPUT_JOIN_TIMEOUT_S`), adds an additive `thread_alive` stat that reads the thread itself (keeping `running` unchanged, since the loop's own `while` reads it), logs a timed-out join at WARNING naming the budget/topic/remedy, and widens the re-entry guard so a timed-out stop stays stoppable. The guard captures `self._thread` into a local before testing it, so there is no TOCTOU on the None check, and `_thread` is never reset to `None`, so the new `stats` read cannot NPE. This mirrors the shape `TeleopMixin.stop_teleoperate` / `_TELEOP_JOIN_TIMEOUT_S` already carry, so it converges a sibling rather than inventing a new convention.

## What's good

- Regression suite parks the loop inside `get_action()` deterministically (event-signalled, no sleeps), shrinks the budget for the wedged cells, and keeps one cell on the shipped 2.0s value so the budget itself stays graded; the mutation table in the description shows the ordering and the budget are load-bearing at runtime, not only in the AST assertion.
- `running` keeps its meaning instead of being widened - the two-flag split (`running` = session flag, `thread_alive` = thread read) is documented in the `stats` docstring, the surface a caller actually reads.
- Changelog fragment under `changelog.d/`, behaviour-describing test names, ASCII-only log strings, `time.monotonic()` throughout - all per AGENTS.md.

PR #2816, opened 2026-08-27
fix(drivers/g1): read whether the control loop actually halted before releasing the wire

cagataycali:
## Problem

`_ControlLoop.stop` signals the loop, joins its thread within a budget, and
**returns whether the thread joined**. Its own docstring says why that return
value exists:

> `False` when the loop is still running - a caller-supplied policy that
> outlasts the join budget (a remote inference call is the ordinary case) needs
> the honest `stopped=False` in the `stop_task` envelope rather than a `success`
> claim the payload's `running=True` contradicts.

Three call sites, one of which read it:

| caller | reads the outcome | closes `_pubs` after |
|---|---|---|
| `stop_task` | yes | no |
| `G1Driver.stop` | **no** | no |
| `G1Driver.cleanup` | **no** | **yes** |

For `cleanup` that was load-bearing. `_emit_zero_torque` reads
`self._driver._pubs` and returns silently when it is `None`, so closing the
publisher under a live thread does not merely release a resource early - it
discards the shutdown frame. `cleanup`'s own docstring names that outcome as the
thing the path exists to prevent:

> Closing `_pubs` under a live 500 Hz thread would drop the loop into its
> `publish` branch and skip the zero-torque frame - the fall this whole path
> exists to prevent.

## Measured

One rollout, one `cleanup()`, two policies. The zero-torque frame is the one
commanding zero stiffness on the commanded joint.

| | policy returns at once | policy outlasts the join budget |
|---|---|---|
| `cleanup()` returned after | 0.05 s | 2.00 s |
| loop still running | False | **True** |
| `_pubs` | `None` | `None` |
| publisher closed | True | **True, under the live thread** |
| zero-torque frames at that point | 1 | 0 |
| **zero-torque frames after the policy returns** | 1 | **0** |

The frame is not delayed, it is lost: the loop reaches its own `finally`, calls
`_emit_zero_torque`, finds `_pubs is None` and returns. The recorded
`exit_reason` is `stop_task`, so the run also reports a clean caller-driven stop.

After this change the right-hand column reads `_pubs` kept, `closed False`, and
**1** zero-torque frame once the policy returns.

## Change

`cleanup` releases the publisher only once the loop is provably gone. A loop that
is still running keeps it, so the frame that loop publishes from its own
`finally` reaches the wire, and the overrun is reported at ERROR naming the
remedy. The subscribers close either way - the loop never reads them - and a
second `cleanup` releases the rest.

`stop` carries no envelope and releases nothing, so its frame was never at risk.
It now logs the overrun rather than returning from shutdown in silence while a
thread still holds the wire. The two are deliberately different: one is a
behaviour change, the other a diagnostic.

## Why nothing caught it

The zero-torque contract already has a test, and its policy returns
immediately:

```python
def policy(_obs: Any) -> dict[str, float]:
    return {"left_knee": 0.0}
```

so its join always succeeded and the discarded value was always `True`. The fast
path cannot distinguish a checked halt from an unchecked one. Every regression
cell here blocks a policy past the join budget - the regime `stop_task`'s
docstring calls ordinary - which is why they cost about two seconds each.

Measured directly: on the pre-fix tree the whole defect fires **0** of the
**685** pre-existing cells in `tests/drivers/`, `tests/tools/g1/` and
`tests/test_driver_seam.py`.

## Tests

Pre-fix split, source reverted with the new file kept: **7 failed / 12 passed**.
By class - 5/5 `TestCleanupKeepsThePublisherForAnUnjoinedLoop`, 1/1
`TestStopReportsAnUnjoinedLoop`, 1/2 `TestEveryTeardownPathReadsTheOutcome`, and
**0 of 3 `TestThePremise`, 0 of 3 `TestNothingLeaksEitherWay`, 0 of 5
`TestTheJoinedPathIsUnchanged`**. The headline failure is the defect verbatim:
`assert 0 == 1` on the zero-torque frame count.

`TestNothingLeaksEitherWay` holds on both trees by design: deferring the release
is only defensible if the resources are still released.

### Mutations

`mine` = failures in the new file, `coll` = cells collected, `sib` = failures in
the three pre-existing suites.

| mutation | mine | coll | sib |
|---|---|---|---|
| control (no change) | 0 | 19 | 0 |
| both discards restored (raw revert) | 7 | 19 | 0 |
| `cleanup` discards the outcome | 6 | 19 | 0 |
| `stop` discards the outcome | 2 | 19 | 0 |
| `cleanup` reads it but closes `_pubs` anyway | 3 | 19 | 0 |
| overrun reported at DEBUG, not ERROR | 2 | 19 | 0 |
| report drops "did not join" | 1 | 19 | 0 |
| report drops the remedy | 1 | 19 | 0 |
| subscribers held back too (over-reach) | 1 | 19 | 0 |
| `joined` starts `False` (over-reach) | 3 | 19 | **1** |

10 mutations, 10 distinct firing sets, none undetected, `collected` stable at 19
so no row hid a deselection, source restored md5-identical.

Two rows carry most of the argument. **`cleanup` reads it but closes `_pubs`
anyway** is the tempting smaller fix - log the overrun and release regardless -
and it still fires the three frame-and-publisher cells, so deferring the release
is the behavioural content rather than the log. **`joined` starts `False`** is
the over-reach direction and it trips a pre-existing cell
(`test_cleanup_is_idempotent_with_no_loop`), because a driver with no loop at all
must still release everything.

Complementarity is exact: the `cleanup`-only revert fires 6, the `stop`-only
revert fires 2, they share the derived structural cell, and the union is the 7 of
the full revert.

`TestEveryTeardownPathReadsTheOutcome` is derived from the module's own AST, so a
fourth teardown path cannot land discarding the join result, with a non-vacuity
cell asserting the banned shape is expressible in the first place.

## Gate

Identical 430-file selection - all of `tests/drivers/`, `tests/tools/g1/`,
`tests/test_driver_seam.py` plus every suite that walks the tree, parses source
or reads docstrings - with the new file excluded from both trees and every path
verified present on both: identical **20680 collected** and
`20 failed, 20587 passed, 73 skipped` on each, 20 identical failing ids, `comm`
empty in both directions, two distinct rootdirs. All 20 are pre-existing: 17 need
`awsiot`/`awscrt` (absent here, declared in the `[mesh-iot]` extra) and 3 are the
lerobot-from-source `encode()` drift.

tests_integ`. mypy **24 == 24** against a worktree of the merge base, normalized
message sets byte-identical in both directions, 0 attributable to the changed
files.

No visual artifact: this is a teardown-reporting change with no policy,
simulation, rendering, recording or asset behaviour in it. The measured
before/after frame counts are the evidence.

### DIFF
--- a/changelog.d/2816-cleanup-reads-the-loop-halt-outcome.md
+++ b/changelog.d/2816-cleanup-reads-the-loop-halt-outcome.md
@@ -1,0 +1,33 @@
+### Fixed: the G1 zero-torque shutdown frame survives a policy that overruns the stop budget
+
+`_ControlLoop.stop` signals the control loop, joins its thread within a budget,
+and returns whether the thread actually joined. `G1Driver.stop_task` reads that
+value and reports `stopped=False` when the join times out - its docstring
+explains why, and names the case: a caller-supplied policy that outlasts the
+budget, "a remote inference call is the ordinary case". The two teardown paths
+called the same method and dropped the answer.
+
+For `cleanup` that was load-bearing rather than cosmetic. It halted the loop and
+then closed `_pubs` and set it to `None` regardless of whether the halt had
+taken, so on an overrun it removed the publisher from underneath a live 500 Hz
+thread. `_emit_zero_torque` reads `self._driver._pubs` and returns silently when
+it is `None`, so the zero-torque shutdown frame was never published: not by
+`cleanup`, and not later when the policy finally returned and the loop reached
+its own `finally`. That is the fall `cleanup`'s docstring says the path exists to
+prevent, and the loop still recorded a clean caller-driven `exit_reason` of
+`stop_task` on the way out.
+
+`cleanup` now releases the publisher only once the loop is provably gone. A loop
+that is still running keeps it, so the frame that loop publishes from its own
+`finally` reaches the wire, and the overrun is reported at ERROR naming the
+remedy - call `cleanup` again once the loop has exited. The subscribers close
+either way, because the loop never reads them, and a second `cleanup` releases
+the rest. `stop` carries no envelope and releases nothing, so its frame was
+never at risk; it now logs the overrun rather than returning from shutdown in
+silence while a thread still holds the wire.
+
+The zero-torque contract already had a test, and it drives a policy that returns
+immediately. Its join therefore always succeeded and the discarded value was
+always `True`, so the fast path could not distinguish a checked halt from an
+unchecked one. The new cells block a policy past the join budget, which is the
+regime `stop_task` calls ordinary.
--- a/strands_robots/drivers/g1.py
+++ b/strands_robots/drivers/g1.py
@@ -502,12 +502,25 @@ class G1Driver:
         its thread before returning - a controlled stop rather than
         abrupt frame cessation mid-policy on a robot whose FSM has just
         gone away.  Idempotent: no running task returns immediately.
+
+        A loop that outlasts the join budget is reported rather than
+        returned: this signature carries no envelope, and returning from
+        shutdown while a thread still holds the wire is exactly what the
+        log exists to say.  Unlike :meth:`cleanup` this leaves the
+        publisher open either way, so that loop still publishes its
+        zero-torque frame when the policy finally returns.
         """
         with self._task_admission:
             loop = self._loop
         if loop is not None and loop.is_running:
             logger.debug("%s.stop() halting control loop", self._tool_name)
-            loop.stop("stop_task")
+            if not loop.stop("stop_task"):
+                logger.error(
+                    "%s.stop(): control loop did not join within the stop budget "
+                    "and still holds the wire; its zero-torque frame publishes "
+                    "when the policy returns",
+                    self._tool_name,
+                )
 
     def cleanup(self) -> None:
         """Release every DDS subscriber and publisher. Idempotent.
@@ -517,17 +530,38 @@ class G1Driver:
         ``_pubs`` under a live 500 Hz thread would drop the loop into
         its ``publish`` branch and skip the zero-torque frame - the fall
         this whole path exists to prevent.
+
+        Halting is not the same as having halted.
+        :meth:`_ControlLoop.stop` reports whether the thread actually
+        joined, and a caller-supplied policy that outlasts the join
+        budget - a remote inference call is the ordinary case - leaves it
+        running.  The publisher is therefore released only once the loop
+        is provably gone: a loop that is still running keeps it, so the
+        zero-torque frame that loop publishes from its own ``finally``
+        reaches the wire instead of being dropped by
+        :meth:`_ControlLoop._emit_zero_torque`'s ``pubs is None`` return.
+        The subscribers close either way - the loop never reads them - and
+        a second ``cleanup()`` once the loop has exited releases the rest.
         """
         with self._task_admission:
             loop = self._loop
+        joined = True
         if loop is not None and loop.is_running:
-            loop.stop("stop_task")
+            joined = loop.stop("stop_task")
         if self._subs is not None:
             self._subs.close()
             self._subs = None
-        if self._pubs is not None:
-            self._pubs.close()
-            self._pubs = None
+        if joined:
+            if self._pubs is not None:
+                self._pubs.close()
+                self._pubs = None
+        else:
+            logger.error(
+                "%s.cleanup(): control loop did not join within the stop budget, "
+                "so the publisher stays open for its zero-torque frame; call "
+                "cleanup() again once the loop has exited",
+                self._tool_name,
+            )
         self._connected = False
 
     # ------------------------------------------------------------------ #
--- a/tests/drivers/test_g1_cleanup_reads_the_loop_halt_outcome.py
+++ b/tests/drivers/test_g1_cleanup_reads_the_loop_halt_outcome.py
@@ -23,14 +23,62 @@ cost about two seconds each.
 from __future__ import annotations
 
 import logging
+import sys
 import threading
 import time
+import types
 from typing import Any
 
 import pytest
 
 import strands_robots.drivers.g1 as g1_mod
-from tests.drivers.test_g1_control_loop import _ControlLoop, _RecordingPublisher
+from tests.drivers.test_g1_control_loop import (
+    _ControlLoop,
+    _RecordingPublisher,
+    _StubCRC,
+    _StubLowCmd,
+)
+
+
+@pytest.fixture(autouse=True)
+def _stub_unitree_sdk(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Install a ``unitree_sdk2py`` stub for the duration of one test.
+
+    The loop's write path and its zero-torque ``finally`` both call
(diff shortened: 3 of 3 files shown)


yinsong1986 (dismissed):
## Summary

This PR makes the two G1 teardown paths (`G1Driver.stop` and `G1Driver.cleanup`) read the boolean `_ControlLoop.stop` already returns instead of discarding it. For `cleanup` the discard was load-bearing: closing `_pubs` under a loop whose policy outlasts the 2 s join budget silently dropped the zero-torque shutdown frame, because `_emit_zero_torque` returns without publishing when `self._driver._pubs is None` -- exactly the fall the method's own docstring says the path exists to prevent. The fix defers the publisher release until the loop is provably gone (a second `cleanup` releases it, and the ERROR log names that remedy), while subscribers still close either way and `_connected` still drops, so further writes are refused through `_check_motion_gates`. I verified the two claims the change rests on against the head tree: the loop's `finally` reads `_pubs` at emit time (so the kept publisher is what lets the deferred frame reach the wire), and the loop never reads `_subs` (it consumes only the cached `_mode_machine`/`_fsm_id`/`_battery`/`_imu` observations plus `_pubs`), so closing subscribers under the live thread is safe. The `stop()` half is logging-only, no signature or envelope changes anywhere, and the loop's `finally` clears `self._driver._loop` under `_task_admission`, so the second-`cleanup` release path is real rather than aspirational.

## What's good

- Changelog fragment present (`changelog.d/2816-...`), per AGENTS.md step 3.
- The regression suite drives the actual overrun regime (policy parked past the join budget) that the shipped fast-path test could never distinguish, with premise cells, over-reach controls (`TestTheJoinedPathIsUnchanged`, the `joined = False` mutation tripping a pre-existing idempotency cell), and leak guards that hold on both trees.
- `TestEveryTeardownPathReadsTheOutcome` is AST-derived with a non-vacuity cell, so a fourth teardown path discarding the join result is graded on arrival.
- Log strings are plain ASCII, deadlines use `time.monotonic()`, test names describe behaviour -- all per AGENTS.md conventions.

yinsong1986 (approved):
## Summary

This PR makes both G1 teardown paths read the boolean `_ControlLoop.stop` already returns instead of discarding it. For `cleanup` the discard was load-bearing: when a caller-supplied policy outlasts the 2 s join budget, closing `_pubs` under the still-live 500 Hz thread silently dropped the zero-torque shutdown frame, because `_emit_zero_torque` returns without publishing when `pubs is None`. The fix defers the publisher release until the loop is provably gone (kept publisher -> the loop's own `finally` publishes the frame when the policy returns; a second `cleanup` releases it, and the ERROR log names that remedy), while subscribers still close either way (the loop never reads them - verified against the head tree: `_call_policy` consumes only the cached `_mode_machine`/`_fsm_id`/`_battery`/`_imu` plus `_pubs`) and `_connected` still drops, so the loop's per-step re-gate exits it cleanly on its next iteration. I traced the join-order invariant the release depends on: `loop.stop(...) is True` implies the thread joined, so its `finally` (including the zero-torque emit and the `_loop` clear under `_task_admission`) has completed before `cleanup` touches `_pubs` - no window where a joined path closes the publisher under a live emit. The `stop()` half is logging-only; no public API, wire-format, or envelope changes anywhere. The delta since the prior (dismissed) review is the `_stub_unitree_sdk` autouse fixture, which uses `monkeypatch.setitem(sys.modules, ...)` per the AGENTS.md sys.modules rule and mirrors the sibling suite's stub so the frame-counting cells run the production publish lane on an SDK-less box.

## What's good

- Changelog fragment present (`changelog.d/2816-...`), per AGENTS.md step 3.
- The regression suite drives the actual overrun regime (policy parked past the join budget) that the shipped fast-path test could never distinguish, with premise cells, over-reach controls (`TestTheJoinedPathIsUnchanged`, `TestNothingLeaksEitherWay` holding on both trees), and a mutation table with 10 distinct firing sets.
- `TestEveryTeardownPathReadsTheOutcome` is AST-derived with a non-vacuity cell, so a later teardown path discarding the join result is graded on arrival.
- Log strings are plain ASCII, deadlines use `time.monotonic()`, test names describe behaviour - all per AGENTS.md conventions.

cagataycali:
## Independent verification of `e3551702`, and why the stub had to be re-declared

The `Run tests` failure on `dce79410` reproduces here and the cause is confirmed: `pytest` binds an `autouse` fixture to the module that **declares** it, so importing `_ControlLoop` and `_RecordingPublisher` from the control-loop suite did not bring `_stub_unitree_sdk` with them. Without it `_build_zero_torque_lowcmd` returns `(None, "unitree_sdk2py is not installed: ...")`, `_emit_zero_torque` logs at debug and returns, and the two cells read 0 published frames.

CI reported one failure only because `-x` stopped the run. With the SDK hidden there were **two**:

```
FAILED ...::TestCleanupKeepsThePublisherForAnUnjoinedLoop::test_the_zero_torque_frame_reaches_the_wire - assert 0 == 1
FAILED ...::TestTheJoinedPathIsUnchanged::test_the_zero_torque_frame_still_goes_out - assert 0 == 1
```

### The fix is verified in both dependency states

`unitree_sdk2py` is installed on this box, so a plain local run cannot see this defect at all. Measurements below hide the module at both layers a suite can consult (`find_spec` returns `None`, and an import raises `ModuleNotFoundError` with `.name` set), with `numpy` and the pytest plugins confirmed still resolvable as controls.

| selection | SDK hidden | SDK present |
|---|---|---|
| the changed file | 19 passed | 19 passed |
| `tests/drivers/` | 548 passed, 33 skipped | 578 passed, 3 skipped |

The skip arithmetic reconciles: `548 + 33 == 578 + 3 == 581`.

Wider gate, 180 files (all of `tests/drivers/`, `tests/tools/g1/`, and every suite that walks the tree, parses source or reads docstrings), base `dce79410` against `e3551702`:

| tree | result |
|---|---|
| `dce79410` | `2 failed, 7376 passed, 96 skipped` |
| `e3551702` | `7378 passed, 96 skipped` |

New failures: none. Fixed: exactly the two cells above. `7376 + 2 == 7378`.

`ruff check` and `ruff format --check` over `strands_robots tests tests_integ` are clean, 1626 files already formatted.

### Re-declaring the fixture is the more robust of the two options

The smaller edit is to import the sibling's fixture for its `autouse` effect. That works, and it is measurably more fragile - it depends on the sibling keeping `autouse=True`:

| mutation | fixture imported for its effect | fixture re-declared locally (this branch) |
|---|---|---|
| control | 0 failed / 19 collected | 0 failed / 19 collected |
| the sibling's fixture stops being `autouse` | **2 failed** | **0 failed** |

Re-declaring costs a second copy of the eight-name registration list, so a ninth SDK submodule would need two edits rather than one. Both cells fail loudly in that case, so it is a maintenance cost rather than a detection gap - and it buys independence from a marker in another module, which is the failure this PR exists to fix. Sharing `_StubLowCmd` and `_StubCRC` while re-declaring only the fixture keeps the stub *classes* single-sourced, which is the part that has to agree with the IDL.

### One classification, so it is not chased

A wider run on either tree shows four failures in `tests/test_doctor.py` and friends with `PackageNotFoundError: No package metadata was found for strands-robots`. They appear identically on `dce79410` and `e3551702` and are an artefact of running from a bare `git worktree`, where the editable install's distribution metadata is not present. Not attributable to either commit.

PR #2817, opened 2026-08-27
fix(drivers): name a robot's native driver instead of lerobot's vocabulary

cagataycali:
## What is wrong

`Robot("vx300s", mode="real")` is refused with a list of the sixteen robot types
lerobot knows, none of which is `vx300s` — and never mentions that this package
ships the driver that builds it. Four registry entries are in that position.

```
ValueError: Unsupported robot type: 'vx300s'. Known lerobot robot types:
['bi_openarm_follower', 'bi_rebot_b601_follower', 'bi_so_follower',
 'earthrover_mini_plus', 'hope_jr_arm', 'hope_jr_hand', 'koch_follower',
 'lekiwi', 'lekiwi_client', 'omx_follower', 'openarm_follower',
 'rebot_b601_follower', 'reachy2', 'so100_follower', 'so101_follower',
 'unitree_g1']
```

Two registries answer "what builds this robot": lerobot's `RobotConfig`
ChoiceRegistry, and this package's native-driver registry. A robot can be in the
second and not the first — that gap is what a native driver exists to close, and
`drivers/reachy.py` records it in as many words:

> the Reachy Mini has no lerobot robot type, so before this driver `mode="real"`
> raised `ValueError: Unsupported robot type: 'reachy_mini'`.

The Reachy Mini also declares `hardware.driver="strands"`, so `resolve_driver`
sends it to its driver and it never meets that refusal. The four robots the
Dynamixel driver serves that have no lerobot type declare nothing, so the default
routes them to lerobot, which cannot build them — and the refusal they get points
nowhere a caller can go.

Measured on `main`, over every registry robot:

| | count |
|---|---|
| registry robots reaching this refusal | 55 |
| of those, a native driver exists in this package | **4** |
| of those, no driver of either kind (listing is the right answer) | 51 |

The four are `vx300s`, `wx250s`, `trossen_wxai` and `dynamixel_2r`. All four
build fine today with `driver="strands"` — there is simply no way for a caller to
learn that.

## The change

The site already had this shape for the other wrong entry point. A leader arm is
a lerobot *teleoperator*, and `teleoperator._other_lerobot_kind_refusal` names it
as one rather than answering with follower types. Its docstring is where the rule
is written down:

> a request for one kind that names a device of the other kind is a wrong entry
> point, not a typo — and answering it with the names of the kind it is not
> answers the wrong question. […] `None` means the name is genuinely unknown, so
> the caller keeps its own listing of the kind that was asked for.

`drivers.registry._native_driver_refusal` is that function's sibling for a
natively driven robot: same site, same report-or-`None` contract, so a name with
no native driver keeps the listing that is right for it.

```
ValueError: Unsupported robot type: 'vx300s'. lerobot has no robot type for it,
but this package ships a native driver for it (DynamixelDriver): build it with
Robot('vx300s', mode='real', driver='strands', ...). To make that the default
for this robot, declare hardware.driver='strands' on its registry entry.
```

**Deliberately unchanged: which driver wins.** Resolution precedence is
untouched and no registry entry gains a declaration. `koch` and `aloha` have both
a working lerobot type and a native driver; whether they should prefer the native
one is a preference, and `unitree_g1` shows the registry is where such a
preference is declared. This changes only what a caller is told when the driver
they were routed to cannot build the robot at all.

### Why the native check goes first

The two checks' populations overlap, so the order is a decision rather than a
free choice: `unitree_g1` is a lerobot teleoperator type *and* a natively driven
robot. For a name in that overlap the driver is what a `Robot()` caller asked
for, not an instruction to build the leader that teleoperates it.

Nothing observable turns on the order today — lerobot's robot registry knows
`unitree_g1` too, so it resolves and never reaches this handler — which is
exactly when a silent reordering would go unnoticed. It is pinned structurally
for that reason (`b3` below).

### One docstring corrected

`registry.get_driver` claimed `hardware.driver` is "absent everywhere in the
package registry, because every robot here is driven through lerobot". Measured:
two robots declare it, and six have native drivers. It now says that an absent
declaration means "no preference", not "no native driver" — the distinction this
fix turns on. The same stale belief appeared as an assertion message in
`test_driver_seam.py` and is corrected there.

## Tests

`tests/drivers/test_native_driver_named_when_lerobot_has_no_type.py`, 59 cells.
The four robots are named as literals so a narrowed rule fails loudly instead of
deselecting a case; a separate derived cell grades the rule itself, so a fifth
robot arriving in this position is caught there.

Pre-fix split, behavioural revert (the helper kept, its call site removed):
**14 failed / 45 passed**, and the headline failure is the defect itself:

```
assert 'DynamixelDriver' in "Unsupported robot type: 'vx300s'. Known lerobot
robot types: ['bi_openarm_follower', ...]"
```

| class | failed / cells | role |
|---|---|---|
| `TestANativelyDrivenRobotIsNamed` | 12 / 16 | regression |
| `TestTheOrderIsStated` | 2 / 2 | structural |
| `TestThePremise` | 0 / 16 | premise |
| `TestTheDerivedPopulationIsExactlyThese` | 0 / 2 | derived inventory |
| `TestTheGenericListingSurvivesWhereItIsTheRightAnswer` | 0 / 8 | over-reach |
| `TestTheTeleoperatorRefusalIsUnchanged` | 0 / 3 | control |
| `TestTheHelperReportsRatherThanRaises` | 0 / 6 | unit |
| `TestNothingElseChanged` | 0 / 6 | over-reach |

### Mutations

`own` counts cells in the new file; `sib` counts the 618 pre-existing cells in
`tests/drivers/`, `tests/test_driver_seam.py` and
`tests/test_leader_arm_is_not_a_robot.py`. `collected` was 59 on every row, so no
row hid a deselection. Source restored byte-identical after each.

| # | mutation | own | sib |
|---|---|---|---|
| b0 | control, no change | 0 | 0 |
| b1 | the call site removed (behavioural revert) | 14 | 14 |
| b2 | helper answers `None` for a robot that has a driver | 27 | 34 |
| b3 | native check moved **after** the teleoperator arm | 1 | 1 |
| b4 | native check moved **after** the generic listing | 14 | 14 |
| b5 | message drops the driver class name | 7 | 7 |
| b6 | message drops the `driver='strands'` call | 5 | 5 |
| b7 | message drops the robot asked for | 0 | 0 |
| b8 | helper raises for a robot with **no** driver (over-reach) | 11 | 17 |
| b9 | lookup skips `resolve_name` (alias miss) | 2 | 7 |
| b10 | message appends lerobot's listing anyway | 4 | 4 |

Ten mutations, nine detected, eight distinct firing sets. Notes on the rest:

- **b1 and b4 fire the identical 14 cells** (verified as sets, not counts): a
  guard placed after the `raise` is unreachable, so it is the same as absent.
- **b7 fires nothing, and that is correct.** The message interpolates the robot
  name twice — once in the summary and once in the `Robot(...)` retry — so
  dropping the first leaves the property "names the robot asked for" true.
- **b8 is the over-reach row**: making the helper *raise* rather than report
  breaks the 51-robot control set and trips 17 pre-existing cells, which is why
  it reports.
- **b6 originally fired 1**, because `"driver='strands'" in refusal` was
  satisfied by the second, registry-declaration mention that a caller cannot
  copy. The cell now pins the whole retry call, and the row fires 5.
- **b3 originally fired 0**: the ordering was ungraded until the structural pin
  was added.

## Gate

  tests_integ` (1626 files).
- mypy **24 == 24** against a worktree of the merge base, normalised message sets
  byte-identical in both directions, 0 attributable to the changed files
  (`checked 1617` vs `1616`, the +1 being the new test file).
- Paired run over an identical 612-file selection (every suite naming the driver
  seam, the registries, the refusals, plus every tree-walking guard), the new
  file excluded from both trees: identical **27133 collected** and
  `24 failed, 26911 passed, 198 skipped` on each, **24 identical failing ids with
  both `comm` directions empty**, distinct rootdirs. Of the 24: 17 need `awsiot`
  (absent here, declared in the `mesh-iot` extra — confirmed with `find_spec` and
  `tomllib`), 4 are a large-selection module-caching artifact that passes **13/13
  in isolation on both trees**, 3 are the lerobot-from-source
  `encode() takes 1 positional argument` drift.
- Rebased onto `7d80ff38` afterwards; `tests/drivers/`, the seam and the leader
  suite give **694 passed, 3 skipped** there, including the guards that base
  added.

No visual artifact: this changes a refusal string and the order two diagnostics
are consulted in, none of the policy, simulation, rendering, recording or asset
surfaces. The measured 55/4/51 table and the before/after refusals are the
evidence.

### DIFF
--- a/changelog.d/2817-native-driver-named-not-lerobots-vocabulary.md
+++ b/changelog.d/2817-native-driver-named-not-lerobots-vocabulary.md
@@ -1,0 +1,47 @@
+### Fixed: a robot with a native driver is named as one, not answered with lerobot's list
+
+Two registries answer "what builds this robot": lerobot's `RobotConfig`
+ChoiceRegistry and this package's native-driver registry. A robot can be in the
+second and not the first - that gap is what a native driver exists to close, and
+`drivers/reachy.py` records it in as many words: "the Reachy Mini has no lerobot
+robot type, so before this driver `mode="real"` raised `ValueError: Unsupported
+robot type: 'reachy_mini'`".
+
+The Reachy Mini also declares `hardware.driver="strands"`, so `resolve_driver`
+sends it to its driver and it never meets that refusal. Four robots the
+Dynamixel driver serves - `vx300s`, `wx250s`, `trossen_wxai` and `dynamixel_2r` -
+declare nothing, so the default routes them to lerobot, which has no robot type
+for any of them. They reached the generic listing of lerobot's sixteen robot
+types, and that listing never mentioned that this package ships the driver that
+builds them. A caller with no reason to guess `driver="strands"` was at a dead
+end, holding a list of sixteen names none of which was the robot they asked for.
+
+The site already had this shape for the other wrong entry point. A leader arm is
+a lerobot *teleoperator*, and `teleoperator._other_lerobot_kind_refusal` names it
+as one rather than listing follower types - its docstring is where "answering it
+with the names of the kind it is not answers the wrong question" is written down.
+`drivers.registry._native_driver_refusal` is that function's sibling for a
+natively driven robot: consulted at the same site, returning a reason or `None`
+the same way, so a name with no native driver keeps the listing that is the right
+answer for it. Of the fifty-five registry robots that reach this refusal, four
+now name their driver and fifty-one are unchanged - including `robotiq_2f85`,
+which has no driver of either kind.
+
+It is consulted before the teleoperator arm rather than after, because the two
+populations overlap: `unitree_g1` is a lerobot teleoperator type as well as a
+natively driven robot, and for a name in that overlap the driver is what a
+`Robot()` caller asked for, not an instruction to build the leader that
+teleoperates it. Nothing observable turns on that order today - lerobot's robot
+registry knows `unitree_g1` too, so it resolves and never arrives - which is why
+the ordering is pinned structurally.
+
+Resolution precedence is untouched and no registry entry gains a declaration.
+Which driver wins is unchanged: `koch` and `aloha` have both a working lerobot
+type and a native driver, and whether they should prefer the native one is a
+preference the registry is where to declare - `unitree_g1` shows that. This
+changes only what a caller is told when the driver they were routed to cannot
+build the robot at all. `get_driver`'s docstring, which claimed
+`hardware.driver` is "absent everywhere in the package registry, because every
+robot here is driven through lerobot", is corrected alongside: two robots declare
+it, and an absent declaration means "no preference" rather than "no native
+driver".
--- a/strands_robots/drivers/registry.py
+++ b/strands_robots/drivers/registry.py
@@ -154,3 +154,46 @@ def list_native_drivers() -> dict[str, str]:
         Canonical robot name -> driver class name, sorted by robot name.
     """
     return {name: cls.__name__ for name, cls in sorted(_NATIVE_DRIVERS.items())}
+
+
+def _native_driver_refusal(robot_type: str) -> str | None:
+    """Name ``robot_type``'s native driver, or ``None`` when it has none.
+
+    lerobot's ``RobotConfig`` registry and this package's native-driver registry
+    are two answers to "what builds this robot", and a robot may be in the
+    second and not the first: the Reachy Mini and the Trossen arms have no
+    lerobot robot type at all, which is the gap their native drivers exist to
+    close. When ``driver="lerobot"`` is chosen for such a robot -- and it is
+    chosen by default, because :func:`resolve_driver` falls back to
+    :data:`~strands_robots.drivers.base.DEFAULT_DRIVER` for a robot that
+    declares nothing -- lerobot cannot build it, and answering with the names of
+    the robots lerobot *does* know answers the wrong question. The driver that
+    builds this robot ships in this package, one keyword away.
+
+    The sibling of :func:`strands_robots.teleoperator._other_lerobot_kind_refusal`,
+    which does the same for the other kind of wrong entry point: that one names a
+    leader arm as a teleoperator, this one names a robot as natively driven. Both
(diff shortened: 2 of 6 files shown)


yinsong1986 (approved):
## Summary

Routes the `except KeyError` refusal in `hardware_robot._create_minimal_config` through a new `drivers.registry._native_driver_refusal` before the existing teleoperator check and the generic lerobot listing, so the four registry robots that lerobot cannot build but this package drives natively (`vx300s`, `wx250s`, `trossen_wxai`, `dynamixel_2r`) get a refusal naming their driver class and the exact `Robot(..., driver='strands', ...)` retry instead of a dead-end list of sixteen lerobot robot types. Resolution precedence is deliberately untouched — this changes only the diagnostic. The helper mirrors the report-or-`None` contract of its sibling `_other_lerobot_kind_refusal` at the same site, the native-before-teleoperator ordering is pinned structurally via AST (justified by the `unitree_g1` overlap, which nothing observable exercises today), and the stale `get_driver` docstring claim ("absent everywhere") is corrected to match the measured registry state.

## What's good

- Ran the new file plus `tests/drivers/`, `tests/test_driver_seam.py` and `tests/test_leader_arm_is_not_a_robot.py` against the head SHA: 59 + 662 passed, 35 skipped — the over-reach controls confirm the 51 no-driver robots keep the generic listing and `koch`/`aloha` are not diverted.
- The four names are literals with a separate derived-population cell grading the rule itself, so a narrowed guard fails loudly and a fifth robot arriving in this position is caught — plus premise cells so a pass cannot be vacuous.
- Mutation table in the description is honest about the two undetected/degenerate rows (b1==b4 as sets, b7 correctly vacuous) rather than padding the count.
- Refusal strings are plain ASCII, the helper stays out of `__all__`, and the changelog lands as a `changelog.d/2817-*.md` fragment per AGENTS.md step 3.

PR #2818, opened 2026-08-27
fix(registry): declare the joint count each robot's own asset has

cagataycali:
## Problem

Every robot in `registry/robots.json` carries a `joints` figure, and two discovery surfaces report it verbatim: `get_robot` returns it, and `list_robots` prints it in the `Joints` column an agent reads to size an action vector. Two entries disagreed with a sibling built from an **indistinguishable** compiled model - same joint names, same joint types, same actuator names - so one compiled shape was described by two different numbers.

| robot | declared | asset `njnt` | movable | actuators | sibling declares |
|---|---|---|---|---|---|
| `ur5e` | **8** | 6 | 6 | 6 | `ur10e` -> 6 |
| `unitree_a1` | **16** | 13 | 12 | 12 | `aliengo`, `go1` -> 13 |

`ur5e` is a six-axis arm with no gripper and no floating base. Its own description says so ("6-DOF industrial"), and `ur10e` declared `6`. `unitree_a1` shares its model with `aliengo` and `go1`, which both declared `13` - that model's `njnt` of twelve movable joints plus the floating base - and `anymal_b`/`anymal_c`, a separate quadruped family whose model has the same 13/12/12 shape, declare `13` too.

Both wrong figures were also rendered in user-facing tables, immediately adjacent to the siblings' correct rows: `docs/robots/arms.md:38` (`8`, one line below `ur10e`'s `6`) and `docs/robots/mobile.md:36` (`16`, between `aliengo` and `go1` at `13`).

## What is deliberately not decided

What `joints` counts registry-wide is left alone. `docs/robots/arms.md` says "Joint counts include any free joints / gripper actuators", which reads as MuJoCo's `njnt` and holds for `anymal_b`/`anymal_c` (13 against a 12-DOF description, the extra one being the base). But `panda` declares `7` against an `njnt` of 9 - the arm without its two finger joints - and `arx_l5`/`piper` both declare `11` against an `njnt` of 8. **Of the 50 registry robots whose asset loads, 22 declare a figure that is neither their `njnt` nor their movable-joint count.** Choosing one convention would rewrite those 22 numbers on a guess about what each was counting.

So the regression test grades a weaker property that needs no such decision:

> two robots whose compiled models are indistinguishable must be described by the same number, whatever that number is counting.

That holds under every convention above - `njnt`, movable joints, actuated DOF, hardware DOF - because the models agree on all of them. It is the in-family control that makes these two figures decidable while the registry-wide question stays open.

## The families

The assets group the 59 resolvable entries into six such families:

| family | declared | `njnt`/movable/`nu` | |
|---|---|---|---|
| `aliengo`, `go1`, `unitree_a1` | 13, 13, **16** | 13/12/12 | fixed here |
| `ur10e`, `ur5e` | 6, **8** | 6/6/6 | fixed here |
| `anymal_b`, `anymal_c` | 13, 13 | 13/12/12 | already agreed |
| `arx_l5`, `piper` | 11, 11 | 8/8/7 | already agreed |
| `toddlerbot_2xc`, `toddlerbot_2xm` | 45, 45 | 45/44/30 | already agreed |
| `vx300s`, `wx250s` | 19, 16 | 8/8/7 | **unresolved, recorded** |

`vx300s`/`wx250s` is the same defect, but unlike the two fixed above nothing there says which figure is right, or whether either is - a two-member family has no majority, and both descriptions state the same shape ("6-DOF + gripper"). It is recorded as an unresolved family with that reason rather than guessed at; removing it from that set is how the convention decision gets enforced.

## Tests

`tests/registry/test_asset_family_joint_counts.py`, 16 cells in two layers because the oracle is not available everywhere:

- **`TestEveryAssetFamilyAgrees`** reads `robots.json` alone, so it grades the property on any install - including one with no MuJoCo and no downloaded assets, which is where the registry is most often read.
- **`TestTheFrozenFamiliesStillMatchTheAssets`** re-derives the grouping from the real compiled models and asserts the frozen table still describes them, so the hand-maintained copy cannot drift into agreeing with a wrong registry. It skips where assets are absent rather than downloading them. This layer earned its place during development: it caught that my first frozen table was **missing the `aliengo`/`go1`/`unitree_a1` family entirely**, because I had walked the raw asset cache instead of `resolve_model_path`.
- **`TestTheFamilyTableIsUsable`** holds the premises - every family has >=2 members, every named robot is in the registry, the graded set is non-empty and still contains both corrected families, and no exemption names a family that is not declared.

Pre-fix split, with the registry values reverted and the test kept: **5 failed / 11 passed**, all five in the regression class, `0` of the 4 premise cells and `0` in the fidelity layer.

### Mutations

`collected` is recorded because two rows change it - a silently narrowed parametrization runs fewer tests while still reporting success.

| mutation | fails | collected | firing |
|---|---|---|---|
| control (no mutation) | 0 | 16 | - |
| `ur5e` reverted to 8 | 3 | 16 | agreement + value + discovery |
| `unitree_a1` reverted to 16 | 3 | 16 | agreement + value + discovery |
| both reverted (raw pre-fix) | 5 | 16 | union; `3 + 3 - 1` shared |
| family AGREES at the wrong value (`ur10e` -> 8) | 3 | 16 | value cells only |
| quadrupeds AGREE at the wrong value (all -> 16) | 4 | 16 | value cells only |
| UR family exempted | 1 | **15** | premise: must stay graded |
| UR family dropped from the table | 2 | **15** | premise + **fidelity layer** |
| exemption for a family that is not declared | 1 | 16 | premise |
| typo in a family member name | 3 | 16 | premise + fidelity + agreement |

9 mutations, **9 distinct firing sets**, control clean, both files restored byte-identical.

Two rows carry the design. The **"family agrees at the wrong value"** rows show agreement alone is satisfied by moving the majority to the outlier, so the value assertions are load-bearing rather than decorative. The **"family dropped from the table"** row is why the fidelity layer exists: a hand-maintained table that omits a family is silent on it, and only re-deriving from the assets notices.

## Gate

- Paired selection of **560 files** (all of `tests/registry/`, every suite walking the tree / parsing source / reading docstrings, `robots.json`, `docs/robots/` or `changelog.d/`, plus everything naming the affected robots), the new file excluded from both trees and every path verified present on both: identical **`23 failed, 24492 passed, 195 skipped`** on branch and base, **23 identical failing ids, `comm` empty in both directions**, distinct rootdirs.
- Those 23 are pre-existing and environmental: **17** need `awsiot`/`awscrt` (`find_spec` False locally; declared in the `[mesh-iot]` extra, which CI installs), **4** in `test_recording_frame_loss_is_not_tolerated.py` are a large-selection module-caching artifact (**13 passed in isolation** on both trees), **2** are lerobot-from-source `TypeError: encode()` drift.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine worktree of the same base, normalized message sets identical in both directions, **0 attributable** to the changed files.
- `ruff check` and `ruff format --check` clean over 1620 files.
- Registry + changelog suites on the rebased base: **841 passed**.

No visual artifact: this changes two reported integers and two documentation tables. Nothing renders, simulates, records or moves differently - the measured tables above are the evidence.

### DIFF
--- a/changelog.d/2818-registry-joint-counts-match-their-asset.md
+++ b/changelog.d/2818-registry-joint-counts-match-their-asset.md
@@ -1,0 +1,51 @@
+### Fixed: two registry entries declare the joint count their own asset has
+
+Every robot in `registry/robots.json` carries a `joints` figure, and two
+discovery surfaces report it verbatim: `get_robot` returns it, and `list_robots`
+prints it in the `Joints` column an agent reads to size an action vector. Two
+entries disagreed with a sibling built from an indistinguishable model - same
+joint names, same joint types, same actuator names - so one compiled shape was
+described by two different numbers:
+
+```
+robot        declared   asset njnt   movable   actuators   sibling declares
+ur5e         8          6            6         6           ur10e -> 6
+unitree_a1   16         13           12        12          aliengo, go1 -> 13
+```
+
+`ur5e` is a six-axis arm with no gripper and no floating base; its own
+description says so ("6-DOF industrial"), and `ur10e` - whose compiled model is
+byte-identical in joint names, joint types and actuator names - declared `6`.
+`unitree_a1` shares its model with `aliengo` and `go1`, which both declared `13`,
+the model's `njnt` of twelve movable joints plus the floating base;
+`anymal_b` and `anymal_c`, a separate quadruped family whose model has the same
+13/12/12 shape, declare `13` too. Both figures are now what the shared asset has,
+and `docs/robots/arms.md` and `docs/robots/mobile.md` - which rendered `8` and
+`16` in tables directly above and below the siblings' correct rows - agree.
+
+What `joints` counts registry-wide is deliberately not settled here.
+`docs/robots/arms.md` says "Joint counts include any free joints / gripper
+actuators", which reads as MuJoCo's `njnt` and holds for `anymal_b`/`anymal_c`
+(13 against a 12-DOF description, the extra one being the base). But `panda`
+declares `7` against an `njnt` of 9 - the arm without its two finger joints - and
+`arx_l5`/`piper` both declare `11` against an `njnt` of 8. Of the 50 registry
+robots whose asset loads, 22 declare a figure that is neither their `njnt` nor
+their movable-joint count. Choosing one convention would rewrite those 22 numbers
+on a guess about what each was counting, so the regression test grades a weaker
+property that needs no such decision: two robots whose compiled models are
+indistinguishable must be described by the same number, whatever that number is
+counting. That holds under every convention above, because the models agree on
+all of them - and it is the in-family control that makes these two figures
+decidable while the registry-wide question stays open.
+
+The assets group the 59 resolvable entries into six such families. Four already
+agreed; the two that did not are corrected here, except `vx300s`/`wx250s`, which
+declare `19` and `16` against an `njnt` of 8 while describing the same shape as
+each other ("6-DOF + gripper"). Unlike the two fixed above, nothing there says
+which figure is right, or whether either is - a two-member family has no majority
+- so the pair is recorded as an unresolved family with that reason rather than
+guessed at, and removing it from that set is how the convention decision gets
+enforced. The test re-derives the grouping from the compiled models and fails if
+a family it names no longer groups together, or if the assets group robots that
+no family names, so the frozen table cannot drift into agreeing with a wrong
+registry.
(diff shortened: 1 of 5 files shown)


yinsong1986 (approved):
## Summary

Corrects two registry `joints` figures that disagreed with siblings compiled from indistinguishable MJCF models: `ur5e` 8 -> 6 (matching `ur10e` and its own "6-DOF industrial" description) and `unitree_a1` 16 -> 13 (matching `aliengo`/`go1` and the asset's `njnt`), updates the two docs tables that rendered the wrong figures, and pins the fix with `tests/registry/test_asset_family_joint_counts.py` - a registry-only agreement layer that runs on any install, a fidelity layer that re-derives the family grouping from the compiled models (skipping where assets are absent), and premise cells that keep the frozen table from going vacuous. The registry-wide question of what `joints` counts is deliberately left open, with the `vx300s`/`wx250s` disagreement recorded as an unresolved family rather than guessed at. Nothing in the tree asserts or hardcodes the old values (`embodiments.json`'s own note for `ur5e` already said "6 named joints"), so the change is a pure data correction with no downstream breakage.

## What's good

- MuJoCo enum/type reads in the fidelity layer are value-based (`int(model.jnt_type[i])`), per the enum-comparison rule.
- Test names describe behaviour, not provenance; both new files are pure ASCII; no host paths.
- The mutation table in the description shows the value assertions are load-bearing (agreement alone is satisfiable at the wrong number), and the exemption set is itself graded so the `vx300s`/`wx250s` deferral cannot silently widen.

PR #2819, opened 2026-08-27
fix(mesh): announce a safety audit record that can never be written

cagataycali:
## What is wrong

`SensorLoopsMixin.publish_safety_event` sends one event two ways: to the mesh
wire, and to the local audit log. Its own docstring records which copy carries
what happened:

> `severity`: The real severity. It reaches the audit record only - the wire copy
> is uniformly `"info"` (issue #272) - **so the audit record is the only surviving
> copy**

`session._report_unencodable_payload` states the rule both halves are held to:

> Every transport's `put` is fire-and-forget and `MeshTransport.put` scopes that
> tolerance to a TRANSIENT failure - a closed session, a dropped broker, a
> socket-level write - which the next tick retries. A payload the JSON encoder
> refuses is not transient: it fails identically forever [...]
> **Reporting it at DEBUG left the two halves of one call disagreeing.**

That raised the wire half to ERROR. The audit half still reported its own loss at
`logger.debug`, so the two halves of one call were still disagreeing - and the
half still at DEBUG is the one holding the only copy of the real severity.

## Measured

Driven through the public method with the audit write failing the way a full
volume makes it fail (`OSError(28)`), and separately with a payload the encoder
refuses:

| half that loses the event | reported at | an operator at the default level (WARNING) sees |
|---|---|---|
| wire (payload the encoder refuses) | `ERROR` | yes |
| audit (write to disk fails) | `DEBUG` | **no** |

For `publish_safety_event(event_type="remote_resume_applied", severity="critical", ...)`
with the audit write failing, before this change:

```
raised out of publish_safety_event : None
wire put() calls                   : 1        (severity "info")
records at >= WARNING              : 0
all records                        : [('DEBUG', '[mesh] alice: audit log write failed: [Errno 28] No space left on device')]
```

`severity="critical"` reached nothing an operator would see, and the record that
was lost is the only one that ever carried that word.

An audit write is permanent in a stronger sense than the encoder case the rule was
written for. `publish_safety_event` is called once per safety transition - all 17
call sites in `mesh/core.py` are single calls at a single transition, and no loop
re-publishes a safety event - so there is no "next tick" to retry it at all.

`remote_resume_applied` is emitted by `_on_safety_resume` **after**
`self._estop_lockout.clear()` has already run, so the lost record is a fleet that
resumed motion with nothing on the forensic trail above DEBUG to say so.

## Change

`logger.debug` -> `logger.error`, and the report now names the event type and the
real severity, because the wire copy names neither. The docstring records why the
level is what it is and cites the same rule rather than restating it.

Deliberately unchanged: the call still does not raise (fire-and-forget is the
contract), the wire copy still goes out with its uniform `"info"` severity, the
caller's payload is still left unedited, and the `except Exception` stays wide for
the reason the surrounding code gives - an audit backend's raise set is not
enumerable at this seam.

## Why nothing caught it

Nine of the eleven `publish_safety_event` call sites in `mesh/core.py` wrap the
call in `except (TypeError, ValueError, OSError)` "best-effort ... must never
block the safety path", so the callers had already priced an audit failure as
survivable. What none of them could see is how the failure is *announced*: that
is decided inside `publish_safety_event`, and no test drove its `except` branch.

## Pre-fix split

Source reverted, tests kept: **6 failed / 9 passed**.

|---|---|---|
| `TestAPermanentlyLostSafetyRecordIsAnnounced` | regression | 5 / 5 |
| `TestTheTwoHalvesOfOneCallAgree` | regression + wire-half premise | 1 / 2 |
| `TestWhatIsUnchanged` | over-reach controls | 0 / 6 |
| `TestThePremisesHold` | premises | 0 / 2 |

Headline failures: `assert {10} == {40}` (DEBUG vs ERROR) and
`assert 'critical' in ''` - the real severity was named nowhere.

## Break table

8 mutations, 8 distinct firing sets, control clean, source restored
md5-identical after every row. `sib` counts failures in six pre-existing safety
suites (211 cells).

| mutation | fires | collected | sib |
|---|---|---|---|
| control (unmodified) | 0 | 15 | 0 |
| level back to DEBUG (the raw revert) | 6 | 15 | 0 |
| level softened to WARNING | 5 | 15 | 0 |
| report drops the event type | 1 | 15 | 0 |
| report drops the real severity | 1 | 15 | 0 |
| report drops the peer | 1 | 15 | 0 |
| report drops the reason the write failed | 1 | 15 | 0 |
| handler narrowed so an `OSError` escapes the safety path | 10 | 15 | 3 |
| over-reach: an ordinary publish also announces at ERROR | 3 | 15 | 0 |

The narrowed-handler row is the useful one for scope: it fires the
`test_the_call_still_does_not_raise` control **and trips three pre-existing
safety cells**, which is the evidence that the width of the catch is load-bearing
and this change must not touch it. The over-reach row fires the
healthy-publish-is-silent control, so "ERROR must not become routine" is graded
rather than asserted.

## Gate

- `ruff check` / `ruff format --check` clean over `strands_robots tests tests_integ` (1630 files).
  merge-base, normalized message sets identical in both directions, 0 attributable.
- Paired run over an identical 373-file selection (all of `tests/mesh/` plus every
  suite that walks the tree, parses source, reads docstrings, or names
  `publish_safety_event` / `log_safety_event` / `_report_unencodable_payload`), the
  new file excluded from both trees: identical **11261 collected** and
  `21 failed, 11152 passed, 64 skipped, 24 errors` on each, **45 identical failing
  ids with an empty `comm` diff in both directions**. All 45 are
  `tests/mesh/test_iot_*` needing `awsiot` / `awscrt`, both `find_spec=False` here
  and both declared in the `[mesh-iot]` extra.
- `tests/mesh/test_zenoh_transport_security.py` is excluded from the pair (it opens
  a real session) and verified alone: `9 passed, 2 skipped` on both trees.
- Changed file plus five sibling safety suites: **191 passed** on the rebased tree.

No visual artifact: this changes a reporting level and a message, not a policy,
simulation, rendering, recording or asset. The measured before/after log capture
above is the evidence.

### DIFF
--- a/changelog.d/2819-safety-audit-write-failure-is-announced.md
+++ b/changelog.d/2819-safety-audit-write-failure-is-announced.md
@@ -1,0 +1,23 @@
+### Fixed: a safety audit record that can never be written is announced rather than logged at DEBUG
+
+`SensorLoopsMixin.publish_safety_event` sends one event two ways - to the mesh
+wire and to the local audit log - and its own docstring records that the
+`severity` argument "reaches the audit record only": the wire copy is uniformly
+`info` (issue #272), so the audit record is the only copy that carries what
+actually happened.
+
+`session._report_unencodable_payload` states the rule both halves are held to. A
+transport's fire-and-forget tolerance is scoped to a TRANSIENT failure - "a
+closed session, a dropped broker, a socket-level write - which the next tick
+retries" - and a loss that no retry can undo is reported at ERROR instead,
+"because reporting it at DEBUG left the two halves of one call disagreeing". That
+raised the wire half. The audit half still reported its own loss at DEBUG, below
+the default level, so an operator saw nothing at all when the only surviving copy
+of a `critical` event failed to reach the disk.
+
+An audit write is permanent in the stronger sense: a safety event is published
+once, at one lockout transition, so there is no later tick to retry it. The
+failure is now reported at ERROR and names the event type and the real severity,
+because the wire copy names neither. Everything else about the call is unchanged
+- it still does not raise, the wire copy still goes out with its uniform `info`
+severity, and the caller's payload is still left unedited.
--- a/strands_robots/mesh/sensors.py
+++ b/strands_robots/mesh/sensors.py
@@ -720,6 +720,17 @@ class SensorLoopsMixin:
         value that tripped a limit, the distance that closed), and every other
         record this mixin sends to the wire is coerced before it goes.
 
+        A failed audit write is reported at ERROR for the same reason
+        :func:`~strands_robots.mesh.session._report_unencodable_payload` gives for
+        the wire half: a transport fire-and-forget tolerance is scoped to a TRANSIENT
+        failure that the next tick retries, and this is not that. A safety event is
+        published once at one transition, so there is no later tick, and the audit
+        copy is the only one carrying the real severity. Reporting the loss at DEBUG
+        left the two halves of one call disagreeing about how a permanently lost
+        safety record is announced, which is the disagreement that report was raised
+        to ERROR to close. The report names the event type and the real severity
+        because the wire copy names neither.
+
         Coerced into a copy rather than in place, so the caller keeps the mapping
         they passed unedited - the same reason :func:`_coerce_record` replaces a
         nested container instead of editing the provider's own.
@@ -770,5 +781,13 @@ class SensorLoopsMixin:
                 # first so the keys that process decided win a name collision.
                 payload={**record, "severity": severity},
             )
-        except Exception as exc:  # noqa: BLE001
-            logger.debug("[mesh] %s: audit log write failed: %s", self.peer_id, exc)
+        except Exception as exc:  # noqa: BLE001 - see the reporting-level note in the docstring
+            logger.error(
+                "[mesh] %s: the audit record for a %s safety event could not be written, so the "
+                "only copy carrying its real severity (%s) is lost - the wire copy is uniformly "
+                "info-level and a safety event is published once, so no later call rewrites it: %s",
+                self.peer_id,
+                event_type,
+                severity,
+                exc,
+            )
--- a/tests/mesh/test_safety_audit_write_failure_is_reported.py
+++ b/tests/mesh/test_safety_audit_write_failure_is_reported.py
@@ -1,0 +1,245 @@
+"""A safety record that can never be written is announced, not logged at DEBUG.
+
+:meth:`~strands_robots.mesh.sensors.SensorLoopsMixin.publish_safety_event` sends
+one event two ways: to the wire, and to the local audit log. Its own docstring
+records that the ``severity`` parameter "reaches the audit record only - the wire
+copy is uniformly ``info``", so the audit record is the only copy that carries
+what actually happened.
+
+:func:`~strands_robots.mesh.session._report_unencodable_payload` states the rule
+both halves are held to. A transport's fire-and-forget tolerance is scoped to a
+TRANSIENT failure - "a closed session, a dropped broker, a socket-level write -
+which the next tick retries" - and a permanent loss is reported at ERROR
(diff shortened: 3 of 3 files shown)


yinsong1986 (dismissed):
## Summary

This PR raises the reporting level of a failed audit write in `SensorLoopsMixin.publish_safety_event` from DEBUG to ERROR, closing the asymmetry with the wire half: `session._report_unencodable_payload` already reports a permanent loss at ERROR, while the audit half - the only copy carrying the event's real severity, since the wire copy is uniformly `info` per issue #272 - announced its own permanent loss below the default level. The new report names the event type, the real severity, the peer, and the underlying reason, all of which the wire copy omits. The accompanying test suite pins the escalation from both directions (the loss is announced at ERROR and visible at the WARNING default; a healthy publish stays silent so ERROR does not become routine) and pins what must not change: the call still never raises, the wire copy still goes out on audit failure, the wire severity stays uniformly `info`, the caller's payload is left unedited, and the stopped-mesh early return still says nothing. Verified against the head SHA: the new suite plus the two adjacent suites (`test_sensor_readers.py`, `test_unencodable_payload_is_reported_not_dropped.py`) pass, ruff check/format are clean, and both files are plain ASCII. The pre-existing `except Exception` scope is unchanged - only the report inside it moved - and a log message is not a wire format or persisted schema, so nothing here is a one-way door.

## What's good

- The negative-space tests (`TestWhatIsUnchanged`) pin the fire-and-forget contract alongside the escalation, so the fix cannot regress into breaking the safety publish path.
- `test_a_healthy_publish_announces_nothing` guards against ERROR becoming routine noise, and `TestThePremisesHold` asserts the cited rule in `_report_unencodable_payload` still says what the docstring claims, so the citation cannot rot silently.
- Test class and function names describe behaviour, not provenance, per AGENTS.md > Review Learnings (#85) > Testing.

One non-blocking note: the diff carries no `changelog.d/2819-*.md` fragment; the CI-enforced fragment check (AGENTS.md > PR Workflow step 3) will flag it, and it is self-clearing with a one-file addition.

yinsong1986 (approved):
## Summary

This PR raises the reporting level of a failed audit write in `SensorLoopsMixin.publish_safety_event` from DEBUG to ERROR, closing the remaining half of the asymmetry `session._report_unencodable_payload` already closed for the wire: the audit record is the only copy carrying the event's real severity (the wire copy is uniformly `info` per issue #272), and a safety event is published once per transition, so a failed write is a permanent loss with no later tick to retry it. The new report names the event type, the real severity, the peer, and the reason - none of which the wire copy carries. Everything load-bearing is deliberately unchanged and pinned as such: the call still never raises (fire-and-forget), the wide `except Exception` stays (the PR's break table shows narrowing it trips three pre-existing safety cells), the wire copy still goes out with its uniform `info` severity, and the caller's payload is left unedited. Verified against the head SHA: the new 15-test suite passes, the adjacent `test_unencodable_payload_is_reported_not_dropped.py` and `test_sensor_readers.py` suites pass (114 tests), ruff check/format are clean, both new/changed files are plain ASCII, the format string's four placeholders match its four arguments, and every seam the tests patch (`core.put`, `sensors.log_safety_event`, `session._SESSION`, `_unencodable_topics_warned`) exists as named. A log level and message are neither a wire format nor a persisted schema, so nothing here is a one-way door; the changelog fragment (`changelog.d/2819-*.md`) is present per PR Workflow step 3.

## What's good

- The over-reach controls (`test_a_healthy_publish_announces_nothing`, `test_a_stopped_mesh_still_publishes_nothing`) grade that ERROR does not become routine, so the escalation cannot quietly widen.
- `TestThePremisesHold` asserts the cited rule in `_report_unencodable_payload` still says what the docstring claims, so the citation cannot rot silently.
- Test names describe behaviour, not provenance, and the docstring cites the shared rule rather than restating it, per AGENTS.md conventions.

PR #2820, opened 2026-08-27
fix(tools): refuse the Feetech broadcast where serial_tool is about to read one reply

cagataycali:
## Problem

The Feetech Protocol 1 ID byte carries every unicast address and one address that is no servo: `0xfe` is the broadcast. `serial_tool` bounded `motor_id` at `254` for a reason keyed to the field's *width*:

```
"motor_id": (1, 254, "the packet carries the ID in one byte, and 255 is the frame header")
```

That is true of the byte and says nothing about which of its values a servo can hold. So `action="feetech_ping"` accepted the broadcast, wrote `FF FF FE 02 01 FE`, read ten bytes back and reported them as one motor's answer.

Every servo on the bus receives that frame, and because `PING` is answered every one of them replies at once. On a half-duplex bus those replies collide, so the bytes read back belong to no single servo and the ID quoted beside them belongs to none either. With one servo attached it would appear to work, which is worse: it reads like a discovery feature.

The same intent through the two surfaces in this repository, measured:

| Surface | `motor_id=254` |
|---|---|
| `serial_tool` `action="feetech_ping"` | accepted -> `FF FF FE 02 01 FE` on the wire, reported `Feetech Motor 254 responded: FFFF010200FC` |
| `drivers.feetech.protocol.ping_packet` | `ValueError: motor_id=0xfe is the broadcast; this instruction expects a reply` |

`drivers/feetech/protocol.py` (landed as #2786) already owns this question. It declares `MAX_UNICAST_ID = 0xFD` ("Highest ID a specific servo may hold. 0xFE is the broadcast") and refuses the broadcast from `build_packet` unless a caller opts in `allow_broadcast=True` "for `SYNC_WRITE` and other reply-less instructions that legitimately target the broadcast ID. Every other caller passes `False` so a slip that addresses a broadcast where a reply is required is refused here rather than by a servo that never answers."

## Change

`serial_tool` now reads `BROADCAST_ID` and `MAX_UNICAST_ID` from that codec rather than restating them, and applies the same rule at its own boundary, before the port is opened.

A reply-less write keeps the whole range. `feetech_position` and `feetech_velocity` never read a status packet, so addressing every servo with one frame means exactly what it says. Only an action that reads a reply is held to a single servo, which mirrors the codec's own per-instruction `allow_broadcast` distinction.

| `motor_id` | `feetech_ping` | `feetech_position` | `feetech_velocity` |
|---|---|---|---|
| 1 .. 252 | accept | accept | accept |
| 253 (`MAX_UNICAST_ID`) | accept | accept | accept |
| 254 (`BROADCAST_ID`) | **refuse** | accept | accept |
| 255 | refuse (field bound) | refuse | refuse |

The refusal:

```
feetech_ping: motor_id 0xfe is the broadcast, which every servo on the bus answers
at once, so no single reply can be read back; address one servo in [1, 253], got 254.
```

The reply-expecting set is derived from the tool's own body - an action that both reads the port back and consumes `motor_id` - so an action added later that reads a reply for a caller-supplied ID is held to the same rule the hour it lands. Of the eight actions, `read` / `send_read` / `monitor` read without addressing a servo and `feetech_position` / `feetech_velocity` address one without reading, which is what makes the discriminator precise rather than a naming heuristic.

## Why nothing caught it

`tests/tools/test_serial_tool_numeric_domain.py` sweeps `motor_id` refusals over `(0, 255, 256, 300, -1, True, 2.5, "1")`. The broadcast is *inside* the ID byte, so it is absent from a sweep of values outside it. Its acceptance sweep listed `254` under the name `test_an_addressable_motor_id_still_reaches_the_wire` - and the broadcast is the one in-range value that is not addressable, so that name recorded the opposite. That case now sweeps `253`, the highest ID a servo may hold, and what the broadcast may and may not be used for is graded in the new file.

## Tests

`tests/tools/test_feetech_broadcast_is_not_a_reply_address.py`, 24 cells. Reverting the source and keeping the tests gives **12 failed / 12 passed**, and no over-reach control is among the failures:

| Class | Role | pre-fix failed/cells |
|---|---|---|
| `TestAReplyExpectingActionRefusesTheBroadcast` | regression | 3/3 |
| `TestTheRefusalNamesWhatWentWrong` | regression (message) | 5/5 |
| `TestAReplyLessWriteStillAddressesTheWholeBus` | over-reach control | **0/7** |
| `TestTheAddressSpaceIsTheCodecs` | premise + agreement | 3/7 |
| `TestEveryReplyExpectingActionIsHeldToASingleServo` | derived drift guard | 1/2 |
| **total** | | **12/24** |

The headline pre-fix failures are the tool's own output: `assert 'feetech_ping' in 'Feetech Motor 254 responded: FFFF010200FC'` and `assert [b'\xff\xff\xfe\x02\x01\xfe'] == []`.

Mutation table. `new` counts failing cells in the new file, `sib` in the 162 pre-existing cells of `test_serial_tool_numeric_domain.py` / `test_feetech_velocity_direction_bit.py` / `test_feetech_status_packet_framing.py`. `collected` was 24 on every row, so no row hid a deselection.

| Row | Mutation | new | sib |
|---|---|---|---|
| b1 | the broadcast guard is dropped | 8 | 0 |
| b2 | the reply-expecting set is emptied | 9 | 0 |
| b3 | the reply-less writes are held to a reply too (over-reach) | 5 | 0 |
| b4 | the field ceiling narrowed to `MAX_UNICAST_ID` for every action | 7 | 0 |
| b5 | the broadcast is refused only after the port is opened | 1 | 0 |
| b6 | the comparison widened to `>=` the broadcast | 0 | 0 |
| b7 | the refusal stops naming the range that can answer | 1 | 0 |
| b8 | the refusal stops saying why no reply can be read | 1 | 0 |
| b9 | the field reason reverts to the byte width alone | 1 | 0 |
| b10 | the address space is hardcoded instead of read from the codec | 1 | 0 |

11 rows, 9 firing, **9 distinct firing sets, no two rows share one**, control clean, source restored byte-identically.

Rows worth reading:

- **b1 is a strict subset of b2**, and the one cell b2 adds is `test_the_module_holds_every_action_that_reads_a_reply` - the derived guard is what separates "the rule was removed" from "the rule is intact and reaches nothing".
- **b3 and b4 are the two tempting over-reaches** and both fire the reply-less-write controls. b4 is the uniform narrowing - holding *every* action to a single servo - which would refuse a broadcast write that means what it says.
- **b5 fires exactly one cell**, the placement assert. Refusing after the port is open is not equivalent: the guard belongs where every sibling option is already checked.
- **b6 fires 0 and is a measured semantic no-op**: the field bound already refuses anything above 254, so only the broadcast can reach the comparison and `>=` is `==` over the reachable domain.
- **b10 fires 1 and closed a real gap in my own test.** The value-agreement cell passes when the constants are hardcoded to the same numbers - equal values are exactly what the previous state looked like - so a structural cell reads the import instead.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1630 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a worktree of the merge base, normalized message sets byte-identical in both directions, 0 attributable to the changed files.
- Paired run over an identical 462-file selection (all of `tests/tools/` and `tests/drivers/`, plus every suite that walks the tree, parses source, reads docstrings or reads `docs/`), with the new file excluded from both trees: `20 failed, 21837 passed, 78 skipped` on **both**, the same 20 failing ids, `comm` empty in both directions. 17 need `awsiot` (declared in the `[mesh-iot]` extra, absent here) and 3 are the lerobot-from-source `encode() takes 1 positional argument` drift.

No visual artifact: this changes an option domain and a refusal message, not a policy, simulation, rendering, recording or robot asset. The measured acceptance table and the two-surface comparison above are the evidence.

## Docs

`docs/hardware/tools.md` documented the same byte-width reason in its `motor_id` row. The row and a new paragraph now state which address is a servo, which is the whole bus, and why only a reply-expecting action is held to one.

### DIFF
--- a/changelog.d/2820-feetech-broadcast-is-not-a-reply-address.md
+++ b/changelog.d/2820-feetech-broadcast-is-not-a-reply-address.md
@@ -1,0 +1,32 @@
+### Fixed: a broadcast is refused where the tool is about to read one servo's reply
+
+The Feetech Protocol 1 ID byte carries every unicast address and one address that
+is no servo: `0xfe` is the broadcast. `serial_tool` bounded `motor_id` at 254 for
+a reason keyed to the field's width - "the packet carries the ID in one byte, and
+255 is the frame header" - which is true of the byte and says nothing about which
+of its values a servo can hold. So `action="feetech_ping"` accepted the
+broadcast: the tool wrote `FF FF FE 02 01 FE`, read ten bytes back and reported
+`Feetech Motor 254 responded: <hex>`. Every servo on the bus receives that frame
+and, because `PING` is answered, every one of them replies at once; on a
+half-duplex bus those replies collide, so the bytes read back belong to no single
+servo and the ID quoted beside them belongs to none either. With a single servo
+attached it would appear to work, which is worse - it reads like a discovery
+feature.
+
+The address space is not this module's to decide.
+`strands_robots.drivers.feetech.protocol` declares `BROADCAST_ID` and
+`MAX_UNICAST_ID` ("Highest ID a specific servo may hold. 0xFE is the broadcast")
+and already refuses the same intent from `build_packet` unless a caller passes
+`allow_broadcast=True` "for `SYNC_WRITE` and other reply-less instructions that
+legitimately target the broadcast ID". The tool now reads those two constants
+rather than restating them, and applies that same rule at its own boundary,
+before the port is opened.
+
+A reply-less write keeps the whole range. `feetech_position` and
+`feetech_velocity` never read a status packet, so addressing every servo with one
+frame means exactly what it says and is still accepted; only an action that reads
+a reply is held to a single servo. That set is derived from the tool's own body -
+an action both reading the port back and consuming `motor_id` - so an action
+added later that reads a reply for a caller-supplied ID is held to the same rule
+the hour it lands. The refusal names the broadcast, why no single reply can come
+back, and the range that can answer.
--- a/docs/hardware/tools.md
+++ b/docs/hardware/tools.md
@@ -92,9 +92,21 @@ speed in the opposite direction, and `velocity=32768` read as magnitude zero,
 stopping a servo the caller had just asked to run. `position` was already inside
 that limit at 4095; `velocity` now is too.
 
+`motor_id` takes more than the byte width for a different reason: the ID byte
+carries one address that is no servo. `0xfe` is the broadcast, which every servo
+on the bus receives, and for an instruction that expects a reply every servo
+answers at once - on a half-duplex bus those replies collide, so what
+`action="feetech_ping"` reads back belongs to no single servo. A reply-less
+write to the broadcast means what it says and is still accepted, so
+`feetech_position` and `feetech_velocity` take the whole `[1, 254]`; only a
+reply-expecting action is held to a single servo. The Protocol 1 codec applies
+the same rule to the frames it builds
+(`build_packet(..., allow_broadcast=False)`), so the tool and the driver cannot
+disagree about which address is a servo and which is the whole bus.
+
 | Option | Accepted | Why the bound is where it is |
 |--------|----------|------------------------------|
-| `motor_id` | integer in `[1, 254]` | the frame carries the ID in one byte, and `255` is the header value |
+| `motor_id` | integer in `[1, 254]`, or `[1, 253]` for an action that reads a reply | the frame carries the ID in one byte, of which `0xfd` is the highest a servo may hold and `0xfe` is the broadcast, while `0xff` is the header value |
 | `position` | integer in `[0, 4095]` | `Goal_Position` is a 12-bit register - the same full scale the reported angle divides by |
 | `velocity` | integer in `[0, 32767]` | `Goal_Velocity` is sign-magnitude with bit 15 the direction bit, so a larger magnitude commands the opposite direction |
 | `baudrate` | positive integer | pyserial coerces rather than checks, so `2.7` opens the port at 2 baud |
--- a/strands_robots/tools/serial_tool.py
+++ b/strands_robots/tools/serial_tool.py
@@ -21,6 +21,15 @@ opposite direction, and ``velocity=32768`` read as magnitude zero -- stopping a
 servo the caller had just asked to run -- both reported as success quoting the
 number supplied.
 
+``motor_id`` takes more than the byte width too, because the ID byte carries
+one address that is not a servo: ``0xfe`` is the broadcast, which every servo
+on the bus receives and, for an instruction that expects a reply, every servo
+answers at once. On a half-duplex bus those replies collide, so the bytes
+``action="feetech_ping"`` reads back belong to no single servo -- and it
+reported them as one, quoting an ID no servo holds. A reply-less write to the
+broadcast means exactly what it says and stays accepted; the highest address a
+servo may hold is ``0xfd``, which is what a reply-expecting action is held to.
+
 ``baudrate`` and ``read_bytes`` are coerced rather than checked by pyserial
 (``2.7`` becomes 2 baud, and a non-positive ``read_bytes`` reads nothing and
 reports success, which is indistinguishable from a timed-out read on a healthy
@@ -42,6 +51,7 @@ import serial
 import serial.tools.list_ports
 from strands import tool
 
+from strands_robots.drivers.feetech.protocol import BROADCAST_ID, MAX_UNICAST_ID
 from strands_robots.utils import (
     finite_number_error,
     non_negative_count_error,
@@ -64,7 +74,12 @@ _MAX_MAGNITUDE = (1 << _DIRECTION_BIT) - 1
 # surface uses; only the ceiling is decided here, because it is a property of
 # the packet field this module encodes into.
 _REGISTER_FIELDS: dict[str, tuple[int, int, str]] = {
-    "motor_id": (1, 254, "the packet carries the ID in one byte, and 255 is the frame header"),
+    "motor_id": (
+        1,
+        BROADCAST_ID,
+        f"the packet carries the ID in one byte, of which {MAX_UNICAST_ID:#x} is the highest a servo "
+        f"may hold and {BROADCAST_ID:#x} is the broadcast, while 0xff is the frame header",
+    ),
     "position": (
         0,
         4095,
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

`serial_tool` accepted `motor_id=254` for `feetech_ping` because the field bound was keyed to the ID byte's width, so the tool would write a broadcast PING, read back whatever collided on the half-duplex bus, and report it as one servo's answer. This PR imports `BROADCAST_ID` / `MAX_UNICAST_ID` from `strands_robots.drivers.feetech.protocol` (the codec that already owns this rule via `allow_broadcast=False`) and refuses the broadcast in `_motor_id_error`, at the same pre-port-open boundary where every other option domain is checked. The reply-less writes (`feetech_position`, `feetech_velocity`) deliberately keep the full `[1, 254]`, mirroring the codec's per-instruction distinction. Verified locally at head `4f2bb342`: the new file plus the numeric-domain suite pass (101 tests); reverting `serial_tool.py` to the merge base reproduces the claimed 12 failed / 12 passed split, with all over-reach controls (broadcast writes stay accepted) among the passes; `ruff check` / `ruff format --check` clean on the changed files; the tree-wide graders for tool parameter descriptions, ASCII tool-result strings, and `Raises:` completeness all pass.

## What's good

- Constants are imported from the codec rather than restated, and `TestTheAddressSpaceIsTheCodecs` pins the import structurally (not just value agreement), closing the drift the PR's own b10 mutation exposed.
- The reply-expecting set is derived from the tool's body (reads the port back AND consumes `motor_id`), so a future reply-reading action is graded on arrival; the over-reach direction is pinned in both the derived guard and the reply-less-write controls.
- The refusal is placed with the other option-domain checks so no port is opened and no frame reaches the wire on a refused call; both properties are asserted directly.
- Changelog fragment, docs table, and the `Args:` docstring entry all updated in the same PR, per AGENTS.md rules 3 and 13; all new user-facing strings are ASCII.

cagataycali:
## The vendor draws this line at "reads a reply", not at "addresses a servo" - measured

This PR refuses `motor_id=254` for `feetech_ping` and leaves the full `[1, 254]` domain on
`feetech_position` / `feetech_velocity`. Recording the measurement behind that split, because the
obvious alternative - refuse the broadcast for every motor-addressed action - looks tidier and is
the wrong line.

`scservo_sdk` is the oracle, and it guards exactly two of the four operations these actions perform:

| `scservo_sdk` operation | guard present | source |
|---|---|---|
| `ping` | **yes** | `if scs_id >= BROADCAST_ID: return model_number, COMM_NOT_AVAILABLE, error` |
| `readTxRx` | **yes** | `if scs_id >= BROADCAST_ID: return data, COMM_NOT_AVAILABLE, 0` |
| `writeTxRx` | **no** | no id check; builds the packet and calls `txRxPacket` |
| `write2ByteTxRx` | **no** | `return self.writeTxRx(port, scs_id, address, 2, data_write)` |

And the reason the write needs no guard is in `txRxPacket`, which carries a broadcast branch of its
own: `if (txpacket[PKT_ID] == BROADCAST_ID):` returns without waiting for a reply. A broadcast write
is a supported operation with a defined meaning - drive every servo, expect nothing back. A
broadcast *read* has no defined meaning, which is why the two read paths refuse it up front.

So the vendor's line is drawn by **whether the operation reads a reply**, not by whether it names a
single servo. `feetech_ping` is this tool's only action that writes a request and then reads one
reply back (`ser.write(pkt)`, `sleep`, `ser.read(10)`), which is why it is the only action narrowed
here.

Two consequences worth having on the record:

1. **A uniform narrowing would refuse an operation the vendor supports.** `feetech_position` at the
   broadcast is a broadcast register write - `writeTxRx` with no guard, `txRxPacket` returning early.
   Refusing it removes a real capability and is not what the SDK does.
2. **A claim that the SDK guards all three operations is false for the write.** Neither `writeTxRx`
   nor `write2ByteTxRx` carries a `scs_id >= BROADCAST_ID` check, so a test asserting the vendor
   refuses a single-register write to the broadcast would be pinning an agreement that does not
   exist in that direction.

The break table already ships the uniform-narrowing variant as a row (it fires 7 cells, including
the reply-less-write controls). This comment is the vendor-side half of that argument, so the scope
decision reads as measured rather than chosen.

cagataycali:
Reviewing this after arriving at the same defect independently (#2822, which I am closing as the duplicate - this one is earlier and better scoped). Three measurements, one of which is a real cost.

## Your scoping is right and mine was an over-reach

I refused the broadcast for all three motor-addressed actions. `_REPLY_EXPECTING_ACTIONS = frozenset({"feetech_ping"})` is the correct line, and the reason is in this tool's own code: `feetech_position` and `feetech_velocity` do `ser.write(packet)` and never read a status packet, so a broadcast write there means exactly what it says - which is what the vendor's own `SYNC_WRITE` uses `BROADCAST_ID` for. Only `feetech_ping` reads a reply back, and that is the one a broadcast cannot answer.

Worth noting for the record that the SDK's three refusal sites are not all reply-less writes:

| `protocol_packet_handler.py` | operation | refuses `scs_id >= BROADCAST_ID` |
|---|---|---|
| `:214` | ping | yes |
| `:245` | single write **expecting a status packet** (`writeTxRx`) | yes |
| `:286` | single read | yes |
| `:435`, `:454` | `SYNC_READ` / `SYNC_WRITE` | uses the broadcast |

So the SDK refuses a broadcast for a write too - but only for the `TxRx` form that waits for a reply, which is not the form this tool uses. Your set is correct for this tool as written; if `feetech_position` ever starts reading its status packet back, it joins `_REPLY_EXPECTING_ACTIONS`. A one-line comment saying that would make the set self-maintaining.

## The import costs 1137 modules and numpy

```python
from strands_robots.drivers.feetech.protocol import BROADCAST_ID, MAX_UNICAST_ID
```

Measured on this branch, at module scope:

```
values: 0xfe/0xfd   import 1697 ms   new modules 1137   numpy pulled in: True
```

Naming the submodule does not avoid it: Python executes `strands_robots/drivers/__init__.py` first, which registers every driver, and `drivers/g1.py` there imports `strands_robots.tools.g1`. So a lazily-loaded agent tool now pays 1.7 s and numpy for two integers.

Single-sourcing them is the right instinct - two modules disagreeing about this ceiling is how the defect existed. The cheaper form is to restate the two values with the reason in a comment and pin the agreement in a test, which can afford the import:

```python
def test_the_ceiling_is_the_codecs_max_unicast_id(self) -> None:
    from strands_robots.drivers.feetech import MAX_UNICAST_ID as codec_ceiling

    _, ceiling, _ = serial_mod._REGISTER_FIELDS["motor_id"]
    assert ceiling == codec_ceiling
```

That keeps the two from drifting without putting the driver registry on the tool's import path. Your call - it is a cost, not a defect.

## docs/hardware/tools.md

You caught `docs/hardware/tools.md:97` and I did not. That table is the surface a caller reads, so it mattered.

## Verified on your head

`tests/tools/` is **3216 passed, 9 skipped** on `4f2bb342`.

One suggestion for the sibling suite. `tests/tools/test_serial_tool_numeric_domain.py` parametrized 254 as an addressable ID inside a class named `TestARegisterFieldIsNeverSilentlyTruncated`; you changed that case, which is right. Since your scoping keeps 254 valid for the two write actions, the case reads correctly as-is - but a cell asserting that `feetech_position` with `motor_id=254` still reaches the wire would pin the half you deliberately kept, so a later reader cannot narrow the field bound to 253 and call it a tightening. That is the mutation I would most want caught here: on your branch, narrowing `_REGISTER_FIELDS["motor_id"]` to `MAX_UNICAST_ID` is behaviour-changing and, as far as I can see, not currently refused by a test.

PR #2821, opened 2026-08-27
fix(hardware): do not close the devices under a teleop loop that is still writing

cagataycali:
## The defect

`stop_teleoperate` reports its join outcome honestly. A leader whose `get_action()` is still blocking - a serial read on a wedged bus is the ordinary case a join budget exists for - yields `status="error"` with `stopped: False`, and that branch **deliberately leaves the attached devices connected**. Its docstring says why, and names the precedent:

> The attached devices are disconnected only once the loop has joined. Tearing the bus down under a live writer is what `G1Driver.cleanup` refuses for the same reason.

`cleanup()` called `stop_teleoperate()`, discarded the envelope, and closed the motors bus anyway. So the two functions disagreed about the same state, and `cleanup()` undid the protection `stop_teleoperate` had just provided. Measured on a one-camera arm whose leader is wedged, with the port's exclusivity modelled:

| call | bus closed | port held | teleop thread alive |
|---|---|---|---|
| `stop_teleoperate()` alone | no | yes | yes |
| `cleanup()` | **yes** | **no** | yes |

## Why this is a defect and not a trade-off

Closing there is worse than deferring on **both** of the counts `cleanup()` cares about.

**The release does not hold.** `send_action` re-opens the robot lazily on a command that finds it disconnected (`hardware_robot.py:2854`), and the live loop's next write goes through it. Nothing remains to close the port again - `cleanup()` is terminal and the executor is already down:

| after the live write | before | after |
|---|---|---|
| bus `connect_calls` | 2 (re-opened) | 1 |
| port held afterwards | yes, for the process lifetime | yes, never closed |

That is exactly the harm the comment above `_disconnect_devices()` already describes, and the teleop loop is **first in its own list** of command sources:

> `send_action` re-opens the robot lazily on a command that finds it disconnected, so a port closed before the teleop loop, the task executor, the mesh and the ROS bridge have stopped can be re-opened behind this teardown.

**The torque disable is undone.** `_disconnect_devices` prefers the driver's own `disconnect()` because "that is where a follower disables torque and releases the gripper". The loop's write lands *after* it, leaving the arm energised at a fresh commanded position.

## The change

`cleanup()` reads the outcome and holds the devices when the loop is still running, recording the reason and the remedy at ERROR - the only report it has, since it returns `None`.

The outcome is read from `stopped`, not from `status`. The status is derived from the session counters, so a session whose every frame errored reports `"error"` after a perfectly clean join; a caller keying on it would read a healthy teardown as a live loop. `_stop_reported_stopped` lives in `teleop_mixin`, which owns the envelope, and defaults to "stopped" for an envelope that carries no flag (the "no active teleoperation" reply has nothing to stop).

Deliberately unchanged: the software teardown still runs, a clean stop still closes every device exactly once, and a `stop_teleoperate` that **raises** leaves the outcome unknown and keeps the existing contract that "a teleop teardown failure must not block the rest of hardware cleanup" (pinned by `TestCleanupFailSoft`).

## Why nothing caught it

`test_hardware_cleanup_disconnects` is the suite dedicated to this property, and it has a class named `TestTheDevicesCloseLast` whose docstring is *"A port closed while any command source is still live gets re-opened."* It grades the mesh, the ROS bridge and the draining executor. The teleop loop - the one command source that is a live **writer** to the bus being closed, and the only one whose shutdown the ordering does not guarantee - appears in that file **0 times**.

## Verification

Pre-fix split (behavioural revert - the gate dropped, the reader kept): **9 failed / 19 passed**

|---|---|---|
| `TestTheCloseIsDeferredUnderALiveWriter` | regression | 9 / 9 |
| `TestThePremiseStopTeleoperateAlreadyDeclinesTheTeardown` | premise | 0 / 2 |
| `TestDeferringTheCloseCostsNothingElse` | over-reach | 0 / 2 |
| `TestACleanStopStillClosesEverything` | over-reach | 0 / 3 |
| `TestAnUnknownOutcomeKeepsThePreExistingContract` | control | 0 / 2 |
| `TestTheOutcomeIsReadFromStoppedNotFromStatus` | reader unit | 0 / 9 |
| `TestTheDecisionIsSingleSourced` | structural | 0 / 1 |

A raw revert is a collection error instead, because the tests import the new helper.

Mutation table - 11 rows, control clean, all 10 detected, **10 distinct firing sets**, `collected` stable at 28, source restored byte-identical. `sib` is failures in the five pre-existing suites (`test_hardware_cleanup_disconnects`, `test_stop_teleoperate_reports_the_join_outcome`, `test_hardware_robot_lifecycle`, `test_teleop`, `test_hardware_cleanup_survives_a_failed_robot_init`):

| row | mutation | fires | sib |
|---|---|---|---|
| b0 | control - no mutation | 0 | 0 |
| b1 | the gate dropped: close unconditionally (the defect) | 9 | 0 |
| b2 | reader keys on `status` instead of `stopped` | 9 | 0 |
| b3 | reader defaults to `False` when the flag is absent | 5 | 0 |
| b4 | reader reads the first `json` block without checking for the flag | 1 | 0 |
| b5 | the deferral is silent (ERROR dropped) | 2 | 0 |
| b6 | the ERROR names the state but not the remedy | 1 | 0 |
| b7 | a `stop_teleoperate` that raises is treated as a live loop (over-reach) | 1 | 0 |
| b8 | the gate also skips the mesh / bridge / executor teardown (over-reach) | 3 | 0 |
| b9 | `cleanup()` re-derives the outcome from the thread handle (drift) | 1 | 0 |
| b10 | reader coerces with a truthiness test on the whole payload | 10 | 0 |

`sib` is 0 on every row, which is the measured form of the section above: none of the pre-existing suites can see any of this. b7 is the scope decision - the pre-existing fail-soft test asserts only `stopped == ["mesh"]`, so it never grades the device close on the raise path, and the control in this file is what does.

Gate, paired against `efd18075` on an identical 441-file selection (every suite naming teleop, cleanup, the hardware wrapper, or walking the tree / reading source and docstrings), the new file excluded from both trees and every path verified present on both:

| | branch | base |
|---|---|---|
| collected | 21008 | 21008 |
| result | 20 failed, 20915 passed, 73 skipped | 20 failed, 20915 passed, 73 skipped |
| failing ids | 20 | 20 |

`comm` is empty in both directions. The 20 are environmental and identical on both trees: 17 in `tests/mesh/test_iot_*` need `awsiot`/`awscrt` (declared in the `[mesh-iot]` extra, absent here, installed in CI) and 3 in `tests/training/test_lerobot*` are the lerobot-from-source `encode() takes 1 positional argument` drift.

`ruff check` and `ruff format --check` clean over 1630 files. mypy 24 == 24 against a worktree of the merge base, normalized message sets byte-identical in both directions, 0 attributable to the changed files.

No visual artifact: this changes a teardown ordering and its report, not a policy, simulation, render, recording or asset. The measured tables above are the evidence.

### DIFF
--- a/changelog.d/2821-cleanup-defers-the-close-under-a-live-teleop-writer.md
+++ b/changelog.d/2821-cleanup-defers-the-close-under-a-live-teleop-writer.md
@@ -1,0 +1,46 @@
+### Fixed: `cleanup()` does not close the devices under a teleop loop that is still writing
+
+`stop_teleoperate` reports its join outcome honestly: a leader whose
+`get_action()` is still blocking - a serial read on a wedged bus is the ordinary
+case a join budget exists for - yields `status="error"` with `stopped: False`,
+and that branch deliberately leaves the attached devices connected. Its
+docstring says why, and names the precedent: "Tearing the bus down under a live
+writer is what `G1Driver.cleanup` refuses for the same reason." `cleanup()`
+called it, discarded the envelope, and closed the motors bus anyway - so the two
+functions disagreed about the same state and `cleanup()` undid the protection
+`stop_teleoperate` had just provided.
+
+Closing there is worse than deferring on both of the counts `cleanup()` cares
+about, which is what makes this a defect rather than a trade-off. The release
+does not hold: `send_action` re-opens the robot lazily on a command that finds it
+disconnected, the live loop's next write goes through it, and nothing remains to
+close the port again because `cleanup()` is terminal and the executor is already
+down - exactly the harm the comment above `_disconnect_devices()` describes, with
+the teleop loop first in its own list of command sources. And the torque disable
+is undone: `_disconnect_devices` prefers the driver's own `disconnect()` because
+that is where torque disable and gripper release live, and the loop's write then
+lands after it. Measured on a one-camera arm whose leader is wedged, with the
+port's exclusivity modelled:
+
+```
+                                        before      after
+stop_teleoperate() alone: bus closed    no          no
+cleanup(): bus closed                   yes         no
+cleanup(): port held afterwards         no          yes
+cleanup(): driver disconnect() calls    1           0
+teleop thread alive throughout          yes         yes
+bus connect_calls after the live write  2           1
+command lands after torque disable      yes         no
+deferral reported                       (nothing)   ERROR + remedy
+```
+
+`cleanup()` now reads the outcome through a `_stop_reported_stopped` helper on
+the envelope's `stopped` key and holds the devices when the loop is still
+running, recording the reason and the remedy at ERROR - the only report it has,
+since it returns `None`. The key is `stopped` rather than `status`, because the
+status is derived from the session counters: a session whose every frame errored
+reports `"error"` after a perfectly clean join, and a caller keying on it would
+read a healthy teardown as a live loop. The software teardown is unchanged, a
+clean stop still closes every device exactly once, and a `stop_teleoperate` that
+raises leaves the outcome unknown and keeps the existing "a teleop teardown
+failure must not block the rest of hardware cleanup" contract.
--- a/strands_robots/hardware_robot.py
+++ b/strands_robots/hardware_robot.py
@@ -39,7 +39,7 @@ from strands.types.tools import ToolResult, ToolSpec, ToolUse
 from strands_robots._serial_discovery import describe_serial_candidates, scan_serial_devices
 from strands_robots.bus_access import read_observation, write_action
 from strands_robots.ros_telemetry import ROS2_SYSTEM_INSTALL_HINT
-from strands_robots.teleop_mixin import TeleopMixin
+from strands_robots.teleop_mixin import TeleopMixin, _stop_reported_stopped
 from strands_robots.utils import (
     boolean_flag_error,
     dds_domain_id_error,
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

This PR closes a real disagreement between `stop_teleoperate` and `cleanup()` about the same state: `stop_teleoperate` already declines to close the devices when the teleop loop outlives its join budget (a wedged leader's blocking `get_action()`), but `cleanup()` discarded that envelope and closed the motors bus anyway - after which the loop's next write re-opened the port through `send_action`'s lazy connect (so the release did not hold) and landed after the driver's `disconnect()` had disabled torque. The fix threads the join outcome through a small private helper (`_stop_reported_stopped`) that reads the envelope's `stopped` flag - correctly keying on `stopped` rather than the counter-derived `status`, so a session whose every frame errored but joined cleanly still closes - and defers `_disconnect_devices()` only on a positive report of a live loop, recording the reason and remedy at ERROR. A `stop_teleoperate` that raises keeps the pre-existing warn-and-continue contract (outcome unknown, devices still closed), the software teardown (shutdown latch, executor, mesh, ROS bridge) is unchanged in both branches, and the logged remedy (re-join then `cleanup()` again) is itself pinned to work by a test. No public API, wire format, or env-var surface changes; the helper is private and not exported.

## What's good

- The 28-test regression suite covers the premise, both directions of the deferral, the unknown-outcome (raise) path, the `stopped`-vs-`status` discrimination table, and an AST guard that `cleanup()` does not re-derive the outcome from the thread handle.
- Consequences are asserted rather than calls: port exclusivity is modelled, so the tests pin "the write did not have to re-open the port" and "no command landed after torque disable" directly.
- The failure direction is conservative: a race where the loop exits just after a failed join leaves devices open with an actionable ERROR, never a close under a live writer.
- Changelog fragment named per convention; ruff check/format clean; log strings ASCII; test class names describe behaviour per AGENTS.md.

## Verification suggestions

Beyond CI: `pytest tests/test_cleanup_defers_the_close_under_a_live_teleop_writer.py tests/test_hardware_cleanup_disconnects.py tests/test_stop_teleoperate_reports_the_join_outcome.py tests/test_teleop.py tests/test_hardware_robot_lifecycle.py -q` - all pass at this head (187 tests).

PR #2822, opened 2026-08-27
fix(tools/serial_tool): refuse the bus broadcast as a single servo's ID

cagataycali:
## What is wrong

`serial_tool` bounds `motor_id` at 254 and gives the byte width as the reason:

```
"motor_id": (1, 254, "the packet carries the ID in one byte, and 255 is the frame header"),
```

Two of the 256 values that byte holds address no servo at all. `0xFF` is the frame header, and `0xFE` is the bus broadcast, which every servo on the bus reads. Only the header was excluded, so the broadcast was accepted as if it named one motor.

All three actions that take `motor_id` write to a single servo and expect its reply, so 254 was not a wider version of the request but a different one.

## Measured

Driven through the tool with a recording stand-in for `serial.Serial`, so the bytes that reach the wire are observable:

| `motor_id` | before | after |
|---|---|---|
| 1, 2, 100, 252 | frame written, ID byte carries the value | unchanged |
| **253** | frame written | unchanged |
| **254** | **frame written with ID byte `0xFE`; `status="success"`** | refused, port never opened |
| 255 | refused (frame header) | unchanged |

Before, each of the three actions reported success for 254:

```
feetech_position  -> success  "Feetech Motor 254 -> Position 100 (8.8 deg)"
feetech_velocity  -> success  "Feetech Motor 254 -> Velocity 100"
feetech_ping      -> writes FF FF FE 02 01 ... then reads a reply
```

`feetech_position` therefore drove every joint of the arm to the same angle while the message named one motor; `feetech_ping` made every servo answer at once, colliding on the half-duplex bus, and read the collision back as one reply.

After:

```
feetech_position: motor_id must be at most 253 (0xFE is the bus broadcast, so a frame
carrying it is read by every servo rather than the one this message names, and 0xFF is
the frame header), got 254.
```

## The vendor draws the same line

`scservo_sdk` declares `BROADCAST_ID = 0xFE` and refuses it for exactly the three operations these actions perform:

| SDK site | operation | on `scs_id >= BROADCAST_ID` |
|---|---|---|
| `protocol_packet_handler.py:214` | ping | `return model_number, COMM_NOT_AVAILABLE, error` |
| `protocol_packet_handler.py:245` | single register write | `return COMM_NOT_AVAILABLE` |
| `protocol_packet_handler.py:286` | single register read | `return data, COMM_NOT_AVAILABLE, 0` |
| `:187` | any broadcast tx | returns without waiting for a reply |
| `:435`, `:454` | `SYNC_READ` / `SYNC_WRITE` | the only places it *uses* the broadcast |

The two sync instructions are the legitimate use of a broadcast, and this tool has no action for either: its actions are `list_ports`, `send`, `read`, `send_read`, `monitor`, `feetech_position`, `feetech_velocity`, `feetech_ping`.

`strands_robots.drivers.feetech` already states the same ceiling as `MAX_UNICAST_ID = 0xFD` beside its own `BROADCAST_ID = 0xFE`, and its `_validate_id(motor_id, allow_broadcast)` refuses the broadcast unless an instruction opts in. So the package answered this question in one module and not in the other.

## Restated rather than imported, and why

Importing that constant would single-source it, but `from strands_robots.drivers.feetech import MAX_UNICAST_ID` executes `drivers/__init__`, which registers every driver: measured at **1097 new modules and 1.8 s, including numpy**, for one integer, on a tool the package loads lazily. `drivers/g1.py` also imports `strands_robots.tools.g1`, so the direction is already occupied.

The value is restated with the reason in a comment, and `TestTheToolAndTheCodecAgree` asserts the tool's ceiling equals `MAX_UNICAST_ID` and that the excluded value equals `BROADCAST_ID`. A test can afford the import; the tool cannot.

## No capability is removed

`send` writes the bytes it is given and claims to address nothing, so a caller who wants a broadcast frame still has one. `TestBroadcastingIsStillReachable` grades that a hand-built `FF FF FE ...` frame reaches the wire, and that `send` does not read `motor_id` at all. It holds before and after, so the refusal cannot later be widened into a ban on addressing the bus.

## Why nothing caught it

`tests/tools/test_serial_tool_numeric_domain.py` pinned the defect. Inside a class named `TestARegisterFieldIsNeverSilentlyTruncated`:

```python
@pytest.mark.parametrize("motor_id", (1, 2, 6, 254))
def test_an_addressable_motor_id_still_reaches_the_wire(...)
```

254 is not an address. That case now uses 253, the largest ID that names one servo; the broadcast case moved to the new file, which names it for what it is. This is the same shape as the `Goal_Velocity` ceiling in #2747, where the suite likewise parametrized the out-of-domain value as valid.

## Tests

Pre-fix split, source reverted and tests kept: **18 failed / 46 passed**.

| class | failed / cells | role |
|---|---|---|
| `TestTheBroadcastIsNotAUnicastTarget` | 16 / 16 | regression |
| `TestTheToolAndTheCodecAgree` | 2 / 2 | drift guard |
| `TestEveryUnicastIdIsStillAccepted` | 0 / 30 | control |
| `TestTheFrameHeaderIsStillRefused` | 0 / 6 | control |
| `TestBroadcastingIsStillReachable` | 0 / 2 | over-reach control |
| `TestThisSuiteStatesTheWireSemanticsIndependently` | 0 / 2 | premise |
| `TestTheVendorSdkDrawsTheSameLine` | 0 / 2 | premise |
| `TestTheContractCoversEveryMotorAddressedAction` | 0 / 4 | derived non-vacuity |

Mutation table. `mine` is the new file, `coll` its collected count, `sib` the four pre-existing `serial_tool` suites (245 cells):

| row | mine | coll | sib | mutation |
|---|---|---|---|---|
| b0 | 0 | 64 | 0 | control, no mutation |
| b1 | 17 | 64 | 0 | ceiling reverted to the broadcast (the defect) |
| b2 | 11 | 64 | 1 | over-reach: ceiling narrowed one further, to 252 |
| b3 | 3 | 64 | 0 | reason no longer names the broadcast |
| b4 | 3 | 64 | 0 | reason no longer names the frame header |
| b5 | 0 | 64 | 0 | ceiling written as a bare literal, not the named constant |
| b6 | 3 | 64 | 54 | options checked after the port is opened |
| b7 | 1 | **47** | 8 | one action stops declaring `motor_id` |
| b8 | 1 | 64 | 0 | documented range reverted to `[1, 254]` |

Seven firing rows, seven distinct firing-ID sets, control clean, source restored byte-identical.

Two rows are worth reading together. **b2** is the over-reach check: narrowing one value too far costs the 253 controls here *and* trips the pre-existing suite's `[253]` case, so the boundary is graded from both sides. **b7** fires only one cell but drops `collected` from 64 to 47 - removing an action from the option map silently deselects 17 cases, which is why the swept set is pinned against locally-named actions as well as derived from the module.

**b5 fires nothing, correctly**: a bare `0xFD` literal is the same value, and the tests grade the ceiling's value rather than its spelling. The named constant is there to carry the reason, not to change behaviour.

## Gate

- `ruff check` and `ruff format --check` over `strands_robots tests tests_integ`: clean, 1632 files.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a worktree of the merge base, normalized message sets identical in both directions, nothing attributable to the changed files.
- Paired run over an identical 546-file selection (all of `tests/tools/`, plus every suite that walks the tree, parses source, or reads docstrings, `Args:`/`Raises:` or `changelog.d/`), the new file excluded from both trees: identical **25469 collected** and `20 failed, 25341 passed, 108 skipped` on each, with the failing-ID sets identical in both directions. Of the 20, 17 need `awsiot`/`awscrt` (`find_spec` is `None` here; declared in the `[mesh-iot]` extra, which CI installs) and 3 are the known lerobot-from-source `encode()` drift.
- The changed files plus the two guards that landed on `main` during this work: 310 passed.

No visual artifact: this changes a refusal and a documented range, not a policy, a simulation, a render, a recording or an asset. The measured before/after table above is the evidence.

### DIFF
--- a/changelog.d/2822-feetech-motor-id-excludes-the-broadcast.md
+++ b/changelog.d/2822-feetech-motor-id-excludes-the-broadcast.md
@@ -1,0 +1,30 @@
+### Fixed: `serial_tool` no longer accepts the Feetech bus broadcast as a single servo's ID
+
+`motor_id` was bounded at 254 and gave the byte width as the reason ("the packet carries the
+ID in one byte, and 255 is the frame header"). Two of the 256 values that byte holds address
+no servo at all: `0xFF` is the frame header, and `0xFE` is the bus broadcast, which every
+servo on the bus reads. Only the header was excluded.
+
+All three of the tool's motor-addressed actions write to one servo and expect its reply, so
+`motor_id=254` was not a wider version of the request but a different one. `feetech_position`
+put a broadcast frame on the wire, driving every joint of the arm to the same angle while
+reporting `Feetech Motor 254 -> Position ...` as though one had moved; `feetech_velocity` did
+the same with a speed; and `feetech_ping` made every servo answer at once, colliding on the
+half-duplex bus, then read the collision back as one reply. Each returned `status="success"`.
+
+The vendor draws the line in the same place. `scservo_sdk` declares `BROADCAST_ID = 0xFE` and
+returns `COMM_NOT_AVAILABLE` for `scs_id >= BROADCAST_ID` in each of the three operations
+these actions perform - a ping, a single register read and a single register write - reserving
+the broadcast for `SYNC_READ` / `SYNC_WRITE`, which this tool has no action for.
+`strands_robots.drivers.feetech` states the same ceiling as `MAX_UNICAST_ID` beside its own
+`BROADCAST_ID`; the value is restated in the tool rather than imported because importing it
+executes `drivers/__init__`, which registers every driver and pulls in numpy, and a test pins
+the two together so they cannot drift.
+
+The ceiling is now 253, the largest ID that names one servo. The 253 values below it are
+unchanged, 255 is still refused for the reason it always was, and broadcasting is unchanged:
+`send` writes the bytes it is given and claims to address nothing, so a caller who wants a
+broadcast frame still has one. The tool's documented range is corrected to `[1, 253]`.
+
+`tests/tools/test_serial_tool_numeric_domain.py` parametrized 254 as an addressable ID, in a
+class named `TestARegisterFieldIsNeverSilentlyTruncated`; that case now uses 253.
--- a/strands_robots/tools/serial_tool.py
+++ b/strands_robots/tools/serial_tool.py
@@ -21,6 +21,17 @@ opposite direction, and ``velocity=32768`` read as magnitude zero -- stopping a
 servo the caller had just asked to run -- both reported as success quoting the
 number supplied.
 
+``motor_id`` takes more than the byte width to bound for a different reason: two
+of the 256 values the byte holds do not address a servo at all. ``0xFF`` is the
+frame header, and ``0xFE`` is the bus broadcast -- a frame carrying it is read by
+every servo on the bus. Each action here that takes ``motor_id`` addresses one
+servo and expects its reply, so a broadcast is not a wider version of the request
+but a different one: ``feetech_position`` drove every joint of the arm to the same
+angle while reporting ``Feetech Motor 254 -> Position ...`` as though it had moved
+one, and ``feetech_ping`` made every servo answer at once, colliding on the
+half-duplex bus. Broadcasting is still reachable through ``send``, which writes
+the bytes it is given and claims to address nothing.
+
 ``baudrate`` and ``read_bytes`` are coerced rather than checked by pyserial
 (``2.7`` becomes 2 baud, and a non-positive ``read_bytes`` reads nothing and
 reports success, which is indistinguishable from a timed-out read on a healthy
@@ -58,13 +69,30 @@ _DIRECTION_BIT = 15
 # Largest magnitude either register carries with ``_DIRECTION_BIT`` still clear.
 _MAX_MAGNITUDE = (1 << _DIRECTION_BIT) - 1
 
+# Largest *unicast* ID, which is the ceiling rather than the byte's maximum: 0xFE
+# is the bus broadcast and 0xFF is the frame header, so neither names one servo.
+# The vendor SDK draws the same line -- ``scservo_sdk`` returns ``COMM_NOT_AVAILABLE``
+# for ``scs_id >= BROADCAST_ID`` on a ping, on a single read and on a single write,
+# and reserves the broadcast for SYNC_READ / SYNC_WRITE, which this module has no
+# action for. ``strands_robots.drivers.feetech`` states it as ``MAX_UNICAST_ID``;
+# the value is restated here rather than imported because importing it executes
+# ``drivers/__init__``, which registers every driver and pulls in numpy -- 1097
+# modules and 1.8 s for one integer, on a tool the package loads lazily. A test
+# pins the two values together so they cannot drift apart.
+_MAX_UNICAST_MOTOR_ID = 0xFD
+
 # Inclusive bounds and the reason for each ceiling, keyed by the parameter that
 # carries the field. The floor and the type are delegated to the shared count
 # domains so an off-type or negative value is reported in the words every other
 # surface uses; only the ceiling is decided here, because it is a property of
 # the packet field this module encodes into.
 _REGISTER_FIELDS: dict[str, tuple[int, int, str]] = {
-    "motor_id": (1, 254, "the packet carries the ID in one byte, and 255 is the frame header"),
+    "motor_id": (
+        1,
+        _MAX_UNICAST_MOTOR_ID,
+        "0xFE is the bus broadcast, so a frame carrying it is read by every servo "
+        "rather than the one this message names, and 0xFF is the frame header",
+    ),
     "position": (
         0,
         4095,
@@ -211,7 +239,8 @@ def serial_tool(
             pyserial's non-blocking mode (return what is already buffered)
         data: String data to send
         hex_data: Hex string data to send (e.g., "FF FF 01 04 03 00 64 92")
-        motor_id: Motor ID for Feetech commands; an integer in [1, 254]
+        motor_id: Motor ID for Feetech commands; an integer in [1, 253]. 254 is the
+            bus broadcast and 255 the frame header, so neither addresses one servo
         position: Target position for Feetech motors; an integer in [0, 4095]
         velocity: Target velocity for Feetech motors; an integer in [0, 65535]
         read_bytes: Number of bytes to read; a positive integer
--- a/tests/tools/test_feetech_motor_id_excludes_the_broadcast.py
+++ b/tests/tools/test_feetech_motor_id_excludes_the_broadcast.py
@@ -44,6 +44,7 @@ value, the one that means "everybody".
 
 from __future__ import annotations
 
+import inspect
 from typing import Any
 
 import pytest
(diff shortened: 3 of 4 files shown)


cagataycali:
## This fixes the same defect as #2820, and the two conflict

Both branches refuse the Feetech `0xFE` broadcast as a `motor_id`. Measured against their merge base `b8222418`:

| measurement | value |
|---|---|
| `git merge-tree` conflict markers | **4** |
| shared modified paths | `strands_robots/tools/serial_tool.py`, `tests/tools/test_serial_tool_numeric_domain.py` |
| new test file, this branch | `tests/tools/test_feetech_motor_id_excludes_the_broadcast.py` |
| new test file, #2820 | `tests/tools/test_feetech_broadcast_is_not_a_reply_address.py` |

Because the two chose different names for the new test, the duplicate-addition sweep reports `unique-additions` over the queue: its key is the set of **added** paths, and this pair collides on **modified** ones. So nothing on the board flags the pair automatically.

### The scope decisions differ, and that is the part worth deciding

They are not two spellings of one fix. They disagree about whether a broadcast write is ever legitimate:

- **#2820** keeps the broadcast **accepted** for a reply-less write and refuses it only for a reply-expecting action, holding those to `0xFD`. It reads `BROADCAST_ID` and `MAX_UNICAST_ID` from the Feetech codec merged in #2786 rather than restating them.
- **This branch** refuses the broadcast for **every** action taking `motor_id`, leaving `send` as the only route to it, and defines the boundary locally.

The second is the wider refusal. Whether a reply-less broadcast write should survive is a real call about the tool's contract, not a detail - a sync-write to every servo is the one case where the broadcast is the point.

### Current state of the two

| | #2820 | this branch |
|---|---|---|
| `reviewDecision` | APPROVED | REVIEW_REQUIRED |
| `call-test-lint` | pending | **failure** |
| changelog fragment | present | **absent** |

Both have auto-merge armed, so whichever lands first leaves the other conflicting on those two files.

Given #2820 is approved, carries a fragment, and sources the two boundary constants from the codec instead of re-deriving them, the cheaper reconciliation is to land #2820 and close this one - unless the all-actions refusal is the intended contract, in which case that scope argument belongs on #2820 where the ceiling is chosen.

yinsong1986 (dismissed):
## Summary

Narrows the `motor_id` domain on `serial_tool`'s three motor-addressed actions (`feetech_ping` / `feetech_position` / `feetech_velocity`) from `[1, 254]` to `[1, 253]`, excluding `0xFE` -- the Feetech bus broadcast -- which previously passed validation and drove every servo on the bus while the tool reported success for a single motor. The refusal is loud (error dict naming the broadcast and the new ceiling `at most 253`), the tool docstring's `Args:` entry is updated in the same change so the schema an agent reads matches what the tool enforces, and broadcasting stays reachable through `send`, so no capability is removed. The new ceiling `0xFD` matches `strands_robots.drivers.feetech.MAX_UNICAST_ID` and the vendor SDK's own `scs_id >= BROADCAST_ID` refusal; the value is restated rather than imported (documented lazy-load rationale) and a test pins the two together so they cannot drift.

## What's good

- The regression suite derives its swept action set from `_OPTIONS_BY_ACTION` rather than a hand-copied list, and separately pins the known set so a dropped action fails instead of silently deselecting -- both AGENTS.md testing patterns.
- Controls assert the 253 unicast IDs below the broadcast still reach the wire with the exact ID in the frame, and that `send` still emits a hand-built broadcast frame, so the narrowing cannot be widened into a capability loss unnoticed.
- Refusal happens before the port is opened (pinned), consistent with the module's validate-before-side-effect contract.
- Verified locally at head `04a16686`: 139 passed, 2 skipped (optional `scservo_sdk` not installed).

cagataycali:
## Measured: the lint failure, the ungraded suite, and one claim the vendor does not support

Read this branch's CI and the vendor together, because three things here are worth having before the
next push.

### 1. `Run tests` never ran, so this suite is ungraded

The required check failed at **step 8 `Run lint`** and **step 9 `Run tests` was skipped**. The 275-line
test file has therefore never been executed by CI - a green-looking "lint fix" round will be the first
time anything here is graded.

The named error:

```
tests/tools/test_feetech_motor_id_excludes_the_broadcast.py:275: error: Skipping analyzing
"scservo_sdk": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

`scservo_sdk` is not in the `[[tool.mypy.overrides]]` third-party-without-stubs list (0 of the 14
blocks name it). Adding it there is the repo's own convention for this shape, and
`tests/test_mypy_untyped_third_party_imports.py` reads that list, so it wants updating in the same
change. I did not work out why mypy attributes the error to a comment line rather than to an import,
so that part is unexplained rather than diagnosed.

### 2. The vendor guard count is 3, and all three are ping/read paths

`test_the_sdk_refuses_the_broadcast_for_reply_expecting_operations` asserts

```python
assert source.count("if scs_id >= BROADCAST_ID:") >= 3
```

Measured against the installed SDK, that count **is** 3, so the assertion passes. But the three
functions carrying the guard are `ping`, `readTx` and `readTxRx` - a ping and **two read paths**. The
comment above the assertion says "ping, a single read and a single write each bail out", and no write
does:

| `scservo_sdk` operation | `scs_id >= BROADCAST_ID` guard |
|---|---|
| `ping` | present |
| `readTx` | present |
| `readTxRx` | present |
| `writeTxRx` | **absent** |
| `write2ByteTxRx` | **absent** (delegates to `writeTxRx`) |

`txRxPacket` explains the asymmetry: it carries `if (txpacket[PKT_ID] == BROADCAST_ID):` and returns
without waiting for a reply. A broadcast **write** is a supported operation with a defined meaning;
a broadcast **read** has none, which is why only the read paths refuse it up front.

So the assertion passes for a different reason than the comment gives, and the docstring's three-row
table is vendor-backed for `feetech_ping` only - not for `feetech_position` or `feetech_velocity`.
The test name is the accurate part: *reply-expecting* operations are what the vendor refuses.

### 3. #2820 changes the same two files

#2820 (`fix(tools): refuse the Feetech broadcast where serial_tool is about to read one reply`) is
already approved and modifies `strands_robots/tools/serial_tool.py` and
`tests/tools/test_serial_tool_numeric_domain.py` - both of the files this branch modifies. It narrows
`feetech_ping` alone and keeps the full `[1, 254]` domain on the two write actions, which is where
the measurement above puts the line.

Worth deciding which scope is wanted before both land, since the two disagree about whether a
broadcast register write should be refused. Note also that the two branches create *differently
named* new test files and only one carries a changelog fragment, so
`scripts/check_duplicate_claim.py --all-open` reports `unique-additions` over this pair (15 pairs
compared, no overlap found) - the relation is keyed on created paths, and these two overlap only on
modified ones.

cagataycali:
Closing as a duplicate of #2820, which was opened 13 minutes earlier and is the better change.

Two things theirs does that this one does not:

- It refuses the broadcast only for **reply-expecting** actions (`_REPLY_EXPECTING_ACTIONS = frozenset({"feetech_ping"})`), keeping a reply-less broadcast write available. That is correct and this PR over-reached: `feetech_position` and `feetech_velocity` write and never read a status packet, so a broadcast there means what it says, and it is what the vendor's own `SYNC_WRITE` uses `BROADCAST_ID` for. Only the ping needs one servo to answer.
- It corrects `docs/hardware/tools.md:97`, which documents the range as `[1, 254]`. This PR missed that surface.

Measurements from this branch that are worth keeping have been posted to #2820: the SDK's four broadcast sites and which of them wait for a reply, the 1137-module / 1697 ms / numpy cost of importing the constants at module scope, and one mutation their suite does not currently refuse.

No code from here is needed. #2820 covers the defect.

PR #2824, opened 2026-08-27
fix(rendering): check a background's alignment numbers instead of coercing them

cagataycali:
## Problem

`strands_robots.rendering` closed its **posture flags** on `boolean_flag_error` — a domain whose own docstring says it is "for a flag that selects a posture *rather than scaling a quantity*", and whose suite states the boundary outright: "these four were the remaining **posture flags** in `strands_robots.rendering` that did not."

The quantities were the other half. Every scalar number `PanoramaBackground` and `GsplatBackground` take was handed to a bare `float()`, which accepts `nan` and `inf`.

`PanoramaBackground.rotation_deg` builds the `Rz` every output ray is turned by, so a non-finite yaw turned each world direction into `nan`, the equirectangular lookup sampled nothing, and `render()` returned a **uniformly black backdrop** — while reporting no error at all:

| `rotation_deg` | before | after |
|---|---|---|
| `0.0` (default) | rendered, mean 80.59, 101 distinct | unchanged, md5 `440ab1cb8d30` |
| `90.0` | rendered, mean 80.65, 102 distinct | unchanged, md5 `dbe0aaa1bc71` |
| `nan` | **rendered, mean 0.00, 1 distinct** | `ValueError: PanoramaBackground: rotation_deg must be a finite number, got nan.` |
| `inf` / `-inf` | **rendered, mean 0.00, 1 distinct** | refused, naming the parameter |
| `True` | silently a 0.0175 rad yaw | refused |
| `"90"` / `None` / `10**400` | bare `TypeError` from the coercion | refused, naming the parameter |

The silence is what made it costly. `GsplatBackground`'s own path comment describes the mechanism from the other side: in app contexts the photoreal background "sits inside a catch-all that demotes the photoreal background to a procedural fallback". A black frame raises nothing, so that fallback could not fire.

`GsplatBackground`'s numbers feed the fitted `world_from_gs`. Measured through the pure-NumPy skybox fit, a non-finite value in any of them produces a 4x4 with non-finite cells, so every gaussian is placed nowhere:

| parameter | `nan` through the fit | `True` |
|---|---|---|
| `up_sign` | 12 non-finite cells | silently `1.0` |
| `yaw_deg` | 12 non-finite cells | silently a 1-degree yaw |
| `radius` | 12 non-finite cells | silently 1 m |
| `floor_z` | 1 non-finite cell | silently 1 m |
| `backdrop_radius` | 12 non-finite cells | silently 1 m |
| `min_opacity` | filter skipped — `nan > 0` is `False` | silently 1.0 |
| `floor_pct` | already refused by `numpy.percentile` | silently 1.0 |

`min_opacity` is the sharpest: the guard reads `if min_opacity > 0`, and `nan > 0` is `False`, so a non-finite threshold skips the opacity filter the value asks to apply.

## Change

Every caller-supplied scalar number in `backgrounds.py` now goes through `finite_number_error`, the shared signed domain for "a physical quantity a caller supplies verbatim". It is the same rule `_shadow_plane_z_error` states one module over for a plane height — *"only a finite number can be honored: a non-finite plane intersects no ray and the shadow pass would silently never fire."*

`up_sign` and `clip_below` carry a documented `None` sentinel (auto-detect / drop nothing), so the domain applies to the supplied number and the sentinel passes through untouched.

**Bounds are deliberately not imposed.** Whether `min_opacity` belongs in `[0, 1]` and `floor_pct` in `[0, 100]` is a policy question this change does not answer, and `numpy.percentile` already refuses a non-finite `floor_pct` loudly on its own. Vector options (`center`, `backdrop_center`, `up_axis`, `major_axis`) are tuples and want an element-wise domain rather than this scalar one.

## Why nothing caught it

`test_panorama_yaw_rotation_changes_the_background` drives `rotation_deg` at `0.0` and `90.0` — the two values that work. The sibling class has a dedicated domain suite (`test_gsplat_background_posture_flag_domain.py`); the numbers had none. Every mutation below fires **0** of the 900 pre-existing cells in `tests/rendering/` except the two over-reach rows.

## Tests

New `tests/rendering/test_background_numeric_option_domain.py`, 125 cells, dependency-free (both constructors defer every `gsplat`/`torch` import to the first render; the transform consequence is measured through the pure-NumPy fit).

Pre-fix split — **83 failed / 42 passed**:

|---|---|---|
| `TestEveryScalarNumberIsCheckedOnTheSharedDomain` | 3 / 5 | derived inventory |
| `TestANonFiniteYawRenderedAnEntirelyBlackBackdrop` | 10 / 10 | regression |
| `TestANonFiniteAlignmentNumberPoisonedTheFittedTransform` | 70 / 70 | regression |
| `TestWhatTheRefusalMustNotCost` | **0 / 18** | over-reach controls |
| `TestThePremisesTheseTestsRestOn` | **0 / 22** | premises |

The inventory is derived from the annotation (`float`, `float | None`) rather than a hardcoded list, so a background that grows a tenth number is held to the same rule the hour it lands — `b10` below is that property.

### Mutation table

`mine` = failures in the new file, `sib` = failures in the other 900 cells of `tests/rendering/`. `collected` stayed 125 on every row, so no row hid a deselection.

| # | mutation | mine | sib |
|---|---|---|---|
| b0 | control — no mutation | 0 | 0 |
| b1 | panorama `rotation_deg` guard removed | 11 | 0 |
| b2 | gsplat plain-numeric loop removed | 55 | 0 |
| b3 | gsplat sentinel loop removed | 17 | 0 |
| b4 | all three removed (raw revert) | 82 | 0 |
| b5 | panorama guard moved **after** the coercion | 4 | 0 |
| b6 | domain swapped to `positive_finite_number_error` | 66 | **50** |
| b7 | sentinel loop drops the `is not None` exemption | 17 | **40** |
| b8 | domain re-implemented locally instead of imported | 75 | 0 |
| b9 | guard reports the context where the parameter belongs | 11 | 0 |
| b10 | `min_opacity` dropped from the guarded numerics | 10 | 0 |

10 distinct firing sets; `b1 | b2 | b3 == b4` exactly, so each guard is independently pinned. Source restored byte-identical after every row.

`b6` and `b7` are the two rows that matter for scope: each trips **50** and **40** pre-existing rendering tests, which is the measurement that the domain must be the *signed* one and that the `None` sentinel must pass through. `b5` shows the placement is load-bearing — `float(np.deg2rad(nan))` is a perfectly good float, so a guard after the coercion inspects a value the defect has already made indistinguishable from a measurement. `b8` fires because a local re-implementation holding today's identical rule is exactly what the unguarded state looked like, so the *import* is the property.

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1635 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a worktree of the merge base, normalized message sets identical in both directions, **0 attributable**.
- Paired run over an identical 518-file selection (all of `tests/rendering/` plus every guard that walks the tree, parses source, or reads docstrings / `__all__` / `changelog.d`), the new file excluded from both trees: identical **24315 collected** and `18 failed, 24157 passed, 140 skipped` on each, **18 identical failing ids with `comm` empty in both directions**. Of those 18, 17 need `awsiot`/`awscrt` (`find_spec` is `None` here; both are declared in the `[mesh-iot]` extra, which CI installs) and 1 is the lerobot-from-source `encode()` drift.
- After rebasing onto `4b91089a`, ruff re-run clean and `tests/rendering/` plus the three test files `main` gained: **902 passed, 3 skipped**. Break rows `b0`/`b1`/`b4` re-measured unchanged (0 / 11 / 82).

## Artifact

One capture script, both trees, one camera. Columns 1 and 2 are byte-identical before and after (md5 `440ab1cb8d30` and `dbe0aaa1bc71`), so nothing legitimate moved; column 3 is the defect and column 4 is the refusal that replaces it. Procedural panorama, pure NumPy — no `sim-gs` extra.

panorama yaw domain

### DIFF
--- a/changelog.d/2824-background-numbers-are-checked-not-coerced.md
+++ b/changelog.d/2824-background-numbers-are-checked-not-coerced.md
@@ -1,0 +1,29 @@
+### Fixed: a background's alignment numbers are checked instead of coerced
+
+`PanoramaBackground` and `GsplatBackground` handed every scalar number a caller
+supplies to a bare `float()`, which accepts `nan` and `inf`. A non-finite
+`rotation_deg` built the rotation matrix every output ray is turned by, so each
+world direction became `nan`, the equirectangular lookup sampled nothing, and
+`render()` returned a uniformly black backdrop -- `rgb.mean() == 0.0` with a
+single distinct value -- while reporting no error at all. That last part is what
+made it costly: in app contexts the photoreal background sits inside a catch-all
+that demotes it to a procedural fallback, and a silent black frame never raises,
+so the fallback could not fire.
+
+The gsplat numbers feed the fitted `world_from_gs`. A non-finite `up_sign`,
+`yaw_deg`, `radius`, `floor_z` or `backdrop_radius` produced a 4x4 with
+non-finite cells, so every gaussian was placed nowhere; `min_opacity` was worse
+than nowhere, because `nan > 0` is `False` and that skips the opacity filter the
+value asks to apply.
+
+The posture flags beside these were already closed on `boolean_flag_error`, a
+domain documented as being for a flag that selects a posture rather than scaling
+a quantity. The quantities were the other half, and they now take
+`finite_number_error`, the shared signed domain for a physical quantity a caller
+supplies verbatim -- the same rule `_shadow_plane_z_error` states one module
+over for a plane height. `up_sign` and `clip_below` carry a documented `None`
+sentinel, so the domain applies to the number and the sentinel passes through.
+
+Bounds are deliberately not imposed: whether `min_opacity` belongs in `[0, 1]`
+and `floor_pct` in `[0, 100]` is a separate question, and `numpy.percentile`
+already refuses a non-finite `floor_pct` on its own.
--- a/strands_robots/rendering/backgrounds.py
+++ b/strands_robots/rendering/backgrounds.py
@@ -37,7 +37,7 @@ from typing import Any, Protocol
 
 import numpy as np
 
-from strands_robots.utils import boolean_flag_error, require_optional
+from strands_robots.utils import boolean_flag_error, finite_number_error, require_optional
 
 from .camera import CameraParams
 
@@ -95,6 +95,13 @@ class PanoramaBackground:
         rotation_deg: yaw rotation of the panorama in degrees (rotates around
             the world +Z axis). Useful for aligning the panorama with the
             scene without re-rendering it.
+
+    Raises:
+        ValueError: if ``rotation_deg`` is not a finite number. The yaw builds
+            the rotation matrix every output ray is turned by, so a non-finite
+            angle turns every ray into ``nan`` and the panorama samples nothing
+            -- the backdrop renders black rather than rotated, and nothing
+            raises to trigger a caller's fallback.
     """
 
     name = "panorama"
@@ -105,6 +112,17 @@ class PanoramaBackground:
         rotation_deg: float = 0.0,
     ) -> None:
         self._image_path = Path(image_path) if image_path else None
+        # Checked before the coercion, on the shared signed-finite domain, for
+        # the reason :func:`~strands_robots.rendering.compositor._shadow_plane_z_error`
+        # gives for a plane height: only a finite number can be honored. The
+        # yaw builds the ``Rz`` every world ray is turned by, so ``nan``/``inf``
+        # turned every direction into ``nan``, the equirect lookup sampled
+        # nothing, and the backdrop came back uniformly black -- with no
+        # exception, so a caller's photoreal-to-procedural fallback never fired.
+        # A signed domain because both senses of yaw are legitimate and a
+        # rotation wraps, so there is no floor or ceiling to impose.
+        if text := finite_number_error(rotation_deg, "rotation_deg", "PanoramaBackground"):
+            raise ValueError(text)
         self._rotation_rad = float(np.deg2rad(rotation_deg))
         self._panorama: np.ndarray | None = None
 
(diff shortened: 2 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Closes the numeric half of the posture-flag rule for the two rendering backgrounds: `PanoramaBackground.rotation_deg` and the eight scalar alignment numbers `GsplatBackground` takes (`backdrop_radius`, `yaw_deg`, `radius`, `floor_z`, `min_opacity`, `floor_pct`, plus the sentinel-bearing `up_sign` / `clip_below`) are now checked on the shared `finite_number_error` domain before the bare `float()` coercion that used to accept `nan`/`inf`. The failure being closed is the silent one the module's own path-validation comment warns about: a non-finite yaw or fit input produced a uniformly black backdrop or a non-finite `world_from_gs` with nothing raised, so the app-side photoreal-to-procedural fallback never fired. The guards sit ahead of the coercions, the `None` sentinels pass through untouched, and no signature changes.

## What's good

- Reuses the shared domain from `strands_robots/utils.py` (5+ existing call sites) rather than adding a local guard, per the AGENTS.md value-domain rule; `ruff check`/`format` clean on both files.
- The panorama guard is placed before `np.deg2rad`, where a post-coercion check would inspect an already-laundered value, and the test pins that ordering.
- Test inventory is derived from the constructors' annotations and the guard calls' own arguments (AST), so a background that grows a new scalar is graded without an edit; over-reach controls pin that every documented spelling (`None` sentinels, NumPy scalars, negative yaws, the declared defaults) still constructs.
- Verified locally at `0755322f`: the new suite passes (125/125) and the rest of `tests/rendering/` passes (the one `gsplat_rasterizer_available` subprocess-probe timeout on my box is environmental and untouched by this diff); `tests/test_raises_docstring_completeness.py` passes with the new `Raises:` block.

Note: no `changelog.d/2824-*.md` fragment is in the diff yet - the `changelog-fragment` workflow will flag it, so just adding the file clears that check.

yinsong1986 (approved):
## Summary

Closes the numeric half of the rendering backgrounds' input-domain rule: `PanoramaBackground.rotation_deg` and `GsplatBackground`'s eight scalar alignment numbers (`backdrop_radius`, `yaw_deg`, `radius`, `floor_z`, `min_opacity`, `floor_pct`, plus the sentinel-bearing `up_sign` / `clip_below`) are now checked on the shared `finite_number_error` domain ahead of the bare `float()` coercion that used to accept `nan`/`inf`. The defect being closed is the silent one: a non-finite yaw rendered a uniformly black backdrop with nothing raised, so the app-side photoreal-to-procedural fallback could never fire, and a non-finite fit input poisoned the whole `world_from_gs`. The guards sit before the coercions, the documented `None` sentinels pass through untouched, no signatures change, and bounds are deliberately left out of scope. Since the prior (dismissed) review at `0755322f`, the changelog fragment it noted as missing landed (`5ee90871`) and `8929b841` renamed the numeric loops' variables off the boolean loop's narrowed names.

## What's good

- Reuses the shared domain from `strands_robots/utils.py` (30+ existing call sites) rather than a local guard, per the AGENTS.md value-domain rule; the mutation table's `b8` row pins that the import itself is the property.
- The panorama guard precedes `np.deg2rad`, where a post-coercion check would inspect an already-laundered value, and a test pins that ordering.
- The test inventory is derived from the constructors' annotations and the guard calls' own arguments via AST, so a background that grows a new scalar is graded without an edit; over-reach controls pin that the `None` sentinels, NumPy scalars, negative yaws and the declared defaults all still construct.
- Verified at `8929b841`: the new suite passes (125/125), all of `tests/rendering/` is green, and the tree-walking guards (`test_raises_docstring_completeness`, `test_changelog_fragments`, `test_docstring_xref_roles_resolve`, `test_log_strings_are_ascii`, `test_no_host_paths`, `test_test_case_names_describe_behaviour`) pass; `ruff check` and `ruff format --check` are clean on both changed files.

PR #2825, opened 2026-08-27
fix(tools/pose_tool): report the joints an interpolated move actually commanded

cagataycali:
## The defect

`MotorController._smooth_move` ended in a literal `return True` and threw away
every `move_motor` result, so the interpolating pose path reported success
whatever reached the bus. Two joints could go missing silently:

- **A motor whose current position did not arrive** has no start point, so the
  loop builds no trajectory for it and never commands it. It was dropped from
  the move without a word.
- **A write the bus refused** (a pulled cable, a bus held by another process) was
  logged by `move_motor` and then discarded by the loop.

## Every sibling already answered for itself

| method | outcome of `move_motor` |
|---|---|
| `move_multiple_motors(smooth=False)` | `if not self.move_motor(...): success = False` |
| `disable_torque` | returns the motors it could not reach |
| `incremental_move` | `return self.move_motor(...)`, and refuses outright when the position is unreadable |
| `_smooth_move` | discarded, then `return True` |

So one method carried two contracts a single `smooth` flag apart - and `smooth`
defaults to `True`, which made the default path the one that could report a pose
it had not commanded a single packet towards. `reset_to_home` passes
`smooth=True` itself and had no honest branch to take at all, so the action an
operator reaches for to put the arm somewhere known was the one that could not
decline to interpolate.

`disable_torque`'s docstring already argues the disposition: a caller told the
whole arm was released when part of it is still driven is worse than no stop at
all. `read_motor_position`'s docstring already claimed the interpolating paths
"refuse to build a trajectory from it" - true of the trajectory, and the joint
was still reported as moved.

## Measured

A bus that accepts writes and answers no position read is an ordinary
half-duplex symptom. Driving `pose_tool` against one, counting the
`Goal_Position` (`0x2A`) packets that reach it:

| call | before | after |
|---|---|---|
| `move_multiple`, `smooth=True` (default) | `status="success"`, both target angles listed, **0 packets** | `status="error"`, **0 packets** |
| `move_multiple`, `smooth=False`, same bus | `status="success"`, 2 packets | unchanged |
| `reset_to_home` (passes `smooth=True`) | `"Robot moved to home position"`, **0 packets** | `status="error"`, **0 packets** |

At the controller, with two joints requested and one silent:

| bus | before | after | packets |
|---|---|---|---|
| no motor answers a read | `True` | `False` | 0, unchanged |
| one motor silent | `True` | `False` | only the readable joint, unchanged |
| the position write is refused | `True` | `False` | only the readable joint, unchanged |
| healthy | `True` | `True` | `steps + 1` per joint, unchanged |

The packets are identical in every row. Only the reported outcome changed, which
is what the tool's own handlers already turn into an error envelope for the
one-shot branch.

## Scope

Every motor is still attempted, matching `disable_torque` and the one-shot
branch - reporting the failure must not abandon the joints that can still move,
and that is pinned. A step count that runs no increments is deliberately still a
success: `_smooth_move_option_error` refuses `steps <= 0` before any caller of
`pose_tool` reaches the loop, and asking for no increments is not the same
failure as a joint that could not move. Conflating them fires a pre-existing test
(`test_a_negative_step_count_writes_nothing_yet_reports_success`).

Which packets are written is unchanged. Whether a partial pose should be written
at all is a separate question and is not answered here.

## Why nothing caught it

`reset_to_home`'s existing ASCII test drove `fake_serial`, whose `read` answers
`b""`, so it asserted `status="success"` on a call that wrote nothing to the bus.
It now takes the answering double - promoted into the shared `tests/tools`
conftest, whose own docstring exists for exactly this ("several modules now need
them"), removing the duplicate copy in the interpolation module - and asserts the
bus was really driven.

The new tests assert against the **bytes** that reach the bus rather than the
status text, for the reason the sibling emergency-stop module gives: the wording
was already right, and only the packets say whether the arm moved.

## Gate

Pre-fix split, source reverted and tests kept: **8 failed / 181 passed** -
`TestTheInterpolatedMoveAnswersForWhatItCommanded` 5/5,
`TestTheToolReportsIt` 2/2, the derived scan 1/2 (its non-vacuity cell passes),
and **0 of 7** in `TestWhatIsUnchanged`. The headline failure reads
`AssertionError: Robot moved to home position`.

Mutation table, 12 rows, control clean, `collected` stable at 15 on every row:

|---|---|---|---|
| b0 | 0 | 0 | control |
| b1 | 8 | 0 | the whole method reverted |
| b2 | 4 | 0 | the uncommanded-joint half of the verdict dropped |
| b3 | 2 | 0 | the failed-write half of the verdict dropped |
| b4 | 3 | 0 | `move_motor`'s outcome discarded again |
| b5 | 6 | 9 | the uncommanded set computed the wrong way round |
| b6 | 1 | 0 | the error names no joint |
| b7 | 1 | 0 | the loop abandons the remaining joints at the first failure |
| b8 | 3 | 0 | an unreadable joint refuses the whole move instead of reporting it |
| b9 | 1 | 1 | an empty loop conflated with a joint that could not move |
| b10 | 3 | 0 | b4 composed, scan derived |
| b11 | 2 | 0 | b4 composed, scan hardcoded to today's caller names |

`b2` and `b3` are **disjoint**, their union is a subset of `b1`, and the two
`b1` cells outside it are the derived scan and the joint naming - so each half of
the verdict is independently pinned. `b7` and `b8` fire `TestWhatIsUnchanged`,
so the "attempt everything" contract is graded rather than asserted. `b9` fires a
pre-existing test, so the empty-loop boundary is not mine to move. `b10` vs `b11`
is why the scan derives its callers instead of listing them: with a discard
reintroduced, deriving names it and a hardcoded list goes silent.

Paired against `4b91089a` over an identical 455-file selection (all of
`tests/tools`, every suite walking the tree or parsing source, everything naming
`pose_tool`), the new file excluded from both trees: identical **21714
collected** and `20 failed, 21616 passed, 78 skipped` on each, **20 identical
failing ids with `comm` empty in both directions**. All 20 are pre-existing: 17
need `awsiot`/`awscrt` (both absent locally, both declared in `[mesh-iot]`) and 3
are lerobot-from-source `encode()` drift. `ruff check` and `ruff format --check`
clean over 1635 files; `mypy` **24 == 24** with the normalized message sets
differing by one `[annotation-unchecked]` note and **0 errors** attributable.

No visual artifact: this changes a hardware tool's reported outcome and its
logging, not a policy, simulation, render, recording or asset. The measured
packet counts above are the evidence.

### DIFF
--- a/changelog.d/2825-interpolated-move-reports-what-it-commanded.md
+++ b/changelog.d/2825-interpolated-move-reports-what-it-commanded.md
@@ -1,0 +1,42 @@
+### Fixed: an interpolated pose move reports the joints it actually commanded
+
+`MotorController._smooth_move` ended in a literal `return True` and threw away
+every `move_motor` result, so the interpolating path answered success whatever
+reached the bus. Two joints could go missing silently. A motor whose current
+position did not arrive has no start point, so no trajectory is built for it and
+the write loop never commands it - it was dropped from the move without a word.
+And a write the bus refused was logged by `move_motor` and then discarded by the
+loop.
+
+Every other commanding method on the class already answered for what it did.
+`move_multiple_motors(smooth=False)` reads each `move_motor` outcome,
+`disable_torque` returns the motors it could not reach - its docstring explains
+why, that a caller told the whole arm was released when part of it is still
+driven is worse than no stop at all - and `incremental_move` refuses outright
+when the current position is unreadable. So one method carried two contracts a
+single `smooth` flag apart, and `smooth` defaults to `True`: the default path was
+the one that could report a pose it had not commanded a single packet towards.
+`reset_to_home` passes `smooth=True` itself and had no honest branch to take,
+which made the action an operator reaches for to put the arm somewhere known the
+one that could not decline to interpolate.
+
+On a bus that accepts writes and answers no position read - an ordinary
+half-duplex symptom - `pose_tool(action="move_multiple")` listed both requested
+angles under `status="success"` with zero `Goal_Position` packets written, while
+the same call with `smooth=False` on the same bus wrote both and reported
+honestly. The interpolating branch now returns `False` when a joint was left
+uncommanded or a write failed, and logs which joints at ERROR, so the tool's own
+handlers turn it into an error envelope the way they already do for the one-shot
+branch.
+
+Every motor is still attempted, matching `disable_torque` and the one-shot
+branch: the reported outcome changed, the packets did not. A step count that runs
+no increments is deliberately still a success - `_smooth_move_option_error`
+refuses `steps <= 0` before any caller of `pose_tool` reaches the loop, and
+asking for no increments is not the same failure as a joint that could not move.
+
+The suite could not see any of this. `reset_to_home`'s existing ASCII test drove a
+serial double that answers no read, so it asserted `status="success"` on a call
+that wrote nothing; it now uses the answering double, promoted into the shared
+`tests/tools` conftest that a third module needed, and asserts the bus was really
+driven.
--- a/strands_robots/tools/pose_tool.py
+++ b/strands_robots/tools/pose_tool.py
@@ -602,8 +602,9 @@ class MotorController:
             verified reply from this motor arrived. ``None`` is the established
             "could not read" signal every caller already handles:
             :func:`pose_tool`'s ``read_position`` turns it into an error envelope
-            instead of quoting a number, and the interpolating paths refuse to
-            build a trajectory from it.
+            instead of quoting a number, and the interpolating paths build no
+            trajectory from it and report the joint as uncommanded rather than
+            leaving it out of the move in silence.
         """
         if not self.serial_conn or not self.serial_conn.is_open:
             return None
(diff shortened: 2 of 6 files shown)


yinsong1986 (approved):
## Summary

This PR fixes a silent-success bug in `MotorController._smooth_move`: the interpolating branch ended in a literal `return True`, so a joint whose current position never arrived was dropped from the move without a word, and a `move_motor` write the bus refused was discarded. The fix returns `not unreadable and not failed`, names the affected joints at ERROR, and leaves the packet behaviour unchanged (every reachable joint is still driven, matching `disable_torque`'s continue-past-failure contract). The `pose_tool` handlers for `move_multiple`, `load_pose` and `reset_to_home` already turn a `False` into an error envelope, so the honest verdict reaches the caller with no tool-surface change. This closes exactly the class of failure AGENTS.md's no-silent-defaults rule names - a success reported for a pose the arm was never commanded towards.

## What's good

- Regression tests assert on the bytes that reach the bus (`Goal_Position` packet counts per motor id), not the status text, and separate the two failure halves (unreadable read vs refused write) with independent knobs.
- The `TestWhatIsUnchanged` class pins that the trajectory, the healthy path, the one-shot branch, and the empty-loop success are all untouched - the reported outcome is the only behaviour change.
- The AST scan (`TestEveryCommandingMethodAnswersForItself`) grades the rule over the class rather than a method list, so a fourth commanding path is held to it on arrival.
- Conventions followed: changelog fragment (not a `CHANGELOG.md` append), shared test doubles promoted to the `tests/tools` conftest with the F811 rationale updated, ASCII-only log strings, no host paths, behaviour-named tests. Ran the three touched test modules at the head commit: 180 passed; `ruff check` and `ruff format --check` clean.

PR #2826, opened 2026-08-27
fix(teleop): join the teleop loop before detach_teleop disconnects what it reads

cagataycali:
## The defect

`_teleop_loop` indexes `self._teleops[name]` on every tick. `detach_teleop`
removes entries from that mapping - and disconnects each device - and only then
stopped the loop, discarding the outcome. So a detach that would leave the loop
with nothing to drive tore the leader down under a loop still parked reading from
it, and answered `status="success"`.

`stop_teleoperate` declines exactly that teardown for itself: on a leader whose
`get_action()` is still blocking it answers `status="error"` with `stopped: False`
and deliberately leaves the devices connected, because "Tearing the bus down under
a live writer is what `G1Driver.cleanup` refuses for the same reason". The same
state through each verb, on one wedged leader:

| verb | status | `stopped` | leader disconnected | loop alive |
|---|---|---|---|---|
| `stop_teleoperate()` | `error` | `False` | no | yes |
| `detach_teleop()` | **`success`** | - | **YES** | yes |

`detach_teleop`'s own docstring already said "Stops the local loop first".

Three things were wrong, and joining first repairs all three:

- **the verdict.** The detach answered `success` where its own callee answered
  `error`, and carried no `json` block - so `_stop_reported_stopped` read `True`
  on the detach envelope and a caller had no way to tell.
- **the teardown.** The leader was disconnected, undoing the protection
  `stop_teleoperate` had just decided to apply.
- **the diagnosis.** Because the entries were popped first, the refusal that got
  discarded described "the loop is still polling `[]`" while claiming "The devices
  are left connected", which this caller had already falsified.

## Two routes in, and the second is the documented remedy

`stop_teleoperate` clears `_teleop_running` *before* it joins and keeps the thread
handle when the join fails. So after a failed stop - the state whose own message
says to call it again - the session flag alone reports an idle session while the
loop is still writing:

| route | `_teleop_running` | `_teleop_thread is not None` |
|---|---|---|
| loop running | `True` | `True` |
| after a failed stop | **`False`** | `True` |

The guard therefore tests the thread handle as well as the flag. Keyed on the flag
alone it sees the first route only - which is what break row b2 measures.

## Before / after

```
                                          before      after
detach under a running loop: status       success     error
  leader disconnected                     yes         no
  streams left attached                   0 of 1      1 of 1
  detached: [] reported                   (no json)   yes
detach after a failed stop: status        success     error
  leader disconnected                     yes         no
forwarded diagnosis names the device      devices: [] devices: ['leader']
```

A live loop now refuses the whole detach: `status="error"` with `detached: []`,
every stream left attached and connected so a later call re-joins that same loop,
and the callee's reason forwarded rather than re-invented. The outcome is read
through the shared `_stop_reported_stopped` helper from a single call site that
precedes the pop - on `stopped` rather than `status`, because the status is derived
from the session counters and a session whose every frame errored answers
`"error"` after a perfectly clean join, which would refuse a detach that is safe.

## Why nothing caught it

The whole defect fires **0** of the 7 pre-existing teleop suites (`sib` column
below). The existing coverage grades the two extremes - a `FakeTeleop` that returns
instantly, so its join always succeeds and the discarded value is always `True` -
and `detach_teleop("")`, which selects nothing and so cannot reach the branch.

## Pre-fix split

`13 failed / 11 passed`, by class:

|---|---|---|
| `TestThePremiseStopTeleoperateDeclinesThisTeardown` | premise | 0 / 4 |
| `TestALiveWriterMakesTheDetachRefuse` | regression (route 1) | 7 / 7 |
| `TestTheFailedStopRemedyPathRefusesToo` | regression (route 2) | 2 / 2 |
| `TestTheOutcomeIsReadFromOneSingleSourcedCallSite` | structural | 4 / 4 |
| `TestWhatIsUnchanged` | over-reach controls | 0 / 7 |

One cell initially passed pre-fix and was tightened: `"['leader']"` is also in the
`Detached: ['leader']` success message this used to return, so it now matches the
phrase only the forwarded diagnosis produces.

## Mutation table

10 mutations, 10 firing, **10 distinct firing sets**, control clean, `collected`
stable at 24, source restored byte-identical. `sib` = failures in the 7 pre-existing
teleop suites with this file deselected.

|---|---|---|---|
| b0 | 0 | 0 | control |
| b1 | 13 | 0 | raw revert: no pre-flight join at all |
| b2 | 3 | 0 | guard keyed on `_teleop_running` only (drops the thread handle) |
| b3 | 2 | 0 | verdict read from `status` instead of the shared helper |
| b4 | 5 | 0 | guard moved AFTER the pop loop (original ordering, but checked) |
| b5 | 2 | 0 | refusal reports `success` |
| b6 | 3 | 0 | refusal carries no `json` block |
| b7 | 2 | 0 | refusal drops the forwarded reason |
| b8 | 3 | **1** | OVER-REACH: refuse any detach while the loop runs |
| b9 | 13 | **2** | the empties-the-registry test is inverted |
| b10 | 9 | 0 | ALTERNATIVE: log the live loop but detach anyway |

The scope rows are the load-bearing ones. **b8** trips
`test_an_empty_name_does_not_end_a_running_session` - a too-wide guard breaks the
membership contract a previous change pinned. **b9** trips that plus
`test_detach_all_stops_running_loop`. **b10** shows that reporting alone is not the
fix: logging and detaching anyway still fires 9, so declining the detach is the
behavioural content.

## Gate

- `ruff check` + `ruff format --check` over `strands_robots tests tests_integ`: clean, 1635 files
  merge-base, normalized message sets identical both directions, **0 attributable**
- paired run, identical 460-file selection (every teleop suite plus every guard that
  walks the tree, parses source, or reads docstrings / `docs/` / `changelog.d`), new
  file excluded from both and every path verified present on both:

```
                     branch                              base
collected            21559                               21559
result               24 failed, 21459 passed, 76 skipped 24 failed, 21459 passed, 76 skipped
failing ids          24                                  24
comm both directions EMPTY
```

The 24 are classified rather than assumed: **17** need `awsiot`/`awscrt`
(`find_spec` False for both, both declared in the `[mesh-iot]` extra, which CI
installs), **3** are the lerobot-from-source `encode()` drift, and **4** in
`test_recording_frame_loss_is_not_tolerated.py` pass **13/13 in isolation on both
trees** - a large-selection module-caching artifact.

## Scope

A partial detach - one of several devices, leaving the loop with something to drive
- is deliberately out of scope and pinned unchanged. The loop's `except Exception`
counts a popped-device `KeyError` as an error frame, so whether the loop should be
re-selected or the detach refused there is a contract question about multi-device
sessions rather than this ordering defect.

`describe()`'s method line and `docs/hardware/teleoperation.md` both documented the
old order ("disconnects each and stops the loop"); both are corrected here.

No visual artifact: this is a teardown-ordering and reporting fix with no policy,
simulation, rendering, recording or asset behaviour. The measured tables above are
the evidence.

### DIFF
--- a/changelog.d/2826-detach-teleop-joins-before-it-disconnects.md
+++ b/changelog.d/2826-detach-teleop-joins-before-it-disconnects.md
@@ -1,0 +1,62 @@
+### Fixed: `detach_teleop` joins the teleop loop before it disconnects what that loop reads
+
+`_teleop_loop` indexes `self._teleops[name]` on every tick, and `detach_teleop`
+removes entries from that mapping. It removed them - and disconnected each
+device - and only then stopped the loop, discarding the outcome. So a detach that
+would leave the loop with nothing to drive tore the leader down under a loop that
+was still parked reading from it, and answered `status="success"`. The method's
+own docstring already said "Stops the local loop first".
+
+`stop_teleoperate` declines exactly that teardown for itself: on a leader whose
+`get_action()` is still blocking it answers `status="error"` with `stopped: False`
+and deliberately leaves the devices connected, because "Tearing the bus down
+under a live writer is what `G1Driver.cleanup` refuses for the same reason". The
+same state through each verb, on one wedged leader:
+
+```
+verb                  status   stopped   leader disconnected   loop alive
+stop_teleoperate()    error    False     no                    yes
+detach_teleop()       success  -         YES                   yes
+```
+
+Three things were wrong and the order repairs all three. The verdict: the detach
+answered `success` where its own callee answered `error`, and carried no `json`
+block, so `_stop_reported_stopped` read `True` on it and a caller had no way to
+tell. The teardown: the leader was disconnected, undoing the protection
+`stop_teleoperate` had just decided to apply. And the diagnosis: because the
+entries were popped first, the refusal that got discarded described "the loop is
+still polling `[]`" while claiming "The devices are left connected", which this
+caller had already falsified.
+
+There are two routes in and the second is the one an operator is told to take.
+`stop_teleoperate` clears `_teleop_running` *before* it joins and keeps the thread
+handle when the join fails, so after a failed stop - the state whose message says
+to call it again - the session flag alone reports an idle session while the loop
+is still writing. The guard therefore tests the thread handle as well as the flag;
+keyed on the flag alone it sees the first route only.
+
+```
+                                          before      after
+detach under a running loop: status       success     error
+  leader disconnected                     yes         no
+  streams left attached                   0 of 1      1 of 1
+  detached: [] reported                   (no json)   yes
+detach after a failed stop: status        success     error
+  leader disconnected                     yes         no
+forwarded diagnosis names the device      devices: [] devices: ['leader']
+```
+
+A live loop now refuses the whole detach: `status="error"` with `detached: []`,
+every stream left attached and connected so a later call re-joins that same loop,
+and the callee's reason forwarded rather than re-invented. The outcome is read
+through the shared `_stop_reported_stopped` helper from a single call site that
+precedes the pop - on `stopped` rather than `status`, because the status is
+derived from the session counters and a session whose every frame errored reports
+`"error"` after a perfectly clean join, which would refuse a detach that is safe.
+
+Unchanged: a clean stop still detaches and disconnects exactly as before, a name
+matching no attached stream still gets the not-found refusal, and `detach_teleop("")`
+still leaves a running session alone. A partial detach - one of several devices,
+leaving the loop with something to drive - is deliberately out of scope and pinned
+unchanged; whether the loop should be re-selected or the detach refused there is a
+contract question about multi-device sessions rather than this ordering defect.
--- a/docs/hardware/teleoperation.md
+++ b/docs/hardware/teleoperation.md
@@ -86,7 +86,7 @@ Every hardware `Robot` and `Simulation` host exposes:
 |--------|------|
 | `attach_teleop(device_or_spec, *, name=None, method=None, map_fn=None, **kwargs)` | Register an input stream (lazy - no hardware touched). `device_or_spec` is a built teleop instance **or** a type string built via `Teleoperator(**kwargs)`. |
 | `teleoperate(*, names=None, robot_name=None, hz=50.0, publish=False, block=False, duration=None)` | Run the control loop. |
-| `detach_teleop(name=None)` | Remove one (or all) attached streams. |
+| `detach_teleop(name=None)` | Remove one (or all) attached streams. Stops the loop before touching a device when the detach would leave it with nothing to drive, and refuses with `detached: []` if that loop does not stop. |
 | `stop_teleoperate()` | Stop the loop, any mesh publishers, and disconnect devices. Reports `status="error"` with `stopped: false` when the loop outlasts its 3 s join budget - the devices are left connected rather than torn down mid-write, and a second call re-joins the same loop. |
 
 ### `attach_teleop`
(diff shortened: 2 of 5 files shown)


yinsong1986 (dismissed):
## Summary

Fixes an ordering defect in `TeleopMixin.detach_teleop`: it previously popped entries out of `self._teleops` (which `_teleop_loop` indexes on every tick) and disconnected each device *before* stopping the loop, then discarded `stop_teleoperate()`'s outcome - so a detach that left the loop with nothing to drive tore the leader down under a live writer and still answered `status="success"`, contradicting the protection `stop_teleoperate` applies to itself. The fix hoists the stop ahead of the pop, gates it on the thread handle as well as the session flag (covering the failed-stop remedy path where `_teleop_running` is already `False`), reads the outcome through the shared `_stop_reported_stopped` helper (on `stopped` rather than the counter-derived `status`), and refuses the whole detach atomically - `status="error"`, `detached: []`, every stream left attached and connected, callee's reason forwarded - when the join fails. A clean stop, the not-found refusal, the empty-name membership contract, and partial detach are all pinned unchanged.

## What's good

- The new test module covers both routes in (live loop and post-failed-stop), the atomicity of the refusal, the machine-readable payload, and structurally pins the single non-discarded `stop_teleoperate()` call site preceding the pop.
- Refusal chosen over `stopped` rather than `status` is correct and tested: a clean join whose every frame errored still detaches.
- Partial detach is explicitly scoped out and pinned unchanged rather than silently altered.
- Changelog fragment per AGENTS.md step 3; docs and `describe()` text updated in the same PR; all added user-facing strings are ASCII.
- Verified locally at head `646a5623`: the new suite (24 tests) plus the existing teleop suites (224 tests) pass, and `ruff check` / `ruff format --check` are clean on the touched files.

yinsong1986 (approved):
## Summary

Fixes the teardown-ordering defect in `TeleopMixin.detach_teleop`: it previously popped entries out of `self._teleops` (which `_teleop_loop` indexes on every tick) and disconnected each device *before* calling `stop_teleoperate()`, whose outcome it then discarded - so a detach that left the loop with nothing to drive tore the leader down under a live writer and answered `status="success"`, contradicting both its own docstring ("Stops the local loop first") and the protection `stop_teleoperate` applies to itself. The fix hoists the join ahead of the pop, gates it on the thread handle as well as the session flag (covering the failed-stop remedy path where `_teleop_running` is already `False` but the handle is retained), reads the outcome through the shared `_stop_reported_stopped` helper on `stopped` rather than the counter-derived `status`, and refuses the whole detach atomically on a failed join: `status="error"` with `detached: []`, every stream left attached and connected, and the callee's reason forwarded verbatim. `describe()` and `docs/hardware/teleoperation.md` are corrected to the new order in the same PR.

## What's good

- Verified at head `24f662b6`: the new suite (24 tests) plus all 7 pre-existing teleop suites (246 tests) pass, and `ruff check` / `ruff format --check` are clean on the touched files.
- The guard's edge routes all behave: live loop (refuse), failed-stop remedy path (refuse), dead-thread-handle after a natural `duration` end (join is instant, detach proceeds), partial detach (untouched, pinned), unknown/empty name (not-found refusal, session survives - the membership contract from the prior selector fix is preserved).
- Keying the verdict on `stopped` rather than `status` is correct and tested: a clean join whose every frame errored still detaches.
- Structural tests pin the single non-discarded `stop_teleoperate()` call site preceding the pop, so the discarded-return drift cannot silently return.
- Changelog fragment per AGENTS.md step 3; error dict returned rather than raised per the tool-surface contract; all added user-facing strings are ASCII; new `Args:`/`Returns:` docstring entries are complete.

PR #2827, opened 2026-08-27
fix(isaac): refuse a non-finite body placement instead of measuring around it

cagataycali:
## What is wrong

`load_mjcf_scene_objects` composes a `SceneObject` from three placements: the top-level body's `pos`
becomes the object's `position`, its orientation becomes `quat`, and each nested body's `pos` becomes
the running offset every geom below it is measured from.

`_geom_aabb` refuses a non-finite value on all four of a *geom*'s parse paths (#2813). The *body*
placements the same bound is composed from were read straight through `_parse_xyz` /
`_parse_orientation`, whose contract is to fall back to a documented default on an unreadable
attribute and to return whatever a readable one parsed to.

The nested case is the one nothing can detect. A non-finite offset makes every bound below it
non-finite, and the running `min`/`max` in `_recursive_collision_aabb` orders a NaN as neither smaller
nor larger than anything, so the whole subtree disappears from the union while the walk still reports
that it found analytic geometry.

Measured through the public loader on a table fixture whose leg hangs off a nested body. All ten rows
compile in MuJoCo:

| input | outcome | non-finite fields reported | reported `size` |
|---|---|---|---|
| finite placement (control) | success | none | `(0.8, 0.8, 0.76)` |
| nested body `pos="0 0 nan"` | success | **none** | **`(0.8, 0.8, 0.04)`** |
| nested body `pos="0 0 inf"` | success | position, size, offset | `(0.8, 0.8, inf)` |
| top body `pos="nan 0 0.75"` | success | position | `(0.8, 0.8, 0.76)` |
| top body `pos="inf 0 0.75"` | success | position | `(0.8, 0.8, 0.76)` |
| top body `quat="nan 0 0 1"` | success | quat | `(0.8, 0.8, 0.76)` |
| top body `euler="nan 0 0"` | success | quat | `(0.8, 0.8, 0.76)` |
| top body `axisangle="0 0 1 nan"` | success | quat | `(0.8, 0.8, 0.76)` |
| geom `pos="nan 0 0"` | **refused** | - | - |

The second row is the defect worth leading with. `(0.8, 0.8, 0.04)` is a 4 cm slab where the file
declares a 76 cm table, it is **byte-identical to the same fixture with the leg deleted**, and every
field the loader reports is finite - so unlike the geom spellings a consumer cannot screen for it. The
last row is the in-file control: the sibling guard already covers a geom's own placement, which is why
only the body placements remained.

Reachability is **wider** here than for the geom spellings, and this is the one place I do not simply
inherit that finding's argument. #2813 is scoped to fixtures because a non-finite geom makes the
inertia MuJoCo derives non-finite, so MuJoCo refuses a body with a free joint. A body's own `pos` is
not an input to that derivation, so MuJoCo compiles a non-finite body placement **with or without** a
free joint - movable task objects as well as static fixtures. Verified both ways.

## The change

The finiteness test moves into `_refuse_non_finite_placement`, the single owner that the geom locator
and a new body locator both delegate to, so neither can drift into tolerating what the other refuses.
`_refuse_non_finite_geom` keeps its name, signature and message, and all four of its call sites are
untouched. Three new call sites: the nested body's `pos` in `_recursive_collision_aabb`, and the
top-level body's `pos` and orientation in `load_mjcf_scene_objects`. The orientation refusal names the
spelling the file used, which is the rule `_geom_aabb` already follows.

The disposition follows the sibling rather than inventing one. `None` from these readers means "this
attribute was unreadable, use the documented default", and a caller cannot tell that apart from "this
attribute was read and is not a number".

`_geom_aabb`'s own docstring already named `_recursive_collision_aabb`'s accumulator as the step that
drops a non-finite value one level up - while that accumulator's own input went ungraded. Both
docstring claims are corrected in the same change, as is `load_mjcf_scene_objects`' `Raises:` block.

### Scope boundary

`_find_body_mesh` composes a nested body's `pos` into the mesh offset the same way, and it is
deliberately **not** guarded. Both walks descend the same `body_el.findall("body")` traversal from the
same root, so every body the mesh walk reads is a body the guarded accumulator also reads: a second
guard there would be unreachable, and because the mesh walk runs first it would mask the one whose
harm this change measures. That traversal identity is pinned by a test rather than asserted here.

## Evidence

Pre-fix split (behavioural revert - the helpers kept, the three call sites removed): **15 failed / 66
passed**, all 15 in the regression class, and **0** in all 6 premise cells, **0** in all 4 controls,
**0** in the scope-boundary class and **0** in every one of the 5 pre-existing geom classes. A raw
revert is a collection error instead, since the suite imports the helpers the change introduces.

Mutation table - 11 mutations, **9 distinct firing sets**, control 0, `collected` stable at 37 on every
row (so no row hides a deselection):

| mutation | fires | sibling suites | note |
|---|---|---|---|
| control | 0 | 0 | |
| nested-body `pos` guard dropped | 9 | 0 | |
| top-level `pos` guard dropped | 2 | 0 | |
| top-level orientation guard dropped | 4 | 0 | union of the three = **15**, disjoint, = the pre-fix split |
| body locator reports `geom` | 3 | 0 | |
| unnamed body borrows the geom type fallback | 1 | 0 | |
| orientation spelling hardcoded to `quat` | 3 | 0 | |
| finiteness test weakened to `any()` | 13 | **14** | it is the shared owner, so it breaks the geom paths too |
| body wrapper raises its own wording | 2 | **2** | trips the pre-existing one-owner test |
| guard reads only the first component | 9 | 0 | |
| guard placed after the offset is composed | **0** | 0 | measured no-op: it still refuses before the recursion |

Gate: `ruff check` and `ruff format --check` clean over 1635 files; `mypy` 24 errors, byte-identical to
the pre-change baseline on this tree with **0 attributable** to the changed files;
`tests/simulation/isaac/` **1804 passed, 5 skipped, 0 failed**.

## Visual

The reported collision proxy drawn over the geometry it claims to bound. One camera, one lighting
setup, one crop and one alpha for all three panels; the middle proxy leaves **8,106 of 16,782** of the
table's silhouette (48%) outside the box that is supposed to contain it, against **0 px** for the
finite placement.

reported collision proxy against the geometry it bounds

### DIFF
--- a/changelog.d/2827-non-finite-body-placement-is-refused.md
+++ b/changelog.d/2827-non-finite-body-placement-is-refused.md
@@ -1,0 +1,36 @@
+### Fixed: a body declaring a placement that is not finite is refused, not measured around
+
+`load_mjcf_scene_objects` composes a `SceneObject` from three placements: the top-level body's `pos`
+becomes the object's `position`, its orientation becomes `quat`, and each nested body's `pos` becomes
+the running offset every geom below it is measured from. `_geom_aabb` refuses a non-finite value on
+all four of a *geom*'s parse paths. The *body* placements the same bound is composed from were read
+straight through `_parse_xyz` / `_parse_orientation`, whose contract is to fall back to a documented
+default on an unreadable attribute and to return whatever a readable one parsed to.
+
+The nested case is the one nothing can detect. A non-finite offset makes every bound below it
+non-finite, and the running `min`/`max` in `_recursive_collision_aabb` orders a NaN as neither
+smaller nor larger than anything, so the whole subtree disappears from the union while the walk still
+reports that it found analytic geometry:
+
+    table, leg body pos="0 0 -0.37"  ->  size (0.8, 0.8, 0.76)
+    table, leg body pos="0 0 nan"    ->  size (0.8, 0.8, 0.04)
+    table with the leg deleted       ->  size (0.8, 0.8, 0.04)
+
+A 4 cm slab where the file declares a 76 cm table, byte-identical to the same fixture with the leg
+deleted, under a successful load - and with every field the loader reports finite, which is what
+separates it from the geom spellings: a consumer cannot screen for it. The other placements fail more
+loudly but no more correctly. A non-finite top-level `pos` or orientation is reported verbatim as the
+object's own, and `pos="0 0 inf"` on a nested body reports an infinite extent, because `inf - inf` is
+a NaN the outer accumulator drops in turn.
+
+Reachability is wider here than for the geom spellings. That finding is scoped to fixtures, because a
+non-finite geom makes the inertia MuJoCo derives non-finite and MuJoCo refuses a body with a free
+joint. A body's own `pos` is not an input to that derivation, so MuJoCo compiles a non-finite body
+placement with or without a free joint - movable task objects as well as the tables and cabinets
+whose footprint a manipulation scene is planned against.
+
+The disposition follows the sibling rather than inventing one: refuse, naming the body and the
+attribute the file used. The finiteness test moves into `_refuse_non_finite_placement`, the single
+owner the geom locator and the new body locator both delegate to, so neither can drift into
+tolerating what the other refuses; the guard's own docstring already named this accumulator as the
+step that drops a non-finite value, while the accumulator's own input went ungraded.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -1831,6 +1831,61 @@ def _segment_aabb(
     return center, half  # type: ignore[return-value]
 
 
+def _refuse_non_finite_placement(where: str, attribute: str, values: Sequence[float]) -> None:
+    """Refuse a placement or extent that parsed but is not finite.
+
+    The single owner of the finiteness test, so the geom paths and the body
+    paths cannot drift into tolerating what the other refuses. ``where`` is the
+    locator its caller built - :func:`_refuse_non_finite_geom` for a geom,
+    :func:`_refuse_non_finite_body` for a body - because the two elements are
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Extends the existing non-finite-refusal contract from geom placements/extents to the three body placements the same collision bound is composed from: a top-level body's `pos` and orientation (which become the `SceneObject`'s own `position`/`quat`), and each nested body's `pos` (the running offset `_recursive_collision_aabb` measures every geom below it from, where a NaN silently drops the whole subtree via the running `min`/`max` while every reported field stays finite). The finiteness test moves into a single owner (`_refuse_non_finite_placement`) that the existing geom locator and the new body locator both delegate to, with AST tests pinning that neither locator can drift into carrying its own wording. The refusal is an explicit `ValueError` naming the body and the attribute spelling the file used, matching the sibling's disposition; the `Raises:` block on `load_mjcf_scene_objects` is updated in the same change.

## What's good

- Verified locally at the head SHA: the two touched test files pass (81 tests), and the full `tests/simulation/isaac/` suite's delta against the merge base is zero (the 3 failures on this host pre-exist on `main` and are environment-only, missing Isaac Sim).
- The `is not finite` guard only grades values that parsed; the documented fall-back-to-default contract for an *unreadable* attribute is pinned unchanged (`test_an_unparseable_placement_still_falls_back_to_its_default`), and extreme-but-finite values stay accepted.
- Premise tests establish reachability (MuJoCo compiles a non-finite body placement, with and without a free joint) rather than asserting it in prose, and the mesh-walk coverage question is settled structurally (both walks descend the same traversal, so one guard answers for both).
- AGENTS.md compliance: changelog fragment present, ASCII-only strings, no host paths, behaviour-named tests, every new helper parameter documented under `Args:`, ruff check/format clean.

PR #2828, opened 2026-08-27
fix(drivers/g1): the stop verb reports the halt outcome instead of asserting one

cagataycali:
## The stop verb an agent reaches could not say the stop had failed

`G1Driver.stream({"action": "stop"})` built its envelope beside `await self.stop()` and hardcoded `status="success"`. Two separate things were wrong with the text it carried.

The **first** is what this PR opened for: it claimed no motion path was wired. That is stale - the transport primitive shipped some time ago. `DDSPublisher` publishes on `rt/lowcmd`, `send_action` builds and publishes a `LowCmd_`, the 500 Hz `_ControlLoop` rolls a policy out, and the loop publishes a zero-torque frame on exit.

The **second** is that the envelope could not report what the stop achieved. `stop` is the protocol's shutdown hook and returns `None`, so an envelope written next to it can only restate the intent. Replacing a stale refusal with `"control loop halted"` swaps a claim that is always out of date for one that is false on the case `_ControlLoop.stop`'s own docstring names:

> Returns `True` when the thread joined within `timeout`. `False` when the loop is still running - a caller-supplied policy that outlasts the join budget (a remote inference call is the ordinary case) needs the honest `stopped=False` in the `stop_task` envelope **rather than a `success` claim the payload's `running=True` contradicts**.

`stop_task` reads that verdict. `cleanup` and `stop` were taught to read it. The verb an *agent* reaches was the one surface left that could not.

### Measured, same driver, same wedge

A policy parked past the 2.0 s join budget, on a driver in the state healthy hardware produces (`_connected`, `_mode_machine=9`, battery 92 %, FSM in the allowed set):

| surface | elapsed | status | payload |
|---|---|---|---|
| `stream({"action": "stop"})` | 2.00 s | **`success`** | text only; loop still running |
| `stop_task()` | 2.00 s | `error` | `stopped=False`, `running=True`, reason names the timeout |

One driver, two stop surfaces, two contracts. `stop` does log the overrun at ERROR, so the information existed and the envelope discarded it.

### The change

The verb returns `stop_task`'s envelope, so the verdict has one owner rather than two:

- a joined stop is still `success` and now carries `stopped=True`
- an unjoined one is an `error` naming the timeout
- a driver with nothing running says so, instead of claiming a halt it did not perform

`stop` keeps its signature: it is the `HardwareDriver` shutdown hook the mesh calls, and `-> None` is why it cannot answer here. Re-deriving the verdict from `self._loop.is_running` was the alternative and is worse - it puts the decision in two places and can drift from the surface a caller polls.

The `inputSchema` description promises the report the verb now delivers.

### Why nothing caught it

The shipped cell drives `G1Driver(tool_name="g1", port="1.2.3.4")` - a driver with no loop at all - so it grades the string and cannot reach either running outcome. That is also why `"control loop halted"` was its own second falsity there: nothing was halted. The new cells park a policy past the budget, which costs about two seconds each; the fast path cannot distinguish a reported outcome from an asserted one.

### Gate

- **Pre-fix split** (source reverted to this branch's previous commit): **9 failed / 81 passed** - 9/9 regression, **0 of 9** premise + over-reach control.
- **Mutations**: 8 rows, **7 distinct firing sets**, none undetected, control 0, `collected` stable at 90, restored md5-identical. Failures in the three sibling g1 suites: **0 on every row**.

| row | fires | mutation |
|---|---|---|
| b0 | 0 | control |
| b1 | 8 | revert the branch: `await stop()` + hardcoded success text |
| b2 | 1 | delegate, then force `status="success"` (the "a tool must not error" fix) |
| b3 | 1 | delegate, then drop the timeout reason |
| b4 | 1 | `await stop()` first, then delegate - the double stop makes a joined rollout report "no task is running" |
| b5 | 5 | delegate but flatten the payload to text only |
| b6 | 1 | schema description reverted (behaviour untouched) |
| b7 | 1 | delegate but restate the halt claim in the branch |

`b1 | b6` is exactly the 9 pre-fix failures, disjoint - the behaviour and the schema promise are independently pinned.

- **Paired suite**, identical 353-file selection (all `tests/drivers/`, `tests/tools/g1/`, and every test that walks the tree, parses source, or reads docstrings), the new file excluded from both trees: identical **17494 collected** and `18 failed, 17413 passed, 63 skipped` on both, **18 == 18 failing ids with `comm` empty in both directions**, distinct rootdirs. Of the 18, 17 need `awsiot`/`awscrt` (`find_spec` False locally, declared in the `mesh-iot` extra) and 1 is the lerobot-from-source `encode()` drift.
- **mypy** `24 == 24` against this branch's previous commit, normalized message sets byte-identical in both directions, **0 attributable**.
- `tests/drivers/ tests/tools/g1/`: **735 passed / 3 skipped** with the SDK present, **698 passed / 40 skipped / 0 failed** with `unitree_sdk2py` hidden behind a `find_spec` + `meta_path` blocker, so the CI box's disposition is measured rather than assumed (698 + 40 == 735 + 3: same collection, only disposition moves).

No visual artifact: this changes a tool envelope, not a policy, simulation, render, recording or asset.

### Note on the preceding commit

`f0d70c42` wrapped the over-long text line to fix `Run lint`. This change deletes that block, so the line cannot return; the commit is preserved in history and `upstream/main` is merged in rather than rebased over it.

### DIFF
--- a/changelog.d/2828-g1-stop-reports-the-halt-outcome-not-a-stale-refusal.md
+++ b/changelog.d/2828-g1-stop-reports-the-halt-outcome-not-a-stale-refusal.md
@@ -1,18 +1,34 @@
-### Docs: g1 driver's ``stop`` verb reports the halt outcome, not a stale #358 refusal
+### Fixed: g1 driver's ``stop`` verb reports the halt outcome instead of asserting one
 
-``G1Driver.stream({"action": "stop"})`` mapped to a text envelope claiming
-"no motion path wired yet (issue #358)". That claim was stale: issue #361
-already landed the transport primitive - :class:`DDSPublisher` on
-``rt/lowcmd``, :meth:`send_action` publishing ``LowCmd_``, the 500 Hz
-control loop, and :meth:`stop_task` publishing a zero-torque frame on the
-way out. The verb was calling the wired :meth:`stop` and lying about it.
+``G1Driver.stream({"action": "stop"})`` built its envelope beside
+``await self.stop()`` and hardcoded ``status="success"``. Two things were wrong
+with the text it carried. It named issue #358 and claimed no motion path was
+wired, which is stale - #361 landed :class:`DDSPublisher` on ``rt/lowcmd``,
+:meth:`send_action`, the 500 Hz control loop and the zero-torque frame the loop
+publishes on exit. And the envelope could not report what the stop achieved,
+because ``stop`` is the protocol's shutdown hook and returns ``None``: an
+envelope written next to it can only restate the intent.
 
-Now the envelope names what the code does: "control loop halted; a running
-task publishes a zero-torque frame on exit". The ``inputSchema`` description
-for the ``stop`` action is updated to match, and
-``test_stream_stop_action_calls_stop`` pins both the new text and a guard
-against the pre-#361 refusal text sneaking back through a rebase.
+That mattered on the case :meth:`_ControlLoop.stop`'s own docstring names. It
+returns whether the thread joined, and a caller-supplied policy that outlasts
+the join budget - a remote inference call is the ordinary case - leaves the loop
+publishing frames. ``stop_task`` already reads that verdict and answers
+``status="error"`` with ``stopped=False`` and ``running=True``; ``cleanup`` and
+``stop`` were taught to read it too. The verb an agent reaches was the one
+surface left that could not say the stop had failed:
 
-No behaviour change - the wire path was already live. This is a docs and
-observability fix so an agent that reads the tool spec sees the surface it
-gets.
+```text
+policy parked past the 2.0s join budget, same driver, same wedge
+
+  stream({"action": "stop"})   status="success"   text only, loop still running
+  stop_task()                  status="error"     stopped=False, running=True
+```
+
+The verb now returns ``stop_task``'s envelope, so the verdict has one owner
+rather than two: a joined stop is still ``success`` and carries
+``stopped=True``, an unjoined one is an ``error`` naming the timeout, and a
+driver with nothing running says so instead of claiming a halt it did not
+perform. The ``inputSchema`` description promises the report the verb now
+delivers, and ``tests/drivers/test_g1_stream_stop_reports_the_halt_outcome.py``
+parks a policy past the budget to grade it - the fast path the shipped suite
+drives cannot distinguish a reported outcome from an asserted one.
--- a/changelog.d/2828-g1-stop-reports-the-halt-outcome-not-a-stale-refusal.md
+++ b/changelog.d/2828-g1-stop-reports-the-halt-outcome-not-a-stale-refusal.md
@@ -1,0 +1,18 @@
+### Docs: g1 driver's ``stop`` verb reports the halt outcome, not a stale #358 refusal
+
+``G1Driver.stream({"action": "stop"})`` mapped to a text envelope claiming
+"no motion path wired yet (issue #358)". That claim was stale: issue #361
+already landed the transport primitive - :class:`DDSPublisher` on
+``rt/lowcmd``, :meth:`send_action` publishing ``LowCmd_``, the 500 Hz
+control loop, and :meth:`stop_task` publishing a zero-torque frame on the
+way out. The verb was calling the wired :meth:`stop` and lying about it.
+
+Now the envelope names what the code does: "control loop halted; a running
+task publishes a zero-torque frame on exit". The ``inputSchema`` description
+for the ``stop`` action is updated to match, and
+``test_stream_stop_action_calls_stop`` pins both the new text and a guard
+against the pre-#361 refusal text sneaking back through a rebase.
+
+No behaviour change - the wire path was already live. This is a docs and
+observability fix so an agent that reads the tool spec sees the surface it
+gets.
--- a/strands_robots/drivers/g1.py
+++ b/strands_robots/drivers/g1.py
@@ -349,7 +349,8 @@ class G1Driver:
                                 "description": (
                                     "sensors: return the latest cached IMU/battery/lidar; "
                                     "status: report connection and FSM; "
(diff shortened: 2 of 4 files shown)


yinsong1986 (dismissed):
## Summary

Docs/observability-only fix: the `G1Driver.stream({"action": "stop"})` envelope, the `inputSchema` description for the `stop` action, and the `stream()` docstring all claimed `stop` was an unwired no-op pending #358, while the code path already calls the wired `stop()` (control-loop join + zero-torque frame via `stop_task`, landed in #361). The PR replaces the stale refusal text with an accurate description of the halt outcome and pins the new text in `test_stream_stop_action_calls_stop`, including a guard against the pre-#361 text returning through a rebase. Verified against the head tree that `stream()`'s `stop` branch does call `await self.stop()` and that `stop()` performs the controlled halt the new text describes - so this is genuinely a no-behaviour-change PR, matching the description.

## What's good

- Changelog fragment added under `changelog.d/` per AGENTS.md PR workflow step 3, rather than editing `CHANGELOG.md`.
- The updated envelope/schema strings are plain ASCII (project Unicode-hygiene rule).
- The test pins both the new text and the absence of the stale text, so the fix cannot silently regress through a rebase.
- Tight scope: 3 files, no behaviour change, no wire/schema shape change (the `enum` and parameter names in the tool spec are untouched; only the human-readable description changed).

cagataycali:
`call-test-lint` failed at step 8 `Run lint`, which means step 9 `Run tests` was
**skipped** - so this branch's test state has not been graded by anything. Measured
locally on the pushed head `bbe4d423`:

**The blocker is `ruff format`, not `ruff check` and not mypy.** One line:

```
ruff check strands_robots tests tests_integ   -> All checks passed!
ruff format --check ...                       -> Would reformat: strands_robots/drivers/g1.py
                                                 1 file would be reformatted, 1633 already formatted
```

```diff
@@ -406,7 +406,9 @@
             await self.stop()
             envelope = {
                 "status": "success",
-                "content": [{"text": "stop: control loop halted; a running task publishes a zero-torque frame on exit"}],
+                "content": [
+                    {"text": "stop: control loop halted; a running task publishes a zero-torque frame on exit"}
+                ],
             }
```

**The tests that CI skipped are green, in both dependency modes.** `unitree_sdk2py`
is not installable in CI, so the mode that matters there is the second one:

```
tests/drivers/ + tests/tools/g1/ + tests/test_driver_seam.py

  SDK importable   763 passed,  3 skipped
  SDK unimportable 726 passed, 40 skipped
```

The totals agree (`763 + 3 == 726 + 40 == 766`), so the same cells were collected
either way and only their disposition moved - the 37 additional skips are the
wire-level assertions that need the SDK.

So this is one `ruff format strands_robots/drivers/g1.py` away from green. Worth
noting before that push lands: the approval on this PR is dismissed by any push
(`dismiss_stale_reviews_on_push`), so it will need a fresh one afterwards.

Not pushing it myself - the branch has commits in flight.

yinsong1986 (dismissed):
## Summary

Docs/observability-only fix: the `stop` verb of `G1Driver.stream()` was calling the wired `stop()` path (control-loop join, with the loop publishing a zero-torque frame from its own `finally` - landed in #361) while the tool-result envelope, the `inputSchema` description, and the `stream()` docstring all still claimed "no motion path wired yet (issue #358)". The PR replaces the stale refusal text with an accurate description of the halt outcome in all three places, removes a duplicated stale section header, and updates `test_stream_stop_action_calls_stop` to pin the new text plus a guard against the pre-#361 refusal returning through a rebase. Verified against the head tree that `stop()` (g1.py:503) does what the new text says: it halts a running `_ControlLoop` and that loop publishes `_build_zero_torque_lowcmd` on exit via `stop_task`'s path. No behaviour change, no wire-format or schema-shape change (the `enum`/parameter names are untouched; only the human/agent-readable `description` and result text move).

## What's good

- Changelog fragment under `changelog.d/` per AGENTS.md step 3, rather than an append to `CHANGELOG.md`.
- The regression test pins both directions: the new text is present, and the stale refusal text (`"no motion path wired"`, `"#358"`) is asserted absent, so a rebase cannot silently reintroduce the lie.
- New tool-result strings are plain ASCII, matching the string-hygiene rule for text an agent reads back out of its own tool call.
- Scope discipline: the only non-doc line touched is the envelope literal the PR exists to fix.

yinsong1986 (approved):
## Summary

The agent-facing `stop` verb of `G1Driver.stream()` previously built a hardcoded `status="success"` envelope beside `await self.stop()` - a shutdown hook that returns `None` - so it could neither drop the stale pre-#361 "no motion path wired" claim nor report the one failure `_ControlLoop.stop`'s docstring names: a policy that outlasts the 2.0 s join budget leaves the loop publishing frames while the agent reads success. The fix delegates the branch to `stop_task()`, which performs the same halt and already owns the verdict, so a joined stop stays `success` with `stopped=True`, an unjoined one is an `error` naming the timeout with `stopped=False`/`running=True`, and a driver with nothing running says so. The `inputSchema` description and `stream()` docstring are updated to match, and the new `tests/drivers/test_g1_stream_stop_reports_the_halt_outcome.py` parks a policy past the join budget to grade both running outcomes - the path the shipped fast-path suite cannot reach.

## What's good

- The verdict is single-sourced: the verb returns `stop_task`'s envelope rather than re-deriving `stopped` from `self._loop.is_running`, so the two stop surfaces an agent and a caller poll cannot drift apart.
- `stop()` keeps its `-> None` `HardwareDriver` signature; the protocol hook is untouched and the error envelope stays inside the tool-result contract (no raise across the tool surface).
- The regression suite pins both directions - unjoined-reports-error and joined-still-reports-success over-reach controls - plus premise cells and stale-text guards, and reuses the sibling's `_Rollout`/SDK-stub fixtures instead of forking them.
- Changelog fragment present under `changelog.d/2828-...` per AGENTS.md step 3; test classes are behaviour-named; new strings are ASCII.

PR #2830, opened 2026-08-27
fix(ci): collide two branches on the change they describe, not only on the file they create

### LINKED ISSUE #2823: ci(duplicate-claim): the added-path key is silent on two branches that fix one defect in one file - #2820 and #2822 reported unique-additions
## What happened

#2820 and #2822 were opened 13 minutes apart, fixed the same defect in the same file, and both modified `strands_robots/tools/serial_tool.py`. `scripts/check_duplicate_claim.py --all-open` reported `unique-additions` over that open set while both were in it:

```
Outcome: **unique-additions**
No two of the 4 open pull requests create the same file or name the same changelog entry (6 pair(s) compared).
```

#2822 has since been closed as the duplicate. The check was run against the live open set at the time both were open, so this is the relation's answer to a real instance of what it exists to find, not a reconstruction.

## Why it is silent

Both keys read **added** paths only:

| key | #2820 | #2822 | collides |
|---|---|---|---|
| added path | `tests/tools/test_feetech_broadcast_is_not_a_reply_address.py` | `tests/tools/test_feetech_motor_id_excludes_the_broadcast.py` | no |
| changelog slug | `feetech-broadcast-is-not-a-reply-address` | `feetech-motor-id-excludes-the-broadcast` | no |

Two agents describing one defect pick two names for the test and two slugs for the entry, so neither key touches. What they cannot avoid sharing is the file the defect is in:

```
shared ADDED    : []
shared MODIFIED : ['strands_robots/tools/serial_tool.py',
                   'tests/tools/test_serial_tool_numeric_domain.py']
```

Both also corrected the same case in the same pre-existing test file, independently.

## The exclusion is wider than the reason given for it

`ADDED_CHANGE_TYPE` documents why the other change types are not read:

> GitHub's enum also carries `MODIFIED`, `REMOVED`, `RENAMED`, `COPIED` and `CHANGED`, and every one of those describes a file that already exists on the base - which is the sibling sweep's composition question rather than this one. Reading them here is what turns a 2-of-1802 relation into a 117-of-1802 one.

Both halves are true of **any** changed path, which is what that 117-of-1802 measurement covers. Neither is a statement about a modified path under `strands_robots/`. A pair sharing an added path is two branches creating one file; a pair modifying one **source** file is not obviously the composition question either - the composition sweep asks whether two changes interact, and this asks whether they are the same change.

## Candidate relation, precision unmeasured

Two co-open pull requests that both modify the same file under `strands_robots/`.

For this pair that is exactly one path, `strands_robots/tools/serial_tool.py`. I have not measured its precision over the corpus, and it should not be added until someone has - the existing 117-of-1802 figure is the reason to be careful, not a reason to skip the measurement. The recipe is the one already in this module's history: enumerate co-open pairs as those whose `[createdAt, closedAt or now]` intervals overlap, take each pull request's files with `changeType`, and count how many selected pairs are genuine duplicates against how many are two unrelated changes to a shared module. `strands_robots/utils.py`, `simulation/base.py` and `mesh/core.py` are the paths most likely to make it noisy, and an exclusion list for high-traffic modules may be what makes it usable.

Worth checking at the same time whether the slug key can be strengthened. The two slugs here share the token `feetech` and both share `broadcast`; a shared-token relation over slugs is cheaper than a path relation and reads only data the sweep already fetches, but "two slugs share a word" is plainly noisier than "two slugs are equal" and needs the same corpus measurement.

## Not filing a pull request

Which relation to add is a precision question with a number attached, and I do not have the number. Adding a key that fires on every pair of branches touching `utils.py` would make the sweep worse than silent. Recording the instance and the candidate so the measurement can be made against a real example rather than a hypothetical one.

cagataycali:
## The defect

`scripts/check_duplicate_claim.py --all-open` pairs open branches on what they
**create** — a path, or the slug of a changelog fragment. That key reads a *name*,
and a name is exactly what two authors describing one change need not share.

#2820 and #2822 fixed one defect thirteen minutes apart, both modifying
`strands_robots/tools/serial_tool.py`, and named the change
`feetech-broadcast-is-not-a-reply-address` and
`feetech-motor-id-excludes-the-broadcast`. Two names, no collision. Run against
the live open set with both in it, the sweep answered:

```
Outcome: **unique-additions**
No two of the 4 open pull requests create the same file or name the same
changelog entry (6 pair(s) compared).
```

#2822 was later closed as the duplicate. So this is the relation's answer to a
real instance of what it exists to find.

## What the second key is

Two branches whose changelog fragments share at least **two words** *and* which
both edit at least one **pre-existing test**. Two authors fixing one defect
describe the same subject and correct the same case. The conjunction is the
relation — neither half is usable alone.

## Measurement

Ground truth is the **eleven** pairs in #2345 through #2825 whose closed half
names, in its own closing comment, the pull request that superseded it (`#2373 ->
#2370`, `#2383 -> #2384`, `#2389 -> #2388`, `#2431 -> #2429`, `#2571 -> #2570`,
`#2707 -> #2708`, `#2766 -> #2767`, `#2785 -> #2792`, `#2787 -> #2792`, `#2790 ->
#2792`, `#2822 -> #2820`). All eleven were open at the same instant. #2646 is
excluded: its comment says "superseded by the 0.5 -> 1.0 backlog sequencing, not
because it is blocked", which is a scope change rather than a duplicate.

Over the 2199 pairs open at the same instant in that window, and over 374
replayed sweeps — one at each moment a pull request was opened with another
already open, which is what `--all-open` actually reads (median open set: six):

| relation | pairs | precision | recall | sweeps firing |
|---|---|---|---|---|
| both edit a source file | 80 | 8.8% | 63.6% | 29.9% |
| fragments share one word | 151 | 6.6% | 90.9% | — |
| fragments share two words | 33 | 30.3% | 90.9% | 37.2% |
| both edit one test | 26 | 11.5% | 81.8% | — |
| **two words AND one test** | **14** | **64.3%** | **81.8%** | **5.6%** |
| created path (the existing key) | 7 | 71.4% | 45.5% | 1.3% |
| both keys together | 15 | 66.7% | 90.9% | 5.9% |

The new key reaches four pairs the created-path key structurally cannot:
`#2370/#2373`, `#2383/#2384`, `#2429/#2431`, `#2820/#2822`.

## Two results decided the shape

**Pairing on a shared edited source file is refused, with a number.** It is the
obvious widening: 8.8% precise, and it fires on nearly a third of sweeps.
`strands_robots/simulation/mujoco/simulation.py` alone accounts for 17 of its 80
pairs. That is worse than the 117-of-1802 figure `ADDED_CHANGE_TYPE` already
refuses on precision grounds, so it is asserted from both sides rather than left
as a comment.

**One shared word is the repository's house style, not a subject.** `names` and
`the` appear in 39 and 20 of the window's 401 fragments, so a one-word rule fires
on 37.2% of sweeps. Three words loses a third of the duplicates (63.6% recall).
Only the conjunction is precise, and it needs no stop list, no path exclusions and
no parameter tuned against the corpus it was measured on — a frequency-derived
stop list is a refinement rather than a requirement (43.5% against 30.3% at the
same recall) and cannot be computed where this runs, since the sweep sees a median
of six pull requests rather than a corpus.

## The two keys are reported apart

They support different conclusions, so they are not merged into one list:

- a shared **created path** is a fact — neither file exists on the base, so one of
  the two branches is redundant;
- a shared **description over a shared test** is a question. At 64.3% precision
  roughly one pair in three shares a subject without sharing a change, so the
  report says so and asks the reader to read both descriptions.

The existing report's reasoning ("Neither branch's file exists on the base, so
this is not a merge order to decide") is *false* of an edited test, so it is not
attached to an echo. Of the five non-duplicate pairs the new key selects, two are
members of the `#2785`/`#2787`/`#2790`/`#2792` supersede cluster and one
(`#2713`/`#2714`) is two branches on one subject that both shipped.

## A denominator correction

Eleven declared duplicate pairs is more than the five this module counted before,
because a closing comment reaches pairs neither key found: `#2370/#2373`,
`#2383/#2384` and `#2429/#2431` are three duplicate pairs in the earlier window
that the module's own table does not list. So the created-path key's recall is
45.5% of the declared set, not most of it. Both the module docstring and AGENTS.md
step 1 now say so.

## Tests

`tests/test_duplicate_work_described_change_key.py`, 57 cases. Every path in
`_ECHO_PAIRS` is the real one from `pulls/<n>/files`, which outlives the state
change that closed one half of each pair.

Pre-fix split, with the relation neutered and every new symbol kept:
**18 failed / 38 passed**, and the failures are 12 in
`TestTheMeasuredEchoPairsAreReported`, 4 in `TestTheTwoKeysReportApart`, 2 in
`TestDeterminism` — **0 in all four premise and control classes**. Against the
current `main` script the whole file is 56 failures, which is the collection-error
shape of a raw revert.

`TestTheCreatedPathKeyCannotSeeThesePairs` is the non-vacuity half: every pair
here reports `unique-additions` when only what the branches create is read, so a
test that the sweep reports them cannot pass on the first key alone.

Mutation table — 12 mutations, 12 detected, control clean, `collected` stable at
57 in every row, source restored byte-identically:

| mutation | new file | sibling suites |
|---|---|---|
| control | 0 | 0 |
| the relation reports nothing | 18 | 0 |
| floor lowered to one word | 2 | 0 |
| floor raised to three words | 14 | 0 |
| shared-test half dropped (words alone) | 3 | **1** |
| shared-words half dropped (test alone) | 1 | 0 |
| reads an edited source file instead of a test | 19 | 0 |
| echoes merged into the created-path list | 1 | 0 |
| a fragment contributes its whole slug, not words | 21 | 0 |
| a reserved name contributes words | 1 | 0 |
| the fragment number becomes a word | 2 | 0 |
| a non-fragment path contributes words | 1 | 0 |

Dropping the shared-test half trips a pre-existing test as well, which is the
over-reach evidence for keeping the conjunction.

The last row was 0 until a `docs/robots/2820-feetech-broadcast.md` exemplar was
added: `docs/robots/` is exactly as long as `changelog.d/`, so that path's tail is
a well-formed fragment name and the directory check is load-bearing rather than
belt-and-braces.

## Gate

  tests_integ` and the script; `mypy` **`Success: no issues found in 1632 source
  files`**.
- Paired against the merge-base on an identical 434-file selection — every suite
  that walks the tree, parses source, or reads docstrings, `AGENTS.md`,
  `changelog.d/` or `scripts/` — with the new file excluded from both:
  **20492 collected and `5 failed, 20414 passed, 73 skipped` on each**, the same
  five failing ids, both `comm` directions empty. All five are pre-existing on a
  pristine `main` worktree: they assert `rclpy` is absent and it resolves here.
- Run against the live open set, the sweep reports `unique-additions` over four
  pull requests and six pairs, exit 0 — no false positive on the current queue.

No visual artifact: this changes a repository check's report, not a policy,
simulation, rendering, recording or asset path. The measured tables are the
evidence.

Closes #2823.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -318,6 +318,21 @@ hatch run format            # ruff check --fix, ruff format
    that window which add a fragment there are **350 distinct slugs**, and the only
    two used twice are the two duplicate pairs above that reused one.
 
+   That key reads a **name**, though, and a name is what two authors describing one
+   change need not share. #2820 and #2822 fixed one defect thirteen minutes apart
+   and called it `feetech-broadcast-is-not-a-reply-address` and
+   `feetech-motor-id-excludes-the-broadcast`; two names, so the sweep reported
+   `unique-additions` with both open (#2823). So the same command asks a second,
+   weaker question beside it: do two branches' fragments share at least **two
+   words** *and* both edit one pre-existing test? Two authors fixing one defect
+   describe the same subject and correct the same case, and the conjunction is what
+   makes that usable - measured over the 2199 co-open pairs in #2345 through #2825,
+   against the eleven pairs whose closed half names its supersedor, shared words
+   alone is 30.3% precise and fires on 37.2% of sweeps, a shared edited test alone
+   is 11.5%, and together they are **64.3% precise at 81.8% recall**, firing on
+   5.6%. Read the report's second section as a pair to check rather than a fact:
+   roughly one in three is two branches on one subject without one change.
+
    Two branches *editing* one file is a different question with a different remedy
    (a merge order, possibly one composition run), and
    `scripts/check_merge_base_overlap.py --github-repo <owner/name> --all-open`
--- a/changelog.d/2830-duplicate-sweep-reads-a-described-change.md
+++ b/changelog.d/2830-duplicate-sweep-reads-a-described-change.md
@@ -1,0 +1,31 @@
+### Fixed: the duplicate-work sweep collides two branches on the change they describe, not only on the file they create
+
+`scripts/check_duplicate_claim.py --all-open` paired open branches on what they
+create - a path, or the slug of a changelog fragment. That key reads a name, and a
+name is what two authors describing one change need not share: #2820 and #2822
+fixed one defect thirteen minutes apart and called it
+`feetech-broadcast-is-not-a-reply-address` and
+`feetech-motor-id-excludes-the-broadcast`, so the sweep reported
+`unique-additions` while both were open.
+
+The sweep now asks a second question beside the first: do two branches' fragments
+share at least two words *and* do both edit one pre-existing test? Two authors
+fixing one defect describe the same subject and correct the same case. The
+conjunction is the relation - neither half is usable alone, and the report keeps
+the two keys apart because they support different conclusions. A shared created
+path is a fact about the branches; a shared description over a shared test is a
+pair to read.
+
+Measured over the 2199 pairs open at the same instant in #2345 through #2825,
+against the eleven pairs whose closed half names, in its own closing comment, the
+pull request that superseded it: the created-path key alone reaches five of the
+eleven, the new key reaches nine, and the two together reach ten. Shared words
+alone would select 33 pairs and fire on 37.2% of replayed sweeps - the repository
+names its changes in a house style, so `names` and `the` collide constantly - and
+a shared edited source file, the obvious widening, is 8.8% precise and fires on
+29.9%. The conjunction selects 14 pairs at 64.3% precision and fires on 5.6%.
+
+The measurement also corrects a denominator: eleven declared duplicate pairs is
+more than the five this module counted before, because a closing comment reaches
+pairs neither key found. The created-path key's recall is 45.5% of that set, not
+most of it.
--- a/scripts/check_duplicate_claim.py
+++ b/scripts/check_duplicate_claim.py
@@ -82,9 +82,9 @@ Three questions, two keys
     always shares its own claim.
 
 ``--all-open``
-    Review, keyed on what a branch creates. Do two open pull requests create the
-    same file, or two fragments naming one changelog entry? Needs no issue
-    number, which is the point: see below.
+    Review, keyed on what a branch creates and what it says. Do two open pull
+    requests create the same file, name one changelog entry, or describe one
+    change over one test? Needs no issue number, which is the point: see below.
 
 The key a claim-free pair collides on
 -------------------------------------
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Adds a second, deliberately weaker key to `scripts/check_duplicate_claim.py --all-open`: two open branches whose changelog fragments share at least two slug words *and* which both edit one pre-existing test under `tests/` are reported as an "echo" (one change described twice), covering the #2820/#2822 shape the created-path key cannot see because the two authors chose different fragment slugs. The conjunction is measured (64.3% precision at 81.8% recall over the 2199 co-open pairs in #2345..#2825), the rejected alternatives (one shared word, shared edited source file, three-word floor) are documented with their measured numbers, and the two keys are reported apart in the output because they support different conclusions - the echo section explicitly labels itself the weaker key and does not inherit the created-path reasoning. Mechanically the change is contained: `added_paths`/`resolve_open_additions` become `file_sets`/`resolve_open_file_sets` returning a `PullFiles` (created + edited) read from the same GraphQL response, with truncation refused once for both keys, and no caller outside the script and its two test files references the renamed functions. Exit-status semantics are unchanged in shape (finding -> 1, unreadable -> 0). Verified locally at the head SHA: both test files pass (145 tests), `ruff check` and `ruff format --check` are clean, and no non-ASCII characters appear in the report strings.

## What's good

- The `edits`-as-part-of-one-argument design (`classify_additions` takes `Mapping[int, PullFiles]`, not an optional second mapping) closes the silent-no-op failure mode the module refuses everywhere else, and `TestTheNodeReaderSplitsOneFileListByChangeType` pins that a truncated list is refused for both keys at once.
- `TestTheCreatedPathKeyCannotSeeThesePairs` pins non-vacuity - each measured echo pair reports `unique-additions` on the first key alone - so the second key cannot quietly become a tightening of the first.
- Both halves of the conjunction have load-bearing controls in each direction, including the `docs/robots/` prefix-length trap and the assembler-authority reserved-name case.
- The change follows AGENTS.md step 3 (fragment, no direct `CHANGELOG.md` edit) and updates the AGENTS.md step-1 guidance in the same PR.

yinsong1986 (approved):
## Summary

Adds a second, deliberately weaker key to `scripts/check_duplicate_claim.py --all-open`: two open branches whose changelog fragments share at least `FRAGMENT_TOKEN_FLOOR` (2) slug words *and* which both edit at least one pre-existing test under `tests/` are reported as an echo (one change described twice), covering the #2820/#2822 shape the created-path key cannot see. The two keys are kept apart in `AdditionVerdict` (`collisions` vs `echoes`) and in the rendered report, so the created-path reasoning ("neither file exists on the base") is never attached to an echo, and the echo section states its own 64.3% precision so the reader treats it as a pair to check rather than a fact. Verified at head: `fragment_tokens` slicing is sound against the assembler's `FRAGMENT_NAME` regex (number excluded, no empty tokens, reserved names and same-length non-fragment directories filtered); nothing outside this script and its two test suites calls the renamed `resolve_open_additions`/`added_paths` surfaces and no workflow invokes the script, so the CLI flags, outcome tokens, and exit-code semantics are the only contract and they are unchanged; a truncated file list is still refused once for both keys rather than read short; the 145 tests in the two duplicate-work suites and the 100 in the neighbouring claim-check suites pass, and ruff check/format are clean on every touched file. Changelog fragment present per AGENTS.md step 3, and the AGENTS.md step-1 addition matches the shipped behaviour.

PR #2831, opened 2026-08-27
fix(device-connect): report an emergency stop the robot refused instead of discarding it

cagataycali:
## The defect

`RobotDeviceDriver.onEmergencyStop` called `self._robot.stop_task()` and dropped the envelope.

Two of the four robot surfaces this driver can wrap answer a stop with an affirmative "I did not stop", and each was written deliberately to do so:

- `G1Driver.stop_task` returns `status="error"` with `stopped=False` when its control loop outlasts the join budget. Its own docstring says why: *"so the caller cannot read 'success' while the payload's own `running=True` says the loop is still writing frames"*, and the inline comment adds *"a caller that reads only `status` cannot count the task as stopped"*.
- `ReachyDriver.stop_task` reports a daemon that refused the stop.

This handler was the caller that read neither field.

## Measured, on a robot that affirmatively refused

Driving `onEmergencyStop` from an authorized source, with the robot answering `{"status": "error", "content": [{"json": {"stopped": false, "running": true, "steps": 4211, "reason": "control loop did not join within timeout"}}]}`:

| | before | after |
|---|---|---|
| `stop_task` called | 1 | 1 |
| handler returned | `None` | `None` |
| records at ERROR or above | **0** | 1 |
| anything naming the refusal | **nothing** | the reason, the source, and the remedy |

Before, the only record was `WARNING  Emergency stop received from safety-controller-1 - stopping task` — the intent, not the outcome. After:

```
CRITICAL [safety] emergency stop from safety-controller-1: the robot reported it did NOT
         stop: stop_task: control loop did not join within timeout; policy is likely
         blocking. It may still be executing; use a hardware cutoff.
```

## The in-tree control: one fleet, one envelope, two answers

`Mesh.emergency_stop` already grades this verdict for every peer it fans out to. Passing the **same envelope** through both paths:

| path | verdict |
|---|---|
| `mesh.core._peers_that_did_not_stop` | `{'g1-01'}` — reports it did not stop |
| `RobotDeviceDriver.onEmergencyStop` | discarded, nothing above WARNING |

`_peers_that_did_not_stop`'s docstring gives the reasoning this change adopts: *"An emergency stop is only as trustworthy as its accounting"*, and it explicitly enumerates *"a hardware stop that failed answers `{"ok": False, "error": "stop_task failed: ..."}`"* as a shape that must not be counted as an acknowledgement. An operator's accounting should not depend on whether the stop arrived over the mesh or over Device Connect.

## Why both signals are read

`teleop_mixin._stop_reported_stopped` asks the same question of a `stop_teleoperate` envelope and is **deliberately not reused**. Measured against the two real `stop_task` failure shapes:

| envelope | `status` | `_stop_reported_stopped` | teleop reader alone |
|---|---|---|---|
| G1 join timeout | `error` | `False` | catches it |
| Reachy daemon refused | `error` | `True` | **misses it** |
| hardware idle | `success` | `True` | correct |
| G1 clean stop | `success` | `True` | correct |

It answers `True` for an envelope with no `json` block, which is right there (nothing was teleoperating) and wrong here, because that is exactly the shape a refused `stop_task` arrives in. So an explicit `stopped` flag is authoritative where a driver supplies one, and the envelope's status answers for the drivers that report through `_refuse`.

## Conservatism

Only an envelope that **affirmatively** reports a failure is flagged, for the reason `_peers_that_did_not_stop` states: a false "did not stop" on the safety path trains operators to ignore the warning. Left alone: a robot that stopped, one that had nothing to stop, one reporting an error beside a payload saying the loop stopped, a driver returning `None`, and a test double. The authorization guard still runs first, so reading the verdict does not widen who can halt a task.

## Why nothing caught it

The three suites that already drive `onEmergencyStop` assert `stop_task.assert_called_once()` or `robot.stopped is True` — that the **call was made**, never what it **answered** — and their robot doubles are `MagicMock`s whose `stop_task` returns a `MagicMock`, which carries no verdict to read. Every mutation below fires **0** of those 175 pre-existing cells except the one that reorders the authorization guard.

## Pre-fix split

`7 failed / 10 passed`, by class:

|---|---|---|
| `TestARobotThatDidNotStopIsReported` | regression | **6 / 6** |
| `TestTheHandlerReadsTheVerdictRatherThanDiscardingIt` | structural | 1 / 2 |
| `TestTheRefusalShapesAreReal` | premise | 0 / 3 |
| `TestAStopThatHappenedIsQuiet` | over-reach control | 0 / 4 |
| `TestTheAuthorizationGuardStillComesFirst` | control | 0 / 2 |

The structural class's second cell is the mesh control, which correctly holds both ways.

## Mutation table

11 mutations, control clean, **10 distinct firing sets**, `collected` stable at 18 on every row (so no row hides a deselection). `sib` counts failures in the three pre-existing `onEmergencyStop` suites.

| mutation | fires | sib |
|---|---|---|
| control (no mutation) | 0 | 0 |
| verdict discarded again (the base handler) | 7 | 0 |
| reader always answers "stopped" | 6 | 0 |
| status fallback dropped (the teleop reader's logic) | 2 | 0 |
| `stopped`-flag branch dropped (status only) | 1 | 0 |
| conservatism dropped (no envelope-shape guard) | 1 | 0 |
| reported at WARNING rather than CRITICAL | 6 | 0 |
| hardware-cutoff remedy dropped | 1 | 0 |
| source of the stop dropped | 1 | 0 |
| robot's own reason not carried | 2 | 0 |
| stop read moved ahead of the authorization check | 1 | **2** |
| reader always answers "refused" (inverse) | 5 | 0 |

The rows that carry the design: the status fallback dropped fires exactly the two **daemon-refused** cells, which is the measured argument against reusing the teleop reader; the inverse row fires 5 including three over-reach controls, so that class is load-bearing; and moving the read ahead of the authorization check is the only row that trips a pre-existing test, so the ordering is graded at runtime rather than only in an assertion. Two rows share a firing set for a structural reason — not computing the refusal and not logging it loudly both make it invisible above WARNING.

## Gate

- Paired 550-file selection (everything naming `device_connect`, `stop_task`, `emergencyStop` or the mesh e-stop reader, plus every tree-walking guard), the new file excluded from both trees and every path verified present on both: identical **25175 collected** and `22 failed, 25043 passed, 110 skipped` on each, **22 identical failing ids with `comm` empty in both directions**, distinct rootdirs. All 22 classified: 17 need `awsiot` (declared in the `[mesh-iot]` extra), 3 are lerobot-from-source drift, 2 need `device_connect_agent_tools` (declared in `[device-connect]`).
- `mypy strands_robots tests tests_integ`: **24 == 24**, normalized message sets byte-identical in both directions, **0 attributable** to the changed files.
- `ruff check` and `ruff format --check` clean over 1638 files.

No visual artifact: this is a reporting change on a teardown path, not a policy, simulation, rendering, recording or asset change. The measured log capture above is the evidence.

## Scope

`HardwareRobot.cleanup` also discards a `stop_task` envelope, and is deliberately untouched: `HardwareRobot.stop_task` has no error path at all, so there is no verdict there to read. `SimulationDeviceDriver.onEmergencyStop` clears `policy_running` directly and produces no envelope, so it is a different shape.

### DIFF
--- a/changelog.d/2831-estop-reports-a-robot-that-did-not-stop.md
+++ b/changelog.d/2831-estop-reports-a-robot-that-did-not-stop.md
@@ -1,0 +1,45 @@
+### Fixed: an emergency stop the robot refused is reported instead of discarded
+
+`RobotDeviceDriver.onEmergencyStop` called `self._robot.stop_task()` and dropped
+the envelope. Two of the four robot surfaces the driver can wrap answer a stop
+with an affirmative "I did not stop", and each was written deliberately to do
+so. `G1Driver.stop_task` returns `status="error"` with `stopped=False` when its
+control loop outlasts the join budget, and its own docstring says why: "so the
+caller cannot read 'success' while the payload's own `running=True` says the
+loop is still writing frames". `ReachyDriver.stop_task` reports a daemon that
+refused the stop. This handler was the caller that read neither field.
+
+So an operator's emergency stop was answered by a single WARNING line stating
+that the stop was being attempted - the intent, not the outcome - with nothing
+recorded above WARNING while the loop kept publishing. `Mesh.emergency_stop`
+grades exactly that verdict for every peer it fans out to, logs one that did not
+stop at CRITICAL, and names a hardware cutoff as the remedy, on the reasoning
+`_peers_that_did_not_stop` states: an emergency stop is only as trustworthy as
+its accounting. The same envelope reaching the mesh was reported and reaching
+Device Connect was not, so an operator's accounting depended on which transport
+the stop arrived over.
+
+The refusal is now read and, when the robot reports one, logged at CRITICAL
+carrying the source of the stop, the reason the robot itself gave, and the same
+cutoff remedy the mesh path names. Both spellings are read, because they differ:
+an explicit `stopped` flag is authoritative where a driver supplies one, and the
+envelope's status answers for the drivers that report through `_refuse`, whose
+envelope carries text and no flag at all. `teleop_mixin._stop_reported_stopped`
+asks the same question of a `stop_teleoperate` envelope and is deliberately not
+reused - it answers `True` for an envelope with no `json` block, which is right
+there (nothing was teleoperating) and wrong here, because that is the shape a
+refused `stop_task` arrives in.
+
+The reader is conservative for the reason `_peers_that_did_not_stop` gives: only
+an envelope that affirmatively reports a failure is flagged, because a false "did
+not stop" on the safety path trains operators to ignore the warning. A robot that
+stopped, one that had nothing to stop, one reporting an error beside a payload
+that says the loop stopped, and a driver that returns nothing at all are all left
+alone. The authorization guard still runs first, so reading the verdict does not
+widen who can halt a task.
+
+The three suites that already drive `onEmergencyStop` assert
+`stop_task.assert_called_once()` or `robot.stopped is True` - that the call was
+made, never what it answered - and their robot doubles are `MagicMock`s whose
+`stop_task` returns a `MagicMock`, which carries no verdict to read. That is why
+the discard was invisible.
--- a/strands_robots/device_connect/robot_driver.py
+++ b/strands_robots/device_connect/robot_driver.py
@@ -25,6 +25,59 @@ from strands_robots.mesh.security import is_safe_policy_provider
 logger = logging.getLogger(__name__)
 
 
+def _stop_task_refusal(envelope: object) -> str | None:
+    """Read whether a ``stop_task`` envelope reports a stop that did not happen.
+
+    Two of the robot surfaces :class:`RobotDeviceDriver` can wrap answer a stop
+    with an affirmative "I did not stop", and they spell it differently.
+    ``G1Driver.stop_task`` reports ``status="error"`` and carries ``stopped``
+    in a ``json`` block when its control loop outlasts the join budget, while
+    ``ReachyDriver.stop_task`` reports a daemon that refused the stop through
+    ``_refuse``, whose envelope carries a ``text`` block and no ``stopped`` key
+    at all. So both signals are read: an explicit ``stopped`` flag is
+    authoritative where the driver supplies one, and ``status`` answers for the
+    envelopes that do not.
+
+    ``teleop_mixin._stop_reported_stopped`` asks the same question of a
+    ``stop_teleoperate`` envelope and is deliberately not reused: it returns
+    ``True`` for an envelope with no ``json`` block, which is right there
+    (nothing was teleoperating) and wrong here, because that is the shape a
+    refused ``stop_task`` arrives in.
+
+    Deliberately conservative, for the reason
+    ``mesh.core._peers_that_did_not_stop`` gives: only an envelope that
+    AFFIRMATIVELY reports a failure is flagged. Anything else -- a driver that
+    returns nothing, a test double, a shape this function does not recognise --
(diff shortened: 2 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Closes a silent-discard on the Device Connect emergency-stop path: `RobotDeviceDriver.onEmergencyStop` previously called `self._robot.stop_task()` and dropped the envelope, so a G1 whose control loop outlasted its join budget (`status="error"`, `stopped=False`) or a Reachy whose daemon refused the stop was recorded as a single WARNING saying the stop was being attempted. The new module-private `_stop_task_refusal` reads the verdict — an explicit `stopped` flag is authoritative where present, `status` answers for text-only refusal envelopes — and a robot that affirmatively reported it did not stop is now logged at CRITICAL naming the source, the robot's own reason, and the hardware-cutoff remedy, matching the accounting `Mesh.emergency_stop` already gives the same operator request over the mesh transport.

## What's good

- The helper is deliberately conservative in the same direction as `mesh.core._peers_that_did_not_stop` — only an affirmative failure is flagged, so `None` returns, MagicMock doubles, and unrecognised shapes stay quiet; the test suite pins both directions (refusals reported, accounted stops silent).
- The authorization guard still runs before the stop, and the tests pin that a spoofed source neither halts the task nor produces a false "did not stop" record.
- The docstring explains why `teleop_mixin._stop_reported_stopped` is not reused (its no-json-block default is `True`, which is exactly the shape a refused `stop_task` arrives in), and a premise test pins that divergence so the two readers cannot silently drift together.
- The two graded envelope shapes are verified against the real drivers (`_refuse` carries text only; G1's `stop_task` reports a `stopped` flag), so the fixtures cannot rot away from the code they model.
- CRITICAL log message is plain ASCII and names an actionable remedy, per AGENTS.md string-hygiene and the mesh's own wording.

yinsong1986 (approved):
## Summary

Closes a silent discard on the Device Connect emergency-stop path: `RobotDeviceDriver.onEmergencyStop` called `self._robot.stop_task()` and dropped the envelope, so a robot that affirmatively reported "I did not stop" (G1's join-timeout `stopped=False`, Reachy's daemon refusal via `_refuse`) left nothing above WARNING while the loop kept executing. The new module-private `_stop_task_refusal` reads both spellings of the verdict - an explicit `stopped` flag where a driver supplies one, `status` for text-only refusal envelopes - and a refused stop is now logged at CRITICAL naming the source, the robot's own reason, and the hardware-cutoff remedy, matching the accounting `Mesh.emergency_stop` already gives the same operator request. The reader is conservative in the safe direction (only an affirmative failure is flagged, so clean stops, nothing-to-stop, `None` returns and MagicMock doubles stay quiet), the authorization guard still runs first, and the helper is module-private so nothing here is a one-way door. Premises verified at head: `teleop_mixin._stop_reported_stopped`, `mesh.core._peers_that_did_not_stop`, and both drivers' `_refuse`/`stop_task` shapes are as the tests state them, and the head commit differs from the previously reviewed one only by the changelog fragment.

## What's good

- Changelog fragment per AGENTS.md step 3; CRITICAL message is plain ASCII with an actionable remedy; no host paths in the test file.
- The test suite pins both directions (refusals reported, accounted stops silent), grades the two real envelope shapes against the drivers that produce them so the fixtures cannot rot, and pins the authorization ordering at runtime.
- Scope discipline: `HardwareRobot.cleanup` and `SimulationDeviceDriver.onEmergencyStop` are left alone with the reason stated (no verdict to read there).

PR #2832, opened 2026-08-27
fix(tools/pose_tool): answer for every joint a whole-arm reading was asked for

cagataycali:
## What was wrong

`MotorController.read_all_positions` skips a motor whose reply did not verify, so
it returns a subset of `motor_configs` carrying no record of what fell out. Both
tool actions that consume it guarded only the all-empty case (`if positions:`,
`if not current_positions:`), so a truthy partial reading passed straight
through.

Measured on a six-joint SO-101 with one dead servo (`wrist_flex`, id 4, every
other joint answering):

| action | before | after |
|---|---|---|
| `read_all` | `success`, 5 joints, "Current robot positions" | `error`, names `wrist_flex`, still carries the 5 |
| `store_pose` | `success`, **persisted 5 of 6 under the name** | `error`, refuses |
| `show_pose` | `success`, reports it as the pose | not found - never stored |
| `load_pose` | `success`, **"Moved to pose 'home'"** | not found - never stored |

The read layer does log `No verified reply from motor wrist_flex (id 4)`, but the
envelope said success, and `store_pose` turned a transient bus fault into a
durable artifact: a pose named `home` missing a joint, which every later
`load_pose` drives towards while answering "Moved to pose". `validate_pose`
checks bounds rather than arity, so nothing downstream could notice.

## Why this disposition

`emergency_stop` had already settled it for a partial result over the configured
motor set - it answers `status="error"` and names the joints plus what that
means:

```
EMERGENCY STOP INCOMPLETE - torque is still enabled on: <joints>.
Those joints are still driven; use the hardware power cutoff.
```

`read_all` now answers the same way and still carries the positions that did
arrive, which are what a caller diagnosing a dead servo needs.

`store_pose` refuses instead, because that one persists. `incremental_move`
refuses on an unreadable position for the same reason, and a stored pose
misrepresents itself on every later load rather than once. The mutation that
reports but persists anyway fires 5 cells, so the refusal is the behavioural
content here rather than the wording.

The gap is derived by one helper both readers share, the same way `_smooth_move`
derives its own (`sorted(set(expected) - set(got))`). Every motor is still
attempted: the reported outcome changes, the packets do not.

## Why nothing caught it

The two buses already pinned are the extremes:

| test | bus |
|---|---|
| `test_pose_tool_read_all_formats_every_motor` | every servo answers |
| `test_pose_tool_read_all_reports_failure_without_responses` | none answers |
| `test_pose_tool_store_pose_captures_current_positions` | every servo answers |
| `test_pose_tool_store_pose_failure_when_no_positions_read` | none answers |

Those are exactly the two buses on which a partial-drop guard and its absence are
indistinguishable. The mixed bus is the discriminating one, and no test drove it.

## Verification

Pre-fix split, source at the base with the new tests kept - **12 failed / 8
passed**:

|---|---|---|
| `TestThePremisesTheMixedBusRestsOn` | premise | **0 / 3** |
| `TestAnIncompleteReadingIsReportedAsOne` | regression | 5 / 5 |
| `TestAnIncompletePoseIsNotPersisted` | regression | 5 / 5 |
| `TestBothReadersShareOneAccountOfCompleteness` | structural | 2 / 2 |
| `TestAHealthyArmIsUnchanged` | control | **0 / 3** |
| `TestTheAllSilentBusKeepsItsOwnDiagnosis` | over-reach | **0 / 2** |

Mutations, all against the 20 cells plus the four pre-existing `pose_tool`
suites, `collected` stable at 20 on every row:

| mutation | fires | sib |
|---|---|---|
| control (none) | 0 | 0 |
| `read_all` guard removed | 6 | 0 |
| `store_pose` guard removed | 6 | 0 |
| both guards removed | 11 | 0 |
| helper compares the reading against itself | 10 | 0 |
| `read_all` reports but keeps `status="success"` | 1 | 0 |
| `store_pose` reports but persists anyway | 5 | 0 |
| helper returns an unsorted set | 0 (see below) | 0 |
| `read_all` drops the `unread` key | 2 | 0 |
| `store_pose` message drops the joint names | 2 | 0 |
| each site re-derives the gap inline (drift) | 1 | 0 |
| over-reach: all-silent routed to the new message | 1 | **1** |

`b1 | b2 == b3` holds, and the two halves share exactly one cell - the structural
one, which either half correctly fires - so 6 + 6 - 1 = 11 exactly.

The over-reach row is the only one that trips a pre-existing test, which is what
pins the all-silent bus keeping its own diagnosis rather than being folded into
the new message.

The unsorted-set row is reported honestly as intermittent rather than as a gap:
`list(set(names))` can coincide with `sorted(names)`, and measuring across eight
hash seeds it does so on 4 of 8 for a two-name gap. The shipped assertion is
deterministic on correct code; only the mutation's detection depends on the
process hash seed. The sort is kept because it matches `_smooth_move`'s form and
costs nothing.

Gate: `ruff check` and `ruff format --check` clean over 1640 files. mypy
**24 == 24** against a pristine worktree of the same base, normalized message
sets byte-identical in both directions, **0 attributable**. Paired run over an
identical 233-file selection with the new file excluded from both trees:
identical **10084 collected** and `10019 passed, 65 skipped, 0 failed` on each,
distinct rootdirs. Rebased onto `49eec31f` after #2826 and #2830 landed, then
re-ran my files together with the two guard files main added (358 passed).

No visual artifact: this changes what a hardware tool reports, not a policy,
simulation, render, recording or asset. The measured before/after table above is
the evidence.

## Deliberately out of scope

A pose already on disk that is incomplete - written before this change, or by
hand - still loads and drives the joints it names. `validate_pose` could check
arity, but that is a third decision about what an existing stored pose means, and
`store_pose` refusing is what stops new ones being created.

### DIFF
--- a/changelog.d/2832-whole-arm-reading-names-a-silent-joint.md
+++ b/changelog.d/2832-whole-arm-reading-names-a-silent-joint.md
@@ -1,0 +1,28 @@
+### Fixed: a whole-arm reading answers for every joint it was asked for
+
+`MotorController.read_all_positions` skips a motor whose reply did not verify, so
+it returns a subset of `motor_configs` carrying no record of what fell out. Both
+tool actions that consume it guarded only the all-empty case, so a truthy partial
+reading passed straight through. On a six-joint SO-101 with one dead servo,
+`read_all` reported five positions as "Current robot positions" and `store_pose`
+persisted those five under a name, which every later `load_pose` then drove
+towards while answering "Moved to pose". `validate_pose` checks bounds rather
+than arity, so nothing downstream could notice.
+
+`emergency_stop` had already settled the disposition for a partial result over
+the configured motor set: it answers `status="error"` and names the joints plus
+what that means, because a caller told the whole arm was released when part of it
+is still driven is worse than no stop at all. `read_all` now answers the same way
+and still carries the positions that did arrive, which are what a caller
+diagnosing a dead servo needs. `store_pose` refuses instead, because that one
+persists - `incremental_move` refuses on an unreadable position for the same
+reason, and a stored pose misrepresents itself on every load rather than once.
+
+The gap is derived by one helper shared by both readers, the same way
+`_smooth_move` derives its own: compare what came back against what was
+expected. A bus where nothing answers keeps its own diagnosis, and every motor is
+still attempted - the reported outcome changes, the packets do not.
+
+The two buses already pinned were the extremes, all servos answering and none
+answering, and those are exactly the two on which a partial-drop guard and its
+absence are indistinguishable. The mixed bus is now covered.
--- a/strands_robots/tools/pose_tool.py
+++ b/strands_robots/tools/pose_tool.py
@@ -230,6 +230,30 @@ _DEFAULT_MOTOR_CONFIGS: dict[str, MotorConfig] = {
     "gripper": {"id": 6, "range": (0, 100), "resolution": 4095},
 }
 
+
+def _joints_that_did_not_answer(controller: "MotorController", positions: dict[str, float]) -> list[str]:
+    """Name the configured joints missing from a whole-arm reading.
+
+    :meth:`MotorController.read_all_positions` skips a motor whose reply did
+    not verify, so its result is a subset of ``motor_configs`` and carries no
+    record of what fell out. The gap is derived the same way
+    :meth:`MotorController._smooth_move` derives its own: compare what came
+    back against what was expected.
+
+    One helper for both whole-arm readers, so ``read_all`` and ``store_pose``
+    cannot come to disagree about what a complete reading is.
+
+    Args:
+        controller: The arm whose ``motor_configs`` defines the expected set.
+        positions: The reading returned for that arm.
+
+    Returns:
+        Sorted names of the configured joints absent from ``positions``, empty
+        when every joint answered.
+    """
+    return sorted(set(controller.motor_configs) - set(positions))
+
+
 # The degree-valued target each action reads. An action absent from this map
 # commands no joint and is never refused here.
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes the partial-reading gap in `pose_tool`: `MotorController.read_all_positions` drops any motor whose reply did not verify, and both consumers (`read_all`, `store_pose`) only guarded the all-empty case, so a bus with one dead servo reported a 5-of-6 reading as the arm's positions and `store_pose` persisted it as a durable named pose that every later `load_pose` then drove toward. The fix adds one shared helper (`_joints_that_did_not_answer`) that derives the gap against `motor_configs`; `read_all` now reports the incomplete reading as an error while still carrying the positions that did arrive, and `store_pose` refuses outright since the artifact persists. The dispositions match the existing precedents in the same file (`emergency_stop` reports a partial set as an error; `incremental_move` refuses on an unreadable position).

## What's good

- The mixed-bus test suite covers exactly the discriminating case the two pre-existing extremes (all-answer / none-answer) could not, with premise, regression, control, and over-reach classes -- and the all-silent bus keeps its original diagnosis rather than being folded into the new message.
- The structural test (`TestBothReadersShareOneAccountOfCompleteness`) pins that both `read_all_positions` readers consult the single helper, so the two surfaces cannot drift apart on what a complete reading is.
- All user-facing strings are plain ASCII, the error paths return error dicts rather than raising, the changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention, and the healthy-arm behaviour is pinned unchanged.
- Verified locally at head `d4cf266a`: the 20 new tests pass, and the full pose_tool-related selection (281 tests) passes with no regression.

PR #2834, opened 2026-08-27
fix(hardware): abandon a rollout whose bring-up a shutdown has already cancelled

cagataycali:
## Problem

`cleanup()` sets `_shutdown_event` and only then calls `stop_task()`, gated on
`status == RUNNING`. A teardown that lands while a task is still in `CONNECTING`
— a motors-bus handshake plus per-camera warmup, seconds on a real arm and
longer on a multi-camera rig — therefore sets **no stop latch at all**, and the
shutdown event is the only record of it.

`_execute_task_async` calls `_honor_stop_request()` at each stage boundary, and
that gate read only `_stop_requested`. So both stage checks passed and the
rollout went on to finish the bring-up it had just been asked to abandon.

Three places decide whether the rollout in flight may continue. Two of them read
both latches; the bring-up gate read one:

| reader | `_stop_requested` | `_shutdown_event` |
|---|---|---|
| control loop exit condition | yes | yes |
| terminal-status discriminator | yes | yes |
| `_honor_stop_request()` (bring-up gate) | yes | **no** |

## Why this is a defect and not a cosmetic status issue

The reported status was already correct — the terminal discriminator reads both
latches, so such a rollout reports `STOPPED` with 0 steps and never commands the
arm. What ran anyway was the rest of the bring-up.

`_shutdown_error` refuses a task started *after* `cleanup()` rather than merely
relabelling it, and its docstring states the reason: *"Bring-up is not free of
side effects, which is what makes this a refusal rather than a cosmetic status
fix."* It then lists them. The in-flight producer reached the same ones.

Measured on the existing two-device double, with `cleanup()` landing while the
bring-up is parked inside `connect()`:

| | before | after |
|---|---|---|
| policy-server dial / checkpoint load | 1 | **0** |
| observation read off the bus and cameras | 1 | **0** |
| `Policy.reset()` on the caller's own object | 1 | **0** |
| `cleanup()` blocked on that work (`start_task` path) | 0.703 s | **0.402 s** |
| task status / steps | stopped / 0 | stopped / 0 |
| arm commanded | no | no |

The reset is the sharpest: a caller may drive one policy object through several
tasks — the documented `run_policy(policy_object=...)` reuse pattern — so
clearing its action-chunk cache and sampler RNG while backing out of a cancelled
rollout corrupts a rollout running elsewhere.

## Why nothing caught it

`tests/test_hardware_after_shutdown.py` already frames both producers, and its
docstring concluded *"so the terminal block has to treat `_shutdown_event` the
way it already treats the stop latch."* That is true and it fixes the report. An
entry-point guard genuinely cannot cover this producer, but the **in-flight stage
gate** can, and it was reading one latch.

Three producers ask the rollout to stop. Only one had these properties ungraded:

| producer | reached the dial / read / reset | pinned by |
|---|---|---|
| operator `stop_task()` during `CONNECTING` | no | `test_hardware_stop_during_connect.py::TestStopWhileConnecting::test_the_policy_is_never_initialized` |
| task started after `cleanup()` | no | `test_hardware_after_shutdown.py::TestRefusalHappensBeforeAnyDeviceEffect` |
| `cleanup()` during `CONNECTING` | **yes** | nothing |

## Change

`_rollout_stop_latched` becomes the single predicate all three readers consult.
Rewriting the loop condition and the terminal discriminator in its terms is a
provable no-op (De Morgan on the same two latches); the behavioural content is
the gate. A test walks the module to check the disjunction has exactly one
owner, because asserting the values agree cannot see this class of drift — two
copies of one disjunction agree right up to the moment one of them is edited.

Docstrings corrected where they now understate or overstate: the gate's claim
that it is "the only thing that has to run for a stop pressed at any point
during bring-up to be honored", `cleanup()`'s summary, and the terminal block's
comment. `docs/hardware/robot-control.md` already claimed *"a shutdown truncates
a task exactly as `stop_task()` does"* — true of the report, and now true of the
work as well.

## Tests

Pre-fix split on the changed suite: **8 failed / 17 passed** — 4 behavioural
regression cells and 4 structural, with 0 failures in the premise, control,
boundary and pre-existing classes.

Mutation table, 9 rows, control clean, **8 distinct firing sets**, `collected`
stable at 25 on every row, source restored byte-identically. `sib` counts the
seven sibling hardware suites:

| mutation | fires (suite) | sib |
|---|---|---|
| b0 control | 0 | 0 |
| b1 gate reads only the stop latch (the defect) | 4 | 0 |
| b2 shared predicate returns only the stop latch | 7 | 0 |
| b3 shared predicate returns only the shutdown latch | 1 | **7** |
| b4 shared predicate ANDs instead of ORs | 7 | 7 |
| b5 loop condition re-derives it inline | 1 | 0 |
| b6 terminal discriminator re-derives it inline | 2 | 0 |
| b7 gate moved after the policy build | 1 | 27 |
| b8 instead: `cleanup()` widens its own `stop_task` guard | 7 | 0 |

- **b3** is the scope row: dropping the *stop* latch fires six cells in
  `test_hardware_stop_during_connect.py`, so that half is load-bearing and
  pinned elsewhere.
- **b5** is behaviourally a no-op and is caught only structurally — which is
  exactly what the drift looked like before it drifted.
- **b8** rules out the alternative fix by measurement rather than preference: it
  trips the pre-existing
  `TestNoFalseTerminalStatus::test_a_shutdown_during_bring_up_reports_stopped_not_completed`,
  which asserts `_stop_requested.is_set() is False` after that teardown.
  Reading the latch at the gate also covers any setter of `_shutdown_event`, not
  just `cleanup()`'s one branch.

## Scope

Unchanged: an operator `stop_task()` during bring-up behaves exactly as before, a
healthy rollout still resets its policy once and reads its observations, and the
terminal discriminator remains the backstop for a shutdown landing after the last
stage check.

Deliberately out of scope, and pinned as such by
`test_the_devices_the_parked_bring_up_opens_are_still_left_open`: the devices the
parked `connect()` opens are still open afterwards. That call is already inside
its blocking handshake when the teardown lands, so it completes either way, and
for a rollout `cleanup()` did not submit it completes after
`_disconnect_devices()` has run. Closing them needs `cleanup()` to wait for a
rollout running on a caller's thread, which is a different fix.

## Gate

- `ruff check` + `ruff format --check` clean over `strands_robots tests tests_integ` (1641 files).
- `mypy` **24 == 24** against a worktree of the merge-base, normalized message
  sets byte-identical in both directions, **0 attributable** to the changed files.
- Paired suite over an identical 774-file selection (everything naming the
  hardware robot, the task lifecycle, the latches or teleop, plus every guard
  that walks the tree, parses source or reads docstrings and `docs/`), with the
  changed test file deselected from both trees: **36 failed, 29766 passed, 167
  skipped on each**, 29967 collected on each, **36 == 36 identical failing IDs,
  `comm` empty in both directions**. 17 need `awsiot`/`awscrt` (absent locally,
  declared in the `[mesh-iot]` extra); 4 pass 13/13 in isolation on both trees;
  the rest are the lerobot-from-source `encode()` drift and HF-hub reachability.
- Changed suite plus the seven sibling hardware suites and the four guards main
  gained since the merge-base: **308 passed**.

No visual artifact: this changes teardown control flow, not a policy, a
simulation behaviour, rendering, recording or a robot asset. The measured
side-effect table above is the evidence.

### DIFF
--- a/changelog.d/2834-a-shutdown-during-bring-up-stops-the-rollout.md
+++ b/changelog.d/2834-a-shutdown-during-bring-up-stops-the-rollout.md
@@ -1,0 +1,59 @@
+### Fixed: a `cleanup()` that lands during bring-up abandons the rollout instead of finishing it
+
+`cleanup()` sets `_shutdown_event` and only then calls `stop_task()`, gated on
+`status == RUNNING`. A teardown arriving while the task is still in `CONNECTING`
+- a motors-bus handshake plus per-camera warmup, seconds on a real arm and
+longer on a multi-camera rig - therefore sets no stop latch at all, and the
+shutdown event is its only record. `_execute_task_async` checks
+`_honor_stop_request()` at each stage boundary and that gate read only the stop
+latch, so both checks passed and the rollout went on to finish the bring-up it
+had just been asked to abandon.
+
+The reported status was already right. `_shutdown_error` refuses a task started
+*after* `cleanup()` rather than merely relabelling it, and its docstring says
+why: "Bring-up is not free of side effects, which is what makes this a refusal
+rather than a cosmetic status fix." It then lists them. The in-flight producer
+reached the same ones, and the terminal-status discriminator - which does read
+both latches - could only correct what the rollout said about itself, not what
+it did. Measured on the two-device double, with `cleanup()` landing while the
+bring-up is parked inside `connect()`:
+
+```
+                                              before      after
+policy-server dial / checkpoint load          1           0
+observation read off the bus and cameras      1           0
+Policy.reset() on the caller's own object     1           0
+cleanup() blocked on that work (start_task)   0.703s      0.402s
+task status / steps                           stopped / 0 stopped / 0
+arm commanded                                 no          no
+```
+
+The reset is the sharpest of them: a caller may drive one policy object through
+several tasks - the documented `run_policy(policy_object=...)` reuse pattern -
+so clearing its action-chunk cache and sampler RNG while backing out of a
+cancelled rollout corrupts a rollout running elsewhere. The sibling producer's
+`reset_calls == 0` was already pinned; this one's was not.
+
+Three readers decide whether the rollout in flight may continue, and they must
+agree: the bring-up gate, the control loop's exit condition, and the terminal
+discriminator. The last two read both latches and the first read one. Asserting
+their values agree cannot see that - two copies of the same disjunction agree up
+to the moment one is edited, which is how the drift arose - so all three now
+read a single `_rollout_stop_latched` predicate, and a test walks the module to
+check the disjunction has exactly one owner.
+
+Widening `cleanup()`'s own `stop_task()` guard to cover `CONNECTING` would also
+stop the side effects, and is ruled out rather than merely disfavoured: it trips
+`test_a_shutdown_during_bring_up_reports_stopped_not_completed`, which asserts
+`_stop_requested.is_set() is False` after that teardown. Reading the latch at
+the gate also covers any setter of `_shutdown_event`, not just that one branch.
+
+Unchanged: an operator `stop_task()` during bring-up behaves exactly as before,
+a healthy rollout still resets its policy once and reads its observations, and
+the terminal discriminator remains the backstop for a shutdown landing after the
+last stage check. Deliberately out of scope and pinned as such: the devices the
+parked `connect()` opens are still open afterwards. That call is already inside
+its blocking handshake when the teardown lands, so it completes either way, and
+for a rollout `cleanup()` did not submit it completes after `_disconnect_devices()`
+has run. Closing them needs `cleanup()` to wait for a rollout running on a
+caller's thread, which is a different fix.
--- a/docs/hardware/robot-control.md
+++ b/docs/hardware/robot-control.md
@@ -50,7 +50,7 @@ robot.cleanup()
 | `start_task(instruction, policy_port, policy_host, policy_provider, duration)` | Async; returns immediately. |
 | `stop_task()` | Halt the current task. Covers a task still in `CONNECTING` (bring-up): the rollout is abandoned before the arm is commanded. |
 | `get_task_status()` | Returns `RobotTaskState` (status, step count, error). |
-| `cleanup()` | Stop tasks, disconnect the robot (motors bus + every camera), stop mesh. Terminal - see below. |
+| `cleanup()` | Stop tasks, disconnect the robot (motors bus + every camera), stop mesh. Abandons a task still in `CONNECTING` at its next stage check, like `stop_task()`. Terminal - see below. |
 | `stop()` | Async spelling of `cleanup()`; delegates to it off the event loop. Terminal. |
 
 One rollout at a time: the arm has a single command bus, so `start_task` /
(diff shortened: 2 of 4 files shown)


yinsong1986 (dismissed):
## Summary

This PR closes the remaining half of the shutdown-during-bring-up gap: a `cleanup()` landing while a task is still in `CONNECTING` sets only `_shutdown_event` (its `stop_task()` call is gated on `status == RUNNING`), and the bring-up stage gate `_honor_stop_request` read only `_stop_requested` -- so the rollout finished its bring-up (policy-server dial, observation read, `Policy.reset()` on a possibly-shared policy object) after the teardown was latched, even though the terminal status was already reported correctly. The fix introduces a single private predicate `_rollout_stop_latched` OR-ing the two latches and routes all three readers (bring-up gate, loop exit condition, terminal-status discriminator) through it, so the drift that produced this bug is structurally closed.

## What's good

- The predicate is private and behaviour-preserving for the loop condition and terminal block (they already read both latches); only the bring-up gate is deliberately widened, and the widened case is exactly the one the PR describes.
- Tests pin both producers (`started-after-cleanup` and `cleanup-during-bring-up`) side by side, drive the race deterministically via a gated `connect()` double rather than sleeps, and add an AST check that the two-latch disjunction has exactly one owner -- the drift this bug came from cannot silently return.
- The known boundary that is *not* moved (devices opened by a parked `connect()` are left open) is stated and pinned rather than assumed, per the repo's no-silent-defaults conventions.
- Docs (`robot-control.md`, `cleanup()` docstring) updated in the same change; new strings are plain ASCII.

yinsong1986 (approved):
## Summary

This PR closes the in-flight half of the shutdown-during-bring-up gap: `cleanup()` sets `_shutdown_event` and only then calls `stop_task()` gated on `status == RUNNING`, so a teardown landing while a task is still in `CONNECTING` set no stop latch, and the bring-up stage gate `_honor_stop_request` read only `_stop_requested` -- the rollout therefore finished its bring-up (policy-server dial, observation read, `Policy.reset()` on a possibly-shared policy object) after the teardown was latched, even though the terminal status was already reported correctly. The fix routes all three readers (bring-up gate, loop exit condition, terminal-status discriminator) through one private `_rollout_stop_latched` predicate, so the only behavioural change is the gate now honouring the shutdown latch, and the drift that produced the bug is structurally closed by an AST test asserting the two-latch disjunction has exactly one owner.

## What's good

- The loop condition and terminal discriminator rewrites are provable no-ops (they already read both latches); the behavioural content is confined to the gate, and the widened case is exactly the one described.
- New tests drive the race deterministically via a gated `connect()` double rather than sleeps, pin both producers side by side, assert the premise (`_stop_requested` unset, `_shutdown_event` set) inside the helper, and pin the deliberately-unmoved boundary (devices the parked `connect()` opens stay open) rather than leaving it assumed.
- Verified locally at head `cc99114e`: the changed suite (25) plus `test_hardware_stop_during_connect.py` and five other sibling hardware suites (165 total) all pass, and `ruff check` is clean on the changed files.
- Conventions followed: changelog fragment (not `CHANGELOG.md`), docs updated in the same PR, `time.monotonic()` for the test's wait budget, `monkeypatch.setattr` for the stand-in, plain-ASCII strings, no host paths.

Issue #2833, opened 2026-08-27
simulation: a cooperative stop issued before a rollout's first frame is discarded, and every stop path reports the halt


cagataycali:
## Summary

A cooperative stop issued after `start_policy` returns but before the rollout's
first frame is discarded, and every stop path in the tree reports that the
rollout halted. The rollout then runs to its full duration.

The window is not microseconds. Measured on an idle executor, `start_policy`
returns in **1.14 ms** and the rollout raises `policy_running` **749.77 ms**
later, so the window is **748.63 ms** wide. An operator who stops a rollout
within three quarters of a second of starting it is told it stopped, and it
does not.

## Reproduction

```python
"""A cooperative stop issued before a rollout's first frame is discarded."""

import time

from strands_robots.simulation import create_simulation
from strands_robots.simulation.model_registry import resolve_model_path

sim = create_simulation("mujoco")
sim.create_world()
sim.add_robot(name="arm", urdf_path=str(resolve_model_path("so101")))
robot = sim._world.robots["arm"]

sim.start_policy("arm", policy_provider="mock", duration=2.0, control_frequency=50.0)
time.sleep(0.10)  # well inside the window; measured at 748 ms on an idle executor

print("list_policies_running:", sim.list_policies_running()["content"][0]["text"])
print("stop_policy          :", sim.stop_policy("arm")["content"][0]["text"])

sim._policy_threads["arm"].result(timeout=20)
print("steps executed after the stop:", robot.policy_steps)
sim.cleanup()
```

Output on `8c196945`, `MUJOCO_GL=egl`:

```
list_policies_running: Active policies (1):
  - arm
stop_policy          : Was not running on 'arm'
steps executed after the stop: 100
```

`duration=2.0` at `control_frequency=50.0` is 100 steps, so the rollout
executed in full.

## Two readers of one question, disagreeing at the same instant

Both answer `status="success"`:

| reader | answer |
|---|---|
| `list_policies_running()` | `Active policies (1): - arm` |
| `_active_policy_robots()` | `['arm']` |
| `stop_policy("arm")` | `Was not running on 'arm'` |
| `list_policies_running()` after that stop | `Active policies (1): - arm` |

One surface says a rollout is in flight; its sibling says there was nothing to
stop. The information needed to tell the truth is in the same object.

## Mechanism

`start_policy` registers the future and returns:

* `simulation/mujoco/simulation.py:5143` submits `run_policy` to the executor
[...]

PR #2835, opened 2026-08-27
fix(device-connect): halt teleoperation as well as policies on a simulation emergency stop

cagataycali:
## The defect

`SimulationDeviceDriver.onEmergencyStop` wrote `robot.policy_running = False`
over `world.robots.values()` and stopped there. A `Simulation` mixes in
`TeleopMixin`, so a leader arm can be driving it from a thread that flag says
nothing about: the teleop loop polls `get_action()` and applies the result
through `send_action` on its own thread.

Measured on a real MuJoCo simulation with a leader attached, driving at 50 Hz:

| | leader polls | teleop session |
|---|---|---|
| before the emergency stop | 20 | `running=True` |
| after it returned (base) | 46 (+26) | `running=True, thread_alive=True` |
| after it returned (this branch) | 21 (+1, the in-flight tick) | `running=False` |

So an operator's emergency stop on a teleoperated simulation reported the halt
and left the leader driving the follower.

## Why this is the handler's gap and not a new policy

The simulation's own `cleanup` already stops teleoperation, under the very guard
this handler now uses:

```python
if getattr(self, "_teleop_running", False) or getattr(self, "_teleops", None):
    with contextlib.suppress(Exception):
        self.stop_teleoperate()
```

It does that because teleoperation is a motion source it cannot leave running.
An emergency stop cannot leave it out.

Both sibling handlers already hold an emergency stop to every source they own.
`reachy_mini_driver.onEmergencyStop` attempts torque-off AND stop-motion, records
each failure, and logs at CRITICAL naming what did not complete.
`robot_driver.onEmergencyStop` reads the stop verdict rather than discarding it
(#2831). This brings the third handler to the same accounting.

## The change

* Teleoperation is stopped as well as policies, under the guard `cleanup` uses.
* Policies go first. That is a flag write which cannot block, so the unblockable
  kill lands even when the bounded teleop join outlasts its budget -- the reason
  the reachy handler runs torque-off before its REST stop.
* Both stops are attempted even if one raises, and a source that did not stop is
  logged at CRITICAL naming it and quoting the reason the stop gave.
* The verdict is read through `teleop_mixin._stop_reported_stopped` rather than
  from the envelope's `status`. `_teleop_stats` derives that status from the
  session counters, so a session whose every frame errored answers `"error"`
  after a perfectly clean join, and a status-keyed read would raise a CRITICAL
  for a loop that did stop. This is the right reader here, where the envelope
  really is a `stop_teleoperate` one; `robot_driver` deliberately does not reuse
  it because it grades a `stop_task` envelope.
* The receipt moves from `print()` to `logger.warning`. A safety receipt written
  to stdout carries no level to alert on, and both siblings log theirs.

Scope: the ordinary `stop` RPC is left alone. Its docstring says "Stop all
running policies", which is what it does; an emergency stop is the one that
promises everything.

## Why the existing suites were silent

Three files drive this handler. They assert the POLICY flag went down
(`robot.policy_running is False`) or that a rogue caller was refused. None of
them attaches a teleoperator, so the second motion source is outside every
fixture: `grep -c teleop` over them is 0.

## Tests

21 cases. Pre-fix split: **10 failed / 11 passed**, with 7 of 7 in the regression
class and **0 of 5 controls, 0 of 4 premises**. The two cells that hold either way
are labelled as such in their class docstrings -- one is the over-reach guard
(adding the teleop stop must not cost the policy stop), one the drift guard (a
future edit that calls the stop and drops its answer).

The premise cells read the tree rather than describing it: that `MuJoCoSimEngine`
really is a `TeleopMixin` subclass, that `cleanup` really stops teleoperation
under this guard, that the shared reader grades `stopped` and not `status`, and
that the reachy handler really escalates a failed stop.

The simulation stand-in answers the envelope shape `TeleopMixin` really produces
-- a `json` block carrying `stopped` -- rather than being a `MagicMock`. A
`MagicMock` answers every call with a truthy `MagicMock` and carries no verdict to
read, which is why the pre-existing fixtures could not see this.

### Mutation table

11 mutations, 10 distinct firing sets, control clean, `collected` stable at 21 on
every row, source restored byte-identical. `sib` is the 5 pre-existing
device-connect suites (203 cases); it reads 3 on every row INCLUDING the control,
and those 3 are `ModuleNotFoundError: No module named 'device_connect_agent_tools'`
on pristine `main` -- the `[device-connect]` extra, absent locally and installed
in CI. So no mutation disturbs a sibling.

| mutation | fires |
|---|---|
| control (no mutation) | 0 |
| teleop stop block removed (the defect) | 10 |
| raw revert to the pre-fix source | 11 |
| verdict discarded (bare statement) | 4 |
| failures logged at WARNING not CRITICAL | 4 |
| a raising teleop stop escapes the handler | 2 |
| guard dropped (always call the teleop stop) | 1 |
| keyed on `status` instead of the `stopped` flag | 1 |
| teleop stop moved ahead of the policy sweep | 1 |
| policy sweep no longer guarded | 1 |
| receipt reverted to stdout | 1 |
| failure string drops the reason the stop gave | 1 |

The `status`-keyed row is worth a note: it fired 0 until a fixture was added where
the two readings DISAGREE -- a clean join whose session counters make the status
`"error"`. Without that case the fixture set could not tell the two readers apart,
so the choice of reader was untested.

## Gate

* `ruff check` + `ruff format --check` over `strands_robots tests tests_integ`:
  clean, 1643 files.
* `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine
  worktree of the merge base, normalized message sets identical in both
  directions, **0 attributable** to the changed files.
* Paired run over an identical 441-file selection (every device-connect suite plus
  every guard that walks the tree, parses source, or reads docstrings and
  `changelog.d`), the new file excluded from both trees: identical **21028
  collected** and `22 failed, 20926 passed, 80 skipped` on each, **22 identical
  failing ids with `comm` empty in both directions**. The 22 are environmental and
  present on the base: 17 need `awsiot` (`[mesh-iot]`), 2 need
  `device_connect_agent_tools` (`[device-connect]`), 3 are lerobot-from-source
  `encode()` drift.

No visual artifact: this changes a teardown path's reach and its reporting, not a
policy, a simulation behaviour, rendering, a recording, or an asset. The measured
leader-poll table above is the evidence.

## Related

While measuring this I found that a cooperative stop issued before a rollout's
first frame is discarded by every stop path, and filed it as #2833. That is a
separate defect with a separate cause (a flag raised on the worker rather than the
caller) and its fix is a cross-backend contract decision. This PR does not depend
on it, and does not close it: the teleop loop raises its own flag synchronously,
so `stop_teleoperate` genuinely stops it -- measured, frames 46 then 46.

### DIFF
--- a/changelog.d/2835-estop-halts-every-simulation-motion-source.md
+++ b/changelog.d/2835-estop-halts-every-simulation-motion-source.md
@@ -1,0 +1,26 @@
+### Fixed: an emergency stop on a simulation halts teleoperation as well as policies
+
+`SimulationDeviceDriver.onEmergencyStop` wrote `robot.policy_running = False`
+over every robot in the world and stopped there. A `Simulation` mixes in
+`TeleopMixin`, so a leader arm can be driving it from a thread that flag says
+nothing about: the teleop loop polls `get_action()` and applies the result
+through `send_action` on its own thread. On a teleoperated simulation the leader
+was polled 26 more times after the emergency stop returned, with the teleop
+session still marked running and the handler reporting the halt.
+
+The simulation's own `cleanup` already stops teleoperation under the same guard
+this handler now uses, because it is a motion source that cannot be left
+running. Both sibling handlers hold an emergency stop to every source they own:
+`reachy_mini_driver.onEmergencyStop` attempts torque-off and stop-motion and
+logs a failure at CRITICAL, and `robot_driver.onEmergencyStop` reads the stop
+verdict rather than discarding it.
+
+Policies are stopped first, since that is a flag write which cannot block, so
+the unblockable kill lands even when the bounded teleop join outlasts its
+budget. Both stops are attempted even if one raises, and a source that did not
+stop is logged at CRITICAL naming it and quoting the reason the stop gave. The
+verdict is read through `teleop_mixin._stop_reported_stopped` rather than from
+the envelope's status, because `_teleop_stats` derives that status from the
+session counters and a session whose frames errored answers `"error"` after a
+perfectly clean join. The receipt moves from `print()` to a log record: a safety
+receipt written to stdout carries no level to alert on.
--- a/strands_robots/device_connect/sim_driver.py
+++ b/strands_robots/device_connect/sim_driver.py
@@ -19,6 +19,7 @@ from device_connect_edge.types import DeviceIdentity, DeviceStatus
 
 from strands_robots.device_connect._authz import authz_error, is_authorized_caller
 from strands_robots.mesh.security import is_safe_policy_provider
+from strands_robots.teleop_mixin import _stop_reported_stopped
 
 logger = logging.getLogger(__name__)
 
@@ -204,6 +205,20 @@ class SimulationDeviceDriver(DeviceDriver):
     async def onEmergencyStop(self, device_id: str, event_name: str, payload: dict[str, Any]) -> None:
         """React to emergencyStop from an authorized safety controller.
 
+        Halts EVERY motion source the simulation has, not only its policies. A
+        ``Simulation`` mixes in :class:`~strands_robots.teleop_mixin.TeleopMixin`,
+        so a leader arm can be driving it from a thread the ``policy_running``
+        flag says nothing about; stopping policies alone left that loop polling
+        ``get_action()`` and applying the result through ``send_action`` after an
+        operator's emergency stop, with the handler reporting the halt.
+
+        Both stops are attempted even if one fails, and a source that did not
+        stop is logged at CRITICAL naming it -- the accounting
+        ``reachy_mini_driver.onEmergencyStop`` gives its two stop actions, and
+        that ``robot_driver.onEmergencyStop`` gives the stop verdict it reads.
+        A stop that arrives here rather than over the mesh is the same operator
+        request, so it gets the same accounting.
+
         Security hardening: only act on emergency-stop events whose source is
         in the emergency-stop allowlist, so a spoofed event from an arbitrary
         device cannot interrupt operations.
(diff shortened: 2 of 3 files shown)


yinsong1986 (dismissed):
## Summary

Widens `SimulationDeviceDriver.onEmergencyStop` from a policy-flag sweep to an attempt-every-source halt: policies first (an unblockable flag write), then `stop_teleoperate()` when the simulation's `TeleopMixin` state says a teleop session exists, with each source guarded independently, the teleop stop's verdict read via the shared `_stop_reported_stopped` reader (keyed on the `stopped` flag rather than the counter-derived `status`), and any incomplete halt escalated at CRITICAL naming the source. This mirrors `reachy_mini_driver.onEmergencyStop`'s two-action shape and the guard `MuJoCoSimEngine.cleanup` already uses for the same reason, and it replaces the old `print` receipt with a leveled log record.

## What's good

- The new suite fails 11/21 on the pre-fix handler and passes 21/21 on the fix, so the regression is genuinely pinned rather than asserted (AGENTS.md > Testing > "Pin every reviewed fix with a regression test").
- Controls pin the over-reach directions: an errored-frames-but-clean-join session raises no false CRITICAL, an unauthorized source still stops nothing, and a no-teleop simulation is not asked to stop one.
- The broad `except Exception` handlers carry the same noqa + attempt-every-source justification as the reachy precedent, and the ordering (flag write before bounded join) is pinned structurally.
- `ruff check` / `ruff format --check` clean; test-name, host-path, and ASCII hygiene meta-tests pass on the branch.

## Verification suggestions

- CI note: no `changelog.d/2835-*.md` fragment is in the diff, so `.github/workflows/changelog-fragment.yml` will flag it; adding the fragment is self-clearing and not a code concern.

yinsong1986 (approved):
## Summary

Widens `SimulationDeviceDriver.onEmergencyStop` from a policy-flag sweep to an attempt-every-source halt: the unblockable `policy_running` flag write lands first, then `stop_teleoperate()` runs under the same `_teleop_running` / `_teleops` guard `MuJoCoSimEngine.cleanup` already uses, each source is guarded independently so one failure cannot skip the other, the teleop verdict is read through the shared `_stop_reported_stopped` reader (keyed on the `stopped` flag rather than the counter-derived `status`, so an errored-frames-but-clean-join session raises no false CRITICAL), and an incomplete halt is escalated at CRITICAL naming the source and quoting the stop's own reason. The stdout `print` receipt becomes a leveled log record. This brings the third device-connect handler to the same accounting as `reachy_mini_driver.onEmergencyStop` (attempt-both, CRITICAL on failure) and `robot_driver.onEmergencyStop` (verdict read, not discarded), and I verified those precedents, the reader's semantics, and the `cleanup` guard against the head tree. No public API, wire format, or persisted schema changes; the authz gate stays the first check and an unauthorized source still stops nothing.

## What's good

- The regression class, over-reach guards (a failing teleop stop does not cost the policy sweep and vice versa), drift guards (verdict may not be discarded), and premise cells (the mixin relationship, the `cleanup` guard, the reader's `stopped`-not-`status` grading) are all pinned, with a fixture that answers the real envelope shape instead of a `MagicMock` (AGENTS.md > Testing > "Pin every reviewed fix with a regression test").
- The broad `except Exception` handlers carry the noqa + attempt-every-source justification the reachy precedent established, and a raise inside the verdict read fails in the safe direction (a false CRITICAL, never a missed halt).
- The `changelog.d/2835-*.md` fragment is in the diff, so the fragment gate is satisfied.
- The synchronous bounded join (3.0s) inside the async handler matches `robot_driver.onEmergencyStop`'s sync `stop_task()` call, and ordering the flag write first is pinned structurally so a wedged leader cannot delay the cheap kill.

## Verification suggestions

- The new suite needs the `[device-connect]` extra (`device_connect_edge`); without it all 21 cases error at fixture setup exactly as the pre-existing `test_device_connect_hardening.py` does, so a plain-install run of just these files is not a signal either way. CI's `call-test-lint` (in progress at review time) is the authoritative run.

yinsong1986 (approved):
## Summary

Widens `SimulationDeviceDriver.onEmergencyStop` from a policy-flag-only sweep to an attempt-every-source halt. The unblockable `policy_running = False` write lands first, then `stop_teleoperate()` runs under the same `_teleop_running` / `_teleops` guard `MuJoCoSimEngine.cleanup` already uses (verified in-tree), each source guarded independently so one failure cannot skip the other. The teleop verdict is read through the shared `_stop_reported_stopped` reader keyed on the envelope's `stopped` flag rather than its counter-derived `status`, so a clean join with unhealthy frame counters does not raise a false CRITICAL, and a source that did not stop is escalated at CRITICAL naming it and quoting the stop's own reason -- the same accounting both sibling handlers (`reachy_mini_driver`, `robot_driver`) already give. The authz gate (`scope="estop"`) is unchanged ahead of any action, the two broad catches are documented safety-handler recovery paths matching the reachy precedent, the receipt moves from stdout to a leveled log record, and the ordinary `stop` RPC is deliberately left alone. The 364-line suite covers the regression, over-reach and drift guards, and reads its premises (mixin subclassing, the `cleanup` guard, the reader's `stopped`-vs-`status` distinction, the sibling's CRITICAL escalation) from the tree rather than restating them. Changelog fragment present under the `<pr>-<slug>.md` convention. No blockers found.

PR #2836, opened 2026-08-27
fix(isaac): check a visual mesh geom's placement in the reader that reports it

cagataycali:
## The defect

`load_mjcf_scene_objects` reports two placements per object. The collision proxy -
`position` and `size` - comes from `_recursive_collision_aabb`. Where the *visual
asset* hangs on that proxy comes from `_find_body_mesh`, which returns the
`mesh_pos` and `mesh_quat` a `SceneObject` carries, and read both straight through
`_parse_xyz` / `_parse_orientation`.

`_geom_aabb` refuses a non-finite geom `pos`, `size`, orientation or `fromto` at the
parse. It never saw the geom this reader picks. Driving one MJCF fixture through the
public loader, four body shapes x two mesh positions:

| body shape | mesh `pos` | before |
|---|---|---|
| contact-free mesh + collidable box | `0 0 0.02` | accepted, `mesh_pos=(0.0, 0.0, 0.02)` |
| **contact-free mesh + collidable box** | **`0 0 nan`** | **accepted, `mesh_pos=(0.0, 0.0, nan)`** |
| collidable mesh + collidable box | `0 0 nan` | refused |
| contact-free mesh, no other geom | `0 0 nan` | refused |
| collidable mesh, no other geom | `0 0 nan` | refused |

One row of four, and it is the dominant convention. Two independent reasons keep the
existing guard away from it:

- `_body_collision_aabb` walks `for collidable_only in (True, False)` and **returns
  as soon as a pass finds a bound**, so on a body that owns a collision primitive the
  second pass never runs and a contact-free geom is never handed to `_geom_aabb` at
  all. Instrumented, the calls for the leaking fixture are `['hull']`.
- A `type="mesh"` geom has no analytic AABB, so even when it *is* handed over the
  answer is `None`.

`_find_body_mesh` prefers precisely the geom the first reason excludes: its strongest
rank is `_MESH_VISUAL_RANK_NON_COLLIDING`, because MuJoCo Menagerie marks a visual
geom `contype="0" conaffinity="0"` and declares the collision geom first. So the one
geom whose placement was *reported* was the one geom whose placement was *unchecked*.

`position` and `size` stayed finite and correct in the leaking row, because the
accumulator supplies them from the collision primitive. The object's physics read
healthy while its visual asset was hung at a coordinate that is not a coordinate -
the wrong half to lose for a reader whose own docstring gives the reason it prefers
the visual asset: "the visual asset is the one a pixel-conditioned policy was trained
on".

A single bad component was not contained either. `_parse_orientation` normalizes, so
`quat="1 0 nan 0"` reported `mesh_quat=(nan, nan, nan, nan)`.

## Reachability

MuJoCo refuses a non-finite geom on a body with a free joint (`mass and inertia of
moving bodies must be larger than mjMINVAL`) and **compiles** the same geom on a body
without one - and a body without a free joint is what this loader reports as a
fixture. So the scenes that reach the reader are precisely the ones that load, the
same boundary `test_non_finite_geom_is_refused_not_measured_around.py` already
establishes. Both premises are pinned.

Across the 570 MJCF files cached on the measuring machine: 3054 bodies, 2855 with a
mesh geom, **2053 in the leaking shape** - `so101_new_calib.xml` contributes
`shoulder`, `upper_arm`, `lower_arm`, `wrist` and `gripper`, i.e. every link of an arm
the registry ships. **None carries a non-finite value today**, so the defect is
latent, and **60 of 60** resolvable registry assets load unchanged after the fix.

## The change

Both quantities go through `_refuse_non_finite_geom` - the module's existing owner for
this wording, so no spelling drifts - at the parse rather than at the return. The
placement is deliberate: `_geom_cannot_collide` is true for exactly the geoms
`_mesh_geom_visual_rank` ranks strongest, and the reader returns on the first of
those, so the geom the bound cannot vouch for is always the one the reader picks and
there is no "checked when it wins" placement covering a different set. Pinned.

Deliberately unchanged, and pinned: an attribute that cannot be *parsed* still falls
back to its documented default; finite values are untouched however extreme; and the
nested-body `pos` this reader folds into its offset is **not** re-guarded here,
because `_recursive_collision_aabb` refuses it on the same `findall("body")`
traversal.

### A shipped test class asserted the coverage that was missing

`TestTheAccumulatorsGuardCoversTheMeshWalkToo` reasoned that "every body the mesh walk
reads is a body the guarded accumulator also reads - a second guard there would be
unreachable", and pinned the absence with
`test_the_mesh_walk_carries_no_finiteness_test_of_its_own`. The reason is true and
reaches the *body* offsets; the exclusion it licensed was wider, because the same
function also reads each mesh geom's own `pos` and orientation, which the accumulator
never parses. The class is renamed to what it establishes and split into the division
of labour it supports - one cell for the offset it must not re-guard, one for the
placement only this reader sees. That is a strengthening: 2 cells become 3.

## Verification

Pre-fix split, per class (behavioural revert of the two guards):

| class | failed/total | role |
|---|---|---|
| `TestANonFiniteVisualMeshPlacementIsRefused` | 12/12 | regression |
| `TestOneOwnerForTheWording` | 2/3 | 2 regression, 1 over-reach guard |
| `TestTheAccumulatorsGuardCoversTheMeshWalksBodyOffsets` | 1/3 | the new geom-guard cell |
| `TestThePremisesTheFindingRestsOn` | 0/23 | premises, clean |
| `TestWhatIsDeliberatelyUnchanged` | 0/13 | controls, clean |
| `TestTheInputsTheGeomGuardAlreadyReached` | 0/4 | controls, clean |
| **total** | **15/73** | |

Mutation table. `mine` = the two test files; `sib` = the four pre-existing suites for
this reader, its ranking and the bound (`collected` stable at 73 on every row, so no
row hides a deselection; source restored byte-identical):

| mutation | mine | sib |
|---|---|---|
| b0 control | 0 | 0 |
| b1 `pos` guard removed | 8 | 0 |
| b2 orientation guard removed | 5 | 0 |
| b3 both removed (raw revert) | 15 | 0 |
| b4 `pos` guard placed *after* the offset composition | 1 | 0 |
| b5 guard reads `geom.attrib` rather than the resolved `attrs` | 2 | 0 |
| b6 orientation spelling hardcoded to `quat` | 4 | 0 |
| b7 reader raises its own wording instead of the shared owner | 12 | **1** |
| b8 guard moved to the winning-geom return | 1 | 0 |
| b9 nested-body `pos` re-guarded here (over-reach) | 1 | 0 |
| b10 shared owner `all()` -> `any()` (over-reach) | 23 | **27** |

11 mutations, 9 distinct firing sets, control clean. Complementarity is exact:
`b1` and `b2` are **disjoint** (8 and 5, intersection empty), `b1 | b2` is a subset of
`b3`, and the 2 cells needing both gone are the two structural ones (either guard
present satisfies "the reader delegates"). b7 and b10 each trip a pre-existing suite,
which is what pins the shared wording and the guard's own polarity.

b8 fires only the ordering cell, and the reason is structural rather than a gap: the
contact-free set *is* the strongest-ranked set, so a geom that is both invisible to
the bound and a ranking loser cannot exist. The cell that looks like it should catch
b8 - a collidable loser beside a good winner - passes before and after, because the
loser *is* offered to `_geom_aabb`; it is filed with the controls and says so.

  (1639 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine worktree of
  this PR's base, normalized message sets byte-identical in both directions, **0
  attributable** to the three changed files (`checked 1639` vs `1638`).
- `tests/simulation/isaac/` **1840 passed, 5 skipped**.
- Paired gate on an identical 600-file selection (all of `tests/simulation/isaac/`
  plus every suite that walks the tree, parses source, reads docstrings or names this
  reader), the one branch-only file excluded from both: **20 failed on each tree**,
  failing-ID sets identical with `comm` empty in both directions, distinct rootdirs,
  `27191` vs `27190` collected (`+1` = the cell added to the sibling file). All 20 are
  environmental and measured as such: 17 need `awsiot`/`awscrt` (`find_spec` false for
  both, declared in the `[mesh-iot]` extra), 3 are lerobot-0.6.2-from-source drift.

No visual artifact: this is a validation change on the Isaac loader path, adding a
refusal rather than altering a rendered result, and Isaac Sim is not installed on the
machine that measured it - so the reported-fields tables above are the evidence rather
than a render of a consumer that cannot run here.

### DIFF
--- a/changelog.d/2836-visual-mesh-placement-is-checked-by-its-reader.md
+++ b/changelog.d/2836-visual-mesh-placement-is-checked-by-its-reader.md
@@ -1,0 +1,39 @@
+### Fixed: an Isaac scene object no longer reports a visual mesh placement that is not a placement
+
+`load_mjcf_scene_objects` reports two placements per object. The collision proxy -
+`position` and `size` - comes from `_recursive_collision_aabb`. Where the visual
+asset hangs on that proxy comes from `_find_body_mesh`, which returns the
+`mesh_pos` and `mesh_quat` a `SceneObject` carries, and read both straight through
+`_parse_xyz` / `_parse_orientation`.
+
+`_geom_aabb` refuses a non-finite geom `pos`, `size`, orientation or `fromto` at
+the parse, and never saw the geom this reader picks. `_body_collision_aabb` walks
+`for collidable_only in (True, False)` and returns as soon as a pass finds a
+bound, so on a body that owns a collision primitive the second pass never runs and
+a contact-free geom is never handed over; a `type="mesh"` geom has no analytic AABB
+in any case. `_find_body_mesh` prefers precisely that geom, because MuJoCo
+Menagerie marks a visual geom `contype="0" conaffinity="0"` and declares the
+collision geom first. The one geom whose placement was reported was the one geom
+whose placement was unchecked.
+
+One of four body shapes leaked, and it is the Menagerie one - a contact-free mesh
+geom beside a collidable sibling reported `mesh_pos=(0.0, 0.0, nan)` under a
+successful load, while the other three were already refused. `position` and `size`
+stayed finite, because the accumulator supplies them from the collision primitive,
+so the object's physics read healthy while its visual asset was hung at a
+coordinate that is not a coordinate - the wrong half to lose for a reader that
+exists to prefer the asset a pixel-conditioned policy was trained on. A single bad
+component was not contained either: `_parse_orientation` normalizes, so
+`quat="1 0 nan 0"` reported `mesh_quat=(nan, nan, nan, nan)`.
+
+Both quantities now go through `_refuse_non_finite_geom`, the module's existing
+owner for this wording, at the parse rather than at the return - the contact-free
+set is exactly the strongest-ranked set, so the geom the bound cannot vouch for is
+always the one the reader picks. An attribute that cannot be parsed still falls
+back to its documented default, finite values are untouched however extreme, and
+the nested-body `pos` this reader folds into its offset is left to
+`_recursive_collision_aabb`, which refuses it on the same `findall("body")`
+traversal. Of the 570 MJCF files cached on the measuring machine, 2053 of 3054
+bodies are in the leaking shape - every link of the `so101` the registry ships
+included - and none carries a non-finite value today, so the defect was latent;
+60 of 60 resolvable registry assets load unchanged.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -1670,6 +1670,21 @@ def _find_body_mesh(
     that decide the rank are read through :func:`_class_attrs`, so a geom that
     inherits them from a ``<default class="visual">`` is ranked as MuJoCo reads
     it rather than being taken for a collision geom.
+
+    The geom's own ``pos`` and orientation are refused when they parse to a
+    non-finite value (:func:`_refuse_non_finite_geom`), because this reader is
+    the only one that sees them. :func:`_geom_aabb` refuses the same quantities,
+    but :func:`_body_collision_aabb` calls it for the geoms MuJoCo can collide
+    and returns as soon as it has a bound - so a *visual* mesh geom beside a
+    collision primitive, which is the Menagerie convention this ranking exists
+    to prefer, is never handed to it. The nested-body ``pos`` folded into the
+    offset needs no test here: :func:`_recursive_collision_aabb` refuses it on
+    the same ``findall("body")`` traversal, so a second one would be
+    unreachable.
+
+    Raises:
+        ValueError: If a mesh geom's ``pos`` or orientation resolves to a value
+            that is not finite (:func:`_refuse_non_finite_geom`).
     """
     best: tuple[str, tuple[float, float, float], tuple[float, float, float, float], int] | None = None
     childclass = body_el.get("childclass") or childclass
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

Adds finiteness guards to `_find_body_mesh` so the visual mesh geom's own `pos` and orientation - the two quantities only this reader parses - are refused when non-finite, closing the one body shape (contact-free mesh beside a collidable sibling, the Menagerie convention) that slipped past `_geom_aabb`'s existing guard because `_body_collision_aabb` returns before ever offering the visual geom. Both refusals delegate to the module's existing wording owner (`_refuse_non_finite_geom`), the `pos` guard sits before the offset composition so the message names the attribute rather than the sum, and the orientation refusal names the spelling the file used, mirroring `_geom_aabb`'s established pattern. The prior AST pin that asserted the *absence* of any finiteness test in the mesh walk is correctly narrowed to the body-offset quantity it actually establishes, and the new suite pins premises, regression, controls, and the deliberately-unchanged behaviours (unparseable-attribute fallback, extreme-but-finite values, nested-body pos left to the accumulator). Verified locally: the two touched test files pass 73/73 at head, and the full `tests/simulation/isaac/` run's 3 failures reproduce identically on the merge base, so none is attributable to this PR. No public API, wire format, or schema change; changelog fragment present; `Raises:` block added per convention.

## What's good

- The fix routes through the shared refusal owner instead of raising its own wording, so no spelling drift (AGENTS.md > one owner for a value-domain guard).
- The shipped test class that had pinned the missing coverage is renamed and split to what it actually establishes, rather than deleted - the body-offset non-re-guard rule survives.
- Reachability is pinned as tests (MuJoCo compiles the leaking fixture on a fixture body, refuses it on a moving one), so the premise cannot rot silently.

PR #2837, opened 2026-08-27
fix(tools/pose_tool): hold a stored pose to the travel its joints are driven within

cagataycali:
## The defect

`load_pose` read a pose off disk and drove its positions straight into
`move_multiple_motors`. `MotorController.degrees_to_position` clamps every target
into the joint's configured `range` before scaling it onto the 12-bit
`Goal_Position` register, so a stored value outside that range was not refused but
silently rewritten to the mechanical limit -- while the envelope said the arm went
where the file asked.

The same value supplied as an argument was already refused. One tool, one
controller, one conversion, two answers depending on whether the number arrived in
the call or in a file:

| surface | target | verdict | `Goal_Position(shoulder_pan)` | envelope |
|---|---|---|---|---|
| `move_multiple` | `999.0` (argument) | `error` | none written | — |
| `load_pose` | `999.0` (stored) | **`success`** `"Moved to pose 'wide'"` | **4095 = +180.00 deg** | echoes `999.0` |
| `load_pose` | `nan` (stored) | **`success`** `"Moved to pose 'nanpose'"` | **4095 = +180.00 deg** | echoes `nan` |

`shoulder_pan` is configured `(-180, 180)`, so `999` is an **819-degree** silent
rewrite, at servo speed, indistinguishable on the wire from a deliberate
`position=180`. `nan` gets to the same encoding through the clamp itself, since
`min(max_deg, nan)` returns `max_deg` -- the guard that looks like a safety net is
what fabricates the command.

`_joint_target_error`'s own docstring names this harm for the argument path: *"the
arm travels to the end stop, and the caller is told it moved to the value it asked
for, because the success text echoes the request."* On the stored path the echo is
`target_positions`, copied verbatim out of the artifact.

## Why the existing deferral could not cover it

When the argument guard landed, the stored pose was deliberately excluded and
deferred to `PoseManager.validate_pose`. That deferral does not hold, and its own
wording contains the hole -- *"already refuses an out-of-bounds position **but only
when the pose file carries `safety_bounds`**"*:

- `safety_bounds` is `None` for every pose this tool writes. There is exactly one
  `store_pose` call site and it passes three positional arguments; no caller
  anywhere supplies the field.
- `validate_pose` therefore returns `(True, "No safety bounds defined")`
  unconditionally on the load path -- measured on a pose the tool itself stored.

So a check does run before the drive, and it can never refuse anything.

The pose file is also a plain-JSON artifact: `_load_poses` reads it back through
`RobotPose.from_dict` with nothing checked, and Python's `json` writes and reads
`NaN` as a literal, so a non-finite target persists across a round trip.

## The change

`load_pose` routes its stored positions through `_joint_target_error` -- the same
authority the `position` / `positions` / `delta` arguments are held to -- and
refuses before the port is opened, for the reason the argument check gives above:
the arm must not travel to an end stop on a target that could not be honored.

The new helper restates no bound of its own; that is asserted structurally rather
than left to convention, because a second copy of the travel could disagree with
the one the servo is driven from. The refusal names the pose as well as the joint
and the bound, since a stored value is corrected in a file rather than in the next
call:

```
load_pose: pose 'wide' positions['shoulder_pan'] must be within [-180, 180] degrees
(the configured travel of 'shoulder_pan'), got 999.0.
```

`reset_to_home` stays out of scope and is now pinned by **reading its literals**
rather than by assertion, so retuning one out of range fires. A stored motor with
no configured `range` still has no travel to be held to and is left to the mover
that cannot address it; finiteness, which needs no range, still applies.

## Tests

`tests/tools/test_pose_tool_stored_pose_target_domain.py`, 36 cases in 7 classes.
Behavioural pre-fix split (helper kept, call site removed, so the split is not an
import error): **14 failed / 22 passed**, 14/14 in the two regression classes and
**0** in all five premise, structural, control and boundary classes. A raw revert
is a collection error, since the tests import the new symbol.

| mutation | new file | `collected` | siblings |
|---|---|---|---|
| control (no mutation) | 0 | 36 | 0 |
| call site removed (behavioural revert) | 14 | 36 | 0 |
| guard moved after the port is opened | 1 | 36 | 0 |
| bounds restated locally instead of delegated | 3 | 36 | 0 |
| deferred to `validate_pose` instead | 17 | 36 | 0 |
| logged and driven anyway | 14 | 36 | 0 |
| label drops the pose name | 2 | 36 | 0 |
| label drops the joint name | 1 | 36 | 0 |
| helper never reports (vacuous) | 16 | 36 | 0 |
| over-reach: an unknown motor refused for its range | 1 | 36 | 0 |
| `motor_name` slipped to the pose name | 10 | 36 | 0 |

10 mutations, all 10 detected, 9 distinct firing sets, `collected` stable at 36 on
every row, source restored byte-identical. The two rows sharing a set are the
behavioural revert and log-and-drive-anyway: both mean the pose is driven, so the
same cells report it -- which is the measured argument that reporting alone is not
the fix.

`siblings` is failures in the five pre-existing `pose_tool` suites (277 cases): **0
on every row**, so nothing that shipped could observe any of this. The reason is
visible in the boundary class that pinned the exclusion, and in the argument
guard's own parametrized targets -- all of them arguments, never a stored pose.

One of my own cells was measured vacuous and repaired rather than kept: asserting
the joint name in the whole message is satisfied by the shared domain's own
`(the configured travel of 'X')` clause, so dropping it from the label fired
nothing. The label is now graded alone, on the text before `must be`.

## Gate

- Paired, identical 236-file selection (all of `tests/tools/` plus every suite
  walking the tree, parsing source, or reading docstrings and `changelog.d`), new
  file excluded from both trees and every path verified present on both -- exactly
  one exclusion. Identical **10187 collected** and `10122 passed, 65 skipped, 0
  failed` on each, distinct rootdirs.
  merge base, both `comm` directions over the normalized message sets empty, **0
  attributable** to the changed files.
- `ruff check` and `ruff format --check` clean over 1643 files.
- No visual artifact: this is a refusal-domain change, not a policy, simulation,
  rendering, recording or asset one. The measured `Goal_Position` table above is
  the evidence.

### DIFF
--- a/changelog.d/2837-stored-pose-target-domain.md
+++ b/changelog.d/2837-stored-pose-target-domain.md
@@ -1,0 +1,36 @@
+### Fixed: a stored pose is held to the travel its joints are driven within
+
+`load_pose` read a pose off disk and drove its positions straight into
+`move_multiple_motors`. `MotorController.degrees_to_position` clamps every
+target into the joint's configured `range` before scaling it onto the 12-bit
+`Goal_Position` register, so a stored value outside that range was not refused
+but silently rewritten to the mechanical limit -- while the envelope reported
+`Moved to pose '<name>'` and echoed `target_positions` verbatim from the
+artifact. Measured on `shoulder_pan` (configured `(-180, 180)`), a pose naming
+`999` drove `Goal_Position` 4095, which is +180 degrees, and reported `999`; a
+stored `NaN` reached the same end stop, because `min(max_deg, nan)` returns
+`max_deg`, and `NaN` round-trips through the pose file as a JSON literal.
+
+The identical value supplied as an argument was already refused:
+`_joint_target_error` holds `position`, `delta` and the values of `positions` to
+each joint's travel. So one tool answered two ways for one number depending on
+whether it arrived in the call or in a file, driving the same servo through the
+same conversion. `load_pose` now routes its stored positions through that same
+authority, restating no bound of its own, and refuses before the port is opened
+for the reason the argument check gives -- the arm must not travel to an end stop
+on a target that could not be honored. The refusal names the pose as well as the
+joint and the bound, because a stored value is corrected in a file rather than in
+the next call.
+
+The deferral this replaces could not fire, which corrects the record left when
+the argument guard landed: `PoseManager.validate_pose` consults the pose's own
+optional `safety_bounds` and answers `No safety bounds defined` when the field is
+absent, and the single `store_pose` call site never supplies it -- so every pose
+this tool writes passed that check unconditionally. The arm's travel is a
+property of the arm, not an annotation a pose file has to carry.
+
+`reset_to_home` stays out of scope and is now pinned by reading its literals
+rather than by assertion: it supplies its own targets and every one of them is
+inside its joint's travel. A stored motor with no configured `range` still has no
+travel to be held to and is left to the mover that cannot address it, while
+finiteness -- which needs no range -- still applies.
--- a/strands_robots/tools/pose_tool.py
+++ b/strands_robots/tools/pose_tool.py
@@ -414,6 +414,38 @@ def _pose_target_error(
     return _joint_delta_error(action, motor_name, delta)
 
 
+def _stored_pose_target_error(pose: RobotPose) -> str | None:
+    """Error text when a stored pose names a target its joint cannot be driven to.
+
+    A pose read back from disk reaches a servo through the same
+    :meth:`MotorController.degrees_to_position` as a caller-supplied one, so it
+    is bounded by the same configured travel. The bounds are not restated here:
+    every position is delegated to :func:`_joint_target_error`, so a stored
+    target and an argument target are held to one authority and cannot drift.
+
+    ``PoseManager.validate_pose`` is not this check. It consults the pose's own
+    optional ``safety_bounds``, and returns "No safety bounds defined" when the
+    field is absent -- which it is for every pose this tool writes, because
+    ``store_pose`` is only ever called without it. The arm's travel is a
+    property of the arm, not an annotation a pose file has to carry.
+
+    The label names the pose as well as the joint: the value lives in a
+    persisted artifact rather than in the call, so a caller who is refused needs
+    to know which stored pose to correct.
+
+    Args:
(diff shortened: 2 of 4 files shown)


yinsong1986 (approved):
## Summary

`load_pose` previously drove stored pose positions straight into `move_multiple_motors`, where `degrees_to_position`'s clamp silently rewrote any out-of-travel (or `NaN`) target to the mechanical end stop while the success envelope echoed the stored value verbatim. This PR routes the stored positions through `_joint_target_error` -- the same authority the `position` / `positions` / `delta` arguments are already held to -- and refuses before the port is opened, so a refused pose opens no port and writes no `Goal_Position`. The new `_stored_pose_target_error` helper restates no bound of its own (asserted structurally via AST in the new tests), and the refusal names the pose, the joint, and the configured travel so the caller knows which persisted artifact to correct. The stale deferral to `PoseManager.validate_pose` (which can never fire, since no `store_pose` call site supplies `safety_bounds`) is corrected in the boundary-pinning docstring of `test_pose_tool_target_domain.py`, and `reset_to_home`'s out-of-scope status is now pinned by reading its literals rather than by assertion.

## What's good

- Guard placement is correct: after `validate_pose`, before `controller.connect()`, so a refused pose has no partial side effect and cannot hold the bus.
- One authority for the bound: the helper delegates to `_joint_target_error` and the delegation is pinned structurally (`TestTheBoundsHaveOneAuthority` asserts no numeric constant appears in the helper body), so the stored path and the argument path cannot drift.
- Verified on the head commit: the 36 new tests pass, the full `tests/tools/` suite passes (3273 passed, 23 skipped), the tree-wide grading suites (Args/Raises docstring completeness, ASCII tool-result strings, host paths, except-tuple scope, xref resolution) pass, and `ruff check` / `ruff format --check` are clean on all three changed Python files.
- Changelog fragment present (`changelog.d/2837-stored-pose-target-domain.md`), per AGENTS.md step 3.
- Every test reaching the motor path passes an explicit fake `port`, so the suite cannot drive a real arm plugged into the CI host.
- No public API, wire format, or persisted schema changes: the helper is private, the pose file format is untouched, and the behaviour change is an explicit refusal naming the remedy rather than a silent default change.

PR #2838, opened 2026-08-27
fix(isaac): refuse a non-finite mesh scale instead of measuring through it

cagataycali:
## Problem

`load_mesh_geometry` already refuses a vertex coordinate that is not finite, and `_non_finite_vertex_error` states exactly why: `min`/`max` order a NaN as neither smaller nor larger than anything, so the running comparison `mesh_aabb` uses "drops it and returns the bounds of the vertices that are finite - the same numbers a mesh declaring only those would produce, under no error at all". That rule was applied where the four parse paths meet.

`scale` reaches the same comparison and did not go through it. `mesh_aabb` measures `vertex * scale`, so a non-finite component poisons every vertex on that axis and the comparison drops all of them, leaving the axis at its `inf`/`-inf` seed. One function, two inputs: the coordinates it reads from disk were checked, the transform it applies to them was not - and that one arrives from the caller.

The reachable route is an MJCF `<asset><mesh scale=...>`. `_parse_mjcf_mesh_assets` reads it through `_parse_axis`, which returns three floats, so `nan` and `inf` pass. MuJoCo compiles such a model, so the reader cannot defer the question to the format's owner. `load_mjcf_scene_objects` then measures the mesh's own bounds for a **mesh-only** body - the `elif mesh_path is not None` branch whose own comment says it exists "so a mesh-only body can fall back to the mesh's own bounds".

## Measured

A 0.2 m mesh on a mesh-only body, `MUJOCO_GL=egl`:

| declared scale | MuJoCo | reported `position` | reported `size` | non-finite reported fields |
|---|---|---|---|---|
| `1 1 1` | compiles | `(0.1, 0.1, 0.6)` | `(0.2, 0.2, 0.2)` | none |
| `nan 1 1` | compiles | `(nan, 0.1, 0.6)` | `(0.0001, 0.2, 0.2)` | `position` |
| `nan nan nan` | compiles | `(nan, nan, nan)` | `(0.0001, 0.0001, 0.0001)` | `position` |
| `inf 1 1` | compiles | `(inf, 0.1, 0.6)` | `(nan, 0.2, 0.2)` | `position`, `size` |

The NaN row is the worse half, and it is the one a consumer cannot screen for. Because the reported extent is floored at `1e-4`, a NaN axis leaves the **size finite** - a plausible 0.1 mm, 2000x smaller per axis than the asset - so screening the reported fields for non-finite values catches the centre and not the size.

The in-scene control is the same declaration on a body that also carries a collidable box: the analytic bound wins, the measurement is never reached, and `position`/`size` are correct. The leak was specific to the branch that exists to prefer the asset itself.

## Change

`mesh_aabb` refuses a non-finite `scale` through `_non_finite_scale_error` - a sibling to the vertex wording rather than a reuse of it, because a bad coordinate in the file and a bad transform from the caller are two causes and so two messages. The check precedes the parse: it needs nothing from the file, and a transform that can never be applied is not worth reading a mesh for. A caller whose path is wrong and whose scale is fine still gets the file contract's own message.

A finite scale is untouched, including the two spellings that look degenerate and are not. `0.0` is a finite request to flatten an axis, where the `1e-4` floor is doing the job it is there for; a negative scale mirrors the asset, where the running comparison is order-independent and the bound is already right.

Both of those are real rather than hypothetical. Across the 811 MJCF models cached on the measuring machine there are **17322** `<mesh>` registry entries, **1631** declare a non-unit scale, and those include `(-1, -1, -1)`, `(-0.001, 0.001, 0.001)` and millimetre scales - so the accepted boundary is what the shipped corpus actually uses.

## Safety

**0** of the 17322 entries is non-finite, so the defect is latent. **9** of the 59 loadable registry robots reach this measurement, one of them through a millimetre scale, so the branch is live. All **60** loadable registry scenes report byte-identical objects before and after (`md5 e64a4b72b6de6e8e325e161d9ba1cd87` on both trees, 0 scenes changed).

## Tests

`tests/simulation/isaac/test_mesh_asset_scale_must_be_finite.py`, 39 cases. Pre-fix **22 failed / 17 passed**, with **0 of 6 premise, 0 of 6 control and 0 of 2 boundary cells** failing - the premises (MuJoCo compiles the model; the asset's own vertices are finite; the extent floor is what made the NaN case unscreenable) and the accepted-scale controls hold either way, which is what makes them premises and controls.

11 mutations, 10 detected, **10 distinct firing sets**, control clean, `collected` stable at 39 on every row, and **0 failures in 6 pre-existing suites** on every row:

| mutation | fires | note |
|---|---|---|
| guard removed | 21 | the behavioural revert |
| checks `scale[0]` only | 8 | the axis-1 and axis-2 cells |
| `all` -> `not any` | 16 | refuses only an all-non-finite scale |
| guard moved after the parse | 2 | the ordering |
| over-reach: `0.0` refused too | 2 | fires the flattened control and a premise |
| over-reach: negative refused | 3 | fires the mirrored control and a premise |
| message drops the asset path | 1 | |
| message drops the `1e-4` cause | 1 | |
| raises inline, not via the factory | 3 | structural |
| reuses the vertex wording | 20 | one owner per cause |

The two over-reach rows are the ones that make the accepted boundary load-bearing: a guard that also refused `0.0` or a negative scale would break spellings 1631 shipped `<mesh>` entries use.

The structural class derives its universe from the module rather than listing it, so a third non-finite cause is held to the same rule when it lands instead of inheriting an exemption by being absent from a tuple.

## Gate

`ruff check` and `ruff format --check` clean over 1646 files. `mypy strands_robots tests tests_integ`: `Success: no issues found in 1643 source files`. Paired run over an identical 571-file selection (all of `tests/simulation/isaac/` plus every suite that walks the tree, parses source, or reads docstrings) with the new file excluded from both trees: **25792 collected and `7 failed, 25667 passed, 118 skipped` on each**, the same 7 ids, both `comm` directions empty. Those 7 are the measuring machine's own environment - 5 assert `rclpy` is absent where it resolves from a system install, and 2 need `docker`.

## Visual

Rendered from one MJCF scene, `MUJOCO_GL=egl`, one camera and one light for all panels. Masks come from separate segmentation passes per geom group, so the proxy's transparency cannot affect a pixel count.

non-finite mesh scale

At unit scale the reported proxy encloses the asset: **0** of its 15294 pixels fall outside its own bound. For `scale="nan nan nan"` the reported bound occupies **0 pixels** and **all 15294** asset pixels fall outside it. The reported centre is NaN and a renderer cannot place it, so panel 2 draws the bound at the asset's own centre - a declared substitution, with only the extent being the measurement.

### DIFF
--- a/strands_robots/simulation/isaac/mesh_assets.py
+++ b/strands_robots/simulation/isaac/mesh_assets.py
@@ -123,6 +123,42 @@ def _non_finite_vertex_error(mesh_path: str, index: int, vertex: tuple[float, fl
     )
 
 
+def _non_finite_scale_error(mesh_path: str, scale: tuple[float, float, float]) -> ValueError:
+    """The refusal for a mesh scale whose components are not all finite.
+
+    :func:`load_mesh_geometry` already refuses a vertex that is not finite,
+    for the reason :func:`_non_finite_vertex_error` states: ``min``/``max``
+    order a NaN as neither smaller nor larger than anything, so the running
+    comparison drops it and reports the bounds of what remains. ``scale``
+    reaches that same comparison - :func:`mesh_aabb` measures
+    ``vertex * scale`` - so a non-finite component there poisons every
+    vertex on that axis and the comparison drops all of them, leaving the
+    axis at its ``inf``/``-inf`` seed. The asset's own coordinates being
+    finite is therefore not enough: the transform applied to them has to be
+    finite too, and it arrives from the caller rather than from the file.
+
+    The two components fail differently and neither is screenable:
+
+    * A NaN axis reports a centre of NaN and, because the reported extent is
+      floored at ``1e-4``, a **finite** size of 0.1 mm. A consumer that
+      screens the reported fields for non-finite values catches the centre
+      and not the size, so a metre-scale asset passes that screen as a
+      plausible sub-millimetre one.
+    * An infinite axis reports an infinite centre and a size of NaN, because
+      the extent is ``inf - inf``.
+
+    A scale of ``0.0`` stays accepted: it is a finite request to flatten the
+    asset on that axis, and the ``1e-4`` floor is doing the job it is there
+    for. A negative scale stays accepted too - it mirrors the asset, and the
+    running comparison is order-independent, so the bound is right.
+    """
+    return ValueError(
+        f"mesh {mesh_path}: scale has a component that is not finite ({scale!r}) - a mesh measured "
+        f"through it reports a centre that is not a position and an extent floored at 1e-4, so the "
+        f"asset is refused instead of being sized"
+    )
+
+
 def load_mesh_geometry(
     mesh_path: str,
 ) -> tuple[list[tuple[float, float, float]], list[int], list[int]]:
@@ -391,7 +427,10 @@ def mesh_aabb(
         Filesystem path to a ``.obj``, ``.stl`` or ``.msh`` (legacy MuJoCo binary mesh) file.
     scale : tuple[float, float, float]
         Per-axis scale applied to the vertices before measuring (an MJCF
-        ``<mesh scale=...>``). Defaults to unit scale.
+        ``<mesh scale=...>``). Defaults to unit scale. Every component must
+        be finite; it is checked before the asset is parsed, because a
+        transform that can never be applied is not worth reading a mesh
+        for, and the check needs nothing from the file.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


yinsong1986 (approved):
## Summary

Adds a finiteness guard on the `scale` argument of `mesh_aabb` in `strands_robots/simulation/isaac/mesh_assets.py`: a NaN or infinite component is now refused with a `ValueError` from a new named factory (`_non_finite_scale_error`) before the mesh file is parsed, instead of poisoning the running min/max comparison and reporting a NaN centre alongside a floored, plausible-looking 0.1 mm extent. The guard closes the gap left by the existing vertex-finiteness check (the file can be clean while the caller-supplied transform is not), matches AGENTS.md's raise-on-fatal / no-silent-defaults rules, and deliberately keeps `0.0` (flatten) and negative (mirror) scales accepted. Both callers were checked: the MJCF mesh-only fallback in `loaders.py` is the intended reachable route and is pinned end-to-end; `simulation.py`'s call uses the default unit scale and is unaffected.

## What's good

- The regression test (`tests/simulation/isaac/test_mesh_asset_scale_must_be_finite.py`, 39 cases) pins every axis and both non-finite values, the end-to-end MJCF route, the premises that make the guard necessary, and controls proving finite scales (unit, non-uniform, mirrored, flattened, millimetre) measure exactly what they did before.
- The `Raises` docstring block was updated in the same change, per AGENTS.md's "a `Raises:` block names every refusal the function itself makes".
- Guard-before-parse is pinned structurally (AST ordering), so a caller with a bad scale and no file on disk gets the accurate diagnosis.
- Behaviour-describing test names, `tmp_path`-scoped fixtures, ASCII-only messages, ruff clean.

PR #2839, opened 2026-08-27
fix(mesh): report a timed drive whose trailing stop was refused

cagataycali:
## Problem

A timed or repeated `drive` owns its own stop. `MobileBaseRobot`, `RosbridgeRobot` and `AckermannRosRobot` all follow a non-zero command with a single zero one from `finally`, so the zero goes out even when the main publish raised.

Sending it is not the same as landing it. That zero is a second command over the same transport, and every shipped graph tool reports a refusal - a declined operator approval, a rate limit, a transport failure - as an error envelope rather than by raising. So the refusal was a value that had to be read, and three of the four drive-owning classes dropped it inside the `finally` and returned the hold's own success.

Both discarding sites carried a comment naming the exact consequence:

> Carries the same context as the command it undoes: a trailing zero that could not reach the gate would leave the robot latched at speed.

## Measurement

Against the real `use_rosbridge` gate, with an operator who approves the hold and declines the stop, on a blocklisted `/cmd_vel`:

| | before | after |
|---|---|---|
| operator prompts | 2 | 2 |
| `drive()` status | **`success`** | `error` |
| `drive()` text | `published 2 message(s) to /turtle1/cmd_vel` | names the topic, the still-live velocity, `Halt it with stop`, and the operator's decision |
| messages on the wire | 2 | 2 |
| last message | **`linear.x = 1.0`** | `linear.x = 1.0` |
| trailing zero sent | no | no |

The robot is left moving at the commanded velocity and the caller is told the drive succeeded. The `drive_<node>` tool description promises a timed command "stops automatically afterwards", so a caller reading `success` never issues `stop`. The gate prompts once per publish - pinned by `test_the_trailing_zero_of_a_timed_drive_reaches_the_same_operator` - so approving a hold and declining its stop is an ordinary pair of decisions rather than a contrived input.

`RtpsRobot` reproduces it through the shared base: `drive(linear=1.0, count=2)` with the same two decisions answered `success` / `published 2 message(s)`.

## Change

`AckermannRosRobot` already kept the verdict and compared it against the hold's own result, which is why this is a consolidation rather than a new rule. Its docstring claims the "same contract as `RosBridgedRobot.drive`" - the reverse of the truth.

The rule now lives once as `_mobile_base.failed_halt_error`, beside the drive contract it belongs to, and all three drive owners consult it. The car's message is byte-identical, which is what its 61 existing tests measure. Two decisions carry across from the exemplar: a hold that failed is still the cause, because a stop it never got to undo is a consequence of it; and a single-shot command owes no stop, so no verdict is read and it latches by contract as before.

Prose corrected where a reader learns the contract. The shared safety contract stated the guarantee as fact - "a timed drive cannot leave a robot with a live velocity latched" - and now says the stop's verdict is read rather than dropped. `docs/rosbridge-integration.md` says the same.

## Why nothing caught it

The rosbridge gate suite covers *approve both* (`test_the_trailing_zero_of_a_timed_drive_reaches_the_same_operator`) and *decline the hold* (`test_a_declined_hold_never_latches_a_velocity`). Approve the hold and decline the stop - the case that test's own docstring names, "one that could not reach the gate would leave the robot latched at the speed of a hold the operator did approve" - had no cell.

## Tests

`tests/mesh/test_drive_reports_a_refused_trailing_stop.py`, 21 cases. Pre-fix split, with the helper and prose kept and only the three call sites reverted to discarding: **6 failed / 15 passed** - 4 of 4 in the regression class, 1 of 1 on the shared base, 1 of 3 structural, and **0 of 4 controls, 0 of 1 premise, 0 of 6 rule-level, 0 of 2 prose**. A raw revert is a collection error, since the helper does not exist on the base.

Mutation table, `mine` = this file, `sib` = the 13 pre-existing mesh drive and gate suites:

| row | mutation | mine | coll | sib |
|---|---|---|---|---|
| b0 | control | 0 | 21 | 0 |
| b1 | shared base discards the verdict again | 2 | 21 | 0 |
| b2 | rosbridge discards it again | 5 | 21 | 0 |
| b3 | ackermann discards it again | 1 | 21 | **1** |
| b4 | precedence dropped: a failed hold no longer wins | 1 | 21 | **1** |
| b5 | no-stop-owed guard dropped (a single-shot is judged) | 2 | 21 | **72** |
| b6 | a landed stop is reported as failed too | 2 | 21 | **12** |
| b8 | message drops the topic | 2 | 21 | 0 |
| b9 | message drops the imperative | 2 | 21 | 0 |
| b10 | message drops the stop's own cause | 3 | 21 | 0 |
| b11 | subject hardcoded, ignoring the platform | 1 | 21 | 0 |
| b12 | no-detail fallback dropped | 1 | 21 | 0 |

Every mutation is detected and the control is clean. b3 and b4 trip the Ackermann car's own pre-existing tests, which is the evidence the consolidation preserved its contract; b5 and b6 trip 72 and 12 sibling cases, which is what makes the two guards load-bearing fleet-wide rather than defensive.

Two cells were found vacuous by the table and tightened: the hold's success text is `published N message(s) to <topic>`, so `topic in text` passed on the discarding code, and `"stop"` also appears in the refusal's own subject ("the trailing stop failed"), so a bare `"stop" in text` survived the imperative being removed. Both now pin the clause only the refusal builds.

The structural inventory is derived, not listed, so a fifth bridge answers on arrival. Composed proof that this matters: revert rosbridge with the inventory derived, 5 cells fire including the structural one; revert rosbridge with the inventory hardcoded to `_mobile_base.py`, 4 fire and the structural cell goes silent; hardcode with nothing reverted, 0.

## Gate

- `ruff check` and `ruff format --check` clean over 1646 files.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a worktree of the merge-base, normalized message sets identical in both directions, 0 attributable.
- Paired run over an identical 535-file selection (all of `tests/mesh` plus every suite that walks the tree, parses source, or reads docstrings and `docs/`), this file excluded from both trees: identical **22401 collected** and `24 failed, 22282 passed, 71 skipped, 24 errors` on each, **48 identical failing ids with both `comm` directions empty**, distinct rootdirs. Of the 48, 45 need `awsiot`/`awscrt` (both absent locally, declared in the `[mesh-iot]` extra) and 3 are the lerobot-from-source `encode()` drift. `tests/mesh/test_zenoh_transport_security.py` is excluded from the pair and verified alone: 9 passed, 2 skipped on both.
- Rebased onto `2d151363` and re-verified there, including the two guard files the base added: 471 passed, 6 skipped.

No visual artifact: this changes what a refusal reports, not a policy, simulation, render, recording or asset. The measured before/after table is the evidence.

### DIFF
--- a/changelog.d/2839-drive-reports-a-refused-trailing-stop.md
+++ b/changelog.d/2839-drive-reports-a-refused-trailing-stop.md
@@ -1,0 +1,36 @@
+### Fixed: a timed drive whose trailing stop was refused reports it
+
+A timed or repeated `drive` owns its own stop: `MobileBaseRobot`, `RosbridgeRobot`
+and `AckermannRosRobot` all follow a non-zero command with a single zero one from
+`finally`, so the zero goes out even when the main publish raised. Sending it is
+not the same as landing it. That zero is a second command over the same
+transport, and every shipped graph tool reports a refusal - a declined operator
+approval, a rate limit, a transport failure - as an error envelope rather than by
+raising, so the refusal was a value that had to be read. Three of the four
+drive-owning classes dropped it inside the `finally` and returned the hold's own
+success.
+
+Measured against the real `use_rosbridge` gate with an operator who approves the
+hold and declines the stop, on a blocklisted `/cmd_vel`: `drive(linear=1.0,
+duration=1.0)` answered `status="success"` with "published 2 message(s)", two
+messages reached the wire, the last of them `linear.x = 1.0`, and no zero
+followed. The robot is left moving at the commanded velocity and the caller is
+told the drive succeeded - and the agent-facing tool description promises a timed
+command "stops automatically afterwards", so a caller reading `success` never
+issues `stop`. The gate prompts once per publish, so approving a hold and
+declining its stop is an ordinary pair of operator decisions. `RtpsRobot` and
+`RosBridgedRobot` reproduced it through the shared base.
+
+`AckermannRosRobot` already kept the verdict and compared it against the hold's
+own result, which is why this is a consolidation rather than a new rule - its
+docstring claimed the "same contract as `RosBridgedRobot.drive`", which was the
+reverse of the truth. The rule now lives once as
+`_mobile_base.failed_halt_error`, beside the drive contract it belongs to, and
+the car's message is unchanged. A hold that failed is still the cause, because a
+stop it never got to undo is a consequence of it; a single-shot command owes no
+stop, so no verdict is read and it latches by contract as before.
+
+The shared safety contract stated the guarantee as fact - "a timed drive cannot
+leave a robot with a live velocity latched" - so it now says that the stop's
+verdict is read rather than dropped, and `docs/rosbridge-integration.md` says the
+same where a reader learns the fleet contract.
--- a/docs/rosbridge-integration.md
+++ b/docs/rosbridge-integration.md
@@ -161,9 +161,12 @@ scan = robot.get_scan()  # read one laser scan (error if no scan_topic)
 - **Timed-command trailing zero**: every drive with a `duration` argument and a
   non-zero command (or multi-message publish) automatically publishes a single
   zero Twist afterwards - even if the main publish failed - so a timed drive
-  cannot leave the robot with a live velocity. This was previously this bridge's
-  alone; the ROS 2 and RTPS bridges inherit it from the shared mobile base now,
-  so a timed drive self-stops on all three.
+  does not leave the robot with a live velocity. That zero is itself a gated
+  command, so when it is the call that fails - a declined approval, a rate limit,
+  a transport error - `drive` returns that failure instead of the hold's success,
+  naming the still-live topic and telling you to halt with `stop`. This was
+  previously this bridge's alone; the ROS 2 and RTPS bridges inherit it from the
+  shared mobile base now, so a timed drive self-stops on all three.
 - **stop() needs no prior state**: the stop method publishes a zero Twist
   regardless of whether an earlier command succeeded, and there is no enable
   handshake to satisfy first. It is *not* exempt from the operator gate: that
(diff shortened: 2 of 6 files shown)


yinsong1986 (approved):
## Summary

This PR fixes a real safety-reporting gap: a timed/repeated `drive` sends its trailing zero from `finally`, but that zero is itself a gated command whose refusal arrives as an error envelope, and `MobileBaseRobot`, `RosbridgeRobot` (and `RtpsRobot`/`RosBridgedRobot` through the shared base) discarded that verdict and returned the hold's success - leaving the robot moving while the caller is told the drive stopped itself. The fix consolidates the verdict-read that `AckermannRosRobot` already had into one shared helper (`_mobile_base.failed_halt_error`) consulted by all three drive owners, with the exemplar's precedence preserved (a failed hold wins; a single-shot owes no stop and reads no verdict). The `finally` guarantee itself is untouched, so the stop still goes out even when the main publish raised, and the exception path is unaffected (the post-`try` verdict read is unreachable when the publish raises, so no unbound `result`). The helper lives in a private module and is not exported, so no public API surface changes; the tool-result envelope shape is unchanged and the success-to-error flip is exactly the reported bug. Verified independently: the new 21-test suite passes at the head, the pre-existing ackermann/rosbridge/rtps/fleet-scope drive suites pass unchanged (115 passed, 7 skipped), the full `tests/mesh` run's 45 failures/errors are byte-identical to the merge base (`2d151363`, all missing `awsiot`/`awscrt` extras), the repo-wide graders (ASCII tool-result strings, host paths, changelog fragment, docstring `Args:`/`Raises:` completeness, xref resolution, test naming) pass, and `ruff check`/`ruff format --check` are clean on the touched files.

## What's good

- The structural test (`TestOneOwnerForTheRule`) derives the drive-owner inventory rather than listing it, so a fifth bridge is graded on arrival - and the PR's composed proof that the derived inventory is load-bearing is the right kind of evidence.
- Two vacuous assertions (`topic in text`, `"stop" in text`) were found by the mutation table and tightened to the clause only the refusal builds - the kind of self-audit AGENTS.md's regression-pin rule asks for.
- Precedence decisions (failed hold wins; no stop owed means no verdict) are documented in the helper's docstring with the reasoning, and pinned by direct rule-level tests.
- Changelog fragment present per step 3; prose corrected in both the module contract and `docs/rosbridge-integration.md` where the old text overstated the guarantee.

PR #2840, opened 2026-08-27
docs(simulation/mujoco): say that the grow path carries the joint-force buffer it carries

cagataycali:
## What is wrong

`_recompile_preserving_state` carries both applied-force buffers across a scene grow, and its own docstring says one of them is not carried:

```
* ``qfrc_applied``, its joint-space sibling, is deliberately NOT carried
  here: nothing in this package ever writes a non-zero value into it.
  ... A joint-force API added later must add
  the carry with it -- ``_SceneState`` already carries this buffer on the
  eject path, where it is free because that path snapshots joints anyway.
```

The carry is at lines 915 and 956 of the same function (`_snapshot_joint_forces` / `_restore_joint_forces`). So one docstring makes three claims:

| claim | where | true? |
|---|---|---|
| "Both are therefore snapshotted by name before the recompile and re-applied after it" | earlier in the same docstring | yes |
| "``xfrc_applied`` ... IS carried here" | the bullet above | yes |
| "``qfrc_applied`` ... is deliberately NOT carried here" | the bullet below | no |

The false half is not inert: it closes with an instruction to add a carry that is already there. `_SceneState`'s docstring, meanwhile, has the right answer and even names the helper ("on the grow path by `_snapshot_joint_forces` and `_snapshot_body_wrenches`, since `spec.recompile` carries neither").

## Where it came from

The bullet was written when the claim was true, and its reasoning is what the later measurement contradicted.

| commit | did |
|---|---|
| `b3939ca8` | added the bullet, reasoning that nothing in the package writes the buffer, so there is nothing for the recompile to lose |
| `f034ccde` | measured what that reasoning missed and shipped `_snapshot_joint_forces` with its restore |

`f034ccde`'s own measurement: `remove_robot('doomed')` kept a 0.5 N m latched torque and `add_object('crate')` dropped it, so a hinge stopped turning the moment anything entered the scene under a `"status": "success"`. Its commit message notes it was itself correcting a stale prose claim about the same pair. It updated the earlier paragraph and `_SceneState`'s docstring, and left this bullet.

## The change

Corrects the bullet to say the buffer is carried, by which helper, and why carrying it is right even though no public setter writes it (the eject path snapshots it with the joint state, and `save_state` checkpoints it so a restore reinstates latched forces rather than only positions). Also names the helper pair in the `xfrc_applied` bullet, matching what `_SceneState`'s docstring already does. No behaviour changes.

## Why nothing caught it

`f034ccde`'s suite is entirely behavioural: it measures the joint's angle, which cannot see a sentence. Running it against this docstring in either state gives the same result.

## The guard

Both halves are derived from the module rather than listed, so a buffer or a carrying function added later is graded on arrival:

- the functions checked are those that call a `_snapshot_*` helper
- the buffers each one carries are the `data.<buffer>` fields those helpers read

Measured on the module: `_recompile_preserving_state` carries exactly `{qfrc_applied, xfrc_applied}`, attributed to `_snapshot_joint_forces` and `_snapshot_body_wrenches` respectively; the discovery pass reaches 8 functions, 3 of which carry a buffer.

Only the negative claim is graded: a carried buffer must not be documented as uncarried. The positive form, that every carrying helper be named in its caller's docstring, is deliberately not required. `_snapshot_scene_state` carries `xfrc_applied` under a one-line docstring and three spec-rollback callers reach `_snapshot_spec` without naming it, so demanding a helper name from each would be a rule about docstring length rather than a check on whether the prose is true.

A sentence is the unit rather than a bullet, because the docstring's bullets run into the prose that follows them and a denial in a trailing paragraph would otherwise be attributed to the bullet above it.

## Pre-fix split

With the stale bullet restored and the guard in place: **2 failed, 9 passed**.

|---|---|---|
| `TestNoCarriedBufferIsDocumentedAsUncarried` | regression | 2 / 2 |
| `TestTheCarriedSetIsDerivedFromTheCode` | premise | 0 / 3 |
| `TestTheRuleIsNotVacuous` | constructed exemplars | 0 / 6 |

The failure names the offending sentence verbatim rather than reporting a count.

## Mutations

9 mutations, control clean, all 8 detected, `collected` stable at 11 on every row, and `sib` (failures in the four pre-existing suites over this module's docstrings and force buffers) 0 on every row.

|---|---|---|---|
| b0 | 0 | 0 | control, no change |
| b1 | 2 | 0 | the stale bullet restored, i.e. the defect |
| b2 | 1 | 0 | an Nth carrying function denies its buffer, discovery derived |
| b3 | 0 | 0 | the same plant, discovery hardcoded to today's names |
| b4 | 2 | 0 | negation widened to what the compiler drops |
| b5 | 1 | 0 | buffer derivation reads any attribute, not only `data.*` |
| b6 | 4 | 0 | buffer derivation reads `model.*` instead of `data.*` |
| b7 | 1 | 0 | discovery narrowed to the rebuild alone |
| b8 | 1 | 0 | sentence scope replaced by the whole docstring |

`b2` against `b3` is why the inventory is derived: the same planted function that denies a buffer it carries is named by the derived scan and is completely invisible to a hardcoded list equal to today's set.

`b4` is why the negation pattern is narrow. The same docstring correctly says `spec.recompile` "transfers neither applied-force buffer at all" and that both "come back entirely zero" - true statements about the compiler, and the reason the carry exists. A pattern reading "not transferred" or "dropped" as a denial flags those, and the exemplar that states the compiler's behaviour catches it.

`b8` was measured at 0 on a first pass, which showed the sentence scope was ungraded: no exemplar could distinguish it from reading the docstring whole. Added one where a true denial about an uncarried buffer sits beside a carried buffer's bullet; the row then fires 1.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1646 files)
- `mypy strands_robots tests tests_integ`: 24 errors on both this branch and the merge-base, normalized message sets byte-identical in both directions, 0 attributable to either changed file. The only differing line is the summary's file count (`checked 1643` against `1642`, the one new test file).
- paired run against the merge-base over an identical 378-file selection (all of `tests/simulation/mujoco/` plus every suite that walks the tree, parses source or reads docstrings), the new file excluded from both: identical **13957 collected** and **13882 passed, 75 skipped, 0 failed** on each tree, from distinct rootdirs.

No visual artifact: this changes a docstring and adds a test. Nothing about a policy, a simulation behaviour, rendering, recording or a robot asset moves, and the paired gate above is the evidence for that. The measured evidence for the change itself is the pre-fix split and the mutation table.


yinsong1986 (approved):
## Summary

Corrects the one false bullet in `_recompile_preserving_state`'s docstring: `qfrc_applied` has been carried on the grow path since `_snapshot_joint_forces`/`_restore_joint_forces` landed, and the stale bullet closed with an instruction to add a carry that already exists. I verified the correction against the code at this head: the rebuild calls both snapshot helpers before `spec.recompile` and both restores after, and the helpers read exactly `data.qfrc_applied` and `data.xfrc_applied` respectively, so every claim in the rewritten bullets is true of the tree it documents. The new guard derives both halves from the module (functions that call a `_snapshot_*` helper; the `data.<buffer>` fields those helpers read) rather than hardcoding today's names, grades the negative claim only, and scopes to the sentence rather than the whole docstring -- I reconstructed the stale bullet and confirmed the guard names it verbatim, and reports nothing on the corrected prose. All 11 tests pass on the head tree with no `mujoco` dependency (source-text only, as the module docstring says), the constructed exemplars pin both accept and reject outcomes including the true-statements-about-the-compiler case that a naive denial pattern would flag, and the changelog fragment follows the step-3 convention. No behaviour changes; nothing here touches a public API, wire format, or persisted schema.

## What's good

- The guard's inventory is derived, not listed -- a buffer or carrying function added later is graded on arrival, and the exemplar pair in `TestTheRuleIsNotVacuous` measures the derived-vs-hardcoded difference the PR claims.
- The denial pattern is deliberately narrow, with an exemplar pinning that the docstring's true statements about what `spec.recompile` drops are not misread as denials of the carry.
- Test class names describe behaviour, every helper has full `Args:` entries, no host paths, and the scope note documents why the positive form (naming helpers in every caller) is deliberately not required.

PR #2841, opened 2026-08-27
test(drivers/g1): pin the battery floor as unreachable behind the un-wired FSM

cagataycali:
## Summary

`G1Driver._check_motion_gates` refuses with `FSM id unknown - motion-switcher source has not been wired` on every connected driver, because `_fsm_id` has no producer: the decoder that would write it is the deferral the refusal itself names, and #2765 carries the wire-side decision. Every guard behind that check is therefore unreachable through the verb surface an agent calls, including the `battery_floor_pct` guard the constructor asks the caller to think about and `get_status` reports back.

Nothing graded that. The shipped suite injects `_fsm_id` on every driver it builds (`tests/drivers/test_g1_driver.py`, nine sites), so it reads the battery guard past a gate the vendor cannot currently open. A caller who read those tests green over `send_action` had no way to tell that on hardware the battery guard never fires.

## Contracts pinned

Three literal contracts, so a change to any of them fires this file rather than a distant one.

1. **Reachability.** On a fully-healthy driver except for `_fsm_id` — `_connected=True`, `_mode_machine` from a real `rt/lowstate` decode, `_battery` from a real `rt/lf/bmsstate` decode — `send_action` at 1.0% against a 15.0% floor refuses with `FSM id unknown`. The refusal does not mention `battery` and does not carry the pack percentage.

2. **Producer count.** `_fsm_id` has exactly one assignment in `strands_robots.drivers.g1` and its assigned value is the `None` literal. The walk is over the parsed source, so a docstring mentioning `self._fsm_id = 500` is not counted as a write. The value is graded as an `ast` node rather than by substring, because the snippet carries the annotation `int | None` and so still reads `None` under `self._fsm_id: int | None = 501` — the silent gate-open the contract has to refuse.

3. **Ordering.** The FSM check precedes the battery-floor comparison inside the gate. Contract (1) reads whichever gate fires first, so the order is what makes it a statement about the battery floor rather than about whichever guard happened to be reached. Stating it separately means a reordering fails a test whose name says `ordering`, instead of only failing (1) with a message about `FSM id unknown` for a change that was about the order.

## Break-table

| mutation to `strands_robots/drivers/g1.py` | (1) reachability | (2) producer | (3) ordering |
|---|---|---|---|
| pristine tree | pass | pass | pass |
| battery check moved above the FSM check | fail | pass | fail |
| a second floor comparison added after it | pass | pass | fail |
| `self._fsm_id: int \| None = 501` | fail | fail | pass |
| the `None` initialiser deleted | fail | fail | pass |

Row 2 is the failure the ordering contract exists to name: without (3) it reports `assert 'FSM id unknown' in 'battery 1.0% is under floor 15.0%'`, which reads as a reachability regression for a change that was about the order. Row 3 is detection (1) does not have — a second comparison is harmless for the refusal this file drives, so only (3) sees it. Rows 4 and 5 share a firing set because both mean the same thing: the single write site is no longer the `None` literal. `collected` is 3 on every row, and the source is restored byte-identical afterwards.

## What stays open

Whether FSM-before-battery is the right order. The gate justifies it with "the caller has already been told the FSM if the FSM is the reason", which holds for the FSM-*value* refusal (`FSM 500 refuses arm writes`) and not for the FSM-*unknown* branch that fires on every driver today — there the FSM is precisely what the caller was not told. That is a maintainer's call about which fact is more urgent to surface, and this file records the current arrangement rather than arguing for it.

## Sunset condition

The day a real motion-switcher source writes `_fsm_id`, contracts (1) and (2) fire — which is the correct signal, because on the same day the battery floor becomes reachable through the verb surface. The failure message names the second write site so the replacement is targeted: this file is then replaced by one grading the reachable battery floor.

## Gate

- the three contracts: 3 passed
- `tests/drivers/`, `tests/tools/g1/`, `tests/test_driver_seam.py` plus the guard `main` added in #2838: 822 passed, 3 skipped, 0 failed
- `ruff check` clean, `ruff format --check` clean over 1646 files
- mypy at the tree's standing count with 0 errors in the changed file
- the changelog-fragment gates and the docstring reference guards: 40 passed

Refs #2765.

### DIFF
--- a/changelog.d/2841-g1-battery-floor-reachability-pinned.md
+++ b/changelog.d/2841-g1-battery-floor-reachability-pinned.md
@@ -1,0 +1,34 @@
+### Added: the g1's battery floor is pinned as unreachable behind the un-wired FSM
+
+``G1Driver._check_motion_gates`` refuses with ``FSM id unknown - motion-switcher
+source has not been wired`` on every connected driver because ``_fsm_id`` has no
+producer -- the decoder that would write it is the exact deferral harness#361
+PR-C and #2765 name. Every guard behind that refusal, including the
+``battery_floor_pct`` guard the caller is asked to think about, was
+unreachable through the verb surface, and no test asserted it. The shipped
+suite injected ``_fsm_id`` on every driver it built and read the battery
+guard past a gate the vendor cannot open, so a caller who read the tests
+green over ``send_action`` had no way to tell that on hardware the battery
+guard never fires.
+
+Two contracts are pinned as literals so a change to either fires this file,
+not a distant one:
+
+- On a fully-healthy driver except for ``_fsm_id`` (``_connected=True``,
+  ``_mode_machine`` from a real ``rt/lowstate`` decode, ``_battery`` from a
+  real ``rt/lf/bmsstate`` decode), ``send_action`` at 1.0% against a 15.0%
+  floor refuses with ``FSM id unknown``. The refusal does not mention
+  ``battery`` and does not carry the pack percentage -- the caller learns
+  about the FSM, not the pack, because the FSM gate fires first.
+- ``_fsm_id`` has exactly one assignment in ``strands_robots.drivers.g1`` and
+  it is the ``None`` initialiser. The check is an :mod:`ast` walk over the
+  shipped source, so a comment that mentions ``self._fsm_id = 500`` in a
+  docstring is not counted as a write. The single site's snippet is asserted
+  to contain ``None`` so a non-``None`` default that silently opened the gate
+  is graded.
+
+The day a real motion-switcher decoder writes ``_fsm_id``, both assertions
+fire -- correctly, because on the same day the battery floor becomes
+reachable and this test file must be replaced by one that grades the new
+reachability. The failure message names the second write site so the
+replacement is targeted, not exploratory.
--- a/changelog.d/2841-g1-battery-floor-reachability-pinned.md
+++ b/changelog.d/2841-g1-battery-floor-reachability-pinned.md
@@ -11,8 +11,8 @@ guard past a gate the vendor cannot open, so a caller who read the tests
 green over ``send_action`` had no way to tell that on hardware the battery
 guard never fires.
 
-Two contracts are pinned as literals so a change to either fires this file,
-not a distant one:
+Three contracts are pinned as literals so a change to any of them fires this
+file, not a distant one:
 
 - On a fully-healthy driver except for ``_fsm_id`` (``_connected=True``,
   ``_mode_machine`` from a real ``rt/lowstate`` decode, ``_battery`` from a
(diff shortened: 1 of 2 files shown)


yinsong1986 (dismissed):
## Summary

Test-only PR that pins two literal contracts around `G1Driver._check_motion_gates`: (1) on an otherwise-healthy driver (`_connected=True`, populated `_mode_machine` and `_battery`) with `_fsm_id` at its `None` initialiser, `send_action` refuses with the `FSM id unknown` message and never surfaces the battery floor, even at 1.0% against a 15.0% floor; and (2) an `ast` walk asserting `_fsm_id` has exactly one write site in `strands_robots.drivers.g1` whose assigned value is the `None` literal. Verified against the head: `g1.py` does carry exactly one `self._fsm_id` assignment (the annotated `None` initialiser in `__init__`), the FSM gate does fire ahead of the battery guard, and both new tests pass. The AST check correctly grades the value node rather than substring-matching the snippet, so a mutated non-`None` default with an unchanged `int | None` annotation is caught. The sunset condition (both assertions fire the day a real motion-switcher decoder lands, prompting replacement with a reachability-grading file) is stated in the module docstring and in both failure messages, so a future writer is steered to replace the file rather than silence it.

## What's good

- Both contracts are stated as literals (`_CRITICAL_PCT`, `_HEALTHY_FLOOR_PCT`) rather than derived from module constants, so a rename in the driver cannot silently change what the test compares.
- Changelog fragment follows `changelog.d/<pr>-<slug>.md` with a `### Added:` heading; no direct `CHANGELOG.md` edit.
- Test names describe behaviour, docstrings carry provenance (harness#361 PR-C, #2765), ASCII-only throughout, and the repo's meta-tests (host paths, test-name conventions, docstring cross-references) pass with the new file in the tree.
- `ruff check` and `ruff format --check` clean; `tests/drivers/` neighbours green (624 passed, 35 skipped in this environment).

yinsong1986 (approved):
## Summary

Test-only PR that pins three literal contracts around `G1Driver._check_motion_gates`: (1) reachability — on an otherwise-healthy driver (`_connected=True`, populated `_mode_machine` and `_battery`) with `_fsm_id` left at its `None` initialiser, `send_action` refuses with `FSM id unknown` and never surfaces the battery floor, even at 1.0% against a 15.0% floor; (2) producer count — an `ast` walk asserting `_fsm_id` has exactly one write site in `strands_robots.drivers.g1` whose assigned value is graded as an `ast.Constant(None)` node rather than by substring, so `self._fsm_id: int | None = 501` is caught despite the annotation carrying `None`; and (3) ordering — the FSM check precedes the battery-floor comparison inside the gate, so a reordering fails a cell named for ordering instead of masquerading as a reachability regression. Verified against the head source: `g1.py` carries exactly one `self._fsm_id` assignment (the annotated `None` initialiser), the `_fsm_id is None` refusal fires ahead of the `_battery_floor_pct` comparison, and the module-wide walk correctly excludes the runner class's `self._driver._fsm_id` reads (target must be a bare `self`). All three new tests pass, the neighbouring `tests/drivers/` suite is green with the file in the tree, and the repo's meta-tests (test-name conventions, host paths) pass. Relative to the earlier dismissed review at `ac6e1c28`, the head adds the third (ordering) contract, which closes the break-table row that (1) alone misreported.

## What's good

- Contracts stated as literals (`_CRITICAL_PCT`, `_HEALTHY_FLOOR_PCT`) rather than derived from module constants, so a driver-side rename cannot silently change what the test compares.
- The ordering contract identifies both checks by what they compare (`ast.unparse` of the `If` test) and asserts uniqueness of each, so a second floor comparison is a named failure rather than an ambiguity.
- Sunset condition is explicit in the module docstring and in both failure messages — a future motion-switcher writer is steered to replace the file, not silence it.
- Changelog fragment follows `changelog.d/<pr>-<slug>.md`; no direct `CHANGELOG.md` edit. ASCII-only throughout, behaviour-describing test names, provenance in docstrings (harness#361 PR-C, #2765).

PR #2842, opened 2026-08-27
fix(mesh): count a teleop frame the follower refused instead of reporting it applied

cagataycali:
## Problem

A host reports "I did not apply this" in two shapes, and only one of them is an exception.

`HardwareRobot.send_action` catches every exception and returns `{"status": "error"}`. Its own docstring gives the reason:

> Returns: Status dict (`success`/`error`) **so the teleop loop can count errors without exceptions tearing down the hot loop**.

A simulation host answers the same way for an action key it cannot resolve to an actuator, or for a `robot_name` that is not in the world. `InputReceiver._on_input` read only the exception, so the shape a host is *designed* to produce went uncounted — and both apply routes dropped the verdict before it could be read:

- `InputReceiver._default_apply` was `-> None` and discarded what `bus_access.write_action` had already carried out for it ("Returns: Whatever the driver's `send_action()` returns").
- The closure `TeleopMixin.start_teleop_receive` installs for a simulation host discarded it too.

Driving 30 frames through that closure into a MuJoCo `so101`, at the receiver's own nominal rate:

| action keys | joints reached | `frames_received` | `errors` | `rejected` | `slew_rejected` |
|---|---|---|---|---|---|
| `1` / `2` / `3` | `+0.0981` (driven) | 30 | 0 | 0 | 0 |
| `shoulder_pan.pos` / … | `+0.0123` (sagged) | 30 | **0** | 0 | 0 |

Both rows commanded `+1.5000` on three joints. The second is a real SO-101 leader's own key spelling reaching a follower that names its joints `1`..`6` — the mismatch `attach_teleop`'s `map_fn` exists to bridge on the local path, and for which the mesh path has no equivalent — so every frame was refused and the arm only sagged under gravity, `8.0x` short of the driven pose. **The two rows were identical in the report**: 30 frames at 49 Hz, zero errors, zero rejected, zero rate-dropped, zero slew-rejected. Nothing in `stats` separated a stream the follower applied in full from one it refused in full.

On hardware the same silence is reached by a much more ordinary route, because the conversion above is deliberate: a follower whose bus write fails on every frame returns 30 error envelopes and read as a healthy 49 Hz stream.

## Why the local path already had this

`TeleopMixin._teleop_loop` reads the envelope and counts it, and `_teleop_stats` writes the vocabulary down:

> soft: `send_action` returns `{"status": "error"}` -> errors += 1 AND frames += 1 (an unpowered follower gives errors == frames)

That comment ends by naming the outcome its derivation exists to refuse — *"0 frames, 0 errors and `success`: a silent no-op"* — which is what the receive path reported. And `_teleop_loop`'s own comment claims the two paths judge a leader frame identically, *"whether it reaches a follower over the network or on this host"*. That held for the per-joint slew bound and not for the outcome.

## Change

Both routes carry the verdict out, and `_on_input` counts it the way the local loop does, so `errors` and `frames_received` both advance. A follower that refuses everything now reads as `errors == frames_received` rather than as a clean session, and the refusal reason reaches the log naming the leader it came from.

`_refusal_text` reads every text block rather than indexing the first, because the envelope shape is the host's choice: an empty `content`, or a leading `json` block, must not turn a log line about a refused frame into an exception on the apply path — which would then be counted a second time, as the raising failure it is not.

## Deliberately unchanged, and pinned

- The frame still counts in `frames_received` — it was delivered and attempted, exactly as the local path counts it.
- The slew baseline still advances, identically to the local path, which merges it after counting the error. Changing it here would be a second decision about a safety bound, made inside an accounting change.
- `rejected` is untouched: that total names a guard on the *receiving* side which refused the frame and never applied it. The host's own verdict is not one of those guards.
- The check reads `== "error"`, not `!= "success"`. The envelope vocabulary in this package is wider than those two — `ok`, `running`, `idle` and `timeout` are all statuses it returns, and `_teleop_stats` adds `degraded` — and the local loop reads `== "error"` too.

## Verification

Pre-fix split (behavioural revert — the three behaviours undone, the new symbols kept): **8 failed / 22 passed**, headline `assert 0 == 3` on the error count. 5/5 in `TestAFrameTheHostRefusedIsCounted`, 3/3 in `TestBothApplyRoutesCarryTheVerdict`, and **0 of 10** in `TestWhatIsUnchangedEitherWay`, **0 of 2** in `TestARaisingApplyIsUnchanged`, **0 of 2** premise cells. A raw revert is a collection error instead, since the tests read `_refusal_text`.

Mutation table — 12 rows, all 11 non-control mutations detected, **11 distinct firing sets**, control clean, `collected` stable at 30 on every row, sources restored md5-identical:

| mutation | fires | in sibling suites |
|---|---|---|
| control | 0 | 0 |
| `_on_input` stops counting the envelope | 5 | 0 |
| `_default_apply` drops the verdict | 7 | 0 |
| the sim closure drops the verdict | 1 | 0 |
| all three reverted | 8 | 0 |
| `isinstance` guard dropped | 1 | **1** |
| counted as `rejected`, not `errors` | 6 | 0 |
| a refused frame is not a received frame | 2 | 0 |
| `_refusal_text` indexes the first block | 6 | 0 |
| `_refusal_text` yields empty for no reason | 4 | 0 |
| `!= "success"` instead of `== "error"` | 4 | 0 |
| the `stats` docstring paragraph removed | 2 | 0 |

`b1 | b2 | b3 == b4` exactly (8 ids), so each route is independently pinned. Dropping the `isinstance` guard also trips a pre-existing test, because a caller's own `apply_fn` returning `None` then raises — which is why that guard is load-bearing rather than defensive.

Paired gate on an identical 765-file selection (every `tests/mesh` file, every suite naming the receiver / the mixin / `send_action` / `write_action`, and every tree-walking guard), the new file excluded from both trees: **29407 collected and `28 failed, 29131 passed, 224 skipped, 24 errors` on each side**, 52 failing/erroring ids each, `comm` empty in both directions. All 52 classified: 45 need `awsiot`/`awscrt` (both `find_spec` `None`, declared in the `mesh-iot` extra), 4 in `test_recording_frame_loss_is_not_tolerated.py` pass 13/13 in isolation on both trees, 3 are lerobot-from-source `TypeError: encode() takes 1 positional argument` reproduced on the base alone.

`ruff check` and `ruff format --check` clean over 1649 files. `mypy strands_robots tests tests_integ`: 5 on the branch against 5 on a pristine worktree of the merge-base, normalized message sets identical in both directions, none attributable.

## Artifact

Both columns rendered from the simulation the frames were applied to, headlight lifted identically in all four captures. The renders are **byte-identical between the two trees** (`max|delta| = 0`) — nothing physical changes here; what changes is whether the report can tell the two streams apart.

the same two streams, before and after

### DIFF
--- a/changelog.d/2842-input-receive-counts-a-host-refusal.md
+++ b/changelog.d/2842-input-receive-counts-a-host-refusal.md
@@ -1,0 +1,50 @@
+### Fixed: a teleop frame the follower refused is counted, not reported applied
+
+A host reports "I did not apply this" in two shapes, and only one of them is an
+exception. `HardwareRobot.send_action` catches every exception and returns
+`{"status": "error"}` - its docstring gives the reason, "so the teleop loop can
+count errors without exceptions tearing down the hot loop" - and a simulation
+host answers the same way for an action key it cannot resolve to an actuator, or
+for a `robot_name` that is not in the world. `InputReceiver._on_input` read only
+the exception, so the shape a host is *designed* to produce went uncounted. Both
+apply routes dropped the verdict on the way out:
+`InputReceiver._default_apply` discarded what
+`bus_access.write_action` had carried out for it ("Returns: Whatever the driver's
+`send_action()` returns"), and the closure `start_teleop_receive` installs for a
+simulation host discarded it too.
+
+Driving 30 frames into a MuJoCo `so101` through that closure, at the receiver's
+own nominal rate:
+
+| action keys | joints reached | frames_received | errors |
+|---|---|---|---|
+| `1` / `2` / `3` | 0.098 (driven) | 30 | 0 |
+| `shoulder_pan.pos` / ... | 0.012 (sagged) | 30 | 0 |
+
+Both rows commanded 1.5. The second is a real SO-101 leader's own key spelling
+reaching a follower that names its joints `1`..`6` - the mismatch
+`attach_teleop`'s `map_fn` exists to bridge on the local path, for which the mesh
+path has no equivalent - so every frame was refused and the arm only sagged under
+gravity. The two rows were identical in the report: 30 frames at 49 Hz, zero
+errors, zero rejected, zero rate-dropped, zero slew-rejected. Nothing in `stats`
+separated a stream the follower applied in full from one it refused in full.
+
+The local teleop loop already read that envelope, and
+`TeleopMixin._teleop_stats` writes the vocabulary down: "soft: `send_action`
+returns `{"status": "error"}` -> errors += 1 AND frames += 1 (an unpowered
+follower gives errors == frames)". Its comment ends by naming the outcome the
+derivation exists to refuse - "0 frames, 0 errors and 'success': a silent no-op"
+- which is what the receive path reported. `_teleop_loop`'s own comment claims
+the two paths judge a leader frame identically, "whether it reaches a follower
+over the network or on this host"; that held for the per-joint slew bound and not
+for the outcome. Both routes now carry the verdict out and `_on_input` counts it
+the way the local loop does, so a follower that refuses everything reads as
+`errors == frames_received`.
+
+Deliberately unchanged, and pinned: the frame still counts in `frames_received`
+(it was delivered and attempted, exactly as the local path counts it), the slew
+baseline still advances, and `rejected` is untouched - that total names a guard on
+the receiving side which refused the frame and never applied it. The check reads
+`== "error"` rather than `!= "success"`, because the envelope vocabulary here is
+wider than those two (`ok`, `running`, `idle`, `timeout` are all statuses this
+package returns) and the local loop reads `== "error"` too.
--- a/strands_robots/mesh/input.py
+++ b/strands_robots/mesh/input.py
@@ -138,6 +138,30 @@ def _input_audit_every() -> int:
     return val if val > 0 else 0
 
 
+def _refusal_text(envelope: dict[str, Any]) -> str:
+    """Read the human-readable reason out of a host's error envelope.
+
+    Reads every text block rather than indexing the first, because the shape is
+    the host's choice: a ``content`` list that is empty, or whose leading block
+    carries ``json`` instead of ``text``, must not turn a log line about a
+    refused frame into an exception on the apply path -- that would be counted a
+    second time, as the raising failure it is not.
+
+    Args:
+        envelope: A tool-result envelope whose ``status`` is ``"error"``.
+
+    Returns:
+        The joined text blocks, or ``"no reason given"`` when the envelope
+        carries none.
+    """
+    texts = [
+        block["text"]
+        for block in envelope.get("content", [])
+        if isinstance(block, dict) and isinstance(block.get("text"), str) and block["text"]
+    ]
+    return "; ".join(texts) if texts else "no reason given"
+
+
 class InputPublisher:
     """Publishes teleoperator actions to the mesh at a fixed rate.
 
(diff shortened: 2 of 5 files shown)


yinsong1986 (dismissed):
## Summary

The mesh teleop receive path (`InputReceiver._on_input`) only counted an apply failure when the apply *raised*, but `HardwareRobot.send_action` (by design) and the simulation hosts report a refusal as a returned `{"status": "error"}` envelope -- so a stream the follower refused in full reported the same stats as a healthy one. This PR reads the apply's return value, counts an error envelope into `errors` (with the same rate-limited log budget as the raising shape), and threads the verdict through both apply routes: `_default_apply` now returns what `write_action` already carried out, and the sim closure `start_teleop_receive` installs returns `host.send_action(...)` instead of dropping it. The diff matches the title and description.

## What's good

- The `apply_fn` signature widening (`-> None` to `-> Any`) is backward compatible, and a `None`-returning custom apply is explicitly pinned as not-an-error, so no existing caller's behaviour changes.
- The check reads `== "error"` rather than `!= "success"`, with a parametrized test over the package's wider status vocabulary (`ok`, `running`, `idle`, `timeout`) pinning that narrowing -- matching how `TeleopMixin._teleop_loop` reads the same envelope, which a structural test also grades so the two paths cannot silently diverge again.
- Deliberately-unchanged boundaries (`frames_received` still advances, slew baseline still merges, `rejected` untouched) are each pinned by a test with the reasoning recorded, per AGENTS.md's regression-pinning convention.
- `_refusal_text` reads every text block defensively rather than indexing `content[0]`, so a host's envelope shape cannot turn a log line into a second counted error on the hot path -- and the edge shapes are tested.
- All 68 tests in the two touched suites pass against the head SHA; ruff check and format are clean on all four files.

Note: no `changelog.d/2842-*.md` fragment is in the diff; the `changelog-fragment` workflow will surface that, so it is left to CI rather than raised as a blocker here.

yinsong1986 (approved):
## Summary

`InputReceiver._on_input` counted an apply failure only when the apply *raised*, while `HardwareRobot.send_action` (by documented design) and the simulation hosts report a refusal as a returned `{"status": "error"}` envelope -- so a stream the follower refused in full was indistinguishable in `stats` from one it applied in full. This PR threads the verdict through both apply routes (`_default_apply` now returns what `bus_access.write_action` already carried out; the sim closure in `start_teleop_receive` returns `host.send_action(...)` instead of dropping it) and counts an error envelope the way the local teleop loop does: `errors` and `frames_received` both advance, so a follower refusing everything reads as `errors == frames_received`. The diff matches the title and description.

## What's good

- The `== "error"` read (not `!= "success"`) matches `TeleopMixin._teleop_loop`, is parametrized over the package's wider status vocabulary, and the local-loop side of the parity is graded structurally so the two paths cannot silently diverge again.
- Deliberately-unchanged boundaries -- `frames_received` still advances, the slew baseline still merges, `rejected` untouched, a `None`-returning custom `apply_fn` is not an error -- are each pinned with the reasoning recorded, per AGENTS.md's regression-pinning convention.
- `_refusal_text` reads every text block defensively rather than indexing `content[0]`, so a host's envelope shape cannot turn a log line into a second counted failure on the hot path; the edge shapes are tested, and the log budget is shared with the raising path so neither shape can spam at 50 Hz.
- The `apply_fn` signature widening (`-> None` to `-> Any`) is backward compatible, and the premise test (`TestARealHostRefusesThisWay`) grades that a shipped simulation host actually answers in this envelope shape, gated behind `importorskip`.
- Changelog fragment present (`changelog.d/2842-input-receive-counts-a-host-refusal.md`), all strings ASCII, test names describe behaviour, no host paths.

PR #2843, opened 2026-08-27
fix(inference): hold the wire lock across the connect replay

cagataycali:
## Problem

`RemotePolicy` serialises its single WebSocket with one lock, and the wire helper
states the precondition in the section header directly above it:

```python
# -- wire helpers (call while holding ``self._lock``) ---------------------

def _request(self, message: dict[str, Any]) -> dict[str, Any]:
```

Four of `_request`'s seven call sites held it. The three inside `_connect` did
not, and those three run *after* `self._ws` is live: the handshake read, then the
replay of any state keys, control frequency and pending reset set before the
connection existed. `_get_actions_blocking` even calls `_ensure_connected()`
*outside* the lock and takes it only for its own request, so the connect sequence
ran its whole replay unlocked on an already-usable connection.

| `_request` call site | lock held |
|---|---|
| `_connect` (state-keys replay) | no |
| `_connect` (control-frequency replay) | no |
| `_connect` (pending-reset replay) | no |
| `set_robot_state_keys` | yes |
| `set_control_frequency` | yes |
| `reset` | yes |
| `_get_actions_blocking` | yes |

## Measured consequence

A server parked inside the replay, a client whose config was set before
connecting, and a second thread on the same policy:

| | before | after |
|---|---|---|
| `self._ws` during the replay | live | live |
| second thread's `reset(seed=7)` | `ConcurrencyError: cannot call recv while another thread is already running recv or recv_streaming` | returns |
| second thread while the replay is open | raised | waiting on the lock |
| connections opened by 8 racing first-callers | up to 8 | 1 |

The report names the transport's internals, out of a call that never mentions
threads, and nothing a caller passed is wrong. The same window is reachable from
`set_robot_state_keys` and from `get_actions`.

## Reachability

Two threads on one policy is the ordinary case here rather than a contrived one.
Every policy coroutine resolves through `_async_utils`' reused worker thread, and
the async-RTC path in `simulation.policy_runner` submits prefetch inference to
its own `rtc-prefetch` worker while the rollout thread carries on stepping - so a
first connect and another call on the same policy land on different threads by
design.

## Why nothing caught it

`tests/inference/` does use threads, but on the *server*: every existing test
runs `PolicyServer` in a background thread and drives the client from one thread
throughout. No test drives the client concurrently, so the suite structurally
could not observe an overlap on the client's own connection.

## Change

`_ensure_connected` holds `self._lock` across `_connect`, with the unlocked fast
path kept and the state re-checked under the lock. None of its three callers
holds the lock, so nothing deadlocks; a concurrent wire user waits for the replay
and then proceeds, which is what the lock was there to do.

Reusing that lock rather than adding a connect-only one is the point, and it is
the row below that measures it: a dedicated lock makes connects mutually
exclusive and still leaves them racing `reset` and `get_actions`.

The precondition is graded at runtime rather than by reading the source. A
witness lock records whether it was held when `_request` ran, so the coverage
`_connect` inherits from its caller counts - a lexical scan would report those
three calls as unlocked either way.

## Mutation table

Every row against the new file (`mine`) and the rest of `tests/inference/` with
it deselected (`sib`); `coll` is the collected count, stable throughout, so no
row hides a deselection.

| id | mutation | mine | coll | sib | firing |
|---|---|---|---|---|---|
| b0 | control, no mutation | 0 | 11 | 0 | - |
| b1 | fix reverted (the shipped defect) | 6 | 11 | 0 | serialisation + precondition + racing |
| b2 | lock held, inner re-check dropped | 1 | 11 | 0 | racing only |
| b3 | lock taken then released before connect | 5-6 | 11 | 0 | serialisation + precondition (+ racing) |
| b4 | unlocked fast path dropped, always acquire | 0 | 11 | 0 | - |
| b5 | a separate `_connect_lock`, not the wire lock | 5 | 11 | 0 | serialisation + precondition |

`b2 | b3 == b1` and b2/b3 are disjoint: the double check and the lock coverage
are independently pinned.

b4 is a measured no-op with a reason - always acquiring is behaviourally
identical to the fast path, only slower - and is reported rather than dropped.

b5 fires b3's set because a connect-only lock serialises connects (so the racing
cell passes) and not the other wire users. That is the distinction between "a
lock" and "the lock", and it is why the fix reuses `self._lock`.

The racing cell detects probabilistically pre-fix: the window is the connect
handshake itself, so staggered starts let the GIL serialise the callers by luck.
With a barrier so all eight arrive together it detects 6 of 8 runs pre-fix and
passes 8 of 8 post-fix; the barrier is in the test for that reason.

## Pre-fix split

6 failed / 5 passed, with roles kept honest:

| class | role | pre-fix |
|---|---|---|
| `TestAConcurrentWireUserWaitsForTheConnectReplay` | regression | 4 of 4 fail |
| `TestEveryWireRequestRunsUnderTheLock` | regression + non-vacuity | 1 of 2 fails |
| `TestTwoRacingFirstCallersOpenOneConnection` | regression | 1 of 1 fails |
| `TestTheConnectReplayIsAWireUser` | premise | 0 of 2 |
| `TestWhatIsUnchanged` | control | 0 of 2 |

The one cell in `TestEveryWireRequestRunsUnderTheLock` that passes both ways is
its non-vacuity half: it asserts the replay's own requests are in the graded set,
which holds either way and is what stops the precondition test passing over a
narrowed drive.

## Gate

* `ruff check` + `ruff format --check` over `strands_robots tests tests_integ`: clean, 1649 files.
* `mypy strands_robots tests tests_integ`: 24 errors in 9 files, the same count as the merge base, 0 in either changed file.
* Paired run over an identical 543-file selection (all of `tests/inference/` plus every guard that walks the tree, parses source, or reads docstrings and `changelog.d`), the new file excluded from both trees, every path verified present on both: identical **24888 collected**, **20 failed on each**, failing-id sets identical, `comm` empty in both directions, distinct rootdirs. The 20 are pre-existing and environmental - 17 need `awsiot`/`awscrt` (declared in the `mesh-iot` extra, absent locally) and 3 are the lerobot-from-source drift in `tests/training/test_lerobot*`.
* `tests/inference/` whole: 135 passed.

The first branch-side run of that selection reported 23, and the three extra were
`tests/test_codeql_query_filters.py`'s derived census of non-reraising
`except BaseException` handlers - the new file added two. They are narrowed to
`except Exception` in the third commit (the failure under test is a
`ConcurrencyError`, a `RuntimeError`), which is what the 20-vs-20 run above
measures.

No visual artifact: this changes when a lock is held and what a concurrent caller
is told, not a policy, simulation, rendering, recording or asset.

### DIFF
--- a/changelog.d/2843-connect-replay-holds-the-wire-lock.md
+++ b/changelog.d/2843-connect-replay-holds-the-wire-lock.md
@@ -1,0 +1,39 @@
+### Fixed: the remote-policy connect replay holds the wire lock the other callers hold
+
+`RemotePolicy` serialises its WebSocket with one lock, and its wire helper says so
+in the section header above it: "call while holding `self._lock`". Four of
+`_request`'s seven call sites did. The three inside `_connect` did not, and those
+three run *after* `self._ws` is live - the handshake read, then the replay of any
+state keys, control frequency and pending reset that were set before the
+connection existed. The connect sequence is the widest wire user in the class and
+it was the one outside the lock.
+
+A second thread reaching the wire in that window overlapped the read, and
+`websockets` refuses an overlapping read. Measured with a server parked mid-replay
+and a client whose config was set before connecting: `reset(seed=7)` raised
+`ConcurrencyError: cannot call recv while another thread is already running recv
+or recv_streaming` - a report naming the transport's internals rather than
+anything the caller passed, out of a call that never mentions threads. The same
+window is reachable from `set_robot_state_keys` and from `get_actions`.
+
+Two threads on one policy is the ordinary case rather than a contrived one. Every
+policy coroutine resolves through `_async_utils`' reused worker thread, and the
+async-RTC path in `simulation.policy_runner` submits prefetch inference to its own
+`rtc-prefetch` worker while the rollout thread carries on stepping, so a
+first connect and another call on the same policy land on different threads by
+design.
+
+`_ensure_connected` now holds `self._lock` across `_connect`. The unlocked fast
+path stays, with the state re-checked under the lock, so two racing first-callers
+open one connection rather than two; none of its three callers holds the lock, so
+nothing deadlocks. A concurrent wire user waits for the replay and then proceeds,
+which is what the lock was there to do. Reusing that lock rather than adding a
+connect-only one is the point: a dedicated lock would make connects mutually
+exclusive and still leave them racing `reset` and `get_actions`.
+
+The precondition is now graded at runtime rather than by reading the source. A
+witness lock records whether it was held when `_request` ran, so the coverage
+`_connect` inherits from its caller counts, where a lexical scan would report
+those three calls as unlocked either way. The pre-existing suite could not see
+this: its threads run the *server*, and the client is driven from one thread
+throughout.
--- a/strands_robots/inference/client.py
+++ b/strands_robots/inference/client.py
@@ -209,8 +209,37 @@ class RemotePolicy(Policy):
             self._reset_pending = False
 
     def _ensure_connected(self) -> None:
-        if self._ws is None:
-            self._connect()
+        """Open the connection if it is not open yet, serialised with the other wire users.
+
+        Holds ``self._lock`` across :meth:`_connect` because the connect
+        sequence is itself a wire user, and the widest one: it reads the
+        handshake and then replays up to three pending-config requests through
+        :meth:`_request`, every one of them after ``self._ws`` is already live.
+        Left unlocked, those sends interleave with a concurrent :meth:`reset`,
+        :meth:`set_robot_state_keys` or :meth:`get_actions` on the same
+        connection, and ``websockets`` refuses the overlapping read with a
+        ``ConcurrencyError`` naming its own internals rather than anything the
+        caller passed - a report no ``RemotePolicy`` caller can act on.
+
+        Two threads on one policy is the ordinary case here, not a contrived
+        one: every policy coroutine resolves through
+        :mod:`strands_robots._async_utils`' reused worker thread, and the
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Closes the one unlocked window on `RemotePolicy`'s single WebSocket: the connect sequence replays up to three pending-config requests through `_request` after `self._ws` is live, and it was the only `_request` caller running outside `self._lock`. `_ensure_connected` now holds the wire lock across `_connect`, keeping the unlocked fast path with a re-check under the lock so two racing first-callers open one connection. Reusing the wire lock (rather than a connect-only lock) is the correct choice and the mutation table's b5 row demonstrates why. Independently verified: the 11 new tests pass at head and 5-6 of 11 fail with the fix reverted (matching the description's note that the 8-way racing cell detects probabilistically pre-fix); the rest of `tests/inference/` passes unchanged; no re-entrant acquisition path exists (none of `_ensure_connected`'s three callers holds the lock, and `_request` never takes it itself), and `close()` just waits out a parked replay bounded by the request timeout. The prior CodeQL `BaseException` findings are resolved by the third commit's narrowing to `except Exception`. Internal-only change - no public API, wire-format, or persisted-schema surface moves.

## What's good

- The regression tests park the server mid-replay via an event rather than sleeping, so the race window is held open deterministically; the witness-lock grading covers the lock coverage `_connect` inherits from its caller, which a lexical scan cannot see.
- Changelog fragment follows the `changelog.d/<pr>-<slug>.md` convention; test names describe behaviour, not provenance; premise and control classes keep the pre-fix split honest.
- The mutation table distinguishes "a lock" from "the lock" (b5) and pins the double-check independently of the serialisation (b2/b3 disjoint).

PR #2844, opened 2026-08-27
fix(simulation): fill the undriven state columns of a synchronized rollout too

cagataycali:
## Problem

`start_recording` declares `observation.state` over **every** robot in the scene, prefixing each column with its robot's name (`alice__shoulder_pan`). A rollout's recording hook supplies only the observation of the robots it drives, so every column belonging to another robot is absent from the frame and `DatasetRecorder.add_frame` writes it as `0.0` — in the same column, with the same dtype, as a measurement, for every frame, under `status="success"`. Nothing downstream can tell that zero from a real reading.

`undriven_robot_state` exists to fill those columns from the engine instead, and the three single-policy recording hooks go through it. `run_multi_policy` does not. Its merge loop iterates the keys of `policies`:

```python
for rname in policies:
    if multi_robot:
        for k, v in per_robot_obs[rname].items():
            merged_obs[f"{rname}__{k}"] = v
```

and `SimEngine.run_multi_policy` requires each key to name a robot in the scene *without* requiring the keys to cover it — the constraint runs one way only. Meanwhile `multi_robot = len(self._world.robots) > 1` is computed over the whole scene, so the schema is prefixed for a bystander whose columns the merge never produces. A synchronized call that drives a subset therefore left every other robot to the fill, exactly as a single-policy hook did before the helper existed.

The helper's own docstring credited the synchronized loop with already having the guarantee, on the strength of its merging every robot *it drives*.

## Measured

Two-robot MuJoCo scene, `bob` parked at `shoulder_pan 0.9` / `elbow -0.7`, `policies={"alice": MockPolicy()}`, read back through `LeRobotDataset`:

| | before | after |
|---|---|---|
| declared columns | `alice__shoulder_pan, alice__elbow, bob__shoulder_pan, bob__elbow` | same |
| reported | `status="success"`, 6 frames, MP4 written | same |
| `bob__shoulder_pan` | `0.0` every frame | `0.9000` every frame |
| `bob__elbow` | `0.0` every frame | traces 2.52 rad of real gravity swing |
| largest reading the fill replaced | 2.2629 rad | — |
| last frame vs the engine's own reading | disagrees | agrees to within 1e-3 rad |
| driven robot's columns | — | bit-identical to before |

## Change

`undriven_robot_state` now takes the collection of robots a frame drives rather than a single name; both `run_multi_policy` implementations (MuJoCo and Isaac) merge its result **before** their own driven columns, so driven keys still win any collision. A bare `str` is refused: a string is iterable per character, which would turn the skip into a substring test and drop a robot named `ali` as though `alice` drove it — the very fill the helper exists to avoid.

The **action** half of the same frame is unchanged and deliberately so: no command was issued to a robot this call does not drive, so its action columns have no truthful value to write. Pinned, so widening the fill there is a visible decision rather than a side effect.

## Tests

Added beside the single-policy cells that pin the same property, so the parity is visible. Behavioural pre-fix split **4 failed / 21 passed**, with 0 failures in every premise, control, boundary and pre-existing class. The whole defect fires **0** of 233 pre-existing cells in the surrounding `run_multi_policy` and recording suites — nothing else could see it.

The structural survey that held each recording entry point to the shared helper covered only the single-policy hooks. It is now derived from the backend registry `create_simulation` itself resolves, and covers `run_multi_policy` for each backend that implements one, so **Isaac's half of the fix is graded on a machine that cannot run Isaac Sim**.

| mutation | fires | collected | note |
|---|---|---|---|
| control | 0 | 25 | |
| both synchronized fills dropped | 4 | 25 | the defect |
| MuJoCo fill only | 3 | 25 | 2 behavioural + structural[mujoco] |
| Isaac fill only | 1 | 25 | structural[isaac] alone |
| bare-`str` refusal dropped | 1 | 25 | |
| skip reverted to equality against the collection | 3 | 25 | |
| fill merged after the driven loop | 0 | 25 | measured no-op: the two key sets are disjoint by construction |
| Isaac fill dropped **and** the survey narrowed to the hooks | 0 | 22 | Isaac's defect is silent without the derived survey |
| the survey narrowed, no defect | 0 | 22 | the narrowing alone fires nothing |
| both fills dropped **and** the survey narrowed | 2 | 22 | only the MuJoCo-reachable behavioural cells |

`MuJoCo fill only` and `Isaac fill only` are disjoint and their union is exactly the full defect, so each backend's half is independently pinned.

## Gate

- `ruff check` / `ruff format --check` clean over 1648 files.
- `mypy` **24 == 24** against a pristine base worktree, normalized message sets byte-identical both directions, **0** attributable to the changed files.
- Paired run over an identical 600-file selection (all of `tests/simulation/` plus every tree-walking, source-parsing and docstring guard), the changed test file excluded from both trees: identical **20713 collected** and `5 failed, 20461 passed, 247 skipped` on each, failing-ID sets identical with `comm` empty both directions. All 5 pass **29/29 in isolation** — a large-selection module-caching artifact in `test_recording_frame_loss_is_not_tolerated.py` plus one wall-clock pacer, present on both trees.

## Artifact

One synchronized episode recorded twice from an identical scene, MuJoCo headless (EGL). Every number in the figure is asserted from the two captures before it is drawn.

undriven state columns of a synchronized rollout

### DIFF
--- a/changelog.d/2844-synchronized-rollout-fills-the-undriven-columns.md
+++ b/changelog.d/2844-synchronized-rollout-fills-the-undriven-columns.md
@@ -1,0 +1,46 @@
+### Fixed: a synchronized rollout that drives a subset of the scene records the rest as measurements
+
+`start_recording` declares `observation.state` over EVERY robot in the scene,
+prefixing each column with its robot's name (`alice__shoulder_pan`). A rollout's
+recording hook supplies only the observation of the robots it drives, so every
+column belonging to another robot is absent from the frame and
+`DatasetRecorder.add_frame` writes it as `0.0` - in the same column, with the
+same dtype, as a measurement, for every frame of the episode, under
+`status="success"`. Nothing downstream can tell that zero from a real reading: a
+policy trained on the dataset learns the other robot is permanently at its zero
+pose. `undriven_robot_state` exists to fill those columns from the engine
+instead, and the three single-policy recording hooks went through it.
+
+`run_multi_policy` did not. Its merge loop iterates the keys of `policies`, and
+`SimEngine.run_multi_policy` requires each key to name a robot in the scene
+without requiring the keys to cover it - the constraint runs one way only. So a
+synchronized call that drives a subset of the scene left every other robot to
+the fill, exactly as a single-policy hook did before the helper existed. The
+helper's own docstring credited the synchronized loop with already having the
+guarantee, on the strength of its merging every robot *it drives*.
+
+Measured on a two-robot MuJoCo scene with `bob` parked at `shoulder_pan 0.9`,
+`elbow -0.7` and `policies={"alice": MockPolicy()}`: the schema declared
+`bob__shoulder_pan` and `bob__elbow`, the rollout reported `status="success"`
+with 6 frames and a written MP4, and both of `bob`'s columns were `0.0` in every
+frame - the largest reading the fill replaced being 2.2629 rad. Reading the same
+dataset back after the fix, `bob__shoulder_pan` holds `0.9` across the episode
+and `bob__elbow` traces the 2.52 rad of gravity swing the joint really performed,
+agreeing with the engine's own reading to within 1e-3 rad. The driven robot's
+columns are bit-identical across the two recordings.
+
+The helper now takes the collection of robots a frame drives rather than a single
+name, and both `run_multi_policy` implementations (MuJoCo and Isaac) merge its
+result before their own driven columns, so driven keys still win any collision.
+A bare `str` is refused rather than accepted: a string is iterable per character,
+which would turn the skip into a substring test and drop a robot named `ali` as
+though `alice` drove it - the very fill the helper exists to avoid.
+
+The *action* half of the same frame is unchanged and deliberately so. No command
+was issued to a robot this call does not drive, so its action columns have no
+truthful value to write, and they are left where #1715 left them. The structural
+survey that held each recording entry point to the shared helper covered only the
+single-policy hooks; it is now derived from the backend registry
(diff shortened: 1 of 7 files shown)


yinsong1986 (approved):
## Summary

Closes the synchronized half of the undriven-state-column defect: `run_multi_policy` in both backends that implement it (MuJoCo, Isaac) now merges `undriven_robot_state` into the recorded frame before the driven columns, so a rollout that drives a subset of the scene records the bystanders' declared `observation.state` columns as engine measurements instead of `add_frame`'s silent `0.0` fill. The helper's signature widens from one name to a collection with a bare-`str` refusal (a string would turn the skip into a per-character membership test), and all four in-tree call sites are updated to `(name,)` tuples. Verified independently: the full test file passes 25/25 on `9c621735`, and dropping the MuJoCo fill line fires exactly the 3 cells the PR's mutation table reports, so the behavioural pinning is real. `multi_robot` is scene-scoped in both loops, matching the fill's prefixed keys; the fill-first merge order keeps driven keys winning collisions and is itself pinned by the precedence test.

## What's good

- The `TypeError` on a bare `str` carries an actionable remedy and a matching `Raises:` entry (AGENTS.md > "A `Raises:` block names every refusal the function itself makes").
- Every recording test names its own dataset root (`root=str(root)`), per AGENTS.md rule 15.
- The structural survey is derived from the backend registry rather than a hand-kept list, so Isaac's half is graded on a machine without Isaac Sim, and a future backend is covered on arrival.
- The action-half boundary is pinned (`test_the_undriven_robots_action_columns_are_left_to_the_fill`), so widening the fill to action columns is a visible decision.
- Changelog fragment present; additions are ASCII-clean; no host paths in tests.

PR #2845, opened 2026-08-27
fix(isaac): refuse a non-finite mesh scale wherever it is carried, not only where it is measured

cagataycali:
## The defect

`mesh_aabb` refuses a `<mesh scale=...>` whose components are not all finite (#2838), because it measures `vertex * scale` and `min`/`max` order a NaN as neither smaller nor larger than anything.

`load_mjcf_scene_objects` reaches that measurement on **one** of its three branches - the `elif mesh_path is not None` fallback for a body with no collidable analytic geom. It carries the scale onto the `SceneObject` on **all** of them, and the Isaac realization applies it to the visual prim's xform.

So a body declaring both a collidable geom and a visual mesh took its bound from the geom, never measured, and reported the scale unchecked:

| `<mesh scale>` | body has a collidable geom | MuJoCo | loader before | loader after |
|---|---|---|---|---|
| `1 1 1` | no | compiles | success, `(1.0, 1.0, 1.0)` | unchanged |
| `1 1 1` | yes | compiles | success, `(1.0, 1.0, 1.0)` | unchanged |
| `nan 1 1` | no | compiles | refused (#2838) | refused |
| `nan 1 1` | **yes** | compiles | **success, `mesh_scale=(nan, 1.0, 1.0)`** | refused |
| `inf 1 1` | **yes** | compiles | **success, `mesh_scale=(inf, 1.0, 1.0)`** | refused |

In the leaking rows `position`, `size` and `offset` were all finite and correct, because the collidable geom supplied them - `size=(0.2, 0.2, 0.1)`, the box. Every field a consumer would screen for a non-finite value was healthy, and the only bad one was the field nothing measures.

**53 of the 63** mesh-carrying bodies in the shipped registry are in that shape (`mesh_aabb` is called for 10 of them), so the measurement covered the minority.

## Why the scale belongs to the same test as its two neighbours

The realization authors the visual prim with one call:

```python
self._author_local_xform(
    stage.GetPrimAtPath(visual_path),
    translate=local_pos,      # from mesh_pos
    orient_wxyz=obj.mesh_quat,
    scale=obj.mesh_scale,
)
```

Two of those three arguments are already refused when they are not finite. Measured on the same body with a collidable box:

| xform argument | from field | non-finite spelling | before |
|---|---|---|---|
| `translate=` | `mesh_pos` | `<geom mesh pos="nan 0 0">` | refused |
| `orient_wxyz=` | `mesh_quat` | `<geom mesh quat="nan 1 0 0">` | refused |
| `scale=` | `mesh_scale` | `<mesh scale="nan 1 1">` | **success, `(nan, 1.0, 1.0)`** |

#2838 pinned the boundary with the reason "refusing it would be a claim about a consumer this reader does not own". That reason covers the *measurement*, and the reader already refuses two arguments of that consumer's own call for exactly this property - so the third is held to the same test rather than to the reach of the measurement. #2838's class is repaired rather than deleted: the claim it gets right (the bound comes from the collidable geom) is kept and strengthened to a finite non-unit scale, where the mesh would measure 0.6 m and the box is what is reported.

## The change

`refuse_non_finite_scale` becomes the single owner of the scale finiteness test in `mesh_assets`; `mesh_aabb` delegates to it, and the loader reaches it where the scale is read - beside the missing-file contract, for the same reason that check sits there: an asset declaration a body reaches has to be usable. The refusal's wording gains the xform consequence alongside the sizing one.

## What is unchanged

A finite scale is untouched on both body shapes, including the negative, zero and non-unit ones. Across the shipped registry: **1746 mesh assets declared, 0 non-finite, 250 finite non-unit** - the population the guard must not touch. After the change all **60** resolvable registry models still load, 0 failures, every reported `mesh_scale` finite. The defect is latent on shipped assets; what it changes is which declarations a scene may be built from.

## Why nothing caught it

`_MESH_ONLY` and `_WITH_COLLISION` templates both existed, and the with-collision one was driven only to assert that the bound came from the geom and that the scale rode along - the second of which is the defect, pinned as intended behaviour.

## Tests

New `tests/simulation/isaac/test_mesh_asset_scale_is_refused_wherever_it_is_carried.py`, 41 cases: the refusal on both templates for both non-finite values on every axis; that the two siblings of the same xform call were already refused; that a finite scale is unchanged on both templates; the premises (MuJoCo compiles the model, the collidable body really takes its bound from its geom, the mesh-only body really measures through the scale); and that the finiteness test has one owner both consumers reach, ahead of both the measurement and the object.

Pre-fix behavioural split (the owner kept importable, the loader's call site undone): **13 failed / 68 passed** - 11 regression (the 9 with-collision refusal cells plus the 2 message cells; the 9 mesh-only cells correctly pass, #2838 already covered them) and 2 structural, with **0** in the premise, control, sibling-xform and repaired classes. A raw revert is a collection error, since the tests import the new owner.

### Mutation table

`mine` is the two scale suites, `sib` is the seven pre-existing isaac mesh and finiteness suites.

| mutation | mine | collected | sib |
|---|---|---|---|
| control | 0 | 81 | 0 |
| loader call site removed (the defect) | 13 | 81 | 0 |
| `mesh_aabb` delegation removed | 16 | 81 | 0 |
| both consumers stop reaching the owner | 43 | 81 | 0 |
| guard moved to the measurement branch | 11 | 81 | 0 |
| owner accepts if **any** component is finite | 37 | 81 | 0 |
| owner checks the first axis only | 20 | 81 | 0 |
| message drops the visual-xform clause | 1 | 81 | 0 |
| owner returns instead of raising | 41 | 81 | 0 |
| two copies of the test instead of one owner | 9 | 81 | 0 |

Ten mutations, nine firing, **nine distinct firing sets**, control clean, `collected` stable at 81 on every row, source restored byte-identical.

The two per-consumer reverts are **disjoint** and their union is a strict subset of the both-consumers revert (13 + 16 = 29 against 43). The 14 cells that need both gone are the mesh-only rows, which either guard alone still catches - the redundancy the belt-and-braces design intends - while the with-collision rows are reachable only through the loader.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1649 files).
- `mypy strands_robots tests tests_integ`: **branch-only message set empty**, 0 attributable. (Raw counts differ only by the cold-cache `examples/isaac_gs/` artifact in the base worktree.)
- Paired 288-file selection - all of `tests/simulation/isaac/` plus every suite walking the tree, parsing source or reading docstrings, `Args:`/`Raises:` and `__all__` - identical on both trees with the new file excluded from each: **0 failures on either side**, both failing-ID sets empty, distinct rootdirs. Collected 14082 against 14081; the `+1` localises to `test_mesh_asset_scale_must_be_finite.py` (39 -> 40), where one test was split into two.
- `tests/simulation/isaac/` alone: **1921 passed, 5 skipped**.

No visual artifact: the consumer this value reaches is Isaac Sim's xform authoring, which needs the `sim-isaac` extra, and MuJoCo compiles the declaration with only a warning rather than rendering a refusal. The measured refusal and corpus tables above are the evidence.

### DIFF
--- a/changelog.d/2845-mesh-scale-refused-wherever-it-is-carried.md
+++ b/changelog.d/2845-mesh-scale-refused-wherever-it-is-carried.md
@@ -1,0 +1,27 @@
+### Fixed: a non-finite MJCF mesh scale is refused on every body that carries it
+
+`mesh_aabb` refuses a `<mesh scale=...>` whose components are not all finite,
+because it measures `vertex * scale` and `min`/`max` order a NaN as neither
+smaller nor larger than anything. `load_mjcf_scene_objects` reaches that
+measurement only on the branch where no collidable analytic geom supplied the
+bound - yet it carries the scale onto the `SceneObject` on every branch, and the
+Isaac realization applies it to the visual prim's xform.
+
+So a body declaring both a collidable geom and a visual mesh took its bound from
+the geom, never measured, and reported `mesh_scale=(nan, 1.0, 1.0)` under
+`success`. `position`, `size` and `offset` all stayed finite and correct, because
+the collidable geom supplied them, so every field a consumer would screen for a
+non-finite value was healthy. 53 of the 63 mesh-carrying bodies in the shipped
+registry are in that shape.
+
+The value lands in one `_author_local_xform(translate=..., orient_wxyz=...,
+scale=...)` call whose other two arguments come from `mesh_pos` and `mesh_quat`,
+both already refused when they are not finite. The finiteness test now lives in
+`refuse_non_finite_scale`, the single owner both consumers call, and the loader
+reaches it where the scale is read - beside the missing-file contract, for the
+same reason: an asset declaration a body reaches has to be usable. The refusal's
+wording gains the xform consequence alongside the sizing one.
+
+A finite scale is untouched, including the 250 non-unit and the negative and zero
+ones the shipped corpus declares. All 60 resolvable registry models still load,
+and none declares a non-finite scale anywhere.
--- a/strands_robots/simulation/isaac/loaders.py
+++ b/strands_robots/simulation/isaac/loaders.py
@@ -2248,7 +2248,12 @@ def load_mjcf_scene_objects(path: str) -> list[SceneObject]:
         measured around one reports a collision bound for geometry the scene
         does not declare, so it is refused rather than approximated).
     """
-    from strands_robots.simulation.isaac.mesh_assets import MESH_EXTENSIONS, USD_EXTENSIONS, mesh_aabb
+    from strands_robots.simulation.isaac.mesh_assets import (
+        MESH_EXTENSIONS,
+        USD_EXTENSIONS,
+        mesh_aabb,
+        refuse_non_finite_scale,
+    )
 
     _require_existing_file(path, "MJCF scene")
     root = _parse_xml(path, "MJCF scene")
@@ -2307,6 +2312,15 @@ def load_mjcf_scene_objects(path: str) -> list[SceneObject]:
             asset = mesh_registry.get(mesh_name)
             if asset is not None:
                 candidate_path, mesh_scale = asset
+                # Refused here rather than at the measurement below, because the
+                # scale rides onto the scene object whichever branch supplies the
+                # bound: a body with collidable geometry takes the analytic bound
+                # and never measures, yet the realization still applies this scale
+                # to the visual prim's xform, beside the mesh geom's position and
+                # orientation - which are already refused when they are not
+                # finite. Placed beside the missing-file contract for the same
+                # reason: an asset declaration a body reaches has to be usable.
+                refuse_non_finite_scale(candidate_path, mesh_scale)
                 ext = os.path.splitext(candidate_path)[1].lower()
                 if ext in MESH_EXTENSIONS or ext in USD_EXTENSIONS:
                     if not os.path.isfile(candidate_path):
--- a/strands_robots/simulation/isaac/mesh_assets.py
+++ b/strands_robots/simulation/isaac/mesh_assets.py
@@ -56,6 +56,7 @@ __all__ = [
     "load_mesh_geometry",
     "mesh_aabb",
     "mesh_usd_cache_dir",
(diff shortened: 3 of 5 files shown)


yinsong1986 (approved):
## Summary

Closes the gap left by #2838: the mesh-scale finiteness check lived only in `mesh_aabb`, which `load_mjcf_scene_objects` reaches on just one of its three branches, while `mesh_scale` rides onto the `SceneObject` (and into the realization's `_author_local_xform` call) on all of them. The fix hoists the test into a single owner, `refuse_non_finite_scale`, delegates `mesh_aabb` to it, and calls it in the loader where the scale is read - so a body whose bound comes from a collidable geom no longer reports `mesh_scale=(nan, 1.0, 1.0)` under success. The guard sits before both the measurement and the `SceneObject` construction, the refusal is loud (`ValueError`) with the xform consequence added to the wording, and finite scales - negative, zero and non-unit - are pinned unchanged on both body shapes. This brings the scale to parity with its two xform siblings (`mesh_pos`, `mesh_quat`), which were already refused.

## What's good

- Verified locally at head: both scale suites pass (81/81), and the tree-wide graders touched by this diff (`Raises:` completeness, `__all__` definedness/privacy, test-name behaviour rule) are green. The 3 failures in `tests/simulation/isaac/` on a host without Isaac Sim pre-exist identically on the merge base - not attributable to this branch.
- Single-owner design is pinned structurally (AST scan over `mesh_assets` for `isfinite`, plus per-consumer reach tests), so a third consumer or a drifted copy fails rather than silently tolerating what the other refuses.
- The invalidated premise test (`test_the_declared_scale_still_rides_along` asserting the NaN carry) was repaired rather than deleted, keeping the claim it got right (the bound comes from the geom) on a finite non-unit scale - exactly the AGENTS.md "replace rather than delete an invalidated premise test" pattern.
- Changelog fragment under `changelog.d/`, ASCII-clean strings, no host paths, ruff clean on the touched files.
- The new `ValueError` is raised by a delegated helper whose own `Raises:` block documents it, and the loader's fail-loud contract already covers the class; the behaviour change is loud, not silent.

PR #2846, opened 2026-08-27
fix(policies/motionbricks): refuse a non-finite direction goal and name the key it came from

cagataycali:
## Problem

`MotionBricksPolicy` resolves two of the issue #300 well-known goal keys — `target_velocity` and `target_heading` — through one reader, `_unit_direction`, which normalises each to a unit direction in the world XY plane. It falls back to "walk straight ahead" for a command with no direction, and its own docstring gave the reason: so an all-zero command "does not produce a NaN direction".

That fallback is a magnitude test, so it could not cover a non-finite component. `nan < 1e-6` is `False`, and `inf` divided by its own norm is `nan`, so either one fell straight through it and produced exactly the outcome the fallback exists to prevent.

Measured on `f35e4f2f`, through the public builder:

| `target_velocity` | main | this branch |
| --- | --- | --- |
| `[0.5, 0.0]` | `movement_direction=[1.0, 0.0, 0.0]` | identical |
| `[0.0, 2.0]` | `[0.0, 1.0, 0.0]` | identical |
| `[0.0, 0.0]` | `[1.0, 0.0, 0.0]` (fallback fires) | identical |
| `[nan, 0.0]` | **`[nan, nan, nan]`**, reported success | refused, names `target_velocity` |
| `[inf, 0.0]` | **`[nan, 0.0, 0.0]`**, reported success | refused, names `target_velocity` |
| `[0.5, 0.0, 0.0, 0.0]` | accepted, read for its first two entries | refused, names `target_velocity` |
| `"abc"` | numpy's `could not convert string to float: 'abc'` | `'target_velocity' elements must be numbers, got 'abc'` |

The same held for `target_heading` and the `facing_direction` it drives.

Two further things about that read were ungraded.

**The refusal named neither key.** One reader serves both, and both answered `direction vector must have 2 or 3 entries`, so a caller passing both could not tell which one had been refused. The sibling locomotion family already names its key in every refusal — `WBCPolicy._validate_velocity` answers `target_velocity must have at least 3 elements [vx, vy, omega]` — and the policy contract's stated reason for leaving the component count out of the goal vocabulary is that "each receiver states its own arity and refuses a shape it cannot use", which needs the refusal to say which receiver and which key. `mesh.security.validate_command` defers the count here deliberately ("the component COUNT is not checked against any receiver's arity here"), so this reader is the only place it is checked at all.

**The message stated a rule the check did not enforce.** Five statements in the package gave the accepted shape as two or three components — the reader's docstring, the module docstring, the constructor's `target_velocity` entry, the refusal text itself and `docs/policies/motionbricks.md` — while the check tested the lower bound alone.

And the constructor's `target_velocity` default went through no domain at all, then was injected into every call that passed none.

## Change

`_unit_direction` now takes the calling `method` and the `param_name` it is reading, consults the shared `utils.finite_vector_error` for the per-component domain, and checks the count against a closed range named by two module constants. Both call sites pass the key they read. The constructor's default goes through the same reader, so a default that cannot be honoured is reported at construction rather than on the first step of a started rollout — the ordering the sibling family's own constructor comment gives as its reason.

Reusing `finite_vector_error` rather than restating the rule is what makes a `bool` refused by name instead of read as `1.0`, and keeps the wording identical to the simulation setters. The ordering of the two checks is observable and is pinned: a value with no readable length has no count to compare, so the arity check running first would answer `got None` — reporting a domain check that never ran — while the component domain says what actually happened.

## Why nothing caught it

The suite stated the property without driving the input that violates it. Two shipped cells say so in their own words:

- `test_build_control_signals_rejects_short_direction` — "the builder rejects it instead of **fabricating a NaN/garbage heading**"
- `test_build_control_signals_zero_velocity_falls_back_to_forward` — "an all-zero command **must not yield a NaN direction**"

Both pass only the inputs where the fallback does fire. Each has been narrowed to the half it grades and now points at the new cells for the other; the first also asserts the key is named, which is why it is the one sibling cell that fails on the pre-fix tree.

## Tests

`tests/policies/motionbricks/test_direction_goal_keys_are_held_to_one_domain.py`, 78 cells over both keys and both doors. Pre-fix split, source reverted and tests kept: **47 failed / 31 passed**.

| class | role | pre-fix |
| --- | --- | --- |
| `TestANonFiniteComponentIsRefused` | regression | 14/14 |
| `TestTheRefusalNamesTheKeyItCameFrom` | regression | 14/14 |
| `TestTheStatedArityIsTheEnforcedArity` | regression | 11/11 |
| `TestTheConstructorDefaultIsHeldToTheSameDomain` | regression | 5/5 |
| `TestTheReaderIsSingleSourced` | regression (structural) | 3/3 |
| `TestWhatIsUnchanged` | over-reach control | **0 of 5** |
| `TestThePremises` | premise | **0 of 3** |

Mutation table, each row against the new file, `collected` stable at 78 and the control clean:

| mutation | fires |
| --- | --- |
| control (nothing changed) | 0 |
| component domain call removed | 31 |
| arity range check removed | 17 |
| both removed (behavioural revert) | 48 |
| arity weakened to a floor only | 9 |
| message drops the key it came from | 7 |
| both call sites pass one key name | 8 |
| constructor default not validated | 6 |
| domain restated locally instead of consulted | 8 |
| arity checked before the components | 4 |
| finiteness narrowed to `nan` only | 15 |

Each half is independently pinned: the two single-half reverts are **disjoint** and their union is exactly the both-halves revert (31 + 17 = 48).

## Gate

- `ruff check` / `ruff format --check` clean over 1597 files.
- `mypy strands_robots tests tests_integ`: **24 == 24** against a worktree of the merge base, normalized message sets byte-identical in both directions, 0 attributable (`checked 1649` vs `1648`, the one extra file being the new test).
- Paired run, identical 216-file selection (all of `tests/policies/` plus every guard walking the tree, parsing source or reading docstrings, `Args:`, `Raises:`, `__all__` or `docs/`), the new file excluded from both trees: identical **8656 collected**, `8600 passed, 56 skipped, 0 failed` on each, distinct rootdirs.

## Artifact

`motionbricks` is not installed on this node, so the generator itself cannot be stepped. The quantity the fix governs is the direction vector handed to it, and that is what is rendered — one capture script driving the real reader against both trees, with every claim in the caption asserted from the two captures before anything is drawn.

target_velocity resolved to movement_direction, main vs this branch

## Scope

`target_heading_angle` is a scalar rather than a direction vector and does not go through this reader; it is pinned as unchanged. The sibling family's own arity stays as it is — the contract deliberately does not fix a component count across receivers, and this change makes that reason true for this receiver rather than making the two agree.

### DIFF
--- a/changelog.d/2846-motionbricks-direction-goal-domain.md
+++ b/changelog.d/2846-motionbricks-direction-goal-domain.md
@@ -1,0 +1,53 @@
+### Fixed: a non-finite direction goal is refused, and the refusal names the key it came from
+
+`MotionBricksPolicy` resolves two of the issue #300 well-known goal keys -
+`target_velocity` and `target_heading` - through one reader that normalises each
+to a unit direction in the world XY plane. That reader falls back to "walk
+straight ahead" for a command with no direction, and its own docstring gave the
+reason: so an all-zero command "does not produce a NaN direction".
+
+The fallback is a magnitude test, so it could not cover a non-finite component.
+`nan < 1e-6` is `False`, and `inf` divided by its own norm is `nan`, so either
+one fell straight through it and produced exactly the outcome the fallback exists
+to prevent: `target_velocity=[nan, 0.0]` resolved to a `movement_direction` of
+`[nan, nan, nan]`, handed to the generator by a call that reported success. The
+same held for `target_heading` and the `facing_direction` it drives. The
+per-component domain is now the shared `finite_vector_error` the simulation
+setters already hold their own vectors to, so a `nan`/`inf` component, a
+non-numeric element and a `bool` are each refused in the library's own words
+rather than read as `1.0` or surfaced as numpy's `could not convert string to
+float`.
+
+The refusal named neither key. One reader serves both, and both answered
+`direction vector must have 2 or 3 entries`, so a caller passing both could not
+tell which one had been refused. Each call site now passes the key it reads. The
+sibling locomotion family already did this - `WBCPolicy._validate_velocity`
+answers `target_velocity must have at least 3 elements [vx, vy, omega]` - and the
+policy contract's stated reason for leaving the component count out of the goal
+vocabulary is that "each receiver states its own arity and refuses a shape it
+cannot use", which needs the refusal to say which receiver and which key. The
+mesh wire validator defers the count here deliberately ("the component COUNT is
+not checked against any receiver's arity here"), so this reader is the only place
+it is checked at all.
+
+And the message stated a rule the check did not enforce. Five statements in the
+package gave the accepted shape as two or three components - the reader's
+docstring, the module docstring, the constructor's `target_velocity` entry, the
+refusal text itself and `docs/policies/motionbricks.md` - while the check tested
+the lower bound alone. A four-, six- or twenty-component vector was accepted and
+read for its first two entries, so a caller who packed a six-component spatial
+twist got planar motion and no error. The count is now the closed range every
+one of those statements already described.
+
+The constructor's `target_velocity` default went through no domain at all and was
+injected into every call that passed none. It now goes through the same reader,
+so a default that cannot be honoured is reported at construction rather than on
+the first step of a started rollout - the ordering the sibling family's own
+constructor comment gives as its reason.
+
+Nothing caught any of this because the suite stated the property without driving
+the input that violates it. Two shipped cells say so in their own words - "the
+builder rejects it instead of fabricating a NaN/garbage heading", "an all-zero
+command must not yield a NaN direction" - and both pass only the inputs where the
+fallback does fire. Both have been narrowed to the half they grade and now point
+at the new cells for the other.
--- a/docs/policies/motionbricks.md
+++ b/docs/policies/motionbricks.md
@@ -138,15 +138,25 @@ policy = create_policy(
 
 ### Per-call goal kwargs
 
-| kwarg | Meaning |
-| --- | --- |
-| `style` / `mode` | Clip mode - an `int` index or `str` name (e.g. `"walk"`, `"stealth_walk"`, `"walk_boxing"`, `"hand_crawling"`). |
-| `target_velocity` | `[vx, vy]` desired planar movement direction (world frame); only the direction is used. |
-| `target_heading` | `[hx, hy]` facing direction, or `target_heading_angle` (radians). |
+| kwarg | Accepted | Meaning |
+| --- | --- | --- |
+| `style` / `mode` | `int` index or `str` name | Clip mode (e.g. `"walk"`, `"stealth_walk"`, `"walk_boxing"`, `"hand_crawling"`). |
+| `target_velocity` | 2 or 3 entries, every component a finite number | `[vx, vy]` (or `[vx, vy, vz]`) desired planar movement direction (world frame); only the direction is used, and a third entry is projected away. |
+| `target_heading` | 2 or 3 entries, every component a finite number | `[hx, hy]` facing direction, or `target_heading_angle` (radians, any finite float). |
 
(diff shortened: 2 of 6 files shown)


yinsong1986 (dismissed):
## Summary

Closes the non-finite hole in `MotionBricksPolicy`'s direction-goal reader: `_unit_direction`'s near-zero fallback is a magnitude test, so `nan`/`inf` components fell through it and resolved to a `[nan, nan, nan]` `movement_direction`/`facing_direction` from a call that reported success. The reader now consults the shared `finite_vector_error` domain (finite, numeric, `bool` refused by name), enforces the documented 2-or-3 arity as a closed range instead of a floor, names the goal key (`target_velocity` / `target_heading`) in every refusal, and routes the constructor's `target_velocity` default through the same reader so an unusable default is reported at construction rather than mid-rollout. The previously-accepted 4+-component vector is now refused, but with an explicit `ValueError` and matching updates to the module docstring, constructor docs, `docs/policies/motionbricks.md`, and a changelog fragment - the documented shape was always 2-or-3, so this closes an enforcement gap rather than changing the contract.

## What's good

- Verified locally at head `8aed3486`: all 279 `tests/policies/motionbricks/` tests pass; `ruff check`, `ruff format --check`, and `mypy` are clean on the changed files; the docstring cross-reference and `Raises:`/`Args:` completeness gates pass (the new docstrings cite `validate_command` and `WBCPolicy._validate_velocity`, both of which exist).
- The new test module drives both keys through both doors (per-call kwarg and constructor default), pins the check ordering (component domain before arity, so a length-less value is not reported as `got None`), and carries over-reach controls asserting every legitimate shape (tuples, ints, numpy scalars/arrays, third-entry projection, near-zero fallback, `target_heading_angle`) is unchanged.
- Domain reuse follows AGENTS.md's shared-guard rule: the reader consults `finite_vector_error` rather than restating it, and a structural cell pins that the constructor delegates to the same reader so the two doors cannot drift.
- Changelog fragment (not a direct `CHANGELOG.md` edit), ASCII-only refusal text, no host paths, private helpers stay out of `__all__`.

yinsong1986 (approved):
## Summary

Closes the non-finite hole in `MotionBricksPolicy`'s direction-goal reader: `_unit_direction`'s near-zero fallback is a magnitude test, so a `nan`/`inf` component in `target_velocity` or `target_heading` fell through it and resolved to a `[nan, nan, nan]` direction handed to the generator from a call that reported success. The reader now consults the shared `finite_vector_error` domain, enforces the closed `[2, 3]` arity that five existing statements already described (previously only the floor was checked), and every refusal names the calling method and the goal key it read — matching the sibling `WBCPolicy` convention and the policy contract's stated reason for leaving arity to each receiver. The constructor's `target_velocity` default goes through the same reader, so an unusable default is refused at construction with an explicit `ValueError` rather than surfacing on the first rollout step.

## What's good

- The domain is consulted, not restated: one owner (`finite_vector_error`) for the per-component rule, structurally pinned so a local copy cannot drift — exactly the shared-guard convention in AGENTS.md.
- Check ordering (component domain before arity) is deliberate and pinned, so a value with no readable length is reported for what actually happened rather than as `got None`.
- 78-cell regression file with over-reach controls (zero-vector fallback, numpy arrays, third-entry projection, `target_heading_angle` all pinned unchanged), premise cells, and a disjoint mutation table (31 + 17 = 48).
- The two shipped cells that stated the property without driving the violating input are narrowed honestly and cross-reference the new file.
- Docs, changelog fragment, and the constructor's `Args:` entry all updated in the same PR; repo guard suites (Raises/Args docstring completeness, xref resolution, test naming, ASCII, host paths) verified green on the head.
- Scope discipline: the scalar `target_heading_angle` path and the sibling family's arity are explicitly left alone, with the reason stated.

PR #2847, opened 2026-08-27
test(drivers/g1): grade the rt/lowcmd wire-frame contract on CI, not only where the SDK is installed

cagataycali:
## The wire-frame contract was graded by nothing in CI

`tests/drivers/test_g1_driver.py` covers `send_action`'s write path and the two `LowCmd_` builders in 20 cells, and every one of them sat behind `skipif(not _HAS_SDK)`.

`unitree-sdk2` is in no extra and in no core dependency, and `hatch`'s default env resolves `features = ["all"]`, so `_HAS_SDK` is False on the runner that gates a merge. Modelling that (both layers a suite can consult: `find_spec` returns `None`, and a real import raises `ModuleNotFoundError`), the suite reads:

| suite | SDK present | SDK hidden, i.e. `call-test-lint` |
|---|---|---|
| `test_g1_driver.py` | 73 passed | **53 passed, 20 skipped** |
| `test_g1_control_loop.py` | 35 passed | 35 passed, 0 skipped |

Same 73 collected either way, so the cells exist and their disposition moves. Among the 20 CI never runs are the four fields the G1 firmware validates:

- `test_build_lowcmd_sets_a_matching_crc_the_firmware_will_accept`
- `test_build_lowcmd_echoes_mode_machine_and_pins_mode_pr_to_zero`
- `test_build_lowcmd_enables_the_touched_slot_and_leaves_the_rest_disabled`
- `test_build_zero_torque_stamps_crc_and_enables_every_named_slot`

A frame with a wrong CRC, a wrong `mode_machine`, `mode_pr = 1` or a Disable byte is one the robot silently ignores or acts on with the wrong physical meaning, and `send_action` reports `success` either way. Those are the regressions the required check could not see.

## The rule is already stated in-tree, for the same builder

`tests/drivers/test_g1_per_joint_gains` grades `_build_lowcmd_from_action` and writes its layering down:

> The contract cells need no `unitree_sdk2py`, so `call-test-lint` grades them in CI, where the SDK is not installed. `unitree-sdk2` is not a declared dependency of this project, so a contract asserted only behind `skipif(not _HAS_SDK)` is asserted by nothing in CI.
> The wire cells, which do need the SDK to build a real `LowCmd_`, confirm the contract cells describe what actually lands on the topic.

Four sibling suites act on it by installing a `unitree_sdk2py` stub on `sys.modules` and driving the production lane against it. Across the 12 g1 suites, 8 run fully in CI; this module loses the most (20 of 73).

Its section comment gave the reason for skipping instead: "a missing SDK short-circuits the whole class (the driver refuses with the SDK's own reason before it would build a LowCmd_)". That is true of the driver, and it is an argument for supplying a `LowCmd_`-shaped object -- which is what the siblings do -- rather than for leaving the contract ungraded.

## The change

Eighteen of the twenty need only that shape to write into. They now take the stub `tests.drivers.test_g1_control_loop` installs, imported rather than copied (the idiom `test_g1_cleanup_reads_the_loop_halt_outcome` already uses) so every suite grading this builder writes into the same object.

Two cells keep the marker, and this is the split the sibling names: they recompute the SDK's own CRC as an independent oracle, and under a stub whose `Crc` returns a constant, `cmd.crc == CRC().Crc(cmd)` compares that constant against itself and grades nothing.

One cell is added. `test_build_zero_torque_stamps_crc_and_enables_every_named_slot` does four jobs and only one is a CRC question; the Enable bound over the 29 named slots and the reserved tail is decidable against any `LowCmd_`-shaped object. The break table below found that gap -- without the new cell, a stop frame left at Disable was still undetectable in CI.

The fixture is **opt-in per cell rather than autouse**, because this module also pins the SDK-*absent* refusals (`test_ensure_dds_reports_missing_sdk`, `test_connect_eagerly_reports_reason_without_sdk`). A module-wide stub would have made those unreachable while leaving them green. Both pass in both dependency modes here.

No production code changes.

## What CI can now detect

Each mutation is a real wire-frame regression, applied to `strands_robots/drivers/g1.py` and run with the SDK hidden -- the tree `call-test-lint` sees. `collected` is stable at 73 (base) and 74 (branch) on every row, so no row hides a deselection.

| mutation | before | after | cell that catches it |
|---|---|---|---|
| control, no mutation | 0 | 0 | - |
| CRC never stamped | **0** | 1 | `..._echoes_cached_mode_machine_not_fsm_id` (`crc != 0`) |
| `mode_pr = 1` (remaps four ankle indices) | **0** | 2 | `..._echoes_mode_machine_and_pins_mode_pr_to_zero` |
| `mode_machine` not echoed | **0** | 2 | `..._echoes_mode_machine_and_pins_mode_pr_to_zero` |
| Enable byte left Disable | **0** | 1 | `..._enables_the_touched_slot_and_leaves_the_rest_disabled` |
| per-joint gains collapsed to a scalar | **0** | 1 | `..._applies_default_gains_for_scalar_targets` |
| stop frame keeps its stiffness | **0** | 1 | `..._zero_torque_lowcmd_zeroes_every_motor_field` |
| stop frame slot left Disable | **0** | 1 | `..._zero_torque_enables_every_named_slot_and_leaves_the_tail` (added) |

Seven regressions, undetectable before, detected after. The source was restored byte-identically after every row.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1650 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a pristine worktree of the merge-base, normalized message sets byte-identical in both directions, 0 attributable to the changed file.
- Paired gate, identical 196-file selection (all of `tests/drivers/` plus every suite that walks the tree, parses source, or reads docstrings), the changed file excluded from both trees, every path verified present on both: identical **7811 collected** and `7745 passed, 66 skipped, 0 failed` on each, distinct rootdirs.
- `tests/drivers/` on the rebased tree, with the guard `main` added since: **699 passed, 3 skipped**.
- The suite itself: unchanged at 73 passed where the SDK is present (74 with the added cell), and `53 passed, 20 skipped` becomes `72 passed, 2 skipped` where it is not.

No visual artifact: this is a test-visibility change with no rendering, recording, policy or asset behaviour. The break table is the evidence.

### DIFF
--- a/changelog.d/2847-g1-wire-frame-contract-graded-on-ci.md
+++ b/changelog.d/2847-g1-wire-frame-contract-graded-on-ci.md
@@ -1,0 +1,44 @@
+### Tests: the g1's ``rt/lowcmd`` wire-frame contract is graded in CI, not only where the SDK is installed
+
+Twenty cells in ``tests/drivers/test_g1_driver.py`` covered ``send_action``'s
+write path and the two ``LowCmd_`` builders, and every one of them sat behind
+``skipif(not _HAS_SDK)``. ``unitree-sdk2`` is in no extra and in no core
+dependency, and ``hatch``'s default env resolves ``features = ["all"]``, so
+``call-test-lint`` skipped all twenty: the CRC, the ``mode_machine`` echo,
+``mode_pr`` and the Enable byte -- the four fields the G1 firmware validates,
+and the four a review of the write path asked for by name -- were asserted by
+nothing on the runner that gates a merge. The suite reported ``53 passed, 20
+skipped`` there while reporting ``73 passed`` on a box that happens to carry
+the SDK.
+
+``tests/drivers/test_g1_per_joint_gains`` already states the rule for the same
+builder: a contract asserted only behind ``skipif(not _HAS_SDK)`` is asserted
+by nothing in CI. Four sibling suites act on it by installing a
+``unitree_sdk2py`` stub on ``sys.modules`` and driving the production lane
+against it. This module was the one that skipped instead, and its section
+comment gave the reason as "a missing SDK short-circuits the whole class" --
+true of the driver, and an argument for supplying a ``LowCmd_``-shaped object
+rather than for leaving the contract ungraded.
+
+Eighteen of the twenty need only that shape to write into, so they now take the
+stub ``tests.drivers.test_g1_control_loop`` installs, imported rather than
+copied so every suite grading this builder writes into the same object. The two
+cells that recompute the SDK's own CRC as an independent oracle keep the marker:
+under a stub whose ``Crc`` returns a constant, ``cmd.crc == CRC().Crc(cmd)``
+compares that constant against itself and grades nothing. One added cell takes
+the half of the stop-frame oracle that is not a CRC question -- the Enable bound
+over the 29 named slots and the reserved tail -- so a stop frame that lands as a
+wire-side no-op is refused by CI rather than by a box with the SDK.
+
+The fixture is opt-in per cell rather than autouse because this module also pins
+the SDK-*absent* refusals (``test_ensure_dds_reports_missing_sdk``,
+``test_connect_eagerly_reports_reason_without_sdk``); a module-wide stub would
+have made those unreachable while leaving them green.
+
+On the tree CI runs, seven wire-frame regressions were undetectable before and
+are detected now: a dropped CRC stamp, ``mode_pr = 1`` (which silently remaps
+four ankle indices), a dropped ``mode_machine`` echo, an Enable byte left at
+Disable in either builder, the per-joint gain table collapsed to a scalar, and a
+stop frame that keeps its stiffness. The suite is unchanged where the SDK is
+present (73 passed, now 74 with the added cell) and moves from ``53 passed, 20
+skipped`` to ``72 passed, 2 skipped`` where it is not.
--- a/tests/drivers/test_g1_driver.py
+++ b/tests/drivers/test_g1_driver.py
@@ -9,6 +9,7 @@ that is a class problem, not a robotics problem.
 from __future__ import annotations
 
 import asyncio
+import sys
 import types
 from typing import Any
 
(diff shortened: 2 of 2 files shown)


yinsong1986 (approved):
## Summary

Test-only change that moves 18 of the 20 `skipif(not _HAS_SDK)` cells in `tests/drivers/test_g1_driver.py` onto an opt-in `unitree_sdk2py` stub fixture, so the `rt/lowcmd` wire-frame contract (CRC stamped, `mode_machine` echo, `mode_pr` pinned to 0, Enable byte bounds) is graded by `call-test-lint`, where the SDK is not installed. One cell is added to grade the stop frame's Enable bound without the SDK; the two cells that recompute the SDK's own CRC as an independent oracle correctly keep the marker, since a stub CRC would compare a constant against itself. No production code changes.

## What's good

- Verified locally on the head tree with the SDK absent: `tests/drivers/test_g1_driver.py` reads `72 passed, 2 skipped` (matching the description), and the full `tests/drivers/` selection is green.
- The fixture is opt-in per cell rather than autouse, and the two SDK-*absent* refusal cells (`test_ensure_dds_reports_missing_sdk`, `test_connect_eagerly_reports_reason_without_sdk`) still pass with it in place - the exact hazard a module-wide stub would have introduced is named and avoided in the section comment.
- Stub classes are imported from `tests.drivers.test_g1_control_loop` rather than copied, so every suite grading `_build_lowcmd_from_action` writes into the same `LowCmd_` shape; `monkeypatch.setitem` restores `sys.modules` per AGENTS.md > Testing Patterns.
- The mutation break table in the description is the right evidence for a visibility change, and the changelog fragment is named for the PR number per step 3 of the PR workflow.

## Verification suggestions

- `python -m pytest tests/drivers/test_g1_driver.py -q` on a box **with** `unitree-sdk2` installed, to confirm the 74-passed/0-skipped side of the table (CI can only exercise the SDK-absent side).

PR #2849, opened 2026-08-27
fix(training/rl): hold num_envs to the shared count domain on both backends

cagataycali:
## What is wrong

`RLTrainSpec.num_envs` is the third caller-supplied factor of the one loop bound every RL backend derives:

```
steps_per_iter = rollout_steps * num_envs
num_iters      = max(1, total_timesteps // steps_per_iter)
for it in range(num_iters):   # collect, update
```

Its two siblings go through `positive_count_error` in `rl_run_size_problems`, whose docstring works through why a bare `value <= 0` test on a *derived* bound is weaker than the same test on a field read straight off the spec.

`num_envs` was deliberately left out of that gate, and the reason holds: which **counts** are usable differs per backend - PPO parallelizes over any positive count, the MuJoCo-backed FastSAC is single-env and requires exactly `1` - so that half is not one shared rule, and each backend stated its own with a bare comparison (`< 1`, `!= 1`).

What the reason does not cover is whether the value is a count at all. That half is identical on both backends and was stated by neither, so the third factor of this product sat outside the domain its two siblings are held to.

## Measured

`total_timesteps=1000`, `rollout_steps=64` - which asks for 15 iterations of 64 steps.

| `num_envs` | PPO `< 1` | FastSAC `!= 1` | `num_iters` | consequence |
|---|---|---|---|---|
| `nan` | accepted | refused | **1000** | `success`, announced `"1000 iterations x 1 steps complete"` while each iteration collected 64 steps - 66x the budget, and the message misdescribes both factors |
| `inf` | accepted | refused | **1** | `success`, announced `"1 iterations x inf steps complete"` |
| `2.5` | accepted | refused | `6.0` | `TypeError: 'float' object cannot be interpreted as an integer` out of `range()`, after `setup` built the env, networks and optimizers |
| `1.0` | accepted | **accepted** | `15.0` | same raise, same depth |
| `True` | accepted | **accepted** | 15 | length is right; a value reading as a flag was accepted as a count of one |
| `"4"` / `None` | **raises** | refused | - | `TypeError: '<' not supported between instances of 'str' and 'int'` out of a `Trainer.validate` documented to *return* problems |
| `4` | accepted | refused | 3 | the per-backend line, and it is unchanged |

`100.5` is worth separating from the float-bound row: `1000 // 6432.0` is `0.0`, which the clamp replaces with the int `1`, so `range()` is happy and the run is one iteration instead of fifteen. A float is not always a float bound, and the two harms are pinned separately.

## The change

Both backends consult the shared domain first and ask their own count rule only of a count - the idiom each already uses one statement later for a relation between two counts (`positive_count_error(spec.rollout_steps, ...) is None and <relation>` in PPO, the same for `batch_size` in FastSAC), with the same stated reason: it holds wherever in the preflight it sits.

- The per-backend line is unchanged: `4` is usable for the parallel backend, refused as single-env by the other.
- A non-count is no longer reported as the wrong *number* of environments.
- The refusal names the backend that refused, which the bare comparison's message did not.
- PPO keeps no second comparison: over the values the domain accepts, every positive `int` parallelizes, so a `< 1` branch there would be unreachable, and dead code that reads as a rule is worse than none. This changes PPO's message for `0` / `-1` from `num_envs must be >= 1` to `ppo: num_envs must be a positive integer`, so the two pins of the old wording in `test_rl_ppo.py` move with it - and in that file's own parametrized table `num_envs` now reads the same way its two sibling factors already did.

Two prose corrections in the same change, because both invited the gap to stay open:

- `rl_run_size_problems`' scope paragraph said the whole field was excluded. It now says which half.
- `TestNumEnvsIsNotInTheSharedDomain`'s non-vacuity test claimed "excluding it from the domain did not leave it unchecked" and drove only `0` - the one unusable value a bare comparison already caught. It is widened to the values that survive one, on both backends.

## Why nothing caught it

The gate's own docstring enumerates, with measurements, exactly the values a bare comparison lets through for the two sibling factors - `True`, a fraction, `nan`, `inf`, a non-numeric - and `num_envs` is guarded by a bare comparison one statement after the sentence that condemns one. The non-vacuity test that exists to say the excluded field is still checked graded the single value the comparison does catch.

## Verification

Pre-fix split over the three touched test files: **75 failed / 367 passed**, with 0 in every control and premise class - `TestTheHarmTheDomainReplaces` 0/7, `TestThePerBackendCountRuleIsUnchanged` 0/9, `TestThePremises` 0/4, and 0 in each of the 8 unrelated classes in `test_rl_run_size_domain.py` (156, 48, 32, 29, 10, 6, 4 cases) and in `test_rl_fast_sac.py` (0/17).

Mutation table, `collected` stable at 426 on every row and 0 collection errors:

|---|---|---|---|
| b0 | 0 | 0 | control |
| b1 | 48 | 0 | ppo: shared domain dropped, bare `< 1` restored |
| b2 | 27 | 0 | fast_sac: shared domain dropped, bare `!= 1` restored |
| b3 | 75 | 0 | both dropped (the full revert) |
| b4 | 20 | 0 | fast_sac: count rule asked first, domain after |
| b5 | 1 | 0 | fast_sac: `elif` becomes `if` (both complaints for a non-count) |
| b6 | 24 | 0 | ppo: weaker domain (`finite_number_error`, accepts a float) |
| b7 | 51 | 0 | ppo: verdict computed and discarded |
| b8 | 51 | 0 | ppo: domain told the wrong field name |
| b9 | 11 | 0 | ppo: domain told a literal context, not the backend |
| b10 | 5 | **1** | over-reach: fast_sac drops its OWN single-env count rule |
| b11 | 1 | 0 | the scope paragraph reverted (prose only) |

11 distinct firing sets of 12. b7 and b8 share one for a structural reason - both make PPO silent about `num_envs`, one by dropping the verdict and one by putting it under another name. **b1 and b2 are disjoint and `b1 | b2 == b3` exactly** (48 + 27 = 75), so each backend's half is independently pinned. b10 is the only row that trips a sibling suite, which is where `test_rl_fast_sac.py` pins the single-env message - the boundary this change must not cross.

Paired gate, 259-file identical selection (all of `tests/training/` plus every suite walking the tree, parsing source, or reading docstrings / `Args:` / `Raises:` / `changelog.d`, plus everything naming `num_envs` / `positive_count_error` / `steps_per_iter`), the new file excluded from both trees:

- branch `3 failed, 13557 passed, 61 skipped` (13621 collected); base `3 failed, 13535 passed, 61 skipped` (13599 collected)
- identical failing ids, **both `comm` directions empty**, distinct rootdirs
- `+22 collected` is exactly the widened `TestNumEnvsIsNotInTheSharedDomain` cells
- the 3 are the lerobot-from-source `TypeError: encode() takes 1 positional argument` drift, reproduced on a pristine base worktree alone

`tests/training/` whole: base `3 failed, 3519 passed`, branch `3 failed, 3638 passed` - `+119` = the 97 new cells plus the 22 widened ones.

`ruff check` and `ruff format --check` clean over 1652 files. mypy **24 == 24** against a pristine worktree of the same base, normalized message sets byte-identical in both directions, 0 attributable.

No visual artifact: this is a read-only preflight over a spec field, not a policy, simulation, rendering, recording or asset change. The measured tables above are the evidence.

### DIFF
--- a/changelog.d/2849-rl-env-count-is-a-count-on-every-backend.md
+++ b/changelog.d/2849-rl-env-count-is-a-count-on-every-backend.md
@@ -1,0 +1,59 @@
+### Fixed: `num_envs` is held to the shared count domain on both RL backends
+
+`RLTrainSpec.num_envs` is the third caller-supplied factor of the one loop bound
+every RL backend derives, `max(1, total_timesteps // (rollout_steps * num_envs))`
+iterated as `range()`. Its two siblings are held to
+`strands_robots.utils.positive_count_error` by `rl_run_size_problems`, whose own
+docstring works through why a bare `value <= 0` test on a *derived* bound is
+weaker than the same test on a field read straight off the spec: the `max(1, ...)`
+clamp turns every value that survives the comparison but cannot divide into a
+silent wrong-length run.
+
+`num_envs` was deliberately left out of that gate, for a reason that holds: which
+*counts* are usable genuinely differs per backend - PPO parallelizes over any
+positive count where the MuJoCo-backed FastSAC is single-env and requires exactly
+`1` - so that half is not one shared rule, and each backend stated its own with a
+bare comparison. What the reason does not cover is whether the value is a count
+at all. That half is identical on both backends and was stated by neither, so the
+third factor of this product sat outside the domain its two siblings are held to.
+
+Measured over `total_timesteps=1000` and `rollout_steps=64`, which asks for 15
+iterations of 64 steps:
+
+- `nan` passed PPO's `< 1`. `64 * nan` is `nan`, `max(1, nan)` keeps the `1`, and
+  `1000 // 1` is `1000`: the run reported `success` and announced `"1000
+  iterations x 1 steps complete"` while each of those 1000 iterations collected
+  `rollout_steps` steps - 66x the requested budget, under a message that
+  misdescribes both factors.
+- `inf` passed the same test and ran **one** iteration, announcing `"1 iterations
+  x inf steps complete"`.
+- `1.0` and `2.5` kept the bound a float, which raises `TypeError: 'float' object
+  cannot be interpreted as an integer` out of `range()` - after `setup` had built
+  the environment, the networks and the optimizers, which is the cost a read-only
+  preflight exists to precede. `1.0` reached that through FastSAC's `!= 1` as
+  well. A large float does not: `100.5` makes `1000 // 6432.0` the float `0.0`,
+  which the clamp replaces with the int `1`, so it runs one iteration instead.
+- `True` passed both backends - it satisfies `< 1` being false and `!= 1` being
+  false - and is numerically `1`, so the run length was right and what was wrong
+  is that a value reading as a flag was accepted as a count. That is the reason
+  the shared domain refuses a `bool` rather than testing a bound.
+- `"4"` and `None` raised `TypeError: '<' not supported between instances of
+  'str' and 'int'` out of PPO's comparison itself, from a `Trainer.validate`
+  documented to *return* problems.
+
+Both backends now consult the shared domain first and ask their own count rule
+only of a count, matching the idiom each already uses one statement later for a
+relation between two counts. The per-backend line is unchanged and pinned: `4` is
+still usable for the parallel backend and still refused as single-env by the
+other, and a non-count is no longer reported as the wrong *number* of
+environments. PPO keeps no second comparison, because over the values the domain
+accepts every positive `int` parallelizes, so a `< 1` branch there would be
+unreachable - and the refusal now names the backend that refused, which the bare
+comparison's message did not.
+
+The scope paragraph in `rl_run_size_problems` said the whole field was excluded;
+it now says which half, so the next reader is not invited to leave the shared half
+open again. `TestNumEnvsIsNotInTheSharedDomain`'s non-vacuity test asserted that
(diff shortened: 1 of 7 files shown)


yinsong1986 (approved):
## Summary

Closes the type half of the `num_envs` preflight on both RL backends: each backend now consults the shared `positive_count_error` domain before asking its own per-backend count rule, so `nan` / `inf` / integral floats / `bool` / `str` / `None` are returned as problems instead of producing a silent wrong-length run, a `TypeError` out of `range()` after setup, or a raise out of a `validate` documented to return problems. The per-backend scope line is genuinely unchanged (verified end-to-end: `4` is still accepted by PPO and refused as single-env by FastSAC, `1` accepted by both), and the change matches the idiom both backends already use one statement later for their relation-between-counts rules. Verified against the head SHA: the four touched training suites pass (443 passed), `ruff check` and `ruff format --check` are clean on every changed file, all strings are ASCII, no host paths in the new test file, and the changelog fragment is present under `changelog.d/` per AGENTS.md step 3.

## What's good

- The refusal now names the backend and the field via the shared domain's message, and the FastSAC `elif` keeps the single-env message reserved for actual counts - a non-count is no longer reported as the wrong number of envs.
- The two prose pins (`rl_run_size_problems` scope paragraph, the widened non-vacuity test in `test_rl_run_size_domain.py`) close the documentation gap that let the field sit outside the domain in the first place.
- Test updates in `test_rl_ppo.py` move exactly the two pins of the old wording, nothing else.

PR #2850, opened 2026-08-27
test(drivers/g1): grade the per-joint gain wire cells in CI, and hold the gating rule to itself

cagataycali:
## What this changes

`tests/drivers/test_g1_per_joint_gains` is the module that writes the rule down:

> `unitree-sdk2` is not a declared dependency of this project, so a contract asserted only behind `skipif(not _HAS_SDK)` is asserted by nothing in CI.

Five of its own cells sat behind that marker. They are the whole wire half of the per-joint gain contract, and none of them needs the real SDK.

| where | SDK present | SDK absent (every `call-test-lint` runner) |
|---|---|---|
| `test_g1_per_joint_gains.py`, before | 40 passed, 0 skipped | **35 passed, 5 skipped** |
| `test_g1_per_joint_gains.py`, after | 40 passed, 0 skipped | **40 passed, 0 skipped** |

`35 + 5 == 40 + 0` on the same 40 collected, so the same cells are collected either way and only their disposition moves.

The five: every slot's gains on a full frame, a knee and an arm carrying different gains in one frame, the partial fallback where a caller tunes `kp` and leaves `kd`, a supplied gain overriding the table, and the stop frame staying soft.

## Why the marker was not the right tool for them

The module's own layering note explains it: the vendor gain tuples are stated locally there *precisely* so the value cells are an independent oracle rather than a tautology. So the oracle is those tuples, not anything the SDK computes, and what the cells need is a `LowCmd_`-**shaped** object to write into. `_StubLowCmd` already carries a 35-slot `motor_cmd` whose entries have `kp`/`kd`, so they now take the shape from the stub `tests.drivers.test_g1_control_loop` installs.

An AST read of the whole test tree separates the two kinds with no judgement calls:

| gated node | cells | imports the SDK inside itself |
|---|---|---|
| `test_g1_driver.py::test_build_lowcmd_sets_a_matching_crc_the_firmware_will_accept` | 1 | yes, `unitree_sdk2py.utils.crc` |
| `test_g1_driver.py::test_build_zero_torque_stamps_crc_and_enables_every_named_slot` | 1 | yes, `unitree_sdk2py.utils.crc` |
| `test_g1_per_joint_gains.py::TestTheWireFrameCarriesTheSlotsOwnGains` | 3 | no |
| `test_g1_per_joint_gains.py::TestWhatTheTableDoesNotChange` | 2 | no |

The two that import the SDK recompute a value it computes, so a stub `Crc` returning a constant would compare that constant against itself. They keep the marker.

The fixture is opt-in per class rather than autouse: a module-wide stub would quietly make an SDK-*absent* refusal cell unreachable if one is added here later, the way it would in `test_g1_driver`.

## What CI can now catch

Real gain regressions in `strands_robots/drivers/g1.py`, measured with the SDK hidden, against the module before and after. `collected` is 40 on every row and the control is clean both ways.

| # | regression | before | after |
|---|---|---|---|
| b0 | control, nothing changed | 0 | 0 |
| b1 | `kp` default collapsed to one value for all 29 slots | 1 | 3 |
| b2 | `kd` default collapsed to one value for all 29 slots | 1 | 3 |
| b3 | both tables read slot 0 instead of the joint's own slot | **0** | 3 |
| b4 | the gain table leaks into the zero-torque stop frame | **0** | 1 |
| b5 | a supplied `kp` no longer wins over the table | **0** | 2 |

b3 is the sharpest of the three that were invisible: it keeps the `_SDK_KP[...]` subscript the source-reading contract cell looks for, and changes only which slot's gains land on the wire. b4 makes a *stop* frame stiff. The firmware validates `crc`, `mode_machine` and the Enable byte, not gains, so each of these publishes with a success envelope and leaves every joint at whatever stiffness was sent.

## The derived guard

This class has now been closed one file at a time twice, so `tests/drivers/test_sdk_gated_cells_carry_their_own_oracle.py` states it once: a node gated on the SDK probe has to import the SDK inside itself. The scan walks every suite, so the next gated cell is graded wherever it lands.

| # | mutation | fires | collected |
|---|---|---|---|
| g0 | control | 0 | 6 |
| g1 | the gain module's markers restored, scan derived | **1**, naming both classes | 6 |
| g2 | markers restored **and** the scan's universe hardcoded to today's modules | **0, silent** | 6 |
| g3 | scan hardcoded, markers not restored (g2's control) | 0 | 6 |
| g4 | the oracle test always answers yes, so the rule accepts anything | 2 | 6 |
| g5 | the scan narrowed until it finds nothing | **1**, the non-vacuity cell | 6 |
| g6 | the scan narrowed **and** the non-vacuity floor removed | **0, silent** | 6 |

g1/g2/g3 are the reason the inventory is derived rather than listed: a hardcoded list equal to today's set finds nothing and still looks like a working test. g5/g6 are the same argument for the non-vacuity floor. Because the tree has no violator left, the rule's own two outcomes are graded on constructed exemplars rather than on the corpus, with a meta-cell asserting both outcomes are reached.

## Gate

- `ruff check` and `ruff format --check` clean over `strands_robots tests tests_integ` (1653 files).
- `mypy strands_robots tests tests_integ`: **24 == 24** against a worktree of the merge base, normalized message sets identical in both directions, 0 attributable to either changed file.
- Paired run over an identical 451-file selection (all of `tests/drivers/`, all of `tests/tools/g1/`, and every suite that walks the tree, parses source, or reads docstrings), with the new file excluded from both trees: identical **21301 collected** and `20 failed, 21208 passed, 73 skipped` on each, the same 20 failing ids, `comm` empty in both directions. Of the 20, 17 need `awsiot`/`awscrt` (`find_spec` is `None` for both; declared in the `mesh-iot` extra, which CI installs) and 3 are the `lerobot`-from-source drift in `tests/training/test_lerobot*`.
- `tests/drivers/` whole directory: 658 passed / 3 skipped with the SDK present, 651 passed / 10 skipped with it hidden, 661 collected either way.
- No visual artifact: this changes which box grades an existing contract, and adds no policy, simulation, rendering, recording or asset behaviour. The measured tables above are the evidence.

## Follow-up, not in scope here

Four modules in `tests/drivers/` now need the stub and there is no `conftest.py`, so the registration list exists in two copies (`test_g1_control_loop` autouse, `test_g1_driver` opt-in) and this adds a third. Consolidating it is a refactor with no behaviour change, and it has to keep the opt-in shape, because two suites in that directory pin SDK-*absent* refusals that a directory-wide autouse fixture would make unreachable.

### DIFF
--- a/changelog.d/2850-g1-gain-wire-cells-graded-in-ci.md
+++ b/changelog.d/2850-g1-gain-wire-cells-graded-in-ci.md
@@ -1,0 +1,41 @@
+### Tests: the g1 per-joint gain wire cells are graded in CI, and the gating rule now holds for the module that states it
+
+``tests/drivers/test_g1_per_joint_gains`` is where this project writes the rule
+down: ``unitree-sdk2`` is not a declared dependency, so a contract asserted only
+behind ``skipif(not _HAS_SDK)`` is asserted by nothing in CI. Five of that
+module's own cells sat behind that marker, and they are the whole wire half of
+the gain contract -- every slot's gains on a full frame, a knee and an arm
+carrying different gains in one frame, the partial fallback where a caller tunes
+``kp`` and leaves ``kd``, a supplied gain overriding the table, and the stop
+frame staying soft. The module reported ``35 passed, 5 skipped`` where the SDK is
+absent, which is every runner ``call-test-lint`` uses, against ``40 passed``
+where it happens to be present.
+
+None of the five needs the real SDK. The module's own layering note explains why:
+the vendor gains are stated locally there precisely so the value cells are an
+independent oracle rather than a tautology, so what the cells read back is
+compared against those local tuples and not against anything the SDK computes.
+What they need is a ``LowCmd_``-*shaped* object to write into. They now take that
+shape from the stub ``tests.drivers.test_g1_control_loop`` installs, which is
+what puts them in front of CI. The fixture is opt-in per class rather than
+autouse, so an SDK-absent refusal cell added to this module later is not quietly
+made unreachable.
+
+The class had been closed one file at a time twice, so a derived guard is added
+beside the cells: every node gated on the SDK probe has to import the SDK inside
+itself. That is what an independent-oracle cell does to obtain its reference
+value, so the import is the evidence the marker is load-bearing rather than
+inherited. The two ``crc`` cells in ``tests/drivers/test_g1_driver`` keep their
+marker and satisfy the rule -- a stub whose ``Crc`` returns a constant would
+compare that constant against itself. The scan walks every suite, so the next
+gated cell is graded wherever it lands.
+
+Five real gain regressions were measured against the tree CI runs. Three of them
+were previously undetectable there: both tables read one slot's gains for every
+joint (the flat-default failure mode, which keeps the subscript syntax the
+source-reading contract cell looks for), the gain table leaking into the
+zero-torque stop frame so a stop lands stiff, and a supplied ``kp`` no longer
+winning over the table. The other two -- ``kp`` and ``kd`` each collapsed to a
+single value -- went from one cell to three. The firmware validates ``crc``,
+``mode_machine`` and the Enable byte, not gains, so each of those publishes with
+a success envelope and leaves the joints at whatever stiffness was sent.
--- a/tests/drivers/test_g1_per_joint_gains.py
+++ b/tests/drivers/test_g1_per_joint_gains.py
@@ -26,14 +26,23 @@ Layering, and why it is this way:
   them in CI, where the SDK is not installed.  ``unitree-sdk2`` is not a
   declared dependency of this project, so a contract asserted only behind
   ``skipif(not _HAS_SDK)`` is asserted by nothing in CI.
-* The wire cells, which do need the SDK to build a real ``LowCmd_``, confirm the
-  contract cells describe what actually lands on the topic.
+* The wire cells confirm the contract cells describe what actually lands on the
+  topic.  They need a ``LowCmd_``-*shaped* object to write into, which is not
+  the same as needing the SDK: the gains they read back are compared against the
+  vendor tuples stated locally below, so the oracle is those tuples rather than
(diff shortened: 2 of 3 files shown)


yinsong1986 (approved):
## Summary

Test-only PR that puts the five SDK-gated gain wire cells in `tests/drivers/test_g1_per_joint_gains.py` in front of CI by swapping the `skipif(not _HAS_SDK)` markers for the shared `unitree_sdk2py` stub (opt-in `usefixtures` per class, classes imported from `tests.drivers.test_g1_control_loop` rather than re-copied), and adds a derived guard (`tests/drivers/test_sdk_gated_cells_carry_their_own_oracle.py`) requiring every remaining SDK-gated node to import the SDK inside itself, so a gated cell whose marker is not load-bearing is named the hour it lands. The oracle argument holds up under inspection: the wire cells compare against vendor tuples stated locally in the module, so a `LowCmd_`-shaped stub does not weaken them, while the two `crc` cells in `test_g1_driver` legitimately keep their marker (a stub CRC would compare a constant against itself) and satisfy the new guard by importing `unitree_sdk2py.utils.crc` in their own bodies. Verified locally at the head commit on an SDK-less box: the two changed/added modules plus the neighbouring g1 suites report 153 passed / 2 skipped (the two crc cells, as intended), the guard's scan is non-vacuous (finds the 2 gated nodes) and exercises both of its own outcomes on constructed exemplars without false-positiving on its own string literals, and `ruff check` / `ruff format --check` are clean on both files.

## What's good

- The stub fixture follows AGENTS.md > Testing Patterns > "Restore a `sys.modules` entry you remove" (`monkeypatch.setitem` throughout), and reuses the single stub source instead of adding a fifth copy of the classes.
- Opt-in rather than autouse is deliberate and documented, so a future SDK-absent refusal cell in the module stays reachable.
- The guard is non-vacuous by construction (`_MINIMUM_GATED_NODES`), grades both outcomes on built source since the tree currently satisfies the rule, and its failure message names the remedy.
- Changelog fragment present per AGENTS.md PR workflow step 3; no production code touched.

PR #2851, opened 2026-08-27
fix(assets): let a caller resolve an asset path without fetching the corpus

cagataycali:
## The problem

`resolve_model_path` downloads whatever it cannot find on disk. That is the right
default for a caller about to load the model — a path whose meshes are absent is
useless to MuJoCo, so fetching them is the helpful thing.

It is the wrong default for a caller that *reports* on assets rather than loading
them, and there was no way to ask for the other one. `is_robot_asset_present`
answers **whether** an asset is on disk without a download; nothing answered
**where** under the same terms.

Two callers wanted it, and both had already noticed:

**`list_available_robots`** hand-rolled a guard for the first of the resolver's two
download triggers — `# Only resolve full path when asset is present - avoids
download attempts` — and still fetched for the second. A cached XML whose meshes
are missing passes the presence check, so building a status listing downloaded
once per such robot:

```
list_available_robots(), 63 rows, download hook instrumented
  empty cache  -> 0 attempts   (the presence check does stop this one)
  this cache   -> 4 attempts   aliengo, asimov_v0, jvrc, unitree_a1
```

**`tests/registry/test_asset_family_joint_counts.py`** re-derives the asset-family
grouping by walking all 72 registry entries, so it fetched every asset the machine
did not already have — which is exactly the case where there is no grouping to
re-derive:

```
_auto_download_robot attempts over the registry walk
  empty cache  -> 63 of 72
  this cache   ->  7
```

That is what exceeded the file's `120s` `pytest-timeout` budget on `5d64a915`.
That run's log records `58` × `attempting auto-download` and `11` real `Cloning`
operations, and the file's docstring claimed the opposite: *"It skips where the
assets are absent rather than downloading them."*

The two commits after it passed, so this is a latent timeout rather than a
permanent one - it fires whenever the fetches are slow enough. It is worth being
precise about what those green runs do and do not show: pytest only dumps a
captured log for a test that **fails**, and the download lines in the red run sit
inside its `Captured log call` section. A passing run therefore prints none of
them whether or not it fetched, so it is no evidence that it did not. What is
measurable either way is that the walk asks for the corpus, and how long that
takes is the network's business rather than the test's.

## The change

`resolve_model_path` takes a keyword-only `allow_download`, defined as **exactly a
download that declines**: an absent XML still returns `None`, and a mesh-less XML
still returns its first candidate — the two outcomes a failed download already
produces. It introduces no third outcome, so the 46 existing call sites keep the
downloading default unchanged, and a caller that only reports can opt out of the
network entirely. A parametrized test pins that equivalence on both triggers.

Both reporting callers now use it.

## A second finding, in the same test

The re-derivation never saw the real assets. `tests/registry/conftest.py` repoints
`STRANDS_ASSETS_DIR` at a per-test temp directory — correctly, so a developer's
user robots cannot leak into registry assertions — and that also empties the asset
cache, the one input this class needs. Under it the grouping was unconfirmable on
every machine, and the downloading default hid that by fetching the corpus *into
the temp directory* to build the oracle instead of reporting that there was nothing
to read.

Declining alone would therefore have traded a red test for a permanently skipping
one, silently deleting the protection the class exists for. It now reads the host
cache, so it gets the two outcomes its docstring describes: confirm where the
assets are present, skip where they are not. On a machine with the assets it
confirms all six families — coverage it never actually had.

## Declining changes no answer

Same walk, downloads allowed vs declined:

```
signatures compiled : 58  vs  58
derived families    :  6  vs   6      IDENTICAL: True
lost from signatures: none
```

All six declared families are still re-derived. Every registry robot still
resolves and loads through the unchanged default: `resolve_model_path("so101")`
returns a byte-identical path either way, and `Robot("so101", mode="sim")` builds a
`MuJoCoSimEngine`, runs a 60-step SmolVLA rollout (7.722 rad of joint travel) and
records a LeRobot dataset that reopens with all three
`observation.images.camera{1,2,3}` streams at `[3, 256, 256]`, 60 frames, 0 frozen
and 0 black frames.

## Tests

Each new cell fails before the change and passes after. Reverting only the two
gates (keeping the signature, so the failures are the behaviour rather than a
`TypeError`):

```
tests/registry/test_asset_family_joint_counts.py::TestTheRederivationReadsDiskOnly
  test_no_download_when_the_cache_is_empty     FAILED  tried to download 63 assets
  test_no_download_when_the_cache_is_partial   FAILED  tried to download 7 assets
  test_the_walk_asks_the_resolver_not_to_fetch PASSED  (grades the caller's source)

tests/test_asset_manager.py::TestDownloadCanBeDeclined
  test_absent_xml_reports_a_miss_without_attempting     FAILED  ['unitbot'] == []
  test_missing_meshes_return_the_xml_without_attempting FAILED  ['unitbot'] == []
  test_declining_matches_a_download_that_fails[both]    FAILED  ['unitbot'] == []
  test_the_downloading_default_is_unchanged             PASSED  (over-reach control)
```

Both download triggers are graded separately, because guarding on
`is_robot_asset_present` alone stops only the first — and the robots it leaves
fetchable include members of a graded family. The over-reach control pins that the
default still downloads, and the structural cell pins *why* the count is zero, so
a passing run cannot be an accident of what happens to be on disk.

## Gate

`ruff check` and `ruff format --check` clean over 1653 files. `mypy strands_robots
tests tests_integ` — `Success: no issues found in 1650 source files`.
`tests/test_asset_manager.py` + `tests/registry/` — 851 passed. A 501-file
selection of every guard that walks the tree, parses source, reads docstrings or
names the resolver — 24149 passed, 5 failed; those 5 assert `rclpy` is absent and
reproduce identically on a pristine `main` worktree on this machine, where it
resolves.

The re-derivation itself goes from a `>120s` timeout to 19 passed in 23s, with no
skips and no network - so its cost stops depending on how fast 63 fetches happen
to be.

### DIFF
--- a/strands_robots/assets/manager.py
+++ b/strands_robots/assets/manager.py
@@ -32,20 +32,36 @@ except ImportError:
     _auto_download_robot_impl = None  # type: ignore[assignment]
 
 
-def _lookup(name: str) -> dict | None:
+def _lookup(name: str, *, allow_discovery: bool = True) -> dict | None:
     """Resolve a registry entry for *name*, with ``robot_descriptions`` fallback.
 
     A curated ``robots.json`` entry always wins. When a name is unknown to the
     curated registry, fall back to :func:`strands_robots.registry.discovery`,
     which synthesizes an entry for any MJCF-capable ``robot_descriptions`` robot
-    (the long tail: ``go2``, ``spot``, ``h1``, ...). Discovery imports the
+    (the long tail: ``gen3``, ``iiwa14``, ``viper``, ...). Discovery imports the
     description module, which can trigger an asset download on first use, so
     this helper is used only by download-capable resolvers - never by the
     side-effect-free presence check.
+
+    Args:
+        name: Robot name (canonical or alias).
+        allow_discovery: When False, answer from the curated registry alone. A
+                         resolver whose caller declined the download must pass
+                         False: the discovery import *is* the fetch, because
+                         ``robot_descriptions`` calls ``clone_to_cache`` at
+                         module scope, so consulting it would perform the exact
+                         download the caller declined - including on the miss
+                         path, where the answer is ``None`` either way.
+
+    Returns:
+        A registry-style entry, or ``None`` when the name is unknown (and, with
+        ``allow_discovery=False``, when only discovery could have synthesized one).
     """
     info = get_robot(name)
     if info is not None:
         return info
+    if not allow_discovery:
+        return None
     from strands_robots.registry.discovery import discover_robot
 
     return discover_robot(name)
@@ -187,21 +203,31 @@ def resolve_model_path(
     ``robot_descriptions`` before returning - unless ``allow_download`` declines.
 
     ``allow_download=False`` makes this a pure filesystem lookup: no network, no
-    ``robot_descriptions`` import. It is exactly equivalent to a download that
-    declines, so it cannot change the answer for an asset already on disk - an
-    absent XML still returns ``None`` and a mesh-less XML still returns its first
-    candidate, the same two outcomes a failed download produces today.
+    ``robot_descriptions`` import. Within the curated registry it is exactly
+    equivalent to a download that declines, so it cannot change the answer for an
+    asset already on disk - an absent XML still returns ``None`` and a mesh-less
+    XML still returns its first candidate, the same two outcomes a failed download
+    produces today.
+
+    It declines the discovery fallback too, so a name only
+    :mod:`strands_robots.registry.discovery` could synthesize an entry for reports
+    a miss instead of importing a description module to find out. That import is
+    itself the fetch - ``robot_descriptions`` calls ``clone_to_cache`` at module
+    scope - so consulting discovery clones the upstream asset repository on a cold
+    cache, including on the miss path where the answer is ``None`` either way.
 
     Prefer it in any caller that reports on assets rather than loading them.
-    :func:`is_robot_asset_present` answers *whether* an asset is on disk without
-    a download; this answers *where* under the same terms.
+    :func:`is_robot_asset_present` answers *whether* an asset is on disk without a
+    download; this answers *where* on the same terms and over the same names -
+    both read the curated registry alone.
 
     Args:
         name: Robot name (canonical or alias).
         prefer_scene: If True, return scene XML (with ground/lights)
                       instead of bare model XML.
-        allow_download: When False, never attempt an asset download; resolve
-                        from what is already on disk or report a miss.
+        allow_download: When False, never attempt an asset download and never
+                        consult discovery; resolve from what is already on disk
+                        for a curated name, or report a miss.
 
     Returns:
         Path to the MJCF XML file, or None if not found.
@@ -212,7 +238,7 @@ def resolve_model_path(
         resolve_model_path("so100", prefer_scene=True)  # → .../trs_so_arm100/scene.xml
         resolve_model_path("franka")            # → .../franka_emika_panda/panda.xml
     """
-    info = _lookup(name)
+    info = _lookup(name, allow_discovery=allow_download)
     if not info or "asset" not in info:
         # DEBUG, not WARNING: resolve_model() probes several candidate names
         # (incl. suffix-stripped variants) and handles a None return cleanly,
--- a/strands_robots/assets/manager.py
+++ b/strands_robots/assets/manager.py
@@ -176,18 +176,32 @@ def is_robot_asset_present(name: str) -> bool:
 def resolve_model_path(
     name: str,
     prefer_scene: bool = False,
+    *,
+    allow_download: bool = True,
 ) -> Path | None:
     """Resolve a robot name to its MJCF model XML path.
 
     Looks up the robot in ``registry/robots.json``, then searches
-    the asset directories for the actual file.  If XML is found but
-    mesh files are missing, automatically downloads them via
-    ``robot_descriptions`` before returning.
+    the asset directories for the actual file.  If no XML is found, or XML is
+    found but mesh files are missing, downloads the asset via
+    ``robot_descriptions`` before returning - unless ``allow_download`` declines.
+
+    ``allow_download=False`` makes this a pure filesystem lookup: no network, no
+    ``robot_descriptions`` import. It is exactly equivalent to a download that
+    declines, so it cannot change the answer for an asset already on disk - an
+    absent XML still returns ``None`` and a mesh-less XML still returns its first
+    candidate, the same two outcomes a failed download produces today.
+
+    Prefer it in any caller that reports on assets rather than loading them.
+    :func:`is_robot_asset_present` answers *whether* an asset is on disk without
+    a download; this answers *where* under the same terms.
 
     Args:
         name: Robot name (canonical or alias).
         prefer_scene: If True, return scene XML (with ground/lights)
                       instead of bare model XML.
+        allow_download: When False, never attempt an asset download; resolve
+                        from what is already on disk or report a miss.
 
     Returns:
         Path to the MJCF XML file, or None if not found.
@@ -236,7 +250,7 @@ def resolve_model_path(
     # Search standard paths with traversal protection
     candidates.extend(_resolve_candidates(asset_dir_name, xml_file, name))
 
-    if not candidates:
+    if not candidates and allow_download:
         # No XML found at all - try auto-download, then re-search
         logger.info("No XML found for %s, attempting auto-download...", name)
         if _auto_download_robot(name, info):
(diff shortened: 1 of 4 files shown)


strands_robots/assets/manager.py
```diff
@@ -176,18 +176,32 @@ def is_robot_asset_present(name: str) -> bool:
...
 
     Looks up the robot in ``registry/robots.json``, then searches
-    the asset directories for the actual file.  If XML is found but
-    mesh files are missing, automatically downloads them via
-    ``robot_descriptions`` before returning.
+    the asset directories for the actual file.  If no XML is found, or XML is
+    found but mesh files are missing, downloads the asset via
+    ``robot_descriptions`` before returning - unless ``allow_download`` declines.
+
+    ``allow_download=False`` makes this a pure filesystem lookup: no network, no
```
yinsong1986, line 205: [MUST FIX] This contract — "pure filesystem lookup: no network, no ``robot_descriptions`` import" — is not honored for names outside the curated registry, and the gap is silent. The first statement of this function is `info = _lookup(name)`, which runs before `allow_download` is ever consulted, and `_lookup` falls back to `strands_robots.registry.discovery.discover_robot(name)` for any name `get_robot` doesn't know. `discover_robot` imports the description module, which per `discovery.py`'s own docstring "makes ``robot_descriptions`` clone the upstream asset repository on first use" and is to be called "only from asset-resolution paths that are allowed to download" — `_lookup`'s docstring says the same ("used only by download-capable resolvers"). So on a cold cache, `resolve_model_path("go2", allow_download=False)` clones the whole Menagerie asset repo and returns a path: the exact side effect the flag exists to decline, performed with nothing raised and nothing in the result distinguishing it from a disk hit. The two callers this PR converts happen to pass only curated names, which is why every new test passes without seeing it.

Why this can't be deferred: the keyword ships as public API in this PR with these documented semantics (and the docstring actively recommends it to reporting callers). Gating the discovery fallback later flips the observable result for a discoverable name from "clone + `Path`" to `None` — a breaking change to a documented contract once external callers exist. Before the flag ships, the same fix is free. That's the one-way-door bucket.

Resolution: don't let a declined resolve reach discovery — e.g. thread the flag through (`_lookup(name, allow_discovery=allow_download)`), or call `get_robot(name)` here and consult `discover_robot` only when `allow_download` is true, mirroring `is_robot_asset_present`, which uses `get_robot` directly for exactly this reason.

yinsong1986 (commented):
## Summary

Adds a keyword-only `allow_download=True` to `resolve_model_path`, defined as exactly equivalent to a download that fails (no third outcome), and converts the two reporting callers — `list_available_robots` and the asset-family re-derivation in `tests/registry/test_asset_family_joint_counts.py` — to decline the fetch. The equivalence claim is pinned on both download triggers, the downloading default is pinned unchanged by an over-reach control, and the family test additionally gains a structural (AST) cell so a passing run cannot be an accident of what happens to be on disk. One blocker inline: the flag's "no network, no `robot_descriptions` import" contract does not survive the `_lookup` → `discover_robot` fallback for non-curated names, and that gap becomes a breaking change to fix once the keyword ships.

## What's good

- The knob is keyword-only and defined by an equivalence to an existing outcome, so the 46 downstream call sites are untouched by construction, and `test_declining_matches_a_download_that_fails` pins that equivalence on both triggers.
- Changelog fragment present (`changelog.d/2851-...`), per AGENTS.md step 3.
- The `TestTheRederivationReadsDiskOnly` behavioural + structural pairing means the no-fetch property is graded even on a machine that already holds every asset.

## Verification suggestions

- After the discovery-fallback gap is addressed, spot-check with a name only discovery can resolve: on a machine with `robot_descriptions` installed and a cold cache, `resolve_model_path("go2", allow_download=False)` should return without importing `robot_descriptions.go2_mj_description` (assert via a `sys.modules` check), not clone the Menagerie repo.

strands_robots/assets/manager.py
cagataycali: Confirmed and fixed in `2d96820e`. The mechanism is exactly as described; one detail of the example differs, and chasing it made the finding sharper.

## The example, and the real population

`go2` is a curated entry here, so `_lookup` returns from `get_robot` and never reaches discovery — same for `spot` and `h1`. The population is the MJCF-discoverable names `get_robot` does **not** know:

```
curated entries                       : 72
MJCF-discoverable                     : 58
discoverable AND unknown to get_robot :  9
  gen3, iiwa14, mujoco_humanoid, n1, openarm_v1, rizon4, so_arm101, viper, widow
```

## Measured on those, before the fix

Spying on `robot_descriptions._cache.clone_to_cache` (patched before any description module binds it) rather than only on the import:

| `resolve_model_path(name, allow_download=False)` | returned | `clone_to_cache` |
|---|---|---|
| `gen3` | `.../kinova_gen3/gen3.xml` | `['mujoco_menagerie']` |
| `iiwa14` | `.../kuka_iiwa_14/iiwa14.xml` | `['mujoco_menagerie']` |
| `viper` | `.../trossen_vx300s/vx300s.xml` | `['mujoco_menagerie']` |
| `widow` | `.../trossen_wx250s/wx250s.xml` | `['mujoco_menagerie']` |
| `rizon4` | `None` | `['mujoco_menagerie']` |

The clone is reached on the **miss** path too — `rizon4` pays for the repository and then reports the `None` the caller would have had for free. `gen3_mj_description.py` calls `_clone_to_cache("mujoco_menagerie")` at module scope, so the import is the fetch, not merely a precursor to one.

## The sharper framing

The docstring offers this as the *where* to `is_robot_asset_present`'s *whether*, "under the same terms". For all nine names the two disagreed in **both** directions — it did more work *and* answered about more robots:

```
name    curated  is_robot_asset_present  declined resolve
gen3    False    False                   Path        <- before
gen3    False    False                   None        <- after
so101   True     True                    Path        (unchanged)
```

## The change

Your first resolution: `_lookup(name, *, allow_discovery=True)`, and `resolve_model_path` passes its own `allow_download`. The lookup keeps sole ownership of the curated-then-discovery ladder, so `resolve_model_dir` and `get_robot_info` are untouched by the default. After it, all nine return `None` with zero `clone_to_cache` calls, and the parity holds.

I narrowed three docstring claims to what the code now supports, rather than leaving the equivalence stated unconditionally: it is exact *within the curated registry*, and a name only discovery could synthesize an entry for now reports a miss instead of importing to find out.

## Grading

Eleven cells in `TestDecliningAlsoDeclinesDiscovery`. The seam is the fallback itself (`_lookup` imports `discover_robot` inside its body, so patching the module attribute is what the call site reads), so the contract is graded on a core install too; one cell re-asks the question of the real registry as an independent oracle, which is the `sys.modules` check you suggested.

Two cells first passed against the defect, because a synthesized entry pointing at nothing on disk returns `None` for the wrong reason. They now place the asset on disk under a name only discovery supplies — the shape the real registry has — so the declined `None` is the decision rather than an absence.

Behavioural revert of the gate alone: **5 failed / 43 passed**, 5 of 5 regression cells, 0 of 4 over-reach controls, 0 of 1 premise.

| mutation | fires | note |
|---|---|---|
| the gate removed | 5 | |
| the forward dropped | 6 | + the structural cell |
| forward keeps the keyword, hardcodes `True` | 6 | same set — the structural cell reads the keyword's *value* |
| the gate defaults closed | 2 | over-reach: both sibling resolvers |
| the gate inverted | 8 | over-reach, **and trips `test_resolve_model_path_uses_discovery_for_uncurated_robot`** |
| the gate placed after the consultation | 4 | |
| the gate placed after the module import | 0 | measured no-op: importing `registry.discovery` is cheap, only the call clones |

The inverted-gate row tripping that pre-existing test is what pins the boundary from the other side — the default must keep resolving the long tail.

## Gate

`ruff check` and `ruff format --check` clean over 1653 files. `mypy strands_robots tests tests_integ`: 24 errors on this branch and 24 on a pristine worktree of the merge base, normalized message sets identical in both directions, 0 attributable. The 398-file selection of every guard that walks the tree, parses source or names the resolver: **18 failed on both trees, identical failing-id sets, `comm` empty both ways** — 17 need `awsiot`/`awscrt` (declared in `[mesh-iot]`, absent here) and one is the lerobot-from-source drift. The `+20` collected localises to `test_asset_manager.py` (31 -> 48) and `test_asset_family_joint_counts.py` (16 -> 19).

`main` is merged in, so the head is current.

yinsong1986 (approved):
## Summary

Adds a keyword-only `allow_download` to `resolve_model_path`, defined as exactly equivalent to a download that fails (no third outcome), and converts the two reporting callers - `list_available_robots` and the asset-family re-derivation test - to decline the fetch. The prior review's must-fix (the discovery fallback in `_lookup` running before `allow_download` was consulted, so a non-curated name still triggered the module-scope `clone_to_cache`) is resolved at this head: the gate lives on the shared lookup (`_lookup(name, allow_discovery=allow_download)`), the declined path returns `None` cleanly when no candidates exist, and `TestDecliningAlsoDeclinesDiscovery` pins the fix behaviourally (miss path, parity with `is_robot_asset_present`, a real-registry oracle asserting no `robot_descriptions` module is imported) and structurally (the forward is graded by AST, so re-deriving the decision locally fails the test).

## What's good

- The equivalence claim ("declining is a download that fails") is tested on both download triggers, plus an over-reach control pinning that the default still fetches - so the 46 existing call sites are covered against accidental narrowing.
- Sibling resolvers (`resolve_model_dir`, `get_robot_info`) are explicitly pinned as unchanged, keeping the scope of the new keyword exactly where the PR says it is.
- The `host_asset_cache` fixture documents why it restores the host cache for the re-derivation class alone, and the download-attempt cells hold regardless of what that cache contains, so the zero cannot be an accident of disk state (the AST cell pins the reason).
- Changelog fragment per AGENTS.md step 3; new user-facing strings are ASCII; `Args:` entries cover the new parameters on both `resolve_model_path` and `_lookup`.

Issue #1810, opened 2026-07-31
ci(codeql): note-severity idiom alerts are merge-blocking, contradicting codeql.yml's stated intent - 32 of 32 examined instances are false positives


cagataycali:
## What

`.github/workflows/codeql.yml` states, in a comment on the `analyze` step:

> Findings appear in the Security tab; PRs are **not blocked** on CodeQL alerts (false-positive rate makes hard-fail noisy).

That intent is not what happens. It is defeated by an interaction the workflow cannot see:

1. `github-advanced-security` opens a **PR review thread** for each new alert on a PR.
2. The `default` branch ruleset sets `required_review_thread_resolution: true`.

So every new CodeQL alert on a PR *is* a hard merge gate, regardless of severity. The workflow comment describes a policy the repository does not implement.

## Measured cost

Two PRs were sitting APPROVED, with `call-test-lint / Test and Lint` green and `mergeable: MERGEABLE`, and were unmergeable purely on these threads:

| PR | blocking threads | rule | verdict |
|----|------------------|------|---------|
| #1110 | 2 | `py/import-and-import-from` | false positive |
| #1722 | 4 | `py/ineffectual-statement` | false positive |

Both were cleared by hand (alerts dismissed as false positive, threads resolved); #1110 has merged. That hand-clearing is the recurring cost this issue is about, not a one-off.

It is already recurring. There are **27 alerts dismissed by hand** on this repo, and the same two rules dominate them: 5x `py/ineffectual-statement`, 9x `py/import-and-import-from`. The same two classes are being swatted one alert at a time, per PR, indefinitely.

## The two rules, enumerated

271 open alerts. Three note-severity rules are 180 of them (66%): `py/empty-except` 89, `py/import-and-import-from` 64, `py/ineffectual-statement` 27.

**`py/ineffectual-statement` (27 open).** Every one is `...` used as a typing-construct body, which is the only way to write it:

- `typing.Protocol` method bodies (PEP 544) - `mesh/transport/base.py` x5, `policies/motionbricks/policy.py` x2
- `@abstractmethod` bodies - `simulation/base.py` x13
- `@overload` signatures - `robot.py:166,179,192`
- `TYPE_CHECKING` cross-mixin stubs so mypy accepts the call - `simulation/mujoco/randomization.py:68,69`, `simulation/mujoco/rendering.py:199`
- an `abc.ABC` stub inside a test - `tests/policies/lerobot_local/test_resolution.py:1388`

[...]

PR #2852, opened 2026-08-27
fix(policies/lerobot_local): emit the state-key-mismatch warning through the format sink

cagataycali:
## The problem

CodeQL alert 949 (`py/log-injection`, error severity) fires on `strands_robots/policies/lerobot_local/policy.py:2461` because the warning message it emits is built by concatenating strings that originate in the live observation dict:

```python
msg = (
    f"None of the configured robot_state_keys {shown}{ellipsis} are present "
    f"in the observation. Observed joint/state keys: {scalar_keys}. "
    + _state_key_cause(self.robot_state_keys)
    + " "
    + state_key_remedy(scalar_keys)  # <- reads from the observation
)
...
logger.warning(msg)  # <- taint sink
```

`scalar_keys` is the observation dict's own scalar keys, so under CodeQL's taint model the untrusted content reaches the log-format sink. It is a repo-wide `security-and-quality` gate, and the `default` branch ruleset requires the review thread it opens to be resolved before merge (that is the mechanic issue #1810 documented).

## The change

One line, at the warning call site: `logger.warning(msg)` → `logger.warning("%s", msg)`. The two shapes render *byte-identically* through any handler that has a formatter (`LogRecord.getMessage` computes `msg % args` when `args` is non-empty), so the operator-facing warning text is unchanged. The difference CodeQL grades is *where in the record the untrusted content lives*:

| Shape | `LogRecord.msg` | `LogRecord.args` | `.getMessage()` |
|---|---|---|---|
| `logger.warning(msg)` (pre) | full payload | `()` | full payload |
| `logger.warning("%s", msg)` (post) | `"%s"` | `(payload,)` | full payload |

CodeQL grades the pre shape as a `py/log-injection` sink (msg is the sink); the post shape moves the untrusted content into `args`, which is not a sink under the same rule.

A co-located comment names *why* the shape matters, so a future edit that reverts to the direct-substitution shape sees the reason at the call site rather than in a commit message.

## Tests

`tests/policies/lerobot_local/test_state_key_mismatch_log_injection.py`, three cells, graded at the `LogRecord` level rather than `.message`:

- **`test_msg_is_format_string_and_args_is_the_payload`** — the shape grader: `LogRecord.msg == "%s"` and `args == (payload,)`. Pre-fix: fails with an actionable diagnostic that names the received `msg`. Post-fix: passes. A revert to `logger.warning(msg)` re-opens the alert AND fails this cell.

- **`test_the_rendered_message_still_names_the_configured_and_observed_keys`** — over-reach control: pins that `.getMessage()` still names every observed joint (`shoulder_pan..gripper`) AND the configured (generic) key (`joint_0`), so a rewrite that keeps `msg == "%s"` but drops the payload from `args` fails this cell. Without this, the fix could be "delete the warning entirely" and the shape grader still passes.

- **`test_the_one_shot_spam_guard_still_fires_exactly_once`** — pins that the `_state_key_mismatch_warned` flag survives the shape edit, so a rewrite that accidentally moves the `logger.warning` call outside the guard fails this cell. A regression here would be a log-volume regression from `O(1)` to `O(steps)` on a long rollout.

### Regression check

`tests/policies/lerobot_local/test_state_key_mismatch.py` — **6/6 pass** unchanged. That file grades on `caplog.records[i].message`, which renders identically under both shapes.

### Pre-fix baseline (verified locally)

Reverting only the one line at `policy.py:2472` from `logger.warning("%s", msg)` back to `logger.warning(msg)`:

```
tests/policies/lerobot_local/test_state_key_mismatch_log_injection.py
  test_msg_is_format_string_and_args_is_the_payload                FAILED  (msg=<full payload>, expected '%s')
  test_the_rendered_message_still_names_the_configured_and_observed_keys  PASSED  (rendered text unchanged)
  test_the_one_shot_spam_guard_still_fires_exactly_once            PASSED  (guard unchanged)

= 1 failed, 2 passed =
```

Post-fix: **3/3 pass**. The failure diagnostic names the exact received `msg` value, so a future contributor who breaks the shape gets an actionable pointer.

## Not shipped, and why

The other open CodeQL alerts on `main` (from `gh api /repos/strands-labs/robots/code-scanning/alerts`):

- 5× `py/empty-except` (`mesh/core.py:1091`, `mesh/sensors.py:473`, `simulation/isaac/loaders.py:612`, `policies/lerobot_local/policy.py:960`, `simulation/mujoco/backend.py:398`) — note severity, all intentional soft-fallback catch sites.
- 1× `py/cyclic-import` (`simulation/policy_runner.py:3548`) — note severity, documented in the file's own comments as the accepted cost of avoiding a worse alert.
- 2× `py/unused-global-variable` and 1× `py/unused-import` — note severity, all in modules that use runtime-conditional imports.
- 1× `py/mixed-returns`, 1× `py/comparison-of-identical-expressions` — in test files, both intentional.

None of them are error-severity, and every one needs either a maintainer scope decision (widen `.github/codeql/codeql-config.yml`, which is pinned by `tests/test_codeql_query_filters.py` to exactly the two rule ids it currently excludes) or a semantics-changing edit. This PR ships only the single error-severity alert (949) — the one that gates every downstream merge until the review thread is resolved.

## Gate

The changelog fragment name (`2852-...`) is a best-effort predictor; happy to rename if this lands as a different number.

### DIFF
--- a/changelog.d/2852-lerobot-local-state-key-mismatch-log-format-sink.md
+++ b/changelog.d/2852-lerobot-local-state-key-mismatch-log-format-sink.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **policies/lerobot_local**: the state-key-mismatch warning now emits its message through the ``%s`` format sink (``logger.warning("%s", msg)``) rather than substituting the untrusted content into the format string. The observed warning text is byte-identical -- the ``args`` tuple carries the same payload the ``msg`` argument did -- but the observation-derived key names now live outside the log-format taint sink, so CodeQL's ``py/log-injection`` grades the call site clean without changing what a rollout's operator sees in a warning. Graded on the two shapes at the ``LogRecord`` level (``msg == "%s"``, ``args == (payload,)``) plus an over-reach control that pins the rendered message still names both the configured and observed keys.
--- a/strands_robots/policies/lerobot_local/policy.py
+++ b/strands_robots/policies/lerobot_local/policy.py
@@ -2456,9 +2456,20 @@ class LerobotLocalPolicy(Policy):
         # Surface the degraded binding as machine-detectable telemetry
         # (run_policy / eval_policy read generic_state_keys_used) and warn at
         # most once per policy so a long rollout is not spammed.
+        #
+        # The message is emitted through the ``%s`` format sink rather than
+        # substituted into the format string.  ``msg`` names the observation's
+        # own keys (``scalar_keys`` originates from a live observation dict),
+        # and the standard logging idiom is to keep untrusted values in the
+        # ``args`` position so a value that contains ``%s`` cannot re-enter
+        # the interpolation - the same rule CodeQL's ``py/log-injection``
+        # graduates on.  Under a live rollout this reads identically to a
+        # direct ``logger.warning("%s", msg)`` (both render the same string through
+        # the same handler); the difference is only observable to a caller
+        # who inspects ``LogRecord.args`` before formatting.
         self.generic_state_keys_used = True
         if not self._state_key_mismatch_warned:
-            logger.warning(msg)
+            logger.warning("%s", msg)
             self._state_key_mismatch_warned = True
         return drop_velocity_siblings(scalar_keys)
 
@@ -2531,7 +2542,7 @@ class LerobotLocalPolicy(Policy):
                 raise ValueError("strict_keys=True: " + msg)
             self.missing_state_keys_used = True
             if not self._state_missing_keys_warned:
-                logger.warning(msg)
+                logger.warning("%s", msg)
                 self._state_missing_keys_warned = True
         return values
 
--- a/tests/policies/lerobot_local/test_state_key_mismatch_log_injection.py
+++ b/tests/policies/lerobot_local/test_state_key_mismatch_log_injection.py
@@ -1,0 +1,183 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+"""The state-key-mismatch warning must not interpolate the observation.
+
+CodeQL's ``py/log-injection`` alert (rule id ``py/log-injection``, error
+severity, alert 949 on ``main = 10ed6913``) fires on
+:meth:`~strands_robots.policies.lerobot_local.policy.LerobotLocalPolicy._resolve_state_order`
+because the warning message it emits is built by concatenating strings that
+originate in the live observation dict (``scalar_keys`` is the observation's
+own scalar keys, and ``_state_key_cause`` and ``state_key_remedy`` are read
+from configured/observed keys respectively). The mitigation the ``logging``
+module documents is to keep the untrusted value in the ``args`` tuple:
+``logger.warning("%s", msg)`` rather than ``logger.warning(msg)``. The two
+call shapes render *identically* through any handler that has a formatter,
+because ``LogRecord.getMessage`` computes ``msg % args`` when ``args`` is
+non-empty, so the observable warning text is byte-identical. The difference
+CodeQL grades is *where in the record the untrusted content lives*: in
+``msg`` (a taint sink) or in ``args`` (not a sink).
+
+This module pins the format-sink shape.
+
+Two cells that grade the shape, one over-reach control:
+
+  * *args carries the payload* -- ``LogRecord.msg == "%s"`` and
+    ``LogRecord.args == (msg,)``. A revert to ``logger.warning(msg)`` fails
+    this cell because ``msg`` would then hold the full string and ``args``
+    would be empty.
+
+  * *the rendered text is unchanged* -- ``LogRecord.getMessage()`` still
+    contains every observed joint name and every configured (generic) key,
+    so a downstream reader that inspects ``.message`` reads identical bytes
+    under either shape. Grading this here makes sure the shape edit did not
+    accidentally regress the message content (which is what
+    ``test_state_key_mismatch.py`` grades separately -- this cell is the
+    local guard, so a future edit sees the two constraints together).
+
+  * *the one-shot spam guard still fires exactly once per policy* -- the
+    format-sink edit is co-located with the ``_state_key_mismatch_warned``
+    flag, so a rewrite that moves the ``logger.warning`` call outside the
+    guard would regress a rollout's log volume from ``O(1)`` to ``O(steps)``.
+    Grading it here surfaces that regression at the same point the shape
+    is graded.
+"""
+from types import SimpleNamespace
+from unittest.mock import patch
+
+import numpy as np
+import torch
+
+from strands_robots.policies.lerobot_local.policy import LerobotLocalPolicy
+
+
+NAMED_JOINTS = ["shoulder_pan", "shoulder_lift", "elbow_flex", "wrist_flex", "wrist_roll", "gripper"]
+
+
+def _visual(shape=(3, 224, 224)):
+    return SimpleNamespace(type=SimpleNamespace(name="VISUAL"), shape=shape)
+
+
+def _state(dim=6):
+    return SimpleNamespace(type=SimpleNamespace(name="STATE"), shape=(dim,))
+
+
+def _policy_with_generic_keys():
+    """A policy whose configured keys match no observation key.
+
+    The generic ``joint_0..N`` names are what the policy auto-fills when a
+    caller does not supply ``robot_state_keys`` and no embodiment hint
+    resolves; a NAMED observation (``shoulder_pan`` etc.) then triggers the
+    mismatch path. ``_load_model`` is patched so the constructor does not
+    reach the network.
+    """
+    with patch.object(LerobotLocalPolicy, "_load_model"):
+        policy = LerobotLocalPolicy(
+            pretrained_name_or_path=None,
+            policy_type="molmoact2",
+            strict_keys=False,
+        )
+    policy._input_features = {
+        "observation.images.base": _visual(),
+        "observation.state": _state(6),
+    }
+    policy._device = torch.device("cpu")
+    policy.robot_state_keys = [f"joint_{i}" for i in range(6)]
+    return policy
+
+
+def _named_obs():
+    obs: dict[str, object] = {"base": np.zeros((224, 224, 3), np.uint8)}
+    obs.update({name: 0.5 for name in NAMED_JOINTS})
+    return obs
+
+
+class TestArgsCarriesThePayload:
+    """The warning is emitted through the ``%s`` format sink."""
+
+    def test_msg_is_format_string_and_args_is_the_payload(self, caplog):
+        """``LogRecord.msg == "%s"`` and the payload lives in ``args[0]``.
+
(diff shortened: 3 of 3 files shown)


yinsong1986 (approved):
## Summary

Narrow, correctly-scoped fix for the one error-severity CodeQL alert (949, `py/log-injection`): the two state-key warning call sites in `LerobotLocalPolicy` now pass the observation-derived message through the `%s` format sink (`logger.warning("%s", msg)`) instead of substituting it into the format string, moving the untrusted content from `LogRecord.msg` into `LogRecord.args` while rendering byte-identically through `getMessage()`. Verified on the head commit: the new shape-grading test file and the existing `test_state_key_mismatch.py` suite pass together (9/9), the changelog fragment is present, and the diff touches nothing else. Note the second changed call site (`_collect_state_values`, the missing-keys warning) is the same class of sink and a sensible companion edit even though the description narrates only the mismatch site.

## What's good

- The regression test grades at the `LogRecord` level (`msg == "%s"`, `args == (payload,)`) rather than on the rendered text, so a revert re-fails it even though the rendered message is identical under both shapes.
- The over-reach control (rendered message still names both configured and observed keys) closes the "delete the warning" escape, and the spam-guard cell pins the one-shot `_state_key_mismatch_warned` behaviour through the edit.
- Changelog fragment per AGENTS.md step 3; no direct `CHANGELOG.md` edit.
- Scope discipline: the other open alerts on `main` are named and deliberately left out, with reasons.

Two non-blocking observations for the author's convenience (CI will surface the first): `ruff` flags the new test file (`I001` import-block ordering plus a missing blank line after the module docstring under `ruff format --check`), so `call-test-lint` will want a `hatch run format` pass; and the co-located comment in `_resolve_state_order` says the fixed shape "reads identically to a direct ``logger.warning(\"%s\", msg)``" where it means the *pre-fix* `logger.warning(msg)` -- worth correcting while the lint fix is being pushed, since the comment exists precisely to inform a future edit.

cagataycali:
## Measured: this change moves the sink rather than clearing it, and its own CodeQL run says so

The fix rests on one mechanism claim, stated in the code comment and again in the changelog fragment:

> CodeQL grades `msg` as a taint sink for `py/log-injection`; `args` is not a sink under the same rule.

That is not what the rule does. `args` is a sink too, and there are two independent measurements.

### 1. This branch's own CodeQL run re-flagged both sites it changed

`Analyze (Python)` completed on `bae02e8d` at 18:08:13Z and created two new alerts, at exactly the two lines this PR rewrote:

| ref | alert | line | sink columns | source there |
|---|---|---|---|---|
| `refs/heads/main` | 949 | 2461 | 28-31 | `logger.warning(msg)` |
| `refs/heads/main` | 714 | 2534 | 28-31 | `logger.warning(msg)` |
| `refs/pull/2852/merge` | 1104 | 2472 | 34-37 | `logger.warning("%s", msg)` |
| `refs/pull/2852/merge` | 1105 | 2545 | 38-41 | `logger.warning("%s", msg)` |

Columns 34-37 of line 2472 are `msg)` - the value in the `args` position. The alert did not clear; it followed the value.

Both new alerts opened a `github-advanced-security` review thread (`isResolved: false`, `isOutdated: false`) at `policy.py:2472` and `policy.py:2545`. With `required_review_thread_resolution: true` on the default ruleset those are a hard merge gate, so the net effect on this rule is two alerts tracked on `main` becoming two alerts that block this branch.

### 2. Six of the eight open alerts on `main` were already in the shape this PR adopts

`py/log-injection` has eight open alerts on `refs/heads/main`. Six of them already pass the untrusted value as an argument with a literal format string as `msg`:

| alert | site | shape | sink |
|---|---|---|---|
| 949 | `lerobot_local/policy.py:2461` | `logger.warning(msg)` | `msg` |
| 714 | `lerobot_local/policy.py:2534` | `logger.warning(msg)` | `msg` |
| 715 | `lerobot_local/policy.py:2622` | `logger.warning("Camera '%s' ... '%s'.", k, feat)` | `k`, an argument |
| 716 | `lerobot_local/policy.py:2715` | `logger.debug("VLA tokenized instruction: '%s...' -> %s tokens", instruction[:50], tokens.shape)` | `instruction[:50]`, an argument |
| 717 | `lerobot_local/policy.py:2910` | `logger.warning("Camera '%s' ... '%s'.", cam, feat)` | `cam`, an argument |
| 710 | `curobo/policy.py:789` | `logger.warning("... world_update=%r ignored", sorted(...))` | an argument |
| 711 | `curobo/policy.py:993` | `logger.warning("... observation.state=%r (%s); ...", state, e)` | `state`, an argument |
| 712 | `moveit2/policy.py:327` | `logger.warning("... observation.state=%r (%s); ...", state, e)` | `state`, an argument |

Alert 716 is the sharpest of these: it is already `logger.debug("<literal with %s>", <untrusted>)`, which is the exact shape this PR produces, and it is open at the same severity in the same file.

Three of them (710, 711, 712) go further and use `%r`, which does neutralise the newline this rule is about:

```
>>> "\n" in ("%s" % payload)
True
>>> "\n" in ("%r" % payload)
False
>>> "%r" % payload
'"configured=[\'a\'] observed=[\'b\']\\nWARNING evil forged line"'
```

So even rendering through `repr`, which removes the forged line break, leaves the alert open. The rule is not grading which argument position the value occupies.

### What is right in the change, and worth keeping

Two claims here hold, and I measured both:

- The two shapes render identically. `LogRecord("...", msg=payload, args=())` and `LogRecord("...", msg="%s", args=(payload,))` return the same string from `getMessage()` - `renders byte-identical: True`. The operator-facing text is unchanged, as stated.
- Keeping an untrusted value in `args` is better logging practice on its own merits: a payload containing `%s` cannot re-enter interpolation. That sentence in the comment is accurate and is a reason to make this edit regardless of CodeQL.

So the edit is not harmful. It just does not do the thing the fragment says it does, and the fragment is the part a future reader will act on.

### The test pins the shape the rule flags

`test_msg_is_format_string_and_args_is_the_payload` asserts `msg == "%s"` and `args == (payload,)`. After this run that is a shape carrying two open alerts, so the test would record it as the resolved form of a defect that is still open at both sites. The next person to see `py/log-injection` in this file has a passing test telling them the shape is already correct.

### Direction, without guessing at the remedy

CodeQL's own help for `py/log-injection` documents the fix as removing the line breaks before logging:

```python
name = name.replace('\r\n', '').replace('\n', '')
```

There is no such helper in the tree today. `simulation/safe_output.sanitize_name_component` is the only `sanitize*` function and it is a filesystem path-component guard, not a log sanitizer - it would reject or rewrite characters that belong in a diagnostic message.

That makes the real change a shared log-sanitizing helper applied at all eight sites across the three policy providers, which is a wider unit than this PR and worth its own scope decision. The alternative - deciding this rule is not worth eight sanitizer calls on operator diagnostics and dismissing the alerts - is a maintainer call rather than a code change. Either way the deciding fact is measurable, and the measurement above is that the argument position is not the axis.

I have not opened a competing change. Suggested next step for this branch: keep the two `logger.warning("%s", msg)` edits and the rendering-parity and spam-guard tests on their own merits, drop the `py/log-injection` claim from the fragment and the comment, and drop or rename the shape-grading test so it does not assert a resolution the rule has not granted. The two bot threads will not resolve on this branch as it stands.

Auto-merge is armed on this PR but cannot complete while those two threads are unresolved, so there is no urgency beyond the fragment text.

cagataycali:
Measured this branch at `bae02e8d` because `Run lint` failed and step 9 `Run tests` was **skipped** — which means the suite has not been graded by anyone. Two things worth having before the next push.

### The reported blocker is one auto-fixable rule, and `--fix` alone will not clear it

CI stops at the first lint command, so its log shows only half the problem:

```
cmd [1] | ruff check strands_robots tests tests_integ
I001 [*] Import block is un-sorted or un-formatted
  --> tests/policies/lerobot_local/test_state_key_mismatch_log_injection.py:43:1
Found 1 error.
[*] 1 fixable with the `--fix` option.
```

`ruff format --check` objects to the **same file** for a second, independent reason, and the lint step never reaches that command:

| command | before `--fix` | after `ruff check --fix` |
|---|---|---|
| `ruff check strands_robots tests tests_integ` | `Found 1 error` (I001) | clean |
| `ruff format --check strands_robots tests tests_integ` | `Would reformat: …test_state_key_mismatch_log_injection.py` | **still `Would reformat`** |

The residue is a missing blank line after the module docstring:

```diff
     Grading it here surfaces that regression at the same point the shape
     is graded.
 """
+
 from types import SimpleNamespace
```

So `ruff check --fix` on its own produces a second red round at the same step. `ruff check --fix && ruff format` on that one file clears both; afterwards `ruff format --check` reports `1656 files already formatted` with nothing outstanding.

### The suite CI skipped does pass

Because the lint step failed, step 9 never ran, so the 183-line test file this branch adds is ungraded on the merge queue. Running it here:

```
tests/policies/lerobot_local/test_state_key_mismatch_log_injection.py  3 passed in 0.10s
```

The contribution itself is sound — only the two formatting rules are holding it.

### One note on sequencing

This branch is currently `APPROVED`. `dismiss_stale_reviews_on_push` is active on the default branch, so the next push — including a formatting-only one — dismisses that approval and a fresh one is needed. That cost is unavoidable here (the branch cannot merge while the required check is red), but it is worth landing both formatting fixes in a **single** push rather than two, so only one approval is spent.

cagataycali:
Closing this rather than spending another round on it, because its own CI refuted the premise before it could merge, and #2855 supersedes it.

The description here claims that moving the observation-derived payload from `LogRecord.msg` into `LogRecord.args` takes it out of `py/log-injection`'s sink. `Analyze (Python)` on this head disagreed: it opened two new alerts (1104, 1105) at lines 2472 and 2545, which are exactly the two rewritten `logger.warning("%s", msg)` calls. The edit did not clear alert 949 - it moved it to a new line number under a new id, and brought the merge-gating review threads with it.

That is now written up in #2853 with the whole eight-alert census, and it holds in two more places that need no branch to check: alert 716 is already `logger.debug("<literal with %s>", <untrusted>)` and open, and 710/711/712 already render through `%r` - which *does* neutralise the break - and are also open. Neither the argument position nor `repr` rendering is the axis.

Measuring what the axis actually is changed the shape of the fix enough that a follow-up commit here would have been the wrong diff, not a smaller one. Driving all eight sinks and reading `LogRecord.getMessage()`: three could forge a record (bare `%s` on a bare `str` - two camera-name sinks and the instruction slice) and five could not, because their value passes through a container `repr` or `tolist()` first. This PR's two sites are in the second group, so its production delta was a no-op against both the rule and the defect, and its 183 test lines graded the record's field layout rather than the property that a record cannot split. Nothing worth carrying forward.

#2855 applies one sanitizer at all eight sinks and grades the property instead: three cells that fail on the pre-change tree by naming the record they still split, three that state the invariant at the sinks where it was only incidental, and one that holds the sanitized set to the census. Keeping the untrusted value in the `args` position is still better logging practice on its own merits and #2853 says so - it is just not a fix for this alert class, and should not ship claiming to be one.

